PR #5006, opened 2024-10-07
CI: add 3rd party tests to all LLVM 10-19 tests

certik:


certik:
The slowdown due to this is not too bad, about 3 min or so. We can speed up some of these tests later. For now it's good enough and we now thoroughly test all 3rd party codes with all LLVM versions.

Linux only, and only LFortran compiled in Debug mode, but that should be robust enough for now. The LFortran in Release mode skips some asserts and verify checks, so that should work, and we test macOS for LLVM 11 and 19, so while in theory things could break on macOS for some LLVM versions in between, the chance is low. 

We still run tests on all platforms with the actual conda packaging, which contains LFortran in release mode, and various LLVM versions (whatever is the default in conda-forge on a given platform), so overall I think we can reasonably guarantee that things will just work on all platforms.

We have to get the integration tests working on Windows, but we'll do it later.

PR #5009, opened 2024-10-08
force replace a FunctionCall when original target and --realloc-lhs is true

gxyd:
I'll add a test case once all the tests on CI pass, and I'll decided whether to add that test case in *simplifier_pass* branch or *main*


gxyd:
I'll undo the change of MRE from this PR once tests pass, as I've already created a PR for MRE against *main* branch here: https://github.com/lfortran/lfortran/pull/5012. This will allow us to have a smaller *diff* on *simplifier_pass* branch when merging into *main*.

gxyd:
Hopefully once https://github.com/lfortran/lfortran/pull/5010 is merged in *simplifier_pass* branch, the *expected checks* which never run will disappear.

gxyd:
E.g.; I hadn't noticed that those CI jobs aren't *required* for merging until you pointed it out.

certik:
The change looks good.

certik:
I added checks from master, and remove checks that were removed from master. You should synchronize the CI yaml file. Things should work now.

PR #5012, opened 2024-10-08
test: add MRE for dftatom as integration test case

gxyd:
this helps have the test case directly in *main* instead of *simplifier_pass* branch

### DIFF
--- a/integration_tests/binop_of_struct_instance_in_function_call.f90
+++ b/integration_tests/binop_of_struct_instance_in_function_call.f90
@@ -8,6 +8,7 @@ contains
 function func(x0)
     real(4), intent(in) :: x0(:)
     real(4) :: func(size(x0))
+    func = x0 + 1
 end function
 
 end module module_binop_of_struct_instance_in_function_call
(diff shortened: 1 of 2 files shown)


integration_tests/binop_of_struct_instance_in_function_call.f90
```diff
@@ -0,0 +1,24 @@
...
+
+program binop_of_struct_instance_in_function_call
+    use module_binop_of_struct_instance_in_function_call
+    implicit none
+    real(4), target :: A(2), B(2)
+    type(custom_data_type) :: d
+    real(4), dimension(2) :: tmp
+    d%x => A
+    d%y => B
+    tmp = func(d%x - d%y)
```
Pranavchiku, line 23: please test value of tmp by passing concrete values of `A` and `B`

Pranavchiku (changes requested):
I think it is fine, you may also test it by assigning values and then doing `if(condition) error stop `

integration_tests/binop_of_struct_instance_in_function_call.f90
gxyd: I've improved the test case and added an `if (...) error stop` condition as well with it.

certik (approved):

gxyd:
I checked the updated test case as well, that it works with *simplifier_pass* branch.

PR #4923, opened 2024-09-26
CI: pin micromamba to version '1.5.10-0'

certik:
Fixes #4919.

PR #2971, opened 2023-12-13
cmake: bring back relative paths

junghans:
For https://github.com/lfortran/lfortran/pull/2932#discussion_r1420713619

PR #5013, opened 2024-10-08
CI: Use micromamba 2.0 via setup-micromamba 2

jjerphan:
Follow-up of #4923.

Fix https://github.com/mamba-org/mamba/issues/3475.


certik:
Thanks for trying it. It looks like there are some issues with the newer version of setup-micromamba.

jjerphan:
micromamba 2.0.4 should have fixed all the issues, but one might need to clean the caches so that it gets pulled and used instead.

Could you clean the caches and restart the workflow when you have some time? Thank you.

certik:
@jjerphan thanks for fixing it!

How do I clean up caches?

jjerphan:
On GitHub UI, they are accessible here:

https://github.com/lfortran/lfortran/actions

alternatively one can use `gh` to access them and to remove them. For instance, one can remove all of them using:

```
gh cache list --limit 1000 | tail -n +5 | awk '{print $1}' | tr '\n' '\0' | xargs -P20 -0 -n1 gh cache delete
```

Some integration problem on `setup-micromamba` might need to be addressed first though (see https://github.com/mamba-org/setup-micromamba/pull/246 and https://github.com/mamba-org/setup-micromamba/pull/247).

certik:
On this page: https://github.com/lfortran/lfortran/actions/caches?query=branch%3Amain, I can see 18 results. Do I remove all of them?

jjerphan:
Removing caches is not required actually: the last version of `setup-micromamba` solves it.

The last failure is unrelated: a few adaptations to the "Check Out-of-Source Debug build" workflow are required since paths changed.

For instance, this environment specification:

https://github.com/lfortran/lfortran/blob/7eb6c7cecec92ca83c6ef091bb9da0947d54d82f/.github/workflows/CI.yml#L332

is not there (anymore).

Interestingly this workflow passes on `main` (potentially due to some cache?. I believe someone from lfortran might be able to update this workflow better and faster than I can. :slightly_smiling_face:

certik:
@jjerphan anything you tell me I am happy to do. Why don't we disable caches in this PR, get all tests to pass, merge with main, and only then re-enable caches?

certik:
Ah I see, the file `src/ci/environment_linux.yml` does not exist. This change was done in #2971.

So this only works with
```yaml
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
          path: src
```

However it seems this change is still there, here is the relevant part:
```yaml
      - uses: actions/checkout@v4
        with:           
          fetch-depth: 0
          path: src                            
                                                              
      - uses: mamba-org/setup-micromamba@v1.8.0
        with:                                       
          environment-file: src/ci/environment_linux.yml
          create-args: >-      
            python=3.10                     
```
So could it be that micromamba changed how it handles these path changes?

.github/workflows/CI.yml
```diff
@@ -38,9 +38,8 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         with:
-          micromamba-version: '1.5.10-0'
```
certik, line 43: This version pin was introduced in #4923 to ensure we are not affected by a failure if upstream changes the micromamba version. So I think we should keep it.

.github/workflows/CI.yml
```diff
@@ -323,12 +320,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          path: src
```
certik, line 326: Why is this change needed?

jjerphan:
> So could it be that micromamba changed how it handles these path changes?

Let me try to revert the "Check Out-of-Source Debug build" workflow to its previous state to use micromamba 1.5.10-0, addressing your review comments at the same time.

certik:
Everything passes there is just one failure left: https://github.com/lfortran/lfortran/actions/runs/12035248619/job/33553754028?pr=5013
```
Error: Could not determine environment name from file src/ci/environment_linux.yml
Error: Error: Error: ENOENT: no such file or directory, open 'src/ci/environment_linux.yml'
Error: If your environment file is not a YAML file containing `name` at the top level, please specify the environment name directly.
Error: ENOENT: no such file or directory, open 'src/ci/environment_linux.yml'
```

jjerphan:
OK, so the problem is not related to micromamba 2.0, but rather to paths which have changed as first mentioned in https://github.com/lfortran/lfortran/pull/5013#issuecomment-2500282250.

.github/workflows/CI.yml
jjerphan: Done in c9e92e5.

certik:
It says:
```

Invalid workflow file: .github/workflows/CI.yml#L332
The workflow is not valid. .github/workflows/CI.yml (Line: 332, Col: 11): 'micromamba-version' is already defined
```

.github/workflows/CI.yml
```diff
@@ -908,10 +906,9 @@ jobs:
             make=4.3
             nodejs=18.20.2
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         if: "!(contains(matrix.llvm-version, '10') || contains(matrix.llvm-version, '19'))"
         with:
-          micromamba-version: '1.5.10-0'
```
certik, line 914: This probably should be kept.

certik:
It looks good. It seems that micromamba 2.0 doesn't support the `path: src` that micromamba 1.5.10 supported? The diff seems as simple as it can get. I am not familiar what exactly the `path: src` does, but if it is not supported anymore, then we'll have to migrate away from it.

jjerphan:
> It seems that micromamba 2.0 doesn't support the path: src that micromamba 1.5.10 supported?

Sorry is has been quite confusing. But this is unrelated to micromamba but to nonexistent files, see https://github.com/lfortran/lfortran/pull/5013#issuecomment-2500282250.

(For details, eddf355 shows that the problem also exists with 1.5.10.)

certik:
micromamba 1.5.10 works in `main`, so whatever is causing the failure somehow doesn't affect `main`. Just upgrading micromamba from 1.5.10 to 2.0 makes the out-of-source CI fail. So we need to resolve this one way or the other before we can upgrade. Do you have a recommendation what we should do?

jjerphan:
Let's close this PR and continue the discussions without confusion of any of us on https://github.com/lfortran/lfortran/pull/5552.

PR #5016, opened 2024-10-08
Mod intrinsic 

AnshMehta1:


certik:
You committed a binary file into the PR. Please be extremely careful with this. Go ahead and remove it from history as well, so that we don't accidentally merge it.

certik:
@AnshMehta1 do you want to fix this issue: https://github.com/lfortran/lfortran/issues/5018, and then test it on this PR to ensure it catches it?

certik:
Important: you need to remove it from the history, not just a commit, otherwise we cannot merge it, only squash. However, let's fix #5018, and ensure it can catch it here.

AnshMehta1:
> You committed a binary file into the PR. Please be extremely careful with this. Go ahead and remove it from history as well, so that we don't accidentally merge it.

I removed the file but was not able to delete its history. Anyways , I opened this PR only for reporting changes that I made. I will open another PR to finaly fix the issue.

AnshMehta1:
> Important: you need to remove it from the history, not just a commit, otherwise we cannot merge it, only squash. However, let's fix #5018, and ensure it can catch it here.

Surely , I will try this.

HarshitaKalani:
```console
git reset --soft HEAD~1
git restore --staged .
```
Using these commands, you can undo the number of commits you want. Here `1` is used to undo 1 commit.

certik:
@assem2002 I think it works! Thank yo for adding this check.

@AnshMehta1 go ahead and fix this PR now, you can remove the binary from history.

AnshMehta1:
> @assem2002 I think it works! Thank yo for adding this check.
> 
> @AnshMehta1 go ahead and fix this PR now, you can remove the binary from history.

Surely , I will do that shortly

Issue #4593, opened 2024-08-05
Intrinsics TODO


certik:
* [ ] https://github.com/lfortran/lfortran/issues/4398
* [ ] https://github.com/lfortran/lfortran/issues/4591 (`program_01.f90`)
* [ ] `arrays_intrin_02.f90`
* [ ] `any_01.f90`

Thirumalai-Shaktivel:
Are the tests failing in the main branch?

HarshitaKalani:
Yes

harperjf:
Another intrinsic that seems to need fixing is ```transfer```. I reported the problem in #3961.

harperjf:
A problem that may or may not be in the intrinsic ```huge``` is #3906. I added a comment to it today. It did hit me in actual code.
So did #3917 where the trouble with ```execute_command_line``` arose in a subroutine. It did not arise when that intrinsic was called in this 2-line main program in a Linux system.
``` 
  call execute_command_line('whoami')
end program
```

certik:
Another TODO:
* [ ] remove `runtime/builtin/lfortran_intrinsic_builtin.f90`
* [ ] remove `runtime/custom/lfortran_intrinsic_custom.f90`
* [ ] remove `runtime/impure/lfortran_intrinsic_math.f90`
* [ ] remove `runtime/pure/lfortran_intrinsic_string.f90`

What will stay:
```
lfortran_intrinsic_ieee_arithmetic.f90
lfortran_intrinsic_iso_c_binding.f90
lfortran_intrinsic_iso_fortran_env.f90
```
And I think that's it!

harperjf:
Other things TODO are 
(1) implementing all standard-comforming Fortran statements 
(2) implementing everything that iso_fortran_env and ieee_arithmetic should provide  
(3) implementing all the edit descriptors valid in formats 
(4) improving error messages so that users can see which lines in their Fortran programs caused the trouble (I don't know how to get that info from hexadecimal local addresses in conda-root or libc.so.6 .) 
And one day (5) allowing more than 2 real kinds. All my other Fortran compilers have either 3 or 4.

certik:
> Other things TODO are
> (1) implementing all standard-conforming Fortran statements

What are the ones that are missing? We should have a pretty good support for all Fortran statements. If you know some that we missed, let us know. This a separate issue from intrinsics.

> (2) implementing everything that iso_fortran_env and ieee_arithmetic should provide

Yes, good point. Here are our current implementations:

[...]

PR #5017, opened 2024-10-08
Implement `len` intrinsic and remove it from old runtime lib

HarshitaKalani:
Towards: #4593 

Some of the tests will fail. I'm working on those.

### DIFF
--- a/a.f90
+++ b/a.f90
@@ -1,0 +1,29 @@
+program string_04
+    implicit none
+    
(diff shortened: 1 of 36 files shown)


HarshitaKalani:
`len` is implemented with the help of a node and I think that implementation is good enough. I've removed the old runtime implementation of `len` in https://github.com/lfortran/lfortran/pull/5023. Hence closing this.

PR #4461, opened 2024-07-08
Handle optional args for `minloc` and `maxloc`

HarshitaKalani:
Towards: #4017, #4456
Fixes: https://github.com/lfortran/lfortran/issues/4411

Issue #4456, opened 2024-07-05
ASR for minloc (optional arguments) design


certik:
```
    | IntrinsicArrayFunction(int arr_intrinsic_id, expr* args, int overload_id, ttype type, expr? value)
```
Minloc has the following arguments:
```
minloc(array, dim, mask, kind, back)
```
Of these:
* `kind` is encoded in `ttype type` in ASR.
*  The `back` argument could possibly be always present (true/false), but since it must be part of `expr* args`, so it would require us to create `LogicalConstant`, so it's some overhead for a rarely used argument.

I think we want to use `overload_id=0` for the most common case `minloc(array)`.

We have two options:
* Either `overload_id=1` for all other cases, and encode all arguments explicitly, even if they are not used, say `back=LogicalConstant`, and if `mask` is present, it is an integer expression, and if it is not present, we could use `LogicalConstant(true/false)` or something like that.
* Or we use:
  *  `overload_id=1` for minloc(array, dim)
  *  `overload_id=2` for minloc(array, mask)
  *  `overload_id=3` for minloc(array, back)
  *  `overload_id=4` for minloc(array, dim, mask)
  *  `overload_id=5` for minloc(array, dim, back)
  *  `overload_id=6` for minloc(array, mask, back)
  *  `overload_id=7` for minloc(array, dim, mask, back)
* Or some combination of the above two options

HarshitaKalani:
Here’s an alternative design I’ve implemented using bitwise flags to represent the optional arguments, which I believe is cleaner and more scalable than the previous approach.
PR: https://github.com/lfortran/lfortran/pull/8006

PR #5019, opened 2024-10-08
fix: optional arguments in findLoc()

### LINKED ISSUE #4679: `gfortran` treats the next logical arg as `mask` instead of `dim`
RESULT = FINDLOC(ARRAY, VALUE, DIM [, MASK] [,KIND] [,BACK])

```fortran
program main
    logical :: x(3) = [.true., .false., .true.], y = .true., back = .true.
    print *, findloc(x, y, back)
end program
```
```console
$ gfortran a.f90 && ./a.out
           1
$ lfortran a.f90
semantic error: `dim` should be a scalar integer type
 --> a.f90:3:14
  |
3 |     print *, findloc(x, y, back)
  |              ^^^^^^^^^^^^^^^^^^^ 


```

jinangshah21:
Fixes #4679.
Still not Complete. I am keeping it in draft.

### DIFF
--- a/integration_tests/intrinsics_301.f90
+++ b/integration_tests/intrinsics_301.f90
@@ -22,7 +22,10 @@ program intrinsics_301
     logical :: mask2(3) = [.true., .false., .true.]
     logical :: back1 = .false.
     logical :: back2 = .true.
-    
+
+    ! findloc(Array, value, mask)
+    logical :: x4(3) = [.true., .false., .true.], y4 = .true., mask3 = .true.
+
     print *, findloc(x1, y1)
     if (any(findloc(x1, y1) /= 3)) error stop
     print *, findloc(x2, y2)
@@ -43,7 +46,9 @@ program intrinsics_301
     if (any(findloc(x2, y2, mask = mask1, back = back2) /= 5)) error stop
     print *, findloc(x1, y1, mask = mask1, back = back2, kind = 8)
     if (any(findloc(x1, y1, mask = mask1, back = back2, kind = 8) /= 5)) error stop
-    
+    print *, findloc(x4, y4, mask3)
+    if (any(findloc(x4, y4, mask3) /= 1)) error stop
+
     print *, i1
     if (i1(1) /= 2) error stop
     print *, i2
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -3364,9 +3364,16 @@ namespace FindLoc {
     ASR::expr_t *mask = args[3];
     ASR::expr_t *back = args[5];
     if (overload_id == 1) {
+        ASR::expr_t *mask_new = nullptr;
+        if( ASRUtils::is_array(ASRUtils::expr_type(mask)) ){
+            mask_new = ArrayItem_02(mask, i);
+        }
+        else{
+            mask_new = mask;
+        }
         body.push_back(al, b.Assignment(result, b.i_t(0, return_type)));
         body.push_back(al, b.DoLoop(i, b.i_t(1, type), UBound(array, 1), {
-            b.If(b.And(b.Eq(ArrayItem_02(array, i), value), b.Eq(ArrayItem_02(mask, i), b.bool_t(1, logical))), {
+            b.If(b.And(b.Eq(ArrayItem_02(array, i), value), b.Eq(mask_new, b.bool_t(1, logical))), {
                 b.Assignment(result, i),
                 b.If(b.NotEq(back, b.bool_t(1, logical)), {
                     b.Return()
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -3236,18 +3236,29 @@ namespace FindLoc {
         int dim = 0, kind = 4; // default kind
         ASR::expr_t *dim_expr = nullptr;
         ASR::expr_t *mask_expr = nullptr;
-        if (args[2]) {
-            dim_expr = args[2];
-            if ( !ASR::is_a<ASR::Integer_t>(*expr_type(args[2])) ) {
-                append_error(diag, "`dim` should be a scalar integer type", loc);
+        
+        // Checking for type findLoc(Array, value, mask)
+        if( args[2] && !args[3] && is_logical(*expr_type(args[2])) ){
+            dim_expr = nullptr;
+            mask_expr = args[2]; 
+        }
+        else { 
+            dim_expr = args[2]; 
+            mask_expr = args[3]; 
+        }
+       
+        if (dim_expr) {
+            if ( !ASR::is_a<ASR::Integer_t>(*expr_type(dim_expr)) ) {
+                dim = ASR::down_cast<ASR::IntegerConstant_t>(dim_expr) -> m_n;
+                append_error(diag, "`dim` should be a scalar integer type", dim_expr->base.loc);
                 return nullptr;
-            } else if (!extract_value(expr_value(args[2]), dim)) {
-                append_error(diag, "Runtime values for `dim` argument is not supported yet", loc);
+            } else if (!extract_value(expr_value(dim_expr), dim)) {
+                append_error(diag, "Runtime values for `dim` argument is not supported yet", dim_expr->base.loc);
                 return nullptr;
             }
             if ( dim < 1 || dim > n_dims ) {
                 append_error(diag, "`dim` argument of `findloc` is out of "
-                    "array index range", loc);
+                    "array index range", dim_expr->base.loc);
                 return nullptr;
             }
             if ( n_dims == 1 ) {
@@ -3264,7 +3275,7 @@ namespace FindLoc {
                     result_dims.push_back(al, tmp_dim);
                 }
             }
-            m_args.push_back(al, args[2]);
+            m_args.push_back(al, dim_expr);
         } else {
             ASR::dimension_t tmp_dim;
             tmp_dim.loc = args[0]->base.loc;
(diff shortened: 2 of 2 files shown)


certik:
Add a test, so that we can see what exact features you are trying to implement.

Pranavchiku (changes requested):
Check if mask is an array then only create an array item

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3236,18 +3236,29 @@ namespace FindLoc {
         int dim = 0, kind = 4; // default kind
         ASR::expr_t *dim_expr = nullptr;
         ASR::expr_t *mask_expr = nullptr;
-        if (args[2]) {
-            dim_expr = args[2];
-            if ( !ASR::is_a<ASR::Integer_t>(*expr_type(args[2])) ) {
-                append_error(diag, "`dim` should be a scalar integer type", loc);
+        
+        // Checking for type findLoc(Array, value, mask)
+        if( args[2] && !args[3] && is_logical(*expr_type(args[2])) ){
```
gxyd, line 3241: The problem with an approach like this is that a similar problem exists for a lot of intrinsic functions, like: `minval`, `maxval`, `findloc` etc.

E.g.; documentation of GFortran here for `maxval`: https://gcc.gnu.org/onlinedocs/gfortran/MAXVAL.html with syntax:
```f90
Syntax:
RESULT = MAXVAL(ARRAY, DIM [, MASK])
RESULT = MAXVAL(ARRAY [, MASK])
```

So, a similar check might be needed for a lot of other intrinsic functions as well, where the `dim` and `mask` are assigned depending on "position" and "type".

With the approach described above, it would lead to a lot of repeated code and hence a possibility to  miss something or cause a bug.

I ideally would be more happy with a general approach here.

integration_tests/intrinsics_301.f90
```diff
@@ -43,7 +46,9 @@ program intrinsics_301
     if (any(findloc(x2, y2, mask = mask1, back = back2) /= 5)) error stop
     print *, findloc(x1, y1, mask = mask1, back = back2, kind = 8)
     if (any(findloc(x1, y1, mask = mask1, back = back2, kind = 8) /= 5)) error stop
-    
+    print *, findloc(x4, y4, mask3)
+    if (any(findloc(x4, y4, mask3) /= 1)) error stop
```
gxyd, line 50: Also, we don't have a test case for when `dim` i.e. scalar is passed as an argument, there are test cases for when `dim` is passed as a keyword-argument though.

gxyd: i.e. something like:
```f90
findloc(x4, y4, 1)
```

src/libasr/pass/intrinsic_array_function_registry.h
gxyd: @HarshitaKalani thoughts?

HarshitaKalani: I agree completely. I would also like to have a general approach for this. 
Also, the current approach will have high chances of bugs. For example, the current implementation would not handle a case like:
```fortran
program main
    print *, findloc([1, 2, 3], 2, .true., .true.)
end program
```
```console
$ gfortran a.f90 && ./a.out
a.f90:2:35:

    2 |     print *, findloc([1, 2, 3], 2, .true., .true.)
Error: ‘dim’ argument of ‘findloc’ intrinsic at (1) must be INTEGER
$ lfortran a.f90
  Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2668, in main
    return main_app(argc, argv);
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2611, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
...

  File "/home/harshita/Desktop/lfortran/src/libasr/pass/intrinsic_array_function_registry.h", line 3252, in LCompilers::ASRUtils::FindLoc::create_FindLoc(Allocator&, LCompilers::Location const&, LCompilers::Vec<LCompilers::ASR::expr_t*>&, LCompilers::diag::Diagnostics&)
    dim = ASR::down_cast<ASR::IntegerConstant_t>(dim_expr) -> m_n;
AssertFailed: is_a<T>(*f)
```

certik: The right design in my opinion is to use the "overload_id" in ASR for this. We discussed exactly this issue for a similar function, I can't remember which one. You decide on the signatures that you will support, let's just say the following:
```
minloc(array) ! overload_id = 0
minloc(array, dim) ! overload_id = 1
minloc(array, dim, mask) ! overload_id = 2
```
And for each case (`overload_id`) all arguments are required and never null (it could be that some arguments have to have their default values specified). So you never write code like this:
```
 if( args[2] && !args[3] && is_logical(*expr_type(args[2])) 
```
Rather, you write code like `if (overload_id == 2) { /* array, dim, mask is present and valid, handle this case */ }`.

Let me know if you have any questions about this, or want to discuss more.

HarshitaKalani: Yes, we did it for `minloc` here: https://github.com/lfortran/lfortran/issues/4456. If we follow this, the implementation would be more clean and robust.
One can refer: https://github.com/lfortran/lfortran/pull/4461.

certik: Yes, that was it, thanks for finding it. @jinangshah21 go ahead and follow the same design here.

jinangshah21: > I agree completely. I would also like to have a general approach for this. Also, the current approach will have high chances of bugs. For example, the current implementation would not handle a case like:
> 
> ```fortran
> program main
>     print *, findloc([1, 2, 3], 2, .true., .true.)
> end program
> ```
> 
> ```
> $ gfortran a.f90 && ./a.out
> a.f90:2:35:
> 
>     2 |     print *, findloc([1, 2, 3], 2, .true., .true.)
>       |                                   1
> Error: ‘dim’ argument of ‘findloc’ intrinsic at (1) must be INTEGER
> $ lfortran a.f90
> Internal Compiler Error: Unhandled exception
> Traceback (most recent call last):
>   Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
>   File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
>   File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
>   File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2668, in main
>     return main_app(argc, argv);
>   File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2611, in main_app(int, char**)
>     err = compile_src_to_object_file(arg_file, tmp_o, false,
> ...
> 
>   File "/home/harshita/Desktop/lfortran/src/libasr/pass/intrinsic_array_function_registry.h", line 3252, in LCompilers::ASRUtils::FindLoc::create_FindLoc(Allocator&, LCompilers::Location const&, LCompilers::Vec<LCompilers::ASR::expr_t*>&, LCompilers::diag::Diagnostics&)
>     dim = ASR::down_cast<ASR::IntegerConstant_t>(dim_expr) -> m_n;
> AssertFailed: is_a<T>(*f)
> ```

GFortran Gives same error. I had tried all cases with Gfortran, only case with findloc(array, value, mask) was not implemented. So do we have to implement cases that Gfortran not support?

jinangshah21: > The right design in my opinion is to use the "overload_id" in ASR for this. We discussed exactly this issue for a similar function, I can't remember which one. You decide on the signatures that you will support, let's just say the following:
> 
> ```
> minloc(array) ! overload_id = 0
> minloc(array, dim) ! overload_id = 1
> minloc(array, dim, mask) ! overload_id = 2
> ```
> 
> And for each case (`overload_id`) all arguments are required and never null (it could be that some arguments have to have their default values specified). So you never write code like this:
> 
> ```
>  if( args[2] && !args[3] && is_logical(*expr_type(args[2])) 
> ```
> 
> Rather, you write code like `if (overload_id == 2) { /* array, dim, mask is present and valid, handle this case */ }`.
> 
> The general design of ASR is that we do not have defaults or optional arguments. Everything is always present and explicit. So there is no branching logic and trying to figure out the defaults, like you discovered in this PR. All these decisions are made by the frontend. Once ASR is constructed, it is already set it stone what exact operations we need to do.
> 
> Let me know if you have any questions about this and if you know how to apply this to this PR, or want to discuss more.

By Checking `if( args[2] && !args[3] && is_logical(*expr_type(args[2])) ` I am differentiating between overload_id for (Array, value, dim)(i.e. 1) and (Array, value, mask)(i.e. 2). So it uses overload_id already.

certik: Only do cases that GFortran supports.

I see, this is in the `create` function, which determines the overload_id. In that case this is fine I think!

certik (approved):
I think this is fine then. It fixes an issue, test is added.

The concern is how to expand this to other functions without repeating too much code.

I would say we can merge it, and then figure out a general solution if there is any.

@gxyd, @HarshitaKalani let me know what you want to do here.

Pranavchiku (approved):
I think we will always have this problem to adjust variables. The approach uses overload_id just swaps two arguments as one is nullptr and other is not. The only way to generalise this is to create some form of mapping in the frontend for each intrinsic so that we do not have to worry about it in `create_*` function.

certik:
The `create` function is called by the frontend, isn't it? It's not called after ASR is constructed.

Pranavchiku:
Yes, it is called in frontend.

gxyd:
> I would say we can merge it, and then figure out a general solution if there is any.

I agree with this.

> @gxyd, @HarshitaKalani let me know what you want to do here.

We can maybe open a separate issue describing in detail about the general issue.

PR #5022, opened 2024-10-09
test: add MRE for dftatom as integration test case

gxyd:
### Why this test case?

this tests to make sure that when a StructInstanceMember is passed as an argument to a subroutine call (with intent(out)), then the changes are indeed made to the StructInstanceMember directly

### MRE

Initially reported here https://github.com/lfortran/lfortran/issues/4761#issuecomment-2401815359


certik (approved):
Looks great. Good test case.

Issue #5018, opened 2024-10-08
CI: check that no binary file was added


certik:
To avoid: https://github.com/lfortran/lfortran/issues/3589.

To do that, let's take a function to detect a binary file, such as:
```python
def is_file_binary(file_path: str) -> bool:
    try:
        with open(file_path, 'r') as fp:
            fp.read(16)
            return False
    except UnicodeDecodeError:
            return True
```
Then:
* run it on all files in the checked out repository
* iterate on all commits of a given branch / PR to catch cases when a binary was added and removed in a subsequent commit
* run it at the CI automatically

PR #5025, opened 2024-10-10
CI : Add test for checking if there's a pushed binary file in git commit history

assem2002:
closes https://github.com/lfortran/lfortran/issues/5018

I don't know if it should be a dedicated job to test for binary files. Anyways, I tried to not interrupt the flow of the CI and did a checkout back to latest commit when the test succeeds. maybe we can call github action `checkout@v4` again right after running the test.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,6 +14,17 @@ env:
   MACOSX_DEPLOYMENT_TARGET: 14.0
 
 jobs:
+  Check_for_Added_Binary_Files:
+    name: Check for Added Binary Files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          
+      - name: Check for Added Binary Files
+        run: python3 check_binary_file_in_git_history.py
+
   Build:
     name: LFortran CI (${{ matrix.python-version }}, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
@@ -33,12 +44,6 @@ jobs:
           environment-file: ci/environment.yml
           create-args: >-
             python=${{ matrix.python-version }}
-
-      - name: Check Any Added Binary File
-        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
-        run: | 
-          python3 check_binary_file_in_git_history.py
-            
 
       - name: Install Windows Packages
         if: contains(matrix.os, 'windows')
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Check Any Added Binary File
         if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
         run: | 
-          python3 Check_binary_file_in_git_history.py
+          python3 check_binary_file_in_git_history.py
             
 
       - name: Install Windows Packages
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,6 +34,12 @@ jobs:
           create-args: >-
             python=${{ matrix.python-version }}
 
+      - name: Check Any Added Binary File
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
+        run: | 
+          python3 Check_binary_file_in_git_history.py
+            
+
       - name: Install Windows Packages
         if: contains(matrix.os, 'windows')
         shell: bash -e -l {0}
--- a/check_binary_file_in_git_history.py
+++ b/check_binary_file_in_git_history.py
@@ -1,47 +1,64 @@
 import subprocess
 
-def is_file_binary(path: str) -> bool:
+# Open file_path, Read it with mode 'r' which expects a utf-8 encode.
+# If the file is binary it'd conflict with
+# text mode 'r', leading to raised error `UnicodeDecodeError`.
+def is_file_binary(file_path: str) -> bool:
     try:
-        with open(path, 'r') as fp:
+        with open(file_path, 'r') as fp:
             fp.read(16)
             return False
     except UnicodeDecodeError:
             return True
-    
-def get_commit_hashes_PR_against_main() -> list : # returns list of strings
-    completedProcess_object =  subprocess.run(["git" ,"rev-list" ,"origin/main..HEAD"], stdout=subprocess.PIPE, text=True)
+
+
+# Get all commit hashes difference between `origin/main` and current `HEAD`
+def get_commit_hash_differences() -> list :
+    # Run command `git rev-list origin/main..HEAD`
+    # Use `subporcess.PIPE` as the output buffer to fetch the list of commit hashes
+    completedProcess_object =  subprocess.run(["git" ,"rev-list" ,"origin/main..HEAD"],
+                                                stdout=subprocess.PIPE
+                                                , text=True)
+    # Create list out of the resulting commit hases. 
     commit_hashes_listed  : list  = completedProcess_object.stdout.split("\n")
-    if(commit_hashes_listed[-1] == ""): commit_hashes_listed.remove("")
-    print("number of commit from current PR branch against origin/main is :", len(commit_hashes_listed))
-    print (completedProcess_object.stdout)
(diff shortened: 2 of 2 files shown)


check_binary_file_in_git_history.py
```diff
@@ -0,0 +1,47 @@
...
+    file_paths : list = completedProcess_object.stdout.split("\n")
+    if(file_paths[-1] == "" ): file_paths.remove("")
+    return file_paths
+
+
+
+def run() :
+    commit_hashes_list = get_commit_hashes_PR_against_main()
+
+    for commit_hash in commit_hashes_list:
```
gxyd, line 30: I think this is nice. We don't want a binary file commit-ed in any of the commits, even if in one commit we add the binary file and the next one we undo that in the same PR, as if we *merge and commit* the PR, then we do have the binary file in our commit history.

So, I think this approach is good. It would be better maybe to have some simple comment about something for this?

check_binary_file_in_git_history.py
```diff
@@ -0,0 +1,47 @@
...
+def get_commit_hashes_PR_against_main() -> list : # returns list of strings
+    completedProcess_object =  subprocess.run(["git" ,"rev-list" ,"origin/main..HEAD"], stdout=subprocess.PIPE, text=True)
+    commit_hashes_listed  : list  = completedProcess_object.stdout.split("\n")
+    if(commit_hashes_listed[-1] == ""): commit_hashes_listed.remove("")
+    print("number of commit from current PR branch against origin/main is :", len(commit_hashes_listed))
+    print (completedProcess_object.stdout)
+    return commit_hashes_listed
+
+def get_changed_files() -> list : # returns list of strings.
+    completedProcess_object = subprocess.run(["git","diff","--name-only","origin/main", "--diff-filter=d"], stdout=subprocess.PIPE, text=True)
```
gxyd, line 20: Maybe we can wrap long lines so that no line is too long?

.github/workflows/CI.yml
```diff
@@ -34,6 +34,12 @@ jobs:
           create-args: >-
             python=${{ matrix.python-version }}
 
+      - name: Check Any Added Binary File
```
gxyd, line 37: A better name could be? *Check for Added Binary File(s) in Pull Request*

check_binary_file_in_git_history.py
assem2002: Okay. I'll to try polish the script and some comments as you suggested.
Do you think I better push some binary file in this PR to test whether the script does it work or not?

gxyd:
For a PR like this I would definitely like to see if it works with an example for it. (Below is a suggestion of how this probably can be achieved):

You merge this PR in your own *main* (i.e. https://github.com/assem2002/lfortran) and then create a PR against your own repository (i.e. against *main* of https://github.com/assem2002/lfortran), with the PR containing a binary file, and if the CI fails on that PR, then we can be a little certain that this actually works.

Let me know if there is a simple way to test the whole thing.

check_binary_file_in_git_history.py
gxyd: You can do that as well actually, at the end you would need to just do a force-push to remove the binary file from commit history altogether.

gxyd:
> For a PR like this I would definitely like to see if it works with an example for it. (Below is a suggestion of how this probably can be achieved):
> 
> You merge this PR in your own _main_ (i.e. https://github.com/assem2002/lfortran) and then create a PR against your own repository (i.e. against _main_ of https://github.com/assem2002/lfortran), with the PR containing a binary file, and if the CI fails on that PR, then we can be a little certain that this actually works.
> 
> Let me know if there is a simple way to test the whole thing.

I thin your comment here https://github.com/lfortran/lfortran/pull/5025#discussion_r1794776253 could be a nice solution if we do a force-push at the end I think.

assem2002:
Yeah that's much better. I'll wrap your suggestions, then will push 2 commits (one pushing a binary file, one reverting it) into another branch an submit a PR.

assem2002:
> > For a PR like this I would definitely like to see if it works with an example for it. (Below is a suggestion of how this probably can be achieved):
> > You merge this PR in your own _main_ (i.e. https://github.com/assem2002/lfortran) and then create a PR against your own repository (i.e. against _main_ of https://github.com/assem2002/lfortran), with the PR containing a binary file, and if the CI fails on that PR, then we can be a little certain that this actually works.
> > Let me know if there is a simple way to test the whole thing.
> 
> I thin your comment here [#5025 (comment)](https://github.com/lfortran/lfortran/pull/5025#discussion_r1794776253) could be a nice solution if we do a force-push at the end I think.

I'll use your approach, so we can make sure the 2 branches have identical changes (the python script file + CI.yml) by observing the commit hashes.

Pranavchiku:
> For a PR like this I would definitely like to see if it works with an example for it. (Below is a suggestion of how this probably can be achieved):
> 
> You merge this PR in your own _main_ (i.e. https://github.com/assem2002/lfortran) and then create a PR against your own repository (i.e. against _main_ of https://github.com/assem2002/lfortran), with the PR containing a binary file, and if the CI fails on that PR, then we can be a little certain that this actually works.
> 
> Let me know if there is a simple way to test the whole thing.

OR simply push a commit like above :)

Pranavchiku:
It fails as expected: https://github.com/lfortran/lfortran/actions/runs/11272405282/job/31347255418?pr=5025

Pranavchiku (commented):
This PR is good, I agree with Gaurav, let's polish the script and add it as a separate CI check, we don't need it for all just `ubuntu-latest` will work. 
TODO:
* [ ] Apply code review
* [ ] Merge this PR
* [ ] Send a PR with binary against main to assure it works

assem2002:
Thanks guys for the feedback if anyone could give quick feedback on some needed comments or clarifications about the python script, please inform me.

assem2002:
Fails here #5030

check_binary_file_in_git_history.py
```diff
@@ -0,0 +1,64 @@
...
+    for commit_hash in commit_hashes_list:
+        subprocess.run(["git","checkout",commit_hash], 
+                       stdout=subprocess.PIPE,stderr=subprocess.PIPE,text=True) 
+        print("checked out to commit --> ",commit_hash)
+
+        file_paths : list = get_changed_files() # fetch changed files from current-checkout-to commit.
+        for file_path in file_paths:
+            if (is_file_binary(file_path)):
+                raise SystemExit(f"ERROR : FOUND PUSHED BINARY FILE --> ({file_path}) IN COMMIT ({commit_hash})")
+
```
certik, line 56: Put `print("OK --- no binary file found")` here.

check_binary_file_in_git_history.py
```diff
@@ -0,0 +1,64 @@
...
+    # Run git `checkout` to each commit_hash between `origin/main` --> `HEAD`
+    for commit_hash in commit_hashes_list:
+        subprocess.run(["git","checkout",commit_hash], 
+                       stdout=subprocess.PIPE,stderr=subprocess.PIPE,text=True) 
+        print("checked out to commit --> ",commit_hash)
+
+        file_paths : list = get_changed_files() # fetch changed files from current-checkout-to commit.
+        for file_path in file_paths:
+            if (is_file_binary(file_path)):
+                raise SystemExit(f"ERROR : FOUND PUSHED BINARY FILE --> ({file_path}) IN COMMIT ({commit_hash})")
```
certik, line 55: And here put a print statement above the throw with `print("FAIL --- binary file detected")`.

certik (approved):
Left two comments, otherwise I think it's good to merge and we can iterate on this as needed.

certik:
We should also check that this PR can find the binary file in https://github.com/lfortran/lfortran/pull/5016.

certik:
Let's merge this?

assem2002:
I've applied your code changes. It should be ready to get merged.

gxyd:
Currently "CI / Check for Added Binary Files (pull_request)" job isn't marked as "Required", maybe that needs to be done from your end @certik? 

That would ensure that we can't merge the PR if that CI job fails.

gxyd (approved):
LGTM!

certik:
Thanks! Let me enable the test to be required now.

PR #5031, opened 2024-10-11
Add libunwind as a dependency

### LINKED ISSUE #4980: test_llvm.cpp fails for LLVM 17 - 19
```console
(lf19) ~/.../lfortran/lfortran(main)$ ctest
Test project /home/ondrej/repos/lfortran/lfortran
    Start 1: test_cwrapper
1/3 Test #1: test_cwrapper ....................   Passed    0.00 sec
    Start 2: test_stacktrace
2/3 Test #2: test_stacktrace ..................   Passed    0.04 sec
    Start 3: test_lfortran
3/3 Test #3: test_lfortran ....................***Exception: SegFault  0.08 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =   0.13 sec

The following tests FAILED:
          3 - test_lfortran (SEGFAULT)
Errors while running CTest
Output from these tests are in: /home/ondrej/repos/lfortran/lfortran/Testing/Temporary/LastTest.log
(lf19) ~/.../lfortran/lfortran(main)$ ./src/lfortran/tests/test_lfortran
/home/ondrej/repos/lfortran/lfortran/src/lfortran/tests/test_llvm.cpp:23:
TEST CASE:  llvm 1

/home/ondrej/repos/lfortran/lfortran/src/lfortran/tests/test_llvm.cpp:23: FATAL ERROR: test case CRASHED: SIGSEGV - Segmentation violation signal

[doctest] test cases: 16 | 15 passed | 1 failed | 79 skipped
[doctest] assertions: 57 | 57 passed | 0 failed |
[doctest] Status: FAILURE!
```

advikkabra:
Towards supporting LLVM >= 12.
Fixes #4980


certik:
Let's enable the `ctest` test, to ensure this fixes it?

Thirumalai-Shaktivel:
It works fine on macOS, now.

advikkabra:
Is there something else I have to change to get the Docker test to pass?

Thirumalai-Shaktivel:
I think you need to make modifications in Dockerfile in the home directory.

Either download libunwind-dev or add it the environment_linux.yml. Or Use the same yml file to create conda env in DockerFile as you do the CI.yml

certik:
Can you please enable the CI test for "ctest" on LLVM > 17 to ensure they now work? We need to ensure this change actually fix the problem.

certik:
Unfortunately it looks like this fix didn't fix the problem for LLVM >=17.

Adding a new dependency is a big deal, so I am hoping to avoid it. Especially when it doesn't actually fix our problem, I think we should not add it.

advikkabra:
I'll look into why it doesn't work for LLVM >= 17.

advikkabra:
@Thirumalai-Shaktivel  Is this PR ready now?

Thirumalai-Shaktivel:
Yes, I think so. I'm trying to figure out any other way to fix this seg fault.

Thirumalai-Shaktivel:
@certik, do you know which `unwind.h` was used here?

https://github.com/lfortran/lfortran/blob/969922035020eb4abaa3c5dbc06b19d271e587cc/src/libasr/stacktrace.cpp#L32

certik (approved):
This is currently the only way we know how to get everything (including `ctest`) working for LLVM 10 - 19. Given that, the highest priority is to get `ctest` working at the CI, so I think this is good enough to merge as is.

There are two bugs:

* unwind dependency needed (sometimes)
* LLVM >= 17 fix for pointers

The second fix is ok. The first fix to add unwind as a dependency is unsatisfactory, since we need to keep the list of our dependencies small, and not arbitrarily add more, and it is also not clear when exactly this dependency must be added. 

But given that this at least gives us a path forward, I think it's ok to merge, and investigate this after merging.

PR #5035, opened 2024-10-12
Raise unsuccessful systemExit

assem2002:


assem2002:
I've missed a critical point. I didn't raise an unsuccessful system exit call. This PR corrects that.

certik:
Thanks!

PR #5037, opened 2024-10-12
fix: isnan() supported for compile time values

### LINKED ISSUE #4548: `isnan` does not work for compile time values
```fortran
program main
    logical, parameter :: l1 = isnan(-1.0)
    print *, l1
end
```
```console
$ lfortran a.f90
$ gfortran a.f90 && ./a.out
 F
```

jinangshah21:
Fixes #4548.

### DIFF
--- a/integration_tests/intrinsics_233.f90
+++ b/integration_tests/intrinsics_233.f90
@@ -1,6 +1,7 @@
 program intrinsics_233
     implicit none
     real :: x
+    real :: z = -1.0
     real :: y(3)
     x = -1.0
     x = sqrt(x)
@@ -18,4 +19,11 @@ program intrinsics_233
     print*, y(3)
     print *, isnan(y(3))
     if (isnan(y(3)) .neqv. .true.) error stop
+
+    !tests for compile time argument
+    print *, isnan(1.0)
+    if (isnan(1.0) .neqv. .false.) error stop
+    print *, isnan(sqrt(z))
+    if (isnan(sqrt(z)) .neqv. .true.) error stop
+
 end program
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -6055,6 +6055,18 @@ public:
                     } else {
                         tmp = create_func(al, x.base.base.loc, args, diag);
                     }
+                    if(var_name == "isnan" && are_all_args_evaluated){
+                        ASR::IntrinsicElementalFunction_t* intr_func = ASR::down_cast<ASR::IntrinsicElementalFunction_t>(ASRUtils::EXPR(tmp));
+                        ASR::expr_t * m_value = intr_func->m_value;
+                        double rv = ASR::down_cast<ASR::RealConstant_t>(m_value)->m_r;
+                        if(rv == 0){
+                            m_value = ASRUtils::EXPR(ASR::make_LogicalConstant_t(al, x.base.base.loc, 0, intr_func->m_type));
+                        } else{
+                            m_value = ASRUtils::EXPR(ASR::make_LogicalConstant_t(al, x.base.base.loc, 1, intr_func->m_type));
+                        }
+                        intr_func->m_value = m_value;
+                        tmp = (ASR::asr_t*)intr_func;
+                    }
                 } else if ( ASRUtils::IntrinsicArrayFunctionRegistry::is_intrinsic_function(var_name) ) {
                     if(var_name == "dot_product"){
                         ASR::expr_t *matrix_a = args[0], *matrix_b = args[1];


integration_tests/intrinsics_233.f90
```diff
@@ -1,6 +1,7 @@
 program intrinsics_233
     implicit none
     real :: x
+    real :: z = -1.0
```
Pranavchiku, line 4:
```suggestion
    real, parameter :: z = -1.0
```

Pranavchiku: can you please check if this works? if you are assigning to `z` somewhere in the body, please create a new variable `real, paramete :: xx = -1.0` and use it.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6055,6 +6055,18 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     } else {
                         tmp = create_func(al, x.base.base.loc, args, diag);
                     }
+                    if(var_name == "isnan" && are_all_args_evaluated){
```
Pranavchiku, line 6058: the problem is that we do not have any runtime logic to compute if it is nan or not, we call the `c` runtime library. Check `lfortran_intrinsic.c` and there will be a `C` function for `isnan`.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6055,6 +6055,18 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        ASR::IntrinsicElementalFunction_t* intr_func = ASR::down_cast<ASR::IntrinsicElementalFunction_t>(ASRUtils::EXPR(tmp));
+                        ASR::expr_t * m_value = intr_func->m_value;
+                        double rv = ASR::down_cast<ASR::RealConstant_t>(m_value)->m_r;
+                        if(rv == 0){
+                            m_value = ASRUtils::EXPR(ASR::make_LogicalConstant_t(al, x.base.base.loc, 0, intr_func->m_type));
+                        } else{
+                            m_value = ASRUtils::EXPR(ASR::make_LogicalConstant_t(al, x.base.base.loc, 1, intr_func->m_type));
+                        }
```
Pranavchiku, line 6066: I am not exactly sure what gets stored in ASR when you do pass `NaN`. We need to check it

integration_tests/intrinsics_233.f90
jinangshah21:
``` diff
semantic error: Argument of `sqrt` has a negative argument
  --> integration_tests/intrinsics_233.f90:26:20
   |
26 |     print *, isnan(sqrt(z))
   |                    ^^^^^^^ 
```
on declaring z as parameter.

src/lfortran/semantics/ast_common_visitor.h
jinangshah21: https://github.com/lfortran/lfortran/blob/969922035020eb4abaa3c5dbc06b19d271e587cc/src/libasr/pass/intrinsic_functions.h#L403

This is compile time eval function used for isnan and it returns a realConstant, so if realconstant is 0 that value is not NaN else it is NaN.

src/lfortran/semantics/ast_common_visitor.h
jinangshah21: I am checking only for compile time value.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: The approach seems correct, but I would prefer to do it inside `intrinsic_functions.h`

Pranavchiku: @HarshitaKalani can you please check and help in here

integration_tests/intrinsics_233.f90
Pranavchiku: Yes, that is expected, this means the check you are doing by passing `z` is runtime and not compile time.

HarshitaKalani: I don't this there is a way to pass nan compile time here and that's why it's tricky to implement isnan() compile time values.

src/lfortran/semantics/ast_common_visitor.h
HarshitaKalani: @Pranavchiku does Fortran has a way to use nan in code? If not, how can we test isnan for compile time?

integration_tests/intrinsics_233.f90
HarshitaKalani: Can you think of a program where we're passing a compile time value to isnan and it return True?

jinangshah21: For isnan(0.0/0.0), 
GFortran gives : 
```
(lf) jinu2110@JNSLAP:/mnt/c/lfortran$ gfortran a.f90
a.f90:11:20:

   11 |   result = isnan(0.0/0.0)
      |                    1
Error: Division by zero at (1)
```
LFortran gives true at compile time.

src/lfortran/semantics/ast_common_visitor.h
jinangshah21: https://github.com/lfortran/lfortran/blob/969922035020eb4abaa3c5dbc06b19d271e587cc/src/libasr/pass/intrinsic_functions.h#L403

Here The only issue is eval_##X returns a realConstant and by explicitly checking function name we can return LogicalConstant but how to check function name. The Second way is to have different namespace for Isnan.

integration_tests/intrinsics_233.f90
jinangshah21: After applying my changes, I got true for isnan(0.0/0.0).

HarshitaKalani: That's nice. Let's confirm if we want the error, otherwise it's good (it's giving true for a compile time value).

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: Check this integration test: https://github.com/lfortran/lfortran/blob/93dc209708f900481a05487466e43beba9d43321/integration_tests/nan_handling_01.f90#L4

jinangshah21: > Check this integration test:
> 
> https://github.com/lfortran/lfortran/blob/93dc209708f900481a05487466e43beba9d43321/integration_tests/nan_handling_01.f90#L4

Here also isnan(b) is giving true. So I think Logic is correct but it can be moved to `intrinsic_functions.h` by creating a separate eval function for Isnan.

Pranavchiku (changes requested):
```diff
diff --git a/src/libasr/pass/intrinsic_functions.h b/src/libasr/pass/intrinsic_functions.h
index 9f2f55e19..93fede97b 100644
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -417,13 +417,25 @@ namespace X {
     }                                                                           \
 } // namespace X
 
+namespace Isnan {
+    static inline ASR::expr_t *eval_IsNan(Allocator &al, const Location &loc,     
+            ASR::ttype_t *t, Vec<ASR::expr_t*> &args,                           
+            diag::Diagnostics& /*diag*/) {                                      
+        double rv = ASR::down_cast<ASR::RealConstant_t>(args[0])->m_r;          
+        ASRUtils::ASRBuilder b(al, loc);                                        
+        return b.bool_t(std::isnan(rv), t);                                       
+    }
+
+    // complete instantiate
+}
+
 create_unary_function(Trunc, trunc, trunc)
 create_unary_function(Gamma, tgamma, gamma)
 create_unary_function(LogGamma, lgamma, log_gamma)
 create_unary_function(Log10, log10, log10)
 create_unary_function(Erf, erf, erf)
 create_unary_function(Erfc, erfc, erfc)
-create_unary_function(Isnan, isnan, is_nan)
+// create_unary_function(Isnan, isnan, is_nan)
 
 namespace ObjectType {   ` 
```

Pranavchiku (approved):

PR #5039, opened 2024-10-13
Remove arr printing in pass `print_arr`

### LINKED ISSUE #4523: `print_arr` pass prints array with format element by element not by `do loop`
It prints it element by element + creating integerop hierarchy of adding 1. It should get replaced with do loop. 
```Fortran
PROGRAM example
    ! integer :: i
    character ::arr(10000)  
    arr = '*'
    10 FORMAT ("whatever", 6(A1)) 
    print 10,  arr 
 END PROGRAM example
```

```clj
(Print
                        [(StringFormat
                            (StringConstant
                                "\"whatever\", 6(A1)"
                                (Character 1 17 ())
                            )
                            [(ArrayItem
                                (Var 2 arr)
                                [(()
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )
                            (ArrayItem
                                (Var 2 arr)
                                [(()
                                (IntegerBinOp
                                    Add
                                    (Integer 4)
                                    ()
                                )
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )
                            (ArrayItem
                                (Var 2 arr)
                                [(()
                                (IntegerBinOp
                                    (IntegerBinOp
                                        Add
                                        (Integer 4)
                                        ()
                                    )
                                    Add
                                    (Integer 4)
                                    ()
                                )
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )
                            (ArrayItem
                                (Var 2 arr)
                                [(()
                                (IntegerBinOp
                                    (IntegerBinOp
                                        (IntegerBinOp
                                            Add
                                            (Integer 4)
                                            ()
                                        )
                                        Add
                                        (Integer 4)
                                        ()
                                    )
                                    Add
                                    (Integer 4)
                                    ()
                                )
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )
                            (ArrayItem
                                (Var 2 arr)
                                [(()
                                (IntegerBinOp
                                    (IntegerBinOp
                                        (IntegerBinOp
                                            (IntegerBinOp
                                                Add
                                                (Integer 4)
                                                ()
                                            )
                                            Add
                                            (Integer 4)
                                            ()
                                        )
                                        Add
                                        (Integer 4)
                                        ()
                                    )
                                    Add
                                    (Integer 4)
                                    ()
                                )
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )
                            (ArrayItem
                                (Var 2 arr)
                                [(()
                                (IntegerBinOp
                                    (IntegerBinOp
                                        (IntegerBinOp
                                            (IntegerBinOp
                                                (IntegerBinOp
                                                    Add
                                                    (Integer 4)
                                                    ()
                                                )
                                                Add
                                                (Integer 4)
                                                ()
                                            )
                                            Add
                                            (Integer 4)
                                            ()
                                        )
                                        Add
                                        (Integer 4)
                                        ()
                                    )
                                    Add
                                    (Integer 4)
                                    ()
                                )
                                ())]
                                (Character 1 1 ())
                                RowMajor
                                ()
                            )]
                            FormatFortran
                            (Character -1 0 ())
                            ()
                        )

```

### LINKED ISSUE #4719: missing space or delimiter in allocatable array output
```fortran
program main
  integer, allocatable :: a(:)
  integer :: i = 1, n = 10
  allocate(a(n))
  a  =  37 * (/ (i - 1, i = 1, n) /)
  print *, a
  deallocate(a)
end program
```
```console
$ gfortran a.f90 && ./a.out
           0          37          74         111         148         185         222         259         296         333
$ lfortran a.f90
03774111148185222259296333
```

### LINKED ISSUE #4798: bug: assignment to `reshape` is not correct
### MRE

```fortran
program array_reshape_14
   implicit none

   integer :: i23(2, 3) = reshape([-14, 3, 0, -2, 19, 1], [2, 3])
   print *, i23
end program
```

### Output

```console
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran ./examples/example.f90
-14    3    -14    3
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran ./examples/example.f90 && ./a.out
         -14           3           0          -2          19           1
```

### LINKED ISSUE #4860: Seg.fault with ```print ...``` but not ```write(*, ...```
An attempt to simplify my test program in #4837 showed that one can print a whole array with format ```(A,*(1X,I0))``` but not an array section. The program below is intended to output an array section with that format. Using ```write``` was OK but not using ```print```:
```
program segfaulta
  integer:: array(3) = [1,2,3]
  write(*,"(A,*(1X,I0))") 'array section =',array(1:2)
  print   "(A,*(1X,I0))", 'array section =',array(1:2)
end program segfaulta
```
Run-time output from lfortran 0.40.:
```
array section = 1 2
array section = 
```

assem2002:
Fixes #4860 
Fixes #4798 
Fixes #4719 
Fixes #4523 
Fixes ##5739 (The issue now complains about wrong printing with provided formmating)
towards ##5672


tests/reference/asr-int_01-9bf3eb9.stderr
```diff
@@ -1,5 +0,0 @@
-semantic error: kind argument to int(a, kind) is not a constant integer
- --> tests/errors/int_01.f90:2:36
-  |
-2 |     real(8), parameter :: ar1(3) = int([1, 2, 3], [8, 8, 8])
-  |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^
```
assem2002, line 5: This file is removed due to changes in PR #5029. I left a comment https://github.com/lfortran/lfortran/pull/5029#discussion_r1798310567 there.

certik:
@assem2002 let's finish this?

assem2002:
I'll put it as next to work on. It's tricky, but It will be good to have this issue resolved.

assem2002:
The major issue is with the various backends we support. This PR removes the implementation in `print_arr.cpp` pass that iterates on the array and print it element by element in the ASR, instead the llvm backend can handle such a thing and that how things go with `write` it uses the backend printing not the ASR.
The CI fails because we don't support array printing in the backend with c++ and WASM. Can we make the pass use the removed implementation with all other backends but LLVM?

certik:
You can keep the old pass and run it for other backends, but add a new pass or whatever new way just for the LLVM backend. That way things wouldn't change for the other backends and we can merge it.

assem2002:
The failing test works for me locally.

assem2002:
I test all the issues linked to this PR and they work as intended.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -783,7 +783,7 @@ char** parse_fortran_format(char* format, int64_t *count, int64_t *item_start) {
 
 struct array_iteration_state{
     //Preserve array size and current element index
-    int64_t array_size;
+     int32_t array_size;
```
certik, line 786: Why did we move from 64bit to 32bit?

32bit is faster, and it's ok to use by default, but if we only support 1 integer size, I think we should support 64bit, otherwise we can't index larger arrays than 2GB, so it is going to bite us later.

I think we either support just one int, in that case it must be 64 bit. Or we support both, then the default can be 32 bit for speed, as long as users can select 64 bit.

src/libasr/pass/print_arr.cpp
```diff
@@ -162,7 +163,11 @@ class PrintArrVisitor : public PassUtils::PassVisitor<PrintArrVisitor>
             empty_print_endl = ASRUtils::STMT(ASR::make_Print_t(al, x.base.base.loc, empty_space));
             ASR::StringFormat_t* format = ASR::down_cast<ASR::StringFormat_t>(x.m_text);
             for (size_t i=0; i<format->n_args; i++) {
-                if (PassUtils::is_array(format->m_args[i])) {
+                bool print_not_handled_by_backend =   (pass_options.backend != Backend::llvm) ||
+                                                ASR::is_a<ASR::StructType_t>(*
+                                                ASRUtils::type_get_past_array_pointer_allocatable(
+                                                ASRUtils::expr_type(format->m_args[i])));
+                if (PassUtils::is_array(format->m_args[i]) && print_not_handled_by_backend ) {
```
certik, line 170: This is problematic. I think we should not be hardwiring backend information into ASR passes. I think ASR passes should not care what current and future backends you use them with. 

Rather, we want to pass in the option to either transform the array or not, as a regular option into the pass constructor. Then at the upper level you decide what passes and what options to call.

src/libasr/utils.h
```diff
@@ -61,6 +65,7 @@ struct PassOptions {
     bool openmp = false;
     bool experimental_simplifier = false;
     bool enable_gpu_offloading = false;
+    Backend backend = Backend::llvm;
```
certik, line 68: I think this is problematic. I think we don't care in the compiler options which backends we use. For example, users can write their own backends, and they want to call certain passes.  We should not hardwire backends like this, nor make any decisions based on what backends are run.

Rather, the compilation is composed of various passes with options, and users assemble their compilation pipeline as they see fit.

src/bin/lfortran.cpp
```diff
@@ -64,10 +64,8 @@ namespace {
 
 using LCompilers::endswith;
 using LCompilers::CompilerOptions;
+using LCompilers::Backend;
 
-enum Backend {
-    llvm, c, cpp, x86, wasm, fortran, mlir
-};
```
certik, line 70: I think this should stay here. It's the highest-level driver that should decide what pipelines and backends will get run. The internals of the compiler should not care.

certik (commented):
Thanks for the fixes! I think there are some design choices that I think we should do differently, I left comments. Let me know what you think.

Let's also add tests that previously segfaulted but now work.

Why does it fail with FAST?

src/libasr/runtime/lfortran_intrinsics.c
assem2002: Yeah I think it kinda of important thing to handle. We have issue with handling size node with return type of kind 8 . I'll invoke the error and let you know and maybe create and MRE and report the issue. I would like also to mention that allocatable arrays don't handle size of `i64` the array_descriptor just have `i32`.

src/libasr/pass/print_arr.cpp
assem2002: That's great thing to know about. So you suggest passing a flag to the pass saying (okay we support printing array in the backend we're targeting), rather than making the passes themselves try to make decisions based on the passed backend?

src/bin/lfortran.cpp
assem2002: If that is https://github.com/lfortran/lfortran/pull/5039#discussion_r1878752235 what we should do, so this change won't be needed.

src/libasr/pass/print_arr.cpp
certik: Yes, exactly. You tell the ASR pass exactly what you want to get done. If we are not transforming the printing, is the `print_arr` pass even doing anything? If so, just pass in a flag to customize what it is doing.

src/libasr/runtime/lfortran_intrinsics.c
assem2002: I addressed the issue here https://github.com/lfortran/lfortran/issues/5647. I'm working on it.

src/libasr/pass/print_arr.cpp
assem2002: I just want to make sure we're on the same track. I'll just create a `bool backend_supports_array_printing=false` as member of the `PassOptions` struct and the `print_arr` acts upon that flag. Is that correct?

assem2002: for now we support printing arrays in LLVM for fortran primitive types (character, integer, real, complex, ...) but not for arrays of type struct (Fortran `type`). for any other backend we don't yet support printing arrays in the backend.
So, We need to always use the pass with all the backends; but with LLVM we have to use the pass for only the case of array of type struct (Fortran type).

assem2002: The actual motive behind printing arrays in the backend instead of using the ASR was returning correct output when using `fmt` as it could be runtime value so we had to print the array at the runtime instead of replacing the print node for the whole array with arrayitems, so we gave everything to the backend (the `fmt` and all the arguments) to handle the formatting at runtime.

now I'm questioning the way we went through! Direct array printing statement seems to be a thing to be handled by the frontend of the compiler as we could be using low-level backend like LLVM or C or C++ that doesn't know how to print an array. but we still face the issue of the runtime formatting on the args, The ASR can't spread the array into separate printable elements for the string_format function to recognize -- print_arr pass loops on the array based on the runtime size of the array.

We need to find a way that makes the runtime-formatting function match against the actual element (runtime element) but also to not make the backend really concerned about proper looping and knowing the type of the thing it's trying to print ( EX : An array of struct `A` ( struct `A` has struct `B` as member -- would be a messy message to pass)

PR #5043, opened 2024-10-14
CI: update the CI job for fastGPT

gxyd:
I've used the Fortran compiler flag *--realloc-lhs*

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -865,78 +865,78 @@ jobs:
             make -f Makefile.manual F90="lfortran --skip-pass=inline_function_calls,fma --fast"
             make -f Makefile.manual test
 
-      # - name: Test fastGPT ( ubuntu-latest )
-      #   shell: bash -e -x -l {0}
-      #   if: contains(matrix.os, 'ubuntu')
-      #   run: |
-      #       git clone https://github.com/certik/fastGPT.git
-      #       cd fastGPT
-      #       git checkout -t origin/lf6
-      #       git checkout bc04dbf476b6173b0bb945ff920119ffaf4a290d
-      #       echo $CONDA_PREFIX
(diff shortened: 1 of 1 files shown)


Pranavchiku (approved):
This looks good but previously then why it worked without `--realloc-lhs` flag? We need to understand and document how `--realloc-lhs` affects simplifier pass.

gxyd:
> previously then why it worked without --realloc-lhs flag? We need to understand and document how --realloc-lhs affects simplifier pass.

We had a sort of an ad hoc condition for replicating what "--realloc-lhs" does, so internally the main would eventually create a `ReAlloc` node (in `array_op` pass) in an MRE like https://github.com/lfortran/lfortran/issues/4761#issuecomment-2408887020
```f90
program main
   implicit none

   integer :: i

   type :: string
   end type

   type(string), allocatable :: tokens(:)

   allocate(tokens(5))
   do i = 1, 5
      tokens(i)%s = "a"
   end do

   ! current *main* is smart to identify that "ReAlloc" node needs to be created for a var like below
   tokens = merge_pair(tokens, 1)

contains

   function merge_pair(intokens, idx) result(tokens)
      ! Merge the pair `idx`
      type(string), intent(in) :: intokens(:)
      integer, intent(in) :: idx
      type(string), allocatable :: tokens(:)
      type(string) :: merged_token
      merged_token%s = intokens(idx)%s // intokens(idx+1)%s
      ! The segmentation fault occurs at the below line
      tokens = [intokens(:idx-1), merged_token, intokens(idx+2:)]
   end function

end program main
```

Pranavchiku:
Okay, let's add this as a test in main with `--realloc-lhs` and then merge main into it or simplifier to main ;)

gxyd:
> Okay, let's add this as a test in main with --realloc-lhs and then merge main into it or simplifier to main ;)

Yes, let me do that.

I'll remember that at some point we eventually try to remove this `--realloc-lhs` compiler flag from fastGPT once we manage to merge *simplifier_pass* into *main*

PR #5048, opened 2024-10-14
CI: uncomment windows CI job to see if it passes

gxyd:


Pranavchiku:
Yes, I also attempted it here https://github.com/lfortran/lfortran/pull/5032, it fails

gxyd:
> Yes, I also attempted it here #5032, it fails

I see. I think the way to go here would be look one by one at the CI log and fix it in our branch directly, and push the changes to see where it fails again.

Pranavchiku:
> I think the way to go here would be look one by one

Precisely, that is how you fix it :)

Pranavchiku:
If you have access to windows, then better is to reproduce it locally, attempts to check directly on CI takes a lot of time to iterate. Else, comment out other CI tests, just keep windows, this way you can reduce load on runner and make tests run faster.

gxyd:
The tests passed in the last commit, so I've uncommented all the other CI jobs and enabled auto-merge (nothing important to be reviewed)

PR #5049, opened 2024-10-14
Throw error for non scalar kind arg in int intrinsic

HarshitaKalani:
Fixing: https://github.com/lfortran/lfortran/pull/5029#discussion_r1798310567


certik (approved):
I think this looks good.

PR #5051, opened 2024-10-14
CI: Consolidate the two workflows

certik:


certik:
I had my `ondrej` remote set to `lfortran/lfortran` by a mistake instead of `certik/lfortran`, my apologies for pushing this branch upstream.

PR #5061, opened 2024-10-15
rem: present from runtime library as implemented as an intrinsic

Pranavchiku:


HarshitaKalani (approved):
Thanks @Pranavchiku for this! :tada: 
I think this is great.

certik (approved):
I think this looks good. It makes it an intrinsic, and later transforms it with a pass.

PR #5067, opened 2024-10-16
chore: create temporary for a function call only when the return type is an Array

gxyd:
## Description

Helps fix the MRE present here: https://github.com/lfortran/lfortran/issues/4761

## Testing

Test program for this has been sent as a PR against *main* here: https://github.com/lfortran/lfortran/pull/5068. Once that's merged, I'll merge *main* into *simplifier_pass*.

Currently, I've ensured locally (on macOS m1) that this change helps us run the test program in https://github.com/lfortran/lfortran/pull/5068


gxyd:
Seems like fastGPT errors with this change.

PR #5071, opened 2024-10-16
chore: replace intrinsic elemental function with temporary when return type is StructType

gxyd:
## Description

We need to add an MRE present here: https://github.com/lfortran/lfortran/issues/4761#issuecomment-2416936436 to the *main* branch, once the tests on this PR succeed.


Pranavchiku (approved):
Looks fine, if tests passes let's merge it.

PR #5077, opened 2024-10-17
chore: fix return type of elemental function (could be of generic procedure)

gxyd:
## Description

If atleast one of the argument is an array, we should create the return type accordingly

## Why PR against *simplifier_pass*?

This PR is ideally to be created against *main* and not *simplifier_pass*, but I'll like to do that just before *simplifier_pass* is ready to be merged.

### DIFF
--- a/integration_tests/elemental_function_scalar_array_arg.f90
+++ b/integration_tests/elemental_function_scalar_array_arg.f90
@@ -1,5 +1,9 @@
 module mod_elemental_function_scalar_array_arg
     implicit none
+    interface count
+        module procedure count_char_char
+    end interface
+
     contains
 
     !> Returns an integer
@@ -17,10 +21,15 @@ program elemental_function_scalar_array_arg
     use mod_elemental_function_scalar_array_arg
     implicit none
     character(len=128) :: string
-    integer :: count(3)
+    integer :: count_value1(3)
+    integer :: count_value2(3)
 
     string = "How much wood would a woodchuck chuck if a woodchuck could chuck wood?"
     print *, count_char_char(string, ["would", "chuck", "could"])
-    count = count_char_char(string, ["would", "chuck", "could"])
-    if (any(count /= [1, 1, 1])) error stop
+    count_value1 = count_char_char(string, ["would", "chuck", "could"])
+    if (any(count_value1 /= [1, 1, 1])) error stop
+
+    print *, count(string, ["would", "chuck", "could"])
+    count_value2 = count(string, ["would", "chuck", "could"])
+    if (any(count_value2 /= [1, 1, 1])) error stop
 end program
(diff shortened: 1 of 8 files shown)


gxyd:
This PR might need updating the reference tests

gxyd:
The expected tests pass, so the change is definitely good.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4313,12 +4313,23 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    break;
+                }
+            }
+        }
+        if( is_elemental && func->n_args >= 1 && any_array_arg) {
+            LCOMPILERS_ASSERT(first_array_arg)
             ASR::dimension_t* array_dims;
             size_t array_n_dims = ASRUtils::extract_dimensions_from_ttype(
-            ASRUtils::expr_type(args[0].m_value), array_dims);
+                ASRUtils::expr_type(first_array_arg), array_dims);
```
Pranavchiku, line 4332: If it is an elemental, we shall use broadcast and not return an array

integration_tests/elemental_function_scalar_array_arg.f90
```diff
@@ -0,0 +1,35 @@
...
+program elemental_function_scalar_array_arg
+    use mod_elemental_function_scalar_array_arg
+    implicit none
+    character(len=128) :: string
+    integer :: count_value1(3)
+    integer :: count_value2(3)
+
+    string = "How much wood would a woodchuck chuck if a woodchuck could chuck wood?"
+    print *, count_char_char(string, ["would", "chuck", "could"])
+    count_value1 = count_char_char(string, ["would", "chuck", "could"])
```
Pranavchiku, line 29: This shall be handled by `array_op` and it shall broadcast

Pranavchiku (approved):
I see it makes senes, this PR finds the array argument and uses its dimension.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4313,12 +4313,23 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+        ASR::expr_t* first_array_arg = nullptr;
+        if (is_elemental && func->n_args >= 1) {
+            for (size_t i=0; i < args.size(); i++) {
+                if (args[i].m_value && ASRUtils::is_array(ASRUtils::expr_type(args[i].m_value))) {
+                    any_array_arg = true;
+                    first_array_arg = args[i].m_value;
+                    break;
+                }
+            }
+        }
```
certik, line 4327: I see this code repeated 3x, could it be extracted into a function?

certik (approved):

src/lfortran/semantics/ast_common_visitor.h
gxyd: Yes, definitely I would do that.

PR #5078, opened 2024-10-17
ci: enable ctest on stdlib

Pranavchiku:


a.f90
```diff
@@ -0,0 +1,10 @@
+program test
```
gxyd, line 1: was this file added mistakenly?

Issue #4222, opened 2024-06-10
Duplicate functions in interface not raising error


assem2002:
related to https://github.com/lfortran/lfortran/issues/4007

```Fortran
MODULE input_module
    implicit none
    INTERFACE test_interface
       MODULE PROCEDURE test_01, test_01, test_01
    END INTERFACE test_interface
 CONTAINS
 
    SUBROUTINE test_01 (x)
        implicit none
        integer , intent(in):: x 
    END SUBROUTINE test_01

 END MODULE input_module

```
This should raise an error as there are 3 modules with the same name.
there's no harm in duplicating it, but it's not a good practice. 
If we raise an ambiguity error when issue https://github.com/lfortran/lfortran/issues/4007 is resolved, then this error should be a must.

mohitdhingra22:
I have created a new test for the duplicate module procedure names and registered it in tests.toml. The PR is ready for review. Here is the link to the PR:[mohitdhingra22:fix_duplicate_procedures](url)

NishantBansal2003:
I was using `std::set<ASR::symbol_t*> symbols_set;` (following this https://github.com/lfortran/lfortran/pull/4732#pullrequestreview-2275807890) to find unique symbols and throwing an error like this:

```cpp
throw SemanticError("Entity " + correct_pname + " is already present in the interface", x->base.loc);
```
However, the error message for the test appears as follows:

```vbnet
semantic error: Entity test_01 is already present in the interface
  --> duplicate.f90:8:5 - 12:26
   |
 8 |        SUBROUTINE test_01 (x)
   |        ^^^^^^^^^^^^^^^^^^^^^^...
...
   |
12 |        END SUBROUTINE test_01
   | ...^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```
Is there a way I can get the error location to appear more like the following `gfortran` error format?

```vbnet
duplicate.f90:4:41:

    4 |        MODULE PROCEDURE test_01, test_01, test_01
      |                                         1
Error: Entity 'test_01' at (1) is already present in the interface
```
cc: @kmr-srbh

assem2002:
@NishantBansal2003. I think you can push a PR, It'd be more convenient to see where you've made the changes in the codebase.

NishantBansal2003:
[...]

PR #5081, opened 2024-10-17
Raise error for duplicate functions in interface

NishantBansal2003:
### PR Description:
This PR addresses an issue where duplicate symbols in the `add_generic_procedures()` function were not being properly caught and reported, potentially leading to incorrect compilation results. The problem occurred while pushing symbols into the `Vec<ASR::symbol_t*>` structure in `ast_symboltable_visitor.cpp`.
fixes: #4222 

### Key Changes:
Added logic to detect and report duplicate symbols in `add_generic_procedures()`.
### Testing:
Verified with test cases that the system now correctly handles and reports duplicates.

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1770,13 +1770,24 @@ public:
             if (AST::is_a<AST::InterfaceModuleProcedure_t>(*item)) {
                 AST::InterfaceModuleProcedure_t *proc
                     = AST::down_cast<AST::InterfaceModuleProcedure_t>(item);
+                std::set<std::string> items_set;
                 for (size_t i = 0; i < proc->n_names; i++) {
                     /* Check signatures of procedures
                     * to ensure there are no two procedures
                     * with same signatures.
                     */
                     char *proc_name = proc->m_names[i];
-                    proc_names.push_back(std::string(proc_name));
+                    std::string item_proc_name = std::string(proc_name);
+                    if (items_set.find(item_proc_name) == items_set.end()) {
+                        proc_names.push_back(item_proc_name);
+                        items_set.insert(item_proc_name);
+                    } else {
+                        diag.semantic_error_label("Entity " + item_proc_name
+                                                      + " is already present in the interface",
+                                                  { item->base.loc },
+                                                  " ");
+                        throw SemanticAbort();
+                    }
                 }
             } else if(AST::is_a<AST::InterfaceProc_t>(*item)) {
                 visit_interface_item(*item);
@@ -1917,7 +1928,6 @@ public:
             loc.first = 1;
             loc.last = 1;
             Vec<ASR::symbol_t*> symbols;
-            std::set<ASR::symbol_t*> symbols_set;
             symbols.reserve(al, proc.second.size());
             for (auto &pname : proc.second) {
                 std::string correct_pname = pname;
@@ -1931,14 +1941,7 @@ public:
                 // lower case the name
                 name = s2c(al, to_lower(name));
                 x = resolve_symbol(loc, name);
-                if (symbols_set.find(x) == symbols_set.end()) {
-                    symbols.push_back(al, x);
-                    symbols_set.insert(x);
-                } else {
-                    throw SemanticError("Entity " + correct_pname
-                                            + " is already present in the interface",
-                                        x->base.loc);
-                }
+                symbols.push_back(al, x);
             }
             std::string sym_name_str = proc.first;
             if( current_scope->get_symbol(proc.first) != nullptr ) {
@@ -1963,14 +1966,7 @@ public:
                             ASRUtils::symbol_name(gp->m_procs[i]));
                         if (s != nullptr) {
                             // Append all the module procedure's in the scope
-                            if (symbols_set.find(s) == symbols_set.end()) {
-                                symbols.push_back(al, s);
-                                symbols_set.insert(s);
-                            } else {
-                                throw SemanticError("Entity " + sym_name_str
-                                                        + " is already present in the interface",
-                                                    s->base.loc);
-                            }
+                            symbols.push_back(al, s);
                         } else {
                             // If not available, import it from the module
                             // Create an ExternalSymbol using it
@@ -1986,15 +1982,7 @@ public:
                                         fn->m_name, dflt_access);
                                 current_scope->add_symbol(fn->m_name, ep_s);
                                 // Append the ExternalSymbol
-                                if (symbols_set.find(ep_s) == symbols_set.end()) {
-                                    symbols.push_back(al, ep_s);
-                                    symbols_set.insert(ep_s);
-                                } else {
-                                    throw SemanticError(
-                                        "Entity " + sym_name_str
-                                            + " is already present in the interface",
-                                        ep_s->base.loc);
-                                }
+                                symbols.push_back(al, ep_s);
                             }
                         }
                     }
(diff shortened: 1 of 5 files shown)


NishantBansal2003:
Please address this issue: https://github.com/lfortran/lfortran/issues/4222#issuecomment-2418802943 so that I can make the necessary changes in this PR.
Thanks!

NishantBansal2003:
I'm not sure why `CI / Build LFortran to WASM and Upload` is failing. Is the failure due to flakiness, or did I make some errors?

NishantBansal2003:
Please review this PR and provide suggestions on where changes are needed. Thanks...
CC: @certik, @Pranavchiku, @kmr-srbh

tests/reference/asr-duplicate_module_procedures-3eaed23.stderr
```diff
@@ -0,0 +1,9 @@
+semantic error: Entity test_01 is already present in the interface
+  --> tests/errors/duplicate_module_procedures.f90:8:5 - 12:26
+   |
+ 8 |        SUBROUTINE test_01 (x)
+   |        ^^^^^^^^^^^^^^^^^^^^^^...
+...
+   |
+12 |        END SUBROUTINE test_01
```
certik, line 8: This error message is reported at the wrong place. Instead, here is how it should look like:
```
a.f90:4:41:

    4 |        MODULE PROCEDURE test_01, test_01, test_01
      |                                         1
Error: Entity 'test_01' at (1) is already present in the interface
```

It needs to highlight the second `test_01` and report that it is already present.

certik:
@NishantBansal2003 awesome, thanks for this PR! Let's improve the error message based on my comment above, and then we can merge it.

tests/reference/asr-duplicate_module_procedures-3eaed23.stderr
NishantBansal2003: Exactly, I was facing this issue (https://github.com/lfortran/lfortran/issues/4222#issuecomment-2418802943). I'm having trouble identifying the correct location of the error. I have tried several possible locations, but none have worked as expected. Could you assist me with this?

NishantBansal2003: This section is also marked with FIXME LOCATION: https://github.com/lfortran/lfortran/blob/7fa4bc901ab2f729968cbddc21028c2767a9119d/src/lfortran/semantics/ast_symboltable_visitor.cpp#L1913-L1920
 However, I'm not sure if this is related to the issue.

certik: Print the part of AST that corresponds to `MODULE PROCEDURE test_01, test_01, test_01`. In there, let's ensure we can see the three `test_01` each with different location information. 

* If we can't, then we need to fix AST until we can
* Once we can, you simply select the second `test_01` to report the error

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1930,7 +1931,14 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 // lower case the name
                 name = s2c(al, to_lower(name));
                 x = resolve_symbol(loc, name);
-                symbols.push_back(al, x);
+                if (symbols_set.find(x) == symbols_set.end()) {
+                    symbols.push_back(al, x);
+                    symbols_set.insert(x);
```
certik, line 1936: You are never resetting `symbols_set`, so I think this will lead to bugs if the source code repeats the interface.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1930,7 +1931,14 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                } else {
+                    throw SemanticError("Entity " + correct_pname
+                                            + " is already present in the interface",
+                                        x->base.loc);
```
certik, line 1940: Use SemanticAbort instead, we are slowly getting rid of SemanticError. See other code how to use it.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1971,7 +1986,15 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
-                                symbols.push_back(al, ep_s);
+                                if (symbols_set.find(ep_s) == symbols_set.end()) {
+                                    symbols.push_back(al, ep_s);
+                                    symbols_set.insert(ep_s);
+                                } else {
+                                    throw SemanticError(
+                                        "Entity " + sym_name_str
+                                            + " is already present in the interface",
+                                        ep_s->base.loc);
+                                }
```
certik, line 1997: Why do we need to perform this check at 3 places?

NishantBansal2003:
I have fixed the required changes. Please take a look, @certik. Also, could you provide a little information about this?
> You are never resetting `symbols_set`, so I think this will lead to bugs if the source code repeats the interface.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1770,13 +1770,24 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                    if (items_set.find(item_proc_name) == items_set.end()) {
+                        proc_names.push_back(item_proc_name);
+                        items_set.insert(item_proc_name);
+                    } else {
+                        diag.semantic_error_label("Entity " + item_proc_name
+                                                      + " is already present in the interface",
+                                                  { item->base.loc },
+                                                  " ");
+                        throw SemanticAbort();
+                    }
```
certik, line 1790: The `items_set` is only inserted into, but never reset. So it will get corrupted once you have multiple interfaces. For example consider:
```fortran
    interface test_interface1
       module procedure test_01, test_02
    end interface
    interface test_interface2
       module procedure test_01, test_02
    end interface
```
This should fail your check, but in fact should succeed.

tests/errors/duplicate_module_procedures.f90
```diff
@@ -0,0 +1,20 @@
...
+        print *, x
+    END SUBROUTINE test_01
+
+END MODULE input_module
+
+
+PROGRAM main
+    USE input_module
+    CALL test_interface(1)
+END PROGRAM main
```
certik, line 20: Let's use all lower case, as is the habit in modern Fortran.

src/lfortran/semantics/ast_symboltable_visitor.cpp
NishantBansal2003: Since I am declaring `items_set` for a particular instance of `AST::interface_item_t`, the `items_set` will reset itself due to redeclaration when multiple interfaces are declared.
Output from your example:
```fortran
module input_module
    implicit none
    interface test_interface
       module procedure test_01, test_02
    end interface test_interface
    interface test_interface2
       module procedure test_01, test_02
    end interface
contains

    subroutine test_01 (x)
        implicit none
        integer, intent(in) :: x
        print *, x
    end subroutine test_01

    subroutine test_02 (x, y)
        implicit none
        integer, intent(in) :: x
        integer, intent(in) :: y
        print *, x, y
    end subroutine test_02

end module input_module

program main
    use input_module
    call test_interface(1)
    call test_interface2(1, 1)
end program main
```
Output:

```bash
╰─ ./src/bin/lfortran duplicate.f90                ─╯
1
1    1
```

certik: You are right, my bad. I think this is the way to implement it.

certik (approved):
Thanks for the fix!

The `item` has incorrect location information, it should just highlight that one item, not the whole line, but that has to be fixed likely in the parser.

NishantBansal2003:
> Thanks for the fix!
> 
> The `item` has incorrect location information, it should just highlight that one item, not the whole line, but that has to be fixed likely in the parser.

Should I open a new issue for this?

certik:
> Should I open a new issue for this?

Yes, please do.

Issue #1941, opened 2023-07-06
ASR passes refactoring


certik:
* [ ] https://github.com/lfortran/lfortran/pull/1937
* [ ] https://github.com/lfortran/lfortran/pull/1938
* [ ] Only keep actual passes in the `pass/` directory, all registered in `gen_pass.py`, everything else should move one level up
* [ ] automatically register all passes in the main program (expose on a command line)
* [ ] automatically register all passes in the pass manager
* [ ] refer to passes using Enum, not a string
* [ ] have the pass manager maintain a list of passes (and later how they should be run if they have any options), allow the backend to modify this list, as well as allow to modify it from the command line, and print it, etc.
* [ ] print each pass to a file
* [ ] each pass has to have a short and long description, that one can print from the main driver
* [ ] generate documentation

PR #5094, opened 2024-10-18
Add experimental-simplifer flag 

assem2002:
Add experimental-simplifer flag to alter between two list of ASR passes to be used

### DIFF
--- a/src/libasr/pass/pass_manager.h
+++ b/src/libasr/pass/pass_manager.h
@@ -71,7 +71,8 @@ namespace LCompilers {
 
     class PassManager {
         private:
-
+        std::vector<std::string> _passes_with_experimental_simplifier;
+        std::vector<std::string> _with_optimization_passes_for_experimental_simplifier;
         std::vector<std::string> _passes;
         std::vector<std::string> _with_optimization_passes;
         std::vector<std::string> _user_defined_passes;
@@ -208,7 +209,7 @@ namespace LCompilers {
 
         PassManager(): apply_default_passes{false},
             c_skip_pass{false} {
-            _passes = {
+            _passes_with_experimental_simplifier = {
                 "global_stmts",
                 "init_expr",// This pass shouldn't be needed.
                 "function_call_in_declaration",
@@ -244,7 +245,7 @@ namespace LCompilers {
                 "insert_deallocate",
             };
 
-            _with_optimization_passes = {
+            _with_optimization_passes_for_experimental_simplifier = {
                 "global_stmts",
                 "init_expr",// This pass shouldn't be needed.
                 "function_call_in_declaration",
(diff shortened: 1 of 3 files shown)


certik (approved):
I think this is the way to go. To test it out, port the simplifier pass itself (two files) and enable it. Then we can check this PR out, and see if it enables it for some simple example. Then we'll go from there.

assem2002:
This PR #5095 does the same with main + porting the simplifier pass and some utility function it needs + using flag `experimental-simplifier` with two integration tests just for checking out that it works.

assem2002:
I think this one is ready for merge so we can start use to port chunks in small PR into main branch.

src/libasr/utils.h
```diff
@@ -59,6 +59,7 @@ struct PassOptions {
     bool with_intrinsic_mods = false;
     bool c_mangling = false;
     bool openmp = false;
+    bool experimental_simplifier = true;
```
certik, line 62: Eventually this will be the only change in the simplifier PR. :) 
```diff
-    bool experimental_simplifier = false;
+    bool experimental_simplifier = true;
```
Everything else will already be in master.

And we'll have to wait with merging it until this flip from "false" to "true" does not bring up significant regressions.

src/libasr/pass/pass_manager.h
```diff
@@ -313,10 +387,14 @@ namespace LCompilers {
...
-                    apply_passes(al, asr, _with_optimization_passes, pass_options,
+                if( pass_options.fast && pass_options.experimental_simplifier ) {
+                    apply_passes(al, asr, _with_optimization_passes_for_experimental_simplifier, pass_options,
                         diagnostics);
-                } else {
+                } else if (!pass_options.fast && pass_options.experimental_simplifier) {
+                    apply_passes(al, asr, _passes_with_experimental_simplifier, pass_options, diagnostics);
+                } else if (pass_options.fast && !pass_options.experimental_simplifier){
+                    apply_passes(al, asr, _with_optimization_passes, pass_options, diagnostics);
+                } else if (!pass_options.fast && !pass_options.experimental_simplifier) {
```
certik, line 397: This is fine for now, but it's getting really complex. We really need to refactor all this according to https://github.com/lfortran/lfortran/issues/1941.

certik (approved):
This can be merged as you see fit.

src/libasr/pass/pass_manager.h
assem2002: The refactoring doesn't seem so much hard work. It should be easy to start on it as I think.

certik: Yes, we'll refactor it in a separate PR.

certik (approved):

PR #5097, opened 2024-10-19
Handle array broadcast for functionCall returning scalar

HarshitaKalani:
Simplifier PR: https://github.com/lfortran/lfortran/pull/5096


Pranavchiku (changes requested):
I think to fix third party test failures, just skip character arrays.

integration_tests/array_op_07.f90
```diff
@@ -0,0 +1,48 @@
...
+      real, intent(in) :: a(:), b(:)
+      real :: temp(size(a))
+      array_function = 0.0
+  
+      temp = a + sqrt(b)
+      array_function = temp(0)
+    end function
+  
+end program array_op_07
+
```
gxyd, line 48: I tried running this program with the current *main*, and it passes without any problem, and I get the below output:
```console
~/OpenSource/lfortran main !1 ?10 ❯ lfortran test.f90                                                                                                                                                                                                         55s  lf 02:26:51 PM
Result:    6.25000000e+00
Result (z):    0.00000000e+00    0.00000000e+00    0.00000000e+00    0.00000000e+00    0.00000000e+00
array(    1    ) squared =    1.00000000e+00
array(    2    ) squared =    1.00000000e+02
array(    3    ) squared =    9.00000000e+00
array(    4    ) squared =    1.00000000e+02
array(    5    ) squared =    2.50000000e+01
```

Can you help me point to an example that doesn't work or works incorrectly with the current main?

gxyd: and GFortran produces the below output:
```console
~/OpenSource/lfortran main !1 ?11 ❯ gfortran test.f90 -o a.out && ./a.out                                                                                                                                                                                          lf 02:29:15 PM
test.f90:44:28:

   44 |       array_function = temp(0)
      |                            1
Warning: Array reference at (1) is out of bounds (0 < 1) in dimension 1
 Result:   6.25000000    
 Result (z):   0.00000000       0.00000000       0.00000000       0.00000000       0.00000000    
 array(           1 ) squared =   1.00000000    
 array(           2 ) squared =   100.000000    
 array(           3 ) squared =   9.00000000    
 array(           4 ) squared =   100.000000    
 array(           5 ) squared =   25.0000000
```

which seems the same.

HarshitaKalani: Yes, it works with the main and produces correct output. I just pushed the changes done for the simplifier pass branch (https://github.com/lfortran/lfortran/pull/5096) to main as well.

integration_tests/array_op_07.f90
```diff
@@ -0,0 +1,48 @@
...
+  
+    ! Declare variables
+    integer, dimension(5) :: array
+    real :: scalar_value
+    integer :: i
+    real, dimension(5) :: x, y, z
+  
+    array = (/ 1, 0, 3, 0, 5 /)
+    x = (/ 1.0, 2.0, 3.0, 4.0, 5.0 /)
+    y = (/ 2.0, 4.0, 6.0, 8.0, 10.0 /)
```
certik, line 12: Use `[]` for arrays.

integration_tests/array_op_07.f90
Pranavchiku: > I am confused --- if the test above works already in main, than the change below is not needed?

This is weird, in release mode with `0.41.0` it fails but works with latest main.

certik: I think we should understand what the code is doing and why (if at all) it is needed.

gxyd:
Since there are some comments which need to be addressed first, I'll move this to *draft mode* for now

certik:
Let's finish this.

HarshitaKalani:
Closing as completed in https://github.com/lfortran/lfortran/pull/5452

Issue #5092, opened 2024-10-18
Test that work on simplifier branch but no on main


Pranavchiku:
```fortran
program main
    implicit none
    integer :: res(4)
    real :: x(1,4), center, order
    integer, parameter :: dim = 1
    order = 2
    x = 1
    center = 0
    res = sum((x - center), dim) / size(x, dim)
    print *, res
end program
```

```console
% lfortran --version
LFortran version: 0.41.0-488-g7d99f976d
Platform: macOS ARM
Default target: arm64-apple-darwin23.5.0
% lfortran a.f90 
1    0    0    0
 % lfortran --version
LFortran version: 0.41.0
Platform: macOS ARM
Default target: arm64-apple-darwin23.5.0
% lfortran a.f90 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x1000266b3
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x100010e4b
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x10001d55b
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x1005587bb
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x101774037
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x100062517
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x1012a17c3
  Binary file "/Users/pranavchiku/mambaforge/envs/lf2/bin/lfortran", local address: 0x1012bdd77
LCompilersException: Number of arguments in the new function call Sum_4_2_1 doesn't satisfy min_args <= new_args.size() <= max_args, 9 <= 7 <= 9

```

Pranavchiku:
We can also revert: https://github.com/czgdp1807/stdlib/commit/645b0939cc8249fae606dae7a829116392c7426f#diff-d5b2252f72add3783967ad55cb0296db75f2ce752d3d921ac8406839933a26a2L268

certik:
```fortran
program main
implicit none
integer, parameter :: N = 10
real :: x(N), y(N)
x = 3
y = 4
[...]

PR #5099, opened 2024-10-20
Fix issues with ``ArrayBroadcast`` node

czgdp1807:
https://github.com/lfortran/lfortran/issues/5092#issuecomment-2422798263
https://github.com/lfortran/lfortran/issues/5092#issuecomment-2422723589


certik:
Thanks! Tests currently fail, so they need to be fixed.

PR #5105, opened 2024-10-21
Undo changes which aren't needed any more in simplifier_pass

gxyd:
## Description

This PR introduces the below changes:

1. undo the workaround in nbody.f90, as that workaround isn't needed anymore
2. use `ASR::make_Variable_t` instead of `ASRUtils::make_Variable_t_util`, as there isn't any difference between the two anymore (initially there was)

### DIFF
--- a/integration_tests/nbody.f90
+++ b/integration_tests/nbody.f90
@@ -104,14 +104,7 @@ program nbody
         real(kind=dp), dimension(3) :: p
         integer :: i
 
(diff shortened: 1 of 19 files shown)


gxyd:
Tests pass, and this PR just undoes some of the changes. So, I'm merging it without needing a review for it.

certik:
Excellent!

PR #4700, opened 2024-08-28
Simplifier Pass

certik:
This is a long-running branch and a PR with a simplifier pass. To collaborate on this, simply send a PR against this `simplifier_pass` branch and if merged, it will automatically appear in this PR. We can use this PR to easily see what the changes are against `main` and our goal is to get these changes to the point that we can merge it. The CI should always be green in this PR.

TODO to merge *simplifier_pass* into *main*:

- [ ] failure of 4 integration tests
- [x] stdlib compilation and run
- [x] fastGPT compilation and run
- [x] updating the reference tests
- [x] Build and test LFortran on Windows (relevant PR: https://github.com/lfortran/lfortran/pull/5048)
- [x] Test WASM backend https://github.com/lfortran/lfortran/pull/5085
- [x] Test C/C++ backend  https://github.com/lfortran/lfortran/pull/5079 (relevant PR: https://github.com/lfortran/lfortran/pull/5046)
- [x] Benchmark SNAP: https://github.com/lfortran/lfortran/issues/4631

---

### sync w main

* [ ] .github/workflows/CI.yml
* [x] src/lfortran/semantics/ast_body_visitor.cpp https://github.com/lfortran/lfortran/pull/5118
* [x] src/lfortran/semantics/ast_common_visitor.h https://github.com/lfortran/lfortran/pull/5118
* [x] src/libasr/ASR.asdl #5161 
* [x] src/libasr/CMakeLists.txt https://github.com/lfortran/lfortran/pull/5118
* [x] src/libasr/asdl_cpp.py #5160 
* [x] src/libasr/asr_builder.h https://github.com/lfortran/lfortran/pull/5118
* [x] src/libasr/asr_utils.cpp https://github.com/lfortran/lfortran/pull/5118
* [x] src/libasr/asr_utils.h https://github.com/lfortran/lfortran/pull/5118 + https://github.com/lfortran/lfortran/pull/5120
* [x] src/libasr/asr_verify.cpp #5136 
* [x] src/libasr/codegen/asr_to_llvm.cpp #5168 
* [x] src/libasr/codegen/asr_to_wasm.cpp #5171 
* [x] src/libasr/gen_pass.py #5146 
* [x] src/libasr/intrinsic_func_registry_util_gen.py #5159 
* [x] src/libasr/pass/array_op.cpp #5125 
* [x] src/libasr/pass/implied_do_loops.cpp https://github.com/lfortran/lfortran/pull/5115
[...]

PR #5115, opened 2024-10-21
sync:`implied_do_loops`, `init_expr`, `pass_utils.h/.cpp` with simplifier

Pranavchiku:
Towards #4700

### DIFF
--- a/src/libasr/pass/init_expr.cpp
+++ b/src/libasr/pass/init_expr.cpp
@@ -26,14 +26,15 @@ class ReplaceInitExpr: public ASR::BaseExprReplacer<ReplaceInitExpr> {
     ASR::cast_kindType cast_kind;
     ASR::ttype_t* casted_type;
     bool perform_cast;
+    bool pass_simplifier = false;
 
     ReplaceInitExpr(
         Allocator& al_,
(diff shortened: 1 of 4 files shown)


src/libasr/pass/init_expr.cpp
```diff
@@ -63,10 +64,12 @@ class ReplaceInitExpr: public ASR::BaseExprReplacer<ReplaceInitExpr> {
         if( casted_type != nullptr ) {
             casted_type = ASRUtils::type_get_past_array(casted_type);
         }
-        PassUtils::ReplacerUtils::replace_ArrayConstructor(x, this,
+        if (pass_simplifier) PassUtils::ReplacerUtils::replace_ArrayConstructor_(al, x, result_var, result_vec,
+            current_scope, perform_cast, cast_kind, casted_type);
+        else PassUtils::ReplacerUtils::replace_ArrayConstructor(x, this,
```
certik, line 69: Btw, it's a good habit in C++ to use:
```
if () {
...
} else {
...
}
```
even if it is just one line, as it's less prone to user mistakes later when modifying or adding statements. This can be improved later.

certik (approved):
I think these changes are fine. Just in case make sure there is no ASR ballooning happening in the default path.

PR #5118, opened 2024-10-21
Sync : ast_body_visitor, ast_common_visitor, CMakeLists, asr_builder,…

assem2002:
… asr_utils.h, asr_utils.cpp, simplifier.cpp

---
lots of the changes are just empty spacing and tabs (I'm keeping them to match simplifier branch).


assem2002:
`asr_utils.h` has small hunk left to sync but it leads to changes in the reference test. It causes no issues with the default branching we have. I preferred to push that hunk in another PR, It would be subsequent to this current PR as it depends on it.

certik (approved):
These changes look good to me. Thanks!

certik:
Yes, good idea to do more smaller PRs. Easy to review, easy to merge.

PR #5119, opened 2024-10-21
Sync : Port some visitors + call the class to start visiting the tran…

assem2002:
…slation unit


certik (approved):
This should not affect the default route, so I think we can merge it.

PR #5120, opened 2024-10-22
Sync : asr_util.h

assem2002:


assem2002:
Should we branch on this implementation using `compiler_option.po.experimental_simplifier`

gxyd:
I don't think this PR balloons the temporary variables created, because:
1. So I went through the `--dump-all-passes` of `template_matrix_01.f90` and `template_04.f90` as well (the files whose ASR changed with this PR's changes), and I don't see any new temporary variables being created (after comparing the ASR of the current *main* and this PR's ASR dump).
2. Secondly, the changes in this PR just updated the `ttype` of `FunctionCall` ASR node

src/libasr/asr_utils.h
```diff
@@ -5720,6 +5720,37 @@ static inline ASR::asr_t* make_FunctionCall_t_util(
...
+                        m_dim_vec.m_start = i32one;
+                        m_dim_vec.m_length = ASRUtils::EXPR(ASR::make_ArraySize_t(al, m_dims[j].loc,
+                            a_args[i].m_value, i32one, ASRUtils::expr_type(i32one), nullptr));
+                        m_dims_vec.push_back(al, m_dim_vec);
+                    }
+                    m_dims = m_dims_vec.p;
+                    n_dims = m_dims_vec.size();
+                }
+                a_type = ASRUtils::make_Array_t_util(al, type->base.loc, type_, m_dims, n_dims,
+                    ASR::abiType::Source, false, ASR::array_physical_typeType::DescriptorArray, true);
```
gxyd, line 5748: What I'm concerned about is the use of `DescriptorArray` here, since I've a doubt if this should be `DescriptorArray`

tests/reference/asr-template_03b-0778e50.stdout
```diff
@@ -146,8 +151,13 @@
...
-                                                                [(()
-                                                                ())]
+                                                                [((IntegerConstant 1 (Integer 4) Decimal)
+                                                                (ArraySize
+                                                                    (Var 8 y)
+                                                                    (IntegerConstant 1 (Integer 4) Decimal)
+                                                                    (Integer 4)
+                                                                    ()
+                                                                ))]
                                                                 DescriptorArray
```
gxyd, line 161: Considering the updated `Array` ASR node, I'm not really sure if this should be a DescriptorArray.

gxyd:
Overall, I don't this PR has any ballooning of temporaries problem, so we don't need to use `compiler_option.po.experimental_simplifier`. Only problem being the ArrayPhysicalType being used in here, which instead of it being `DescriptorArray` probably should be `PointerToDataArray` I think.

Pranavchiku:
> so we don't need to use compiler_option.po.experimental_simplifier.

Let's use it, keep things separate, we can clean it up later. @certik what do you think?

tests/reference/asr-template_04-f41dd3e.stdout
```diff
@@ -355,8 +355,62 @@
...
                                                             (Integer 4)
-                                                            [(()
-                                                            ())]
+                                                            [((IntegerConstant 1 (Integer 4) Decimal)
+                                                            (IntegerBinOp
+                                                                (IntegerBinOp
+                                                                    (IntegerBinOp
+                                                                        (ArrayBound
+                                                                            (StructInstanceMember
+                                                                                (Var 198 solved)
```
Pranavchiku, line 364: this reference, do we actually need this to make the test working or it works without it? If it works without, then we need to guard the change.

gxyd: We need this update in reference tests following the change in `asr_utils.h`

Pranavchiku: Yes, I get that, but does the test pass without it as well? Like if there are any tests which fail without it and thus we need this patch, then let's keep it on by default, else let's guard it with `if(compiler_options.experimental_simplifier)`

assem2002:
@certik this is ready to merge.

src/libasr/asr_utils.h
```diff
@@ -51,6 +51,8 @@
 namespace LCompilers  {
 
     namespace ASRUtils  {
+    
+    static bool use_experimental_simplifier;
```
certik, line 55: I don't like using global variables, but in this case I think it's fine; it's only temporary (hopefully!), we will eventually split the utils in such a way that they are independent of this global flag. Alternatively we need to propagate this flag into function arguments.

However, there is another problem: the libasr is used in many different ways, such as lpython, or you construct ASR directly, not via the AST visitors. This flag will thus not be initialized, or initialized incorrectly. We are also breaking the premise that ASR behavior will not depend on any global external state.

I think the only way forward here is to expose it as an argument. How often is `make_FunctionCall_t_util` used? Just modify all cases, if it is too many, let me know, we'll have to figure out another solution.

assem2002: It exists in total of 31 places in 12 files.

certik: Ok, here is how to proceed.

* Create a global variable, set it to false
* In the main program when the experimental flag is set in `compiler_options`, also set it in this global variable
* create an issue documenting this, and explaining what needs to be done to remove it
* add a TODO comment next to this variable pointing to this issue.

Then we can merge this.

Pranavchiku: > Create a global variable, set it to false

We need this global variable for intrinsics as well.

certik: That's fine for now to unblock us to merge. Then we'll fix it after merging.

assem2002:
reminder to me : Update the code snippets in #5144 after merging this.

certik (approved):

PR #5122, opened 2024-10-22
syn: intrinsic_functions.h with main

Pranavchiku:
Towards #4700

### DIFF
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -670,7 +670,7 @@ namespace Aimag {
         declare_basic_variables("_lcompilers_aimag_" + type_to_str_python(arg_types[0]));
         fill_func_arg("x", arg_types[0]);
         auto result = declare(fn_name, return_type, ReturnVar);
-        body.push_back(al, b.Assignment(result, EXPR(ASR::make_ComplexIm_t(al, loc, 
+        body.push_back(al, b.Assignment(result, EXPR(ASR::make_ComplexIm_t(al, loc,
             args[0], return_type, nullptr))));
         ASR::symbol_t *f_sym = make_ASR_Function_t(fn_name, fn_symtab, dep, args,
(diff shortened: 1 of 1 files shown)


src/libasr/pass/intrinsic_functions.h
```diff
@@ -763,9 +763,9 @@ namespace Abs {
                 std::to_string(input_kind) + " output kind: " + std::to_string(output_kind),
                 loc, diagnostics);
         } else {
-            ASRUtils::require_impl(ASRUtils::check_equal_type(input_type, output_type, true),
-                "The input and output type of elemental intrinsics must exactly match, input type: " +
-                input_type_str + " output type: " + output_type_str, loc, diagnostics);
+            // ASRUtils::require_impl(ASRUtils::check_equal_type(input_type, output_type, true),
+            //     "The input and output type of elemental intrinsics must exactly match, input type: " +
+            //     input_type_str + " output type: " + output_type_str, loc, diagnostics);
```
certik, line 768: Why don't we need this check? If the `input_type` is not complex, we are now not checking anything. Is that correct?

certik:
Otherwise it's fine, but the above seems like a change we don't want.

src/libasr/pass/intrinsic_functions.h
gxyd: I'll check this locally to see if we need this condition in *simplifier_pass* branch or not, if we don't I'll send a PR to remove this from *simplifier_pass* branch and then undo it from here.

gxyd: The CI for PR here: https://github.com/lfortran/lfortran/pull/5132 uncommenting this change, it passes CI.

gxyd (approved):
LGTM!

src/libasr/pass/intrinsic_functions.h
certik: Excellent!

certik:
Good.

PR #5123, opened 2024-10-22
Merge main into simplifier pass

gxyd:


.github/workflows/CI.yml
```diff
@@ -1281,6 +1289,8 @@ jobs:
             cd integration_tests
             ./run_tests.py -b fortran -j1
             ./run_tests.py -b fortran -f -j1
+            ./run_tests.py -b fortran -j1 --experimental-simplifier
```
gxyd, line 1292: @certik @Pranavchiku , just a gentle ping. This is **expected**.

These changes came directly after merge (as there were no merge conflicts for these changes), and slowly diff between *main* and *simplifier_pass* will vanish.

integration_tests/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ endif ()
```
gxyd, line 9: @assem2002 , can you please give a quick look to see that the tests actually run with *simplifier_pass* turned on right?

gxyd: I didn't phrase the question correctly, but my point is simply, that if you can check that the changes after merging *main* here look alright to you?

certik:
This is great. Yes, changes should slowly vanish, until only one change remains: making the experimental_simplifier=true, otherwise everything else is merged.

integration_tests/CMakeLists.txt
assem2002: They exist at the same line number in main branch. 
https://github.com/lfortran/lfortran/blob/21630bc62c7fc0e98fdafe65873910f865bde5b1/integration_tests/CMakeLists.txt#L12C1-L14C1

assem2002: Yeah. This `--experimental_simplifier` option is useless on simplifier branch as has `compiler_option.po.experimental_simplifier` set to  TRUE by default.

src/libasr/pass/simplifier.cpp
```diff
@@ -904,7 +904,7 @@ class ArgSimplifier: public ASR::CallReplacerOnExpressionsVisitor<ArgSimplifier>
 
     ArgSimplifier(Allocator& al_, ExprsWithTargetType& exprs_with_target_, bool realloc_lhs_) :
         al(al_), current_body(nullptr), exprs_with_target(exprs_with_target_),
-        realloc_lhs(realloc_lhs_) {}
+        realloc_lhs(realloc_lhs_) {(void)realloc_lhs; /*Silence-Warning*/}
```
assem2002, line 907: maybe you can undo that to remain as a diff between the two branch (maybe you can leave it, It causes no harm)

gxyd: I think this change is in *main* as well, it makes sense to me to have this here as well, reducing the diff size between *main* and *simplifier_pass* branches

PR #5128, opened 2024-10-23
Sync : Port class 'ArgSimplifier'

assem2002:


assem2002:
Should we separate PR into maybe 2 smaller ones?

integration_tests/CMakeLists.txt
```diff
@@ -419,7 +419,7 @@ RUN(NAME arrays_01 LABELS gfortran cpp llvm llvmStackArray wasm)
 RUN(NAME arrays_01_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
 RUN(NAME arrays_02_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
 RUN(NAME arrays_03_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran)
-RUN(NAME matrix_01_transpose LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME matrix_01_transpose LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER)
```
gxyd, line 422: Can you help me understand:
*  so before this PR, since it had no `NO_EXPERIMENTAL_SIMPLIFIER` label, so this **was** getting tested with `--experimental-simplifier` flag (like e.g. `./run_tests.py -b llvm --experimental-simplifier`)?
* but now with `NO_EXPERIMENTAL_SIMPLIFIER`, this isn't tested with `experimental_simplifier`?

assem2002: Yeah exaclty. They fail with the changes ported from the simplifier branch so I added a flag to not test them with the simplifier pass( They should work) after porting all changes. Only the commented out tests on the simplifier branch are the one that should really be labeled with NO-EXPERIMENTAL-SIMPLIFIER

gxyd: > Only the commented out tests on the simplifier branch are the one that should really be labeled with NO-EXPERIMENTAL-SIMPLIFIER

makes sense.

gxyd:
> Should we separate PR into maybe 2 smaller ones?

I'm happy with the changes as is.

gxyd (approved):
LGTM!

PR #5130, opened 2024-10-23
chore: sync `where` pass with main

kmr-srbh:


kmr-srbh:
The CI will fail for now as the new rewritten `where` pass depends on the new `array_op_simplifier` pass. It will work once we have the `array_op_simplifier` changes in place. Let's wait with these changes for now.

kmr-srbh:
The `array_op_simplifier` pass is merged now, let's see if the CI passes.

gxyd:
We'll need a corresponding PR against the *simplifier_pass* branch, where to create `where_simplifier.cpp` and sync *simplifier_pass* branch's `where.cpp` with the current *main*, so that only `where_simplifier.cpp` (or `replace_where_simplifier.cpp`) are the only new files for it.

gxyd:
> We'll need a corresponding PR against the _simplifier_pass_ branch, where to create `where_simplifier.cpp` and sync _simplifier_pass_ branch's `where.cpp` with the current _main_, so that only `where_simplifier.cpp` (or `replace_where_simplifier.cpp`) are the only new files for it.

That will actually test for us, whether the things work or not.

kmr-srbh:
@gxyd `where_simplifier.cpp` is the same file as `where.cpp` on `simplifier_pass` branch.

certik (approved):
I think this is fine, as it adds the correct where pass that works with simplifier. We need that. The pass is small, so shouldn't be a big deal.

Then in the simplifier branch, I think we can introduce the same structure.

PR #5148, opened 2024-10-24
CI: update link for stdlib which contains two more workarounds

gxyd:
## Description

this is being done, as the *experimental simplifier* (i.e. one with simplifier pass) doesn't compile stdlib without the newly added two workarounds in stdlib

The two workarounds in stdlib being:

* XX: workaround for Casting error (i.e. commit hash: f06d8)
* XX: workaround to avoid asr_verify error in simplifier_pass branch (i.e. commit hash: 57c4)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -720,12 +720,12 @@ jobs:
       - name: Test stdlib
         shell: bash -e -x -l {0}
         run: |
-            git clone https://github.com/czgdp1807/stdlib.git
+            git clone https://github.com/Pranavchiku/stdlib-fortran-lang.git stdlib
             cd stdlib
             export PATH="$(pwd)/../src/bin:$PATH"
 
-            git checkout lf20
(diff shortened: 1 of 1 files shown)


certik:
So this unfortunately regresses in main as well.

Why don't we simply check out the modified version only for the simplifier tests? That should work.

gxyd:
That's a valid point, I think once we test third party tests with `--experimental-simplifier` in CI in *main*, we can do that.

So, until then, this PR isn't needed, so I'll close this for now.

certik:
Actually we do not even check stdlib with the simplifier branch on, correct? So this PR is not necessary, since it will not break.

Instead, if we want to test stdlib, right after it is finished with the default path, check out the modified version and test it with the simplifier. However, I would do this after we are done merging, since it would I think break anyway right now.

gxyd:
> Actually we do not even check stdlib with the simplifier branch on, correct?

That's right. We only test with the *default passes* (i.e. without simplifier pass) the third party tests currently.

PR #5149, opened 2024-10-24
Mod refactoring changes and extended logic to real

AnshMehta1:
In continuation of :- #4645 
Basically , the logic is extended to promote lower kind in case of real.
Also , some refactoring changes are made like removeing ```create_type``` function which was added to create type with given kind but the workaround achieved by a earlier function ```set_kind_to_ttype_t```.

### DIFF
--- a/integration_tests/intrinsics_78.f90
+++ b/integration_tests/intrinsics_78.f90
@@ -1,8 +1,8 @@
 program intrinsics_78
     implicit none
     integer :: ix, iy, iresult
-    integer(8) :: i
-    integer(4) :: j
+    integer(8) :: j
+    integer(4) :: i
     real(8) :: a
     real(4) :: b
     real :: rx, ry, rresult
@@ -92,7 +92,7 @@ program intrinsics_78
     ! Test integer values with different kinds
     i = 121
     j = 121
-    iresult = kind(mod(i,j))
+    iresult = kind(mod(j,i))
     print *, "Test 14: kind(mod(", i, ",", j, ")) = ", iresult 
     if(iresult /= 8) error stop "Test 14 failed"
 
--- a/integration_tests/intrinsics_78.f90
+++ b/integration_tests/intrinsics_78.f90
@@ -1,8 +1,10 @@
 program intrinsics_78
     implicit none
     integer :: ix, iy, iresult
-    integer(8) :: j
-    integer(4) :: i
+    integer(8) :: i
+    integer(4) :: j
+    real(8) :: a
+    real(4) :: b
     real :: rx, ry, rresult
     double precision :: dx, dy, dresult
 
@@ -90,7 +92,14 @@ program intrinsics_78
     ! Test integer values with different kinds
     i = 121
     j = 121
-    iresult = kind(mod(j,i))
+    iresult = kind(mod(i,j))
     print *, "Test 14: kind(mod(", i, ",", j, ")) = ", iresult 
     if(iresult /= 8) error stop "Test 14 failed"
+
+    ! Test real values with different kinds
+    a = 121.882
+    b = 121.882
+    iresult = kind(mod(a,b))
+    print *, "Test 15: kind(mod(", a, ",", b, ")) = ", iresult 
+    if(iresult /= 8) error stop "Test 15 failed"
 end program intrinsics_78
--- a/src/libasr/intrinsic_func_registry_util_gen.py
+++ b/src/libasr/intrinsic_func_registry_util_gen.py
@@ -1042,7 +1042,7 @@ def add_create_func_return_src(func_name):
             src += indent * 3 + f"upper_kind = std::max(upper_kind,ASRUtils::extract_kind_from_ttype_t(expr_type(args[i])));\n"
             src += indent * 2 + f"}}\n"
             src += indent * 2 + f"ASR::ttype_t* type_ = expr_duplicator.duplicate_ttype(ASRUtils::type_get_past_array_pointer_allocatable(expr_type(args[0])));\n"
-            src += indent * 2 + f"type_ = b.create_type(type_, upper_kind);\n"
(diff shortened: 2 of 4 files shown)


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -1042,7 +1042,7 @@ def add_create_func_return_src(func_name):
             src += indent * 3 + f"upper_kind = std::max(upper_kind,ASRUtils::extract_kind_from_ttype_t(expr_type(args[i])));\n"
             src += indent * 2 + f"}}\n"
             src += indent * 2 + f"ASR::ttype_t* type_ = expr_duplicator.duplicate_ttype(ASRUtils::type_get_past_array_pointer_allocatable(expr_type(args[0])));\n"
-            src += indent * 2 + f"type_ = b.create_type(type_, upper_kind);\n"
+            src += indent * 2 + f"type_ = b.duplicate_type_change_kind(type_, upper_kind);\n"
```
gxyd, line 1045: I think we already have a function in `asr_utils.h` called `set_kind_to_ttype_t`, which probably does the same thing as you are looking for. Can you try and see if that works for you?

If `set_kind_to_ttype_t` works for you, maybe, document it a little so that it becomes clear for it's usage?

integration_tests/intrinsics_78.f90
```diff
@@ -90,7 +92,14 @@ program intrinsics_78
     ! Test integer values with different kinds
     i = 121
     j = 121
-    iresult = kind(mod(j,i))
+    iresult = kind(mod(i,j))
```
gxyd, line 95: do we already have a test which ensures that the kind of `kind(mod(j, i))` is also `8`? If not, then we can have two tests?

src/libasr/pass/intrinsic_functions.h
```diff
@@ -3276,32 +3276,45 @@ namespace Mod {
...
+                ASR::expr_t* arg0 = b.r2r_t(args[0], new_type);
+                ASR::expr_t* arg1 = b.r2r_t(args[1], new_type);
 
-                    body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.Div(arg0, arg1))))); 
+                if(upper_kind == 4){
+                    body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.i2r_t(b.r2i_t(b.Div(arg0, arg1), real32), real32)))));
                 }
                 else{
-                    body.push_back(al, b.Assignment(result, b.Sub(args[0], b.Mul(args[1], b.Div(args[0], args[1])))));
+                    body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.i2r_t(b.r2i_t(b.Div(arg0, arg1), real64), real64)))));
```
gxyd, line 3292: It's difficult to read long lines of code, but I don't have a nice solution without wrapping it making it occupy 20 lines, so I'm fine as it is as well.

src/libasr/intrinsic_func_registry_util_gen.py
AnshMehta1: ok , I will try this out

integration_tests/intrinsics_78.f90
AnshMehta1: I think this two tests are sufficient 
```
! Test integer values with different kinds
    i = 121
    j = 121
    iresult = kind(mod(i,j))
    print *, "Test 14: kind(mod(", i, ",", j, ")) = ", iresult 
    if(iresult /= 8) error stop "Test 14 failed"

    ! Test real values with different kinds
    a = 121.882
    b = 121.882
    iresult = kind(mod(a,b))
    print *, "Test 15: kind(mod(", a, ",", b, ")) = ", iresult 
    if(iresult /= 8) error stop "Test 15 failed"
```

certik: Let's not modify the original test, which was good.

Add a new test that fails in main, and this PR fixes it.

AnshMehta1: Ok , will do so .

AnshMehta1: Done with changes , new test that the PR fixes is for real with different kinds.

certik (approved):

Pranavchiku (changes requested):
```cpp
        int upper_kind = std::max(kind, kind2);
        if (is_real(*arg_types[1])) {
            ASR::ttype_t* new_type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, upper_kind));
            ASR::expr_t* arg0 = b.r2r_t(args[0], new_type);
            ASR::expr_t* arg1 = b.r2r_t(args[1], new_type);
            if(upper_kind == 4){
                body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.i2r_t(b.r2i_t(b.Div(arg0, arg1), real32), real32)))));
            }
            else{
                body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.i2r_t(b.r2i_t(b.Div(arg0, arg1), real64), real64)))));
            }
        } else {
            ASR::ttype_t* new_type = ASRUtils::TYPE(ASR::make_Integer_t(al, loc, upper_kind));
            ASR::expr_t* arg0 = b.i2i_t(args[0], new_type);
            ASR::expr_t* arg1 = b.i2i_t(args[1], new_type);
            body.push_back(al, b.Assignment(result, b.Sub(arg0, b.Mul(arg1, b.Div(arg0, arg1)))));
        }
```

This shall be sufficient

Pranavchiku (approved):

Pranavchiku:
From next PR, let's try to make meaningful commit messages, you can use space and avoid `_`. I am squashing this PR.

PR #5156, opened 2024-10-24
Mod intrinsic with different kinds for simpilier pass

AnshMehta1:
PR for simplifier pass , based on PR #5149


gxyd:
Could you please link the PR which is/was created against *main*?

PR #5157, opened 2024-10-24
DEV: Use ``FunctionParam`` in FunctionCall type

czgdp1807:
Next steps,

Replace `FunctionParam` in `FunctionCall` type after the temporaries are given to the arguments.


certik:
This is a temporary solution, the full solution will be https://github.com/lfortran/lfortran/issues/5201.

PR #5125, opened 2024-10-22
chore: sync `array_op` pass with main

kmr-srbh:

PR #5160, opened 2024-10-24
chore: sync `asdl_cpp.py` with main

kmr-srbh:


kmr-srbh:
Some changes related to this were pushed as part of #5125.

kmr-srbh:
The `asdl_cpp.py` file introduces some new changes. The integration tests pass, so I think it's safe.

certik:
This is hard for me to quickly review. Does this change any ASR in default mode?

src/libasr/asdl_cpp.py
```diff
@@ -791,14 +804,14 @@ def visitField(self, field):
                 self.emit("for (size_t i=0; i<x.n_%s; i++) {" % field.name, level)
                 if field.type in products:
                     if field.type == "expr":
-                        self.insert_call_replacer_code(field.name, level + 1, "[i]")
-                        self.emit("if( x.m_%s[i] )" % (field.name), level)
+                        self.insert_call_replacer_code(field.name, level + 1, field.opt, "[i]")
+                        self.emit("if( x.m_%s[i] && visit_expr_after_replacement )" % (field.name), level)
```
certik, line 808: Is `visit_expr_after_replacement` true in default path?

src/libasr/asdl_cpp.py
```diff
@@ -584,7 +588,10 @@ def visitField(self, field):
                     if field.type != "symbol":
                         self.used = True
                         if field.opt:
-                            self.emit("if (x.m_%s)" % field.name, 2)
+                            if field.name == "value":
+                                self.emit("if (x.m_%s && visit_compile_time_value)" % field.name, 2)
+                            else:
+                                self.emit("if (x.m_%s)" % field.name, 2)
```
certik, line 594: Things do not change in default path until here, so the above is good.

src/libasr/asdl_cpp.py
```diff
@@ -772,11 +780,16 @@ def make_visitor(self, name, fields):
...
+    def insert_call_replacer_code(self, name, level, opt, index=""):
+        one_or_zero = (name == "value" or name == "symbolic_value") and opt
+        if (name == "value" or name == "symbolic_value") and opt:
+            self.emit("if (call_replacer_on_value) {", level)
+        self.emit("ASR::expr_t** current_expr_copy_%d = current_expr;" % (self.current_expr_copy_variable_count), level + one_or_zero)
+        self.emit("current_expr = const_cast<ASR::expr_t**>(&(x.m_%s%s));" % (name, index), level + one_or_zero)
+        self.emit("self().call_replacer();", level + one_or_zero)
+        self.emit("current_expr = current_expr_copy_%d;" % (self.current_expr_copy_variable_count), level + one_or_zero)
+        if (name == "value" or name == "symbolic_value") and opt:
+            self.emit("}", level)
```
certik, line 792: I think this part changes default path. What are the changes? I think it changes something with if value is replaced.

certik (approved):
If `call_replacer_on_value` and `visit_expr_after_replacement` are true in default path, then this is good.

Otherwise let's discuss what changes this does.

gxyd:
@kmr-srbh after this change in `asdl_cpp.py` is merged, can you confirm that there is no change needed to be ported from *simplifier_pass* branch for `asdl_cpp.py` file?

gxyd:
Before we merge this PR, I'll try **locally** to merge this PR into *main* and then try to merge *main* into *simplifier_pass*, that will help me see if any problems occur or not. I'll let you know how that goes with the change.

gxyd:
> If `call_replacer_on_value` and `visit_expr_after_replacement` are true in default path, then this is good.
> 
> Otherwise let's discuss what changes this does.

Yes, that's the case @certik

gxyd (approved):
the changes look good to me. I'm merging this.

src/libasr/asdl_cpp.py
kmr-srbh: Apart from the `if` statements, we have this in `main` also. The change here was to visit compile-time  values (both `m_value` and `m_symbolic_value`) optionally. https://github.com/lfortran/lfortran/commit/1a618ad08eba76ec4a9bb985cc67396fd9f2b713.

src/libasr/asdl_cpp.py
kmr-srbh: Yes. It was introduced in #5125.

PR #5161, opened 2024-10-24
chore: sync `ASR.asdl` with main

kmr-srbh:


kmr-srbh:
'temporary needed' reworded to 'temporary created'.

certik:
We should not merge this. The `ASR.asdl` should be without comments to keep it short.

We can document this in our documentation separately.

Issue #5173, opened 2024-10-26
Bad character declaration allowed


harperjf:
The f2023 standard C805 says this  program is not valid because linelen is not a specification expression. But lfortran 0.41.0 prints '?' with it.
```
  integer            :: linelen = 80
  character(linelen) :: line
  line = '?'
  print *, trim(line)
end program
```

kmr-srbh:
Thanks for reporting this! 

It looks like we are allowing variables as length and dimension expressions (by not disallowing it). Seems like our current handling of dimension expressions is weak. Here is another MRE:

```fortran
program main
    integer, dimension("abcd") :: line
end program
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran ./examples/example.f90
Bus error (core dumped)
```

kmr-srbh:
@AnshMehta1 would you like to take a look into this?

We need to take care of this between AST -> ASR. The AST for the simplified MRE:

```fortran
program main
    integer :: linelen = 2
    character, dimension(2) :: line
end program
```
is

```clojure
(TranslationUnit
    [(Program
        main
        ()
        []
        []
        [(Declaration
            (AttrType
                TypeInteger
                []
                ()
                ()
                None
            )
            []
            [(linelen
            []
            []
            ()
            2
            Equal
            ())]
            ()
        )
        (Declaration
            (AttrType
                TypeCharacter
                []
                ()
                ()
                None
            )
            [(AttrDimension
                [(1
                linelen
                DimensionExpr)]
            )]
            [(line
            []
            []
            ()
            ()
            None
            ())]
            ()
        )]
        []
        []
    )]
)
```

Note the `AttrDimension` attribute inside the `Declaration` node. We are concerned with it. Here, the start dimension is `1` and end dimension is `linelen`, which is a non-constant variable. We essentially want to allow only constant values here. So the only correct dimension values can be either a `parameter` annotated integer variable, or a constant integer. We process the dimensions while declaration through the function `process_dims()` inside `ast_common_visitor.h`. We need to throw an error when the above stated condition is false, for both the start and end dimension.

AnshMehta1:
> @AnshMehta1 would you like to take a look into this?
> 
> We need to take care of this between AST -> ASR. The AST for the simplified MRE:
> 
> ```fortran
> program main
>     integer :: linelen = 2
>     character, dimension(2) :: line
> end program
> ```
> 
> is
> 
> ```clojure
[...]

PR #5176, opened 2024-10-27
Adding Semantic Check for non-constant variable in dimension expression

AnshMehta1:
Towards :- #5173


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1340,7 +1340,24 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             if (m_dim[i].m_start) {
                 this->visit_expr(*m_dim[i].m_start);
                 dim.m_start = ASRUtils::EXPR(tmp);
-            } else {
+                ASR::expr_t* start = dim.m_start;
+                if(ASR::is_a<ASR::Var_t>(*start)) {
```
certik, line 1344: Put space after `if` and before `(`. The same for all ifs.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1355,6 +1372,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         ASRUtils::expr_value(end) == nullptr) ) {
                             end = get_transformed_function_call(end_sym);
                         }
+                    else if(ASR::is_a<ASR::Variable_t>(*end_sym)) {
```
certik, line 1375: Don't know what is going on here, but it should look like this `} else if (...) {`, to make it clear which if statement this belongs to. It might be the previous `}` is incorrectly indented.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1340,7 +1340,24 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        ASR::Variable_t* start_variable = ASR::down_cast<ASR::Variable_t>(start_sym);
+
+                        if(start_variable->m_storage != ASR::storage_typeType::Parameter) {
+
+                            if(!in_Subroutine && (symbol_scope->counter == current_scope->counter) && !ASR::is_a<ASR::IntegerConstant_t>(*start)) {
+                                throw SemanticError("Explicit shaped array with nonconstant bounds at ",m_dim[i].m_start->base.loc);
+                            }
+                        }
+                    }
+                }
```
kmr-srbh, line 1359: We need to throw a semantic error for the MRE in https://github.com/lfortran/lfortran/issues/5173#issuecomment-2439503478 also. Please add an `else` clause here which checks for `IntegerConstant`.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1355,6 +1372,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        if(end_variable->m_storage != ASR::storage_typeType::Parameter) {
+
+                            if(!in_Subroutine && (symbol_scope->counter == current_scope->counter) && !ASR::is_a<ASR::IntegerConstant_t>(*end)) {
+                                throw SemanticError("Explicit shaped array with nonconstant bounds at ",m_dim[i].m_end->base.loc);
+                            }
+                        }
+                    }
                 } else if(ASR::is_a<ASR::IntegerBinOp_t>(*end)) {
                     end = convert_integer_binop_to_function_call(end, is_argument);
                 }
```
kmr-srbh, line 1387: Same for the end dimension also.

kmr-srbh (changes requested):

src/lfortran/semantics/ast_common_visitor.h
AnshMehta1: Some changes are to be made , will work on these changes today.

src/lfortran/semantics/ast_common_visitor.h
AnshMehta1: Their was intendation problem , rectified it.

AnshMehta1:
> Test has to be added.

Added

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1354,10 +1409,58 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    }
+                    else {
+
+                        if ((end->type != ASR::exprType::ArraySize) && (end->type != ASR::exprType::StringLen) 
+                        && (end->type != ASR::exprType::IntegerConstant) && (end->type != ASR::exprType::IntegerBinOp) 
+                        && (end->type != ASR::exprType::IntegerUnaryMinus)) {
+                            throw SemanticError("Expecting a scalar integer or parameter annotated integer variable at ",m_dim[i].m_end->base.loc);
+                        }
+                    }
+                }
```
certik, line 1463: This seems like a lot of duplicated code (12x) for one error message. 

First of all, it should not say `at` at the end. We can just manually modify all the 12 occurrences above.

But let's refactor this. What exactly is the condition under which this error message should be reported?

Is the reason to just extract the type? I think we already have a function for this in ASR utils. If not, let's write one. 

There should be just one place to report this error, not 12x.

AnshMehta1: For parameter annoted integer variable and IntegerConstant , error must not be thrown. But their are some functions like len for string and size for array where array shouldn't be thrown as their return type is integer.
Also , their are some functions for array like maxval having return type integer for which array is not thrown .
Basically , for functions with return type integer , error shouldn't be thrown. This thing is enlisted using else ifs

AnshMehta1: Ok , figured out refactoring , will make the changes .

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1337,10 +1337,38 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        ASR::Variable_t* start_variable = ASR::down_cast<ASR::Variable_t>(start_sym);
+
+                        if (start_variable->m_storage != ASR::storage_typeType::Parameter) {
+
+                            if(!in_Subroutine && (symbol_scope->counter == current_scope->counter)) {
+                                throw SemanticError("Expecting a scalar integer or parameter annotated integer variable ",m_dim[i].m_start->base.loc);
+                            }
+                        }
+                    }
+                    else {
```
certik, line 1359: Use this syntax:
```suggestion
                    } else {
```
The same at other places below.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1337,10 +1337,38 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         for (size_t i=0; i<n_dim; i++) {
             ASR::dimension_t dim; dim.m_length = nullptr; dim.m_start = nullptr;
             dim.loc = m_dim[i].loc;
-            if (m_dim[i].m_start) {
+           if (m_dim[i].m_start) {
```
certik, line 1340: This seems incorrectly indented.

certik:
This is better. But the same error is still reported 6 times. Given that we are not distinguishing the different cases, why not write a function that determines if the given entity satisfies the requirements? It will return true/false. Then we do `if (!satisfy_requirements(x)) report the error`.

AnshMehta1:
> This is better. But the same error is still reported 6 times. Given that we are not distinguishing the different cases, why not writing a function that determines if the given entity satisfies the requirements? It will return true/false. Then we do `if (!satisfy_requirements(x)) report the error`.

Idea seems good , will do the change tomorrow morning.

AnshMehta1:
I added the function for testing error check. Now , error is reported 2 times based on function . 1 is error check for start dimension and other is error check for end dimension. Let me know if their is a need to change the name of function or any other thing.

src/lfortran/semantics/ast_common_visitor.h
AnshMehta1: Done

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1327,6 +1327,35 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             left, end_bin_op->m_op, right, end_bin_op->m_type, end_bin_op->m_value));
     }
 
+    bool dimensionAttribute_error_check(ASR::expr_t* dim_expr) {
```
certik, line 1330: Use the following naming convention to be consistent with the rest of the code:
```suggestion
    bool dimension_attribute_error_check(ASR::expr_t* dim_expr) {
```

certik (approved):
I think this looks good! Left some minor comments to fix, after that we can merge.

certik (approved):

PR #5179, opened 2024-10-27
chore: bring back changes from older `main` using `use_experimental_simplifier` guard

kmr-srbh:


kmr-srbh:
#5178 and #5179 are the same changes. I opened a PR for both `main` and `simplifier_pass` so we do not increase the diff.

PR #5180, opened 2024-10-27
chore: shorten diff between `main` and `simplifier_pass`

kmr-srbh:


kmr-srbh:
Yes, I added the whitespace which `main` already had so the diff could be small.

PR #5155, opened 2024-10-24
Merge main into simplifier pass

gxyd:

PR #5186, opened 2024-10-28
sync where.cpp

assem2002:


gxyd:
you need to use `where_simplifier` in tests.toml file, instead of `where` now, and then use the updated reference files. This is something similar to what we did in `array_op_simplifier.cpp`

gxyd:
make sure to revert the changes to older `where` reference files, we wouldn't want to have those updated (any longer)

gxyd:
this might help: https://github.com/lfortran/lfortran/pull/5155#issuecomment-2436851486

assem2002:
yeah I just realized that we have reference tests for `where` pass. I'll update these.

tests/reference/pass_where-where_02-1839d96.stdout
```diff
@@ -7,6 +7,22 @@
                     (SymbolTable
                         2
                         {
+                            __1_k:
```
gxyd, line 10: I would restore these `where_*` test file, only have the new `where_simplifier_*` reference files added.

gxyd: These would add un-necessary diff, which don't make sense to have on *simplifier_pass* branch

assem2002: I think that's what you've warned me about here  https://github.com/lfortran/lfortran/pull/5186#issuecomment-2440630917. I just reverted the reference tests to be matching the main. This should result in unnecessary diff.

assem2002: If you do `rm tests/reference/*` in the simplifier branch and then run `python3 run_tests -u` you would find the following output.
```
       deleted:    tests/reference/pass_array_op-array15-079d0cd.json
	deleted:    tests/reference/pass_array_op-array15-079d0cd.stdout
	deleted:    tests/reference/pass_array_op-arrays_op_10-be689f7.json
	deleted:    tests/reference/pass_array_op-arrays_op_10-be689f7.stdout
	deleted:    tests/reference/pass_array_op-modules_43-6930881.json
	deleted:    tests/reference/pass_array_op-modules_43-6930881.stdout
``` 
So I followed the same approach of leaving reverting back the reference files to match the main branch.

gxyd: Ah, I see.

gxyd: This looks good then, perfect.

kmr-srbh: Please do `git checkout simplifier_pass -- ./tests/reference`. This will revert changes to modified files. You can then do `./run_tests.py -u` to create the reference files for `where_simplifier`.

kmr-srbh: As this PR was merged, using `simplifier_pass` won't work. Please use the commit hash before you updated the tests. I think it's this one - https://github.com/lfortran/lfortran/pull/5186/commits/aaaf9d8455e5bbb542291d41e1b352c41fd71c95.

kmr-srbh: I am confused, but why do we need to update the reference files for `where` pass?

gxyd: I don't think there is anything to be done here.

PR #5187, opened 2024-10-28
use 'external' variable instead of 'static' variable

gxyd:
## Description

Because of use of `static` variable, `make_FunctionCall_t_util` always uses the default value set for it, we need to make it an `external` variable instead, that's one solution.

Other solution to the problem could be use to setter and getter for `use_experimental_simplifier` variable.

## TODO

- [ ] test third party packages

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -382,10 +382,10 @@ RUN(NAME functions_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME functions_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
-RUN(NAME functions_06 LABELS gfortran llvm NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME functions_06 LABELS gfortran llvm)
 RUN(NAME functions_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME functions_09 LABELS gfortran llvm NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME functions_09 LABELS gfortran llvm)
 RUN(NAME functions_10 LABELS gfortran llvm)
 RUN(NAME functions_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME functions_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran)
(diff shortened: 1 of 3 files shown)


gxyd:
I'll make the remaining change once the initial review is done

integration_tests/CMakeLists.txt
```diff
@@ -1801,12 +1801,12 @@ RUN(NAME openmp_24 LABELS gfortran llvm_omp GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_25 LABELS gfortran llvm_omp GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_26 LABELS gfortran llvm_omp GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_27 LABELS gfortran llvm_omp GFORTRAN_ARGS -fopenmp)
-RUN(NAME openmp_28 LABELS gfortran llvm_omp GFORTRAN_ARGS -fopenmp NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME openmp_28 LABELS gfortran llvm_omp NO_EXPERIMENTAL_SIMPLIFIER GFORTRAN_ARGS -fopenmp)
```
gxyd, line 1804: this test doesn't work on *simplifier_pass* branch, that's why we need to use `NO_EXPERIMENTAL_SIMPLIFIER` label.

integration_tests/CMakeLists.txt
```diff
@@ -382,22 +382,22 @@ RUN(NAME functions_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
...
 RUN(NAME functions_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran)
 RUN(NAME functions_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran)
 RUN(NAME functions_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_16 LABELS gfortran llvm)
 RUN(NAME functions_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME functions_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER)
-RUN(NAME functions_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER)
-RUN(NAME functions_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME functions_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME functions_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
gxyd, line 397: this is one example of a test case, which currently doesn't work in *main* when `--experimental-simplifier` flag is specified, i.e.:
```console
$ lfortran integration_tests/functions_19.f90 --experimental-simplifier
...
some error
...
```

with the change to make `use_experimental_simplifier` as `extern` instead of `static` helps fix the error.

gxyd: though the test case `integration_tests/functions_19.f90` already works in `simplifier_pass` branch (with/without `--experimental-simplifier` flag)

assem2002: Yeah that's great catch! I haven't notice that the flag get wrong value as the flag gets created for each file.

assem2002: `extern` seems to be the best option and declaring `use_experimental_simplifier` variable in `asr_utils.cpp` is perfect.

assem2002 (approved):

kmr-srbh (approved):
I think this change looks good! Using `extern` adds true global nature to the variable. Great catch!

integration_tests/CMakeLists.txt
```diff
@@ -862,9 +862,9 @@ RUN(NAME intrinsics_260 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # sum
 RUN(NAME intrinsics_261 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # ceiling
 RUN(NAME intrinsics_262 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # floor
 RUN(NAME intrinsics_263 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # maxval
-RUN(NAME intrinsics_264 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # sin
-RUN(NAME intrinsics_265 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # cos
-RUN(NAME intrinsics_266 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # tan
+RUN(NAME intrinsics_264 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER) # sin
```
gxyd, line 865: The reason to add `NO_EXPERMENTAL_SIMPLIFIER` flag here is as this doesn't pass with `--fast` flag currently when experimental simplifier is enabled.

gxyd:
Thanks for the review @kmr-srbh @assem2002 , I'll merge the PR once the tests pass.

PR #5189, opened 2024-10-28
sync: pass_manager with 'main' branch

gxyd:


gxyd:
I'll re-think about the changes once I'm back from lunch.

src/libasr/pass/pass_manager.h
```diff
@@ -393,13 +394,9 @@ namespace LCompilers {
                 apply_passes(al, asr, _user_defined_passes, pass_options,
                     diagnostics);
             } else if( apply_default_passes ) {
-                if( pass_options.fast && pass_options.experimental_simplifier ) {
```
gxyd, line 396: Ah yes, so in *simplifier_pass* we've two conditions repeated here, there are only 4 possible conditions with all combinations of `pass_options.fast` and `pass_options.experimental_simplifier`

kmr-srbh (approved):

PR #5191, opened 2024-10-28
Merge main into simplifier pass

gxyd:


src/bin/lfortran.cpp
```diff
@@ -2264,7 +2264,10 @@ int main_app(int argc, char *argv[]) {
     app.get_formatter()->column_width(25);
     app.require_subcommand(0, 1);
     CLI11_PARSE(app, argc, argv);
-    LCompilers::ASRUtils::use_experimental_simplifier = compiler_options.po.experimental_simplifier;
+    // this is hard-code here on 'simplifier_pass' branch
+    // intentionally, to make sure that all test on this branch
+    // run with experimental simplifier enabled
+    LCompilers::ASRUtils::use_experimental_simplifier = true;
```
gxyd, line 2270: creating this change exclusively only on *simplifier_pass* branch.

gxyd: what this mean is, whether we run `lfortran test.f90 --experimental-simplifier` or `lfortran test.f90`, will always produce the same result and we just need to change this to `compiler_options.use_experimental_simplifier` to get the behavior same as what we would get on *main*.

PR #5182, opened 2024-10-27
Simplifier pass PR

certik:
GitHub made a mistake and changed the branch in #4700, and there does not seem to be a way to fix it, so I am opening a new PR.

PR #5194, opened 2024-10-28
Enable all intrinsics for simplifier

certik:
These all pass in the simplifier PR #5182, so they should pass in `main` also.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -798,7 +798,7 @@ RUN(NAME intrinsics_196 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # bessel_
 RUN(NAME intrinsics_197 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # bessel_y0
 RUN(NAME intrinsics_198 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # sum
 RUN(NAME intrinsics_199 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # count
-RUN(NAME intrinsics_200 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # pack
+RUN(NAME intrinsics_200 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran NO_EXPERIMENTAL_SIMPLIFIER) # pack
 RUN(NAME intrinsics_201 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # any
 RUN(NAME intrinsics_202 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #parameter
 RUN(NAME intrinsics_203 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # asind
@@ -885,7 +885,7 @@ RUN(NAME intrinsics_283 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # asinh
 RUN(NAME intrinsics_284 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # acosh
 RUN(NAME intrinsics_285 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # real
 RUN(NAME intrinsics_286 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # atanh
-RUN(NAME intrinsics_288 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # spread
+RUN(NAME intrinsics_288 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran NO_EXPERIMENTAL_SIMPLIFIER) # spread
 RUN(NAME intrinsics_289 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # ishftc
 RUN(NAME intrinsics_290 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # log10
 RUN(NAME intrinsics_291 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # gamma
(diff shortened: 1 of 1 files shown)


kmr-srbh:
These tests are yet to be run with the `simplifier_pass` branch - https://github.com/lfortran/lfortran/pull/5182/files#diff-67c8fb3f4b6629c703ecb1c20661e195a435fe18141d370cd76cd2abc39021a8. I'll try testing them and see if they pass now.

integration_tests/CMakeLists.txt
```diff
@@ -598,7 +598,7 @@ RUN(NAME intrinsics_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # sin
 RUN(NAME intrinsics_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # cos
 RUN(NAME intrinsics_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # tan
-RUN(NAME intrinsics_04s LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER) # ctan
+RUN(NAME intrinsics_04s LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # ctan
```
gxyd, line 601: This will not pass, cause this enabled `--fast` test as well with `--experimental-simplifer`, which currently fails and is disabled in *simplifier_pass* branch here: https://github.com/lfortran/lfortran/pull/5182/files#diff-67c8fb3f4b6629c703ecb1c20661e195a435fe18141d370cd76cd2abc39021a8R602

integration_tests/CMakeLists.txt
```diff
@@ -646,7 +646,7 @@ RUN(NAME intrinsics_45 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortra
 RUN(NAME intrinsics_46 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # ichar & iachar
 RUN(NAME intrinsics_47 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # all
 RUN(NAME intrinsics_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # log_gamma
-RUN(NAME intrinsics_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER) # cmplx, sin
+RUN(NAME intrinsics_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # cmplx, sin
```
gxyd, line 649: The same is the case with this as well, the only reason to keep `NO_EXPERIMENTAL_SIMPLIFIER` is to disable `--fast` test here with `--experimental-simplifier`.

gxyd: One obvious solution seems like to have something like:
```
RUN(NAME intrinsics_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER) 
RUN(NAME intrinsics_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NOFAST)
```

but this isn't allowed because of repeated `intrinsics_49` in above.

gxyd:
We've used `NO_EXPERIMENTAL_SIMPLIFIER` for cases like when unless `--realloc-lhs` flag is used the test case fails with `--experimental-simplifier` enabled, e.g.; see: https://github.com/lfortran/lfortran/pull/5182/files#diff-67c8fb3f4b6629c703ecb1c20661e195a435fe18141d370cd76cd2abc39021a8R912 (i.e. `intrinsics_311`)

integration_tests/CMakeLists.txt
```diff
@@ -757,7 +757,9 @@ RUN(NAME intrinsics_155 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_156 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # blt
 RUN(NAME intrinsics_157 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # ble
 RUN(NAME intrinsics_158 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # bge
-RUN(NAME intrinsics_159 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # matmul
+# The test `intrinsics_159` below has `NO_EXPERIMENTAL_SIMPLIFIER` label because it fails when the `--fast` flag is used
```
gxyd, line 760: Yeah, I think this kind of a comment would help for each of the test case, for why `NO_EXPERIMENTAL_SIMPLIFIER` flag is being used.

integration_tests/CMakeLists.txt
kmr-srbh: @gxyd I observed only 1 failure of all the tests (both with and without `--fast`). It was an issue with `intrinsics_159` failing when `--fast` was used with `--experimental-simplifier`.

integration_tests/CMakeLists.txt
gxyd: For most of integration tests which currently use `NO_EXPERIMENTAL_SIMPLIFER` you would be able to find some change in *simplifier_pass* PR, which will give you an idea for why `NO_EXPERIMENTAL_SIMPLIFIER` label is needed here.

integration_tests/CMakeLists.txt
gxyd: Oh, if that's the case, then those errors are specific to macOS, you can ofcourse uncomment it for whatever passes, e.g. if `intrinsics_49.f90` passes for you with `--fast`, it doesn't pass for me neither on *main* nor with `--experimental-simplifier`:

```console
~/OpenSource/lfortran simplifier_pass ?12 ❯ lfortran integration_tests/intrinsics_49.f90 --fast                                                                                                                                                                    lf 10:20:48 PM
Assertion failed: (PointeeType == cast<PointerType>(Ptr->getType()->getScalarType())->getElementType()), function Create, file Instructions.h, line 941.
[1]    36992 abort      /Users/gxyd/OpenSource/lfortran/src/bin/lfortran  --fast
```

these the above issues are specific to macOS not on Linux then.

gxyd: Please go ahead and uncomment, whatever intrinsics or other integration tests that work for you without needing `NO_EXPERIMENTAL_SIMPLIFIER`, but please make sure to remove `NOFAST` (or the corresponding backend) from `simplifier_pass` branch as well.

E.g; if you remove `NO_EXPERIMENTAL_SIMPLIFIER` from *main` for `intrinsics_49`, please remove `NOFAST` from here as well: https://github.com/lfortran/lfortran/pull/5182/files#diff-67c8fb3f4b6629c703ecb1c20661e195a435fe18141d370cd76cd2abc39021a8R650

kmr-srbh: Aren't these differences between Linux an macOS output coming up too often now? I think we should investigate the cause. :detective:

gxyd: I've a feeling this particular test might not be difficult to fix, as it's in ASR->LLVM and not a segfault, which makes it easy to understand.

gxyd: Btw, this has nothing do with the changes in *simplifier_pass* branch, because `intrinsics_49.f90` with `--fast` fails even on *main* for me as well (on macOS m1), with the same error as I printed out above

integration_tests/CMakeLists.txt
certik: I see. That's not a good sign, there is something wrong. But for now I think we are fine, we'll investigate this later.

certik:
Looks like this might work, so we'll merge it.

PR #5206, opened 2024-10-29
ci: attempt to reduce number of jobs by lumping tests

### LINKED ISSUE #5203: CI: merge the following CI tests into one
image

Pranavchiku:
Closes #5203.


certik:
That;s good. Maybe move the upload tarball first, so that when we release, it uploads it quickly.

.github/workflows/CI.yml
```diff
@@ -614,11 +561,11 @@ jobs:
             $(pwd)/../src/bin/lfortran ./examples/example_lmder1.f90 --legacy-array-sections
 
             # Test with experimental simplifier
-            $(pwd)/../src/bin/lfortran ./src/minpack.f90 -c --legacy-array-sections --experimental-simplifier
-            $(pwd)/../src/bin/lfortran ./examples/example_hybrd.f90 --legacy-array-sections --experimental-simplifier
-            $(pwd)/../src/bin/lfortran ./examples/example_hybrd1.f90 --legacy-array-sections --experimental-simplifier
-            $(pwd)/../src/bin/lfortran ./examples/example_lmdif1.f90 --legacy-array-sections --experimental-simplifier
-            $(pwd)/../src/bin/lfortran ./examples/example_lmder1.f90 --legacy-array-sections --experimental-simplifier
+            $(pwd)/../src/bin/lfortran ./src/minpack.f90 -c --legacy-array-sections
```
Pranavchiku, line 564: is this expected?

Pranavchiku:
We can do the following things:
- [ ] Merge `CI / Build LFortran to WASM and Upload (pull_request)` with others
- [ ] Try to merge `CI / Documentation (pull_request)` with others

.github/workflows/CI.yml
gxyd: We should be keeping the `--experimental-simplifier` flag, right?

Pranavchiku: I am not sure why we have this, because of rebase I think.

.github/workflows/CI.yml
```diff
@@ -951,21 +898,6 @@ jobs:
             examples/example_primes
             ctest
 
-            cd ..
```
gxyd, line 954: all of this is intentional to make sure that third party packages are well tested with or without enabling *experimental simplifier*, so I think we would want to restore these.

.github/workflows/CI.yml
Pranavchiku: I know why it is there, let me fix it.

Pranavchiku:
I cancelled workflow, as I am experimenting conditional workflow.

.github/workflows/CI.yml
```diff
@@ -524,9 +524,74 @@ jobs:
...
+              if [[ "$file" == src/libasr/codegen/* ]] && ! echo "$IGNORED_CODEGEN_FILES" | grep -q -F "$file"; then
+                echo "Unignored codegen file changed ($file), setting run_tests=true"
+                echo "run_tests=true" >> $GITHUB_ENV
+                exit 0
+              fi
+            done
+
+            # If no condition is met, set run_tests to false
+            echo "No relevant changes detected, setting run_tests=false"
+            echo "run_tests=false" >> $GITHUB_ENV
```
Pranavchiku, line 590: so this checks if changes in file can affect LLVM backend or not, if yes, it executes else skips. For example any change only in ASR to Fortran backend shall not go through this check and we can confidently skip it with a review.

certik:
Can we just consolidate the tests but otherwise not do any changes in this PR, then merge it?

Then in a next PR we can try to skip some tests, if it is feasible. I am worried about having complicated logic in the CI that is itself not tested. Any bug there will skip tests and we will never know. Either way, let's separate this enhancement to a separate PR?

Pranavchiku:
Yes, I agree.

Pranavchiku:
@certik I think you'll have to merge this.

certik:
I need to carefully review this, it will take me some time. We can't make a mistake here.

.github/workflows/CI.yml
```diff
@@ -1375,120 +1362,122 @@ jobs:
...
           create-args: >-
             llvm-openmp
+            nodejs=18.13.0
 
       - uses: hendrikmuhs/ccache-action@main
         with:
           variant: sccache
           key: ${{ github.job }}-${{ matrix.os }}
 
+      - name: Upload Tarball
```
gxyd, line 1372: I'm curious, now these jobs are clubbed together, if one of these like `Upload Tarball` fails (let's say, the upload doesn't work for some reason), the entire CI job will stop to continue (just an observation).

So, do you think the ordering of the *Test integration tests with GFortran*, *Upload Tarball*, *Upload Tarball to release* etc. should be thought about?

gxyd:
> That;s good. Maybe move the upload tarball first, so that when we release, it uploads it quickly.

Ah, I missed this comment.

gxyd:
While this PR has its benefits, I have some concerns about combining CI jobs into a single large job:

* **Increased Re-run Time**: If one part of the combined job (e.g., Test C/C++ backend) fails, say due to a timeout, we now have to re-run the entire job, which takes around 19 minutes, instead of just re-running Test C/C++ backend, which currently takes about 4 minutes on its own.
* **Longer Debugging Process**: When adding new integration tests (e.g., RUN(NAME program_cmake_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)), if the CI fails initially with gfortran and I push a fix, I may discover additional issues in the cpp backend only in the next run. With combined jobs, this makes the debugging process longer. If the jobs remain separate, it’s much easier to see which ones fail individually and address issues faster.

These CI changes are definitely good, when all of the tests work and nothing goes wrong, but in case something fails the debug process could be a little longer.

Pranavchiku:
Timeout usually happens during runner picking up job, waiting for server, etc i.e. job beginning, so if it happens, build Linux will get affected and not others. So I don’t think it is a problem. To answer problem with longer debugging process, in general if a test works with LLVM it is supposed to pass with other backends like c/c++ as seen in most case. The one you are mentioning are corner ones, which we prefer to just remove a test for that backend and merge. To debug, in next run just comment CI for others and run it.  Also, the overall 19min is less than 26min ( for windows CI and Scipy build and test ).

.github/workflows/CI.yml
```diff
@@ -1375,120 +1362,122 @@ jobs:
...
+            ./run_tests.py
+            ./run_tests.py -vh
+            cd integration_tests
+            ./run_tests.py -m
+            ./run_tests.py -b llvm
+            ./run_tests.py -b llvm -f
+            ./run_tests.py -b llvm --experimental-simplifier
+            ./run_tests.py -b llvm -f --experimental-simplifier
+
+      - name: Test OpenMP
```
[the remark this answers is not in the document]
Pranavchiku: Not necessarily integration tests, will also have reference tests. So I’ll prefer to keep it generic. Or any-other test.

Pranavchiku:
So having individual job is best, but not scalable at the moment. One more solution is to do a build linux in one job, let other depend on it and execute in parallel via multiple jobs, this is same problem. We eventually wish the runner to be free, now instead of 8 tests we have 1. Meaning, other PRs won’t be queued for so long as they do right now.

.github/workflows/CI.yml
Pranavchiku: Resolving with reference to https://github.com/lfortran/lfortran/pull/5206#issuecomment-2453034518

.github/workflows/CI.yml
```diff
@@ -1375,120 +1362,122 @@ jobs:
...
+              -DWITH_RUNTIME_STACKTRACE=yes \
               -DCMAKE_PREFIX_PATH="$CONDA_PREFIX" \
               -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
               -DCMAKE_C_COMPILER_LAUNCHER=sccache \
               -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
 
             cmake --build . -j16 --target install
 
-      - name: Test Linux
+      - name: Test Debug Build Linux
```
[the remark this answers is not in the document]
Pranavchiku: I don’t think so, in general it is python 3.10 and fixing version is a good habit.

certik:
@Pranavchiku how much time do we safe by lumping the tests together?

Pranavchiku:
No direct time saving. This reduces load on runner so lowers queuing time. So if in main, runner had to pickup 8 tests, now it has to pickup 1 and also process 1.

certik:
I think it's still worth doing. @gxyd what do you think?

gxyd:
> I think it's still worth doing. @gxyd what do you think?

Yes, I agree, I'm happy with the changes, and this is better than the current CI. I went through the changes as well, and I didn't PR *untesting* what is already being tested in our current CI.

certik:
I rebased on top of the latest master. I think it is fine, but I haven't gone over every change carefully. It's so easy to make a mistake.

Pranavchiku:
We can take time over this PR, no worries.

certik (approved):
Just triple check everything. But it's ok from me to merge as is. 

We can iterate on it after merging if needed.

certik:
I adjusted the required CI tests to pass for a PR. So things should be ok. Let's keep an eye on these changes and ensure they actually work.

Pranavchiku:
One thing to note is, even with this merging, if one PR has checks running, few checks in other PR are still in queue. So, maybe we need to figure out other checks we can lump in. I’ll spend some time or if you have any ideas, you’re welcome

certik:
Yes, we need to lump a few more.

Pranavchiku:
I forgot to remove `Check Debug Build` from jobs, we lumped it.

PR #5011, opened 2024-10-08
feat: Collapse clause support

adit4443ya:
Towards [Issue](https://github.com/lfortran/lfortran/issues/4818)
Starting with minimal changes and two different type of test cases

PR #5210, opened 2024-10-30
chore: use int instead of unsigned int in comparison operation

adit4443ya:
Towards [bug fix](https://github.com/lfortran/lfortran/pull/5011#discussion_r1821262606)
in continuation of #5011

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -30,10 +30,10 @@ public:
     bool from_block;
     std::set<std::string> labels;
     size_t starting_n_body = 0;
-    size_t loop_nesting = 0;
-    size_t pragma_nesting_level = 0;
+    int loop_nesting = 0;
+    int pragma_nesting_level = 0;
     bool openmp_collapse = false;
-    size_t collapse_value=0;
+    int collapse_value=0;
     Vec<ASR::do_loop_head_t> do_loop_heads_for_collapse;
     Vec<ASR::stmt_t*> do_loop_bodies_for_collapse;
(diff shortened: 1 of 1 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3447,7 +3447,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 do_concurrent->m_head = do_concurrent_head.p;
                 do_concurrent->n_head = do_concurrent_head.size();
                 tmp = (ASR::asr_t*) do_concurrent;
-            } else if (openmp_collapse == true && !omp_constructs.empty() && collapse_value>loop_nesting-1-omp_constructs.size()+1) {
+            } else if (openmp_collapse == true && !omp_constructs.empty() && collapse_value > loop_nesting - 1 - static_cast<int>(omp_constructs.size()) + 1) {
```
gxyd, line 3450: Two considerations/questions:
1. maybe wrap the condition after `&&` over to the next line?
2. any reason to have non-simplified `-1` and `+1` on the rhs of the greater than comparison operator?

Pranavchiku:
Let's make a practice of having meaningful commit messages and PR names, this will avoid squash and merge in PRs and also help in preserving commit history.

certik (approved):
I think this is the right fix. Thanks!

@gxyd feedback can be addressed in another PR if needed, I think it's minor.

certik:
I missed that "small fix" commit log. As @Pranavchiku said, please write a meaningful commit log messages. Instead of "small fix", write what was fixed and why.

PR #4918, opened 2024-09-26
enh : Create physical string descriptor to be used with allocatable strings 

assem2002:
Towards #2257 
In separate PR I'll utilize `size` and `capacity` in `lfortran_alloc()` , `lfortran_strcpy()`, `lfortran_string_write()`

PR #5047, opened 2024-10-14
test: add MRE for fastGPT with '--realloc-lhs' flag

gxyd:

PR #5216, opened 2024-10-31
Fix array constructor function

assem2002:
The issue really appeared (the test failed) when I started utilizing the string descriptor built on top of this PR #4918; So I needed to do this minor fix.

### DIFF
--- a/integration_tests/assign_allocatable_array_of_struct_instances.f90
+++ b/integration_tests/assign_allocatable_array_of_struct_instances.f90
@@ -1,31 +1,34 @@
 program assign_allocatable_array_of_struct_instances
     implicit none
- 
+
     integer :: i
- 
+
     type :: string
-       character(:), allocatable :: s
+        character(:), allocatable :: s
     end type
- 
+
     type(string), allocatable :: tokens(:)
- 
+
     allocate(tokens(5))
     do i = 1, 5
-       tokens(i)%s = "a"
+        tokens(i)%s = "a"
     end do
- 
+
     tokens = merge_pair(tokens, 1)
- 
+
 contains
- 
+
     function merge_pair(intokens, idx) result(tkns)
-       ! Merge the pair `idx`
-       type(string), intent(in) :: intokens(:)
-       integer, intent(in) :: idx
-       type(string), allocatable :: tkns(:)
-       type(string) :: merged_token
-       merged_token%s = intokens(idx)%s // intokens(idx+1)%s
-       tkns = [intokens(:idx-1), merged_token, intokens(idx+2:)]
+        ! Merge the pair `idx`
+        type(string), intent(in) :: intokens(:)
+        integer, intent(in) :: idx
+        type(string), allocatable :: tkns(:)
+        type(string) :: merged_token
+        merged_token%s = intokens(idx)%s // intokens(idx+1)%s
+        ! The segmentation fault occurs at the below line, when used with
+        ! `--experimental-simplifier` flag (i.e. simplifier pass enabled) and
+        ! no `--realloc-lhs` flag is enabled for it
+        tkns = [intokens(:idx-1), merged_token, intokens(idx+2:)]
     end function
- 
+
 end program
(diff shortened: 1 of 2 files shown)


integration_tests/assign_allocatable_array_of_struct_instances.f90
```diff
@@ -25,7 +25,6 @@ function merge_pair(intokens, idx) result(tkns)
        type(string), allocatable :: tkns(:)
        type(string) :: merged_token
        merged_token%s = intokens(idx)%s // intokens(idx+1)%s
-       ! The segmentation fault occurs at the below line
        tkns = [intokens(:idx-1), merged_token, intokens(idx+2:)]
```
Pranavchiku, line 28: can we register this test now?

assem2002: It's registered here https://github.com/lfortran/lfortran/pull/5047. I don't get how we registered it and the comment says it should segfault. Do you know what this is about?

Pranavchiku: Check with `--fast`, I believe it segfaults with it.

assem2002: It's an issue related to llvm17 changes we made recently.

assem2002: maybe @gxyd can tell us what exactly the issue was.

Pranavchiku: Ahh okay, let's remove NOFAST and test on our CI. If it works, we can merge it.

certik: It still fails when `NOFAST` is removed.

certik: There is no segfault, but we get an ICE, assertion fail during compilation with LLVM >= 17.

certik (approved):

certik:
Let's merge this after #4918.

integration_tests/assign_allocatable_array_of_struct_instances.f90
gxyd: > It's an issue related to llvm17 changes we made recently. when Gaurav found it fails with `--fast` with llvm >=17 versions, He added a `NOFAST` label on the test in the `integration_tests/CMakeLists.txt`. So, the comment and `--fast` flag aren't relates as I think.

That's right: here's the original PR where the test case was added: https://github.com/lfortran/lfortran/pull/5047#discussion_r1807839553

gxyd: The comment above, i.e.:

> ! The segmentation fault occurs at the below line

Is actually only applicable to *simplifier_pass* branch (which now is applied via `--experimental-simplifier`) and the segmentation fault can be noticed when it's executed as below:
```console
$ lfortran integration_tests/assign_allocatable_array_of_struct_instances.f90 --experimental-simplifier
$ echo $?
11

$ lfortran integration_tests/assign_allocatable_array_of_struct_instances.f90 --experimental-simplifier --realloc-lhs
$ echo $?
0
```

So, the issue only occurs when `--experimental-simplifier` flag is used and **without** use of `--realloc-lhs` flag.

@assem2002 , could you please help modify comment here to say:
```f90
! The segmentation fault occurs at the below line, when used with
! `--experimental-simplifier` flag (i.e. simplifier pass enabled) and
! no `--realloc-lhs` flag is enabled for it
```

Pranavchiku:
Merging as #4918 is merged.

integration_tests/assign_allocatable_array_of_struct_instances.f90
gxyd: I hope that's alright, that I've pushed a commit addressing that comment.

Pranavchiku (approved):

integration_tests/assign_allocatable_array_of_struct_instances.f90
certik: That's fine, thanks.

Issue #5212, opened 2024-10-30
tokenizer error with format 'tr1' 


gxyd:
## Minimal Reproducible Example

for the below fortran program in fixed-form, with a `format` statement

```f90
      subroutine f ()
      implicit none

  1   format (tr1,a)

      end
```

on execution of the above program with lfortran as below:
```console
$ lfortran test.f --fixed-form --cpp
tokenizer error: Token 't' is not recognized in `format` statement
 --> test.F:1:8
  |
1 |       subroutine f ()
  |        ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

## Other information

We need the fix to be able to move forward with compilation of [POT3D](https://github.com/lfortran/lfortran/issues/2862). Please see the below comment https://github.com/lfortran/lfortran/issues/5212#issuecomment-2446428446, that the issue isn't restricted to fixed-form programs

gxyd:
Seems like this issue exists even free-form as well:
```f90
      subroutine f ()
      implicit none

  1   format (tr1,a)

      end

      program main
        implicit none
      end program main

```

on it's execution with lfortran as below:
```console
> lfortran test.F
tokenizer error: Token 't' is not recognized in `format` statement
 --> test.F:1:2
  |
1 |       subroutine f ()
  |  ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

certik:
We might not support the "t" token in the format. I am not sure what it does, so we need to look it up, and then improve the tokenizer.

PR #5219, opened 2024-11-01
bug: fix regex for 'TR'/'TL' format specification in tokenizer

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5212. This PR only fixes the problem in tokenization of 'TL'/'TR' format, and I think there are some issues in runtime with this format as well, I'll open separate issues for them there and handle them.


src/lfortran/parser/tokenizer.re
```diff
@@ -777,7 +777,7 @@ void lex_format(unsigned char *&cur, Location &loc,
                 ;
 
             position_edit_desc
-                = 'T' whitespace? int ('L' | 'R')? whitespace? int
+                = 'T' whitespace? ('L' | 'R')? whitespace? int
```
gxyd, line 780: Below is a screenshot from *Fortran 2023 Interpretation document J3/24-007*, describing what the different formats look like:

Screenshot 2024-11-01 at 10 08 33 AM

certik (approved):
So there was a bug that we had "int" in there by a mistake?

gxyd:
> So there was a bug that we had "int" in there by a mistake?

Yes, that's it.

There is a separate issue with runtime for `TR/TL` formats as well, where it raises an unexpected error, I'll raises separate issue after this for it.

Issue #5224, opened 2024-11-01
Implement 'TR' format descriptor


gxyd:
## Minimal Reproducible Example

for the below fortran program in free-form:
```f90
PROGRAM main
    IMPLICIT NONE

    ! printing with just TR5 - should skip 5 spaces
    ! before printing "Hello, God!"
    PRINT '(TR5, A)', "Hello, God!"
END PROGRAM main
```

on it's execution with LFortran as:
```console
$ lfortran --version
LFortran version: 0.41.0-173-g8f7a93466-dirty
Platform: macOS ARM
LLVM: 11.0.1
Default target: arm64-apple-darwin23.6.0

$ lfortran test.f90
Error: Positive width required with 'T' descriptor in format string

$ echo $?
1
```

## Output with GFortran

```console
> gfortran test.f90 -o a.out && ./a.out
     Hello, God!
```

certik:
I think that's why it had the "int" in the parser in it. Make sure "T <int>" is not allowed.

PR #5225, opened 2024-11-01
enh: implement "TL/TR" format specifiers

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5224.

Since this would anyway be eventually needed for compilation of POT3D third party package, I picked it up.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -717,10 +717,17 @@ char** parse_fortran_format(char* format, int64_t *count, int64_t *item_start) {
                 *item_start = format_values_count;
                 break;
             case 't' :
+                // handle 'T', 'TL' & 'TR' editing see section 13.8.1.2 in 24-007.pdf
                 start = index++;
-                // raise error when "T" is specified itself or with non-positive width
+                if (tolower(format[index]) == 'l' || tolower(format[index]) == 'r') {
+                     index++;  // move past 'L' or 'R'
+                }
+                // raise error when "T/TL/TR" is specified itself or with
+                // non-positive width
                 if (!isdigit(format[index])) {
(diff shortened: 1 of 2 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1046,25 +1053,51 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(int count, const char* form
...
+                        // Truncate the string to the length specified by Tn
+                        // if the current position exceeds it
+                        if (tab_position < current_length) {
+                            // Truncate the string at the position specified by Tn
+                            result[tab_position] = '\0';
+                        }
                     }
                 }
             } else {
-                if(count <= 0) break;
```
gxyd, line 1066: Only formatting has been improved here

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1046,25 +1053,51 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(int count, const char* form
...
+                    int tab_left_pos = atoi(value + 2);
+                    int current_length = strlen(result);
+                    if (tab_left_pos > current_length) {
+                        result[0] = '\0';
+                    } else {
+                        result[current_length - tab_left_pos] = '\0';
+                    }
+                } else if (tolower(value[1]) == 'r') {
+                    // handle "TR" format specifier
+                    int tab_right_pos = atoi(value + 2);
```
gxyd, line 1067: There definitely is some common code between 'TR' and 'T', but I wouldn't want to refactor it, as it would end up leading to more `if .. else..`, and current code is readable considering I've used more case specific variable names.

Let me know if there is any refactoring expected still.

certik (approved):
I think it looks great, thanks for implementing it!

Issue #5220, opened 2024-11-01
fixed form program with do concurrent statement doesn't end


gxyd:
## Minimal Reproducible Example

for the below fixed form Fortran program:

```f90
      subroutine f ()
            implicit none
            integer :: i
            do concurrent (i=1:2)
            enddo
      end subroutine
```

on executing the above fortran program as below:
```console
$ lfortran --version
LFortran version: 0.41.0-172-gedb225517-dirty
Platform: macOS ARM
LLVM: 11.0.1
Default target: arm64-apple-darwin23.6.0

$ lfortran test.f --fixed-form --show-ast
# the program doesnt print anything nor does it end
$ echo $?
130
```

PR #5229, opened 2024-11-02
chore: tokenize 'do concurrent' in fixed-form program

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5220

### DIFF
--- a/src/lfortran/parser/fixedform_tokenizer.cpp
+++ b/src/lfortran/parser/fixedform_tokenizer.cpp
@@ -956,6 +956,11 @@ struct FixedFormRecursiveDescent {
             return true;
         }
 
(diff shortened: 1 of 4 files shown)


gxyd:
Ideally, something like `fixed_free_equal_ast` or `fixed_free_equal_tokens` (currently there is `fixed_free_equal_asr`, which I'm not sure if it does anything at all) would be nice, as that will allow us to use existing programs to ensure that free-form and fixed-form behave the same.

certik (approved):

certik:
We could, but I think the current AST tests are enough. We could in the future take our free-form tests, automatically generate a fixed-form code, reparse it, and compare the AST. Let's do that after beta.

Issue #5036, opened 2024-10-12
Towards error resilient frontend


Pranavchiku:
Picked from PR: #1098:

This example shows how to recover from errors, it creates an AST:
```console
tests/errors(error2) $ lfortran assignment_01.f90 --show-ast --indent
ERROR 3
ERROR 2
(TranslationUnit 
    [(Module 
        assignment_01 
        (TriviaNode 
            [(EndOfLine)
            (EndOfLine)] 
            []
        ) 
        [] 
        [] 
        [] 
        [(Subroutine 
            f 
            [(a)
            (b)] 
            [] 
            () 
            (TriviaNode 
                [] 
                [(EndOfLine)
                (EndOfLine)]
            ) 
            [] 
            [] 
            [] 
            [(Declaration 
                (AttrType 
                    TypeInteger 
                    [] 
                    () 
                    None
                ) 
                [(AttrIntent 
                    InOut
                )] 
                [(a 
                [] 
                [] 
                () 
                None 
                ())
                (b 
                [] 
                [] 
                () 
                None 
                ())] 
                (TriviaNode 
                    [] 
                    [(EndOfLine)
                    (EndOfLine)]
                )
            )] 
            [(= 
                0 
                b 
                b 
                (TriviaNode 
                    [] 
                    [(EndOfLine)
                    (EndOfLine)]
                )
            )] 
            []
        )]
    )]
)
```

TODO:
* [x] Put the syntax error into the diagnostics, but continue (https://github.com/lfortran/lfortran/pull/4774)
* [ ] Add a compiler option that would keep these errors but it would not exit after AST, but to ASR as well, just ignoring/recovering from the error
* [ ] Create InvalidStatement, or InvalidExpr AST nodes, and insert them into the tree with location information and any other info that we might have
* [ ] Benchmark speed for valid code to ensure there is no regression (I don't expect a significant slow down, but let's check)

certik:
@Pranavchiku as we discussed:
* [ ] use the SemanticAbort exception, but catch at at the statement level. Keep constructing ASR. 

I think we should for now not add error nodes into the AST/ASR, that can be done later.

Pranavchiku:
Yes, I just copied it over so that I can brainstorm and document design ideas here whenever I do.

PR #5230, opened 2024-11-02
make array_02 error resilient

AnshMehta1:
Towards :- #5036 

array_02
```
program array_02
implicit none

print *, [[[], [[]]], [[]], []]
print * , "made error resilient"

end program
```

LFortran outcome 
```
(lf) ansh_mehta@DESKTOP-VJGRLON:/mnt/d/lfortran2/lfortran$ ./src/bin/lfortran try.f90 --continue-compilation
semantic error: Empty array constructor is not allowed
 --> try.f90:4:12
  |
4 | print *, [[[], [[]]], [[]], []]
  |            ^^


made error resilient
```


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -118,7 +118,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             // Visit the statement
             LCOMPILERS_ASSERT(current_body != nullptr)
             tmp = nullptr;
-            this->visit_stmt(*m_body[i]);
+            try {
+                this->visit_stmt(*m_body[i]);
+            } catch (const SemanticAbort &a) {
+                if (!compiler_options.continue_compilation) throw a;
+            }
```
certik, line 125: Can it happen that `tmp` is not null but error was raised? Would be safer to set `tmp=nullptr` here to be sure.

certik: Same with `tmp_vec`.

tests/reference/run-array_02-becdf83.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Empty array constructor is not allowed
+ --> tests/errors/array_02.f90:4:12
+  |
+4 | print *, [[[], [[]]], [[]], []]
+  |            ^^
```
Pranavchiku, line 5: I think something is off with this reference, it should not stop here and atleast produce some output.

src/lfortran/semantics/ast_body_visitor.cpp
Pranavchiku: > Can it happen that tmp is not null but error was raised?

I don't think so, `tmp` is usually set at very end and if `SemanticAbort` is thrown `tmp` would still be nullptr.

AnshMehta1: Ok , will work on it and made the required changes.

AnshMehta1: > > Can it happen that tmp is not null but error was raised?
> 
> I don't think so, `tmp` is usually set at very end and if `SemanticAbort` is thrown `tmp` would still be nullptr. But it'll be good to set `tmp` and `tmp_vec` as nullptr if `--continue-compilation` is on.

Made tmp as null and tmp_vec as empty . This changes are made within catch block so as to not to hinder the behaviour of code flow when their is no Semantic abort.

certik: Thanks, looks good.

tests/tests.toml
```diff
@@ -3549,6 +3549,7 @@ asr = true
 [[test]]
 filename = "errors/array_02.f90"
 asr = true
+continue_compilation = true
```
certik, line 3552: Instead of modifying existing test, add a new test and ensure multiple errors are reported (make two errors on consecutive lines).

Pranavchiku: Let's make a convention then, `array_02_cc.f90`.

tests/reference/run-array_02_cc-634bad3.json
```diff
@@ -0,0 +1,13 @@
...
+    "cmd": "lfortran --continue-compilation --no-color {infile}",
+    "infile": "tests/errors/array_02_cc.f90",
+    "infile_hash": "a34069990d9e3e23ea76584253a86585a9892ddc087ec9b4b011a78b",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": "run-array_02_cc-634bad3.stdout",
+    "stdout_hash": "a779ee4f7341ddce58a41cee3d249ecf1b37216d7d4f16eb4e0f528e",
+    "stderr": "run-array_02_cc-634bad3.stderr",
+    "stderr_hash": "e5dac3833910ebb373f5c18d40cbb77da46d1932776c3970b6a1c2f5",
+    "returncode": 0
```
certik, line 12: I just realized that this is a mistake --- it needs to return non-zero `returncode`m since there were errors.

https://github.com/lfortran/lfortran/issues/5248

certik: I answered here: https://github.com/lfortran/lfortran/issues/5248#issuecomment-2455277639.

Issue #5231, opened 2024-11-03
CI: Test fastGPT ( macos-latest ) job under 'Test LLVM' doesn't run


gxyd:
## Description

Let's look at the `Test LLVM` CI job here: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L878-L879, where it explicitly mentions:
```yml
runs-on: ubuntu-latest
```

and under that we've `Test fastGPT ( macos-latest)` here: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L1163, which will probably never run because of the main job already requiring the os to be `ubuntu-latest`.

I'm not really sure if that's intentional or not, if it is intentional, then maybe we can have a comment describing that it's there just for the sake of repeating the code repeated from https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L496 (i.e. `Check Third Party Code Compilation - OS (${{ matrix.os }}), LLVM (${{ matrix.llvm-version }})`) or the other is we actually test it on macOS as well.

gxyd:
I went through a few of the PR's where `Test LLVM` CI job is run, e.g.; https://github.com/lfortran/lfortran/actions/runs/11645376083/job/32428340408?pr=5230

there's nothing which runs under:
* `Test fastGPT (ubuntu-latest)` 
* `Test fastGPT (macos-latest)`
* `Test fpm`

under none of them there is any console output (see attached screenshot as well), which I think indicates that they aren't actually being run.

![Screenshot 2024-11-03 at 10 27 14 AM](https://github.com/user-attachments/assets/31b9b241-33a2-4719-8131-e231bf224a6f)

Pranavchiku:
There is separate mac-os job for FastGPT

gxyd:
> There is separate mac-os job for FastGPT

I'm sorry, I don't understand. Are you saying that we do run the macOS job for fastGPT at the CI correctly? The links in the issue description suggest otherwise to me.

Pranavchiku:
Yes, refer: https://github.com/lfortran/lfortran/pull/5232#discussion_r1826917386

Pranavchiku:
> The links in the issue description suggest otherwise to me.

We made a choice not to run third party test on all llvm version but only 11 and 19, both on mac and ubuntu.

Pranavchiku:
[...]

PR #5232, opened 2024-11-03
CI: fix bug with fastGPT ubuntu & fpm in Test LLVM CI job

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5231

some of third party packages in it need a specific version i.e. either macOS or ubuntu, this PR makes sure that the OS is used as a variable, where the different third party can decide whether something needs to be run with 'macOS' or 'ubuntu' or both

before this commit, 'Test fpm', 'Test fastGPT ( ubuntu-latest )' or 'Test fastGPT ( macos-latest )' weren't being run under 'Test LLVM' CI job as 'matrix.os' was always empty

Below are the third party packages, which are now tested twice for each OS (ubuntu and macOS), and for each LLVM version (earlier they are were tested on only ubuntu):
* Test Legacy Minpack (SciPy)
* Test Modern Minpack (Fortran-Lang)
* Test Modern Minpack (check results)
* Test dftatom
* Test stdlib
* Test SNAP

unless needed, we can skip their testing with macOS and only do that on ubuntu, as that will save sometime on the CI. @certik thoughts?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -875,8 +875,8 @@ jobs:
             ./gsnap ../qasnap/sample/inp out
 
   test_llvm:
-    name: Test LLVM ${{ matrix.llvm-version }}
-    runs-on: ubuntu-latest
+    name: Test LLVM ${{ matrix.llvm-version }} - OS (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
@@ -887,6 +887,7 @@ jobs:
         # exactly the same manner, to ensure that users can just change the
         # `llvmdev` version in their conda environment and everything will just
         # work.
+        os: ["macos-latest", "ubuntu-latest"]
         llvm-version: ["10", "11", "12", "13", "14", "15", "16", "17", "18", "19"]
     steps:
       - uses: actions/checkout@v4
@@ -907,6 +908,7 @@ jobs:
 
       - name: Build Linux
         shell: bash -e -l {0}
+        if: contains(matrix.os, 'ubuntu')
         run: |
             ./build0.sh
             cmake . -GNinja \
@@ -922,9 +924,17 @@ jobs:
 
             cmake --build . -j16 --target install
 
+      - name: Build MacOS
+        shell: bash -e -l {0}
+        if: contains(matrix.os, 'macos')
+        run: |
+            ./build0.sh
+            ./build1.sh
+
       # LLVM 10-19 all work in exactly the same way, so the test is identical
       - name: Test Linux LLVM 10-19
         shell: bash -e -l {0}
+        if: contains(matrix.os, 'ubuntu')
         run: |
             ctest --output-on-failure
             cd integration_tests
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -875,8 +875,8 @@ jobs:
             ./gsnap ../qasnap/sample/inp out
 
   test_llvm:
-    name: Test LLVM ${{ matrix.llvm-version }} - OS (${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
+    name: Test LLVM ${{ matrix.llvm-version }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -887,7 +887,6 @@ jobs:
         # exactly the same manner, to ensure that users can just change the
         # `llvmdev` version in their conda environment and everything will just
         # work.
-        os: ["macos-latest", "ubuntu-latest"]
         llvm-version: ["10", "11", "12", "13", "14", "15", "16", "17", "18", "19"]
     steps:
       - uses: actions/checkout@v4
@@ -908,7 +907,6 @@ jobs:
 
       - name: Build Linux
         shell: bash -e -l {0}
-        if: contains(matrix.os, 'ubuntu')
         run: |
             ./build0.sh
             cmake . -GNinja \
@@ -924,17 +922,9 @@ jobs:
 
             cmake --build . -j16 --target install
 
-      - name: Build MacOS
-        shell: bash -e -l {0}
-        if: contains(matrix.os, 'macos')
-        run: |
-            ./build0.sh
-            ./build1.sh
-
       # LLVM 10-19 all work in exactly the same way, so the test is identical
       - name: Test Linux LLVM 10-19
         shell: bash -e -l {0}
-        if: contains(matrix.os, 'ubuntu')
         run: |
             ctest --output-on-failure
             cd integration_tests
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -880,8 +880,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["macos-latest", "ubuntu-latest"]
-        python-version: ["3.10"]
         # Note: The LLVM 11 test is redundant, since it is already tested by
         # most other tests (since LLVM 11 is our default version), but we have
         # it here as well for consistency, and that way other tests can freely
@@ -889,6 +887,7 @@ jobs:
         # exactly the same manner, to ensure that users can just change the
         # `llvmdev` version in their conda environment and everything will just
         # work.
+        os: ["macos-latest", "ubuntu-latest"]
         llvm-version: ["10", "11", "12", "13", "14", "15", "16", "17", "18", "19"]
     steps:
       - uses: actions/checkout@v4
@@ -900,12 +899,7 @@ jobs:
           micromamba-version: '1.5.10-0'
           environment-file: ci/environment_linux_llvm.yml
           create-args: >-
-            python=${{ matrix.python-version }}
             llvmdev=${{ matrix.llvm-version }}
-            bison=3.4
-            openblas=0.3.21
-            cmake=3.26.4
-            make=4.3
 
       - uses: hendrikmuhs/ccache-action@main
         with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -880,6 +880,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        os: ["macos-latest", "ubuntu-latest"]
+        python-version: ["3.10"]
         # Note: The LLVM 11 test is redundant, since it is already tested by
         # most other tests (since LLVM 11 is our default version), but we have
         # it here as well for consistency, and that way other tests can freely
@@ -887,7 +889,6 @@ jobs:
         # exactly the same manner, to ensure that users can just change the
         # `llvmdev` version in their conda environment and everything will just
         # work.
-        os: ["macos-latest", "ubuntu-latest"]
         llvm-version: ["10", "11", "12", "13", "14", "15", "16", "17", "18", "19"]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


gxyd:
Even though this is a small change, it makes me a little nervous about making changes in CI, considering that there is another big change coming in from another CI change here: https://github.com/lfortran/lfortran/pull/5206.

I'm really in no hurry to get this change merged, cause I believe it's better to make one change to the CI in *main* and wait (for a few days maybe) to see if things look good and everything seems fine, and after that we go ahead with the next change.

But ofcourse, I'm happy to receive feedback on this PR.

Pranavchiku:
We test FastGPT with macos

.github/workflows/CI.yml
```diff
@@ -875,11 +875,13 @@ jobs:
             ./gsnap ../qasnap/sample/inp out
 
   test_llvm:
-    name: Test LLVM ${{ matrix.llvm-version }}
-    runs-on: ubuntu-latest
+    name: Test LLVM ${{ matrix.llvm-version }} - OS (${{ matrix.os }})
```
Pranavchiku, line 878: We don't need it here, we already test all third party tests both on macos and ubuntu latest with LLVM 11 and LLVM 19:

Refer: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L496-L503

We don't want to test on multiple LLVM versions, 11 and 19 suffice.

Pranavchiku (commented):
> before this commit, 'Test fpm', 'Test fastGPT ( ubuntu-latest )' or 'Test fastGPT ( macos-latest )' weren't being run under 'Test LLVM' CI job as 'matrix.os' was always empty

fastGPT ubuntu: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L653
fastGPT macos: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L778
fpm ubuntu: https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L809

Reason why fpm is on ubuntu is because of different build.sh requirement, we can set it up though.

.github/workflows/CI.yml
gxyd: So @Pranavchiku , what are you recommending is we go with the first option listed here in the issue (i.e.):

> I'm not really sure if that's intentional or not, if it is intentional, then maybe we can have a comment describing that it's there just for the sake of repeating the code repeated from https://github.com/lfortran/lfortran/blob/main/.github/workflows/CI.yml#L496

Pranavchiku:
> Below are the third party packages, which are now tested twice for each OS (ubuntu and macOS), and for each LLVM version (earlier they are were tested on only ubuntu):

> Test Legacy Minpack (SciPy)
Test Modern Minpack (Fortran-Lang)
Test Modern Minpack (check results)
Test dftatom
Test stdlib
Test SNAP

All these are already tested with mac-os, refer: https://github.com/lfortran/lfortran/actions/runs/11643270291/job/32423821523

.github/workflows/CI.yml
Pranavchiku: In short, what this PR is aiming to do is already being done at CI. It was intentional to only test with LLVM 19 and 11 to avoid lots of CI runs and hence macos-latest is not put in `Test LLVM`.

gxyd: Why do we've all of these third party tests under *Test LLVM* if we already test all of the third party tests with LLVM 19 and 11 under `third_party_code_compile`?

gxyd:
Earlier, we used test all the third party packages (except `fpm`, `fastGPT with ubuntu` and `fastGPT with macOS` weren't run at all) with all LLVM versions ie. 10 through 19.

Now we test all the third party packages with either ubuntu (all except `fastGPT with macOS`) or macOS (but not both) with all the LLVM versions

.github/workflows/CI.yml
Pranavchiku: I hope https://github.com/lfortran/lfortran/issues/5231#issuecomment-2453326257 answers.

Pranavchiku:
> (except fpm, fastGPT with ubuntu and fastGPT with macOS weren't run at all) with all LLVM versions ie. 10 through 19.

Please refer: https://github.com/lfortran/lfortran/actions/runs/11643270291/job/32423821523, we do run it on macos.

Pranavchiku (commented):
> 5 successful, 19 in progress, 22 queued, and 10 expected checks

I am -1 to add these many checks on CI.

gxyd:
> I am -1 to add these many checks on CI.

That's a valid concern. Let's only do that with ubuntu then, I'll push a change.

Pranavchiku:
Cancelled runner with reference to above comment.

gxyd:
> I am not sure what do you mean by https://github.com/lfortran/lfortran/pull/5232#issuecomment-2453326044, if you look at https://github.com/lfortran/lfortran/actions/runs/11643270291/job/32423821523, we already test all third party codes with macos on CI

I don't agree with this explanation, because with that logic, we already test all the third party packages with ubuntu as well, so no need to have any third party packages under *Test LLVM* then.

The only reason to not have *macOS* here for me is, it increases the number of CI jobs to run.

Pranavchiku:
Precisely summarised https://github.com/lfortran/lfortran/issues/5231#issuecomment-2453326257!

Pranavchiku:
The whole argument was on "Do we need to test all LLVM versions for macos-latest?", no, because LLVM 11 & 19 suffice considering constraints with runner. For more clarification, please refer: https://github.com/lfortran/lfortran/pull/5006#issuecomment-2396980711

gxyd:
> The whole argument was on "Do we need to test all LLVM versions for macos-latest?",

That wasn't initially the argument, but it ended up being a side effect of my changes.

gxyd:
Let me make some changes, which will have a positive change.

Pranavchiku:
> That wasn't initially the argument, but it ended up being a side effect of my changes.

Ahh, I am sorry for not clearly stating about it at the beginning. Let's take this PR to a state we both approve on and take a final review from Ondrej :)

.github/workflows/CI.yml
```diff
@@ -1193,7 +1195,6 @@ jobs:
 
       - name: Test fpm
         shell: bash -e -x -l {0}
-        if: contains(matrix.os, 'ubuntu')
```
gxyd, line 1196: this job was earlier never run at all all, see screenshot in: https://github.com/lfortran/lfortran/issues/5231#issuecomment-2453297833

.github/workflows/CI.yml
```diff
@@ -1160,6 +1159,9 @@ jobs:
 
             rm -rf fastGPT/
 
+      # The below job isn't every run because this run on 'macOS' and it
+      # isn't worth the extra runners on CI to for testing fastGPT on macOS
+      # with all the LLVM versions
       - name: Test fastGPT ( macos-latest )
         shell: bash -e -x -l {0}
         if: contains(matrix.os, 'macos')
```
Pranavchiku, line 1167: we can also remove it from here

.github/workflows/CI.yml
```diff
@@ -1037,7 +1037,6 @@ jobs:
 
       - name: Test fastGPT ( ubuntu-latest )
```
gxyd, line 1038: this job was earlier never run at all, see: https://github.com/lfortran/lfortran/issues/5231#issuecomment-2453297833

.github/workflows/CI.yml
```diff
@@ -1037,7 +1037,6 @@ jobs:
...
         shell: bash -e -x -l {0}
-        if: contains(matrix.os, 'ubuntu')
```
Pranavchiku, line 1040: let's keep this, because this is specific to `ubuntu`, if anyone tried to reproduce it locally on mac, it will throw segfault because of different binary. So to avoid guiding astray, it is good to keep it.

.github/workflows/CI.yml
Pranavchiku, line 1196: same comment for this

.github/workflows/CI.yml
gxyd: if we do, then this job will be tried to be run with `ubuntu`, we don't want that.

.github/workflows/CI.yml
gxyd: No, we can't because the value of `matrix.os` will be an empty string, that's a bug which we currently have.

Pranavchiku: Yup, let's add a matrix for os with `ubuntu` in it or add a comment specifying it.

gxyd: That makes sense.

Pranavchiku: The reason I am emphasising on it is because I have struggled / stuck on compilation because of this minor details, see https://lfortran.zulipchat.com/#narrow/channel/197339-General/topic/fpm.20compilation/near/424530202.

gxyd: I agree on this.

.github/workflows/CI.yml
Pranavchiku: I mean remove entire code block for `Test fastGPT ( macos-latest )`

.github/workflows/CI.yml
```diff
@@ -898,7 +901,13 @@ jobs:
           micromamba-version: '1.5.10-0'
           environment-file: ci/environment_linux_llvm.yml
           create-args: >-
+            python=${{ matrix.python-version }}
             llvmdev=${{ matrix.llvm-version }}
+            bison=3.4
+            openblas=0.3.21
+            llvm-openmp=14.0.4
+            cmake=3.26.4
+            make=4.3
```
gxyd, line 910: these dependencies are need for `fastGPT (with ubuntu)`

.github/workflows/CI.yml
gxyd: I've reverted this change.

.github/workflows/CI.yml
gxyd: I've reverted this change

Pranavchiku (approved):
Awesome, this fixes #5231 without any new load on runner and aligning with https://github.com/lfortran/lfortran/pull/5006#issuecomment-2396980711. It was a good discussion, thanks for it and making this PR reach at a state where we both agree :)

gxyd:
> I saw this PR and it introduced macos to third party test, so that was my argument since beginning

Ah yes, I didn't realize the impact that it would've and in the later commits I saw the CI and realized that it wasn't the way to go forward with it.

Now, I'm also convinced that just having a comment above it, is actually the way to go, among the two options which I listed here: https://github.com/lfortran/lfortran/issues/5231#issue-2630968337

Sorry, if it was a little confusing here @Pranavchiku

Pranavchiku:
No worries, happy to discuss and come up with a solution :)

gxyd:
As this is a change in CI, we'll ofcourse wait for @certik 's final comments.

Pranavchiku:
Also for future optimisations it is worth to note the time it takes to run these longer CI tests like SciPy, Test LLVM, third party code compilation, etc. 

one suggestion @gxyd we can avoid `CI / Check Third Party Code Compilation - OS (ubuntu-latest), LLVM (11) (pull_request)`and `CI / Check Third Party Code Compilation - OS (ubuntu-latest), LLVM (19) (pull_request)` in subsequent PRs, as we already check it in `Test LLVM`

.github/workflows/CI.yml
certik: Yes, I think you can remove this whole block. we still test on macOS separately in the 3rd party tests.

.github/workflows/CI.yml
certik: Why is this needed now, and it was not needed before?

I don't see any changes to the CI that would suddenly require this.

certik:
As it is now, I think it's good. The main bug it fixes is that it introduces `python-version` and `os`, especially the `os` was used in the tests, so they were run twice.

I am worried about https://github.com/lfortran/lfortran/pull/5232/files#r1827004029, as this was not needed before. 

I am also worried about `llvm-openmp=14.0.4` accidentally installing LLVM 14 everywhere --- must carefully check.

certik:
I see, fastGPT wasn't tested at all before.

certik (approved):
I think this is good. I checked and the proper LLVM version is being used.

Issue #2257, opened 2023-08-26
Redesigning strings


certik:
There are two main issues with strings:

* Right now a lot of our string operations leak memory. We need to treat strings like allocatable arrays, which get automatically deallocated when they go out of scope, and have copy semantics.
* The handling is quite complicated in the backend, with the associated bugs. We should treat strings like we do arrays, with multiple physical types.

There are three main physical types for strings (and two options, total of possible 5 physical types):

* pointer to a *null*-terminated (C)
* size + pointer to *non-null*-terminated string (Fortran, as well as `std::string_view` in C++); optionally also having a "capacity", allowing fast appending (like `std::string` and `std::vector` and our List)
* size + pointer to a *null*-terminated string (`std::string` in C++); optionally also having a "capacity", allowing fast appending (like `std::string` and `std::vector` and our List)

We need the first to interoperate with C. The second and third one are generally faster, especially the second one if it is just a non-owning view (although for that we would have to do something like a "pointer string", and also make sure it doesn't segfault). The third one is an owning string, so it's almost "free" to add the null at the end, which is what C++ does, then one can pass it to functions that operate on the first physical type. I believe in Fortran traditionally the second is the default, we can probably also do that. We probably want to always use the capacity as well, which makes it very fast for appending, which is quite a common operation for strings, just like it is for lists.

Overall, this is low priority, let's tackle this after fastGPT is done, and probably even later, once we get more codes working.

Thirumalai-Shaktivel:
@certik Do you think we can handle memory leaks using this approach?
I think I have to spend some time in LPython to handle the `strings_to_be_deallocated` in the backend: https://github.com/lcompilers/lpython/pull/2416. Instead, I thought that completely redesigning the strings would be worth spending time on.

certik:
Let's rename Character to String (at least in our minds only initially) to drive the point home that we only have a String type, not a char type. A single character can always be represented by `u8` or `i8` which are existing types in ASR.

For a String type, we need the following physical types:

* (pointer to a null-terminated string), non-owning: needed for LC's C strings (`char *s`) and for bind(c)
* (size, capacity, pointer to null-terminated string), owning: needed for LC's `std::string`; this can be rewritten with a pass to use a list of u8.
* (size, pointer to non-null terminated string), non-owning: LC's `std::string_view`, `std::span` for strings. --- not sure if we need to support it
* (size, pointer to non-null terminated string), owning: this is exactly a Fortran string

certik:
[...]

PR #5239, opened 2024-11-03
Utilize size and capacity for optimizing string allocation and copying

assem2002:
Towards #2257 
- Introduce extend_string() --> To only extend strings with correct capacity that can hold the required size.
- Utilize lfortran_strcpy_descriptor_string --> to either allocate or extend the string.
- Utilize lfortran_strcpy_pointer_strign --> to just copy from src to dest
- Utilize lfortran_alloc --> To create initial capacity for descriptor strings(allocatables)
- Utilize lfortran_string_write --> To copy string appropriatlely


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2008,29 +2008,91 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
...
 }
+// Allocate + Extend String ----------------------------------------------------------- 
+
+void extend_string(char** ptr, int32_t new_size /*Null-Character Included*/,
+    int64_t* current_size, int64_t* string_capacity){
+    ASSERT_MSG(current_size != NULL,"%s", "string size is NULL");
+    ASSERT_MSG(string_capacity != NULL, "%s", "string capacity is NULL");
+    ASSERT_MSG(*string_capacity >= *current_size, "%s", "compiler-behavior error : string capacity < string size");
+/* Remove this (It's for debugging purposes) */
+    ASSERT_MSG(strlen(*ptr) == (*current_size),"%s","Compiler-Internal Error : Size in string descriptor doesn't match actual size of char*");
```
assem2002, line 2019: I think this assertion should be removed as `strlen` would be expensive for long strings. We even could utilize the size in the descriptor string to be the return for such statement
```fortran
character(:), allocatable :: str
print *, len(str) ! HERE
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2008,29 +2008,91 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
...
+            }
+            *ptr = (char*)malloc(inital_capacity);
+            *string_capacity = inital_capacity;
+            const int8_t null_terminated_char_len = 1;
+
+            *string_size = desired_size - null_terminated_char_len;
+        } else if(*ptr != NULL && *string_capacity != 0){
+            // printf("runtime error: Attempting to allocate already allocated variable\n");
+            // exit(1);
+        } else {
```
assem2002, line 2052: This should be uncommented, but we face an issue with correct deallocations. I'll open an issue with an MRE. Once we can fix the MRE, We can uncomment this line.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2008,29 +2008,91 @@ LFORTRAN_API void _lfortran_strcat(char** s1, char** s2, char** dest)
...
-            x[0][i] = ' ';
+            extend_string(x, y_len+1, x_string_size, x__string_capacity);
+        }
+    }
+
+    if(!is_descriptor_string){
+        x_len = strlen(*x); // Stick to LHS-string-fixed size.
+    } else {
+        x_len = y_len; // LHS follows RHS size. 
+    }
```
certik, line 2084: I don't like this logic if it is descriptor or not.

In ASR, the assignment should only be between two fixed size strings, or two descriptor strings.

So in the backend, just write two functions: "copy_fixed_size" and "copy_descriptor", and in the LLVM backend dispatch correctly.

assem2002: I like that approach.

src/libasr/runtime/lfortran_intrinsics.c
assem2002: Related to this #5249. Once that is fixed we can turn that on.

src/libasr/runtime/lfortran_intrinsics.c
certik: I think this looks a lot better now!

certik:
I think it looks in the right direction now.

gxyd (approved):
This looks good to me.

@assem2002 , are you going to push anymore changes to this?

certik (approved):

PR #5241, opened 2024-11-04
chore: clear external_procedure after visiting subroutine

### LINKED ISSUE #5238: 'no implicit return type' error for an external function

For the below fortran program:
```f90
subroutine map_xi_to_c (cs, cv)
    USE iso_fortran_env
    implicit none
    real(REAL32), dimension(2), intent(in) :: cs
    real(REAL32), intent(out) :: cv
    real(REAL32), external :: fold
    cv=fold(cs(1))
    return
end

function fold (x0)
    use iso_fortran_env
    implicit none
    real(REAL32) :: fold
    real(REAL32) :: x0
end

program main
end program main

```

on it's execution with LFortran as below:
```console
$ lfortran --version
LFortran version: 0.41.0-184-g30e8beb84-dirty
Platform: macOS ARM
LLVM: 11.0.1

$ lfortran test.f90 --implicit-interface --implicit-typing
semantic error: No implicit return type available for `fold`
  --> test.f90:11:1 - 16:3
   |
11 |    function fold (x0)
   |    ^^^^^^^^^^^^^^^^^^...
...
   |
16 |    end
   | ...^^^ 


```

I've tried using the different LFortran compiler options i.e. `--implicit-interface` and `--implicit-typing`, and none of them work.

Pranavchiku:
Fixes #5238

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1496,6 +1496,7 @@ RUN(NAME external_05 LABELS gfortran llvmImplicit)
 RUN(NAME external_06 LABELS gfortran llvmImplicit)
 RUN(NAME external_07 LABELS gfortran llvmImplicit)
 RUN(NAME external_08 LABELS gfortran llvmImplicit)
+RUN(NAME external_09 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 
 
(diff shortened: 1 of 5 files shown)


gxyd:
This is a nice clean fix, looks good to me.

certik:
Great catch. A typical bug in a visitor pattern of not cleaning up the state. I wonder if there is an idiom to ensure these bugs don't happen.

Issue #5233, opened 2024-11-03
CI: remove `ubuntu-latest` from `third_party_compile` job


Pranavchiku:
This is because we already test all third party code under `Test LLVM` with all LLVM versions.

certik:
Well, but do we truly test it equivalently? So far we didn't. :)

Also I think I would like to make the LLVM 10 - 19 tests execute much faster, not be as thorough, but have these 3rd party tests thorough.

gxyd:
> tests execute much faster, not be as thorough

@certik , could you please explain what do you mean by "not be as thorough" here? Do you mean, not to test all LLVM versions between 10 through 19?

certik:
The LLVM 10-19 tests should execute the 3rd party tests quickly. The dedicated macos/linux 3rd party tests should test the 3rd party tests thoroughly, which can take a lot longer (e.g., dftatom), than we would do for LLVM 10-19.

Pranavchiku:
Ahh okay, I see.

Pranavchiku:
Linking https://github.com/lfortran/lfortran/pull/5206#issuecomment-2453953388, also here are a few observations:

* Test LLVM takes 23 minutes to execute
* Most time is taken by `dftatom`
* There were a few redundancy like downloading model in fastGPT, git cloning stdlib again, removed at #5242 
* Check Debug Build was not removed although merged in one test ( https://github.com/lfortran/lfortran/pull/5206#issuecomment-2453953388 )
* Merged `build_wasm_and_upload`

certik:
dftatom should be changed to only run a few tests for LLVM 10 - 19.

Pranavchiku:
Another idea is to combine `test_llvm_wasm` and `test_llvm` this will free two runners and running `test_llvm_wasm` portion only for LLVM version 10 and 16 inside `test_llvm`. I think it is possible.

Pranavchiku:
If we see https://github.com/lfortran/lfortran/pull/5281, we don't need to run any of LLVM backend tests, it is solely related to MLIR, although rare but checking these may be good way to optimise.

PR #5242, opened 2024-11-04
ci: refactor Test LLVM to speedup fastGPT and stdlib

Pranavchiku:


.github/workflows/CI.yml
```diff
@@ -1228,21 +1140,16 @@ jobs:
             FC=lfortran cmake . -DTEST_DRIVE_BUILD_TESTING=OFF -DBUILD_EXAMPLE=ON -DCMAKE_Fortran_COMPILER_WORKS=TRUE -DCMAKE_Fortran_FLAGS="--cpp --realloc-lhs"
             make -j8
             ctest
-            ./build_test_gf.sh
 
             # Test with experimental simplifier
-            cd ..
```
gxyd, line 1234: do we want to keep in sync, *Test stdlib* in `Test LLVM` and `third_party_code_compile` in sync with each other?

Pranavchiku: Refering to https://github.com/lfortran/lfortran/issues/5233#issuecomment-2453849959, no.

gxyd:
The changes look good me, only a few things:
* the changes made to third party code in `Test LLVM` can be replicated to `third_party_code_compile`?
* Currently there are two commits, the first commit makes some changes and the second one partially reverts some of the changes from the first one. Instead, I believe, it will be nicer to have the one commit deleting the *Check Debug build* job, one commit for refactoring fastGPT to not download the model again and one commit for stdlib changes, as the changes to CI are usually delicate, this will make it easy for us in the future look back and see what changes were done. Maybe you were experimenting with changes, that's why the changes are mostly clubbed into a single commit.

Pranavchiku:
> the changes made to third party code in Test LLVM can be replicated to third_party_code_compile?

It think we can, but let's keep those verbose so that if anyone is willing to execute specific portion can do. For test LLVM we can optimise.

gxyd:
> It think we can, but let's keep those verbose so that if anyone is willing to execute specific portion can do. For test LLVM we can optimise.

Sounds good to me.

certik (approved):
Ok, I think these changes are fine. There is an issue of only doing "git clean -dfx" inside the build directory that it assumes now extra files were created outside of it, which usually there shouldn't be, but sometimes there are, that would be a bug in the 3rd party build system.

Pranavchiku:
> There is an issue of only doing "git clean -dfx" inside the build directory that it assumes now extra files were created outside of it

We do create this outside, like `model.dat` file is stored outside. Reason is to reuse it and not download everytime.

PR #5252, opened 2024-11-05
enh: declare implicit variable logic to handle parameter value

### LINKED ISSUE #5251: Symbol cannot be found in module when using Fortran 77 style PARAMETER
Compiling the following fortran code:

```fortran
MODULE FIRST
  PARAMETER (X=90)
END MODULE FIRST

MODULE SECOND
  USE FIRST, ONLY: X
END MODULE SECOND
```

gives the following error:

```
semantic error: The symbol 'x' not found in the module 'first'
 --> SIZ.f90:6:3
  |
6 |   USE FIRST, ONLY: X
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 

```

I have tested it to work fine with gfortran.

Replacing the `PARAMETER (X=90)` with the more modern ` INTEGER, PARAMETER :: X = 90` works. The older `PARAMETER` statement should still be valid Fortran 2018.

Pranavchiku:
Fixes #5251


certik (approved):
I think that looks good, thanks for fixing it!

PR #5254, opened 2024-11-05
ci: merge `test_llvm_wasm` to `test_llvm` and test WASM for LLVM 10 and 19

Pranavchiku:
This PR does the following changes:

* Merges `Test LLVM WASM` to `Test LLVM` thus freeing up 2 jobs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1060,17 +1060,21 @@ jobs:
             git checkout 6e0f6e7506440b042aae93e57b10444ee6c36f7a
             export PATH="$(pwd)/../src/bin:$PATH"
             make -f Makefile.manual
+            make -f Makefile.manual test
 
             # Test with experimental simplifier
             git clean -dfx
             make -f Makefile.manual F90="lfortran --experimental-simplifier"
+            make -f Makefile.manual test
 
             git clean -dfx
             make -f Makefile.manual F90="lfortran --skip-pass=inline_function_calls,fma --fast"
+            make -f Makefile.manual test
 
             # Test with experimental simplifier
             git clean -dfx
             make -f Makefile.manual F90="lfortran --experimental-simplifier --skip-pass=inline_function_calls,fma --fast"
+            make -f Makefile.manual test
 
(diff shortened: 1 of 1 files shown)


Pranavchiku:
Cancelled workflow as both changes work, pushing correct patch.

certik:
Can you keep the original name for tests? We can rename it in a separate PR, but it will make things easier on me for this PR.

.github/workflows/CI.yml
```diff
@@ -976,21 +1060,17 @@ jobs:
...
-            make -f Makefile.manual test
 
             git clean -dfx
             make -f Makefile.manual F90="lfortran --skip-pass=inline_function_calls,fma --fast"
-            make -f Makefile.manual test
 
             # Test with experimental simplifier
             git clean -dfx
             make -f Makefile.manual F90="lfortran --experimental-simplifier --skip-pass=inline_function_calls,fma --fast"
-            make -f Makefile.manual test
```
certik, line 993: This should still run a test, but possibly just one. There is a way to do it. Let's do it in a separate PR.

certik:
Also let's do the WASM check for LLVM 10 and 19. The idea is to always do the latest and oldest LLVM version.

.github/workflows/CI.yml
```diff
@@ -845,6 +845,37 @@ jobs:
...
+          environment-file: ci/environment_linux_llvm.yml
+          create-args: >-
+            python=${{ matrix.python-version }}
+            llvmdev=${{ matrix.llvm-version }}
+            bison=3.4
+            openblas=0.3.21
+            llvm-openmp=14.0.4
+            cmake=3.26.4
+            make=4.3
+            nodejs=18.20.4
```
Pranavchiku, line 860: If we use `nodejs=18.13.0` it throws version conflicts and if we use `nodejs=18.20.4` with `llvmdev=10` again version conflicts hence have to separate two jobs.

certik: That's good.

certik (approved):
This looks good, thanks for doing it. I'll merge all other PRs first, then we'll merge this one.

PR #5255, opened 2024-11-05
fix: return non-zero exit code with `--continue-compilation`

### LINKED ISSUE #5248: Return non-zero exit code for --continue-compilation if there are errors
https://github.com/lfortran/lfortran/pull/5230/files#r1827993610

Pranavchiku:
Fixes #5248

We now return non-zero code with `--show-llvm`, `--show-fortran`, `--show-asr`, binary generation and execution when `--continue-compilation` is on and there are error messages.

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -628,6 +628,7 @@ int python_wrapper(const std::string &infile, std::string array_order,
     LCompilers::diag::Diagnostics diagnostics;
     LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
         r = fe.get_asr2(input, lm, diagnostics);
+    bool has_error_w_cc = diagnostics.has_error() && compiler_options.continue_compilation;
     std::cerr << diagnostics.render(lm, compiler_options);
     if (!r.ok) {
         LCOMPILERS_ASSERT(diagnostics.has_error())
@@ -663,7 +664,7 @@ int python_wrapper(const std::string &infile, std::string array_order,
         std::cout << LCompilers::pickle(*asr, compiler_options.use_colors, compiler_options.indent,
                 compiler_options.po.with_intrinsic_mods) << std::endl;
     }
-    return 0;
+    return has_error_w_cc;
 }
 
 int emit_cpp(const std::string &infile, CompilerOptions &compiler_options)
@@ -766,7 +767,7 @@ int emit_fortran(const std::string &infile, CompilerOptions &compiler_options) {
     std::cerr << diagnostics.render(lm, compiler_options);
     if (src.ok) {
         std::cout << src.result;
-        return 0;
+        return diagnostics.has_error() && compiler_options.continue_compilation;
     } else {
         LCOMPILERS_ASSERT(diagnostics.has_error())
         return 1;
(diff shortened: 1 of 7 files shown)


certik (approved):
Thanks. I think this is better. This needs to be rebased and updated, since I merged the other PR that added another test.

gxyd:
Nice, this is exactly what I was think about the other PR I just merged for the `echo $?` returning `0`, even after a raised `SemanticError`

src/bin/lfortran.cpp
```diff
@@ -628,6 +628,7 @@ int python_wrapper(const std::string &infile, std::string array_order,
     LCompilers::diag::Diagnostics diagnostics;
     LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
         r = fe.get_asr2(input, lm, diagnostics);
+    bool has_error_w_cc = diagnostics.has_error() && compiler_options.continue_compilation;
```
gxyd, line 631: How about if we do the `compiler_options.continue` first, i.e.:
```diff
diff --git a/src/bin/lfortran.cpp b/src/bin/lfortran.cpp
index 0cdefdf7f..4319366db 100644
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -628,7 +628,7 @@ int python_wrapper(const std::string &infile, std::string array_order,
     LCompilers::diag::Diagnostics diagnostics;
     LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
         r = fe.get_asr2(input, lm, diagnostics);
-    bool has_error_w_cc = diagnostics.has_error() && compiler_options.continue_compilation;
+    bool has_error_w_cc = compiler_options.continue_compilation && diagnostics.has_error();
     std::cerr << diagnostics.render(lm, compiler_options);
     if (!r.ok) {
         LCOMPILERS_ASSERT(diagnostics.has_error())
```

?

that will present checking the `Diagnostics` object for error messages.

src/bin/lfortran.cpp
```diff
@@ -766,7 +767,7 @@ int emit_fortran(const std::string &infile, CompilerOptions &compiler_options) {
     std::cerr << diagnostics.render(lm, compiler_options);
     if (src.ok) {
         std::cout << src.result;
-        return 0;
+        return diagnostics.has_error() && compiler_options.continue_compilation;
```
gxyd, line 770: Same with other places, i.e. having `compiler_options.continue_compilation` before `diagnostics.has_error()` call.

gxyd (commented):
Other that one comment, this looks good to me.

Also, I merged another PR: https://github.com/lfortran/lfortran/pull/5257, which might make this PR require merging *main* into this and updating this PR again, sorry, I didn't see this PR before I merged the other PR.

src/bin/lfortran.cpp
Pranavchiku: Yes it makes sense, I'll do it, but I think we are over engineering ( over fundamentiling ) things here a bit XD

certik: It sounds like a minor thing, but actually @gxyd's point is spot on. Consider `has_error`:
```c++
bool Diagnostics::has_error() const {
    for (auto &d : this->diagnostics) {
        if (d.level == Level::Error) return true;
    }
    return false;
}
```
So in the original code we would now be calling this every time we compile any code, thus slowing the compilation down. In the new code this is avoided. 

Yes, it might seem like minor, but all these little things add up. The key is not "not pessimize" the code. So if we know that something is expensive, and another solution is less expensive, we should always use the less expensive version by default.

By consistently doing this in the whole code base ensures good speed.

Pranavchiku: Yes, but the fact that diagnostics will only have 1 error message by nature of it because it throws an exception and compilation stops, it is just like checking if the single element diagnostic has Level::Error.

Maybe I am wrong if we have warnings, style suggestions then it will hurt us. I forgot about this case previously and came to me while writing. 

So if we forget the warnings, style suggestions, then either of them is same but that is not the case so yeah, the suggested is better.

PR #5256, opened 2024-11-05
Add compiler flag for LLVM stacktraces

advikkabra:
Adds `WITH_LFORTRAN_LLVM_STACKTRACE` flag to optionally compile with LLVM stacktraces.
I have currently either required WITH_BFD to explicitly be set as yes, or LLVM stacktraces are used by default.


certik:
Excellent, thanks for working on this!

CMakeLists.txt
```diff
@@ -303,7 +305,9 @@ if (WITH_STACKTRACE)
         endif()
     else()
         set(WITH_LINKH yes)
-        set(WITH_BFD yes)
+        if (NOT WITH_BFD)
```
gxyd, line 308: So now, the only way to set `WITH_BFD` as true is when in the build configuration someone specifies `DWITH_BFD=yes`? (Earlier, when a LFortran developer on Linux machine would've specified `DWITH_STACKTRACE` it would've been set as `true`)

I'm curious, would someone explicitly do that?

gxyd: Apologies for not being familiar with which libraries are needed for stack traces on different platforms.

advikkabra: Yes, if DWITH_BFD is not explicitly set, LLVM stacktraces are used. I think LLVM stacktraces are better in most usecases, so I kept it like that. Maybe I can set WITH_BFD to automatically be yes when WITH_LLVM is set as no?

gxyd: That expanded explanation helps understanding for why we did what we've done here. Thanks @certik .

PR #5257, opened 2024-11-05
Error resilient array_04

AnshMehta1:
Toward :- #5036 

I am not sure that should we handle this error file.

Error message for this file is thrown through asr_utils and so , didn't break the semantic error into diagnostic and semantic abort and instead used catch for semantic error

```
File "/mnt/d/lfortran2/lfortran/src/libasr/asr_utils.h", line 4752, in LCompilers::ASR::expr_t* LCompilers::ASRUtils::get_bound<LCompilers::LFortran::SemanticError>(LCompilers::ASR::expr_t*, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Allocator&)
    throw SemanticError(msg, arr_expr->base.loc);
semantic error: Type member a is not an array so it cannot be indexed.
 --> try.f90:8:1
  |
8 | b%a(:) = 1
  | ^^^^^^


```

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -126,6 +126,14 @@ public:
                     tmp = nullptr;
                     tmp_vec.clear();
                 }
+            } catch (const SemanticError &e) {
+                if (!compiler_options.continue_compilation) {
+                    throw e;
+                } else {
+                    diag.add(e.d);
+                    tmp = nullptr;
+                    tmp_vec.clear();
+                }
             }
             if (tmp != nullptr) {
                 ASR::stmt_t* tmp_stmt = ASRUtils::STMT(tmp);
--- a/tests/errors/array_04_cc.f90
+++ b/tests/errors/array_04_cc.f90
@@ -1,0 +1,11 @@
+program array_04
+implicit none
+
+type :: t
+    integer :: a
+end type
+type(t) :: b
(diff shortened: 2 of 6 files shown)


certik:
I think that's fine for now. The SemanticError is passed as a template parameter, so libasr does not actually depend on LFortran.

We can later refactor it for `get_bound` to return true/false to indicate if there was an error message, maybe we can give it a diagnostic argument so that it can append the message there, and then let the caller handle raising an exception if needed.

Pranavchiku (approved):
I think this is fine, we are skipping an entire statement if there are any errors in it. We can make it more granular in future but this works for now.

gxyd:
@certik Unrelated to the above PR query I've:

> The SemanticError is passed as a template parameter, so libasr does not actually depend on LFortran.

I see, so `SemanticError` is never used in asr_utils.h directly but passed in as template parameter in it. The comment for `SemanticError` says (here: https://github.com/lfortran/lfortran/blob/main/src/lfortran/semantics/semantic_exception.h#L8-L9):
```
// This exception is only used internally in the lfortran/semantics/ directory
// and in lfortran/asr_utils.h/cpp. Nowhere else.
```

So, we don't actually use `SemanticError` in asr_utils.h either, right? It's always passed in as a template parameter.

gxyd (approved):
The PR looks good to me.

certik:
@gxyd yes, that's correct.

PR #5265, opened 2024-11-06
allow 'derived type' without start 'program' in parser

### LINKED ISSUE #5250: type cannot be declared without program statement
test code:
```fortran
type :: test_t
  real :: a
end type
end
```

output:
```
$ lfortran --version
LFortran version: 0.41.0
Platform: Linux
$ lfortran test_program.f90 -o test_program
syntax error: Token '::' is unexpected here
 --> test_program.f90:1:6
  |
1 | type :: test_t
  |      ^^ 

syntax error: Token 'end type' is unexpected here
 --> test_program.f90:3:1
  |
3 | end type
  | ^^^^^^^^ 


```

To circumvent the issue, one can add program statement in the beginning, but that should not be necessary:
```fortran
program test
type :: test_t
  real :: a
end type
end
```

gxyd:
## Description

Fixes #5250


tests/parser/derived_type_without_start_program.f90
```diff
@@ -0,0 +1,8 @@
+! Test to ensure that 'type' declaration without
+! a 'program' statement at the beginning parses
+! correctly
+    implicit none
```
Pranavchiku, line 4:
```suggestion
```

Why do we need an `implicit none`?

tests/parser/derived_type_without_start_program.f90
```diff
@@ -0,0 +1,8 @@
+! Test to ensure that 'type' declaration without
+! a 'program' statement at the beginning parses
+! correctly
```
Pranavchiku, line 3:
```suggestion
```

Let's test exact code, or assure that comments do not change code behavior

Pranavchiku (changes requested):
requested changes and left a question above

tests/parser/derived_type_without_start_program.f90
gxyd: We don't necessarily need it, but that doesn't matter whether we've it or not, cause both will cause the same error as is reported in the original issue.

src/lfortran/parser/parser.yy
```diff
@@ -610,6 +610,7 @@ script_unit
     | use_statement
     | implicit_statement
```
Pranavchiku, line 611: This worries me, because now `implicit none` is as `implicit_statement`, so we are mixing two script_unit

tests/parser/derived_type_without_start_program.f90
Pranavchiku: Refer: https://github.com/lfortran/lfortran/pull/5265/files#r1831201797

Pranavchiku (approved):
Looks good, thanks for fixing it!

certik (approved):
I think this looks good. It adds the derived type declaration to `script_unit` which allows to use it interactively as well as when "program" is missing, as it gets converted later.

So this is the way to fix it.

PR #5277, opened 2024-11-07
enh: frontend to support keyword argument in array reshape

### LINKED ISSUE #5258: (PRIMA) Keyword Arguments are not working properly in Reshape
Towards #4397 

MRE : 
```fortran
program main
integer :: A(2,2)
A = reshape(source = [1,2,2,3], shape = shape(A))
print *, A
end program
```

LFortran Error : 
```console
semantic error: reshape accepts only 2 arguments, got 0 arguments instead.
 --> a.f90:3:5
  |
3 | A = reshape(source = [1,2,2,3], shape = shape(A))
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```
Works with GFortran.

Pranavchiku:
Fixes #5258. With this we can revert https://github.com/Pranavchiku/prima/commit/0d4240bc9094159260ff1d2334f1e8d4356b5782


gxyd (approved):
I guess, we can have more than two kwargs to [`ArrayReshape`](https://gcc.gnu.org/onlinedocs/gcc-4.4.3/gfortran/RESHAPE.html) with `pad` and `order` are also possible, but I guess, that's something to handle later. This looks good to me.

Issue #5260, opened 2024-11-05
enh: fortran backend to support openmp dump


Pranavchiku:
This is much required feature for anyone to understand how openmp pass works and therefore I'll ask @adit4443ya to prioritise it before getting on to implement `collapse` in openmp pass. This will help him to understand how openmp pass works and help streamline development of it.

```fortran
program openmp_05
use omp_lib
integer :: n = 10, i

call omp_set_num_threads(4)

!$omp parallel private(i) shared(n)
!$omp do
do i = 1, n
    print *, "xyz"
    print *, "i = ", i, "from thread = ", omp_get_thread_num()
    if (omp_get_thread_num() > 4) error stop
end do
!$omp end do
!$omp end parallel

print *, "end program"

end program
```

```console
(lf) pranavchiku@pranavchikus-MacBook-Air:~/repos/lfortran(ci-speedup-2) % lfortran a.f90 --dump-all-passes-fortran
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2673, in main
    return main_app(argc, argv);
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2467, in main_app(int, char**)
    dump_all_passes(arg_file, compiler_options);
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 789, in (anonymous namespace)::dump_all_passes(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, LCompilers::CompilerOptions&)
    LCompilers::Result<LCompilers::ASR::TranslationUnit_t*> asr = fe.get_asr2(input, lm, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 270, in LCompilers::FortranEvaluator::get_asr2(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics);
[...]

PR #5278, opened 2024-11-07
enh: Fortran backend to support openmp dump

adit4443ya:
Towards [ Fortran Backend](https://github.com/lfortran/lfortran/issues/5260)

Starting with minimal fix

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1831,6 +1831,7 @@ RUN(NAME openmp_36 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_37 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_38 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
+RUN(NAME openmp_40 LABELS gfortran fortran GFORTRAN_ARGS -fopenmp)
 
 RUN(NAME compiler_version_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1831,7 +1831,8 @@ RUN(NAME openmp_36 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_37 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_38 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
-RUN(NAME openmp_40 LABELS gfortran fortran GFORTRAN_ARGS -fopenmp)
+
+RUN(NAME fortran_01 LABELS gfortran llvm fortran)
 
 RUN(NAME compiler_version_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
--- a/integration_tests/openmp_40.f90
+++ b/integration_tests/openmp_40.f90
@@ -1,7 +1,7 @@
-program openmp_40
-    use omp_lib
-    integer :: i
-    do i = 1, 5
-        print *, i
-    end do 
+program main
+    use iso_c_binding, only: c_ptr!, c_loc
+    implicit none
+    integer, pointer :: x
+    type(c_ptr) :: ptr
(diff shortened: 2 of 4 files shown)


gxyd:
A test case would be nice here.

adit4443ya:
It will work on every openmp testcase
If there is some new testcase needed then i will add it 

> A test case would be nice here.

adit4443ya:
For time being we can consider this as an example 
``` fortran
program openmp_36
integer :: i
do concurrent ( i =1:5 )
    print *, i
end do 
end program
```

> A test case would be nice here.

gxyd:
> For time being we can consider this as an example
> 
> ```fortran
> program openmp_36
> integer :: i
> do concurrent ( i =1:5 )
>     print *, i
> end do 
> end program
> ```
> 

I see, that's a good start, once you get any of the program present in `integration_tests/CMakeLists.txt` (pertaining to openmp), you can probably just add a label `fortran` to it (those can act as a good test case for this).

I moved this PR to draft mode until then.

Pranavchiku:
Let's add a test which this PR fixes, a minimal, close this and do improvements in next PR. Let's not carry longer PRs.

adit4443ya:
This fix now generates dump passes which uses `omp_lib`
```fortran
program openmp_40
    use omp_lib
    integer :: i
    do i = 1, 5
        print *, i
    end do 
end program
```
So i can add this test if required and then we will move on to merge it and then I will fix for the `DoConcurrentLoop` in next PR

certik:
The error is:
```
CMakeFiles/openmp_40.dir/openmp_40.f90.o.tmp.f90:95:20:

   95 |         type(c_ptr), value :: data
      |                    1
Error: Derived type ‘c_ptr’ at (1) is being used before it is defined
```

integration_tests/openmp_40.f90
```diff
@@ -0,0 +1,7 @@
+program openmp_40
+    use omp_lib
+    integer :: i
+    do i = 1, 5
+        print *, i
+    end do 
+end program
```
Pranavchiku, line 7:
```suggestion
program main
    use iso_c_binding, only: c_ptr!, c_loc
    implicit none
    integer, pointer :: x
    type(c_ptr) :: ptr
    !ptr = c_loc(x)
end program
```

Pranavchiku: Use this instead, you'll need `c_loc` as well so slowly keep adding all these. These are minor changes and can be done in small but multiple PRs

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -246,6 +246,9 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
                     import_struct_type.push_back(struct_name);
                 }
                 break;
+            } case ASR::ttypeType::CPtr: {
+                r = "type(c_ptr)";
```
Pranavchiku, line 250: The fact that if the code has `c_ptr` usage it will eventually be doing `use iso_c_binding`, so it should work, maybe the example you used is complex.

Pranavchiku (changes requested):
left few comments to get this PR in shape

integration_tests/CMakeLists.txt
```diff
@@ -1831,6 +1831,7 @@ RUN(NAME openmp_36 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_37 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_38 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
+RUN(NAME openmp_40 LABELS gfortran fortran GFORTRAN_ARGS -fopenmp)
```
Pranavchiku, line 1834:
```suggestion
RUN(NAME openmp_40 LABELS gfortran llvm fortran)
```

integration_tests/openmp_40.f90
```diff
@@ -0,0 +1,7 @@
+program main
```
Pranavchiku, line 1:
```suggestion
program fortran_01
```

Rename the test case

integration_tests/CMakeLists.txt
Pranavchiku: use the renamed case here

Pranavchiku (changes requested):
left few comments

integration_tests/openmp_40.f90
adit4443ya: For now i guess this PR is good to go, in next PR i would handle `c_loc` case

certik:
Let's get into the habit of creating a nice set of commits. Or squash it.

PR #5281, opened 2024-11-07
[MLIR][Refactor] Rename b (OpBuilder) to builder

Thirumalai-Shaktivel:


src/libasr/codegen/asr_to_mlir.cpp
```diff
@@ -41,7 +41,7 @@ class ASRToMLIRVisitor : public ASR::BaseVisitor<ASRToMLIRVisitor>
     std::string src;
 
     std::unique_ptr<mlir::MLIRContext> context;
-    std::unique_ptr<mlir::OpBuilder> b;
+    std::unique_ptr<mlir::OpBuilder> builder;
```
Thirumalai-Shaktivel, line 44: Minor:

@certik, I feel there is no need for builder to be unique_ptr, what do you think?

Pranavchiku:
Always feel good to see your PRs :)

Pranavchiku (approved):

src/libasr/codegen/asr_to_mlir.cpp
certik: Probably it doesn't need to be `unique_ptr`.

Thirumalai-Shaktivel:
Thanks for the reviews!

Issue #5286, opened 2024-11-07
ICE on missing file


pjh40:
Trying to run 
```sh
src/bin/lfortran no_such_file.f90
```
where `no_such_file.f90` doesn't exist results in:
```sh
Internal Compiler Error: Unhandled exception
```
on `main` branch, SHA 2d0376d42c22b5b026

pjh40:
`main` is catching the `LCompilers::LCompilersException("File does not exist: " + arg_file)` exception thrown in `main_app`, but the ICE message is way too "scary" for a missing input file.

harperjf:
This bug was reported in #3796 in April and I thought it had been fixed. It hasn't.

certik:
Fixed by https://github.com/lfortran/lfortran/pull/5289.

Issue #3796, opened 2024-04-06
ICE for nonexistent program file


harperjf:
Having heard many times that internal compiler errors should always be reported to those responsible for compilers whether the program is valid or not, I give lfortran's otherwise correct response when it was asked to compile a nonexistent file.
```
(lf2) john:~/Jfh$ lfortran Foobar.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x726858
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29e3f
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29d8f
LCompilersException: File does not exist: Foobar.f90
(lf2) john:~/Jfh$
```

certik:
Thanks! Yes, this needs to be a proper error message.

certik:
Fixed by https://github.com/lfortran/lfortran/pull/5289.

PR #741, opened 2022-09-23
fix: throw error when infile does not exist (#740)

lucifer1004:
Fix #740

PR #5289, opened 2024-11-07
Reduce severity of missing input file

pjh40:
Converts the response to a specified but missing input file from an `LCompilersException` to a message on `std::cerr` and a non-zero return code, both modeled on clang's response.  This addresses issue #5286 (which is a duplicate of issue #3796)


certik (approved):
Awesome, thank you @pjh40 for fixing this!

This was initially fixed by https://github.com/lfortran/lfortran/pull/741, but we should have just printed the error and exit like in this PR, not raise an exception.

Issue #5294, opened 2024-11-08
subroutine call with ExternalSymbol raises asr verify error


gxyd:
## Minimal Reproducible Example

for the below fortran program, where the `ExternalSymbol`: `g`, is passed as an argument to subroutine call `sub2` raises asr verify error, when `--implicit-interface` flag is used, otherwise, it works without error.

```f90
module vars
    implicit none
    integer :: g = 2
end module

subroutine sub1
    use vars
    implicit none
    call sub2 (g)
end subroutine

subroutine sub2 (x)
    implicit none
    integer :: x
end
```

on it's execution with LFortran as:
```console
> lfortran --version
LFortran version: 0.41.0-281-g00ab9ca50-dirty
Platform: macOS ARM
LLVM: 11.0.1
Default target: arm64-apple-darwin23.6.

> lfortran test.f90 --implicit-interface
ASR verify pass error: ASR verify: The symbol table was not found in the scope of `symtab`.
 --> /Users/gxyd/OpenSource/lfortran/test.f90:3:16
  |
3 |     integer :: g = 2
  |                ^^^^^ failed here

ASR verify pass error: Var::m_v `g` cannot point outside of its symbol table
 --> /Users/gxyd/OpenSource/lfortran/test.f90:9:16
  |
9 |     call sub2 (g)
  |                ^ failed here


> lfortran -c test.f90     # execution without `--implicit-interface` flag
> echo $?
0
```

PR #5295, opened 2024-11-08
chore: don't call replacer on ExternalSymbol with the same underling symbol

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5294

The program linked in the issue (or the one added as a test case in this PR) already works without any use of `--implicit-interface` (and it's expected to work), but this PR focuses on making that program work with use of `--implicit-interface` as well, as *POT3D* needs `--implicit-interface` flag necessarily.

## Changes in ASR

Currently the MRE in the linked issue when run with `--implicit-interface` on the current-main raises the below error:
```console
ASR verify pass error: ASR verify: The symbol table was not found in the scope of `symtab`.
 --> /Users/gxyd/OpenSource/lfortran/test.f90:3:16
  |
3 |     integer :: g = 2
  |                ^^^^^ failed here

ASR verify pass error: Var::m_v `g` cannot point outside of its symbol table
 --> /Users/gxyd/OpenSource/lfortran/test.f90:9:16
  |
9 |     call sub2 (g)
  |                ^ failed here

```

### ASR in *main*
so to still be able to get the ASR, I commented out ASR verify check on *main* local branch and below is the ASR generated by `lfortran test.f90 --impliicit-interface --show-asr`:
```clj
(TranslationUnit
    (SymbolTable
        1
        {
            sub:
                (Function
                    (SymbolTable
                        4
                        {
                            arg:
                                (Variable
                                    4
                                    arg
                                    []
                                    Unspecified
                                    ()
                                    ()
                                    Default
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    sub
                    (FunctionType
                        [(Integer 4)]
                        ()
                        Source
                        Implementation
                        ()
                        .false.
                        .false.
                        .false.
                        .false.
                        .false.
                        []
                        .false.
                    )
                    []
                    [(Var 4 arg)]
                    []
                    ()
                    Public
                    .false.
                    .false.
                    ()
                ),
            vars:
                (Module
                    (SymbolTable
                        2
                        {
                            var_g:
                                (Variable
                                    2
                                    var_g
                                    []
                                    Local
                                    (IntegerConstant 2 (Integer 4) Decimal)
                                    (IntegerConstant 2 (Integer 4) Decimal)
                                    Save
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    vars
                    []
                    .false.
                    .false.
                ),
            wrapper:
                (Function
                    (SymbolTable
                        3
                        {
                            var_g:
                                (ExternalSymbol
                                    3
                                    var_g
                                    2 var_g
                                    vars
                                    []
                                    var_g
                                    Public
                                )
                        })
                    wrapper
                    (FunctionType
                        []
                        ()
                        Source
                        Implementation
                        ()
                        .false.
                        .false.
                        .false.
                        .false.
                        .false.
                        []
                        .false.
                    )
                    [sub]
                    []
                    [(SubroutineCall
                        1 sub
                        ()
                        [((Var 2 var_g))]     ;; notice this part of the ASR only which is concerned
                        ()
                    )]
                    ()
                    Public
                    .false.
                    .false.
                    ()
                )
        })
    []
)
```

### ASR in this PR

generated by `lfortran test.f90 --implicit-interface --show-asr`
```clj
(TranslationUnit
    (SymbolTable
        1
        {
            sub:
                (Function
                    (SymbolTable
                        4
                        {
                            arg:
                                (Variable
                                    4
                                    arg
                                    []
                                    Unspecified
                                    ()
                                    ()
                                    Default
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    sub
                    (FunctionType
                        [(Integer 4)]
                        ()
                        Source
                        Implementation
                        ()
                        .false.
                        .false.
                        .false.
                        .false.
                        .false.
                        []
                        .false.
                    )
                    []
                    [(Var 4 arg)]
                    []
                    ()
                    Public
                    .false.
                    .false.
                    ()
                ),
            vars:
                (Module
                    (SymbolTable
                        2
                        {
                            var_g:
                                (Variable
                                    2
                                    var_g
                                    []
                                    Local
                                    (IntegerConstant 2 (Integer 4) Decimal)
                                    (IntegerConstant 2 (Integer 4) Decimal)
                                    Save
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    vars
                    []
                    .false.
                    .false.
                ),
            wrapper:
                (Function
                    (SymbolTable
                        3
                        {
                            var_g:
                                (ExternalSymbol
                                    3
                                    var_g
                                    2 var_g
                                    vars
                                    []
                                    var_g
                                    Public
                                )
                        })
                    wrapper
                    (FunctionType
                        []
                        ()
                        Source
                        Implementation
                        ()
                        .false.
                        .false.
                        .false.
                        .false.
                        .false.
                        []
                        .false.
                    )
                    [sub]
                    []
                    [(SubroutineCall
                        1 sub
                        ()
                        [((Var 3 var_g))]   ;; notice the diff, now the SymbolTable is '3' which is correct
                        ()
                    )]
                    ()
                    Public
                    .false.
                    .false.
                    ()
                )
        })
    []
)
```

the ASR generated above is the same as what it would be without use of `--implicit-interface` flag

### DIFF
--- a/src/libasr/asr_utils.cpp
+++ b/src/libasr/asr_utils.cpp
@@ -202,7 +202,7 @@ void update_call_args(Allocator &al, SymbolTable *current_scope, bool implicit_i
                 ASR::symbol_t* arg_sym = arg_var->m_v;
                 ASR::symbol_t* arg_sym_underlying = ASRUtils::symbol_get_past_external(arg_sym);
                 ASR::symbol_t* sym = fetch_sym(arg_sym_underlying);
-                if (sym != arg_sym) {
+                if (sym != arg_sym_underlying) {
                     ASR::expr_t** current_expr_copy = current_expr;
                     current_expr = const_cast<ASR::expr_t**>((expr_to_replace));
                     this->call_replacer_(sym);
@@ -255,7 +255,7 @@ void update_call_args(Allocator &al, SymbolTable *current_scope, bool implicit_i
                     ASR::symbol_t* arg_sym = arg_var->m_v;
                     ASR::symbol_t* arg_sym_underlying = ASRUtils::symbol_get_past_external(arg_sym);
                     ASR::symbol_t* sym = fetch_sym(arg_sym_underlying);
-                    if (sym != arg_sym) {
+                    if (sym != arg_sym_underlying) {
                         ASR::expr_t** current_expr_copy = current_expr;
                         current_expr = const_cast<ASR::expr_t**>(&(func->m_args[i]));
                         this->call_replacer_(sym);
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1534,6 +1534,7 @@ RUN(NAME external_06 LABELS gfortran llvmImplicit)
 RUN(NAME external_07 LABELS gfortran llvmImplicit)
 RUN(NAME external_08 LABELS gfortran llvmImplicit)
 RUN(NAME external_09 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
+RUN(NAME external_10 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 
 
 RUN(NAME interface_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/external_10.f90
+++ b/integration_tests/external_10.f90
@@ -1,0 +1,25 @@
+module external_10_module
+    implicit none
+    integer :: g = 2
+end module
+
+subroutine sub1_external_10
+    use external_10_module
+    implicit none
+    call sub2_external_10 (g)
+    print *, g
+    if (g /= 10) error stop
+end subroutine
+
+subroutine sub2_external_10 (x)
+    implicit none
+    integer, intent(inout) :: x
+    print *, x
+    if (x /= 2) error stop
+    x = 10
+end subroutine
+
+program external_10
+    implicit none
+    call sub1_external_10
+end program


src/libasr/asr_utils.cpp
```diff
@@ -202,7 +202,7 @@ void update_call_args(Allocator &al, SymbolTable *current_scope, bool implicit_i
                 ASR::symbol_t* arg_sym = arg_var->m_v;
                 ASR::symbol_t* arg_sym_underlying = ASRUtils::symbol_get_past_external(arg_sym);
                 ASR::symbol_t* sym = fetch_sym(arg_sym_underlying);
-                if (sym != arg_sym) {
+                if (sym != arg_sym_underlying) {
```
Pranavchiku, line 205: we cannot do this, it will always be true ( if function got updated ). The aim is to check if newly created function is updated or not ( not sure but something like that ). 

`arg_sym` is what we have as an argument, the external function is `arg_sym_underlying`, now as `--implicit-interface` have empty args and thus as soon as it encounters call to it with `n` args we change the external function. This causes a problem that now the call is to `old` function which is `arg_sym_underlying`. So, we then replace it with `fetch_sym(arg_sym_underlying)` which fetches the new sym which must be used.

Pranavchiku (changes requested):
If I take the MRE:

```fortran
module vars
    implicit none
    integer :: g = 2
end module

subroutine sub1
    use vars
    implicit none
    call sub2 (g)
end subroutine

subroutine sub2 (x)
    implicit none
    integer :: x
end
```

This shall not even need `--implicit-interface` but POT3D needs it, so if it fails it means with the flag we create an empty interface for `sub2` which we shall not. So, we need to understand how it works without `--implicit-interface` flag and then fix it. I don't think this PR is a correct fix for it right now. Feel free to ping me for any questions in case you need further clearance, I'll be happy to answer.

gxyd:
@certik @Pranavchiku , kindly review the PR again, and I've updated the PR description to describe a little more in detail for what this PR changes, I haven't made any changes since I last submitted the PR for review when comments by Pranav were left. I've gone through the changes to get a better understanding of the change made and I feel confident that this might actually be the right fix for it.

gxyd:
>  @gxyd does't it?

I thought about the change properly this time, it's possible that maybe Pranav misunderstood the change, that's where the updated PR description might help understanding the change this PR brings in. Happy to discuss on this again.

gxyd:
@Pranavchiku 's comment above which briefly mentioned:
> So, we need to understand how it works without --implicit-interface flag and then fix it.

This is what this PR does, i.e. ensures that the ASR generated with/without `--implicit-interface` flag is same, currently in *main*, it isn't.

src/libasr/asr_utils.cpp
Pranavchiku: I am thinking on this, give me more time

Pranavchiku (approved):
I am convinced, so the things is we need to check if `old_function` ( `arg_sym_underlying` if it is in Module ) is different to new one i.e. `sym`.

gxyd:
> I am convinced, so the things is we need to check if `old_function` ( `arg_sym_underlying` if it is in Module ) is different to new one i.e. `sym`.

Thanks for the review @Pranavchiku , really appreciate it.

Also, I'm glad that my understanding here was along the lines of how you described it now.

gxyd:
I'll go ahead with the merge here.

certik (approved):
Thanks! Thanks @Pranavchiku for the review.

PR #5296, opened 2024-11-08
fix: PR#5247 by applying review

Pranavchiku:
Fixes #5247

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1656,11 +1656,18 @@ public:
                                           size_t &value_index) {
         ASR::ImpliedDoLoop_t *implied_do_loop = ASR::down_cast<ASR::ImpliedDoLoop_t>(implied_do_loop_expr);
 
-        ASR::expr_t* loop_start_expr = ASRUtils::expr_value(implied_do_loop->m_start);
-        ASR::expr_t* loop_end_expr = ASRUtils::expr_value(implied_do_loop->m_end);
+        ASR::expr_t* loop_start_expr = implied_do_loop->m_start;
+        ASR::expr_t* loop_end_expr = implied_do_loop->m_end;
         ASR::expr_t* loop_increment_expr = implied_do_loop->m_increment;
-        if (loop_increment_expr) {
-            loop_increment_expr = ASRUtils::expr_value(loop_increment_expr);
+        ASR::expr_t* start_expr_value = ASRUtils::expr_value(loop_start_expr);
+        ASR::expr_t* end_expr_value = ASRUtils::expr_value(loop_end_expr);
+        if (!start_expr_value) {
+            throw SemanticError("The start variable of the data implied do loop must be constants",
+                loop_start_expr->base.loc);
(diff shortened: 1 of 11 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1656,11 +1656,18 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                           size_t &value_index) {
         ASR::ImpliedDoLoop_t *implied_do_loop = ASR::down_cast<ASR::ImpliedDoLoop_t>(implied_do_loop_expr);
 
-        ASR::expr_t* loop_start_expr = ASRUtils::expr_value(implied_do_loop->m_start);
-        ASR::expr_t* loop_end_expr = ASRUtils::expr_value(implied_do_loop->m_end);
+        ASR::expr_t* loop_start_expr = implied_do_loop->m_start;
```
gxyd, line 1659: I'm sorry, but I still don't understand the fundamental problem that this PR fixes with #5247 .

I see this PR only adding semantic error when a non-parameter start/end/increment is passed for an ImpliedDoLoop in a data statement. Cause, even I wasn't replacing start/end/increment with `expr_value` in-place of the `ImpliedDoLoop` either.

Could you give me something concrete that this PR *helps fix*, which the other PR was doing incorrectly? (let's say: doing a fortran dump or something else maybe?)

gxyd: The difference I see is, mine was:
```cpp
ASR::expr_t* loop_start_expr = ASRUtils::expr_value(implied_do_loop->m_start);
```

the change here is:
```cpp
ASR::expr_t* loop_start_expr = implied_do_loop->m_start;
ASR::expr_t* start_expr_value = ASRUtils::expr_value(loop_start_expr);
```
it seems like, you are doing the same thing that I was, but just in two steps.

Pranavchiku: I agree and I know this / that PR was not affecting the node in any way unlike what we discussed in PR. So, this PR is just intended to isolate two things and throw error messages :)

certik (approved):
Seems fine to merge. More work is needed to keep the original expression in ASR, which we can do later.

Pranavchiku:
> More work is needed to keep the original expression in ASR, which we can do later.

Right now we just skip that and add bunch of assignments as shown below:

```console
% cat b.f90 
program data_13
    implicit none
    integer, parameter :: k = 3
    integer :: i, iarx(k), iary(2*k)
    data(iarx(i), i=1, k) / 1, 2, 3 /
end program data_13
% lfortran b.f90 --show-fortran
program data_13
implicit none
integer(4) :: i
integer(4), dimension(3) :: iarx
integer(4), dimension(6) :: iary
integer(4), parameter :: k = 3
iarx(1) = 1
iarx(2) = 2
iarx(3) = 3
end program data_13
```

So here doing `expr_value(loop_start_expr)` won't hurt us. Yes, we need to find a way to unroll this after generating ASR so that we can generate correct code in ASR to Fortran.

PR #5300, opened 2024-11-08
Error resilient array constructor with asterisk in type spec

parth121101:
Towards: https://github.com/lfortran/lfortran/issues/5036


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4225,8 +4225,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             ASR::String_t* char_type_spec = ASR::down_cast<ASR::String_t>(type);
             // e.g. [character(len=*) :: "a", "apple"], this isn't allowed
             if (char_type_spec->m_len == -1) {
-                throw SemanticError("Type-spec cannot contain an asterisk for a type "
-                    "parameter", char_type_spec->base.base.loc);
+                diag.add(Diagnostic("Type-spec cannot contain an asterisk for a type "
+                    "parameter", Level::Error, Stage::Semantic, {}));
+                throw SemanticAbort();
```
Pranavchiku, line 4230:
```suggestion
                if (!compiler_options.continue_compilation) throw SemanticAbort();
```

tests/errors/array_constructor_with_asterisk_in_type_spec_cc.f90
```diff
@@ -0,0 +1,4 @@
+program array_constructor_with_asterisk_in_type_spec_cc
+    print *, [character(*) :: "a", "b", "ball", "cat"]
```
Pranavchiku, line 2:
```suggestion
    print *, [character(*) :: "a", "b", "ball", "cat"]
    print *, [character(*) :: "a2", "b2", "ball2", "cat2"]
```
So that we get 2 error and be sure that it is working,

Pranavchiku (changes requested):
left changes above

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: Also you'll have to add `try expect` block where you are calling `check_if_type_spec_has_asterisk`

tests/errors/array_constructor_with_asterisk_in_type_spec_cc.f90
parth121101: Sure @Pranavchiku , this will ensure that it is working as expected. Incorporated this comment.

src/lfortran/semantics/ast_common_visitor.h
parth121101: @Pranavchiku I think check for continue compilation flag should be made will calling the function. As suggested I added the function call inside try catch block and inside this I am checking for the continue compilation flag.

Please let me know if I am incorrect at any of the place.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4259,8 +4260,13 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
         // (e.g. "[real :: 1, 2, 3, 4]")
         bool is_type_spec_ommitted { type == nullptr };
-        check_if_type_spec_has_asterisk(type);
-
+        try {
+            check_if_type_spec_has_asterisk(type);
+        } catch (const SemanticError &e) {
+            if ( compiler_options.continue_compilation ) diag.add(e.d);
+            else throw e;
+        }
```
Pranavchiku, line 4268: you need to catch `SemanticAbort` as well because you are throwing that,

tests/reference/run-array_constructor_with_asterisk_in_type_spec_cc-fc032e2.stderr
```diff
@@ -0,0 +1,3 @@
+semantic error: Type-spec cannot contain an asterisk for a type parameter
+
```
Pranavchiku, line 2: missing:

```console
 --> tests/errors/array_constructor_with_asterisk_in_type_spec.f90:2:14
  |
2 |     print *, [character(*) :: "a", "b", "ball", "cat"]
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: In this routine it is fine, but general idea is to stop only when it is a visitor for example if you catch an error in `visit_Assignment`, you'll simply skip this and `throw SemanticAbort()`, but for any other routine you let it run through, because that may generate other possible errors which we'll report.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3277,6 +3277,9 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             } catch (const SemanticError &e) {
                 if ( compiler_options.continue_compilation ) diag.add(e.d);
                 else throw e;
+            } catch (const SemanticAbort &a) {
+                if ( !compiler_options.continue_compilation ) throw a;
+                return;
```
parth121101, line 3282: Returning is required, this will ensure that once an error is thrown we are out of the function and the code wont break at line number : 3285 due to null value of tmp.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4260,7 +4262,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         // (e.g. "[real :: 1, 2, 3, 4]")
         bool is_type_spec_ommitted { type == nullptr };
         check_if_type_spec_has_asterisk(type);
-
+
```
certik, line 4265: Don't add random white space.

certik:
I think it's good, let's merge it. We can send more PRs if needed.

Pranavchiku (commented):
This is good, thanks for merging!

Issue #5305, opened 2024-11-08
Non-standard character-context continuation behavior; do we care?


pjh40:
While testing a fix for issue #678, I realized that the `parse_string` in `src/lfortran/parser/parser.cpp` exhibits a variety of non-standard behaviors regarding line continuation in both fixed and free format.  In fixed format, if the parser has not found a single closing quote before the newline, it will simply move to column 7 of the next line, regardless of the state of any of the control columns 1-6, including the lack of a continuation character in column 6.  GCC flags this as an error.

In free format, both LFortran and GCC will compile the program
```f90
program foo
  print  *, "this &
     is a string continuation test"
end program foo
```
by continuing the string to the next non-blank character on the next line ('I').  The standard requires that there is some later continuation line that begins with `&`, and that parsing of the string will continue immediately after that:
 ```f90
program foo
  print  *, "this &
             &is a string continuation test"
end program foo
```

So, the question is, "Do we care?"  I know that a lot of Fortran coders have no idea what the string continuation rules actually are, so I imagine that there is a lot of non-conforming code out there.  At worst, the current implementation accepts some non-conforming code, but it basically "does the right thing", and fixing it is a pretty big pain. 

Thoughts?

Issue #678, opened 2022-08-26
Discard columns 72+ in fixed form


konradha:
The fixed form parser currently just accepts the rest of the line after the 72nd column -- all characters after that should be discarded though.

konradha:
Needed for #480

PR #5306, opened 2024-11-08
Enforce 72-column limit on fixed-form inputs

pjh40:
This changes the fixed-form parser to completely ignore inputs beyond  column 72, which addresses issue #678.  This PR also adjusts an existing fixed-form test (common2) to conform to the 72 column restriction, and adds a new test case to ensure c72 behavior. 

Note that this PR leaves the string-parsing behavior discussed in issue #5305 unchanged: those issues will be addressed in another PR.   Ideas for future improvement include:
* add a command-line option to adjust the fixed-form line length
* warn when input is being discarded (a la GCC -Wall).
* consider the 2018 limit of 132 characters for free-form lines (removed in 2023).

### DIFF
--- a/src/lfortran/parser/parser.cpp
+++ b/src/lfortran/parser/parser.cpp
@@ -321,27 +321,52 @@ void skip_rest_of_line(const std::string &s, size_t &pos)
 
 // Parses string, including possible continuation lines
 void parse_string(std::string &out, const std::string &s, size_t &pos,
-    bool fixed_form)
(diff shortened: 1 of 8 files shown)


Pranavchiku (approved):
This looks good, thanks for doing it! I'll wait for one more approval before merging this.

certik (approved):
Indeed this looks great, thank you so much for implementing it. Everything looks clean to me.

PR #5307, opened 2024-11-09
feat: error resilient arithmetic if

parth121101:
Towards: https://github.com/lfortran/lfortran/issues/5036

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3350,7 +3350,11 @@ public:
         bool is_int  = ASR::is_a<ASR::Integer_t>(*test_int_type);
         bool is_real = ASR::is_a<ASR::Real_t>(*test_int_type);
         if (!is_int && !is_real) {
-            throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
+            diag.add(diag::Diagnostic(
+                "Arithmetic if (x) requires an integer or real for `x`",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("", {x.base.base.loc})}));
+           if (!compiler_options.continue_compilation) throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
         }
         ASR::expr_t *test_lt, *test_gt;
         int kind = ASRUtils::extract_kind_from_ttype_t(test_int_type);
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3353,8 +3353,8 @@ public:
             diag.add(diag::Diagnostic(
                 "Arithmetic if (x) requires an integer or real for `x`",
                 diag::Level::Error, diag::Stage::Semantic, {
-                    diag::Label("", {x.base.base.loc})}));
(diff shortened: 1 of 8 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3347,7 +3347,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         bool is_int  = ASR::is_a<ASR::Integer_t>(*test_int_type);
         bool is_real = ASR::is_a<ASR::Real_t>(*test_int_type);
         if (!is_int && !is_real) {
-            throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
+            diag.add(diag::Diagnostic(
+                "Arithmetic if (x) requires an integer or real for `x`",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("", {x.base.base.loc})}));
+           if (!compiler_options.continue_compilation) throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
```
certik, line 3354: This logic reports it twice, as you can see it below.

Instead, the logic should be to throw a SemanticAbort here if continue_compilation is false. If it is true, I think we should not continue with this function. So I would just throw it in every case:

```suggestion
           throw SemanticAbort();
```

In general, we only should consult `continue_compilation` in the statements section, not in every node.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3347,7 +3347,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         bool is_int  = ASR::is_a<ASR::Integer_t>(*test_int_type);
         bool is_real = ASR::is_a<ASR::Real_t>(*test_int_type);
         if (!is_int && !is_real) {
-            throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
+            diag.add(diag::Diagnostic(
+                "Arithmetic if (x) requires an integer or real for `x`",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("", {x.base.base.loc})}));
+            throw SemanticAbort();
```
Pranavchiku, line 3354: Where are we catching it?

certik: In the transform_statements, the place that handles all statements.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3347,7 +3347,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         bool is_int  = ASR::is_a<ASR::Integer_t>(*test_int_type);
         bool is_real = ASR::is_a<ASR::Real_t>(*test_int_type);
         if (!is_int && !is_real) {
-            throw SemanticError("Arithmetic if (x) requires an integer or real for `x`", test_int->base.loc);
+            diag.add(diag::Diagnostic(
+                "Arithmetic if (x) requires an integer or real for `x`",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("", {x.base.base.loc})}));
```
certik, line 3353: You can see from the error message that this location is too broad. Use the original `test_int->base.loc`.

Pranavchiku (approved):

PR #5310, opened 2024-11-09
ref: frontend to not create Enum_t instead add symbols in parent_scope

### LINKED ISSUE #5290: Simplify resolve_variable() in common visitor
Currently it contains this code:
```c++
        // Check for the variable in enum symtab, if enum is declared
        bool from_enum = false;
        {
            int i = 1;
            std::string enum_name = "_nameless_enum";
            std::string selected_enum_name = "";
            while (v == nullptr && scope->resolve_symbol(std::to_string(i) +
                    enum_name) != nullptr) {
                ASR::symbol_t *enum_s = scope->resolve_symbol(std::to_string(i)
                    + enum_name);
                ASR::Enum_t *enum_ = ASR::down_cast<ASR::Enum_t>(enum_s);
                v = enum_->m_symtab->get_symbol(var_name);
                from_enum = v;
                selected_enum_name = std::to_string(i) + enum_name;
                i++;
            }
            if( v && from_enum ) {
                std::string var_name_ = current_scope->get_unique_name(std::string("1_") + var_name);
                v = ASR::down_cast<ASR::symbol_t>(ASR::make_ExternalSymbol_t(al, loc, current_scope,
                        s2c(al, var_name_), v, s2c(al, selected_enum_name), nullptr, 0, s2c(al, var_name),
                        ASR::accessType::Public));
                current_scope->add_symbol(var_name_, v);
            }
        }
        if (var_name == "c_null_ptr") {
            // Check if c_null_ptr is imported from iso_c_binding (intrinsic module)
            if (v && ASR::is_a<ASR::ExternalSymbol_t>(*v)) {
                std::string m_name = ASR::down_cast<ASR::ExternalSymbol_t>(v)->m_module_name;
                if (startswith(m_name, "lfortran_intrinsic")) {
                    ASR::ttype_t *type_ = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
                    tmp = ASR::make_PointerNullConstant_t(al, loc, type_);
                    return tmp;
                }

            }
        }
```
This is very slow. We should simply lookup the symbol, not doing some iterative while loop string concatenation lookup for the first part.

The second part about `"c_null_ptr"` also I think should just be in the symbol table.

Pranavchiku:
Fixes #5290

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1109,33 +1109,6 @@ public:
             }
         }
 
-        // TODO: the code below should be simplified:
-        // https://github.com/lfortran/lfortran/issues/5290
-
-        // Check for the variable in enum symtab, if enum is declared
-        bool from_enum = false;
-        {
-            int i = 1;
-            std::string enum_name = "_nameless_enum";
-            std::string selected_enum_name = "";
-            while (v == nullptr && scope->resolve_symbol(std::to_string(i) +
-                    enum_name) != nullptr) {
-                ASR::symbol_t *enum_s = scope->resolve_symbol(std::to_string(i)
-                    + enum_name);
-                ASR::Enum_t *enum_ = ASR::down_cast<ASR::Enum_t>(enum_s);
-                v = enum_->m_symtab->get_symbol(var_name);
-                from_enum = v;
-                selected_enum_name = std::to_string(i) + enum_name;
-                i++;
-            }
-            if( v && from_enum ) {
-                std::string var_name_ = current_scope->get_unique_name(std::string("1_") + var_name);
-                v = ASR::down_cast<ASR::symbol_t>(ASR::make_ExternalSymbol_t(al, loc, current_scope,
-                        s2c(al, var_name_), v, s2c(al, selected_enum_name), nullptr, 0, s2c(al, var_name),
-                        ASR::accessType::Public));
-                current_scope->add_symbol(var_name_, v);
-            }
-        }
         if (var_name == "c_null_ptr") {
             // Check if c_null_ptr is imported from iso_c_binding (intrinsic module)
             if (v && ASR::is_a<ASR::ExternalSymbol_t>(*v)) {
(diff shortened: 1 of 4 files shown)


Pranavchiku (commented):
We lose the information we need to reconstruct Fortran code, see the following:

```console
% cat a.f90     
program enum_01
    implicit none
    enum, bind(c)
        enumerator :: red, blue = 3
        enumerator yellow
    end enum
    enum, bind(c)
        enumerator :: green = 5
    end enum
    print *, red, blue, yellow, green
end program enum_01
% lfortran a.f90 --show-fortran
program enum_01
implicit none
integer(4), parameter :: blue = 3
integer(4), parameter :: green = 5
integer(4), parameter :: red = 0
integer(4), parameter :: yellow = 4
print *, red, blue, yellow, green
end program enum_01
```

certik (approved):
It's weird that it wasn't needed at all. But we test enums in `enum_*.f90` and they seem to work.

certik:
I see. Hold on on merging this yet.

certik:
The current implementation is incorrect, since the whole point of having a symbol table as a hashtable is immediate lookup for any symbol. We specifically do not loop over all of ASR to try to find the symbol. So this PR fixes this wrong design by removing it.

However, we still have the problem how to actually implement the numerations. It's a problem that these enumerators are global, but also part of the anonymous enumeration. The current design is incorrect anyway:
```
                            1_blue:
                                (ExternalSymbol
                                    2
                                    1_blue
                                    3 blue
                                    1_nameless_enum
                                    []
                                    blue
                                    Public
                                ),
                            1_green:
                                (ExternalSymbol
                                    2
                                    1_green
                                    4 green
                                    2_nameless_enum
                                    []
                                    green
                                    Public
                                ),
                            1_nameless_enum:
                                (Enum
                                    (SymbolTable
                                        3
                                        {
                                            blue:
                                                (Variable
```
It's using an ExternalSymbol to go into the enumerator, but ExternalSymbol should only be used for another module.

So I would say let's merge this, but open up an issue to properly design this.

certik:
Here is the issue for a proper fix: https://github.com/lfortran/lfortran/issues/5314.

Issue #480, opened 2022-08-15
NIST F77 validation suite, updated for Fortran 95


certik:
*Original issue*: https://gitlab.com/lfortran/lfortran/-/issues/627

Gary Klimowicz has just uploaded [FCVS_95](https://github.com/gklimowicz/FCVS_95) to GitHub, which is the NIST F77 validation suite, updated for Fortran 95 (still using fixed source form). It could be part of the test suite for LFortran, either directly or after translating to free source form.

konradha:
This is a great benchmark we should push to support.

PR #5318, opened 2024-11-09
Adjust DATA statement parsing

pjh40:
Trying to run the FCVS_95 validation suite in #480, I found that some `DATA` statement constructs were not supported in LFortran, such as positive signed integers and complex:
```f90
  data ivon06,ivon07,ivon08,ivon09,ivon10 /+3,+76,+587,+9999,+21111/
  data axva,axvb /(-234.23, 3), (+2, -3.0)/
```
While I was working on fixing those, I noted that the `data_stmt_repeat` grammar rule was too permissive in some cases, and that the `data_stmt_constant` was not supporting _scalar-constant-subobject_, so I tried to fix those as well.  Added a new test as well.

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -363,6 +363,8 @@ void yyerror(YYLTYPE *yyloc, LCompilers::LFortran::Parser &p,
 
 // Nonterminal tokens
 
+%type <ast> designator
+%type <ast> signed_numeric_constant
 %type <ast> expr
 %type <vec_ast> expr_list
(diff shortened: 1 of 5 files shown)


certik:
The macOS error is unrelated: https://github.com/lfortran/lfortran/issues/5321.

certik (approved):
I think the fix is good, thanks for implementing it.

I am going to merge https://github.com/lfortran/lfortran/pull/5322 first and then I'll rebase this PR on top, which should hopefully fix the macOS failure.

PR #5320, opened 2024-11-09
Chore : Add runtime error when allocating already allocated string

assem2002:


certik (approved):
I think this is good, thanks for the PR! It behaves just like GFortran.

certik:
We need to merge #5322 first.

PR #5322, opened 2024-11-09
CI: set the packages to specific versions

### LINKED ISSUE #5321: CI: macOS test sometimes fails with `--debug-with-line-column`
The macOS test sometimes fails with:
```
[  0%] Building Fortran object CMakeFiles/program_cmake_01.dir/program_cmake_01.f90.o
semantic error: The `emit_debug_line_column` is not enabled; please use the `--debug-with-line-column` option to get the correct location information
```

https://github.com/lfortran/lfortran/actions/runs/11760037574/job/32760145302?pr=5318

This error comes from using `-g` with lfortran:
```console
$ lfortran integration_tests/program_cmake_01.f90 -g
semantic error: The `emit_debug_line_column` is not enabled; please use the `--debug-with-line-column` option to get the correct location information


```
And not adding the `--debug-with-line-column`:
```console
$ lfortran integration_tests/program_cmake_01.f90 -g --debug-with-line-column
Hello
```
But the issue is that we should not be testing with `-g`, as not all tests pass yet with it.

The bug is likely that the `cmake` version being used might be too new and it adds `-g` by default to LFortran or something like that.

certik:
These versions are known to work. Install a specific cmake version, which should fix #5321.

Fixes #5321.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -18,7 +18,6 @@ dependencies:
   - zlib=1.2.13
   - rapidjson=1.1.0
   - zstd-static=1.5.5
-  - pandoc=3.1.13
   - cmake=3.29.1
 #  - bison=3.4  [not win]
 #  - m2-bison=3.4  [win]
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -3,21 +3,22 @@ channels:
   - conda-forge
 dependencies:
   - llvmdev=11.1.0
-  - toml
-  - pytest
-  - jupyter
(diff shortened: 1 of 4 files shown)


certik:
It looks like it will go through this time. I had to leave `zlib` unspecified, since there is no single version that works for all LLVM 10-19 versions. But zlib usually doesn't cause issues, so I think it's fine to leave it unspecified. The alternative is that we would need to set it together with LLVM, which would be extra work to do. So let's leave it unspecified.

PR #5323, opened 2024-11-10
CI: speedup dftatom tests

certik:
Several changes were made:

* Use the upstream master unmodified
* In LLVM 10-19 only test one test, not the full testsuite
* In 3rd party tests only test simplifier without `--fast`.


certik:
It seems to work now, it reduces the dftatom times by about 6 minutes each.

PR #5328, opened 2024-11-10
feat: error resilient array with different character length

parth121101:
Can't really figure out what has just happened. Why test references of other files updated. 
@Pranavchiku can you help me here.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -121,14 +121,15 @@ public:
                 } else {
                     tmp = nullptr;
                     tmp_vec.clear();
+                    return;
                 }
             } catch (const SemanticError &e) {
                 if (!compiler_options.continue_compilation) {
                     throw e;
                 } else {
-                    diag.add(e.d);
                     tmp = nullptr;
                     tmp_vec.clear();
+                    return;
                 }
             }
             if (tmp != nullptr) {
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -121,15 +121,14 @@ public:
                 } else {
                     tmp = nullptr;
                     tmp_vec.clear();
-                    return;
                 }
             } catch (const SemanticError &e) {
                 if (!compiler_options.continue_compilation) {
                     throw e;
                 } else {
+                    diag.add(e.d);
                     tmp = nullptr;
                     tmp_vec.clear();
-                    return;
                 }
             }
             if (tmp != nullptr) {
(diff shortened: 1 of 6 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -125,6 +125,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 } else {
                     tmp = nullptr;
                     tmp_vec.clear();
+                    return;
```
Pranavchiku, line 128: because of these I suspect

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -133,6 +134,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     diag.add(e.d);
                     tmp = nullptr;
                     tmp_vec.clear();
+                    return;
```
Pranavchiku, line 137: and this

parth121101: Thanks for the help here. It worked

tests/reference/run-array_constructor_with_different_char_lengths_cc-746cb03.stdout
```diff
@@ -0,0 +1,3 @@
+ball
+ball1
```
Pranavchiku, line 2: this is interesting. Why did it print these?

Pranavchiku (approved):
I think this is good, let's wait for one more approval.

tests/reference/run-array_constructor_with_different_char_lengths_cc-746cb03.stdout
HarshitaKalani: +1

HarshitaKalani (approved):
I'm also curious about this:https://github.com/lfortran/lfortran/pull/5328#discussion_r1835649736

Otherwise this looks good to me too.

certik:
This PR doesn't add any new fix or implementation. It only adds tests. Do we need more tests right now?

It does expose a bug though (the spurious prints), so that has to be fixed separately.

Right now let's only add tests to test new functionality, not current functionality unless we have to, since it slows down our testing runs. So we should spare our "testing budget" for new features. Once we reach beta, then let's add thorough tests for every compiler error (but not as one test per error, but probably like 100 errors per file, all reported at once, for testing performance reasons).

parth121101:
> This PR doesn't add any new fix or implementation. It only adds tests. Do we need more tests right now?
> 
> It does expose a bug though (the spurious prints), so that has to be fixed separately.
> 
> Right now let's only add tests to test new functionality, not current functionality unless we have to, since it slows down our testing runs. So we should spare our "testing budget" for new features. Once we reach beta, then let's add thorough tests for every compiler error (but not as one test per error, but probably like 100 errors per file, all reported at once, for testing performance reasons).

Yes @certik . We will add a single test file to test more errors. And regarding the bug present here, I would work on fixing it in this PR itself.

certik:
Yes. In fact once we have one file with errors, then I think you can add more error messages tests (like in this PR), as the overhead should be minimal.

Pranavchiku (approved):

PR #5329, opened 2024-11-10
revert: creating `Enum_t` node and link variables via `ExternalSymbol`

### LINKED ISSUE #5314: Design enumerations properly
See https://github.com/lfortran/lfortran/pull/5310#issuecomment-2466309316.

The big problem is that enumerators are both part of the enumeration as well as global.

The old design was almost correct. We need to use either ExternalSymbol or an equivalent feature to store enumerators globally, but make them point inside the enumeration. So we use regular lookups in the symbol table to find them.

A relatively clean design could be:

* Introduce EnumeratorLink symbol, which would point to the actual enumeration. Then we look it up as usual, and use in expressions.
* We have a choice:
  * Either we always go into the enumeration and pull the actual enumerator
  * Or we use EnumeratorLink in expressions via Var. But it would be quite painful I think, so probably always following up might not be a bad idea; this would happen at the same place where the old code in #5310 was, but we simply check if the symbol is EnumeratorLink, and if so, follow up through and use the enumerator from enumeration's symbol table.

When an enumeration is declared in Fortran, then the frontend (AST->ASR) will bring all enumerators into the current scope via EnumeratorLink. In LPython and other frontends we usually don't want to do that and let users reference the enumeration explicitly, but since Fortran is designed this way, then we can implement this global feature via EnumeratorLink.

Pranavchiku:
Fixes #5314. This is the same design mentioned at #5314 but due to https://github.com/lfortran/lfortran/issues/5314#issuecomment-2466649139 we cannot pull enum variable. This asks us to create an `EnumeratorLink` node, doing precisely the same thing as `ExternalSymbol` but we need to write exact same code and handler for `EnumeratorLink` as we do for `ExternalSymbol` so I think using `ExternalSymbol` is a way here to avoid duplication.

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -3430,6 +3430,23 @@ public:
     }
 
     void visit_Enum(const AST::Enum_t &x) {
+        SymbolTable *parent_scope = current_scope;
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+        std::string sym_name = "_nameless_enum";
+        {
+            int i = 1;
+            while (parent_scope->get_symbol(std::to_string(i) +
+                    sym_name) != nullptr) {
+                i++;
+            }
+            sym_name = std::to_string(i) + sym_name;
+        }
+        Vec<char *> m_members;
+        m_members.reserve(al, 4);
+        ASR::ttype_t *type = ASRUtils::TYPE(ASR::make_Integer_t(al,
+            x.base.base.loc, compiler_options.po.default_integer_kind));
+
+        ASR::abiType abi_type = ASR::abiType::BindC;
         if ( x.n_attr == 1 ) {
             if ( AST::is_a<AST::AttrBind_t>(*x.m_attr[0]) ) {
                 AST::Bind_t *bind = AST::down_cast<AST::Bind_t>(
@@ -3457,6 +3474,27 @@ public:
         for ( size_t i = 0; i < x.n_items; i++ ) {
             this->visit_unit_decl2(*x.m_items[i]);
         }
+
+        for( auto sym: current_scope->get_scope() ) {
+            ASR::Variable_t* member_var = ASR::down_cast<
+                ASR::Variable_t>(sym.second);
+            m_members.push_back(al, member_var->m_name);
+        }
+
+        ASR::enumtypeType enum_value_type = ASR::enumtypeType::IntegerConsecutiveFromZero;
+        ASRUtils::set_enum_value_type(enum_value_type, current_scope);
+
+        tmp = ASR::make_Enum_t(al, x.base.base.loc, current_scope,
+            s2c(al, sym_name), nullptr, 0, m_members.p, m_members.n, abi_type,
+            dflt_access, enum_value_type, type, nullptr);
+        parent_scope->add_symbol(sym_name, ASR::down_cast<ASR::symbol_t>(tmp));
+        for (auto it: current_scope->get_scope()) {
+            ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(it.second);
+            parent_scope->add_symbol(var->m_name, ASR::down_cast<ASR::symbol_t>(ASR::make_ExternalSymbol_t(al,
+                        var->base.base.loc, parent_scope, s2c(al, var->m_name), it.second,
+                        s2c(al, sym_name), nullptr, 0, var->m_name, var->m_access)));
+        }
+        current_scope = parent_scope;
     }
 
 };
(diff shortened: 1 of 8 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3457,6 +3474,27 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        tmp = ASR::make_Enum_t(al, x.base.base.loc, current_scope,
+            s2c(al, sym_name), nullptr, 0, m_members.p, m_members.n, abi_type,
+            dflt_access, enum_value_type, type, nullptr);
+        parent_scope->add_symbol(sym_name, ASR::down_cast<ASR::symbol_t>(tmp));
+        for (auto it: current_scope->get_scope()) {
+            ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(it.second);
+            parent_scope->add_symbol(var->m_name, ASR::down_cast<ASR::symbol_t>(ASR::make_ExternalSymbol_t(al,
+                        var->base.base.loc, parent_scope, s2c(al, var->m_name), it.second,
+                        s2c(al, sym_name), nullptr, 0, var->m_name, var->m_access)));
+        }
```
Pranavchiku, line 3496: This is the only new part, rest is copied from deleted changes. See, we now don't need the `resolve_variable()` logic.

certik:
Let's figure out how to test this. I think either ASR test for enumerations, possibly also ASR->Fortran in tests/, driving the point home that unless we do this PR, we can't represent the enumeration anymore.

Right now as it is, this PR adds code that is not ideal, doesn't change any tests, so it seems like a net negative. But we know (or hope) it is not, due to improved ASR. So let's test it.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3430,6 +3430,23 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        SymbolTable *parent_scope = current_scope;
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+        std::string sym_name = "_nameless_enum";
+        {
+            int i = 1;
+            while (parent_scope->get_symbol(std::to_string(i) +
+                    sym_name) != nullptr) {
+                i++;
+            }
+            sym_name = std::to_string(i) + sym_name;
```
certik, line 3442: I don't like this. Is the idea here to create a unique name that cannot be used by the user?

If so, let's create a function for this in ASR utils. In fact I think we already have a function for this.

src/lfortran/semantics/ast_symboltable_visitor.cpp
certik: Add a comment here what this does:
```
// Expose all enumerators into the parent scope as ExternalSymbols pointing into the enumeration, which is the semantics of Fortran enums
// That way `resolve_variable()` can resolve the automatically.
// In ASR->Fortran we do not create any Fortran code for these ExternalSymbols, since they are implicit. But in ASR we need to represent them explicitly.

**Comment by certik** [comment]
Thanks @Pranavchiku I think the general idea in this PR is correct, and it's ok to use ExternalSymbols for this, but in that case let's document it. And also, let's start to fix some of the technical debt:

* Add ASR verify check that ensures that whatever ExternalSymbol points to is either separate module, or enum (in this PR), nothing else.

File: src/lfortran/semantics/ast_symboltable_visitor.cpp (hunk 2 above)
Comment (Pranavchiku) [reply src/lfortran/semantics/ast_symboltable_visitor.cpp:3442]: Yes, I see, we can now use that function because we don’t rely on the pattern to lookup the enum variables.

**Comment by Pranavchiku** [comment]
Makes sense, let me also enable Fortran backend test for enum, that will be a minor fix that way we can be sure that this PR is in right direction.

File: src/lfortran/semantics/ast_symboltable_visitor.cpp (hunk 1 above)
Comment (certik) [reply src/lfortran/semantics/ast_symboltable_visitor.cpp:3489]: @Pranavchiku can you add this comment?

**Comment by certik** [comment]
Can you add a reference test for the Fortran output, ensuring that previously we couldn't see the enum, but now we do?

**Review by certik — APPROVED** [review]
Looks great I think!

Make sure to add this comment (or equivalent): https://github.com/lfortran/lfortran/pull/5329/files#r1835712738.


Then we can merge.

File: src/lfortran/semantics/ast_symboltable_visitor.cpp [hunk 3]
```diff
@@ -3481,6 +3481,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             s2c(al, sym_name), nullptr, 0, m_members.p, m_members.n, abi_type,
             dflt_access, enum_value_type, type, nullptr);
         parent_scope->add_symbol(sym_name, ASR::down_cast<ASR::symbol_t>(tmp));
+        // Expose all enumerators into the parent scope as ExternalSymbols pointing into the enumeration, which is the semantics of Fortran enums
+        // That way `resolve_variable()` can resolve the automatically.
```
certik, line 3485: the -> them. Add a dot at the end of the previous line.

Pranavchiku: Done through https://github.com/lfortran/lfortran/pull/5394

PR #5330, opened 2024-11-10
feat: error resilient dim in size intrinsic

HarshitaKalani:
Towards: #5036

### DIFF
--- a/tests/reference/asr-array_size_02-94fb4d3.stderr
+++ b/tests/reference/asr-array_size_02-94fb4d3.stderr
@@ -1,4 +1,4 @@
-semantic error: dim has already been specified as a positional/keyword argument to size.
+semantic error: dim has already been specified as a positional/keyword argument to size
  --> tests/errors/array_size_02.f90:5:14
   |
 5 |     size_a = size(a, 1, dim=1)
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -5358,10 +5358,13 @@ public:
                 if( !raise_error ) {
                     return false;
                 }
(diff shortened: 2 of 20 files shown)


tests/reference/asr-array_size_02-94fb4d3.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: dim has already been specified as a positional/keyword argument to size.
+semantic error: dim has already been specified as a positional/keyword argument to size
```
Pranavchiku, line 1: tada! you missed a period :)

Pranavchiku (approved):
Fix the missing `.` issue and this PR is good to merge.

tests/reference/asr-array_size_02-94fb4d3.stderr
HarshitaKalani: I removed that intentionally. Ending with a period is making the message feel like a sentence rather than a quick note on an issue. So I felt it's better to remove that. What do you feel about this?

HarshitaKalani: And we're generally not using a period after an error message, so that's an another reason.

Pranavchiku: I think it is fine then

parth121101 (approved):

tests/reference/asr-array_size_02-94fb4d3.stderr
certik: Yes, probably without a period is better. We should write guidelines how to write good error messages. Until then let's do our best to be consistent.

certik (approved):

PR #5331, opened 2024-11-10
feat: error resilient array with different kind

parth121101:
What I am doing here is I am trying to add the error in diagnostics and then throwing the semantic abort.
But while compiling the error file ```tests/errors/array_constructor_with_different_kind_cc.f90``` using ```continue-compilation``` option I am getting some errors.
I am investigating the stack trace. If anyone knows the problem please let me know.


kmr-srbh:
This gives the correct expected output on my machine (Ubuntu). Could you please try building again and pushing the new reference file?

tests/tests.toml
```diff
@@ -4017,6 +4017,10 @@ asr = true
 filename = "errors/array_constructor_with_different_kind.f90"
 asr = true
 
+[[test]]
+filename = "errors/array_constructor_with_different_kind_cc.f90"
+continue_compilation = true
```
certik, line 4022: Instead of adding a new short file, let's start adding all these tests into just one file, that will eventually have 100 or so error tests.

parth121101: Sure

parth121101:
@certik @Pranavchiku can we merge this PR ?

tests/errors/continue_compilation_2.f90
```diff
@@ -83,5 +84,14 @@ program continue_compilation_2
     !atomic_01
     call atomic_add (atom[1], this_image())
     call atomic_add (atom[2], this_image())
+    !array_constructor_with_asterisk_in_type_spec
```
Pranavchiku, line 87: remove comments
```suggestion
```

parth121101: I have added comments to make a note of the tests that are covered. Once all tests are covered I will remove them.

certik (approved):
I think that this is fine.

What is the meaning of the comments? I don't understand how to interpret them.

Issue #3770, opened 2024-04-02
ASR: idea: add explicit nodes to free non-primitive types


certik:
We have "primitive" types like Integer, Real, Complex.

And non-primitive types: Array, List, Dict, Set, String, Class, Struct

The non-primitive types require some kind of deallocation or freeing, even if they are not "allocatable".

The Allocatable type's semantic is that the entity is not even allocated, and one must call "Allocate" to create it. Non-allocatable types are always created (whether on stack or a heap, that's an implementation detail).

However, non-primitive types have to be freed if the backend chooses to implement them on a heap.

The logic of freeing is complicated (many ways to escape from a block), so it should be explicit in ASR and checked by ASR verify.

Another idea: we can have a general "OutOfScope" ASR node that will list all variables that go out of scope at this point, including real and integer. The backend will then choose which of these variables should be freed and how.

It seems that Allocatable type (primitive or non-primitive) has exactly the same handling as a non-allocatable non-primitive type, except that there must be an explicit allocation and the deallocation must check at runtime if the type was allocated.

certik:
To make progress on this, minimal changes that LFortran needs (LPython will need more) are:

* All allocatable entities need a Deallocate node: will deallocate if allocated
* Array, String, Class, Struct need OutOfScope node: the backend will always "free" memory if it was using heap (say for array, string, etc.)
* The OutOfScope for Class/Struct (and arrays of Class/Struct) must recursively call Deallocate on all members that are allocatable. There are multiple ways to implement it. I think we know at compile time what entity is allocatable, however it could be nested a few times, so it seems the easiest is to simply recurse at runtime and deallocate all allocatables.

certik:
It's still not clean when it comes to finalizers and the backend having to do work and logic. Here is another iteration of the above that is cleaner.

Let's create the following ASR nodes to finalize all non-primitive types: 

* `StringFinalize`
* `ArrayFinalize`
* `ListFinalize`
* `ClassFinalize`
* `StructFinalize`
* ...

And the backend must implement these to call `*Finalize` nodes for each non-allocatable member of the non-primitive type and `Deallocate` on each allocatable member, and then if the type itself was allocated on heap, to free it, otherwise do nothing. The order must be like in C++, so first the destructor is called on the child class, then the parent, then all memory is freed. In Python it's the [__del__](https://docs.python.org/3/reference/datamodel.html#object.__del__) method (delete, finalizer, destructor).

[...]

PR #5333, opened 2024-11-10
refactor: remove redundant `ImplicitDeallocate` node addition during AST -> ASR transformation

### LINKED ISSUE #2763: Duplicate deallocate in the ASR
```fortran
program main
integer, allocatable :: d(:)
allocate(d(3))
end program
```
ASR after passes:
```clojure
...
(ImplicitDeallocate
    [(Var 2 d)]
)
(ImplicitDeallocate
    [(Var 2 d)]
)
...
```

The ImplicitDeallocate nodes are inserted in `array_op` and `insert_deallocate` passes. I think we have to remove this in `array_op`.

kmr-srbh:
Cleanup as part of #3770. Fixes #2763.

We handle this explicitly in the `insert_deallocate` pass, so this is not required here.

### DIFF
--- a/tests/reference/asr-allocate_01-f3446f6.stdout
+++ b/tests/reference/asr-allocate_01-f3446f6.stdout
@@ -986,9 +986,6 @@
                                     )
                                     (ExplicitDeallocate
                                         [(Var 3 c_copy)]
-                                    )
-                                    (ImplicitDeallocate
-                                        [(Var 3 c_copy)]
                                     )]
                                     ()
                                     Public
@@ -1633,11 +1630,6 @@
                             ()
                         )]
                         []
-                    )
-                    (ImplicitDeallocate
-                        [(Var 2 a)
-                        (Var 2 b)
-                        (Var 2 c)]
                     )]
                 )
         })
--- a/tests/reference/llvm-allocate_03-495d621.stdout
+++ b/tests/reference/llvm-allocate_03-495d621.stdout
@@ -288,32 +288,9 @@ else13:                                           ; preds = %ifcont11
   br label %ifcont14
 
 ifcont14:                                         ; preds = %else13, %then12
-  %167 = load %array*, %array** %c, align 8
-  %168 = getelementptr %array, %array* %167, i32 0, i32 0
-  %169 = load i32*, i32** %168, align 8
-  %170 = ptrtoint i32* %169 to i64
-  %171 = icmp ne i64 %170, 0
-  br i1 %171, label %then15, label %else16
-
-then15:                                           ; preds = %ifcont14
-  %172 = getelementptr %array, %array* %167, i32 0, i32 0
-  %173 = load i32*, i32** %172, align 8
-  %174 = alloca i8*, align 8
-  %175 = bitcast i32* %173 to i8*
-  store i8* %175, i8** %174, align 8
-  %176 = load i8*, i8** %174, align 8
-  call void @_lfortran_free(i8* %176)
-  %177 = getelementptr %array, %array* %167, i32 0, i32 0
-  store i32* null, i32** %177, align 8
-  br label %ifcont17
-
-else16:                                           ; preds = %ifcont14
-  br label %ifcont17
-
-ifcont17:                                         ; preds = %else16, %then15
   br label %return
 
-return:                                           ; preds = %ifcont17
+return:                                           ; preds = %ifcont14
   ret i32 0
 }
 
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -1183,30 +1183,6 @@ public:
                                     stat, errmsg, source);
     }
 
-// If there are allocatable variables in the local scope it inserts an ImplicitDeallocate node
-// with their list. The ImplicitDeallocate node will deallocate them if they are allocated,
-// otherwise does nothing.
-    ASR::stmt_t* create_implicit_deallocate(const Location& loc) {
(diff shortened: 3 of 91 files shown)


tests/reference/asr-allocate_01-f3446f6.stdout
```diff
@@ -1633,11 +1630,6 @@
                             ()
                         )]
                         []
-                    )
-                    (ImplicitDeallocate
-                        [(Var 2 a)
-                        (Var 2 b)
-                        (Var 2 c)]
```
kmr-srbh, line 1640: We do not remove this, just that instead of being added here after AST -> ASR transformation, we add this in the `insert_deallocate` pass.

certik:
We have to be very careful here.

First, the `insert_deallocate` doesn't seem to insert the deallocate correctly:
```c++
        void visit_Symbol(const T& x) {
            Vec<ASR::expr_t*> to_be_deallocated;
            to_be_deallocated.reserve(al, 1);
            for( auto& itr: x.m_symtab->get_scope() ) {
                if( ASR::is_a<ASR::Variable_t>(*itr.second) &&
                    ASR::is_a<ASR::Allocatable_t>(*ASRUtils::symbol_type(itr.second)) &&
                    ASRUtils::is_array(ASRUtils::symbol_type(itr.second)) &&
                    ASRUtils::symbol_intent(itr.second) == ASRUtils::intent_local ) {
                    to_be_deallocated.push_back(al, ASRUtils::EXPR(
                        ASR::make_Var_t(al, x.base.base.loc, itr.second)));
                }
            }
            if( to_be_deallocated.size() > 0 ) {
                T& xx = const_cast<T&>(x);
                Vec<ASR::stmt_t*> body;
                body.from_pointer_n_copy(al, xx.m_body, xx.n_body);
                body.push_back(al, ASRUtils::STMT(ASR::make_ImplicitDeallocate_t(
                    al, x.base.base.loc, to_be_deallocated.p, to_be_deallocated.size())));
                xx.m_body = body.p;
                xx.n_body = body.size();
            }
        }
```
This code only appends it the given body. However, one must actually properly insert this into every place that can exit the function (if statements, etc.).

Second, we really want to enforce this using ASR verify, to check that all things are properly deallocated and that leaks cannot happen. However that means we need to do this when constructing the ASR. Alternatively, we need to create a pass and run it just to fix this. However for compilation speed reasons, we should do it in one pass.

We can have a helper pass to use for debugging or to get things done quickly, but it seems like a wrong design to have a pass that one has to call every time to "fix up the ASR".

An alternative design would be that deallocation is not enforced via ASR verify, nor required. In that case, the semantics would be that things get deallocated automatically. Then we could have a pass that inserts the deallocation nodes correctly. It might not be a bad design, since the deallocation logic would be concentrated in a pass. But we need to design this well.

assem2002:
If we go with `insert_deallocate` pass, This PR seems to be a good step towards handling the finalization issue.
If we go with ASR_verify(), So we gotta keep that for sure.

certik:
Keep the PR open for now. We'll figure out a design very soon.

tests/reference/llvm-allocate_03-495d621.stdout
```diff
@@ -288,32 +288,9 @@ else13:                                           ; preds = %ifcont11
...
-  %176 = load i8*, i8** %174, align 8
-  call void @_lfortran_free(i8* %176)
-  %177 = getelementptr %array, %array* %167, i32 0, i32 0
-  store i32* null, i32** %177, align 8
-  br label %ifcont17
-
-else16:                                           ; preds = %ifcont14
-  br label %ifcont17
-
-ifcont17:                                         ; preds = %else16, %then15
```
certik, line 313: Why were these lines removed?

kmr-srbh: This is correct actually, the code was redundant. Before this PR, we were adding the `ImplicitDeallocate` node twice, first in AST -> ASR, then in the pass. In this PR, I removed exactly the same logic from AST -> ASR as had been implemented in the pass, so we only remove redundancy, Currently, the case of `SubroutineCall` is still handled during AST -> ASR. In the next PR, I will port those changes here.

Talking of redundancy, we need to take care of the user allocatables being deallocated twice, first through `ExplicitDeallocate` node inserted during AST -> ASR, then through `ImplicitDeallocate` inserted in this pass. As `ImplicitDeallocate` does not throw an error for deallocating unallocated `allocatable` we do not get an error, but this still adds redundancy.

```fortran
program main
    implicit none
        integer, allocatable :: c(:)
        allocate(c(3))
        deallocate(c)
end program
```
```clojure
;; ASR after applying the pass: insert_deallocate
(TranslationUnit
    (SymbolTable
        1
        {
            main:
                (Program
                    (SymbolTable
                        2
                        {
                            c:
                                (Variable
                                    2
                                    c
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Allocatable
                                        (Array
                                            (Integer 4)
                                            [(()
                                            ())]
                                            DescriptorArray
                                        )
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    main
                    []
                    [(Allocate
                        [((Var 2 c)
                        (IntegerConstant 3 (Integer 4) Decimal))]
                        ()
                        ())]
                        ()
                        ()
                        ()
                    )
                    (ExplicitDeallocate
                        [(Var 2 c)]
                    )
                    (ImplicitDeallocate
                        [(Var 2 c)]
                    )]
                )
        })
    []
)
```

certik:
Ok. Go ahead and merge this.

Issue #5285, opened 2024-11-07
(PRIMA) Passing array in abs and comparing not supported


jinangshah21:
Towards #4397 

MRE : 
```fortran
program bobyqa_exmp
real :: A(2) = [2,2]
logical :: B(2)
B = abs(A) > maxval(abs(A))
end program bobyqa_exmp
```

LFortran Error : 
```console
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/mnt/c/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-LcI20x/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 2677, in main
    return main_app(argc, argv);
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 2619, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 1045, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/mnt/c/lfortran/src/lfortran/fortran_evaluator.cpp", line 377, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    compiler_options, run_fn, infile);
  File "/mnt/c/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 10317, in LCompilers::asr_to_llvm(LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LCompilers::PassManager&, LCompilers::CompilerOptions&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[...]

PR #5336, opened 2024-11-10
Fix: Handling maxval array function with comparision operator

AnshMehta1:
Fix :- #5285
Earlier , we were not visiting arguments of Array Function which can be an IntrinsicElementalFunction and thus , this PR tries to incorporate this thing in array_op.


certik:
Let's add a test that now works.

Pranavchiku (approved):
Looks good to me, thank you!

certik:
Thanks!

PR #5339, opened 2024-11-11
add mpi.f90 for declarations of subroutines in Open MPI

gxyd:


src/runtime/openmpi/mpi.f90
```diff
@@ -0,0 +1,162 @@
+! https://docs.open-mpi.org/en/v5.0.0/man-openmpi/man3
```
gxyd, line 1: should we name this as `mpi_lib.f90` or `openmpi_lib.f90` maybe?

src/runtime/openmpi/mpi.f90
```diff
@@ -0,0 +1,172 @@
...
+! NOTE: this document currently only declares openmpi
+! subroutines or constants which are used in the repository
+! [POT3D](https://github.com/predsci/POT3D)
+module mpi
+    use, intrinsic :: iso_c_binding
+    implicit none
+
+    type, bind(C) :: MPI_Datatype
+        type(C_PTR) :: c_ptr
```
gxyd, line 10: If I don't use this, it raises a warning with GFortran as:
```console
src/runtime/openmpi/mpi.f90:12:29:

   12 |     type, bind(C) :: MPI_Comm
Warning: Derived type 'mpi_comm' with BIND(C) attribute at (1) is empty, and may be inaccessible by the C companion processor
src/runtime/openmpi/mpi.f90:9:33:
```

src/runtime/openmpi/mpi.f90
```diff
@@ -0,0 +1,172 @@
...
+    type, bind(C) :: MPI_Request
+        type(C_PTR) :: c_ptr
+    end type MPI_Request
+
+    type, bind(C) :: MPI_Status
+        integer(C_INT) :: MPI_SOURCE, MPI_TAG, MPI_ERROR
+    end type MPI_Status
+
+    type(MPI_Comm) :: MPI_COMM_WORLD
+    type(MPI_Datatype), bind(C, name="MPI_INT") :: MPI_Integer
```
gxyd, line 34: this on compilation with LFortran, raises a semantic error:
```console
semantic error: Attribute type not implemented yet
  --> src/runtime/openmpi/mpi.f90:28:5
   |
28 |     type(MPI_Datatype), bind(C, name="MPI_INT") :: MPI_Integer
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

gxyd:
We need to define the (atleast) the below as well for POT3D, which aren't subroutines but objects:
```
MPI_COMM_TYPE_SHARED
mpi_info_null
mpi_real4
mpi_real8
mpi_thread_funneled
mpi_wtime
mpi_in_place
mpi_sum
mpi_proc_null
mpi_statuses_ignore
```

gxyd:
E.g.; there is no mention of `MPI_REAL4` in [Open MP documentation](https://docs.open-mpi.org/en/v5.0.x/), and I tried searching in [POT3D repo](https://github.com/search?q=repo%3Apredsci%2FPOT3D%20MPI_Real4&type=code) as well, and search results weren't of much help either to me.

src/runtime/openmpi/mpi.f90
```diff
@@ -0,0 +1,172 @@
+! https://docs.open-mpi.org/en/v5.0.0/man-openmpi/man3
+! NOTE: this document currently only declares openmpi
+! subroutines or constants which are used in the repository
+! [POT3D](https://github.com/predsci/POT3D)
+module mpi
+    use, intrinsic :: iso_c_binding
```
Pranavchiku, line 6:
```suggestion
    use, intrinsic :: iso_c_binding, only: c_ptr, c_int, c_bool
```

src/runtime/openmpi/mpi.f90
Pranavchiku, line 10:
```suggestion
        type(c_ptr) :: c_ptr
```

src/runtime/openmpi/mpi.f90
```diff
@@ -0,0 +1,172 @@
...
+    implicit none
+
+    type, bind(C) :: MPI_Datatype
+        type(C_PTR) :: c_ptr
+    end type MPI_Datatype
+
+    type, bind(C) :: MPI_Comm
+        type(C_PTR) :: c_ptr
+    end type MPI_Comm
+
```
Pranavchiku, line 16: I would keep it as simple as I can in the beginning. Else all these may create design challenges like you'll need to then handle `visit_Struct`.

src/runtime/openmpi/mpi.f90
Pranavchiku, line 34:
```suggestion
    type(c_ptr), bind(C, name="MPI_INT") :: MPI_Integer
```
I would rather do this

src/runtime/openmpi/mpi.f90
Pranavchiku: same comment elsewhere

src/runtime/openmpi/mpi.f90
Pranavchiku: This might fix:

```
semantic error: Attribute type not implemented yet
  --> /Users/runner/work/lfortran/lfortran/lfortran-0.41.0-354-gacac1d5ae/src/runtime/openmpi/mpi.f90:34:5
   |
34 |     type(MPI_Datatype), bind(C, name="MPI_INT") :: MPI_Integer
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

src/runtime/openmpi/mpi.f90
Pranavchiku: let's do `mpi_lib.f90` we can add APIs for multiple library in same place that is fine.

src/runtime/CMakeLists.txt
```diff
@@ -51,6 +53,7 @@ install(
   ${CMAKE_Fortran_MODULE_DIRECTORY}/lfortran_intrinsic_ieee_arithmetic.mod
   ${CMAKE_Fortran_MODULE_DIRECTORY}/lfortran_intrinsic_iso_c_binding.mod
   ${CMAKE_Fortran_MODULE_DIRECTORY}/omp_lib.mod
+  ${CMAKE_Fortran_MODULE_DIRECTORY}/mpi.mod
```
certik, line 56: This should not go into our runtime library. Rather, it should be a separate library from LFortran. So just add it as a regular integration_test. In the 3rd party code, you simply add this `mpi.f90` to the code, as a temporary workaround until openmpi compiles with LFortran.

src/runtime/openmpi/mpi.f90
certik: This should be called `mpi.f90`, since that is what the 3rd party code uses.

src/runtime/openmpi/mpi.f90
certik: Is it because we don't have C interop for structs yet?

certik:
Also I am not sure the functions like `    subroutine MPI_Allgather(sendbuf, sendcount, sendtype, recvbuf, recvcount, recvtype, comm, ierror) bind(C, name="MPI_Allgather")` should be `bind(c)`. We definitely need an interface to the C-based MPI, so this can be *that* interface. However, I think we also need regular Fortran functions for the `mpi.f90` with regular Fortran types, not `iso_c_binding` types. We need to check the mpich/openmpi implementation as well as the MPI standard.

To get started, I would simply take the 3rd party code and create such `mpi.f90` file that makes the 3rd party code compile with GFortran. Then to make it run in parallel, inside the `mpi.f90` make it call the C MPI and link everything together. If it turns out that we can just make `MPI_Allgather` to be a `bind(c)` into the C MPI directly, then I think we can do it. Otherwise you need to add a Fortran wrapper.

Pranavchiku:
I agree, let's only port things which are required. These implementations need a thorough background research.

gxyd:
> To get started, I would simply take the 3rd party code and create such mpi.f90 file that makes the 3rd party code compile with GFortran. 

Alright, that makes sense. Though I added the mpi.f90 in this PR with only the functions/objects needed in POT3D repo, but instead of that I'll add changes with focus to get it working with GFortran first.

Issue #5341, opened 2024-11-11
Possibly incorrect way of data initialization


gxyd:
## Minimal Reproducible Example

The below fortran program is derived from [tests/fixedform_doloop3.f](https://github.com/lfortran/lfortran/blob/main/tests/fixedform_doloop3.f), which I think has an incorrect initialization:
```f90
SUBROUTINE START1()
    IMPLICIT NONE
    INTEGER :: L
    INTEGER :: N
    REAL F, ONE=1, TWO=2
END

program main
end program main

```

the original line in the program `tests/fixedform_doloop3.f` is:
```f90
        REAL F(L), ONE=1, TWO=2
```
here: https://github.com/lfortran/lfortran/blob/main/tests/fixedform_doloop3.f#L7

## GFortran output
```console
> gfortran test.f90 -o a.out && ./a.out
test.f90:5:16:

    5 |     REAL F, ONE=1, TWO=2
      |                1
Error: Syntax error in data declaration at (1)
```

## Concern

The concern of this issue *isn't* the **error not reported** bug, but rather the fact that this allows an incorrect way to initialize variables and hence might disturb other correct fixes to the parser. With this issue the main concern is to correct the `tests/fixedform_doloop3.f`, below is a screenshot from https://bayanbox.ir/view/3820206914900746365/StatLab-Applied-Statistics-Algorithms.pdf, where the program `tests/fixedform_doloop3.f` comes from:

![Screenshot 2024-11-11 at 6 14 28 PM](https://github.com/user-attachments/assets/16ed9290-fb07-4715-9129-cfe595d7881b)

harperjf:
Is this the same bug as #3945 ?

pjh40:
As @harperjf said, I would guess that it is the same as #3945... I don't see any incorrect initialization apart from the missing `::`.

gxyd:
The underlying issue is the same here, i.e. we shouldn't allow a syntax like `REAL ONE = 1`, which wasn't necessarily the intention of this issue to fix it. The focus of this issue was to rectify the wrong test case in our test-suite where a bad syntax like `REAL ONE = 1` was there and instead use a correct syntax.

So, I think we can still keep the issue #3945 open, as even after closing this issue we don't raise an error for the incorrect syntax yet.

certik:
Everything should be fixed now.

PR #5342, opened 2024-11-11
use correct way of initialization in test case

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5341 (considering that issue is valid)

### DIFF
--- a/tests/fixedform_doloop3.f
+++ b/tests/fixedform_doloop3.f
@@ -4,7 +4,8 @@ C       ALGORITHM AS 93.1 APPL. STATIST. (1976) VOL.25, NO.1
 C
 C       GENERATES A 1,N ANSARI-BRADLEY DISTRIBUTION IN F.
 C
-      REAL F(L), ONE=1, TWO=2
+      REAL F(L), ONE, TWO
+      DATA ONE, TWO /1.0, 2.0/
       LOUT = 1 + N / 2
       DO 1 I = 1, LOUT
 1     F(I) = TWO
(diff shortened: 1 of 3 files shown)


tests/fixedform_doloop3.f
```diff
@@ -4,7 +4,8 @@ SUBROUTINE START1(N, F, L, LOUT)
 C
 C       GENERATES A 1,N ANSARI-BRADLEY DISTRIBUTION IN F.
 C
-      REAL F(L), ONE=1, TWO=2
```
Pranavchiku, line 7: My question is why `ONE=1` incorrect? because is shall be `one=1.0`? if so, then let's use that instead.

gxyd: No, the issue is with the syntax, i.e. we can't have `REAL F(L), ONE=1, TWO=2` as the correct syntax is: `REAL :: F(L), ONE=1, TWO=2` (i.e. use of `::`) or as is done currently by use of `data` statement, the only reason to use the data statement is as that's what's used in the original test case which this is being updated to.

Please also see the linked issue.

certik: We need to double check with GFortran, but I think so.

gxyd: Yes, I initially mentioned this in the linked issue.

gxyd: I'll let one of you resolve the conversation if this PR looks good.

certik (approved):
I think this is correct.

| Fixed-form | Free-form |
| --- | --- |
| integer x | integer :: x |
| integer x = 5 | integerx = 5 |
| integer :: x = 5 | integer :: x = 5 |

These are not allowed in declarations:
```
integer A(10), x = 5
integer x = 5
```

PR #4774, opened 2024-09-10
Fault Tolerant Parsing

Vipul-Cariappa:

PR #5343, opened 2024-11-11
fix: parser to throw multiple error messages only when `--continue-compilation` is enabled

### LINKED ISSUE #5332: Syntax errors now print all errors
It seems we now print all syntax errors, so we often get hundreds of error messages.

Instead, by default let's only print exactly one error message and exit. Only with `--continue-compilation` let's print all of them. We can also later control how many to print with a compiler option, but for now I think a good default is to print just one error message and exit.

Also, as a user, I don't think I want to run the code if there are errors. But I still might want to print all errors. We should figure out the CompilerOptions flags how to control this well.

Pranavchiku:
Fixes #5332. While fixing this PR, I noticed that in https://github.com/lfortran/lfortran/pull/4774 we have "not" achieved error resiliency. That PR only collects error messages and dumps at the end, not skip statements. So the AST generated is incorrect. This means we cannot generate a valid ASR. We need to fix that.


tests/reference/run-multi_error1-2ef69ef.stderr
```diff
@@ -0,0 +1,24 @@
...
+  |
+5 | print *, x +
+  |             ^ 
+
+syntax error: Token 'y' (of type 'identifier') is unexpected here
+ --> tests/multi_error1.f90:6:12
+  |
+6 | print *, x y
+  |            ^ 
+clang: error: no such file or directory: 'tests/multi_error1.tmp.o'
```
Pranavchiku, line 24: This is expected

tests/reference/run-multi_error1-2ef69ef.stdout
```diff
@@ -0,0 +1 @@
+The command 'clang -o multi_error1.out tests/multi_error1.tmp.o  -L"$DIR/src/bin/../runtime" -Wl,-rpath,"$DIR/src/bin/../runtime" -llfortran_runtime -lm' failed.
```
Pranavchiku, line 1: This is also expected because of the header comment in PR, AST is no valid and thus `ast.ok` results to false resulting in compiler not progress to compute ASR thus LLVM despite `--continue-compilation`.

Pranavchiku: I mean these are bugs, which will be removed once we make parser truly error resilient.

tests/tests.toml
```diff
@@ -4271,7 +4271,8 @@ run = true
 
 [[test]]
 filename = "multi_error1.f90"
-ast = true
+ast = true # TODO: remove this and only use continue compilation
+continue_compilation = true
```
Pranavchiku, line 4275: once the above comments are addressed ( in subsequent PRs ) this will be converted to simply

```suggestion
continue_compilation = true
```

This will generate valid AST -> ASR -> LLVM -> binary -> execute it. RIght now AST is not valid.

run_tests.py
```diff
@@ -136,14 +136,24 @@ def is_included(backend):
...
+                    extra_args)
+            else:
+                run_test(
+                    filename,
+                    "ast",
+                    "lfortran --show-ast --no-color {infile} -o {outfile}",
+                    filename,
+                    update_reference,
+                    verify_hash,
+                    extra_args)
```
Pranavchiku, line 156: same comment as above, we can revert this change once we fix the above cases.

Pranavchiku:
I opened https://github.com/lfortran/lfortran/issues/5344 for it.

tests/reference/run-multi_error1-2ef69ef.stderr
certik: Let's stop running the code, I think as a user I never want to run the code if there are errors.

tests/reference/run-multi_error1-2ef69ef.stdout
certik: Yes, we'll fix it later.

certik:
@Pranavchiku let's get this merged, I need this to work on parser fixes.

Pranavchiku:
> Let's stop running the code, I think as a user I never want to run the code if there are errors. So the default should be not to run it. We can add a separate flag to also run the code.

shall we merge this and then I can submit a patch to fix this? I am in middle of a code debug and will need a bit of back and froth.

certik:
Yes, I think this is good to merge. Thanks!

Pranavchiku:
I opened https://github.com/lfortran/lfortran/issues/5351 for it

PR #5345, opened 2024-11-11
feat: error resilient kind parameter for intrinsics

HarshitaKalani:
Towards: #5036

### DIFF
--- a/tests/errors/array_size_05_cc.f90
+++ b/tests/errors/array_size_05_cc.f90
@@ -1,7 +1,9 @@
-program array_size_05
+program array_size_05_cc
     implicit none
     integer :: a(3)
     integer :: size_a
     integer :: kindvar = 4
     size_a = size(a, kind=kindvar, dim=1)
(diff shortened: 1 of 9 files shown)


Thirumalai-Shaktivel (commented):
.

tests/errors/array_size_05_cc.f90
```diff
@@ -0,0 +1,9 @@
+program array_size_05_cc
+    implicit none
+    integer :: a(3)
+    integer :: size_a
+    integer :: kindvar = 4
+    size_a = size(a, kind=kindvar, dim=1)
+    size_a = size(a, kind=kindvar)
```
Pranavchiku, line 7: Let's create a `tests/errors/continue_compilation.f90` and add all these statements there. So that we won't need to add multiple tests.

HarshitaKalani: I'll add more errors in the single test file from now on.

parth121101 (approved):

PR #5347, opened 2024-11-11
feat: error resilient atmoic ops and add a single test file

HarshitaKalani:
Towards: #5036


tests/errors/atomic_01_cc.f90
```diff
@@ -0,0 +1,7 @@
+program atomic_01_cc
```
Pranavchiku, line 1: Let's lump this as well

HarshitaKalani: I faced some issue with the use_fortran_env, so added this separately. Maybe, I'll look for that later if that's fine?

gxyd:
Just a general comment: we might want to delete the stale reference files as well (e.g. reference files of `tests/errors/arithmetic_if1_cc.f90`), which wouldn't be needed anymore. If you are already aware of that, that's all good I guess.

certik:
I think that looks great, exactly what we should do.

HarshitaKalani:
Thanks @gxyd for pointing it out. I forgot to. I'll just push that.

certik (approved):
I think this looks great to me.

parth121101 (approved):

PR #5359, opened 2024-11-11
feat : error resilient data implied do

AnshMehta1:


tests/errors/continue_compilation_3.f90
```diff
@@ -0,0 +1,20 @@
...
+    data(iarx(i), i=1, k) / 1, 2, 3 /
+    print *, "First error skipped"
+
+    ! data_implied_do2
+    data(iarx(i), i=1, 3, k) / 1, 2, 3 /
+    print *, "Second error skipped"
+
+    ! data_implied_do3
+    data(iarx(i), i=k, 3) / 1, 2, 3 /
+    print *, "Third error skipped"
```
certik, line 18: Let's add these into `tests/errors/array_xx_cc.f90`. Let's not add new files.

tests/tests.toml
```diff
@@ -4313,4 +4313,4 @@ asr = true
 
 [[test]]
 filename = "errors/data_implied_do3.f90"
-asr = true
+asr = true
```
certik, line 4316: remove this change, it's causing conflicts

tests/errors/array_xx_cc.f90
```diff
@@ -69,4 +71,16 @@ program array_xx_cc
...
+data(iarx(i), i=1, k) / 1, 2, 3 /
+print *, "compilation continued despite errors"
+
+! data_implied_do2
+data(iarx(i), i=1, 3, k) / 1, 2, 3 /
+print *, "compilation continued despite errors"
+
+! data_implied_do3
+data(iarx(i), i=k, 3) / 1, 2, 3 /
+print *, "compilation continued despite errors"
```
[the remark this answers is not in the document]
certik: @AnshMehta1 let's simplify the above?

tests/tests.toml
```diff
@@ -4349,4 +4349,4 @@ asr = true
 
 [[test]]
 filename = "errors/init1.f90"
-ast = true
+ast = true
```
certik, line 4352: Let's remove this unrelated change?

In a PR, only modify the files that are relevant.

tests/errors/array_xx_cc.f90
```diff
@@ -69,4 +71,13 @@ program array_xx_cc
...
 print *, "compilation continued despite errors"
 
+! data_implied_do1
+data(iarx(i), i=1, k) / 1, 2, 3 /
+
+! data_implied_do2
+data(iarx(i), i=1, 3, k) / 1, 2, 3 /
+
+! data_implied_do3
+data(iarx(i), i=k, 3) / 1, 2, 3 /
```
parth121101, line 81: Add a log here to check if compilation was continued.

certik: No need, that's already tested in the test. Rather, let's make it even simpler: https://github.com/lfortran/lfortran/pull/5359#discussion_r1838317550.

tests/tests.toml
Pranavchiku: I think this is removing end of line, but we should avoid these changes.

Pranavchiku (approved):

Pranavchiku:
mark it "ready for review" once this is ready

HarshitaKalani (approved):
Looks good to me. Thanks!

certik:
This wasn't addressed: https://github.com/lfortran/lfortran/pull/5359#discussion_r1839020789

AnshMehta1:
> This wasn't addressed: [#5359 (comment)](https://github.com/lfortran/lfortran/pull/5359#discussion_r1839020789)

Print statements were removed as per comment and only error statements are kept

AnshMehta1:
> The comment suggested to just do:
> 
> ```
> data(iarx(i), i=1, k) / 1, 2, 3 /
> data(iarx(i), i=1, 3, k) / 1, 2, 3 /
> data(iarx(i), i=k, 3) / 1, 2, 3 /
> ```

Just commented file names from which the error statement is taken , to keep the track of files . Anyways , can make these changes in any other error resiliency Pr .

certik:
> file names from which the error statement is taken

I see. I think we don't need the original filenames, do we? If the error is reported with `--continue-compilation`, then it should report without this flag also, right? If it failed, the failure mode is elsewhere in the code base that handles the `--continue-compilation` flag, the failure mode is not at the error generation site in AST->ASR.

So I think we can consolidate all our error tests into just a few files. And remove such comments.

CC @Pranavchiku, what do you think?

parthlambdatest:
> > file names from which the error statement is taken
> 
> I see. I think we don't need the original filenames, do we? If the error is reported with `--continue-compilation`, then it should report without this flag also, right? If it failed, the failure mode is elsewhere in the code base that handles the `--continue-compilation` flag, the failure mode is not at the error generation site in AST->ASR.
> 
> So I think we can consolidate all our error tests into just a few files. And remove such comments.
> 
> CC @Pranavchiku, what do you think?

Yes @certik, Agreed.

Pranavchiku:
I agree, we don't need any file names.

PR #5372, opened 2024-11-12
TEST: Remove ``NO_EXPERIMENTAL_SIMPLIFIER`` from tests which are working

czgdp1807:


integration_tests/CMakeLists.txt
```diff
@@ -494,7 +494,7 @@ RUN(NAME arrays_31 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray
 RUN(NAME arrays_32 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 # The test `arrays_33` below has `NO_EXPERIMENTAL_SIMPLIFIER` label because it fails when the `--fast` flag is used
 # with the `--experimental-simplifier` flag. It executes correctly with `--experimental-simplifier`
-RUN(NAME arrays_33 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME arrays_33 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
```
Pranavchiku, line 497: This test does not work with `--experimental-simplifier --fast`

certik:
> the tests have different behavior on macOS than Linux

It's just a bug, probably something is not initialized properly. We'll fix it once we are ready to enable the simplifier by default.

czgdp1807:
For me the opposite happens. On macOS they pass but on Linux they fail. Anyways, there are many theoretical bugs which are uncovered with the use of `simplifier`. I am fixing them. I will complete it soon. What's left for `simplifier` is simply removing `NO_EXPERIMENTAL_SIMPLIFIER` from `integration_tests/CMakeLists.txt` and running third party tests with `simplifier` pass.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -532,20 +532,48 @@ class ReplaceFunctionCallReturningArrayVisitor : public ASR::CallReplacerOnExpre
...
+    PassUtils::UpdateDependenciesVisitor w(al); \
     w.visit_TranslationUnit(unit);
+
+    if( !ASRUtils::use_experimental_simplifier ) {
+        {
+            apply_intrinsic_function_pass_macro
+        }
+        {
+            apply_intrinsic_function_pass_macro
+        }
```
certik, line 572: Is this calling the pass twice? Why?

I don't think we should be doing that, it slows down the compilation. We need to revert this I think.

src/libasr/pass/pass_manager.h
```diff
@@ -310,6 +310,7 @@ namespace LCompilers {
                 "symbolic",
                 "intrinsic_function",
                 "intrinsic_subroutine",
+                "implied_do_loops",
```
certik, line 313: This is calling the `implied_do_loops` pass twice. Let's not do that.

We should not be calling passes twice, it slows down the compilation. We need to design our pass pipeline in such a way so that each pass is only called once, ideally.

certik:
@czgdp1807 please don't merge PRs without review. I think this PR introduces some regressions that we should not do. It calls passes several times, which will be hard to undo later as it will break tests. These are not the right fixes. 

Do you know how to quickly fix this? If not then let's revert this PR.

integration_tests/CMakeLists.txt
```diff
@@ -900,7 +900,7 @@ RUN(NAME intrinsics_291 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # gamma
 RUN(NAME intrinsics_292 LABELS gfortran llvm) # command_argument_count
 RUN(NAME intrinsics_293 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # dgamma
 RUN(NAME intrinsics_294 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # log_gamma
-RUN(NAME intrinsics_295 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # exp
+RUN(NAME intrinsics_295 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NOFAST) # exp
```
certik, line 903: Adding all these NOFAST is not ideal either.

czgdp1807: Well they fail on `main` as well at least on my macOS.

src/libasr/pass/pass_manager.h
czgdp1807: Its on older version of LFortran (without `simplifier` pass). They don’t happen with `simplifier` pass. Now, it would be very difficult to keep LFortran optimal in both the cases, i.e., with and without `simplifier` pass). Do we want to keep both cases in the long run? If we want to shift entirely to `simplifier` pass based workflow then I wouldn’t be focussing much on keeping without `simplifier` based workflow perfect. I would focus more on fixing `simplifier` pass and just workaround things in the other workflow. That’s what I did here. If we want to keep both then IMHO, that shouldn’t the goal. Without `simplifier` things are unmanageable.

certik: Yes, the goal is to turn simplifier on by default, but we don't have a plan in place when that will be. So we *must* plan for the future that it won't happen for another 6 months (hopefully not, but it's a possibility). So until simplifier is on by default, we can't compromise the current code, otherwise we can't deliver beta quality. It should be easy not to compromise it: simply use `if( ASRUtils::use_experimental_simplifier ) {` and do all these changes under that that would otherwise break the default path. We can meet to figure it out together if you want.

integration_tests/CMakeLists.txt
certik: I see. Why don't they fail at the CI on macOS?

Issue #5319, opened 2024-11-09
Implement initializer for `save` array and complex variables


certik:
The simplest cases of integer/real/logical variables were fixed by #5316 and #5317. What remains is to fix complex, array and derived type variables. Here is a simple example that works with GFortran, but fails with LFortran:
```fortran
subroutine foo(reference)
integer, intent(in) :: reference(2)
integer, save :: calls(2) = [1, 1]
calls = calls + 1
print *, 'FOO: ', calls, reference
if (all(calls /= reference)) error stop
end subroutine foo

program test_save
integer i
do i=2,4
    call foo([i,i])
end do
end program test_save
```
This gives:
```console
$ lfortran a.
f90
FOO:     2    2    2    2
FOO:     2    2    3    3
ERROR STOP
```
To fix this, one must initialize these global LLVM variables in the main program.

PR #5373, opened 2024-11-13
Initalize save variables in program for arrays and complex

assem2002:
Towards #5319.
The fix doesn't work for `structTypes` yet. but it should be easy after this PR to do so.

I don't like the workarounds in `init_expr` and `simplifier` pass to make this feature work. They simplify things for the backend, but that's the way I found it would work.

Things that gets done in `program` instead of locally in the function or while creating the module (like allocating an array desc) are getting a lot. maybe we need to look into that matter to simplify things further. maybe use lambda functions to preserve these actions so it get applied one by one without caring about preserving the arguments and calling the needed function.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3635,7 +3635,16 @@ public:
                 strings_to_be_deallocated.push_back(al, llvm_utils->CreateLoad2(v->m_type, target_var));
             }
         } else {
-            builder->CreateStore(init_value, target_var);
+            if (v->m_storage == ASR::storage_typeType::Save
+                && v->m_value
+                && (ASR::is_a<ASR::Integer_t>(*v->m_type)
+                || ASR::is_a<ASR::Real_t>(*v->m_type)
+                || ASR::is_a<ASR::Logical_t>(*v->m_type))) {
+                // Do nothing, the value is already initialized
+                // in the global variable
+            } else {
+                builder->CreateStore(init_value, target_var);
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3774,61 +3836,14 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    variable_inital_value_vec.push_back({v, target_var});   
                 } else {
-                    this->visit_expr_wrapper(v->m_symbolic_value, true);
-                }
-                llvm::Value *init_value = tmp;
-                if( ASRUtils::is_array(v->m_type) &&
-                    ASRUtils::is_array(ASRUtils::expr_type(v->m_symbolic_value)) &&
-                    (ASR::is_a<ASR::ArrayConstant_t>(*v->m_symbolic_value) ||
-                    (v->m_value && ASR::is_a<ASR::ArrayConstant_t>(*v->m_value))) ) {
-                    ASR::array_physical_typeType target_ptype = ASRUtils::extract_physical_type(v->m_type);
```
assem2002, line 3791: Moving the whole code into dedicated function is good in general. We should do that frequently with `process_variable()` as it's a big function and reading through it is kinda hard.

certik (approved):
If this works, it's fine for now. Later we'll have to rework it to allow partial compilation module at a time to object files. We could do that by emitting some kind of "module initialization function", that later gets linked into the main program.

However for now we don't support it anyway, so I think this solution is fine.

certik:
I think we should likely do these transformations in ASR itself, so that we don't need to pollute the backend with it. Although your changes don't seem too bad, so I don't know. For now this looks ok.

Pranavchiku (approved):

PR #5349, opened 2024-11-11
enh: intrinsic function pass to allocate temporary of intrinsic Transpose

Pranavchiku:
Fixes #5334.

PR #5377, opened 2024-11-13
enh: set_allocation_size for transpose intrinsic array function

gxyd:
## Description

I'll first want to see if the tests pass or not. The idea is to make the test case being added here: https://github.com/lfortran/lfortran/pull/5349 to pass with `--experimental-simplifier`.

Currently, I've commented out the test case because it doesn't work without *experimental-simplifier*.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -932,6 +932,7 @@ RUN(NAME intrinsics_323 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # dot_pro
 RUN(NAME intrinsics_324 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # max
 RUN(NAME intrinsics_325 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # spread
 RUN(NAME intrinsics_326 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # minloc
+RUN(NAME intrinsics_327 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # transpose
 
 
 RUN(NAME la_constants LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NOFAST) # LAPACK constants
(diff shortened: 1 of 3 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -930,6 +930,7 @@ RUN(NAME intrinsics_322 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_323 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # dot_product
 RUN(NAME intrinsics_324 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # max
 RUN(NAME intrinsics_325 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # spread
+# RUN(NAME intrinsics_326 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # transpose
```
gxyd, line 933: currently there is no label which allows to run a test with only *experimental-simplifier* (i.e. not run the test case unless *simplifier* pass is used)

gxyd: https://github.com/lfortran/lfortran/pull/5349 fixes this issue without use of *simplifier pass*.

certik: go ahead and rebase this PR and enable the test.

integration_tests/CMakeLists.txt
```diff
@@ -523,7 +523,7 @@ RUN(NAME arrays_55 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 # is enabled i.e. works on `simplifier_pass` branch
 # RUN(NAME arrays_56 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_57 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME arrays_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_EXPERIMENTAL_SIMPLIFIER)
+RUN(NAME arrays_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
gxyd, line 526: the fix in this PR, enables this with `--experimental-simplifier` flag

PR #5340, opened 2024-11-11
enh: Fortran Backend : Handling C_loc

adit4443ya:
In continuation of #5278 
And Towards #5260 
 
 Generates proper fortran code for `ptr = c_loc(x)`

PR #5378, opened 2024-11-13
enh: Fortran Backend : Handling C_loc

adit4443ya:
In continuation of #5278
And Towards #5260

Generates proper fortran code for ptr = c_loc(x)

### DIFF
--- a/integration_tests/fortran_01.f90
+++ b/integration_tests/fortran_01.f90
@@ -1,7 +1,7 @@
 program fortran_01
-    use iso_c_binding, only: c_ptr!, c_loc
+    use iso_c_binding, only: c_ptr, c_loc
     implicit none
     integer, pointer :: x
     type(c_ptr) :: ptr
-    ! ptr = c_loc(x)
+    ptr = c_loc(x)
 end program
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -619,20 +619,23 @@ public:
     void visit_ExternalSymbol(const ASR::ExternalSymbol_t &x) {
         ASR::symbol_t *sym = down_cast<ASR::symbol_t>(
             ASRUtils::symbol_parent_symtab(x.m_external)->asr_owner);
+        if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
+            SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
+            ASR::symbol_t *sym = st->resolve_symbol("iso_c_binding");
+            if ( sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
+                src = indent;
+                src += "use ";
+                src += "iso_c_binding";
+                src += ", only: ";
+                src.append(x.m_original_name);
+                src += "\n";
+                return;
+            }
+        }
         if (!is_a<ASR::Struct_t>(*sym) && !is_a<ASR::Enum_t>(*sym)) {
             src = indent;
             src += "use ";
-            if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
-                SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
-                ASR::symbol_t *sym = st->resolve_symbol(x.m_original_name);
-                if (sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
-                    src += "iso_c_binding";
-                } else {
-                    src.append(x.m_module_name);
-                }
-        } else {
-                src.append(x.m_module_name);
-            }
+            src.append(x.m_module_name);
             src += ", only: ";
             src.append(x.m_original_name);
             src += "\n";
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -620,8 +620,6 @@ public:
         ASR::symbol_t *sym = down_cast<ASR::symbol_t>(
             ASRUtils::symbol_parent_symtab(x.m_external)->asr_owner);
         if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
-            SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
-            ASR::symbol_t *sym = st->resolve_symbol("iso_c_binding");
             if ( sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
                 src = indent;
                 src += "use ";
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -623,8 +623,14 @@ public:
             src = indent;
             src += "use ";
             if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
-                src += "iso_c_binding";
-            } else {
+                SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
+                ASR::symbol_t *sym = st->resolve_symbol(x.m_original_name);
+                if (sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
+                    src += "iso_c_binding";
+                } else {
+                    src.append(x.m_module_name);
+                }
+        } else {
                 src.append(x.m_module_name);
             }
             src += ", only: ";
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -1850,7 +1850,10 @@ public:
 
     // void visit_CLoc(const ASR::CLoc_t &x) {}
 
-    // void visit_PointerToCPtr(const ASR::PointerToCPtr_t &x) {}
+    void visit_PointerToCPtr(const ASR::PointerToCPtr_t &x) {
+        visit_expr(*x.m_arg);
+        src = "c_loc(" + src + ")";
+    }
 
     // void visit_GetPointer(const ASR::GetPointer_t &x) {}
 
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -619,16 +619,15 @@ public:
     void visit_ExternalSymbol(const ASR::ExternalSymbol_t &x) {
         ASR::symbol_t *sym = down_cast<ASR::symbol_t>(
             ASRUtils::symbol_parent_symtab(x.m_external)->asr_owner);
-        if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
-            if ( sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
-                src = indent;
-                src += "use ";
-                src += "iso_c_binding";
-                src += ", only: ";
-                src.append(x.m_original_name);
-                src += "\n";
-                return;
-            }
+        if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0 &&
+            sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
+            src = indent;
+            src += "use ";
+            src += "iso_c_binding";
+            src += ", only: ";
+            src.append(x.m_original_name);
+            src += "\n";
+            return;
         }
         if (!is_a<ASR::Struct_t>(*sym) && !is_a<ASR::Enum_t>(*sym)) {
             src = indent;
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -378,7 +378,7 @@ public:
 
     void visit_Module(const ASR::Module_t &x) {
         std::string r;
-        if (strcmp(x.m_name,"lfortran_intrinsic_iso_c_binding")==0) {
+        if (strcmp(x.m_name,"lfortran_intrinsic_iso_c_binding")==0 && x.m_intrinsic) {
             return;
         }
         r = "module";
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -378,6 +378,9 @@ public:
 
     void visit_Module(const ASR::Module_t &x) {
         std::string r;
+        if (strcmp(x.m_name,"lfortran_intrinsic_iso_c_binding")==0) {
+            return;
+        }
         r = "module";
         r += " ";
         r.append(x.m_name);
@@ -619,7 +622,11 @@ public:
         if (!is_a<ASR::Struct_t>(*sym) && !is_a<ASR::Enum_t>(*sym)) {
             src = indent;
             src += "use ";
-            src.append(x.m_module_name);
+            if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
+                src += "iso_c_binding";
+            } else {
+                src.append(x.m_module_name);
+            }
             src += ", only: ";
             src.append(x.m_original_name);
             src += "\n";


adit4443ya:
With this `visit_PointerToCPtr` fix Now the `--show-fortran` is running with this example 
Now the Error in CI test
```console
[100%] Building Fortran object CMakeFiles/fortran_01.dir/fortran_01.f90.o
CMakeFiles/fortran_01.dir/fortran_01.f90.o.tmp.f90:57:6:

   57 | ptr = c_loc(x)
      |      1
Error: Cannot convert INTEGER(4) to TYPE(c_ptr) at (1)
```

As per this [comment](https://github.com/lfortran/lfortran/pull/5340#pullrequestreview-2427272474) 
The problem here is that 
We are assigning `INTEGER(4)`  (which is address ) to the `TYPE(c_ptr)` which is like user defined struct which have member named `ptr`

It can be seen here 
```fortran
module lfortran_intrinsic_iso_c_binding
implicit none
type :: c_funptr
    integer(4) :: ptr
end type c_funptr
type :: c_ptr
    integer(4) :: ptr        ! This is the main Line
end type c_ptr
integer(4), parameter :: c_bool = 4
integer(4), parameter :: c_char = 1
integer(4), parameter :: c_double = 8
integer(4), parameter :: c_double_complex = 8
integer(4), parameter :: c_float = 4
integer(4), parameter :: c_float_complex = 4
integer(4), parameter :: c_int = 4
integer(4), parameter :: c_int16_t = 2
integer(4), parameter :: c_int32_t = 4
integer(4), parameter :: c_int64_t = 8
integer(4), parameter :: c_int8_t = 1
integer(4), parameter :: c_long = 4
integer(4), parameter :: c_long_double = -1
integer(4), parameter :: c_long_double_complex = -1
integer(4), parameter :: c_long_long = 8
character(len=1, kind=1), parameter :: c_null_char = ""
type(c_ptr), parameter :: c_null_ptr = c_ptr(0)
integer(4), parameter :: c_short = 2
integer(4), parameter :: c_size_t = 8
interface
    logical(4) function c_associated(c_ptr_1)
        import c_ptr
        type(c_ptr), intent(in) :: c_ptr_1
    end function c_associated

    subroutine c_f_pointer(cptr, fptr, shape)
        import c_ptr
        type(c_ptr), intent(in) :: cptr
        integer(4), pointer, intent(out) :: fptr
        integer(4), dimension(:), intent(in), optional :: shape
    end subroutine c_f_pointer

    integer(4) function c_funloc(x)
        integer(4), intent(in) :: x
    end function c_funloc

    integer(4) function c_loc(x)
        integer(4), intent(in) :: x
    end function c_loc
end interface
end module lfortran_intrinsic_iso_c_binding

program fortran_01
use lfortran_intrinsic_iso_c_binding, only: c_loc
use lfortran_intrinsic_iso_c_binding, only: c_ptr
implicit none
type(c_ptr) :: ptr
integer(4), pointer :: x
ptr = c_loc(x)
end program fortran_01
```

adit4443ya:
Since there is no compilation error in compiling and running this `fortran_01.f90` example
Therefore i Had an idea that  we need to find a spot where i can do something like 
```shell
ASR::ttype_t* cptr = ASRUtils::TYPE(ASR::make_CPtr_t(al, arg->base.loc));
```
Which is basically like 
`ptr=c_ptr(c_loc(x))`  (This is just Bad way to show it, But just for understanding that what we need)

certik:
In the ASR->Fortran, do not print the `iso_c_binding`, since that is our own internal implementation that will not work with other compilers. Simply emit `use iso_c_binding` when you see that intrinsic module in ASR. Ideally importing only things that you need.

Thirumalai-Shaktivel:
Right, Replace `lfortran_intrinsic_iso_c_binding` with `iso_c_binding`.
Also, I think we shouldn't generate the `lfortran_intrinsic_iso_c_binding` module itself, as we don't use it.

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -1832,7 +1832,10 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
 
     // void visit_CLoc(const ASR::CLoc_t &x) {}
 
-    // void visit_PointerToCPtr(const ASR::PointerToCPtr_t &x) {}
+    void visit_PointerToCPtr(const ASR::PointerToCPtr_t &x) {
+        visit_expr(*x.m_arg);
+        src = "c_loc(" + src + ")";
```
Thirumalai-Shaktivel, line 1837: Also, in ASR `c_loc` is represented as `PointerToCPtr` but not an ExternalSymbol, so I think you need to insert an `use` statement in this visitor. See how `math.h` is inserted in C backend.

Pranavchiku:
> In the ASR->Fortran, do not print the iso_c_binding

So, let's break down this, right now we print entire module and do `use lfortran_intrinsic_iso_c_binding` as soon as we encounter

```clojure
                            c_loc:
                                (ExternalSymbol
                                    2
                                    c_loc
                                    4 c_loc
                                    lfortran_intrinsic_iso_c_binding
                                    []
                                    c_loc
                                    Public
                                )
```

```fortran
module lfortran_intrinsic_iso_c_binding
implicit none
type :: c_funptr
    integer(4) :: ptr
end type c_funptr
type :: c_ptr
    integer(4) :: ptr        ! This is the main Line
end type c_ptr
integer(4), parameter :: c_bool = 4
integer(4), parameter :: c_char = 1
integer(4), parameter :: c_double = 8
integer(4), parameter :: c_double_complex = 8
integer(4), parameter :: c_float = 4
integer(4), parameter :: c_float_complex = 4
integer(4), parameter :: c_int = 4
integer(4), parameter :: c_int16_t = 2
integer(4), parameter :: c_int32_t = 4
integer(4), parameter :: c_int64_t = 8
integer(4), parameter :: c_int8_t = 1
integer(4), parameter :: c_long = 4
integer(4), parameter :: c_long_double = -1
integer(4), parameter :: c_long_double_complex = -1
integer(4), parameter :: c_long_long = 8
character(len=1, kind=1), parameter :: c_null_char = ""
type(c_ptr), parameter :: c_null_ptr = c_ptr(0)
integer(4), parameter :: c_short = 2
integer(4), parameter :: c_size_t = 8
interface
    logical(4) function c_associated(c_ptr_1)
        import c_ptr
        type(c_ptr), intent(in) :: c_ptr_1
    end function c_associated

    subroutine c_f_pointer(cptr, fptr, shape)
        import c_ptr
        type(c_ptr), intent(in) :: cptr
        integer(4), pointer, intent(out) :: fptr
        integer(4), dimension(:), intent(in), optional :: shape
    end subroutine c_f_pointer

    integer(4) function c_funloc(x)
        integer(4), intent(in) :: x
    end function c_funloc

    integer(4) function c_loc(x)
        integer(4), intent(in) :: x
    end function c_loc
end interface
end module lfortran_intrinsic_iso_c_binding
```

So, you need to do the following:
* [ ] Go to `visit_Module`, check if it is `IntrinsicModule`, and name is `lfortran_intrinsic_iso_c_binding`. Then skip this

```clojure
iso_c_binding:
    (IntrinsicModule lfortran_intrinsic_iso_c_binding)
```

* [ ] In `visit_ExternalSymbol`, if `m_module_name` is `lfortran_intrinsic_iso_c_binding` then just append `use_iso_c_binding, only: xx` instead of `use lfortran_intrinsic_iso_c_binding`

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -378,6 +378,9 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
 
     void visit_Module(const ASR::Module_t &x) {
         std::string r;
+        if (strcmp(x.m_name,"lfortran_intrinsic_iso_c_binding")==0) {
+            return;
+        }
```
certik, line 383: Even better is to *also* check the "Intrinsic" flag of the Module ASR, in case the user had a user module of this name (as unlikely as it is).

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -619,7 +622,11 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
         if (!is_a<ASR::Struct_t>(*sym)) {
             src = indent;
             src += "use ";
-            src.append(x.m_module_name);
+            if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
```
certik, line 625: Same here.

Pranavchiku:
Any updates over this PR?

src/libasr/codegen/asr_to_fortran.cpp
adit4443ya: done

src/libasr/codegen/asr_to_fortran.cpp
adit4443ya: I think External symbol do not have `intrinsic` flag

adit4443ya:
> Any updates over this PR?

I think it's now ready

src/libasr/codegen/asr_to_fortran.cpp
Pranavchiku:
```diff
diff --git a/src/libasr/codegen/asr_to_fortran.cpp b/src/libasr/codegen/asr_to_fortran.cpp
index e3df26cf9..e1448af73 100644
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -616,6 +616,17 @@ public:
     void visit_ExternalSymbol(const ASR::ExternalSymbol_t &x) {
         ASR::symbol_t *sym = down_cast<ASR::symbol_t>(
             ASRUtils::symbol_parent_symtab(x.m_external)->asr_owner);
+        if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
+            SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
+            ASR::symbol_t *sym = st->resolve_symbol(x.m_original_name);
+            if ( sym && ASR::is_a<ASR::Module_t>(*sym) && ASR::down_cast<ASR::Module_t>(sym)->m_intrinsic) {
+                src += "iso_c_binding";
+            } else {
+                src.append(x.m_module_name);
+            }
+        } else {
+            src.append(x.m_module_name);
+        }
         if (!is_a<ASR::Struct_t>(*sym) && !is_a<ASR::Enum_t>(*sym)) {
             src = indent;
             src += "use ";
```

integration_tests/fortran_01.f90
```diff
@@ -1,7 +1,7 @@
 program fortran_01
```
Pranavchiku, line 1: > % lfortran ./integration_tests/fortran_01.f90 --show-fortran 


```fortran
program fortran_01
use lfortran_intrinsic_iso_c_binding, only: c_loc
use lfortran_intrinsic_iso_c_binding, only: c_ptr
implicit none
type(c_ptr) :: ptr
integer(4), pointer :: x
ptr = c_loc(x)
end program fortran_01
```

This is incorrect

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -619,7 +622,17 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
         if (!is_a<ASR::Struct_t>(*sym) && !is_a<ASR::Enum_t>(*sym)) {
             src = indent;
             src += "use ";
-            src.append(x.m_module_name);
+            if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
+                SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
+                ASR::symbol_t *sym = st->resolve_symbol(x.m_original_name);
```
Thirumalai-Shaktivel, line 627:
```suggestion
```
Maybe, this might fix it?

The symbol name in the SymbolTable is "iso_c_binding", but not "lfortran_intrinsic_iso_c_binding". So, I think let's use the existing the asr_owner (sym) at line 620.

src/libasr/codegen/asr_to_fortran.cpp
Thirumalai-Shaktivel: My suggestion is required only when we apply unused_functions pass for Fortran backend.

src/libasr/codegen/asr_to_fortran.cpp
Thirumalai-Shaktivel: Or try resolving the symbol using "iso_c_binding".

Pranavchiku (approved):
I observed a few things in this PR, the diff was applied incorrectly, `if ... else` was messed up signifying unnecessary hurry to wrap this PR. I don't think you have tested the code with `lfortran ./integration_tests/fortran_01.f90 --show-fortran` either and tried to check if it produces output that we expect. Please keep all these in mind and try to check locally before pushing changes. I made changes, output is expected. The PR looks good.

src/libasr/codegen/asr_to_fortran.cpp
Pranavchiku: Yes, absolutely that fixed it. The diff was to make him understand what to do, I didn't pay attention that much while commenting. @Thirumalai-Shaktivel feel free to review it once again and if looks good this PR is ready to merge. I'll prefer `squash and merge` for this.

src/libasr/codegen/asr_to_fortran.cpp
Pranavchiku: >  so I think you need to insert an use statement in this PointerToCPtr  node

Assuming the Fortran code user wrote is correct. We'll always have a `use iso_c_binding` or similar in ASR so `visit_Use` shall tackle this, no?

Thirumalai-Shaktivel: Yes, but if we apply the unused_function pass, the ExternalSymbol gets removed. As we use PointerToCPtr, instead of a symbol from lfortran_intrinsic_iso_c_binding module. So, the ExternalSymbol becomes unused.

See the generated fortran code, after applying the pass:
```fortran
! Fortran code after applying the pass: unique_symbols

program fortran_01
use iso_c_binding, only: c_ptr
implicit none
type(c_ptr) :: ptr
integer(4), pointer :: x
ptr = c_loc(x)
end program fortran_01
```
And it will fail with gfortran.

It's not a blocker as we don't apply passes for fortran backend. So, it's fine for now.

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -616,6 +619,19 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
     void visit_ExternalSymbol(const ASR::ExternalSymbol_t &x) {
         ASR::symbol_t *sym = down_cast<ASR::symbol_t>(
             ASRUtils::symbol_parent_symtab(x.m_external)->asr_owner);
+        if (strcmp(x.m_module_name,"lfortran_intrinsic_iso_c_binding")==0) {
+            SymbolTable* st = ASRUtils::symbol_parent_symtab(sym);
+            ASR::symbol_t *sym = st->resolve_symbol("iso_c_binding");
```
Thirumalai-Shaktivel, line 624: Instead, we can directly use the sym variable at line 620, right?

Thirumalai-Shaktivel (approved):
LGTM, the below suggestion can be resolved in the subsequent PRs.

Thanks for the work

src/libasr/codegen/asr_to_fortran.cpp
Thirumalai-Shaktivel: It seems to be symbol of `asr_owner` for the `symbol_parent_symtab`, which would be a module, right?

adit4443ya:
> I observed a few things in this PR, the diff was applied incorrectly, `if ... else` was messed up signifying unnecessary hurry to wrap this PR. I don't think you have tested the code with `lfortran ./integration_tests/fortran_01.f90 --show-fortran` either and tried to check if it produces output that we expect. Please keep all these in mind and try to check locally before pushing changes. I made changes, output is expected. The PR looks good.

I really aplogise for this negligence ,, actually i checked it but i forgot to build it again n just checked with `--show-fortran`  and it gave correct output , in hurry i pushed changes.
I am really sorry for this, i will keep things and your suggestions in mind

adit4443ya:
Thanks to both of you @Pranavchiku  and @Thirumalai-Shaktivel  for guiding me through this , as it was somehow confusing and it was unclear from my side that what do we want here,,,

src/libasr/codegen/asr_to_fortran.cpp
Pranavchiku: Oh yes, I agree. Thanks for this!

Pranavchiku:
No worries, everyone learn through mistakes :)

certik (approved):
I think that this is fine.

gxyd (approved):
LGTM.

gxyd:
As the changes aren't too big and very much related to each other, I'll squash and merge this PR.

PR #5381, opened 2024-11-13
ref: combined errors in single file

parth121101:
Towards: #5036


certik (approved):
I think this looks great, much better!

PR #5383, opened 2024-11-13
Enh : Utilize stringDescriptor size for `len()`

assem2002:
Instead of relaying on runtime C's `strlen` to return the size of the string by iterating over the consecutive characters till it reaches the null character, We simply use the size if the string we have is of physical type stringDescriptor `{char_ptr, size, capacity}`


certik (approved):
I think this is fine.

It looks like in the past you only allowed PointerString in StringLen, while now you also allow DestriptorString, correct?

Later we should extend ASR verify to ensure for each string node what physical types are allowed.

assem2002:
Yeah exactly. The backend can handle both of them now.

PR #5392, opened 2024-11-15
Fixes for simplifier pass

czgdp1807:


czgdp1807:
I don't know how the `Fortran compiler FLAGS` is set to `-g` in the following configuration,

```zsh
Configuration results
---------------------
Fortran compiler: /home/runner/work/lfortran/lfortran/src/bin/lfortran
C compiler      : /usr/bin/cc
Build type: Debug
Fortran compiler flags: -g
C compiler flags      : -g
Installation prefix: /usr/local
LFORTRAN_BACKEND: llvm
FAST: no
EXPERIMENTAL_SIMPLIFIER: no
```

czgdp1807:
So I don't know why the above configuration is happening on two of the failing CI jobs. Very weird.

certik:
Cmake does it automatically I think. We need to downgrade cmake version. I've seen this in other PRs too.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -532,6 +532,26 @@ class ReplaceFunctionCallReturningArrayVisitor : public ASR::CallReplacerOnExpre
                 if (x.m_overloaded) {
                     visit_stmt(*x.m_overloaded);
                 }
+
+                /*
+                Uncomment in case we encounter a case where target and value ranks are different.
```
certik, line 537: You can put an assert statement here that would trigger if target /= value rank to notify us of this issue.

certik:
I think this looks good, it doesn't seem to break anything in the default path.

certik:
Just rebase or merge main and the CI should pass.

src/libasr/pass/intrinsic_function.cpp
czgdp1807: First, this is not needed in `simplifier`, only needed in older LFortran (without `simplifier` pass). And it won't happen here. When I will add a pass to replace expressions with their compile time values then we will put this code there. And `intrinsic_functions` pass will be called after that, so it will work. I am just keeping it here for reference because I know we will need this again soon after `simplifier` is over. Because I will be adding the said pass after `simplifier` is complete.

Issue #5393, opened 2024-11-15
procedure with an implementation raises semantic error


gxyd:
## Minimal Reproducible Example

```f90
module nf_layer

    !! This module provides the `layer` type that is part of the public
    !! user-facing API.
  
    ! use nf_base_layer, only: base_layer
    implicit none
  
    private
    public :: layer
  
    type :: layer

        contains
        ! Specific subroutines for different array ranks
        procedure, private :: backward_1d
        generic :: backward => backward_1d

    end type layer

    interface backward

        pure module subroutine backward_1d(self, previous, gradient)
            class(layer), intent(in out) :: self
            class(layer), intent(in) :: previous
            real, intent(in) :: gradient(:)
        end subroutine backward_1d
    end interface backward

end module nf_layer

```

For the above fortran program, where the procedure `backward_1d` has an implementation in the interface `backward`, on program's execution with LFortran, raises the below error:
```console
> lfortran --version
LFortran version: 0.41.0-477-g56816414a-dirty
Platform: macOS ARM
LLVM: 11.0.1
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
semantic error: PROCEDURE(interface) should be declared DEFERRED
  --> test.f90:16:9
   |
16 |         procedure, private :: backward_1d
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

gxyd:
I'll look into a fix for the issue.

PR #5399, opened 2024-11-16
fix: no semantic error for class procedure without "deferred" and without implementation

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/5393.

Now, we don't raise semantic error when a class procedure without *DEFERRED* and without an implementation is used. The implementation for the procedure might be present in some other file as well.

Ideally, the linker should raise an error for when an implementation isn't provided for a class procedure without *DEFRRED*.


tests/procedure_without_deferred_impl.f90
```diff
@@ -0,0 +1,31 @@
+! test to ensure that class procedure without "deferred"
+! and without implementation doesn't rasie semantic error
```
Pranavchiku, line 2:
```suggestion
! and without implementation doesn't raise semantic error
```

tests/procedure_without_deferred_impl.f90
```diff
@@ -0,0 +1,31 @@
...
+module nf_layer
```
Pranavchiku, line 3:
```suggestion
module module_procedure_without_deferred_impl
```
Use this and elsewhere

tests/tests.toml
```diff
@@ -2644,6 +2644,10 @@ filename = "../integration_tests/class_04.f90"
 asr = true
 llvm = true
 
+[[test]]
+filename = "procedure_without_deferred_impl.f90"
+asr = true
```
Pranavchiku, line 2649: why only ASR, does it not compile to LLVM and binary?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2135,10 +2135,6 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                     }
                 }
                 ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(proc_sym);
-                if (!is_deferred &&
-                    ASRUtils::get_FunctionType(*func)->m_deftype == ASR::deftypeType::Interface) {
-                    throw SemanticError("PROCEDURE(interface) should be declared DEFERRED", loc);
-                }
```
Pranavchiku, line 2141: I will not remove it rather comment and say strengthen this check because we might need it someday later.

Pranavchiku (changes requested):
left a few comments above

tests/tests.toml
Thirumalai-Shaktivel: Yes, if it generates the binaries, let's move the test to integration_tests.

Mostly we add test in tests when there is no support for LLVM or if ASR representation shouldn't get modified in the future.

src/lfortran/semantics/ast_symboltable_visitor.cpp
gxyd: I'm not sure what you mean by *strengthening* this check, cause, I believe it's a mistake to raise a semantic error of any such sort here. For a non-deferred and non-implementation defined procedure in a semantic check.

tests/tests.toml
gxyd: > does it not compile to LLVM

It gives the below LLVM IR:
```ll
; ModuleID = 'LFortran'
source_filename = "LFortran"
```

(seems like our LLVM backend doesn't generated any code for a non-deferred procedure without any implementation)

I'll check for why that's the case.

In cases when an implementation actually is present, that is, in case the below program is used (which includes an implementation of `backward_1d`:
```f90
! test to ensure that class procedure without "deferred"
! and without implementation doesn't rasie semantic error
module nf_layer
    implicit none

    private

    public :: layer

    type, abstract :: layer
        real, allocatable :: weights(:)
        real :: learning_rate = 0.01

        contains

        ! class procedure without "DEFERRED" and without
        ! an implementation
        procedure(backward_1d), private, deferred :: backward_1d
        generic :: backward => backward_1d
    end type

    interface backward
        module subroutine backward_1d(self, previous, gradient)
            class(layer), intent(inout) :: self
            class(layer), intent(in) :: previous
            real, intent(in) :: gradient(:)
        end subroutine
    end interface

    contains

    module subroutine backward_1d(self, previous, gradient)
        class(layer), intent(inout) :: self
        class(layer), intent(in) :: previous
        real, intent(in) :: gradient(:)
        print *, "Hello world"
    end subroutine
end module
```

then the below LLVM IR is generated:
```ll
; ModuleID = 'LFortran'
source_filename = "LFortran"

%layer_polymorphic = type { i64, %layer* }
%layer = type { %array*, float }
%array = type { float*, i32, %dimension_descriptor*, i1, i32 }
%dimension_descriptor = type { i32, i32, i32 }
%__vtab_layer = type { i64 }

@0 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@1 = private unnamed_addr constant [12 x i8] c"Hello world\00", align 1
@2 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1
@3 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@4 = private unnamed_addr constant [12 x i8] c"Hello world\00", align 1
@5 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define void @__module_nf_layer_backward_1d(%layer_polymorphic* %self, %layer_polymorphic* %previous, %array* %gradient) {
.entry:
  %0 = alloca %__vtab_layer, align 8
  %1 = getelementptr %__vtab_layer, %__vtab_layer* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([5 x i8], [5 x i8]* @2, i32 0, i32 0), i8* getelementptr inbounds ([12 x i8], [12 x i8]* @1, i32 0, i32 0), i8* getelementptr inbounds ([2 x i8], [2 x i8]* @0, i32 0, i32 0))
  br label %return

  ret void
}

define void @__module_nf_layer_backward_1d_real____2(%layer_polymorphic* %self, %layer_polymorphic* %previous, float* %gradient, i32* %__1gradient, i32* %__2gradient) {
.entry:
  %0 = alloca %__vtab_layer, align 8
  %1 = getelementptr %__vtab_layer, %__vtab_layer* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([5 x i8], [5 x i8]* @5, i32 0, i32 0), i8* getelementptr inbounds ([12 x i8], [12 x i8]* @4, i32 0, i32 0), i8* getelementptr inbounds ([2 x i8], [2 x i8]* @3, i32 0, i32 0))
  br label %return

  ret void
}

declare void @_lfortran_printf(i8*, ...)
```

gxyd: Seems like `unused_functions` pass completely remove `backward` GenericProcedure, because it isn't being used.

Pranavchiku: Modify the test to use it, also pass a variable and test it inside. That way we can be assured that everything works.

src/lfortran/semantics/ast_symboltable_visitor.cpp
Pranavchiku: > I'm not sure what you mean by strengthening this check

Narrowing down the scope to throw SemanticError.

> , I believe it's a mistake to raise a semantic error of any such sort here,

I haven't checked semantics yet, let me do that.

tests/tests.toml
gxyd: I'll test it by adding the implementation to the a new module I guess.

gxyd:
The added integration test case currently raises an error (aware of that), I need to fix that as well:
```console
code generation error: Subroutine code not generated for 'backward_1d'



```

integration_tests/procedure_07_module.f90
```diff
@@ -0,0 +1,32 @@
...
+    public :: layer
+
+    type :: layer
+        real, allocatable :: weights(:)
+        real :: learning_rate = 0.01
+
+        contains
+
+        ! class procedure without "DEFERRED" and without
+        ! an implementation
```
gxyd, line 18: this has an implementation now in `procedure_07.f90`, I'll update this comment

Pranavchiku:
> code generation error: Subroutine code not generated for 'backward_1d'

Exactly, something is tend to fail because we now support something that we have not handled in either passes or backend.

src/lfortran/semantics/ast_symboltable_visitor.cpp
certik: First let's write tests for this error message, possibly as a separate PR first. Let's ensure those tests are good.

Then we can update this PR and we need to ensure the error tests still report errors.

PR #5403, opened 2024-11-16
Ref: append errors in diagnostics for `ast_common_visitor` file and removed Semantic Errors

parth121101:
Towards: #5036


parth121101:
Initially, we were throwing semantic errors in the `ast_common_visitor` file in about 180 different locations. 165 of these 180 have been refactored.
Rest 15 have some issues. While do that in a separate PR.

certik (approved):
I think this looks good, thank you for doing it.

certik:
@parth121101 watch the CI in `main`, there might be some failures, as we merged several big PRs that might introduce some failures. If so, it should not be difficult to fix.

parth121101:
> @parth121101 watch the CI in `main`, there might be some failures, as we merged several big PRs that might introduce some failures. If so, it should not be difficult to fix.

Sure @certik

PR #5406, opened 2024-11-17
Enh: refactor semantic errors in ast_symboltable_visitor file

HarshitaKalani:
Towards: #5036

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -24,7 +24,7 @@
 namespace LCompilers::LFortran {
 
 template <typename T>
-void extract_bind(T &x, ASR::abiType &abi_type, char *&bindc_name) {
+void extract_bind(T &x, ASR::abiType &abi_type, char *&bindc_name, diag::Diagnostics &diag) {
     if (x.m_bind) {
         AST::Bind_t *bind = AST::down_cast<AST::Bind_t>(x.m_bind);
         if (bind->n_args == 1) {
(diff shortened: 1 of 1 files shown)


HarshitaKalani:
I've refactored the rest of the files too. Will push that after this gets merged.

parth121101:
> I've refactored the rest of the files too (except common_visitor which @parth121101 is working on). Will push that after this gets merged.

@HarshitaKalani even common_visitor is refactored. Look PR 5403

HarshitaKalani:
I was talking about the rest 15 you've mentioned here: https://github.com/lfortran/lfortran/pull/5403#issuecomment-2480688554

parth121101 (approved):

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -36,16 +36,25 @@ void extract_bind(T &x, ASR::abiType &abi_type, char *&bindc_name) {
                 } else if (to_lower(std::string(name->m_id)) == "js") {
                     abi_type=ASR::abiType::BindJS;
                 } else {
-                    throw SemanticError("Unsupported language in bind()",
-                        x.base.base.loc);
+                    diag.add(diag::Diagnostic(
+                        "Unsupported language in bind()",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {x.base.base.loc})}));
+                    throw SemanticAbort();
```
gxyd, line 43: So, the idea is that with an incorrect specification of `bind` (e.g.; `bind(name=1)`) we raise only *one* error and not more than one?

As `bind(name=1)` has two possible errors for it, an unspecified binding specifier (like 'C') and incorrect value of `name` (as it only expects a string). So, we only intend to raise one error and exit out of that?

gxyd (commented):
I'm not really always comfortable with change, unless I also see test cases for it. And I think it's possible with these changes that we might throw one error and exit out (as there is a use of `SemanticAbort()`) if my understand of the code here is correct.

Please correct me if I'm wrong in understanding here.

certik (approved):
It seems this PR is not changing any behavior. Only allowing (optionally) to continue compilation and getting rid of SemanticError, which we should get rid of completely. It seems it is behaving exactly as before.

HarshitaKalani:
Yes @gxyd , the PR is to needed for error resiliency. Here, we're just getting rid of semanticError so that we can continue to compile even when there's error. I hope this clears things.

PR #5407, opened 2024-11-17
Fixed a bug in pass class constructor

### LINKED ISSUE #5357: bug: `class_constructor` pass removes statement `fn(cls(), cls())` from ASR

```fortran
program main
    implicit none
    
    type :: Circle
        real :: radius = 1.0
    end type Circle
    
    call print_radius(Circle(1.5), Circle())

contains

    subroutine print_radius(m, n)
        type(Circle), intent(in) :: m, n
        print *, m%radius, n%radius
    end subroutine

end program main
```

#### GFortran

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran ./examples/example.f90 && ./a.out
   1.50000000       1.00000000 
```

#### LFortran

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran ./examples/example.f90
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ 
``` 

This happens because the subroutine call is removed in the `class_constructor` pass. It occurs only for cases when the class instances passed directly to the subroutine call are more than one; for a single argument (`fn(cls())`), the program executes correctly.

### ASR after applying `class_constructor` pass

```clojure
;; ASR after applying the pass: class_constructor
(TranslationUnit
    (SymbolTable
        1
        {
            main:
                (Program
                    (SymbolTable
                        2
                        {
                            1_circle_radius:
                                (ExternalSymbol
                                    2
                                    1_circle_radius
                                    3 radius
                                    circle
                                    []
                                    radius
                                    Public
                                ),
                            circle:
                                (Struct
                                    (SymbolTable
                                        3
                                        {
                                            radius:
                                                (Variable
                                                    3
                                                    radius
                                                    []
                                                    Local
                                                    (RealConstant
                                                        1.000000
                                                        (Real 4)
                                                    )
                                                    (RealConstant
                                                        1.000000
                                                        (Real 4)
                                                    )
                                                    Save
                                                    (Real 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                )
                                        })
                                    circle
                                    []
                                    [radius]
                                    Source
                                    Public
                                    .false.
                                    .false.
                                    []
                                    ()
                                    ()
                                ),
                            print_radius:
                                (Function
                                    (SymbolTable
                                        4
                                        {
                                            1_circle_radius:
                                                (ExternalSymbol
                                                    4
                                                    1_circle_radius
                                                    3 radius
                                                    circle
                                                    []
                                                    radius
                                                    Public
                                                ),
                                            m:
                                                (Variable
                                                    4
                                                    m
                                                    []
                                                    In
                                                    ()
                                                    ()
                                                    Default
                                                    (StructType
                                                        2 circle
                                                    )
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                ),
                                            n:
                                                (Variable
                                                    4
                                                    n
                                                    []
                                                    In
                                                    ()
                                                    ()
                                                    Default
                                                    (StructType
                                                        2 circle
                                                    )
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                )
                                        })
                                    print_radius
                                    (FunctionType
                                        [(StructType
                                            2 circle
                                        )
                                        (StructType
                                            2 circle
                                        )]
                                        ()
                                        Source
                                        Implementation
                                        ()
                                        .false.
                                        .false.
                                        .false.
                                        .false.
                                        .false.
                                        []
                                        .false.
                                    )
                                    []
                                    [(Var 4 m)
                                    (Var 4 n)]
                                    [(Print
                                        (StringFormat
                                            ()
                                            [(StructInstanceMember
                                                (Var 4 m)
                                                4 1_circle_radius
                                                (Real 4)
                                                ()
                                            )
                                            (StructInstanceMember
                                                (Var 4 n)
                                                4 1_circle_radius
                                                (Real 4)
                                                ()
                                            )]
                                            FormatFortran
                                            (String -1 0 () PointerString)
                                            ()
                                        )
                                    )]
                                    ()
                                    Public
                                    .false.
                                    .false.
                                    ()
                                ),
                            temp_struct_var__:
                                (Variable
                                    2
                                    temp_struct_var__
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (StructType
                                        2 circle
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    main
                    []
                    [(Assignment
                        (StructInstanceMember
                            (Var 2 temp_struct_var__)
                            2 1_circle_radius
                            (Real 4)
                            ()
                        )
                        (RealConstant
                            1.500000
                            (Real 4)
                        )
                        ()
                    )]
                )
        })
    []
)
```

Found while studying the case of directly passing class instances to functions, as part of #3770.

tanay-man:
fixes #5357


kmr-srbh (approved):
Thanks for this! The changes look good to me. Please add a reference test with `run = true ` and `pass = "class_constructor"` and we should be able to merge this.

kmr-srbh:
With the changes in this PR, we create a new auxiliary variable for every `StructConstructor` we encounter. This has a drawback for cases where every `StructConstructor` is same - `call print_radius(Circle(1.5), Circle(1.5))`. I think we should ideally have only 1 common auxiliary variable and reuse it every time we encounter the constructor. I think it would require checking the equality of the current `result_var` value with the value of the `StructInstanceMember`. I do not know how much of an impact this comparison can have on performance. @certik could you please guide? Should we do this?

tests/tests.toml
```diff
@@ -4282,6 +4282,10 @@ filename = "multi_error1.f90"
 ast = true # TODO: remove this and only use continue compilation
 continue_compilation = true
 
+[[test]]
+filename = "multiple_objects_args.f90"
+run =true
+
```
kmr-srbh, line 4288:
```suggestion
pass = "class_constructor"
```
Let's keep a reference of the ASR after the `class_constructor` pass.

gxyd:
Hi @tanay-man , I get the below reference updates when I run `./run_tests.py -u` on your branch locally on my machine:
```console
~/OpenSource/lfortran constructor_pass !2 ?30 ❯ git status                                                                                                                                                                                                    ~/OpenSource/lfortran constructor_pass ?30 ❯ git status                                                                                                                                                                                                            lf 03:37:21 PM
On branch constructor_pass
Your branch is up to date with 'tanay-man/constructor_pass'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	tests/reference/pass_class_constructor-multiple_objects_args-6e697b8.json
	tests/reference/pass_class_constructor-multiple_objects_args-6e697b8.stdout

```

I think you just need to do `git add` the reference files.

tanay-man:
Thanks @gxyd, I had run it before adding the change `pass = "class_constructor"` .  I will do it right away 👍 .

tests/tests.toml
```diff
@@ -4282,6 +4282,11 @@ filename = "multi_error1.f90"
 ast = true # TODO: remove this and only use continue compilation
 continue_compilation = true
 
+[[test]]
+filename = "multiple_objects_args.f90"
+run =true
```
certik, line 4287: Instead of `run` true, let's add this as an integration test and ensure the result is correct in the test itself via `error stop`.

The `pass` test below is fine, as it checks ASR, which was incorrect before I am assuming.

certik:
@kmr-srbh this PR makes code that didn't run before to run?  Is there existing code that ran before, but now has an extra variable?

In other words, it seems that this PR moves us in the right direction without making anything slower. So we can merge it, and then investigate if there is a way to reuse variables. However for performance, it might not matter much if they can be optimized out.

kmr-srbh:
> @kmr-srbh this PR makes code that didn't run before to run? 

Yes. The code before had allowed creating only 1 temporary (`temp_struct_var__`) for `StructConstructor` in `stmt`. It failed for the linked issue, and for cases like:

```fortran
program main
    implicit none
    
    type :: Circle
        real :: radius = 1.0
    end type Circle

    type :: Square
        real :: side = 2.0
    end type Square
    
    print *, Circle(), Square(1.5)

end program main
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran ./examples/example.f90 && ./a.out
   1.00000000       1.50000000    
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran ./examples/example.f90
code generation error: circle doesn't have any member named side
  --> ./examples/example.f90:12:24
   |
12 |     print *, Circle(), Square(1.5)
   |                        ^^^^^^^^^^^ 


```

> Is there existing code that ran before, but now has an extra variable?

No, as the previous code worked for `stmt` nodes with only 1 `StructConstructor`. With the changes in this PR, we always create a new variable for every `StructConstructor`.

certik:
I think this looks good. @tanay-man do you want to address the comment https://github.com/lfortran/lfortran/pull/5407#discussion_r1846906025 so that we can merge it?

tanay-man:
I haven't tested the performance regression yet, but I think that it should be a problem as this PR only allows multiple temp vars when it is necessary.

tests/tests.toml
tanay-man: Ok I will make the change 👍

tanay-man: Are the test flags correct?

src/libasr/pass/pass_utils.h
```diff
@@ -608,8 +608,8 @@ namespace LCompilers {
                 }
                 return ;
             }
-            if( replacer->result_var == nullptr ) {
-                std::string result_var_name = replacer->current_scope->get_unique_name("temp_struct_var__", false);
+            if( replacer->result_var == nullptr ) {
```
certik, line 611:
```suggestion
            if( replacer->result_var == nullptr ) {
```

certik (approved):
I think it looks good now, thanks!

certik:
The fortran backend failed, just remove the label for now.

certik (approved):

PR #5417, opened 2024-11-19
feat: add more error tests with continue compilation flag

parth121101:
Towards: #5036

### DIFF
--- a/tests/errors/continue_compilation_2.f90
+++ b/tests/errors/continue_compilation_2.f90
@@ -49,6 +49,8 @@ program continue_compilation_2
     integer :: i_2, j_1
     integer :: i_3
     complex :: a_4
+    complex :: a_5
+    real :: y_4
 
     ! c_f_pointer_01
     call c_f_pointer(queries_1, y_1, [2])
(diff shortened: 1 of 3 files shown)


tests/errors/continue_compilation_2.f90
```diff
@@ -140,5 +142,19 @@ program continue_compilation_2
...
+    !flush_invalid_kwarg
+    FLUSH(unit=10, start=100)
+    !func_parameter_type
+    print *, f(42.9)
+
+    CONTAINS
+    LOGICAL FUNCTION f(x)
+        INTEGER, INTENT(IN), OPTIONAL :: x
+        f = PRESENT(x)
+    END FUNCTION
```
certik, line 158: Use lower case

tests/errors/continue_compilation_2.f90
```diff
@@ -49,6 +49,8 @@ program continue_compilation_2
     integer :: i_2, j_1
     integer :: i_3
     complex :: a_4
+    complex :: a_5
```
certik, line 52: why do we need adding new variables here, why cannot we reuse variables? Also we should add variables on the same line, that way we don't constantly move the line numbers in the output.

PR #5423, opened 2024-11-20
build: use `ninja` for build

kmr-srbh:
Use `ninja` for generating build files and `ccache` for caching compilation.

### DIFF
--- a/build1.sh
+++ b/build1.sh
@@ -14,5 +14,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+-G Ninja \
     .
-cmake --build . -j16 --target install
+cmake --build . --target install
--- a/build1.bat
+++ b/build1.bat
@@ -7,4 +7,4 @@ cmake ^
     -DCMAKE_INSTALL_PREFIX=%cd%/inst ^
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes ^
     .
-cmake --build . --config Release -j2 --target install
+cmake --build . --config Release --target install
--- a/build_release.sh
+++ b/build_release.sh
@@ -15,5 +15,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+-G Ninja \
     .
-cmake --build . -j16 --target install
+cmake --build . --target install


kmr-srbh:
The CI failed because it could not find `ccache`. How do I install it here?

gxyd:
I thin our environment file used in CI is `ci/environment_linux.yml`

certik:
Separate the ccache into a different PR, that should not be on by default, otherwise we have to add a dependency on it.

However we should switch to ninja by default.

gxyd:
I think you added `ccache` to just `environment_linux.yml`

certik:
If ccache really helps, we can use it, but we need to figure out a solution so that it is not on by default, we need to keep our dependencies down, and ccache is a big one and I think it can sometimes fail also.

kmr-srbh:
> If ccache really helps, we can use it, but we need to figure out a solution so that it is not on by default, we need to keep our dependencies down, and ccache is a big one

I completely agree that `ccache` introduces a dependency burden.

> I think it can sometimes fail also

Could you please elaborate on how it might fail sometimes?

gxyd:
> but we need to figure out a solution so that it is not on by default

This might help with that: https://theunixzoo.co.uk/blog/2023-10-24-llvm-ccache-distcc.html#:~:text=LLVM%20uses%20cmake%20as%20its,when%20you%20configure%20the%20build.

gxyd:
> Could you please elaborate on how it might fail sometimes?

Build's can be a bit tricky, I feel the more vanilla one keeps it's the more easier it is avoid issues resulting from cache corruption or disk issues and ofcourse debugging can be difficult with ccache.

I haven't personally worked with ccache, but considering the stats which I posted on [zulip thread](https://lfortran.zulipchat.com/#narrow/channel/452774-Internal/topic/Speeding.20up.20time.20to.20compile.20LFortran/near/483616894), I feel like atleast here, using ccache could be good.

kmr-srbh:
The CI fails due to these warnings:

```console
In file included from /home/runner/work/lfortran/lfortran/src/lfortran/tests/test_serialization.cpp:4:
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:43:19: warning: shift count >= width of type [-Wshift-count-overflow]
    bytes[0] = (i >> 56) & 0xFF;
                  ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:44:19: warning: shift count >= width of type [-Wshift-count-overflow]
    bytes[1] = (i >> 48) & 0xFF;
                  ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:45:19: warning: shift count >= width of type [-Wshift-count-overflow]
    bytes[2] = (i >> 40) & 0xFF;
                  ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:46:19: warning: shift count >= width of type [-Wshift-count-overflow]
    bytes[3] = (i >> 32) & 0xFF;
                  ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:91:31: warning: shift count >= width of type [-Wshift-count-overflow]
    return (((uintptr_t)p[0]) << 56) |
                              ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:92:31: warning: shift count >= width of type [-Wshift-count-overflow]
           (((uintptr_t)p[1]) << 48) |
                              ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:93:31: warning: shift count >= width of type [-Wshift-count-overflow]
           (((uintptr_t)p[2]) << 40) |
                              ^  ~~
/home/runner/work/lfortran/lfortran/src/libasr/bwriter.h:94:31: warning: shift count >= width of type [-Wshift-count-overflow]
           (((uintptr_t)p[3]) << 32) |
                              ^  ~~
8 warnings generated.
```

It is strange that the CI did not fail before with this warning. Is it related to this PR?

Thirumalai-Shaktivel:
I think it's not the warning, maybe the following:
```
FAILED: src/bin/lfortran.js 
: && /home/runner/ext/emsdk/upstream/emscripten/em++ -Wall -Wextra -fexceptions -g0 -Oz -fexceptions --preload-file asset_dir -Wall -Wextra --no-entry -s ASSERTIONS -s ALLOW_MEMORY_GROWTH=1 -s WASM_BIGINT -s "EXPORTED_RUNTIME_METHODS=['cwrap']" src/bin/CMakeFiles/lfortran.dir/lfortran.cpp.o -o src/bin/lfortran.js  src/lfortran/liblfortran_lib.a  src/libasr/libasr.a  src/runtime/legacy/liblfortran_runtime_static.a && :
error: asset_dir does not exist
em++: error: '/home/runner/ext/emsdk/upstream/emscripten/tools/file_packager src/bin/lfortran.data --from-emcc --preload asset_dir' failed (returned 1)
```

kmr-srbh:
@Thirumalai-Shaktivel what could be the reason for this? I see that it does not fail for the other PRs. Which change in this PR could be causing it?

build1.sh
```diff
@@ -11,8 +11,9 @@ cmake \
...
     -DWITH_LSP=no \
     -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH_LFORTRAN;$CONDA_PREFIX" \
-    -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
+    -DCMAKE_INSTALL_PREFIX=$(pwd)/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+-G Ninja \
     .
-cmake --build . -j16 --target install
+ninja -j16 install
```
certik, line 19: I think ninja runs in parallel by default, correctly using the number of CPUs on your computer:
```suggestion
ninja install
```

Furthermore, let's use the `cmake` to invoke it, like this:
```
cmake --build . --parallel --target install
```
I am not sure if the `--parallel` is needed.

kmr-srbh: We can use either `--parallel N` or `-jN`. Using them without `N` will lead to infinite jobs with `make`, with `ninja`, we use the max thread count for the CPU.

#### make
```console
-j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
```

#### ninja
```console
-j N     run N jobs in parallel (0 means infinity) [default=8 on this system]
```

kmr-srbh: On my system, the `--parallel` option causes the memory to fill rapidly and eventually hang the system.

build1.sh
```diff
@@ -14,5 +14,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+-G Ninja \
     .
-cmake --build . -j16 --target install
+ninja install
```
certik, line 19: There is no way to use `cmake --build` so that it creates an equivalent of `ninja install` which runs in parallel based on the number of cores that you have?

kmr-srbh: I found a way to do this using CMake environment variables. Pushing changes.

certik:
I just tested it, the following diff works:
```diff
--- a/build1.sh
+++ b/build1.sh
@@ -14,5 +14,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+    -G Ninja \
     .
-cmake --build . -j16 --target install
+cmake --build . --target install
```
Ninja runs in parallel by default. So the above is perfect, it will use all the cores that you have.

build1.sh
```diff
@@ -3,6 +3,8 @@
 set -e
 set -x
 
+export CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
+
```
certik, line 7: remove this

build1.sh
```diff
@@ -14,5 +16,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
     -DCMAKE_INSTALL_LIBDIR=share/lfortran/lib \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+-G Ninja \
     .
-cmake --build . -j16 --target install
+cmake --build . --target install --parallel
```
certik, line 21: remove `--parallel`

kmr-srbh: Okay. I am pushing the changes now.

certik (approved):

Thirumalai-Shaktivel:
**Note:** `make -j` doesn't work any more. We need to use `ninja` or `ninja -j16`

certik:
If you want, you can just modify the `cmake` line to use make, then `make -j` works.

Issue #5420, opened 2024-11-19
`--lookup-name` sometimes returns the location of a symbol declaration instead of its definition, and the wrong symbol name.


dylon:
For the test defined in `tests/function_call1.f90`, `lfortran --lookup-name` returns the location of the declaration for `eval_1d` instead of the location of its definition, and it consistently returns the wrong name (it always returns `a.first`).

```shell
./src/bin/lfortran --lookup-name --line=18 --column=22 tests/function_call1.f90 | jq
```
```json
[
  {
    "kind": 1,
    "location": {
      "range": {
        "start": {
          "character": 7,
          "line": 3
        },
        "end": {
          "character": 27,
          "line": 3
        }
      },
      "uri": "uri"
    },
    "name": "a.first"
  }
]
```

PR #5424, opened 2024-11-20
enh: `--lookup-name` to return symbol name

Pranavchiku:
Fixes the hardwired returning of `a.first`. Refer #5420

### DIFF
--- a/src/lfortran/tests/test_ast.cpp
+++ b/src/lfortran/tests/test_ast.cpp
@@ -144,5 +144,48 @@ end program
     }
 }
 
+TEST_CASE("Variable Location 2") {
+    Allocator al(4*1024);
+
+    std::string src = R"""(
+program expr2
+implicit none
+integer :: x
+x = (2+3)*5
+print *, x
+end program
+)""";
+
+    LCompilers::diag::Diagnostics diagnostics;
+    CompilerOptions compiler_options;
+    compiler_options.lookup_name = true;
+    compiler_options.line = "1";
+    compiler_options.column = "9";
+
+    LCompilers::LocationManager lm;
+    {
+        LCompilers::LocationManager::FileLocations fl;
+        fl.out_start0 = {};
+        fl.in_filename = "input.f90";
+        lm.files.push_back(fl);
+    }
+    FortranEvaluator e(compiler_options);
+    LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
+        r = e.get_asr2(src, lm, diagnostics);
+    ASR::asr_t* asr2 = e.handle_lookup_name(r.result, lm.linecol_to_pos(2, 12));
+    LCOMPILERS_ASSERT(ASR::is_a<ASR::symbol_t>(*asr2));
+    ASR::symbol_t* sym = ASR::down_cast<ASR::symbol_t>(asr2);
+    std::string symbol_name = ASRUtils::symbol_name(sym);
+    LCOMPILERS_ASSERT(symbol_name == "expr2");
+    std::vector<diag::Span> spans2 = diag::Label("", {asr2->loc}).spans;
+    for( auto it: spans2 ) {
+        populate_span(it, lm);
+        CHECK(it.first_line == 2);
+        CHECK(it.first_column == 1);
+        CHECK(it.last_line == 7);
+        CHECK(it.last_column == 11);
+    }
+}
+
 
 } // namespace LCompilers::LFortran
--- a/src/lfortran/tests/test_ast.cpp
+++ b/src/lfortran/tests/test_ast.cpp
@@ -3,7 +3,7 @@
 #include <sstream>
 
 #include <iostream>
-
+#include <libasr/lsp.cpp>
 #include <lfortran/fortran_evaluator.h>
 #include <libasr/codegen/evaluator.h>
 #include <libasr/alloc.h>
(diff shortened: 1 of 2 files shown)


src/libasr/lsp.cpp
```diff
@@ -251,7 +251,13 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
             LCompilers::ASR::asr_t* asr = fe.handle_lookup_name(x.result, pos);
             LCompilers::document_symbols loc;
             // TODO: make sure it returns us the symbol name
-            std::string symbol_name = "a.first";
+            std::string symbol_name;
+            try {
+                ASR::symbol_t* s = ASR::down_cast<ASR::symbol_t>(asr);
+                symbol_name = ASRUtils::symbol_name( s );
+            } catch (std::exception &e) {
```
certik, line 258: Just like in Python, never catch an exception like this.

In this case do not use exceptions at all. What is the failure mode here? We have to handle it explicitly.

What does `a.first` mean here? Why would we ever return it?

Also remove the TODO item.

certik:
Also we need to test the LSP in Fortran, right now it is not tested, or we don't have the test for this.

I would probably add a "ctest" test for this --- put the source code there, and check that the result has the proper name.

Pranavchiku:
> we don't have the test for this.

We test it via ctest

src/libasr/lsp.cpp
Pranavchiku: > What does a.first mean here? Why would we ever return it?

I'm not exactly sure if `asr` is symbol always. Hence had to do this, we can do if/else

Pranavchiku:
Now using `ASR::is_a` to test if it is symbol and then fetching symbol name.

src/libasr/lsp.cpp
certik: Then see `handle_lookup_name`, document what it should return, and you can put an assert inside it to ensure it is a symbol if it is not clear.

If it can be other things too, then you need to use `is_a` to check for a symbol, and return an error if it is not --- however I recommend to simply assert it is a symbol if it passes all tests.

Then if we discover a case where this fails, we can then figure out how to handle the error.

certik:
> We test it via ctest

No test was changed in this PR, which means that this path is not tested yet. Let's add a test.

src/libasr/lsp.cpp
```diff
@@ -251,7 +251,13 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
             LCompilers::ASR::asr_t* asr = fe.handle_lookup_name(x.result, pos);
             LCompilers::document_symbols loc;
             // TODO: make sure it returns us the symbol name
```
certik, line 253:
```suggestion
```

Pranavchiku:
If it is not symbol we shall not populate list and simply return.

certik:
> If it is not symbol we shall not populate list and simply return.

Yes, this is one way to handle the error properly. Let's do it.

certik (approved):
This is better. 

What does `handle_lookup_name` return if it is a

* symbol --- it returns the ASR node of the declaration?
* "=" (Assignment) --- it returns that assignment?

We can further refactor this in subsequent PRs.

However let's add a ctest for this, that fails in main, but now works. After that we can merge.

Pranavchiku:
> it returns the ASR node of the declaration?

Yes

> "=" (Assignment) --- it returns that assignment?

Yes

Pranavchiku:
> However let's add a ctest for this, that fails in main,

Yes, I am trying to figure that out. It will be straight forward.

src/lfortran/tests/test_ast.cpp
```diff
@@ -144,5 +144,48 @@ end program
...
+        lm.files.push_back(fl);
+    }
+    FortranEvaluator e(compiler_options);
+    LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
+        r = e.get_asr2(src, lm, diagnostics);
+    ASR::asr_t* asr2 = e.handle_lookup_name(r.result, lm.linecol_to_pos(2, 12));
+    LCOMPILERS_ASSERT(ASR::is_a<ASR::symbol_t>(*asr2));
+    ASR::symbol_t* sym = ASR::down_cast<ASR::symbol_t>(asr2);
+    std::string symbol_name = ASRUtils::symbol_name(sym);
+    LCOMPILERS_ASSERT(symbol_name == "expr2");
```
Pranavchiku, line 179: this will fail in main

certik: The test is in the right direction. Just make sure it fails in main, to ensure it is actually testing the fix in this PR.

src/lfortran/tests/test_ast.cpp
```diff
@@ -144,5 +144,37 @@ end program
...
+        LCompilers::LocationManager::FileLocations fl;
+        fl.out_start0 = {};
+        fl.in_filename = "input.f90";
+        lm.files.push_back(fl);
+    }
+#ifdef HAVE_LFORTRAN_RAPIDJSON
+    std::string json = get_definitions_source_code(src, compiler_options);
+    std::string expected_json = R"""([{"kind":1,"location":{"range":{"start":{"character":1,"line":1},"end":{"character":11,"line":6}},"uri":"uri"},"name":"expr2"}])""";
+    LCOMPILERS_ASSERT(json == expected_json)
+#endif
```
Pranavchiku, line 176: We won't need this after #5438

src/libasr/lsp.cpp
```diff
@@ -323,6 +326,105 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+            x = fe.get_asr2(input, lm, diagnostics);
+        if (x.ok) {
+            // populate_symbol_lists(x.result, lm, symbol_lists);
+            uint16_t l = std::stoi(compiler_options.line);
+            uint16_t c = std::stoi(compiler_options.column);
+            uint64_t pos = lm.linecol_to_pos(l, c);
+            LCompilers::ASR::asr_t* asr = fe.handle_lookup_name(x.result, pos);
+            LCompilers::document_symbols loc;
+            if (!ASR::is_a<ASR::symbol_t>(*asr)) {
+                return "{}";
```
dylon, line 354: If the return value of the function should be an array, then please return an empty array instead of an object:

```c++
return "[]";
```

src/libasr/lsp.cpp
```diff
@@ -323,6 +326,105 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+                last_column, filename);
+            loc.first_column = first_column;
+            loc.last_column = last_column;
+            loc.first_line = first_line-1;
+            loc.last_line = last_line-1;
+            loc.symbol_name = symbol_name;
+            loc.filename = filename;
+            symbol_lists.push_back(loc);
+        } else {
+            return "{}";
```
dylon, line 375: As mentioned above, please return an empty array instead of an object:

```c++
return "[]";
```

dylon (changes requested):
Other than a couple minor change requests, this looks good to me!

src/libasr/lsp.cpp
```diff
@@ -323,6 +326,105 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+        test_capture.AddMember("kind", rapidjson::Value().SetInt(1), allocator);
+        test_capture.AddMember("location", location_object, allocator);
+        test_capture.AddMember("name", rapidjson::Value().SetString(name.c_str(), allocator), allocator);
+        test_output.PushBack(test_capture, test_output.GetAllocator());
+    }
+    rapidjson::StringBuffer buffer;
+    buffer.Clear();
+    rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+    test_output.Accept(writer);
+    std::string resp_str( buffer.GetString() );
```
certik, line 422: This should just generate strings, so that we do not depend on rapidjson. Then we can enable LSP by default.

Pranavchiku:
closing with reference to https://github.com/lfortran/lfortran/pull/5456

PR #5425, opened 2024-11-20
fix: incorrect location of procedure calls

### LINKED ISSUE #5420: `--lookup-name` sometimes returns the location of a symbol declaration instead of its definition, and the wrong symbol name.

Pranavchiku:
Fixes #5420


certik (approved):
I think that this is fine.

PR #5438, opened 2024-11-21
enh: enable LSP by default

Pranavchiku:
Requested at https://github.com/lfortran/lfortran-lsp/issues/14#issuecomment-2489809495

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -2498,24 +2498,14 @@ int main_app(int argc, char *argv[]) {
     }
     lfortran_pass_manager.parse_pass_arg(arg_pass, skip_pass);
     if (compiler_options.lookup_name) {
-#ifdef HAVE_LFORTRAN_RAPIDJSON
         return get_definitions(arg_file, compiler_options);
-#else
-        return emit_asr(arg_file, lfortran_pass_manager,
-                compiler_options);
-#endif
     }
     if (show_asr) {
         return emit_asr(arg_file, lfortran_pass_manager,
                 compiler_options);
     }
     if (show_document_symbols) {
-#ifdef HAVE_LFORTRAN_RAPIDJSON
         return get_symbols(arg_file, compiler_options);
-#else
-        std::cerr << "Compiler was not built with LSP support (-DWITH_LSP), please build it again." << std::endl;
-        return 1;
-#endif
     }
 
     if (show_errors) {


Pranavchiku:
If this or #5424 gets merged then references needs to be updated in other PR else it will fail.

gxyd:
I tried to build locally on this branch, and I got the below error:
```console
Assuming default values of intrinsic_function_registry_util.h
+ cd src/lfortran
+ re2c -W -b parser/tokenizer.re -o parser/tokenizer.cpp
+ cd src/lfortran
+ re2c -W -b parser/preprocessor.re -o parser/preprocessor.cpp
+ cd src/lfortran/parser
+ bison -Wall -d -r all parser.yy
+ grep -n ''\''' src/lfortran/parser/parser.yy
+ echo OK
OK
-- Found LLVM 11.0.1
-- Using LLVMConfig.cmake in: /Users/gxyd/anaconda3/envs/lf/lib/cmake/llvm
CMake Error at /Users/gxyd/anaconda3/envs/lf/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find RapidJSON (missing: RAPIDJSON_HEADER)
Call Stack (most recent call first):
  /Users/gxyd/anaconda3/envs/lf/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindRapidJSON.cmake:4 (find_package_handle_standard_args)
  CMakeLists.txt:264 (find_package)

```

Seems like enabling LSP by default will put a dependency on rapidjson, and considering the discussion somewhere else, as we usually intend to keep the dependency list small, we would need to discuss this maybe?

gxyd:
Oops, I see it already added as a dependency in `environment_linux.yml` and also in `ci/environment.yml`, that's why the tests in this PR succeed.

gxyd (approved):
Looks good to me

src/libasr/lsp.cpp
```diff
@@ -8,17 +8,101 @@
...
+            }
+        }
+
+        void SetArray() {
+            value = "[";
+        }
+
+        void AddMember(std::string key, int v) {
+            // find `{` and add key value pair
+            int end_pos = value.find_last_of("}");
```
Pranavchiku, line 41: We can optimise it by tracking everytime we add `}`

src/libasr/lsp.cpp
```diff
@@ -8,17 +8,101 @@
...
+};
+
+class LFortranJSON {
+    public:
+        std::string value;
+        LFortranJSONType type;
+
+        LFortranJSON(LFortranJSONType type) : type(type) {}
+
+        void SetObject() {
```
certik, line 27: Let's use `set_object` style like in the rest of LFortran. The same for other methods here.

src/libasr/lsp.cpp
```diff
@@ -8,17 +8,101 @@
...
+            // find `[` and add value
+            int end_pos = value.find_last_of("]");
+            // If ']' is not found, then add value at the end and add ']'
+            // If ']' is found, then add `, value` before ']'
+            if (end_pos == -1) {
+                value.insert(value.length(), v.GetValue() + "]");
+            } else {
+                value.insert(end_pos, "," + v.GetValue());
+            }
+        }
```
certik, line 97: This implementation seems very fragile --- what happens if you have "}" or "]" inside strings in JSON?

Also this constant finding and insert is slow.

We should construct JSON just like we print ASR, just recursively construct all the strings in the right order. Then you never have to replace anything.

certik (approved):
I think this is great. It removes the dependency on rapidjson, which didn't see a release in a very long time, so we should not depend on it. It enables LSP by default, which is good.

I think the JSON implementation is not optimal, see my comments, but this can be fixed in subsequent PRs. 

I think this is good to merge as is.

Pranavchiku:
Opened https://github.com/lfortran/lfortran/issues/5455 for improving JSON

PR #5441, opened 2024-11-21
feat: introduce and implement `--rename-symbol`

### LINKED ISSUE #5419: Need command-line option to contextually rename symbols
For our LSP integration, we will need a command-line option to contextually rename symbols. It would best if we could rename them across a project rather than just the current file, but we must be able to rename them within at least the current file.

Support for this will require indexing a symbol's locations within the current file and possibly within its project. We may then use that same index for LSP's various symbol reference functionalities.

Requirements:
1. Need a command-line option, named `--rename-symbol` or the like.
2. It will accept the following parameters:
    1. `--line` := line on which the symbol to rename appears (accepts int param; must be within the respective range)
    2. `--column` := column on which the symbol to rename appears (accepts int param; must be within the respective range)
    3. `--new-name` (or something like that) := what to replace the symbol with (accepts string param; must be valid identifier)
    4. The final, positional parameter will be the path to the file whose symbol should be renamed (as is the current behavior).

Pranavchiku:
Fixes #5419. This PR is built on #5424, and #5438 hence CI shall fail. Once those are merged, we need to rebase this over top of latest main.


src/libasr/lsp.cpp
```diff
@@ -323,6 +328,94 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+            LCompilers::ASR::asr_t* asr = fe.handle_lookup_name(x.result, pos);
+            LCompilers::document_symbols loc;
+            if (!ASR::is_a<ASR::symbol_t>(*asr)) {
+                std::cout << "{}";
+                return 0;
+            }
+            // TODO: make sure it returns us the symbol name
+            std::string symbol_name = ASRUtils::symbol_name(ASR::down_cast<ASR::symbol_t>(asr));
+            LCompilers::LFortran::OccurenceCollector occ(symbol_name, symbol_lists, lm);
+            occ.visit_TranslationUnit(*x.result);
```
Pranavchiku, line 363: Here is the core logic, use line and column to compute `symbol_name` and then pass it to `OccurenceCollector` which will collect all occurrence.

src/libasr/lsp.cpp
```diff
@@ -323,6 +328,94 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+            x = fe.get_asr2(input, lm, diagnostics);
+        if (x.ok) {
+            // populate_symbol_lists(x.result, lm, symbol_lists);
+            uint16_t l = std::stoi(compiler_options.line);
+            uint16_t c = std::stoi(compiler_options.column);
+            uint64_t pos = lm.linecol_to_pos(l, c);
+            LCompilers::ASR::asr_t* asr = fe.handle_lookup_name(x.result, pos);
+            LCompilers::document_symbols loc;
+            if (!ASR::is_a<ASR::symbol_t>(*asr)) {
+                std::cout << "{}";
```
dylon, line 357: Lines 357 and 365 should return an empty array instead of an object:

```c++
std::cout << "[]";
```

src/libasr/lsp.cpp
```diff
@@ -323,6 +328,94 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+                return 0;
+            }
+            // TODO: make sure it returns us the symbol name
+            std::string symbol_name = ASRUtils::symbol_name(ASR::down_cast<ASR::symbol_t>(asr));
+            LCompilers::LFortran::OccurenceCollector occ(symbol_name, symbol_lists, lm);
+            occ.visit_TranslationUnit(*x.result);
+        } else {
+            std::cout << "{}";
```
dylon, line 365: As mentioned above, please return an empty array instead of an object when the return type should be an array:

```c++
std::cout << "[]";
```

dylon (changes requested):
With the exception of a couple minor changes, this looks good to me!

Pranavchiku:
closing with reference to https://github.com/lfortran/lfortran/pull/5458

PR #5442, opened 2024-11-21
chore: free memory allocated by `_lfortran_malloc`

kmr-srbh:
Part of #3770

### DIFF
--- a/tests/reference/run-continue_compilation_2-3499257.stderr
+++ b/tests/reference/run-continue_compilation_2-3499257.stderr
@@ -246,3 +246,5 @@ semantic error: Type mismatch in argument at argument (1); passed `real` to `int
     |
 152 |     print *, f(42.9)
     |                ^^^^ 
+free(): double free detected in tcache 2
+Aborted (core dumped)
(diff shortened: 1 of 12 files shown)


kmr-srbh:
With this PR, we free cases like:

```fortran
program main
   type t
      integer :: i
   end type t

   type(t), allocatable :: x
   allocate(x)
end program
```

tests/reference/run-continue_compilation_2-3499257.stderr
```diff
@@ -246,3 +246,5 @@ semantic error: Type mismatch in argument at argument (1); passed `real` to `int
     |
 152 |     print *, f(42.9)
     |                ^^^^ 
+free(): double free detected in tcache 2
+Aborted (core dumped)
```
kmr-srbh, line 250: I am not sure if this failure is correct. This was an incorrect code that ran because of `--continue-compilation`.

certik:
I think we have to call the finalization pass, insert proper deallocation, and in the backend just call it. The backend should not have logic when things get deallocated.

kmr-srbh:
> I think we have to call the finalization pass, insert proper deallocation, and in the backend just call it. The backend should not have logic when things get deallocated.

Okay, let's keep only the freeing logic here and collect the variables in the pass. Closing this.

PR #5447, opened 2024-11-21
Parser: error at the statement level

certik:
Now we have just one error recovery at the statement level, which is easier to mmaintain and it is able to print AST, since we simply skip the statement.

This provides an improvement over the initial implementation in https://github.com/lfortran/lfortran/pull/4774. The test introduced there still works.

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -681,7 +681,6 @@ endinterface
     | endinterface0 KW_OPERATOR "(" operator_type ")"
     | endinterface0 KW_OPERATOR "(" "/)"
     | endinterface0 KW_OPERATOR "(" TK_DEF_OP ")"
-    | error
     ;
 
 endinterface0
@@ -723,7 +722,6 @@ endenum
 enum_var_modifiers
     : %empty { LIST_NEW($$); }
     | var_modifier_list { $$ = $1; }
-    | error { LIST_NEW($$); }
     ;
 
 derived_type_decl
@@ -757,7 +755,6 @@ require_decl
 unit_require_plus
     : unit_require_plus "," unit_require { $$ = $1; LIST_ADD($$, $3); }
     | unit_require { LIST_NEW($$); LIST_ADD($$, $1); }
-    | error { LIST_NEW($$); }
     ;
 
 unit_require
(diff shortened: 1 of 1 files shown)


certik:
Later we still have to add `error` into a list of statements in if statements, otherwise it will delete the otherwise good if statement if there is a single syntax error in the "then" or "else" branches, but for now it's good enough.

src/lfortran/parser/parser.yy
```diff
@@ -1635,6 +1612,7 @@ sep_one
 decl_statements
     : decl_statements decl_statement { $$ = $1; LIST_ADD($$, $2); }
     | %empty { LIST_NEW($$); }
+    | decl_statements error sep_one { $$ = $1; }
```
certik, line 1615: Right now this just skips the statement. Later, we can instead insert a `StatementError` statement AST node, and insert the location information, which might be useful for LSP, we can for example try to parse this statement with some heuristics. For now I think discarding is a good way to get started.

certik:
The current test for multiple syntax errors still works:
```console
$ lfortran tests/multi_error1.f90 --continue-compilation
 --> tests/multi_error1.f90:3:15
  |
3 | integer :: x =
  |               ^ 

 --> tests/multi_error1.f90:4:15
  |
4 | integer :: y =
  |               ^ 

 --> tests/multi_error1.f90:5:13
  |
5 | print *, x +
  |             ^ 

syntax error: Token 'y' (of type 'identifier') is unexpected here
 --> tests/multi_error1.f90:6:12
  |
6 | print *, x y
  |            ^ 


clang: error: no such file or directory: 'tests/multi_error1.tmp.o'
The command 'clang -o multi_error1.out tests/multi_error1.tmp.o  -L"/Users/ondrej/repos/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/ondrej/repos/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
```

Pranavchiku:
Woah, this is cool. Thanks for doing this!

src/lfortran/parser/parser.yy
gxyd: This is informative, I usually prefer to have this kind of useful information in the form of a comment somewhere, but this looks good.

gxyd (approved):
Looks good to me.

Issue #5093, opened 2024-10-18
set allocation size not set for IntrinsicElementalFunction 'Sqrt'


gxyd:
## MRE

```f90
program array_operations
    implicit none
  
    ! Declare variables
    integer, dimension(5) :: array
    real :: scalar_value
    integer :: i
    real, dimension(5) :: x, y, z

    array = (/ 1, 0, 3, 0, 5 /)
    x = (/ 1.0, 2.0, 3.0, 4.0, 5.0 /)
    y = (/ 2.0, 4.0, 6.0, 8.0, 10.0 /)
    
    scalar_value = 2.5
    
    print *, "Result:", square(scalar_value)

    z = array_function(x, y)
    print *, "Result (z):", z

    where (array == 0)
      array = 10
    end where

    do i = 1, size(array)
      print *, "array(", i, ") squared =", square(real(array(i)))
    end do

  contains

    real function square(a)
      real, intent(in) :: a
      square = a * a
    end function

    real function array_function(a, b)
      real, intent(in) :: a(:), b(:)
      real :: temp(size(a))
      array_function = 0.0

      temp = a + sqrt(b)
      array_function = temp(0)
    end function
  
end program array_operations

```

## Output with *simplifier_pass* branch

On execution of the above fortran program on *simplifier_pass* branch:
```console
> lfortran test.f90
LCOMPILERS_ASSERT failed: /Users/gxyd/OpenSource/lfortran/src/libasr/pass/simplifier.cpp
function set_allocation_size(), line number 595 at 
false
ERROR MESSAGE:
ASR::IntrinsicElementalFunctions::Sqrt not handled yet in set_allocation_size
[1]    98858 abort      /Users/gxyd/OpenSource/lfortran/src/bin/lfortran test1.f90
```

## Output with current *main*

while it's execution with current main with commit: `474790bc55619460146d8e8ffc5d8ad36897705f` as the top, I get the correct output:
```console
> lfortran test.f90
Result:    6.25000000e+00
Result (z):    0.00000000e+00    0.00000000e+00    0.00000000e+00    0.00000000e+00    0.00000000e+00
array(    1    ) squared =    1.00000000e+00
array(    2    ) squared =    1.00000000e+02
array(    3    ) squared =    9.00000000e+00
array(    4    ) squared =    1.00000000e+02
array(    5    ) squared =    2.50000000e+01
```

HarshitaKalani:
```diff
$ git diff
[...]

PR #5452, opened 2024-11-22
DEV: Fix for Assignment to scalar expression

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/5093


integration_tests/arrays_60.f90
```diff
@@ -0,0 +1,45 @@
+program arrays_60
+    implicit none
+
+    ! Declare variables
+    integer, dimension(5) :: array
+    real :: scalar_value
+    integer :: i
+    real, dimension(5) :: x, y, z
+
+    array = (/ 1, 0, 3, 0, 5 /)
```
certik, line 10: Use `[]`:
```suggestion
    array = [ 1, 0, 3, 0, 5 ]
```
This is minor, you can fix it after merging, or even let be. But we should use `[]`.

PR #5458, opened 2024-11-23
feat: introduce and implement `--rename-symbol`

### LINKED ISSUE #5419: Need command-line option to contextually rename symbols

Pranavchiku:
Fixes #5419


src/libasr/lsp.cpp
```diff
@@ -9,7 +9,7 @@
 #include <libasr/location.h>
 #include <lfortran/fortran_evaluator.h>
 #include <libasr/lsp_interface.h>
-
+#include "libasr/asr_lookup_name.cpp"
```
certik, line 12: What is going on here? We should not be including `.cpp` files.

src/libasr/lsp.cpp
```diff
@@ -391,5 +391,88 @@ int get_definitions(const std::string &infile, LCompilers::CompilerOptions &comp
...
+
+        range_object.SetObject();
+
+        start_detail.SetObject();
+        start_detail.AddMember("character", start_character);
+        start_detail.AddMember("line", start_line);
+        range_object.AddMember("start", start_detail);
+
+        end_detail.SetObject();
+        end_detail.AddMember("character", end_character);
```
certik, line 457: Let's call this `column`?

src/libasr/lsp.cpp
Pranavchiku: We need this for `OccurenceCollector` visitor. It is implemented in `libasr/asr_lookup_name.cpp`

src/libasr/lsp.cpp
Pranavchiku: We can, but then we need to change handling at `lfortran-lsp`

certik: Ok. We can do that later.

src/libasr/lsp.cpp
certik: Ok.

Pranavchiku:
Is this good to merge?

src/libasr/lsp.cpp
certik: Can you implement it in a header file? I think it's bad form to be including `.cpp` files. Maybe you can just rename `asr_lookup_name.cpp` to `asr_lookup_name.h`.

certik (approved):
I think that's fine. But if you can fix the `cpp` issue based on my comment above, that would be great. You can do it in a subsequent PR.

src/libasr/asr_lookup_name.h
```diff
@@ -0,0 +1,140 @@
+#include <iostream>
```
certik, line 1: You have to put guards into the header file.

certik:
The error is just a temporary error I think:
```
  Transaction starting
  critical libmamba Download error (28) Timeout was reached [https://conda.anaconda.org/conda-forge/osx-arm64/llvm-tools-11.1.0-hfa12f05_5.tar.bz2]
      Operation too slow. Less than 30 bytes/sec transferred the last 60 seconds
  Error: The process '/Users/runner/micromamba-bin/micromamba' failed with exit code 1
```

certik (approved):

certik (approved):
I think this looks great, thanks!

PR #5466, opened 2024-11-23
Break out type_spec grammar rules

### LINKED ISSUE #5426: Syntax vs. Semantic error
There are a set of error tests that look something like:
```f90
program kind2
implicit none

real(*) a

end program
```
which currently returns
```
semantic error: Expected initialization expression for kind
 --> tests/errors/kind2.f90:4:6
  |
4 | real(*) a
  |      ^ 
```
I'm working on tightening up the parser in places, and, according to the standard, this statement is _syntactically_ invalid, not semantically.  With rules like the standard's _kind-selector_ in place, I get a parsing error of:
```
syntax error: Token ')' is unexpected here
 --> tests/errors/kind2.f90:4:7
  |
4 | real(*) a
  |       ^ 
```
since `*` is not a valid _scalar-int-constant-expr_.  How do people feel about this sort of change?  I'm tagging @gxyd as they put in the `kind_star_of_int` test that first caught this change.

### LINKED ISSUE #5464: Parser requires kind on data-implied-do
Attempting to compile:
```f90
program foo
  integer, parameter :: n = 4
  real, dimension(n) :: arr
  data (arr(i), integer :: i = 1,n) /0.0, 0.0, 0.0, 0.0/
  print *, arr
end program foo
```
leads to 
```
 --> data.f90:4:30
  |
4 |   data (arr(i), integer :: i = 1,n) /0.0, 0.0, 0.0, 0.0/
  |                              ^ 
```
Because the `integer_type` Bison rule requires `integer "(" kind_arg_list ")"`.  This is related to #4033.

pjh40:
This PR separates the rules _declaration-type-spec_, _intrinsic-type-spec_, and _integer-type-spec_ into their own Bison rules for consistent  use in other parts of the grammar.  For example `integer_type_spec` is now used in the implied-do `data_object`, and `declaration_type_spec` can be used in IMPLICIT statements.  This does _not_ affect the generalized parsing of KIND attributes.

This closes #5464.
This close #5426.


certik:
Thanks! Looks like it's getting:
```
syntax error: Internal Compiler Error: syntax is ambiguous in the parser
 --> /Users/runner/work/lfortran/lfortran/integration_tests/types_19.f90:2:3
  |
2 |   type(integer) :: a
  |   ^^^^^^^^^^^^^ 
```

pjh40:
That's so weird... I swear that I ran all the tests successfully! But it reproduces for me as well, will fix.  Sorry!

pjh40:
Okay, I must be missing some testing step.  I was doing:
```
ctest
./run_tests.py
```
but the `types_19` error looks like it is coming from some other testing. @certik what am I missing?

certik:
The error comes from integration tests:
```
cd integration_tests
./run_tests.py -j16
```
We have a large test suite there that we compile end do end with various backends and options there.

certik (approved):
I think this looks great, thanks for this. Let's add tests from the issues that it closes, let's put them into integration_tests.

PR #5468, opened 2024-11-23
Finalization : deallocate allocatables

assem2002:
starting out by freeing strings for the most simple case (Intent-local-allocatable variable pushed at end of program and functions )
The main goal here is to free allocatable `strings` as they don't get freed which causes memory leaks

### DIFF
--- a/src/libasr/pass/insert_deallocate.cpp
+++ b/src/libasr/pass/insert_deallocate.cpp
@@ -25,7 +25,7 @@ class InsertDeallocate: public ASR::CallReplacerOnExpressionsVisitor<InsertDeall
             for( auto& itr: x.m_symtab->get_scope() ) {
                 if( ASR::is_a<ASR::Variable_t>(*itr.second) &&
                     ASR::is_a<ASR::Allocatable_t>(*ASRUtils::symbol_type(itr.second)) &&
-                    ASRUtils::is_array(ASRUtils::symbol_type(itr.second)) &&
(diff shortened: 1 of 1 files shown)


assem2002:
Maybe we need a reference?

assem2002:
This would deallocate scalar variable such as `integer, allocatable :: num`. Should we avoid that? It's totally unnecessary.

certik:
The scalar allocatable variables like `integer, allocatable :: x` don't have to be allocatable. But if they are, we have to free them, since they are allocated on a heap.

assem2002:
for now I think we don't allocate on heap it's done on the stack.

assem2002:
This change would fail for test like that 
```fortran
module a
    contains
subroutine ff(ref) 
    integer :: ref
    character(:) ,allocatable, save :: f
    if(ref == 1) then
        print *, allocated(f)
        f = "hi"
    else 
        print *, allocated(f)
        f = f // ", there"
    end if
    print *, f
end subroutine ff
end module a
program b
    use a 
    integer :: i
    call ff(1)
    call ff(2)
end program 
```
is it imports deallocation at the end though the variable is `save` storage.

assem2002:
I'm closing that and pushing cleaner PR

assem2002:
@certik Do we need a reference test to show we deallocate allocatable-strings for sure?

certik:
I think it would be good to have a reference tests. I can't think of another good way to test it.

PR #5476, opened 2024-11-24
feat: add error tests with continue-compilation flag

HarshitaKalani:
Towards: #5036 
With this, we now have template_error_*.f90 tests left which I think will need to be handled separately.


parth121101 (approved):
Thanks @HarshitaKalani

parth121101 (commented):
@HarshitaKalani can we add this tests in continue_compilation_1.f90 file. I think this file was created by you in your previous PR.

certik (approved):
I think it's fine. Later we can merge `continue_compilation_*` files into just one or fewer files.

PR #5483, opened 2024-11-25
Finalization : Finalize with return exit

assem2002:
- insert `implicitDeallcoate` node in `program`, `function`, `block`, `if`, `doWhile`, `do` bodies.
- create a stack to hold the current implicit node that needs to be inserted by if,doWhile, do visitors, nested functions.
- Create visitors for if, doWhile, do to look up for insider returns or exits.
- Create some functions for redundant usage.
- make sure avoid unneceessary insertion for the implicitDeallocate node if we encounter a dead code.

### DIFF
--- a/src/libasr/pass/insert_deallocate.cpp
+++ b/src/libasr/pass/insert_deallocate.cpp
@@ -4,6 +4,7 @@
 #include <libasr/asr_utils.h>
 #include <libasr/asr_verify.h>
 #include <libasr/pass/insert_deallocate.h>
+#include<stack>
 
 
 namespace LCompilers {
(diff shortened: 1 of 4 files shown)


assem2002:
I don't know if using a stack to hold the needed `implicitDeallocate` statement that needs to be inserted by if, doWhile, Do -- is considered a good thing to use.

certik:
We probably can use Vec instead of stack? I am hoping to reduce dependency on C++ stdlib, to speedup compilation.

certik (approved):
I think this is good, we can further improve it.

PR #5486, opened 2024-11-25
enh: free allocatable `StructInstanceMember`

kmr-srbh:
Part of #3770 

We now free allocatables inside `StructType` like:

```fortran
program main
    type my_type
        integer, allocatable :: int(:)
    end type

    type(my_type) :: x

    allocate(x%int(2))
end program
```

Along with this, we also free other scalar allocatables like `integer, allocatable :: int`.


src/libasr/pass/insert_deallocate.cpp
```diff
@@ -78,32 +76,44 @@ class InsertDeallocate: public ASR::CallReplacerOnExpressionsVisitor<InsertDeall
...
+                        if (ASR::is_a<ASR::StructInstanceMember_t>(*allocate_stmt->m_args[i].m_a)) {
+                            this->visit_expr(*allocate_stmt->m_args[i].m_a);
+                            ASR::StructInstanceMember_t* s = ASR::down_cast<ASR::StructInstanceMember_t>(allocate_stmt->m_args[i].m_a);
+                            if (ASRUtils::is_allocatable(ASRUtils::symbol_type(s->m_m))
+                                && ASRUtils::symbol_StorageType(ASRUtils::symbol_get_past_external(s->m_m)) != ASR::storage_typeType::Save) {
+                                default_storage_variables_to_deallocate.push_back(al, allocate_stmt->m_args[i].m_a);
+                            }
+                        }
+                    }
+                }
```
kmr-srbh, line 105: This is the new change.

kmr-srbh:
I feel the logic of collecting variables can be improved. I will try to look into it.

src/libasr/pass/insert_deallocate.cpp
certik: How does this work --- will it deallocate the members when the struct goes out of scope, even inside functions (not just the main program)?

certik:
Also add a test for a struct in a function and call the function multiple times; let's ensure it gets deallocated each time.

src/libasr/pass/insert_deallocate.cpp
kmr-srbh: With the current design we can only free for the program scope, but after #5483 is merged, we can also free at the end of other scopes like `function`. I think we can improve the design further to make it easier to handle the scopes.

certik:
Rebase it on top of main, get the CI to work. Let's ensure the leak is fixed via valgrind. Then we can merge, and further improve with subsequent PRs.

kmr-srbh:
The current failure is due to this case:

```fortran
program class_11
    implicit none
    type t_1
        character(:), allocatable :: beta
    end type t_1

    type t
        class(t_1), allocatable :: alpha
    end type t

    type(t) :: gamma
    allocate(gamma%alpha)
    allocate(character(4) :: gamma%alpha%beta)

    ! Added manually to create MRE
    deallocate(gamma%alpha%beta)
    deallocate(gamma%alpha)
end program class_11
```

We have 2 things here:
1. We have a bug with handling `deallocate(gamma%alpha%beta)` when `alpha` is `ClassType`. This is unrelated to changes here.
2. We need to figure out the logic to deallocate implicitly exactly in the order above as `gamma%alpha%beta` is dependent on `gamma%alpha`. We need to figure out the correct dependency.

1 came up now when we tried to deallocate it implicitly. For now, we can disable this test, fix 1 and then work on 2.

certik:
Yes, you can disable the test if all 3rd party codes work. If they fail, then we need to skip this case inside our lfortran code.

assem2002:
For struct case condition. We have to change function `get_allocatable_default_storage_variables()` in `insert_deallocate.cpp` to do one more step and to test for struct variable, then it should call a function that recursively wraps struct members in `StructInstanceMember` node.
So a change in   `get_allocatable_default_storage_variables()` to detect struct variables + function with similar signature as this ` Vec<ASR::expr_t*> fetch_allocatables_in_structs(ASR::struct_t* ss)` it will call itself to create such expression `aa%bb%cc%dd` then you deallocate them all

kmr-srbh:
Yes @assem2002. I think you did something similar for SNAP. I will try to take a look into that code and make progress. :+1:

kmr-srbh:
Ready.

kmr-srbh:
Some failures occur, I will take a look into it tomorrow morning (IST).

kmr-srbh:
The failure is caused due to cases of nested `allocatable` derived types. This is a bit involved. Working on it.

kmr-srbh:
The failure in `derived_types_32.f90` will be fixed using #5512.

kmr-srbh:
The current failure is due to the `simplifier` pass. I will take a look into this today and try to wrap up this PR.

PR #5485, opened 2024-11-25
feat: introduce and implement `--semantics-only`

Pranavchiku:
Fixes https://github.com/lfortran/lfortran/issues/5480

PR #5499, opened 2024-11-26
test: do not check no llvm for semantics only

Pranavchiku:
Fixes https://github.com/lfortran/lfortran/pull/5485#discussion_r1858862450


certik (approved):
Thanks!

PR #5503, opened 2024-11-26
CMake: use libunwind automatically for LLVM >= 12

certik:
And only on linux. This allows unified treatment of the unwind dependency across all platforms: on linux and LLVM >= 17 we install it, and it gets automatically picked up, on other platforms or LLVM < 17 we do not install it and do not pick it up.

This was initially introduced in https://github.com/lfortran/lfortran/pull/5031.


certik:
Looks like this is required already for LLVM >= 12.

PR #5504, opened 2024-11-26
Build on Windows using pixi and shell

certik:


certik:
Usage:
```
pixi r -e llvm11 build
```
Currently it gets far in compilation but fails with:
```
C:\Users\ondrejcertik\repos\lfortran\src\lfortran\semantics\ast_body_visitor.cp
p(1,1): error C1128: number of sections exceeded object file format limit: comp
ile with /bigobj [C:\Users\ondrejcertik\repos\lfortran\src\lfortran\lfortran_li
b.vcxproj]
```

PR #5489, opened 2024-11-25
Add environments to pixi

certik:

PR #5505, opened 2024-11-27
Add environments to pixi.toml

certik:
The CI works, but I left it disabled for now to save runners. Once we can replace some existing jobs, then we'll switch.

This ports changes from #5489, except `pixi.lock`, since it's a huge file.


certik:
This doesn't touch any existing functionality, so we'll merge it.

PR #5509, opened 2024-11-27
Use Warnings instead of Errors

Thirumalai-Shaktivel:


Thirumalai-Shaktivel:
This was required for godbolt PR.

They enable -g option by default, currently I made the changes to overwrite it. But, in future we need to add it.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -367,11 +367,9 @@ Result<std::unique_ptr<LLVMModule>> FortranEvaluator::get_llvm3(
                 "The `emit_debug_line_column` is not enabled; please use the "
                 "`--debug-with-line-column` option to get the correct "
                 "location information",
-                LCompilers::diag::Level::Error,
+                LCompilers::diag::Level::Warning,
```
Thirumalai-Shaktivel, line 370: This is fine. I can add the option `--debug-with-line-column` to be default

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -353,8 +353,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                 break;
             }
-            default : throw LCompilersException("Debug information for the type: `"
-                + ASRUtils::type_to_str_python(t) + "` is not yet implemented");
+            default:
+                diag.add(diag::Diagnostic("Debug information for the type: `"
+                    + ASRUtils::type_to_str_python(t) +
+                    "` is not yet implemented",
+                    diag::Level::Warning, diag::Stage::CodeGen, {}
+                ));
```
Thirumalai-Shaktivel, line 361: But, this has to be a Warning. Because godbolt would be useless until we support all the cases.

I don't know how much time would take to complete this.
Maybe for now we can have warning and change to error when most of cases are handled.

src/bin/lfortran.cpp
```diff
@@ -1056,14 +1056,11 @@ int compile_src_to_object_file(const std::string &infile,
     if (compiler_options.emit_debug_info) {
 #ifndef HAVE_RUNTIME_STACKTRACE
         diagnostics.add(LCompilers::diag::Diagnostic(
-            "The `runtime stacktrace` is not enabled. To get the stack traces "
-            "or debugging information, please re-build LFortran with "
-            "`-DWITH_RUNTIME_STACKTRACE=yes`",
-            LCompilers::diag::Level::Error,
+            "The `runtime stacktrace` is not enabled. To get the stack traces, "
+            "please re-build LFortran with `-DWITH_RUNTIME_STACKTRACE=yes`",
```
Thirumalai-Shaktivel, line 1060: I had some issues in making `llvm-dwarfdump` to be available for lfortran.

Or maybe we can have an option to disable it.
In the meantime, I will try to make llvm-dwarf-dump available

src/lfortran/fortran_evaluator.cpp
certik: Do you remember why we added this separate option?

src/bin/lfortran.cpp
certik: I see, the `-g` does not work unless `WITH_RUNTIME_STACKTRACE` is enabled, right?

certik: If we change it to a warning, then we'll be getting this warning with every compilation right? So it's not really useful. We need to build lfortran with stacktrace support.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, warning is fine here.

src/lfortran/fortran_evaluator.cpp
Thirumalai-Shaktivel: We made a decision here, https://github.com/lcompilers/lpython/pull/1240#discussion_r1008479887.
Basically, it's because of the performance.

For generating the line and column information, we need to read the source code and compute the information, which is slower.

Linear location is faster.

src/bin/lfortran.cpp
Thirumalai-Shaktivel: Yes, I thought the same.

So, here is the problem: there is no binaries available in the path, where the lfortran binary resides and it's not able to find the llvm-dwarfdump.
Maybe we can fix this, as we do the clang. We need to use some env variables, `export LFORTRAN_CC=<path>/clang`.

Do you think of any other solutions?

src/lfortran/fortran_evaluator.cpp
certik: I see. However the debugging tools can't work with it anyway, so if we use `-g`, then we probably should use the slow way by default.

src/bin/lfortran.cpp
certik: Rather, we need to implement everything in C++, not call an external program that might or might not be available.

Currently, we do:
```c++
            cmd += "dsymutil " + file_name + ".out && llvm-dwarfdump --debug-line "                                      
                + file_name + ".out.dSYM > ";       
#else                                             
            cmd += "llvm-dwarfdump --debug-line " + file_name + ".out > ";
#endif                                    
            std::string dwarf_scripts_path = LCompilers::LFortran::get_dwarf_scripts_dir();                              
            cmd += file_name + "_ldd.txt && (" + dwarf_scripts_path + "/dwarf_convert.py "                                 
                + file_name + "_ldd.txt " + file_name + "_lines.txt "
                + file_name + "_lines.dat && " + dwarf_scripts_path + "/dat_convert.py "                                      
                + file_name + "_lines.dat)";
```
Which creates the file `lines.dat` with debugging information in a form that we can use. This is obviously fragile, as it requires our Python script and `llvm-dwarfdump`. And btw, `llvm-dwarfdump` recently changed how it prints the output (https://github.com/lfortran/lfortran/pull/5497), so our script doesn't even work with every version.


So I would say, let's keep `WITH_RUNTIME_STACKTRACE` off by default and for godbolt. We'll enable it when we have a pure C++ implementation of the above.

src/lfortran/fortran_evaluator.cpp
Thirumalai-Shaktivel: Exactly.

src/bin/lfortran.cpp
Thirumalai-Shaktivel: > We'll enable it when we have a pure C++ implementation of the above.

You mean, not using the `dwarf_convert.py` and `llvm-dwarfdump`?

certik: Yes, we should implement some C++ solution that doesn't require to call external tools like these.

PR #5512, opened 2024-11-28
enh : Get subroutine_from_function pass to handle strings (right owner)

assem2002:
Towards #2257 (https://github.com/lfortran/lfortran/issues/2257#issuecomment-2041693935) + towards finalization
This handles the descriptorStrings `(character(:), alloctable :: str)` to transform the following code :
```fortran 
program p
    print *, func_str()    

    contains 

    function func_str() result(ret_str)
        character(:), allocatable :: ret_str
    end function
end program  
```
to 
```fortran
program p
implicit none
character(len=:, kind=1), allocatable :: __libasr__created__var__0__func_call_res
call func_str(__libasr__created__var__0__func_call_res)
print *, __libasr__created__var__0__func_call_res
deallocate(__libasr__created__var__0__func_call_res) ! Implicit deallocate

contains

subroutine func_str(ret_str)
    character(len=:, kind=1), allocatable, intent(out) :: ret_str
end subroutine func_str

end program p
```

I think we should do the same with normal strings (pointerString) `character(10) :: str` as it's being heap allocated so it needs to have proper owner then later in `insert_deallocate` pass we can properly deallocate it.

### DIFF
--- a/integration_tests/operator_overloading_07.f90
+++ b/integration_tests/operator_overloading_07.f90
@@ -1,0 +1,50 @@
+! Overloading the assignment operator `=` with a function call that would be handled by
+! `subroutine_from_function` pass. So we need to make sure that the pass 
+! actually replaces the function call in the overloaded expression.
+module operator_overloading_07_mod1
+    implicit none
+    interface assignment(=)
+        module procedure :: assign_string_char
+    end interface assignment(=)
+
+    type string_type
+        character(:), allocatable :: raw
+    end type
+
+    contains
+
+    subroutine assign_string_char(lhs, rhs)
+        type(string_type), intent(inout) :: lhs
+        character(len=*), intent(in) :: rhs
+        lhs%raw = rhs
+    end subroutine assign_string_char
+
+end module operator_overloading_07_mod1
+
+
+module operator_overloading_07_mod2
+    use operator_overloading_07_mod1
+    implicit none
+
+    type(string_type) :: instance
+    
+    contains
+    function func() result(ret_arr)
+        character(:), allocatable :: ret_arr
+        allocate(character(15) :: ret_arr)
+        ret_arr = "Hello World!"
+    end function
+
+    subroutine ff()
+        character(:), allocatable :: inp
+        instance = func() ! pass `subroutine_from_function` takes action here. It must handle the overloading of `=`. 
+        print *, instance%raw
+    end subroutine ff
+end module
+
+
+program operator_overloading_07
+    use operator_overloading_07_mod2, only : ff
+    implicit none
+    call ff()
+end program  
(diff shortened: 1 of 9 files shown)


assem2002:
the `simplifier` path fails due to that it uses special version of subroutine_from_function -- `subroutine_from_function_simplifer.cpp`, This pass doesn't have the logic that exists in normal `subroutine_from_function.cpp` as it depends on the `simplifier.cpp` to simplify the things beforehand.
I'll try to port the logic from `subroutine_from_function.cpp` to --> `subroutine_from_function_simplifer.cpp` for just strings.

assem2002:
The CI should pass after fixing this issue https://github.com/lfortran/lfortran/issues/5564

assem2002:
I tried to make the commits descriptive as much as I can. now we can appropriately free allocatable strings that are `returnVar`. 
If a reference test is needed please inform me.

certik:
The CI fails with:
```
[ 41%] Building Fortran object test/string/CMakeFiles/test_string_strip_chomp.dir/test_string_strip_chomp.f90.o
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/../libasr/asr_utils.h
function Call_t_body(), line number 5973 at 
ASRUtils::check_equal_type(arg_type, orig_arg_type)
ERROR MESSAGE:
ASRUtils::check_equal_type(string_type, str)
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/../libasr/asr_utils.h
function Call_t_body(), line number 5973 at 
ASRUtils::check_equal_type(arg_type, orig_arg_type)
ERROR MESSAGE:
ASRUtils::check_equal_type(string_type, str)
Aborted (core dumped)
make[2]: *** [test/string/CMakeFiles/test_string_match.dir/build.make:75: test/string/CMakeFiles/test_string_match.dir/test_string_match.f90.o] Error 134
```

integration_tests/operator_overloading_07.f90
```diff
@@ -0,0 +1,51 @@
+! Overloading the assignment operator `=` with a function call that would be handled by
+! `subroutine_from_function` pass. So we need to make sure that the pass 
+! actually replaces the function call in the overloaded expression.
+module operator_overloading_07_mod1
+    implicit none
```
assem2002, line 5: It's not relevant test to what the PR is trying to do, rather it's a test to make sure that overload expr in `assignment_t` node is actually replaced correctly. as it took me a while to figure out that I forget about the case of a replaceable functionCall in the `overload` expr.

assem2002:
Ready for review.

certik (approved):
I think this looks good, it moves things in the right direction. 

The deallocate statement should be inserted at the end via "insert_deallocate", but we'll get to it later.

assem2002:
We already insert the deallocation node for allocatable strings .

PR #5514, opened 2024-11-29
TESTS: Remove `NO_EXPERIMENTAL_SIMPLIFIER` flag from working tests

czgdp1807:


czgdp1807:
I don't know what's wrong here. With `arrays_33` and LLVM 19 things work fine on my Linux machine. Here its failing.

czgdp1807:
Fixed the bug. :-)

czgdp1807:
Like this, the last few bugs will be debugged directly on CI if they cannot be reproduced locally. Will try to run the CI minimally. We should have `[skip ci]` functionality enabled so that we can skip running CI for draft PRs. Also, we can do `[skip job_name]` to skip jobs which we don't want to run. This is there in SciPy. We can have it here as well. Not urgent but a suggestion.

Thirumalai-Shaktivel:
Yes, this feature is very much required.

czgdp1807:
Yes. Something like, `[skip llvm19] [skip llvm10] .. ` in the commit message. So that we can debug on our own. In fact, in SciPy we can run CI on our own fork. I am not sure if we can do that in LFortran. If not then that should be supported as well.

certik:
You can modify the CI.yml to adjust it as needed for debugging.

czgdp1807:
Yes. That works too and is quick. Will do it for now.

PR #5517, opened 2024-11-29
Use Ninja as a build system

certik:


gxyd:
I see that these changes are present in our repository already, maybe we don't need this PR anymore?

gxyd:
I'll convert this to *draft mode* for now, and you can let me know if anything else is remaining here. As I see that we already use ninja as our build system now.

certik:
Yes, we merged another PR for that already.

PR #5523, opened 2024-11-29
Allow parser to accept `IMPLICIT NONE ()`

### LINKED ISSUE #5522: Empty parens not allowed on IMPLICIT NONE
Trying to compile
```f90
subroutine foo
  implicit none ()
end subroutine foo
```
with `lfortran --show-ast` erroneously results in:
```
syntax error: Token ')' is unexpected here
 --> implicit.f90:2:18
  |
2 |   implicit none ()
  |                  ^ 
```
R866 of the F23 draft gives the grammar as `IMPLICIT NONE [ ( [` _implicit-none-spec-list_  `] ) ]`, so an empty pair of parens should be accepted, and treated the same as if no parens were given (or `IMPLICIT NONE (TYPE)`).

pjh40:
Simple parser change to allow `IMPLICIT NONE ()` to be accepted.  AST->ASR already handles this case.

* Rename `implicit_none_spec_list` to `implicit_none_spec_star`
* Add `%empty` branch to `implicit_none_spec_star`
* Update `implicit1` test to cover this case.

This closes #5522

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -554,7 +554,7 @@ void yyerror(YYLTYPE *yyloc, LCompilers::LFortran::Parser &p,
 %type <ast> implicit_statement
 %type <vec_ast> implicit_statement_star
 %type <ast> implicit_none_spec
-%type <vec_ast> implicit_none_spec_list
+%type <vec_ast> implicit_none_spec_star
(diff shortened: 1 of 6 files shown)


Thirumalai-Shaktivel (approved):
LG, thank you!

certik (approved):
Thanks!

Issue #5526, opened 2024-11-30
Possible fixes for linker issues


Thirumalai-Shaktivel:
We need to support the following options how to link:
* gcc
* clang
* msvc
* LLVM lld (linked with LFortran)

We need to make it configurable, allow the user to tell LFortran how he wants the linking to be done, and where the linker.

I recommend not doing `which`, but rather just calling `gcc` or `clang` from PATH. The user is responsible for putting it there. But we can then tell lfortran where it is via environment variables and/or command line options.

For the most smooth experience, let's get the `lld` working --- it's an extra dependency and it does make our lfortran binary quite large (last time I tried I think it was something like extra 50MB), but it would remove the dependency on an external linker, which would fix all these issues. For now let's make the `lld` optional, that way we can enable it at godbolt and other places where the extra 50MB is not a big deal.

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/4764#issuecomment-2508050473_

Thirumalai-Shaktivel:
Also, see: https://github.com/lfortran/lfortran/pull/5527#issuecomment-2515693269

PR #5527, opened 2024-11-30
fix: Add a new option --linker-path

Thirumalai-Shaktivel:
Now, the user has two option to specify linker path:
- create an environment variable LFORTRAN_CC with the path
  to the linker, gcc or clang
- pass an option `--linker-path` with the path

Towards: https://github.com/lfortran/lfortran/issues/5526

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1741,11 +1741,15 @@ int link_executable(const std::vector<std::string> &infiles,
             } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
                  CC = env_CC;
             } else {
+#ifdef __clang__
+                CC = "clang";
+#else
                 std::cerr << "Error: linker not specified, "
                     "use --linker-path=<CC> or create an environment variable "
                     "`export LFORTRAN_CC=<CC>`, where CC is path to "
                     "clang or gcc ";
                 return 1;
+#endif
             }
 
             if (compiler_options.target != "" &&
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1741,8 +1741,10 @@ int link_executable(const std::vector<std::string> &infiles,
             } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
                  CC = env_CC;
             } else {
-#ifdef __clang__
+#if defined(__clang__)
                 CC = "clang";
+#elif defined(__GNUC__)
+                CC = "gcc";
 #else
                 std::cerr << "Error: linker not specified, "
                     "use --linker-path=<CC> or create an environment variable "
@@ -2715,7 +2717,7 @@ int main_app(int argc, char *argv[]) {
         err_ = err;
         if (!err) object_files.push_back(tmp_o);
     }
-    if (object_files.size() == 0) { 
+    if (object_files.size() == 0) {
         return err_;
     } else {
         return err_ + link_executable(object_files, outfile, runtime_library_dir, backend, static_link, shared_link,
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1736,6 +1736,12 @@ int link_executable(const std::vector<std::string> &infiles,
             std::string options;
             std::string runtime_lib = "lfortran_runtime";
 
+std::cout << system("which clang");
+#ifndef __clang__
+#define __clang__ 0
+#endif
+std::cout << __clang__ << ": __clang__\n";
+std::cout << __GNUC__ << ": __GNUC__\n";
             if (!linker_path.empty()) {
                 CC = linker_path;
             } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1736,28 +1736,20 @@ int link_executable(const std::vector<std::string> &infiles,
             std::string options;
             std::string runtime_lib = "lfortran_runtime";
 
-std::cout << system("which clang");
-#ifndef __clang__
-#define __clang__ 0
-#endif
-std::cout << __clang__ << ": __clang__\n";
-std::cout << __GNUC__ << ": __GNUC__\n";
             if (!linker_path.empty()) {
                 CC = linker_path;
             } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
                  CC = env_CC;
             } else {
-#if defined(__clang__)
-                CC = "clang";
-#elif defined(__GNUC__)
-                CC = "gcc";
-#else
-                std::cerr << "Error: linker not specified, "
-                    "use --linker-path=<CC> or create an environment variable "
-                    "`export LFORTRAN_CC=<CC>`, where CC is path to "
-                    "clang or gcc ";
-                return 1;
-#endif
+                if (std::ifstream c_path{"/usr/bin/clang"}) {
+                    CC = "clang";
+                } else {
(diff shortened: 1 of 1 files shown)


src/bin/lfortran.cpp
```diff
@@ -1731,21 +1731,30 @@ int link_executable(const std::vector<std::string> &infiles,
...
 
-            if (link_with_gcc) {
-                CC = "gcc";
+            char *env_CC = std::getenv("LFORTRAN_CC");
+            if (env_CC) {
+                 CC = env_CC;
+            } else if (!linker_path.empty()) {
+                CC = linker_path;
             } else {
+#ifdef __clang__
```
Thirumalai-Shaktivel, line 1745: Is it fine to use this macro?

Thirumalai-Shaktivel:
I think we need to use environment variable on Linux

src/bin/lfortran.cpp
```diff
@@ -1731,21 +1731,37 @@ int link_executable(const std::vector<std::string> &infiles,
...
                 CC = "clang";
+#elif defined(__GNUC__)
+                CC = "gcc";
+#else
+                std::cerr << "Error: linker not specified, "
+                    "use --linker-path=<CC> or create an environment variable "
+                    "`export LFORTRAN_CC=<CC>`, where CC is path to "
+                    "clang or gcc ";
+                return 1;
+#endif
```
Thirumalai-Shaktivel, line 1760: Or we can also do the following:
```suggestion
                if (std::ifstream c_path{"/usr/bin/clang"}) {
                    CC = "clang";
                } else {
                    std::cerr << "Error: linker not specified, "
                        "use --linker-path=<CC> or create an environment variable "
                        "`export LFORTRAN_CC=<CC>`, where CC is path to "
                        "clang or gcc ";
                    return 1;
                }
```

src/bin/lfortran.cpp
```diff
@@ -1731,21 +1731,37 @@ int link_executable(const std::vector<std::string> &infiles,
...
+std::cout << __GNUC__ << ": __GNUC__\n";
+            if (!linker_path.empty()) {
+                CC = linker_path;
+            } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
+                 CC = env_CC;
             } else {
+#if defined(__clang__)
                 CC = "clang";
+#elif defined(__GNUC__)
+                CC = "gcc";
```
Thirumalai-Shaktivel, line 1753:
```suggestion
```

Let's not use `gcc` as a default linker

src/bin/lfortran.cpp
```diff
@@ -1731,21 +1731,29 @@ int link_executable(const std::vector<std::string> &infiles,
...
-            if (link_with_gcc) {
-                CC = "gcc";
+            if (!linker_path.empty()) {
+                CC = linker_path;
+            } else if (char *env_CC = std::getenv("LFORTRAN_CC")) {
+                 CC = env_CC;
             } else {
-                CC = "clang";
+                if (std::ifstream c_path{"/usr/bin/clang"}) {
+                    CC = "clang";
```
Thirumalai-Shaktivel, line 1745: This is using the default clang path. If it is not available in the path, the user can specify it using --link-path=<path>

src/bin/lfortran.cpp
Thirumalai-Shaktivel: See: https://github.com/lfortran/lfortran/pull/5527#issuecomment-2508932555

Thirumalai-Shaktivel:
The CI is green.

src/bin/lfortran.cpp
certik: The problem is that if it is available, it will be used, but we are now touching filesystem for every linker invocation, slowing things down.

certik (commented):
How is MSVC linker handled?

How can you easily select gcc vs clang? Or do you need to provide full path to them?

Thirumalai-Shaktivel:
> How is MSVC linker handled?

I don't have Windows machine, right now, to test it. Maybe someone with Windows machine could help here.
According to this patch, I think it's using `clang`

> How can you easily select gcc vs clang?

The user can specify `--linker-path=gcc<or>clang` to switch between the compilers. Or one can also use the env variable, `export LFORTRAN_CC=gcc`.

--linker-path has higher priority to choose the compiler.
> do you need to provide full path to them?

I think it's not required, if it is available in the path. If it says `clang not found`, then user should make sure to give the correct path to the compiler, right?

Thirumalai-Shaktivel:
I'm confused about the requirements for this PR.

1. The user specify the compiler to choose:
    - Should we check the availability of the compiler and report the user if not available 
    - Should we get the full path of the linker compiler?
2. What needs to be used by default?
3. If we not use default, should the user or developer use `--linker-path=<>` or `LFORTRAN_CC` be set every time?

certik:
I see, I was confused by the name `--linker-path` that you need to provide a full path. Let's just call it `--linker`. There is another problem --- each linker has different command line handling, so LFortran needs to know the type. On Windows use use the MSVC `link`.

I recommend the following:
* `--linker=clang/gcc/msvc/lld` specifies the type of the linker, so that LFortran knows.
* `--linker-path` can provide optional full path to the linker executable, otherwise just the one from path is used
* We do not check if it is available (wastes time), we simply call it with the options we understand about the linker (for example for `msvc` we use the `msvc` options, and if the user supplied clang instead, it will fail). We check the exit code, if it failed, we exit. That's it.
* The default should be `msvc` on Windows, `clang` on macOS and freeBSD and either `clang` or `gcc` on Linux (not sure which compiler is the more common default system compiler, likely `gcc`, so let's use `gcc` on Linux).

By default things should just work on all platforms with no extra options. Sometimes the user might use `--linker` to select a different linker (say `clang` on Linux from the default `gcc`).

Regarding `LFORTRAN_CC`, we should also support it, but we need to somehow tell LFortran what linker it is. So how about `LFORTRAN_LINKER` and `LFORTRAN_LINKER_PATH` which would be equivalent to `--linker` and `--linker-path`?

certik:
P.S. You pushed this branch upstream instead of your fork. Do you want to move it to your fork?

Thirumalai-Shaktivel:
Yes, sorry about that.
I will move it to my fork. 

I think I need to close this PR and create a new PR. I will push it with the required changes soon.

certik:
Post a link here for the new PR.

Thirumalai-Shaktivel:
> * We cannot use gcc as a default linker, as we get the following error while compiling third party application, see: [lfortran/lfortran/actions/runs/12095481717/job/33728713390?pr=5527](https://github.com/lfortran/lfortran/actions/runs/12095481717/job/33728713390?pr=5527)
>   ```
>   /home/runner/micromamba/envs/lf/bin/../lib/gcc/x86_64-conda-linux-gnu/14.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: /home/runner/work/lfortran/lfortran/src/bin/../runtime/liblfortran_runtime.so: undefined reference to `logf@GLIBC_2.27'
>   /home/runner/micromamba/envs/lf/bin/../lib/gcc/x86_64-conda-linux-gnu/14.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: /home/runner/work/lfortran/lfortran/src/bin/../runtime/liblfortran_runtime.so: undefined reference to `exp@GLIBC_2.29'
>   ```

Using gcc to be default has some problem. First I will try to make the required changes, create a PR and apply the changes based on the CI's response.

Thirumalai-Shaktivel:
Continued here: https://github.com/lfortran/lfortran/pull/5605

PR #5530, opened 2024-11-30
Add `--realloc-lhs` to tests which need allocation of user defined variables by LFortran

czgdp1807:


src/libasr/pass/simplifier.cpp
```diff
@@ -1298,7 +1298,7 @@ class ReplaceExprWithTemporary: public ASR::BaseExprReplacer<ReplaceExprWithTemp
             ASR::expr_t* target = target_info.first; targetType target_Type = target_info.second; \
             if( ASRUtils::is_allocatable(ASRUtils::expr_type(target)) && \
                 target_Type == targetType::OriginalTarget && \
-                realloc_lhs ) { \
+                (realloc_lhs || ASR::is_a<ASR::IntrinsicArrayFunction_t>(*(*current_expr))) ) { \
```
czgdp1807, line 1301: Why we always need to allocate for target if there is `IntrinsicArrayFunction` on RHS? Following are the reasons,

1. GFortran does it always irrespective of whether you provide `-fno-realloc-lhs`. You provide this flag in `arrays_44`. GFortran will give correct result. However if you remove the call to `pack` and do `x_tmp = x(j1, :)`. Then it will segfault if you provide `-fno-realloc-lhs`. Otherwise it will work. This shows that GFortran always reallocates for `IntrinsicArrayFunction` and not for any other expression.
2. LFortran without simplifier pass also does the same. That's why `arrays_44` works without `--experimental-simplifier`.

So, IMO this change is harmless and is compatible with GFortran and LFortran without simplifier pass.

czgdp1807 (commented):
Explained the change below.

czgdp1807:
All tests pass. 

P.S. We should have an option to test for `--experimental-simplifier --fast` but `--fast` shouldn't be enabled alone. `arrays_44` doesn't work with `--fast` on CI but it might work with `--experimental-simplifier --fast` on CI. However both cases work for me locally. So need to be tested only on CI.

src/libasr/pass/simplifier.cpp
certik: It's unprincipled though --- the code in question is:
```
integer, allocatable :: x_tmp(:)

do j1 = 1, size(x, 1)
    x_tmp = pack(x(j1, :), mask(j1, :))
end do
res = sum(x_tmp)
```
Here we should give an error message in Debug mode when this is used without `--realloc-lhs`. So the test should be modified to pass "realloc-lhs" to both gfortran and lfortran.

So let's modify the test to pass realloc-lhs, then things should work. For performance reasons, we do not want to allocate automatically, let the user do it.

@czgdp1807 what do you think?

czgdp1807: Makes sense. GFortran enables `—realloc-lhs` by default. So it automatically allocates every time. Very opaque behaviour. We can add EXTRA_ARGS to the tests for LFortran.

czgdp1807: However, I am bit surprised that LFortran used to do auto-allocation of user defined variables before. Anyways, discovered and fixed now.

PR #5536, opened 2024-11-30
Support _implicit-spec-list_ in `IMPLICIT` statements

### LINKED ISSUE #3113: Implicit statement syntax error
MWE from the Netlib library dcdflib (https://netlib.org/random/):

```fortran
      DOUBLE PRECISION FUNCTION dstrem(z)
      IMPLICIT DOUBLE PRECISION (a-h,o-p,r-z),INTEGER (i-n),LOGICAL (q)
      dstrem = z
      END
```

```
$ lfortran -c --implicit-typing --fixed-form dcdflib_1.f 
 --> dcdflib_1.f:2:46
  |
2 |       IMPLICIT DOUBLE PRECISION (a-h,o-p,r-z),INTEGER (i-n),LOGICAL (q)
  |                                              ^ 
```

A temporary workaround is to separate these into individual statements:
```fortran
      DOUBLE PRECISION FUNCTION dstrem(z)
      IMPLICIT DOUBLE PRECISION (a-h,o-p,r-z)
      IMPLICIT INTEGER (i-n)
      IMPLICIT LOGICAL (q)
      dstrem = z
      END
```

This appears to be the only error blocking this library of statistical distribution functions.

### LINKED ISSUE #5352: IMPLICIT statement parsing error
Attempting to parse:
```f90
subroutine foo
  implicit integer(Y, V-X), real(M)
end subroutine foo
```
with LFortran SHA 4068df32e leads to the error:
```
 --> func.f90:2:27
  |
2 |   implicit integer(Y, V-X), real(M)
  |                           ^ 
```
I'd guess that the grammar doesn't accept an _implicit-spec-list_, as is specified in the standard.

Also, _kind-selector_ inputs that aren't an `id` or `KW_INTEGER` are improperly rejected:
```f90
implicit real(4+4)(q-s)
implicit integer(kind=2)(i)
```

pjh40:
Working with F77 validation suite FCVS_95 (see #480 ) demonstrated that our support for `IMPLICIT` statements was incomplete, particularly in that we only allow one _implicit-spec_ per _implicit-stmt_, instead of an _implicit-spec-list_.  An example statement that is now properly accepted is:
```f90
implicit real(i), integer(b)
```

 This PR adds that support, with changes in the AST grammar, the parser, AST symbol table visitor, and AST->src conversion.  AST->ASR conversion already supported this structure.

The PR also updates the test reference AST files, and adds the new test `implicit13`.  After these changes, FCVS_95 tests that use _implicit-spec-list_ no longer fail (on those statements, at least).

This closes #5352 
This closes #3113


tests/reference/asr-implicit13-4b9822a.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Implicit typing is not allowed, enable it by using --implicit-typing
```
gxyd, line 1: This file probably isn't needed, as it pertains to an error raised without use of `--implicit-typing` LFortran flag

certik:
Awesome, thanks for this fix @pjh40. I'll review it carefully tomorrow, unless others beat me to it.

Can this parse `implicit real(dp) (f)` ?

What is the ambiguous case if we want to reuse the `declaration_type_spec` rule?

I can't promise for sure, but so far we have been able to figure out how to make the grammar unambiguous in free form with the simple tokenizer.

gxyd (approved):
Otherwise, this looks quite good, the documentation helped reference things for a review.

gxyd:
> Can this parse `implicit real(dp) (f)` ?

The below seems to work:
```f90
implicit real (8) (f)
```

as AST:
```clj
[(ImplicitSpec
                (AttrType
                    TypeReal
                    [(()
                    8
                    Value)]
                    ()
                    ()
                    None
                )
                []
                [(LetterSpec
                    ()
                    f
                )]
            )]
```

but the below programs raises error:
```f90
program main
    implicit real (dp) (f)
    f = 1.2
    print *, f
end program main
```

the error is:
```console
syntax error: Token 'real' is unexpected here
  --> test.f90:24:14
   |
24 |     implicit real (dp) (f)
   |              ^^^^ 



```

and so does the below:
```f90
program main
    implicit real (kind=4) (f)
    f = 1.2
    print *, f
end program main

```

the error raised is:
```console
  --> test.f90:23:24
   |
23 |     implicit real (kind=4) (f)
   |                        ^ 


```

tests/tests.toml
```diff
@@ -3411,6 +3411,16 @@ ast_disable_style = true
...
 asr_implicit_typing = true
 
+[[test]]
+filename = "implicit12.f90"
+asr = true
+asr_implicit_typing = true
+
+[[test]]
+filename = "implicit13.f90"
+asr = true
```
Pranavchiku, line 3421:
```suggestion
```

tests/tests.toml
```diff
@@ -3411,6 +3411,16 @@ ast_disable_style = true
 filename = "implicit11.f90"
 asr_implicit_typing = true
 
+[[test]]
+filename = "implicit12.f90"
+asr = true
```
Pranavchiku, line 3416:
```suggestion
```

Pranavchiku: I see, `asr=true` might be used to test if this file provides correct error or not.

Pranavchiku: Resolving as no change required.

grammar/AST.asdl
```diff
@@ -99,7 +99,10 @@ import_modifier
 
 implicit_statement
     = ImplicitNone(implicit_none_spec* specs, trivia? trivia)
-    | Implicit(decl_attribute type, letter_spec* kind, letter_spec* specs, trivia? trivia)
+    | Implicit(implicit_spec* specs, trivia? trivia)
+
+implicit_spec
+    = ImplicitSpec(decl_attribute type, letter_spec* kind, letter_spec* specs)
```
Pranavchiku, line 105: I think it is fine. I am unable to think any other design better than this.

certik: The only other design is `implicit_spec = (decl_attribute type, letter_spec* kind, letter_spec* specs)`, or merging `implicit_spec` and `implicit_none_spec`. But I think the above is fine for now.

tests/reference/asr-implicit13-4b9822a.stderr
certik: If we don't have a test for this error message yet, then we should keep it.

tests/tests.toml
certik: Actually this `implicit12` test should either be removed, or the other one moved here, since it is already registered on line 3441.

tests/tests.toml
certik: Yes, I think we only need to test this error message once and we do so for implicit12, so we can remove it from here.

pjh40: I'm not sure why I can't respond to the comments above about the _kind-selector_ behavior, but you're correct, it is broken.  There isn't a difference in the _declaration-type-spec_ accepted by this PR and that accepted by the `main` branch:  I only made the grammar accept a list of those (broken) specs.  I mentioned in #5352 that I was going to file a separate issue about this, because I'm not sure how to resolve the "essential" shift/reduce conflicts that arise.  I would suggest that we can discuss the s/r problems in that issue.

pjh40: Thanks for your careful reviews, I need that!  I believe that the new commit SHA 196948cbed216427c6d612ec3005a83a42867e15 fixes the testing issue.

certik (approved):
Thanks, I think it looks great to merge now.

PR #5547, opened 2024-12-02
fix: use `fscanf()` for reading character arrays

### LINKED ISSUE #5545: read for character array
- `lfortran` version
```
LFortran version: 0.42.0
Platform: Linux
LLVM: 19.1.1
```
lfortran will exceed the character len
- code
```fortran
program main
   implicit none
    character(len=1)::s(2)
    read(*,*)s
    print*,s(1)
    print*,s(2)
end program main
```
``` sh
$ gfortran 1.f90 && ./a.out
ab cd
a
c
$lfortran 1.f90
ab cd
ab
cd
```
 if use `character(len=1)::s` ,it is right

kmr-srbh:
fixes #5545 

We can read character arrays with character length more than 1 if we initialize the elements, something similar to `string_init()`. Currently it segfaults as `*p` points to an uninitialized memory location.


file_21_data.txt
```diff
@@ -0,0 +1 @@
+ab cd
```
certik, line 1: This file should be removed.

certik:
Otherwise I think it's fine.

certik (approved):

gxyd (approved):
Looks good to me.

Issue #5549, opened 2024-12-03
IMPLICIT declaration_type_spec


pjh40:
Currently, the `parser.yy` grammar uses rules local to `implicit_spec` for imitating the _kind-selector_ component of an (intrinsic) _declaration-type-spec_ in an IMPLICIT statement.  While these rules handle some simple cases correctly (e.g. `implicit real(8) (a)`), it is easy to generate examples that should be accepted, but are not:
```f90
implicit real(kind=8) (a)
implicit real(4+4) (b)
implicit character(len=20, kind=c_char) (c)
```

Furthering the problem are inputs that are accepted using the `"(" letter_spec_list ")" "(" letter_spec_list ")"` style of rules, which are never checked or resolved as _scalar-int-constant-expr_:
```f90
program bad_kind
  use constants
  implicit real(r8k) (a)
  implicit real(dp) (b)
  implicit real(q-l) (c)
  print *, kind(a), kind(b), kind(c)
end program bad_kind
```
prints `4    4    4`, without even attempting to lookup values for `r8k` or `dp` or compute the value `q-l`.  Note that these problems are exhibited both before (SHA c75347178f1982e6) and after (SHA aed88c2ec20c7e4591) my commit to add _implicit-spec-list_ 

The appropriate solution to these problems is to change the grammar to simply use the `declaration_type_spec` rule, which handles all of these cases.  That would look like:
```bison
implicit_statement
    : KW_IMPLICIT KW_NONE sep { $$ = IMPLICIT_NONE(TRIVIA_AFTER($3, @$), @$); }
    | KW_IMPLICIT KW_NONE "(" implicit_none_spec_star ")" sep {
            $$ = IMPLICIT_NONE2($4, TRIVIA_AFTER($6, @$), @$); }
    | KW_IMPLICIT implicit_spec_list sep {
            $$ = IMPLICIT($2, TRIVIA_AFTER($3, @$), @$); }
    ;

implicit_spec_list
    : implicit_spec_list "," implicit_spec { $$ = $1; LIST_ADD($$, $3); }
    | implicit_spec { LIST_NEW($$); LIST_ADD($$, $1); }
    ;

implicit_spec
    : declaration_type_spec "(" letter_spec_list ")" {
           $$ = IMPLICIT_SPEC($1, $3, @$); }
    ;
```
The complete implementation of this change can be seen in the draft PR linked below.

[...]

PR #5550, opened 2024-12-03
Demonstration of using IMPLICIT declaration-type-spec

pjh40:
*DO NOT MERGE*

This is an example of how to use the `declaration_type_spec` rule in the `implicit_spec` to improve parsing of the _kind-selector_.  As discussed in #5549, this code introduces 5 essential shift/reduce conflicts, which need to be addressed.  This PR is intended just for allowing developers to see the complete code.


certik:
I can verify that this introduces 5 essential conflicts.

PR #5552, opened 2024-12-03
ci: Use micromamba 2.0 via setup-micromamba 2

jjerphan:
Alternative to https://github.com/lfortran/lfortran/pull/5013 with unaltered changes and thus caches for the "Check Out-of-Source Debug build" workflow.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,9 +38,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         with:
-          micromamba-version: '1.5.10-0'
+          micromamba-version: '2.0.4-0'
           environment-file: ci/environment.yml
           create-args: >-
             python=${{ matrix.python-version }}
@@ -123,9 +123,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         with:
-          micromamba-version: '1.5.10-0'
+          micromamba-version: '2.0.4-0'
           environment-file: ci/environment_linux.yml
           create-args: >-
             python=3.10
@@ -245,9 +245,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         with:
-          micromamba-version: '1.5.10-0'
+          micromamba-version: '2.0.4-0'
           environment-file: ci/environment_linux.yml
           create-args: >-
             python=3.10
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -894,10 +894,10 @@ jobs:
             make=4.3
             nodejs=18.20.4
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
         if: contains(matrix.llvm-version, '10')
         with:
-          micromamba-version: '1.5.10-0'
+          microconda-version: '2.0.4-0'
```
certik, line 900: Is this a typo? What is `microconda`?

certik:
This PR looks good!

How does the 'cache for the "Check Out-of-Source Debug build" workflow.' work? Why does it work here, but not in the previous PR?

.github/workflows/CI.yml
jjerphan: Yes, just a typo.

jjerphan: Fixed by 2353620.

.github/workflows/CI.yml
```diff
@@ -38,9 +38,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@v2
```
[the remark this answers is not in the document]
jjerphan: Fixed by 2353620.

jjerphan:
> How does the 'cache for the "Check Out-of-Source Debug build" workflow.' work? 

A cache is generally used for each workflow. Such a cache is identified by a key which is built using information of the context (e.g. version of the GHA, of the parameters, etc.). The construction of the key is workflow-specific, you should be able to browse [the ones of this PR (the one identified by `branch:refs/pull/5552/merge`)](https://github.com/lfortran/lfortran/actions/caches?query=branch%3Arefs%2Fpull%2F5552%2Fmerge+). I do not know how `lfortran` sets them exactly.

Note that I assume this is due to caches (which this PR does not modify anyway) based on what I have observed, that is that some paths referenced by this workflow have changed (this is for instance the case of this environment specification):

https://github.com/lfortran/lfortran/blob/7eb6c7cecec92ca83c6ef091bb9da0947d54d82f/.github/workflows/CI.yml#L332

What I know is that by not modifying the workflow, I do not invalidate any potential out-of-date caches which might still relying on a previous state of lfortran's workdirectory structure. This is not the case of #5013 which modify the context of this workflow.

I think the root cause of this issue is unrelated to micromamba 2.0 and better be solved in a dedicated PR.

certik (approved):
This PR upgrades micromamba from 1.5.10 to 2.0.4, no other changes and all the bugs that we have seen before were fixed upstream, so everything works.

There is one bug remaining: there is some cache problem as described in https://github.com/lfortran/lfortran/pull/5552#issuecomment-2514817244, but that is unrelated to this PR, it was just discovered while working on upgrading micromamba. We should investigate it and fix it. For now I made an issue for it at https://github.com/lfortran/lfortran/issues/5561.

@jjerphan thank you so much for helping us upgrade micromamba, we really appreciate it!

jjerphan:
You are welcome.

Thank you for your patience.

Issue #5558, opened 2024-12-03
Pass location information of function / subroutine / program / module name to ASR


Pranavchiku:
With `--rename-symbol` we intend to return all occurrences of a symbol and when it comes to `Unit` i.e. `function / subroutine / program / modules / etc ` the location information contains entire start line and end line but we only need the `id`. This was discussed as a problem before and raised at https://github.com/lfortran/lfortran/issues/5524. A potential solution is to pass location information of `id` to ASR, for example:

```fortran
function xyz
...
end function xyz
```

We can store location of `xyz` in AST and then pass it to ASR.

certik:
@Pranavchiku what needs to be done to finish this?

Pranavchiku:
I think we can close this with reference to #5602

PR #5560, opened 2024-12-03
feat: propagate location information of name in `unit`

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/5558

We have location information but it is incorrect.

```fortran
real function xyz()
        print *, "abc"
end function xyz

```

```console
% lfortran a.f90 --show-asr
Function: xyz
loc->first: 53763168
loc->last: 24576
[...]
```

### DIFF
--- a/grammar/AST.asdl
+++ b/grammar/AST.asdl
@@ -53,7 +53,7 @@ program_unit
     | Function(identifier name, arg* args, decl_attribute* attributes,
         expr? return_var, bind? bind, trivia? trivia, unit_decl1* use, import_statement* import,
         implicit_statement* implicit, unit_decl2* decl, stmt* body,
-        program_unit* contains, identifier* temp_args)
+        program_unit* contains, identifier* temp_args, Location start_name, Location end_name)
 
 unit_decl1
     = Use(decl_attribute* nature, identifier module, use_symbol* symbols,
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -988,25 +988,28 @@ function
         sep use_statement_star import_statement_star implicit_statement_star decl_statements
         contains_block_opt
         end_function sep {
+            YYLTYPE start_name_loc = @2;
             LLOC(@$, @12); $$ = FUNCTION0($2, $4, nullptr, nullptr,
                 TRIVIA($6, $13, @$), $7, $8, $9, SPLIT_DECL(p.m_a, $10),
-                SPLIT_STMT(p.m_a, $10), $11, $12, @$); }
+                SPLIT_STMT(p.m_a, $10), $11, $12, @$, &start_name_loc, nullptr); }
     | KW_FUNCTION id "(" id_list_opt ")"
         bind
         result_opt
         sep use_statement_star import_statement_star implicit_statement_star decl_statements
         contains_block_opt
         end_function sep {
+            YYLTYPE start_name_loc = @2;
             LLOC(@$, @14); $$ = FUNCTION0($2, $4, $7, $6, TRIVIA($8, $15, @$),
-                $9, $10, $11, SPLIT_DECL(p.m_a, $12), SPLIT_STMT(p.m_a, $12), $13, $14, @$); }
+                $9, $10, $11, SPLIT_DECL(p.m_a, $12), SPLIT_STMT(p.m_a, $12), $13, $14, @$, &start_name_loc, nullptr); }
     | KW_FUNCTION id "(" id_list_opt ")"
         result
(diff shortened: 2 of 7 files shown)


src/lfortran/parser/parser.yy
```diff
@@ -988,25 +988,28 @@ function
         sep use_statement_star import_statement_star implicit_statement_star decl_statements
         contains_block_opt
         end_function sep {
+            YYLTYPE start_name_loc = @2;
             LLOC(@$, @12); $$ = FUNCTION0($2, $4, nullptr, nullptr,
                 TRIVIA($6, $13, @$), $7, $8, $9, SPLIT_DECL(p.m_a, $10),
-                SPLIT_STMT(p.m_a, $10), $11, $12, @$); }
+                SPLIT_STMT(p.m_a, $10), $11, $12, @$, &start_name_loc, nullptr); }
```
Pranavchiku, line 994: @certik is this the correct way to pass location information of `id` i.e. `xyz`?

certik: I am trying to minimize "coding" in `parser.yy`, that's why we use the macros to hide all "coding". Why cannot you pass `@2` directly?

src/libasr/asdl.py
```diff
@@ -223,7 +223,9 @@ def tokenize_asdl(buf):
     for lineno, line in enumerate(buf.splitlines(), 1):
         for m in re.finditer(r'\s*(\w+|--.*|.)', line.strip()):
             c = m.group(1)
-            if c[0].isalpha():
+            if c == "Location":
+                yield Token(TokenKind.TypeId, c, lineno)
+            elif c[0].isalpha():
```
certik, line 228: I think you modified a wrong place. Just add `location` into the line 36 above:
```
builtin_types = {'identifier', 'string', 'int', 'bool', 'float', 'node', 'symbol_table', 'void'}
```
That should do it.

grammar/AST.asdl
```diff
@@ -53,7 +53,7 @@ program_unit
     | Function(identifier name, arg* args, decl_attribute* attributes,
         expr? return_var, bind? bind, trivia? trivia, unit_decl1* use, import_statement* import,
         implicit_statement* implicit, unit_decl2* decl, stmt* body,
-        program_unit* contains, identifier* temp_args)
+        program_unit* contains, identifier* temp_args, Location start_name, Location end_name)
```
certik, line 56: Per my comment below:
```suggestion
        program_unit* contains, identifier* temp_args, location start_name, location end_name)
```

certik: Also change:
```
-- We extend these by:
--   * node (any ASDL node)
--   * bool
```
by adding `* location (for storing location information only)` in there, or something similar.

certik:
I think this is great, exactly as I was thinking. I left some comments to clean it up.

grammar/AST.asdl
certik: Idea: instead of `location`, why not just use `identifier`? I think that will work perfectly in this case and we already have support for it.

src/lfortran/parser/parser.yy
certik: If my idea above (https://github.com/lfortran/lfortran/pull/5560/files#r1868016909) works, then just pass in `$2`.

grammar/AST.asdl
certik: This might be all that is needed:
```suggestion
        program_unit* contains, identifier* temp_args, identifier? end_name)
```

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1299,6 +1299,18 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            Location* loc = x.m_start_name;
+            std::cout<<"loc->first: "<<(*loc).first<<std::endl;
+            std::cout<<"loc->last: "<<(*loc).last<<std::endl;
+
+            Location* end_loc = x.m_end_name;
+            std::cout<<"end_loc->first: "<<(*end_loc).first<<std::endl;
+            std::cout<<"end_loc->last: "<<(*end_loc).last<<std::endl;
+        } else {
+            std::cout<<"nullptr hai"<<std::endl;
+        }
```
Pranavchiku, line 1313: Now we have correct location information. If the design is fine we can proceed with passing it to ASR.

Pranavchiku:
Applied code review https://github.com/lfortran/lfortran/pull/5572 and hence closing with reference to it.

PR #5565, opened 2024-12-03
Allow for allocatable file names in file inquire

### LINKED ISSUE #5559: (PRIMA) Handle allocatable file name in `inquire()`
MRE :
```fortran
program test 
    character(len=:), allocatable :: fname
    logical :: fexist
    fname = "a.f90"
    inquire(file = fname, exist = fexist)
    print *, fexist
end program
```

LFortran Error : 
```console
  %fname = alloca %string_descriptor, align 8
 i8*  call void @_lfortran_inquire(%string_descriptor* %fname, i1* %fexist, i32 -1, i1* %13)
```

advikkabra:
Fixes #5559

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8004,6 +8004,12 @@ public:
         if (x.m_file) {
             this->visit_expr_wrapper(x.m_file, true);
             f_name = tmp;
+            if (ASRUtils::is_allocatable(ASRUtils::expr_type(x.m_file))) {
+                llvm::Type *string_desc = llvm_utils->get_type_from_ttype_t_util(ASRUtils::expr_type(x.m_file), module.get());
+                llvm::Type *i8_pointer = llvm::Type::getInt8Ty(context)->getPointerTo();
+                f_name = llvm_utils->CreateLoad2(i8_pointer, llvm_utils->create_gep2(string_desc, f_name, 0));
+            }
+
         } else {
             f_name = llvm::Constant::getNullValue(character_type);
         }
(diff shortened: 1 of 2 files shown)


certik:
I think the fix is fine. Can you please add a test that works both with GFortran and LFortran in this PR, but fails in main?

gxyd (approved):
Looks good to me. Thanks @advikkabra

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8004,6 +8004,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         if (x.m_file) {
             this->visit_expr_wrapper(x.m_file, true);
             f_name = tmp;
+            if (ASRUtils::is_allocatable(ASRUtils::expr_type(x.m_file))) {
+                llvm::Type *string_desc = llvm_utils->get_type_from_ttype_t_util(ASRUtils::expr_type(x.m_file), module.get());
+                llvm::Type *i8_pointer = llvm::Type::getInt8Ty(context)->getPointerTo();
+                f_name = llvm_utils->CreateLoad2(i8_pointer, llvm_utils->create_gep2(string_desc, f_name, 0));
+            }
```
certik, line 8011: This looks like a cast --- it should be fixed like fix in https://github.com/lfortran/lfortran/pull/5555 I think.

@kmr-srbh what do you think?

kmr-srbh: Yes, we should handle this in the ASR using `StringPhysicalCast`.

@advikkabra please see `visit_Inquire()` in `ast_body_visitor.cpp`. We need to create a `StringPhysicalCast` node casting the `file` argument's `physical_type` to `PointerString` if it is already not so.

advikkabra: Right. I'll make another PR to do that.

Issue #4770, opened 2024-09-09
CharacterString with '&!' as substring in it raises syntax error


gxyd:
## Minimal Reproducible Example

```f90
program main
    character(len=20) :: textblock
    textblock = '&!'
end program main
```

On compiling the above program with LFortran, it raises an error:
```zsh
~/Doc/lfortran main !1 ?4 ❯ main_lfortran test.f90 --show-ast          lf 10:36:53 PM
tokenizer error: Token ''' is not recognized
 --> test1.f90:3:17
  |
3 |     textblock = '&!'
  |                 ^ token not recognized


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

gxyd:
If I use the compiler option `--no-prescan`, the program works without any problem:
```console
> lfortran test.f90 --show-ast --no-prescan
(TranslationUnit
    [(Program
        main
        ()
        []
        []
        [(Declaration
            (AttrType
                TypeCharacter
                [(len
                20
                Value)]
                ()
                ()
                None
            )
            []
            [(textblock
            []
            []
            ()
            ()
            None
            ())]
            ()
        )]
        [(Assignment
            0
            textblock
            (String "&!")
            ()
        )]
        []
    )]
)
```

and the output of `--show-prescan` is:
```console
> lfortran test.f90 --show-prescan
program main
    character(len=20) :: textblock
    textblock = 'end program main

```

probably that's where the problem is. I don't fully understand the meaning of *pre-scan* though, I'll check about that.

PR #5569, opened 2024-12-04
fix: in a string with '&' and '!' together isn't a comment

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/4770

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1239,6 +1239,7 @@ RUN(NAME derived_types_37 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME line_continuation_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME line_continuation_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME line_continuation_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
+RUN(NAME line_continuation_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 
 RUN(NAME program_01 LABELS gfortran)
 RUN(NAME init_values LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/line_continuation_04.f90
+++ b/integration_tests/line_continuation_04.f90
@@ -1,0 +1,32 @@
+program line_continuation_04
+    implicit none
+    integer :: result, a, b, c
+    character(len=20) :: textblock
+    textblock = '&!'
+    print *, textblock
+    if (textblock /= "&!") error stop
+
+    textblock = "apple&
+!apple
+"
+    print *, textblock
+    if (textblock /= "apple") error stop
+
+    ! textblock = "apple'&!apple"
+    ! if (textblock /= "apple'&!apple") error stop
+
+    a = 1
+    b = 2
+    c = 3
+    ! the below line doesn't have &! in a string
+    ! hence, everything after ! is treated as a
+    ! comment
+    result = a + b +&!addition here
+    c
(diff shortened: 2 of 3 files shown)


gxyd:
Actually, I figured another test case, which even this PR doesn't fix:
```f90
textblock = "apple'&!apple"
```

I'm looking at some other issue, but I'll have to come back to why that fails (probably because of `in_string`), happy to receive any help on this PR.

kmr-srbh:
You are right, this happens due to `in_string` only, when we set it here:

```c++
if (s[pos] == '\'' || s[pos] == '"') {
  in_string = !in_string;
}
```

I think we need to remember the correct opening quotes. I am able to pass this using the below hack. It is not clean.

```diff
diff --git a/src/lfortran/parser/parser.cpp b/src/lfortran/parser/parser.cpp
index 6b57bfd6d..efa201a4d 100644
--- a/src/lfortran/parser/parser.cpp
+++ b/src/lfortran/parser/parser.cpp
@@ -246,6 +246,7 @@ void cont1(
 ) {
     ws_or_comment = true;
     bool in_comment = false;
+
     while (s[pos] != '\n') {
         // in a string if '&!' appear together then it isn't a comment
         if (s[pos] == '!' && (!in_string || s[pos - 1] != '&')) {
@@ -627,9 +628,16 @@ std::string prescan(const std::string &s, LocationManager &lm,
         std::string out;
         size_t pos = 0;
         bool in_comment = false, newline = true, in_string = false;
+        char quote = '\0';
         while (pos < s.size()) {
-            if (s[pos] == '\'' || s[pos] == '"') {
-                in_string = !in_string;
+            if ((s[pos] == '\'' || s[pos] == '"')) {
+                if (quote == '\0') {                // string starts, store it's opening quote
+                    in_string = true;
+                    quote = s[pos];
+                } else if (quote == s[pos]) {       // string ends, reset quote
+                    in_string = false;
+                    quote = '\0';
+                }
             }
             if (newline && is_include(s, pos)) {
                 int col = 0; // doesn't matter
```

gxyd:
I also tried a solution to the above (I haven't pushed a solution yet) and I also tried yours as well, and then the below fails:
```f90
    textblock = "apple'&!apple"
    if (textblock /= "apple'&!apple") error stop   ! this is where it fails
```

gxyd:
Minimal example:
```f90
program line_continuation_04
    implicit none
    character(len=20) :: textblock
    textblock = "apple&
!apple
"
    print *, textblock

    textblock = "apple'&!apple"
    if (textblock /= "apple'&!apple") error stop
end program line_continuation_04

```

certik:
Those are some real corner cases, didn't even know about this. Just fix them one by one, add a test, etc.

certik:
CI has to pass.

gxyd:
> Those are some real corner cases, didn't even know about this. Just fix them one by one, add a test, etc.

Makes sense, I'll open a separate issue for once I get the PR in good shape and if that allows us to move forward with compilation of `fman` third party repository.

gxyd:
I've mostly focused on free form prescanning in this PR, as these are quite a few corner cases, probably I'll open separate issue for free form as well, though I'll not work on it currently as it isn't needed for any third party compilation yet.

gxyd:
So, we've fixed all the tests in this PR now, including the MRE here: https://github.com/lfortran/lfortran/pull/5569#issuecomment-2517138029

gxyd:
Seems like the last fix was causing some issues in our third party tests, so I've reverted it. I'll handle it in a separate issue, so that this is ready for a review now.

certik (approved):
I think this is good. The logic is getting quite complicated there, but I think it is correct and the overhead shouldn't be large here.

gxyd:
> The logic is getting quite complicated there

Agreed, there a few corner cases, I should try to simplify the logic as well, once I've identified all the failing cases and some working solution to them (without overhead), after that I'll focus on simplifying the solution to it as well.

Issue #5524, opened 2024-11-29
`--rename-symbol` sometimes returns incorrect ranges


dylon:
The new `--rename-symbol` option sometimes return the wrong ranges for symbols. For example, when I want to rename `eval_1d` in `tests/function_call1.f90`, lfortran gives me the following ranges:

```shell
./src/bin/lfortran --rename-symbol --line=18 --column=22 tests/function_call1.f90 | jq
```

```json
[
  {
    "kind": 1,
    "location": {
      "range": {
        "start": {
          "character": 5,
          "line": 8
        },
        "end": {
          "character": 25,
          "line": 12
        }
      },
      "uri": "uri"
    },
    "name": "eval_1d"
  },
  {
    "kind": 1,
    "location": {
      "range": {
        "start": {
          "character": 7,
          "line": 4
        },
        "end": {
          "character": 28,
          "line": 4
        }
      },
      "uri": "uri"
    },
    "name": "eval_1d"
  }
]
```

If we look at the text of [the file](https://github.com/lfortran/lfortran/blob/main/tests/function_call1.f90):

```fortran
module module_function_call1
    type :: softmax
    contains
      procedure :: eval_1d
    end type softmax
  contains
  
    pure function eval_1d(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
    end function eval_1d
  
    pure function eval_1d_prime(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
      res = self%eval_1d(x)
    end function eval_1d_prime
end module module_function_call1
```

, the correct ranges should be the following:
1. start: line=3, character=19; end: line=3, character=26
2. start: line=7, character=18; end: line=7, character=25
3. start: line=11, character=17; end: line=11, character=24
4. start: line=17, character=17; end: line=17, character=24

certik:
Now it returns 3 things:
```
$ lfortran --rename-symbol --line=18 --column=22 tests/function_call1.f90 | jq
[
  {
    "kind": 12,
    "location": {
      "range": {
        "start": {
          "character": 19,
          "line": 8
        },
        "end": {
          "character": 26,
          "line": 8
        }
      },
      "uri": "uri"
    },
    "name": "eval_1d"
  },
  {
    "kind": 12,
    "location": {
      "range": {
        "start": {
          "character": 18,
          "line": 12
        },
        "end": {
[...]

PR #5572, opened 2024-12-04
feat: propagate location information of name in `Function`

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/5558

This now logs correct information of `xyz`, although there few places are missing that is a separate bug not related to this PR. Almost fixed https://github.com/lfortran/lfortran/issues/5524, still not fully

We have location information which seems correct to me

```fortran
module module_function_call1
    type :: softmax
    contains
      procedure :: eval_1d
    end type softmax
  contains
  
    pure function eval_1d(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
    end function eval_1d
  
    pure function eval_1d_prime(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
      res = self%eval_1d(x)
    end function eval_1d_prime
end module module_function_call1
```

```console
% lfortran --rename-symbol --line=18 --column=22 tests/function_call1.f90                 
[{"kind":1,"location":{"range":{"start":{"character":19,"line":8},"end":{"character":26,"line":8}},"uri":"uri"},"name":"eval_1d"},{"kind":1,"location":{"range":{"start":{"character":18,"line":12},"end":{"character":25,"line":12}},"uri":"uri"},"name":"eval_1d"},{"kind":1,"location":{"range":{"start":{"character":7,"line":4},"end":{"character":28,"line":4}},"uri":"uri"},"name":"eval_1d"}]
```

We just miss `eval_1d` from `line=18`, it is not related to this PR, separate bug. I know how to fix it.


certik (approved):
I reviewed it, I think it looks good. It gets the job done.

If we later decide that we don't need a dedicated type just for location information, we can upgrade this. I think it's no harm to have this capability, and it seems faster to do it this way than also carrying around the string itself that we don't need for compilation.

Pranavchiku:
Thanks for the review! I am merging this and will add this functionality for `Program / Module / Subroutine` in subsequent PRs

PR #5573, opened 2024-12-04
fix: retain symbol name for `ExternalSymbol` during import if `m_external` is `ASR::Function_t`

### LINKED ISSUE #5366: ExternalSymbol is stored instead of GenericProcedure where multiple modules exists
```fortran 
module inf_mod_1
interface is_finite
    module procedure is_finite_1
end interface is_finite
contains
subroutine is_finite_1() 
end subroutine is_finite_1
end module inf_mod_1

module infnan_mod_1
use, non_intrinsic :: inf_mod_1, only : is_finite
contains
subroutine is_finite_1(x)
integer , intent(in) :: x
print *, x 
end subroutine is_finite_1
end module infnan_mod_1


program bobyqa_exmp
use, non_intrinsic :: infnan_mod_1
call is_finite_1(1)
end program bobyqa_exmp
```

LFortran Error :
```console
semantic error: More actual than formal arguments in procedure call
  --> integration_tests/a.f90:22:18
   |
22 | call is_finite_1(1)
   |                  ^ 
```

Here issue is in SymbolTable of program `bobyqa_exmp` where `is_finte_1` is stored as ExternalSymbol for is_finite_1 of inf_mod_1 and we have no information of is_finite_1 of infnan_mod_1.

kmr-srbh:
fixes #5366

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1042,6 +1042,7 @@ RUN(NAME modules_56 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NOFAST)
 RUN(NAME modules_57 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME modules_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES modules_58_module.f90)
 RUN(NAME modules_59 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME modules_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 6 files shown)


kmr-srbh:
The issue here was that subroutine `is_finite_1` from `infnan_mod_1` was getting overwritten by `is_finite_1` from `inf_mod_1` during import. This happened because we always import `ExternalSymbol` to the symbol table based on `m_original_name`, which in this case, was same.

With the changes in this PR, we use the symbol name for importing the symbol if it is a function as it can be part of a `GenericProcedure` and have the same name. We thus avoid the function from getting overwritten.

certik (approved):
Thanks! The ASR output looks good, it's a little convoluted example, so I hope I didn't make a mistake, but it looks like it uses ExternalSymbols as it should and imports the right symbols at the right place.

PR #5577, opened 2024-12-04
Create PhysicalCast node for allocatable file names in inquire

advikkabra:
Revert changes in #5565 , and insert an ASR node instead.


certik (approved):
I think this looks good, much better.

This I think ultimately should have been caught by ASR verify(), but I guess we are not as strict yet.

Issue #5463, opened 2024-11-23
Continue Compilation flag fails for some errors


parth121101:
I will keep on modifying this list that contains the error tests that does not skip error statements despite of continue-compilation flag.

- [x] tests/errors/int_01.f90
- [x] tests/errors/integer16.f90
- [x] tests/errors/common1.f90
- [x] tests/errors/kind_float_of_int.f90
- [x] tests/errors/kind_invalid_int_of_complex.f90
- [x] tests/errors/kind_invalid_int_of_int.f90
- [x] tests/errors/kind_invalid_int_of_logical.f90
- [x] tests/errors/kind_star_of_complex.f90
- [x] tests/errors/kind_star_of_int.f90
- [x] tests/errors/kind_star_of_logical.f90
- [x] tests/errors/kind_string_of_int.f90
- [x] tests/errors/kind_var_of_int.f90
- [x] tests/errors/kind1.f90
- [x] tests/errors/kind2.f90
- [x] tests/errors/type_casting_01.f90
- [x] tests/errors/type_conflict1.f90
- [x] tests/errors/init1.f90
- [x] tests/errors/program_without_line_01.f90
- [x] tests/errors/program_without_line_02.f90
- [x] tests/errors/program_without_line_03.f90
- [x] tests/errors/program_without_line_04.f90

Pranavchiku:
Nice, if you can send a PR at the same time to fix that would be awesome!

parth121101:
All the tasks of this issue are solved. Closing the issue

PR #5578, opened 2024-12-04
fix: resolved failing of continue compilation flag for error files for different kinds

parth121101:
Towards: #5463 #5036

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3064,9 +3064,15 @@ public:
                         s_intent == ASRUtils::intent_inout));
                 }
                 ASR::symbol_t *type_declaration;
-                ASR::ttype_t *type = determine_type(x.base.base.loc, sym, x.m_vartype, is_pointer,
+                ASR::ttype_t *type = nullptr;
+                try {
+                type = determine_type(x.base.base.loc, sym, x.m_vartype, is_pointer,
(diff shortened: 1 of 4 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3064,9 +3064,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         s_intent == ASRUtils::intent_inout));
                 }
                 ASR::symbol_t *type_declaration;
-                ASR::ttype_t *type = determine_type(x.base.base.loc, sym, x.m_vartype, is_pointer,
+                ASR::ttype_t *type = nullptr;
+                try {
+                type = determine_type(x.base.base.loc, sym, x.m_vartype, is_pointer,
```
certik, line 3069: I think the indentation here is incorrect?

certik (commented):
I think that this is fine. I am not super happy for adding all these try/catch everywhere --- is there a way to catch it at an upper level?

parth121101:
> I think that this is fine. I am not super happy for adding all these try/catch everywhere --- is there a way to catch it at an upper level?

Will figure this out @certik

Pranavchiku:
> is there a way to catch it at an upper level?

Yes, let's catch it inside `ast_symboltable_visitor` where we run a loop and  `visit_unit_decl**`  for each `unit`

certik (approved):
Thanks!

Issue #5563, opened 2024-12-03
`subroutine_from_functions_simplifier` dumps some statements  


assem2002:
MRE 
```fortran 
module gg
    contains
    function func_arr3(inp_ref) result(ret_arr)
        integer, intent(in) :: inp_ref
        integer,allocatable :: ret_arr(:)
        ! print *, allocated(ret_arr)
        allocate(ret_arr(1))
        ret_arr = [inp_ref]
        print *, "worked", inp_ref
    end function
end module


program p
    use gg
    implicit none
    integer, allocatable :: inp(:)
    integer :: i =77
    
    do i= 1, 3
        inp = func_arr3(i)
    end do 
end program  
end program  
```
```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt6.f90 && ./a.out 
 worked           1
 worked           2
 worked           3
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt6.f90 --experimental-simplifier
worked    77
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt6.f90 
worked    1
worked    2
worked    3





```

harperjf:
I tried a number of changes to that program.  The results from gfortran were OK, but from lfortran 0.42.0 were not. The program:
```
module gg ! file alloc.f90
    contains
    function func_arr3(inp_ref) result(ret_arr)
        integer, intent(in) :: inp_ref
        integer,allocatable :: ret_arr(:)
        print *, 'allocated on entry?', allocated(ret_arr)
!!        allocate(ret_arr(1))
        ret_arr = [inp_ref,inp_ref+1]
        print *, "worked", ret_arr !! inp_ref
        print *,'allocated on exit ?',allocated(ret_arr)
    end function
end module

program p
    use gg, only: func_arr3
    implicit none
    integer, allocatable :: inp(:)
    integer :: i !! =77
    
    do i= 1, 3
       inp = func_arr3(i)
       print *,'inp = ',inp
    end do 
end program  
```
Results:
```
(lf) john:~$ gfortran alloc.f90 && ./a.out
 allocated on entry? F
 worked           1           2
 allocated on exit ? T
 inp =            1           2
 allocated on entry? F
 worked           2           3
 allocated on exit ? T
 inp =            2           3
 allocated on entry? F
 worked           3           4
 allocated on exit ? T
 inp =            3           4
(lf) john:~$ lfortran alloc.f90 
[...]

PR #5580, opened 2024-12-04
Fix : Flush pass_result into the body if we'll overwrite it in `subroutine_from_function_simplifier`.

assem2002:
Fixes https://github.com/lfortran/lfortran/issues/5563

The fix does the same as other passes do, it makes sure to add the statements pushed in `Vec<ASR::stmt_t*> pass_result` in the appropriate body once we make a nested call that will overwrite the `pass_result` as it's shared by all visitors in the class.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -350,7 +350,7 @@ RUN(NAME doloop_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c cpp wasm)
 RUN(NAME doloop_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # nested subroutine
 RUN(NAME doloop_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm) # implicit cast loop variable
 RUN(NAME doloop_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm EXTRA_ARGS --use-loop-variable-after-loop)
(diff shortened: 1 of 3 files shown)


certik (approved):
I think that's fine, thanks!

integration_tests/CMakeLists.txt
```diff
@@ -350,6 +350,7 @@ RUN(NAME doloop_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c cpp wasm)
 RUN(NAME doloop_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # nested subroutine
 RUN(NAME doloop_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm) # implicit cast loop variable
 RUN(NAME doloop_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm EXTRA_ARGS --use-loop-variable-after-loop)
+RUN(NAME doloop_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
```
gxyd, line 353: Seems like a failure in WASM backend for `do_loop_12`:
```console
  throw err;
  ^

Error: Cannot find module '/home/runner/work/lfortran/lfortran/integration_tests/test-wasm/doloop_12.js'
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.20.4
```

assem2002: I don't get what is the issue with WASM. so I'll just remove wasm test for now

PR #5542, opened 2024-12-01
feat: Multiple Headed DoConcurrentLoop Total Iterations

adit4443ya:
Towards #4900

Intended towards correctly calculating total number of iterations in Multiple Headed DoConcurrentLoop.

Issue #4900, opened 2024-09-24
Support multi control statement inside `DoConcurrentLoop`


Pranavchiku:
```fortran
program main
integer :: i, j, res = 0
do concurrent ( i =1:5, j = 1:6 )
    res = res + i * j
end do 
end program
```

Pranavchiku:
Update `ASR.asdl`

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 642534bea..4d0362869 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -35,7 +35,7 @@ stmt
     | Cycle(identifier? stmt_name)
     | ExplicitDeallocate(expr* vars)
     | ImplicitDeallocate(expr* vars)
-    | DoConcurrentLoop(do_loop_head head, expr* shared, expr* local, reduction_expr* reduction, stmt* body)
+    | DoConcurrentLoop(do_loop_head* head, expr* shared, expr* local, reduction_expr* reduction, stmt* body)
     | DoLoop(identifier? name, do_loop_head head, stmt* body, stmt* orelse)
     | ErrorStop(expr? code)
     | Exit(identifier? stmt_name)
```

@certik do the new definition looks fine to you?

ivan-pi:
Attaching this here (this kernel appears in a LU factorization)
```fortran
! trailing_update.f90
subroutine trailing_update(m,n,A,lda,b,c,k)
    integer, parameter :: dp = kind(1.0d0)
    integer, intent(in) :: m, n, lda, k
    real(dp), intent(inout) :: A(lda,n)
    real(dp), intent(in) :: b(m), c(:)
    integer :: i, j
    do concurrent(j=k+1:n,i=k+1:m)
        A(i,j) = A(i,j) - b(i) * c(j)
    end do
end subroutine
```

```
(lf) ~/fortran/reclu$ lfortran -c trailing_update.f90 
semantic error: Do concurrent: exactly one control statement is required for now
  --> trailing_update.f90:8:5 - 10:10
   |
 8 |        do concurrent(j=k+1:n,i=k+1:m)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
10 |        end do
   | ...^^^^^^^^^^ 


Note: Please report unclear or confusing messages as bugs at
https://github.com/lfortran/lfortran/issues.
(lf) ~/fortran/reclu$ lfortran --version
[...]

Issue #4818, opened 2024-09-16
Implementation of collapse() clause for OpenMp support


adit4443ya:
```shell
(lf) aditya-trivedi@Observer:~/thats_me$ ./lfortran/src/bin/lfortran --openmp --link-with-gcc ./testing_codes/t.f90 -o a --openmp-lib-dir=${CONDA_PREFIX}/lib && ./a
semantic error: The clause collapse is not supported yet
  --> ./testing_codes/t.f90:11:5
   |
11 |     !$omp parallel do collapse(2) reduction(+:result)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Pranavchiku:
As discussed we can represent the following OpenMP collapse, please check syntax:

```fortran
program main
integer :: i, j, l
!$omp parallel do private(i)
!$omp do collapse(2) private(j)
!$omp do private(l)
!$omp end do
!$omp end do
!$omp end parallel do
end program
```

TO

```fortran
program main
        integer :: i, j, k, l
        do concurrent (i=1:2)
                do concurrent ( j = 2:4, i=1:2)
                        do concurrent ( l = 12:14)
                        end do 
                end do 
        end do 
end program
```

The point of observation is to use multiple heads at the 2nd level of nesting only to recognise it in openmp pass.

certik:
Isn't it rather equivalent to:
```
program main
        integer :: i, j, k, l
        do concurrent (i=1:2)
                do concurrent ( j = 2:4, l=1:2)

                end do 
        end do 
end program
```

adit4443ya:
As I expected and also verified regarding the usage of the `collapse` clause with conventions:

1. There should not be any intermediate statements between inner nested loops.
2. Therefore, we should only be concerned about the body of the last nested loop.

For more information, refer to the article on [Coding Habits](https://www.admin-magazine.com/HPC/Articles/OpenMP-Coding-Habits-and-GPUs).

Additionally, I checked various other sources, and developers typically use the `collapse` clause with the following assumptions:
- The iterations will be independent of each other.
- There won't be any intermediate code between loops.

adit4443ya:
Also now i would like to propose a way to represent this in the ASR
Let's say for example
we have code
```fortran
[...]

PR #5582, opened 2024-12-04
Feat: Multiple Headed DoConcurrentLoop: Update of each loop variables

adit4443ya:
This PR aims to correctly update the loop variables for each head of DoConcurrentLoop,  continuing #5542 and hence moving towards #4900 and #4818 completion

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1859,6 +1859,7 @@ RUN(NAME openmp_37 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_38 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_40 LABELS llvm_omp)
+RUN(NAME openmp_41 LABELS llvm_omp)
 
 RUN(NAME fortran_01 LABELS gfortran llvm fortran)
 RUN(NAME fortran_02 LABELS gfortran fortran)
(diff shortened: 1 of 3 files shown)


certik:
Is the idea here that we are flattening the nested parallel loops?

adit4443ya:
> Is the idea here that we are flattening the nested parallel loops?

Yeah we can say that 
As we are representing collapse with multiple headed doconcurrent loops

certik (approved):
I think that this is fine.

adit4443ya:
> The code can be further improved as needed, I think it's good to merge.

If this gets merged then both issues 
- Support multi control statement inside DoConcurrentLoop #4900
- Implementation of collapse() clause for OpenMp support #4818
 are resolved as the main aim is to have properly functioning multiple headed doconcurrent loop

certik:
We should add more thorough tests, such as assigning some value to a multidimensional array and such, which might reveal bugs that need to be fixed. But the initial implementation is done, thank you!

adit4443ya:
Yeah I will look at it !! And will report bugs and resolve them right away if there comes any

adit4443ya:
> We should add more thorough tests, such as assigning some value to a multidimensional array and such, which might reveal bugs that need to be fixed. But the initial implementation is done, thank you!
> 
> For example I noticed without `--openmp` it still fails, but it shouldn't. We need to properly implement `DoConcurrent` using nested regular (serial) loops without `--openmp`.



This might be because in the 'do_loop.cpp' we are replacing doconcurrent with do loops and do loops only have one head and hence it is failing..

adit4443ya:
> We should add more thorough tests, such as assigning some value to a multidimensional array and such, which might reveal bugs that need to be fixed. But the initial implementation is done, thank you!
> 
> For example I noticed without `--openmp` it still fails, but it shouldn't. We need to properly implement `DoConcurrent` using nested regular (serial) loops without `--openmp`.

PR #5586

certik:
Thanks!

Pranavchiku (commented):
Thanks for the PR! Looks good to me.

PR #5583, opened 2024-12-04
fix: use a temporary buffer with `fgets()` for formatted read

### LINKED ISSUE #5576: Reading more than 2 lines from a file fails
MRE:
```fortran
program file_read
implicit none
integer :: i
character(len=10) :: tmp
integer :: u
open(newunit=u, file="input.txt", status='old')
do i = 1, 3
    read(u, '(A)') tmp
    print *, i, tmp
end do
close(u)
end program
```
with `input.txt`:
```
xx
yy
zz
```
Behaves like this:
```console
$ gfortran a.f90 && ./a.out
           1 xx        
           2 yy        
           3 zz        
$ lfortran a.f90 
1    xx
2    yy
3    
```
The first two lines are read correctly, but the last line fails.

kmr-srbh:
fixes #5576 and https://github.com/lfortran/lfortran/issues/5545#issuecomment-2513454447.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -3319,13 +3319,35 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
 
     int n = strlen(*arg);
     *arg = (char*)malloc(n * sizeof(char));
+    const char SPACE = ' ';
 
     if (unit_num == -1) {
         // Read from stdin
-        *iostat = !(fgets(*arg, n, stdin) == *arg);
-        (*arg)[strcspn(*arg, "\n")] = 0;
-        va_end(args);
-        return;
+        char *buffer = (char*)malloc((n + 1) * sizeof(char));
+        if (fgets(buffer, n + 1, stdin) == NULL) {
+            *iostat = -1;
+            va_end(args);
+            free(buffer);
+            return;
+        } else {
+            if (streql(buffer, "\n")) {
+                *iostat = -2;
+            } else {
+                *iostat = 0;
+            }
+
+            (buffer)[strcspn(buffer, "\n")] = 0;
+
+            size_t input_length = strlen(buffer);
+            *chunk = input_length;
+            while (input_length < n) {
+                strncat(buffer, &SPACE, 1);
+                input_length++;
+            }
+            strcpy(*arg, buffer);
+            va_end(args);
+            free(buffer);
+        }
     }
 
     bool unit_file_bin;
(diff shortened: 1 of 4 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3320,10 +3320,19 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
...
+        char *tmp_buffer = (char*)malloc((n + 1) * sizeof(char));
+        *iostat = !(fscanf(stdin, "%s", tmp_buffer) == 1);
+        size_t input_length = strlen(tmp_buffer);
+        strcpy(*arg, tmp_buffer);
+        while (input_length < n) {
+            strncat(*arg, &SPACE, 1);
+            input_length++;
+        }
+        (*arg)[n] = '\0';
+
```
certik, line 3335: Should we free `tmp_buffer` here?

certik (approved):
I think this looks great, thanks for fixing it!

Let's merge it, and we can improve upon this with subsequent PRs, such as fix any memory leaks if there are any.

certik:
CI must be fixed.

kmr-srbh:
Yes this CI failure is due to handling of `iostat`. There is some issue. I will look into it.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: Ah, yes. Adding it.

kmr-srbh:
Will finish this PR for `"(a)"` format specifier today. We can later look into supporting the other format specifiers in subsequent PRs if it is high priority.

certik:
`(A)` is good for now.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3319,30 +3319,69 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
...
+            (buffer)[strcspn(buffer, "\n")] = 0;
+
+            size_t input_length = strlen(buffer);
+            *chunk = input_length;
+            while (input_length < n) {
+                strncat(buffer, &SPACE, 1);
+                input_length++;
+            }
+            strcpy(*arg, buffer);
```
gxyd, line 3347: I see that `*arg` is allocate above as `*arg = (char*)malloc(n * sizeof(char));`, while buffer is `char *buffer = (char*)malloc((n + 1) * sizeof(char));`

and the syntax for `strcpy` says:
```c
char *strcpy(char *dest, const char *src)
```

It probably maybe wrong to copy size `n + 1` (`buffer`) to size `n` (`*arg`)

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3319,30 +3319,69 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
...
+            } else {
+                *iostat = 0;
+            }
+
+            (buffer)[strcspn(buffer, "\n")] = 0;
+
+            size_t input_length = strlen(buffer);
+            *chunk = input_length;
+            while (input_length < n) {
+                strncat(buffer, &SPACE, 1);
```
gxyd, line 3344: I'm looking at the documentation for `strncat` https://en.cppreference.com/w/c/string/byte/strncat, it says:
```
Return value
1) returns a copy of dest
2) returns zero on success, returns non-zero on error. Also, on error, writes zero to dest[0] (unless dest is a null pointer or destsz is zero or greater than RSIZE_MAX).
```

which makes me think that this isn't an efficient way of padding spaces. Why don't we simply use a `for loop` and assign the remaining character of `buffer` as `buffer[i] = ' ';` I think that would be more efficient than using `strncat`

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3319,30 +3319,69 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
 
     int n = strlen(*arg);
     *arg = (char*)malloc(n * sizeof(char));
+    const char SPACE = ' ';
```
gxyd, line 3322: could you please help me understand, that if this function `_lfortran_formatted_read` is run for every line we encounter?

e.g. for
```
xx
yy
zz
```
so this function `_lfortran_formatted_read` will be run three times?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3319,30 +3319,69 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
...
+        } else {
+            if (streql(buffer, "\n")) {
+                *iostat = -2;
+            } else {
+                *iostat = 0;
+            }
+
+            (buffer)[strcspn(buffer, "\n")] = 0;
+
+            size_t input_length = strlen(buffer);
```
gxyd, line 3341: since we already used `strcspn` above to find the character where the first `\n` occurs, we can find `input_length` from that?

gxyd (commented):
I left a few comments, every time I read C code, it usually takes me a long time to review it, because of not so ease with debugging. I'm sure that I haven't reviewed it carefully, but left a few comments we can think about.

certik (approved):
I think it's good enough to merge, since it fixes a bug that now works. However, let's improve the code based on @gxyd's suggestions in subsequent PRs.

PR #5585, opened 2024-12-05
Adds ASR type to document_symbols for LSP

dylon:
Addresses respective TODO from https://github.com/lfortran/lfortran-lsp/issues/6.


certik:
This is probably fine, but see my comment here: https://github.com/lfortran/lfortran-lsp/pull/19#discussion_r1870471310.

src/libasr/lsp.cpp
```diff
@@ -13,6 +14,50 @@
...
+    case ASR::symbolType::Variable:
+        return LSP::SymbolKind::Variable;
+    case ASR::symbolType::Class:
+        return LSP::SymbolKind::Class;
+    case ASR::symbolType::ClassProcedure:
+        return LSP::SymbolKind::Method;
+    case ASR::symbolType::Template:
+        return LSP::SymbolKind::TypeParameter;
+    default:
+        return LSP::SymbolKind::Function;
```
certik, line 57: Here we should probably throw an exception in default, so that if we add more symbols, it shows the exception instead of silently returning Function.

certik (approved):
I think this is good. It nicely maps into the LSP definitions, so it is invariant of our internal ASR data structures, so we have the freedom to change them, without breaking this API.

PR #5590, opened 2024-12-05
preprocessing: use `in_string` only when position has moved forward

gxyd:
## Description

This PR has the changes from https://github.com/lfortran/lfortran/pull/5569 and then builds on top of it, this is definitely not intended to be reviewed until https://github.com/lfortran/lfortran/pull/5569 has been reviewed and merged.

## TODO

I'll add test cases as well for this PR

### DIFF
--- a/src/lfortran/parser/parser.cpp
+++ b/src/lfortran/parser/parser.cpp
@@ -688,7 +688,7 @@ std::string prescan(const std::string &s, LocationManager &lm,
                 process_include(out, s, lm, pos, fixed_form, include_dirs, col);
             }
             newline = false;
-            if (s[pos] == '!') in_comment = true;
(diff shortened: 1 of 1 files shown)


Thirumalai-Shaktivel:
Pushing with no commits closes the PR.

Try to push with some commits to the same branch.
I have reopened this PR.

gxyd:
> Pushing with no commits closes the PR.

Ah yes, thank you. I think now it has the change that I intend to test it at the CI, and seems like the CI is running, so it's good now. Thanks @Thirumalai-Shaktivel

gxyd:
There is a definitely a test case which we need to extract here, but unfortunately it's on ubuntu, and I've a macOS machine.

PR #5593, opened 2024-12-05
Rename Union/UnionType

certik:
Make UnionType the `type`, and Union the `symbol`.


Pranavchiku (approved):
You may need to update test references

certik:
No test references changed. I think we only test this in LPython, not LFortran (yet). We should add Fortran extensions and test this, but that's for later.

Issue #5292, opened 2024-11-08
Give an error if the variable in `parameter` is not declared


certik:
```fortran
program foo2
implicit none
parameter(foo=1)
end program
```
GFortran:
```console
$ gfortran a.f90
a.f90:3:13:

    3 | parameter(foo=1)
      |             1
Error: Symbol ‘foo’ at (1) has no IMPLICIT type
```
LFortran:
```console
$ lfortran a.f90
$
```

captainhaddock18:
@certik 
Hi there, I am trying to solve this issue. is this right way to start? I saw a similar for the save attribute. I am a little confused what needs to be done if implicit none is active and type is also specified.
![image](https://github.com/user-attachments/assets/7157ca5c-f0c0-4757-a3e9-997d3c59c987)

captainhaddock18:
I think this works for parameter attribute. `implicit-typing` can now be enabled for parameter attribute as well. Also, the error is thrown if not declared earlier
![image](https://github.com/user-attachments/assets/575d6da6-90f4-4b4f-8faf-fb65ba8a39ee)

Pranavchiku:
The only concern is if we use `implicit none` with `--implicit-typing` the `implicit_dictionary` will be empty so you need to throw error there as well, rest it looks good. Nice catch.

PR #5574, opened 2024-12-04
Fix issue #5292: Error Handling When Parameter Not Declared

captainhaddock18:
Error handling check implemented when the parameter is not declared. implemented using the `implicit-typing` option. Fixes issue #5292.

PR #5594, opened 2024-12-05
fix: parameter not declared

captainhaddock18:
Fixes the error #5292 . Checks if not declared and also enables `--implicit-typing`. tests for this has also been added.


certik:
Can you please address my questions from https://github.com/lfortran/lfortran/pull/5574 ?

captainhaddock18:
> Can you please address my questions from #5574 ?

Apologies, I missed it there 

1) keeping `Parameter is not declared` is a great option but the error should also make the user aware that he can use the --implicit-typing option with the parameter.  

2) the error `Parameter `" + sym + "` has no IMPLICIT Type` arises when one uses both --implicit-typing option and implicit none together.
 
for example, the below code
`program test_implicit_typing_disabled
    implicit none
    parameter (a = 3.14)
end program
`
gives an output
image

certik:
We currently don't suggest to the user that `--implicit-typing` can be used when a variable is not declared. So to be consistent, we should not here either. Either do it for all, or for none.

Is this the right default choice? I think so, since implicit typing should not be used for new codes. And for old codes, typically you are aware of this. Where we should suggest it is for "implicit real ..." implicit statements. But just for not declaring a variable I think we should not, since the user should declare it. And if the user knows about implicit typing, he can enable it, but we should not encourage it.

captainhaddock18:
> We currently don't suggest to the user that `--implicit-typing` can be used when a variable is not declared. So to be consistent, we should not here either. Either do it for all, or for none.
> 
> Is this the right default choice? I think so, since implicit typing should not be used for new codes. And for old codes, typically you are aware of this. Where we should suggest it is for "implicit real ..." implicit statements. But just for not declaring a variable I think we should not, since the user should declare it. And if the user knows about implicit typing, he can enable it, but we should not encourage it.
> 
> Regarding 2, given that `implicit none` was used, it disabled implicit typing, so I think the error message should still say "not declared".

Got it. I'll make these corrections. Both the errors would now ask to declare the variable and not encourage to use the implicit-typing. Thanks a lot!

integration_tests/implicit_typing_03.f90
```diff
@@ -1,5 +1,5 @@
 module implicit_typing_03_first
-  parameter (x=90)
+  integer,parameter :: x=90
```
certik, line 2: We still need to test the original file, so just add this new code as a new test.

tests/errors/parameter_declared2.f90
```diff
@@ -0,0 +1,5 @@
+program parameter_declared2
+    implicit none
+    real, parameter :: a = 3.14
+    print *, a
+end program
```
certik, line 5: These two tests have to be registered in `tests/tests.toml`.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2470,7 +2470,29 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                                    }));
+                                                throw SemanticAbort();
+                                            }
+                                        } else {
+                                            diag.add(Diagnostic(
+                                                "Parameter `" + sym + "` is not declared`",
+                                                Level::Error, Stage::Semantic, {
+                                                    Label("",{x.base.base.loc})
+                                                }));
+                                            throw SemanticAbort();
```
certik, line 2494: Let's add tests for all three branches here:
* the `if/then if/then`
* the `if/then if/else`
* the `if/else`

I think you probably have the tests, but they need to be registered and we need to ensure they actually test all three branches.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2470,7 +2470,29 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                        if (compiler_options.implicit_typing) {
+                                            ASR::ttype_t* type = implicit_dictionary[std::string(1,sym[0])];
+                                            if (type) {
+                                                sym_ = declare_implicit_variable(x.m_syms[i].loc, sym, ASR::intentType::Local);
+                                                ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym_);
+                                                v->m_storage = ASR::storage_typeType::Parameter;
+                                                assgnd_storage[sym] = std::make_pair(ASR::storage_typeType::Parameter, s.m_initializer);
+                                            } else {
+                                                diag.add(Diagnostic(
+                                                    "Parameter `" + sym + "` is not declared",
```
certik, line 2482: This error happens with `--implicit-typing` and having `implicit none`?

If so, maybe the error can be:
```suggestion
                                                    "Parameter `" + sym + "` is not declared and no implicit type specified",
```
To distinguish it from the error below.

certik:
@captainhaddock18 thanks for working on this. I left some comments how to finish the PR. Let me know if you have any questions.

Issue #5595, opened 2024-12-05
Split asr.h into smaller parts


certik:
Split `asr.h` into smaller parts, and only use what we need; the stock `asr.h` should only contain the ASR definitions that are needed in all of the code base, but specialied visitors, possibly all visitors should be separate.

We have the following parts that we need to split:
```
$ grep "^class" -- src/libasr/asr.h
class BaseVisitor
class BaseWalkVisitor : public BaseVisitor<StructType>
class TreeBaseVisitor : public BaseVisitor<StructType>
class PickleBaseVisitor : public BaseVisitor<StructType>
class JsonBaseVisitor : public BaseVisitor<StructType>
class SerializationBaseVisitor : public BaseVisitor<StructType>
class DeserializationBaseVisitor : public BaseVisitor<StructType>
class DefaultLookupNameVisitor : public BaseVisitor<StructType>
class ASRPassBaseWalkVisitor : public BaseVisitor<StructType>
class BaseExprStmtDuplicator {
class BaseExprReplacer {
class BaseStmtReplacer {
class CallReplacerOnExpressionsVisitor : public BaseVisitor<StructType>
```

certik:
These were all split in https://github.com/lfortran/lfortran/pull/5597.
```
$ grep "^class" -- src/libasr/asr.h
$
```

PR #5597, opened 2024-12-06
Split asr.h into individual header files

certik:
The main file went from 50K lines to less than 5K lines (over 10x reduction in size). Currently most of the visitor header files are still included in `asr_utils.h`, so the compilation speed reduction is not huge, but now we can refactor `asr_utils.h` and move most visitors into `.cpp` files.

Towards https://github.com/lfortran/lfortran/issues/5595.


Pranavchiku (commented):
Looks good, thank you!

PR #5605, opened 2024-12-07
fix: Add two new options `--linker` and `--linker-path`

Thirumalai-Shaktivel:
Now, there are two ways to specify the linker and their path:
- Using `--linker` and `--linker-path` options
- Using environment variable `LFORTRAN_LINKER` and `LFORTRAN_LINKER_PATH`

Usage:
```console
$ lfortran x.f90 --linker=gcc
$ lfortran x.f90 --linker-path=/usr/bin
$ lfortran x.f90 --linker=gcc --linker-path=/usr/bin
```

```console
$ export LFORTRAN_LINKER=gcc
$ export LFORTRAN_LINKER_PATH=/usr/bin/
$ lfortran x.f90
```

Towards: https://github.com/lfortran/lfortran/issues/5526

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1597,8 +1597,8 @@ int link_executable(const std::vector<std::string> &infiles,
     const std::string &outfile,
     const std::string &runtime_library_dir, Backend backend,
     bool static_executable, bool shared_executable,
-    bool link_with_gcc, bool kokkos, bool verbose,
-    const std::vector<std::string> &lib_dirs,
+    std::string linker, std::string linker_path, bool kokkos,
+    bool verbose, const std::vector<std::string> &lib_dirs,
     const std::vector<std::string> &libraries,
     const std::vector<std::string> &linker_flags,
     CompilerOptions &compiler_options)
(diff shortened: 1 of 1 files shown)


Thirumalai-Shaktivel:
Yes, on Linux the gcc fails, see: https://github.com/lfortran/lfortran/actions/runs/12210214344/job/34066039614?pr=5605, so using clang as default for all the platform

certik (approved):
I think that's good, thanks!

src/bin/lfortran.cpp
```diff
@@ -1731,21 +1731,35 @@ int link_executable(const std::vector<std::string> &infiles,
...
+
+            if (!linker.empty()) {
+                CC += linker;
+            } else if (char *env_linker = std::getenv("LFORTRAN_LINKER")) {
+                CC += env_linker;
+            } else {
+                // TODO: Add support for msvc linker for Windows
+                // TODO: Add support for lld linker
+                // Default linker to be used
+                CC += "clang";
```
certik, line 1758: We need to fix Windows, I am surprised that this works. I think we are still using `msvc` elsewhere in the driver.

PR #4913, opened 2024-09-25
feat: add `./.vscode/launch.json` for setting debugger

Pranavchiku:

PR #5610, opened 2024-12-07
chore: untrack `.vscode/launch.json`

kmr-srbh:


kmr-srbh:
I think we should not track the editor files with git. When using the debugger with changes to `launch.json`, I always had to worry about not pushing it with the commit. Now we can freely use a debugger according to our needs.

gxyd:
I did highlight a possibility of similar concern here: https://github.com/lfortran/lfortran/pull/4913#pullrequestreview-2335708410, when we commited this change to our repo.

> When using the debugger with changes to launch.json, I always had to worry about not pushing it with the commit.

Do you often use `git add .`? Because, I personally never have had any problem with it's presence.

kmr-srbh:
> Do you often use git add .? Because, I personally never have had any problem with it's presence.

Yes, when committing test references, I stage all of them at once. I have never accidentally committed changes to `launch.json`, but it adds some friction. The real problem occurs when rebasing branches, when I have to undo the changes and make them again.

certik:
I would keep it in the repos, since this makes debugging work. Without it I always struggle how to set it up.

What modifications do you need to do in this file?

When creating commits, if I have dirty changes, I use `git add -p` to only select changes that I want.

kmr-srbh:
> What modifications do you need to do in this file?

I modify the `args` field and sometimes use `gdb` instead of `lldb`.

If we need this file, we may surely keep it.

gxyd:
> I modify the args field and sometimes use gdb instead of lldb.

That I often do as well, the modification of the filename to debug.

certik:
I think you can add a section specifically for `gdb`, then you can select it from VSCode.

Regarding the binary to debug and command line arguments (if any), maybe there is some way to do this without modifying the file. I usually modify it for that, as I don't know any other way.

kmr-srbh:
Closing this PR as it is not required. I will open a new PR for adding `gdb` as a debugger.

PR #5612, opened 2024-12-07
fix: Don't Remove ExternalSymbol which is used in `m_type_declaration` of variable 

### LINKED ISSUE #5603: (PRIMA) use actual function instead of its type in `m_type_declaration` of Variable
It is just one step ahead of #5500 

MRE : 
```fortran
module procedure_07_module
contains
    subroutine cb(x)
        implicit none
        integer, intent(in), optional :: x(:)
    end subroutine cb

    subroutine calfun(x)
        implicit none
        integer, intent(in), optional :: x(:)
        logical :: y
        y = present(x)
        if (y) error stop
    end subroutine calfun
end module procedure_07_module

program procedure_07
    use procedure_07_module
    call temp(calfun)
contains
    subroutine temp(call_back)
        implicit none
        procedure(cb) :: call_back
        call call_back()   
    end subroutine temp
end program procedure_07
```

LFortran Error : 
```console
code generation error: Subroutine code not generated for 'cb'
```

jinangshah21:
Fixes #5603.
But there is one issue at https://github.com/lfortran/lfortran/issues/5603#issuecomment-2525200055 which can be looked up in next PR.


certik (approved):
I think that this is fine, thanks!

PR #5613, opened 2024-12-07
Fix `ArrayPhysicalCast` in `pass_array_by_data` pass

### LINKED ISSUE #5521: (PRIMA) Nested function calls with array as parameters Causes Error
MRE : 
```fortran
program test 
    implicit none
    real(8) :: y(2,2)
    real(8) :: x(2) = [1,2]
    real(8) :: z(size(y,2))
    integer :: j = 1
    z = matprod12(x, y)
    print *, z

contains
function matprod12(x, y) result(z)
    implicit none
    real(8), intent(in) :: x(:)
    real(8) :: z(size(y, 2))
    integer :: j 
    j = size(y,2)
    z(j) = inprod(x, y(:, j))
end function matprod12

    implicit none
    real(8), intent(in) :: x(:)
    real(8), intent(in) :: y(:)
    real(8) :: z 
end function inprod
end program
```

LFortran Error : 
```console
double* %x
 %array*  %47 = call double @inprod(double* %x, %array* %46)
```

advikkabra:
1. Set the `m_old` parameter to the new physical type of the array.
2. Set the dimensions to the array dimensions if necessary.

Creates valid ASR instead of wrong type being passed.

Fixes #5521


certik (approved):
I think that this is good, thanks for the fix!

PR #5615, opened 2024-12-07
Remove the cpptranslate binary

certik:
This functionality wasn't finished anyway. And we should use `lfortran` itself to do the C++ translation (currently via the `--show-cpp` option).


Thirumalai-Shaktivel (approved):

gxyd (approved):
LGTM!

PR #5616, opened 2024-12-07
CMake: Do not use CMAKE_VERBOSE_MAKEFILE

certik:
This option gives a warning when we use Ninja.


gxyd (approved):
Can confirm that this works, as on *main* when building, I get the below warning:
```console
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_VERBOSE_MAKEFILE

```

while on this branch, I don't get any such warning.

Do we want to do the same in CI.yml file as well? I see it being used at two places and they also generate a similar warning. (Ofcourse, these changes are more important than CI.yml changes, as we use this build locally more often ourselves)

certik:
Yes, we can do it at the CI also.

PR #5617, opened 2024-12-07
Add WITH_BENCHMARKS cmake option

certik:
Keep this option off by default for speed of compilation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,6 +282,9 @@ if (WITH_FMT)
     find_package(fmt)
 endif()
 
+set(WITH_BENCHMARKS no
+    CACHE BOOL "Build benchmarks")
+
 # Stacktrace
 set(WITH_UNWIND no
     CACHE BOOL "Build with unwind support")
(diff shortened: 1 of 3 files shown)


gxyd:
> Keep this option off by default for speed of compilation

I tested it on my machine as well, I get some improvement in build time, on current *main*:
```console
> time ./build1.sh
./build1.sh  250.35s user 21.04s system 492% cpu 55.142 total
```

while on this branch:
```console
> time ./build1.sh
./build1.sh  245.00s user 14.97s system 518% cpu 50.092 total
```

gxyd:
The CI failed for *Test GFortran, Debug Build, Fortran, OpenMP, C/C++ backend, Upload Tarball, CPP Build, WASM* CI job, specifically in *Test CPP*, so I've pushed one commit, let's see if the CI passes or not now.

gxyd (approved):
The CI passes, the change that you made @certik looked good to me. Once you're happy with the small fix that I pushed, we can go ahead with this PR's merge.

certik (commented):
This is great, thanks for the fix @gxyd, now the benchmarks are still tested, which is good, but not on by default.

Yes, it's a small improvement in the speed of compilation, but all these little improvements will add up.

PR #5618, opened 2024-12-07
Support full KIND specifications in IMPLICIT type declarations

### LINKED ISSUE #5549: IMPLICIT declaration_type_spec

pjh40:
This PR addresses deficiencies in the parsing of `implicit_spec` rules, where kind arguments had to either be an integer or an identifier.  While it is not possible to simply use a grammar rule like:
```bison
implicit_spec
    : declaration_type_spec "(" letter_spec_list ")" {
           $$ = IMPLICIT_SPEC($1, $3, @$); }
    ;
``` 
due to essential shift/reduce conflicts in the parser, we can use `kind_arg_list` instead of `letter_spec_list`, and produce rules like:
```bison
implicit_spec
    : KW_INTEGER "(" kind_arg_list ")" "(" kind_arg_list ")" {
            $$ = IMPLICIT_SPEC(ATTR_TYPE_KIND(Integer, $3, @$), $6, @$); }
    | KW_INTEGER "*" TK_INTEGER "(" kind_arg_list ")" {
            $$ = IMPLICIT_SPEC(ATTR_TYPE_INT(Integer, $3, @$), $5, @$);
	    WARN_INTEGERSTAR($3, @2);}
    | KW_INTEGER "(" kind_arg_list ")" {
            $$ = IMPLICIT_SPEC(ATTR_TYPE(Integer, @$), $3, @$); }
    ;
```
Note that the _declaration-type-spec_ portion of the `implicit_spec` *must* be included directly in each alternative: trying to move them to another rule brings back the shift/reduce conflicts.  However, the only thing this affects is the format of warnings on Sun-style `*` declarations (e.g. `INTEGER*4`).

The basic changes are:

1.  The `ImplicitSpec` no longer contains an a `kind` description: all kind information is embedded in the Type AST.
2.  Rather than converting from letter-spec to a simplified kind, we convert from a `kind_arg` to a `letter_spec`
3. Warnings are now enabled on Sun-style declarations in implicit
4. The full `KIND` grammar is accepted for all declaration types
5. Errors in `COMPLEX*<n>` warning messages have been corrected. 

Thanks to @certik for all the guidance on this effort!

Closes #5549


tests/reference/asr-implicit10-23a5156.stderr
```diff
@@ -4,14 +4,14 @@ style suggestion: Use real(8) instead of real*8
 6 | real*8 :: g, h, i
   | ^^^^^^ help: write this as 'real(8)'
 
-style suggestion: Use complex(8) instead of complex*8
+style suggestion: Use complex(4) instead of complex*8
```
gxyd, line 7: I'm sorry, could you help me understand why the new style suggestion is to be used instead of the older one here?

gxyd: This seems very much intentional, as mentioned in the PR description:
> Errors in COMPLEX*<n> warning messages have been corrected.

I'm just not able to understand, why this is a correct change.

gxyd: And a declaration like `real * 8` is what you call a *sun-style* declaration?

src/lfortran/parser/semantics.h
```diff
@@ -446,20 +446,58 @@ ast_t* data_implied_do(Allocator &al, Location &loc,
 #define IMPLICIT_NONE_EXTERNAL(l) make_ImplicitNoneExternal_t(p.m_a, l, 0)
 #define IMPLICIT_NONE_TYPE(l) make_ImplicitNoneType_t(p.m_a, l)
 
+
+static inline char * id_if_Name_t(expr_t const * exprNode) {
+  char *id = nullptr;
```
gxyd, line 451: The indentation in this function and `vec_kind_item2ast` isn't consistent, would be easier to read if it was though.

gxyd (approved):
Looks really good to me, left one question and one request for a change for indentation.

tests/reference/asr-implicit10-23a5156.stderr
pjh40: Yes, the syntax like `real*8` was introduced in a Sun Microsystems Fortran-77 compiler a long time ago: it has never been part of the Fortran standard (unlike `character*12`).  As for the `complex*16` -> `complex(8)` change, `complex*16` meant a complex that occupied 16 bytes: what we would call a double precision complex.  The kind value for that type in LFortran is 8, so you would rewrite `complex*16` as `complex(kind=8)`.

src/lfortran/parser/semantics.h
pjh40: I'm struggling to find an indentation style that is used consistently.  Does the project have a target style (GNU, LLVM, etc)? Is the `.clang-format` file representative of your preferred style?  I'm happy to change it once I know what you're looking for.

src/lfortran/parser/semantics.h
```diff
@@ -1778,14 +1816,14 @@ void add_ws_warning(const Location &loc,
         } else if (end_token == yytokentype::KW_COMPLEX) {
             if (a_kind == 16) {
                 diagnostics.parser_style_label(
-                "Use complex(16) instead of complex*16",
+                "Use complex(8) instead of complex*16",
```
certik, line 1819: Ah, is `complex(8)` composed of two `real(8)`, but that is specified as `complex*16`? I think you are right, thanks for spotting this!

tests/implicit1.f90
```diff
@@ -40,5 +40,14 @@ program implicit1
 
 IMPLICIT TYPE(BLOB) (A)
 IMPLICIT class(X) (A-b)
-IMPLICIT procedure(Y) (A-b)
```
certik, line 43: I think we still need to test for `IMPLICIT procedure(Y) (A-b)`.

tests/reference/asr-implicit10-23a5156.stderr
certik: I think we do it correctly in AST->ASR, but we gave a wrong warning in the parser. Thanks for correcting it!

tests/implicit1.f90
pjh40: Is there a non-standard use case for this? It isn't legal in an IMPLICIT statement...

certik (approved):
Awesome, thanks for figuring this out and fixing it! There are some minor indentation issues, that can be fixed later.

I think we want to put this test case back:
https://github.com/lfortran/lfortran/pull/5618/files#r1874978450
But that can be done later also.

So I am going to merge it. Thank you again and thanks @gxyd for the review!

tests/implicit1.f90
certik: I see. If it is not legal, then let's not do it. I might have put it in as an example of a more complex implicit statement, not realizing that it is not standard. So let's not parse it, if we discover it in the wild, we can add it as an extension later.

src/lfortran/parser/semantics.h
certik: Yes, we should add some `.clang-format`. We use 4 spaces and the `} else {` style, other than that different contributors have different styles.

PR #5620, opened 2024-12-08
DEV: Duplicate type to avoid mis-updates

czgdp1807:
Using pointers is always risky in C++. Duplicating types in helps in avoiding same pointer present as types of two different ASR objects.

We need to have an ASR verify check to make sure one pointer is present at one place for all ASR objects except symbols. Symbol pointers can be present at multiple places.


czgdp1807 (commented):
Enabling auto merge on this PR because it is a harmless change. In fact it increases safety of code written. Avoids in a ASR::ttype pointer present at two places in ASR.

certik:
It's fine. I think types should either be duplicated, or we need some mechanism to reuse it. I think they can probably be reused, if it speeds up the compilation. But I would play with this after beta. Until beta our job is to get things done robustly and correctly, which this PR does I think.

Issue #4397, opened 2024-06-26
Compiling PRIMA


shivimathur:
To compile PRIMA with GFortran : 
```console
git clone --depth 1 https://github.com/shivimathur/prima
cd prima 
git checkout -t origin/lf
cmake -S . -B build -DCMAKE_INSTALL_PREFIX=install -DCMAKE_Fortran_COMPILER=gfortran
cmake --build build --target install
```

HarshitaKalani:
@shivimathur , as you proceed, can you keep pushing the workarounds in your branch and keep linking the related issues with this issue?

certik:
@shivimathur awesome, thanks! Great progress. Go ahead and submit the PR to LFortran with the missing functions.

HarshitaKalani:
Great job @shivimathur .
Apart from the iso_c_binding, I'm getting the following errors:
```console
semantic error: The symbol 'c_sizeof' not found in the module 'iso_c_binding'
   --> memory_preprocessed.F90:134:1
    |
134 | use iso_c_binding, only: c_sizeof
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
semantic error: Module 'memory_mod' not declared in the current source and the modfile was not found
   --> fortran/common/string.f90:200:1
    |
200 | use, non_intrinsic :: memory_mod, only : safealloc
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
semantic error: Module 'memory_mod' not declared in the current source and the modfile was not found
    --> fortran/common/linalg.f90:1138:1
     |
1138 | use, non_intrinsic :: memory_mod, only : safealloc
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
semantic error: Module 'linalg_mod' not declared in the current source and the modfile was not found
  --> fortran/common/evaluate.f90:33:1
   |
33 | use, non_intrinsic :: linalg_mod, only : trueloc
[...]

PR #5624, opened 2024-12-09
enh: udpate function variable when `m_type_declaration` is updated

jinangshah21:
Towards #4397
Previously function Variable was only updating when call is made but now it updates when we declare `procedure(cb) :: call_back`.


certik (approved):
I think that looks good, thanks!

certik:
(the first commit has a typo in the commit log "functin" -> "function")

PR #5631, opened 2024-12-09
Parse more fixed-form function return types

### LINKED ISSUE #5350: Infinite loop in FixedFormRecursiveDescent
Trying to parse:
```fortran
      CHARACTER*10 FUNCTION CF716(IVD001)
      END                               
```
LFortran SHA 4068df32ee4 enters an infinite loop in `LCompilers::LFortran::FixedFormRecursiveDescent`. It appears that the `FixedFormRecursiveDescent::is_declaration` function is trying to match against an input string of `0functioncf716(`.  This seems to point to two issues:

1. The `CHARACTER*<N>` parser is not greedy enough getting the repeat count (1 instead of 10?); and
2. There doesn't seem to be a non-match exit criterion from `is_declaration`.

pjh40:
This PR allows the fixed form parser to parse declarations of the form `<type>*<n> ... function` where `<n>` can have an arbitrary number of digits.  Previously, the number of digits was hard-coded to each type, and would enter an infinite loop if it encountered an unexpected value.

Also adds a test for these declarations.

This allows test `FM715.f` from the FCVS_95 suite to run successfully.

Closes #5350


certik (approved):
Thanks, this looks good!

certik:
I gave you push access, you can merge it if tests pass.

pjh40:
Need to also add `character*(*)`!

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1626,14 +1626,26 @@ struct FixedFormRecursiveDescent {
...
         std::vector<std::string> decls{keywords.begin(), keywords.end()};
         while(decls.size() != 0) {
             for (unsigned int i=0;i<decls.size();++i) {
                 if (next_is(cpy, decls[i])) {
                     kw_found.push_back(decls[i]);
                     cpy += decls[i].size();
+                    if (decls[i].back() == '*') {
+ 		        if (cpy[0] == '(' &&
+			    cpy[1] == '*' &&
+			    cpy[2] == ')') {
```
certik, line 1640: You have to check that `cpy[i]` is not `'\0'`, otherwise you are reading unallocated memory.

certik (approved):
Otherwise looks great, thanks!

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -331,25 +331,23 @@ struct FixedFormRecursiveDescent {
...
-            }
-            cur2++;
-        }
-        std::string next_str = std::string((char *)tok, cur2 - tok);
-        return next_str == str;
+    constexpr bool next_is(unsigned char const *cur, const std::string &str) {
+	for(const char s : str) {
+	    if (*cur++ != s) return false;
+	}
+	return true;
```
certik, line 338: I think you need to keep checking for `'\0'` like the original code, since the input can end at any point.

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -331,25 +331,23 @@ struct FixedFormRecursiveDescent {
...
-        } else {
-            return false;
-        }
+    // Are the next characters in the `cur` stream equal to `str`?
+    constexpr bool next_is(unsigned char const * cur, const char *str) {
+	while(*str) {
+	    if (*cur++ != *str++) return false;
+	}
+	return true;
+    }
```
certik, line 347: As long as both `cur` and `str` are null terminated, then this probably cannot fail, although it's really easy to make a mistake in code like this. :)

pjh40: Yeah, if either is not null-terminated, you're in trouble regardless of how you do it.

src/lfortran/parser/fixedform_tokenizer.cpp
pjh40: You beat me to it... there is a fix coming in for this.

pjh40: Now, if `*cur` is `\0` (at a point where there are still characters in `str`), it can't match anything in `str`, so you return false.

certik (approved):
I think that looks good, thanks for the fixes!

You can merge it if it is ready.

Issue #5516, opened 2024-11-29
(PRIMA) Arraysection inside Array Reference throws error 


jinangshah21:
MRE : 
```fortran
program test 
integer :: A(3) = [1,2,3]
integer :: X(2) = [1,2]
integer :: Y = 2
A(X(:)) = Y
end program
```

LFortran Error : 
```console
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/mnt/c/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-LcI20x/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 2725, in main
    return main_app(argc, argv);
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 2664, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/mnt/c/lfortran/src/bin/lfortran.cpp", line 1070, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/mnt/c/lfortran/src/lfortran/fortran_evaluator.cpp", line 382, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    compiler_options, run_fn, infile);
  File "/mnt/c/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 10383, in LCompilers::asr_to_llvm(LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LCompilers::PassManager&, LCompilers::CompilerOptions&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    v.visit_asr((ASR::asr_t&)asr);
[...]

PR #5633, opened 2024-12-10
Handling Assignment for Array with Array Argument to represent Array index

AnshMehta1:
Towards #5516 

Tries to make assignment with Array Item as Array reference.

MRE 
```
program array_indices_array_item_assignment 
integer :: A(3) = [1,2,3]
integer :: X(2) = [1,2]
integer :: Y = 2
A(X) = Y
print * , A
end program
```

LFortran result
```
(lf) lordansh@LordAnsh:~/dev/lfort3/lfortran$ ./src/bin/lfortran try.f90
2    2    3
```

Still Array Section (like A(X(:)) = Y ) is not addressed in this PR , which I will address in subsequent PR.

### DIFF
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -2035,7 +2035,68 @@ class ArrayOpVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisit
                     pass_result.push_back(al, auxiliary_assign_stmt_);
                     resultvar2value[replacer.result_var] = *current_expr;
                 } else {
-                    resultvar2value[replacer.result_var] = array_broadcast->m_array;
+                    if (ASR::is_a<ASR::ArrayItem_t>(*x.m_target)) {
+                        ASR::ArrayItem_t *array_item = ASR::down_cast<ASR::ArrayItem_t>(x.m_target);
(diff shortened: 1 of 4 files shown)


certik (approved):
I think that this is fine. Wasn't this already fixed?

AnshMehta1:
Will extend this logic towards Array Section as Array reference in next PR , as that is our main concern for PRIMA

src/libasr/pass/array_op.cpp
```diff
@@ -2035,7 +2035,68 @@ class ArrayOpVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisit
                     pass_result.push_back(al, auxiliary_assign_stmt_);
                     resultvar2value[replacer.result_var] = *current_expr;
                 } else {
-                    resultvar2value[replacer.result_var] = array_broadcast->m_array;
+                    if (ASR::is_a<ASR::ArrayItem_t>(*x.m_target)) {
+                        ASR::ArrayItem_t *array_item = ASR::down_cast<ASR::ArrayItem_t>(x.m_target);
```
Pranavchiku, line 2039: Handle `ArraySection_t` here, make a generic `expr_target` and assign it accordingly. Also take care of `lbound` and `rbound` if specified in `ArraySection`.

PR #5638, opened 2024-12-10
rem: unnecessary granular catch at `visit_Name`

Pranavchiku:
I think this should fix https://github.com/lfortran/lfortran-lsp/issues/21


certik (approved):
I think this is good. Can this test be simply added to our "continue_compilation" large files? We don't need a specific file for this, do we?

Pranavchiku:
Yes, we don't need it, I'll do it.

gxyd (approved):
Looks good to me.

I wanted to test whether it actually fixes the linked issue in lfortran-lsp or not, but I couldn't properly install it, so posted my query here: https://lfortran.zulipchat.com/#narrow/channel/452774-Internal/topic/LSP/near/488569721

Issue #5647, opened 2024-12-11
Bug : `Arraysize` on arrays with variable dimensions fails at when the variable doesn' t match size kind


assem2002:
MRE
```fortran
PROGRAM subroutines_12

    IMPLICIT NONE
    
    INTEGER(2) :: nang = 32767
    
    CALL my_subroutine
    
    CONTAINS
    
        SUBROUTINE my_subroutine
        
            INTEGER, DIMENSION(nang) :: fxhv
            ! both size prints fail
            print *, size(fxhv, kind = 1)
            print *, size(fxhv, kind = 4) 
            
        END SUBROUTINE my_subroutine

    
    END PROGRAM subroutines_12
    
```

They fail due to incompatibility with the variables the we fetch the actual dimension length from and the kind of the `size` return type.

PR #5648, opened 2024-12-11
Fix : Get dimension length kind and Array size kind compatible

### LINKED ISSUE #5646: Bug : calculating the size for array based on kind=8 variable fails
MRE 
```fortran
PROGRAM subroutines_12

    IMPLICIT NONE
    
    INTEGER(8) :: nang = 2
    
    CALL my_subroutine
    
    CONTAINS
    
        SUBROUTINE my_subroutine
    
            INTEGER, DIMENSION(nang) :: fxhv
    
        END SUBROUTINE my_subroutine
    
    END PROGRAM subroutines_12
    
```

```console
; ModuleID = 'LFortran'
source_filename = "LFortran"

@subroutines_12.nang = internal global i64 2

.entry:
  call void @my_subroutine()
  br label %return

  ret i32 0
}

define void @my_subroutine() {
.entry:
  %0 = load i64, i64* @subroutines_12.nang, align 4
  %1 = mul i32 1, i64 %0
  %2 = mul i32 %1, 4
  %3 = call i8* @_lfortran_malloc(i32 %2)
  %4 = bitcast i8* %3 to i32*
  br label %return

  call void @_lfortran_free(i8* %3)
  ret void
}

declare i8* @_lfortran_malloc(i32)

declare void @_lfortran_free(i8*)

  %1 = mul i32 1, i64 %0

  %1 = mul i32 1, i64 %0



```

assem2002:
fixes https://github.com/lfortran/lfortran/issues/5647
fixes  #5646


certik (approved):
As a temporary fix this is fine, however can you open up an issue and document there how to fix this properly and link it from the source code?

We need to start reducing this technical debt, so let's at least have it documented.

I think the ASR itself should have this figured out (either all 32 bit, or all 64 bit, and explicit casts in ASR itself).

assem2002:
I don't get what actually to get documented, is that we don't support array size of int64 in llvm `array_descriptor`?
If it's about ASR node `ArraySize` , I think we can't cast this explicitly in the ASR the node looks like this :
```clj
(ArraySize
  (Var 3 fxhv)
  ()
  (Integer 2)
  ()
)
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3724,6 +3724,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 ptr_loads = 2;
                 for( size_t i = 0; i < n_dims; i++ ) {
                     this->visit_expr_wrapper(m_dims[i].m_length, true);
+
+                    // Make dimension length and return size compatible.(TODO : array_size should be of type int64)
+                    if(ASRUtils::extract_kind_from_ttype_t(
+                        ASRUtils::expr_type(m_dims[i].m_length)) > 32){
```
certik, line 3730:
```suggestion
                        ASRUtils::expr_type(m_dims[i].m_length)) > 4){
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3724,6 +3724,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        tmp = builder->CreateTrunc(tmp, llvm::IntegerType::get(context, 32));
+                    } else if (ASRUtils::extract_kind_from_ttype_t(
+                        ASRUtils::expr_type(m_dims[i].m_length)) < 32){
```
certik, line 3733:
```suggestion
                        ASRUtils::expr_type(m_dims[i].m_length)) < 4){
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3724,6 +3724,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 ptr_loads = 2;
                 for( size_t i = 0; i < n_dims; i++ ) {
                     this->visit_expr_wrapper(m_dims[i].m_length, true);
+
+                    // Make dimension length and return size compatible.(TODO : array_size should be of type int64)
```
certik, line 3728: Create an issue an link it here

assem2002: opened one https://github.com/lfortran/lfortran/issues/5679

src/libasr/codegen/asr_to_llvm.cpp
assem2002: done here https://github.com/lfortran/lfortran/pull/5648/commits/d19a917f94b9aa4ea76638914ab51aad2181cbaa

src/libasr/codegen/asr_to_llvm.cpp
assem2002: done here https://github.com/lfortran/lfortran/pull/5648/commits/d19a917f94b9aa4ea76638914ab51aad2181cbaa

src/libasr/codegen/asr_to_llvm.cpp
certik: You should link it from the source code too.

certik (approved):

PR #5654, opened 2024-12-11
Fixes for `integration_tests/matmul_01.f90`  in C backend

czgdp1807:


czgdp1807 (commented):
Another PR where the bug was not in the `simplifier` pass. It was in the C backend but in older LFortran it never showed up. Enabling auto-merge here as it is harmless and in fact improvises the C backend.

src/libasr/asr_utils.h
```diff
@@ -6385,6 +6385,9 @@ class RemoveArrayProcessingNodeReplacer: public ASR::BaseExprReplacer<RemoveArra
     }
 
     void replace_ArrayPhysicalCast(ASR::ArrayPhysicalCast_t* x) {
+        if( x->m_new == ASR::array_physical_typeType::SIMDArray ) {
+            return ;
+        }
         ASR::BaseExprReplacer<RemoveArrayProcessingNodeReplacer>::replace_ArrayPhysicalCast(x);
         if( !ASRUtils::is_array(ASRUtils::expr_type(x->m_arg)) ) {
             *current_expr = x->m_arg;
```
certik, line 6393: What is the `RemoveArrayProcessingNodeReplacer` doing and why do we need to skip the `simd` array physical type casts?

PR #5658, opened 2024-12-11
fix: extract value from first argument in findloc and use it

gxyd:
## Description

The below Fortran program doesn't work in *main*, this PR fixes it:
```f90
program main
    implicit none
    integer, parameter :: x1(5) = [1, 3, 2, 4, 2]
    integer, parameter :: y1 = 2
    integer, parameter :: i14(1) = findloc(x1, y1)
end program main

```

### DIFF
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -3304,11 +3304,12 @@ namespace FindLoc {
         ASR::expr_t* dim = args[2];
         if (!array) return nullptr;
         if (!value) return nullptr;
-        if (extract_n_dims_from_ttype(expr_type(args[0])) == 1) {
+        if (extract_n_dims_from_ttype(expr_type(array)) == 1) {
             int arr_size = 0;
+            ASR::expr_t* array_value = ASRUtils::expr_value(array);
             ASR::ArrayConstant_t *arr = nullptr;
-            if (ASR::is_a<ASR::ArrayConstant_t>(*array)) {
-                arr = ASR::down_cast<ASR::ArrayConstant_t>(array);
+            if (array_value && ASR::is_a<ASR::ArrayConstant_t>(*array_value)) {
+                arr = ASR::down_cast<ASR::ArrayConstant_t>(array_value);
                 arr_size = ASRUtils::get_fixed_size_of_array(arr->m_type);
             } else {
                 return nullptr;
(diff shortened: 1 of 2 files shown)


src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3345,11 +3346,11 @@ namespace FindLoc {
...
                 double ele = 0;
                 if (is_integer(*ASRUtils::expr_type(args[1]))) {
-                    ele = ASR::down_cast<ASR::IntegerConstant_t>(args[1])->m_n;
+                    ele = ASR::down_cast<ASR::IntegerConstant_t>(ASRUtils::expr_value(args[1]))->m_n;
                 } else if (is_real(*ASRUtils::expr_type(args[1]))) {
-                    ele = ASR::down_cast<ASR::RealConstant_t>(args[1])->m_r;
+                    ele = ASR::down_cast<ASR::RealConstant_t>(ASRUtils::expr_value(args[1]))->m_r;
                 } else if (is_logical(*ASRUtils::expr_type(args[1]))) {
-                    ele = ASR::down_cast<ASR::LogicalConstant_t>(args[1])->m_value;
+                    ele = ASR::down_cast<ASR::LogicalConstant_t>(ASRUtils::expr_value(args[1]))->m_value;
```
certik, line 3353: Is there always `expr_value` to be guaranteed to be present (i.e., not `nullptr`)?

gxyd: Actually no, I was probably being a little lazy here, I'll write some good tests and improve the logic here.

gxyd:
I'll move this PR to *draft mode*, as I think there are a few more bugs and I'll try to fix them.

HarshitaKalani:
```
$ gfortran a.f90 && ./a.out
           3
$ lfortran a.f90
0
```
The example above indeed gives wrong answer with the main. This PR I think solves the issue. Can we complete this?

src/libasr/pass/intrinsic_array_function_registry.h
gxyd: I think this is now resolved in the new changes, I've made sure to exit early when `expr_value` return `nullptr` for `value` argument, hence, it's handled now.

gxyd:
Apologies for the delay here, I think this should is now available for a review.

gxyd:
There are a few more issues with `findloc`, I'll open issues for them.

Issue #5642, opened 2024-12-10
Missing unrestricted intrinsic function


pjh40:
Issue #492 shows `IDNINT` as implemented, but
```f90
program test
  print *, IDNINT(1.1d0)
end program test
```
yields
```
semantic error: Function 'idnint' not found (not user defined nor intrinsic)
 --> intrinsics.f90:2:12
  |
2 |   print *, IDNINT(1.1d0)
  |            ^^^^^^^^^^^^^ 
```

certik:
This is now fixed by #5659, thanks @parth121101 !
```console
$ lfortran a.f90 
1
```

PR #5659, opened 2024-12-11
feat: implement idnint intrinsic

parth121101:
Towards: #5642


src/libasr/pass/intrinsic_functions.h
```diff
@@ -2506,6 +2507,48 @@ namespace Nint {
...
 } // namespace Nint
 
+namespace Idnint {
+
+    static ASR::expr_t *eval_Idnint(Allocator &al, const Location &loc,
+            ASR::ttype_t* arg_type, Vec<ASR::expr_t*> &args, diag::Diagnostics& diag) {
+        if (ASRUtils::extract_kind_from_ttype_t(expr_type(args[0])) != 8 ) {
+            diag.semantic_error_label("`idnint` takes argument of kind 8", {loc}, "");
+        }
+        double rv = ASR::down_cast<ASR::RealConstant_t>(args[0])->m_r;
```
gxyd, line 2517: this probably needs use of `ASRUtils::expr_value` to get this working for parameterised constants like `real, parameter :: x = 2`

integration_tests/intrinsics_329.f90
```diff
@@ -0,0 +1,82 @@
+program intrinsics_329
+    use iso_fortran_env, only: dp => real64
+    real(dp) :: w = -5.1d0
```
gxyd, line 3: It would be great to have parameterised tests, to ensure that values (wherever expected) are evaluated at compile time.

certik: We need to test both runtime variables, as well as compile time values.

For compile time values, I thought just `idnint(5.8_dp)` would be enough, but @gxyd you discovered a case where using `x` as a parameter will break? In that case, let's add a test with `real, parameter :: x` as well in here and ensure it works.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -2506,6 +2507,48 @@ namespace Nint {
...
 } // namespace Nint
 
+namespace Idnint {
+
+    static ASR::expr_t *eval_Idnint(Allocator &al, const Location &loc,
+            ASR::ttype_t* arg_type, Vec<ASR::expr_t*> &args, diag::Diagnostics& diag) {
+        if (ASRUtils::extract_kind_from_ttype_t(expr_type(args[0])) != 8 ) {
+            diag.semantic_error_label("`idnint` takes argument of kind 8", {loc}, "");
+        }
+        double rv = ASR::down_cast<ASR::RealConstant_t>(ASRUtils::expr_value(args[0]))->m_r;
```
gxyd, line 2517: Sorry to bug you, but I think this will now fail for non-parameterised values, probably return as soon as `ASRUtils::expr_value(arg)` is `nullptr`

parth121101: its not breaking @gxyd

certik: Isn't the `eval_Idnint` only called if each argument has a value? If so, then `expr_value` should not fail.

certik: Also, isn't it called with the values already substituted for the arguments?

gxyd: Ah, I see

parth121101: right

parth121101: sure

Pranavchiku (approved):
Thank you! Looks good to me.

src/libasr/pass/intrinsic_functions.h
certik: The conclusion seems to be that it does not need `ASRUtils::expr_value` and it will work for parameter constants too.

certik (approved):
Thanks!

Issue #5639, opened 2024-12-10
semantic error on nullifying a pointer


gxyd:
## Minimal Reproducible Example

```f90
module p
    implicit none
    integer, parameter, private :: mxdim = 3
    type :: rp1d
        real(8), dimension(:), pointer :: f
    end type

    type :: sds
        integer :: ndim
        integer, dimension(mxdim) :: dims
        logical :: hdf32
        type(rp1d), dimension(mxdim) :: scales
        real(8), dimension(:,:,:), pointer :: f
    end type
end module

program main
    use p
    implicit none
    type(sds) :: s
    nullify(s%scales(1)%f)
end program

```

The above program on it's compilation with LFortran raises the below error:
```console
> lfortran test.f90
semantic error: Only a pointer variable symbol can be nullified.
  --> test.f90:21:13
   |
21 |     nullify(s%scales(1)%f)
   |             ^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

while the above program works without any issue with GFortran.

## Updated MRE

Further simplified Fortran program, which produces the same error:

```f90
module p
    implicit none

    type :: scaled_data_structure
        real, dimension(:), pointer :: one_d_real_pointer
    end type
end module

program main
    use p, only : scaled_data_structure
    implicit none
    type(scaled_data_structure) :: s
    nullify(s%one_d_real_pointer)
end program

```

adit4443ya:
I applied just small diff 
```diff
 aditya-trivedi   lfortran    main ↓12  ?2 ~1    git diff
diff --git a/src/lfortran/semantics/ast_body_visitor.cpp b/src/lfortran/semantics/ast_body_visitor.cpp
index 45ec21387..337d821d9 100644
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -4407,14 +4407,20 @@ public:
         for( size_t i = 0; i < x.n_args; i++ ) {
             this->visit_expr(*(x.m_args[i]));
             ASR::expr_t* tmp_expr = ASRUtils::EXPR(tmp);
+            std::cout<<tmp_expr->type<<std::endl;
             if( tmp_expr->type != ASR::exprType::Var ) {
-                diag.add(Diagnostic(
-                    "Only a pointer variable symbol "
-                    "can be nullified.",
[...]

Issue #2862, opened 2023-11-17
Compiling POT3D


certik:
Here is a good code to test LFortran on:

https://github.com/predsci/POT3D/blob/53b0550902a512c357de736ef10b9196089ac53e/src/stdpar/pot3d.F

Pranavchiku:
This is weird, it shows me:

```console
% lfortran a.f --fixed-form
tokenizer error: Expecting terminating symbol for program
  --> a.f:44:7
   |
44 |       module ident
   |       ^ 


Note: Please report unclear or confusing messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

certik:
Our fixed-form tokenizer must be improved. I can look at it later, it should not be difficult.

sumseq:
Any updates of the compilation of POT3D with LFortran?

certik:
@sumseq I haven't worked on POT3D specifically, but we can now compile the SNAP code (with just some small workarounds) and also can parallelize do concurrent (via LLVM OpenMP runtime), so I am sure we are much closer to compile it. We are looking for beta testers and users. If you think you would use LFortran regularly if we can compile it, we can prioritize it. Here is our current code progression: https://github.com/lfortran/lfortran/issues/3806.

sumseq:
We probably would only use LFortran if it supported DC GPU offload to AMD GPUs, or ran the code significantly faster than GCC/IFX.
That said, POT3D is part of the SPEChpc2021 benchmark (spec.org/hpc2021) and I think having LFortran successfully compile the complete SPEChpc2021 benchmark would be a really great milestone/achievement for the compiler (since Fortran is a heavily used for HPC) and convince people to give it a try.

certik:
@sumseq offloading to AMD (#4535) and NVIDIA (#4497) GPUs is planned, but not implemented yet. Regarding speed, we are currently usually within a factor of 2x slower than GCC. I have not used ifx yet, but my experience with ifort was that it is usually faster than GCC, so I am guessing ifx will eventually match ifort's speed. Right now our main focus is on compiling codes, not necessarily runtime performance, but once we reach beta quality for code compilation, we'll focus on performance.

Regarding AMD GPU support, what is the best way for a compiler to offload it? If you have any feedback, let us know in #4535.

sumseq:
What are you thoughts on supporting the full SPEChpc2021 benchmark?
At least on the CPU?

certik:
[...]

Issue #5655, opened 2024-12-11
Allocation of struct pointer


adit4443ya:
MRE
```fortran
module p
    implicit none
    integer, parameter, private :: mxdim = 3
    type :: rp1d
        real(8), dimension(:), pointer :: f
    end type

    type :: sds
        integer :: ndim
        integer, dimension(mxdim) :: dims
        logical :: hdf32
        type(rp1d), dimension(mxdim) :: scales
        real(8), dimension(:,:,:), pointer :: f
    end type
end module

program main
    use p
    implicit none
    type(sds) :: s
    allocate(s%scales(1)%f(2))
    s%scales(1)%f=[1,2]
    print *, 'Scale 0:', s%scales(1)%f
    nullify(s%scales(1)%f)
end program
```
on Lfortran compilation
```shell
 aditya-trivedi   lfortran    main ≡  ?1 ~1    lfortran ./integration_tests/nullify_03.f90
lfortran: /home/aditya-trivedi/conda_root/envs/lf/include/llvm/IR/Instructions.h:939: static llvm::GetElementPtrInst* llvm::GetElementPtrInst::Create(llvm::Type*, llvm::Value*, llvm::ArrayRef<llvm::Value*>, const llvm::Twine&, llvm::Instruction*): Assertion `PointeeType == cast<PointerType>(Ptr->getType()->getScalarType())->getElementType()' failed.
Traceback (most recent call last):
  Binary file "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran.cpp", line 2732, in main
    return main_app(argc, argv);
  File "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran.cpp", line 2670, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
[...]

PR #5660, opened 2024-12-11
Fix: Handles StructInstanceMember Pointer to be nullified

adit4443ya:
Towards #5639 and hence #2862

I guess the expected ASR is correct But it's giving llvm error at compilation

Please Refer [comment](https://github.com/lfortran/lfortran/issues/5639#issuecomment-2534190101)


tests/tests.toml
```diff
@@ -2883,6 +2883,12 @@ ast = true
 asr = true
 llvm = true
 
+[[test]]
+filename = "../integration_tests/nullify_03.f90"
+ast = true
```
certik, line 2888: Don't test AST here, unless there is a parsing issue, but I think this parses just fine.

adit4443ya: ok

tests/tests.toml
```diff
@@ -2883,6 +2883,12 @@ ast = true
...
+asr = true
+llvm = true
```
certik, line 2890: Either it compiles and works, then just make it a regular integration test.

Or it doesn't not yet, in which case just test ASR here.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4408,13 +4408,17 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                    "can be nullified.",
-                    Level::Error, Stage::Semantic, {
-                        Label("",{tmp_expr->base.loc})
-                    }));
-                throw SemanticAbort();
+                if (tmp_expr->type == ASR::exprType::StructInstanceMember && ASR::is_a<ASR::Pointer_t>(*ASR::down_cast<ASR::StructInstanceMember_t>(tmp_expr)->m_type)) {
+                    arg_vec.push_back(al, ASR::down_cast<ASR::StructInstanceMember_t>(tmp_expr)->m_m);
+                } else {
+                    diag.add(Diagnostic(
+                        "Only a pointer variable symbol1 "
```
certik, line 4415:
```suggestion
                        "Only a pointer variable symbol "
```

tests/tests.toml
certik: Then just test "asr" here, nothing else.

integration_tests/nullify_03.f90
```diff
@@ -0,0 +1,25 @@
...
+        type(rp1d), dimension(mxdim) :: scales
+        real(8), dimension(:,:,:), pointer :: f
+    end type
+end module
+
+program nullify_03
+    use p
+    implicit none
+    type(sds) :: s
+    ! allocate(s%scales(1)%f(2))
```
adit4443ya, line 21: This is separate Issue for allocating the pointer of struct 
i opened it here 
#5655

gxyd (commented):
I tried to get the ASR of the below Fortran program:
```f90
module p
    implicit none

    type :: scaled_data_structure
        real, dimension(:), pointer :: one_d_real_pointer
    end type
end module

program main
    use p, only : scaled_data_structure
    implicit none
    type(scaled_data_structure) :: s
    type(scaled_data_structure) :: s2
    nullify(s%one_d_real_pointer)
    nullify(s2%one_d_real_pointer)
end program

```

the output is:
```clj
(TranslationUnit
    (SymbolTable
        1
        {
            main:
                (Program
                    (SymbolTable
                        4
                        {
                            1_scaled_data_structure_one_d_real_pointer:
                                (ExternalSymbol
                                    4
                                    1_scaled_data_structure_one_d_real_pointer
                                    3 one_d_real_pointer
                                    scaled_data_structure
                                    []
                                    one_d_real_pointer
                                    Public
                                ),
                            s:
                                (Variable
                                    4
                                    s
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (StructType
                                        4 scaled_data_structure
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            s2:
                                (Variable
                                    4
                                    s2
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (StructType
                                        4 scaled_data_structure
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            scaled_data_structure:
                                (ExternalSymbol
                                    4
                                    scaled_data_structure
                                    2 scaled_data_structure
                                    p
                                    []
                                    scaled_data_structure
                                    Public
                                )
                        })
                    main
                    [p]
                    [(Nullify
                        [4 1_scaled_data_structure_one_d_real_pointer]
                    )
                    (Nullify
                        [4 1_scaled_data_structure_one_d_real_pointer]
                    )]
                ),
            p:
                (Module
                    (SymbolTable
                        2
                        {
                            scaled_data_structure:
                                (Struct
                                    (SymbolTable
                                        3
                                        {
                                            one_d_real_pointer:
                                                (Variable
                                                    3
                                                    one_d_real_pointer
                                                    []
                                                    Local
                                                    ()
                                                    ()
                                                    Default
                                                    (Pointer
                                                        (Array
                                                            (Real 4)
                                                            [(()
                                                            ())]
                                                            DescriptorArray
                                                        )
                                                    )
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                )
                                        })
                                    scaled_data_structure
                                    []
                                    [one_d_real_pointer]
                                    Source
                                    Public
                                    .false.
                                    .false.
                                    []
                                    ()
                                    ()
                                )
                        })
                    p
                    []
                    .false.
                    .false.
                )
        })
    []
)
```

You can see that there are two `nullify` nodes:
```clj
                    (Nullify
                        [4 1_scaled_data_structure_one_d_real_pointer]
                    )
                    (Nullify
                        [4 1_scaled_data_structure_one_d_real_pointer]
                    )
```

even though they apply on different instances of `scaled_data_structure`, but produce the same ASR, which I think is incorrect. Something is wrong there.

adit4443ya:
> You can see that there are two `nullify` nodes:
> 
> ```clojure
>                     (Nullify
>                         [4 1_scaled_data_structure_one_d_real_pointer]
>                     )
>                     (Nullify
>                         [4 1_scaled_data_structure_one_d_real_pointer]
>                     )
> ```
> even though they apply on different instances of `scaled_data_structure`, but produce the same ASR, which I think is incorrect. Something is wrong there.

I see that good catch,, 
I guess this is because of the fact that there is only one symbol for both the StructInstanceMember 
There must be something like 
1_s_scaled_data_structure_one_d_real_pointer and 
1_s1_scaled_data_structure_one_d_real_pointer

in the ASR so that we can differentiate between the symbols of two different variables declared in the Program

Then it would be 

> ```clojure
>                     (Nullify
>                         [4 1_s_scaled_data_structure_one_d_real_pointer]
>                     )
>                     (Nullify
>                         [4 1_s1_scaled_data_structure_one_d_real_pointer]
>                     )
> ```

Is this correct @gxyd

integration_tests/nullify_03.f90
```diff
@@ -0,0 +1,25 @@
+module p
```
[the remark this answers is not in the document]
adit4443ya: Are you suggesting here to use this MRE for this PR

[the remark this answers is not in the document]
gxyd: Yes, I think we can, because I believe the fundamental problem is same in both.

gxyd:
> Then it would be
> 
> > ```clojure
> >                     (Nullify
> >                         [4 1_s_scaled_data_structure_one_d_real_pointer]
> >                     )
> >                     (Nullify
> >                         [4 1_s1_scaled_data_structure_one_d_real_pointer]
> >                     )
> > ```
> 
> Is this correct @gxyd

Something like that yes.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4408,13 +4408,17 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             this->visit_expr(*(x.m_args[i]));
             ASR::expr_t* tmp_expr = ASRUtils::EXPR(tmp);
             if( tmp_expr->type != ASR::exprType::Var ) {
```
certik, line 4410: First of all, we should be using `ASR::is_a<ASR::Var_t>(*tmp_expr)` here.

certik:
So the rule is that nullify() can only have variables or structure members as arguments, not (other) expressions (so no function results, or other arithmetic expressions). The argument can be any pointer variable (including a `procedure(), pointer`).

So the logic should be:

* Let's use `expr_type` and check that each argument is a `Pointer` (I don't know if we can just use `is_a<Pointer_t>` on the type, or if we need to call an auxiliary function that will determine if the type is a Pointer, as the Pointer node might not be an upper level node, say with Array or other more complex types). If it is not a pointer, then we give a nice error message.
* After that, let's check that the argument is either `Var_t`, or struct component; Otherwise give a nice error message.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4408,13 +4408,17 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                diag.add(Diagnostic(
-                    "Only a pointer variable symbol "
-                    "can be nullified.",
-                    Level::Error, Stage::Semantic, {
-                        Label("",{tmp_expr->base.loc})
-                    }));
-                throw SemanticAbort();
+                if (tmp_expr->type == ASR::exprType::StructInstanceMember && ASR::is_a<ASR::Pointer_t>(*ASR::down_cast<ASR::StructInstanceMember_t>(tmp_expr)->m_type)) {
+                    arg_vec.push_back(al, ASR::down_cast<ASR::StructInstanceMember_t>(tmp_expr)->m_m);
```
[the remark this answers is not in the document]
certik: So that instead of:
```
                    (Nullify
                        [4 1_s_scaled_data_structure_one_d_real_pointer]
                    )
                    (Nullify
                        [4 1_s1_scaled_data_structure_one_d_real_pointer]
                    )
```
we get (roughly):
```
                    (Nullify
                        [4 (StructInstanceMember
                            (Var 4 s1)
                            4 1_s_a
                            (Integer 4)
                            ()
                        )]
                    )
                    (Nullify
                        [4 (StructInstanceMember
                            (Var 4 s1)
                            4 1_s_a
                            (Integer 4)
                            ()
                        )]
                    )
```
The point is that we need to have nullify of `StructInstanceMember` here.

adit4443ya:
> * Let's use `expr_type` and check that each argument is a `Pointer` (I don't know if we can just use `is_a<Pointer_t>` on the type, or if we need to call an auxiliary function that will determine if the type is a Pointer, as the Pointer node might not be an upper level node, say with Array or other more complex types). If it is not a pointer, then we give a nice error message.

I agree on this point but expr_type don't have `Pointer` in enum ,,, as you said it's not upper level node 
In this case upper level node is the `StructInstanceMember` so we need to get from this that if it's a pointer or not 
same with `Var_t` ....

src/lfortran/semantics/ast_body_visitor.cpp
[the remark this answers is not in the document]
adit4443ya: For that we need to change the Nullify_t something like this 
```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 36deafc6c..323a67ae8 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -60,7 +60,7 @@ stmt
     | IntrinsicImpureSubroutine(int sub_intrinsic_id, expr* args, int overload_id)
     | Where(expr test, stmt* body, stmt* orelse)
     | WhileLoop(identifier? name, expr test, stmt* body, stmt* orelse)
-    | Nullify(symbol* vars)
+    | Nullify(expr* vars)
     | Flush(int label, expr unit, expr? err, expr? iomsg, expr? iostat)
     | ListAppend(expr a, expr ele)
     | AssociateBlockCall(symbol m)
(END)
```
Then we can have structInstancemember itself in the list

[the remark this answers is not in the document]
certik: I see. Yes, it needs to be an `expr`, not a `symbol`.

certik:
If it is not upper level in `ttype_t`, then write a function that determines if it is a Pointer or not.

adit4443ya:
Goog thing is now ASR is correctly generated As what we discussed yesterday
```clojure
 aditya-trivedi   lfortran    null ≢    lfortran ./integration_tests/nullify_04.f90 --show-asr
(TranslationUnit
    (SymbolTable
        1
        {
            nullify_04:
                (Program
                    (SymbolTable
                        4
                        {
                            1_scaled_data_structure_one_d_real_pointer:
                                (ExternalSymbol
                                    4
                                    1_scaled_data_structure_one_d_real_pointer
                                    3 one_d_real_pointer
                                    scaled_data_structure
                                    []
                                    one_d_real_pointer
                                    Public
                                ),
                            s:
                                (Variable
                                    4
                                    s
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (StructType
                                        4 scaled_data_structure
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            s2:
                                (Variable
                                    4
                                    s2
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (StructType
                                        4 scaled_data_structure
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            scaled_data_structure:
                                (ExternalSymbol
                                    4
                                    scaled_data_structure
                                    2 scaled_data_structure
                                    p
                                    []
                                    scaled_data_structure
                                    Public
                                )
                        })
                    nullify_04
                    [p]
                    [(Nullify
                        [(StructInstanceMember
                            (Var 4 s)
                            4 1_scaled_data_structure_one_d_real_pointer
                            (Pointer
                                (Array
                                    (Real 4)
                                    [(()
                                    ())]
                                    DescriptorArray
                                )
                            )
                            ()
                        )]
                    )
                    (Nullify
                        [(StructInstanceMember
                            (Var 4 s2)
                            4 1_scaled_data_structure_one_d_real_pointer
                            (Pointer
                                (Array
                                    (Real 4)
                                    [(()
                                    ())]
                                    DescriptorArray
                                )
                            )
                            ()
                        )]
                    )]
                ),
                
```
For this case 
```fortran
module p
    implicit none

    type :: scaled_data_structure
        real, dimension(:), pointer :: one_d_real_pointer
    end type
end module

program nullify_04
    use p, only : scaled_data_structure
    implicit none
    type(scaled_data_structure) :: s
    type(scaled_data_structure) :: s2
    nullify(s%one_d_real_pointer)
    nullify(s2%one_d_real_pointer)
end program
```

gxyd:
The CI fails, it will be helpful if we can fix that first, then it will be a little easier to review I think.

integration_tests/nullify_03.f90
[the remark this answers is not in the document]
gxyd: The current CI failure of job `Check Release Build` is because of the module name `p`:
```console
FAIL: Found a duplicate module name
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 110, in <module>
Name: p
Filename: nullify_03.f90
    main()
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 93, in main
    check_module_names()
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 71, in check_module_names
    raise Exception("Duplicate module names")
Exception: Duplicate module names
```

probably use `module_nullify_03` (or something like that)?

integration_tests/nullify_04.f90
```diff
@@ -0,0 +1,16 @@
+module p
```
gxyd, line 1: similarly, the module name here can be `module_nullify_04`

gxyd (approved):
Looks good to me. Any reason to still keep this in *draft mode*?

adit4443ya:
> Looks good to me. Any reason to still keep this in _draft mode_?

Ready!!
The LLVM error is coming for only StructInstancemember 
But ASR is generating well
I will make another PR to fix this LLVM error
If it's good then , it's ready to merge

certik (approved):
I think this looks good!

PR #5665, opened 2024-12-12
enh:Fortran Backend: Nullify Visitor

adit4443ya:
This PR adds a Nullify visitor in the fortran backend
```fortran
 aditya-trivedi   lfortran    fortran ≢    cat ./integration_tests/nullify_02.f90 
program nullify_02
implicit none

   real, pointer :: p1
   real, target :: t1 
   integer, pointer :: p2
   integer, target :: t2
   
   p1=>t1
   p1 = 1
   p2=>t2
   p2 = 2.
   nullify(p1, p2) 
   
   
end program nullify_02
 aditya-trivedi   lfortran    fortran ≢    lfortran ./integration_tests/nullify_02.f90 --show-fortran --show-stacktrace
program nullify_02
implicit none
real(4), pointer :: p1
integer(4), pointer :: p2
real(4) :: t1
integer(4) :: t2
p1 => t1
p1 = real(1, kind=4)
p2 => t2
p2 = int(2.00000000e+00, kind=4)
nullify (p1, p2)
end program nullify_02
 aditya-trivedi   lfortran    fortran ≢   
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1917,6 +1917,9 @@ RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_40 LABELS llvm_omp llvm)
 RUN(NAME openmp_41 LABELS llvm_omp llvm)
 
+RUN(NAME nullify_01 LABELS gfortran fortran llvm)
+RUN(NAME nullify_02 LABELS gfortran fortran llvm)
+
 RUN(NAME fortran_01 LABELS gfortran llvm fortran)
 RUN(NAME fortran_02 LABELS gfortran fortran)
 
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -1275,11 +1275,10 @@ public:
         std::string r = indent;
         r += "nullify (";
         for (int i = 0; i < static_cast<int>(x.n_vars); i++) {
-            if(x.m_vars[i]->type == ASR::Variable) {
-                r += ASRUtils::symbol_name(x.m_vars[i]);
-                if(i != static_cast<int>(x.n_vars-1)) {
-                    r += ", ";
-                }
+            visit_expr(*x.m_vars[i]);
+            r += src;
+            if(i != static_cast<int>(x.n_vars-1)) {
+                r += ", ";
             }
         }
         r += ")";
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -721,6 +721,9 @@ public:
         if (x.m_value_attr) {
             r += ", value";
         }
+        if (x.m_target_attr) {
+            r += ", target";
+        }
         r += " :: ";
         r.append(x.m_name);
         if (x.m_symbolic_value && x.m_value && ASR::is_a<ASR::StringChr_t>(*x.m_symbolic_value) && ASR::is_a<ASR::StringConstant_t>(*x.m_value)) {
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -1271,7 +1271,22 @@ public:
         src = r;
     }
 
-    // void visit_Nullify(const ASR::Nullify_t &x) {}
+    void visit_Nullify(const ASR::Nullify_t &x) {
+        std::string r = indent;
+        r += "nullify (";
+        for (int i = 0; i < static_cast<int>(x.n_vars); i++) {
+            if(x.m_vars[i]->type == ASR::Variable) {
+                r += ASRUtils::symbol_name(x.m_vars[i]);
+                if(i != static_cast<int>(x.n_vars-1)) {
+                    r += ", ";
+                }
+            }
+        }
+        r += ")";
+        handle_line_truncation(r, 2);
+        r += "\n";
+        src = r;
+    }
 
     // void visit_Flush(const ASR::Flush_t &x) {}
 


adit4443ya:
Error in CI indicates 
```clojure
6 | p1 => t1
      |      1
Error: Pointer assignment target is neither TARGET nor POINTER at (1)
CMakeFiles/nullify_01.dir/nullify_01.f90.o.tmp.f90:7:6:

    7 | p2 => t1
      |      1
Error: Pointer assignment target is neither TARGET nor POINTER at (1)
```
It's due to the fact the in fortran backend 
`integer, target :: t2` is transformed into 
`integer(4) :: t2`

gxyd:
That's a tricky one. I've always wanted that problem to be fixed though, since a lot of time output program(s) `--dump-all-passes-fortran` can't be run directly without having `target` for some of the variables.

Something to keep in mind:
```f90
program nullify_01
    implicit none
    integer, pointer :: p(:)       ! Pointer to an integer array
    integer, pointer :: q(:)
    allocate(p(5))
    p = [1, 2, 3, 4, 5]
    q => p
end program

```

In the above program, even though `p` appears on the RHS, it can't have an *explicit* target attribute.

Pranavchiku:
Woah, LFortran treats `integer, target :: x` same as `integer :: x`. We need to distinguish it somehow. 


```console
        (Declaration
            (AttrType
                TypeInteger
                []
                ()
                ()
                None
            )
            [(SimpleAttribute
                AttrTarget
            )]
            [(t2
            []
            []
            ()
            ()
            None
            ())]
            ()
        )
        (Declaration
            (AttrType
                TypeInteger
                []
                ()
                ()
                None
            )
            []
            [(t3
            []
            []
            ()
            ()
            None
            ())]
            (TriviaNode
                []
                [(EndOfLine)
                (EndOfLine)]
            )
        )]
```

We can use 

```console
(SimpleAttribute
                AttrTarget
            )
```
to set `target_attr`

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 36deafc6c..e8f8d0d09 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -18,7 +18,7 @@ symbol
     | Struct(symbol_table symtab, identifier name, identifier* dependencies, identifier* members, abi abi, access access, bool is_packed, bool is_abstract, call_arg* initializers, expr? alignment, symbol? parent)
     | Enum(symbol_table symtab, identifier name, identifier* dependencies, identifier* members, abi abi, access access, enumtype enum_value_type, ttype type, symbol? parent)
     | Union(symbol_table symtab, identifier name, identifier* dependencies, identifier* members, abi abi, access access, call_arg* initializers, symbol? parent)
-    | Variable(symbol_table parent_symtab, identifier name, identifier* dependencies, intent intent, expr? symbolic_value, expr? value, storage_type storage, ttype type, symbol? type_declaration, abi abi, access access, presence presence, bool value_attr)
+    | Variable(symbol_table parent_symtab, identifier name, identifier* dependencies, intent intent, expr? symbolic_value, expr? value, storage_type storage, ttype type, symbol? type_declaration, abi abi, access access, presence presence, bool value_attr, bool target_attr)
     | Class(symbol_table symtab, identifier name, abi abi, access access)
     | ClassProcedure(symbol_table parent_symtab, identifier name, identifier? self_argument, identifier proc_name, symbol proc, abi abi, bool is_deferred, bool is_nopass)
     | AssociateBlock(symbol_table symtab, identifier name, stmt* body)
```

@certik shall we add this or is there any other way to propagate?

gxyd:
If I think about it, we can have the `target` attribute with every possible declarations when generating Fortran code (ASR -> Fortran), except ofcourse when it isn't legit like in case of `pointer` attribute present for them (or other few cases).

Consider the below C++ code:
```cpp
#include <iostream>

int main() {
    int* p1 = nullptr; 
    int* p2 = nullptr;
    int t1;
    int t2;

    p1 = &t1; 
    p2 = &t1;

    *p1 = 1;

    p1 = nullptr;
    p2 = nullptr;


    if (p1 == nullptr && p2 == nullptr) {
        std::cout << "p1 and p2 are nullified." << std::endl;
    }

    return 0;
}
```

Even though `t1` and `t2` don't have any difference in there declaration, we can take `&t1` or `&t2` whenever we feel like, so by default having `target` attribute in ASR -> Fortran whenever we can, is one solution.

Pranavchiku:
```console
program nullify_02
implicit none

   real, pointer :: p1
   real, target :: t1 
   integer, pointer :: p2
   integer, target :: t2
   integer :: t3
   
   p1=>t1
   p1 = 1
   p2=>t3
   p2 = 2.
   nullify(p1, p2) 
   
   
end program nullify_02
```

> we can have the target attribute with every possible declarations

no then `p2=>t3` will be valid but it should not be. Also, it will be divergence from the motivation of Fortran backend i.e. "as close as user code"

adit4443ya:
I guess we need target attribute to distinguish which variable should be pointed by a pointer...
If not used then any variable would be targetted which is incorrect. 
In Cpp we have liberty to have pointer of nearly everything that's why we do not need any specific attribute to declare it to be targetted,,

gxyd:
I read about `target`, I think there can be many cases where having `target` wouldn't be good, so I think what you recommended `target` property as a part of `Variable` node makes more sense, with default as `true` maybe.

> no then p2=>t3 will be valid but it should not be. Also, it will be divergence from the motivation of Fortran backend i.e. "as close as user code"

Yes, I think you are right.

adit4443ya:
I tried to set the default value of `target_attr` to `true`, but `Variable_t` has many dependencies, hence we need a track of where `target_attr` should be false. For instance, we cannot use `real(4), pointer, target :: p1`; thus, we must ensure `target_attr` is false whenever we encounter a pointer.

One solution is to implement a single visitor function just before ASR generation to set `target_attr` to false whenever `ASR::is_a<Pointer_t>(*var->m_type)` is true.

Tell me how that sounds @Pranavchiku @gxyd

certik:
I would worry about correct printing of "target" later if it is difficult now.

We might need to track it in ASR itself and enforce using ASR verify.

Pranavchiku:
>  but Variable_t has many dependencies

The way I do it is I create a `ASRUtils::make_Variable_t_util` that takes in all arguments and `target_attr = false` by default. Then search and replace `make_Variable_t`, this way nothing breaks. Then add `target_attr` at required places.

adit4443ya:
That sounds good 
I will make another PR for that 
Here it is #5698 Linked with Issue #5697

adit4443ya:
I think this can be merged now

certik (approved):

PR #5680, opened 2024-12-13
Fix: Use Var name in SubCall instead type_decl name

### LINKED ISSUE #5603: (PRIMA) use actual function instead of its type in `m_type_declaration` of Variable

jinangshah21:
Fixes #5603 
Refer https://github.com/lfortran/lfortran/issues/5603#issuecomment-2525200055.

### DIFF
--- a/integration_tests/procedure_08.f90
+++ b/integration_tests/procedure_08.f90
@@ -2,16 +2,17 @@ module procedure_08_module
 contains
     subroutine cb(x)
         implicit none
-        integer, intent(in), optional :: x(:)
+        integer, intent(inout), optional :: x(:)
     end subroutine cb
 
     subroutine calfun(x)
         implicit none
-        integer, intent(in), optional :: x(:)
+        integer, intent(inout), optional :: x(:)
         logical :: y
+        x = 2
         y = present(x)
         print *, y
-        if (y) error stop
+        if (.not. y) error stop
     end subroutine calfun
 end module procedure_08_module
 
@@ -25,7 +26,12 @@ contains
     subroutine temp(call_back)
         implicit none
         procedure(cb), optional :: call_back
-        if(present(call_back)) call call_back()   
+        integer :: x(4) 
+        if(present(call_back)) then
+            call call_back(x) 
+            print *, x
+            if(x(1) /= 2) error stop
+        end if  
     end subroutine temp
     subroutine temp2(call_back)
         implicit none
--- a/integration_tests/procedure_08.f90
+++ b/integration_tests/procedure_08.f90
@@ -10,6 +10,7 @@ contains
         integer, intent(in), optional :: x(:)
         logical :: y
         y = present(x)
+        print *, y
         if (y) error stop
     end subroutine calfun
 end module procedure_08_module
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3938,7 +3938,7 @@ public:
 
     bool is_function_variable(const ASR::Variable_t &v) {
         if (v.m_type_declaration) {
-            return ASR::is_a<ASR::Function_t>(*v.m_type_declaration);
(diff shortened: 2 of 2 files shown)


Pranavchiku (approved):

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9289,13 +9289,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             proc_sym = clss_proc->m_proc;
         } else if (ASR::is_a<ASR::Variable_t>(*proc_sym)) {
             ASR::symbol_t *type_decl = ASR::down_cast<ASR::Variable_t>(proc_sym)->m_type_declaration;
-            LCOMPILERS_ASSERT(type_decl);
```
gxyd, line 9292: these changes are straightforward and look good.

integration_tests/procedure_08.f90
```diff
@@ -10,6 +10,7 @@ subroutine calfun(x)
         integer, intent(in), optional :: x(:)
         logical :: y
         y = present(x)
+        print *, y
```
gxyd, line 13: Is this acting as a test case here?

gxyd (commented):
I see a change, but I don't see any case which will fail in current *main*, but pass in this PR.

gxyd:
I'm sorry, there is a long discussion in the linked issue, but if could help me understand what exactly acts as an MRE here, that would be helpful.

jinangshah21:
Consider MRE in https://github.com/lfortran/lfortran/issues/5603#issue-2722844558
In this when we do `call call_back()` in `temp` function it calls `cb` but in actual case it should call `calfun`. Previously call was pointing to `m_type_declaration` (which is `cb` in our MRE), but now it points to variable which will call `calfun` at runtime.

integration_tests/procedure_08.f90
jinangshah21: It is just to make sure that `calfun` is called in our test instead of `cb`.

gxyd:
If that's the case, do you think we should be able to add a test case signifying that it actually calls `calfun` and not `cb`?

gxyd:
Like, maybe make `intent(inout)` in both `calfun` and `cb` and they both set the value for `x` differently, and in program use `if (...) error stop` to ensure that it's set correctly.

certik:
Let's add an MRE that fails in main and works here.

gxyd (dismissed):
Looks good to me.

integration_tests/procedure_09.f90
```diff
@@ -0,0 +1,27 @@
...
+    use procedure_09_module
+    call temp(calfun)
+contains
+    subroutine temp(call_back)
+        implicit none
+        procedure(cb) :: call_back
+        integer :: x(4)
+        call call_back(x)   
+        print *, x
+        if(x(1) /= 2) error stop
```
gxyd, line 25: Wait a second, why can't all of this be done in `procedure_08.f90`? the file `procedure_09.f90` looks very similar.

certik (approved):
I think this is good, thanks!

integration_tests/procedure_09.f90
jinangshah21: I have done it.

certik (approved):

Pranavchiku:
Shall we merge this?

PR #5681, opened 2024-12-13
Remove ``NO_EXPERIMENTAL_SIMPLIFIER`` completely from all integration_tests

czgdp1807:
Nothing to do with simplifier I think. `openmp` is applied before `implied_do_loops` in LFortran without simplifier pass. So, I did the same with simplifier pass and it works. Nothing related to simplifier pass.

I checked and it generates similar code for `openmp` with and without simplifier pass. Would request someone from the team to verify.

### DIFF
--- a/src/libasr/pass/pass_manager.h
+++ b/src/libasr/pass/pass_manager.h
@@ -222,8 +222,8 @@ namespace LCompilers {
                 "global_stmts",
                 "init_expr",// This pass shouldn't be needed.
                 "function_call_in_declaration",
-                "implied_do_loops", // Should be implemented when optimisations for ImpliedDoLoop are possible in LFortran, until then not needed.
                 "openmp",
+                "implied_do_loops", // Should be implemented when optimisations for ImpliedDoLoop are possible in LFortran, until then not needed.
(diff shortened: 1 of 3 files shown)


src/libasr/pass/pass_manager.h
```diff
@@ -222,8 +222,8 @@ namespace LCompilers {
                 "global_stmts",
                 "init_expr",// This pass shouldn't be needed.
                 "function_call_in_declaration",
-                "implied_do_loops", // Should be implemented when optimisations for ImpliedDoLoop are possible in LFortran, until then not needed.
                 "openmp",
+                "implied_do_loops", // Should be implemented when optimisations for ImpliedDoLoop are possible in LFortran, until then not needed.
```
gxyd, line 226: Maybe these comments next to it can be removed?

gxyd: Considering that we do actually need this pass.

czgdp1807: Yes.

Pranavchiku:
If CI runs, let's do in subsequent PRs as it is minor and simplifier is actively worked on. Let's not waste one more run and around 30 minutes here. I am fine to do it in this PR itself as well.

czgdp1807:
> I am fine to do it in this PR itself as well.

What? Removal of comments? Yes. Will do later.

certik (approved):

PR #5682, opened 2024-12-13
fix: `is_statement_function` to not consider scalar constants arguments

Pranavchiku:
While experimenting on examples with error resiliency, I found:

```fortran
program main
integer :: xy
xy(1) = 1
end program
```

Not working with `--continue-compilation` and `xy` being treated as `statement` function. This PR strengthens the logic for checking `statement` function, and makes the example error resilient.

### DIFF
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -93,6 +93,8 @@ call bpe()
 i = bpe()
 print *, xx
 test_re = 1245.13
+c(1) = 1
+
 contains
 subroutine bpe()
     print *, size(bpe)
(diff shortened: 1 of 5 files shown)


tests/errors/continue_compilation_3.f90
```diff
@@ -9,6 +9,7 @@ program continue_compilation_3
 rea :: test_real(12)
 real :: test_re()
 int ::
+integer :: xy
```
certik, line 12: If you reuse `c` from above, then you get no line changes difference in the output.

certik (approved):
I think this is fine.

parth121101 (approved):

Issue #4319, opened 2024-06-19
Add MLIR backend


certik:
First let's generate text, later on we'll upgrade it to generate MLIR bytecode directly.

First example:
```mlir
func.func @main() -> i32 {
  %arg0 = arith.constant 7 : i32
  %0 = math.ctlz %arg0 : i32
  func.return %0 : i32
}
```
This can run as follows:
```console
$ mlir-opt a.mlir --pass-pipeline="builtin.module(convert-math-to-funcs{convert-ctlz}, func.func(convert-scf-to-cf,convert-arith-to-llvm), convert-func-to-llvm, convert-cf-to-llvm, reconcile-unrealized-casts)" | mlir-cpu-runner -e main -entry-point-result=i32
29
```

Thirumalai-Shaktivel:
https://github.com/lfortran/lfortran/pull/4576#issuecomment-2263426643

PR #5609, opened 2024-12-07
feat: [MLIR] Add basic fixed array support

Thirumalai-Shaktivel:
Towards: https://github.com/lfortran/lfortran/issues/4319

PR #5691, opened 2024-12-14
[MLIR] Handle and fix Print and FileWrite node

Thirumalai-Shaktivel:
Towards: https://github.com/lfortran/lfortran/issues/4319

### DIFF
--- a/src/libasr/codegen/asr_to_mlir.cpp
+++ b/src/libasr/codegen/asr_to_mlir.cpp
@@ -234,7 +234,9 @@ public:
         if (ASR::is_a<ASR::StringFormat_t>(*x.m_text)) {
             ASR::StringFormat_t *sf = ASR::down_cast<ASR::StringFormat_t>(
                 x.m_text);
-            args.reserve(al, sf->n_args);
+            args.reserve(al, sf->n_args+1);
+            // Later used by `printf_fmt`
+            args.push_back(al, nullptr);
             for (size_t i=0; i<sf->n_args; i++) {
                 ASR::ttype_t *t = ASRUtils::expr_type(sf->m_args[i]);
                 this->visit_expr(*sf->m_args[i]);
@@ -276,8 +278,9 @@ public:
             loc, global_str);
         globalPtr = builder->create<mlir::LLVM::GEPOp>(loc, llvmI8PtrTy,
             globalPtr, mlir::ValueRange{zero, zero});
-        builder->create<mlir::LLVM::CallOp>(loc, fn, mlir::ValueRange{globalPtr,
-            args[0]});
+        args.p[0] = globalPtr;
+        builder->create<mlir::LLVM::CallOp>(loc, fn,
+            mlir::ValueRange{args.as_vector()});
     }
 
 };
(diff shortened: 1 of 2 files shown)


src/libasr/codegen/asr_to_mlir.cpp
```diff
@@ -226,58 +233,76 @@ class ASRToMLIRVisitor : public ASR::BaseVisitor<ASRToMLIRVisitor>
...
                 }
             }
         } else {
-            throw CodeGenError("Unsupported expression as formatter in print",
-                x.base.base.loc);
+            throw CodeGenError("Unsupported expression as formatter in print", l);
         }
         fmt += "\n";
 
+        static int itr = 0; ++itr;
```
Thirumalai-Shaktivel, line 263: I hope, `static` is okay here, otherwise we can use a global variable.

src/libasr/codegen/asr_to_mlir.cpp
```diff
@@ -226,58 +233,76 @@ class ASRToMLIRVisitor : public ASR::BaseVisitor<ASRToMLIRVisitor>
...
-        mlir::LLVM::LLVMVoidType voidTy = mlir::LLVM::LLVMVoidType::get(context.get());
-        mlir::LLVM::LLVMFunctionType llvmFnType = mlir::LLVM::LLVMFunctionType::get(
-            voidTy, llvmI8PtrTy, true);
-        mlir::LLVM::LLVMFuncOp fn = builder0.create<mlir::LLVM::LLVMFuncOp>(
-            loc, "printf", llvmFnType);
+        // FIXME: Add "\00" to end of `builder->getStringAttr(fmt)` string.
+        // Now, the fmt string value is " %d\n", but it needs to be " %d\n\00"
+        mlir::LLVM::GlobalOp global_str = builder0.create<mlir::LLVM::GlobalOp>(
+            loc, arrayI8Ty, false, mlir::LLVM::Linkage::Private,
+            "printf_fmt_" + std::to_string(itr), builder->getStringAttr(fmt));
```
Thirumalai-Shaktivel, line 281: There is a issue here in `builder->getStringAttr(fmt))`,
Consider an example:
```fortran
program expr2
integer :: x

x = (2+3)*5
print *, x
write(*,*) x

end program
```
For which, I get the output as,
```console
$ lfortran examples/expr2.f90 --backend=mlir  
 25
 25
 -28655536
```
Where, ` -28655536` shouldn't be printed

The reason is, the global string `printf_fmt_2` and `printf_fmt_1` value doesn't have a null char `\00` at the end.
```mlir
module @LFortran {
  llvm.mlir.global private @printf_fmt_2(" %d\0A")
  llvm.func @printf(!llvm.ptr<i8>, ...)
  llvm.mlir.global private @printf_fmt_1(" %d\0A")
  llvm.func @main(%arg0: !llvm.ptr<i8>, ...) -> i32 {
  // [...]
  }
}
```

I will submit a new PR to fix this.

Thirumalai-Shaktivel:
I need this PR for https://github.com/lfortran/lfortran/pull/5609, so, enabling auto-merge. 
If anything needs to be fixed here, please write it down here.
I will fix them and submit a new PR.

PR #5704, opened 2024-12-15
[CI] Debug and Fix the DockerFile CI failure 

Thirumalai-Shaktivel:


Dockerfile
```diff
@@ -8,7 +8,8 @@ RUN apt install curl git build-essential binutils-dev zlib1g-dev clang libunwind
 RUN curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
 RUN bash Miniforge3-$(uname)-$(uname -m).sh -b
 
-RUN /root/miniforge3/bin/mamba init bash
```
gxyd, line 11: So this command used to work earlier, and now it fails, what changed that this command fails now?

Dockerfile
```diff
@@ -8,7 +8,8 @@ RUN apt install curl git build-essential binutils-dev zlib1g-dev clang libunwind
 RUN curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
```
gxyd, line 8: I see the changelog for miniforge to have this:
https://github.com/conda-forge/miniforge/compare/24.11.0-0...24.11.0-1, and one of the commit says:
https://github.com/conda-forge/miniforge/commit/b1f83a9c232ae1c375ac96184795988c7dc640d7

`Update to mamba 2.0.4`. Maybe new version of mamba doesn't support this?

gxyd: The older version of mamba used was 1.5.11 and now it's 2.0.4 I think

gxyd (approved):
The changes look good to me. Thanks @Thirumalai-Shaktivel 

As a separate question, what resource did you use to figure out the solution here? I'm interested to know, for me the solution wasn't too obvious.

Thirumalai-Shaktivel:
I was able to reproduce locally using UTM. And tried every steps possible. Didn't use any resource though.

Thirumalai-Shaktivel:
Thanks for the review!

PR #5705, opened 2024-12-15
Update installation.md

saikarna913:
The extra `\` is causing the command to fail when it is run immediately after being copied.


gxyd (approved):
Thanks for the fix here @saikarna913.

The CI failure is unrelated to your changes, and I'll rebase it on top of *main*

gxyd:
Sorry, I merged *main*, instead of rebase, but I guess that will also do.

PR #5708, opened 2024-12-15
fix: add `AST` node for `allocate` in programs without `program` line

### LINKED ISSUE #5707: allocate statement disallowed!
This program uses integer arrays ```x(0:2)``` and allocatable ```y(:)```, which is allocated with the same bounds.
```
  integer x(0:2)
  integer,allocatable::y(:)
  x(:) = [1,2,3]
  x(1:) = [10,20]
  print "(3(1X,I0))",x
  allocate(y(0:2))
  y(:) = [1,2,3]
  y(1:) = [10,20]
  print "(3(1X,I0))",y
end program
```
Lfortran 0.42.0 misinterorets the allocate statement.
```
 + lfortran --std=f23 -o badarray.f90oL badarray.f90
syntax error: Statement or Declaration expected inside program, found Expression
 --> badarray.f90:6:3
  |
6 |   allocate(y(0:2))
  |   ^^^^^^^^^^^^^^^^ 
```
Run-time outputs from gfortran, g95, ifort, ifx and AMD flang are identical and correct:
```
 1 10 20
 1 10 20
```

kmr-srbh:
fixes #5707

- When no `program` line is used, we handle the `allocate` statement as a general `AST::FuncCallOrArray_t` node, which is also a `AST::expr_t`.

- Previously, we handled `AST::expr_t` only for `AST::Name_t` and skipped this case, causing errors.

- In the current implementation, we create the required `AST::Allocate_t` node and add to the program body.


tests/reference/ast-allocate_statement_without_program_line-7b3307b.stdout
```diff
@@ -0,0 +1,168 @@
...
+                0)]
+                []
+                []
+                []
+            )
+            ()
+            0)]
+            []
+            ()
+        )
```
kmr-srbh, line 116: This is the added AST node.

tests/allocate_statement_without_program_line.f90
```diff
@@ -0,0 +1,10 @@
+integer x(0:2)
+integer,allocatable::y(:)
+x(:) = [1,2,3]
+x(1:) = [10,20]
+print "(3(1X,I0))",x
+allocate(y(0:2))
+y(:) = [1,2,3]
+y(1:) = [10,20]
+print "(3(1X,I0))",y
+end program
```
Thirumalai-Shaktivel, line 10: NIT:

Always add an empty new line to the end of the file.

tests/tests.toml
```diff
@@ -4435,3 +4435,8 @@ fortran = true
 [[test]]
 filename = "polymorphic_class_in_derived_type.f90"
 asr = true
+
+[[test]]
+filename = "allocate_statement_without_program_line.f90"
+ast = true
+run = true
```
Thirumalai-Shaktivel, line 4442: It is recommended to add the test to integration_tests, if we want to check the runtime output.
If you are checking the output printing, then this is fine.

tests/reference/run-allocate_statement_without_program_line-e7c2de5.stdout
```diff
@@ -0,0 +1,5 @@
+ 10 20 4
```
Thirumalai-Shaktivel, line 1: Here is the issue for the failure. I'm getting the following in MacOS
```console
$ lfortran tests/allocate_statement_without_program_line.f90 
 10 20 0
 1 
 10 
 20 
```

tests/tests.toml
Thirumalai-Shaktivel: I'm not sure, if we properly support `integer x(0:2)`. I think the output is wrong here.
See, GFortran:
```console
$ gfortran tests/allocate_statement_without_program_line.f90 && ./a.out          
 1 10 20
 1 10 20
```

tests/allocate_statement_without_program_line.f90
Thirumalai-Shaktivel: How about we only add allocatable and allocate(..) here. And only add AST test in tests.toml.

Thirumalai-Shaktivel (approved):
LGTM. Let's fix the failure and merge this.

tests/reference/run-allocate_statement_without_program_line-e7c2de5.stdout
kmr-srbh: Yes, we have an issue with handling assignments to arrays with lower bound lesser than 1. I opened an issue here - #5715.

Thirumalai-Shaktivel: Great, let's fix the failure then.

tests/allocate_statement_without_program_line.f90
kmr-srbh: > How about we only add allocatable and allocate(..) here. And only add AST test in tests.toml.

Yes, I will reduce the test size.

tests/tests.toml
kmr-srbh: > It is recommended to add the test to integration_tests, if we want to check the runtime output. If you are checking the output printing, then this is fine.

Yes, I had wished to check the AST output for this, so I added it here.

Thirumalai-Shaktivel:
Let me explain, how this handled in the parser.yy
```
units
    : units script_unit  %dprec 9  { RESULT($2); }
    | script_unit        %dprec 10 { RESULT($1); }
    | sep
    ;

script_unit
    : module
    | submodule
    | block_data
    | program
    | subroutine
    | procedure
    | function
    | use_statement
    | implicit_statement
    | var_decl           %dprec 9
    | derived_type_decl
    | enum_decl
    | statement          %dprec 7
    | expr sep           %dprec 8
    | KW_END_PROGRAM sep { $$ = SYMBOL($1, @$); }
    ;
```
This lines are responsible for recognising all the nodes in the TranslationUnit. 
So for your case as it doesn't have the program keyword. It won't be parsed through
```
script_unit
    : module
    | submodule
    | block_data
    | program     <-----
    | subroutine
[...]
``` 
But instead, get parsed through:
```
    | enum_decl
    | statement          %dprec 7
    | expr sep           %dprec 8.   <-----
```
Here, as you can see expr has higher precedence than statement. So, it is parsed as expr -> FuncCallOrArray.

Thirumalai-Shaktivel:
Wait a minute, after thinking about this, reducing the precedence value actually fixes the issue.
```diff
diff --git a/src/lfortran/parser/parser.yy b/src/lfortran/parser/parser.yy
index cf952e6bc..5047d3a06 100644
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -616,8 +616,8 @@ script_unit
     | var_decl           %dprec 9
     | derived_type_decl
     | enum_decl
-    | statement          %dprec 7
-    | expr sep           %dprec 8
+    | statement          %dprec 8
+    | expr sep           %dprec 7
     | KW_END_PROGRAM sep { $$ = SYMBOL($1, @$); }
     ;
```

Can you check if it works for all your cases and does it passes the tests?
If so, then I think this is fine.

kmr-srbh:
Thanks! The above works correctly for all the cases! :)

Thirumalai-Shaktivel:
Excellent. @certik, can you also review this before merging.

kmr-srbh:
Ah, we have some failures. Looking into it.

kmr-srbh:
Here is the current failure:

```shell
--------------------------------------------------------------------------------
SRC:
exit
    enddo
SRC -> AST:
(Exit 0 () ())
AST -> SRC:
exit
enddo
/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:236:
TEST CASE:  exit

/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:242: ERROR: CHECK( p(al, R"(exit
    enddo)") == "exit" ) is NOT correct!
  values: CHECK( (Exit 0 () ()) == exit )

--------------------------------------------------------------------------------
SRC:
cycle
    enddo
SRC -> AST:
(Cycle 0 () ())
AST -> SRC:
cycle
enddo
/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:248:
TEST CASE:  cycle

/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:251: ERROR: CHECK( p(al, R"(cycle
    enddo)") == "cycle" ) is NOT correct!
  values: CHECK( (Cycle 0 () ()) == cycle )

--------------------------------------------------------------------------------
SRC:
return
    enddo
SRC -> AST:
(Return 0 () ())
AST -> SRC:
return
enddo
/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:257:
TEST CASE:  return

/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/tests/test_pickle.cpp:260: ERROR: CHECK( p(al, R"(return
    enddo)") == "return" ) is NOT correct!
  values: CHECK( (Return 0 () ()) == return )
```

These 3 tests fail because the `AST -> SRC` does not match the provided value. For the first test, the code
```fortran
exit
enddo
```
returns the same code for `AST -> SRC` as source. Is the test probably incorrect? How are the above codes compiled?

kmr-srbh:
Are `exit` and `return` handled as variables to ease the internal implementation for the REPL, or can we not handle `ASR::stmt_` there?

src/lfortran/parser/parser.yy
```diff
@@ -616,8 +616,8 @@ script_unit
     | var_decl           %dprec 9
     | derived_type_decl
     | enum_decl
-    | statement          %dprec 7
-    | expr sep           %dprec 8
+    | statement          %dprec 8
+    | expr sep           %dprec 7
```
certik, line 620: As @Shaikh-Ubaid said, we want to be very careful here, most likely this is not correct.

A loose `return` we want to interpret as a variable, because in the global scope this will most likely be a variable. Can it even be a "return"?

How is `allocate` currently parsed? We might need to do some logic there to fix it.

certik:
The fundamental ambiguity is that in interactive mode, at the global scope, `allocate(a(5))` can either be the `allocate` statement, or an array expression or a function call. We can't resolve this at the AST level, but parsing it as an expression should work, and then in AST->ASR we can decide what to do.

What should we do? Some options:

* consult the symbol table, and if `allocate` is not an array, interpret it as an allocate statement. If it is an array, we either interpret as an array, or maybe we'll always interpret as an allocate statement.
* we always interpret as an allocate statement. If you call your array `allocate` (extremely rare), then you can't print it interactively like this, since it collides with a statement, you would have to use "print *, allocate(a(5))", which should work.

I think we should always interpret global `allocate` as a statement. This can be done in AST->ASR. Other things like `exit`, `return`, `cycle` I think can't appear in the global scope anyway, they are always inside a block, so at the global scope that should be interpreted as variables.

src/lfortran/parser/parser.yy
kmr-srbh: When no `program` line is used, we parse it as `AST::expr_t`.

Yes, the initial fix I had pushed was creating the `AST::Allocate_t` node for this, but we found we would need to do this for many other statements, so we tried to think more generally. Do we want to handle this for every such node specifically? 

Discussed in https://github.com/lfortran/lfortran/pull/5708#issuecomment-2543888561 and the comments following it.

certik: I think I discussed the options here: https://github.com/lfortran/lfortran/pull/5708#issuecomment-2556304213, let me know if you have any questions.

kmr-srbh: Thanks! I reread the idea you proposed and my questions are resolved now. I have a fair idea how to proceed. I will push the changes on Monday.

kmr-srbh:
Ready.

Issue #5690, opened 2024-12-13
codegen error for an array initialized with scalar value in module


gxyd:
## Minimal Reproducible Example

For the below Fortran program:
```f90
module meshdef
    implicit none
    integer, dimension(2) :: drratio=0 ! if this was [1, 2] it works
end module

program main
    implicit none
end program main

```

on executing it with LFortran, I get the below error:
```console
> lfortran test.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

@drratio = global [2 x i32] zeroinitializer

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  call void @_lpython_free_argv()
  br label %return

return:                                           ; preds = %.entry
  ret i32 0
}

declare void @_lpython_call_initial_functions(i32, i8**)

declare void @_lpython_free_argv()
asr_to_llvm: module failed verification. Error:
Global variable initializer type does not match global variable type!
[2 x i32]* @drratio

code generation error: asr_to_llvm: module failed verification. Error:
Global variable initializer type does not match global variable type!
[2 x i32]* @drratio



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

while the above program works without error with GFortran.

Pranavchiku:
The difference is with `[1,2]` it produces:

```llvm
@drratio = global [2 x i32] [i32 1, i32 2]
```

and with `0` it gives

```llvm
@drratio = global [2 x i32] zeroinitializer
```

But we need 

```
@drratio = global [2 x i32] [i32 0, i32 0]
```

kmr-srbh:
The issue here is related specifically to initializing with `0`. This is a special case because of the way LLVM handles it. When we initialize the array in the above MRE, we loop over the elements in the `ArrayConstant` and add to a vector as `llvm::Constant*`. Post this we initialize the global variable with a `llvm::ConstantArray` created from the same vector. In the above case, that vector contains only 0s, so LLVM initializes the variable with `zeroinitializer` instead of `[i32 0, i32 0]`. The following diff mitigates the issue:
```diff
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index 3c2dddb8f..a8da0e4c2 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
[...]

PR #5710, opened 2024-12-15
Fix: Handle if array initializer is zero

adit4443ya:
Towards #5690 
In this PR i have changed 
```fortran
integer, dimension(2) :: drratio=0 --------> integer,dimension(2) :: drratio
real, dimension(2) :: drratior=0.0 ---------> real, dimension(2) :: drratior
```
On ASR level 
Which prevents this llvm error 
```clojure
aditya-trivedi   lfortran    main ≡    lfortran integration_tests/array_init_with_scalar.f90 --show-llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

@drratio = global [2 x i32] zeroinitializer
@drratior = global [2 x float] zeroinitializer

.entry:
  br label %return

  ret i32 0
}

[2 x i32]* @drratio
[2 x float]* @drratior

[2 x i32]* @drratio
[2 x float]* @drratior



```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1887,6 +1887,8 @@ RUN(NAME openmp_41 LABELS llvm_omp llvm)
 RUN(NAME fortran_01 LABELS gfortran llvm fortran)
 RUN(NAME fortran_02 LABELS gfortran fortran)
 
+RUN(NAME array_init_with_scalar_main LABELS gfortran llvm)
+
 RUN(NAME compiler_version_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME exit_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -2855,6 +2855,7 @@ public:
                 ASR::presenceType s_presence = dflt_presence;
                 ASR::storage_typeType storage_type = dflt_storage;
                 bool value_attr = false;
+                bool is_initializer_zero = false;
                 AST::AttrType_t *sym_type =
                     AST::down_cast<AST::AttrType_t>(x.m_vartype);
                 bool is_char_type = sym_type->m_type == AST::decl_typeType::TypeCharacter;
@@ -3293,6 +3294,10 @@ public:
                         );
                         LCOMPILERS_ASSERT(ASR::is_a<ASR::ArrayConstant_t>(*init_expr));
                         value = init_expr;
+                        if((ASR::is_a<ASR::IntegerConstant_t>(*tmp_init) && ASR::down_cast<ASR::IntegerConstant_t>(tmp_init)->m_n==0) ||
+                            (ASR::is_a<ASR::RealConstant_t>(*tmp_init) && ASR::down_cast<ASR::RealConstant_t>(tmp_init)->m_r==0.0)) {
+                            is_initializer_zero = true;
+                        }
                     }
                     ASR::ttype_t *init_type = ASRUtils::expr_type(init_expr);
 
@@ -3612,6 +3617,11 @@ public:
                     if ( !is_implicitly_declared && !is_external) {
                         SetChar variable_dependencies_vec;
                         variable_dependencies_vec.reserve(al, 1);
+                        if(is_initializer_zero) {
+                            value = nullptr;
+                            init_expr = nullptr;
+                            storage_type=ASR::storage_typeType::Default;
+                        }
                         ASRUtils::collect_variable_dependencies(al, variable_dependencies_vec, type, init_expr, value);
                         ASR::asr_t *v = ASR::make_Variable_t(al, s.loc, current_scope,
                                 s2c(al, to_lower(s.m_name)), variable_dependencies_vec.p,
--- a/integration_tests/array_init_with_scalar.f90
+++ b/integration_tests/array_init_with_scalar.f90
@@ -1,0 +1,9 @@
+module array_init_with_scalar_0
+    implicit none
+    integer, dimension(2) :: drratio=0
+    real, dimension(2) :: drratior=0.0
+end module
+
+program array_init_with_scalar_main
+implicit none
+end program array_init_with_scalar_main


integration_tests/CMakeLists.txt
```diff
@@ -1887,6 +1887,8 @@ RUN(NAME openmp_41 LABELS llvm_omp llvm)
 RUN(NAME fortran_01 LABELS gfortran llvm fortran)
 RUN(NAME fortran_02 LABELS gfortran fortran)
 
+RUN(NAME array_init_with_scalar_main LABELS gfortran llvm)
```
kmr-srbh, line 1890:
```suggestion
RUN(NAME array_init_with_scalar LABELS gfortran llvm)
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3293,6 +3294,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         );
                         LCOMPILERS_ASSERT(ASR::is_a<ASR::ArrayConstant_t>(*init_expr));
                         value = init_expr;
+                        if((ASR::is_a<ASR::IntegerConstant_t>(*tmp_init) && ASR::down_cast<ASR::IntegerConstant_t>(tmp_init)->m_n==0) ||
+                            (ASR::is_a<ASR::RealConstant_t>(*tmp_init) && ASR::down_cast<ASR::RealConstant_t>(tmp_init)->m_r==0.0)) {
+                            is_initializer_zero = true;
+                        }
```
kmr-srbh, line 3300: I think doing it this way is simpler than setting `init_expr` and `value` to `nullptr` here and handling work related to the RHS array later, which can get more complicated.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3612,6 +3617,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     if ( !is_implicitly_declared && !is_external) {
                         SetChar variable_dependencies_vec;
                         variable_dependencies_vec.reserve(al, 1);
+                        if(is_initializer_zero) {
+                            value = nullptr;
+                            init_expr = nullptr;
+                            storage_type=ASR::storage_typeType::Default;
```
kmr-srbh, line 3623: I think this change was made because of issue #5714. I think we do not want to change the storage type.

kmr-srbh (approved):

gxyd:
I'm curious, why do we've a special handling for just zero? It can be initialized to non-zero value as well, right? Like:

```f90
module array_init_with_scalar_0
    implicit none
    integer, dimension(2) :: drratio=5
end module
```

Any reason for just a special treatment to zero?

Also, I'm not sure if this approach takes into account that an integer value might be on the right hand side of a real variable declaration as well.

adit4443ya:
> I'm curious, why do we've a special handling for just zero? It can be initialized to non-zero value as well, right? Like:

For your mentioned case 
This branch outputs this ASR which is correct
```clojure
 aditya-trivedi   lfortran    scalar ≢  ~1    lfortran integration_tests/array_init_with_scalar.f90 --show-asr
(TranslationUnit
    (SymbolTable
        1
        {
            array_init_with_scalar:
                (Program
                    (SymbolTable
                        3
                        {
                            
                        })
                    array_init_with_scalar
                    []
                    []
                ),
            array_init_with_scalar_0:
                (Module
                    (SymbolTable
                        2
                        {
                            drratio:
                                (Variable
                                    2
                                    drratio
                                    []
                                    Local
                                    (ArrayConstant
                                        8
                                        [5, 5]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    (ArrayConstant
                                        8
                                        [5, 5]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    Save
                                    (Array
                                        (Integer 4)
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                ),
                            drratior:
                                (Variable
                                    2
                                    drratior
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Array
                                        (Real 4)
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
                        })
                    array_init_with_scalar_0
                    []
                    .false.
                    .false.
                )
        })
    []
)
 aditya-trivedi   lfortran    scalar ≢  ~1   

```

adit4443ya:
For your Question of why special treatment of Zero
Refer this [comment](https://github.com/lfortran/lfortran/issues/5690#issuecomment-2543099284)

adit4443ya:
> Also, I'm not sure if this approach takes into account that an integer value might be on the right hand side of a real variable declaration as well.
Yeah i will look for this

src/lfortran/semantics/ast_common_visitor.h
adit4443ya: Yeah that's what I encountered

Thirumalai-Shaktivel:
Can you please share the LLVM IR for an MRE using this PR.

I feel this comment: https://github.com/lfortran/lfortran/issues/5690#issuecomment-2542846818 would be the right way to fix this. What do you think?

adit4443ya:
>Can you please share the LLVM IR for an MRE using this PR.
```clojure
 aditya-trivedi   lfortran    scalar ≢  ~2    lfortran integration_tests/array_init_with_scalar.f90 --show-llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

@drratio = global [2 x i32] zeroinitializer
@drratior = global [2 x float] zeroinitializer

.entry:
  br label %return

  ret i32 0
}

```

Thirumalai-Shaktivel:
I'm confused, both the LLVM IR (before and after this PR) seems exactly the same. How is the issue fixed here?

Maybe I think the right way to fix this is:
```
@drratio = global [2 x i32] [i32 0, i32 0]
```

adit4443ya:
On main we get this error
```clojure
 aditya-trivedi   lfortran    main ↓2    lfortran integration_tests/array_init_with_scalar.f90 --show-llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

@drratio = global [2 x i32] zeroinitializer
@drratior = global [2 x float] zeroinitializer

.entry:
  br label %return

  ret i32 0
}

[2 x i32]* @drratio
[2 x float]* @drratior

[2 x i32]* @drratio
[2 x float]* @drratior



```

adit4443ya:
> On main we get this error
> 
> > I'm confused, both the LLVM IR (before and after this PR) seems exactly the same. How is the issue fixed here?
> 
> ```clojure
>  aditya-trivedi   lfortran    main ↓2    lfortran integration_tests/array_init_with_scalar.f90 --show-llvm
> ; ModuleID = 'LFortran'
> source_filename = "LFortran"
> 
> @drratio = global [2 x i32] zeroinitializer
> @drratior = global [2 x float] zeroinitializer
> 
> define i32 @main(i32 %0, i8** %1) {
> .entry:
>   call void @_lpython_call_initial_functions(i32 %0, i8** %1)
>   call void @_lpython_free_argv()
>   br label %return
> 
> return:                                           ; preds = %.entry
>   ret i32 0
> }
> 
> declare void @_lpython_call_initial_functions(i32, i8**)
> 
> declare void @_lpython_free_argv()
> asr_to_llvm: module failed verification. Error:
> Global variable initializer type does not match global variable type!
> [2 x i32]* @drratio
> Global variable initializer type does not match global variable type!
> [2 x float]* @drratior
> 
> code generation error: asr_to_llvm: module failed verification. Error:
> Global variable initializer type does not match global variable type!
> [2 x i32]* @drratio
> Global variable initializer type does not match global variable type!
> [2 x float]* @drratior
> 
> 
> 
> Note: Please report unclear, confusing or incorrect messages as bugs at
> https://github.com/lfortran/lfortran/issues.
> ```


As you can see here , there is type mismatch here , as LLVM initializes the variable with `zeroinitializer` instead of [i32 0, i32 0], raising this error

Thirumalai-Shaktivel:
You shared the LLVM IR using this PR: https://github.com/lfortran/lfortran/pull/5710#issuecomment-2543888029 as:
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

@drratio = global [2 x i32] zeroinitializer
@drratior = global [2 x float] zeroinitializer
; [...]
```

Which also does the `zeroinitializer`, why is it working in this PR?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3293,6 +3294,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         );
                         LCOMPILERS_ASSERT(ASR::is_a<ASR::ArrayConstant_t>(*init_expr));
                         value = init_expr;
+                        if((ASR::is_a<ASR::IntegerConstant_t>(*tmp_init) && ASR::down_cast<ASR::IntegerConstant_t>(tmp_init)->m_n==0) ||
+                            (ASR::is_a<ASR::RealConstant_t>(*tmp_init) && ASR::down_cast<ASR::RealConstant_t>(tmp_init)->m_r==0.0)) {
```
Thirumalai-Shaktivel, line 3298: This would fail for:
```fortran
module expr2
integer :: x(2) = [0,0]
end
```

Maybe we can handle this case in the backend itself, what do you think?

Thirumalai-Shaktivel:
Okay, I made some analysis in the LLVM backend. And it seems the passed values to
https://github.com/lfortran/lfortran/blob/381124d216e562cae68d9a008e556925637d93d2/src/libasr/codegen/asr_to_llvm.cpp#L2802
is 
```
@drratio = global [2 x i32] [i32 0, i32 0]
```
But LLVM represents it as 
```
@drratio = global [2 x i32] zeroinitializer
```

Maybe you can do this. 
Check if all the elements are zero if so then set it to NULL (aka zeroinitializer => getNullValue(...)).

Let me know your thoughts on this.

Thirumalai-Shaktivel:
Also, after finishing this. Please, verify it always prints `0` instead of some garbage value.

src/lfortran/semantics/ast_common_visitor.h
adit4443ya: I checked it this fails in main too..

adit4443ya:
>Let me know your thoughts on this.

I agree with your arguments , But only reason why i chose this approach is, if we make that check in backend then we are iterating through each element in LLVM backend again, which will be slower if the array is large and can hinder the speed,
But if we handle this in ASR , then it would save us cost of iterations, 

> ```fortran
> module expr2
> integer :: x(2) = [0,0]
> end
> ```

Also I would incorporate this case in this PR too,

Let me know your thoughts on this
@Thirumalai-Shaktivel @kmr-srbh on this

kmr-srbh:
Yes, I suggested the ASR changes due to the additional overhead of looping through each element (which is 0) again in the back-end. We loop over them during AST -> ASR, which is already a bottleneck. Please see #5718.

```fortran
module expr2
integer :: x(2) = [0,0]
end
```

The above is still an issue with this approach. We cannot be sure of each element being a 0 unless we loop over it, which is again slow.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku, line 3298: I am -1 with this approach of hardwiring values

kmr-srbh:
Please see https://github.com/lfortran/lfortran/pull/5719#discussion_r1886617563. I think modifying the ASR is not correct as it changes the compile-time nature of the array.

adit4443ya:
Okk It seems 5719 is correctly handling This issue 
I will close this PR

Issue #5713, opened 2024-12-15
program without line does not generate ast with continue-compilation flag


parth121101:
```
cat tests/errors/program_without_line_01.f90
print *, "OK"
contains
```

```
lfortran tests/errors/program_without_line_01.f90 --continue-compilation --show-ast
syntax error: Expected program end
 --> tests/errors/program_without_line_01.f90:1:1 - 2:8
  |
1 |    print *, "OK"
  |    ^^^^^^^^^^^^^...
...
  |
2 |    contains
  | ...^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Following are the files that have the issue.

- [ ] tests/errors/program_without_line_01.f90
- [ ] tests/errors/program_without_line_02.f90
- [ ] tests/errors/program_without_line_03.f90
- [ ] tests/errors/program_without_line_04.f90

Will try to fix this

Pranavchiku:
Closing with reference to #5752

PR #5716, opened 2024-12-15
fix: program without line generates ast with continue compilation flag

parth121101:
Towards: #5713

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -54,6 +54,7 @@ def single_test(test: Dict, verbose: bool, no_llvm: bool, skip_run_with_dbg: boo
     asr_implicit_interface_and_typing_with_llvm = is_included("asr_implicit_interface_and_typing_with_llvm")
     continue_compilation = is_included("continue_compilation")
     semantics_only_cc = is_included("semantics_only_cc")
+    syntax_only_cc = is_included("syntax_only_cc")
     asr_use_loop_variable_after_loop = is_included("asr_use_loop_variable_after_loop")
     asr_preprocess = is_included("asr_preprocess")
     asr_indent = is_included("asr_indent")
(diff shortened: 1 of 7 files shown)


parth121101:
with this PR `program_without_line.f90` files generates ast when continue-compilation flag is used. Here's the example:
```
cat tests/errors/program_without_line_01.f90
print *, "OK"
contains
```

```
lfortran tests/errors/program_without_line_01.f90 --con
tinue-compilation --show-ast
syntax error: Expected program end
 --> tests/errors/program_without_line_01.f90:1:1 - 2:8
  |
1 |    print *, "OK"
  |    ^^^^^^^^^^^^^...
...
  |
2 |    contains
  | ...^^^^^^^^ 


(TranslationUnit
    [(Print
        0
        ()
        [(String "OK")]
        ()
    )
    contains]
)
```

Thirumalai-Shaktivel:
I think AST is invalid here, don't you think?

Is this required in interactive shell?
I think the statements needs to be included inside a Module, Function or Subroutine. So, previously it used to throw correct error only right?

Would this change generate an ASR, can you please share it?

parth121101:
> I think AST is invalid here, don't you think?
> 
> Is this required in interactive mode? I think the statements needs to be included inside a Module, Function or Subroutine. So, previously it used to throw correct error.
> 
> Would this PR generate an ASR, can you please share it?

@Thirumalai-Shaktivel this does not generate ASR. Even the main branch is unable to generate the ASR at this point of time. So I believe this needs to be fixed as well.

Thirumalai-Shaktivel:
I have some doubts here, can you please clarify them?
From the issue, I can see you are trying to generate AST for the following tests:
```
tests/errors/program_without_line_01.f90
tests/errors/program_without_line_02.f90
tests/errors/program_without_line_03.f90
tests/errors/program_without_line_04.f90
```
- Why do we need to generate AST for these tests using continue-compilation flag?
- Where is it used?
- I checked GFortran, all the tests seems to be failing as invalid syntax, why generate ASR?

Thirumalai-Shaktivel:
Oh okay. But, I was curious to know where would this be used.

Thanks for the response.

certik:
Should we add some test for this?

parth121101:
> Should we add some test for this?

Sure @certik

Pranavchiku:
I think a new file will works.

certik:
Thanks!

PR #5719, opened 2024-12-15
fix: handle zero initialization of array in LLVM backend

### LINKED ISSUE #5690: codegen error for an array initialized with scalar value in module

gxyd:
## Description

Fixes #5690

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3266,8 +3266,10 @@ public:
                     }
                     if (!is_compile_time && ASR::is_a<ASR::Array_t>(*type)
                         && (ASR::is_a<ASR::IntegerConstant_t>(*tmp_init) || ASR::is_a<ASR::RealConstant_t>(*tmp_init)
-                            || ASR::is_a<ASR::RealUnaryMinus_t>(*tmp_init) || ASR::is_a<ASR::IntegerUnaryMinus_t>(*tmp_init)
-                            || ASR::is_a<ASR::StringConstant_t>(*tmp_init) || ASR::is_a<ASR::LogicalConstant_t>(*tmp_init))) {
+                            || ASR::is_a<ASR::ComplexConstant_t>(*tmp_init) || ASR::is_a<ASR::RealUnaryMinus_t>(*tmp_init)
+                            || ASR::is_a<ASR::IntegerUnaryMinus_t>(*tmp_init) || ASR::is_a<ASR::StringConstant_t>(*tmp_init)
+                            || ASR::is_a<ASR::LogicalConstant_t>(*tmp_init)))
+                    {
                         /*
                             Case: integer :: x(2) = 1
                             which is equivalent to x(2) = [1,1]
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -2799,7 +2799,11 @@ public:
                         }
                     }
                     llvm::ArrayType* arr_type = llvm::ArrayType::get(type, arr_const_size);
-                    module->getNamedGlobal(x.m_name)->setInitializer(llvm::ConstantArray::get(arr_type, arr_elements));
+                    llvm::Constant* initializer = llvm::ConstantArray::get(arr_type, arr_elements);
+                    if (initializer->isElementWiseEqual(llvm::ConstantArray::getNullValue(arr_type))) {
+                        initializer = llvm::ConstantArray::getNullValue(type);
+                    }
+                    module->getNamedGlobal(x.m_name)->setInitializer(initializer);
                 } else {
                     module->getNamedGlobal(x.m_name)->setInitializer(llvm::ConstantArray::getNullValue(type));
                 }
--- a/integration_tests/module_array_initialization.f90
+++ b/integration_tests/module_array_initialization.f90
@@ -1,0 +1,18 @@
+module mod_module_array_init
+    implicit none
+    integer, dimension(2) :: arr_integer_with_zero = 0
+    real, dimension(2) :: arr_real_with_zero = 0
+    complex, dimension(2) :: arr_complex_with_zero = (0.0, 0.0)
+end module
+
+program module_array_init
+    use mod_module_array_init
+    implicit none
+    if (any(arr_integer_with_zero /= [0, 0])) error stop
+    if (any(arr_real_with_zero /= [0., 0.])) error stop
+    if (any(arr_complex_with_zero /= [(0., 0.), (0., 0.)])) error stop
+
+    if (kind(arr_integer_with_zero(1)) /= 4) error stop
+    if (kind(arr_real_with_zero(1)) /= 4) error stop
+    if (kind(arr_complex_with_zero(1)) /= 4) error stop
+end program


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2799,7 +2799,11 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         }
                     }
                     llvm::ArrayType* arr_type = llvm::ArrayType::get(type, arr_const_size);
-                    module->getNamedGlobal(x.m_name)->setInitializer(llvm::ConstantArray::get(arr_type, arr_elements));
+                    llvm::Constant* initializer = llvm::ConstantArray::get(arr_type, arr_elements);
+                    if (initializer->isElementWiseEqual(llvm::ConstantArray::getNullValue(arr_type))) {
+                        initializer = llvm::ConstantArray::getNullValue(type);
+                    }
+                    module->getNamedGlobal(x.m_name)->setInitializer(initializer);
```
kmr-srbh, line 2806: This approach is correct, but will be slow for large arrays. Although I think it is the right way to be sure of cases like
```fortran
module expr2
integer :: x(2) = [0,0]
end
```
where we need to know every element for using `zeroinitializer`. Ideally, letting LLVM directly use `zeroinitializer` for these cases would have been faster.

I think we can do both approaches:
1. Setting `m_value=nullptr` for arrays initialized using scalar 0. These arrays can be fairly large, so looping will be a bottleneck.
2. Using the above element-wise comparison approach for initialization using `ArrayConstant`.

Please let me know your thoughts on this.

gxyd:
I checked the time taken for the below Fortran program:
```f90
module mod_module_array_init
    implicit none
    integer, dimension(1000) :: arr_integer_with_zero = 0
end module

program module_array_init
    use mod_module_array_init
    implicit none
end program

```

it's full execution on this branch and below are the statistics for it, and did it for GFortran as well:
```console
> time gfortran test.f90 -o a.out && ./a.out
gfortran test.f90 -o a.out  0.09s user 0.05s system 111% cpu 0.121 total
> time lfortran test.f90  # it's on this branch
/Users/gxyd/OpenSource/lfortran/src/bin/lfortran test.f90  0.15s user 0.06s system 68% cpu 0.306 total
```

So, LFortran is slow, but not that slow compared to GFortran.

Now, let's see if this PR gets us slow or not. Consider the below Fortran program:
```f90
module mod_module_array_init
    implicit none
    integer, dimension(1000) :: arr_integer_with_zero = 1
end module

program module_array_init
    use mod_module_array_init
    implicit none
end program

```

I change the initialization above from `0` to `1` so that also works on our current *main* as well:
```console
> time main_lfortran test.f90  # our current main
/Users/gxyd/Documents/lfortran/src/bin/lfortran test.f90  0.16s user 0.02s system 120% cpu 0.151 total

> time lfortran test.f90   # this branch
/Users/gxyd/OpenSource/lfortran/src/bin/lfortran test.f90  0.15s user 0.03s system 110% cpu 0.155 total
```

so, not much of a slowdown from this PR.

src/libasr/codegen/asr_to_llvm.cpp
gxyd: The concern for checking is valid though. I'm not saying that I like the approach of ASR, but I can improve upon the changes already made in this PR.

kmr-srbh: Arrays of 0 are a special case actually, so I mentioned about the ASR approach for simplicity. Checking for each element is correct for initialization with `ArrayConstant`, but for scalars, we can save compile-time at 2 places by setting `m_value=nullptr` - `init_expr` pass and the backend.

gxyd: I'm sorry, I'm still not convinced by the argument 😄

kmr-srbh: I think the `init_expr` pass is the main bottleneck. If we can fix it, we can take either path. :)

gxyd: > I think the init_expr pass is the main bottleneck. If we can fix it, we can take either path. :)

Yes, I think that's probably the case.

Pranavchiku:
All this is completely fine, my argument is 

When `integer, dimension(2) :: drratio=[1,2]` gives the following ASR

```clojure
(Variable
                                    2
                                    drratio
                                    []
                                    Local
                                    (ArrayConstant
                                        8
                                        [1, 2]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    (ArrayConstant
                                        8
                                        [1, 2]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    Save
                                    (Array
                                        (Integer 4)
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
```

and

`integer, dimension(2) :: drratio=0`, gives exactly same ASR, why does it not create `@drratio = global [2 x i32] [i32 0, i32 0]`


```clojure
(Variable
                                    2
                                    drratio
                                    []
                                    Local
                                    (ArrayConstant
                                        8
                                        [0, 0]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    (ArrayConstant
                                        8
                                        [0, 0]
                                        (Array
                                            (Integer 4)
                                            FixedSizeArray
                                        )
                                        ColMajor
                                    )
                                    Save
                                    (Array
                                        (Integer 4)
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                )
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3266,8 +3266,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     }
                     if (!is_compile_time && ASR::is_a<ASR::Array_t>(*type)
                         && (ASR::is_a<ASR::IntegerConstant_t>(*tmp_init) || ASR::is_a<ASR::RealConstant_t>(*tmp_init)
-                            || ASR::is_a<ASR::RealUnaryMinus_t>(*tmp_init) || ASR::is_a<ASR::IntegerUnaryMinus_t>(*tmp_init)
-                            || ASR::is_a<ASR::StringConstant_t>(*tmp_init) || ASR::is_a<ASR::LogicalConstant_t>(*tmp_init))) {
+                            || ASR::is_a<ASR::ComplexConstant_t>(*tmp_init) || ASR::is_a<ASR::RealUnaryMinus_t>(*tmp_init)
+                            || ASR::is_a<ASR::IntegerUnaryMinus_t>(*tmp_init) || ASR::is_a<ASR::StringConstant_t>(*tmp_init)
+                            || ASR::is_a<ASR::LogicalConstant_t>(*tmp_init)))
```
Pranavchiku, line 3271: Why do we need this for this PR?

Pranavchiku (changes requested):
Left comments above.

src/lfortran/semantics/ast_common_visitor.h
gxyd: This change is need to handle the creation of Array from ComplexConstant, which wasn't handle before, it needs to be handle exactly the same way as `Real`, `Integer`, etc. are handled already.

The added test-case here: https://github.com/lfortran/lfortran/pull/5719/files#diff-813042eadcfdd746271040fb223da591f8bed094f5926951c95360941f9273b9R5, ensures that the RHS when it's a `ComplexConstant`, an array is created (earlier it wasn't the case).

So, this bring a minor change in the ASR, just for `ComplexConstant`

Pranavchiku: Let's add  tests for others as well i.e. `RealUnaryMinus_t`, etc.

gxyd:
> integer, dimension(2) :: drratio=0, gives exactly same ASR, why does it not create @drratio = global [2 x i32] [i32 0, i32 0]


It doesn't need to create `@drratio = global [2 x i32] [i32 0, i32 0]`. Earlier, the *printed* value was `zeroinitializer` which was created using:
```c++
initializer = llvm::ConstantArray::get(arr_type, arr_elements);
```

and now we are creating initializer using:
```c++
initializer = llvm::ConstantArray::getNullValue(type);
```
though, both in the output LLVM IR are printed the same i.e. `zeronitializer`, the difference lies in internal type that the LLVM IR expects, that's what I think is happening, I read the documentation here:
[getNullValue](https://www.llvm.org/doxygen/classllvm_1_1Constant.html#aa6574d526b3e38a28f688a7bb4325c2c).

I'm not claiming to understand why it expects `getNullValue` and not `llvm::ConstantArray::get`. But, definitely printing-wise there is no change in LLVM IR here.

Pranavchiku:
I think we have same understanding here, it makes sense.

src/libasr/codegen/asr_to_llvm.cpp
Pranavchiku: > Setting m_value=nullptr for arrays initialized using scalar 0. These arrays can be fairly large, so looping will be a bottleneck.

We cannot set this as it will change example from compile-time to run-time which is not what we want.

We can do this if it is not `parameter`, in that case symbolic value can be used.

No, evaluated symbolic expression is contained in `value`, so we cannot change that.

Pranavchiku: I think approach in this PR is fine.

Pranavchiku (approved):
Looks good to me. I am convinced with this approach.

src/lfortran/semantics/ast_common_visitor.h
gxyd: Yes, I think it's good add those, I'll do that.

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: > > Setting m_value=nullptr for arrays initialized using scalar 0. These arrays can be fairly large, so looping will be a bottleneck.
> 
> We cannot set this as it will change example from compile-time to run-time which is not what we want.
> 
> We can do this if it is not `parameter`, in that case symbolic value can be used.
> 
> No, evaluated symbolic expression is contained in `value`, so we cannot change that.

Ah, yes, that would certainly make the change from compile-time to runtime. I agree with the current argument of handling in the backend. Thanks a lot for stating this!

integration_tests/module_array_init.f90
```diff
@@ -0,0 +1,36 @@
+module mod_module_array_init
+    implicit none
+    integer, dimension(2) :: arr_integer_with_zero = 0
+    integer, dimension(3) :: arr_integer_with_unary_minus = -1
+    real, dimension(2) :: arr_real_with_zero = 0
+    real, dimension(4) :: arr_real_with_unary_minus = -1
+    complex, dimension(2) :: arr_complex_with_zero = (0.0, 0.0)
+    logical, dimension(2) :: arr_logical_with_false = .false.
+    logical, dimension(2) :: arr_logical_with_true = .true.
+    logical, dimension(2) :: arr_logical_with_true_false = [.true., .false.]
```
gxyd, line 10: even this fails in *main*, and generates an LLVM IR like:
```ll
@e = global [2 x i1] [[2 x i1] true, [2 x i1] false]


[2 x i1]* @e
```

certik (approved):
I think that's fine.

PR #5722, opened 2024-12-16
enh: set_allocation_size for 'spread' intrinsic

gxyd:
## Description

The removed `NO_EXPERIMENTAL_SIMPLIFIER` acts as a test to ensure that the added change works alright.


Pranavchiku (approved):
looks good

src/libasr/pass/simplifier.cpp
```diff
@@ -540,6 +540,37 @@ bool set_allocation_size(
                     }
                     break;
                 }
+                case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Spread): {
+                    size_t n_dims = ASRUtils::extract_n_dims_from_ttype(
+                        ASRUtils::expr_type(intrinsic_array_function->m_args[0]));
+                    LCOMPILERS_ASSERT(intrinsic_array_function->n_args = 3);
+
+                    // we just set it as 'dim = 1' for now
+                    size_t dim = 1;
```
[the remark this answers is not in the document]
gxyd: I've actually removed the use of `dim` argument, as we don't need it. The allocation size only depends on the rank of the input array and `ncopies` value.

src/libasr/pass/simplifier.cpp
```diff
@@ -540,6 +540,23 @@ bool set_allocation_size(
                     }
                     break;
                 }
+                case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Spread): {
+                    size_t n_dims = ASRUtils::extract_n_dims_from_ttype(
+                        ASRUtils::expr_type(intrinsic_array_function->m_args[0])
+                    );
```
Pranavchiku, line 546:
```suggestion
                    size_t n_dims = ASRUtils::extract_n_dims_from_ttype(intrinsic_array_function->m_type);
```

src/libasr/pass/simplifier.cpp
```diff
@@ -540,6 +540,23 @@ bool set_allocation_size(
...
+                    LCOMPILERS_ASSERT(intrinsic_array_function->n_args = 3);
+                    ASR::expr_t* ncopies = intrinsic_array_function->m_args[2];
+                    allocate_dims.reserve(al, n_dims + 1);
```
Pranavchiku, line 549:
```suggestion
                    allocate_dims.reserve(al, n_dims);
```

src/libasr/pass/simplifier.cpp
```diff
@@ -540,6 +540,23 @@ bool set_allocation_size(
                     }
                     break;
                 }
+                case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Spread): {
+                    size_t n_dims = ASRUtils::extract_n_dims_from_ttype(
+                        ASRUtils::expr_type(intrinsic_array_function->m_args[0])
+                    );
+                    LCOMPILERS_ASSERT(intrinsic_array_function->n_args = 3);
```
Pranavchiku, line 547: This will be ensure before creating intrinsic so I think we can remove this

src/libasr/pass/simplifier.cpp
```diff
@@ -540,6 +540,23 @@ bool set_allocation_size(
...
+                    ASR::expr_t* ncopies = intrinsic_array_function->m_args[2];
+                    allocate_dims.reserve(al, n_dims + 1);
+
+                    for (size_t i = 1; i <= n_dims + 1; i++) {
```
Pranavchiku, line 551:
```suggestion
                    for (size_t i = 1; i <= n_dims; i++) {
```

Pranavchiku (changes requested):
left minor comments

src/libasr/pass/simplifier.cpp
gxyd: Please refer to `spread` intrinsic documentation: https://gcc.gnu.org/onlinedocs/gfortran/SPREAD.html, which says:
```
*Result value*:
 The result is an array of the same type as SOURCE and has rank n+1 where n equals the rank of SOURCE.

```

src/libasr/pass/simplifier.cpp
gxyd: I don't like the suggested formatting, the current one is better to me.

src/libasr/pass/simplifier.cpp
gxyd: see: https://github.com/lfortran/lfortran/pull/5722/files#r1886428037

src/libasr/pass/simplifier.cpp
Pranavchiku: Yes yes, all this is already handled in `create_Spread`, so why repeat logic when we have computed stuff in `intrinsic_array_function->m_type`

src/libasr/pass/simplifier.cpp
Pranavchiku: refer: https://github.com/lfortran/lfortran/pull/5722/files#r1886515462

Pranavchiku (changes requested):
left comments above

src/libasr/pass/simplifier.cpp
gxyd: I see, yes, you are right about that.

src/libasr/pass/simplifier.cpp
gxyd: Yes, I've made the appropriate change here now.

src/libasr/pass/simplifier.cpp
gxyd: Done, I've reverted this as well.

gxyd:
Probably needs a review again now, and it should be hopefully good now @Pranavchiku

Pranavchiku (approved):
Perfect, thanks for doing this!

gxyd:
Nice, thanks for the review here @Pranavchiku, needed more comment that it should've but I'm happy that this looks good now.

Pranavchiku:
Haha, more precision from now ;)

PR #5730, opened 2024-12-16
CI: Revert "[CI] Fix the command to initialize SHELL for mamba"

gxyd:
## Description

The CI job *build-and-push-image* seems to be failing again, yesterday the fix was made here https://github.com/lfortran/lfortran/pull/5704 and merged.

with this PR I attempt to see if reverting that change fix it now or not. This reverts commit 2f3fe4c0a1cebbed4faca30100b790fdf51c1af9.

PR #5740, opened 2024-12-17
CI: fix version of miniforge in Dockerfile

gxyd:
## Description

This is an attempt to fix the version of miniforge to use and hence the version of mamba installed to make sure that we don't get CI failure of *build-and-push* CI job, like we got in https://github.com/lfortran/lfortran/pull/5730.

I'll first check to see if the tests pass or not

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,8 @@ USER root
 RUN apt update
 RUN apt install curl git build-essential binutils-dev zlib1g-dev clang libunwind-dev -y
 
-RUN curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
+# fix version of miniforge to 24.11.0-0 as that fixes the version of mamba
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,7 +5,8 @@ USER root
 RUN apt update
 RUN apt install curl git build-essential binutils-dev zlib1g-dev clang libunwind-dev -y
 
-RUN curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
+# fix version of miniforge to 24.11.0-0 as that fixes the version of mamba
+RUN curl -L -O "https://github.com/conda-forge/miniforge/releases/download/24.11.0-0/Miniforge3-$(uname)-$(uname -m).sh"
```
Pranavchiku, line 9: this is a good practice to fix version of external dependencies to avoid failures as observed in #5730

Pranavchiku (approved):

certik (approved):
Perfect!

certik:
Yes, the lesson is that one has to fix the version of absolutely everything, in order to minimize CI breakage, and then upgrade manually as needed.

PR #5745, opened 2024-12-17
build: add support for cached building using `ccache` through a CMake option

kmr-srbh:
Add support for build caching using `ccache`. 

We do not want to add `ccache` as a project dependency, so we do not build with it by default. The developer can install `ccache` on their own to do a cached build, with this PR we only enable them to do so.

See https://github.com/lfortran/lfortran/pull/5423#issuecomment-2489109206.


kmr-srbh:
See https://github.com/lfortran/lfortran/pull/5423#issuecomment-2489109206.

certik (approved):
I think that's fine.

PR #4183, opened 2024-06-08
Improve first build experience

serge-sans-paille:
When trying to build lfortran after a clone and a simple 

```
cmake .. -DLFORTRAN_BUILD_ALL=1 
```

I got a few errors (missing packages, version script not working) which should be fixed by these commits

PR #5747, opened 2024-12-17
Properly detect re2c and bison when building under LFORTRAN_BUILD_ALL

certik:
This was extracted from #4183.

I benchmarked the speed of compilation, didn't notice any slowdown.


Pranavchiku (approved):
Works fine for me as well without any noticeable slowdown

PR #5750, opened 2024-12-17
enh: improved LFortranJSON to cover edge cases

### LINKED ISSUE #5455: Improve `LFortranJSON` implementation
Refer https://github.com/lfortran/lfortran/pull/5438#discussion_r1854148516

captainhaddock18:
fixes #5455

### DIFF
--- a/src/libasr/lsp.cpp
+++ b/src/libasr/lsp.cpp
@@ -69,15 +69,15 @@ enum LFortranJSONType {
 };
 
 class LFortranJSON {
-
 private:
     LFortranJSONType type;
     std::string json_value;
     std::vector<std::pair<std::string, std::string>> object_members;
     std::vector<std::string> array_values;
+    bool rebuild_needed; 
 
 public:
-    LFortranJSON(LFortranJSONType type) : type(type) {
+    LFortranJSON(LFortranJSONType type) : type(type), rebuild_needed(true) {
         if (type == kArrayType) {
             json_value = "[]";
         } else {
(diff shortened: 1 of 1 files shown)


src/libasr/lsp.cpp
```diff
@@ -63,89 +69,75 @@ enum LFortranJSONType {
...
+        array_values.clear();
+        json_value = "[]";
+    }
+    void AddMember(std::string key, int v) {
+        object_members.push_back({key, std::to_string(v)});
+        RebuildObject();
+    }
+    void AddMember(std::string key, uint32_t v) {
+        object_members.push_back({key, std::to_string(v)});
+        RebuildObject();
```
Pranavchiku, line 103: Let's rebuild it only when user calls `GetValue`

certik:
I think it's fine to store things, but we must only build once at the end, not every time.

Also, does this work with arbitrary nesting of JSON?

captainhaddock18:
> I think it's fine to store things, but we must only build once at the end, not every time.
> 
> Also, does this work with arbitrary nesting of JSON?

I have revised the code to run build JSON only when the GetValue is called. Yeah, it works for arbitrary nesting of JSON.

for example 
`int main() {

    LFortranJSON level3(kObjectType);
    level3.AddMember("key3", "value3");

    LFortranJSON level2(kObjectType);
    level2.AddMember("key2", level3);

    LFortranJSON level1(kObjectType);
    level1.AddMember("key1", level2);

    LFortranJSON root(kObjectType);
    root.AddMember("root", level1);

    std::cout << root.GetValue() << std::endl;

    return 0;
}` 
 output: 
image

src/libasr/lsp.cpp
```diff
@@ -63,89 +69,83 @@ enum LFortranJSONType {
...
-            // If '}' is found, then add `, "key": value` before '}'
-            if (end_pos == -1) {
-                value.insert(value.length(), "\"" + key + "\":\"" + v + "\"}");
-            } else {
-                value.insert(end_pos, ",\"" + key + "\":\"" + v + "\"");
-            }
+    }
+    void SetObject() {
+        type = kObjectType;
+        object_members.clear();
```
Pranavchiku, line 89: We cannot clear it

src/libasr/lsp.cpp
```diff
@@ -63,89 +69,83 @@ enum LFortranJSONType {
...
+        json_value = "{}";
+        rebuild_needed = false;
+    }
+    void SetArray() {
+        type = kArrayType;
+        array_values.clear();
```
Pranavchiku, line 95: same comment here

certik:
@Pranavchiku I missed it, sorry about that. Hopefully it's not hard to fix.

Pranavchiku:
I think this is fine. We need to do this.

Issue #5709, opened 2024-12-15
call to FLUSH raises undefined intrinsic


gxyd:
## Minimal Reproducible Example

For the below Fortran program:
```f90
program POT3D
    use iso_fortran_env, only : OUTPUT_UNIT
    implicit none
    print *, OUTPUT_UNIT
    call FLUSH(OUTPUT_UNIT) ! I used 'call FLUSH(OUTPUT_UNIT)' instead of just 'FLUSH(OUTPUT_UNIT)'
end

```

on it's execution with LFortran, I get the below error:
```console
> lfortran --version
LFortran version: 0.42.0-526-g75e4b770b
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
semantic error: Function 'flush' not found (not user defined nor intrinsic)
  --> test.f90:92:5
   |
92 |     call FLUSH(OUTPUT_UNIT)
   |     ^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

if I change `call FLUSH(OUTPUT_UNIT)` to `FLUSH(OUTPUT_UNIT)` I get a different error:
```console
6
Specified UNIT 6 in FLUSH is not connected.

> echo $?
1
```

in `POT3D` code, I only see it's call as `call FLUSH(...)`, that's why reporting it here in that way.

gxyd:
I think my understanding of the issue was incorrect in terms of what *POT3D* needs. From the documentation of `FLUSH` from GFotran docs (ref: https://gcc.gnu.org/onlinedocs/gfortran/FLUSH.html), it says:
> Beginning with the Fortran 2003 standard, there is a FLUSH statement that should be preferred over the FLUSH intrinsic.

and a call like:
```f90
CALL FLUSH(9)
```

is actually a call to `FLUSH` intrinsic, while that doesn't work in LFortran, but using FLUSH statement does work well in LFortran, so we can replace `CALL FLUSH(...)` with `FLUSH` statements (i.e. `FLUSH(...)`), see commit: https://github.com/gxyd/POT3D/commit/fe44c91df5d96856b87dc5df92ce0818948155ea, which can act as a *better* workaround. By *better*, I mean, earlier I commented out the lines containing `CALL FLUSH`, which isn't the best thing.

gxyd:
[...]

PR #5751, opened 2024-12-17
CI: Test POT3D with LLVM backend with a few workarounds

gxyd:
## Description

There are two main workarounds which have been made use of in this:
* First is https://github.com/gxyd/POT3D/commit/4a85416220bfe18150b4a0d76d747b1bfbfd0ffa, the issue for this is already reported here: https://github.com/lfortran/lfortran/issues/5709, we'll try to fix this as well soon
* second is: https://github.com/gxyd/POT3D/commit/db7edd3a493ae4650c1942f257a60dd29958b6c1, uncommenting all calls to write subroutines for HDF5 format, we should be able to fix this well, I'll report an issue for this soon.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -515,12 +515,12 @@ jobs:
             git clone https://github.com/gxyd/pot3d.git
             FC="$(pwd)/src/bin/lfortran"
             cd pot3d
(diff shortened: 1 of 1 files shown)


certik (approved):
Great progress!

PR #5752, opened 2024-12-17
fix: asr generates with cc flag for program without line files

parth121101:
Towards: #5036 #5713 #5463


certik (approved):
I think that's fine, thanks!

PR #5753, opened 2024-12-18
enh: add support for automatic allocation for `allocatable` with `--std=f23`

kmr-srbh:
As seen in https://wg5-fortran.org/N2201-N2250/N2212.pdf.

'The new features of Fortran 2023' paper linked above talks about the introduction of automatic allocation of character variables but GFortran seems to support this for all variables.


Pranavchiku (approved):
I think it is fine. I'll wait for one more approval before merging this

certik (approved):
Thanks for adding it in.

PR #5754, opened 2024-12-18
enh : Avoid segfaults on strcpy with RHS-unallocated string

assem2002:
MRE
```Fortran
program string_43

    implicit none
    character(:), allocatable :: a
    character(:), allocatable :: b

    a  =  b

end program string_43
```


assem2002:
gfortran sometimes fails for this program so we won't test with gfortran.

integration_tests/string_43.f90
```diff
@@ -0,0 +1,13 @@
...
+
+    implicit none
+    character(:), allocatable :: a
+    character(:), allocatable :: b
+
+    a = "Hello"
+    a  =  b
+
+    print *, a
+    if(a /= "") error stop
```
Pranavchiku, line 11: My only concern is `""` is this always true?

kmr-srbh:
I think the above program is incorrect. For deferred length character variables, GFortran allocates the required length on assignment. In the provided MRE, `a` is assigned to `b`, but `b` points to uninitialized memory as it is unallocated, and so segfaults.

From https://wg5-fortran.org/N2201-N2250/N2212.pdf, section 2.2

> 2.2 US 14. Automatic allocation of lengths of character variables
When a deferred-length allocatable variable is defined by intrinsic assignment, as in the example
character(:), allocatable :: quotation
:
quotation = ’Now is the winter of our discontent.’
it is allocated by the processor to the correct length. This behaviour is extended to messages
returned through iomsg and errmsg specifiers, writing to a scalar character variable as an
internal file, and intent out and inout scalar character arguments of intrinsic procedures, such
as in call get command(command).

integration_tests/string_43.f90
kmr-srbh: This program works for me with GFortran, but `b` is still unitialized here.

kmr-srbh:
From ChatGPT:

```text
Assignment behavior:

In the first program, the assignment a = b tries to copy the contents of b into a. 
Since b is unallocated, this operation attempts to access an invalid memory location,
causing a segmentation fault.

In the second program, b is initially unallocated, but the assignment a = "Hello" ensures 
that a is valid and allocated before any operation involving b. By the time a = b is executed,
b remains unallocated, but the compiler or runtime system has a more predictable state, possibly
handling the unallocated b gracefully (e.g., treating it as an empty string, depending on the compiler
implementation). However, this behavior is not guaranteed by the Fortran standard.
```

integration_tests/string_43.f90
assem2002: actually your questions lead to admitting that this program should have unidentified behavior and shouldn't be used as a test on our CI (though using allocatable on the LHS instantly allocates memory for LHS -if NULL- no matter what is the RHS).

kmr-srbh: Yes, the memory is allocated for LHS on assignment, but as pointed by ChatGPT, GFortran might be doing some special handling for the uninitialized `b` when it is passed for assignment to a variable which has memory allocated to it. As it points, this is not correct behavior.

assem2002: I think maybe we call handle the unintended behavior this by nullifying the LHS if the RHS is `NULL`. It some how gonna be predictable thing to the user that "yeah, now I have an LHS that's not allocated like the RHS"

assem2002: > Yes, the memory is allocated for LHS on assignment according to the length of the RHS, but as pointed by ChatGPT, GFortran might be doing some special handling for the uninitialized `b` when it is passed for assignment to a variable which has memory allocated to it. As it points, this is not correct behavior.

on my machine we I run the executable of gfortran  multiple times it raises a segfaults.

kmr-srbh: Yes, we can give a runtime error if RHS is NULL in an assignment.

integration_tests/CMakeLists.txt
```diff
@@ -1363,6 +1363,7 @@ RUN(NAME string_39 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_40 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
 RUN(NAME string_41 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
 RUN(NAME string_42 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME string_43 LABELS llvm llvm_wasm llvm_wasm_emcc)
```
certik, line 1366: Add gfortran

integration_tests/string_43.f90
```diff
@@ -0,0 +1,13 @@
+program string_43
+
+    implicit none
+    character(:), allocatable :: a
+    character(:), allocatable :: b
+
+    a = "Hello"
+    a  =  b
```
certik, line 8: Remove this line

assem2002: Which one?

certik: The `a = b` line is incorrect, you can't assign unallocated string.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2058,8 +2058,13 @@ LFORTRAN_API void _lfortran_strcpy_descriptor_string(char** x, char *y, int64_t*
     ASSERT_MSG(((*x != NULL) && (*x_string_size <= (*x_string_capacity - 1))) ||
         (*x == NULL && *x_string_size == 0 && *x_string_capacity == 0) , "%s",
     "compiler-behavior error : string x_string_capacity < string size");
+    
     size_t y_len, x_len; 
-    y_len= strlen(y); 
+    if(y == NULL){
+        y_len = 0;
```
certik, line 2064: Shouldn't this give a runtime error?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2080,13 +2085,18 @@ LFORTRAN_API void _lfortran_strcpy_descriptor_string(char** x, char *y, int64_t*
 
 LFORTRAN_API void _lfortran_strcpy_pointer_string(char** x, char *y)
 {
-    size_t y_len = strlen(y);
+    size_t y_len;
+    if(y == NULL){
+        y_len = 0;
```
certik, line 2090: Runtime error?

integration_tests/string_43.f90
```diff
@@ -0,0 +1,11 @@
...
+
+    implicit none
+    character(:), allocatable :: a
+
+    a = "Hello"
+
+    print *, a
+    if(a /= "Hello") error stop
+
+end program string_43
```
certik, line 11: This test already works in master, so it is insufficient to test this PR.

certik (commented):
I think this should give a runtime error. What exact bug is this fixing? The test already works in master.

integration_tests/string_43.f90
assem2002: I removed this test, and used reference test instead.

src/libasr/runtime/lfortran_intrinsics.c
assem2002: done here https://github.com/lfortran/lfortran/pull/5754/commits/3f21f89b20717648eaa80b8a2e17802d915ce8b8

src/libasr/runtime/lfortran_intrinsics.c
assem2002: done here https://github.com/lfortran/lfortran/pull/5754/commits/3f21f89b20717648eaa80b8a2e17802d915ce8b8

certik (approved):
I think this looks great! Thanks.

PR #5756, opened 2024-12-18
Set of fixes for compiling PRIMA with `--experimental-simplifier`

czgdp1807:
PRIMA still doesn't compile with `--experimental-simplifier` but I have fixed a bunch of issues. Will send more tomorrow. We are close. Only PRIMA is left.


src/libasr/pass/simplifier.cpp
```diff
@@ -13,6 +13,34 @@
...
+        case ASR::exprType::ComplexUnaryMinus:
+        case ASR::exprType::Var: {
+            return true;
+        }
+        default: {
+            return false;
+        }
+    }
+}
+
```
czgdp1807, line 43: We will need this function in future inside `create_and_allocate_temporary_variable`. Will help us in creating temporaries only for expressions where are not vectorise-able. There is one function below, `is_temporary_needed`, where we should use this. Not going to do it now but later after compiling PRIMA.

czgdp1807:
I am merging this PR. All the changes are tested so nothing to worry about.

PR #5762, opened 2024-12-19
remove : avoid repeated checks in convert_call_args

jinangshah21:


Pranavchiku (commented):
Does this PR fix any issue?
Does these checks harm us someway? If no, I think we can keep these as is.

jinangshah21:
This doesn't fix any issue. It just avoid multiple checks for same thing.

certik (approved):
I think this looks good, removing code is always good.

PR #5764, opened 2024-12-19
fix: decrement loop variable to retain correct value when using loop variable after loop

### LINKED ISSUE #5761: Bad loop variable
Lfortran 0.42.0 gives wrong results from this program with every option tried. The issue arose from a comment by @certik in #5739. The loop variable ```i``` has the wrong value after one of the loops; which loop it is depends on an option.
```
  integer i
  do i = 1,2
     print "(A,I0)",'In loop 1 i=',i
  end do
  print "(A,I0,A)",'After loop 1 i=',i,merge(' good',' bad ',i==3)
  do i = 1,2
     print "(A,I0)",'In loop 2 i=',i
     if (i<2) exit
  end do
  print "(A,I0,A)",'After loop 2 i=',i,merge(' good',' bad ',i==1)
  end program
```
Output with no options or ```--std=f23```
```
In loop 1 i=1
In loop 1 i=2
After loop 1 i=2 bad 
In loop 2 i=1
After loop 2 i=1 good
```
Output with ```--use-loop-variable-after-loop``` with or without ```--std=f23```
```
In loop 1 i=1
In loop 1 i=2
After loop 1 i=3 good
In loop 2 i=1
After loop 2 i=2 bad 
```
Output (correct IMHO) from 5 other compilers: gfortran, g95, ifx, ifort, AMD flang
```
In loop 1 i=1
In loop 1 i=2
After loop 1 i=3 good
In loop 2 i=1
After loop 2 i=1 good
```

kmr-srbh:
fixes #5761 

When breaking from the loop, we decrement the loop variable by the loop increment value, so when we break out of loop, the default increment by `--use-loop-variable-after-loop` fixes the value.

### DIFF
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -4461,3 +4461,8 @@ show_errors = true
 filename = "document_symbols1.f90"
 document_symbols = true
 extrafiles = "document_symbols1_module.f90"
+
+[[test]]
+filename = "../integration_tests/loop_var_use_after_loop.f90"
+pass = "do_loops"
+options = "--use-loop-variable-after-loop"
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1493,6 +1493,8 @@ RUN(NAME div_to_mul LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME fma LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME loop_unroll_small LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME loop_unroll_large LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME loop_var_use_after_loop LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm
+        EXTRA_ARGS --use-loop-variable-after-loop)
 RUN(NAME sign_from_value LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 
 RUN(NAME rewind_inquire_flush LABELS gfortran)
--- a/integration_tests/loop_var_use_after_loop.f90
+++ b/integration_tests/loop_var_use_after_loop.f90
@@ -1,0 +1,58 @@
+program main
+   implicit none
+
+   integer :: i
+
+   ! Loop 1
+   do i = 1, 2
+      print *,'In loop 1 i =', i
+      if (i < 2) exit
+   end do
+   print *,'After loop 1 i = ', i
+   if (i /= 1) error stop
+
+   ! Loop 2
+   print *
+   do i = 1, 3
+      print *,'In loop 2 i = ', i
+      if (i < 2) then
+         print *, "if (i < 2) then"
+      else
+         exit
+      end if
+   end do
+   print *,'After loop 2 i = ', i
+   if (i /= 2) error stop
+
+
+   ! Loop 3
+   print *
+   do i = 1, 6
+      print *,'In loop 3 i = ', i
+      if (i <= 4) then
+         if (i - 1 == 3) then
+            if (i - 1 == 2) exit
+            exit
+         end if
+      end if
+   end do
+   print *,'After loop 3 i = ', i
+   if (i /= 4) error stop
+
+   ! Loop 4
+   print *
+   do i = 1, 6
+      print *,'In loop 4 i = ', i
+      if (i <= 4) then
+         if (i - 1 == 3) then
+            if (i - 1 == 2) exit
+            exit
+         else
+            print *, "i - 1 /= 3"
+         end if
+      end if
+   end do
+   print *,'After loop 4 i = ', i
+   if (i /= 4) error stop
+
+end program
--- a/src/libasr/pass/pass_utils.cpp
+++ b/src/libasr/pass/pass_utils.cpp
@@ -1121,6 +1121,101 @@ namespace LCompilers {
             }
         }
 
+        Vec<ASR::stmt_t*> insert_if_stmts_in_loop_body(Allocator& al,
+                                                       ASR::If_t* if_stmt,
+                                                       ASR::stmt_t* decrement_stmt)
+        {
+            Vec<ASR::stmt_t*> body; body.reserve(al, 0);
+            Vec<ASR::stmt_t*> if_stmt_body; if_stmt_body.reserve(al, 0);
+            Vec<ASR::stmt_t*> else_stmt_body; else_stmt_body.reserve(al, 0);
+
+            for (size_t i = 0; i < if_stmt->n_body; i++) {
+                if (ASR::is_a<ASR::If_t>(*if_stmt->m_body[i])) {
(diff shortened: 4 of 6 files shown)


kmr-srbh:
The code:

```fortran
program main
   implicit none

   integer :: i

   do i = 1, 6
      print *,'In loop i = ', i
      if (i <= 4) then
         if (i - 1 == 3) then
            if (i - 1 == 2) exit
            exit
         end if
      end if
   end do
   print *,'After loop i = ', i
end program
```

is transformed to:

```fortran
! Fortran code after applying the pass: do_loops
program main
implicit none
integer(4) :: i
i = 1 - 1
do while (i + 1 <= 6)
    i = i + 1
    print *, "In loop i = ", i
    if (i <= 4) then
        if (i - 1 == 3) then
            if (i - 1 == 2) then
                i = i - 1 ! i decremented to make up for increment later
                exit
            end if
            i = i - 1 ! i decremented to make up for increment later
            exit
        end if
    end if
end do
i = i + 1
print *, "After loop i = ", i
end program main
```

kmr-srbh:
There is a minor issue, where we are skipping some statements in the `else` block. Will push a fix in a while.

tests/tests.toml
```diff
@@ -4444,3 +4444,8 @@ fortran = true
 [[test]]
 filename = "polymorphic_class_in_derived_type.f90"
 asr = true
+
+[[test]]
+filename = "../integration_tests/loop_var_use_after_loop.f90"
+asr = true
+options = "--use-loop-variable-after-loop"
```
kmr-srbh, line 4451: Need to push the ASR after the `do_loop` pass actually.  Will make the changes tomorrow morning (IST).

certik:
The approach is correct, the default mode is not affected. But it does have one issue: it will slow down the loop execution (I think) in "standard mode".

I think there is a way to implement the loops so that the variable holds the value of the last iteration even after exiting the loop, and doing it without too many extra checks. But we can try to figure it out later, I think at least getting the right answer first is more important.

kmr-srbh:
While thinking of a fix, I had 2 ideas. One of them was the implementation in this PR, where we decrement at every possible point of breaking from the loop, which makes up for the increment outside. The other one was as you pointed above, where we just want to retain the value of the loop variable after exit, but this would require runtime knowledge of the control flow, as in conditional statements like `if - else`, we do not know which path the program execution would take. Also, we do not want to remove the increment outside the loop as it handles a normal loop completion.

Could you please elaborate the idea you are suggesting?

certik:
I would say for now fix it the way you did in this PR, so at least we have the correct behavior and a test.

To figure out the right answer, the issue is that we need to generate such an LLVM code so that it gets properly vectorized and gets the highest performance. It could be that maybe we will do something for the most common case of compile-time increment 1, then something slightly more complex for runtime increment, and something even more complex if the variable has to have the right value after the loop.

kmr-srbh:
Okay, I will try to move ahead with the implementation in this PR. I’m not very familiar with the concept of vectorization, so I wasn’t able to consider a solution along those lines. The topic sounds intriguing, and I think it would make for a good pastime read. :)

certik:
We just have to generate code that LLVM can optimize, so as simple as possible, no if statements (or as little as possible) inside the loop, etc.

kmr-srbh:
Here is the MRE for the SciPy failure:

```fortran
program main

contains

   SUBROUTINE CCHG

      REAL(8) :: A, CFAC
      INTEGER :: N
      DO N=0, 10
         IF (.true.) THEN
            CALL CGAMA
            IF (.true.) CFAC=DCOS(A)
         END IF
      END DO
   END SUBROUTINE CCHG


   SUBROUTINE CGAMA
   END SUBROUTINE CGAMA

end program main

```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran -c --generate-object-code --use-loop-variable-after-loop test.f90
ASR verify pass error: Function cchg doesn't depend on cgama but is found in its dependency list.
  Binary file "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2732, in main
    std::vector<LCompilers::StacktraceItem> d = e.stacktrace_addresses();
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2625, in main_app(int, char**)
    std::cerr << "The -c option requires the LLVM backend to be enabled. Recompile with `WITH_LLVM=yes`." << std::endl;
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 1049, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
  File "/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/fortran_evaluator.cpp", line 382, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    compiler_options, run_fn, infile);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 10445, in LCompilers::asr_to_llvm(LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LCompilers::PassManager&, LCompilers::CompilerOptions&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    skip_optimization_func_instantiation.push_back(static_cast<int64_t>(
  File "/home/saurabh-kumar/Projects/System/lfortran/src/libasr/pass/pass_manager.h", line 399, in LCompilers::PassManager::apply_passes(Allocator&, LCompilers::ASR::TranslationUnit_t*, LCompilers::PassOptions&, LCompilers::diag::Diagnostics&)
    apply_passes(al, asr, _passes, pass_options, diagnostics);
LCompilersException: Verify failed in the pass: do_loops
```

https://github.com/lfortran/lfortran/pull/5764#issuecomment-2554638483 and this failure is the same. It is good that this is a minor issue where we are only missing statements. The SciPy failure gave no hint towards this.

kmr-srbh:
Ready.

tests/tests.toml
kmr-srbh: Done.

certik (approved):
I think this fix looks good to me: it doesn't affect the default path, and it fixes the `--use-loop-variable-after-loop` path.

I am a little worried the new code is not tested enough --- we probably need to run all integration tests with `--std=f23` as well, to ensure everything actually works both ways. We can do that later (https://github.com/lfortran/lfortran/issues/5815), for now this is good.

kmr-srbh:
Thank you for reviewing this!

PR #5767, opened 2024-12-20
enh: `--show-errors` to throw multiple errors for LSP to pick

Pranavchiku:
Fixes https://github.com/lfortran/lfortran-lsp/issues/22


tests/reference/llvm-arrays_06-538e67d.stdout
```diff
@@ -1639,9 +1639,9 @@ declare i8* @_lcompilers_string_format_fortran(i32, i8*, ...)
...
 declare void @_lfortran_printf(i8*, ...)
 
-; Function Attrs: nounwind readnone speculatable willreturn
+; Function Attrs: nocallback nofree nosync nounwind speculatable willreturn memory(none)
 declare float @llvm.pow.f32(float, float) #0
 
 declare void @_lpython_free_argv()
 
-attributes #0 = { nounwind readnone speculatable willreturn }
+attributes #0 = { nocallback nofree nosync nounwind speculatable willreturn memory(none) }
```
certik, line 1647: These changes seem unrelated?

tests/reference/llvm-arrays_06-538e67d.stdout
```diff
@@ -1639,9 +1639,9 @@ declare i8* @_lcompilers_string_format_fortran(i32, i8*, ...)
 
 declare void @_lfortran_printf(i8*, ...)
 
-; Function Attrs: nounwind readnone speculatable willreturn
+; Function Attrs: nocallback nofree nosync nounwind speculatable willreturn memory(none)
```
Pranavchiku, line 1642: These are weird, I am not sure why are these present.

tests/reference/asr-show_errors1-cc473c3.stdout
```diff
@@ -0,0 +1 @@
+{"uri":"uri","diagnostics":[{"source":"lpyth","range":{"start":{"line":8,"character":5},"end":{"line":8,"character":5}},"message":"Newline is unexpected here","severity":1},{"source":"lpyth","range":{"start":{"line":11,"character":10},"end":{"line":11,"character":11}},"message":"Variable 'xx' is not declared","severity":1},{"source":"lpyth","range":{"start":{"line":12,"character":10},"end":{"line":12,"character":11}},"message":"Variable 'xz' is not declared","severity":1},{"source":"lpyth","range":{"start":{"line":13,"character":10},"end":{"line":13,"character":18}},"message":"Variable 'fasldjkfh' is not declared","severity":1}]}
```
certik, line 1: This is unrelated, but instead of `"source":"lpyth"` we need to put in the correct filename, in this case it would be `tests/errors/show_errors1.f90`. We can fix that later.

certik (approved):
Thanks for fixing it, looks good!

Check the LLVM reference tests changes in this PR, it seems a mistake.

tests/reference/llvm-arrays_06-538e67d.stdout
certik: It failed as it should, so that's good.

Likely you are using a different LLVM version than 11.

Pranavchiku: > These changes seem unrelated?

Yes, it is present in main as well. I checked with clean build.

Pranavchiku:
```console
% llvm-config --version

16.0.0
```

certik: Yes, you can't use any other LLVM than 11. We'll eventually upgrade, but not right now.

PR #5768, opened 2024-12-20
ci: enable LLVM19 for PRIMA

Pranavchiku:
This PR checks if PRIMA works with `llvm19` or not.


certik:
I am guessing it will fail. If so, let's create an MRE, fix it and add it to our integration tests.

Pranavchiku:
Yes, it failed.

Issue #5735, opened 2024-12-16
bug: self-referencing procedure pointers not supported


kmr-srbh:
### Minimal Reproducible Example

```fortran
module main_module

contains

   recursive subroutine sub(int)
      integer, intent(in) :: int
      procedure(sub), pointer :: proc => sub

      if (int > 10) return

      if (mod(int, 2) /= 0) then
         print *, "Odd: ", int
         call proc(int + 1)
      else
         proc => alt_sub
         call proc(int)
      end if
   end subroutine sub

   subroutine alt_sub(int)
      integer, intent(in) :: int
      print *, "Even: ", int
      call sub(int + 1)
   end subroutine alt_sub

end module main_module


program main
   use main_module
   call sub(1)
end program main
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
 Odd:            1
 Even:            2
 Odd:            3
 Even:            4
 Odd:            5
 Even:            6
 Odd:            7
 Even:            8
 Odd:            9
 Even:           10
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
semantic error: Procedure type 'sub' not declared
 --> test.f90:7:7
  |
7 |       procedure(sub), pointer :: proc => sub
  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #5770, opened 2024-12-20
fix: visit procedure declaration after visiting func/subrout in `ast_symboltable_visitor`

### LINKED ISSUE #5757: bug: defining procedure pointer with `subroutine` not allowed

```fortran
program main

   procedure(say_something), pointer :: proc
   proc => say_something

   call proc("Hello from proc!")

contains

   subroutine say_something(message)
      character(len=*),intent(in) :: message
      print *, message
   end subroutine

end program main
```

### Output

```console
/usr/bin/ld: warning: /tmp/ccltcx46.o: requires executable stack (because the .note.GNU-stack section is executable)
 Hello from proc!
semantic error: Procedure type 'say_something' not declared
 --> test.f90:3:4
  |
3 |    procedure(say_something), pointer :: proc
  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

jinangshah21:
Fixes #5757

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -560,6 +560,7 @@ public:
         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
         current_module_dependencies.reserve(al, 4);
+        Vec<size_t> procedure_decl_indices; procedure_decl_indices.reserve(al, 0);
         if (compiler_options.implicit_typing) {
             Location a_loc = x.base.base.loc;
             populate_implicit_dictionary(a_loc, implicit_dictionary);
@@ -586,6 +587,29 @@ public:
             }
         }
         for (size_t i=0; i<x.n_decl; i++) {
+            if(AST::is_a<AST::Declaration_t>(*x.m_decl[i])) {
+                AST::Declaration_t* decl = AST::down_cast<AST::Declaration_t>(x.m_decl[i]);
+                if(decl->m_vartype) {
+                    AST::AttrType_t* type = AST::down_cast<AST::AttrType_t>(decl->m_vartype);
+                    if(type && type->m_type == AST::decl_typeType::TypeProcedure) {
+                        procedure_decl_indices.push_back(al, i);
+                        continue;
+                    }
+                }
+            }
+            try {
+                visit_unit_decl2(*x.m_decl[i]);
+            } catch (SemanticAbort &e) {
+                if ( !compiler_options.continue_compilation ) throw e;
+            }
+        }
+        for (size_t i=0; i<x.n_contains; i++) {
+            bool current_storage_save = default_storage_save;
+            default_storage_save = false;
+            visit_program_unit(*x.m_contains[i]);
+            default_storage_save = current_storage_save;
+        }
+        for (size_t i : procedure_decl_indices) {
             try {
                 visit_unit_decl2(*x.m_decl[i]);
             } catch (SemanticAbort &e) {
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -586,19 +586,43 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                    if(type && type->m_type == AST::decl_typeType::TypeProcedure) {
+                        try {
+                            visit_unit_decl2(*x.m_decl[i]);
+                        } catch (SemanticAbort &e) {
+                            if ( !compiler_options.continue_compilation ) throw e;
+                        }
+                    }
+                }
+            }
+        }
```
kmr-srbh, line 624: We have some redundancy here, also let's not loop over all declarations again. Can we add declarations with `AST::decl_typeType::TypeProcedure` to a vector `Vec<AST::unit_decl2_t>`, loop over them and visit?

jinangshah21: Thanks ! I will do it.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -586,19 +587,36 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
-        process_simd_variables();
         for (size_t i=0; i<x.n_contains; i++) {
             bool current_storage_save = default_storage_save;
             default_storage_save = false;
             visit_program_unit(*x.m_contains[i]);
             default_storage_save = current_storage_save;
         }
+        for (size_t i=0; i<procedure_decl_indices.size(); i++) {
+            try {
+                visit_unit_decl2(*x.m_decl[i]);
```
kmr-srbh, line 614:
```suggestion
        for (size_t i : procedure_decl_indices) {
            try {
                visit_unit_decl2(*x.m_decl[i]);
```

jinangshah21: Oh yes! Sorry.

kmr-srbh (approved):
Thanks for this! The changes in the PR look good.

We need to handle this for other language elements that support declarations. Please see `AST.asdl` for nodes with `unit_decl2*`. `AST::Module_t` can be handled in the same way as this PR. 

For others however, like `AST::Subroutine_t`, `AST::Function_t` we probably would need to extend the logic or do something different, as these can have self-referencing procedures. We have an issue open for this at #5735. Another case is `AST::DerivedType_t`, where we declare the procedure in the scope of the derived type, but have the subroutine/function declared outside. I think we can open a separate issue for this.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -586,19 +587,36 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             }
         }
         for (size_t i=0; i<x.n_decl; i++) {
+            if(x.m_decl[i]->type == AST::unit_decl2Type::Declaration) {
+                AST::Declaration_t decl = (const AST::Declaration_t &)*x.m_decl[i];
```
certik, line 591: Use `AST::down_cast`

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -586,19 +587,36 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                if(decl.m_vartype) {
+                    AST::AttrType_t* type = AST::down_cast<AST::AttrType_t>(decl.m_vartype);
+                    if(type && type->m_type == AST::decl_typeType::TypeProcedure) {
```
certik, line 594: Use `AST::is_a`

certik (approved):
Otherwise I think it's fine.

src/lfortran/semantics/ast_symboltable_visitor.cpp
jinangshah21: We can't use `is_a` here because we don't have specific node for `TypeProcedure`.

certik (approved):

certik (approved):

Issue #4649, opened 2024-08-14
reshape with shape as reshape function call segfaults


gxyd:
### Minimal Reproducible Example

```fortran
program main
    implicit none

    complex(8) :: my_complex(2, 3)
    my_complex = reshape([(1, 2), (3, 4), (5, 6), (7, 8), (9, 10), (11, 12)], reshape([2, 3], [2]))
    print *, my_complex
end program
```

The above fortran program causes segfault with LFortran.

While the below program, where a variable is assigned for the shape (i.e. `reshape` function call), it works well, i.e.:
```fortran
program main
    implicit none
    complex(8) :: my_complex(2, 3)
    integer(4) :: shape(2) = reshape([2, 3], [2])
    my_complex = reshape([(1, 2), (3, 4), (5, 6), (7, 8), (9, 10), (11, 12)], shape)
    print *, my_complex
end program

```

the below program produces the correct output:
```bash
(1.000000,2.000000) (5.000000,6.000000) (9.000000,10.000000)
(3.000000,4.000000) (7.000000,8.000000) (11.000000,12.000000)
```

gxyd:
This issue seems to have been fixed in the latest *main*:
```console
> lfortran --version
LFortran version: 0.42.0-459-g9b2bcaccb-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
1.00000000000000000e+00    2.00000000000000000e+00    3.00000000000000000e+00    4.00000000000000000e+00    1.00000000000000000e+00    2.00000000000000000e+00    3.00000000000000000e+00    4.00000000000000000e+00
```

though the output *printed* isn't correct because of some issue with printing (that's a separate issue).

I think we just need to add a test case here, and we should be good to close this issue.

dsaw:
i can take a shot at this one if no one is working on it yet.

dsaw:
i can take this up, if no one has done this yet.

gxyd:
Yes, you can send in a PR.

Jaysukh-409:
With lastest main also above code is giving error.

```
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/lfortran$ lfortran --version
LFortran version: 0.43.0-98-ge0a1a30f9
Platform: Linux
LLVM: 11.0.1
Default target: x86_64-unknown-linux-gnu
[...]

PR #5780, opened 2024-12-21
Test for  reshape call as a parameter inside reshape function

dsaw:
#4649


tests/tests.toml
```diff
@@ -1614,6 +1614,11 @@ filename = "builtin1.f90"
 ast = true
 ast_f90 = true
 
+[[test]]
+filename = "reshape_param_shape.f90"
+run = true
+
+
```
kmr-srbh, line 1621: We do not need a reference test for this case. Please add it as an integration test by registering it inside `integration_tests/CMakeLists.txt`. Please delete the generated reference too.

tests/reshape_param_shape.f90
```diff
@@ -0,0 +1,8 @@
+! Issue 4649: reshape called in shape parameter of reshape function 
+program main
```
kmr-srbh, line 2: Rename the program to something else, like `reshape_param_shape`.

tests/reshape_param_shape.f90
```diff
@@ -0,0 +1,8 @@
...
+    implicit none
+
+    complex(8) :: my_complex(2, 3)
+    my_complex = reshape([(1, 2), (3, 4), (5, 6), (7, 8), (9, 10), (11, 12)], reshape([2, 3], [2]))
+    print *, my_complex
```
kmr-srbh, line 7: Please add an `if (any(... /= ...)) error stop` block for this. Currently, we do not support direct comparison with output of `reshape`, so you will need to do it like `if (any(my_complex /= my_reshape)) error stop`, where `my_reshape` contains the value which you get from `reshape` on `stdout` output by GFortran.

tests/reshape_param_shape.f90
dsaw: i might have misunderstood, `reshape_param_shape` is the same name as the file?

tests/tests.toml
```diff
@@ -1614,6 +1614,7 @@ filename = "builtin1.f90"
 ast = true
 ast_f90 = true
 
+
```
certik, line 1617: remove white space change.

PR #5797, opened 2024-12-23
fix: Update real function to provide correct kind

### LINKED ISSUE #4629: `real` does not give the correct kind for output
```fortran
program main
    complex(8) :: arr2(3) = [(5._8, 7._8), (8.2_8, 7.12_8), (1.2_8, 3.4_8)]
    print *, kind(real(arr2))
end program 
```
```console
$ lfortran a.f90
4
$ gfortran a.f90 && ./a.out
           8
```

HarshilShah1804:
Fix #4629 
The real function is updated so that if a kind is specified, the output is of that kind, otherwise the same kind as the input. 

For
```
program main
    complex(8) :: arr2(3) = [(5._8, 7._8), (8.2_8, 7.12_8), (1.2_8, 3.4_8)]
    print *, kind(real(arr2))
end program 
```

Output
```
$ lfortran a.f90
8
```


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -850,13 +850,6 @@
             "return": "int32",
         },
     ],
-    "Real": [
-        {
-            "args": [("int",), ("real",), ("complex",)],
-            "return": "real32",
```
certik, line 856: I wonder if we can do something like `"ret_type_arg_idx": 0` but do it for "kind".

@HarshitaKalani ?

certik:
Also add a test for this.

src/libasr/intrinsic_func_registry_util_gen.py
HarshilShah1804: @certik Sir, implementing this change would affect the output kind for all data types: real, integer, and complex. However, I noticed that in gfortran, the output kind remains the same as the input only for complex data types, while for real and integer types, it defaults to 4 unless explicitly specified.

certik: I see. Let's also check what the standard says here --- I know that gfortran behaves as above, but is this the right behavior?

certik: I checked godbolt and all other compilers (ifx, gfortran, flang) return 8 here, so we should as well, as you did.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -3936,6 +3947,66 @@ namespace Real {
...
+        if (all_args_evaluated(m_args)) {
+            Vec<ASR::expr_t*> args_values; args_values.reserve(al, 1);
+            args_values.push_back(al, expr_value(m_args[0]));
+            m_value = eval_Real(al, loc, return_type, args_values, diag);
+            if (diag.has_error()) {
+                return nullptr;
+            }
+        }
+        return ASR::make_IntrinsicElementalFunction_t(al, loc, static_cast<int64_t>(IntrinsicElementalFunctions::Real), m_args.p, m_args.n, 0, return_type, m_value);
+    }
```
certik, line 4008: How did you create these two functions --- you copied the autogenerated functions and just inserted a simple modification for complex?

Is there a way to do this automatically with our script? It might be a simple change.

src/libasr/intrinsic_func_registry_util_gen.py
certik: Can we do maybe `real32_except_complex` and add special handling for complex there? It might be a simple change. I don't like that we have to now maintain the complex logic manually of "verify" and "create".

src/libasr/pass/intrinsic_functions.h
HarshilShah1804: Yes, I copied them from the autogenerated functions and inserted the modification for complex.

certik:
Per https://github.com/lfortran/lfortran/pull/5797#discussion_r1896906495, see if you can modify the generator script to insert the modification for complex correctly, using the existing mechanism. That way we don't need to duplicate / manually maintain the two functions.

src/libasr/intrinsic_func_registry_util_gen.py
HarshitaKalani: `real32_except_complex` - Yes, I think this would be better. We should be avoiding the manual logic. @HarshilShah1804 can you try it in the `src/libasr/intrinsic_func_registry_util_gen.py` file only?

HarshitaKalani:
@HarshilShah1804 just a heads up. If you check `aimag` intrinsic, it already follows what we're trying to do here:
```fortran
program main
    complex(8) :: arr2(3) = [(5._8, 7._8), (8.2_8, 7.12_8), (1.2_8, 3.4_8)]
    print *, kind(real(arr2))
end program 
```
```console
$ lfortran a.f90
8
$ gfortran a.f90 && ./a.out
           8
```
You can follow its implementation and it'll be a simple fix.

src/libasr/intrinsic_func_registry_util_gen.py
HarshilShah1804: I have update the `src/libasr/intrinsic_func_registry_util_gen.py` file as told by you. Please check.

integration_tests/intrinsics_334.f90
```diff
@@ -0,0 +1,33 @@
+
+program intrinsics_332
```
HarshitaKalani, line 2:
```suggestion
program intrinsics_334
```

integration_tests/intrinsics_334.f90
```diff
@@ -0,0 +1,33 @@
...
+    if (kind(real(r1)) /= 4) error stop
+    print *, kind(real(r2))
+    if (kind(real(r2)) /= 4) error stop
+
+    
+    print *, kind(real(c1))
+    if (kind(real(c1)) /= 4) error stop
+    print *, kind(real(c2))
+    if (kind(real(c2)) /= 8) error stop
+end program
```
HarshitaKalani, line 33: @HarshilShah1804 I think we're not checking the test case which is mentioned in the issue. All the cases mentioned here already work fine. Can you please add a test for the issue which you just solved. That'll be great.

src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -852,10 +852,11 @@
     ],
     "Real": [
         {
-            "args": [("int",), ("real",), ("complex",)],
+            "args": [("real",), ("int",), ("complex",)],
```
HarshitaKalani, line 855: Why this change? If it was not intentional, can you please revert it?

HarshitaKalani (commented):
I've left some comments. Can you please address them?
This looks good otherwise. Thanks for fixing it!

src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -1067,6 +1068,15 @@ def add_create_func_return_src(func_name):
...
         src += indent * 2 + "}\n"
+    real_32_except_complex = arg_infos[0].get("real_32_except_complex", False)
+    if real_32_except_complex:
+        src += indent * 2 + "else { \n"
+        src += indent * 3 + "ASR::ttype_t* arg_type = ASRUtils::expr_type(args[0]);\n"
+        src += indent * 3 + "if (is_complex(*arg_type)) { \n"
+        src += indent * 4 + "int kind = ASRUtils::extract_kind_from_ttype_t(arg_type); \n"
+        src += indent * 4 + "set_kind_to_ttype_t(return_type, kind); \n"
+        src += indent * 3 + "} \n"
+        src += indent * 2 + "} \n"
```
certik, line 1079: Perfect. If this is the only change that is needed to fix this, then this is clearly easier to maintain than the two full functions in the previous version of the PR.

src/libasr/intrinsic_func_registry_util_gen.py
HarshilShah1804: Done

integration_tests/intrinsics_334.f90
HarshilShah1804: Changed.

integration_tests/intrinsics_334.f90
HarshilShah1804: Added the test

HarshitaKalani (approved):
Looks good to me. Thanks!

Issue #5728, opened 2024-12-16
bug: `StructInstanceMember` not supported for `newunit` argument to `open` statement


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   type :: file_t
      integer :: iunit = 0
   end type file_t

   type(file_t) :: f
   open(newunit=f%iunit, file='test.txt')
end program main
```

> test.txt

```
Hello, world!
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
semantic error: `newunit`/`unit` must be a variable of type, Integer or IntegerPointer
 --> test.f90:7:4
  |
7 |    open(newunit=f%iunit, file='test.txt')
  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #5803, opened 2024-12-23
Fix : use any variable as unit, not just `ASR::Var` + support different kinds integers to be used as unit arg.

assem2002:
fixes https://github.com/lfortran/lfortran/issues/5728

### DIFF
--- a/src/runtime/custom/lfortran_intrinsic_custom.f90
+++ b/src/runtime/custom/lfortran_intrinsic_custom.f90
@@ -1,24 +1,55 @@
 module lfortran_intrinsic_custom
 implicit none
 
-contains
+interface  newunit
+    procedure :: newunit_int_1
+    procedure :: newunit_int_2
+    procedure :: newunit_int_4
+    procedure :: newunit_int_8
+end interface 
 
-    subroutine newunit(unit)
-        implicit none
+contains
+    function get_valid_newunit() result(lun)
         integer, parameter :: LUN_MIN=0, LUN_MAX=1000
         logical :: opened
-        integer :: lun
-        integer, intent(out) :: unit
-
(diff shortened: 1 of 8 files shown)


assem2002:
The addressed issue is resolved only with this commit https://github.com/lfortran/lfortran/pull/5803/commits/0d8f729d9227e2b9626a6f77962045f568fa6e30. I just noticed that we can't support other integer kinds but 4 as unit, so the subsequent commits fix that.

src/runtime/custom/lfortran_intrinsic_custom.f90
```diff
@@ -1,24 +1,55 @@
...
         integer, parameter :: LUN_MIN=0, LUN_MAX=1000
         logical :: opened
-        integer :: lun
-        integer, intent(out) :: unit
-
+        integer(4) :: lun ! no need for kind 8, we have limit of 1000 for now.
         do lun=LUN_MIN,LUN_MAX
             inquire(unit=lun,opened=opened)
             if (.not. opened) then
-                unit = lun
```
certik, line 16: This is problematic: it is using a loop variable as a return value. I would do the original, `unit = lun`, that seems clearer. Loop variables should not be used outside of the loop, and the above usage seems confusing to me.

certik (approved):
I think this is good. Left a minor comment that you can fix in a subsequent PR.

src/runtime/custom/lfortran_intrinsic_custom.f90
assem2002: done here https://github.com/lfortran/lfortran/pull/5803/commits/a1f7b3a839defa4eec8df756f1bc1baa302e5e1b

certik (approved):

Issue #5813, opened 2024-12-24
Fortran code snippets that do not generate ASR and AST with cc flag


parth121101:
```
program circular_reference
    type :: node
      type(node), pointer :: next  ! Circular reference without allocation
      integer :: value
    end type node
  
    type(node) :: head
    head%value = 1
    print *, head%value
  end program circular_reference
  ```
  
  
  ```
   program invalid_select_type
    class(*), pointer :: obj
    integer :: x
    obj => x
    select type (obj)
    type is (integer)
      print *, "This is an integer"
    type is (real)  ! Error: A variable cannot have multiple types
      print *, "This is a real"
    end select
  end program invalid_select_type
  ```
  
  
  ```
  program unterminated_string
    print *, "Hello, world  ! Error: Missing closing quote
  end program unterminated_string
```

```
  program my_program
  print *, "Hello, World!"
end program another_name  ! Error: The name "another_name" does not match "my_program"
```

```
subroutine my_subroutine()
    print *, "Inside subroutine"
  end subroutine different_name  ! Error: "different_name" does not match "my_subroutine"
```
  

```
  function my_function() result(res)
    integer :: res
    res = 42
  end function not_my_function  ! Error: "not_my_function" does not match "my_function"
```

```
  module my_module
    integer :: x = 10
  end module wrong_module_name  ! Error: "wrong_module_name" does not match "my_module"
```

```
  program incorrect_pragma
    integer :: i
    !LF$unroll 4  ! Error: Missing space after `!LF$`
    do i = 1, 10
      print *, i
    end do
  end program incorrect_pragma
```

```
!only ASR not generated
  program bad_implicit
    implicit real(a-z)  ! Error: "a-z" is invalid because it includes the entire alphabet without a valid specification
    integer :: x
    x = 10
    print *, x
  end program bad_implicit
```

```
  program invalid_range
    implicit integer(p-t-z)  ! Error: Invalid range "p-t-z" contains multiple dashes
  end program invalid_range
```
  

```
!Only ASR not generated
  program overlapping_ranges
    implicit integer(a-f), real(e-z)  ! Error: Overlap between "a-f" and "e-z" for letters "e-f"
  end program overlapping_ranges
```

```
  program invalid_letters
    implicit real(1-5)  ! Error: "1-5" is invalid as it uses numbers instead of letters
  end program invalid_letters
  ```

parth121101:
Will keep this list updated and will be working on fixing them.

PR #5816, opened 2024-12-24
fix: End function name mismatch with function name generates ast with cc

parth121101:
Towards: #5813

### DIFF
--- a/tests/reference/asr-continue_compilation_2-a6145a1.stderr
+++ b/tests/reference/asr-continue_compilation_2-a6145a1.stderr
@@ -4,6 +4,46 @@ syntax error: Invalid syntax for variable initialization (try inserting '::' aft
 34 |     integer init_x = 1
    |     ^^^^^^^^^^^^^^^^^^ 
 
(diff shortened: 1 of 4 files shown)


tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -4,6 +4,46 @@ syntax error: Invalid syntax for variable initialization (try inserting '::' aft
...
 
+syntax error: End module name does not match module name
+   --> tests/errors/continue_compilation_2.f90:331:1 - 333:28
+    |
+331 |    module my_module
+    |    ^^^^^^^^^^^^^^^^...
+...
+    |
+333 |    end module wrong_module_name 
+    | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
certik, line 15: This can be done later: I think we now know the location of the actual name itself, so we should use it to highlight just `my_module`, not the whole module.

certik (approved):

PR #5822, opened 2024-12-25
Fix for simplification of `Where`

czgdp1807:
Let's see if PRIMA compiles and runs. If so, then I will make simplifier default in the next PR.


czgdp1807:
Ok PRIMA works but `where_15` fails with FAST (something to be looked at after this). Added NOFAST tag.

PR #5824, opened 2024-12-25
feat: Support (`ArrayItem` = `ArrayConstant`) type of Assignments with array indices

jinangshah21:
Here all cases are covered for (`ArrayItem` = `ArrayConstant`) type of Assignments where LHS contains Array Indices.


jinangshah21:
In pass_utils.h , I have specified explicit cases (i.e. types of lhs and rhs) of Assignments in `create_do_loop_assign` for clarity but many things are common, So afterwards we can shorten code.

certik (approved):

Issue #5825, opened 2024-12-25
syntax error for a program containing '&'


gxyd:
## Minimal Reproducible Example

For the below Fortran program:
```f90
program m
   implicit none
   write(*,*)'*prototype_and_cmd_args_to_nlist* error obtaining argument ',&
      &'target length=5'

!!     call set_args(' &

contains
   function sgs()
      character(len=:),allocatable         :: sgs(:)
   end function sgs

end program m

```

on it's execution with LFortran as below, I get an error:
```console
> lfortran --version

> lfortran test.f90
syntax error: Token 'sgs' (of type 'identifier') is unexpected here
 --> test.f90:9:13
  |
9 |    function sgs()
  |             ^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

gxyd:
Seems like the *prescan* output of the above program is incorrect:
```console
> lfortran test.f90 --show-prescan
program m
   implicit none
   write(*,*)'*prototype_and_cmd_args_to_nlist* error obtaining argument ','target length=5'

!!     call set_args(' contains
   function sgs()
      character(len=:),allocatable         :: sgs(:)
   end function sgs

end program m

```

which has `contains` on the same line as the line to be commented.

gxyd:
Gentle ping @saikarna913,  in case you are interested in fixing this issue.

saikarna913:
@gxyd Thanks for mentioning me. I Will work on the issue

gxyd:
Fixed by https://github.com/lfortran/lfortran/pull/5834

PR #5827, opened 2024-12-25
&bug

saikarna913:
This PR fixes the bug #5825

The code has been tested with the following program:
```
program m
   implicit none
   write(*,*)'*prototype_and_cmd_args_to_nlist* error obtaining argument ',&
      &'target length=5'

!!     call set_args(' &

contains
   function sgs()
      character(len=:),allocatable         :: sgs(:)
   end function sgs

end program m
```

and output is: 
```
*prototype_and_cmd_args_to_nlist* error obtaining argument     target length=5
```


saikarna913:
@gxyd Please let me know if any changes are required.

src/lfortran/parser/parser.cpp
```diff
@@ -730,8 +729,11 @@ std::string prescan(const std::string &s, LocationManager &lm,
                     newline = true;
                 }
             }
-            out += s[pos];
+             if (!(s[pos] == '&' &&  previous_nonspace_character(s, pos) == '\n')) {
+                out += s[pos];
             pos++;
+                    }
+            else {pos++;}
```
certik, line 736: The indentation is wrong, let's fix it. Also should be `} else {`.

src/lfortran/parser/parser.cpp
```diff
@@ -730,8 +729,10 @@ std::string prescan(const std::string &s, LocationManager &lm,
                     newline = true;
                 }
             }
-            out += s[pos];
-            pos++;
+             if (!(s[pos] == '&' &&  previous_nonspace_character(s, pos) == '\n')) {
```
gxyd, line 732: This change logically looks good to me, but I would just make the below diff to this PR:
```diff
diff --git a/integration_tests/line_continuation_03.f90 b/integration_tests/line_continuation_03.f90
index 8eee10c56..579dacf66 100644
--- a/integration_tests/line_continuation_03.f90
+++ b/integration_tests/line_continuation_03.f90
@@ -68,4 +68,9 @@ print *, "some string &
 
     &is continued"
 
+print *, '*t',&
+&''
+!!     call set_args ( \'  &
+!! & \n
+
 end program
diff --git a/src/lfortran/parser/parser.cpp b/src/lfortran/parser/parser.cpp
index 627dc1016..cef712910 100644
--- a/src/lfortran/parser/parser.cpp
+++ b/src/lfortran/parser/parser.cpp
@@ -729,10 +729,10 @@ std::string prescan(const std::string &s, LocationManager &lm,
                     newline = true;
                 }
             }
-             if (!(s[pos] == '&' &&  previous_nonspace_character(s, pos) == '\n')) {
+            if (!(s[pos] == '&' &&  previous_nonspace_character(s, pos) == '\n')) {
                 out += s[pos];
-               pos++;
-            } else pos++;
+            }
+            pos++;
         }
         // set the size of the last interval
     //    lm.in_size.push_back(pos-lm.in_start[lm.in_start.size()-1]);
```

gxyd: and run `./run_tests.py -u` to update the reference tests

integration_tests/line_continuation_06.f90
```diff
@@ -0,0 +1,13 @@
+program m
```
gxyd, line 1: Once I make the below diff that I suggested below, we wouldn't need this test file added here i.e. `integration_tests/line_continuation_06.f90` wouldn't be needed.

gxyd (commented):
Just recommended a minor change to parser.cpp and instead of having a new test file, we can test the change in `integration_tests/line_continuation_03.f90` instead of a new file `integration_tests/line_continuation_06.f90`

python
```diff
@@ -0,0 +1,6500 @@
+-- The C compiler identification is GNU 13.3.0
```
gxyd, line 1: Mistakenly added file?

src/lfortran/parser/parser.cpp
```diff
@@ -730,8 +729,8 @@ std::string prescan(const std::string &s, LocationManager &lm,
                     newline = true;
                 }
             }
-            out += s[pos];
-            pos++;
+             if (!(s[pos] == '&' &&  previous_nonspace_character(s, pos) == '\n')) out += s[pos];
```
gxyd, line 732: I generally always prefer to wrap the code within the `if`'s body within `{ }` as that make it much easier to read and debug.

src/lfortran/parser/parser.cpp
saikarna913:
```
print *, '*t',&
&''
!!     call set_args ( \'  &
!! & \n

```

Added the above test case in line_continuation_03.f90
updated the test references

tests/reference/run_dbg-runtime_stacktrace_01-dcc746e.stderr
```diff
@@ -1,9 +1 @@
-code generation error: asr_to_llvm: module failed verification. Error:
```
gxyd, line 1: this isn't a relevant change, maybe you built LFortran without stacktrace disabled.

saikarna913:
@gxyd  i am really sorry. I have tried to make all the commits into a single commit so that the test will only run once but I don't know what happened, can you help me to revert the change that I have done

certik:
Please link the next PR when you close the old one: https://github.com/lfortran/lfortran/pull/5834.

PR #5830, opened 2024-12-26
Switch to simplifier by default and add `--no-experimental-simplifier`

czgdp1807:


czgdp1807:
The question is either,

1. Update reference tests according to experimental simplifier pass.
2. Or use `--no-experimental-simplifier` in `run_tests.py`

I would go for the first.

tests/tests.toml
czgdp1807, line 1: The reference tests which I have removed from this file are already tested in `integration_tests/CMakeLists.txt` (with `llvm` and `gfortran` backend). Keeping them here as well increases redundancy and increases size of the LFortran repository. 

Especially those marked with `llvm` should be removed and only tested in `integration_tests/CMakeLists.txt`. Its safe I think.

.github/workflows/CI.yml
```diff
@@ -792,18 +792,18 @@ jobs:
...
         if: contains(matrix.os, 'ubuntu')
         run: |
-            git clone https://github.com/certik/fpm.git
+            git clone https://github.com/czgdp1807/fpm.git
             cd fpm
-            git checkout lfortran_build_3
-            git checkout a63eb805747f717d420c12de1984e5e15cdb6984
+            git fetch origin lfortran_build_4
+            git checkout lfortran_build_4
+            git checkout 910c461f04506bf87a05a8fbaf7d1b0a79f0bd48
```
Pranavchiku, line 799: FPM is fine but just checking if we are not adding any workarounds here?

tests/reference/asr-derived_types_04-da02dd9.stdout
```diff
@@ -1020,9 +1020,7 @@
                                                             ()
                                                         )]
                                                         0
-                                                        (Allocatable
-                                                            (String 1 -2 () PointerString)
-                                                        )
+                                                        (String 1 -2 () PointerString)
```
Pranavchiku, line 1023: This needs to be checked. Why it is not `Allocatable` now?

tests/reference/asr-derived_types_06-847ca73.stdout
```diff
@@ -1020,9 +1020,7 @@
                                                             ()
                                                         )]
                                                         0
-                                                        (Allocatable
-                                                            (String 1 -2 () PointerString)
-                                                        )
+                                                        (String 1 -2 () PointerString)
```
Pranavchiku, line 1023: same comment here

tests/reference/asr-template_04-f41dd3e.stdout
```diff
@@ -362,8 +362,62 @@
                                                         ))]
                                                         (Array
                                                             (Integer 4)
-                                                            [(()
-                                                            ())]
+                                                            [((IntegerConstant 1 (Integer 4) Decimal)
+                                                            (IntegerBinOp
+                                                                (IntegerBinOp
+                                                                    (IntegerBinOp
```
Pranavchiku, line 368: At some places we remove these and use empty dimensions while here / other we add these dimensions. Why is that so?

.github/workflows/CI.yml
czgdp1807: No work around. Just `build.sh` is modified to use the new flags.

Pranavchiku (commented):
Left initial set of comments, looks fine to me. Let's wait for CI to run and others to review.

tests/reference/asr-template_04-f41dd3e.stdout
czgdp1807: The way handle sizes before was incorrect (didn't matter if we used simplifier pass or not). Doing something like, `ArraySize(FunctionCall(...))` inside a type is incorrect way of doing things. So whenever you have to do `ArraySize(FunctionCall(..))` or `ArraySize(IntrinsicArrayFunction(...))` in a ASR type then it means that the size couldn't be figured out, so we should leave the dimensions empty. `simplifier` would figure out and create an allocatable for it (if needed). 

In fact, the way we handle `ArraySize` creation also needs to be refined and perfect it (like leave the dimension empty only if we are sure that we must create an allocatable for a given non-Var expression).

tests/reference/asr-derived_types_04-da02dd9.stdout
czgdp1807: Most probably this means we don't need an Allocatable type for it.

czgdp1807: We don't need an Allocatable type for it because it is the type of `IntrinsicElementalFunction` and it can be vectorised easily by `array_op_simplifier`. The change is safe and reduces Allocatable types in the ASR.

czgdp1807:
There were many reference tests which were redundant. I removed them from here because they were interfering with analysing the reference tests output. Especially, some LLVM reference tests generate huge files and very large diffs and generate LLVM code according to the OS. I am on macOS and for some tests it generates slightly different code.

In future too I recommend to remove those reference tests which are already tested in `integration_tests/CMakeLists.txt`. LLVM should be removed from reference tests altogether. Reference tests should be only for ASR and ASR to ASR passes. Everything else should be tested via `integration_tests`.

Thirumalai-Shaktivel (approved):
LGTM, thank you!

HarshitaKalani (approved):
Thanks @czgdp1807 ! Looks good to me too.

src/bin/lfortran.cpp
```diff
@@ -2268,7 +2269,7 @@ int main_app(int argc, char *argv[]) {
     app.add_flag("--ignore-pragma", compiler_options.ignore_pragma, "Ignores all the pragmas");
     app.add_flag("--stack-arrays", compiler_options.stack_arrays, "Allocate memory for arrays on stack");
     app.add_flag("--wasm-html", compiler_options.wasm_html, "Generate HTML file using emscripten for LLVM->WASM");
-    app.add_flag("--experimental-simplifier", compiler_options.po.experimental_simplifier, "Use experimental simplifier pass");
+    app.add_flag("--no-experimental-simplifier", no_experimental_simplifier, "Do not use experimental simplifier pass");
```
certik, line 2272: If we merge this PR, it's not an experimental simplifier anymore, right?
```suggestion
    app.add_flag("--no-default-simplifier", no_experimental_simplifier, "Do not use the default simplifier pass");
```

certik (approved):
I think that this is fine. I am going to run some benchmarks first to make sure, if it looks good, let's merge it. Wait for my comment first.

certik:
LFortran in release mode. SNAP:
```
$ time make -j8 FORTRAN=lfortran FFLAGS= MPI=no OPENMP=no
1.04s user 0.25s system 115% cpu 1.119 total
$ time make -j8 FORTRAN=lfortran FFLAGS=--experimental-simplifier MPI=no OPENMP=no
0.96s user 0.25s system 117% cpu 1.031 total
```
So compilation speed is good. Right now I can't compile with `--fast`, investigating.

certik:
For dftatom unfortunately there is a big regression:
```
$ make -f Makefile.manual F90=lfortran F90FLAGS="-I../../src --skip-pass=inline_function_calls,fma --fast --experimental-simplifier"
$ time ./tests/atom_U/F_atom_U
0.27s user 0.01s system 99% cpu 0.287 total
```
vs:
```
$ make -f Makefile.manual F90=lfortran F90FLAGS="-I../../src --skip-pass=inline_function_calls,fma --fast"
$ time ./tests/atom_U/F_atom_U
0.17s user 0.01s system 98% cpu 0.178 total
```
So that's 61.2% slowdown. That's a big slowdown.

certik:
I merged it. I think SNAP is more important and it is clearly faster there.

We need to investigate the regression for dftatom and fix it, but I think it's not a blocker; for now we'll use `--no-experimental-simplifier`.

PR #5832, opened 2024-12-26
fix: missing location information of variables

### LINKED ISSUE #5829: global module variables do not work with `--lookup-name`
```fortran
MODULE sn_module

REAL, ALLOCATABLE, DIMENSION(:,:,:) :: ec

CONTAINS

SUBROUTINE sn_expcoeff ( ndimen )
  INTEGER, INTENT(IN) :: ndimen
  ec(:,1,5) = 1.0
END SUBROUTINE sn_expcoeff


END MODULE sn_module
```

```console
% lfortran a.f90 --line=9 --column=4 --lookup-name
[]
```

Pranavchiku:
Fixes #5829. Although it introduces a regression where `Go To Definition` of `procedure` takes us to variable declaration and not `Function` itself.

### DIFF
--- a/src/libasr/asdl_cpp.py
+++ b/src/libasr/asdl_cpp.py
@@ -362,7 +362,7 @@ class DefaultLookupNameVisitor(ASDLVisitor):
         self.emit("uint32_t last = loc.last;", 2)
         self.emit("if (first <= pos && pos <= last) {", 2)
         self.emit("uint32_t span = last - first;", 3)
-        self.emit("if (span <= min_span) {", 3)
+        self.emit("if (span < min_span) {", 3)
         self.emit("min_span = span;", 4)
         self.emit("return true;", 4)
         self.emit("}", 3)
--- a/src/libasr/asr_lookup_name.h
+++ b/src/libasr/asr_lookup_name.h
@@ -18,6 +18,16 @@ namespace LCompilers::LFortran {
             LookupNameVisitor(uint16_t pos) {
                 this->pos = pos;
             }
+            void visit_ExternalSymbol(const ASR::ExternalSymbol_t &x) {
+                if ((bool&)x) { } // Suppress unused warning
+                if (test_loc_and_set_span(x.base.base.loc)) {
+                    const ASR::symbol_t* sym = this->symbol_get_past_external_(x.m_external);
+                    this->handle_symbol(sym);
+                    if ( ASR::is_a<ASR::ClassProcedure_t>(*sym) ) {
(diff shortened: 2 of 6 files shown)


src/libasr/asdl_cpp.py
```diff
@@ -362,7 +362,7 @@ def visitModule(self, mod):
         self.emit("uint32_t last = loc.last;", 2)
         self.emit("if (first <= pos && pos <= last) {", 2)
         self.emit("uint32_t span = last - first;", 3)
-        self.emit("if (span <= min_span) {", 3)
+        self.emit("if (span < min_span) {", 3)
```
Pranavchiku, line 365: I remember intentionally keeping it `<=` but I suspected this to be wrong.

tests/reference/lookup_name-function_call1-c233951.stdout
```diff
@@ -1 +1 @@
-[{"kind":12,"location":{"range":{"start":{"character":5,"line":8},"end":{"character":24,"line":12}},"uri":"uri"},"name":"eval_1d","filename":"tests/function_call1.f90"}]
\ No newline at end of file
+[{"kind":6,"location":{"range":{"start":{"character":7,"line":4},"end":{"character":27,"line":4}},"uri":"uri"},"name":"eval_1d","filename":"tests/function_call1.f90"}]
```
Pranavchiku, line 1: Here is the regression, now it points to `procedure :: eval_1d` instead of `function eval_1d(x) ... end function`

certik: Why is it doing it? How do we fix it?

certik (approved):
I think the PR is ok, but since there is a regression, I am going to keep it open for now until we decide how to fix it.

Pranavchiku:
I need to check that, looking at it.

tests/reference/lookup_name-function_call1-c233951.stdout
Pranavchiku: Explained at https://github.com/lfortran/lfortran/pull/5832#issuecomment-2563396424

Pranavchiku:
These location information bug are tricky to debug.

certik:
> When we visit `self%eval_1d(x)` in `ast_body_visitor`, `eval_1d` does not exist, so we create an `ExternalSymbol` with same location information as `self%eval_1d(x)` and `DefaultLookupNameVisitor` visits symbols first so it hits `ExternalSymbol` and marks line column of `procedure :: eval_1d`.

That's unrelated, but I don't understand why it should be an external symbol. It should link directly to the method.

PR #5834, opened 2024-12-26
syntax error for a program containing '&'

saikarna913:
This PR fixes the bug #5825

The code has been tested with the following program:

```
program m
   implicit none
   write(*,*)'*prototype_and_cmd_args_to_nlist* error obtaining argument ',&
      &'target length=5'

!!     call set_args(' &

contains
   function sgs()
      character(len=:),allocatable         :: sgs(:)
   end function sgs

end program m
```
Added the test case in the line_continuation_05.f90

### DIFF
--- a/integration_tests/line_continuation_03.f90
+++ b/integration_tests/line_continuation_03.f90
@@ -68,4 +68,9 @@ print *, "some string &
 
     &is continued"
 
+print *, '*t',&
+&''
+!!     call set_args ( \'  &
+!! & \n
+
 end program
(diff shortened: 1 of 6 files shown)


saikarna913:
@gxyd, please review this; the previous one was closed.

gxyd (approved):
Looks much cleaner now. Thanks for fixing this @saikarna913

gxyd:
I squashed and merged the commits, as the commit message weren't the best, and the change overall was small here.

PR #5846, opened 2024-12-27
enh: support multiple integer types for `_lfortran_str_item`

### LINKED ISSUE #5839: bug : different integer kinds passed to `lfortran_str_item` parameter `idx` fails
MRE
```fortran
program test
    character(len=3),parameter :: color = "red"
    integer(2) :: i
    i = 1
    print *, color(i:i)
end program test
```
```console
  %4 = load i16, i16* %i1, align 2
 i32  %5 = call i8* @_lfortran_str_item(i8* getelementptr inbounds ([4 x i8], [4 x i8]* @2, i32 0, i32 0), i16 %4)

```

kmr-srbh:
fixes #5839

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -2292,20 +2292,26 @@ LFORTRAN_API char* _lfortran_strrepeat_c(char* s, int32_t n)
 }
 
 // idx starts from 1
-LFORTRAN_API char* _lfortran_str_item(char* s, int32_t idx) {
-
-    int s_len = strlen(s);
-    int original_idx = idx - 1;
-    if (idx < 1) idx += s_len;
-    if (idx < 1 || idx >= s_len + 1) {
-        printf("String index: %d is out of Bounds\n", original_idx);
-        exit(1);
+#define DEFINE_LFORTRAN_STR_ITEM(type)                                                             \
+    LFORTRAN_API char* _lfortran_str_item_##type(char* s, type idx)                                \
+    {                                                                                              \
+        int s_len = strlen(s);                                                                     \
+        int original_idx = idx - 1;                                                                \
+        if (idx < 1) idx += s_len;                                                                 \
+        if (idx < 1 || idx >= s_len + 1) {                                                         \
+            printf("String index: %d is out of Bounds\n", original_idx);                           \
+            exit(1);                                                                               \
+        }                                                                                          \
+        char* res = (char*) malloc(2);                                                             \
+        res[0] = s[idx - 1];                                                                       \
+        res[1] = '\0';                                                                             \
+        return res;                                                                                \
     }
-    char* res = (char*)malloc(2);
-    res[0] = s[idx-1];
-    res[1] = '\0';
-    return res;
-}
+
+DEFINE_LFORTRAN_STR_ITEM(int8_t)
+DEFINE_LFORTRAN_STR_ITEM(int16_t)
+DEFINE_LFORTRAN_STR_ITEM(int32_t)
+DEFINE_LFORTRAN_STR_ITEM(int64_t)
 
 // idx1 and idx2 both start from 1
 LFORTRAN_API char* _lfortran_str_copy(char* s, int32_t idx1, int32_t idx2) {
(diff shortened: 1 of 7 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2292,20 +2292,26 @@ LFORTRAN_API char* _lfortran_strrepeat_c(char* s, int32_t n)
...
+        if (idx < 1) idx += s_len;                                                                 \
+        if (idx < 1 || idx >= s_len + 1) {                                                         \
+            printf("String index: %d is out of Bounds\n", original_idx);                           \
+            exit(1);                                                                               \
+        }                                                                                          \
+        char* res = (char*) malloc(2);                                                             \
+        res[0] = s[idx - 1];                                                                       \
+        res[1] = '\0';                                                                             \
+        return res;                                                                                \
+    }
```
certik, line 2309: The motivation to have 4 implementations of this function is performance. Otherwise it would be a lot easier to just use 64 bit index, correct?

However, this function is extremely slow for the following reasons:

* It has bounds checking that cannot be turned off even in Release mode
* it allocates a 2 byte string for the output (!)

So this optimization implemented in this PR in my opinion does not make anything faster, just more code to maintain.

To actually speed it up, we need to get rid of the above two issues first.

certik (approved):
The change itself is fine, but it adds a lot more code to maintain for no obvious reason, see my comment.

Why not just cast the index to 64 bit?

I am ok with merging it, if you feel strongly this is the way to fix this, but I'll leave that up to you to merge. :)

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: > Otherwise it would be a lot easier to just use 64 bit index, correct?

Yes, my initial thought was to do the casting only, but it would have led to 8 byte integer index allocations for small strings also. Then `s_len` would have to be `int64_t` too and we would end up using more memory on the stack. 

We have many cases where we need to handle other integer types and the macro would lead to increase in the runtime library binary. So if the memory usage is not an issue, we can surely do the casting.

> It has bounds checking that cannot be turned off even in Release mode

We can handle incorrect indices by returning an empty string, but I think we need the bound checking to avoid undefined behaviors. We do this for other string operations like string slicing also. For string slicing, GFortran allows incorrect indices and prints random garbage values. LFortran handles this gracefully and returns an empty string, like CPython.

```fortran
program main
   character(len=5) :: color
   integer(4) :: k = 2000

   color = "green"

   print *, color(20:k)
   print *, len(color(20:k))
end program main
```
```console
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
 ����������|���������Bg��\����k�J�8�hmg��\0P�|�k�M�8��k!8/)n������+��vp��{��|���hmg��\����Bg��\�@g��\����@g��\���8�%���%���%��L&��_&��s&���&���&���&��'��='��p'��w'���'���'���'���'��(��W(���(���(���(��)��)��})���)���)���)��N*��i*��*���*���*���*��'+��=+��s+���+���+���+���+��,��',��8,��Q3��k3���3��4��14���4���4���4���4���4��5��15��X5��}5���5���5��6��5�L�|6����@g��\��6���6���6���6��>7��L7��W7��_7���7���7���7���7���7��8��Y8��x8���8��9��39��Y9���9���9���:���:��>��>��A>��w>���>���>��?��_?���?���?���?��!pL�|3����d@0g��\8
              �
��9���?��I���+��v?U�5�x86_64
        1981
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90

0
```

> it allocates a 2 byte string for the output (!)

Yes, this can lead to memory fragmentation for many calls, but I think we want null-terminated strings, so we need a string here?

certik: GFortran doesn't seem to do bounds checking on the above example even with `-fcheck=all`.

For LFortran, when we enable bounds checking (by default in Debug mode), it should check all such usage. But it should not check it in Release mode.

kmr-srbh: If the changes look good here, can we merge this PR and handle the bound-checking problem as a separate issue? We will need to do this for other operations also.

certik: Is there any downside of having just 64bit indexing in the runtime library, and always cast it to 64 bit first? I think it would be a lot easier. Given that this function is inefficient anyway, it seems wasteful to try to optimize something that shouldn't exist in the first place.

kmr-srbh: Not a big downside with 64-bit integers. Comparing the trade-offs, the casting methods seems correct for now. On a side note, I think we require the function to return a string, else we will have segmentation fault or incorrect behavior with stuff like printing due to the absence of the null-character. I had tried this implementation and faced this issue.

kmr-srbh:
Ready.

PR #5847, opened 2024-12-27
enh: use delta between input and output pos to determine node

### LINKED ISSUE #5833: location information of `Var` in `RealBinOp`  is incorrect
```fortran
module time_module_xx
  IMPLICIT NONE
  PUBLIC
  REAL :: tparset_ = 0.0
  REAL :: wtime_ = 0.0
END module time_module_xx


module plib_module_xx
use time_module_xx, only: wtime_
PUBLIC
INTEGER :: iproc = 0
INTEGER :: root = 0
INTEGER :: pinit = 0
INTEGER :: pcomm_set = 0
end module

PROGRAM snap_main
  USE time_module_xx, ONLY: tparset_

  USE plib_module_xx, ONLY: pinit,         &
    pcomm_set
  tparset_ = tparset_ + 1.0
END PROGRAM snap_main
```

```console
% lfortran a.f90 --line=23 --column=17 --lookup-name
[]
```

If I use the following diff it works:

```diff
diff --git a/a.f90 b/a.f90
index a8a3493a2..0e060bda8 100644
--- a/a.f90
+++ b/a.f90
@@ -18,7 +18,6 @@ end module
 PROGRAM snap_main
   USE time_module_xx, ONLY: tparset_
 
-  USE plib_module_xx, ONLY: pinit,         &
-pcomm_set
+  USE plib_module_xx, ONLY: pinit, pcomm_set
   tparset_ = tparset_ + 1.0
 END PROGRAM snap_main
```

Pranavchiku:
Fixes #5833. This was super awesome to debug!


certik (approved):
I think it's fine, but I don't understand how it works. Why is delta needed?

Pranavchiku:
It has to do with output_to_input mapping that we do in LocationManager. So the output and input may have a delta, we just use it to identify and then once identified use the inbuilt lm.output_to_input_pos function to render correct location information

Issue #5845, opened 2024-12-27
AssertionFailure on string concatenation of allocatable array


gxyd:
## Minimal Reproducible Example

For the below Fortran program extracted from *fman* repository:

```f90
function f() result(textblock)
    character(len=256), allocatable :: textblock(:)
    textblock=[character(len=len(textblock)+1) :: ' ',textblock]
    textblock=' '//textblock
end function

```

On it's execution with LFortran, I get an error:
```console
> lfortran --versionLFortran version: 0.42.0-785-ga97b0da4b-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran -c test.f90 --show-asr
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2729
    LCompilers::print_stack_on_segfault();
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2554
    return emit_asr(arg_file, lfortran_pass_manager,
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 655
    r = fe.get_asr2(input, lm, diagnostics);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 275
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 297
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 104
    if (res.ok) {
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 4648
    b.is_body_visitor = true;
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 156
    visit_ast(*x.m_items[i]);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/ast.h", line 4943
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/ast.h", line 4899
[...]

Issue #4889, opened 2024-09-24
Length of character variable wrong


harperjf:
This program is a variant of the one in #4888. It compiles but has a run-time error:
```
 program writetestb
  character:: c(6)*3 = 'ab'
  write(*,"(*(A))") 'c = "',c ,'"'
end program writetestb
```
Run-time output from lfortran 0.40.0:
```
c = "abababababab"
```
Run-time output, which is correct, from gfortran, ifx or ifort:
```
c = "ab ab ab ab ab ab "
```

kmr-srbh:
Thanks for reporting this bug! This seems to be an issue with how we pad the character array data when the LHS and RHS lengths are not same.

harperjf:
This version of my program may help in the bug-hunt. It finds ```c``` in three ways that give the wrong values, then the wrong length, then the right answer, when compiled with lfortran. But I would really like to be able to use all the methods.
```
program writeteste
  character:: c(6)*3
  data c/'ab ','cd ','ef ','gh ','ij ','kl '/
  write(*,"(*(A))") 'c = "',c,'"'
  c = ['ab','cd','ef','gh','ij','kl']
  write(*,"(*(A))") 'c = "',c,'"'
  c = ['ab ','cd ','ef ','gh ','ij ','kl ']
  write(*,"(*(A))") 'c = "',c,'"'
end program writeteste
```
Run-time output from lfortran:
```
c = "ab ab ab ab ab ab "
c = "abcdefghijkl"
c = "ab cd ef gh ij kl "
```
Run-time output from AMD flang, gfortran, ifx or ifort:
```
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
```

kmr-srbh:
Thanks for the comprehensive test! :+1:

HarshitaKalani:
This works with main. Let's add a test to close this.
```console
% gfortran a.f90 && ./a.out
c = "ab ab ab ab ab ab "
% lfortran a.f90 
c = "ab ab ab ab ab ab "
```
```console
 % lfortran b.f90
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
% gfortran b.f90 && ./a.out
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
c = "ab cd ef gh ij kl "
```

Issue #3862, opened 2024-04-12
padding of character array elements


gxyd:
```fortran
program testoddity
    character(8) a(2)
    a = ['aa','a ']
    print *, len(a(1))
end program
```

the above program gives the output `2` by LFortran. While the output by GFortran is `8`.

PR #5848, opened 2024-12-27
enh: handle character array for string concat

HarshitaKalani:
Fixes: https://github.com/lfortran/lfortran/issues/5845

### DIFF
--- a/integration_tests/string_43.f90
+++ b/integration_tests/string_43.f90
@@ -1,0 +1,16 @@
+
+program string_43
+    character(len=2) :: textblock(2)
+    textblock = f()
+    print*, "lf"//textblock
+
+    contains
+    function f() result(textblock)
+        character(len=256), allocatable :: textblock(:)
+        allocate(textblock(2))
+        textblock(1) = "lf"
+        textblock(2) = "gf"
+        textblock=[character(len=len(textblock)+1) :: ' ',textblock]
+        textblock=' '//textblock
+    end function
+end program
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1382,7 +1382,7 @@ RUN(NAME string_39 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_40 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
 RUN(NAME string_41 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
 RUN(NAME string_42 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME string_43 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran ONLY_EXPERIMENTAL_SIMPLIFIER NOFAST)
+RUN(NAME string_43 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc ONLY_EXPERIMENTAL_SIMPLIFIER NOFAST)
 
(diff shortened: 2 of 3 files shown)


HarshitaKalani:
I don't understand the issue:
```
make[2]: *** [CMakeFiles/string_43_no_simplifier.dir/build.make:75: CMakeFiles/string_43_no_simplifier.dir/string_43.f90.o] Aborted (core dumped)
make[1]: *** [CMakeFiles/Makefile2:26701: CMakeFiles/string_43_no_simplifier.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
Can someone please suggest how do I reproduce this?

kmr-srbh:
Could you please try `lfortran ./integeration_tests/string_43.f90 --no-experimental-simplifier` ?

HarshitaKalani:
Yes, thanks for the suggestion @kmr-srbh :)

integration_tests/string_43.f90
```diff
@@ -0,0 +1,16 @@
...
+    textblock = f()
+    print*, "lf"//textblock
+
+    contains
+    function f() result(textblock)
+        character(len=256), allocatable :: textblock(:)
+        allocate(textblock(2))
+        textblock(1) = "lf"
+        textblock(2) = "gf"
+        textblock=[character(len=len(textblock)+1) :: ' ',textblock]
```
gxyd, line 13: I printed the value of `len(textblock)` here, and the value printed just above this line of code by LFortran is `2` (it's printing the length of newly assigned 'lf' / 'gf') for `print *, len(textblock)`, which is probably incorrect.

The correct value should be `256`.

Even though this doesn't seem like an obvious problem but this ultimately will lead to incorrect printing above.

integration_tests/string_43.f90
```diff
@@ -0,0 +1,16 @@
+
+program string_43
+    character(len=2) :: textblock(2)
+    textblock = f()
+    print*, "lf"//textblock
```
gxyd, line 5: there is no `if (...) error stop` here, because if there is one it will reveal that this will print an incorrect value with LFortran vs GFortran

gxyd (commented):
The added fix would definitely allow us to move forward with compilation of *fman* third party repository (which is definitely good).

A further fix is needed to test with `if (...) error stop condition` as there is another underlying bug here with regards to `len` on concatenation of Array.

Overall, I'm happy with the changes in this PR to be merged as it is, given that we work on `integration_tests/string_43.f90` further in subsequent PR's maybe.

HarshitaKalani:
Yes, that's correct. But I think that's a different problem which I've faced earlier as well and some issue is already open for it. I can open up a new issue for that if that sounds good. The current PR handles the string concatenation for character arrays.

integration_tests/string_43.f90
HarshitaKalani: Let's open a new issue for that maybe. That's more of handling of strings.

gxyd:
> But I think that's a different problem which I've faced earlier as well and some issue is already open for it.

Understood. I tried searching for an open issue for this, I couldn't find it, maybe if you can find it that would be great and link it here, otherwise we can open a separate issue and link that here.

gxyd (approved):

HarshitaKalani:
I think this is one of them: https://github.com/lfortran/lfortran/issues/4889#issuecomment-2370396656

HarshitaKalani:
A more similar issue: https://github.com/lfortran/lfortran/issues/3862

gxyd:
> A more similar issue: https://github.com/lfortran/lfortran/issues/3862

Nice, I'll merge this PR then.

HarshitaKalani:
Thanks for the review @gxyd :)

PR #5853, opened 2024-12-28
feat: implement `intput_to_output_pos` in `LocationManager`

Pranavchiku:
Cleanup of #5847


certik (approved):
I think this looks good, but I haven't checked all the details, only roughly, and it seems like the right approach now.

I would add some tests if there is a way to test this.

PR #5858, opened 2024-12-28
chore: add `.well-known/funding-manifest-urls` file

kmr-srbh:


.well-known/funding-manifest-urls
```diff
@@ -0,0 +1 @@
+https://www.lfortran.org/funding.json/
```
certik, line 1: This url does not work. Correct url is:

```suggestion
https://www.lfortran.org/funding.json
```

Let me know if this url still works.

kmr-srbh: Thanks for catching this!

certik (approved):

Issue #5849, opened 2024-12-27
Assignment issues in simplifier pass


jinangshah21:
Test 1: 
```fortran
program test1
    integer :: B(4) = [1,2,3,4]
    B([2,1,3,4]) = B
    print *, B
end program
```

Test 2: 
```fortran
program test2
    integer :: A(2,2) = reshape([1,2,3,4],[2,2])
    A(:, :) = A(:, [1,2])
    print *, A
end program
```

Test 3 : 
```fortran
program test3
    integer :: A(2,2) = reshape([1,2,3,4],[2,2])
    A(:, [2,1]) = A(:, [1,2])
    print *, A
end program
```

jinangshah21:
These are some common issues with array indexed assignments. Please add if anyone finds other assignment issues.

AnshMehta1:
> Test 1:
> 
> ```fortran
> program test1
>     integer :: B(4) = [1,2,3,4]
>     B([2,1,3,4]) = B
>     print *, B
> end program
> ```
> 
> Test 2:
> 
> ```fortran
> program test2
>     integer :: A(2,2) = reshape([1,2,3,4],[2,2])
>     A(:, :) = A(:, [1,2])
>     print *, A
> end program
> ```
> 
> Test 3: (BinOp)
> 
> ```fortran
> program test3
>     integer :: A(2,2) = reshape([1,2,3,4],[2,2])
>     integer :: A2(2,2) = reshape([4,5,6,7],[2,2])
>     A([2,1],[2,1]) = A - A2
>     print *, A
> end program
> ```
> 
> Test 4: (Reshape)
> 
> ```fortran
> program test4
>     integer :: A(2,2) = reshape([1,2,3,4],[2,2])
>     integer :: B(4) = [1,2,3,4]
>     A([2,1],[2,1]) = reshape(B, [2,2])
>     print *, A
> end program
> ```

For , the first test , I handled assignment `A([1,2]) = A([2,1])` in main in #5786 , I think similarly this case can be handled.

jinangshah21:
Simplifer pass handles it but just we have to create temporary for RHS wherever LHS and RHS has same variables. 
Test 3 and Test 4 seems to be handled in `array_op_simplifier`

AnshMehta1:
> Simplifer pass handles it but just we have to create temporary for RHS wherever LHS and RHS has same variables. Test 3 and Test 4 seems to be handled in `array_op_simplifier`

Ahh, okay.

PR #5859, opened 2024-12-28
fix: Correct Loop bounds Logic for loop generated for `array_indexed` array reference  

jinangshah21:
With this only one issue #5849 is pending for Assignments statements in simplifier pass.


jinangshah21:
Test for this PR is failing for no_experimental_simplifier because this test is only supported with simplifier.

certik:
I think you can register a test to be simplifier only.

certik (approved):

PR #5875, opened 2024-12-30
fix: use 1 as length of newline character in `new_line()` intrinsic

### LINKED ISSUE #5869: (PRIMA) Handle runtime `new_line('')` argument passed in a function
MRE : 
```fortran
program test
    character(len=:), allocatable :: string 
    character(len=*), parameter :: newline = new_line('')
    integer :: j
    string = "hello" // new_line('') // "world"
    j = index(string, newline) 
    print *, j
end program 
```

Here j should be 6 but output is 1. 

Also if we do `j = index(string, new_line(' '))`  then it works so I think runtime arg should be handled.

jinangshah21:
Fixes #5869

### DIFF
--- a/integration_tests/intrinsics_218.f90
+++ b/integration_tests/intrinsics_218.f90
@@ -1,8 +1,15 @@
-program main
-implicit none
-
-integer, parameter :: new_len = len(new_line('a'))
-print *, new_len
-if (new_len /= 1) error stop
-print*, "Hello, World!", new_line('a')
-end
+program intrinsics_218
+    implicit none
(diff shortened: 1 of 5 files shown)


jinangshah21:
In Fortran Backend "/n" is considered as newline (`character(len=1, kind=1), parameter :: str = "`) but we need actual "/n". (character(len=1, kind=1), parameter :: str = "/n"`)

certik:
I think that's fine. CI has to pass first.

Pranavchiku (approved):

gxyd (approved):
Looks good. Neat change.

PR #5879, opened 2024-12-31
fix: Use variable in Subroutinecall instead its `m_type_declaration` and create load when function variable is local

### LINKED ISSUE #5868: (PRIMA) Function Var Points to its `m_type_declaration` when passed in a function which passes it to another function
MRE : 
```fortran 
program procedure_08
    call temp(calfun)
contains

    subroutine calfun(x) !! Actual function to be called
        implicit none
        integer, intent(inout) :: x(:)
        x = 2
    end subroutine calfun

    subroutine cb(x) !! Called function
        implicit none
        integer, intent(inout) :: x(:)
        x = 1
    end subroutine cb

    subroutine temp(call_back)
        implicit none
        procedure(cb) :: call_back
        call temp2(call_back)
    end subroutine temp

    subroutine temp2(call_back)
        implicit none
        procedure(cb) :: call_back
        integer :: x(4) 
        call call_back(x) 
        print *, x
        if(x(1) /= 2) error stop
    end subroutine temp2
end program procedure_08
```

Here `call call_back(x)` in `temp2` should call `calfun` instead of `cb`.

jinangshah21:
Fixes #5868

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8854,18 +8854,9 @@ public:
                                     tmp = llvm_utils->CreateLoad2(arg->m_type, tmp);
                                 }
                             } else {
-                                if(arg->m_type_declaration && ASR::is_a<ASR::Function_t>(
-                                        *ASRUtils::symbol_get_past_external(arg->m_type_declaration))){
-                                    // (FunctionType)** --> (FunctionType)*
-                                    ASR::Function_t* fn = ASR::down_cast<ASR::Function_t>(
-                                        symbol_get_past_external(arg->m_type_declaration));
-                                    uint32_t h = get_hash((ASR::asr_t*)fn);
-                                    if (ASRUtils::get_FunctionType(fn)->m_deftype == ASR::deftypeType::Implementation) {
-                                        LCOMPILERS_ASSERT(llvm_symtab_fn.find(h) != llvm_symtab_fn.end());
-                                        tmp = llvm_symtab_fn[h];
-                                    } else if(llvm_symtab_fn_arg.find(h) != llvm_symtab_fn_arg.end()) {
-                                        tmp = llvm_symtab_fn_arg[h];
-                                    }
+                                if( arg->m_intent == intent_local && ASR::is_a<ASR::FunctionType_t>(*arg->m_type) ) {
+                                    // (FunctionType**) --> (FunctionType*)
+                                    tmp = llvm_utils->CreateLoad2(arg->m_type, tmp);
                                 }
                                 if( orig_arg &&
                                     !LLVM::is_llvm_pointer(*orig_arg->m_type) &&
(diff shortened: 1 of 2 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8854,18 +8854,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                                    ASR::Function_t* fn = ASR::down_cast<ASR::Function_t>(
-                                        symbol_get_past_external(arg->m_type_declaration));
-                                    uint32_t h = get_hash((ASR::asr_t*)fn);
-                                    if (ASRUtils::get_FunctionType(fn)->m_deftype == ASR::deftypeType::Implementation) {
-                                        LCOMPILERS_ASSERT(llvm_symtab_fn.find(h) != llvm_symtab_fn.end());
-                                        tmp = llvm_symtab_fn[h];
-                                    } else if(llvm_symtab_fn_arg.find(h) != llvm_symtab_fn_arg.end()) {
-                                        tmp = llvm_symtab_fn_arg[h];
-                                    }
+                                if( arg->m_intent == intent_local && ASR::is_a<ASR::FunctionType_t>(*arg->m_type) ) {
```
kmr-srbh, line 8857: Could you help me understand the need for checking `intent_local` here? My understanding is that a `procedure` variable inside a subroutine has `Intent::Unspecified`. I think this block will be reached in cases where a local auxiliary variable is created for handling optional arguments, or a local `procedure` variable is passed as argument to the subroutine.

jinangshah21: Yes, you are correct. Previously we were using function pointer directly which pointed to its `m_type_declaration`. Now we just convert double pointer to single pointer if it is auxilliary local varible.

kmr-srbh: Okay, but how does it work for the MRE in #5868?

jinangshah21: If you see part I deleted, it was passing `m_type_declaration` as subroutine call arg but now it passes variable. That's why #5868 works

kmr-srbh: Yes, but when calling `temp2` inside `temp` as `call temp2(call_back)`, the variable `call_back` has `Intent` as `Unspecified`.

jinangshah21: Yes, we don't want it to CreateLoad2 it is for cases handled by deleted part

kmr-srbh: I see. So the code added by you is for preventing the incorrect behavior by not doing anything for the `procedure` variables which are subroutine parameters, as the previous code handled the `m_type_declaration` always. For auxiliary and local procedure variables, you load the pointer. Makes sense.

kmr-srbh (approved):

certik (approved):

PR #5884, opened 2025-01-01
fix :Visit internals of the symbol instead of passing it if it's identified as classProcedure

### LINKED ISSUE #5844: bug : `pass_by_array` doesn't replace the name for the changed function with use of `class`
MRE
```fortran
module modd
    interface oo
    procedure :: ff
    end interface oo
    contains 
    function ff(rr) result(r)
        integer, allocatable :: r(:)
        integer :: rr(:)
    end function
end module modd

program test
    use modd
    
    type :: m
    end type m
    integer, allocatable :: ll(:)

    type(m) :: tt2
    contains
    subroutine ffo(dummy) 
        class(m):: dummy
        integer, allocatable :: r(:)
        integer, dimension(5) :: rr
        r = oo(rr)
    end subroutine ffo
end program test
```
```console
ASR verify pass error: ExternalSymbol::m_original_name ('ff') + scope_names not found in a module 'modd'
  Binary file "/home/assem/lfortran_test/lfortran/src/bin/lfortran", in _start
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 2729, in main
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 2521, in main_app(int, char**)
    }
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 822, in (anonymous namespace)::dump_all_passes(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::CompilerOptions&)
    pass_manager.dump_all_passes(al, asr.result, compiler_options.po, diagnostics, lm);
LCompilersException: Verify failed in the pass: pass_array_by_data

```

If we change `class` into --> `type`, it works. That's due to some externalSymbol that gets created when we use `class` instead of `type`. after `pass_by_array` run, it changes function `ff` name into another name which raises that error we have.

assem2002:
fixes #5844

### DIFF
--- a/src/libasr/pass/pass_array_by_data.cpp
+++ b/src/libasr/pass/pass_array_by_data.cpp
@@ -707,6 +707,9 @@ class RemoveArrayByDescriptorProceduresVisitor : public PassUtils::PassVisitor<R
             for( auto& item: current_scope->get_scope() ) {
                 if (ASR::is_a<ASR::Function_t>(*item.second)) {
                     ASR::FunctionType_t* func_type = ASRUtils::get_FunctionType(item.second);
+                    SymbolTable* current_scope_copy = current_scope;
+                    visit_symbol(*item.second);
+                    current_scope = current_scope_copy;
                     if (func_type->n_arg_types >= 1 && ASR::is_a<ASR::ClassType_t>(*func_type->m_arg_types[0])) {
                         continue;
                     }
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1614,6 +1614,7 @@ RUN(NAME pass_array_by_data_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llv
 RUN(NAME pass_array_by_data_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm)
 RUN(NAME pass_array_by_data_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME pass_array_by_data_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
+RUN(NAME pass_array_by_data_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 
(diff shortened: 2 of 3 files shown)


src/libasr/pass/pass_array_by_data.cpp
```diff
@@ -707,6 +707,9 @@ class RemoveArrayByDescriptorProceduresVisitor : public PassUtils::PassVisitor<R
             for( auto& item: current_scope->get_scope() ) {
                 if (ASR::is_a<ASR::Function_t>(*item.second)) {
                     ASR::FunctionType_t* func_type = ASRUtils::get_FunctionType(item.second);
+                    SymbolTable* current_scope_copy = current_scope;
+                    visit_symbol(*item.second);
+                    current_scope = current_scope_copy;
                     if (func_type->n_arg_types >= 1 && ASR::is_a<ASR::ClassType_t>(*func_type->m_arg_types[0])) {
                         continue;
                     }
```
assem2002, line 715: This `if` condition workaround the function that are to be considered `ClassProcedure` as `pass_array_by_data` pass can't handled such function appropriatly due to symbol dependencies (ExternalSymbol depending on ClassProcedure), The condition just skips deleting the function though the new one exists. The issue raised when the condition identified a function that isn't used as classProcedure to be classProcedure and neglected its internals that had some symbols that used to point to old functions and they should be removed but no they don't get removed.

assem2002:
I'll push another PR to handle the symbols pointing to the old functions replaced by `pass_array_by_data` pass, to get the pass to handle `ExternalSymbol->ClassProcedure->Function` symbols and don't use the old ones.

maybe we can wait for this PR till I handle that PR.

gxyd (approved):
I went through the changes introduced in the ASR pass by this PR, and I noticed the below changes:
```diff
<                                                     Private
<                                                 ),
<                                             interface_1@ff:
<                                                 (ExternalSymbol
<                                                     6
<                                                     interface_1@ff
<                                                     2 ff
<                                                     pass_array_by_data_09_mod
<                                                     []
<                                                     ff
```

which look good to me. I would always love to see this kind of information in the PR description it would really help me better understand the underlying changes, as sometimes words aren't easy to interpret, seeing the ASR changes make me feel confident that the changes are in the right direction.

assem2002:
I'll try to always do that. The did seem to be hard to follow just using words.

Issue #5867, opened 2024-12-29
Figure out how to use a newer cmake in our tests (`-g` issue)


certik:
See e.g. https://github.com/lfortran/lfortran/issues/5865.

PR #5886, opened 2025-01-01
explicitly set CMAKE_Fortran_FLAGS as empty in debug mode

gxyd:
## Description

The solution in this PR isn't something I really like, but I would like to see with the tests first if this works or not, for that that I've temporarily upgraded the version of cmake to 3.31.1, with which I was able to reproduce the issue locally as well, and I'll try to come up with a better solution if possible.

intends to fix: https://github.com/lfortran/lfortran/issues/5867


gxyd:
I'll have to first fix error:
```console
  error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ cmake =3.31.1 * is installable and it requires
      │  └─ libuv >=1.49.2,<2.0a0 *, which can be installed;
      └─ nodejs =18.20.2 * is not installable because it requires
         └─ libuv >=1.48.0,<1.49.0a0 *, which conflicts with any installable versions previously reported.
  critical libmamba Could not solve for environment specs
```

certik:
You can upgrade nodejs to 20, that should fix it.

Yes, we need to unset whatever cmake sets by default, one way or the other.

PR #5892, opened 2025-01-02
fix: 'intrinsic' attribute refers to intrinsic procedure

### LINKED ISSUE #5857: intrinsic variable conflicts with variable of same name

For the below Fortran program, extracted from *fman* repository:
```f90

module M_match
   implicit none

   integer,parameter :: byte=kind(1)

   integer(kind=byte),parameter :: LETA=97_byte

   integer(kind=byte),parameter :: CHAR=LETA

contains

   function r2f(ints)
      integer,intent(in)          :: ints(:)
      character(len=size(ints)-1) :: r2f
      integer                     :: i
      intrinsic char
      r2f(i:i)=char(ints(i))
   end function r2f

end module M_match
```

on it's execution with LFortran, gives the below semantic error:
```console
> lfortran --version
LFortran version: 0.42.0-817-g8892f4ab1-dirty
Platform: macOS ARM
LLVM: 11.1.0

> lfortran test.f90
semantic error: char was declared as a variable, it can't be called as a function
  --> bootstrap/fman.f90:18:16
   |
18 |       r2f(i:i)=char(ints(i))
   |                ^^^^^^^^^^^^^ help: use the compiler option "--implicit-interface" to use intrinsic functions


```

gxyd:
## Description

earlier 'intrinsic' attribute had almost no effect in ASR creation, this PR makes sure that a declaration like 'intrinsic abs' uses the intrinsic procedure 'abs' even in case of an existing declaration of variable 'abs'

Fixes #5857

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -103,15 +103,23 @@ public:
     ASR::ttype_t *tmp_type;
 
     SymbolTableVisitor(Allocator &al, SymbolTable *symbol_table,
-        diag::Diagnostics &diagnostics, CompilerOptions &compiler_options, std::map<uint64_t, std::map<std::string, ASR::ttype_t*>> &implicit_mapping,
-        std::map<uint64_t, ASR::symbol_t*>& common_variables_hash, std::map<uint64_t, std::vector<std::string>>& external_procedures_mapping,
+        diag::Diagnostics &diagnostics, CompilerOptions &compiler_options,
+        std::map<uint64_t, std::map<std::string, ASR::ttype_t*>> &implicit_mapping,
+        std::map<uint64_t, ASR::symbol_t*>& common_variables_hash,
+        std::map<uint64_t, std::vector<std::string>>& external_procedures_mapping,
+        std::map<uint64_t, std::vector<std::string>>& explicit_intrinsic_procedures_mapping,
         std::map<uint32_t, std::map<std::string, ASR::ttype_t*>> &instantiate_types,
         std::map<uint32_t, std::map<std::string, ASR::symbol_t*>> &instantiate_symbols,
         std::map<std::string, std::map<std::string, std::vector<AST::stmt_t*>>> &entry_functions,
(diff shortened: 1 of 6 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -103,15 +103,23 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
         std::map<uint32_t, std::map<std::string, ASR::ttype_t*>> &instantiate_types,
         std::map<uint32_t, std::map<std::string, ASR::symbol_t*>> &instantiate_symbols,
         std::map<std::string, std::map<std::string, std::vector<AST::stmt_t*>>> &entry_functions,
         std::map<std::string, std::vector<int>> &entry_function_arguments_mapping,
         std::vector<ASR::stmt_t*> &data_structure, LCompilers::LocationManager &lm)
-      : CommonVisitor(al, symbol_table, diagnostics, compiler_options, implicit_mapping, common_variables_hash, external_procedures_mapping,
-                      instantiate_types, instantiate_symbols, entry_functions, entry_function_arguments_mapping, data_structure, lm) {}
+      : CommonVisitor(
+            al, symbol_table, diagnostics, compiler_options, implicit_mapping,
+            common_variables_hash, external_procedures_mapping,
```
gxyd, line 118: I also did some formatting of code here, trying to reduce the code line length, that makes it easier to read and understand.

gxyd:
Two things:
1. Either need to fix the *fortran* backend or disable it for the test case
2. appropriate place to use `erase_explicit_intrinsic_procedures` function, which I've added in this PR.

gxyd:
The CI error is because of the output of `fun2` function being:
```f90
integer(4) function fun2() result(i)
    print *, "intrinsic elemental char(65): ", Char(65)
    if (Char(65) /= "A") then
        error stop
    end if
end function fun2
```

and as there is no explicit `intrinsic char`, we get that error because of an existing `char` variable.

certik (approved):
I think that this is fine. I don't know if this fixes every possible combination, but if we discover further issues, we can improve the logic.

Issue #5890, opened 2025-01-02
Compile time evaluation got slower


certik:
Using the benchmark from: https://lfortran.org/blog/2024/06/lfortran-the-fastest-open-source-compiler-in-compile-time-evaluation-of-an-array-benchmark/, we are now slower than GFortran for N = 65535.
```
$ time lfortran a.f90 -o a.out
lfortran a.f90 -o a.out  0.84s user 0.07s system 102% cpu 0.884 total
$ time gfortran a.f90 -o a.out
gfortran a.f90 -o a.out  0.48s user 0.05s system 98% cpu 0.539 total
```
We are doing something inefficiently, let's fix it. It should take roughly 0.08s, so 10x faster.

czgdp1807:
`lfortran --show-asr --no-indent ../debug.f90  12.45s user 0.06s system 99% cpu 12.514 total` - All the time is spent in AST to ASR. :-).

czgdp1807:
Yes, confirmed its AST to ASR which takes all the time.

certik:
With #5893 I am getting:
```
$ time lfortran a.f90 -o a.out
lfortran a.f90 -o a.out  0.08s user 0.05s system 121% cpu 0.104 total
```

certik:
With https://github.com/lfortran/lfortran/pull/5907 I am getting for N=10000000:
```console
$ time lfortran a.f90 -o a.out
lfortran a.f90 -o a.out  1.10s user 0.24s system 102% cpu 1.319 total
```
and with N=65535:
```console
$ time lfortran a.f90 -o a.out
lfortran a.f90 -o a.out  0.07s user 0.04s system 123% cpu 0.091 total
```
and with N=75000:
```console
$ time lfortran a.f90 -o a.out
lfortran a.f90 -o a.out  0.06s user 0.04s system 124% cpu 0.083 total
# most are around 0.097s, so that's probably a more accurate number, but I did get 0.083s once
```

certik:
Note: I don't know why it says "124% cpu", since it means we are running more than one thread. We don't do it directly in LFortran, so likely some dependency is doing it, either LLVM, or maybe the clang linker, or some C++ standard library data structure.

PR #5893, opened 2025-01-02
Speeding up compile time evaluation of ImpliedDoLoop

czgdp1807:
https://github.com/lfortran/lfortran/issues/5890

The problem was this below line,

https://github.com/lfortran/lfortran/blob/c219981e5502807afab4ab963b11b7171516176f/src/lfortran/semantics/ast_common_visitor.h#L7333

The above line is executed inside a loop (which is of size `65535`). So in each iteration the `visitor` object gets created and destroyed leading to slow down. I moved the class out, and the `visitor` object creation out of the loop. Then it speeds up.

On my system,

**Before**

`lfortran ../debug.f90 -o a.out  12.52s user 0.24s system 100% cpu 12.745 total`

**After**

`lfortran ../debug.f90 -o a.out  0.37s user 0.11s system 86% cpu 0.557 total`

**Code**

```fortran
program compile_time
implicit none
integer, parameter :: N = 65535
integer, parameter :: sp = kind(1.0)
integer, parameter :: dp = kind(1.d0)
real(dp), parameter :: pi = 2*asin(1._dp)
real(dp), parameter :: a = 0, b = pi
real(dp), parameter :: dx = (b-a)/N
integer :: i
real(dp), parameter :: X(*) = [(sin(a+(b-a)*i/N), i = 1, N)]
real(dp), parameter :: S = sum(X)*dx
logical, parameter :: l = S < 2
real(kind=merge(sp, dp, l)) :: y
if (kind(y) == sp) then
print *, "true"
else
print *, "false"
end if
print *, S
end program
```

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -90,6 +90,193 @@ static inline int64_t stmt_label(AST::stmt_t *f)
     }
 }
 
+struct IntrinsicSignature {
+    std::vector<std::string> kwarg_names;
+    int positional_args, max_args;
+
+    IntrinsicSignature(std::vector<std::string> kwarg_names_,
+        int positional_args_, int max_args_): kwarg_names(kwarg_names_),
+        positional_args(positional_args_), max_args(max_args_) {}
+};
+
+template <typename T>
+class ImpliedDoLoopValuesVisitor : public ASR::BaseWalkVisitor<ImpliedDoLoopValuesVisitor<T>> {
+    public:
+    Allocator &al;
+    std::vector<ASR::symbol_t*>& loop_vars;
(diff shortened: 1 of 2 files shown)


kmr-srbh:
On Ubuntu, AMD Ryzen 5 3500U, 8 GB RAM

### Before
```console
true
1.99999999961701591e+00

real    0m16.003s
user    0m15.643s
sys     0m0.194s
```

### After

```console
true
1.99999999961701591e+00

real    0m0.580s
user    0m0.417s
sys     0m0.120s
```

certik (approved):
This fixes the main issue, so I am going to merge it. But it seems we are still slower than we used to be: https://github.com/lfortran/lfortran/issues/5890#issuecomment-2567905063.

PR #5894, opened 2025-01-02
build: handle stacktraces before setting `LFORTRAN_LLVM_COMPONENTS`

kmr-srbh:
This makes sure we have a value for `WITH_LLVM_STACKTRACE` to handle appending `symbolize object`. Previously, this variable was defined later in the scope and we did not append the component causing a build failure.


gxyd:
> This did not occur previously because the variable WITH_LLVM_STACKTRACE was undefined until this scope, causing build failure later.

I'm sorry, I don't understand this part, could you please explain a little more on this? The first part, I'm able to follow from here: https://github.com/lfortran/lfortran/blob/main/CMakeLists.txt#L155-L157

gxyd:
Oh, seems like this isn't ready for a review, let me know once this is ready for a review.

gxyd:
Btw, I personally don't see the benefit of force-pushing the changes, as it gets rid of the older changes completely, making the review a little difficult, we can always squash the commits when merging if you don't like one commit or the other.

kmr-srbh:
> > This did not occur previously because the variable WITH_LLVM_STACKTRACE was undefined until this scope, causing build failure later.
> 
> I'm sorry, I don't understand this part, could you please explain a little more on this? The first part, I'm able to follow from here: https://github.com/lfortran/lfortran/blob/main/CMakeLists.txt#L155-L157

Previously the variable `WITH_LLVM_STACKTRACE` was defined later in the file and was used as `false` by default in the `if` block, hence the `symbolize object` was not appended to `LFORTRAN_LLVM_COMPONENTS`, causing the build failure.

kmr-srbh:
> Oh, seems like this isn't ready for a review, let me know once this is ready for a review.

The PR is ready for review. The previous fix set `WITH_LLVM_STACKTRACE` to `true`, but I saw that we want to allow the user to choose whether they want to use `BFD` or LLVM stacktraces. The previous fix set the LLVM stacktrace to default, which was not correct.

> Btw, I personally don't see the benefit of force-pushing the changes, as it gets rid of the older changes completely, making the review a little difficult, we can always squash the commits when merging if you don't like one commit or the other.

I force pushed over the previous commit due to the above reason. :)

kmr-srbh:
Not sure why it fails for LLVM 12 - 19.

.github/workflows/CI.yml
```diff
@@ -949,7 +949,7 @@ jobs:
             nodejs=18.20.2
 
       - uses: mamba-org/setup-micromamba@v2.0.2
-        if: "!(contains(matrix.llvm-version, '10') || contains(matrix.llvm-version, '19'))"
+        if: ${{! (contains(matrix.llvm-version, '10') || contains(matrix.llvm-version, '19'))}}
```
gxyd, line 952: This semantically is the same, but since it gets rid of the error locally on my vscode:
```console
Unexpected symbol: '"'. Located at position 1 within expression: "!(contains(matrix.llvm-version, '10') || contains(matrix.llvm-version, '19'))"
```

I think this is a good change after referring to documentation here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#about-expressions.

kmr-srbh: Yes, looking through the `CI.yml` file I found this, and thought maybe this missed compilation was causing the error, but it is something else. I have no idea why it segfaults. I tested with LLVM 19 locally and it passed.

gxyd: Oh, interesting. I'm checking it and see if I pick on something causing the current failure.

kmr-srbh: Thanks @gxyd!

CMakeLists.txt
```diff
@@ -145,6 +145,76 @@ set(WITH_TARGET_AARCH64 no CACHE BOOL "Enable target AARCH64")
...
+    CACHE BOOL "Build with dwarfdump support")
+set(WITH_LINKH no
+    CACHE BOOL "Build with link.h support")
+set(WITH_MACHO no
+    CACHE BOOL "Build with mach-o support")
+set(WITH_STACKTRACE no
+    CACHE BOOL "Build with stacktrace support (requires binutils-dev)")
+set(WITH_LIBUNWIND no
+    CACHE BOOL "Build with libunwind (required libunwind-dev)")
+if (WITH_LIBUNWIND_FORCE)
```
gxyd, line 164: I think it's because of this variable, earlier we used to set it later.

Since some of these variables are latter changed, you would probably need to move "if ... endif()" conditions back to where they were earlier

gxyd: `WITH_LIBUNWIND_FORCE` is one part of the problem I believe, and think similarly might be others?

gxyd: e.g. if you notice the usage of `WITH_LIBUNWIND_FORCE` in the CMakeLists.txt you would notice that:
```cmake
    if(CMAKE_SYSTEM_NAME MATCHES "Linux" AND LLVM_VERSION_MAJOR GREATER_EQUAL 12)
        set(WITH_LIBUNWIND_FORCE yes)
```

and once `WITH_LIBUNWIND_FORCE` is set to `yes`, there is no condition below it anymore to set `WITH_LIBUNWIND` as `yes` anymore, hence causing the failure for versions >= 12 but not for versions < 11, as `libunwind` **isn't** a dependency for LLVM versions < 12

gxyd: for some of the variables, having a double `if ... endif()` might be needed, one where it's in the current *main* and the one where you put in this PR.

kmr-srbh: Thanks a lot for the review @gxyd! This may have caused the issue actually. It is strange why it did not happen locally. I pushed a commit to test at the CI. :)

gxyd: > It is strange why it did not happen locally.

It sure is, for me, as I'm on macOS, I wouldn't be able to do that locally though.

Seems the CI fails again, maybe as I mentioned above:
> WITH_LIBUNWIND_FORCE is one part of the problem I believe, and think similarly might be others?

Do you think if it's the case?

kmr-srbh: I checked for the other variables, but they are not used anywhere else. Moreover, I think only this variable is handled according to the LLVM version?

gxyd: I pushed one commit, let's see if it passes or not

CMakeLists.txt
```diff
@@ -371,6 +368,9 @@ set(WITH_LIBUNWIND no
 if (WITH_LIBUNWIND_FORCE)
     set(WITH_LIBUNWIND yes)
 endif()
+if (WITH_LIBUNWIND)
```
gxyd, line 371: The reason, I think this change is needed is because, `LIBUNWIND_LIBRARY ` set's the location for libunwind library, and it is set to *NOT FOUND*, when it's used above, hence it isn't able to find this library.

.github/workflows/CI.yml
certik: I think the new change is better based on the documentation. How is it that it was working before?

certik (approved):

.github/workflows/CI.yml
gxyd: > I think the new change is better based on the documentation. How is it that it was working before?

It's semantically the same and not an *incorrect syntax* per say, just that it isn't recommended to use that syntax.

gxyd: > This semantically is the same, but since it gets rid of the error locally on my vscode:

The VS code extensions probably is a little harsher on it, and showing it as an *error* instead of a *warning*

Issue #5891, opened 2025-01-02
(PRIMA) Incorrect size allocation in spread intrinsic


jinangshah21:
MRE :
```fortran
program temp
    integer :: A(2) = [1,2]
    integer :: npt = 3
    print *, spread(A, dim=1, ncopies=npt)
end program temp
```

GFortran Output :
```console
(lf) jinu2110@JNSLAP:/mnt/c/lfortran$ gfortran d.f90 && ./a.out
           1           1           1           2           2           2
```

LFortran Output :
```console
(lf) jinu2110@JNSLAP:/mnt/c/lfortran$ lfortran d.f90 --skip-pass=print_arr 
1    1    1    2    2    2    0    0    0
```

jinangshah21:
This is issue with simplifier. It works without Simplifier.

czgdp1807:
Consider the following example,

```fortran
program temp
     integer :: i
     integer :: A(2, 3, 4) = reshape([(i, i = 1, 24)], [2, 3, 4])
     integer :: npt = 7
     print *, shape(spread(A, dim=1, ncopies=npt))
     print *, shape(spread(A, dim=2, ncopies=npt))
     print *, shape(spread(A, dim=3, ncopies=npt))
     print *, shape(spread(A, dim=4, ncopies=npt))
end program temp
```

The implementation of `spread` (`instantiate_Spread`) is incorrect. The output shape can be figured via a loop. There is no invariant expression that I can figure out right now.

czgdp1807:
Basically we need to fix `spread` end to end - the function we generate for it, the return type, etc.

czgdp1807:
https://github.com/lfortran/lfortran/pull/5903#discussion_r1901525586 contains the way forward for creating output shape of `spread`. :-) 

Fixing implementation of `spread` is also needed.

Pranavchiku:
check if it can be fixed via #6154

jinangshah21:
Closing it as done via #6154.

PR #5903, opened 2025-01-02
fix: set allocation size for 'spread' intrinsic correctly

gxyd:
## Description

https://github.com/lfortran/lfortran/issues/5891

### DIFF
--- a/src/libasr/pass/simplifier.cpp
+++ b/src/libasr/pass/simplifier.cpp
@@ -288,7 +288,9 @@ void set_allocation_size_elemental_function(
 bool set_allocation_size(
     Allocator& al, ASR::expr_t* value,
     Vec<ASR::dimension_t>& allocate_dims,
-    size_t target_n_dims
+    size_t target_n_dims,
+    Vec<ASR::stmt_t*>* current_body=nullptr,
+    SymbolTable* scope=nullptr
 ) {
     if ( !ASRUtils::is_array(ASRUtils::expr_type(value)) ) {
         return false;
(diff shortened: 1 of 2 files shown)


gxyd:
**NOTE** for the reviewer: Please review the PR once the tests, but no need to merge this PR, and I'll make some minor refinements tomorrow or the day after (I might not work tomorrow).

certik:
Looks good to me, but @czgdp1807 should review this one, since it changes the simplifier.

czgdp1807 (commented):
Does the example I added in https://github.com/lfortran/lfortran/issues/5891#issuecomment-2567809334 works?

czgdp1807:
I think this PR doesn't fix the issue #5891.

src/libasr/pass/simplifier.cpp
```diff
@@ -564,33 +577,59 @@ bool set_allocation_size(
...
+                            )
+                        ));
+                        ASR::expr_t* merge_i = ASRUtils::EXPR(
+                            ASRUtils::make_IntrinsicElementalFunction_t_util(
+                                al, loc, static_cast<int64_t>(ASRUtils::IntrinsicElementalFunctions::Merge),
+                                merge_i_args.p, merge_i_args.size(), 0, ASRUtils::expr_type(int32_one),
+                                nullptr
+                            )
+                        );
+                        allocate_dim.m_length = merge_i;
```
czgdp1807, line 632: We have `create_Merge`. However for the general case (example in https://github.com/lfortran/lfortran/issues/5891#issuecomment-2567809334) will not work with this fix. Basically the new shape is determined by the following logic,

```fortran
new_shape = [(0, i = 1, input_array_rank + 1)]
i = 1
j = 1
do i = 1, input_array_rank + 1
    if i != dim then
        new_shape[i] = original_shape[j]
        j = j + 1
    else
        new_shape[i] = ncopies
    end if
end do
```
 There are two parts of the above, `i == dim` and `i != dim`. We need to increase j only if `i != dim`. The `merge` invariant here can be, 

`merge(original_shape(input_array, j), ncopies, i != dim)`

You can use the above `merge` invariant in `allocate`.

`original_shape` can be defined as,

```fortran
integer function original_shape(input_array, j)
integer :: input_array(:)
integer, intent(out) :: j
original_shape = size(input_array, j)
j = j + 1
end function
```

The `original_shape` will be called when `i != dim` and `j` will be updated only in that case. So we are safe. No loops needed.

czgdp1807 (commented):
Merge this PR only if you need it. Otherwise its not a fix just a workaround. We need to fix implementation of `spread` (basically the code we generate in `instantiate_Spread`)

src/libasr/pass/simplifier.cpp
czgdp1807: I see, `merge` evaluates both before picking the right expression and what we need is update only one `true` expression is picked.

czgdp1807: Instead of using `merge` we should update by defining our own function. Something like, `new_shape`,

`new_shape(input_array, j, ncopies, i, dim)`

This invariant should work. See full code below,

```fortran
program debug
implicit none

integer :: array(5, 6, 7)
integer, allocatable :: spread_array(:, :, :, :)
integer :: j, ncopies, dim

dim = 3
j = 1
ncopies = 8
allocate(spread_array(&
         new_shape(array, j, ncopies, 1, dim),&
         new_shape(array, j, ncopies, 2, dim),&
         new_shape(array, j, ncopies, 3, dim),&
         new_shape(array, j, ncopies, 4, dim)))

print *, shape(spread_array)

contains

integer function new_shape(input_array, j, ncopies, i, dim)
integer :: input_array(:, :, :), i, ncopies, dim
integer, intent(out) :: j
    if( i .ne. dim ) then
        new_shape = size(input_array, j)
        j = j + 1
    else
        new_shape = ncopies
    end if
end function
end program
```

Output,

```zsh
5           6           8           7
```

We just need to generate `new_shape` in `Spread` namespace in `src/libasr/pass/intrinsic_array_function_registry.h` and then use it in `simplifier`. Should be doable and not that hard logic.

gxyd:
I'm working on the improvement suggested above.

Issue #4248, opened 2024-06-12
Implement compiler_version() and compiler_options() intrinsics


certik:
We'll have users for this feature:

https://fortran-lang.discourse.group/t/retrieve-compiler-info-at-run-time/8192

PR #5918, opened 2025-01-05
enh: Implement intrinsic `compiler_options`

HarshitaKalani:
Towards: https://github.com/lfortran/lfortran/issues/4248


HarshitaKalani:
```console
$ lfortran integration_tests/intrinsics_337.f90 --no-experimental-simplifier
--no-experimental-simplifier
```

certik:
The Fortran backend is printing the line incorrectly:
```
CMakeFiles/intrinsics_337.dir/intrinsics_337.f90.o.tmp.f90:37:31:

   37 | print *, compiler_options(use lfortran_intrinsic_iso_fortran_env, only: real_kinds
Error: Syntax error in argument list at (1)
CMakeFiles/intrinsics_337.dir/intrinsics_337.f90.o.tmp.f90:38:1:

   38 | )
      | 1
Error: Invalid character in name at (1)
The command 'gfortran -fno-backtrace -o CMakeFiles/intrinsics_337.dir/intrinsics_337.f90.o -c CMakeFiles/intrinsics_337.dir/intrinsics_337.f90.o.tmp.f90' failed.
```

src/libasr/compiler_options.h
```diff
@@ -0,0 +1,27 @@
...
+        this->options = options;
+    }
+
+    const std::vector<std::string>& get_options() const {
+        return options;
+    }
+
+private:
+    CompilerOptionsState() = default;
+    std::vector<std::string> options;
```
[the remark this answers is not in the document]
HarshitaKalani: Actually LFORTRAN_VERSION macro is defined at compile time in config.h. The use of the macro avoids linkage issues because it’s not a variable but a preprocessor static constant whereas compilerOptions is dynamic depending on the runtime arguments. That's the challenge here.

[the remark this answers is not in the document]
Pranavchiku: > Why not just do the same approach as LFORTRAN_VERSION?

I agree. @HarshitaKalani take a look how and where are we setting `LFORTRAN_VERSION`. In similar way, generate a string from `compiler_options` that you return as output and save it as `LCOMPILER_OPTIONS`. Then you can directly leverage `LCOMPILER_OPTIONS`.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -1218,6 +1220,43 @@ namespace CompilerVersion {
...
+        ASRUtils::ASRBuilder b(al, loc);
+
+        // const std::vector<std::string>& options = CompilerOptionsState::instance().get_options();
+        // std::string options_str = "";
+        // for (const auto& option : options) {
+        //     if (!options_str.empty()) options_str += " ";
+        //     if (option != "lfortran" && (option.size() < 4 || option.substr(option.size() - 4) != ".f90")) {
+        //         options_str += option;
+        //     } 
+        // }
```
Pranavchiku, line 1242: let's remove this

integration_tests/intrinsics_338.f90
```diff
@@ -0,0 +1,4 @@
+program intrinsics_338
+    use iso_fortran_env
+    print *, compiler_options()
```
Pranavchiku, line 3: test output of compiler options?

Pranavchiku (approved):
just few minor commits, else looks good.

src/libasr/pass/intrinsic_functions.h
HarshitaKalani: Oops yes

integration_tests/intrinsics_338.f90
HarshitaKalani:
```
$ lfortran integration_tests/intrinsics_338.f90  --no-experimental-simplifier
--no-experimental-simplifier 
```

Pranavchiku: I mean modify test by using `EXTRA_ARGS` and use `if ( xx ) error stop`

Issue #3806, opened 2024-04-07
TODO for beta


certik:
In order to reach beta quality, we need to implement the following items:

* [ ] All instrinsic functions working (compile time, runtime): @HarshitaKalani, @parth121101
* [ ] array_op redesign: @czgdp1807 
* [ ] https://github.com/lfortran/lfortran/issues/2257: @Shaikh-Ubaid 
* [ ] https://github.com/lfortran/lfortran/issues/3581: @Pranavchiku 
* [ ] robust deallocation (https://github.com/lfortran/lfortran/issues/3770)
* [ ] most OOP features
* [ ] `--fast` working in all cases
* [ ] compile at least 10 codes without any workaround (except the items listed below)

After we reach beta, we will tackle the following:

* coarrays
* submodules
* common block and other F77 features that should not be used in new codes (but we have to support to compile old codes)
* separate compilation (generate intermediate object files, link everything at the end)
* Other backends (WASM, C, C++)

thoth291:
I think that it would be better not to wait for everything to be done to make it work with Jupyter kernel. I can see how it can be useful to have a testing repo with Binder and Colab integration. That would help anyone to test recent version and report back issues if any discovered.

certik:
@thoth291 the Jupyter kernel already works. The above issue is to focus on ironic all bugs there, which we should do after we can compile standalone codes (i.e., after beta). If you discover a bug in the kernel, go ahead and report it, we fix them as we have time.

krystophny:
| Status | Track | Issues | PRs |
|---|---|---|---|
| needs review | ASR docs | [#7006](https://github.com/lfortran/lfortran/issues/7006) | [#9808](https://github.com/lfortran/lfortran/pull/9808) |
| needs review | M_framework | [#10885](https://github.com/lfortran/lfortran/issues/10885) | [#11796](https://github.com/lfortran/lfortran/pull/11796) |
| needs completion | Warnings | [#10605](https://github.com/lfortran/lfortran/issues/10605) | [#10696](https://github.com/lfortran/lfortran/pull/10696) |
| needs completion | Compiler options | [#7937](https://github.com/lfortran/lfortran/issues/7937) | [#8105](https://github.com/lfortran/lfortran/pull/8105) |
[...]

PR #5922, opened 2025-01-06
Fix: raise error on improper use of `class` keyword in declarations

### LINKED ISSUE #5872: using `class` keword for declaring variable should raise an error
MRE
```fortran
program main
   type tt
   end type tt
   class(tt) :: inst_tt
 end program main
```

```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt5.f90 
tt5.f90:4:23:

    4 |    class(tt) :: inst_tt
      |                       1
Error: CLASS variable ‘inst_tt’ at (1) must be dummy, allocatable or pointer

```

ejovo13:
Fix #5872

Todo:

- [ ] Add test in `tests/errors`

### DIFF
--- a/integration_tests/template_vector.f90
+++ b/integration_tests/template_vector.f90
@@ -59,7 +59,7 @@ contains
 
     subroutine main()
         instantiate vector_t(integer), only: IntVector => Vector
-        class(IntVector) :: v
+        type(IntVector) :: v
         integer :: i
         v = IntVector()
         ! call v%push_back(v, 1)
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3303,6 +3303,21 @@ public:
                         throw SemanticAbort();
                     }
                 }
+
+                bool is_class_type = sym_type->m_type == AST::decl_typeType::TypeClass; 
+                // Assert class declaration variable is dummy, pointer or allocatable
+                if (is_class_type) {
+                    if (!(is_argument | is_pointer | is_allocatable)) {
+                        diag.add((Diagnostic(
+                            "Class variable must be a dummy argument, allocatable, or pointer.",
+                            Level::Error, Stage::Semantic, {
+                                Label("",{x.base.base.loc})
+                            }
+                        )));
+                        throw SemanticAbort();
+                    }
+                }
+
                 if (s.n_dim > 0) {
                     if (dims.size() > 0) {
                         // This happens for:
--- a/tests/reference/asr-template_vector-140858c.json
+++ b/tests/reference/asr-template_vector-140858c.json
@@ -2,11 +2,11 @@
     "basename": "asr-template_vector-140858c",
     "cmd": "lfortran --show-asr --no-color {infile} -o {outfile}",
     "infile": "tests/../integration_tests/template_vector.f90",
-    "infile_hash": "23c3387fefdedf9e792fc7b6ddfec9102541f4501a300a3641871023",
+    "infile_hash": "c90086bccd7acd20362f5bcb800e4030a60fc54abf4ba105cc7ceef6",
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-template_vector-140858c.stdout",
-    "stdout_hash": "7d84da782c723456ced90da045b54ed65335ba698ba443e6909d6bac",
+    "stdout_hash": "89d5560d4d7cafae90ddffccb0aff6019cfe25e5565fc2f6e9e8ed63",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-template_vector-140858c.stdout
+++ b/tests/reference/asr-template_vector-140858c.stdout
@@ -639,7 +639,7 @@
                                                     ()
                                                     ()
                                                     Default
-                                                    (ClassType
+                                                    (StructType
                                                         7 intvector
                                                     )
                                                     ()


integration_tests/template_vector.f90
```diff
@@ -59,7 +59,7 @@ subroutine resize(this, n)
 
     subroutine main()
         instantiate vector_t(integer), only: IntVector => Vector
-        class(IntVector) :: v
+        type(IntVector) :: v
```
certik, line 62: Above it says `class(Vector)`, but here it is required `type(IntVector)`? That seems inconsistent.

What are the rules about when you use `class` and when `type`?

tests/errors/class_declaration_01.f90
```diff
@@ -0,0 +1,7 @@
+! https://github.com/lfortran/lfortran/issues/5872
+program main
+   type tt
+   end type tt
+   ! Class declaration must be dummy, allocatable, or pointer
+   class(tt) :: inst_tt
+ end program main
```
certik, line 7: We recently switched to add all error messages into `tests/errors/continue_compilation_*.f90` files and each error must be recoverable. Go ahead and add the above error in there and update tests via `./run_tests.py -u`.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3303,6 +3303,21 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         throw SemanticAbort();
                     }
                 }
+
+                bool is_class_type = sym_type->m_type == AST::decl_typeType::TypeClass;
```
certik, line 3307: Use `AST::is_a<>`, see other places in the file for examples how to use it.

certik:
Awesome, thanks for adding the error message. I left some comments.

integration_tests/template_vector.f90
ejovo13: I am a complete novice so take my words with a grain of salt but what I've gathered is that you can only use the `class` keyword for "polymorphic" variables that will be "resolved" at runtime: 
- when used as an argument (i.e. dummy variable) in a procedure
- when tagged with the `pointer` attribute
- when tagged with the `allocatable` attribute

You can't (as in gfortran disallows this, I'm not sure what the precise verbiage of the Fortran standards has to say) just use `class` when declaring a new concrete variable.

Above in the file the class keyword is used to declare the polymorphic type of the procedure's arguments. Doing so allows the procedure to accept types that are extended from `IntVector`. In the body of the `main` subroutine we must use `type` instead of `class` because we are dealing with a local variable that's _not_ a part of the routine's signature.

See [this](https://stackoverflow.com/questions/29652620/difference-between-type-and-class-in-fortran-2003) stack overflow answer for more detail.


#### Examples

The following subroutine compiles with gfortran:

```fortran
    subroutine some_subroutine(self)
        
        ! "Polymorphic" variable
        class(my_type) :: self

        ! Local variable
        type(my_type) :: local_var

        ! allocatable
        class(my_type), allocatable :: allocatable_var

        ! pointer
        class(my_type), pointer :: pointer_var

    end subroutine
```

Whereas the following subroutine does not:

```fortran
    subroutine some_subroutine(self)
        
        ! "Polymorphic" variable, ok
        class(my_type) :: self

        ! "Concrete" local variable can't be declared with class!
        class(my_type) :: not_valid

    end subroutine

```

```
(lf-build) ejovo@envy:~/Programming/Fortran/lfortran$ gfortran -c ./b_class.f90
./b_class.f90:24:35:

   24 |         class(my_type) :: not_valid
Error: CLASS variable ‘not_valid’ at (1) must be dummy, allocatable or pointer
```

src/lfortran/semantics/ast_common_visitor.h
ejovo13: Sounds good, I'll replace it with `AST::is_a<>`. I actually just copied the `is_char_type` declaration from a little higher up in the function ;). 

While I have you here, if you have a second could you try and explain what the `AST::down_cast<>` functions do? I'll admit I kept jumping back and forth between the ASDL declaration, it's generated code (both the Types enum and the `*_t` structs, and it's usage within the visitor code and I was getting really confused about what types I was actually using in the client code.

ejovo13: I believe that there is no way to use `AST::is_a<>` instead because in this case the ASDL node has no constructors.

AST.asdl:
```
decl_type
    = TypeClass
    | TypeCharacter
    | TypeComplex
    | TypeDoublePrecision
    | TypeDoubleComplex
    | TypeInteger
    | TypeLogical
    | TypeProcedure
    | TypeReal
    | TypeType
```
Generated code:

```
enum decl_typeType // Simple Sum
{ // Types
    TypeClass, TypeCharacter, TypeComplex, TypeDoublePrecision, TypeDoubleComplex, TypeInteger, TypeLogical, TypeProcedure, TypeReal, TypeType
};
```

Other places in the file that use `is_a<>` are exclusively dealing with the actual Constructor types:
```
...
            if (AST::is_a<AST::AttrType_t>(*args[i])) {
...
            } else if (AST::is_a<AST::AttrNamelist_t>(*args[i])) {
```
```
                } else if (AST::is_a<AST::AttrEquivalence_t>(*x.m_attributes[i])) {
```

whose template parameters are actual structs and not just enum values:
```
    struct AttrEquivalence_t // Constructor
    {
        const static decl_attributeType class_type = decl_attributeType::AttrEquivalence;
        typedef decl_attribute_t parent_type;
        decl_attribute_t base;
        equi_t* m_args; size_t n_args; // Sequence
    };
```

I'm going to leave it as is (following the `is_char_type` idiom that appears in the same function):
```
                bool is_char_type = sym_type->m_type == AST::decl_typeType::TypeCharacter;
```

### Remark

Note that when we are operating on the _ASR_, we should indeed favor to use the `is_a<>` template function because the ASDL includes _constructors_ for each variable type:

`ast_common_visitor` around line 2885:
```
                            bool is_char_type = ASR::is_a<ASR::String_t>(*v->m_type);
```

Which works because the `ASR.asdl` grammar is different:
```
ttype
    = Integer(int kind)
    | UnsignedInteger(int kind)
    | Real(int kind)
    | Complex(int kind)
    | String(int kind, int len, expr? len_expr, string_physical_type physical_type)
    | Logical(int kind)
    | Set(ttype type)
    | List(ttype type)
...
```

A little long winded and rambling, but I hope what I've written is clear.

ejovo13:
Ok I added some tests. 

LLVM generation failed when trying to _execute_ this subroutine (which I believe is normal, as OOP features haven't been fully implemented #3806):

```fortran
subroutine some_subroutine(self)
    
    ! _Polymorphic_ variable
    class(my_type) :: self

    ! Local variable
    type(my_type) :: local_var

    ! allocatable
    class(my_type), allocatable :: allocatable_var

    ! pointer
    class(my_type), pointer :: pointer_var

end subroutine
```

So I only declared the subroutine in `tests/class_declarations.f90` removed the program that called it.

As a new contributor it wasn't obvious to me that I was supposed to add test configurations to `tests/test.toml`. It might be a good idea to open up a new ticket to document adding new tests (i.e. errors in `continue_compile_*.f90`, adding tests to `test.toml`, how the options change the testing behavior, etc).

Let me know if things look good or if there's anything you'd like me to change!

tests/errors/class_declaration_01.f90
ejovo13: I added some tests, see the commits below. I'm not entirely sure what you mean (in terms of implementation) by 
> each error must be recoverable
I'm assuming that the error I've added will only be thrown after the entire source file has been parsed (AST) and only once converting to ASR by adding in semantic information/checks. Is that recoverable? LMK.

tests/reference/asr-continue_compilation_4-246178a.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Class variable must be a dummy argument, allocatable, or pointer.
+ --> tests/errors/continue_compilation_4.f90:6:4
+  |
+6 |    class(tt) :: inst_tt
+  |    ^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
ejovo13: I think it'd be really nice to provide a more descriptive explanation of _why_ it's an error to use the `class` keyword outside of a polymorphic context, but I'm having trouble coming up with a clean explanation inline. Here's my attempt so far:

```
semantic error: Class variable must be a dummy argument, allocatable, or pointer.
  --> ./tests/errors/continue_compilation_5.f90:13:22
   |
13 |         class(tt) :: local_var
   |                      ^^^^^^^^^ 
   |
13 |         class(tt) :: local_var
   |         ~~~~~~~~~ Replace class(tt) with type(tt) if the type of local_var is known at compile time. For polymorphic contexts, add an 'allocatable' or 'pointer' attribute instead.
```
While the error message "Class variable must be a dummy argument, allocatable, or pointer" is technically sufficient, it doesn't really explain _why_ using `class` instead of `type` is a problem. I can easily imagine someone reporting this error message as unclear or confusing (as evidenced by your own confusion in the above discussion as well as my own).

Adding a second label starts to clutter the output but I retain that the one-line description is not clear enough.

### Proposal

What do you think of adding an index of error codes like [rustc](https://doc.rust-lang.org/error_codes/error-index.html) and an `--explain` flag to lfortran to provide a lot more context and detailed explanations for various errors? One of the most useful features of a compiler is having excellent error reporting and I think it would be interesting and powerful feature to explore.

#### Prior Art

Rust's error reporting

```
error[E0623]: lifetime mismatch
 --> main.rs:8:9
  |
8 |         Foo { x: "hello" }
  |         ^^^ lifetime `'static` required
  |
  = help: consider giving it a 'static lifetime
  = note: for more information, see https://doc.rust-lang.org/error-index.html#E0623
```

ejovo13:
Thanks for the comments. I've been bouncing around the airport and hotel with delayed, canceled flights. I'll have this cleaned up and ready to merge when I've landed tomorrow. Cheers.

PR #5925, opened 2025-01-06
enh: handling of 'dble', 'float', 'dfloat', 'shifta' as intrinsics

### LINKED ISSUE #5923: interface conflicts with intrinsic procedure

For the below fortran program with use of intrinsic procedure `dble` and an interface `dble` defined in it:
```f90
MODULE M_strings
implicit none

interface dble
   module procedure dble_s2v
end interface

CONTAINS

   subroutine a2d()
      integer                      :: intg
      print *, dble(intg)
   end subroutine a2d

   doubleprecision function dble_s2v()
   end function dble_s2v

end module M_strings


```

On it's execution with LFortran, raises the below error:
```console
> lfortran --version
LFortran version: 0.43.0-34-gb55e36587-dirty
Platform: macOS ARM
LLVM: 11.1.0

> lfortran test.f90
semantic error: Function 'dble' not found (not user defined nor intrinsic)
  --> test.f90:12:16
   |
12 |       print *, dble(intg)
   |                ^^^^^^^^^^ 


```

gxyd:
## Description

Fixes #5923

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -7066,6 +7066,9 @@ public:
         return nullptr;
     }
 
+    // special handling of 'dble', 'float', 'dfloat', 'shifta' intrinsics
+    // maybe once those are moved to IntrinsicElementalFunction, this might
+    // not be needed
     ASR::asr_t* handle_intrinsics_dble_float_dfloat_shifta(
         const AST::FuncCallOrArray_t &x,
         Allocator &al
@@ -7664,6 +7667,16 @@ public:
                     throw SemanticAbort();
                 }
             } else if (compiler_options.implicit_interface) {
+
+                bool is_function = true;
+                // NOTE: ideally this shouldn't be needed, this is only to handle
+                // 'dble', 'shifta', 'float', 'dfloat', which aren't currently
+                // implemented as intrinsic elemental function
+                intrinsic_as_node(x, is_function);
+                if (!is_function) {
+                    return;
+                }
+
                 // If implicit interface is allowed, we have to handle the
                 // following case here:
                 // real :: x
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -5141,6 +5141,10 @@ public:
                     false);
         if( idx == -1 ) {
             bool is_function = true;
+            if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+                tmp = result;
+                return tmp;
+            }
             v = intrinsic_as_node(x, is_function);
             if( !is_function ) {
                 return tmp;
@@ -5255,6 +5259,10 @@ public:
                     false);
             if( idx == -1 ) {
                 bool is_function = true;
+                if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+                    tmp = result;
+                    return tmp;
+                }
                 v = intrinsic_as_node(x, is_function);
                 if( !is_function ) {
                     return tmp;
@@ -7061,6 +7069,29 @@ public:
         return nullptr;
     }
 
+    ASR::asr_t* handle_intrinsics_dble_float_dfloat_shifta(
+        const AST::FuncCallOrArray_t &x,
+        Allocator &al
+    ) {
+        ASR::asr_t* asr_node { nullptr };
+        std::string var_name = to_lower(x.m_func);
+        Vec<ASR::call_arg_t> args;
+        if (var_name == "dble") {
+            visit_expr_list(x.m_args, x.n_args, args);
+            asr_node = handle_intrinsic_dble(al, args, x.base.base.loc);
+        } else if (var_name == "float" ) {
+            visit_expr_list(x.m_args, x.n_args, args);
+            asr_node = handle_intrinsic_float_dfloat(al, args, x.base.base.loc, 4);
+        } else if (var_name == "dfloat" ) {
+            visit_expr_list(x.m_args, x.n_args, args);
+            asr_node = handle_intrinsic_float_dfloat(al, args, x.base.base.loc, 8);
+        } else if (var_name == "shifta") {
+            visit_expr_list(x.m_args, x.n_args, args);
+            asr_node = create_Shifta(x.base.base.loc, args);
+        }
+        return asr_node;
+    }
+
     ASR::asr_t* handle_intrinsic_dble(Allocator &al, Vec<ASR::call_arg_t> args,
                                         const Location &loc) {
         ASR::expr_t *arg = nullptr, *value = nullptr;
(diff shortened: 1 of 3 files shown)


Pranavchiku (commented):
I think the way it should be fixed is by understanding why the following works in latest main

```fortran
program main
real :: x
x = 1.0
print *, dble()
CONTAINS
function dble() result(x)
real :: x
x = 129.0
end function dble
end program
```

When we visit `print *, dble()` it automatically resolves `dble` with line `v = current_scope->resolve_symbol(var_name);` and hence it never goes inside `if (!v || (v && (is_external_procedure || is_explicit_intrinsic)))`

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7581,25 +7612,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                Vec<ASR::call_arg_t> args;
-                visit_expr_list(x.m_args, x.n_args, args);
-                tmp = handle_intrinsic_float_dfloat(al, args, x.base.base.loc, 8);
-                return;
-            } else if (var_name == "shifta") {
-                Vec<ASR::call_arg_t> args;
-                visit_expr_list(x.m_args, x.n_args, args);
-                tmp = create_Shifta(x.base.base.loc, args);
+            if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+                tmp = result;
```
Pranavchiku, line 7616: Well first things first, this refactoring misguides reviewer to think that changing this might have fixed the code. That is what happened, but it has nothing to do with fix. Please do not mix changes, especially when fixing it or put a comment that states it is just refactoring on github.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5141,6 +5141,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     false);
         if( idx == -1 ) {
             bool is_function = true;
+            if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+                tmp = result;
+                return tmp;
+            }
```
Pranavchiku, line 5147: I would not do it here rather make it a part of `intrinsic_as_node`. 

```diff
     ASR::symbol_t* intrinsic_as_node(const AST::FuncCallOrArray_t &x,
                                      bool& is_function) {
         std::string var_name = to_lower(x.m_func);
         std::string specific_var_name = var_name;
         bool is_specific_type_intrinsic = intrinsic_mapping.count(var_name);
+        if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+            return ASR::down_cast<ASR::symbol_t>(result);
+        }
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5255,6 +5259,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     false);
             if( idx == -1 ) {
                 bool is_function = true;
+                if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+                    tmp = result;
+                    return tmp;
+                }
```
Pranavchiku, line 5265: same comment as above

Pranavchiku (changes requested):
left a few comments above

gxyd:
> When we visit `print *, dble()` it automatically resolves `dble` with line `v = current_scope->resolve_symbol(var_name);` and hence it never goes inside `if (!v || (v && (is_external_procedure || is_explicit_intrinsic)))`

That's right, a user-defined function takes higher priority compared to an intrinsic procedure of the same name. That's why this is indeed correct.

src/lfortran/semantics/ast_common_visitor.h
gxyd: That's a little debatable. It doesn't make reviewing the changes a little difficult for sure. Locally, when I made the changes, I first re-used the code and once that started working, I put those staged those changes (but didn't commit, probably I should've done that).

src/lfortran/semantics/ast_common_visitor.h
gxyd: > I would not do it here rather make it a part of intrinsic_as_node.

Fair enough.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7581,25 +7612,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                return;
-            } else if (var_name == "dfloat" ) {
-                Vec<ASR::call_arg_t> args;
-                visit_expr_list(x.m_args, x.n_args, args);
-                tmp = handle_intrinsic_float_dfloat(al, args, x.base.base.loc, 8);
-                return;
-            } else if (var_name == "shifta") {
-                Vec<ASR::call_arg_t> args;
-                visit_expr_list(x.m_args, x.n_args, args);
-                tmp = create_Shifta(x.base.base.loc, args);
```
kmr-srbh, line 7602: Why do we handle these intrinsics specially? The issue which the PR fixes does not occur for other intrinsics like `int` or `real`. Currently we use type casting for the intrinsics `dble`, `float` and `dfloat` while others like `int` are handled as `IntrinsicElementalFunction`. Similarly `shiftl` and `shiftr` are handled as `IntrinsicElementalFunction` but `shifta` is implemented through `IntegerBinOp` using `BitRShift`.

For now, the current approach is fine, but do we probably need to implement these intrinsics through the new approach?

kmr-srbh: I saw the code through the `Files changed` section, so did not get to know that a review has occurred. Please disregard any comment that sounds out-of-context.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: > It does make reviewing the changes a little difficult for sure. 

It changes the entire context. So if you see my comment, unless I drill down into it, I have the intuition that code flows into this which means somehow `v` is nullptr which should not be the case, else it is a different bug.

Now if I drill down, then I understand, no, it is not. It means this PR needs a considerable amount of time to review which is absolutely fine but for a fix that only needs addition of 9 lines I think it is not good.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: > we probably need to implement these intrinsics through the new approach?

Yes, if I remember correctly we came to the conclusion that we will right away add a cast node for these but then for cases like `int` and `real` the argument was that those are elemental and thus wrapping those with `IntrinsicElementalFunction` automatically handles the broadcast. Same is with `dble` and hence I think we should turn these to `IntrinsicElementalFunction`.

src/lfortran/semantics/ast_common_visitor.h
gxyd: > code flows into this which means somehow v is nullptr which should not be the case

I don't think the logic I added does anything with that. It might non-nullptr and still enter this, and I'm not sure if my logic changes any of that anyway.

Pranavchiku: > It might non-nullptr and still enter this,

only if it is an external procedure or intrinsic, which it is not. So, no :)

> I'm not sure if my logic changes any of that anyway.

Exactly that is when it is refactoring. Neither I am telling it changes anything. Simply, please do not add refactoring changes with fixes it messes up the thought process behind code review.

Pranavchiku (approved):
The cleanup pushed in latest commits is good. Thanks for doing this!

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6897,6 +6897,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
     ASR::symbol_t* intrinsic_as_node(const AST::FuncCallOrArray_t &x,
                                      bool& is_function) {
         std::string var_name = to_lower(x.m_func);
+        if (ASR::asr_t* result = handle_intrinsics_dble_float_dfloat_shifta(x, al)) {
+            is_function = false;
+            tmp = result;
+            return nullptr;
```
Pranavchiku, line 6903: Why are we returning a nullptr? This may cause trouble in cases where `tmp` is not used and rather output of this is used. As far as I remember the diff was to return  `result`. @gxyd please address this in subsequent PRs.

gxyd: > As far as I remember the diff was to return result

Ah yes, but after a bit of thinking and how `intrinsic_as_node` is written, it made more sense to return `nullptr`.

Please see this: https://github.com/lfortran/lfortran/blob/main/src/lfortran/semantics/ast_common_visitor.h#L7017 in `intrinsic_as_node`, which shows that `nullptr` is returned most of the times except for the case when `is_intrinsic_registry_function(var_name)` isn't true.

Hence, I decided to return `nullptr` in that case. Maybe the logic can be improved, but I didn't divert from how it's currently already done.

gxyd: Ah yes, a better answer to `return nullptr` is, since there is no *Symbol* for an intrinsic function to return (e.g. in cases of `aint`, `complex`, `log` etc.), so we simply can't return a symbol and instead specify that `is_function = false` and hence the current approach isn't a bad one actually.

gxyd: Let me know what you think @Pranavchiku

Pranavchiku: Makes sense, yes.

PR #5927, opened 2025-01-06
CI: Update CI for PRIMA

jinangshah21:
With this `bobyqa_example_1` and `bobyqa_example_2` works well and give same output as GFortran.

PR #5937, opened 2025-01-07
CI: Update CI for PRIMA

jinangshah21:
With this `bobyqa_example_1` and `newuoa_example_1` gives same output as GFortran.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -503,10 +503,11 @@ jobs:
         run: |
             git clone https://github.com/Pranavchiku/prima.git
             cd prima
-            git checkout -t origin/lf25
-            git checkout 811948c319aa8f05e6cfb2faedc0334124d77670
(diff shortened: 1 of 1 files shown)


jinangshah21:
`bobyqa_example_2` in #5927 is having some issues, I will check it.

Pranavchiku (dismissed):
remove tests for `nf`, `f` and only check values of `x`.

gxyd (approved):
> remove tests for `nf`, `f` and only check values of `x`.

Seems like the *Test PRIMA* CI job does exactly that now. Looks good to me.

Pranavchiku (commented):
Looks good, thanks for review and merging this!

jinangshah21:
For `bobyqa_example_1` nf is working properly in current main also.

jinangshah21:
Also I have corrected some workarounds and correct value of nf is 22. LFortran also gives 22 which is same as Gfortran with PRIMA main. I will that the nf test in next PR.

PR #5856, opened 2024-12-28
fix: tokenizer errors generates ast and asr with cc flag

parth121101:

PR #5940, opened 2025-01-08
fix: tokenizer errors generates ast and asr with cc flag - 2

parth121101:
This PR resolves many issues that we were facing in PR : #5856 . The only problem left here is it's throwing error twice for some of the error tests.

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -439,7 +439,7 @@ int emit_tokens(const std::string &infile, bool line_numbers, const CompilerOpti
             compiler_options.fixed_form, include_dirs);
     }
     auto res = LCompilers::LFortran::tokens(al, input, diagnostics, &stypes, &locations,
-        compiler_options.fixed_form);
+        compiler_options.fixed_form, compiler_options.continue_compilation);
     lm.init_simple(input);
     lm.file_ends.push_back(input.size());
     std::cerr << diagnostics.render(lm, compiler_options);
--- a/src/lfortran/parser/parser_exception.h
+++ b/src/lfortran/parser/parser_exception.h
@@ -11,6 +11,8 @@ namespace parser_local {
     // It is not propagated outside.
     // This file is included in parser.tab.cc (via semantics.h)
     // And in parser.cpp. Nowhere else.
+    class TokenizerAbort {
+    };
 
     class TokenizerError
     {
--- a/src/lfortran/parser/tokenizer.re
+++ b/src/lfortran/parser/tokenizer.re
@@ -5,6 +5,11 @@
 #include <lfortran/parser/parser.tab.hh>
 #include <libasr/bigint.h>
 
+using LCompilers::diag::Level;
+using LCompilers::diag::Stage;
+using LCompilers::diag::Label;
+using LCompilers::diag::Diagnostic;
+
 namespace LCompilers::LFortran {
 
 void Tokenizer::set_string(const std::string &str)
@@ -262,11 +267,7 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
                     }));
 
                 if(!continue_compilation) {
-                    throw parser_local::TokenizerError(diag::Diagnostic(
-                        "Token '" + t + "' is not recognized",
-                        diag::Level::Error, diag::Stage::Tokenizer, {
-                        diag::Label("token not recognized", {loc})
-                    }));
+                    throw parser_local::TokenizerAbort();
                 } else {
                     continue;
                 }
--- a/src/lfortran/parser/tokenizer.re
+++ b/src/lfortran/parser/tokenizer.re
@@ -144,7 +144,7 @@ void Tokenizer::add_rel_warning(diag::Diagnostics &diagnostics, bool fixed_form,
     }
 }
 
-int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnostics &diagnostics)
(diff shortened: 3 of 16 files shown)


src/lfortran/parser/tokenizer.re
```diff
@@ -255,12 +255,21 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
...
-                    })
-                );
+                    }));
+
+                if(!continue_compilation) {
+                    throw parser_local::TokenizerError(diag::Diagnostic(
+                        "Token '" + t + "' is not recognized",
+                        diag::Level::Error, diag::Stage::Tokenizer, {
+                        diag::Label("token not recognized", {loc})
+                    }));
```
Pranavchiku, line 269: The reason is you are adding to diagnostics and also throwing the error. Hence it throws twice. By default at the end we do `diagnostics.render()` which prints all the errors.

tests/reference/ast-esub2-248f0df.stderr
```diff
@@ -3,3 +3,9 @@ tokenizer error: Token '~' is not recognized
   |
 3 | x = ~1
   |     ^ token not recognized
+
+tokenizer error: Token '~' is not recognized
+ --> tests/errors/esub2.f90:3:5
+  |
+3 | x = ~1
+  |     ^ token not recognized
```
Pranavchiku, line 11: Shall be fixed by comment above

src/bin/lfortran.cpp
```diff
@@ -438,7 +438,7 @@ int emit_tokens(const std::string &infile, bool line_numbers, const CompilerOpti
             compiler_options.fixed_form, include_dirs);
     }
     auto res = LCompilers::LFortran::tokens(al, input, diagnostics, &stypes, &locations,
-        compiler_options.fixed_form);
+        compiler_options.fixed_form, compiler_options.continue_compilation);
```
Pranavchiku, line 441: I do not think you need this now. Just assure if `p.continue_compilation` is set and everything will be fine.

Pranavchiku (changes requested):
left a few comments above

src/bin/lfortran.cpp
Pranavchiku: Can you not use `po.continue_compilation`? The only place you need to check and skip throwing error is inside `int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnostics &diagnostics)` which you call from `int yylex(LCompilers::LFortran::YYSTYPE *yylval, YYLTYPE *yyloc,
    LCompilers::LFortran::Parser &p)` so just pass `p.continue_compilation` inside `Tokenizer::lex`. That would do.

parth121101:
I think now this PR is ready for a review.
Cc: @certik @Pranavchiku

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -10,44 +10,56 @@ syntax error: kind 16 is not supported yet.
...
+   --> tests/errors/continue_compilation_2.f90:325:17
+    |
+325 |     integer  :: ?
+    |                 ^ token not recognized
+
+syntax error: Newline is unexpected here
+   --> tests/errors/continue_compilation_2.f90:325:18
+    |
+325 |     integer  :: ?
+    |                  ^
```
Pranavchiku, line 23: the reason for two error is because we still tokenize this and hence it also gives us syntax error,

src/bin/lfortran.cpp
Pranavchiku: @parth121101 can you please address this?

tests/reference/asr-continue_compilation_2-a6145a1.stderr
parth121101: So is this an expected behaviour ?
I think yes

src/bin/lfortran.cpp
parth121101: @Pranavchiku the function where I have added a new parameter `LCompilers::LFortran::tokens` is present in parser.cpp
This function directly calls `int Tokenizer::lex` function and not via `int yylex`

Pranavchiku:
I think this is good otherwise

src/bin/lfortran.cpp
Pranavchiku: I’ll check this thoroughly, we can do it in subsequent PRs

tests/reference/asr-continue_compilation_2-a6145a1.stderr
parth121101: Resolved this

src/lfortran/parser/parser_exception.h
```diff
@@ -2,6 +2,7 @@
 #define LFORTRAN_PARSER_PARSER_EXCEPTION_H
 
 #include <libasr/exception.h>
+#include <iostream>
```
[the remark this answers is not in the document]
parth121101: Done

parth121101:
@Pranavchiku We can merge this now

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -171,6 +177,15 @@ semantic error: Kind 16 is not supported for Real
 322 |     real*16 :: unsupported_kind
     |     ^^^^^^^ 
 
+semantic error: Symbol is already declared in the same scope
```
Pranavchiku, line 180: I think this is creating a symbol x, let’s use a variable name that does not conflict

parth121101: sorted

Pranavchiku (approved):

PR #5941, opened 2025-01-08
chore: convert 'FLUSH' intrinsic to 'FLUSH' statement

gxyd:
## Description

'FLUSH' intrinsic and 'FLUSH' statement have identifical effect. Other compilers ifx, ifort, gfortran support 'FLUSH' intrinsic as well

We raise a warning to encourage users to use 'FLUSH' statement instead of 'FLUSH' intrinsic.

Fixes https://github.com/lfortran/lfortran/issues/5709

### DIFF
--- a/integration_tests/flush_02.f90
+++ b/integration_tests/flush_02.f90
@@ -1,0 +1,8 @@
+program flush_02
+    implicit none
+    open(10, file="file_01_data.txt")
+    open(11, file="file_02_data.txt")
+    call FLUSH()
+    close(10)
+    close(11)
+end program flush_02
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3280,13 +3280,6 @@ public:
     */
     void flush_subroutine_to_flush_statement(const AST::SubroutineCall_t &x) {
         LCOMPILERS_ASSERT(to_lower(x.m_name) == "flush")
-        // encourage users to use `FLUSH` statement instead,
-        // as 'FLUSH' subroutine call isn't in Fortran standard
-        diag.semantic_warning_label(
-            "The 'flush' intrinsic function is an LFortran extension",
-            {x.base.base.loc},
-            "help: use the 'flush' statement instead"
-        );
 
         // for FLUSH subroutine call, the only argument 'unit' is optional
         ASR::expr_t* unit_flush = nullptr;
@@ -3318,7 +3311,7 @@ public:
             );
         } else {
             diag.add(Diagnostic(
-                "Expected 0 or 1 argument, instead got " + std::to_string(x.n_args + x.n_keywords) +
+                "Expected 0 or 1 arguments, got " + std::to_string(x.n_args + x.n_keywords) +
                 " arguments instead.",
                 Level::Error, Stage::Semantic, {
                     Label("",{x.base.base.loc})
@@ -3329,6 +3322,13 @@ public:
         tmp = ASR::make_Flush_t(
             al, x.base.base.loc, 0, unit_flush, nullptr,
             nullptr, nullptr
+        );
+        // encourage users to use `FLUSH` statement instead,
+        // as 'FLUSH' subroutine call isn't in Fortran standard
+        diag.semantic_warning_label(
+            "The 'flush' intrinsic function is an LFortran extension",
+            {x.base.base.loc},
+            "help: use the 'flush' statement instead"
         );
     }
 
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3297,40 +3297,25 @@ public:
             visit_expr(*x.m_keywords[0].m_value);
             unit_flush = ASRUtils::EXPR(tmp);
         } else if (x.n_args == 0 && x.n_keywords == 0) {
-            // when no argument to 'FLUSH' subroutine call is specified,
-            // we assign `OUTPUT_UNIT` from 'iso_fortran_env'
-            std::string output_unit_str = "output_unit";
-            std::string current_sym_name = ASRUtils::symbol_name(
-                ASR::down_cast<ASR::symbol_t>(current_scope->asr_owner)
-            );
-            std::string mangled_output_unit_name = current_scope->get_unique_name(
-                "__external_" + output_unit_str + "_" + current_sym_name, "false"
-            );
-            std::string iso_fortran_env_module_str = "lfortran_intrinsic_iso_fortran_env";
-            SymbolTable *tu_symtab = current_scope;
-            while (tu_symtab->parent != nullptr) {
-                tu_symtab = tu_symtab->parent;
-            }
-            ASR::Module_t* iso_fortran_env_module = ASRUtils::load_module(
-                al, tu_symtab, iso_fortran_env_module_str, x.base.base.loc,
-                true, compiler_options.po, true,
-                [&](const std::string &/*msg*/, const Location &/*loc*/) { }, lm
-            );
-
-            ASR::symbol_t* module_output_unit_sym = iso_fortran_env_module->m_symtab->resolve_symbol(output_unit_str);
-
-            ASR::symbol_t* output_unit_sym = ASR::down_cast<ASR::symbol_t>(
-                ASR::make_ExternalSymbol_t(
-                    al, x.base.base.loc, current_scope, s2c(al, mangled_output_unit_name),
-                    module_output_unit_sym, s2c(al, iso_fortran_env_module_str), nullptr,
-                    0, s2c(al, output_unit_str), ASR::accessType::Private
+            // when 'FLUSH' intrinsic is called with no argument,
+            // then all open units need to be 'FLUSH'ed
+            ASR::ttype_t* int_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4));
+            ASR::expr_t* one = ASRUtils::EXPR(
+                ASR::make_IntegerConstant_t(
+                    al, x.base.base.loc, 1, int_type
                 )
             );
-            current_scope->add_symbol(mangled_output_unit_name, output_unit_sym);
-            ASR::expr_t* output_unit_expr = ASRUtils::EXPR(
-                ASR::make_Var_t(al, x.base.base.loc, output_unit_sym)
+            ASR::expr_t* minus_one = ASRUtils::EXPR(
+                ASR::make_IntegerConstant_t(
+                    al, x.base.base.loc, -1, int_type
+                )
             );
-            unit_flush = output_unit_expr;
+            // we assign an integer -1 unit to that
+            unit_flush = ASRUtils::EXPR(
+                ASR::make_IntegerUnaryMinus_t(
+                    al, x.base.base.loc, one, int_type, minus_one
+                )
+            );
         } else {
             diag.add(Diagnostic(
                 "Expected 0 or 1 argument, instead got " + std::to_string(x.n_args + x.n_keywords) +
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3278,14 +3278,14 @@ public:
     /*
         Function to convert 'FLUSH' subroutine call to 'FLUSH' ASR node
     */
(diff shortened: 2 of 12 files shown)


tests/flush2.f90
```diff
@@ -0,0 +1,6 @@
+program flush2
+    implicit none
+    call flush()
```
gxyd, line 3: this call currently raises a runtime error, since `FLUSH(OUTPUT_UNIT)` (i.e. `FLUSH` statement with `OUTPUT_UNIT` as unit) raises runtime error:
```console
Specified UNIT 6 in FLUSH is not connected.
```

I think that's anyway a separate issue, since GFortran doesn't raise a runtime error, if needed we can have a separate issue for it.

gxyd: I'm not sure if the Fortran standard specifies anything on this, for what should be it's behavior.

gxyd: Seems like, ifx doesn't error either: https://godbolt.org/z/da46h3sh8

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3275,6 +3275,72 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        if (x.n_args == 1 && x.n_keywords == 0) {
+            visit_expr(*x.m_args[0].m_end);
+            unit_flush = ASRUtils::EXPR(tmp);
+        } else if (x.n_args == 0 && x.n_keywords == 1) {
+            visit_expr(*x.m_keywords[0].m_value);
+            unit_flush = ASRUtils::EXPR(tmp);
+        } else if (x.n_args == 0 && x.n_keywords == 0) {
+            // when no argument to 'FLUSH' subroutine call is specified,
+            // we assign `OUTPUT_UNIT` from 'iso_fortran_env'
+            std::string output_unit_str = "output_unit";
```
kmr-srbh, line 3302: This will cause an issue if we call `flush` without any argument more than once. Let's check for the presence of this symbol before importing.

kmr-srbh, line 3302: Let's mangle this name, so it does not interfere with some other symbol named `output_unit`.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3275,6 +3275,72 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            std::string iso_fortran_env_module_str = "lfortran_intrinsic_iso_fortran_env";
+            SymbolTable *tu_symtab = current_scope;
+            while (tu_symtab->parent != nullptr) {
+                tu_symtab = tu_symtab->parent;
+            }
+            ASR::symbol_t* module_sym = (ASR::symbol_t*)(ASRUtils::load_module(
```
kmr-srbh, line 3308: Let's check for the module also before importing it.

kmr-srbh (changes requested):

src/lfortran/semantics/ast_body_visitor.cpp
kmr-srbh, line 3308: Please use `down_cast`.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3275,6 +3275,72 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                al, tu_symtab, iso_fortran_env_module_str, x.base.base.loc,
+                true, compiler_options.po, true,
```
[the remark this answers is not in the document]
gxyd: No, I don't think so, the idea behind that logic is defined in a comment above it, which says: https://github.com/lfortran/lfortran/blob/8a46b442c7067be9720b9a7ac9aca29db29c7c55/src/libasr/asr_utils.cpp#L315-L316

when someone passes `intrinsic = false`, it first tries searching it as a regular module, and if that didn't workout, it explicitly searches it as an intrinsic module. In my case, I already know it's an intrinsic module (hence `intrinsic = true`) and it will already correctly find and load the module here above it: https://github.com/lfortran/lfortran/blob/8a46b442c7067be9720b9a7ac9aca29db29c7c55/src/libasr/asr_utils.cpp#L313-L314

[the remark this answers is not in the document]
gxyd: also, requiring `intrinsic = false` would mean that there is something wrong in the logic of `load_module` as it intuitively doesn't make sense for `iso_fortran_env` module to require `intrinsic = false`, as it's an intrinsic module.

src/lfortran/semantics/ast_body_visitor.cpp
gxyd: Ah yes, makes sense.

src/lfortran/semantics/ast_body_visitor.cpp
gxyd: `down_cast` is needed when a generic `symbol` is returned and we wanted to *cast* it to a specific type (e.g. `Module`), in this case a `Module` is returned and instead we want to upcast it to `symbol` instead.

So, I'm doing the opposite of down casting here. On second thought, I shouldn't be doing anything here, neither down-casting nor upcasting as below, I anyway use it as a `Module`. I'll revert this.

tests/flush2.f90
gxyd: I messed it up, the meaning of `call FLUSH()`, i.e. without the optional argument is:
```
    Flushes Fortran unit(s) currently open for output. Without the optional argument, all units are flushed, otherwise just the unit specified.
```

So, we need to close all open units, when `call FLUSH()` is used.

gxyd:
I'm sorry, I think I had some misunderstanding of what `call FLUSH()` without an argument meant, and so I've now rewritten the part to handle it in our runtime.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3275,6 +3275,63 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    al, x.base.base.loc, one, int_type, minus_one
+                )
+            );
+        } else {
+            diag.add(Diagnostic(
+                "Expected 0 or 1 argument, instead got " + std::to_string(x.n_args + x.n_keywords) +
+                " arguments instead.",
+                Level::Error, Stage::Semantic, {
+                    Label("",{x.base.base.loc})
+                })
```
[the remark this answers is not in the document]
gxyd: @Pranavchiku @parth121101 , I would like to add a test case for this to our test-suite, making use of `continue_compilation` compiler option, which file should I add a test case like `call FLUSH(1, 2)` (as that's expected to raise a semantic error)?

I see `tests/errors/continue_compilation_01.f90`, and similarly `..02.f90` and `...03.f90`, should I add to one of those or create a new file? I'm not sure of how it's being done currently.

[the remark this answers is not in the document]
gxyd: If there is no convention, I can use my own judgement for it.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2926,13 +2926,23 @@ LFORTRAN_API int64_t _lfortran_open(int32_t unit_num, char *f_name, char *status
...
+    if (unit_num == -1) {
+        for (int i = 0; i <= last_index_used; i++) {
+            if (unit_to_file[i].filep != NULL) {
+                fflush(unit_to_file[i].filep);
+            }
+        }
+    } else {
+        bool unit_file_bin;
+        FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin);
+        get_file_pointer_from_unit(unit_num, &unit_file_bin);
```
kmr-srbh, line 2939: Is it intentional to call the function here again, disregarding it's return value?

src/lfortran/semantics/ast_body_visitor.cpp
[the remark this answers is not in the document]
Pranavchiku: Add a line in any of those files. The convention was meant to avoid merge conflicts among people working on continue compilation at start.

src/libasr/runtime/lfortran_intrinsics.c
gxyd: Oh no this is a mistake, I'll fix this. There is no reason to do that.

src/lfortran/semantics/ast_body_visitor.cpp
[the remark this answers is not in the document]
gxyd: added a test case now for this.

src/libasr/runtime/lfortran_intrinsics.c
gxyd: I've fixed this

integration_tests/flush_02.f90
```diff
@@ -0,0 +1,8 @@
+program flush_02
```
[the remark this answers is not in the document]
gxyd: So, you are saying that if there was no `FLUSH` subroutine call above, the program would fail? It doesn't fail with GFortran for it. i.e.:
```f90
program test
    implicit none
    integer :: unit = 10
 
    open(unit=unit, file='test.txt')
 
    write(unit, '(A)') "Hello, world!"
    ! call flush(unit)
 
    rewind(unit)
 
    allocate(character(len=25) :: str)
    read(unit, '(A)') str
    print *, str
 
    if (str /= "Hello, world!") error stop
end program test

```

this program gives the same output, after commenting `call FLUSH(unit)` above. So, I'm not sure how it's testing whether `FLUSH` worked or not, because `FLUSH` statement was there or not, the impact is still the same.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3275,6 +3275,63 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
 
+    /*
+        Function to convert 'FLUSH' subroutine call to 'FLUSH' ASR node
+    */
+    void flush_subroutine_to_flush_statement(const AST::SubroutineCall_t &x) {
+        LCOMPILERS_ASSERT(to_lower(x.m_name) == "flush")
+
+        // for FLUSH subroutine call, the only argument 'unit' is optional
+        ASR::expr_t* unit_flush = nullptr;
+        if (x.n_args == 1 && x.n_keywords == 0) {
```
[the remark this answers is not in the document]
gxyd: that's not the focus of this PR, because we get the same error with `FLUSH` statement as well. That can be a separate issue for sure.

integration_tests/flush_02.f90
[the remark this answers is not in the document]
kmr-srbh: I looked into it, the `read` operation is causing the buffered data to be flushed automatically, so the data is read correctly. 

We can observe the effect with/without `flush` through this program:

```fortran
program writer
   implicit none
   integer :: sleep_duration = 10
   integer :: start, finish
   integer :: count_rate = 1000

   open(10, file='test.txt')

   write(10, '(A)') "Hello, world!"
   ! call flush(10)

   print *, "Writing done. Sleeping for ", sleep_duration, " seconds."

   ! call system_clock(count_rate=count_rate)
   call system_clock(count=start)

   do
      call system_clock(count=finish)
      if (finish - start >=  sleep_duration * count_rate) exit
   end do
end program writer
```

We can use `sleep()` in GFortran for simplicity, but we do not support this extension. Also, we have an error with querying the `count_rate`, so I got the value for my CPU using GFortran and hardcoded it. 

The buffered data is written to the file when the program exits, but using `flush()` flushes the data before. LFortran works correctly, but testing this will require a parallel reader program to read the flushed data while the writer program waits.

```fortran
program reader
   implicit none
   character(len=100) :: line
   integer :: ios

   open(10, file='test.txt', status="old", action="read")

   read(10, '(A)', iostat=ios) line
   if (ios /= 0) then
      print *, "Unable to read: File is empty or buffered."
   else
      print *, "Read line: ", trim(line)
   end if
end program reader
```

So, testing this at the CI is very tricky and probably not worth the effort. The implementation works correctly so we can be sure of it. We could maybe add this inside `examples/` where we have another difficult to test program related to reading user input.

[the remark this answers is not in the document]
kmr-srbh: Ah! Here is a simpler program:

```fortran
program writer
   implicit none
   character(len=25) :: c

   open(10, file='test.txt')

   write(10, '(A)') "Hello, world!"
   ! call flush()

   print *, "Written to buffer. Press ENTER to write to file."
   read(*, *) c

end program writer
```

Issue #5871, opened 2024-12-29
--openmp fails


certik:
The example from https://gist.github.com/hakostra/ac5ed9279136fe1e7f6f217f2561f08e, modified to compile: https://gist.github.com/certik/924cf554b0144f00d15937970310fab7 works like this:
```console
$ lfortran --cpp -D_USE_DO_CONCURRENT_ a.f90 -o a.out
$ lfortran --cpp  a.f90 -o a.out 
```
But fails to compile when you add `--openmp` to either line above:
```console
$ lfortran --cpp -D_USE_DO_CONCURRENT_ --openmp a.f90 -o x.omp
zsh: segmentation fault  lfortran --cpp -D_USE_DO_CONCURRENT_ --openmp a.f90 -o x.omp
```
LFortran was compiled in Release mode using `./build_release.sh`. In Debug mode we get:
```
$ lfortran --cpp -D_USE_DO_CONCURRENT_ --openmp a.f90 -o x.omp
Undefined symbols for architecture arm64:
  "_GOMP_barrier", referenced from:
      _lcompilers_function in a.tmp.o
      _lcompilers_function1 in a.tmp.o
  "_GOMP_parallel", referenced from:
      _main in a.tmp.o
      _main in a.tmp.o
  "_omp_get_max_threads", referenced from:
      _lcompilers_function in a.tmp.o
      _lcompilers_function1 in a.tmp.o
  "_omp_get_thread_num", referenced from:
      _lcompilers_function in a.tmp.o
      _lcompilers_function1 in a.tmp.o
  "_zatan2", referenced from:
      _lcompilers_function in a.tmp.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
The command 'clang -o x.omp a.tmp.o  -L"/Users/ondrej/repos/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/ondrej/repos/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
```

We should tell the user to use the following option:
```
[...]

PR #5946, opened 2025-01-09
Fix: Handle User-Defined Internal Function in DoConcurrent

adit4443ya:
Towards #5871 
Details can be check in this [comment](https://github.com/lfortran/lfortran/issues/5871#issuecomment-2567365691)

Im still getting this error
```clojure
 aditya-trivedi   lfortran    pure ≢  ?1 ~2    lfortran ./integration_tests/openmp_43.f90 --openmp  --linker=gcc  --openmp-lib-dir=${CONDA_PREFIX}/lib --dump-all-passes
visit_FunctionCall 25
Killed
```


Pranavchiku:
Yes, you need to debug ASR pass output for that.

src/libasr/pass/openmp.cpp
```diff
@@ -360,6 +360,27 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
             ASR::Function_t* new_func = ASR::down_cast<ASR::Function_t>(func_sym);
             new_func->m_body = nullptr; new_func->n_body = 0;
             ASR::down_cast<ASR::FunctionType_t>(new_func->m_function_signature)->m_deftype = ASR::deftypeType::Interface;
+            // Steps:
+            // Create a module add it to current scope->parent symtab
+            // Add func to that module symtab
+            // add External symbol to current_symtab symtab
+            char* module_name = strdup((std::string(func->m_name) + "_m").c_str());
```
gxyd, line 367: we usually use `s2c` (i.e. std::string to char*) in LFortran for this, see other places of its use

src/libasr/pass/openmp.cpp
```diff
@@ -360,6 +360,27 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
...
+            ASR::asr_t *mo = ASR::make_Module_t(al,func->base.base.loc, current_scope->parent,module_name, nullptr,0,false,false);
+            current_scope->parent->add_symbol(module_name, ASR::down_cast<ASR::symbol_t>(mo));
```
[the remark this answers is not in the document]
adit4443ya: Here is the ASR without that line , (which means without New module in symtab), I'm Just pasting a section of openmp pass dump

```clojure
(TranslationUnit
    (SymbolTable
        1
        {
           lcompilers_function:
                   (Function
                     ! Other symbols
                     
                            fun:
                                (Function
                                    (SymbolTable
                                        28
                                        {
                                            fun:
                                                (Variable
                                                    28
                                                    fun
                                                    []
                                                    ReturnVar
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                ),
                                            x:
                                                (Variable
                                                    28
                                                    x
                                                    []
                                                    In
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                )
                                        })
                                    fun
                                    (FunctionType
                                        [(Integer 4)]
                                        (Integer 4)
                                        Source
                                        Interface
                                        ()
                                        .false.
                                        .true.
                                        .false.
                                        .false.
                                        .false.
                                        []
                                        .false.
                                    )
                                    []
                                    [(Var 28 x)]
                                    []
                                    (Var 28 fun)
                                    Public
                                    .false.
                                    .false.
                                    ()
                                ),
                            fun_t:
                                (ExternalSymbol
                                    27
                                    fun
                                    2 fun
                                    fun_m
                                    []
                                    fun
                                    Public
                                ),
                              
```

src/libasr/pass/openmp.cpp
kmr-srbh, line 367: Let's handle the module name after we have the global scope access. We can then do `char* module_name = s2c(al, current_scope->parent->get_unique_name("lcompilers_user_defined_functions"))` to mangle the symbol name.

src/libasr/pass/openmp.cpp
kmr-srbh, line 369: The code is setting the module's symbol table to `current_scope->parent`, which, given how `lcompilers_function` is defined, is the global scope. Let's create a symbol table for the module and use it.

```c++
SymbolTable* module_scope = al.make_new<SymbolTable>(current_scope->parent);

// define the module here 
...

// we do not need to check for the presence of this symbol due to the use of `get_unique_name()`
current_scope->parent->add_symbol(...);
```

Also, let's check for the presence of this module before creating it.

src/libasr/pass/openmp.cpp
```diff
@@ -360,6 +360,27 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
...
+                /* a_symtab */ current_scope,
+                /* a_name */ func->m_name,
+                (ASR::symbol_t*)func,
+                module_name, nullptr, 0, func->m_name,
+                ASR::accessType::Public
+                );
+            std::string sym = to_lower(func->m_name);
+            sym += "_t";
+            current_scope->add_symbol(sym, ASR::down_cast<ASR::symbol_t>(fn));
+            std::cout<<"visit_FunctionCall 25"<<std::endl;
```
kmr-srbh, line 383: Let's do this:

- Duplicate the function inside the module.
- Check if the `asr_owner` of `x.m_name` is `ASR::Program_t`. If yes, `down_cast` the `asr_owner` symbol to `ASR::Program_t` and overwrite the function `ExternalSymbol` inside the symbol table of the program.
- Update `x.m_name` and `x.m_original_name` to point to the new symbols.

After this change, you will need to handle the logic inside the `if` block to use the correct symbol. We have an issue at #5951, which I think should get fixed by the changes we make as part of this implementation.

kmr-srbh (changes requested):
I think you should make these changes outside before the `if` block.

adit4443ya:
Now At this point we have , Please have a look at it

<details>
<summary>Openmp pass ASR</summary>

```clojure
 lcompilers_function:
  (Function
      ! Other symbols
              (Print
                   (StringFormat
                            ()
                            [(StringConstant
                                "Fun->"
                                (String 1 5 () PointerString)
                            )
                            (FunctionCall     // This is the x_copy which is we have inside the function visit_Functioncall
                                  2 fun         // This is giving error as there is no `2 fun` symbol in this symtab of lcompiler_functions, Question is What should it point to ?
                                  2 fun
                                  [((Var 27 i))]
                                  (Integer 4)
                                  ()
                                  ()
                              )]
                              FormatFortran
                              (String -1 0 () PointerString)
                              ()
                          )
                      )
lcompilers_user_defined_functions:
              (Module
                  (SymbolTable
                      29
                      {
                          fun:
                              (Function
                                  (SymbolTable
                                      30
                                      {
                                          fun:
                                              (Variable
                                                  30
                                                  fun
                                                  []
                                                  ReturnVar
                                                  ()
                                                  ()
                                                  Default
                                                  (Integer 4)
                                                  ()
                                                  Source
                                                  Public
                                                  Required
                                                  .false.
                                                  .false.
                                              ),
                                          x:
                                              (Variable
                                                 30
                                                  x
                                                  []
                                                  In
                                                  ()
                                                   ()
                                                  Default
                                                  (Integer 4)
                                                  ()
                                                  Source
                                                   Public
                                                  Required
                                                  .false.
                                                  .false.
                                              )
                                      })
                                  fun
                                  (FunctionType
                                      [(Integer 4)]
                                      (Integer 4)
                                        Source
                                        Implementation
                                        ()
                                        .false.
                                        .true.
                                        .false.
                                        .false.
                                        .false.
                                        []
                                        .false.
                                    )
                                    []
                                    [(Var 30 x)]
                                    [(Assignment
                                        (Var 3 fun)          // This change of symbols would be handled by #5968 
                                        (IntegerBinOp
                                            (Var 3 x)
                                            Mul
                                            (Var 3 x)
                                            (Integer 4)
                                            ()
                                        )
                                        ()
                                    )]
                                    (Var 30 fun)
                                    Public
                                    .false.
                                    .false.
                                    ()
                                )
                        })
                    lcompilers_user_defined_functions
                    []
                    .false.
                    .false.
                ),
openmp_43:
(Program
                            fun:
                                (ExternalSymbol
                                    2
                                    fun
                                    29 fun
                                    lcompilers_user_defined_functions
                                    []
                                    fun
                                    Public
                                ),
! body of program 
(Print
                        (StringFormat
                            ()
                            [(StringConstant
                                "Fun->"
                                (String 1 5 () PointerString)
                            )
                            (FunctionCall
                             2 fun    
                             ()
                             [((IntegerConstant 2 (Integer 4) Decimal))]
                             (Integer 4)
                             ()
                             ()
                         )
```

</details>

kmr-srbh:
With the following patch:

```diff
diff --git a/src/libasr/pass/openmp.cpp b/src/libasr/pass/openmp.cpp
index bf164206e..a7a1f2d34 100644
--- a/src/libasr/pass/openmp.cpp
+++ b/src/libasr/pass/openmp.cpp
@@ -348,11 +348,43 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
 
     void visit_FunctionCall(const ASR::FunctionCall_t &x) {
         ASR::FunctionCall_t* x_copy = const_cast<ASR::FunctionCall_t*>(&x);
+        // Steps:
+        // Create a module add it to current_scope->parent symtab
+        // Add func to that module symtab
+        // Overwrite External symbol to x's asr_owner's symtab
+        ASRUtils::SymbolDuplicator duplicator(al);
+        SymbolTable* module_scope = al.make_new<SymbolTable>(current_scope->parent);
+        ASR::asr_t* asr_owner = ASRUtils::symbol_parent_symtab(x.m_name)->asr_owner;
+        char* module_name = s2c(al, current_scope->parent->get_unique_name("lcompilers_user_defined_functions"));
+        ASR::asr_t* mo = ASR::make_Module_t(
+            al, x.base.base.loc, module_scope, s2c(al, module_name), nullptr, 0, false, false);
+        current_scope->parent->add_symbol(module_name, ASR::down_cast<ASR::symbol_t>(mo));
+        ASR::Module_t* module = ASR::down_cast<ASR::Module_t>(ASR::down_cast<ASR::symbol_t>(mo));
+        ASR::symbol_t* fun_sym_for_module = duplicator.duplicate_Function(
+            ASR::down_cast<ASR::Function_t>(x_copy->m_name), module_scope);
+        module->m_symtab->add_symbol(ASRUtils::symbol_name(x_copy->m_name), fun_sym_for_module);
+        ASR::asr_t* fn = ASR::make_ExternalSymbol_t(
+                        al,
+                        x.base.base.loc,
+                        ASRUtils::symbol_parent_symtab(x.m_name),
+                        ASRUtils::symbol_name(x_copy->m_name),
+                        fun_sym_for_module,
+                        s2c(al, module_name),
+                        nullptr,
+                        0,
+                        x_copy->m_original_name 
+                                        ? ASRUtils::symbol_name(x_copy->m_original_name)
+                                        : ASRUtils::symbol_name(x_copy->m_name),
+                        ASR::accessType::Public);
+        if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))) {
+            ASR::Program_t* program = ASR::down_cast<ASR::Program_t>(ASR::down_cast<ASR::symbol_t>(asr_owner));
+            program->m_symtab->add_or_overwrite_symbol(ASRUtils::symbol_name(x_copy->m_name),
+                                                       ASR::down_cast<ASR::symbol_t>(fn));
+        }
         ASR::symbol_t* func_sym = current_scope->get_symbol(ASRUtils::symbol_name(x.m_name));
         if (func_sym == nullptr) {
-            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(x.m_name);
+            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(ASRUtils::symbol_get_past_external(x.m_name));
             // this means we have a user defined function and need to create an interface for it
-            ASRUtils::SymbolDuplicator duplicator(al);
             // This will create trouble when we have something like x(n) inside function
             // duplicator does not duplicate type of variables inside function
             func_sym = duplicator.duplicate_Function(func, current_scope);
@@ -360,31 +392,6 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
             ASR::Function_t* new_func = ASR::down_cast<ASR::Function_t>(func_sym);
             new_func->m_body = nullptr; new_func->n_body = 0;
             ASR::down_cast<ASR::FunctionType_t>(new_func->m_function_signature)->m_deftype = ASR::deftypeType::Interface;
-            // Steps:
-            // Create a module add it to current_scope->parent symtab
-            // Add func to that module symtab
-            // Overwrite External symbol to x's asr_owner's symtab
-            SymbolTable* module_scope = al.make_new<SymbolTable>(current_scope->parent);
-            ASR::asr_t* asr_owner = ASRUtils::symbol_parent_symtab(x.m_name)->asr_owner;
-            char* module_name = s2c(al, current_scope->parent->get_unique_name("lcompilers_user_defined_functions"));
-            ASR::asr_t *mo = ASR::make_Module_t(al,func->base.base.loc, module_scope,s2c(al,module_name), nullptr,0,false,false);
-            current_scope->parent->add_symbol(module_name, ASR::down_cast<ASR::symbol_t>(mo));
-            ASR::Module_t* module = ASR::down_cast<ASR::Module_t>(ASR::down_cast<ASR::symbol_t>(mo));
-            ASR::symbol_t* fun_sym_for_module = duplicator.duplicate_Function(func, module_scope);
-            module->m_symtab->add_symbol(func->m_name, fun_sym_for_module);
-            ASR::asr_t *fn = ASR::make_ExternalSymbol_t(
-                al, func->base.base.loc,
-                ASRUtils::symbol_parent_symtab(x.m_name),
-                func->m_name,
-                fun_sym_for_module,
-                s2c(al,module_name), nullptr, 0, func->m_name,
-                ASR::accessType::Public
-                );
-            if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))){
-                ASR::Program_t* program = ASR::down_cast<ASR::Program_t>(ASR::down_cast<ASR::symbol_t>(asr_owner));
-                program->m_symtab->add_or_overwrite_symbol(func->m_name, ASR::down_cast<ASR::symbol_t>(fn));
-            }
-            func_sym = ASR::down_cast<ASR::symbol_t>(fn);
         }
         LCOMPILERS_ASSERT(func_sym != nullptr);
         x_copy->m_name = func_sym;
@@ -420,23 +427,31 @@ public:
     void replace_Var(ASR::Var_t *x) {
         x->m_v = fn_scope.get_symbol(ASRUtils::symbol_name(x->m_v));
     }
+
+    void replace_FunctionCall(ASR::FunctionCall_t* x) {
+        x->m_name = fn_scope.get_symbol(ASRUtils::symbol_name(x->m_name));
+        if (x->m_original_name) x->m_original_name = fn_scope.get_symbol(ASRUtils::symbol_name(x->m_original_name));
+    }
 };
 
 // Expression visitor to call the replacer
-class DoConcurrentBodyVisitor:
-    public ASR::CallReplacerOnExpressionsVisitor<DoConcurrentBodyVisitor> {
+class ReplaceSymbolsVisitor:
+    public ASR::CallReplacerOnExpressionsVisitor<ReplaceSymbolsVisitor> {
 
 private:
     ReplaceSymbols replacer;
 
 public:
-    DoConcurrentBodyVisitor(SymbolTable &fn_scope): replacer(fn_scope) { }
+    ReplaceSymbolsVisitor(SymbolTable &fn_scope): replacer(fn_scope) { }
 
     void call_replacer() {
         replacer.current_expr = current_expr;
         replacer.replace_expr(*current_expr);
     }
 
+    void visit_FunctionCall(const ASR::FunctionCall_t &x) {
+        replacer.replace_expr(&const_cast<ASR::FunctionCall_t*>(&x)->base);
+    }
 };
 
 class DoConcurrentVisitor :
@@ -1367,7 +1382,7 @@ class DoConcurrentVisitor :
                     fn_args.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, loc, sym)));
                 }
 
-                DoConcurrentBodyVisitor v(*fn_scope);
+                ReplaceSymbolsVisitor v(*fn_scope);
                 v.visit_DoConcurrentLoop(x);
 
                 Vec<ASR::stmt_t *> fn_body; fn_body.reserve(al, 1);
@@ -1627,6 +1642,17 @@ class DoConcurrentVisitor :
             current_scope = current_scope_copy;
         }
 
+        void visit_FunctionCall(const ASR::FunctionCall_t &x) {
+            SymbolTable* current_scope_copy = current_scope;
+            current_scope = ASRUtils::symbol_parent_symtab(x.m_name);
+            if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(current_scope->asr_owner))) {
+                ReplaceSymbolsVisitor sym_replacer(*current_scope);
+                sym_replacer.visit_FunctionCall(x);
+            }
+
+            current_scope = current_scope_copy;
+        }
+
         void visit_DoLoop(const ASR::DoLoop_t &x) {
             ASR::DoLoop_t& xx = const_cast<ASR::DoLoop_t&>(x);
 

```


we are able to generate the correct ASR:

<details>

```clojure
;; ASR after applying the pass: openmp
(TranslationUnit
    (SymbolTable
        1
        {
            iso_c_binding:
                (IntrinsicModule lfortran_intrinsic_iso_c_binding),
            lcompilers_function:
                (Function
                    (SymbolTable
                        27
                        {
                            ;; other symbols
                            fun:
                                (Function
                                    (SymbolTable
                                        30
                                        {
                                            fun:
                                                (Variable
                                                    30
                                                    fun
                                                    []
                                                    ReturnVar
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                ),
                                            x:
                                                (Variable
                                                    30
                                                    x
                                                    []
                                                    In
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                )
                                        })
                                    fun
                                    (FunctionType
                                        [(Integer 4)]
                                        (Integer 4)
                                        Source
                                        Interface
                                        ()
                                        .false.
                                        .true.
                                        .false.
                                        .false.
                                        .false.
                                        []
                                        .false.
                                    )
                                    []
                                    [(Var 30 x)]
                                    []
                                    (Var 30 fun)
                                    Public
                                    .false.
                                    .false.
                                    ()
                                ),
                        ;; other module statements
                        (Print
                            (StringFormat
                                ()
                                [(StringConstant
                                    "Fun->"
                                    (String 1 5 () PointerString)
                                )
                                (FunctionCall
                                    27 fun
                                    27 fun
                                    [((Var 27 i))]
                                    (Integer 4)
                                    ()
                                    ()
                                )]
                                FormatFortran
                                (String -1 0 () PointerString)
                                ()
                            )
                        )]
                        []
                    )
                    (SubroutineCall
                        27 gomp_barrier
                        ()
                        []
                        ()
                    )]
                    ()
                    Public
                    .false.
                    .false.
                    ()
                ),
            lcompilers_user_defined_functions:
                (Module
                    (SymbolTable
                        28
                        {
                            fun:
                                (Function
                                    (SymbolTable
                                        29
                                        {
                                            fun:
                                                (Variable
                                                    29
                                                    fun
                                                    []
                                                    ReturnVar
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                ),
                                            x:
                                                (Variable
                                                    29
                                                    x
                                                    []
                                                    In
                                                    ()
                                                    ()
                                                    Default
                                                    (Integer 4)
                                                    ()
                                                    Source
                                                    Public
                                                    Required
                                                    .false.
                                                    .false.
                                                )
                                        })
                                    fun
                                    (FunctionType
                                        [(Integer 4)]
                                        (Integer 4)
                                        Source
                                        Implementation
                                        ()
                                        .false.
                                        .true.
                                        .false.
                                        .false.
                                        .false.
                                        []
                                        .false.
                                    )
                                    []
                                    [(Var 29 x)]
                                    [(Assignment
                                        (Var 29 fun)
                                        (IntegerBinOp
                                            (Var 29 x)
                                            Mul
                                            (Var 29 x)
                                            (Integer 4)
                                            ()
                                        )
                                        ()
                                    )]
                                    (Var 29 fun)
                                    Public
                                    .false.
                                    .false.
                                    ()
                                )
                        })
                    lcompilers_user_defined_functions
                    []
                    .false.
                    .false.
                ),
            ; module omp_lib
            openmp_43:
                (Program
                    (SymbolTable
                        2
                        {
                            ;; other symbols
                            fun:
                                (ExternalSymbol
                                    2
                                    fun
                                    28 fun
                                    lcompilers_user_defined_functions
                                    []
                                    fun
                                    Public
                                ),
                    ;; other program statements
                    (Print
                        (StringFormat
                            ()
                            [(StringConstant
                                "Fun->"
                                (String 1 5 () PointerString)
                            )
                            (FunctionCall
                                2 fun
                                ()
                                [((IntegerConstant 2 (Integer 4) Decimal))]
                                (Integer 4)
                                ()
                                ()
                            )]
                            FormatFortran
                            (String -1 0 () PointerString)
                            ()
                        )
                    )]
                ),
            ;; thread_data_module module
        })
    []
)

```

</details>

But the interface function `fun` which we create in the pass, still remains undefined. Why do we need to do so? How is the required function linked to it at link time?

kmr-srbh:
I had pulled changes from #5968 to duplicate the function correctly. I will try to resolve the failure in that PR so we could merge it soon.

kmr-srbh:
I realized `interface` here did not refer to link time function resolution, like `bind(c)`, but to resolving the correct function in the ASR. With [fad5259](https://github.com/lfortran/lfortran/pull/5946/commits/fad52594bc64d63210b25691b9f148f587472f46),  we can now successfully compile and execute the program in this PR and https://github.com/lfortran/lfortran/issues/5871#issuecomment-2566392831.

The output is slightly off due to the below bug unrelated to this PR:

```fortran
program main
   implicit none

   integer :: i

   do concurrent (i = 1:4)
      print *, i
   end do

end program main

```
```console
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90 --openmp --openmp-lib-dir=${CONDA_PREFIX}/lib
3
4
1
2
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90 --openmp --openmp-lib-dir=${CONDA_PREFIX}/lib
4
3
1
2
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90 --openmp --openmp-lib-dir=${CONDA_PREFIX}/lib
4
1
2
3
```

Pranavchiku:
> due to the below bug unrelated to this PR:

This is not a bug. We never know which thread picks what chunk of the loop size, it is purely based on heuristics that GNU guys have determines. The final output after synchronising all the threads should match.

integration_tests/openmp_43.f90
```diff
@@ -0,0 +1,17 @@
+program openmp_43
+   implicit none
+
+   integer :: i
+
+   do concurrent (i = 1:4)
+      print *,"Fun->", FUN(i)
```
Pranavchiku, line 7: Do not check it like this, do a sum and then at the end test the value of it

src/libasr/pass/openmp.cpp
```diff
@@ -395,23 +453,58 @@ class ReplaceSymbols: public ASR::BaseExprReplacer<ReplaceSymbols> {
...
+    void replace_FunctionCall(ASR::FunctionCall_t* x) {
+        x->m_name = fn_scope.get_symbol(ASRUtils::symbol_name(x->m_name));
+        if (x->m_original_name) x->m_original_name = fn_scope.get_symbol(ASRUtils::symbol_name(x->m_original_name));
+    }
 };
 
 // Expression visitor to call the replacer
-class DoConcurrentBodyVisitor:
-    public ASR::CallReplacerOnExpressionsVisitor<DoConcurrentBodyVisitor> {
+class ReplaceSymbolsVisitor:
```
Pranavchiku, line 464: Why do we need this renaming? What we are trying to do is visit the body of do concurrent loop and replacing it with appropriate expressions, no?

src/libasr/pass/openmp.cpp
```diff
@@ -395,23 +453,58 @@ class ReplaceSymbols: public ASR::BaseExprReplacer<ReplaceSymbols> {
...
         replacer.current_expr = current_expr;
         replacer.replace_expr(*current_expr);
     }
 
+    void visit_FunctionCall(const ASR::FunctionCall_t &x) {
+        replacer.replace_expr(&const_cast<ASR::FunctionCall_t*>(&x)->base);
+    }
+};
+
+class ReplaceStatements: public ASR::BaseStmtReplacer<ReplaceStatements> {
```
Pranavchiku, line 483: Let's be specific to name the visitors. This will get unnecessary tricky to understand what this visitor does later on.

src/libasr/pass/openmp.cpp
```diff
@@ -1342,7 +1435,7 @@ class DoConcurrentVisitor :
                     fn_args.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, loc, sym)));
                 }
 
-                DoConcurrentBodyVisitor v(*fn_scope);
+                ReplaceSymbolsVisitor v(*fn_scope);
```
Pranavchiku, line 1438: I'll prefer to keep it as `DoConcurrentBodyVisitor`

src/libasr/pass/openmp.cpp
kmr-srbh: Previously this visitor was used for only `DoConcurrentLoop`, but now we use it in general for replacing other expressions like `FunctionCall`, which can be present outside the loop body also, like `Program` for the test in this PR. So I renamed it for clarity.

src/libasr/pass/openmp.cpp
```diff
@@ -346,26 +346,84 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
...
+                            ASRUtils::symbol_parent_symtab(x.m_name),
+                            ASRUtils::symbol_name(x_copy->m_name),
+                            fun_sym_for_module,
+                            s2c(al, module_name),
+                            nullptr,
+                            0,
+                            x_copy->m_original_name
+                                ? ASRUtils::symbol_name(x_copy->m_original_name)
+                                : ASRUtils::symbol_name(x_copy->m_name),
+                            ASR::accessType::Public);
```
Pranavchiku, line 381: Why do you need to create a module, add it to symtab and create an external symbol? Why can't you just use the logic that was previously written, which creates interface of the function by finding the function definition and adds it to symtab? This interface will serve as a linking step for backend.

src/libasr/pass/openmp.cpp
```diff
@@ -346,26 +346,84 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
...
-            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(x.m_name);
-            ASRUtils::SymbolDuplicator duplicator(al);
             // This will create trouble when we have something like x(n) inside function
             // duplicator does not duplicate type of variables inside function
-            func_sym = duplicator.duplicate_Function(func, current_scope);
-            current_scope->add_symbol(ASRUtils::symbol_name(func_sym), func_sym);
-            ASR::Function_t* new_func = ASR::down_cast<ASR::Function_t>(func_sym);
-            new_func->m_body = nullptr; new_func->n_body = 0;
-            ASR::down_cast<ASR::FunctionType_t>(new_func->m_function_signature)->m_deftype = ASR::deftypeType::Interface;
+            if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))) {
```
Pranavchiku, line 392: I think this logic shall suffice, no?

src/libasr/pass/openmp.cpp
kmr-srbh: With the previous approach when we created an `interface` function, the test in this PR had passed. But for the test case in https://github.com/lfortran/lfortran/issues/5871#issuecomment-2567365691, the function `fun` got removed by `unused_functions` pass, so we had no function definition to link the interface with. Adding to a module and using `ExternalSymbol` ensures we can have access to the function.

kmr-srbh:
The SciPy failure is due to the changes in #5968 which I mistakenly pushed along with the latest commit. I am removing it.

src/libasr/pass/openmp.cpp
Pranavchiku: Then it is a bug in `unused_function` pass. I do not think this is a right approach, because the code that this pass generates is a valid Fortran code that any user can write and we cannot tell that you need to wrap it inside module.

kmr-srbh: It could be so. Can `FunctionCall` inside a global function resolve the function to be called from `Program`?

Pranavchiku: > Can FunctionCall inside a global function resolve the function to be called from Program?

Yes, even a functioncall inside a function which is nested to global function resolve the function to be called from Program, refer test added in https://github.com/lfortran/lfortran/pull/5980/files

Pranavchiku: Only thing you need is to make backend understand that whatever you are passing has a implementation somewhere, not even in given file maybe linked externally and that shall work.

Pranavchiku: The fix you are looking here is to add the udf into dependencies which I suspect might be empty in the `_lcompilers_function` we create and thus the `unused_function` pass won't be able to identify this being used.

kmr-srbh: Ah, by global functions I had meant functions declared solely in the global scope.

I think the `openmp` pass transforms code roughly like:

```fortran
integer function lcompilers_function()
   implicit none
   integer :: i

   interface
      integer function fun(x)
         integer, intent(in) :: x
      end function fun
   end interface

   ! do loop for reference only
   do i = 1, 4
      print *, "lcompilers_function_fun: ", fun(i)
   end do

   lcompilers_function = 1
end function lcompilers_function


program main
   implicit none

   interface
      integer function lcompilers_function()
      end function lcompilers_function
   end interface

   ! function invocation for reference only
   print *, lcompilers_function()
   print *, "main_fun: ", fun(2)

contains

   integer function fun(x)
      integer, intent(in) :: x
      fun = x * x
   end function fun

end program main

```

Here, the global function `lcompilers_function` invokes `fun`, but cannot resolve it because it is present inside `program`. GFortran has the same behavior.

```console
/usr/bin/ld: /tmp/cc7Ge3xD.o: in function `lcompilers_function_':
test.f90:(.text+0x69): undefined reference to `fun_'
collect2: error: ld returned 1 exit status
/usr/bin/ld: test.tmp.o: in function `lcompilers_function':
LFortran:(.text+0x43): undefined reference to `fun'
The command 'clang -o test.out test.tmp.o  -L"/home/saurabh-kumar/Projects/System/lfortran/src/bin/../runtime" -Wl,-rpath,"/home/saurabh-kumar/Projects/System/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
```

I think the behavior is correct given the fact that `program` can access `lcompilers_function` through the `interface` due to the function's global nature, but `lcompilers_function` does not know where `fun` comes from because `fun` is local to `program`.

LFortran has a bug in this regard which can be observed by uncommenting the line `! print *, "main_fun: ", fun(2)`.

```console
/usr/bin/ld: /tmp/ccWjJWbd.o: in function `lcompilers_function_':
test.f90:(.text+0x87): undefined reference to `fun_'
collect2: error: ld returned 1 exit status
lcompilers_function_fun:     1
lcompilers_function_fun:     4
lcompilers_function_fun:     9
lcompilers_function_fun:     16
1
main_fun:     4
```

GFortran produces an error because `fun` inside `lcompilers_function` still could be resolved, but LFortran produced an output. 

This is because we always define functions in the LLVM back-end globally. In the first case, we got a linker error because the `unused_functions` pass removed `fun` from `program` and it was absent from the LLVM IR. In the second case, `fun` was not removed, got defined in the LLVM IR and the linker resolved it.

I think we need to declare program functions with `internal` linkage and/or mangle them. 

So considering the above behavior, I think the transformation by `openmp` pass probably needs to be changed.

src/libasr/pass/openmp.cpp
kmr-srbh: Please see https://github.com/lfortran/lfortran/pull/5946#discussion_r1912427677.

integration_tests/CMakeLists.txt
```diff
@@ -1952,6 +1952,7 @@ RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_40 LABELS llvm_omp llvm)
 RUN(NAME openmp_41 LABELS llvm_omp llvm)
 RUN(NAME openmp_42 LABELS llvm_omp llvm)
+RUN(NAME openmp_43 LABELS llvm_omp)
```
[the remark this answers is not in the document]
kmr-srbh: It runs with `llvm`, but produces incorrect output due to #6006.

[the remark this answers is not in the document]
adit4443ya: It does run But gives value of 
```
 aditya-trivedi   lfortran    pure ↑10 ↓6  ~1    lfortran integration_tests/openmp_43.f90 
FUN sum->    30
SUB sum->    5  // This is problem
```

Which is I'm not sure about that it's right or not
We are correct in the case of `--openmp` (e.g. DoConcurrentLoop)
The iterating variable goes into subroutine call and then it gets updated which is summed over and displayed as `30`. 

But in case of not using `--openmp` we replace it via normal `DoLoop` Hence Now It's property should be followed
And as per this , Do loop iterator should not be modified. I opened Up the Issue at #6006 Please Check it out.

Check These two lines on this link https://support.nag.com/nagware/np/r72_doc/nag_f2018.html

- These locality specifiers determine how a variable can be used within and without the loop, and cannot be applied to the loop index variables, which are always effectively LOCAL.

- ... if LOCAL, it begins each iteration undefined, and if LOCAL_INIT it begins each iteration with the value of the outside variable. This ensures that LOCAL and LOCAL_INIT variables cannot cause any dependency between iterations.

This means that after each iteration the iterator variable gets the loop_count value which gets updated by FUN(). Which is right about DoConcurrenLoop
But for the DoLoop the loop_count and the iterator variable are dependent so we can't alter the loop variable, which again corresponds to the #6006 

This have been also discussed over Zulip here https://lfortran.zulipchat.com/#narrow/channel/452774-Internal/topic/Modification.20of.20iterating.20var.20in.20loops/near/491426869

adit4443ya:
I think now the test is good.
@certik 
```clojure
 aditya-trivedi   lfortran    pure ≡  ~1    lfortran integration_tests/openmp_43.f90
FUN sum->    30
SUB sum->    30
 aditya-trivedi   lfortran    pure ≡  ~1    lfortran ./integration_tests/openmp_43.f90 --openmp  --linker=gcc  --openmp-lib-dir=${CONDA_PREFIX}/lib 
FUN sum->    30
SUB sum->    30
 aditya-trivedi   lfortran    pure ≡  ~1    gfortran integration_tests/openmp_43.f90 -o ../a && ../a
 FUN sum->          30
 SUB sum->          30
 ```

File: integration_tests/CMakeLists.txt [hunk 11]
```diff
@@ -1952,6 +1952,7 @@ RUN(NAME openmp_39 LABELS gfortran GFORTRAN_ARGS -fopenmp)
 RUN(NAME openmp_40 LABELS llvm_omp llvm)
 RUN(NAME openmp_41 LABELS llvm_omp llvm)
 RUN(NAME openmp_42 LABELS llvm_omp llvm)
+RUN(NAME openmp_43 LABELS llvm_omp llvm)
```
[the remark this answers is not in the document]
adit4443ya: We only use `reduce(+:x)` in LFortran with doconcurrent for summing operation, GFortran does not require it and hence gives error for it.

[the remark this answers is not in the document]
adit4443ya: Yes I checked it .. in above comment I showed you output for GFortran after removing that reduce.

src/libasr/pass/openmp.cpp
kmr-srbh: Can we call it `FixStmtScopeVisitor`?

kmr-srbh (approved):
The changes look good. The rationale for the design change stems from #6009.

In a later PR, we can handle the case of global functions correctly. The code:

```fortran
PURE integer function FUN(x)
   integer, intent(in) :: x
   FUN = x * x
end function FUN


program openmp_43
   implicit none

   interface
      PURE integer function FUN(x)
         integer, intent(in) :: x
      end function FUN
   end interface

   integer :: i

   do concurrent (i = 1:4)
      print *,"Fun->", FUN(i)
   end do

end program openmp_43
```

produces ASR which is bloated. Currently we create a module, duplicate the interface `FUN` inside it, and import it as `ExternalSymbol` inside `lcompilers_function` and `program`. This produces correct output, but is not good design. 

I think due to the global nature of `FUN`, we can duplicate the interface inside `lcompilers_function` and it should work.

```diff
diff --git a/src/libasr/pass/openmp.cpp b/src/libasr/pass/openmp.cpp
index 93fc8f3e0..12f9b4cc3 100644
--- a/src/libasr/pass/openmp.cpp
+++ b/src/libasr/pass/openmp.cpp
@@ -351,15 +351,18 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
         T* x_copy = const_cast<T*>(&x);
         ASR::Function_t* fn = ASR::down_cast<ASR::Function_t>(
                             ASRUtils::symbol_get_past_external(x_copy->m_name));
+        ASR::FunctionType_t* fn_type = ASR::down_cast<ASR::FunctionType_t>(
+                                        fn->m_function_signature);
         ASR::asr_t* asr_owner = ASRUtils::symbol_parent_symtab(x.m_name)->asr_owner;
         ASR::symbol_t* fun_sym_for_module = nullptr;
         char* module_name = nullptr;
+        ASRUtils::SymbolDuplicator duplicator(al);
         // Steps:
         // Create a module add it to current_scope->parent symtab
         // Add func to that module symtab
         // Overwrite External symbol to x's asr_owner's symtab
-        if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))) {
-            ASRUtils::SymbolDuplicator duplicator(al);
+        if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))
+            && fn_type->m_deftype != ASR::deftypeType::Interface) {
             SymbolTable* module_scope = al.make_new<SymbolTable>(current_scope->parent);
 
             module_name = s2c(al, current_scope->parent->get_unique_name("lcompilers_user_defined_functions"));
@@ -396,7 +399,8 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
 
         ASR::symbol_t* func_sym = current_scope->get_symbol(ASRUtils::symbol_name(x.m_name));
         if (func_sym == nullptr) {
-            if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))) {
+            if (ASR::is_a<ASR::Program_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))
+                && fn_type->m_deftype != ASR::deftypeType::Interface) {
                 ASR::asr_t* ext_fn = ASR::make_ExternalSymbol_t(
                                         al,
                                         x.base.base.loc,
@@ -415,6 +419,10 @@ class DoConcurrentStatementVisitor : public ASR::CallReplacerOnExpressionsVisito
                 func_sym = current_scope->get_symbol(fn->m_name);
             } else if (ASR::is_a<ASR::Module_t>(*ASR::down_cast<ASR::symbol_t>(asr_owner))) {
                 func_sym = current_scope->resolve_symbol(fn->m_name);
+            } else {
+                // we are dealing with an interface
+                func_sym = duplicator.duplicate_Function(fn, current_scope);
+                current_scope->add_symbol(fn->m_name, func_sym);
             }
         }
         LCOMPILERS_ASSERT(func_sym != nullptr);
```

Thanks for working on this @adit4443ya!

adit4443ya:
>Thanks for working on this @adit4443ya!

Thanks to You too @kmr-srbh !!!!!

src/libasr/pass/openmp.cpp
Pranavchiku: > I think the openmp pass transforms code roughly like:

The `DoConcurrentSymbol` collector will collect `fun` from the body and will put it inside `lcompilers_thread_data_module` from which we can easily access it. So the conversion roughly looks like:

```fortran
module lcompilers_thread_data_module
integer :: i
integer :: x
contains

integer function fun(x)
integer, intent(in) :: x
fun = x * x
end function fun

end module

integer function lcompilers_function()
   use lcompilers_thread_data_module
   implicit none

   ! do loop for reference only
   do i = 1, 4
      print *, "lcompilers_function_fun: ", fun(i)
   end do

   lcompilers_function = 1
end function lcompilers_function


program main
   implicit none

   interface
      integer function lcompilers_function()
      end function lcompilers_function
   end interface

   ! function invocation for reference only
   print *, lcompilers_function()
   print *, "main_fun: ", fun(2)

contains

   integer function fun(x)
      integer, intent(in) :: x
      fun = x * x
   end function fun

end program main
```

Where you populate `lcompilers_thread_data_module` as soon as you visit do concurrent loop.

src/libasr/pass/openmp.cpp
Pranavchiku: I do not think that is correct, please refer: https://github.com/lfortran/lfortran/pull/5946#discussion_r1918515609

Pranavchiku:
Thanks for getting this PR to finish line! However, I am still not satisfied with the approach used in this PR therefore let's improve on this. The changes pushed in this PR is a workaround and not an actual fix.

Pranavchiku:
Opened https://github.com/lfortran/lfortran/issues/6032 for this

src/libasr/pass/openmp.cpp
kmr-srbh: Ah, I checked this on an older branch, the `thread_data_module` does not contain the symbol `fun`. Could that have been the main issue? #6009 is a valid bug though.

kmr-srbh:
As you say. :slightly_smiling_face:

PR #5956, opened 2025-01-10
tests: uncomments tests and remove TODO

gxyd:
## Description

we probably forgot to uncomment these test cases, this probably should've been done once we used label *ONLY_EXPERIMENTAL_SIMPLIFIER*


Pranavchiku (approved):
Thanks for doing this!

PR #5968, opened 2025-01-10
refactor: perform scoped duplication of function body and arguments while duplicating `ASR::Function_t`

kmr-srbh:
The duplication of `ASR::stmt_t` nodes like `ASR::Assignment_t` was done incorrectly before causing expressions like `ASR::Var_t` to point to the original scope. We fix it using scoped duplication. Rest of the change is cleanup due to the use of new method.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -4477,14 +4477,16 @@ class SymbolDuplicator {
         SymbolTable* function_symtab = al.make_new<SymbolTable>(destination_symtab);
         duplicate_SymbolTable(function->m_symtab, function_symtab);
         Vec<ASR::stmt_t*> new_body;
-        new_body.reserve(al, function->n_body);
-        ASRUtils::ExprStmtDuplicator node_duplicator(al);
-        node_duplicator.allow_procedure_calls = true;
-        node_duplicator.allow_reshape = false;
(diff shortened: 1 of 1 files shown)


Pranavchiku:
yup, not only these some `ArraySize` and type nodes used to have `Var xx` pointing to old symtab

Pranavchiku (commented):
changes look good, see if you can fix CI failure then will have a final review.

kmr-srbh:
The issue occurs with external/statement functions used as argument for a subroutine.

```fortran
SUBROUTINE DJCKZ

   EXTERNAL DPVB

END

SUBROUTINE DJCKM(FCN)

   EXTERNAL FCN

   EXTERNAL DPVB

   CALL DPVB(FCN)

END

```

I have a fix for this.

kmr-srbh:
Prior to this PR, function duplication was incorrect in sense that `ASR::Var_t` was duplicated as it is because we did not enforce the scope. So, when we duplicated a function, all `Var_t` nodes pointed to the original scope. This issue was not caught before because `duplicate_function()` is used only once in the code-base in the `openmp` pass.

kmr-srbh:
Due to no use of this function anywhere other than once in the `openmp` pass, I think it is not possible to test this currently.

PR #5984, opened 2025-01-12
tests: fix scale factor in format specifier

kmr-srbh:


integration_tests/format_04.f90
```diff
@@ -27,8 +27,8 @@ program format_04
 print "(12(i3))", 1,2,3,4,5,6,7,8,9,10,11,12
 print "(4(i3),' hello')", 1,2,3,4,5,6,7,8,9,10,11,12,13,14
 print '(i0)', f, -f
-print '(d0.0,1x,d0.1,1x,d0.2)',a,b,c
```
gxyd, line 30: The only concerning format specifier here is `D` format specifier with a specification like:
```
d0.0
```

other specifications like `d0.1`, `d0.2` etc. are valid specifications.

I don't see any need to *untest* whatever is being tested already with it.

Refer: the Fortran standard 24-007.pdf pg. 277:
Screenshot 2025-01-13 at 11 58 09 AM

gxyd: And if we do remove the `d0.0` format specifier, because it should ideally raise an error, it would be nice to open an issue for it.

kmr-srbh: Sure. I added the changes from here in #5983 so I will update it there. Thanks for pointing this out! :+1:

gxyd:
I think we can close this PR, because when we merged https://github.com/lfortran/lfortran/pull/5983, the change introduced in this PR has already been merged.

PR #5987, opened 2025-01-12
Fix : insert `implicit_deallocate` before `return` only in block constructs

### LINKED ISSUE #5986: bug : `insert_deallocte` inserts deallocation before hitting `exit` statement
MRE
```.f90
program prog
  integer :: i
  allocate(arr(10)) 
  arr = [1,2,3,4,5,6,7,8,9,10]
  do i =1,2
    print *, "dummy do loop" 
    exit
  end do
  print *, arr
  if (any(arr /= [1,2,3,4,5,6,7,8,9,10])) error stop
end program prog
```
```Console
dummy do loop
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt3.f90  && ./a.out 
 dummy do loop
           1           2           3           4           5           6           7           8           9          10

```

assem2002:
fixes #5986


assem2002:
only `block` uses `exit` to terminate the construct ( a construct that has declared variable in it - and needs to be deallocated if necessary)
e.g.

https://github.com/lfortran/lfortran/blob/e0a1a30f96e76b86c3378a7b56c312df745ec13e/tests/finalize_01.f90#L36-L47

PR #5995, opened 2025-01-13
build: use "find_package" instead of "find_library" for libunwind

gxyd:
## Description

Addresses the comment left here: https://github.com/lfortran/lfortran/pull/5031

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,7 +369,7 @@ if (WITH_LIBUNWIND_FORCE)
     set(WITH_LIBUNWIND yes)
 endif()
 if (WITH_LIBUNWIND)
-    find_library(LIBUNWIND_LIBRARY unwind)
+    find_package(LIBUNWIND_LIBRARY unwind)
 endif()
 
 set(WITH_KOKKOS no CACHE BOOL "Detect Kokkos location for cpp backend")


Pranavchiku (approved):

CMakeLists.txt
```diff
@@ -369,7 +369,7 @@ if (WITH_LIBUNWIND_FORCE)
     set(WITH_LIBUNWIND yes)
 endif()
 if (WITH_LIBUNWIND)
-    find_library(LIBUNWIND_LIBRARY unwind)
+    find_package(LIBUNWIND_LIBRARY unwind)
```
kmr-srbh, line 372:
```suggestion
    find_package(libunwind REQUIRED)
```

kmr-srbh (changes requested):
We also need to update `src/lfortran/CMakeLists.txt` and `src/libasr/CMakeLists.txt` to link the library correctly.

CMakeLists.txt
kmr-srbh: Okay, so I think this won't work with only the above changes. `libunwind` is actually a library, not a package. `conda` provides it as `libunwind.so`. So if we want to really use `find_package()` to resolve it, we will also need to provide CMake a `Findlibunwind.cmake` file. Something like - https://github.com/v6d-io/v6d/blob/main/cmake/FindLibUnwind.cmake.

I think Advik had rightly used `find_library()` and CMake had resolved it simply from `CONDA_PREFIX`.

kmr-srbh: I think we can manually throw an error if the library is not found.

```cmake
if (NOT LIBUNWIND_LIBRARY)
  message(FATAL_ERROR "libunwind library not found. Please install libunwind.")
endif()
```

kmr-srbh: Yes, we can do that. https://github.com/lfortran/lfortran/pull/5995#discussion_r1913452034 contains link to one such file which we can use if it works for us.

gxyd:
Ah yes, apologies for the quick changes here, I didn't even test them locally before submitting the PR. Probably just rushed it, I'll address the comments left on this.

gxyd:
@kmr-srbh , would you want to take up this issue instead maybe? Because I don't have access to ubuntu machine (I thought I will have it for today, but I was wrong), since `libunwind` is a dependency only on Linux machine, I'll not be able to test it on macOS M1 machine.

kmr-srbh:
Thanks @gxyd! I will work on it. :)

kmr-srbh:
Ready.

cmake/Findlibunwind.cmake
```diff
@@ -0,0 +1,8 @@
+find_path(LIBUNWIND_INCLUDE_DIR libunwind.h)
```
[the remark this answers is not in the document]
kmr-srbh: Done in https://github.com/lfortran/lfortran/commit/0e48cae22ec8eb1c15d90841ccd8cb7f3f9461ae.

cmake/Findlibunwind.cmake
```diff
@@ -0,0 +1,8 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(LIBUNWIND DEFAULT_MSG LIBUNWIND_INCLUDE_DIR LIBUNWIND_LIBRARY)
+
+add_library(p::libunwind INTERFACE IMPORTED)
+set_property(TARGET p::libunwind PROPERTY INTERFACE_INCLUDE_DIRECTORIES
+    ${LIBUNWIND_INCLUDE_DIR})
```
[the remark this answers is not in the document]
kmr-srbh: Done in https://github.com/lfortran/lfortran/commit/0e48cae22ec8eb1c15d90841ccd8cb7f3f9461ae.

Issue #5998, opened 2025-01-13
ci: fix `gcc` version in `Check Release build` job


kmr-srbh:
Currently, we are having to downgrade from the default version `13.3.0` to `11.4.0`. This was done chiefly due to 2 reasons:
- bypass "uninitialized variable" warning which is found and generated due to the use of `-Werror` flag
- `subprocess_aborted` for various integration tests (https://github.com/lfortran/lfortran/actions/runs/12742005151/job/35513044996).

This was due to the use of `gcc 13.3.0` for the release build. Debug builds work fine, but release builds fail.

We need to figure out the way to do a successful release build with the default compiler version on the CI.

kmr-srbh:
Fixing issues with uninitialized variables will help us use the `-Werror` flag again for building. Currently, using this flag causes issues in both debug and release build with the default `gcc` compiler.

PR #6003, opened 2025-01-14
fix: allocate space for null terminator in string allocations

kmr-srbh:
Part of #5998


kmr-srbh:
I am opening a different PR to fix the "uninitialized variables" issue.

PR #6004, opened 2025-01-14
fix: initialize variables to avoid undefined behavior

### LINKED ISSUE #5998: ci: fix `gcc` version in `Check Release build` job

kmr-srbh:
resolves #5998


kmr-srbh:
Since we use the `-Werror` flag to build at the CI, I think we should consider using it for local builds also?

Issue #5938, opened 2025-01-07
Bug : `spread` can't accept `allocatable` arrays


assem2002:
MRE
```f90
program test
  integer, allocatable :: i(:)
  integer :: ii(4,4)
  allocate(i(4))
  ii =  spread(i,dim = 2, ncopies=4) ! Code generation error while trying to pass `i` to `lcompilers_spread`
  print *, spread(i,dim = 2, ncopies=4) ! LLVM IR assertion error while trying to handle `i` that's becuase of the temp created as I think.
end program test
```

Both errors
```Console
code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %array_descriptor = alloca %array, align 8
 %array**  call void @_lcompilers_spread(%array** %i, i32* %call_arg_value, i32* %call_arg_value1, %array* %array_descriptor)



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt3.f90 
lfortran: /home/assem/conda_root/envs/lf/include/llvm/IR/Instructions.h:897: llvm::Type* llvm::checkGEPType(Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
Traceback (most recent call last):
  Binary file "/home/assem/lfortran_test/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 2733, in main
    return main_app(argc, argv);
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 2671, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/home/assem/lfortran_test/lfortran/src/bin/lfortran.cpp", line 1051, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
[...]

PR #6012, opened 2025-01-15
fix: handle allocatable arrays with spread

HarshitaKalani:
Towards: https://github.com/lfortran/lfortran/issues/5938

There's one more issue to be handled.

### DIFF
--- a/integration_tests/intrinsics_342.f90
+++ b/integration_tests/intrinsics_342.f90
@@ -1,0 +1,9 @@
+program intrinsics_342
+    integer, allocatable :: i(:)
+    integer :: ii(4,4)
+    allocate(i(4))
+    i = [1, 2, 3 ,4]
+    ! ii =  spread(i, dim = 2, ncopies=4) ! Code generation error while trying to pass `i` to `lcompilers_spread`
+    print *, spread(i, dim = 2, ncopies=4)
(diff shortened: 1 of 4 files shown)


integration_tests/intrinsics_342.f90
```diff
@@ -0,0 +1,8 @@
+program intrinsics_342
+    integer, allocatable :: i(:)
+    integer :: ii(4,4)
+    allocate(i(4))
+    ! ii =  spread(i, dim = 2, ncopies=4) ! Code generation error while trying to pass `i` to `lcompilers_spread`
+    print *, spread(i, dim = 2, ncopies=4)
```
gxyd, line 6: we haven't assigned any values to the array `i`, so this might fail sometimes and become *flaky* for us. It may pass right now, but I've noticed these kind of tests fails sometimes at the CI.

Also, to test that it works well, I wouldn't use all values as 0's, as re-arrangement of ten 0's in (5, 2) shape will be same as any permutation of 0's in (5, 2) shape, while the same may not be true for re-arrangement of an array with values `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]` in (5, 2) shape.

gxyd (approved):
Looks good to me

gxyd:
> There's one more issue to be handled.

Do you intend to handle that in this PR?

HarshitaKalani:
Will pick that separately.

gxyd:
> Will pick that separately.

Very well then, I'll merge this one. Thanks @HarshitaKalani .

Issue #5990, opened 2025-01-12
Bug : array as arrayIndex in `arrayItem` in `arrayConstrcutror` fail


assem2002:
MRE
```.f90
program prog
  integer :: i
  integer :: arr(10)
  integer :: arr_indecies(10)
  integer,allocatable :: arr_res(:)
  allocate(arr_res(10))

  do i =1,10
    arr(i) = i
    arr_indecies(i) = i
  end do
  arr_res = [arr(arr_indecies),11] ! <<<<<<<<<<<<<<<< Bug
  print *, arr_res
  do i =1,11
    if (arr_res(i) /= i) error stop
  end do
end program prog
```
I think this due to `implied_do_loop.cpp` pass as the resulting temp array that replaces the `arrayConstructor` has a fixedSize of `2`.
```clj
__libasr__created__var__0__array_constructor_:
  (Variable
      2
      __libasr__created__var__0__array_constructor_
      []
      Local
      ()
      ()
      Default
      (Array
          (Integer 4)
          [((IntegerConstant 1 (Integer 4) Decimal)
          (IntegerConstant 2 (Integer 4) Decimal))]
          FixedSizeArray
      )
      ()
      Source
      Public
      Required
      .false.
      .false.
  ),
```

AnshMehta1:
I will work on it .

AnshMehta1:
I have got a fix . Will be implementing it .

AnshMehta1:
For this MRE 
```
program prog
  integer :: i
  integer :: arr(10)
  integer :: arr_indecies(10)
  integer,allocatable :: arr_res(:)
  allocate(arr_res(10))

  do i =1,10
    arr(i) = i
    arr_indecies(i) = i
  end do
  arr_res = [arr(arr_indecies),11] ! <<<<<<<<<<<<<<<< Bug
  print *, arr_res
  do i =1,11
    if (arr_res(i) /= i) error stop
  end do
end program prog
```

I guess , it should be ` allocate(arr_res(11))` instead of ` allocate(arr_res(10))` . My changes is fixing this MRE . Opening a PR in a while

assem2002:
is this issue fixed? @AnshMehta1

AnshMehta1:
> is this issue fixed? @AnshMehta1

1st Mre is fixed but second Mre you posted is left .
I will work on it Tommorow.

Issue #3203, opened 2024-01-25
An example for compile time evaluation


certik:
We can try to compile this example: https://fortran-lang.discourse.group/t/computing-at-compile-time/3044
```fortran
program compile_time
implicit none
integer, parameter :: N = 65535
integer, parameter :: sp = kind(1.0)
integer, parameter :: dp = kind(1.d0)
real(dp), parameter :: pi = 2*asin(1._dp)
real(dp), parameter :: a = 0, b = pi
real(dp), parameter :: dx = (b-a)/N
integer :: i
real(dp), parameter :: X(*) = [(sin(a+(b-a)*i/N), i = 1, N)]
real(dp), parameter :: S = sum(X)*dx
logical, parameter :: l = S < 2
real(kind=merge(sp, dp, l)) :: y
if (kind(y) == sp) then
    print *, "true"
else
    print *, "false"
end if
print *, S
end program
```
And the thread above has other similar examples.

certik:
Current status: the following compiles and runs:
```
program compile_time
implicit none
integer, parameter :: N = 10
integer, parameter :: sp = kind(1.0)
integer, parameter :: dp = kind(1.d0)
real(dp), parameter :: pi = 2*asin(1._dp)
real(dp), parameter :: a = 0, b = pi
real(dp), parameter :: dx = (b-a)/N
integer :: i
real(dp), parameter :: X(N) = [(sin(a+(b-a)*i/N), i = 1, N)]
!real(dp), parameter :: S = sum(X)*dx
!logical, parameter :: l = S < 2
!real(kind=merge(sp, dp, l)) :: y
!if (kind(y) == sp) then
!    print *, "true"
!else
!    print *, "false"
!end if
!print *, S
print *, X
end program
```
But `X` does not have a compile time ArrayConstant associated with it, so the next line fails with:
```console
$ lfortran a.f90           
semantic error: Initialization of `S` must reduce to a compile time constant.
  --> a.f90:11:1
   |
11 | real(dp), parameter :: S = sum(X)*dx
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

There is also a bug with using `real(dp), parameter :: X(*)` instead of `X(N)`, where we get:
```console
$ lfortran a.f90
Traceback (most recent call last):
  File "/Users/ondrej/repos/lfortran/lfortran/src/bin/lfortran.cpp", line 2386
    LCompilers::print_stack_on_segfault();
[...]

PR #6013, opened 2025-01-15
enh : Handling ArrayItem in Array Constructor

AnshMehta1:
Towards #5990 

Size of ArrayConstructor `[arr(arr_indecies),11]` was not getting intialized properly and PR aims to achieve this.
 
MRE :- 
```
program array_76
  integer :: i
  integer :: arr(10)
  integer :: arr_indecies(10)
  integer,allocatable :: arr_res(:)
  allocate(arr_res(11))

  do i =1,10
    arr(i) = i
    arr_indecies(i) = i
  end do

  arr_res = [arr(arr_indecies),11]
  print *, arr_res
end program
```

LFortran Output :- 
```
(lf) lordansh@LordAnsh:~/dev/lfort1/lfortran$ lfortran try.f90
1
2
3
4
5
6
7
8
9
10
11
```


gxyd (approved):
Looks good to me.

gxyd:
As a separate comment, we've `get_ArrayConstructor_size` in both `implied_do_loops.cpp` and in `asr_utils.cpp`, some difference in signature, but we're going maintain both? (Probably needs a separate issue to discuss)

gxyd:
> We can handle this in separate issues as per MREs

Oh no, that might just be refactoring (I'm not sure if it can done or not)

kmr-srbh:
We need to handle this for allocatable arrays also (the case where `arr` is allocatable). If this is not required for PRIMA, we can skip it for now.

Why do we wait until the `implied_do_loops` pass to set the array size correctly? The initial ASR and the passes before it use the incorrect size and this can cause issues in the future.

AnshMehta1:
> We need to handle this for allocatable arrays also (the case where `arr` is allocatable). If this is not required for PRIMA, we can skip it for now.
> 
> Why do we wait until the `implied_do_loops` pass to set the array size correctly? The initial ASR and the passes before it use the incorrect size and this can cause issues in the future.

Array constructor was getting created in implied_do_loops pass with incorrect size , so , modified that pass.

AnshMehta1:
> We need to handle this for allocatable arrays also (the case where `arr` is allocatable). If this is not required for PRIMA, we can skip it for now.
> 
> Why do we wait until the `implied_do_loops` pass to set the array size correctly? The initial ASR and the passes before it use the incorrect size and this can cause issues in the future.

Their is one MRE in #5990 for allocatable arrays. Will handle that in my next PR.

kmr-srbh:
> Array constructor was getting created in implied_do_loops pass with incorrect size , so , modified that pass.

Ah it is created after the body visitor, so the array size before this pass is still incorrect. I think we should probably handle it there to fix this correctly? 

CC @gxyd

AnshMehta1:
This is pass_04
```
! Fortran code after applying the pass: openmp
program prog
implicit none
integer(4), dimension(10) :: arr
integer(4), dimension(10) :: arr_indecies
integer(4), dimension(:), allocatable :: arr_res
integer(4) :: i
allocate(arr_res(10))
do i = 1, 10
    arr(i) = i
    arr_indecies(i) = i
end do
arr_res = [arr(arr_indecies), 11]
print *, arr_res
end program prog
```

This is pass_05
```
! Fortran code after applying the pass: implied_do_loops
program prog
implicit none
integer(4) :: __1__libasr_index_
integer(4) :: __1_t
integer(4), dimension(11) :: __libasr__created__var__0__array_constructor_
integer(4), dimension(10) :: arr
integer(4), dimension(10) :: arr_indecies
integer(4), dimension(:), allocatable :: arr_res
integer(4) :: i
allocate(arr_res(10))
do i = 1, 10
    arr(i) = i
    arr_indecies(i) = i
end do
__1__libasr_index_ = lbound(__libasr__created__var__0__array_constructor_, 1)
do __1_t = lbound(arr_indecies, 1), ubound(arr_indecies, 1)
    __libasr__created__var__0__array_constructor_(__1__libasr_index_) = arr(arr_indecies(__1_t))
    __1__libasr_index_ = __1__libasr_index_ + 1
end do
__libasr__created__var__0__array_constructor_(__1__libasr_index_) = 11
__1__libasr_index_ = __1__libasr_index_ + 1
arr_res = __libasr__created__var__0__array_constructor_
print *, arr_res
end program prog

```

AnshMehta1:
Earlier , `integer(4), dimension(11) :: __libasr__created__var__0__array_constructor_` was `integer(4), dimension(2) :: __libasr__created__var__0__array_constructor_` in pass_05 as array_item was not visited for getting array_constructor_size.
That's why changes made here.

AnshMehta1:
Actually , size of `arr_res = [arr(arr_indecies),11]` was being evaluated as 2 as array_item arr was not getting visited.

kmr-srbh:
Oh no, I am not talking about this. The expression `[arr(arr_indecies),11]` is `ArrayConstructor`. The node can be viewed in the ASR through `--show-asr`. Generally for debugging purposes, the ASR is preferred over the Fortran code output.

AnshMehta1:
> Oh no, I am not talking about this. The expression `[arr(arr_indecies),11]` is `ArrayConstructor`. The node can be viewed in the ASR through `--show-asr`. Generally for debugging purposes, the ASR is preferred over the Fortran code output.

I understant your point but for this assignment `arr_res = [arr(arr_indecies),11]` , we are using a temporary `integer(4), dimension(11) :: __libasr__created__var__0__array_constructor_` in  implied_do_loops pass. Size of this temporary is not getting set correctly which is handled in this pass.

kmr-srbh:
Yes, the fix in itself is correct. :)

I meant to say that we can set the array size correctly for the main `ArrayConstructor` node, so the temporary takes care of itself. As @gxyd pointed out above, we have an ASRUtils function for this, so I think improving it and using it in the body visitor will get us the correct value. In fact the `get_ArrayConstructor_size()` function in the pass is essentially an improved version of the ASRUtils function. I am guessing it may have been added here for #3203.

AnshMehta1:
> Yes, the fix in itself is correct. :)
> 
> I meant to say that we can set the array size correctly for the main `ArrayConstructor` node, so the temporary takes care of itself. As @gxyd pointed out above, we have an ASRUtils function for this, so I think improving it and using it in the body visitor will get us the correct value. In fact the `get_ArrayConstructor_size()` function in the pass is essentially an improved version of the ASRUtils function. I am guessing it may have been added here for #3203.

Your point seems reasonable . In the pass , we are retrieving ArrayConstructor size by `get_ArrayConstructor_size()` defined in pass. So, we can directly modify `get_ArrayConstructor_size()` function of asr_utils and directly use that function in pass instead of redefining that function in pass .

AnshMehta1:
For now , let us keep things as it is but we can make refactoring changes in future PRs.

Issue #5494, opened 2024-11-26
‘kind’ argument of ‘verify’ intrinsic at must be a scalar


HarshitaKalani:
```fortran
program main
	character(len=5) :: string = "hello"
	character(len=1) :: set(2) = ["l", "h"]
	print*, verify(string, set, kind= [4, 4] )
end
```
```console
$ lfortran b.f90
1    2
$ gfortran b.f90 && ./a.out
b.f90:4:35:

    4 |         print*, verify(string, set, kind= [4, 4] )
      |                                          1
Error: ‘kind’ argument of ‘verify’ intrinsic at (1) must be a scalar
```

pjh40:
I think the above list of intrinsics is missing `bound`, `shape`, `size`, `stopped_images`, `ubound` and `ieee_real`.  All of these intrinsic procedures have an optional `KIND` argument that is a scalar integer constant expression.

harperjf:
Thank you @pjh40. The first item in your list should be ```lbound``` not ```bound```.

pjh40:
Sorry, darn autocorrect...

Issue #4379, opened 2024-06-25
Throw error if args to `and`/`or`/`xor` intrinsics is array, must be scalar


HarshitaKalani:
```fortran
program main
    print *, and([1, 2, 3], [1, 2, 3])
end program main
```
```console
$ lfortran a.f90
1 2 3 
$ gfortran a.f90 && ./a.out
a.f90:2:17:

    2 |     print *, and([1, 2, 3], [1, 2, 3])
      |                 1
Error: ‘i’ argument of ‘and’ intrinsic at (1) must be a scalar
```

HarshilShah1804:
@HarshitaKalani can you suggest what can I use here to check whether the input is an array or not ? I have tried `ASR::is_a<ASR::Array_t>`, `extract_dim_value_int`, `extract_n_dims_from_ttype`, but none of them seems to work here

gxyd:
you can check whether an input is an array or not using `ASRUtils::is_array` which underneath it uses `extract_dimensions_from_ttype`, and what you used i.e. `extract_n_dims_from_ttype` should also work.

If `extract_n_dims_from_ttype` isn't working, then it's possible that the `args[0]` (or some other argument) is already not an array, I'll suggest to debug it, I've a feeling that's probably the case if you are checking in `create_And` inline function in `namespace And`

gxyd:
See the screenshot below:

![Screenshot 2025-01-02 at 4 21 53 PM](https://github.com/user-attachments/assets/7d2dfa4b-6bc7-424f-8859-45ae0b53c4c2)

HarshilShah1804:
> you can check whether an input is an array or not using `ASRUtils::is_array` which underneath it uses `extract_dimensions_from_ttype`, and what you used i.e. `extract_n_dims_from_ttype` should also work.
> 
Yes, I have checked that also but it wasn't working.

> If extract_n_dims_from_ttype isn't working, then it's possible that the args[0] (or some other argument) is already not an array, I'll suggest to debug it, I've a feeling that's probably the case if you are checking in create_And inline function in namespace And

Yes, I was modifying the create_And function in the namespce Ans only.

gxyd:
You can refer GFortran's documentation for 'And' intrinsic here: https://gcc.gnu.org/onlinedocs/gfortran/AND.html, which will give some idea for what is allowed and what isn't.

gxyd:
[...]

PR #6014, opened 2025-01-15
fix: throw error for non scalar args in or/and/xor and non scalar kind arg

HarshitaKalani:
Fixes: https://github.com/lfortran/lfortran/issues/4379
Fixes: https://github.com/lfortran/lfortran/issues/5494

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -6001,7 +6001,16 @@ public:
         for( size_t i = 0; i < x.n_args; i++ ) {
             this->visit_expr(*x.m_args[i].m_end);
             args.p[i] = ASRUtils::EXPR(tmp);
-            if(ASRUtils::is_descriptorString(ASRUtils::expr_type(args.p[i]))){
+            if (intrinsic_name == "and" || intrinsic_name == "or" || intrinsic_name == "xor") {
+                if( ASRUtils::is_array(ASRUtils::expr_type(args[i]))) {
+                    diag.add(diag::Diagnostic(
+                    "arguments of `" + intrinsic_name + "` intrinsic must be scalar",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("", {x.base.base.loc})}));
+                    throw SemanticAbort();
+                }
+            }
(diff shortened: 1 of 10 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6589,19 +6598,44 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            if (args[3]) {
+                if (ASRUtils::is_array(ASRUtils::expr_type(args[3]))) {
+                    diag.add(Diagnostic(
+                        "`kind` argument of `" + name + "` intrinsic must be a scalar",
+                        Level::Error, Stage::Semantic, {
+                            Label("",{args[3]->base.loc})
+                        }));
+                    throw SemanticAbort();
+                }
+            }
```
gxyd, line 6638: we've some repeated code here, whenever a new intrinsic with new position of `kind` argument is added, we'll have to re-use: `std::find(optional_kind_arg_at_index_3.begin(), optional_kind_arg_at_index_3.end(), name) != optional_kind_arg_at_index_3.end() ... { } ` code block.

Instead of above, why don't we save an unordered map (something like):
```cpp
std::unordered_map<std::string, int> kind_arg_index_map = {
        {"logical", 1}, {"storage_size", 1}, {"anint", 1}, {"nint", 1}

...
...
}
```

and then we can iterate over `intrinsic` name and kind_index over it. And will have to raise error at just one place.

gxyd (approved):
Otherwise looks good to me.

src/lfortran/semantics/ast_common_visitor.h
HarshitaKalani: Yes, thanks for the suggestion. I'll do that.

tests/errors/intrinsics17.f90
```diff
@@ -0,0 +1,3 @@
+program intrinsics17
+    print *, and([1, 2, 3], [1, 2, 3])
+end program
```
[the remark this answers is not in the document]
HarshitaKalani: Sure

gxyd (approved):
Looks good now, thanks @HarshitaKalani

PR #6016, opened 2025-01-15
CI: update CI for `cobyla_example_1` PRIMA

jinangshah21:


gxyd:
Since there is an LLVM 19 CI failure, which I believe you are already looking into, so I'll move this PR to draft mode for now.

Issue #6020, opened 2025-01-15
Run all integration tests and 3rd party tests with `--std=f23`


certik:
There are other combinations of options possible, but being able to just test the default and `--std=f23` should cover most cases. In order to save CI resources, we can test this on just Ubuntu and just one LLVM version.

Currently we only test the options on a few tests, but we need to test all of them, to prevent bugs like https://github.com/lfortran/lfortran/issues/6017.

Third party repositories:

- [ ] Numerical Methods Fortran (PR: https://github.com/lfortran/lfortran/pull/7385)
- [ ] POT3D (PR: https://github.com/lfortran/lfortran/pull/7384)
- [ ] PRIMA (PR: https://github.com/lfortran/lfortran/pull/7388)
- [ ] Legacy Minpack (SciPy) (PR: https://github.com/lfortran/lfortran/pull/7391)
- [ ] Modern Minpack (Fortran-Lang)
- [ ] Modern Minpack (check results)
- [ ] dftatom
- [ ] fastGPT ( ubuntu-latest )
- [ ] fastGPT ( macos-latest )
- [ ] fpm
- [ ] stdlib
- [ ] SNAP (tracking issue: https://github.com/lfortran/lfortran/issues/7445)

PR #6022, opened 2025-01-15
enh (tests): add support for running integration tests with the requested language standard

kmr-srbh:
Part of #6020 

We introduce a new `--std=TEXT` option in the `run_tests.py` script.
  - The option controls 2 boolean CMake variables `STD_F23` and
    `STD_LEGACY`.
  - Full support is added for testing with the `f23` standard.
  - Preliminary groundwork is laid for supporting `legacy` standard
    in the future.


kmr-srbh:
I ran it locally and we have some failing tests.

PR #6027, opened 2025-01-16
fix: use fixed size ArrayConstructor directly instead of ArrayPhysicalCast

### LINKED ISSUE #5996: (PRIMA) Handle `ArrayConstructor` with Array passed as argument in function
MRE 1 : 
```fortran
program test
    real :: x(4) = [1.0, 2.0, 3.0, 4.0]
    print *, maxval([1.0,x])
end program
```
MRE 2 : 
```fortran
program test
    real :: x(4) = [1.0, 2.0, 3.0, 4.0]
    print *, max_func([1.0,x])
contains 
    function max_func(x) result(y)
      real, intent(in) :: x(:)
      real :: y
      y = maxval(x)
    end function
end program
```

GFortran Output : `4.00000000`
LFortran Output : `1.00000000e+00`

jinangshah21:
Fixes #5996

### DIFF
--- a/src/libasr/pass/implied_do_loops.cpp
+++ b/src/libasr/pass/implied_do_loops.cpp
@@ -460,6 +460,7 @@ class ReplaceArrayConstant: public ASR::BaseExprReplacer<ReplaceArrayConstant> {
     }
 
     void replace_ArrayPhysicalCast(ASR::ArrayPhysicalCast_t* x) {
+        bool is_arr_construct_arg = ASR::is_a<ASR::ArrayConstructor_t>(*x->m_arg);
         ASR::BaseExprReplacer<ReplaceArrayConstant>::replace_ArrayPhysicalCast(x);
         if( ASRUtils::use_experimental_simplifier ) {
             if( x->m_old != ASRUtils::extract_physical_type(ASRUtils::expr_type(x->m_arg)) ) {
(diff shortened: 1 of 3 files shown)


src/libasr/pass/implied_do_loops.cpp
```diff
@@ -460,11 +460,16 @@ class ReplaceArrayConstant: public ASR::BaseExprReplacer<ReplaceArrayConstant> {
...
+        [[maybe_unused]] bool is_arr_construct_arg = ASR::is_a<ASR::ArrayConstructor_t>(*x->m_arg);
         ASR::BaseExprReplacer<ReplaceArrayConstant>::replace_ArrayPhysicalCast(x);
         if( ASRUtils::use_experimental_simplifier ) {
             if( x->m_old != ASRUtils::extract_physical_type(ASRUtils::expr_type(x->m_arg)) ) {
                 x->m_old = ASRUtils::extract_physical_type(ASRUtils::expr_type(x->m_arg));
             }
+            if( (is_arr_construct_arg && ASRUtils::is_fixed_size_array(ASRUtils::expr_type(x->m_arg))) ){
+                *current_expr = x->m_arg;
+                return;
+            }
```
gxyd, line 472: Do the tests fail if we take this condition out and enable it for `--no-experimental-simplifier` as well? If the tests continue to pass, I think we can use this condition for that as well.

gxyd (approved):
This looks good to me. I'm guessing the tests fail when the condition is enabled for `--no-experimental-simplifier`, probably that's the case.

src/libasr/pass/implied_do_loops.cpp
gxyd: Just checked it locally, seems like the tests do fail, so this is no problem.

jinangshah21: No if we keep it outside then it fails for `--no-experimental-simplifier`

jinangshah21:
> This looks good to me. I'm guessing the tests fail when the condition is enabled for `--no-experimental-simplifier`, probably that's the case.

Yes it fails without Simplifier.

Issue #3204, opened 2024-01-25
Add exhaustive test for elemental functions


certik:
The issue https://github.com/lfortran/lfortran/issues/3182 revealed that our tests are not complete. Here is what we need to test for an elemental function `f(a, b)` (and the same for a subroutine):

* `f([1.], [1.])`
* `f([1.], 1.)`
* `f(1., [1.])`
* `f(1., 1.)`

And then:

* `f([1.], [1.], 1.)`
* `f([1.], 1., 1.)`
* `f(1., [1.], 1.)`
* `f(1., 1., 1.)`
* `f([1.], [1.], [1.])`
* `f([1.], 1., [1.])`
* `f(1., [1.], [1.])`
* `f(1., 1., [1.])`

Then we need to do the same and replace `[1.]` with `[1., 2., 3.]` and finally replace it with a 2D array, and then a 3D. That should do it.

dsaw:
I can take a look at this if possible.

certik:
@dsaw don't post into issues if you can take a look. You can take a look at any issue. Just send a PR.

PR #6029, opened 2025-01-16
test: add test for elemental function for 1D array

Jaysukh-409:
Towards #3204 

- Add a test for a elemental function with array parameter as `[ 1.0, 2.0, 3.0 ]` and scalar parameter as `1.0`.


gxyd (approved):
Looks good to me. Thanks @Jaysukh-409

Issue #6024, opened 2025-01-16
'get_environment_variable' intrinsic subroutine always gives the same value


gxyd:
## Description

For the below Fortran program:
```f90
PROGRAM test_getenv
    character(len=255) :: pwd
    character(len=255) :: homedir
    character(len=255) :: my_env_var
    CALL get_environment_variable("PWD", pwd)
    CALL get_environment_variable("HOME", homedir)
    CALL get_environment_variable("MY_ENV_VAR", my_env_var)
    print *, pwd
    print *, homedir
    print *, my_env_var
END PROGRAM
```

the output produced by LFortran is:
```bash
# version of LFortran being used
> lfortran --version
LFortran version: 0.43.0-129-g5294decf7-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

# to create a new environment variable 'MY_ENV_VAR'
> echo 'export MY_ENV_VAR="HelloWorld"' >> ~/.zshrc
> source ~/.zshrc
> echo $MY_ENV_VAR
HelloWorld

# running the program 'test.f90' (the one I mentioned above)
> lfortran test.f90
/Users/gxyd
/Users/gxyd
/Users/gxyd
```

while the output produced by GFortran is:
```bash
> gfortran test.f90 -o a.out && ./a.out
 /Users/gxyd/OpenSource/lfortran                                                                                                                                                                                                                                
 /Users/gxyd                                                                                                                                                                                                                                                    
 HelloWorld
```

which seems correct

gxyd:
I'll take a look for why this isn't working correctly.

PR #6030, opened 2025-01-16
fix: 'get_environment_variable' intrinsic subroutine

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6024

This isn't ready for a review, because even though this PR gives correct output for issue https://github.com/lfortran/lfortran/issues/6024, there is one underlying question here: https://github.com/lfortran/lfortran/pull/6010#discussion_r1917799894, which needs to be addressed.


gxyd:
Third party test fails in *stdlib*, I'll extract the MRE, shouldn't be difficult here.

Issue #5912, opened 2025-01-04
cshift giving wrong answers


HarshilShah1804:
Cshift function is giving wrong answers
```fortran
program main
    integer, allocatable :: b(:)
    allocate(b(3))
    b = reshape([1,2,3], shape=[3])
    b = cshift(b, 2)
    print *, b
end program
```
```
$ lfortran a.f90
3
3
3

$ gfortran a.f90 && ./a.out
           3           1           2
```

PR #6033, opened 2025-01-16
Fix ``cshift`` for 1D case

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/5912

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -429,7 +429,7 @@ RUN(NAME functions_26 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_27 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_29 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


HarshitaKalani (approved):
Looks great! Thanks @czgdp1807

czgdp1807:
@HarshitaKalani Now we need to focus on multi-dimensional case.

1. If shfit is an array then input array can only be of 2 rank.
2. If shift is a scalar then input array can be of any rank.

We need to first implement `cshift` by hand in Fortran and check if our generalised are logics are correct. Define a user function, say (`cshift_manual`) and compare its results with intrinsic `cshift`.

One test case,

```fortran
program test_cshift
    integer, dimension(3,5, 2) :: a
    a = reshape( (/ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 /), (/ 3, 5, 2 /))
    print '(3i3)', a(1,:, :)
    print '(3i3)', a(2,:, :)
    print '(3i3)', a(3,:, :)
    a = cshift(a, SHIFT=1, DIM=3)
    print *
    print '(3i3)', a(1,:, :)
    print '(3i3)', a(2,:, :)
    print '(3i3)', a(3,:, :)
end program test_cshift
```

HarshitaKalani:
Yes. Thanks for the help @czgdp1807 :). I'll look into handling the `dim` argument for the array intrinsics starting with `cshift`.

Issue #5969, opened 2025-01-11
ieee_arithmetic provided dp but shouldn't.


harperjf:
This program is the same one as in #3789 but it now exposes a different lfortran error: pretending falsely that I had used an ```only``` clause.
```
program infnan
  use,intrinsic:: ieee_arithmetic
  implicit none
  integer     ,parameter:: dp = kind(1d0)
  character(*),parameter:: fmt = "(A,4L2)" ! lfortran can't use (A,:,4L2)
  real(dp):: xdp(4)
  xdp = [ ieee_value(1d0, ieee_positive_inf),&
       ieee_value(1d0, ieee_negative_inf),&
       ieee_value(1d0, ieee_quiet_nan),&
       ieee_value(1d0, ieee_signaling_nan) ]

  print fmt,repeat(' ',16)//'Inf NaN'
  print fmt,repeat(' ',16)//'+ - Q S' 
  print fmt,'isinfdp(xdp)   ',isinfdp(xdp)
  print fmt,'isposinfdp(xdp)',isposinfdp(xdp)
  print fmt,'isneginfdp(xdp)',isneginfdp(xdp)
  print fmt,'isnandp(xdp)   ',isnandp(xdp)
  print fmt,'isfinitedp(xdp)',isfinitedp(xdp)
contains

  elemental logical function isnandp(x)
    real(dp),intent(in)::            x
    isnandp = (x/=x)
  end function isnandp

  elemental logical function isinfdp(x)
    real(dp),intent(in)::            x
    isinfdp = (abs(x)>huge(x))
  end function isinfdp

  elemental logical function isposinfdp(x)
    real(dp),intent(in)::            x
    isposinfdp = (x>huge(x))
  end function isposinfdp

  elemental logical function isneginfdp(x)
    real(dp),intent(in)::            x
    isneginfdp = (x<-huge(x))
  end function isneginfdp

  elemental logical function isfinitedp(x)
    real(dp),intent(in)::            x
    isfinitedp = (abs(x)<=huge(x))
  end function isfinitedp

end program infnan
```
Result of l;fortran 0.43.0:
```
semantic error: Symbol is already declared in the same scope
 --> infnan.f90:5:28
  |
5 |   integer     ,parameter:: dp = kind(1d0)
  |                            ^^^^^^^^^^^^^^ redeclaration

 --> /home/john/lfortran-0.43.0/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90:2:60
  |
[...]

PR #5873, opened 2024-12-30
fix (AST -> ASR): import `ExternalSymbol` with `m_name` instead of `m_original_name`

kmr-srbh:
fixes #5870 

Using `m_original_name` causes incorrect variable names for aliases.

Issue #5870, opened 2024-12-29
Importing a kind from a module does not work


certik:
MRE:
```fortran
module a
use iso_fortran_env, only: dp => real64
end module

program test_dp
use a
integer(dp) :: i
end program
```
This gives:
```console
$ gfortran a.f90 
$ lfortran a.f90 
semantic error: Variable 'dp' is not declared
 --> a.f90:7:9
  |
7 | integer(dp) :: i
  |         ^^ 'dp' is undeclared
```

harperjf:
Providing ```, only: dp``` made the program compile. I also changed it to use ```dp``` as a real kind:
```
module a
use iso_fortran_env, only: dp => real64
end module a

program test_dp
use a, only: dp
real(dp) :: x = 42
print *,x
end program test_dp
```
Output with lfortran 0.42.0 was ```4.20000000000000000e+01```

certik:
Thanks @harperjf  for the details. I think this bug will not be difficult to fix.

PR #6039, opened 2025-01-17
fix: ieee_arithmetic providing sp and dp

Jaysukh-409:
Fix: #5969

- In `ieee_arithmetic` instead of `sp` and `dp` used `real32` and `real64`.


integration_tests/test_ieee_arithmetic_02.f90
```diff
@@ -0,0 +1,6 @@
+program dpcheck
```
gxyd, line 1: We use program names same as the test program name i.e. in this case `program test_ieee_arithmetic_02`

tests/tests.toml
```diff
@@ -4427,4 +4427,8 @@ asr = true
 
 [[test]]
 filename = "../integration_tests/polymorphic_arguments_02.f90"
+asr = true
+
+[[test]]
+filename = "../integration_tests/test_ieee_arithmetic_02.f90"
```
gxyd, line 4433: We probably don't need to test this in `tests/tests.toml`, cause testing it as a part of integration tests, makes sure that the semantic error doesn't occur.

gxyd (commented):
Other the above two comment, this looks good. The tests probably will fail right now, but once you undo the changes made in this PR to `tests/tests.toml`, I think it will be fine.

gxyd (approved):
Looks good now, thanks @Jaysukh-409 .

btw, you can always leave it upto the review to *resolve conversations*, that helps me look at it quickly, and if it looks good to me, I'll resolve them.  😄

gxyd:
I squashed the commits, as one commit is good for this PR.

src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90
```diff
@@ -1,5 +1,5 @@
 module lfortran_intrinsic_ieee_arithmetic
```
kmr-srbh, line 1: This is not the actual issue here. The root cause of the issue stems from the way we import the symbols from a module as `ExternalSymbol`. I am explaining the problem.

kmr-srbh: For the below Fortran program derived from the MRE in #5870:

```fortran
module a
   use iso_fortran_env, only: dp => real64
end module

program test_dp
   use a
   implicit none
   print *, real64
end program
```
the output from GFortran and an older release version of LFortran (v0.42.0) is:

```console
(lf_release) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
test.f90:8:18:

    8 |    print *, real64
      |                  1
Error: Symbol ‘real64’ at (1) has no IMPLICIT type
(lf_release) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
8
```

It was because we imported the symbol using it's original name, which in this case was `real64`. This was not the correct behavior due to the use of the alias. It was fixed in #5873 and the symbol is now imported with name `dp`.

`lfortran_intrinsic_ieee_arithmetic` imports `real32` and `real64` with aliases `sp` and `dp`. Now when the user imports the `ieee_arithmetic` module completely, the symbol `sp` and `dp` also gets imported into the scope. This causes issue when a user declares a variable of the name same as the alias.

Renaming, like the changes in this PR, will still cause the same issue, just not for variable names `sp` and `dp`.

```fortran
program main
   use, intrinsic :: ieee_arithmetic
   implicit none
   integer, parameter :: real32 = 42
   print *, real32
end program main
```
```console
          42
semantic error: Symbol is already declared in the same scope
 --> test.f90:4:26
  |
4 |    integer, parameter :: real32 = 42
  |                          ^^^^^^^^^^^ redeclaration

 --> /home/saurabh-kumar/Projects/System/lfortran/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90:2:46
  |
2 |     use, intrinsic :: iso_fortran_env, only: real32, real64
  |                                              ~~~~~~ original declaration


```

kmr-srbh: I am not sure what should be a correct fix for this as this behavior is disallowed for general modules:

```fortran
module super
   integer, parameter :: super_dp = 4
end module super

module base
   use super, only: base_dp => super_dp
   integer(base_dp) :: i
end module base

module main
   use base
   integer, parameter :: base_dp = 42
end module main
```
```console
test.f90:12:32:

   11 |    use base
      |       2                         
   12 |    integer, parameter :: base_dp = 42
      |                                1
Error: Symbol ‘super_dp’ at (1) conflicts with symbol from module ‘super’, use-associated at (2)
semantic error: Symbol is already declared in the same scope
  --> test.f90:12:26
   |
12 |    integer, parameter :: base_dp = 42
   |                          ^^^^^^^^^^^^ redeclaration
   |
 6 |    use super, only: base_dp => super_dp
   |                     ~~~~~~~~~~~~~~~~~~~ original declaration


```

Probably, I think we should improve how we import things from an `IntrinsicModule`.

Pranavchiku: > the symbol sp and dp also gets imported into the scope.

Which should be imported unless you set those as private.

```fortran
module xx
    use iso_fortran_env, only: dp => real64
end module

program main
    use xx, only : dp
    implicit none
    print *, kind(dp)
end program
```

```console
% lfortran c.f90   
4
% gfortran c.f90 && ./a.out
           4
```

Pranavchiku: > Renaming, like the changes in this PR, will still cause the same issue, just not for variable names sp and dp.


Yes, I know this but this is towards right direction and if not a fix, it is towards a fix, not even a workaround I would say. The actual fix will be to remove global import and import it inside where we use simple as that.

Pranavchiku (changes requested):
```diff
diff --git a/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90 b/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90
index 0a25d58be..0a0e0b617 100644
--- a/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90
+++ b/src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90
@@ -1,5 +1,4 @@
 module lfortran_intrinsic_ieee_arithmetic
-    use, intrinsic :: iso_fortran_env, only: real32, real64
     implicit none
 
     type ieee_class_type
@@ -64,6 +63,7 @@ module lfortran_intrinsic_ieee_arithmetic
     contains
 
     elemental function spieee_class(x) result(y)
+        use iso_fortran_env, only: real32
         real(real32), intent(in) :: x
         type(ieee_class_type) :: y
     end function
```

@Jaysukh-409 do it like this, it will solve the problem mentioned in https://github.com/lfortran/lfortran/pull/6039/files#r1920080258

src/runtime/pure/lfortran_intrinsic_ieee_arithmetic.f90
Pranavchiku: > Probably, I think we should improve how we import things from an IntrinsicModule.

The comment: https://github.com/lfortran/lfortran/pull/6039#pullrequestreview-2559142284 fixes it

```fortran
program dpcheck
use, intrinsic:: ieee_arithmetic
integer, parameter :: real32 = kind(1.0e0)
print *, 1/3.0_dp
end program dpcheck
````

```console
% lfortran a.f90 
3.33333333333333315e-01
```

kmr-srbh: Thanks for clarifying this. I thought we handled `IntrinsicModule` differently.

Pranavchiku (approved):
I think it is good, thank you!

Issue #4002, opened 2024-05-14
dshiftr not implemented


harperjf:
This program shows that ```dshiftr``` is not implemented, though the corresponding program with ```dshiftl```compiled and ran correctly.
``` 
(lf2) john:~/Test$ cat dshiftrtest.f90
program dshiftrtest
  integer,parameter:: ik1 = selected_int_kind(1)
  integer(ik1),parameter:: i0 = 0, i1 = 1
  print "(A,I0)", &
       'kind(i1)               = ',kind(i1),&
       'dshiftr(i1,i0,2)       = ',dshiftr(i1,i0,2)
end program dshiftrtest
[1]+  Done                    emacs dshiftrtest.f90
(lf2) john:~/Test$ lfortran dshiftrtest.f90
semantic error: Function 'dshiftr' not found or not implemented yet (if it is intrinsic)
 --> dshiftrtest.f90:6:36
  |
6 |        'dshiftr(i1,i0,2)       = ',dshiftr(i1,i0,2)
  |                                    ^^^^^^^^^^^^^^^^ 
```

PR #6043, opened 2025-01-17
fix: implementation of `dshiftr` for different kind value

Jaysukh-409:
Fixes: #4002

- Change the `k_val` variable for different kind of integer as `k_val = kind * 8` to ensure it's correctness with gfortran results.
- Updated the `dshiftr` evaluation implementation.

### DIFF
--- a/integration_tests/test_dshiftr.f90
+++ b/integration_tests/test_dshiftr.f90
@@ -1,0 +1,6 @@
+program test_dshiftr
+  integer, parameter:: ik1 = selected_int_kind(1)
+  integer(ik1), parameter:: i0 = 0, i1 = 1
+  integer(ik1):: expected = 64
+  if (abs(dshiftr(i1, i0, 2) - expected) > 1e-6) error stop
+end program test_dshiftr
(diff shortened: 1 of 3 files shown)


integration_tests/test_dshiftr.f90
```diff
@@ -0,0 +1,6 @@
+program test_dshiftr
+  integer, parameter:: ik1 = selected_int_kind(1)
+  integer(ik1), parameter:: i0 = 0, i1 = 1
+  integer(ik1):: expected = 64
+  if (abs(dshiftr(i1, i0, 2) - expected) > 1e-6) error stop
```
Pranavchiku, line 5:
```suggestion
  if (dshiftr(i1, i0, 2) /= expected) error stop
```

For integers we can do exact comparison, for real numbers due to floating point precision errors we need to compare difference.

Pranavchiku (approved):
I think this is good!

Pranavchiku:
Please avoid doing `git merge main` instead do `git rebase main`.

Issue #6046, opened 2025-01-17
Two bugs reading 'NaN'


harperjf:
These two programs both try to read 'NAN' into a real variable in the same way, but ```readnan1.f90``` does it in a subroutine and ```readnan2.f90``` does it in the main line. #3789 may be relevant here although it used IEEE and the present programs do not.

Both programs are wrong in different ways with lfortran 0.44.0.
Both programs are right with gfortran. 

readnan1.f90:
```
! Fortran 2003 free source form program readnan1.f90 by J F Harper
program readnan1
  call testnan
contains
  subroutine testnan
    real:: xnan
    character:: cni*4, msg*200
    integer :: ios
    cni = 'NAN '
    read(cni,*,iostat=ios,iomsg=msg) xnan
    if(ios/=0) then
       print *,"Reading 'NAN ' into xnan failed."
       print *,'iomsg = '//trim(msg)
    else
       print "(A,L2,1X,F0.0)",' NaN/=NaN?',xnan/=xnan,xnan
    end if
  end subroutine testnan
end program readnan1
```
Lfortran 0.44.0 output:
```
 Reading 'NAN ' into xnan failed.
 iomsg = 
```
readnan2.f90:
```
! Fortran 2003 free source form program readnan2.f90 by J F Harper
program readnan2
  real:: xnan
  character:: cni*4, msg*200
  integer :: ios
  cni = 'NAN '
  read(cni,*,iostat=ios,iomsg=msg) xnan
  if(ios/=0) then
     print *,"Reading 'NAN ' into xnan failed."
     print *,'iomsg = '//trim(msg)
  else
     print "(A,L2,1X,F0.0)",' NaN/=NaN?',xnan/=xnan,xnan
  end if
end program readnan2
```
Lfortran 0.44.0 output:
```
 NaN/=NaN? F -9223372036854775808.n
 ```
 Gfortran output from both programs:
 ```
  NaN/=NaN? T NaN
 ```

**Comment by kmr-srbh** [comment]
Regarding `readnan2`, reading `NaN` is correct:

```fortran
program readnan2
   real:: xnan
   character:: cni*4
   cni = 'NAN '
   read(cni,*) xnan
   print *, 'NaN/=NaN?',xnan/=xnan, xnan
end program readnan2
```
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
NaN/=NaN?    F              nan
```

With this program, we have the following 2 bugs:
[...]

PR #2283, opened 2023-08-30
Perform ordered comparison for real numbers

Pranavchiku:
Fixes: https://github.com/lfortran/lfortran/pull/2268#issuecomment-1696854990

PR #2240, opened 2023-08-24
Use `fcmp oeq` to handle NaN like GFortran

certik:

PR #6050, opened 2025-01-18
fix: do unordered comparison for `/=` to handle `NaN`

kmr-srbh:
Resolves floating point comparison issue in #6046.

We previously used the `fcmp one` instruction which does
'ordered' comparison and yields `false` if either
operand is `QNaN`. This was incorrect behavior given the
IEEE 754 standard for comparison between `NaN`.

We now use the `fcmp une` instruction which yields `true`
if either operand is a `QNaN`.

Reference:
  [1] https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN
  [2] https://llvm.org/docs/LangRef.html#fcmp-instruction


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5941,7 +5941,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 break;
             }
             case (ASR::cmpopType::NotEq) : {
-                tmp = builder->CreateFCmpONE(left, right);
+                tmp = builder->CreateFCmpUNE(left, right);
```
Pranavchiku, line 5944: I remember we faced similar issue while compiling SciPy: https://github.com/lfortran/lfortran/pull/2240, so this looks like a correct fix

Pranavchiku: Can you test with one nan and other real value?

kmr-srbh: Yes.

```fortran
   print *, 'NaN /= NaN?', xnan /= 3.0
```
```console
 NaN /= NaN? T
NaN /= NaN?    T
```

For the other binary operators, doing unordered comparison will give incorrect answer as the `fcmp u*` instructions check for `QNAN` using the `or` comparison, while the `fcmp o*` ones use `and`. Given the standard for NaN comparison, the `fcmp une` instruction fits. It was rightly fixed in #2283.

Pranavchiku: And also SciPy works without breaking anything so I think it is fine.

Pranavchiku (approved):
I think changes in this PR is good, thank you!

PR #6065, opened 2025-01-20
feat: Support all types of indices in `ArraySection` + `Arrayitem` on lhs and rhs

jinangshah21:


src/libasr/pass/simplifier.cpp
```diff
@@ -1677,8 +1678,14 @@ class ReplaceExprWithTemporary: public ASR::BaseExprReplacer<ReplaceExprWithTemp
...
-            if( exprs_with_target.find(*current_expr) == exprs_with_target.end() &&
-                !is_assignment_target_array_section_item ) {
+            ASR::expr_t* temp = x->m_v;
+            ASR::symbol_t* rhs_var_sym = nullptr;
+            if( ASR::is_a<ASR::Var_t>(*temp) ) {
+                rhs_var_sym = ASR::down_cast<ASR::Var_t>(temp)->m_v;
+            }
+            if( (exprs_with_target.find(*current_expr) == exprs_with_target.end() &&
+                !is_assignment_target_array_section_item) || 
+                (rhs_var_sym && lhs_var_sym && lhs_var_sym == rhs_var_sym)) {
```
czgdp1807, line 1688: I have already implemented a function for checking this. Please see, `is_common_symbol_present_in_lhs_and_rhs` and use it here.

czgdp1807 (changes requested):

src/libasr/pass/simplifier.cpp
jinangshah21: Done.

jinangshah21:
It is failing for PRIMA with LLVM 19. Let me check and create workaround for that.

jinangshah21:
I have reverted a workaround from PRIMA which was causing LLVM 19 failure for this PR (I will create MRE for that also) and is working in current main.

PR #6070, opened 2025-01-20
fix: remove tokenizerError from tokenizer.re file

parth121101:


parth121101:
@certik we need to get rid of all TokenizerErrors and add those errors in diagnostics followed by throwing TokenizerAbort, right??

PR #6073, opened 2025-01-21
fix : Handling array with array size passed through subroutine and cause fix-sized array declaration

AnshMehta1:


kmr-srbh:
The changes here are not correct actually. This will work only when the subroutine parameter and the variable declared in the outer scope are equal. It worked here because both the variables share the same name `x`.

AnshMehta1:
> The changes here are not correct actually. This will work only when the subroutine parameter and the variable declared in the outer scope have the same name. It worked here because both the variables share the name `x`.

Ya , their are other issues also with this PR . Its a temporary fix. I am working on it and will improve it in subsequent PR.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1668,6 +1668,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
                             end = get_transformed_function_call(end_sym);
+                    } else if (ASR::is_a<ASR::Variable_t>(*end_sym)) {
+                        ASR::Variable_t* end_sym_varible = ASR::down_cast<ASR::Variable_t>(end_sym);
+                        ASR::symbol_t* end_parent_sym = current_scope->parent->get_symbol(end_sym_varible->m_name);
+                        if (in_Subroutine && end_sym_varible->m_value == nullptr && end_parent_sym != nullptr && ASR::is_a<ASR::Variable_t>(*end_parent_sym)) {
+                            ASR::Variable_t* end_parent_sym_variable = ASR::down_cast<ASR::Variable_t>(end_parent_sym);
+                            if (end_parent_sym_variable != nullptr && end_parent_sym_variable->m_value != nullptr && ASR::is_a<ASR::IntegerConstant_t>(*end_parent_sym_variable->m_value)) {
+                                end = end_parent_sym_variable->m_value;
+                            }
+                        }
```
kmr-srbh, line 1679: This will work only when the variable names are same. 

It took me a while with this PR and I think this is tricky because we have some other issues also. When we pass `x` to `temp`, the back-end passes a pointer to `x` in the function call. The function `temp` also expects a pointer to `x` and this is correct because we want to pass by reference for intent `inout`. However, this also means that modifying this value will affect all the other places which use it, like the dimension here. 

Due to the above reason, I think we cannot set the dimension directly using `x` as a change in value will affect the dimension in other operations later, like printing. In some ASR pass we can transform the code to something like:

```fortran
program test 
   integer :: x
   x = 2
   call temp(x, x)

contains
   subroutine temp(x, y)
     integer, intent(inout) :: x
     integer, intent(in) :: y
     integer :: keep(y)
     keep = [1,2]
     print *, keep
     x = 1
     print *, keep
   end subroutine
end program
```

Currently this does not work correctly because we pass `y` also by reference and enforce the intent in the ASR, like not allowing assignment to variable. So the value of `y` gets affected due to the change in value of `x`.

```console
1
2

1
```

kmr-srbh: Looks like GFortran, ifx and flang also modify the intent `in` variable like this!

```fortran
program main
   integer :: var
   var = 2
   call temp(var, var)

contains
   subroutine temp(x, y)
      integer, intent(inout) :: x
      integer, intent(in) :: y
      print *, "x: ", x
      print *, "y: ", y
      x = 1
      print *, "x: ", x
      print *, "y: ", y
   end subroutine
end program
```
```console
 x:            2
 y:            2
 x:            1
 y:            1
```

This is unlike pass by value in C, so I think the above method will also not work.

Pranavchiku:
The way to fix is to create a deep copy of register corresponding to `x` inside `asr_to_llvm` and use it to declare size instead of `x`.

PR #5791, opened 2024-12-22
refactor (ASR): add field `ttype* member_types` to `StructType`

kmr-srbh:
Part of #4263

In this PR, we introduce the new `member type` field to `StructType` which stores the `ttype` of the `Struct` members. The derived symbol is still present as part of the node though. I will work on refactoring it in the upcoming PRs.

PR #6075, opened 2025-01-21
Syncing Struct and StructType with LPython

tanay-man:


src/libasr/ASR.asdl
```diff
@@ -193,7 +193,7 @@ ttype
     | Set(ttype type)
     | List(ttype type)
     | Tuple(ttype* type)
-    | StructType(symbol derived_type)
+    | StructType(ttype* data_member_types, ttype* member_function_types, bool is_cstruct, symbol derived_type)
```
kmr-srbh, line 196: Porting this design to LFortran is not feasible actually. Please see https://github.com/lfortran/lfortran/pull/5791#issuecomment-2561364819. The test failures in `derived_types_06.f90` and `modules_31.f90` are the same.

PR #6085, opened 2025-01-23
enh: add support for `status` parameter to `close` statement

### LINKED ISSUE #6082: close(unit,status='delete') regression
The program below gave correct output in my Ubuntu system with lfortran 0.42.0 but 0.43.0 and 0.44.0 were wrong in different ways. Files with names beginning ```xxx_``` were deleted manually just before compiling the program (otherwise files xxx_b, xxx_c, ... might have been created.)
```
program newfilemin ! Do files still exist after closing with delete?
  implicit none    ! newfile.f90 used execute_command_line, avoided here 
  character(:),allocatable:: xxxfile, letters
  integer :: n, newu
  logical :: ex
  letters = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
  do n = 1,52
     xxxfile='xxx_'//letters(n:n)
     inquire(file=xxxfile,exist=ex)
     if(.not.ex) exit
     if(n==52) stop 'Too many file names begin with xxx'
  end do  ! file xxxfile should be nonexistent now because it will be deleted
  open(newunit=newu,file=xxxfile)
  write(newu,"(A)") xxxfile       ! xxxfile now has content = its own name
  close(newu,status='delete')  
  inquire(file=xxxfile,exist=ex)
  print "(3A,L2)",'Was ',xxxfile,' deleted?', .not.ex
end program newfilemin
```
Output from lfortran 0.44.0: (answer should have been T)
```
Was xxx_a deleted? F
```
Output from  lfortran 0.43.0: (what happened to the file's name?)
```
Was  deleted? T
```
Output from lfortran 0.42.0 or gfortran: (correct: file was deleted but not its name)
```
Was xxx_a deleted? T

### PR DESCRIPTION
**Pull request by kmr-srbh**
resolves #6082

### CODE REVIEW

File: integration_tests/file_close_01.f90 [hunk 1]
```diff
@@ -0,0 +1,17 @@
+program file_close_01 ! Do files still exist after closing with delete?
```
gxyd, line 1: If that's a question you are inquiring about, then: if `DELETE` is specified, the file will not exist after the execution of a `CLOSE` statement.

integration_tests/file_close_01.f90
```diff
@@ -0,0 +1,17 @@
...
+   character(len=16) :: str
+   integer :: newu = 10
+   logical :: ex
+   file_name = "file_close_1.txt"
+   open(newunit=newu,file=file_name)
+   write(newu, *) file_name       ! file_name now has content = its own name
+   rewind(newu)
+   read(newu,  *) str
+   print *, "contents: ", str
+   close(newu,status='delete')
```
gxyd, line 13: Can we also have a test case where `status` is specified via a variable (i.e. a default-char-expr, as it's called by the Fortran standard)?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2788,6 +2788,7 @@ LFORTRAN_API int64_t _lpython_open(char *path, char *flags)
 
 struct UNIT_FILE {
     int32_t unit;
+    char* filename;
```
gxyd, line 2791: I'm usually a little skeptical about using `char*`, as we'll have to manually take care of freeing up the memory, I mostly prefer fixed size char, but the Fortran standard doesn't specify any maximum length of file-name. So, I'm not really sure.

gxyd:
Left a few comments to think about, but otherwise, this looks pretty good to me.

integration_tests/file_close_01.f90
kmr-srbh: Sure! I am pushing the test case.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: Ah, this is only a pointer actually. The actual string is the one created in the back-end. It maybe allocated on a stack and freed when the function returns, or be a constant and freed at program exit by the OS, or be a malloc'ed string (`allocatable character`) which we free through the `ImplicitDeallocate` node like `file_name` here. :)

Issue #5900, opened 2025-01-02
Handle function result allocatable arrays without `--realloc-lhs` better


certik:
Currently it seems `--realloc-lhs` is needed if a function returns an allocatable array.

Given that function results get transformed into `intent(out)` arguments, the `--realloc-lhs` should not be needed, since `intent(out), allocatable` doesn't require any runtime checks, the logic is that the input argument gets deallocated, then the function gets called.

There are likely some issues related to the new simplifier pass and temporaries, but it should be possible to implement this without any extra runtime checks, so it should work without `--realloc-lhs`.

certik:
Note that a minimal example like this already works correctly with LFortran:
```
program test_alloc

integer, allocatable :: x(:)

x = f(2)
print *, x
x = f(3)
print *, x

contains

    function f(n) result(r)
    integer, intent(in) :: n
    integer, allocatable :: r(:)
    allocate(r(n))
    r = 5
    end function

end program
```
But the fastGPT example doesn't. We need to create an MRE first.

certik:
The `--realloc-lhs` was introduced for fastGPT in ffff79cb87d402d28243efd99b9ec337fec6e752 as part of https://github.com/lfortran/lfortran/pull/5190. There is a useful comment https://github.com/lfortran/lfortran/pull/5190#discussion_r1819460627 that says `--realloc-lhs` is needed, but doesn't say why.

certik:
In fastGPT the bug happens here:
```fortran
function merge_utf8_pairs(intokens) result(tokens)
! Merge all UTF-8 character pairs
type(string), intent(in) :: intokens(:)
type(string), allocatable :: tokens(:)
integer :: i, j
logical :: one_more_pass
print *, "Y1"
allocate(tokens(size(intokens)))
print *, "Y2"
tokens = intokens
print *, "Y3"
```
The `Y2` gets printed, but `Y3` doesn't, it segfaults on the line `tokens = intokens`. However when trying to reproduce this with a simple MRE it works. So it is probably some corner case when this happens inside a function or something like that.

certik:
Found it. Here is a small reproducible example:
```fortran
program test_alloc
implicit none

type :: string
    character(:), allocatable :: s
end type


type(string), allocatable :: x(:)
type(string), allocatable :: y(:)
integer :: i

x = f(5)
allocate(y(size(x)))
x = g(x)
do i = 1, size(x)
    print *, i, x(i)%s, y(i)%s
end do

contains

[...]

PR #6090, opened 2025-01-23
Remove `--realloc-lhs` flag from fastGPT CI

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/5900


czgdp1807:
Seems like, the Github Actions CI is down. We will have to wait until it clears up.

czgdp1807:
This is a working solution. However we need to clearly work on the rules on what to allocate and what not to allocate in LFortran (with simplifier_pass). Then implement those rules and add clear fixes.

Issue #6051, opened 2025-01-18
Segfault on assigning function return to allocatable array with simplifier


HarshitaKalani:
```fortran
program test
  implicit none
  integer, allocatable :: array(:)
  array = arr()
  print *, array
  
  contains
  function arr()
    integer arr(1)
    arr(1) = 1
  end function
end program test
```
```console
$ lfortran b.f90
Segmentation fault (core dumped)
$ gfortran b.f90 && ./a.out
           1
$ lfortran b.f90 --no-experimental-simplifier
1
```

czgdp1807:
I am not sure whether to classify this as a bug because `gfortran` also fails if you disable reallocation of LHS in it (In `gfortran`, unlike `lfortran` reallocation of LHS is enabled by default, you can disable it by `-fno-realloc-lhs`). See below,

```zsh
(lf) 23:06:19:~/lfortran_project/lfortran % gfortran ../debug.f90 -fno-realloc-lhs
(lf) 23:06:29:~/lfortran_project/lfortran % ./a.out

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x102fc6d2f
#1  0x102fc5d83
#2  0x189a92e03
#3  0x102ddfe47
#4  0x102ddfecf
zsh: segmentation fault  ./a.out
(lf) 23:06:31:~/lfortran_project/lfortran % lfortran ../debug.f90                                         
(lf) 23:06:43:~/lfortran_project/lfortran % lfortran ../debug.f90 --realloc-lhs                           
1
```

In older LFortran, many things were auto reallocated. We can do the same with simplifier pass as well but I am not sure about how right it would be.

HarshitaKalani:
Oh I see. Thanks for the information. We can close this as this is not a bug.

PR #6093, opened 2025-01-23
Fix segfault on assigning function return to allocatable array with simplifier

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/6051


czgdp1807:
I agree. I would follow the rule if `gfortran -fno-realloc-lhs` is not working then LFortran by default should not work as well.

Issue #6079, opened 2025-01-22
bug: disallow `exit` statement outside loop or `select case` constructs


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none
   print *, "hello"
   exit
   print *, "world"
end program
```

### Output

```console
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
Traceback (most recent call last):
  Binary file "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2732, in main
    return main_app(argc, argv);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2670, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 1050, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/fortran_evaluator.cpp", line 382, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    compiler_options, run_fn, infile);
[...]

PR #6118, opened 2025-01-25
fix: exit outside any loop or block

sriganeshres:
I have added error handling to visit_Exit(const ASR::Exit_t &x) function so that when encountered a exit outside any loop or block will throw a error.

Refers #6079


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6220,6 +6220,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     }
 
     void visit_Exit(const ASR::Exit_t &x) {
+        if (loop_or_block_end.empty()) {
+            throw CodeGenError("Error: 'exit' statement used outside of any loop or block.",
+                               x.base.base.loc);
+        }
```
kmr-srbh, line 6226: Can we do this in the front-end?

tests/errors/exit_outside_loop_or_block.f90
```diff
@@ -0,0 +1,6 @@
+program main
+   implicit none
+   print *, "hello"
+   exit
+   print *, "world"
+end program
```
kmr-srbh, line 6: Please add a test case inside `tests/continue_compilation_1.f90`. I think just adding the `exit` statement in program scope would work?

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh, line 6226: Can we throw a semantic error in the front-end at ASR level?

sriganeshres: The thing is asr is generated as well as ast when ran with those flags --show-asr and --show-ast but running without those is throwing this error and it is from derefernecing null pointer in asr_to_llvm.cpp so i added the like there.

tests/errors/exit_outside_loop_or_block.f90
kmr-srbh: Yes. I meant to say that we should add this test case inside the above mentioned file. It is one of the files that we now use for all new error references, instead of separate ones.

sriganeshres: ok

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: @certik  @kmr-srbh  Can you give more information about what to be done.

sriganeshres: Ok.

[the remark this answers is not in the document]
sriganeshres: I thought it is not hurting any thing why to break so i left it like that.

tests/reference/pass_insert_deallocate-finalize_01-3efebdd.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: EXIT statement outside of loop
+  --> tests/finalize_01.f90:41:9
+   |
+41 |         exit bl ! This will have `impliciteDeallocate` inserted before it.
+   |         ^^^^^^^
```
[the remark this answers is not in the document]
sriganeshres: If i have ensured block safety then i guess this error will go away

tests/errors/continue_compilation_1.f90
```diff
@@ -178,6 +178,9 @@ subroutine my_func(x, y)
     print *, reshape(.true., [2, 3])
     print *, reshape([1, 2, 3, 4], "hello")
     print *, reshape([1, 2, 3, 4], .false.)
+   print *, "hello"
+   exit
+   print *, "world"
```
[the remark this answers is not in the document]
sriganeshres: ok

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -27,6 +27,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
     std::set<std::string> labels;
     size_t starting_n_body = 0;
     int loop_nesting = 0;
+    int all_loop = 0;
```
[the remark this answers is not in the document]
sriganeshres: Done

sriganeshres:
What to do next will you merge this PR.

tests/reference/pass_insert_deallocate-finalize_01-3efebdd.stderr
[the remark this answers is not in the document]
sriganeshres: Ok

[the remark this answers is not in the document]
sriganeshres: I think it is not caused by my changes instead by the block statement i guess

Issue #6086, opened 2025-01-23
bug: findloc giving wrong answers with `dimension` argument


HarshilShah1804:
```fortran
program main
    integer, dimension(2, 3) ::  arr = reshape([1, 2, 3, 4, 5, 6], [2, 3])
    integer :: x = 1
    print *, findloc(arr, x, dim=1)
end program
```

```
lfortran a.f90
1    1    1

gfortran a.f90 && ./a.out
           1           0           0
```

PR #6119, opened 2025-01-25
fix: handle 'dim' argument of 'findloc' in 'intrinsic_function' pass

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6086

This PR currently breaks `findloc` for 1-dim array's, that's why the integration test will currently fail, I'll fix that tomorrow.


gxyd:
Fixing the bug wasn't the highest of the priority here, but I did this PR because it gave me a better understanding on how the `intrinsic_function` pass is used to write Fortran code, and possibly the compilations involved with getting intrinsics to work with higher-dimensions input array.

PR #6155, opened 2025-01-30
CI : `lincoa 2` correct output

assem2002:


assem2002:
if it happens that we face failing CI cause of inconsistent segfault we can just revert that commit a16d3ec9c6e816b2524aab5f77a57c6f80d089d8

PR #6163, opened 2025-01-30
enh: Support assignment when lhs var is common to funcCall argument var

### LINKED ISSUE #6148: Handle `Arraysection` in funcCall on rhs if argument variable of funcCall is same as lhs
MRE:
```fortran 
program test 
    implicit none
    real(8) :: x(4) = [1.0_8,2.0_8,3.0_8,4.0_8]
    call temp(x)
contains

    implicit none
    real(8), intent(in) :: x(:)
    real(8), intent(in) :: y(:)
    real(8) :: z 
    integer :: i

    z = 0.0_8
    do i = 1, size(x)
        z = z + x(i) * y(i)
    end do
end function inprod
  
subroutine temp(x)
    implicit none
    real(8), intent(in) :: x(:)
    real(8) :: z2(size(x))
    z2 = x
    z2(1:2) = inprod(z2(1:2), z2(1:2))
    print *, z2
end subroutine temp
end program
```

GFortran Output :  ` 5.0000000000000000        5.0000000000000000        3.0000000000000000        4.0000000000000000 ` 

LFortran Output : 
```console
5.00000000000000000e+00
2.90000000000000000e+01
3.00000000000000000e+00
4.00000000000000000e+00
```

jinangshah21:
Fixes #6148 

Removes : https://github.com/Pranavchiku/prima/commit/3c7bd0c24886677af39057598e39f53e009f2295

Previously we were only checking common variable for lhs and rhs arrays, now we also check if funcCall present on rhs contains common array variable to lhs or not.


integration_tests/functions_30.f90
```diff
@@ -62,6 +67,15 @@ subroutine check_for_args(x, y)
       if(maxval([1, -y]) /= 3) error stop
       print *, maxval([1, y + y([2,1,3,4])])
       if(maxval([1, y + y([2,1,3,4])]) /= 8) error stop
+      
+      x2 = x
+      x2 = x2 / norm(x2)
```
Pranavchiku, line 72:
```suggestion
      x2 = x / norm(x2)
```

check if this works, then we can merge it.

Pranavchiku (approved):
left a minor comment, else looks good.

integration_tests/functions_30.f90
jinangshah21: Done.

HarshitaKalani (approved):

PR #6182, opened 2025-02-01
feat: Add compile-time constant check for array bounds

awsomecty:
- Enhance is_compile_time_constant to handle function calls
- Add support for intrinsic functions like abs, min, max, etc.
- Fix array bounds checking for function call expressions


HarshitaKalani (changes requested):
Hey @awsomecty , I believe you formatted the code for the whole file which makes it difficult to review.
Can you please just do the required changes for the issue you wish to solve and then push the PR?

Issue #6127, opened 2025-01-27
throw error for invalid value of shape arg in intrinsic reshape


HarshitaKalani:
```fortran
program test
    print *, reshape([1,2,3,4],[3,2])
end program
```
```console
$ gfortran a.f90 && ./a.out
a.f90:2:21:

    2 |     print *, reshape([1,2,3,4],[3,2])
      |                     1
Error: Without padding, there are not enough elements in the intrinsic RESHAPE source at (1) to match the shape
$ lfortran a.f90
1    2    3    1    2    3
```

PR #6193, opened 2025-02-02
fix: correct compile time error message for reshape

Jaysukh-409:
Fixes: #6127 

- Throw correct compile time message for the `reshape` intrinsic when provided shape argument having greater size(product of shape array element) then size of source array

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -352,3 +352,9 @@ semantic error: reshape accepts arrays for `shape` argument, found bool instead.
     |
 147 |     print *, reshape([1, 2, 3, 4], .false.)
     |                                    ^^^^^^^ 
+
+semantic error: reshape accepts `source` array with size greater than or equal to size of `shape` array
+   --> tests/errors/continue_compilation_1.f90:149:14
+    |
+149 |     print *, reshape([1, 2, 3, 4], [2, 3])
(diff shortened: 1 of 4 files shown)


tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -352,3 +352,9 @@ semantic error: reshape accepts arrays for `shape` argument, found bool instead.
     |
 147 |     print *, reshape([1, 2, 3, 4], .false.)
     |                                    ^^^^^^^ 
+
+semantic error: reshape accepts `source` array with size greater than or equal to size of `shape` array
+   --> tests/errors/continue_compilation_1.f90:149:14
+    |
+149 |     print *, reshape([1, 2, 3, 4], [2, 3])
+    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `source` array has size 4, but `shape` array has size 6
```
HarshitaKalani, line 360:
```suggestion
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `shape` specifies size of 6, which exceeds the `source` array size of 4
```

I think ``shape` array has size 6` is misleading. The above can be a better label.

HarshitaKalani (commented):
If you can please address the comment. Otherwise it is good.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
HarshitaKalani: The previous text says `shape` array's size is 6, whereas `shape` size is 2, it specifies the output size to be 6. So I thought this makes more sense.

PR #6194, opened 2025-02-02
Throw error for compile time division by zero

### LINKED ISSUE #4339: bug: catch division by 0 in a regular division expression

```fortran
PROGRAM main
   IMPLICIT NONE
   INTEGER :: i
   INTEGER :: j

   i = 1
   j = 0

   PRINT *, i/j

END PROGRAM main
```
#### LFortran
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran ./examples/example.f90
-1
```
It is strange that GFortran does not catch this error either and fails with a `Floating point exception (core dumped)`.
#### GFortran
```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ ./a.out

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x798077623c5d in ???
#1  0x798077622d85 in ???
#2  0x79807724531f in ???
        at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#3  0x5d640d5921d1 in ???
#4  0x5d640d59223c in ???
#5  0x79807722a1c9 in __libc_start_call_main
        at ../sysdeps/nptl/libc_start_call_main.h:58
#6  0x79807722a28a in __libc_start_main_impl
        at ../csu/libc-start.c:360
#7  0x5d640d5920b4 in ???
#8  0xffffffffffffffff in ???
Floating point exception (core dumped)
```

GFortran raises an error for compile-time values, but LFortran does not.

xaerru:
Fix #4339


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -12,6 +12,7 @@
 #include <libasr/pass/intrinsic_subroutine_registry.h>
 #include <lfortran/utils.h>
 #include <lfortran/semantics/comptime_eval.h>
+#include <lfortran/semantics/asr_implicit_cast_rules.h>
```
HarshitaKalani, line 15: Do we need to do this?

tests/errors/division_by_zero_02.f90
```diff
@@ -0,0 +1,9 @@
+program main
+implicit none
+
+integer, parameter :: zero = 0
+integer :: x = 1
+
+print *, x/zero
+
+end program
```
HarshitaKalani, line 9: Instead of adding these new test files, can you please add the tests in [tests/errors/continue_compilation_1.f90](https://github.com/lfortran/lfortran/pull/6014/files#diff-0e0e3b466065f1803f01e3370df0c3239313c0c7bf679eebdfeee8ef9691cfef) file?

And then run
```
rm -rf ./tests/reference/*
./run_tests.py -u
```

src/lfortran/semantics/ast_common_visitor.h
xaerru: Without it, my LSP (clangd) complained that `ImplicitCastRules` class is undeclared.

HarshitaKalani: >Without it, my LSP (clangd) complained that ImplicitCastRules class is undeclared.

Try removing it. If you face issue, can you paste the error here?

xaerru: image

It is not a compiler error but a LSP(auto complete) error. Everything compiles even without the include though.

Issue #6202, opened 2025-02-03
`shiftl` is not throwing error for invalid shift value


Jaysukh-409:
MRE:
```fortran
program main
    integer, parameter:: ik1 = selected_int_kind(1)
    integer(ik1), parameter:: val1 = 2, val2 = 9
    print *, shiftl(val1, val2) 
end program
```

Output:
```console
$ lfortran a.f90
0

$ gfortran a.f90 && ./a.out
a.f90:4:26:

    4 |     print *, shiftl(val1, val2)
      |                          1
Error: ‘SHIFT’ at (1) must be less than or equal to BIT_SIZE(‘I’)
```

Jaysukh-409:
The similar error is occurs in other shift operations also. Like, `shiftr`, `rshift`, `lshift`.

PR #6206, opened 2025-02-03
throw correct error message for `shiftl`, `shiftr` and `rshift`

Jaysukh-409:
Fixes: #6202 

- Throw correct error message for `shiftl`, `shiftr` and `rshift` intrinsic when provided incorrect `shift` value.


Jaysukh-409:
One thing I observe while resolving this was when we call `lshift` then also `shiftl` is called. Is there any specific reason for that ?
```console
$ lfortran a.f90
semantic error: The shift argument of 'shiftl' intrinsic must be non-negative integer
 --> a.f90:4:27
  |
4 |     print *, lshift(val1, -5)
  |     
```

Issue #5431, opened 2024-11-20
Throw error for type mismatch in initialisation


HarshitaKalani:
```fortran
program main
    implicit none
    real :: adwf = .true.
end program
```
```console
$ lfortran a.f90 
; ModuleID = 'LFortran'
source_filename = "LFortran"

@main.adwf = internal global float true

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  br label %return

return:                                           ; preds = %.entry
  ret i32 0
}

declare void @_lpython_call_initial_functions(i32, i8**)
asr_to_llvm: module failed verification. Error:
Global variable initializer type does not match global variable type!
float* @main.adwf

code generation error: asr_to_llvm: module failed verification. Error:
Global variable initializer type does not match global variable type!
float* @main.adwf



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$ gfortran a.f90 && ./a.out
a.f90:3:18:

    3 |     real :: adwf = .true.
      |                  1
Error: Cannot convert LOGICAL(4) to REAL(4) at (1)
```

Same needs to be done for other type mismatches as well.

HarshitaKalani:
```console
% lfortran
 b.f90
semantic error: Cannot convert LOGICAL to REAL
 --> b.f90:3:20
  |
3 |     real :: adwf = .true.
  |                    ^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
% gfortran b.f90 && ./a.out
b.f90:3:18:

    3 |     real :: adwf = .true.
      |                  1
Error: Cannot convert LOGICAL(4) to REAL(4) at (1)
```

PR #6208, opened 2025-02-04
Raise error for LOGICAL to REAL conversion

mohitdhingra22:
Fixed -#5431
- This PR modifies the error handling for invalid LOGICAL to REAL assignments.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3556,7 +3556,13 @@ public:
                         value = init_expr;
                     }
                     ASR::ttype_t *init_type = ASRUtils::expr_type(init_expr);
-
+                    // if (type=real and init_type=logical) Raise a semantic error
+                    diag.add(Diagnostic(
+                        "Cannot convert LOGICAL to REAL at assignment",
+                        Level::Error, Stage::Semantic, {
+                            Label("", {init_expr->base.loc})
+                        }));
+                    throw SemanticAbort();
                     size_t rhs_rank = ASRUtils::extract_n_dims_from_ttype(init_type);
                     size_t lhs_rank = ASRUtils::extract_n_dims_from_ttype(type);
 


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3556,7 +3556,13 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         value = init_expr;
                     }
                     ASR::ttype_t *init_type = ASRUtils::expr_type(init_expr);
-
+                    // if (type=real and init_type=logical) Raise a semantic error
```
gxyd, line 3559: You'll need to convert this pseudo-code condition into an `if` condition in C++, and only then raise the error.

gxyd (changes requested):
Also, you can always run integration tests locally on your machine to see if the tests pass or not by running:
```console
> cd integration_tests
> python run_tests.py -j=8
```

src/lfortran/semantics/ast_common_visitor.h
HarshitaKalani, line 3559: This needs to be removed.

tests/errors/continue_compilation_1.f90
```diff
@@ -182,6 +180,7 @@ subroutine my_func(x, y)
     print *, product(r1)
     print *, minval(q1)
     print *, maxval(r1)
-
+    
     print *, sum([1, 2, 3], mask = [1, 2, 3])
+    x_real = y
```
gxyd, line 185: Does testing it via `z1 = y` work?

mohitdhingra22:
@certik i did the above changes u mentioned :)

tests/errors/continue_compilation_1.f90
```diff
@@ -3,6 +3,7 @@ program continue_compilation_1
 
     integer :: a(3), b(3), b1(3, 3), a3(3, 3, 3), b4(3, 3, 3, 3), a5, c5, i, arr1(3), arr2(2, 3), arr3(2, 1, 3)
     character :: a1(3, 3)
+    logical :: y
```
gxyd, line 6: Can we move this declaration on the same line as:
```f90
logical :: a2(3, 3), mask1(3), mask2(2, 3), mask3(2, 1, 3), mask4(3, 2), mask5(2, 3, 1)
```
?

tests/errors/continue_compilation_1.f90
mohitdhingra22: ---
> 183 |     z1 = y 
>     |     ^^   ^ type mismatch (real and logical)

gxyd: Yes, but with your current change you introduce a new variable, which isn't really necessary in this case I believe.

mohitdhingra22: so need to stick with `z1`?

tests/errors/continue_compilation_1.f90
mohitdhingra22: yes, we can!

tests/errors/continue_compilation_1.f90
gxyd: Yes, I believe that should work.

gxyd: along with the fix for : https://github.com/lfortran/lfortran/pull/6208#discussion_r1948660070, this should be good.

HarshitaKalani (approved):
Looks good to me.

gxyd (approved):

gxyd:
Though this PR changes the line information of a few tests, but I went through the reference test changes and I think they are fine.

I think we can merge this now, thanks @mohitdhingra22

PR #6209, opened 2025-02-04
fix: consider only `intent(in)` scalar argument while creating dimension for temporary

### LINKED ISSUE #6205: Size allocation to temporary is incorrect when it depends on `intent(inout)` integer
MRE:
```fortran 
module function_32_mod

contains
function matprod(x, y) result(z)
    real(8), intent(in) :: x(:)
    real(8) :: z(size(y, 2))
    z = 0.0_8
end function matprod
end module

program function_32
    real(8) :: A(5,3)
    integer :: iact(3)
    integer :: nact
    iact = [2,1,3]
    A = 1.0_8
    nact = -1
    call func(A, iact, nact)
contains 
    subroutine func(A, iact, nact)
        use function_32_mod
        real(8), intent(in) :: A(:,:)
        integer, intent(inout) :: iact(:)
        integer, intent(inout) :: nact
        real(8) :: dnew(size(A,2))
        real(8) :: cviol
        nact = 1
        dnew(1:nact) = -matprod(dnew, A(:, 1:nact)) !! This segfaults
    end subroutine
end program
``` 

LFortran Segfaults.

Towards : https://github.com/Pranavchiku/prima/commit/00caff859ce08337105899aff29d8e84d0390130
Towards : https://github.com/Pranavchiku/prima/commit/6f6353e5e29f24796d03cf8991882d3618aed44e
Towards : https://github.com/Pranavchiku/prima/commit/027724063ed06a96a05e3cd77909c91d0f9ed3fd

jinangshah21:
Fixes #6205 

Removes : https://github.com/Pranavchiku/prima/commit/c79a877e98ffc6c9b9bd33d77ff5024e01d92716
Removes : https://github.com/Pranavchiku/prima/commit/99dbeb7fd7ee5e5635df08d3510bc41cf91437f9

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -2401,6 +2401,11 @@ static inline bool is_only_upper_bound_empty(ASR::dimension_t& dim) {
     return (dim.m_start != nullptr && dim.m_length == nullptr);
 }
 
+inline bool is_array(ASR::ttype_t *x) {
+    ASR::dimension_t* dims = nullptr;
+    return extract_dimensions_from_ttype(x, dims) > 0;
+}
+
 class ExprDependentOnlyOnArguments: public ASR::BaseWalkVisitor<ExprDependentOnlyOnArguments> {
 
     public:
@@ -2413,7 +2418,11 @@ class ExprDependentOnlyOnArguments: public ASR::BaseWalkVisitor<ExprDependentOnl
         void visit_Var(const ASR::Var_t& x) {
             if( ASR::is_a<ASR::Variable_t>(*x.m_v) ) {
                 ASR::Variable_t* x_m_v = ASR::down_cast<ASR::Variable_t>(x.m_v);
-                is_dependent_only_on_argument = is_dependent_only_on_argument && ASRUtils::is_arg_dummy(x_m_v->m_intent);
+                if ( ASRUtils::is_array(x_m_v->m_type) ) {
+                    is_dependent_only_on_argument = is_dependent_only_on_argument && ASRUtils::is_arg_dummy(x_m_v->m_intent);
+                } else {
+                    is_dependent_only_on_argument = is_dependent_only_on_argument && (x_m_v->m_intent == ASR::intentType::In);
+                }
             } else {
                 is_dependent_only_on_argument = false;
             }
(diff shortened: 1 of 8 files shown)


src/libasr/asr_utils.h
```diff
@@ -2413,7 +2418,11 @@ class ExprDependentOnlyOnArguments: public ASR::BaseWalkVisitor<ExprDependentOnl
         void visit_Var(const ASR::Var_t& x) {
             if( ASR::is_a<ASR::Variable_t>(*x.m_v) ) {
                 ASR::Variable_t* x_m_v = ASR::down_cast<ASR::Variable_t>(x.m_v);
-                is_dependent_only_on_argument = is_dependent_only_on_argument && ASRUtils::is_arg_dummy(x_m_v->m_intent);
+                if ( ASRUtils::is_array(x_m_v->m_type) ) {
+                    is_dependent_only_on_argument = is_dependent_only_on_argument && ASRUtils::is_arg_dummy(x_m_v->m_intent);
+                } else {
+                    is_dependent_only_on_argument = is_dependent_only_on_argument && (x_m_v->m_intent == ASR::intentType::In);
```
Pranavchiku, line 2424: why can't it be `inout`?

Pranavchiku (changes requested):
left a question

src/libasr/asr_utils.h
jinangshah21: Because `inout` can be changed during procedure now consider below example :
```fortran
integer, intent(inout) :: nact
 nact = 2
 bmat(:, 1:nact) = -matprod(yzmat_c, x(:, 1:nact))
 ```
 
 During simplifier pass it makes temporary for `matprod(yzmat_c, x(:, 1:nact))` where return size of matprod is `size(2nd argument, 2)`
 
 So it do something like :
 ```fortran
integer, intent(inout) :: nact
integer, dimension(nact) :: func_call_temp
 nact = 2
func_call_temp = -matprod(yzmat_c, x(:, 1:nact))
 bmat(:, 1:nact) = func_call_temp
 ```
 but instead we should make an allocatable temporary.
 For deciding between allocatable and fixed size, simplifier sees if all size arguments are dependent only on function arguments then it directly define dimension but for `inout` we can't directly assign dimension.

Pranavchiku: okay, I'll take it. Thanks for fixing it!

Pranavchiku (approved):

PR #6210, opened 2025-02-04
throw error when mask is not logical

### LINKED ISSUE #6201: throw error if mask to `sum` is not logical
```fortran
program main
    print *, sum([1, 2, 3], mask = [1, 2, 3])
end program
```
```console
$ gfortran a.f90 && ./a.out
a.f90:11:35:

   11 |     print *, sum([1, 2, 3], mask = [1, 2, 3])
Error: ‘mask’ argument of ‘sum’ intrinsic at (1) must be LOGICAL
$ lfortran a.f90
6
```

sauravsoni6377:
Fixes  #6201

- Throw error message  mask argument to sum is not logical.

in src/libasr/pass/intrinsic_array_function_registry.h:

`else if (args[2]) {
                if (is_logical(*ASRUtils::expr_type(args[2]))) {
                    mask = args[2];
                } else {
                    diag.add(diag::Diagnostic("'mask' argument of 'sum' intrinsic at (1) must be LOGICAL",
                        diag::Level::Error, diag::Stage::Semantic, {diag::Label("", {loc})}));
                    return nullptr;
                }
    }`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -473,6 +473,7 @@ jobs:
             openblas=0.3.21
             llvm-openmp=14.0.4
             make=4.3
+            openmpi=5.0.6
 
       - uses: hendrikmuhs/ccache-action@main
         with:
@@ -530,15 +531,22 @@ jobs:
       - name: Test POT3D
         shell: bash -e -l {0}
         run: |
-            git clone https://github.com/gxyd/pot3d.git
             FC="$(pwd)/src/bin/lfortran"
+            git clone https://github.com/gxyd/pot3d.git
             cd pot3d
-            git checkout -t origin/lf3
-            git checkout 81ff23264eb31860dbc0e156050e877f03696d2d
+            git checkout -t origin/build_pot3d_custom_mpi_wrappers_and_without_hdf5
+            git checkout 502271c6ae591893d8132ed08be40c3637e47893
             cd src
-            $FC -c mpi.f90
-            $FC -c psi_io.f90
-            $FC pot3d.F --cpp --implicit-interface --fixed-form --show-llvm
+            if [[ "${{matrix.os}}" == "macos-latest" ]]; then
+              CC=clang
+            else
+              CC=gcc
+            fi
+            ${CC} -I$CONDA_PREFIX/include -c mpi_wrapper.c
+            ${FC} -c mpi_c_bindings.f90
+            ${FC} -c mpi.f90
+            ${FC} -c psi_io.f90
+            ${FC} -c --cpp --implicit-interface pot3d.F90 --show-asr
 
       - name: Test Legacy Minpack (SciPy)
         shell: bash -e -x -l {0}


src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -691,8 +691,14 @@ static inline ASR::asr_t* create_ArrIntrinsic(
...
-    } else if (args[2] && is_logical(*ASRUtils::expr_type(args[2]))) {
-                mask = args[2];
+    } else if (args[2]) {
+                if (is_logical(*ASRUtils::expr_type(args[2]))) {
+                    mask = args[2];
+                } else {
+                    diag.add(diag::Diagnostic("'mask' argument of 'sum' intrinsic at (1) must be LOGICAL",
+                        diag::Level::Error, diag::Stage::Semantic, {diag::Label("", {loc})}));
+                    return nullptr;
+                }
```
kmr-srbh, line 701: The `else if` condition checked for `logical` already, so can we add an `else` block and throw this error? I think we are doing the same with the current changes, but with some nesting.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -691,8 +691,14 @@ static inline ASR::asr_t* create_ArrIntrinsic(
                 args[1]->base.loc);
             return nullptr;
         }
-    } else if (args[2] && is_logical(*ASRUtils::expr_type(args[2]))) {
-                mask = args[2];
+    } else if (args[2]) {
+                if (is_logical(*ASRUtils::expr_type(args[2]))) {
+                    mask = args[2];
+                } else {
+                    diag.add(diag::Diagnostic("'mask' argument of 'sum' intrinsic at (1) must be LOGICAL",
```
kmr-srbh, line 698: Please use all lower-case. Also, do not use the text `"(1)"`. The `Location loc` informs the user of the place of occurrence.

tests/reference/run-allocate_01-d7a2337.stderr
```diff
@@ -0,0 +1,4 @@
+sh: 1: clang: not found
+The command 'clang -o allocate_01.out tests/errors/allocate_01.tmp.o  -L"$DIR/src/bin/../runtime" -Wl,-rpath,"$DIR/src/bin/../runtime" -llfortran_runtime -lm' failed.
+Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
+Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
```
kmr-srbh, line 4: This reference is not required actually. By default, we need `clang` to link the executable. I think we currently do not support using the `--link-with-gcc` when running the reference tests. Please install `clang` and do the following:

1. Pull latest changes from `main`.
2. Do `git checkout main tests/reference/` to reset the test references.
3. Run `./run_tests.py -u` to generate references.
4. Check the references for anomalies like error stacktraces or the one here. Ideally, we should have references for only the semantic error we add here.
5. Push the new references.

tests/errors/mask_not_logical.f90
```diff
@@ -0,0 +1,3 @@
+program main
+    print *, sum([1, 2, 3], mask = [1, 2, 3])
+end program
```
kmr-srbh, line 3: Please add this test case inside `continue_compilation_01.f90`.

kmr-srbh (changes requested):

kmr-srbh:
Please click on "Ready for review" when the PR is ready.

src/libasr/pass/intrinsic_array_function_registry.h
kmr-srbh: Isn't this nesting doing the same thing as a non-nested `else if` and `else` block?

tests/reference/asr-mask_not_logical-1c7b2a8.json
```diff
@@ -0,0 +1,13 @@
...
+    "infile": "tests/errors/mask_not_logical.f90",
+    "infile_hash": "7c67278ecbb5a744cc477e7bc962e73be7353eb0713848c4091a1ae7",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "asr-mask_not_logical-1c7b2a8.stderr",
+    "stderr_hash": "eb2ed7b520d66648e33cae7a992ae6f7659a7254dbdc31db6b8c1030",
+    "returncode": 2
+}
```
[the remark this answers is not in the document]
sauravsoni6377: okay, i'll do it right now.

PR #6218, opened 2025-02-04
enh: support `allocatable` variables in list directed read

kmr-srbh:


kmr-srbh:
The implementation for reading boolean values is just a placeholder currently. It does not give correct output. I am trying to understand how GFortran does this. For any substring of `".false."` in same order from `"f"`, the string is read as `.false.` and vice versa for `".true."`. Is there a standard governing this behavior?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3589,6 +3589,11 @@ LFORTRAN_API void _lfortran_string_read_str(char *str, char *format, char **s) {
     sscanf(str, format, *s);
 }
 
+LFORTRAN_API void _lfortran_string_read_bool(char *str, char *format, int32_t *i) {
+    sscanf(str, format, i);
+    printf("%s\n", str);
```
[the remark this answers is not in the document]
kmr-srbh: Thanks for catching this. I was debugging the behavior for boolean read and added this print statement. I will remove it in the implementation for reading `logical` types.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8077,18 +8079,26 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                 llvm::Function::ExternalLinkage, runtime_func_name, *module);
                     }
                     llvm::Value *fmt = nullptr;
-                    if (ASR::is_a<ASR::Integer_t>(*type)) {
-                        ASR::Integer_t* int_type = ASR::down_cast<ASR::Integer_t>(type);
+                    if (ASR::is_a<ASR::Integer_t>(*ASRUtils::type_get_past_array(
```
gxyd, line 8082: I believe `ASRUtils::extract_type` does the same work as `ASRUtils::type_get_past_array(ASRUtils::type_get_past_allocatable_pointer`?

gxyd (commented):
Do you know from which file is this MRE extracted from? Is it pot3d.cpp itself?

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: Oh! I did not know about this. I recall there used to be a utility function by the name `type_get_past_array_pointer_allocatable()`, but I do not know where it disappeared. I think it was renamed to this function. I will make the changes in a different PR.

kmr-srbh:
> Do you know from which file is this MRE extracted from? Is it pot3d.cpp itself?

Yes. There are some more input related issues uncovered from the same file.

src/libasr/codegen/asr_to_llvm.cpp
gxyd: Sorry about that, they had the same purpose, and that's why I remove `type_get_past_array_pointer_allocatable` to avoid any confusion that might arise from it. Since, `extract_type` had been there already before `type_get_past_array_pointer_allocatable`, that's why I removed it.

gxyd:
> Yes. There are some more input related issues uncovered from the same file.

I'm curious, because there is some code in pot3d.cpp, which I wrote to remove dependency on HDF5, is from the code that I wrote? (if you can check by `git annotate`)

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: Oh, there were duplicates. :)

kmr-srbh:
> I'm curious, because there is some code in pot3d.F90, which I wrote to remove dependency on HDF5, is it from the code that I wrote? (if you can check by git annotate)

I am sorry I got confused here I see. The file is `pot3d_cpp.f90`. :smile:

gxyd:
So, it's simply that the file `pot3d_cpp.f90` is created from `pot3d.F90` by C-preprocessing, hence the suffix `_cpp` (i.e. using `--cpp` flag in LFortran and -cpp in GFortran, by replacing a few `#ifdef` with appropriate values), so it's mostly the same files.

Simply put `pot3d_cpp.f90` is generated when building, while `pot3d.F90` is the concerned file. (you can see: https://github.com/predsci/POT3D/tree/main/src, there is no file `pot3d_cpp.f90` present by default)

kmr-srbh:
>  the file pot3d_cpp.f90 is created from pot3d.F90 by C-preprocessing, hence the suffix _cpp

I see. Going through the commit history, I see that the issue is coming from the workaround done for namelists.

gxyd:
> I see. Going through the commit history, I see that the issue is coming from the workaround done for namelists.

It works with GFortran, so there's that.

gxyd:
>  I see that the issue is coming from the workaround done for namelists

Or maybe have a better way to read the `.bin` file save the variables.

Issue #6225, opened 2025-02-05
Codegen error on read operation in a subroutine


gxyd:
## Minimal Reproducible Program

For the below Fortran program:
```f90
module g
    contains
subroutine f
    integer :: ierr
    character(80) :: line
    character(20) :: key
    character(60) :: value
    integer :: pos_eq
    integer, save, dimension(3) :: nprocs=(/-1,-1,-1/)
    integer, save :: nr=0
    ierr = 0
    line = '  nr=133'
    pos_eq = 5
    key = "nr"
    value = "133"
    select case (trim(key))
        case ("nprocs")
            read(value, *) nprocs
        case ("nr")
            read(value, *) nr
        case default
            write(*, *) "WARNING: Unknown key:", trim(key)
    end select
    print *, key
end subroutine f
end module

program main
    use g
    implicit none
    call f
end program main

```

I get the below error on compilation with LFortran:
```console
> lfortran --version
LFortran version: 0.45.0-181-g7d41fc7b2-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90 --show-llvm

...
...

code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
i32* @f.nr
 [3 x i32]*  call void @_lfortran_string_read_i32(i8* %17, i8* getelementptr inbounds ([3 x i8], [3 x i8]* @6, i32 0, i32 0), i32* @f.nr)



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

while with GFortran:
```console
> gfortran test.f90 -o a.out && ./a.out
 nr
```

gxyd:
It might be possible to create a workaround for this also, if possible we'll probably do that. I'll check tomorrow for the exact cause of failure here.

PR #6228, opened 2025-02-06
fix: separate runtime functions to read string into array values

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6225


gxyd:
I'll need to fix the error, shouldn't difficult.

gxyd:
I've re-assigned a new string (with integer value inside it) to `str` variable.

I'll merge the PR once the CI succeeds.

PR #6235, opened 2025-02-06
Handle IntrinsicArrayFunction in m_length of Array in `function_call_in_declaration` pass

AnshMehta1:

PR #6245, opened 2025-02-07
enh : Add method to find global scope of any scope in SymbolTable

AnshMehta1:
Based on comment https://github.com/lfortran/lfortran/pull/6235#discussion_r1945303607

### DIFF
--- a/src/libasr/asr_scopes.h
+++ b/src/libasr/asr_scopes.h
@@ -50,6 +50,14 @@ struct SymbolTable {
         return scope[name];
     }
 
+    SymbolTable* get_global_scope(SymbolTable* current_scope) {
+        SymbolTable* global_scope = current_scope;
+        while (global_scope->parent) {
+            global_scope = global_scope->parent;
+        }
+        return global_scope;
+    }
+
     const std::map<std::string, ASR::symbol_t*>& get_scope() const {
         return scope;
     }
--- a/src/libasr/pass/function_call_in_declaration.cpp
+++ b/src/libasr/pass/function_call_in_declaration.cpp
@@ -148,10 +148,7 @@ public:
         std::vector<ArgInfo> indicies;
         get_arg_indices_used(x, indicies);
 
-        SymbolTable* global_scope = current_scope;
-        while (global_scope->parent) {
-            global_scope = global_scope->parent;
-        }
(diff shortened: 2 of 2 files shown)


gxyd (approved):
This looks good, I've noticed this block of code being used quite often in our codebase. This utility function will help.

Thanks.

src/libasr/asr_scopes.h
```diff
@@ -50,6 +50,14 @@ struct SymbolTable {
         return scope[name];
     }
 
+    SymbolTable* get_global_scope(SymbolTable* current_scope) {
```
Pranavchiku, line 53: do not pass `current_scope`, we already have it as we can call it like `current_scope->get_global_scope`

src/libasr/pass/function_call_in_declaration.cpp
```diff
@@ -148,10 +148,7 @@ class ReplaceFunctionCall : public ASR::BaseExprReplacer<ReplaceFunctionCall>
         std::vector<ArgInfo> indicies;
         get_arg_indices_used(x, indicies);
 
-        SymbolTable* global_scope = current_scope;
-        while (global_scope->parent) {
-            global_scope = global_scope->parent;
-        }
+        SymbolTable* global_scope = current_scope->get_global_scope(current_scope);
```
Pranavchiku, line 151: please address the changes

AnshMehta1: Surely , I will address this.

AnshMehta1: In struct SymbolTable , we don't have any pointer to access current_scope (or SymbolTable ) . Thus , we need to pass current_scope as argument . I don't see this as a good approach . Surely , something can be done to handle this .

Pranavchiku: The moment you define `SymbolTable* current_scope` you can do `current_scope->get_global_scope()` and now you do the following:

```cpp
    SymbolTable* get_global_scope() {
        SymbolTable* global_scope = this;
        while (global_scope->parent) {
            global_scope = global_scope->parent;
        }
        return global_scope;
    }
```

AnshMehta1: Ahh , got it . `this` just went of my mind.

Issue #6251, opened 2025-02-07
Register keywords for all intrinsics


HarshitaKalani:
```fortran
program main
  print *, abs(a = 4)
end program
```
```console
$ lfortran a.f90
semantic error: No matching signature found for intrinsic abs
 --> a.f90:2:12
  |
2 |   print *, abs(a = 4)
  |            ^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$ gfortran a.f90 && ./a.out
           4
```
This issue would be there with all the intrinsics with number of arguments = 1.
The reason being they are not registered here:
(line 1006 in ast_common_visitor.h file)
```
std::map<std::string, IntrinsicSignature> name2signature = {
        {"any", IntrinsicSignature({"mask", "dim"}, 1, 2)},
        {"all", IntrinsicSignature({"mask", "dim"}, 1, 2)},
...
```
The task is to register them here with the correct keyword corresponding to the intrinsic name. eg, "a" for abs (refer: https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gfortran/ABS.html).

nilay-v3rma:
Hey, I would like to work on this issue. I have already made the required changes for the `abs` function. To extend this fix to other intrinsic functions that take exactly one argument, Is there an exhaustive list of such functions available? If not, I can raise a PR with the signatures for the functions I could find, and someone can review it. Let me know how to proceed. Thanks!

HarshitaKalani:
You can check the list at: intrinsic_function_registry.h, intrinsic_array_function_registry.h, intrinsic_subroutine_registry.h file. You'll find all the intrinsics registered in these 3 files.

HarshitaKalani:
Here are some: 
abs(x), acos(x), asin(x), atan(x), cos(x), cosh(x), erf(x), erfc(x), exp(x), gamma(x), log(x), log10(x), sin(x), sinh(x), sqrt(x), tan(x), tanh(x), iabs(x), not(x), adjustl(x), adjustr(x), len(x), trim(x), kind(x), precision(x), range(x), tiny(x), huge(x), epsilon(x), digits(x), radix(x), present(x), associated(x), allocated(x), shape(x), size(x), leadz(i), trailz(i)

Note: add them one by one and test each locally.

harperjf:
[...]

PR #6253, opened 2025-02-07
Register keywords for 10 intrinsics

galactus3050:
towards #6251


HarshitaKalani (commented):
I think this is good. However, there are more such functions. Some of which I can think of are trailz, leadz etc. 
Also, can you please add a test to assure it is working fine. Thanks

galactus3050:
can anyone please guide me why it is failing on macos only?

Issue #6157, opened 2025-01-30
intrinsic reshape not accepting `pad` and `order` argument


Jaysukh-409:
MRE:
```fortran
program test
    print *, reshape([1,2,3,4],[2,2],order=[2, 1])
end program
```

Output:
```console
$ gfortran a.f90 && ./a.out
           1           3           2           4

$ lfortran a.f90
semantic error: reshape accepts only 2 arguments, got 2 arguments instead.
 --> a.f90:2:14
  |
2 |     print *, reshape([1,2,3,4],[2,2],order=[2, 1])
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

Beliavsky:
The error message

```
semantic error: reshape accepts only 2 arguments, got 2 arguments instead.
 --> a.f90:2:14
  |
2 |     print *, reshape([1,2,3,4],[2,2],order=[2, 1])
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```
should say `got 3 arguments`

PR #6254, opened 2025-02-07
fix: taking order and pad arguments in the reshape intrinsic

HarshilShah1804:
Towards #6157 

Current Status:
- [x] taking the arguments order and pad in the reshape function
- [x] add the validation for order to be the same size as shape and pad to be of same type as array
- [x] stop throwing error when number of elements in array is less than that after reshape if pad argument is given
- [x] implementing the logic for order and pad

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -6299,7 +6299,7 @@ public:
 
     ASR::asr_t* create_ArrayReshape(const AST::FuncCallOrArray_t& x) {
         if( x.n_args + x.n_keywords < 2 || x.n_args + x.n_keywords > 4 ) {
-            diag.add(Diagnostic("reshape expects number of arguments to be between, got " +
+            diag.add(Diagnostic("reshape expects number at least 2 and at most 4 arguments, got " +
                                 std::to_string(x.n_args) + " arguments instead.",
                                 Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
             throw SemanticAbort();
@@ -6319,7 +6319,7 @@ public:
         }
         if (x.n_args > 3){
             order = x.m_args[2].m_end;
-            pad = x.m_args[2].m_end;
+            pad = x.m_args[3].m_end;
         }
         for( size_t i=0;i<x.n_keywords;i++ ) {
             if( to_lower(x.m_keywords[i].m_arg) == "source" ) {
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6298,37 +6298,38 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
     }
 
     ASR::asr_t* create_ArrayReshape(const AST::FuncCallOrArray_t& x) {
-        if( x.n_args + x.n_keywords != 2) {
-            diag.add(Diagnostic("reshape accepts only 2 arguments, got " +
+        if( x.n_args + x.n_keywords < 2 || x.n_args + x.n_keywords > 4 ) {
+            diag.add(Diagnostic("reshape expects number of arguments to be between, got " +
                                 std::to_string(x.n_args) + " arguments instead.",
```
HarshitaKalani, line 6303:
```suggestion
            diag.add(Diagnostic("reshape expects atleast 2 and atmost 4 arguments, got " +
                                std::to_string(x.n_args + x.n_keywords) + " arguments instead.",
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6298,37 +6298,38 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-            if ( to_lower(x.m_keywords[1].m_arg) == "source" ) {
-                source = x.m_keywords[1].m_value;
-            } else if ( to_lower(x.m_keywords[1].m_arg) == "shape" ) {
-                shape = x.m_keywords[1].m_value;
+        if (x.n_args > 2){
+            pad = x.m_args[2].m_end;
+        }
+        if (x.n_args > 3){
+            order = x.m_args[2].m_end;
+            pad = x.m_args[2].m_end;
```
HarshitaKalani, line 6322: Why args[2] for both?

HarshilShah1804: Oh, it should be 3 for the `pad`. I'll correct this

Pranavchiku (commented):
I'll wait for you to address the comments above. Else if you have any other specific thing you wish me to review, please let me know

HarshilShah1804:
@HarshitaKalani  Okay. The implementation of the `order` and `pad` functionality is left. I will add the tests once the full implementation is completed as the added commits are mainly for validation checks.

HarshilShah1804:
The order and pad are now working. And the package numerical-methods-fortran is getting compiled with lfortran with order argument also. I will add the tests in some time.

Pranavchiku (approved):
This approach will only work for compile time values it is fine for now, please open issue for runtime values.

HarshitaKalani (approved):
Thanks @HarshilShah1804 ! Great work.

PR #6257, opened 2025-02-08
update name2signature for intrinsics with 1 arguement

nilay-v3rma:
towards issue [#6251](url) . Are any other changes required?


nilay-v3rma:
I initially had the test file intrinsics_1arg in the tests/errors folder, but since I couldn't find a CMake file there, I moved the test file to integration_tests. is it fine?

Additionally, for string functions, I am using the variable name str, and for arrays, I have used x, as shown in the following documentation. Please let me know if any changes are needed.


array function  doc- https://gcc.gnu.org/onlinedocs/gfortran/SHAPE.html
string function doc - https://gcc.gnu.org/onlinedocs/gcc-4.4.2/gfortran/TRIM.html#:~:text=7.219%20TRIM%20%E2%80%94%20Remove%20trailing%20blank%20characters%20of%20a%20string&text=Removes%20trailing%20blank%20characters%20of%20a%20string.&text=Shall%20be%20a%20scalar%20of,the%20number%20of%20trailing%20blanks.

nilay-v3rma:
working on it

nilay-v3rma:
I went through the details of the failing tests and found a common error in all of them,
image

So I tried the same thing in the repl mode and noticed that `string` (even though in the name2signature definition i have use `str`) works for len but not for other string function. And `str` works for other string function as defined in name2signatures (shown in the image attached below). I would appreciate if you have any advice for this problem. 

image

integration_tests/intrinsics_1arg.f90
```diff
@@ -0,0 +1,129 @@
...
+    ! Integer functions
+    i1 = iabs(a = -5)
+    print *, i1
+    if (i1 /= 5) error stop
+
+    i1 = not(i = 0)
+    print *, i1
+
+    ! String functions
+    str = adjustl(str = "  text")
```
HarshitaKalani, line 77: adjustl takes "string" as keyword and not "str": https://gcc.gnu.org/onlinedocs/gfortran/ADJUSTL.html.
Same for adjustr.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1108,6 +1108,44 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+        {"sin", IntrinsicSignature({"x"}, 1, 1)},
+        {"sinh", IntrinsicSignature({"x"}, 1, 1)},
+        {"sqrt", IntrinsicSignature({"x"}, 1, 1)},
+        {"tan", IntrinsicSignature({"x"}, 1, 1)},
+        {"tanh", IntrinsicSignature({"x"}, 1, 1)},
+        {"iabs", IntrinsicSignature({"a"}, 1, 1)},
+        {"not", IntrinsicSignature({"i"}, 1, 1)},
+        {"adjustl", IntrinsicSignature({"str"}, 1, 1)},
+        {"adjustr", IntrinsicSignature({"str"}, 1, 1)},
+        {"len", IntrinsicSignature({"str"}, 1, 1)},
```
HarshitaKalani, line 1132: len takes 1 or 2 arguments. We want to register only the ones which accept exactly one argument.

nilay-v3rma:
The tests are failing due to the following error,

image

name2signature for `allocated` function,
```
{"allocated", IntrinsicSignature({"x"}, 1, 1)}
```

Test for `allocated` function,
```
    allocate(arr(5))
    print *, allocated(x = arr)
    deallocate(arr)
```
  
  `allocated` function doc - https://gcc.gnu.org/onlinedocs/gfortran/ALLOCATED.html

Could you help me understand the reason for the failure?
Since x = arr, I am only passing one argument, which matches the expected signature.
Your advice would be appreciated.

HarshitaKalani:
https://gcc.gnu.org/onlinedocs/gfortran/ALLOCATED.html
From here, the keyword for it is "array" or "scalar", right? And you're passing "x"

nilay-v3rma:
@Pranavchiku I did not understand the changes you made. Could you please guide if there is anything that I need to do?

PR #6279, opened 2025-02-10
enh: add GenericCompare

### LINKED ISSUE #4532: ICE with templates
```fortran
module std_prop_m
    implicit none
    public

    requirement magma_r(T, bin)
        type, deferred :: T
        pure elemental function bin(x, y) result(bin)
            type(T), intent(in) :: x
            type(T), intent(in) :: y
            type(T) :: bin
        end function
    end requirement

    template commutative_prop(T,bin)
        require :: magma_r(T,bin)
      contains
        pure function commutative_p(x, y) result(prop)
            type(T), intent(in) :: x, y
            type(logical) :: prop

            prop = bin(x,y) == bin(y,x)
        end function
    end template

    template alt_commutative_prop(bin)
        require :: magma_r(integer,bin)
      contains
        pure function commutative_p(x, y) result(prop)
            type(integer), intent(in) :: x, y
            type(logical) :: prop

            prop = bin(x,y) == bin(y,x)
        end function
    end template
    
end module std_prop_m


program test_intrinsic_p
  use std_prop_m
  instantiate alt_commutative_prop(operator(+)), only: plus_comm => commutative_p
  instantiate commutative_prop(integer,operator(+)), only: int_plus_comm => commutative_p
  print *, "test commutative"
  print *, "plus_comm: ", plus_comm(3,4)
  ! print *, "int_plus_comm: ", int_plus_comm(3,4)
end program
```

This gives:
```console
$ lfortran a.f90
```

xaerru:
Fix #4532 

Add `GenericCompare` to compare two `TypeParameter` types.

When instantiated it is duplicated or replaced to the respective Compare expr.


xaerru:
Added error message for comparing two generic types without overloading.

> We associate the operator(+) with add. In ASR, we simply use add directly (I think). We do not have GenericBinOp in ASR, we just use add.

> Note: I can see this is a little problematic if we want to print the Fortran code back as Fortran, I guess it would print it as add(res, A(i)) instead of res + A(i). So I think we should figure out how to preserve this information in ASR somehow.

There was a bug in `visit_BinOp2` which caused an early return. That skipped the creation of `OverloadedBinOp`. I removed the early return and also added `template_simple_04.f90` to `tests.toml`.

I removed `GenericCompare` because it isn't needed. `OverloadedBinOp` provides information about the operator which was overloaded.

PR #6282, opened 2025-02-10
fix: incorrect c api call to random seed

Pranavchiku:
With this and https://github.com/Pranavchiku/prima/commits/test-bobyqa-2/ we get test_bobyqa working.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1032,10 +1032,10 @@ RUN(NAME intrinsics_351 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_352 LABELS gfortran llvm) # random_seed
 RUN(NAME intrinsics_353 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # random_number
 RUN(NAME intrinsics_354 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # minval
-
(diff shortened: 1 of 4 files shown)


HarshitaKalani (approved):

AnshMehta1:
> The error seems real:
> 
> ```
> semantic error: No implicit return type available for `x`
>   --> /home/runner/work/lfortran/lfortran/integration_tests/intrinsics_358.f90:9:1 - 13:18
>    |
>  9 |    function rand0() result(x)
>    |    ^^^^^^^^^^^^^^^^^^^^^^^^^^...
> ...
>    |
> 13 |    end function rand0
>    | ...^^^^^^^^^^^^^^^^^^ 
> 
> 
> Note: Please report unclear, confusing or incorrect messages as bugs at
> ```

Actually , we need to add impure to function declaration . With that it will work.

AnshMehta1:
> We should change the error message, I don't understand what it means.

I agree , it should be something that indicates error in function declaration .

AnshMehta1:
> @AnshMehta1 go ahead an open up an issue for the pure/impure thing. Why is impure needed?

Actually , as per `std=f23` , we can mention in declaration that whether our function is pure ( doesn't cause any side effects ) or impure ( can have side effects ) . But as per my knowledge , I am not sure that is it compulsory to mention pure/impure . Because gfortran doesn't raise any error as the one raised here https://github.com/lfortran/lfortran/pull/6282#issuecomment-2648971678

PR #6286, opened 2025-02-11
Fix : Handle more complex expressions for FMA optimizaiton

assem2002:
The fix is towards compiling PRIMA with `--fast`
This fix is considered as a workaround as the pass should be using the callExpressionReplacer class that's is used by most of the other passes to replace expression effectively .


assem2002:
if changes in the refernce file is overwhelming we can move the test in a standalone file instead of update an existing one.

assem2002:
any other complex expressions would fail. for example 
```.f90
program fma_02

    implicit none
    integer ::b(10)
    integer ::bb(10)
    real :: r
    bb(1) =  100
    b = 1
    r = 1
    b(1) = bb(r + r*r)
    print *, b(1)   

end program 
```

assem2002:
the `fma` instruction doesn't seem to be a bad optimization. I think just using the `callReplacerExpression` class in that pass would make things robust and clean, so I'll do that in subsequent PR with some integration tests to ensure it works as intended . If you think the whole pass isn't necessary, so maybe we can only use it with flag then.

PR #6285, opened 2025-02-11
Fix : update type of `UnaryMinus` parent from array to type of the array

assem2002:
Towards compiling PRIMA with `--fast`. With #6284 and this PR, We compile `cobyla_1`, `cobyla_2`, `lincoa_2`
for this expression `b = -b + 1.0` :
ASR Before : 
```clj
                            (RealBinOp
                                (RealUnaryMinus
                                    (ArrayItem
                                        (Var 2 b)
                                        [(()
                                        (Var 2 __libasr_index_0_2)
                                        ())]
                                        (Real 4)
                                        ColMajor
                                        ()
                                    )
                                    (Array
                                        (Real 4)
                                        [((IntegerConstant 1 (Integer 4) Decimal)
                                        (IntegerConstant 10 (Integer 4) Decimal))]
                                        FixedSizeArray
                                    )
                                    ()
                                )
                                Add
                                (RealConstant
                                    1.000000
                                    (Real 4)
                                )
                                (Array ; <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<This
                                    (Real 4)
                                    [((IntegerConstant 1 (Integer 4) Decimal)
                                    (IntegerConstant 10 (Integer 4) Decimal))]
                                    FixedSizeArray
                                )
                                ()
                            )
                            ()
                        )
```
ASR After : 
```clj
                            (RealBinOp
                                (RealUnaryMinus
                                    (ArrayItem
                                        (Var 2 b)
                                        [(()
                                        (Var 2 __libasr_index_0_2)
                                        ())]
                                        (Real 4)
                                        ColMajor
                                        ()
                                    )
                                    (Real 4)
                                    ()
                                )
                                Add
                                (RealConstant
                                    1.000000
                                    (Real 4)
                                )
                                (Real 4) ; <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<This
                                ()
                            )
                            ()
                        )
```

PR #6287, opened 2025-02-11
CI : Run PRIMA with `--fast`

assem2002:
after merging these #6285 ,#6285. we'll be able to compile and all tests, but 3 tests are failing (they're commented out in `.yml` file)


assem2002:
Expected to fail.

assem2002:
Update we skip pass `fma` for now to all examples working. I'll work on it.

PR #6289, opened 2025-02-11
ci: setup prima tests on CI

Pranavchiku:
Branch: https://github.com/Pranavchiku/prima/commits/lf-prima-2

Please note this has `lf-cleaned_branch-10` as base and `lf-cleaned_branch-11` throws error with `cobyla` test.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -826,44 +826,44 @@ jobs:
             ./project3
 
 
-Commented out to save CI runners:
- test-pixi:
-   name: Pixi Build (${{ matrix.os }})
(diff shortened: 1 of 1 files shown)


Pranavchiku:
Yes, I have `test_cobyla` commented at commit we test

Pranavchiku:
We also observed with @jinangshah21, @kmr-srbh and others who tried to run test on linux that it halts. I think same is happening with CI. I am disabling for linux, we can test it on mac.

Pranavchiku:
> run tests that do not use GB of memory.

How do we check that?

PR #6295, opened 2025-02-12
Initial implementation of language server

dylon:
Major changes:
* Adds initial implementation of language server
* Adds support for LSP protocol but has structure to support others
* Generates LSP data model, data transforms, and language server stubs from Microsoft spec (`src/server/metaModel.json`)
    * See: [Language Server Protocol Specification - 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/)
* Has multithreaded/asynchronous architecture
* Supports boilerplate logic for maintaining an in-memory corpus of document changes and configuration caching
* Adds support for validating documents to the lfortran language server

Communication works as follows:
* Language Client (VSCode) writes to the `stdin` stream of the language server and reads from its `stdout`.
* Requests and notifications may be sent from either the client-to-server or server-to-client.
* Requests and notifications must be responded to in mostly the same order in which they were received. Some requests and notifications may be responded-to out-of-order but llanguage-server responds to them in-order.
* Receiving a response from the client does not require the server to provide additional acknowledgement.
* Messages are passed in a format much like [HTML messages](https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages)
    * No start line is required (unlike HTML)
    * The only supported headers are `Content-Type` and `Content-Length`
    * `Content-Type` is mostly ignored since the LSP protocol is a superset of JSON-RPC. The `Content-Type` is assumed to be JSON and non-JSON requests will result in an error.
    * `Content-Length` describes how many bytes (chars) the JSON message contains.
    * Newlines are represented as `\r\n` regardless the operating system.
    * A single, empty line serves as the delimiter between the headers and message body.
    * Exactly `Content-Length` characters are read from `stdin` for the message body.
    * The message body is not followed by a newline.

Requests and notifications from client-to-server are fulfilled as follows:
1. The `CommunicationProtocol` streams messages from the client off `stdin` with an implementation of `MessageStream`.
2. Each message is enqueued into a `MessageQueue` for processing by the `LanguageServer`.
3. The language server listens for messages on the same `MessageQueue` and fulfills them from one of the available request threads.
4. The appropriate handler is dispatched and its response is enqueued into another `MessageQueue` for messages from server-to-client. The response will reference any request id from the client.
5. The `CommunicationProtocol` listens for messages to send to the client and writes them to `stdout`.

Requests and notifications from server-to-client are fulfilled as follows:
1. The `LanguageServer` enqueues a request or notification into the `MessageQueue` to the client. The request id will be recorded so any response from the client can be dispatched accordingly.
2. The `CommunicationProtocol` listens for messages to send to the client and writes them to `stdout`.
3. The client performs the requested action and responds on `stdin`.
4. The `CommunicationProtocol` listens for messages on `stdin` and enqueues them into the `MessageQueue` going to the `LanguageServer`.
5. The `LanguageServer` listens for messages on its `MessageQueue` and if they are a response type, looks up their respective handlers and dispatches them accordingly.

Utilities include:
* A worker thread-pool shared among requests (for such things as validating documents asynchronously).
* A JSON parser (JSON string -> `LSPAny`) and serializer (`LSPAny` -> JSON string).
* A thread-safe logger that writes to both `stderr` and a log file.
* Fixed-size, thread-safe message queues.
* A helpful exception type that can be thrown from anywhere and returned as the error for a response.
* Auto-generated transforms for translating messages between LSP representations (`LSPAny`, which is basically JSON, and more specific LSP types like `LogMessageParams`).

The language server base class `LanguageServer` defines the interface all language servers must implement. It is small and can be used to support protocols other than LSP without having to change the rest of the architecture.

A subtype of `LanguageServer`, `LspLanguageServer`, is generated from the LSP specification. It contains some boilerplate logic and method stubs for each of the LSP requests and notifications. It implements basic handlers for `initialize`, `initialized`, `shutdown`, and `exit`.

`BaseLspLanguageServer` extends `LspLanguageServer` with some useful boilerplate logic that should be helpful to all language servers regardless the target language. At the moment, that logic includes maintaining an in-memory corpus of files being edited by the client and configuration caching.

`LFortranLanguageServer` extends `BaseLspLanguageServer` and adds handlers for validating documents as they are opened and edited. It also supports pulling lfortran compiler flags from the client's configuration (such as includes) and changing the log level from the client's configuration.

The method handlers are virtual so to add support for one, just `override` its definition and it will be used to dispatch relevant requests.


dylon:
> All the files are very small, there are only two large files:
> 
> * `generate_lsp_code.py` (5,600 lines) --- that's fine for now. Later on we might be able to figure out how to make it even smaller
> * `metaModel.json` (14,835 lines) --- truly massive, but probably not much we can do, since the information there is the true complexity of the LSP itself. How is this file generated? Or is it copied from somewhere?

I'm sure we can make the code generator smaller. The JSON data model comes from the LSP spec: [Meta Model](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#metaModel). I considered downloading it as part of the build process but we wouldn't be able to build offline.

CMakeLists.txt
```diff
@@ -344,7 +344,8 @@ set(WITH_LSP no CACHE BOOL "Build with LSP support")
 if (WITH_JSON)
     find_package(RapidJSON REQUIRED)
     set(HAVE_LFORTRAN_RAPIDJSON yes)
-endif()
+endif(
+)
```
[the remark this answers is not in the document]
dylon: This is not something I would have done intentionally, it might be from a reformatting bug in my editor. I will fix it.

src/bin/lfortran.cpp
```diff
@@ -2254,61 +2246,24 @@ int main_app(int argc, char *argv[]) {
...
-    app.add_flag("--all-mangling", compiler_options.po.all_symbols_mangling, "Mangles all possible symbols");
-    app.add_flag("--bindc-mangling", compiler_options.po.bindc_mangling, "Mangles functions with abi bind(c)");
-    app.add_flag("--apply-fortran-mangling", compiler_options.po.fortran_mangling, "Mangle symbols with Fortran supported syntax");
-    app.add_flag("--mangle-underscore", compiler_options.po.mangle_underscore, "Mangles with underscore");
-    app.add_flag("--legacy-array-sections", compiler_options.legacy_array_sections, "Enables passing array items as sections if required");
-    app.add_flag("--ignore-pragma", compiler_options.ignore_pragma, "Ignores all the pragmas");
-    app.add_flag("--stack-arrays", compiler_options.stack_arrays, "Allocate memory for arrays on stack");
-    app.add_flag("--wasm-html", compiler_options.wasm_html, "Generate HTML file using emscripten for LLVM->WASM");
-    app.add_option("--emcc-embed", compiler_options.emcc_embed, "Embed a given file/directory using emscripten for LLVM->WASM");
-    app.add_flag("--mlir-gpu-offloading", compiler_options.po.enable_gpu_offloading, "Enables gpu offloading using MLIR backend");
```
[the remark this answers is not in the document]
dylon: Yes. I intended to make that clearer in my comment above the call to that function but I will document the change better. I needed a way to initialize an instance of `CompilerOptions` in the same way it would be initialized from the command-line. Any changes made to the command-line API should be available to the language server.

CMakeLists.txt
```diff
@@ -40,6 +40,11 @@ if (NOT WITH_XEUS)
     endif ()
 endif(NOT WITH_XEUS)
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
```
gxyd, line 44: I'm sorry, what's the use of '/bigobj' here?

src/bin/CMakeLists.txt
```diff
@@ -1,6 +1,33 @@
...
+if (WITH_LSP)
+    add_compile_definitions(WITH_LSP)
+    set(LFORTRAN_SRC
+        server/lfortran_lsp_language_server.cpp
+        server/interface.cpp
+        ${LFORTRAN_SRC}
+    )
+    set(LFORTRAN_LINK_LIBRARIES
+        ${LFORTRAN_LINK_LIBRARIES}
+        llanguage_server
```
gxyd, line 18: The use of "llanguage", is the initial "l" for LFortran's "L"?

src/bin/server/interface.cpp
```diff
@@ -0,0 +1,538 @@
...
+    auto CommandLineInterface::validateAndSetConfigSection() -> ExitCode {
+        if (isValidConfigSection(args.configSection)) {
+            opts.configSection = args.configSection;
+            return ExitCode::SUCCESS;
+        }
+        return ExitCode::INVALID_VALUE;
+    }
+
+    auto CommandLineInterface::prepare(CLI::App &app) -> CLI::App & {
+        CLI::App &server = *app.add_subcommand(
```
gxyd, line 260: with this subcommand being here, we probably wouldn't even show this when `lfortran --help` is done, when LFortran is built without LSP support. Is that allright?

Also, this is the only sub_command not present in `lfortran.cpp`, I'm not exactly sure if this is the best place for it.

gxyd (commented):
Left a few questions

src/bin/lfortran.cpp
```diff
@@ -2304,61 +2296,26 @@ int main_app(int argc, char *argv[]) {
...
     app.add_flag("--show-errors", show_errors, "Show errors when LSP is running in the background");
     app.add_flag("--show-document-symbols", show_document_symbols, "Show symbols in lfortran file");
 
+    CompilerOptions compiler_options;
+
+    // NOTE: All compiler_options initialization has been moved to the following
+    // utility function. This allows the language server to initialize them the
+    // same way as the command-line and it will keep its options up-to-date. All
+    // previous options have been moved here.
+    LCompilers::CommandLineInterface::init_compiler_options(compiler_options, app);
```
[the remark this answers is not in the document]
dylon: This has been resolved with the merging of https://github.com/dylon/lfortran/pull/1

CMakeLists.txt
dylon: There are a few hundred types generated from the LSP spec and MSVC complains about the size of the generated file without the `/bigobj` flag enabled.

src/bin/CMakeLists.txt
dylon: Yes, we are going to create a new repo for it like we have for libasr. It will be the LLanguage Server since we prefix most of our projects with `L` (LFortran, LPython, and LC are all part of LCompilers, and we will now have the LLanguage Server).

src/bin/server/interface.cpp
dylon: The subcommand is added to the interface from here and is available at the command-line menu. I added it to a separate file to avoid making `lfortran.cpp` too large. This module serves to initialize the server and is not part of the standard compilation pipeline of LFortran. It is tangential to the compiler so I wanted to isolate its logic in a separate file.

gxyd:
I went through the PR, the PR looks good to me overall. I agree that we probably should make LSP off by default, considering that it takes much longer now to build on this branch currently. On my machine, it's quite apparent.

PR #6296, opened 2025-02-12
Remove ``NOFAST`` from ``integration_tests/CMakeLists.txt``

czgdp1807:


czgdp1807:
I have removed `NOFAST` from `integration_tests/CMakeLists.txt` from wherever possible (without touching anything in the code base). Once this PR is merged (and the tests pass), I will remove those optimization passes which cause the remaining `NOFAST` flags to exist.

assem2002:
Do you think even though some llvm version work and some don't, that the issue still resides in our passes? not from llvm itself?

czgdp1807:
LLVM 15 and LLVM 16 are hanging on CI. Yesterday with the same commit, these two were passing on CI.

czgdp1807:
Seems like its good to go. More `NOFAST` labels will be removed in another PR.

Issue #6291, opened 2025-02-11
LFortran Raises Semantic Error for function declaration without pure/impure keyword for `std=f23` flag


AnshMehta1:
Based on comment https://github.com/lfortran/lfortran/pull/6282#issuecomment-2650498434

MRE :-

```.f90
program test
    implicit none
    real :: A(3)  
    A = rand1()

contains

    function rand1() result(x)
        implicit none
        real :: x(3)
        x = [1.1, 1.2, 1.3]
    end function

end program
```

LFortran Error with `std=f23` :- 

```console
(lf) lordansh@LordAnsh:~/dev/lfort3/lfortran$ lfortran try.f90 --std=f23
semantic error: No implicit return type available for `x`
  --> try.f90:8:5 - 12:16
   |
 8 |        function rand1() result(x)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
12 |        end function
   | ...^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Works with gfortran , lfortran without `std=f23` flag and also , it is not compulsory to mention pure/impure at function declaration as per fortran Standards .

HarshitaKalani:
Hey @kmr-srbh what gfortran version are you having?
With the -std=f2023, I get the following
```
$ gfortran a.f90 -std=f2023 && ./a.out
gfortran: error: unrecognized command-line option ‘-std=f2023’; did you mean ‘-std=f2003’?
$ gfortran a.f90 -std=f2018 && ./a.out
$ gfortran --version
GNU Fortran (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

kmr-srbh:
The above output was from GFortran 14.2.0.

PR #6298, opened 2025-02-12
fix: find return_type from attribute only when return var is null

HarshitaKalani:
Fixes: https://github.com/lfortran/lfortran/issues/6291


Pranavchiku (approved):

Pranavchiku:
Yes, I think so.

HarshitaKalani:
Yes, that's it.

PR #6302, opened 2025-02-13
ci: test if prima linux tests works

Pranavchiku:
We are able to run only one test on CI for linux: `test_uobyqa.f90`


Pranavchiku:
Yes, we can merge it.

PR #6310, opened 2025-02-14
Added ``replace_with_compile_time_values`` pass for replacing expressions with their compile time values

czgdp1807:
One example with `integration_tests/parameter_13.f90`,

```zsh
[(If (IntegerCompare (TypeInquiry Kind (Real 4) (Var 2 y) (Integer 4) (IntegerConstant 4 (Integer 4) Decimal)) NotEq (Var 2 sp) (Logical 4) (LogicalConstant .false. (Logical 4))) [(ErrorStop ())] []) (Print (StringFormat () [(Var 2 s)] FormatFortran (String -1 0 () PointerString) ())) (If (RealCompare (IntrinsicElementalFunction Abs [(RealBinOp (Var 2 s) Sub (RealConstant 2.000000 (Real 8)) (Real 8) (RealConstant 0.000000 (Real 8)))] 0 (Real 8) (RealConstant 0.000000 (Real 8))) Gt (RealConstant 0.000000 (Real 8)) (Logical 4) (LogicalConstant .false. (Logical 4))) [(ErrorStop ())] [])])}) [])
```

is reduced to,

```zsh
[(If (LogicalConstant .false. (Logical 4)) [(ErrorStop ())] []) (Print (StringFormat () [(RealConstant 2.000000 (Real 8))] FormatFortran (String -1 0 () PointerString) ())) (If (LogicalConstant .false. (Logical 4)) [(ErrorStop ())] [])])}) [])
```

This helps the latter passes to not process unnecessary information in the ASR.

I will apply it only in `--fast` mode. Just testing without `--fast` mode to cover corner cases.

### DIFF
--- a/tests/reference/llvm-array_bound_1-6741f43.json
+++ b/tests/reference/llvm-array_bound_1-6741f43.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-array_bound_1-6741f43.stdout",
-    "stdout_hash": "b20c2bd6981a514817950b10ce3721a1e33066979c78462977bd1b53",
+    "stdout_hash": "8f62d45e1c30066cad28a73dc1df90f1a0236be93461b5cdcc35ed5e",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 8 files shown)


czgdp1807:
> For now make sure this does not run by default, to keep our default pass pipeline as short as possible.

Sure. Let me make it only for `--fast` mode and remove it from default mode. Will that work?

Issue #6313, opened 2025-02-14
bug: `std::exception` with intrinsic `pack` when argument `array` is `ArrayConstructor` and `mask` is `ArrayConstant`


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none
   integer :: k

   print *, pack( [ (k, k = 1, 2) ], [.true., .true.] )
end program main
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
           1           2
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
std::exception: cannot create std::vector larger than max_size()
```

PR #6319, opened 2025-02-15
fix: handle compile time values correctly in `pack` intrinsic

HarshitaKalani:
Fixes: https://github.com/lfortran/lfortran/issues/6313

### DIFF
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -4818,15 +4818,18 @@ namespace Pack {
         if (is_type_allocatable) {
             ret_type = TYPE(ASRUtils::make_Allocatable_t_util(al, loc, ret_type));
         }
-        Vec<ASR::expr_t*> m_args; m_args.reserve(al, 2);
+        Vec<ASR::expr_t*> arg_values; arg_values.reserve(al, 3);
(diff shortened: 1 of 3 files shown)


src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -4818,15 +4818,18 @@ namespace Pack {
         if (is_type_allocatable) {
             ret_type = TYPE(ASRUtils::make_Allocatable_t_util(al, loc, ret_type));
         }
-        Vec<ASR::expr_t*> m_args; m_args.reserve(al, 2);
+        Vec<ASR::expr_t*> arg_values; arg_values.reserve(al, 3);
+        arg_values.push_back(al, expr_value(array)); arg_values.push_back(al, expr_value(mask));
```
Pranavchiku, line 4822: why do we need `arg_values`? Why can't just `m_args.reserve(al, 3)` and pushback `expr_value(vector)` work?

HarshitaKalani: We need expr_value of the arguments for eval_function but not instantiate.

Pranavchiku: okay, looks good.

Pranavchiku (approved):

Issue #5711, opened 2024-12-15
fix findloc to handle different  (conformable) types of `array` and `value` argument


HarshitaKalani:
```fortran
program main
    real(8), dimension(2) :: dc = [1.3, 2.3]
    real :: x = 1
    print *, findloc(dc, x)
end program
```
```console
$ gfortran a.f90 && ./a.out
           0
(lf) (base) harshita@harshita-linux:~/Desktop/lfortran$ lfortran a.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2732, in main
    return main_app(argc, argv);
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2670, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
...
  File "/home/harshita/Desktop/lfortran/src/libasr/pass/intrinsic_function.cpp", line 137, in LCompilers::ReplaceIntrinsicFunctions::replace_IntrinsicArrayFunction(LCompilers::ASR::IntrinsicArrayFunction_t*)
    ASR::expr_t* current_expr_ = instantiate_function(al, x->base.base.loc,
  File "/home/harshita/Desktop/lfortran/src/libasr/../libasr/pass/intrinsic_array_function_registry.h", line 3572, in LCompilers::ASRUtils::FindLoc::instantiate_FindLoc(Allocator&, LCompilers::Location const&, LCompilers::SymbolTable*, LCompilers::Vec<LCompilers::ASR::ttype_t*>&, LCompilers::ASR::ttype_t*, LCompilers::Vec<LCompilers::ASR::call_arg_t>&, long)
    b.If(b.Eq(ArrayItem_02(array, i), value), {
AssertFailed: check_equal_type(expr_type(left), expr_type(right))
```

HarshitaKalani:
The following should work too.
```fortran
program main
    integer(8), dimension(2) :: dc = [1.4, 2.3]
    real :: x = 1.3
    print *, findloc(dc, x)
end program
```
```console
$ gfortran a.f90 && ./a.out
           1
[...]

PR #6331, opened 2025-02-16
enh: enhance finddloc to work for same type but different kinds of arguments

HarshilShah1804:
Towards #5711

### DIFF
--- a/integration_tests/intrinsics_360.f90
+++ b/integration_tests/intrinsics_360.f90
@@ -1,0 +1,11 @@
+program intrinsics_360
+    real(8), dimension(2) :: dc_r = [1.3, 2.3]
+    real :: x_r = 1.3
+    print *, findloc(dc_r, x_r)
+    if (any(findloc(dc_r, x_r) /= 1)) error stop
+
+    integer, dimension(2) :: dc_i = [1, 2]
+    integer(8) :: x_i = 2
+    print *, findloc(dc_i, x_i)
+    if (any(findloc(dc_i, x_i) /= 2)) error stop
+end program
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -3407,6 +3407,29 @@ namespace FindLoc {
                 "real, logical, character or complex type", loc);
             return nullptr;
         }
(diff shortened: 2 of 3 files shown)


integration_tests/intrinsics_360.f90
```diff
@@ -0,0 +1,11 @@
+program intrinsics_360
+    real(8), dimension(2) :: dc_r = [1.3, 2.3]
+    real :: x_r = 1.3
+    print *, findloc(dc_r, x_r)
+    if (any(findloc(dc_r, x_r) /= 1)) error stop
+
+    integer, dimension(2) :: dc_i = [1, 2]
+    integer(8) :: x_i = 2
```
gxyd, line 8: Fortran standard and GFortran compiler doesn't allow these declarations after any statements in a program. Hence, we would need to move these declarations from here to the top along with the rest of the declarations.

Currently LFortran doesn't raise an error for this, which it should, and it's a known issue.

gxyd (commented):
The approach is good in this PR, just that we need to re-use existing code (which I can think might be possible here) and fix the test case to pass with GFortran (which it currently fails)

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3407,6 +3407,29 @@ namespace FindLoc {
...
+                    value = EXPR(ASR::make_IntegerConstant_t(
+                        al, loc, ASR::down_cast<ASR::IntegerConstant_t>(value)->m_n,
+                        ASRUtils::TYPE(ASR::make_Integer_t(al, loc, extract_kind_from_ttype_t(array_type)))));
+                } else {
+                    value = EXPR(ASR::make_Cast_t(
+                        al, loc, value, ASR::cast_kindType::IntegerToInteger,
+                        ASRUtils::TYPE(ASR::make_Integer_t(al, loc, extract_kind_from_ttype_t(array_type))), nullptr));
+                }
+            } 
+        }
```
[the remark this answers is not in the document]
HarshilShah1804: Okay. I'll move it to `asr_utils.h`. Then modify it to use it here also.

integration_tests/intrinsics_360.f90
HarshilShah1804: Changed. Thanks.

PR #6332, opened 2025-02-16
don't merge: run validation script for POT3D on macOS

gxyd:
## Description

with this PR, I only would like to see the stacktrace error that we get when the validation fails on macOS

this PR isn't intended to be merged at all


gxyd:
Closing this PR, as it's not needed anymore.

PR #6315, opened 2025-02-14
fix: create temporary when `ArraySection` pointer is passed to funccall 

jinangshah21:
Fixes #6214

PR #6336, opened 2025-02-16
fix: create temporary for non-contiguous descriptor passed in function

### LINKED ISSUE #6214: Array passed in chain of function as argument  gets corrupted
MRE:
```fortran 
module function_32_mod
contains
function istril(y) result(z)
    logical :: z
    z = .true.
    print *, y(:, 1)
    print *, y(:, 2)
end function
function matprod(y) result(z)
    real(8) :: z(size(y, 2))
    if(istril(y)) print *, 'yes'
    z = 0.0_8
end function matprod
end module

program function_32
    real(8) :: A(5,3)
    A = 1.0_8
    A(1,2) = 0
    A(2,2) = 0
    call func(A)
contains 
    subroutine func(A)
        use function_32_mod
        real(8), intent(in) :: A(:,:)
        real(8) :: dnew(size(A,2))
        integer :: nact
        nact = 2
        print *, A(1:nact, 1)
        print *, A(1:nact, 2)
        dnew(1:nact) = -matprod(A(1:nact,1:nact))
    end subroutine
end program 
```

GFortran Output: 
```console
1.0000000000000000        1.0000000000000000     
   0.0000000000000000        0.0000000000000000     
   1.0000000000000000        1.0000000000000000     
   0.0000000000000000        0.0000000000000000     
 yes
```
LFortran Output : 
```console
1.00000000000000000e+00    1.00000000000000000e+00
0.00000000000000000e+00    0.00000000000000000e+00
1.00000000000000000e+00    1.00000000000000000e+00
1.00000000000000000e+00    1.00000000000000000e+00
yes
```

### LINKED ISSUE #6329: When a temporary has to be created
```fortran
program test_temporaries
implicit none
real(dp) :: A(5,3)
A = 1
A(1,2) = 2
A(2,2) = 3
print *, A(1:2,1:2)
call f(A(1:2,1:2))

contains

    subroutine f(y)
    real(dp), intent(in) :: y(:, :)
    call g(size(y,1), size(y, 2), y)
    end subroutine

    subroutine g(n, m, y)
    integer, intent(in) :: n, m
    real(dp), intent(in) :: y(n, m)
    print *, y(2, 2)
    end subroutine

end program
```
This gives:
```console
$ gfortran -Wall -fcheck=all a.f90
$ ./a.out
   1.0000000000000000        1.0000000000000000        2.0000000000000000        3.0000000000000000
At line 15 of file a.f90
Fortran runtime warning: An array temporary was created for argument 'y' of procedure 'g'
   3.0000000000000000
```
Line 15 is the line
```
call g(size(y,1), size(y, 2), y)
```

Now when you change `call f(A(1:2,1:2))` to `call f(A)` then it runs without creating any temporary.

From this it follows:
* When compiling the function `f`, we accept a descriptor array and it can be strided or not, and it is a runtime decision (in general), since we can for example read an input file and decide if to call `f` using `call f(A)` or using `call f(A(1:2,1:2))`.
* On the line `call g(size(y,1), size(y, 2), y)` we are converting a descriptor array to a pointer (contiguous) array as follows, based on the run-time check of the descriptor:
  * If the array is not strided, then we pass the pointer directly
  * if the array is strided, we make a copy --- and we copy the result back after the function is done for `intent(out)` and `intent(inout)` arguments

jinangshah21:
Fixes #6214
Closes #6329

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -515,8 +515,8 @@ jobs:
         run: |
             git clone https://github.com/Pranavchiku/prima.git
             cd prima
-            git checkout -t origin/lf-prima-6
-            git checkout c66cf5ecd63caf94e3559899a6997a59c1909c63
+            git checkout -t origin/lf-prima-7
+            git checkout a2a741bec839182a18426dd2ea27721a9ffb4422
             git clean -dfx
             FC="$(pwd)/../src/bin/lfortran --cpp" cmake -S . -B build -DCMAKE_INSTALL_PREFIX=$(pwd)/install -DCMAKE_Fortran_FLAGS=""  -DCMAKE_SHARED_LIBRARY_CREATE_Fortran_FLAGS=""  -DCMAKE_MACOSX_RPATH=OFF -DCMAKE_SKIP_INSTALL_RPATH=ON  -DCMAKE_SKIP_RPATH=ON && cmake --build build --target install
             ./build/fortran/example_bobyqa_fortran_1_exe


jinangshah21:
Transforms :
```fortran
subroutine matprod(y)
    real(8), intent(inout) :: y(:, :)
    y(1,2) = 2
    call istril(y)
end subroutine
```
To: 
```fortran
subroutine matprod(y)
    real(8), dimension(:, :), pointer :: y_tmp
    real(8), dimension(:, :), intent(inout) :: y
    y(1, 2) = real(2, kind=8)
    if (.true.) then
        deallocate(y_tmp)
        allocate(y_tmp(size(y, 1), size(y, 2)))
        y_tmp = y
    else
y_tmp => y
    end if
    call istril(y_tmp)
    if (.true.) then
        y = y_tmp
    end if
end subroutine matprod
```

jinangshah21:
Currently `functions_22` fails and it is related to `pass_array_by_data`, I think it is a bug exposed by this PR, we need to fix it there.

jinangshah21:
PRIMA also fails, it is also related to `pass_array_by_data`.
I will look it tomorrow.

jinangshah21:
Consider below test :
```
module present_01_module
implicit none

contains

   subroutine foo1(a)
      real, intent(in), optional :: a(:)
      call foo2(a)  
   end subroutine

   subroutine foo2(a)
      real, intent(in), optional :: a(..)
      if (present(a)) error stop
   end subroutine

end module

program present_01
use present_01_module
implicit none

   call foo1()

end
```

When we call `foo2(a)` we can't use temporary (`a_tmp`) if a is not present in `foo1`.

jinangshah21:
Issue in `pass_array_by_data`: as our temporary (created in this PR) is a pointer and `pass_array_by_data` didn't convert calls who's arguments are pointer. So everytime we have pointer as arg, so we are never able to convert calls with array arguments that pass.

jinangshah21:
Above 2 are the current failure in tests, other than `integration_tests` passes. 
Then we have PRIMA failure which is likely same as issue with `pass_array_by_data`

jinangshah21:
With latest commit `lincoa` examples are segfaulting. All other examples work fine.

jinangshah21:
@Pranavchiku, can you look `cobyla` segfault that is only coming for mac?

Pranavchiku:
Please find https://github.com/lfortran/lfortran/pull/6341 with equivalent status as #6315.

jinangshah21:
I have addressed https://github.com/lfortran/lfortran/pull/6336#issuecomment-2662318254 by doing something like :

```fortran 
real, pointer :: tmp(:)
if(present(a))
   if(.true.) 
      allocate(a(size(a,1))
      tmp = a
   else
      tmp => a
   end if
else 
   tmp => null()
end if 

call foo2(tmp)
```

But there is bug exposed in doing this (Reported at #6354). I am currently disabling `present_01`, once #6354 will be fixed we can enable it.

PR #6341, opened 2025-02-17
enh: handling of array section passed to function call

### LINKED ISSUE #6214: Array passed in chain of function as argument  gets corrupted

### LINKED ISSUE #6329: When a temporary has to be created

Pranavchiku:
Fixes #6329. Fixes #6214.

TODO:

If willing to merge PR without fix, just do `b.logical_false()` and everything should work.

If willing to merge with Fix, wait for `is_contiguous`

### DIFF
--- a/src/libasr/pass/array_struct_temporary.cpp
+++ b/src/libasr/pass/array_struct_temporary.cpp
@@ -1520,20 +1520,31 @@ class ArgSimplifier: public ASR::CallReplacerOnExpressionsVisitor<ArgSimplifier>
         LCOMPILERS_ASSERT(!x.m_dt || !ASRUtils::is_array(ASRUtils::expr_type(x.m_dt)));
         Vec<ASR::call_arg_t> x_m_args; x_m_args.reserve(al, x.n_args);
         std::vector<bool> is_arg_intent_out;
-        ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(ASRUtils::symbol_get_past_external(x.m_name));
-        ASR::FunctionType_t* func_type = ASR::down_cast<ASR::FunctionType_t>(func->m_function_signature);
-        bool is_func_bind_c = func_type->m_abi == ASR::abiType::BindC || func_type->m_deftype == ASR::deftypeType::Interface;
-        for (size_t i = 0; i < func->n_args; i++ ) {
-            if ( ASR::is_a<ASR::Var_t>(*func->m_args[i]) ) {
-                ASR::Variable_t* var = ASRUtils::EXPR2VAR(func->m_args[i]);
-                is_arg_intent_out.push_back(var->m_intent == ASR::intentType::Out ||
-                    var->m_intent == ASR::intentType::InOut);
-            } else {
-                is_arg_intent_out.push_back(false);
+        if ( ASR::is_a<ASR::Function_t>(*ASRUtils::symbol_get_past_external(x.m_name)) ) {
+            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(ASRUtils::symbol_get_past_external(x.m_name));
+            ASR::FunctionType_t* func_type = ASR::down_cast<ASR::FunctionType_t>(func->m_function_signature);
+            bool is_func_bind_c = func_type->m_abi == ASR::abiType::BindC || func_type->m_deftype == ASR::deftypeType::Interface;
+            for (size_t i = 0; i < func->n_args; i++ ) {
+                if ( ASR::is_a<ASR::Var_t>(*func->m_args[i]) ) {
+                    ASR::Var_t* var_ = ASR::down_cast<ASR::Var_t>(func->m_args[i]);
+                    if ( ASR::is_a<ASR::Variable_t>(*var_->m_v) ) {
+                        ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(var_->m_v);
+                        is_arg_intent_out.push_back(var->m_intent == ASR::intentType::Out ||
+                            var->m_intent == ASR::intentType::InOut);
+                    } else {
+                        is_arg_intent_out.push_back(false);
+                    }
+                } else {
+                    is_arg_intent_out.push_back(false);
+                }
             }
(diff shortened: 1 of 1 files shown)


jinangshah21:
I think this is better then what I did in #6336, we are not creating less temporary here so it will fast also.

jinangshah21:
Here `A(1:2,1:2)` will be directly passed to `is_contiguous`.

Pranavchiku:
I think so yes.

Pranavchiku:
It will work, `print *, is_contiguous(A(1:2, 1:2))` gives `False` with GFortran so we are good.

jinangshah21:
Then we have to investigate POT3D failure

Pranavchiku:
It is fine I guess, it will automatically work the moment we put in `is_contigous(xx)` check. Or not maybe, I think we need to investigate it.

tests/reference/llvm-allocate_03-495d621.stdout
```diff
@@ -1,7 +1,7 @@
 ; ModuleID = 'LFortran'
 source_filename = "LFortran"
 
-%array = type { i32*, i32, %dimension_descriptor*, i1, i32 }
+%array = type { ptr, i32, ptr, i1, i32 }
```
Pranavchiku, line 4: These changes are weird. I don't think these should have been there, probably incorrect LLVM version

Pranavchiku: Yes it is using LLVM 19

gxyd (commented):
Looks pretty good, but I think we should add a test case for the last commit pushed.

src/libasr/pass/array_struct_temporary.cpp
gxyd, line 1: For this commit i.e. dc7f9e585339f9fffb717fb53a711072b6eb5a60 we haven't added a test case.

Pranavchiku:
> I think we should add a test case for the last commit pushed.

Will be waiting for POT3D MRE :)

src/libasr/pass/array_struct_temporary.cpp
Pranavchiku: Yes, we have it, will do once this PR is polished.

src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -832,8 +850,49 @@ ASR::expr_t* create_and_allocate_temporary_variable_for_array(
...
+            alloc_arg.m_a = array_var_temporary;
+            alloc_arg.m_dims = allocate_dims.p;
+            alloc_arg.n_dims = allocate_dims.size();
+            alloc_arg.m_len_expr = nullptr;
+            alloc_arg.m_type = nullptr;
+            alloc_args.push_back(al, alloc_arg);
+        
+            Vec<ASR::expr_t*> dealloc_args; dealloc_args.reserve(al, 1);
+            dealloc_args.push_back(al, array_var_temporary);
+            ASR::expr_t* is_contiguous = b.logical_true();
```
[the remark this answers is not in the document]
Pranavchiku: We will not be merging it till we have `is_contihuous`

czgdp1807 (approved):

PR #6364, opened 2025-02-19
CMake enhancements - file generation, out of source, presets

Mazrog:
Adding custom commands for file generation, getting the version from git, removing the need not to forget to start build0.sh/bat when touching a non-source file.

Adding basis for CMake presets, allowing to define profiles and variables in a shareable way instead of command line or custom scripts (having to maintain them in win/linux for instance).

Draft: as I am trying to get it to work with clang on windows, I still have some compilation issue left to fix atm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -201,6 +201,7 @@ MANIFEST
 *~
 
 ## Build Files
+build/
 */bin/lfortran
 output
 *.o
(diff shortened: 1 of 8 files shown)


Mazrog:
Why are we creating source tarballs just to unpack them right after?
The building scripts are all over the place and all different, it makes it very difficult to follow :(

PR #6371, opened 2025-02-20
Use values with `icmp` instruction in llvm not pointers

assem2002:


assem2002:
It used to work due to the unnecessary `load` --> `  %13 = load i8*, i8* %11, align 8`

Issue #6370, opened 2025-02-20
intrinsic: generated code for `cshift` accesses out of bound index


gxyd:
## Minimal Reproducible Example

For the below Fortran program (let's call it `pot3d.f90`):

```f90
program pot3d
      implicit none
      integer(4), parameter :: dp = 8
      integer(4), save :: n = 62
      real(8), dimension(:), allocatable :: r_g
      allocate(r_g(n))
      call genmesh(n, r_g)
      deallocate(r_g)

contains

subroutine genmesh(nc, c)
    integer(4), intent(in) :: nc
    real(8), dimension(nc), intent(inout) :: c
    real(8), dimension(nc) :: dc
    integer(4) :: v
    v = int(1, kind=4) - 1
    do while (v + 1 <= nc + 1 - 1)
        v = v + 1
        c(v) = real(0, kind=8)
    end do
    c(nc) = 2.5000000000000000e+00_8
    dc = cshift(c, -1)
end subroutine genmesh
end program pot3d

```

### Output with GFortran
```console
> gfortran -fcheck=all -g -c pot3d.f90
> gfortran pot3d.o -o pot3d
> ./pot3d
> echo $?
0
```

Now, the file to be used which actually has errors in it is created via:
```console
> lfortran pot3d.f90 --dump-all-passes-fortran --apply-fortran-mangling
> gfortran -fcheck=all -g -c pass_fortran_31_insert_deallocate.f90
> gfortran pass_fortran_31_insert_deallocate.o -o pass_fortran_31_insert_deallocate
> ./pass_fortran_31_insert_deallocate
At line 55 of file pass_fortran_31_insert_deallocate.f90
Fortran runtime error: Index '63' of dimension 1 of array 'result' above upper bound of 62

Error termination. Backtrace:
#0  0x10312fcf7
#1  0x103130b07
#2  0x103130ebf
#3  0x10294b267
#4  0x10294b78b
#5  0x10294b4bb
#6  0x10294b7df
```

### Reference

The `-fcheck` option mentioned above is from : https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html, which says:

* Enable the generation of run-time checks; the argument shall be a comma-delimited list of the following keywords. Prefixing a check with no- disables it if it was activated by a previous specification.*

HarshitaKalani:
Closing this as completed with: #6372
and opened https://github.com/lfortran/lfortran/issues/6374 for the new issue.

[...]

PR #6372, opened 2025-02-20
fix: handle indices correctly in `cshift` intrinsic 

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/6370

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -2573,11 +2573,11 @@ public:
                 break;
             }
             case ASR::array_physical_typeType::FixedSizeArray:
+            case ASR::array_physical_typeType::SIMDArray:
                 tmp = llvm::ConstantInt::get(context, llvm::APInt(1, 1));
                 break;
             case ASR::array_physical_typeType::PointerToDataArray:
             case ASR::array_physical_typeType::StringArraySinglePointer:
-            case ASR::array_physical_typeType::SIMDArray:
                 tmp = llvm::ConstantInt::get(context, llvm::APInt(1, 0));
                 break;
             default:
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -1727,11 +1727,12 @@ namespace Cshift {
         }
         ));
         body.push_back(al, b.Assignment(i, b.i32(1)));
+        
         body.push_back(al, b.DoLoop(j, b.Add(shift_val, b.i32(1)), UBound(args[0], 1), {
             b.Assignment(b.ArrayItem_01(result, {i}), b.ArrayItem_01(args[0], {j})),
             b.Assignment(i, b.Add(i, b.i32(1))),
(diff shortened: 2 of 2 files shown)


HarshitaKalani:
With this, the error is now different:
```console
$ cat lf.sh
lfortran a.f90 --dump-all-passes-fortran --apply-fortran-mangling
gfortran -fcheck=all -g -c pass_fortran_31_insert_deallocate.f90
gfortran pass_fortran_31_insert_deallocate.o -o pass_fortran_31_insert_deallocate
./pass_fortran_31_insert_deallocate
$ sh lf.sh 
At line 10 of file pass_fortran_31_insert_deallocate.f90
Fortran runtime error: Attempt to DEALLOCATE unallocated 'r_g'

Error termination. Backtrace:
#0  0x7a2747823960 in ???
#1  0x7a27478244d9 in ???
#2  0x7a2747824ad6 in ???
#3  0x5953aa69777e in pot3d
        at /home/harshita/Desktop/lfortran/pass_fortran_31_insert_deallocate.f90:10
#4  0x5953aa697a2a in main
        at /home/harshita/Desktop/lfortran/pass_fortran_31_insert_deallocate.f90:10
```

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1727,11 +1727,12 @@ namespace Cshift {
         }
         ));
         body.push_back(al, b.Assignment(i, b.i32(1)));
+        
         body.push_back(al, b.DoLoop(j, b.Add(shift_val, b.i32(1)), UBound(args[0], 1), {
             b.Assignment(b.ArrayItem_01(result, {i}), b.ArrayItem_01(args[0], {j})),
             b.Assignment(i, b.Add(i, b.i32(1))),
         }, nullptr));
-        body.push_back(al, b.DoLoop(j, LBound(args[0], 1), b.Add(shift_val, b.i32(1)), {
+        body.push_back(al, b.DoLoop(j, LBound(args[0], 1), shift_val, {
```
gxyd, line 1735: Looks good this one.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2573,11 +2573,11 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 break;
             }
             case ASR::array_physical_typeType::FixedSizeArray:
+            case ASR::array_physical_typeType::SIMDArray:
```
gxyd, line 2576: why do we need this change?

gxyd (commented):
The change looks good to me. I'm still thinking if there is a way to add a test case here.

Also, I checked it locally on my machine and this change allows us to remove the workaround from POT3D

src/libasr/codegen/asr_to_llvm.cpp
gxyd: Ah, I see. Would it be alright to create a separate PR for this instead? I was a little confused by this change. Since the changes aren't related, unless you think they are, let me know.

Issue #6312, opened 2025-02-14
bug: cannot initialize a `parameter` annotated variable with `reshape`


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none
   integer, parameter :: arr(*) = reshape( [1, 2], [2] )
   print *, arr
end program main
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
           1           2
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
Traceback (most recent call last):
  Binary file "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2769, in main
    return main_app(argc, argv);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2707, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 1008, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    result = fe.get_asr2(input, lm, diagnostics);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/lfortran/fortran_evaluator.cpp", line 275, in LCompilers::FortranEvaluator::get_asr2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
[...]

PR #6373, opened 2025-02-20
Fixed initialisation of ``parameter`` attributed deferred shape array with ``reshape``

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/6312

Simple fix. Two things were needed,

1. Handling `value = nullptr` correctly.
2. Setting shape if right size is `ArrayReshape` and the LHS is a `parameter` array variable.


czgdp1807:
I made this PR while I was free.

PR #6379, opened 2025-02-22
DON'T MERGE: CI: check to see if the workaround can be removed

gxyd:
## Description

This PR isn't intended to be merged at all, only intends to check whether the workaround https://github.com/gxyd/POT3D/commit/432ee83a1a33f0e066063c3c4082eee2b80b6e48 can be removed or not.

At the CI, even if the tests pass once on this PR, I'll re-run it to make sure that no random segfault occurs, and if the segfault does occur even once, it will be a sign that we need to create an MRE and then fix the issue as well.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -565,8 +565,8 @@ jobs:
             FC="$(pwd)/src/bin/lfortran"
             git clone https://github.com/gxyd/pot3d.git
             cd pot3d
-            git checkout -t origin/build_pot3d_custom_mpi_wrappers_and_without_hdf5
-            git checkout de3890cc12cb6a3c1b15710bb82f4c3e816cd59f
+            git checkout -t origin/check_workaround_removal
+            git checkout 3adb758855198924704a4972a07da5191c0aa783
             cd src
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -565,8 +565,8 @@ jobs:
             FC="$(pwd)/src/bin/lfortran"
             git clone https://github.com/gxyd/pot3d.git
             cd pot3d
-            git checkout -t origin/build_pot3d_custom_mpi_wrappers_and_without_hdf5
-            git checkout de3890cc12cb6a3c1b15710bb82f4c3e816cd59f
+            git checkout -t origin/check_workaround_removal
```
gxyd, line 568: This branch will anyway never be used, and since I don't like force pushing, I created this *commit reversal* on a separate branch, so I never have to make a bad commit on our long running `build_pot3d_custom_mpi_wrappers_and_without_hdf5` branch.

gxyd:
Closing this PR, as the purpose of opening this PR i.e. checking if the tests fail on Linux or not is fulfiled

Issue #6297, opened 2025-02-12
`pad` and `order` argument not working for runtime cases


HarshilShah1804:
MRE:
```fortran
program main
    implicit none
    integer :: original(4) = [1, 2, 3, 4]
    integer :: reshaped(2, 3)
    reshaped = reshape(original, shape=[2, 3], pad=[0])
    print *, reshaped
end program 
```
```console
lfortran a.f90
1    2    3    4    1039285880    32764
```
```console 
gfortran a.f90 && ./a.out
           1           2           3           4           0           0
```

Pranavchiku:
@HarshilShah1804 are you working on this?

HarshilShah1804:
> @HarshilShah1804 are you working on this?

Yes, @Pranavchiku I am working on it.

PR #6385, opened 2025-02-23
enh: implement `pad` in `reshape` for runtime cases

HarshilShah1804:
Towards #6297 
We need to strengthen the error condition in the error message. As it gives error for arrays other than 1-D in runtime cases. I have commented it out for now.


Pranavchiku (commented):
You also need to implement it for other physical types. The current one is for `DescriptorArray`. For one example, does it sometimes work sometime segfaults? or if it works it works, for the cases it does not work it segfaults?

HarshilShah1804:
> You also need to implement it for other physical types. The current one is for `DescriptorArray`. For one example, does it sometimes work sometime segfaults? or if it works it works, for the cases it does not work it segfaults?

For any example, it sometimes works and sometimes give segfaults randomly ?

HarshilShah1804:
> This may happen due to multiple reasons, are you initialising new variables and using it? Then make sure to correctly initialise it with a value, else it will have junk and show the behaviour you stated above. There might be cases when accessing descriptor is incorrect, check that as well.

Ok. I'll check that.

Pranavchiku (commented):
- submit the example as a part of this PR
- add comments in your code to explain what the snippet is doing ( write corresponding Fortran / C code in comment )

PR #6391, opened 2025-02-24
Implement --time-report for LLVM compilation also

certik:


src/bin/lfortran.cpp
```diff
@@ -1047,8 +1064,11 @@ int compile_src_to_object_file(const std::string &infile,
         return 1;
 #endif
     }
+    t2 = std::chrono::high_resolution_clock::now();
```
gxyd, line 1067:
```suggestion
    t1 = std::chrono::high_resolution_clock::now();
```

src/bin/lfortran.cpp
```diff
@@ -1094,6 +1120,22 @@ int compile_src_to_object_file(const std::string &infile,
...
+    if (time_report) {
+        std::cout << "Allocator usage of last chunk (MB): "
+            << fe.get_al().size_current() / (1024. * 1024) << std::endl;
+        std::cout << "Allocator chunks: " << fe.get_al().num_chunks() << std::endl;
+        std::cout << std::endl;
+        std::cout << "Time report:" << std::endl;
+        std::cout << "File reading:" << std::setw(5) << time_file_read << " ms" << std::endl;
+        std::cout << "Src -> ASR:  " << std::setw(5) << time_src_to_asr << " ms" << std::endl;
+        std::cout << "ASR -> mod:  " << std::setw(5) << time_save_mod << " ms" << std::endl;
+        std::cout << "ASR -> LLVM: " << std::setw(5) << time_asr_to_llvm << std::endl;
```
gxyd, line 1132: Can we add an " ms" after each of these as well?

src/bin/lfortran.cpp
```diff
@@ -1094,6 +1120,22 @@ int compile_src_to_object_file(const std::string &infile,
...
+        std::cout << "LLVM opt:    " << std::setw(5) << time_opt << std::endl;
```
gxyd, line 1133: Here also, would be nice.

src/bin/lfortran.cpp
```diff
@@ -1047,8 +1064,11 @@ int compile_src_to_object_file(const std::string &infile,
         return 1;
 #endif
     }
+    t2 = std::chrono::high_resolution_clock::now();
     LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
         res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
```
gxyd, line 1069: Maybe a separate query from me, we already do fast optimizations here:
https://github.com/lfortran/lfortran/blob/5e38b24b684d3a438398ead62a5b676c5f5feb48/src/lfortran/fortran_evaluator.cpp#L390-L392

Do we end up doing optimizing twice when compiling and running a file? Because, below this also we've a code block:
```c++
if (compiler_options.po.fast) {
    e.opt(*m->m_m);
}
```

Is this intentional?

gxyd (commented):
Other than a few comments, this looks good to me.

src/bin/lfortran.cpp
gxyd: Yes, I've reported an issue for this bug here: https://github.com/lfortran/lfortran/issues/6405

Issue #6318, opened 2025-02-15
Potential issue with loading binary file


certik:
This is the second time I've seen this error: https://github.com/lfortran/lfortran/actions/runs/13316578118/job/37191966530?pr=6295
```
/home/runner/work/lfortran/lfortran/src/bin/lfortran --realloc-lhs CMakeFiles/test_chat.dir/tests/test_chat.f90.o -o test_chat   -L/usr/lib/gcc/x86_64-linux-gnu/13  -Wl,-rpath,/home/runner/micromamba/envs/lf/lib libfastgpt.a /home/runner/micromamba/envs/lf/lib/libopenblas.so /home/runner/micromamba/envs/lf/lib/libomp.so -lgcc -lgcc_s -lc -lgcc -lgcc_s
make[2]: Leaving directory '/home/runner/work/lfortran/lfortran/fastGPT/lf'
[100%] Built target test_chat
make[1]: Leaving directory '/home/runner/work/lfortran/lfortran/fastGPT/lf'
/home/runner/micromamba/envs/lf/bin/cmake -E cmake_progress_start /home/runner/work/lfortran/lfortran/fastGPT/lf/CMakeFiles 0
+ ln -s ../model.dat .
+ ./gpt2
ERROR STOP Invalid fastGPT model file
Loading the model...
Found:    1329865020
Expected:    262477463
```

I am opening this issue to start tracking it.

gxyd:
All the four times that you reported the issue, it occured on ubuntu OS, not on macOS.

gxyd:
And shasum is wrong here: 
```console
+ shasum -a 256 model.dat
983919968e402aeafe98a1bb41fb240d59ceb7db59a792cc2631c4e6b3e41bc8  model.dat
```

in all the passing jobs it's correct (i.e. `11f6f018794924986b2fdccfbe8294233bb5e8ba28d40ae971dec3adbdc81ad7`)

PR #6392, opened 2025-02-24
CI: print the SHA256 of model.dat to check it

certik:
That way we can ensure that the downloaded file is correct.

Towards #6318.


gxyd (approved):
Looks good to me.

PR #6368, opened 2025-02-20
Completely remove `NOFAST` tags from `integration_tests/CMakeLists.txt`

czgdp1807:
To merge this PR following needs to be done (taken from https://github.com/lfortran/lfortran/pull/6357)

- [ ] Remove the application of optimisation passes from in between the normal passes. Append them to the normal passes.
- [ ] Remove NOFAST tags from working tests. This takes time because of CI. I can make two PRs today to figure out the erroring tests.
- [ ] Then comment out optimisation passes and remove NOFAST from tests which error.
- [ ] Finally remove --epxerimental-fast and merge it with --fast. Remove the additional optimisation passes which fail the 3rd party tests

Figuring out exactly which optimisation passes work and which don't will take the maximum time out of these.

https://github.com/lfortran/lfortran/pull/6357 should be merged first, I will rebase this PR on top of `main` afterwards.

Closes https://github.com/lfortran/lfortran/issues/6333

PR #6394, opened 2025-02-24
Remove ``NOFAST_TILL_LLVM16`` from ``cmd_02.f90``

czgdp1807:
Fixes LLVM error in https://github.com/lfortran/lfortran/pull/6368#issuecomment-2671886744


czgdp1807:
We were loading, `i8**` twice, it was being caught in LLVM 11-16 `—fast` mode.

czgdp1807:
LLC for LLVM 11 was catching the bug but LLVM 11 itself was letting the code go through its optimisation passes. May be a bug in LLVM 11 C++ API.

PR #6395, opened 2025-02-24
CI: remove `--realloc-lhs` from `test_llvm` CI job for fastGPT

gxyd:
## Description

I think this was forgotten to be done in the PR: https://github.com/lfortran/lfortran/pull/6090. Let's see if the tests pass or there is a specific version of LLVM for which this might fail.


gxyd:
Apologies for the force push, as I mistakenly created a branch with the same name as this one, and hence ended up force-pushing the wrong changes.

Now the changes are alright, after a double force-push on this PR.

PR #6406, opened 2025-02-25
CI: use git-lfs instead of curl to get fastGPT model file

gxyd:
## Description

An attempt to fix https://github.com/lfortran/lfortran/issues/6318


gxyd:
The run passed for the first time here, I'll re-run this to see if it succeeds again or not.

gxyd:
Second time also, the run passed, I'll do a re-run once again now.

gxyd:
git lfs method is a much more reliable (as it's specifically for large files like model files) method than using `curl`, where we'll also need to ensure that the sha of the downloaded file is correct or not, and it resumes download automatically if the download fails.

Also, in our case downloading the entire repository i.e. https://huggingface.co/datasets/certik/fastGPT/tree/main doesn't add any overhead compared to just downloading the model file, because the repository contains just two other small files < 3 kb.

Albeit, this PR adds a extra git commands, hence may not be as fast as using `curl`, but I find git lfs method to be more reliable than using `curl` to obtain the model file.

Issue #6229, opened 2025-02-06
reading comma separated string into an array of values


gxyd:
## Minimal Reproducible Example

```f90
program main
    implicit none
    character(20) :: key
    character(60) :: value
    integer, save, dimension(3) :: nprocs=[-1,-1,-1]
    integer, save :: nr=0
    key = "nprocs"
    value = "133,134,135"
    select case (key)
        case ("nr")
            read(value, *) nr
        case ("nprocs")
            read(value, *) nprocs
        case default
            write(*, *) "WARNING: Unknown key:", key
    end select
    print *, nprocs
end program main

```

For the above Fortran program, on it's execution with LFortran, I get a codegen error:
```console
> lfortran --version
LFortran version: 0.42.0-1297-gf7629f240-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

@main.nr = internal global i32 0
@main.nprocs = internal global [3 x i32] zeroinitializer
@0 = private unnamed_addr constant [7 x i8] c"nprocs\00", align 1
@1 = private unnamed_addr constant [12 x i8] c"133,134,135\00", align 1
@2 = private unnamed_addr constant [3 x i8] c"nr\00", align 1
@3 = private unnamed_addr constant [3 x i8] c"%d\00", align 1
@4 = private unnamed_addr constant [7 x i8] c"nprocs\00", align 1
@5 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@6 = private unnamed_addr constant [22 x i8] c"WARNING: Unknown key:\00", align 1
@7 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1
@8 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@9 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define i32 @main(i32 %0, i8** %1) {
.entry:
  %array_bound16 = alloca i32, align 4
  %array_bound12 = alloca i32, align 4
  %array_bound8 = alloca i32, align 4
  %array_bound = alloca i32, align 4
  %__1_k = alloca i32, align 4
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %__1_k1 = alloca i32, align 4
  %key = alloca i8*, align 8
  %2 = call i8* @_lfortran_malloc(i32 21)
[...]

PR #6411, opened 2025-02-25
enh: read comma separated string into an array

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6229


integration_tests/string_48.f90
```diff
@@ -0,0 +1,32 @@
...
+    integer, save, dimension(3) :: nprocs_int32 = [-1,-1,-1]
+    integer(8), save, dimension(4) :: nprocs_int64 = [-1,-1,-1,-1]
+    real, save, dimension(3) :: nprocs_real32 = [-1,-1,-1]
+    real(8), save, dimension(5) :: nprocs_real64 = [-1,-1,-1,-1,-1]
+    val1 = "133,134,135"
+    read(val1, *) nprocs_int32
+    print *, nprocs_int32
+    if (any(nprocs_int32 /= [133, 134, 135])) error stop
+
+    val2 = "133,134,135,136"
```
gxyd, line 18: the only reason to have `val1`, `val2` etc. is because there is another bug here, because of which the value of `val` after the first `read` is not what one expects, I'll open a separate issue for that.

gxyd:
> Excellent, good progress! Let's get the CI to run and see if this fixes the workaround.

It doesn't do that, I've an MRE extracted from POT3D which doesn't work with LFortran, but works with GFortran:
```f90
program main
    implicit none
    real(8), dimension(10) :: drratio=0.
    character(len=60) :: val
    integer :: i
    integer :: ierr
    integer :: n_values = 1
    val = "2.5"
    read(val, *, iostat=ierr) (drratio(i), i = 1, n_values)
end program main
```

I'll fix this.

PR #6423, opened 2025-02-26
ci: test `1e-16` precision for prima examples

Pranavchiku:
Okay, for the examples that did not match bit-to-bit with GFortran on Mac, there might be two cases

- We checking wrong values ( will be proved if CI passes now )
- GFortran gives different output on Mac and Linux ( will be proved if Linux fails )


Pranavchiku:
closing with reference to #6426

PR #6431, opened 2025-02-26
Handle memory leaks for `str_item`, `str_slice` and `printf`

assem2002:
I don't know the best approach to handle those leaks, I was just trying to free them to see if the leak spikes would drop, and it did.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -552,6 +552,7 @@ jobs:
             if [[ "${{matrix.os}}" == "ubuntu-latest" ]]; then
               cd fortran/
               test_name=test_uobyqa.f90 FC="$(pwd)/../../src/bin/lfortran" ./script.sh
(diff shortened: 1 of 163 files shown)


assem2002:
Yeah, the addressed places in this PR (where things leak)  are all related to the string redesign https://github.com/lfortran/lfortran/issues/2257. I think we should do the following ( introduce StringPhysicalType that acts like `string_view` as mentioned here https://github.com/lfortran/lfortran/issues/2257#issue-1868295228 + make the `simplifier` pass (`array_struct_temporary`) create temp for `str_slice` and `str_item`)

assem2002:
As a reminder for myself : 
- `str_cat` in C library leaks
- The following example leaks also :
```.f90

module name
  implicit none
  contains
  function ff() result(r)
    character(10000) :: r ! <<<<<<<<<<<<<< This memory won't get freed
  end function ff
  
end module name

program prog
  use name
  character(20) :: x
  integer :: i
  do i =0 , 1000000
    x = ff()
  end do

end program prog
```

PR #6432, opened 2025-02-26
fix : pass pointer to pointer in subroutine for intent inout or out mode

AnshMehta1:


AnshMehta1:
I will handle the failures shortly

AnshMehta1:
Currently , Intent Unspecified is not taken into consideration for passing Pointer to Pointer as a lot of things needs to be taken care of for that and also , in PRIMA , all intents are well specified and we don't see any intent unspecified.

AnshMehta1:
1. Currently , added integration test is getting semantic error in LFortran std = f23 but I tested the test using `gfortran -std=f2018` ( Not having latest version of gfortran for `-std=f2023` ) and their it was running fine . So , their might be some error in LFortran std = f23 .
2. Other error is regarding wasm backend , which I am not sure why we are getting this error because things are running well with llvm backend

AnshMehta1:
> If you can't reproduce it locally, you can:
> 
>     * Try to disable the WASM tests if they fail for this new test.
> 
>     * You can add `NO_STD_F23` and see if it fixes the issues.
> 
> 
> Get the CI to work.
> 
> Then at least we have the option to merge.

Done with changes , lets see if CI passes ( it should pass ) . If it fails then , will handle the failure tomorrow .

AnshMehta1:
> Excellent. Does this fix allow us to remove the workaround in PRIMA?

We need to check that . I will check it and update the status tomorrow .

PR #6439, opened 2025-02-27
Fix: Handle external symbol in nullify and update prima CI

### LINKED ISSUE #6440: Handle external procedure pointer in nullify
MRE:
```fortran
module temp 
contains
 subroutine OBJ()
 end subroutine 
end module

module ext 
    use temp
    procedure(OBJ), pointer :: calfun => null()
end module

program test
  use ext
  nullify(calfun) 
end program
```

LFortran Error:
```console 
Operand is null
  store void ()* null, <null operand!>, align 8
```

jinangshah21:
Fixes #6440 and remove workaround from PRIMA.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -516,8 +516,8 @@ jobs:
         run: |
             git clone https://github.com/Pranavchiku/prima.git
             cd prima
-            git checkout -t origin/lf-prima-11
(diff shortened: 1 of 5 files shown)


jinangshah21:
Here `nullify_03` and `nullify_04` tests were not working before but with this PR they also works.

Pranavchiku (approved):
Awesome! With this we have PRIMA working without any workarounds aligning bit-to-but with GFortran. Super!

PR #6450, opened 2025-02-28
fix : Handle AttrIntent declaration in different line

### LINKED ISSUE #5637: declaration attributes on different lines raises semantic error
## Description

```f90
program main
    implicit none
    integer :: x = 1
    call f(x)

    contains

    subroutine f(a)
        integer :: a ! type is declared here for 'a'
        intent(in) :: a ! intent is declared here for 'a'
    end subroutine
end program main

```

probably used in only legacy codebase to define attribute on separate line. The above Fortran program on compilation with LFortran, raises the below error:
```console
> lfortran test.f90
  --> test.f90:10:9
   |
10 |         intent(in) :: a ! intent is declared here for 'a'
   |         ^^^^^^^^^^^^^^^ 


```

while it's compilation with GFortran works without any issue.

AnshMehta1:
fix #5637 

MRE :- 

```.f90
program subroutines_19
    implicit none
    integer :: x = 1
    call f(x)

    contains

    subroutine f(a)
        integer :: a 
        intent(in) :: a 
        print * , a
        if (a /= 1) error stop
    end subroutine
end program
```

Intent updation in ASR 

```console
a:
    (Variable
        3
        a
        []
        In
        ()
        ()
        Default
        (Integer 4)
        ()
        Source
        Public
        Required
        .false.
        .false.
        .false.
    )
```

### DIFF
--- a/integration_tests/subroutines_18.f90
+++ b/integration_tests/subroutines_18.f90
@@ -1,4 +1,4 @@
-program main
+program subroutines_18
    procedure(OBJ), pointer :: proc_ptr
 
    call sub(proc_ptr)


AnshMehta1:
> This is fine to unblock us.
> 
> However this fix is not considering this case:
> 
> ```
> intent(in) :: a
> integer :: a
> ```
> 
> It will ignore the `intent(in)` line in this case. The proper fix is to create a placeholder for `a` in the symbol table, and then fill it with more info later, and give an error message if the type of `a` is not later declared. I opened up an issue for this at #6452.

I truely agree with your point . We need a robust implementation to handle attribute declaration in different lines as per your suggestion .

Issue #6445, opened 2025-02-28
Array expression resulting from slicing an array of structures is not recognized in intrinsic `sum`


difference-scheme:
Testcode:
```
program array_of_struct

   type :: struct
      real :: field = 5.0
   end type struct

   type(struct) :: array(3)
   
   write(*,*) sum(array(:)%field)

end program array_of_struct
```
LFortran's output:
```
semantic error: Argument to intrinsic `Sum` is expected to be an array, found: real
 --> array_of_struct.f90:9:15
  |
9 |    write(*,*) sum(array(:)%field)
  |               ^^^^^^^^^^^^^^^^^^^ 



```

kmr-srbh:
Thanks for the bug report @difference-scheme!

HarshitaKalani:
I think the issue is not with sum but the wau we handle struct.
```fortran
program array_of_struct

   type :: struct
      real :: field = 5.0
   end type struct

   type(struct) :: array(3)
   integer :: arr(3)
   
   print*, array(:)%field

end program array_of_struct
```
```console
$ gfortran a.f90 && ./a.out
   5.00000000       5.00000000       5.00000000    
$ lfortran a.f90
lfortran: /home/harshita/miniconda3/envs/lf/include/llvm/IR/Instructions.h:939: static llvm::GetElementPtrInst* llvm::GetElementPtrInst::Create(llvm::Type*, llvm::Value*, llvm::ArrayRef<llvm::Value*>, const llvm::Twine&, llvm::Instruction*): Assertion `PointeeType == cast<PointerType>(Ptr->getType()->getScalarType())->getElementType()' failed.
Traceback (most recent call last):
  Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2819, in main
    return main_app(argc, argv);
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2757, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, time_report, false,
[...]

PR #6471, opened 2025-03-03
Correct m_type of StructInstanceMember for ArraySection

AnshMehta1:
towards #6445 

Mre :- 
```.f90
program struct_type_02

   type :: struct
      real :: field = 5.0
   end type struct

   type(struct) :: array(3)
   print*, array(:)%field

end program
```

LFortran Output :- 
```console
5.00000000e+00    5.00000000e+00    5.00000000e+00
```


AnshMehta1:
Still original MRE of #6445 is not working due to some other issue . Will fix that in next PR.

PR #6475, opened 2025-03-04
Subroutine_from_function on fixed strings

assem2002:


assem2002:
The failures are due to how the `result_var` -the variable that will be passed to hold the result of the function- is getting initialized, it should be allocatable string (the same as how simplifier is handling arrays)

PR #6476, opened 2025-03-04
CI: update hash commit to test with POT3D

gxyd:
## Description

This removes a workaround or two from POT3D and now tests them with LFortran


gxyd:
The CI passes, I'll not wait for a review here, as it's not something major to review.

Issue #4189, opened 2024-06-08
Derived type initialisation issue


nikabot:
The following shouldn't work:
```fortran
program test

type t
   integer :: i
   integer :: j
end type t

type(t) :: t_ = t(i=13)

print *, t_

end program test
```

Both the arguments has to be passed to initialise.

See GFortran:
```console
$ gfortran examples/expr2.f90  && ./a.out
examples/expr2.f90:75:16:

   75 | type(t) :: t_ = t(i=13)
      |                1
Error: No initializer for component 'j' given in the structure constructor at (1)
```

kmr-srbh:
This is an easy fix. I am pushing the changes now.

kmr-srbh:
I worked on a fix for this. Turns out that the way we handle the arguments is not very clean and comparing the number of passed arguments to expected number of arguments does not work for all cases. When the derived type has a default value for all the arguments, we do not take this into account. 

Here is the MRE:
```fortran
program test

   type :: date
      integer :: year = 0
      integer :: month = 0
      integer :: day = 0
   end type

   type(date) :: empty_date = date()

   print *, empty_date

end program test
```

Due to these errors, I am not opening a PR for this. I thought the comparison would work, but it does not. I am sorry, unfortunately I would not be able to devote more time to investigate for a fix.

certik:
Right. One must take into account the default arguments. We can fix that after beta.

PR #6489, opened 2025-03-05
Improve derived type default initialization

xaerru:
Fix https://github.com/lfortran/lfortran/issues/4189

The first commit fixes `template_vector.f90` test. Without this commit, the second commit will give ASR error for `sz` not being specified. Declaring class improperly should raise an error after (https://github.com/lfortran/lfortran/pull/5922) is merged.

The second commit uncomments a block of code which was commented out because

> The following loop takes the default initial value expression
        and puts it into the constructor call. The issue with this approach
        is that one has to replace all the symbols present in the expression
        with external symbols so that we don't receive out of scope errors
        in ASR verify pass (with the help of a new visitor ReplaceWithExternalSymbolsInExpression).
        So for now its commented out and one can deal with the
        same in the backend itself which appears cleaner to me for now.

I might be wrong but I don't think its a problem anymore. The default initializer values should be compile time constants.

In the test references many `StructConstructor` nodes are replaced with `StructConstant`.

### DIFF
--- a/integration_tests/template_vector.f90
+++ b/integration_tests/template_vector.f90
@@ -59,10 +59,9 @@ contains
 
     subroutine main()
         instantiate vector_t(integer), only: IntVector => Vector
-        class(IntVector) :: v
-        integer :: i
-        v = IntVector()
-        ! call v%push_back(v, 1)
+        type(IntVector) :: v
+        call v%push_back(10)
+        if (v%elements(1) /= 10) error stop
     end subroutine
 
 end module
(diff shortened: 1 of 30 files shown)


xaerru:
> `fpm` is giving problems in ci. The comment was right.

Earlier `m_symbolic_value` was being used as an init expr. Now `m_value` is being used. So no need to make `ReplaceWithExternalSymbolsInExpression` as mentioned in the comment.

After using `m_value`, fpm was giving a semantic error about constructor arguments. Overridden constructor by interface with same name as the type was not being called, the normal constructor was being called. Fixed it in a separate commit.

PR #6493, opened 2025-03-06
fix: update dependencies of DerviedType appropriately

gxyd:
## Description

When it's an allocatable / Array / pointer or a combination of one of these types, update the dependencies.

This *might* help us with issue: https://github.com/lfortran/lfortran/issues/5655

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1965,7 +1965,7 @@ public:
             }
             char* aggregate_type_name = nullptr;
             if (item.first != "~abstract_type") {
-                ASR::ttype_t* var_type = ASRUtils::type_get_past_pointer(ASRUtils::symbol_type(item.second));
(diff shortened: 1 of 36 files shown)


gxyd:
For the failing CI test, below is the MRE extracted from *SNAP* third party repository:
```f90
module p
    implicit none

    TYPE cell_id_type
        INTEGER(4) :: ic, j, k
    END TYPE cell_id_type

    TYPE diag_type
        INTEGER(4) :: len
        TYPE(cell_id_type), ALLOCATABLE, DIMENSION(:) :: cell_id
    END TYPE diag_type

    TYPE(diag_type), ALLOCATABLE, DIMENSION(:) :: diag
end module p

MODULE mkba_sweep_module
    USE p, ONLY: diag
    IMPLICIT NONE

    CONTAINS
    SUBROUTINE mkba_sweep ( )
      INTEGER(4) :: n
        n = diag(1)%len
    END SUBROUTINE mkba_sweep

END MODULE mkba_sweep_module

```

PR #6501, opened 2025-03-06
Fix : Allocate character with integer kind = 8

assem2002:
fixes https://github.com/lfortran/lfortran/issues/6500


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -129,6 +112,25 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                }, false);
+        fn = llvm::Function::Create(function_type,
+                llvm::Function::ExternalLinkage, func_name, module);
+    }
+    std::vector<llvm::Value*> args = {llvm_utils->convert_kind(arg_size,
+                                            llvm::Type::getInt64Ty(context)),
+                                    arg_string};
+    builder.CreateCall(fn, args);
+}
+
```
assem2002, line 133: Just moved the function inside the class to have access to `llvm_utils` object

Issue #1693, opened 2023-05-24
Keyword argument in type-bound procedure leads to LFortran memory allocation failure


difference-scheme:
This bug is truly strange. Upon hitting it during compilation, LFortran ate up all the memory on my system, and forced me to reboot ...
I had to turn off swap in order to investigate it, without freezing the computer, but couldn't reduce it any further than this:

```
module malloc_failure
   
   implicit none
   
   type, public, abstract :: AbsType
   contains
      procedure(callee), deferred, nopass :: callee
   end type AbsType
   
   abstract interface
      subroutine callee(mode)
         integer, intent(in)  :: mode
      end subroutine callee
   end interface
      
   type :: SomeClass
      private
      class(AbsType), allocatable :: object
   contains
      procedure :: caller
   end type SomeClass
   
   interface SomeClass
      procedure :: constructor
   end interface SomeClass
   
contains

   function constructor(object) result(self)
      type(SomeClass)            :: self
      class(AbsType), intent(in) :: object
      self%object = object
   end function constructor
   
   subroutine caller(self)
      class(SomeClass), intent(in)  :: self
      call self%object%callee(mode=0)
   end subroutine caller

end module malloc_failure
```
This compiles with ifort. It seems that LFortran's trouble with it is connected to the use of a keyword argument in the call of the `callee` procedure.

certik:
Thanks! It's some kind of an infinite loop or something. We'll investigate.

HarshitaKalani:
There's this for loop at line 9582 in ast_common_visitor which is going infinite times and hence this issue,
```
for (size_t i = 0; i < fn_n_args - offset - type_bound; i++) {
            ASR::call_arg_t call_arg;
            call_arg.loc = loc;
            call_arg.m_value = nullptr;
            args.push_back(al, call_arg);
        }
```
I'm investigating the reason here.

difference-scheme:
@HarshitaKalani, @certik Thanks for working on this.

PR #6507, opened 2025-03-07
fix: handle keyword arguments correctly for class procedure

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/1693#issuecomment-2703157527


HarshitaKalani:
This solution is fine but not sufficient. We should only subtract n_member if the function definition (fn_def) actually includes those members as arguments. So in this case, the correct solution should be to check if callee is NOPASS, then the function should correctly not subtract type_bound.

HarshitaKalani:
I'll merge this to unblock on that issue. I'm looking for the correct solution meanwhile.

Issue #6479, opened 2025-03-04
Dummy argument not recognized in subroutine with inner function when using implicit interface and implicit typing


difference-scheme:
The following test code compiles without issues using either gfortran, ifx, or flang-20:
```
subroutine outer(x)

   implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)

contains
   
   function inner()
      implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)
      inner = 1
   end function inner

end subroutine outer
```
When compiled with `lfortran -c --implicit-typing --implicit-interface outer.f90` the result is as follows:
```
semantic error: Dummy argument 'x' not defined
  --> outer.f90:1:1 - 12:20
   |
 1 |    subroutine outer(x)
   |    ^^^^^^^^^^^^^^^^^^^...
...
   |
12 |    end subroutine outer
   | ...^^^^^^^^^^^^^^^^^^^^ 

```
Leaving out the `--implicit-interface` compiler option, one encounters Issue #6453. So these two are possibly related.

AnshMehta1:
I will work on it .

PR #6509, opened 2025-03-07
fix : Create copy for implicit dictionary before nesting

AnshMehta1:
fix : #6479

MRE :-  
```.f90
subroutine outer(x)

   implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)

contains
   
   function inner()
      implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)
      inner = 1
   end function inner

end subroutine outer
```

Corrected ASR for x :- 
```.f90
x:
    (Variable
        2
        x
        []
        Unspecified
        ()
        ()
        Default
        (Real 8)
        ()
        Source
        Public
        Required
        .false.
        .false.
        .false.
    )

```

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1102,13 +1102,9 @@ public:
         for (size_t i=0; i<x.n_contains; i++) {
             bool current_storage_save = default_storage_save;
             default_storage_save = false;
-            if ( compiler_options.implicit_typing ) {
-                std::map<std::string, ASR::ttype_t*> implicit_dictionary_copy = implicit_dictionary;
-                visit_program_unit(*x.m_contains[i]);
-                implicit_dictionary = implicit_dictionary_copy;
-            } else {
-                visit_program_unit(*x.m_contains[i]);
-            }
+            std::map<std::string, ASR::ttype_t*> implicit_dictionary_copy = implicit_dictionary;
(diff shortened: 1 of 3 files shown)


HarshitaKalani (commented):
The failing check seem to be unrelated. You can rerun that after all the checks run.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1102,7 +1102,13 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
             bool current_storage_save = default_storage_save;
             default_storage_save = false;
-            visit_program_unit(*x.m_contains[i]);
+            if ( compiler_options.implicit_typing ) {
+                std::map<std::string, ASR::ttype_t*> implicit_dictionary_copy = implicit_dictionary;
+                visit_program_unit(*x.m_contains[i]);
+                implicit_dictionary = implicit_dictionary_copy;
+            } else {
+                visit_program_unit(*x.m_contains[i]);
+            }
```
Pranavchiku, line 1111:
```suggestion
             std::map<std::string, ASR::ttype_t*> implicit_dictionary_copy = implicit_dictionary;
             visit_program_unit(*x.m_contains[i]);
             implicit_dictionary = implicit_dictionary_copy;
```

We can keep it by default, nice catch!

Pranavchiku (approved):
Looks good, check if the simplification works, then feel free to merge this and do in subsequent PR.

PR #6514, opened 2025-03-07
DON'T MERGE CI: test workaround removal from POT3D's branch

gxyd:
## Description

This PR isn't intended to be merged at all, just want to check which CI tests fail and on which platform(s), when the final and only workaround is removal from POT3D


gxyd:
I think the CI will pass now, and if it really does, I think we'll have it.

gxyd:
I'm merging this, and wouldn't wait for a review here, as it isn't necessarily needed.

PR #6515, opened 2025-03-07
fix: implementation of `ieee_is_finite`

Pranavchiku:
With this PR and https://github.com/lfortran/lfortran/pull/6511 we get `numerical-methods-fortran` bit-to-bit aligned without any workarounds!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -527,7 +527,7 @@ jobs:
             ./test_ode.exe
             ./test_probability_distribution.exe
             ./test_sde.exe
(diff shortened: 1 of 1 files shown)


Pranavchiku:
It does not matter, the primary suspicion of mine is using ordered comparison in case of less than makes it strict for handling NaN whereas `/=` uses unordered comparison. When we switch from `<` to `/=` it uses ordered and returns false for `is_finite`. The operation `==` also uses unordered comparison so it gives `true` for `NaN`. Apart from this I see no "logical" difference / preference in using `/=` or `<`.

PR #6518, opened 2025-03-07
CMake: enable standalone parser and C libraries

certik:
This tweaks the cmake to be able to build just the parser and/or C wrapper as standalone libraries.


Pranavchiku (approved):
Tested locally, works fine, no evident delays

PR #6546, opened 2025-03-09
enh: handling allocation of nested struct instance members

Pranavchiku:
Towards #5655.


Pranavchiku:
To get #5655 fully working we need to fix https://github.com/lfortran/lfortran/issues/6547 as well. It is related to printing it.

Issue #5801, opened 2024-12-23
bug : array of characters doesn't do actual stringcopying, It just stores the pointers


assem2002:
MRE 
``` Fortran
program progg
    character(20) :: i(2)
    character(20) :: str
    str = "LFortran"

    i(1)  = str
    print *, i(1)

    str = "whatever"
    print *, str
    print* ,i(1)
end program progg
```

```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt6.f90 
LFortran            
whatever            
whatever            
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt6.f90 && ./a.out 
 LFortran            
 whatever            
 LFortran            

```

assem2002:
related to https://github.com/lfortran/lfortran/issues/5575 https://github.com/lfortran/lfortran/issues/2257

assem2002:
another MRE
```.f90
  program p
    type tt
      character(10) :: str
    end type tt

    character(10) :: fixed_s
    type(tt) :: inst
    
    fixed_s = "helloworld"
    inst%str = fixed_s !BUG : both variables will share same memory
    print *, inst%str
    print *, fixed_s

    fixed_s = "ff"

    print *, inst%str
    print *, fixed_s
  end program p
```
console
```
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt4.f90 && ./a.out 
 helloworld
 helloworld
 helloworld
 ff        
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt4.f90 
helloworld
helloworld
ff        
ff        
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ 

```

assem2002:
This works now.

assem2002:
I'll reopen it. We need to add an integration test for it before closing.

PR #6572, opened 2025-03-11
Fix : Always do string copy with string assignment expression

### LINKED ISSUE #6560: allocatable string passed as argument is memory corrupted
``MRE
module m
    contains
    subroutine strip(x, y)
      implicit none
  
      character(len=*), intent(in) :: x
      character(len=len(x)), intent(out) :: y
      y = trim(x)
  end subroutine strip
  
  
  end module m
  program p
  use m
  character(:), allocatable :: ss
  allocate(character(5) :: s)
  call strip("hello",s)
  print *,"|", s, "|"
  if(s /= "hello") error stop
  end program p
```
```console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt4.f90 && ./a.out 
 |hello|
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranMainStream/lfortran/src/bin/lfortran tt4.f90 
|             |
ERROR STOP
```
This is needed for #6475

assem2002:
Fix #6560


assem2002:
This needs to get resolved first #5801 (some tests should pass)

Issue #6570, opened 2025-03-10
v0.48.0 fails to build with LLVM 20.1.0


junghans:
From https://koji.fedoraproject.org/koji/taskinfo?taskID=130096800:
```
[ 71%] Building CXX object src/libasr/CMakeFiles/asr.dir/codegen/llvm_utils.cpp.o
cd /builddir/build/BUILD/lfortran-0.48.0-build/lfortran-0.48.0/redhat-linux-build/src/libasr && /usr/bin/g++ -DHAVE_TARGET_X86=1 -DHAVE_WHEREAMI=1 -DHAVE_ZLIB=1 -DLCOMPILERS_FAST_ALLOC=1 -I/builddir/build/BUILD/lfortran-0.48.0-build/lfortran-0.48.0/redhat-linux-build/src/libasr/.. -I/builddir/build/BUILD/lfortran-0.48.0-build/lfortran-0.48.0/src/libasr/.. -isystem /usr/lib64/llvm20/include -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -mtls-dialect=gnu2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -DNDEBUG -std=gnu++17 -fPIC -fno-rtti -Wno-deprecated-declarations -MD -MT src/libasr/CMakeFiles/asr.dir/codegen/llvm_utils.cpp.o -MF CMakeFiles/asr.dir/codegen/llvm_utils.cpp.o.d -o CMakeFiles/asr.dir/codegen/llvm_utils.cpp.o -c /builddir/build/BUILD/lfortran-0.48.0-build/lfortran-0.48.0/src/libasr/codegen/llvm_utils.cpp
[...]

PR #6573, opened 2025-03-11
CI: Enable LLVM 20 test

certik:
Towards #6570.


assem2002:
That's great so far. We probably got minor problem with llvm20

PR #6579, opened 2025-03-11
fix: no semantic error for class procedure withour "defered"

HarshilShah1804:
Apply suggestion from https://github.com/lfortran/lfortran/pull/5399#discussion_r1844942969


HarshilShah1804:
I believe the failing tests are unrelated to my PR as they are passed in other CI checks. Moreover, I did clean build on my system by reverting my changes also and it still gives same result. So, I believe it is due to something else.

HarshilShah1804:
I closed this PR as just commenting out that lines does not solve this issue completely. The test mentioned in this PR doesn't accurately check it and other check that was there in Gaurav's PR cause some code generation error in llvm. This has to be fixed in the llvm. The initial error that is identified is that `unused_functions.cpp` pass removes some instances of the interface procedures which is causing some code generation error if we skip pass unused_functions. I will investigate this further how to fix it in llvm and then reopen the complete PR.

Issue #6309, opened 2025-02-14
Complicated initialization bug


harperjf:
This program finds the smallest integer of the form ```x**3 + y**3``` in two different ways. Lfortran can't compile it.  
```
program ramanujan_single
    implicit none
    integer, parameter   :: max_ij  = 13, max_sum = 2 * max_ij ** 3
    integer              :: i, j, k

    integer, parameter :: s2cube(*) = &
	   pack( [ (k, k = 1,max_sum)], &
                [ ( count( [((i**3+j**3, i = 1,j), &
                  j = 1,max_ij)] == k ) > 1, &
                  k = 1,max_sum )] )
    print "(I0)", s2cube
  end program ramanujan_single
```
Output from lfortran 0.45.0 --std=f23:
```
/home/john/bin/lf45: line 3:  7357 Segmentation fault      (core dumped) ~/lfortran-0.45.0/inst/bin/lfortran "$@"
```
Correct output from gfortran and ifx:
```
1729
```

kmr-srbh:
Thanks for reporting this @harperjf! Here is another MRE that reproduces this bug:

```fortran
program ramanujan
   implicit none
   integer :: k

   integer, parameter :: s2cube(*) = pack( [ (k, k = 1,2) ], [ (.true., k = 1, 2) ] )
   print "(I0)", s2cube
end program ramanujan
```

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2769, in main
    return main_app(argc, argv);
  File "/home/saurabh-kumar/Projects/System/lfortran/src/bin/lfortran.cpp", line 2707, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, false,
[...]

PR #6582, opened 2025-03-11
enh: support initial bool support in implied do loop values visitor

jinangshah21:
Towards #6309
Fixes MRE at https://github.com/lfortran/lfortran/issues/6309#issuecomment-2659025725


Pranavchiku (approved):
Good, thank you!

Issue #6589, opened 2025-03-12
add a warning that `Volatile` attribute is not supported yet


Pranavchiku:
https://github.com/lfortran/lfortran/pull/6586#discussion_r1991794377

islml:
Hello, can I work on this issue?

PR #6606, opened 2025-03-13
Add a warning that `Volatile` attribute is not supported yet

islml:
fixes: #6589

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3372,7 +3372,11 @@ public:
                             } else if(sa->m_attr == AST::simple_attributeType
                                 ::AttrNoPass) {
                             } else if (sa->m_attr == AST::simple_attributeType::AttrVolatile) {
-                                // TODO: Implement volatile attribute
+                                diag.add(Diagnostic(
+                                    "Volatile attribute is not supported yet",
+                                    Level::Warning, Stage::Semantic, {
+                                        Label("",{x.base.base.loc})
+                                }));
                             } else {
                                 diag.add(Diagnostic(
                                     "Attribute type not implemented yet " + std::to_string(sa->m_attr),


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3372,7 +3372,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             } else if(sa->m_attr == AST::simple_attributeType
                                 ::AttrNoPass) {
                             } else if (sa->m_attr == AST::simple_attributeType::AttrVolatile) {
-                                // TODO: Implement volatile attribute
```
gxyd, line 3375: We can continue to have the *TODO* note here, as it still isn't implemented yet.

Also, it would be good to test this as well, maybe in `tests/errors/continue_compilation_01.f90` or other relevant file?

gxyd (commented):
We need a test for this warning.

Pranavchiku:
@islml thanks for the PR! Let us also add test, you can take an example using `volatile` and add it inside `tests/errors/continue_compilation_01.f90`, then run the following commands:

```console
rm -rf ./tests/reference/*
./run_tests.py -u
```

Whatever changes are introduced, commit those.

islml:
Done @Pranavchiku

tests/reference/run-array14-c9f6829.stdout
```diff
@@ -1,2 +0,0 @@
-5.00000000e+00    -1.00000000e+00    3.00000000e+00    4.00000000e+00    2.00000000e+00
```
Pranavchiku, line 1: These should not be there. Do a clean build and again update test references, here is how to do:

```
git restore --staged .
git restore .
git clean -dfx
./build1.sh
rm -rf ./tests/reference/*
./run_tests.py -u
```

Make sure the only change is the one pushed in the PR.

islml:
@gxyd can you review the changes please

islml:
@certik done

Issue #6596, opened 2025-03-12
Complicated parameter initialization not recognized


jhykes:
```
$ cat test.f90
module test_mod

implicit none

logical, parameter :: param=ichar(transfer(1,'a')) == 0

end module test_mod


$ lfortran -c test.f90
semantic error: Initialization of `param` must reduce to a compile time constant.
 --> test.f90:5:1
  |
5 | logical, parameter :: param=ichar(transfer(1,'a')) == 0
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This compiles okay with gfortran, ifort, nagfor.

Possibly related to https://github.com/lfortran/lfortran/issues/1484

jinangshah21:
I will try to fix this

HarshitaKalani:
@jinangshah21 let me know if you're working on it. I know how to fix it.

PR #6607, opened 2025-03-13
fix: support character type as mold in bitcast

jinangshah21:
Towards #6596

### DIFF
--- a/integration_tests/transfer_01.f90
+++ b/integration_tests/transfer_01.f90
@@ -2,6 +2,7 @@ program transfer_01
     implicit none
     integer :: i, integer_result, y(3, 2)
     real :: r, result, x(5, 4)
+    character(len=*), parameter :: chr = 'A'
     r = 2.987654
     i = 123456
     x = 3.19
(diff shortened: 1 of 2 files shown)


integration_tests/transfer_01.f90
```diff
@@ -22,4 +23,9 @@ program transfer_01
     result = transfer(y(1,1), x(5,2))
     print *, result
     if (abs(result - 5.60519386E-45) > 1e-6) error stop
+
+    print *, transfer(65, chr)
+    if (transfer(65, chr) /= 'A') error stop
```
gxyd, line 28: This test may not necessarily ensure that `transfer(65, chr)` reduced to a compile time constant.

What we probably need is, something like:
```f90
character(len=*), parameter :: chr = transfer(65, 'A')

print *, chr
if (chr /= 'A') error stop
```

gxyd (approved):
If the intention of this PR is to evaluate BitCast at runtime, then this does it, but I guess the linked issue is about compile-time evaluation.

integration_tests/transfer_01.f90
```diff
@@ -2,6 +2,7 @@ program transfer_01
     implicit none
     integer :: i, integer_result, y(3, 2)
     real :: r, result, x(5, 4)
+    character(len=*), parameter :: chr = 'A'
```
gxyd, line 5: Can we also have a test case where `chr` isn't a parameter value?

gxyd:
I've approved the PR, but it would be ideal to have one more test case for it.

jinangshah21:
Oh, we don't have compile time evaluation for `Bitcast` at all. I will implement it.

Issue #6604, opened 2025-03-13
Handle different kinds of arguments in `shifta`


jinangshah21:
MRE: 
```fortran
program test
    integer(8) :: x = 32
    integer :: y = 2
    x = shifta(x, y)
    print *, x
end program test
```

LFortran Error :
```console
code generation error: asr_to_llvm: module failed verification. Error:
Both operands to a binary operator are not of the same type!
  %4 = ashr i64 %2, i32 %3
```

PR #6614, opened 2025-03-13
Fix handling of different kinds of arguments in shifta intrinsic (#6604)

samimshoaib01:


samimshoaib01:
If left_val is i32 and right_val is i64, we cast right_val to i32 using CreateTrunc.

If left_val is i64 and right_val is i32, we cast right_val to i64 using CreateSExt.

jinangshah21:
You Have to handle it in `ASR` (.i.e. got to `create_Shifta` in `ast_common_visitor` and handle this logic there)

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19,6 +19,7 @@
 #include <map>
 #include <limits>
 
+
```
jinangshah21, line 22: Remove these extra newline.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6769,6 +6769,7 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
                 tmp = builder->CreateAShr(left_val, right_val);
                 break;
             }
+            
         }
```
jinangshah21, line 6773: Same here

jinangshah21 (approved):

jinangshah21:
Go to `CmakeLists.txt`, search `shifta` and add test in that file instead of making new test.

Issue #6405, opened 2025-02-25
fast optimizations are run twice when compiling and running with LLVM backend


gxyd:
I applied the below diff to the LFortran source code:
```diff
diff --git a/src/bin/lfortran.cpp b/src/bin/lfortran.cpp
index c7b09d9a8..3c81a9086 100644
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1078,6 +1078,7 @@ int compile_src_to_object_file(const std::string &infile,
     }
 
     if (compiler_options.po.fast) {
+        std:: cout << "Applying optimizations for second time\n";
         t1 = std::chrono::high_resolution_clock::now();
         e.opt(*m->m_m);
         t2 = std::chrono::high_resolution_clock::now();
diff --git a/src/lfortran/fortran_evaluator.cpp b/src/lfortran/fortran_evaluator.cpp
index fdd48d0da..5305ef683 100644
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -388,6 +388,7 @@ Result<std::unique_ptr<LLVMModule>> FortranEvaluator::get_llvm3(
     }
 
     if (compiler_options.po.fast) {
+        std::cout << "Applying optimizations for first time\n";
         e->opt(*m->m_m);
     }
 
```

and then re-building LFortran and for the below Fortran program:
```f90
program main
    implicit none
    print *, "Hello world"
end program main

```

and running it with LFortran with `--fast` flag, I get the below output:
```console
> lfortran test.f90 --fast
Applying optimizations for first time
Applying optimizations for second time
Hello world
```

which indicates that we apply the fast optimizations twice. Probably a bug

PR #6628, opened 2025-03-15
fix: don't apply LLVM optimizations twice

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6405


src/lfortran/fortran_evaluator.cpp
```diff
@@ -382,16 +389,11 @@ Result<std::unique_ptr<LLVMModule>> FortranEvaluator::get_llvm3(
...
     } else {
         LCOMPILERS_ASSERT(diagnostics.has_error())
         return res.error;
     }
-
-    if (compiler_options.po.fast) {
-        e->opt(*m->m_m);
-    }
-
-    return m;
```
Pranavchiku, line 394: We can't do this because there are places where we simply call `get_llvm3` without going through `get_llvm2`, ex: `compile_src_to_object_file` in `lfortran.cpp`

Pranavchiku (changes requested):
left a comment above

src/lfortran/fortran_evaluator.cpp
gxyd: There are three places where `get_llvm3` is used, i.e.
1. `compile_src_to_object_file` (in lfortran.cpp), which already applies optimizations, hence applied the LLVM optimizations twice in current *main*
2. `evaluate` (in fortran_evaluator.cpp), where LLVM optimizations are applied explicitly in that function itself
3. `get_llvm2` (in fortran_evaluator.cpp), same as `evaluate`, where the LLVM optimizations are applied explicitly now.

gxyd: I couldn't really understand the concern here.

src/bin/lfortran.cpp
```diff
@@ -1204,13 +1204,7 @@ int compile_src_to_object_file(const std::string &infile,
...
     }
 
-    if (compiler_options.po.fast) {
-        t1 = std::chrono::high_resolution_clock::now();
-        e.opt(*m->m_m);
-        t2 = std::chrono::high_resolution_clock::now();
-        time_opt = std::chrono::duration_cast<std::chrono::microseconds>(t2 - t1).count();
-    }
-
+    // time_opt evaluation needs to be done
```
gxyd, line 1207: Will remove this line, but before I do, I'll wait for the CI checks to pass/fail.

src/lfortran/fortran_evaluator.cpp
gxyd: I've attempted to address this comment in the latest few commits.

Pranavchiku:
I’ll review and merge tomorrow morning.

Pranavchiku (approved):

PR #6634, opened 2025-03-16
Error message for wrong number of arguments to intrinsic functions

swamishiju:
#6918


HiteshGhanchi:
does it work ? if no argument i spassed in present?

swamishiju:
@certik  added a couple tests

swamishiju:
@certik  i found some bugs in my fix ill get back when its sorted out

Issue #6198, opened 2025-02-02
enh: raise proper error when number of arguments to intrinsic `present` is not 1


YBansal95:
```
program main
    integer :: a, b
    print *, present(a, b)
end program main
```

```
yashwini@yashwini:~/Desktop/vscode/lfortran$ ./src/bin/lfortran tmp.f90
semantic error: No matching signature found for intrinsic present
 --> tmp.f90:3:14
  |
3 |     print *, present(a, b)
  |              ^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
yashwini@yashwini:~/Desktop/vscode/lfortran$ gfortran tmp.f90
tmp.f90:3:12:

    3 |     print *, present(a, b)
      |            1
Error: Too many arguments in call to ‘present’ at (1)
```

And similarly, for:
```
program main
    integer :: a, b
    print *, present()
end program main
```

```
yashwini@yashwini:~/Desktop/vscode/lfortran$ ./src/bin/lfortran tmp.f90
semantic error: No matching signature found for intrinsic present
 --> tmp.f90:3:14
  |
3 |     print *, present()
  |              ^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
yashwini@yashwini:~/Desktop/vscode/lfortran$ gfortran tmp.f90
tmp.f90:3:12:

    3 |     print *, present()
      |            1
Error: Missing actual argument ‘a’ in call to ‘present’ at (1)
```

swamishiju:
I did get something working with a minor change. It seems to be passing integration tests but im not sure if its a hack

[Initial Code](https://github.com/lfortran/lfortran/blob/1fe60e2b0e80f9596b8e645b1afdd89ccd496680/src/lfortran/semantics/ast_common_visitor.h#L7364-L7368)

to 

```cpp
bool signature_matched = false;
signature_matched = handle_intrinsic_node_args(x, args, signature.kwarg_names, signature.positional_args, signature.max_args, var_name, true);
```

So the errors are generated in handle_intrinsic_node only

HiteshGhanchi:
@swamishiju so did you raise the pr with your changes and is it solved ?

HiteshGhanchi:
`semantic error: Incorrect number of arguments passed to the 'present' intrinsic. It accepts exactly 1 arguments.
[...]

PR #6647, opened 2025-03-17
Fix: Proper error messages for incorrect number of arguments to intrinsic functions

swamishiju:
#6198 

@certik closed to old pr by accident while cleaning


gxyd:
It always helps to link the older PR, I'm doing this time. 

Link to the older PR: https://github.com/lfortran/lfortran/pull/6634

tests/errors/continue_compilation_1.f90
```diff
@@ -284,5 +284,11 @@ program continue_compilation_1
     ! c1 is Character
     print *, c1%mymember
 
+    integer:: a_,b_
```
gxyd, line 287: We already have an integer declared in the program here: https://github.com/lfortran/lfortran/blob/b412b37b7c2cb6b3ebfc25bf5486eec965185e6f/tests/errors/continue_compilation_1.f90#L69, i.e. `x`, you can define one if another integer isn't defined or pass `x` again to `present` below, i.e. `present(x, x)`.

The reason being, we shouldn't be declaring variables after statements in Fortran anyway.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6368,11 +6368,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         Vec<ASR::expr_t*>& args, std::vector<std::string>& kwarg_names,
         size_t min_args, size_t max_args, const std::string& intrinsic_name,
         bool raise_error=true) {
+        // raise_error is used only when the error is not related to number of arguments
         size_t total_args = x.n_args + x.n_keywords;
         if( !(total_args <= max_args && total_args >= min_args) ) {
-            if( !raise_error ) {
-                return false;
-            }
```
gxyd, line 6375: Do you think this is the right fix or the one below?
```diff
diff --git a/src/lfortran/semantics/ast_common_visitor.h b/src/lfortran/semantics/ast_common_visitor.h
index 8a4f5bb42..575e73504 100644
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -7605,7 +7605,7 @@ public:
                 signature_matched = handle_intrinsic_node_args(
                     x, args, signature.kwarg_names,
                     signature.positional_args, signature.max_args,
-                    var_name, false);
+                    var_name, true);
                 check_specific_type_intrinsics(specific_var_name, args, x.base.base.loc);
                 if( !signature_matched ) {
                     diag.add(Diagnostic("No matching signature found for intrinsic " + var_name,
```

gxyd: It would be great if you can help understand with an Fortran program example, for which fix is the correct one.

swamishiju: I think a found a bug with my first fix it was something to do with a program like this I believe it was late at night so I can't be entirely sure

```f90
program main
    exit
end program main
```

I remember this exiting with compiler errors instead of the ones lfortran prints i'll try to recreate and get back

tests/errors/continue_compilation_1.f90
swamishiju: Look good now?

src/lfortran/semantics/ast_common_visitor.h
swamishiju: I can't seem to recreate it as of now both seem to work about the same

swamishiju: But I feel like the second fix is better in case the original author had a case in mind and there is a similar thing done in  https://github.com/lfortran/lfortran/blob/b3300c8f59d3c46d69d9d038fcd3c0249d7744f2/src/lfortran/semantics/ast_common_visitor.h#L6404-L6416

gxyd: Yes, I agree.

swamishiju: So we can go ahead with merge?

Issue #6349, opened 2025-02-18
Unimplemented ```sign``` specifier in ```open``` statement


harperjf:
Having found that lfortran has not implemented the edit descriptor ```SP``` I tried the other way to get the same result in this program:
```
  open(6,sign='plus')
  print '(F5.1)', 1.0 
end program
```
AMD flang, ifx, ifort and gfortran all printed this:
```
 +1.0
```
Lfortran 0.45.0 --std=f23 gave this at compile time:
```
semantic error: Invalid argument `sign` supplied
 --> gf51418b.f90:1:3
  |
1 |   open(6,sign='plus')
  |   ^^^^^^^^^^^^^^^^^^^ 
```

assem2002:
Working on it..

PR #6649, opened 2025-03-17
feat : Introduce `SP` specifier

assem2002:
towards #6349
- Parse `sp` specifier from string format
- Start add `+` (if needed) when the specifier is activated
- Add the plus sign with `I`, `F`, `D`, `E` specifiers (number related specifiers)
- Enhance minor problems in the formatting functions.
- Add reference test to show the `SP` specifier in action.


gxyd:
I initially thought, this PR is about "SP" as in "single precision", then I realized that this PR is about "SP" format specifier. 😄

gxyd (approved):
The changes in this PR look good to me, thanks for the PR @assem2002 . Great work

PR #6658, opened 2025-03-18
fix: ignore case for `c_null_ptr` in Frontend

### LINKED ISSUE #6650: lfortran will not initialize a c_ptr to C_NULL_PTR
I ran into an issue in which lfortran refused to compile something with `type(c_ptr) :: ptr = C_NULL_PTR`.  The initial case was within a derived type, but even as a "raw" object this fails.  Simple reproducer below.

```
% cat simple.f90 
program test
   use, intrinsic :: iso_c_binding
   implicit none
   private
   public :: test_obj

   type :: test_obj
      type(c_ptr) :: ptr = C_NULL_PTR !Fails on this line; does not want to initialize
      ! type(c_ptr) :: ptr ! Without the initialization this will compile
   end type test_obj

   type(test_obj) :: instance

   !type(c_ptr) :: myptr = C_NULL_PTR ! This line will also fail to compile, so it is not due to being in a derived type
end program


% lfortran simple.f90 
semantic error: Named initialization not supported with: C_NULL_PTR
 --> simple.f90:8:7
  |
8 |       type(c_ptr) :: ptr = C_NULL_PTR !Fails on this line; does not want to initialize
  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


% lfortran --version
LFortran version: 0.48.0
Platform: macOS ARM
LLVM: 19.1.1
Default target: arm64-apple-darwin24.3.0

```

jinangshah21:
Fixes #6650

### DIFF
--- a/integration_tests/c_ptr_02.f90
+++ b/integration_tests/c_ptr_02.f90
@@ -3,8 +3,17 @@ program test_c_ptr_02
    implicit none
    complex, target :: c_arr(2)
    real, pointer :: r_ptr(:)
+
+   type :: test_obj
+      type(c_ptr) :: ptr = c_NULL_ptr
+   end type test_obj
+   type(test_obj) :: instance
+   type(c_ptr) :: temp_ptr
+   temp_ptr = instance%ptr
+
    c_arr = (1, 2)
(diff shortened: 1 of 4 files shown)


Pranavchiku:
Rebase and merge this PR.

PR #6671, opened 2025-03-18
fix: handle mask without dimension in minloc/maxloc

### LINKED ISSUE #6664: bug: incorrect error for minloc with mask argument
```fortran
program main
    print*, minloc([1, 2], .true.)  
end program
```
```console
$ lfortran b.f90
semantic error: `dim` should be a scalar integer type
 --> b.f90:2:13
  |
2 |     print*, minloc([1, 2], .true.)  
  |             ^^^^^^^^^^^^^^^^^^^^^^ 


$ gfortran b.f90 && ./a.out
           1
```

HarshilShah1804:
Fixes #6664


HarshitaKalani (commented):
I'm concerned about the following case. With this PR, we make the following work which shouldn't be the case. The following program should throw an error.
@HarshilShah1804 can you think of a more robust approach?
```fortran
program main
  print*, minloc([1, 2], dim = .true.)  
end program
```
```console
$ lfortran a.f90
1
$ gfortran a.f90 && ./a.out
a.f90:14:31:

   14 |   print*, minloc([1, 2], dim = .true.)
Error: ‘dim’ argument of ‘minloc’ intrinsic at (1) must be INTEGER
```

HarshilShah1804:
Oh right. I'll change the approach of working here.

HarshilShah1804:
@HarshitaKalani I have changed the assignments in `ast_common_visitors.h` by matching the conditions. Please review and let me know if any changes are required.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6359,6 +6359,14 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
             }
         }
+        if ((intrinsic_name == "minloc" || intrinsic_name == "maxloc") && x.n_args == 2) {
+            this->visit_expr(*x.m_args[1].m_end);
+            ASR::expr_t* arg = ASRUtils::EXPR(tmp);
+            if (ASRUtils::is_logical(*ASRUtils::expr_type(arg))) {
+                args.p[2] = arg;
+                args.p[1] = nullptr;
+            }
+        }
```
gxyd, line 6369: I think a better solution is to fix this in `create_MaxMinLoc`, instead of fixing it in `ast_common_visitor.h`.

gxyd: But ofcourse, @HarshitaKalani knows much better about intrinsics.

integration_tests/intrinsics_369.f90
```diff
@@ -0,0 +1,4 @@
+program intrinsics_369
+    print *, minloc([1, 2], .true.)
```
gxyd, line 2: Two things to consider:
1. if you want to ensure that computation is done at compile time, it would be better to do something like: `integer, parameter :: x = minloc([1, 2], .true.)`, where `parameter` attribute ensures that it's evaluated at compile-time.
2. Eventually, do something like `if (lhs /= rhs) error stop` to ensure that the computed value as you expect it to be.

src/lfortran/semantics/ast_common_visitor.h
HarshilShah1804: @gxyd I initially tried to fix the issue in the create_MaxMinLoc, but since only arguments are passed to create_MaxMinLoc and not keywords, it was causing this issue https://github.com/lfortran/lfortran/pull/6671#pullrequestreview-2695651123 if the dimension is given with keyword. Therefore, I found it a better solution to assign the arguments to their desired place in the `ast_common_visitors.h` only.

integration_tests/intrinsics_369.f90
HarshilShah1804: I have updated the test. Thanks.

src/lfortran/semantics/ast_common_visitor.h
HarshitaKalani: I agree. We'll need to look for a robust way which works for all the intrinsics.

PR #6698, opened 2025-03-20
Implicitly allocate with StructConstructor 2

### LINKED ISSUE #1726: Run-time instantiation of derived types with structure constructors fails
When attempting to compile the following code
```
module hello_module

   implicit none
   private
   
   public :: HelloClass
   
   type :: HelloClass
   contains
      procedure :: printout
   end type HelloClass
   
contains
   
   subroutine printout(self)
      class(HelloClass), intent(in) :: self
      write(*,*) 'Hello world!'
   end subroutine printout
   
end module Hello_Module

program greetings

   use hello_module, only: HelloClass

   type(HelloClass), allocatable :: hello

   hello = HelloClass()
   
   call hello%printout()
   
end program greetings
```
LFortran crashes with a core dump:
```
lfortran: /usr/include/llvm-13/llvm/IR/Instructions.h:921: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
Aborted (core dumped)
```
Doing the instantiation of object `hello` statically, i.e. by omitting the `allocatable` attribute and the use of the `HelloClass()` structure constructor, the code above compiles and runs correctly. So the problem could be related to either one of the aforementioned two.

Getting this fixed will be required in order to move towards the support of polymorphism by LFortran.

xaerru:
Previous reverted PR: #6683 

Fix #1726 

Added a commit to keep track of allocated symbols in body visitor. So we only allocate if not already allocated.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -37,6 +37,7 @@ public:
     std::vector<ASR::symbol_t*> do_loop_variables;
     std::map<ASR::asr_t*, std::pair<const AST::stmt_t*,int64_t>> print_statements;
     std::vector<ASR::DoConcurrentLoop_t *> omp_constructs;
+    std::set<ASR::symbol_t*> allocated_symbols;
 
     BodyVisitor(Allocator &al, ASR::asr_t *unit, diag::Diagnostics &diagnostics,
         CompilerOptions &compiler_options,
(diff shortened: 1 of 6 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1506,6 +1510,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        if (ASR::is_a<ASR::ArrayItem_t>(*v)) {
+            ASR::ArrayItem_t *arri = ASR::down_cast<ASR::ArrayItem_t>(v);
+            return get_allocate_expr_sym(arri->m_v);
+        }
+        if (ASR::is_a<ASR::ArraySection_t>(*v)) {
+            ASR::ArraySection_t *arrs = ASR::down_cast<ASR::ArraySection_t>(v);
+            return get_allocate_expr_sym(arrs->m_v);
+        }
+        LCOMPILERS_ASSERT(false);
+        return nullptr;
```
Pranavchiku, line 1531: What happens if we do not allocate it right now? If it segfaults, then this assertion is fine else let's not throw and assert failure.

gxyd (approved):
Looks good to me, thanks @xaerru

Issue #6708, opened 2025-03-21
Statement function using passed-in function causes bogus compile errors and ICE.


w6ws:
```
      SUBROUTINE extfun (FUN)
      REAL  X1, X, H
      REAL G8, FUN
      ! Statement function calls passed in function
      G8(X,H)=FUN(X-X1*H)

      END
```
With no compilation options I get:
```
$ lfortran -c *.f 2>&1 | more
semantic error: Statement function needs to be declared.
 --> bsgq8.f:5:7
  |
5 |       G8(X,H)=FUN(X-X1*H)
  |       ^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
However the statement function, G8, has already been declared as having type REAL.

If I then try --implicit-typing:
```
$ lfortran -c --implicit-typing *.f 2>&1 | more
ASR verify pass error: The variable in ArrayItem must be an array, not a sca
lar
 --> bsgq8.f:5:15
  |
5 |       G8(X,H)=FUN(X-X1*H)
  |               ^^^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
Finally, if I specify both --implicit-interface and --implicit-typing, the compiler ICEs:
```
$ lfortran -c --implicit-typing --implicit-interface *.f 2>&1 | more
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/inst/bin/lfortran", 
in _start
  File "/build/glibc-FcRMwW/glibc-2.31/csu/../csu/libc-start.c", line 3
08, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line
 2690, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line
 2573, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, fa
lse,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line
 1165, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, s
td::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::ch
[...]

PR #6712, opened 2025-03-22
fix: accessign variable declared in upper case while declaring statement function

Pranavchiku:
Towards #6708.


Pranavchiku:
This is weird.

Pranavchiku:
It is working fine for me locally, also the PR checks passes. Something is going on with CI.

PR #6719, opened 2025-03-22
fix: ensure array dimensions are compile-time constants

akramhany:
This is a fix for the issue #6179, but there is only one issue, which is that I am not handling when functions are used as an array dimensions, e.g:
```fortran
module my_math
  implicit none
contains

  integer function my_abs(x) result(res)
    implicit none
    integer, intent(in) :: x
    if (x < 0) then
      res = -x
    else
      res = x
    end if
  end function my_abs

end module my_math

program main
  use my_math
  implicit none

  integer :: arr(my_abs(-5))  ! Using my_abs to determine array size

  print *, "Array size: ", size(arr)

end program main
```
The above code would run normally, while it should give an error, and the reason is that when I don't allow functions to be used, the following `ctest` doesn't pass:
```
(lf)  ✘ akramhany@akramhany  ~/Akram/Projects/Fortran/lfortran   ArraysShapeNotConst ±  ctest --rerun-failed --output-on-failure        
Test project /home/akramhany/Akram/Projects/Fortran/lfortran
    Start 3: test_lfortran
1/1 Test #3: test_lfortran ....................***Failed    0.34 sec
<string>:5:9: error: expected instruction opcode
    %1 =x alloca i64
        ^
<string>:5:9: error: expected instruction opcode
    %1 =x alloca i64
        ^
<string>:5:25: error: use of undefined value '@count'
    %1 = load i64, i64* @count
                        ^
0    0    0    0    0    0    0    0    0    0
5    5    5
src/lfortran/tests/ir.ll: error: Could not open input file: No such file or directory
src/lfortran/tests/ir.ll: error: Could not open input file: No such file or directory
/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1229:
TEST CASE:  FortranEvaluator asr verify 2

/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1256: ERROR: CHECK( r.ok ) is NOT correct!
  values: CHECK( false )

/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1265: ERROR: CHECK( r.ok ) is NOT correct!
  values: CHECK( false )

/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1270: ERROR: CHECK( r.result.i32 == 1 ) is NOT correct!
  values: CHECK( 0 == 1 )

/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1274: ERROR: CHECK( r.result.i32 == 1 ) is NOT correct!
  values: CHECK( 0 == 1 )

/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/tests/test_llvm.cpp:1278: ERROR: CHECK( r.result.i32 == 1 ) is NOT correct!
  values: CHECK( 0 == 1 )

[doctest] test cases:  95 |  94 passed | 1 failed | 0 skipped
[doctest] assertions: 673 | 668 passed | 5 failed |
[doctest] Status: FAILURE!


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.34 sec

The following tests FAILED:
	  3 - test_lfortran (Failed)
Errors while running CTest
```

The above error is coming from the following test, which is expected to pass (I don't understand why):
```cpp
TEST_CASE("FortranEvaluator asr verify 2") {
    CompilerOptions cu;
    cu.interactive = true;
    cu.po.runtime_library_dir = LCompilers::LFortran::get_runtime_library_dir();
    FortranEvaluator e(cu);
    LCompilers::Result<FortranEvaluator::EvalResult>
    r = e.evaluate2(R"(
pure function id(x) result(r)
    integer, intent(in) :: x
    integer :: r
    r = x
end function id
)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::none);
    r = e.evaluate2(R"(
subroutine sa(l, a)
    integer, intent(in) :: l
    integer, intent(inout) :: a(id(l))
    
    integer :: i
    
    do i = 1, size(a)
        a(i) = a(i) + 1
    end do
end subroutine sa
)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::none);
    r = e.evaluate2("integer :: arr(3)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::none);
    r = e.evaluate2("arr = 0");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::statement);
    r = e.evaluate2("call sa(3, arr)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::statement);
    r = e.evaluate2("arr(1)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::integer4);
    CHECK(r.result.i32 == 1);
    r = e.evaluate2("arr(2)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::integer4);
    CHECK(r.result.i32 == 1);
    r = e.evaluate2("arr(3)");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::integer4);
    CHECK(r.result.i32 == 1);
}
```

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1728,31 +1728,37 @@ public:
             left, end_bin_op->m_op, right, end_bin_op->m_type, end_bin_op->m_value));
     }
 
-    bool dimension_attribute_error_check(ASR::expr_t* dim_expr) {
-        if (ASR::is_a<ASR::Var_t>(*dim_expr)) {
-            ASR::Var_t* dim_expr_var = ASR::down_cast<ASR::Var_t>(dim_expr);
-            ASR::symbol_t* dim_expr_sym = dim_expr_var->m_v;
-            SymbolTable* symbol_scope = ASRUtils::symbol_parent_symtab(dim_expr_sym);
-            if (ASR::is_a<ASR::Variable_t>(*dim_expr_sym)) {
-                ASR::Variable_t* dim_expr_variable = ASR::down_cast<ASR::Variable_t>(dim_expr_sym);
+	bool var_dimension_error_check(ASR::Var_t* dim_expr_var) {
+        ASR::symbol_t* dim_expr_sym = dim_expr_var->m_v;
+        SymbolTable* symbol_scope = ASRUtils::symbol_parent_symtab(dim_expr_sym);
 
-                if (dim_expr_variable->m_type->type != ASR::ttypeType::Integer) {
+        if (ASR::is_a<ASR::Variable_t>(*dim_expr_sym)) {
+            ASR::Variable_t* dim_expr_variable = ASR::down_cast<ASR::Variable_t>(dim_expr_sym);
+
+            if (dim_expr_variable->m_type->type != ASR::ttypeType::Integer) {
+                return true;
+            } else {
+                if ((dim_expr_variable->m_storage != ASR::storage_typeType::Parameter) && !(in_Subroutine) && (symbol_scope->counter == current_scope->counter)) {
                     return true;
-                } else {
-
-                    if ((dim_expr_variable->m_storage != ASR::storage_typeType::Parameter) && !(in_Subroutine) && (symbol_scope->counter == current_scope->counter)) {
-                        return true;
-                    }
                 }
             }
-        } else {
+		}
 
-            ASR::ttype_t* dim_expr_type = ASRUtils::expr_type(dim_expr);
+		return false;
+	}
 
-            if (dim_expr_type->type != ASR::ttypeType::Integer) {
-                return true;
-            }
+    bool dimension_attribute_error_check(ASR::expr_t* dim_expr) {
+        ASR::ttype_t* dim_expr_type = ASRUtils::expr_type(dim_expr);
+
+		// If any expression doesn't evaluate to integer, return true
+        if (dim_expr_type->type != ASR::ttypeType::Integer) {
+            return true;
         }
+
+		if (ASR::is_a<ASR::Var_t>(*dim_expr)) {
+			ASR::Var_t* dim_expr_var = ASR::down_cast<ASR::Var_t>(dim_expr);
+			return var_dimension_error_check(dim_expr_var);
+		}
 
         return false;
     }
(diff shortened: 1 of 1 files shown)


akramhany:
> ```fortran
> subroutine f(x)
> integer :: x(2)
> integer :: x2(2)
> integer :: arr(kind(x2))
> end subroutine
> ```
> 
> So the logic I think is that either it has to be a compile time constant, such as in `kind(x2)`, or it can be a function whose arguments are only composed of function arguments.

I am sorry but I don't totally understand the above part. How is `kind(x2)` a compile time constant? is it because kind only needs to know the type of the variable? And also if the dimension was a call to a function that has some function arguments passed to it, like this:
```fortran
module y
contains
    integer function array_size(n) result(size_out)
        integer, intent(in) :: n
        size_out = n + 2
    end function array_size

    subroutine f(n)
        integer, intent(in) :: n
        integer :: arr(array_size(n))

        print *, "Array declared with size:", size(arr)
    end subroutine f
end module y

program main
    use y
    call f(3)
end program main
```

Then shouldn't this also not be allowed? I tried it in gfortran and I got some errors specifying that the function `array_size` needs to be pure.

akramhany:
> Consequently, I would not walk the whole expression again as you do in this PR. Rather, I would simply check if the result is a compile time constant. The exception is `integer :: arr(array_size(n))`, but there you want to check that the type of `arr` is constructed using FunctionParam and does not contain `var`. This is already enforced while constructing the type.

I apologize but I don't entirely understand you here. I know that I can see if an expression is compile time constant based on it's `m_value`, which would be `nullptr` if the expression is not compile time constant, but the problem still exists about how can I check that the variable used is not an `intent (in or inout)`? I would still have to go through the whole expression, no?

akramhany:
I have tried thinking of another approach for the problem, but the only solution I can think of would be to determine if the expression is valid as an array dimension or not in the parsing stage, but I don't really now how we can do that, as I don't understand how the parser work, so I guess I will just leave this issue like this for now until someone else comes up with a better solution.

PR #6740, opened 2025-03-26
CI: use the commit of pot3d branch which exits with non-zero status on validation failure

gxyd:
## Description

I've pushed a commit: https://github.com/gxyd/POT3D/commit/e2f2a2b6668e81f5d9f1441c20d64b8c7446aa6b to branch `lf_hdf5_mpi_namelist_global_workarounds`.

This is an important change, we definitely should've it in our CI. The linked commit ensures that when validation has failed, we exit with non-zero status.

I shouldn't have missed this change. We can possibly squash this commit with the commit prior to this on the same branch, we also would need to update the commit hash in the blog post as well (can't think of anywhere else though).


gxyd:
I missed porting this change from our other branch where the development happened, I've tested it locally on my machine by first wrongly update the `testsuite/validation/validation/pot3d.out` and seeing if `./validate.sh` exits with non-zero status now or not, and it does exit with non-zero status now. (that wasn't the case before, I know it's a big miss and could've costed us)

Issue #6720, opened 2025-03-22
Bind(C) with `c_ptr` as the return value


adit4443ya:
# MRE 
---
```fortran
module mpi
    contains

    function c_mpi_comm_f2c(comm_f) bind(C, name="MPI_Comm_f2c")
        use iso_c_binding, only: c_int, c_ptr
        integer(c_int), value :: comm_f
        type(c_ptr) :: c_mpi_comm_f2c  ! MPI_Comm as pointer
    end function c_mpi_comm_f2c

    subroutine MPI_Barrier(comm, ierror)
        use iso_c_binding, only: c_int, c_ptr
        integer, intent(in) :: comm
        integer, intent(out), optional :: ierror
        type(c_ptr) :: c_comm

        c_comm = c_mpi_comm_f2c(comm)
    end subroutine MPI_Barrier
end module



program barrier_1
    use mpi
    implicit none
    integer, parameter :: MPI_COMM_WORLD = 0
    integer :: ierr
    call MPI_Barrier(MPI_COMM_WORLD, ierr)
end program
```

<details>
<summary>LFortran Output</summary>

```console
 aditya-trivedi   tests    detached at f16bc9f ≢  ?7 ~3    lfortran -c MRE.f90 
Traceback (most recent call last):
  Binary file "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran.cpp", line 2690, in main
    return main_app(argc, argv);
  File "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran.cpp", line 2573, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/aditya-trivedi/thats_me/main/lfortran/src/bin/lfortran.cpp", line 1165, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
[...]

PR #6744, opened 2025-03-26
Handle type(C_PTR) return type from Bind(C) FunctionCalls

adit4443ya:
Towards #6720

### DIFF
--- a/integration_tests/bindc_05.f90
+++ b/integration_tests/bindc_05.f90
@@ -1,28 +1,55 @@
-module mpi_barr
-    contains
+module mpi
+    use iso_c_binding, only: c_int, c_ptr, c_f_pointer, c_null_ptr, c_associated
+    implicit none
 
-    function c_mpi_comm_f2c(comm_f) bind(C, name="MPI_Comm_f2c")
-        use iso_c_binding, only: c_int, c_ptr
-        integer(c_int), value :: comm_f
-        type(c_ptr) :: c_mpi_comm_f2c  ! MPI_Comm as pointer
-    end function c_mpi_comm_f2c
+    interface
+        function ax(comm_f) bind(C, name="ax")
+            import :: c_int, c_ptr
+            integer(c_int), value :: comm_f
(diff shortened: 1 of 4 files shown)


integration_tests/bindc_05.f90
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+program bindc_05
+    use mpi_barr
+    implicit none
+    integer, parameter :: MPI_COMM_WORLD = 0
+    integer :: ierr
+    call MPI_Barrier(MPI_COMM_WORLD, ierr)
+end program
```
gxyd, line 28: This is great, but I would write a C function in an extra file with this integration test, something like:
```c
#include <stdio.h>
#include <stdlib.h>

void* ax(int comm_f) {
    int* p = (int*)malloc(sizeof(int));
    if (p == NULL) {
        fprintf(stderr, "Memory allocation failed\n");
        return NULL;
    }
    *p = comm_f;
    printf("C function received: %d, returning pointer: %p\n", comm_f, (void*)p);
    return (void*)p;
}

```

and then at the Fortran end, I would use `c_f_pointer` and store the value underneath `c_comm` (which is type `c_ptr`), and store it's value in a Fortran integer, and make sure it's value is what I passed for `comm` (i.e. `MPI_Comm_World`, I would actually assign `MPI_Comm_WORLD`  a non-zero value, unlike zero as it's currently done also)

gxyd (approved):
The fix looks good to me, only I think needs improvement in the added test case.

integration_tests/bindc_05.f90
gxyd: The reason to assign non-zero value to MPI_COMM_WORLD is because, I've noticed non-initialized and random values also giving a zero value in Fortran.

gxyd (approved):
I've pushed a minor modification to use `error stop`, this looks pretty good now.

Also, I request that you don't do a force push on PR (avoid rebase, prefer *Merge* if necessary), that just makes the review a little tricky.

Issue #6750, opened 2025-03-27
[stdlib: logger] handle keywords correctly for class procedure


HarshitaKalani:
```fortran
module kwargs_02_stdlib_logger
    type :: logger_type
    contains
        private
        procedure, public, pass(self) :: add_log_file
    
    end type logger_type
    contains
        subroutine add_log_file(self, filename, unit)
            class(logger_type), intent(inout) :: self
            character(*), optional :: filename
            integer, optional :: unit
        end subroutine add_log_file
end module kwargs_02_stdlib_logger
    
program main
    use kwargs_02_stdlib_logger
    type(logger_type) :: logger
    integer :: unit
    call logger % add_log_file(unit=unit)
end program
```
```console
$ gfortran a.f90 && ./a.out
$ lfortran a.f90
semantic error: Argument was not specified
  --> a.f90:20:5
   |
20 |     call logger % add_log_file(unit=unit)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 0-th argument not specified for add_log_file


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #6751, opened 2025-03-27
fix: handle optional keyword args correctly for class procedure

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6750

### DIFF
--- a/integration_tests/class_15.f90
+++ b/integration_tests/class_15.f90
@@ -10,8 +10,8 @@ module class_15_mod
             class(logger_type), intent(inout) :: self
             character(*), optional :: filename
             integer, optional :: unit
-            filename = "lfortran"
-            unit = 10
+            if (present(filename)) filename = "lfortran"
+            if (present(unit)) unit = 10    
         end subroutine add_log_file
 end module
     
--- a/integration_tests/class_15.f90
+++ b/integration_tests/class_15.f90
@@ -1,0 +1,32 @@
+module class_15_mod
+    type :: logger_type
+    contains
+        private
+        procedure, public, pass(self) :: add_log_file
+    
+    end type logger_type
+    contains
+        subroutine add_log_file(self, filename, unit)
+            class(logger_type), intent(inout) :: self
+            character(*), optional :: filename
+            integer, optional :: unit
+            filename = "lfortran"
+            unit = 10
+        end subroutine add_log_file
+end module
+    
+program class_15
+    use class_15_mod
+    type(logger_type) :: logger
+    integer :: unit
+    character(len=100) :: filename
+    call logger % add_log_file(unit=unit) !> which is already done
+    call logger % add_log_file(filename=filename)
+    call logger % add_log_file(unit=unit, filename=filename)
+    call logger % add_log_file(filename=filename, unit=unit) ! reverse the order
+    call logger % add_log_file()
+    print *, filename
+    if (filename /= "lfortran") error stop
+    print *, unit
+    if (unit /= 10) error stop
+end program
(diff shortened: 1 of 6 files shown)


integration_tests/class_15.f90
```diff
@@ -0,0 +1,21 @@
...
+    end type logger_type
+    contains
+        subroutine add_log_file(self, filename, unit)
+            class(logger_type), intent(inout) :: self
+            character(*), optional :: filename
+            integer, optional :: unit
+        end subroutine add_log_file
+end module
+    
+program main
```
gxyd, line 16: Probably use the program name as `class_15`?

integration_tests/class_15.f90
```diff
@@ -0,0 +1,21 @@
...
+    use class_15
+    type(logger_type) :: logger
+    integer :: unit
+    call logger % add_log_file(unit=unit)
```
gxyd, line 20: Now that we're at it, would be great to have different combination of testing it, e.g;

```f90
call logger % add_log_file(unit=unit) !> which is already done
call logger % add_log_file(filename=filename)
call logger % add_log_file(unit=unit, filename=filename)
call logger % add_log_file(filename=filename, unit=unit) ! reverse the order
call logger % add_log_file()
```

also, define  a new subroutine `add_log_file_with_one_filename_optional_argument_and_fixed_unit_argument`, and then have a combination of possible argument calls for it, 
```f90
call logger % add_log_file(unit=unit)
call logger % add_log_file(filename=filename) !> this should raise an error
call logger % add_log_file(unit=unit, filename=filename)
call logger % add_log_file(filename=filename, unit=unit) ! reverse the order
call logger % add_log_file() !> this should raise an error
```

gxyd (commented):
I've left some comments requesting for more test cases, the reason I request for that is because now that you've fixed one bug in it, and you understand the logic well to fix other possible bugs which are there in this block of code (which I think there are more bugs, I've a feeling, it just didn't feel right, but I'm happy to be proved wrong for it.)

integration_tests/class_15.f90
gxyd, line 20: Also, I would like to enhance the subroutine in such a way that we test the output value from the subroutine as well, doing nothing within  a subroutine doesn't test it as well as it would if we check the value returned by it or modified by it.

HarshitaKalani:
@gxyd can you please review it and let me know if this looks good now? For the cases I haven't included yet, there already exists test for that.

tests/errors/continue_compilation_1.f90
```diff
@@ -21,8 +27,15 @@ subroutine display(self, extra_arg)
...
-end module
+    subroutine add_log_file(self, filename, unit)
+        class(logger_type), intent(inout) :: self
+        character(*), optional :: filename
+        integer :: unit
+        filename = "lfortran"
+        unit = 10
+    end subroutine add_log_file
 
+end module
```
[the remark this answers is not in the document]
HarshitaKalani: Done.

tests/errors/continue_compilation_1.f90
```diff
@@ -304,4 +320,6 @@ program continue_compilation_1
     allocate(arr4(3), source=reshape([1, 2, 3, 4, 5, 6], [2, 3]))
     allocate(arr4, source=7)
 
+    call logger % add_log_file(filename=filename) !> this should raise an error
+    call logger % add_log_file() !> this should raise an error
```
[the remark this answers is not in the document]
HarshitaKalani: Yep.

gxyd (approved):
Looks good to me. Thanks @HarshitaKalani for resolving my comments as well.

PR #6780, opened 2025-03-30
Sync Libasr with LPython

Shaikh-Ubaid:


Shaikh-Ubaid:
Yes, the LibASRs are same. 

>If so, use a submodule -- point it to upstream LFortran, and use the hash from your PR, it will work.

On it.

Shaikh-Ubaid:
>v

Shaikh-Ubaid:
>cd libasr

For now, we are just calling it as `lfortran`. If it is fine, we can have a subsequent PR changing the name from lfortran to libasr.

Shaikh-Ubaid:
```bash
cd libasr
git checkout 4bb120effcca83388f9c6697266cc5cd425c6467
cd ..
git commit -a
```
Are we sure this will work? This commit is not available in the https://github.com/lfortran/lfortran repository. It is from my forked upstream https://github.com/Shaikh-Ubaid/lfortran.

Shaikh-Ubaid:
>Yes, I am sure. @Pranavchiku and I tested it.

Seems like it works.

Issue #6760, opened 2025-03-28
[stdlib: io] Handle allocatable array inside `cmplx` intrinsic


jinangshah21:
```fortran
program test
   real, allocatable :: a(:), b(:) 
   complex :: c(2)
    allocate(a(2))
    allocate(b(2))
    a = 1
    b = 2
   c = cmplx(a,b)
   print *, c
end program
```

LFortran Error :
```console
 b.visit_TranslationUnit(ast);
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 167, in LCompilers::LFortran::BodyVisitor::visit_TranslationUnit(LCompilers::LFortran::AST::TranslationUnit_t const&)
    visit_ast(*x.m_items[i]);
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/ast.h", line 4969, in LCompilers::LFortran::AST::BaseVisitor<LCompilers::LFortran::BodyVisitor>::visit_ast(LCompilers::LFortran::AST::ast_t const&)
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/ast.h", line 4925, in void LCompilers::LFortran::AST::visit_ast_t<LCompilers::LFortran::BodyVisitor>(LCompilers::LFortran::AST::ast_t const&, LCompilers::LFortran::BodyVisitor&)
    case astType::mod: { v.visit_mod((const mod_t &)x); return; }
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/ast.h", line 4972, in LCompilers::LFortran::AST::BaseVisitor<LCompilers::LFortran::BodyVisitor>::visit_mod(LCompilers::LFortran::AST::mod_t const&)
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/ast.h", line 4549, in void LCompilers::LFortran::AST::visit_mod_t<LCompilers::LFortran::BodyVisitor>(LCompilers::LFortran::AST::mod_t const&, LCompilers::LFortran::BodyVisitor&)
    case modType::Program: { v.visit_Program((const Program_t &)x); return; }
[...]

PR #6786, opened 2025-03-30
fix: Handle allocatable array inside cmplx intrinsic

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6760


HarshitaKalani:
Yep, I agree.

PR #6793, opened 2025-03-31
Dumps LSP server stdout and stdin logs on test failure

dylon:
Changes:
* Specifies utf-8 as byte encoding
* Migrates several buffers from StringIO to BytesIO
* Adds SpyIO to pipe the server's stdin and stdout to their respective log files

### DIFF
--- a/src/server/tests/src/llanguage_test_client/lsp_json_stream.py
+++ b/src/server/tests/src/llanguage_test_client/lsp_json_stream.py
@@ -1,44 +1,49 @@
 import io
 import json
 import time
+from io import BytesIO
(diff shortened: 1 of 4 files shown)


dylon:
The error, `Cannot choose between targets "x86-64" and "x86-64"`, seems to be coming from here: https://github.com/llvm/llvm-project/blob/b9b9addae6c79c72394c8260106f8ba5281e9747/llvm/lib/MC/TargetRegistry.cpp#L178-L179. LLVM seems to have redundant entries in its target registry for `x86-64`, for some reason.

Issue #6794, opened 2025-03-31
fix: `len` does not work for array input


HarshitaKalani:
```fortran
program test
    implicit none
    character(len=4) :: array(2)
    print*, len(array)
end program test
   
```
```console
$ lfortran c.f90
Segmentation fault (core dumped)
$ gfortran c.f90 && ./a.out
           4
```

PR #6795, opened 2025-03-31
fix: handle array input for len intrinsic

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6794


gxyd (approved):
This is good, thanks for the fix @HarshitaKalani

Issue #6770, opened 2025-03-29
download from source tarball wastes space


harperjf:
Just downloaded lfortran 0.50.0 and found that like 0.49.0 it has 2 versions of the compiler in directories ```src```  and ```inst```. That  wastes a lot of space. Evidence:
```
(lf) john:~/lfortran-0.50.0/src/bin$ ls -o ../../*/bin/lfortran
-rwxr-xr-x 1 john 129491904 Mar 29 19:42 ../../inst/bin/lfortran
-rwxrwxr-x 1 john 129491904 Mar 29 19:42 ../../src/bin/lfortran
(lf) john:~/lfortran-0.50.0/src/bin$
```

harperjf:
I used https://github.com/lfortran/lfortran/releases/tag/v0.50.0 and clicked on Releases
Under Assets, clicked on lfortran-0.50.0.tar.gz
That downloaded  'lfortran-0.50.0(1).tar.gz' into directory ~/Downloads
because I had downloaded lfortran-0.50.0.tar.gz two days ago.
The two tarballs appear to be identical. Proof:
```
(lf) john:~/Downloads$ ls -o lfortran-0.50.*
-rw-rw-r-- 1 john 2129421 Mar 31 13:10 'lfortran-0.50.0(1).tar.gz'
-rw-rw-r-- 1 john 2129421 Mar 29 19:18  lfortran-0.50.0.tar.gz
(lf) john:~/Downloads$ diff lfortran-0.50.*
(lf) john:~/Downloads$ 
```
In my system ```lo``` is aliased to ```ls -taro```, and ~/helpful.hin is a non-executable file of helpful hints that I wrote to remind myself of various things. Between Mar 29 19:18 and 19:30 I was doing things with ifx, irrelevant to lfortran. Then my history file contained
```
  815  29/03 19:30 cp lfortran-0.50.0.tar.gz ..
  816  29/03 19:30 cd
  817  29/03 19:30 emacs *hin &
  818  29/03 19:32 conda install libunwind
  819  29/03 19:33 conda update -n base -c conda-forge conda
  820  29/03 19:33 tar xzf lfortran-0.50.0.tar.gz
  821  29/03 19:33 cd lfortran-0.50.0
  822  29/03 19:34 lo
  823  29/03 19:34 cmake -DWITH_LLVM=yes -DCMAKE_INSTALL_PREFIX=`pwd`/inst .
  824  29/03 19:34 make -j4
  825  29/03 19:42 make install
[...]

PR #6802, opened 2025-03-31
Do not install LFortran when building it

certik:
Towards #6770.

### DIFF
--- a/build1.sh
+++ b/build1.sh
@@ -11,9 +11,8 @@ cmake \
     -DWITH_RUNTIME_STACKTRACE=yes \
     -DWITH_LSP=no \
     -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH_LFORTRAN;$CONDA_PREFIX" \
-    -DCMAKE_INSTALL_PREFIX=`pwd`/inst \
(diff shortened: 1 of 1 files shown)


harperjf:
I followed the https://docs.lfortran.org/en/installation/ instructions on using a source tarball. The relevant part seems to be
```
cmake -DWITH_LLVM=yes -DCMAKE_INSTALL_PREFIX=`pwd`/inst .
make -j8
make install

This will install lfortran into inst/bin.
```
What should the instructions have said, to leave me with one copy of the compiler not two? (I know much more about Fortran than cmake and make.)

harperjf:
From my point of view the instructions were misleading because I had no idea what ```make install``` would do. My suggestions (including where to find the compiler, which was not where the instructions said it would be) actually worked.
Also, my test programs hit fewer bugs in lfortran 0.50.0 than in 0.49.0 . Thanks to all who made that happen!

harperjf:
I now know the disadvantage of make install. But I suppose it was in the instructions because it had some benefit. What?

gxyd (approved):
Looks good to me, as it follows from the discussion in this issue and the linked issue, we definitely need to fix the instructions also to fix the linked issue.

Issue #6769, opened 2025-03-29
Warnings in a build


certik:
Not sure how they are not caught, but the [conda build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1210789&view=logs&j=be60363f-e4b7-5b4d-cc0b-2e04eaf5ba1a&t=a50466e8-94fb-5b34-007c-74688e22e6ce) has the following warnings. Let's fix them:
```
-- Found OpenSSL: $PREFIX/lib/libcrypto.so (found version "3.4.1")
CMake Warning (dev) at /home/conda/feedstock_root/build_artifacts/lfortran_1743220652778/_build_env/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (LIBUNWIND)
  does not match the name of the calling package (libunwind).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/Findlibunwind.cmake:5 (find_package_handle_standard_args)
  CMakeLists.txt:377 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found LIBUNWIND: $PREFIX/include

[  2%] Building CXX object src/libasr/CMakeFiles/lfortran_utils.dir/diagnostics.cpp.o
/home/conda/feedstock_root/build_artifacts/lfortran_1743220652778/work/src/libasr/runtime/lfortran_intrinsics.c: In function 'get_stacktrace_addresses':
/home/conda/feedstock_root/build_artifacts/lfortran_1743220652778/work/src/libasr/runtime/lfortran_intrinsics.c:4153:5: warning: implicit declaration of function '_Unwind_Backtrace' [-Wimplicit-function-declaration]
 4153 |     _Unwind_Backtrace(unwind_callback, &d);
      |     ^~~~~~~~~~~~~~~~~

[  9%] Building C object src/runtime/legacy/CMakeFiles/lfortran_runtime.dir/__/__/libasr/runtime/lfortran_intrinsics.c.o
/home/conda/feedstock_root/build_artifacts/lfortran_1743220652778/work/src/libasr/runtime/lfortran_intrinsics.c: In function 'get_stacktrace_addresses':
[...]

PR #6803, opened 2025-04-01
fix release build warning from macOS

gxyd:
## Description

Fixes one of the warnings present in https://github.com/lfortran/lfortran/issues/6769


Pranavchiku (approved):

gxyd:
> It's weird that it doesn't fail at our CI --- do we not test with `-Werror` in Release mode, when the ASSERT is not present?

@certik , we *do* test with `-Werror`, it just happened that this warning is generated with LLVM > 16 but at the CI we only test with LLVM version 11, and hence the warning isn't detected at the CI.

I think we should release build with LLVM version > 16 as well. Wdyt @certik ?

PR #6839, opened 2025-04-04
fix: wrap with Array ttype when declaration of `cptr` is passed

gxyd:
## Description

also make sure to load the ArrayItem in visit_CPtrToPointer

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4476,6 +4476,10 @@ public:
                     al, loc, type, dims.p, dims.size(), abi, is_argument);
             } else if (v && ASRUtils::is_c_ptr(v, derived_type_name)) {
                 type = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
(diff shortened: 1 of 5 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4476,6 +4476,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     al, loc, type, dims.p, dims.size(), abi, is_argument);
             } else if (v && ASRUtils::is_c_ptr(v, derived_type_name)) {
                 type = ASRUtils::TYPE(ASR::make_CPtr_t(al, loc));
+                type = ASRUtils::make_Array_t_util(
+                    al, loc, type, dims.p, dims.size(), abi, is_argument,
+                    ASR::array_physical_typeType::DescriptorArray, false, is_dimension_star
+                );
```
gxyd, line 4482: earlier `type(c_ptr) :: c_requests(2)` wouldn't wrap `c_requests` ttype with an `Array` ttype, and hence in current *main*, it's simply `CPtr` instead of `Array(CPtr)`

PR #6847, opened 2025-04-04
enh: enhance cshift to work for n-D arrays

HarshilShah1804:
`cshift` now works for any n-dimensional array without explicit dimension checking. 

The `dim` argument of `cshift` still doesn't work appropriately. I will implement that in a subsequent PR.


HarshitaKalani (approved):
This is great. Great job @HarshilShah1804 !

PR #6850, opened 2025-04-05
Input Handling - Integer and Float

Th3C0d3Mast3r:
This is a new PR-in extension to the previous [PR](https://github.com/lfortran/lfortran/pull/6730/).

Solved the issues, and created proper .txt file which has the testcase; and using `run=true` in the `tests.toml` file. 

Kindly check the following and please do let me know if any further changes needed.


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3392,11 +3393,32 @@ LFORTRAN_API void _lfortran_backspace(int32_t unit_num)
...
 }
 
+// Improved input validation for integer reading
+// - Prevents auto-casting of invalid inputs to integers
 LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num)
 {
     if (unit_num == -1) {
-        // Read from stdin
-        (void)!scanf("%d", p);
+        char buffer[100];
```
Pranavchiku, line 3401: Why it is `100`  in `buffer[100]` ?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3392,11 +3393,32 @@ LFORTRAN_API void _lfortran_backspace(int32_t unit_num)
...
+
+        // Use strtok() to extract only the first token before any whitespace
+        char *token = strtok(buffer, " \t\n");
+        if (token == NULL) {
+            fprintf(stderr, "Error: Invalid input for int32_t.\n");
+            exit(1);
+        }
+
+        char *endptr;
+        *p = strtol(token, &endptr, 10);
```
Pranavchiku, line 3415:
```suggestion
        *p = (int32_t) strtol(token, &endptr, 10);
```

It is fine though.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3408,17 +3430,40 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num)
...
+        }
+
+        char *token = strtok(buffer, " \t\n");
+        if (token == NULL) {
+            fprintf(stderr, "Error: Invalid input for int64_t.\n");
+            exit(1);
+        }
+
+        char *endptr;
+        *p = strtoll(token, &endptr, 10);
```
Pranavchiku, line 3461:
```suggestion
        *p = (int64_t) strtoll(token, &endptr, 10);
```

Pranavchiku (approved):
Left few comments, rest looks good!

src/libasr/runtime/lfortran_intrinsics.c
Th3C0d3Mast3r: I was testing out this with a bit larger string test-cases, just to see if the code breaks or what happens-thus, took a buffer for 100 character length, just to be safe.

Yes, I suppose, if not required, we could even use `char buffer[25] ` or `char buffer[30]`  _(as for int32 we would require about 12 and for int64, it would be 20 characters, leaving an additional space of 18 and 10 characters respectively)_

tests/tests.toml
```diff
@@ -4526,3 +4522,11 @@ asr = true
...
 filename = "declaration1.f90"
 asr = true
+
+[[test]]
+filename = "errors/integerStringInput.f90"
+run = true
+
+[[test]]
+filename = "errors/floatStringInput.f90"
+run = true
```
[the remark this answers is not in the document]
Th3C0d3Mast3r: Extremely sorry for causing the misunderstanding.
I mistakenly repeated two files _(as was testing it once, and overlooked removing it)_ -thus, the total intended files are supposed to be 3- dealing with int32, int64 and float _(which will be using `invalidInput_integer.txt` and `invalidInput_float.txt` files respectively for testing)_

invalidInput_integer.txt
```diff
@@ -0,0 +1,23 @@
+xasxacscsd123s15x6
```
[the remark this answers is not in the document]
Th3C0d3Mast3r: yes, as error occurs and exited

Pranavchiku (changes requested):
Let's get the CI passing and then we can review.

PR #6852, opened 2025-04-05
enh: lhs reallocation for intrinsic types

xaerru:
Add automatic allocation for allocatable `integer`, `real`, `complex` and `logical`.

### DIFF
--- a/integration_tests/automatic_allocation_02.f90
+++ b/integration_tests/automatic_allocation_02.f90
@@ -1,0 +1,23 @@
+program automatic_allocation_02
+    implicit none
+
+    integer, allocatable :: i
+    real, allocatable :: r
+    complex, allocatable :: c
+    logical, allocatable :: l
+
+    i = 10
+    r = 4.4
+    c = (1, 2)
+    l = .true.
+
+    if (i /= 10) error stop
+    if (r /= 4.4) error stop
+    if (c /= (1, 2)) error stop
+    if (l .neqv. .true.) error stop
+
+    deallocate(i)
+    deallocate(r)
+    deallocate(c)
+    deallocate(l)
+end program automatic_allocation_02
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -37,7 +37,6 @@ public:
     std::vector<ASR::symbol_t*> do_loop_variables;
     std::map<ASR::asr_t*, std::pair<const AST::stmt_t*,int64_t>> print_statements;
     std::vector<ASR::DoConcurrentLoop_t *> omp_constructs;
-    std::set<ASR::symbol_t*> allocated_symbols;
 
     BodyVisitor(Allocator &al, ASR::asr_t *unit, diag::Diagnostics &diagnostics,
         CompilerOptions &compiler_options,
@@ -1482,9 +1481,6 @@ public:
         tmp = ASR::make_Allocate_t(al, x.base.base.loc,
                                     alloc_args_vec.p, alloc_args_vec.size(),
                                     stat, errmsg, source);
-        for (ASR::alloc_arg_t alloc_arg : alloc_args_vec) {
-            allocated_symbols.insert(get_allocate_expr_sym(alloc_arg.m_a));
-        }
 
         if (source) {
             current_body->push_back(al, ASRUtils::STMT(ASR::make_Allocate_t(al, x.base.base.loc,
@@ -3097,25 +3093,7 @@ public:
                     Label("",{x.base.base.loc})
                 }));
             throw SemanticAbort();
-        } else if (compiler_options.po.realloc_lhs &&
-                 ASR::is_a<ASR::Allocatable_t>(*ASRUtils::expr_type(target)) &&
-                 ASR::is_a<ASR::StructConstructor_t>(*value) &&
-                 (allocated_symbols.find(get_allocate_expr_sym(target)) == allocated_symbols.end())) {
-            // Implicitly allocate the Struct by pushing an Allocate_t to the body
-            Vec<ASR::alloc_arg_t> alloc_args; alloc_args.reserve(al, 1);
-            ASR::alloc_arg_t alloc_arg;
-            alloc_arg.loc = x.base.base.loc;
-            alloc_arg.m_a = target;
-            alloc_arg.m_dims = nullptr;
-            alloc_arg.n_dims = 0;
-            alloc_arg.m_type = nullptr;
-            alloc_arg.m_len_expr = nullptr;
-            alloc_args.push_back(al, alloc_arg);
-            current_body->push_back( al,
-                ASRUtils::STMT(
-                    ASR::make_Allocate_t(al, x.base.base.loc, alloc_args.p, 1, nullptr, nullptr, nullptr)));
         }
-
         check_ArrayAssignmentCompatibility(target, value, x);
 
         if( overloaded_stmt == nullptr ) {
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5677,6 +5677,7 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
     // Checks if target_expr is allocated and if not then allocate
     // Used for compiler_options.po.realloc_lhs
     void check_and_allocate(ASR::expr_t *target_expr) {
+        LCOMPILERS_ASSERT(compiler_options.po.realloc_lhs);
         ASR::ttype_t *asr_type = ASRUtils::type_get_past_pointer(
             ASRUtils::type_get_past_allocatable(
             ASRUtils::expr_type(target_expr)));
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -9384,7 +9384,9 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
             } else {
                 ASR::ttype_t* arg_type = expr_type(x.m_args[i].m_value);
                 int64_t ptr_loads_copy = ptr_loads;
-                ptr_loads = !LLVM::is_llvm_struct(arg_type);
+                if (LLVM::is_llvm_struct(arg_type)) {
+                    ptr_loads = 1;
+                }
                 this->visit_expr_wrapper(x.m_args[i].m_value);
 
                 if( x_abi == ASR::abiType::BindC ) {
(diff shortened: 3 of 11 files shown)


integration_tests/automatic_allocation_02.f90
```diff
@@ -0,0 +1,23 @@
...
+
+    integer, allocatable :: i
+    real, allocatable :: r
+    complex, allocatable :: c
+    logical, allocatable :: l
+
+    i = 10
+    r = 4.4
+    c = (1, 2)
+    l = .true.
```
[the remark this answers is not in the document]
xaerru: It will only allocate once. I added a class member variable named `allocated_symbols` in `BodyVisitor` which keeps track of all the allocated symbols. Before we allocate it we check if it is already in the set, and after allocating we add the symbol to the set.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3097,25 +3093,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-            alloc_arg.loc = x.base.base.loc;
-            alloc_arg.m_a = target;
-            alloc_arg.m_dims = nullptr;
-            alloc_arg.n_dims = 0;
-            alloc_arg.m_type = nullptr;
-            alloc_arg.m_len_expr = nullptr;
-            alloc_args.push_back(al, alloc_arg);
-            current_body->push_back( al,
-                ASRUtils::STMT(
-                    ASR::make_Allocate_t(al, x.base.base.loc, alloc_args.p, 1, nullptr, nullptr, nullptr)));
```
Pranavchiku, line 3116: let's add test that covers this. Find a case that triggered this code and check if it works with this PR.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5316,6 +5319,17 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
             builder->CreateStore(tmp, llvm_symtab[target_h]);
             return ;
         }
+
+        // When assigning to a StructInstanceMember, whose instance is allocatable
+        // Check if the underlying struct instance is allocated, if not allocate
+        if (compiler_options.po.realloc_lhs &&
+            ASR::is_a<ASR::StructInstanceMember_t>(*x.m_target)) {
+            ASR::StructInstanceMember_t *sim = ASR::down_cast<ASR::StructInstanceMember_t>(x.m_target);
+            if (ASR::is_a<ASR::Allocatable_t>(*expr_type(sim->m_v))) {
```
Pranavchiku, line 5328:
```suggestion
            if (ASRUtils::is_allocatable(ASRUtils::expr_type(sim->m_v))) {
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5645,14 +5654,53 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
             llvm_utils->dict_api->write_item(pdict, key, value, module.get(),
                                 dict_type->m_key_type,
                                 dict_type->m_value_type, name2memidx);
-        } else if (ASR::is_a<ASR::Allocatable_t>(*target_type) && ASR::is_a<ASR::StructType_t>(*value_type)) {
+        } else if (ASR::is_a<ASR::Allocatable_t>(*target_type)) {
```
Pranavchiku, line 5657: use `ASRUtils::is_allocatable`

integration_tests/automatic_allocation_02.f90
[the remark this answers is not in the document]
Pranavchiku: But below you removed `allocated_symbols`

Pranavchiku (commented):
This PR needs a concrete test on struct instance members and also justification on 

> What happens if you assign more than once? Will it allocate the already allocated variable again? 


Because as per the response, `allocated_symbols` is supposed to handle it but this PR removes that as well.

xaerru:
The ASR checking of allocation status is incorrect. But we merged it in https://github.com/lfortran/lfortran/pull/6698. The test for struct exists in that merged PR. It is `derived_types_45.f90`. In the test the `myint(44)` `StructConstructor` is converted to a `StructInstanceMember` assignment after the `class_constructor` pass.

We need to check at runtime if it is allocated or not, by adding a llvm if statement before every assignment. So I removed the ASR checking. I've linked a zulip message with more information below.

[#Internal > Finalization @ 💬](https://lfortran.zulipchat.com/#narrow/channel/452774-Internal/topic/Finalization/near/511230954)

integration_tests/automatic_allocation_02.f90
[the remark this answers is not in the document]
xaerru: Yes, because we can't check just in the body visitor or even in ASR. With the previous approach consider this case:
```fortran
program my
    integer, allocatable :: i

    integer :: n
    read *, n

    if (n == 0) then
        ! We insert allocate before this line
        i = 2
    end if

    ! We do not insert allocate before this line, this is a problem if n != 0
    i = 3

    print *, i

end program
```
I'm now checking if it is allocated or not in the llvm backend before every assignment when the option is on.

Pranavchiku:
The check if the variable is allocated is done at runtime. For every assignment with `--realloc-lhs` this introduces a `if, else` block. Is it not possible do it at compile time? If no, why?

Pranavchiku:
In that case I think this PR is good, I did a few checks to reassign a variable and it works except a case:

```fortran
program automatic_allocation_02
    implicit none

    integer, allocatable :: i
    real, allocatable :: r
    complex, allocatable :: c
    logical, allocatable :: l

    i = 10
    r = 4.4
    c = (1, 2)
    l = .true.

    if ( abs(r - 4.4) <= 1.0e-6 ) then
        l = .true.
    end if

    l = .true.

    print *, "i = ", i
    if (i /= 11) error stop
    print *, "r = ", r
    if (r /= 4.4) error stop
    print *, "c = ", c
    if (c /= (1, 2)) error stop
    print *, "l = ", l
    if (l .neqv. .true.) error stop

    deallocate(i)
    deallocate(r)
    deallocate(c)
    deallocate(l)
end program automatic_allocation_02
```

```console
% lfortran a.f90 --realloc-lhs
[...]
  %11 = fsub float* %10, float 0x40119999A0000000
  store float* %11, float* %call_arg_value, align 8
 float


```

xaerru:
Fixed that. But don't know why LLVM 16 in CI is failing.

xaerru:
The CI failure was unrelated. https://github.com/lfortran/lfortran/issues/7023

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1019,7 +1019,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                     llvm_utils->string_init(alloc_size, ptr_to_init);
                 } else if(ASR::is_a<ASR::StructType_t>(*curr_arg_m_a_type) ||
                           ASR::is_a<ASR::ClassType_t>(*curr_arg_m_a_type) ||
-                          ASR::is_a<ASR::Integer_t>(*curr_arg_m_a_type)) {
+                          ASR::is_a<ASR::Integer_t>(*curr_arg_m_a_type) ||
+                          ASR::is_a<ASR::Real_t>(*curr_arg_m_a_type) ||
+                          ASR::is_a<ASR::Complex_t>(*curr_arg_m_a_type) ||
+                          ASR::is_a<ASR::Logical_t>(*curr_arg_m_a_type)) {
```
gxyd, line 1025: I don't think that any of these are tested currently with the added test cases, would be good to have that.

In case this change isn't needed as a part of the PR, then maybe don't have the change at all unless we can add a test case here.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5316,6 +5319,17 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
...
             return ;
         }
+
+        // When assigning to a StructInstanceMember, whose instance is allocatable
+        // Check if the underlying struct instance is allocated, if not allocate
+        if (compiler_options.po.realloc_lhs &&
+            ASR::is_a<ASR::StructInstanceMember_t>(*x.m_target)) {
+            ASR::StructInstanceMember_t *sim = ASR::down_cast<ASR::StructInstanceMember_t>(x.m_target);
+            if (ASRUtils::is_allocatable(ASRUtils::expr_type(sim->m_v))) {
+                check_and_allocate(sim->m_v);
```
gxyd, line 5329: Neither is this block of code currently tested.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5645,14 +5654,53 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
...
+
             target = llvm_utils->CreateLoad(target);
             builder->CreateStore(value, target);
         } else {
             builder->CreateStore(value, target);
         }
     }
 
+    // Checks if target_expr is allocated and if not then allocate
+    // Used for compiler_options.po.realloc_lhs
```
gxyd, line 5678: Does this mean that `check_and_allocate` is only to be used when `compiler_options.po.realloc_lhs` is true?

If that's the case, we can have an explicit `LCOMPILERS_ASSERT` statement within this function indicating the same.

gxyd (commented):
I left some comments.

I usually feel more at ease when we know that we save the LLVM IR (of atleast one of the test cases) as a reference test, in this case we don't, I think it would be good to do that. We can consult @certik if it's a good idea if unsure.

src/lfortran/semantics/ast_body_visitor.cpp
gxyd: Would be nice to have a confirmation if this comment is resolved before we merge.

integration_tests/automatic_allocation_02.f90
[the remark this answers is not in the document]
gxyd: > It will only allocate once. I added a class member variable named `allocated_symbols` in `BodyVisitor` which keeps track of all the allocated symbols. Before we allocate it we check if it is already in the set, and after allocating we add the symbol to the set.

I'm not sure how was this resolved or resolved at all. The git's commit history doesn't make it obvious for me here, whether the initial approach in the PR was any different than what it is currently. (we have a force-push here)

src/libasr/codegen/asr_to_llvm.cpp
xaerru: This is tested in `derived_types_45.f90`. Added a llvm test for it as well.

src/libasr/codegen/asr_to_llvm.cpp
xaerru: Added a test.

https://github.com/lfortran/lfortran/pull/6852/commits/0ab0c9242f6b437e562d89f6155909931187fb85

src/lfortran/semantics/ast_body_visitor.cpp
xaerru: Yes, it can be resolved. These are changes made in #6698. This approach was not correct.

Pranavchiku:
Checking in the status for this? Are we good to merge?

Pranavchiku (approved):

gxyd:
> Checking in the status for this? Are we good to merge?

From my end, I just need to check if the conversation that @xaerru  resolved are actually resolved, if they are, I'm happy with it. I'll take a look

Issue #6837, opened 2025-04-04
Handle array indices inside struct array reference


jinangshah21:
```fortran
program temp 
    type :: line
        integer :: first
    end type

    integer :: arr(5) = [1, 2, 3, 4, 5]
    integer :: shift(5)
    type(line) :: l_arr(5)
    shift(:) = l_arr(arr)%first - 1
end program
```

LFortran Error:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90
ASR verify pass error: ArrayItem::m_type with array indices must be an array.
  --> a.f90:33:16
   |
33 |     shift(:) = l_arr(arr)%first - 1
   |                ^^^^^^^^^^^^^^^^ failed here
```

PR #6884, opened 2025-04-07
enh: handle array indices inside struct array

xaerru:
Fix https://github.com/lfortran/lfortran/issues/6837


jinangshah21 (approved):
I think this is good

Issue #6875, opened 2025-04-06
LSP: syntax error on hover


certik:
When you hover over `x` in a valid program (examples/expr2.f90`, it gives:
```
2025-04-06 11:26:34.882 [info] syntax error: Statement or Declaration expected inside program, found Variable name
 --> /Users/ondrej/repos/lfortran/examples/expr2.f90:1:1
  |
1 | program expr2
  | ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.

```

The reason has to do with some formatting, but we should not be trying to parse some invalid code, or if we do, we need to ignore the errors. We should not be outputting it like this, since this signals that something is wrong.

PR #6911, opened 2025-04-08
Runs LFortran in interactive mode

dylon:
Runs LFortran in interactive mode so it does not err on unexpected top-level block types when formatting code

Should address https://github.com/lfortran/lfortran/issues/6875


dylon:
We have refactored the code such that this is not an issue anymore. We will reconsider what to do if the problem arises again.

PR #6914, opened 2025-04-08
fix : Get rid of Allocatable in m_type of BitCast

### LINKED ISSUE #6805: [stdlib:hashmaps] Length of allocatable array is not remaining deferred on using `transfer` function
MRE :- 
```.f90
module mod
    use iso_fortran_env, only: int8
    implicit none
    integer(int8), allocatable :: val(:)

contains
    subroutine sub( value )
        character(*), intent(in) :: value
        val = transfer(value, val, 1 * len(value))
    end subroutine sub
end module

program main
end program
```

Lfortran Error :- 
```console
ASR verify pass error: Length of allocatable should be deferred (empty).
 --> try.f90:4:5
  |
4 |     integer(int8), allocatable :: val(:)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


```

AnshMehta1:
fix #6805


Pranavchiku (commented):
Add the test: https://github.com/lfortran/lfortran/issues/6805#issuecomment-2839579602 with `--realloc-lhs` flag.

Pranavchiku (approved):

Issue #5418, opened 2024-11-19
spread generates semantically incorrect fortran code


Pranavchiku:
```fortran
program main
        integer :: a(2)
        print *, spread(a, 1, 2)
end program
```

with `--dump-all-passes-fortran` we get

```fortran
! Fortran code after applying the pass: intrinsic_function
program main
implicit none
integer(4), dimension(2, 2) :: temporary_spread
integer(4), dimension(2) :: a
call spread_(a, 1, 2, temporary_spread)
print *, temporary_spread

contains

subroutine spread_(source, dim, ncopies, result)
    integer(4), intent(in) :: dim
    integer(4) :: ele
    integer(4) :: i
    integer(4) :: j
    integer(4) :: k
    integer(4), intent(in) :: ncopies
    integer(4), dimension(2, 2), intent(out) :: result
    integer(4), dimension(:), intent(in) :: source
    i = 1
    if (dim == 1) then
        do j = 1, ubound(source, 1)
            ele = source(j)
            do k = 1, ncopies
                result(i) = ele
                i = i + 1
            end do
        end do
    else
        do j = 1, ncopies
            do k = 1, ubound(source, 1)
                ele = source(k)
                result(i) = ele
                i = i + 1
            end do
        end do
    end if
    return
end subroutine spread_

end program main
```

```console
% gfortran b.f90 && ./a.out
b.f90:25:22:

   25 |                 result(i) = ele
      |                      1
Error: Rank mismatch in array reference at (1) (1/2)
b.f90:33:22:

   33 |                 result(i) = ele
      |                      1
Error: Rank mismatch in array reference at (1) (1/2)
```

tanay-man:
I am getting this as the `pass_fortran_15_intrinsic_function.f90`:
```fortran
! Fortran code after applying the pass: intrinsic_function
program main
implicit none
integer(4) :: __1_k
integer(4), dimension(2, 2) :: __libasr__created__var__0__lcompilers_spread_res
integer(4), dimension(2), save :: a
__1_k = lbound(a, 1)
a(__1_k) = 0
__1_k = __1_k + 1
a(__1_k) = 0
__1_k = __1_k + 1
call _lcompilers_spread(a, 1, 2, __libasr__created__var__0__lcompilers_spread_res)
print *, __libasr__created__var__0__lcompilers_spread_res

contains

subroutine _lcompilers_spread(source, dim, ncopies, result)
    integer(4), intent(in) :: dim
    integer(4) :: ele
[...]

PR #6945, opened 2025-04-09
Update ``spread`` implementation to always produce correct results for any dimensions

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/5418

Implemented https://github.com/lfortran/lfortran/issues/5418#issuecomment-2787424746 (quoted below),

There are two problems with implementation of spread intrinsic in LFortran,

> 1. The shape of resulting array is generated incorrectly (especially if the input is fixed size array) - The correct > way is, size of i-th dimension of resulting array would be defined as `merge(ncopies, merge(original_shape[i], > original_shape[merge(1, i - 1, i == 1)], i < dim), i == dim)`.
>
> 2. The implementation doesn't need a loop, we need to do,
> 
> ```
> do i = lbound(result, selected_dim), ubound(result, selected_dim)
>    result(..., :, :, :, i, :, :, :, ...) = a
> end do
> ```
> Simply copy paste the original array to the output array while iterating over the dim.
>
> Both these are needed to make spread work for general case specified in https://gcc.gnu.org/onlinedocs/gfortran/SPREAD.html. Both are easy to do, I will do it day after tomorrow.

### DIFF
--- a/src/libasr/asr_builder.h
+++ b/src/libasr/asr_builder.h
@@ -295,7 +295,7 @@ class ASRBuilder {
     }
 
     inline ASR::expr_t* ArraySize(ASR::expr_t* x, ASR::expr_t* dim, ASR::ttype_t* t) {
-        return EXPR(make_ArraySize_t_util(al, loc, x, dim, t, nullptr));
+        return EXPR(make_ArraySize_t_util(al, loc, x, dim, t, nullptr, false));
     }
 
     inline ASR::expr_t* Ichar(std::string s, ASR::ttype_t* type, ASR::ttype_t* t) {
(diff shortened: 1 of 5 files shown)


src/libasr/asr_builder.h
```diff
@@ -867,6 +867,8 @@ class ASRBuilder {
     }
 
     ASR::expr_t *ArrayItem_01(ASR::expr_t *arr, std::vector<ASR::expr_t*> idx) {
+        // LCOMPILERS_ASSERT(
+        //     ASRUtils::extract_n_dims_from_ttype(ASRUtils::expr_type(arr)) == idx.size());
```
czgdp1807, line 871: Forgot to uncomment this. :-(.

czgdp1807: But anyways not needed now. Will uncomment in another PR for safety.

Issue #6892, opened 2025-04-07
Progress : Refactor string and string array (LLVM backend)


assem2002:
- [ ] Refactor Non-allocatable string arrays
    - [ ] Use 2D array to represent array of strings ( + null char ).
    - [ ] Fill the whole memory with empty space.
    - [ ]  Modify `Visit_arrayItem` to embed a `\0` char at end of any string ( for now - as we rely on null termination).
    - [ ]  Make print (C stringFormat function) Handle strings without null char by serializing the string length. This is the only that doesn't access string through `arrayItem`.
- [ ]  Refactor allocatable string array
    - [ ]  
- [ ] Start Using strings with no null char (Non Null-terminated)
- [ ] Transform any function return that's string or array of strings into a subroutine to have a proper owner of the return.

assem2002:
**FixedSizeString** :
- Its size is known at compile time and allocated as an array in the back end.
- allocated on the stack as user intended.
- No need to use `alloca` as it may not exist in all back ends.
- It's non-null terminated.
- ASR knows about its length.
- **ASR Verify** should verify that its length is known at compile-time
- **ASR Verify** should verify that it can't be allocatable

**PointerString**: 
- It's an `i8*` that's non-null terminated.
- You can access its size through ASR (either from constant compile-time value or expression). 
- It's an intermediate representation of the strings it can't be a physical type of a variable
- **ASR Verify** should verify that it can't be a type of a variable.

**DescriptorString** : 
- It's a `{i8*, length, capacity}` descriptor
- Used when the length isn't known at compile-time.
- Used with `allocatable` and `non-allocatable` variables. 

**CString** :
- Size isn't a problem with this type, You can fetch it with call to `strlen`
- It's used with c_bind Functions.
- It's **null-terminated**.
- ASR Verify should verify that the variable with CString physicaltype always used with C bind
- Any string variable that's not a c bind shouldn't have a physical type of **CString**.




- `Character(10) :: str` -> fixedSizeArrayString -> `[10 x i8]` + length is constant compile time value
-  `Character(:), alloctable:: str` -> DescriptorString -> `{i8*, length, capacity}`
-  `Character(x), alloctable:: str` -> DescriptorString -> `{i8*, length, capacity}` (Raise an error if allocated with different length variable or constant)
- `character(*) :: str` -> DescriptorString -> `{i8*, length, capacity}`
-  character(x) :: str` ->  DescriptorString -> `{i8*, length, capacity}`
-
[...]

PR #6954, opened 2025-04-10
Refactor : Make string type aware of its length to shift into non-null terminated strings + Push that length to runtime print function 

assem2002:
Towards https://github.com/lfortran/lfortran/issues/6892#issuecomment-2799027443

**Mainly it's concerned about making sure that we know about the length of string and not depending on (C's `strlen` + null-termianted strings) to tell us the length.** as we're moving towards not using non null-terminated strings.

***
**How different physical types know about the length**
- DescriptorString-> it's a descriptor so length is encapsulated along with the string itself
- `PointerString` -> We will treat it as intermediate type of strings, so length is known by calling intrinsic function `len` over the original string.
- `Cstring` is a null-terminated C string, so we fetch the length by calling C's `strlen` function.

***
**What is in this PR**
- Introducing string physical type -> `CString` to only work with CBind character variables (needs more work still).
- Making stringPhysicaltype-> `PointerString` able to get the length of the string if the string isn't compile-time known length.
- also making sure that  `DescriptorString` PhysicalType knows the length if it's being created out of `StringPhysicalCast` from `PointerString` or `CString`
- Tweaking C's runtime printing function to know about the length of the passed string and stop depending on `strlen` to know the length. We have many functions depending on `strlen`, We'll come to those in subsequent PR.
- Tweaking some ASR intrinsic function to use correct string physical type with appropriate length (needs more work still).


assem2002:
Yes

tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.stdout
```diff
@@ -106,6 +106,60 @@
...
+                                                    (String 1 1 () PointerString)
+                                                    ()
+                                                    Source
+                                                    Public
+                                                    Required
+                                                    .false.
+                                                    .false.
+                                                    .false.
+                                                ),
+                                            __libasr_created_variable_5:
```
gxyd, line 146: I'm sorry, why are these new variables now created?

tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.stdout
```diff
@@ -164,7 +218,42 @@
...
+                                                        ()
+                                                        [((StringConstant
+                                                            "build"
+                                                            (String 1 5 () PointerString)
+                                                        ))
+                                                        ((StringConstant
+                                                            ".gitignore"
+                                                            (String 1 10 () PointerString)
+                                                        ))
+                                                        ((Var 7 __libasr_created_variable_3))
```
gxyd, line 233: We probably should be using `__libasr_created_variable_` instead of the newly created `__libasr_created_variable_3`?

I'm a little concerned about newly created temporary variables.

tests/reference/asr-allocate_04-68facec.stdout
```diff
@@ -163,7 +163,11 @@
                                 DescriptorString
                                 PointerString
                                 (Allocatable
-                                    (String 1 -2 () PointerString)
+                                    (String 1 -1 (StringLen
```
gxyd, line 166: So, I'm looking at asr.md file, which briefly explains the meaning of `len` in Character node:
```
**`len`** in Character:
- $>=0$ ... the length of the string, known at compile time
- $-1$ ... character( * ), i.e., inferred at runtime
- $-2$ ... character(:), allocatable (possibly we might use -1 for that also)
- $-3$ ... character(n+3), i.e., a runtime expression stored in `len_expr`

```

So, the reason for change from `-2` to `-1` is because we do really know the length at runtime (once the `len_expr` is computed) and hence the change from `-2` to `-1` for `len` attribute of `Character` node?

gxyd (commented):
Left a few comments.

tests/reference/asr-allocate_04-68facec.stdout
gxyd: Could you please help me understand from the design in this PR, can it be concluded that we can determine the `len` of `String` ASR node depending on `string_physical_type`?

and what will the value of `CString` and `PointerString` 's for `len` be?

assem2002:
I found out that I might need multiple iterations to get strings refactored, and each time a big chunk of the ASR reference test will be changed, hence exhausting review process (not to mention some issues I push that I mean to fix in a subsequent PR).
So maybe I can pile up multiple refactoring PR in one single PR, to speed up the reviewing process. I'll open up another PR that's concerned with mulitple refactorings

assem2002:
was merged within this PR #7048

PR #6985, opened 2025-04-13
Optional args fixed [LP]

swamishiju:
This is a libasr PR change which currently shouldn't effect LF but make optional_args functional again in LP


swamishiju:
@certik  I did put a PR in LP but doesn't this need to be accepted for integration tests to work?

swamishiju:
@certik I couldn't figure out how to do that, could you help?

Issue #6976, opened 2025-04-12
Give a nice error message not Internal Compiler Error for incompatible mod files


difference-scheme:
The following code used to compile, but doesn't do so any more with the latest main branch:
```
module test
contains
   subroutine open_file()     
      integer :: unit
      open(newunit=unit,file='testfile',form='unformatted')      
   end subroutine open_file
end module test
```
LFortran's output:
```
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
LCompilersException: Incompatible format: LFortran Modfile was generated using version '0.48.0-7-g4ef9b6a31-dirty', but current LFortran version is '0.51.0-77-gf4a3246b4-dirty'

```

difference-scheme:
@certik What do you mean? 

If one tries to compile (with the latest main branch) the simple test code that I gave above, then one gets the ICE that I posted. 

There's no recompilation involved in this.

difference-scheme:
Since this ICE looked pretty strange, I got myself a fresh git clone of the lfortran respository, and rebuilt LFortran from scratch. Which, in fact, made the problem vanish.

So it looks as if it was a configuration issue in the previous copy of the repo that I used to build the compiler. I had previously used `git pull` (followed by `./build0.sh`) to upgrade to, and to configure and compile LFortran's latest version of the main branch. But apparently this is not sufficient to do a working recompilation.

I'll always use a fresh `git clone`, instead, from now on. Sorry, my bad. Closing this one.

difference-scheme:
> I think we might be using this Fortran module for the newunit behavior: https://github.com/lfortran/lfortran/blob/059a4a513abca484ca984df1342a0272b1a58a96/src/runtime/custom/lfortran_intrinsic_custom.f90, and it was compiled with an older version of LFortran. I think when you upgraded, it didn't recompile it.

Ah, yes, that would make sense.

> What I do is git clean -dfx when I pull in changes, that way all old stuff is gone, and the build has none of these issues.

Thanks for the hint.

> However the real bug here is that the error message should be better, not an ICE, so I'll keep the issue open.

I see. Ok, then let's keep this open.

PR #6994, opened 2025-04-14
enh: catch LCompilersException while loading mod file and throw nice error message

Pranavchiku:
Towards #6976. I was unable to reproduce #6976 and hence not marking this PR "fixes" it.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -8725,14 +8725,20 @@ public:
 
         SymbolTable *tu_symtab = ASRUtils::get_tu_symtab(current_scope);
 
-        ASR::Module_t *m = ASRUtils::load_module(al, tu_symtab, module_name,
-                loc, true, compiler_options.po, true,
-                [&](const std::string &msg, const Location &loc) {
-                        diag.add(Diagnostic(msg, Level::Error, Stage::Semantic, {Label("", {loc})}));
-                        throw SemanticAbort();
-                    }, lm, compiler_options.separate_compilation
-                );
-
+        ASR::Module_t *m = nullptr;
+        try {
+            m = ASRUtils::load_module(al, tu_symtab, module_name,
+                    loc, true, compiler_options.po, true,
+                    [&](const std::string &msg, const Location &loc) {
+                            diag.add(Diagnostic(msg, Level::Error, Stage::Semantic, {Label("", {loc})}));
+                            throw SemanticAbort();
+                        }, lm, compiler_options.separate_compilation
+                    );
+        } catch ( LCompilersException &e ) {
+            diag.add(Diagnostic(e.msg(),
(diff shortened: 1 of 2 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -8725,14 +8725,20 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    [&](const std::string &msg, const Location &loc) {
+                            diag.add(Diagnostic(msg, Level::Error, Stage::Semantic, {Label("", {loc})}));
+                            throw SemanticAbort();
+                        }, lm, compiler_options.separate_compilation
+                    );
+        } catch ( LCompilersException &e ) {
+            diag.add(Diagnostic(e.msg(),
+                Level::Error, Stage::Semantic, {Label("", {loc})}));
+            throw SemanticAbort();
+        }
```
gxyd, line 8741: What this would mean is that any LCompilersException thrown by `load_module` will raise a semantic error, which I I'm not sure should be the case.

Example, even when: https://github.com/lfortran/lfortran/blob/main/src/libasr/modfile.cpp#L144-L147 is to be raised.

If we do actually think that SemanticError needs to be raised, maybe we should be raising SemanticError in the first place and not raising ICE first and then throwing it as SemanticError.

What do you think?

Pranavchiku: We don't have access to `diagnostics` and no way to `throw SemanticAbort()` in there. Hence I had to catch it where is done in this PR.

gxyd: > We don't have access to `diagnostics` and no way to `throw SemanticAbort()` in there. Hence I had to catch it where is done in this PR.

I don't even think that the error that needs to be reported (for the linked issue) should fall under `SemanticError` at all really.

Pranavchiku: I agree, we can later rename it to anything that suites, this is anyway minor. The point to figure out is how to add the error in diagnostics.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -8725,14 +8725,20 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-
+        ASR::Module_t *m = nullptr;
+        try {
+            m = ASRUtils::load_module(al, tu_symtab, module_name,
+                    loc, true, compiler_options.po, true,
+                    [&](const std::string &msg, const Location &loc) {
+                            diag.add(Diagnostic(msg, Level::Error, Stage::Semantic, {Label("", {loc})}));
+                            throw SemanticAbort();
+                        }, lm, compiler_options.separate_compilation
+                    );
```
certik, line 8736: Cannot you just call this callback inside?

Issue #6773, opened 2025-03-29
[stdlib: logger] handle optional arguments in present correctly


HarshitaKalani:
```fortran
module stdlib_logger    
    implicit none
contains
subroutine add_log_unit(stat )
        integer, intent(out), optional    :: stat
        call validate_unit()
    contains
        subroutine validate_unit()
                if ( present(stat) ) then
                    stat = 5
                end if
        end subroutine validate_unit
    end subroutine add_log_unit
end module
program main
end program
```
```console
$ gfortran b.f90 && ./a.out
$ lfortran b.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2690, in main
    return main_app(argc, argv);
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2618, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 1165, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/home/harshita/Desktop/lfortran/src/lfortran/fortran_evaluator.cpp", line 382, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    compiler_options, run_fn, infile);
[...]

PR #6988, opened 2025-04-13
fix: handle present in nested functions correctly

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/6773

PR #6995, opened 2025-04-14
fix: transform optional arguments and then apply nested_vars pass

HarshitaKalani:
fixes: #6773


HarshitaKalani:
If these changes are fine, we can revert the changes done in this PR: https://github.com/lfortran/lfortran/pull/6988. 
We won't need to look for the variables in the module created by the nested_vars pass.

PR #6998, opened 2025-04-14
fix: import error for `compiler_version` from `iso_fortran_env`

### LINKED ISSUE #4248: Implement compiler_version() and compiler_options() intrinsics

### LINKED ISSUE #4928: Missing interface for compiler_version intrinsic
The intrinsic function `compiler_version` was added in #4428 as a built-in function, but it was not added to the intrinsic module `iso_fortran_env`. Its absence was also noted in https://github.com/lfortran/lfortran/issues/4742.

Here is a test which exposes the lack of interface:
```fortran
module string_mod
    public
contains
    !> Fixed-length string
    function string()
       character(len=6) :: string
       string = "abcdef"
    end function
end module

program test
    use string_mod, only: string

! semantic error: The symbol 'compiler_version' not found in the module 'iso_fortran_env'
    !use, intrinsic :: iso_fortran_env, only: compiler_version
    
    character(len=len(compiler_version())) :: a
    character(len=len(string())) :: b
    character(len=:), allocatable :: c

    a = compiler_version()
    print *, len(a), a
    
    b = string()
    print *, len(b), b
    
    c = compiler_version()
    print *, len(c), c

end program
```

The module `iso_fortran_env` (https://github.com/lfortran/lfortran/blob/main/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90) should probably contain something like:
```fortran
interface
   function compiler_version() bind(c,"_lfortran_compiler_version")
      character(len=????) :: compiler_version
   end function
   function compiler_options() bind(c,"_lfortran_compiler_options")
      character(len=????) :: compiler_options
   end function
end interface
```

The `compiler_options` functions requires special handling, because the length of those come from the actual command-line (sum over a subset of `argv[i]`).

### LINKED ISSUE #6553: The symbol 'compiler_options' not found in the module 'iso_fortran_env'
`use iso_fortran_env, only: compiler_options` doesn't work:
```
$ lfortran binop_02.f90
semantic error: The symbol 'compiler_options' not found in the module 'iso_fortran_env'
 --> binop_02.f90:2:32
  |
2 |     use iso_fortran_env, only: compiler_options
  |                                ^^^^^^^^^^^^^^^^
```

As a workaround, one must use `use iso_fortran_env`.

### LINKED ISSUE #6731: [stdlib:selection] compiler_version not found in the module 'iso_fortran_env'
```fortran
program main
  use, intrinsic :: iso_fortran_env, only: compiler_version
  print*, compiler_version()
end program
```
```console
$ lfortran a.f90
semantic error: The symbol 'compiler_version' not found in the module 'iso_fortran_env'
 --> a.f90:2:44
  |
2 |   use, intrinsic :: iso_fortran_env, only: compiler_version
  |                                            ^^^^^^^^^^^^^^^^ 


```

### LINKED ISSUE #6981: The symbol 'compiler_version' not found in the module 'iso_fortran_env'
```
program main
use, intrinsic :: iso_fortran_env, only: &
   compiler_version, compiler_options
end program
```
Gives
```
semantic error: The symbol 'compiler_version' not found in the module 'iso_fortran_env'
   --> a.f90:128:4
    |
128 |    compiler_version, compiler_options
    |    ^^^^^^^^^^^^^^^^
```
I thought we already fixed it, but maybe not this case.

Pranavchiku:
Fixes #6981. Fixes #6731. Fixes #6553. Fixes #4928. Closes #4248.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1158,7 +1158,7 @@ public:
         {"dim", IntrinsicSignature({"X", "Y"}, 2, 2)},
         {"selected_real_kind", IntrinsicSignature({"p", "r", "radix"}, 0, 3)},
         {"nearest", IntrinsicSignature({"x", "s"}, 2, 2)},
-        {"lfortran_compiler_version", IntrinsicSignature({}, 0, 0)},
+        {"_lfortran_compiler_version", IntrinsicSignature({}, 0, 0)},
         {"compiler_options", IntrinsicSignature({}, 0, 0)},
         {"command_argument_count", IntrinsicSignature({}, 0, 0)},
         {"ishftc", IntrinsicSignature({"i", "shift", "size"}, 2, 3)},
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1158,7 +1158,7 @@ public:
         {"dim", IntrinsicSignature({"X", "Y"}, 2, 2)},
         {"selected_real_kind", IntrinsicSignature({"p", "r", "radix"}, 0, 3)},
         {"nearest", IntrinsicSignature({"x", "s"}, 2, 2)},
-        {"compiler_version", IntrinsicSignature({}, 0, 0)},
+        {"lfortran_compiler_version", IntrinsicSignature({}, 0, 0)},
         {"compiler_options", IntrinsicSignature({}, 0, 0)},
         {"command_argument_count", IntrinsicSignature({}, 0, 0)},
         {"ishftc", IntrinsicSignature({"i", "shift", "size"}, 2, 3)},
--- a/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
+++ b/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
@@ -23,7 +23,7 @@ integer, parameter :: iostat_end = -1
 contains
 function compiler_version() result(version)
     character(len=:), allocatable :: version
-    version = lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
+    version = _lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
 end function compiler_version
 
 
--- a/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
+++ b/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
@@ -22,7 +22,7 @@ integer, parameter :: iostat_end = -1
 
 contains
 function compiler_version() result(version)
-    character(len=100) :: version
+    character(len=:), allocatable :: version
     version = lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
 end function compiler_version
 
--- a/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
+++ b/src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
@@ -20,4 +20,11 @@ integer, parameter :: logical_kinds(1) = [4]
 
 integer, parameter :: iostat_end = -1
 
+contains
+function compiler_version() result(version)
+    character(len=100) :: version
+    version = lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
+end function compiler_version
+
+
 end module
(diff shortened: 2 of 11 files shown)


src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
```diff
@@ -20,4 +20,11 @@ module lfortran_intrinsic_iso_fortran_env
 
 integer, parameter :: iostat_end = -1
 
+contains
+function compiler_version() result(version)
+    character(len=100) :: version
```
gxyd, line 25: I think this should be `allocatable` instead of a fixed length `100`, because if it's fixed length, the Fortran standard will necessitate it to have succeeding empty space, which will make it look odd.

E.g.; something like:
```f90
program main
   use, intrinsic :: iso_fortran_env, only: compiler_version
   print *, 'This file was compiled by ', compiler_version(), ' using some options '
end program

```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1158,7 +1158,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         {"dim", IntrinsicSignature({"X", "Y"}, 2, 2)},
         {"selected_real_kind", IntrinsicSignature({"p", "r", "radix"}, 0, 3)},
         {"nearest", IntrinsicSignature({"x", "s"}, 2, 2)},
-        {"compiler_version", IntrinsicSignature({}, 0, 0)},
+        {"lfortran_compiler_version", IntrinsicSignature({}, 0, 0)},
```
gxyd, line 1161: I'm a little concerned about use of `lfortran_` prefix here, as this would almost make it an exception to this rule, all other checks which are done just use the intrinsic name, so I'm a little concerned about the prefix here.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
```diff
@@ -20,4 +20,11 @@ module lfortran_intrinsic_iso_fortran_env
...
+    version = lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
```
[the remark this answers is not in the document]
gxyd: Do you see any problems with this way?

gxyd (commented):
Left a few comments.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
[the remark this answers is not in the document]
Pranavchiku: > I'm a little concerned about use of lfortran_ prefix here, as this would almost make it an exception to this rule, all other checks which are done just use the intrinsic name, so I'm a little concerned about the prefix here.

Haha, you answered it.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: What is your say on the way forward?

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
Pranavchiku: Agreed, can you add a suggestion and change the definition for better understanding?

gxyd: We can't have a fixed length, we should make it `allocatable` and that's it, i.e.:
```f90
function compiler_version() result(version)
    character(:), allocatable :: version
    version = lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
end function compiler_version
```

that will always ensure that it's assigned a length which `lfortran_compiler_version()` is exactly equal to at runtime.

src/lfortran/semantics/ast_common_visitor.h
gxyd: Was there any issue with use of just `compiler_version` (i.e. without any prefix)?

Pranavchiku: Yes, here are the set of issues:

> Fixes https://github.com/lfortran/lfortran/issues/6981. Fixes https://github.com/lfortran/lfortran/issues/6731. Fixes https://github.com/lfortran/lfortran/issues/6553. Fixes https://github.com/lfortran/lfortran/issues/4928. Closes https://github.com/lfortran/lfortran/issues/4248.

Pranavchiku:
@gxyd applied suggestion, can you please review again?

gxyd (approved):
The changes look good.

The below imperfections can possibly be addressed later:
* presence of `compiler_version` intrinsic as a part of elemental function (it's an inquiry function)
* use of `lfortran_` prefix

I don't know the solution to either of them above currently.

Pranavchiku:
let's not merge this, I'll wait for it to be discussed and reviewed in our meeting tonight with @certik

PR #7007, opened 2025-04-14
enh: initial implementation advance keyword for fileread

### LINKED ISSUE #7000: `iostat` for end of record is not proper
```fortran
program test
integer :: io, stat
character(len=4096) :: buffer
open(newunit=io, status="scratch")
write(io, "(a)") repeat("abc", 10), repeat("def", 100), repeat("ghi", 1000)
rewind(io)
read(io, '(a)', advance='no', iostat=stat) buffer
print *, stat
close(io)
end program
```

```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
0
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran b.f90 && ./a.out
          -2
```

jinangshah21:
Fixes #7000


gxyd:
Could you please help understand how is `advance` keyword related to `end` keyword?

From my understand, what it seems like you are doing in this PR is, when `advanced="no"` is specified in the user call, you set `end=""` (empty string), while when `advance="yes"` you are setting `end="\n"` (newline character).

I see that there is possibility of presence of both `advance` and `end` keyword in `FileRead` (see screenshot below from 24-007.pdf J3 documentation):
Screenshot 2025-04-15 at 12 39 42 PM


If you can kindly refer me some documentation for how `advance` and `end` are set and related to each other.

Also, I don't see any test case for `advance="yes"`, so we probably need that as well.

jinangshah21:
Thanks @gxyd for reference, I was not aware of `end` keyword in read. I made a seperate argument for advance and added test cases for `advance="yes"` also.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3905,7 +3905,7 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
             free(buffer);
             return;
         } else {
-            if (streql(buffer, "\n")) {
+            if (streql(buffer, "\n") || (streql(advance, "no") && strcspn(buffer, "\n") != n)) {
```
gxyd, line 3908: What does the condition `strcspn(buffer, "\n") != n` really imply for our code?

I know what it does in `C` but I don't really know what it really means for LFortran's functionality here.

src/libasr/ASR.asdl
```diff
@@ -48,7 +48,7 @@ stmt
     | Print(expr text)
     | FileOpen(int label, expr? newunit, expr? filename, expr? status, expr? form, expr? access, expr? iostat, expr? iomsg)
     | FileClose(int label, expr? unit, expr? iostat, expr? iomsg, expr? err, expr? status)
-    | FileRead(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? size, expr? id, expr* values, stmt? overloaded, bool is_formatted)
+    | FileRead(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? advance, expr? size, expr? id, expr* values, stmt? overloaded, bool is_formatted)
```
gxyd, line 51: If the value of `advance` is always going to be either "yes" or "no" then why don't we set it as a boolean value? Dealing with boolean values is easier than character strings and fast comparisons in our runtime library
from pg 245 of J3/24-007, I've attached a screenshot

Screenshot 2025-04-15 at 6 18 56 PM

src/libasr/runtime/lfortran_intrinsics.c
jinangshah21: This condition just checks if `\n` is present in string or not. 
Consider if I am reading with `advance="no"` to varible of length 2. But line in file has length 10. So in this case we have not reached End of line. So for this case we need to check if `\n` is present in buffer or not.

jinangshah21: I have also added comment in code

Issue #6978, opened 2025-04-12
lfortran lacks ability to generate object files


kargl:
Here's a weird one.  I have no idea how to describe the outcome. 
I've built lfortran for FreeBSD, copied my m90 library to a new
directory, and altered the Makefile to use lfortran.  Upon excluding
the files that give lfortran heartburn building the library and ,  I

    %../work/bin/lfortran --version
    LFortran version: 0.51.0-77-gf4a3246b4-dirty
    Platform: FreeBSD
    LLVM: 15.0.7
    Default target: x86_64-portbld-freebsd15.0
    % cd m90
    % make
    % ls -l lu3m*
    -rw-r--r--  1 kargl kargl  1925 Feb  9 08:59 lu3m.F90
    -rw-r--r--  1 kargl kargl 22293 Apr 12 10:40 lu3m.mod
    -rw-r--r--  1 kargl kargl   480 Apr 12 10:40 lu3m.o
    -rw-r--r--  1 kargl kargl   435 Feb  9 08:59 lu3mi.fin
    -rw-r--r--  1 kargl kargl   656 Feb  9 08:59 lu3mr.fin
    % objdump -d lu3m.o

    lu3m.o:     file format elf64-x86-64-freebsd

    % ls -l | grep 480 | wc -l
      57

That is, there are 57 empty object files.  The Makefile causes the following
compiler invocation is 

    % rm lu3m.o
    % make
    ../work/bin/lfortran -c --cpp -I. --no-style-warnings --no-warnings lu3m.F90

Is there some other magic compiler option needed to actually cause code
generation?

PS: adding the `--show-asm` to the compiler options leads to

     % rm lu3m.o
     % make
     ../work/bin/lfortran -c --cpp -I. --show-asm --no-style-warnings --no-warnings  lu3m.F90
     Internal Compiler Error: Unhandled exception
     Traceback (most recent call last):
     LCompilersException: visit_DoLoop() not implemented
     *** Error code 1

PPS:  Just found the ` --generate-object-code` option.  That seems to be needed to
to actually cause code generation.   One might expect the `-c` to imply this option
to give the behavior one sees with every other compiler on the planet.

kargl:
Well, the `--generate-object-code` is a bust.  After adding it to the Makefile build infrastructure,

    .SUFFIXES: .c .F90 .f90 .mod .fin

    .F90.o:
	${F95} -c --cpp --generate-object-code ${FFLAGS} ${MODPATHS} $<

[...]

Issue #6894, opened 2025-04-07
Separate compilation design


certik:
When separate compilation mode is enabled via a compiler flag, do:

* store the flag in compiler_options
* when "using" a module that must be loaded from a `.mod` file, we still load it to ASR, but mark everything in the module as Interactive, via the `SymbolTable::mark_all_variables_external()` function (I think we first load the given module to a separate TranslationUnit, so you can probably just call this function on this unit).
* The same with main program that uses modules --- they will all be marked as external

Idea: possibly when saving `.mod` files we can already mark it as external, then we avoid fixing it when loading it.

To test it, add it to our integration_tests, as separate files and ensure they can be linked. I think we can probably run our all testsuite with `--separate-compilation` just like we run it with `--std=f23`. It should work either way. We already have tests for modules. Initially not everything will be working, so just add a label (or reverse label, whatever is easier) to only run a subset of tests with separate compilation, and we will keep increasing the coverage until everything works.

certik:
@Pranavchiku let's document how it works. Couple questions. 

* If we have global subroutines and use separate compilation, then it works. Correct?
* If we have global subroutines and use direct compilation, what exactly happens --- do we generate any object code for them, or will they not work? What exactly is the status there.
* The separate compilation and direct compilation do not mix, right? That is, you either have to compile the whole project with separate compilation, or no files at all?
* To consider: rename `--generate-object-code` to `--separate-compilation`. Not sure yet.

We have several ways forward for the second option. One way is to not support it, in which case we should give an error message, saying to enable separate compilation. Another is to automatically generate object code just for those subroutines.

After the above is clear, let's document this somewhere exactly how it works.

certik:
We can now close this, separate compilation fully works.

PR #7011, opened 2025-04-15
feat: auto infer global procedures and set separate compilaton to true

Pranavchiku:
Towards #6894. With this we have #6978 working as well.


Pranavchiku:
This is high priority if you can do it now, I'll have time to make changes.

gxyd:
> This is high priority if you can do it now, I'll have time to make changes.

I'm really sorry, I wouldn't be able to look today, but it's ok to merge it, I'll take a look later and come back with any comments I've. I couldn't understand how separate compilation works.

PR #7012, opened 2025-04-15
Revert 6988

HarshitaKalani:
https://github.com/lfortran/lfortran/pull/6995#issuecomment-2802653085


gxyd (approved):
Looks good to me.

gxyd:
One thing I would like to point is to not just use "Revert 6988" in commit message, provide a little more information.

gxyd:
> > One thing I would like to point is to not just use "Revert 6988" in commit message, provide a little more information.
> 
> You could have fixed it before merging. :)

Yes, I also made a mess of it, it was the moment right after merge I realized that. 😞

PR #7019, opened 2025-04-17
enh: introduce `c_null_funptr` to intrinsic `iso_c_binding`

### LINKED ISSUE #6864: Missing c_null_funptr in iso_c_binding
While working with iso_c_binding ,  I noticed that c_null_funptr is not declared in LFortran.
This leads to a semantic error during compilation, even though it is a standard as per Fortran 2003.

MRE : 
```
program test_null_funptr
  use iso_c_binding
  implicit none

  type(c_funptr) :: fptr
  fptr = c_null_funptr

  print *, "Null function pointer assigned successfully."
end program test_null_funptr
```

Error : 
```
semantic error: Variable 'c_null_funptr' is not declared
 --> ../codes/test/c_null_funptr.f90:6:10
  |
6 |   fptr = c_null_funptr
  |          ^^^^^^^^^^^^^ 'c_null_funptr' is undeclared
```

Expected Behaviour (gfortran result)  : 
`
Null function pointer assigned successfully.
`

Pranavchiku:
Fixes #6864.

### DIFF
--- a/tests/reference/asr-c_ptr_02-fce1b0e.stdout
+++ b/tests/reference/asr-c_ptr_02-fce1b0e.stdout
@@ -252,6 +252,16 @@
                                     c_null_char
                                     Public
                                 ),
+                            c_null_funptr:
+                                (ExternalSymbol
+                                    2
+                                    c_null_funptr
+                                    4 c_null_funptr
+                                    lfortran_intrinsic_iso_c_binding
+                                    []
+                                    c_null_funptr
+                                    Public
+                                ),
                             c_null_ptr:
                                 (ExternalSymbol
                                     2
--- a/tests/reference/asr_openmp-openmp_39-aaf2ba8.stdout
+++ b/tests/reference/asr_openmp-openmp_39-aaf2ba8.stdout
@@ -219,6 +219,16 @@
                                     c_null_char
                                     Public
                                 ),
+                            c_null_funptr:
+                                (ExternalSymbol
+                                    4
+                                    c_null_funptr
+                                    17 c_null_funptr
+                                    lfortran_intrinsic_iso_c_binding
+                                    []
+                                    c_null_funptr
+                                    Public
+                                ),
                             c_null_ptr:
                                 (ExternalSymbol
                                     4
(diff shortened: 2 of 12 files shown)


tests/reference/asr-c_ptr_02-fce1b0e.stdout
```diff
@@ -252,6 +252,16 @@
...
+                                (ExternalSymbol
+                                    2
+                                    c_null_funptr
+                                    4 c_null_funptr
+                                    lfortran_intrinsic_iso_c_binding
+                                    []
+                                    c_null_funptr
+                                    Public
+                                ),
                             c_null_ptr:
```
gxyd, line 265: this indicates that we should just modify `integration_tests/c_ptr_02.f90` to only import the necessary variables

Ofcourse one solution is to let this ASR change be as it is OR fix it in `integration_tests/c_ptr_02.f90`

tests/reference/asr_openmp-openmp_39-aaf2ba8.stdout
```diff
@@ -983,6 +993,16 @@
...
+                            c_null_funptr:
+                                (ExternalSymbol
+                                    2
+                                    c_null_funptr
+                                    17 c_null_funptr
+                                    lfortran_intrinsic_iso_c_binding
+                                    []
+                                    c_null_funptr
+                                    Public
+                                ),
```
gxyd, line 1005: Similarly, we need to fix `omp_lib` to only import what's needed from `iso_c_binding`.

We either fix it now, or if we don't do it now, then in the future these kind of spurious changes to ASR will pop up again when do implement something like `C_F_PROCPOINTER` etc. (and many more).

gxyd (approved):
Pretty much happy with the changes, upto you to take a call on how you would like to address the spurious ASR changes we see in this PR.

tests/reference/asr_openmp-openmp_39-aaf2ba8.stdout
Pranavchiku: I agree, let's open an issue for this.

Another POV: These changes serve as a validation that the change we made is correct and we didn't miss anything. Although we do add a test corresponding to the fix.

Pranavchiku: I'll not block this PR and prefer to address in subsequent PRs.

gxyd: > Another POV: These changes serve as a validation that the change we made is correct and we didn't miss anything. Although we do add a test corresponding to the fix.

Exactly what I thought, then I thought that maybe we can keep it one of the integration test cases, but remove it from `omp_lib.f90`.

gxyd: > I agree, let's open an issue for this.

Opened it here: https://github.com/lfortran/lfortran/issues/7021

PR #7037, opened 2025-04-19
fix: logical data type

Th3C0d3Mast3r:
## ERROR BEFORE CHANGES:-
Before making the changes, when tried to run a boolean input using lfortran, there was an error file, as shown in the image
image

## POST CHANGES:-
After making the `case` for logical datatype in the `asr_to_llvm.cpp` file, and even adding the `logical read` in `intrinsics.c` (along with adding its function in the header file `intrinsics.h`, there is a proper read for logical type, with error handling when wrong input is entered.
image

If any changes needed, please do let me know
Thank you.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8207,6 +8207,32 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
                 }
                 break;
             }
+            // adding case of boolean type input (TO DO:- boolean array)
+            case (ASR::ttypeType::Logical):{
+                std::string runtime_func_name;
+                llvm::Type *type_arg;
+                int a_kind = ASRUtils::extract_kind_from_ttype_t(type);
+
+                if (a_kind == 4) {
+                    runtime_func_name = "_lfortran_read_logical";
+                    type_arg = llvm::Type::getInt1Ty(context); // LLVM boolean type (1 bit)
+                } else {
+                    throw CodeGenError("Read Logical function not implemented for kind: " + std::to_string(a_kind));
+                }
+
+                fn = module->getFunction(runtime_func_name);
+                if (!fn) {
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8207,6 +8207,32 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
...
+                            llvm::Type::getVoidTy(context), {
+                                type_arg->getPointerTo(),
+                                llvm::Type::getInt32Ty(context)
+                            }, false);
+                    fn = llvm::Function::Create(function_type,
+                            llvm::Function::ExternalLinkage, runtime_func_name, *module);
+                }
+                break;
+            }
+
```
gxyd, line 8235: Could you provide test cases, which allow us to now correctly compile valid Fortran program because of the changes in this PR?

tests/errors/floatStringInput.f90
```diff
@@ -0,0 +1,15 @@
...
+    do
+        read(10, *, end=100) x
+        print *, "Read real number:", x
+    end do
+
+100 continue
+    close(10)
+    stop
+
+end program realStringInput
```
gxyd, line 15: I'm not sure if this test program follows the naming convention that we use for test files in LFortran, please see the how the naming is done for other files present in `tests/error/` directory. None of them follow camel case naming for test programs.

gxyd (commented):
It's a little difficult to properly review the PR without some test cases which are valid Fortran program, I only see addition of programs which leads to error with your changes.

gxyd:
Kindly use a more descriptive PR title.

gxyd:
In the PR description you've added screenshots of compiling and run them with LFortran with a different behavior on *main* vs this PR, maybe that can act as test case?

gxyd:
The failure observed in CI for windows, seems like a genuine failure.

Issue #6736, opened 2025-03-25
[stdlib: bitsets] Handle calling procedure for extended type


jinangshah21:
```fortran
module bit_type 
    type, abstract :: bitset_type
    contains
        procedure(from_string_abstract), deferred, pass(self)   :: from_string 
    end type bitset_type
    type, extends(bitset_type) :: bitset_large
    contains
        procedure, pass(self)  :: from_string => from_string_large
    end type bitset_large

    abstract interface
        subroutine from_string_abstract(self)
            import :: bitset_type
            class(bitset_type), intent(out) :: self
        end subroutine from_string_abstract
    end interface 

   contains

    subroutine from_string_large(self)
        class(bitset_large), intent(out) :: self
        print *, "hi"
    end subroutine from_string_large
end module

program test
    use bit_type
    type(bitset_large) :: b
    call b%from_string()
end program
```

Lfortran outptuts nothing, while it shoud print `hi`.

xaerru:
Working on this.

PR #7038, opened 2025-04-19
fix: subroutine runtime polymorphism for derived types declared in a program

xaerru:
Fix https://github.com/lfortran/lfortran/issues/6736

We now create a vtab for all `StructType` and `ClassType` variables. Earlier we were creating a vtab only for `ClassType` variables.

We also pass the polymorphic arg to `visit_RuntimePolymorphicSubroutineCall`. Earlier it was assumed that `x.m_dt` is a `ClassType` but it can be a `StructType`.


HarshitaKalani (approved):
Looks good to me! We'll need to resolve the conflicts.

xaerru:
Second commit is fixing a bug I found, added the corresponding test in the same commit.

Issue #7044, opened 2025-04-21
CI: `semantic error: Module 'modules_09_b' not declared in the current source and the modfile was not found`


certik:
For some reason the `modules_09_b` is not found in https://github.com/lfortran/lfortran/actions/runs/14565834999/job/40854922061?pr=6850:
```
-- Build files have been written to: /home/runner/work/lfortran/lfortran/integration_tests/test-llvm_sc
[  1%] Generating arrays_50_mod.o
[  3%] Generating modules_22_module.o
[  5%] Generating lcompilers_test_module.o
[  8%] Generating modules_07_module.o
[ 10%] Generating modules_23_module.o
[ 10%] Generating modules_08_a.o
[ 12%] Generating modules_08_b.o
[ 14%] Generating modules_09_b.o
[ 16%] Generating modules_09_a.o
[ 17%] Building Fortran object CMakeFiles/modules_07.dir/modules_07.f90.o
[ 19%] Building Fortran object CMakeFiles/modules_08.dir/modules_08.f90.o
semantic error: Module 'modules_09_b' not declared in the current source and the modfile was not found
 --> /home/runner/work/lfortran/lfortran/integration_tests/modules_09_a.f90:2:1
  |
2 | use modules_09_b, only: b, i
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
make[2]: *** [CMakeFiles/modules_09.dir/build.make:74: modules_09_a.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```
It only happens sometimes.

Given that this is the `llvm_sc` tag, I wonder if it has something to do with @Pranavchiku's recent refactoring.

HarshitaKalani:
Yeah, the same issue was observed earlier: https://github.com/lfortran/lfortran/issues/7023

PR #7052, opened 2025-04-22
test: reorder modules_09_b to generate object files for separate compilation

Pranavchiku:
Potential fix for #7044


Pranavchiku:
With `llvm_sc` enabled, files are compiled separately and hence the ordering matters unlike the direct compilation where we can do `lfortran a.f90 b.f90 c.f90`.

Pranavchiku:
We don't need this now, ported this as part of https://github.com/lfortran/lfortran/pull/7076

Issue #6700, opened 2025-03-20
tracking issue for compiling POT3D with --std=f23


gxyd:
## Description

This is a tracking issue for compiling POT3D with `--std=f23` flag, the current error is:
```console
...
++ uname
+ [[ Darwin == \L\i\n\u\x ]]
+ CC=clang
+ cd src
+ clang -I/Users/gxyd/miniforge3/envs/lf_ci_pot3d_llvm11/include -c mpi_wrapper.c
+ /Users/gxyd/OpenSource/lfortran/pot3d/../src/bin/lfortran --std=f23 -c mpi_c_bindings.f90
+ /Users/gxyd/OpenSource/lfortran/pot3d/../src/bin/lfortran --std=f23 -c mpi.f90
+ /Users/gxyd/OpenSource/lfortran/pot3d/../src/bin/lfortran --std=f23 -c psi_io.f90 --no-style-warnings --no-warnings
+ /Users/gxyd/OpenSource/lfortran/pot3d/../src/bin/lfortran --std=f23 -c --cpp --implicit-interface pot3d.F90 --no-style-warnings --no-warnings
semantic error: Variable 'm' is not declared
    --> pot3d.F90:4745:21
     |
4745 |       integer :: JA(M),IA(N+1),A_da(N),iw(N)
     |                     ^ 'm' is undeclared


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.

```

### Issues

- [ ] https://github.com/lfortran/lfortran/issues/6699

PR #7060, opened 2025-04-23
CI: enable testing POT3D with --std=f23 standard at the CI

gxyd:
## Description

Once the PR: https://github.com/lfortran/lfortran/pull/7059 is accepted and merged, and this PR is rebased on top of *main*, the CI for this PR will also pass.

Fixes https://github.com/lfortran/lfortran/issues/6700


ci/test_third_party_codes.sh
```diff
@@ -133,6 +133,9 @@ time_section "🧪 Testing POT3D" '
   print_subsection "Building with optimization flags"
   FC="$FC --fast --skip-pass=dead_code_removal" ./build_and_run.sh
 
+  print_subsection "Building with optimization flags and f23 standard"
+  FC="$FC --std=f23 --fast --skip-pass=dead_code_removal" ./build_and_run.sh
```
gxyd, line 137: I'm wondering, along with this, should we enable `--std=f23` with default flags as well? 

The only reason, I didn't do this is because, if it adds a good value or not, considering testing all third party repositories with their fast and no-fast versions both with `--std=f23` will increase the time for CI testing.

But, if we skip it, it might help us catch a bug which otherwise we might introduce.

gxyd:
Closing this PR in favor of https://github.com/lfortran/lfortran/pull/7384

PR #7065, opened 2025-04-23
feat: enable separate compilation using `--generate-object-code`

Pranavchiku:
Towards #6894.


Pranavchiku:
The following commit is now used which fixed the LAPACK failure: https://github.com/Pranavchiku/lapack/commit/2e2144d65348b83423d4ebaf4ce26a378f80ab0c

PR #7077, opened 2025-04-26
(a) format supported in formatted_read

HiteshGhanchi:
In the following program:
```
program test_stdin
    implicit none
    character(len=10) :: a
    read(*,'(a)') a
    print *, "From stdin:", a
end program
```
an initial runtime error occurred:
`No file found with given unit`
This issue happened because, after handling the stdin case (unit_num == -1), the control flow continued into the file input handling section.
Since no valid file pointer existed for unit_num == -1, the program exited with the above error.

The root cause was the absence of a return statement after successfully reading from stdin.
Without an explicit return, execution proceeded further unnecessarily, resulting in the error.

Adding a return after completing the stdin read ensures that the program exits correctly from the _lfortran_formatted_read function and does not attempt further file operations.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -3997,7 +3997,9 @@ LFORTRAN_API void _lfortran_formatted_read(int32_t unit_num, int32_t* iostat, in
             strcpy(*arg, buffer);
             va_end(args);
             free(buffer);
+            return;
         }
+
     }
 
     bool unit_file_bin;


HiteshGhanchi:
Sure

HiteshGhanchi:
@certik This case handles user input, and runtime error for that. But in tests.toml i chose the option of run, but how do I handle this user input case, cause taking input from some file will not work in this case, and I saw tests are taking input fro m txt files.

jinangshah21:
```fortran 
program detect_input
    use iso_c_binding
    implicit none
    character(len=21) :: line
    interface
        function isatty(fd) bind(C)
            import :: c_int
            integer(c_int), value :: fd
            integer(c_int) :: isatty
        end function
    end interface

    if (isatty(0_c_int) == 1) then
        call execute_command_line("lfortran a.f90 < a.f90")
    else
        read(*, '(a)') line
        if (trim(line) /= "program detect_input") error stop
    end if
end program
```

We can do something like this to test `stdin`. Here we pass program file itself as `stdin`.

HiteshGhanchi:
@certik @jinangshah21 The approach fails some checks in CI 
My test works locally because stdin is an interactive terminal (isatty(0)=1). But in CI/CD (GitHub Actions), I think stdin is non-interactive (isatty(0)=0), so the test fails because the program expects input differently. I need advice on how we want to handle stdin simulation in CI properly.
In test.toml, I tried adding stdin and passing arguments through it, but it is not being recognized.
I think the best approach will be changes in run_tests.py to handle stdin cases too.
This will help in future tescases too.

jinangshah21:
You need to add test in `intergration_tests` folder and register in `CMakeLists` present in that folder.

jinangshah21:
You didin't added test file in `integration_tests` folder. Please add that.

HiteshGhanchi:
Added it now.

integration_tests/formatted_read_1.f90
```diff
@@ -0,0 +1,24 @@
...
+            import :: c_int
+            integer(c_int), value :: fd
+            integer(c_int) :: isatty
+        end function
+    end interface
+
+    istty = isatty(0_c_int)
+
+    if (istty == 1) then
+        call execute_command_line("./src/bin/lfortran ./tests/formatted_read_1.f90 < ./tests/formatted_read_1.f90")
```
jinangshah21, line 18: File path is incorrect.

HiteshGhanchi: How do we run this integration test locally ? What is the command for it

jinangshah21: `./integration_tests/run_test.py`

integration_tests/formatted_read_1.f90
```diff
@@ -0,0 +1,24 @@
...
+            import :: c_int
+            integer(c_int), value :: fd
+            integer(c_int) :: isatty
+        end function
+    end interface
+
+    istty = isatty(0_c_int)
+
+    if (istty == 1) then
+        call execute_command_line("./src/bin/lfortran ./integration_tests/formatted_read_1.f90 < ./tesintegration_teststs/formatted_read_1.f90")
```
jinangshah21, line 18:
```suggestion
        call execute_command_line("./src/bin/lfortran formatted_read_1.f90 < formatted_read_1.f90")
```

HiteshGhanchi:
sorry i didnt read that , my bad

jinangshah21:
Refer `integration_tests/file_27.f90`, and execute command (like `gfortran a.f90 && ./a.out < a.f90`) for GFortran to ensure it works for GFortran also. 
Also you can disable `llvm_wasm llvm_wasm_emcc` from tests for now.

jinangshah21:
@certik, locally tests works for GFortran but on CI it doesn't run, how to test at CI?

HiteshGhanchi:
@jinangshah21 @certik any update regarding this ?

jinangshah21:
@certik can we ignore Gfortran for this one as GFortran is working locally and also this is small change and valid change.

integration_tests/CMakeLists.txt
```diff
@@ -2201,3 +2201,5 @@ RUN(NAME separate_compilation_02 LABELS gfortran llvm llvm_goc EXTRAFILES separa
 
 RUN(NAME no_explicit_return_type_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc
     EXTRA_ARGS --implicit-typing)
+
+RUN(NAME formatted_read_1 LABELS gfortran llvm)
```
gxyd, line 2205: We aren't using `--cpp` flag appropriately, we need to use `EXTRA_ARGS` as `--cpp` for preprocess C preprocessor directive.

integration_tests/formatted_read_1.f90
```diff
@@ -0,0 +1,33 @@
...
+
+#ifdef __GFORTRAN__
+    is_gfortran = .true.
+#endif
+
+    istty = isatty(0_c_int)
+
+    if (istty == 1) then
+        if (is_gfortran) then
+            call execute_command_line("gfortran formatted_read_1.f90 && ./a.out < formatted_read_1.f90")
```
gxyd, line 24: I don't like the idea of using a default name `a.out`, maybe something else needs to be used.

And again, the path needs to be fixed, see the other comment

integration_tests/formatted_read_1.f90
```diff
@@ -0,0 +1,33 @@
...
+            integer(c_int), value :: fd
+            integer(c_int) :: isatty
+        end function
+    end interface
+
+#ifdef __GFORTRAN__
+    is_gfortran = .true.
+#endif
+
+    istty = isatty(0_c_int)
```
gxyd, line 20: can we please print the value of "istty"? I will help us see the output both when running at the CI and at terminal
i.e.

```f90
    print "istty: ", istty
```

that should do.

integration_tests/formatted_read_1.f90
gxyd: We aren't passing `-cpp` here? Maybe we don't need to?

HiteshGhanchi: If i dont pass --cpp here in lfortran, i get error, as we are using #ifdef and such , hence they need this --cpp flag for processing it

integration_tests/formatted_read_1.f90
```diff
@@ -0,0 +1,33 @@
...
+
+    if (istty == 1) then
+        if (is_gfortran) then
+            call execute_command_line("gfortran formatted_read_1.f90 && ./a.out < formatted_read_1.f90")
+        else
+            call execute_command_line("./src/bin/lfortran --cpp formatted_read_1.f90 < formatted_read_1.f90")
+        end if
```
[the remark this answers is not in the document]
HiteshGhanchi: Assuming I am in integration_test folder this time , and i will add paths according to it this time.
Should work I suppose.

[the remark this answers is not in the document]
HiteshGhanchi: I did this based on the advice given by jinang.
I used the code snippet code he gave to simulate user input.

HiteshGhanchi:
@gxyd
I believe the issue is that in CI environments like GitHub Actions, there's no interactive terminal , it's just running in the background. So isatty(0) always returns 0, whereas locally it returns 1, which is why things work fine on our terminals.

Because of this, the program tries to read from stdin in CI, expecting user input, but hits EOF immediately since there's nothing to read — leading to a runtime error.

I couldn’t find any existing test case where user input via stdin is actually tested. However, my changes require testing that interaction. Also, as @certik mentioned, using execute_command_line isn't ideal here.

I think the better solution might be to modify run_tests.py to allow passing custom input to stdin —,so we can test interactive behavior for future testcases too.

HiteshGhanchi:
So basically it only manifests when reading interactively, due to lack of return statement.

HiteshGhanchi:
@certik Should I remove the tests and push again with just the initial change?

gxyd:
Sorry to have kept a hold on this PR because of the test case.

Issue #7066, opened 2025-04-24
function call to an interface while building the symbol table raise wrong error


assem2002:
```.F90


!!!!!!!!!!
  ! CIRCULAR DEPENDENCY !
!!!!!!!!!!

module mm

interface wow  !>>>>>>>>>>>>>>>>>>>>>>>>>>>>> Depends on both `wow_func` and `wow_func2` to be declared 
  procedure :: wow_func
  procedure :: wow_func2
end interface

contains

pure function wow_func(int) result(length)
integer, intent(in) :: int
integer :: length(wow(int)) !>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> Depends on `Wow` to be fully declared
length = int*2
end function 

pure function wow_func2(int) result(length)
integer, intent(in) :: int
integer :: length
length = int*2
end function 

function maybe(int) result(string)
  integer :: int
  character(10) :: string(wow(int)) 
end function maybe

end module mm


program pp
  use mm
  print *, maybe(10)

end program 




```

```console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranUpStream/lfortran/src/bin/lfortran tt4.f90 
semantic error: Function 'wow' not found (not user defined nor intrinsic)
  --> tt4.f90:18:27
   |
18 |   character(10) :: string(wow(int))
   |                           ^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

assem2002:
This showed up while refactoring strings, before we used to neglect any function calls passed to `len` while declaring a character type variable (ASR didn't have any information about it), now with the refactorings being done we're correcting that to actually use the functionCall.

assem2002:
A potential way of fixing that would be delaying the evaluation of these variable-declaration-related-expressions to be evaluated at the very end, maybe by using list of lambdas that capture the needed variables. and we just invoke them all after finishing the whole symbolTable creation phase, similar to what we do to `genericProcedure` right now we delay its symbolTable creation till the very end.

Pranavchiku:
This fails because we resolve function from `GenericProcedure`, we can delay that and have a separate function fixing these once ASR is created.

assem2002:
I'm fixing it, as it's needed for the string refactoring

PR #7081, opened 2025-04-27
Fix : [GenericProcedure] Build the symbolTable is this order (Functions -> GenericProcedures -> Variables depending on calls to genericProcedures)

assem2002:
Fixes https://github.com/lfortran/lfortran/issues/7066

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -2824,7 +2824,7 @@ public:
         }
         /*
             Check for generic procedure call in the symbol dimension
-            e.g. : `integer :: arr(func_generic(),20)
+            e.g. : `integer :: arr(func_generic(),20)`
         */
         for(size_t i = 0; i < x.n_syms; i++){
             AST::var_sym_t &s = x.m_syms[i];
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1402,6 +1402,16 @@ public:
     std::vector<ASR::stmt_t*> &data_structure;
     LCompilers::LocationManager &lm;
 
+    std::map<std::string, std::vector<std::string>> generic_procedures;
+    /*
+        Preserves AST::declaration node + its scope that should be evaluated in.
+        Used to evaluate the `AST::Declaration_t` nodes after resolving `ASR::GenericProcedure_t` node,
+        as these declaration nodes depends on calls to GenericProcedure node.
+        Evalution Flow (of building symbolTables) :
+            Functions -> GenericProcedure(Interface) -> Calls to GenericProcedure.
+    */
+    std::vector<std::pair<SymbolTable*, const AST::Declaration_t*>> delayed_generic_procedure_calls;
+
     // global save variable
     bool is_global_save_enabled = false;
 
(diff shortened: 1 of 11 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -649,7 +649,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             throw SemanticAbort();
         }
         handle_save();
+        // Build : Functions --> GenericProcedure(Interface) -> variable_decl calling GenericProcedure.
         add_generic_procedures();
+        Evaluate_delayed_generic_procedure_calls();
```
Pranavchiku, line 654: This may fail if program is not present.

tests/reference/asr-string_17-29e01e3.stdout
```diff
@@ -163,7 +163,8 @@
                                                 (Variable
                                                     5
                                                     maybe_string
-                                                    [string]
+                                                    [len_string
```
Pranavchiku, line 166: I'm concerned here, it shall not have `len_string` as dependency.

Pranavchiku (approved):
Left a concern, rest all looks good.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2790,7 +2800,54 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            }
+        }
+        return false; // Default if no genericProcedureCall found.
+    }
+
     void visit_DeclarationUtil(const AST::Declaration_t &x) {
+        if(is_declaration_depending_on_unresolved_genericProcedure_call(x)){
+            delayed_generic_procedure_calls.emplace_back(current_scope, &x);
+            return;
+        }
```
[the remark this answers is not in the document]
Pranavchiku: This won't affect cases without generic procedure ( as `if(generic_procedures.empty()) return false` will return false ).

src/lfortran/semantics/ast_symboltable_visitor.cpp
assem2002: Yeah it shouldn't fail. but we may not need to use these two calls after evaluating `AST::Program_t`
```
        add_generic_procedures();
        Evaluate_delayed_generic_procedure_calls();
```
I think these calls are just needed within `ASR::Module_t`

src/lfortran/semantics/ast_common_visitor.h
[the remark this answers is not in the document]
assem2002: Yeah it just delays any `AST::Declaration_t` node that depends on call to some `GenericProcedure`, and after finishing the build of almost every thing (module's functions and other constructs + resolving genericProcedure) we come back to those delayed declaration nodes and evaluate them.

tests/reference/asr-string_17-29e01e3.stdout
assem2002: I'm not 100% sure about the job of the dependency member in this node and other nodes as well, but what I get is -> They are some ASR nodes that are needed by this current ASR node ( and they're not explicitly referenced to in the fortran syntax  as a dependency, unlike using keyword `use`).

so the code here is depending on call to generic function `len` that get resolved to `len_string`:
```.f90
module string_17_string_module
implicit none
    type :: string_type
        sequence
        private
        character(len=:), allocatable :: raw
    end type string_type

    interface len
        module procedure :: len_string
    end interface

contains

    elemental function len_string(string) result(length)
            type(string_type), intent(in) :: string
            integer :: length

            if (allocated(string%raw)) then
                length = len(string%raw)
            else
                length = 0
            end if

    end function len_string

    pure function maybe(string) result(maybe_string)
            type(string_type), intent(in) :: string
            character(len=len(string)) :: maybe_string
            if (allocated(string%raw)) then
                maybe_string = string%raw
            else
                maybe_string = ''
            end if
    end function maybe

end module
```

Let me know if I understand that correctly. @Pranavchiku

src/lfortran/semantics/ast_symboltable_visitor.cpp
assem2002, line 654:
```suggestion
        evaluate_delayed_generic_procedure_calls();
```

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -496,6 +495,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
         }
         current_module_sym = nullptr;
         add_generic_procedures();
+        Evaluate_delayed_generic_procedure_calls();
```
assem2002, line 498:
```suggestion
        evaluate_delayed_generic_procedure_calls();
```

assem2002:
Great! Thanks for reviewing it! 
Can we leave it unmerged for now, I found a the case that's failing (specifically when we have a variable declaration of the function return + it depends on a call to some genericProcedure.
I'll find out if we can merge that and handle that case in another PR. or the needed change is critical that it will make lots of changes to this PR.

tests/reference/asr-string_17-29e01e3.stdout
Pranavchiku: Okay, I think it is fine. I’ll give a detailed look today evening, feel free to merge this and make progress.

Pranavchiku: I think it is fine to have it. We are specific on what function it will be using at compile time which is good.

PR #7083, opened 2025-04-27
ref: introduce `reallo_lhs` field in `Assignment` ASR node 

### LINKED ISSUE #7069: Realloc LHS TODO
Let's add a flag to Assignment like this:
```diff
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -30,7 +30,7 @@ stmt
     = Allocate(alloc_arg* args, expr? stat, expr? errmsg, expr? source)
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
-    | Assignment(expr target, expr value, stmt? overloaded)
+    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs)
     | Associate(expr target, expr value)
     | Cycle(identifier? stmt_name)
     | ExplicitDeallocate(expr* vars)
```
And set it to true if `--realloc-lhs` is true *and* the LHS is allocatable. Otherwise false.

In the backend if `realloc_lhs` is true, then it will insert `check_and_allocate` in LLVM.

Note: any ASR pass will preserve `realloc_lhs`. If it knows or can prove that some assignment has the LHS already allocated and to the right shape, then it can set `realloc_lhs=false`, as an optimization.

Pranavchiku:
Fixes #7069


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1383,7 +1383,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 body.push_back(al, associate_stmt);
             } else {
                 ASRUtils::make_ArrayBroadcast_t_util(al, tmp_expr->base.loc, target_var, tmp_expr);
-                ASR::stmt_t* assign_stmt = ASRUtils::STMT(ASR::make_Assignment_t(al, tmp_expr->base.loc, target_var, tmp_expr, nullptr));
+                ASR::stmt_t* assign_stmt = ASRUtils::STMT(ASRUtils::make_Assignment_t_util(al, tmp_expr->base.loc, target_var, tmp_expr, nullptr, compiler_options.po.realloc_lhs));
```
gxyd, line 1386: If we are using `realloc_lhs` in creating an ASR, so that means it isn't just a pass option anymore?

currently it's `compiler_options.po.realloc_lhs` where `po` means `PassOptions`.

src/libasr/ASR.asdl
```diff
@@ -30,7 +30,7 @@ stmt
     = Allocate(alloc_arg* args, expr? stat, expr? errmsg, expr? source)
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
-    | Assignment(expr target, expr value, stmt? overloaded)
+    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs)
```
gxyd, line 33: I would like to understand, what's the downside of having:
```asdl
Assignment(expr target, expr value, bool realloc_lhs, stmt? overloaded)
```

?

i.e. having `bool realloc_lhs` before `stmt? overloaded`, considering that `overloaded` is optional.

src/libasr/pass/pass_utils.h
```diff
@@ -11,6 +11,10 @@ namespace LCompilers {
 
     namespace PassUtils {
 
+        ASR::asr_t* make_Assignment_t_util(Allocator &al, const Location &a_loc,
+            ASR::expr_t* a_target, ASR::expr_t* a_value,
+            ASR::stmt_t* a_overloaded, bool a_realloc_lhs=false);
+
```
gxyd, line 17: I'm sorry, why do we need this?

src/libasr/pass/pass_utils.cpp
```diff
@@ -567,6 +567,15 @@ namespace LCompilers {
...
         }
 
+        ASR::asr_t* make_Assignment_t_util(Allocator &al, const Location &a_loc,
+            ASR::expr_t* a_target, ASR::expr_t* a_value,
+            ASR::stmt_t* a_overloaded, bool a_realloc_lhs) {
+            a_realloc_lhs = a_realloc_lhs && ( ASRUtils::is_allocatable(a_target) || ASRUtils::is_struct(*ASRUtils::expr_type(a_target)) ||
+                            ASR::is_a<ASR::StructInstanceMember_t>(*a_target) );
+            return ASR::make_Assignment_t(al, a_loc, a_target, a_value,
+                a_overloaded, a_realloc_lhs);
+        }
```
gxyd, line 577: I'm sorry, why do we need this?

src/libasr/ASR.asdl
Pranavchiku: No downside I would say, we can rearrange these.

src/libasr/pass/pass_utils.cpp
Pranavchiku: passes don't have access to `ASRUtils` name space and hence have to copy it over here.

src/lfortran/semantics/ast_body_visitor.cpp
Pranavchiku: Yup, all the options that needs to be used in passes "as well" are put in `PassOptions`.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2097,8 +2097,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             ASR::expr_t* expression_value = ASRUtils::expr_value(value);
             if (expression_value) {
                 ASRUtils::make_ArrayBroadcast_t_util(al, x.base.base.loc, object, expression_value);
-                ASR::stmt_t* assignment_stmt = ASRUtils::STMT(ASR::make_Assignment_t(al, x.base.base.loc,
-                                            object, expression_value, nullptr));
+                ASR::stmt_t* assignment_stmt = ASRUtils::STMT(ASRUtils::make_Assignment_t_util(al, x.base.base.loc,
+                                            object, expression_value, nullptr, compiler_options.po.realloc_lhs));
```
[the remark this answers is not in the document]
Pranavchiku: `ASRUtils::make_Assignment_t_util` will do that check, here we only pass necessary information to it.

src/libasr/pass/pass_utils.h
```diff
@@ -596,6 +600,18 @@ namespace LCompilers {
...
+                    ASR::Assignment_t& xx = const_cast<ASR::Assignment_t&>(x);
+                    ASR::expr_t* a_target = xx.m_target;
+                    bool is_allocatable = ASRUtils::is_allocatable(a_target);
+                    if ( ASR::is_a<ASR::StructInstanceMember_t>(*a_target) ) {
+                        ASR::StructInstanceMember_t* a_target_struct = ASR::down_cast<ASR::StructInstanceMember_t>(a_target);
+                        is_allocatable |= ASRUtils::is_allocatable(a_target_struct->m_v);
+                    }
+                    xx.m_realloc_lhs &= is_allocatable;
+                    BaseWalkVisitor<UpdateDependenciesVisitor>::visit_Assignment(x);
+                }
```
Pranavchiku, line 614: This is done because in `array_op` pass we do an inplace update, for ex: `A = 9.0` where `A` is an allocatable array gets converted to:

```
do i = lbound(A, 1), ubound(A,2)
A(i) = 9.0
```

Here the assignment node `A(i) = 9.0` is the same node as `A = 9.0` and hence it retains `.true.` for `x.m_realloc_lhs`.

Pranavchiku: > Or I guess this only sets it if it was allocatable before, due to the &= operator, right?

yes, correct.

Pranavchiku:
@gxyd @czgdp1807 your review will be very helpful, please do it once you have time!

gxyd:
> @gxyd @czgdp1807 your review will be very helpful, please do it once you have time!

Apologies, I missed it. I'll do that today.

src/libasr/asr_utils.cpp
```diff
@@ -409,6 +409,19 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     return mod2;
 }
 
+ASR::asr_t* make_Assignment_t_util(Allocator &al, const Location &a_loc,
+    ASR::expr_t* a_target, ASR::expr_t* a_value,
+    ASR::stmt_t* a_overloaded, bool a_realloc_lhs) {
+    bool is_allocatable = ASRUtils::is_allocatable(a_target);
+    if ( ASR::is_a<ASR::StructInstanceMember_t>(*a_target) ) {
+        ASR::StructInstanceMember_t* a_target_struct = ASR::down_cast<ASR::StructInstanceMember_t>(a_target);
+        is_allocatable |= ASRUtils::is_allocatable(a_target_struct->m_v);
```
czgdp1807, line 418: Is reallocation allowed when LHS is a pointer? If so then that case should be handled here.

src/libasr/pass/arr_slice.cpp
czgdp1807, line 1: By the way this pass is not being used now because we don't need it anymore. So let's remove this file altogether. :-)

src/libasr/pass/array_struct_temporary.cpp
czgdp1807, line 1: I remember using `realloc_lhs` here in this pass. May be now we should update it to use, `realloc_lhs` flag in Assignment statement?

src/libasr/pass/pass_utils.cpp
czgdp1807: As far as I remember passed do have access to `ASRUtils`. Like you can include `asr_utils.h` (must be already included) and then use inside this file.

czgdp1807 (commented):
May be the real use case would be to actually just use `realloc_lhs` in compiler options till AST to ASR. Afterwards we should use `realloc_lhs` in `Assignment` node. That would clear up how we benefit from the addition of `realloc_lhs` in `Assignment` node.

Issue #7073, opened 2025-04-25
[stdlib: bitsets] Unsupported or unrecognized `B` in format string


HarshitaKalani:
```fortran
program test_binary_format
  implicit none
  integer :: x
  x = 42
  print '(B8)', x
end program test_binary_format

```
```console
$ gfortran a.f90 && ./a.out
  101010
$ lfortran a.f90
Unsupported or unrecognized `B` in format string
```

harperjf:
Reported in #3813, still there.

Issue #3813, opened 2024-04-08
No b (binary) format


harperjf:
This program tried unsuccessfully to use b32 and b0 formats:
``` 
(lf2) john:~/Test/Lfortran$ cat testbfmtmin.f90
program testbfmtmin
  print "(a,b32,a)",'-1 in b32 format: "', -1,'"'
  print "(a,b0 ,a)",'-1 in b0  format: "', -1,'"' ! valid f95; AMD gives "" 
end program testbfmtmin
(lf2) john:~/Test/Lfortran$ lfortran testbfmtmin.f90
Unsupported or unrecognized `b` in format string
(lf2) john:~/Test/Lfortran$ 
```

Issue #5961, opened 2025-01-10
Support `(b64.64)` string formatting


Pranavchiku:
```fortran
program demo_epsilon
use,intrinsic :: iso_fortran_env, only : dp=>real64,sp=>real32
implicit none
real(kind=dp) :: t
real(kind=dp) :: my_dp_eps

   ! starting with a value of 1, keep dividing the value
   ! by 2 until no change is detected. Note that with
   ! infinite precision this would be an infinite loop.
   ! 1/2 of a non-zero number is always smaller than the number, right?
   my_dp_eps = 1.0d0
   INFINITE: do
               my_dp_eps = my_dp_eps/2.0d0
               t = 1.0d0 + my_dp_eps
               if (t <= 1.0d0) exit
   enddo INFINITE
   my_dp_eps = 2.0d0*my_dp_eps
   print *, my_dp_eps
   print *, 'weird coincidence, that is the same as ',epsilon(0.0d0)
   print *, 'congratulations! You calculated the epsilon value of a machine the hard way!'
   print '(b64.64)',my_dp_eps

end program demo_epsilon
```

```console
% lfortran d.f90 
style suggestion: Use 'end do' instead of 'enddo'
  --> d.f90:16:4
   |
16 |    enddo INFINITE
   |    ^^^^^ help: write this as 'end do'


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
x.m_name: my_dp_eps
x.m_name: t
x.m_name: character_kinds
x.m_name: error_unit
x.m_name: input_unit
x.m_name: int16
x.m_name: int32
x.m_name: int64
x.m_name: int8
x.m_name: integer_kinds
x.m_name: iostat_end
x.m_name: logical_kinds
x.m_name: output_unit
x.m_name: real128
x.m_name: real32
x.m_name: real64
x.m_name: real_kinds
2.22044604925031308e-16
weird coincidence, that is the same as     2.22044604925031308e-16
congratulations! You calculated the epsilon value of a machine the hard way!
Unsupported or unrecognized `b` in format string ------------------------------->
```

HarshitaKalani:
The output with the latest main:
```console
$ lfortran a.f90
2.22044604925031308e-16
weird coincidence, that is the same as     2.22044604925031308e-16
congratulations! You calculated the epsilon value of a machine the hard way!
0011110010110000000000000000000000000000000000000000000000000000
$ gfortran a.f90 && ./a.out
[...]

PR #7085, opened 2025-04-27
Support b (binary) format

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/7073
fixes: https://github.com/lfortran/lfortran/issues/3813


gptsarthak:
We need to add support for this format for REAL type as well, see #5961

integration_tests/CMakeLists.txt
```diff
@@ -698,6 +698,7 @@ RUN(NAME format_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+RUN(NAME format_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
```
gxyd, line 701: I think, what's more important for tests like these is that they produce the same output, i.e. having them tested in `tests.toml`, just testing it as part of integration tests doesn't make sure that they produced the same output (considering we don't have any `if (...) error stop` conditions in it.

integration_tests/format_22.f90
```diff
@@ -0,0 +1,18 @@
...
+  print '(B4)', x
+  print '(B5)', x
+  print '(B6)', x
+  print '(B7)', x
+  print '(B8)', x
+  print '(B9)', x
+  print '(B10)', x
+  print "(a,b32,a)",'-1 in b32 format: "', -1,'"'
+  print "(a,b0 ,a)",'-1 in b0  format: "', -1,'"'
+end program format_22
```
gxyd, line 18: A little more variety with different integer values would be good, like some ending with zeros, etc.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -892,25 +893,27 @@ char primitive_enum_to_format_specifier(Primitive_Types primitive_enum){
...
     if(lowered_format_value == 'd' || lowered_format_value == 'e'){
         lowered_format_value = 'f';
     }
     // Special conditions that are allowed by gfortran.
     bool special_conditions = (lowered_format_value == 'l' && current_arg_correct_format == 'a') ||
-                              (lowered_format_value == 'a' && current_arg_correct_format == 'l');
-    if(lowered_format_value != current_arg_correct_format && !special_conditions){
+                               (lowered_format_value == 'a' && current_arg_correct_format == 'l');
+
+    bool b_format_for_integer = (lowered_format_value == 'b' && current_arg_correct_format == 'i');
```
gxyd, line 908: If we don't yet support "real, string" types, I think we should leave a comment, because I think we might be able to simplify the logic here.

These seem very much corner conditions right now.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1480,6 +1483,51 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, const c
...
+                        if ((uval >> bit) & 1) {
+                            started = true;
+                        }
+                        if (started) {
+                            binary_str[idx++] = ((uval >> bit) & 1) ? '1' : '0';
+                        }
+                    }
+
+                    if (idx == 0) {
+                        binary_str[idx++] = '0'; // If number is 0
```
gxyd, line 1510: Can you please give an example of when this will be true? and add a test case for it?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1480,6 +1483,51 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, const c
                 } else if (tolower(value[0]) == 'i') {
                     // Integer Editing ( I[w[.m]] )
                     handle_integer(value, integer_val, &result, is_SP_specifier);
+                } else if (tolower(value[0]) == 'b') {
+                    int width = 0;
+                    if (strlen(value) > 1) {
+                        width = atoi(value + 1); // Get width after 'B'
+                    }
+
+                    int bit_size = 32; // default INTEGER*4 in Fortran
```
gxyd, line 1492: Please if we can leave a comment to deal with other types

gxyd (commented):
I'm happy if we address implementation of "real" and or other types in subsequent PR's, but we should leave some more informative comments about that.

gxyd:
Even though there is an approval from @certik , I would request if we can address some comments in this PR, kindly let me know how you would like to proceed here.

Pranavchiku:
In general I'm happy to merge and do improvements in subsequent PRs.

HarshitaKalani:
I'm fine either way. Maybe, since the changes are not difficult, I'll make them in this PR and then set it to review again.
For supporting the real type, I'll open another PR.

integration_tests/CMakeLists.txt
HarshitaKalani: I see format_01, 02, 03, 11, 12 and 13 registered in tests.toml. However, the others are not. I'll register this one (format_22) in tests.toml too. What is your say on the rest of the tests, should I include them too?

HarshitaKalani:
@gxyd I've addressed the comments. Can you please review? Thanks!

Pranavchiku (approved):
I think this is good. @gxyd what do you say?

HarshitaKalani:
I've the implementation of real ready. Once we merge this, I'll push a PR for that.

PR #7089, opened 2025-04-27
`show-ast` flag return non zero exit code if error is encountered in fortran file

parth121101:


gxyd (approved):
Looks good to me, thanks @parth121101

PR #7092, opened 2025-04-27
Handle BitCast for Array Type in llvm backend

AnshMehta1:
Towards #6805 

MRE :- 
```.f90
program test
    implicit none
    real :: value(5) = [1.1, 1.2, 1.3, 1.4, 1.5]
    integer :: val(5)
    val = transfer(value, val, 1 * size(value))
    print * , val
end program
```

LFortran Output :- 
```console
1066192077    1067030938    1067869798    1068708659    1069547520
```


AnshMehta1:
Currently, Bitcast (transfer function) is still not working with allocatable arrays. Will handle this in next PR which would eventually fix the associated issue #6805 .

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7901,6 +7901,10 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
         llvm::Type* target_base_type = llvm_utils->get_type_from_ttype_t_util(x.m_type, module.get());
```
gxyd, line 7901: if you need to need `target_base_type` overriding the Array from over it, why not just do:
```suggestion
        llvm::Type* target_base_type = llvm_utils->get_type_from_ttype_t_util(ASRUtils::type_get_past_array(x.m_type), module.get());
```

that's a cleaner solution as I understand it.

gxyd (commented):
I think we can simplify the change as it's introduced in this PR, once that's done I believe this is good.

gxyd:
> Currently, Bitcast (transfer function) is still not working with allocatable arrays. Will handle this in next PR which would eventually fix the associated issue #6805 .

Understood, it seems like ASR created for it might have some issue?

gxyd (approved):
Looks perfect, thanks for addressing the comment @AnshMehta1

Issue #7094, opened 2025-04-28
LSP timeout failure in concurrent mode


certik:
This blocked the CI for 5 hours:

https://github.com/lfortran/lfortran/actions/runs/14690524288/job/41225147609?pr=7083

```
+ pytest -vv --showlocals --timeout=10 --execution-strategy=concurrent tests/server
============================= test session starts ==============================
platform darwin -- Python 3.10.14, pytest-8.3.5, pluggy-1.5.0 -- /Users/runner/micromamba/envs/lf/bin/python3.10
cachedir: .pytest_cache
rootdir: /Users/runner/work/lfortran/lfortran/tests/server
plugins: anyio-4.9.0, xonsh-0.16.0, timeout-2.3.1
timeout: 10.0s
timeout method: signal
timeout func_only: False
collecting ... collected 13 items

tests/server/tests/test_document_manipulation.py::test_document_manipulation PASSED [  7%]
tests/server/tests/test_features.py::test_goto_definition PASSED         [ 15%]
tests/server/tests/test_features.py::test_diagnostics PASSED             [ 23%]
tests/server/tests/test_features.py::test_configuration_caching PASSED   [ 30%]
tests/server/tests/test_features.py::test_rename PASSED                  [ 38%]
tests/server/tests/test_features.py::test_document_highlight ERROR       [ 46%]
tests/server/tests/test_features.py::test_document_hover PASSED          [ 53%]
tests/server/tests/test_features.py::test_symbol_tree PASSED             [ 61%]
tests/server/tests/test_features.py::test_semantic_highlighting PASSED   [ 69%]
tests/server/tests/test_features.py::test_code_completion PASSED         [ 76%]
tests/server/tests/test_features.py::test_whole_document_formatting PASSED [ 84%]
tests/server/tests/test_features.py::test_partial_document_formatting PASSED [ 92%]
tests/server/tests/test_features.py::test_telemetry_event PASSED         [100%]

==================================== ERRORS ====================================
__________________ ERROR at setup of test_document_highlight ___________________
[...]
=========================== short test summary info ============================
ERROR tests/server/tests/test_features.py::test_document_highlight - RuntimeError: Timed-out after 3.0 seconds while reading from the stream:
========================= 12 passed, 1 error in 16.22s =========================
```

I uploaded the full log at https://gist.github.com/certik/bf60ce8c3af5d8b7a069d93ee5c7a515.

gxyd:
> This blocked the CI for 5 hours:

[...]

PR #7102, opened 2025-04-29
Adds additional layers of timeouts to prevent LSP bugs from blocking the pipeline

dylon:
Adds the following levels of timeouts:
1. `pytest` is run with `timeout` which will send it a `SIGTERM` after 60 seconds and a `SIGKILL` 10 seconds later (if it hasn't terminated).
2. The whole step is run with a timeout of 5 minutes. Github will kill the process after 5 minutes if it has not terminated.


gxyd:
Is this PR intended to fix the issue: https://github.com/lfortran/lfortran/issues/7094 ?

dylon:
It is intended to prevent the unit tests from blocking the pipeline. I will need to deep dive into what's happening to identify the root cause. I should be able to spend some time on it tomorrow.

PR #7105, opened 2025-04-29
Fix : function return

assem2002:
I'm testing CI for now.

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1313,8 +1313,8 @@ public:
         is_global_save_enabled = is_global_save_enabled_copy;
     }
 
-    AST::AttrType_t* find_return_type(AST::decl_attribute_t** attributes,
-            size_t n, const Location &loc, std::string &return_var_name, ASR::symbol_t* return_var_sym) {
+    AST::AttrType_t* get_return_type_or_implict_type(AST::decl_attribute_t** attributes,
+            size_t n, const Location &loc, std::string &return_var_name, bool return_var_symbol_exist_in_function_scope) {
         AST::AttrType_t* r = nullptr;
         bool found = false;
         bool are_all_attributes_simple = true;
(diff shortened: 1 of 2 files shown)


integration_tests/template_03b.f90
```diff
@@ -4,7 +4,7 @@ program template_03
         type, deferred :: T
         type, deferred :: U
         type, deferred :: V
-        elemental function op(a, b) result(op)
+        elemental function op(a, b)
```
Pranavchiku, line 7: is it incorrect Fortran?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1539,149 +1539,101 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        if (x.m_return_var) { // Case -> `function foo() result(ret)`
             if (x.m_return_var->type == AST::exprType::Name) {
                 return_var_name = to_lower(((AST::Name_t*)(x.m_return_var))->m_id);
+                if(current_scope->get_symbol(to_lower(x.m_name))){
+                    diag.add(diag::Diagnostic(
+                        "Function already have result variable set",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {x.m_return_var->base.loc})}));
+                    throw SemanticAbort();
+                }
```
Pranavchiku, line 1558: Let's add a reference test for this error

tests/reference/asr-class_02-b56b852.stdout
```diff
@@ -117,7 +117,7 @@
                                                     (Real 4)
                                                     ()
                                                     Source
-                                                    Public
+                                                    Private
```
Pranavchiku, line 120:
```fortran
module class_circle2
  implicit none
  private
  public :: main
  real :: pi = 3.1415926535897931d0 ! Class-wide private constant

  type, public :: Circle
     real :: radius
   contains
     procedure :: circle_area  !=> circle_area ! Can you point to implementation in diff module?
     procedure :: circle_print
  end type Circle
...
```

I think `circle_area` shall be public as we can access it by creating instance of `Circle`

Pranavchiku (commented):
If it takes too much time and is not a blocker, I would prefer to mark this as draft and let it wait unless absolutely necessary.

assem2002:
It's not necessary, it's just a refactor that I thought is good to merge.

Pranavchiku:
That is fine, marking this PR as draft for now.

PR #7106, opened 2025-04-29
Removed unused flag `fixed_form_cc` and add `fixed_form_cc_ast` flag

parth121101:


parth121101:
It seems I have discovered a new bug here. I have started losing some errors and warnings.

run_tests.py
```diff
@@ -423,11 +423,11 @@ def is_included(backend):
                 verify_hash,
                 extra_args)
             
-    if fixed_form_cc:
+    if fixed_form_cc_ast:
```
Pranavchiku, line 426: Why are we doing it, fixed_form_cc is not used?

run_tests.py
```diff
@@ -423,11 +423,11 @@ def is_included(backend):
...
         run_test(
                 filename,
-                "asr",
-                "lfortran --fixed-form --semantics-only --continue-compilation --no-color {infile}",
+                "ast",
+                "lfortran --fixed-form --semantics-only --continue-compilation --show-ast --no-color {infile} -o {outfile}",
```
Pranavchiku, line 430:
```suggestion
                "lfortran --fixed-form --semantics-only --continue-compilation --show-asr--no-color {infile} -o {outfile}",
```

tests/reference/ast-continue_compilation_ff_1-cdae745.stderr
```diff
@@ -75,21 +75,3 @@ syntax error: Newline is unexpected here
     |
 119 |       CALL FUNC( 5, 6,
     |                       ^ 
-
-warning: This equivalence statement is not implemented yet, for now we will ignore it
-   --> tests/errors/continue_compilation_ff_1.f:108:7
```
Pranavchiku, line 80: I think we throw this warning in AST->ASR

run_tests.py
Pranavchiku: continue compilation automatically stops at ASR

Pranavchiku:
> It seems I have discovered a new bug here. I have started losing some errors and warnings.

This is because you now force it to stop at AST and the errors and warnings were thrown in ASR.

What is the purpose to stop it at AST?

Pranavchiku (changes requested):
left a few questions and suggestions above

run_tests.py
parth121101: Yes, its no longer used. It was just used for 1 file before my PR. So decided just return AST for all the files as commented by Ondrej here https://github.com/lfortran/lfortran/pull/7042#discussion_r2064501487

parth121101:
> > It seems I have discovered a new bug here. I have started losing some errors and warnings.
> 
> This is because you now force it to stop at AST and the errors and warnings were thrown in ASR.
> 
> What is the purpose to stop it at AST?

So you are saying that if I use `--show-ast` compilation would stop at AST, right?
If this is the case then the lost errors and warnings are expected. So this is not a bug. Is my understanding correct ?

tests/reference/ast-continue_compilation_ff_1-cdae745.stderr
```diff
@@ -75,21 +75,3 @@ syntax error: Newline is unexpected here
...
-
-semantic error: Variable 'integerklm' is not declared
-  --> tests/errors/continue_compilation_ff_1.f:99:7
-   |
-99 |       INTEGER K L M = 5
-   |       ^^^^^^^^^^^^^ 'integerklm' is undeclared
-
-semantic error: Variable 'n5' is not declared
-   --> tests/errors/continue_compilation_ff_1.f:100:15
-    |
```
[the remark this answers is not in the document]
parth121101: With the recent commit I am no longer facing problem of losing errors

Issue #6771, opened 2025-03-29
--realloc-lhs flag causes error in reshape


gxyd:
## Minimal Reproducible Example

```f90
module lincoa_mod
    implicit none

    contains

    subroutine get_lincon(amat)
        implicit none
        real, intent(out), allocatable :: amat(:, :)
        integer :: m, n
        m = 3
        n = 2
        allocate(amat(m, n))
        amat = reshape([1., 2., 3., 4., 5., 6.], shape(amat))
    end subroutine get_lincon
end module lincoa_mod

program lincoa_exmp
    use lincoa_mod
    implicit none
    real, allocatable :: a(:, :)
    call get_lincon(a)
    print *, a
end program lincoa_exmp


```

The above Fortran program, when executed as below, gives the below output / errors:
```console
> lfortran test.f90  # this works without any issue
1.00000000e+00    2.00000000e+00    3.00000000e+00    4.00000000e+00    5.00000000e+00    6.00000000e+00

> lfortran --version
LFortran version: 0.49.0-153-gd33392b96
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90 --realloc-lhs
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2687
    LCompilers::print_stack_on_segfault();
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2618
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 1165
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 382
    compiler_options, run_fn, infile);
  File "/Users/gxyd/OpenSource/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 10896
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/Users/gxyd/OpenSource/lfortran/src/libasr/pass/pass_manager.h", line 310
    apply_passes(al, asr, _passes, pass_options, diagnostics);
  File "/Users/gxyd/OpenSource/lfortran/src/libasr/pass/pass_manager.h", line 190
[...]

PR #7110, opened 2025-04-29
Fix `--realloc-lhs` if `shape` is dependent on target in `ArrayReshape`

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/6771


HarshitaKalani (approved):
Looks good to me. Thanks @czgdp1807 !

PR #7114, opened 2025-04-30
Fix : GenericProcedure calls in variable declaration [postpone the `AST::FuncCallOrArray` node instead of the whole `AST::Declaration `node]

### LINKED ISSUE #7095: Bug : genericProcedure call while declaring a returnVar is raising false error
```.f90
module mm
  interface generic
     procedure :: ff
  end interface
  contains

  pure function ff() result(ret)
    integer :: ret
    ret  = 190
  end function

  function foo() result(arr) 
    integer :: arr(generic()) ! >>>>>>>>>>>>>>>>>>>>>>>>> HERE
  end function

end module

program pop
  use mm
  print *, size(foo())
end program pop
```

Current main
```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran  tt4.f90 
semantic error: Return type not specified
  --> tt4.f90:12:3 - 14:14
   |
12 |      function foo() result(arr) 
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
14 |      end function
   | ...^^^^^^^^^^^^^^ 


```
Before this commit 023b108f576247b438f3e117f5eda704ee278b50

```Console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranUpStream/lfortran/src/bin/lfortran tt4.f90 
semantic error: Function 'generic' not found (not user defined nor intrinsic)
  --> tt4.f90:13:20
   |
13 |     integer :: arr(generic())
   |                    ^^^^^^^^^ 


```

assem2002:
Fix #7095 
- It's an enhancement to this PR https://github.com/lfortran/lfortran/pull/7081
### What it it fixes : 
GenericProcedure calls while declaring a variable `integer :: ret(genericProcedure())`

### What is the issue ? : 
We used to postpone the evaluation of the whole `AST::Declaration` node till we Evaluate symbolTable of both the Functions and then -> `AST::Interface`.
That led to some `ASR::Function` node being built with incomplete signature, which invoked semantic errors see #7095.

### The Fix :
Just postpone the `AST::FuncCallOrArray` node instead of the whole `AST::Declaration` node, then evaluate the node at the end (after having complete `ASR::Function` + `ASR::GenericProcedure` nodes).

### DIFF
--- a/integration_tests/modules_15b.f90
+++ b/integration_tests/modules_15b.f90
@@ -241,7 +241,7 @@ interface
     ! int f_string(char *s)
     integer(c_int) function f_string0(s) result(r) bind(c, name="f_string")
     import :: c_int, c_char
-    character(len=1, kind=c_char), intent(in) :: s(*)
+    character(len=1, kind=c_char), intent(in) :: s
     end function
 
     integer(c_int) function call_fortran_i32(i) result(r) bind(c)
--- a/tests/reference/asr-array12-cb81afc.stdout
+++ b/tests/reference/asr-array12-cb81afc.stdout
@@ -172,7 +172,18 @@
                                         (FunctionCall
                                             2 getcwd
                                             ()
-                                            [((Var 12 cpath))
+                                            [((ArrayPhysicalCast
+                                                (Var 12 cpath)
+                                                DescriptorArray
+                                                StringArraySinglePointer
+                                                (Array
+                                                    (String 1 1 () PointerString)
+                                                    [(()
+                                                    ())]
+                                                    StringArraySinglePointer
+                                                )
+                                                ()
+                                            ))
                                             ((Var 12 buffersize))]
                                             (CPtr)
                                             ()
@@ -223,7 +234,12 @@
                                                     ()
                                                     ()
                                                     Default
-                                                    (String 1 1 () PointerString)
+                                                    (Array
+                                                        (String 1 1 () PointerString)
+                                                        [(()
+                                                        ())]
+                                                        StringArraySinglePointer
+                                                    )
                                                     ()
                                                     BindC
                                                     Public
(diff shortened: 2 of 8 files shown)


integration_tests/modules_15b.f90
```diff
@@ -241,7 +241,7 @@ subroutine sub_int_double_value_name(a, b, r) &
     ! int f_string(char *s)
     integer(c_int) function f_string0(s) result(r) bind(c, name="f_string")
     import :: c_int, c_char
-    character(len=1, kind=c_char), intent(in) :: s(*)
+    character(len=1, kind=c_char), intent(in) :: s
```
assem2002, line 244: the type of this node used to be like this `(String 1 1 () PointerString)` --- with the fix in this PR the type got corrected to an **array of type string** ( I don't know why :) ).

The C function is declared as : 
```C
int f_string(char *s) {
    return strlen(s);
}
```

So I just fixed the test

tests/reference/asr-array12-cb81afc.stdout
```diff
@@ -223,7 +234,12 @@
                                                     ()
                                                     ()
                                                     Default
-                                                    (String 1 1 () PointerString)
+                                                    (Array
+                                                        (String 1 1 () PointerString)
+                                                        [(()
+                                                        ())]
+                                                        StringArraySinglePointer
+                                                    )
```
assem2002, line 242: That's the correct node. I don't know why exactly it got corrected :)

integration_tests/modules_15b.f90
assem2002: opened an issue here https://github.com/lfortran/lfortran/issues/7136. Gfortran casts a string to an array of strings if needed, while we don't.

PR #7130, opened 2025-05-01
Remove extra parentheses after format specification

### LINKED ISSUE #6748: Extra characters after format specification
F2023 13.2.2 paragraph 1 says "A character expression used as a format in a formatted input/output statement shall evaluate to a character string whose leading part is a valid format specification."

Paragraph 2 sentence 2 says "Character positions, if any, following the right parenthesis that ends the format specification need not be defined and may contain any character data with no effect on the interpretation of the format specification."

This program has one extra character ```)``` after its format specification ```(A)``` that causes lfortran trouble.
```
program extraformat
  print "(A))", ' foobar'
end program extraformat
```
Several compilers print
```
 foobar
```
Some of them also give warnings about extra characters in the format.
Lfortran 0.49.0 output disobeys the standard:
```
Unsupported or unrecognized `)` in format string
```
Some other extra characters do not trigger this bug.

gptsarthak:
Fixes #6748


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1221,6 +1221,34 @@ void print_into_string(Serialization_Info* s_info,  char* result){
 
 }
 
+void strip_outer_paranthesis(const char* str, int len, char* output) {
```
HarshitaKalani, line 1224:
```suggestion
void strip_outer_parenthesis(const char* str, int len, char* output) {
```

HarshitaKalani (approved):
Just a minor change which can be corrected while resolving the conflicts. Thanks @gptsarthak !

gptsarthak:
Sorry for the delay, will merge as soon as the checks are passed

PR #7147, opened 2025-05-02
Fix a bug in arrays_reshape_23 test

### LINKED ISSUE #7142: flaky GFortran test
Run: https://github.com/lfortran/lfortran/actions/runs/14792086554/job/41531256497

certik:
We cannot call `shape(amat)` since `amat` is not allocated.

Also added actual tests for the output.

Fixes #7142.

### DIFF
--- a/integration_tests/arrays_reshape_23.f90
+++ b/integration_tests/arrays_reshape_23.f90
@@ -19,7 +19,7 @@ module lincoa_mod
         integer :: m, n
         m = 3
         n = 2
-        amat = reshape([1., 2., 3., 4., 5., 6.], shape(amat))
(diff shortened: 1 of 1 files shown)


czgdp1807:
You can close this PR. I am sending another one where I will remove `get_lincon1` function because the reason why I added it is no longer valid. So keeping it in the test doesn't make sense anymore.

czgdp1807:
The bug being fixed here is different. I will open a separate issue for it. Meanwhile let me remove `get_lincon1`.

czgdp1807:
Opened https://github.com/lfortran/lfortran/issues/7151

PR #7149, opened 2025-05-02
Bug Fix : Iostat value correctly handled after successfull read from file

HiteshGhanchi:
```
program read_from_file
    implicit none
    character(len=20) :: s 
    integer :: ios
    character(len=100) ::sa

    open(unit=10, file="../codes/input.txt", status="old", iostat=ios)
    if (ios /= 0) then
        print *, "Error opening the file."
        stop
    end if

    print * , ios
    read(10, *, iostat=ios) s
    print * , ios



    print *, trim(s)
    close(10)

end program read_from_file
```

For the above code eveb after successfull read from the file , the value of iostat was -1  and not 0 (which indicates sucess).

Lfortran Output : 
```
0
-1
stringInput
```

Expected Output:
```
0
0
 stringInput
```
Also I observed that iostat value is only changed "_lfortran_empty_read" for unformatted read, but it should be done in every function ideally.
For now, the chnages I made still mimics the sucessfull output.


HiteshGhanchi:
Should I add testcase for this in integration_test folder ?

Pranavchiku:
@jinangshah21 can you please review this?

HiteshGhanchi:
@certik 
During a test, I noticed that the logic is not correct , for correct implementation I need to handle iostat value in the lfortran_read_* functions in intrinsic.cpp file.
Should I do that for all the functions or do one , have it reviewed and then proceed further ?

HiteshGhanchi:
Closing this PR in favor of a new one with a clean and correct implementation. I will reference it.

Issue #7067, opened 2025-04-24
The `print_area` method should be virtual


certik:
This example behaves differently in GFortran and LFortran. In GFortran the `print_area` is virtual, but in LFortran it is not.
```fortran
module shape_mod
    implicit none

    type :: Shape
    contains
        procedure :: print_area
    end type Shape

    type, extends(Shape) :: Circle
        real :: radius
    contains
        procedure :: print_area => circle_print_area
    end type Circle

contains

    subroutine print_area(self)
        class(Shape), intent(in) :: self
        print *, "Base shape: unknown area"
    end subroutine print_area

    subroutine circle_print_area(self)
        class(Circle), intent(in) :: self
        real :: area
        area = 3.14159 * self%radius**2
        print *, "Circle area:", area
    end subroutine circle_print_area

    subroutine mine(self)
        class(Shape), intent(in) :: self
        call self%print_area()
    end subroutine mine
end module shape_mod


program test_polymorphism
    use shape_mod
    implicit none

    type(Circle) :: c
    c%radius = 2.0

    call mine(c)

end program test_polymorphism
```

Question:

* Is the line `self%print_area()` where `self` is of type `class(Shape)` a virtual function call?

Note: The `deferred` keyword is like pure virtual function in C++.

PR #7152, opened 2025-05-02
fix: use runtime polymorphism when variable is ClassType

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7067

> About procedure resolution: https://j3-fortran.org/doc/year/24/24-007.pdf#subsection.2943
> 
> If the binding-name in a procedure-designator (R1522) is that of a specific type-bound procedure, the procedure
> referenced is the one bound to that name in the dynamic type of the data-ref .
> 
> Also from 5th point of 7.3.2.3, 
> 
> The dynamic type of a nonallocatable nonpointer polymorphic dummy argument is the dynamic type of its effective argument. 

 _Originally posted by @xaerru in [#7068](https://github.com/lfortran/lfortran/issues/7068#issuecomment-2845498140)_


xaerru:
This should be fine to merge for now. We do runtime polymorphism when the variable is `ClassType` or if it is marked as `deferred`.

We currently mark all procedures which override the deferred procedure as deferred too. That is incorrect. If we don't mark the overrides as deferred, then stdlib fails.

After this diff on this branch.
```diff
diff --git a/src/lfortran/semantics/ast_symboltable_visitor.cpp b/src/lfortran/semantics/ast_symboltable_visitor.cpp
index 48369dba4..e6dbf2e01 100644
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -2511,17 +2511,9 @@ public:

     bool check_is_deferred(const std::string& pname, ASR::Struct_t* clss) {
         auto& cdf = class_deferred_procedures;
-        while( true ) {
-            std::string proc = clss->m_name;
-            if(cdf.count(proc) && cdf[proc].count(pname) && cdf[proc][pname].count("deferred")) {
-                return true;
-            }
-            ASR::symbol_t* clss_sym = ASRUtils::symbol_get_past_external(clss->m_parent);
-            if( !clss_sym ) {
-                break;
-            }
-            LCOMPILERS_ASSERT(ASR::is_a<ASR::Struct_t>(*clss_sym));
-            clss = ASR::down_cast<ASR::Struct_t>(clss_sym);
+        std::string proc = clss->m_name;
+        if(cdf.count(proc) && cdf[proc].count(pname) && cdf[proc][pname].count("deferred")) {
+            return true;
         }
         return false;
     }
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index ccf64e130..f8e3ff05d 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -9977,8 +9977,7 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
         if( ASR::is_a<ASR::ClassProcedure_t>(*proc_sym) ) {
             ASR::ClassProcedure_t* class_proc =
                 ASR::down_cast<ASR::ClassProcedure_t>(proc_sym);
-            is_runtime_polymorphism = ASR::is_a<ASR::ClassType_t>(*ASRUtils::extract_type(ASRUtils::expr_type(x.m_dt))) ||
-                class_proc->m_is_deferred;
+            is_runtime_polymorphism = ASR::is_a<ASR::ClassType_t>(*ASRUtils::extract_type(ASRUtils::expr_type(x.m_dt)));
             proc_sym_name = class_proc->m_name;
             is_nopass = class_proc->m_is_nopass;
         }
```

This MRE (from stdlib)
```fortran
module a_m
    implicit none

    type, abstract :: base
    contains
        procedure(base_m), deferred, pass(self) :: m
        generic :: m_gen => m
    end type

    contains
        procedure, pass(self) :: m => derived_m
    end type

    abstract interface
        subroutine base_m(self, outi)
            import :: base
            class(base) :: self
            integer :: outi
        end subroutine
    end interface

    contains
        subroutine derived_m(self, outi)
            class(derived) :: self
            integer :: outi
            outi = 2
            print *, outi
        end subroutine
end module

program my
    use a_m
    implicit none

    type(derived) :: d

    integer :: outi
    call d%m_gen(outi)
end program
```
gives
```console
(lf)
[xaerru@nienna:~/projects/lfortran]$ gfortran a.f90 && ./a.out
           2
(lf)
[xaerru@nienna:~/projects/lfortran]$ ./src/bin/lfortran a.f90
code generation error: Subroutine code not generated for 'base_m'


```

This is a problem with us not inheriting `GenericProcedure` from the parent struct symbol. If we do inherit it in ASR, when converting back to Fortran we will have a duplicate generic statement in the child type. This will need some changes in `GenericProcedure` to determine which procedure to call at compile time, I'll work on it.

Pranavchiku (approved):
I think this is good.

xaerru:
`Class` symbol is not used in LFortran ASR, a `Struct` symbol is used with a symbol table which contain `ClassProcedure` symbols which are methods.

Issue #4692, opened 2024-08-25
G format unsupported


harperjf:
This 3-line program is valid Fortran 2008 but lfortran can't compile it.
```
program gformat
  print "(G0.17)", 1d0/3d0
end program gformat
```
Compile-time output from lfortran 0.39.0 either with or without --std=f23
```
Unsupported or unrecognized `G` in format string
```
Run-time output from gfortran
```
0.33333333333333331
```

PR #7157, opened 2025-05-03
enh: support g format

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4692


tests/reference/run-format_24-257ddde.stdout
```diff
@@ -0,0 +1,16 @@
+test:  1.0000 
+0.33333333333333331
+12346.
+1.2345669999999999E-08
```
HarshitaKalani, line 4: gfortran for this test case gives:
```console
$cat a.f90
print "(G0.17)", 0.00000001234567d0 
$ gfortran a.f90 && ./a.out
0.12345669999999999E-7
```
Are we fine with `1.2345669999999999E-08` or the output should be same as gfortran?

integration_tests/format_24.f90
```diff
@@ -0,0 +1,26 @@
+program format_24
+  implicit none
+  complex(8) :: z
+  real(8) :: rvals(3)
+  character(len=*), parameter :: FMT = '(*(g0.5,1x))'
+  integer, parameter :: WP = selected_real_kind(15,307)
+  rvals = [1.1d0, 2.2d0, 3.3d0]
+  z = (1.23d0, -4.56d0)
+  real(WP) :: g = 1
```
HarshitaKalani, line 9: This should have been above line 7. I'll make that change.

tests/reference/run-format_24-257ddde.stdout
Pranavchiku: We should be same as GFortran

Pranavchiku (approved):
If you can fast follow the divergence with GFortran then I am fine to merge but in that case please do not close the target issues.

PR #7158, opened 2025-05-03
enh: range check for int32 and int64

Th3C0d3Mast3r:
While testing the integer input, I found the lack of bound checks (when input beyond the limit is entered, the compiler doesn't thrown any error, but allows it to be parsed. While GFortran detects the error)

image
---
After making the proper changes in the `intrinsics.c` file, the following is the output when entering overflowing values for int32 (and int64). The image shows the stdout by GFortran, followed by LFortran:

image

Kindly do let me know if any more changes required for the same.
Thank you.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -3721,18 +3721,13 @@ LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
             exit(1);
         }
     } else {
-        // if (fscanf(filep, "%" PRId64, p) != 1) {
-        //     fprintf(stderr, "Error: Invalid input for int64_t from file.\n");
-        //     exit(1);
-        // }
-        long long temp;
-        if (fscanf(filep, "%lld", &temp) != 1) {
+        int64_t temp;
+        if (fscanf(filep, "%" PRId64, &temp) != 1) {
             fprintf(stderr, "Error: Invalid input for int64_t from file.\n");
             exit(1);
         }
-
         if (temp < INT64_MIN || temp > INT64_MAX) {
-            fprintf(stderr, "Error: Value %lld is out of integer(8) range (file).\n", temp);
+            fprintf(stderr, "Error: Value %" PRId64 " is out of integer(8) range (file).\n", temp);
             exit(1);
         }
 


Pranavchiku (commented):
Let's get CI passing and then we can have review

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3694,10 +3721,18 @@ LFORTRAN_API void _lfortran_read_int64(int64_t *p, int32_t unit_num)
             exit(1);
         }
     } else {
-        if (fscanf(filep, "%" PRId64, p) != 1) {
+        long long temp;
+        if (fscanf(filep, "%lld", &temp) != 1) {
```
kmr-srbh, line 3725: Let's keep `PRId64` for portability.

tests/errors/integer_overflow.f90
```diff
@@ -0,0 +1,25 @@
...
+        print *, "Failed to open scratch file."
+        stop
+    end if
+
+    ! writing in file
+    write(unit_no, *) 1729
+    write(unit_no, *) 2147483648_8  ! using _8 (for int64) because, when writing in int32 mode, while writing, it is parsed wrong
+
+    rewind(unit_no)
+
```
kmr-srbh, line 17: Please create a file with the current numbers you are writing. Let's use it in the test instead of a temporary file.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3625,14 +3627,23 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num)
...
-        // If any junk remains in the token, reject the input
-        if (*endptr != '\0') {
+        if (endptr == token || *endptr != '\0') {
             fprintf(stderr, "Error: Invalid input for int32_t.\n");
             exit(1);
         }
+
+        // check for overflow (when input value is more than the int32 limit)
+        if (errno == ERANGE || long_val < INT32_MIN || long_val > INT32_MAX) {
+            fprintf(stderr, "Error: Value %ld is out of int32_t range.\n", long_val);
```
kmr-srbh, line 3641: Should we say "int32_t" here? What can be a better error message?

kmr-srbh (dismissed):

tests/errors/integer_overflow.f90
kmr-srbh: As the test exits with an error message before the file is closed, the file remains undeleted. So I recommended to use a simple text file to take the input.

PR #7167, opened 2025-05-03
enh: resolve generic procedures considering overrides at compile time

xaerru:
Addressing https://github.com/lfortran/lfortran/pull/7152#issuecomment-2848624514

After this PR, we only do runtime polymorphism for `ClassType` variables.


xaerru:
Yes, that's not a correct error, I added a FIXME.

PR #7169, opened 2025-05-03
Consolidate read_file() implementations

certik:
Previously there were 3 duplicate implementations. The most general implementation returns a success/fail boolean status, and a string. This covers all use cases, so we use that.

For backwards compatibility we also provide read_file_ok() that returns the file content, but aborts on failure to open the file.


Pranavchiku (commented):
looks good

PR #7171, opened 2025-05-04
chore: fix error message for non-existent file path

### LINKED ISSUE #7170: Improve an error message for non-existent file
This should say "file `expr23.f90` does not exist", not some stuff about a linker:
```
$ lfortran examples/expr23.f90 --show-ast
error: 'examples/expr23.f90': linker input file not found: No such file or directory
```

kmr-srbh:
closes #7170

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -2434,8 +2434,7 @@ int main_app(int argc, char *argv[]) {
     }
 
     if (CLI::NonexistentPath(opts.arg_file).empty()) {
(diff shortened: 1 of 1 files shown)


Pranavchiku (approved):
let's add a reference test

kmr-srbh:
I tried to create a reference for this with the `run = true` option, but it seems the test runner requires a file to be present to create the hashes, and so fails.

https://github.com/lfortran/lfortran/blob/049e2672ded04232bfde38401be1520836860ccc/src/libasr/compiler_tester/tester.py#L149-L176

Do you know some other way to test this?

PR #7176, opened 2025-05-05
CI: use github model release link associated with certik/fastGPT

gxyd:
## Description

For the reviewer, atleast the following things need be probably done:
* Check from the release made at: https://github.com/certik/fastGPT/releases/tag/v1.0-model-github-release, whether any improvements are desired? e.g.; the release name, description, etc., so that we can fix it

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -361,8 +361,10 @@ time_section "🧪 Testing fastGPT" '
         git clean -dfx
         git checkout -t origin/namelist
         git checkout d3eef520c1be8e2db98a3c2189740af1ae7c3e06
+        # NOTE: the release file link below wouldn't necessarily
+        # need to be updated if the commit hash above is updated
         curl -f -L -o model.dat \
-                https://github.com/gxyd/gpt/releases/download/v2.0/model_fastgpt_124M_v1.dat
+            https://github.com/certik/fastGPT/releases/download/v1.0-model-github-release/model_fastgpt_124M_v1.dat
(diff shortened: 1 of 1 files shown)


ci/test_third_party_codes.sh
```diff
@@ -361,8 +361,10 @@ time_section "🧪 Testing fastGPT" '
         git clean -dfx
         git checkout -t origin/namelist
         git checkout d3eef520c1be8e2db98a3c2189740af1ae7c3e06
+        # NOTE: the release file link below wouldn't necessarily
```
gxyd, line 364: I think the first commit failed CI check with the message (not very informative):
```
===============================
🧪 Testing fastGPT
ci/test_third_party_codes.sh: eval: line 48: syntax error: unexpected end of file
===============================

Error: Process completed with exit code 2.
```

because of the presence of `'` in this comment here, it wasn't very obvious, still not sure if that's the reason, if the CI fails again, I'll push a commit removing `'`

gxyd: I think that's probably the reason, I've use `would not` instead of `wouldn't` in the 3rd commit now. If the CI check passes now, then that's the reason of failure, which probably would indicate a bug in our test_third_party_codes.sh script.

gxyd:
> I think the ubuntu and macos test case can be greatly consolidated, the code is almost identical in both. That's a separate issue.

But the branch used in macOS is different: https://github.com/lfortran/lfortran/blob/011df7a5b7492672d7bfbfd91690717fd412a9b8/ci/test_third_party_codes.sh#L356-L364

from the branch used in testing on ubunt: https://github.com/lfortran/lfortran/blob/011df7a5b7492672d7bfbfd91690717fd412a9b8/ci/test_third_party_codes.sh#L378-L384

PR #7177, opened 2025-05-05
fix: match G0 format output with gfortran for small magnitudes

HarshitaKalani:
Addressed: https://github.com/lfortran/lfortran/pull/7157#discussion_r2072349061


tests/reference/run-format_24-257ddde.stdout
```diff
@@ -1,7 +1,7 @@
 test:  1.0000 
 0.33333333333333331
 12346.
-1.2345669999999999E-08
+0.12345670000000000E-7
```
HarshitaKalani, line 4: Not the output is in same format as GFortran but there's difference in precision:
```
$cat a.f90
print "(G0.17)", 0.00000001234567d0 
$ gfortran a.f90 && ./a.out
0.12345669999999999E-7
```
Is this an issue with the implementation?

HarshitaKalani: The output now is exactly same as GFortran.

Issue #7185, opened 2025-05-06
Keyword Argument in Generic Procedures


adit4443ya:
## MRE

```fortran
module custom_deque
  implicit none
  type :: Complex32Deque
    integer :: size_ = 0
  contains
    procedure :: Complex32Deque_resize_default
    generic :: resize => Complex32Deque_resize_default
  end type Complex32Deque

contains

  subroutine Complex32Deque_resize_default(this, count, value)
    class(Complex32Deque), intent(inout) :: this
    integer, intent(in) :: count
    integer, optional, intent(in) :: value

    ! Simplified implementation to avoid undefined dependencies
    if (count >= 0) then
      this%size_ = count
      if (present(value)) then
        ! Simulate setting value (not implemented)
      end if
    end if
  end subroutine Complex32Deque_resize_default

end module custom_deque

program test_deque
  use custom_deque
  implicit none

  type(Complex32Deque) :: d
  integer :: rc=0

  call d%resize(5, value=5)
  print *, "Size:", d%size_, "RC:", rc
end program test_deque
```

LFortran Output

```console
 aditya-trivedi   lfortran    main ≡  ?3    lfortran integration_tests/resize_default.f90 
semantic error: complex32deque_resize_default is not a function/subroutine.
  --> integration_tests/resize_default.f90:35:3
   |
35 |   call d%resize(5, value=5)
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Gfortran output

```fortran
 aditya-trivedi   lfortran    main ≡  ?3    gfortran integration_tests/resize_default.f90 -o a && ./a
 Size:           5 RC:           0
```

Here if we drop value=val and just write val then it works,

gxyd:
Closing this as fixed by #7202

PR #7189, opened 2025-05-06
Fix: Handle args for Generic ClassProcedures with this as an starting argument

adit4443ya:
Towards https://github.com/lfortran/lfortran/issues/7185

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3702,6 +3702,7 @@ public:
         }
         Vec<ASR::call_arg_t> args;
         visit_expr_list(x.m_args, x.n_args, args);
+        bool nopass=false;
         if (x.n_keywords > 0) {
             ASR::symbol_t* f2 = ASRUtils::symbol_get_past_external(original_sym);
             if ( ASR::is_a<ASR::Variable_t>(*f2) ) {
@@ -3744,7 +3745,7 @@ public:
                 }
             } else if (ASR::is_a<ASR::GenericProcedure_t>(*f2)) {
                 bool is_ClassProcedure = false;
-                bool is_ClassProcedure_With_Self_Argument = false;
+                bool is_nopass=false;
                 ASR::GenericProcedure_t* f3 = ASR::down_cast<ASR::GenericProcedure_t>(f2);
                 bool function_found = false;
                 for( size_t i = 0; i < f3->n_procs && !function_found; i++ ) {
@@ -3762,13 +3763,8 @@ public:
                         is_ClassProcedure = true;
                         ASR::ClassProcedure_t* f5 = ASR::down_cast<ASR::ClassProcedure_t>(f4);
                         f4 = f5->m_proc;
-                        if (x.n_member >= 1) {
-                            ASR::call_arg_t this_arg;
-                            this_arg.loc = x.m_member[0].loc; // Location of the first member (object)
-                            this_arg.m_value = v_expr;        // The object expression (`this`)
-                            args.push_front(al, this_arg);     // Prepend `this` to the args vector
-                            is_ClassProcedure_With_Self_Argument = true;
-                        }
+                        is_nopass = f5->m_is_nopass;
+                        nopass = is_nopass;
                     }
                     ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(f4);
                     diag::Diagnostics diags;
@@ -3776,7 +3772,13 @@ public:
                     args_.from_pointer_n_copy(al, args.p, args.size());
                     visit_kwargs(args_, x.m_keywords, x.n_keywords,
                         f->m_args, f->n_args, x.base.base.loc, f,
-                        diags, x.n_member, is_ClassProcedure_With_Self_Argument);
+                        diags, x.n_member, is_nopass);
+                    if (x.n_member >= 1 && is_ClassProcedure && !is_nopass) {
+                        ASR::call_arg_t this_arg;
+                        this_arg.loc = x.m_member[0].loc;
+                        this_arg.m_value = v_expr;
+                        args_.push_front(al, this_arg);
+                    }
                     if( !diags.has_error() ) {
                         if( ASRUtils::select_generic_procedure(args_, *f3, x.base.base.loc,
                             [&](const std::string &msg, const Location &loc) {
@@ -3790,7 +3792,7 @@ public:
                             false) != -1 ) {
                             function_found = true;
                             args.n = 0;
-                            if (is_ClassProcedure) {
+                            if (is_ClassProcedure && !is_nopass) {
                                 args.from_pointer_n_copy(al, args_.p+1, args_.size()-1);
                             } else {
                                 args.from_pointer_n_copy(al, args_.p, args_.size());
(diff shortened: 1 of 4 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3756,13 +3758,25 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        }
+                    }
                     ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(f4);
                     diag::Diagnostics diags;
                     Vec<ASR::call_arg_t> args_;
                     args_.from_pointer_n_copy(al, args.p, args.size());
                     visit_kwargs(args_, x.m_keywords, x.n_keywords,
                         f->m_args, f->n_args, x.base.base.loc, f,
-                        diags, x.n_member);
+                        diags, x.n_member, is_ClassProcedure_With_Self_Argument);
```
gxyd, line 3779: First, you are incorrectly passing `is_ClassProcedure_With_Self_Argument`'s value to `is_nopass` in `visit_kwargs`: https://github.com/lfortran/lfortran/blob/011df7a5b7492672d7bfbfd91690717fd412a9b8/src/lfortran/semantics/ast_common_visitor.h#L10031-L10033

Notice that they are both the 10th argument for them.

Secondly, the purpose for which you've introduced `is_ClassProcedure_With_Self_Argument` is actually really supposed to be taken care of by `is_nopass` itself. So, what you actually need to do is modify the value of `is_nopass` to be passes in call to `visit_kwargs`, when `this` is expected as the first argument, make sure that `is_nopass` is set appropriately.

gxyd (commented):
It would be good to have more test case for it, e.g.; with:
```
procedure, nopass :: Complex32Deque_resize_default
```

etc. 

and also use `if (...) error stop` condition to ensure that things work fine.

src/lfortran/semantics/ast_body_visitor.cpp
gxyd: My point is, I don't think should really be a need for a new argument to be passes to `visit_kwargs` at all really.

adit4443ya: Ahh yes!!
pushed commit  to address your review

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3756,13 +3759,26 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                     visit_kwargs(args_, x.m_keywords, x.n_keywords,
                         f->m_args, f->n_args, x.base.base.loc, f,
-                        diags, x.n_member);
+                        diags, x.n_member, is_nopass);
+                    if (x.n_member >= 1 && is_ClassProcedure && !is_nopass) {
+                        ASR::call_arg_t this_arg;
+                        this_arg.loc = x.m_member[0].loc;
+                        this_arg.m_value = v_expr;
+                        args_.push_front(al, this_arg);
+                    }
```
gxyd, line 3781: I'm sorry, I don't like this approach of having to explicitly push `this_arg` as the first argument, this seems a little hackish to me.

Can you please check how this is handled in case of `ClassProcedure` (i.e. the block of code above this), when `nopass` attribute is used and when not used, and how the `args` is handled in that case?

src/lfortran/semantics/ast_body_visitor.cpp
gxyd: This part is resolved

src/lfortran/semantics/ast_body_visitor.cpp
adit4443ya: Yes sure 
The `ClassProcedure` block relies on `args_with_mdt` to include `this` later, but for `GenericProcedure`, we need args to include `this` during `visit_kwargs` because `select_generic_procedure` performs signature matching immediately. This is why explicit insertion is needed in the `GenericProcedure` loop., same logic as `args_with_mdt`, just applied earlier for `GenericProcedure`

Also Let me take an example here for both the cases with `nopass` and without `nopass`

when used with `nopass` attribute we do not add `this` argument explicitly it is already in included in the `visit_expr_list(x.m_args, x.n_args, args);` function call
So we just need to avoid adding `this`  to the arguments list , 

And when `nopass` attribute is not used we do need to explicitly include the `this` argument in the args list and it needs to be in order hence i used `push_front()`

The condition here `if (x.n_member >= 1 && is_ClassProcedure && !is_nopass)` handles both the cases
,,

Please ask if anything is unclear regarding it

gxyd: > when used with nopass attribute we do not add this argument explicitly it is already in included in the visit_expr_list(x.m_args, x.n_args, args); function call
So we just need to avoid adding this to the arguments list ,

That's an incorrect assumption, you are assuming that `this` always be passed when `nopass` attribute is used.

E.g.;

```f90
module derived_types
  implicit none

  type :: type_nopass
    integer :: value
  contains
    procedure, nopass :: set_value_nopass
    generic :: set_value => set_value_nopass
  end type

contains

  subroutine set_value_nopass(value)
    integer, intent(inout) :: value
    value = 25
  end subroutine

end module

program main
  use derived_types
  implicit none
  integer :: value

  type(type_nopass) :: obj_nopass

  value = 5
  call obj_nopass%set_value(value)
  obj_nopass%value = value
end program
```

currently fails with an error:
```console
(lf) gxyd@Gauravs-MacBook-Air OpenSource/lfortran (gen_proc_1) » lfortran integration_tests/generic_procedure_1.f90
semantic error: Arguments do not match for any generic procedure, set_value
  --> integration_tests/generic_procedure_1.f90:28:3
   |
28 |   call obj_nopass%set_value(value)
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3702,6 +3702,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         }
         Vec<ASR::call_arg_t> args;
         visit_expr_list(x.m_args, x.n_args, args);
+        bool nopass=false;
```
gxyd, line 3705: What does this `nopass` for a subroutine call actually mean? It will remain set to `false`, when there is no keyword arguments present in subroutine call, even though the underlying subroutine implementation has use of `nopass` attribute with it.

E.g.;, consider the pseudo-code:
```f90
  type :: type_nopass
    integer :: value
  contains
    procedure, nopass :: set_value_nopass
    generic :: set_value => set_value_nopass
  end type


call obj_nopass%set_value(value)
```

gxyd:
I've sent a PR here https://github.com/lfortran/lfortran/pull/7202, which intends to fix the same issue that this PR intends to fix.

That PR uses a little more simpler logic to assign `nopass`, and I've added more test cases.

src/lfortran/semantics/ast_body_visitor.cpp
adit4443ya: I do update nopass bool for the case when  keyword argument is used, But not for when it's not used, that's why the example which you gave in the prev comment where 
```
call obj_nopass%set_value(value)
```
was used it gave error, it's easy fix though can be done in subsequent pr or in this one itself

adit4443ya:
>I've sent a PR here https://github.com/lfortran/lfortran/pull/7202, which intends to fix the same issue that this PR intends to fix.

okay  :smiling_face_with_tear:

gxyd:
> > I've sent a PR here #7202, which intends to fix the same issue that this PR intends to fix.
> 
> okay 🥲

No, aim is to receive feedback on that PR, and we can decide if it needs more improvements or something that I might have missed. Considering how much GenericProcedure is used in pFUnit repository, it would make sense if we can test more and more to ensure things work, instead of spending time creating MRE's, which isn't very easy to do with pFUnit, as you also mentioned from your experience with it.

adit4443ya:
> instead of spending time creating MRE's, which isn't very easy to do with pFUnit, as you also mentioned from your experience with it.

Yes I totally agree with this

gxyd:
> What is the status of this? Looks like some comments have to be addressed and conflicts resolved.

I discussed this with @adit4443ya  and a little improved PR i.e. https://github.com/lfortran/lfortran/pull/7202 was merged, and hence we can safely close this PR now.

PR #7201, opened 2025-05-07
Sync

swamishiju:
Working libasr sync


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1947,9 +1947,9 @@ namespace LCompilers {
                 {
                     llvm::Value* i = LLVMUtils::CreateLoad2(llvm::Type::getInt32Ty(context), idx);
                     llvm::Value* l = LLVMUtils::CreateLoad2(llvm::Type::getInt8Ty(context), create_ptr_gep2(
-                        llvm::Type::getInt8Ty(context)->getPointerTo(), left, i));
+                        llvm::Type::getInt8Ty(context), left, i));
                     llvm::Value* r = LLVMUtils::CreateLoad2(llvm::Type::getInt8Ty(context), create_ptr_gep2(
-                        llvm::Type::getInt8Ty(context)->getPointerTo(), right, i));
+                        llvm::Type::getInt8Ty(context), right, i));
```
assem2002, line 1952: That's totally the correct way of creating `GEP` over `char*` to fetch the value of each character.

assem2002 (approved):
It looks good to me.
Most of the changes related to strings are about correcting type when creating `GEP` instruction, which seems good. I don't know how llvm<17 didn't complain before about that while creating IR, but I'm guessing it wasn't tested (no integration tests used these instructions).
I don't see `DescriptorString` in sight, but if it got used by these functions llvm will complain, so that's good!

swamishiju:
> It looks good to me. Most of the changes related to strings are about correcting type when creating `GEP` instruction, which seems good. I don't know how llvm<17 didn't complain before about that while creating IR, but I'm guessing it wasn't tested (no integration tests used these instructions). I don't see `DescriptorString` in sight, but if it got used by these functions llvm will complain, so that's good!

@assem2002 why did we go from create_gep to create_gep2 when I looked at create_gep it seemed to get the type without explicitly mentioning

assem2002:
I think it's because `craete_gep` depends on the metadata of pointer types that used to exist in llvm<17 versions, so we used to be able to fetch the type of the passed pointer; but `create_gep2` is forcing the user of the function to pass the type of the pointer. I hope that makes sense.

PR #6638, opened 2025-03-16
fix: implement mold argument in allocate for Arrays

HarshilShah1804:
Towards #6592

In subsequent work, I will implement it for strings and structs instances also.

PR #7208, opened 2025-05-08
chore: add errors for unsupported `mold` argument in allocate statements

HarshilShah1804:
Apply suggestion from https://github.com/lfortran/lfortran/pull/6638#pullrequestreview-2823049951

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -58,7 +58,7 @@ program continue_compilation_1
     character :: c1
     complex :: c = (1.0, 2.0)
     real a_real(0)
-    integer, allocatable ::  shape_(:), arr4(:)
+    integer, allocatable ::  shape_(:), arr4(:), arr5(:)
     integer, dimension(2, 3) :: matrix
     integer, dimension(4) :: source = [1, 2, 3, 4]
     allocate(shape_(2))
@@ -77,7 +77,7 @@ program continue_compilation_1
     character(len=100) :: filename
     type(MyClass), parameter :: myclass_array(2) = [1, MyClass(10)]
     type(MyClass), parameter :: myclass_array2(2) = [MyClass(1), MyClass(q1)]
-    integer, allocatable :: arr5(:)
+    
 
 
 
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -77,7 +77,7 @@ program continue_compilation_1
     character(len=100) :: filename
     type(MyClass), parameter :: myclass_array(2) = [1, MyClass(10)]
     type(MyClass), parameter :: myclass_array2(2) = [MyClass(1), MyClass(q1)]
-
+    integer, allocatable :: arr5(:)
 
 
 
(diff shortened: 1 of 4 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -77,7 +77,7 @@ program continue_compilation_1
     character(len=100) :: filename
     type(MyClass), parameter :: myclass_array(2) = [1, MyClass(10)]
     type(MyClass), parameter :: myclass_array2(2) = [MyClass(1), MyClass(q1)]
-
+    integer, allocatable :: arr5(:)
```
gxyd, line 80: instead of using a newline do you think it will be better to have this declaration of `arr5(:)` along with `integer, allocatable ::  shape_(:), arr4(:)` here: https://github.com/lfortran/lfortran/blob/47475f0c46c97eec433bec9bc6c084f20832123d/tests/errors/continue_compilation_1.f90#L61 ?

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -685,3 +685,15 @@ semantic error: Required argument `unit` is missing in function call
     |
 308 |     call logger % add_log_file()
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+semantic error: `allocate` statement only accepts four keyword arguments,`stat`, `errmsg`, `source` and `mold`
```
gxyd, line 689: Fix it to?
```suggestion
semantic error: `allocate` statement only accepts four keyword arguments: `stat`, `errmsg`, `source` and `mold`
```

gxyd (commented):
So, I guess this PR addresses both the comments left here: https://github.com/lfortran/lfortran/pull/6638#pullrequestreview-2823049951, right?

tests/errors/continue_compilation_1.f90
gxyd: Ofcourse, you'll have to check that the suggested change doesn't cause un-necessary reference updates.

You current change does ensure that there aren't any unnecessary reference updates, but the only reason to suggest this change is not use a new unnecessary line for the declaration.

HarshilShah1804:
@gxyd I have applied the changes suggested by you. Please review and tell me if anything else is required.

gxyd (approved):
Looks good to me, thanks @HarshilShah1804 for addressing the comments.

tests/errors/continue_compilation_1.f90
```diff
@@ -77,7 +77,7 @@ program continue_compilation_1
     character(len=100) :: filename
     type(MyClass), parameter :: myclass_array(2) = [1, MyClass(10)]
     type(MyClass), parameter :: myclass_array2(2) = [MyClass(1), MyClass(q1)]
-
+
```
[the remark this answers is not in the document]
HarshilShah1804: Sorry, Will take care about this next time onwards.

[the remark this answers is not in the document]
gxyd: Ah, missed this in my review.

Issue #7103, opened 2025-04-29
`Token_loc` function throws error while storing location information for files having procedures used without contains


parth121101:
Related links: 
https://github.com/lfortran/lfortran/pull/7042
https://github.com/lfortran/lfortran/pull/7042#discussion_r2064270939

Definition of `token_loc` function:
```
loc.first = tok-string_start;
loc.last = cur-string_start-1;
```

While the one I needed to use was:
```
loc.first = cur - string_start;
loc.last = cur - string_start;
```

Need to find a bug in `token_loc` function implementation.

PR #7227, opened 2025-05-10
Used implemented `token_loc` function to find location information

parth121101:
Issue: https://github.com/lfortran/lfortran/issues/7103

Definition of token_loc function:
It takes `tok`, `cur` and `loc` arguments and sets the first and last position in `loc` variable.
It sets the first and last position in the following manner:
```
loc.first = tok-string_start;
loc.last = cur-string_start-1;
```

For the following error test:
```
PROGRAM continue_compilation_ff_2
print *, "This is a test program"
SUBROUTINE faulty_subroutine(a, b, c)
INTEGER, INTENT(IN) :: sub_a
END SUBROUTINE faulty_subroutine
END PROGRAM continue_compilation_ff_2
```

As suggested here (https://github.com/lfortran/lfortran/pull/7042#discussion_r2056862832) I tried to use `token_loc` function by minor manipulation.


HarshitaKalani:
The CI checks failed due to: https://github.com/lfortran/lfortran/issues/7230

Issue #7245, opened 2025-05-11
Comparison when Lhs and Rhs are pointers


adit4443ya:
## MRE

```fortran
program test
  implicit none

  real, pointer :: d1, d2
  real, target :: t1, t2

  ! Initialize target objects
  t1 = 4
  t2 = 5
  ! Assign pointers
  d1 => t1
  d2 => t2

  if ( d1 > d2) error stop "Test 1 failed"

end program test
```

in main

```shell
 aditya-trivedi   lfortran    main ≡  ?1 ~1    lfortran --version
LFortran version: 0.52.0-89-g6668fd8d9-dirty
Platform: Linux
LLVM: 11.0.1
Default target: x86_64-unknown-linux-gnu
 aditya-trivedi   lfortran    main ≡  ?1 ~1    lfortran integration_tests/complex_op.f90 
semantic error: Compare: only Integer or Real can be on the LHS and RHS. If operator is .eq. or .neq. then Complex type is also acceptable
  --> integration_tests/complex_op.f90:46:7
   |
46 |   if (d1 > d2) then
   |       ^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
 aditya-trivedi   lfortran    main ≡  ?1 ~1    gfortran integration_tests/complex_op.f90 
 aditya-trivedi   lfortran    main ≡  ?1 ~1    echo $?
0
```

adit4443ya:
Closing as fixed by #7246

PR #7239, opened 2025-05-10
fix: allow pointer comparison of Real, Ints and Complex

gxyd:
## Description

This is an MRE extracted from pFUnit third party repository (towards: https://github.com/lfortran/lfortran/issues/6930)

```f90
program compare_pointers_01
    implicit none

    integer, target :: int_num1, int_num2
    integer, pointer :: int_ptr1 => null(), int_ptr2 => null()

    int_num1 = 1
    int_num2 = 2

    int_ptr1 => int_num1
    int_ptr2 => int_num2

    print *, int_ptr1 < int_ptr2
end program compare_pointers_01

```

For the above Fortran program, LFortran raises a semantic error currently:
```console
semantic error: Compare: only Integer or Real can be on the LHS and RHS. If operator is .eq. or .neq. then Complex type is also acceptable
  --> test.f90:13:14
   |
13 |     print *, int_ptr1 < int_ptr2
   |              ^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

This PR intends to fix that issue.

PR #7246, opened 2025-05-11
Handle the case of LHS and RHS being pointers in less than condition

adit4443ya:
Towards #7245

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -668,8 +668,8 @@ inline static void visit_Compare(Allocator &al, const AST::Compare_t &x,
         }
     }
     // Cast LHS or RHS if necessary
-    ASR::ttype_t *left_type = ASRUtils::type_get_past_allocatable(ASRUtils::expr_type(left));
-    ASR::ttype_t *right_type = ASRUtils::type_get_past_allocatable(ASRUtils::expr_type(right));
+    ASR::ttype_t *left_type = ASRUtils::type_get_past_allocatable_pointer(ASRUtils::expr_type(left));
+    ASR::ttype_t *right_type = ASRUtils::type_get_past_allocatable_pointer(ASRUtils::expr_type(right));
     ASR::ttype_t *left_type2 = ASRUtils::type_get_past_array(left_type);
     ASR::ttype_t *right_type2 = ASRUtils::type_get_past_array(right_type);
 
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -319,6 +319,7 @@ RUN(NAME volatile_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 
 RUN(NAME cond_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)
 RUN(NAME cond_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c fortran)
+RUN(NAME cond_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 
 RUN(NAME expr_01 FAIL LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)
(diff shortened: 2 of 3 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -670,6 +670,8 @@ inline static void visit_Compare(Allocator &al, const AST::Compare_t &x,
     // Cast LHS or RHS if necessary
     ASR::ttype_t *left_type = ASRUtils::type_get_past_allocatable(ASRUtils::expr_type(left));
     ASR::ttype_t *right_type = ASRUtils::type_get_past_allocatable(ASRUtils::expr_type(right));
+    left_type = ASRUtils::type_get_past_pointer(left_type);
+    right_type = ASRUtils::type_get_past_pointer(right_type);
```
kmr-srbh, line 674:
```suggestion
    ASR::ttype_t *left_type = ASRUtils::type_get_past_allocatable_pointer(ASRUtils::expr_type(left));
    ASR::ttype_t *right_type = ASRUtils::type_get_past_allocatable_pointer(ASRUtils::expr_type(right));
```

kmr-srbh (approved):
The changes look good, left a minor suggestion.

src/lfortran/semantics/ast_common_visitor.h
adit4443ya: yes sure
One CI was failing for the C-backend which i checked it gave similar errors for pointer MREs in main too (as i think C-backend is immature i guess so) so i skipped C backend in here for now

kmr-srbh: Yes, the C backend is not complete yet.

kmr-srbh (approved):

kmr-srbh:
The `wasm` backend fails:

```console
18/226 Test  #20: cond_03 ..........................***Failed    0.18 sec
  throw err;
  ^

Error: Cannot find module '/home/runner/work/lfortran/lfortran/integration_tests/test-wasm/cond_03.js'
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.20.4
```

kmr-srbh:
I think you are already aware of it, but I am still posting it here on the thread. Here is a case that needs to be handled in the `llvm` backend:

```fortran
program cond_03
  implicit none

  real, pointer :: d1, d2
  real, target :: t1, t2

  ! Initialize target objects
  t1 = 4
  t2 = 5
  ! Assign pointers
  d1 => t1
  d2 => t2

  print *, d1 > d2

end program cond_03
```
```console
 F
(lf) saurabhkumar@Mac lfortran % lfortran test.f90            
.
.
.
Invalid operand types for FCmp instruction
  %4 = fcmp ogt float* %2, %3

Invalid operand types for FCmp instruction
  %4 = fcmp ogt float* %2, %3



```

If this is not required for `pFUnit`, you may open up an issue for this. I think we need to handle the condition evaluation similar to how we handle it in the `if` clause.

adit4443ya:
As Discussed with @kmr-srbh I'm currently disabling wasm backend for this test 
rest of the things works , ready to merge i guess after CI passes

adit4443ya:
>If this is not required for pFUnit, you may open up an issue for this. I think we need to handle the condition evaluation similar to how we handle it in the if clause.

As for this I will investigate it once this PR is merged.(As it is not required for pfUnit for now but still it should work as told by saurabh (Note:- the error which is shown in above saurabh's comment comes on this branch (After processing the comparison condition (which is done correctly by this fix), on main it gives same error as #7245 which this PR solves)

kmr-srbh:
I see that the test added in #7239 had a `print` statement and a comment stating that it fails. Maybe Gaurav would be able to tell if he found a usage. If not, I think we can open an issue and move ahead with the compilation.

gxyd:
> I see that the test added in #7239 also had a `print` statement and a comment stating that it fails. Maybe Gaurav would be able to tell if he found a usage. If not, I think we can open an issue and move ahead with the compilation.

No, there was no usage of printing of pointer comparison in pFUnit compilation, it was a second nature for me to write a print statement (as I have always done with other test cases in Lfortran).

So yes, can have a separate issue and keep this PR separate from it.

kmr-srbh (approved):

Issue #3382, opened 2024-02-09
compiling stdlib


Pranavchiku:
Master issue to track compilation of various modules, test-drive part of `stdlib`.

## Modules

There are total of 21 modules, listed below:

* [ ] array
* [ ] ascii
* [ ] bitsets
* [ ] error
* [ ] hash_procedures
* [ ] hashmaps
* [ ] io
* [ ] linalg #3323
* [ ] logger
* [ ] math #3046
* [ ] optval
* [ ] quadrature
* [ ] random
* [ ] selection
* [ ] sorting
* [ ] specialfunctions_gamma #3000
* [ ] stats #3052
* [ ] stats_distribution_exponential
* [ ] stats_distribution_normal
* [ ] stats_distribution_uniform
* [ ] string_type #3201

## CI
Of these 21 packages, the listed modules are tested at our CI ( or PRs are ready to test those ):

* linalg #3381
* specialfunctions_gamma #3036
* math #3322

czgdp1807:
I think better metric would be to list how many tests we can compile and run.

Thirumalai-Shaktivel:
# Stdlib Milestone
Currently, we are able to compile `27` out of `33` modules and run `8` out of `20` tests. Out of which, `8` modules compile without workarounds. Note: Some of the modules that compile are not tested; it might require workarounds if we enable tests.

## Modules

- [x] stdlib_error
  - [ ] estop
- [x] stdlib_ansi (doesn't require workaround)
  - [x] stdlib_ansi_operator
  - [x] stdlib_ansi_to_string
- [x] stdlib_array
- [x] stdlib_ascii
- [x] stdlib_bitsets (doesn't require workaround)
  - [x] stdlib_bitsets_64
  - [x] stdlib_bitsets_large
- [ ] stdlib_hash_32bit
  - [ ] stdlib_hash_32bit_fnv
  - [ ] stdlib_hash_32bit_nm
  - [ ] stdlib_hash_32bit_water
- [ ] stdlib_hash_64bit
  - [ ] stdlib_hash_64bit_fnv
  - [ ] stdlib_hash_64bit_pengy
  - [ ] stdlib_hash_64bit_spookyv2
- [ ] stdlib_hashmaps
  - [ ] stdlib_hashmap_chaining
  - [ ] stdlib_hashmap_open
- [ ] stdlib_hashmap_wrappers
- [x] stdlib_io
- [x] stdlib_io_npy
  - [x] stdlib_io_npy_load
[...]

PR #7253, opened 2025-05-11
Iostat

HiteshGhanchi:
#7149


jinangshah21:
`loadnpy` test from stdlib is failing, we need to fix that.

HiteshGhanchi:
Yeah 
Is there a way to know what code that is ?
From CI , I can't get why it is failing.
Thanks

jinangshah21:
You can find steps to compile stdlib at #3382

PR #7273, opened 2025-05-13
fix: [fortran_mpi] use MPI wrappers from lfortran/fortran_mpi

gxyd:
## Description

this commit deletes the existing wrappers in POT3D's branch and instead use the cloned repo to obtain MPI wrappers, currently this is to showcase that CI works with the latest main of lfortran/fortran_mpi, maybe we resort to use a release of lfortran/fortran_mpi?

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -148,8 +148,13 @@ time_section "🧪 Testing Numerical Methods Fortran" '
 time_section "🧪 Testing POT3D" '
   git clone https://github.com/gxyd/pot3d.git
   cd pot3d
-  git checkout -t origin/lf_hdf5_mpi_namelist_global_workarounds
-  git checkout 83e1e90db7e7517fcbe8c7bce5ba309addfb23f6
+  git checkout -t origin/lf_hdf5_fortranMPI_namelist_global_workarounds
+  git checkout fc2788ca07c065152b2cd6f3d0eb1abfbee9ffcf
+
+  git clone https://github.com/lfortran/fortran_mpi
+  cp fortran_mpi/src/mpi.f90 ./src/
+  cp fortran_mpi/src/mpi_c_bindings.f90 ./src/
+  cp fortran_mpi/src/mpi_constants.c ./src/
 
   print_subsection "Building with default flags"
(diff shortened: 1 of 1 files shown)


gxyd:
If the CI passes, then we'll have a question to whether we make a release of [lfortran/fortran_mpi](https://github.com/lfortran/fortran_mpi) repository and use that instead.

The benefit of using a release of fortran_mpi instead of the latest clone will be that, let's say for reason a change in fortran_mpi (passes fortran_mpi's own CI), but when we clone it here, it fails here (very much possible also), then we'll have a trouble with PR's failing CI checks and wasting time of devs. I'm in favor of using release here, instead of what's currently done in this PR.

gxyd:
Failing checks, fixing it.

adit4443ya:
I guess the SegFault error comes only with MPI and not with MPICH
Might be related MPI_HANDLE_KIND ?

gxyd:
> I guess the SegFault error comes only with MPI and not with MPICH
> Might be related MPI_HANDLE_KIND ?

I think you are right, we've OpenMPI installed in CI.yml configuration, but I used MPICH locally on my machine. This is helpful.

gxyd:
Thank you so much for your help @adit4443ya , I was a little confused when I saw the CI failing here earlier.

ci/test_third_party_codes.sh
```diff
@@ -148,14 +148,22 @@ time_section "🧪 Testing Numerical Methods Fortran" '
...
+  cd fortran_mpi
+  git checkout 31033d3c8af32c4c99fac803c161e6731bc39a78
+  cp src/mpi.f90 ../src/
+  cp src/mpi_c_bindings.f90 ../src/
+  cp src/mpi_constants.c ../src/
+  cd ..
 
   print_subsection "Building with default flags"
-  FC=$FC ./build_and_run.sh
+  FC="$FC --cpp -DOPEN_MPI=yes" ./build_and_run.sh
```
Pranavchiku, line 163: can we not test `FC="$FC" ./build_and_run.sh` now? If we can, let's test that as well.

gxyd: No, we've need C preprocessing, and hence can't test without `--cpp`

Issue #7275, opened 2025-05-13
no matching function found for generic procedure


gxyd:
## Minimal Reproducible Example

```f90
module gFTL2_Integer32Complex32Map
   use iso_fortran_env
   implicit none

   type KeywordEnforcer
   end type KeywordEnforcer

   public :: Integer32Complex32Map
   public :: Integer32Complex32Pair

   type :: Integer32Complex32Pair
      integer(kind=INT32) :: first
      complex(kind=REAL32) :: second
   end type Integer32Complex32Pair

   type, abstract :: map_s_BaseNode
   end type map_s_BaseNode

   type, extends(map_s_BaseNode) ::  map_s_Node
      type(map_s_Node), pointer :: parent => null()
   end type map_s_Node

   type :: map_Set
      private
      class(map_s_BaseNode), allocatable :: root
   contains
      procedure :: insert_single => map_s_insert_single
      generic :: insert => insert_single
   end type map_Set

   type :: Integer32Complex32Map
      private
      type(map_Set) :: tree
   contains
      procedure :: of => map_of
   end type Integer32Complex32Map
   contains

   recursive subroutine map_s_insert_single(this, value, unused, is_new)
      class(map_Set), target, intent(inout) :: this
      type(Integer32Complex32Pair), intent(in) :: value
      type (KeywordEnforcer), optional :: unused
      logical, optional, intent(out) :: is_new
   end subroutine map_s_insert_single

   subroutine map_of(this, key)
       class(Integer32Complex32Map), target, intent(inout) :: this
       integer(kind=INT32), intent(in) :: key
       type(Integer32Complex32Pair) :: p
       logical :: is_new
       p%first= key
       call this%tree%insert(p, is_new=is_new)
   end subroutine map_of
end module gFTL2_Integer32Complex32Map

```

For the above Fortran program, on it's compilation I get the below error:
```console
(lf) gxyd@Gauravs-MacBook-Air Downloads/pFUnit-4.12.0 (lf3) » lfortran new.f90
semantic error: No matching function found for the call to generic procedure, insert
  --> new.f90:52:8
   |
52 |        call this%tree%insert(p, is_new=is_new)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.

[...]

PR #7276, opened 2025-05-13
fix: correctly assign args size when visiting kwargs

gxyd:
## Description
using is_method is correct, as we only need to do offseting on the basis of whether a procedure needs to account for "this/self" or not

Fixes https://github.com/lfortran/lfortran/issues/7275

I'll keep this PR in draft mode, and try to see if I can add more tests for it, once I'm back to work in a hour or not.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -10139,7 +10139,7 @@ public:
         std::vector<std::string> fn_args2 = convert_fn_args_to_string(fn_args, fn_n_args);
 
         int offset = args.size();
-        for (int i = 0; i < (int)fn_n_args - offset - (int)type_bound; i++) {
+        for (int i = 0; i < (int)fn_n_args - offset - is_method; i++) {
             ASR::call_arg_t call_arg;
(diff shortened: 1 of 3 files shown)


gxyd:
This is now ready for review.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10139,7 +10139,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         std::vector<std::string> fn_args2 = convert_fn_args_to_string(fn_args, fn_n_args);
 
         int offset = args.size();
-        for (int i = 0; i < (int)fn_n_args - offset - (int)type_bound; i++) {
+        for (int i = 0; i < (int)fn_n_args - offset - is_method; i++) {
```
gxyd, line 10142: `type_bound` is an integer, denoting level of nesting e.g. if the subroutine (or function) call is of the form `this % x % y % FUNC (...)` the value of `type_bound` is 3, i.e. one each for level of `%`

We only need to account for whether we need to offset for `this` or not, for that knowing whether atleast one `%` is present is enough.

Issue #7264, opened 2025-05-12
comparison of pointer and value of custom type fails


gxyd:
## Minimal Reproducible Example

```f90
module gFTL2_Integer32Complex32Map
   implicit none

   type :: Integer32Complex32Pair
      integer :: first
      complex :: second
      contains
   end type Integer32Complex32Pair

   interface operator(==)
      module procedure map_p_equal
   end interface operator(==)

   contains

   logical function map_p_equal(a, b) result(equal)
      type(Integer32Complex32Pair), target, intent(in) :: a
      type(Integer32Complex32Pair), target, intent(in) :: b
   end function map_p_equal

   subroutine map_find_basic(value)
      type(Integer32Complex32Pair), intent(in) :: value
      type(Integer32Complex32Pair), pointer :: ptr
      if (ptr==value) print *, "Viola"
    end subroutine map_find_basic
end module gFTL2_Integer32Complex32Map

```

For the above Fortran program, it fails with the below error:
```console
(lf) gxyd@Gauravs-MacBook-Air Downloads/pFUnit-4.12.0 (lf3) » lfortran new.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2685
    LCompilers::print_stack_on_segfault();
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2616
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 1111
    lcompilers_unique_ID = compiler_options.generate_object_code ? get_unique_ID() : "";
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 112
    if (res.ok) {
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 5181
    b.is_body_visitor = true;
[...]

PR #7279, opened 2025-05-13
fix: do pointer comparison of custom types

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/7264

The fix is pretty much similar to what we did in: https://github.com/lfortran/lfortran/pull/7246


gxyd:
Fortran backend fails here, I'll disable that.

gxyd:
> Or just implement the Fortran backend's `visit_OverloadedCompare()` instead of disabling it. You can disable here and send a subsequent PR adding this.

Yup, I'll do that.

PR #7290, opened 2025-05-13
Test : Use a test in both integration and reference tests

assem2002:
Minor change, suggested here https://github.com/lfortran/lfortran/pull/7272#discussion_r2086953782


assem2002:
It's a minor, so I think I'll enable auto merge for it.

assem2002:
> Do we need to test LLVM here?
actually no.

PR #7293, opened 2025-05-14
enh: enable commented tests for EN format

HarshitaKalani:


tests/format3.f90
```diff
@@ -19,12 +18,12 @@ program format3
     print '(EN12.4)', num1
     print '(EN12.4)', num2
     print '(EN10.4)', num3
-    ! print '(EN10.4)', small_num4
-    ! print '(EN10.4)', under_power_10
+    print '(EN10.4)', small_num4
```
gxyd, line 21: Same for this as well, can we please have more tests for this?

tests/format3.f90
```diff
@@ -19,12 +18,12 @@ program format3
...
+    print '(EN10.4)', under_power_10
```
gxyd, line 22: Can we please have more variety of tests for `under_power_10`?

I went through the logic and I've a feeling it might be working well for this particular format description with these specific values, but may not work for many other.

gxyd: Ok, I tried these values locally :
```console
   print '(EN10.1)', under_power_10
```

and the output via LFortran is:
```
1000.0E+00
```

while, considering that it's an EN notation, I expect it to be more like 1.0E+03 or something like that.

gxyd: Maybe you can open up a separate issue for this where you fix it and test it with different variations of `x` and `y` in `ENx.y`?

gxyd (approved):
I'm approving this with a hope that the comments will be addressed by opening a separate issue and resolved in subsequent PR. Thanks @HarshitaKalani

tests/format3.f90
HarshitaKalani: Thanks for the review, @gxyd! You're absolutely right — there can be so many edge cases for ENw.dEe depending on the specific values being printed that it’s easy to miss some. My plan is to first get a reasonably complete and consistent base implementation for all the formats that works across the common cases. Once that’s in place, I’ll start building out a more robust solution with thorough test coverage to ensure all edge cases are handled properly.

Do you think that approach makes sense, or would it be better to aim for a more comprehensive implementation from the start?

gxyd:
> I think this failure is unrelated.

Yes, that's right.

tests/format3.f90
gxyd: > Thanks for the review, @gxyd! You're absolutely right — there can be so many edge cases for ENw.dEe depending on the specific values being printed that it’s easy to miss some. My plan is to first get a reasonably complete and consistent base implementation for all the formats that works across the common cases. Once that’s in place, I’ll start building out a more robust solution with thorough test coverage to ensure all edge cases are handled properly.

Agreed.

I'll merge this PR as soon as the CI at the *main* is fixed.

HarshitaKalani: Perfect. Thanks!

gxyd: If you could please open an issue for one example that I mentioned, that would be great, and I'll resolve this conversation as well. Thank you.

gxyd:
CI failure issue: https://github.com/lfortran/lfortran/issues/7295

tests/format3.f90
HarshitaKalani: Done: https://github.com/lfortran/lfortran/issues/7300

gxyd: That's perfect.

gxyd:
I've rebased the changes on top of main now.

PR #7294, opened 2025-05-14
fix: cast integer constant properly when passing to API

jinangshah21:
Fixes LLVM 20 failure for `stdlib` (Refer CI update PR: #7261)


gxyd (commented):
With this as a separate PR without any test case, and only a link to the failing CI job, i.e. no CI changes in this PR either, how do I know whether this actually fixes the linked CI failure?

jinangshah21:
I don't have MRE, but it fixes stdlib segfault (which I checked locally). Also that failure was only for LLVM 20, So I suspected a small issue and fixing that works for me locally, so I didn't wasted my time to create MRE. The change in this PR is logical also.

PR #7303, opened 2025-05-14
ci: set `fortran_mpi` and `POT3D` with `fortran_mpi`

Pranavchiku:


adit4443ya:
This is good way to test it (As a separate 3rd party test) .

Pranavchiku:
> We can add MPICH test in subsequent PRs?

Yes, once this gets merged, you can send a new PR.

Issue #7305, opened 2025-05-14
semantic error: function not found


gxyd:
## Minimal Reproducible Example

```f90
module gFTL2_Integer32Complex32OrderedMap
   use iso_fortran_env
   implicit none
   type KeywordEnforcer
   end type KeywordEnforcer
   private
   public :: Integer32Complex32OrderedMap

   type :: omap_I_Set
   contains
   end type omap_I_Set

   type :: omap_Map
      private
      type(omap_I_Set) :: tree
   contains
      procedure :: at_rc => omap_I_at_rc
      generic :: at => at_rc
   end type omap_Map

   type :: omap_Vector
   contains
      procedure :: at_default => omap_K_at_default
      generic :: at => at_default
   end type omap_Vector
   type :: Integer32Complex32OrderedMap
      private
      type(omap_Map) :: map
   contains
      procedure :: insert_key_value => omap_insert_key_value
      generic :: insert => insert_key_value
   end type Integer32Complex32OrderedMap
   contains
      function omap_I_at_rc(this, key, rc) result(res)
         complex(kind=REAL32), pointer :: res
         class(omap_Map), target, intent(in) :: this
         integer(kind=INT32), intent(in) :: key
         integer, intent(out), optional :: rc
      end function omap_I_at_rc
   function omap_K_at_default(this, i, rc) result(res)
      class(omap_Vector), target, intent(in) :: this
      integer, intent(in) :: i
      integer(kind=INT32), pointer :: res
      integer, intent(out) :: rc
   end function omap_K_at_default
      recursive subroutine omap_insert_key_value(this, key, value)
         class (Integer32Complex32OrderedMap), intent(inout) :: this
         integer(kind=INT32), intent(in) :: key
         complex(kind=REAL32), intent(in) :: value
         integer :: status
         complex(kind=REAL32), pointer :: tmp
         tmp => this%map%at(key, rc=status)
      end subroutine omap_insert_key_value
end module gFTL2_Integer32Complex32OrderedMap

```

For the above Fortran program, LFortran fails to compile it:
```console
(pfunit) gxyd@Gauravs-MacBook-Air Downloads/pFUnit-4.12.0 (lf3) » lfortran new.f90
semantic error: at_rc is not a function.
  --> new.f90:52:17
   |
52 |          tmp => this%map%at(key, rc=status)
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


[...]

PR #7309, opened 2025-05-14
Fix: Handle FunctionCall to ClassProcedures for Generics 

adit4443ya:
Towards #7305

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2130,6 +2130,8 @@ RUN(NAME polymorphic_arguments_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME polymorphic_class_compare LABELS gfortran)
 
+RUN(NAME nested_struct_proc_01 LABELS gfortran)
+
 RUN(NAME binop_of_struct_instance_in_function_call gfortran llvm fortran)
 
 RUN(NAME test_call_arg_structinstance_member gfortran llvm fortran)
--- a/integration_tests/nested_struct_proc_01.f90
+++ b/integration_tests/nested_struct_proc_01.f90
@@ -1,0 +1,70 @@
+module gFTL2_Integer32Complex32OrderedMap
+   use iso_fortran_env
+   implicit none
+
+   integer, parameter :: I32 = int32
+   integer, parameter :: R32 = real32
+
+   type :: omap_I_Set
+   contains
+   end type omap_I_Set
+
+   type :: omap_Map
+      private
+      type(omap_I_Set) :: tree
+      complex(kind=R32) :: stored_value  ! Module-level target
+   contains
+      procedure :: at_rc => omap_I_at_rc
+      generic :: at => at_rc
+   end type omap_Map
+
+   type :: Integer32Complex32OrderedMap
+      private
+      type(omap_Map) :: map
+   contains
+      procedure :: insert_key_value => omap_insert_key_value
+      generic :: insert => insert_key_value
+   end type Integer32Complex32OrderedMap
+
+contains
+
+   function omap_I_at_rc(this, key, rc) result(res)
+      class(omap_Map), target, intent(in) :: this
+      integer(kind=I32), intent(in) :: key
+      integer, intent(out) :: rc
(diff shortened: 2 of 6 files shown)


Pranavchiku (commented):
Let's get the CI green, marking it to draft.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5077,6 +5077,10 @@ ptr_type[ptr_member] = llvm_utils->get_type_from_ttype_t_util(
                         ASR::is_a<ASR::FunctionType_t>(*value_type)) {
                 llvm_value = llvm_utils->CreateLoad(llvm_value);
                 builder->CreateStore(llvm_value, llvm_target);
+            } else if (ASR::is_a<ASR::Pointer_t>(*value_type) && ASR::is_a<ASR::Pointer_t>(*target_type) &&
+                       ASR::is_a<ASR::FunctionCall_t>(*x.m_value)) {
+                builder->CreateStore(llvm_value, llvm_target);
+
```
adit4443ya, line 5083: The need for this Change is that First of all it gave me error like this 

```
  store %complex_4 %21, %complex_4** %tmp, align 1
 %complex_4*
 ```
 
 the reason is quite simple when we debug its generated LLVM IR then we can see that 
 the statement 
 ```
 tmp => this%map%at(key, rc=status) // where tmp and functionCall Return Type are both Pointers
 ```
 is An Associate Node with This ASR
 
 
 
 ```shell
 (Associate
                                        (Var 10 tmp)
                                        (FunctionCall
                                            10 1_omap_i_at_rc
                                            7 at
                                            [((StructInstanceMember
                                                (Var 10 this)
                                                10 1_integer32complex32orderedmap_map
                                                (StructType
                                                    []
                                                    []
                                                    .true.
                                                    2 omap_map
                                                )
                                                ()
                                            ))
                                            ((Var 10 key))
                                            ((Var 10 status))]
                                            (Pointer
                                                (Complex 4)
                                            )
                                            ()
                                            ()
                                        )
                                        
```

Generates This Code 

```
%20 = call %complex_4* @__module_gftl2_integer32complex32orderedmap_omap_i_at_rc(%omap_map_polymorphic* %17, i32* %key, i32* %status)

//Below 2 Instructions Are Generated By Associate Node
  %21 = load %complex_4, %complex_4* %20, align 1 // This Load Instruction is not required here
  store %complex_4 %21, %complex_4** %tmp, align 1 //This is Where error occurs due to type mismatch
  ```
  The extra load condition is not required as in the function return we already do a load 
  
  ```
  define %complex_4* @__module_gftl2_integer32complex32orderedmap_omap_i_at_rc(%omap_map_polymorphic* %this, i32* %key, i32* %rc) {
  // its body
  %5 = load %complex_4*, %complex_4** %res, align 8
  ret %complex_4* %5
  ```


 I Hope I explained it correctly

**Comment by adit4443ya** [comment]
Can We do one thing at a time ?
Make this PR to just generate the ASR,
and make other PR in which i handle the llvm error?

**Comment by gxyd** [comment]
> Can We do one thing at a time ?
> Make this PR to just generate the ASR,
> and make other PR in which i handle the llvm error?

That would be a good approach.

**Comment by adit4443ya** [comment]
Ok I got the origin of error it is totally unrelated to our Issue
The CI failure was because of the print statement when used with --fast flag 

this is the small MRE for the same

```fortran
program main 
    implicit none
    complex, target :: a
    complex, pointer :: b
    a=(1.0, 2.0)
    b => a
    print *, "a = ", b
    print *, "a = ", real(b), aimag(b)
end program main
```
```console
 aditya-trivedi   lfortran    main ≡  ?2    lfortran integration_tests/pointer_03.f90 --fast 
a =     (1.00000000e+00, 2.00000000e+00)
Illegal instruction (core dumped)
```

So to unblock us I will re-add that commit which makes the MRE in this pr to compile and run and will drop ASR test .
And in the MRE in this PR i will just remove this print statement which causes error with --fast flag
and in this way we can go ahead

adit4443ya:
It's here https://github.com/lfortran/lfortran/pull/7321

adit4443ya:
I opened issue at https://github.com/lfortran/lfortran/issues/7322

Issue #7297, opened 2025-05-14
Handle `binOp` array passed as argument to a function


jinangshah21:
MRE:
```fortran
program test_peek
  implicit none
  integer, parameter :: n = 5
  integer :: offset(n)
  integer :: pos
  integer :: offset_date

  offset = (/1, 2, 3, 4, 5/)
  pos = 10
  offset_date = 3

  print *, peek(pos + offset(:offset_date))

contains

function peek(x) result(y)
    integer, intent(in) :: x(:)
    integer :: y(size(x))
    y = x * 10
  end function peek

end program test_peek
```

LFortran Error:
```console
(lf20) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90
LCOMPILERS_ASSERT failed: /home/jinang_shah/Desktop/lfortran/src/libasr/asr_utils.cpp
function make_ArraySize_t_util(), line number 2227 at 
ASR::is_a<ASR::Var_t>(*a_v) || ASR::is_a<ASR::StructInstanceMember_t>(*a_v) || ASR::is_a<ASR::FunctionParam_t>(*a_v)
ERROR MESSAGE:
Found ASR::exprType::76
^[[ATraceback (most recent call last):
  Binary file "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2688, in main
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2616, in main_app(int, char**)
    if (endswith(arg_file, ".f90") || endswith(arg_file, ".f") ||
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 1110, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    t1 = std::chrono::high_resolution_clock::now();
[...]

PR #7317, opened 2025-05-15
Fix for `BinOp(ArraySection, Scalar)` 

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/7297


src/libasr/asr_utils.cpp
```diff
@@ -2190,8 +2190,11 @@ ASR::asr_t* make_ArraySize_t_util(
         }
     }
     if( is_binop_expr(a_v) && for_type ) {
-        if( ASR::is_a<ASR::Var_t>(*extract_member_from_binop(a_v, 1)) ) {
-            return make_ArraySize_t_util(al, a_loc, extract_member_from_binop(a_v, 1), a_dim, a_type, a_value, for_type);
+        if( !ASR::is_a<ASR::ArrayBroadcast_t>(*extract_member_from_binop(a_v, 1)) &&
+            (ASR::is_a<ASR::Var_t>(*extract_member_from_binop(a_v, 1)) ||
+            ASR::is_a<ASR::ArraySection_t>(*extract_member_from_binop(a_v, 1))) ) {
+            return make_ArraySize_t_util(al, a_loc, ASRUtils::get_past_array_broadcast(
+                extract_member_from_binop(a_v, 1)), a_dim, a_type, a_value, for_type);
```
gxyd, line 2197: If above we already have a condition that `!ASR::is_a<ASR::ArrayBroadcast_t>(*extract_member_from_binop(a_v, 1))` is true, then why do we need to get past array broad cast, i.e. we don't really need `ASRUtils::get_past_array_broadcast` call over `extract_member_from_binop(a_v, 1)` maybe?

czgdp1807: Ah! Yes. Forgot to remove. Let me fix.

PR #7321, opened 2025-05-16
Fix: Handle Associate when Value is FunctionCall with return type as pointer

adit4443ya:
Towards https://github.com/lfortran/lfortran/pull/7309#issuecomment-2885633480


adit4443ya:
Can we merge this ? 
CC:@gxyd @certik

gxyd (approved):
Yes, looks good to me, thanks @adit4443ya

PR #7323, opened 2025-05-16
enh: catch ICE to throw a nice error

sriganeshres:
fixes: #6976


sriganeshres:
@certik Can You review this?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -8944,13 +8945,19 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    loc, true, compiler_options.po, true,
+                    [&](const std::string &msg, const Location &loc) {
+                            diag.add(Diagnostic(msg, Level::Error, Stage::Semantic, {Label("", {loc})}));
+                            throw SemanticAbort();
+                        }, lm, compiler_options.separate_compilation
+                    );
+        } catch ( LCompilersException &e ) {
+            std::cerr << e.name() + ": " << e.msg() << std::endl;
+            std::abort();
+        }
```
gxyd, line 8960: This approach may not be the best, because it will catch any possible ICE (if you go inside `load_module` there are other possible places where ICE is raised, we wouldn't want to change that for everything) and throw it now as a `std::cerr`, which isn't what we want.

This is the place where ICE is being throw from currently: https://github.com/lfortran/lfortran/blob/aba1a616dccc299442489745254333f291df758e/src/libasr/modfile.cpp#L149-L151

maybe throw an error there itself, would it work?

gxyd: If you think that will work, maybe then you can also change the title of the PR, to something like: `fix: raise standard error when incompatible mod file is used`?

gxyd:
Kindly move this to ready for review, once this is.

src/libasr/modfile.cpp
```diff
@@ -147,7 +147,10 @@ inline void load_serialised_asr(const std::string &s, std::string& asr_binary,
     }
     std::string version = b.read_string();
     if (version != LFORTRAN_VERSION) {
-        throw LCompilersException("Incompatible format: LFortran Modfile was generated using version '" + version + "', but current LFortran version is '" + LFORTRAN_VERSION + "'");
+        std::cerr << "Incompatible format: LFortran Modfile was generated using version '"
+                      + version + "', but current LFortran version is '" + LFORTRAN_VERSION + "'"
+                  << std::endl;
+        std::abort();
```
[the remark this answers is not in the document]
sriganeshres: Like throw an error and catch it else where. Is that what you are suggesting?

sriganeshres:
@certik @gxyd I know this is not the final chages. But this is what i understood from propogating the error without exceptions. If this is correct approach i will try to propgate it to ast_common_vistor file. If not let me know.

src/libasr/modfile.cpp
```diff
@@ -135,7 +136,7 @@ std::string save_pycfile(const ASR::TranslationUnit_t &m, LCompilers::LocationMa
 }
 
 inline void load_serialised_asr(const std::string &s, std::string& asr_binary,
-                                LCompilers::LocationManager &lm) {
+                                LCompilers::LocationManager &lm, std::string& error_message, int& error_code) {
```
certik: Return `error_code` as a return value bool true/false.

src/libasr/modfile.cpp
```diff
@@ -147,7 +148,10 @@ inline void load_serialised_asr(const std::string &s, std::string& asr_binary,
     }
     std::string version = b.read_string();
     if (version != LFORTRAN_VERSION) {
-        throw LCompilersException("Incompatible format: LFortran Modfile was generated using version '" + version + "', but current LFortran version is '" + LFORTRAN_VERSION + "'");
+        error_message = "Incompatible format: LFortran Modfile was generated using version '"
+                        + version + "', but current LFortran version is '" + LFORTRAN_VERSION + "'";
+        error_code = 1;
+        return;  // Error code for incompatible version
```
certik:
```suggestion
        return false;
```

src/libasr/modfile.cpp
```diff
@@ -216,7 +220,13 @@ inline void load_serialised_asr(const std::string &s, std::string& asr_binary,
...
         bool load_symtab_id, SymbolTable &symtab, LCompilers::LocationManager &lm) {
     std::string asr_binary;
-    load_serialised_asr(s, asr_binary, lm);
+    std::string error_message;
+    int error_code = 0;
+    load_serialised_asr(s, asr_binary, lm, error_message, error_code);
+    if (error_code != 0) {
+        std::cerr << "Error loading modfile: " << error_message << std::endl;
+        std::abort();
+    }
```
certik: Change the return type from `ASR::TranslationUnit_t*` to `Result<ASR::TranslationUnit_t*>` and:
```suggestion
    int error_code = 0;
    if (!load_serialised_asr(s, asr_binary, lm, error_message)) {
        return Error(error_message);
    }
```
let the caller handle the error state.

src/libasr/modfile.cpp
```diff
@@ -227,7 +237,13 @@ ASR::TranslationUnit_t* load_modfile(Allocator &al, const std::string &s,
...
         bool load_symtab_id, LCompilers::LocationManager &lm) {
     std::string asr_binary;
-    load_serialised_asr(s, asr_binary, lm);
+    std::string error_message;
+    int error_code = 0;
+    load_serialised_asr(s, asr_binary, lm, error_message, error_code);
+    if (error_code != 0) {
+        std::cerr << "Error loading modfile: " << error_message << std::endl;
+        std::abort();
+    }
```
certik, line 240: Same here.

src/libasr/modfile.cpp
sriganeshres: removed it as we changed the return type of bool

src/libasr/modfile.cpp
sriganeshres: Done.

src/libasr/modfile.cpp
sriganeshres: Done.

src/libasr/modfile.cpp
sriganeshres: Done.

src/libasr/asr_utils.cpp
```diff
@@ -478,11 +479,16 @@ ASR::TranslationUnit_t* find_and_load_module(Allocator &al, const std::string &m
...
-                set_intrinsic(asr);
+            Result<ASR::TranslationUnit_t*> res = load_modfile(al, modfile, false, symtab, lm);
+            if (res.ok) {
+                ASR::TranslationUnit_t* asr = res.result;
+                if (intrinsic) {
+                    set_intrinsic(asr);
+                }
+                return asr;
+            } else {
+                std::cerr << res.error.message << std::endl;
```
certik: Let the caller handle the error:
```suggestion
                return res;
```

src/libasr/exception.h
```diff
@@ -49,6 +49,12 @@ namespace LCompilers {
     below.
 */
 struct Error {
+    std::string message;
+
+    Error() = default;
+    Error(const std::string& msg)
+        : message(msg)
+    {}
```
certik: Don't add it here, if it is only used at the place you used it, since the `message` is not used at the other cases. Maybe create a new one ErrorMessage.

sriganeshres: Since we have a default constuctor it doesnot matter.

certik: It's confusing to the user if the message will contain something or not, and if it needs to be set. I also don't know the performance implications right now. 

For now create a separate ErrorMessage. We can later merge it if needed.

src/libasr/asr_utils.cpp
sriganeshres: Done.

src/libasr/exception.h
sriganeshres: Done.

sriganeshres:
@certik I have added all the suggested changes. Please review it and merge this.

src/libasr/asr_utils.cpp
```diff
@@ -478,14 +480,19 @@ ASR::TranslationUnit_t* find_and_load_module(Allocator &al, const std::string &m
...
+                }
+                return asr;
+            } else {
+                return res;
             }
-            return asr;
         }
     }
-    return nullptr;
+    return Result<ASR::TranslationUnit_t*, ErrorMessage>(ErrorMessage("Module '" + msym + "' modfile was not found"));
```
certik: This should work:
```suggestion
    return ErrorMessage("Module '" + msym + "' modfile was not found");
```

src/libasr/asr_utils.cpp
```diff
@@ -1,5 +1,6 @@
 #include <unordered_set>
 #include <map>
+#include <iostream>
```
certik: This is not needed, right?

src/lfortran/tests/test_serialization.cpp
```diff
@@ -122,8 +122,16 @@ void asr_mod(const std::string &src) {
...
+    LCompilers::Result<LCompilers::ASR::TranslationUnit_t*, LCompilers::ErrorMessage> res
+        = LCompilers::load_modfile(al, modfile, true, symtab, lm);
+    LCompilers::ASR::TranslationUnit_t* asr2 = nullptr;
+    if (res.ok) {
+        asr2 = res.result;
+    } else {
+        std::cerr << "Error loading modfile: " << res.error.message << std::endl;
+        CHECK(false);
+        return;
+    }
```
certik: This seems easier:
```suggestion
    CHECK(res.ok);
    LCompilers::ASR::TranslationUnit_t* asr2 = res.result;
```

src/libasr/exception.h
```diff
@@ -51,47 +51,83 @@ namespace LCompilers {
...
-        Error error;
+        E error;
     };
     // Default constructor
     Result() = delete;
     // Success result constructor
-    Result(const T &result) : ok{true}, result{result} {}
+    Result(const T& res)
+        : ok{ true } {
+        new (&result) T(res);
```
certik: Don't use `new` here, that seems incorrect. Slower than it was before. Why is `new` needed here? I would use the previous code.

src/libasr/exception.h
```diff
@@ -51,47 +51,83 @@ namespace LCompilers {
...
+    }
     // Error result constructor
-    Result(const Error &error) : ok{false}, error{error} {}
+    Result(const E& err)
+        : ok{ false } {
+        new (&error) E(err);
```
certik: Don't use `new`.

src/libasr/exception.h
```diff
@@ -51,47 +51,83 @@ namespace LCompilers {
...
+    }
     // Move assignment
-    Result<T>&& operator=(T &&other) = delete;
+    Result& operator=(Result&& other) {
+        if (this != &other) {
+            this->~Result();
+            new (this) Result(std::move(other));  // placement new move
+        }
+        return *this;
+    }
```
certik, line 95: I don't like all these changes here. Are they needed? I would use the original code, but you can template E, that's fine.

src/libasr/modfile.cpp
```diff
@@ -6,6 +6,7 @@
 #include <libasr/modfile.h>
 #include <libasr/serialization.h>
 #include <libasr/bwriter.h>
+#include <iostream>
```
certik: Not needed?

src/libasr/exception.h
sriganeshres: New is needed as we need to use this for string in ErrorMessage.

src/libasr/exception.h
sriganeshres: This is needed for ErrorMessage struct to work with Result.

src/libasr/modfile.cpp
sriganeshres: Yes we can remove this. It was added for cerr

src/lfortran/tests/test_serialization.cpp
sriganeshres: Done.

src/libasr/modfile.cpp
sriganeshres: Done.

src/libasr/asr_utils.cpp
sriganeshres: We can remove that but it will safe gaurd rest cerr in this file are compiled in all conditions.

sriganeshres: Done.

src/libasr/exception.h
sriganeshres: Without this new Template will not compile.

src/libasr/asr_utils.cpp
sriganeshres: Done.

src/libasr/exception.h
sriganeshres: Done.

src/libasr/exception.h
sriganeshres: Done.

src/libasr/exception.h
sriganeshres: It is neccesary to accomodate Errormessage struct. I am leaving this suggestion out.

src/libasr/exception.h
```diff
@@ -67,31 +75,53 @@ struct Result {
...
+    // Move constructor for success
+    Result(T&& res)
+        : ok(true) {
+        new (&result) T(std::move(res));
+    }
+    // Move constructor for error (optional)
+    Result(E&& err)
+        : ok(false) {
+        new (&error) E(std::move(err));
+    }
```
certik: Cannot these be replaced with:
```c++
Result(T&& res) : ok(true), result{std::move(res)} {}
Result(E&& err) : ok(false), error{std::move(err)} {}
```

src/libasr/exception.h
certik: What code breaks with the original `Result<T>&& operator=(T &&other) = delete;` ?

src/libasr/exception.h
sriganeshres: Yes They can be replaced with those. Before i was reassigning which made it fail. Now with the current changes this will work.

src/libasr/exception.h
sriganeshres: I was re assigning in my way of implementation which made it neccesary. now we can use the old way.

src/lfortran/tests/test_serialization.cpp
sriganeshres, line 1: Here before your suggestion i was reassigning due to which i had modify move assignemnet and move constructors.

src/libasr/exception.h
```diff
@@ -67,31 +75,41 @@ struct Result {
...
+        : ok(other.ok) {
+        if (ok) {
+            new (&result) T(std::move(other.result));
+        } else {
+            new (&error) E(std::move(other.error));
+        }
+    }
     // Move assignment
-    Result<T>&& operator=(T &&other) = delete;
+    Result<T>&& operator=(T&& other) = delete;
```
certik: The changes above are not good, it allows Result to be assigned to, but you do not want to do that. You should only return it directly from a function and handle it in the caller.

certik:
I fixed it up myself, however it is not catching the error mode correctly. It is reporting the error as:
```
semantic error: Module 'iso_fortran_env' not declared in the current source and the modfile was not found
 --> /Users/ondrej/repos/lfortran/integration_tests/use_02.f90:2:5
  |
2 |     use iso_fortran_env, only: dp => real64
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
But that is an incorrect error.

certik:
Ok, previously it wasn't actually fixing the bug. Now it is giving a decent error message. In some error modes the error message has changed, and I think the error handling logic should be further reworked and simplified. But I think it might be good enough to merge now.

Issue #7298, opened 2025-05-14
Failed PARAMETER statement using %re or %im


JCollinsSimCon:
The following code fails to compile under lfortran:
```
   COMPLEX(8),PARAMETER :: x = (1.0D0,2.0D0)
   REAL(8),PARAMETER :: re = x%re
   REAL(8),PARAMETER :: ri = x%im
```
It doesn't compile under ifx either.  Works for gfortran.

kargl:
Note, gfortran's behavior is the expected conforming behavior.  In simplest terms, `x` is a named constant with a type of `complex`.  `x%re` and `x%im` are subobjects of the named constant.  One now goes to F2023, 10.1.12. and reads.

PR #7337, opened 2025-05-17
fix: set compile time value for ComplexIm

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7298


integration_tests/complex_20.f90
```diff
@@ -0,0 +1,9 @@
+program complex_20
+    complex(8), parameter :: x = (1.0D0, 2.0D0)
```
Pranavchiku, line 2:
```suggestion
    complex(8), parameter :: x = (1.129D0, -2.12D0)
```

integration_tests/complex_20.f90
```diff
@@ -0,0 +1,9 @@
...
+
+    real(8), parameter :: re = x%re
+    real(8), parameter :: ri = x%im
+
+    if (abs(re - 1) > 1e-5) error stop
+    if (abs(ri - 2) > 1e-5) error stop
+end program
```
Pranavchiku, line 9:
```suggestion
    if (abs(re - 1.129D0) > 1e-8) error stop
    if (abs(ri - (-2.12D0)) > 1e-8) error stop
end program
```

Pranavchiku (approved):
This is good, let's improve test in subsequent PR.

PR #7338, opened 2025-05-17
tests: better numbers and checking in complex_20

xaerru:
Address comments from https://github.com/lfortran/lfortran/pull/7337.

### DIFF
--- a/integration_tests/complex_20.f90
+++ b/integration_tests/complex_20.f90
@@ -1,9 +1,9 @@
 program complex_20
-    complex(8), parameter :: x = (1.0D0, 2.0D0)
+    complex(8), parameter :: x = (1.129D0, -2.12D0)
 
     real(8), parameter :: re = x%re
     real(8), parameter :: ri = x%im
 
-    if (abs(re - 1) > 1e-5) error stop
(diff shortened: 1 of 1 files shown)


integration_tests/complex_20.f90
```diff
@@ -1,9 +1,9 @@
 program complex_20
-    complex(8), parameter :: x = (1.0D0, 2.0D0)
+    complex(8), parameter :: x = (1.129D0, -2.12D0)
 
     real(8), parameter :: re = x%re
     real(8), parameter :: ri = x%im
 
-    if (abs(re - 1) > 1e-5) error stop
-    if (abs(ri - 2) > 1e-5) error stop
+    if (abs(re - 1.129D0) > 1e-8) error stop
```
Pranavchiku, line 7:
```suggestion
    if (abs(re - 1.129D0) > 1e-15) error stop
```

I'm sorry I missed that these were double precision numbers. For double precision we try to reach `1e-15` if not then `1e-8` is fine.

Pranavchiku (approved):
Please do a small change, I apologise I oversaw these were double precision numbers! Rest all is good.

Issue #7328, opened 2025-05-16
3rd-party code separate compilation


certik:
These codes at our CI must be fixed to work with separate compilation:

* [ ] PRIMA https://github.com/lfortran/lfortran/issues/7692
* [x] stdlib ( in sync with default mode )
* [x] fastGPT https://github.com/lfortran/lfortran/pull/7339
* [x] fortran_primes https://github.com/lfortran/lfortran/pull/7381

The following 3rd party codes are already tested with separate compilation:

* POT3D
* fortran_mpi
* dftatom
* scipy
* minpack (both variants)
* numerical_methods_fortran
* SNAP

PR #7339, opened 2025-05-17
ci: setup fastGPT with separate compilation

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/7328

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -432,6 +432,17 @@ time_section "🧪 Testing fastGPT" '
         ./test_basic_input
         ./test_more_inputs
         cd ..
+
+        mkdir lf-goc
+        cd lf-goc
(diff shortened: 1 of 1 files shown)


ci/test_third_party_codes.sh
```diff
@@ -432,6 +432,17 @@ time_section "🧪 Testing fastGPT" '
         ./test_basic_input
         ./test_more_inputs
         cd ..
+
+        mkdir lf-goc
+        cd lf-goc
+        FC="$FC --generate-object-code --rtlib" CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS -DCMAKE_BUILD_TYPE=Debug ..
```
Pranavchiku, line 438: This is sort of workaround, actual fix requires to get `integration_tests/file_23` compiled. Once that is done, we can remove this `--rtlib` flag.

Pranavchiku: With https://github.com/lfortran/lfortran/pull/7341 we might be able to remove it.

PR #7342, opened 2025-05-17
Ref: removed `ParserError` completely from `parser.cpp` file and add `ParserAbort` class

parth121101:


Pranavchiku (approved):
Makes sense to me, thank you!

Issue #7318, opened 2025-05-15
Handle `subroutine call` for nested `StructInstanceMembers`


jinangshah21:
MRE:
```fortran
module toml_mod
    type :: toml_value
    contains 
        procedure :: destroy
    end type
    type :: toml_node
      class(toml_value), allocatable :: val
    end type toml_node
contains
    subroutine destroy(self)
        class(toml_value), intent(inout) :: self
    end subroutine
end module

program test_type
    use toml_mod
    type(toml_node) :: tmp
    call tmp%val%destroy
end program
```

LFortran Error:
```console
base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2688, in main
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2616, in main_app(int, char**)
    if (endswith(arg_file, ".f90") || endswith(arg_file, ".f") ||
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 1166, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    }
  File "/home/jinang_shah/Desktop/lfortran/src/lfortran/fortran_evaluator.cpp", line 393, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, int*)
    compiler_options, run_fn, "", infile);
[...]

PR #7343, opened 2025-05-17
Fix: Handle polymorphic classType StructInstanceMember as a `self` arg SubroutineCall

adit4443ya:
Towards #7318


integration_tests/class_19.f90
```diff
@@ -0,0 +1,19 @@
+module toml_mod_19
```
Pranavchiku, line 1:
```suggestion
module toml_mod_19_class_19
```

Pranavchiku (approved):
looks good, just fix the integration test and merge it.

integration_tests/class_19.f90
adit4443ya: I forced pushed the branch with this change, Now we can merge this?

Issue #6467, opened 2025-03-03
'PF' format descriptor bug


gxyd:
## Minimal Reproducible Example

```f90
program scale_f_example
    implicit none
    real :: y
    y = 123.456

    ! Default scaling (0P)
    print '(F10.4)', y

    ! 1P: Multiply by 10^1
    print '(1PF10.4)', y

    ! -2P: Divide by 10^2
    print '(-2PF10.4)', y
end program scale_f_example
```

For the above Fortran program, the output produced by LFortran is:
```console
> lfortran --version
LFortran version: 0.42.0-1672-gce53ab904-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
  123.4560
  123.4560
  123.4560

> gfortran test.f90 -o test.o && ./test.o
  123.4560
 1234.5600
    1.2346
```

Issue #6308, opened 2025-02-14
Bad parameter values and bad E11.3 format


harperjf:
This program runs correctly with gfortran, gives an ICE with ifx, and lfortran gives wrong run-time output.
```
program test ! Modified from an Intel bug report by @Baschti 08-12-2024
  real, parameter :: a(4) = [ 1.0, 2.0, 3.0, 4.0 ]
  real, parameter :: b(4) = a / SUM(a)
  print "(4E11.3)", b
end program test
```
Lfortran 0.45.0 --std=f23 output: b(1) and b(4) wrong format, and wrong value of every element of b
```
  .0000E+00 -0.206E+07  0.415E-41  .0000E+00
```
Gfortran correct output (the leading zero is optional by F2023 13.7.2.3.3 paragraph 4):
```
  0.100E+00  0.200E+00  0.300E+00  0.400E+00
```

HarshilShah1804:
With the latest main, this program works well.
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
  0.100E+00  0.200E+00  0.300E+00  0.400E+00
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 --std=f23
  0.100E+00  0.200E+00  0.300E+00  0.400E+00
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
  0.100E+00  0.200E+00  0.300E+00  0.400E+00
```

HarshitaKalani:
Yes, that's right @HarshilShah1804 . Thanks!

PR #7344, opened 2025-05-18
fix: support scale factor in formats with float handling

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6467
closes: https://github.com/lfortran/lfortran/issues/6308


Pranavchiku (approved):
thank you!

PR #7364, opened 2025-05-19
ref: remove `parserError` class from file `parser_exception.h`

parth121101:


Pranavchiku (commented):
wow, thank you for this!

PR #7384, opened 2025-05-22
CI: test POT3D compilation with --std=f23 flag

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/6700

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -84,8 +84,14 @@ time_section "🧪 Testing POT3D with fortran_mpi" '
   print_subsection "Building with default flags"
   FC="$FC --cpp -DOPEN_MPI=yes" ./build_and_run.sh
 
+  print_subsection "Building with default flags and f23 standard"
+  FC="$FC --cpp -DOPEN_MPI=yes --std=f23" ./build_and_run.sh
+
   print_subsection "Building with optimization flags"
   FC="$FC --cpp --fast --skip-pass=dead_code_removal -DOPEN_MPI=yes" ./build_and_run.sh
+
+  print_subsection "Building with optimization flags and f23 standard"
+  FC="$FC --cpp --fast --skip-pass=dead_code_removal -DOPEN_MPI=yes --std=f23" ./build_and_run.sh
 
   print_subsection "Building POT3D in separate compilation mode"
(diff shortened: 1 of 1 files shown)


ci/test_third_party_codes.sh
```diff
@@ -84,9 +84,15 @@ time_section "🧪 Testing POT3D with fortran_mpi" '
...
+  FC="$FC --cpp -DOPEN_MPI=yes --std=f23" ./build_and_run.sh
+
   print_subsection "Building with optimization flags"
   FC="$FC --cpp --fast --skip-pass=dead_code_removal -DOPEN_MPI=yes" ./build_and_run.sh
 
+  print_subsection "Building with optimization flags and f23 standard"
+  FC="$FC --cpp --fast --skip-pass=dead_code_removal -DOPEN_MPI=yes --std=f23" ./build_and_run.sh
+
   print_subsection "Building POT3D in separate compilation mode"
   FC="$FC --cpp --generate-object-code -DOPEN_MPI=yes" ./build_and_run.sh
```
gxyd, line 97: Maybe we should also test "separate compilation mode" with `--std=f23` as well.

gxyd:
> One issue is that POT3D is really slow to run (many minutes), so this greatly increases our CI runs.

That's correct.



> I think we need to modify the test case in POT3D to run quickly (in few seconds, not minutes). So we need about 60x speedup in the tests. Is there a way to modify the test input files to run quicker? I created an issue for this at #7386.

Let me think about it and come back to you on this.

gxyd:
> To move forward, how much slower does this PR make the CI?

Quite slow, earlier I had posted the *duration* of different third party tests at LFortran's CI here: https://gist.github.com/gxyd/666361ad6e1161c42e8257d7a0656516, earlier *POT3D* would take (as mentioned in the gist) ~680 seconds and not it would take 1080 seconds, which isn't good.

gxyd:
> So this PR adds 400s = 6min 40s to every run of the 3rd party tests, which I think is every LLVM 10 - 20, for example?

Yes, that's right.

gxyd:
> I don't think we can merge it then.
> 
> I think we really need to fix #7386 first.

Agreed, I'll move this PR to draft mode for now, and send PR's first to simplify the validation program which takes lesser time with GFortran and LFortran both at fortran_mpi repository by testing it there first.

PR #7404, opened 2025-05-23
fix: handle `move_alloc` for unallocated variables

### LINKED ISSUE #7400: bug: segfault in `move_alloc` if target is not allocated
MRE
```fortran
program move_alloc
  implicit none
  integer, allocatable :: a(:), b(:)

  allocate(a(3))
  a = [1, 2, 3]

  call move_alloc(a, b)
  print *, b

end program move_alloc

```

LFortran
```console
```

GFortran
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
           1           2           3
```

If we allocate `b` before calling `move_alloc`, then it does not segfault.

HarshilShah1804:
Fixes #7400 

I am not sure if this is the best way to do it but we handle the `move_alloc` logic similarly by the function `instantiate_moveAlloc`. 

We can add semantic checks in the function `handle_moveAlloc` itself to prevent any invalid code from executing, and throw errors for wrong arguments like we do in `create_MoveAlloc` function.

### DIFF
--- a/integration_tests/intrinsics_378.f90
+++ b/integration_tests/intrinsics_378.f90
@@ -1,6 +1,6 @@
 program intrinsics_378
   implicit none
-  integer, allocatable :: a(:), b(:)
+  integer, allocatable :: a(:), b(:), c(:)
 
   allocate(a(3), b(3))
   a = [1, 2, 3]
@@ -8,5 +8,19 @@ program intrinsics_378
   call move_alloc(a, b)
 
   print *, allocated(a)
+  print *, b
+
+  if (.not. allocated(b)) error stop
+  if (allocated(c)) error stop
+  if (any(b /= [1, 2, 3])) error stop
+    
+  call move_alloc(b, c)
+
+  print *, allocated(b)
+  print *, c
+
+  if (allocated(b)) error stop
+  if (.not. allocated(c)) error stop
+  if (any(c /= [1, 2, 3])) error stop
 
 end program intrinsics_378
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3625,6 +3625,30 @@ public:
                     ASR::asr_t* func_call = create_func(al, x.base.base.loc, args, diag);
                     Vec<ASR::expr_t*> explicit_deallocate_args; explicit_deallocate_args.reserve(al, 1);
                     explicit_deallocate_args.push_back(al, args[0]);
+                    if (ASRUtils::is_array(ASRUtils::expr_type(args[0]))) {
+                        int n_dims = ASRUtils::extract_n_dims_from_ttype(ASRUtils::expr_type(args[0]));
+                        Vec<ASR::dimension_t> alloc_dims; alloc_dims.reserve(al, n_dims);
+                        ASR::ttype_t* integer_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4));
+                        for(int i=0; i<n_dims; i++) {
+                            ASR::dimension_t dim;
+                            dim.loc = x.base.base.loc;
+                            dim.m_start = ASRUtils::EXPR(ASR::make_IntegerConstant_t(
+                                al, x.base.base.loc, 1, integer_type));
+                            dim.m_length = ASRUtils::EXPR(ASR::make_ArraySize_t(
+                    al, x.base.base.loc, args[0], ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, x.base.base.loc, i+1, integer_type)), integer_type, nullptr));
+                            alloc_dims.push_back(al, dim);
+                        }
+                        Vec<ASR::alloc_arg_t> alloc_args; alloc_args.reserve(al, 1);
+                        ASR::alloc_arg_t alloc_arg;
+                        alloc_arg.loc = x.base.base.loc;
+                        alloc_arg.m_a = args[1];
+                        alloc_arg.m_dims = alloc_dims.p;
+                        alloc_arg.n_dims = alloc_dims.n;
+                        alloc_arg.m_len_expr = nullptr;
+                        alloc_arg.m_type = nullptr;
+                        alloc_args.push_back(al, alloc_arg);
+                        current_body->push_back(al, ASRUtils::STMT(ASR::make_Allocate_t(al, x.base.base.loc, alloc_args.p, alloc_args.n, nullptr, nullptr, nullptr)));
+                    }
                     ASR::stmt_t* explicit_deallocate = ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(al, x.base.base.loc, explicit_deallocate_args.p, explicit_deallocate_args.n));
                     tmp = ASRUtils::make_Assignment_t_util(al, x.base.base.loc, args[1], ASRUtils::EXPR(func_call), nullptr, compiler_options.po.realloc_lhs);
                     current_body->push_back(al, ASRUtils::STMT(tmp));


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3576,13 +3576,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 if( ASRUtils::IntrinsicElementalFunctionRegistry::is_intrinsic_function(var_name) ) {
                     fill_optional_kind_arg(var_name, args);
 
-                    ASRUtils::create_intrinsic_function create_func =
-                        ASRUtils::IntrinsicElementalFunctionRegistry::get_create_function(var_name);
```
Pranavchiku, line 3580: this means we are completely removing the intrinsic function call, why?

Pranavchiku (commented):
something is weird, this PR removes the call to intrinsic `move_alloc` but no test fails

src/lfortran/semantics/ast_body_visitor.cpp
HarshilShah1804: Ultimately, we achieve the same goal using the `instantiate_MoveAlloc` function like this:
`body.push_back(al, b.Assignment(result, args[0]));`

The primary motivation for this approach was to ensure automatic allocation of the variable if it is not already allocated—since, by default, assignment causes reallocation of the LHS, if LHS is allocatable. This addresses that issue.

While we could also perform manual allocation before using `instantiate_MoveAlloc`, runtime dimension information would not be available in that case. 

I chose this method because it effectively addresses both concerns. Additionally, it resolves issues related to derived types in `move_alloc`, which have been reported in stdlib and other packages like toml-f (see #7408, #7357). These issues arise from the assertion `derived_type_sym != nullptr`, which fails because the type symbol isn't available in the function scope.

Interestingly, such issues don’t occur with direct assignment. This is because `move_alloc` essentially just transfers the value from the from variable to the to variable. Therefore, the functionality of `move_alloc` can be conceptually simplified as:

```fortran
allocate(b)  ! to the shape of a (reallocation due to assignment handles this)
b = a
deallocate(a)
```

HarshilShah1804:
> something is weird, this PR removes the call to intrinsic `move_alloc` but no test fails

The reason for this is probably that we do assignment only by the function `instantiate_MoveAlloc` and moreover, we don't have any test which matches the ASR with any refrence.

integration_tests/intrinsics_378.f90
```diff
@@ -1,12 +1,17 @@
...
   a = [1, 2, 3]
 
   call move_alloc(a, b)
 
   print *, allocated(a)
+  print *, b
+
+  call move_alloc(b, c)
+  print *, allocated(b)
+  print *, c
```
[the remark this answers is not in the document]
HarshilShah1804: Updated the test.

src/lfortran/semantics/ast_body_visitor.cpp
HarshilShah1804: For the second point,
```cpp
ASR::asr_t* make_Assignment_t_util(Allocator &al, const Location &a_loc,
    ASR::expr_t* a_target, ASR::expr_t* a_value,
    ASR::stmt_t* a_overloaded, bool a_realloc_lhs) {
    bool is_allocatable = ASRUtils::is_allocatable(a_target);
    if ( ASR::is_a<ASR::StructInstanceMember_t>(*a_target) ) {
        ASR::StructInstanceMember_t* a_target_struct = ASR::down_cast<ASR::StructInstanceMember_t>(a_target);
        is_allocatable |= ASRUtils::is_allocatable(a_target_struct->m_v);
    }
    a_realloc_lhs = a_realloc_lhs && is_allocatable;
    return ASR::make_Assignment_t(al, a_loc, a_target, a_value,
        a_overloaded, a_realloc_lhs);
}
```

I think when we have allocatable as LHS, we have `realloc_lhs = true` by default.

HarshilShah1804: Regarding the performance, by using `dump-all-passes-fortran` with the latest main on the fortran code:
```fortran
program expr2
  integer, allocatable :: a(:), b(:)
  allocate(a(3), b(3))
  a = [1, 2, 3]
  call move_alloc(a, b)
end program
```
we do something like `b = move_alloc(a, b)` after `ast_body_visitors.cpp`, which is then replaced after intrinsic_functions pass to:

```fortran
do __libasr_index_0_2 = lbound(b, 1), ubound(b, 1)
    b(__libasr_index_0_2) = _lcompilers_move_alloc_i32(a(__libasr_index_0_3), b(__libasr_index_0_4))
    __libasr_index_0_3 = __libasr_index_0_3 + 1
    __libasr_index_0_4 = __libasr_index_0_4 + 1
end do

integer(4) function _lcompilers_move_alloc_i32(from, to)
    integer(4), intent(in) :: from
    integer(4), intent(in) :: to
    _lcompilers_move_alloc_i32 = from
end function _lcompilers_move_alloc_i32

```

So, it is like we are doing the assignment only, we call the function `_lcompilers_move_alloc_i32` and just return the `from` value only.

So, technically, `b(__libasr_index_0_2) = _lcompilers_move_alloc_i32(a(__libasr_index_0_3),` is just `b(__libasr_index_0_2) = a(__libasr_index_0_3)`, which is what would also been done by the assignment.

This is the code after pass from my branch:
```fortran
do __libasr_index_0_2 = lbound(b, 1), ubound(b, 1)
    b(__libasr_index_0_2) = a(__libasr_index_0_3)
    __libasr_index_0_3 = __libasr_index_0_3 + 1
end do
```

So, I don't think it should hamper the performance. I guess it should improve the performance as we are removing the redundant function calls. Please provide your suggestions on this.

HarshilShah1804: Got it. So, we have to explicitely allocate `b` to the size of `a` instead of doing direct assignment.

HarshilShah1804: @certik I have added allocation of target variable to same as the size of source.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3625,6 +3625,30 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                     explicit_deallocate_args.push_back(al, args[0]);
+                    if (ASRUtils::is_array(ASRUtils::expr_type(args[0]))) {
+                        int n_dims = ASRUtils::extract_n_dims_from_ttype(ASRUtils::expr_type(args[0]));
+                        Vec<ASR::dimension_t> alloc_dims; alloc_dims.reserve(al, n_dims);
+                        ASR::ttype_t* integer_type = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4));
+                        for(int i=0; i<n_dims; i++) {
+                            ASR::dimension_t dim;
+                            dim.loc = x.base.base.loc;
+                            dim.m_start = ASRUtils::EXPR(ASR::make_IntegerConstant_t(
+                                al, x.base.base.loc, 1, integer_type));
```
Pranavchiku, line 3636:
```suggestion
                            dim.m_start = ASRUtils::EXPR(ASR::make_ArrayBound_t(
                                xx));
```

Pranavchiku: I think it is fine as we are using `ArraySize` as `m_length`

Pranavchiku (changes requested):
It is fine to merge but this is not a correct fix, the argument by @HarshilShah1804 is valid ( why can't we have a direct assignment as `args[1] = args[0]` ). 

The correct fix is to have `move_alloc` as `IntrinsicSubroutine` and not as an `IntrinsicElementalFunction`. I'm fine to merge this but let's also do a proper fix in subsequent PR.

Pranavchiku (approved):
open an issue for the requested change, and you can follow in subsequent PR

PR #7405, opened 2025-05-23
impl: Added Implementation for Allocatble Cptr

sriganeshres:


tests/tests.toml
```diff
@@ -4634,3 +4634,7 @@ run = true
 [[test]]
 filename = "errors/integer_overflow.f90"
 run = true
+
+[[test]]
```
Pranavchiku, line 4638: you can remove this and add a test in `integration_tests/CMakeLists.txt`

tests/allocatble_c_ptr.f90
```diff
@@ -0,0 +1,7 @@
+program main
```
Pranavchiku, line 1: let's add it in integration test

Pranavchiku (dismissed):
changes look good, just transfer test to integration tests and it should be good to merge.

tests/tests.toml
sriganeshres: Can you tell me what Labels to be added.

tests/tests.toml
```diff
@@ -4633,4 +4633,4 @@ run = true
 
 [[test]]
 filename = "errors/integer_overflow.f90"
-run = true
+run = true
```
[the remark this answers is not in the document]
sriganeshres: I have added and removed the lines so i did not see the whitespaces. I will make sure for the next.

PR #7418, opened 2025-05-24
feat: add support for global `struct` initialization 

### LINKED ISSUE #7394: Incorrect default initialization of module-level derived type variables
Another reason why test for `stdlib_logger` failing. When defining variable of same type inside the module itself which is publicly accessible, it's value is not printed as expected. 

MRE:
```fortran
module temp_module
    public :: myType, global_myType

    type :: myType
        private 
        logical :: a = .true.
        integer :: i = 2

        contains
            private
            procedure, public, pass(self) :: config
    end type myType

    type(myType) :: global_myType

contains
    subroutine config(self)
        class(myType), intent(in) :: self
        print *, "a = ", self%a
        print *, "i = ", self%i        
    end subroutine config

end module temp_module

program main
    use temp_module, global => global_myType
    call global % config()
end program main
```

Output:
```console
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ lfortran test/logger/temp2.f90
a =     F
i =     0
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ gfortran test/logger/temp2.f90 && ./a.out
 a =  T
 i =            2
```

Here, LFortran prints the variable `a = False` and `i = 0` while it is defined as the `True` & `2` respectively.

kmr-srbh:
resolves #7394

### DIFF
--- a/tests/reference/llvm-class_01-82031c0.json
+++ b/tests/reference/llvm-class_01-82031c0.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-class_01-82031c0.stdout",
-    "stdout_hash": "0f26a06768d32c2d8676838083a67562c192531b2776961317f1c8d7",
+    "stdout_hash": "c601227acd7214209f7bba6a6278ed86783d5634bcdc33f39d58dd95",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-class_01-82031c0.stdout
+++ b/tests/reference/llvm-class_01-82031c0.stdout
@@ -2,7 +2,7 @@
 source_filename = "LFortran"
 
 %circle_polymorphic = type { i64, %circle* }
-%circle = type <{ float }>
+%circle = type { float }
 %__vtab_circle = type { i64 }
 
 @pi = global float 0x400921FB60000000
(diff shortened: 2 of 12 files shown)


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -244,7 +244,7 @@ namespace LCompilers {
                 name2memidx[der_type_name][std::string(member->m_name)] = member_idx;
                 member_idx++;
             }
-            (*der_type_llvm)->setBody(member_types, true);
+            (*der_type_llvm)->setBody(member_types);
```
assem2002, line 247: As you mentioned the commented out test is caused due to using **unpacked** structs, Does the PR really need to use unpacked structs?
Reverting this would make test `print_06` work .

assem2002 (approved):
The implementation is clean. Using global values instead of stack ones is good choice.
Thanks for fixing that @kmr-srbh

src/libasr/codegen/llvm_utils.cpp
kmr-srbh: Yes, I had faced memory alignment issues with packed structs when dealing with strings and string variables. I will look into it again. :)

Pranavchiku: Why do we need to do that? Does this PR not only intend to do `struct` initialization?

kmr-srbh:
Issue fixed in #8022.

PR #7444, opened 2025-05-26
CI: build and run dftatom third party repository with --std=f23

gxyd:
## Description

Once the CI checks pass on this PR, we should wait to see how much time it takes at LFortran's CI now, compared to what it used to take earlier.

In one of the earlier CI runs, we noticed that dftatom on current *main* takes ~75 seconds (see: https://gist.github.com/gxyd/666361ad6e1161c42e8257d7a0656516)

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -498,6 +498,22 @@ time_section "🧪 Testing dftatom" '
   git clean -dfx
   make -f Makefile.manual F90=$FC F90FLAGS="-I../../src --generate-object-code --fast"
   make -f Makefile.manual quicktest
+
+  git clean -dfx
+  make -f Makefile.manual F90=$FC F90FLAGS="-I../../src --std=f23"
+  make -f Makefile.manual quicktest
+
+  git clean -dfx
+  make -f Makefile.manual F90=$FC F90FLAGS="-I../../src --fast --std=f23"
+  make -f Makefile.manual quicktest
+
+  git clean -dfx
+  make -f Makefile.manual F90=$FC F90FLAGS="-I../../src --generate-object-code --std=f23"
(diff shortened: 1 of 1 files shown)


gxyd:
Even in release mode with just the use of `--std=f23`, the runtime speed is a little slow though.

I'm concerned about the increasing CI time because of these additional tests, do you think if we enable these added CI checks in this PR for only a specific version of LLVM and specific OS (as mentioned here: https://github.com/lfortran/lfortran/issues/6020#issue-2790936267)?

gxyd:
> So your total estimate is 190-75=115s. From your more granular numbers of compile time + runtime, I get less than 7+10+10+14+10+3+10+3=67s. So there is still about 48s discrepancy.

That initial timing I posted, which is it taking 190 seconds is at LFortran's CI, not on my machine, which was posted to give an estimate of how slower all of our CI jobs will run compared to what they currently do.

Pranavchiku:
I mean this bogs us very down, I've to wait for 3 hours almost to get a PR in and quick fixes or progress are just not possible. I'm fine to merge this but we need to figure out a way to club these. I'll say we need to atleast test separate compilation and default fully or maybe merge separate compilation with --std=f23 and test these two fully. This approach although robust but makes PR merging cycle massive.

gxyd:
In this PR we run new combination of third party tests with `--std=f23` for *all possible LLVM versions 10,, 11, .., 20` and for both macOS and ubuntu.

Maybe we do this for just one specific LLVM version and on one platform? (As was suggested here: https://github.com/lfortran/lfortran/issues/6020#issue-2790936267)

gxyd:
> I think CI speed is a priority. Let's not merge this, but keep it open, until we figure out a way to do this.

Agreed.

Issue #7128, opened 2025-05-01
bug: Allocate giving wrong output with `source` argument


HarshilShah1804:
MRE:
```fortran
program expr2
    integer, allocatable :: a(:), b(:)
    a = [1, 2, 3, 4, 5]
    allocate(b(10), source = a)
    print *, b
end program

```

LFortran:
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
1    2    3    4    5
```

GFortran:
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
           1           2           3           4           5           0           0           0           0           0
```

PR #7448, opened 2025-05-26
`ReAlloc` only when the target is not allocated

czgdp1807:
Closes https://github.com/lfortran/lfortran/issues/7128

### DIFF
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -854,9 +854,9 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
 
     void insert_realloc_for_target(ASR::expr_t* target, ASR::expr_t* value, Vec<ASR::expr_t**>& vars) {
         ASR::ttype_t* target_type = ASRUtils::expr_type(target);
+        bool array_copy = ASR::is_a<ASR::Var_t>(*value) && ASR::is_a<ASR::Var_t>(*target);
         if( (realloc_lhs == false || !ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
-            !(ASR::is_a<ASR::Var_t>(*value) && ASR::is_a<ASR::Var_t>(*target) &&
-              ASRUtils::is_allocatable(target_type)) ) {
+            !(array_copy && ASRUtils::is_allocatable(target_type)) ) {
             return ;
         }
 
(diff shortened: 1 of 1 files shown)


HarshilShah1804:
@czgdp1807 adding a `if (.not. allocated())` for deciding reallocation may fail some cases like this.
```fortran
program test
    integer, allocatable :: b(:)
    allocate(b(2))
    b = [1, 2, 3, 4, 5]
    print *, b
end program
```

LFortran
```console
1    2
```

GFortran
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran -fcheck=all -Wall a.f90 && ./a.out
           1           2           3           4           5
```

Moreover, the issue that this PR is targeting has wrong MRE which fails with GFortran when `-fcheck=all` is enabled. Refer https://github.com/lfortran/lfortran/pull/7129#issuecomment-2864535055

czgdp1807:
In my new approach I am not using `is_allocated`. I think directly putting assignment statement in AST to ASR. It modifies the semantics of the input code. Like if you convert the output of AST to ASR back to Fortran you will get an additional `b = a` statement which was not there in the Fortran code.

I have handled the `source` think in `array_op` ASR pass and kept the AST to ASR as it is.

src/libasr/pass/array_op.cpp
```diff
@@ -896,7 +896,51 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+                Vec<ASR::expr_t**> vars1;
+                vars1.reserve(al, 1);
+                if( ASRUtils::is_array(
+                        ASRUtils::expr_type(assignment_t->m_value)) ) {
+                    vars1.push_back(al, &(assignment_t->m_value));
+                }
+                vars1.push_back(al, &(assignment_t->m_target));
+
+                // TODO: Remove the following. Instead directly handle
+                // allocate with source in the backend.
```
czgdp1807, line 929: If `allocate` has `source` then we should be using `memcpy` instead of putting `source` into `assignment` statement. Would be faster (`memcpy` is faster than looping I think). @certik What do you say?

czgdp1807: Also note - https://github.com/lfortran/lfortran/pull/7448#issuecomment-2910527834

Issue #7366, opened 2025-05-19
[OPENMP] SECTIONS Construct


adit4443ya:
Reference :- [Examples Document](https://www.openmp.org/wp-content/uploads/openmp-examples-6.0.pdf)
Page :-40

Other links: [Reference Guide](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [Specifications](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [More Examples](https://passlab.github.io/Examples/contents/Title_Page.html).

Use Cases:-
- The sections construct forms a region that contains one or more structured blocks. Each block of a sections directive is constructed with a section construct, and executed once by one of the threads (any one) in the team.
- Parallelizes independent tasks with different computational requirements

### 1
```fortran
!$OMP PARALLEL SHARED(array1, array2)
  !$OMP SECTIONS
  !$OMP SECTION
  do i = 1, n
    array1(i) = array1(i) * 2.0
  end do
  !$OMP SECTION
  do i = 1, n
    array2(i) = array2(i) * 3.0
  end do
  !$OMP END SECTIONS
  !$OMP END PARALLEL
```

Description:- Here each section block is exec uted by single thread but all sections in the PARALLEL region are executed simultaneously (concurrently). Hence kind of  Independent Task Parallelisim

Issue #7332, opened 2025-05-17
Design OPENMP ParallelRegion ASR Ndoe


adit4443ya:
# Proposal: Introduce `OMPRegion` ASR Node for Enhanced OpenMP Support

This issue proposes a new ASR node, `OMPRegion`, to improve OpenMP support in LFortran. It aims to handle a wider range of OpenMP 6.0 constructs and clauses in a clear and flexible way.

## Proposed ASR Node Design
The `OMPRegion` node is defined as:

```
stmt
  = ...
  | OMPRegion(omp_region_type region, omp_clause* clauses, stmt* body)

omp_region_type
  = Parallel | Do | ParallelDo | Sections | Single | Task | Simd | Teams | Target | TargetData

omp_clause
  = OMPPrivate(expr* vars)
  | OMPShared(expr* vars)
  | OMPFirstPrivate(expr* vars)
  | OMPLastPrivate(expr* vars)
  | OMPReduction(reduction_op operator, expr* vars)
  | OMPCollapse(expr count)
  | OMPIf(expr condition)
  | OMPNumThreads(expr num_threads)
  | OMPSchedule(schedule_type kind, expr? chunk_size)
  | OMPNowait()
  | OMPMap(map_type type, expr* vars)
  | OMPDevice(expr device_id)
  | OMPThreadLimit(expr limit)
  | OMPEnterData(expr* vars)
  | OMPExitData(expr* vars)
  | OMPUpdateData(expr* vars)
  | OMPUseDevicePtr(expr* vars)

reduction_op
  = ReduceAdd | ReduceSub | ReduceMul | ReduceMIN | ReduceMAX

schedule_type
  = Static | Dynamic | Guided | Auto | Runtime

map_type
  = To | From | ToFrom | Alloc
```

This design keeps OpenMP constructs organized and easy to extend.

## Sample ASR
For this Fortran code:

```fortran
!$omp teams num_teams(4) thread_limit(8)
    !$omp parallel do collapse(2)
    do i = 1, 100
        do j = 1, 100
            a(i, j) = i + j
        end do
    end do
!$omp end teams
```

The ASR would be (Roughly):

```clojure
OMPRegion(
    region = Teams,
    clauses = [
        OMPNumTeams(IntegerConstant(4, type=Integer(4))),
        OMPThreadLimit(IntegerConstant(8, type=Integer(4)))
    ],
    body = [
        OMPRegion(
            region = ParallelDo,
[...]

PR #7449, opened 2025-05-26
[OPENMP] Feat: Introduce OMPRegion Node in ASR and Represent Sections Construct in ASR

adit4443ya:
Towards #7366 
Follows design of #7332 

I have picked Sections Construct to implement First due to these reasons 
- `Parallel Do` and `Parallel Sections` are different types of construct one is loop based and other is non-loop based.
- Current Design is more prone to accept the `Parallel Do` only and represent it directly as `DoConcurrentLoop` , Hence if i had gone to shift `parallel do` in currently proposed design it would take long PRs (As it would need to adapt openmp pass to visit OMPRegion too)
- With the selection of Sections construct we can go ahead and represent it in ASR without disrupting `Parallel Do` construct's test case which we compile now. (Hence no regression)
- Once this is merged we can easily extend the OPENMP pass to process OMPRegion ASR Node

I hope i clarified it completely


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -134,20 +136,55 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     tmp_vec.clear();
                 }
             }
-            if (tmp != nullptr) {
-                ASR::stmt_t* tmp_stmt = ASRUtils::STMT(tmp);
-                if (tmp_stmt->type == ASR::stmtType::SubroutineCall) {
-                    ASR::stmt_t* impl_decl = create_implicit_deallocate_subrout_call(tmp_stmt);
-                    if (impl_decl != nullptr) {
-                        body.push_back(al, impl_decl);
+            bool in_sections = !omp_regions.empty() && omp_regions.back()->m_region == ASR::omp_region_typeType::Sections;
```
adit4443ya, line 139: I will improvise this later on after i finish to represent Other constructs' too in ASR (e.g. teams,tasks etc).
This will need refactoring in future

Pranavchiku (commented):
I'll do a careful review in sometime.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -134,20 +136,55 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                current_body = &section_body;
+                i++; // Move to next statement
+                while (i < n_body && !(m_body[i]->type == AST::stmtType::Pragma &&
+                    AST::down_cast<AST::Pragma_t>(m_body[i])->m_type == AST::OMPPragma && 
+                    (to_lower(AST::down_cast<AST::Pragma_t>(m_body[i])->m_construct_name) == "section" ||
+                    AST::down_cast<AST::Pragma_t>(m_body[i])->m_end))) {
+                    this->visit_stmt(*m_body[i]);
+                    if (tmp) {
+                        section_body.push_back(al, ASRUtils::STMT(tmp));
+                        tmp = nullptr;
```
Pranavchiku, line 155: we don't need this all of these in `transform_stmts`, let's follow an approach similar to `omp_constructs` and a hashmap to temporarily store intermediate statements as discussed over call.

Pranavchiku (changes requested):
The reference ASR looks good to me, let's refactor the ASR creation approach and this PR shall be good.

src/libasr/ASR.asdl
```diff
@@ -246,4 +247,19 @@ arraybound = LBound | UBound
...
+  | OMPFirstPrivate(expr* vars)
+  | OMPLastPrivate(expr* vars)
+  | OMPReduction(reduction_op operator, expr* vars)
+  | OMPCollapse(expr count)
+  | OMPIf(expr condition)
+  | OMPNumThreads(expr num_threads)
+  | OMPSchedule(schedule_type kind, expr? chunk_size)
+  | OMPNowait()
+schedule_type
+  = Static | Dynamic | Guided | Auto | Runtime
```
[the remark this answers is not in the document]
adit4443ya: Thanks for the review!!
I will do this in subsequent PR, if not required to get this PR merged?

PR #7459, opened 2025-05-27
Added support for nullify with class variables

### LINKED ISSUE #7350: Handle `ClassType` variables inside `nullify`
MRE:
```fortran
program nullify_class
    type :: toml_table
        integer :: x
    end type
    class(toml_table), pointer :: temp
    nullify(temp)
end program
```

LFortran Error:
```console
  store %toml_table_polymorphic* null, %toml_table_polymorphic* %temp, align 8
 %toml_table_polymorphic = type { i64, %toml_table* }
```

swamishiju:
Fixes #7350

### DIFF
--- a/integration_tests/nullify_06.f90
+++ b/integration_tests/nullify_06.f90
@@ -2,6 +2,7 @@ program nullify_class
     type :: toml_table
         integer :: x
     end type
+    
     class(toml_table), pointer :: temp
     nullify(temp)
 end program


swamishiju:
IR now looks like

```
store %tt* null, %tt** %5, align 8
store %tt_polymorphic zeroinitializer, %tt_polymorphic* %temp, align 8
```

integration_tests/nullify_06.f90
```diff
@@ -0,0 +1,7 @@
+program nullify_class
+    type :: toml_table
+        integer :: x
+    end type
+    class(toml_table), pointer :: temp
+    nullify(temp)
```
gxyd, line 6: Maybe we need something like `if (associated(temp)) error stop` condition to ensure that this actually work? i.e. it really nullifies it.

integration_tests/nullify_06.f90
```diff
@@ -0,0 +1,7 @@
+program nullify_class
```
gxyd, line 1: name the program as `nullify_06` itself?

gxyd (commented):
Overall this looks good to me.

integration_tests/nullify_06.f90
swamishiju: @gxyd it appears associated doesn't work for class pointers, I checked-out main and ran

```f90
program nullify_06
    type :: toml_table
        integer :: x
    end type
    class(toml_table), pointer :: temp
    if (associated(temp)) error stop
end program
```

without nullify it gave codegen error

```
PtrToInt source must be pointer
  %7 = ptrtoint %toml_table_polymorphic %6 to i64
PtrToInt source must be pointer
  %9 = ptrtoint %toml_table_polymorphic %6 to i64
```

gxyd (approved):

gxyd:
Seems like PRIMA is failing here.

integration_tests/nullify_06.f90
swamishiju: #7461

gxyd: > PRIMA fails right?

Yes, seems like it. Maybe we can wait and see if it fails for other LLVM versions or not.

swamishiju: This is very surprising it only fails on 11, why 11?

swamishiju: @gxyd it seems to be flaky test..

gxyd:
> If this is some random failure, please open up an issue for this to track it. It will happen again and we'll have to fix it. Is there a possibility this random failure is caused by this PR?

This is my concern as well, it's possible that this PR causes it, but we aren't sure about it. Maybe once we open an issue we can link this PR that this failure was noticed for the first time here, so we know it's first place of observation (unless there is already an issue open for this)

PR #7462, opened 2025-05-27
Set intrinsics

swamishiju:
Implemented set intrinsics

- Constant
- Len
- Add

excuse the commits i had some nightmare rebasing


swamishiju:
Another libasr bug 👍

swamishiju:
This bug has been a headache I still haven't thought of a fix, I put a message on zulip

swamishiju:
> In this screenshot:
> image
> 
> I think the `NOLLVM17` should be yes, not no.

I think it should be no only, I didnt add the flag to wasm tests the wasm llvm 19 was were it failed I have fix for that I sent some code on zulip, I wanted to confirm before pushing I don't want to bog up the CI

swamishiju:
@certik I've fixed the libasr issues, was working on that today morning it is somewhat big

swamishiju:
A lot of the issues where from being unable to get `list_type` from `ptr` so I've created a couple new functions to handle that given the contained type in some form

PR #7463, opened 2025-05-27
enh: add support for passing `class` arguments  to functions expecting `type`

### LINKED ISSUE #7423: Handle `ClassType` variable passed to subroutine which expects a `StructType`
MRE:
```fortran
module toml_val_mod
  implicit none
  type :: val_type
        integer :: origin = 3
        integer :: dest = 2
    end type

end module toml_val_mod

program merge_example
    use toml_val_mod
    class(val_type), allocatable :: val1
    call merge(val1)
contains

    subroutine merge(val1)
        type(val_type), intent(inout) :: val1
    end subroutine merge

end program merge_example
```

LFortran Error:
```console
  %val1 = alloca %val_type_polymorphic, align 8
 %val_type*  call void @merge(%val_type_polymorphic* %val1)
```

kmr-srbh:
fixes #7423

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1716,6 +1716,8 @@ RUN(NAME class_27 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_29 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_30 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


integration_tests/class_29.f90
```diff
@@ -0,0 +1,22 @@
...
+   allocate(val1)
+
+   call sub(val1)
+   print *, "Outside val1%origin: ", val1%origin
+   if (val1%origin /= 5) error stop
+contains
+
+   subroutine sub(val)
+      type(val_type), allocatable, intent(out) :: val
+      allocate(val)
```
kmr-srbh, line 18: I had to change the test here to not check for `val%origin == 3` because the allocate statement initialises the values to zero. The added non-allocatable test `class_28.f90` work fine with this check though.

kmr-srbh: #7530 is the bug report for this.

Issue #7386, opened 2025-05-22
CI: speedup POT3D tests


certik:
Currently the POT3D tests runs take minutes. Instead, they should take seconds, so we need about 60x speedup of the runtime. To do that, we need to device a testcase that is smaller.

That way we can focus on compiling POT3D in many different ways via LFortran, and spend CI time on that, and not waste it actually running POT3D. We only need to run POT3D to check that it was compiled correctly, and we can do that with a smaller test case.

PR #7467, opened 2025-05-28
CI: use validation dataset which runs faster

gxyd:


ci/test_third_party_codes.sh
```diff
@@ -71,7 +71,7 @@ time_section "🧪 Testing POT3D with fortran_mpi" '
   git clone https://github.com/gxyd/pot3d.git
   cd pot3d
   git checkout -t origin/lf_hdf5_fortranMPI_namelist_global_workarounds
-  git checkout 695841f041300ee75fdfdd1da5d0fc6fe66f5717
+  git checkout 9bf5d4784581ce83e2df13b828de86950ba88902
```
gxyd, line 74: Changes introduced by this commit can be seen at: https://github.com/gxyd/POT3D/commit/9bf5d4784581ce83e2df13b828de86950ba88902

gxyd:
As seen here: [one of the CI job](https://productionresultssa14.blob.core.windows.net/actions-results/0e5e9bc1-bb67-42d4-8fe1-7c390d423d1c/workflow-job-run-60549d08-c86f-5cfb-bddc-6be86f6bc48f/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-05-28T04%3A45%3A26Z&sig=dshtTeN8nmQEVHV2ebzb3Q%2BT0MddJohTagU2ZnlzU8g%3D&ske=2025-05-28T15%3A16%3A06Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-05-28T03%3A16%3A06Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-05-28T04%3A35%3A21Z&sv=2025-05-05), it now takes ~112 seconds to run, compared to ~356 seconds to 400 seconds earlier.

gxyd:
@certik , do you think if we can consider the linked issue (i.e. #7386 ) fixed? Or would you expect any further improvements in timings?

gxyd:
> So it now takes 30s to run with 1 MPI rank, unoptimized build. Ideally we should run in 1s. To achieve that, we need to reduce all array sizes to absolute minimum. It could be that the POT3D code might not easily allow that, I don't know.

I did try reducing the sizes of `np` and `nt` further, the parameters used in POT3D program for *size*, and it actually worked as well with *mpif90* (i.e. without our custom MPI wrappers and uses HDF5), but it failed with GFortran + our custom MPI wrappers + binary read (instead of HDF5), so I do believe it's possible to further reduce it, I'll try that as well.

Issue #4263, opened 2024-06-14
Refactor StructType


certik:
StructType should contain the types of the members (but not names) and not reference the symbol table.

See:

https://github.com/lfortran/lfortran/issues/1162#issuecomment-1374249861

certik:
The ASR of the `StructType` ttype should be changed as follow:
```diff
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -193,7 +193,7 @@ ttype
     | Set(ttype type)
     | List(ttype type)
     | Tuple(ttype* type)
-    | StructType(symbol derived_type)
+    | StructType(ttype* member_types, bool packed)
     | EnumType(symbol enum_type)
     | UnionType(symbol union_type)
     | ClassType(symbol class_type)
```
The `packed` boolean can be added later, I just added it here as an example.

xaerru:
> StructType ttype has to be updated to have a list of types, and no reference to the symbol table.

We can do this if we don't want the symbol in `StructType`:

1. We store the name of the struct as a string in `StructType`, if we don't, we lose nominal typing. Two types with same type members and different names would be considered as the same type.
2. The `Struct` symbol contains the member variables, class procedures and parent type symbol. We need to know what `Struct` symbol the `StructType` variable belongs to. This can be done by resolving the name of the type in the variable's symbol table, the struct symbol would be imported as an `ExternalSymbol`.

Regarding directly putting all the members as a `ttype` we lose information which is stored in the `Variable` symbol like the default value. I don't think this can be handled with just `ttype`. If we can find a solution to this then we can consider this design.

We need to understand the tradeoffs between having the symbol directly in the `StructType` and replacing the symbol with a string name.

My opinion is that the current design is good. We can remove the `ttype` list and only have the struct symbol. It represents correctly and minimally the information that is needed for the `StructType` variable to be translated. I agree with improving on the `Struct` symbol though.

Also in my opinion `StructType` and `ClassType` should be different ASR nodes because they represent different types. Having only a `StructType` and a field to indicate whether it is a class makes it confusing.

For reference, the current ASR design:
<details>
<summary>Fortran</summary>

```fortran
module test_module
    implicit none

    type base 
        integer :: base_member = 0
    end type

    type, extends(base) :: derived
        integer :: derived_member = 1
    end type
end module

program test_program
    use test_module
    implicit none

    type(derived) :: d_ins

end program
```
</details>

<details>
<summary>ASR</summary>

[...]

PR #7469, opened 2025-05-28
refactor(ASR): remove `ClassType`

kmr-srbh:
Part of #4263 

- We use the `is_cstruct` field to differentiate between
  classes and structs. We add a utility function
  `ASRUtils::is_class_type()` for this.

- For code logic where only a regular struct is required,
  we add an additional check `!ASRUtils::is_class_type()`.

- For code logic where the struct and class was handled in
  the same way, we remove code for the `ClassType`.

- For other cases, we decided the code logic based on the
  value returned by `ASRUtils::is_class_type()`.


kmr-srbh:
Integrations tests go till 80%. We have a failure related to templates. Debugging it.

Issue #4545, opened 2024-07-28
Throw error for unequal character lengths in `merge` intrinsic


HarshitaKalani:
```fortran
program main
    print*, merge("okay", "ok", .true.)
end
```
```console
$ gfortran b.f90 && ./a.out
b.f90:2:18:

    2 |     print*, merge("okay", "ok", .true.)
      |                  1
Error: Unequal character lengths (4/2) in MERGE intrinsic at (1)
$ lfortran b.f90
okay
```

HarshitaKalani:
Closing this as it works and test also exists.
```console
% gfortran a.f90 && ./a.out
a.f90:2:16:

    2 |   print*, merge("okay", "ok", .true.)
      |                1
Error: Unequal character lengths (4/2) in MERGE intrinsic at (1)
% lfortran a.f90           
semantic error: Unequal character lengths in MERGE intrinsic
 --> a.f90:2:11
  |
2 |   print*, merge("okay", "ok", .true.)
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #7470, opened 2025-05-28
Append error for unequal character length for `merge` intrinsic

parth121101:
Towards: #4545


Pranavchiku (approved):

parth121101:
> If you have commits that add and remove the same files, you should squash it, to not add unused size to the git repository, since it won't improve the clarity anyway.
Sure, will keep this in mind

PR #7477, opened 2025-05-28
CI: run fortran primes with --std=f23 flag

gxyd:
## Description


Pranavchiku (commented):
we are holding this PRs as it will increase CI timings.

PR #7479, opened 2025-05-28
ci: enable other stdlib packages with separate compilation

Pranavchiku:
Towards #7328.


ci/test_third_party_codes.sh
```diff
@@ -118,7 +118,8 @@ time_section "🧪 Testing stdlib (Less Workarounds)" '
 
   git clean -dfx
   git restore .
-  git checkout 9e54a1115c9b5d5558e9db6f295849912b1202d5
+  git checout sc-lf-1
```
Pranavchiku, line 121:
```suggestion
  git checkout sc-lf-1
```

PR #7480, opened 2025-05-28
tests: uncomment integration tests which work with --std=f23 flag

gxyd:
## Description

For the integration tests which still continue to have use of `NO_STD_F23`, I'll open an issue to track them and resolve them one by one


integration_tests/CMakeLists.txt
```diff
@@ -1832,13 +1831,13 @@ RUN(NAME external_05 LABELS gfortran llvmImplicit)
 RUN(NAME external_06 LABELS gfortran llvmImplicit)
 RUN(NAME external_07 LABELS gfortran llvmImplicit)
 RUN(NAME external_08 LABELS gfortran llvmImplicit)
-RUN(NAME external_09 LABELS gfortran llvm EXTRA_ARGS --implicit-interface NO_STD_F23)
+RUN(NAME external_09 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
```
gxyd, line 1834: A lot of new integration tests are are now run with `--std=f23` flag, the reason for which PR or commit fixed them isn't easy to know for every single one of them, but I figured out using git bisect that this particular test (i.e. `external_09.f90`) was actually fixed by PR: https://github.com/lfortran/lfortran/pull/6298

It probably isn't worth the effort to know which PR fixed it for all the tests which are now run with `--std=f23`.

PR #7493, opened 2025-05-29
feat: set_allocation_size for matmul intrinsic array function

gxyd:
## Description

Documentation for matmul: https://gcc.gnu.org/onlinedocs/gfortran/MATMUL.html


HarshitaKalani (approved):
Looks good to me!

PR #7504, opened 2025-05-29
fix: do not resolve `~abstract_type` symbol  in current scope

### LINKED ISSUE #7492: [stdlib:hashmaps] Passing Derived type having `unlimited polymorphic member` in Subroutine as intent out causes Error
MRE :- 
```.f90
module stdlib_hashmap_wrappers

    implicit none

    type :: other_type
    end type other_type

end module stdlib_hashmap_wrappers


program test_open_maps

    use stdlib_hashmap_wrappers
    implicit none

    type(other_type) :: other
    call get_other_open_data( other)

contains

    subroutine get_other_open_data( other )
        type(other_type), intent(out)           :: other
    end subroutine get_other_open_data

end program test_open_maps
```

LFortran Error :- 
```console
ASR verify pass error: ExternalSymbol::m_module_name `stdlib_hashmap_wrappers` must match external's module name `other_type`
 --> try.f90:6:9
  |
6 |         class(*), allocatable :: value
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


```

MRE works well with GFortran.

kmr-srbh:
resolves #7492 

Currently, we create the `Struct` node for `~abstract_type` inside the scope which declares the unlimited polymorphic variable, here, another `Struct`. In this case we cannot import it into another scope, so the symbol table was always incorrect since we did not enforce ASR Verify on `ClassType`. This was found after the recent refactor. The fix in this PR is a hack and will eventually be removed in the upcoming symbol refactor for `StructType`.


AnshMehta1:
Thanks @kmr-srbh  for the PR. Will help me make progress in stdlib. Waiting for it to get merged.

tests/abstract_types_02.f90
```diff
@@ -0,0 +1,16 @@
...
+   type(other_type) :: other
+   call get_other_open_data( other)
+
+contains
+
+   subroutine get_other_open_data( other )
+      type(other_type), intent(out)           :: other
+   end subroutine get_other_open_data
+
+end program abstract_types_02
```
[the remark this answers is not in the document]
AnshMehta1: You can just remove the tests. I have commits ready for llvm failures, will push it  with tests after your pr gets merged @kmr-srbh .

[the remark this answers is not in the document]
kmr-srbh: Got an MRE for this. I will fix it soon.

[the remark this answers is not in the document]
kmr-srbh: > Can these tests be regular integration tests? Just enable gfortran for now.

Okay, I will register this as an integration test and also keep as a reference test to test the ASR generation.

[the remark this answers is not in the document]
kmr-srbh: > You can just remove the tests. I have commits ready for llvm failures, will push it with tests after your pr gets merged @kmr-srbh .

Okay.

src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -2376,12 +2376,17 @@ class TransformVariableInitialiser:
 
     void visit_StructType(const ASR::StructType_t& x) {
         std::string derived_type_name = ASRUtils::symbol_name(x.m_derived_type);
-        if( x.m_derived_type == current_scope->resolve_symbol(derived_type_name) ) {
-            return ;
+        if (!x.m_is_cstruct && derived_type_name == "~abstract_type") {
+            // do nothing for class(*) type
```
certik: Are you sure there is nothing to check for `class(*)`?

AnshMehta1:
Actually, I did checkout your PR, but still I am getting same `ASR verify pass error` on MRE reported in #7492 .

src/libasr/pass/array_struct_temporary.cpp
Pranavchiku: Yeah, this is not good. We rely on `~abstract_type` for this check at all places.

kmr-srbh: As `ClassType` had not been enforced by ASR verify before, the case of a `class(*)` defined inside a struct caused failure when we enforced it using checks for `StructType` as the symbol resolution was incorrect. The changes in this PR are one way to fix this - to ignore only this particular case from ASR Verify, as we will eventually remove the symbol in the upcoming refactor. Another way is to resolve the symbol correctly. I tried the first method as it seemed easier, but it seems it is harder to find every case where we need to resolve the symbol and ignore it (the CI passed, but Ansh's MRE, which was working before, failed) . I found a way to fix this through the latter. I will push the new changes.

certik:
I would add some tests that fail in main, but work here.

src/libasr/pass/array_struct_temporary.cpp
kmr-srbh: Done.

tests/reference/asr-class_abstract_type-cfa0047.stdout
```diff
@@ -0,0 +1,224 @@
...
+                                (ExternalSymbol
+                                    5
+                                    other_type_~abstract_type
+                                    3 ~abstract_type
+                                    other_type
+                                    []
+                                    ~abstract_type
+                                    Public
+                                )
+                        })
```
kmr-srbh, line 123: `~abstract_type` is imported and resolved correctly now.

tests/class_abstract_type.f90
```diff
@@ -0,0 +1,25 @@
...
+    type(other_type) :: other
+    call get_other_open_data( other)
+
+contains
+
+    subroutine get_other_open_data( other )
+        type(other_type), intent(out) :: other
+    end subroutine get_other_open_data
+
+end program class_abstract_type
```
certik, line 25: Add this to integration tests and run gfortran on it, to ensure the test is good. If we can't compile it yet, just skip the `llvm` tag.

AnshMehta1: PR #7558 handles LLVM errors. Once that pr gets merged, we can move the tests to integration tests with all the required labels .

AnshMehta1: PR #7580 moves the test of PR to integration tests.

kmr-srbh: Thanks for doing this @AnshMehta1! :)

Issue #7484, opened 2025-05-28
Speedup CI


certik:
Right now it seems it takes 3h to finish a PR: https://github.com/lfortran/lfortran/pull/7444#issuecomment-2915956186, that's too slow. It should take 30 min max.

Looking at the speeds here: https://github.com/lfortran/lfortran/actions/runs/15295842846/usage it seems most actually finish under 30 min, and those few that go a few minutes over we should speed up.

What causes the 3h delay then? If it is waiting for other commits to finish, then I think we need to cancel the CI run when a new commit is pushed, that should help. And then we need to probably consolidate runs, and have fewer tests, with each with duration of 30 min max.

gxyd:
> How often does a bug only fail in the other versions but NOT in 11 nor 20?

I was going through the [All workflows](https://github.com/lfortran/lfortran/actions) and found one run where LLVM 10 and LLVM 19 failed and rest passed, see: https://github.com/lfortran/lfortran/actions/runs/15282795638/job/42985688071. I'm not sure why though.

gxyd:
With the PR: https://github.com/lfortran/lfortran/pull/7840 merged.

Now we've 6 CI jobs in *Quick Checks*, namely:
* LFortran CI (OS=ubuntu, LLVM=11)
* LFortran CI (OS=ubuntu, LLVM=20)
* LFortran CI (OS=macOS, LLVM=11)
* LFortran CI (OS=macOS, LLVM=20)
* LFortran CI (OS=windows, LLVM=11)
* Build LFortran to WASM and Upload

And the rest are a part of *Exhaustive Checks*.

The next thing we need to do is atleast for a week or so, let both *Quick Checks* and *Exhaustive Checks* be run as usual on all the PR's, and notice if on any PR if *Quick Checks* pass but the *Exhaustive Checks* fail, or any different failure noticed on *Exhaustive Checks* which isn't producible in *Quick Checks*, and that will help us find anything which might be untested with *Quick Checks* and we *might* like to test with *Quick Checks* and not just *Exhaustive Checks*.

Once the above is done, a week or so after that we'll only run *Quick Checks* by default and *Exhaustive Checks* will need to be run manually.


[...]

PR #7509, opened 2025-05-30
CI: refactor CI into quick check and exhaustive check

gxyd:
## Description

This wouldn't work currently, so I'll keep it under draft mode for now

Towards: https://github.com/lfortran/lfortran/issues/7484


.github/workflows/CI.yml
```diff
@@ -591,6 +591,7 @@ jobs:
         shell: bash -e -x -l {0}
         run: |
             export PATH="$(pwd)/src/bin:$PATH"
+            export LFORTRAN_RUN_EXHAUSTIVE_TESTS=0
```
gxyd, line 594: these changes are currently added only to make sure that it passes `LFORTRAN_RUN_EXHAUSTIVE_TESTS` to `ci/test_third_party_codes.sh`.

ci/new_environment.yml
```diff
@@ -0,0 +1,21 @@
...
+  - xonsh=0.16.0
+  - re2c=3.1
+  - numpy=1.26.4
+  - rapidjson=1.1.0
+  - ninja=1.11.1
+  - zlib
+  - cmake=3.29.1
+  - shell=0.2.0
+#  - bison=3.4  [not win]
+#  - m2-bison=3.4  [win]
```
gxyd, line 21: eventually, this file will replace `ci/environment.yml` file

.github/workflows/EXHAUSTIVE.yml
```diff
@@ -0,0 +1,646 @@
+# https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow
+name: Exhaustive check
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'Pull Request number (optional, defaults to main)'
+        required: false
+
```
gxyd: @certik , can you help me understand why I'm not able to manually trigger this workflow in LFortran's repository?

Is it because it has to be on *main* branch for us to be able to trigger it manually? On LFortran's repository the below screenshot is what I see at the link: https://github.com/lfortran/lfortran/actions/workflows/EXHAUSTIVE.yml
Screenshot 2025-06-04 at 1 35 49 PM


the two runs that you see are actually from the past from this PR when I had intentionally made EXHAUSTIVE.yml to be run without any condition (i..e. not needing manual workflow run)


Also, I'm able to see the option to run the workflow on my own repository (i.e. gxyd/lfortran's `main`, keep in mind that it's on main of `gxyd/lfortran`), see: 
Screenshot 2025-06-04 at 1 37 36 PM for link: https://github.com/gxyd/lfortran/actions/workflows/gxyd_exhaustive.yml

gxyd: Ah yes, this is confirmed by https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually

which says:
> To trigger the workflow_dispatch event, your workflow must be in the default branch. For more information about configuring the workflow_dispatch event, see [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch).

~~So, I'll need to first have EXHAUSTIVE.yml in *main* branch, and then I can update that in this PR.~~

~~I'll create a PR for this.~~

I'll make exhaustive.yml first as to run without `workflow-displatch` and make it's run pass and then later I can make it with `workflow-dispatch`

gxyd:
The only failing CI check is *Quick check / LFortran CI (ubuntu-latest, LLVM 11)* for the Fortran backend for the only test `format_08.f90`, and it isn't a random failure, even previous failures of CI were just that. I'll try to figure out why that's failing.

Everything else is passing, indicating that things looks good.

gxyd:
> The only failing CI check is _Quick check / LFortran CI (ubuntu-latest, LLVM 11)_ for the Fortran backend for the only test `format_08.f90`, and it isn't a random failure, even previous failures of CI were just that. I'll try to figure out why that's failing.
> 
> Everything else is passing, indicating that things looks good.

I wasn't sure for why it failed on ubuntu with LLVM version 11 for `format_08.f90` and it doesn't run on LLVM 20 or macOS either, but they run fine on my local machine (macOS).

So, I've enabled it on macOS in the last commit the running of `Test Fortran backend`, and if it passes, it's possible that I'll need help from someone with access to an ubuntu machine to check for me, if it fails for them locally or not.

gxyd:
With the lastest commit, we've a failure of `Build Linux` (with ubuntu + LLVM 11) with the below error:
```
thread 'main' panicked at crates/shell/src/main.rs:191:82:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I know that such kind of errors occurs when `shell` can't find the file it's passed i.e. in this case `ci/build_tmp.sh`, which is an odd failure in this case as it passes clearly on macOS, and even with ubuntu + LLVM 20

gxyd:
> thread 'main' panicked at crates/shell/src/main.rs:191:82:
> called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

This is a little tricky to understand for why this fails on ubuntu with just LLVM 11 (I don't have access to ubuntu either), so I'm trying out a few things in the PR to understand the reason of failure.

gxyd:
Closing this PR in favor of https://github.com/lfortran/lfortran/pull/7637

PR #7514, opened 2025-05-30
 fix: handle parameter which is instatiated using structConstructor

### LINKED ISSUE #7422: Pass pointer to value inside `merge` intrinsic when `StructInstanceMember` is passed
MRE:
```fortran
module toml_val_mod
  implicit none
  type :: val_type
        integer :: origin = 3
        integer :: dest = 2
    end type
  type(val_type), parameter :: val = val_type()

end module toml_val_mod

program merge_example
    use toml_val_mod, only: val
    integer :: stat
    logical :: temp 
    stat = merge(val%origin, stat, temp)
end program merge_example
```

LFortran Error:
```console
i32 3
 i32*  %2 = call i32 @_lcompilers_merge_i32(i32 3, i32* %stat1, i1* %temp)
```

### LINKED ISSUE #7427: Handle struct parameters members inside a `functionCall` and `print`
MRE:
```fortran 
program test_program
    type :: toml_value
        character(len=1) :: key = "i"
    end type
    type(toml_value), parameter :: temp = toml_value()
    print *, char_num(temp%key)
contains 
    function char_num(c) result(x)
    character(len=1), intent(in) :: c 
    logical :: x
    end function
end program
```

LFortran Error:
```console
i8* getelementptr inbounds ([2 x i8], [2 x i8]* @1, i32 0, i32 0)
 i8**  call void @_lfortran_strcpy_pointer_string(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @1, i32 0, i32 0), i8* getelementptr inbounds ([2 x i8], [2 x i8]* @2, i32 0, i32 0))
i8* getelementptr inbounds ([2 x i8], [2 x i8]* @4, i32 0, i32 0)
 i8**  %6 = call i1 @char_num(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @4, i32 0, i32 0))
```

jinangshah21:
Fixes #7422
Fixes #7427

### DIFF
--- a/integration_tests/class_21.f90
+++ b/integration_tests/class_21.f90
@@ -4,10 +4,14 @@ program class_21
     end type
     integer :: stat
     class(val_type), allocatable :: val
+    logical :: temp = .false.
+    type(val_type), parameter :: val_par = val_type()
     type(val_type) :: tmp
     allocate(val)
     stat = merge(val%origin, stat, .true.)
     tmp = val_type(merge(val%origin - 1, stat, any([stat, val%origin] /= 0)))
     if (stat /= 3) error stop
+    stat = merge(val_par%origin, stat - 1, temp)
+    if (stat /= 2) error stop
     if (tmp%origin /= 2) error stop
 end program


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3750,7 +3750,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                 is_struct_const = true;
                             }
                         }
-                        if (is_derived_type) {
+                        if (is_derived_type || storage_type == ASR::storage_typeType::Parameter) {
```
gxyd, line 3753: Why do we need the condition that it's declared as a parameterized constant here?

jinangshah21: Because if struct variable is parameter than initialization is should be `StructConstant`. So for that we need to do `is_struct_const = true`

gxyd: Ah, I see. So the logic is that for a declaration as `TypeType`, we just check if it's a parameterized declaration in a program, then we set `is_struct_constant = true`. Makes sense now.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -10103,7 +10103,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                 !(orig_arg && !LLVM::is_llvm_pointer(*orig_arg->m_type) &&
                                 LLVM::is_llvm_pointer(*arg_type) &&
                                 !ASRUtils::is_character(*orig_arg->m_type) &&
-                                ASRUtils::is_descriptorString(arg_type)) && !ASR::is_a<ASR::StructInstanceMember_t>(*x.m_args[i].m_value) ) {
+                                ASRUtils::is_descriptorString(arg_type)) &&
+                                (!ASR::is_a<ASR::StructInstanceMember_t>(*x.m_args[i].m_value) || ASRUtils::is_value_constant(x.m_args[i].m_value)) ) {
```
gxyd, line 10107: I'm a little skeptical when handling this kind of code in any backend itself, why don't we replace this directly in the ASR or ASR pass itself?

This works sure, but I'm not sure if this is the best approach. What I see is an assignment node like:
```clj
                    [(Assignment
                        (Var 4 stat)
                        (FunctionCall
                            1 _lcompilers_merge_i32
                            1 _lcompilers_merge_i32
                            [((StructInstanceMember
                                (Var 4 val)
                                4 1_val_type_origin
                                (Integer 4)
                                (IntegerConstant 3 (Integer 4) Decimal)
                            ))
                            ((Var 4 stat))
                            ((Var 4 temp))]
                            (Integer 4)
                            ()
                            ()
                        )
                        ()
                        .false.
                    )]
```

Where the `StructInstanceMember` already has a value, maybe we should just replace it with it's value?

jinangshah21: Function expects pointer to value, so we need to do that.

gxyd: Can you please tell me exactly which are you referencing? A small Fortran program to understand would be nice.

jinangshah21:
```fortran
 type(val_type), parameter :: val_par = val_type()
stat = merge(val_par%origin, stat - 1, temp)
```

Consider this here merge expects a `i32 *` pointer so we can't directly pass value of `val_par%origin`

PR #7520, opened 2025-05-30
feat: add support for class to class variable assignment

### LINKED ISSUE #7511: Separate compilation fails to generate object code for object composition
The following MRE is of a truly nasty bug, that makes it impossible to use (polymorphic) object composition with LFortran:
$ cat test.f90
```
module test_module

   type, abstract :: Composed
   end type Composed
   
   type :: Wrapper
      class(Composed), allocatable :: obj
   end type Wrapper
   
contains

   function init(obj) result(self)
      class(Composed), intent(in) :: obj
      type(Wrapper)               :: self
      self%obj = obj
   end function init
   
end module test_module
```
$ lfortran -c --generate-object-code test.f90
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

%composed_polymorphic = type { i64, %composed* }
%composed = type <{}>
%wrapper = type <{ %composed_polymorphic }>
%__vtab_composed = type { i64 }

define void @__module_test_module_init(%composed_polymorphic* %obj, %wrapper* %self) {
.entry:
  %0 = alloca %__vtab_composed, align 8
  %1 = getelementptr %__vtab_composed, %__vtab_composed* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  %2 = getelementptr %wrapper, %wrapper* %self, i32 0, i32 0
  %3 = load %composed_polymorphic, %composed_polymorphic* %2, align 8
  store %composed_polymorphic* %obj, %composed_polymorphic %3, align 8
  br label %return

  ret void
}
  store %composed_polymorphic* %obj, %composed_polymorphic %3, align 8

  store %composed_polymorphic* %obj, %composed_polymorphic %3, align 8



```

kmr-srbh:
resolves #7511

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5482,7 +5482,7 @@ public:
         bool is_value_set = ASR::is_a<ASR::Set_t>(*asr_value_type);
         bool is_target_struct = ASR::is_a<ASR::StructType_t>(*asr_target_type) && !ASRUtils::is_class_type(asr_target_type);
         bool is_value_struct = ASR::is_a<ASR::StructType_t>(*asr_value_type) && !ASRUtils::is_class_type(asr_value_type);
-        // a class variable is always either allocatable or pointer
+        // a class variable is always either an allocatable, a pointer, or a dummy argument of a procedure
         bool is_target_class = ASR::is_a<ASR::StructType_t>(
                                    *ASRUtils::type_get_past_allocatable_pointer(asr_target_type))
                                && ASRUtils::is_class_type(
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5482,6 +5482,15 @@ public:
         bool is_value_set = ASR::is_a<ASR::Set_t>(*asr_value_type);
         bool is_target_struct = ASR::is_a<ASR::StructType_t>(*asr_target_type) && !ASRUtils::is_class_type(asr_target_type);
         bool is_value_struct = ASR::is_a<ASR::StructType_t>(*asr_value_type) && !ASRUtils::is_class_type(asr_value_type);
+        // a class variable is always either allocatable or pointer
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5447,6 +5447,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         bool is_value_set = ASR::is_a<ASR::Set_t>(*asr_value_type);
         bool is_target_struct = ASR::is_a<ASR::StructType_t>(*asr_target_type) && !ASRUtils::is_class_type(asr_target_type);
         bool is_value_struct = ASR::is_a<ASR::StructType_t>(*asr_value_type) && !ASRUtils::is_class_type(asr_value_type);
+        // a class variable is always either allocatable or pointer
```
gxyd: Is this verified by our asr_verify? If not, maybe we can have an `LCOMPILERS_ASSERT`?

kmr-srbh: We need to make many changes to the front-end for `class` variables, so maybe we could address it there?

difference-scheme: Just for the sake of completeness: as per the standard, a variable that is declared with the `class`specifier must either be an `allocatable`, a `pointer`, or the dummy argument of a procedure (Constraint C708 in the Fortran 2018 standard).

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5471,6 +5471,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
```
```suggestion
        // a class variable is always either an allocatable, a pointer, or a dummy argument of a procedure
```

kmr-srbh:
> Can this be merged? 

Yes.

jinangshah21:
Is this PR fixing #7411, than I will close #7521

kmr-srbh:
> Is this PR fixing #7411, than I will close #7521

Yes, the MRE works.

PR #7529, opened 2025-05-31
More intrinsics

swamishiju:
- set_item,
-  get_item
-  list_reverse
-  list_count

I have a problem with `_lfortran_set_item` it sounds like a intrinsic related to set. 
I also fixed some bugs from my previous PR's libasr commits which i'm surprised didn't show up earlier.


kmr-srbh:
> I have a problem with _lfortran_set_item it sounds like a intrinsic related to set.

Is `_lfortran_set_item` generic? If not and it works only for lists, you may call it `_lfortran_set_list_item` to avoid ambiguity. Likewise for `_lfortran_get_item`.

certik:
> previous PR's libasr commits 

It means it wasn't tested. That's why we are adding these tests to thoroughly test everything.

PR #7542, opened 2025-06-01
fix: initialize char length=1 if only kind is provided

jinangshah21:
Fixes #7541

Issue #7496, opened 2025-05-29
compiling `fortran-shlex`


Pranavchiku:
```
git clone https://github.com/perazz/fortran-shlex.git
cd fortran-shlex
fpm --compiler=lfortran build
```

```console
% fpm --compiler=lfortran build
 + mkdir -p build/dependencies
shlex_module.f90                       failed.
[ 50%] Compiling...
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2603
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 1128
    result = fe.get_asr2(input, lm, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 112
    if (res.ok) {
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 5267
    b.is_body_visitor = true;
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 166
    visit_ast(*x.m_items[i]);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4969
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4924
    case astType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4972
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
[...]

PR #7560, opened 2025-06-02
CI: update ci for fortran-shlex build

jinangshah21:
After #7542 merge this will be good.

towards #7496 
Towards #6840


jinangshah21:
`fortran-shlex` is failing with macOS, can anyone with mac check?

Update: it fails for Ubuntu also, need to look.

jinangshah21:
There are recent commits added in fortran-shlex and there are some errors there. So currenty just compiling, not testing

PR #7561, opened 2025-06-02
CI: use a new github group on PR's merged in main

gxyd:
## Description

Currently, the CI for *main* branch is run in single group named `'main'` in a queue, and that saves resources for PR's, but that approach seems to fail when PR's are merged without much time gap between them, and the older PR's CI get's cancelled on *main* in that case (which is bad), e.g.;

for the below three CI runs merged in *main* (see screenshot below) in chronological order from bottom to top, the three PR's were merged at a time gap of 3 minutes, but the CI run of the bottom two PR's on *main* was cancelled altogether with the error:

```console
Canceling since a higher priority waiting request for CI-main exists
```

Screenshot 2025-06-02 at 10 09 23 PM

Hence, this PR makes sure that this doesn't happen by alloting to each commit merged in *main*, it's own unique group.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,16 +10,15 @@ on:
     branches:
       - main
 
-# For a PR #7488 against main branch, the `group` will become: CI-{{ 7488 || 'main' }}
-# which eventually evaluates to: CI-7488 and 'main' isn't used.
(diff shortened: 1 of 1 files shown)


certik:
It might be working:

image

It used to cancel all jobs in main, but now it queued the last two.

PR #7563, opened 2025-06-02
fix: don't deepcopy allocable string which is not allocated in struct

### LINKED ISSUE #7548: `Struct` type used when `SIM` is passed as argument in subroutine is incorrect
Towards #7495 

MRE:
```fortran
program string_len
     type :: regex_token
        character(len=:), allocatable :: ccl
    end type regex_token

    type :: regex_pattern
        type(regex_token), dimension(2) :: pattern
    end type regex_pattern

    type(regex_pattern) :: my_pattern
    type(regex_token) :: token
    token = my_pattern%pattern(1)
end program
```

LFortran Error:
```console
Runtime Error : RHS allocatable-character variable must be allocated before assignment.
```

jinangshah21:
Fixes #7548


assem2002:
I can say, try out to actually neglect the error (Remove that check). instead of checking at runtime.
The string check seems unreasonable to some extent.

jinangshah21:
@assem2002, for LLVM > 16, strings are represented as `ptr` now how to check if it is allocated or not? Because when we are doing deepcopy then we don't want to copy str which is not allocated.

assem2002:
I'm not quite sure I get the issue here, but what I'm trying to suggest is to remove the whole check that raises the error you're trying to bypass, it doesn't seem a right check to do.

PR #7565, opened 2025-06-02
chore: Added integration test for hashmap struct

### LINKED ISSUE #7393: [stdlib:hashmaps] Handle `Subroutine Call` which passes a polymorphic argument for a derived type to a subroutine accepting type variable for same derived type
MRE :- 
```.f90
module stdlib_hashmaps

    implicit none

    type :: open_hashmap_type
        integer(4) :: slots
    end type open_hashmap_type

end module stdlib_hashmaps


program main
    use stdlib_hashmaps

    implicit none
    type(open_hashmap_type)  :: map

    call in_open_map(map)

contains

    subroutine expand_open_slots(map)
        type(open_hashmap_type), intent(inout) :: map
    end subroutine expand_open_slots


    subroutine in_open_map(map)
        call expand_open_slots(map)
    end subroutine in_open_map
end program
```

LFortran CodeGen Error :-  

```console
%open_hashmap_type_polymorphic* %map
 %open_hashmap_type*  call void @expand_open_slots(%open_hashmap_type_polymorphic* %map)
```

MRE works well with GFortran .

sriganeshres:
fixes #7393

### DIFF
--- a/integration_tests/hashmap_struct_01.f90
+++ b/integration_tests/hashmap_struct_01.f90
@@ -1,0 +1,25 @@
+module stdlib_hashmaps_hashmap_struct_01
+    implicit none
+    type :: open_hashmap_type
+        integer(4) :: slots
+    end type open_hashmap_type
+end module stdlib_hashmaps_hashmap_struct_01
+
+program hashmap_struct_01
+    use stdlib_hashmaps_hashmap_struct_01
+    implicit none
+    type(open_hashmap_type) :: map
+    call in_open_map(map)
+
+contains
+
+    subroutine expand_open_slots(map)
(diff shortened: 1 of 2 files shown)


integration_tests/hashmap_struct_01.f90
```diff
@@ -0,0 +1,25 @@
+module stdlib_hashmaps
```
AnshMehta1: Append test file name at the end of module name like `stdlib_hashmaps_<file_name>`.

AnshMehta1 (approved):

integration_tests/hashmap_struct_01.f90
sriganeshres: Done.

Issue #6067, opened 2025-01-20
bug: `c_long_double` not supported


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
    use iso_c_binding, only: c_long_double
    implicit none

    print *, c_long_double
end program main
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
          10
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
-1
```

Likewise for `c_long_double_complex`.

vmagnin:
Indeed, `c_long_double` and `c_long_double_complex` are not yet implemented in https://github.com/lfortran/lfortran/blob/cc213c5a2b7aece217d15911cb4045ae8437fe95/src/runtime/pure/lfortran_intrinsic_iso_c_binding.f90

```fortran
integer, parameter :: c_long_double = -1

integer, parameter :: c_long_double_complex = -1
```

The fix is to be patient :-) 
I am now preparing CMake gtk-fortran scripts and `./alt_build_test.sh` for when LFortran will be ready.

vmagnin:
In my `gtk4-vmagnin` development branch, I have updated the script `src/alt_build_test.sh` to filter GTK flags unsupported by LFortran 0.52.0 (in Fedora Rawhide): `-msse2 -msse -mfpmath -pthread`

The first error encountered is concerning `iso_c_binding`:
```bash
$ FC="lfortran" ./alt_build_test.sh 
>>> Building gtk-4-fortran with these options:
[...]

PR #7566, opened 2025-06-02
enh: auto-generate `lfortran_intrinsic_iso_c_binding.f90` for portability

kmr-srbh:
Part of #6067


kmr-srbh:
Some things to consider:

- Constants like `c_long_double` have value 8 on macOS vs 10 on Linux. There can be more differences like this, so we will have to implement this accordingly in the future. Currently our front-end catches unsupported kind values and throws an error, so the implementation in the `iso_c_binding` file is standalone. Some kind parameters for character are left to be implemented.
- The `iso_c_binding` webpage by GFortran says the constants `c_x_complex` where x is `float`, `double` or `long double` are equivalent to `x _Complex` in C, but the values returned by the GNU C compiler and GFortran do not agree. For example, the value returned by GNU C compiler for `sizeof(float _Complex)` is 8 compared to 4 for `c_float_complex`, by GFortran on macOS.

src/runtime/pure/generate_iso_c_binding.c
```diff
@@ -0,0 +1,165 @@
...
+const size_t c_double = sizeof(double);
+
+const size_t c_long_double = sizeof(long double);
+
+// complex types
+const size_t c_float_complex = sizeof(float);
+
+const size_t c_double_complex = sizeof(double);
+
+const size_t c_long_double_complex = sizeof(long double);
```
kmr-srbh, line 65: This is currently set to `sizeof(float)` etc so GFortran agrees. This is not the standard though. We need to figure this out correctly.

build0.sh
```diff
@@ -18,6 +18,9 @@ python src/libasr/wasm_instructions_visitor.py
 # Generate the intrinsic_function_registry_util.h (C++)
 python src/libasr/intrinsic_func_registry_util_gen.py
 
+# Generate file lfortran_intrinsic_iso_c_binding.f90 (Fortran)
+clang src/runtime/pure/generate_iso_c_binding.c && ./a.out
```
certik, line 22: This adds fragility to our build system.

The other way to do it is to let LFortran decide, at runtime, what the values should be.

How do other compilers handle it?

PR #7570, opened 2025-06-03
CI: queue and group management in main branch

gxyd:
## Description

Try to see if we can have a single queue in *main* with all of the commit's merged in *main* belonging with to a single group `CI-main` but also at the same time making sure that older CI runs are preferred to run over the newer CI runs

Currently we've three workflows running for three PR's which were merged recently to *main* and because of that my PR's are all waiting to be *queued*.

Screenshot 2025-06-03 at 2 52 48 PM

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,15 +10,16 @@ on:
     branches:
       - main
 
-# For a PR #7488 against main branch, the `group` will become: CI-{{ 7488 || github.sha }}
-# which eventually evaluates to: CI-7488 and 'sha' isn't used.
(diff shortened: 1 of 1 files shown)


certik:
I would prefer not to depend on a 3rd party for a critical piece of infrastructure like this. The package also doesn't seem to have very many recent commits. I think the current solution in main seems fine.

gxyd:
> I would prefer not to depend on a 3rd party for a critical piece of infrastructure like this.

Makes sense.

I think I should close this PR then, I'm hoping that the PR: https://github.com/lfortran/lfortran/pull/7509 gives us enough improvements in CI timings that maybe we don't need something like this anymore at all.

Issue #7569, opened 2025-06-03
CI: Consolidate the build script


certik:
Some ideas:
* [ ] Use Ninja on all platforms
* [ ] Switch from Xonsh to Shell
* [ ] Move lines like `$IS_MAC = platform.system() == "Darwin"` out of the script and into the CI yaml, as env variables

gxyd:
> * Switch from Xonsh to Shell

I see, the initial work as done in PR: https://github.com/lfortran/lfortran/pull/6064

PR #7572, opened 2025-06-03
CI: use shell build script instead of xonsh for macOS and Linux

gxyd:
## Description

Towards: https://github.com/lfortran/lfortran/issues/7569#issue-3112926459

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -91,7 +91,8 @@ jobs:
         run: |
             export CXXFLAGS="-Werror"
             export CFLAGS="-Werror"
-            xonsh ci/build_tmp.xsh
+            export WIN=0
+            shell ci/build_tmp.sh
 
       - name: Build (Windows)
         if: contains(matrix.os, 'windows')
--- a/ci/build_tmp.sh
+++ b/ci/build_tmp.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env shell
+
+set -ex
+
+echo "Running SHELL"
+
+echo "CONDA_PREFIX=$CONDA_PREFIX"
+llvm-config --components
+
+# Generate the `version` file
+bash ci/version.sh
+
+# Generate a Fortran AST from AST.asdl (C++)
+python src/libasr/asdl_cpp.py grammar/AST.asdl src/lfortran/ast.h
+# Generate a Fortran ASR from ASR.asdl (C++)
+python src/libasr/asdl_cpp.py src/libasr/ASR.asdl src/libasr/asr.h
+# Generate a wasm_visitor.h from src/libasr/wasm_instructions.txt (C++)
+python src/libasr/wasm_instructions_visitor.py
+# Generate the intrinsic_function_registry_util.h (C++)
+python src/libasr/intrinsic_func_registry_util_gen.py
+
+# Generate the tokenizer and parser
+echo $(pwd)
+cd src/lfortran/parser && re2c -W -b tokenizer.re -o tokenizer.cpp && cd ../../..
+cd src/lfortran/parser && re2c -W -b preprocessor.re -o preprocessor.cpp && cd ../../..
+cd src/lfortran/parser && bison -Wall -d -r all parser.yy && cd ../../..
+
+# Generate the LSP sources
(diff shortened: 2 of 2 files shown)


ci/build_tmp.sh
```diff
@@ -0,0 +1,76 @@
...
+cd src/lfortran/parser && re2c -W -b tokenizer.re -o tokenizer.cpp && cd ../../..
+cd src/lfortran/parser && re2c -W -b preprocessor.re -o preprocessor.cpp && cd ../../..
+cd src/lfortran/parser && bison -Wall -d -r all parser.yy && cd ../../..
+
+# Generate the LSP sources
+python src/server/generator/generate_lsp_code.py --schema src/server/generator/metaModel.json --target-language c++ --output-dir src/server
+
+pandoc --standalone --to man doc/man/lfortran.md -o doc/man/lfortran.1
+
+lfortran_version=$(cat version)
```
gxyd, line 34: this command is where it creates problems for `shell`, but it works just fine with `bash`. `shell` behaves in an odd manner with this. I'll need to figure out a fix.

But I ensure that if I explicitly set `lfortran_version=0.52.0-550-g9887efa53-dirty`, then the complete script works with `shell` and `bash` both.

.github/workflows/CI.yml
```diff
@@ -91,7 +91,8 @@ jobs:
         run: |
             export CXXFLAGS="-Werror"
             export CFLAGS="-Werror"
-            xonsh ci/build_tmp.xsh
+            export WIN=0
+            shell ci/build_tmp.sh
```
gxyd, line 95: Currently, I've only made use of this on linux/macOS, but once it works on both of them, I'll try replacing it on windows as well.

ci/build_tmp.sh
gxyd: As seen on the CI run: https://github.com/lfortran/lfortran/actions/runs/15418282704/job/43386252624?pr=7572, it sets `lfortran_version` as:
```console
lfortran_version=+ cat version 0.52.0-563-gda2c6ac5d
```

ci/build_tmp.sh
```diff
@@ -44,18 +21,19 @@ python src/libasr/wasm_instructions_visitor.py
...
 python src/server/generator/generate_lsp_code.py --schema src/server/generator/metaModel.json --target-language c++ --output-dir src/server
 
 pandoc --standalone --to man doc/man/lfortran.md -o doc/man/lfortran.1
 
-$lfortran_version=$(cat version).strip()
-$dest="lfortran-" + $lfortran_version
-bash ci/create_source_tarball0.sh
+lfortran_version=$(cat version)
+dest="lfortran-$lfortran_version"
+dest=$dest bash ci/create_source_tarball0.sh
```
gxyd: this line seems to be passing `dest` correctly on macOS/Linux while passing it as empty on Windows (or not passing it at all really)

gxyd:
Struggling with fixing the current failure on Windows, and I'm doubting if it's worth moving from xonsh to shell at all really, since I don't think that the current script I've used for shell is wrong as it's working on macOS / Linux but failing on Windows 😫

ci/build_tmp.sh
```diff
@@ -88,9 +67,11 @@ jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=120 --ou
 jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=120 --output Variables_out.ipynb Variables.ipynb
 cd ../../..
 
-cp lfortran-$lfortran_version/test-bld/src/bin/lfortran src/bin
-if $IS_WIN:
+if [[ $WIN == "1" ]]; then # Windows
+    cp lfortran-$lfortran_version/test-bld/src/bin/lfortran.exe src/bin
```
gxyd, line 71: Earlier we were using `cp` for `lfortran` file which works to make sure that it copies lfortran.exe in xonsh, but we need to be explicit about the complete name in `shell`, my guess was right about it here.

ci/build_tmp.sh
certik: Typically this would work, but we have to try:
```
set +x
lfortran_version=$(cat version)
set -x
```

Issue #6554, opened 2025-03-09
error not reported: compiler_options must be imported first


certik:
```fortran
program test_compiler_options
implicit none
print *, compiler_options()
end program
```
This gives:
```console
$ lfortran a.f90
/Users/ondrej/bin/lfortran
$ gfortran a.f90
a.f90:3:8:

    3 | print *, compiler_options()
      |        1
Error: Function ‘compiler_options’ at (1) has no IMPLICIT type
```

One must `use iso_fortran_env`, then it works. LFortran should report an error without using the module.

PR #7573, opened 2025-06-03
fix: import error for compiler_options from iso_fortran_env

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6554
fixes: https://github.com/lfortran/lfortran/issues/6981

### DIFF
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -1393,7 +1393,8 @@ public:
     }
 
     void visit_IntrinsicElementalFunction_helper(std::string &out, std::string func_name, const ASR::IntrinsicElementalFunction_t &x) {
-        if ( x.m_intrinsic_id == static_cast<int64_t>(ASRUtils::IntrinsicElementalFunctions::CompilerVersion) ) {
+        if ( x.m_intrinsic_id == static_cast<int64_t>(ASRUtils::IntrinsicElementalFunctions::CompilerVersion) ||
+             x.m_intrinsic_id == static_cast<int64_t>(ASRUtils::IntrinsicElementalFunctions::CompilerOptions) ) {
             src = "";
             visit_expr(*x.m_value); // we will always have a value
             out += src;
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "5381406e8c856ea96f950057abdc161c1f0c7fea195533659bc22149",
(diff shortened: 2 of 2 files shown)


Pranavchiku (commented):
CI fails, marking it draft

HarshitaKalani:
The test ./build/fortran/example_bobyqa_fortran_2_exe passes in local for me.

integration_tests/CMakeLists.txt
```diff
@@ -2184,7 +2184,7 @@ RUN(NAME arrayprint_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME print_arr_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME binop_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME binop_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME binop_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NOFAST_TILL_LLVM16)
```
Pranavchiku, line 2187:
```suggestion
RUN(NAME binop_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST)
```

Pranavchiku: The third party test fails with `--fast` so you'll have to fix this to get those working.

HarshitaKalani:
There's an issue doing it this way. The mod file (iso_fortran_env) generated will not have the compiler options provided by user at runtime. So we'll not be able to return the same. We don't return correct output with the changes in this PR.
We need to look for a better way of doing this.

certik:
I see --- it stores the compiler options when the `iso_fortran_env.mod` file was created, now when the user actually compiles his code at runtime.

I guess the question is --- should the `compiler_options` be replaced with the string of the compiler options when this part of source code is compiled to a mod file, or when the main program is compiled?

I would say it's when that part of the source code gets compiled. The user can then put this into the main program or whichever place he wants to be tracked.

I would recommend creating CompilerOptions ASR node, but set it inplace where `compiler_options` is used. Something like this:
```fortran
subroutine f()
use iso_fortran_env, only: compiler_options
print *, compiler_options()
end subroutine
```
and it would become:
```clojure
(Function ...
[(Print (CompilerOptions "-O3 --implicit-typing"))]
)
```
I would set the string directly in CompilerOptions. That way, when printing Fortran source code, we have the option to convert back to either compiler_options (if the user wanted that), or to a string of the actual options used (if the user wanted that instead). In the LLVM backend we simply use the string inside CompilerOptions.

This design stores the compiler options used when compiling the `compiler_options()` intrinsic. However it leaves the door open to later replace it with compiler options used to compile some later stage, for example some lfortran command line option can select this. I don't know if that would ever be useful, but it could be done. So I think the above design seems good enough that doesn't close doors.

I think it was me who suggested to use the intrinsic function `_lfortran_compiler_options`, but I think it won't work, so I apologize for that. I think in AST->ASR, we need to replace `compiler_options` directly with `CompilerOptions`.

I made this https://github.com/lfortran/lfortran/issues/7730.

Issue #7587, opened 2025-06-04
use `m_intrinsic` to identify if a module is intrinsic or not


Pranavchiku:
Refer https://github.com/lfortran/lfortran/pull/7586/files#r2126695871

PR #7589, opened 2025-06-04
use `m_intrinsic` to check module is intrinsic or not

Jaysukh-409:
Fix: #7587


Pranavchiku (approved):
Thank you, looks good to me!

PR #7592, opened 2025-06-04
fix: don't create load when funtion return type in pointer string

### LINKED ISSUE #7591: Handle `character` pointer as function return type
MRE:
```fortran
program string_len
    character(len=:), pointer :: temp
    temp => cast_string()
contains 
    function cast_string() result(ptr)
        character(:), pointer :: ptr
    end function cast_string
end program
```

LFortran error:
```console
Function return type does not match operand type of return inst!
  ret i8* %0
 i8**
```

jinangshah21:
Fixes #7591

### DIFF
--- a/integration_tests/class_21.f90
+++ b/integration_tests/class_21.f90
@@ -7,6 +7,7 @@ program class_21
     use class_21_mod
     integer :: stat
     class(val_type), allocatable :: val
+    class(val_type), pointer :: val2
     logical :: temp = .false.
     type(val_type), parameter :: val_par = val_type()
     type(val_type) :: tmp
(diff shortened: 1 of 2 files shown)


integration_tests/string_56.f90
```diff
@@ -0,0 +1,11 @@
...
+    character(len=:), pointer :: temp
+    temp => cast_string()
+    if (temp /= "Hello") error stop
+contains 
+    function cast_string() result(ptr)
+        character(:), pointer :: ptr
+        character(len=5), save, target ::tmp = "Hello"
+        ptr => tmp
+    end function cast_string
+end program
```
certik: I think this test is not registered.

jinangshah21: Sorry, I have done it now.

certik:
The new test fails:
```
1128/1615 Test #1128: string_56_FAST ..........................................***Failed    0.01 sec
ERROR STOP
```

Issue #4147, opened 2024-06-04
incorrect parse of `!$omp parallel`


Pranavchiku:
With #4140, the following example gives:

```fortran
program openmp
use omp_lib

integer :: thread_id

!$omp parallel private(thread_id)

thread_id = omp_get_thread_num()
print *, "hello from process: ", thread_id
!$omp end parallel
end program

```

> lfortran a.f90 --openmp --show-asr

```clojure
                    [(Assignment
                        (Var 2 thread_id)
                        (FunctionCall
                            2 omp_get_thread_num
                            ()
                            []
                            (Integer 4)
                            ()
                            ()
                        )
                        ()
                    )
                    (Print
                        [(StringConstant
                            "hello from process: "
                            (Character 1 20 ())
                        )
                        (Var 2 thread_id)]
                        ()
                        ()
                    )
                    (DoConcurrentLoop
                        (()
                        ()
                        ()
                        ())
                        []
                        [(Var 2 thread_id)]
                        []
                    )]
```

The above ASR is incorrect and we cannot use a `DoConcurrentLoop` here.

How about introducing a new node `OMPParallel`

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index d1356151b..5dbba5ee1 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -35,6 +35,7 @@ stmt
     | Cycle(identifier? stmt_name)
     | ExplicitDeallocate(expr* vars)
     | ImplicitDeallocate(expr* vars)
+    | OMPParallel(expr* shared, expr* local, stmt* body)
     | DoConcurrentLoop(do_loop_head head, expr* shared, expr* local, stmt* body)
     | DoLoop(identifier? name, do_loop_head head, stmt* body, stmt* orelse)
     | ErrorStop(expr? code)
```

cc @nikabot @certik

PR #7593, opened 2025-06-04
[OPENMP] FEAT: Replace DonConcurrentLoop conversion with OMPRegion Node for the ParallelDo region type and extend OpenMp pass to visit OMPRegion ASR Node

adit4443ya:
Towards #7332

This PR intends to Represent `OMP PARALLEL DO` in `OMPREGION` ASR node instead of DoConcurrentLoop.
In this PR I did the following changes:

- Removed the older logic of converting the `parallel do` pragmas to `DonConcurrentLoop` in the AST->ASR conversion.
- Extended the Openmp pass to visit the `OMPRegion` ASR node and handle `parallel do` constructs just like we do in the DoConcurrentLoop
- I am making sure that we do not face any kind of regression on the OMP testcases which we compile till now

Benefits of current approach in the openmp:-

- Now we can handle standalone `parallel` constructs instead of just `parallel do`. That means:

<details>

```fortran
program main
    !$omp parallel
        print*,"hi"
    !$omp end parallel
end program
```

```clojure
 aditya-trivedi   lfortran    omp_refactor ≢    lfortran --openmp --openmp-lib-dir={$CONDA_PREFIX}/include --linker=gcc --dump-all-passes  integration_tests/openmp_31.f90
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
hi
```

</details>

- Further this can be extended to multiple constructs and also we can handle combined constructs too for example:
Nested one:

<details>

```fortran
!$omp parallel
!$omp do
DO LOOP
!$omp end do
!$omp end parallel
```

</details>

Combined one:

<details>

```fortran
!$omp parallel do
DOLOOP
!$omp end parallel do
```

</details>

Similarly for sections too it can be extended and any other constructs

- Intermediate statements are also handled now with the benefit of OMPRegion ASR node design and Openmp pass extension to support it, (Towards #4147 )

Examples like these are also handled now:

<details>

```fortran
!$omp parallel
    !$omp do
        DO LOOP
    !$omp end do

   print *,"hi" // this is intermdediate stmt we strugged to represent this print stmt in asr as seen on and openmp_06.f90
!$omp end parallel
```

</details>

For Sections i have left placeholder Functions just to showcase how would it be handled too similar to ParallelDo and Do region_type, i will implement it in next PR

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -28,9 +28,13 @@ public:
     size_t starting_n_body = 0;
     int loop_nesting = 0;
     int all_loops_blocks_nesting = 0;
+    int all_blocks_nesting = 0;
     int pragma_nesting_level = 0;
     int pragma_nesting_level_2 = 0;
     bool openmp_collapse = false;
+    bool pragma_in_do_loop=false;
+    bool do_in_pragma=false;
+    int nesting_lvl_inside_pragma=0;
     int collapse_value=0;
     Vec<ASR::do_loop_head_t> do_loop_heads_for_collapse;
     Vec<ASR::stmt_t*> do_loop_bodies_for_collapse;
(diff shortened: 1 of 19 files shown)


Pranavchiku (approved):
This is good, all tests are working, reference tests look good.

Pranavchiku:
With this PR we migrated to newer design for `DoConcurrent` and `ParallelDo`, right?

adit4443ya:
Yes, correct !!

Issue #3734, opened 2024-03-28
Throw error for argument of `ichar` not equal to length 1


HarshitaKalani:
```fortran
program main
  character(5):: x = "Hello"
  print*, ichar(x)
end
```
```console
$ lfortran a.f90
72
$ gfortran a.f90 && ./a.out
a.f90:3:16:

    3 |   print*, ichar(x)
      |                1
Error: Argument of ichar at (1) must be of length one
```
LFortran should be throwing error if the argument provided to `ichar` has length not equal to 1.

HarshitaKalani:
The same needs to be done for `iachar`
```fortran
program main
  character(5):: x = "Hello"
  print*, iachar(x)
end
```
```console
$ lfortran a.f90
72
$ gfortran a.f90 && ./a.out
a.f90:3:17:

    3 |   print*, iachar(x)
      |                 1
Error: Argument of iachar at (1) must be of length one
```

PR #7611, opened 2025-06-06
Fix: throw error for argument of `ichar` intrinsic having length not equal to 1

parth121101:
Resolves: https://github.com/lfortran/lfortran/issues/3734

### Enhancements to ASR utilities:
* Added the `get_fixed_string_len` function in `src/libasr/asr_utils.h` to retrieve the fixed length of a string type, or return `-1` otherwise.

### Intrinsic function validation improvements:
* Introduced a `char_len_validation` field in `src/libasr/intrinsic_func_registry_util_gen.py` to specify constraints on string argument lengths for intrinsic functions.
* Added logic in `add_create_func_arg_type_src` to enforce string length constraints during argument validation, leveraging the `get_fixed_string_len` utility.

The design structure
```
"char_len_validation": [{0: {0: {"min": 1, "max": 1}}}]
```

First Key 0 -> It tells us the index of the argument whose char length needs to be validated. 
The key–value pair `{0: {"min": 1, "max": 1}}` indicates that the 0th argument is expected to have min length 1 and max length 1.

### DIFF
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -4998,6 +4998,15 @@ namespace Ichar {
         fill_func_arg("str", ASRUtils::TYPE(ASR::make_String_t(al, loc, 1, nullptr, ASR::string_length_kindType::AssumedLength, ASR::string_physical_typeType::PointerString)));
         auto result = declare("result", return_type, ReturnVar);
         auto itr = declare("i", int32, Local);
+
+        ASR::expr_t* len_expr = ASR::down_cast<ASR::String_t>(ASR::down_cast<ASR::Variable_t>(ASR::down_cast<ASR::Var_t>(args[0])->m_v)->m_type)->m_len;
+        if (ASR::is_a<ASR::IntegerConstant_t>(*len_expr) ) {
+            int64_t len = ASR::down_cast<ASR::IntegerConstant_t>(len_expr)->m_n;
+            if (len != 1) {
+                throw LCompilersException("Argument to Ichar must have length one");
+            }
+        }
+
         body.push_back(al, b.Assignment(itr, b.i32(1)));
         body.push_back(al, b.Assignment(result, b.i2i_t(
             ASRUtils::EXPR(ASR::make_Ichar_t(al, loc, ASRUtils::EXPR(ASR::make_StringItem_t(al, loc, args[0], itr,
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -4999,18 +4999,13 @@ namespace Ichar {
         auto result = declare("result", return_type, ReturnVar);
         auto itr = declare("i", int32, Local);
 
-        if (ASR::is_a<ASR::Var_t>(*args[0])) {
-            ASR::symbol_t* sym_t = ASR::down_cast<ASR::Var_t>(args[0])->m_v;
-            if (ASR::is_a<ASR::Variable_t>(*sym_t)) {
-                ASR::ttype_t* type_t = ASR::down_cast<ASR::Variable_t>(sym_t)->m_type;
-                if (ASR::is_a<ASR::String_t>(*type_t)) {
-                    ASR::expr_t* len_expr = ASR::down_cast<ASR::String_t>(type_t)->m_len;
-                    if (len_expr && ASR::is_a<ASR::IntegerConstant_t>(*len_expr) ) {
-                        int64_t len = ASR::down_cast<ASR::IntegerConstant_t>(len_expr)->m_n;
-                        if (len != 1) {
-                            throw LCompilersException("Argument to Ichar must have length one");
-                        }
-                    }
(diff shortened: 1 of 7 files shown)


src/libasr/pass/intrinsic_functions.h
```diff
@@ -4948,9 +4948,26 @@ namespace Ichar {
...
+        if (ASR::is_a<ASR::Var_t>(*args[0])) {
+            ASR::symbol_t* sym_t = ASR::down_cast<ASR::Var_t>(args[0])->m_v;
+            if (ASR::is_a<ASR::Variable_t>(*sym_t)) {
+                ASR::ttype_t* type_t = ASR::down_cast<ASR::Variable_t>(sym_t)->m_type;
+                if (ASR::is_a<ASR::String_t>(*type_t)) {
+                    ASR::expr_t* len_expr = ASR::down_cast<ASR::String_t>(type_t)->m_len;
+                    if (len_expr && ASR::is_a<ASR::IntegerConstant_t>(*len_expr) ) {
+                        int64_t len = ASR::down_cast<ASR::IntegerConstant_t>(len_expr)->m_n;
+                        if (len != 1) {
+                            throw LCompilersException("Argument to Ichar must have length one");
```
Pranavchiku: I think we've to find a way to append error to diagnostics, but this may be a major change as we need to update arguments of `instantiate_xx` function.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -4948,9 +4948,26 @@ namespace Ichar {
...
         declare_basic_variables("_lcompilers_ichar_" + type_to_str_python(arg_types[0]));
-        fill_func_arg("str", ASRUtils::TYPE(ASR::make_String_t(al, loc, 1, nullptr, ASR::string_length_kindType::AssumedLength, ASR::string_physical_typeType::PointerString)));
+        fill_func_arg("str",arg_types[0]);
         auto result = declare("result", return_type, ReturnVar);
         auto itr = declare("i", int32, Local);
+
+        if (ASR::is_a<ASR::Var_t>(*args[0])) {
+            ASR::symbol_t* sym_t = ASR::down_cast<ASR::Var_t>(args[0])->m_v;
+            if (ASR::is_a<ASR::Variable_t>(*sym_t)) {
+                ASR::ttype_t* type_t = ASR::down_cast<ASR::Variable_t>(sym_t)->m_type;
```
Pranavchiku:
```suggestion
                ASR::ttype_t* type_t =ASRUtils::expr_type(args[0]);
```

Pranavchiku (changes requested):
left enhancements

src/libasr/pass/intrinsic_functions.h
```diff
@@ -4948,9 +4948,21 @@ namespace Ichar {
...
         auto result = declare("result", return_type, ReturnVar);
         auto itr = declare("i", int32, Local);
+
+        ASR::ttype_t* type_t =ASRUtils::expr_type(args[0]);
+        if (ASR::is_a<ASR::String_t>(*type_t)) {
+            ASR::expr_t* len_expr = ASR::down_cast<ASR::String_t>(type_t)->m_len;
+            if (len_expr && ASR::is_a<ASR::IntegerConstant_t>(*len_expr) ) {
+                int64_t len = ASR::down_cast<ASR::IntegerConstant_t>(len_expr)->m_n;
+                if (len != 1) {
+                    throw LCompilersException("Argument to Ichar must have length one");
```
certik: This is fine for now, it's an ICE.

The caller that constructs ASR must construct it correctly. We don't need to return any error messages here into diagnostics.

certik:
I would put an assert there instead.

But the actual MRE is for a user error, so we must give a real error message in AST->ASR.

parth121101:
To address the review, 
I will work on reporting this error in AST->ASR

parth121101:
For throwing errors in AST->ASR we need modify create_functions.
Since create_functions are automatically generated using a python script.
This PR now has an initial design to check for character lengths of the function arguments. 
Please review this and let me know if these has some problems.

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -509,7 +509,7 @@ semantic error: Type mismatch in argument at argument (1); passed `real` to `int
 334 |     print *, f(42.9)
     |                ^^^^ 
 
-semantic error: Argument to Ichar must have length one
+semantic error: Argument 0 to Ichar must have length 1
```
certik, line 512: How many arguments does ichar have? 

Also, start counting arguments from 1, not from 0.

certik: Also, where is the old message being emitted? Should that code be removed now?

parth121101: ichar have 2 arguments character and kind
Old messages are emitted from eval_ichar function. Yes we can remove that now, its redundant.

parth121101:
PR continued here:
https://github.com/lfortran/lfortran/pull/7809

PR #7614, opened 2025-06-06
chore: sanity check value to not be nullptr before accessing

### LINKED ISSUE #7599: reshape on character array fails
```fortran
module regex_test_1
   !  use regex_module
    implicit none
    public
    character(len=*), parameter :: test1data(4,1) = reshape([ character(len=30) :: &
        "YES", "\d                           ", "5                             ", "1 "],[4,1])
end module regex_test_1
```

```console
% lfortran a.f90
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2726
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2653
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 1128
    result = fe.get_asr2(input, lm, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 78
    if (res.ok) {
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 4127
    v.visit_TranslationUnit(ast);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 82
    visit_ast(*x.m_items[i]);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4969
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4924
    case astType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4972
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4546
    case modType::Module: { v.visit_Module((const Module_t &)x); return; }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 388
    visit_ModuleSubmoduleCommon<AST::Module_t, ASR::Module_t>(x);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 333
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4983
    void visit_unit_decl2(const unit_decl2_t &b) { visit_unit_decl2_t(b, self()); }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/ast.h", line 4575
    case unit_decl2Type::Declaration: { v.visit_Declaration((const Declaration_t &)x); return; }
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 1692
    visit_DeclarationUtil(x);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_common_visitor.h", line 3929
    ASR::String_t *lhs_type = ASR::down_cast<ASR::String_t>(
  File "/Users/pranavchiku/repos/lfortran/src/libasr/asr_utils.h", line 361
    return ASR::expr_type0(f);
  File "/Users/pranavchiku/repos/lfortran/src/libasr/asr_expr_type_visitor.h", line 20
    LCOMPILERS_ASSERT(f != nullptr);
AssertFailed: f != nullptr
```

Pranavchiku:
Fixes #7599.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -564,6 +564,7 @@ RUN(NAME arrays_reshape_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran
 RUN(NAME arrays_reshape_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_reshape_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_reshape_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
+RUN(NAME arrays_reshape_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_elemental_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran)
 RUN(NAME arrays_03_func LABELS gfortran cpp llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm)
 RUN(NAME arrays_03_func_pass_arr_dims LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm)
--- a/integration_tests/arrays_reshape_24.f90
+++ b/integration_tests/arrays_reshape_24.f90
@@ -1,0 +1,17 @@
+module regex_test_1_arrays_reshape_24
+   !  use regex_module
+    implicit none
+    public
+    character(len=*), parameter :: test1data(4,1) = reshape([ character(len=30) :: &
+        "YES", "\d                           ", "5                             ", "1 "],[4,1])
+end module regex_test_1_arrays_reshape_24
+
+program arrays_reshape_24
+    use regex_test_1_arrays_reshape_24
+    implicit none
+    print *, test1data
+    if (test1data(1,1) /= "YES") error stop
+    if (test1data(2,1) /= "\d                           ") error stop
+    if (test1data(3,1) /= "5                             ") error stop
+    if (test1data(4,1) /= "1                             ") error stop
+end program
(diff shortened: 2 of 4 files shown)


certik:
If it is a parameter, doesn't it have to have a non-null value?

So this PR is masking a bug?

Pranavchiku:
> So this PR is masking a bug?

Yes, we don't create a compile-time value for reshape.

certik:
I think we have to, since one could use the result if this compile time variable in places like `kind` where we need to know the value at compile time.

So instead of masking it like this, we should rather compute the compile time value for reshape.

Pranavchiku:
> So instead of masking it like this, we should rather compute the compile time value for reshape.

Yes, that might not be difficult to do. I'll do it under https://github.com/lfortran/lfortran/issues/7616. If it is fine, can we get this in, will need to make progress on `fortran-regex`?

certik:
Well, if it is urgent, you can merge it, however I really do not like this fix. Rather, if the parameter is set, it must have a `value` and it should be enforced by ASR verify also. This PR goes in exactly the other direction and it will make it harder to fix it, as now you will also have to fix reshape.

I would say: let's fix the reshape, that's the right fix. If this is an urgent blocker for a code with no workaround, then you can merge it, but can you please do the proper fix and undo this PR in a matter of days? Otherwise it will never get done.

Pranavchiku:
sure, I’ll not merge and first fix reshape correctly.

PR #7621, opened 2025-06-07
Dicts

swamishiju:
WIP


certik:
Looks good to me. Is this ready?

swamishiju:
It is a very bare bones implementation I was thinking i'd add getter, setter then undraft

PR #7624, opened 2025-06-07
fix: get type of polymorphic function caller

jinangshah21:


jinangshah21:
MRE for this works with #7625, if that is ok, I will close this PR.

Issue #7334, opened 2025-05-17
[stdlib:hashmaps] Handle `Deallocation` of allocatable member of Derived type which is also a Derived type


AnshMehta1:
MRE :- 

```.f90
module stdlib_hashmaps

    implicit none
    
    type :: open_map_entry_type
        integer :: slots
    end type open_map_entry_type

    type :: open_map_entry_pool
        type(open_map_entry_type), allocatable :: more_map_entries(:)
    end type open_map_entry_pool

    type :: open_hashmap_type
        type(open_map_entry_pool), pointer :: cache => null()
    end type open_hashmap_type

end module stdlib_hashmaps


program test_open_maps
    use stdlib_hashmaps

    implicit none

    type(open_hashmap_type)   :: map
    type(open_map_entry_pool), pointer :: map_entry_pool_head

    do while(associated(map % cache))
        deallocate( map_entry_pool_head % more_map_entries )
    end do

end program test_open_maps
```

LFortran Error :- 
```console
(lf) aaravmehta@LAPTOP-C2CK4KGC:~/lfort3/lfortran$ lfortran try.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/aaravmehta/lfort3/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/aaravmehta/lfort3/lfortran/src/bin/lfortran.cpp", line 2696, in main
    return main_app(argc, argv);
  File "/home/aaravmehta/lfort3/lfortran/src/bin/lfortran.cpp", line 2624, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/aaravmehta/lfort3/lfortran/src/bin/lfortran.cpp", line 1168, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
[...]

PR #7626, opened 2025-06-07
fix: correctly generate LLVM for deallocating allocatable component of nested derived type (stdlib:hashmaps)

sriganeshres:
fixes: #7334

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1356,7 +1356,6 @@ public:
                     ASR::ttype_t* new_caller_type = ASRUtils::type_get_past_pointer(
                         ASRUtils::type_get_past_allocatable(
                         ASRUtils::expr_type(sm->m_v)));
-                    std::cout << "New Caller Type: " << new_caller_type->type << std::flush;
                     if (ASR::is_a<ASR::StructType_t>(*new_caller_type)) {
                         struct_sym = ASRUtils::symbol_get_past_external(
                             ASR::down_cast<ASR::StructType_t>(new_caller_type)->m_derived_type);
@@ -1385,12 +1384,7 @@ public:
                     builder->CreateStore(dt_val, alloca_tmp);
                     dt = alloca_tmp;
                 }
-                if (!dt->getType()->isPointerTy()) {
-                    std::cerr << "Expected pointer type for GEP, got: ";
-                    dt->getType()->print(llvm::errs());
-                    std::cerr << std::endl;
-                    throw CodeGenError("GEP requires a pointer value to struct.");
-                }
+                LCOMPILERS_ASSERT(dt->getType()->isPointerTy());
                 llvm::Value* dt_1 = builder->CreateGEP(name2dertype[curr_struct], dt, idx_vars);
 #if LLVM_VERSION_MAJOR > 16
                 llvm::Type *dt_1_type = llvm_utils->get_type_from_ttype_t_util(
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -263,13 +263,10 @@ namespace LCompilers {
         } else if (_type->type == ASR::ttypeType::Pointer) {
             ASR::Pointer_t* ptr = ASR::down_cast<ASR::Pointer_t>(_type);
             ASR::ttype_t* ptr_type = ASRUtils::type_get_past_pointer(ptr->m_type);
-            if (ASR::is_a<ASR::StructType_t>(*ptr_type)) {
-                ASR::StructType_t* der = ASR::down_cast<ASR::StructType_t>(ptr_type);
-                ASR::symbol_t* der_sym = ASRUtils::symbol_get_past_external(der->m_derived_type);
-                der_type = ASR::down_cast<ASR::Struct_t>(der_sym);
-            } else {
-                throw CodeGenError("Expected a struct type, but found: " + ASRUtils::type_to_str_fortran(ptr_type));
-            }
+            LCOMPILERS_ASSERT(ASR::is_a<ASR::StructType_t>(*ptr_type));
+            ASR::StructType_t* der = ASR::down_cast<ASR::StructType_t>(ptr_type);
+            ASR::symbol_t* der_sym = ASRUtils::symbol_get_past_external(der->m_derived_type);
+            der_type = ASR::down_cast<ASR::Struct_t>(der_sym);
         } else {
             LCOMPILERS_ASSERT(false);
             return nullptr; // silence a warning


sriganeshres:
@certik @Pranavchiku can you review this

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1366,6 +1379,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    llvm::Value* alloca_tmp = builder->CreateAlloca(name2dertype[curr_struct]);
+                    builder->CreateStore(dt_val, alloca_tmp);
+                    dt = alloca_tmp;
+                }
+                if (!dt->getType()->isPointerTy()) {
+                    std::cerr << "Expected pointer type for GEP, got: ";
+                    dt->getType()->print(llvm::errs());
+                    std::cerr << std::endl;
+                    throw CodeGenError("GEP requires a pointer value to struct.");
+                }
```
certik: When would this happen? I think these should just be asserts.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -262,6 +262,16 @@ namespace LCompilers {
...
             der_type = ASR::down_cast<ASR::Struct_t>(der_sym);
+        } else if (_type->type == ASR::ttypeType::Pointer) {
+            ASR::Pointer_t* ptr = ASR::down_cast<ASR::Pointer_t>(_type);
+            ASR::ttype_t* ptr_type = ASRUtils::type_get_past_pointer(ptr->m_type);
+            if (ASR::is_a<ASR::StructType_t>(*ptr_type)) {
+                ASR::StructType_t* der = ASR::down_cast<ASR::StructType_t>(ptr_type);
+                ASR::symbol_t* der_sym = ASRUtils::symbol_get_past_external(der->m_derived_type);
+                der_type = ASR::down_cast<ASR::Struct_t>(der_sym);
+            } else {
+                throw CodeGenError("Expected a struct type, but found: " + ASRUtils::type_to_str_fortran(ptr_type));
```
certik: When would this happen? I think this should just be an assert.

tests/tests.toml
```diff
@@ -4686,3 +4686,7 @@ llvm = true
 [[test]]
 filename = "../integration_tests/volatile_03.f90"
 llvm = true
+
+[[test]]
+filename = "../integration_tests/hashmap_nested_dealloc_derived_pointer.f90"
+asr = true
```
certik: This PR is not modifying ASR --- what exactly are we testing here? If the above code compiles and runs, that seems enough?

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: Just in case for safety.

src/libasr/codegen/llvm_utils.cpp
sriganeshres: We are getting type past pointer and if it happens to be not a struct then it is wrongly classified as struct Instance.

tests/tests.toml
sriganeshres: i added it just for reference.

sriganeshres: If it is not required we can remove it.

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: Instead of a nasty stack trace error( Seg Fault in LLVM) it is better to stop here. This will not happen but in case we are able to find some thing close to this then it throws a simple error instead.

src/libasr/codegen/llvm_utils.cpp
sriganeshres: Im extracting the struct and making symbol name for that.

tests/tests.toml
certik: I would remove it. We only want to add ASR reference if it is something complex, or something new that doesn't work in LLVM yet. Otherwise we just add regular tests like you did here.

sriganeshres: Yes removed it.

sriganeshres:
@certik i have removed the asr of test.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1352,6 +1352,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         dt = llvm_utils->CreateLoad2(dt_type->getPointerTo(), llvm_utils->create_gep(dt, 1));
                     }
                 }
+                if(caller_type->type == ASR::ttypeType::Pointer) {
+                    ASR::ttype_t* new_caller_type = ASRUtils::type_get_past_pointer(
+                        ASRUtils::type_get_past_allocatable(
+                        ASRUtils::expr_type(sm->m_v)));
+                    std::cout << "New Caller Type: " << new_caller_type->type << std::flush;
```
certik: Remove the debug print:
```suggestion
```

src/libasr/codegen/asr_to_llvm.cpp
certik: Then just use:
```suggestion
                // GEP requires a pointer value to struct
                LCOMPILERS_ASSERT(dt->getType()->isPointerTy());
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -260,6 +260,16 @@ namespace LCompilers {
...
```
certik:
```suggestion
            LCOMPILERS_ASSERT(ASR::is_a<ASR::StructType_t>(*ptr_type));
            ASR::StructType_t* der = ASR::down_cast<ASR::StructType_t>(ptr_type);
            ASR::symbol_t* der_sym = ASRUtils::symbol_get_past_external(der->m_derived_type);
            der_type = ASR::down_cast<ASR::Struct_t>(der_sym);
```

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: Done.

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: Done.

src/libasr/codegen/llvm_utils.cpp
sriganeshres: Wouldn't it better throw a CodeGen simply rather than Assert ans stacktrace.

certik: Can this error happen for a user code? If so, it should be caught in AST->ASR. If not, it should be an assert, to catch internal errors.

sriganeshres:
@certik Done all the suggested chnages.

Issue #7365, opened 2025-05-19
[OPENMP] TASK Construct


adit4443ya:
Reference :- [Examples Document](https://www.openmp.org/wp-content/uploads/openmp-examples-6.0.pdf)
Page :-87, 83

Other links: [Reference Guide](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [Specifications](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [More Examples](https://passlab.github.io/Examples/contents/Title_Page.html).

## Use Cases

- Dynamic Load Balancing/Scheduling: It makes the queue of tasks (Enqueued/Dequeued) under the parallel region and any thread which becomes free picks up the task, so there is no static allocaton of unit of work/ chunk of iterations to specific threads it is done in dynamic manner..
- Ofcourse Task Parallelism.
- Recursive Task Management

### 1
```fortran
!$OMP PARALLEL SHARED(array)
  !$OMP MASTER
  do i = 1, n
    !$OMP TASK PRIVATE(i) FIRSTPRIVATE(array)
    array(i) = array(i) * real(i)
    !$OMP END TASK
  end do
  !$OMP END MASTER
  !$OMP END PARALLEL
```
Description:- This MRE scales each element of an array by its index using tasks. The PARALLEL region creates a thread team, and the MASTER  (SINGLE can also be used in place of MASTER, would do the same work) directive ensures the master thread generates tasks for each array index. Each task is executed by an available thread, dynamically scheduled.

### 2
```fortrn
recursive function fib(n) result(res)
  use omp_lib
  implicit none
  integer :: n, res, left, right
  if (n <= 1) then
    res = n
  else
    !$OMP TASK SHARED(left) FIRSTPRIVATE(n)
    left = fib(n-1)
    !$OMP END TASK
    !$OMP TASK SHARED(right) FIRSTPRIVATE(n)
    right = fib(n-2)
    !$OMP END TASK
    !$OMP TASKWAIT
    res = left + right
  end if
end function fib

! probably in main
!$OMP PARALLEL
  !$OMP SINGLE
  result = fib(n)
  !$OMP END SINGLE
  !$OMP END PARALLEL
```

[...]

PR #7638, opened 2025-06-09
[OPENMP] FEAT: Implement Single/Master Construct

adit4443ya:
Towards #7332 and #7365

This will form base for implementing the `TASK` Construct in subsequent PR, I just thought to keep it separate so as to keep reviewing simpler hence make separate PR

### DIFF
--- a/integration_tests/openmp_48.f90
+++ b/integration_tests/openmp_48.f90
@@ -1,0 +1,14 @@
+program openmp_48
+    use omp_lib
+    implicit none
+    integer :: i=0
+    call omp_set_num_threads(8)
+
+    !$omp parallel reduction(+:i)
+        !$omp single
+            i=i+1 !This will be done by one thread only hence 1 will be added to i
+        !$omp end single
+        i=i+1 !This will be done by all threads hence eventually 8 will be added to i
+    !$omp end parallel 
+    if(i/=9) error stop
+end program openmp_48
--- a/src/libasr/pass/openmp.cpp
+++ b/src/libasr/pass/openmp.cpp
@@ -2017,6 +2017,11 @@ class ParallelRegionVisitor :
                 visit_OMPParallelSections(x);
                 break;
 
+                case ASR::omp_region_typeType::Single:
+                case ASR::omp_region_typeType::Master:
+                visit_OMPSingleThread(x);
+                break;
+
                 default:
                     // for now give error for constructs which we do not support
                     break;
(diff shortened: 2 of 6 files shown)


integration_tests/openmp_48.f90
```diff
@@ -0,0 +1,15 @@
+program openmp_48
+    use omp_lib
+    implicit none
+    call omp_set_num_threads(8)
+    !$omp parallel
+        !$omp single
+            print *,"In single thread: ", omp_get_thread_num()
+            if (omp_get_thread_num() /= 0) error stop
```
certik: This condition seems incorrect?
```
1690/1739 Test #1692: openmp_48 ..........................................***Failed    0.04 sec
 In single thread:            7
ERROR STOP 

Error termination. Backtrace:
#0  0x7f33b9423e59 in ???
#1  0x7f33b9424a71 in ???
#2  0x7f33b942610b in ???
#3  0x557b6d204352 in MAIN__._omp_fn.0
	at /home/runner/work/lfortran/lfortran/integration_tests/openmp_48.f90:8
#4  0x7f33b93cc79e in ???
#5  0x7f33b909caa3 in ???
#6  0x7f33b9129c3b in ???
#7  0xffffffffffffffff in ???
```
Consequently, why is it passing with LFortran?

src/libasr/pass/openmp.cpp
```diff
@@ -2367,6 +2371,39 @@ class ParallelRegionVisitor :
...
+                    this->visit_stmt(*x.m_body[i]);
+                    for (size_t j = 0; j < nested_lowered_body.size(); j++) {
+                        single_body.push_back(nested_lowered_body[j]);
+                    }
+                    nested_lowered_body = body_copy;
+                } else {
+                    stmt_visitor.visit_stmt(*x.m_body[i]);
+                    single_body.push_back(x.m_body[i]);
+                }
+            }
```
Pranavchiku, line 2403: let's write a utility for this

integration_tests/openmp_48.f90
Pranavchiku: Thread number can be anything, the test is wrong.

src/libasr/pass/openmp.cpp
```diff
@@ -2367,6 +2371,39 @@ class ParallelRegionVisitor :
...
             visit_OMPParallel(x);
         }
+
+        void visit_OMPSingle(const ASR::OMPRegion_t &x) {
+            nested_lowered_body = {};
+            Location loc = x.base.base.loc;
+            ASRUtils::ASRBuilder b(al, loc);
+
+            // Restrict execution to thread 0
+            ASR::expr_t* condition = b.Eq(b.i32(0),
```
Pranavchiku, line 2382: execution is not necessarily on thread 0, it can be on any thread as shown in the comment above where thread number is `7`.

integration_tests/openmp_48.f90
adit4443ya: Yes, I apologise for putting this condition in it, It's wrong test i guess. 
In GFortran Any thread from pool [0,num_threads-1] can execute this block,
in this pr's implementation , I did implemented it by fixing the threadID to 0, e.g, 

if(ThreadID==0) then execute this block

That's why LFortran apssing it and gfortran isn't

adit4443ya:
I checked the documentation - [REF](https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-6-0.pdf) page 405

image

it says that choice of , which thread can execute it , is implementation defined. Hence i thought of fixing it to threadID 0

I checked GFortran i didn't found any deterministic way to do it, Hence i Guess GFortran randomly selects from RANGE[0,num_threads-1] and executes that block with that thread

Also when i looked deep into it,
I got another construct `!$OMP MASTER` which is simialr to single except that instead of ANY thread it executes only on ThreadID = 0

That means My current implementation matches with what is required in the `!$OMP MASTER` construct.


So we have two options

- Implement `MASTER` and `SINGLE` differently, (randomly select threadID from available thread)
OR 
- Implement both of them exactly Same as what is done in this PR (Fix threadID to 0) (In which there is no harm to standards i guess as what we eventually need is only one thread executing that block, also it's been said in doc that choice is implementation specific..so)) 

Let me know which one of this sounds good to go, then i would make changes

Pranavchiku:
I think it is fine to have it `0` but we can't test it this way. Just figure out a correct way to test and then this PR is fine.

src/libasr/pass/openmp.cpp
adit4443ya: I think it will need some refactoring in future (unrelated to this PR for now) ( i found some MREs where something is fishy will open up Issue/PR ) ,when i implement `TASK` construct , let me do it next PR where i implement `TASK` construct. Let's leave it for this PR. 
Then i will add utility which can be used in all places.

adit4443ya:
I think it's tests are good now CI is passing.
New Changes includes
- Allow thread with ID==0 to execute the Single block
- Handle `MASTER` and `SINGLE`, as same for now.
- Updated tests

PR #7641, opened 2025-06-09
fix: set return type correctly for `ClassProcedure`

### LINKED ISSUE #7636: Segfault of code generation on a derived type due to use of the `--std=f23` option
A segfault, when compiling the following code, is curiously triggered by the use of the `--std=f23` option.  
$ cat test.f90 
```
module test
      
   type :: SomeType
   contains
      procedure :: method1
      procedure :: method2
   end type SomeType
   
contains
   
   subroutine method1(self,arr)
      class(SomeType), intent(in) :: self
      integer,         intent(in) :: arr(:)
      integer, allocatable :: local(:)
      local = self%method2(arr)
   end subroutine method1
   
   function method2(self,a) result(out)
      class(SomeType), intent(in) :: self
      integer,         intent(in) :: a(:)
      integer                     :: out(size(a))
   end function method2

end module test
```
$ lfortran -c --std=f23 --generate-object-code test.f90

The segfault vanishes if one omits the `--std=f23` option.

kmr-srbh:
fixes #7636

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1741,6 +1741,7 @@ RUN(NAME class_42 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_43 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_44 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_45 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_46 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
 
 RUN(NAME class_procedure_args_01 LABELS gfortran llvm)
 
--- a/integration_tests/class_46.f90
+++ b/integration_tests/class_46.f90
@@ -1,0 +1,39 @@
+module class_46_mod
+
+   type :: SomeType
+   contains
+      procedure :: method1
+      procedure :: method2
+   end type SomeType
+
+contains
+
+   subroutine method1(self, arr)
+      class(SomeType), intent(in) :: self
+      integer, intent(in) :: arr(:)
+      integer, allocatable :: local(:)
+
+      local = self%method2(arr)
+      print *, "local: ", local
+
+      if (.not. all(local == [1, 2, 3])) error stop
+   end subroutine method1
+
+   function method2(self, a) result(out)
+      class(SomeType), intent(in) :: self
+      integer, intent(in) :: a(:)
+      integer :: out(size(a))
+
+      out = a
+   end function method2
+
+end module class_46_mod
+
+program class_46
+   use class_46_mod
+
+   type(SomeType) :: obj
+   integer :: arr(3) = [1, 2, 3]
+
+   call obj%method1(arr)
+end program class_46


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5864,13 +5864,26 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
             type = ASRUtils::EXPR2VAR(func->m_return_var)->m_type;
-            type = node_duplicator.duplicate_ttype(type);
+            if (!v_class_proc->m_is_nopass) {
+                ASR::call_arg_t self_arg;
+                self_arg.loc = func->m_args[0]->base.loc;
+                self_arg.m_value = func->m_args[0];
+                args = {};
+                args.reserve(al, func->n_args);
+                visit_expr_list(m_args, n_args, args);
+                args.push_front(al, self_arg);  // push self arg to fulfill correct number of args in definition
```
kmr-srbh, line 5875: This is required for the function call to `handle_return_type()`. We reset the `args` vector to it's former value below.

kmr-srbh:
> I think that this is fine. What does it have to do with `--realloc-lhs`?

We need it for the assignment `local = self%method2(arr)`.

certik:
Ok, and in the original issue, why does the segfault disappears with removing `--std=f23`?

kmr-srbh:
We enable the `--realloc-lhs` option with the `--std=f23` flag. Using this flag led to segfault in the `pass_array_by_data` pass when replacing the RHS of `local = self%method2(arr)` where the symbol in the `Var` expression pointed outside it's scope (it pointed to variable `a` in `method2` due to type duplication). We could reproduce the segfault without the `--std=f23` flag using just this flag also.

```gdb
(lldb) r test.f90 -c --generate-object-code --realloc-lhs
Process 14672 launched: '/Users/saurabhkumar/miniconda3/envs/lf_release/bin/lfortran' (arm64)
Process 14672 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xc)
    frame #0: 0x00000001018fdb34 lfortran`LCompilers::EditProcedureReplacer::replace_Var(LCompilers::ASR::Var_t*) + 60
lfortran`LCompilers::EditProcedureReplacer::replace_Var:
->  0x1018fdb34 <+60>: ldr    w22, [x21, #0xc]
    0x1018fdb38 <+64>: ldr    x8, [x0, #0x10]
    0x1018fdb3c <+68>: mov    x9, x21
    0x1018fdb40 <+72>: cmp    w22, #0x5
Target 0: (lfortran) stopped.
```

certik:
I see. However the issue https://github.com/lfortran/lfortran/issues/7636 said that the segfault only happens with `--std=f23`, but vanishes without it. This is exactly the opposite to what you wrote, so that is why I am confused if we understand the issue and if it is truly fixed.

kmr-srbh:
> This is exactly the opposite to what you wrote

I am sorry for the confusion. I meant that:
- `--std=f23` brought the `--realloc-lhs` option, so the segfault occurred. 
- The segfault did not occur when neither `--std=f23` nor `--realloc-lhs` was used. 
- The segfault occurred when only `--realloc-lhs` was used.

certik:
Ah, perfect! So I think this should fix it. Thanks!

difference-scheme:
Thanks for the fix, indeed!

PR #7646, opened 2025-06-09
Feat: throw `Semantic Errors` inplace of` LcompilerError` for instantiate functions for intrinsics

parth121101:
This PR tries to throw semantic errors inplace of throwing lcompiler error. This even eliminates the printing of the stack trace.
For example
Consider the following example:
```
program idint_example
    implicit none
    real(4) :: x
    integer :: i

    x = 42.98
    i = idint(x)
end program idint_example
```

Currently we throw lcompiler exception and print the stack trace
```
 lfortran a.f90                                                                
  File "/Users/parth1211/repos/lfortran/src/bin/lfortran.cpp", line 2726
  File "/Users/parth1211/repos/lfortran/src/bin/lfortran.cpp", line 2653
  File "/Users/parth1211/repos/lfortran/src/bin/lfortran.cpp", line 1184
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
  File "/Users/parth1211/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 391
    = asr_to_llvm(asr, diagnostics,
  File "/Users/parth1211/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 11739
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/pass_manager.h", line 318
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/pass_manager.h", line 189
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/intrinsic_function.cpp", line 354
    v.visit_TranslationUnit(unit);
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 39
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_base_visitor.h", line 345
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_base_visitor.h", line 30
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 51
    self().transform_stmts(xx.m_body, xx.n_body);
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 33
    self().visit_stmt(*m_body[i]);
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_base_visitor.h", line 361
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_base_visitor.h", line 54
    case stmtType::Assign: { v.visit_Assign((const Assign_t &)x); return; }
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 267
    current_expr = const_cast<ASR::expr_t**>(&(x.m_value));
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/intrinsic_function.cpp", line 144
    replacer.replace_expr(*current_expr);
  File "/Users/parth1211/repos/lfortran/src/libasr/asr_expr_base_replacer_visitor.h", line 2078
    self().replace_IntrinsicElementalFunction(down_cast<ASR::IntrinsicElementalFunction_t>(x));
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/intrinsic_function.cpp", line 75
    global_scope, arg_types, type, new_args, x->m_overload_id);
  File "/Users/parth1211/repos/lfortran/src/libasr/pass/intrinsic_functions.h", line 3158
    throw LCompilersException("first argument of `idint` must have kind equals to 8");
LCompilersException: first argument of `idint` must have kind equals to 8
```

This PR tries to provide a better error message by elimination of returning stack trace and throwing of lcompiler exception. It is replaced with semantic error. For example:

```
lfortran a.f90                
semantic error: first argument of `idint` must have kind equals to 8
 --> a.f90:1:1 - 8:25
  |
1 |    program idint_example
  |    ^^^^^^^^^^^^^^^^^^^^^...
...
  |
8 |    end program idint_example
  | ...^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

This PR is an initial draft of the implementation and it might require significant refactoring, improvements and addition of error tests.

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1180,19 +1180,18 @@ int compile_src_to_object_file(const std::string &infile,
         return 1;
 #endif
     }
-    try {
         LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
         res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
-        if (res.ok) {
+    if (diagnostics.has_error()) {
+        std::cerr << diagnostics.render(lm, compiler_options);
+        return 5;
+    }
+    if (res.ok) {
         m = std::move(res.result);
     } else {
         LCOMPILERS_ASSERT(diagnostics.has_error())
         return 5;
     }
-    } catch (...) {
-        std::cerr << diagnostics.render(lm, compiler_options);
-        return 5;
-    }   
 
     // LLVM -> Machine code (saves to an object file)
     if (assembly) {
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1180,17 +1180,21 @@ int compile_src_to_object_file(const std::string &infile,
         return 1;
 #endif
     }
+    try {
         LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
         res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
-    if (diagnostics.has_error()) {
         std::cerr << diagnostics.render(lm, compiler_options);
-        return 5;
-    }
-    if (res.ok) {
-        m = std::move(res.result);
-    } else {
-        LCOMPILERS_ASSERT(diagnostics.has_error())
-        return 5;
+        if (res.ok) {
+            m = std::move(res.result);
+        } else {
+            LCOMPILERS_ASSERT(diagnostics.has_error())
+            return 5;
+        }
+    } catch (const LCompilers::RuntimeException &e) {
+        if (diagnostics.has_error()) {
+            std::cerr << diagnostics.render(lm, compiler_options);
+            return 5;
+        }
     }
 
     // LLVM -> Machine code (saves to an object file)
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -1180,15 +1180,19 @@ int compile_src_to_object_file(const std::string &infile,
         return 1;
 #endif
     }
-    LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
+    try {
+        LCompilers::Result<std::unique_ptr<LCompilers::LLVMModule>>
         res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
-    std::cerr << diagnostics.render(lm, compiler_options);
-    if (res.ok) {
+        if (res.ok) {
         m = std::move(res.result);
     } else {
         LCOMPILERS_ASSERT(diagnostics.has_error())
         return 5;
     }
+    } catch (...) {
+        std::cerr << diagnostics.render(lm, compiler_options);
+        return 5;
+    }   
 
     // LLVM -> Machine code (saves to an object file)
     if (assembly) {
--- a/src/libasr/exception.h
+++ b/src/libasr/exception.h
@@ -165,6 +165,10 @@ static inline T TRY(Result<T> result) {
     }
 }
 
+class RuntimeException
+{
+};
+
 
 } // namespace LCompilers
 
--- a/src/libasr/pass/intrinsic_function.cpp
+++ b/src/libasr/pass/intrinsic_function.cpp
@@ -80,10 +80,10 @@ class ReplaceIntrinsicFunctions: public ASR::BaseExprReplacer<ReplaceIntrinsicFu
         } catch (const LCompilersException& e) {
             global_diagnostics->add(diag::Diagnostic(
                 std::string(e.what()),
-                diag::Level::Error, diag::Stage::Semantic, {
+                diag::Level::Error, diag::Stage::Runtime, {
                 diag::Label("", {x->base.base.loc})
             }));
-            return;
+            throw RuntimeException();
         }
     }
 
--- a/src/libasr/pass/intrinsic_function.cpp
+++ b/src/libasr/pass/intrinsic_function.cpp
@@ -363,9 +363,27 @@ void pass_replace_intrinsic_function(Allocator &al, ASR::TranslationUnit_t &unit
         throw LCompilers::LFortran::SemanticAbort();
     }
     ReplaceFunctionCallReturningArrayVisitor u(al, func2intrinsicid);
-    u.visit_TranslationUnit(unit);
+    try {
+        u.visit_TranslationUnit(unit);
+    } catch (const LCompilersException& e) {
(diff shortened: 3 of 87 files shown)


gxyd:
The current error at the CI occurs because of the below failure:
```console
[ 64%] Building CXX object CMakeFiles/asr.dir/pass/intrinsic_function.cpp.o
/home/runner/work/lfortran/lfortran/build/libasr/pass/intrinsic_function.cpp:10:10: fatal error: lfortran/semantics/semantic_exception.h: No such file or directory
   10 | #include <lfortran/semantics/semantic_exception.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [CMakeFiles/asr.dir/build.make:622: CMakeFiles/asr.dir/pass/intrinsic_function.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:111: CMakeFiles/asr.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

which you can reproduce locally using `shell` (using `conda install shell=0.2.0`), using `shell ci/test_cpp_version.sh` at: https://github.com/lfortran/lfortran/blob/cfe8af374f693fe0624071078d64d3f49d4b8881/ci/test_cpp_version.sh#L21-L28, which tests that libasr can be independently used, but the changes in your PR make it depend on other non-libasr related files, and hence the CI fails.

Pranavchiku (commented):
As far as I remember we are doing this to throw semantic error in instantiate function. I discussed with @HarshitaKalani, we don't need to throw error in instantiate function, it can be done in `create_*` functions which already have diagnostic information. So we may not need these change. 


However, let's keep this PR in draft mode if the above way does not fix it, then we can use this approach. Thanks for this and I'm sorry for guiding you astray!

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -71,9 +72,19 @@ class ReplaceIntrinsicFunctions: public ASR::BaseExprReplacer<ReplaceIntrinsicFu
...
             global_scope, arg_types, type, new_args, x->m_overload_id);
-        *current_expr = current_expr_;
+            *current_expr = current_expr_;
+        } catch (const LCompilersException& e) {
+            global_diagnostics->add(diag::Diagnostic(
+                std::string(e.what()),
+                diag::Level::Error, diag::Stage::Semantic, {
+                diag::Label("", {x->base.base.loc})
+            }));
+            return;
```
Pranavchiku: I see, this is good.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -71,9 +72,19 @@ class ReplaceIntrinsicFunctions: public ASR::BaseExprReplacer<ReplaceIntrinsicFu
...
+        ASR::expr_t* current_expr_ = nullptr;
+         try {
+            current_expr_ = instantiate_function(al, x->base.base.loc,
             global_scope, arg_types, type, new_args, x->m_overload_id);
-        *current_expr = current_expr_;
+            *current_expr = current_expr_;
+        } catch (const LCompilersException& e) {
+            global_diagnostics->add(diag::Diagnostic(
+                std::string(e.what()),
+                diag::Level::Error, diag::Stage::Semantic, {
```
Pranavchiku:
```suggestion
                diag::Level::Error, diag::Stage::XX, {
```

Do you think `Semantic` is good here? or we need to have some other name?

parth121101:
> As far as I remember we are doing this to throw semantic error in instantiate function. I discussed with @HarshitaKalani, we don't need to throw error in instantiate function, it can be done in `create_*` functions which already have diagnostic information. So we may not need these change.
> 
> However, let's keep this PR in draft mode if the above way does not fix it, then we can use this approach. Thanks for this and I'm sorry for guiding you astray!
> 
> Edit: Also, @parth121101 can you identify if we throw any `LCompilersException` in `instantiate_*` functions, if yes, then we may need this under different `Label`.

So currently we throw Lcompiler exception for 13 different instantiate function.

src/libasr/pass/intrinsic_function.cpp
parth121101: We can change these if required.

src/bin/lfortran.cpp
```diff
@@ -1180,14 +1180,21 @@ int compile_src_to_object_file(const std::string &infile,
...
+            m = std::move(res.result);
+        } else {
+            LCOMPILERS_ASSERT(diagnostics.has_error())
+            return 5;
+        }
+    } catch (const LCompilers::RuntimeException &e) {
+        if (diagnostics.has_error()) {
+            std::cerr << diagnostics.render(lm, compiler_options);
+            return 5;
+        }
```
certik, line 1197: Don't catch nor use exceptions. Instead use the technique from this PR: https://github.com/lfortran/lfortran/pull/7323, return the error state in the Result.

src/libasr/exception.h
```diff
@@ -165,6 +165,10 @@ static inline T TRY(Result<T> result) {
     }
 }
 
+class RuntimeException
+{
+};
```
certik, line 170: Don't add a new exception.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -71,9 +72,19 @@ class ReplaceIntrinsicFunctions: public ASR::BaseExprReplacer<ReplaceIntrinsicFu
...
-        *current_expr = current_expr_;
+            *current_expr = current_expr_;
+        } catch (const LCompilersException& e) {
+            global_diagnostics->add(diag::Diagnostic(
+                std::string(e.what()),
+                diag::Level::Error, diag::Stage::Runtime, {
+                diag::Label("", {x->base.base.loc})
+            }));
+            throw RuntimeException();
+        }
```
certik, line 87: Don't throw exceptions.

The fundamental issue to decide here is if these are errors in user Fortran code. If so, ideally they should be caught in AST->ASR. We can use some helper functions here, but the error reporting logic should be in AST->ASR.

If they are not, then they are ICE (internal compiler error) and giving a stacktrace is perfect then. So just use an ASSERT.

parth121101: Sure, so lets keep this. Will work on error reporting for such cases in AST->ASR.

parth121101:
Closing this PR

PR #7653, opened 2025-06-10
More list stuff

swamishiju:


swamishiju:
`lp_list_test_01_.f90` is an almost full port of `test_list_01.py` I removed the negative indexing test because I'm pretty sure something in LP handles that

Issue #7369, opened 2025-05-20
LFortran with LSP leaves around empty lfortran-language-server.log


certik:
```console
fortran/examples(main)$ ls
asr.f90  chat_01.f90  cmdline.f90  expr2.f90  project1  project2  project3
lfortran/examples(main)$ lfortran expr2.f90
25
lfortran/examples(main)$ ls
asr.f90      cmdline.f90  expr2.out    lfortran-language-server.log  project2
chat_01.f90  expr2.f90    expr2.tmp.o  project1                      project3
```

The LSP log should only be emitted when LSP is used, and probably also only when a command line option is present.

gxyd:
> The LSP log should only be emitted when LSP is used

Is there a very specific flag in LFortran which indicates we want to use LSP specific options?

> when a command line option is present

Is there already a command line option to indicate that?

certik:
Right now LSP should only be used when `lfortran server` is used.

We can add a specific option under `lfortran server` to emit or not emit the log file.

TODO:

* Figure out why is LSP code being run in the first place by default ---- no LSP should run by default
* Remove any LSP calls, which should avoid creating the log file

Optionally we can make it configurable, but that's minor. We can keep the log file in `lfortran server` mode for now.

gxyd:
> * Figure out why is LSP code being run in the first place by default ---- no LSP should run by default

I built LFortran in debug mode with `-DWITH_LSP=no` (as it currently is in our `build1.sh`), and on execution of program with LFortran, I don't see any generation of `lfortran-language-server.log` file.

It's only when I use `-DWITH_LSP=yes`, I see generation of `lfortran-language-server.log` file, which I think you are suggesting shouldn't happen either.

gxyd:
So, I'm not sure why this issue blocks us from making a release of LFortran, if the log-file `lfortran-language-server.log` isn't created unless `-DWITH_LSP=yes` is explicitly specified.

certik:
The reason this issue is a blocker is because the conda package builds LFortran with `-DWITH_LSP=yes` and thus it creates this log file for EVERY compilation. Which is insane and must be fixed in the next release and we must release soon.

gxyd:
> The reason this issue is a blocker is because the conda package builds LFortran with `-DWITH_LSP=yes` and thus it creates this log file for EVERY compilation. Which is insane and must be fixed in the next release and we must release soon.

Understood. 

[...]

PR #7658, opened 2025-06-10
LSP: add an option to enable logging (enabled by default)

gxyd:
## Description

I'm not exactly sure how to run the lfortran language server locally yet, I'm trying to find the exact commands to test it.

Towards: https://github.com/lfortran/lfortran/issues/7369

### DIFF
--- a/src/bin/language_server_interface.cpp
+++ b/src/bin/language_server_interface.cpp
@@ -274,6 +274,12 @@ namespace LCompilers::LLanguageServer::Interface {
             "Number of threads that will handle sub-tasks of requests."
         )->capture_default_str();
 
+        opts.enableLogging = true;
+        server->add_flag(
+            "--enable-logging", opts.enableLogging,
+            "Enable or disable logging to a file."
+        )->capture_default_str();
+
         opts.configSection = "LFortran";
         server->add_option(
             "--config-section", opts.configSection,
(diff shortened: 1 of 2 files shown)


certik:
`lfortran server` should run it. You can run tests via Python, see our CI how to run them. You can also test it in VSCode.

certik:
Make this off by default and check if the log file is still created all the time.

If not, then at least it's a workaround, until we figure out why any LSP code is run at all.

gxyd:
With the changes in this PR, no `lfortran-language-server.log` is created anymore. I checked the LSP commands from CI.yml and they seem to pass locally as well.

Even with all of that, I think this needs a little more work with regards to testing, @certik can you help with making a request to this `lfortran server` locally? Or could you point me to where can I specifically read more about it? E.g. in file `tests/server/tests/conftest.py` I see in `client` function the definition of `config` is defined as:
```json
        "LFortran": {
            "openIssueReporterOnError": False,
            "maxNumberOfProblems": 100,
            "trace": {
                "server": "verbose",
            },
            "compiler": {
                "path": "lfortran",
                "flags": [],
            },
            "log": {
                "path": server_log_path,
                "level": "all",
                "prettyPrint": False,
            },
            "indentSize": 4,
            # NOTE: Regarding timing-out and retrying requests, a timeout of 0
            # means do not time-out (or retry):
            "timeoutMs": 0,
            "retry": {
                "maxAttempts": 3,
                "minSleepTimeMs": 10,
                "maxSleepTimeMs": 300,
            },
            "telemetry": {
                "enabled": True,
                "frequencyMs": 1000,
            },
        }

```
I'm not sure what changes are expected in it to reflect the same (if any)

src/bin/language_server_interface.cpp
```diff
@@ -274,6 +274,12 @@ namespace LCompilers::LLanguageServer::Interface {
             "Number of threads that will handle sub-tasks of requests."
         )->capture_default_str();
 
+        opts.enableLogging = false;
+        server->add_flag(
+            "--enable-logging", opts.enableLogging,
+            "Enable or disable logging to a file."
```
certik, line 280: This does not disable logging, only enables, correct?

certik: Fixed in https://github.com/lfortran/lfortran/pull/7734.

PR #7660, opened 2025-06-10
List stuff

swamishiju:
Added the clear intrinsic and ported `test_list_06.py`


integration_tests/lp_list_test_06_.f90
```diff
@@ -0,0 +1,109 @@
+module list_tests
```
certik, line 1: Use the test name in the name, to avoid collisions:
```suggestion
module list_tests_06_mod
```

swamishiju: I wasn't aware modules worked like that across files 😅  will do

PR #7663, opened 2025-06-10
fix: make is_cstruct true for class variables

### LINKED ISSUE #7644: Handle class variable passed into subroutine call as nested subroutine var
MRE:
```fortran
module toml_module
    implicit none

    type :: toml_lexer
        integer :: dummy = 0
    end type toml_lexer

contains

    subroutine temp2(lex)
        class(toml_lexer), intent(in) :: lex
        print *, "In temp2: lexer%dummy = ", lex%dummy
    end subroutine temp2

end module toml_module

program test_lexer
    use toml_module
    implicit none

    type(toml_lexer) :: my_lexer

    my_lexer%dummy = 123
    call wrapper(my_lexer)

contains

    subroutine wrapper(lexer)
        class(toml_lexer) :: lexer
        call temp()

        contains

            subroutine temp()
                call temp2(lexer)
            end subroutine temp

    end subroutine wrapper

end program test_lexer

```

LFortran Error:
```console
void** @__lcompilers_created__nested_context__wrapper_lexer
 %toml_lexer_polymorphic*  call void @__module_toml_module_temp2(void** @__lcompilers_created__nested_context__wrapper_lexer)
```

jinangshah21:
Fixes #7644

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4805,7 +4805,7 @@ public:
                 current_scope = parent_scope;
             }
             // this is class variable declaration
-            type = ASRUtils::TYPE(ASRUtils::make_StructType_t_util(al, loc, v, true));
+            type = ASRUtils::TYPE(ASRUtils::make_StructType_t_util(al, loc, v, false));
             type = ASRUtils::make_Array_t_util(
                 al, loc, type, dims.p, dims.size(), abi, is_argument);
             if (is_pointer) {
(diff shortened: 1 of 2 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4805,7 +4805,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 current_scope = parent_scope;
             }
             // this is class variable declaration
-            type = ASRUtils::TYPE(ASRUtils::make_StructType_t_util(al, loc, v, false));
+            type = ASRUtils::TYPE(ASRUtils::make_StructType_t_util(al, loc, v, true));
```
kmr-srbh: I feel it is not right to do this. We set the `is_cstruct` field to `false` to denote a `class` variable declaration and `true` for `type` variables. Why do we need this?

jinangshah21: Yes, it is incorrect, I am working on this.

PR #7668, opened 2025-06-10
fix: correct the mold condition in allocate for checking if length in dim is not null

### LINKED ISSUE #7665: Code generation fails for `allocate` statement with `mold` clause
$ cat test.f90 
```
module test

contains
   
   subroutine molded_allocation(arr)
      real(8), intent(in)  :: arr(:,:)
      real(8), allocatable :: local_arr(:,:)
      allocate(local_arr, mold = arr)
   end subroutine molded_allocation

end module test

```

$ lfortran -c --generate-object-code test.f90 
```
code generation error: Internal error: x is nullptr


```

HarshilShah1804:
Fixes #7665


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1589,7 +1589,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     ASR::ttype_t* a_type = ASRUtils::type_get_past_allocatable(ASRUtils::expr_type(alloc_args_vec[i].m_a));
                     if ( ASRUtils::check_equal_type(mold_type, a_type) ) {
                         if (ASRUtils::is_array(mold_type)) {
-                            if (ASR::is_a<ASR::Array_t>(*mold_type)) {
+                            if (ASR::is_a<ASR::Array_t>(*mold_type) && ASR::down_cast<ASR::Array_t>(mold_type)->m_dims[0].m_length != nullptr) {
```
HarshilShah1804, line 1592: I have a doubt here. Is it possible if the `dim[0]` has non-null length attribute and some other dim has null. In that case, we will have to check for all till `n_dims`

gxyd: I don't think Fortran allows declarations like:
```f90
    real, dimension(2, :) :: x
```

(neither with `allocatable` attribute present), in subroutines we can have assumed-size array where one of the dimensnion length is known while the other is unknown (e.g. `real :: x(2, *)`), but such arrays aren't allowed as values for `mold` specifier, so I think this condition is fine.

integration_tests/allocate_21.f90
```diff
@@ -0,0 +1,27 @@
...
+      allocate(local_arr, mold = arr)
+
+      local_arr = arr
+      if (size(local_arr, 1) /= size(arr, 1) .or. &
+          size(local_arr, 2) /= size(arr, 2)) error stop
+   end subroutine molded_allocation
+
+end module allocate_module
+
+program main
```
gxyd: Can we use a more specific program name i.e. `allocate_21`?

integration_tests/allocate_21.f90
```diff
@@ -0,0 +1,27 @@
...
+   use test
```
gxyd: We need to use the correct module name here.

integration_tests/allocate_21.f90
```diff
@@ -0,0 +1,27 @@
+module allocate_module
```
gxyd: I would prefer mangling the module with use of the test file name itself, e.g. `mod_allocate_21` or something similar

integration_tests/allocate_21.f90
```diff
@@ -0,0 +1,27 @@
...
+contains
+
+   subroutine molded_allocation(arr)
+      real(8), intent(in)  :: arr(:,:)
+      real(8), allocatable :: local_arr(:,:)
+
+      allocate(local_arr, mold = arr)
+
+      local_arr = arr
```
gxyd: I don't think that this assignment is necessary to actually test this. I would remove it

gxyd (commented):
The change itself in this PR looks fine to me, but maybe needs changes in the test added. Once those are fixed, the CI should probably pass as well.

src/lfortran/semantics/ast_body_visitor.cpp
HarshilShah1804: Thanks @gxyd for clarification.

integration_tests/allocate_21.f90
HarshilShah1804: Yes, I'll change that.

integration_tests/allocate_21.f90
HarshilShah1804: Yes, that's the reason the CI is failing. I will change that.

integration_tests/allocate_21.f90
HarshilShah1804: Ok. I'll change that to `mod_allocate_21`.

PR #7661, opened 2025-06-10
fix: handle load of class and struct member in `deallocate`

jinangshah21:
Fixes #7648

PR #7669, opened 2025-06-10
fic: handle nested SIM in deallocate

jinangshah21:
Apply Suggestion on https://github.com/lfortran/lfortran/pull/7661#discussion_r2138475042


jinangshah21:
It means `StructInstanceMember`.

Issue #7640, opened 2025-06-09
Segfault of code generation on a derived type that makes use of an embedded polymorphic object


difference-scheme:
The present issue might have the same cause as #7636, although there's no `--std=f23` compiler option involved in the present case:
$ cat test.f90 
```
module test
   
   type, abstract :: AbsType
   contains
      procedure(method2), deferred :: method2
   end type AbsType

   abstract interface
      function method2(self,arr) result(a)
         import
         class(AbsType), intent(in) :: self
         integer,        intent(in) :: arr(:)
         integer                    :: a(size(arr))
      end function method2
   end interface

   type :: SomeType
      integer,        allocatable :: arr(:)
      class(AbsType), allocatable :: obj
   contains
      procedure :: method1
   end type SomeType

contains

   subroutine method1(self)
      class(SomeType), intent(inout) :: self
      self%arr = self%obj%method2(self%arr)
   end subroutine method1

end module test
```

$ lfortran -c --generate-object-code test.f90
`Segmentation fault (core dumped)`

kmr-srbh:
Thanks for this bug report @difference-scheme! This compiles with #7641, but we have an issue:

```fortran
module test
   
   type, abstract :: AbsType
   contains
      procedure(method2), deferred :: method2
   end type AbsType

   abstract interface
      function method2(self,arr) result(a)
         import
         class(AbsType), intent(in) :: self
         integer,        intent(in) :: arr(:)
         integer                    :: a(size(arr))
      end function method2
   end interface

   type :: SomeType
      integer,        allocatable :: arr(:)
      class(AbsType), allocatable :: obj
   contains
      procedure :: method1
   end type SomeType

   type, extends(AbsType) :: MyType
    contains
        procedure :: method2 => my_method2
    end type MyType

contains

   subroutine method1(self)
      class(SomeType), intent(inout) :: self
      self%arr = self%obj%method2(self%arr)
   end subroutine method1

   function my_method2(self, arr) result(a)
        class(MyType), intent(in) :: self
        integer, intent(in) :: arr(:)
        integer :: a(size(arr))
        integer :: i
        do i = 1, size(arr)
            a(i) = arr(i) * 2
        end do
    end function my_method2

end module test

program test_main
    use test
    implicit none

    class(SomeType), allocatable :: s
    
    allocate(s)
    allocate(s%arr(3))
    allocate(MyType :: s%obj)
    s%arr = [1, 2, 3]

    call s%method1()
[...]

PR #7670, opened 2025-06-10
fix: load `allocatable` array member of `class` when passes to an assumed-shape dummy array

kmr-srbh:
Towards #7640


integration_tests/class_47.f90
```diff
@@ -0,0 +1,65 @@
...
+   end interface
+
+   type, extends(AbsType) :: MyType
+   contains
+      procedure :: method2 => my_method2
+   end type MyType
+
+   type :: SomeType
+      integer,        allocatable :: arr(:)
+      class(MyType), allocatable :: obj
```
kmr-srbh, line 24: This particular line is the main reason why this code works correctly and the one in https://github.com/lfortran/lfortran/issues/7640#issuecomment-2956212678 does not. Here I have changed the type of `obj` to a concrete type implementation of `AbsType` - `MyType`. 

In the former code, the issue occurs below at line 34 when we try to find the struct to call `method2`.
```fortran
self%arr = self%obj%method2(self%arr)
```
We currently do this using the compile-time type, so the struct for `self%obj` becomes `AbsType` which is abstract. This eventually leads to no function call and we return. This PR fixes a related bug. To fix #7640 we will need access to the current runtime type for `self%obj` where `obj` is declared using the abstract type as
```fortran
```
and allocated to the correct concrete type implementation somewhere.

PR #7684, opened 2025-06-12
Parser changes

swamishiju:
- Made parser changes for better tuple, dict syntax 
- Introduced tuple type

### DIFF
--- a/integration_tests/lp_tuple_test_01_.f90
+++ b/integration_tests/lp_tuple_test_01_.f90
@@ -1,0 +1,40 @@
+module tuple_test
+   implicit none
+
+contains
+
+   subroutine f()
+      type(_lfortran_tuple(integer, real, character(len=:), logical)) :: t1, t2
+      type(_lfortran_tuple(integer, integer)) :: t3
+      type(_lfortran_tuple(real, real)) :: t4
+
+      integer :: a, b
+      real :: af, bf
+
+      t1 = _lfortran_tuple_constant(1, 2.0, "3", .true.)
+      t2 = _lfortran_tuple_constant(2, 3.0, "3", .false.)
+
+      if (_lfortran_get_item(t2, 0) - _lfortran_get_item(t1, 0) /= 1) error stop
+      if (abs(_lfortran_get_item(t2, 1) - _lfortran_get_item(t1, 1) - 1.0) > 1e-12) error stop
+      if (_lfortran_get_item(t1, 2) /= _lfortran_get_item(t2, 2)) error stop
+      if (.not. (_lfortran_get_item(t1, 3) .or. _lfortran_get_item(t2, 3))) error stop
+
+      t3 = _lfortran_tuple_constant(1, 2)
+      t4 = _lfortran_tuple_constant(1.0, 2.0)
+
+      a = _lfortran_get_item(t3, 0)
+      b = _lfortran_get_item(t3, 1)
+      af = _lfortran_get_item(t4, 0)
+      bf = _lfortran_get_item(t4, 1)
+
+      if (af /= real(a)) error stop
+      if (bf /= real(b)) error stop
+   end subroutine
+
+end module
+
+program main
+   use tuple_test
+   implicit none
+   call f()
+end program
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4471,6 +4471,31 @@ public:
         AST::decl_attribute_t* decl_attribute, bool is_pointer,
         bool is_allocatable, Vec<ASR::dimension_t>& dims,
         ASR::symbol_t *&type_declaration, ASR::abiType abi, bool is_argument=false, bool is_dimension_star=false) {
+
+        if (AST::is_a<AST::AttrTypeList_t>(*decl_attribute)) {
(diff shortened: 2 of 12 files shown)


integration_tests/lp_tuple_test_01_.f90
```diff
@@ -0,0 +1,40 @@
+module tuple_test
+   implicit none
+
+contains
+
+   subroutine f()
+      type(_lfortran_tuple(integer, real, character(len=:), logical)) :: t1, t2
+      type(_lfortran_tuple(integer, integer)) :: t3
```
certik, line 8: Also test things like:
```suggestion
      type(_lfortran_tuple(integer(4), integer(8))) :: t3
```

certik: Also, in the way this is implemented in the parser, I think the following is also now supported, so we should add tests for it:
```
      _lfortran_tuple(integer(4), integer(8)) :: t3
```

The other thing to consider is to always give a warning that `_lfortran_tuple` is an LFortran extension, and we would need to add some compiler option to disable the warning. Let's just think about how this can work, don't do it right now. I think the `_lfortran_tuple` is sufficiently unique and descriptive that people will not use it by mistake without realizing it's an LFortran extension. However, later on we want to rename `_lfortran_tuple` to just `tuple`, and then we want to communicate that it's an extension somehow.

certik:
From the LLVM reference test update, it seems you are using LLVM >= 16, which introduced the `*ptr` thing. Instead, use LLVM 11.

certik:
Given the complexity of this PR, can you split it into two or more PRs. Keep this one open, but send a new PR that only does the parser changes, but keeps everything the same (I think you need to update AST->ASR to use the new AST changes, but don't do anything else). Let's carefully review the parser changes and get them in.

Then send a PR that introduces the new dict and tuple features and tests and backend changes.

integration_tests/lp_tuple_test_01_.f90
swamishiju: > Also, in the way this is implemented in the parser, I think the following is also now supported, so we should add tests for it:

Yes that was one of the good things to simply work with this approach

swamishiju:
> From the LLVM reference test update, it seems you are using LLVM >= 16, which introduced the `*ptr` thing. Instead, use LLVM 11.

Oh that makes sense I was using the wrong conda env

swamishiju:
> Given the complexity of this PR, can you split it into two or more PRs. Keep this one open, but send a new PR that only does the parser changes, but keeps everything the same (I think you need to update AST->ASR to use the new AST changes, but don't do anything else). Let's carefully review the parser changes and get them in.
> 
> Then send a PR that introduces the new dict and tuple features and tests and backend changes.

Will do

swamishiju:
Ok ye after the llvm changes it seems to be fine, I'll open another PR for parser changes specifically link it here and send a dm

swamishiju:
@certik https://github.com/lfortran/lfortran/pull/7698

src/libasr/codegen/llvm_utils.h
```diff
@@ -606,6 +606,12 @@ namespace LCompilers {
             llvm::Value* read_item(llvm::Value* llvm_tuple, size_t pos,
                                    bool get_pointer=false);
 
+            llvm::Value* read_item2(llvm::Type* el_type, llvm::Value* llvm_tuple, llvm::Value* pos,
+                                   bool get_pointer=false);
+
+            llvm::Value* read_item2(llvm::Type* el_type, llvm::Value* llvm_tuple, size_t pos,
+                                   bool get_pointer=false);
```
certik, line 613: I don't like these overloaded methods, it's not clear which ones should be used. I would find a good descriptive name for each. Grok suggested:

GetTupleElementByValue (dynamic position, untyped)
GetTupleElementByIndex (static position, untyped)
GetTypedTupleElementByValue (dynamic position, typed)
GetTypedTupleElementByIndex (static position, typed)

But I am not sure I like it. But you get the idea. Let's do this in a subsequent PR.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7410,7 +7403,22 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 
                 throw SemanticAbort();
             }
-            return ASR::make_ListItem_t(al, x.base.base.loc, args[0], args[1],
+
+            ASR::expr_t *index = ASRUtils::EXPR(tmp);
+            ASR::expr_t* val = ASRUtils::expr_value(index);
+            if (val && ASR::is_a<ASR::IntegerConstant_t>(*val)) {
+                if (ASR::down_cast<ASR::IntegerConstant_t>(val)->m_n < 0) {
+                    // Replace `x[-1]` to `x[len(x)+(-1)]`
```
certik, line 7411: Let's add some tests for the -1 index also.

swamishiju: Shouldn't we handle it the back-end though? Run-time -ve indexing would be nice unless there was sm problem i'm not aware of. It would slightly increase time for each look up but other than that?

src/libasr/codegen/llvm_utils.h
swamishiju: Those were just to internally change things like `CreateLoad` -> `CreateLoad2` it was just smtg quick I could slowly incorporate without tons of refactoring, will do.

PR #7685, opened 2025-06-12
fix : Create Load Statement for StructInstanceMember of Pointer type

AnshMehta1:


gxyd (approved):
Looks good to me, thanks @AnshMehta1

integration_tests/derived_types_61.f90
```diff
@@ -0,0 +1,35 @@
+module stdlib_hashmaps_derived_types_61
+
+    implicit none
+
+    type :: open_map_entry_type
+        integer  :: slot = 0
```
gxyd, line 6: Actually, do you mind if you can change this value to some non-zero value and test the value below with that?

I've often noticed that even non-initialized values are mostly set to zero in Fortran and then some of the times they attain non-zero values.

certik: Yes, definitely. An uninitialized variable's value is usually 0, since that is the most common value in memory in practice, so the chance is high that if you set it to 0 but somehow it is actually not initialized, the test will pass anyway. So if you set this to something specific, say 13, then the random chance if it being 13 is very very low.

AnshMehta1:
> The fix 0 -> 13 can be done in another PR.

Yes, will bring this change in my next PR.

Issue #7683, opened 2025-06-11
cmake LIBUNWIND warning


certik:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1266659&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=ae18cb70-17c3-5a15-782b-9596d963d812

```
+ cmake -DCMAKE_AR=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-ar -DCMAKE_CXX_COMPILER_AR=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-gcc-ar -DCMAKE_C_COMPILER_AR=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-gcc-ar -DCMAKE_RANLIB=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-ranlib -DCMAKE_CXX_COMPILER_RANLIB=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-gcc-ranlib -DCMAKE_C_COMPILER_RANLIB=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-gcc-ranlib -DCMAKE_LINKER=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86_64-conda-linux-gnu-ld -DCMAKE_STRIP=/home/conda/feedstock_root/build_artifacts/lfortran_1749671241190/_build_env/bin/x86...
-- The C compiler identification is GNU 13.3.0
-- The CXX compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ - skipped
-- Detecting CXX compile features
[...]

PR #7687, opened 2025-06-12
fix CMake warning for libunwind package name

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/7683
CMake is case sensitive, hence we should use the same case in both cmake/Findlibunwind.cmake and CMakeLists.txt

to be consistent with other packages like BFD, LINKH, MACHO etc. it's better to update it in CMakeLists.txt to use `LIBUNWIND` (uppercase) instead of `libunwind` (lowercase)

Once the CI checks complete, I'll check the logs in *Test LLVM x* (where `x > 11`) to see if we get any warning or not.

### DIFF
--- a/cmake/FindLIBUNWIND.cmake
+++ b/cmake/FindLIBUNWIND.cmake
@@ -1,0 +1,11 @@
+find_path(LIBUNWIND_INCLUDE_DIR libunwind.h)
+find_library(LIBUNWIND_LIBRARY unwind)
+
+include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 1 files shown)


gxyd:
I also should rename `Findlibunwind.cmake` to `FindLIBUNWIND.cmake`, just the way we've `FindLINKH.cmake`, `FindBDF.cmake` etc.

gxyd:
I don't have access to ubuntu machine, so I've to use the CI to see if things actually work or not.

gxyd:
I verified at the CI run: https://github.com/lfortran/lfortran/actions/runs/15610564955/job/43970262026?pr=7687, which runs it with `WITH_LIBUNWIND=yes`, and the CMake doesn't give any such warning, hence this PR fixes the warning.

I'll auto-merge this PR.

PR #7690, opened 2025-06-12
enh: added constant folding for speeding loops

sriganeshres:
This is to speed up loops.


certik:
This adds complexity to our backend, and LLVM optimizes this anyway in Release mode. Also very rarely is the condition simply false. It can end up being false at runtime, but one has to do a full analysis, which LLVM does.

Also, if we are going to add such optimizations, then they need to happen at the ASR level as optional passes. However, let's do that after beta.

sriganeshres:
Yeah I just had this in one of my stashes. Thought i would make a PR and leave it there.

Issue #7679, opened 2025-06-11
Handle default `StructConstructor` call when customized constructors are defined


jinangshah21:
MRE:
```fortran
module datetime_mod
    implicit none
    type :: toml_val 
    end type
    type :: toml_datetime
        character(len=:), allocatable :: date
    end type toml_datetime

    interface toml_datetime
        module procedure :: toml_datetime_ctor
    end interface

contains
    function toml_datetime_ctor(date_str, time_str) result(dt)
        character(len=*), intent(in) :: date_str, time_str
        type(toml_datetime) :: dt

        dt%date = date_str
    end function toml_datetime_ctor
end module datetime_mod

program main
    use datetime_mod
    implicit none
    type(toml_datetime) :: ts1
    character(len=10) :: tmp = "2025-06-11"
    ts1 = toml_datetime(tmp)
    print *, "Date: ", ts1%date
    ! print *, "Time: ", ts1%time
end program main
```

LFortran Error:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
semantic error: Function 'toml_datetime' not found (not user defined nor intrinsic)
  --> b.f90:31:11
   |
31 |     ts1 = toml_datetime(tmp)
   |           ^^^^^^^^^^^^^^^^^^ 
```

PR #7708, opened 2025-06-13
fix: try if FuncCallOrArray matches struct constructor call after interface GenericProcedure

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/7679

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -9090,7 +9090,7 @@ public:
             // for a GenericProcedure and a derived type with the same name,
             // we prefer GenericProcedure matching first, if it doesn't workout
             // we eventually try matching it with derived type *constructor* as well
-            // e.g.; see: integration_tests/derived_types_62.f90
+            // e.g.; see: integration_tests/derived_types_64.f90
             v = current_scope->resolve_symbol("~" + var_name);
             if (!v) {
                 v = current_scope->resolve_symbol(var_name);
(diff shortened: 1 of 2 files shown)


integration_tests/derived_types_62.f90
```diff
@@ -0,0 +1,34 @@
+module mod_derived_types_62
+    implicit none
+    
+    ! derived type with the same name as interface
+    type :: toml_datetime
```
gxyd: Ideally, I should *also* test this by using `toml_datetime` -> `count` (yes, the intrinsic function name), to make sure that this still works in that case.

I'm a little tired today, so I'll do that tomorrow either in this PR or next PR (in case someone else reviews this and merges this)

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -9237,13 +9253,6 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     return;
                 }
             }
-        } else if (ASR::is_a<ASR::Struct_t>(*f2)) {
-            // Check for any interface overriding a constructor for the struct
-            ASR::symbol_t* interface_override_s = current_scope->resolve_symbol("~" + var_name);
-            if (interface_override_s) {
-                v = interface_override_s;
-                f2 = ASRUtils::symbol_get_past_external(interface_override_s);
-            }
```
gxyd, line 9247: I didn't like this approach, which was introduced in https://github.com/lfortran/lfortran/pull/6489, as I think we should do this sooner rather than later.

This PR didn't need this change though for fixing the test case introduced in this PR.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5844,6 +5844,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     },
                     false);
         if( idx == -1 ) {
+            // if no GenericProcedure is matches, we try matching it with
```
certik:
```suggestion
            // if no GenericProcedure matches, we try matching it with
```

certik:

PR #7710, opened 2025-06-13
enh: set `m_side_effect_free` correctly

Pranavchiku:
Towards https://github.com/lcompilers/lpython/issues/1266

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -1066,6 +1066,7 @@ public:
         for( std::uint32_t i = 0; i < n_values; i++ ) {
             this->visit_expr(*m_values[i]);
             ASR::expr_t* expr = ASRUtils::EXPR(tmp);
+            is_side_effect_free = is_side_effect_free && !ASRUtils::does_expression_contain_global_var(expr);
             if(ASRUtils::is_descriptorString(ASRUtils::expr_type(expr))){
                 expr = ASRUtils::cast_string_descriptor_to_pointer(al, expr);
             }
@@ -1185,6 +1186,7 @@ public:
     }
 
     void visit_Write(const AST::Write_t& x) {
+        is_side_effect_free = false;
         create_read_write_ASR_node(x.base, x.class_type);
     }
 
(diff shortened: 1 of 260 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -141,6 +141,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     tmp_vec.clear();
                 }
             }
+            is_deterministic = is_deterministic && !ASRUtils::does_node_contain_global_var(tmp, current_scope);
```
certik, line 144: What is the performance implication of this? For each statement, we now call an extra visitor. So still O(n) processing, but roughly twice slower, as now we need to go over each statement twice, correct?

Cannot we collect the information about each symbol (if it is global or local) as we are constructing it?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1185,6 +1187,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
     }
 
     void visit_Write(const AST::Write_t& x) {
+        is_side_effect_free = false;
```
certik, line 1190: Unless it's a "write" to a local variable, which is used for string formatting. This can be fixed later.

src/libasr/asr_utils.cpp
```diff
@@ -282,6 +282,35 @@ void update_call_args(Allocator &al, SymbolTable *current_scope, bool implicit_i
...
+            ASR::symbol_t* sym = x.m_v;
+            if (sym == nullptr) {
+                return;
+            }
+            if (ASRUtils::symbol_parent_symtab(ASRUtils::symbol_get_past_external(sym))->counter != current_scope->counter) {
+                contain_global_var = true;
+                return;
+            }
+        }
+    };
```
certik, line 305: I think we need to determine this for almost every single Var in ASR --- since Var is in a body and body is in a function. Why not determine this while creating Var, and set internal variable "using_global". We also need to determine if you are reading or writing to global --- I think you don't do that yet, correct?

Pranavchiku:
I'm marking it as draft.

PR #7713, opened 2025-06-13
feat: add support for setting array dimension using pure deferred user-defined functions

### LINKED ISSUE #2882: Deferred type-bound procedure is not recognized
This is another bug regarding type-bound procedures. In the following code, `obj` is reported by LFortran as having no type-bound procedure named `nelements` (both ifort and gfortran compile this code correctly):
```
module test_method

   type, abstract :: abstract_type
   contains
      procedure(nelements), deferred :: nelements
   end type abstract_type

   abstract interface
      pure function nelements(self) result(n)
         import; implicit none
         class(abstract_type), intent(in) :: self
         integer                          :: n
      end function nelements
   end interface
  
   type :: Wrapper
      private
      class(abstract_type), allocatable :: obj 
   contains
      procedure :: caller
   end type Wrapper
   
contains

   subroutine caller(self)
      class(Wrapper), intent(in) :: self      
      real(8), dimension(self%obj%nelements()) :: a
   end subroutine caller

end module test_method
```
Here's LFortran's output:
```
semantic error: Variable 'obj' doesn't have any member named, 'nelements'.
  --> test.f90:27:26
   |
27 |       real(8), dimension(self%obj%nelements()) :: a
   |                          ^^^^^^^^^^^^^^^^^^^^ 

```

kmr-srbh:
Resolves #2882

### DIFF
--- a/integration_tests/class_52.f90
+++ b/integration_tests/class_52.f90
@@ -1,0 +1,82 @@
+module class_52_mod
+
+   type, abstract :: abstract_type
+   contains
+      procedure(nelements), deferred :: nelements
+   end type abstract_type
+
+   abstract interface
+      pure function nelements(self) result(n)
+         import;
+         implicit none
+         class(abstract_type), intent(in) :: self
+         integer                          :: n
+      end function nelements
+   end interface
+
+   type, extends(abstract_type) :: concrete_type
+   contains
+      procedure :: nelements => concrete_nelements
+   end type concrete_type
+
+   type :: Wrapper
+      class(concrete_type), allocatable :: obj
+   contains
+      procedure :: caller_struct_member, caller_simple_subrout, caller_simple_func
+   end type Wrapper
+
+contains
+
+   subroutine caller_struct_member(self)
+      class(Wrapper), intent(in) :: self
+      real(8), dimension(self%obj%nelements()) :: a
+
+      a = [1.0, 2.0, 3.0, 4.0, 5.0]
+      print *, a
+
+      print *, size(a)
+      if (size(a) /= 5) error stop
+   end subroutine
+
+   subroutine caller_simple_subrout(self)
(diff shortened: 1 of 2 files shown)


kmr-srbh:
Using deferred user-defined functions for regular local variables is fully supported. Function arguments and return variable needs to be handled. They can be implemented in the same way as we handle function dependencies. I will push the code for it. The CI will fail for now, so I am cancelling it.

integration_tests/class_52.f90
```diff
@@ -0,0 +1,82 @@
...
+      end function nelements
+   end interface
+
+   type, extends(abstract_type) :: concrete_type
+   contains
+      procedure :: nelements => concrete_nelements
+   end type concrete_type
+
+   type :: Wrapper
+      class(concrete_type), allocatable :: obj
```
kmr-srbh, line 23: We have an issue related to abstract type allocation of struct member not persisting through scopes, so I had to change the type of `obj` from `abstract_type` to `concrete_type` here. I will open an issue for this.

kmr-srbh:
I will resume work on this after I am done with the `StructType` refactoring.

src/libasr/asr_verify.cpp
```diff
@@ -1119,6 +1119,9 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         require(x.m_name,
             "FunctionCall::m_name must be present");
         variable_dependencies.push_back(std::string(ASRUtils::symbol_name(x.m_name)));
+        if (x.m_dt && ASR::is_a<ASR::StructInstanceMember_t>(*x.m_dt)) {
+            visit_expr(*x.m_dt);
+        }
```
kmr-srbh, line 1124: ASR Verify fix.

Issue #7647, opened 2025-06-09
Handle deepcopy of structType variable where there is nested subroutine which contains SIM in assignment


jinangshah21:
MRE:
```fortran
module merge_example
    implicit none
    type :: enum_policy
      integer :: append = 3
   end type enum_policy
   type(enum_policy), parameter :: merge_policy = enum_policy()
   type :: toml_merge_config
      integer :: table = merge_policy%append
   end type toml_merge_config

   interface toml_merge_config
      module procedure :: new_merge_config
   end interface toml_merge_config


contains

function new_merge_config(table) result(config)
   character(*), intent(in), optional :: table

   type(toml_merge_config) :: config

   if (present(table)) config%table = merge_policy%append

contains

   subroutine set_enum(enum, str)
      character(*), intent(in) :: str
      integer, intent(inout) :: enum
      enum = merge_policy%append
   end subroutine set_enum

end function new_merge_config

end module merge_example

program test
    use merge_example
    type(toml_merge_config) :: config2, config3
    config2 = config3
end program
```

LFortran Error:
```console
    this->visit_stmt(*x.m_body[i]);
  File "/home/jinang_shah/Desktop/lfortran/src/libasr/../libasr/asr_base_visitor.h", line 361, in LCompilers::ASR::BaseVisitor<LCompilers::ASRToLLVMVisitor>::visit_stmt(LCompilers::ASR::stmt_t const&)
    void visit_stmt(const stmt_t &b) { visit_stmt_t(b, self()); }
  File "/home/jinang_shah/Desktop/lfortran/src/libasr/../libasr/asr_base_visitor.h", line 55, in void LCompilers::ASR::visit_stmt_t<LCompilers::ASRToLLVMVisitor>(LCompilers::ASR::stmt_t const&, LCompilers::ASRToLLVMVisitor&)
    case stmtType::Assignment: { v.visit_Assignment((const Assignment_t &)x); return; }
  File "/home/jinang_shah/Desktop/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5654, in LCompilers::ASRToLLVMVisitor::visit_Assignment(LCompilers::ASR::Assignment_t const&)
    llvm_utils->deepcopy(value_struct, target_struct,
[...]

PR #7722, opened 2025-06-14
fix: in deepcopy iterate through structs' members through m_members rather than m_symtab

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7647

### DIFF
--- a/integration_tests/derived_types_63.f90
+++ b/integration_tests/derived_types_63.f90
@@ -1,0 +1,43 @@
+module derived_types_63_m
+    implicit none
+    type :: enum_policy
+      integer :: append = 3
+   end type enum_policy
+   type(enum_policy), parameter :: merge_policy = enum_policy()
+   type :: toml_merge_config
+      integer :: table = merge_policy%append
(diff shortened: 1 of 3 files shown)


integration_tests/derived_types_63.f90
```diff
@@ -0,0 +1,43 @@
+module derived_types_63_m
+    implicit none
+    type :: enum_policy
+      integer :: append = 3
+   end type enum_policy
```
certik, line 5: The indentation is inconsistent.

integration_tests/derived_types_63.f90
```diff
@@ -0,0 +1,43 @@
...
+end function new_merge_config
+
+end module derived_types_63_m
+
+program derived_types_63
+    use derived_types_63_m
+    type(toml_merge_config) :: config2, config3
+    config2 = config3
+
+    if (config2%table /= config3%table) error stop
```
certik, line 42: I would also check that it is actually equal to 3, both config2 and config3, to ensure things got assigned and copied correctly.

PR #7726, opened 2025-06-14
enh: assert pos before accessing from `Vec`

Pranavchiku:


src/libasr/containers.h
```diff
@@ -161,6 +161,8 @@ struct Vec {
     }
 
     const T& operator[](size_t pos) const {
+        LCOMPILERS_ASSERT(pos >= 0);
```
certik: The error at the CI is:
```
/home/runner/work/lfortran/lfortran/lfortran-0.53.0-54-ga9796672a/src/libasr/../libasr/containers.h:164:31: error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
  164 |         LCOMPILERS_ASSERT(pos >= 0);
```
That's a good point, `pos` is always nonnegative, so we don't need this assert here.

certik:
```suggestion
```

certik:
Looks like this uncovered a lot of bugs. We'll fix them all. A quick workaround is to create an empty / default element if `pos > n`, since that is what it was doing before implicitly --- just open up an issue for each case and we can fix the actual bug later. I think it's important to get this PR in, as it will prevent new bugs and help with debugging in the future.

Pranavchiku:
I've opened issues:

* [ ] #7737 
* [ ] #7738 
* [ ] #7739 
* [ ] #7740 
* [ ] #7741

src/libasr/containers.h
```diff
@@ -161,6 +161,13 @@ struct Vec {
     }
 
     const T& operator[](size_t pos) const {
+        if ( pos >= n ) {
+            // TODO: remove this once everything is fixed.
+            // create a dummy object to avoid undefined behavior
+            for (size_t j = n; j <= pos; j++) {
+                std::memcpy(&p[j], &p[n-1], sizeof(T));
```
certik, line 168: This has to be done in the caller. I would add the assert, and then start fixing the bugs.

krystophny:
Closing this PR as the underlying Vec out-of-bounds bugs have been fixed in PR #9040. The workaround approach in this PR (expanding the array to hide bugs) is no longer needed - proper size checks were added to the affected code.

krystophny:
@Pranavchiku The issues this PR was addressing (#7738, #7739, #7740, #7741, #7833) have been fixed by #9040, which added proper size checks in `eval_ArrIntrinsic` and the `LCOMPILERS_ASSERT(pos < n)` debug check.

This PR's workaround approach (silently copying data on OOB access) is no longer needed. Can this PR be closed?

certik:
This can be closed now, as https://github.com/lfortran/lfortran/pull/9040 implemented a proper fix.

Issue #7732, opened 2025-06-14
CI: dyld[6900]: Library not loaded: @rpath/libgfortran.5.dylib


certik:
https://github.com/lfortran/lfortran/actions/runs/15655622886/job/44106323500?pr=7731

```
dyld[6900]: Library not loaded: @rpath/libgfortran.5.dylib
  Referenced from: <B0F1412E-0434-3796-8844-C45D6BA33301> /Users/runner/micromamba/pkgs/fpm-0.12.0-hcfc1361_0/bin/fpm
  Reason: tried: '/Users/runner/micromamba/pkgs/fpm-0.12.0-hcfc1361_0/bin/libgfortran.5.dylib' (no such file), '/Users/runner/micromamba/pkgs/fpm-0.12.0-hcfc1361_0/bin/../lib/libgfortran.5.dylib' (no such file), '/Users/runner/micromamba/pkgs/fpm-0.12.0-hcfc1361_0/bin/libgfortran.5.dylib' (no such file), '/Users/runner/micromamba/pkgs/fpm-0.12.0-hcfc1361_0/bin/../lib/libgfortran.5.dylib' (no such file), '/usr/local/lib/libgfortran.5.dylib' (no such file), '/usr/lib/libgfortran.5.dylib' (no such file, not in dyld cache)
ci/test_third_party_codes.sh: line 49:  6900 Abort trap: 6           fpm build --compiler=$FC --profile release --flag "--cpp -DREAL32" --verbose
```

Other such failures:

* https://github.com/lfortran/lfortran/actions/runs/15631417248/job/44036430895

It only happens sometimes, so restarting the job usually fixes it.

certik:
Here is the fpm recipe: https://github.com/conda-forge/fpm-feedstock/blob/7368a57cbd4a1ddd19bc667a6462552b00f3e95b/recipe/meta.yaml, the package depends on gfortran.

When install fpm using conda, it downgrades `libgfortran` from 14.2.0 to 5.0.0, not sure if they just changed version scheme, or it is an actual downgrade:
```
  Updating specs:

   - fpm


  Package         Version  Build                Channel           Size
────────────────────────────────────────────────────────────────────────
  Install:
────────────────────────────────────────────────────────────────────────

  + fpm            0.12.0  hcfc1361_0           conda-forge      722kB

  Change:
────────────────────────────────────────────────────────────────────────

  - libgfortran5   14.2.0  h2c44a93_105         conda-forge     Cached
[...]

PR #7733, opened 2025-06-14
CI: Debug fpm installation

certik:
Towards #7732.


certik:
On both linux and macOS the `fpm` binary is in the conda environment as it should be, not a link to the unpacked package path. So rpath should resolve against that. In the failed CI log it however is resolving it against the unpacked package path, for whatever reason.

Issue #7609, opened 2025-06-06
Code Generation Error due to no implicit type promotion


Jaysukh-409:
- While running example of `radix_sort` in `stdlib` sorting package, due to `bin_idx = arr(1)` I am getting code generation failure. Since, GFortran do implicit type promotion and that's why for it works fine. But not in LFortran.

MRE:
```fortran
program main
    use iso_fortran_env, only: int8, int64
    implicit none
    
    integer, parameter :: int_size = int64
    integer(int8), allocatable :: arr(:)
    integer(kind=int_size) :: N = 5

    arr = [-128, 127, 0, -1, 1]
    call radix_sort(N, arr)

contains
    subroutine radix_sort(N, arr)
        integer(kind=int_size), intent(in) :: N
        integer(kind=int8), dimension(N), intent(inout) :: arr
        integer :: bin_idx
 
        bin_idx = arr(1)

    end subroutine radix_sort
end program main
```

GFortran Output:
```console
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ gfortran a.f90 && ./a.out
```

LFortran Output:
```console
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ lfortran a.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

%array = type { i8*, i32, %dimension_descriptor*, i1, i32 }
%dimension_descriptor = type { i32, i32, i32 }

@main.n = internal global i64 5

define i32 @main(i32 %0, i8** %1) {
.entry:
  %array_bound = alloca i32, align 4
  %array_size11 = alloca i32, align 4
  %2 = alloca i32, align 4
  %3 = alloca i32, align 4
  %array_size = alloca i32, align 4
  %__1_k = alloca i32, align 4
  %__lcompilers_i_0 = alloca i32, align 4
  %__libasr_index_0_ = alloca i32, align 4
  %__libasr_index_0_1 = alloca i32, align 4
  %int_size = alloca i32, align 4
  store i32 8, i32* %int_size, align 4
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %__1_k1 = alloca i32, align 4
  %__lcompilers_i_02 = alloca i32, align 4
  %__libasr__created__var__0__array_constant_ = alloca [5 x i8], align 1
  %__libasr_created__subroutine_call_radix_sort = alloca %array*, align 8
[...]

PR #7745, opened 2025-06-15
test: add test for implicit type promotion

Jaysukh-409:
FIX: #7609

### DIFF
--- a/integration_tests/implicit_type_promotion.f90
+++ b/integration_tests/implicit_type_promotion.f90
@@ -1,0 +1,21 @@
+program implicit_type_promotion
+    use iso_fortran_env, only: int8
+    implicit none
+    integer(int8), allocatable :: arr(:)
+    integer :: N = 5
+
+    arr = [-128, 127, 0, -1, 1]
+    call radix_sort(N, arr)
+
+contains
+    subroutine radix_sort(N, arr)
+        integer, intent(in) :: N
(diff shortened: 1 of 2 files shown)


integration_tests/implicit_type_promotion.f90
```diff
@@ -0,0 +1,21 @@
...
+        integer, intent(in) :: N
+        integer(kind=int8), dimension(N), intent(inout) :: arr
+        integer :: bin_idx
+ 
+        bin_idx = arr(1)
+
+        if (bin_idx /= -128) error stop
+
+    end subroutine radix_sort
+end program implicit_type_promotion
```
gxyd, line 21: I don't this is the correct test program for the MRE present in issue: https://github.com/lfortran/lfortran/issues/7609#issue-3124308617.

I checkout the commit: 76068256f4923e208ba3f1ddc844686350b3656e from *main* branch of LFortran (one of the commit before the linked issue was created), and I built LFortran with it, and I ran the linked issue's MRE (not the one from this PR) and it failed as expected, while the MRE in this PR worked without any issue. Hence this MRE isn't the right test program to add, something is missing in this.

HarshitaKalani: @Jaysukh-409 can you consider adding the exact MRE mentioned in the issue: https://github.com/lfortran/lfortran/issues/7609?

PR #7748, opened 2025-06-15
CI: Enable more test for stdlib_sorting after reverting workaround

Jaysukh-409:


Pranavchiku (changes requested):
can you please do it over `n-lf-13`? Let's sync if it is not.

Issue #7743, opened 2025-06-15
Handle labels directly declare by if stmt


jinangshah21:
MRE:
```fortran
program test_named_block
    implicit none

    integer :: basein_local
    basein_local = 1  ! Try 1 or 2 to test both branches

ALL: if (basein_local < 0) then
        print *, "Inside IF block: basein_local =", basein_local
    else
        TMP: block 
            exit ALL
        end block TMP
    end if ALL
end program test_named_block
```

LFortran Error:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran d.f90
code generation error: Could not find block or loop named ALL in parent scope to exit from.
  --> d.f90:11:13
   |
11 |             exit ALL
   |             ^^^^^^^^ 
```

Pranavchiku:
We don't store label information in ASR

```clojure
                    (If
                        (IntegerCompare
                            (Var 2 basein_local)
                            Lt
                            (IntegerConstant 0 (Integer 4) Decimal)
                            (Logical 4)
                            ()
                        )
                        [(Print
                            (StringFormat
                                ()
                                [(StringConstant
                                    "Inside IF block: basein_local ="
                                    (String 1 (IntegerConstant 31 (Integer 4) Decimal) ExpressionLength PointerString)
                                )
                                (Var 2 basein_local)]
                                FormatFortran
                                (String 1 () ExpressionLength CString)
                                ()
                            )
                        )]
                        [(BlockCall
                            -1
                            2 TMP
                        )]
                    )
```

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 60edd0959..7997d1edb 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -43,7 +43,7 @@ stmt
     | ForEach(expr var, expr container, stmt* body)
     | GoTo(int target_id, identifier name)
     | GoToTarget(int id, identifier name)
-    | If(expr test, stmt* body, stmt* orelse)
+    | If(identifier? name, expr test, stmt* body, stmt* orelse)
     | IfArithmetic(expr test, int lt_label, int eq_label, int gt_label)
     | Print(expr text)
     | FileOpen(int label, expr? newunit, expr? filename, expr? status, expr? form, expr? access, expr? iostat, expr? iomsg)
```

And then need to process it the way we do for `WhileLoop`

Pranavchiku:
[...]

PR #7751, opened 2025-06-15
feat: add label to if statement

Pranavchiku:
Towards #7743.

### DIFF
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -43,7 +43,7 @@ stmt
     | ForEach(expr var, expr container, stmt* body)
     | GoTo(int target_id, identifier name)
     | GoToTarget(int id, identifier name)
-    | If(expr test, stmt* body, stmt* orelse)
+    | If(identifier? name, expr test, stmt* body, stmt* orelse)
     | IfArithmetic(expr test, int lt_label, int eq_label, int gt_label)
     | Print(expr text)
     | FileOpen(int label, expr? newunit, expr? filename, expr? status, expr? form, expr? access, expr? iostat, expr? iomsg)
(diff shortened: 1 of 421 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1390,6 +1390,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 
     ASR::accessType dflt_access = ASR::Public;
     bool in_module = false;
+    int if_label_go_to_id = 0; // Used to generate unique labels for GoTo statements
+    std::unordered_map<std::string, ASR::stmt_t*> if_label_mapping; // maps if label to corresponding exit GoToTarget
```
certik: Why do we need this for `if`, but not for `do` and `while`?

src/libasr/ASR.asdl
```diff
@@ -43,7 +43,7 @@ stmt
     | ForEach(expr var, expr container, stmt* body)
     | GoTo(int target_id, identifier name)
     | GoToTarget(int id, identifier name)
-    | If(expr test, stmt* body, stmt* orelse)
+    | If(identifier? name, expr test, stmt* body, stmt* orelse)
```
certik, line 46: I think that this is needed, just like we already have it for `DoLoop` and `WhileLoop`, so that `Exit` can jump to it.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: We can extend it. It is fine.

src/libasr/ASR.asdl
Pranavchiku: Yup.

Issue #7707, opened 2025-06-13
Initializing polymorphic variable with the user-defined constructor of a concrete extended type does not work


difference-scheme:
This is a variation of #7538 that still fails:
$ cat test.f90
```
module types_module

   type, abstract :: AbsType
   end type AbsType
   
   type, extends(AbsType) :: MyType
   end type MyType

   interface MyType
      procedure :: init
   end interface MyType

contains

   function init() result(self)
      type(MyType) :: self
   end function init

end module types_module

program test

   use types_module
   
   class(AbsType), allocatable :: obj

   obj = MyType()

end program test
```

$ lfortran -c --std=f23 --generate-object-code  test.f90 
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

%mytype = type <{ %abstype }>
%abstype = type <{}>
%__vtab_mytype = type { i64 }
%__vtab_abstype = type { i64 }
%abstype_polymorphic = type { i64, %abstype* }

define void @__module_types_module_init(%mytype* %self) {
.entry:
  %0 = alloca %__vtab_mytype, align 8
  %1 = getelementptr %__vtab_mytype, %__vtab_mytype* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  br label %return

return:                                           ; preds = %.entry
  ret void
}

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %2 = alloca %__vtab_abstype, align 8
  %3 = getelementptr %__vtab_abstype, %__vtab_abstype* %2, i32 0, i32 0
  store i64 1, i64* %3, align 4
  %4 = alloca %__vtab_mytype, align 8
  %5 = getelementptr %__vtab_mytype, %__vtab_mytype* %4, i32 0, i32 0
  store i64 0, i64* %5, align 4
  %obj = alloca %abstype_polymorphic, align 8
  %6 = getelementptr %abstype_polymorphic, %abstype_polymorphic* %obj, i32 0, i32 0
  store i64 1, i64* %6, align 4
  %7 = getelementptr %abstype_polymorphic, %abstype_polymorphic* %obj, i32 0, i32 1
  store %abstype* null, %abstype** %7, align 8
  %8 = getelementptr %abstype_polymorphic, %abstype_polymorphic* %obj, i32 0, i32 1
  %9 = load %abstype*, %abstype** %8, align 8
[...]

PR #7752, opened 2025-06-15
fix: bitcast class allocatable to required type and do automatic allocation of StructType in backend instead of ASR

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7707

Also revert a change from https://github.com/lfortran/lfortran/pull/7700, which inserted `Allocate` in ASR. Instead use `check_and_allocate()` in llvm backend.


gxyd (approved):
Looks good to me, thanks @xaerru

kmr-srbh:
The `check_and_allocate()` function creates the `Allocate` ASR node in the LLVM back-end and visits it for allocation. I think we should not create ASR nodes in the back-end and instead be explicit in the ASR itself. Currently, we know that we allocate `class` variables on assignment with `--realloc-lhs`, but we do not show that in the ASR. I think we should have it in the front-end, or some ASR pass, instead of the back-end.

difference-scheme:
Thanks for the fix. This has eliminated a number of error messages for me.

There's a related case that still fails, though. I've opened Issue #7758 for it.

Issue #7758, opened 2025-06-15
Initializing a wrapped polymorphic object with the user-defined constructor of an extended type does not work


difference-scheme:
The following is a variation of #7707 that still fails:
```
$ cat test.f90 
module testmod
   
   type, abstract :: AbsType
   end type AbsType
   
   type, extends(AbsType) :: MyType
   end type MyType
   
   interface MyType
      procedure :: init
   end interface MyType

   type :: Wrapper
      class(AbsType), allocatable :: obj
   contains
      procedure :: client
   end type Wrapper
   
contains

   function init() result(self)
      type(MyType) :: self
   end function init

   subroutine client(self)
      class(Wrapper), intent(inout) :: self
      self%obj = MyType()
   end subroutine client

end module testmod
```

```
$ lfortran -c --std=f23 --generate-object-code test.f90 
; ModuleID = 'LFortran'
source_filename = "LFortran"

%wrapper_polymorphic = type { i64, %wrapper* }
%wrapper = type <{ %abstype_polymorphic }>
%abstype_polymorphic = type { i64, %abstype* }
%abstype = type <{}>
%__vtab_wrapper = type { i64 }
%mytype = type <{ %abstype }>
%__vtab_mytype = type { i64 }

define void @__module_testmod_client(%wrapper_polymorphic* %self) {
.entry:
  %0 = alloca %__vtab_wrapper, align 8
  %1 = getelementptr %__vtab_wrapper, %__vtab_wrapper* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  %2 = getelementptr %wrapper_polymorphic, %wrapper_polymorphic* %self, i32 0, i32 1
  %3 = load %wrapper*, %wrapper** %2, align 8
  %4 = getelementptr %wrapper, %wrapper* %3, i32 0, i32 0
  %5 = getelementptr %wrapper_polymorphic, %wrapper_polymorphic* %self, i32 0, i32 1
  %6 = load %wrapper*, %wrapper** %5, align 8
  %7 = getelementptr %wrapper, %wrapper* %6, i32 0, i32 0
  %8 = getelementptr %abstype_polymorphic, %abstype_polymorphic* %7, i32 0, i32 1
  %9 = load %abstype*, %abstype** %8, align 8
  %10 = icmp eq %abstype* %9, null
  %11 = getelementptr %abstype_polymorphic, %abstype_polymorphic* %7, i32 0, i32 0
  %12 = load i64, i64* %11, align 4
[...]

PR #7762, opened 2025-06-16
fix: convert class to type when arg is StructInstanceMember

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7758


difference-scheme:
Thanks a lot for the fix. It eliminated a few more errors for me.

Issue #7754, opened 2025-06-15
Parser: add tests for `_lfortran_dict(integer, real(8)) :: x`


certik:
As a follow up on https://github.com/lfortran/lfortran/pull/7698, that PR likely already supports the syntax without `type`:
```
_lfortran_dict(integer, real(8)) :: x
```

Add tests for it.

certik:
See also https://github.com/lfortran/lfortran/pull/7684#discussion_r2142595025.

PR #7763, opened 2025-06-16
Updated syntax for LF specific types

swamishiju:
Towards #7754


integration_tests/lp_set_test_01.f90
```diff
@@ -1,18 +1,18 @@
...
-    if (_lfortran_len(test_set) /= 5) error stop
-     
-    ! Add other intrinsics later
+   test_set = _lfortran_set_constant(1, 2, 3, 4)
+   if (_lfortran_len(test_set) /= 4) error stop
+   call _lfortran_set_add(test_set, -50)
+   if (_lfortran_len(test_set) /= 5) error stop
+
+   ! Add other intrinsics later
 end program
```
certik, line 22: I think now the other style `type(_lfortran_set(integer))` is not tested, correct?

We need to test both styles. I would recommend keeping the old ones intact, but adding a few new declarations for the `_lfortran_set(integer)` style. That should do it.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3450,8 +3450,20 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                if (AST::is_a<AST::AttrType_t>(*x.m_vartype))
+                    sym_type = AST::down_cast<AST::AttrType_t>(x.m_vartype);
+                else if (AST::is_a<AST::AttrTypeList_t>(*x.m_vartype))
+                    sym_type = AST::down_cast<AST::AttrType_t>(
+                                AST::down_cast<AST::decl_attribute_t>(
+                                    AST::make_AttrType_t(
+                                    al, x.m_vartype->base.loc, 
+                                    AST::decl_typeType::TypeType,
+                                    nullptr, 0, x.m_vartype, 
+                                    nullptr, AST::symbolType::None)));
```
certik, line 3464: Add {} around the body. What happens if neither if statement is true?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -450,7 +450,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                    } else if  (AST::is_a<AST::AttrTypeList_t>(*decl->m_vartype)){
+                        type = AST::down_cast<AST::AttrType_t>(
+                                AST::down_cast<AST::decl_attribute_t>(
+                                    AST::make_AttrType_t(
+                                    al, decl->m_vartype->base.loc, 
+                                    AST::decl_typeType::TypeType,
+                                    nullptr, 0, decl->m_vartype, 
+                                    nullptr, AST::symbolType::None)));
+                        
+                    }
```
certik: Same here, what happens if neither if statement executes, then `type` is a nullptr, so it will fail later?

src/lfortran/semantics/ast_common_visitor.h
swamishiju: Ideally it shouldn't reach there I'll put a semantic error or maybe an assert?

certik: Assert is good. Put {} around the if statement body.

certik: You can do it in a subsequent PR.

Issue #7750, opened 2025-06-15
`open` statement with untrimmed `status` string fails


Jaysukh-409:
When providing the status with the not trimmed length string and valid status, it causes error with LFortran and not open the file. But in GFortran it automatically trimmed while compiling so it works fine.

MRE:
```fortran
program main
    implicit none
    integer :: unit, iostat
    character(128) :: iomsg
    
    open(newunit=unit, file='temp.txt', form='formatted', status='replace    ', iostat=iostat, iomsg=iomsg)

    print *, unit
    print *, iostat
    if (iostat /= 0) print *, iomsg
end program main
```

Output:
```console
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ lfortran b.f90
0
5002
STATUS specifier in OPEN statement has invalid value.                                                                           
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ gfortran b.f90 && ./a.out
         -10
           0
```

kmr-srbh:
@Jaysukh-409 we had handled this for `file` in #6283. I think it can be handled similarly for `form` and `status`.

PR #7765, opened 2025-06-16
enh: trim trailing spaces from the form & status of `open`

Jaysukh-409:
Fixes: #7750 

- Remove the trailing spaces from the `form` and `status` both argument of `open` statement


Pranavchiku (approved):
It is fine with me, will wait for one more review.

PR #7771, opened 2025-06-16
Phasing out typecodes

swamishiju:
wip

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -5098,9 +5098,9 @@ namespace LCompilers {
                           ASR::ttype_t* asr_type, llvm::Module* module,
                           std::map<std::string, std::map<std::string, int>>& name2memidx) {
         std::string type_code = ASRUtils::get_type_code(asr_type);
-        int type_size = std::get<1>(typecode2listtype[type_code]);
-        llvm::Type* el_type = std::get<2>(typecode2listtype[type_code]);
-        llvm::Type* list_type = std::get<0>(typecode2listtype[type_code]);
+        llvm::Type* el_type = llvm_utils->get_type_from_ttype_t_util(asr_type, module);
+        int type_size = el_type->getScalarSizeInBits()/8;
+        llvm::Type* list_type = llvm_utils->get_type_from_ttype_t_util(asr_type, module);
 
 
         llvm::Value* current_end_point = llvm_utils->CreateLoad2(llvm::Type::getInt32Ty(context),
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -1,3 +1,4 @@
+#include "libasr/asr.h"
 #include <libasr/assert.h>
 #include <libasr/codegen/llvm_utils.h>
 #include <libasr/codegen/llvm_array_utils.h>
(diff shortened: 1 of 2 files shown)


certik:
@kmr-srbh is the design good here, accomplishing what you wanted to get done?

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -5097,20 +5124,18 @@ namespace LCompilers {
...
                           ASR::ttype_t* asr_type, llvm::Module* module,
                           std::map<std::string, std::map<std::string, int>>& name2memidx) {
-        std::string type_code = ASRUtils::get_type_code(asr_type);
-        int type_size = std::get<1>(typecode2listtype[type_code]);
-        llvm::Type* el_type = std::get<2>(typecode2listtype[type_code]);
-        llvm::Type* list_type = std::get<0>(typecode2listtype[type_code]);
-
+        llvm::Type* el_type = llvm_utils->get_type_from_ttype_t_util(asr_type, module);
+        int type_size = std::get<1>(type2listtype[asr_type]);
+        llvm::Type* list_type = get_list_type_from_ttype(asr_type);
```
kmr-srbh: Let's fetch all values from the map. Also, we need to make sure that the key passed to `type2listtype` is equal to the element type of the list. So if `asr_type` here refers to the ttype `ASR::List_t`, we need to pass in `asr_type->m_type` which is the element type. If it is already the element type, then it's fine.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2326,6 +2330,29 @@ namespace LCompilers {
         return list_desc;
     }
 
+    llvm::Type* LLVMList::get_list_type_from_ttype(ASR::ttype_t* type, int type_size) {
```
kmr-srbh:
```suggestion
    llvm::Type* LLVMList::get_list_type_from_ttype(ASR::ttype_t* el_ttype, int type_size) {
```

src/libasr/codegen/llvm_utils.h
```diff
@@ -446,18 +446,19 @@ namespace LCompilers {
             llvm::IRBuilder<>* builder;
 
             std::map<std::string, std::tuple<llvm::Type*, int32_t, llvm::Type*>> typecode2listtype;
+            std::map<ASR::ttype_t*, std::tuple<llvm::Type*, int32_t, llvm::Type*>> type2listtype;
```
kmr-srbh: This is good. I was a bit skeptical at first to have a pointer as the key, but it looks like we already have such maps in the codebase (with keys `ASR::expr_t*` and `ASR::symbol_t*`). Hopefully we will not encounter any issues with dangling pointers.

kmr-srbh (commented):
The changes are in the right direction. Thanks for this @swamishiju! It will be very helpful.

src/libasr/codegen/llvm_utils.h
swamishiju: Yes I was a bit averse to it too, but I don't think there is a better alternative.

certik:
What is the status here --- why was this PR closed?

kmr-srbh:
I think Swami force pushed his branch, causing the commits to disappear, hence closing the PR. From what we observed with the `StructType` refactor PR, we should be able to merge without the LPython data-structures also. The type-codes will be required for testing the C/CPP back-end though (same reason, they are required for the types in the back-end). So I think we can narrow the scope of this refactor.

PR #7776, opened 2025-06-17
feat: allow 'optional' attribute with dummy Function argument

gxyd:
## Description

Example Fortran program where `optional` dummy Function argument is used:
```f90
module optional_06_mod1
    implicit none
    public :: cauchy

    interface cauchy
        module procedure cauchy_sngl_opt
    end interface cauchy

contains
    subroutine cauchy_sngl_opt(i)
        integer, intent(out) :: i
        i = 6
    end subroutine cauchy_sngl_opt
end module optional_06_mod1

module optional_06_mod2
    use optional_06_mod1
    implicit none
contains
    subroutine wind0_sngl(i, f)
        implicit none
        integer, intent(out) :: i
        interface
            function f(z) result(r)
                integer :: r
                complex, intent(in) :: z
            end function f
        end interface
        optional :: f     ! optional dummy Function argument
        complex :: z

        z = (1.0, 2.0)
        i = 0

        if (present(f)) then
            i = f(z)
        else
            call cauchy(i)
        end if
    end subroutine wind0_sngl
end module optional_06_mod2

```

### DIFF
--- a/src/lfortran/tests/test_llvm.cpp
+++ b/src/lfortran/tests/test_llvm.cpp
@@ -378,7 +378,7 @@ end function)";
     LCompilers::LocationManager lm;
     LCompilers::ASR::TranslationUnit_t* asr = TRY(LCompilers::LFortran::ast_to_asr(al, *tu,
         diagnostics, nullptr, false, compiler_options, lm));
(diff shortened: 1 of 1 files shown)


certik:
The `optional_06.f90` test is not registered.

certik:
I think `f` should be a Variable, of type FunctionType. The Variable should have `presence` and it already does. The Variable can possibly also link to a Function symbol for additional info. But I think Function itself shouldn't have `presence`?

gxyd:
> The `optional_06.f90` test is not registered.

That was intentional to not register it in integration tests, as it fails in our ASR pass currently.



> I think `f` should be a Variable, of type FunctionType. The Variable should have `presence` and it already does. The Variable can possibly also link to a Function symbol for additional info.

I think you are right, we can have other attributes along with `f` like `volatile`, `value`, `target` etc., hence it makes sense to have it as a Variable instead of `Function`.

> But I think Function itself shouldn't have presence?

I think you are right.

PR #7777, opened 2025-06-17
CI: run "LFortran CI" job with ubuntu + LLVM 20 as well

gxyd:
## Description

Modify "LFortran CI" job to include LLVM 20 run as well.

More changes are definitely needed at the CI currently for this to pass though


gxyd:
I get an error:
```console
  error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ llvmdev =20 * is installable and it requires
      │  └─ zstd >=1.5.7,<1.6.0a0 *, which can be installed;
      └─ zstd-static =1.5.5 * is not installable because it requires
         └─ zstd ==1.5.5 hfc55251_0, which conflicts with any installable versions previously reported.
  critical libmamba Could not solve for environment specs
  Error: The process '/home/runner/micromamba-bin/micromamba' failed with exit code 1
```

On the CI job: *LFortran CI (3.10, ubuntu-latest, LLVM 20)*, to resolve that it will need changes in zstd-static version for LLVM 20

gxyd:
Now the latest failure with LLVM 20 is:
```console
subroutine3.f90 * llvm
Reference JSON: tests/reference/llvm-subroutine3-fb7e0b8.json
Output JSON:    tests/output/llvm-subroutine3-fb7e0b8.json
Omitting 10 identical items
Differing items:
{'stdout_hash': 'f059d770482bda5a0f5ef2f3334a96fc45d083270eb003761bf693cb'} != {'stdout_hash': '203462336eb580ad635038b53ab2235056bb0572890be6c2f2c99fbc'}
Diff against: tests/reference/llvm-subroutine3-fb7e0b8.stdout
7c7
<   store i32 42, i32* %f, align 4
---
>   store i32 42, ptr %f, align 4
11c11
<   %0 = load i32, i32* %f, align 4
---
>   %0 = load i32, ptr %f, align 4

```

because, we shouldn't really run reference tests with LLVM 20, just LLVM 11. Next change will be to fix it.

gxyd:
This PR in-effect introduces extra testing, only for LLVM 20 with ubuntu, i.e. we need to confirm we want to run the below tests with LLVM 20 with ubuntu or not:

✅  Test (Linux / macOS)
✅  Test LFortran's Command Line Interface (Linux / macOS)  -> this anyways takes very less time
✅  Test with specific Fortran standard
❌  LSP Test (Linux / macOS)  -> Is there a need to test this with LLVM 20 as well?

I've put ✅  or ❌  depending on whether I think we should test with LLVM 20 with ubuntu or not.

certik:
We don't need to test LSP for both LLVM.

gxyd:
> We don't need to test LSP for both LLVM.

Makes sense, I've disabled LSP on LLVM 20

.github/workflows/CI.yml
```diff
@@ -62,16 +73,27 @@ jobs:
...
+        run: |
+          micromamba install bison=3.4
+          if [[ "${{ matrix.llvm-version }}" == "11" ]]; then
+            micromamba install zstd-static=1.5.5 zlib=1.2.13
+          elif [[ "${{ matrix.llvm-version }}" == "20" ]]; then
+            micromamba install zstd-static=1.5.7 zlib=1.3.1
+          fi
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" && "${{ matrix.llvm-version }}" != "11" ]]; then
+            micromamba install libunwind=1.7.2
+          fi
```
gxyd, line 96: I verified that zlib is automatically installed but the version by-default is different from what's installed with LLVM 11 here, and zstd-static isn't installed at all, unless explicitly specified.

So, I think it's better as it currently is.

certik:
Ok, let's get the CI to catch up, and the other PRs to go in first. Then we'll rebase this and I'll merge it and update the required tests.

certik:
Ok, all other PRs merged, I am rebasing this. If it passes, I'll merge and update the required tests.

certik:
Ok, merged, required tests updated.

PR #7778, opened 2025-06-17
chore: accept "extends" attribute of Struct ASR node in Fortran backend

gxyd:


src/libasr/asr_utils.h
```diff
@@ -997,7 +998,38 @@ static inline std::pair<char**, size_t> symbol_dependencies(const ASR::symbol_t
...
+            }
+
+            size_t total_deps = unique_deps.size();
+            char** merged_deps = new char*[total_deps];
+            size_t idx = 0;
+            for (const auto& dep : unique_deps) {
+                merged_deps[idx] = strdup(dep.c_str());
+                ++idx;
+            }
+            return std::make_pair(merged_deps, total_deps);
```
gxyd, line 1032: the above diff allows us to make sure correctly order the dependencies of Struct, which are defined using `m_parent` (i.e. using `extends` in Fortran code), which is what was failing from the first commit for `class_allocate_01.f90`

gxyd:
Thanks for the review.

PR #7786, opened 2025-06-17
refactor: simplify LLVM utility functions by removing versions overloaded with `ASR::ttype_t*` parameter

### LINKED ISSUE #7785: refactor: remove LLVM utility functions overloaded with a `ASR::ttype_t*` parameter
This refactor is required for helping the `StructType` refactor. The chief issue which the overloaded functions cause, are related to `StructType` handling. Initially, all operations on expressions/variables of type `StructType` depended on extracting the struct symbol from the type and working with it. With #5791, we remove the struct symbol from the type. This means that now we need either the `Variable` symbol, or an `ASR::expr_t*` expression to work with entities of type `StructType`. Given this context, some LLVM utility functions overloaded with versions having a `ASR::ttype_t* type` parameter create an issue. Such functions pass the `type` parameter to the `get_type_from_ttype_t_util()` function to get the `llvm::Type` to use later. With #5791, the `get_type_from_ttype_t_util()` function now requires an `ASR::expr_t* expr` parameter to be passed to it, but the afore mentioned utility functions did not have one to provide it. This expression is solely used to resolve the LLVM type for `StructType` using a helper function to fetch the struct symbol. Faced with this issue, we have 2 ways forward:
- Add a `ASR::expr_t* expr` parameter to the LLVM utility functions and refactor all places where it is called.
- Remove the overloaded functions and refactor all places where it is called.

As one can convert a `ASR::ttype_t*` to `llvm::Type` using the `get_type_from_ttype_t_util()` function, the overloaded functions are left redundant, so it seems ideal to choose the second option.

This refactor will require:
- Finding places where these overloaded functions are used. This will be done automatically by a language server like `clangd`.
- Converting the passed `ASR::ttype_t*` value to `llvm::Type` using `get_type_from_ttype_t_util()` and passing it to the same function (as the overloaded functions have similar signature).

kmr-srbh:
Resolves #7785

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1253,7 +1253,9 @@ public:
 
 
             if(ASRUtils::is_array(ASRUtils::expr_type(x.m_vars[i]))){
-                llvm::Value* target_ = llvm_utils->CreateLoad2(ASRUtils::expr_type(x.m_vars[i]), target);
+                llvm::Type* x_m_vars_type = llvm_utils->get_type_from_ttype_t_util(
+                    ASRUtils::expr_type(x.m_vars[i]), module.get());
+                llvm::Value* target_ = llvm_utils->CreateLoad2(x_m_vars_type, target);
                 llvm::Value* data_ptr = arr_descr->get_pointer_to_data(ASRUtils::expr_type(x.m_vars[i]), target_ , module.get());
                 builder->CreateStore(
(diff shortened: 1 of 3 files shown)


kmr-srbh:
As copy-pasting the variable definitions was easy, I ended up creating variables for most of the changes until I realised it. Even though these variables will be allocated on the stack and freed later, I will make sure to pass the function call in-place when handling the changes made here in the `StructType` refactor PR.

Issue #7377, opened 2025-05-20
Error when calling nested subroutine without providing Type as argument


Jaysukh-409:
The similar problem I was facing while compiling `stdlib_logger`. This is the simpler version of the main problem I was facing there.

MRE:
```fortran
module temp_module
    public :: myType

    type :: myType
        integer :: i = 4
        real :: r = 3.14

        contains
            private
            procedure, public, pass(self) :: get_i
    end type myType

contains
    subroutine get_i(self)
        class(myType), intent(inout) :: self
        print *, "i = ", self % i
        call get_r()

        contains
            subroutine get_r()
                print *, "r = ", self % r
            end subroutine get_r
    end subroutine get_i

end module temp_module

program test
    use temp_module
end program test
```

Output:
```console
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ gfortran test/logger/jay.f90 && ./a.out
(lf) jaysukh@Jaysukh-Ubuntu:~/Desktop/stdlib-fortran-lang$ lfortran test/logger/jay.f90
Traceback (most recent call last):
  Binary file "/home/jaysukh/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/jaysukh/Desktop/lfortran/src/bin/lfortran.cpp", line 2696, in main
    return main_app(argc, argv);
  File "/home/jaysukh/Desktop/lfortran/src/bin/lfortran.cpp", line 2624, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/jaysukh/Desktop/lfortran/src/bin/lfortran.cpp", line 1168, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
[...]

PR #7789, opened 2025-06-18
enh: Added calling a subroutine without type

sriganeshres:
Fixes: #7377


integration_tests/call_subroutine_without_type.f90
```diff
@@ -0,0 +1,23 @@
...
+            procedure, pass(self) :: get_i
+    end type myType
+
+contains
+    subroutine get_i(self)
+        class(myType), intent(inout) :: self
+        call get_r()
+        contains
+            subroutine get_r()
+                print *, "r = ", self % r
```
kmr-srbh, line 16:
```suggestion
                print *, "r = ", self % r
                if (self%r /= 1.0) error stop
```

integration_tests/call_subroutine_without_type.f90
```diff
@@ -0,0 +1,23 @@
...
+            end subroutine get_r
+    end subroutine get_i
+
+end module module_call_subroutine_without_type
+
+program call_subroutine_without_type
+    use module_call_subroutine_without_type
```
kmr-srbh, line 24: We need to call the subroutine `get_i()` to test this. Please add something like:
```suggestion
    use module_call_subroutine_without_type
    
    class(myType), allocatable :: obj
    allocate(obj)
    obj%r = 1.0
    call obj%get_i()
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3173,18 +3173,24 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        if (type->isPointerTy()) {
+                            module->getNamedGlobal(llvm_var_name)->setInitializer(
+                                llvm::ConstantPointerNull::get(static_cast<llvm::PointerType*>(type))
+                            );
+                        } else {
+                            // Avoid initializing non-pointers to null
+                            // Use UndefValue instead or don't set initializer
+                            module->getNamedGlobal(llvm_var_name)
+                                ->setInitializer(llvm::Constant::getNullValue(type));
+                        }
```
kmr-srbh:
```suggestion
                        module->getNamedGlobal(llvm_var_name)
                                ->setInitializer(llvm::Constant::getNullValue(type));
```

tests/reference/llvm-call_subroutine_without_type-297d9da.stdout
```diff
@@ -0,0 +1,78 @@
...
+  %1 = getelementptr %__vtab_mytype, %__vtab_mytype* %0, i32 0, i32 0
+  store i64 0, i64* %1, align 4
+  %2 = getelementptr %mytype_polymorphic, %mytype_polymorphic* %self, i32 0, i32 0
+  %3 = load i64, i64* %2, align 4
+  store i64 %3, i64* getelementptr inbounds (%mytype_polymorphic, %mytype_polymorphic* @__lcompilers_created__nested_context__get_i_self, i32 0, i32 0), align 4
+  %4 = getelementptr %mytype_polymorphic, %mytype_polymorphic* %self, i32 0, i32 1
+  %5 = load %mytype*, %mytype** %4, align 8
+  %6 = load %mytype, %mytype* %5, align 1
+  %7 = load %mytype*, %mytype** getelementptr inbounds (%mytype_polymorphic, %mytype_polymorphic* @__lcompilers_created__nested_context__get_i_self, i32 0, i32 1), align 8
+  store %mytype %6, %mytype* %7, align 1
```
kmr-srbh: These instructions are created due to deep copying the class variable on the assignment:

```c++
__lcompilers_created__nested_context__get_i_self = self
```
done by the `nested_vars` pass. This will lead to a segfault on the last instruction as `%mytype* %7` is null at this point. We will need to handle this deep copying correctly for such cases.

kmr-srbh (dismissed):

kmr-srbh:
Please click on "Ready for review" when ready.

src/libasr/codegen/asr_to_llvm.cpp
sriganeshres: Done.

integration_tests/call_subroutine_without_type.f90
sriganeshres: Done.

integration_tests/call_subroutine_without_type.f90
sriganeshres: Why is this needed?

kmr-srbh: The above will be replaced by the global variable in the `nested_pass`. This check will verify that the value is assigned correctly.

sriganeshres: Ok. Done.

sriganeshres:
@kmr-srbh Can you tell me how to populate ptr_type properly. As this fails on this (LLVM >= 17).

kmr-srbh:
> @kmr-srbh Can you tell me how to populate ptr_type properly. As this fails on this (LLVM >= 17).

It will be similar to https://github.com/lfortran/lfortran/pull/7788/commits/41625362e53f6a9b08771e9cf0d7f72020ae26b5. You need to find the correct place to put this block.

sriganeshres:
Ok. I tried a workaround if that doesnot work i will use that

sriganeshres:
@kmr-srbh All tests are passed.

integration_tests/call_subroutine_without_type.f90
```diff
@@ -0,0 +1,30 @@
...
+    end type myType
+
+contains
+    subroutine get_i(self)
+        class(myType), intent(inout) :: self
+        call get_r(self)  ! pass self explicitly
+        
+        contains
+            subroutine get_r(s)
+                class(myType), intent(inout) :: s
```
kmr-srbh: This is a different test than the one in the issue. We need to ensure that we get the value of `r` from the outer subroutine only. We should not declare a new dummy argument to `get_r()`.

sriganeshres:
@kmr-srbh I changed it because of no initialisation It is giving a segmentation fault without initialisation. There is no mistake in this test.

kmr-srbh:
> @kmr-srbh I changed it because of no initialisation It is giving a segmentation fault without initialisation. There is no mistake in this test.

The test of course is correct, but it is different to what we are fixing here. To test that, we need to address https://github.com/lfortran/lfortran/pull/7789#discussion_r2155206392 and https://github.com/lfortran/lfortran/pull/7789#discussion_r2154118595. You can try fixing the segfault and post questions on Zulip if you get stuck.

certik:
So the other issue now is that the transformed ASR code with the separate module with the global variable is not initialized correctly. So to fix that issue, let's print (or create by hand) the equivalent Fortran code, that has the same issue, and then let's work on fixing it directly.

sriganeshres:
@certik I will change the test to run for compile time. And i will make another issue for the runtime error. I will link that issue here as well.

sriganeshres:
refers: #7816 for runtime error.

sriganeshres:
@kmr-srbh please merge this PR. i have made a new PR for runtime. PR: #7822

sriganeshres:
@Pranavchiku can you merge this.

tests/tests.toml
```diff
@@ -4691,3 +4691,7 @@ llvm = true
 filename = "codegen_function_polymorphic.f90"
 llvm = true
 options = "-c --generate-object-code"
+
+[[test]]
+filename = "../integration_tests/call_subroutine_without_type.f90"
+llvm = true
```
Pranavchiku: does it not run? if it runs remove it from here

Pranavchiku (dismissed):
Looks good, one comment, then we can merge this

tests/tests.toml
sriganeshres: It runs. Since i have added a new change in llvm. I have added a reference to it. If it is not needed we can remove it.

sriganeshres: Removed it.

sriganeshres:
@Pranavchiku I have removed it please merge this.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1664,8 +1664,13 @@ namespace LCompilers {
...
         llvm::Type *type = nullptr;
-        if (ptr_type.find(x) != ptr_type.end()) {
-            type = ptr_type[x];
+        auto it = ptr_type.find(x);
+        if (it != ptr_type.end()) {
+            type = it->second;
+        } else if (auto* AI = llvm::dyn_cast<llvm::AllocaInst>(x)) {
+            type = AI->getAllocatedType();
+        } else if (auto* GV = llvm::dyn_cast<llvm::GlobalVariable>(x)) {
+            type = GV->getValueType();
```
certik: We should not be using `dyn_cast`, and ideally we should not be "querying" LLVM to find the type, we should know the type. Can you explain why the `ptr_type[x]` doesn't have the correct type? Let's find another solution here.

tests/tests.toml
```diff
@@ -4691,3 +4691,4 @@ llvm = true
 filename = "codegen_function_polymorphic.f90"
 llvm = true
 options = "-c --generate-object-code"
+
```
certik: Unrelated change.

certik:
I think that this looks good. I really don't like https://github.com/lfortran/lfortran/pull/7789/files#r2159634225, is there another way to do it?

src/libasr/codegen/llvm_utils.cpp
sriganeshres: Like we are some times missing those types by doing this We can find types from global as well.

tests/tests.toml
sriganeshres: I actually removed a test so this happened. It will rectify in the next pr.

sriganeshres:
> I think that this looks good. I really don't like https://github.com/lfortran/lfortran/pull/7789/files#r2159634225, is there another way to do it?

I tried diiferent ways but global type was not captured so this is the way im doing. For now lets do this if there is a good way i will refactor it.

src/libasr/codegen/llvm_utils.cpp
certik: Let's discuss at Zulip. The above fix is problematic.

certik: Here is how to fix this.

1. Put `abort()` or failing assert under `type = AI->getAllocatedType();` and print the stacktrace of the calls that lead to this
2. Higher up in the stack find a place that calls it, and instead obtain the proper LLVM type from ASR type there and call `CreateGEP2`.
3. Then do the same for `GV->getValueType();`.
4. Remove the above if statements. Instead, add an assert for `it != ptr_type.end()`.

sriganeshres: Ok. I will do it.

sriganeshres: The thing is this is not directly called it is via createLoad and create load uses creategep.

sriganeshres:
@certik I have added type extraction from `asr_to_llvm.cpp` directly and used `create_gep2` instead of `create_gep`

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2871,8 +2871,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                llvm::Type* type = nullptr;
+#if LLVM_VERSION_MAJOR <= 16
+                type = tmp->getType()->getContainedType(0);
                 tmp = llvm_utils->CreateLoad2(
-                    name2dertype[current_der_type_name]->getPointerTo(), llvm_utils->create_gep(tmp, 1));
+                    name2dertype[current_der_type_name]->getPointerTo(), llvm_utils->create_gep2(type, tmp, 1));
+#else
+                if (auto* GV = llvm::dyn_cast<llvm::GlobalVariable>(tmp)) {
+                    type = GV->getValueType();
+                }
```
certik: This is in a good direction. However we need to check ASR for the type. Does the line `type = tmp->getType()->getContainedType(0);` not work in LLVM > 17 ?

If so, it shows we should not use it. In general, we should not use `lvm::dyn_cast`, or query LLVM.

As a rule of thumb --- write code that works for all LLVM. The dispatch on LLVM <> 17 is done in create_gep2, so we shoudn't do this dispatch anywhere else. And if we do, write a utility function where internally you do the dispatch, but then just use it.

Above, you need to get the LLVM type from the ASR type, in a way that works for all LLVM versions. You might need to go one level up in the call chain.

sriganeshres: No it doesn't work in llvm > 16 as they make opaque pointers. Even functions like getpointerelement are removed. That is why we are using a map of own.

certik: Right. And we should not use the map anyway (https://github.com/lfortran/lfortran/issues/7836#issuecomment-2993664330), but that's for later.

To finish this PR, we should not use `type = tmp->getType()->getContainedType(0);` at all, nor `auto* GV = llvm::dyn_cast<llvm::GlobalVariable>(tmp)`. Either insert the correct type into the map, or pass it through the call chain (this would be ideal and that's what we need to do eventually, but I am not sure how much work is involved, we want to be able to merge this PR soon).

sriganeshres: `type = tmp->getType()->getContainedType(0);` this is the correct way to do in llvm < 16. I dont think this is wrong.

certik: This is wrong because it doesn't work for all LLVM versions. Quoting myself from above:

> As a rule of thumb --- write code that works for all LLVM. The dispatch on LLVM <> 17 is done in create_gep2, so we shoudn't do this dispatch anywhere else. And if we do, write a utility function where internally you do the dispatch, but then just use it.

sriganeshres: Ok. So you are suggesting to extract type directly from asr and do not depend on llvm versions like removing those #if. Am I right?

sriganeshres: I guess i have done it.

sriganeshres:
@certik Please have a look at new changes.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, or convert to LLVM type once and pass it through the call chain, in a local manner relative to the instruction considered.

sriganeshres:
Once check https://github.com/lfortran/lfortran/pull/7822 aswell. I will rebase it once this is merged.

sriganeshres:
@certik I guess you have to remove change requests.

PR #7795, opened 2025-06-18
feat: throw kind mismatch of `idint`function arguments in AST->ASR

parth121101:
This pull request introduces a new validation mechanism for the `Idint` intrinsic function to ensure that its arguments have the correct "kind" value.

### Validation Logic for `Idint` Function
* Added a `kind_validation` field to the `Idint` function definition in `src/libasr/intrinsic_func_registry_util_gen.py` to specify the required kind for arguments (`first` argument must have kind `8`).
* Implemented logic in `add_create_func_arg_type_src` to enforce `kind_validation` rules for specific functions by checking the kind of arguments and generating appropriate error messages.

### Code Simplification and Refactoring
* Removed redundant kind-checking logic from the `eval_Idint` and `instantiate_idint` methods in `src/libasr/pass/intrinsic_functions.h`, as this validation is now centralized in the registry.

The design structure
```
"kind_validation": [{"first":{0: 8}}],
```

Keyword `first` -> It would be used to generate appropriate error message. 
For example
`first argument must have kind 8`

The key–value pair {0: 8} indicates that the 0th argument is expected to have kind 8.

### DIFF
--- a/src/libasr/intrinsic_func_registry_util_gen.py
+++ b/src/libasr/intrinsic_func_registry_util_gen.py
@@ -1006,13 +1006,11 @@ def add_create_func_arg_type_src(func_name):
             src += 4 * indent + f'append_error(diag, "Kind of all the arguments of {func_name} must be the same", loc);\n'
             src += 4 * indent + f'return nullptr;\n'
             src += 3 * indent + '}\n'
-        
-        # Add kind validation for specific functions
         kind_validation_info = arg_info.get("kind_validation", [])
         for validation_item in kind_validation_info:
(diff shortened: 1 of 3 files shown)


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -500,6 +500,7 @@
     "Idint": [
         {
             "args": [("real",)],
+            "kind_validation": [{"first":{0: 8}}],
```
Pranavchiku, line 503:
```suggestion
            "args": [("real(8)",)],
```

PR #7798, opened 2025-06-19
enh: add `len` argument for `allocate` in fortran backend

HarshilShah1804:


Pranavchiku (changes requested):
Let's add a test here

integration_tests/allocate_23.f90
```diff
@@ -0,0 +1,10 @@
+program allocate_23
+    implicit none
+    character(len=:), allocatable :: names(:)
+    integer :: i, n
+    n = 3
+    allocate(character(len=1) :: names(n))
```
HarshilShah1804: @Pranavchiku Shall I add a test here with runtime value of length or is this fine ?

integration_tests/CMakeLists.txt
```diff
@@ -1528,6 +1528,7 @@ RUN(NAME allocate_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+RUN(NAME allocate_23 LABELS gfortran fortran)
```
certik: Don't we already have a test for this?

If not, can you also please add the `llvm` test label?

HarshilShah1804: Yes, I found the test `allocate_08.f90` testing the character arrays only. I enabled fortran backend for that test. Also, there is a test `allocate_06.f90` testing character arrays but that is failing in fortran backend due to implicit deallocate getting inserted at wrong position. I am looking for that and will enable fortran backend in that once it is fixed.

PR #7784, opened 2025-06-17
fix: handle assignment in select type where value is class

jinangshah21:
Towards #7782

PR #7805, opened 2025-06-19
test: add integration test to test subroutine call select type

### LINKED ISSUE #7773: Handle `class(*)` array passed to function in select type statement to function which expects concrete type
MRE:
```fortran
program class_optional_demo
    implicit none
    integer :: x(3)
    call print_any(x)
contains

    subroutine print_any(val)
        class(*) :: val(:)
        select type(val)
        type is (character(len=*));  call print_any2(val)
        end select
    end subroutine print_any

    subroutine print_any2(val)
        character(len=*)  :: val(:)
    end subroutine print_any2

end program class_optional_demo
```

LFortran Error:
```console
 %array.0*  call void @print_any2(%array* %array_descriptor)
```

jinangshah21:
Closes #7773 

It was solved by #7784

### DIFF
--- a/integration_tests/select_type_07.f90
+++ b/integration_tests/select_type_07.f90
@@ -9,21 +9,28 @@ program select_type_07
     end type structType
 
     type(structType) :: err
+    integer :: y = 0
 
-    call print_obj(42)
-    call print_obj(3.14)
+    call print_obj(42, y)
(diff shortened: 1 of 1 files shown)


gxyd (approved):
LGTM

PR #7813, opened 2025-06-20
Porting Dict tests

swamishiju:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -3784,12 +3784,15 @@ namespace LCompilers {
          * return value_list[pos];
          */
 
+
+        std::string key_type_code = ASRUtils::get_type_code(key_asr_type);
+        std::string value_type_code = ASRUtils::get_type_code(value_asr_type);
```
certik, line 3789: I am not happy with introducing all these strings here.

Do you understand why the `get_pointer_to_capacity_using_typecode` below requires a string type code? We are taking `key_asr_type`, converting it to string, then converting this string back to `llvm::Value`. That's inefficient. We should convert it directly to a value. It seems the current design requires it, so probably it's needed, but it doesn't seem like an optimal design.

swamishiju: Most of the new `using_typecode` functions can be easily refactored, `get_pointer_to_capacity` specifically is a bit harder because its a virtual function we are overriding. One of them uses get_list_type the other get_dict_type so refactoring it needs both of them to work with Type instead of type_codes.

Once the refactoring is complete imo it should in general have a better design with all the strings gone.

swamishiju:
> Looks good. Thanks for testing it. Are the tests exercising collisions, to ensure they work?

Just to clarify - what collisions are you referring to?

certik:
The dictionary is using linear probing, so when the key has collisions, it finds the next available slot. We should add tests for this case, to ensure the runtime code works well.

swamishiju:
> The dictionary is using linear probing, so when the key has collisions, it finds the next available slot. We should add tests for this case, to ensure the runtime code works well.

Oh yes there are some key collision tests but its not extensive

Issue #7815, opened 2025-06-20
interface operator raises AssertionFailure


gxyd:
### Minimal Reproducible Example

```f90
module julienne_test_diagnosis_m
  implicit none

  public :: operator(.and.)

  interface operator(.and.)
    elemental module function and(lhs, rhs) result(diagnosis)
      implicit none
      integer, intent(in) :: lhs, rhs
      integer diagnosis
    end function
  end interface

end module julienne_test_diagnosis_m

```

For the above Fortran program, on it's execution with LFortran and GFortran, I get the below output/error:
```console
(lf) gxyd@Gauravs-MacBook-Air OpenSource/lfortran (main) » lfortran -c test.f90
proc.first: 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 2603
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/Users/gxyd/OpenSource/lfortran/src/bin/lfortran.cpp", line 1129
    lcompilers_unique_ID = compiler_options.generate_object_code ? get_unique_ID() : "";
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 78
    if (res.ok) {
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 4184
    v.visit_TranslationUnit(ast);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 83
    visit_ast(*x.m_items[i]);
  File "/Users/gxyd/OpenSource/lfortran/src/lfortran/ast.h", line 4990
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
[...]

PR #7817, opened 2025-06-20
enh: populate all intrinsic op types in hashmap

Pranavchiku:
Fixes #7815.

PR #7820, opened 2025-06-20
fix: allow intrinsic operator overloading for "OR", "AND"

gxyd:
## Description

Fixes https://github.com/lfortran/lfortran/issues/7815

We should add more tests for other intrinsic operator overloaded functions, like the ones which are added in this PR, and they might reveal more bugs.

I'll take a look at some after this.


gxyd:
I'll close this PR in favor of https://github.com/lfortran/lfortran/pull/7817

Issue #2925, opened 2023-11-30
Typed allocation for derived types isn't recognized


difference-scheme:
It seems that typed allocation of derived types, that were extended from a base type, isn't supported yet by LFortran. The following valid code, that attempts to allocate an `array` of `class(Base)` objects, with its single element/object being of the specific type `Extended`, fails to compile:
```
module typed_allocation

   type :: Base
   end type Base
   
   type, extends(Base) :: Extended
   end type Extended
   
contains
   
   subroutine allocator(array)
      class(Base), allocatable, intent(out) :: array(:)
      allocate( Extended :: array(1) )
   end subroutine allocator

end module typed_allocation
```
LFortran's output is:
```
semantic error: `The type-spec: Extended` is not supported yet
  --> typed_allocation.f90:13:17
   |
13 |       allocate( Extended :: array(1) )
   |                 ^^^^^^^^ 


```

difference-scheme:
With the latest main branch this now compiles when using the following options:
```
lfortran -c typed_allocation.f90
```
But it fails with separate compilation, i.e. when using
```
lfortran -c --generate-object-code typed_allocation.f90
```
LFortran's output then is as follows:
```
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
LCompilersException: Not implemented class base

```
Could someone from the team put either the "separate compilation" label on it, and/or adjust the title to something that is now more appropriate?

difference-scheme:
This bug also qualifies for the OOP label. Having done more testing, I also believe that together with bug #7707 it should be given the highest priority for OOP support.

certik:
Thanks @difference-scheme, I told our team the above two issues are the highest priority.

difference-scheme:
Thanks @certik. 

Once these two bugs are fixed we'll be able to compile (including code generation) 80% of that OO application. Getting the remaining 20% to compile should then be easier (modulo the final run-time debugging, that I expect will have to be done in any case).

HarshitaKalani:
```console
$ cat a.f90
module typed_allocation

   type :: Base
   end type Base
   
   type, extends(Base) :: Extended
   end type Extended
   
contains
   
   subroutine allocator(array)
      class(Base), allocatable, intent(out) :: array(:)
      allocate( Extended :: array(1) )
   end subroutine allocator

end module typed_allocation
$ lfortran -c --generate-object-code a.f90
$ lfortran --version
LFortran version: 0.52.0-769-gda75b0e82
[...]

PR #7831, opened 2025-06-20
Support accessing derived type inside polymorphic wrapper struct

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/2925

### DIFF
--- a/integration_tests/struct_type_07.f90
+++ b/integration_tests/struct_type_07.f90
@@ -1,0 +1,29 @@
+module typed_allocation
+
+   type :: Base
+      real :: data = 2.0
+   end type Base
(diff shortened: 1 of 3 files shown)


integration_tests/struct_type_07.f90
```diff
@@ -0,0 +1,29 @@
...
+
+program test
+
+   use typed_allocation
+
+   class(Base), allocatable :: array(:)
+
+   call allocator(array)
+
+!    print *, array(1)%data
```
HarshitaKalani, line 27: This currently gives a segfault. I'm trying to understand the cause.

PR #7832, opened 2025-06-20
[OPENMP] FEAT: Add CRITICAL, BARRIER construct and enhance Shared/Private vars handling in Parallel Region

adit4443ya:
Towards #7332,


src/libasr/pass/openmp.cpp
```diff
@@ -437,17 +437,67 @@ class InvolvedSymbolsCollector:
     public ASR::BaseWalkVisitor<InvolvedSymbolsCollector>
 {
     private:
-        std::map<std::string, ASR::ttype_t*> &symbols;
+        std::map<std::pair<std::string, bool>, ASR::ttype_t*> &symbols;
```
Pranavchiku:
```suggestion
        std::map<int64_t, std::map<std::string, bool>> &variable_access_mapping;
        std::map<std::string, ASR::ttype_t*> &symbols;
```

adit4443ya: I updated it. Trying till tonight if I can push commit that can make it work with `TASK`, too (with `SHARED/PRIVATE`) . If not then i will make another PR, and then this one would be good to merge. (As this one fulfills all points as mentioned in first comment )

adit4443ya: Let's merge this, TASK would take more time i guess, Let's lock in what we have currently
CC: @Pranavchiku

Issue #7376, opened 2025-05-20
throw error for wrong format style


HarshitaKalani:
```fortran
program main
    print "(2E13)", epsilon(1e0)
end program
```
```console
$ lfortran b.f90
Segmentation fault (core dumped)
$ gfortran b.f90 && ./a.out
b.f90:2:15:

    2 |     print "(2E13)", epsilon(1e0)
      |               1
Error: Period required in format specifier E at (1)
```

HarshitaKalani:
Closing as completed in https://github.com/lfortran/lfortran/pull/7837
```
 % lfortran a.f90
Error: Invalid format descriptor E - Proper Format is E<width>.<number of digits>
Period required in format specifier
% gfortran a.f90 && ./a.out
a.f90:2:15:

    2 |     print "(2E13)", epsilon(1e0)
      |               1
Error: Period required in format specifier E at (1)
```

PR #7837, opened 2025-06-21
fix: error for wrong format style

Th3C0d3Mast3r:
## Fix for Issue #7376 

Added a small check before passing the `format`  to `handle_decimal()` or `handle_en()` function to prevent the Segmentation Fault.

After changes, the output looks like:-
image

If the changes seem proper, please do let me know, so I may add test for the same.

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -205,8 +205,8 @@ time_section "🧪 Testing stdlib (Less Workarounds)" '
   cd stdlib-fortran-lang
   export PATH="$(pwd)/../src/bin:$PATH"
 
-  git checkout n-lf-15
-  git checkout be22847d0c0d337d698b3304edd9989e9ee52b07
+  git checkout n-lf-16
+  git checkout 666e4cd4bcfdbb1f73bf1cb8e4f5bd16d9c888ea
   micromamba install -c conda-forge fypp
 
   git clean -fdx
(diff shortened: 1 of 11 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1804,6 +1804,13 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, const c
                     double val = *(double*)s_info.current_arg_info.current_arg;
                     handle_decimal(value, double_val, scale, &result, "D", is_SP_specifier);
                 } else if (tolower(value[0]) == 'e') {
+                    // check for decimal presence before passing, else leads to segmentation fault
+                    // as FORTRAN seeks for E<width>.<number of digits>
+                    if (strchr(value, '.') == NULL) {
+                        fprintf(stderr, "\033[1;31mError:\033[0m Invalid format descriptor E - Proper Format is E<width>\033[1;31m.<number of digits>\033[0m\n");
```
certik: I would not print in colors yet --- but this is a great idea and we should do that, but it needs to be configurable: https://github.com/lfortran/lfortran/issues/7838.

Th3C0d3Mast3r: have removed the colors and submitted the final one with test. Please do let me know, if anything else to be added as well?

Issue #7655, opened 2025-06-10
[stdlib:hashmaps] `intent out` attribute on a `derived type` argument should reinitialize its parameter


AnshMehta1:
MRE :- 
```.f90
module stdlib_hashmaps        
    implicit none

    type :: hashmap_type
        integer(4) :: num_entries = 0
    end type hashmap_type

end module stdlib_hashmaps


program test_open_maps

    use stdlib_hashmaps
    implicit none

    type(hashmap_type)   :: map

    call init_open_map(map)
    map % num_entries = map % num_entries + 1
    call init_open_map(map)

    print *, map % num_entries
    
contains

    subroutine init_open_map( map )
        class(hashmap_type), intent(out)      :: map
    end subroutine init_open_map

end program test_open_maps
```

GFortran Output :- 
```console
(lf) lordansh@LordAnsh:~/dev/lfort2/lfortran$ gfortran test.f90 && ./a.out
           0
```

LFortran Output :- 
```console
(lf) lordansh@LordAnsh:~/dev/lfort2/lfortran$ lfortran test.f90
1
```

kmr-srbh:
I am not sure, but barring the `allocatable` attribute this maybe related to #7530.

Issue #7530, opened 2025-05-31
bug: allocation of `type` variables ignore default initialisation


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none

   type :: val_type
      integer :: origin = 3
   end type

   type(val_type), allocatable :: val
   allocate(val)

   print *, val%origin

end program main
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
           3
(lf) saurabhkumar@Mac lfortran % lfortran test.f90           
0
```

PR #7842, opened 2025-06-21
feat: reinitialize members of StructType when intent is intent_out

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7655, Fix https://github.com/lfortran/lfortran/issues/7530


certik:
Can you please explain what exactly this PR does --- what does "reinitialize" mean? Can you give an example?

For `intent(out)` structs, we simply should pass a pointer and the struct is allocated in the caller. We should not do any kind of allocation or "reinitialization". I don't quite understand the changes, so perhaps it's doing the right thing. If you could please clarify what it was doing before, what the bug was and what it is doing now, I would appreciate it.

xaerru:
In this MRE.
```fortran
module stdlib_hashmaps        
    implicit none

    type :: hashmap_type
        integer(4) :: num_entries = 0
    end type hashmap_type

end module stdlib_hashmaps


program test_open_maps

    use stdlib_hashmaps
    implicit none

    type(hashmap_type)   :: map

    call init_open_map(map)
    map % num_entries = map % num_entries + 1
    call init_open_map(map)

    ! map%num_entries should be 0 here
    print *, map % num_entries
    
contains

    subroutine init_open_map( map )
        class(hashmap_type), intent(out)      :: map
    end subroutine init_open_map

end program test_open_maps
```

When we pass in a `StructType` to a subroutine/function where its intent is out. We reinitialize the members of that `StructType` to their default values in the definition, so any changes done to that member earlier are lost.

certik:
I was going to say that can't be the right behavior, but it looks like it's passing with GFortran. Investigating.

certik:
I see -- the `intent(out)` means the struct is initialized in the subroutine and returned. So indeed when the subroutine enters, the member must be initialized to its default value, and then returned.

What I was describing was for `intent(inout)`.

certik:
@swamishiju given this subtle behavior above, I think in LPython we should only have intent `In` and `InOut`, but not `Out`. I opened up an issue for this at https://github.com/lcompilers/lpython/issues/2858.

swamishiju:
> @swamishiju given this subtle behavior above, I think in LPython we should only have intent `In` and `InOut`, but not `Out`. I opened up an issue for this at [lcompilers/lpython#2858](https://github.com/lcompilers/lpython/issues/2858).

I'll look into that

Issue #7699, opened 2025-06-12
[stdlib: string_type] unformatted read/write for derived type causes segfault


HarshilShah1804:
MRE
```fortran
module m
    type :: mytype
        character(len=:), allocatable :: raw
    end type

    interface write(unformatted)
        module procedure :: my_write
    end interface

contains

    subroutine my_write(x, unit, iostat, iomsg)
        class(mytype), intent(in) :: x
        integer, intent(in) :: unit
        integer, intent(out) :: iostat
        character(len=*), intent(inout) :: iomsg

        ! This is what crashes LFortran
        write(unit, iostat=iostat, iomsg=iomsg) len(x%raw)
    end subroutine
end module

program test
    use m
    type(mytype) :: x
    integer :: u, iostat
    character(len=100) :: iomsg

    x%raw = "hello"

    open(newunit=u, form="unformatted", status="scratch")
    write(u, iostat=iostat, iomsg=iomsg) x
    close(u)
end program
```

LFortran 
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
Segmentation fault (core dumped)
```

MRE Works well with GFortran

Pranavchiku:
```diff
diff --git a/src/libasr/runtime/lfortran_intrinsics.c b/src/libasr/runtime/lfortran_intrinsics.c
index 08c35ca7e..fe1d015f6 100644
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -4865,6 +4865,7 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const
         filep = stdout;
     }
     if (unit_file_bin) {
+        printf("Binary file writing is not supported yet.\n");
         fseek(filep, 0, SEEK_END);
         va_list args;
         va_start(args, format);
```

Use these kind of print statements to identify exact point where it segfaults and then fix it.

PR #7844, opened 2025-06-21
fix: Fix unformatted write segfault

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/7699

### DIFF
--- a/integration_tests/write_09.f90
+++ b/integration_tests/write_09.f90
@@ -1,0 +1,33 @@
+module write_09_m
+    type :: mytype
+        character(len=:), allocatable :: raw
+    end type
+
+    interface write(unformatted)
+        module procedure :: my_write
+    end interface
+
+contains
+
+    subroutine my_write(x, unit, iostat, iomsg)
+        class(mytype), intent(in) :: x
+        integer, intent(in) :: unit
+        integer, intent(out) :: iostat
+        character(len=*), intent(inout) :: iomsg
+        write(unit, iostat=iostat, iomsg=iomsg) len(x%raw)
+    end subroutine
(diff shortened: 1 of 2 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4884,6 +4884,10 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const
             void* ptr = va_arg(args, void*);
 
             data[count].ptr = ptr;
+            if (data[count].ptr == NULL) {
+                printf("Error: NULL pointer passed to _lfortran_file_write.\n");
+                exit(1);
```
certik, line 4889: Is this a user error, or an ICE? I.e., can this be triggered by Fortran code, or it can only be triggered if we have a bug in our LLVM backend and pass a nullptr by a mistake?

If it is a user error, then let's add a test for it. If it is an ICE, then I would say so in the error ("Internal Compiler Error: NULL pointer passed to _lfortran_file_write.").

integration_tests/write_09.f90
```diff
@@ -0,0 +1,33 @@
...
+    implicit none
+    type(mytype) :: x
+    integer :: u, iostat
+    character(len=100) :: iomsg
+
+    x%raw = "hello"
+
+    open(newunit=u, form="unformatted", status="scratch")
+    write(u, iostat=iostat, iomsg=iomsg) x
+    close(u)
```
certik, line 32: Let's also add a test to read this and check that it was read correctly (which would also check that it was written correctly)?

Issue #7825, opened 2025-06-20
BinOp supports only 1 proc


Pranavchiku:
```fortran
module julienne_string_m
  use iso_c_binding, only : c_bool
  implicit none
  private
  public :: string_t
  public :: operator(.separatedBy.)

  type string_t
    private
    character(len=:), allocatable :: string_
  end type
  
  interface operator(.csv.)

    pure module function strings_with_comma_separator(strings) result(csv)
      implicit none
      type(string_t), intent(in) :: strings(:)
      type(string_t) csv
    end function

  end interface

  interface operator(.separatedBy.)
  end interface

  
contains


  module procedure strings_with_comma_separator
  end procedure


   
end module julienne_string_m


program create_markdown_table

  use julienne_string_m, only : string_t, operator(.separatedBy.)
  implicit none

contains

  pure function markdown_table(column_header) result(lines)
    type(string_t), intent(in) :: column_header(:)
    character(len=1), parameter :: column_separator = "|"
    type(string_t) lines(5)

    lines(1) = column_header .separatedBy. column_separator

  end function

end program
```

```console
% lfortran c.f90 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2653
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 1128
    result = fe.get_asr2(input, lm, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 112
    if (res.ok) {
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 5498
    b.is_body_visitor = true;
[...]

PR #7848, opened 2025-06-21
feat: Adding multiple BinOp processes

sriganeshres:
fixes: #7825

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -10414,86 +10414,104 @@ public:
         }
 
         ASR::CustomOperator_t* gen_proc = ASR::down_cast<ASR::CustomOperator_t>(op_sym);
-        LCOMPILERS_ASSERT(gen_proc->n_procs == 1)
-        ASR::symbol_t* proc;
-        if ( ASR::is_a<ASR::ClassProcedure_t>(*gen_proc->m_procs[0]) ) {
-            proc =  ASRUtils::symbol_get_past_external(
-                ASR::down_cast<ASR::ClassProcedure_t>(
-                gen_proc->m_procs[0])->m_proc);
-        } else {
-            proc = gen_proc->m_procs[0];
-        }
-        switch(proc->type) {
-            case ASR::symbolType::Function: {
-                ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(proc);
-                std::string matched_func_name = "";
-                if( func->n_args == 2 ) {
-                    ASR::ttype_t* left_arg_type = ASRUtils::expr_type(func->m_args[0]);
-                    ASR::ttype_t* right_arg_type = ASRUtils::expr_type(func->m_args[1]);
-                    if( ASRUtils::check_equal_type(left_arg_type, left_type) &&
-                        ASRUtils::check_equal_type(right_arg_type, right_type) ) {
-                        Vec<ASR::call_arg_t> a_args;
-                        a_args.reserve(al, 2);
-                        ASR::call_arg_t left_call_arg, right_call_arg;
-
-                        left_call_arg.loc = left->base.loc;
-                        left_call_arg.m_value = left;
-                        a_args.push_back(al, left_call_arg);
-
-                        right_call_arg.loc = right->base.loc;
-                        right_call_arg.m_value = right;
-                        a_args.push_back(al, right_call_arg);
-
-                        std::string func_name = to_lower(func->m_name);
-                        if( current_scope->resolve_symbol(func_name) ) {
-                            matched_func_name = func_name;
-                        } else {
-                            std::string mangled_name = func_name + "@" + std::string(x.m_op);
-                            matched_func_name = mangled_name;
-                        }
-                        ASR::symbol_t* a_name = current_scope->resolve_symbol(matched_func_name);
-                        if( a_name == nullptr ) {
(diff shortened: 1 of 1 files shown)


sriganeshres:
@Pranavchiku @certik 
This is intermediate changes but when i am running it Im getting for the linked test this as output
```
semantic error: No matching procedure found in `.def_op.` with compatible argument types
  --> test.f90:48:18
   |
48 |       lines(1) = column_header .separatedBy. column_separator
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

If this is correct direction, tell me what to fill in the implementation so that i can proceed and complete the whole implementation and  test aswell.

certik:
You can do what GFortran does.

Pranavchiku (commented):
Mark "Ready to Review" once this PR is ready.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2804,7 +2804,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                     m->m_name, nullptr, 0, gp->m_name,
                     dflt_access
                     );
-                std::string sym = to_lower(gp->m_name);
+                std::string sym = gp->m_name;
```
certik, line 2807: Is this because `gp->m_name` is already lowercase?

certik:
Thanks for finishing the PR! Can you describe the changes in `ast_common_visitor.h`? Why was the previous code incorrect and what you did to fix it.

@Pranavchiku if you could please review it, I would appreciate it!

sriganeshres:
> Thanks for finishing the PR! Can you describe the changes in `ast_common_visitor.h`? Why was the previous code incorrect and what you did to fix it.
> 

So the problem was while importing symbols they are done in lower case but when using we are not using the lowercased one. And if you can see the function `string_with_seperator` it returns a type(string_t) object. But the function has the struct which in module scope and when we use that function after importing the type in program scope. They are not matching with each other. ModuleScoped Struct in function and Programscoped struct are causing the issue. So when making the function we have changed return type struct to point the program scope with `fix_scoped_struct`. It will look into the function scope as well where we are calling the operator.

integration_tests/multiple_procs_in_bin_op.f90
```diff
@@ -0,0 +1,62 @@
...
+    module procedure strings_with_comma_separator
+    end procedure
+  
+    module procedure strings_with_separator
+    end procedure
+  
+  end module julienne_string_m_multiple_procs_in_bin_op
+  
+  
+  program create_markdown_table
```
Pranavchiku, line 46:
```suggestion
program create_markdown_table
```

Fix indentation here and elsewhere.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10583,109 +10583,105 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                proc = gen_proc->m_procs[i];
+            }
 
-                            second_call_arg.loc = second_operand->base.loc;
-                            second_call_arg.m_value = second_operand;
-                            a_args.push_back(al, second_call_arg);
-                        } else {
-                            a_args.reserve(al, 1);
-                            ASR::call_arg_t call_arg;
+            if (proc->type != ASR::symbolType::Function) {
```
Pranavchiku, line 10600:
```suggestion
            if (!ASR::is_a<ASR::Function_t>(*proc)) {
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10583,109 +10583,105 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
 
         ASR::CustomOperator_t* gen_proc = ASR::down_cast<ASR::CustomOperator_t>(operator_sym);
-        ASR::symbol_t* proc = nullptr;
-        if (ASR::is_a<ASR::ClassProcedure_t>(*gen_proc->m_procs[0])) {
-            proc = ASRUtils::symbol_get_past_external(
-                ASR::down_cast<ASR::ClassProcedure_t>(gen_proc->m_procs[0])->m_proc
-            );
-        } else {
-            proc = gen_proc->m_procs[0];
-        }
```
Pranavchiku, line 10593: Perfect, we now loop over `gen_proc->m_procs` and handle those.

src/lfortran/semantics/ast_symboltable_visitor.cpp
Pranavchiku: +1 to this, why do we need this?

Pranavchiku (approved):
rest all looks good, feel free to do in subsequent PR but address https://github.com/lfortran/lfortran/pull/7848/files#r2188039168 and then we can merge this.

sriganeshres:
@Pranavchiku should i make those changes in this pr or let it get merged then address those above issues in other prs.

src/lfortran/semantics/ast_common_visitor.h
sriganeshres: Yes.

src/lfortran/semantics/ast_symboltable_visitor.cpp
sriganeshres: yes it is already lowercase when we get to this point.

sriganeshres:
@certik Please merge this.

PR #7849, opened 2025-06-22
fix: get current type in visit_cast for select type stmts

### LINKED ISSUE #7810: Handle `class(*)` current type in `visit_cast` in backend
MRE:
```fortran
program main
    implicit none
    call handle_error(12)
contains
    subroutine handle_error(obj)
        implicit none
        class(*)  :: obj
        doubleprecision :: obj2
        integer :: valu
        select type(obj)
        type is (integer)
            obj2 = obj
        end select
    end subroutine handle_error
end program main
```

LFortran Error:
```console
SIToFP source must be integer or integer vector
  %5 = sitofp %"~abstract_type_polymorphic"* %obj to double
```

jinangshah21:
Fixes #7810

### DIFF
--- a/integration_tests/select_type_07.f90
+++ b/integration_tests/select_type_07.f90
@@ -30,19 +30,24 @@ contains
         integer, intent(out) :: y
         integer :: val1
         real :: val2
+        doubleprecision :: dbval
         character(len=:), allocatable :: val3
         type(structType) :: val4
 
         select type(x)
         type is (integer)
             val1 = x
+            dbval = x
             print *, val1
             if (val1 /= 42) error stop
+            if (dbval /= 42.0d0) error stop
             y = 1
         type is (real)
             val2 = x
+            dbval = x
             print *, val2
             if (val2 /= 3.14) error stop
+            if (dbval /= 3.14) error stop
             y = 2
         type is (character(*))
             call check_char(x, y)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8477,6 +8463,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
         }
         this->visit_expr_wrapper(x.m_arg, true);
+
+        int arg_kind = ASRUtils::extract_kind_from_ttype_t(ASRUtils::expr_type(x.m_arg));
+        if (ASRUtils::is_class_type(ASRUtils::extract_type(ASRUtils::expr_type(x.m_arg))) &&
+            ASRUtils::EXPR2VAR(x.m_arg)->m_name == current_selector_var_name) {
+            tmp = get_current_select_type(ASRUtils::expr_type(x.m_arg),
+                        tmp, module.get(), current_select_type_block_type);
+            tmp = llvm_utils->CreateLoad2(current_select_type_block_type, tmp);
+            arg_kind = llvm_utils->get_kind_from_llvm_val(tmp);
```
certik, line 8473: It would be better to extract this from ASR. Why is it different to `arg_kind`, already calculated above?

jinangshah21: Consider below example;

```fortran
class(*) :: val

select type(val) 
type is (integer) 
val2 = val
end select
```

Here in ASR `val` is a class(*) type node so inside select type we need to get kind from runtime type as from ASR node we don't know it's current selected type.

Issue #7512, opened 2025-05-30
bug: polymorphic variable not unlinked from allocated type


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none

   type :: base
      integer :: x
   end type

   type, extends(base) :: derived
      real :: r
   end type

   class(base), allocatable :: var

   allocate(derived :: var)

   select type(var)
    type is (base)
      print *, "base"
    type is (derived)
      print *, "derived"
   end select

   deallocate(var)

   select type(var)
    type is (base)
      print *, "base"
    type is (derived)
      print *, "derived"
   end select

end program main
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out 
 derived
 base
(lf) saurabhkumar@Mac lfortran % lfortran test.f90            
derived
derived
```

This is because we do not store back the old class hash into the polymorphic struct when deallocating it.

certik:
I think using `var` in select type after it was deallocated is not allowed (#8655).

certik:
@kmr-srbh do you know which 3rd party code was using the above?

kmr-srbh:
I remember I had opened this issue after I saw the implementation of polymorphic deallocation and noticed it was not resetting the class hash. This was not created from a 3rd party code.

I think this should not throw an error as the `select type` construct is based on the type of the selector. With `var` deallocated, we have it back as type `base`, which is its declared type, so `type is(base)` matches.

From Fortran 2018 implementation document, clause 7.3.2.3 CLASS type specifier, point 7:


> The dynamic type of an unallocated allocatable object or a disassociated pointer is the same as its declared type.

certik:
I see, thanks. Let me know if this assessment is correct: https://github.com/lfortran/lfortran/issues/8655#issuecomment-3362695677.

PR #7850, opened 2025-06-22
fix class deallocate and implement source argument in allocate for types

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7512

After the first commit, `dealloc_04.f90` was failing because source argument in allocate wasn't implemented in the llvm backend. Second commit implements it.


xaerru:
Before this PR, source argument in allocate was being ignored in the backend. After adding initial implementation, tests with `class(*)` were failing. Some tests in stdlib and `class_allocate_02.f90`.

I'll implement `class(*)` features in a separate PR.

certik:
So stdlib now works, but `class_allocate_02.f90` doesn't. I don't know if it is a problem that `class_allocate_02.f90` doesn't work, since it's a regression. How difficult is to fix it to run --- the test seems very simple, isn't there some quick workaround that you can add to LFortran to make it succeed, so that we can merge this PR?

This PR implements an important feature, but I don't know if that regression will break some things that used to work.

xaerru:
The test `class_allocate_02.f90` was added to check allocation status of a `class(*)` variable after an `Allocate` stmt with the source argument. It was mistakenly assumed to be working when it was added.

Before this PR, source argument in `Allocate` was being ignored in the llvm backend. In the main branch `class_allocate_02.f90` assumes `dummy` is a `StructType`, gets the size of `~abstract_type`, and calls `lfortran_malloc(0)` and stores that pointer. So it shows as allocated because the pointer is not `nullptr`.

After this PR, we skip the case of `class(*)`. So the test fails.

I think this is not a regression, it's fine to merge this.

integration_tests/dealloc_05.f90
```diff
@@ -0,0 +1,30 @@
...
+
+    select type(var)
+        type is (derived)
+        class default
+            error stop
+    end select
+
+    deallocate(var)
+
+    select type(var)
```
certik, line 24: I think this should give an error in Debug mode at runtime that `var` is used, but not allocated (#8655).

Issue #7845, opened 2025-06-21
Read function is failing when used with string slicing


akramhany:
## MRE

First create any text file and write anything in it, and change the filename in the below fortran code.

```fortran
program write_and_read_temp
    character(len=5) :: buf
    integer :: stat

    open(unit=10, file="t.txt", status='old', access="stream")

    read(10, iostat=stat) buf(1:4)

    print *, "Status code:", stat
    print *, "From file:", buf

    close(10)
end program write_and_read_temp
```

## Error

```sh
asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %7 = call i8* @_lfortran_str_slice(i8* %6, i32 0, i32 4, i32 1, i1 true, i1 true)
 i8**  call void (i8**, i32, ...) @_lfortran_read_char(i8* %7, i32 10, i32 %11)

code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %7 = call i8* @_lfortran_str_slice(i8* %6, i32 0, i32 4, i32 1, i1 true, i1 true)
 i8**  call void (i8**, i32, ...) @_lfortran_read_char(i8* %7, i32 10, i32 %11)



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

akramhany:
We can notice that the problem is that `@_lfortran_read_char` expects the first parameter to be of type `i8**`, but it's getting a type `i8*`, so I tried to convert `i8* %7` to be of type `i8**`, by the following code:
```cpp
                        if (str->getType()->isPointerTy() && str->getType()->getPointerElementType()->isIntegerTy(8)) {
                            llvm::Type *i8_ptr_ty = llvm::PointerType::getInt8PtrTy(builder->getContext());
                            llvm::Value *descriptor = llvm_utils->CreateAlloca(i8_ptr_ty);

                            builder->CreateStore(str, descriptor);
                            str = descriptor;
                        }

                        visit_StringLen(*strlen);
                        builder->CreateCall(fn, {str, unit_val, tmp}); tmp = nullptr;
```

where I added the above `if` block, and when I tried to compile the MRE again, it compiled successfully, but it is giving a wrong output now:
```sh
Status code:    0
From file:         
```

assem2002:
It's consisting of 2 major issues
[...]

Issue #7084, opened 2025-04-27
Handle str slice as read variable


jinangshah21:
MRE:
```fortran
program write_and_read_temp
    character(len=20) :: buf
    integer(8) :: stat, major, temp = 1
    character(len=20) :: filename

    filename = "tempfile.txt"

    open(unit=10, file=filename, status='replace', access="stream", form="unformatted")

    write(10) "HELLOWORLD"
    close(10)
    open(unit=10, file=filename, status='old', access="stream", form="unformatted")
    major = 2   ! Try changing this to 1

    read(10, iostat=stat) buf(1:merge(4, 2, major > 1))

    print *, "Status code:", stat
    print *, "From file:", trim(buf)

    close(10, status="delete")
end program write_and_read_temp
```

lfortran error:
```
lfortran: /home/jinang_shah/conda_root/envs/lf/include/llvm/IR/Instructions.h:1181: void llvm::ICmpInst::AssertOK(): Assertion `getOperand(0)->getType() == getOperand(1)->getType() && "Both operands to ICmp instruction are not of the same type!"' failed.
```

PR #7855, opened 2025-06-22
Fix: merge compare condition between variable and constant with read

akramhany:
Towards #7084
The next issue to solve is #7845.

### DIFF
--- a/tests/reference/llvm-read_04-581ec8b.stderr
+++ b/tests/reference/llvm-read_04-581ec8b.stderr
@@ -1,0 +1,5 @@
+code generation error: asr_to_llvm: module failed verification. Error:
+Call parameter type does not match function signature!
+  %10 = call i8* @_lfortran_str_slice(i8* %6, i32 0, i32 %9, i32 1, i1 true, i1 true)
+ i8**  call void (i8**, i32, ...) @_lfortran_read_char(i8* %10, i32 10, i32 %17)
+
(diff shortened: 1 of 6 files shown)


tests/reference/llvm-read_04-581ec8b.stderr
```diff
@@ -0,0 +1,5 @@
+code generation error: asr_to_llvm: module failed verification. Error:
+Call parameter type does not match function signature!
+  %10 = call i8* @_lfortran_str_slice(i8* %6, i32 0, i32 %9, i32 1, i1 true, i1 true)
+ i8**  call void (i8**, i32, ...) @_lfortran_read_char(i8* %10, i32 10, i32 %17)
```
certik, line 4: If the module failed verification then I think we should not be testing it. It doesn't work yet.

akramhany: I added the test only to mark the current issue, and it should be passing if we solved #7845.
Should I remove it? and if yes then how should I test the changes?

certik:
Is there a test that the 3 line change in this PR makes working (at least generates LLVM without errors)? If so, add it and we can merge it. If not, then keep the 3 line change in this PR and we'll keep it open until a more complete fix is available so that at least some test starts working.

akramhany:
I tried making a test for these 3 lines alone, but I can't seem to find one, as the problem in this issue happens when `read` is used with `merge`, but `read` itself contains many other issues than this one, so I couldn't find a separate test.
I think it would be better to leave this PR until issue #7845 is solved, then maybe we can merge this PR after that.

certik:
That works.

PR #7790, opened 2025-06-18
fix: don't create temporary variable when intent is inout or out

xaerru:
Fix https://github.com/lfortran/lfortran/issues/7562

Also uncomment a todo in `class_38.f90`.

PR #7858, opened 2025-06-23
refactor: make utility function for getting variable from var with checks

xaerru:
Address comments from https://github.com/lfortran/lfortran/pull/7790.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -345,6 +345,21 @@ static inline void set_kind_to_ttype_t(ASR::ttype_t* type, int kind) {
     }
 }
 
+static inline ASR::Variable_t* expr_to_variable_or_null(ASR::expr_t* expr) {
+    if (!expr || !ASR::is_a<ASR::Var_t>(*expr)) {
+        return nullptr;
+    }
+
+    ASR::symbol_t* sym = ASRUtils::symbol_get_past_external(
+        ASR::down_cast<ASR::Var_t>(expr)->m_v);
+
+    if (!ASR::is_a<ASR::Variable_t>(*sym)) {
+        return nullptr;
+    }
+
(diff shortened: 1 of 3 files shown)


src/libasr/asr_utils.h
```diff
@@ -345,6 +345,21 @@ static inline void set_kind_to_ttype_t(ASR::ttype_t* type, int kind) {
...
+    if (!ASR::is_a<ASR::Variable_t>(*sym)) {
+        return nullptr;
+    }
+
+    return ASR::down_cast<ASR::Variable_t>(sym);
+}
+
 static inline ASR::Variable_t* EXPR2VAR(const ASR::expr_t *f)
 {
     return ASR::down_cast<ASR::Variable_t>(symbol_get_past_external(
```
Pranavchiku, line 365: rather improve this to have checks

src/libasr/asr_utils.h
```diff
@@ -347,8 +347,18 @@ static inline void set_kind_to_ttype_t(ASR::ttype_t* type, int kind) {
...
+    }
+
+    ASR::symbol_t* sym = ASRUtils::symbol_get_past_external(
+        ASR::down_cast<ASR::Var_t>(f)->m_v);
+
+    if (!ASR::is_a<ASR::Variable_t>(*sym)) {
+        return nullptr;
+    }
+
+    return ASR::down_cast<ASR::Variable_t>(sym);
```
certik, line 360: This changes the contract what this function does. In main we assume `f` is not null and it is a Var pointing to a variable, perhaps via external, and we want to just cast it to a Variable and fail otherwise (in Debug mode), but be very fast in Release mode. I think the previous function was thus correct.

This seems like a new function with a different contract -- it will allow accepting null pointers and also Vars not pointing to Variables and it will handle it gracefully and return nullptr in that case.

So I would name it `expr_to_variable_or_null` or something like that.

PR #7873, opened 2025-06-24
Misc fixes for tuple

swamishiju:
- Added simple list of tuple test
- Removed `create_gep` for tuple functions in favor of `create_gep2`

### DIFF
--- a/src/libasr/codegen/llvm_utils.h
+++ b/src/libasr/codegen/llvm_utils.h
@@ -616,16 +616,13 @@ namespace LCompilers {
                             ASR::Tuple_t* tuple_type, llvm::Module* module,
                             std::map<std::string, std::map<std::string, int>>& name2memidx);
 
-            llvm::Value* read_item(llvm::Value* llvm_tuple, llvm::Value* pos,
+            llvm::Value* read_item(llvm::Value* llvm_tuple, ASR::Tuple_t* tuple_type,
+                                   size_t pos, bool get_pointer=false);
+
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/llvm_utils.h
```diff
@@ -616,16 +616,13 @@ namespace LCompilers {
...
-                                   bool get_pointer=false);
+            llvm::Value* read_item(llvm::Value* llvm_tuple, ASR::Tuple_t* tuple_type,
+                                   size_t pos, bool get_pointer=false);
 
-            llvm::Value* read_item_using_typecode(llvm::Type* el_type, llvm::Value* llvm_tuple, llvm::Value* pos,
+            llvm::Value* read_item(llvm::Type* el_type, llvm::Value* llvm_tuple, ASR::Tuple_t* tuple_type, llvm::Value* pos,
                                    bool get_pointer=false);
 
-            llvm::Value* read_item_using_typecode(llvm::Type* el_type, llvm::Value* llvm_tuple, size_t pos,
+            llvm::Value* read_item(llvm::Type* el_type, llvm::Value* llvm_tuple, ASR::Tuple_t* tuple_type, size_t pos,
```
certik, line 625: Why was this renamed back?

swamishiju: Those functions don't use typecodes anywhere, I could get them working purely with `ASR::ttype_t` so I thought the naming wasnt appropriate.

certik: Fine for now, but in general I prefer clear names what the function does and why you would prefer one over the other. I don't like overloading functions.

PR #7874, opened 2025-06-24
enh: implement `read` and `readwrite` argument for `inquire`

HarshilShah1804:


certik:
GFortran fails with the new test:
```
1311/1792 Test #1311: inquire_03 .........................................***Failed    0.01 sec
 NO          YES         NO          
ERROR STOP 

Error termination. Backtrace:
#0  0x7f71a4823e59 in ???
#1  0x7f71a4824a71 in ???
#2  0x7f71a482610b in ???
#3  0x55aa69c9a41e in inquire_03
```

HarshilShah1804:
@certik When I run the new test with GFortran locally, it passes. It just fails on CI. Please suggest how to fix this.

certik:
We need to debug it to understand why it fails. What version of GFortran do you use locally and on what platform. What version is being used at the CI and what platform.

Found it, it's due to https://github.com/lfortran/lfortran/pull/7874/files#r2167153933.

integration_tests/inquire_03.f90
```diff
@@ -1,8 +1,31 @@
 program inquire_03
     integer :: unit = 2
-    character(12) :: filename, specifier
+    character(12) :: filename, specifier_r, specifier_w, specifier_rw
     filename = "dummy.txt"
+
+    open(unit, file=filename, action='write')
+    inquire(unit, write=specifier_w, read=specifier_r, readwrite=specifier_rw)
+    print *, specifier_r, specifier_w, specifier_rw
+    if (specifier_r /= "No" .or. specifier_w /= "Yes ".or. specifier_rw /= "No") error stop
```
certik: GFortran seems to return NO, YES, NO, in uppercase. Does your local GFortran version return No, Yes, No?

If this is Fortran version dependent, then let's check what the standard says, and if it is case insensitive, then we need to improve this test to be case insensitive.

HarshilShah1804: Oh yes my bad. My local also returns is capital. Than I will modify it both in the `lfortran_intrinsics.c` and the test. Thank you

PR #7877, opened 2025-06-24
Tests : Add tests for C interoperability specifically for characters

assem2002:
- Test Using C functions from Fortran side
- Testing using Fortran functions from C side
- Test different cases for `c_char` characters in both sides

We're only testing them with gfortran as we have many bugs from our end. Part of the `c_char` interoperability works, But due to trying making every test comprehensive we end up with every test failing in LFortran for now.


integration_tests/character_10.f90
```diff
@@ -0,0 +1,101 @@
+
+module mm
```
certik: Use unique module name:
```suggestion
module character_10_mm
```

integration_tests/character_10c.c
```diff
@@ -0,0 +1,90 @@
...
+#include <memory.h>
+
+typedef struct {
+    void *base_addr;  // Pointer to data
+    size_t elem_len; 
+    // int rank;
+    // CFI_type_t type;
+    // int attribute;
+    // CFI_dim_t dim[CFI_MAX_RANK];
+} CFI_cdesc_t;
```
certik: This you have to include from a header, as this struct is compiler-dependent.

assem2002:
Thanks for the review. I'll resolve the comments and merge.

PR #7881, opened 2025-06-24
feat: throw type mismatch error for `min` and `max` intrinsic in `AST->ASR`

parth121101:
This pull request improves type-checking for intrinsic functions `min0` and `max0` in `src/libasr/pass/intrinsic_functions.h` and updates related test cases to ensure stricter validation of argument types in AST->ASR.

### DIFF
--- a/tests/reference/asr-continue_compilation_2-a6145a1.json
+++ b/tests/reference/asr-continue_compilation_2-a6145a1.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_2-a6145a1",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_2.f90",
-    "infile_hash": "b749230c3b25ef5b8ecf476d06112aea104ef18e68174445caba6a3c",
+    "infile_hash": "be6881cd51af5dde4cb08499496f44b18b18f90bd2b47b356f3834ff",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_2-a6145a1.stderr",
-    "stderr_hash": "8f9aa73cc6fc17846a246af8775da77dcf085c0fde8cc71761539ab3",
+    "stderr_hash": "011dcd9801e0cf0c4275776e961972454a65fd09a5372f00ed8073da",
     "returncode": 1
 }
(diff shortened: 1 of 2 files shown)


src/libasr/pass/intrinsic_functions.h
```diff
@@ -6220,10 +6236,11 @@ namespace Max {
             for (size_t i = 0; i < new_args.size(); i++) {
                 fill_func_arg("x" + std::to_string(i), ASRUtils::TYPE(ASR::make_Integer_t(al, loc, kind)));
             }
-        } else {
-            throw LCompilersException("Arguments to max0 must be of real, integer or character type");
```
certik, line 6224: Instead of adding the assert above, add it to this else branch. That way if we add more "else if" branches, we can just add them, instead of also need to modify the assert above.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6248,8 +6265,6 @@ namespace Max {
                 EXPR(ASR::make_StringLen_t(al, loc, new_args[0].m_value, int32, nullptr)),
                 ASR::string_length_kindType::ExpressionLength,
                 ASR::string_physical_typeType::PointerString));
-        } else {
-            throw LCompilersException("Arguments to max0 must be of real, integer or character type");
```
certik, line 6252: Same here.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6387,10 +6417,11 @@ namespace Min {
             for (size_t i = 0; i < new_args.size(); i++) {
                 fill_func_arg("x" + std::to_string(i), ASRUtils::TYPE(ASR::make_Integer_t(al, loc, kind)));
             }
-        } else {
-            throw LCompilersException("Arguments to min0 must be of real, integer or character type");
```
certik, line 6391: And here.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6415,8 +6446,6 @@ namespace Min {
                 EXPR(ASR::make_StringLen_t(al, loc, new_args[0].m_value, int32, nullptr)),
                 ASR::string_length_kindType::ExpressionLength,
                 ASR::string_physical_typeType::PointerString));
-        } else {
-            throw LCompilersException("Arguments to min0 must be of real, integer or character type");
```
certik, line 6419: And here.

parth121101:
> Looks good. Left some comments, you can fix it in a new PR.

sure

PR #7885, opened 2025-06-25
create copy of format_08s which follows Fortran standard

gxyd:
## Description

The test case `integration_tests/format_08.f90` isn't a valid fortran and is considered non-conforming, see what GFortran says about it: https://gcc.gnu.org/onlinedocs/gfortran/Commas-in-FORMAT-specifications.html

with that in mind, in this PR we remove `fortran` label from it from integration tests.


integration_tests/CMakeLists.txt
```diff
@@ -719,7 +719,10 @@ RUN(NAME format_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME format_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+# we might eventually need to remove "gfortran" label from format_08, once
+# we start using GFortran version 15
+RUN(NAME format_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
```suggestion
RUN(NAME format_08 LABELS llvm llvm_wasm llvm_wasm_emcc)
```

PR #7906, opened 2025-06-26
fix: use function return type for operator overloaded compare

### LINKED ISSUE #7876: Handle Struct array compare using operator overloading
MRE:
```fortran
module string_ops
  type :: string_t
  end type
  interface operator(==)
    module procedure string_array_equal
  end interface
contains
  logical function string_array_equal(a, b)
    type(string_t), intent(in), allocatable :: a(:), b(:)
  end function
end module string_ops


program test
  use string_ops
  type(string_t), allocatable :: a(:), b(:)
 if (a == b) print *, "Hi"
end program test
```

LFortran Error:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
semantic error: Expected logical expression in if statement, but recieved logical[:] instead
  --> b.f90:18:6
   |
18 |  if (a == b) print *, "Hi"
   |      ^^^^^^ logical[:] expression, expected logical
```

jinangshah21:
Fixes #7876

MRE to be handled in upcoming PR: 
```fortran
module operator_overloading_15_mod
  type :: string_t
  end type
  interface operator(==)
    module procedure string_array_equal
  end interface
contains
  logical function string_array_equal(a, b)
    type(string_t), intent(in) :: a(:), b(:)
    integer :: i
    string_array_equal = .true.
    do i = 1, size(a)
      if (a(i)%str /= b(i)%str) then
        string_array_equal = .false.
        return
      end if
    end do
  end function
end module operator_overloading_15_mod

program operator_overloading_15
  use operator_overloading_15_mod
  type(string_t), allocatable :: a(:), b(:)
  allocate(a(1), b(1))
  a(1)%str = "Hello"
  b(1)%str = "HelloWorld"
  if (a == b) error stop
end program
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1267,6 +1267,7 @@ RUN(NAME modules_59 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME modules_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_05_module3 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES
     operator_overloading_05_module1.f90 operator_overloading_05_module2.f90)
+RUN(NAME operator_overloading_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME associate_06 LABELS gfortran EXTRAFILES
         associate_06_module.f90)
 
--- a/integration_tests/operator_overloading_15.f90
+++ b/integration_tests/operator_overloading_15.f90
@@ -1,0 +1,29 @@
+module operator_overloading_15_mod
+  type :: string_t
+    character(len=:), allocatable :: str
+  end type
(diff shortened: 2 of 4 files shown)


integration_tests/template_commutative.f90
```diff
@@ -12,7 +12,8 @@ pure elemental function bin(x, y) result(bin)
         pure elemental function equal(x, y) result(equal)
             type(T), intent(in) :: x
             type(T), intent(in) :: y
-            type(T) :: equal
+            ! type(T) :: equal
+            logical :: equal
```
jinangshah21, line 16: I changed this because in `visit_Compare` in AST->ASR we were directly making type `logical` but if it is an operator overloaded function we should use its return type.

certik: That's fine --- make sense for `equal` to be logical.

certik:
Looks good. I think you need to update reference results.

gxyd (approved):
Looks good to me, thanks @jinangshah21

PR #7907, opened 2025-06-27
Refactor : Refactor intrinsic subroutine (cbind returning string)

assem2002:
### The main issue it resolves :
is that cbind function can't return `char*` so we pass a variable to receive the return (like how we do with pass `subroutine_from_function`)
***
### This refactors implementation for : 
- get_environment_variable
- data_and_time
- get_command

***
- Refactor the interface declaration return `void` instead of `char*` and received the return in some passed argument to the C function.
- Declare the interface parameter types correctly (we're concerned about how to represent `char*` in C with proper C interoperability)

### DIFF
--- a/src/libasr/asr_builder.h
+++ b/src/libasr/asr_builder.h
@@ -137,6 +137,20 @@ class ASRBuilder {
         return make_Array_t_util(al, loc, type, m_dims.p, m_dims.n);
     }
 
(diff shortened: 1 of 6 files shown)


certik:
The failure is unrelated: https://github.com/lfortran/lfortran/issues/7916

assem2002:
I think the error might be related. I'll investigate.

certik:
I restarted it and it failed again, so it might be related.

PR #7910, opened 2025-06-27
fix: pass is_module to Function ASR node correctly

gxyd:
## Description

In this PR we also raise error when the implementation of Function doesn't match the declaration of interface'd Function

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1678,7 +1678,7 @@ public:
                     ASRUtils::get_FunctionType(f2)->m_deftype == ASR::deftypeType::Interface) {
                     if (!ASRUtils::types_equal(f2->m_function_signature, func->m_function_signature)) {
                         diag.add(diag::Diagnostic(
-                            "Arguments or return var mismatch in interface and implementation",
+                            "Argument(s) or return type mismatch in interface and implementation",
                             diag::Level::Error, diag::Stage::Semantic, {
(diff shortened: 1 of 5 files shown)


tests/reference/asr-submodule_02-9152b7b.stdout
```diff
@@ -169,7 +169,7 @@
                                         ()
                                         .false.
                                         .false.
-                                        .false.
+                                        .true.
```
gxyd, line 172: this update is corresponding to `point_dist_func` declared as a *Module* function, see: https://github.com/lfortran/lfortran/blob/b8a6b659ce9f4fcf501661aa6ecc7037d833b33c/integration_tests/submodule_02.f90#L18-L21

gxyd: similarly, the other ASR changes follow the same suit.

gxyd:
A separate concern: I think we should probably try to refactor `visit_Function` and `visit_Subroutine`, that will save a lot of time fixing bugs which are needed to be separately fixed for both `Function` and `Subroutine`, and hence create double effort for us.

E.g.; we already pass `is_module` correctly for Subroutine, hence if the code is refactored to use the same code it will save hours for us trying to fix the same bugs in Function / Subroutine.

certik:
> A separate concern: I think we should probably try to refactor visit_Function and visit_Subroutine, that will save a lot of time fixing bugs which are needed to be separately fixed for both Function and Subroutine, and hence create double effort for us.

Create an issue for this.

Issue #7836, opened 2025-06-20
LLVM: get rid of all calls to CreateGEP, replace with CreateGEP2; Replace CreateLoad->CreateLoad2


certik:
Then remove the `ptr_type` structure, that is a temporary structure we created to work with LLVM 17, but after moving to CreateGEP2 it will not be needed anymore.

To fix this, one must pass in the proper LLVM type in the call chain.

certik:
Furthermore we shouldn't need the global map of pointer types (since it is fragile and it has bugs), I think we should pass the ASR (or LLVM) type down the call chain as needed.

certik:
Here is how to work on this issue:

* Create lf11 and lf20 conda environments with LLVM 11 and LLVM 20 in them.
* When you create a change, build LFortran in both lf11 and lf20 and run all integration tests. 
* That way you quickly find if there is an issue. If integration tests work in both lf11 and lf20, there is a good chance everything will work, so you submit a PR (or a commit to an existing PR) to run the full CI suite.

certik:
Current status:
```console
$ git grep -n "\<CreateGEP\>"
codegen/asr_to_llvm.cpp:1651:                llvm::Value* dt_1 = builder->CreateGEP(name2dertype[curr_struct], dt, idx_vars);
codegen/asr_to_llvm.cpp:2905:        // llvm::Value *p = CreateGEP(str, idx_vec);
codegen/asr_to_llvm.cpp:8743:        tmp = builder->CreateGEP(
codegen/llvm_utils.cpp:1362:        return LLVMUtils::CreateGEP(ds, idx_vec);
codegen/llvm_utils.cpp:1369:        return LLVMUtils::CreateGEP(ds, idx_vec);
codegen/llvm_utils.cpp:1515:    llvm::Value* LLVMUtils::CreateGEP(llvm::Value *x,
codegen/llvm_utils.cpp:1522:        return builder->CreateGEP(t2, x, idx);
codegen/llvm_utils.cpp:1530:        return builder->CreateGEP(type, x, idx);
codegen/llvm_utils.cpp:1536:        return builder->CreateGEP(t, x, idx);
codegen/llvm_utils.cpp:1964:        llvm::Value* desired_element =  builder->CreateGEP(llvm::Type::getInt8Ty(context), string_data, actual_idx);
[...]

PR #7912, opened 2025-06-27
More create_gep refactoring

swamishiju:
Towards https://github.com/lfortran/lfortran/issues/7836

- The union type stuff is irrelevant that was just backlog from another PR


swamishiju:
Its also possible a lot of the typecode stuff can be refactored into type/ttype but I'll look into that some other time

PR #7914, opened 2025-06-27
fix: don't expose private functions of module and handle scope of private

### LINKED ISSUE #7878: private function of a module is exposed via use
MRE:
```fortran
module greetings_mod
  implicit none
  private
contains
  subroutine greet(x)
    integer, intent(inout) :: x
  end subroutine
end module greetings_mod

program test
  use greetings_mod
  implicit none
contains
  subroutine greet(x)
    integer, intent(out) :: x
  end subroutine
end program
```

LFortran Error: 
```console
semantic error: Subroutine already defined greet
 --> b.f90:1:1 - 8:24
  |
1 |    module greetings_mod
  |    ^^^^^^^^^^^^^^^^^^^^...
...
  |
8 |    end module greetings_mod
  | ...^^^^^^^^^^^^^^^^^^^^^^^^
```

jinangshah21:
Fixes #7878


jinangshah21:
I have corrected many template tests as all was having private `testcall` and we were calling it outside the module, so it should be public.

jinangshah21:
I will look into failure tomorrow.

jinangshah21:
In `stdlib` there was incorrect workaround done for submodules.  I had just corrected it at https://github.com/Pranavchiku/stdlib-fortran-lang/commit/8cf4ed68b2b6f89501583358512e86d8549ce590

Issue #1162, opened 2023-01-05
ASR: Implement FunctionType node to represent a function signature


certik:
Use it to represent a function type, in callbacks like `procedure(my_funtion_type)`. Probably `Function` should just reference this `FunctionType` as well.

Some details to figure out:

* What should be part of FunctionType: probably arguments and return value (as well as type_params and restrictions ?)
* What about side-effect-free / deterministic, and other flags like inline, static, abi, access, bindc, elemental, pure, module ?
* When does a given function satisfy the type, say for a callback: arguments must exactly match. Does "elemental" have to match? What about other flags?

This FunctionType should probably be a type, like Integer.

certik:
Here is one idea to get started:
```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 450004d0a..8af10a9c0 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -84,10 +84,7 @@ symbol
         stmt* body)
     | Module(symbol_table symtab, identifier name, identifier* dependencies,
         bool loaded_from_mod, bool intrinsic)
-    | Function(symbol_table symtab, identifier name, identifier* dependencies, expr* args, stmt* body,
-        expr? return_var, abi abi, access access, deftype deftype,
-        string? bindc_name, bool elemental, bool pure, bool module, bool inline,
-        bool static, ttype* type_params, symbol* restrictions, bool is_restriction)
+    | Function(symbol_table symtab, ttype function_signature, stmt* body)
     | GenericProcedure(symbol_table parent_symtab, identifier name,
         symbol* procs, access access)
     | CustomOperator(symbol_table parent_symtab, identifier name,
@@ -98,6 +95,10 @@ symbol
     | StructType(symbol_table symtab, identifier name, identifier* dependencies,
         identifier* members, abi abi, access access, bool is_packed,
         expr? alignment, symbol? parent)
+    | FunctionType(symbol_table symtab, identifier name, identifier* dependencies, expr* args,
+        expr? return_var, abi abi, access access, deftype deftype,
+        string? bindc_name, bool elemental, bool pure, bool module, bool inline,
+        bool static, ttype* type_params, symbol* restrictions, bool is_restriction)
     | EnumType(symbol_table symtab, identifier name, identifier* dependencies,
         identifier* members, abi abi, access access, enumtype enum_value_type,
         ttype type, symbol? parent)
[...]

PR #7935, opened 2025-06-30
Union access

swamishiju:
- Added support for accessing union members


certik:
Looks good so far!

swamishiju:
I think there is some work to be done in how UnionInstanceMember is handled in the backend, ive implemented a hacky work-around in test_02 for now it should work, I'll look into the backend part.

Doing `int()` to it causes a SIGTRAP ill look into the llvm and see if I can find a fix

swamishiju:
@certik https://github.com/lfortran/lfortran/issues/1162#issuecomment-1374388817 You discussed some changes to union under ASR refactoring should I work on that? I need to look into the backend either way

certik:
I think that comment is already implemented, except the part to not link to the symbol from a type, but I would fix that later after we refactor the struct: https://github.com/lfortran/lfortran/pull/5791.

On second thought, I think it would be good to fix it, the changes should be mostly orthogonal to #5791. Yes, if you can upgrade it just like the `StructType` in #5791, that would be great.

Issue #7706, opened 2025-06-13
Refactor and fix `--realloc-lhs` (make it only work for arrays, scalars always reallocate)


certik:
@difference-scheme reported at https://github.com/lfortran/lfortran/issues/2940#issuecomment-2970101356:

But in order to get the fixed behavior one needs to compile with either the option --realloc-lhs or --std=f23 (which includes the former). I can foresee this need for specifying an extra compiler option to become a major trouble-spot for compiling object-oriented code in the future.

Such code will almost certainly make extensive use of polymorphic assignments (like the one for object adder above) and having to provide an extra compiler option for this, that the user might be unaware of, will become the source of endless future complaints. I believe  --realloc-lhs should be made the default behavior in LFortran.

certik:
We already make an exception for allocatable strings, so we can also make an exception for polymorphic assignments.

The main motivation for `--realloc-lhs` to be off by default is performance for array assignments, as well as that automatic LHS reallocation hides bugs (the user might think an array was pre-allocated correctly, but the compiler silently reallocates). I think we should thus keep `--realloc-lhs` off by default, however the following features should just work, by default:

* allocatable strings (we already do this)
* allocatable polymorphic types (both scalars, possibly also in arrays)

Another idea is to always do automatic reallocation of LHS for scalars, but not arrays. That might be a simple rule that might cover the cases above.

certik:
In either case, we have to have two modes configurable with `--realloc-lhs` and `--no-realloc-lhs`. By default we will start with `--no-realloc-lhs`, which leaves the door open to later switch to `--realloc-lhs`; the alternative is to make `--realloc-lhs` the default, but that closes the door to later switch, as it will break user's code, so we won't start with this.

No matter the default, the two modes have to be supported. In `--realloc-lhs`, every LHS gets reallocated. In `--no-realloc-lhs`, we will still reallocate strings and scalar polymorphic types. But will will not reallocate arrays. To be decided what to do with reallocatable scalars.

And we have to add checks for out of bounds and non-correctly-allocated LHS: https://github.com/lfortran/lfortran/issues/7891.

An example how GFortran handles the two modes:
```fortran
$ cat expr2.f90
program expr2
implicit none
integer, allocatable :: x(:)
integer :: y(3)
allocate(x(10))
y = 3
x = y
end program
```
output:
```console
$ gfortran -fcheck=all -g expr2.f90 && ./a.out
[...]

Issue #7891, opened 2025-06-25
Enable bounds checking, add tests


certik:
It's already implemented in `compiler_options.enable_bounds_checking`, but not exposed in the frontend, it's only tested in LPython:
```
src/bin/lpython.cpp:        app.add_flag("--enable-bounds-checking", compiler_options.enable_bounds_checking, "Turn on index bounds checking");
src/bin/lpython.cpp:        if( compiler_options.po.fast && compiler_options.enable_bounds_checking ) {
src/bin/lpython.cpp:            compiler_options.enable_bounds_checking = true;
```

Add tests. In LPython I only found 1 test: `runtime_errors/test_list_02.py`, the runtime error is:
```
IndexError: List index is out of range. Index range is (0, 2), but the given index is 3
```

We simply add `run` tests into `tests.toml`, we only want to do bounds checking with an option enabled (GFortran uses `-fcheck=all`, or a more granular option). So likely `run_bounds_checking` TOML label.

This array bounds checking will check:

* all arrays indexing out of bounds (at runtime)
* array assignment of incompatible array lengths
* assignment to unallocated array
* passing unallocated or wrong size array to a function that expects an array

In LFortran, by default it will compile in Debug mode (without `--fast`), it will enable this bounds checking.

xaerru:
For Debug mode runtime errors. I think it is best to generate code for them in `visit_Assignment()` in the llvm backend.

But the problem is the `array_op` pass replaces the assignment with a loop, then while generating llvm we don't know the LHS and RHS for checking bounds. It also inserts `Realloc` without proper checking for compiler options, we need to fix that.

For example:
```fortran
program main
   implicit none

   integer, allocatable :: x(:)
   allocate(x(3))

   x = [5, 6, 7, 8]
end program main
```
After the `array_op` pass:
```fortran
! Fortran code after applying the pass: array_op
program main
    implicit none
    integer(4) :: __1_k
    integer(4), dimension(4) :: __libasr__created__var__0__array_constant_
    integer(4) :: __libasr_index_0_
    integer(4) :: __libasr_index_0_1
    integer(4), dimension(:), allocatable :: x
    allocate(x(3))
    __1_k = 1
    __libasr__created__var__0__array_constant_(__1_k) = 5
    __1_k = 2
    __libasr__created__var__0__array_constant_(__1_k) = 6
    __1_k = 3
    __libasr__created__var__0__array_constant_(__1_k) = 7
    __1_k = 4
[...]

PR #7936, opened 2025-06-30
feat: always reallocate scalars

xaerru:
Towards https://github.com/lfortran/lfortran/issues/7706

1. Remove checks for `realloc_lhs` option in `asr_to_llvm.cpp`.
2. Rename `check_and_allocate()` to `check_and_allocate_scalar()`.
3. Array reallocation is not handled in the backend. It is done in passes.


src/bin/lfortran_command_line_parser.cpp
```diff
@@ -118,7 +118,8 @@ namespace LCompilers::CommandLineInterface {
         app.add_flag("--dump-all-passes", compiler_options.po.dump_all_passes, "Apply all the passes and dump the ASR into a file");
         app.add_flag("--dump-all-passes-fortran", compiler_options.po.dump_fortran, "Apply all passes and dump the ASR after each pass into fortran file");
         app.add_flag("--cumulative", compiler_options.po.pass_cumulative, "Apply all the passes cumulatively till the given pass");
-        app.add_flag("--realloc-lhs", compiler_options.po.realloc_lhs, "Reallocate left hand side automatically");
+        app.add_flag("--realloc-lhs", compiler_options.po.realloc_lhs, "Reallocate left hand side automatically for arrays");
+        app.add_flag("--no-realloc-lhs", compiler_options.po.no_realloc_lhs, "Reallocate left hand side automatically for scalars");
```
certik: If we have this option, it needs to be named differently, perhaps `--realloc-lhs-scalars`.

However, I would just do it automatically and not have it as an option, I think it's a good design to always do it for scalars.

Unless the documentation is incorrect.

Also, while doing this, I would rename `--realloc-lhs` to `--realloc-lhs-arrays`, to make this clear. And scalars we always reallocate. We can add `--no-realloc-lhs-arrays` to turn it off, whe it is enabled with, say, `--std=f23`. However, I think for now I wouldn't bother with the `no` option. We need to refactor all our options soon anyway, but that's a separate issue (https://github.com/lfortran/lfortran/issues/7937).

certik:
In the past I made a mistake reviewing PRs that touch `--realloc-lhs` just using the web interface, without actually manaully trying all combinations that they still work and I made a mistake and let regressions slip in.

I can't easily review this, without trying locally. I can do that later.

However, we urgently need Debug time tests for all cases that `--realloc-lhs` does not enable (https://github.com/lfortran/lfortran/issues/7891). Do you think you could please fix that issue first?

Then this PR will be very easy to review --- every possible combination of Fortran code and `--realloc-lhs` on/off is either tested in `tests` for runtime errors, or it is tested in integration tests. We can move cases between the two as we iterate on things, but that way we know there is no regression.

certik:
Let's get #7955 in first and more Debug time checks as tests. Then we'll get this PR in, as it will be easy to ensure it works as expected.

Update July 5: I made this a draft, so that I don't have it in my review queue. Once #7955 is merged, let's open this up.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7444,11 +7439,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
     // Checks if target_expr is allocated and if not then allocate
     // Used for compiler_options.po.realloc_lhs
-    void check_and_allocate(ASR::expr_t *target_expr, ASR::expr_t *value_expr = nullptr, ASR::ttype_t *value_struct_type = nullptr) {
-        LCOMPILERS_ASSERT(compiler_options.po.realloc_lhs);
+    void check_and_allocate_scalar(ASR::expr_t *target_expr, ASR::expr_t *value_expr = nullptr, ASR::ttype_t *value_struct_type = nullptr) {
         ASR::ttype_t *asr_ttype =ASRUtils::expr_type(target_expr);
         ASR::ttype_t *asr_type = ASRUtils::type_get_past_pointer(
             ASRUtils::type_get_past_allocatable(asr_ttype));
+        if (ASRUtils::is_array(asr_ttype)) {
+            throw CodeGenError("Target is an array, it should be scalar");
```
certik, line 7455: Note: This is an ICE.

PR #7939, opened 2025-06-30
doc: separate compilation as a new mode in design

Pranavchiku:
Addresses questions asked at https://github.com/lfortran/lfortran/issues/6894#issuecomment-3019694879

### DIFF
--- a/doc/src/design.md
+++ b/doc/src/design.md
@@ -83,7 +83,7 @@ The ASR is designed to have the following features:
 
 We support two compilation modes:
 
-1. Default mode: The standard compilation mode. 
+1. Monolithic Compilation Mode: The standard compilation mode. 
 In this mode, LFortran produces empty object files (`.o` files) with `-c` flag (the only reason to produce those is to satisfy existing build systems that typically expect `.o` files to be created). The object code is generated only when main program is encountered: all modules are loaded from `.mod` files and everything compiled and linked at once. When a module is compiled, only a `.mod` file is generated with full code. For files with global procedures, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates object code only for global procedures (you must thus link these generated `.o` object files with the main program), and rest of the modules are serialized to `.mod` files which do not contain global procedure.
 
 2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag. In this mode, LFortran generates full code for each file into object files (`.o` files) with full symbol information. This


doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+
+To cater to the above problems, we introduced a separate compilation mode in LFortran. In particular,
+on spotting global routines in ASR, we set `--separate-compilation` as true. Once either of the flags are set,
+LFortran will generate object files (`.o` files) with full symbol information. In the main file, it will mark all the
+global routines as `external` by defining `abi` as `ExternalUndefined` in the ASR. With this `abi`, backend will not
+generate any code for the global routines, but will generate the symbol information in the object file.
+
+Note: **_If you enabled separate compilation mode, you've to enable it for all the files in the compilation unit._**
+
+### Why two different flags?
```
Pranavchiku:
```suggestion
#### Why two different flags?
```

doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+
+2. Separate compilation mode: This mode is enabled with `--generate-object-code` flag.
+In this mode, LFortran produces object files (`.o` files) with full symbol information. This
+is default mode in other Fortran compilers such as GFortran. In this mode, LFortran
+also produces `.mod` files, which contain the ASR information for the module. User is responsbile
+to link the object files together at the end, either manually or using a build system such as CMake.
+
+### Problems with Default Mode
+
+The major problem lies in handling global routines (functions and subroutines) in the default mode.
```
Pranavchiku:
```suggestion
The major problem for this mode is handling of global routines (functions and subroutines).
```

doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+In the default mode, LFortran does not generate object files with full symbol information, which
```
Pranavchiku:
```suggestion
In this mode, LFortran does not generate object files with full symbol information, which
```

doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+means that global routines cannot be compiled and linked separately. Also, the `.mod` files generated in this mode
+do not contain any information about global routines.
+
+Also, before separate compilation, if we used `--generate-object-code` flag, LFortran produced symbols for
```
Pranavchiku:
```suggestion
Also, before separate compilation, if we used `--generate-object-code` flag, LFortran produces symbols for
```

doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+global routines in both remote and main file, which caused duplication of symbols while linking.
+
+### Separate compilation Details
+
+To cater to the above problems, we introduced a separate compilation mode in LFortran. In particular,
+on spotting global routines in ASR, we set `--separate-compilation` as true. Once either of the flags are set,
+LFortran will generate object files (`.o` files) with full symbol information. In the main file, it will mark all the
```
Pranavchiku:
```suggestion
on spotting global routines in ASR, we set `--separate-compilation` as true. In the main file, it will mark all the
```

doc/src/design.md
```diff
@@ -79,6 +79,46 @@ The ASR is designed to have the following features:
...
+global routines as `external` by defining `abi` as `ExternalUndefined` in the ASR. With this `abi`, backend will not
+generate any code for the global routines, but will generate the symbol information in the object file.
```
Pranavchiku:
```suggestion
generate any implementation code for the global routines, but will generate the symbol information in the object file.
```

certik:
Thanks. This still leaves a lot of questions:

* The default behavior is inconsistent: The `--separate-compilation` is enabled for files that have global procedures, but not for other files, which is inconsistent with "Note: **_If you enabled separate compilation mode, you've to enable it for all the files in the compilation unit._**". So either the documentation is wrong, or the design is wrong. Answer: below.
* `--separate-compilation` is thus not an option, but rather in the CompilerOption struct? Answer: yes.
* We don't need to describe the history of development. Just the current state. Is it accurate to say that the current state both direct compilation and separate compilation fully works for all cases? Answer: direct compilation works in all cases (as far as we know), the separate compilation works in all cases in main (all integration tests and all 3rd party), but there might be still some bugs that are undiscovered, and will get triggered as we compile more of fpm, we suspect related to OOP stuff. We will fix those as we create MREs for them.
*  The direct compilation uses the `--separate-compilation` flag (how confusing!!) and the separate compilation uses `--generate-object-code`? Answer: yes, fixed in #7957.
* When I say "separate compilation" I mean `--generate-object-code`, so I think the flag should be renamed to `--separate-compilation`. Answer: yes, fixed.
* The current `--separate-compilation` CompilerOption should be renamed to `generate_code_for_global_procedures` or something like that. Answer: yes, fixed.
* Only global procedures get generated, or the whole file, including all modules in the file? Answer: only global procedures get generated, all other modules only get saved to `.mod` file but no object code is generated for them, in direct mode.
* `compilation unit` is typically one file, so maybe you meant to write "all files"? Answer: yes, compilation unit=translation unit is only one file that is being compiled.

More to document:

* In separate compilation mode, we create object code, and we still create `.mod` files for modules and they contain everything just like for direct mode --- but when any `.mod` file is loaded, we change all symbols to `ExternalUndefined` ABI, then everything works. We don't change the ABI for `bind(c)` (since those are undefined already in object code, the user is responsible to provide an implementation at link time).

Pranavchiku:
> a.f90

```fortran
program separate_compilation_02
   use separate_compilation_02_module
   implicit none
   integer :: i

   interface
      subroutine sub
      end subroutine sub

      integer function fn()
      end function fn
   end interface

   call sub
   call m_sub
   i = fn()
   print *, i
   if ( i /= 19 ) error stop
end program separate_compilation_02
```

> b.f90

```fortran
module separate_compilation_02_module
contains
   subroutine m_sub
      print *, "m_sub"
   end subroutine m_sub
end module separate_compilation_02_module

subroutine sub
   print *, "sub"
end subroutine sub

integer function fn()
   use separate_compilation_02_module
   call m_sub
   fn = 19
end function fn
```


```
% lfortran -c b.f90 
% nm b.o
                 U ___module_separate_compilation_02_module_m_sub
                 U __lfortran_printf
0000000000000000 T _fn
0000000000000024 T _sub
0000000000000058 s l___unnamed_1
000000000000005a s l___unnamed_2
000000000000005e s l___unnamed_3
0000000000000000 t ltmp0
0000000000000058 s ltmp1
0000000000000068 s ltmp2
00000000000000a8 s ltmp3
% lfortran -c a.f90 
% nm a.o
0000000000000000 T ___module_separate_compilation_02_module_m_sub
                 U __lcompilers_print_error
                 U __lcompilers_string_format_fortran
                 U __lfortran_printf
                 U __lpython_call_initial_functions
                 U __lpython_free_argv
                 U _exit
                 U _fn
0000000000000034 T _main
                 U _sub
00000000000000d4 s l___unnamed_1
00000000000000d6 s l___unnamed_2
00000000000000dc s l___unnamed_3
00000000000000e1 s l___unnamed_4
00000000000000e6 s l___unnamed_5
00000000000000f6 s l___unnamed_6
00000000000000eb s l___unnamed_7
00000000000000f8 s l___unnamed_8
00000000000000e3 s l_serialization_info
0000000000000000 t ltmp0
00000000000000d4 s ltmp1
0000000000000100 s ltmp2
0000000000000140 s ltmp3
% lfortran a.o b.o
sub
m_sub
m_sub
19
```

certik:
So the answer to the first question:

* The `--separate-compilation` is a flag in `CompilerOptions`, but it is not exposed to the user as a command line option
* It is enabled in AST->ASR for a given translation unit if it contains global procedures
* The `.mod` file is still created for any module(s) in the given unit, but it odes not contain the global procedures
* The LLVM backend is invoked only for those procedures and object file is created only for those
* The user is responsible to link this file with the final program (otherwise the compilation will fail at linking with missing symbols)
* The rest of the code (besides the global procedures) is compiled in direct mode via `.mod` file at the very end.
* To answer `"Note: If you enabled separate compilation mode, you've to enable it for all the files in the compilation unit.".`: this `--separate-compilation` compilation flag (automatically set) does *not* have to be set for every compilation unit.

This is a correct design, no changes needed. We should rename the option to something more descriptive (#7957), otherwise good.

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
   construction, the ASR classes always contain valid Fortran code and the rest
   of LFortran can depend on it.
 
+## Modes:
+
+We support two compilation modes:
+
+1. Default mode: The mode in which LFortran runs by default. In this mode, LFortran
+produces empty object files (`.o` files) with `-c` flag, and everything is linked
+at the end via `.mod` files
```
Pranavchiku:
```suggestion
at the end via `.mod` files.
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
   construction, the ASR classes always contain valid Fortran code and the rest
   of LFortran can depend on it.
 
+## Modes:
+
+We support two compilation modes:
+
+1. Default mode: The mode in which LFortran runs by default. In this mode, LFortran
```
Pranavchiku:
```suggestion
1. Default mode: The standard compilation mode. In this mode, LFortran
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+Specifically for files with global procedure, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` ( not exposed to user ), which then generates
+object code only for global procedures, and rest of the modules are serialized to `.mod` files which does not contain global procedure.
+
+2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag.
+In this mode, LFortran produces object files (`.o` files) with full symbol information. This
+is default mode in other Fortran compilers such as GFortran.
+
+In this mode, we create object code, and we still create `.mod` files for modules and they contain everything just like for direct mode --- but when any `.mod` file is loaded, we change all symbols to `ExternalUndefined` ABI. We don't change the ABI for `bind(c)` (since those are undefined already in object code, the user is responsible to provide an implementation at link time).
+
+Note: **_If you enabled separate compilation mode, you've to enable it for all the files._**
```
Pranavchiku:
```suggestion
Note: **_If you enable separate compilation mode, you've to enable it for all the files._**
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
 
+## Modes:
+
+We support two compilation modes:
+
+1. Default mode: The standard compilation mode. In this mode, LFortran
+produces empty object files (`.o` files) with `-c` flag, and everything is linked
+at the end via `.mod` files.
+
+Specifically for files with global procedure, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` ( not exposed to user ), which then generates
```
Pranavchiku:
```suggestion
Specifically for files with global procedure, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+object code only for global procedures, and rest of the modules are serialized to `.mod` files which does not contain global procedure.
```
Pranavchiku:
```suggestion
object code only for global procedures, and rest of the modules are serialized to `.mod` files which do not contain global procedure.
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+
+2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag.
+In this mode, LFortran produces object files (`.o` files) with full symbol information. This
+is default mode in other Fortran compilers such as GFortran.
+
+In this mode, we create object code, and we still create `.mod` files for modules and they contain everything just like for direct mode --- but when any `.mod` file is loaded, we change all symbols to `ExternalUndefined` ABI. We don't change the ABI for `bind(c)` (since those are undefined already in object code, the user is responsible to provide an implementation at link time).
```
Pranavchiku:
```suggestion
In this mode, LFortran produces object files (`.o` files) with full symbol information. This
is default mode in other Fortran compilers such as GFortran.

We create object code, and we still create `.mod` files for modules and they contain everything just like for direct mode but when any `.mod` file is loaded, we change all symbols to `ExternalUndefined` ABI. We don't change the ABI for `bind(c)` (since those are undefined already in object code, the user is responsible to provide an implementation at link time).
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
   construction, the ASR classes always contain valid Fortran code and the rest
   of LFortran can depend on it.
 
+## Modes:
+
+We support two compilation modes:
+
+1. Default mode: The standard compilation mode. In this mode, LFortran
+produces empty object files (`.o` files) with `-c` flag, and everything is linked
```
Pranavchiku:
```suggestion
1. Default mode: The standard compilation mode. 
In this mode, LFortran produces empty object files (`.o` files) with `-c` flag, and everything is linked
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
 
+## Modes:
+
+We support two compilation modes:
+
+1. Default mode: The standard compilation mode. 
+In this mode, LFortran produces empty object files (`.o` files) with `-c` flag, and everything is linked
+at the end via `.mod` files.
+
+Specifically for files with global procedure, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates
```
Pranavchiku:
```suggestion
For files with global procedures, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+For files with global procedures, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates
+object code only for global procedures, and rest of the modules are serialized to `.mod` files which do not contain global procedure.
+
+2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag.
+In this mode, LFortran produces object files (`.o` files) with full symbol information. This
+is default mode in other Fortran compilers such as GFortran.
+
+We create object code, and we still create `.mod` files for modules and they contain everything just like for direct mode but when any `.mod` file is loaded, we change all symbols to `ExternalUndefined` ABI. We don't change the ABI for `bind(c)` (since those are undefined already in object code, the user is responsible to provide an implementation at link time).
+
+Note: **_If you enable separate compilation mode, you've to enable it for all the files._**
```
certik:
```suggestion
Note: **_If you enable separate compilation mode, you have to enable it for all the files._**
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
   construction, the ASR classes always contain valid Fortran code and the rest
   of LFortran can depend on it.
 
+## Modes:
```
certik:
```suggestion
## Compilation Modes
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+
+We support two compilation modes:
+
+1. Default mode: The standard compilation mode. 
+In this mode, LFortran produces empty object files (`.o` files) with `-c` flag, and everything is linked
+at the end via `.mod` files.
```
certik:
```suggestion
In this mode, LFortran produces empty object files (`.o` files) with `-c` flag (the only reason to produce those is to satisfy existing build systems that typically expect `.o` files to be created). The object code is generated only when main program is encountered: all modules are loaded from `.mod` files and everything compiled and linked at once. When a module is compiled, only a `.mod` file is generated with full code.
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+
+For files with global procedures, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates
+object code only for global procedures, and rest of the modules are serialized to `.mod` files which do not contain global procedure.
```
certik:
```suggestion
object code only for global procedures (you must thus link these generated `.o` object files with the main program), and rest of the modules are serialized to `.mod` files which do not contain global procedure.
```

doc/src/design.md
```diff
@@ -79,6 +79,25 @@ The ASR is designed to have the following features:
...
+
+2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag.
+In this mode, LFortran produces object files (`.o` files) with full symbol information. This
+is default mode in other Fortran compilers such as GFortran.
```
certik:
```suggestion
In this mode, LFortran generates full code for each file into object files (`.o` files) with full symbol information. This
is usually the default mode used by most other Fortran compilers.
```

Pranavchiku:
With latest push, this is how it is previewed

image

doc/src/design.md
```diff
@@ -79,6 +79,18 @@ The ASR is designed to have the following features:
   construction, the ASR classes always contain valid Fortran code and the rest
   of LFortran can depend on it.
 
+## Compilation Modes
+
+We support two compilation modes:
+
+1. Default mode: The standard compilation mode.
```
certik: Maybe this can be called "Unified compilation mode"?

I found something similar here:

* https://firefox-source-docs.mozilla.org/build/buildsystem/unified-builds.html
* https://docs.chef.io/unified_mode/

It would be good to find out how compilers like Zig call this.

Pranavchiku:
How about `Monolithic Compilation Mode`?

---

A few suggestions

| **Name**                        | **Why It Fits**                                                                            |
| ------------------------------- | ------------------------------------------------------------------------------------------ |
| **Deferred Codegen Mode**       | Code generation is deferred until main program or global procedures are encountered.       |
| **Monolithic Compilation Mode** | All code is compiled and linked in one go when the main program is seen.                   |
| **Lazy Object Mode**            | `.o` files are only generated when absolutely needed; module files are used otherwise.     |
| **Metadata-First Mode**         | Focuses on generating `.mod` files rather than object code.                                |
| **Build-Friendly Mode**         | Meant to work nicely with build systems expecting `.o` files, but without true modularity. |
| **Single-Pass Mode**            | Compilation happens in one pass at final compilation time (like Zig).                      |
| **Hybrid Compilation Mode**     | Because it uses both `.mod` files and selectively `.o` files.                              |
| **Top-Down Mode**               | Main program drives code generation; modules are just metadata unless needed.              |
| **Symbolic Compilation Mode**   | Emphasizes generating symbolic (non-object) representations like `.mod` files.             |


Inspired by zig:

| **Name**                          | **Inspired by**                                      | **Why It Fits LFortran's Default Mode**                                             |
| --------------------------------- | ---------------------------------------------------- | ----------------------------------------------------------------------------------- |
| **Single Unit Mode**              | Zig compiles everything as one unit by default       | Like Zig, LFortran defers codegen until the main program is known                   |
| **Whole Program Mode**            | Zig doesn't do per-module `.o` linking by default    | `.o` files generated only when global procs or main exist                           |
| **Unified Build Mode**            | Reflects Zig’s centralized build                     | All modules are handled together in a unified step                                  |
| **One-Shot Mode**                 | Zig doesn't recompile piecemeal                      | LFortran waits to compile all together in a single shot                             |
| **Flat Compilation Mode**         | Zig flattens the module graph                        | No separate compilation steps; everything is resolved top-down                      |
| **Lazy Linkage Mode**             | Zig defers linking logic to runtime in some cases    | LFortran produces `.o` only for what's strictly needed                              |
| **Inline Module Mode**            | Zig treats imported modules as part of the main unit | LFortran processes `.mod` files but no separate object generation unless necessary  |
| **Just-in-Time Compilation Mode** | Zig has runtime evaluation                           | LFortran waits until all pieces (main + modules) are present before generating code |
| **Graph-Resolved Mode**           | Zig builds a full dependency graph                   | LFortran collects all modules, then emits objects as needed                         |

certik:
Let's go with `Monolithic Compilation Mode`, it looks like that term is also used a bit online, so it should be understandable what it is.

doc/src/design.md
Pranavchiku:
```suggestion
1. Monolithic Compilation Mode: The standard compilation mode. 
```

doc/src/design.md
```diff
@@ -79,6 +79,18 @@ The ASR is designed to have the following features:
...
 
+## Compilation Modes
+
+We support two compilation modes:
+
+1. Monolithic Compilation Mode: The standard compilation mode. 
+In this mode, LFortran produces empty object files (`.o` files) with `-c` flag (the only reason to produce those is to satisfy existing build systems that typically expect `.o` files to be created). The object code is generated only when main program is encountered: all modules are loaded from `.mod` files and everything compiled and linked at once. When a module is compiled, only a `.mod` file is generated with full code. For files with global procedures, LFortran identifies those automatically and sets `generate_code_for_global_procedures` compiler option to `true` (not exposed to user), which then generates object code only for global procedures (you must thus link these generated `.o` object files with the main program), and rest of the modules are serialized to `.mod` files which do not contain global procedure.
+
+2. Separate compilation mode: This mode is enabled with `--separate-compilation` flag. In this mode, LFortran generates full code for each file into object files (`.o` files) with full symbol information. This
```
certik, line 89: The capitalization is inconsistent. I just pushed a fix into main directly, to save our CI: https://github.com/lfortran/lfortran/commit/4b9ddb9fd75a827e2fcf0aa6b670bcd0159fdeca.

PR #7946, opened 2025-07-01
Gep refactor

sriganeshres:
towards: #7836

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -11243,11 +11243,11 @@ public:
         llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
         llvm::Type* i64 = llvm::Type::getInt64Ty(context);
         llvm::Type* llvm_dt_type = nullptr;
-#if LLVM_MAJOR_VERSION > 16
+#if LLVM_VERSION_MAJOR > 16
         llvm_dt_type = llvm_utils->get_type_from_ttype_t_util(
             ASRUtils::extract_type(ASRUtils::expr_type(x.m_dt)), module.get());
 #else
-        llvm_dt_type = llvm_dt->getType()->getPointerElementType();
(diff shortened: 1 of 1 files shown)


sriganeshres:
@certik Im recovering this are some small tested chnages. I will be pushing more as i recover them

certik:
When you work on this issue, I recommend following this workflow: https://github.com/lfortran/lfortran/issues/7836#issuecomment-3026115185.

sriganeshres:
@certik Please merge this i  will make more prs.

Pranavchiku (approved):

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11350,19 +11358,26 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         tmp = llvm_utils->CreateAlloca(*builder, llvm_utils->get_type_from_ttype_t_util(x.m_type, module.get()));
         llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
         llvm::Type* i64 = llvm::Type::getInt64Ty(context);
+        llvm::Type* llvm_dt_type = nullptr;
+#if LLVM_VERSION_MAJOR > 16
+        llvm_dt_type = llvm_utils->get_type_from_ttype_t_util(
+            ASRUtils::extract_type(ASRUtils::expr_type(x.m_dt)), module.get());
+#else
+        llvm_dt_type = llvm_dt->getType()->getContainedType(0);
+#endif
```
certik, line 11367: This is not correct, this is a step back: it introduces an ifdef and `getContainedType`. We don't want to do that.

We want all code to be uniform and work in all versions, for example the first branch of the ifdef:
```
        llvm_dt_type = llvm_utils->get_type_from_ttype_t_util(
            ASRUtils::extract_type(ASRUtils::expr_type(x.m_dt)), module.get());
```
is good, but we need to fix things so that it works for LLVM <= 16 also.

Issue #7947, opened 2025-07-02
CI: speedup quick tests


certik:
Right now the "quick" tests run over 1h each.

For example Ubuntu LLVM 11 takes almost 1h 20 min:

https://github.com/lfortran/lfortran/actions/runs/16004203997/job/45146689548

Goal: it should run 30 min max.

The 5 longest tasks:

* 3rd-party code compilation: 25 min
* SciPy: 14 min
* Test: 8 min
* Build: 5 min
* Test Debug Build Linux: 4 min

Here is how to fix each:

* Most 3rd party tests run in few seconds (10s), but a few take many minutes. We must remove running the test suite if it takes longer than a few seconds to run. Ideal run for a testsuite is 1s. We can tolerate on the order of 10s. But not 100s.
* Move SciPy to exhaustive checks
* Move some of the many tests into exhaustive checks
* Speedup the build by moving things to cpp files, split header files
* Move some or all of these tests to exhaustive checks

More will be needed to get to 30 min, but we can start with the above.

gxyd:
Do you think disabling warnings and style suggestions when building third party repositories, help speed them up at CI?

gxyd:
On macOS, *Test (Linux/macOS)* takes around 24 minutes, e.g. https://github.com/lfortran/lfortran/actions/runs/16007365617/job/45157026128, while on Linux, it takes around 8 to 9 minutes.

certik:
> Do you think disabling warnings and style suggestions when building third party repositories, help speed them up at CI?

I doubt it, but you can try it.

gxyd:
> I doubt it, but you can try it.

Got it, I'll try that for some, my assumption was that it might take sometime to log the (style) warnings.

certik:
It should be negligible.

PR #7948, opened 2025-07-02
CI: run integration tests in parallel at CI

gxyd:
## Description

Towards: https://github.com/lfortran/lfortran/issues/7947

Locally, running this definitely speeds up `shell ci/test.sh`, let's wait and see what happens at the CI.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -39,6 +39,13 @@ src/bin/lfortran integration_tests/intrinsics_04.f90 -o intrinsics_04
 # Run all tests (does not work on Windows yet):
 cmake --version
 if [[ $WIN != "1" ]]; then
+    # using debugging option i.e. `-x` causes incorrect assignment
+    set +x
+    NPROC=$(nproc)
+    # we turn on the debugging again
+    set -x
+    echo "NPROC: ${NPROC}"
+
     if [[ $LFORTRAN_LLVM_VERSION == "11" ]]; then
         ./run_tests.py
     fi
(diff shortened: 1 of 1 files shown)


gxyd:
As mentioned in issue: https://github.com/lfortran/lfortran/issues/7947, that *Test (Linux/macOS)* takes 8 minutes 53 seconds in current *main*, with this PR (see: https://github.com/lfortran/lfortran/actions/runs/16018744496/job/45190514044?pr=7948) it takes around 5 minutes 37 seconds or 6 minutes 4 seconds. So, an improvement of around 2 to 3 minutes.

gxyd:
On macOS we've a failure with the error:
```
Error running link command: resource temporarily unavailablemake[2]: *** [max_01] Error 2
Error running link command: resource temporarily unavailablemake[2]: *** [if_02] Error 2
```

I'll try re-running, and if I get the same error, then I'll try reducing the number of parallel runners on macOS at the CI to half what I set now.

gxyd:
The tests seems to be passing, below are the timings *current main* vs *this PR*:

### Benchmark Comparison (Before vs After)

| Platform | LLVM Version | Current main     | This PR      | Improvement |
|----------|--------------|------------|------------|-------------|
| macOS    | 11           | 24m 11s    | 20m 04s    | ↓ ~4m 07s   |
| macOS    | 20           | 23m 41s    | 19m 53s    | ↓ ~3m 48s   |
| Ubuntu   | 11           | 8m 40s     | 5m 54s     | ↓ ~2m 46s   |
| Ubuntu   | 20           | 8m 29s     | 5m 34s     | ↓ ~2m 55s   |

certik:
Ok, we can try it. Does the GitHub Action VM give us more than 1 core to use?

gxyd:
@certik : https://stackoverflow.com/a/78056861 , it has been updated to 4 cores for public repositories (for ubuntu). Details are in github's documentation.

certik:
Ah, great! Based on the github documentation, it's 4 cores on Windows also.

gxyd:
> Ah, great! Based on the github documentation, it's 4 cores on Windows also.

That's right. We also have 3 on macOS (M1) as well (as we use `macos-latest`), with this PR we use 2, so if we need further improvements, we can considering setting `NPROC=3` as well (instead of `NPROC=2`).


*[EDIT]*: Sorry, I think with 3 is what we observed an error on macOS at the CI, cause `${nproc}` is 3 on it, so we can't probably use that.

Issue #7363, opened 2025-05-19
[OPENMP] TEAM Construct


adit4443ya:
Reference :- [Examples Document](https://www.openmp.org/wp-content/uploads/openmp-examples-6.0.pdf)
Page :-20

Other links: [Reference Guide](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [Specifications](https://www.openmp.org/wp-content/uploads/OpenMP-RefGuide-6.0-OMP60SC24-web.pdf), [More Examples](https://passlab.github.io/Examples/contents/Title_Page.html).

## Use Cases

- Device Offloading: Combined with the target construct to distribute work across compute units on devices like GPUs, such as streaming multiprocessors.
- TOPOLOGICAL / NESTED / Hierarchical Parallelism: Creates nested levels of parallelism, where teams handle coarse-grained tasks, and threads within teams manage fine-grained tasks., (Its is similar to that of Cartesian of processes in the MPI which helps in topological processing of computation Load).

### 1

```fortran
!$omp teams num_teams(num_teams) thread_limit(block_threads) reduction(+:sum)
    !$omp distribute
       do i0=1,N, block_size
          !$omp parallel do reduction(+:sum)
          do i = i0, min(i0+block_size,N)
             sum = sum + B(i) * C(i)
          end do
       end do
    !$omp end teams
```

Description:- Computes the dot product of arrays B and C using hierarchical parallelism. Teams distribute blocks of the loop via distribute, and threads within each team compute partial sums in parallel using parallel do.

### 2

```fortran
!$OMP TARGET MAP(TOFROM:matrix)
  !$OMP TEAMS NUM_TEAMS(4) THREAD_LIMIT(128)
  !$OMP DISTRIBUTE PARALLEL DO PRIVATE(i,j) SHARED(matrix) SCHEDULE(STATIC) COLLAPSE(2)
  do i = 1, n
    do j = 1, n
      matrix(i,j) = matrix(i,j) * real(omp_get_team_num() + 1)
    end do
  end do
  !$OMP END DISTRIBUTE PARALLEL DO
  !$OMP END TEAMS
  !$OMP END TARGET
```

Description:- Offloads a matrix scaling operation to a device (e.g., GPU). Teams distribute iterations of the nested loops, with threads in each team handling matrix elements, scaled by a team-specific factor.

### 3
```fortran
!$OMP TEAMS NUM_TEAMS(2)
[...]

PR #7951, opened 2025-07-02
FEAT: Implement TaskLoop and TEAMS constructs

adit4443ya:
Towards #7365 and #7363 and hence contributing to #7332

This PR intends to 
- Implement TaskLoop construct (For which i used the existing implementation of TASK for now as they both don't have much difference except for 1/2 clauses which are specific for taskloop which would be added if needed in future)
- Implement TEAMS construct , This encountered various blockers but i got it and got all existing clauses work with it.
- To make reduction, Shared, Private clause work in TEAMS as well as Parallel region inside Teams.


adit4443ya:
I will Add the implementation of `DISTRIBUTE` construct too in this PR to include good number of tests for `TEAMS`. Hence keeping it as draft for now, meanwhile feel free to have a review.( As all mentioned things are pretty working)

adit4443ya:
```
OMP: Warning #96: Cannot form a team with 3 threads, using 1 instead.
OMP: Hint Consider unsetting KMP_DEVICE_THREAD_LIMIT (KMP_ALL_THREADS), KMP_TEAMS_THREAD_LIMIT, and OMP_THREAD_LIMIT (if any are set).
```
this is the CI warning regarding setting some env variable, and hence is not able to set `thread_limit` dynamically and hence opting to keep default as 1, hence values are coming wrong. (It passes on my local machine)

I'm confused as well as curious as , do we use `libomp` when we say `llvm_omp` (coz it's giving warning containing some kmp, which is libomp based), and if yes then how the linking till now was done for calls of GOMP_*, as it requires `libgomp`.

@Pranavchiku @certik  would you answer this?

I checked the CI, it uses `llvm-openmp=*`, which is `libomp`

Pranavchiku:
`libomp` provides both GOMP_* and kmp_* routines. Hence we opted that. Try setting `OMP_THREAD_LIMIT`.

adit4443ya:
How do we put env variable for any test in our cmake? We can't use `EXTRA_ARGS` for that right?

I found that required env variable it is `export KMP_TEAMS_THREAD_LIMIT=32`, if we set this those error will not come and result will be correct

certik:
We would need to add it in the test suite. However, isn't there a way to rework the test, so that the environment variable is not needed?

You can also set an environment variable inside the program itself using Fortran intrinsics. I don't know if it has to be set before the program launches or not.

adit4443ya:
I checked it, there is no way to set it using fortran intrinsics (like those of `call omp_set_num_threads`) 

Also the warning it is giving says that it can max take upto 1 thread per TEAM , if i rework test to keep just 1 thread per thread, it will work but we wouldn't be able to check the parallelism inside the team, that  is, multiple threads inside a TEAM. So I insist if i could add that ENV variable then it would be good for testings

MREs like these would need multiple threads per team so that we can check that is it really working well

```
!$omp teams num_teams(2) thread_limit(5)
!$omp distribute
do i = 1, 1000, 100
    print*,omp_get_num_threads(), omp_get_max_threads()
  !$omp parallel do
  do j = i, min(i+99, 1000)
    array(j) = j * 3
  end do
  !$omp end parallel do
end do
!$omp end distribute
!$omp end teams
```

I have implemented `DISTRIBUTE` construct too , will push commit with it's tests too if this env variable can be set vi acmake somewhow

certik:
It looks like standard intrinsics indeed do not allow to set an environment variable, only to read it. Here is one way to do it on Linux:
```fortran
PROGRAM set_env
  USE ISO_C_BINDING
  IMPLICIT NONE

  INTERFACE
    FUNCTION setenv(name, value, overwrite) BIND(C, NAME='setenv')
      USE ISO_C_BINDING
      CHARACTER(KIND=C_CHAR), INTENT(IN) :: name(*), value(*)
      INTEGER(C_INT), VALUE :: overwrite
      INTEGER(C_INT) :: setenv
    END FUNCTION setenv
  END INTERFACE

  CHARACTER(KIND=C_CHAR, LEN=100) :: name = 'MY_VAR' // C_NULL_CHAR
  CHARACTER(KIND=C_CHAR, LEN=100) :: value = '42' // C_NULL_CHAR
  INTEGER(C_INT) :: overwrite = 1  ! Non-zero to overwrite existing variable
  INTEGER(C_INT) :: status

  status = setenv(name, value, overwrite)
  IF (status == 0) THEN
    PRINT *, 'Environment variable set successfully.'
  ELSE
    PRINT *, 'Failed to set environment variable.'
  END IF
END PROGRAM set_env
```

Does the above fail with gfortran also, without the environment variable set?

Can you try the above `setenv` approach and see if it works with GFortran? Then this is one way to do it.

Finally, we can add it as an option to cmake also. But I think the above approach might be better, to make the code self-sufficient.

adit4443ya:
I got your point , And tried this updated MRE(setenv) with lfortran and gfortran , (As your given code gave llvm error due to some function signature mismatch, which i do not want to indulge in right now)

<details><summary>code</summary>

```fortran
program set_env
  use iso_c_binding
  implicit none

  ! Define a new interface for setenv accepting c_ptr
  interface
    function setenv_cptr(name, value, overwrite) bind(C, name="setenv")
      import :: c_ptr, c_int
      type(c_ptr), value :: name, value
      integer(c_int), value :: overwrite
      integer(c_int) :: setenv_cptr
    end function setenv_cptr
  end interface

  ! Declare c_ptr variables
  type(c_ptr) :: var_ptr, val_ptr
  integer(c_int) :: stat
  ! Static strings with null termination
  character(kind=c_char, len=24), target :: var = 'OMP_THREAD_LIMIT'//c_null_char
  character(kind=c_char, len=3), target :: val = '32'//c_null_char

  ! Convert to c_ptr using c_loc
  var_ptr = c_loc(var)
  val_ptr = c_loc(val)

  ! Call setenv with c_ptr arguments
  stat = setenv_cptr(var_ptr, val_ptr, 1)

  if (stat == 0) then
    print *, '✅ Environment variable set successfully.'
  else
    print *, '❌ Failed to set environment variable.'
  end if
  print*,stat
end program set_env
```

</details> 

But it gave me this

<details><summary>Output</summary>

```clojure
 aditya-trivedi   lfortran    team_taskloop ≢  ?3 ~6    lfortran --openmp --openmp-lib-dir={$CONDA_PREFIX}/include --linker=gcc integration_tests/openmp_60.f90 --dump-all-passes
❌ Failed to set environment variable.
-1
 aditya-trivedi   lfortran    team_taskloop ≢  ?3 ~6    gfortran -fopenmp integration_tests/openmp_60.f90 -o a && ./a && nm a | grep GOMP_ && rm a ✅ Environment variable set successfully.
           0
```

</details> 

As this is separate issue to investigate it, i will open up issue for it.

For now to move forward/unblock we can try something like this

```
elseif (LFORTRAN_BACKEND STREQUAL "llvm_omp")
            add_executable(${name} ${file_name}.f90 ${extra_files})
            target_compile_options(${name} PUBLIC --openmp --openmp-lib-dir=$ENV{CONDA_PREFIX}/lib ${extra_args})
            target_link_options(${name} PUBLIC --openmp --openmp-lib-dir=$ENV{CONDA_PREFIX}/lib ${extra_args})
            add_test(${name} ${CURRENT_BINARY_DIR}/${name})
            set_tests_properties(${name} PROPERTIES
                ENVIRONMENT
                    "KMP_TEAMS_THREAD_LIMIT=32"
            )
```

This will just set the env for the `llvm_omp` backend, wouldn't disturb other ones.
It is working on my machine, i will push it with changes of `Distribute` construct in a while

certik:
> For now to move forward/unblock we can try something like this
> 
> ```
> elseif (LFORTRAN_BACKEND STREQUAL "llvm_omp")
>             add_executable(${name} ${file_name}.f90 ${extra_files})
>             target_compile_options(${name} PUBLIC --openmp --openmp-lib-dir=$ENV{CONDA_PREFIX}/lib ${extra_args})
>             target_link_options(${name} PUBLIC --openmp --openmp-lib-dir=$ENV{CONDA_PREFIX}/lib ${extra_args})
>             add_test(${name} ${CURRENT_BINARY_DIR}/${name})
>             set_tests_properties(${name} PROPERTIES
>                 ENVIRONMENT
>                     "KMP_TEAMS_THREAD_LIMIT=32"
>             )
> ```

Yes, try it. If it works, that will unblock us.

adit4443ya:
Pushed it, CI passing ,Ready to merge

PR #7955, opened 2025-07-03
feat: add runtime bound checking for ArrayItem and functions

xaerru:
Towards https://github.com/lfortran/lfortran/issues/7891

1. Correctly handle `realloc_lhs` in `array_op` pass.
2. Add a utility function `llvm_utils->generate_runtime_error()`.
3. Generate the runtime error if `!compiler_options.po.realloc_lhs && !compiler_options.po.fast` for finding out which integration tests give bound check runtime error.
4. Check if the same integration tests give runtime error with gfortran with `gfortran -fcheck=all -g -fno-realloc-lhs file.f90 && ./a.out`
5. If gfortran and lfortran both give bound check runtime errors then enable `--realloc-lhs` on those tests.
6. If gfortran doesn't give bound check error and lfortran does then fix those tests. These tests are commented with FIXME in `CMakeLists.txt`.
7. The change in `asr_utils.h` fixes arrays with negative lbound and ubound, haven't added a test yet.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -560,7 +560,7 @@ RUN(NAME arrays_04_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackA
 RUN(NAME arrays_05_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_06_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_07_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME arrays_08_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --no-array-bounds-checking)
+RUN(NAME arrays_08_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_09_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_10_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_11_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
@@ -583,7 +583,7 @@ RUN(NAME arrays_01_real LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackA
 RUN(NAME arrays_01_complex LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
 RUN(NAME arrays_01_logical LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm)
 RUN(NAME arrays_01_multi_dim LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
-RUN(NAME integer_bin_op_dim_external_module LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray EXTRA_ARGS --no-array-bounds-checking)
+RUN(NAME integer_bin_op_dim_external_module LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME array_bound_1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
 RUN(NAME array_bound_2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
 RUN(NAME array_bound_3 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran)
@@ -665,11 +665,11 @@ RUN(NAME arrays_44 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran
 RUN(NAME arrays_45 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_46 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_47 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME arrays_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
+RUN(NAME arrays_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_50 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES arrays_50_mod.f90)
 RUN(NAME arrays_51 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME arrays_52 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
+RUN(NAME arrays_52 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_53 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_54 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_55 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
@@ -731,7 +731,8 @@ RUN(NAME array_section_is_non_allocatable LABELS gfortran llvm llvm_wasm llvm_wa
 RUN(NAME array_indices_array_section LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_indices_array_section_assignment LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_constructor_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME array_constructor_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+# FIXME: fails with bounds checking
+RUN(NAME array_constructor_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --no-array-bounds-checking)
 RUN(NAME allocatble_c_ptr LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME hashmap_struct_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME hashmap_nested_dealloc_derived_pointer LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
@@ -796,8 +797,7 @@ RUN(NAME format_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME format_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME format_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-# FIXME: Make this work with --fast for llvm >= 17, https://github.com/lfortran/lfortran/issues/8155
-RUN(NAME format_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST)
+RUN(NAME format_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME format_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME format_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
@@ -988,8 +988,7 @@ RUN(NAME intrinsics_144 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_145 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # dot_product
 RUN(NAME intrinsics_146 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # dprod
 RUN(NAME intrinsics_147 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # pack
-# FIXME: Fails with array bounds checking. Multi dimensional ArrayConstant is not supported.
-RUN(NAME intrinsics_148 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking) # pack
+RUN(NAME intrinsics_148 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # pack
 RUN(NAME intrinsics_149 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # unpack
 RUN(NAME intrinsics_150 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # maskr
 RUN(NAME intrinsics_151 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # maskl
@@ -1243,10 +1242,10 @@ RUN(NAME integer_boz_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME test_dshiftr LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
-RUN(NAME passing_array_01 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
-RUN(NAME passing_array_02 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
-RUN(NAME passing_array_03 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
-RUN(NAME passing_array_04 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --no-array-bounds-checking)
+RUN(NAME passing_array_01 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME passing_array_02 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME passing_array_03 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME passing_array_04 LABELS gfortran fortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME parameter_01 LABELS gfortran llvm)
 RUN(NAME parameter_02 LABELS gfortran llvm)
--- a/integration_tests/arrays_op_20.f90
+++ b/integration_tests/arrays_op_20.f90
@@ -3,7 +3,7 @@ implicit none
 
 real, allocatable :: array(:, :), arrayoutput(:, :)
 
-allocate(array(5, 5))
+allocate(array(3, 3))
 arrayoutput = f(5, array)
 print *, size(arrayoutput)
 if( size(arrayoutput) /= 24 ) error stop
@@ -12,7 +12,7 @@ contains
 
 function f(m, input) result(output)
 integer :: m
-real :: input(m, m)
+real :: input(m)
 real :: output(2:m, m:2*m)
 end function
 
--- a/integration_tests/fortuno_01.f90
+++ b/integration_tests/fortuno_01.f90
@@ -7,7 +7,7 @@ module fortuno_basetypes
   
   
     type :: test_ptr_item
-        class(test_base), pointer :: item
+      class(test_base), pointer :: item => null()                                 
     end type test_ptr_item
   
   
@@ -25,9 +25,6 @@ module fortuno_basetypes
     subroutine test_list_free(this)
       class(test_list), intent(inout) :: this                                     
   
-      type(test_ptr_item), target :: d(1)
-      this%storage_ => d
-
       select type (item => this%storage_(1)%item)                                 
       class default
       end select


xaerru:
The test `integration_tests/arrays_08_size.f90` doesn't give bound check error with gfortran but it does with lfortran and flang-new.

MRE of `arrays_08_size.f90` for bound check error:
```fortran
program arrays_08_size
   integer :: y = 2
   call temp(y)

contains
   subroutine temp(x)
        integer, intent(inout) :: x
        integer :: result(x)
        x = 1
        ! This assignment is not valid, results size is 2 and trueloc returns an array of size 1
        result = trueloc()
        print * , size(result)
   end subroutine

   function trueloc() result(loc)
        integer(4), allocatable :: loc(:)
        allocate(loc(1))
   end function trueloc
end program
```
```shell
(lf)
[xaerru@nienna:~/projects/lfortran/bound-check]$ ./src/bin/lfortran a.f90
Runtime error: Array '__libasr_created__function_call_trueloc' index out of bounds.

Tried to access index 2 of dimension 1, but valid range is 1 to 1.
(lf)
[xaerru@nienna:~/projects/lfortran/bound-check]$ flang-new a.f90 && ./a.out

fatal Fortran runtime error(/home/xaerru/projects/lfortran/bound-check/a.f90:10): Assign: mismatching element counts in array assignment (to 2, from 1)
Aborted (core dumped)
(lf)
[xaerru@nienna:~/projects/lfortran/bound-check]$ gfortran -fcheck=all -g -fno-realloc-lhs a.f90 && ./a.out
           2
```

This seems to be a problem with gfortran. We should apply the following diff to this test to correct it.
```diff
diff --git a/integration_tests/arrays_08_size.f90 b/integration_tests/arrays_08_size.f90
index 28f30df45..a91b2b18e 100644
--- a/integration_tests/arrays_08_size.f90
+++ b/integration_tests/arrays_08_size.f90
@@ -6,7 +6,7 @@ contains
    subroutine temp(x)
         integer, intent(inout) :: x
         logical :: keep(x)
-        integer :: result(x)
+        integer, allocatable :: result(:)
         keep = [.false., .true.]
         x = 1
         result = trueloc(keep)
```

xaerru:
Now only `integration_tests/intrinsics_288.f90` fails.

MRE is
```fortran
program intrinsics_288
    integer, dimension(3, 2) :: result

    result = spread([1, 4, 5], 2, 2) ! [[1, 1], [4, 4], [5, 5]]
    print *, result
end program
```

The `IntrinsicArrayFunction` is evaluated at compile time and stored as an `ArrayConstant`.
```clojure
  (ArrayConstant
      24
      [1, 4, 5, 1, 4, 5]
      (Array
          (Integer 4)
          (IntegerConstant 6 (Integer 4) Decimal))]
          FixedSizeArray
      )
      ColMajor
  )
```

The dimensions should be [(1, 3), (1, 2)] but they are [(1, 6)], this is because we don't store multi-dimensional arrays in `ArrayConstant`.

So the assignment in the MRE is replaced by the following after the `array_op` pass.
```fortran
integer(4), dimension(3, 2) :: result
result(1) = 1
result(2) = 4
result(3) = 5
result(4) = 1
result(5) = 4
result(6) = 5
```

So we throw a runtime error.
```shell
Runtime error: Array 'result' index out of bounds.

Tried to access index 4 of dimension 1, but valid range is 1 to 3.
```

This works in main because in the llvm backend we support `ArrayItem` assignment in `ColMajor` order with only a single index like the pass generates. But with bounds checking we throw a runtime error.

There are 2 solutions I could think of:
1. We add a reshape outside of `spread` by default. This is simple but we can't reshape at compile time.
2. We allow `ArrayConstant` to have have multi-dimensional arrays. This would be complicated and would require modifications to `array_op` and also `ArrayConstructor` node which flattens `ArrayConstant`.

certik:
Yes, https://github.com/lfortran/lfortran/pull/7955#issuecomment-3035217649 I think should trigger a runtime bounds error.

Regarding `ArrayConstant`, I think we need to extend it to work for any multidimensional array if we assign it to a multidimensional array. Otherwise we need to reshape it.

As a short-term workaround to unblock this PR, you can add a special case for this and skip a bounds check.

ci/test_third_party_codes.sh
```diff
@@ -779,7 +779,7 @@ time_section "🧪 Testing fastGPT" '
 
         cd lf-fast
         git clean -dfx
-        FC="$FC --fast" CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS -DCMAKE_BUILD_TYPE=Release ..
+        FC="$FC --realloc-lhs --fast" CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS -DCMAKE_BUILD_TYPE=Release ..
```
certik: Note that fastGPT does not need `--realloc-lhs`. It allocates all arrays explicitly.

First get the PR to pass CI, then we can look at this why this is needed.

xaerru: It is related to the incorrect checking of `realloc_lhs` flag in the `array_op` pass.

After this diff in main, fastGPT throws a segfault.
```diff
diff --git a/src/libasr/pass/array_op.cpp b/src/libasr/pass/array_op.cpp
index 24d3a31b5..ce83bd94f 100644
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -855,7 +855,8 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
     void insert_realloc_for_target(ASR::expr_t* target, ASR::expr_t* value, Vec<ASR::expr_t**>& vars) {
         ASR::ttype_t* target_type = ASRUtils::expr_type(target);
         bool array_copy = ASR::is_a<ASR::Var_t>(*value) && ASR::is_a<ASR::Var_t>(*target);
-        if( (realloc_lhs == false || !ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
+        if (!realloc_lhs) return;
+        if( (!ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
             !(array_copy && ASRUtils::is_allocatable(target_type)) ) {
             return ;
         }
```

certik:
TODO:
* fix macos
* Reproduce ubuntu llvm 11 issue and fix.

xaerru:
I've added bounds checking when arrays are passed into functions. Bound checking is generated in the llvm backend when visiting `FunctionCall` or `SubroutineCall`.  If there is an `ArrayPhysicalCast` node in the arguments, we take its type and expect the array to have those dimensions.

I've added two new visitors for dealing with arrays whose dimensions depend on input arguments or other symbols:
1. `ReplaceFunctionParamWithArg`
This visitor replaces `FunctionParam` with the particular argument in that `FunctionCall` or `SubroutineCall`.
2. `CheckSymbolReplacer`
This visitor is for checking whether the symbols in the dimension expression can be accessed through the scope of the `FunctionCall` or `SubroutineCall`. 

Some cases which require the second visitor:
```fortran
pure integer function fn1() result(r)
    implicit none
    r = 1
end function

subroutine my_sub(res)
    implicit none
    real, intent(out) :: res(fn1())

    res = [1]
end subroutine

program functions_22
    implicit none
    real :: res(1)

    interface
        subroutine my_sub(res)
            implicit none

            interface
                pure integer function fn1()
                    implicit none
                end function
            end interface

            real, intent(out) :: res(fn1())
        end subroutine
    end interface

    ! fn1() interface is not available here so do not set dimension
    call my_sub(res)
end program
```
```fortran
module module_arrays_47
    integer :: nx = 4
    integer :: ny = 3
    integer :: nz = 4
contains

    subroutine a(cs)
        real, dimension(nx, ny, nz), intent(in) :: cs
        print *, sum(cs)
        if (abs(sum(cs) - 6343.99072) > 1e-8) error stop
    end subroutine
end module module_arrays_47

program arrays_47
    use module_arrays_47
    real, dimension(4, 4, 4) :: cs
    cs = 99.1248
    ! Replace nx, ny, nz symbols with corresponding external symbols here
    call a(cs)
end program
```

Should we keep this visitor or ignore the dimensions when they have symbols in them?
In main we currently ignore the dimensions if they are not compile time constants.

certik:
It looks like one solution is to evaluate the expression for the dimension, and if it contains a function call `fn1`, then just call it, in the caller. That gives you an LLVM value for the dimension that the function itself expects at runtime. Then compare this value (at runtime) with the actual size of the array, as computed at the call site --- if there is a physical cast, simply treat it as a special case (similar to https://github.com/lfortran/lfortran/issues/8005) and go to the inner expression, which typically will be some kind of a temporary array, and extract the size of it, from its descriptor, or otherwise (perhaps using ArraySize or whatever it takes).

Note: all physical casts always have to have the dimension filled in, except one case: a physical cast as an argument to a function call. In that case we don't always know how to fill in the dimensions, so we don't. Then we have a special handling of this case in the backend. We should enforce this by verify() that all casts are always well-formed including the dimensions, except this one special case.

certik:
Looks like there are two failures:
```
<ERROR> Execution for object " splpak_test_linear " returned exit code  11
<ERROR> Execution for object " splpak_test " returned exit code  11
STOP 11
```
and
```
=== DYNAMIC Schedule Test ===
Runtime error: Array 'thread_iterations' index out of bounds.

Tried to access index 0 of dimension 1, but valid range is 1 to 8.

      Start 79: openmp_68
      Start 80: openmp_69
74/80 Test #76: openmp_65 ........................***Failed    0.02 sec
```

src/libasr/utils.h
```diff
@@ -114,7 +114,8 @@ struct CompilerOptions {
     bool disable_style = false;
     bool logical_casting = false;
     bool no_error_banner = false;
-    bool enable_bounds_checking = false;
+    bool enable_bounds_checking = true;
+    bool disable_bounds_checking = false;
```
certik: Remove the `disable_bounds_checking`, and maybe call `enable_bounds_checking` just `bounds_checking`?

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -161,6 +161,8 @@ namespace LCompilers::CommandLineInterface {
         app.add_flag("--realloc-lhs", compiler_options.po.realloc_lhs, "Reallocate left hand side automatically")->group(group_miscellaneous_options);
         app.add_flag("--ignore-pragma", compiler_options.ignore_pragma, "Ignores all the pragmas")->group(group_miscellaneous_options);
         app.add_flag("--stack-arrays", compiler_options.stack_arrays, "Allocate memory for arrays on stack")->group(group_miscellaneous_options);
+        app.add_flag("--array-bounds-checking", compiler_options.enable_bounds_checking, "Enables runtime array bounds checking")->group(group_miscellaneous_options);
```
certik:
```suggestion
        app.add_flag("--array-bounds-checking", compiler_options.bounds_checking, "Enables runtime array bounds checking")->group(group_miscellaneous_options);
```

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -161,6 +161,8 @@ namespace LCompilers::CommandLineInterface {
...
+        app.add_flag("--no-array-bounds-checking", compiler_options.disable_bounds_checking, "Disables runtime array bounds checking")->group(group_miscellaneous_options);
```
certik:
```suggestion
        app.add_flag("--no-array-bounds-checking", disable_bounds_checking, "Disables runtime array bounds checking")->group(group_miscellaneous_options);
```
And declare `bool disable_bounds_checking` as a local variable.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -238,6 +240,10 @@ namespace LCompilers::CommandLineInterface {
             );
         }
 
+        if (compiler_options.disable_bounds_checking) {
+            compiler_options.enable_bounds_checking = false;
+        }
```
certik:
```suggestion
        if (disable_bounds_checking) {
            compiler_options.bounds_checking = false;
        }
```

ci/test_third_party_codes.sh
```diff
@@ -712,7 +712,7 @@ time_section "🧪 Testing fastGPT" '
         git checkout -t origin/lf6
         git checkout bc04dbf476b6173b0bb945ff920119ffaf4a290d
         echo $CONDA_PREFIX
-        FC=$FC CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS .
+        FC="$FC --realloc-lhs --no-array-bounds-checking" CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS .
```
certik: This must compile without `--realloc-lhs`, we are allocating everything explicitly in fastGPT.

integration_tests/arrays_08_size.f90
```diff
@@ -6,7 +6,7 @@ program arrays_08_size
    subroutine temp(x)
         integer, intent(inout) :: x
         logical :: keep(x)
-        integer :: result(x)
+        integer, allocatable :: result(:)
```
certik: This changes the test. Add a new test instead.

integration_tests/array_indices_array_item.f90
```diff
@@ -52,8 +52,8 @@ program array_indices_array_item
     print *, rank_val
     if (rank_val /= 1) error stop
 
-    allocate(arr(1, 3))
-    arr = reshape([1, 2, 3], shape(arr))
+    allocate(arr(2, 3))
+    arr = reshape([1, 2, 3, 4, 5, 6], shape(arr))
```
certik: This changes the test. Add a new test or add more lines here.

integration_tests/arrays_51.f90
```diff
@@ -2,7 +2,7 @@
 module arrays_51_mod
     IMPLICIT NONE
     integer :: i = 4
-    integer :: arr_mod(4,4)
+    integer :: arr_mod(4,5000)
```
certik: This is a big array, let's keep the original?

certik:
I have two concerns:

* We should not modify existing tests, unless they are wrong. Instead, add new tests or extend the current ones to test more cases.
* We should not require `--realloc-lhs` for fastGPT and perhaps other codes. This seems to suggest that some logic was changed regarding this that shouldn't have.

Let's resolve these two issues, then I'll review again.

If this change is too complex, then let's split this PR into smaller pieces that we can get in one by one.

src/libasr/pass/array_op.cpp
```diff
@@ -850,7 +855,10 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
         ASR::ttype_t* target_type = ASRUtils::expr_type(target);
         bool array_copy = ASR::is_a<ASR::Var_t>(*value) && ASR::is_a<ASR::Var_t>(*target);
-        if( (realloc_lhs == false || !ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
+        if (!realloc_lhs) {
+            return;
+        }
+        if( (!ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
             !(array_copy && ASRUtils::is_allocatable(target_type)) ) {
             return ;
         }
```
xaerru: This is the change related to `--realloc-lhs`. Old code inserts a realloc before the assignment loop even if `--realloc-lhs` is not enabled.

On this code:
```fortran
program main
   implicit none

   integer, allocatable :: x(:)
   allocate(x(3))

   x = [5, 6, 7, 8]
end program main
```
After the array op pass, we get a reallocate statement without passing in `--realloc-lhs`.
```fortran
! Fortran code after applying the pass: array_op
program main
implicit none
integer(4) :: __1_k
integer(4), dimension(4) :: __libasr__created__var__0__array_constant_
integer(4), dimension(:), allocatable :: x
allocate(x(3))
__1_k = 1
__libasr__created__var__0__array_constant_(__1_k) = 5
__1_k = 2
__libasr__created__var__0__array_constant_(__1_k) = 6
__1_k = 3
__libasr__created__var__0__array_constant_(__1_k) = 7
__1_k = 4
__libasr__created__var__0__array_constant_(__1_k) = 8
reallocate(x(size(__libasr__created__var__0__array_constant_, 1)))
__libasr_index_0_1 = lbound(__libasr__created__var__0__array_constant_, 1)
do __libasr_index_0_ = lbound(x, 1), ubound(x, 1)
    x(__libasr_index_0_) = __libasr__created__var__0__array_constant_(__libasr_index_0_1)
end do
end program main
```

The new change removes this reallocate statement. But we can't detect the bounds error here yet. I plan to do that after this PR, it will require adding a new node `DebugCheckArrayBounds` before the `Assignment` node.

I can revert this change and include it in the next PR, but it will cause some `ArrayItem` tests to not give a runtime error.

For example:
```fortran
program array_bounds_check_01
    integer, allocatable :: x(:)
    allocate(x(10))
    ! realloc(x(3)) will be inserted here
    x = [1, 2, 3]
end program
```
If we include the reallocate statement then everything is in bounds, but if we don't have it then we throw an `ArrayItem` bounds check error.

certik: I would say to split this logic into a separate PR. We will have to add a lot more tests to exactly guarantee to either work or give a runtime error, and for both with and without realloc LHS.

In this PR, just implement what works, and skip what doesn't work. And remove the above change. And hopefully no changes are needed to the CI.

Then in subsequent PRs we can iterate on this and make it better.

ci/test_third_party_codes.sh
```diff
@@ -837,15 +837,15 @@ time_section "🧪 Testing SNAP" '
...
     ./gsnap ../qasnap/sample/inp out
 
     make clean
-    make -j8 FORTRAN=$FC FFLAGS="--separate-compilation" MPI=no OPENMP=no
+    make -j8 FORTRAN=$FC FFLAGS="--separate-compilation --no-array-bounds-checking" MPI=no OPENMP=no
     ./gsnap ../qasnap/sample/inp out
 
     make clean
-    make -j8 FORTRAN=$FC FFLAGS="--fast" MPI=no OPENMP=no
+    make -j8 FORTRAN=$FC FFLAGS="--fast --no-array-bounds-checking" MPI=no OPENMP=no
```
certik: Doesn't `--fast` imply `--no-array-bounds-checking`? If it doesn't, then it should. Then this is not needed.

certik:
Perfect. Now my worry is this one:

* Why do so many 3rd party codes require `--no-array-bounds-checking`? Why does old F77 style code like SciPy fail? I am less worried there. But even modern Fortran code like fastGPT fail with the bounds checking. That's worrying, because there should be no bugs.

Let me understand the issue first. But I think we need to implement this bounds checking in such a way that it can always be on, unless the Fortran code has bugs, in which case we just have to fix it (but I don't think that's the case here).

You are probably still doing too much in this PR and either there are bugs or it is too aggressive that even valid code doesn't pass, or something like that. Do you understand what might be going on?

Let's figure this out.

certik:
I marked it as draft, go ahead and click on "Ready for review" when it's ready for another review.

integration_tests/arrays_08_size.f90
xaerru: Reverted my change and added `--no-array-bounds-checking` to this test. The assignment from the `FunctionCall` to result had different sizes.

integration_tests/array_indices_array_item.f90
xaerru: The earlier test was incorrect and throws index out of bounds. The `arr_idx`  array is `[2, 1]`. So we need to allocate the first dimension of `arr` to atleast 2.

ci/test_third_party_codes.sh
```diff
@@ -236,7 +236,7 @@ time_section "🧪 Testing stdlib (Less Workarounds)" '
   FC=$FC cmake . \
       -DTEST_DRIVE_BUILD_TESTING=OFF \
       -DBUILD_EXAMPLE=ON -DCMAKE_Fortran_COMPILER_WORKS=TRUE \
-      -DCMAKE_Fortran_FLAGS="--cpp --separate-compilation --realloc-lhs --no-warnings --use-loop-variable-after-loop -I$(pwd)/src -I$(pwd)/subprojects/test-drive/"
+      -DCMAKE_Fortran_FLAGS="--cpp --separate-compilation --realloc-lhs --no-warnings --use-loop-variable-after-loop --no-array-bounds-checking -I$(pwd)/src -I$(pwd)/subprojects/test-drive/"
```
certik: Why doesn't stdlib run with bounds checking?

ci/test_third_party_codes.sh
```diff
@@ -195,13 +195,13 @@ time_section "🧪 Testing POT3D with fortran_mpi" '
...
   print_subsection "Building with default flags"
-  FC="$FC --cpp -DOPEN_MPI=yes" ./build_and_run_lfortran.sh
+  FC="$FC --cpp -DOPEN_MPI=yes --no-array-bounds-checking" ./build_and_run_lfortran.sh
 
   print_subsection "Building with optimization flags"
   # FC="$FC --cpp --fast --skip-pass=dead_code_removal -DOPEN_MPI=yes" ./build_and_run_lfortran.sh
 
   print_subsection "Building POT3D in separate compilation mode"
-  FC="$FC --cpp --separate-compilation -DOPEN_MPI=yes" ./build_and_run_lfortran.sh
+  FC="$FC --cpp --separate-compilation --no-array-bounds-checking -DOPEN_MPI=yes" ./build_and_run_lfortran.sh
```
certik: Why doesn't POT3D run with bounds checking?

ci/test_third_party_codes.sh
```diff
@@ -837,11 +837,11 @@ time_section "🧪 Testing SNAP" '
     git checkout lf11
     git checkout 169a9216f2c922e94065a519efbb0a6c8b55149e
     cd ./src
-    make -j8 FORTRAN=$FC FFLAGS= MPI=no OPENMP=no
+    make -j8 FORTRAN=$FC FFLAGS="--no-array-bounds-checking" MPI=no OPENMP=no
     ./gsnap ../qasnap/sample/inp out
 
     make clean
-    make -j8 FORTRAN=$FC FFLAGS="--separate-compilation" MPI=no OPENMP=no
+    make -j8 FORTRAN=$FC FFLAGS="--separate-compilation --no-array-bounds-checking" MPI=no OPENMP=no
```
certik: Why doesn't SNAP work with bounds checking?

certik:
I noticed SNAP, POT3D and stdlib still don't work with bounds checking. Do you understand why? I think it's a bug in the bounds checker, as these codes are valid, modern Fortran codes, so should pass.

Then old F77 minpack doesn't pass, and neither does SciPy. I am less worried there, however I would still like to understand the reason --- do you know a simple MRE for them, that doesn't pass our bounds checker? If it is something with "legacy array sections", we should just turn off the bounds checker for that, as that usage (I think) cannot easily be checked.

I would start with a subset and ensure all codes run with bounds checker on. Then slowly start expanding it. The goal should be that LFortran with no options (all defaults) should always run bounds checker and everything should pass for codes that it can compile. Then if you start adding options, such as "--legacy-array-sections" then it will turn off the bounds checker for that feature (if it makes sense).

certik:
I also noticed some tests were modified --- if they were invalid code, then they should be modified. Otherwise let's add new tests, don't change existing tests.

src/libasr/pass/array_op.cpp
```diff
@@ -851,7 +856,7 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
     void insert_realloc_for_target(ASR::expr_t* target, ASR::expr_t* value, Vec<ASR::expr_t**>& vars) {
         ASR::ttype_t* target_type = ASRUtils::expr_type(target);
         bool array_copy = ASR::is_a<ASR::Var_t>(*value) && ASR::is_a<ASR::Var_t>(*target);
-        if( (realloc_lhs == false || !ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
+        if( (realloc_lhs = false || !ASRUtils::is_allocatable(target_type) || vars.size() == 1) &&
```
certik: I think this is a bug. If you meant it this way, then assign `realloc_lhs = false` above this on a standalone line.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -2376,7 +2376,7 @@ namespace Eoshift {
             b.Assignment(b.ArrayItem_01(result, {i}), b.ArrayItem_01(args[0], {j})),
             b.Assignment(i, b.Add(i, b.i32(1))),
         }, nullptr));
-        body.push_back(al, b.DoLoop(j, LBound(args[0], 1), b.Add(shift_val, b.i32(1)), {
+        body.push_back(al, b.DoLoop(j, LBound(args[0], 1), shift_val, {
```
certik, line 2382: Is this a bug in the shift intrinsic?

ci/test_third_party_codes.sh
xaerru: It's a bug in POT3D.

The error is:
```shell
Runtime error: Array shape mismatch in subroutine 'genmesh'

Tried to match size 151 of dimension 1 of argument number 12, but expected size is 150
```

This call is responsible: https://github.com/gxyd/POT3D/blob/lf_hdf5_fortranMPI_namelist_global_workarounds/src/pot3d.F90#L2671

`npm1_g` is 150, `p_g` is supposed to have dimension `npm1_g`. But `p_g` has size 151.

ci/test_third_party_codes.sh
xaerru: It's a bug in SNAP.

The error is:
```shell
$ ./gsnap ../qasnap/sample/inp out                                                                                                                                    (lf)

   *WARNING: INPUT_CHECK: MULTISWP reset to 0 for serial

          keyword Iteration Monitor
********************************************************************************
  Outer
Runtime error: Array shape mismatch in subroutine 'outer_src_calc'

Tried to match size 1 of dimension 1 of argument number 8, but expected size is 0
```

In this call: https://github.com/certik/SNAP/blob/lf11/src/outer.f90#L223.
The last argument's expected size in dimension 1 is `cmom-1` from https://github.com/certik/SNAP/blob/lf11/src/outer.f90#L256.

At the point of the call `cmom` is 1, so expected size is 0.

ci/test_third_party_codes.sh
xaerru: It doesn't work with separate compilation (`sc-lf-9` branch). stdlib has a bug.

```shell
$ ./example/array/example_falseloc                                                                                                                      (lf)
Runtime error: Array 'loc' index out of bounds.

Tried to access index 2 of dimension 1, but valid range is 1 to 1.
```

This is the test:
```fortran
program example_falseloc
  use stdlib_array, only: falseloc
  implicit none
  real, allocatable :: array(:)
  allocate (array(200))
  call random_number(array)
  array(falseloc(array < 0.5, lbound(array, 1))) = 0.0
end program example_falseloc
```
And definition of `falseloc` is:
```fortran
  pure function falseloc(array, result_size, lbound) result(loc)
    !> Mask of logicals
    logical, intent(in) :: array(:)
    integer, intent(in) :: result_size
    !> Lower bound of array to index
    integer, intent(in), optional :: lbound
    !> Locations of false elements
    integer :: loc(result_size)

    call logicalloc(loc, array, .false., lbound)
  end function falseloc
```

We are not passing `result_size` argument in `example_falseloc`.

This diff fixes it. But it also outputs garbage, we can't know the `result_size` because we are randomly generating the integers.
```diff
diff --git a/example/array/example_falseloc.f90 b/example/array/example_falseloc.f90
index a02b165a..39af4c09 100644
--- a/example/array/example_falseloc.f90
+++ b/example/array/example_falseloc.f90
@@ -4,5 +4,5 @@ program example_falseloc
   real, allocatable :: array(:)
   allocate (array(200))
   call random_number(array)
-  array(falseloc(array < 0.5, lbound(array, 1))) = 0.0
+  print *, falseloc(array < 0.5, 200, lbound(array, 1))
 end program example_falseloc
```

src/libasr/pass/intrinsic_array_function_registry.h
xaerru: Yes, we get bounds check error without the fix on 2 integrations tests.
```shell
$ ./src/bin/lfortran integration_tests/intrinsics_248.f90                                                                                                (lf)
3    4    10    10
4.00000000e+00    1.00000000e+01    1.00000000e+01    1.00000000e+01
ab    cd    de    gt    tg
3    4    0    0
4.00000000e+00    0.00000000e+00    0.00000000e+00    0.00000000e+00
ab    cd
10    10    57    62
1.00000000e+01    1.00000000e+01    1.00000000e+01    3.10200005e+01
ab    cd    de    gt    tg
0    0    1    2
0.00000000e+00    0.00000000e+00    0.00000000e+00    1.00000000e+00
ab    cd
Runtime error: Array 'result' index out of bounds.

Tried to access index 5 of dimension 1, but valid range is 1 to 4.
$ ./src/bin/lfortran integration_tests/intrinsics_384.f90                                                                                                (lf)
Runtime error: Array 'result' index out of bounds.

Tried to access index 4 of dimension 1, but valid range is 1 to 3.
```

certik: Great job!

ci/test_third_party_codes.sh
certik: Consider the following program:
```fortran
program sizes
implicit none

integer :: x(6)

x = 6
call f(x)

contains

    subroutine f(A)
    integer, intent(in) :: A(3)
    print *, A
    end subroutine

end
```

This compiles and runs with:
```console
$ gfortran -W -Wall  -fcheck=all a.f90 && ./a.out
           6           6           6
```
But if you change `x(6)` to `x(2)` you get:
```console
$ gfortran -W -Wall  -fcheck=all a.f90 && ./a.out
a.f90:7:7:

    7 | call f(x)
      |       1
Error: Actual argument contains too few elements for dummy argument ‘a’ (2/3) at (1)
```
In this case a compile-time error, but with runtime dimensions we would get a runtime bounds check error.


I think we already discussed this case. The rule seems to be that the passed in array can be longer than what the parameter expects, but not shorter. If the input argument length is `N` and the parameter expects `M`, then the relationship must be `N >= M`. I think currently you are checking `N == M`.

I actually like to have an option to enforce `N == M`, in fact I think it probably even should be the default, since `N > M` is almost always a bug, and the code can be changed to ensure `N == M` always holds.

However, right now we are just starting, and we should enforce what other compilers do and the standard allows, which is `N >= M`. That way all the codes should pass, and we can merge the PR.

Later let's figure out how to enforce `N == M`, we'll have to make it configurable. I think later we can make `N == M` the default, and you will be able to relax it to `N >= M` with a compiler option for codes like SNAP (the default error message will say if your code depends on `N > M`, you can relax this check by using `--allow-larger-array-arguments`). But for now let's always do `N >= M` to get us unblocked and so that we can move forward, in the spirit of first getting some subset working, then iterating.

ci/test_third_party_codes.sh
certik: This is another case of `N > M` see below: https://github.com/lfortran/lfortran/pull/7955#discussion_r2248351125.

I agree that likely this is a bug. But since this is allowed by the standard and other compilers, let's for now only enforce `N >= M`, let's ensure POT3D fully passes and let's get this PR merged.

See the other comment how we can later make this check stricter.

ci/test_third_party_codes.sh
certik: Can you report this analysis into an stdlib issue and link it here? It looks like they are passing "lbound" as the second argument by a mistake.

Until they fix it upstream, can you make this workaround in the stdlib branch and change it in this PR to ensure otherwise stdlib works? This is an important test for this PR to ensure that all the checks that we implemented here actually work in practice on real code.

ci/test_third_party_codes.sh
xaerru: Changed the condition to `N >= M`. 

Also in the latest commit I added checking for rank at runtime. If the ranks are not equal then the sizes should be equal else we throw a runtime error. Let's see if the CI passes.

Pranavchiku:
Is it ready to be reviewed?

ci/test_third_party_codes.sh
xaerru: The size argument is a workaround done in LFortran's stdlib. stdlib upstream doesn't have the size argument.

certik:
It starts to look pretty good! The CI fails.

src/libasr/asr_utils.h
```diff
@@ -5024,6 +5067,66 @@ class SymbolDuplicator {
 
 };
 
+// This replacer is for checking if symbols in the input expression are accessible from current_scope.
+// If they are accessible as external symbols then replace them by the corresponding ExternalSymbol.
+// If atleast one symbol is not accessible then result is false.
```
certik, line 5072:
```suggestion
// If at least one symbol is not accessible then result is false.
```

Issue #7924, opened 2025-06-29
Implement function loc(x)


certik:
https://gcc.gnu.org/onlinedocs/gfortran/LOC.html

An example from https://fortran-lang.discourse.group/t/bug-on-gfortran-15-1-on-mac/9876/11:

```fortran
program main
    implicit none
    integer :: i
    integer, save :: j

    print*, "main not saved", loc(i)
    print*, "main     saved", loc(j)
    call sub()
contains
    subroutine sub()
        integer :: i
        integer, save :: j

        print*, "sub  not saved", loc(i)
        print*, "sub      saved", loc(j)
    end
end program
```

harperjf:
Fortran 2023 seems to have no intrinsic called loc. Where is it defined?

certik:
I think it's not in the standard, but apparently supported by GFortran, ifx and Flang, so we should also support it. People use it, as evidenced by the link to Fortran Discourse.

harperjf:
Gfortran and ifx do implement ```loc()``` but in different ways. Which do you want lfortran to do? (My own choice would be to implement all of Fortran 95 before adding extensions that are different with different compilers.) My test program and what
three compilers did with it:
```
  print *, loc(Z'1111'), loc(42)
end program
```
Ifx output:
```
               4628492               4628496
```
gfortran output:
```
loctest.f90:1:15:

    1 |   print *, loc(Z'1111'),loc(42)
      |               1
Error: ‘x’ argument of ‘loc’ intrinsic at (1) must be a variable
loctest.f90:1:15:

    1 |   print *, loc(Z'1111'),loc(42)
      |               1
Error: A BOZ literal constant at (1) cannot appear as an actual argument in a function reference
loctest.f90:1:28:

    1 |   print *, loc(Z'1111'),loc(42)
      |                            1
Error: ‘x’ argument of ‘loc’ intrinsic at (1) must be a variable
```
AMD flang output:
```
F90-S-0033-Illegal use of constant 4369 (loctest.f90: 1)
  0 inform,   0 warnings,   1 severes, 0 fatal for MAIN
```
Note that int(z'1111') is 4369.

certik:
I think `loc` is meant to be used with variables. Then all compilers that support it behave the same.

Of course the question then becomes what happens if the variable is optimized out, but maybe calling `loc` prevents that.

I agree this is a low priority for us.

certik:
After #7956, the code above now compiles and prints:
```
main not saved    1841771004
[...]

PR #7956, opened 2025-07-03
Implement `loc` intrinsic

HarshitaKalani:
Towards: https://github.com/lfortran/lfortran/issues/7924


HarshitaKalani:
This implementation works but still needs improvement to correctly return distinct loc() values for SAVE variables. Also, can we get negative values for loc?

PR #7957, opened 2025-07-03
ref: separate compilation flags

Pranavchiku:
Addressing https://github.com/lfortran/lfortran/pull/7939#issuecomment-3020198327

### DIFF
--- a/examples/project2/CMakeLists.txt
+++ b/examples/project2/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 project(project2 C Fortran)
 
+if(CMAKE_VERSION VERSION_LESS 3.31.4)
+  # CMake 3.31.0 through 3.31.3 hard-coded this flag.
+  # It does not work with modules, so remove it.
+  string(REPLACE "--generate-object-code" "" CMAKE_Fortran_COMPILE_OBJECT "${CMAKE_Fortran_COMPILE_OBJECT}")
+endif()
+
 add_executable(doconcurrent
     doconcurrent.f90
 )


src/libasr/utils.h
```diff
@@ -109,7 +109,7 @@ struct CompilerOptions {
 
         It is overridden by `generate_object_code` option.
```
certik, line 110: fix this too.

Pranavchiku:
The failure was due to https://github.com/Pranavchiku/minpack/commit/45801cf882871ea8a668213e8cf90b5817877484

certik:
In our next release we should document in the release notes this breaking change. But I think we should do it.

Also with a bit of luck, the direct compilation should now work, including in the cmake tests with global subroutines, automatically. So the `--separate-compilation` will not be as needed, only when users really want that.

examples/project1/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 project(project1 C Fortran)
 
+set(CMAKE_Fortran_COMPILER_WORKS TRUE CACHE INTERNAL "")
+
 if(CMAKE_VERSION VERSION_LESS 3.31.4)
   # CMake 3.31.0 through 3.31.3 hard-coded this flag.
```
Pranavchiku: We need to send a patch to CMake to remove `--generate-object-code` and use `--separate-compilation`

assem2002, line 8: CI is operating on 3.31.2 at that moment of testing. We end up using old flag `generate-object-code`

Pranavchiku:
@certik are we good to merge this?

examples/project2/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 project(project2 C Fortran)
 
+if(CMAKE_VERSION VERSION_LESS 3.31.4)
+  # CMake 3.31.0 through 3.31.3 hard-coded this flag.
+  # It does not work with modules, so remove it.
```
certik, line 7: In the subsequent PR, let's fix this documentation. The old flag `--generate-object-code` is the same as the new flag `--separate-compilation` and of course it works with modules. So the documentation needs to be changed.

Pranavchiku: Yes, I'll complete a refactoring allocate and then open PR for it.

Issue #7871, opened 2025-06-23
Submodules : Handle Function declaration in Module interface having implementation in Submodule


AnshMehta1:
Towards #536

MRE :- 
```.f90
module stdlib_hash_32bit

    implicit none

    type :: key_type
        integer(1), allocatable :: value(:)
    end type key_type

    interface fnv_1_hash
        pure module function int8_fnv_1( key ) result(hash_code)
            integer(1), intent(in) :: key(:)
            integer(4)             :: hash_code
        end function int8_fnv_1
    end interface fnv_1_hash

end module stdlib_hash_32bit


submodule(stdlib_hash_32bit) stdlib_hash_32bit_fnv

    implicit none

contains

    pure module function int8_fnv_1( key ) result(hash_code)
        integer(1), intent(in) :: key(:)
        integer(4)             :: hash_code
    end function int8_fnv_1

end submodule stdlib_hash_32bit_fnv

program main

    use stdlib_hash_32bit

    implicit none

    integer(4) :: fnv_1_hasher
    type(key_type) :: key

    fnv_1_hasher = fnv_1_hash( key % value )
end program
```

LFortran Error :- 
```console
(lf) lordansh@LordAnsh:~/dev/lfort1/lfortran$ lfortran try.f90
/usr/bin/ld: /tmp/try.tmp.o: in function `main':
LFortran:(.text+0x58): undefined reference to `int8_fnv_1_integer____0'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
The command 'clang -o try.out /tmp/try.tmp.o  -L"/home/lordansh/dev/lfort1/lfortran/src/bin/../runtime" -Wl,-rpath,"/home/lordansh/dev/lfort1/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
```

MRE works well with GFortran .

Issue #536, opened 2022-08-15
Implement submodules


certik:
*Original issue*: https://gitlab.com/lfortran/lfortran/-/issues/694

Current status: they parse to ASR, but as separate modules, and then they compile to LLVM as unrelated functionality, so the interface in the parent module does not have an implementation and fails at linking. A simple test was added in https://gitlab.com/lfortran/lfortran/-/merge_requests/1736:

```
$ lfortran submodule_03.f90
Undefined symbols for architecture arm64:
  "___module_submodule_03_mod_f", referenced from:
      _main in a.out.tmp.o
ld: symbol(s) not found for architecture arm64
```

AnshMehta1:
Features of Submodule :- 
  1. Provides implementation to interface subroutines/functions of parent module. (Should  use same program_unit_t nodes)

  2. Interfaces are not needed in submodules — only implementations are written there

  3. Can access all the entities (public/private) of Parent Module.

  4. Functions/subroutines present in Submodule can only be used if its present in Module interface. (We won't get compilation error if module interface doesn't have function/subroutine present in Submodule).

  5. Parent module can compile without submodule but using its subroutine or function in interface will cause the error.

  6. Submodule should always compile after the compilation of parent module as submodule uses modfile (.mod) of parent module for compilation.

certik:
For monolithic compilation when we load a module at a `use` statement in the main program (and recursively) we need to also load any submodules. Some paths forward:

* For each module load we scan the include paths for submodule `.mod` files (we can even use `.smod` extension to make it easier) and use the parent module name to find those and load those
* We store the information for each parent module which submodules it has, either by modifying the parent module `.mod` file or storing this separately in a file somehow.
* The user can tell us explicitly when compiling the main program that there is a submodule, for example by adding the `submodule.mod` file on the command line; then we simply load it. In cmake build system we'll have to add these extra options when compiling the main program, whether listing the `.mod` file or by dedicated option like `--add-submodule my_submodule.mod`. 

I like the last option the most. We can start with that, and later on we can investigate if some combination of the previous two options could be better.

AnshMehta1:
Submodules in monolithic compilation aren't working. 
For eg. 

a.f90
```.f90
module mod_separate_compilation_16
    implicit none

    interface
        module subroutine map_open_entry(key)
            integer, intent(inout) :: key
        end subroutine map_open_entry
    end interface
    
end module mod_separate_compilation_16
```

b.f90
```.f90
[...]

PR #7961, opened 2025-07-03
feat : Implement Submodule in ASR by linking Parent Module Functions with corresponding Submodule Functions

### LINKED ISSUE #7270: [stdlib:hashmaps] Calling type-bound procedure declared via module and implemented in submodule is causing `code generation error`
MRE :- 
```.f90
module stdlib_hashmaps
    implicit none
    private

    public ::  open_hashmap_type

    type :: open_hashmap_type
    contains
        procedure :: map_entry => map_open_entry
    end type open_hashmap_type

    interface
        module subroutine map_open_entry(map, key)
            integer, intent(in) :: key
        end subroutine map_open_entry
    end interface
    
end module stdlib_hashmaps

submodule(stdlib_hashmaps) stdlib_hashmap_open
    implicit none
contains

    module subroutine map_open_entry(map, key)
        integer, intent(in) :: key
    end subroutine map_open_entry

end submodule stdlib_hashmap_open

program main
    use stdlib_hashmaps, only : open_hashmap_type

    implicit none

    type(open_hashmap_type)   :: map
    integer :: key
    call map % map_entry( key )

end program
```

LFortran CodeGen Error :- 
```console
(lf) lordansh@LordAnsh:~/dev/lfort4/lfortran$ lfortran test.f90
code generation error: Subroutine code not generated for 'map_open_entry'


```

AnshMehta1:
Toward #536 
Fixes #7270 #7871 

Features of PR :- 
1. Once the ASR (body, args) of Submodule Functions/Subrountines is generated, we populate their corresponding Functions/Subroutines (based on name) in Parent Module.
2. In Parent Module Functions/Subroutines, we convert deftype to implementation from interface.
3. Then, we add variables and ExternalSymbols, made in Submodule Function node to symbol table of parent module Function node.
4. We also add variables and ExternalSymbols, of Submodule to Symbol Table of Parent Module
5. We then populate body of Parent Module function by duplicating statements of Submodule function to update the scopes using statement duplicator based on , first,  module scope and then function scope.
6. Some Utility Functions are also created to support this duplication process to update SymbolTable.
7. Finally, m_loaded_from_mod flag of parent module is turned to false to re-save updated Parent Module ASR.

### DIFF
--- a/tests/reference/asr-derived_types_05-35150cd.json
+++ b/tests/reference/asr-derived_types_05-35150cd.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-derived_types_05-35150cd.stdout",
-    "stdout_hash": "b763f78cdada445f0bb0300fef642262dde56f98cc31d88be338702e",
+    "stdout_hash": "5780fa6e39458db0cdfe8cf67b7bf8fd51928fce2c6e1d65157ec5c4",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-derived_types_05-35150cd.stdout
+++ b/tests/reference/asr-derived_types_05-35150cd.stdout
@@ -81,7 +81,7 @@
                                             2 string_type
                                         )
                                         Source
-                                        Interface
+                                        Implementation
                                         ()
                                         .true.
                                         .true.
@@ -163,7 +163,7 @@
                                             2 string_type
                                         )
                                         Source
-                                        Interface
+                                        Implementation
                                         ()
                                         .true.
                                         .true.
@@ -245,7 +245,7 @@
                                             2 string_type
                                         )
                                         Source
-                                        Interface
+                                        Implementation
                                         ()
                                         .true.
                                         .true.
@@ -327,7 +327,7 @@
                                             2 string_type
                                         )
                                         Source
-                                        Interface
+                                        Implementation
                                         ()
                                         .true.
                                         .true.
--- a/tests/reference/asr-submodule_02-9152b7b.json
+++ b/tests/reference/asr-submodule_02-9152b7b.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-submodule_02-9152b7b.stdout",
-    "stdout_hash": "05779c58b6dfd1dbc35ed694e8c0361083aa458e02ca67179b288073",
+    "stdout_hash": "7db832925719b7acb68c35533ff8b75303166181702628176a432192",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-submodule_02-9152b7b.stdout
+++ b/tests/reference/asr-submodule_02-9152b7b.stdout
@@ -75,6 +75,26 @@
                                     (SymbolTable
                                         8
                                         {
+                                            1_point_x:
+                                                (ExternalSymbol
+                                                    8
+                                                    1_point_x
+                                                    7 x
+                                                    point
+                                                    []
+                                                    x
+                                                    Public
+                                                ),
+                                            1_point_y:
+                                                (ExternalSymbol
+                                                    8
+                                                    1_point_y
+                                                    7 y
+                                                    point
+                                                    []
+                                                    y
+                                                    Public
+                                                ),
                                             a:
                                                 (Variable
                                                     8
@@ -165,7 +185,7 @@
                                         )]
                                         (Real 4)
                                         Source
-                                        Interface
+                                        Implementation
                                         ()
                                         .false.
                                         .false.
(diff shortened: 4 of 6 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -2250,6 +2297,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         }
 
         create_and_replace_structType();
+        populate_parent_module(v);
```
certik, line 2331: Question: The parent module has to be compiled first, right? So the `.mod` file is already created.

Then we handle the submodule, we save it to a `.mod` file. That means these modifications to the parent module in ASR will get lost, correct? Is that a problem? Why do we need modifications to the parent in the first place?


In other words, I am worried that the above change only works if the parent module and the submodule are both in the same file, but will fail if they are compiled in separate files. If that is the case, we need to find another solution.

AnshMehta1: Firstly, we always import Parent Module in `use` statement. So, Whenever,  we make Subroutine call or Function Call, it will be a call to `Function_t*` node of parent and thus, we need to update the parent function. Otherwise, we will need to update multiple things that is connected to parent function in `ASR_to_LLVM` phase.

AnshMehta1: For Separate files, we can make `m_loaded_from_mod` flag false. I think , by doing this ,  we can again save updated mod file.

certik: I don't understand why the parent Function_t has to be modified. Normally you call it via ExternalSymbol.

But I would say, get it working with both parent/sub module in a single file as well as in separate files, and get it working with both direct and separate compilation. As long as that works, then I think the design might work.

AnshMehta1: Ohhh... Ya, I forgot about ExternalSymbol. Yes, It would make the work a lot easier. Will make changes accordingly.

AnshMehta1: We can convert Parent Function to ExternalSymbol pointing to Submodule Function only when, we have submodule compiled. This means that, anyways, we need to again save the mod file for parent module.

Pranavchiku:
Take a look at the case below where Function signature is present as an interface in parent module and body is linked via submodule. This are cases similar to `procedure` , here termed as `module procedure`

> a.f90

```fortran

module stdlib_error
implicit none

interface ! f{08,18}estop.f90
    module subroutine error_stop(msg, code)
        !! version: experimental
        !!
        !! Provides a call to `error stop` and allows the user to specify a code and message
        !! ([Specification](..//page/specs/stdlib_error.html#description_1))
        character(*), intent(in) :: msg
        integer, intent(in), optional :: code
    end subroutine error_stop
end interface

end module


program main
    use stdlib_error
    implicit none

    call error_stop("This is an error message", 2)

end program main

```

> b.f90

```fortran

implicit none

contains

module procedure error_stop
! Aborts the program with nonzero exit code
!
! The "stop <character>" statement generally has return code 0.
! To allow non-zero return code termination with character message,
! error_stop() uses the statement "error stop", which by default
! has exit code 1 and prints the message to stderr.
! An optional integer return code "code" may be specified.
!
! Example
! -------
!
! call error_stop("Invalid argument")

if(present(code)) then
  write(*,*) msg
  error stop code
else
  error stop msg
endif
end procedure

end submodule f18estop
```


```console
% gfortran a.f90 b.f90 && ./a.out
 This is an error message
ERROR STOP 2

Error termination. Backtrace:
#0  0x104f32d2f
#1  0x104f338d7
#2  0x104f34b1f
#3  0x104903e77
#4  0x104903d9b
#5  0x104903dd7
```

src/lfortran/semantics/ast_body_visitor.cpp
Pranavchiku: If we use `ExternalSymbol` in parent module then we will lose function signature.  Example at https://github.com/lfortran/lfortran/pull/7961#issuecomment-3036569212

AnshMehta1: So, the best way was our earlier implementation where we populate m_body of Parent Function.

AnshMehta1:
Currently, Submodules are implemented for a Single file. For Separate files, changes made is breaking stdlib. Will handle Separate files with Submodules in a separate PR, or else this PR will get extended unnecessarily.

certik:
Ok. Get the PR to pass the CI, we can merge it in pieces.

src/lfortran/semantics/ast_body_visitor.cpp
certik: The parent module has interfaces, right? Why cannot we simply use interfaces like other interfaces that already work? Why do we need to modify it?

AnshMehta1: Our Current Interfaces have corresponding Functions in same Module. That Functions are always complete, has function_signature (args, type) etc. But for Submodules Function, It is not necessary to have signature. Just Body can also be present in Submodule as seen in https://github.com/lfortran/lfortran/pull/7961#issuecomment-3036569212

AnshMehta1: Either we, populate submodule functions with signature when required and point them using external signature or transfer body to parent. 2nd Approach is generalized and so, I decided to go on with it.

certik: I am worried that modifying the parent module might not work super well, especially if there are multiple nested modules, etc.

Instead, why cannot we treat the parent module's function as interfaces?

certik: It's more important to get it working, and this approach does seem quite simple --- you just fill it in. But it might lead to a dead-end, where with multiple nesting and separate compilation we might fail. For example, parent module A, doubly nested modules B and C. Separate compilation. If both B and C fill A, how exactly will the separate compilation work? Unless you have a clear plan how this will work, let's go with the other approach:

With my interface approach above, it will work, B, C will generate object code, then everything gets linked.

AnshMehta1: For Interface approach, what parent module Function should Point? Like I didn't get it. Also, I am not much clear about Functioning of Separate Compilation, I didn't got the Actual Problem in nesting modules.

certik: I think even with parent A and submodule B and separate compilation: we only compile B, since A is already compiled to object code. And when we link everything together, I think it won't work well.

AnshMehta1:
PR with updated approach is #8010

PR #7964, opened 2025-07-04
Replace create_gep with create_gep2

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/7836


HarshitaKalani:
ALthough I tried this with llvm version 19.
```
(lf19) harshita@harshita-linux:~/Desktop/lfortran$ ./integration_tests/run_tests.py
.
.
.
100% tests passed, 0 tests failed out of 1726

Label Time Summary:
c                 =   0.47 sec*proc (80 tests)
c_nopragma        =   0.02 sec*proc (2 tests)
cpp               =   0.27 sec*proc (44 tests)
fortran           =   4.51 sec*proc (723 tests)
gfortran          =  10.11 sec*proc (1654 tests)
llvm              =  10.54 sec*proc (1726 tests)
llvm2             =   0.06 sec*proc (11 tests)
llvmImplicit      =   0.01 sec*proc (2 tests)
llvmStackArray    =   0.72 sec*proc (123 tests)
llvm_nopragma     =   0.04 sec*proc (3 tests)
llvm_omp          =   0.05 sec*proc (8 tests)
llvm_wasm         =   9.58 sec*proc (1564 tests)
llvm_wasm_emcc    =   9.57 sec*proc (1563 tests)
mlir              =   0.07 sec*proc (12 tests)
mlir_llvm_omp     =   0.00 sec*proc (1 test)
mlir_omp          =   0.00 sec*proc (1 test)
wasm              =   1.28 sec*proc (227 tests)
x86               =   0.12 sec*proc (19 tests)

Total Test time (real) =   2.15 sec
(lf19) harshita@harshita-linux:~/Desktop/lfortran$ llvm-config --version
19.1.7
```

EDIT: I see, I need to build LFortran in both the envs.

HarshitaKalani:
I know we don't want to use ifdef at multiple places. Can we use it in one function in llvm_utils and use that everywhere? Although, I'm myself not convinced.
```
llvm::Type* LLVMUtils::get_pointee_type(llvm::Value* val) {
    llvm::Type* ty = val->getType();
    if (!ty->isPointerTy()) {
        throw std::runtime_error("Expected a pointer type");
    }
#if LLVM_VERSION_MAJOR >= 17
    return llvm::cast<llvm::PointerType>(ty)->getNonOpaquePointerElementType();
#else
    return llvm::cast<llvm::PointerType>(ty)->getPointerElementType();
#endif
}
```

certik:
We can use ifdefs inside a function, and we should collect all such functions into one place.

However, we should not be using code like `llvm::cast<llvm::PointerType>(ty)->getPointerElementType();`, because there is no equivalent in LLVM >= 17. Pointers do not have type there. Consequently, we must pass the type down the call chain.

HarshitaKalani:
Closing as not needed.

PR #7966, opened 2025-07-04
Union ASR refactor

swamishiju:


swamishiju:
I've made some changes to Union ASR im still working on integrating like List types into union, I also had some thoughts about security concerns in Union about `ptr` types in general,  we should probably make it "tagged" union and a corresponding `is_instance` type intrinsic. 

There could be maybe an error/warning when you trying accessing the wrong member and you can turn it off in the syntax maybe like
```f90
type(test_type) :: tagged_union
type(test_type), untagged :: untagged_union

tagged_union%int32 = 10
! Can give arbitrary memory write
_lfortran_list_append(tagged_union%list_i32, 10)
! Warning tagged_union is currently tagged as integer, not list
```

src/libasr/ASR.asdl
```diff
@@ -163,7 +163,7 @@ expr
     | StructInstanceMember(expr v, symbol m, ttype type, expr? value)
     | StructStaticMember(expr v, symbol m, ttype type, expr? value)
     | EnumStaticMember(expr v, symbol m, ttype type, expr? value)
-    | UnionInstanceMember(expr v, symbol m, ttype type, expr? value)
+    | UnionInstanceMember(expr v, identifier member_name, ttype type, expr? value)
```
certik: This should be a symbol, like StructInstanceMember. We should not use strings in ASR, but direct pointers to things. The job of AST->ASR is to resolve all these strings and turn them into pointers and integers (enums).

src/libasr/ASR.asdl
```diff
@@ -208,7 +208,7 @@ ttype
     | Tuple(ttype* type)
     | StructType(ttype* data_member_types, ttype* member_function_types, bool is_cstruct, symbol derived_type)
     | EnumType(symbol enum_type)
-    | UnionType(symbol union_type)
+    | UnionType(identifier name, ttype* members, identifier* member_names, abi abi, access access, symbol? parent)
```
certik: Names are not part of a ttype and parent I think also just has to be at ttype.
```suggestion
    | UnionType(ttype* member_types, ttype? parent)
```
Everything else is stored in the symbol table.

certik: I think the parent should also not be there:https://github.com/lfortran/lfortran/pull/5791/files#r2188038351, so I would start with just:

```suggestion
    | UnionType(ttype* member_types)
```

swamishiju: For a given UnionType how would we get corresponding Union from symtab?

certik:
> For a given UnionType how would we get corresponding Union from symtab?

You have to know the variable. You follow the variable and Var points to a symbol table. The Union symbol has all the information like names, etc.

swamishiju:
There is going to be some major refactoring need for that which is in the #5791 PR, should I put this one on hold until that is merged?

Mainly its just one function `getUnion` which is in some heavily used functions like `get_type_from_ttype`. The fix for the analogous `getStructType` in #5791 was to refactor `get_type_from_ttype` to include a `expr_t` argument and a new function called `get_struct_sym_from_struct_expr`. The refactoring needed for that is a lot so imo I should wait until that is merged then it should be easy to get this working.

certik:
Yes, I would wait until the struct refactor it done, then you can do it.

When you are implementing the union features, just keep this in mind. Also coordinate with them, so that we can finish the struct refactor soon.

swamishiju:
The commit history is a bit messy - i somehow got a commit by jinang while merging with main

integration_tests/lp_union_test_03.f90
```diff
@@ -0,0 +1,26 @@
+module lp_union_test_03_mod
```
certik: Let's not use lp, just remove this prefix from everywhere.

integration_tests/lp_union_test_03.f90
```diff
@@ -0,0 +1,26 @@
...
+    implicit none
+    _lfortran_union_type :: test_type
+        integer                              :: int_
+        real                                 :: float_
+        _lfortran_list(integer)              :: list
```
certik: How does deallocation work for union with lists in it?

I thought unions in C++ can only be for types that do not require deallocation. Because we do not know if the "int" or "list" is used, we can't deallocate "list", since it might not have been used.

So I think we can't use lists in unions.

certik:
I think this looks good otherwise.

integration_tests/lp_union_test_03.f90
swamishiju: Mm true that does make sense - I did it since it was part of my original GSOC proposal. It could be possible if we somehow made tagged unions. Idk how that would work but knowing the type maybe we can do smtg.

certik: I would not do this at all for now. First we need to implement proper deallocation and we can come back to this, but yes, it would have to be a tagged union, like `std::variant`, so that the compiler knows which member is active.

swamishiju:
@certik I removed the list in union and commented out the related tests. The CI failure seems to be a random error - should I open an issue for that?

certik:
I reported it at https://github.com/lfortran/lfortran/issues/8238 and restarted the job.

integration_tests/CMakeLists.txt
```diff
@@ -2576,6 +2576,7 @@ RUN(NAME lp_dict_test_04_ LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 # LFortran extensions (union)
 RUN(NAME lp_union_test_01 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran 
 RUN(NAME lp_union_test_02 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran 
+# RUN(NAME lp_union_test_03 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran  #TODO: Uncomment after proper de-alloc
```
certik: You need to enable this, and remove the list in union below.

certik:
> I removed the list in union and commented out the related tests.

That's not a good fix, as this will not test the changes in this PR. We should not support lists in unions at all, as it is not clear how to do. There is no such things as "proper dealloc". I think this can only be done for tagged unions, but not for the union as we have it. So we should not support it. But you need to still test the union in LFortran.

So let's remove the list from union, and then test everything. Also please remove all the "lp" stuff from tests names.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -163,7 +163,7 @@ namespace LCompilers {
                 break ;
             }
             case ASR::ttypeType::UnionType: {
-                llvm_mem_type = getUnion(mem_type, module);
+                llvm_mem_type = getUnion(ASR::down_cast<ASR::Union_t>(member->m_type_declaration), module);
```
swamishiju, line 162: @certik Adding/Removing lists from union is simply putting an extra case statement here. 
I've also added an error if you use python types inside union.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2077,22 +2083,36 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                } else if ( ASR::is_a<ASR::List_t>(*var_type)
+                            || ASR::is_a<ASR::Dict_t>(*var_type) 
+                            || ASR::is_a<ASR::Set_t>(*var_type)
+                            || ASR::is_a<ASR::Tuple_t>(*var_type)) {
+
+                    diag.add(diag::Diagnostic(
+                        "Type `" + ASRUtils::type_to_str_fortran(var_type) + "` is not allowed inside Union",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {})}));
+                    throw SemanticAbort();
```
swamishiju, line 2110: The aforementioned error

swamishiju:
@certik  Is this another flaky test or some issue in the parser refactor?

https://github.com/lfortran/lfortran/actions/runs/16766756070/job/47473248359

certik:
It's not a flaky test, you have to regenerate the diff in `ci/parser.yy.patch` when you modify parser.yy.

swamishiju:
> It's not a flaky test, you have to regenerate the diff in `ci/parser.yy.patch` when you modify parser.yy.

How do i do that?

certik:
Currently by hand. Take your `parser.yy`, and do the modifications in the patch by hand. Then `git diff > ci/parser.yy.patch`.

certik:
The last commit is incorrect. We can write a script to generate the patch file automatically, but right now you need to reproduce what it is doing by hand.

certik:
Just use this script: https://github.com/lfortran/lfortran/pull/8270, it will do it automatically.

swamishiju:
@certik The script worked great, I think its good to go

Issue #7963, opened 2025-07-03
Array constructor with derived types causes semantic error


HarshilShah1804:
MRE
```fortran
module string_module
  type :: string_t
    character(len=:), allocatable :: val
  end type

  type :: key_string_array_pair_t
  contains
    procedure :: get_json_value
  end type

contains

  function get_json_value(this, key, mold) result(string_array)
    class(key_string_array_pair_t), intent(in) :: this
    character(len=*), intent(in) :: key
    type(string_t), dimension(:), intent(in) :: mold
    type(string_t), allocatable, dimension(:) :: string_array

    allocate(string_array(1))
    string_array(1)%val = "example"
  end function

end module

program test_semantic_error
  use string_module
  type(key_string_array_pair_t) :: key_string_array_pair
  type(string_t), allocatable, dimension(:) :: string_array

  associate(string_array => key_string_array_pair%get_json_value(key="lead singer", mold=[string_t::]))
    print *, string_array(1)%val
  end associate
end program
```

LFortran
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
semantic error: Empty array constructor is not allowed
  --> examples/expr2.f90:31:90
   |
31 |   associate(string_array => key_string_array_pair%get_json_value(key="lead singer", mold=[string_t::]))
   |                                                                                          ^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

GFortran
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
 example
```

HarshilShah1804:
Simlar to #7944

HarshilShah1804:
Simplified MRE
```fortran
MODULE m
        TYPE tp
           INTEGER(4) i
        END TYPE tp
END MODULE m

PROGRAM t
        USE m
        CALL s([tp :: ])
END PROGRAM t

SUBROUTINE s(va)
        USE m
        type(tp) va(:)
        WRITE(*,*)va
END SUBROUTINE s
```

PR #7976, opened 2025-07-05
fix: don't allow type declaration in array constructor

### LINKED ISSUE #7944: Use `[tp :: v1,v2,v3]` instead of `[TYPE(tp) :: v1,v2,v3]` for array constructor
Use `type-spec` instead of `declaration-type-spec` in array constructor:

https://fortran-lang.discourse.group/t/derived-type-declaration-in-array-constructor/9897/8

Example:
```fortran
MODULE m
        TYPE tp
           INTEGER(4) i
        END TYPE tp
END MODULE m

PROGRAM t
        USE m
        TYPE (tp) v1,v2,v3
        v1%i = 1
        v2%i = 2
        v3%i = 3
        CALL s([TYPE(tp) :: v1,v2,v3]) ! incorrect
        CALL s([tp :: v1,v2,v3]) ! correct
END PROGRAM t

SUBROUTINE s(va)
        USE m
        TYPE(tp) va(3)
        WRITE(*,*)va
END SUBROUTINE s
```

HarshilShah1804:
Fixes #7944 
Towards #7963


tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -757,3 +757,9 @@ semantic error: Variable protected_module_my_class_obj is PROTECTED and cannot a
     |
 341 |     protected_module_my_class_obj%value = 42
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+semantic error: Invalid syntax for array constructor
+   --> tests/errors/continue_compilation_1.f90:343:11
+    |
+343 |     arr = [type(MyClass) :: v1, v2, v3]
+    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
certik: Since we are spending time to give an error message here, let's give a great error message:

* `semantic error: Invalid syntax of derived type type for array constructor`
* highlight just the `type(MyClass)`
* Add a secondary label `help: use just the derived type name 'MyClass', without the keyword 'type'`.

That should make it very clear what is wrong with the syntax and how to fix it.

HarshilShah1804: Applied the suggestions.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -757,3 +757,9 @@ semantic error: Variable protected_module_my_class_obj is PROTECTED and cannot a
     |
 341 |     protected_module_my_class_obj%value = 42
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+semantic error: Invalid syntax of derived type type for array constructor
```
HarshilShah1804: type is present twice.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5631,6 +5631,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             ASR::symbol_t *type_declaration;
             type = determine_type(x.base.base.loc, sym, x.m_vartype, false, false,
                 dims, type_declaration, ASR::abiType::Source);
+            if (ASR::is_a<ASR::StructType_t>(*type)) {
+                std::string derived_type_name = ASRUtils::symbol_name(ASR::down_cast<ASR::StructType_t>(type)->m_derived_type);
```
HarshitaKalani, line 5635: Although we're removing the `m_derived_type` with Struct refactoring, but it's fine, we'll update this while rebasing.

HarshitaKalani (approved):

src/lfortran/semantics/ast_common_visitor.h
certik: Thanks! I think we have access to the symbol in `type_declaration`, so hopefully it's not complicated.

certik: @HarshilShah1804 how does this catches `type(tp)`, but not `tp`? Don't they both have a `StructType` ttype?

HarshilShah1804: > @HarshilShah1804 how does this catches type(tp), but not tp? Don't they both have a StructType ttype?

This is the AST for ArrayInitializer:
```
ArrayInitializer(decl_attribute? vartype, identifier? classtype, expr* args)
```

Here, the `type(tp)` is treated as the `vartype` attribute and `tp` is treated as `classtype`. Therefore, we can check if `vartype` attribute is a `StructType` and throw error based on that.

certik: Ok. I added a test for this in https://github.com/lfortran/lfortran/pull/7991 to ensure we test both ways (maybe we already had it, but I couldn't find it). So if we decide to refactor this, we test both paths and ensure both the correct and the error path keeps working.

PR #7977, opened 2025-07-05
chore: gep to gep2 in visit_RuntimePolymorphicSubroutineCall and visi…

sriganeshres:
…t_RuntimePolymorphicFunctionCall


sriganeshres:
@certik I removed ifdef and used bitcast to change. The problem was pointer was of polymorphic and pointee was polymoprhic extended. Same inside so they have been bitcasted.

sriganeshres:
Will use this to make more gep to gep2 changes.

PR #7985, opened 2025-07-06
fix: use m_classtype name to get correct type for struct array initializer

HarshilShah1804:
Fixes #7963 
Addresses https://github.com/lfortran/lfortran/issues/7944#issuecomment-3040973524

PR #7991, opened 2025-07-06
Add a test for [MyClass :: v1, v2, v3]

certik:
A test for #7944.


HarshilShah1804:
@certik I have also added a test for this in the PR #7985. So, we can remove the test from either one of the PRs.

certik:
@HarshilShah1804 I just saw your PR. Let's keep both.

HarshilShah1804:
Okay.

PR #7992, opened 2025-07-06
fix: make assoc variable pointer to structType in select type

### LINKED ISSUE #7984: Handle association of `m_assoc_name` variable to `structInstanceMember` in select type statements
MRE:
```fortran
module m_values
  implicit none

  type, abstract :: base
  end type base

  type, extends(base) :: int_val
    integer :: i
  end type int_val

  type :: temp
    class(base), allocatable :: bas
  end type

contains

  subroutine print_value(val, x)
    class(temp), intent(in), target :: val
    class(real_val), intent(out) :: x
    select type(v => val%bas)
    type is(int_val)
    end select
  end subroutine print_value

end module m_values

program test_select_type
  use m_values
  implicit none

  class(temp), allocatable :: val
  class(real_val), allocatable :: r_val

    call print_value(val, r_val)

end program test_select_type
```

LFortran Error:
```console
  store %base* %22, %base_polymorphic* %v, align 8
 %base_polymorphic = type { i64, %base* }
```

jinangshah21:
Fixes #7984


certik:
The error seems real:
```
#20 5.056 ../integration_tests/select_type_02.f90 * asr
#20 5.056 The JSON metadata differs against reference results
#20 5.056 Reference JSON: tests/reference/asr-select_type_02-6e04a0b.json
#20 5.056 Output JSON:    tests/output/asr-select_type_02-6e04a0b.json
#20 5.056 Omitting 10 identical items
#20 5.056 Differing items:
#20 5.056 {'stdout_hash': '5e13c325be4dcb753ff9618c93b1acac0e3d6e522a553a9bb011331d'} != {'stdout_hash': '11aab1936975ae228e2fd32af9a02aba777c22b364a5f27e738b148e'}
#20 5.056 Diff against: tests/reference/asr-select_type_02-6e04a0b.stdout
#20 5.056 355c355
#20 5.056 <                                                             .false.
#20 5.056 ---
#20 5.056 >                                                             .true.
#20 5.056 449,450c449,450
#20 5.056 <                                                             .false.
#20 5.056 <                                                             4 point
#20 5.056 ---
#20 5.056 >                                                             .true.
#20 5.056 >                                                             4 color_point
```

tests/reference/asr-select_type_02-6e04a0b.stdout
```diff
@@ -446,8 +446,8 @@
                                                         (StructType
                                                             []
                                                             []
-                                                            .false.
-                                                            4 point
+                                                            .true.
+                                                            4 color_point
```
certik, line 450: This change is correct?

jinangshah21: Yes, because when we use `type is` in select type assoc variable should not be class type. Instead it should be of type defined in type is (I.e. here `color_point`)

PR #7993, opened 2025-07-06
Revert "fix: use m_classtype name to get correct type for struct array initializer"

certik:
Reverts lfortran/lfortran#7985


certik:
This should pass all tests, but the fix in https://github.com/lfortran/lfortran/pull/7994 is even better, it keeps all integration tests, just relaxes the too strong error for now.

certik:
Fixed in https://github.com/lfortran/lfortran/pull/7994.

PR #8002, opened 2025-07-08
LP type argument handling for C backend (LP sync)

swamishiju:


swamishiju:
Typically, the conversion from function to subroutine was only applied to functions that return structs and, I believe, descriptorstrings. When I was working on exposing LP types, I modified the system so that functions returning lists, dictionaries, and other similar types would also trigger this conversion. However, this change wasn’t reflected in the C backend, likely because those types are handled differently in that context.

I’m not sure why this conversion was originally only enabled for CPtr types, though.

swamishiju:
Some intrinsics don't have return type as pointer so we likely need to refactor those

Ex: 
```c
void __lpython_overloaded_1__polar(double_complex_t x, struct tuple_r64_r64 _lpython_return_variable)
```

certik:
Let's add tests for this. 

Is there a test that was previously failing, but now it works?

swamishiju:
C tests weren't enabled for those in LF. I'll enable them after I get this working again on the LP side

certik:
I see -- you are testing each commit here in https://github.com/lcompilers/lpython/pull/2860? Very good, yes, get the LPython sync PR working first with any changes you need here. Then if things pass here, we can merge without tests, then you can depend on "main" from LPython and finish the sync PR there.

Only then you can add some tests here.

src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -1392,7 +1391,11 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
         } else if ( is_target_tup && is_value_tup ) {
             ASR::Tuple_t* tup_target = ASR::down_cast<ASR::Tuple_t>(ASRUtils::expr_type(x.m_target));
             std::string dc_func = c_ds_api->get_tuple_deepcopy_func(tup_target);
-            src += indent + dc_func + "(" + value + ", &" + target + ");\n";
+            // TODO: Proper intent checking instead of target name
+            if (target == "_lpython_return_variable")
```
kmr-srbh, line 1395: Please see https://github.com/lfortran/lfortran/pull/8011#discussion_r2195583718 for a way to do this.

Issue #7972, opened 2025-07-05
support keyword argument for `execute_command_line`


Pranavchiku:
```fortran
program intrinsics_387
    implicit none
    character(len=:), allocatable :: temp
    integer :: stat = -1

    temp = "echo Hello from Fortran!"
    call execute_command_line(temp)

    temp = "export LFORTRAN_TEST_VAR_INTRINSIC=42; echo $LFORTRAN_TEST_VAR_INTRINSIC"
    call execute_command_line(temp)

    call execute_command_line(temp, exitstat=stat)
    print *, "Exit status:", stat ! TODO: support keyword arguments in execute_command_line
    ! if ( stat /= 0 ) error stop ! TODO: uncomment once keyword arguments are supported

end program intrinsics_387

```

@HarshitaKalani @parth121101 either of you can pick this up.

PR #8006, opened 2025-07-08
support keyword args in execute_command_line intrinsic

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/7972


Pranavchiku (approved):
Nice, thank you! Let's enable this for all intrinsics and test a plethora of combinations.

PR #8010, opened 2025-07-09
feat : Implement Submodules by mangling parent module name as prefix for module Functions

AnshMehta1:


AnshMehta1:
Tests added also works well with Separate Compilation and Separate File Compilation.

AnshMehta1:
For #7270, more changes needs to be done which I will handle in next PR. 
Also, for https://github.com/lfortran/lfortran/pull/7961#issuecomment-3036569212, we need to modify ASR of submodule while compiling it, which I will handle in subsequent PRs.

PR #8011, opened 2025-07-09
dict_deepcopy in C backend

swamishiju:
Did the same change to `dict_deepcopy` as was done to `tuple_deepcopy` #8002 

The condition is still some-what hacky, but I think thats a lower priority issue [Fixed]


src/libasr/codegen/asr_to_c_cpp.h
```diff
@@ -1399,7 +1399,11 @@ PyMODINIT_FUNC PyInit_lpython_module_)" + fn_name + R"((void) {
         } else if ( is_target_dict && is_value_dict ) {
             ASR::Dict_t* d_target = ASR::down_cast<ASR::Dict_t>(ASRUtils::expr_type(x.m_target));
             std::string dc_func = c_ds_api->get_dict_deepcopy_func(d_target);
-            src += indent + dc_func + "(&" + value + ", &" + target + ");\n";
+            // TODO: Proper intent checking instead of target name
+            if (target == "_lpython_return_variable")
```
kmr-srbh: If `target` is `ASR::Var_t` and it's `m_v` field is a `ASR::Variable_t` node, then you can down cast it to `ASR::Variable_t` and check it's `intent`.

PR #8012, opened 2025-07-10
ref: rename `~abstract_type` to `~unlimited_polymorphic_type`

Pranavchiku:
This PR patches a parallel refactoring done as a part of #5791 ( struct refactoring ). We don't want to manage two refactoring in same PR, this leads to lot of manual effort while rebasing to get LFortran binary working.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -3635,8 +3635,8 @@ inline bool is_parent(ASR::Struct_t* a, ASR::Struct_t* b) {
 
 inline bool is_derived_type_similar(ASR::Struct_t* a, ASR::Struct_t* b) {
     return a == b || is_parent(a, b) || is_parent(b, a) ||
-        (std::string(a->m_name) == "~abstract_type" &&
-        std::string(b->m_name) == "~abstract_type");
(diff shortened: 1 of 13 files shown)


src/libasr/asr_utils.h
```diff
@@ -5332,6 +5332,44 @@ static inline ASR::Enum_t* get_Enum_from_symbol(ASR::symbol_t* s) {
...
+    if (!ASRUtils::is_class_type(type)) {
+        return false;
+    }
+    return (st->n_data_member_types == 0 && st->n_member_function_types == 0
+            && ASRUtils::symbol_name(ASRUtils::symbol_get_past_external(
+                   st->m_derived_type))
+                   == std::string("~unlimited_polymorphic_type"));
+
+}
+
```
kmr-srbh, line 5372: Is it possible to remove the function `is_abstract_class_type()` completely? I think it is redundant. I had changed this function to the above during the refactor actually.

Pranavchiku: Yes, nice catch, done via https://github.com/lfortran/lfortran/pull/8014

PR #8013, opened 2025-07-10
fix : Do not return from instantiate function and populate llvm_symtab_fn for class interfaces

### LINKED ISSUE #7270: [stdlib:hashmaps] Calling type-bound procedure declared via module and implemented in submodule is causing `code generation error`

AnshMehta1:
Towards #536 
Fixes #7270
Addresses https://github.com/lfortran/lfortran/pull/8010#pullrequestreview-3004223332

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4668,7 +4668,7 @@ public:
             F = llvm_symtab_fn[h];
         } else {
             llvm::FunctionType* function_type = llvm_utils->get_function_type(x, module.get());
-            if( ASRUtils::get_FunctionType(x)->m_deftype == ASR::deftypeType::Interface ) {
+            if( ASRUtils::get_FunctionType(x)->m_deftype == ASR::deftypeType::Interface && !ASRUtils::get_FunctionType(x)->m_module ) {
                 ASR::FunctionType_t* asr_function_type = ASRUtils::get_FunctionType(x);
                 for( size_t i = 0; i < asr_function_type->n_arg_types; i++ ) {
                     if( ASRUtils::is_class_type(asr_function_type->m_arg_types[i]) ) {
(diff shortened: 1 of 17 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4668,7 +4668,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             F = llvm_symtab_fn[h];
         } else {
             llvm::FunctionType* function_type = llvm_utils->get_function_type(x, module.get());
-            if( ASRUtils::get_FunctionType(x)->m_deftype == ASR::deftypeType::Interface ) {
+            if( ASRUtils::get_FunctionType(x)->m_deftype == ASR::deftypeType::Interface && !ASRUtils::get_FunctionType(x)->m_module ) {
```
Pranavchiku, line 4671: Ahh I see, you are handling cases for class. PR name suggests that we generate object code for `module ` interfaces.

AnshMehta1: Ya, I named it generally, like whenever we have module interfaces, then we don't return from instantiate function. Will change the name.

PR #8018, opened 2025-07-10
ichar fix for ichar(char(0)) and changed the branch

### LINKED ISSUE #7576: bug: incorrect error for `ichar(char(0))`
```fortran
program main
    print *, ichar(char(0))
end program
```
```console
$ lfortran b.f90
semantic error: Argument to Ichar must have length one
 --> b.f90:2:14
  |
2 |     print *, ichar(char(0))
  |              ^^^^^^^^^^^^^^ 


$ gfortran b.f90 && ./a.out
           0
```

galactus3050:
Fixes #7576

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -343,4 +343,6 @@ program continue_compilation_1
     arr = [type(MyClass) :: v1, v2, v3]
 
     arr = [NonExistingType :: v1, v2, v3]
+
+    print *, ichar(char(0))
 end program
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -4951,11 +4951,9 @@ namespace Ichar {
 
     static ASR::expr_t *eval_Ichar(Allocator &al, const Location &loc,
             ASR::ttype_t* t1, Vec<ASR::expr_t*> &args, diag::Diagnostics& /*diag*/) {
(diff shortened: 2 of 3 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -343,4 +343,6 @@ program continue_compilation_1
     arr = [type(MyClass) :: v1, v2, v3]
 
     arr = [NonExistingType :: v1, v2, v3]
+
+    print *, ichar(char(0))
```
certik, line 347: This needs to go into integration tests, since it succeeds and we want it to.

HarshilShah1804, line 347:
```suggestion
```
This line would not throw any errors. So, you can move this test from `continue_compilation_1.f90` to integration tests.

HarshilShah1804: duplicate.

integration_tests/intrinsics_46.f90
```diff
@@ -119,4 +119,6 @@ program intrinsics_46
     print *, ichar(e)
     if (any(ichar(e) /= x)) error stop
 
+    print *, ichar(char(0))
```
certik, line 122: Can you also add a check what the result should be?
```
if (ichar(char(0)) /= ...)) error stop
```

tests/errors/continue_compilation_1.f90
certik: Remove this from here.

PR #8022, opened 2025-07-11
FEAT: default initialization of module-level derived type variables

sriganeshres:
fixes: #7394 

* Added a default initialize (`init_value`) for structs so that default values can be intialized.

### DIFF
--- a/src/libasr/codegen/llvm_utils.h
+++ b/src/libasr/codegen/llvm_utils.h
@@ -306,6 +306,9 @@ namespace LCompilers {
 
             llvm::Value* lfortran_str_cmp(llvm::Value* left_arg, llvm::Value* right_arg,
                                           std::string runtime_func_name, llvm::Module& module);
+            // Converts a constant ASR expression into an equivalent LLVM Constant.
+            // Supports integer, logical, and real constants used in struct initializers.
+            // Throws an exception for unsupported expression types.
             llvm::Constant* create_llvm_constant_from_asr_expr(ASR::expr_t* expr,
                                                                llvm::Module* module);
 


sriganeshres:
@kmr-srbh Did you have this in your mind?

sriganeshres:
@certik can you review this?

src/libasr/codegen/llvm_utils.h
```diff
@@ -306,6 +306,8 @@ namespace LCompilers {
 
             llvm::Value* lfortran_str_cmp(llvm::Value* left_arg, llvm::Value* right_arg,
                                           std::string runtime_func_name, llvm::Module& module);
+            llvm::Constant* create_llvm_constant_from_asr_expr(ASR::expr_t* expr,
+                                                               llvm::Module* module);
```
certik, line 313: Document what this function is doing. I think it expects a IntegerConstant or RealConstant or LogicalConstant (nothing else, not even an expression of these --- but this is not needed, since the `value` member already contains a constant) and it returns an LLVM constant.

integration_tests/derived_type_with_default_init.f90
```diff
@@ -0,0 +1,28 @@
...
+        print *, "a = ", self%a
+        print *, "i = ", self%i        
+    end subroutine config
+
+end module module_derived_type_with_default_init
+
+program main
+    use module_derived_type_with_default_init, global => global_myType
+    call global % config()
+end program main
```
certik: As we discussed yesterday, please create an MRE from this. It looks like this is an issue with global structs, or possibly any structs initialization.

Also, test in the test using `if (...) error stop` to ensure the values got initialized correctly.

certik:
I think this approach works! Just improve the test (see my comment) and I think this is good to go.

integration_tests/derived_type_with_default_init.f90
sriganeshres: yes i have made a test for global structs i have added this single test beacuse i felt it compasses both global struct and function call. Yes i will make it if error stop.

sriganeshres:
I have added a new test.

src/libasr/codegen/llvm_utils.h
sriganeshres: Like it will create a constant for all primitive types. like we created a = .true. so we create a new constant.
i = 2 it will create a integer constant.

integration_tests/derived_type_with_default_init.f90
```diff
@@ -0,0 +1,28 @@
...
+            procedure, public, pass(self) :: config
+    end type myType
+
+    type(myType) :: global_myType
+
+contains
+    subroutine config(self)
+        class(myType), intent(in) :: self
+        print *, "a = ", self%a
+        print *, "i = ", self%i
```
certik: Add
```
    if (self % a .neqv. .true.) error stop
    if (self % i .ne. 2) error stop
```

src/libasr/codegen/llvm_utils.h
certik: Document it as a comment in the header file.

integration_tests/derived_type_with_default_init.f90
sriganeshres: Done.

sriganeshres:
I have added if else in the first case as well. Once CI passes i will ping ypu please merge this.

certik:
Please fix this too: https://github.com/lfortran/lfortran/pull/8022#discussion_r2201420524

sriganeshres:
I have added a 3 line comment.

kmr-srbh:
> @kmr-srbh Did you have this in your mind?

Please see #7418 for improving the constant handling. As @certik mentioned in https://github.com/lfortran/lfortran/pull/8022#discussion_r2201195186, I think we do not need a helper function for creating the constants. You may check if initialising character and character arrays work. If they work, that's great, if not, you could open up an issue specifically for that case when we merge this.

PR #8049, opened 2025-07-13
doc: remove incorrect information

Pranavchiku:


doc/src/asr/asr_nodes/statement_nodes/allocate.md
```diff
@@ -1,6 +1,6 @@
 # Allocate
 
-Allocate function, a **statement (stmt)** node.
+Allocate, a **statement (stmt)** node.
```
certik: I think that's still incorrect. This might be better:
```suggestion
Allocate an allocatable variable.
```

PR #8050, opened 2025-07-13
feat: group compiler options and flags

parth121101:
I refactored the logging of compiler options to group and log them by category. This allows us to log only the relevant category of options instead of all at once.
The behaviour would be like following:

Screenshot 2025-07-13 at 10 22 35 PM

We can print options with group name just like below:

Screenshot 2025-07-13 at 10 23 31 PM


tests/reference/mod_to_asr-mod1-14-bb0bd17.stderr
```diff
@@ -1,2 +1 @@
 The following argument was not expected: --no-indent
-Run with --help for more information.
```
certik, line 2: This might be another but that it shouldn't return an error. We should investigate.

certik:
I think the idea is good, but CI has to pass, then we have to carefully review that nothing got broken.

Issue #7006, opened 2025-04-14
Idea: generate ASR.asdl, asr_verify.cpp and doc/src/asr/ from the same source


certik:
Say for the `Array` ASR node we have the following verifier:
```
void visit_Array(const Array_t& x) {
        require(!ASR::is_a<ASR::Allocatable_t>(*x.m_type),
            "Allocatable cannot be inside array");
        visit_ttype(*x.m_type);
        require(x.n_dims != 0, "Array type cannot have 0 dimensions.")
        require(!ASR::is_a<ASR::Array_t>(*x.m_type), "Array type cannot be neste
d.")
        _processing_dims = true;
        for (size_t i = 0; i < x.n_dims; i++) {
            visit_dimension(x.m_dims[i]);
        }
        _processing_dims = false;
    }
```
The following ASR definition:
```
| Array(ttype type, dimension* dims, array_physical_type physical_type)
```
And currently no documentation.

One should put these together and ensure the node is well designed and well tested. For example it seems we are currently not checking in the verifier that the `dims` correspond to the physical_type correctly (such as that DescriptorArray has each dimension equal to None). By having them together we can ensure that we test every combination and that the documentation is current.

I would generate the code directly into git, so nothing has to change in our build system.

Another idea is to use our `asr_verify.cpp` as the "master source" that would have C++ comments with the ASR definition of `Array` above `visit_Array`, and documentation for `Array`, then we have a Python script that extracts these comments to create `ASR.asdl` and or documentation in `doc/`.

Another idea: we keep `ASR.asdl` and `asr_verify.cpp` as master sources, without any comments, to keep them short and maintainable. Then we have a separate Markdown document that lists the name and documentation, but does not list the ASR definition nor the `verify_*` function. We add a Python script that will take this Markdown, looks up the definition in ASR.asdl as well as the verifier in `asr_verify.cpp`, and puts them all together into our documentation in `doc/`. 

Actually, probably even the best way is the previous point, but the Markdown file will be directly in `doc/`, but the script would go in and update the ASR definition as well as a verifier (if it exists) for each node file in place, overwriting what was there.

## Summary

Summary of the best idea so far is:

* we keep ASR.asdl and asr_verify.cpp as master sources, without any comments, to keep them short and maintainable. 
* Then we have a separate Markdown document in `doc/` that lists the name and documentation, and uses some structure to list the ASR definition and the verifier C++ code, directly as text.
[...]

PR #8052, opened 2025-07-13
feat: generate reference document from `ASR.asdl` and `asr_verify.cpp`

Pranavchiku:
Towards #7006

### DIFF
--- a/generate_reference_doc.py
+++ b/generate_reference_doc.py
@@ -1,0 +1,368 @@
+import os
+import re
+import hashlib
+import sys
+import difflib
+
+from pathlib import Path
+from collections import defaultdict
+
+ASDL_PATH = "src/libasr/ASR.asdl"
+VERIFY_CPP_PATH = "src/libasr/asr_verify.cpp"
+DOCS_DIR = Path("doc/reference")
+NODE_TYPES = ["unit", "symbol", "stmt", "expr", "ttype", "omp_clause", "schedule_type"]
+
+RED = "\033[91m"
+GREEN = "\033[92m"
+CYAN = "\033[96m"
+BOLD = "\033[1m"
+RESET = "\033[0m"
+BLUE = "\033[94m"
+
+
+def parse_asdl(asdl_text):
+    nodes = defaultdict(list)     # concrete nodes with parentheses
+    enums = defaultdict(list)     # enum values (no parentheses)
+    structs = {}                  # structs
+
+    current_section = None
+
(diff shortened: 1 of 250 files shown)


generate_reference_doc.py
```diff
@@ -0,0 +1,368 @@
...
+                        if ")" in next_line:
+                            break
+                if "(" in part:
+                    name = part.split("(")[0].strip()
+                    nodes[current_section].append((name, part))
+                else:
+                    enums[current_section].append(part)
+        line_index += 1
+
+    return nodes, enums, structs
```
certik: Why not reusing the ASDL parser that we already have, so that we do not need to maintain the code above, that seems quite fragile?

certik:
Thanks a lot for this! I'll play with this soon and provide more feedback.

generate_reference_doc.py
Pranavchiku: Yup, I’ll use that.

generate_reference_doc.py
```diff
@@ -0,0 +1,325 @@
...
+    return restriction_map
+
+
+
+def generate_enum_file(lhs, values):
+    output_path = DOCS_DIR / f"{lhs}.md"
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+
+    md = [
+        "<!-- This is an automatically generated file. Do not edit it manually. -->",
```
certik, line 116: I suggest we edit this manually to actually put useful documentation into these files. Then the script should merge it.

generate_reference_doc.py
```diff
@@ -0,0 +1,325 @@
...
+        "",
+        f"# {lhs}",
+        "",
+        f"`{lhs}` is an enum with the following values:",
```
certik, line 120: I think it's not an enum. `ttype` is a "base class" in a sense, but not an enum.

generate_reference_doc.py
Pranavchiku: Then once the script runs again it will overwrite those. Or maybe we modify script to not override it rather if a file exists then populate other information automatically. We need to set some rules then to manually edit files like put a comment <!— Manual Start —> …. <!— Manual Ends —> for helping script to understand that this portion needs to be untouched.

certik: The script should override the automatic parts, but merge user parts of the output.

Pranavchiku:
I'm marking this as draft, low priority at the moment.

PR #8055, opened 2025-07-14
Unsigned int print fix

swamishiju:
Previously print would just give `Unknown Type` since `uint` wasn't handled in `print_into_string`


certik:
This is fine. Let's also expose unsigned in the frontend and add tests.

PR #8056, opened 2025-07-14
Support for string desc in C backend

swamishiju:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -688,7 +687,11 @@ namespace LCompilers {
                 ASR::String_t* v_type = ASR::down_cast<ASR::String_t>(asr_type);
                 a_kind = v_type->m_kind;
                 if (arg_m_abi == ASR::abiType::BindC) {
-                    type = character_type;
+                    if (v_type->m_physical_type == ASR::string_physical_typeType::DescriptorString) {
+                        type = string_descriptor;
```
swamishiju, line 691: Set type according to physical type

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -10372,7 +10372,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                             tmp = llvm_utils->CreateLoad(tmp);
                                         }
                                     } else {
-                                        if (!arg->m_value_attr && !ASR::is_a<ASR::String_t>(*arg_type)) {
+                                        if (!arg->m_value_attr && !ASR::is_a<ASR::String_t>(*ASRUtils::extract_type(arg_type))) {
```
swamishiju, line 10375: Since string descriptor is usually inside allocatable this would not normally be true for it

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -688,7 +687,11 @@ namespace LCompilers {
...
+                    } else {
+                        type = character_type;
+                    }
```
certik, line 694: @assem2002 is this a correct change?

assem2002: That seems a bit awkward, I think `BindC` in main uses llvm type = `character_type` and the physicalType should've always been `PointerString`.
If that unblocks some needed work, You can go for it.

assem2002:
May you mention the issue this PR is trying to fix @swamishiju

src/libasr/codegen/llvm_utils.cpp
certik: @swamishiju does this unblock you in some way? Is there some Fortran code that shows what this fixes?

swamishiju: This was a fix for `@ccall` in LP

```py
@ccall 
def f_str_i32(x:str) -> i32 : 
    pass
```

```llvm
declare i32 @f_str_i32(i8**)
...
  %casted_string_ptr_to_desc = alloca %string_descriptor, align 8
 i8**  %32 = call i32 @f_str_i32(%string_descriptor* %casted_string_ptr_to_desc)
```

Initially it generated an llvm function with `i8**` for BindC functions with string arguments.

I had two options, create another cast for this case to convert string descriptor to pointer string, or change the argument based on physical type. After that the llvm test works but CPython and C tests fail.

The C-failure is unrelated its just physical cast not being implemented.
The CPython failure is some size initialization issue i need to look into

swamishiju: It was for this PR but it needs some more changes https://github.com/lcompilers/lpython/pull/2864

certik:
@swamishiju go ahead and fix the LPython PR first to see that it works. Then we can consider merging this. 

Even better would be to submit a test case that breaks in LFortran and this PR fixes it.

Pranavchiku:
conflicts need to be resolved

swamishiju:
I need to look into this again, I have a feeling these changes might not be required - but ye I need to look into it.

PR #8066, opened 2025-07-15
Relocated semantic check for bindC character length

rahulrangers:
This PR moves the existing semantic check for BIND(C) procedures to the end of the `TypeCharacter` handling block in the `determine_type()`


assem2002:
Thanks for moving it

assem2002 (approved):

Issue #8070, opened 2025-07-15
bug: setting array dimensions using non-constant variable allowed inside `ASR::Struct_t`


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   implicit none

   integer :: n = 2

   type :: matrix
      integer :: elements(n)
   end type

   type(matrix) :: m
   m%elements = [1, 2]
end program
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
test.f90:49:26:

   49 |       integer :: elements(n)
      |                          1
Error: Variable 'n' cannot appear in the expression at (1)
(lf) saurabhkumar@Mac lfortran % lfortran test.f90           
Assertion failed: (PointeeType == cast<PointerType>(Ptr->getType()->getScalarType())->getElementType()), function Create, file Instructions.h, line 941.
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2653
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 1185
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 394
    if (res.ok) {
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 12316
    v.visit_asr((ASR::asr_t&)asr);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 345
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 320
    case asrType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 346
    void visit_unit(const unit_t &b) { visit_unit_t(b, self()); }
[...]

PR #8074, opened 2025-07-16
Report Error for non-parameter variables in derived type bounds.

rahulrangers:
fixes https://github.com/lfortran/lfortran/issues/8070


tests/errors/type_dim_01.f90
```diff
@@ -0,0 +1,12 @@
...
+
+   integer :: n = 2
+
+   type :: matrix
+      integer :: elements(n)
+   end type
+
+   type(matrix) :: m
+   m%elements = [1, 2]
+end program
```
HarshilShah1804: Please move this test to `continue_compilation_*.f90` files instead of adding a separate test.

tests/reference/asr-type_dim_01-34f98ab.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Variable `n` cannot appear in the expression
```
certik: The error should say why it cannot appear in the expression.

tests/reference/asr-type_dim_01-34f98ab.stderr
```diff
@@ -0,0 +1,5 @@
...
+ --> tests/errors/type_dim_01.f90:7:7
+  |
+7 |       integer :: elements(n)
```
certik: Highlight just `n`.

tests/errors/type_dim_01.f90
```diff
@@ -0,0 +1,12 @@
+program main
+   implicit none
+
+   integer :: n = 2
```
certik: Also add an integration test with `integer, parameter :: n = 2`, which *is* allowed. I think you might not be checking for this case. Update: you are checking for it `if (dim_var->m_storage != ASR::storage_typeType::Parameter) {`. See if we have a test for it, if not, add it.

tests/errors/type_dim_01.f90
rahulrangers: I tried having the test in `continue_compilation_*.f90`. But I am getting this error :
```
ASR verify pass error: The Program::m_symtab->parent is not the right parent
   --> tests/errors/continue_compilation_1.f90:64:1 - 350:11
    |
 64 |    program continue_compilation_1
    |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
    |
350 |    end program
    | ...^^^^^^^^^^^ failed here```

Comment (HarshilShah1804) [reply tests/errors/type_dim_01.f90]: I believe this is a bug. We only encounter this error if we use the flag `--continue-compilation`. We weren't encountering this at first place because we were only running the program, whereas in `continue_compilation_1.f90` we use the flag `--continue-compilation`. I guess we can open a new issue for that.

Comment (certik) [reply tests/errors/type_dim_01.f90]: @rahulrangers please create an MRE from the bug you discovered and open up an issue.

Comment (HarshilShah1804) [reply tests/errors/type_dim_01.f90]: I believe the MRE is that particular test which you have added. Just use the flag `--continue-compilation` while running it.

Comment (rahulrangers) [reply tests/errors/type_dim_01.f90]: Made an issue reporting the above error.

**Comment by certik** [comment]
What is the status of this --- there is no problem in main, but this PR introduces an ASR verify failure? If so, we need to fix the failure first.

File: tests/errors/type_dim_01.f90 (hunk 1 above)
Comment (HarshilShah1804) [reply tests/errors/type_dim_01.f90]: Thanks. @certik Can this be merged with a different error test now if the CI passes or shall we first fix that ASR Verify Pass error ?

Comment (certik) [reply tests/errors/type_dim_01.f90]: Is the ASR verify issue in main also, or is it only triggered by this PR?

Comment (certik) [reply tests/errors/type_dim_01.f90]: See https://github.com/lfortran/lfortran/issues/8085#issuecomment-3084578963. If it doesn't trigger in main, then we need to fix it as part of this PR, which causes it.

Comment (rahulrangers) [reply tests/errors/type_dim_01.f90]: It is only triggered by this PR in main we are getting this
```Abort: Signal SIGABRT (abort) received```

Comment (rahulrangers) [reply tests/errors/type_dim_01.f90]: @certik I will check Why I am getting this error and try to fix in this PR itself.

**Comment by rahulrangers** [comment]
@certik  Resolved the `ASR verify pass error` by aborting derived type construction as soon as a member caused a semantic error. This prevented partial or invalid ASR from being registered.

File: integration_tests/derived_types_71.f90 [hunk 5]
```diff
@@ -17,4 +17,4 @@ program derived_type_71
     x => z
     y = build_target_ptr(x)
     if (y%ptr%key /= 12) error stop
-end program  
\ No newline at end of file
+end program
```
certik: Don't modify this file, I think it also caused a conflict.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1894,7 +1894,12 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
         is_derived_type = true;
         ASR::accessType dflt_access_copy = dflt_access;
         for (size_t i=0; i<x.n_items; i++) {
+            try {
             this->visit_unit_decl2(*x.m_items[i]);
```
certik: I think the indentation here is wrong.

tests/reference/asr-template_error_06-1cbc8a9.json
```diff
@@ -5,9 +5,9 @@
...
     "outfile": null,
     "outfile_hash": null,
-    "stdout": null,
-    "stdout_hash": null,
+    "stdout": "asr-template_error_06-1cbc8a9.stdout",
+    "stdout_hash": "830f5d5f71b7b2e193a50a725c71eea040dad33965f830de0abfc6a5",
     "stderr": "asr-template_error_06-1cbc8a9.stderr",
     "stderr_hash": "bb2807cb7bd5ec770c93e19e96535a404d5ffd6277814d2f741d1b06",
-    "returncode": 2
+    "returncode": 0
```
certik: I think this now gave an error message, but succeeded. So that's a bug.

src/lfortran/semantics/ast_symboltable_visitor.cpp
certik: I wonder if this change is incorrect, as it is now "passing" incorrect code that should fail, see below.

rahulrangers:
@certik All queries are now resolved. The reason for the previous returncode being 0 was that I was returning quickly instead of throwing the error.

PR #8081, opened 2025-07-16
feat : Implement Submodules for separate files without `separate compilation` flag using command line submodule arguments

AnshMehta1:


AnshMehta1:
In my system, this tests are passing. I don't know why do they fail here.

certik:
The failure is:
```
[ 96%] Building Fortran object CMakeFiles/separate_compilation_21.dir/separate_compilation_21.f90.o
semantic error: Module 'submod_separate_compilation_22' modfile was not found
 --> /home/runner/work/lfortran/lfortran/integration_tests/separate_compilation_22.f90:1:1 - 8:11
  |
1 |    program separate_compilation_22
  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
  |
8 |    end program
  | ...^^^^^^^^^^^ 


```

That should be possible to reproduce locally.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -457,6 +457,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            ASR::Module_t* submodule = ASRUtils::load_module(al, tu_symtab, submodule_name, x.base.base.loc,
+                                                                false, compiler_options.po, true,
+                                                                [&](const std::string &msg, const Location &loc) {
+                                                                    diag.add(diag::Diagnostic(
+                                                                        msg, diag::Level::Error, diag::Stage::Semantic, {
+                                                                            diag::Label("", {loc})}));
+                                                                    throw SemanticAbort();
+                                                                }, lm, false
+                                                            );
+            LCOMPILERS_ASSERT(submodule != nullptr && std::string(submodule->m_name) == submodule_name);
```
Pranavchiku, line 473:
```suggestion
            LCOMPILERS_ASSERT(submodule != nullptr && std::string(submodule->m_name) == submodule_name);
            if ( compiler_options.separate_compilation && !startswith(submodule->m_name, "lfortran_intrinsic") ) {
                submodule->m_symtab->mark_all_variables_external(al);
            }
```

integration_tests/CMakeLists.txt
```diff
@@ -2456,6 +2456,10 @@ RUN(NAME separate_compilation_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc E
 RUN(NAME separate_compilation_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_17a.f90 separate_compilation_17b.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_18a.f90 separate_compilation_18b.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_19a.f90 separate_compilation_19b.f90 EXTRA_ARGS --separate-compilation)
+RUN(NAME separate_compilation_20 LABELS gfortran llvm NO_LLVM_GOC EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90 EXTRA_ARGS --add-submodule submod_separate_compilation_20.mod)
```
Pranavchiku, line 2459:
```suggestion
RUN(NAME separate_compilation_20 LABELS gfortran llvm EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90 EXTRA_ARGS --add-submodule submod_separate_compilation_20.mod)
```

Pranavchiku:
CMake is unable to identify that `separate_compilation_23b.f90` is required for compiling `separate_compilation_23.f90`. It only that `separate_compilation_23a.f90` is needed. Hence the order in which files compile is:

1. separate_compilation_23a.f90
2. separate_compilation_23.f90

And `separate_compilation_23b.f90` is independent of these two.

However using the following diff you can force CMake to depend on `separate_compilation_23b.f90`

```diff
diff --git a/integration_tests/separate_compilation_23.f90 b/integration_tests/separate_compilation_23.f90
index edfde8199..270d55298 100644
--- a/integration_tests/separate_compilation_23.f90
+++ b/integration_tests/separate_compilation_23.f90
@@ -1,5 +1,7 @@
 program separate_compilation_23
     use mod_separate_compilation_23, only : open_hashmap_type
+    use submod_separate_compilation_23
+
 
     implicit none
 
```


CMake with GFortran exhibits similar behaviour just that GFortran knows how to link submodule files.

Pranavchiku:
Summary: GFortran knows while compiling a module that it will have a submodule, with this it generates `<module-name>.smod` and later uses it to check for submodules and link those together. More details in section below.

So, we can add a flag `has_submodule`, if it is `True` then we load all submodules as those will be prefixed with parent module name.

```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 374c34915..842333c21 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -10,7 +10,7 @@ unit
 
 symbol
     = Program(symbol_table symtab, identifier name, identifier* dependencies, stmt* body, location start_name, location end_name)
-    | Module(symbol_table symtab, identifier name, identifier parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, location start_name, location end_name)
+    | Module(symbol_table symtab, identifier name, identifier parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, bool has_submodule, location start_name, location end_name)
     | Function(symbol_table symtab, identifier name, ttype function_signature, identifier* dependencies, expr* args, stmt* body, expr? return_var, access access, bool deterministic, bool side_effect_free, string? module_file, location start_name, location end_name)
     | GenericProcedure(symbol_table parent_symtab, identifier name, symbol* procs, access access)
     | CustomOperator(symbol_table parent_symtab, identifier name, symbol* procs, access access)
```

----

Here is how GFortran handles it:

> b.f90

```fortran
module mod_separate_compilation_23
    implicit none
    private

    interface
        subroutine map_open_entry_sc_23()
        end subroutine map_open_entry_sc_23
    end interface

end module mod_separate_compilation_23
```


For this normal module file, when we do `gfortran -c b.f90` it generates `mod_separate_compilation_23.mod` only.

```console
% gfortran -c b.f90 
% ls *.mod
mod_separate_compilation_23.mod
% ls *.smod
zsh: no matches found: *.smod
```

While, if we add prefix `module` into `subroutine map_open_entry_sc_23()` and then it generates `mod_separate_compilation_23.smod` along with `mod_separate_compilation_23.mod`

```console
% gfortran -c b.f90
% ls *.mod
mod_separate_compilation_23.mod
% ls *.smod
mod_separate_compilation_23.smod
```

Now moving ahead to `c.f90` which contains submodule:

> c.f90

```fortran
submodule(mod_separate_compilation_23) submod_separate_compilation_23
    implicit none
contains

    module subroutine map_open_entry_sc_23()
    end subroutine map_open_entry_sc_23

end submodule submod_separate_compilation_23
```

on compiling `c.f90` we get, which shows generates a `.smod ` file whose name is prefixed with parent module name i.e. `mod_separate_compilation_23@submod_separate_compilation_23.smod`

```console
% gfortran -c c.f90 
% ls *.mod
mod_separate_compilation_23.mod
% ls *.smod
mod_separate_compilation_23.smod
mod_separate_compilation_23@submod_separate_compilation_23.smod
```


CMake behaves same with both GFortran and LFortran. I think GFortran when loading modules looks if `<name>.smod` exists or not? If it does then it loads all files matching with `<name>@<xx>.smod`. I think we should do the same.

certik:
I see. It looks like that indeed we can tell if a submodule will be needed. If so, then let's mark this in ASR for the parent module. I think last time we discussed it, we assumed that there is no way to tell. But it looks like there is a way to tell, using the `module` keyword. When we process the module, if we discover the `module` keyword (and/or other hints), we simply set the "has_submodule=True".

Then when the parent module is loaded in the main program, if it has `has_submodule=True`, then we need to start looking for all submodules, that we can load using the parent name. We either save submodules into `.smod`, or just use `.mod` and use the parent module mangling prefix, to load them.

So:

* For monolithic compilation, we see `has_submodule=True`, and we load all submodules automatically from their `.mod` files. We can consider using `.smod` for those. Specifically, when the main program is compiling, and any module has `has_submodule=True`, at this point we load the submodule. We do not load it right away, only when compiling the main program.
* For separate compilation I think we don't worry, we simply compile each parent / submodule into their `.mod` / `.smod` files and the user links all object files for us.

Issue #8084, opened 2025-07-17
Incorrect decleration of ArrayReshape node


sriganeshres:
MRE:
```fortran
program arrays_op_8
   implicit none
   integer(8) :: e1d(500), shape3d(3)
   print *, reshape(e1d, shape3d)
end program
````

**Comment by sriganeshres** [comment]
```
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2653
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 1185
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 394
    if (res.ok) {
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 12320
    v.visit_asr((ASR::asr_t&)asr);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 345
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 320
    case asrType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 346
    void visit_unit(const unit_t &b) { visit_unit_t(b, self()); }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 22
    case unitType::TranslationUnit: { v.visit_TranslationUnit((const TranslationUnit_t &)x); return; }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 1030
    visit_symbol(*item.second);
[...]

PR #8087, opened 2025-07-17
chore: gep to gep2

sriganeshres:
*  in visit_ArrayPhysicalCastUtil

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -6528,7 +6528,7 @@ public:
                 !ASR::is_a<ASR::ArrayConstant_t>(*ASRUtils::expr_value(m_arg))) ||
                 ASRUtils::expr_value(m_arg) == nullptr) &&
                 !ASR::is_a<ASR::ArrayConstructor_t>(*m_arg) ) {
-                tmp = llvm_utils->create_gep(tmp, 0);
+                tmp = llvm_utils->create_gep2(arr_type, tmp, 0);
             }
         } else if (
             m_new == ASR::array_physical_typeType::SIMDArray &&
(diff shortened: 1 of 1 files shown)


sriganeshres:
The Rest create_gep except for `visit_Assignment` are not being used in integration_tests.

sriganeshres:
And we have a issue for replacing in `visit_Assignment`. Issue: #8084.

tests/reference/llvm-call_subroutine_without_type_01-1c100d1.stdout
```diff
@@ -113,22 +113,23 @@ define i32 @main(i32 %0, i8** %1) {
   %16 = load i64, i64* %15, align 4
   %17 = getelementptr %mytype_polymorphic, %mytype_polymorphic* %obj, i32 0, i32 1
   %18 = load %mytype*, %mytype** %17, align 8
-  %19 = getelementptr %__vtab_mytype, %__vtab_mytype* %2, i32 0, i32 0
-  %20 = load i64, i64* %19, align 4
-  %21 = icmp eq i64 %16, %20
-  br i1 %21, label %then, label %else
+  %19 = bitcast %__vtab_mytype* %2 to <{ i64 }>*
+  %20 = getelementptr <{ i64 }>, <{ i64 }>* %19, i32 0, i32 0
```
certik, line 117: Why is:
```
%19 = getelementptr %__vtab_mytype, %__vtab_mytype* %2, i32 0, i32 0
```
replaced by:
```
  %19 = bitcast %__vtab_mytype* %2 to <{ i64 }>*
  %20 = getelementptr <{ i64 }>, <{ i64 }>* %19, i32 0, i32 0
```

I want to understand why it is needed and if there is any performance issue.

certik:
Otherwise good, if the CI passes.

tests/reference/llvm-call_subroutine_without_type_01-1c100d1.stdout
sriganeshres: No if the struct from symbol is empty in vtab we have a integer. So in this case we encounter a type of this vtab_base = <{i64}>. I didnot know how to get it other than symbol but symbol gives <{}>. So i made a type on the fly and finished it. No performace issues as of now.

certik: Ok, it looks like both will generate equivalent assembly. So I think it's fine for now.

certik: There might be some difference in alignment, thus in optimization opportunities (https://grok.com/share/bGVnYWN5_c00335b6-88a6-4978-b0d2-c3dc5f0d40b5), but I think that's not important right now.

sriganeshres:
@certik @Pranavchiku Can you merge this.

Issue #7839, opened 2025-06-21
Invalid casting of BOZ Literals


YashNagda17:
Consider the following program

```
program a
  IMPLICIT NONE
  
  integer :: my_int = Z'01'
  real, parameter :: result = 1 * Z'01'
  print *,1
end program
```

Compiling on gfortran 

```
$gfortran a.f90
a.f90:4:21:

    4 |   integer :: my_int = Z'01'
      |                     1
Error: BOZ literal constant at (1) is neither a data-stmt-constant nor an actual argument to INT, REAL, DBLE, or CMPLX intrinsic function [see '-fallow-invalid-boz']
a.f90:5:29:

    5 |   real, parameter :: result = 1 * Z'01'
      |                             1
Error: Operands of binary numeric operator '*' at (1) are INTEGER(4)/BOZ
```
Whereas program on lfortran runs without any errors

```
$ lfortran a.f90
1
```

The cause of the issue is likely because the compiler should ideally know whether to treat the bits of Z'01' like BOZ Values as INT, REAL etc. There should be an explicit conversion as, `INT(Z'01')` before using it in an operation with Base 10 literals declared.

Kindly correct me if I am wrong. Can I take up this issue?

harperjf:
This part of Fortran has changed significantly since F2018. The  F2023 standard 7.7 C7119 (R772) says: 
"A _boz-literal-constant_ shall appear only as a _data-stmt-constant_ in a DATA statement, as the
_initialization_ for a named constant or variable of type integer or real, as the _expr_ in an intrinsic assignment
whose _variable_ is of type integer or real, as an _ac-value_ in an array constructor with a _type-spec_ that
specifies type integer or real, as the _scalar-expr_ in an enum constructor, or where explicitly allowed in
16.9 as an actual argument of an intrinsic procedure.10.2.1.2 ".

F2018 7.7 C7019 (R764) allowed only the first and last of the F2023 possibilities.

Gfortran objected to line 4 of the test program, which is bad F2018 but good F2023. Line 5  is bad F2018 and also bad F2023. On looking at the Gfortran website just now I found that it claims F2018 but not F2023 compliance. Is it worth reporting a gfortran bug? It would certainly be useful to see what happens if the test program is compiled with the  ```-fallow-invalid-boz``` option, preferably if ```my_int``` and ```result``` were printed instead of just ```1```.

YashNagda17:
Thanks for the details analysis... 

[...]

PR #8095, opened 2025-07-19
Adding BOZ Support in Data Literals

YashNagda17:
Towards #7839

Building on #8035 , the BOZ Integration into Data Statements is complete. Support for Real & Integer BOZ Literals in implied-do statements & Array Assignments, inside the Data Statements, is complete (Scalar Support was already addressed in #8035).

This shall make usage of BOZ Compliant with first condition of Fortran 2023 as follows:

_A boz-literal-constant shall appear only as a data-stmt-constant in a DATA statement._


YashNagda17:
As per #7839 , gfortran hasn't integrated the compliance towards Fortran 2023 standard, which this request is trying to achieve. Kindly suggest how to correctly proceed further, since the check with gfortran is not a valid comparison.

integration_tests/CMakeLists.txt
```diff
@@ -1895,7 +1895,7 @@ RUN(NAME bits_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME cpu_time_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 # RUN(NAME cpu_time_02_wasm LABELS wasm)
 RUN(NAME system_clock_01 LABELS gfortran llvm)
-RUN(NAME boz_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME boz_01 LABELS llvm llvm_wasm llvm_wasm_emcc) #remove gfortran label, as functionality is not supported by gfortran
```
certik: Is this change correct Fortran? Either way, keep the original test, since it works with GFortran and add a new test for this.

certik:
@YashNagda17 keep the original test since it works with GFortran, and add this as a new test that only works with LFortran.

YashNagda17:
Ohk, working to keep test-case files separate. 

Meanwhile, addressing the question of correctness of the code, here I am attaching the outputs of the real test case, with the modified lfortran and AMD Fortran Compiler for reference. AMD flang has already integrated these changes, so considering it as reference. 

```
program a
implicit none

integer :: i
real :: boz_1, matrix1(5), matrix2(2), arr1(2)
integer :: arr2(2)

!Check Scalar Assigment
data boz_1 /b'01011101'/  

!Check Multiple Assigments of Real array with BOZ Values
data matrix1 / b'10', 2*o'100', 3.12, z'ABC' / 

!Check Data Broadcasting to all elements of array
data matrix2 / 2*b'10' /

!Check Implied Do loop, with real assigments along with integer assignments
data (arr1(i), arr2(i), i=1,2) / 1.2, z'02', z'01', 3 /

print *, boz_1
print *, matrix1
print *, matrix2
print *, arr1, arr2

end program
```

Lfortran Output:
```
$ lfortran a.f90
1.30320757e-43
2.80259693e-45    8.96831017e-44    8.96831017e-44    3.11999989e+00    3.85076818e-42
2.80259693e-45    2.80259693e-45
1.20000005e+00    1.40129846e-45    2    3
```

Flang Output
```
$ flang a.f90 & ./a.out
[1] 124042
1.30320757e-43
2.80259693e-45    8.96831017e-44    8.96831017e-44    3.11999989e+00    3.85076818e-42
2.80259693e-45    2.80259693e-45
1.20000005e+00    1.40129846e-45    2    3
```

certik:
Excellent! Yes, since GFortran can't compile it yet, if you could please separate the test, that way we keep using GFortran for what it can handle, and the new functionality will be tested just with LFortran, without GFortran.

integration_tests/boz_01.f90
```diff
@@ -2,23 +2,10 @@ program boz_01
...
-data boz_4 /b'01011101'/
-data matrix1/ b'10', 2*o'100', 3.12, z'ABC' /
-data matrix2/ 5*b'10'/
-data (arr1(i), arr2(i), i=1,2) / 1.2, z'02', z'01', 3 /
 
-print *, boz_1, boz_2, boz_3, boz_4
-print *, matrix1
-print *, matrix2
-print *, arr1,arr2
+print *, boz_1, boz_2, boz_3
```
certik, line 10: Please keep the original file without modifications. It works with both GFortran and LFortran, so it should keep working.

certik: Ah I see --- it didn't compile with GFortran in the first place?

integration_tests/boz_02.f90
```diff
@@ -0,0 +1,25 @@
...
+!Check Data Broadcasting to all elements of array
+data matrix2 / 2*b'10' /
+
+!Check Implied Do loop Assignments, with real assigments along with integer assignments
+data (arr1(i), arr2(i), i=1,2) / 1.2, z'02', z'01', 3 /
+
+print *, boz_1
+print *, matrix1
+print *, matrix2
+print *, arr1, arr2
```
certik, line 23: Which of these cases didn't work in main, and now they do?

integration_tests/boz_01.f90
YashNagda17: Yes, I have kept only those cases in boz_01.f90 tests, which work in gfortran. 

Currently, gfortran only supports conversion from BOZ to INTEGER Variables, so I moved all tests with BOZ to REAL Conversions in boz_02.f90

integration_tests/boz_02.f90
YashNagda17: All of the cases. In PR #8035 , I added basic functionality for BOZ to Real Conversions, and tested on Scalar Assignments of BOZ inside Data Statements. Since gfortran doesn't allow this functionality yet, so I have added test-case for the same in the new test-case file, boz_02.f90. 

So, after this PR, any BOZ Assignment to REAL or INTEGER Variables inside the Data Statements can be assigned correctly as per Fortran 2023 Standard. Integer support was already present in main, Real Support is now added via #8035 and this PR. 

Sorry, I should have mentioned this in detail earlier.

PR #8100, opened 2025-07-20
fix: fix printing for allocatable scalars

### LINKED ISSUE #8071: Wrong value of allocatable scalar integer
This program was inspired by the #8060 issue. There was no segfault but the print statament prints a different wrong value each time the program runs with lfortran.
```
  integer,allocatable:: x
  x = 666
  print "(I0)", x
  deallocate(x)
end program
```
Output from 3 other compilers:
```
666
```
Output from lfortran 0.54.0
```
1152250592
```
Output from a second run
```
2115191520
```

HarshilShah1804:
Fixes #8071 

This bug was probably introduced as a regression from #7917


certik:
@HarshilShah1804 great job, thanks for fixing it!

Issue #8098, opened 2025-07-19
No Test Coverage for All functions


sriganeshres:
* [ ] Function `fill_array_details`
* [ ] Function `nested_struct_rd`
* [ ] Function `lfortran_intrinsic`
* [x] Function `call_lcompilers_free_strings` has no body  
* [ ] Function `visit_IntegerBitLen`  
* [ ] Function `visit_ListCompare`  
* [ ] Function `generate_ListIndex`  
* [ ] Function `generate_Exp2`  
* [ ] Funcion `generate_Expm1`  
* [ ] Function `generate_ListPop_0`  
* [ ] Function `generate_ListReserve`  
* [ ] Function `generate_DictElems`  
* [ ] Function `generate_SetRemove`  
* [ ] Function `visit_ListRepeat`  
* [ ] Function `visit_TupleLen`  
* [ ] Function `visit_ArraySection`  
* [ ] Function `lookup_EnumValue`  
* [ ] Function `visit_EnumValue`  
* [ ] Function `visit_EnumName`  
* [ ] Function `visit_EnumConstructor`  
* [ ] Function `visit_SizeOfType`  
* [ ] Function `visit_StructInstanceMember`  
* [ ] Function `visit_Enum`  
* [ ] Function `is_nested_pointer` - @certik @czgdp1807 if the issue is fixed this can be removed.  
* [ ] Function `visit_CLoc`  
* [ ] Function `visit_OverloadedStringConcat`  
* [ ] Function `visit_UnsignedIntegerCompare`  
* [ ] Function `visit_CPtrCompare`  
* [ ] Function `visit_If`  
* [ ] Function `visit_StringRepeat`  
* [ ] Function `visit_StringOrd`  
* [ ] Function `visit_OverloadedUnaryMinus`  
* [ ] Function `visit_UnsignedIntegerBitNot`  
* [ ] Function `generate_flip_sign`  
* [ ] Function `generate_fma`  
* [ ] Function `generate_sign_from_value`  
* [ ] Function `generate_optimization_instructions`  
* [x] Function `handle_bitwise_args`  
* [x] Function `handle_bitwise_xor`  
* [x] Function `handle_bitwise_and`  
* [x] Function `handle_bitwise_or`

sriganeshres:
I have added all those functions which are not used in `asr_to_llvm.cpp` file during integration tests.
cc: @certik

Pranavchiku:
[...]

PR #8111, opened 2025-07-22
removed call_lcompilers_free_strings

sriganeshres:
towards: #8098


certik:
@assem2002 should we remove this, or keep it since we will fix this after the string refactor is done?

assem2002 (approved):

assem2002:
It's not needed. I recommended to remove it.

Issue #8109, opened 2025-07-21
bug: lfortran prints junk value for allocatable variable


HarshitaKalani:
```fortran
program main
implicit none
integer(4), allocatable :: x
allocate(x)
x = 5
print *, x
end program main
```
```console
$ lfortran a.f90
-813944096
$ gfortran a.f90 && ./a.out
           5
```

kmr-srbh:
This is working correctly for me on Mac:

```console
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
5
```

HarshilShah1804:
On Linux also, this works correctly. 
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
5
```

This issue was fixed in #8100 . You can try pulling the latest changes from the main.

harperjf:
It would help us all when various lfortran versions give different outputs, as they seem to with this issue, if everyone said which version they were using. Also how does one get the "main" ? I have been getting the 0.n.0 version when a new value of n appears in the list of tags. At any given time is the main somewhere between version 0.n.0 and 0.(n+1).0 ? Sorry to be asking very elementary questions: I am not an expert in compiler-writing, but I have reported many bugs in Fortran compilers over the years.

assem2002:
@harperjf You can check the instructions here https://docs.lfortran.org/en/installation/#build-from-git. They're straight forward initializing a conda env, then building the project right away.

HarshitaKalani:
Oh right. This now works. However, I came across this while solving the following bug:
```fortran
program main
implicit none
integer(4), allocatable :: x
allocate(x, source = 5)
print *, x
end program main
```
The above program used to give a junk value while printing x. Now, it gives 0. I've the solution, I'll raise a PR for the same. Thanks!

PR #8114, opened 2025-07-22
fix: handle assignment when allocate is provided with source keyword

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8109


certik:
I think that this is fine. I would document that now the "source" in Allocate will *not* do a copy, it will only use it for a runtime decision how to allocate.

HarshitaKalani:
Sorry, I did not understand. Here I'm copying the contents of source into the variable being allocated.

certik:

Pranavchiku (approved):

PR #8117, opened 2025-07-22
fix: handle array args in convert to polymorphic

jinangshah21:
Uncomments `class_58` integration test

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -2573,12 +2573,11 @@ public:
             v = ASRUtils::EXPR2VAR(x.m_v);
             uint32_t v_h = get_hash((ASR::asr_t*)v);
             array = llvm_symtab[v_h];
-            // TODO after strucType Refactor: uncomment this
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2573,12 +2573,11 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             v = ASRUtils::EXPR2VAR(x.m_v);
             uint32_t v_h = get_hash((ASR::asr_t*)v);
             array = llvm_symtab[v_h];
-            // TODO after strucType Refactor: uncomment this
-            // if (ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(v->m_type))) {
-            //     ASR::Struct_t* der_symbol = ASR::down_cast<ASR::Struct_t>(
-            //         ASRUtils::symbol_get_past_external(v->m_type_declaration));
-            //     current_der_type_name = std::string(der_symbol->m_name);
-            // }
```
kmr-srbh, line 2581: Ah! You already had a variant for the new change. We missed it.

Issue #7897, opened 2025-06-26
Linker error for module functions defined inside interface block


HarshilShah1804:
MRE
```fortran
module julienne_command_line_m
  implicit none
  private
  public :: command_line_t
  type command_line_t
  contains
    procedure, nopass :: flag_value
  end type

  interface
    module function flag_value(flag)
      implicit none
      character(len=*), intent(in) :: flag
      character(len=:), allocatable :: flag_value
    end function

  end interface

contains

  module procedure flag_value
    flag_value=""
  end procedure

end module

program get_command_line_flag_value
  use julienne_command_line_m, only : command_line_t
  implicit none
  type(command_line_t) command_line
  print *, command_line%flag_value("--foo")
end program
```

```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
/usr/bin/ld: /tmp/expr2.tmp.o: in function `main':
LFortran:(.text+0x80): undefined reference to `flag_value'
/usr/bin/ld: LFortran:(.text+0xbc): undefined reference to `flag_value'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
The command 'clang -o expr2.out /tmp/expr2.tmp.o  -L"/home/harshil-shah/lfortran/src/bin/../runtime" -Wl,-rpath,"/home/harshil-shah/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
```

MRE works without any errors with GFortran.

Pranavchiku:
You'll have to implement `visit_Procedure` written in `ast_common_visitor.h`. For this pickup function signature from interface and body from module procedure and then register a Function with `abiType=Implementation`.

HarshilShah1804:
This MRE now works with LFortran. So, I am closing this issue.

Issue #7890, opened 2025-06-25
CodeGenError for module function inside interface block


HarshilShah1804:
MRE

```fortran
module julienne_command_line_m
  implicit none

  private
  public :: command_line_t

  type command_line_t
  contains
    procedure, nopass :: argument_present
  end type

  interface

    module function argument_present(acceptable_argument) result(found)
      implicit none
      character(len=*), intent(in) :: acceptable_argument(:)
      logical found
    end function

  end interface

contains

  module procedure argument_present
      found = .false.
  end procedure

end module

program check_for_command_line_argument
  use julienne_command_line_m, only : command_line_t
  implicit none

  type(command_line_t) command_line

  if (command_line%argument_present(["--some-argument"])) then
     print '(a)', new_line('') // "argument 'some-argument' present" // new_line('')
  else
     print '(a)', new_line('') // "argument 'some-argument' not present" // new_line('')
  end if
end program
```

LFortran
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran examples/expr2.f90 
code generation error: Function code not generated for 'argument_present'


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

MRE works without any errors with GFortran

Pranavchiku:
Quick workaround:

```fortran
module julienne_command_line_m
  implicit none

  private
  public :: command_line_t

  type command_line_t
  contains
    procedure, nopass :: argument_present
  end type

  interface

    module function argument_present(acceptable_argument) result(found)
      implicit none
      character(len=*), intent(in) :: acceptable_argument(:)
      logical found
    end function

  end interface

contains

    module function argument_present(acceptable_argument) result(found)
      implicit none
      character(len=*), intent(in) :: acceptable_argument(:)
      logical found
      found = .false.
  end

end module

program check_for_command_line_argument
  use julienne_command_line_m, only : command_line_t
  implicit none

  type(command_line_t) command_line

  if (command_line%argument_present(["--some-argument"])) then
     print '(a)', new_line('') // "argument 'some-argument' present" // new_line('')
  else
     print '(a)', new_line('') // "argument 'some-argument' not present" // new_line('')
  end if
end program
```

HarshitaKalani:
```console
% lfortran a.f90

argument 'some-argument' not present

% gfortran a.f90 && ./a.out

argument 'some-argument' not present

[...]

PR #8126, opened 2025-07-24
feat : Implement Module Procedures for Submodules by creating a new ASR Function

AnshMehta1:
Towards #536 
Addresses https://github.com/lfortran/lfortran/pull/7961#issuecomment-3036569212

Features of PR :-
1. For module procedure in submodules (representing Implementation for some interface), we create an ASR Function in AST SymbolTable Visitor using the Function Signature and args of interface from the parent module.
2. We also use symbol and expr duplicators for updating scopes of args.
3. Then, in AST Body Visitor, we handle body of module procedure and populate our Function (representing module procedure) with ASR body Statements.

Currently, this PR is also able to handle Module Procedures in a Module (#7890 , #7897), but I think we should handle module procedures in a module separately (directly update Parent Function instead of creating new Function). Will handle that in next PR.

Also, currently, declaration made in module procedures are not handled like :- 
```.f90
implicit none

contains

    module procedure error_stop
        integer, parameter :: i = 3 --------------------> not handled yet
        code = i
    end procedure

end submodule f18estop
```

Will handle this in my next pr.


Pranavchiku (commented):
I'll review this in a while.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -819,6 +819,60 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        SymbolTable* parent_scope = current_scope;
+        current_scope = al.make_new<SymbolTable>(parent_scope);
+
+        ASRUtils::SymbolDuplicator symbol_duplicator(al);
+        ASRUtils::ExprStmtWithScopeDuplicator exprstmt_duplicator(al, current_scope);
+
+        for (auto &item : proc_interface->m_symtab->get_scope()) {
+            if (ASR::is_a<ASR::Variable_t>(*item.second)) {
+                ASR::Variable_t* proc_interface_vari = ASR::down_cast<ASR::Variable_t>(item.second);
+                ASR::symbol_t* new_sym = symbol_duplicator.duplicate_Variable(proc_interface_vari, current_scope);
```
Pranavchiku, line 845: check there is a method to directly duplicate symtab

AnshMehta1: Ok, will check for it and make changes in next pr.

PR #8132, opened 2025-07-24
Fix for LP list regression

swamishiju:


certik:
The CI failed with:
```
1746/1746 Test #1736: lp_list_test_02__LLVM_GOC ...................................***Exception: SegFault  0.10 sec
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5036,9 +5036,17 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                    is_argument))); 
+                    is_argument);
+
+            // Implicitly consider deferred length descriptor string as allocatable
+            if (ASRUtils::is_deferredLength_string(type) && ASRUtils::is_descriptorString(type))
+                type = determine_type(loc, sym, sym_type->m_attr,
+                                is_pointer, true, dims, var_sym, type_declaration, abi,
+                                is_argument);
+
+            return ASRUtils::TYPE(ASR::make_List_t(al, loc, type));
```
assem2002, line 5140: I think this is fine. @certik Do you have worries about this?

src/libasr/asr_utils.h
```diff
@@ -3329,6 +3329,18 @@ static inline bool is_deferredLength_string(ASR::ttype_t* t){
     return false;
 }
 
+static inline bool is_allocated_descriptor_string(ASR::ttype_t* t) {
```
assem2002: Shouldn't that be named `is_allocatable_descriptor_string` (according to the functionality of it)

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2374,52 +2505,11 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
-                llvm::Value* l = LLVMUtils::CreateLoad2(llvm::Type::getInt8Ty(context), create_ptr_gep2(
-                    llvm::Type::getInt8Ty(context), left, i));
-                llvm::Value* r = LLVMUtils::CreateLoad2(llvm::Type::getInt8Ty(context), create_ptr_gep2(
-                    llvm::Type::getInt8Ty(context), right, i));
-                return builder->CreateICmpEQ(l, r);
+                ASR::string_physical_typeType string_type = ASR::down_cast<ASR::String_t>(asr_type)->m_physical_type;
+                if (string_type == ASR::string_physical_typeType::CChar)
+                    return is_equal_pointer_string(left, right);
+                else 
+                    return is_equal_descriptor_string(left, right);
```
assem2002: Are these equivalent to `stringcompare` and using the runtime C function `str_compare`?

src/lfortran/semantics/ast_common_visitor.h
certik: I think this is related to the fact that we can't easily specify that the character should be allocatable in Fortran syntax, so for now we just make it. But only for lists and sets. Correct @swamishiju ?

swamishiju: Yes, in LP we can do this in annotation directly but in LF we have to implicitly convert

src/libasr/asr_utils.h
swamishiju: Oh yes its type 😅  I'll refactor

src/libasr/codegen/llvm_utils.cpp
swamishiju: I'm not entirely sure, I think these are only used in LP types but the C equivalent would be just looping over corresponding i8* - the only difference is probably the final `i1` value.

swamishiju:
The test is working for me locally on a Mac (`./run_tests -b llvm -sc`), im assuming CI runs on Linux can someone check

assem2002:
I can reproduce that locally.
Here is valgrind output
```console 
(lf) assem@assem-PC:~/lfortran_test/lfortran/integration_tests$ lfortran lp_list_test_02_.f90 -o pp
(lf) assem@assem-PC:~/lfortran_test/lfortran/integration_tests$ valgrind ./pp 
==182271== Memcheck, a memory error detector
==182271== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==182271== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==182271== Command: ./pp
==182271== 
==182271== Conditional jump or move depends on uninitialised value(s)
==182271==    at 0x484DC80: realloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==182271==    by 0x486CFAB: _lfortran_strcpy (lfortran_intrinsics.c:2783)
==182271==    by 0x40209B: __module_list_test_02_mod_test_list_01 (in /home/assem/lfortran_test/lfortran/integration_tests/pp)
==182271==    by 0x402F45: __module_list_test_02_mod_tests (in /home/assem/lfortran_test/lfortran/integration_tests/pp)
==182271==    by 0x402F5A: main (in /home/assem/lfortran_test/lfortran/integration_tests/pp)
```
sometimes `realloc` function raises an error `invalid pointer`
If you need some help, please ask. I left it for now as you're more familiar with the implementation of list methods, so you'll probably catch the segfault faster.

swamishiju:
@assem2002 Could you take help with this PR if possible? I'm facing a couple different bugs even after commenting stuff out and I'm a bit lost as to why it doesn't work. 
Also it not giving any error on my machine is really annoying.

assem2002:
@swamishiju? Were you able to resolve the issue with test `lp_list_test_02_`? If not, I can start off from there.

swamishiju:
> @swamishiju? Were you able to resolve the issue with test `lp_list_test_02_`? If not, I can start off from there.

Nope I currently just commented it out - If i remember correctly the issue arises in `_lfortran_list_set_item`

assem2002:
Good. I'll keep investigating.

swamishiju:
This looks great I think only wasm tests fail now. @assem2002 could you give me a brief of the fix if possible? From what i understand it was some issue `set_string_on_heap` not setting length.

assem2002:
yeah that was partly the problem, but with `allocatable-deferred-length` strings, you just need to set length to 0 and the ptr to `nullptr` (ptr location to `0`).
That's why I was making sure to set any allocated memory for `list` with ZEROES, that sets up the string properly for futher usages (typically using C runtime `lfortran_strpcy`).

For the current error, I don't actually know the bug yet.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2429,85 +2430,105 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
-        llvm::Value* left_string_size = CreateLoad2(llvm::Type::getInt64Ty(context),
-                                           create_gep2(string_descriptor, left, 1));
-        llvm::Value* right_string_body = CreateLoad2(llvm::Type::getInt8Ty(context)->getPointerTo(),
-                                           create_gep2(string_descriptor, right, 0));
-        llvm::Value* right_string_size = CreateLoad2(llvm::Type::getInt64Ty(context),
-                                           create_gep2(string_descriptor, right, 1));
+        llvm::Value* l_str_data, *l_str_len;
+        llvm::Value* r_str_data, *r_str_len;
+        std::tie(l_str_data, l_str_len) = get_string_length_data(type, left);
+        std::tie(r_str_data, r_str_len) = get_string_length_data(type, right);
```
swamishiju, line 2452: These look significantly better 😅 
I wrote those before getting familiar with the API

assem2002: Yeah. I thought the same. 
I tried to modify the existing implementation as it was good, but I eventually needed to change the logic of how we check the strings just to avoid some subtle issues like accessing wrong memory when `i` gets larger than length, or to not access anything if the length is `0`. They were subtle things, but it made the tests fail.

PR #8128, opened 2025-07-24
fix: custom operator with the same name as another variable

akramhany:
Fixes #7884

Also, right now I am using (`~~`) as a symbol to distinct custom operators from other symbols, maybe we need to use something else.

PR #8137, opened 2025-07-25
refactor: improve handling of custom operator names

akramhany:
Refactored the updating of custom operator names in PR #8128

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -10539,8 +10539,8 @@ public:
         std::string new_op = op;
 
         // Append "~~" to the begining of any custom defined operator, that is not inside a struct
-        if (should_change_custom_op_name(op) && first_struct == nullptr) {
-            new_op = "~~" + op;
+        if (first_struct == nullptr) {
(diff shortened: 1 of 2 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -11583,15 +11583,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         }
     }
 
-    // check if we should append "~~" to an operator name
-    bool should_change_custom_op_name(const std::string &op) {
-        static const std::string operator_chars = "+-*/~<=>";
-
-        if (op.find_first_of(operator_chars) != std::string::npos) {
-            return false;
```
akramhany, line 11591: At first, I thought overloaded operators like "+" ("-", etc..), where actually used as they are in the AST and ASR.
After looking again, I found that we replace them with a function like `add` for `+`, and append this function name with `~`, to distinct it from other symbols.
I don't think it is necessary to change these operators (`~add`, `~sub`, etc), as they are already distinct, so now I only check if the operator name contains a `~`, and return it if it does, or update it with `~~` if it doesn't.

PR #8142, opened 2025-07-26
enh : Support Declarations made in Module Procedures

AnshMehta1:
Addresses https://github.com/lfortran/lfortran/pull/8126#discussion_r2231316807
PR also adds support for declarations made in module procedures. For eg :- 
```.f90
implicit none

contains

    module procedure error_stop
        integer, parameter :: i = 3 
        code = i
    end procedure

end submodule f18estop
```

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -838,15 +838,7 @@ public:
 
         ASRUtils::SymbolDuplicator symbol_duplicator(al);
         ASRUtils::ExprStmtWithScopeDuplicator exprstmt_duplicator(al, current_scope);
-
-        for (auto &item : proc_interface->m_symtab->get_scope()) {
-            if (ASR::is_a<ASR::Variable_t>(*item.second)) {
-                ASR::Variable_t* proc_interface_vari = ASR::down_cast<ASR::Variable_t>(item.second);
-                ASR::symbol_t* new_sym = symbol_duplicator.duplicate_Variable(proc_interface_vari, current_scope);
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -838,22 +838,26 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            if (!AST::is_a<AST::Require_t>(*x.m_decl[i])) {
+                try {
+                    visit_unit_decl2(*x.m_decl[i]);
+                } catch (SemanticAbort &e) {
+                    if ( !compiler_options.continue_compilation ) throw e;
+                }
+            }
+            is_Function = false;
+        }
+
```
Pranavchiku, line 860: you may wish to preserve last state of `is_Function`

```suggestion
        bool is_Function_copy = is_Function;
        is_Function = true;
        for (size_t i=0; i<x.n_decl; i++) {

            if (!AST::is_a<AST::Require_t>(*x.m_decl[i])) {
                try {
                    visit_unit_decl2(*x.m_decl[i]);
                } catch (SemanticAbort &e) {
                    if ( !compiler_options.continue_compilation ) throw e;
                }
            }

        }
        is_Function = is_Function_copy;
```

AnshMehta1: I don't think `is_Function` is required anywhere else apart from DeclarationUtil. So, I don't think that we need to preserve previous state. For `visit_Subroutine` and `visit_Function` also, we do the similar thing instead of creating a copy.

Pranavchiku: okay, still please bring the `is_Function = true` outside loop.

PR #8147, opened 2025-07-26
refactor: pass `Allocator` instance through call chain for expression node creation

kmr-srbh:


src/libasr/codegen/c_utils.h
```diff
@@ -848,15 +848,14 @@ class CCPPDSUtils {
                                 + struct_type_str + "* dest)";
             func_decls += "inline " + signature + ";\n";
             std::string tmp_generated = indent + signature + " {\n";
-            Allocator al(4*1024);
             for(size_t i=0; i < struct_t->n_members; i++) {
                 std::string mem_name = std::string(struct_t->m_members[i]);
                 ASR::symbol_t* member = struct_t->m_symtab->get_symbol(mem_name);
                 ASR::expr_t* member_expr = ASRUtils::EXPR(ASR::make_Var_t(al, member->base.loc, member));
```
certik, line 854: So all this PR is needed for this one line, correct? It uses `al`, so we need to get it to it.

certik: What exactly is this function doing? I can see a lot of string manipulation. Do we create `member_expr` just to generate some C code? That seems wasteful ---- is there another way of doing it?

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -241,16 +242,15 @@ namespace LCompilers {
...
                 member_types.push_back(par_llvm);
                 dertype2parent[der_type_name] = std::string(par_der_type->m_name);
                 member_idx += 1;
             }
-            Allocator al(1024);
             for( size_t i = 0; i < der_type->n_members; i++ ) {
                 std::string member_name = der_type->m_members[i];
                 ASR::Variable_t* member = ASR::down_cast<ASR::Variable_t>(der_type->m_symtab->get_symbol(member_name));
-                llvm::Type* llvm_mem_type = get_type_from_ttype_t_util(ASRUtils::EXPR(ASR::make_Var_t(
+                llvm::Type* llvm_mem_type = get_type_from_ttype_t_util(al, ASRUtils::EXPR(ASR::make_Var_t(
```
certik, line 253: Here is another reason -- we create Var just to get the LLVM type. Is there another way of doing it?

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2728,7 +2727,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
                                 llvm::ConstantInt::get(llvm::Type::getInt64Ty(context), llvm::APInt(64, 0)));
                         }
                         create_if_else(is_allocated, [&]() {
-                            deepcopy(ASRUtils::EXPR(ASR::make_Var_t(al, mem_sym->base.loc, mem_sym)), src_member, dest_member,
+                            deepcopy(al, ASRUtils::EXPR(ASR::make_Var_t(al, mem_sym->base.loc, mem_sym)), src_member, dest_member,
```
certik, line 2730: I think this Var is the last one. Is there another way of doing it?

certik:
Can you check that this PR fixes the memory issue? This is a huge change, so let's make sure it actually fixes the problem.

It seems in all these cases the ASR that is created is short-lived --- the lifetime is for the duration of the function it exists in. I think we should use some kind of a temporary Arena for this particular case. The way to do it is to create a local array as a variable, so it lives on the stack of the function. Then tell the Arena to live in this array. Then create Var. When the function finishes, the Var will get destroyed automatically.

A related question is why we need to create these Var nodes in the first place --- why is it needed and is there no other way?

Issue #8148, opened 2025-07-27
ICE: `present()` on optional argument inside internal subroutine causes crash


HarshitaKalani:
```fortran
module bspline_sub_module
    contains
    subroutine check_inputs(nx)
        implicit none
        integer(4),intent(in),optional :: nx
        contains
            subroutine check(n)
                implicit none
                integer(4),intent(in),optional :: n     
                print *, present(n)
            end subroutine check
    end subroutine check_inputs
end module bspline_sub_module

program main
    use bspline_sub_module
    implicit none
    integer(4) :: nx
    call check_inputs(nx)
end program
```
```console
$ lfortran a.f90 --cpp
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/harshita/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 392, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshita/Desktop/lfortran/src/bin/lfortran.cpp", line 2729, in main
    return main_app(argc, argv);
.
.
.
  File "/home/harshita/Desktop/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 2325, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplacePresentCalls>::replace_expr(LCompilers::ASR::expr_t*)
    self().replace_StringFormat(down_cast<ASR::StringFormat_t>(x));
  File "/home/harshita/Desktop/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 1006, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplacePresentCalls>::replace_StringFormat(LCompilers::ASR::StringFormat_t*)
    self().replace_expr(x->m_args[i]);
  File "/home/harshita/Desktop/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 2105, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplacePresentCalls>::replace_expr(LCompilers::ASR::expr_t*)
    self().replace_IntrinsicElementalFunction(down_cast<ASR::IntrinsicElementalFunction_t>(x));
[...]

PR #8150, opened 2025-07-27
enh: handle optional args in nested subroutines

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8148

### DIFF
--- a/integration_tests/intrinsics_392.f90
+++ b/integration_tests/intrinsics_392.f90
@@ -1,0 +1,21 @@
+module bspline_sub_module
+    contains
+    subroutine check_inputs(nx)
+        implicit none
+        integer(4),intent(in),optional :: nx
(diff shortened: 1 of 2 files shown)


Pranavchiku (approved):

integration_tests/intrinsics_392.f90
```diff
@@ -0,0 +1,21 @@
...
+    contains
+    subroutine check_inputs(nx)
+        implicit none
+        integer(4),intent(in),optional :: nx
+        call check(nx)
+        contains
+            subroutine check(n)
+                implicit none
+                integer(4),intent(in),optional :: n     
+                print *, present(n)
```
certik, line 11: It would be good to add some "if () error stop" style checks here to ensure it is returning the correct result.

PR #8152, opened 2025-07-27
refactor: removed handle_bitwise_args, handle_bitwise_or, handle_bitwise_and, handle_bitwise_xor

sriganeshres:
towards: #8098

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -12083,18 +12083,6 @@ public:
                 throw CodeGenError("FunctionCall: StructType symbol type not supported");
             }
         }
-        if( ASRUtils::is_intrinsic_function2(s) ) {
-            std::string symbol_name = ASRUtils::symbol_name(x.m_name);
-            if( startswith(symbol_name, "_bitwise_xor") ) {
-                return ;
-            }
-            if( startswith(symbol_name, "_bitwise_and") ) {
-                return ;
-            }
-            if( startswith(symbol_name, "_bitwise_or") ) {
-                return ;
-            }
-        }
 
         bool intrinsic_function = ASRUtils::is_intrinsic_function2(s);
         uint32_t h;


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12089,15 +12060,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                handle_bitwise_xor(x);
                 return ;
             }
             if( startswith(symbol_name, "_bitwise_and") ) {
-                handle_bitwise_and(x);
                 return ;
             }
             if( startswith(symbol_name, "_bitwise_or") ) {
-                handle_bitwise_or(x);
                 return ;
```
certik: Are these not used in LPython either? How is the `_bitwise_and` created? We might need to add more tests into LFortran to exercise these. If it is not needed, can you find the PR that makes this code obsolete?

This PR still leaves their implementation, but now makes them a no-op. That doesn't seem like the right fix. Instead, we should remove more code.

sriganeshres: Yes it would be good to remove the block completety. I was told that they are being handled else where so it is not a problem to remove this.

sriganeshres: I will try to find where they are lowered.

sriganeshres: I guess we are lowering them directly in this
`void visit_LogicalBinOp(const ASR::LogicalBinOp_t &x)` function without going to `handle_bitwise_XX`

sriganeshres: I have removed that block as it it not being reached by control.

certik:
@swamishiju are these used in LPython, or can they be removed?

swamishiju:
> @swamishiju are these used in LPython, or can they be removed?

No its not used in LP specifically but a quick grep gave
```c++
void visit_FunctionCall : asr_to_llvm.cpp
...
        if( ASRUtils::is_intrinsic_function2(s) ) {
            std::string symbol_name = ASRUtils::symbol_name(x.m_name);
            if( startswith(symbol_name, "_bitwise_xor") ) {
                handle_bitwise_xor(x);
                return ;
            }
            if( startswith(symbol_name, "_bitwise_and") ) {
                handle_bitwise_and(x);
                return ;
            }
            if( startswith(symbol_name, "_bitwise_or") ) {
                handle_bitwise_or(x);
                return ;
            }
        }

```

this is the only place I can find it

certik:
Thanks. Then let's remove it.

sriganeshres:
@certik Tests have passed please merge this

Issue #8143, opened 2025-07-26
Flaky CI test for stdlib sorting


Pranavchiku:
For macos LLVM 20: https://github.com/lfortran/lfortran/actions/runs/16540721847/job/46781559168

```
99% tests passed, 1 tests failed out of 200

Label Time Summary:
quadruple_precision    =   0.00 sec*proc (2 tests)

Total Test time (real) =   3.48 sec

The following tests did not run:
	  1 - always_skip (Skipped)

The following tests FAILED:
	 20 - sorting (SEGFAULT)
```

This passes on rerun.

PR #8153, opened 2025-07-28
create `alloca` at the entry point when used to change function return type into pointer

assem2002:
- Make sure to allocate stack memory at entry point (for a temporary that isn't part of the ASR, hence could lead to stack overflow).
- Add an integration test
***
 closes https://github.com/lfortran/lfortran/issues/8143

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -12195,7 +12195,12 @@ public:
             // The convention we use is that any strings is a pointer to the underlying physicalType
             // Example of StringPhysicalTypes -> `string_descriptor*`, `i8*`
             if(ASRUtils::is_string_only(return_var_type0)){ 
-                llvm::Value* string_ptr = llvm_utils->CreateAlloca(*builder, llvm_utils->get_StringType(return_var_type0));
+                // Make sure to use `alloca` at the entry point.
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12195,7 +12195,12 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             // The convention we use is that any strings is a pointer to the underlying physicalType
             // Example of StringPhysicalTypes -> `string_descriptor*`, `i8*`
             if(ASRUtils::is_string_only(return_var_type0)){ 
-                llvm::Value* string_ptr = llvm_utils->CreateAlloca(*builder, llvm_utils->get_StringType(return_var_type0));
+                // Make sure to use `alloca` at the entry point.
+                llvm::Value* string_ptr = llvm_utils->CreateAlloca(
+                                            llvm_utils->get_StringType(return_var_type0),
+                                            nullptr,
+                                            "string_ret_const"
+                                        );
```
assem2002, line 12203: The API that operates on strings is always expecting a pointer to the underlying physicaltype (e.g. `string_descriptor*`), so we need to add that line of code. I've added a low priority task here https://github.com/lfortran/lfortran/issues/6892#issue-2977394717 to make the API tolerate both pointer to the physicalType (ptr) and also the physicalType itself (register-value).

PR #8159, opened 2025-07-28
feat : Implement Submodules in separate files for monolithic compilation 

AnshMehta1:
Towards #536 

Features of PR :-
1. Parent Module Flag in Module ASR is modified to string from Symbol to simplify deserialization process while loading a submodule. For Parent Module, now, this flag would be an empty string while for submodule, it would be parent module name.
2. A new flag `has_submodules` is introduced in Module ASR which represnets whether a module can have submodules in fututre or not. For a module, we set this flag to true if we find a `Function Node` in module having `m_module` field in `FunctionType` as true. For Submodule, we set this as false.
3. We add parent module name as prefix while saving modfile (`.smod`) for submodules. We also create an empty modfile (`.mod`) for submodule using only Submodule name to satisy CMAKE condition.
4. Two utility functions `load_submodules` and `find_and_load_submodules` are created to load all submodules corresponding to a parent module using parent module name.
5. When we encounter a use statement, we check if this is in program scope and loaded parent module (correspondng to use statement) has flag `has_submodules` as true. If these conditions satisfy, then we use `load_submodules` function to load all the corresponding submodules in global scope.
6. A new label `llvm_submodule` is created for testing submodules in separate files for monolithic compilation.
7. CI Exhaustive checks are also updated to run integration tests with `llvm_submodule` label.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2518,11 +2518,11 @@ RUN(NAME separate_compilation_13 LABELS gfortran llvm EXTRAFILES separate_compil
 
 RUN(NAME separate_compilation_14 LABELS gfortran llvm EXTRAFILES separate_compilation_14a.f90)
 RUN(NAME separate_compilation_15 LABELS gfortran llvm EXTRAFILES separate_compilation_15a.f90)
-RUN(NAME separate_compilation_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_16a.f90 separate_compilation_16b.f90 EXTRA_ARGS --separate-compilation)
-RUN(NAME separate_compilation_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_17a.f90 separate_compilation_17b.f90 EXTRA_ARGS --separate-compilation)
-RUN(NAME separate_compilation_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_18a.f90 separate_compilation_18b.f90 EXTRA_ARGS --separate-compilation)
-RUN(NAME separate_compilation_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_19a.f90 separate_compilation_19b.f90 EXTRA_ARGS --separate-compilation)
-RUN(NAME separate_compilation_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90 EXTRA_ARGS --separate-compilation)
+RUN(NAME separate_compilation_16 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_16a.f90 separate_compilation_16b.f90)
+RUN(NAME separate_compilation_17 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_17a.f90 separate_compilation_17b.f90)
+RUN(NAME separate_compilation_18 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_18a.f90 separate_compilation_18b.f90)
+RUN(NAME separate_compilation_19 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_19a.f90 separate_compilation_19b.f90)
+RUN(NAME separate_compilation_20 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90)
 
 
 
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -370,6 +370,8 @@ jobs:
             ./run_tests.py -b llvm -sc
             ./run_tests.py -b llvm llvmImplicit
             ./run_tests.py -b llvm llvmImplicit -f -nf16
+            ./run_tests.py -b llvm_submodule
+            ./run_tests.py -b llvm_submodule -sc
 
       - name: Test Linux LLVM 17-20
         if: ${{(contains(matrix.llvm-version, '17') || contains(matrix.llvm-version, '18') || contains(matrix.llvm-version, '19') || contains(matrix.llvm-version, '20'))}}
@@ -380,6 +382,8 @@ jobs:
             ./run_tests.py -b llvm -sc
             ./run_tests.py -b llvm llvmImplicit
             ./run_tests.py -b llvm llvmImplicit -f
+            ./run_tests.py -b llvm_submodule
+            ./run_tests.py -b llvm_submodule -sc
 
       - name: Check SciPy Build and Test Run - Test SciPy
         shell: bash -e -x -l {0}


certik:
Is this ready for review? I saw it was marked Draft again --- are there some issues to resolve first?

AnshMehta1:
> Is this ready for review? I saw it was marked Draft again --- are there some issues to resolve first?

No, there are just some tests that are left to be added. I just added 1 test for reviewing our new llvm_submodule label. I will add remaining test tomorrow and then, pr will be ready for merging.

src/libasr/asr_utils.cpp
```diff
@@ -969,6 +969,71 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
...
+            diag::Diagnostics diagnostics;
+            if (!asr_verify(*tu, true, diagnostics)) {
+                std::cerr << diagnostics.render2();
+                throw LCompilersException("Verify failed");
+            };
+    #endif
+        }
+        symtab->asr_owner = orig_asr_owner;
+    }
+}
```
certik: How does this differ from `load_module`? You don't need to load other modules recursively here?

src/libasr/ASR.asdl
```diff
@@ -10,7 +10,7 @@ unit
 
 symbol
     = Program(symbol_table symtab, identifier name, identifier* dependencies, stmt* body, location start_name, location end_name)
-    | Module(symbol_table symtab, identifier name, identifier* dependencies, symbol? parent_module, bool loaded_from_mod, bool intrinsic, location start_name, location end_name)
+    | Module(symbol_table symtab, identifier name, identifier parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, bool has_submodules, location start_name, location end_name)
```
certik: Instead of passing in an empty string, why not:
```suggestion
    | Module(symbol_table symtab, identifier name, identifier? parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, bool has_submodules, location start_name, location end_name)
```

src/libasr/asr_utils.cpp
AnshMehta1: In load module, we can just load one module, whose name is provided in argument. But, I wanted to load all the submodules for a parent module using its name.

AnshMehta1: > How does this differ from `load_module`? You don't need to load other modules recursively here?

load_submodules function doesn’t load other modules. It just load all the submodules for a parent module.

src/libasr/ASR.asdl
AnshMehta1: Yes, i think we can do this. Will update this in the pr.

src/libasr/asr_utils.cpp
certik: Yes, but submodules can use other modules, right? So I think we need to load them recursively here also.

AnshMehta1: Like, how should we load them recursively. Actually, i didn’t got the exact idea

AnshMehta1: > Yes, but submodules can use other modules, right? So I think we need to load them recursively here also.

Also, i am just loading submodule in global scope. I am not importing any entity of submodule in program or wherever load_submodules is called. So, I don’t think it will cause the issue

certik: In the other module loading function, when you load one `.mod` file, it also loads all other modules that this modfile "uses". And it does it recursively.

My question is why is this not needed for submodules? If a submodule B uses a module C, and module C is not used from anywhere else, how do we ensure that module C gets loaded?

AnshMehta1: Yes, I got the actual question now. I think, I forgot to add this functionality in my new function. Thanks for pointing this out. Will handle this.

src/libasr/ASR.asdl
AnshMehta1: I don't think we can have this change as while building, I am getting following errors :-
```
sic_iso_fortran_env.f90 -o CMakeFiles/lfortran_intrinsic_iso_fortran_env.dir/pure/lfortran_intrinsic_iso_fortran_env.f90.o
std::exception: basic_string: construction from null is not valid
```

src/libasr/asr_utils.cpp
AnshMehta1: Added the implementation to recursively load all the dependant modules for loaded submodules.

src/libasr/ASR.asdl
certik: You have to check if the `parent_module` pointer is null first. If it is, then it is not present, so do not construct a string from it. If it is not null, then there is a parent module present and the string shows its name.

certik:
To move forward, let's do the following.

Split the PR into two. In the first PR just do this change:

```
    | Module(symbol_table symtab, identifier name, identifier? parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, bool has_submodules, location start_name, location end_name)
```

And fill in the `has_submodules` and `parent_module` correctly using the logic from this PR. Update all tests. Use the `identifier?`, not `identifier`. Set it to null if there is no parent module. If there is a parent module, set it to its name. This should be straightforward, and we can merge it.


Then do the rest of this PR in a second PR, which should be quite a small change now, as no tests need to be updated, and we should be able to review & get it in quickly.

src/libasr/asr_utils.cpp
certik: Ok. So at this point, why to duplicate the function with almost the same functionality? 

Instead, why not add a flag/parameter into the same function and use it for both modules and submodules? And the submodule specific logic will be hidden behind a flag?

AnshMehta1: Still we will need a new function `find_and_load_submodules` as their we will load all submodule files instead of just one module. For `load_module` function, the problem is we return the loaded module but we won't be able to do so for all submodules at a time. Also, `load_module` is written just to handle a single module. So, this would create a lot of if-else which would infact duplicate the code within same function. Better to create new functions I feel.

integration_tests/CMakeLists.txt
```diff
@@ -2500,11 +2517,11 @@ RUN(NAME separate_compilation_13 LABELS gfortran llvm EXTRAFILES separate_compil
...
-RUN(NAME separate_compilation_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_19a.f90 separate_compilation_19b.f90 EXTRA_ARGS --separate-compilation)
-RUN(NAME separate_compilation_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90 EXTRA_ARGS --separate-compilation)
+RUN(NAME separate_compilation_16 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_16a.f90 separate_compilation_16b.f90)
+RUN(NAME separate_compilation_17 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_17a.f90 separate_compilation_17b.f90)
+RUN(NAME separate_compilation_18 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_18a.f90 separate_compilation_18b.f90)
+RUN(NAME separate_compilation_19 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_19a.f90 separate_compilation_19b.f90)
+RUN(NAME separate_compilation_20 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_20a.f90 separate_compilation_20b.f90)
```
certik, line 2525: Do these run with both separate compilation and monolithic compilation? It seems to me they now run with monolithic compilation only, not separate compilation anymore.

I think we need to test both.

certik:
This now looks good, the changes are quite minimal and reusing existing machinery. I have just one (important) question: https://github.com/lfortran/lfortran/pull/8159#discussion_r2245737967.

integration_tests/CMakeLists.txt
certik: I think you need to keep the original, but also add `llvm_submodule` in addition to all the other labels.

certik:
@AnshMehta1 let's enable both separate and monolithic compilation tests and let's get this merged.

AnshMehta1:
> @AnshMehta1 let's enable both separate and monolithic compilation tests and let's get this merged.

We are running both separate and monolithic compilation tests currently using llvm submodule label. I have added both the scripts in ci exhaustive checks. -sc tag represents separate compilation. We can merge the pr

integration_tests/CMakeLists.txt
certik: I see, the `-sc` flag does it. I think that's fine, although it would be good to verify that this is indeed the case by hand (later).

This PR removed `llvm_wasm llvm_wasm_emcc`, but for now it's fine.

PR #8184, opened 2025-07-30
enh : Modify `Parent Module` flag in Module ASR and Add a new flag `has submodules` in Module ASR

AnshMehta1:
Towards #536 
Addresses https://github.com/lfortran/lfortran/pull/8159#issuecomment-3137015970

### DIFF
--- a/tests/reference/asr-array12-cb81afc.json
+++ b/tests/reference/asr-array12-cb81afc.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-array12-cb81afc.stdout",
-    "stdout_hash": "bf59a3a7e78e04c63e60a69ed315badae39ceab61fc2d27a4463daa8",
+    "stdout_hash": "522dc05ed3c7ca29bb4e828d296079d128a19c013af096b86dbd1752",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-array12-cb81afc.stdout
+++ b/tests/reference/asr-array12-cb81afc.stdout
@@ -339,9 +339,10 @@
                                 )
                         })
                     array12
+                    ()
                     [iso_c_binding
                     array12]
-                    ()
+                    .false.
                     .false.
                     .false.
                 ),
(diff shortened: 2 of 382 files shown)


AnshMehta1:
Fixing the Failures in a while.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -115,6 +115,25 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        for(auto &item : m->m_symtab->get_scope()){
+            if (ASR::is_a<ASR::Function_t>(*item.second)) {
+                ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(item.second);
+                ASR::FunctionType_t* func_type = ASR::down_cast<ASR::FunctionType_t>(func->m_function_signature);
+                if (func_type->m_module) {
+                    is_parent_module = true;
+                    break;
+                }
+            }
+        }
```
certik, line 133: Note: for now I think it's fine, but this loop can be avoided by simply using a visitor variable `is_parent_module` and fill it when visiting the function. However I think the loop above might be quite quick, it just iterates over the symbol table, so I think we can keep it for now, as it's simpler.

certik:
This is good, it provides the foundation for the second PR.

For the failure at the CI I think you need to check if the identifier is null first before using it as a string.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -115,6 +115,25 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
         // To Be Implemented
     }
 
+    void initialize_has_submodules(ASR::Module_t* m) {
+        if (!std::string(m->m_parent_module).empty()) {
```
certik: You need to fix this by first checking if `m->m_parent_module` is null.

AnshMehta1: Done. Actually, I thought I handled this and it just remained in the back of my mind which caused all the trouble.

AnshMehta1:
Once this gets merged, then I will rebase #8159 with the remaining commits.

AnshMehta1:
> I think that this is good. If CI passes, you can merge it.
> 
> Make sure you modified the reference tests in only one commit, since it's a lot of files, so that we don't have multiple commits doing and undoing changes there. As of now you did in just one commit, so it's good to merge.

Yes, here we have all reference updation in one commit.

PR #8186, opened 2025-07-30
fix: some regression tests in string refactor

akramhany:
This PR should fix the following regression tests in string refactor:
- `string_negative_start`
- `allocate_01`


assem2002 (approved):
Great work. Thanks for fixing that.

assem2002:
`allocate_10_LLVM_GOC` is failing, Does it work normally without `--separate-compilation` flag? or we have actual bug with the normal compilation?

tests/reference/asr-string_negative_start_index-643e9c0.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: The first index in string section is less than 1
+ --> tests/errors/string_negative_start_index.f90:5:15
+  |
+5 | print*, "s:", s(-1:4)
+  |               ^^^^^^^
```
certik, line 5: We should remove this test as it is already tested in continue_compilation. In subsequent PRs.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1928,6 +1928,22 @@ namespace LCompilers {
...
+        llvm::Value* null_ptr = llvm::ConstantPointerNull::get(static_cast<llvm::PointerType*>(str_data->getType()));
+
+        llvm::Value* is_not_null = builder->CreateICmpNE(str_data, null_ptr);
+
+        create_if_else(is_not_null, [&]() {
+            llvm::Value *fmt_ptr = builder->CreateGlobalStringPtr("runtime error: Attempting to allocate already allocated variable!\n");
+            print_error(context, *module, *builder, {fmt_ptr});
+            const int exit_code_int = 1;
+            llvm::Value *exit_code = llvm::ConstantInt::get(context, llvm::APInt(32, exit_code_int));
+            exit(context, *module, *builder, exit_code);
```
certik, line 1942: Later: we should call our C runtime for this, and we should also pass in location information, that we know at compile time here, and if it fails, print at least where it fails, and later a full stacktrace.

PR #8197, opened 2025-08-01
fix: create copy of rhs class when class type is not same at lhs and rhs

jinangshah21:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6237,23 +6237,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
 
             llvm::Value* value_class_ptr = llvm_utils->create_gep2(value_llvm_type, value_struct, 1);
             value_class_ptr = llvm_utils->CreateLoad2(wrapper_value_llvm_type->getPointerTo(), value_class_ptr);
             // bitcast to the correct type
-            value_class_ptr = builder->CreateBitCast(value_class_ptr, wrapper_target_llvm_type->getPointerTo());
-            value_class_ptr = llvm_utils->CreateLoad2(wrapper_target_llvm_type, value_class_ptr);
-
             llvm::Value* target_class_ptr = llvm_utils->create_gep2(target_llvm_type, target_struct, 1);
-            target_class_ptr = llvm_utils->CreateLoad2(wrapper_target_llvm_type->getPointerTo(), target_class_ptr);
+            if (val_struct != tar_struct) {   // Create copy of rhs type if both class type is not equal
```
certik, line 6253: Shouldn't there rather be an explicit cast in ASR if the types are not equal?

jinangshah21: ASR `Cast` node only supports below conversions:
```cpp
enum cast_kindType
{ // Types
    RealToInteger, IntegerToReal, LogicalToReal, RealToReal, IntegerToInteger, RealToComplex, IntegerToComplex, IntegerToLogical, RealToLogical, StringToLogical, StringToInteger, StringToList, ComplexToLogical, ComplexToComplex, ComplexToReal, ComplexToInteger, LogicalToInteger, RealToString, IntegerToString, LogicalToString, UnsignedIntegerToInteger, UnsignedIntegerToUnsignedInteger, UnsignedIntegerToReal, UnsignedIntegerToLogical, IntegerToUnsignedInteger, RealToUnsignedInteger, CPtrToUnsignedInteger, UnsignedIntegerToCPtr, IntegerToSymbolicExpression, ListToArray, PointerToInteger
};
```

certik: It needs to be extended, or a new node for casting classes is needed. These kinds of things have to be figured out in ASR. In the backend we should only be assigning and passing around the same type.

This change is small, so if it unblocks you, you can merge it, but please create an issue for this. Perhaps we need to design the class support in ASR better, but I think we should not be doing the above in the backend.

certik:
You can merge this, per https://github.com/lfortran/lfortran/pull/8197#discussion_r2249262792, but create an issue for the proper fix.

PR #8209, opened 2025-08-02
Added string_index_kind to ASR

swamishiju:
Added `string_index_kind` (name subject to change) enum in ASR for string section handling

### DIFF
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -136,7 +136,7 @@ expr
     | StringRepeat(expr left, expr right, ttype type, expr? value)
     | StringLen(expr arg, ttype type, expr? value)
     | StringItem(expr arg, expr idx, ttype type, expr? value)
-    | StringSection(expr arg, expr? start, expr? end, expr? step, ttype type, expr? value)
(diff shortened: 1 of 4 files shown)


src/libasr/ASR.asdl
```diff
@@ -136,7 +136,7 @@ expr
     | StringRepeat(expr left, expr right, ttype type, expr? value)
     | StringLen(expr arg, ttype type, expr? value)
     | StringItem(expr arg, expr idx, ttype type, expr? value)
-    | StringSection(expr arg, expr? start, expr? end, expr? step, ttype type, expr? value)
+    | StringSection(expr arg, string_index_kind kind, expr? start, expr? end, expr? step, ttype type, expr? value)
```
certik, line 139: Document in doc/ what the two options do and how the sections differ in Fortran / Python.

certik:
As discussed, this should not be needed. To go forward, we should just adjust the frontend to use the same ASR, preferably. I would start with that.

Later, we can represent some of this more exactly in ASR by adding `start_index` and `last_bound_exclusive` specific tags, not Python/Fortran tags.

Issue #8192, opened 2025-07-31
Handle associate variable properly in select type (Refactor Regression)


jinangshah21:
MRE:
```fortran
module m_types
  implicit none

  type :: fpm_cmd_settings
  end type

  type, extends(fpm_cmd_settings) :: fpm_new_settings
     integer :: version
  end type fpm_new_settings

end module m_types


program test_settings
  use m_types
  implicit none

  class(fpm_cmd_settings), allocatable :: cmd_settings

  select type(settings => cmd_settings)
    type is (fpm_new_settings)
      print *, "Version =", settings%version
    end select
end program test_settings
```

LFortran Error:
```console
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store %fpm_cmd_settings* %17, %fpm_cmd_settings_polymorphic* %settings, align 8
 %fpm_cmd_settings_polymorphic = type { i64, %fpm_cmd_settings* }
```

PR #8221, opened 2025-08-03
fix: check class type assignment in intrinsics

jinangshah21:
Towards #8192


certik:
Do you mind fixing the conflicts please? Sorry about that.

PR #8222, opened 2025-08-03
Intent in fix

### LINKED ISSUE #8219: Bug: Compiler does not diagnose modification of intent(in) argument via a nested subroutine call
LFortran fails to produce a compile-time error when an intent(in) dummy argument is passed to another subroutine that expects an intent(inout) or intent(out) argument. This allows the read-only argument to be illegally modified at runtime.

```
program test_intent_chain
  implicit none
  integer :: x = 10
  call sub_a(x)
contains
  subroutine sub_a(val_a)
    integer, intent(in) :: val_a

    ! This call is illegal because sub_b expects to modify its argument,
    ! but val_a is a read-only intent(in) variable.
    call sub_b(val_a)
  end subroutine sub_a

  subroutine sub_b(val_b)
    integer, intent(inout) :: val_b
    val_b = 20 
  end subroutine sub_b
end program test_intent_chain
```

**Observed Behavior with LFortran**
LFortran incorrectly compiles the code without any errors. When run, the program illegally modifies the variable x.
```
In main, before call, x =    10
In sub_a, after call, val_a =    20
In main, after call,  x =    20
```

**Gfortran behaviour:**
```
gfortran ../test_intent_in.f90 -o ../test_intent_in
../test_intent_in.f90:17:15:

   17 |     call sub_b(val_a)
      |               1
Error: Dummy argument ‘val_a’ with INTENT(IN) in variable definition context (actual argument to INTENT = OUT/INOUT) at (1)

```
Related to #8005

HiteshGhanchi:
Fixes #8219


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4216,6 +4216,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                    Level::Error, Stage::Semantic, {
+                                        Label("Argument passed here", {passed_arg_expr->base.loc})
+                                    }));
+                                throw SemanticAbort();
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
kmr-srbh: Let's add this block inside `Call_t_body()` in `asr_utils.h`. This will ensure we handle this for both subroutines and functions.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4216,6 +4216,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+
+                        // This checks only for the intent(in) violation.
+                        if (passed_var->m_intent == ASR::intentType::In) {
+                            ASR::Variable_t* callee_param = ASRUtils::EXPR2VAR(f->m_args[i]);
+                            if (callee_param->m_intent == ASR::intentType::Out ||
+                                callee_param->m_intent == ASR::intentType::InOut) {
+                                
+                                diag.add(Diagnostic(
+                                    "Argument '" + std::string(passed_var->m_name) +
+                                    "' has intent(IN) but is passed to a dummy argument with a modifying intent (OUT or INOUT)",
```
kmr-srbh:
```suggestion
                                    "Argument `" + std::string(passed_var->m_name) +
                                    "` with intent(in) passed to a dummy argument with modifying intent",
```

@certik what could be a better error message? Should we get the exact modifying intent, convert to string, and display in the error message?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4216,6 +4216,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                    Level::Error, Stage::Semantic, {
+                                        Label("Argument passed here", {passed_arg_expr->base.loc})
```
kmr-srbh: I think we probably do not need the secondary label here, since we already highlight the passed argument. We use the secondary label to provide some extra information, like help snippets.

tests/errors/intent_chain.f90
```diff
@@ -0,0 +1,26 @@
+program test_intent_chain
```
kmr-srbh: Let's add this test case inside `continue_compilation_02.f90`.

tests/errors/intent_chain.f90
```diff
@@ -0,0 +1,26 @@
...
+  integer :: x = 10
+
+  print *, "In main, before call, x =", x
+
+  call sub_a(x)
+
+  print *, "In main, after call,  x =", x
+
+contains
+  subroutine sub_a(val_a)
```
kmr-srbh: Please include a test case for `function` after moving the code inside `Call_t_body()`.

src/lfortran/semantics/ast_body_visitor.cpp
kmr-srbh: GFortran's current error message is:

```console
test.f90:502:15:

  502 |     call sub_b(val_a)
      |               1
Error: Dummy argument 'val_a' with INTENT(IN) in variable definition context (actual argument to INTENT = OUT/INOUT) at (1)
```

tests/errors/intent_chain.f90
HiteshGhanchi: Ok, will do so.
Is there a reason for adding in this particular file. In one of my other PR also, I was told to add testcases there.
How does one decide where to add which ?
Thanks.

certik: Always add everything in one of the continue_compilation file. In the past we couldn't recover from errors, but now we do, so we should ensure that all new errors that we add can be recovered from.

HiteshGhanchi: Understood.
Thanks.

kmr-srbh: We test the `continue_compilation_*.f90` files with the `--continue-compilation` flag, which helps us test additionally if we can recover from that error during compilation to move ahead and catch the other compilation errors. We report all the errors together at the end. So, without `--continue-compilation` flag, we stop at the first error, but with it, we catch and report the other errors as well.

src/lfortran/semantics/ast_body_visitor.cpp
HiteshGhanchi: The problem with this is that `diag` is not available in the `Call_t_body()` function. 
I was thinking of making another util function for this and add that check in both subroutine and functions.
If it is fine, I will do that.

tests/errors/intent_chain.f90
HiteshGhanchi: Ok, makes sense.
Thanks.

src/lfortran/semantics/ast_body_visitor.cpp
HiteshGhanchi: @kmr-srbh Since, I can't do it in `Call_t_body()`, I made a util function, but there is no `FuntionCall` function like we had for `SubroutineCall`.
So how do I add the check for function.

kmr-srbh: Please see `visit_FuncCallOrArray()` in `ast_common_visitor.h`. The function call is assigned to `tmp` by the `create_FunctionCallWithASTNode()` call.

kmr-srbh: Please rebase your commits on top of latest `main`. Currently, the PR includes many other commits also.

kmr-srbh: > Please rebase your commits on top of latest `main`. Currently, the PR includes many other commits also.

Thanks. This is done.

HiteshGhanchi: Sorry about that.
I wasn't paying attention while merging.
My bad.

HiteshGhanchi: @kmr-srbh I have added the logic for functions too

src/lfortran/semantics/ast_body_visitor.cpp
HiteshGhanchi: Removed the label

tests/errors/intent_chain.f90
HiteshGhanchi: Added the test

tests/errors/continue_compilation_2.f90
```diff
@@ -462,5 +462,7 @@ logical function f(x)
         integer, intent(in), optional :: x
         f = PRESENT(x)
     end function
+    call sub_a(intent_bug_sub_x)
+    print *, outer_func(intent_bug_sub_x)
```
kmr-srbh: Please move this outside the `contains` block and update the reference.

HiteshGhanchi: Done.

kmr-srbh: Thanks! This looks good!

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -260,6 +260,18 @@ semantic error: Cannot assign to an intent(in) variable `y`
 41 |         y = 99  
    |         ^ 
 
+semantic error: Argument `val_a` with intent(in) passed to a dummy argument with modifying intent
+  --> tests/errors/continue_compilation_2.f90:63:20
+   |
+63 |         call sub_b(val_a)
+   |                    ^^^^^ 
+
+semantic error: Argument `val_a` with intent(in) passed to a dummy argument with modifying intent
```
certik, line 269: I don't like the term `modifying` intent. Instead let's say:
```suggestion
semantic error: Argument `val_a` with intent(in) passed to a dummy argument with intent(inout)
```
And change `inout` to `out` on a case-by-case basis.

This can be done in a subsequent PR.

certik:
Let's wait if the CI passes here. If so, I'll squash merge it. The above improvement can be done in a subsequent PR.

PR #8225, opened 2025-08-03
fix : Use ExprStmtWithScopeDuplicator to update scope using destination symtab while duplicating variable

AnshMehta1:


certik:
I rebased it, as we merged a few PRs, just in case.

PR #8226, opened 2025-08-03
Fix an essential parser conflict for the common block parsing

### LINKED ISSUE #7702: Parser: remove 3 essential conflicts
See https://github.com/lfortran/lfortran/pull/7698#issuecomment-2968951922. They were introduced by a mistake in https://github.com/lfortran/lfortran/pull/5811, where I think I forgot to check for essential conflicts.

We need to either remove the changes, or change them so that there are no conflicts.

certik:
The grammar might work, there are no essential conflicts, but we now need to update the action rules to handle this changed grammar.

Once finished, it will fix #7702.

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -1395,14 +1395,7 @@ common_block_list_top
     | common_block_start common_block_object {
         COMMON_BLOCK_2($$, $1, $2, @$) }
     | common_block_list_top "," common_block_object {
-        $$ = $1;
-        LCompilers::LFortran::AST::common_block_t last = $$.back();
-        Vec<LCompilers::LFortran::AST::var_sym_t> v;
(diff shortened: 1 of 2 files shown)


certik:
The parser actions are not completely correct yet. Then we need to refactor them into nice macros.

certik:
The CI canceled on its own (https://github.com/lfortran/lfortran/issues/8349), I restarted the skipped jobs.

certik:
The parser actions should now be correct, all tests pass for me locally. The essential conflict is fixed.

certik:
The code in `semantics.h` can be cleaned up, but I think it's good enough to merge. The parser changes should be ok, all tests pass, the AST should be equivalent as before. I am going to merge it, since it makes the parser LALR(1) again, no essential conflicts remain.

Issue #8204, opened 2025-08-02
Need to create warning/error for `unsigned` as derived type name


swamishiju:
I've implemented in `unsigned` type in #8149 with the following syntax

```f90
unsigned :: x
type(unsigned) :: y
```

Given we didn't choose to call it `_lfortran_unsigned` this opens up possible issues if we create a derived type with the name `unsigned` so we should probably report and error/warning when this happens.

https://github.com/lfortran/lfortran/pull/8149#discussion_r2236771189

harperjf:
I suggest that it be a warning, not an error, because the Fortran standard would allow it, and that even that warning should not appear if the --std=f23 option is used.

PR #8227, opened 2025-08-03
Addresses #8204

yuvrajsinghgmx:
Addresses #8204 

This adds a test for unsigned

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -2036,6 +2036,15 @@ public:
             }
             vars_with_deferred_struct_declaration.erase(to_lower(x.m_name));
         }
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2036,6 +2036,15 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             }
             vars_with_deferred_struct_declaration.erase(to_lower(x.m_name));
         }
+        std::string derived_type_name = to_lower(std::string(x.m_name));
+        if (to_lower(derived_type_name) == "unsigned") {
+            diag.add(diag::Diagnostic(
+                "Defining a derived type named 'unsigned' is not allowed.",
```
certik, line 2042: It's allowed to define an unsigned derived type. It will then take precedence over the builtin unsigned type.

PR #8229, opened 2025-08-04
tests: fix incorrect tests with errors found from bounds checking

xaerru:
These errors were found on https://github.com/lfortran/lfortran/pull/7955.

The test `lp_dict_test_01.f90` also fails with bounds checking, but I don't know the best way to remove the error here.
In `lp_dict_test_01.f90`:
```fortran
   if (abs(_lfortran_get_item(dict_i, -14) - 6.4) < eps) error stop
   if (abs(_lfortran_get_item(dict_i, 2) - 2.5) < eps) error stop
```
Is the sign supposed to be `>` or is it correct?
In `dict_i` key `-14` doesn't exist, so it throws
```shell
KeyError: The dict does not contain the specified key
```

### DIFF
--- a/integration_tests/arrays_op_20.f90
+++ b/integration_tests/arrays_op_20.f90
@@ -3,7 +3,7 @@ implicit none
 
 real, allocatable :: array(:, :), arrayoutput(:, :)
 
-allocate(array(5, 5))
+allocate(array(3, 3))
 arrayoutput = f(5, array)
 print *, size(arrayoutput)
 if( size(arrayoutput) /= 24 ) error stop
@@ -12,7 +12,7 @@ contains
 
 function f(m, input) result(output)
 integer :: m
-real :: input(m, m)
+real :: input(m)
 real :: output(2:m, m:2*m)
 end function
 
--- a/integration_tests/passing_array_01.f90
+++ b/integration_tests/passing_array_01.f90
@@ -57,7 +57,7 @@ MODULE passing_array_01_mod
       len = 20
       arr = [(i, i = 1, 40)]
       ! All calls would accept the passed array,
-      ! because the passed array's length is greater than expected in the function
+      ! because the passed array's length is greater than or equal to the expected length in the function
 
       ret =  test_01_interface (len,arr)
       print * , ret
--- a/integration_tests/passing_array_03.f90
+++ b/integration_tests/passing_array_03.f90
@@ -12,7 +12,7 @@ MODULE passing_array_03_mod
     END FUNCTION test_01
 
     subroutine test_entry
-        INTEGER, DIMENSION(4,2) :: value
+        INTEGER, DIMENSION(6,2) :: value
         integer :: ret
         ret =  test_01(value)
         print *, ret
--- a/integration_tests/passing_array_04.f90
+++ b/integration_tests/passing_array_04.f90
@@ -13,12 +13,12 @@ MODULE passing_array_04_mod
 
     subroutine test_entry
         integer :: len
-        INTEGER, DIMENSION(4, 2) :: value
+        INTEGER, DIMENSION(6, 2) :: value
         INTEGER :: ret
-        len = 2
+        len = 3
         ret = test_01(len,value)
         print * , ret
-        if(ret /= 8) error stop 
+        if(ret /= 12) error stop 
     end subroutine test_entry
  
 END MODULE passing_array_04_mod


integration_tests/array_indices_array_item.f90
```diff
@@ -52,8 +52,8 @@ program array_indices_array_item
     print *, rank_val
     if (rank_val /= 1) error stop
 
-    allocate(arr(1, 3))
-    arr = reshape([1, 2, 3], shape(arr))
+    allocate(arr(2, 3))
+    arr = reshape([1, 2, 3, 4, 5, 6], shape(arr))
```
certik, line 56: Why is the original invalid? It seems valid to me.

integration_tests/arrays_45.f90
```diff
@@ -22,5 +22,5 @@ program arrays_45
 cs = [1.0, 2.0, 3.0, 4.0]
 cs2 = 12.87
 call a(cs)
-call b(cs)
+call b(cs2)
```
certik, line 25: This is correct. `b` expects either (4, 4) array, or (16,) array, but it was provided (4,) array, so that's wrong.

integration_tests/arrays_51.f90
```diff
@@ -7,13 +7,13 @@ module arrays_51_mod
...
-        if(size(arr) /= 20000) error stop
+        if(size(arr) /= 16) error stop
     end subroutine sub
 end module arrays_51_mod
 
 
 program arrays_51
     use arrays_51_mod
-    i = 5000
+    i = 4
```
certik: Let's use 6 instead of 4.

certik: Change `arr_mod` to (4, 6).

integration_tests/array_indices_array_item.f90
xaerru: On line no. 57 `arr` is indexed by `arr_idx` in the first dimension. `arr_idx` is `[2, 1]`. So the first dimension should at least have size 2.

integration_tests/array_03.f90
```diff
@@ -6,7 +6,7 @@ program array_03
     real(RP):: zmat(1:1, 1:1) 
     knew = 1
     zmat = reshape([1.0_RP], [1, 1])
-    G = planerot(zmat(knew, [1_IK, 2]))
+    G = planerot(zmat(knew, [1_IK, 1]))
```
certik, line 9: Good.

integration_tests/arrays_08_size.f90
```diff
@@ -7,11 +7,11 @@ subroutine temp(x)
         integer, intent(inout) :: x
         logical :: keep(x)
         integer :: result(x)
-        keep = [.false., .true.]
+        keep = [.true., .true.]
         x = 1
         result = trueloc(keep)
         print * , result
-        if (result(1) /= 2) error stop
+        if (any(result /= [1, 2])) error stop
```
certik, line 14: The previous was incorrect? It assigned length one array from `trueloc` into length two array in `result`, right? That is indeed incorrect. If you could confirm this, that would be great.

integration_tests/arrays_51.f90
```diff
@@ -2,18 +2,18 @@
...
-        if(size(arr) /= 20000) error stop
+        if(size(arr) /= 24) error stop
     end subroutine sub
 end module arrays_51_mod
 
 
 program arrays_51
     use arrays_51_mod
-    i = 5000
+    i = 6
```
certik, line 17: This change is good --- the `arr` parameter array used to be bigger than `arr_mod` argument, which is wrong.

integration_tests/arrays_52.f90
```diff
@@ -1,10 +1,10 @@
 ! ! Check that lcompilers_get_i() returns mutated value of i not just hard coded value of 4
 program arrays_52
     implicit none
-    integer(4), dimension(6) :: cs
+    integer(4), dimension(11) :: cs
     integer(4), save :: nx = 4
  
-    cs = [1, 2, 3, 4, 5, 6]
+    cs = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
```
certik, line 7: This change is good --- the second `test_2` is accepting 10 elements, and we were passing only 6, which was wrong.

integration_tests/arrays_op_20.f90
```diff
@@ -12,7 +12,7 @@ program arrays_op_20
 
 function f(m, input) result(output)
 integer :: m
-real :: input(m)
+real :: input(m, m)
```
certik: This original code is correct I think. The original array is (3,3), total 9 elements. We are passing it to `f` that is accepting 5 elements. This is allowed. Reshaping is also allowed.

Your change is also good, but that is a separate test (if needed, add this as a separate test if we don't already have it).

integration_tests/bindc4.f90
```diff
@@ -39,7 +39,7 @@ program bindc4
     end do
 end do
 
-if (c_associated(queries, c_loc(x(0, 0)))) error stop
+if (.not. c_associated(queries, c_loc(x(1, 1)))) error stop
```
certik, line 42: Why is this needed --- and gfortran was passing both before and after this change? That's weird.

certik: I see, `x` starts from 1. So the previous code was indeed incorrect.

certik: I think this change is good.

integration_tests/passing_array_01.f90
```diff
@@ -51,12 +51,13 @@ END FUNCTION test_03
...
-      INTEGER, DIMENSION(6) :: arr
+      INTEGER, DIMENSION(40) :: arr
+      INTEGER :: i
       INTEGER :: ret
       len = 20
-      arr = [1,2,3,4,5,6]
+      arr = [(i, i = 1, 40)]
       ! All calls would accept the passed array,
-      ! as they don't have compile-time array size to compare against.
+      ! because the passed array's length is greater than expected in the function
```
certik:
```suggestion
      ! because the passed array's length is greater than or equal to the expected length in the function
```

integration_tests/passing_array_02.f90
```diff
@@ -14,25 +14,23 @@ MODULE passing_array_02_mod
...
+        ! Slicing is premesible (3 <= 30), so accept it.  
         INTEGER, DIMENSION(3,1),intent(in) :: value 
         INTEGER :: res
         res = size(value)
     END FUNCTION test_02
 
     subroutine test_entry
         integer:: len
-        INTEGER, DIMENSION(6,2) :: arr
+        INTEGER, DIMENSION(3, 10) :: arr
```
certik, line 33: I think that's good.

integration_tests/passing_array_01.f90
```diff
@@ -51,12 +51,13 @@ END FUNCTION test_03
     SUBROUTINE test_entry
       implicit none
       INTEGER :: len
-      INTEGER, DIMENSION(6) :: arr
+      INTEGER, DIMENSION(40) :: arr
+      INTEGER :: i
       INTEGER :: ret
       len = 20
-      arr = [1,2,3,4,5,6]
+      arr = [(i, i = 1, 40)]
```
certik, line 58: I think that's good.

integration_tests/passing_array_03.f90
```diff
@@ -12,7 +12,7 @@ FUNCTION test_01 (value) result(res)
     END FUNCTION test_01
 
     subroutine test_entry
-        INTEGER, DIMENSION(6,2) :: value
+        INTEGER, DIMENSION(4,2) :: value
```
certik: The original was good. The size is 12, and the function accepts 8. That must work.

integration_tests/passing_array_04.f90
```diff
@@ -6,21 +6,19 @@ MODULE passing_array_04_mod
...
-        ! Run-time size (40000/12).
         INTEGER, DIMENSION(2,2,len),intent(in) :: value 
         INTEGER :: res
         res = size(value)
     END FUNCTION test_01
 
     subroutine test_entry
         integer :: len
-        INTEGER, DIMENSION(6,2) :: value
+        INTEGER, DIMENSION(4, 2) :: value
```
certik: Keep the original, total length 12.

integration_tests/passing_array_04.f90
```diff
@@ -6,21 +6,19 @@ MODULE passing_array_04_mod
...
         INTEGER :: ret
-        len = 10000
+        len = 2
```
certik: Use:
```suggestion
        len = 3
```
So the total length inside the function will be (2, 2, 3) total of 12.

integration_tests/passing_array_04.f90
```diff
@@ -6,21 +6,19 @@ MODULE passing_array_04_mod
...
         ret = test_01(len,value)
         print * , ret
-        if(ret /= 40000) error stop 
+        if(ret /= 8) error stop
```
certik:
```suggestion
        if(ret /= 12) error stop 
```

integration_tests/passing_array_04.f90
```diff
@@ -6,21 +6,19 @@ MODULE passing_array_04_mod
...
-        INTEGER, DIMENSION(6,2) :: value
+        INTEGER, DIMENSION(4, 2) :: value
         INTEGER :: ret
-        len = 10000
+        len = 2
         ret = test_01(len,value)
         print * , ret
-        if(ret /= 40000) error stop 
+        if(ret /= 8) error stop 
     end subroutine test_entry
```
certik, line 22: Otherwise good. The function was expecting (2, 2, 10000) total of 40,000, and we were passing in 12, which is wrong.

certik:
Ok I finished it. Most of these changes are good, but some are not good. So please remove the not good changes, but do it by pushing more commits. Don't force push, don't rebase. Otherwise we will lose my comments and I will have to do the review again, and it takes a very long time.

integration_tests/array_indices_array_item.f90
certik: That is correct. I see.

Ok, the previous code was indeed incorrect, now it is correct. The change is good.

integration_tests/arrays_08_size.f90
xaerru: Yes, that's why the previous test was incorrect.

certik:
I think this is good as it currently is. Let's get it in.

PR #8236, opened 2025-08-04
Fix Regression ( `pass_array_by_data_05`)

assem2002:
checking CI


assem2002:
apparently the bug is still there, and it's subtle :)

certik:
It fails with:
```
1484/1782 Test #1484: pass_array_by_data_05_FAST ..............................***Failed    0.01 sec
3
1.00677021e+31
ERROR STOP
```

Pranavchiku (approved):

PR #8244, opened 2025-08-04
use `np=1,2` with POT3D validation test ( Use the new branch )

assem2002:


certik:
I think you now need to enable the `--fast` checks at the CI, to ensure they actually work.

assem2002:
Ah, that's really really odd. It used to pass with `np = 2` now it fails, and only passes with `np=1`. and some llvm versions passed. 
It's really a subtle one.

PR #8245, opened 2025-08-04
Turn off Exhaustive Checks on pull requests only 

assem2002:


.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -6,9 +6,9 @@ on:
       - main
     tags:
       - 'v*'
-  pull_request:
-    branches:
-      - main
+  # pull_request:
+  #   branches:
+  #     - main
```
certik, line 11: This turns off all testing on pull requests.

Instead, there was a way to turn it off by default, but making it optional. You click some button and it will run. But I can't find it right now how to do it....

PR #8254, opened 2025-08-05
Throw Error when referencing BOZ with greater than 64 Bits

### LINKED ISSUE #8253: Stoull Error when passing BOZ Literals larger than 64 Bits
MRE:
```
program long_boz
implicit none
integer::a = INT(z'F111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111')
PRINT *,a
end program
```

```
$ lfortran a.f90
std::exception: stoull
```

```
$gfortran a.f90 
a.f90:4:132:
    4 | nteger::a = INT(z'F111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111')
      |                                                                                                                                   1
Error: Line truncated at (1) [-Werror=line-truncation]
a.f90:4:132:
    4 | nteger::a = INT(z'F111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111')
      |                                                                                                                                   1
Error: Illegal character in BOZ constant at (1)
```

Stoull is a C++ Function, that converts BOZ to Unsigned Bits, which accepts only upto 64 Bits characters. So, an error message needs to be added, for passing BOZ Literals larger than 64 Bits.

YashNagda17:
Fixes #8253 

Added a check, which throws semantic error, when large BOZ Literals are encountered. This gives a better Error Message than `std::exception: stoull` which main currently throws.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -11086,6 +11086,16 @@ public:
             throw SemanticAbort();
         }
         std::string boz_str = s.substr(2, s.size() - 2);
+        //Check if BOZ string has more than 64 Bits, else stoull Throws Error
+        if ((((s[0]=='b') || (s[0]=='B')) && (boz_str.size()> 65)) || 
+            (((s[0]=='o') || (s[0]=='O')) && (boz_str.size()> 22)) || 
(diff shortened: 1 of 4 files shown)


kargl:
No. It is not good.  The Fortran standard requires that the string representation
is truncated to an appropriate bit width.

Issue #8241, opened 2025-08-04
String Refactor: Null Terminated to Non Null Terminated


sriganeshres:
**String Refactor Preparation Plan**

* Goals:

  * Know string length at all times (runtime or compile-time)
  * Use non-null-terminated strings eventually
  * Depend on length instead of null character
  * **Do not** remove null character yet — just prepare functions
  * Avoid `strlen` / null-terminated assumptions
  * Use length-aware APIs or copy to C-style string if needed

* Functions to refactor:

  * [x] `_lfortran_printf` (https://github.com/lfortran/lfortran/pull/8242)
  * [x] `remove_spaces_except_quotes` (https://github.com/lfortran/lfortran/pull/8259)
  * [x] `find_matching_parentheses` (#8280)
  * [x] `parse_fortran_format` (#8280)
  * [ ] `_lfortran_strrepeat`
  * [x] `_lfortran_open` (#8289)
  * [x] `_lfortran_inquire` (#8311)
  * [x] `_lfortran_read_char` (#8322)
  * [x] `_lfortran_formatted_read` (#8322)
  * [x] `_lfortran_file_write` (#8354)
  * [x] `_lfortran_string_write` (#8354)
  * [x] `_lfortran_string_read_i32` (#8289)
  * [x] `_lfortran_string_read_i64` (#8289)
  * [x] `_lfortran_string_read_f32` (#8289)
  * [x] `_lfortran_string_read_f64` (#8289)
  * [x] `_lfortran_string_read_bool` (#8321)
  * [x] `remove_whitespace` (#8321)
  * [x] `print_stacktrace_addresses` (#8321)
  * [x] `_lfortran_close` (#8321)

harperjf:
The length of a Fortran allocatable string is usually not knowable at compile time and it may change more than once at run time.

certik:
Yes, the length is not known at compile time, but we always have a way to find it at run time. Yes, it can change. We handle all that already, now it's just updating our runtime functions to not be using null termination but rather work with this (runtime) length.

harperjf:
One case where lfortran gets the length wrong is revealed by the first print statement in this program:
```
program allocstring
  implicit none
  integer:: maxlen = range(1)+2, n = 42
  character(:), allocatable:: string
  allocate(character(maxlen):: string) ! needed in F2018 but not in F2023
  write(string,"(I0)") n 
  print "(3A,I0)", 'string = "',string,'" len(string) = ',len(string)
[...]

PR #8259, opened 2025-08-05
refactor: make remove_spaces_except_quotes null termination independent 

akramhany:
Towards #8241

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -708,17 +708,16 @@ within character string edit descriptor
 
 E.g.; "('Number : ', I 2, 5 X, A)" becomes '('Number : ', I2, 5X, A)'
 */
-char* remove_spaces_except_quotes(const char* format) {
-    int len = strlen(format);
-    char* cleaned_format = malloc(len + 1);
+char* remove_spaces_except_quotes(const char* format, const int64_t len, int* cleaned_format_len) {
+    char* cleaned_format = malloc(len + 1);   // +1 for optional '\0' at end
 
-    int i = 0, j = 0;
+    int j = 0;
     // don't remove blank spaces from within character
     // string editor descriptor
     bool in_quotes = false;
     char current_quote = '\0';
 
-    while (format[i] != '\0') {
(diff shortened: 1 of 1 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1553,12 +1551,12 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, int64_t
     int64_t format_values_count = 0,item_start_idx=0;
     char** format_values;
     char* modified_input_string;
-    char* cleaned_format = remove_spaces_except_quotes(format);
+    int len = 0;
+    char* cleaned_format = remove_spaces_except_quotes(format, format_len, &len);
```
assem2002, line 1555: It's not necessary to make the return of `remove_spaces_except_quotes` null-independent, as the return isn't consumed by Fortran (it's a utility in the C runtime library). Anyways it's not a bit deal.

Why not to change the type ` char*` -> `int8_t*` whenever we have a fortran string data, that way we have some way to distinguish between Fortran strings (no-null-terminated) and C-Strings (null-terminated); It means the signature would be  
```C
char* remove_spaces_except_quotes(const int8_t* format, const int64_t len)
```

assem2002 (approved):
The changes looks great and ready to merge. Thanks you.
Left a minor suggestion, If you agree, let's make it the standard and share that thought in Zulip for others to follow

src/libasr/runtime/lfortran_intrinsics.c
akramhany: Indeed, changing `char*` to `int8_t*` is a very good way to distinguish between Fortran strings and C-Strings, thanks for suggestion!
I think we can even make a `typdef` for `int8_t` to something like `fchar` so that it would be more clear that this is a Fortran string, what do you think?

We should have something like this:
```c
typedef int8_t fchar;
char* remove_spaces_except_quotes(const fchar* format, const int64_t len)
```

assem2002: even better. :rocket:

Pranavchiku:
Is this good to merge?

PR #8262, opened 2025-08-06
feat: implement type to string functionality for `ASR::StructType_t` using cached `ASR::StructType_t*` nodes

kmr-srbh:
We create a set of all struct names and a map from `ASR::ttype_t*` to `std::string*`. The idea behind this implementation is that a `ASR::ttype_t*` can be associated to the struct it was created from, and hence get access to the name.

### Program
```fortran
program main
   implicit none

   type :: Home
      integer :: x
      integer :: y
   end type

   type :: School
      integer :: y
   end type

   type(Home) :: h
   type(School) :: s

   h = s
end program
```

#### On `main`

```console
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
  --> test.f90:16:4
   |
16 |    h = s
   |    ^   ^ type mismatch (derived type and derived type)


```

#### On branch

```console
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
  --> test.f90:16:4
   |
16 |    h = s
   |    ^   ^ type mismatch (home and school)


```

### DIFF
--- a/src/libasr/asr_utils.cpp
+++ b/src/libasr/asr_utils.cpp
@@ -65,6 +65,14 @@ public:
         ASRUtils::struct_names.insert(x.m_name);
         ASRUtils::map_struct_name_to_type(x.m_name, x.m_struct_signature, true);
         ASRUtils::map_struct_type_to_name(x.m_struct_signature, x.m_name);
+
+        for (auto& var : x.m_symtab->get_scope()) {
+            if (ASR::is_a<ASR::Variable_t>(*var.second)
+                && (ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(
+                    ASR::down_cast<ASR::Variable_t>(var.second)->m_type)))) {
+                StructTypeToNameVisitor::visit_symbol(*var.second);
+            }
+        }
     }
 
     void visit_Variable(const ASR::Variable_t& x)
@@ -76,7 +84,7 @@ public:
             ASRUtils::map_struct_name_to_type(
                 ASRUtils::symbol_name(ASRUtils::symbol_get_past_external(x.m_type_declaration)),
                 &struct_type->base,
-                true);
+                struct_type->m_is_cstruct);
             ASRUtils::map_struct_type_to_name(
                 &struct_type->base,
                 ASRUtils::symbol_name(ASRUtils::symbol_get_past_external(x.m_type_declaration)));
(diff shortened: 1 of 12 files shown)


kmr-srbh:
The implementation works correctly, until we encounter newly created types of the same struct. In that case, we do not have the type pointer in our map. The solution to this issue is to reuse the same type which is created from a struct.

kmr-srbh:
@certik could you please review the design here? Does it look good to you?

certik:
I don't think that's a good design:

> The idea behind this implementation is that a ASR::ttype_t* can be associated to the struct it was created from, and hence get access to the name.

The StructType ttype, especially after we deduplicate it, doesn't have a 1:1 correspondence with a Struct symbol. For example, the two structs:
```
   type :: A
      integer :: x
      integer :: y
   end type

   type :: B
      integer :: y
      integer :: z
   end type
```
Will have two different Struct symbols in the symbol table, but the same ttype, after deduplication. The structural type is the same for A and B. They are equal types, in terms of structural types.

I think the solution to print A and B in error messages is that you must consult the symbol table, not just the ttype.

kmr-srbh:
> Will have two different Struct symbols in the symbol table, but the same ttype, after deduplication.

We are storing pointers for the types actually. So, the ttype created for `A` is located at a different address (on heap) and `B` at a different one. Here is an example after #8269 was merged:

```fortran
program main
   implicit none

   type :: A
      integer :: x
      integer :: y
   end type

   type :: B
      integer :: y
      integer :: z
   end type

   type(A) :: A_var
   type(B) :: B_var

   A_var = B_var
end program
```
```console
test.f90:17:11:

   17 |    A_var = B_var
      |           1
Error: Cannot convert TYPE(b) to TYPE(a) at (1)
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
  --> test.f90:17:4
   |
17 |    A_var = B_var
   |    ^^^^^   ^^^^^ type mismatch (a and b)


```

By reusing types, I meant to refer to reusing the ttype pointer.
- A `StructType` node is created on heap for a new struct declaration.
- Reuse this same ttype pointer for every expression related to that struct instead of allocating a new ttype.

certik:
> Reuse this same ttype pointer for every expression related to that struct instead of allocating a new ttype.

Yes, we have to do that.

src/libasr/asr_utils.cpp
```diff
@@ -19,6 +19,79 @@ namespace LCompilers {
...
 
+std::map<ASR::ttype_t*, const std::string*> struct_type_to_struct_name;
+
+// A struct type variable can have 2 types - non-polymorphic and polymorphic, so
+// store both as value in an array. 
+// We use index 0 for `is_cstruct == false` and 1 for `true`.
+std::map<const std::string, ASR::ttype_t* [2]> struct_name_to_struct_type;
+
+// Stores names of all declared `ASR::Struct_t` objects.
+std::unordered_set<std::string> struct_names;
```
certik, line 30: I don't like global variables --- why do we need them? What if we do multiple compilations, such at dev.lfortran.org or VSCode server?

What if we run in parallel, such as for the VSCode server?

kmr-srbh:
Fixed in #8329.

PR #8268, opened 2025-08-06
fix: incorrect usage of registered allocated at nested branch for struct vtab

### LINKED ISSUE #8240: Class vtab ptr got out of scope when another type is passed as argument in type bound procedure
MRE:
```fortran
module installer_mod
    implicit none

    type :: installer_t
    contains
        procedure :: install_header
        procedure :: install
    end type installer_t

    type :: error_t
    end type

contains

    subroutine install_header(self, error)
        class(installer_t), intent(inout) :: self
        type(error_t), allocatable, intent(out) :: error
        if (.true.) then
            call self%install()
        end if
        call self%install()
    end subroutine install_header

    subroutine install(self)
        class(installer_t), intent(in) :: self
    end subroutine install

end module installer_mod


program test_installer
    use installer_mod
end program test_installer
```

LFortran Error:
```console
Instruction does not dominate all uses!
  %2 = alloca %__vtab_installer_t, align 8
  %19 = bitcast %__vtab_installer_t* %2 to <{ i64 }>*
```

Pranavchiku:
Fixes #8240.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -147,6 +147,7 @@ public:
     std::map<uint64_t, llvm::Value*> llvm_symtab_fn_arg;
     std::map<uint64_t, llvm::BasicBlock*> llvm_goto_targets;
     std::set<uint32_t> global_string_allocated;
+    std::vector<ASR::symbol_t*> struct_type_sym_alloca_in_branch;
     const ASR::Function_t *parent_function = nullptr;
 
     std::vector<llvm::BasicBlock*> loop_head; /* For saving the head of a loop,
@@ -4405,6 +4406,7 @@ public:
             llvm::APInt(64, get_class_hash(struct_type_sym)));
         builder->CreateStore(struct_type_hash, struct_type_hash_ptr);
         type2vtab[struct_type_sym][symtab] = vtab_obj;
+        struct_type_sym_alloca_in_branch.push_back(struct_type_sym);
         ASR::symbol_t* struct_type_ = struct_type_sym;
         bool base_found = false;
         while( !base_found ) {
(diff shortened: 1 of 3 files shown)


jinangshah21:
I was working on this one only. It was just due to updation a common variable which was causing this. Below diff in `asr_to_llvm` will solve it and this will create alloca for all v_tabs at starting of function only so in `if`/`else` conditions we will use that common v_tab.

```diff
@@ -4520,13 +4841,20 @@ public:
                     bool is_vtab_needed = false;
-                    while( !is_vtab_needed && struct_type ) {
-                        if( struct_type == class_sym ) {
+                    ASR::symbol_t* temp_struct_type = struct_type;
+                    while( !is_vtab_needed && temp_struct_type ) {
+                        if( temp_struct_type == class_sym ) {
                             is_vtab_needed = true;
                         } else {
-                            struct_type = ASR::down_cast<ASR::Struct_t>(
+                            temp_struct_type = ASR::down_cast<ASR::Struct_t>(
                                 ASRUtils::symbol_get_past_external(struct_type))->m_parent;
                         }
                     }
```
 
I think fix done in this PR will allocate `v_tabs` in every if else condition?

Pranavchiku:
> I think fix done in this PR will allocate v_tabs in every if else condition?

Yup, it will do it in every branch change. When I now think of it considering `block`, `entry`, etc ( although old ), we may need this. 

I'm fine either ways, we will be refactoring all this.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7586,9 +7588,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            // remove all `struct_type_sym` from `type2vtab`
+            for (ASR::symbol_t* sym : struct_type_sym_alloca_in_branch) {
+                if (type2vtab.find(sym) != type2vtab.end()) {
+                    type2vtab.erase(sym);
+                }
+                if (class2vtab.find(sym) != class2vtab.end()) {
+                    class2vtab.erase(sym);
+                }
+            }
+            struct_type_sym_alloca_in_branch = struct_type_sym_alloca_in_branch_copy;
```
certik, line 7605: This seems quite inefficient. If @jinangshah21's solution does the same thing, I would go with his.

certik:
@jinangshah21 can you submit a PR for this?

jinangshah21:
Opened PR: #8286

certik:
This was fixed in https://github.com/lfortran/lfortran/pull/8286.

Issue #7949, opened 2025-07-02
bug: initialisation of pointer variable during declaration not allowed


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main

   type :: my_type
      integer :: x
   end type my_type

   type(my_type), target :: my_var
   type(my_type), pointer :: p => my_var
end program main
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
(lf) saurabhkumar@Mac lfortran % lfortran test.f90
semantic error: Named initialization not supported with: my_var
 --> test.f90:8:4
  |
8 |    type(my_type), pointer :: p => my_var
  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #8271, opened 2025-08-07
fix: allow pointer initialization during variable declaration

HiteshGhanchi:
#7949 
I want to add this testcase, but where do I add it.
I see different ways in the repo. One is inside the tests folder and add it in tests.toml file , other is in integration_tests folder.
How is it decided which tests are added where and where do I add the MRE in the Issue.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2634,3 +2634,4 @@ RUN(NAME union_test_03 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 
 # LFortran extensions (unsigned int)
 RUN(NAME test_unsigned LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran 
+RUN(NAME associate_21 LABELS gfortran llvm)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1691,6 +1691,7 @@ RUN(NAME associate_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES a
 RUN(NAME associate_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME associate_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME associate_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES associate_21_mod.f90)
+RUN(NAME associate_22 LABELS gfortran llvm)
 RUN(NAME attr_dim_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME attr_dim_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME attr_dim_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
@@ -2634,4 +2635,3 @@ RUN(NAME union_test_03 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 
 # LFortran extensions (unsigned int)
 RUN(NAME test_unsigned LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran 
-RUN(NAME associate_21 LABELS gfortran llvm)
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4066,12 +4066,53 @@ public:
                             }
 
                         } else {
-                            diag.add(Diagnostic(
-                                "Named initialization not supported with: " + sym_name,
-                                Level::Error, Stage::Semantic, {
-                                    Label("",{x.base.base.loc})
-                                }));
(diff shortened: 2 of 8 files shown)


swamishiju:
You can put the MRE in a program, put it in the integration tests folder and add a test in integration_test/CMakeList.txt to register it

integration_tests/CMakeLists.txt
```diff
@@ -2617,3 +2617,4 @@ RUN(NAME lp_union_test_02 LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 
 # LFortran extensions (unsigned int)
 RUN(NAME test_unsigned LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran 
+RUN(NAME pointer_assignment_01 LABELS gfortran llvm)
```
Pranavchiku:
```suggestion
RUN(NAME associate_xx LABELS gfortran llvm)
```

Let's have this name

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4062,12 +4062,46 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             }
 
                         } else {
-                            diag.add(Diagnostic(
-                                "Named initialization not supported with: " + sym_name,
-                                Level::Error, Stage::Semantic, {
-                                    Label("",{x.base.base.loc})
-                                }));
-                            throw SemanticAbort();
+                            ASR::symbol_t *target_sym = current_scope->resolve_symbol(sym_name);
```
Pranavchiku, line 4069: Why do we need to put it inside `else` block?

Pranavchiku:
```
(Declaration
            (AttrType
                TypeType
                []
                ()
                my_type
                None
            )
            [(SimpleAttribute
                AttrPointer
            )]
            [(p
            []
            []
            ()
            my_var
            Arrow
            ())]
            ()
        )
```

Shouldn't `my_var` be in `Name_t`? I think something is off here.

HiteshGhanchi: > Why do we need to put it inside `else` block?

The if/else structure here is designed to handle all initializers that are an AST::Name_t. The if branch handles the single special intrinsic name, c_null_ptr (for => null()). Then the else block is handling all the other general cases. Hence I put it there.

HiteshGhanchi:
I had to change the code because the `check_equal_type()` was changed in a recent commit.
Everything else works. Also I have implemented the changed asked.

certik:
@HiteshGhanchi can you create a good title describing what bug was fixed and how?

HiteshGhanchi:
> @HiteshGhanchi can you create a good title describing what bug was fixed and how?

The pr resolves the bug mentioned in #7949.
The code initially handled the case of `p => null()` (pointer assignment to a null pointer), and in the else case gave an error saying `named initialization is not supported`.
I added the logic to handle general pointer assignment, also the asr_verify check was modified to not throw error for this error.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4104,10 +4145,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             }));
                         throw SemanticAbort();
                     }
-
+
```
certik, line 4145: Don't add whitespace.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4104,10 +4145,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             }));
                         throw SemanticAbort();
                     }
-
+                    
                     value = ASRUtils::expr_value(init_expr);
                     if ( init_expr ) {
                         if( ASRUtils::is_value_constant(value) ) {
+                        } else if (is_pointer) {
                         } else if( ASRUtils::is_value_constant(init_expr) ) {
```
certik, line 4150: This looks like an empty if statement --- can you add a comment inside `// do nothing, because ....` and explain why.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4066,12 +4066,53 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            ASR::symbol_t* target_sym_unwrapped = ASRUtils::symbol_get_past_external(target_sym);
+                            if (!ASR::is_a<ASR::Variable_t>(*target_sym_unwrapped)) {
+                                diag.add(Diagnostic("'" + sym_name + "' is not a variable and cannot be a pointer target",
+                                    Level::Error, Stage::Semantic, {Label("not a variable", {s.m_initializer->base.loc})}));
+                                throw SemanticAbort();
+                            }
+                            ASR::Variable_t *target_var = ASR::down_cast<ASR::Variable_t>(target_sym_unwrapped);
+
+                            if (!target_var->m_target_attr && !ASRUtils::is_pointer(target_var->m_type)) {
+                                diag.add(Diagnostic("The target of '=>' must have the TARGET or POINTER attribute",
```
certik, line 4086: Use lowercase in diagnostic messages.

certik (commented):
Can you also please add tests for every new compiler error / warning that you added?

@Pranavchiku do you mind please reviewing this?

PR #8274, opened 2025-08-07
testing CI triggering on specific label

assem2002:


assem2002:
Exhaustive checks must run and pass, as they're required.

certik:
The failure is unrelated: https://github.com/lfortran/lfortran/issues/8285.

certik:
When rebasing with a button, it will not run the exhaustive checks even though the label is on. I must manually remove the label and add it again.

PR #8279, opened 2025-08-07
ref: simplify vtab from `type {i64}` to `i64`

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/7068


Pranavchiku:
Please do not merge this PR! This makes the `__vtab` difficult to understand. Previous PRs used `__vtab_xx` while bitcasting and this now does not making it difficult to understand usage of this.

PR #8282, opened 2025-08-07
chore: use `Union` symbol for type checking

kmr-srbh:


certik:
Do you mind please rebasing this on top of the latest main? I merged a quick fix, to get us unblocked, and your fix is probably better.

Issue #5901, opened 2025-01-02
allocatable scalar regression


harperjf:
This program:
```
module integertostring
  implicit none
  integer :: n(2) = [42,-1]
contains
  pure function i0(n)
    character(:),allocatable :: i0
    integer,intent(in)       :: n
    character(range(n)+2)    :: result
    write(result,"(I0)") n ! write(i0, ...) with i0 unallocated is bad f2018
    i0 = trim(adjustl(result))
  end function i0
end module integertostring
program initialtest
! Does gfortran warn when it shouldn't with an allocatable scalar?
  use integertostring, only: i0, n
  implicit none
  character(:),allocatable:: string
  string = 'The answer is '//merge(i0(n(1)),i0(n(2)),n(1)>0)//'.'
  print *,'"'//string//'"'
end program initialtest
```
gave this output with lfortran 0.42.0-835-ge375d9063 :
```
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```
With lfortran 0.42.0 it gave this correct output:
```
 "The answer is 42."
```

harperjf:
Here is a shorter program that gives this bug in Linux Ubuntu. I hope it helps in the debugging:
```
program whatprec
  character(20) fmt
  write(fmt,"(A,I0,A)") '(A,I0,A,F0.',precision(1d0),')'
  print *,'fmt is ',trim(fmt)
end program whatprec
```
Lfortran 0.42.0-835-ge375d9063 output:
```
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```
Gfortran 14.2.0 output:
```
 fmt is (A,I0,A,F0.15)
```

HarshitaKalani:
I'm on Linux Ubuntu and tested it with the version 0.42.0-835-ge375d9063 - gives me correct output though.
```console
$ lfortran --version
LFortran version: 0.42.0-835-ge375d9063
Platform: Linux
LLVM: 11.0.1
Default target: x86_64-unknown-linux-gnu
$ lfortran a.f90
"The answer is 42."
```

tanay-man:
With the latest main, I too am getting the correct output.

gxyd:
We can probably just add a test case for this to LFortran and close this issue. Tagging this issue as *easy to fix*

krystophny:
This was fixed by #8100 (merged 2025-07-20), which fixed printing for allocatable scalars.

[...]

PR #8287, opened 2025-08-08
test: added test print_arr_08

shubham050802:
fixes: #5901


integration_tests/CMakeLists.txt
```diff
@@ -356,6 +356,7 @@ RUN(NAME print_arr_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm2 fortra
 RUN(NAME print_arr_04 LABELS llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --apply-fortran-mangling)
 RUN(NAME print_arr_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME print_arr_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran mlir)
+RUN(NAME print_arr_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
```
assem2002, line 359: Modify test registration, as CI is failing with Fortran backend.
```suggestion
RUN(NAME print_arr_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```

assem2002 (approved):
Looks good. Thanks for fixing that.

integration_tests/CMakeLists.txt
kmr-srbh: I was looking at that too! :) Here is a patch https://github.com/lfortran/lfortran/pull/8287#pullrequestreview-3100717628 for the `fortran` back-end. We should not add the `C` back-end label as you rightly mentioned above, as it fails too.

PR #8289, opened 2025-08-08
refactor: _lfortran_string_read_i32

sriganeshres:
towards: #8241

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5082,7 +5082,11 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
 }
 
 LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format, int32_t *i) {
-    sscanf(str, format, i);
+    char* tmp = (char*)malloc(len + 1);
+    strncpy(tmp, str, len);
+    tmp[len] = '\0';
+    sscanf(tmp, format, i);
+    free(tmp);
 }
 
 LFORTRAN_API void _lfortran_string_read_i64(char *str, int64_t len, char *format, int64_t *i) {


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5082,19 +5082,35 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
...
+    strncpy(tmp, str, len);
+    tmp[len] = '\0';
+    sscanf(tmp, format, i);
+    free(tmp);
 }
 
 LFORTRAN_API void _lfortran_string_read_f32(char *str, int64_t len, char *format, float *f) {
-    sscanf(str, format, f);
+    char* tmp = (char*)malloc(len + 1);
+    strncpy(tmp, str, len);
```
assem2002: Don't we have any other way to handle this function without using C-style strings?

sriganeshres: We can but we need to use a bounded writer. Do you want me to do that.

sriganeshres: I am doing that.

assem2002: Can you explain further? I tried to pass the problem to some GPTs and obviously the options out there aren't different (performance wise) from what we have here.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5082,19 +5082,35 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
 }
 
 LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format, int32_t *i) {
-    sscanf(str, format, i);
+    char* tmp = (char*)malloc(len + 1);
+    strncpy(tmp, str, len);
```
assem2002: Can you please use `memcpy`?, as `strncpy` has implication that it stops whenever encounters a `\0` and Fortran strings can have such characters as part of the string.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5082,19 +5106,165 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
...
+    int sign = 1;
+    if (p < end && *p == '-') { sign = -1; p++; }
+    else if (p < end && *p == '+') { p++; }
+
+    int32_t value = 0;
+    while (p < end && isdigit((unsigned char)*p)) {
+        value = value * 10 + (*p - '0');
+        p++;
+    }
+    *i = value * sign;
```
certik: I don't like this. From here and below we are expanding a single line `sscanf` into 20+ lines, copy & pasting the same code (possibly with some tiny changes --- hard to tell) all over.

My first question is: what is the motivation, is sscanf not enough for what we need? Is there some Fortran test case that didn't work in main, but now works? If so, the test should be added in this PR.

If the answer is that it is not good enough, then we need to refactor all this code into functions to parse integers, don't repeat it. And then we need to add tests for these parsing routines, as it is easy to make mistakes in them, and forget about some corner case, like an empty string, invalid string, etc. The best way to test it is to call the runtime from our `ctest` infrastructure and just add tests specifically for the parsing. Then use those well-tested functions here.

sriganeshres: No I wanted to use non c style strings that why i mdid not make any tmp C style strings and made it to mimic sscanf as asked here https://github.com/lfortran/lfortran/pull/8289#discussion_r2262818697. If you dont want this and just change `strncpy` to `memcpy` I can do that.

sriganeshres: No problem i have changed `strncpy` to `memcpy`

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5082,19 +5106,40 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
 }
 
 LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format, int32_t *i) {
-    sscanf(str, format, i);
+    char *buf = (char*)malloc(len + 1);
+    if (!buf) return; // handle allocation failure if needed
```
certik: What is the purpose of this comment --- either handle it, or explain why it doesn't need to be handled.

sriganeshres: it is just a malloc error. Comment is nothing we can ignore that.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5082,19 +5106,40 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
...
 }
 
 LFORTRAN_API void _lfortran_string_read_f32(char *str, int64_t len, char *format, float *f) {
-    sscanf(str, format, f);
+    char *buf = (char*)malloc(len + 1);
+    if (!buf) return;
+    memcpy(buf, str, len);
+    buf[len] = '\0';
+    sscanf(buf, format, f);
+    free(buf);
```
certik, line 5133: This should be completely redesigned: https://github.com/lfortran/lfortran/issues/8310, but let's do it in a subsequent PR.

src/libasr/runtime/lfortran_intrinsics.c
certik: Then remove it. Don't add comments that are "nothing".

```suggestion
    if (!buf) return;
```

PR #8297, opened 2025-08-08
fix: do not return struct symbol past external 

### LINKED ISSUE #8291: Compilation of `associate` statement leads to late compiler segfault
This seems to be a regression after some of the recent refactoring. The following test code, made up of these three files, 
 `$ cat module1.F90`
```
module module1

   type, abstract :: AbsType
   end type AbsType

end module module1
```

 `$ cat module2.F90`
```
module module2

   use module1, only: AbsType
   
   type :: MyType
      class(AbsType), allocatable :: arr(:)
   contains
      procedure :: method
   end type MyType

contains
   
   subroutine method(self)
      class(MyType), intent(inout) :: self
      associate ( element => self%arr(size(self%arr)) )
      end associate
   end subroutine method

end module module2
```
`$ cat module3.F90` 
```
module module3

   use module2, only: MyType

end module module3
```
gives the following behavior upon compilation:
`$ lfortran --cpp -c module1.F90`
`$ lfortran --cpp -c module2.F90 `
`$ lfortran --cpp -c module3.F90`

A segfault happens during the compilation of `module3.F90`, triggered by an `associate` statement (that operates on an array) which is used within `module2.F90`.

kmr-srbh:
Resolves #8291


kmr-srbh:
Some more cases need to have `ASRUtils::symbol_get_past_external()` removed from the return statement. We should not return the symbol past external as it may be outside our scope. I will make sure to correct the rest of them later.

PR #8301, opened 2025-08-09
LP Sync

swamishiju:
- Fix for LP `str()`


certik:
Let's figure out tests to add into LFortran for all these fixes, otherwise they will get broken again.

swamishiju:
> Let's figure out tests to add into LFortran for all these fixes, otherwise they will get broken again.

Most of the stuff in this PR is a bit awkward to test for in LF. The easiest is probably the unsigned arrays - but others like `int->str` have completely different nodes in Fortran

certik:
Can you please fix this PR, I think it has unrelated changes.

Regarding int->str, we should be using exactly the same nodes in both LFortran and LPython, and add tests in LFortran.

swamishiju:
Closed in favor of #8418

Issue #7937, opened 2025-06-30
Refactor all compiler options


certik:
We need to decide:

* [ ] if we stick with `-` for single options and `--` for longer options, or use the technique `-fcheck=all` style in gcc/clang
* [ ] how to group all options into classes, such as: warnings, optimization, code generation, language features, etc. (For example `-W`, `-f`, `-O`, etc., depending on the previous point)
* [ ] Add "no" option for every option that enables something, so that one can for example do `--std=f23`, but then turn off some options
* [ ] Warnings need to be more granular (related to https://github.com/lfortran/lfortran/issues/564)

We should do this before beta.

(I thought we already had some issue for this, but I can't find it.)

certik:
For reference, here is how other compilers do it:

* MSVC: https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-by-category?view=msvc-170
* Clang: https://clang.llvm.org/docs/ClangCommandLineReference.html
* GCC: https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html
* ifx: https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-0/alphabetical-option-list.html
* Classic Flang: https://github.com/flang-compiler/flang/wiki/Using-Flang
* Cray compiler: https://docs.lumi-supercomputer.eu/development/compiling/cce/
* NAG compiler: https://support.nag.com/nagware/np/r71_doc/manual/compiler_2_4.html
* Zig: https://ziglang.org/learn/build-system/
* Rust: https://doc.rust-lang.org/beta/rustc/command-line-arguments.html
* Go: https://pkg.go.dev/cmd/compile

PR #8308, opened 2025-08-09
add enable style compiler option

parth121101:
https://github.com/lfortran/lfortran/issues/7937#issue-3188997915

This PR is the initial PR to address the third point of the above comment.
If someone wants to use `--std=f23` flag with style suggestion enable, then this is the flag that can be used. 
I have tested this. Following are the results.
Using only `--std=f23` flag would not show style suggestions. 

```
lfortran /Users/parth1211/repos/lfortran/tests/save4.f90 --std=f23
warning: Assuming implicit save attribute for variable declaration
 --> /Users/parth1211/repos/lfortran/tests/save4.f90:9:20
  |
9 |         integer :: x = 5
  |                    ^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement


 y is 1
 ```
 
 With ` --style-warnings` flag one can use `--std=f23` flag and still can see style warnings
 
 ```
 lfortran /Users/parth1211/repos/lfortran/tests/save4.f90 --std=f23 --style-warnings
style suggestion: Use '==' instead of '.eq.'
 --> /Users/parth1211/repos/lfortran/tests/save4.f90:4:11
  |
4 |     if (y .eq. 1) then
  |           ^^^^ help: write this as '=='

style suggestion: Use '==' instead of '.eq.'
  --> /Users/parth1211/repos/lfortran/tests/save4.f90:10:15
   |
10 |         if (x .eq. 1) then
   |               ^^^^ help: write this as '=='

warning: Assuming implicit save attribute for variable declaration
 --> /Users/parth1211/repos/lfortran/tests/save4.f90:9:20
  |
9 |         integer :: x = 5
  |                    ^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement


 y is 1
 ```

### CODE REVIEW

File: src/libasr/utils.h [hunk 1]
```diff
@@ -113,6 +113,7 @@ struct CompilerOptions {
     bool generate_code_for_global_procedures = false;
     bool no_warnings = false;
     bool disable_style = false;
+    bool enable_style = false;
```
certik: Don't add both flags. Just use one option `style_warnings = true/false` and that's it.

parth121101: Due you also want me to remove this two flag as well and combine them. 
```
app.add_flag("--no-style-warnings", compiler_options.disable_style, "Turn off style suggestions")->group(group_warning_options);
app.add_flag("--style-warnings", compiler_options.enable_style, "Enable style suggestions")->group(group_warning_options);
```

Or you want me to remove enable_style from compiler options and just keep one compiler option for style warning ?

parth121101: https://github.com/lfortran/lfortran/pull/8308/commits/2b63de3a67318915ddb6152334d2215fc3130556
For this commit I have removed the enable_style that I recently added in compiler option.

parth121101: I have removed the enable_style option from compiler option. 
Now we have only single variable for compiler option and keeping two on and off flags for command line.

parth121101:
Added a simple test to validate the implemented functionality.

HarshitaKalani (approved):
lgtm

certik:
I had to revert this due to https://github.com/lfortran/lfortran/issues/8326.

PR #8309, opened 2025-08-09
Tests : Trigger workflow with changes in commit history (rebase, new commits, ..)

assem2002:
This PR just handles the cases of (PR sync, reopen, unlabel, once opened).
The case of having label `Tests:Run-Exhaustive` already set, and then adding another random label then the workflow happens to get triggered again unnecessarily **isn't handled in this PR**
as well the case of removing some random label will also the trigger the workflow to run unnecessarily, It's not a bit deal now as these cases doesn't happen so frequently (Our dependency on labels isn't intensive yet).

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -9,6 +9,10 @@ on:
   pull_request:
     types: 
       - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - reopened
     branches:
       - main
 


certik:
This will run exhaustive checks on opening a PR?

And on removing any label? That seems we don't need that, right?

certik:
How about this: https://grok.com/share/bGVnYWN5_e1e55f11-ec35-4b80-b362-c292a46d7732

Add the following to every exhaustive job:
```
if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-exhaustive') }}
```

That should do it, right?

assem2002:
> How about this: https://grok.com/share/bGVnYWN5_e1e55f11-ec35-4b80-b362-c292a46d7732
> 
> Add the following to every exhaustive job:
> 
> ```
> if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-exhaustive') }}
> ```
> 
> That should do it, right?

if we use that with the general workflow trigger we used before, it will work as we intend yeah. 
Here : 
```yml
on:
  pull_request:
    branches:
      - main
```
++
```yml
if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-exhaustive') }}
```

**Update** : 
the general trigger only triggers (open, sync) it doesn't trigger the labeling activity.
I think what we have now in the PR is good enough, and it's also explicit.

assem2002:
> This will run exhaustive checks on opening a PR?
> 
> And on removing any label? That seems we don't need that, right?

let's assume the case where you want to stop the tests from running.

certik:
You have to read the full answer in Grok. You still need:
```
types:
      - labeled
```
And maybe also `unlabeled` to stop it from running, but I don't think this cancels it, just doesn't run it if it didn't start yet. But that might never happen, as usually it "starts" right away, and waits, at least I think.

So just start with `labeled`.

assem2002:
Oh, Yeah Grok's answer mentioned my concern.
So now things got confusing, Can you list the requirements we want to achieve?

certik:
Ok, I just did my idea myself: https://github.com/lfortran/lfortran/pull/8315. I think that's what Grok had in mind.

It turns out you already had an "if" statement there with a check for the label. Why didn't it work? Is it because the workflow only triggers on new labels being added?

So this PR here adds more cases to trigger the run, and it then gets narrowed down with the if statement?

assem2002:
> It turns out you already had an "if" statement there with a check for the label. Why didn't it work? Is it because the workflow only triggers on new labels being added?

Yes, this PR adds more activity triggers (open, repopen, etc.).
The If statement (in current main) is responsible of making the jobs run if and only if (label=`Tests::Run-Exhaustive`|| event=`push`). 
So the triggering activity whenever happens it invokes the workflow, and then we do the checks we just mentioned.

For now I think this PR fulfills most of what we want to achieve. (proper run of the exhaustive checks as long as the label is set)
There are some minor issues. For example, Adding random label like `bug` meanwhile the label `Tests::Run-Exhaustive` was already there and the tests were already running, that would invoke the workflow again and run the tests again totally unnecessarily; But these are minor ones that we won't face a lot.

assem2002:
BTW, The `open` trigger made the CI in this PR have this ugly (yet informative) `skipped` jobs.

PR #8311, opened 2025-08-09
refactor: make _lfortran_inquire null independent

akramhany:
Towards #8241

refactors `_lfortran_inquire`.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -3573,8 +3573,7 @@ trim_trailing_spaces(char** str, int64_t* len, bool init)
 }
 
 
-static char*
-to_c_string(char* src, int64_t len)
+static char* to_c_string(const fchar* src, int64_t len)
 {
     char* buf = (char*) malloc(len + 1);
     if (!buf)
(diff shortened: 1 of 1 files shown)


Pranavchiku (approved):

assem2002 (approved):
Looks good to me. Thanks.

assem2002:
Would you mind using that same `make_c_string` function in this PR https://github.com/lfortran/lfortran/pull/8280/files that you've merged lately, It doesn't actually use null-terminated C-strings like you do in this PR.

akramhany:
> Would you mind using that same `make_c_string` function in this PR https://github.com/lfortran/lfortran/pull/8280/files that you've merged lately, It doesn't actually use null-terminated C-strings like you do in this PR.

Where exactly? If it is in the `parse_fortran_format` when I converted the `fchar*` to `char*`, then I didn't make it null-terminated because inside the function we don't do something that require the string to be null-terminated, as everything is done on length, but indeed it would be better to make it null-terminated, so I will do that.
(please tell me if this is not what you meant)

assem2002:
Oh, I see. So yeah that's fine, though a variable named `C_*` implies a null-terminated string, but that's fine for now. 
No need for changes now.

akramhany:
I noticed that we were creating C strings in `_lfortran_open` but not using them, so I used them instead of the Fortran strings. 
I also fixed an error that was in the `_lfortran_open` function where we didn't update the length of the `f_name`.

akramhany:
@assem2002 Can you please run the exhaustive checks again?

PR #8315, opened 2025-08-10
CI: Conditions for running exhaustive checks

certik:


Pranavchiku:
This only checks Quick checks, which is good! Is this PR meant for further testing?

certik:
I don't know if we want it or not. The CI still doesn't (or didn't) behave properly, see https://github.com/lfortran/lfortran/issues/8350.

PR #8329, opened 2025-08-12
Better derived type error message

certik:
Now it prints:
```console
$ lfortran der.f90
  --> der.f90:15:1
   |
15 | x = y
   | ^   ^ type mismatch (derived type 'a' and derived type)


```
For:
```fortran
program expr2
implicit none

type :: A
    integer :: aa
end type

type :: B
    integer :: bb
end type

type(A) :: x
type(B) :: y

x = y

end program
```

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -1744,8 +1744,8 @@ public:
                 ASR::ttype_t* source_type = ASRUtils::expr_type(source);
                 ASR::ttype_t* var_type = ASRUtils::expr_type(alloc_args_vec.p[i].m_a);
                 if (!ASRUtils::check_equal_type(source_type, var_type, source, alloc_args_vec.p[i].m_a)) {
-                    std::string source_type_str = ASRUtils::type_to_str_fortran(source_type, source);
-                    std::string var_type_str = ASRUtils::type_to_str_fortran(var_type, alloc_args_vec.p[i].m_a);
+                    std::string source_type_str = ASRUtils::type_to_str_fortran_expr(source_type, source);
+                    std::string var_type_str = ASRUtils::type_to_str_fortran_expr(var_type, alloc_args_vec.p[i].m_a);
                     diag.add(Diagnostic(
                         "Type mismatch: The `source` argument in `allocate` must have the same type as the allocated variable.\n"
                         "Expected type: " + var_type_str + ", but got: " + source_type_str + ".",
@@ -1883,7 +1883,7 @@ public:
             } else {
                 diag.add(Diagnostic(
                     "Cannot deallocate variables in expression " +
-                    ASRUtils::type_to_str_python(ASRUtils::expr_type((tmp_expr)), tmp_expr),
+                    ASRUtils::type_to_str_python_expr(ASRUtils::expr_type((tmp_expr)), tmp_expr),
                     Level::Error, Stage::Semantic, {
                         Label("",{tmp_expr->base.loc})
                     }));
@@ -3152,8 +3152,8 @@ public:
                                         ASRUtils::expr_type(value),ASRUtils::expr_type(to_return), diag);
         if (!ASRUtils::check_equal_type(ASRUtils::expr_type(to_return),
                                     ASRUtils::expr_type(value), to_return, value)) {
-            std::string ltype = ASRUtils::type_to_str_fortran(ASRUtils::expr_type(to_return), to_return);
-            std::string rtype = ASRUtils::type_to_str_fortran(ASRUtils::expr_type(value), value);
+            std::string ltype = ASRUtils::type_to_str_fortran_expr(ASRUtils::expr_type(to_return), to_return);
+            std::string rtype = ASRUtils::type_to_str_fortran_expr(ASRUtils::expr_type(value), value);
             diag.semantic_error_label(
                 "Type mismatch in statement function, the types must be compatible",
                 {to_return->base.loc, value->base.loc},
@@ -3473,8 +3473,8 @@ public:
                     }
                 }
             }
-            std::string ltype = ASRUtils::type_to_str_fortran(ASRUtils::expr_type(target), target);
-            std::string rtype = ASRUtils::type_to_str_fortran(ASRUtils::expr_type(value), value);
+            std::string ltype = ASRUtils::type_to_str_fortran_expr(ASRUtils::expr_type(target), target);
+            std::string rtype = ASRUtils::type_to_str_fortran_expr(ASRUtils::expr_type(value), value);
             if (!ASRUtils::check_equal_type(ASRUtils::expr_type(target),
                                         ASRUtils::expr_type(value), target, value)) {
                 if(value->type == ASR::exprType::ArrayConstant) {
@@ -3700,8 +3700,8 @@ public:
                 ASR::ttype_t *arg_type = ASRUtils::expr_type(args[1]);
                 ASR::ttype_t *contained_type = ASRUtils::get_contained_type((ASR::ttype_t *)list_type);
                 if (!ASRUtils::check_equal_type(contained_type, arg_type, nullptr, nullptr)) {
-                    std::string contained_type_str = ASRUtils::type_to_str_fortran(contained_type, (ASR::expr_t*)nullptr);
-                    std::string arg_type_str = ASRUtils::type_to_str_fortran(arg_type, (ASR::expr_t*)nullptr);
+                    std::string contained_type_str = ASRUtils::type_to_str_fortran_expr(contained_type, nullptr);
+                    std::string arg_type_str = ASRUtils::type_to_str_fortran_expr(arg_type, nullptr);
                     diag.add(Diagnostic(
                         "Type mismatch in " + var_name + ", the types must be compatible",
                         Level::Error, Stage::Semantic, {
@@ -3726,8 +3726,8 @@ public:
                 ASR::ttype_t *contained_type = ASRUtils::get_contained_type(ASRUtils::expr_type(args[0]));
 
                 if (!ASRUtils::check_equal_type(contained_type, ASRUtils::expr_type(args[2]), nullptr, nullptr)) {
-                    std::string contained_type_str = ASRUtils::type_to_str_fortran(contained_type, (ASR::expr_t*)nullptr);
-                    std::string arg_type_str = ASRUtils::type_to_str_fortran(ASRUtils::expr_type(args[2]), (ASR::expr_t*)nullptr);
+                    std::string contained_type_str = ASRUtils::type_to_str_fortran_expr(contained_type, nullptr);
+                    std::string arg_type_str = ASRUtils::type_to_str_fortran_expr(ASRUtils::expr_type(args[2]), nullptr);
                     diag.add(Diagnostic(
                         "Type mismatch in " + var_name + ", the types must be compatible",
                         Level::Error, Stage::Semantic, {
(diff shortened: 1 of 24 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3521,6 +3521,15 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                 } else {
+                        ASR::ttype_t *target_type = ASRUtils::expr_type(target);
+                        if (ASR::is_a<ASR::StructType_t>(*target_type)) {
+                            ASR::symbol_t *target_symbol = ASRUtils::get_struct_sym_from_struct_expr(target);
+                            LCOMPILERS_ASSERT(target_symbol);
+                            LCOMPILERS_ASSERT(ASR::is_a<ASR::Struct_t>(*target_symbol));
+                            ASR::Struct_t *target_Struct = ASR::down_cast<ASR::Struct_t>(target_symbol);
+                            std::string target_struct_name = target_Struct->m_name;
+                            ltype = "derived type '" + target_struct_name + "'";
+                        }
```
certik: @kmr-srbh refactor this into a dedicated function in ASR utils.

kmr-srbh:
Yes. I will complete this PR.

src/libasr/asr_utils.h
```diff
@@ -878,7 +880,10 @@ static inline void encode_dimensions(size_t n_dims, std::string& res,
     }
 }
 
-static inline std::string type_to_str_fortran(const ASR::ttype_t *t)
+// Templated argument `struct_type_name_helper` should be either an `ASR::expr_t*` or
+// `ASR::symbol_t*`.
```
certik: Why cannot this always be `symbol_t`?

certik:
The CI fails.

Cannot we do this step by step? The original PR was almost ready, just a small refactoring.

Related: https://github.com/lfortran/lfortran/issues/8355.

src/libasr/asr_utils.h
kmr-srbh: We mostly have an `ASR::expr_t*` available, so we use that. But for some cases, where it was not available, I have overloaded this function with a `ASR::symbol_t*` variant. This variant is also used a lot. I have made this change locally. I will push it with the rest of the changes.

kmr-srbh:
> Cannot we do this step by step? The original PR was almost ready, just a small refactoring.

The change is simple actually. We have either a symbol or expression every where. I am almost done with refactoring `type_to_str_python()` also. I am pushing the changes in a while.

kmr-srbh:
The LSP regression is also now resolved.

kmr-srbh:
Similar to the changes done for `types_equal()`, I have passed `nullptr` for cases where we do not encounter a `StructType`.

src/libasr/codegen/c_utils.h
```diff
@@ -334,7 +334,7 @@ namespace CUtils {
                 break;
             }
             default: {
-                throw CodeGenError("Type " + ASRUtils::type_to_str_python(t) + " not supported yet.");
+                throw CodeGenError("Type " + std::to_string(t->type) + " not supported yet.");
```
certik, line 337: Later: at least print the ttype name as a string, not a number.

src/libasr/asr_utils.h
```diff
@@ -530,14 +532,17 @@ static std::string intent_to_str(ASR::intentType intent) {
     }
 }
 
-static inline std::string type_to_str_fortran(const ASR::ttype_t *t);
+static inline std::string type_to_str_fortran(const ASR::ttype_t *t, ASR::expr_t* expr);
```
certik:
```suggestion
static inline std::string type_to_str_fortran_expr(const ASR::ttype_t *t, ASR::expr_t* expr);
```

src/libasr/asr_utils.h
```diff
@@ -530,14 +532,17 @@ static std::string intent_to_str(ASR::intentType intent) {
...
+
+static inline std::string type_to_str_fortran(const ASR::ttype_t *t, ASR::symbol_t* sym);
```
certik:
```suggestion
static inline std::string type_to_str_fortran_symbol(const ASR::ttype_t *t, ASR::symbol_t* sym);
```

src/libasr/asr_utils.h
```diff
@@ -530,14 +532,17 @@ static std::string intent_to_str(ASR::intentType intent) {
     }
 }
 
-static inline std::string type_to_str_fortran(const ASR::ttype_t *t);
```
certik, line 533: Maybe implement this one using `type_to_str_fortran_symbol(t, nullptr)`?

certik: It turns out, this function is never needed, we always want to call either `type_to_str_fortran_expr` or `type_to_str_fortran_symbol`.

certik: So keep it removed.

src/libasr/asr_utils.h
```diff
@@ -878,7 +883,7 @@ static inline void encode_dimensions(size_t n_dims, std::string& res,
     }
 }
 
-static inline std::string type_to_str_fortran(const ASR::ttype_t *t)
+static inline std::string type_to_str_fortran(const ASR::ttype_t* t, ASR::expr_t* expr)
```
certik:
```suggestion
static inline std::string type_to_str_fortran_expr(const ASR::ttype_t* t, ASR::expr_t* expr)
```

src/libasr/asr_utils.h
```diff
@@ -878,7 +883,7 @@ static inline void encode_dimensions(size_t n_dims, std::string& res,
...
 {
```
certik, line 888: Cannot this function be implemented as:
```c++
ASR::symbol_t sym = nullptr;
if (ASR::is_a<ASR::StructType_t>(*expr)) {
    sym = ASRUtils::get_struct_sym_from_struct_expr(expr));
    // TODO: possibly this if statement is already present inside the function. Then it can be removed from here.
}
type_to_str_fortran_symbol(t, sym)
```

src/libasr/asr_utils.h
```diff
@@ -964,13 +969,92 @@ static inline std::string type_to_str_fortran(const ASR::ttype_t *t)
     }
 }
 
-static inline std::string type_to_str_with_type(const ASR::ttype_t *t) {
-    std::string type = type_to_str_fortran(t);
-    std::string kind = std::to_string(extract_kind_from_ttype_t(t));
-    return type + "(" + kind + ")";
+static inline std::string type_to_str_fortran(const ASR::ttype_t* t, ASR::symbol_t* sym)
+{
```
certik, line 973: Put assert here:
* if `ttype` is StructType, then `sym` is not null, it points to a valid symbol
* otherwise `sym` is nullptr

src/libasr/asr_utils.h
```diff
@@ -964,13 +969,92 @@ static inline std::string type_to_str_fortran(const ASR::ttype_t *t)
     }
 }
 
-static inline std::string type_to_str_with_type(const ASR::ttype_t *t) {
-    std::string type = type_to_str_fortran(t);
-    std::string kind = std::to_string(extract_kind_from_ttype_t(t));
-    return type + "(" + kind + ")";
+static inline std::string type_to_str_fortran(const ASR::ttype_t* t, ASR::symbol_t* sym)
```
certik:
```suggestion
static inline std::string type_to_str_fortran_symbol(const ASR::ttype_t* t, ASR::symbol_t* sym)
```

src/libasr/asr_utils.h
certik: And if this is the case, let's rename `sym` to `struct_sym`.

kmr-srbh:
Review suggestions addressed in https://github.com/lfortran/lfortran/pull/8329/commits/e01a2817bd0531829ed70b2897b1e392d6f4401d. :)

certik:
I think that https://github.com/lfortran/lfortran/pull/8329/files#diff-da3dce35379cdaaf286c90ae38fe3d27efda9a299f60c682dacc97ad1eca5ce7 is not addressed yet?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3819,8 +3819,8 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     ASR::ttype_t *contained_type = ASRUtils::get_contained_type((ASR::ttype_t *)list_type);
 
                     if (!ASRUtils::check_equal_type(contained_type, arg_type, nullptr, nullptr)) {
-                        std::string contained_type_str = ASRUtils::type_to_str_fortran(contained_type);
-                        std::string arg_type_str = ASRUtils::type_to_str_fortran(arg_type);
+                        std::string contained_type_str = ASRUtils::type_to_str_fortran_expr(contained_type, nullptr);
+                        std::string arg_type_str = ASRUtils::type_to_str_fortran_expr(arg_type, nullptr);
```
certik, line 3823: There are a lot of places that use this nullptr. Why not keep the original `type_to_str_fortran` and just implemented by calling `ASRUtils::type_to_str_fortran_expr(arg_type, nullptr)`? And furthermore, it should really call `ASRUtils::type_to_str_fortran_symbol(arg_type, nullptr)` instead, as that's the main function.

tests/server/tests/test_features.py
```diff
@@ -198,8 +198,7 @@ def test_document_hover(client: LFortranLspTestClient) -> None:
     assert doc.preview == Hover(
         contents=MarkupContent(
             kind=MarkupKind.Markdown,
-            # TODO: Reset string "derived type" to "softmax"
-            value="```fortran\nfunction eval_1d(self, x) result(res)\n    derived type, intent(in) :: self\n    real[:], intent(in) :: x\n    real[:] :: res\nend function\n```"
+            value="```fortran\nfunction eval_1d(self, x) result(res)\n    softmax, intent(in) :: self\n    real[:], intent(in) :: x\n    real[:] :: res\nend function\n```"
```
certik, line 201: This is not correct, I think it has to be `type(softmax)`?

certik:
I think it's good enough. I rebased it, let's see if tests pass. If so, let's merge it.

tests/server/tests/test_features.py
kmr-srbh: Yes, it should have been that. When I had encountered this failure, it was set to only `softmax`, so I set it back again. I do not know if the LSP currently supports writing `type(softmax)` and `class(softmax)`.

certik: Can you please open up an issue for this?

kmr-srbh:
> I think that https://github.com/lfortran/lfortran/pull/8329/files#diff-da3dce35379cdaaf286c90ae38fe3d27efda9a299f60c682dacc97ad1eca5ce7 is not addressed yet?

I am unable to access this comment somehow. Could you please post what it was? I am sorry.

src/lfortran/semantics/ast_body_visitor.cpp
kmr-srbh: So, we want a third function `type_to_str_fortran(ASR::ttype_t*)` which internally calls `type_to_str_fortran_expr(ttype, nullptr)`, right?

tests/server/tests/test_features.py
kmr-srbh: Done in #8363.

src/lfortran/semantics/ast_body_visitor.cpp
certik: That's what I was thinking -- it should simplify a lot of the code, no?

kmr-srbh: We eventually would need to pass an expression there actually. For example, when we support list of struct types, dictionary with struct type object as key or value, etc. We can add this TODO somewhere, likely in an issue, and implement the above idea suggested by you. Should I open up the issue and PR for this?

PR #8335, opened 2025-08-12
test: update error stop condition and uncomment test

HarshitaKalani:
Uncomment test failing after struct refactor.

### DIFF
--- a/integration_tests/intrinsics_211.f90
+++ b/integration_tests/intrinsics_211.f90
@@ -64,7 +64,7 @@ program intrinsics_211
     
     do i = 1, size(x)
         print *, bessel_yn(n(i), x(i)), "i = ", i
-        if (abs(bessel_yn(n(i), x(i)) - expected(i)) > 1e-12_dp) error stop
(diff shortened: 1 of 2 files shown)


integration_tests/intrinsics_211.f90
```diff
@@ -64,7 +64,7 @@ program intrinsics_211
     
     do i = 1, size(x)
         print *, bessel_yn(n(i), x(i)), "i = ", i
-        if (abs(bessel_yn(n(i), x(i)) - expected(i)) > 1e-12_dp) error stop
+        if (abs((bessel_yn(n(i), x(i)) - expected(i))/bessel_yn(n(i), x(i))) > 1e-12_dp) error stop
```
certik: Usually I would do:

```suggestion
        if (abs((bessel_yn(n(i), x(i)) - expected(i))/expected(i)) > 1e-12_dp) error stop
```
Since `expected` is already computed, while the function `bessel_yn` is not. But this is minor.

PR #8336, opened 2025-08-12
Add enable style compiler option

parth121101:
I will investigate the past failure in this PR.


tests/reference/asr_no_warnings-style1-3d16af1.stderr
```diff
@@ -0,0 +1,5 @@
+style suggestion: Use '==' instead of '.eq.'
+ --> tests/style1.f90:4:11
+  |
+4 |     if (y .eq. 1) then
+  |           ^^^^ help: write this as '=='
```
certik, line 5: Shouldn't this give no warnings?

parth121101: This are style warnings, so I think this should be there. 
One think I can add here is,
use --std=f23 flag along with --style-warnings flag so that we can validate that style warnings appears even tough we have used --std=f23.

parth121101: Now I have used `--std=f23` flag. The default style suggestion of this flag is set to `false`. But now with `--style-warnings` flag one can enable style suggestions keeping `--std=f23` flag.

certik:
I am still confused when it should and should not give warnings. Let's discuss tomorrow.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -216,6 +218,9 @@ namespace LCompilers::CommandLineInterface {
             // The default LFortran behavior, do nothing
         } else if (opts.arg_standard == "f23") {
             compiler_options.disable_style = true;
+            if (style_warnings) {
+                compiler_options.disable_style = false;
```
certik, line 222: Let's rename this to `show_style_warnings`. Later.

tests/tests.toml
```diff
@@ -3684,6 +3684,11 @@ filename = "save4.f90"
 asr_no_warnings = true
 asr_disable_style_and_warnings = true
 
+[[test]]
+filename = "style1.f90"
+asr_no_warnings = true
+asr_enable_style_warnings = true
```
```suggestion
asr_enable_style_suggestion = true
```

certik: Also add a third test to print a warning.

tests/style1.f90
```diff
@@ -0,0 +1,7 @@
+program style1
+    implicit none
+    integer :: y = 1
+    if (y .eq. 1) then
+        print *, "y is 1"
+    end if
```
certik, line 6: Add something here to print a warning.

certik: And then check that this warning is *not* printed for `asr_no_warnings=true` test. And is printed for `asr_enable_style_warnings=true` test.

run_tests.py
```diff
@@ -258,6 +259,19 @@ def is_included(backend):
             update_reference,
             verify_hash,
             extra_args)
+    
+    if asr_enable_style_warnings:
```
certik:
```suggestion
    if asr_enable_style_suggestion:
```

tests/style1.f90
parth121101: Updated the tests so that it has some warnings which is returned with `asr_enable_style_warnings=true` while it is not returned with `asr_no_warnings=true`.

tests/tests.toml
parth121101: `asr_enable_style_suggestion=true` itself throws a warning so the third test is not required i think.

src/bin/lfortran_command_line_parser.cpp
parth121101: Will address this in coming PRs.

parth121101: Creating an issue for this and will link a PR for it.

PR #8339, opened 2025-08-13
[OPENMP]DOC: OMPRegion ASR node description

adit4443ya:
Towards #7332
I'm open to ideas of what we should include.
CC:@certik @Pranavchiku


adit4443ya:
Yes, Now it works!!

Issue #8323, opened 2025-08-10
Implied-do wrong in a parameter, OK elsewhere


harperjf:
This program shows that implied-do is wrong in a parameter array constructor but right in a non-parameter array constructor or an I/O list when using lfortran.
```
program array_implied_do
  integer i
  integer, parameter:: nmax = 5, a(nmax) = [1,2,4,8,8], &
       b(nmax-1) = [ ( a(i)-a(i-1), i=2,nmax ) ]
  integer:: c(nmax-1) =  [ ( a(i)-a(i-1), i=2,nmax ) ]
  print "(A,5(1X,I0))", 'a =' ,a
  print "(A,4(1X,I0))", 'b = ',b
  print "(A,4(1X,I0))", 'c = ',c
  print "(A,4(1X,I0))", '? = ', ( a(i)-a(i-1), i=2,nmax )
```
Output from lfortran 0.55.0 (bad)
```
a = 1 2 4 8 8
b =  1 1 1 1
? =  1 2 4 0
c =  1 2 4 0
```
Output from gfortran, g95, ifx or AMD flang (good)
```
a = 1 2 4 8 8
b =  1 2 4 0
c =  1 2 4 0
? =  1 2 4 0
```

PR #8345, opened 2025-08-13
fix: handle compile-time evaluation of implied-do loops for parameter arrays

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8323

### DIFF
--- a/tests/reference/asr-implied_do_loops1-25d26db.stdout
+++ b/tests/reference/asr-implied_do_loops1-25d26db.stdout
@@ -141,17 +141,7 @@
                                                 (IntegerConstant 3 (Integer 4) Decimal)
                                                 ()
                                                 (Integer 4)
-                                                (ArrayConstant
-                                                    12
-                                                    [1, 2, 3]
-                                                    (Array
-                                                        (Integer 4)
-                                                        [((IntegerConstant 1 (Integer 4) Decimal)
-                                                        (IntegerConstant 3 (Integer 4) Decimal))]
-                                                        FixedSizeArray
-                                                    )
-                                                    ColMajor
-                                                )
+                                                ()
                                             )]
                                             FormatFortran
                                             (Allocatable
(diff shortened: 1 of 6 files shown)


tests/reference/asr-implied_do_loops1-25d26db.stdout
```diff
@@ -141,17 +141,7 @@
...
-                                                    [1, 2, 3]
-                                                    (Array
-                                                        (Integer 4)
-                                                        [((IntegerConstant 1 (Integer 4) Decimal)
-                                                        (IntegerConstant 3 (Integer 4) Decimal))]
-                                                        FixedSizeArray
-                                                    )
-                                                    ColMajor
-                                                )
+                                                ()
```
certik, line 144: This is concerning --- is this not the correct compile-time value of this implied do loop? Why was it removed?

HarshitaKalani: Previously, the compiler was storing the loop indices rather than the evaluated values when handling compile-time expressions. This PR updates the implementation to correctly store the evaluated values of the implied do loop elements instead of their indices.

HarshitaKalani: For the runtime, do we need to do this evaluation in the body_visitor? It is later handled with init_expr and the implied_do_loops pass.

certik: I see. I have a couple questions then:

1. Is this compile-time value correct that was removed?
2. It was removed because it was easier to finish this PR?
3. All compile-time values for implied do loops are handled in `init_expr` and `implied_do_loops` passes?
4. Is this a good design that we construct an ASR and we do not have compile-time values set correctly?

HarshitaKalani:
```
program implied_do_loop1
    integer :: j
    integer :: a(10)=(/(j,j=1,10)/).  -----> first implied do loop
    call s()
    print*, (a(j),j=1,10).  --------> second implied do loop

contains
    subroutine s()
        print *, (j, j = 1, 3). ------> third implied do loop
    end subroutine s
end program
```
Here is the program that we're seeing currently. The first implied do loop is evaluated in the body visitor, since it is compile time. The evaluation for second and third implied do loops are removed from this ASR as they are runtime and they are evaluated in the init_expr and implied_do_loops passes.

Issue #8264, opened 2025-08-06
Submodules : Submodules are not  getting loaded in program file in `monolithic` compilation when parent module is loaded using nesting modules


AnshMehta1:
Towards #536 

a.f90
```.f90
module stdlib_sorting
    implicit none

    interface
        module subroutine sort(n)
            integer, intent(inout) :: n
        end subroutine
    end interface
    
end module stdlib_sorting
```

b.f90
```.f90
submodule(stdlib_sorting) radix_sort
    implicit none
contains

    module subroutine sort(n)
        integer, intent(inout) :: n
        n = 5
    end subroutine

end submodule radix_sort
```

c.f90
```.f90
module tester
    use stdlib_sorting
    implicit none
end module tester

program main
    use tester

    implicit none

    integer :: n = 1
    call sort( n )

    print *, n
end program
```

Steps to reproduce the error :- 
```console
lfortran a.f90 -c
lfortran b.f90 -c
lfortran c.f90 -c
lfortran a.o b.o c.o
```

LFortran Error :- 
```console
/usr/bin/ld: c.o: in function `main':
LFortran:(.text+0x14): undefined reference to `__module_stdlib_sorting_sort'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
The command 'clang -o a.out a.o b.o c.o  -L"/home/lordansh/dev/lfort4/lfortran/src/bin/../runtime" -Wl,-rpath,"/home/lordansh/dev/lfort4/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
```

PR #8347, opened 2025-08-13
fix : Load all the dependent Submodules for a module that is already loaded in global scope

AnshMehta1:
Fixes https://github.com/lfortran/lfortran/issues/8264

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -3352,12 +3352,9 @@ public:
         current_module_dependencies.push_back(al, msym_cc);
 
         ASR::symbol_t *t = current_scope->resolve_symbol(msym);
+        SymbolTable *tu_symtab = current_scope->get_global_scope();
+        bool load_submodules = (!compiler_options.separate_compilation && in_program);
         if (!t) {
-            SymbolTable *tu_symtab = current_scope;
-            while (tu_symtab->parent != nullptr) {
-                tu_symtab = tu_symtab->parent;
-            }
-            bool load_submodules = (!compiler_options.separate_compilation && in_program);
             t = (ASR::symbol_t*)(ASRUtils::load_module(al, tu_symtab,
                 msym, x.base.base.loc, false, compiler_options.po, true,
                 [&](const std::string &msg, const Location &loc) {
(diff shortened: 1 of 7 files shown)


AnshMehta1:
@certik, can you see if the approach I used in this PR is correct for fixing the associated issue??

src/libasr/asr_utils.cpp
```diff
@@ -1039,6 +1040,41 @@ ASR::Module_t* load_module(Allocator &al, SymbolTable *symtab,
     return mod2;
 }
 
+void load_dependant_submodules(Allocator &al, SymbolTable *symtab,
```
certik:
```suggestion
void load_dependent_submodules(Allocator &al, SymbolTable *symtab,
```

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3375,6 +3373,35 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            PassUtils::UpdateDependenciesVisitor v(al);
+            v.visit_TranslationUnit(*tu);
+            #if defined(WITH_LFORTRAN_ASSERT)
+                diag::Diagnostics diagnostics;
+                if (!asr_verify(*tu, true, diagnostics)) {
+                    std::cerr << diagnostics.render2();
+                    throw LCompilersException("Verify failed");
+                };
+            #endif
+            tu_symtab->asr_owner = orig_asr_owner;
```
certik: I think this kind of code would be better handled in `serialization.cpp`

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3375,6 +3373,35 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             throw SemanticAbort();
         }
         ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(t);
+        if (load_submodules) {
+            ASRUtils::load_dependant_submodules(al, tu_symtab, m, x.base.base.loc,
```
certik:
```suggestion
            ASRUtils::load_dependent_submodules(al, tu_symtab, m, x.base.base.loc,
```

AnshMehta1: Done.

src/libasr/asr_utils.cpp
AnshMehta1: Done.

AnshMehta1:
I created a set to keep track of modules whose submodules are loaded. This way we can get rid of reloading our submodules in case of multiple use statements in program.

certik:
We merged many PRs, so I rebased this one, and I am running all checks. If the CI passes, let's merge.

PR #8031, opened 2025-07-11
enh: added default initliazation for string constant and array constant types

sriganeshres:
I have added support for `ArrayConstant_t` and `StringConstant`

PR #8348, opened 2025-08-13
enh: added default initialization for string constant and array constant types

sriganeshres:
from: #8031


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3379,6 +3379,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         }
     }
 
+    static bool isStringDescriptorTy(llvm::Type* T) {
+        llvm::StructType* ST = llvm::dyn_cast<llvm::StructType>(T);
+        if (!ST)
+            return false;
```
certik, line 3385: Always use `{}` around the `then` clause in if statements.

certik:
CI fails.

PR #8354, opened 2025-08-14
refactor: convert some string runtime functions to be null-independent

akramhany:
Towards #8241

refactors `_lfortran_file_write` and `_lfortran_string_write`

### DIFF
--- a/tests/reference/llvm-derived_types_32-4684b97.json
+++ b/tests/reference/llvm-derived_types_32-4684b97.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-derived_types_32-4684b97.stdout",
-    "stdout_hash": "a9d31c3ad0082954d9e8170006574055702db43685eed271fe306d22",
+    "stdout_hash": "b76c8cd05d62fcd726eba493589f74fd3fc415f8f658460ee46f1e4a",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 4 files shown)


certik:
independant -> independent

akramhany:
@assem2002 can you review this please?

assem2002:
Will review it first thing tomorrow.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5029,38 +5042,51 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
...
-    char* str;
-    // int64_t str_len; >> TODO: Pass length
-    char* end_data = ""; int64_t end_len;
-    if(strcmp(format, "%s%s") == 0){
+    char* str = "";
+    int64_t str_len = 0;
+    char* end_data = "";
+    int64_t end_len = 0;
+
+    char *c_format = to_c_string((const fchar*)format, format_len);
```
assem2002, line 5050: I think that conversion (copy) isn't necessary.
The usage of `to_c_string`.

assem2002:
The changes are good as intended, just left minor comments.

src/libasr/runtime/lfortran_intrinsics.c
akramhany: You mean because we create format so that it always ends with a `\0`? I think you are right yes, but doesn't this mean that the `format_len` won't be used?

assem2002:
This will unnecessarily hurt performance, but it's fine for now.

certik:
> This will unnecessarily hurt performance, but it's fine for now.

Can you open up an issue what hurts performance? 

Sorry about that, I thought it's ready to go in.

akramhany:
> > This will unnecessarily hurt performance, but it's fine for now.
> 
> Can you open up an issue what hurts performance? 
> 
> Sorry about that, I thought it's ready to go in.

The problem is that I was converting the Fortran strings to C strings, where there was another way of refactoring without the need to do that (thanks @assem2002 for pointing that).

I will open a new PR with these changes.

certik:
Perfect, go ahead and open a PR with the proper fix.

PR #8366, opened 2025-08-15
ref: rename `no_warnings` compiler option to `show_warnings`

parth121101:
Towards: #8365


src/bin/lfortran_command_line_parser.cpp
```diff
@@ -66,9 +67,9 @@ namespace LCompilers::CommandLineInterface {
 
         // LFortran specific options
         // Warning-related flags
-        app.add_flag("--no-warnings", compiler_options.no_warnings, "Turn off all warnings")->group(group_warning_options);
+        app.add_flag("--no-warnings", disable_warnings, "Turn off all warnings")->group(group_warning_options);
         app.add_flag("--no-style-warnings", opts.disable_style_suggestions, "Turn off style suggestions")->group(group_warning_options);
-        app.add_flag("--style-warnings", style_warnings, "Enable style suggestions")->group(group_warning_options);
+        app.add_flag("--style-warnings", style_suggestions, "Enable style suggestions")->group(group_warning_options);
```
certik, line 72: Should the option be called `--style-suggestions`?

parth121101: Yes sure, will address this in future PR

Issue #8295, opened 2025-08-08
Bounds checking for `y=x`


certik:
This should fail by default:
```fortran
program expr2
implicit none

integer, allocatable :: x(:), y(:)
allocate(x(4), y(3))

x = 2
y = 3
print *, x, y
y = x
print *, x, y

end program
```
But should work with `--realloc-lhs`.

This should always fail:
```fortran
program expr2
implicit none

integer :: x(3)
integer, allocatable :: y(:)
allocate(y(4)) ! works with 3, must fail with 2 or 4

x = 2
y = 3
print *, x, y
x = y
print *, x, y

end program
```
It fails with 2, and works with 3, which is both correct. But it should also fail with 4.

PR #8367, opened 2025-08-15
feat: add bounds checking for array assignment

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8295

- Inserts a `DebugCheckArrayBounds` node after inserting the `reallocate` statement in the `array_op` pass.
- Made a function `get_expr_size_var()` to get the `Var` inside the `ArraySize` expression. I've done it using if-else statements. It might not be the best way, we need to cover all the `expr` nodes.

I haven't fixed the `realloc_lhs` bug yet, I'll fix it after I get third party passing in CI with existing changes.

### DIFF
--- a/src/libasr/asr_verify.cpp
+++ b/src/libasr/asr_verify.cpp
@@ -949,6 +949,23 @@ public:
         BaseWalkVisitor<VerifyVisitor>::visit_ArraySize(x);
     }
 
+    void visit_DebugCheckArrayBounds(const ASR::DebugCheckArrayBounds_t& x) {
+        if (check_external) {
+            require(ASR::is_a<ASR::Var_t>(*x.m_target) ||
+                    ASR::is_a<ASR::ArrayPhysicalCast_t>(*x.m_target) ||
+                    ASR::is_a<ASR::StructInstanceMember_t>(*x.m_target) ||
+                    ASR::is_a<ASR::BitCast_t>(*x.m_target) ||
+                    ASR::is_a<ASR::ArrayConstant_t>(*x.m_target), "DebugCheckArrayBounds::m_target must be Var, ArrayPhysicalCast, StructInstanceMember, BitCast, or ArrayConstant");
+
+            require(ASR::is_a<ASR::Var_t>(*x.m_value) ||
+                    ASR::is_a<ASR::ArrayPhysicalCast_t>(*x.m_value) ||
+                    ASR::is_a<ASR::StructInstanceMember_t>(*x.m_value) ||
+                    ASR::is_a<ASR::BitCast_t>(*x.m_value) ||
+                    ASR::is_a<ASR::ArrayConstant_t>(*x.m_value), "DebugCheckArrayBounds::m_value must be Var, ArrayPhysicalCast, StructInstanceMember, BitCast, or ArrayConstant");
+        }
+        BaseWalkVisitor<VerifyVisitor>::visit_DebugCheckArrayBounds(x);
+    }
+
     void visit_ArraySection(const ArraySection_t &x) {
         require(
             ASR::is_a<ASR::Array_t>(*x.m_type),
--- a/src/libasr/asr_utils.cpp
+++ b/src/libasr/asr_utils.cpp
@@ -3280,19 +3280,24 @@ ASR::expr_t* get_compile_time_array_size(Allocator& al, ASR::ttype_t* array_type
 
 template<typename T>
 ASR::expr_t* get_binop_size_var(ASR::expr_t* x) {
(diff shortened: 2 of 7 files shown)


src/libasr/ASR.asdl
```diff
@@ -33,6 +33,7 @@ stmt
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
     | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs)
+    | DebugCheckArrayBounds(expr target, expr value)
```
certik, line 36: Let's add asr_verify() checks for this node and document there the expected constraints. For example I am guessing `value` must evaluate to an integer number at runtime, correct?

Is `target` just a `Var`, or can it be other nodes also? The verify should enforce these restrictions.

certik:
I checked the overall approach, it looks good, didn't see anything major to fix. Thanks for this.

Let's get the CI to pass, then I'll do a final detailed review.

certik:
The CI failure on macOS LLVM 11 is unrelated: https://github.com/lfortran/lfortran/issues/8374. Once fixed, just rebase this PR.

The other failures are due to the same assert failure, so that should be fixed.

src/libasr/ASR.asdl
xaerru: The expressions `target` and `value` are the duplicated from the original `Assignment`. The `ArraySize` node outside them is created in the backend and then visited.

Should I create an `ArraySize` where I create the `DebugCheckArrayBounds` node?

certik: Is this the relevant line that creates ASR in the backend: https://github.com/lfortran/lfortran/pull/8367/files#diff-48a7161397f9eb8b730f0cbec2f11ce1471aac961dc6b16dc558cf161c375054R6883?

In general we should not do that. It would be better to either do it as part of some pass, or even better in AST->ASR. However, I think we can upgrade this implementation later, the code is short in the backend, so shouldn't be a blocker for a merge.

xaerru: > Is this the relevant line that creates ASR in the backend

Yes.

I've added the `asr_verify.cpp` checks, shifting the creation of `ArraySize` can be done after this PR.

src/libasr/asr_verify.cpp
```diff
@@ -949,6 +949,23 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
...
+                    ASR::is_a<ASR::ArrayPhysicalCast_t>(*x.m_target) ||
+                    ASR::is_a<ASR::StructInstanceMember_t>(*x.m_target) ||
+                    ASR::is_a<ASR::BitCast_t>(*x.m_target) ||
+                    ASR::is_a<ASR::ArrayConstant_t>(*x.m_target), "DebugCheckArrayBounds::m_target must be Var, ArrayPhysicalCast, StructInstanceMember, BitCast, or ArrayConstant");
+
+            require(ASR::is_a<ASR::Var_t>(*x.m_value) ||
+                    ASR::is_a<ASR::ArrayPhysicalCast_t>(*x.m_value) ||
+                    ASR::is_a<ASR::StructInstanceMember_t>(*x.m_value) ||
+                    ASR::is_a<ASR::BitCast_t>(*x.m_value) ||
+                    ASR::is_a<ASR::ArrayConstant_t>(*x.m_value), "DebugCheckArrayBounds::m_value must be Var, ArrayPhysicalCast, StructInstanceMember, BitCast, or ArrayConstant");
```
certik, line 964: Also I think we require that if it is a Var, it must be an array.

certik:
Not sure what the failure is:
```
[  0%] Building Fortran object CMakeFiles/program_cmake_01.dir/program_cmake_01.f90.o
[  0%] Linking Fortran executable program_cmake_01
[  0%] Built target program_cmake_01
[  0%] Building Fortran object CMakeFiles/program_cmake_02.dir/program_cmake_02.f90.o
style suggestion: Use '==' instead of '.eq.'
 --> /home/runner/work/lfortran/lfortran/integration_tests/program_cmake_02.f90:4:49
  |
4 |       integer stop ; stop = 1 ; do while ( stop .eq. 0 ) ; end do
  |                                                 ^^^^ help: write this as '=='


[  0%] Linking Fortran executable program_cmake_02
make[2]: *** [CMakeFiles/program_cmake_02.dir/build.make:96: program_cmake_02] Terminated
make[1]: *** [CMakeFiles/Makefile2:1585: CMakeFiles/program_cmake_02.dir/all] Terminated
make: *** [Makefile:101: all] Terminated
```

xaerru:
I couldn't reproduce it on ubuntu. Can you try rerunning it?

certik:
It might have been a GitHub issue. I rebased everything, let's see what happens.

PR #8338, opened 2025-08-12
fix: append unique string to intermediate object file name

kmr-srbh:
resolves #8294

> test.f90
```fortran
program main
   implicit none

   print *, 1
end program
```

> test_compiler.sh
```shell
seq 1 10 | parallel -j 10 "
    mkdir -p compilertest_{};
    cd compilertest_{};
    cp /Users/saurabhkumar/OpenSource/lfortran/test.f90 .;
    lfortran test.f90
"
```

#### Before
```console
1
1
1
1
clang: error: no such file or directory: '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
clang: error: no such file or directory: '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Tip: If there is a linker issue, switch the linker using --linker=<CC> option or create an environment variable `export LFORTRAN_LINKER=<CC>`, where CC is clang or gcc
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker execuatable
1
1
clang: error: no such file or directory: '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp.o'
[...]

PR #7036, opened 2025-04-19
feat: mark modules as external if global procedures are present and no program is present

Pranavchiku:
Towards #6894.

PR #1898, opened 2023-06-29
Implementation of min0() using IntrinsicFunction 

HarshitaKalani:
Implemented min0

PR #8370, opened 2025-08-15
refactor: differentiate compilation ID creation for monolithic and separate compilation

kmr-srbh:
Implementing suggestion in https://github.com/lfortran/lfortran/pull/8338#discussion_r2276618717.

- We use `LCOMPILERS_UNIQUE_ID` for uniquely naming intermediate object files during both compilation modes. 
- We use `lcompilers_unique_ID_separate_compilation` for uniquely naming symbols during separate compilation.

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -80,7 +80,7 @@
 #define MAGENTA "\033[35m"   // Magenta for 'File reading' and 'Src -> ASR'
 #define RED        "\033[31m"   // Red for 'Time taken by pass' and 'ASR -> ASR passes'
 
-extern std::string lcompilers_unique_ID;
+extern std::string lcompilers_unique_ID_separate_compilation;   // Used in naming unique symbols during separate compilation
 extern std::string lcompilers_commandline_options;
 
 namespace {
@@ -129,7 +129,8 @@ std::string get_unique_ID() {
 }
 
 // The unique compilation ID for this invocation of the compiler.
-const std::string LFORTRAN_COMPILATION_ID = get_unique_ID();
+// Used in naming unique intermediate object files during both compilation modes.
+const std::string LCOMPILERS_UNIQUE_ID = get_unique_ID();
 
 void print_one_component(std::string component) {
     std::istringstream ss(component);
@@ -1146,7 +1147,7 @@ int compile_src_to_object_file(const std::string &infile,
     t1 = std::chrono::high_resolution_clock::now();
     LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
         result = fe.get_asr2(input, lm, diagnostics);
-    lcompilers_unique_ID = compiler_options.separate_compilation ? get_unique_ID() : "";
+    lcompilers_unique_ID_separate_compilation = compiler_options.separate_compilation ? LCOMPILERS_UNIQUE_ID : "";
 
     time_src_to_asr = std::chrono::duration_cast<std::chrono::microseconds>(t2 - t1).count();
     bool has_error_w_cc = compiler_options.continue_compilation && diagnostics.has_error();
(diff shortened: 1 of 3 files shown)


src/bin/lfortran.cpp
```diff
@@ -1146,7 +1144,6 @@ int compile_src_to_object_file(const std::string &infile,
     t1 = std::chrono::high_resolution_clock::now();
     LCompilers::Result<LCompilers::ASR::TranslationUnit_t*>
         result = fe.get_asr2(input, lm, diagnostics);
-    lcompilers_unique_ID = compiler_options.separate_compilation ? get_unique_ID() : "";
```
kmr-srbh, line 1149: I deleted this statement because we are setting the value once when we call `main_app()` and have it available globally.

kmr-srbh:
The CI failed as:

```console
../integration_tests/derived_types_19.f90 * asr
Reference JSON: tests/reference/asr-derived_types_19-26385d4.json
Output JSON:    tests/output/asr-derived_types_19-26385d4.json
Omitting 10 identical items
Differing items:
{'stdout_hash': '15df11a9eb868d2f862a088d39b75197f5f6e9d5296b07a39b3aa605'} != {'stdout_hash': '8adc9850110270921fcde3d59da320fa244d7d14a38475908a23d732'}
Diff against: tests/reference/asr-derived_types_19-26385d4.stdout
73c73
<                                             ~select_type_block_:
---
>                                             ~select_type_block__CRlVr50XXpUIeAxkB4rxib:
89c89
<                                                     ~select_type_block_
---
>                                                     ~select_type_block__CRlVr50XXpUIeAxkB4rxib
145c145
<                                                     7 ~select_type_block_
---
>                                                     7 ~select_type_block__CRlVr50XXpUIeAxkB4rxib

```

I see that the variable `lcompilers_unique_id` is used in implementing the function `get_unique_name()` which is used a lot in our codebase, hence the error.

One way to address this issue, is by creating a copy of`lcompilers_unique_id`, setting original `lcompilers_unique_id` based on the compiler options like we previously did, then after the ASR stage is done, we store back the copied value into `lcompilers_unique_id`. A drawback is, that we will have to do this for 13 cases where we can go through the ASR phase.

src/bin/lfortran.cpp
certik: Is this and the similar change below causing the tests to fail?

certik: This was introduced in https://github.com/lfortran/lfortran/pull/7036/files#r2054526796 and I objected against it there, but as usual I said "let's fix it later" and it never got fixed and now it bit us again.

certik:
I did some history research:
* The initial implementation of the pattern `compilers_unique_ID = compiler_options.generate_object_code ? get_unique_ID() : "";` was done in https://github.com/lfortran/lfortran/pull/1898, this PR also duplicated the `get_unique_ID()` implementation in the code base
* The same pattern was then copied over in https://github.com/lfortran/lfortran/pull/7036
* In several other PRs the code has been moved around, or more conditions added

To fix this, I recommend the following:

* [ ] define `get_unique_ID()` just once in the code base
* [x] Assign it in the main program into the global variable `lcompilers_unique_ID =  get_unique_ID()`
* [x] On the next line create a new global variable `lcompilers_unique_ID_separate_compilation = (parser.opts.compiler_options.separate_compilation || compiler_options.generate_code_for_global_procedures ? lcompilers_unique_ID : ""`.
* [x] Use `lcompilers_unique_ID_separate_compilation` at the places that requires this modified logic.
* [ ] Later: remove `lcompilers_unique_ID_separate_compilation` and replace it with an if statement to `compiler_options.generate_code_for_global_procedures`, etc.

kmr-srbh:
That looks good! I am pushing the changes.

kmr-srbh:
Done in https://github.com/lfortran/lfortran/pull/8370/commits/ed34c80263ad6647d89001474d76fb7cc5cac55f. Also updated the PR description.

src/bin/lfortran.cpp
```diff
@@ -129,7 +129,8 @@ std::string get_unique_ID() {
 }
 
 // The unique compilation ID for this invocation of the compiler.
-const std::string LFORTRAN_COMPILATION_ID = get_unique_ID();
+// Used in naming unique intermediate object files during both compilation modes.
+const std::string LCOMPILERS_UNIQUE_ID = get_unique_ID();
```
kmr-srbh, line 133: I used all-caps to name this to signify the const nature. I can make it all lowercase if you say.

kmr-srbh:
CI failure is due to #8374.

src/bin/lfortran.cpp
kmr-srbh: > define get_unique_ID() just once in the code base

We are calling `get_unique_ID()` only once actually, at this line. After that, at other places, we use the value of `LCOMPILERS_UNIQUE_ID`.

PR #8380, opened 2025-08-16
enh: get a full example working with new classes

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/7068.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -70,6 +70,7 @@ if [[ $WIN != "1" ]]; then
     ./run_tests.py -b llvm llvm2 llvm_rtlib llvm_nopragma
     ./run_tests.py -b llvm2 llvm_rtlib llvm_nopragma -f
     ./run_tests.py -b llvm -f -nf16
+    ./run_tests.py -b llvm_new_classes
     cd ..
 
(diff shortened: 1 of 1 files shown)


tests/classes1.f90
```diff
@@ -13,11 +13,14 @@ subroutine show_x(this)
...
+    b%x = 10
+    call b%show_x()
+    print *, b%x
+    if ( b%x /= 12 ) error stop
+end program
 
-!     type(base) :: b
-!     b%x = 10
-!     call b%show_x()
-! end program
```
certik, line 23: Why don't you add this to integration tests, since you added the new label there and get it running? Then we can slowly start adding the label to more tests.

certik:
Is the generated LLVM now the same as Clang's?

tests/classes1.f90
Pranavchiku: The test class_01 is more complicated than this, so I thought to avoid adding this.

Pranavchiku:
Yes. I’ve got a few more working, will push in a while. I’ve changes ready. Let’s merge this if CI gets through.

PR #8382, opened 2025-08-17
enh: support association of base to derived type

Pranavchiku:
Towards https://github.com/lfortran/lfortran/issues/7068


tests/classes2.f90
```diff
@@ -0,0 +1,45 @@
...
+use Defs
+implicit none
+
+    real :: res
+    class(point), pointer :: ptr
+
+    type(point2d), target :: p2d = point2d( 3, 4 )
+
+    ptr => p2d
+    ! res = ptr%radius()
```
Pranavchiku, line 44: need to understand how clang does this, won't be difficult, will be doing in subsequent PRs.

PR #8385, opened 2025-08-17
feat: add new `disable_implicit_typing` CLI option

parth121101:
Towards: #7937 
This pull request introduces a new CLI option, `--disable-implicit-typing`, to disable implicit typing. With this change, one can now combine the `--std=f23` or `--std=legacy` CLI flags with the `--disable-implicit-typing` option.

Test Cases:
I have added two test cases:
1) Using `--std=f23` with `--disable-implicit-typing` (disables implicit typing while retaining other default options).
2) Using `--disable-implicit-typing` with `--implicit-typing` (this is an error condition).

### DIFF
--- a/src/bin/lfortran_command_line_parser.cpp
+++ b/src/bin/lfortran_command_line_parser.cpp
@@ -78,6 +78,7 @@ namespace LCompilers::CommandLineInterface {
         app.add_flag("--fixed-form-infer", opts.fixed_form_infer, "Use heuristics to infer if a file is in fixed form")->group(group_language_options);
         app.add_option("--std", opts.arg_standard, "Select standard conformance (lf, f23, legacy)")->group(group_language_options);
         app.add_flag("--implicit-typing", compiler_options.implicit_typing, "Allow implicit typing")->group(group_language_options);
+        app.add_flag("--disable-implicit-typing", opts.disable_implicit_typing, "Disable implicit typing")->group(group_language_options);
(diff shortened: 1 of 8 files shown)


src/bin/lfortran_command_line_parser.cpp
```diff
@@ -217,14 +218,6 @@ namespace LCompilers::CommandLineInterface {
             app.parse(args);
         }
 
-        if (opts.disable_style_suggestions) {
-            compiler_options.show_style_suggestions = false;
-        }
-
-        if (disable_warnings) {
-            compiler_options.show_warnings = false;
-        }
```
certik, line 226: I can't tell easily if this change breaks something or not. Is this needed for this PR?

parth121101: Yes, this changes is not required. Will remove this.

parth121101: I had a small confusion regarding what will happen if both flags (disable and enable) are used together. So I shifted this logic below this.
```
if (opts.disable_style_suggestions && style_suggestions) {
            throw lc::LCompilersException("Cannot use --no-style-warnings and --style-warnings at the same time");
}
```

But yes this will work without the above change as well. To test this I will add an error test case in future PR just the way I did here. 
```
Cannot use --disable-implicit-typing and --implicit-typing at the same time
```

PR #8389, opened 2025-08-17
fix: add struct name to make unique StructInstanceMember of different types

### LINKED ISSUE #8378: Handle successive call of function with same name but from different derived type
MRE:
```fortran
module m_toml_array
    type :: toml_array
    contains
        procedure :: get
    end type toml_array
contains
    subroutine get(this, n)
        class(toml_array), intent(inout) :: this
        integer, intent(in)              :: n
        print *, "array_get called with n=", n
    end subroutine get
end module m_toml_array

module m_toml_table
    type :: toml_table
    contains
        procedure :: get
    end type toml_table
contains
    subroutine get(this, key)
        class(toml_table), intent(inout) :: this
        character(len=*), intent(in) :: key
        print *, "table_get called with key=", key
    end subroutine get
end module m_toml_table

program test_mre
    use m_toml_array
    use m_toml_table
    implicit none

    type(toml_table) :: tbl
    type(toml_array) :: arr

    call arr%get(5)
    call tbl%get("Hello")
end program test_mre
```

LFortran Error:
```console
%string_descriptor* @string_const.5
 i32*  call void @__module_m_toml_array_get(%toml_array_polymorphic* %9, %string_descriptor* @string_const.5)
```

jinangshah21:
Fixes #8378

### DIFF
--- a/tests/reference/asr-class_01-704dee8.json
+++ b/tests/reference/asr-class_01-704dee8.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-class_01-704dee8.stdout",
-    "stdout_hash": "b927d39609410a79de8b443e94fefe89b7f1e9037a22ef0156f9077b",
+    "stdout_hash": "b7e9f4876106ffe16c07db363ad4774db50f6a40f2d3dc760dfd312d",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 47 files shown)


certik:
There are two kinds of errors at the CI:

* sorting, should be resolved now after #8391 is in.
* `FAILED tests/server/tests/test_features.py::test_code_completion - ValueError: list.remove(x): x not in list`, we should investigate what is going on --- likely the Python test suite fails, but probably due to some underlying change in LFortran.

PR #8390, opened 2025-08-17
fix: handle nested structconstructor in declaring global vars

jinangshah21:


src/libasr/codegen/llvm_utils.h
```diff
@@ -274,6 +274,9 @@ namespace LCompilers {
             // Converts a constant ASR expression into an equivalent LLVM Constant.
             // Supports integer, logical, and real constants used in struct initializers.
             // Throws an exception for unsupported expression types.
+            void get_type_default_field_values(ASR::symbol_t* struct_sym,
+                llvm::Module* module, std::vector<llvm::Constant*>& field_values);
```
certik, line 278: I think you put this function into a wrong place, doesn't the comment above it apply to the function below?

Issue #8340, opened 2025-08-13
Temps For Strings


assem2002:
Main goal is to dissolve memory ownership ambiguities. 
### StringItem
- Stringview over the original string. (after null-independent refactor. Refer #6892, point 2) 
- No need for temp.
### StringSlice
- Stringview over the original string.  (after null-independent refactor. Refer #6892, point 2)
- No need for temp.
### StringChr
- Creates string type. Creates memory. Needs owner.
- Change implementation into intrinsic.
### StringConcat
- Creates string type. Creates memory. Needs owner.
- Change implementation into intrinsic. 
### StringFormat
- Creates memory. Needs owner.
- PROBLEM : There's no Fortran syntax equivalent to the `StringFormat` node. So we can't have `(Assignment (Var emp_str ) (StringFormat (...))`
- SOLUTION : 
    - StringFormat is always dependent on print, we can make print responsible of the ownership. We can enforce that by adding asr_verify check (print's m_arg must be `stringFormat`). Not the most perfect solution, but StringFromatting still needs improvements anyways.
    - Remove that node, and add string formatting + args directly into `print` node
### String Returns From Functions
- It's handled by subroutine from function.
- Better get handled by the `array_struct_temporary`( simplifier ) pass.

***
Steps : 
- [x] Change `ASR::StringConcat` into intrinsic + remove that node. (#8393)
- [x] Change `ASR::StringChr` into intrinsic + remove that node. https://github.com/lfortran/lfortran/pull/8534
- [ ] `ASR::Print` sometimes accepts `var` node of character, make it accept `ASR::StringFormat` only + Make its implementation free the passed ptr + enforce `(print (stringFormat ...)...)` using `ASR_verify`
- [ ] ~Port the creation of temporaries for String-returing functions from `subroutine_from_function` into `array_struct_temporoary` + make it correspond to `intrinsicFunction` also~
- [x] Move `subroutine_from_function` pass after `intrinsic_function` pass, in order to properly handle any intrinsicFunction that eventually ends up creating  String-returning function.
- [ ] Some `setup_string` usages in the codebase isn't the best usecase. Sometimes we want to create LFortran-String based on the physicalType to hold some value resulting from evaluating some expression, `setup_string` allocates memory unnecessarily and some times we leak due to it (Refer to how we handle string-return `ASR::BitCast` expression )

swamishiju:
[...]

PR #8393, opened 2025-08-18
Add IntrinsicElementalFunction `stringconcat` 

assem2002:
as part of building robust memory ownership for strings
towards #8340

### DIFF
--- a/tests/reference/asr-derived_types_04-da02dd9.json
+++ b/tests/reference/asr-derived_types_04-da02dd9.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-derived_types_04-da02dd9.stdout",
-    "stdout_hash": "ef6398ea67b42799668e257f6a08029b808872b120396339e97cf2c9",
+    "stdout_hash": "fdafc9b0feabc1ad7dc2763afd9517930f80e528f4d079cf995ee337",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-derived_types_06-847ca73.json
+++ b/tests/reference/asr-derived_types_06-847ca73.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-derived_types_06-847ca73.stdout",
-    "stdout_hash": "53698d7410ef5db15c6fec8f1d7db753870c15ed6fb3ac78cbfc585f",
+    "stdout_hash": "a1671ac9cb7ee011281e779bd10f7d064cd63f1cd95669ab07d11d3b",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-string_03-2cd8fec.json
+++ b/tests/reference/llvm-string_03-2cd8fec.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-string_03-2cd8fec.stdout",
-    "stdout_hash": "a3c7abdd70e8bcb2337bd98947c6edae2f22618c78f9449836b41aef",
+    "stdout_hash": "edf6173b398fe9cab2d433bb9fcc158269729f51488416abef266f3a",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-string_03-2cd8fec.stdout
+++ b/tests/reference/llvm-string_03-2cd8fec.stdout
@@ -3,7 +3,6 @@ source_filename = "LFortran"
 
 %string_descriptor = type <{ i8*, i64 }>
 
-@0 = private unnamed_addr constant [67 x i8] c"runtime error: Attempting to allocate already allocated variable!\0A\00", align 1
 @string_const_data = private constant [9 x i8] c"learned \00"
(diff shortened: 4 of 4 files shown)


assem2002:
Thanks for the review @HarshitaKalani. Could be please approve it?

HarshitaKalani (approved):
Looks good to me. Thanks!

src/libasr/asr_utils.h
```diff
@@ -1867,6 +1867,15 @@ static inline bool extract_value(ASR::expr_t* value_expr, T& value) {
     return true;
 }
 
+template <typename T,
+    typename = typename std::enable_if<
+        std::is_same<T, std::complex<double>>::value == false &&
+        std::is_same<T, std::complex<float>>::value == false>::type>
```
certik, line 1873: I don't like this. Is there a way to do the same without this magic? Can we use asserts?

src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -184,12 +184,6 @@
             "return": "int32"
         },
     ],
-    "Repeat": [
-        {
-            "args": [("char", "int")],
-            "return": "allocatable_deferred_string()"
-        }
-    ],
```
certik, line 192: Why is this removed from here?

src/libasr/pass/intrinsic_functions.h
```diff
@@ -4887,6 +4888,156 @@ namespace Adjustr {
...
+                character(*) :: s2
+                integer(4) :: s1_len
+                integer(4) :: s2_len
+                character(:), allocatable :: concat_result
+                s1_len = len(s1)
+                s2_len = len(s2)
+                allocate(character(s1_len + s2_len)::concat_result)
+                concat_result(1:s1_len) = s1
+                concat_result(s1_len+1: len(concat_result)) = s2
+            end function
```
certik, line 5011: I am not sure if the result can be allocatable and if there is any downside. I think the implementation has to be as follows:
```
            function lcompilers_stringconcat(s1, s2) result(concat_result)
                character(len=*), intent(in) :: s1, s2
                character(len=len(s1)+len(s2)) :: concat_result
                concat_result(1:len(s1)) = s1
                concat_result(len(s1)+1:len(concat_result)) = s2
            end function
```
No allocatable strings.

I think the downside of your implementation is that you have to allocate inside the function. My implementation doesn't do any string allocation inside the function. We have to support code like the above anyway, so might as well use it for the implementation.

src/libasr/intrinsic_func_registry_util_gen.py
assem2002: We don't use the generator anymore for `repeat` as return type `string` has various length possibilities

- known  `(character(10)::s1 + character(20)::s2)` OR `(character(10)::s1 + character(foo())::s2)`
- of deferred-length` (character(:), allocatable :: s1 + character(10) :s2`

The generator can't don that for now.

src/libasr/asr_utils.h
assem2002: I'll check. I don't like this C++ magic either.
Anyways I made it to raise error when the expr has no value (compile-time value), I can check from the caller side, no problem.

src/libasr/pass/intrinsic_functions.h
assem2002: It's doable, I just wanted to avoid the functionParam problems. 
I'll push the change, If it passes, let's go for it.

assem2002: @certik Could be please explain, why using `allocate` + `allocatable` type here is a bad practice?

certik: > Could be please explain, why using allocate + allocatable type here is a bad practice?

Yes. Here is the way I see it:

1. Using allocate inside a function when it is not needed is bad practice, because we are instantly slowing things down. And for intrinsic functions we should have such design that doesn't do any unnecessary allocations. It's kind of like using malloc or std::unique_ptr in C++ inside functions. We should not do that, unless there is no other way.

2. The allocatable return type is bad practice for two reasons:
  2a. It forces to allocate inside the function, which is bad practice due to point 1. 
  2b. It also *might* force allocatable on the caller in some cases, which is even worse. I know in some cases the allocatable is not required in the caller and just a copy happens. However I am currently not sure if it works like that in all cases. I think there might be some subtle differences in some corner cases. So this point might or might not apply, would require deep investigation. But if we find in the future that in some corner case it applies, then it will be too late to change.

assem2002: Got it.
1 - Performance concerns
2 - Avoiding potential performance down ( If we happen to allocate on the caller side)

I'd like to mention that we use `malloc` for strings. so both `character(foo()) :: s1` and `character(:),allocatable :: s2` are both allocating memory using syscall `malloc`.
NOTE : We can shift to stack allocation easily for non-allocatable strings.

certik:
To move forward --- can you figure out how to make it work without allocatable result? I think if we require allocatable we are fundamentally closing the door on maximum performance. I rather start with non-allocatable.

assem2002:
I'm handling it. It's simple change, but the functionParam just needs to get handled.

assem2002:
@certik `allocatable` INTO `non-allocatable` handled in this commit https://github.com/lfortran/lfortran/pull/8393/commits/4b664ffdb7dfa4e5ee81d282762bf005bbae05a8

Issue #5904, opened 2025-01-02
read iostat=ios regression


harperjf:
With --std=f23 lfortran 0.42.0 compiled and ran this program to test various separators between input numbers, some standard-conforming, some not. With bad inputs ```ios``` was always 0, but gfortran gave nonzero values in those cases.
Lfortran 0.42.0-835-ge375d9063 won't compile it.
```
  implicit none
  real x(2)
  character(*),parameter:: punc = ",; ."
  integer,parameter:: lpunc =len(punc)
  integer ios,i,j
  character:: msg*80, input(lpunc**2)*4 = &
       [(("2"//punc(i:i)//"5"//punc(j:j),i=1,lpunc),j=1,lpunc)]
  
  do i = 1,size(input)
     print *,'i=',i,'input(i) = "',input(i),'"'
     read(input(i),*,decimal='point',iostat=ios,iomsg=msg) x
     print *,'with decimal=point x(:) =',x,' ios=',ios
     if(ios/=0) print *,trim(msg)
     x(2) = 666
     read(input(i),*,decimal='comma',iostat=ios,iomsg=msg) x
     print *,'with decimal=comma x(:) =',x,' ios=',ios
     if(ios/=0) print *,trim(msg)
  end do
end program
```

certik:
Here is the error I am getting with 0.42.0-835-ge375d9063:
```console
$ lfortran a.f90 
ASR verify pass error: Initialisation of input must reduce to a compile time constant.
 --> a.f90:6:23 - 7:63
  |
6 |      character:: msg*80, input(lpunc**2)*4 = &
  |                          ^^^^^^^^^^^^^^^^^^^^^...
...
  |
7 |           [(("2"//punc(i:i)//"5"//punc(j:j),i=1,lpunc),j=1,lpunc)]
  | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```
The same error with the latest (0.43.0-11-g9465ce08a).

PR #8401, opened 2025-08-18
fix: fix the verify condition for checking compile-time constant arrays

HarshilShah1804:
Towards #5904 

The reported MRE still fails, `LFORTRAN ERROR: Reading into an array of float is not supported.`. Will fix it in subsequent a PR.


HarshilShah1804:
I am not sure why that extra check was present. Will look into the history of the file to get the commit from where that check was introduced.

certik:
That would be great. Then document it here. That way we'll make sure we didn't introduce a bug.

Issue #8319, opened 2025-08-10
Assumed rank array not supported


HarshilShah1804:
MRE
```fortran
program assumed_rank
    implicit none
    real :: a(3) = [1,2,3]
    real :: b(2, 2) = reshape([1, 2, 3, 4], [2, 2])
    call show(a)
    call show(b)
contains
    subroutine show(x)
        real :: x(..)
        print *, size(x)
    end subroutine
end program
```

LFortran:
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
Traceback (most recent call last):
  Binary file "/home/harshil-shah/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2745, in main
    return main_app(argc, argv);
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2669, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 1200, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
  File "/home/harshil-shah/lfortran/src/lfortran/fortran_evaluator.cpp", line 393, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
    compiler_options, run_fn, "", infile);
[...]

PR #8405, opened 2025-08-18
fix: handle using reshape with casting

akramhany:
Towards #8319
This is what I reached for now, we successfully compile the code but the output contains some errors.

#### Code
```fortran
program assumed_rank
    implicit none
    real :: b(3, 2) = reshape([-1, -3, 6, 7, 8, 9], [3, 2])

    print *, b
end program assumed_rank
```

#### Lfortran Output
```sh
3.00000000e+00    2.00000000e+00    6.00000000e+00    7.00000000e+00    8.00000000e+00    9.00000000e+00
```

As seen above, the first 2 numbers are wrong, and it seems that the first 2 numbers of the array in reshape are replaced with the shape.


certik:
@jinangshah21 can you please review it, you fixed many similar bugs recently.

@akramhany, go ahead and add the above MRE as an integration test, and insert "if (..) error stop" checks there, but comment them out since it is currently returning incorrect answers. Then we can merge. Later you can fix the run time also and enable the if checks.

jinangshah21 (approved):
I think this is fine.

certik:
Thanks @jinangshah21. @akramhany please finish the PR per my comment above.

Issue #8383, opened 2025-08-17
Handle StructInstanceMember association in select type


jinangshah21:
MRE:
```fortran

module container_mod
    type, abstract :: base
    end type base
    type, extends(base) :: child
        integer :: x = 42
    end type child
    type :: container
        class(base), allocatable :: val
    end type container
end module container_mod

program demo
    use container_mod
    type(container) :: c
    type(child) :: ch
    class(container), allocatable :: self
    allocate(self)
    allocate(child::self%val)
    select type(val => self%val)
    type is(child)
        print *, "child%x =", val%x
    end select
end program demo
```

LFortran Error:
```console
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store %base* %49, %base_polymorphic* %val, align 8
 %base_polymorphic = type { i64, %base* }
```

jinangshah21:
Done in `select_type_09.f90`

PR #8409, opened 2025-08-19
Fix: Handle StructInstanceMember association in select type

adit4443ya:
Towards #8383


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5959,16 +5959,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                 llvm::Value* casted_val_ptr = builder->CreateBitCast(val_data_ptr, struct_type->getPointerTo());
-
                 llvm::Value* struct_value = builder->CreateLoad(struct_type, casted_val_ptr);
+                llvm::Type* llvm_target_type = llvm_utils->get_type_from_ttype_t_util(x.m_target, target_type, module.get());
+                for(auto map_:name2dertype) {
+                    // for the case where the target is polymorphic
+                    if(llvm_target_type == map_.second && map_.first.find("_polymorphic") != std::string::npos) {
+                        llvm_target = llvm_utils->create_gep2(llvm_target_type, llvm_target, 1);
+                    }
+                }
```
certik, line 5975: Are you searching for `llvm_target_type` in `name2dertype`? If so, you should search directly (`O(1)`), not in a linear fashion (`O(n)`) like this, as that doesn't scale.

adit4443ya: Ohh yes, i changed it with latest commit

certik: How does the new logic work:
```
name2dertype[polymorphic_struct_name] && llvm_target_type == name2dertype[polymorphic_struct_name]
```
Is this testing that `name2dertype[polymorphic_struct_name]` is not nullptr? But then it is testing that it is equal to llvm_target_type?
But `llvm_target_type` can't be null, can it?

Also what if `polymorphic_struct_name` is not in `name2dertype`? Then I think it's undefined.

adit4443ya: Yes,
It is first checking that does `name2dertype[polymorphic_struct_name]` exist (which is not needed i suppose, can be removed), 
main goal is to check `llvm_target_type == name2dertype[polymorphic_struct_name]`, if it's yes then we need to GEP its `base*` from `%base_polymorphic = type { i64, %base* }`

>But llvm_target_type can't be null, can it?

No , it can't be , it's just `llvm::Type` of  `x.m_target`.

> Also what if `polymorphic_struct_name` is not in `name2dertype`

Yes, it won't go in the if loop.

kmr-srbh (changes requested):
@adit4443ya could you please apply the following diff on top of latest `main`? I think that will fix the linked issue. :)

```diff
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index 9408f0f5b..1d12dcbb7 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5878,7 +5878,7 @@ public:
             ASR::ttype_t *type = ASRUtils::get_contained_type(target_type);
             type = ASRUtils::type_get_past_allocatable(type);
             [[maybe_unused]] bool is_target_class = ASRUtils::is_class_type(
-                ASRUtils::type_get_past_pointer(target_type));
+                ASRUtils::type_get_past_allocatable_pointer(target_type));
             [[maybe_unused]] bool is_value_class = ASRUtils::is_class_type(
                 ASRUtils::type_get_past_pointer(
                     ASRUtils::type_get_past_allocatable(value_type)));
```

certik:
I think this is good. Let's send a new PR with https://github.com/lfortran/lfortran/pull/8409#pullrequestreview-3141312911.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5959,16 +5959,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                     ASRUtils::symbol_get_past_external(ASRUtils::symbol_get_past_external(ASRUtils::get_struct_sym_from_struct_expr(x.m_value))));
                 LCOMPILERS_ASSERT(ASRUtils::is_derived_type_similar(struct_type_t, class_type_t));
                 llvm::Type* struct_type = llvm_utils->getStructType(struct_type_t, module.get(), true);
                 llvm::Value* casted_val_ptr = builder->CreateBitCast(val_data_ptr, struct_type->getPointerTo());
-
                 llvm::Value* struct_value = builder->CreateLoad(struct_type, casted_val_ptr);
+                llvm::Type* llvm_target_type = llvm_utils->get_type_from_ttype_t_util(x.m_target, target_type, module.get());
+                // This ensures the case where we are storing the base to base_polymorphic, and hence we need to load base from the later one.
+                std::string polymorphic_struct_name = std::string(struct_type_t->m_name) + "_polymorphic";
+                if(llvm_target_type == name2dertype[polymorphic_struct_name]) {
```
kmr-srbh, line 5973: On a side note, please avoid using the `[]` operator on maps for item access. The operator is overloaded, so it does item insertion with null values if the key does not exist. We can use the `insert()` and `at()` methods for item insertion and item access respectively. :)

certik: I always forget about this. Indeed. Let's fix that in a subsequent PR also.

PR #8410, opened 2025-08-19
fix: ensure valid check for `m_value` in `ReplaceModuleVarWithValue`

kmr-srbh:


certik:
That's fine. Let's add a test?

PR #8418, opened 2025-08-21
LPython sync

swamishiju:
Clean git history PR

Previous #8301


swamishiju:
Seems to work good in LP - I'll add exhaustive tests 

https://github.com/lcompilers/lpython/pull/2866

swamishiju:
> I think it's fine. Let's merge this.
> 
> But we really need tests for all these, otherwise it will get broken again.

I'll open an issue for that to make sure I wont forget

Issue #8180, opened 2025-07-30
Handle string array passed to a `BindC` functionCall


jinangshah21:
MRE:
```fortran 
program test_get_current_directory
  use iso_c_binding, only: c_char
  implicit none

  character(len=1) :: result(2)
  integer :: n
  n = getcwd(result)
contains

  function getcwd(buf) result(n) bind(c, name="getcwd")
    character(kind=c_char,len=1), dimension(:), intent(out) :: buf
    integer :: n
  end function getcwd

end program test_get_current_directory
```

LFortran Error:
```console
LCOMPILERS_ASSERT failed: /home/jinang_shah/Desktop/lfortran/src/libasr/codegen/asr_to_llvm.cpp
function convert_call_args(), line number 10693 at 
ASRUtils::is_character_phsyical_types_matched( ASRUtils::extract_type(ASRUtils::expr_type(x.m_args[i].m_value)), ASRUtils::extract_type(orig_arg->m_type))
ERROR MESSAGE:
Unmatching String Physical Types
Traceback (most recent call last):
  Binary file "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2729, in main
    return main_app(argc, argv);
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 2653, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/jinang_shah/Desktop/lfortran/src/bin/lfortran.cpp", line 1184, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
[...]

PR #8421, opened 2025-08-21
FIX: Handle string array passed to a BindC functionCall

adit4443ya:
Towards #8180


integration_tests/bindc_07.f90
```diff
@@ -0,0 +1,23 @@
...
+  use iso_c_binding, only: c_char, c_ptr, c_null_ptr, c_size_t, c_int, c_associated, c_f_pointer, c_null_char
+  implicit none
+
+  
+  character(len=1) :: result(1)
+  character(len=1024, kind=c_char), pointer :: result1
+  type(c_ptr) :: ptr_result
+  integer(c_size_t) :: size
+  ptr_result=c_null_ptr
+   ptr_result = getcwd(result, size)
```
certik: Inconsistent indentation.

integration_tests/bindc_07.f90
```diff
@@ -0,0 +1,23 @@
...
+   
+   if (.not.c_associated(ptr_result)) then
+      error stop "Failed to get current working directory."
+   end if
```
certik: Why not print the `result`? Is getcwd returning a string into `buf`? Is the caller supposed to allocate `buf` and pass in the size of it?  I think `size` is not initialized.

certik:
@assem2002 can you please review this PR --- is it using the new strings correctly in this case?

certik:
If the CI passes, we can merge it. I rebased it, just in case and enabled all tests.

PR #8422, opened 2025-08-22
refactor: using functions which take explicit `llvm::Type*`.

sriganeshres:


certik:
If tests pass, I'll rebase (just in case) and merge. Keep it running for now, don't add commits/rebase, let's see if CI finishes.

certik:
Ok, the CI seems to work. I rebased it. If it passes now, we can merge.

certik:
Everything passes. The `LFortran CI (OS=macos-latest, LLVM=11)` test seems to be stuck on:
```
vardim: N = 16 NPT = 0, Random test 8

Warning: NEWUOA: Invalid NPT; it should be an integer in the interval [N+2, (N+1)(N+2)/2] and less than MAXFUN; it is set to 33.


Warning: NEWUOA: Invalid RHOBEG; it should be a positive number; it is set to  1.0000000000000000E+000.
```
line 26403 in the log.

Update: it just went past it. Weird, it was stock on it for several minutes.

PR #8424, opened 2025-08-22
Enh : Completely Remove null dependency from strings

assem2002:
Towards bullet point **2** in here https://github.com/lfortran/lfortran/issues/6892#issue-2977394717
This PR removes our dependency on null-termination in strings (in LLVM backend)

***
That should make our lives easier with `stringitem` and `stringslice`

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1807,7 +1807,7 @@ RUN(NAME string_66 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_67 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_68 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_69 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
-RUN(NAME string_70 LABELS llvm llvm_wasm llvm_wasm_emcc) # GFortran doesn't agree with us (it tests some non-stadard-conforming code)
+RUN(NAME string_70 LABELS llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/tests/reference/llvm_new_classes-classes1-bfe507d.json
+++ b/tests/reference/llvm_new_classes-classes1-bfe507d.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm_new_classes-classes1-bfe507d.stdout",
-    "stdout_hash": "6d22309f439b4ce02a19c8ae0604e3252b9f1228c304ffdb53589263",
+    "stdout_hash": "54112ce5817458bdccf61c7dd6f882a89251b0cc5697b3ea243baf14",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm_new_classes-classes1-bfe507d.stdout
+++ b/tests/reference/llvm_new_classes-classes1-bfe507d.stdout
@@ -1,11 +1,12 @@
 ; ModuleID = 'LFortran'
 source_filename = "LFortran"
 
-%_VTable_base.type = type { [3 x i8*] }
 %base = type <{ i32 (...)**, i32 }>
 %string_descriptor = type <{ i8*, i64 }>
 
-@_VTable_base = linkonce_odr unnamed_addr constant %_VTable_base.type { [3 x i8*] [i8* null, i8* null, i8* @__module_xx_show_x] }, align 8
+@_Name_base = private unnamed_addr constant [5 x i8] c"base\00", align 1
+@_Type_Info_base = linkonce_odr unnamed_addr constant { i8*, i8* } { i8* getelementptr inbounds ([5 x i8], [5 x i8]* @_Name_base, i32 0, i32 0) }, align 8
+@_VTable_base = linkonce_odr unnamed_addr constant { [3 x i8*] } { [3 x i8*] [i8* null, i8* @_Type_Info_base, i8* bitcast (void (%base*)* @__module_xx_show_x to i8*)] }, align 8
 @0 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
 @serialization_info = private unnamed_addr constant [3 x i8] c"I4\00", align 1
 @1 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1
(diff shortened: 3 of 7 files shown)


assem2002:
- Changes in `llvm_utils.cpp` handles the removal of `\0` from strings + not allocating it.
- Changes in `lfortran_intrinsics.c` handles some minor issues related to the old dependency on null char.
- Other files are just modifications and enhancements + minor changes related to passing length to some runtime functions in our C-library.

integration_tests/string_70.f90
```diff
@@ -0,0 +1,23 @@
...
+    implicit none
+    character(10), allocatable :: a
+    character(:), allocatable :: b
+    character(:), allocatable :: c
+
+    !-----------------------------------------------!
+
+    a = "HelloWorld"
+    if(.not. allocated(a)) error stop
+    a = b
```
certik: GFortran fails on this line --- are you sure it is allowed to assign a non-allocated string from RHS to LHS?

To be conservative, we should probably give a nice runtime error message instead. We can later relax it and support it as an extension if it is allowed, but I think it is not allowed. When you assign something, the RHS has to be allocated. I think this is covered in `Clause 9.7.1.3, "Allocation of allocatable variables":`:

> An allocatable variable has a status of ‘unallocated’ if it is not allocated. The status of an allocatable variable becomes unallocated if it is deallocated (9.7.3) or if it is given that status by the intrinsic subroutine MOVE_ALLOC. **An allocatable variable with this status shall not be referenced or defined**.

tests/reference/run-string_assignment_02-84c751f.stderr
```diff
@@ -1 +0,0 @@
-LFORTRAN ERROR: Run-time Error : Copying from unallocated RHS string
```
certik, line 1: This was a good error. Let's keep it.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13472,6 +13489,9 @@ Result<std::unique_ptr<LLVMModule>> asr_to_llvm(ASR::TranslationUnit_t &asr,
         LCOMPILERS_ASSERT(diagnostics.has_error())
         Error error;
         return error;
+    } catch (...){
+        std::cout<<"UNKNOWN EXCEPTION CAUGHT WHILE VISITING ASR WITH LLVM BACKEND (ASR->LLVM)\n";
+        std::exit(1);
```
certik: This error message has to be better.  I think this will catch `LCompilersException`  too?

I would catch it in the main program I think.

certik:
I think overall it is fine. But I think the first two comments above should be fixed first?

integration_tests/string_70.f90
assem2002: I see. 
It's not the best thing to tolerate a RHS that's not allocated. I didn't find any similar clause to the one you shared while looking up that case, so I went with the convention of moving the **allocation** state from RHS to LHS hence the test `string_70.f90` line `14`.

The problem is that we use that trick (tolerating unallocated RHS) in our `nested_vars` pass, If we instead raise an error, we'll fail to cover some use cases that `nested_vars` was handling correctly (refer to test `allocate_10.f90`); What I mean is that we're inserting that statement by our `nested_var` pass, if we decide to raise an error then our pass will be inserting a wrong statement.

assem2002: Here is the case I'm concerned about :
```.f90
module A_allocate_10
  
    contains
    
    subroutine f()
        character(len=:), allocatable:: argument
        call g()
        print *, allocated(argument)
    
    contains
    
        subroutine g()
            allocate(character(len=10) :: argument)
        end subroutine
    
    end subroutine
    
end module
```

AFTER `nested_vars` PASS : 
```.f90
module A_allocate_10
    character(len=:),allocatable :: temp_argument !!!!!!!!!!! <<<<<<<<<<<<<
    
    contains
    
    subroutine f()
        character(len=:), allocatable :: argument
        temp_argument = argument !!!!!!!!<<<<<<<<<<<<<
        call g()
        argument = temp_argument !!!!!!!!! <<<<<<<<<<<<<
        print *, allocated(argument)
    
    contains
    
        subroutine g()
            allocate(character(len=10) :: temp_argument) !!!!!!!!!!!!! <<<<<<<<<<<<<
        end subroutine
    
    end subroutine
    
end module
```

tests/reference/run-string_assignment_02-84c751f.stderr
assem2002: Will revert it back :+1:

src/libasr/codegen/asr_to_llvm.cpp
assem2002: Yeah I think main catches that one, It's better to remove it yes.

I was just throwing `CodeGenError` from `llvm_utils.cpp` and due to the internalLinkage of the `CodeGenError` class, `asr_to_llvm.cpp` wasn't able to catch it and I kept looking up for the reason and eventually decided to add that block to prevent others from getting confused when they come across my scenario.
I also like to mention that `llvm_utils.cpp` has lots of `throw CodeGenError` which will result in an `unknown expection` message, we need to fix that to prevent confusion and improve debugging.

integration_tests/string_70.f90
assem2002: I think I might have a fix, yes. I'll just insert `if(allocated())` block.

assem2002: resolved in https://github.com/lfortran/lfortran/pull/8424/commits/5081cff900bee76d29eaed391d36a5326951051e

src/libasr/codegen/asr_to_llvm.cpp
assem2002: resolved in commit https://github.com/lfortran/lfortran/pull/8424/commits/5081cff900bee76d29eaed391d36a5326951051e

tests/reference/run-string_assignment_02-84c751f.stderr
assem2002: resolved in commit https://github.com/lfortran/lfortran/pull/8424/commits/5081cff900bee76d29eaed391d36a5326951051e

assem2002:
@certik Would you mind to give it another review ? You can review the last 2 commits only (the new ones).

certik:
On it.

integration_tests/CMakeLists.txt
```diff
@@ -1807,6 +1807,7 @@ RUN(NAME string_66 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_67 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_68 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_69 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
+RUN(NAME string_70 LABELS llvm llvm_wasm llvm_wasm_emcc)
```
certik, line 1810:
```suggestion
RUN(NAME string_70 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2330,15 +2333,15 @@ namespace LCompilers {
                     "Global variable should have constant-compile-time known length"
                 );
                 // Type -> [len x i8]
-                llvm::ArrayType *char_array_type = llvm::ArrayType::get(llvm::Type::getInt8Ty(context), len + 1 /*null-char*/);
-                // [len x i8] c "DATA HERE\00"
+                llvm::ArrayType *char_array_type = llvm::ArrayType::get(llvm::Type::getInt8Ty(context), len);
+                // [len x i8] c "DATA HERE\"
```
certik, line 2337:
```suggestion
                // [len x i8] c "DATA HERE"
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -733,7 +747,7 @@ within character string edit descriptor
 E.g.; "('Number : ', I 2, 5 X, A)" becomes '('Number : ', I2, 5X, A)'
 */
 char* remove_spaces_except_quotes(const fchar* format, const int64_t len, int* cleaned_format_len) {
-    char* cleaned_format = malloc(len + 1);   // +1 for optional '\0' at end
+    char* cleaned_format = malloc(len + 1);
```
certik, line 750: I guess the +1 is not needed now?

PR #8425, opened 2025-08-22
visit_ArrayItem: CreateGEP -> CreateGEP2

certik:
Towards https://github.com/lfortran/lfortran/issues/7836.


certik:
@assem2002 is this the right fix?

assem2002:
According to the current String implementation in LLVM backend, This seems incorrect.
I think we have wrong usage over `ArrayItem` node.
hence I believe that whole block doesn't have proper effect in the codebase.

I'll check locally.

assem2002:
I removed that block of code. Let's see if it passes 3rd party.

assem2002:
@certik May you mention what this PR was attempting to fix.

certik:
This is towards https://github.com/lfortran/lfortran/issues/7836.

PR #8426, opened 2025-08-23
fix: output of reshape used in declarations

akramhany:
A subsequent PR to #8405

Anytime we have the `reshape` function used with casting in an **initialization**, I decided to handle this in the `init_expr` pass, as I would have all required info to do that.

Another option would have been the following:

1. Any use of `reshape` with an array that is a `parameter`, handle it's reshape in ASR before passes, and use this value directly in `init_expr` pass instead of reshaping it in the `init_expr` pass.

   ```fortran
   real, parameter :: b(3, 2) = reshape([-1, -3, 6, 7, 8, 9], [3, 2])
   ```

2. Any use of `reshape` with an array that is not a `parameter`, convert this declaration to 2 lines, where one line is the declaration alone, and the second one is an assignment, and we already handles assignment of `reshape`.

   ```fortran
   real :: b(3, 2) = reshape([-1, -3, 6, 7, 8, 9], [3, 2])
   
   
   ! the above will be converted to the below in init_expr
   real :: b(3, 2) 
   b = reshape([-1, -3, 6, 7, 8, 9], [3, 2])
   ```
   

Do you think I should change the implementation to be like the above option? Or leave it as in this PR? @certik


src/libasr/pass/init_expr.cpp
```diff
@@ -96,6 +96,21 @@ class ReplaceInitExpr: public ASR::BaseExprReplacer<ReplaceInitExpr> {
...
+        // If the array is a Var, get past it to it's value
+        if (ASR::is_a<ASR::Var_t>(*x->m_array)) {
+            ASR::Variable_t *arr_variable = ASRUtils::EXPR2VAR(x->m_array);
+            if (arr_variable != nullptr) {
+                x->m_array = arr_variable->m_symbolic_value;
+            }
+        }
+
+        BaseExprReplacer<ReplaceInitExpr>::replace_ArrayReshape(x);
+    }
```
certik, line 112: Can you explain what exactly is this doing? Is this assigning the reshape as an assignment? Or something else.

akramhany: No, this assigns and casts each value of the reshaped array separately to target array.

For example, in this Fortran code:
```fortran
program reshape_arr
    implicit none
    real :: d(1) = reshape([7], [1])
end program reshape_arr
```

The result of the `init_expr` pass is this:
```sh
;; ASR after applying the pass: init_expr
(TranslationUnit
    (SymbolTable
        1
        {
            assumed_rank:
                (Program
                    (SymbolTable
                        2
                        {
                            __1_k:
                                (Variable
                                    2
                                    __1_k
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                    .false.
                                    .false.
                                    ()
                                    .false.
                                    .false.
                                ),
                            d:
                                (Variable
                                    2
                                    d
                                    []
                                    Local
                                    ()
                                    ()
                                    Save
                                    (Array
                                        (Real 4)
                                        (IntegerConstant 1 (Integer 4) Decimal))]
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                    .false.
                                    .false.
                                    ()
                                    .false.
                                    .false.
                                )
                        })
                    assumed_rank
                    []
                    [(Assignment
                        (Var 2 __1_k)
                        ()
                        .false.
                    )
                    (Assignment
                        (ArrayItem
                            (Var 2 d)
                            [(()
                            (Var 2 __1_k)
                            ())]
                            (Real 4)
                            RowMajor
                            ()
                        )
                        (Cast
                            (IntegerConstant 7 (Integer 4) Decimal)
                            IntegerToReal
                            (Real 4)
                            ()
                        )
                        ()
                        .false.
                    )]
                )
        })
    []
)
```

akramhany:
> I think it's fine to merge, the change is small.
> 
> We likely need to improve this further, but for now it gets us unblocked.

I am actually willing to refactor this or even change the whole implementation if needed, I am just not fully sure what the correct design for this should be.

PR #8427, opened 2025-08-23
refactor: using functions which take explicit `llvm::Type*` and use create_gep2

sriganeshres:
towards: #7836


certik:
Let's wait for the full CI, then we can merge.

PR #8428, opened 2025-08-23
Solved the error of Implied loop array print

HiteshGhanchi:
```
program test_char
  implicit none
  character(len=5) :: s = "Hello"
  integer :: i
  print *, (s, i = 1, 3)
end program test_char
```
Initial Output :
```
Hello
```

New Output:
```
Hello    Hello    Hello
```
I have added this case as test case too.
This error was only visible for arrays created due to implied loop for strings.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -4655,7 +4655,8 @@ public:
 
             tmp = ASR::make_Print_t(al, x.base.base.loc, string_format);
         } else if (!fmt && body.size() == 1
-                        && ASR::is_a<ASR::String_t>(*ASRUtils::expr_type(body[0]))) {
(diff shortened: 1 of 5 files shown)


kmr-srbh:
Thanks for working on fixing this @HiteshGhanchi. Could you please open up an issue stating the bug, and then link this PR as a fix to it? That way we can keep track of the bugs we found and the changes that fixed them easily.

kmr-srbh:
@assem2002 could you please review the changes here? I think you implemented printing an implied do loop containing string.

HiteshGhanchi:
Fixes #8442

HiteshGhanchi:
> Thanks for working on fixing this @HiteshGhanchi. Could you please open up an issue stating the bug, and then link this PR as a fix to it? That way we can keep track of the bugs we found and the changes that fixed them easily.

Done

PR #8429, opened 2025-08-23
fix: reallocating lhs even when flag is false

xaerru:
Fix `--realloc-lhs` bug in `array_op.cpp`.

1. First commit is addressing comments from #8367, moving `ArraySize` creation to the pass from the backend.
2. The integration tests to which I have added `EXTRA_ARGS --realloc-lhs` fail without it with a runtime bounds check error.


tests/reference/run-array_bounds_check_03-15fcd63.stderr
```diff
@@ -0,0 +1,3 @@
+Runtime Error: Size mismatch in assignment to 'x'
```
certik, line 1: Later: this should say "allocate before assigning, or use `--realloc-lhs`", per https://github.com/lfortran/lfortran/issues/8430.

tests/reference/run-array_bounds_check_03-15fcd63.stderr
```diff
@@ -0,0 +1,3 @@
...
+
+LHS size is 1 and RHS size is 3
```
certik, line 3: I think this is incorrect -- the LHS is not allocated, so it is not size 1. It should say "LHS is not allocated".

tests/reference/run-array_bounds_check_01-713dbcf.stderr
```diff
@@ -0,0 +1,3 @@
+Runtime Error: Size mismatch in assignment to 'x'
+
+LHS size is 10 and RHS size is 3
```
certik, line 3: Later: another candidate for https://github.com/lfortran/lfortran/issues/8430, it should add "Allocate LHS to the right size, or use `--realloc-lhs`".

tests/reference/run-array_bounds_check_01-713dbcf.json
```diff
@@ -7,7 +7,7 @@
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
-    "stderr": null,
-    "stderr_hash": null,
-    "returncode": 0
+    "stderr": "run-array_bounds_check_01-713dbcf.stderr",
+    "stderr_hash": "1e4d1c683fef65bfb4cdd657ef165182b8199fba5a14356fed395eac",
+    "returncode": 1
 }
```
certik, line 13: Let's add tests for these that they pass with `--realloc-lhs`. I tested it locally by hand, it works, so we just need to test it also, to lock in the win.

ci/test_third_party_codes.sh
```diff
@@ -704,7 +704,7 @@ time_section "🧪 Testing fastGPT" '
 
         mkdir lf
         cd lf
-        FC=$FC CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS -DCMAKE_BUILD_TYPE=Debug ..
+        FC="$FC --no-array-bounds-checking --realloc-lhs" CMAKE_PREFIX_PATH=$CONDA_PREFIX cmake -DFASTGPT_BLAS=OpenBLAS -DCMAKE_BUILD_TYPE=Debug ..
```
certik, line 707: This is likely a bug in this PR: fastGPT is written to pass with bounds checking on, and to work without `--realloc-lhs`.

certik: I debugged fastGPT, and sure enough there is a bug there:
```
tokens = merge_pair(tokens, merge_pair_idx)
```
The merge_pair returns an allocatable array that is one shorter than the input. The error from LFortran is:
```
Runtime Error: Size mismatch in assignment to 'tokens'

LHS size is 4 and RHS size is 3
```
Which is the correct error, the original `tokens` is size 4, the new `tokens` is size 3.

I have to fix fastGPT.

certik: Everything works with a simple workaround: https://github.com/certik/fastGPT/commit/342b4f9425a69c63de6de0a550abf442bb00bc6a.

However, now I think we actually have a bug in this PR. It has to do with the fact when a function returns an allocatable array, and we are assigning it to an allocatable array. I wonder if the LHS gets deallocated, kind of like for `intent(out)` array, and then allocated from inside the function, effectively.

certik: On further investigation, here is the MRE:
```
program test

allocate(arr(3))
arr = [1, 2, 3]
arr = double_array(arr)  ! Safe: arr is read before being reallocated
print *, arr

contains
    function double_array(input) result(output)
    integer, intent(in) :: input(:)
    integer, allocatable :: output(:)

    ! Always works:
    !allocate(output(size(input)))
    !output = input(:size(input)) * 2

    ! Only works with `--realloc-lhs`:
    allocate(output(size(input)-1))
    output = input(:size(input)-1) * 2
    end function
end program
```
This does *not* work before F2003, since there the LHS must always be allocated to the correct size before the assignment, no matter what the RHS is (array or function result, whether allocatable or not). So without `--realloc-lhs`, this must indeed fail at runtime with a bounds check error. With `--realloc-lhs`, the LHS gets allocated to the function result.

Note: Pre F2003 this would also fail for another reason --- functions were not allowed to return allocatable arrays, which is a separate issue, although obviously related.

certik:
I am testing it here: https://github.com/lfortran/lfortran/pull/8435.

ci/test_third_party_codes.sh
certik: There is another problem: if we require the LHS to be allocated and the right size, then we can't use functions that return allocatable arrays at all, unless you know the size of the result ahead of time, but in that case the function shouldn't have been allocatable, it should have just computed the size of the result and use non-allocatable array. Allocatable result is for cases where you do not know the size ahead of time.

I noticed this PR actually allows an escape hatch: if the LHS is not allocated at all, it will allocate it to the function result. If it is already allocated, then it must be allocated to the correct size, otherwise it fails.

So this works:
```
program test
integer, allocatable :: arr(:), arr2(:)

allocate(arr(3))
arr = [1, 2, 3]
arr2 = double_array(arr)
print *, arr2

contains
    function double_array(input) result(output)
    integer, intent(in) :: input(:)
    integer, allocatable :: output(:)

    allocate(output(size(input)-1))
    output = input(:size(input)-1) * 2
    end function
end program
```

certik:
I forgot the exhaustive checks were not run --- I think they fail.

I first merged with master, to see if it is the merge that fails, but I then reverted the merge, I think it actually fails with your original branch also. That way it's cleaner to debug.

certik:
The CI passed, so I rebased the PR, since it was a few days old.

PR #8435, opened 2025-08-24
fix: reallocating lhs even when flag is false

certik:
The same as https://github.com/lfortran/lfortran/pull/8429, but removing `--no-array-bounds-checking` from fastGPT, as it should not be needed.


certik:
I am getting:
```
Errors while running CTest
	412 - format_20_FAST (SEGFAULT)
```
Which almost seems unrelated.

PR #8443, opened 2025-08-25
fix : Load all the dependent submodules while loading a module through load module function in global scope

### LINKED ISSUE #8436: Submodules : Submodules are not getting loaded in program file in `monolithic compilation` when parent module is a dependency module for a module being loaded
Towards #536 

modules.f90
```.f90
module stdlib_linalg
  implicit none

  interface diag
      module function diag_rsp_mat(A) result(res)
        real, intent(in) :: A(:,:)
        real :: res(minval(shape(A)))
      end function diag_rsp_mat
  end interface
end module stdlib_linalg


module stdlib_stats_corr
  use stdlib_linalg, only: diag
  implicit none

  interface corr
      procedure corr_2_rsp_rsp
  end interface corr

contains

  module function corr_2_rsp_rsp(x, dim) result(res)
    real, intent(in) :: x(:, :)
    integer, intent(in) :: dim
    real :: res(2,2)
    real :: mean_(2)
    mean_ = diag(res)
  end function corr_2_rsp_rsp
end module
```

submodule.f90
```.f90
submodule (stdlib_linalg) stdlib_linalg_diag
  implicit none

contains

  module function diag_rsp_mat(A) result(res)
    real, intent(in) :: A(:,:)
    real :: res(minval(shape(A)))
    integer :: i
    do i = 1, minval(shape(A))
      res(i) = A(i,i)
    end do
  end function diag_rsp_mat
end submodule
```

program.f90
```.f90
program example_corr
  use stdlib_stats_corr, only: corr
  implicit none
  real :: y(2, 3)
  print *, corr(y, 2)           
end program example_corr
```

Steps to reproduce the Error :- 
```console
lfortran modules.f90 -c
lfortran submodule.f90 -c
lfortran program.f90
```

LFortran Error :-
```console
/usr/bin/ld: /tmp/c.tmp_gSm7CsUBxfju1Li3eUz4br.o: in function `__module_stdlib_stats_corr_corr_2_rsp_rsp_real_______0':
LFortran:(.text+0x3a9): undefined reference to `__module_stdlib_linalg_diag_rsp_mat_real_______0'
The command 'clang -o c.out /tmp/c.tmp_gSm7CsUBxfju1Li3eUz4br.o  -L"/home/lordansh/dev/lfort1/lfortran/src/bin/../runtime" -Wl,-rpath,"/home/lordansh/dev/lfort1/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
```

AnshMehta1:
Fixes #8436

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2615,6 +2615,7 @@ RUN(NAME separate_compilation_21 LABELS gfortran llvm_submodule EXTRAFILES separ
 RUN(NAME separate_compilation_22 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_22a.f90 separate_compilation_22b.f90)
 RUN(NAME separate_compilation_23 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_23a.f90 separate_compilation_23b.f90)
 RUN(NAME separate_compilation_24 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_24a.f90 separate_compilation_24b.f90)
+RUN(NAME separate_compilation_25 LABELS gfortran llvm_submodule EXTRAFILES separate_compilation_25a.f90 separate_compilation_25b.f90)
 
 
 
--- a/integration_tests/separate_compilation_25.f90
+++ b/integration_tests/separate_compilation_25.f90
@@ -1,0 +1,11 @@
+program separate_compilation_25
+  use stats_corr_separate_compilation_25, only: corr
+  implicit none
+
+  real :: y(2,2) = reshape([1.0, 2.0, 3.0, 4.0], [2,2])
+  real :: result(2)
+
+  result = corr(y)
+  print *, result
+  if (.not. all(result == [1.0, 4.0])) error stop
+end program
--- a/integration_tests/separate_compilation_25a.f90
+++ b/integration_tests/separate_compilation_25a.f90
@@ -1,0 +1,28 @@
+module linalg_separate_compilation_25
+  implicit none
+
+  interface diag
+      module function diag_real(A) result(res)
+        real, intent(in) :: A(:,:)
+        real :: res(minval(shape(A)))
+      end function diag_real
+  end interface
+end module linalg_separate_compilation_25
+
+
+module stats_corr_separate_compilation_25
+  use linalg_separate_compilation_25, only: diag
+  implicit none
+
+  interface corr
+      procedure corr_real
+  end interface corr
+
+contains
+
+  module function corr_real(x) result(res)
+    real, intent(in) :: x(:, :)
+    real :: res(2)
+    res = diag(x)
+  end function corr_real
+end module
--- a/integration_tests/separate_compilation_25b.f90
+++ b/integration_tests/separate_compilation_25b.f90
@@ -1,0 +1,14 @@
+submodule (linalg_separate_compilation_25) linalg_diag_separate_compilation_25
+  implicit none
+
+contains
+
+  module function diag_real(A) result(res)
+    real, intent(in) :: A(:,:)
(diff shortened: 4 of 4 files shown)


src/libasr/asr_utils.cpp
```diff
@@ -19,6 +19,8 @@ namespace LCompilers {
 
     namespace ASRUtils  {
 
+std::set<std::string> loaded_submodules;
+
```
certik: Let's not use global variables, we need to be thread safe. We are currently not always, but close. This moves us away.

AnshMehta1: Actually, I want a global set to track modules whose submodules are loaded. Is their any other way to do that?

certik: Yes, you have to pass the variable into the visitors and pass it from the symbol table to the body visitor. However are we loading modules in the body visitor at all?

AnshMehta1: Let me Explain the situation.

So, we just need `std::set<std::string> loaded_submodules` in `ast_symboltable_visitor.cpp` as we load submodules only when we encounter use statements which are handled only in sybmoltable visitor.
So, we somehow need to pass this set to `load_module` function. 

But the problem is we use `load_module` function in multiple files. So, just adding this set to arguments of `load_module` function won't work as in other files, we have nothing to pass to satisfy set argument.

certik: Isn't load_module only used in the body/symbol table visitors?

AnshMehta1: No, currently, we are using it at multiple places like 
image

certik: I think those are all used from the two visitors in every usage, right?

So you just have to pass all the information via arguments in the call chain.

AnshMehta1: `Openmp.cpp` and `pass_utils.cpp` don't have access to declarations made in symboltable visitor or ast common visitor.

certik: So those are creating modules on the fly. In those we should not use submodules, to keep things simple.

So can you pass in an empty list to those?

Alternatively, you just need to maintain a list of submodules in ASR? One workaround is to just write a function in utils that gives you this list when you give it a TranslationUnit.

Goal:

* ASR must be self-sufficient, not additional information should be needed to operate on it
* No global variables

AnshMehta1: I think the idea of empty list is good enough and simple. Will implement it.

src/libasr/asr_utils.cpp
```diff
@@ -1097,6 +1085,52 @@ void load_dependent_submodules(Allocator &al, SymbolTable *symtab,
...
+                        if (item == "iso_c_binding"
+                            ||item == "iso_fortran_env") {
+                            Result<ASR::TranslationUnit_t*, ErrorMessage> res
+                                = find_and_load_module(al, "lfortran_intrinsic_" + item,
+                                *symtab, true, pass_options, lm);
+                            if (res.ok) {
+                                mod1 = res.result;
+                            } else {
+                                error_message =  res.error.message;
+                            }
```
certik, line 1117: What is going on here --- this should not be here. In src/libasr we need to be frontend independent.

If iso_c_binding requires a special handling, then filter this in the caller, or figure out some other solution.

AnshMehta1: Its just we trying to load special intrinsic modules.

Actually, I have copied this code from `load_module` function (just above this function) where we were doing same thing for loading dependent modules. Their also, our purpose was same.

certik: Ok, I made this https://github.com/lfortran/lfortran/issues/8477, let's fix it later.

PR #8457, opened 2025-08-26
Migrate some CreateLoad -> CreateLoad2

certik:
Towards #7836.


certik:
It looks like all integration tests pass, but 3rd party tests fail. So this is a great opportunity to create MREs from the failures and improve our integration tests. I think some of the paths in the backend are not tested in integration tests, so we need to fix it.

PR #8458, opened 2025-08-26
Migrate the rest of CreateLoad->CreateLoad2

certik:
Building on top of https://github.com/lfortran/lfortran/pull/8457.

This migrates all CreateLoad in asr_to_llvm. All integration tests pass on LLVM 11 and LLVM 20, but it seems some 3rd party codes fail.


sriganeshres:
You are getting the i8* not compatible with void* issue.

PR #8462, opened 2025-08-27
Add LFortran specific test (`_lfortran_str`)

### LINKED ISSUE #8439: Add tests for LPython functionality
In https://github.com/lfortran/lfortran/pull/8418/files#diff-48a7161397f9eb8b730f0cbec2f11ce1471aac961dc6b16dc558cf161c375054

This  likely corresponds to:
```

i = 3
a = str(i)
print (type(a)) # return str type
```
in LPython. For now we can add a new `_lfortran_str` intrinsic and test it like:
```
integer :: i = 3
character(:), allocatable :: a
a = _lfortran_str(i)
```
That should do it.

assem2002:
- Registered `_lfortran_str` as an LFortran specific function, To avoid treating it as wrong functionCall
- visiting `FuncCallOrArray` of `_lfortran_str` results in `ASR::Cast` node
- Refactored LLVM implementation of the `IntegerToString` cast + `RealToString` cast (Created our string representation and copied the runtime-function-call return into it + freed the return)

closes #8439

### DIFF
--- a/integration_tests/test_str.f90
+++ b/integration_tests/test_str.f90
@@ -1,0 +1,48 @@
+! Testing Python's str casting
+! -- Example --
+! i = 10
+! s = "Hi"
+! s = str(i) # Equivalent in LFortran is `lfortran_str()`
+
+program test_str
+    character(30) :: str
+    integer :: i
+    real  :: r
+
+    i = 11
+    str =  "HI"//_lfortran_str(9) // _lfortran_str(i)
+    print *, str, "|"
+    if(trim(str) /= "HI911") error stop
+  ! --------------------------------------------------------------------------- !
+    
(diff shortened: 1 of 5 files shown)


assem2002:
I think those changes shouldn't break LPython at all.

integration_tests/test_str.f90
```diff
@@ -0,0 +1,48 @@
+! Testing Python's str casting
+! -- Example --
+! i = 10
+! s = "Hi"
+! s = str(i) # Equivalent in LFortran is `lfortran_str()`
```
certik, line 5:
```suggestion
! s = str(i) # Equivalent in LFortran is `_lfortran_str()`
```

PR #8467, opened 2025-08-27
feat: add `vptr` to LLVM structure declaration for structs and classes

kmr-srbh:
A new dummy `vtable` implementation is also added. This in line with keeping the new class design separate from the old one to simplify the refactor.


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4456,6 +4489,30 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+        llvm::Type* vtable_type = llvm_utils->struct_vtable.at(struct_sym);
+        llvm::Value* vtable_obj = llvm_utils->CreateAlloca(*builder, vtable_type);
+
+        llvm::Value* struct_type_hash_ptr = llvm_utils->create_gep2(vtable_type, vtable_obj, 0);
+        llvm::Value* struct_type_hash = llvm::ConstantInt::get(
+            llvm_utils->getIntType(8), llvm::APInt(64, get_class_hash(struct_sym)));
+        builder->CreateStore(struct_type_hash, struct_type_hash_ptr);
+    }
+
```
kmr-srbh, line 4515: We discussed that the class hash will collide in separate compilation as it is scope specific. We will have to use a different hashing technique to fix it. The above implementation currently uses the hashing method from the old design only. Also, it is not used currently apart from just being set in the structure. It will be addressed in future PRs.

certik: How does Clang handle this issue?

Pranavchiku (commented):
The change is incomplete, with following diff:

```diff
% git diff
diff --git a/tests/classes2.f90 b/tests/classes2.f90
index e35edff15..d6dcad0a4 100644
--- a/tests/classes2.f90
+++ b/tests/classes2.f90
@@ -41,5 +41,6 @@ implicit none
     type(point2d), target :: p2d = point2d( 3, 4 )
 
     ptr => p2d
-    ! res = ptr%radius()
+    res = ptr%radius()
+    print *, "Radius: ", res
 end program main
```

```
% gfortran ./tests/classes2.f90 && ./a.out   
 Radius:    5.00000000    
% lfortran ./tests/classes2.f90 --new-classes
Radius:     9.06830374e-29
```

tests/reference/llvm_new_classes-classes1-bfe507d.stdout
```diff
@@ -2,8 +2,8 @@
 source_filename = "LFortran"
 
 %string_descriptor = type <{ i8*, i64 }>
-%base = type <{ i32 }>
-%__vtab_base = type { i64 }
+%base = type <{ %__new_vtab_base*, i32 }>
+%__new_vtab_base = type { i64 }
```
Pranavchiku, line 6: this is an improvement for sure, towards aligning with what clang does but incomplete which can be done in subsequent PRs.

src/libasr/codegen/asr_to_llvm.cpp
Pranavchiku: Let's discuss this in tomorrow's call.

Issue #8448, opened 2025-08-26
Julienne build error with lfortran 0.56.0


rouson:
```
git clone -b lfortran git@github.com:berkeleylab/julienne
cd julienne
fpm test --compiler lfortran --flag --cpp
```
which yields the following trailing output:
```
[ 15%]         julienne_formats_s.F90
sh: line 1:  5971 Segmentation fault: 11  lfortran -c ././src/julienne/julienne_formats_s.F90 --cpp -I././include -Ibuild/dependencies/assert/include -J build/lfortran_3A22F11D36F35A40 -Ibuild/lfortran_3A22F11D36F35A40 -o build/lfortran_3A22F11D36F35A40/julienne/src_julienne_julienne_formats_s.F90.o > build/lfortran_3A22F11D36F35A40/julienne/src_julienne_julienne_formats_s.F90.o.log 2>&1
[ 16%]         julienne_formats_s.F90  done.

<ERROR> Compilation failed for object " src_julienne_julienne_formats_s.F90.o "
<ERROR> stopping due to failed compilation
STOP 1
```
The above file, `src/julienne/julienne_formats_s.F90` is pretty short.  In case it helps, Julienne's [include/language-support.F90[ shows some of the ways we're already attempting other work around issues in compilers, including lfortran.

certik:
Awesome, thanks! We'll fix it.

bonachea:
This issue appears to be subsumed by #7812 and should probably be closed.

bonachea:
Julienne has been working for awhile now, and https://github.com/lfortran/lfortran/pull/12148 integrates the latest Julienne release into third-party CI

PR #8475, opened 2025-08-28
fix: convert procedure name to lower case

akramhany:
Towards #8448

I haven't made an integration test as I thought it might be trivial, should I make one?

### DIFF
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -4783,3 +4783,7 @@ llvm_new_classes = true
 [[test]]
 filename = "implied_do_loop2.f90"
 run = true
+
+[[test]]
+filename = "errors/submodule_01.f90"
+asr = true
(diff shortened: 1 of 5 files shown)


assem2002:
We better have an integration test for each case and program flow to guarantee our integration tests are solid, and to depend less on 3rd party codes (as you see extracting MREs from them is pain)

tests/tests.toml
```diff
@@ -4783,3 +4783,7 @@ llvm_new_classes = true
 [[test]]
 filename = "implied_do_loop2.f90"
 run = true
+
+[[test]]
+filename = "errors/submodule_01.f90"
+asr = true
```
certik, line 4789: We should later figure out how to merge this error message with the continue compilation ones.

Issue #1504, opened 2023-04-06
Complex error


fxcoudert:
```
program test
  complex, parameter :: x(2) = (1.0, 0.1)
  print *, x%re, x%im
end
```

gives

```
Compilation time: 59 ms
Execution time: NaN ms
CompileError: WebAssembly.Module doesn't validate: attempt to use unknown local 1, the number of locals is 1, in function at index 1
ERROR: The code could not be executed. Either there is a runtime error or there is an issue at our end.
```

Shaikh-Ubaid:
It seems the example shared is an array of complex numbers, which I think is not yet supported.

Complex numbers are supported and the following example works:
```fortran
program test
  complex, parameter :: x = (1.0, 0.1)
  print *, x%re, x%im
end
```
Output:
```bash
1.000000000 0.10000000

Compilation time: 71 ms
Execution time: 0 ms
```

certik:
Indeed, an array is not supported yet. @fxcoudert thanks for reporting it, that's a bug.

harperjf:
Good news: the error message is clearer. Bad news: the bug is still there. Lfortran 0.56.0 output from @fxcoudert's program
```
syntax error: Token 'complex' is unexpected here
 --> cplxarray.f90:2:3
  |
2 |   complex, parameter :: x(2) = (1.0, 0.1)
  |   ^^^^^^^ 
```

PR #8481, opened 2025-08-30
Fix complex array member access (%re and %im)

ubaidsk:
fixes https://github.com/lfortran/lfortran/issues/1504

```shell
(lf) lfortran % cat examples/expr2.f90 
program test
  complex, parameter :: x(2) = (1.0, 0.1)
  print *, x%re, x%im
end%                                                                                                                  
(lf) lfortran % gfortran examples/expr2.f90 
(lf) lfortran % ./a.out 
   1.00000000       1.00000000      0.100000001      0.100000001    
(lf) lfortran % lfortran examples/expr2.f90
1.00000000e+00    1.00000000e+00    1.00000001e-01    1.00000001e-01
```

### DIFF
--- a/integration_tests/complex_array_member_access.f90
+++ b/integration_tests/complex_array_member_access.f90
@@ -16,5 +16,6 @@ program test
     if (abs(y_im(1) - 0.3) > 1e-5) error stop "Imaginary part 1 failed"
     if (abs(y_im(2) - 0.3) > 1e-5) error stop "Imaginary part 2 failed"
     
+    print *, x%re, x%im
     print *, "All tests passed!"
 end


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9302,44 +9302,36 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            im = complex_im(tmp, complex_type_4);
+            tmp = builder->CreateFPExt(im, llvm::Type::getDoubleTy(context));
+        } else if (arg_kind == 8 && dest_kind == 4) {
+            // complex(8) -> real(4)
+            im = complex_im(tmp, complex_type_8);
+            tmp = builder->CreateFPTrunc(im, llvm::Type::getFloatTy(context));
+        } else if (arg_kind == 8 && dest_kind == 8) {
+            // complex(8) -> real(8)
+            im = complex_im(tmp, complex_type_8);
+            tmp = im;
```
certik, line 9329: It seems that casting should be done with a casting node in ASR, not here. 

But we can fix that later.

Issue #8459, opened 2025-08-26
Design idea: Allow non-constant elements in FixedSizeArray


certik:
The issue https://github.com/lfortran/lfortran/issues/8452 brought up a design decision to make:

* Currently `FixedSizeArray` has been documented in https://github.com/certik/lfortran/blob/5c5797d711e6787a16ef4e50988276ade33ee8d8/doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md to only allow constant elements. So both the size and elements are compile-time constants.
* In the backend and in ASR however we currently allow `FixedSizeArray` to also have non-constant elements (see #8452 for an example). We still require compile-time constant size.

The proposal to consider:

* Allow `FixedSizeArray` to have non-constant elements. It must still have compile-time constant size.
* The `PointerToDataArray` array must have a size that is not a compile-time constant, it's a runtime expression. The elements consequently cannot be constants, since we do not even have access to them at compile time.
* The `ArrayConstant` must be `FixedSizeArray` physical type, and in addition it must have constant elements.
* The `ArrayConstructor` can now be `FixedSizeArray` but at least one element must not be constant.

I think the proposal is to effectively migrate `[a, b, c]` from `PointerToDataArray` to `FixedSizeArray`. I think currently:

* `FixedSizeArray`: constant size, constant elements
* `PointerToDataArray`: constant or runtime size, at least one element non-constant

And the proposal is:

* `FixedSizeArray`: constant size, elements can be constant or runtime
* `PointerToDataArray`: runtime size (cannot be constant), elements cannot be constant, since we do not even have access to them due to the runtime size

certik:
Note: in the new proposal
* The `PointerToDataArray` has runtime size, so it *must* be represented by a pointer, thus the name. 
* The `FixedSizeArray` has constant size, so it can be represented by a compile-time struct of N elements in LLVM. For small N this might be a good optional optimization. It could even be passed around by value. For now I think the backend still uses a regular pointer, but this could be done, for example if a function is expecting an array of size N=7, then we can pass in 7 runtime values and inline them as individual arguments.

So I think keeping the distinction of compile-time vs runtime *size* is good. The elements can always be runtime for both now. I think that's fine. Finally, ArrayConstant will allow the optimizer to just inline it everywhere, so it might not even be present as a physical type.

assem2002:
[...]

PR #8486, opened 2025-08-30
Doc: Clarify the new design for FixedSizeArray

certik:
Towards https://github.com/lfortran/lfortran/issues/8459.

### DIFF
--- a/doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md
+++ b/doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md
@@ -35,7 +35,24 @@ The return value is an array expression in the _new_ physical type. It's a messa
 
 ### Physical Types
 
-**PointerArray** : The size is not known at compile time, but we know an expression for it at runtime. This type is used for `bind(c)`; represented by just a pointer, the compiler knows expressions dimensions at compile time. We know the expression for all dimensions. The expression can contain runtime parameters. It is represented by just a pointer. The compiler knows the dimensions at compile time, but they are not stored at runtime explicitly. You can call size(A), the compiler will determine the size expression at compile time (using runtime variables).
+**PointerArray** : The size is not known at compile time, but we know an
+expression for it at runtime. This type is used for `bind(c)`; represented by
+just a pointer, the compiler knows expressions dimensions at compile time. We
+know the expression for all dimensions. The expression can contain runtime
+parameters. It is represented by just a pointer. The compiler knows the
+dimensions at compile time, but they are not stored at runtime explicitly. You
+can call size(A), the compiler will determine the size expression at compile
+time (using runtime variables).
+
+Any PointerArray can be cast to a DescriptorArray and DescriptorArray can in
+general be used without any additional code/logic in all places where
+PointerArray is used.
+
+The other direction is more complicated: It is possible to cast a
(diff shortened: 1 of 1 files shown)


doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md
```diff
@@ -51,29 +51,43 @@ real, intent(in) :: A(n, 2*m)
...
+LHS is PointerArray (can also be FixedSizeArray), RHS is FixedSizeArray
 ```fortran
 real :: x(3), a, b, c
-x = [a, b, c]  ! Both LHS and RHS are PointerArray, RHS is
+x = [a, b, c] ! LHS is PointerArray, RHS is FixedSizeArray
 ArrayConstructor
 ```
 
-**FixedSizeArray**: An array that is fully known at compile time: both size and elements. ArrayConstant. In LLVM backend we can store such an array as a variable with known size (say 10). (This type might in principle be possible to use for `A(10, 20)`, but currently we don't. It cannot be used for `A(n, m)`.). It seems we cannot change elements at runtime, they are all constant.
+**FixedSizeArray**: The size is known at compile time. It can contain constant or runtime elements (like variables). An array that is fully known at compile time: both size and elements. ArrayConstant. In LLVM backend we can store such an array as a variable with known size (say 10). (This type might in principle be possible to use for `A(10, 20)`, but currently we don't. It cannot be used for `A(n, m)`.). It seems we cannot change elements at runtime, they are all constant.
```
assem2002:
```suggestion
**FixedSizeArray**: The size is known at compile time. It can contain constant or runtime elements (like variables). 
```

assem2002 (commented):
That should be design yes. 
Left a suggestion

doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md
certik: Fixed in a commit below.

certik:
The CI failure is unrelated: https://github.com/lfortran/lfortran/issues/8493.

doc/src/asr/asr_nodes/expression_nodes/ArrayPhysicalCast.md
```diff
@@ -35,7 +35,24 @@ The return value is an array expression in the _new_ physical type. It's a messa
 
 ### Physical Types
 
-**PointerArray** : This type is used for `bind(c)`; represented by just a pointer, the compiler knows expressions dimensions at compile time. We know the expression for all dimensions. The expression can contain runtime parameters. It is represented by just a pointer. The compiler knows the dimensions at compile time, but they are not stored at runtime explicitly. You can call size(A), the compiler will determine the size expression at compile time (using runtime variables).
+**PointerArray** : The size is not known at compile time, but we know an
+expression for it at runtime. This type is used for `bind(c)`; represented by
```
Pranavchiku, line 39: this can also be used for non `bind(c)` cases.

Pranavchiku, line 39:
```suggestion
expression for it at runtime. This type is also used for `bind(c)`; represented by
```

certik: I am going to wait if the CI job passes first.

Issue #8491, opened 2025-08-30
Void* and i8* not matching in function signature.


sriganeshres:
refers: #7836, #8457 

```fortran
program cptr_example
    use iso_c_binding, only: c_ptr, c_loc, c_f_pointer, c_int
    implicit none

    integer, target :: arr(5) = [1, 2, 3, 4, 5]
    type(c_ptr) :: arr_ptr

    arr_ptr = c_loc(arr)

    call process_array(arr_ptr, size(arr))

contains

    subroutine process_array(ptr, n)
        use iso_c_binding, only: c_ptr, c_f_pointer, c_int
        type(c_ptr), intent(in) :: ptr
        integer, intent(in) :: n
        integer(c_int), pointer :: f_arr(:)
        integer :: i

        ! Convert C pointer back to Fortran pointer
        call c_f_pointer(ptr, f_arr, [n])

        print *, "Array elements from process_array:"
        do i = 1, n
            print *, "Element", i, "=", f_arr(i)
        end do
    end subroutine process_array

end program cptr_example
```

Output
```
asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %42 = load i8*, void** %arr_ptr, align 8
 void*  call void @process_array(i8* %42, i32* %call_arg_value)

code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %42 = load i8*, void** %arr_ptr, align 8
 void*  call void @process_array(i8* %42, i32* %call_arg_value)
```

When we change the void** loads into createload2.

sriganeshres:
In this we `process_array(void*, i32*)` but when we load it using direct type we make call to this `process_array(i8*, i32*)`

Issue #8453, opened 2025-08-26
Incorrect array array physical type in reshape, fails in LLVM when migrating gep->gep2


certik:
This MRE (created from `arrays_reshape_23.f90` where it fails in integration tests):
```fortran
program test
integer :: m = 3, n = 2
integer :: arr(3,2)
arr = reshape([1, 2, 3, 4, 5, 6], [m, n])
print *, arr
end program
```
Currently has the following ASR:
```
[(Assignment
                        (Var 2 arr)
                        (ArrayReshape
                            (ArrayConstant
                                24
                                [1, 2, 3, 4, 5, 6]
                                (Array
                                    (Integer 4)
                                    [((IntegerConstant 1 (Integer 4) Decimal)
                                    (IntegerConstant 6 (Integer 4) Decimal))]
                                    FixedSizeArray
                                )
                                ColMajor
                            )
                            (ArrayPhysicalCast
                                (ArrayConstructor
                                    [(Var 2 m)
                                    (Var 2 n)]
                                    (Array
                                        (Integer 4)
                                        [((IntegerConstant 1 (Integer 4) Decimal)
                                        (IntegerConstant 2 (Integer 4) Decimal))]
                                        FixedSizeArray
                                    )
                                    ()
                                    ColMajor
                                )
                                FixedSizeArray
                                DescriptorArray
                                (Array
                                    (Integer 4)
                                    [((IntegerConstant 1 (Integer 4) Decimal)
                                    (IntegerConstant 2 (Integer 4) Decimal))]
                                    DescriptorArray
                                )
                                ()
                            )
                            (Array
                                (Integer 4)
                                [(()
                                ())
                                (()
                                ())]
                                FixedSizeArray
                            )
                            ()
                        )
                        ()
                        .false.
                    )
```
The main bug is #8452, the type of the `ArrayConstructor` should be `PointerToDataArray`, not `FixedSizeArray`. This in turn makes the following fix in the LLVM backend to migrate from gep to gep2 fail:
```diff
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -6711,7 +6711,10 @@ public:
-                 value = llvm_utils->create_gep(value, 0);
+                llvm::Type* value_llvm_type = llvm_utils->get_type_from_ttype_t_util(x.m_value, value_type, module.get());
+                value = llvm_utils->create_gep2(value_llvm_type, value, 0);
```
[...]

PR #8492, opened 2025-08-30
fixed: makde a void* to fix createload2

sriganeshres:
towards: #7836 
fixes: #8491


sriganeshres:
@certik if this passes then we can rebase #8457 on to this and i think rest are fine.

certik:
@sriganeshres go ahead and try https://github.com/lfortran/lfortran/pull/8457. I still think we need to fix https://github.com/lfortran/lfortran/issues/8453#issuecomment-3238938789, but you can skip that one case and see if everything else works.

Issue #8276, opened 2025-08-07
object oriented design in clang vs LFortran


Pranavchiku:
I'll be write `Fortran` examples, generate their LLVM using LFortran and compare LLVM generated by clang on corresponding `cpp` example. This is an understanding of what should be the design of object oriented code.

For clang we use the following command:

```
clang -S -emit-llvm a.cpp -o a.ll
```

```fortran
module xx
    type :: base
        integer :: i
    end type base
end module

program main
    use xx
    implicit none

    type(base) :: b
end program
```

```llvm
; ModuleID = 'LFortran'
source_filename = "LFortran"

%__vtab_base = type { i64 }
%base = type <{ i32 }>

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %2 = alloca %__vtab_base, align 8
  %3 = getelementptr %__vtab_base, %__vtab_base* %2, i32 0, i32 0
  store i64 0, i64* %3, align 4
  %b = alloca %base, align 8
  %4 = getelementptr %base, %base* %b, i32 0, i32 0
  call void @_lpython_free_argv()
  br label %return

return:                                           ; preds = %.entry
  ret i32 0
}

declare void @_lpython_call_initial_functions(i32, i8**)
```

```cpp
// #include <iostream>

class Base {
    int x;
};

int main() {
    Base b;
    // std::cout << "Base class created." << std::endl;
    return 0;
}
```

```llvm
; ModuleID = 'a.cpp'
source_filename = "a.cpp"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128"
target triple = "arm64-apple-macosx14.0.0"

%class.Base = type { i32 }

; Function Attrs: noinline norecurse nounwind optnone ssp uwtable(sync)
define i32 @main() #0 {
  %1 = alloca i32, align 4
  %2 = alloca %class.Base, align 4
  store i32 0, ptr %1, align 4
  ret i32 0
}

[...]

PR #8498, opened 2025-08-30
feat: add `type_info` object for structs to support `dynamic_cast()` implementation for `select type` construct

kmr-srbh:
Towards #8276

### DIFF
--- a/tests/reference/llvm_new_classes-classes1-bfe507d.json
+++ b/tests/reference/llvm_new_classes-classes1-bfe507d.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm_new_classes-classes1-bfe507d.stdout",
-    "stdout_hash": "f62d81154da3e57ec6d614d0d664e22d59c4ad421dc8f4edbe2cdcbd",
+    "stdout_hash": "9e31876470343d6a14bd00ee5c522b1e03411e3949fea02ab6f40b73",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm_new_classes-classes1-bfe507d.stdout
+++ b/tests/reference/llvm_new_classes-classes1-bfe507d.stdout
@@ -1,11 +1,12 @@
 ; ModuleID = 'LFortran'
 source_filename = "LFortran"
 
-%_VTable_base.type = type { [3 x i8*] }
 %string_descriptor = type <{ i8*, i64 }>
 %base = type <{ i32 (...)**, i32 }>
 
-@_VTable_base = linkonce_odr unnamed_addr constant %_VTable_base.type { [3 x i8*] [i8* null, i8* null, i8* bitcast (void (%base*)* @__module_xx_show_x to i8*)] }, align 8
+@_Name_base = private unnamed_addr constant [5 x i8] c"base\00", align 1
+@_Type_Info_base = linkonce_odr unnamed_addr constant { i8*, i8* } { i8* getelementptr inbounds ([5 x i8], [5 x i8]* @_Name_base, i32 0, i32 0) }, align 8
+@_VTable_base = linkonce_odr unnamed_addr constant { [3 x i8*] } { [3 x i8*] [i8* null, i8* @_Type_Info_base, i8* bitcast (void (%base*)* @__module_xx_show_x to i8*)] }, align 8
 @0 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
 @serialization_info = private unnamed_addr constant [3 x i8] c"I4\00", align 1
 @1 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1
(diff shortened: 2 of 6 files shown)


src/libasr/runtime/type_info.cpp
kmr-srbh, line 1: The ABI implementation is much more complex in `clang`. For now, I have implemented only the logic required for handling single-inheritance. Supporting multiple-inheritance will require adding some new functions and extending the logic in `lfortran_dynamic_cast()` to support the graph algorithm used by `clang`, so the extension is easy when we want to support this in future.

kmr-srbh:
Ready.

certik:
This makes our runtime library C++ instead of C. That's a big design change that we need to consider.

I would not do that for many reasons: smaller runtime, no C++ dependency, smaller wasm, more control, easier to isolate the system API, etc. Eventually we want to get to the point that we do not even ship a runtime if not needed. For the inheritance I think we don't need any system API, so this logic should eventually be part of ASR itself, or ship it as a simple C library.

If this PR works, let's rewrite the logic to C and add it to our existing C file, that way you don't need to modify our build system, WASM related stuff, etc.

kmr-srbh:
Yes, the CPP addition was a concern to me as well. Having this logic as part of the ASR or LLVM phase would be great! The `clang` approach uses a C++ specific library - `typeinfo` from whom the other type-info classes inherit. I will try to rewrite this logic into C and see how it all lands.

jinangshah21:
Here, `dynamic_cast` returns true even if one type is parent of other, correct?

kmr-srbh:
> Here, `dynamic_cast` returns true even if one type is parent of other, correct?

Yes. `select type` handles down casting from parent to derived. So class of object X in `select type(X)` should be Y or an extension of class Y in `class is(Y)`. So in the implementation, we always search up in the inheritance chain for the destination type (class Y).

jinangshah21:
But in `type is(..)` statements we need to allow only if dynamic type is exactly matched. (Type derived from that are not allowed)

kmr-srbh:
> But in `type is(..)` statements we need to allow only if dynamic type is exactly matched. (Type derived from that are not allowed)

Yes, for `type is (...)`, we use a boolean flag `match_exact_type` to only compare the current type-info objects. We then return the value of that comparison and do not check the inheritance chain.

certik:
> Yes, the CPP addition was a concern to me as well. Having this logic as part of the ASR or LLVM phase would be great! The `clang` approach uses a C++ specific library - `typeinfo` from whom the other type-info classes inherit. I will try to rewrite this logic into C and see how it all lands.

Yes, just rewrite it to C and put it into our existing C runtime file. That will be a good start.

Pranavchiku (approved):
logic looks good, let's isolate changes which are not required in this PR and move .cpp to runtime library then we should be good to merge! Very good!

kmr-srbh:
Ready.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5684,3 +5685,76 @@ LFORTRAN_API char *_lfortran_get_env_variable(char *name) {
...
+
+
+
+/*
+    Section below contains code for Run-Time Type Information (RTTI)
+    implementation used by LFortran.
+
+    For general runtime functions, please add them above this section. Please do 
+    not add any unrelated changes below.
+*/
```
kmr-srbh, line 5698: Added comment for the sake of separation of logic. I can remove it if it feels awkward.

certik:
I think you need to update reference results?

integration_tests/CMakeLists.txt
```diff
@@ -1905,17 +1906,17 @@ RUN(NAME class_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes
...
 RUN(NAME class_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
-RUN(NAME class_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
-RUN(NAME class_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
```
certik, line 1919: This is a regression, check with @jinangshah21 if this is ok.

jinangshah21: @kmr-srbh you can rebase and check if `class_19.f90` works or not?

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5684,3 +5685,76 @@ LFORTRAN_API char *_lfortran_get_env_variable(char *name) {
...
+    if (is_equal((const type_info*) dynamic_type, (const type_info*) dst_type)) {
+        dst_ptr = dynamic_ptr;
+    } else if (!match_exact_type) {
+        if (search_dst_type((const __si_class_type_info*) dynamic_type, dst_type)) {
+            dst_ptr = dynamic_ptr;
+        }
+    }
+
+    return (void*) dst_ptr;
+}
```
certik, line 5760: This is great. Much simpler than the C++ version.

Can you document what each function does and when it should be called by the backend? You understand exactly what it does. If this is documented well somewhere else, you can also just point to it, but I can't find this function `search_dst_type` in `https://itanium-cxx-abi.github.io/cxx-abi/abi.html#rtti`, so it should be documented here.

jinangshah21 (approved):
I think this is fine.
We need to store type info correctly when there are no functions in derived type which can be handled in upcoming PRs.

integration_tests/CMakeLists.txt
kmr-srbh: The bus error exists Jinang. I think it is due to changes in this PR, where I touched some of the allocation logic. I found some LLVM instructions are not getting generated now, before call to `store_class_vptr()`. I will try to handle it in the upcoming PR.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: Yes, will add comments for this in the next PR.

PR #8504, opened 2025-08-31
Fix implicit cast for complex numbers

ubaidsk:
fixes https://github.com/lfortran/lfortran/pull/8481#discussion_r2311758687

## Test Plan

- New test added ✅ 
- Existing integration and reference tests pass ✅


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3394,6 +3394,25 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     al, loc, y, im, ASRUtils::expr_type(target), nullptr));
                 value = cmplx;
             }
+        } else if ( ASR::is_a<ASR::ComplexRe_t>(*target) ) {
```
ubaidsk, line 3397: This is implemented below for ComplexIm, but wasn't implemented for ComplexRe. We are fixing it in this PR.

PR #8506, opened 2025-09-01
enh: [new classes] add support for runtime subroutine polymorphism

jinangshah21:
Towards #7068

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3169,7 +3169,11 @@ public:
         } else {
             ptr_loads = 2 - LLVM::is_llvm_pointer(*x_m_v_type);
         }
-        this->visit_expr(*x.m_v);
+        if (compiler_options.new_classes) {
+            this->visit_expr_wrapper(x.m_v, LLVM::is_llvm_pointer(*x_m_v_type));
+        } else {
+            this->visit_expr(*x.m_v);
+        }
         ptr_loads = ptr_loads_copy;
         if (ASRUtils::is_unlimited_polymorphic_type(x.m_v)) {
             if( current_select_type_block_type ) {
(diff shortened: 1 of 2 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12376,6 +12382,57 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            llvm::PointerType *fnPtrPtrPtrTy = llvm::PointerType::get(fnPtrPtrTy, 0);
+
+            // Get VTable pointer
+            llvm::Value* vtable_ptr = builder->CreateBitCast(llvm_dt, fnPtrPtrPtrTy);
+            vtable_ptr = llvm_utils->CreateLoad2(fnPtrPtrTy, vtable_ptr);
+#if LLVM_VERSION_MAJOR > 16
+            ptr_type_deprecated[vtable_ptr] = fnPtrPtrTy;
+#endif
+            // Get function pointer from VTable
+            llvm::Value* fn = (llvm_utils->create_ptr_gep_deprecated(
```
certik: Let's not use deprecated functions, use the create_ptr_gep2 functions.

certik:
The CI fails, so I made it a draft for now.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4187,20 +4194,38 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                return ;
+            }
+            llvm::Type* v_llvm_type = llvm_utils->get_type_from_ttype_t_util(
+                    v->m_type, v->m_type_declaration, module.get());
             if (LLVM::is_llvm_pointer(*v->m_type)) {
-                ptr = llvm_utils->CreateLoad2(llvm_utils->get_type_from_ttype_t_util(
-                    v->m_type, v->m_type_declaration, module.get()), ptr);
+                if (is_unallocated) {
+                    llvm::Value* vptr = builder->CreateAlloca(llvm_utils->vptr_type);
+                    builder->CreateStore(builder->CreateBitCast(vptr, v_llvm_type), ptr);
```
jinangshah21, line 4220: See `class_19.f90` for above change, as we have to store `vptr` for allocatable variable even it is not allocated.
Is there any other better way to do it?

kmr-srbh: I checked the test and it seems it is some sort of a special case, where we are able to access an unallocated member, and call the type-bound procedure? The test works with GFortran even when the `class(toml_value), allocatable :: val` member is made a `type` variable.

I think this change in itself is fine, since the allocatable variables will get their vtable ptr set to the appropriate pointer after allocation because the memory is zeroed and allocated type can be different. :)

PR #8507, opened 2025-09-01
fix : Load all the unloaded submodules of loaded dependent modules of a submodule for `monolithic compilation`

AnshMehta1:


src/libasr/asr_utils.cpp
```diff
@@ -1085,6 +1085,9 @@ void load_dependent_submodules(Allocator &al, SymbolTable *symtab,
         = ASR::down_cast2<ASR::TranslationUnit_t>(ASR::make_TranslationUnit_t(al, loc,
             symtab, nullptr, 0));
 
+    // Keeps track of loaded dependent modules whose submodules are not loaded
+    std::vector<ASR::Module_t*> dependent_modules_with_unloaded_submodules;
```
certik, line 1089: The term "unloaded" suggests that it was loaded, then actively unloaded, and now we are loading it again.

However I think the logic here is that you figure out which submodules need to be loaded. So maybe let's use a different name:


```suggestion
    std::vector<ASR::Module_t*> dependent_modules_with_not_yet_loaded_submodules;
```

Or something like that.

AnshMehta1: Sure, I will make the changes accordingly.

certik: You can do it in the next PR, this is very minor.

Issue #8471, opened 2025-08-28
characters after `achar(0)` is not stored properly


jinangshah21:
MRE:
```fortran
program nul_char_demo
  implicit none
  character(len=:), allocatable :: s
  s = 'ABC' // achar(0) // 'XYZ'
  print *, 's length:', len(s)
  print *, 's: ', s
end program nul_char_demo
```

Outputs:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran c.f90 && ./a.out
 s length:           7
 s: ABCXYZ
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90
s length:    7
s:     ABC
```

assem2002:
I believe it's a bug with how we print strings.
```.f90
program nul_char_demo
  implicit none
  character(len=:), allocatable :: s
  s = 'ABC' // achar(0) // 'XYZ'
  print *, 's length:', len(s)
  print *, 's: ', s(5:5) ! OUTPUTS --> `X`
end program nul_char_demo
```

PR #8508, opened 2025-09-01
fix: handle printing characters after `achar(0)`

akramhany:
Towards #8471


certik:
If there is a way to test it, then add a test. If not, then not.

akramhany:
> If there is a way to test it, then add a test. If not, then not.

I don't think there is any, I will add a test for the whole issue when I fix it in the PR, but for this part alone, we can't really test it.

akramhany:
I intended at first to fix the whole issue in this PR, but now I think it would be better to divide it between more than one PR, so yes I think this is ready for merge.

PR #8511, opened 2025-09-01
fix(parser): accept READ with format literal and no unit (fixes #4527)

### LINKED ISSUE #4527: READ without a unit not allowed
In this program lfortran disallows the READ statement, which was valid Fortran 77 and is still valid Fortran 2023 (12.6.1 allows it without a unit, and 12.6.4.3 says where the reading is from). The program:
```
(lf) john:~/Test$ cat readtest1.f90
implicit none
  character reply
  print *,'Enter one character.'
  READ '(A)', REPLY
  print *,'You said ',reply 
end program
```
AMD flang, gfortran, g95, ifort and ifx all compiled and ran it but this is what lfortran 0.36.0 did in a Ubuntu 22.04 x86_64 system:
```
lf) john:~/Test$ lfortran readtest1.f90
syntax error: Token ''(A)'' (of type 'string') is unexpected here
 --> readtest1.f90:4:8
  |
4 |   READ '(A)', REPLY
  |        ^^^^^ 


(lf) john:~/Test$
```

(lf) john:~/Test$

krystophny:
Allow READ with a format string literal and no unit: \n- New grammar alt:  builds Read_t with format field set and default unit.\n- Mirrors existing  pattern.\n- Keeps GLR conflicts unchanged (no single quotes in grammar).\n\nVerification\n- Before:  failed with token '(A)' unexpected.\n- After: parses and compiles; runtime reads from stdin (EOF in CI).\n


certik:
Add a test into integration_tests. If it doesn't fully compile, then just check that ASR gets constructed by also registering it in `tests/tests.toml`.

PR #8514, opened 2025-09-02
semantics(925): implicit typing for PARAMETER; fold initializer

### LINKED ISSUE #925: Parser: allow parameter before implicit statement
Minimal example:
```fortran
subroutine f()
parameter (NL=100,NP=60)
implicit double precision (a-h,o-z)
end subroutine
```
In GFortran it parser correctly but fails in LFortran:
```console
$ gfortran -c a.f90 -o a.o
$ lfortran -c a.f90 -o a.o
syntax error: Token 'double' is unexpected here
 --> a.f90:3:10
  |
3 | implicit double precision (a-h,o-z)
  |          ^^^^^^ 


Note: if any of the above error or warning messages are not clear or are lacking
context please report it to us (we consider that a bug that needs to be fixed).
```

krystophny:
Summary
- Allow untyped PARAMETER statements to declare entities via IMPLICIT typing (fixes #925).

Scope
- Semantics: when `parameter (x=...)` is used without an explicit type, declare `x` using IMPLICIT rules if allowed, set storage to `parameter`, and fold the initializer to the declared type.
- Also handle the case where a PARAMETER attribute was seen before symbol use: declare on first use under IMPLICIT typing.
- No changes to codegen required; constants flow through `m_value` for PARAMETER.

Verification
- Local smoke:
  - `program p; parameter (x=5); print *, x; end` with `--implicit-typing` prints `5.00000000e+00`.
  - `integration_tests/implicit_typing_03.f90` builds and runs with `--implicit-typing`, prints `90`.
- Behavior under `implicit none`: using untyped PARAMETER without IMPLICIT rules errors cleanly.

Rationale
- Fortran allows `parameter (x=...)` to implicitly type `x` via IMPLICIT rules unless `implicit none` applies. This brings LFortran into line with that behavior while preserving strictness under `implicit none`.


krystophny:
@certik running pre-commit hooks against this file gives immense changes. Is this worth it at this stage for fixing a small bug on top, or should we work against the original formatting? See #8519

krystophny:
> @certik running pre-commit hooks against this file gives immense changes. Is this worth it at this stage for fixing a small bug on top, or should we work against the original formatting? See #8519

... or maybe changing this file is even nonsensical for this bug. Could be a remnant of the rebase and some pre-commit hooks run accidentially. I try to restore something useful again.

PR #8516, opened 2025-09-02
chore(format): clang-format parser.cpp only

krystophny:
Summary
- Format `src/lfortran/parser/parser.cpp` with repository clang-format only.

Scope
- Formatting-only; no behavior changes.

Verification
- Build succeeds; no code changes beyond whitespace/layout.

Rationale
- Isolate formatting churn to simplify review of subsequent functional changes (CRLF handling).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,7 @@ reference how to contribute to the project.
 - Tools: CMake (>=3.10), Ninja, Git, Python (>=3.8), GCC/Clang.
 - Generators: re2c, bison (needed for build0/codegen).
 - Libraries: zlib; optional: LLVM dev, libunwind, RapidJSON, fmt, xeus/xeus-zmq, Pandoc.
+- Pre-commit hooks: see `.pre-commit-config.yaml`; activate with `pre-commit install`.
 
 ## Build, Test, and Development Commands
 - Typical dev config (Ninja + LLVM) is specified in `./build1.sh`:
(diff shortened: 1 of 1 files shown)


krystophny:
Two short comments in AGENTS.md added in addition.

certik:
I don't want to do pre commit hooks. I have lost work in the past due to various bugs in them. For example if you make a change and then clang format reformats the file before committing, the change is lost. You can't easily undo it. I think clang-format results depend on the version also. Then it has to be installed on people's machines, etc.

Right now it adds friction. After we reach beta quality, then we can possibly introduce this. Until then let us move with high velocity with minimal "requirements". LLMs seem to be fine with any format.


Regarding reformatting the file, if it makes life a lot easier for you, we can do it one time, but I would also prefer to do reformatting after we reach beta.

krystophny:
> I don't want to do pre commit hooks. I have lost work in the past due to various bugs in them. For example if you make a change and then clang format reformats the file before committing, the change is lost. You can't easily undo it. I think clang-format results depend on the version also. Then it has to be installed on people's machines, etc.
> 
> Right now it adds friction. After we reach beta quality, then we can possibly introduce this. Until then let us move with high velocity with minimal "requirements". LLMs seem to be fine with any format.
> 
> Regarding reformatting the file, if it makes life a lot easier for you, we can do it one time, but I would also prefer to do reformatting after we reach beta.

OK, then this was a misunderstanding, I assumed since `.pre-commit-config.yaml` exists, it should be activated. Shall we then do the opposite and document that one should not use it? Or remove clang-format there?

PR #8517, opened 2025-09-02
Migrate from deprecated calls

certik:
This approach introduces ifdefs that we can hide into llvm utils later. But it might avoid the reliance on the `ptr_type_deprecated`, removing of which is the first step. Then we can clean up the rest after that.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -6800,8 +6800,18 @@ public:
             llvm::Type* target_el_type = llvm_utils->get_type_from_ttype_t_util(x.m_target, ASRUtils::extract_type(target_type), module.get());
             llvm::Type* value_el_type = llvm_utils->get_type_from_ttype_t_util(x.m_value, ASRUtils::extract_type(value_type), module.get());
             if( is_value_fixed_sized_array && is_target_fixed_sized_array ) {
-                value = llvm_utils->create_gep_deprecated(value, 0);
-                target = llvm_utils->create_gep_deprecated(target, 0);
+                llvm::Type* value_gep_ty = nullptr;
+                llvm::Type* target_gep_ty = nullptr;
+#if LLVM_VERSION_MAJOR <= 16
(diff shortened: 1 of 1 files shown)


certik:
It worked locally for me in LLVM 11 and 20, but failed at the CI with:
```
[ 11%] Building Fortran object CMakeFiles/arrays_op_9.dir/arrays_op_9.f90.o
LCOMPILERS_ASSERT failed: /Users/runner/work/lfortran/lfortran/lfortran-0.56.0-248-g8a1d78970/src/libasr/codegen/llvm_utils.cpp
function get_type_from_ttype_t(), line number 1158 at 
ASRUtils::is_fixed_size_array(v_type->m_dims, v_type->n_dims)
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6838,19 +6848,51 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    llvm::Type* value_gep_ty = nullptr;
+#if LLVM_VERSION_MAJOR <= 16
+                    value_gep_ty = llvm::cast<llvm::PointerType>(value->getType())->getElementType();
+#else
+                    if (llvm::isa<llvm::AllocaInst>(value)) {
+                        value_gep_ty = llvm::cast<llvm::AllocaInst>(value)->getAllocatedType();
+                    } else {
+                        value_gep_ty = llvm_utils->get_type_from_ttype_t_util(x.m_value, value_type, module.get());
+                    }
+#endif
```
ubaidsk, line 6880: Would it be possible or make sense to wrap this up in some function? I have noticed that C++ extensions for IDEs are unable to figure out the block wrap correctly when we have these #if-else macros defined in a block.

By block wrap, I mean the following carret thing that is available with most IDEs. In the following pic, we can see that the wrap stopped before the `#if`. Ideally, I think it should have wrapped until the end of the block scope.

I understand this is a bug in the extension, and not in our approach. Just wondering if there is a way to work around it.

image

Issue #4694, opened 2024-08-26
Improve printing of ASR to be exactly equal to Clojure


certik:
* Improve printing of ASR to be exactly equal to Clojure
  * [ ] Use `:x` instead of `x:`
  * [ ] add an option to print keywords like `(Integer :kind 4 :dim [])`

Maybe a few other changes.

ubaidsk:
To close this issue, we only have https://github.com/lcompilers/lpython/issues/1496 remaining. Do we want to move forward with using `/` as the namespace separator instead of `@` as discussed in the linked issue?

We can do this update in two ways:
- For the whole asr: this means that wherever we have used `@` in our AST->ASR and ASR->ASR for separating a namespace, we would instead use `/`.
- Or we can simply replace `@` with `/` when printing ASR in Clojure format. Meaning, when printing ASR with `--clojure`, we can call a function like `make_sym_clojure_compatible()` (we need to define this function) when printing any symbol and this returns us the clojure compatible symbol (replacing `@` with `/`) which we simply print.

Note that the linked issue discusses about prepending a namespace like `lpython.namespace/pow/pow_i32`. Do we need this namespace as well in the first ASR clojure version?

I have confirmed from https://www.mycompiler.io/new/clojure that using `/` in a clojure variable works fine.

<img width="2996" height="690" alt="Image" src="https://github.com/user-attachments/assets/d547f8a4-fd3c-40c6-9d46-a7a135b1d7fd" />

PR #8520, opened 2025-09-02
Make ASR symbols Clojure compatible

ubaidsk:
Replace @ with / in ASR symbol names and symbol table keys when printing in Clojure format to ensure Clojure compatibility.

This follows the second approach shared [here](https://github.com/lfortran/lfortran/issues/4694#issuecomment-3240455952).

fixes https://github.com/lcompilers/lpython/issues/1496


tests/tests.toml
```diff
@@ -3084,6 +3084,7 @@ asr = true
 [[test]]
 filename = "../integration_tests/functions_10.f90"
```
ubaidsk, line 3085: There are only few existing reference tests that used `@` in their generated ASR. This one was picked out of those few because, its stdout is shorter (might not be the shortest) as compared to other examples that have `@` in their ASR output. The shortest one was not picked so as to ensure we have sufficient/robust no. of `@` replacements done.

ubaidsk:
I think we can squash merge this.

Issue #8524, opened 2025-09-04
MRE for `AssertFailed: left_type2->type < 7`


certik:
I created an MRE from https://fortran-lang.discourse.group/t/polymorphism-and-generic-operator/10384:

```fortran
module test1_mod
    implicit none

    type, abstract :: Vect
    contains
        private
        procedure(addInterface), pass(self), deferred :: add
        generic, public :: operator(+) => add
    end type Vect

    abstract interface
        function addInterface(self, other)
            import :: Vect
            class(Vect), intent(in) :: self
            class(Vect), intent(in) :: other
            class(Vect), allocatable :: addInterface
        end function addInterface
    end interface

    type, extends(Vect) :: Vect2D
        real :: X = 0.0
        real :: Y = 0.0
    contains
        private
        procedure :: add => Vect2DAdd
    end type Vect2D

contains
    function Vect2DAdd(self, other)
        class(Vect2D), intent(in) :: self
        class(Vect), intent(in) :: other
        class(Vect), allocatable :: Vect2DAdd
        select type(o => other)
        type is (Vect2D)
            Vect2DAdd = Vect2D(self%X + o%X, self%Y + o%Y)
        end select
    end function Vect2DAdd
end module

program test1
    use test1_mod
    implicit none

    class(Vect), allocatable :: v1
    type(Vect2D) :: v2D1, v2D2

    v2D1 = Vect2D(0.,1.)
    v2D2 = Vect2D(3.,4.)
    v1 = v2D1 + v2D2   ! fails here

end program test1
```

This gives:
```console
$ lfortran a.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2736
    LCompilers::print_stack_on_segfault();
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2663
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 1138
    lcompilers_unique_ID_separate_compilation = compiler_options.separate_compilation ? LCOMPILERS_UNIQUE_ID : "";
  File "/Users/ondrej/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
[...]

PR #8526, opened 2025-09-04
Give an actual error message

certik:
Towards #8524.

Instead of an internal error, we now get on the example in #8524:
```
$ lfortran a.f90
semantic error: Operator `+` undefined for the types in the expression `vect2d + vect2d`
  --> a.f90:49:10
   |
49 |     v1 = v2D1 + v2D2   ! fails here
   |          ^^^^^^^^^^^
```
Which is better, probably good for now, until we actually implement it.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2008,6 +2008,7 @@ RUN(NAME custom_operator_15 LABELS gfortran llvm)
 RUN(NAME custom_operator_16 LABELS gfortran llvm)
 RUN(NAME custom_operator_17 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME custom_operator_18 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME custom_operator_19 LABELS gfortran llvm)
 
 RUN(NAME types_07 LABELS gfortran)
 RUN(NAME types_08 LABELS gfortran)
--- a/integration_tests/custom_operator_19.f90
+++ b/integration_tests/custom_operator_19.f90
@@ -1,0 +1,50 @@
+module custom_operator_19_inner
+  implicit none
+  interface operator(.myop.)
+    module procedure int_myop
+  end interface
+contains
+  pure function int_myop(a, b) result(r)
+    integer, intent(in) :: a, b
+    logical :: r
+    r = (a == b)
+  end function
+end module
+
+module custom_operator_19_outer
+  use custom_operator_19_inner, only: operator(.myop.)
+  implicit none
+end module
+
+module custom_operator_19_user
+  use custom_operator_19_outer, only: operator(.myop.)
+  implicit none
+  interface operator(.myop.)
+    module procedure logical_myop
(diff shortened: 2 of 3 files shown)


src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -606,6 +606,13 @@ class ImplicitCastRules {
       ASRUtils::type_get_past_pointer(left_type));
     ASR::ttype_t *right_type2 = ASRUtils::type_get_past_array(
       ASRUtils::type_get_past_pointer(right_type));
+
+    // Skip implicit casting for types not covered by the cast rules
+    // (StructType, EnumType, etc. should use overloaded operators instead)
+    if (left_type2->type >= num_types || right_type2->type >= num_types) {
+        return;
+    }
```
certik: @Pranavchiku I am not sure this is the right fix --- should we just quietly return here if the type doesn't fit our table?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10709,7 +10709,28 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    break;
+                case (ASR::Pow):
+                    op_str = "**";
+                    break;
+                default:
+                    LCOMPILERS_ASSERT(false);
+            }
+            diag.add(Diagnostic("Operator `" + op_str + "` undefined for the types in the expression `" + ASRUtils::type_to_str_fortran_expr(left_type, left)
+                                + " " +  op_str + " " + ASRUtils::type_to_str_fortran_expr(right_type, right) + "`", Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
+            throw SemanticAbort();
```
certik: This might be an internal compiler error later on, but right now it triggers, so I think that's fine.

src/lfortran/semantics/asr_implicit_cast_rules.h
certik: We could emit a warning as well.

I think this happens when we have `+` for two structs.

Pranavchiku: Yes, this is not good. We should not be doing `>= num_types` rather explicit type checking must be used.

src/lfortran/semantics/ast_common_visitor.h
Pranavchiku: ~This is redundant, it will hit `LCOMPILERS_ASSERT(false);` making it just a dead code.~

Edit: Ahh I see, makes sense. This is good.

Pranavchiku:
Are you using any AI here?

certik:
Idea: I wonder if we should turn most of these asserts into internal compiler errors ICE with location information, so that they stay in release mode also. I made this https://github.com/lfortran/lfortran/issues/8528.

src/lfortran/semantics/ast_common_visitor.h
certik: No actually it gives a nice error message now https://github.com/lfortran/lfortran/pull/8526#issue-3382508090, so it's not dead code.

PR #8532, opened 2025-09-04
fix: [new_classes] allocate vptr properly while storing it

jinangshah21:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -4252,17 +4252,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                struct_t->m_struct_signature, struct_sym, module.get());
-        if (LLVM::is_llvm_pointer(*struct_t->m_struct_signature)) {
-            if (is_unallocated) {
-                    llvm::Value* vptr = builder->CreateAlloca(llvm_utils->vptr_type);
-                    builder->CreateStore(builder->CreateBitCast(vptr, v_llvm_type), ptr);
-                }
+                struct_t->m_struct_signature, struct_sym, module.get())->getPointerTo();
+        if (is_unallocated) {
+            llvm::Value* vptr = builder->CreateAlloca(llvm_utils->vptr_type);
+            builder->CreateStore(builder->CreateBitCast(vptr, v_llvm_type), ptr);
```
kmr-srbh, line 4263: Ah, I get the issue now. I refactored the function to use the structure symbol, so the type is always just `StructType` with no `Allocatable` wrapper, hence we never reached that deleted block.

Issue #8441, opened 2025-08-25
Allocating an unallocated list with different size and later giving  it elements of different size gives no error.


yuvrajsinghgmx:
Related to  #8342 

MRE :

`program m
  implicit none
  character(len=:), allocatable :: list(:)
  if (.not.allocated(list)) then
    list = [character(len=2):: ]
    list = ["aaaaa","aaaaa","aaaaa"]
  end if
  print *, len(list) 
end program`

Output : 

`(lf) yuvrajsingh@FairyTale:~/OS/lfortran$ ./src/bin/lfortran a.f90
5`

assem2002:
Is that related to what you're working on @xaerru ?

xaerru:
> Is that related to what you're working on @xaerru ?

Yes. 

With https://github.com/lfortran/lfortran/pull/8429 merged, we do throw an error now.

```shell
Runtime Error: Size mismatch in assignment to 'list'

LHS size is 1 and RHS size is 0
```

But it needs to print that LHS is not allocated instead of this size mismatch. I'll fix it.

harperjf:
The lfortran output 5 was correct in any Fortran version 2003 or later because ```list```  was automatically reallocated by the statement
```  list = ["aaaaa","aaaaa","aaaaa"]```
irrespective of what its previous value was. Gfortran gave the same output. So no error should be thrown by that program unless you want Fortran 95 behaviour.

assem2002:
Yeah, that's what we're trying to achieve. We're working on the `realloc-lhs` flag to tolerate the automatic reallocation of the left hand side, and we're also adding proper errors when the flag is not in use, that way we would fulfill the 2 points you mentioned.

PR #8537, opened 2025-09-05
feat: add move assignment for allocatable arrays

xaerru:
Address comments from https://github.com/lfortran/lfortran/pull/8429, Fix https://github.com/lfortran/lfortran/issues/8441

Design:
1. Allow LHS to be not allocated if RHS is a `FunctionCall` returning an allocatable array. The allocatable array returned should be moved from RHS to LHS.
2. If LHS is already allocated and RHS is a `FunctionCall` returning an allocatable array then throw bounds check error if LHS's size is not equal to RHS's size.

Changes done:
1. In `subroutine_from_function` pass, an `Assignment` with RHS being a `FunctionCall` is converted to a `SubroutineCall` with the LHS of the assignment passed in as the last parameter. This PR passes in a temporary in place of the LHS. Also the assignment to LHS from the temporary is converted to a "move" assignment.
2. Created a new function `copy_array_move` in `llvm_array_utils.h`. This function moves the data pointer of src to the dest and stores `nullptr` as the data pointer of the src to prevent double deallocation.


src/libasr/ASR.asdl
```diff
@@ -32,8 +32,8 @@ stmt
     = Allocate(alloc_arg* args, expr? stat, expr? errmsg, expr? source)
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
-    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs)
-    | DebugCheckArrayBounds(expr target, expr value)
+    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs, bool move)
```
certik: Is this to move the allocation? If so, should it be:

```suggestion
    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs, bool move_allocation)
```
?

src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -1053,6 +1053,62 @@ namespace LCompilers {
             builder->CreateStore(n_dims, this->get_rank(dest_ty , dest, true));
         }
 
+        // Copy destination's descriptor to the source descriptor
+        // Move the data pointer from dest to src
+        // And set the src's data pointer to null to prevent double deallocation
+        void SimpleCMODescriptor::copy_array_move(llvm::Type* src_ty, llvm::Value* src, llvm::Type* dest_ty, llvm::Value* dest,
```
certik:
```suggestion
        void SimpleCMODescriptor::copy_array_move_allocation(llvm::Type* src_ty, llvm::Value* src, llvm::Type* dest_ty, llvm::Value* dest,
```

src/libasr/ASR.asdl
```diff
@@ -32,8 +32,8 @@ stmt
     = Allocate(alloc_arg* args, expr? stat, expr? errmsg, expr? source)
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
-    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs)
-    | DebugCheckArrayBounds(expr target, expr value)
+    | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs, bool move_allocation)
+    | DebugCheckArrayBounds(expr target, expr value, bool move_allocation)
```
certik, line 36: Let's document what this new member does in doc/.

src/libasr/asr_utils.h
```diff
@@ -3080,7 +3080,7 @@ ASR::expr_t* get_ArrayConstructor_size(Allocator& al, ASR::ArrayConstructor_t* x
 
 ASR::asr_t* make_Assignment_t_util(Allocator &al, const Location &a_loc,
     ASR::expr_t* a_target, ASR::expr_t* a_value,
-    ASR::stmt_t* a_overloaded, bool a_realloc_lhs);
+    ASR::stmt_t* a_overloaded, bool a_realloc_lhs, bool a_move = false);
```
certik: Let's not use default arguments. Make it explicit  true/false in the caller.

certik:
Let's see if tests pass. The comments can be fixed after it's merged.

PR #8546, opened 2025-09-06
fix: printing a string that contains a null-char inside it

### LINKED ISSUE #8471: characters after `achar(0)` is not stored properly

akramhany:
Fixes #8471

### DIFF
--- a/tests/reference/llvm-boz_01-def9db5.json
+++ b/tests/reference/llvm-boz_01-def9db5.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-boz_01-def9db5.stdout",
-    "stdout_hash": "dfb999750ff5454097513f780b4a8cf4f0778e225d5f12b0291a707f",
+    "stdout_hash": "0fd7ff7331df1cdc55f713e7cb1c750048bb3e1692f2a6f09cd4f5cc",
     "stderr": "llvm-boz_01-def9db5.stderr",
     "stderr_hash": "2c1014e4f04672dfbcc83dde266587a98d7dc9651f6401dcaab51839",
     "returncode": 0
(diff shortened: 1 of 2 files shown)


akramhany:
How can I add test for this PR? Because the test needs to be on the `print` function, and I am not really sure how can I test it.

kmr-srbh:
> How can I add test for this PR? Because the test needs to be on the `print` function, and I am not really sure how can I test it.

@akramhany you can add a reference test, with `run = true`.

akramhany:
@kmr-srbh Great! Thanks Saurabh!

akramhany:
Can I get Exhaustive Checks here?

certik:
@assem2002 do you mind reviewing this PR please?

tests/reference/run-print5-c57814b.stdout
kmr-srbh, line 1: @akramhany the output here says it's binary. Could you please verify the content of this file locally and see if it contains what we want?

akramhany: The output is indeed correct, I think it is presented as binary because the null-char inside it.
image

kmr-srbh: Indeed! Storing outputs from GFortran and Flang gives the same binary file output too! Great work! :)

Screenshot 2025-09-07 at 21 13 06

PR #7625, opened 2025-06-07
enh: simplify `StructInstanceMember` handling for polymorphic `subroutine` and `function` calls

kmr-srbh:
Part of #7578, fixes #7318

Issue #7578, opened 2025-06-03
Segfault of code generation on code that makes use of a wrapped polymorphic type


difference-scheme:
Another code generation bug on polymorphic code:
$ cat test.f90 
```
module test
   
   type, public, abstract :: AbsType
   contains
      procedure(method), deferred :: method
   end type AbsType

   abstract interface
      subroutine method(self)
         import
         class(AbsType), intent(in)  :: self
      end subroutine method
   end interface

   type :: Wrapper
      class(AbsType), allocatable :: obj
   end type Wrapper
   
   type :: Client
      type(Wrapper) :: wrapped
   contains
      procedure :: caller
   end type Client
   
contains

   subroutine caller(self)
      class(Client), intent(in) :: self
      call self%wrapped%obj%method()
   end subroutine caller
   
end module test

```

lfortran -c --generate-object-code test.f90 
`Segmentation fault (core dumped)`

kmr-srbh:
This occurs due to nested `StructInstanceMember` during subroutine call. Working on it.

kmr-srbh:
Thanks for this bug report @difference-scheme! This helped reveal a few more bugs. It took me sometime to iron out all of them. #7625 is the PR for this. :)

difference-scheme:
@kmr-srbh Awesome! Thanks for the PR.

A similar polymorphism-related bug, with multiple different manifestations, is probably underlying #7571 and #7574. I'd appreciate if this one could be looked into next.

difference-scheme:
Oh, wow! 

I missed that your PR apparently partly fixes also #7571 and #7574. Great job!

Would it be too much trouble to now implement full support for polymorphic arrays?

certik:
> Would it be too much trouble to now implement full support for polymorphic arrays?

@kmr-srbh go ahead and open up an issue for this.

@difference-scheme We are systematically getting the OOP in shape, we'll get to it.

difference-scheme:
@certik Great! 

I'll keep reporting any further OOP bugs that I may run into.

certik:
> I'll keep reporting any further OOP bugs that I may run into.

Good news is that it is still easy to find bugs! :)

Issue #7571, opened 2025-06-03
Code generation segfaults for code that uses an array declared with the `class` specifier


difference-scheme:
This is another bug on polymorphic code. In particular, on code that makes use of an extended type, which employs an array of some other type, that is declared in terms of the `class` specifier:
$ cat test.f90
```
module test
   
   type, abstract :: SomeType
   end type SomeType

   type, abstract :: Base
   contains
      procedure(method), deferred :: method
   end type Base

   abstract interface
      subroutine method(self,arr)
         import
         class(Base),     intent(inout) :: self
         class(SomeType), intent(inout) :: arr(:)
      end subroutine method
   end interface

   type, extends(Base) :: Extended
      class(Base), allocatable :: basic
   contains
      procedure :: method => implementation
   end type Extended

contains
   
   subroutine implementation(self,arr)
      class(Extended), intent(inout) :: self
      class(SomeType), intent(inout) :: arr(:)
      call self%basic%method(arr)
   end subroutine implementation

end module test
```
$ lfortran -c --generate-object-code test.f90 
`Segmentation fault (core dumped)
`

If one turns the `arr` variable into a scalar, then the segfault vanishes.

PR #8560, opened 2025-09-09
enh: [new_classes] don't store vtable if variable is not allocated and check at runtime call

jinangshah21:


integration_tests/CMakeLists.txt
```diff
@@ -1923,11 +1923,11 @@ RUN(NAME class_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes
 RUN(NAME class_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
 RUN(NAME class_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
-RUN(NAME class_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs)
-RUN(NAME class_26 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
+RUN(NAME class_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes EXTRA_ARGS --realloc-lhs)
+RUN(NAME class_26 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_27 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
-RUN(NAME class_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_new_classes)
-RUN(NAME class_29 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
jinangshah21, line 1930: `class_26` and `class_28` are easy fix after we support `unlimited_polymorphic_type` for new classes

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12943,37 +12975,64 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-            tmp = builder->CreateCall(fnTy, fn, args);
+
+            // If derived type variable is not allocated use it's declared type VTable
+            llvm::Value* vtable_ptr;
+            llvm::Type* result_type = llvm_utils->get_type_from_ttype_t_util(
+                const_cast<ASR::expr_t*>(&x.base), x.m_type, module.get());
+            tmp = llvm_utils->CreateAlloca(*builder, result_type);
+            llvm_utils->create_if_else(is_allocated, [&]() {
+                // Get Runtime VTable Pointer
+                vtable_ptr = builder->CreateBitCast(llvm_dt, fnPtrPtrPtrTy);
```
jinangshah21:
```fortran
program class_29
    type :: toml_table
    end type

    class(toml_table), allocatable :: temp3
   call temp3%anyfunc()
end program
```

This runtime check is due to above program. As methods can be called even if variable is not allocated.

certik: I made this https://github.com/lfortran/lfortran/issues/8561. It's not clear to me that it's allowed.

certik: @jinangshah21 it seems that this is not allowed per the standard. Is this used in any 3rd party code? If so, point me to it and let me have a look.

jinangshah21: I didn't see this example in any 3rd party code, it was in our integration tests.

certik: We should remove it. Let's also look at the git history to see which PR added it and if there was some motivation that we can't see right now.

certik:
If this whole PR is only supporting a case that shouldn't be supported, then let's not do it.

If it fixes things that we should support, we can merge it, and further iterate on it.

The test of calling a method on unallocated variable should give a nice runtime error in Debug mode.

src/libasr/codegen/asr_to_llvm.cpp
jinangshah21: Tests for this was added in #7625, MREs were: #7571, #7578 where it is not clear in MRE that struct member is allocated or not. I didn't see any motivation from MREs, its seems that GFortran was working fine without allocating so we added the tests.

jinangshah21:
I think, for now we can avoid above special case as we don't have any strong motivation for it. I will revert that runtime check.

integration_tests/class_39.f90
```diff
@@ -28,6 +28,8 @@ subroutine caller(self)
       class(Client), intent(in) :: self
       type(Client) :: type_s
 
+      allocate(self%wrapped%obj)
+      allocate(self%wrapped%obj%nested_obj)
       call self%wrapped%obj%abs_method(42)
```
jinangshah21: I have added explicit allocate statements in tests so we don't allow method call from an unallocated object.

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, GFortran seems to run it, but it shouldn't.

integration_tests/class_42.f90
```diff
@@ -1,6 +1,6 @@
 module class_42_mod
 
-   type, abstract, public :: AbsType
+   type, public :: AbsType
```
certik, line 3: Why was `abstract` removed?

If this is temporary, let's create an issue for this to fix it.

certik: I made this https://github.com/lfortran/lfortran/issues/8563.

PR #8565, opened 2025-09-10
refactor: made functions use create_gep2

sriganeshres:
towards: #7836


sriganeshres:
@certik Only `get_pointer_to_data` and `visit_Assignment` function use deprecated `create_gep_deprecated`.

PR #8567, opened 2025-09-10
fix: using elemental functions with custom typed arrays

### LINKED ISSUE #8550: Use elemental function with custom types
### MRE

```fortran
module greetings_mod
  implicit none
  type string_t
    character(len=:), allocatable :: string_
  end type string_t
contains

  elemental function bracket(x) result(res)
    class(string_t), intent(in) :: x
    type(string_t) :: res

    res%string_ = x%string_ // " called func"
  end function bracket

end module greetings_mod


program main
  use greetings_mod
  implicit none

  type(string_t) :: lines(5)

  lines(1)%string_ = "Hello"
  lines(2)%string_ = "World"
  lines = bracket(lines)

  print *, lines(1)%string_
  print *, lines(2)%string_
end program main
```

### Lfortran Output

```sh
(lf)  🐍 lf  akramhany@akramhany  ~/Akram/Projects/Fortran/lfortran  ↱ JulienneBuildError ±  lfortran ../b.f90
ASR verify pass error: FunctionCall::m_name bracket must be returning a non-void value.
  Binary file "/home/akramhany/Akram/Projects/Fortran/lfortran/src/bin/lfortran", in _start
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/bin/lfortran.cpp", line 2739, in main
    return main_app(argc, argv);
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/bin/lfortran.cpp", line 2663, in main_app(int, char**)
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/bin/lfortran.cpp", line 1193, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 393, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
    compiler_options, run_fn, "", infile);
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 13795, in LCompilers::asr_to_llvm(LCompilers::ASR::TranslationUnit_t&, LCompilers::diag::Diagnostics&, llvm::LLVMContext&, Allocator&, LCompilers::PassManager&, LCompilers::CompilerOptions&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
  File "/home/akramhany/Akram/Projects/Fortran/lfortran/src/libasr/pass/pass_manager.h", line 318, in LCompilers::PassManager::apply_passes(Allocator&, LCompilers::ASR::TranslationUnit_t*, LCompilers::PassOptions&, LCompilers::diag::Diagnostics&)
```

akramhany:
Fix #8550

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -839,7 +839,7 @@ RUN(NAME submodule_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-# RUN(NAME submodule_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME submodule_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME floor_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran) # floor body
 RUN(NAME floor_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran) # floor symboltable
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -1079,6 +1079,13 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
             }
         }
 
+        // If x.m_dt is an array, and the subroutine is elemental, we need to replace the m_dt with an array item
+        // and call the elemental function on each element of the x.m_dt array not on the whole array
+        if (x.m_dt && ASRUtils::is_array(ASRUtils::expr_type(x.m_dt))) {
+            ASR::SubroutineCall_t& xx = const_cast<ASR::SubroutineCall_t&>(x);
+            vars.push_back(al, &(xx.m_dt));
+        }
+
         if( vars.size() == 0 ) {
             return ;
         }


integration_tests/CMakeLists.txt
```diff
@@ -838,7 +838,7 @@ RUN(NAME submodule_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME submodule_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME submodule_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+# RUN(NAME submodule_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
akramhany: I commented this as it is now failing, and I am working on fixing it now.
(Note: this was a test I added that was related to elemental functions too)

integration_tests/elemental_17.f90
```diff
@@ -0,0 +1,31 @@
...
+
+program elemental_17
+  use elemental_17_mod
+  implicit none
+
+  type(string_t) :: lines(3)
+
+  lines(1)%string_ = "Hello"
+  lines(2)%string_ = "World"
+  lines = bracket(lines)
```
akramhany, line 26: Line 26 will get expanded to the following code after the ASR passes:
```fortran
elemental subroutine bracket(x, res)
    type(string_t), intent(out) :: res
    type(string_t), intent(in) :: x
    res%string_ = _lcompilers_stringconcat(x%string_, " called func")
end subroutine bracket

__libasr_index_0_1 = lbound(lines, 1)
__libasr_index_0_ = lbound(lines, 1) - 1
do while (__libasr_index_0_ + 1 <= ubound(lines, 1))
    call bracket(lines(__libasr_index_0_), lines(__libasr_index_0_1))      ! here
end do
```

So is that correct? Passing an object as an `intent(in)` and `intent(out)` at the same time I mean.

akramhany:
Copilot is saying that the failing error is this:
```sh
fatal: unable to access 'https://github.com/Pranavchiku/numerical-methods-fortran.git/': Could not resolve host: github.com
Process completed with exit code 128.
```

But I don't know why I can't find the logs at all.

certik:
You have to click on raw logs: https://productionresultssa5.blob.core.windows.net/actions-results/5e3d9a33-6f05-4628-a4cd-e30809b4fc96/workflow-job-run-3dfe2f80-2dc4-5c42-ae21-543fcd17d68d/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-09-11T00%3A35%3A42Z&sig=FFcmEQEyCXlwpZM%2Bkxilv22osBoi0dFYZuhfmpMVcII%3D&ske=2025-09-11T09%3A17%3A57Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-09-10T21%3A17%3A57Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-09-11T00%3A25%3A37Z&sv=2025-05-05

certik:
Tests pass, but I think https://github.com/lfortran/lfortran/pull/8567#discussion_r2337905754 has to be fixed, so I made it a draft for now.

integration_tests/elemental_17.f90
akramhany: @certik Please tell me what you think about this.

certik: There is probably a way to simplify the ASR that we generate.

Can you formulate a specific question --- do you suspect the code is wrong?

akramhany: No, I am just asking if it is correct to pass a certain variable to a subroutine as an `intent(in)` and `intent(out)` at the same time. For example in this Fortran code:
```fortran
program main_program
  implicit none
  integer :: x

  x = 10
  call process_number(x, x)

  print *, "result: ", x

contains
  subroutine process_number(input_arg, output_arg)
    implicit none
    integer, intent(in)  :: input_arg
    integer, intent(out) :: output_arg

    output_arg = input_arg * 2
  end subroutine process_number
end program main_program
```

Can we pass `x` as an input and output at the same time for the subroutine `process_number`? Or is this considered an invalid Fortran code?

certik: I see --- I think it is not allowed. We should not do it in our passes. However above the arguments are not the same, we have `lines(__libasr_index_0_)` and `lines(__libasr_index_0_1)`, and they are different.

akramhany: Can you please check that again, as I think they are the same. The `__libasr_index_0_1` initial value is `1`, and `__libasr_index_0_` initial value is `1 - 1` which is `0`, but it gets incremented in the beginning of the loop so it becomes `1` too.

akramhany: Also, assuming that they are the same, how should I fix this? I am thinking of making a `tmp` variable to hold the result, and assign it to the `x` variable again after calling the subroutine, would that be correct?

czgdp1807: Logically, first `bracket(lines)` would be evaluated and stored somewhere like, `tmp` and then `lines = tmp` will be executed. Imagine an elemental function `f` which takes two inputs, `x` and `y` and depending upon the value of `y`, does something to `x`. Say `y` is odd then it overwrites `x` otherwise it doesn't do anything. Now doing, `f(x, y)` and storing it to a `tmp` becomes a must. I mean `f` can be anything and there might exist an implementation logic where evaluating it first becomes a must. Here in this case we are evaluating and assigning to the result both at the same time. This idea seems fast and memory efficient but might not work always. It has concerned me several times when I used to work on LFortran.

certik: The line:
```
lines = bracket(lines)
```
Must be reworked like this:
```
allocate(tmp(...)) // temporary
call bracket(tmp, lines)
lines = tmp
```
I can't see any other way to make this work correctly. You will get speedups for `lines2 = bracket(lines)`, but not for `lines = bracket(lines)`.

If this is not what we do, please open up an issue to get it done.

akramhany: @certik I opened an issue for it here #8601, also I mentioned some edges cases that might need separate handling.

src/libasr/pass/array_passed_in_function_call.cpp
```diff
@@ -868,7 +868,6 @@ class CallVisitor : public ASR::CallReplacerOnExpressionsVisitor<CallVisitor>
 
     template <typename T>
     void visit_Call(const T& x, const std::string& name_hint) {
-        LCOMPILERS_ASSERT(!x.m_dt || !ASRUtils::is_array(ASRUtils::expr_type(x.m_dt)));
```
certik, line 871: If this is false, what exactly is the type of `x.m_dt`? Why not testing it explicitly here?

certik:
I am running all tests. If they pass, let's merge it. Improvements can be done in subsequent PRs.

src/libasr/pass/array_passed_in_function_call.cpp
akramhany: `x.m_dt` should be a derived type expression (if it exists), and I think it only exists with methods and type-bound procedures. I am not really sure why there was an assertion on the condition that `x.m_dt` must not be an array, but there is no integration test failing so I don't think this would cause a problem.

Let's wait and see the exhaustive checks, as they might catch something.

certik: I see, I think it's fine then.

Issue #8494, opened 2025-08-30
error: throw error for missing required argument in "index"


HarshitaKalani:
```fortran
program main 
print *, index( substring = 'de', back = .true. )
end program
```
```console
 % lfortran c.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2736
    LCompilers::print_stack_on_segfault();
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2663
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 1138
    lcompilers_unique_ID_separate_compilation = compiler_options.separate_compilation ? LCOMPILERS_UNIQUE_ID : "";
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 112
    if (res.ok) {
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 5816
    b.is_body_visitor = true;
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 175
    visit_ast(*x.m_items[i]);
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/ast.h", line 5016
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/ast.h", line 4971
    case astType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/ast.h", line 5019
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/ast.h", line 4593
[...]

PR #8569, opened 2025-09-12
Fix: throw error for missing required argument in  intrinsic 

mohitdhingra22:
Fixes: #8494 
By adding proper semantic checks for missing required arguments in the index intrinsic.
Previously, calls like index(substring='de', back=.true.) caused an internal compiler error.
Now LFortran reports a clear semantic error.


mohitdhingra22:
> Thanks for the PR. Can you show a code which raises the error you added and on `main` it raises an ICE. Just want to see that the error added here actually gets raised.

Before: 
```fortran
program main
    print *, index(substring = 'de', back = .true.)
end program
```
```bash
(lf) mohitdhingra@Mohits-MacBook-Air lfortran % lfortran missing_index_arg.f90
  File "/Users/mohitdhingra/OpenSource/lfortran/src/bin/lfortran.cpp", line 2736
  File "/Users/mohitdhingra/OpenSource/lfortran/src/bin/lfortran.cpp", line 2663
  File "/Users/mohitdhingra/OpenSource/lfortran/src/bin/lfortran.cpp", line 1138
    lcompilers_unique_ID_separate_compilation = compiler_options.separate_compilation ? LCOMPILERS_UNIQUE_ID : "";
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 112
    if (res.ok) {
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 5838
    b.is_body_visitor = true;
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 175
    visit_ast(*x.m_items[i]);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 5016
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 4971
    case astType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 5019
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 4593
    case modType::BlockData: { v.visit_BlockData((const BlockData_t &)x); return; }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 2356
    transform_stmts(body, x.n_body, x.m_body);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 135
    this->visit_stmt(*m_body[i]);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 5065
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 4729
    case stmtType::Pragma: { v.visit_Pragma((const Pragma_t &)x); return; }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_body_visitor.cpp", line 4640
    this->visit_expr(*x.m_values[i]);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 5117
    void visit_expr(const expr_t &b) { visit_expr_t(b, self()); }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/ast.h", line 4767
    case exprType::Compare: { v.visit_Compare((const Compare_t &)x); return; }
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_common_visitor.h", line 9999
    v = intrinsic_as_node(x, is_function);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/lfortran/semantics/ast_common_visitor.h", line 9270
    tmp = create_func(al, x.base.base.loc, args, diag);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/libasr/pass/intrinsic_function_registry_util.h", line 1809
    ASR::ttype_t *arg_type0 = ASRUtils::expr_type(args[0]);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/libasr/asr_utils.h", line 418
    return ASR::expr_type0(f);
  File "/Users/mohitdhingra/OpenSource/lfortran/src/libasr/asr_expr_type_visitor.h", line 20
    LCOMPILERS_ASSERT(f != nullptr);
AssertFailed: f != nullptr
```
After:
```bash
(lf) mohitdhingra@Mohits-MacBook-Air lfortran % lfortran missing_index_arg.f90
semantic error: Missing actual argument 'string' in call to 'index'
 --> missing_index_arg.f90:2:14
  |
2 |     print *, index(substring = 'de', back = .true.)
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7341,6 +7341,16 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             args.p[kwarg_idx] = ASRUtils::EXPR(tmp);
         }
         fill_optional_args(intrinsic_name, args, x.base.base.loc);
+        // "Missing actual argument 'string' in call to 'index'"
```
HarshitaKalani: Do we need this comment? Can you please remove it?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7341,6 +7341,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
         fill_optional_args(intrinsic_name, args, x.base.base.loc);
+        for (size_t i = 0; i < min_args; ++i) {
+            if (args[i] == nullptr) {
+                std::string missing_name = (i < kwarg_names.size() && !kwarg_names[i].empty()) ? kwarg_names[i] : ("argument(" + std::to_string(i+1) + ")");
+                diag.add(Diagnostic(
+                    "Missing actual argument '" + missing_name + "' in call to '" + intrinsic_name + "'",
+                    Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
+                throw SemanticAbort();
+            }
+        }
```
Pranavchiku, line 7372: This should go inside `verify_args` I suppose.

Pranavchiku (changes requested):
This should be caught inside verify_args, let's check along those lines and then we can merge this.

mohitdhingra22:
> This should be caught inside verify_args, let's check along those lines and then we can merge this.

There is no such a function like `verify_args`in `ast_commonvisitor.h`, its only exist in `instrincs_function_registry.h`
can u tell me which exact line or function have to put this logic inside. @Pranavchiku

Pranavchiku:
`verify_args` for individual intrinsic, it will be present in `instrincs_function_registry.h`.

HarshitaKalani:
I think `astcommonvisitor` is the correct place to do this @Pranavchiku . We don't have keywords handling possible in verify_args. So the changes in this PR look good to me.

mohitdhingra22:
> I think `astcommonvisitor` is the correct place to do this @Pranavchiku . We don't have keywords handling possible in verify_args. So the changes in this PR look good to me.

Thanks for clarifying @Pranavchiku and @HarshitaKalani.
Since keyword argument handling isn’t available inside verify_args in intrinsic_function_registry.h, I’ve kept the missing required argument check inside ast_common_visitor.h. This ensures that cases with keyword arguments (like index(substring='de', back=.true.)) are correctly validated.
Let me know if you’d like me to adjust further, otherwise this should be the right place for the check.

tests/tests.toml
```diff
@@ -4798,3 +4798,6 @@ asr = true
 [[test]]
 filename = "../integration_tests/select_type_13.f90"
 llvm_new_classes = true
+
+[[test]]
+filename = "errors/missing_index_arg.f90"
```
certik: What is going on here -- I think the test is not registered, doesn't have what test should run.

certik:
Add a test to continue compilation that previously failed and now it works.

mohitdhingra22:
> Add a test to continue compilation that previously failed and now it works.

Thanks @certik for the suggestion. I have registered the test in tests/errors/continue_compilation_1.f90 so that the compiler continues after this error as well. I also ran the full test suite and updated the reference outputs — all tests are passing on my side.

tests/tests.toml
```diff
@@ -4798,3 +4798,6 @@ asr = true
 [[test]]
 filename = "../integration_tests/select_type_13.f90"
 llvm_new_classes = true
+
+[[test]]
+filename = "misssing_index_arg.f90"
```
certik: This isn't registered. I think this test shouldn't be here at all?

tests/errors/continue_compilation_1.f90
```diff
@@ -289,6 +289,9 @@ program continue_compilation_1
 
     print *, count(1)
     print *, count([2])
+    
+    ! index_missing_required_arg
+    print *, index(substring = "de", back = .true.)
```
certik: You need to add it at the end, so as to keep the diff small, see the comments in the file explaining how to add tests.

tests/tests.toml
mohitdhingra22: > This isn't registered. I think this test shouldn't be here at all?

https://github.com/lfortran/lfortran/pull/8569#pullrequestreview-3218265649
As Harshita said register in the `test.toml` im gonna register the test.

tests/errors/continue_compilation_1.f90
mohitdhingra22: Alright i move this to the end.

certik:
CI fails.

Issue #8476, opened 2025-08-28
List-directed output of character strings


harperjf:
Lfortran seems to disobey the F2023 standard 13.10.4 by putting in some blanks between character strings written in * (list-directed) format.
13.10.4 begins
"The form of the values produced is the same as that required for input, except as noted otherwise. With the exception of adjacent undelimited character sequences, the values are separated by one or more blanks or by a comma, or a semicolon if the decimal edit mode is COMMA, optionally preceded by one or more blanks and optionally followed by one or more blanks. Two undelimited character sequences are considered adjacent when both were written using list-directed input/output, no intervening data transfer or file positioning operations on that unit occurred, and both were written either by a single data transfer statement, or during the execution of a parent data transfer statement along with its child data transfer statements."

Also, lfortran using no option omits the obligatory blank space (13.10.4 paragraph 13) at the beginning of list-directed output. (The documentation of --print-leading-space mentions to "if format is unspecified". That must refer to unformatted output because * specifies a format. I suggest that the documentation should say "if the format is list-directed, i.e. * " because that has been in the standard ever since F77.)  

Test program
```
program listdirected
  print *, '"','?','"'
end program listdirected
```
Lfortran 0.56.0 output with no option (wrong)
```
"    ?    "
```
Lfortran 0.56.0 output with --std=f23 option (wrong in a different way)
```
 "    ?    "
```
Four other compilers' output (right)
```
 "?"
```

certik:
Good catch, looks like there are several bugs here that we need to fix. Thanks for reporting it.

PR #8572, opened 2025-09-13
Correct list-directed output of character strings

mohitdhingra22:
Fixes: #8476
 - Added a single leading blank at the beginning of list-directed output.
- Ensured no extra blanks are inserted between adjacent character arguments.
- Added integration tests to confirm correct behavior.

Test program 
```fortran
program listdirected
  print *, '"','?','"'
end program listdirected
```
Before : 
```bash
"    ?    "
```
After :
```bash
 "?"
 ```

### CODE REVIEW

File: tests/reference/run-array14-c9f6829.stdout [hunk 1]
```diff
@@ -1,2 +1,2 @@
-5.00000000e+00    -1.00000000e+00    3.00000000e+00    4.00000000e+00    2.00000000e+00
-5.00000000e+00    -1.00000000e+00    3.00000000e+00
+ 5.00000000e+00 -1.00000000e+00 3.00000000e+00 4.00000000e+00 2.00000000e+00
+ 5.00000000e+00 -1.00000000e+00 3.00000000e+00
```
Pranavchiku: why do we have leading spaces?

Pranavchiku (approved):
This looks towards good direction, feel free to merge once exhaustive checks pass.

HarshitaKalani (commented):
Hey @mohitdhingra22 , thanks for the PR. Can you please confirm if the leading spaces are expected? Otherwise the PR looks good.

mohitdhingra22:
> Hey @mohitdhingra22 , thanks for the PR. Can you please confirm if the leading spaces are expected? Otherwise the PR looks good.

Yes, the single leading space is intentional and required for character list-directed output.

tests/reference/run-array14-c9f6829.stdout
mohitdhingra22: The leading space in list-directed output is standard Fortran behaviour. It provides separation before printing values, especially for character strings, improving readability and ensuring consistent formatting across outputs. This matches what major Fortran compilers `(like gfortran)`do.

certik:
This should not print a leading space without `--print-leading-space`, and print the leading space with `--print-leading-space`.

mohitdhingra22:
> This should not print a leading space without `--print-leading-space`, and print the leading space with `--print-leading-space`.

 I have addressed the issue by modifying the formatting function to optionally print a leading space only when the `--print-leading-space` flag is provided. The default behavior now suppresses the leading space.
I updated relevant code, added tests, and verified the output with and without the flag. All tests are passing on my side.
Please let me know if you have any further suggestions.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1501,39 +1501,64 @@ void strip_outer_parenthesis(const char* str, int len, char* output) {
...
 }
 
-void default_formatting(char** result, int64_t *result_size_ptr, struct serialization_info* s_info){
+void default_formatting(char** result, int64_t *result_size_ptr, struct serialization_info* s_info, bool print_leading_space){
     int64_t result_capacity = 100;
     int64_t result_size = 0;
-    const int default_spacing_len = 4;
-    const char* default_spacing = "    ";
+    const int default_spacing_len = 1;
+    const char* default_spacing = " ";
```
kmr-srbh, line 1508:
```suggestion
    const char default_spacing = ' ';
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1501,39 +1501,64 @@ void strip_outer_parenthesis(const char* str, int len, char* output) {
...
     ASSERT(default_spacing_len == strlen(default_spacing));
-    *result = realloc(*result, result_capacity + 1 /*Null Character*/ );
+
+    *result = realloc(*result, result_capacity + 1);
```
kmr-srbh: Please add back the comment. It helps understand why the `+1` is needed. Although, I am not very sure if we even need that now, given we moved to null-independent strings sometime ago.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1501,39 +1501,64 @@ void strip_outer_parenthesis(const char* str, int len, char* output) {
...
+
+        if(current_is_char && *(char**)s_info->current_arg_info.current_arg != NULL){
+            size_to_allocate = (s_info->current_arg_info.current_string_len + default_spacing_len + 1) * sizeof(char);
         } else {
             size_to_allocate = (60 + default_spacing_len) * sizeof(char);
         }
+
         int64_t old_capacity = result_capacity;
-        while(result_capacity <= size_to_allocate + result_size){ // Check if string extension is needed.
+        while(result_capacity <= size_to_allocate + result_size){
```
kmr-srbh: Please add back the comment.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1587,7 +1612,7 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, int64_t
     {fprintf(stderr,"Internal Error : default formatting error\n");exit(1);}
 
     if(format == NULL){
-        default_formatting(&result, result_size, &s_info);
+        default_formatting(&result, result_size, &s_info, false);
```
kmr-srbh: This change passes `false` to the `print_leading_space` argument which is the default behaviour and correct. With `--print-leading-space` flag, this should be passed as `true`, but I don't see a function call for that. This PR is fine with the current changes, but I think we need to make the appropriate function call for the flag as requested in https://github.com/lfortran/lfortran/pull/8572#issuecomment-3299421647. Please open a different PR for that change.

tests/tests.toml
```diff
@@ -4798,3 +4798,6 @@ asr = true
 [[test]]
 filename = "../integration_tests/select_type_13.f90"
 llvm_new_classes = true
+
+[[test]]
+filename = "test_list_directed.f90"
```
kmr-srbh: I think this test was missed during a commit push? :)

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1501,39 +1501,64 @@ void strip_outer_parenthesis(const char* str, int len, char* output) {
...
+    if (*result == NULL) {
+        fprintf(stderr, "Memory allocation failed\n");
+        exit(1);
+    }
+    if (print_leading_space) {
+        (*result)[0] = ' ';
+        (*result)[1] = '\0';
+        result_size = 1;
+    } else {
+        (*result)[0] = '\0';
```
kmr-srbh, line 1521: `result` is passed by reference and this value is already set inside `_lcompilers_string_format_fortran()` before calling `default_formatting()`. Could you please revert this change?

kmr-srbh (approved):
The changes look good! I left some comments.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: @assem2002 could you please help with this?

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: `result` is passed by reference and this value is already set inside `_lcompilers_string_format_fortran()` before calling `default_formatting()`. Could you please revert this change in the `else` block?

src/libasr/runtime/lfortran_intrinsics.c
assem2002: I don't believe we need it, yes. But I think it's causing no harm for now.
So let's keep it that way, Removing that `+1` could lead to some minor bugs that's totally unrelated to this PR.

kmr-srbh: > I don't believe we need it, yes. But I think it's causing no harm for now. So let's keep it that way, Removing that `+1` could lead to some minor bugs that's totally unrelated to this PR.

Yeah, that's right! :)

src/libasr/runtime/lfortran_intrinsics.c
mohitdhingra22: > 

@kmr-srbh I tried changing `default_spacing` to a single `char` as you mentioned, but it caused compilation errors on my side due to incompatible types with `strlen` and `strcpy` calls.
So I kept it as a `const char*` string literal (`" "`) because the code needs it in string form for those functions to work correctly.
Please let me know if you think there’s a better way to handle this or if I missed something.

kmr-srbh: Let's keep the original change then.

kmr-srbh:
@mohitdhingra22 Is this PR ready?

mohitdhingra22:
> @mohitdhingra22 Is this PR ready?

yes!

tests/tests.toml
certik: This test should be removed, or registered correctly.

test_list_directed.f90
```diff
@@ -0,0 +1,3 @@
+program listdirected
+    print *, '"','?','"'
+  end program listdirected
```
certik: This should not be in the root directory.

src/libasr/runtime/lfortran_intrinsics.c
certik: I think this has to be fixed here, since it's a regression, no? Or is this not working in main either.

kmr-srbh: Looking closely, I think the function argument `print_leading_space` and its related handling is not actually required, since the leading space is handled as a separate string with `--print-leading-space`. 

@mohitdhingra22 could you please use this diff to remove the function argument:

```diff
diff --git a/src/libasr/runtime/lfortran_intrinsics.c b/src/libasr/runtime/lfortran_intrinsics.c
index 3ffa7b331..d7ded801f 100644
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -1501,7 +1501,7 @@ void strip_outer_parenthesis(const char* str, int len, char* output) {
     }
 }
 
-void default_formatting(char** result, int64_t *result_size_ptr, struct serialization_info* s_info, bool print_leading_space){
+void default_formatting(char** result, int64_t *result_size_ptr, struct serialization_info* s_info){
     int64_t result_capacity = 100;
     int64_t result_size = 0;
     const int default_spacing_len = 1;
@@ -1513,11 +1513,6 @@ void default_formatting(char** result, int64_t *result_size_ptr, struct serializ
         fprintf(stderr, "Memory allocation failed\n");
         exit(1);
     }
-    if (print_leading_space) {
-        (*result)[0] = ' ';
-        (*result)[1] = '\0';
-        result_size = 1;
-    }
     bool prev_was_char = false;
 
     while(move_to_next_element(s_info, false)){
@@ -1546,7 +1541,7 @@ void default_formatting(char** result, int64_t *result_size_ptr, struct serializ
                 exit(1);
             }
         }
-        if(result_size > (print_leading_space ? 1 : 0) && !(prev_was_char && current_is_char)){
+        if(result_size > 0 && !(prev_was_char && current_is_char)){
             strcpy((*result) + result_size, default_spacing);
             result_size += default_spacing_len;
         }
@@ -1608,7 +1603,7 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, int64_t
     {fprintf(stderr,"Internal Error : default formatting error\n");exit(1);}
 
     if(format == NULL){
-        default_formatting(&result, result_size, &s_info, false);
+        default_formatting(&result, result_size, &s_info);
         free_serialization_info(&s_info);
         return result;
     }

```

mohitdhingra22:
i do not know how other commits are added to this i was just push and commit this https://github.com/lfortran/lfortran/pull/8572/commits/5212dab89b9c993e6b3ea06205a2b246a9c6549c
but along with other commits pushed here im trying to fixing by delete the files but conflicts cant be edited.

PR #8575, opened 2025-09-13
cli: accept --generate-object-code as legacy alias of --separate-compilation (CMake compat)

krystophny:
Context
- CMake’s FortranCInterface currently passes `--generate-object-code` when the compiler ID is `LFortran`.
- LFortran renamed this flag to `--separate-compilation`.

Change
- Accept `--generate-object-code` as a legacy alias of `--separate-compilation` in the CLI.
- Keeps configure-time compatibility with current CMake while upstream adjusts.

Why
- Without this, `try_compile` in FortranCInterface can fail or mis-detect, preventing projects like LAPACK from configuring with LFortran using current CMake.
- Upstream CMake is being updated; this is a temporary shim until those releases are widely available.

Verification
- With this change, FortranCInterface detection and verification succeed when configuring LAPACK.
- Example configure:
  cmake .. -DCMAKE_Fortran_COMPILER=lfortran -DCBLAS=no -DLAPACKE=no -DBUILD_TESTING=no -DBUILD_DOUBLE=no -DBUILD_COMPLEX=no -DBUILD_COMPLEX16=no -DLAPACKE_WITH_TMG=no \
           -DCMAKE_Fortran_PREPROCESS=yes \
           -DCMAKE_Fortran_FLAGS="--fixed-form-infer --implicit-interface --separate-compilation --legacy-array-sections" \
           -DCMAKE_Fortran_COMPILE_OPTIONS_PREPROCESS_ON=--cpp \
           -DCMAKE_Fortran_COMPILE_OPTIONS_PREPROCESS_OFF=--no-cpp

Scope
- File: src/bin/lfortran_command_line_parser.cpp
- Behavior: `--generate-object-code` toggles the same internal option as `--separate-compilation`.

Notes
- This does not change codegen semantics beyond honoring the legacy name.
- Once CMake releases with updated behavior are widely available, this alias can remain for backward compatibility or be reconsidered.


krystophny:
Upstream CMake issue tracking the legacy flag: https://gitlab.kitware.com/cmake/cmake/-/issues/27225

Pranavchiku (commented):
With this we now support `--generate-object-code` as well which is deprecated. I'm not sure if we have sent patch to CMake to support this, if not then this is fine but in long run I'm -1 to keep this alias. I'll loop in @certik for more thoughts.

krystophny:
@Pranavchiku I understand that this was intentionally changed. I am myself not sure what's the best way. The problem right now is that it will take long for the CMake fix to roll out, and in the meantime, people won't be able to use LFortran with CMake in such cases. Let's discuss further, I don't mind closing the PR either and find another workaround.

krystophny:
PS: I have opened a PR on CMake repo already, see https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11184

Pranavchiku:
Perfect, we can then merge this and once https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11184 gets in, we can disable this alias! Thank you so much for creating a PR upstream :)

I'm good to merge now, let's wait and see what @certik has to say.

krystophny:
Thanks @Pranavchiku ! What we could do is add a depreciation warning though. What do you think?

PR #8587, opened 2025-09-15
Change `stringItem` implementation in LLVM backend

assem2002:
Changes :
- Dump the usage of a C-runtime function to return a character corresponding to the addressed string item.
- Simply use a stringView (GEP on string data based on the index + length of 1)
- Add llvm utility `create_stringView` that accepts (`ASR::String` to deduce the physicalType, Data, len)

Benefits :
- No memory leaks with stringItems
- StringItems expression returns a reference to the original string

Tests :
- Registerd test to exercise the return be reference for StringItem expression.


certik:
CI fails.

PR #8585, opened 2025-09-15
Allocate string descriptor on stack instead of heap

assem2002:
We used to create the descriptor on heap to avoid problems related to function-returns.
now : 
- we don't return strings (We use the subroutine and intentOUT arg trick. We still have some cases to go) 
- String returns for **StringDescriptor** in LLVM backend are returns by value (so the main concern is now gone).

This should eliminate string-descriptor-creation leaks.

PR #8588, opened 2025-09-16
Uncomment FAST POT3D test

assem2002:
I think it might be a consequent of changes in this PR #8585. I'll test different LLVM versions In the CI and see how it goes.


assem2002:
I think it's fine to merge.

PR #8578, opened 2025-09-14
Official LAPACK compiling

krystophny:
Now with

```
cmake -DCMAKE_Fortran_COMPILER=lfortran \
            -DCMAKE_Fortran_FLAGS="--fixed-form-infer --implicit-interface" \
            -DBUILD_INDEX64_EXT_API=OFF -DBUILD_COMPLEX=no -DBUILD_COMPLEX16=no ..
```
Only remaining issues are that they use hardcoded `-cpp` for `BLAS` part, and issues with the complex numbers.

PR #8598, opened 2025-09-19
Fix complex captures for legacy third-party builds

krystophny:
## Summary
- force LLVM complex comparisons and logical casts to load real/imag parts using the proper kinded struct type
- keep captured-variable pointers local per function so nested procedures receive valid arguments
- add conservative constant folding fallbacks so complex unary patterns stay evaluable

## Testing
- micromamba run -n llvm20 bash -lc 'export PATH=$PWD/build/src/bin:$PATH; FFLAGS="--debug-with-line-column" ./integration_tests/run_tests.py -b llvm -j8'
- PATH=build/src/bin:$PATH lfortran --legacy-array-sections --fixed-form --implicit-interface /tmp/clasyf_aa.f -c
- micromamba run -n llvm20 bash -lc 'export PATH=/home/ert/code/lfortran/build/src/bin:$PATH; cd /tmp/minpack_modern; lfortran ./src/minpack.f90 -c --legacy-array-sections; lfortran ./examples/example_lmder1.f90 --legacy-array-sections'
- micromamba run -n llvm20 bash -lc 'export PATH=/home/ert/code/lfortran/build/src/bin:$PATH; cd /tmp/lapack-3.12.0; cmake -S . -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE=/tmp/lfortran_toolchain.cmake -DCMAKE_Fortran_FLAGS="--fixed-form-infer --implicit-interface --legacy-array-sections --cpp" -DBUILD_INDEX64_EXT_API=OFF -DBUILD_COMPLEX=OFF -DBUILD_COMPLEX16=OFF -DBUILD_TESTING=OFF -DCBLAS=OFF -DLAPACKE=OFF; cmake --build build -j8'
- micromamba run -n llvm20 bash -lc 'export PATH=/home/ert/code/lfortran/build/src/bin:$PATH; cd /tmp/fortran_mpi/tests; FC="lfortran --cpp" ./run_tests.sh'
- micromamba run -n llvm20 bash -lc 'export PATH=/home/ert/code/lfortran/build/src/bin:$PATH; cd /tmp/fortran_mpi/tests; FC="lfortran --cpp --separate-compilation" ./run_tests.sh'


krystophny:
@certik this is now the relevant PR for LAPACK support. Integration and thirdparty tests pass for LLVM11, LLVM20 and LLVM21. Still waiting for CI and unit tests. Let me know what you think!

krystophny:
Here's a quick rationale for the big semantic block so reviewers can tie it back to the legacy failures we saw:

* **Why:** LAPACK/Minpack routinely pass array elements through implicit interfaces while expecting the callee to see the full descriptor. Example: `call slarft('F','C',n,k,work(1,1),ldwork,...)`. We need to hand the code generator an actual that already behaves like `work(:,1)` (with concrete bounds) so LLVM sees a raw pointer instead of a descriptor slice.

* **Where:** `legacy_array_sections_helper()` now delegates to three small helpers (see `src/lfortran/semantics/ast_common_visitor.h` lines 6738–6908):  
  * `legacy_resolve_formal_dims()` rebuilds the descriptor shape for the callee’s formal by substituting any `FunctionParam`s with the actual call arguments.  
  * `legacy_promote_array_item()` upgrades `ArrayItem` actuals to `ArraySection`—fills in default left/step bounds, then using the resolved descriptor to assign a matching descriptor type.  
  * `legacy_promote_array_expr()` handles the non-`ArrayItem` cases and emits an `ArrayPhysicalCast` when the formal expects a pointer descriptor (pointer array/unbounded pointer array).

* **Result:** By the time we reach `create_Function()` the arg list already contains the correct `ArraySection` or `ArrayPhysicalCast`. Codegen no longer has to guess the bounds or insert casts ad hoc—especially important on LLVM ≤16 where typed pointers require the exact element pointer ahead of time.

This is what keeps legacy CALL sites like `sgejsv/slarft/sorm22` from tripping the old “descriptor pointer mismatch” assertions on LLVM 11 while leaving non‐legacy codepaths untouched.

certik:
@krystophny great job! I am running all tests. @Pranavchiku, @jinangshah21, @assem2002 do you mind reviewing it please?

ci/test_third_party_codes.sh
```diff
@@ -868,6 +868,57 @@ time_section "🧪 Testing SNAP" '
...
+  # Use latest stable release
+  git checkout v3.12.0
+
+  python - <<PY
+from pathlib import Path
+path = Path("LAPACKE/include/CMakeLists.txt")
+needle = "  FortranCInterface_VERIFY()"
+replacement = (
+    "  if (CMAKE_Fortran_COMPILER MATCHES \"lfortran\")\n"
+    "    message(STATUS \"Skipping FortranCInterface_VERIFY for LFortran\")\n"
```
assem2002, line 881: If It's failing due to something that we should support it'd be a good chance to open an issue for that to be fixed later.

krystophny:
The latest Exhaustive checks (run 18061693772 on commit 66a5f451fe5f3d4913799f98cf0205a8d211634b) are failing on all high-LLVM jobs (`Test LLVM 17`, `18`, `19`, `20`, `21`). Each job crashes while building the new FAST integration targets for legacy array sections:

```
legacy_array_sections_02_FAST
legacy_array_sections_05_FAST
```

The build log shows `LCOMPILERS_ASSERT` triggering inside `visit_ArraySection()` in `src/libasr/codegen/asr_to_llvm.cpp:2997`:

```
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/codegen/asr_to_llvm.cpp
function visit_ArraySection(), line number 2997 at
ASR::is_a<ASR::String_t>(*ASRUtils::expr_type(x.m_v)) && n_dims == 0
```

The new legacy lowering now promotes numeric actuals to `ArraySection`, so when the FAST variants of `integration_tests/legacy_array_sections_02.f90` and `_05.f90` are compiled with LLVM ≥17 (opaque pointers), codegen hits this assert because `visit_ArraySection()` still assumes the base expression is a scalar string. LLVM ≤16 jobs (and the quick checks) stay green because they never walk through this path.

To reproduce locally, use the micromamba LLVM 21 environment referenced in the PR description, build, and run the affected integration targets:

```
micromamba activate lf-llvm21
cmake -S . -B build-llvm21 -G Ninja -DWITH_LLVM=ON -DLLVM_VERSION=21
cmake --build build-llvm21 -j
cd integration_tests
FFLAGS="--legacy-array-sections" ./run_tests.py -b llvm -t legacy_array_sections_{02,05}
```

This should reproduce the crash at `visit_ArraySection()` on LLVM 21.

krystophny:
## Additional Fixes in 939c950d0
| Area | Why it exists |
| --- | --- |
| `src/libasr/codegen/asr_to_llvm.cpp` | Adds `create_array_section_descriptor()` so legacy array sections now lower to proper descriptors before LLVM codegen. This lets the LLVM ≥17 backend emit pointer arguments for numeric slices without tripping the old string-only assert. `visit_ArraySection()` falls back to the original string path but delegates all array cases to the new helper, while `handle_array_section_association_to_pointer()` reuses it to keep pointer assignments in sync. |

**Repro** (LLVM ≥17 opaque pointers):
1. `micromamba activate lf-llvm21`
2. `cmake -S . -B build-llvm21 -G Ninja -DWITH_LLVM=ON -DLLVM_VERSION=21`
3. `cmake --build build-llvm21 -j`
4. `build-llvm21/src/bin/lfortran --legacy-array-sections --fast -c integration_tests/legacy_array_sections_02.f90`
5. `build-llvm21/src/bin/lfortran --legacy-array-sections --fast -c integration_tests/legacy_array_sections_05.f90`

The assertions reported in Exhaustive checks disappear after step 3; steps 4–5 confirm the compiler accepts both previously failing MREs.

krystophny:
Quick note on the latest push:

**Summary**
| Area | Why it exists |
| --- | --- |
| `src/libasr/utils.h` | `PassOptions` now tracks `legacy_array_sections`, so downstream passes can respect the flag. |
| `src/libasr/pass/promote_allocatable_to_nonallocatable.cpp` | Skip the promotion pass entirely when legacy sections are active; this avoids the FAST-mode descriptor copies that broke pointer-actual semantics. |
| `src/libasr/codegen/asr_to_llvm.cpp`, `src/libasr/codegen/asr_to_wasm.cpp`, `src/bin/lfortran.cpp`, `src/lfortran/fortran_evaluator.cpp` | Thread the new pass option through every driver that invokes the pass manager (CLI, evaluator, LLVM/WASM backends) so the skip applies consistently. |

**Validation Matrix**
| Toolchain | Status | Notes |
| --- | --- | --- |
| LLVM 21 (`lf-llvm21`) | ✅ | `lfortran --legacy-array-sections --fast integration_tests/legacy_array_sections_{01,02,05}.f90` now prints the updated slices; also recompiled `src/minpack.f90` + `examples/example_lmder1.f90` under the same flags. |

(I didn’t rerun the micromamba-based third-party sweep here; that still depends on the CI env setup.)

ci/test_third_party_codes.sh
krystophny: This should solve itself with newer CMake version. I didn't see it anymore with CMake 4.0 . Some issues with compiler flags that are still assumed to be gfortran ones in older CMake versions.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -354,7 +354,11 @@ namespace LCompilers::CommandLineInterface {
 
         // Decide if a file is fixed format based on the extension
         // Gfortran does the same thing
-        if (opts.fixed_form_infer && endswith(opts.arg_file, ".f")) {
+        if (!compiler_options.fixed_form && (
+                endswith(opts.arg_file, ".f") ||
+                endswith(opts.arg_file, ".F") ||
+                endswith(opts.arg_file, ".for") ||
+                endswith(opts.arg_file, ".FOR"))) {
```
certik, line 368: You can use the `--fixed-form-infer` instead, which already does this. 

The above code makes `--fixed-form-infer` the default, but that has two issues:

* We want to keep the door open to use `.f` for free form (it was proposed several years ago and I still quite like it, I think we should not use fixed form, and if we have to, we can use a compiler option)
* It doesn't seem to allow to treat `.f` as free form at all.

Either way this change should not be part of this PR.

integration_tests/CMakeLists.txt
```diff
@@ -2247,6 +2250,9 @@ RUN(NAME dealloc_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_NEW_CLASSES
 
 RUN(NAME double_complex_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME lapack_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME lapack_03 LABELS gfortran llvm EXTRA_ARGS --fixed-form --implicit-interface --legacy-array-sections GFORTRAN_ARGS -ffixed-form) # LAPACK slarft recursive call with sequence association
+RUN(NAME lapack_04 LABELS gfortran llvm EXTRA_ARGS --fixed-form --implicit-interface --legacy-array-sections GFORTRAN_ARGS -ffixed-form) # LAPACK sorm22 WORK array sequence association
+RUN(NAME lapack_02 LABELS gfortran llvm EXTRA_ARGS --fixed-form --implicit-interface --legacy-array-sections GFORTRAN_ARGS -ffixed-form) # LAPACK sgejsv scalar/array ambiguity with SLASCL
```
certik: Very minor: I would order it by number.

integration_tests/lapack_05.f90
```diff
@@ -0,0 +1,30 @@
+program lapack_name_data
```
certik, line 1: This test is not registered.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1623,6 +1623,7 @@ namespace LCompilers {
 
     llvm::Value* LLVMUtils::CreateGEP2(llvm::Type *t, llvm::Value *x,
             std::vector<llvm::Value *> &idx) {
+        x = builder->CreateBitCast(x, t->getPointerTo());
         return builder->CreateGEP(t, x, idx);
     }
```
certik, line 1635: What implications does this change have? Now it is not just creating a GEP, but also inserting a cast before it. Why is it needed and is this the right place to have it?

Is this needed for every GEP?

src/bin/lfortran_command_line_parser.cpp
krystophny: Changed this already, some cache issue in github diff view. is it fine like that? 

Image

krystophny:
## Analysis: BitCast is Not the Right Solution

After setting up LLVM 21 environment and reproducing the original failure, I've determined that **the BitCast additions are masking a deeper issue in semantic analysis**.

### Root Cause
The problem occurs **before LLVM codegen** - in the semantic transformations. When using `--legacy-array-sections`, the code:
```fortran
call sub(a(1,j))  ! where a is double precision :: a(16,3)
```
Should transform `a(1,j)` (ArrayItem returning scalar f64) into `a(1:16,j)` (ArraySection returning f64[16]) to match the formal parameter `a(16)`.

### Current Issue
The `legacy_array_sections_helper()` transformations aren't working correctly, so `a(1,j)` remains a scalar, causing:
```
Type mismatch in subroutine call, expected `f64[16]`, passed `f64`
```

### Why BitCast "Fixed" It
The BitCast likely masked type mismatches in LLVM codegen, allowing incorrectly typed operations to proceed, but the fundamental semantic issue remained unresolved.

### Recommendation
Instead of adding BitCast to `CreateGEP2` (which is too broad and affects all GEP operations), we should:
1. **Fix the semantic transformations** to properly promote ArrayItem → ArraySection 
2. **Remove the BitCast workarounds** from llvm_utils.cpp

The changes to `CreateGEP2` are too far-reaching and address the wrong layer of the problem.

src/libasr/codegen/llvm_utils.cpp
krystophny: That's really a problem. It hid also another bug which is now in an integration test `lapack_06`. For now I tried some things but I am now stuck unfortunately. Let's see...

certik:
Can you post the old branch somewhere --- don't override it with rebase. All tests passed, so that's a big deal. The fix wasn't quite correct, but it's good to use it for reference.

> call sub(a(1,j))
> Should transform a(1,j) (ArrayItem returning scalar f64) into a(1:16,j) (ArraySection returning f64[16]) to match the formal parameter a(16).

Actually, that's only if you have access to the `sub`'s definition. But typically you don't in these old codes. So you only see `sub(a(1,j))` and you do not know if `sub` accepts an integer, or a slice. As a solution that we used in the past was to use a `bind(c)` and just pass a pointer as `bind(c)`. Then in the `sub` function we dereference the pointer and use either one element, or multiple elements, as needed.

krystophny:
> Can you post the old branch somewhere --- don't override it with rebase. All tests passed, so that's a big deal. The fix wasn't quite correct, but it's good to use it for reference.
> 
> > call sub(a(1,j))
> > Should transform a(1,j) (ArrayItem returning scalar f64) into a(1:16,j) (ArraySection returning f64[16]) to match the formal parameter a(16).
> 
> Actually, that's only if you have access to the `sub`'s definition. But typically you don't in these old codes. So you only see `sub(a(1,j))` and you do not know if `sub` accepts an integer, or a slice. As a solution that we used in the past was to use a `bind(c)` and just pass a pointer as `bind(c)`. Then in the `sub` function we dereference the pointer and use either one element, or multiple elements, as needed.

Clear! The "good" commit is [0f492c2](https://github.com/lfortran/lfortran/pull/8598/commits/0f492c2d9ff1cb7b8612d3d9638adb17d1c2e05a) and still intact for reference. The other branch that did not use legacy array sections and would have made it much more difficult with LLVM11 has intact commit [af66702](https://github.com/lfortran/lfortran/pull/8578/commits/af667029ff361578b1304d943b8a153955eb9454) with passing LLVM20 tests.

certik:
Ok, I pushed it here so that we don't lose it: https://github.com/certik/lfortran/tree/lapack_fix

PR #8599, opened 2025-09-20
Exposed StrOrd to Frontend

swamishiju:
- Added tests and exposed StrOrd to LF frontend

### DIFF
--- a/integration_tests/list_test_09_.f90
+++ b/integration_tests/list_test_09_.f90
@@ -56,7 +56,7 @@ contains
 
       if (_lfortran_len(c) /= 5) error stop
       do i = 0, 4
-         if (iachar(_lfortran_get_item(c, i)) - iachar('a') /= i) error stop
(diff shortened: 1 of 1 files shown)


swamishiju:
Python's `ord` and Fortran's `iachar` serve the same purpose as far as I can tell, do we want to keep both?

integration_tests/list_test_09_.f90
```diff
@@ -56,7 +56,7 @@ subroutine test_list_concat()
 
       if (_lfortran_len(c) /= 5) error stop
       do i = 0, 4
-         if (iachar(_lfortran_get_item(c, i)) - iachar('a') /= i) error stop
+         if (_lfortran_ord(_lfortran_get_item(c, i)) - _lfortran_ord('a') /= i) error stop
```
swamishiju, line 59: *as said in my previous comment

certik: Will the previous code still work?

swamishiju:
Yes the previous code will still work, exposing and fixing `ord` just fixes another LP test

PR #8608, opened 2025-09-22
fix: [new_classes] do deepcopy for derived types assignments

jinangshah21:


certik:
The failure seems unrelated: https://github.com/lfortran/lfortran/issues/8610.

PR #8616, opened 2025-09-23
enh: [new_classes] support `arrayitem` in runtime type polymorphism call

jinangshah21:


certik:
The error is:
```
LCOMPILERS_ASSERT failed: /Users/runner/work/lfortran/lfortran/lfortran-0.56.0-447-g004512f02/src/libasr/codegen/llvm_utils.cpp
function CreateGEPDeprecated(), line number 1615 at 
type
```

Issue #8594, opened 2025-09-18
Using unallocated variable must give an error


certik:
```
program expr2
implicit none

integer, allocatable :: x
integer :: y

y = x

end program
```
This should give a runtime error `x` is not allocated, in Debug mode.

PR #8617, opened 2025-09-23
feat: throw runtime error if scalar variable is read but not allocated

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8594


xaerru:
Yes, with `--fast` we don't throw the runtime error.

PR #8621, opened 2025-09-24
fix: [new_classes] handle runtime polymorphic call in select type

jinangshah21:


certik:
Is this fixing something in 3rd party tests that is not caught by our integration tests? If so, can you please add an integration test for this also?

jinangshah21:
Added test

integration_tests/CMakeLists.txt
```diff
@@ -1400,6 +1400,7 @@ RUN(NAME select_type_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_NEW_CLASSES)
 RUN(NAME select_type_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_type_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --new-classes)
```
certik, line 1403: I am guessing this test doesn't run with old classes, that's why we only need to run it with new classes.

jinangshah21: Yes

PR #8626, opened 2025-09-25
fix: load array var properly in getting vptr on polymorphism call

jinangshah21:


certik:
The error is:
```
ld: pointer not aligned in '_color'+0x105 (/private/var/folders/q0/wmf37v850txck86cpnvwm_zw0000gn/T/tmp.7lgrDB64cK/assert/jonquil/build/lfortran_90027A7654EF83E9/jonquil/test_unit_main.f90.o)
The command 'clang -o build/lfortran_CB1B3243CF22A21F/test/jqtest build/lfortran_90027A7654EF83E9/jonquil/test_unit_main.f90.o build/lfortran_90027A7654EF83E9/jonquil/test_unit_test_lexer.f90.o build/lfortran_90027A7654EF83E9/jonquil/libjonquil.a  -L"/Users/runner/work/lfortran/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/runner/work/lfortran/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13229,7 +13229,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             // Get VTable pointer
             if (ASR::is_a<ASR::ArrayItem_t>(*x.m_dt)) {
                 ASR::ArrayItem_t* array_item = ASR::down_cast<ASR::ArrayItem_t>(x.m_dt);
-                this->visit_expr_wrapper(array_item->m_v, true);
+                ptr_loads = 1;
+                this->visit_expr_wrapper(array_item->m_v, ASR::is_a<ASR::StructInstanceMember_t>(*array_item->m_v));
+                ptr_loads = ptr_loads_copy;
```
kmr-srbh, line 13302: I think we should do this only based on the `--new-classes` flag?

jinangshah21: This will run with `new_classes` only.

certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/8766.

Issue #8601, opened 2025-09-20
Handle calling a function with an argument and assigning the result to the same argument


akramhany:
In below fortran code:
```fortran
module elemental_17_mod
  implicit none
  type string_t
    character(len=:), allocatable :: string_
  end type string_t
contains

  function bracket(x) result(res)
    class(string_t), intent(in) :: x
    type(string_t) :: res

    res%string_ = x%string_ // " called func"
  end function bracket

end module elemental_17_mod


program elemental_17
  use elemental_17_mod
  implicit none

  type(string_t) :: lines

  lines%string_ = "Hello"
  lines = bracket(lines)         ! Here

  print *, lines%string_
end program elemental_17
```

In the line `lines = bracket(lines)`, the `bracket` function would be converted to a subroutine, and the function call would be converted to a subroutine call like this `call bracket(lines, lines)`, but it's not a good idea to pass `lines` both as an input and output at the same time, so it has to be reworked to something like this:
```sh
make a temp var with same type as lines
call bracket(lines, temp)
lines = temp
```

akramhany:
I tried to solve this issue in the `subroutine_from_function` ASR pass, where we convert the function call to a subroutine call, but I encountered some edge cases that need to be handled separately:
- If we are passing an array to a function that it's dimensions are not known yet (e.g. assumed shape), then in compile time we won't be able to know it's dimensions, so we won't be able to create the temp variable:
```fortran
program arrays_58

real :: A(2,2) = reshape([1.0,2.0,3.0,4.0], [2,2])
call trans(A)
print *, A
if ( any(abs( A - 12.91 ) > 1e-8) ) error stop 
contains

subroutine trans(A)
real, intent(inout) :: A(:, :)
A = matprod(transpose(A))           ! HERE we don't know the dimensions of A in compile time

end subroutine

function matprod(x) result(k)
real,intent(in) :: x(:,:)
real :: k(size(x, 1), size(x, 2))
k = 12.91
end function

end program 
```

- If the argument that we should create a temp for (e.g. `lines`) is allocatable, it has 3 cases:
    - It might be allocated before being passed to the subroutine, so we would need to create a temp and allocate it too, before passing it to the subroutine.
    - It might be allocated inside the subroutine, so we don't need to allocate the temp, but we would need to allocate the original variable (`lines`).
    - It might not be allocated anywhere at all, and hence we shouldn't allocate anything too.

  The problem is that we don't know which case of the above 3 it is.

[...]

Issue #8610, opened 2025-09-22
MPI: The PMIx server's listener thread failed to start.


certik:
https://github.com/lfortran/lfortran/actions/runs/17925376034/job/50970052015?pr=8608
```
2025-09-22T19:50:21.5703350Z Element 10:        60.00
2025-09-22T19:50:21.5703500Z All values verified successfully
2025-09-22T19:50:21.5766400Z [0;32mTest allreduce_4 with 4 MPI ranks PASSED![0m
2025-09-22T19:50:21.5815930Z [1;33mCompiling barrier_1...[0m
2025-09-22T19:50:22.2704090Z [1;33mRunning barrier_1 with 1 MPI ranks...[0m
2025-09-22T19:50:23.4110190Z Process     0     reached before the barrier.
2025-09-22T19:50:23.4219110Z Process     0     passed the barrier.
2025-09-22T19:50:23.4336850Z [0;32mTest barrier_1 with 1 MPI ranks PASSED![0m
2025-09-22T19:50:23.4443960Z [1;33mRunning barrier_1 with 2 MPI ranks...[0m
2025-09-22T19:50:24.5865970Z Process     1     reached before the barrier.
2025-09-22T19:50:24.5967280Z Process     1     passed the barrier.
2025-09-22T19:50:24.6023320Z Process     0     reached before the barrier.
2025-09-22T19:50:24.6050850Z Process     0     passed the barrier.
2025-09-22T19:50:24.6051640Z [0;32mTest barrier_1 with 2 MPI ranks PASSED![0m
2025-09-22T19:50:24.6052010Z [1;33mRunning barrier_1 with 4 MPI ranks...[0m
2025-09-22T19:50:25.6908580Z Process     0     reached before the barrier.
2025-09-22T19:50:25.6930810Z Process     1     reached before the barrier.
2025-09-22T19:50:25.6931690Z Process     3     reached before the barrier.
2025-09-22T19:50:25.6933030Z Process     2     reached before the barrier.
2025-09-22T19:50:25.6934500Z Process     0     passed the barrier.
2025-09-22T19:50:25.6936210Z Process     3     passed the barrier.
2025-09-22T19:50:25.6937090Z Process     1     passed the barrier.
2025-09-22T19:50:25.6937410Z Process     2     passed the barrier.
2025-09-22T19:50:25.7003820Z [0;32mTest barrier_1 with 4 MPI ranks PASSED![0m
[...]

PR #8634, opened 2025-09-28
[DON'T MERGE] enh: Create temp var when passing using a var as input and output with functions

akramhany:
Towards #8601.


akramhany:
How can I register reference tests and make them run till all ASR passes are applied? Because using `asr = true` seems to only generate the ASR, but doesn't apply the passes.

akramhany:
Also there is some refactoring that needs to be done (marked by TODO), and I will do it in another PR to not complicate things here.

certik:
The failure is unrelated: https://github.com/lfortran/lfortran/issues/8610, I restarted it.

certik:
Are you able to improve this PR to pass some test that fails in main? To ensure these changes are actually needed.

akramhany:
> Are you able to improve this PR to pass some test that fails in main? To ensure these changes are actually needed.

Yes of course!
I opened an [issue](https://github.com/lfortran/lfortran/issues/8649) for a MRE that is producing wrong output with current main, but is fixed with this PR.
I will add it, along with another test case I found, as integration tests.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -225,18 +225,47 @@ class ReplaceFunctionCallWithSubroutineCallVisitor:
...
                 s_args.push_back(al, fc->m_args[i]);
+
+                if (ASRUtils::expr_equal(target, fc->m_args[i].m_value)) {      //TODO: Write a utility function that checks if 2 expr are the SAME
+                    use_temp_var_for_return = true;
+                }
+            }
+
+            if (fc->m_dt && ASRUtils::expr_equal(target, fc->m_dt)) {       //TODO: Write a utility function that checks if 2 expr are the SAME
+                use_temp_var_for_return = true;
+            }
```
certik, line 242: This determines if to create a temporary. It only does it if any argument is equal to the result. That is correct. I am sure there might be a bunch of corner cases, but to start this is good.

PR #8632, opened 2025-09-27
Issue 8603

akramhany:
Fixes #8603.

This PR:
1. Fixes the MRE listed in the issue.
2. Checks on 2 cases when there are no custom operators for the expression and prints an error instead of assertion failing.

PR #8650, opened 2025-10-01
tests: make the custom_operator_09 test stop if output is wrong

akramhany:
Subsequent PR to #8632.

### DIFF
--- a/integration_tests/custom_operator_09.f90
+++ b/integration_tests/custom_operator_09.f90
@@ -33,6 +33,6 @@ program custom_operator_09
 
   greeting_obj = prefix // name_obj
 
-  print *, "result: ", greeting_obj % value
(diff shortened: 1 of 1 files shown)


integration_tests/custom_operator_09.f90
```diff
@@ -33,6 +33,6 @@ program custom_operator_09
 
   greeting_obj = prefix // name_obj
 
-  print *, "result: ", greeting_obj % value
+  if (greeting_obj%value /= "hello,    world!") error stop
```
certik, line 36: Leave both print and if statement, that way when you run it by hand you can see the output.

certik:
The failure is unrelated https://github.com/lfortran/lfortran/issues/8610.

PR #8653, opened 2025-10-02
Advance `subroutine_from_function` pass after `intrinsic_function` pass

assem2002:
...


src/libasr/ASR.asdl
```diff
@@ -33,7 +33,7 @@ stmt
     | ReAlloc(alloc_arg* args)
     | Assign(int label, identifier variable)
     | Assignment(expr target, expr value, stmt? overloaded, bool realloc_lhs, bool move_allocation)
-    | DebugCheckArrayBounds(expr target, expr value, expr orig_value, bool move_allocation)
+    | DebugCheckArrayBounds(expr target, expr value, expr? orig_value, bool move_allocation)
```
certik, line 36: New semantics:
```suggestion
    | DebugCheckArrayBounds(expr target, expr value, expr? binop_value, bool move_allocation)
```
If the RHS is a binop, then binop_value is always present, otherwise it is never present. Let's enforce this in verify.

`value` is always a type integer, and it's an ArraySize of an array, or an integer Var (or maybe an expression like `n+1`) or IntegerConstant directly.

`target` is also a type integer, a size, just like `value`.

For example:

`x = a+b`,

The `target` will be `ArraySize(x)`, the `binop_value` will `a+b`, and `value` will be only `ArraySize(a)`.
So: `DebugCheckArrayBounds(ArraySize(x), ArraySize(a), a+b, false)`

The idea is that `value` is some simple array whose size we can get at runtime.

Idea: remove binop_value completely, so we'll have just `DebugCheckArrayBounds(expr target, expr value, bool move_allocation)`

Insert another `DebugCheckArrayBounds(ArraySize(a), ArraySize(b), false)`. So the idea is to generate two nodes:
```
DebugCheckArrayBounds(ArraySize(a), ArraySize(b), false, BinOpBoundsError)
DebugCheckArrayBounds(ArraySize(x), ArraySize(a), false, AssignmentBoundsError)
```

certik: I made this https://github.com/lfortran/lfortran/issues/8726.

doc/src/passes/subroutine_from_function.md
```diff
@@ -0,0 +1,74 @@
...
+```
+========   How It Works?   ========
+    1. We use entry static function `Allocate()` to prevent using other functions by mistake.
+    2. Pass to it Var to allocate + FunctionCall we're allocating against its return type + Other helper members.
+    3. We get function return type (the one with FunctionParams).
+    4. We call the replacer on the type.
+    5. The replacer replaces FunctionParam nodes with arguments in the functionCall.
+    6. FunctionCalls arguments are replaced with temporaries in both FunctionParam site and also the functionCall's argument site.
+    7. Any other argument expression is just used normally.
+    8. Then we insert an allocate statement (the main purpose of this class)
```
assem2002, line 55: @certik Do you have any opinion upon the process? Do you think it could be slowing things down?

certik: I am having hard time following this document. It's documenting two passes, implemented by C++ classes?

Yes, this code seems correct to me:
```
    temp_var = f()
    allocate(character(len=funcCall_temp_var) :: return_slot)
    call foo(temp_var, return_slot)
    print *, return_slot
```

Regarding the code of the pass itself, we just need to review it.

assem2002: Documentation is for classes 2 classes -- one that walks through the whole ASR (partially) to transform functions into subroutines, the other one is doing a particular job which is allocating a temp variable with the function return type -- it walks on small portion of the ASR to do the job of allocation properly, We use an entry static function for that class to avoid using visitors mistakenly by the user of the class  .

certik (approved):
I reviewed everything. I didn't see any big issues. I would say let's merge it and if we discover issues, we'll address them later.

Issue #8283, opened 2025-08-07
Julienne blocker: `select rank`


rouson:
Steps to reproduce with the Fortran Package Manager ([`fpm`](https://github.com/fortran-lang/fpm)) installed.
```
git clone https://github.com/berkeleylab/julienne
cd julienne
fpm test --compiler lfortran --flag --cpp
```
which yields
```
LCompilersException: visit_SelectRank() not implemented
```
This is not hard to work around, but working around it does require a laborious chore so it would be great to save some time by having support for this feature.

certik:
I fixed one bug and improved the error message at least: https://github.com/lfortran/lfortran/pull/8284

It looks like we do not support select rank at all right now. If this would be helpful to you, we can implement it. 

`select rank` is a runtime feature, right? It figures out at runtime what rank the array is, and executes a different block based on the rank. It doesn't know at compile time. Unless it generates a different version of the function per rank, but that might be impractical if the function has several arguments and we do select rank on each.

sebastian-mutz:
> If this would be helpful to you, we can implement it.

I just want to flag here that stdlib also uses `select rank` (e.g., in [stdlib_error](https://github.com/fortran-lang/stdlib/blob/stdlib-fpm/src/stdlib_error.f90) ), making it harder to work around in projects with stdlib as a dependency. So if stdlib keeps using it, implementing the feature would at least be indirectly helpful for all projects using stdlib + lfortran.

certik:
We are implementing it, @HarshilShah1804 is working hard to get a prototype working.

bonachea:
This issue appears to be subsumed by #7812 and should probably be closed.

HarshilShah1804:
Yes, Thanks @bonachea.

certik:
Yes, select rank is fully implemented and tested now.

PR #8654, opened 2025-10-02
enh: initial implementation of `select rank`

HarshilShah1804:
Towards #8283

### DIFF
--- a/integration_tests/select_rank_01.f90
+++ b/integration_tests/select_rank_01.f90
@@ -1,0 +1,31 @@
+
+program select_rank_test
+    implicit none
+
+    integer :: b(2) = [1,2]
+    integer :: c(2,2) = reshape([1,2,3,4],[2,2])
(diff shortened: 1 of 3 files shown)


integration_tests/select_rank_01.f90
```diff
@@ -0,0 +1,31 @@
...
+
+    subroutine check(x, expected)
+        integer, intent(in) :: x(..)
+        integer, intent(in) :: expected
+
+        select rank(x)
+            rank(0)
+                if (expected /= 0) error stop
+            rank(1)
+                if (expected /= 1) error stop
```
certik, line 23: In subsequent PRs: Inside the rank(1) you can now access `x` as a 1D array. So to implement it, add a test, and use casting in AST->ASR, and possibly a temporary variable.

HarshilShah1804:
As all checks have passed, I am merging this PR.

PR #8656, opened 2025-10-03
enh: check that target and value expressions are exactly the same

### LINKED ISSUE #8601: Handle calling a function with an argument and assigning the result to the same argument

akramhany:
Subsequent to #8634.
Fixes #8601.


certik:
CI fails

PR #8658, opened 2025-10-03
fix: add error handling for assumed-rank arrays

HarshilShah1804:


certik:
CI fails

HarshilShah1804:
I have fixed the issue causing CI to fail. Now it shall work.

HarshilShah1804:
I don't think we would require exhaustive checks for this. The changes made are quite simple and shall not fail in other LLVM version. What do you say @certik ?

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -115,6 +115,22 @@ semantic error: Dummy argument 'c' not defined
 59 |        END SUBROUTINE faulty_subroutine
    | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 
+semantic error: Assumed-rank variable 'x' must not have any dimensions specified.
+  --> tests/errors/continue_compilation_2.f90:82:20
+   |
+82 |         integer :: x(.., 5)
+   |                    ^^^^^^^^ 
+
+semantic error: Dummy argument 'x' not defined
```
certik, line 124: This error is I think caused by the previous error, so not great to report. But that can be fixed later.

certik:
I restarted the failing test. If it passes, we can merge. I think we don't need exhaustive checks.

tests/reference/asr-continue_compilation_2-a6145a1.stderr
HarshilShah1804: Yes, I also felt so. I'll try to fix that afterwards.

HarshilShah1804:
The checks likely failed due to a network issue while cloning stdlib. Such network errors have become very frequent on the macOS runners over the past few days.

PR #8663, opened 2025-10-04
Remove deprecated functions

certik:

PR #8664, opened 2025-10-04
Remove ptr_type_deprecated

certik:
Builds on top of https://github.com/lfortran/lfortran/pull/8663.


certik:
I think this is a subset of https://github.com/lfortran/lfortran/pull/8665.

PR #8667, opened 2025-10-04
Remove context.setOpaquePointers(false);

certik:


certik:
For LLVM 15 and 16 we get:
```
lfortran: /home/runner/micromamba/envs/lf/include/llvm/IR/Type.h:418: llvm::Type* llvm::Type::getNonOpaquePointerElementType() const: Assertion `NumContainedTys && "Attempting to get element type of opaque pointer"' failed.
```
So likely we are assuming the pointers are typed. We need to change the code that calls it to assume opaque pointers, then it should work.

assem2002:
I think all block of codes that handle opaque pointer logic ( Using map + Calling some different API) have this `if` macro
```cpp
#if LLVM_VERSION_MAJOR > 16
```

it needs to get changed to 
```cpp
#if LLVM_VERSION_MAJOR >= 15
```

certik:
All the map was removed now. But I think the point still applies, the API is slightly different.

certik:
I tried the fix in b3690702fdb84c874787bd17723c44237eeeb8a3, so we'll see if it fixes it.

certik:
The commit didn't work, so more investigation is needed.

certik:
@assem2002 thanks for fixing it! Removing `context.setOpaquePointers(false);` is good, since LLVM said it is not supported super well, only "best effort". So now we use the default, well-supported path in all LLVM versions.

PR #8671, opened 2025-10-05
Add LLVM 7 support

certik:
Builds on top of #8670. The code builds and a simple test runs, with a linker warning. More complicated tests do not run.


certik:
This PR is identical but also tests LLVM 7 at the CI: https://github.com/lfortran/lfortran/pull/8673

PR #8673, opened 2025-10-05
Support LLVM 7

certik:
Some tests pass, but not all.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -491,7 +491,7 @@ jobs:
 
       - name: Test third party codes
         shell: bash -e -x -l {0}
-        if: ${{! (contains(matrix.llvm-version, '7')) }}
+        if: ${{! (matrix.llvm-version == '7') }}
         run: |
             export PATH="$(pwd)/src/bin:$PATH"
             RUNNER_OS="${{matrix.os}}" FC=lfortran ci/test_third_party_codes.sh


src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -9,6 +9,25 @@ namespace LCompilers {
...
+            auto *GV = new llvm::GlobalVariable(
+                module, StrConstant->getType(), true,
+                llvm::GlobalValue::PrivateLinkage, StrConstant, "",  nullptr,
+                llvm::GlobalVariable::NotThreadLocal, 0);
+            GV->setUnnamedAddr(llvm::GlobalValue::UnnamedAddr::Global);
+            return llvm::ConstantExpr::getBitCast(GV, llvm::Type::getInt8PtrTy(context, 0));
+#else
+            return builder.CreateGlobalStringPtr(Str);
+#endif
+        }
```
certik, line 29: I currently don't understand what the issue is with `builder.CreateGlobalStringPtr` in LLVM 7. The above creates a workaround that does seem to work. But this will require to use `CreateGlobalStringPtrSafe` everywhere instead of `builder.CreateGlobalStringPtr`, which is fine, we could do that. But before we do, it would be good to understand exactly the problem and if the above is the best way to fix it. It inserts an extra bitcast in LLVM 7, so it might not be the optimal solution.

src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -19,7 +38,7 @@ namespace LCompilers {
                             llvm::Type::getInt64Ty(context)
                         }, false);
                 fn = llvm::Function::Create(function_type,
-                        llvm::Function::ExternalLinkage, func_name, module);
+                        llvm::Function::ExternalLinkage, func_name, &module);
```
certik, line 41: These changes are probably fine. Alternatively we can create a wrapper and use ifdef inside, but probably not needed in this case.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1536,11 +1536,34 @@ namespace LCompilers {
 
 
     llvm::Value *LLVMUtils::CreateLoad2(llvm::Type *t, llvm::Value *x, bool is_volatile) {
+#if LLVM_VERSION_MAJOR >= 8
         return builder->CreateLoad(t, x, is_volatile);
+#else
+        // LLVM 7: CreateLoad(Value*, bool, Name) signature
+        return builder->CreateLoad(x, is_volatile);
+#endif
```
certik, line 1544: I think this change is related to our recent opaque pointer refactor. In old LLVM versions one didn't specify the type for load. The correct API for all LLVM version is to specify the type, so our wrapper is good and inside it we need to use this ifdef.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1536,11 +1536,34 @@ namespace LCompilers {
...
+                }
+            }
+            if (all_constant) {
+                // Create GEP instruction directly, bypassing constant folder
+                return llvm::GetElementPtrInst::Create(t, x, idx, "", builder->GetInsertBlock());
+            }
+        }
+#endif
         return builder->CreateGEP(t, x, idx);
     }
```
certik, line 1568: I am not sure if this workaround is the right one. There is some issue with `builder->CreateGEP` in LLVM 7 and I am also not completely sure if the above workaround actually fixes the problem, but I think it does. We need to understand exactly what is happening.

src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -9,6 +9,25 @@ namespace LCompilers {
 
     namespace LLVMArrUtils {
 
+        // Workaround for LLVM 7 CreateGlobalStringPtr bug
+        static llvm::Constant* CreateGlobalStringPtrSafe(
```
certik, line 13: I am not sure with the name `Safe`, maybe a different name would be better.

src/libasr/codegen/llvm_utils.h
```diff
@@ -284,7 +284,18 @@ namespace LCompilers {
...
+                        llvm::Constant *StrConstant = llvm::ConstantDataArray::getString(context, message);
+                        auto *GV = new llvm::GlobalVariable(*module, StrConstant->getType(), true,
+                            llvm::GlobalValue::PrivateLinkage, StrConstant, "",
+                            nullptr, llvm::GlobalVariable::NotThreadLocal, 0);
+                        GV->setUnnamedAddr(llvm::GlobalValue::UnnamedAddr::Global);
+                        llvm::Value* formatted_msg = llvm::ConstantExpr::getBitCast(GV,
+                            llvm::Type::getInt8PtrTy(context, 0));
+#else
                         llvm::Value* formatted_msg = builder->CreateGlobalStringPtr(message);
+#endif
```
certik, line 298: We should use `CreateGlobalStringPtrSafe` here instead, which does the same thing.

certik:
@Pranavchiku, @assem2002, @jinangshah21 let me know what you think. I think most of these changes are to figure out a "stable" API that works for all LLVM versions, and the extra ifdef, hidden inside our own API wrapper, should not slow anything down for modern LLVM versions.

Overtime I would like to move to have our own API on top of LLVM that is stable and works with any LLVM version, so these changes are towards such an API.

There are a few places where we need to investigate more exactly what is going on, I commented there.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -908,7 +933,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         character_type->getPointerTo()
                     }, false);
             fn = llvm::Function::Create(function_type,
-                    llvm::Function::ExternalLinkage, runtime_func_name, *module);
+                    llvm::Function::ExternalLinkage, runtime_func_name, module.get());
```
Pranavchiku, line 936: are these necessary?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2348,19 +2373,43 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     void generate_Exp(ASR::expr_t* m_arg) {
         this->visit_expr_wrapper(m_arg, true);
         llvm::Value *item = tmp;
+#if LLVM_VERSION_MAJOR >= 12
         tmp = builder->CreateUnaryIntrinsic(llvm::Intrinsic::exp, item);
+#elif LLVM_VERSION_MAJOR >= 8
+        tmp = builder->CreateIntrinsic(llvm::Intrinsic::exp, {item->getType()}, {item});
```
Pranavchiku, line 2379: I think we can club these do, used to work before as is.

src/libasr/codegen/llvm_array_utils.cpp
Pranavchiku: If it works in LLVM 7 without this change, we should avoid it.

src/libasr/codegen/llvm_array_utils.cpp
assem2002: I'll check that. Chatgpt agrees that llvm < 8 versions have some problems related to GEP with strings. I'll see if we can improve that.

jinangshah21: Inside `builder.CreateGlobalStringPtr` we have `ConstantExpr::getGetElementPtr` call which is Buggy in LLVM 7. Below diff show difference inside `ConstantExpr::getGetElementPtr` function for LLVM 8 from 7.

```diff
-    assert(
-       Ty ==
-        cast<PointerType>(C->getType()->getScalarType())->getContainedType(0u));
+    assert(Ty ==
+           cast<PointerType>(C->getType()->getScalarType())->getElementType());
```
Here `getElementType()` do some additional checks in LLVM 8 which was not handled in LLVM 7.


I think above change is fine as one bitcast is added but `CreateGlobalStringPtr` contains a  `ConstantExpr::getGetElementPtr` which is not here.

src/libasr/codegen/llvm_array_utils.cpp
certik: Here is what I get in LLVM 7 if I remove this change:
```
/workspace/repos/lfortran/src/libasr/codegen/llvm_array_utils.cpp:41:69: error: no viable conversion from 'llvm::Module' to 'Module *'
   41 |                         llvm::Function::ExternalLinkage, func_name, module);
      |                                                                     ^~~~~~
/workspace/repos/lfortran/.pixi/envs/llvm7/include/llvm/IR/Function.h:137:56: note: passing argument to parameter 'M' here
  137 |                           const Twine &N = "", Module *M = nullptr) {
      |                                                        ^
```
I am not quite sure if in modern LLVM versions they support both a pointer and a value for Module.

assem2002: https://llvm.org/doxygen/classllvm_1_1Function.html#ab7351d0170b356c9f3487f8cc8c6cafd

The function is overloaded (both `*` and `&`) probably in the newer versions only.

src/libasr/codegen/llvm_array_utils.cpp
assem2002: apparently integration tests fail with llvm 7 and this PR just compiles with llvm7. The failures I encountered are much related to loading with various kind of pointers (not just for strings). I don't see it's necessary to use that wrapper, LFortran compiles with or witihout the wrapper. I also think the wrapper doesn't cost anything (runtime wise) so it's also fine to keep it.

src/libasr/codegen/evaluator.cpp
```diff
@@ -326,6 +337,25 @@ void LLVMEvaluator::add_module(std::unique_ptr<LLVMModule> m) {
...
+        llvm::SmallVector<char, 128> buf;
+        llvm::raw_svector_ostream dest(buf);
+        llvm::logAllUnhandledErrors(std::move(e), dest, "");
+        std::string msg = std::string(dest.str().data(), dest.str().size());
+        if (msg[msg.size()-1] == '\n') msg = msg.substr(0, msg.size()-1);
+        throw LCompilersException("getAddress() failed for symbol '"
+            + name + "', error: " + msg);
+    }
+    return (intptr_t)addr.get();
+#else
```
assem2002, line 358: This is fine. but could be merged with the code block below it.

src/libasr/codegen/llvm_utils.cpp
assem2002: It's also fine to use this, although we don't test lllvm7. 
What I mean is that maybe llvm7 works just fine with no problem.

src/libasr/codegen/llvm_array_utils.cpp
certik: I'll try to create a minimal PR that compiles and we test it at the CI to ensure we can at least compile. Then we can worry about enabling some of these workaround and adding tests.

certik:
Thanks for the review. I'll try to make a minimal PR that still builds and we can merge that. That doesn't cost us much, and then over time we can get tests working in separate PRs.

Understanding what is involved should also help us with designing a good stable API for LLVM:

https://github.com/lfortran/lfortran/issues/8674

certik:
Here is the minimal PR just to compile: https://github.com/lfortran/lfortran/pull/8676.

I made this PR a draft, so that we don't merge it accidentally. All tests pass, so the changes are good, but let's merge things in step by step, ensuring high quality.

PR #8676, opened 2025-10-06
Get LFortran to compile with LLVM 7

certik:
Tests do not run, but everything compiles and links. CI test for compilation added.


certik:
The failure is unrelated. This PR adds about 5 minutes to the CI with the extra job for LLVM 7, which I think is fine for now. The changes are minimal to get things compiling with LLVM 7, so I think we can merge it.

Now we can focus on investigating proper fixes to run things correctly.

Issue #8674, opened 2025-10-05
Creating stable LLVM API


certik:
I iterated with Grok about how we can create a stable LLVM API that we can use in LFortran, and then provide implementation of it using LLVM 7 - 21 (and later even lower LLVM versions). This will fully isolate LFortran from any particular LLVM dependency. Later we can also provide our own fast Debug implementation.

This refactor we should do after beta, it's not needed right now and it will take some time, although it can be done gradually.

Relevant issues:

* https://github.com/lfortran/lfortran/issues/1905
* https://github.com/lfortran/lfortran/issues/2587
* https://github.com/lfortran/lfortran/issues/1302

---------

# Stable LLVM API Implementation Plan

## Overview

This document outlines a refined plan for creating a stable, version-agnostic C-style API for LLVM (versions 7–21) in LFortran. The API uses opaque struct pointers (e.g., `struct LlvmModule;`) for type safety and seamless migration. This design allows hybrid usage during transition: files can mix stable API calls with direct LLVM calls via temporary bridge functions (e.g., `llvm_unwrap_module`). Once migrated, upstream LLVM headers are removed.

**Goals**:
- Abstract LLVM version differences (e.g., pointer vs. reference overloads, bugs like `CreateGlobalStringPtr`).
- Enable incremental migration of LFortran's large codebase without big-bang rewrites.
- Support a future C++ RAII layer with near-zero overhead.
- Maintain pure C compatibility for portability.

**Key Design Choices**:
- **Opaque Pointers**: Forward-declare `struct LlvmModule;`; implementation aliases it to `llvm::Module` (e.g., `typedef struct LlvmModule llvm::Module;`).
- **Error Handling**: Simple integer returns (0 = success).
- **Ownership**: RAII-like via create/destroy pairs; pointers are cheap.
- **Version Dispatch**: Use `#if LLVM_VERSION_MAJOR >= X` in the .cpp impl.

### Core Principles
- **Minimal Header**: Only forward-decls and function signatures; no LLVM exposure.
- **Typed Handles**: Pointers to opaque structs (e.g., `struct LlvmModule*`) for safety over `void*`.
- **Functions**: Procedural C style with `llvm_` prefix; varargs for flexible ops (e.g., function params).
- **Bridges**: Temporary C++-only unwrap functions for migration (enabled via CMake flag).
- **Extensibility**: Add categories (e.g., Builder ops, Types) iteratively.

### Detailed Header Example (`llvm_api.h`)

```c
[...]

PR #8678, opened 2025-10-06
Migrate `builder->CreateGlobalStringPtr()` to `LCompilers::create_global_string_ptr()`

certik:
This PR gets a simple example working with LLVM 7:
```
$ ./src/bin/lfortran examples/expr2.f90 
/usr/bin/ld: /tmp/expr2.tmp_Ve8JmYiNC9T2eMOTkqDLdI.o: warning: relocation in read-only section `.text'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
25
```

### DIFF
--- a/src/libasr/codegen/llvm_utils.h
+++ b/src/libasr/codegen/llvm_utils.h
@@ -133,7 +133,8 @@ namespace LCompilers {
 
     //! Creates a global string pointer safely across LLVM versions
     static inline llvm::Constant* create_global_string_ptr(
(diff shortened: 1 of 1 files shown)


certik:
This doesn't change anything for LLVM > 7, and gets a simple program working for LLVM 7, with linker warnings. So I think we can merge it. Now we can investigate what the proper fix is for LLVM 7, at least the global string creation is done at one place, and even if we don't need the current LLVM 7 workaround, this PR moves us closer towards #8674.

PR #8679, opened 2025-10-06
Fix: returning a string from a function without declaring returned var

### LINKED ISSUE #8371: character function doesn't default to len=1.
```
module elem_char
  implicit none

contains

character function tolower (c) result (res)
  character, intent(in) :: c

  res = merge (achar (iachar (c) + 32), c, c >= 'A' .and. c <= 'Z')

end function

end module
```
When compiled with lfortran, returns:
```
$ lfortran -c elem_char.f90
ASR verify pass error: String length must be length >= 0
Current length is -> -10
  --> elem_char.f90:6:1 - 11:12
   |
 6 |    character function tolower (c) result (res)
   |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
11 |    end function
   | ...^^^^^^^^^^^^ failed here


$
```
Since it mentions "Current length is ->-10", I'm assuming something is not initialized correctly.  A workaround is to declare the function with ```character(1)```.

Version of lfortran is a snapshot of the trunk from a few days ago:
```
$ lfortran --version
LFortran version: 0.54.0-467-g09add98dd
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$
```

akramhany:
Fixes #8371.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -212,7 +212,7 @@ jobs:
         # exactly the same manner, to ensure that users can just change the
         # `llvmdev` version in their conda environment and everything will just
         # work.
-        llvm-version: ["10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21"]
+        llvm-version: ["7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21"]
         python-version: ["3.10"]
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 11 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1545,7 +1545,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             ASR::ttype_t *type = nullptr;
             int i_kind = compiler_options.po.default_integer_kind;
             int a_kind = 4;
-            int a_len = -10;
+            int a_len = 1;
```
certik, line 1548: Can you examine the git history to find a PR that implemented this?

Let's investigate what the original motivation for `a_len=-10` was. Maybe it was meant as "not initialized", and the bug is somewhere else. Or it was really just a typo and it was meant to be 1, in which case this change is good.

certik:

src/lfortran/semantics/ast_symboltable_visitor.cpp
assem2002: I believe it was some default value that indicates something is wrong about `len` (before : it was either positive integer OR `-3` indicating runtime value OR `-1` indicating a `*` OR `-2` indicating a `:`)

refer to this commit https://github.com/lfortran/lfortran/commit/fd71d22fba070076efa4f04e7cffb38c8a1b9ceb

assem2002:
I believe we'd still have issues with example like this 
```.f90
program pp
  integer :: n
  contains

  character(n) function tolower (c) result (res)
    character, intent(in) :: c
  end function

end program  pp
```

We need to tackle this. I'd also like to not use `extract_len` (we use it in 2 spots), It follows the old length design mentioned here https://github.com/lfortran/lfortran/commit/fd71d22fba070076efa4f04e7cffb38c8a1b9ceb#diff-1599d8226e12e870f72451766a11638199a4cdde172ca0ef9c057912072a5bcfR203

Let's fix that in subsequent PR.

certik:
@assem2002 thanks. I merged it. Can you please create issues and describe what else needs to be now fixed in subsequent PRs?

assem2002:
Opened issue here https://github.com/lfortran/lfortran/issues/8686.
 @akramhany It's quite relevant to this PR, Maybe you can give that a look.

PR #8682, opened 2025-10-06
feat: [new_classes] implement array and scalar struct deepcopy

jinangshah21:
Towards #8630

PR #8690, opened 2025-10-08
Add type asserts to create_gep

certik:
This ensures that types are consistent with the pointer type. These kinds of checks are only needed for LLVM < 15.

Without these asserts LLVM will segfault deep inside it if we call it with incorrect types. So these asserts catch bugs early.

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -3,6 +3,7 @@
 #include <libasr/codegen/llvm_array_utils.h>
 #include <libasr/asr_utils.h>
 #include <libasr/codegen/llvm_compat.h>
+#include <llvm/Support/raw_ostream.h>
 
 namespace LCompilers {
 


certik:
@jinangshah21 here is the motivation for this PR: https://github.com/lfortran/lfortran/pull/8682#issuecomment-3379106345. Let me know if the checks in this PR are correct. They pass all tests in main for me locally, I tested LLVM 8, 11 and 21.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1463,6 +1471,48 @@ namespace LCompilers {
...
+        if (ds->getType()->isPointerTy()) {
+            llvm::Type* ds_pointee_type = ds->getType()->getPointerElementType();
+            // For struct types, the target type must match the pointer's pointee type
+            if (llvm::isa<llvm::StructType>(t)) {
+                std::string target_type_str;
+                llvm::raw_string_ostream target_rso(target_type_str);
+                t->print(target_rso);
+                std::string pointee_type_str;
+                llvm::raw_string_ostream pointee_rso(pointee_type_str);
+                ds_pointee_type->print(pointee_rso);
```
certik, line 1488: In future PRs this should be refactored to only do this work if the condition `ds_pointee_type == t` fails, to not slow down compilation.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1463,6 +1471,48 @@ namespace LCompilers {
     }
 
     llvm::Value* LLVMUtils::create_gep2(llvm::Type *t, llvm::Value* ds, int idx) {
+#if LLVM_VERSION_MAJOR < 15
```
assem2002, line 1474:
```suggestion
#if defined(WITH_LFORTRAN_ASSERT) &&  LLVM_VERSION_MAJOR < 15
```
It's fine to check all the times but I think the checks introduced should be only for debugging, so it's better to call that check off on release mode just for speed nothing less.

certik: Perfect, that's the way to do it. I'll apply this in a subsequent PR, so that we can merge this in and start using it. I need to check with the above addition that the check still triggers, which I'll do in the next PR.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -116,6 +117,13 @@ namespace LCompilers {
                     ASR::is_a<ASR::Allocatable_t>(asr_type)) &&
                     !ASRUtils::is_string_only(const_cast<ASR::ttype_t*>(&asr_type));
         }
+
+        std::string get_type_as_string(llvm::Type* type) {
+            std::string type_str;
+            llvm::raw_string_ostream rso(type_str);
+            type->print(rso);
+            return rso.str();
+        }
```
assem2002, line 126: I think that's not used.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1463,6 +1471,48 @@ namespace LCompilers {
...
     llvm::Value* LLVMUtils::create_gep2(llvm::Type *t, llvm::Value* ds, int idx) {
+#if LLVM_VERSION_MAJOR < 15
+        // Assertion: Verify type consistency to catch bugs early
+        // The GEP target type 't' should match the pointee type of 'ds'
+        // Note: LLVM 15+ uses opaque pointers, so getPointerElementType() doesn't exist
+        // and this type of bug cannot occur.
+        if (ds->getType()->isPointerTy()) {
+            llvm::Type* ds_pointee_type = ds->getType()->getPointerElementType();
+            // For struct types, the target type must match the pointer's pointee type
+            if (llvm::isa<llvm::StructType>(t)) {
```
assem2002, line 1482: I think we don't need to check only for `structType`, Right ?
The check `ds_pointee_type == t` should be for any type.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1463,6 +1471,48 @@ namespace LCompilers {
     }
 
     llvm::Value* LLVMUtils::create_gep2(llvm::Type *t, llvm::Value* ds, int idx) {
+#if LLVM_VERSION_MAJOR < 15
+        // Assertion: Verify type consistency to catch bugs early
+        // The GEP target type 't' should match the pointee type of 'ds'
+        // Note: LLVM 15+ uses opaque pointers, so getPointerElementType() doesn't exist
+        // and this type of bug cannot occur.
+        if (ds->getType()->isPointerTy()) {
```
assem2002, line 1479:
```suggestion
        lcompilers_assert(ds->getType()->isPointerTy())
```

I believe `ds` muust be pointer at all times. I LLVM raises error if not.

assem2002: and we remove the` if `block

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1463,6 +1471,48 @@ namespace LCompilers {
...
+        // Only validate bounds for struct types
+        if (llvm::isa<llvm::StructType>(t)) {
+            llvm::StructType* struct_type = llvm::cast<llvm::StructType>(t);
+            unsigned num_elements = struct_type->getNumElements();
+            LCOMPILERS_ASSERT_MSG(idx >= 0 && (unsigned)idx < num_elements,
+                "Index out of bounds in create_gep2: index " + std::to_string(idx) +
+                " is out of range for struct with " + std::to_string(num_elements) + " elements");
+        }
+#endif
+
```
assem2002, line 1515: Looks good

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1545,11 +1595,35 @@ namespace LCompilers {
 
     llvm::Value* LLVMUtils::CreateGEP2(llvm::Type *t, llvm::Value *x,
             std::vector<llvm::Value *> &idx) {
+#if LLVM_VERSION_MAJOR < 15
+        // Validate that the type parameter matches the pointer's pointee type
+        // Note: LLVM 15+ uses opaque pointers, so this check is not possible/needed
+        if (x->getType()->isPointerTy()) {
```
assem2002, line 1601:
```suggestion
        lcompilers_assert(x->getType()->isPointerTy())
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1558,6 +1632,23 @@ namespace LCompilers {
 
     llvm::Value* LLVMUtils::CreateInBoundsGEP2(llvm::Type *t,
             llvm::Value *x, std::vector<llvm::Value *> &idx) {
+#if LLVM_VERSION_MAJOR < 15
+        // Validate that the type parameter matches the pointer's pointee type
+        // Note: LLVM 15+ uses opaque pointers, so this check is not possible/needed
+        if (x->getType()->isPointerTy()) {
```
assem2002, line 1638:
```suggestion
        lcompilers_assert(x->getType()->isPointerTy())
```

certik:
Great feedback, thanks @assem2002. I'll address it in subsequent PRs.

src/libasr/codegen/llvm_utils.cpp
certik: Fixed in https://github.com/lfortran/lfortran/pull/8695.

certik:
All the suggestions are addressed in https://github.com/lfortran/lfortran/pull/8696.

PR #8695, opened 2025-10-08
Only run checks if WITH_LFORTRAN_ASSERT is defined

certik:


src/libasr/codegen/llvm_utils.cpp
certik, line 1504: I think this is incorrect.

certik: I fixed this in https://github.com/lfortran/lfortran/pull/8696.

PR #8696, opened 2025-10-08
Simplify the assert checks in create_gep

certik:
Depends on https://github.com/lfortran/lfortran/pull/8695


certik:
We can just merge this one.

PR #8700, opened 2025-10-09
fix: [new_classes] pass polymorphic array correctly to function

jinangshah21:


certik:
Tests fail, but the change is fine.

PR #8703, opened 2025-10-10
fix: fix dimension handling in case of assumed rank arrays

HarshilShah1804:
This enables more features of assumed rank arrays like element-accessing.

### DIFF
--- a/integration_tests/select_rank_01.f90
+++ b/integration_tests/select_rank_01.f90
@@ -15,15 +15,25 @@ contains
     subroutine check(x, expected)
         integer, intent(in) :: x(..)
         integer, intent(in) :: expected
+        integer :: i
 
         select rank(x)
             rank(0)
                 if (expected /= 0) error stop
             rank(1)
+                i = x(2)
+                print *, i
+                if (i /= 2) error stop
                 if (expected /= 1) error stop
             rank(2)
+                i = x(1,2)
+                print *, i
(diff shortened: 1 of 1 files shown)


src/libasr/asr_verify.cpp
```diff
@@ -938,7 +938,7 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
                 // TODO: This seems like a bug, we should not use ArrayItem with
                 // strings but StringItem. For now we ignore it, but we should
                 // fix it
-            } else {
+            } else if (!ASRUtils::is_assumed_rank_array(ASRUtils::expr_type(x.m_v))) {
                 require(n_dims > 0,
                     "The variable in ArrayItem must be an array, not a scalar");
             }
```
certik, line 944: I think this is not correct. By the time it arrives to ArrayItem, it must have `n_dims` > 0. It can't be "assumed rank array". It must be properly cast to a particular DescriptorArray with an appropriate `n_dims`.

certik:
Keep this PR up, as a backup if we wanted to quickly fix this as a workaround. But this is not the right design, so I think we should not be weakening the verify. The right design is that we need to insert a cast from assumed rank to a DescriptorArray of a particular dimension, known at compile time.

HarshilShah1804:
Now as we have used the casting mechanism, I am closing this PR.

PR #8707, opened 2025-10-11
fix : Check for Class Assignment compatibility for unlimited polymorphic type by comparing symbol names

### LINKED ISSUE #8702: Subroutine call to a interface involving `unlimited polymorphic argument` causes Semantic Error
MRE :- 
```.f90
module stdlib_hashmap_wrappers
    implicit none

    type :: other_type
    end type other_type

    interface set
        module procedure set_other
    end interface set

contains

    subroutine set_other( other, value )
        type(other_type), intent(out) :: other
        class(*), intent(in)          :: value
        allocate(other % value, source=value)
    end subroutine set_other

end module stdlib_hashmap_wrappers


program test_open_maps

    use stdlib_hashmap_wrappers
    implicit none
    
contains

    subroutine input_random_data( )
        class(*), allocatable :: dummy
        type(other_type) :: other
        call set ( other, dummy )
    end subroutine input_random_data

end program test_open_maps
```

LFortran Error :- 
```console
(lf) lordansh@LordAnsh:~/dev/lfort1/lfortran$ lfortran a.f90
semantic error: Arguments do not match for any generic procedure, set
  --> a.f90:33:9
   |
33 |         call set ( other, dummy )
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

MRE works well with GFortran.

AnshMehta1:
Fixes #8702

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1656,6 +1656,7 @@ RUN(NAME derived_types_79 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_AR
 RUN(NAME derived_types_80 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME derived_types_81 LABELS gfortran llvm)
 RUN(NAME derived_types_82 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays --new-classes)
+RUN(NAME derived_types_83 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME derived_type_with_default_init LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME derived_type_with_default_init_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
--- a/integration_tests/derived_types_83.f90
+++ b/integration_tests/derived_types_83.f90
@@ -1,0 +1,36 @@
+module wrappers_derived_types_83
+    implicit none
+
+    type :: other_type
+        class(*), allocatable :: value
+    end type other_type
+
+    interface set
+        module procedure set_other
+    end interface set
+
+contains
+
+    subroutine set_other( other, value )
+        type(other_type), intent(out) :: other
+        class(*), intent(in)          :: value
+        allocate(other % value, source=value)
+    end subroutine set_other
+
+end module wrappers_derived_types_83
+
+
+program derived_types_83
+
+    use wrappers_derived_types_83
+    implicit none
+    
+contains
+
+    subroutine input_random_data( )
+        class(*), allocatable :: dummy
+        type(other_type) :: other
+        call set ( other, dummy )
+    end subroutine input_random_data
+
+end program


src/libasr/asr_utils.h
```diff
@@ -4379,9 +4379,11 @@ inline bool check_class_assignment_compatibility(ASR::symbol_t* target, ASR::sym
     value = ASRUtils::symbol_get_past_external(value);
     bool is_class_same = false;
     if (ASR::is_a<ASR::Struct_t>(*target) && ASR::is_a<ASR::Struct_t>(*value)) {
+        std::string target_name = std::string(ASRUtils::symbol_name(target));
+        std::string value_name = std::string(ASRUtils::symbol_name(value));
         ASR::Struct_t* tar_struct = ASR::down_cast<ASR::Struct_t>(target);
         ASR::Struct_t* val_struct = ASR::down_cast<ASR::Struct_t>(value);
-        is_class_same = (target == value);
+        is_class_same = (target_name == value_name);
```
certik: I don't like that we are comparing strings here, it seems wrong. For example what if there are two classes of the same name defined in two modules, that happen to have the same name but are not the same?

Is there another way without comparing strings?

certik:
I am going to make it a draft. I think the type comparison should work based on the symbol itself, not the name. Can you explain exactly what the problem is? Let's figure out a better solution than comparing strings.

AnshMehta1:
I am not sure about the exact error but I think, its related to the name `~unlimited_polymorphic_type`.

Like I followed the diff
```diff
diff --git a/src/libasr/asr_utils.h b/src/libasr/asr_utils.h
index 363d8b24d..9d33fec7d 100644
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -4377,6 +4377,9 @@ inline bool check_class_assignment_compatibility(ASR::expr_t* target, ASR::expr_
 inline bool check_class_assignment_compatibility(ASR::symbol_t* target, ASR::symbol_t* value) {
     target = ASRUtils::symbol_get_past_external(target);
     value = ASRUtils::symbol_get_past_external(value);
+    std::cout<<ASRUtils::symbol_name(target)<<std::endl;
+    std::cout<<ASRUtils::symbol_name(value)<<std::endl;
+    std::cout<<(ASRUtils::symbol_name(target) == ASRUtils::symbol_name(value))<<std::endl;
     bool is_class_same = false;
     if (ASR::is_a<ASR::Struct_t>(*target) && ASR::is_a<ASR::Struct_t>(*value)) {
         ASR::Struct_t* tar_struct = ASR::down_cast<ASR::Struct_t>(target);
```

for `class(*)`, this are the results :-
```console
~unlimited_polymorphic_type
~unlimited_polymorphic_type
0
```

But for any regular class like `class(key_type)`, results are fine as :- 
```console
key_type
key_type
1
```

Due to this, `(target == value)` condition is not evaluating to true for `class(*)`, whereas it should be.
This is not occurring when I compare strings rather than `char *`.

AnshMehta1:
I think the problem of `char *` comparison as I described at https://github.com/lfortran/lfortran/pull/8707#issuecomment-3394544703 was limited to `~unlimited_polymorphic_type`. So, instead of comparing every struct using m_name, we should do this only for `~unlimited_polymorphic_type`. We already have a utility function `is_derived_type_similar` for this so I directly used it and updated the PR.

src/libasr/asr_utils.h
```diff
@@ -4381,8 +4381,7 @@ inline bool check_class_assignment_compatibility(ASR::symbol_t* target, ASR::sym
     if (ASR::is_a<ASR::Struct_t>(*target) && ASR::is_a<ASR::Struct_t>(*value)) {
         ASR::Struct_t* tar_struct = ASR::down_cast<ASR::Struct_t>(target);
         ASR::Struct_t* val_struct = ASR::down_cast<ASR::Struct_t>(value);
-        is_class_same = (target == value);
-        is_class_same = is_class_same || ASRUtils::is_parent(tar_struct, val_struct);
+        is_class_same = is_derived_type_similar(tar_struct, val_struct);
     }
```
jinangshah21, line 4386: We can't use `is_derived_type_similar` as it considers both side inheritance as true but we need only one side which was correct previousy. For above example, both target and value are `~unlimited_polymorphic_type` struct but both struct are in different scope. So issue is due to seperate symbols of `unlimited_polymorphic_type` struct in different scope.

certik:
I think this is an issue: https://github.com/lfortran/lfortran/pull/8707#discussion_r2429880666.

certik:
@jinangshah21 how should this be fixed?

AnshMehta1:
> @jinangshah21 how should this be fixed?

Based on @jinangshah21 comment. I think we just need to add an extra condition (in or) in our previous code to check for ~unlimited_polymorphic_type as we added in is_derived_type_similar. Am I right??

jinangshah21:
Solution in https://github.com/lfortran/lfortran/pull/8707#issuecomment-3403772310 is fine to get code work, but optimal fix should be to have a global module for storing `~unlimited_polymorphic_type` struct and use this single symbol everywhere in code.

AnshMehta1:
> Solution in [#8707 (comment)](https://github.com/lfortran/lfortran/pull/8707#issuecomment-3403772310) is fine to get code work, but optimal fix should be to have a global module for storing `~unlimited_polymorphic_type` struct and use this single symbol everywhere in code.

Currently, I am not having the exact knowledge for that global module approach. So, I am just implementing the approach I mentioned at https://github.com/lfortran/lfortran/pull/8707#issuecomment-3403772310 which is used at many places in the codebase. Global module thing can be implemented later on for which I have opened an issue #8738 .

PR #8708, opened 2025-10-12
fix: [new_classes] handle `class(*)` argument conversion and assignments

jinangshah21:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -273,6 +273,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         if (ASR::is_a<ASR::Variable_t>(*sym) && (ASRUtils::symbol_name(sym) == current_selector_var_name)) {        \
             llvm::Type *llvm_type = llvm_utils->get_type_from_ttype_t_util(expr,                                    \
                 ASRUtils::extract_type(current_select_type_block_type_asr), module.get());                          \
+            if (compiler_options.new_classes) {                                                                     \
+                llvm::Type *generic_llvm_type = llvm_utils->get_type_from_ttype_t_util(expr,                        \
+                    ASRUtils::extract_type(ASRUtils::expr_type(expr)), module.get());                               \
+                llvm_value = llvm_utils->create_gep2(generic_llvm_type, llvm_value, 1);                             \
+                llvm_value = llvm_utils->CreateLoad2(llvm_utils->i8_ptr, llvm_value);                               \
+                llvm_value = builder->CreateBitCast(llvm_value, llvm_type->getPointerTo());                         \
+            }                                                                                                       \
```
certik, line 282: We should not be using macros like this. Not related to this PR, you are just following what was there.

PR #8710, opened 2025-10-14
fix: throw error for do loop name already used for another variable

### LINKED ISSUE #8709: Bug if loop name = name of integer variable
This program won't compile with gfortran, ifx or AMD flang because ```i``` is both an integer variable and a loop name but lfortran runs it. I think the other 3 compilers are right.
```
program label_do
  integer i
  i: do i = 1,3
     print *,i
  end do i
end program label_do
```
Output from lfortran 0.57.0
```
1
2
3
```

HarshilShah1804:
FIxes #8709

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -345,4 +345,8 @@ program continue_compilation_1
 
     !Data Statements with different number of arguments on LHS and RHS
     data j2, x2, (y2(i2), i2=1,3), k2 / 1,2,3,4,5,6,7,3*8 /
+
+    q1: do q1 = 1, 3
+        print *, q1
+    end do q1
 end program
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4936,6 +4936,18 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
 
     void visit_DoLoop(const AST::DoLoop_t &x) {
+        char* loop_name = x.m_stmt_name;
+        if (loop_name) {
+            std::string loop_name_str = to_lower(loop_name);
+            if (current_scope->get_symbol(loop_name_str) != nullptr) {
+                diag.add(Diagnostic(
+                    "DO loop label '" + loop_name_str + "' already defined",
+                    Level::Error, Stage::Semantic, {
+                        Label("",{x.base.base.loc})
```
certik, line 4946: Don't use `x`, but highlight the actual label that is already defined.

As a bonus, highlight both the label and the definition. You can do it in a subsequent PR.

Issue #8712, opened 2025-10-14
`function_call_in_declaration` creates a helper function with mismatching funcParams


assem2002:
```.f90
module mm
  contains
  function foo (x) result(res)
    character(*) :: x
    character(10) :: res
  end function 

  pure function boo(x,y) result(res)
    integer, intent(in) :: x,y
    integer :: res
  end function

function ff(line,length) result(strout)  
  character(len=*):: line
  integer :: length
  character(len=boo(length, len(foo(line)))) :: strout
end function 

end module 


program pp
  use mm
end program 
```

**error**
```
LCOMPILERS_ASSERT failed: /home/assem/LFortranUpStream/lfortran/src/libasr/../libasr/asr_utils.h
function Call_t_body(), line number 6728 at 
ASRUtils::check_equal_type(arg_type, orig_arg_type, arg_expr, orig_arg_expr)
ERROR MESSAGE:
ASRUtils::check_equal_type(i32, str)

```

Issue #8713, opened 2025-10-14
Arguments of `IntrinsicElementalFunctions` aren't properly visited in `function_call_in_declaration` pass


assem2002:
```.f90
module mm
  contains
  function foo (x) result(res)
    character(*) :: x
    character(10) :: res
  end function 

function ff(line,length) result(strout)  
  character(len=*):: line
  integer :: length
  character(len=max(length, len(foo(line)))) :: strout
end function 

end module 


program pp
  use mm
end program 
```

```
AssertFailed: parent_body != nullptr
```

call to `foo()` should trigger the `function_call_in_declaration` pass and force it to replace `len(foo(line))` with a call to some helper function.

Pranavchiku:
The provided program works as is:

```console
% lfortran a.f90            
% echo $?
0
```

Pranavchiku:
The following fails:


```fortran
module mm
  contains
  function foo (x) result(res)
    character(*) :: x
    character(10) :: res
  end function 

function ff(line,length) result(strout)  
  character(len=*):: line
  integer :: length
  character(len=max(length, len(foo(line)))) :: strout
end function 

end module 


program pp
  use mm
  print *, ff("hello",5)
end program 
```

```console
% lfortran a.f90 
ASR verify pass error: FunctionCall::m_name foo must be returning a non-void value.
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2737
    LCompilers::print_stack_on_segfault();
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2664
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 1193
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile, &time_opt);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 391
    = asr_to_llvm(asr, diagnostics,
  File "/Users/pranavchiku/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 14377
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/libasr/pass/pass_manager.h", line 318
[...]

PR #8721, opened 2025-10-15
Fix : Make `function_call_in_declaration` pass replace `FunctionParam` correctly

assem2002:
Fixes https://github.com/lfortran/lfortran/issues/8712 -- Needed by #8653

---

### WHAT IT ATTEMPTS TO FIX
- Replace `FunctionParam` expression with arguments based name rather than index found within `FunctionParam` node -- as we're replacing the FunctionParam in another scope (not where it was created originally).
- Enhance some minors + Add Comments for future refactor of the pass
- Add and register integration test.

### DIFF
--- a/src/libasr/pass/function_call_in_declaration.cpp
+++ b/src/libasr/pass/function_call_in_declaration.cpp
@@ -150,7 +150,7 @@ public:
             }
         }
         // If everthing was fine, Function would've returned earlier -- Now it's not so raise ERROR.
-        throw LCompilersException("Argument Not Found -- FuncParam Points to an argument that's proabably isn't in the current scope");
+        throw LCompilersException("Argument Not Found -- FuncParam Points to an argument that is likely not in the current scope");
     }
 
     void replace_FunctionParam_with_FunctionArgs(ASR::expr_t*& value, Vec<ASR::expr_t*>& new_args) {


assem2002:
This issue #8713 is also fixed by this current PR -- but an error emerges from `subroutine_from_function` pass . it's a separate true bug from that pass.
I believe it'll be fixed by this PR #8653 -- I'm working on it.

assem2002:
I'm targeting to refactor `function_call_in_declaration` pass later -- I believe it has so much redundant work and also using only single replacer class makes things a bit complicated when you want to modify the pass (due to state variables that tries to reroute the replacer class).

src/libasr/pass/function_call_in_declaration.cpp
```diff
@@ -104,7 +103,7 @@ private :
 public:
     Allocator& al;
     SymbolTable* new_function_scope = nullptr;
-    SymbolTable* current_scope = nullptr;
+    SymbolTable* &current_scope; // Dependency -- Passed by visitor -- Avoids mantining 2 spearate variables
```
certik:
```suggestion
    SymbolTable* &current_scope; // Dependency -- Passed by visitor -- Avoids maintaining 2 separate variables
```

src/libasr/pass/function_call_in_declaration.cpp
```diff
@@ -129,11 +129,28 @@ private :
...
+        for(size_t i = 0; i < newargsp->n; i++) {
+            char* const argument_name = ASRUtils::symbol_name(down_cast<ASR::Var_t>((*newargsp)[i])->m_v);
+            if( argument_name == parameter_name ){
+                *current_expr = newargsp->p[i];
+                return;
+            }
         }
-        *current_expr = newargsp->p[x->m_param_number];
+        // If everthing was fine, Function would've returned earlier -- Now it's not so raise ERROR.
+        throw LCompilersException("Argument Not Found -- FuncParam Points to an argument that's proabably isn't in the current scope");
```
certik:
```suggestion
        throw LCompilersException("Argument Not Found -- FuncParam Points to an argument that is likely not in the current scope");
```

src/libasr/pass/function_call_in_declaration.cpp
```diff
@@ -129,11 +129,28 @@ private :
...
+        // We'll depend on the fact that variables' names in both HELPER function and the HELPED function are the exact same
+        // So we can pick the correct argument.
+        LCOMPILERS_ASSERT(current_scope && current_scope->asr_owner)
+        ASR::Function_t* func = ASR::down_cast2<ASR::Function_t>(current_scope->asr_owner);
+        ASR::Variable_t* v = ASRUtils::EXPR2VAR(func->m_args[x->m_param_number]);
+        char* const parameter_name = v->m_name;
+        // Match on Symbol name -- Use argument from `newargsp` -- replace current
+        for(size_t i = 0; i < newargsp->n; i++) {
+            char* const argument_name = ASRUtils::symbol_name(down_cast<ASR::Var_t>((*newargsp)[i])->m_v);
+            if( argument_name == parameter_name ){
```
certik, line 147: Here you are comparing pointers, not the strings --- is this intended?

assem2002: Good catch! Yeah that's wrong. I'll push a fix.
It worked because one of those variables is a duplicate of the other -- but apparently the duplicate utility we have doesn't duplicate `char*` it just reuses the pointer.

Issue #3028, opened 2024-01-02
Could NOT find StaticZSTD (missing: zstd_LIBRARY)


jschueller:
the cmake configuration fails on archlinux because there are no zstd static libs, only shared
```
-- The C compiler identification is GNU 13.2.1
-- The CXX compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/sbin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/sbin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found ZLIB: /usr/lib/libz.a (found version "1.3")  
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find StaticZSTD (missing: zstd_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindStaticZSTD.cmake:15 (find_package_handle_standard_args)
  CMakeLists.txt:143 (find_package)


-- Configuring incomplete, errors occurred!
==> ERROR: A failure occurred in build().
    Aborting...
```

https://archlinux.org/packages/core/x86_64/zstd/

Thirumalai-Shaktivel:
I think it is caused by the missing `zstd-static`; try the following and see if it fixes the issue?
```console
$ conda install zstd-static -c conda-forge
```

jschueller:
I'm using archlinux, not conda-forge

certik:
Thanks @jschueller. We'll need to extend our cmake to also allow dynamic `zstd`. This library is needed by LLVM.

carlosal1015:
Any progress here? [`lfortran`](https://aur.archlinux.org/packages/lfortran) does not compile.

certik:
[...]

PR #8723, opened 2025-10-16
Fix: Could NOT find StaticZSTD (#3028).

Syize:
The modified cmake config will search dynamic zstd library first, then static library.


certik:
The dynamic zstd library causes problem when distributing a binary. I would like to first find the static one and use it, only if not available, use the dynamic one.

Syize:
When building distribution files for package manager like apt, pacman, it seems using dynamic one first will be better because the user may already have dynamic library installed in their system.

Perhaps we could check a marco variable like `USE_DYNAMIC_ZSTD` and then decide which one will be used first?

certik:
I see the issue --- in `main` as well as in this PR, it prefers one over the other, which doesn't satisfy all uses. So we need to make it configurable. Why don't we add a cmake option `USE_DYNAMIC_ZSTD` to configure this?

certik:
Thanks for the fix, I need to test it manually, I didn't get to it yet. I'll get to it soon.

krystophny:
Verified that PR #8723 fixes the ZSTD issue on this machine, where I had to disable `WITH_ZSTD` in order to compile with system LLVM 21:

  - OS: CachyOS Linux (Arch‑based), Linux 6.17.9-2-cachyos x86_64
  - Config 1 (LLVM 11, micromamba env `lf-llvm11`):
      - `cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON -DWITH_LSP=ON -DWITH_RUNTIME_STACKTRACE=yes -DWITH_ZSTD=ON`
      - Result: CMake succeeds, finds ZSTD as `mamba/envs/lf-llvm11/lib/libzstd.a`.
  - Config 2 (system LLVM 21):
      - `cmake -S . -B build-llvm21-system -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON -DWITH_LSP=ON -DWITH_RUNTIME_STACKTRACE=yes -DWITH_ZSTD=ON -DUSE_DYNAMIC_ZSTD=ON -DLLVM_DIR=/usr/lib/cmake/llvm`
      - Result: CMake succeeds, finds ZSTD as `/usr/lib/libzstd.so`.

  In both clean configurations, ZSTD detection works with `WITH_ZSTD=ON` and no need to disable static ZSTD.

certik:
I rebased, cleaned it up, see my last commit. I tested it locally, I think it works. The setting is now printed and uses regular cmake variables. Also I renamed the cmake file from FindStaticZSTD to FindLFortranZSTD since it now handles both static and dynamic libraries.

Question: I see you added the `zstd::libzstd` target --- is that needed? I kept it for now.

Syize:
> Question: I see you added the `zstd::libzstd` target --- is that needed? I kept it for now.

I believe it is needed. The target `zstd::libzstd` is referenced when linking against the zstd library, and a CMake error occurs if it is not defined.

I’m currently afk until next Tuesday, but I’ll provide the full error logs once I’m back.

Syize:
> but I’ll provide the full error logs once I’m back.

Here is the error log:

```
CMake Error at /usr/lib/cmake/zstd/zstdTargets.cmake:42 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: zstd::libzstd_shared

  Targets not yet defined: zstd::libzstd

Call Stack (most recent call first):
  /usr/lib/cmake/zstd/zstdConfig.cmake:42 (include)
  /usr/lib/cmake/llvm/LLVMConfig.cmake:294 (find_package)
  CMakeLists.txt:228 (find_package)
```

Add the `zstd:libzstd` target can fix this error.

Issue #8630, opened 2025-09-25
Handle assignment of `StructType` variables where lhs is `StructInstanceMember` of class variable


jinangshah21:
MRE:
```fortran 
module m_diagnostic
    implicit none

    type :: label_t
        integer :: x
    end type label_t

    type :: toml_diagnostic
        type(label_t), allocatable :: label(:)
    end type toml_diagnostic

    type :: toml_parser
        type(toml_diagnostic), allocatable :: diagnostic
    end type toml_parser


contains

    subroutine syntax_error(diag)
        type(toml_diagnostic), allocatable, intent(out) :: diag

        allocate(diag)                    
        allocate(diag%label(1))          
        diag%label(1) = label_t(5)
    end subroutine syntax_error

end module m_diagnostic

program test_diag
    use m_diagnostic
    implicit none
    type(toml_parser), allocatable :: pars
    allocate(pars)
    call temp(pars)

contains

    subroutine temp(parser)
        class(toml_parser), intent(inout) :: parser
        type(toml_diagnostic), allocatable :: tmp_diag

        call syntax_error(tmp_diag)
        parser%diagnostic = tmp_diag
        print *, tmp_diag%label(1)%x
    end subroutine temp
end program test_diag
```

Outputs:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 --new-classes --realloc-lhs-arrays
Runtime error: Array '' index out of bounds.

Tried to access index 1 of dimension 1, but valid range is 32764 to 1223383021.
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran a.f90 && ./a.out
           5
```

jinangshah21:
```fortran
module m_diagnostic
    implicit none

    type :: toml_diagnostic
        integer, allocatable :: label(:)
    end type toml_diagnostic

    type :: toml_parser
        type(toml_diagnostic), allocatable :: diagnostic(:)
    end type toml_parser
end module m_diagnostic

program test_diag
    use m_diagnostic
    implicit none
    type(toml_parser), allocatable :: pars, pars2
    allocate(pars)
    allocate(pars2)
    allocate(pars%diagnostic(1))
    allocate(pars2%diagnostic(1))
    pars%diagnostic(1) = toml_diagnostic([1,2,3])
    pars2 = pars
    pars2%diagnostic(1)%label = [4,5,6]
    print *, pars%diagnostic(1)%label
end program test_diag
```

[...]

PR #8731, opened 2025-10-18
fix: [new_classes] allocate array members of struct on heap

jinangshah21:
Towards #8630


jinangshah21:
This allocates array members of struct on heap, still we need to deallocate struct properly such that all memory allocated gets freed.

PR #8732, opened 2025-10-20
Fix(Minor) : Make sure to compare actual names -- not the pointers.

assem2002:
Fix according to comment here https://github.com/lfortran/lfortran/pull/8721#discussion_r2434724382


certik:
It's concerning that it was working before, likely our test coverage is not great.

assem2002:
One of the variables is a duplicate of the other one -- that's always the case in this current place; but when we duplicate we don't duplicate strings (char*) we just reuse the pointer.
 I don't think it's a testing problem on our end.

certik:
If it's always a duplicate, and the change in this PR cannot be tested, then should we revert it?

assem2002:
No revert needed.
The changed code is used in main by test `string_77`. This PR is just for correctness.

PR #8733, opened 2025-10-20
fix:: Handle Class(*) Print of scalars with --new-classes flag

### LINKED ISSUE #8716: Handle `class(*)` printing in `select_type`
MRE:
```fortran
program demo
    class(*), allocatable :: x

    allocate(x, source=42)

    select type (x)
    type is (integer)
        print *, "Integer value:", x
    class default
        print *, "Unknown type"
    end select
end program
```

Output:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90 --new-classes
Integer value:    4210128
```

YashNagda17:
Fixes #8716 

Added Test-Case:
```
program class_71
    class(*), allocatable :: x
    allocate(x, source=42)
    select type (x)
    type is (integer)
        print *, x
    class default
        print *, "Unknown type"
    end select
end program class_71
```

Updated Main:
```
$ lfortran a.f90 --new-classes
42
```

Gfortran:
```
$ gfortran a.f90 & ./a.out
[1] 149770
          42
```

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14368,18 +14368,14 @@ public:
                     ASR::is_a<ASR::Var_t>(*x.m_args[i]) &&
                     ASRUtils::is_unlimited_polymorphic_type(x.m_args[i])) {
                     // Extract data pointer from polymorphic struct
-#if LLVM_VERSION_MAJOR >= 14
-                    // LLVM 14+ uses opaque pointers, get type from ASR instead
-                    ASR::ttype_t* arg_type = ASRUtils::expr_type(x.m_args[i]);
-                    llvm::Type* polymorphic_struct_type = llvm_utils->get_type_from_ttype_t_util(
-                        x.m_args[i], arg_type, module.get());
+                    // Get struct type from ASR, as LLVM 15+ Doesn't support direct 
+                    // extraction from tmp as getPointerElementType()
+                    ASR::symbol_t* struct_sym = ASRUtils::symbol_get_past_external(
+                        ASRUtils::get_struct_sym_from_struct_expr(x.m_args[i]));
+                    ASR::Struct_t* struct_t = ASR::down_cast<ASR::Struct_t>(struct_sym);
+                    llvm::Type* polymorphic_struct_type = llvm_utils->getClassType(struct_t);
                     llvm::Value* data_ptr = llvm_utils->create_gep2(
                         polymorphic_struct_type, tmp, 1);
-#else
-                    // LLVM < 14: use getPointerElementType()
-                    llvm::Value* data_ptr = llvm_utils->create_gep2(
-                        tmp->getType()->getPointerElementType(), tmp, 1);
-#endif
 #if LLVM_VERSION_MAJOR >= 17
                     // LLVM 17+: use PointerType::getUnqual()
                     tmp = llvm_utils->CreateLoad2(


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14362,6 +14362,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    llvm::Value* data_ptr = llvm_utils->create_gep2(
+                        polymorphic_struct_type, tmp, 1);
+#if LLVM_VERSION_MAJOR >= 17
+                    // LLVM 17+: use PointerType::getUnqual()
+                    tmp = llvm_utils->CreateLoad2(
+                        llvm::PointerType::getUnqual(llvm::Type::getInt8Ty(context)), data_ptr);
+#else
+                    // LLVM < 17: use Type::getInt8PtrTy()
+                    tmp = llvm_utils->CreateLoad2(llvm::Type::getInt8PtrTy(context), data_ptr);
+#endif
```
certik, line 14386: In subsequent PR: create a function for this, and move ifdefs into llvm_utils (I think). The asr_to_llvm should be completely LLVM version independent.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14362,6 +14362,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    ASRUtils::is_unlimited_polymorphic_type(x.m_args[i])) {
+                    // Extract data pointer from polymorphic struct
+                    // Get struct type from ASR, as LLVM 15+ Doesn't support direct 
+                    // extraction from tmp as getPointerElementType()
+                    ASR::symbol_t* struct_sym = ASRUtils::symbol_get_past_external(
+                        ASRUtils::get_struct_sym_from_struct_expr(x.m_args[i]));
+                    ASR::Struct_t* struct_t = ASR::down_cast<ASR::Struct_t>(struct_sym);
+                    llvm::Type* polymorphic_struct_type = llvm_utils->getClassType(struct_t);
+                    llvm::Value* data_ptr = llvm_utils->create_gep2(
+                        polymorphic_struct_type, tmp, 1);
```
certik, line 14378: @jinangshah21 is this the correct fix to be doing this in the LLVM backend? I think the ASR is incorrect for this case above:
```
(TranslationUnit
    (SymbolTable
        1
        {
            class_71:
                (Program
                    (SymbolTable
                        2
                        {
                            x:
                                (Variable
                                    2
                                    x
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Allocatable
                                        (StructType
                                            []
                                            []
                                            .false.
                                            .true.
                                        )
                                    )
                                    2 ~unlimited_polymorphic_type
                                    Source
                                    Public
                                    Required
                                    .false.
                                    .false.
                                    .false.
                                    ()
                                    .false.
                                    .false.
                                ),
                            ~select_type_block_:
                                (Block
                                    (SymbolTable
                                        4
                                        {

                                        })
                                    ~select_type_block_
                                    [(Print
                                        (StringFormat
                                            ()
                                            [(Var 2 x)]
                                            FormatFortran
                                            (Allocatable
                                            )
                                            ()
                                        )
                                    )]
                                ),
                            ~unlimited_polymorphic_type:
                                (Struct
                                    (SymbolTable
                                        3
                                        {

                                        })
                                    ~unlimited_polymorphic_type
                                    (StructType
                                        []
                                        []
                                        .false.
                                        .true.
                                    )
                                    []
                                    []
                                    []
                                    Source
                                    Public
                                    .false.
                                    .true.
                                    []
                                    ()
                                    ()
                                )
                        })
                    class_71
                    []
                    [(Allocate
                        [((Var 2 x)
                        []
                        ()
                        ()
                        ())]
                        ()
                        ()
                        (IntegerConstant 42 (Integer 4) Decimal)
                    )
                    (SelectType
                        (Var 2 x)
                        ()
                        [(TypeStmtType
                            (Integer 4)
                            [(BlockCall
                                -1
                                2 ~select_type_block_
                            )]
                        )]
                        [(Print
                            (StringConstant
                                "Unknown type"
                                (String 1 (IntegerConstant 12 (Integer 4) Decimal) ExpressionLength DescriptorString)
                            )
                        )]
                    )]
                )
        })
    []
)
```

It just prints `(Var 2 x)`, but that variable `x` is a `class(*)` variable, so we can't print it. Instead, we need to cast it to Integer. Then we just print an integer as normal, we don't handle `class(*)` in a format.

YashNagda17: When comparing with old implementation (without --new-classes flag), the asr generated is exactly the same as with --new-classes flag

So I basically took the same path as old implementation, found the problematic part, and added additional logic to handle --new-classes flag. Anyways, got your point...

src/libasr/codegen/asr_to_llvm.cpp
YashNagda17: Ohk, working on it...

src/libasr/codegen/asr_to_llvm.cpp
certik: Yes, the old classes also didn't handle this.

@jinangshah21 are you ok with merging this as an intermediate step to get more tests working, and we'll rework it later to use proper ASR casting, or will it make it harder to rework later?

If this PR doesn't make it harder to implement the proper fix, then I would say let's merge it now, to get more tests working.

jinangshah21: Yes, correct fix is to cast `class(*)` in print. But this is quick fix which will not make any problem later, so I think we can merge this.

certik (approved):

certik:
Awesome, thanks @jinangshah21  for the review and @YashNagda17 for the PR!

PR #8735, opened 2025-10-20
Tests : Add integration test

### LINKED ISSUE #7980: bug : wrong string length (depending on null char)
```.f90
module m
  integer :: i = 10
end module m

program test
use m
call ff()
contains 
subroutine ff()
  character(i), allocatable :: cc
  i = 3
  cc = "as"
  print *, len(cc) ! Used to print the return of `strlen`
  print *, i
end subroutine

end program
```
```console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranUpStream/lfortran/src/bin/lfortran tt3.f90 
2
3
10
3
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt3.f90 && ./a.out 
          10
           3

```

assem2002:
closes #7980


certik:
Thanks!

PR #8740, opened 2025-10-22
Enh : Free Return of calls to runtime function `lfortran_string_format()`

assem2002:
- Introduce `StringFormatReturn` class to manage the return of any call to `lfortran_string_format()`
- Free returned value (call return) after their consumers actually did consume it -- Current consumers are `ASR::Print`, `ASR::FileWrite`
- Do minor changes for some helper-utility functions -- add them to the llvm utility class `llvmUtils`

### DIFF
--- a/src/libasr/codegen/llvm_utils.h
+++ b/src/libasr/codegen/llvm_utils.h
@@ -633,7 +633,7 @@ namespace LCompilers {
             /**
              *@class StringFormatReturn
              *
-             *@brief Hold return of `_lcompilers_string_format_fortran` 
+             *@brief Holds return of `_lcompilers_string_format_fortran` 
              *        for the consumer of the return to free it after consumtion (print, write, ..).
              *
              *@details ** `_lcompilers_string_format_fortran` return a `char*` which needs
@@ -642,16 +642,19 @@ namespace LCompilers {
              */ 
             class StringFormatReturn {
                 LLVMUtils   *llvmUtils_instance_;
-                llvm::Value *return_val = nullptr; // Holds return of call to `_lcompilers_string_format_fortran`
-                StringFormatReturn(LLVMUtils *instance): llvmUtils_instance_(instance) {}
-
-                bool is_freed()           { return return_val == nullptr; }
+                llvm::Value *return_val = nullptr; // Holds `_lcompilers_string_format_fortran()` call return
                 bool previous_one_freed() { return return_val == nullptr; }
 
-                public:
+            public:
 
+                StringFormatReturn(LLVMUtils *instance): llvmUtils_instance_(instance) {}
+
+                /// Check that nothing is pending to be freed.
+                bool all_clean(){ return return_val == nullptr; }
+                
                 void set(llvm::Value* val) {
-                    LCOMPILERS_ASSERT_MSG(previous_one_freed(), "Previous StringFormatReturn not freed")
+                    LCOMPILERS_ASSERT_MSG(previous_one_freed(),
+                                          "Previous StringFormatReturn not freed")
                     return_val = val;
                 }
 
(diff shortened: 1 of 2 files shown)


src/libasr/codegen/llvm_utils.h
```diff
@@ -648,6 +630,46 @@ namespace LCompilers {
...
+                void free() {
+                    if(!return_val) return;
+                    llvmUtils_instance_->lfortran_free(return_val);
+                    return_val = nullptr;
+                }
+
+                ~StringFormatReturn()  {
+                   if(std::uncaught_exceptions() == 0) {
+                        LCOMPILERS_ASSERT_MSG(is_freed(), "StringFormat Return Not Freed");
+                   }
```
certik: We should not be using exceptions for this. Let's figure out another mechanism.

src/libasr/codegen/llvm_utils.h
```diff
@@ -648,6 +630,46 @@ namespace LCompilers {
...
+             *@class StringFormatReturn
+             *
+             *@brief Hold return of `_lcompilers_string_format_fortran` 
+             *        for the consumer of the return to free it after consumtion (print, write, ..).
+             *
+             *@details ** `_lcompilers_string_format_fortran` return a `char*` which needs
+             *              to be freed by the consumer of that string.
+             *         ** It complains if current return wasn't freed.
+             */ 
+            class StringFormatReturn {
```
certik, line 643: Where is this class used? I can't find any usage of it.

src/libasr/codegen/llvm_utils.h
assem2002: I don't like that either :). but I liked the idea of making the class responsible of the whole process, it makes sure if you freed the thing or not. 
Anyways, I'll handle that by using a call somefunction `bool nothing_to_free()` after we finish generating llvm code.

src/libasr/codegen/llvm_utils.h
assem2002: here https://github.com/lfortran/lfortran/pull/8740/files#diff-d662a9d3de5c3ed45765e49102082d3b9e964548f6a0bab7e97d34d5ab5aced2 -- inside `llvm::Value* LLVMUtils::string_format_fortran()` it sets  the return.

In https://github.com/lfortran/lfortran/pull/8740/files#diff-48a7161397f9eb8b730f0cbec2f11ce1471aac961dc6b16dc558cf161c375054 -- used in `handle_prin()t` and `visit_filewrite()` to free the value we just sat.

certik: I see, it's `stringFormat_return`, where there is a global variable `            StringFormatReturn stringFormat_return {this};`. 

I don't like that. We should not be using global variables.

src/libasr/codegen/llvm_utils.h
certik: Yes, that's fine. Put the variable into the visitor, so that it's not global and let's not use exceptions. Other than that, you can use a class, I think that's fine.

src/libasr/codegen/llvm_utils.h
assem2002: the class declaration and the instance are inside `llvmUtils` class, Is that fine? 
We try to create a single instance only and the class is only visible to `llvmUtils`, so we have a `stringFormat_return` instance per `LLVMUtils` instance only.

assem2002:
Latest commit removes the block where we throw/abort in class destructor -- mentioned here https://github.com/lfortran/lfortran/pull/8740#discussion_r2450198460

@certik If you have any other suggestions or enchantments on how to make the class instance more encapsulated, please inform me.

src/libasr/codegen/llvm_utils.h
certik: I see, that's fine.

certik (approved):
I think that's fine now.

PR #8741, opened 2025-10-23
Fix: allow functions' return type to use variables from host scope

### LINKED ISSUE #8686: Function with character return sets length incorrectly
```.f90
program pp
  integer :: n
  contains

  character(n) function tolower (c) result (res)
    character, intent(in) :: c
  end function

end program  pp
```
It set length to `-3`. We need to tackle this.

I'd also like to not use `extract_len` (we use it in 2 spots), It follows the old length design mentioned here https://github.com/lfortran/lfortran/commit/fd71d22fba070076efa4f04e7cffb38c8a1b9ceb#diff-1599d8226e12e870f72451766a11638199a4cdde172ca0ef9c057912072a5bcfR203

akramhany:
Fix #8686

----

I did the following changes:
- refactor `extract_len` function to remove any unused cases.
- modify the `visit_FunctionType` verification check to allow using any variables that are either external or defined in a parent scope.

### DIFF
--- a/src/libasr/asr_verify.cpp
+++ b/src/libasr/asr_verify.cpp
@@ -885,7 +885,7 @@ public:
             s = ASRUtils::symbol_get_past_external(x.m_v);
         }
 
-        // Allow any variable that is either external or is not defined in this scope to pass FunctionType verification
+        // Allow any string variable that is either external or is not defined in this scope to pass FunctionType verification
         if (is_a<ASR::Variable_t>(*s) && (is_a<ASR::ExternalSymbol_t>(*x.m_v) ||
             (_is_return_type_string && !current_symtab->get_symbol(x_mv_name)))) {
             non_global_symbol_visited = false;


certik:
For arrays we insert autogenerated simple functions to handle this case. For strings it looks like we just link the global variable directly for subroutines (that already work), so this PR just follows the same design, which is fine for now.

src/libasr/asr_verify.cpp
```diff
@@ -883,7 +883,9 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         if (check_external) {
             s = ASRUtils::symbol_get_past_external(x.m_v);
         }
-        if (is_a<ASR::Variable_t>(*s) && is_a<ASR::ExternalSymbol_t>(*x.m_v)) {
+
+        // Allow any variable that is either external or is not defined in this scope to pass FunctionType verification
+        if (is_a<ASR::Variable_t>(*s) && (is_a<ASR::ExternalSymbol_t>(*x.m_v) || !current_symtab->get_symbol(x_mv_name))) {
```
certik: Only do it for strings --- check that the type is a character. Disallow for anything else.

src/libasr/asr_verify.cpp
```diff
@@ -883,11 +884,16 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         if (check_external) {
             s = ASRUtils::symbol_get_past_external(x.m_v);
         }
-        if (is_a<ASR::Variable_t>(*s) && is_a<ASR::ExternalSymbol_t>(*x.m_v)) {
+
+        // Allow any variable that is either external or is not defined in this scope to pass FunctionType verification
```
certik: Only strings, right?
```suggestion
        // Allow any string variable that is either external or is not defined in this scope to pass FunctionType verification
```

akramhany: Correct, it should be a string variable.

Issue #8306, opened 2025-08-09
Bounds Checking


certik:
Here are the improvements that we should implement:

* [ ] https://github.com/lfortran/lfortran/issues/10003
* [x] https://github.com/lfortran/lfortran/issues/10812
* [x] https://github.com/lfortran/lfortran/issues/8726#issuecomment-3411817520
* [x] Add line/column number, filename (embed at compile time)
* [x] https://github.com/lfortran/lfortran/issues/8728
* [x] https://github.com/lfortran/lfortran/issues/8750
* [x] https://github.com/lfortran/lfortran/pull/8887
* [x] https://github.com/lfortran/lfortran/issues/8689
* [x] https://github.com/lfortran/lfortran/issues/9028
* [x] https://github.com/lfortran/lfortran/issues/9086
* [x] https://github.com/lfortran/lfortran/issues/8593
* [x] https://github.com/lfortran/lfortran/issues/8926 (this might not be a bug in bounds checker, but it might)
* [x] https://github.com/lfortran/lfortran/issues/7891 (if it is fixed, we can close it)
* [x] https://github.com/lfortran/lfortran/issues/8655
* [x] #8727
* [ ] Better error message (if the source file is available, make it look like a compiler error message, with squiggles of the exact index that is wrong, etc.): https://github.com/lfortran/lfortran/issues/8307
* [x] https://github.com/lfortran/lfortran/issues/9526
* [ ] Consider adding a full span of the index expression that is out of bounds
* [x] Add colors (--runtime-color flag, uses _lfortran_set_use_runtime_colors)
* [x] Check out of bounds at compile time for fixed arrays
* [x]  https://github.com/lfortran/lfortran/issues/9186
* [x] https://github.com/lfortran/lfortran/issues/9512

---

* [ ] Add stacktrace
* [ ] Consider embedding the full line, that way we can always print a full message. This might be hidden behind a compiler option, as the resulting binary might be too big.
* [x] https://github.com/lfortran/lfortran/issues/8701
* [x] https://github.com/lfortran/lfortran/issues/8295
* [x] https://github.com/lfortran/lfortran/issues/8594
* [x] https://github.com/lfortran/lfortran/issues/8687

PR #8749, opened 2025-10-24
feat: add line and column to runtime error messages

xaerru:
Towards https://github.com/lfortran/lfortran/issues/8306

### DIFF
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -111,7 +111,7 @@ Result<FortranEvaluator::EvalResult> FortranEvaluator::evaluate(
 
     // ASR -> LLVM
     Result<std::unique_ptr<LLVMModule>> res3 = get_llvm3(*asr,
-        pass_manager, diagnostics, lm.files.back().in_filename,
+        pass_manager, diagnostics, lm, lm.files.back().in_filename,
         nullptr);
     std::unique_ptr<LCompilers::LLVMModule> m;
     if (res3.ok) {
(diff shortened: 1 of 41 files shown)


src/libasr/codegen/llvm_utils.h
```diff
@@ -280,14 +280,38 @@ namespace LCompilers {
...
+                lm.files.push_back(fl);
+                std::string input;
+                if (!read_file(infile, input)) {
+                    throw CodeGenError("File '" + infile + "' cannot be opened.");
+                }
+                lm.init_simple(input);
+                lm.file_ends.push_back(input.size());
+                lm.pos_to_linecol(lm.output_to_input_pos(loc_first, false),
+                    line, column, fl.in_filename);
+            }
```
certik: You can't do it like that --- you need to pass in the existing LocationManager using an argument. This approach has two huge drawbacks:

* Incorrect if preprocesor, prescanning or include files are used
* Very slow, as it is reading the file every time from scratch, don't want to do that. It's doing it for every debug check.

src/libasr/codegen/llvm_utils.h
```diff
@@ -280,14 +280,38 @@ namespace LCompilers {
...
+            void generate_runtime_error(llvm::Value* cond, std::string message, std::string infile, Location loc, Args... args)
             {
                 llvm::Function *fn = builder->GetInsertBlock()->getParent();
 
                 llvm::BasicBlock *thenBB = llvm::BasicBlock::Create(context, "then", fn);
                 llvm::BasicBlock *mergeBB = llvm::BasicBlock::Create(context, "ifcont");
 
+                uint32_t line, column;
+                if (infile != "" && loc.first != 0 && loc.last != 0) {
+                    get_line_column(loc.first, line, column, infile);
```
certik: Get this information from the existing LocationManager

certik:
Is this ready for review?

src/lfortran/fortran_evaluator.cpp
```diff
@@ -353,13 +353,14 @@ Result<std::unique_ptr<LLVMModule>> FortranEvaluator::get_llvm2(
         i.e. time taken by optimizations, and used when
         `--time-report` flag is used
 */
+// Add LocationManager here
```
certik, line 356: Was this a TODO comment that is not needed anymore?

certik (approved):
I think this is now done correctly, thanks!

Issue #8072, opened 2025-07-15
Bug : global fixedSize or PointerToData array doesn't setup its internals.


assem2002:
```.f90
module mm
    type t 
        integer , allocatable :: arr(:)
    end type
    type(t) :: instance_arr(39)
end module

program pp
    use mm
    print*, allocated(instance_arr(1)%arr)
end program 
```
```console
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranUpStream/lfortran/src/bin/lfortran tt2.f90 
Segmentation fault (core dumped)
```

PR #8751, opened 2025-10-24
fix: check if array is null inside allocated() function

YashNagda17:
Towards #8072
Adding an if condition to check if argument inside allocated() function is not nullptr, which throws segfault...  This is added in LLVM Generation Block... Handle null cases separately
Added Test-Case:
```
module mm
    type t 
        integer,allocatable :: arr(:)
    end type
    type(t) :: instance_arr(39)
end module

program pp
    use mm
    print *, allocated(instance_arr(1)%arr)
    if (allocated(instance_arr(1)%arr)) error stop
    allocate(instance_arr(1)%arr(10))
    print*, allocated(instance_arr(1)%arr)
    if (.not. allocated(instance_arr(1)%arr)) error stop
end program
```

Gfortran:
```
$ gfortran a.f90 &./a.out
[1] 114199
 F
 T
```

Updated Main:
```
$ lfortran a.f90
F
T
```

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -174,6 +174,10 @@ time_section "🧪 Testing M_CLI2" '
   git checkout 600737dc23004c1efa10d2233d4a631d0521fd53
   fpm --compiler=$FC build --flag "--realloc-lhs-arrays"
   fpm --compiler=$FC test --flag "--realloc-lhs-arrays"
+
+  git clean -dfx
+  fpm --compiler=$FC build --flag "--realloc-lhs-arrays --new-classes"
+  fpm --compiler=$FC test --flag "--realloc-lhs-arrays --new-classes"
   print_success "Done with M_CLI2"
   cd ..
 '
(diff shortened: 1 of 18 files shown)


YashNagda17:
The error seems unrelated to the changes that were made

src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -853,12 +868,18 @@ namespace LCompilers {
...
                     llvm::Type::getInt64Ty(context)),
                 builder->CreatePtrToInt(llvm::ConstantPointerNull::get(memory_holder_type),
                     llvm::Type::getInt64Ty(context)));
+            builder->CreateBr(mergeBB);
+            // Merge block with PHI node
+            builder->SetInsertPoint(mergeBB);
+            llvm::PHINode *result = builder->CreatePHI(llvm::Type::getInt1Ty(context), 2, "is_allocated");
+            result->addIncoming(llvm::ConstantInt::getFalse(context), entryBB); // from entry (NULL case)
+            result->addIncoming(is_data_allocated, checkDataBB); // from check_data block
+            return result;
```
certik, line 882: I think this can be refactored to use the "if statement builder" that we have, to reduce all the boilerplate.

PR #8758, opened 2025-10-27
Enh : Finalize Symbol Table in backend (llvm) -- Handle strings only

### LINKED ISSUE #8129: Strings TODO
* [x] Regressions fix: https://github.com/lfortran/lfortran/pull/7861#issuecomment-3110144864
* [x] Remove null character from strings in the runtime

After:

* [x] Proper deallocation, in the backend (when the symbol table goes out of scope) [all strings are allocated on heap + its the string descriptor] [#8758](https://github.com/lfortran/lfortran/pull/8758)

assem2002:
closes #8129
towards #3770

- Feat : Introduce class `LLVMFinalize` : 
    - Handles Finalization of the whole TU.
    - Have 4 major finalize functions (that actually insert finalization instructions into LLVM)
    - For now we have `finalize_string()` implemented. `finalize_struct()`, `finalize_array()`, `finalize_scalar()` to do.
    - The class avoids finalizing any global variable (module variables , TU variables (if any), save variables)
- Map symbols with their return block (llvm basic block) to insert finalization instruction before it.
- Enh : Use `visit_blockCall` instead of doing that manually in some spots, that's to have proper return basic block
- Enh : Use basic blocks with `AssociateBlock`, that's to have proper return basic block.
- Stop handling strings in `insert_deallocate` pass, that's to avoid double freeing.
- Tests : Adjust integration test as it had a dangling pointer bug.
- Tests : Update reference tests + Add integration test for string that should explode the memory if not freed.

### DIFF
--- a/integration_tests/string_37.f90
+++ b/integration_tests/string_37.f90
@@ -14,7 +14,8 @@ program string_37
     end function maybe
     function cast_to_string() result(str)
         character(len=:), pointer :: str
-        character(len=5), target :: str2
+        character(len=5), pointer :: str2
+        allocate(str2)
         str2 = "Hello"
         str => str2
     end function cast_to_string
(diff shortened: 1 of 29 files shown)


integration_tests/string_37.f90
```diff
@@ -14,7 +14,8 @@ function maybe(string) result(maybe_string)
     end function maybe
     function cast_to_string() result(str)
         character(len=:), pointer :: str
-        character(len=5), target :: str2
+        character(len=5), pointer :: str2
+        allocate(str2)
         str2 = "Hello"
         str => str2
```
certik, line 20: Previously we were returning a dangling pointer, so this change is correct.

PR #8759, opened 2025-10-28
enh : Use `realloc_lhs_arrays` flag as a check before allocating struct arrays members in `struct_deepcopy`

### LINKED ISSUE #8683: migrate `realloc_lhs_arrays` to `CompilerOptions` from `PassOptions`
I saw that we have `realloc_lhs_arrays` as a `PassOption` flag and it is implemented in `array_op` pass, but for struct deepcopy we need `realloc_lhs_arrays` flag in ASR -> LLVM also. Consider below example:
```fortran
module A
implicit none

type D
    integer :: member
end type

type E
    type(D), allocatable :: ds(:)
end type

type F
    type(E), allocatable :: es(:)
end type

contains

function get_ds() result(ds)
type(D), allocatable :: ds(:)
ds = [D(1), D(2), D(3)]
end function

subroutine new_ds(self)
type(F), intent(inout) :: self
if( .not. allocated(self%es) ) error stop
self%es(2)%ds = get_ds()
end subroutine
end module

program allocate_12
use A
implicit none
type(F) :: fobj(2)
allocate(fobj(2)%es(3))
tmp = fobj(2)
end program
```

Here we can't allocate struct arrays members at ASR -> ASR Passes (it is runtime). So we need `realloc_lhs_arrays` flag at ASR -> LLVM. For now I am opening PR with assuming `realloc_lhs_arrays` true for struct deepcopy in ASR -> LLVM.

Refer #8682

AnshMehta1:
Closes #8683
Adresses https://github.com/lfortran/lfortran/pull/8682#discussion_r2407720666

I haven't added the check using `realloc_lhs_arrays` for https://github.com/lfortran/lfortran/pull/8682#discussion_r2407717262 because I don't feel we need this check their. Also, it leads to some additional failures. So, I am ignoring this for now.

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -8656,8 +8656,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al, 
                 LCOMPILERS_ASSERT(false);
             }
 
-            // realloc_lhs_arrays
-            if (ASRUtils::is_allocatable(src_expr)) {
+            if (llvm_utils->compiler_options.po.realloc_lhs_arrays && ASRUtils::is_allocatable(src_expr)) {
                 uint64_t data_type_size = data_layout.getTypeAllocSize(llvm_data_type);
                 llvm::Value* total_memory = builder->CreateMul(num_elements,
                     llvm::ConstantInt::get(context, llvm::APInt(32, data_type_size)));


jinangshah21 (commented):
There is one more place where this flag is needed to be added (See https://github.com/lfortran/lfortran/pull/8682#discussion_r2407717262). Can you please add that also?

jinangshah21:
Other than above comment I think this is good.

AnshMehta1:
> There is one more place where this flag is needed to be added (See [#8682 (comment)](https://github.com/lfortran/lfortran/pull/8682#discussion_r2407717262)). Can you please add that also?

I have mentioned my thoughts regarding that at https://github.com/lfortran/lfortran/pull/8759#issue-3561134953.

certik:
We should also add tests for this --- it should give a bounds check error without it, and work with the option.

AnshMehta1:
> We should also add tests for this --- it should give a bounds check error without it, and work with the option.

Yes, the test is already added in #8682 for this purpose. It works perfectly as per your comment.

PR #8765, opened 2025-10-30
Fix nested_16 on macOS

certik:


certik:
It looks like the CI has only one error:
```
~/repos/lfortran/integration_tests$ lfortran class_32.f90 --new-classes
c%x:     1
d%x:     42
c%x after assignment:     42
d%x:     3
c%x:     42
c_base%obj%x:     42
d_base%obj%x:     3
c%x:     20
c_base%obj%x:     42
c_super%x:     1
d_derived%x:     42
class_32.out(42237,0x20c0aa0c0) malloc: Heap corruption detected, free list is damaged at 0x600003de80d0
*** Incorrect guard value: 168384192839680
class_32.out(42237,0x20c0aa0c0) malloc: *** set a breakpoint in malloc_error_break to debug
```
And I can reproduce it locally. Needs to be debugged and fixed.

certik:
The fix is simple, all tests pass, so let's merge this, to unblock us. If we need to further refine it, we can.

PR #8766, opened 2025-10-30
fix: load array var properly in getting vptr on polymorphism call

certik:
Build on top of https://github.com/lfortran/lfortran/pull/8626.


certik:
The added tests here passes in main for me, so I just added it in https://github.com/lfortran/lfortran/pull/8767.

certik:
The 3rd-party tests also both work for me locally, so I enabled those in https://github.com/lfortran/lfortran/pull/8768.

Maybe the changes in this PR are not needed.

certik:
All the individual parts of this PR was merged into master. What is left are a few fixes that don't seem to be needed --- there is no test in master that fails that works here.

Consequently, I am leaving it here for reference, in case there are some tests that we need to add. I'll mark it as draft till then. Otherwise we can close it.

Issue #8763, opened 2025-10-29
bug: cannot print polymorphic variable through `print` statement


kmr-srbh:
### Minimal Reproducible Example

```fortran
program main
   type :: type_a
      integer :: a
   end type type_a

   class(type_a), allocatable :: x

   allocate(x)
   x%a = 10

   select type(x)
    type is (type_a)
      print *, x
    class default
      print *, "default"
   end select
end program
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
clang: warning: overriding deployment version from '16.0' to '26.0' [-Woverriding-deployment-version]
          10
(lf) saurabhkumar@Mac lfortran % lfortran test.f90           
LCOMPILERS_ASSERT failed: /Users/saurabhkumar/OpenSource/lfortran/src/libasr/../libasr/asr_utils.h
function type_to_str_fortran_expr(), line number 992 at 
expr != nullptr
ERROR MESSAGE:
`expr` should be non-null for `StructType`
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2737
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2664
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 1194
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 394
    if (res.ok) {
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 14450
    v.visit_asr((ASR::asr_t&)asr);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 359
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 333
    case asrType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/saurabhkumar/OpenSource/lfortran/src/libasr/asr_base_visitor.h", line 360
[...]

PR #8773, opened 2025-10-31
WIP: Investigation of Polymorphic Class 

opixdown:
solving for issue #8763 

Hi all,
This PR is not a final fix, it is meant for discussion and reference so everyone can understand the issue.
I am still investigating the full solution, but these changes helped identify where the problem actually lies.

The summary and progress till now:

1. Compiler crash fixed
The compiler was crashing because ASRUtils::type_to_str_fortran_expr() was called with a nullptr.
I passed the correct expr instead of nullptr, and the crash is now gone.

2. Code generation bug discovered
I found that user-defined structs/classes were not being handled correctly in:
ASR::is_a<ASR::StructType_t>(*type) && !ASRUtils::is_class_type(type)

The is_class_type check incorrectly prevented struct types from being processed, which caused missing backend IR.
After removing that part, codegen continued.

3.Class wrapper vs actual struct layout
During testing:

I considered the this code as suggested by assem to see whats happening 

type a
  integer :: ii
end type
type(a), allocatable :: x

I found that class(a) in LLVM IR has an extra wrapper:
%a = type { i32 (...)**, i32 }

So the compiler was serializing the wrapper instead of the underlying struct.
I confirmed this by unwrapping the pointer and seeing:

"store i32 10, i32* %23"

which shows the correct data is actually stored.

4. Debug code added to locate the incorrect value being printed

Even after storing 10 correctly, printing showed 0.
To understand why, I added debug print statements around the type-id loading.

This revealed:
•	The actual object’s type-id is loaded as i64 
•	The expected vtable type-id is loaded as i32 

So the compiler ends up comparing i64 vs i32, which will never match.
i think that why it behaves incorrectly and printing fails.

At this point, I’m stuck because of this type id mismatch and need some guidance on how the vtable/type-id fields are intended to be typed and loaded.

Also, thanks to @assem2002 for helping me dig through this layer by layer it made the investigation much clearer.

This PR is here mainly so everyone can see the progress


jinangshah21:
In current codebase, we have two class design working old and new, for new you can see `compiler_options.new_classes` flag in llvm backend. You need to fix it with new classes code. And new classes code doesn't have type id for class type variables. You need to unwrap class type variable at correct place in llvm backend.

opixdown:
@jinangshah21 Thanks for the clarification
I’ll look in the new-classes implementation and find the correct place where unwrapping must happen.

assem2002:
Great work! 
I think you can make things work with this diff 
```diff
diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index 0c073668d..6830db7c5 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14354,9 +14354,7 @@ public:
                 }
                 // Handling for polymorphic class variables in print statements with --new-classes
                 if (compiler_options.new_classes &&
-                    current_select_type_block_type_asr != nullptr &&
-                    ASR::is_a<ASR::Var_t>(*x.m_args[i]) &&
-                    ASRUtils::is_unlimited_polymorphic_type(x.m_args[i])) {
+                    ASR::is_a<ASR::Var_t>(*x.m_args[i])) {
                     // Extract data pointer from polymorphic struct
                     // Get struct type from ASR, as LLVM 15+ Doesn't support direct 
                     // extraction from tmp as getPointerElementType()
@@ -14366,7 +14364,11 @@ public:
                     llvm::Type* polymorphic_struct_type = llvm_utils->getClassType(struct_t);
                     llvm::Value* data_ptr = llvm_utils->create_gep2(
                         polymorphic_struct_type, tmp, 1);
-                    tmp = llvm_utils->CreateLoad2(llvm_utils->i8_ptr, data_ptr);
+                    if(ASRUtils::is_unlimited_polymorphic_type(x.m_args[i])){
+                        tmp = llvm_utils->CreateLoad2(llvm_utils->i8_ptr, data_ptr);
+                    } else {
+                        tmp = data_ptr;
+                    }
                 }
                 if(!tmp->getType()->isPointerTy() ||
                     ASR::is_a<ASR::PointerToCPtr_t>(*x.m_args[i])){
```

opixdown:
@assem2002 thankyou assem for the code, i runned it but there was quick func name issue the function we were calling 

ASRUtils::get_struct_from_expr(al, x.m_args[i]) 

the corect function was 

ASRUtils::get_struct_sym_from_struct_expr(x.m_args[i])

I then runned and it solves the problem!!

certik:
Perfect. Now add a test into integration_tests that fails in main, but works now.

certik:
Register it in cmake, and remove the output (instead check correct results from the code itself via `if (.not. something) error stop`.

Once done, please click on the "Ready for review" button.

certik:
The test case that you added works for me in main already.

Which MRE are you trying to fix in this PR? If the one in https://github.com/lfortran/lfortran/issues/8763, then you need to ensure it still triggers the bug in main, and that it is fixed by your PR.

Likely it might be related to print, so keep the print statements there, but also use the "if" for checks.

opixdown:
Thank you @certik for the detailed guidance. I am still new to contributing, so I appreciate your patience and support. I will update the test to properly reproduce the issue and push the changes.

assem2002:
I think it's good to merge.

src/bin/lfortran.cpp
```diff
@@ -2757,4 +2757,4 @@ int main(int argc, char *argv[])
         return 1;
     }
     return 0;
-}
+}
```
certik, line 2760: Next time just remove all these whitespace changes, only commit actual changes in a PR.

certik (approved):
I think that looks good, I tested that the code actually fails in main, and it works here.

certik:
The exhaustive checks fail, so I had to revert it in https://github.com/lfortran/lfortran/pull/8779.

@opixdown go ahead and open a new PR and we'll run exhaustive checks. You need to fix this failure:

https://github.com/lfortran/lfortran/actions/runs/19053321886/job/54418251645

```
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/../libasr/asr.h
function down_cast(), line number 41 at 
is_a<T>(*f)
```

opixdown:
Thanks for the help! I understand the PR was reverted.
I’ll look into the failing test and open a new PR soon.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14413,19 +14413,33 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                    // Get struct type from ASR, as LLVM 15+ Doesn't support direct 
-                    // extraction from tmp as getPointerElementType()
-                    ASR::symbol_t* struct_sym = ASRUtils::symbol_get_past_external(
-                        ASRUtils::get_struct_sym_from_struct_expr(x.m_args[i]));
-                    ASR::Struct_t* struct_t = ASR::down_cast<ASR::Struct_t>(struct_sym);
-                    llvm::Type* polymorphic_struct_type = llvm_utils->getClassType(struct_t);
-                    llvm::Value* data_ptr = llvm_utils->create_gep2(
-                        polymorphic_struct_type, tmp, 1);
-                    tmp = llvm_utils->CreateLoad2(llvm_utils->i8_ptr, data_ptr);
+                    ASR::is_a<ASR::Var_t>(*x.m_args[i])) {
```
jinangshah21, line 14416: Here just add one more condition `ASR::is_a<ASR::StructType_t>(*extract_type(expr_type(x.m_args[i])))`, this will make things work

PR #8775, opened 2025-11-01
fix: cast assumed rank arrays to descriptor when used as arguments to intrinsic functions

HarshilShah1804:
## Changes Made
- Add `is_assumed_rank` flag in the ASR Node of Array. This is done as assumed rank arrays are casted to descriptors with empty dimensions. So, to decide whether an array originally an  assumed rank array in later passes (for eg. to stop pass_array_by_data), we need some flag.
- Use a map based approach and insert cast node whenever an argument for any intrinsic function is an assumed rank array

Now, we convert the assumed rank array node to an array physical cast node, which casts assumed rank array to a descriptor array of rank same as its select rank block with empty dims
```clojure
(IntrinsicArrayFunction
    MinLoc
    [(ArrayPhysicalCast
        (Var 3 x)
        AssumedRankArray
        DescriptorArray
        (Array
            (Integer 4)
            [(()
            ())]
            DescriptorArray
            .true.
        )
        ()
    )
```


certik (approved):
I think that this is fine, thanks!

PR #8776, opened 2025-11-02
Tests : Uncomment openmp tests

### LINKED ISSUE #8772: tests : Fix openmp after `--new-classes`
```
 The following tests FAILED:
Errors while running CTest
	 66 - openmp_55 (SEGFAULT)
	 67 - openmp_56 (SEGFAULT)
	 68 - openmp_57 (SEGFAULT)
	 69 - openmp_58 (SEGFAULT)
```
https://github.com/lfortran/lfortran/actions/runs/18978527943/job/54204658224?pr=8769

assem2002:
closes #8772


certik:
Thanks for figuring this out!

Was the issue that this bug was always there, but didn't surface until now?

assem2002:
No it's due to new classes. With new classes, llvm structs has increased by 8 bytes.

src/libasr/pass/openmp.cpp
```diff
@@ -3222,6 +3222,7 @@ class ParallelRegionVisitor :
 
         int64_t compute_task_data_size(const ASR::symbol_t* task_data_struct_sym) {
             int64_t total_size = 0;
+            const int STRUCT_VTABLE_SIZE = 8;
```
certik, line 3225: Is there a way to get this number out of the LLVM backend and ensure it is correct? If we change this in the future, so that we are forced to update it here as well.

certik:

I made this https://github.com/lfortran/lfortran/issues/8778.

assem2002:
Yeah that's for sure needs some assertion, or even some delegation to the backend to fill in that piece of information (specially if we have more similar cases).

certik:
Great idea --- let's just move this constant to the backend as a global variable/constant and put some assert in the backend also and we are done.

Issue #8771, opened 2025-10-31
Run `select_type_13.f90`  with new classes and `--fast`


jinangshah21:
Remove `NO_FAST` Label from `select_type_13` integration test (https://github.com/lfortran/lfortran/pull/8769#discussion_r2481707843)

PR #8782, opened 2025-11-04
Remove NO_FAST label

PrakharGoyal28:
#8771 Fixes
Removes NO_FAST label from select_type_13 test in integration_tests/CMakeLists.txt since it works correctly with --fast now.


integration_tests/CMakeLists.txt
```diff
@@ -689,7 +689,7 @@ RUN(NAME arrays_46 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_47 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME arrays_50 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES arrays_50_mod.f90)
+RUN(NAME arrays_50 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
jinangshah21: This change is unrelated, remove it.

PrakharGoyal28: Done

assem2002:
I think you already got the conda environment set -- probably you're using LLVM11 (which is our standard), but the CI is failing with LLVM >=17. Can you use one of those version to test the integration test? We can help you investigating.
Notice : llvm >=17 versions fail due to main core reason which is opaque pointers and how we handle them in the codebase.

PR #8787, opened 2025-11-04
Fix #6052: nested_vars pass assigning to intent(in) variables

### LINKED ISSUE #6052: assigning to intent(in) variable by nested vars pass
```fortran
subroutine add(x, y)
    implicit none
    real, intent(in) :: x, y
    print *, x + y
    call add2()
    contains
        subroutine add2()
            implicit none
            print *, x + y + 2
        end subroutine
end subroutine

program main
    implicit none
    call add(1.0, 2.0)
end program
```

```console
% lfortran c.f90
ASR verify pass error: Assignment target `x` with intent `IN` not allowed
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2738
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2679
    err = compile_src_to_object_file(arg_file, tmp_o, false,
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 1052
    res = fe.get_llvm3(*asr, lpm, diagnostics, infile);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 382
    compiler_options, run_fn, infile);
  File "/Users/pranavchiku/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 10496
  File "/Users/pranavchiku/repos/lfortran/src/libasr/pass/pass_manager.h", line 403
    apply_passes(al, asr, _passes_with_experimental_simplifier, pass_options, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/libasr/pass/pass_manager.h", line 184
    throw LCompilersException("Verify failed in the pass: "
LCompilersException: Verify failed in the pass: nested_vars
```

Bumped to this while debugging #5957

PrakharGoyal28:
Closes #6052
Converts variables with intent(in) to intent(inout) before creating synchronization assignments in the nested_vars pass.
This prevents ASR verification errors like
Assignment target 'x' with intent IN not allowed

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7375,43 +7375,57 @@ public:
     void generate_binop_check(ASR::expr_t* left, ASR::expr_t* right) {
         ASR::ttype_t *type32 = ASRUtils::TYPE(ASR::make_Integer_t(al, left->base.loc, 4));
 
-        ASR::expr_t* left_size = ASRUtils::EXPR(ASR::make_ArraySize_t(al, left->base.loc,
-            left, nullptr, type32, nullptr));
-        visit_expr(*left_size);
-        llvm::Value* left_llvm_size = tmp;
+        ASR::ttype_t* left_type = ASRUtils::expr_type(left);
+        ASR::dimension_t* m_dims = nullptr;
+        size_t rank = ASRUtils::extract_dimensions_from_ttype(left_type, m_dims);
 
-        ASR::expr_t* right_size = ASRUtils::EXPR(ASR::make_ArraySize_t(al, right->base.loc,
-            right, nullptr, type32, nullptr));
-        visit_expr(*right_size);
-        llvm::Value* right_llvm_size = tmp;
+        for (size_t dim = 0; dim < rank; dim++) {
+            ASR::expr_t* dim_asr = ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, m_dims[dim].loc, dim + 1,
+                        ASRUtils::TYPE(ASR::make_Integer_t(al, m_dims[dim].loc, 4))));
 
-        ASR::Variable_t* left_var = ASRUtils::expr_to_variable_or_null(left);
-        ASR::Variable_t* right_var = ASRUtils::expr_to_variable_or_null(right);
+            ASR::expr_t* left_size = ASRUtils::EXPR(ASR::make_ArraySize_t(al, left->base.loc,
+                left, dim_asr, type32, nullptr));
+            visit_expr(*left_size);
+            llvm::Value* left_llvm_size = tmp;
 
-        if (left_var && right_var) {
-            llvm::Value *left_name = LCompilers::create_global_string_ptr(context, *module, *builder, left_var->m_name);
-            llvm::Value *right_name = LCompilers::create_global_string_ptr(context, *module, *builder, right_var->m_name);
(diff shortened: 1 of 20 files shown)


src/libasr/pass/nested_vars.cpp
```diff
@@ -791,6 +791,17 @@ private :
                         } else if (is_procedure_variable) {
                             body.push_back(al, ASRUtils::STMT(ASR::make_Associate_t(al, t->base.loc, target, val)));
                         } else {
+                            if (ASR::is_a<ASR::Var_t>(*val)) {
```
AnshMehta1: I don't think this change is required anymore for fixing this issue as the MRE works correctly with latest main.

tests/errors/a.f90
```diff
@@ -0,0 +1,16 @@
+subroutine add(x, y)
```
AnshMehta1: This test should be added as an integration test with error stop conditions rather than adding it in errors directory.

integration_tests/internal_subr_host_vars_01.f90
```diff
@@ -0,0 +1,26 @@
+! Verify host association for internal subroutines and enforce via error stop
+subroutine add(x, y)
+    implicit none
+    real, intent(in) :: x, y
+    real :: s
+
+    s = x + y
+    if (abs(s - (x + y)) > 1.0e-6) error stop "host add: s mismatch"
```
AnshMehta1: Test must be intended to validate the value of `x + y`.  Add something like `if ( s /= 3.0) error stop`. Don't add any message after error, just a normal error stop is enough.

```suggestion
    if (s /= 3.0) error stop

File: integration_tests/internal_subr_host_vars_01.f90 [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+
+contains
+
+    subroutine add2()
+        implicit none
+        real :: t
+        ! add2 should be able to see x, y from the host
+        t = x + y + 2.0
+        if (abs(t - (x + y + 2.0)) > 1.0e-6) error stop "host add2: t mismatch"
```
AnshMehta1:
```suggestion
        if (t /= 5.0) error stop

File: integration_tests/internal_subr_host_vars_01.f90 (hunk 3 above)
Comment (PrakharGoyal28) [reply integration_tests/internal_subr_host_vars_01.f90]: Got it

File: integration_tests/internal_subr_host_vars_01.f90 [hunk 5]
```diff
@@ -0,0 +1,26 @@
...
+        real :: t
+        ! add2 should be able to see x, y from the host
+        t = x + y + 2.0
+        if (t /= 5.0) error stop
+    end subroutine add2
+
+end subroutine add
+
+
+program main
```
AnshMehta1: Also, always use program name as file name .
```suggestion
program internal_subr_host_vars_01

File: integration_tests/internal_subr_host_vars_01.f90 [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+        if (t /= 5.0) error stop
+    end subroutine add2
+
+end subroutine add
+
+
+program internal_subr_host_vars_01
+    implicit none
+    call add(1.0, 2.0)
+end program main
```
AnshMehta1: This should not be main as program name is not main.
```suggestion 
end program internal_subr_host_vars_01

AnshMehta1 (approved):
Looks fine to me. Thanks

Issue #6776, opened 2025-03-29
Error : Report error for not matching string array types


assem2002:
MRE 
```.f90
program test
  character(2):: x(200)
  character(:),allocatable:: y(:)
  call ss(x)   ! <<<<<<<<<<<<<<<<<< ERROR
  ! call ss(y) ! <<<<<<<<<<<<<<<<<< This works
  contains 
  subroutine ss (xx)
    character(:), allocatable :: xx(:)  
    xx(1) = "AB"
  end subroutine 
end program test
```
```console
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store [3 x i8]* %9, i8*** %10, align 8
 i8**Call parameter type does not match function signature!
  %array_descriptor = alloca %array, align 8
 %array**  call void @ss(%array* %array_descriptor)



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt7.f90 
tt7.f90:16:10:

   16 |   call ss(x)   ! <<<<<<<<<<<<<<<<<< ERROR
      |          1
Error: Actual argument at (1) to allocatable or pointer dummy argument ‘xx’ must have a deferred length type parameter if and only if the dummy has one
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ 

```

galactus3050:
hi @assem2002, 
I would like to work on this issue, can you guide me a little bit please.

assem2002:
Hi @galactus3050. Thanks for your interest in this issue. 
We're now doing some string-related refactoring, so you may will find what I'm about to suggest got changed, but here is the appropriate way to fix that for now.

Here is the ASR node for strings `String(int kind, int len, expr? len_expr, string_physical_type physical_type)`,    `int len` member if set to `-2` it means the length is deferred. You can them check the passed `ASR::Var` node in `ASR::Functioncall` against the actuall `ASR::Variable` type of the function `ss` if `m_len` aren't of the same value you can raise the same error raised by gfortran.

Reference : 
- Check function `inline bool check_equal_type(ASR::ttype_t* x, ASR::ttype_t* y, bool check_for_dimensions)` at file `asr_utils.h` https://github.com/lfortran/lfortran/blob/5026aeca39330a5f862c45111744d98be85a5e7b/src/libasr/asr_utils.h#L3994-L3995.
[...]

PR #8788, opened 2025-11-05
Report error for mismatched deferred-length string arrays (#6776)

swastim01:
### **Title**

Report error for mismatched deferred-length string arrays (#6776)

### **Description**

Adds a semantic check in `types_equal()` (`src/libasr/asr_utils.h`) to ensure that a deferred-length character array (`character(:)`) isn’t passed to or from a fixed-length array (`character(n)`).
Uses `ASRUtils::is_deferredLength_string()` to detect and reject mismatched types, producing a clear compile-time error instead of an LLVM failure.

### **Fixes:**
Fixes:  #6776

### **Example**

```fortran
program test
  character(2):: x(200)
  call ss(x)  ! now correctly reports type mismatch
contains
  subroutine ss(xx)
    character(:), allocatable :: xx(:)
  end subroutine
end program
```

### **Summary**

[x] Adds deferred-length type check
[x] Prevents string array mismatch crashes
[x] Matches gfortran behavior

### Notes:

The CI failures appear to be due to the new deferred-length string check in `types_equal()`.
For example, `integration_tests/modules_15b.f90` now correctly reports a semantic error when mixing `character(n)` with `character(:)`.
This behavior aligns with standard Fortran semantics (as implemented by gfortran).
Please confirm if you'd like me to update or mark the affected tests as expected failures.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1843,6 +1843,7 @@ RUN(NAME string_79 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_80 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_81 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_82 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME string_83 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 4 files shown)


assem2002:
Great work!
This PR enforces the argument and the parameter to be both of `deferredLength`. The actual case is that we only should raise error if parameter is of `deferredLength` and the argument is not. If the opposite occurs we compile just fine.
 
 see this similar but valid example :
 ```.f90
 program test
  character(:), allocatable:: x(:)
  call ss(x)
contains
  subroutine ss(xx)
    character(10) :: xx(100)
  end subroutine
end program
```

we need to handle that as the CI is failing due to that.

The check in function `types_equal()` is good place to add the check, but apparently the error message isn't clarifying enough (it says `Type mismatch in argument at argument (1); passed string to string.`) . We need to add that check before the following check and raise an appropriate error message about the deferredLength thing https://github.com/lfortran/lfortran/blob/4d8cfa4c804049fe7b51df7a852e5980ebbbeab7/src/lfortran/semantics/ast_common_visitor.h#L6761-L6768

assem2002:
Will draft it till it gets ready. Feel free to reach out for any help.

PR #8791, opened 2025-11-05
fix: allocate lhs properly in `move_alloc` and while reallocating scalar `class()` variable

jinangshah21:


src/libasr/asr_builder.h
```diff
@@ -1021,16 +1021,17 @@ class ASRBuilder {
         }
     }
 
-    ASR::stmt_t *Allocate(ASR::expr_t *m_a, Vec<ASR::dimension_t> dims) {
+    ASR::stmt_t *Allocate(ASR::expr_t *m_a, Vec<ASR::dimension_t> dims, ASR::symbol_t* sym_subclass = nullptr) {
```
certik, line 1024: One idea is to create a new function, so that we don't have default values, which are problematic for adding more arguments, etc.

certik (approved):

src/libasr/asr_builder.h
certik: This can be done later, no need now.

Issue #8728, opened 2025-10-16
Missing bounds check in `e = f(a)`


certik:
Found a case where we don't throw a bounds check error. gfortran and flang-new both throw errors.

```fortran
program my
    integer, allocatable :: a(:)
    integer, allocatable :: e(:)
    allocate(a(3))
    allocate(e(3))

    a = 3
    e = f(a)

    print *, e

    contains
        function f(x) result(y)
            integer, intent(in) :: x(:)
            integer :: y(size(x) - 1)
            y = 2
        end function
end program
```

We convert `e = f(a)` to `call f(a, e)`. The second argument of function `f` expects an array of size `size(x)-1`, where x is the first argument. Size of a and e is 3. So `f` expects an array of size 2 as the second argument. We are passing e, which is an array of size 3. We don't throw any error in this case.

This is done because we don't have strict bounds checking, we allow larger arrays to be passed into a subroutine which expects shorter arrays.

Issue #8727, opened 2025-10-16
Add an option for strict bounds checking


certik:
Add an option for strict bounds checking: the passed parameters must be exactly equal to arguments, not just larger than or equal. This is not standards compliant, so must be configurable. We might consider having it by default, but the error message must say to use the given option to enable "greater than" checking.

PR #8804, opened 2025-11-07
feat: add strict bounds checking 

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8727, Fix https://github.com/lfortran/lfortran/issues/8728

1. Strict bounds checking means the size of the array passed in a subroutine must exactly equal the expected size. The standard allows us to pass an array whose size is larger than expected.
2. #8728 is fixed by using strict bounds checking on the return argument when `SubroutineCall` was converted from a `FunctionCall`.

### DIFF
--- a/doc/src/asr/asr_nodes/statement_nodes/subroutinecall.md
+++ b/doc/src/asr/asr_nodes/statement_nodes/subroutinecall.md
@@ -8,7 +8,7 @@ function (`name`) member, a `stmt` node.
 ### Syntax
 
 ```fortran
-SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, bool was_functioncall)
+SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, bool strict_bounds_checking)
 ```
 
 ### Arguments
@@ -17,7 +17,7 @@ SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, boo
 `original_name` contains name present in program unit.
 `args` contains arguments passed to subroutine call.
 `dt` contains expression for variable name, array name, an aseterist e.t.c.
-`was_functioncall` indicates if this SubroutineCall was a FunctionCall before getting converted by subroutine_from_function pass, this is used for bounds checking.
+`strict_bounds_checking` indicates if this SubroutineCall was a FunctionCall before getting converted by subroutine_from_function pass, if it was then we need to do strict_bounds_checking for the return argument.
 
 ### Return values
 
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -61,7 +61,7 @@ stmt
     | Select(identifier? name, expr test, case_stmt* body, stmt* default, bool enable_fall_through)
     | Stop(expr? code)
     | Assert(expr test, expr? msg)
-    | SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, bool was_functioncall)
+    | SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, bool strict_bounds_checking)
     | IntrinsicImpureSubroutine(int sub_intrinsic_id, expr* args, int overload_id)
     | Where(expr test, stmt* body, stmt* orelse)
     | WhileLoop(identifier? name, expr test, stmt* body, stmt* orelse)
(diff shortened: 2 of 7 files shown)


xaerru:
I can't view the logs of the macos fail.

src/libasr/ASR.asdl
```diff
@@ -61,7 +61,7 @@ stmt
     | Select(identifier? name, expr test, case_stmt* body, stmt* default, bool enable_fall_through)
     | Stop(expr? code)
     | Assert(expr test, expr? msg)
-    | SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt)
+    | SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, bool was_functioncall)
```
certik: I am not sure about this. How is bounds checking different if it was a function call before?

It seems the right fix is to insert some new nodes for extra checks. When we have a subroutine, it should not matter if it was a function call before.

xaerru: I've added some comments in both programs below:
```fortran
program my
    integer, allocatable :: a(:)
    integer, allocatable :: e(:)
    allocate(a(3))
    allocate(e(3))

    a = 3
    ! size(e) = 3
    ! size(f(a)) = 2
    ! This must throw an error. gfortran also throws a runtime error here
    ! But after conversion to subroutine it becomes allowed
    e = f(a)

    print *, e

    contains
        function f(x) result(y)
            integer, intent(in) :: x(:)
            integer :: y(size(x) - 1)
            y = 2
        end function
end program
```
After `subroutine_from_function`:
```fortran
! Fortran code after applying the pass: subroutine_from_function
program my
implicit none
integer(4), dimension(:), allocatable :: a
integer(4), dimension(:), allocatable :: e
allocate(a(3))
allocate(e(3))
do __libasr_index_0_ = lbound(a, 1), ubound(a, 1)
    a(__libasr_index_0_) = 3
end do
! size(e) == 3
! Expected size of argument y in subroutine f is size(a) - 1 = 2
! When we check arguments with expected size in the subroutine we can catch this error
! But only if we do strict bounds checking
! e is of size 3 and f expects an array of size 2, this is allowed by the standard
call f(a, e)
print *, e

contains

subroutine f(x, y)
    integer(4), dimension(:), intent(in) :: x
    integer(4), dimension(size(x) - 1), intent(out) :: y
    do __libasr_index_0_ = lbound(y, 1), ubound(y, 1)
        y(__libasr_index_0_) = 2
    end do
end subroutine f

end program my
```

certik: Got it. In this case, I think all we want is to enforce strict `M == N` bounds checking, not just `M >= N` bounds checking. So maybe let's rename `bool was_functioncall` to `bool strict_bounds_checking`. If `true`, the bounds checker is supposed to do strict bounds checking for this. If `false`, it can do `M >= N` bounds checking.

certik:
The error is unrelated (https://github.com/lfortran/lfortran/issues/8610), but we really need to fix it, it's been failing very often.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -184,6 +184,7 @@ namespace LCompilers::CommandLineInterface {
         app.add_flag("--stack-arrays", compiler_options.stack_arrays, "Allocate memory for arrays on stack")->group(group_miscellaneous_options);
         app.add_flag("--array-bounds-checking", compiler_options.po.bounds_checking, "Enables runtime array bounds checking")->group(group_miscellaneous_options);
         app.add_flag("--no-array-bounds-checking", disable_bounds_checking, "Disables runtime array bounds checking")->group(group_miscellaneous_options);
+        app.add_flag("--strict-array-bounds-checking", compiler_options.po.strict_bounds_checking, "Enables strict runtime array bounds checking: Array passed into subroutine must exactly match the expected size")->group(group_miscellaneous_options);
```
certik, line 187: After this is in, let's test it out. But likely we want to make it the default, and add `--relaxed-array-bounds-checking` instead later. But in this PR this is good.

certik:
I am not sure about this failure: https://github.com/lfortran/lfortran/issues/8808, but likely unrelated given it only happens in LLVM 14.

doc/src/asr/asr_nodes/statement_nodes/subroutinecall.md
```diff
@@ -17,7 +17,7 @@ SubroutineCall(symbol name, symbol? original_name, call_arg* args, expr? dt, boo
 `original_name` contains name present in program unit.
 `args` contains arguments passed to subroutine call.
 `dt` contains expression for variable name, array name, an aseterist e.t.c.
-`was_functioncall` indicates if this SubroutineCall was a FunctionCall before getting converted by subroutine_from_function pass, this is used for bounds checking.
+`strict_bounds_checking` indicates if this SubroutineCall was a FunctionCall before getting converted by subroutine_from_function pass, if it was then we need to do strict_bounds_checking for the return argument.
```
certik, line 20: I see, it's only for the return argument, not all arguments. The design is still not clean, but at least it's a start, we can iterate on this later.

certik (approved):
I think this is good to merge as is, if the last LLVM 14 test passes.

Issue #8655, opened 2025-10-02
Give Debug time error `var` not allocated


certik:
```fortran
program dealloc_05
    implicit none

    type :: base
        integer :: x
    end type

    type, extends(base) :: derived
        real :: r
    end type

    class(base), allocatable :: var

    allocate(derived :: var)

    select type(var)
        type is (derived)
        class default
            error stop
    end select

    deallocate(var)

    print *, allocated(var)
    select type(var) ! <--- give an error here, `var` is not allocated
        type is (base)
        class default
            error stop
    end select

end program dealloc_05
```

certik:
@jinangshah21 based on https://github.com/lfortran/lfortran/issues/7512#issuecomment-3362162271, it looks like this:
```fortran
    deallocate(var)

    print *, allocated(var)
    select type(var)
        type is (base)
        class default
            error stop
    end select
```
Should be allowed, and the type of `var` is `base`. But this should not be allowed:
```fortran
    deallocate(var)

    print *, allocated(var)
    select type(var)
        type is (base)
            print *, var%x
        class default
            error stop
    end select
```
Since this is using an unallocated variable.

kmr-srbh:
Yes. In the second code example, the other compilers throw an error at runtime, which we can too. Since the allocation is a runtime event, the code is semantically correct I think.

certik:
@jinangshah21  you closed this as part of https://github.com/lfortran/lfortran/pull/8729, but in order to close it, we need tests for both cases in https://github.com/lfortran/lfortran/issues/8655#issuecomment-3362695677, the first one to work and the second one to fail with a proper runtime error message.

PR #8806, opened 2025-11-07
fix: throw runtime error if a member of an unallocated variable is accessed

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8655


certik:
Important: do these only happen when bounds checking is enabled? Or all the time.

If all the time, we need to go over all bounds checking and ensure these extra runtime LLVM operations are only emitted when bounds checking is on.

xaerru:
I forgot the check in this PR. Added it.

Everywhere else we do check the compiler option `compiler_options.po.bounds_checking` before generating extra llvm.

PR #8809, opened 2025-11-08
fix: cast assumed rank arrays to descriptor correctly before passing to print statements

HarshilShah1804:
## Summary of what is to be done
- We have a map containing the current rank of the assumed rank array (while visiting statements present inside a select rank block) names `assumed_rank_arrays`
- We have to identify all the statements that can have arrays as arguments and insert an `ArrayPhysicalCast` casting assumed rank array to a descriptor array by using rank from the map.
- If the rank is not present in the map, we will throw error as no operations are allowed on assumed rank arrays.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -752,6 +752,7 @@ RUN(NAME assumed_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME global_allocatable_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME global_allocatable_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/select_rank_04.f90
+++ b/integration_tests/select_rank_04.f90
@@ -1,0 +1,23 @@
+program select_rank_04
+    implicit none
+
+    integer :: a(2) = [4, 2]
+
+    call check(a)
+
+contains
+
+    subroutine check(x)
+        integer, intent(in) :: x(..)
+        integer :: l1(1)
+        integer :: l2(2)
+    
+        select rank(x)
+        rank (1)
+            print *, x
+        rank (2)
+            print *, x
+        end select
+    end subroutine check
+
+end program select_rank_04
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -25,10 +25,10 @@ contains
         y = y + 1
     end subroutine intent_inout_test
 
-
-
-
-
+    subroutine assumed_rank(x)
+        integer, intent(in) :: x(..)
+        print *, x
+    end subroutine assumed_rank
 
 
 
--- a/tests/reference/asr-continue_compilation_3-435a232.json
+++ b/tests/reference/asr-continue_compilation_3-435a232.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_3-435a232",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_3.f90",
-    "infile_hash": "626e427aa8377506e049aa3282766ce16b7587f15a76d70eb125f377",
+    "infile_hash": "f5ad8cb455e9394b4de123ee0db1ab73cadba1ee4dea8cffafc292af",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_3-435a232.stderr",
-    "stderr_hash": "e815b1d129c484727e7509ae9ee04a2a89c12dde33fe011f236c0d32",
+    "stderr_hash": "f0012457706685d68645d4645f4f5a306ec93d179181eec426a9649a",
     "returncode": 1
 }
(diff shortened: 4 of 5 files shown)


src/libasr/asr_verify.cpp
```diff
@@ -1267,7 +1271,7 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
     }
 
     void visit_dimension(const dimension_t &x) {
-        if (_inside_array_physical_cast_type && !_inside_call) {
+        if (_inside_array_physical_cast_type && !_inside_call && !_processing_assumed_rank_array) {
```
HarshilShah1804, line 1276: @certik can you please provide your feedback on whether relaxing this verify condition is appropriate or not as casting assumed rank arrays to empty dimension descriptors would always have empty dimensions.

certik: Can you show ASR output for this case? Let's have a look.

I thought DescriptorArray has dimensions but they are each `:`, so nullptr. However this check in master would possibly fail for that case? Or when is this visit_dimension called exactly?

Finally, what Fortran code example are you fixing in this PR that fails in master but now works?

HarshilShah1804: For the MRE:
```fortran
program select_rank_03
    implicit none

    integer :: a(2) = [4, 2]

    call check(a)

contains

    subroutine check(x)
        integer, intent(in) :: x(..)
        integer :: l1(1)
        integer :: l2(2)
    
        select rank(x)
        rank (1)
            print *, x
        rank (2)
            print *, x
        end select
    end subroutine check

end program select_rank_03
```

This is the generated ASR (specifically for the select rank block:
```clojure
~select_rank_block_
    [(Print
        (StringFormat
            ()
            [(ArrayPhysicalCast
                (Var 3 x)
                AssumedRankArray
                DescriptorArray
                (Array
                    (Integer 4)
                    [(()
                    ())]
                    DescriptorArray
                )
                ()
            )]
            FormatFortran
            (Allocatable
            )
            ()
        )
    )]
```




> I thought DescriptorArray has dimensions but they are each `:`, so nullptr. However this check in master would possibly fail for that case? Or when is this visit_dimension called exactly?

Yes, you are correct. But this part of the `visit_dimension` function is called only when we are visiting the `type` variable of the `ArrayPhysicalCast` node. So, I guess till now we expected that the dimensions of the `type` variable of `ArrayPhysicalCast` node to not be null.

I guess this is the correct approach for handling assumed rank arrays as maybe in all the cases that we would have handled till now, we might not have used casting anywhere when the dimensions are not available.

certik: So add the above example into tests.

src/libasr/asr_verify.cpp
```diff
@@ -1046,6 +1047,9 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
                 + " " + std::to_string(ASRUtils::extract_physical_type(ASRUtils::expr_type(x.m_arg))));
             bool _inside_array_physical_cast_type_copy = _inside_array_physical_cast_type;
             _inside_array_physical_cast_type = true;
+            if (x.m_old == ASR::array_physical_typeType::AssumedRankArray) {
+                _processing_assumed_rank_array = true;
+            }
```
certik, line 1053: You never set this to false --- when you set it to true, you need to set it to false in all places correctly. It's a bit fragile, but the current way is almost for sure incorrect.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4760,6 +4760,39 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            if (ASRUtils::is_assumed_rank_array(ASRUtils::expr_type(expr))) {
+                ASR::Var_t* v = ASR::down_cast<ASR::Var_t>(expr);
+                ASR::Variable_t *var = ASR::down_cast<ASR::Variable_t>(v->m_v);
+                std::string var_name = var->m_name;
+                if (assumed_rank_arrays.find(var_name) == assumed_rank_arrays.end()) {
+                    diag.add(diag::Diagnostic(
+                        "Assumed-rank arrays are not supported in print statements",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {expr->base.loc})
+                        }));
```
certik, line 4760: Add a test for this error. Just put it into tests/errors/continue_compilation_1.f90

certik:
Ok, it's probably fine then. Go ahead and fix the other things, then it should be good.

src/libasr/asr_verify.cpp
HarshilShah1804: Oh yes. That's correct. I'll change that.

src/lfortran/semantics/ast_body_visitor.cpp
HarshilShah1804: Ok.

src/libasr/asr_verify.cpp
HarshilShah1804: > You never set this to false --- when you set it to true, you need to set it to false in all places correctly. It's a bit fragile, but the current way is almost for sure incorrect.

Yes, I also think so, as we have already merged a PR doing this array physical cast but haven't encountered this error. I'll investigate it further.

HarshilShah1804:
Ok. I got the reason why we had to relax that verify condition ? Initially, all the cases that we had fixed for assumed rank array casting were related to function calls or subroutine calls where, we had a flag `_inside_call` set to true. So, it didn't checked the dimensions and we didn't get any error then. But for all other statements (like `print` in this case), it required the dimensions of the array to be not null but as we have handled the assumed rank arrays in the backend, I think we would need to relax this condition.

@certik Please provide your suggestions on this.

certik:
The verify handling is probably fine. In order to finish this PR, please fix:

* [x] https://github.com/lfortran/lfortran/pull/8809/files#r2508081950
* [x] https://github.com/lfortran/lfortran/pull/8809/files#r2508081006
* [x] https://github.com/lfortran/lfortran/pull/8809/files#r2508080540

HarshilShah1804:
The failures seem to be unrelated. I'll rerun those tests once the rest of them are completed.

src/libasr/asr_verify.cpp
certik: Ok, I think the logic is now correct, I don't know if `_processing_assumed_rank_array = _processing_assumed_rank_array_copy;` was there from the beginning or if you just added it, but I think that is probably correct.

HarshilShah1804: Yes, I just added it now.

certik:
Unfortunately today the CI is having some trouble, so we'll have to keep restarting the CI until it works.

HarshilShah1804:
> I think this is fine now.
> 
> However, this change that you did for Print we will now have to do for every single ASR node, which will be a lot of changes. One could extract these into some utility function. This is to be done later, right now let's get this PR to merge.

Yes, I also had that in mind. Will do that in an upcoming PR.

PR #8817, opened 2025-11-11
enh: make resolving struct symbols covers more cases

akramhany:
This is an enhancement I noticed that fixed an issue I encountered while compiling the `Julienne` package, but unfortunately I couldn't extract an MRE for the issue.

Now, the resolving of struct symbols totally depends on the `ASRUtils::get_struct_sym_from_struct_expr()` function, which covers all possible cases for a struct expression.


certik (approved):
If it passes tests, we can merge it. Removing code is always good.

Issue #8814, opened 2025-11-10
Handle string kind cast in a comparision expression


jinangshah21:
MRE:
```fortran
program class_77
   integer, parameter :: tfc = selected_char_kind('DEFAULT')
    character(1, tfc) :: hash = tfc_"a"
    print *, tfc_"a" == hash
end program class_77
```

Outputs:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran b.f90 && ./a.out
 T
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
F
```

HarshitaKalani:
```
% lfortran a.f90 
1
3
F
% cat a.f90
program main
    character :: hash = 1_"a"
    print *, len(hash)
    print *, len(1_"a")
    print *, 1_"a" == hash
end program main
```
This is an issue with parser I think. Do we support kind prefix for character type?

PR #8827, opened 2025-11-12
Parse prefix kind for character correctly

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8814


src/lfortran/parser/tokenizer.h
```diff
@@ -48,8 +48,15 @@ class Tokenizer
...
-        s.n = cur-tok-2;
+        unsigned char *start = tok;
+        unsigned char *end = cur;
+        while (start < end && *start != (unsigned char)ch) {
+            start++;
+        }
+        LCOMPILERS_ASSERT(start < end);
+        LCOMPILERS_ASSERT((end - start) >= 2);
+        s.p = (char*) start + 1;
+        s.n = end - start - 2;
```
certik, line 59: Is this logic skipping the prefix?

Are we saving the information into AST? We need to figure out how to store this in AST. Likely the right place to fix this is in the re2c-based tokenizer that recognizes a string without a prefix (already works) and a string with a prefix (in this case it will call a function to extract the prefix and the string, and store the prefix in AST).

HarshitaKalani: > Is this logic skipping the prefix?

Yes. 
You're right, we should be storing it in AST. I'll check that.

Issue #7988, opened 2025-07-06
Handle class pointer in source of `allocate`


jinangshah21:
MRE:
```fortran
program read_int_from_ptr
  type :: base
  end type
  class(base), allocatable :: x
  class(base), pointer :: y
  allocate(x, source=y)
end program read_int_from_ptr
```

LFortran Error:
```console
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store %base_polymorphic* %14, %base_polymorphic* %13, align 8
 %base_polymorphic = type { i64, %base* }
```

HarshitaKalani:
```console
 % gfortran a.f90 && ./a.out
In file 'a.f90', around line 6: Error allocating 15447069372904012797 bytes: Cannot allocate memory

Error termination. Backtrace:
#0  0x104392103
#1  0x104392cd7
#2  0x104392f53
#3  0x104170823
#4  0x10417089f
% lfortran a.f90           
% cat a.f90
program read_int_from_ptr
  type :: base
  end type
  class(base), allocatable :: x
  class(base), pointer :: y
  allocate(x, source=y)
end program read_int_from_ptr%                                    
```
This is the behaviour in main.

HarshitaKalani:
On allocating y, it works fine.
```console
% cat a.f90
program read_int_from_ptr
  type :: base
  end type
  class(base), allocatable :: x
  class(base), pointer :: y
  allocate(y)
  allocate(x, source=y)
end program read_int_from_ptr
% gfortran a.f90 && ./a.out
% lfortran a.f90  
```

PR #8830, opened 2025-11-12
test: add and register test for class pointer in source of allocate

HarshitaKalani:
closes: https://github.com/lfortran/lfortran/issues/7988

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1359,23 +1359,13 @@ public:
 
                         llvm::Value* target_struct_i8 = llvm_utils->CreateLoad2(
                             llvm_utils->i8_ptr, target_addr_as_i8pp);
+                        llvm::Type* target_struct_type = llvm_utils->get_type_from_ttype_t_util(
+                            curr_arg.m_a, curr_arg_m_a_type, module.get());
+                        llvm::Value* target_struct = builder->CreateBitCast(
+                            target_struct_i8, target_struct_type->getPointerTo());
 
-                        llvm::FunctionType* copy_fn_type = llvm_utils->struct_copy_functype;
-                        llvm::PointerType* copy_fn_ptr_type = llvm::PointerType::get(copy_fn_type, 0);
-                        llvm::PointerType* copy_fn_ptr_ptr_type = llvm::PointerType::get(copy_fn_ptr_type, 0);
-                        llvm::PointerType* copy_fn_ptr_ptr_ptr_type = llvm::PointerType::get(copy_fn_ptr_ptr_type, 0);
-
-                        llvm::Value* vtable_copy_ptr = builder->CreateBitCast(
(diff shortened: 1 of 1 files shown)


integration_tests/pointer_04.f90
```diff
@@ -0,0 +1,8 @@
+program pointer_04
+  type :: base
+  end type
+  class(base), allocatable :: x
+  class(base), pointer :: y
+  allocate(y)
+  allocate(x, source=y)
+end program pointer_04
```
jinangshah21, line 22: Can you put some member inside base and point y to some `type(base)` variable and check x's value after allocate?

certik: Good idea, that way we know it works.

HarshitaKalani:
I tried the following and realised this does not return correct output.
```console
% cat a.f90
program pointer_04
    implicit none

    type :: base
        integer :: v
    end type

    class(base), allocatable :: x
    class(base), pointer     :: y
    type(base), target       :: t
    t%v = 42
    y => t
    
    allocate(x, source=y)
    print *, "t%v =", t%v
    print *, "y%v =", y%v
    print *, "x%v =", x%v

    t%v = 100

    print *, "After modifying t:"
    print *, "t%v =", t%v
    print *, "y%v =", y%v
    print *, "x%v =", x%v   ! should remain 42
end program pointer_04

% lfortran a.f90
t%v =    42
y%v =    42
x%v =    0
After modifying t:
t%v =    100
y%v =    100
x%v =    0
% gfortran a.f90 && ./a.out
 t%v =          42
 y%v =          42
 x%v =          42
 After modifying t:
 t%v =         100
 y%v =         100
 x%v =          42
```

HarshitaKalani:
The test now gives correct answer.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1313,8 +1313,74 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            copy_fn_ptr_type, vtable_copy_ptr, 0);
+                        copy_fn_slot = llvm_utils->CreateLoad2(copy_fn_ptr_type, copy_fn_slot);
+
+                        llvm::Value* source_struct_i8 = builder->CreateBitCast(
+                            source_handle, llvm_utils->i8_ptr);
+                        builder->CreateCall(copy_fn_type, copy_fn_slot,
+                                            { source_struct_i8, target_struct_i8 });
+
+                        continue;
+                    }
```
jinangshah21, line 1371: This is fine, but if we can use `llvm_utils->deepcopy()` here, then it is good as it will not cause code duplication. 
Also if `curr_arg_m_a_type` is `c_struct` then we should not use runtime function copy function, which is implemented in `llvm_utils->deepcopy` I think.

HarshitaKalani: Does this look better now?

Issue #8850, opened 2025-11-14
calling `execute_line_comand` twice with different length string corrupts command


jinangshah21:
MRE:
```fortran
program test
    character(:), allocatable :: dir
    integer :: i
    integer :: stat
    character(len=:), allocatable :: redirect, total
    redirect = " >/dev/null 2>&1"
    dir = "fortran_scratch"
    total = "test -d "//dir//redirect
    call execute_command_line(total, exitstat=stat)
    dir = "fortran_scratch2"
    total = "test -d "//dir//redirect
    call execute_command_line(total, exitstat=stat)
end program test
```

LFortran Error:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
sh: 1: Syntax error: Bad fd number
```

HarshitaKalani:
This test is working for me.
```console
% cat a.f90
program test
    character(:), allocatable :: dir
    integer :: i
    integer :: stat
    character(len=:), allocatable :: redirect, total
    redirect = " >/dev/null 2>&1"
    dir = "fortran_scratch"
    total = "test -d "//dir//redirect
    call execute_command_line(total, exitstat=stat)
    dir = "fortran_scratch2"
    total = "test -d "//dir//redirect
    call execute_command_line(total, exitstat=stat)
end program test
% lfortran a.f90
% gfortran a.f90 && ./a.out
% lfortran --version
LFortran version: 0.58.0-25-g584c9a813-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin24.5.0
```

jinangshah21:
```fortran
program test
    character(:), allocatable :: dir
    integer :: i
    integer :: stat
    character(len=:), allocatable :: redirect, total
    redirect = " >/dev/null 2>&1"
    dir = "fortran_scratch2"
    total = "test -d "//dir//redirect
    call execute_command_line(total, exitstat=stat)
end program test
```

```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
Executed: test -d fortran_scratch2 >/dev/null 2>&1q

sh: 1: Syntax error: Bad fd number
```

Here you can see that actual string passed to `_lfortran_exec_command` c instrinsic has some garbage value appended at end, so it might be bad allocation or termination.

assem2002:
As Jinang pointed out, it's a termination problem.
[...]

PR #8852, opened 2025-11-15
test: add test for working execute_command_line

HarshitaKalani:
Towards: https://github.com/lfortran/lfortran/issues/8850


jinangshah21:
The original test in #8850 still fails for me locally on Linux, I guess there is some bad allocation or termination of string there, becasue I tried to print exact command that is passed to `_lfortran_exec_command` and I saw some random characters were append to original command passed.

certik:
I merged it, but kept #8850 open, since it still segfaults for @jinangshah21.

Issue #7868, opened 2025-06-23
Handle `struct` used in another struct which is declared before in module


jinangshah21:
MRE:
```fortran
module derived_types_67
    public :: fpm_registry_settings, fpm_global_settings
    type :: fpm_global_settings
        type(fpm_registry_settings), allocatable :: registry_settings
    end type
    type :: fpm_registry_settings
    end type
end module derived_types_67
```

LFortran Error:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
ASR verify pass error: The symbol table was found, but the symbol `sym` is not in it
 --> b.f90:5:9
  |
5 |         type(fpm_registry_settings), allocatable :: registry_settings
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here

ASR verify pass error: StructType::m_derived_type 'fpm_registry_settings' cannot point outside of its symbol table, owner: fpm_global_settings
 --> b.f90:5:9
  |
5 |         type(fpm_registry_settings), allocatable :: registry_settings
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here
```

HarshitaKalani:
This works for me:
```
 % cat a.f90   
module derived_types_67
    public :: fpm_registry_settings, fpm_global_settings
    type :: fpm_global_settings
        type(fpm_registry_settings), allocatable :: registry_settings
    end type
    type :: fpm_registry_settings
    end type
end module derived_types_67
program main
    use derived_types_67
    implicit none
    type(fpm_global_settings) :: global_settings
    allocate(global_settings%registry_settings)
end program main%                                                                                                                              
% lfortran a.f90
% gfortran a.f90 && ./a.out
```

HarshitaKalani:
This test fails with --std f23 yet

PR #8861, opened 2025-11-16
test: add test for struct used in another struct which is declared before in module

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/7868


HarshitaKalani:
Why would `1204 - associate_08_STD_F23 (SEGFAULT)` fail because of adding a test in this PR?

certik:
Looks like there are several issues here:

* Go ahead and create a new issue for the `associate_08_STD_F23` bug. Looks like we have some flaky tests / bugs.
* I am guessing `--std=f23` doesn't work with this test? If so, do not close the #7868, just say "towards". We need to fix the bug for `--std=f23`.

You can then rebase it, and enable all exhaustive tests too, just to be sure.

certik:
Would you mind please opening up an issue to fix the `--std=f23` problem?

PR #8867, opened 2025-11-17
ci: separate debug and release WASM builds with retention policy

ubaidsk:
Previously, all WASM builds were stored indefinitely in the wasm_builds repository, causing rapid growth toward GitHub's 1GB limit since every main branch merge triggers a build.

This change routes main branch builds (debug mode) to dev/ with a 10-build retention limit, while release tag builds (release mode) go to release/ with no limit. This keeps repository size manageable while preserving all production release builds for users.

### DIFF
--- a/ci/upload_lfortran_wasm.sh
+++ b/ci/upload_lfortran_wasm.sh
@@ -20,9 +20,9 @@ elif [[ ${git_ref:0:11} == "refs/tags/v" ]]; then
     echo "Release tag detected: using dest_dir=${dest_dir}"
 else
     # We are either on a non-main branch or tagged with a tag that does
-    # not start with v*. We skip the upload.
-    echo "Not a main branch, not tagged with v*, skipping upload..."
-    exit 0
+    # not start with v*. We run the script for testing but do not upload.
+    dest_dir="release"
+    echo "Not a main branch, not tagged with v*, using dest_dir=${dest_dir} for testing..."
 fi
 
 lfortran_version=$(<version)
@@ -64,6 +64,19 @@ git commit -m "${COMMIT_MESSAGE}"
 
 git show HEAD -p --stat
 dest_commit=$(git show HEAD -s --format=%H)
+
+if [[ ${git_ref} == "refs/heads/main" ]]; then
+    echo "The pipeline was triggered from the main branch"
+else
+    if [[ ${git_ref:0:11} == "refs/tags/v" ]]; then
+        echo "The pipeline was triggered from a tag 'v*'"
+    else
+        # We are either on a non-main branch, or tagged with a tag that does
+        # not start with v*. We skip the upload.
+        echo "Not a main branch, not tagged with v*, skipping..."
+        exit 0
+    fi
+fi
 
 set +x
 if [[ "${SSH_PRIVATE_KEY_WASM_BUILDS}" == "" ]]; then


ubaidsk:
The dev build corresponds to wasm builds for PR merge to main. We will only be storing dev builds for the past 10 PR merges.

The release build corresponds to wasm builds for actual lfortran github releases. We will be storing unlimited of these. Since new releases are not that frequent, the size of the wasm_builds repo should be manageable.

ubaidsk:
Post this PR merge, I can also go ahead and tidy up the wasm_builds repo. Currently it stores like ~140 release builds. We can reduce this down to ~25, ensuring the version we use at https://github.com/lfortran/lcompilers_frontend still exists and the past 3 actual release builds also exist. Let me know if this sounds good.

certik:
Awesome, thanks for this!

What is the advantage of building in Debug mode --- why not always build it in Release mode?

We don't want to only build in Release mode when a commit gets tagged, because it can fail. Rather, we need to build exactly how we will build it when tagged, but for every commit. Then we know for sure that it will work. So I would just build in Release mode, as it was before, unless there is a good reason to also build in Debug mode, but then we need to test both on every commit.

ubaidsk:
Cool, updated. My only reasoning was that since Release builds already happen for actual releases, a Debug build felt suitable for PR merges to main. If we see any failures like wrong compilation or some bug in the compiler for the release built, we would have a debug build ready for debugging the error.


Your approach makes sense, though — building both would ideally require running Release and Debug for every PR, which is inefficient.

ubaidsk:
Now this will always build in release mode. But the builds for PR merges to main will be saved in the `debug` directory and builds for lfortran releases will be saved in the `release` directory in the wasm_builds repository.

certik:
Perfect. This removes old dev binaries. However, in git when you just remove files, they still stay in the history, so this will not actually fix the git repository blow up, correct?

ubaidsk:
Thanks, good catch. So it seems like we can just keep updating/amending the top most commit for wasm_builds repo. Does this approach sound fine or are there better approaches to avoid increase in git history size?

ubaidsk:
Maybe what we can also do is always maintain just a single orphaned commit in the wasm_builds repository.

Everytime we would like to push to wasm_builds, we do:

```shell
rm -rf .git
git init
# copy the necessary files
git add .
git commit -m "Deploy"
git branch -M main
git remote add origin <ssh-url>
git push --force origin main
```

This should overwrite the commit in remote without storing any history. We always only have a single reachable commit.

The drawback I can see is GitHub might still temporarily store previous objects in the remote repo until its own garbage collection runs (which it seems could days/weeks).

certik:
I think the force push is the way to go.

ubaidsk:
Thanks, just to confirm — when you say “force push is the way to go,” were you referring to the approach where we keep amending the latest commit and force-pushing it, or the approach where we recreate the repo each time so there’s always exactly one orphaned commit?

Both require a git push --force, but the workflow is slightly different, so I wanted to make sure I’m following the one you had in mind.

certik:
I see --- good idea about the orphaned commit (https://github.com/lfortran/lfortran/pull/8867#issuecomment-3544660997). When we do something like that, how long would such a commit stay available, before it gets removed by github?

I guess the idea is similar to pushing branches with the latest code, as well as always overwriting the main branch. So when the next commit overwrites the main branch, we still have a copy of the older commit in a separate branch. So we'll have one branch per commit (plus the latest commit always in main). So the main branch is always small. Then we can keep, say, 10 branches alive and remove older branches to keep the repository small.

I think that's similar to the idea of just force pushing into the main branch. When we recreate the whole repository, who is that idea different?

I don't mind any of these solutions. Whichever you think is the best, most robust, go for it. I think we need to keep all the released versions, and we should be putting up on dev.lfortran.org the released version. And then we just need the latest main, for quick fixes and testing that things are fixed. Once we fix any regressions, we just make an LFortran release and use the released version.

ubaidsk:
Thanks for the clarification — that helps a lot.

Regarding how long GitHub keeps unreachable commits after a force-push, I’m honestly not sure. From what I’ve read, GitHub runs its own garbage collection on its own schedule, and it can take a while, but I don’t think they document an exact timeframe.

I agree with you that this ends up being quite similar to simply amending the top commit and force-pushing it again. In both cases we only keep one reachable commit on main, and GitHub eventually cleans up the old objects.

On the question about “recreating the whole repository”, my understanding is that we wouldn’t really be recreating the repo itself — we rewrite the entire local history to contain just a single orphaned commit, and after force-pushing, the remote main branch also ends up with only that one commit.

Since you mentioned you’re fine with any of the approaches, I’m leaning toward the “single orphaned commit + force push” idea because it seems the simplest to maintain: the `wasm_builds` repo will always contain just one commit.

ci/upload_lfortran_wasm.sh
```diff
@@ -8,7 +8,22 @@ deploy_repo_push="git@github.com:lfortran/wasm_builds.git"
...
+    dest_dir="dev"
+    echo "Main branch detected: using dest_dir=${dest_dir}"
+elif [[ ${git_ref:0:11} == "refs/tags/v" ]]; then
+    dest_dir="release"
+    echo "Release tag detected: using dest_dir=${dest_dir}"
+else
+    # We are either on a non-main branch or tagged with a tag that does
+    # not start with v*. We skip the upload.
+    echo "Not a main branch, not tagged with v*, skipping upload..."
+    exit 0
```
certik: This means the rest of this script only runs on merge to main or on release. So in PRs, such as this one, we do not actually check that it works:
```
+ deploy_repo_pull=https://github.com/lfortran/wasm_builds.git
+ deploy_repo_push=git@github.com:lfortran/wasm_builds.git
++ git rev-parse --short dce692401e104dd013b00db745c9e4081574bff7
+ git_hash=dce692401
+ git_ref=refs/pull/8867/merge
+ [[ refs/pull/8867/merge == \r\e\f\s\/\h\e\a\d\s\/\m\a\i\n ]]
+ [[ refs/pull/8 == \r\e\f\s\/\t\a\g\s\/\v ]]
+ echo 'Not a main branch, not tagged with v*, skipping upload...'
+ exit 0
Not a main branch, not tagged with v*, skipping upload...
```
So if there is any mistake in the script below or the Python script that is called from this script, we do not know.

Instead, run the full script as before. But before setting up ssh keys (which only works on main), exit there, as it used to be.

ubaidsk:
Updated

certik:
I archived the old main at https://github.com/certik/wasm_builds/tree/old-main-3.

ubaidsk:
Looks like it failed, sent a fix PR here https://github.com/lfortran/lfortran/pull/8881.

Issue #8808, opened 2025-11-08
Segmentation fault in PRIMA on LLVM 14


certik:
https://github.com/lfortran/lfortran/actions/runs/19188066955/job/54858342823?pr=8804

```
2025-11-08T05:07:24.3394451Z +++ echo 'FC is lfortran --std=f23'
2025-11-08T05:07:24.3394750Z FC is lfortran --std=f23
2025-11-08T05:07:24.3395004Z +++ lfortran --std=f23 --cpp -c ./common/consts.F90
2025-11-08T05:07:24.3750781Z +++ lfortran --std=f23 --cpp -c ./common/infos.f90
2025-11-08T05:07:24.4087497Z +++ lfortran --std=f23 --cpp -c ./common/debug.F90
2025-11-08T05:07:24.4451557Z +++ lfortran --std=f23 --cpp -c ./common/huge.F90
2025-11-08T05:07:24.4795374Z +++ lfortran --std=f23 --cpp -c ./common/inf.F90
2025-11-08T05:07:24.5161037Z +++ lfortran --std=f23 --cpp -c ./common/infnan.F90
2025-11-08T05:07:24.5527990Z +++ lfortran --std=f23 --cpp -c ./common/memory.F90
2025-11-08T05:07:24.5953455Z +++ lfortran --std=f23 --cpp -c ./common/string.f90
2025-11-08T05:07:24.6489696Z +++ lfortran --std=f23 --cpp -c ./common/linalg.f90
2025-11-08T05:07:24.8067269Z +++ lfortran --std=f23 --cpp -c ./common/univar.f90
2025-11-08T05:07:24.9019520Z +++ lfortran --std=f23 --cpp -c ./common/ratio.f90
2025-11-08T05:07:24.9408698Z +++ lfortran --std=f23 --cpp -c ./common/redrho.f90
2025-11-08T05:07:24.9768108Z +++ lfortran --std=f23 --cpp -c ./common/xinbd.f90
2025-11-08T05:07:25.0653766Z +++ lfortran --std=f23 --cpp -c ./common/history.f90
2025-11-08T05:07:25.1697174Z +++ lfortran --std=f23 --cpp -c ./common/selectx.f90
2025-11-08T05:07:25.2753292Z +++ lfortran --std=f23 --cpp -c ./common/checkexit.f90
2025-11-08T05:07:25.3164222Z +++ lfortran --std=f23 --cpp -c ./common/fprint.f90
[...]

PR #8869, opened 2025-11-17
fix: trigger of exhaustive test on label changes

### LINKED ISSUE #8855: Exhaustive checks rerun when any label is removed
It's fine, but it's wasteful on the CI when we rework our labels, so it would be good to fix that.

Pranavchiku:
Fixes #8855.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -7,9 +7,7 @@ on:
     tags:
       - 'v*'
   pull_request:
-    types:
-      - labeled
-      - unlabeled
+    types: [labeled, unlabeled, synchronize]
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -7,12 +7,9 @@ on:
     tags:
       - 'v*'
   pull_request:
-    types: 
+    types:
       - labeled
       - unlabeled
-      - synchronize
-      - opened
-      - reopened
```
assem2002: Can't we keep those? I think after a sync you'd need to remove and add the label back

certik:
The failure is unrelated: https://github.com/lfortran/lfortran/issues/8610, we'll probably have to disable MPI, since it is failing quite often due to this, unless we can figure out how to fix it.

certik:
Now it will not run when you keep pushing into the PR, correct? Only when you add/remove the label. So you will have to remove/readd the label after every push, corect? I think that might not be ideal.

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -7,12 +7,9 @@ on:
     tags:
       - 'v*'
   pull_request:
-    types: 
+    types:
```
Pranavchiku:
```suggestion
    types: 
```

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -10,9 +10,6 @@ on:
     types: 
       - labeled
       - unlabeled
```
Pranavchiku:
```suggestion
      - unlabeled
      - synchronize
```

Pranavchiku:
Yes one has to add/remove label. Let me fix that. Marking it as draft.

Pranavchiku:
It is now working.

certik:
The failure is unrelated: https://github.com/lfortran/lfortran/issues/8808#issuecomment-3548360182.

PR #8871, opened 2025-11-17
Enh : Finalize Structs

### LINKED ISSUE #8844: Only add finalizer block for non-primitive types
TODO to fix next:

* Only add finalizer block for non-primitive types (so integers do not need a block). Note: allocatable variables are not primitive.

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/8826#issuecomment-3528438072_

assem2002:
closes #8844


src/libasr/pass/insert_deallocate.cpp
```diff
@@ -297,8 +288,8 @@ class LoopTempVarDeallocateVisitor : public ASR::BaseWalkVisitor<LoopTempVarDeal
 
 void pass_insert_deallocate(Allocator &al, ASR::TranslationUnit_t &unit,
                                 const PassOptions &/*pass_options*/) {
-    InsertDeallocate v(al);
-    v.visit_TranslationUnit(unit);
+    // InsertDeallocate v(al);
+    // v.visit_TranslationUnit(unit);
```
certik, line 292: Why not remove it? Does it mean we do not insert deallocation somewhere?

integration_tests/CMakeLists.txt
```diff
@@ -1393,7 +1393,7 @@ RUN(NAME bindc5 LABELS gfortran llvm)
 RUN(NAME bindc_01 LABELS gfortran llvm EXTRAFILES bindc_01b.f90 bindc_01c.c)
 RUN(NAME bindc_02 LABELS gfortran llvm EXTRAFILES bindc_02b.f90 bindc_02c.c)
 RUN(NAME bindc_03 LABELS gfortran llvm EXTRAFILES bindc_03c.c)
-RUN(NAME bindc_04 LABELS gfortran llvm EXTRAFILES bindc_04c.c OLD_CLASSES)
+# RUN(NAME bindc_04 LABELS gfortran llvm EXTRAFILES bindc_04c.c OLD_CLASSES)
```
certik, line 1396: These two are commented out because now they don't run with the old classes, and they still don't run with new classes?

certik (approved):
Thanks, I think that's good.

integration_tests/CMakeLists.txt
assem2002: Yes. `finalize_struct` operates on the new design only.

src/libasr/pass/insert_deallocate.cpp
assem2002: If you asking about why not remove those lines -- I just wanted to turn off the feature nothing more, but yes we can remove the lines.

If you're asking that we don't fully now insert any ASR deallocations, Yes we don't now. Gurarvsingh used this pass to insert some deallocation statements for the `array-lhs-realloc` thing; I think if we can port those deallocations to `array_op`, We can stop using this pass in our default pass list (They're used for code correctness not finalizing).

certik: Yes, we should I think we should stop inserting ASR deallocations, only used ones provided by the user. The rest is done in your finalizers.

assem2002: Yes. that's we do. Any other deallocations are for code correctness while we're applying our passes.

PR #8873, opened 2025-11-18
refactor: make a new util function to copy data from fchar to char

### LINKED ISSUE #8872: Refactor a string operation in the runtime
@akramhany do you mind please refactoring https://github.com/lfortran/lfortran/pull/8864#discussion_r2532445868 into nice string function(s) in the runtime? Over time we will refactor how we handle strings and memory allocation, but until then hide all string operations behind a function call, to make ti clear what the intent is.

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/8864#issuecomment-3548308537_

akramhany:
Fixes #8872

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5856,7 +5856,7 @@ LFORTRAN_API char *_lfortran_get_env_variable(char *name) {
     return getenv(name);
 }
 
-// This function assumes thatt length of src is at least len, and dest is at least len + 1 (1 for '\0').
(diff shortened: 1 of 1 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5856,15 +5856,16 @@ LFORTRAN_API char *_lfortran_get_env_variable(char *name) {
     return getenv(name);
 }
 
+// This function assumes thatt length of src is at least len, and dest is at least len + 1 (1 for '\0').
```
```suggestion
// This function assumes that the length of src is at least len, and dest is at least len + 1 (1 for '\0').
```

PR #8877, opened 2025-11-19
Debugging LSP bug

certik:


certik:
• Hang Analysis

  - test_document_highlight issues textDocument/documentHighlight (request
    id 2) at test_document_highlight-parallel-client.log:69-82. No response
    ever arrives; instead the server keeps emitting telemetry each second.
    outgoingMessages’s seen counter advances (telemetry only) while the request-
    specific response never shows up, so the client blocks until you hit Ctrl‑C.
  - During the hang the telemetry snapshots show requestPool stuck with
    numActive=3/numExecuted=6 and workerPool continuously executing cron
    jobs (e.g. workerPool.numExecuted climbs from 5 to 164). That means the
    request thread handling the highlight call never returns—most likely
    stuck in LFortranLspLanguageServer::receiveTextDocument_documentHighlight
    (src/bin/lfortran_lsp_language_server.cpp:933-982) or deeper in
    LFortranAccessor::getAllOccurrences, so nothing ever reaches the send queue
    and the pytest client waits forever.
  - No server log output was produced (test_document_highlight-parallel-
    server.log is empty), so we’ll need to instrument the highlight handler (or
    LFortranAccessor) next to determine whether we’re hung in ASR generation,
    resolve(), or the logger’s send-order barrier.

  Reproducer

  - Added ci/repro_document_highlight.sh, which stress-runs only
    test_document_highlight under the parallel execution strategy until it fails
    (or until an optional iteration limit is reached). Example:
      - Loop forever: ./ci/repro_document_highlight.sh
      - Stop after N passes: ./ci/repro_document_highlight.sh 25
      - The script uses pixi run -e test -- pytest -vv --showlocals
        --execution-strategy="parallel" tests/server/tests/
        test_features.py::test_document_highlight.
  - Verified it works for a single pass (./ci/repro_document_highlight.sh 1),
    so you can now leave it running to reproduce the hang much faster than
    rerunning the entire suite.

  Next Steps

  - Instrument receiveTextDocument_documentHighlight() (or
    LFortranAccessor::getAllOccurrences()) to log start/end timestamps and the
    path/line being queried so we can confirm where the worker gets stuck.
  - Consider enabling a non-zero timeoutMs in the test configuration to ensure
    the server returns a timeout error instead of hanging indefinitely; that
    would also make the new reproducer terminate on its own when the bug hits.
  - Once we catch a failing run using the new script, inspect the generated
    {client,stdout,stdin,server}.log files for that attempt—they’ll be limited
    to the highlight test, making the hang easier to analyze.

certik:
• Key log from the latest run is test_whole_document_formatting-parallel-
  client.log. It shows:

  - At 01:13:40.314Z the client sent textDocument/formatting (request id 2) for
    tests/function_call1.f90.
  - No response for that id ever arrived. Instead, only telemetry events were
    received once per second. By 01:29:17 the server had emitted 927 telemetry
    messages while requestPool stayed at numActive=3, numExecuted=6—meaning the
    request thread that should have completed id 2 never returned.
  - Because the request handler is effectively just
    LFortranLspLanguageServer::receiveTextDocument_formatting() calling
    lfortran.format(...) and then sending the result, this strongly suggests
    lfortran.format is the piece that hung. The telemetry confirms only the
    worker pool’s cron jobs were running; no new requests were executed.

  So the current failure is confined to the whole-document formatting request;
  our new range-format logging hasn’t even been hit yet. To move forward:

  1. Instrument LFortranAccessor::format() itself (e.g., log the input path,
     whether fe.get_asr2 succeeded, and when the function returns) using
     std::cerr or another guaranteed sink, since the logger file remains
     empty. That will tell us whether the hang occurs inside the ASR build, the
     formatter, or the indentation post-processing.
  2. Optional: temporarily set config["LFortran"]["timeoutMs"] to a non-zero
     value in tests/server/tests/conftest.py so that requests abort instead of
     spinning forever; that will make the reproducer fail fast once we find the
     problematic stage.

  The existing logs already isolate the bug to the textDocument/formatting
  handler, so the next step is to instrument LFortranAccessor::format() and
  rerun until the hang reoccurs.

certik:
• The latest hang is again in test_document_highlight. From
  test_document_highlight-parallel-client.log:

  - The client opened examples/expr2.f90, waited for diagnostics, then sent
    textDocument/documentHighlight (request id 2) at 06:34:10Z.
  - No response ever arrived—only telemetry events streamed in every second (409
    of them) while the request queue stayed idle. That means the worker thread
    handling the highlight request never returned and the client sat waiting
    until the suite stalled.

  Because the request loop is still stuck somewhere inside
  LFortranLspLanguageServer::receiveTextDocument_documentHighlight() (most
  likely in lfortran.getAllOccurrences), I’ve added verbose tracing to make the
  next failure self‑diagnosing:

  - RunTracer now logs every request/notification/response start/end with
    elapsed time to std::cerr, so every LSP method shows up in the per-test
    *-stdout.log files.
  - receiveTextDocument_documentHighlight, receiveTextDocument_formatting, and
    receiveTextDocument_rangeFormatting now log all of their parameters, input
    sizes, and stage timings to std::cerr. LFortranAccessor::format already does
    the same.
  - These traces go straight to stderr, so they’re visible even if the logger is
    configured to write elsewhere.

  After these changes, please re-run the stress script. When the hang manifests
  again, test_document_highlight-parallel-stdout.log will show which stage was
  the last to log (begin request, entering getAllOccurrences, exiting, etc.),
  giving us the exact place where the worker is stuck.

PR #8881, opened 2025-11-20
Fix the branch name in build to wasm

ubaidsk:
Post merging of https://github.com/lfortran/lfortran/pull/8867, our CI in main fails with:

```shell
error: src refspec main does not match any
error: failed to push some refs to 'github.com:lfortran/wasm_builds.git'
```


ubaidsk:
There is a CI failure, but it seems to be unrelated.

certik:
It's concerning that this wasn't caught in the PR --- any idea why?

ubaidsk:
>It's concerning that this wasn't caught in the PR --- any idea why?

It failed during the `git push` step which is the very last step of this script. We do this push only for the `main` branch CI run where the SSH key is present. So, this did not get tested in the PR CI check.

certik:
Ok, perfect!

PR #8884, opened 2025-11-21
Fix : Use Pointer and `associate` instead of `assignment` with nested functions using structType

### LINKED ISSUE #8780: Handle extended type containing array member in nested global variable
The following MRE fails with LFortran. To fix it, find the reason why it segfaults, isolate the error, and then figure out the fix.

MRE:
```fortran
module toml_mod
   implicit none
   type, abstract :: abstract_lexer
   end type abstract_lexer
   type, extends(abstract_lexer) :: mocked_lexer
        integer, allocatable :: token(:)
    end type mocked_lexer
contains
   subroutine xx(lexer)
      class(abstract_lexer), intent(inout) :: lexer
      call nested1()

      contains
         subroutine nested1()
            call semantic_error_tmp(lexer)
         end subroutine nested1
   end subroutine xx
   subroutine semantic_error_tmp(lexer)
      class(abstract_lexer), intent(inout) :: lexer
   end subroutine semantic_error_tmp
end module toml_mod

program test
   use toml_mod
   implicit none
   type(mocked_lexer) :: lexer

   call xx(lexer)
end program test
```

Output:
```console
$ gfortran b.f90 && ./a.out
```

assem2002:
closes #8780

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1881,6 +1881,7 @@ RUN(NAME nested_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME nested_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_vars1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)
 RUN(NAME nested_vars2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)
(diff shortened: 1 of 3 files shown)


certik:
The 3rd party codes fail with:
```
2025-11-21T01:58:04.3352199Z asr_to_llvm: module failed verification. Error:
2025-11-21T01:58:04.3352322Z Stored value type does not match pointer operand type!
2025-11-21T01:58:04.3353086Z   store %enum_token { i32 (...)** bitcast (i8** getelementptr inbounds ({ [4 x i8*] }, { [4 x i8*] }* @_VTable_enum_token, i32 0, i32 0, i32 2) to i32 (...)**), i32 -1, i32 -2, i32 -3, i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19 }, %enum_token** @__lcompilers_created__nested_context__parse_table_header_token_kind, align 8
2025-11-21T01:58:04.3353241Z  %enum_token*Stored value type does not match pointer operand type!
```
It would be good to extract an MRE for this and add it to our integration tests (that pass in this PR), so that it fails early next time already in integration tests.

PR #8887, opened 2025-11-22
fix: prevent passing in unallocated arguments when non-allocatable optional arguments are expected

xaerru:
Fix the `transform_optional_argument_functions` pass for when unallocated arguments are passed into a procedure which expects an optional non-allocatable type argument.

If the argument is allocated then we pass in the argument, if it is not allocated then we pass in a dummy variable declared on the stack of the caller.


xaerru:
Yes, it will. The bounds check PR will require some changes though. I will work on it after CI in this PR passes.

integration_tests/functions_52.f90
```diff
@@ -0,0 +1,12 @@
+program functions_52
+    integer, allocatable :: b
+    b = my()
+
+    if (b /= 44) error stop
```
certik, line 8: We can also assert not allocated before `my()` and allocated after.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12375,6 +12377,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            if (ASRUtils::is_array(orig_arg->m_type) && ASRUtils::is_array(arg->m_type) &&
+                                ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(orig_arg->m_type)) &&
+                                ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(arg->m_type))) {
+                                llvm::Type* llvm_orig_arg_type = llvm_utils->get_type_from_ttype_t_util(ASRUtils::EXPR(ASR::make_Var_t(
+                                    al, orig_arg->base.base.loc, &orig_arg->base)),
+                                    orig_arg->m_type, module.get());
+                                llvm::Type* llvm_arg_type = llvm_utils->get_type_from_ttype_t_util(x.m_args[i].m_value, arg->m_type, module.get());
+                                if (llvm_orig_arg_type != llvm_arg_type) {
+                                    tmp = builder->CreateBitCast(tmp, llvm_orig_arg_type->getPointerTo());
+                                }
```
certik, line 12390: Can you explain why the types would not match?

We are trying to ensure types always match in ASR, so that in the backend we don't need to do any of these checks and casts. Casts should ideally only be done with an explicit ASR casting node. Otherwise we should be able to assume the types always agree between arguments and parameters.

src/libasr/pass/transform_optional_argument_functions.cpp
```diff
@@ -558,13 +637,61 @@ class ReplaceFunctionCallsWithOptionalArgumentsVisitor : public ASR::CallReplace
...
+                        }
+                    }
+                    new_body.push_back(al, body[i]);
+                }
+                if (body.n == new_body.n) {
+                    converge = true;
+                }
+                body.n = new_body.n;
+                body.p = new_body.p;
+            }
```
certik, line 692: Don't add potentially infinite loops like this. Add a for loop with NUM_TRIES=3 (or more?) and if it fails to convert within NUM_TRIES, give a clear error that we failed to convert all FunctionCalls.

Add an ASR test for recursive function calls, so that we can see that it got transformed correctly --- run this pass and check the result.

tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_45-db3a04a.stdout
```diff
@@ -875,7 +1115,97 @@
...
+                                        [(Associate
+                                            (Var 6 __libasr_created_variable_pointer_4)
+                                            (Var 6 file_scope_flags)
+                                        )]
+                                        [(Associate
+                                            (Var 6 __libasr_created_variable_pointer_4)
+                                            (Var 6 __libasr_created_dummy_variable_4)
+                                        )]
+                                    )
+                                    (Assignment
```
certik, line 1208: I checked in `integration_tests/modules_45.f90` and this logic seems correct.

It would be worth adding a dedicated ASR test just for this case that the PR is fixing (allocatable argument to non-allocatable optional parameter), since `modules_45.f90` is quite large.

src/libasr/codegen/asr_to_llvm.cpp
xaerru: For this code:
```fortran
module my_mod
    type :: base
        integer :: x
    end type base

        integer :: y
    end type derived

    contains
        subroutine my(a, b)
            class(base), intent(in) :: a(:)
            integer, optional :: b

            print *, a(1)%x
        end subroutine
end module

program my_program
    use my_mod

    integer, allocatable :: i
    type(derived), allocatable :: my_array(:)

    allocate(my_array(1))
    my_array(1) = derived(33, 44)

    i = 10

    call my(my_array, i)
end program
```

LLVM  of `my` in main:
```llvm
; Defined with these types
define void @__module_my_mod_my_base____0(%base* %a, i32* %__1a, i32* %b, i1* %is_b_present_) {}
; Called like this
call void @__module_my_mod_my_base____0(%base* %384, i32* %call_arg_value, i32* %391, i1* %call_arg_value67)
```
In this PR:
```llvm
%array = type { %base*, i32, %dimension_descriptor*, i1, i32 }
%array.0 = type { %derived*, i32, %dimension_descriptor*, i1, i32 }
define void @__module_my_mod_my(%array* %a, i32* %b, i1* %is_b_present_) {}

; Called like this
%143 = load %array.0*, %array.0** %my_array, align 8
%144 = bitcast %array.0* %143 to %array* ; This is the bitcast being added here
call void @__module_my_mod_my(%array* %144, i32* %145, i1* %call_arg_value)
```

This happens because in main `pass_array_by_data` transforms the `DescriptorArray` to a `PointerArray` with size passed in.

But in this PR we are passing in a `Pointer` type to argument `b` in `my` subroutine, so `pass_array_by_data` doesn't transform it and the descriptor is used. The descriptors have different types because we pass in a `derived` type struct array to a subroutine which expects a `base` type struct array.

src/libasr/pass/transform_optional_argument_functions.cpp
xaerru: Set `NUM_TRIES` to 10 and added `tests/optional_03.f90` which has nested function calls.

tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_45-db3a04a.stdout
xaerru: Added `tests/optional_02.f90`.

src/libasr/codegen/asr_to_llvm.cpp
certik: Ok, so it's just a cast from derived to base?

That's a classic cast and it must be explicit in ASR. The backend should not do any of this casting or worrying that types do not match. 

> The descriptors have different types because we pass in a derived type struct array to a subroutine which expects a base type struct array.

This should not be allowed in ASR, we must cast it properly first.

certik (commented):
I am fine with merging this, if you'll be able to fix the casting issue, as that's an important design --- we should not be doing any such casting in the backend at all.

But I want to lock in the win, by merging this.

@jinangshah21 how difficult do you think it would be to fix the casting here to happen already in ASR, not in the backend?

jinangshah21:
For https://github.com/lfortran/lfortran/pull/8887#discussion_r2572206689 example, currently in backend we have `convert_to_polymorphic_arg` utility which does `derived` to `base_class` conversion. But we can simply Add Cast node in ASR for this also. I think it is not so difficult to do. But It can be handled in separate PR.

certik:
Ok, I made this subsequent work this issue: https://github.com/lfortran/lfortran/issues/8931

PR #8895, opened 2025-11-24
LLVM: use constant initializer for character array descriptors

certik:
The declare_constant_stringArray function was incorrectly using builder->CreateStore() at global scope to set the length field of string descriptors. Store instructions can only exist inside functions, not at the module/global level, causing a segmentation fault during LLVM verification.

This fix replaces the invalid approach with proper constant struct initialization:
- Create global variables with llvm::ConstantStruct::get() initializers
- Set both the data pointer and length field at declaration time
- Match the pattern used by handle_global_nonallocatable_stringArray

The fix allows character arrays in module-level DATA statements to compile successfully to LLVM IR.

Fixes segfault in src/libasr/codegen/llvm_utils.cpp:2298. Test added.


assem2002 (approved):
Looks good. Thanks for the fix.
We need to refactor this function along with `handle_global_nonallocatable_stringArray()`, They're so much alike.

certik:
> Looks good. Thanks for the fix. We need to refactor this function along with `handle_global_nonallocatable_stringArray()`, They're so much alike.

Sure, I made it https://github.com/lfortran/lfortran/issues/8898.

Thanks for the review.

PR #8896, opened 2025-11-24
Fix: crash when assigning to SAVE character substring

certik:
Fixed a bug where substring assignment to a character variable with the SAVE attribute (explicit or implicit) caused a segmentation fault by attempting to write to read-only constant memory.

Root Cause: In visit_StringSection() (asr_to_llvm.cpp:9217), when a StringSection node had a compile-time constant value (x.m_value), the code would always use that constant value regardless of context. When the StringSection appeared as an assignment target (e.g., ch(5:5) = 'X'), this caused the LLVM backend to generate code that attempted to write to a constant string descriptor in read-only memory instead of the actual mutable SAVE variable's data buffer.

The Fix: Modified visit_StringSection() to check the is_assignment_target flag. When a StringSection is used as an assignment target, we now skip the compile-time constant optimization and properly compute the mutable substring location by calling stringSection_helper_fortran().

This ensures the LLVM backend generates correct code that writes to the actual SAVE variable's mutable data buffer.

Testing: Added integration_tests/string_85.f90 which verifies:
- Substring assignment to SAVE variables works without crashing
- Modified values are correctly stored and persist between calls
- Both explicit and implicit SAVE attributes are handled correctly

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7201,6 +7201,16 @@ public:
                                              module.get(), true);
             }
         } else {
+            if (!ASR::is_a<ASR::Var_t>(*x.m_target)) {
+                std::string target_type = "unknown";
+                if (x.m_target->type == ASR::exprType::StringConstant) {
+                    target_type = "StringConstant (this is likely a bug in an ASR pass that incorrectly "
+                                  "constant-folded an assignment target like ch(5:5) = 'X')";
+                } else {
+                    target_type = "type " + std::to_string(x.m_target->type);
+                }
+                throw CodeGenError("Assignment target must be a variable or subscripted expression, got: " + target_type);
+            }
             ASR::Variable_t *asr_target = EXPR2VAR(x.m_target);
             h = get_hash((ASR::asr_t*)asr_target);
             target = llvm_symtab[h];
@@ -9214,11 +9224,14 @@ public:
     }
 
     void visit_StringSection(const ASR::StringSection_t& x) {
-        // Don't use compile-time value when this is an assignment target
+        // Don't use compile-time constant value when this is an assignment target
+        // For assignment targets, we need the actual mutable location in the string,
+        // not a constant value that would cause writes to read-only constant memory
         if (x.m_value && !is_assignment_target) {
             return this->visit_expr_wrapper(x.m_value, true);
         }
 
+        // Compute the actual substring location for reading or writing
         // TODO : Find some way to use the helper functions based on the frontend,
         // We are using fortran only for now.
         if(true){ // Fortran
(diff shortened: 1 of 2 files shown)


assem2002 (approved):
Looks good!

assem2002:
CI failure with added integration test. `--Fast` + LLVM >=17.
probably nothing to do with the fix, The change just invoked another bug.

certik:
I can trigger it locally with LLVM 11 also:
```
$ lfortran integration_tests/string_85.f90 --fast
    if (res.ok) {
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 14814
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 359
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 333
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 360
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 22
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 1270
    visit_symbol(*item.second);
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 4397
    visit_procedures(x);
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5893
    ASR::Function_t *s = ASR::down_cast<ASR::Function_t>(item.second);
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5503
    visit_procedures(x);
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5844
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 54
    case stmtType::Assign: { v.visit_Assign((const Assign_t &)x); return; }
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 7203
    } else {
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_utils.h", line 407
    ASR::down_cast<ASR::Var_t>(f)->m_v));
  File "/Users/ondrej/repos/lfortran/src/libasr/asr.h", line 41
    LCOMPILERS_ASSERT(is_a<T>(*f));
AssertFailed: is_a<T>(*f)
```
Not sure why it doesn't trigger at the CI. Investigating.

certik:
It was the same bug, in optimization pass. I explained the details in the git commit.

assem2002:
Great! 
Why didn't that fail with LLVM < 17 ? Don't we test those with `--fast`?

certik:
>  Why didn't that fail with LLVM < 17 ? Don't we test those with --fast?

We need to check, I am pretty sure we do check it.

certik:
These network failures are annoying. Here is how to fix them: https://github.com/lfortran/lfortran/issues/8900.

Issue #8888, opened 2025-11-23
Complex division wrong


harperjf:
This program gives output wrong in two ways. If ```pr``` is changed from 6 to 15,  for double precision arithmetic, the same errors occur.
```
program complexdiv
  implicit none
  integer,parameter :: pr = 6, wp = selected_real_kind(pr)
  real(wp),parameter:: small = tiny(1.0_wp)
  complex(wp)       :: a, b
  a = cmplx(-2,1,wp)*small
  b = cmplx(-1,2,wp)*small
  print *, a/b
end program complexdiv
```
Output from ifx or AMD flang (correct):
```
 (0.8000000,0.6000000)
```
Gfortran and g95 give answers within the expected floating-point accuracy.
Lfortran 0.58.0 --std=f23 gives this output.
```
 (         -nan,          -nan)
```
The two errors are (1) a/b should be finite, (2) nan should not be printed with a sign, as already reported in #3939. I now use Ubuntu 25.04.

PR #8903, opened 2025-11-25
enh: optimize complex div code to handle very large and small values

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8888


certik (approved):
Excellent, that's good.

Eventually we'll refactor these operations to happen in ASR->LLVM, so that LLVM can optimize out these little operations when possible.

Issue #8812, opened 2025-11-10
Handle string kind in default init value of struct member


jinangshah21:
MRE:
```fortran
program class_77
   integer, parameter :: tfc = selected_char_kind('DEFAULT')
    type :: toml_char
      character(1, tfc) :: hash = tfc_"#"
    end type
    type(toml_char) :: char_kind = toml_char()
    print *, char_kind%hash
end program class_77
```

Outputs:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran b.f90
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ ./a.out
 #
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90
f
```

PR #8904, opened 2025-11-25
enh: support handling of kind prefix with string

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8812
fixes: https://github.com/lfortran/lfortran/issues/8814
closes: https://github.com/lfortran/lfortran/pull/8827

### DIFF
--- a/src/lfortran/parser/parser_stype.h
+++ b/src/lfortran/parser/parser_stype.h
@@ -102,7 +102,6 @@ static_assert(std::is_trivial<YYSTYPE>::value);
 // Ensure the YYSTYPE size is equal to Vec<AST::ast_t*>, which is a required member, so
 // YYSTYPE has to be at least as big, but it should not be bigger, otherwise it
 // would reduce performance.
-// StrSuffix is now: Str (16 bytes) + Str* (8 bytes) = 24 bytes, same as Vec<AST::ast_t*>
 #if !defined(HAVE_BUILD_TO_WASM) && !defined(__ppc__)
 static_assert(sizeof(YYSTYPE) == sizeof(Vec<AST::ast_t*>));
 #endif


src/lfortran/parser/tokenizer.h
```diff
@@ -46,10 +46,27 @@ class Tokenizer
...
+            p++;
+        }
+        
+        // If no kind prefix found, skip just the opening quote
+        if (start == tok) {
+            start = tok + 1;
+        }
+        
+        s.p = (char*) start;
+        s.n = cur - start - 1; // Subtract 1 for closing quote
```
certik: I think I commented on the same change before --- this is not the right fix. We must parse it to AST, and it should be done in the lexer, not in this function.

HarshitaKalani: Please review now.

grammar/AST.asdl
```diff
@@ -216,7 +216,7 @@ expr
     | Num(int n, string? kind)
     | Real(string n)
     | Complex(expr re, expr im)
-    | String(string s)
+    | String(string s, string? kind)
```
certik, line 219: Yes, that's good.

src/lfortran/parser/parser_stype.h
```diff
@@ -45,11 +45,17 @@ struct IntSuffix {
...
+    Str str_s;
+    Str str_kind;
+};
+
 union YYSTYPE {
     int64_t n;
     Str string;
 
     IntSuffix int_suffix;
+    StrSuffix str_suffix;
```
certik, line 58: The change here is probably ok. The alternative design is to pass in the suffix as two tokens. I like this PRs design better, similar to IntSuffix.

tests/reference/ast-program_without_line_cc-78281a4.stdout
```diff
@@ -2,7 +2,7 @@
     [(Print
         0
         ()
-        [(String "OK")]
+        [(String "OK" ())]
```
certik, line 5: I can't figure out why this change is here, as I don't see any change in ASR.asdl.

certik: I think the contents of this file is AST. But why would the file say "asr"?

certik (approved):

tests/reference/ast-program_without_line_cc-78281a4.stdout
HarshitaKalani: Yes, the file name is wrong. It is AST only. I'll correct that.

HarshitaKalani:
Ready for review.

src/lfortran/parser/tokenizer.h
```diff
@@ -54,6 +54,38 @@ class Tokenizer
...
+            p++;
+        }
+        
+        // No kind prefix found - just extract string
+        str_suffix.str_s.p = (char*) tok + 1;
+        str_suffix.str_s.n = cur-tok-2;
+        str_suffix.str_s.p = str_unescape_fortran(al, str_suffix.str_s, ch);
+        str_suffix.str_s.n = strlen(str_suffix.str_s.p);
+    }
+
```
certik, line 94: This is fine for now, but I think it's not very efficient, I think there is a faster way to tokenize this --- in re2c we already identified this as a string with a suffix, so we should not be looping over the whole string again.

src/lfortran/parser/parser_stype.h
```diff
@@ -96,8 +102,10 @@ static_assert(std::is_trivial<YYSTYPE>::value);
 // Ensure the YYSTYPE size is equal to Vec<AST::ast_t*>, which is a required member, so
 // YYSTYPE has to be at least as big, but it should not be bigger, otherwise it
 // would reduce performance.
+// Note: StrSuffix (two Str fields) makes YYSTYPE 32 bytes vs Vec's 24 bytes on 64-bit.
+// This 8-byte difference is acceptable for modern systems.
 #if !defined(HAVE_BUILD_TO_WASM) && !defined(__ppc__)
-static_assert(sizeof(YYSTYPE) == sizeof(Vec<AST::ast_t*>));
+static_assert(sizeof(YYSTYPE) <= sizeof(Vec<AST::ast_t*>) + 8);
```
certik: This will not work, it will slow everything down. I was worried about this.

src/lfortran/parser/parser_stype.h
```diff
@@ -45,11 +45,17 @@ struct IntSuffix {
     Str int_kind;
 };
 
+struct StrSuffix {
+    Str str_s;
+    Str str_kind;
```
certik: Make `str_kind` just a pointer and allocate it using Arena. That should fit into 24 bytes --- Str is a pointer + length (16 bytes), and str_kind just a pointer to Str, so 8 bytes, total 24.

src/lfortran/parser/parser_stype.h
```diff
@@ -96,6 +102,7 @@ static_assert(std::is_trivial<YYSTYPE>::value);
 // Ensure the YYSTYPE size is equal to Vec<AST::ast_t*>, which is a required member, so
 // YYSTYPE has to be at least as big, but it should not be bigger, otherwise it
 // would reduce performance.
+// StrSuffix is now: Str (16 bytes) + Str* (8 bytes) = 24 bytes, same as Vec<AST::ast_t*>
```
certik: This comment doesn't need to be here.

certik:
This looks good. Let me give this one last thorough review, and if it is good, I'll merge it.

src/lfortran/parser/tokenizer.h
```diff
@@ -54,6 +54,44 @@ class Tokenizer
         s.n = strlen(s.p);
     }
 
+    // Parse string literal with optional kind prefix (e.g., tfc_"#" or "#")
+    // Extracts both the string content and the kind prefix (if present)
+    // The kind prefix is allocated in Arena as a Str* to keep StrSuffix at 24 bytes
+    void lex_string(Allocator &al, StrSuffix &str_suffix, char ch) const
```
certik, line 60: It's a prefix, not a suffix, right?
```suggestion
    void lex_string(Allocator &al, StrPrefix &str_prefix, char ch) const
```

src/lfortran/parser/parser.yy
```diff
@@ -118,7 +118,7 @@ void yyerror(YYLTYPE *yyloc, LCompilers::LFortran::Parser &p,
 %token TK_PERCENT "%"
 %token TK_VBAR "|"
 
-%token <string> TK_STRING
+%token <str_suffix> TK_STRING
```
certik, line 121:
```suggestion
%token <str_prefix> TK_STRING
```

src/lfortran/parser/parser_stype.h
certik:
```suggestion
```

Issue #8907, opened 2025-11-26
selected_logical_kind missing


harperjf:
This program uses a Fortran 2023 feature, selected_logical_kind.
```
  integer, parameter :: lk = selected_logical_kind(1)
  print "(*(g0))",'logical kind  ',lk
end program
```
Output from lfortran 0.58.0 --std=f23
```
semantic error: Initialization of `lk` must reduce to a compile time constant.
--> logicalkind.f90:1:3
  |
1 |   integer, parameter :: lk = selected_logical_kind(1)
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

The Fortran 2023 standard requires the ```selected_logical_kind``` intrinsic, but Fortran 2018 did not. The Lfortran Documentation https://docs.lfortran.org/en/ Roadmap item 3a says
```
Implement more Fortran features, until eventually full Fortran 2018 is supported
```
so I'm not sure whether the bug is in lfortran or its documentation. (Is it now hoped that the full ```std=f23``` standard will be supported one day?)

harperjf:
Logical8 is also missing. Test program:
``` 
  use iso_fortran_env, only: logical8
  logical(logical8) x
  x = .true.
  print *,x
end program
```
gfortran output:
```
 T
```
lfortran 0.58.0 output
```
semantic error: The symbol 'logical8' not found in the module 'iso_fortran_env'
 --> logical8.f90:1:30
  |
1 |   use iso_fortran_env, only: logical8
  |                              ^^^^^^^^ 
```

certik:
Both codes in this issue now work. The first one:
```
$ lfortran a.f90
logical kind  1
```
and the second one:
```
$ lfortran b.f90
T
```

PR #8908, opened 2025-11-26
enh: implement selected_logical_kind intrinsic

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/8907

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1275,7 +1275,7 @@ RUN(NAME intrinsics_393 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_394 LABELS gfortran llvm) # execute_command_line
 RUN(NAME intrinsics_395 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # cmplx kind cast regression
 RUN(NAME intrinsics_396 LABELS gfortran llvm)
-RUN(NAME intrinsics_397 LABELS llvm llvm_wasm llvm_wasm_emcc fortran) # selected_logical_kind
+RUN(NAME intrinsics_397 LABELS llvm llvm_wasm llvm_wasm_emcc) # selected_logical_kind
 
 RUN(NAME la_constants LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # LAPACK constants
 


certik:
GFortran failed to compile it --- let's fix the test so that GFortran compiles it.

HarshitaKalani:
I'm unable to reproduce it on macOS. Can this be because of the GFortran version we use in the CI?

HarshilShah1804:
I think it is some version issue with GFortran. I was using GFortran 13.3.0 and fails to compile even this code:
```fortran
program intrinsics_397

    use, intrinsic :: iso_fortran_env
    implicit none

    integer, parameter :: lk1 = selected_logical_kind(1)
end program intrinsics_397
```

When I checked on godbolt, it is not compiling with GFortran version 13 and 14. It works only with version 15. Probably, we use an earlier version at the CI.

HarshitaKalani:
That's what I feel too.

integration_tests/CMakeLists.txt
```diff
@@ -1275,6 +1275,7 @@ RUN(NAME intrinsics_393 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_394 LABELS gfortran llvm) # execute_command_line
 RUN(NAME intrinsics_395 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # cmplx kind cast regression
 RUN(NAME intrinsics_396 LABELS gfortran llvm)
+RUN(NAME intrinsics_397 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # selected_logical_kind
```
```suggestion
RUN(NAME intrinsics_397 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # selected_logical_kind
```

HarshitaKalani:
The program works with the fortran flag. I think the issue is because of the gfortran version. The version that we use at the CI does not support selected_logical_kind and hence the program fails.

certik:
Go ahead and push in a proper fix to get the CI working.

certik:
Don't install gfortran at the CI in this PR. That's a delicate operation that must be done in a dedicated PR and we need to ensure we did it correctly.

In this PR, just skip GFortran tests for this one test, let's get it merged.

HarshitaKalani:
I tried adding gfortran version >= 14.0 in the environment.yaml but it does not work.
The Linux CI fails with a linker error where the system is looking for /lib64/libm.so.6 and /lib64/libmvec.so.1 but can't find them. This is happening because GFortran 15 from conda has different library dependencies than what's available on the GitHub Actions runner.

The selected_logical_kind intrinsic is a Fortran 2023 feature that's only in GCC 14+, so it's expected that CI with older GFortran won't support it. Can we skip the gfortran test label for this specific test?

certik:
Yes, just skip the gfortran test.

We can worry about enabling newer gfortran in a dedicated PR.

PR #8909, opened 2025-11-26
Fix dimension pointer corruption in pass_array_by_data

krystophny:
## Summary
Fix segfault when compiling LAPACK code with mixed whole-array and subscripted array arguments in implicit interface calls.

## Problem
In `pass_array_by_data.cpp`, the pass was copying the `m_dims` pointer from source array to target array. When multiple `ArrayPhysicalCast` nodes reference arrays with different dimensions, this causes pointer corruption.

## Fix
Use `ASRUtils::duplicate_dimensions()` to deep copy dimensions instead of sharing the pointer.

## Test
Added `lapack_02.f90` - minimal reproducer extracted from LAPACK's `sgbtrf.f`.


krystophny:
@certik that's the first small patch. Btw I wrote on Fortran Discourse, in case you want to discuss the bigger picture in a video call, I'm usually available at 11am in Los Alamos time (8pm in Austria).

PR #8910, opened 2025-11-26
Fix fixed-form prescan crash on files without final newline

krystophny:
Merge after #8909

## Summary
Fix crash in fixed-form prescanner when file ends with whitespace but no final newline.

## Problem  
CMake-generated compiler test files can end with trailing whitespace and no newline. This causes out-of-bounds string access in `skip_rest_of_line()` and `copy_rest_of_line()` in `parser.cpp`.

## Fix
Add bounds check `pos < s.size()` before accessing `s[pos]`.

## Test
Added `lapack_03.f90` - minimal reproducer with trailing whitespace.


certik (approved):
Good catch!

PR #8911, opened 2025-11-26
Fix implicit interface type check for FORTRAN 77 sequence association

krystophny:
Merge after #8910

## Summary

Fix compiler crash when calling EXTERNAL subroutines with implicit interfaces in FORTRAN 77 fixed-form code using sequence association.

## Problem

Compiling LAPACK's `slals0.f` triggered a crash:
```
ASRUtils::check_equal_type(i32[:], r32[:, :])
at asr_utils.h:6704
```

The crash occurred when passing `B(PERM(I), 1)` to an EXTERNAL subroutine, where `PERM` is an integer array and `B` is a 2D real array. The compiler incorrectly attempted to enforce strict type checking between the actual argument's physical type and the dummy argument's type, which is incompatible with FORTRAN 77 sequence association semantics.

## Solution

1. **Skip type check for implicit interface functions** (asr_utils.h:6704)
   - Detect implicit interfaces via `func_type->m_deftype == ASR::deftypeType::Interface`
   - Allow F77-style sequence association where array physical types may differ

2. **Add LLVM < 15 bitcast workaround** (asr_to_llvm.cpp:13470-13480)
   - LLVM < 15 uses typed pointers requiring exact type matches
   - Bitcast pointer arguments when types don't match to satisfy LLVM verification

## Testing

**Integration test**: Added `lapack_04.f90` - minimal 14-line MRE from LAPACK SLALS0

**Test results**:
- LLVM 21: 1935/1939 pass (99.8%) - 3 env-var test failures + lapack_04 MRE (expected)
- LLVM 11: 1935/1939 pass (99.8%) - same 3 + lapack_04
- No regressions (class_63 failure is pre-existing)

**Verified**: Fixes #5544

### DIFF
--- a/integration_tests/lapack_04.f90
+++ b/integration_tests/lapack_04.f90
@@ -13,3 +13,27 @@ C MRE: implicit-interface array physical type mismatch from LAPACK SLALS0
    90 CONTINUE
 
       END
+
+      SUBROUTINE SCOPY( N, X, INCX, Y, INCY )
+      INTEGER N, INCX, INCY
+      REAL X(*), Y(*)
+C Stub implementation - just to make test compile and run
+      END
+
+      SUBROUTINE SGEMV( TRANS, M, N, ALPHA, A, LDA, X, INCX,
+     $                  BETA, Y, LDY )
+      CHARACTER TRANS
+      INTEGER M, N, LDA, INCX, LDY
+      REAL ALPHA, BETA
+      REAL A(LDA,*), X(*), Y(*)
+C Stub implementation - just to make test compile and run
+      END
+
+      PROGRAM LAPACK_04
+      INTEGER PERM(2)
+      REAL B(3,2), BX(3,2), WORK(2)
+      PERM(1) = 1
+      PERM(2) = 2
+      CALL SLALS0( 2, B, 3, BX, 3, PERM, 2, WORK )
+      PRINT *, 'PASS'
+      END


src/libasr/codegen/asr_to_llvm.cpp
krystophny, line 1: @certik this looks again a bit edgy for LLVM < 15. Maybe you understand better the consequences. I didn't manage to get this done up to now without a bitcast.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13465,6 +13465,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            llvm::FunctionType* fn_type = fn->getFunctionType();
+            for (size_t i = 0; i < args.size() && i < fn_type->getNumParams(); i++) {
+                llvm::Type* expected_type = fn_type->getParamType(i);
+                llvm::Type* actual_type = args[i]->getType();
+                if (expected_type != actual_type &&
+                    expected_type->isPointerTy() && actual_type->isPointerTy()) {
+                    args[i] = builder->CreateBitCast(args[i], expected_type);
+                }
+            }
+            #endif
```
certik, line 13480: So this just works for LLVM >= 16, but fails for LLVM < 15?

Let me think what is going on here --- if this is hiding a bug, or not. We might be missing a cast in ASR itself.

If we want this, then we'll create a function in utils called "bitcast" that does nothing from LLVM >= 16. So that we don't have ifdefs in asr_to_llvm.

certik:
Tests pass, let me rebase this to simplify the diff and let me run all tests.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: Yes I ran into this problem always when trying to use bare pointers with LLVM<15. I interpret it such that they made the type checks for pointers less strict in later versions. We could add extra guards for the legacy array section flag if this is needed.

src/libasr/asr_utils.h
```diff
@@ -6700,8 +6700,11 @@ static inline void Call_t_body(Allocator& al, ASR::symbol_t* a_name,
                 // TODO: Make this a regular error. The current asr_utils.h is
                 // not setup to return errors, so we need to refactor things.
                 // For now we just do an assert.
+                // Skip type check for implicit interface functions (FORTRAN 77 sequence association)
+                bool is_implicit_interface = (func_type->m_deftype == ASR::deftypeType::Interface);
```
certik, line 6704: This should fix the warning made error:
```suggestion
                [[maybe_unused]] bool is_implicit_interface = (func_type->m_deftype == ASR::deftypeType::Interface);
```

certik:
I think this might not be the right fix. First of all, let's create an MRE for this bug, to make it easier to investigate what went wrong and what the right fix should be. By simplifying the reproducible example I came up with:
```fortran
subroutine scopy(x, y)
  real :: x(*), y(*)
end subroutine

program main
  integer :: perm(1)
  real :: bx(1,1)
  external scopy

  call scopy(bx, bx)
  call scopy(bx(1, 1), bx(perm(1), 1))
end program
```
Which still gives the same error:
```console
$ lfortran c.f90 --implicit-interface --legacy-array-sections
LCOMPILERS_ASSERT failed: /Users/ondrej/repos/lfortran/src/libasr/../libasr/asr_utils.h
function Call_t_body(), line number 6706 at
ASRUtils::check_equal_type(arg_type, orig_arg_type, arg_expr, orig_arg_expr)
ERROR MESSAGE:
ASRUtils::check_equal_type(i32[:], r32[:, :])
    if (res.ok) {
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 14839
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/pass_manager.h", line 318
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/pass_manager.h", line 189
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/nested_vars.cpp", line 930
    w.visit_TranslationUnit(unit);
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/nested_vars.cpp", line 466
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/nested_vars.cpp", line 480
    transform_stmts(xx.m_body, xx.n_body);
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 33
    self().visit_stmt(*m_body[i]);
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 83
    case stmtType::Assert: { v.visit_Assert((const Assert_t &)x); return; }
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/nested_vars.cpp", line 619
    }
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_utils.h", line 6792
    }
  Binary file "/usr/lib/system/libsystem_c.dylib", local address: 0x18037ca3b
  Binary file "/usr/lib/system/libsystem_pthread.dylib", local address: 0x18047388b
```
Now we need to investigate the ASR and ensure it is well formed. All the types have to exactly agree. If done correctly, we shouldn't need to do any of the casts in the backend like this.

certik:
I think this analysis is accurate:
```
Type Mismatch in ASR

  Looking at the ASR output, there are two different scopy definitions that
  don't agree:

  1. Actual scopy subroutine (lines 258-351 of ASR)

  - Parameters x and y are 1D arrays:
  Array(Real 4)
  [(()  ())]  ← Single dimension
  DescriptorArray

  2. Implicit interface scopy in main (lines 64-165 of ASR)

  - Parameters scopy_arg_0 and scopy_arg_1 are 2D arrays:
  Array(Real 4)
  [(()  ())   ← First dimension
   (()  ())]  ← Second dimension
  PointerArray

  Why? Because the first call call scopy(bx, bx) passes a 2D array bx(1,1),
  so the implicit interface is inferred as expecting 2D arrays.

  ---
  The Bug: Second Call Arguments (lines 203-256)

  In call scopy(bx(1, 1), bx(perm(1), 1)):

  First argument (lines 206-240): ✅ OK
  - bx(1, 1) → Array section → Cast to 2D real array
  - Type: r32[:, :] matches the implicit interface

  Second argument (lines 241-253): ❌ WRONG!
  ArraySection(
    Var perm    ← Using the INTEGER array perm!
    ...
    Array(Integer 4)   ← Integer type, not Real!
    [(()  ())]         ← 1D array

  The compiler is creating an ArraySection of perm instead of:
  1. Evaluating perm(1) to get an integer scalar
  2. Using that scalar to index bx

  So it's passing an i32[:] (integer 1D array) where the implicit interface
  expects r32[:, :] (real 2D array).

  This is the exact mismatch in the error message:
  ASRUtils::check_equal_type(i32[:], r32[:, :])
```

So our ASR is incorrect. We need to fix that first.

certik:
I tried to fix it in https://github.com/lfortran/lfortran/pull/8914 and it fixes this particular issue, but it breaks lapack_02.f90 test.

So more work is needed.

The basic idea is this: ASR must be produces valid types that agree. For these implicit interfaces we use BindC, since that seems to be the only way to get `f(A(1))` working for both a scalar and an array section. 

It should be possible to do.

Issue #8764, opened 2025-10-29
Parameterized array of derived type values doesn't compile


w6ws:
```
module types_m
  implicit none

  type int_t
    integer :: i
  end type

  type(int_t), parameter ::  &
    T_ONE = int_t(1),  &
    T_TWO = int_t(2),  &
    T_THREE = int_t(3)

  type(int_t), parameter ::  &
    T_ALL(3) = (/ T_ONE, T_TWO, T_THREE /)

end module
```
No error with gfortran.  Lfortran won't compile it:
```
$ gfortran -c type_array.f90
$ lfortran -c type_array.f90
semantic error: Array members must me of the same type as the struct
  --> type_array.f90:14:19
   |
14 |     T_ALL(3) = (/ T_ONE, T_TWO, T_THREE /)
   |                   ^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
Version of lfortran is from the trunk as of a few days ago:
```$ lfortran --version
LFortran version: 0.57.0-168-g6bd06aa68
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
```
Secondarily, note that the error message has a spelling error:  'me' should be 'be'.

certik:
@w6ws awesome, thanks for both bug reports!

AnshMehta1:
Working on it.

PR #8912, opened 2025-11-27
[Feat] : Support derived type parameter arrays in ASR (ASR only, no LLVM)

opixdown:
## **Issue Fixed**
#8764 Parameterized array of derived type values doesn't compile

## **Summary**

This PR implements ASR support for parameter arrays of derived/struct types by creating an
ASR::ArrayConstructor whose elements are self contained ASR::StructConstant expressions. This avoids packing derived types into ArrayConstant

## **Current Behavior**
```
semantic error: Array members must me of the same type as the struct
  --> type_array.f90:14:19
   |
14 |     T_ALL(3) = (/ T_ONE, T_TWO, T_THREE /)
   |                   ^^^^^ 

```

## **Expected**
Should correctly initialize and access derivedtype array parameters.

## **Note**

-  This PR allow generating a valid ASR. The LLVM lowering for derived types is not yet implemented, therefore output running the program give garbage value. 

- I added a reference ASR test not an integration runtime test to lock the ASR behavior.

## **Request**
Please review the ASR shape. If it correct, we can merge this and then work on adding the LLVM lowering for derived type ArrayConstructor.


certik:
Thanks for the PR! Update reference tests via `./run_tests.py -u` and commit the result to fix the CI failure.

opixdown:
Thanks! I’ve updated the reference tests. Please let me know if anything else is needed

certik:
Don't remove `tests/interop/mod1-14.mod` in this PR, that's a separate issue.

certik:
Move `tests/reference/struct_array_01.f90` into `tests/`, and register it in `tests/tests.toml` and regenerate the tests.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4268,57 +4268,105 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            (name_struct != nullptr && strcmp(name_struct->m_name, sym_type->m_name) != 0)) {
+                            diag.add(Diagnostic(
+                                "Array members must be of the same type as the struct",
+                                Level::Error, Stage::Semantic, {
+                                    Label("",{array_init->m_args[0]->base.loc})
+                                }));
+                            throw SemanticAbort();
+                        }
+                    }
+                    // Detect if this is a derived-type array then build ArrayConstructor instead of ArrayConstant
```
certik, line 4319: Why cannot derived types be part of ArrayConstant?

I think for `parameter`, we need to ensure everything is a compile-time constant, and if we use ArrayConstructor, I think that is fundamentally runtime?

opixdown:
@certik ArrayConstructor can still be compile time evaluable, because all of its elements are structconstant values. So even though it not flattened like ArrayConstant, it still represents a fully constant expression.

To add  support directly inside ArrayConstant, we have to redesign how it packs data since structs are not stored like primitive arrays. That would lead to deeper changes in ASR utilities and codegen.

I chose ArrayConstructor because it already supports derived types without needing to change the internal data layout, and the ASR verifier accepts it as the elements are constant.

just let me know what you think is the better pproach.

opixdown:
@certik the CI is failing due to a github network error should I just re run the CI or is this something that needs to be fixed ?

certik:
@opixdown what is the status of this? Let's finish it? The tests passed in the last commit.

PR #8918, opened 2025-11-28
fix: remove redundant visitor in insert_deallocate pass

xaerru:
Instead of inserting `deallocate` for temporary assignment targets at the exits of loops, insert a `deallocate` before the assignment.

After this there is no need for `insert_deallocate` pass, so remove it from default passes.


tests/reference/pass_insert_deallocate-finalize_01-3efebdd.stdout
```diff
@@ -240,6 +251,10 @@
                         ()
                         .false.
                     )
+                    (ImplicitDeallocate
+                        [(Var 5 arr)
+                        (Var 5 str)]
+                    )
```
certik: How did these get in here? I can't figure out the logic. They seem to be inserted by the insert_deallocate pass, but you just removed this pass.

src/libasr/pass/insert_deallocate.cpp
```diff
@@ -288,8 +288,8 @@ class LoopTempVarDeallocateVisitor : public ASR::BaseWalkVisitor<LoopTempVarDeal
 
 void pass_insert_deallocate(Allocator &al, ASR::TranslationUnit_t &unit,
                                 const PassOptions &/*pass_options*/) {
-    // InsertDeallocate v(al);
-    // v.visit_TranslationUnit(unit);
+    InsertDeallocate v(al);
+    v.visit_TranslationUnit(unit);
```
certik: This is not used, so no consequence here, correct?

src/libasr/pass/array_op.cpp
```diff
@@ -1138,6 +1138,21 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+                    ASRUtils::symbol_StorageType((ASR::symbol_t *)target_variable) == ASR::storage_typeType::Default) {
+                    if (std::string(target_variable->m_name).rfind("__libasr_created") != std::string::npos) {
+                        Vec<ASR::expr_t*> v;
+                        v.reserve(al, 1);
+                        v.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, x.m_target->base.loc, (ASR::symbol_t *)target_variable)));
+
+                        pass_result.push_back(al, ASRUtils::STMT(ASR::make_ImplicitDeallocate_t(
+                            al, xx.base.base.loc, v.p, v.size())));
+                    }
+                }
```
certik: This now always calls deallocate before creating any temporary? The logic is that `ImplicitDeallocate` checks if it is allocated, and deallocates only if allocated?

I am not sure if this is the right design.

Let's add an ASR test for this --- check in reference ASR after this pass, and let's look at actual examples, to ensure this is what we want. And if we want it, it will test the correct behavior.

certik:
@assem2002 can you please also review it?

tests/reference/pass_insert_deallocate-finalize_01-3efebdd.stdout
assem2002: I think that reference test is registered like this in `tests.toml` 
```
Insert_deallocate = true
```
Which uses the pass anyways even it's not used by our default pass path.

certik: I see, that would explain it.

How can we test the pass removal above? It seems it made no difference in our tests, so it means we didn't have explicit tests for it.

I would like to figure out how to test that deallocate is inserted correctly.

assem2002: Why it doesn't have major effect on our reference tests ? -- Yes we didn't. why?
because the pass (now) has 2 functionalities : finalization and deallocate before some assignments.
 - finalization part -> turned off in current main -> hence, no test change
 - deallocation before assignment part -> doesn't have reference tests reflecting that part in **current main** -> hence, no test change.

assem2002: How to test changes in this PR?
As this PR is targetting `array_op` pass which depends on temp pass to work first, We probably won't be able to have a reference test. Reference tests for passes just runs the pass without following our default pass path.

We can test it with `llvm=true` if that helps.

src/libasr/pass/array_op.cpp
assem2002: I don't think it happens before every assignment, it only gets inserted with some specific case (debug check inserted for that assignment node + it's a move assignment).
We can consider it a workaround. it's working around a case where happen to have double temporaries for the same expression. Reference test `do_loop_07` -- We create 2 temporaries for one purpose ( one in **temporary** pass, another in **subroutine_from_function** pass)

assem2002:
@xaerru  I believe what you're targeting here is deallocating the temporary created in the `array_struct_temporary` pass (that pass usually add a deallocate statement, but you handling a case where the pass decides no not insert).
Can we deallocate that variable when it's first inserted? seems cleaner and more obvious?

xaerru:
I'll explain why this deallocate is needed:

`integration_tests/do_loop_07.f90` :
```fortran
program do_loop_07
    integer :: i, s
    do i = 1, 3
        if (i == 1) then
            cycle
        end if

        ! There should be an ImplicitDeallocate for __libasr_created_string_format here
        print *, new_array(i)
    end do

    contains
        function new_array(i) result(t)
            integer, intent(in) :: i
            integer, allocatable :: t(:)
            allocate(t(i))

            t = 1
        end function
end program
```
After `--dump-all-passes-fortran`:
```fortran
! Fortran code after applying the pass: array_struct_temporary
program do_loop_07
implicit none
integer(4), dimension(:), allocatable :: __libasr_created_string_format
integer(4) :: i
integer(4) :: s
do i = 1, 3
    if (i == 1) then
        cycle
    end if
    __libasr_created_string_format = new_array(i)
    print *, __libasr_created_string_format
end do

contains

function new_array(i) result(t)
    integer(4), intent(in) :: i
    integer(4), dimension(:), allocatable :: t
    allocate(t(i))
    t = 1
end function new_array

end program do_loop_07
```
```fortran
! Fortran code after applying the pass: array_op
program do_loop_07
implicit none
integer(4), dimension(:), allocatable :: __libasr_created_string_format
integer(4) :: i
integer(4) :: s
do i = 1, 3
    if (i == 1) then
        cycle
    end if
    __libasr_created_string_format = new_array(i)
    print *, __libasr_created_string_format
end do

contains

function new_array(i) result(t)
    integer(4), intent(in) :: i
    integer(4), dimension(:), allocatable :: t
    allocate(t(i))
    do __libasr_index_0_ = lbound(t, 1), ubound(t, 1)
        t(__libasr_index_0_) = 1
    end do
end function new_array

end program do_loop_07
```
```fortran
! Fortran code after applying the pass: subroutine_from_function
program do_loop_07
implicit none
integer(4), dimension(:), allocatable :: __libasr_created__subroutine_from_function_
integer(4), dimension(:), allocatable :: __libasr_created_string_format
integer(4) :: i
integer(4) :: s
do i = 1, 3
    if (i == 1) then
        cycle
    end if
    deallocate(__libasr_created__subroutine_from_function_) ! Implicit deallocate
    call new_array(i, __libasr_created__subroutine_from_function_)
    __libasr_created_string_format = __libasr_created__subroutine_from_function_
    print *, __libasr_created_string_format
end do

contains

subroutine new_array(i, t)
    integer(4), intent(in) :: i
    integer(4), dimension(:), allocatable, intent(out) :: t
    allocate(t(i))
    do __libasr_index_0_ = lbound(t, 1), ubound(t, 1)
        t(__libasr_index_0_) = 1
    end do
end subroutine new_array

end program do_loop_07
```
```fortran
! Fortran code after applying the pass: array_op
program do_loop_07
implicit none
integer(4), dimension(:), allocatable :: __libasr_created__subroutine_from_function_
integer(4), dimension(:), allocatable :: __libasr_created_string_format
integer(4) :: i
integer(4) :: s
do i = 1, 3
    if (i == 1) then
        cycle
    end if
    deallocate(__libasr_created__subroutine_from_function_) ! Implicit deallocate
    call new_array(i, __libasr_created__subroutine_from_function_)
    deallocate(__libasr_created_string_format) ! Implicit deallocate
    ! CheckBounds LHS = __libasr_created_string_format, RHS = __libasr_created__subroutine_from_function_, 
    __libasr_created_string_format = __libasr_created__subroutine_from_function_
    print *, __libasr_created_string_format
end do

contains

subroutine new_array(i, t)
    integer(4), intent(in) :: i
    integer(4), dimension(:), allocatable, intent(out) :: t
    allocate(t(i))
    do __libasr_index_0_ = lbound(t, 1), ubound(t, 1)
        t(__libasr_index_0_) = 1
    end do
end subroutine new_array

end program do_loop_07
```

When the `array_op` pass is applied for the second time, the `deallocate` above the `! CheckBounds` comment is being inserted. If it is not inserted the temporary `__libasr_created_string_format` would be allocated from the previous loop iteration and `CheckBounds` will fail when it shouldn't.

xaerru:
> Can we deallocate that variable when it's first inserted? seems cleaner and more obvious?

We can try it, need to figure out the conditions.

certik:
I think the implicit deallocate only deallocates if it is allocated. The regular deallocate would give an error (at least it should in our Debug mode).

So this is only needed in a for loop? And it happens only in a for loop? If so, then I think that's fine. The alternative design is to call deallocate at the end of the for loop. But I think for now this is fine.

certik:
> So this is only needed in a for loop? And it happens only in a for loop?

@xaerru please let me know the answers. So that we can get it merged.

xaerru:
> So this is only needed in a for loop? And it happens only in a for loop? If so, then I think that's fine. The alternative design is to call deallocate at the end of the for loop. But I think for now this is fine.

This will insert it even if the assignment is not in a loop.

The deallocation at the end or exit of the loop was implemented in `insert_deallocate` with `LoopTempVarDeallocateVisitor`. This PR is disabling that pass and inserting the deallocation before the assignment for simplicity.

certik:
> This will insert it even if the assignment is not in a loop.

So now before _every_ assignment we insert a deallocate check?

If so, that would negate the work we have been working so hard to achieve, which is no checks like this in Release mode.

xaerru:
A solution is to keep the original visitor from `insert_deallocate` which adds deallocate to the end of the loop and move it to `do_loops` pass.

certik:
Ok, let's keep this PR as a draft for now.

xaerru:
https://github.com/lfortran/lfortran/pull/9039/ added one more visitor `IntentOutDeallocateVisitor`.

Do we still want to remove the `insert_deallocate` pass?

assem2002:
I think, I saw lately that we're using that pass with some cases. I think we'll keep it then.
Okay, so the pass has some code to insert deallocate statements at the end of constructs to finalize the symboltable -- well, that's not needed any more, We should remove that code from the pass if we're gonna keep it in action.

PR #8922, opened 2025-11-28
fix : Resolve Submodule bug which was giving error on repeated runs by adding a small check before adding submodule in Symbol Table

### LINKED ISSUE #8875: submodule function with no result clause
This version of the program in #8857 has no result clause but also gives an ICE with lfortran.
```
! OK with gfortran or ifx but not lfortran
module the_interface
  implicit none
  integer,parameter:: dp = selected_real_kind(15)
  interface
     real(dp) module function pi()
     end function pi
  end interface
end module the_interface

submodule (the_interface) the_calculation
  implicit none
contains
  real(dp) module function pi()
    pi = acos(-1.0_dp)
  end function pi
end submodule the_calculation

program main
  use the_interface
  implicit none
  print *, "pi is ", pi()
end program main
```
Output from lfortran 0.58.0 (bad)
```
LCompilersException: read_string: String is too short for deserialization.
```
Output from gfortran (good)
```
 pi is    3.1415926535897931
```

AnshMehta1:
Fixes #8875

Now, we can run the code multiple times without getting error reported in https://github.com/lfortran/lfortran/issues/8875#issuecomment-3550845990 with deleting any `.mod/.smod` files.


certik (approved):

certik:
Awesome, thanks for fixing it!

PR #8924, opened 2025-11-28
Implicit Interface Fixes and LAPACK_02 Enablement

krystophny:
Summary
- Extend @certik’s implicit-interface work from #8914 and finish LLVM11 support for LAPACK-style sequence association in `integration_tests/lapack_02.f90`.
- Keep the existing Source-ABI descriptor-array semantics for unresolved implicit interfaces, while aligning legacy implicit interfaces with the actual implementation ABI where possible.

Scope
- This PR is **dependent on** #8927 (`fix/modfile-location-offset`) and is intended to be merged **after** that one. #8927 fixes the empty-file/class-63 modfile issues; this PR focuses only on the implicit-interface + LAPACK_02 side.
- No changes to the modfile reader or the empty-file handling are included here; those remain in #8927.

Details
1. Semantics: implicit interfaces for legacy array sections
- File: `src/lfortran/semantics/ast_common_visitor.h`
- In `create_implicit_interface_function`, for `ArrayItem` arguments under `legacy_array_sections`:
  - Resolve the corresponding implementation `Function_t` in an enclosing scope (if it exists).
  - When the implementation dummy is an array, extract its `Array_t::m_physical_type` (DescriptorArray vs PointerArray).
  - Rebuild the implicit-interface dummy array type using that same physical type instead of always forcing a descriptor array.
- Effect: when the implementation uses `PointerArray` (classic Fortran sequence-association ABI), the implicit interface follows that physical representation. For unresolved top-level implicit interfaces, the previous behavior (Source ABI + descriptor arrays) is preserved.

2. Codegen: LLVM11-safe lowering for pointer-array implementations
- File: `src/libasr/codegen/asr_to_llvm.cpp`
- In `convert_call_args<T>`:
  - When a call goes through an interface symbol, walk up the symbol tables to find a corresponding implementation `Function_t` with the same name and `deftype == Implementation`. Use that symbol for ABI decisions (`orig_arg`, `x_abi`, etc.). This is what ties the implicit interface for `sub` in `integration_tests/lapack_02.f90` to the real implementation.
  - For `ArrayPhysicalCast` arguments whose implementation dummies are `PointerArray`:
    - Lower the inner argument (`arr_cast->m_arg`) to LLVM IR.
    - Use the existing `visit_ArrayPhysicalCastUtil` and descriptor helpers to convert fixed/descriptor/pointer arrays into the pointer-array ABI (e.g., `float*`).
    - This ensures that LLVM11 sees calls of the form `call void @sub(float*, float*)` instead of mixing `%array*` descriptors with raw pointers, avoiding verifier failures.
  - As a narrow fallback, if an argument for a `PointerArray` dummy still has a descriptor physical type at the call site, derive the data pointer + offset using the descriptor API (`get_pointer_to_data` / `get_offset`) and pass that pointer instead.
- All descriptor ↔ pointer conversions go through the centralized descriptor utilities (`llvm_array_utils`); there are no hard-coded LAPACK names, and no ad-hoc IR bitcasts based on raw LLVM structures.

3. Tests and references (LLVM11 as the reference backend)
- LLVM11 (`lf-llvm11` environment, `build-llvm11`):
  - `integration_tests/lapack_02.f90` now compiles and runs cleanly using the CI-style CMake/CTest harness.
  - ASR reference tests for:
    - `tests/implicit_interface5.f90`
    - `tests/implicit_interface_allocatable_array.f90`
    - `tests/variable1.f90`
    have been **regenerated using the LLVM11 build only**, so that the reference ASR matches the updated implicit-interface semantics.
- LLVM21 (system build, `build/`):
  - `integration_tests/lapack_02.f90` also passes using a separate out-of-tree build directory and its own `build-lfortran-llvm` tree.

4. Tester tooling improvement
- File: `src/libasr/compiler_tester/tester.py`
- Behavioral tweak to `--update` semantics:
  - Previously, `./run_tests.py -u -t <some-tests>` would first remove all `tests/reference/*` and then regenerate only the selected tests, unintentionally wiping unrelated references.
  - Now, the full wipe (`rm -rf tests/reference/*`) happens only when updating the whole suite (no test/backend filters). When `-t` or backend filters are used, only the selected tests’ references are updated.
- This change was used to regenerate the implicit-interface ASR references with LLVM11 without touching other reference files.

Verification
- LLVM11 and LLVM21 both build successfully.
- `integration_tests/lapack_02.f90` passes under both LLVM11 and LLVM21 with the updated lowering.
- The regenerated implicit-interface ASR reference tests pass with the LLVM11 build, and the same ASR behavior is accepted by the LLVM21 build as well.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,15 +23,11 @@ reference how to contribute to the project.
   - `cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON -DWITH_STACKTRACE=yes`
   - `cmake --build build -j`
 - Release build: `cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DWITH_LLVM=ON`
-- Tests: `./run_tests.py` (compiler); LLVM integration via CMake/CTest:
-  - `cd integration_tests && rm -rf build-lfortran-llvm && mkdir build-lfortran-llvm && cd build-lfortran-llvm`
-  - `cmake -DLFORTRAN_BACKEND=llvm -DCMAKE_Fortran_COMPILER=../../build/src/bin/lfortran -DCMAKE_Fortran_FLAGS_DEBUG=\"\" -DCURRENT_BINARY_DIR=. ..`
-  - `make -j$(nproc)`
-  - `LFORTRAN_TEST_ENV_VAR=\"STATUS OK!\" ctest -L llvm -j$(nproc)`
+- Tests: `./run_tests.py` (compiler); `cd integration_tests && ./run_tests.py -j16`
 - Local integration tip (LLVM): some evaluator/diagnostic paths require precise
   location info. When running integration tests locally with the LLVM backend,
   prefer injecting the flag via environment rather than changing CMake:
-  - `cd integration_tests && FFLAGS=\"--debug-with-line-column\" ./run_tests.py -b llvm`
+  - `cd integration_tests && FFLAGS="--debug-with-line-column" ./run_tests.py -j8`
   This mirrors CI closely while enabling line/column emission.
 
 ## Quick Smoke Test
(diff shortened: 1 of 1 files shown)


krystophny:
@certik the quick checks have passed now. Can you turn on exhaustive checks and take a look? If all goes well, we could merge this after #8927 .

tests/reference/asr-implicit_interface5-fbe2fe9.stdout
```diff
@@ -80,7 +80,7 @@
                                                         PointerArray
                                                     )
                                                     ()
-                                                    BindC
+                                                    Source
```
certik, line 83: Are you sure the implicit interface should be Source and not BindC? I think all implicit interfaces must be BindC, since we don't have the source code and we can't distinguish between arrays and scalars.

krystophny: Not really. Let's wait with that one too, I'll check further.

certik:
The CI has been really unstable the whole day today.

Regarding the changes, they might be good, but we'll have to split this PR into several and not lump unrelated changes together. For example Lapack doesn't use modules, so the modfile changes should not trigger for lapack? So those fixes should go into a separate PR. Etc. The Source->BindC change is worrying, but perhaps it's correct.

The integration tests should be tidied up, etc. 

But overall since tests passed, I think this is on a good way.

PR #8927, opened 2025-11-28
Fix modfile location offset causing crash with --cpp flag

krystophny:
## Summary

When loading modfiles, the `LocationManager` was not adjusting position offsets for the `out_start` vector. While `file_ends` was correctly offset, `out_start` values were appended without adjustment, causing `output_to_input_pos()` to compute invalid bisection results and access out-of-bounds indices.

This manifested as a crash (assertion failure in `std::vector::operator[]`) when compiling files that import modules built with the `--cpp` flag.

## Changes

- Offset `out_start` values by the current file position when loading modfile LocationManager data

## How to reproduce manually

```bash
# Build the compiler
cmake --build build -j

# Create test files
mkdir -p /tmp/modtest && cd /tmp/modtest
cat > mod.f90 << 'FEOF'
module mod
   type, abstract :: AbsType
   end type AbsType
   type :: MyType
      class(AbsType), allocatable :: arr(:)
   contains
      procedure :: method
   end type MyType
contains
   subroutine method(self)
      class(MyType), intent(inout) :: self
      associate ( element => self%arr(size(self%arr)) )
      end associate
   end subroutine method
end module
FEOF

cat > main.f90 << 'FEOF'
program main
   use mod
end program
FEOF

# Build module with --cpp (creates modfile)
/path/to/lfortran --cpp -c mod.f90

# On main branch: crashes with std::vector assertion
# With this fix: compiles successfully
/path/to/lfortran --cpp main.f90 mod.o
```

## Test plan

- [x] `class_63` integration test passes (uses `--cpp` with modfiles)
- [x] Full integration test suite passes on LLVM 21
- [x] Full integration test suite passes on LLVM 11

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -462,22 +462,29 @@ def single_test(test: Dict, verbose: bool, no_llvm: bool, skip_run_with_dbg: boo
                 extra_args)
 
     if show_errors:
+        skip_test = False
         for extrafile in extrafiles:
             extrafile_ = extrafile.rstrip().lstrip()
+
+            if no_llvm and len(extrafile_) > 0:
+                log.info(f"{filename} * show_errors   SKIPPED because LLVM is not enabled")
+                skip_test = True
+                break
 
             if len(extrafile_) > 0:
                 extrafile_ = os.path.join("tests", extrafile_)
                 modfile = extrafile_[:-4] + ".mod"
                 if not os.path.exists(modfile):
                     run_cmd("lfortran {} -c {}".format(options, extrafile_))
-        run_test(
-            filename,
-            "asr",
-            "lfortran --show-errors --continue-compilation --no-color {infile}",
-            filename,
-            update_reference,
-            verify_hash,
-            extra_args)
+        if not skip_test:
+            run_test(
+                filename,
+                "asr",
+                "lfortran --show-errors --continue-compilation --no-color {infile}",
+                filename,
+                update_reference,
+                verify_hash,
+                extra_args)
 
     if syntax_only_cc:
         run_test(filename, "ast", "lfortran --continue-compilation --show-ast --no-color {infile}",


certik:
I think this is a good fix, however we really need to figure out a way to test this. This only manifests if we use `--cpp`? (the `--cpp` option uses one more indirection in the location manager). I would think this would manifest for showing any error message or a warning from another module.

Can you show me how I can reproduce this bug and to check that it's fixed, manually?

Let's then think how to test this at the CI. It might be as simple as showing an error message (which we already have tests for) and just ensuring the source is in another module, compiled separately, possibly enabling `--cpp`. We should set this up in subsequent PRs. For now, just show me how to test this by hand, I'll verify it and we can merge it.

certik:
Currently the following test failed:
```
HEAD is now at 8c19157 revert : sim association in select type
++ fpm --compiler=/home/runner/work/lfortran/lfortran/src/bin/lfortran build --flag '--cpp --realloc-lhs-arrays'
 + mkdir -p build/dependencies
hint: Using 'master' as the name for the initial branch. This default branch name
Initialized empty Git repository in /tmp/tmp.IsPqdMhfPB/assert/toml-f/build/dependencies/test-drive/.git/
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
hint:
hint: Disable this message with "git config set advice.defaultBranchName false"
From https://github.com/fortran-lang/test-drive
 * branch            HEAD       -> FETCH_HEAD
[  0%]                    version.f90
[  2%]                    version.f90  done.
[  2%]                  constants.f90
[  4%]                  constants.f90  done.
[  4%]                      token.f90
[  7%]                      token.f90  done.
[  7%]                      error.f90
[  9%]                      error.f90  done.
[  9%]                   datetime.f90
[ 11%]                   datetime.f90  done.
[ 11%]                         io.f90
[ 14%]                         io.f90  done.
[ 14%]                      utils.f90
[ 16%]                      utils.f90  done.
[ 16%]                        abc.f90
[ 19%]                        abc.f90  done.
[ 19%]                   terminal.f90
[ 21%]                   terminal.f90  done.
[ 21%]                      value.f90
[ 23%]                      value.f90  done.
[ 23%]                 json_lexer.f90
[ 26%]                 json_lexer.f90  done.
[ 26%]                 diagnostic.f90
[ 28%]                 diagnostic.f90  done.
[ 28%]                       list.f90
[ 30%]                       list.f90  done.
[ 30%]                        map.f90
[ 33%]                        map.f90  done.
[ 33%]                       node.f90
[ 35%]                       node.f90  done.
[ 35%]                       sort.f90
[ 38%]                       sort.f90  done.
[ 38%]                     keyval.f90
[ 40%]                     keyval.f90  done.
[ 40%]                ordered_map.f90
[ 42%]                ordered_map.f90  done.
[ 42%]                 array_list.f90
[ 45%]                 array_list.f90  done.
[ 45%]                    context.f90
[ 47%]                    context.f90  done.
[ 47%]                  structure.f90
[ 50%]                  structure.f90  done.
[ 50%]                      lexer.f90
[ 52%]                      lexer.f90  done.
[ 52%]                      array.f90
[ 54%]                      array.f90  done.
[ 54%]                      table.f90
[ 57%]                      table.f90  done.
[ 57%]                       type.f90
[ 59%]                       type.f90  done.
[ 59%]                        ser.f90
[ 61%]                        ser.f90  done.
[ 61%]                     parser.f90
[ 64%]                     parser.f90  done.
[ 64%]                     keyval.f90
[ 66%]                     keyval.f90  done.
[ 66%]                      merge.f90
[ 69%]                      merge.f90  done.
[ 69%]                   json_ser.f90
[ 71%]                   json_ser.f90  done.
[ 71%]                         de.f90
[ 73%]                         de.f90  done.
[ 73%]                      array.f90
[ 76%]                      array.f90  done.
[ 76%]                      table.f90
[ 78%]                      table.f90  done.
[ 78%]                       path.f90
[ 80%]                       path.f90  done.
[ 80%]                      build.f90
[ 83%]                      build.f90  done.
[ 83%]                      tomlf.f90
[ 85%]                      tomlf.f90  done.
[ 85%]                json_parser.f90
[ 88%]                json_parser.f90  done.
[ 88%]                    libtoml-f.a
[ 90%]                    libtoml-f.a  done.
[ 90%]                  json2toml.f90
[ 92%]                  json2toml.f90  done.

<ERROR> Compilation failed for object " test_compliance_json2toml.f90.o "
STOP 1
```

Which might be caused by this PR. I enabled all tests, let's see if other tests fail too.

krystophny:
Heads up: I opened a draft repro PR (https://github.com/lfortran/lfortran/pull/8928) that adds a minimal test (mod_cpp_modfile_crash) showing the modfile/preprocessor crash when modules are built with --cpp. It crashes on current main; fix is on branch fix/modfile-location-offset.

krystophny:
@certik I've simplified the fix significantly - it's now just 6 lines that offset `out_start` positions when loading modfile LocationManager data.

## Manual reproduction

```bash
# Build lfortran
cmake --build build -j

# Create minimal repro
mkdir -p /tmp/modtest && cd /tmp/modtest

# Module with polymorphic array in associate block
cat > mod.f90 << 'FEOF'
module mod
   type, abstract :: AbsType
   end type AbsType
   type :: MyType
      class(AbsType), allocatable :: arr(:)
   contains
      procedure :: method
   end type MyType
contains
   subroutine method(self)
      class(MyType), intent(inout) :: self
      associate ( element => self%arr(size(self%arr)) )
      end associate
   end subroutine method
end module
FEOF

cat > main.f90 << 'FEOF'
program main
   use mod
end program
FEOF

# Build module with --cpp (creates modfile with preprocessor=true)
build/src/bin/lfortran --cpp -c mod.f90

# On main branch: crashes with std::vector operator[] assertion
# With this fix: compiles successfully  
build/src/bin/lfortran --cpp main.f90 mod.o -o main && ./main
```

The existing `class_63` test already covers this case (multi-file compilation with `--cpp`), so no new test is needed.

krystophny:
Interesting... the reproducer also doesn't fail in CI/CD on GitHub before the fix. Maybe it's a local issue (Arch/CachyOS with newest GCC). Edit: Confirmed, it is a hidden bug that is exposed by stricter checks there. See latest commit and PR https://github.com/lfortran/lfortran/pull/8928 (now already merged into this branch here too)

krystophny:
@certik sorry for the noise, this one here should be ready to merge.

integration_tests/CMakeLists.txt
```diff
@@ -2030,6 +2030,16 @@ RUN(NAME class_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
...
+# Reproducer for modfile preprocessor interval bug: compile module with --cpp,
+# then compile a separate main that uses the generated modfile.
+# Only run with lfortran (not gfortran) since it tests lfortran-specific modfile handling.
+if(NOT CMAKE_Fortran_COMPILER MATCHES "gfortran")
+    add_test(NAME mod_cpp_modfile_crash
+        COMMAND bash -lc
+        "${CMAKE_Fortran_COMPILER} --cpp -J ${CMAKE_CURRENT_BINARY_DIR} -I ${CMAKE_CURRENT_BINARY_DIR} -c ${CMAKE_CURRENT_SOURCE_DIR}/mod_cpp_mod.f90 && ${CMAKE_Fortran_COMPILER} --cpp --realloc-lhs-arrays -J ${CMAKE_CURRENT_BINARY_DIR} -I ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/mod_cpp_main.f90 ${CMAKE_CURRENT_BINARY_DIR}/mod_cpp_mod.o -o ${CMAKE_CURRENT_BINARY_DIR}/mod_cpp_main.exe"
+    )
+    set_tests_properties(mod_cpp_modfile_crash PROPERTIES LABELS "llvm")
+endif()
```
certik: It should be possible to just do something like:
```
RUN(NAME XXX LABELS llvm EXTRA_ARGS -c --cpp EXTRAFILES class_63_module_1.f90)
```

krystophny: True, this was just needed before we added the CXX assert flag to CI. Now I have even removed it because class_63 test triggers the bug anyway. (see repro branch, I cherry picked and will close the repro after this here is merged)

krystophny: Btw, the last test failed due to download timeout, can you restart it?

krystophny: > Btw, the last test failed due to download timeout, can you restart it?

... already done by rebase. All tests passed before, there was just some flakiness in micromamba download.

certik: Hold on --- how does `class_63` reproduce this bug? It is already enabled in main and it is passing.

certik: Ah I see --- due to `CXXFLAGS="-Werror -D_GLIBCXX_ASSERTIONS"`.

krystophny: Now it is not reached since `empty.f90` fails earlier in the repro branch. But this is also fixed here in addition, so all passes.

src/libasr/string_utils.cpp
```diff
@@ -139,11 +139,16 @@ bool read_file(const std::string &filename, std::string &text)
...
+    if (size == 0) {
+        text.clear();
+        return true;
+    }
+
+    std::vector<char> bytes(size);
+    ifs.read(bytes.data(), filesize);
 
-    text = std::string(&bytes[0], filesize);
+    text.assign(bytes.data(), size);
```
certik, line 151: Do you know why this change is needed --- was the previous code incorrect? It seems it's an equivalent rewrite?

krystophny: This was needed to get empty files to pass. Maybe it is too complicated, I'll double check now...

certik: Wouldn't this do the job:
```diff
--- a/src/libasr/string_utils.cpp
+++ b/src/libasr/string_utils.cpp
@@ -140,7 +140,10 @@ bool read_file(const std::string &filename, std::string &text)
     ifs.seekg(0, std::ios::beg);

     std::vector<char> bytes(filesize);
-    if (filesize == 0) bytes.reserve(1);
+    if (filesize == 0) {
+        text = "";
+        return true;
+    }
     ifs.read(&bytes[0], filesize);

     text = std::string(&bytes[0], filesize);
```

krystophny: OK here is what happened in the old version and why it triggered the error in the strict checks:

- `std::vector<char> bytes(filesize);` creates an empty vector.
- `bytes.reserve(1);` reserves capacity but `size()` stays 0; there is still no element 0.
- `ifs.read(&bytes[0], filesize);` and `std::string(&bytes[0], filesize);` dereference bytes[0] on an empty vector -> undefined behavior.

krystophny:
> If tests pass then I think it's fine to merge.
> 
> I think we still need better tests --- maybe some reference error tests where the errors are coming from a modfile, to ensure the location is correct. But this can be done later.

Great! This micromamba flakiness is still killing a lot of tests though.

certik:
I think we need actual solid tests for this change as I mentioned. For now this is what I did:
```diff
diff --git a/integration_tests/class_63.f90 b/integration_tests/class_63.f90
index a3ef15df1..d4cacd892 100644
--- a/integration_tests/class_63.f90
+++ b/integration_tests/class_63.f90
@@ -1,9 +1,9 @@
 module class_63_module

-   use class_63_module_2, only: MyType
+   use class_63_module_2, only: MyType, AbsType

 end module

 program class_63
-    use class_63_module
-end program
\ No newline at end of file
+    use class_63_module, only: MyType, AbsType
+end program
```
and
```console
$ lfortran --cpp -c class_63_module_1.f90
$ lfortran --cpp -c class_63_module_2.f90
$ lfortran --cpp class_63.f90
$ lfortran --cpp class_63.f90 --lookup-name --line 8 --column 34
[{"kind":6,"location":{"range":{"start":{"character":1,"line":5},"end":{"character":3,"line":3}},"uri":"class_63_module_2.f90"},"name":"mytype","filename":"class_63_module_2.f90"}]
```
Now the `start` is line 5, col 1 (this is incorrect, should be col 4) and `end` is completely incorrect as line 3, col 3, should be line 9 or 10.

In main, this is what I get:
```console
$ lfortran --cpp class_63.f90 --lookup-name --line 8 --column 34
[{"kind":6,"location":{"range":{"start":{"character":4,"line":5},"end":{"character":1,"line":10}},"uri":"class_63_module_2.f90"},"name":"mytype","filename":"class_63_module_2.f90"}]
```
So start is line 5 (col 4), end is line 10 (col 1), which is correct.



So if this fixes some bug, it introduces other bugs.

What we need to do is to add solid testing for these kinds of things. One way we could do it is to dump the full ASR with location information in file,line, column format as text into reference results. Then we regenerate in your PR and see if the changes are correct (that they fix bugs and not introduce other bugs).

I am going to mark this as Draft for now, until we fix the logic.

krystophny:
@certik the problem should be fixed, and a unit test added (needed some adaption of test scripts to support extra files in this context). Can you take a look and see if you can break it again once tests have passed?

certik:
Thanks for fixing it and for adding tests. I'll review it very soon and see if there are any issues.

certik:
I have an idea how to do this --- let's extract the new tests and send them as a separate PR. Then this PR just becomes a simple fix. I can help with this. Keep this PR open and don't touch the branch, it passes all the tests, so we'll use it as a reference.

I am still confused exactly about what bug we are fixing as it seems this PR fixes several bugs:

* empty files (I want to figure out how to test it)
* `--cpp` logic (this should hopefully change the tests that you added in this PR --- we'll add them into main, then this PR changes them).

It could be that we can't add them to main, without the fix in this PR. I'll play with it. I like to go slow with such changes and ensure we have solid testing, so that it doesn't break in the future and also so that the fix is actually correct.

krystophny:
> I have an idea how to do this --- let's extract the new tests and send them as a separate PR. Then this PR just becomes a simple fix. I can help with this. Keep this PR open and don't touch the branch, it passes all the tests, so we'll use it as a reference.
> 
> I am still confused exactly about what bug we are fixing as it seems this PR fixes several bugs:
> 
> * empty files (I want to figure out how to test it)
> * `--cpp` logic (this should hopefully change the tests that you added in this PR --- we'll add them into main, then this PR changes them).
> 
> It could be that we can't add them to main, without the fix in this PR. I'll play with it. I like to go slow with such changes and ensure we have solid testing, so that it doesn't break in the future and also so that the fix is actually correct.

Fine! Don't worry, this is easily automatable. I trigger now the split into various PRs so that you can check individually the new tests and fixes, and a separate PR that will enable the flag in CI. This means, all tests will pass on main as it is right now, and we can decide on the order of the fixes. I think we can merge the `empty.f90` fix more quickly, as it is trivial, and be more careful with the rest.

certik:
Ok, so the empty file fixes and refactoring was merged in:

* https://github.com/lfortran/lfortran/pull/8960
* https://github.com/lfortran/lfortran/pull/8956

And the bounds checking at the CI is enabled in

* https://github.com/lfortran/lfortran/pull/8958 (simple)
* https://github.com/lfortran/lfortran/pull/8961 (comprehensive)

The tests are added in

* https://github.com/lfortran/lfortran/pull/8955

Finally, just the fix for the location information is in:

* https://github.com/lfortran/lfortran/pull/8957

I am merging all the PRs in order, so we'll get everything in.

certik:
We can close this now, all of this PR was submitted as individual PRs, see my previous comment for links.

PR #8928, opened 2025-11-29
Repro: modfile crash with --cpp modfiles

krystophny:
## Summary
- Add minimal integration test `mod_cpp_modfile_crash` that builds a module with `--cpp` and then compiles a user program that uses the generated modfile.
- On current `main`, this test crashes in `LocationManager::output_to_input_pos` (std::vector::operator[] assert) while mapping locations from the preprocessed modfile.
- Purpose: standalone reproducer for the modfile/preprocessor crash reported in #8927.

## How to reproduce manually
```
# Build the compiler
cmake --build build -j$(nproc)

# Configure a fresh integration build dir if needed
cd integration_tests
rm -rf build-lfortran-llvm && mkdir build-lfortran-llvm && cd build-lfortran-llvm
cmake -DLFORTRAN_BACKEND=llvm \
      -DCMAKE_Fortran_COMPILER=../../build/src/bin/lfortran \
      -DCMAKE_Fortran_FLAGS_DEBUG="" \
      -DCURRENT_BINARY_DIR=. ..

# Run just the repro test
ctest -R mod_cpp_modfile_crash -L llvm -V
```
Expected on `main`: abort with std::vector::operator[] from LocationManager. After the fix, it should emit a normal diagnostic and exit nonzero without crashing.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2030,16 +2030,6 @@ RUN(NAME class_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_61 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME class_62 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_63 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS -c --cpp EXTRAFILES class_63_module_1.f90 class_63_module_2.f90)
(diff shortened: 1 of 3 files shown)


krystophny:
Just confirmed that this looks like a hidden bug that the GitHub CI/CD environment cannot capture, but GCC 15 on CachyOS with newer libstdc++ and other compile flags:

  - On current GCC/libstdc++ with _GLIBCXX_ASSERTIONS (e.g., host GCC 15.2.1) mod_cpp_modfile_crash fails with a bounds assertion in LocationManager::output_to_input_pos called from
    LLVMUtils::generate_runtime_error.
  - In CI (ubuntu-22.04 + LLVM11, GCC 11, assertions off) the same test passes, so the bug is masked there.

PR #8930, opened 2025-11-29
refactor: don't use vptr for cstruct type variable

### LINKED ISSUE #8797: Do not insert vtable pointer to structs with no virtual methods
So derived types with just data members and no functions should behave like C structs, no vtable pointer in the instance. All `bind(c)` structs are like that too. And also any struct that has methods which are not virtual. (By default all methods are virtual in Fortran, but you can make them not virtual by the `nonoverridable` keyword).

jinangshah21:
Fixes #8797

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2046,6 +2046,7 @@ RUN(NAME class_76 LABELS gfortran llvm)
 RUN(NAME class_77 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_78 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_STD_F23)
 RUN(NAME class_79 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_80 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME class_procedure_args_01 LABELS gfortran llvm)
 
--- a/integration_tests/class_80.f90
+++ b/integration_tests/class_80.f90
@@ -1,0 +1,25 @@
+module class_80_mod
+   type :: class_value
+    character(len=:), allocatable :: value
+  end type
+  type :: temp_type
+    character(len=:), allocatable :: value
+  end type
+contains 
+  subroutine cast_to_keyval(ptr, val)
+   class(class_value), intent(in) :: ptr(:)
+   type(temp_type), intent(in), optional :: val(:)
+!    val(1)%value = ptr(1)%value   ! TODO 
+end subroutine cast_to_keyval 
+end module
+
+program class_80
+  use class_80_mod
+  type(temp_type), allocatable :: x(:)
+  type(class_value), allocatable :: y(:)
+  allocate(x(1))
+  allocate(y(1))
+  y(1)%value = "Hello World"
+  call cast_to_keyval(y, x)
+!   if (x(1)%value /= "Hello World") error stop  ! TODO
+end program class_80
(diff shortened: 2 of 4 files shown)


certik:
Thank you, great job!

I will review carefully. It would be helpful if you could describe either in the PR or git commit how it was fixed. I am trying to piece it together from the changes themselves.

certik:
It seems in this PR, we now use a "wrapper" that contains a vptr plus a pointer to a struct, while before the vptr was part of the struct. So if virtual methods are needed, this is now strictly slower than Clang does for C++, correct?

If so, I am not sure this is the right design.

How does GFortran handle this?

certik:
It looks like GFortran inserts vptr directly into the payload, like we do in `main`, and for `bind(c)` it uses direct struct (we currently can't do it?).

So it looks like we do not want to use this wrapper approach, it seems it's not the right design?

jinangshah21:
GFortran follows same design as done in this PR. There is cstruct type for `type` variables and wrapper for `class` variables.

certik:
Unless this is only done for cstruct derived types. We should check if those are allowed at all for `class(t)` variables. Possibly the `bind(c)` are only allowed for `type(t)` variables. We could allow any cstruct derived types using this wrapper approach as a `class(t)` variables, and possibly it would be like an optimization --- or a trade off to use for optimization if it is advantagous (slower access, but smaller memory footprint; but faster for `type(t)`, etc.).

certik:
Ok, I see.

Is this correct: this PR makes NO change to derived types that have virtual functions (so vptr is part of the struct in memory), and any derived type with cstruct=true (currently set in AST->ASR for derived types that have no virtual functions as well as `bind(c)`) now uses the wrapper (and no vptr in the same struct in memory).

jinangshah21:
Variables with `cstruct=true` will not have vptr else we will have vptr (i.e. wrapper). It doesn't depend if struct has virtual functions or not. Because for class type variables we always need vptr for runtime copy function.

certik:
Ok, so I think that's not quite the right design yet. What we want for `cstruct=false` is the vptr to be part of the struct, like Clang does for C++, since that's faster -- no need to dereference a pointer to access the data (I think that's the current design in `main`). For `cstruct=true` we want to use the wrapper for `class(t)` parameters, but for `bind(c)` (and likely also `type(t)` parameters) we do not want to use the wrapper at all.

certik:
Summary: in this PR we follow what GFortran does, which is that all classes (with virtual functions) and structs are declared as C structs with no vptr in them directly (inlined) and then all `type(t)` function parameters pass in this C-style struct directly (no access to vptr), and all `class(t)` function parameters pass in a wrapper `<vptr, *dt>` where the vptr is first, and the second is a pointer to the C-style struct to access members. The `class(*)` function parameter also uses this wrapper and works for all intrinsic types also. Even in the main program, if we have any `class` variables, they will contain the vptr via the wrapper. This design is quite optimal, it's what GFortran does. The only downside is when accessing data members of `base` in `class(base)`, we have to dereference a pointer to access them. There is no downside for `type(base)` variables, and no downside when `bind(c)` interoperability with C.

In the current main (before this PR), we always inlined `<vptr, x1, x2>` even for C-style structs with no virtual functions. This design is not optimal, because it's hard to make the structs interoperate with C via `bind(c)`, as well as it is wasteful in memory if there are no virtual functions.

In the future, we can further optimize the last "unoptimal" case of `class(base)` and accessing the data member variables by inlining the vptr as in `<vptr, x1, x2>` for such classes, which will allow to access the`x1`, `x2` data members directly, and we need to handle every corner case correctly.

certik:
TODO for later:

Identify copies of `type` to `class` and separately `class` to `type` and ensure that data members are copied correctly. Ensure we have integration tests for that, if not, add them. Here is an example of `type` to `class` copy:
```fortran
    subroutine copy_obj(a, b)
        class(base), intent(out) :: a
        type(base), intent(in)  :: b
        a = b 
    end subroutine copy_obj
```

Issue #8925, opened 2025-11-28
ICE with READ or WRITE using an assigned format.


w6ws:
Assigned formats are a legacy construct that was deleted as of F95.  Use of them causes the compiler to ICE, whereas it would be preferable to produce a nice error message.
```
$ cat assigned_fmt.f90
program assigned_fmt
  implicit none
  integer :: i

  ASSIGN 0012 TO I
0012 FORMAT (" **** ASSIGN FORMAT NUMBER TO INTEGER VARIABLE ****" )
  WRITE (6, I)

!  READ also causes an ICE
!  assign 13 to i
!  13 format ()
!  read (5, i)

end program
$
$ lfortran --version
LFortran version: 0.58.0-127-gde1c39095
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$
$ lfortran -c assigned_fmt.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1193, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
[...]

PR #8932, opened 2025-11-30
[enh] : Fix ICE with assigned format in READ/WRITE/PRINT

opixdown:
## **Issue Fixed**
#8925  ICE when using assigned format in WRITE/READ/PRINT statements

## **Summary**

This PR fixes an ICE that occurred when using assigned formats integer variables as format specifiers in WRITE, READ, and PRINT statements.

## **Problem**

When compiling code that uses assigned format feature, LFortran would crash with an ICE instead of providing a clear error message, code example:

```
program assigned_fmt
  implicit none
  integer :: i

  ASSIGN 0012 TO I
   0012 FORMAT (" **** ASSIGN FORMAT NUMBER TO INTEGER VARIABLE ****" )
  WRITE (6, I)

end program
```

## **Error Before this PR**

## **Error After this PR**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran -c tests/errors/assigned_fmt_write.f90
semantic error: Assigned format (using integer variable as format specifier) is not supported. Use character variables instead.
 --> tests/errors/assigned_fmt_write.f90:7:3
  |
7 |   WRITE (6, I)
  |   ^^^^^^^^^^^^ 
```
## **Changes**

- Modified `src/lfortran/semantics/ast_body_visitor.cpp` to add assigned format detection
- Added error tests: `tests/errors/assigned_fmt_write.f90`, `assigned_fmt_print.f90`, `assigned_fmt_read.f90`

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -121,7 +121,6 @@ program continue_compilation_1
 
 
 
-
     ! Use the space above to insert new declarations, and remove the line, so
     ! that the lines below do not shift, to keep the diff minimal.
     !
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
-    "infile_hash": "7d2df4774995ef522322595cfeabf2323121e554c38685e55a945175",
+    "infile_hash": "5309205e77f75af6ea136446f8c6c9986c9e85ca37475b18fe2135e3",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "fbb3a6c2b3b93c8b61ef7e56cbf26e94575e450ff3ec587000b27163",
+    "stderr_hash": "0cbcca7b23d3eb859e3942cb6b8aa9a944ee3dccf9775f2cfef717dd",
     "returncode": 1
 }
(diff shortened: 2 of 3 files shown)


tests/errors/assigned_fmt_print.f90
```diff
@@ -0,0 +1,9 @@
+program assigned_fmt_print
+  implicit none
+  integer :: i
+
+  assign 100 to i
+100 format (A)
+  print i, "test"
+
+end program
```
certik: add these to `tests/errors/continue_compilation_1.f90` and regenerate tests using `./run_tests -u`.

tests/errors/continue_compilation_1.f90
```diff
@@ -349,4 +349,38 @@ program continue_compilation_1
     q1: do q1 = 1, 3
         print *, q1
     end do q1
-end program
+end program 
+
+    ! Test assigned format - WRITE
+    !checks for issue #8925
+    program assigned_fmt_write
+    implicit none
```
certik: This wouldn't work --- you are compiling two programs in the same file, I think this is not allowed. That's another bug in fact, we should give an error message. Go ahead and open up an issue for this.

And to finish this PR, put the variables at the top of the program, and leave the statements here, but in the same program.

opixdown: okay i will open a issue for compiling two programs in the same file.

tests/errors/continue_compilation_1.f90
```diff
@@ -117,6 +117,7 @@ program continue_compilation_1
     type(bspline_3d) :: s3_in_program
     integer :: j2, i2, k2(2), x2(2), y2(3)    
     integer::tt = b'01' * 3
+    integer :: fmt_i1, fmt_i2, fmt_i3 ! for issue #8925
```
HarshitaKalani, line 120: Can you remove one line after this addition so that the updated references would look clean?

opixdown: I’ve removed a blank line after the new declarations, as you suggested.
Please let me know if anything else needs adjustment. Thanks!

HarshitaKalani:
Rebase it on top of main. The CI would be working fine.

certik:
The CI has been unstable today, but we just need to keep restarting until it passes. Then we can merge.

PR #8936, opened 2025-11-30
Add lapack_04: reproducer for UnboundedPointerArray lbound

krystophny:
## Summary
Minimal reproducer from LAPACK `stfsm.f` that triggers assertion in `visit_ArrayBound()` when using `lbound` on 0-based assumed-size array (`A(0:*)`).

## Dependency
**This PR depends on #8924** and should be merged after that one.

## Details
- Fails with: `LCOMPILERS_ASSERT` at `asr_to_llvm.cpp:14746`
- The switch in `visit_ArrayBound()` doesn't handle `UnboundedPointerArray` physical type
- Blocks LAPACK build at ~70% (first failure: `stfsm.f`)

## Test
```
lfortran --fixed-form --implicit-interface --legacy-array-sections integration_tests/lapack_04.f90
```


krystophny:
Status: on fix/arraybound-unbounded-pointer-array all LLVM11 unit tests and the full llvm integration suite now pass via the Python harness, and LAPACK (real-only, complex off) builds up to SGESVD where a mis-lowered SLASCL call is the current blocker; I will fix this regression next while keeping the existing integration tests green.

PR #8939, opened 2025-12-02
CI: cd to root `wasm_builds` dir

ubaidsk:
The WASM build deployment script was executing git operations (git init, git add, git commit) from inside the wasm_builds/docs/ directory, which caused docs/ to become the new repository root during the force push, effectively flattening the directory structure on each deployment. This meant that on the next clone, the repository structure had data.json at the root level instead of inside a docs/ subdirectory, but the script would then create a fresh docs/ directory and look for data.json there, never finding the existing file from the previous deployment. The fix moves the git operations to the parent wasm_builds/ directory by adding cd .. before git init, ensuring that docs/ remains as a subdirectory in the committed repository structure, which allows the script to correctly find and update the existing data.json file across deployments.

### DIFF
--- a/ci/upload_lfortran_wasm.sh
+++ b/ci/upload_lfortran_wasm.sh
@@ -48,6 +48,10 @@ cp $D/src/bin/lfortran.data ${dest_dir}/${git_hash}/lfortran.data
 echo "$git_hash" > ${dest_dir}/latest_commit # overwrite the file instead of appending to it
 python $D/ci/wasm_builds_update_json.py ${dest_dir} ${lfortran_version} ${git_hash}
 
+# Move back to wasm_builds/ directory to perform git operations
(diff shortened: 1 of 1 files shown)


ubaidsk:
The CI at [https://github.com/lfortran/wasm_builds](https://github.com/lfortran/wasm_builds) fails currently as there is no `docs/` directory to publish the webpage. Post this PR merge, I will restore the `wasm_builds` repo back using the archived contents from https://github.com/certik/wasm_builds/tree/old-main-3.

certik:
Thanks for fixing it. Go ahead and fix the repository.

ubaidsk:
The `wasm_builds` repo is fixed. I can see that it is working fine. Past couple of workflow runs have passed successfully.

certik:
Awesome, thank you!

Issue #8921, opened 2025-11-28
EQUIVALENCE with two arrays causes SEGV


w6ws:
This may be a variation of PR #5529, #5481, and of course #2523:
```
$ cat two_array_equiv.f90
      program two_array_equiv
      implicit none
!
! This creates a 6 integer structure with 
! ia2(3) and ia2(4) overlapping ('unioned') with ia1(1) and ia1(2)
      integer ia1(4), ia2(4)
      equivalence (ia1, ia2(3))

      logical :: pf

      ia1 = [11, 12, 13, 14]
      ia2 = [ 1,  2,  3,  4]  ! 3 and 4 replace 11 and 12 in ia1
      print *, 'ia1 =', ia1
      print *, 'ia2 =', ia2
      pf = all (ia1 == [3, 4, 13, 14]) .and.  &
           all (ia2 == [1, 2,  3,  4])
      print *, 'test: ', merge ('passed', 'failed', pf)
      end program
$
$ gfortran two_array_equiv.f90
$ ./a.out
 ia1 =           3           4          13          14
 ia2 =           1           2           3           4
 test: passed
$
```
With lfortran a SEGV is produced:
```
$ lfortran --version
LFortran version: 0.58.0-127-gde1c39095
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$ lfortran -c two_array_equiv.f90
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1137, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
[...]

PR #8947, opened 2025-12-02
equivalence with two array

PrakharGoyal28:
#8921 Fixes 

This PR fixes a segmentation fault caused by EQUIVALENCE generating pointer types with nullptr m_type fields.
ASR verification crashed when type_get_past_pointer() dereferenced these null types.

Key fixes:

Added a null check in type_get_past_pointer() to avoid dereferencing nullptr.

Updated EQUIVALENCE handling to only create pointer types when the base type is non-null

Corrected element-type extraction for array variables, ensuring correct type deduction during EQUIVALENCE

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3578,9 +3578,11 @@ public:
                                         "ignored for now"
                                     );
                                 }
-                                type = ASRUtils::make_Array_t_util(al, asr_eq2->base.loc, type, dim2.p, dim2.size(), ASR::abiType::Source, false, ASR::array_physical_typeType::DescriptorArray, false, false);
-                                ASR::ttype_t* ptr = ASRUtils::TYPE(ASR::make_Pointer_t(al, asr_eq2->base.loc, type));
-                                var__->m_type = ptr;
+                                if (type != nullptr) {
+                                    type = ASRUtils::make_Array_t_util(al, asr_eq2->base.loc, type, dim2.p, dim2.size(), ASR::abiType::Source, false, ASR::array_physical_typeType::DescriptorArray, false, false);
+                                    ASR::ttype_t* ptr = ASRUtils::TYPE(ASR::make_Pointer_t(al, asr_eq2->base.loc, type));
+                                    var__->m_type = ptr;
+                                }
 
                                 Vec<ASR::expr_t*> args;
                                 args.reserve(al, arr->n_dims);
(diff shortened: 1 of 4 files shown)


src/libasr/asr_utils.h
```diff
@@ -244,6 +244,9 @@ static inline ASR::ttype_t *type_get_past_pointer(ASR::ttype_t *f)
 {
     if (ASR::is_a<ASR::Pointer_t>(*f)) {
         ASR::Pointer_t *e = ASR::down_cast<ASR::Pointer_t>(f);
+        if (e->m_type == nullptr) {
+            return f;
+        }
```
certik, line 249: When would this happen? I think the `m_type` always has to be a valid type, it shouldn't be nullptr. This change seems to hide a bug.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3620,8 +3625,10 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                             "ignored for now"
                                         );
                                     }
-                                    ASR::ttype_t* ptr = ASRUtils::TYPE(ASR::make_Pointer_t(al, asr_eq2->base.loc, type));
-                                    var__->m_type = ptr;
+                                    if (type != nullptr) {
```
certik, line 3628: The if logic above always assigns to a `type` so that it is not null. So this if statement shouldn't be needed, or it's hiding some bug.

I would determine the `type` explicitly. What condition causes the `type` to be null after the above if statements?

certik: Oh I see --- it's null because equivalence is not implemented.

If you are implementing equivalence, then remove the warning, since it's now implemented.

certik:
The error it's giving is:
```
semantic error: ArrayInitalizer expressions can only be assigned array references
  --> /home/runner/work/lfortran/lfortran/integration_tests/equivalence_11.f90:17:7
   |
17 |       ia1 = [11, 12, 13, 14]
   |       ^^^^^^^^^^^^^^^^^^^^^^ 
```
Not sure what this error means --- we should improve this error message (that's a separate issue).

certik:
I marked it as draft for now. I would recommend to first get all integration tests passing, test it locally by hand first. Make sure the added test actually passes. Figure out what needs to be done (even with hacks) to get this test working and not breaking any other test. 

Once CI passes, we can then see if we can clean up the code, based on my above comments.

You identified the parts of the code that need fixing, but it's not the right fix yet.

PR #8949, opened 2025-12-02
Enh : Introduce StringToArray casting

### LINKED ISSUE #7136: Bug : LFortran doesn't implicitly create an array for strings
```.f90
program prog
  integer :: intt
  intt = ff("12333")
  contains 
  
  function ff(s) result(r)
    character(*) :: s(*)
    integer :: r
    r  = len(s)
  end function
end program 
```

```Console
  %call_arg_value = alloca i8*, align 8
 %array*  %2 = call i32 @ff(i8** %call_arg_value)
```

assem2002:
closes #7136

In case of passing a string to a function expecting an array of strings, We'll do a cast.

What is the length and size of resulting array? let's consider all possible variations (length->exist or not, size->exist or not).
```
STRING PASSED --> `character(20) :: str`
- `character(10) :: arr(2)`   -> array of size=2, length=10 
- `character(10) :: arr(*)`   -> array of size=20/2, length =10
- `character(*) :: arr(2)`    -> array of size = 1, length = 20/2
- `character(*) :: arr(*)`    -> array of size = 1, length = 20
- `character(*) :: arr(n)`    -> array of size = n, length = 20 (undefined behavior)
```
Notice that almost all of them could lead to wrong behavior if length and size don't fit with the whole length of passed string. Rules here a bit vague, but gfortran can do the cast (it's native to string and array of string actual phsyicalTypes when gfortran lowers the code) + some codes use that feature, so we need to support that weird casting .

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1879,6 +1879,7 @@ RUN(NAME string_83 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_84 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_85 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_86 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME string_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/string_87.f90
+++ b/integration_tests/string_87.f90
@@ -1,0 +1,14 @@
+program string_87
+    character(10) :: str
+    str = "HelloWorld"
+    call ff(str)
+    if(str /= "HelloWTrld") error stop 3 
+    contains
+    
+    subroutine ff(s)
+      character(5) :: s(2)
+      if(s(1) /= "Hello") error stop 1
+      if(s(2) /= "World") error stop 2
+      s(2)(2:2) = "T"
+    end subroutine
+end program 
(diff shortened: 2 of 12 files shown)


certik:
Shouldn't this be:
```
- `character(*) :: arr(2)`    -> array of size = 2, length = 20/2
```

certik:
I didn't know you can do this in Fortran. Which codes use this? I think we need to add proper bounds checking for this in Debug mode. What does "undefined behavior" mean? It should give a Debug-time bounds check error, or it should be well-defined, not undefined.

certik:
Let's see if tests pass. The added test only handles one of the cases above --- are all other cases not implemented yet?

assem2002:
replying on comments : 

- What code uses this feature?
fpm does, with C-bind functions. This PR is mainly fixing these commented out fpm lines we have in our fpm-CI repo -- [HERE](https://github.com/jinangshah21/fpm/commit/98d1c06f1b1dfd53caf39c7a0567f5fcaffe533a)

- How bounds checks should operate here?
dimension(*)   --> No checks -- that's the most common way of using that "String to array" cast btw.
dimension(:)    --> no check -- gfortran doesn't allow casting string to that array btw, though LFortran does .
dimension(10) --> compile time checks if string len is compile time -- not checked currently :)
dimension(n)   --> Check maybe needed.


- Does all mentioned 5 possibilities above work?
No. 4/5 work. We need to support the case with `dimension(n)`

a minor notice 
`character(*) :: arr(2)`   acts like this with gfortran and this PR --> `array of size = 2, length = 20 .` That's wrong, I think our bounds checker could catch it if we force the cast node to result in an array of `length 20 but size=1`.
gfortran doesn't raise any errors with `-fcheck=all`.

assem2002:
I'll merge this and add tests in subsequent PR shortly.

certik:
>  dimension(:) --> no check -- gfortran doesn't allow casting string to that array btw, though LFortran does .

We should not either --- we need to be as strict as we can at the beginning. We can always relax later without breaking people's codes, but not the other way round.

I think it has to do with the fact that `dimension(n)` and `(*)` are contiguous and Fortran allows implicit reshaping of arrays, and so array of strings behaves a little bit like that too, so it's allowed, but for `dimension(:)` the array is not contiguous in general and so reshaping must be done explicitly by the user.

Issue #8357, opened 2025-08-14
Design of assumed rank arrays


certik:
We have at least the following two options:

* Runtime: we store the rank and descriptor at runtime. `size(x)` just queries the descriptor. `select rank` queries the descriptor at runtime. For this design we might need to use the same descriptor for all the ranks.
* Compile-time: it would work like a template --- we would need to keep the function as "generic", not generate any code. And then for each call site we would instantiate the function for the specific ranks of all arguments. Possibly caching the instantiation. One specific approach for this case is detailed at https://github.com/lfortran/lfortran/issues/8319#issuecomment-3180401532.

certik:
We need to support global functions in separate compilation. We must be able to generate runtime code that will work. So let's start with the runtime design. The compile-time design can be done later as an optional optimization for cases where we have access to the source code.

For the runtime design to work we need a descriptor like in https://github.com/lfortran/lfortran/issues/1966:
```
! (injected by the Fortran processor)
desc_a%base_addr = c_loc(a)
desc_a%rank = 1
desc_a%type = CFI_INT
desc_a%dim(1)%lb = 1  ! lower bound 
desc_a%dim(1)%size = size(a) ! size
desc_a%dim(1)%sm = 4 ! stride
```
The first element is a pointer to the data, and the second element is a rank. The rank then determines how many dimension fields are present.

So it looks like the steps to implement this are:

* [x] Ensure the first element is a pointer
* [x] Add rank as the second element into each descriptor. Update: the rank is the last element.
* [x] Then implement `size` of assumed rank array by using the rank and the runtime descriptor
* [x] Implement `select rank` by dispatching on `rank` at runtime

HarshilShah1804:
Assumed-rank arrays only support a limited set of operations directly, namely the inquiry intrinsics `size`, `shape`, `lbound`, `ubound`, `rank`, `allocated`, `associated`, and `present`. For any other operation, or when calling intrinsic functions that require knowledge of the actual rank, a select rank block must be used to resolve the rank at runtime.

This is what GFortran does:
It creates functions similar to what we do for instantiating intrinsic functions for all these shape and then use the required information from the array descriptor to perform the operations of shape, size, etc. like this (for size):
```cpp
size.1 = 1;
idx.3 = 0; 
goto L.2; 
L.1:; 
if (x->dtype.rank + -1 == idx.3 && x->dim[idx.3].ubound == -1) { 
    extent.2 = -1;
} else {
[...]

PR #8950, opened 2025-12-02
fix: cast assumed rank arrays to descriptor of appropriate rank before using in reshape

HarshilShah1804:

### DIFF
--- a/integration_tests/select_rank_05.f90
+++ b/integration_tests/select_rank_05.f90
@@ -1,0 +1,30 @@
+program select_rank_04
+    implicit none
+
+    integer :: a(2) = [4, 2]
+    integer :: b(2, 3) = reshape([1, 2, 3, 4, 5, 6], [2, 3])
+    
+    call check(a)
+    call check(b)
+contains
+
+    subroutine check(x)
+        integer, intent(in) :: x(..)
+        integer :: l1(1)
+        integer :: l2(2)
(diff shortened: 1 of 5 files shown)


HarshilShah1804:
@certik I don't think we require exhaustive checks for this PR. What do you suggest ?

integration_tests/select_rank_05.f90
```diff
@@ -0,0 +1,30 @@
...
+            print *, c
+            if (any(c /= reshape([4, 2], [2, 1]))) error stop
+        rank (2)
+            d = reshape(x, [3, 2])
+            print *, d
+            if (any(d /= reshape([1, 2, 3, 4, 5, 6], [3, 2]))) error stop
+        end select
+    end subroutine check
+
+end program select_rank_04
```
certik, line 30: You need to register this test in cmake for it to run.

certik:
Yes, likely full tests are not needed, but you need to fix https://github.com/lfortran/lfortran/pull/8950/files#r2582599356 first.

integration_tests/select_rank_05.f90
HarshilShah1804: Oh yes. I forgot to add it. I have added it now. You can review.

certik (approved):
I think this is fine to merge, thanks!

Issue #8937, opened 2025-12-01
[enh] : Reject multiple main programs in the same source file


opixdown:
### **Description** ### 
LFortran incorrectly allows compiling multiple program units inside one source file. According to the standard, a .f90 file may contain only one PROGRAM.

### MRE
```
! First program
program A
    print *, "Program A"
end program A

! Second program
program B
    print *, "Program B"
end program B
```
### Current Behaviour 

```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test01.f90 
Program A
```
### Expected Behaviour 
LFortran should emit a clear semantic error such as:
```
error: Multiple program units are not allowed in the same source file
```

harperjf:
That's not quite the right error message because one may compile two or more program units together in different source files. F2023 3.112 defines "program" as

set of Fortran program units (3.113) and entities defined by means other than Fortran that includes exactly one main program (3.97)

How about 
```error: more than one main program found```

harperjf:
#8954 and #8937 use a term defined in the F2023 standard in a nonstandard way, and my comment on #8937 seems not to have been understood. There is some confusion about "main program" and "program unit".

A program unit (3.113) is a main program, external subprogram, module, submodule, or block data program unit. Suppose that file ```a.f90``` consists of a module called ```a``` and a main program using it, that ```b.f90``` is another copy of module ```a```, and that ```c.f90``` is a main program using the module ```a```, and that ```a.f90``` is good Fortran if compiled alone, and ```b.f90``` and ```c.f90``` are good Fortran if compiled together.

"Reject multiple PROGRAM units in the same source file" would appear to reject
```lfortran a.f90``` because it contains two program units. (Recall that PROGRAM and program are equivalent in Fortran.) On the other hand
```lfortran a.f90 c.f90``` should be rejected because it has two main programs even though they are not in the same source file.

PR #8954, opened 2025-12-02
Reject multiple main programs in the same source file

opixdown:
## **Issue Fixed**
#8937 Reject multiple main programs in the same source file

## **Summary**

This PR rejects multiple main programs within a single source file.

## **Problem**

LFortran incorrectly allowed compiling multiple program units inside one source file. According to the standard, a `.f90` file may contain only one program.

Code example:
```
! multi.f90
program A
  print *, "Program A"
end program A

program B
  print *, "Program B"
end program B
```

```
$ lfortran multi.f90
Program A  # Incorrectly compiles and runs only the first program
```

```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran multi.f90 
semantic error: Multiple program units are not allowed in the same source file
 --> multi.f90:2:1 - 4:13
  |
2 |    program A
  |    ^^^^^^^^^...
...
  |
4 |    end program A
  | ...^^^^^^^^^^^^^ first program unit defined here
  |
7 |    program B
  |    ^^^^^^^^^...
...
  |
9 |    end program B
  | ...^^^^^^^^^^^^^ second program unit defined here

```
## **Changes**

- Modified `src/lfortran/semantics/ast_symboltable_visitor.cpp`

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -425,7 +425,7 @@ public:
             first_program_loc = x.base.base.loc;
         } else if (program_count >= 2) {
             diag.add(diag::Diagnostic(
-            "Multiple main programs in the same file",
+            "More than one main program found",
                 diag::Level::Error, diag::Stage::Semantic, {
                     diag::Label("first main program defined here", {first_program_loc}),
                     diag::Label("second main program defined here", {x.base.base.loc})
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "ffac7e2f092abc94903b2ee9b031243c58538b34e74ec89eab17c7e4",
+    "stderr_hash": "bc8780201e40bff1c68d77882dd43f94fcf38fc62b6845fa00cd39f6",
     "returncode": 1
 }
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -129,7 +129,7 @@ semantic error: BOZ literal constant cannot be used in binary operations
 119 |     integer::tt = b'01' * 3
     |                   ^^^^^^^^^ 
 
-semantic error: Multiple main programs in the same file
+semantic error: More than one main program found
    --> tests/errors/continue_compilation_1.f90:64:1 - 367:11
     |
  64 |    program continue_compilation_1


opixdown:
@certik Where should I add the test case for this semantic error check? Should it go in `tests/errors/`  or elsewhere?

certik:
Try to add it to `tests/errors/continue_compilation_1.f90`, so that we test it with other errors also.

certik:
It doesn't work, you committed a stacktrace instead of errors (correct output):
```
< Internal Compiler Error: Unhandled exception
< Traceback (most recent call last):
<   File "$DIR/src/bin/lfortran.cpp", line 2737
<     LCompilers::print_stack_on_segfault();
<   File "$DIR/src/bin/lfortran.cpp", line 2540
<     return run_parser_and_semantics(opts.arg_file, compiler_options);
<   File "$DIR/src/bin/lfortran.cpp", line 727
<     bool has_error_w_cc = compiler_options.continue_compilation && diagnostics.has_error();
<   File "$DIR/src/lfortran/fortran_evaluator.cpp", line 276
<     Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
<   File "$DIR/src/lfortran/fortran_evaluator.cpp", line 298
<     compiler_options.symtab_only, compiler_options, lm);
<   File "$DIR/src/lfortran/semantics/ast_to_asr.cpp", line 112
<     if (res.ok) {
<   File "$DIR/src/lfortran/semantics/ast_body_visitor.cpp", line 6028
<     b.is_body_visitor = true;
<   File "$DIR/src/lfortran/semantics/ast_body_visitor.cpp", line 175
<     visit_ast(*x.m_items[i]);
<   File "$DIR/src/lfortran/ast.h", line 5018
<     void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
<   File "$DIR/src/lfortran/ast.h", line 4973
<     case astType::unit: { v.visit_unit((const unit_t &)x); return; }
<   File "$DIR/src/lfortran/ast.h", line 5021
<     void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
<   File "$DIR/src/lfortran/ast.h", line 4595
<     case modType::BlockData: { v.visit_BlockData((const BlockData_t &)x); return; }
<   File "$DIR/src/lfortran/semantics/ast_body_visitor.cpp", line 2438
<     ASR::symbol_t *t = current_scope->get_symbol(to_lower(x.m_name));
<   File "$DIR/src/libasr/asr.h", line 40
<     LCOMPILERS_ASSERT(f != nullptr);
< AssertFailed: f != nullptr
```

opixdown:
Hi @certik, I tried both approaches:
	•	Keeping throw SemanticAbort() : the test still produces an ICE.
	•	Replacing it with return : the compiler does not emit the “multiple main program” error anymore.

So I’m not sure what the correct approach should be here.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -416,6 +419,20 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            first_program_loc = x.base.base.loc;
+        } else if (program_count >= 2) {
+            diag.add(diag::Diagnostic(
+            "Multiple main programs in the same file",
+                diag::Level::Error, diag::Stage::Semantic, {
+                    diag::Label("first main program defined here", {first_program_loc}),
+                    diag::Label("second main program defined here", {x.base.base.loc})
+                }
+            ));
+            return;
```
assem2002, line 434:
```suggestion
            
```

certik:
The correct approach in statements is to use SemanticAbort and we then catch it and skip the statements.

At the program level we probably don't catch the SemanticAbort, so you need to figure out the logic of the visitor and ensure we just skip the second program (after giving the diagnostic error), so that we can recover from this error correctly.

opixdown:
> The correct approach in statements is to use SemanticAbort and we then catch it and skip the statements.
> 
> At the program level we probably don't catch the SemanticAbort, so you need to figure out the logic of the visitor and ensure we just skip the second program (after giving the diagnostic error), so that we can recover from this error correctly.

Thanks for the clarification ondrej.

harperjf:
#8954 and #8937 use a term defined in the F2023 standard in a nonstandard way, but my comment on #8937 seems not to have been understood.

A program unit (3.113) is a main program, external subprogram, module, submodule, or block data program unit. Suppose that file ```a.f90``` consists of a module called ```a``` and a main program using it, that ```b.f90``` is another copy of module ```a```, and that ```c.f90``` is a main program using the module ```a```, and that ```a.f90``` is good Fortran if compiled alone, and ```b.f90``` and ```c.f90``` are good Fortran if compiled together.

"Reject multiple PROGRAM units in the same source file" would appear to reject
```lfortran a.f90``` because it contains two program units. (Recall that PROGRAM and program are equivalent in Fortran.) On the other hand
```lfortran a.f90 c.f90``` should be rejected because it has two main programs even though they are not in the same source file.

opixdown:
@harperjf Thanks for the explanation! I updated the error message to refer to “main programs” instead of “program units” and kept the rest of the logic the same.

harperjf:
Thank you @opixdown. But there's still a problem with "same source file".
What should be forbidden is 2 or more main programs being accessible to a 
compiler at the same time, not how many source files they are in. (I note 
that the Fortran standard does not define "source file" and mentions one 
only as "something.h" in the context of interoperability with C.)

On Wed, 3 Dec 2025, opixdown wrote:

> Date: Wed, 3 Dec 2025 21:35:12 +0000
> From: opixdown ***@***.***>
> Reply-To: lfortran/lfortran
>     ***@***.***>
> To: lfortran/lfortran ***@***.***>
> Cc: harperjf ***@***.***>, Mention ***@***.***>
> Subject: Re: [lfortran/lfortran] Reject multiple main programs in the same
>     source file (PR #8954)
> 
> opixdown left a comment (lfortran/lfortran#8954)
>
> @harperjf Thanks for the explanation! I updated the error message to refer to “main programs” instead of “program units” and kept the rest of the logic the same.
>
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/lfortran/lfortran/pull/8954#issuecomment-3608934701
> You are receiving this because you were mentioned.
>
> Message ID: ***@***.***>


-- John Harper, School of Mathematics and Statistics
Victoria Univ. of Wellington, PO Box 600, Wellington 6140, New Zealand.
e-mail ***@***.***

certik:
Maybe instead of `semantic error: Multiple main programs in the same file`, it could say `semantic error: Multiple main programs in the same translation unit`, that way it covers both the same file, as well as multiple files accessible to the compiler at the same time.

Even better:

`semantic error: more than one main program found`

As suggested by @harperjf in the other thread in https://github.com/lfortran/lfortran/issues/8937#issuecomment-3599247487

opixdown:
okay @certik i will update the error message to "semantic error: more than one main program found".

PR #8955, opened 2025-12-03
tests: add mod_cpp_modfile and lookup_name regression coverage

krystophny:
Split original PR #8927 (modfile location crash with --cpp) into smaller, focused changes.

Order for review / merge:

1. **#8955 (this PR)** – add tests and references for `mod_cpp_modfile_main` (show_errors with `--cpp` + extrafiles) and `lookup_name_class63` (multi-module abstract type), and extend `run_tests.py` to build required extrafiles with the same options.
2. **#8956** – fix empty-file handling in `read_file` so zero-byte files are safe under libstdc++ assertions.
3. **#8957** – fix modfile location offsets and preprocessor interval bounds so `mod_cpp_modfile` and `lookup_name_class63` use correct, safe locations.
4. **#8958** – enable `-D_GLIBCXX_ASSERTIONS` in Quick-Checks-CI to catch latent bounds/UB issues.

PR #8927 (`krystophny:fix/modfile-location-offset`) is kept open as a reference implementation and should not be modified; the above PRs are all based directly on up-to-date `main`.


certik:
Do any of these reference tests change when we apply the fix from https://github.com/lfortran/lfortran/pull/8957?

krystophny:
> Do any of these reference tests change when we apply the fix from #8957?

I don't think so. I double check right now.

krystophny:
@certik after local rebase on main, nothing changes between the two branches in the references.

certik:
Ok, in that case we can merge these any time, as they are in some sense independent of the actual fixes.

tests/tests.toml
```diff
@@ -4664,6 +4664,14 @@ lookup_name = true
 line = 22
 column = 18
 
+[[test]]
+filename = "lookup_name_class63.f90"
+asr = true
```
certik, line 4669: I think we don't need `asr` here, no need to test it.

tests/tests.toml
```diff
@@ -4736,6 +4744,12 @@ asr = true
 filename = "errors/show_errors1.f90"
 show_errors = true
 
+[[test]]
+filename = "errors/mod_cpp_modfile_main.f90"
+show_errors = true
+options = "--cpp"
+extrafiles = "errors/mod_cpp_modfile_module.f90"
```
certik, line 4751: I don't see any error in the reference results. What is the purpose of this test?

certik:
These changes are hard to review, I need to play with it by hand. It's really easy to make a mistake in the test runner.

I think all we need it to allow extra files for the location information lookup, so that we can add tests. That should do it.

certik:
I think let's not maintain these extra tests logic. I'll keep the PR open, but mark it draft. If we later decide that we need this, then we can resurrect it.

krystophny:
I don't mind, this was just to test the stricter flags and since this is anyway triggered by the other integration tests this should be fine to close too.

krystophny:

PR #8956, opened 2025-12-03
Refactor read_file()

krystophny:
Ensure read_file correctly handles zero-byte files by returning an empty string without touching any buffer element. This removes latent UB and makes the existing empty.f90 test safe under stricter C++ assertions, without changing behavior for non-empty files.

PR #8958, opened 2025-12-03
ci: enable libstdc++ assertions in quick checks

krystophny:
Update Quick-Checks-CI to build and run tests with -D_GLIBCXX_ASSERTIONS in both C and C++. This is intended to land after the tests PR (#8955) and the fixes for empty-file handling (#8956) and modfile locations (#8957), so that CI catches latent bounds/UB issues without new failures.

### DIFF
--- a/.github/workflows/Quick-Checks-CI.yml
+++ b/.github/workflows/Quick-Checks-CI.yml
@@ -102,7 +102,6 @@ jobs:
         shell: bash -e -l {0}
         if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
         run: |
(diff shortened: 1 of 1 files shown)


certik:
It caught the bug at the CI (so it works!), and I now rebased the PR, since the fix is now merged, so it should pass now.

certik:
There were two bugs in main, one was empty files (now fixed), the other in `class_63`, fixed by #8957, which this PR still catches.

certik:
I rebased it, it should pass now.

certik:
Merged in https://github.com/lfortran/lfortran/pull/8961.

PR #8959, opened 2025-12-03
enh: enable select rank construct with fortran backend

HarshilShah1804:


integration_tests/CMakeLists.txt
```diff
@@ -749,11 +749,11 @@ RUN(NAME arrays_99 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_100 LABELS gfortran llvm)
 # RUN(NAME array_init LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME select_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
HarshilShah1804, line 754: The tests `select_rank_02` and `select_rank_03` don't work with fortran backend because of other issues in the `asr_to_fortran` conversion.
- select_rank_02 - because it first casts the integer to real and then does reshape. So, it is converted just as a plain array in the generated fortran code which causes rank mismatch.
- select_rank_03 - because `minloc` is not converted properly to fortran with default arguments.

We can fix those in subsequent PRs

certik (approved):

PR #8961, opened 2025-12-03
CI: enable bounds checking for C++ stdlib

certik:
On macOS use:

    -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG

on Linux use:

    -D_GLIBCXX_ASSERTIONS

And check all Debug builds. The Release build we build without it, to ensure it works too.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -57,7 +57,7 @@ jobs:
         shell: bash -e -l {0}
         run: |
             ( cd src && ./build0.sh )
-            export CXXFLAGS="-Werror"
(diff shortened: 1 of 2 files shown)


certik:
It fails in `class_63`, which might be due to not having the fix:

https://github.com/lfortran/lfortran/pull/8957


Testing this hypothesis in https://github.com/lfortran/lfortran/pull/8964.

certik:
The fix is now in, so I rebased this. The PR https://github.com/lfortran/lfortran/pull/8964 failed with https://github.com/lfortran/lfortran/issues/8969, which might be caused by these bounds checks, but it would have to be indirect, since the failure happens in the compiled code, not LFortran. So it might be a flaky test.

certik:
Everything passes, so it was an unrelated failure.

PR #8962, opened 2025-12-03
fix: cast assumed rank arrays to descriptor before accessing

HarshilShah1804:


certik (approved):
I think this looks good, thanks! 

Thanks for adding tests for both valid Fortran and invalid Fortran, so it covers all paths in the C++ code.

certik:
I resolved the conflicts, so CI should run now.

PR #8964, opened 2025-12-03
CI assert and a fix

certik:


certik:
The CI failed with https://github.com/lfortran/lfortran/issues/8969, but it's not clear if it is caused by this PR.

Issue #8968, opened 2025-12-03
ICE or code generation error w/BLOCK DATA and type COMPLEX.


w6ws:
Depending on how I compile the following, I get either a code generation error or an ICE:
```
      program bd_test
      end program
      BLOCK DATA
      COMPLEX AXVC
      COMMON /BLK9/ AXVC
      data axvc /(234.23, 34.9)/
      END
```
With `--fixed-form`, I get the code generation error:
```
$ lfortran --version
LFortran version: 0.58.0-185-g87999a014
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$
$ lfortran -c --fixed-form block_data.f
code generation error: Non-constant value found in struct initialization


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
If I omit the `--fixed-form`, I get an ICE:
```
$ lfortran -c block_data.f
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1137, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    result = fe.get_asr2(input, lm, diagnostics);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 276, in LCompilers::FortranEvaluator::get_asr2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
[...]

PR #8970, opened 2025-12-04
WIP: Fix ICE in BLOCK DATA 

opixdown:
## **Issue Fixed**
#8968 ICE or code generation error w/BLOCK DATA and type COMPLEX

## **Summary**
This PR prevents ICE that occurred when compiling BLOCK DATA constructs. It ensures the compiler no longer crashes while handling block data; further work is needed to address the code generation.

## **Reference Code**
```
program bd_test
COMPLEX AXVC
COMMON /BLK9/ AXVC

print *, "AXVC from MAIN =", AXVC
end program
BLOCK DATA
COMPLEX AXVC
COMMON /BLK9/ AXVC
data axvc /(234.23, 34.9)/
END
```

## **Progress**
I tried to debug and found out that the symbols in global_scope

```
debug-> Symbols in global_scope:
  - 1_blk9_axvc
  - axvc
  - bd_test
  - blk9
  - file_common_block_blk9   (it is in lowercase)
  - struct_instance_blk9
  ```

But the code was looking for 
```
debug-> looking for module: file_common_block_BLK9 (it is in uppercase)
  ```
The issue was case mismatch.

### DIFF
--- a/integration_tests/block_data_complex.f90
+++ b/integration_tests/block_data_complex.f90
@@ -2,7 +2,10 @@ program bd_test
 implicit none
 complex :: axvc
 common /blk9/ axvc
+if (abs(real(axvc) - 234.23) > 0.01) error stop "Real part incorrect"
+if (abs(aimag(axvc) - 34.9) > 0.01) error stop "Imaginary part incorrect"
 print *, "AXVC from MAIN =", axvc
+print*,"Test Passed"
 end program
 
 block data
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -9939,35 +9939,35 @@ public:
         tmp = complex_from_floats(re2, im2, type);
     }
 
-void visit_ComplexConstant(const ASR::ComplexConstant_t &x) {
-    double re = x.m_re;
-    double im = x.m_im;
-    int a_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
-    llvm::Constant *re2, *im2;
-    llvm::Type *type;
-    switch( a_kind ) {
-        case 4: {
-            re2 = llvm::ConstantFP::get(context, llvm::APFloat((float)re));
-            im2 = llvm::ConstantFP::get(context, llvm::APFloat((float)im));
-            type = complex_type_4;
-            break;
+    void visit_ComplexConstant(const ASR::ComplexConstant_t &x) {
+        double re = x.m_re;
+        double im = x.m_im;
+        int a_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
+        llvm::Constant *re2, *im2;
+        llvm::Type *type;
+        switch( a_kind ) {
+            case 4: {
+                re2 = llvm::ConstantFP::get(context, llvm::APFloat((float)re));
+                im2 = llvm::ConstantFP::get(context, llvm::APFloat((float)im));
+                type = complex_type_4;
+                break;
+            }
+            case 8: {
+                re2 = llvm::ConstantFP::get(context, llvm::APFloat(re));
+                im2 = llvm::ConstantFP::get(context, llvm::APFloat(im));
+                type = complex_type_8;
+                break;
+            }
+            default: {
+                throw CodeGenError("kind type is not supported");
+            }
         }
-        case 8: {
-            re2 = llvm::ConstantFP::get(context, llvm::APFloat(re));
-            im2 = llvm::ConstantFP::get(context, llvm::APFloat(im));
-            type = complex_type_8;
-            break;
-        }
-        default: {
-            throw CodeGenError("kind type is not supported");
-        }
+        // Create a compile-time constant struct for ComplexConstant
+        // Complex numbers are represented as {real, imaginary} structs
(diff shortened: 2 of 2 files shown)


opixdown:
I tried to fix the codegen error the issue was that earlier the complex literals were being built using runtime instructions. I replaced that with an LLVM constant struct, so comple values are now true compile time constants. With this change it fix the code gen error 

the changes i made were in the asr_to_llvm.cpp:
```
-        tmp = complex_from_floats(re2, im2, type);
     }
+    std::vector<llvm::Constant*> elements = {re2, im2};
+    llvm::StructType* struct_type = llvm::cast<llvm::StructType>(type);
+    tmp = llvm::ConstantStruct::get(struct_type, elements);
+}
```
it works the ouput was 

```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran tes.f
AXVC from MAIN =    (2.34229996e+02, 3.49000015e+01)
```

but when i runned `./run_tests.py` it got failed for `./integration_tests/arrays_01_complex.f90` and i am not sure how to tackle it

certik:
What failure do you get with `./run_tests.py` ? If the reference tests changed, you can update them with `./run_tests.py -u`.

opixdown:
> What failure do you get with `./run_tests.py` ? If the reference tests changed, you can update them with `./run_tests.py -u`.

Yes the error was due to updated reference.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9939,31 +9939,35 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            throw CodeGenError("kind type is not supported");
         }
-        tmp = complex_from_floats(re2, im2, type);
     }
+    // Create a compile-time constant struct for ComplexConstant
+    // Complex numbers are represented as {real, imaginary} structs
+    std::vector<llvm::Constant*> elements = {re2, im2};
+    llvm::StructType* struct_type = llvm::cast<llvm::StructType>(type);
+    tmp = llvm::ConstantStruct::get(struct_type, elements);
+}
```
certik: The indentation is incorrect.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9939,31 +9939,35 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        default: {
+            throw CodeGenError("kind type is not supported");
         }
-        tmp = complex_from_floats(re2, im2, type);
     }
+    // Create a compile-time constant struct for ComplexConstant
+    // Complex numbers are represented as {real, imaginary} structs
+    std::vector<llvm::Constant*> elements = {re2, im2};
+    llvm::StructType* struct_type = llvm::cast<llvm::StructType>(type);
+    tmp = llvm::ConstantStruct::get(struct_type, elements);
```
certik: Why cannot the `complex_from_floats` function be used? Why do we need to change the ABI to a struct? Why not change it to a struct in the `complex_from_floats` function?

integration_tests/block_data_complex.f90
```diff
@@ -0,0 +1,12 @@
+program bd_test
+implicit none
+complex :: axvc
+common /blk9/ axvc
+print *, "AXVC from MAIN =", axvc
```
certik, line 7: Also add `if () error stop` style test and check that `axvc` has the correct values, both real and imaginary.

src/libasr/codegen/asr_to_llvm.cpp
opixdown: the  `complex_from_floats()` creates runtime instructions:
```cpp
llvm::AllocaInst *pres = llvm_utils->CreateAlloca(*builder, complex_type);
return llvm_utils->CreateLoad2(complex_type, pres);
```

It gives `llvm::Value*` but when I traced back `visit_StructConstant()`:
```cpp
visit_expr_wrapper(value);
initializer = llvm::dyn_cast(tmp);
if (!initializer) {
    throw CodeGenError("Non-constant value found in struct initialization");
}
```

The `dyn_cast<llvm::Constant>` fails because tmp is an `llvm::Value*`  not a  `llvm::Constant*`.

I think there is no ABI change the complex values remain {real, imaginary} structs only the construction method moved from runtime to compile time ,here also the layout is same like { float, float } or { double, double }.

About changing the function actually it uses builder->CreateAlloca(). For data statements there is  no builder available as we are in global constant. That’s why I had to build the constant struct directly with `llvm::ConstantStruct::get()`.

However, My change affects all complexconstant usage. should I use if statements like to keep existing behavior for runtime code and only use compiletime constants when specifically needed?

integration_tests/block_data_complex.f90
opixdown: okayy

src/libasr/codegen/asr_to_llvm.cpp
certik: > However, My change affects all complexconstant usage. should I use if statements like to keep existing behavior for runtime code and only use compiletime constants when specifically needed?

Can you think of some Fortran code that will trigger this failure mode?

PR #8985, opened 2025-12-07
fix: handle descriptor strings in cptrtoptr and ptrtocptr

jinangshah21:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6111,18 +6118,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                llvm::Value* data_ptr = builder->CreateBitCast(llvm_cptr, i8_ptr_ty);
-                llvm::Value* fptr_cast = builder->CreateBitCast(llvm_fptr, char_desc_type->getPointerTo());
-                llvm::Value* gep_len = llvm_utils->create_gep2(char_desc_type, fptr_cast, 0);  // .len
-                llvm::Value* gep_data = llvm_utils->create_gep2(char_desc_type, fptr_cast, 1); // .ptr
-
-                llvm::Value* len_val = llvm::ConstantInt::get(llvm::Type::getInt32Ty(context), 1); // dummy length
-                builder->CreateStore(len_val, gep_len);
+                llvm::Value* data_ptr = builder->CreateBitCast(llvm_cptr, llvm_utils->i8_ptr);
+                llvm::Value* gep_data = llvm_utils->create_gep2(llvm_utils->string_descriptor, llvm_fptr, 0);
                 builder->CreateStore(data_ptr, gep_data);
```
jinangshah21, line 6123: I don't know why we were using {i32, i8*} as descriptor while converting from cptr to fortran pointer. As cptr should be just i8*.

certik: I don't know either, but all tests pass, so your PR is good. We could look into git history to see.

Issue #8967, opened 2025-12-03
ICE with simple formatted READ


w6ws:
If I change the `read (5,100)` to `read (5,*)` the program compiles.  If I use a formatted `write` instead of `read`, the program also compiles.
```
$ cat read_ice.f90
      program read_ice
      implicit none
      integer :: i
      i = 42
      read (5,100) i
  100 format (I3)

      END
$
$ lfortran --version
LFortran version: 0.58.0-185-g87999a014
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$
$ lfortran -c read_ice.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1193, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 393, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
    compiler_options, run_fn, "", infile, lm);
[...]

PR #9003, opened 2025-12-08
fix : an error catch has been introduced to provide correct explanation (issue #8967)

Atharv1405:
issue #8967 

there was no check for format so check is introduced to give a proper error message in case of a integer is given in place of character 

MRE:
```  
program internal
   implicit none
   integer :: i
   character(8) :: string = '42'
   read (string,100) i
   print 100,i
100 format (I3)
 END program internal
```

BEFORE:
```
LCompilersException: Unsupported string physical type.
```

AFTER :
```
code generation error: use a character format string instead, e.g. READ(5, '(I3)')
 --> t1.f90:5:17
  |
5 |    read (string,100) i
  |                 ^^^ 
```

QUESTIONS:

this format can be supported on lfortran as well like in gfortran but i new to this repository so is there any function available to convert integer to character for llvm backend ?

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -1175,6 +1175,19 @@ public:
         }
 
         if (a_fmt && ASR::is_a<ASR::IntegerConstant_t>(*a_fmt)) {
+
+            if (_type == AST::stmtType::Read) {
+                diag.add(Diagnostic(
+                    "FORMAT statement labels are not supported yet for formatted READ.",
+                    Level::Error, Stage::Semantic, {
+                        Label(
+                            "use a character literal instead, e.g. READ(*,'(I3)')",
+                            {loc}
+                        )
+                    }));
+                throw SemanticAbort();
+            }
+
             ASR::IntegerConstant_t* a_fmt_int = ASR::down_cast<ASR::IntegerConstant_t>(a_fmt);
             int64_t label = a_fmt_int->m_n;
             if (format_statements.find(label) == format_statements.end()) {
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11179,8 +11179,33 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+                args.push_back(fmt_data);
+                args.push_back(fmt_len);
+
+            } else if (ASR::is_a<ASR::IntegerConstant_t>(*x.m_fmt)) {
+                // label like READ(5, 100)
+                throw CodeGenError(
+                    "use a character format string instead, e.g. READ(5, '(I3)')",
+                    x.m_fmt->base.loc
+                );
```
certik: This creates code generation error. Instead we need to emit semantic error. Add this check in AST->ASR, and then here just add an assert that this condition is met.

Also make the error message more clear: explain why it is wrong (is it because we do not support it yet? Or is this wrong in Fortran -- explain why, etc.), and then you can provide a suggestion what the user should do instead.

Atharv1405: ok i will do that

Atharv1405:
@certik Hi , i have tried to add changes as recommended please review them again
**Here is the new error shown**
```
semantic error: FORMAT statement labels are not supported yet for formatted READ.
 --> t2.f90:5:7
  |
5 |       read (5,100) i
  |       ^^^^^^^^^^^^^^ use a character literal instead, e.g. READ(*,'(I3)')


```

please suggest any changes if needed

certik:
Perfect, that looks good. Now add a test for this new error message into continue_compilation_1.f90 and update reference tests.

Issue #8996, opened 2025-12-07
ICE: DATA statement with parameterized repeat count


w6ws:
```
$ cat data_ice.f90
      program data_ice

      real :: array(5)
      integer, parameter :: nelems=5
      DATA (array(I), I=1,5) /nelems*3.1415/

      stop
      end
$ 
$ lfortran --version
LFortran version: 0.58.0-208-ge8cc1488e
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$ 
$ lfortran -c data_ice.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1137, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    result = fe.get_asr2(input, lm, diagnostics);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 267, in LCompilers::FortranEvaluator::get_asr2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
    res = get_ast2(code_orig, lm, diagnostics);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 224, in LCompilers::FortranEvaluator::get_ast2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
[...]

PR #9005, opened 2025-12-08
update in repeat_list_add function to handle datatype error correctly (issue #8996)

Atharv1405:
issue #8996 

**changes**
introduced if-else block in the code to handle type conflict which was occuring earlier  

**MRE**
```
program data_ice

      real :: array(5)
      integer, parameter :: nelems=5
      DATA (array(i), i=1,5) /nelems*3.1415/

      stop
      end
```

**AFTER**
```
 --> t2.f90:5:19
  |
5 |       DATA (array(i), i=1,5) /nelems*3.1415/
  |                   ^ 'i' is undeclared


```

**WARNING SHOWN**
```
/home/atharv/Documents/openSource/lfortran/src/lfortran/ast.h: In instantiation of ‘bool LCompilers::LFortran::AST::is_a(const U&) [with T = Num_t; U = ast_t]’:
/home/atharv/Documents/openSource/lfortran/src/lfortran/parser/semantics.h:1878:78:   required from here
/home/atharv/Documents/openSource/lfortran/src/lfortran/ast.h:32:26: warning: comparison between ‘const enum LCompilers::LFortran::AST::exprType’ and ‘const enum LCompilers::LFortran::AST::astType’ [-Wenum-compare]
   32 |     return T::class_type == x.type;
      |               ~~~~~~~~~~~^~~~~~~~~

```

**QUESTIONS**
still a different error persists after applying , does it occur due the change i have made or its a different error which can be raised as a different issue , and also can you please advise how i can eliminate the warning being shown while building . thanks in advance

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
-    "infile_hash": "d6bed42c137e80709c1b5ca7c93c2e1435612ae6ba2dad88a85c8118",
+    "infile_hash": "e6f658d47bbbb954954fe4c1de9bed77e2dc3819606de30697f1dae3",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "f8036f7d09224862ec0dd3805f4ba2311c9b6368ecead795560381e1",
+    "stderr_hash": "0580ed238c517e6c70cf53f034822cf0189d8394023889a58d976349",
     "returncode": 1
 }
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -129,6 +129,12 @@ semantic error: BOZ literal constant cannot be used in binary operations
 174 |     integer::tt = b'01' * 3
     |                   ^^^^^^^^^ 
 
+semantic error: DATA statement element count mismatch: 5 elements on left-hand side, 1 values on right-hand side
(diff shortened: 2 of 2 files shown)


harperjf:
The semantic error with ```i``` not declared is a different issue from the ```DATA``` one and I shall write one for it.

certik:
Add a test for this into `tests/errors/continue_compilation_1.f90` and ensure that it gives the right error now.

Atharv1405:
@certik i have added the test , can you please check and confirm if i did it correctly , after rebuilding the compiler the error match the above reference and test passed after update . please tell if there is any mistake ,i will try to correct it.

certik:
That's good. Now update reference tests with `./run_tests.py -u`. It should show the new error.


If it doesn't show the error, then your fix doesn't actually work. Then let's figure out why it doesn't work.

Atharv1405:
I think the fix doesn't work because all the test passed after i ran `./run_tests.py -u` but when i manually ran the MRE the above error is shown which is different than the earlier error .

certik:
Yes, so you need to figure out how to make the fix work.

Atharv1405:
yes i have figured out the issue , earlier there was a typecasting issue as seen in the warning .now i have resolved it
error before running `run_tests.py -u`
*error* (as expected)
```
=== STDERR DIFF ===
Reference: tests/reference/asr-continue_compilation_1-04b6d40.stderr
Output:    tests/output/asr-continue_compilation_1-04b6d40.stderr
131a132,137
> semantic error: DATA statement element count mismatch: 5 elements on left-hand side, 1 values on right-hand side
>    --> tests/errors/continue_compilation_1.f90:178:5
>     |
> 178 |     DATA (ary(I), I=1,5) /nelems*3.1415/
>     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
> 
```

ICE is no longer shown while running
after updating the reference using `run_tests.py -u` all test cases got passed

please review and provide feedback

certik:
Go ahead and resolve the conflict --- update against the latest main.

certik:
You need to update reference tests, the new error is not showing.

Atharv1405:
i have added the references test and rebased the branch with lfortran/main . i think it is ready to be merged with main without any conflicts

PR #8998, opened 2025-12-07
fix: inquire(unit, size=) returning 0 instead of actual file size

opixdown:
## **Issue Fixed**
closes #8990 Handle size in inquire when unit is given instead filename

## **Summary**
This PR fixes the inquire statement to correctly return file size when querying by unit number. Earlier the size parameter was  always returning 0 instead of the actual file size.

## **Reference Code**
```
program test_inquire_size
    implicit none
    integer :: fh
    integer :: length
    character(*), parameter :: test_lines = 'build.f90'
    open(newunit=fh, file="test", access="stream", form='unformatted')
    write(fh) test_lines
    inquire(fh, size=length)
    close(fh)
    print *, "File size =", length, "bytes"
end program test_inquire_size
```

## **Behaviour before PR**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90 
File size =    0    bytes
```

## **Behaviour after PR**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90 
 File size =    9    bytes
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran test.f90 && ./a.out
 File size =           9 bytes
```

PR #9006, opened 2025-12-08
Revert "fix: inquire(unit, size=) returning 0 instead of actual file size"

assem2002:
Reverts lfortran/lfortran#8998


certik:
This is probably not needed, as the error is elsewhere.

Issue #9008, opened 2025-12-09
Implicit typing not accepted in an implied do.


harperjf:
This program is valid Fortran but lfortran won't compile it.
 ```
  integer n(3)
  n = [(42, i = 1,size(n))]
  print "(3I3)", n
end program
```
Output from 4 other compilers (gfortran, g95, ifx, AMD flang):
```
 42 42 42
```
Output from lfortran 0.58.0 either with or without ```--std=f23```
```
Semantic error: The implied do loop variable 'i' is not declared
 --> noimplicit.f90:2:8
  |
2 |   n = [(42, i = 1,size(n))]
  |        ^^^^^^^^^^^^^^^^^^^ 
```
There is no ```implicit none``` statement in the program, and so ```i``` should have been treated as a scalar default integer. Some valid Fortran programs use features that the Lfortran developers disapprove of and treat as errors, and using 
```--std=f23``` usually allows such features. Not this one.

**Comment by harperjf** [comment]
Changing from an implied do to an actual do loop as in this program:
```
  integer n(3)
  do i = 1,size(n)
     n(i) = 42
  end do
  print "(3I3)", n
end program
```
made lfortran --std=f23 accept the program and print ``` 42 42 42```, but without that option it would not compile. So the treatment of implied do is wrong.

harperjf:
Similar problem with a ```forall``` statement whose index variable ```i``` was not declared and there was no ```implicit``` statement.

opixdown:
on it!!

PR #9009, opened 2025-12-09
Feat : Implicit typing for undeclared loop variables in do loops

opixdown:
## **Issue Fixed**
#9008 Implicit typing not accepted in an implied do.

## **Summary**

This PR adds implicit declaration support for loop variables in both implied do loops and regular do loops, following Fortran's implicit typing rules. 

## **Problem**

LFortran required loop variables to be explicitly declared in:

- Implied do loops 

- Regular do loops

This violated Fortran standard behavior where loop variables can be implicitly typed according to the implicit typing rules.

## **Reference Code (I)**
```
 integer n(3)
 n = [(42, i = 1,size(n))]
 print "(3I3)", n
end program
```

## **Reference Code (II)**
```
  integer n(3)
  do i = 1,size(n)
     n(i) = 42
  end do
  print "(3I3)", n
end program
```

**CODE I**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran test.f90 && ./a.out
 42 42 42

(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90
Semantic error: The implied do loop variable 'i' is not declared
 --> noimplicit.f90:2:8
  |
2 |   n = [(42, i = 1,size(n))]
  |        ^^^^^^^^^^^^^^^^^^^ 

```

**CODE 2**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran t.f90 && ./a.out
 42 42 42

 --> ./t.f90:2:3 - 4:8
  |
2 |      do i = 1,size(n)
  |      ^^^^^^^^^^^^^^^^...
...
  |
4 |      end do
  | ...^^^^^^^^ 'i' is undeclared
```

**CODE 1**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90
 42 42 42
 ```
**CODE 2**
```
 42 42 42
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2270,6 +2270,7 @@ RUN(NAME write_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME write_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME write_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME write_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME write_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME do_loop_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4008,7 +4008,16 @@ public:
                         )));
                         throw SemanticAbort();
                     }
-
+    
+                    if (is_allocatable && s.m_initializer != nullptr) {
+                        diag.add(Diagnostic(
+                            "An 'allocatable' variable cannot have an initialization expression",
+                            Level::Error, Stage::Semantic, {
+                                Label("", {s.loc})
+                            }));
+                        throw SemanticAbort();
+                    }
+                    
                     if (s_intent == ASRUtils::intent_out && value_attr) {
                         diag.add(Diagnostic(
                             "`value` attribute conflicts with `intent(out)` attribute",
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -175,6 +175,8 @@ program continue_compilation_1
     integer :: fmt_i1, fmt_i2, fmt_i3 ! for issue #8925
     integer, parameter :: arr_implicit(3) = [(42, j_implicit = 1, 3)]
     integer :: n_decl(3) = [(42, k_implicit = 1, 3)] !declaration test (intializtion)
+    integer, allocatable :: allocate_int = 1
+    character(:), allocatable :: allocate_char = "H"
 
 
 
--- a/integration_tests/write_11.f90
+++ b/integration_tests/write_11.f90
@@ -1,0 +1,16 @@
+program write_11
+    implicit none
+
+    character(len=*), parameter :: temp_file = "_lfortran_temporary_file"
+    character(len=*), parameter :: test_line = "HelloWorld!"
+    integer :: unit, ios
+    ios = 0
+    unit = 1
+
+    open(newunit=unit, file=temp_file)
+    write(unit, iostat=ios) test_line
+    close(unit)
+    print *, ios
+    if(ios <= 0) error stop "IOS should be positive -- as a runtime error should occur"
+
+end program
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -12018,9 +12018,13 @@ public:
             args.push_back(end_data);
             args.push_back(end_len);
         }
-        std::string fmt_str;
-        for (size_t i=0; i<fmt.size(); i++) {
-            fmt_str += fmt[i];
+        std::string fmt_str {};
+        if(x.m_is_formatted){
+            for (size_t i=0; i<fmt.size(); i++) {
+                fmt_str += fmt[i];
+            }
+        } else {
+            // Don't provide format.
         }
         llvm::Value *fmt_data = LCompilers::create_global_string_ptr(context, *module, *builder, fmt_str);
         llvm::Value *fmt_len = llvm::ConstantInt::get(context, llvm::APInt(64, fmt_str.length()));
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5184,6 +5184,19 @@ LFORTRAN_API char* _lpython_read(int64_t fd, int64_t n)
     return c;
 }
 
+/**
+ * @brief Checks Format match for File Write statement and OpenFile statement
+ * OpenFile -> 'Unformatted' + FileWrite -> Binary Format (no format provided) => Match
+ * OpenFile -> 'Formatted' + FileWrite -> Formatted Format (format provided) => Match
+ * Otherwise => No Match
+ */
+bool is_write_and_open_format_match(bool unit_file_bin, const char* format_data){
+    ASSERT(format_data)
+    const bool is_openFile_formatted =  unit_file_bin == false;
+    const bool is_fileWrite_formatted =  format_data[0] != '\0';
+    return is_openFile_formatted == is_fileWrite_formatted;
+}
+
 LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const char* format_data, int64_t format_len, ...)
 {
     bool unit_file_bin;
@@ -5191,10 +5204,22 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const 
     bool read_access, write_access;
     int delim;
     FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin, &access_id, &read_access, &write_access, &delim);
+
+    if(!is_write_and_open_format_match(unit_file_bin, format_data)){
+        if(iostat) {
+            *iostat = 5001;
+            return;
+        } else {
+            fprintf(stderr, "Runtime Error: Format mismatch between "
+                "OPEN statement and WRITE statement on unit %d.\n", unit_num);
+            exit(1);
+        }
+    }
+    
     if (!filep) {
         filep = stdout;
     }
-    if (unit_file_bin) {
+    if (unit_file_bin) { // Unformatted
         fseek(filep, 0, SEEK_END);
         va_list args;
         va_start(args, format_len);
@@ -5245,7 +5270,7 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const 
         } else {
             if(iostat != NULL) *iostat = 0;
         }
-    } else {
+    } else { // Formatted
         va_list args;
         va_start(args, format_len);
         char* str = va_arg(args, char*);
--- a/tests/errors/write_01.f90
+++ b/tests/errors/write_01.f90
@@ -1,0 +1,13 @@
+program write_01
+    implicit none
+
+    character(len=*), parameter :: temp_file = "lfortran_temporary_file_write_01.dat"
+    character(len=*), parameter :: test_line = "HelloWorld!"
+    integer :: unit, ios
+    ios = 0
+    unit = 1
+
+    open(newunit=unit, file=temp_file) !formatted
+    write(unit) test_line !unformatted
+    close(unit)
+end program
--- a/tests/errors/write_02.f90
+++ b/tests/errors/write_02.f90
@@ -1,0 +1,13 @@
+program write_02
+    implicit none
+
+    character(len=*), parameter :: temp_file = "lfortran_temporary_file_write_02.dat"
+    character(len=*), parameter :: test_line = "HelloWorld!"
+    integer :: unit, ios
+    ios = 0
+    unit = 1
+
+    open(newunit=unit, form="unformatted", file=temp_file) !unformatted
+    write(unit,*) test_line !formatted
+    close(unit)
+end program
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -3,11 +3,13 @@
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
     "infile_hash": "2db05fd953c1f3a7d75cb4f4422d5d23ec9a3e0aad057ea5d909e0bc",
+    "infile_hash": "ea71c56d04c495d285480a7259a6455b49af4b49f11226dc2f1d5677",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
     "stderr_hash": "b1eaf0df0ecfbf752ed1ac0abdd86c6b6733fd7759f510c0bdb9eab4",
+    "stderr_hash": "7e60bcce38e89a2c372fdb488aecc3e52f03477e954b56adc4948582",
     "returncode": 1
 }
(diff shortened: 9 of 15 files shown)

### CODE REVIEW

File: tests/reference/implicit_do_1.f90 [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+    
+    n = [(42, i = 1, size(n))]
+    
+    if (n(1) /= 42) error stop
+    if (n(2) /= 42) error stop
+    if (n(3) /= 42) error stop
+    
+    print "(3I3)", n
+    
+end program
```
certik, line 16: Move these one level up. In fact move them to integration tests if this is valid code.

Then add error tests in tests/tests.toml.

certik: I see you registered them in integration tests. So move them there.

You should run the tests locally to ensure they actually work before submitting a PR.

opixdown: Thank you for the feedback I will move the test files to the integration_tests/ directory and add error tests in tests/tests.toml for invalid cases.
I apologize for not running the tests locally before submitting.

opixdown:
@certik  i want to confirm that the tests are currently producing the correct output (42 42 42), so I'll move them as-is to the integration tests directory. Is this right?

integration_tests/implicit_do_1.f90
```diff
@@ -0,0 +1,9 @@
+program implicit_do_1
+    implicit none
```
HarshilShah1804:
```suggestion
```
I believe you will have to remove `implicit none` to make the test work correctly with GFortran as mentioned in the issue.

> There is no implicit none statement in the program, and so i should have been treated as a scalar default integer.

integration_tests/implicit_do_2.f90
```diff
@@ -0,0 +1,11 @@
+program implicit_do_2
+    implicit none
```
HarshilShah1804:
```suggestion
```
I believe you will have to remove `implicit none` to make the test work correctly with GFortran as mentioned in the issue.

> There is no implicit none statement in the program, and so i should have been treated as a scalar default integer.

integration_tests/implicit_do_1.f90
HarshilShah1804: The failing check of ubuntu LLVM 11 is because of this. After these changes, it should work fine.

HarshilShah1804:
Also, as you have added the tests in `integration_tests`, you can remove them from `tests/references`.

integration_tests/implicit_do_1.f90
opixdown: thank you for feedback i have commit the changes you suggested

HarshilShah1804 (approved):

integration_tests/CMakeLists.txt
```diff
@@ -1516,6 +1516,8 @@ RUN(NAME implicit_typing_04 LABELS gfortran llvmImplicit EXTRA_ARGS --implicit-i
 RUN(NAME implicit_typing_05 LABELS gfortran llvm  EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
 
+RUN(NAME implicit_do_1 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc)
+RUN(NAME implicit_do_2 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc)
```
certik: I think you need to call llvmImplicit here (look it up).

With `llvm` it should actually fail, but it doesn't. That means we have a bug there --- once you fix it, add it to continue_compilation_01.f90 to check that it says `i` is not declared.

certik:
I think this PR requires more work.

integration_tests/CMakeLists.txt
opixdown: thanks for the clarification i will try to debug it

opixdown:
Behaviour after changes 
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran --implicit-typing t.f90
42    42    42
```

```
semantic error: The implied do loop variable 'i' is not declared
 --> ./t.f90:5:10
  |
5 |     n = [(42, i = 1, 3)]  
  |          ^^^^^^^^^^^^^^ 


```

integration_tests/CMakeLists.txt
```diff
@@ -1515,6 +1515,8 @@ RUN(NAME implicit_typing_04 LABELS gfortran llvmImplicit EXTRA_ARGS --implicit-i
 RUN(NAME implicit_typing_05 LABELS gfortran llvm  EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
 
+RUN(NAME implicit_do_1 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
+RUN(NAME implicit_do_2 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
```
certik: Since these are implicit typing tests, why not append them to the rest of the implicit typing tests?
```suggestion
RUN(NAME implicit_typing_07 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
RUN(NAME implicit_typing_08 LABELS gfortran llvm c llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
```

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5081,7 +5081,37 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        std::string loop_var_name = to_lower(x.m_var);
+        ASR::symbol_t* loop_var_sym = current_scope->resolve_symbol(loop_var_name);
+        if (loop_var_sym == nullptr) {
+        // Check if implicit typing is enabled
+        if (!compiler_options.implicit_typing) {
+            diag.add(Diagnostic("The implied do loop variable '" +
+                loop_var_name + "' is not declared",
+                Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
+            throw SemanticAbort();
+        }
```
certik: The indentation is incorrect.

tests/errors/continue_compilation_1.f90
```diff
@@ -57,6 +57,12 @@ subroutine s(c) bind(c)
         character(len=2, kind=c_char), intent(in) :: c
     end subroutine s
 
+    subroutine test_implicit_loop_var()
+        implicit none
+        integer :: n(3)
+        n = [(42, i = 1, 3)]  
+    end subroutine test_implicit_loop_var
```
certik: First of all, please follow https://github.com/opixdown/lfortran/blob/0a54f17dd4857f60975b1f10eda81c15e5fb2335/tests/errors/continue_compilation_1.f90#L1-L3, to keep the diff small. I see the empty space is running low, I made more space in https://github.com/lfortran/lfortran/pull/9029.

Second, I think this doesn't need to be a function, or does it? Cannot you simply add `n = [(42, i = 1, 3)]` at the end of the main program below? Change `n` and `i`, declare new `n` correctly if needed make sure new `i` is not declared.

Based on my other comment, you actually have to test this implicit implied do loop in both a declaration and a body. So far it's only tested in a body.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5081,7 +5081,37 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            // Implicitly declare the loop variable 
+            ASR::ttype_t* var_type = nullptr;
+            // Apply implicit typing rules 
+            char first_char = loop_var_name[0];
+            if (first_char >= 'i' && first_char <= 'n') {
```
certik: You need to use the implicit typing dictionary, since the variable can be implicitly typed to a different type, so you need to honor it. See how other implicit variables are handled.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10004,13 +10004,42 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                to_lower(x.m_var) + "' is not declared",
-                Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
-            throw SemanticAbort();
+            // Check if implicit typing is enabled
+            if (!compiler_options.implicit_typing) {
+                diag.add(Diagnostic("The implied do loop variable '" +
+                    var_name + "' is not declared",
+                    Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
+                throw SemanticAbort();
+            }
```
certik, line 10146: It looks like we have duplicated code here, once for declarations, one for bodies. Is there a way to reuse the same code for both?

We have no tests for declarations, only for bodies, so this code is not tested.

Also, the error test is only for the body, not a declaration either.

All my comments for the body visitor apply here, since it's the same code. Just extract it to some function or reuse the visitor from common, so that we have just one code and then when you fix my comments above, the fixes will automatically apply here too.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5081,7 +5081,37 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
         if (x.m_var) {
-            var = replace_with_common_block_variables(ASRUtils::EXPR(resolve_variable(x.base.base.loc, to_lower(x.m_var))));
+        std::string loop_var_name = to_lower(x.m_var);
+        ASR::symbol_t* loop_var_sym = current_scope->resolve_symbol(loop_var_name);
+        if (loop_var_sym == nullptr) {
+        // Check if implicit typing is enabled
+        if (!compiler_options.implicit_typing) {
+            diag.add(Diagnostic("The implied do loop variable '" +
+                loop_var_name + "' is not declared",
+                Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
```
certik: Here is how the error currently looks like:
```
semantic error: The implied do loop variable 'i' is not declared
  --> tests/errors/continue_compilation_1.f90:63:14
   |
63 |         n = [(42, i = 1, 3)]  
   |              ^^^^^^^^^^^^^^ 
```
Instead of highlighting the full expression, only highlight `i`. That should be possible to do with something like:

```suggestion
                Level::Error, Stage::Semantic, {Label("", {x.m_var.base.loc})}));
```

opixdown: @certik I tried the suggested change but x.m_var is a char* without location info. The AST only stores x.base.base.loc for the entire expression. we would need to modify the parser to store location info for m_var separately. Should I proceed with modifying the parser ?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2782,6 +2782,36 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            var_type = implicit_dictionary[first_char_str];
+        } else {
+            diag.add(Diagnostic("Cannot determine implicit type for variable '" + var_name + "'",
+                Level::Error, Stage::Semantic, {Label("", {loc})}));
+            throw SemanticAbort();
+        }
+        // Create the variable
+        SetChar variable_dependencies_vec;
+        variable_dependencies_vec.reserve(al, 1);
+        ASR::symbol_t* sym = ASR::down_cast<ASR::symbol_t>(ASR::make_Variable_t(al, loc, scope, s2c(al, var_name), variable_dependencies_vec.p,
```
assem2002: Do you think we need to mangle the variable name ?
assume 
```.f90
integer :: n(3)
  integer :: i
  n = [(42, i = 1, size(n))]
```
will that work properly?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2782,6 +2782,36 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         }
     }
 
+    ASR::symbol_t* create_implicit_loop_variable(const Location &loc,const std::string &var_name,SymbolTable* scope) {
+        // Check if implicit typing is enabled
+        if (!compiler_options.implicit_typing) {
+            diag.add(Diagnostic("The loop variable '" + var_name + "' is not declared",
+                Level::Error, Stage::Semantic, {Label("", {loc})}));
+            throw SemanticAbort();
+        }
```
assem2002: **It's a minor**; `create_implicit_loop_variable` implies `implicit_typing` is already **true**, so I think the caller should use the function only when it's known for sure that we need to create an implicit loop variable + `create_implicit_loop_variable`  will assert that `implicit_typing` is true to avoid wrong usage.

assem2002:
Reference tests need an update.

src/lfortran/semantics/ast_common_visitor.h
opixdown: No, it doesn't work properly. I tested and the implicit loop variable overwrites the user's declared i.
I think we should mangle the name to avoid conflicts?

assem2002: Oh. I think that example is misleading. It should overwrite user's variable, that's correct.
anyways we should mangle the name to avoid conflicts.

assem2002: I dumped the idea for now. I think not mangled names is fine for now.

tests/errors/continue_compilation_1.f90
```diff
@@ -74,44 +74,6 @@ subroutine s(c) bind(c)
...
-
-
-
-
-
-
-
-
-
-
```
certik, line 114: Why did you remove these empty lines? I makes the diff huge below.

integration_tests/implicit_typing_07.f90
```diff
@@ -0,0 +1,8 @@
+program implicit_do_1
+    integer :: n(3)
+    n = [(42, i = 1, size(n))]
+    if (n(1) /= 42) error stop 
+    if (n(2) /= 42) error stop
+    if (n(3) /= 42) error stop
+    print "(3I3)", n
+end program
```
certik, line 16: This tests for a body, but not a declaration, correct? Do we have an already existing test for a declaration?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5093,7 +5093,18 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
         ASR::ttype_t* type = nullptr;
         var = start = end = nullptr;
         if (x.m_var) {
-            var = replace_with_common_block_variables(ASRUtils::EXPR(resolve_variable(x.base.base.loc, to_lower(x.m_var))));
+            const std::string loop_var_name = to_lower(x.m_var);
+            ASR::symbol_t* loop_var_sym = current_scope->resolve_symbol(loop_var_name);
+            if (!loop_var_sym) {
+                if (!compiler_options.implicit_typing) {
+                    diag.add(Diagnostic("The loop variable '" + loop_var_name + "' is not declared",
+                        Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
```
certik, line 5101: The `x.m_var` doesn't have a location information? That would highlight just the symbol.

src/lfortran/semantics/ast_common_visitor.h
certik: I think this comment was "resolved", but not actually addressed.

src/lfortran/semantics/ast_body_visitor.cpp
opixdown: @certik the x.m_var is a char* without location info. The AST only stores x.base.base.loc for the entire expression. we would need to modify the parser to store location info for m_var separately.

certik: Ah, I see! In that case it's fine for now. We can later see if we want to modify the parser for this or not. Don't worry about it now.

integration_tests/implicit_typing_09.f90
```diff
@@ -0,0 +1,10 @@
+program implicit_do_9
+    integer :: n(3) = [(42, i = 1, 3)] !declaration test (initialization) 
+    
+    if (n(1) /= 42) error stop 
+    if (n(2) /= 42) error stop
+    if (n(3) /= 42) error stop
+    
+    print "(3I3)", n
+    
+end program
```
certik: Given how tiny these tests are, can these 3 tests possibly be merged into just one test? To speed up the test suite.

certik:
I think you didn't update the reference data, or perhaps some conflict happened. Go ahead, merge main, resolve conflicts, then update the references and it should be good.

HarshitaKalani:
Hey @opixdown , please feel free rebase the PR with main and update the test references. The CI should go green. We can then merge this.

PR #9014, opened 2025-12-11
Fix legacy_array_sections nested ArrayItem recursion

krystophny:
## Summary

Fix `--legacy-array-sections` to not recurse into nested `ArrayItem` expressions in array indices.

## Problem

For `arr(idx(1))` where `idx` is an integer array and `arr` is a real array, the visitor was recursively replacing the inner `idx(1)` first, losing the outer `arr` reference. This caused type mismatch: `i32[:] vs r32[:]`.

## LAPACK Build Progress

| Version | Files Compiled | Failure Point |
|---------|----------------|---------------|
| **BEFORE** (main) | 841/2119 (40%) | `slals0.f` - type mismatch |
| **AFTER** (this PR) | 1461/2119 (69%) | `stfsm.f` - ArrayBound assertion |

## Test

- `lapack_04.f90` - MRE from LAPACK `slals0.f`

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2346,7 +2346,7 @@ RUN(NAME double_complex_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME lapack_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME lapack_02 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface GFORTRAN_ARGS -ffixed-form)
 RUN(NAME lapack_03 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface --cpp-infer GFORTRAN_ARGS -ffixed-form)
-RUN(NAME lapack_04 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface GFORTRAN_ARGS -ffixed-form)
+RUN(NAME lapack_04 LABELS gfortran llvm EXTRA_ARGS --legacy-array-sections --implicit-interface)
 
 RUN(NAME common_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/lapack_04.f90
+++ b/integration_tests/lapack_04.f90
@@ -1,22 +1,22 @@
-C MRE from LAPACK slals0.f: integer array index into real array
-C causes type check assertion: check_equal_type(i32[:], r32[:])
-C Pattern: arr(idx(i)) passed to implicit-interface subroutine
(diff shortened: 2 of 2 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -2344,6 +2344,7 @@ RUN(NAME double_complex_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME lapack_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME lapack_02 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface GFORTRAN_ARGS -ffixed-form)
 RUN(NAME lapack_03 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface --cpp-infer GFORTRAN_ARGS -ffixed-form)
+RUN(NAME lapack_04 LABELS gfortran llvm EXTRA_ARGS --fixed-form --legacy-array-sections --implicit-interface GFORTRAN_ARGS -ffixed-form)
```
certik: Btw, you can just convert it to free form, since the bug here has nothing to do with fixed form, so we can eliminate that constraint from the test.

krystophny: Done

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6687,12 +6687,30 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                    array_index.m_right = ASRUtils::expr_value(array_index.m_right);
+                                }
+                                array_index.m_step = b.i32( j + 1 );
+                                array_indices.push_back(al, array_index);
+                            }
+                            ASR::ttype_t* new_type = ASRUtils::duplicate_type_with_empty_dims(
+                                al, ASRUtils::expr_type(array_item->m_v));
+                            x.m_args[it.first].m_value = ASRUtils::EXPR(ASR::make_ArraySection_t(
+                                al, array_item->base.base.loc, array_item->m_v,
+                                array_indices.p, array_indices.n, new_type, nullptr));
```
certik, line 6713: Doesn't this now always convert the top array to array section? It should only happen with `legacy_array_sections==true`, but I can't see `if (legacy_array_sections) { ...` anywhere here.

certik: It works as expected:
```
$ lfortran --legacy-array-sections --implicit-interface lapack_04.f90
PASS
$ lfortran  --implicit-interface lapack_04.f90
code generation error: Type mismatch in subroutine call, expected `f32[:]`, passed `f32`
  --> lapack_04.f90:11:14
   |
11 |     call sub(arr(idx(1)))
   |              ^^^^^^^^^^^
```
I don't understand how, but it works. :)

certik:
Let's run all the tests just in case. If they pass, we can merge.

src/lfortran/semantics/ast_common_visitor.h
krystophny: > I don't understand how, but it works. :)

I am not sure if this is a good sign :D

PR #9018, opened 2025-12-11
fix: allocate the variable before assignment if value type is function call

### LINKED ISSUE #9013: Handle allocatable type variable assigned to function call returning derived type
MRE:
```fortran
program mre1
    type :: temp
    end type 
    type(temp), allocatable :: x
    x = parse_temp()
contains
    function parse_temp() result(res)
          type(temp) :: res
          allocate(res%arr(2))
    end function
end program
```

LFortran segfaults on it.

HarshilShah1804:
Fixes #9013

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -560,6 +560,7 @@ RUN(NAME functions_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME functions_50 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_51 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_52 LABELS llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME functions_53 LABELS llvm llvm_wasm llvm_wasm_emcc)
 
 
 RUN(NAME common_01 LABELS gfortran)
--- a/integration_tests/functions_53.f90
+++ b/integration_tests/functions_53.f90
@@ -1,0 +1,14 @@
+program functions_53
+    implicit none
+    type :: temp
+      integer, allocatable :: arr(:)
+    end type 
+    type(temp), allocatable :: x
+    x = parse_temp()
+    if (.not. allocated(x)) error stop
+contains
+    function parse_temp() result(res)
+          type(temp) :: res
+          allocate(res%arr(2))
+    end function
+end program
--- a/tests/reference/asr-modules_28-9506bba.json
+++ b/tests/reference/asr-modules_28-9506bba.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-modules_28-9506bba.stdout",
-    "stdout_hash": "e276c614fa138fc9d41c2c38fc961957a5100bcacb40a575d04bf721",
+    "stdout_hash": "57215dbdb592ab6d0e376ae4b6d9ed97a3b2f7abd3bbee495cf4dd1b",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-modules_28-9506bba.stdout
+++ b/tests/reference/asr-modules_28-9506bba.stdout
@@ -804,7 +804,66 @@
                                     )
                                     []
                                     [(Var 9 self)]
-                                    [(Assignment
+                                    [(If
+                                        ()
+                                        (LogicalNot
+                                            (IntrinsicImpureFunction
+                                                Allocated
+                                                [(StructInstanceMember
+                                                    (Var 9 self)
+                                                    9 1_dependency_config_t_git
+                                                    (Allocatable
+                                                        (StructType
+                                                            [(Allocatable
+                                                                (String 1 () DeferredLength DescriptorString)
+                                                            )
+                                                            (Allocatable
+                                                                (String 1 () DeferredLength DescriptorString)
+                                                            )]
+                                                            []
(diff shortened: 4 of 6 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3710,6 +3710,25 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                alloc_arg.m_len_expr = nullptr;
+                alloc_arg.m_sym_subclass = nullptr;
+                alloc_arg.m_type = nullptr;
+                Vec<ASR::alloc_arg_t> alloc_args;
+                alloc_args.reserve(al, 1);
+                alloc_args.push_back(al, alloc_arg);
+                ASR::stmt_t* alloc_stmt = ASRUtils::STMT(ASR::make_Allocate_t(al, target->base.loc,
+                                            alloc_args.p, alloc_args.n,
+                                            nullptr, nullptr, nullptr));
+                current_body->push_back(al, alloc_stmt);
```
certik: I think this is not the right fix. Consider:
```diff
--- a/integration_tests/functions_53.f90
+++ b/integration_tests/functions_53.f90
@@ -4,10 +4,11 @@ program functions_53
       integer, allocatable :: arr(:)
     end type
     type(temp), allocatable :: x
+    allocate(x)
     x = parse_temp()
 contains
     function parse_temp() result(res)
```
Now there are 2 allocate in ASR:
```
[(Allocate
                        [((Var 2 x)
                        []
                        ()
                        ()
                        ())]
                        ()
                        ()
                        ()
                    )
                    (Allocate
                        [((Var 2 x)
                        []
                        ()
                        ()
                        ())]
                        ()
                        ()
                        ()
                    )
                    (Assignment
                        (Var 2 x)
                        (FunctionCall
                            2 parse_temp
                            ()
                            []
                            (StructType
                                [(Allocatable
                                    (Array
                                        (Integer 4)
                                        [(()
                                        ())]
                                        DescriptorArray
                                    )
                                )]
                                []
                                .true.
                                .false.
                            )
                            ()
                            ()
                        )
                        ()
                        .false.
                        .false.
                    )]
```
And two allocations like this should fail in Debug mode (it currently doesn't but that's another bug: https://github.com/lfortran/lfortran/issues/9028).


This has to be fixed in ASR->LLVM. This is related to automatic reallocation of LHS. We do it for scalars, but not arrays.

HarshilShah1804: Okay. I'll change that.

HarshilShah1804: So, upon deeper inspection, I found out that we are not having any assignment node only in the final ASR. In the `subroutine_from_functions` pass, we transform the assignment statement to a subroutine call. So, I was not able to figure out how to handle it in LLVM. Now, I have thought of two approaches:

1. We can either keep the transformation as it is and then handle the unallocated structs in the generating code subroutine call in the backend. 
2. We keep the assignment statement and handle it while generating code for assignment statement.

We also generate wrong Fortran (and probably the ASR even) as the Fortran code generated using `--dump-all-passes-fortarn` also segfault with GFortran. So, I believe we should keep the assignment statement as it is and to get correct ASR and then handle it in the backend similar to automatic reallocation of scalars.

@certik Can you also please suggest which of the two would be a better approach.

HarshilShah1804: This input code:
```fortran
program mre1
    type :: temp
    end type 
    type(temp), allocatable :: x
    x = parse_temp()
contains
    function parse_temp() result(res)
          type(temp) :: res
          allocate(res%arr(2))
    end function
end program
```
gets transformed into this final code (before LLVM):
```fortran
program mre1
implicit none
type :: temp
    integer(4), dimension(:), allocatable :: arr
end type temp
type(temp), allocatable :: x
call parse_temp(x)

contains

subroutine parse_temp(res)
    type(temp), intent(out) :: res
    allocate(res%arr(2))
end subroutine parse_temp

end program mre1
```

certik: One option is something like this:
```fortran
program mre1
implicit none
type :: temp
    integer(4), dimension(:), allocatable :: arr
end type temp
type(temp), allocatable :: x
if (allocated(x)) deallocate(x)
allocate(x)
call parse_temp(x)

contains

subroutine parse_temp(res)
    type(temp), intent(out) :: res
    allocate(res%arr(2))
end subroutine parse_temp

end program mre1
```
Alternatively:
```fortran
type(temp), allocatable :: x
if (.not. allocated(x)) allocate(x)
call parse_temp(x)
```

This last one is the way to do it I think.

certik: Let's also consider when `x` is an array:
```fortran
program mre1
    type :: temp
    end type 
    type(temp), allocatable :: x(:)
    x = parse_temp(5)
contains
    function parse_temp(n) result(res)
          integer, intent(in) :: n
          integer :: i
          type(temp) :: res(n)
          do i = 1, n
              allocate(res(i)%arr(2))
          end do
    end function
end program
```

With `--realloc-lhs-arrays`, we reallocate `x` every time. Without `--realloc-lhs-arrays` we require `x` to be allocated and have the correct size, otherwise we must give a bounds check error.

Note: this already works. Let's see if we have an integration test and error tests for all these cases.

HarshilShah1804: I have made the changes as we discussed. I'll add the test for arrays in a subsequent PR.

certik:
Let's run all tests first.

If all tests pass, then I think we can merge it, the PR is quite simple now.

PR #9023, opened 2025-12-12
Handle missing debug line info file gracefully

krystophny:
## Summary
- Stacktrace: tolerate missing `<prog>_lines.dat.txt` (fd < 0, size==0, NULL read) and fall back to printing raw addresses with a note.


certik:
Do you have some example how it looks before and after?

PR #9024, opened 2025-12-12
Fix assumed-size dummy ABI for LAPACK compatibility

krystophny:
## Summary
Fix array ABI handling for assumed-size dummy arguments to enable LAPACK compilation.

**Core changes:**
- Handle null dimension start/length in array indexing (`asr_to_llvm.cpp`)
- Use `PointerArray` instead of `DescriptorArray` for assumed-size dummy args (`asr_utils.h`)
- Pass `is_proc_arg` flag through type creation for correct ABI selection (`ast_common_visitor.h`)
- Preserve array physical type during template instantiation (`instantiate_template.cpp`)
- Disable bounds checking for dimensions without length info

**ISO/IEC 1539-1:2018 compliance:**
- Section 5.4.6.2: Assumed-size `*` only valid in last dimension
- Arrays with `*` dimension use pointer-based ABI for F77 sequence association

## Test plan
- Reference tests updated for new array physical types
- Enables vanilla Reference-LAPACK build (see #9025)
- Run full test suite: `./run_tests.py` and integration tests

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -6948,6 +6948,11 @@ public:
                             for (size_t i = 0; i < temp_function_dependencies.n; i++) {
                                 current_function_dependencies.push_back(al, temp_function_dependencies[i]);
                             }
+                        } else {
+                            expected_arg_type = ASRUtils::duplicate_type_with_empty_dims(
+                                al, array_arg_idx[i], expected_phys, true
+                            );
+                            array_t = ASR::down_cast<ASR::Array_t>(expected_arg_type);
                         }
 
                         ASR::asr_t* expected_array = ASR::make_Array_t(al, loc, ASRUtils::type_get_past_array(expected_arg_type),
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2499,6 +2499,7 @@ RUN(NAME legacy_array_sections_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc 
 RUN(NAME legacy_array_sections_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray EXTRA_ARGS --legacy-array-sections)
 RUN(NAME legacy_array_sections_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray EXTRA_ARGS --legacy-array-sections)
 RUN(NAME legacy_array_sections_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray EXTRA_ARGS --legacy-array-sections)
+RUN(NAME legacy_array_sections_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray EXTRA_ARGS --legacy-array-sections)
 
 RUN(NAME cmake_minimal_test_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME char_array_initialization_declaration LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/legacy_array_sections_09.f90
+++ b/integration_tests/legacy_array_sections_09.f90
@@ -1,0 +1,35 @@
+program legacy_array_sections_09
+    implicit none
+
+    integer :: n, lwa
+    real, allocatable :: wa(:)
+
+    n = 3
+    lwa = 10
+    allocate(wa(lwa))
+    wa = 0.0
+
+    call caller(n, lwa, wa)
+
+contains
+
+    subroutine callee(n, diag)
+        implicit none
+
+        integer, intent(in) :: n
+        real, intent(inout) :: diag(n)
+
+        diag = diag
+    end subroutine callee
+
+    subroutine caller(n, lwa, wa)
+        implicit none
+
+        integer, intent(in) :: n
+        integer, intent(in) :: lwa
+        real, intent(inout) :: wa(lwa)
+
+        call callee(n, wa(1))
+    end subroutine caller
+
+end program legacy_array_sections_09


krystophny:
# #9024 review summary (LAPACK assumed-size)

**What this PR is for**
- Make LFortran compatible with real LAPACK patterns around assumed-size dummies (`*`) + sequence association, primarily under `--legacy-array-sections`.

**Main code changes**
- `src/lfortran/semantics/ast_common_visitor.h`: keep dummy-argument context (ABI/physical type) across all declaration routes (attributes-only placeholders, `DIMENSION` statements, etc.). Prevents assumed-size dummies from silently becoming `DescriptorArray`.
- `src/libasr/asr_utils.h`: ensure `DIMENSION(*)` dummy arguments use the expected pointer-style physical type.
- `src/libasr/pass/instantiate_template.cpp`: preserve array physical types through template instantiation/casts (needed by `template_lapack_01`).
- `src/libasr/codegen/asr_to_llvm.cpp`: small supporting adjustments for assumed-size behavior.

**New tests (LAPACK-shaped MREs)**
- `integration_tests/lapack_05.f90`: LAPACK-style `A(LDA, *)`/`B(LDB, *)` dummy ABI.
- `integration_tests/lapack_06.f90`: from real `stfsm.f` style, 0-based assumed-size + `call sub(A(M))` sequence association.

**Evidence (local, LLVM11)**
- `stfsm.f` compiles
- `lapack_05` PASS
- `lapack_06` PASS

**Alternatives considered (prior LAPACK work)**
- Codegen-only bitcast hacks (LLVM11 typed pointers): rejected. Can silence verifier/type mismatches but is semantically wrong (can pass a descriptor address as if it were a data pointer and corrupt memory).
- Disable checks / weaken tests (e.g. bounds checking off): rejected. Hides the bug rather than fixing LAPACK compatibility.
- Sweeping mega-patch for all LAPACK failures: rejected. We stick to the MRE-driven approach (one LAPACK-shaped reproducer + minimal compiler fix), so review stays tractable.
- Touch unrelated subsystems to make CI green: rejected/split. Example: `array_constructor_02` fallout was split into #9035.

certik:
Well done. Lots of changes and I see some changes from Descriptor to Pointer and FixedSize to Pointer, etc. So I will have to review manually. If there is a way to split this big PR into smaller PRs, that would be great. If not, we'll review it as is, but it will take some time.

It looks good as an idea, but we need to make sure we didn't break anything.

krystophny:
> Well done. Lots of changes and I see some changes from Descriptor to Pointer and FixedSize to Pointer, etc. So I will have to review manually. If there is a way to split this big PR into smaller PRs, that would be great. If not, we'll review it as is, but it will take some time.
> 
> It looks good as an idea, but we need to make sure we didn't break anything.

Thanks! Let's see, yes. I fear this cannot become much smaller, because it is one coherent change. Most lines changed are in tests and references, so hope it is not too much effort to check.

krystophny:
Tried some splitting once more into multiple PRs like #9039 but no luck. This is again a monolith.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2117,12 +2125,21 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+	    for (size_t i=0; i<n_dim; i++) {
+	        if (is_argument &&
+	                m_dim[i].m_end_star == AST::dimension_typeType::DimensionStar
+	                && i != n_dim-1) {
+	            diag.add(diag::Diagnostic(
+	                "Assumed-size '*' is only permitted in the last dimension",
+	                diag::Level::Error, diag::Stage::Semantic, {
+	                    diag::Label("", {m_dim[i].loc})}));
+	            throw SemanticAbort();
+	        }
```
certik, line 2138: Let's add a test for this into continue_compilation_1.f90

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4173,6 +4210,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     } else if ( is_implicitly_declared ) {
                         ASR::symbol_t* symbol = current_scope->get_symbol(sym);
                         ASR::Variable_t* symbol_variable = ASR::down_cast<ASR::Variable_t>(symbol);
+                        if (is_argument && is_dimension_star) {
+                            symbol_variable->m_type = type;
+                        } else
                         if ( ASR::is_a<ASR::Array_t>(*symbol_variable->m_type) ) {
```
certik, line 4216: Write it as `else if` on the same line.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4096,6 +4130,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             ASR::Variable_t* symbol_variable = ASR::down_cast<ASR::Variable_t>(symbol);
                             symbol_variable->base.base.loc = s.loc;
                             if ( symbol_variable->m_type ) {
+                                if (is_argument && is_dimension_star) {
+                                    symbol_variable->m_type = type;
+                                } else
                                 if ( ASR::is_a<ASR::Array_t>(*symbol_variable->m_type) ) {
```
certik, line 4136: Write it as `else if` on the same line.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6268,16 +6285,46 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                } else {
+                    llvm::Value *lbound = builder->CreateSExtOrTrunc(lbs.p[i], llvm::Type::getInt32Ty(context));
+                    if (m_dims[i].m_length) {
+                        visit_expr_wrapper(m_dims[i].m_length, true);
+                        llvm::Value *length = builder->CreateSExtOrTrunc(tmp, llvm::Type::getInt32Ty(context));
+                        ubs.p[i] = builder->CreateSub(builder->CreateAdd(lbound, length),
+                            llvm::ConstantInt::get(llvm::Type::getInt32Ty(context), 1));
+                    } else {
+                        // Assumed-size dimension - ubound is undefined, avoid crash and use lbound.
+                        ubs.p[i] = lbound;
```
certik, line 6318: The comment is not clear -- what kind of crash are we avoiding, and is using lbound just a temporary workaround or is that the correct solution and if so, then I would not even mention ubound, since that was incorrect anyway.

src/libasr/pass/inline_function_calls.cpp
```diff
@@ -177,6 +177,14 @@ class InlineFunctionCalls: public ASR::BaseExprReplacer<InlineFunctionCalls> {
                 return false;
             }
 
+            // Don't inline functions with assumed-size array parameters
+            // (PointerArray with empty dimensions)
+            ASR::ttype_t* var_type = ASR::down_cast<ASR::Variable_t>(sym.second)->m_type;
+            if( ASRUtils::is_array(var_type) &&
+                ASRUtils::is_dimension_empty(var_type) ) {
+                return false;
+            }
```
certik, line 186: This stops inlining assumed-shape arrays also. Rather, we only want to stop inlining assumed-size.

src/libasr/asr_utils.h
```diff
@@ -3155,8 +3155,8 @@ inline ASR::ttype_t* make_Array_t_util(Allocator& al, const Location& loc,
                 }
             } else if( !ASRUtils::is_dimension_empty(m_dims, n_dims) ) {
                 physical_type = ASR::array_physical_typeType::PointerArray;
-            } else if ( is_dimension_star && ASRUtils::is_only_upper_bound_empty(m_dims[n_dims-1]) ) {
-                physical_type = ASR::array_physical_typeType::UnboundedPointerArray;
+            } else if (is_dimension_star && is_argument) {
+                physical_type = ASR::array_physical_typeType::PointerArray;
```
certik, line 3159: I think this needs to be:
```suggestion
                physical_type = ASR::array_physical_typeType::UnboundedPointerArray;
```
The physical type for `A(*)` is UnboundedPointerArray.

certik:
There is also an issue that I think we now somehow use 0 as the dimensions or something like that for PointerArray and that we allow the dimensions to be unspecified. All mistakes that we'll need to undo.

PR #9025, opened 2025-12-12
Add vanilla Reference-LAPACK to third-party tests

krystophny:
## Summary
Add Section 14 to third-party tests for vanilla Reference-LAPACK v3.12.0:

- Clone official Reference-LAPACK repository (not patched fork)
- Apply minimal `sed` patch to skip `FortranCInterface_VERIFY`
- Build BLAS and LAPACK libraries with LFortran flags:
  `--fixed-form-infer --implicit-interface --legacy-array-sections --separate-compilation`
- Run DGESV linear solve test to verify runtime correctness

This validates that LFortran can build and use **unmodified** LAPACK.

## Dependencies
- Requires #9024 (Fix assumed-size dummy ABI) to be merged first
- Will need rebase after #9024 merges

## Test plan
- CI will run the LAPACK build and DGESV test
- Verify LAPACK libraries compile without errors
- Verify DGESV solves a 3x3 linear system correctly

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -934,7 +934,7 @@ program test_dgesv
 end program
 TESTEOF
 
-    lfortran --implicit-interface test_dgesv.f90 -L build/lib -lblas -llapack -o test_dgesv
+    lfortran --implicit-interface test_dgesv.f90 -L build/lib -llapack -lblas -o test_dgesv
     ./test_dgesv
 '
 


krystophny:
This should right now fail, until rebased on or getting fix merged from #9024 . That's a good baseline check.

certik:
Yes, let me see the failure, then I'll rebase.

I am also testing the rebased version here: https://github.com/lfortran/lfortran/pull/9065.

certik:
The failure was caused by another PR, I reverted it. I rebased, let's see if it passes now.

certik:
Ok, we are back in the game, the CI works. Just the LAPACK tests don't for some reason.

certik:
I think the failure is regarding cmake version. @krystophny which cmake version used to work for your?

I can reproduce the failure locally too with 3.29.1.

krystophny:
> Ok, we are back in the game, the CI works. Just the LAPACK tests don't for some reason.
> I think the failure is regarding cmake version. @krystophny which cmake version used to work for your?
> 
> I can reproduce the failure locally too with 3.29.1.

4.2.1 . I try to downgrade and fix it.

krystophny:
@certik passes locally for me now with the custom cmake toolchain config. This has to be set before command line flags this is why it is in an extra file. From cmake 3.31 on the correct preprocessor syntax should be recognized correctly.

krystophny:
Done!

certik:
I think it works! That's incredible.

PR #9031, opened 2025-12-12
Default: -g enables line/column debug info

krystophny:
## Summary
- Make `-g` automatically enable `--debug-with-line-column` so debug builds always include line/column location info.
- Prevents “emit_debug_line_column is not enabled” errors when users or CMake add `-g` implicitly.

## Test plan
- `src/bin/lfortran -g -c integration_tests/print_01.f90` (no extra flags; succeeds).
- `cmake --build lfortran/build` (updates binaries with the new default).


krystophny:
Does this make more sense than throwing an error and requiring another CLI flag? I ran into this often when running with cmake `Debug` settings.

certik:
This PR makes `emit_debug_line_column` useless, as it is always on together with `-g`, so just remove it completely. We'll only use `-g` and if enabled, the linear index always gets converted to line/column. It's slower, but `-g` is already slower and we can implement the linear index later --- the idea was that we would only convert the linear index to line/column in our runtime to print a stacktrace. But using the default debug info requires line/column, so let's just use it by default.

krystophny:
> This PR makes `emit_debug_line_column` useless, as it is always on together with `-g`, so just remove it completely. We'll only use `-g` and if enabled, the linear index always gets converted to line/column. It's slower, but `-g` is already slower and we can implement the linear index later --- the idea was that we would only convert the linear index to line/column in our runtime to print a stacktrace. But using the default debug info requires line/column, so let's just use it by default.

Done!

Issue #9030, opened 2025-12-12
`lfortran tests/errors/continue_compilation_1.f90 --continue-compilation` segfaults


certik:
```console
$ lfortran tests/errors/continue_compilation_1.f90 --continue-compilation
...
semantic error: Assigned format (using integer variable as format specifier) is not supported. Use character variables instead.
   --> tests/errors/continue_compilation_1.f90:451:5
    |
451 |     print fmt_i2, "test"
    |     ^^^^^^^^^^^^^^^^^^^^

semantic error: Assigned format (using integer variable as format specifier) is not supported. Use character variables instead.
   --> tests/errors/continue_compilation_1.f90:456:5
    |
456 |     read (5, fmt_i3)
    |     ^^^^^^^^^^^^^^^^


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
Traceback (most recent call last):
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2737
    LCompilers::print_stack_on_segfault();
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2664
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 1194
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/ondrej/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 394
    if (res.ok) {
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 15118
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/pass_manager.h", line 318
    apply_passes(al, asr, _passes, pass_options, diagnostics, cummulative_time_taken_by_passes_in_microseconds);
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/pass_manager.h", line 189
    _passes_db[passes[i]](al, *asr, pass_options);
  File "/Users/ondrej/repos/lfortran/src/libasr/pass/array_op.cpp", line 1231
    v.call_replacer_on_value = false;
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_expr_call_replacer_visitor.h", line 39
    for (auto &a : x.m_symtab->get_scope()) {
[...]

PR #9033, opened 2025-12-12
Fix malformed ASR in continue compilation

certik:
Fixes the MRE in https://github.com/lfortran/lfortran/issues/9030#issuecomment-3647255956, but doesn't fix the full issue, as there is another bug after that.

I verified that the added test catches this segfault if the fix in this PR is not applied.


certik:
The fix just shuffles the code around to only create ASR if it is valid, otherwise not at all, it doesn't change or add any other logic, so I think it's fine to merge. It's tested and all tests pass.

PR #9034, opened 2025-12-12
Fix malformed ASR in continue compilation

### LINKED ISSUE #9030: `lfortran tests/errors/continue_compilation_1.f90 --continue-compilation` segfaults

certik:
Builds on top of #9033. Fixes #9030.

### DIFF
--- a/tests/reference/run-mre_parameter-15f810f.json
+++ b/tests/reference/run-mre_parameter-15f810f.json
@@ -1,0 +1,13 @@
+{
+    "basename": "run-mre_parameter-15f810f",
+    "cmd": "lfortran --continue-compilation --no-color {infile}",
+    "infile": "tests/errors/mre_parameter.f90",
+    "infile_hash": "98f39f3110f80da2ce3dcae7c3d1a1bf53190249d4cb0d1f39f95e4e",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "run-mre_parameter-15f810f.stderr",
+    "stderr_hash": "01c4dff897b57fbb9191fb41026044183854c84d51c90c5b2231e15e",
+    "returncode": 1
+}
--- a/tests/reference/run-mre_parameter-15f810f.stderr
+++ b/tests/reference/run-mre_parameter-15f810f.stderr
@@ -1,0 +1,11 @@
+semantic error: arguments of `repeat` intrinsic must be scalar
+ --> tests/errors/mre_parameter.f90:3:38
+  |
+3 |     character(3), parameter :: ar1 = repeat(["abc", "#^1", "123"], [1, 2, 3])
+  |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+semantic error: Variable 'ar1' is not declared
+ --> tests/errors/mre_parameter.f90:4:14
+  |
+4 |     print *, ar1
+  |              ^^^ 'ar1' is undeclared


certik:
Currently this is using try/except and I don't like that. I would prefer a solution that doesn't need to try/except, but let's at least check if CI passes with this. If it works, then we have a reference implementation here and we can try to fix it in a better way.

certik:
Looks like everything passes. But I don't want to merge it, as I don't like the try/catch. Rather, we need to save the variable to a local temporary until we resolve the initial value, and only then insert into the symbol table. Alternatively we insert it right away, but not set it as parameter, and then if we successfully convert the initial value, we can assign it, and change the variable to a parameter. This should avoid the try/catch approach.

I'll keep this PR open as a reference how to fix it.

certik:
So far I haven't figured out a better solution without a large refactor -- the two use cases to covert: self-referential initialization (the variable must be in a symbol table to succeed) and the above case. The try/catch might be the cleanest solution after all, a long as we use exceptions to propagate errors.

certik (commented):
I am going to merge it --- I can't figure out a better approach, and it's documented what the try/catch is doing and it fixes a bug, so let's merge it. We can always improve it later.

PR #9035, opened 2025-12-12
Fix: correct ArrayConstructor sizing for array args

krystophny:
## Summary

Fix `ArrayConstructor` size calculation when an element is itself an array expression (e.g., an allocatable array component like `target%dependencies`).

**The bug**: Previously, all non-ArraySection elements were counted as `+1` regardless of whether they were scalars or arrays. This caused incorrect RHS shape for downstream codegen/realloc decisions.

**The fix** (in `asr_utils.cpp` and `implied_do_loops.cpp`):
- Check if element is an array type
- If fixed-size array: add full array size to `constant_size`
- If dynamic array: compute size at runtime via `get_size()` 
- Only count `+1` for actual scalar elements

Also fixes assignment bug where `builder.Add()` result was not assigned back.

## Test

`integration_tests/array_constructor_02.f90` - now runs with bounds checking enabled (previously required `--no-array-bounds-checking` workaround).

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -770,7 +770,7 @@ RUN(NAME array_section_is_non_allocatable LABELS gfortran llvm llvm_wasm llvm_wa
 RUN(NAME array_indices_array_section LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_indices_array_section_assignment LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_constructor_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME array_constructor_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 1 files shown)


certik:
Thanks for fixing it!

For some reason the array_constructor_02 fails at the CI now.

krystophny:
Ah our good friend array reallocation

```fortran
allocate(target%dependencies(2))

target%dependencies = [target%dependencies, build_target_ptr(dependency)]
```

Hope it passes now.

certik:
You don't need reallocation in this case, correct? It's allocated to 2 and you are assigning 2. So the `--realloc-lhs-arrays` option is hiding a bug.

krystophny:
> You don't need reallocation in this case, correct? It's allocated to 2 and you are assigning 2. So the `--realloc-lhs-arrays` option is hiding a bug.

Wouldnt that one make a size 3 array? So concatenate the array of size two `target%dependencies` with another single entry `build_target_ptr(dependency)` ? Unfortunately much code in `fpm` and other packages relies on such list extension as far as I remember.

certik:
Oh, wait, I missed it:
```
target%dependencies = [target%dependencies, build_target_ptr(dependency)]
```
It's appending to it. Yes, that requires `--realloc-lhs-arrays`.

Issue #4547, opened 2024-07-28
Should we throw an error for overflow in `log_gamma` or `inf` output works?


HarshitaKalani:
```fortran
program main
    print*, log_gamma(0.0)
end
```
```console
$ gfortran a.f90 && ./a.out
a.f90:2:22:

    2 |     print*, log_gamma(0.0)
      |                      1
Error: Result of LGAMMA overflows its kind at (1)
$ lfortran a.f90 
          inf
```

HarshitaKalani:
>Shall be of type REAL and neither zero nor a negative integer.

The definition here: https://gcc.gnu.org/onlinedocs/gcc-4.9.4/gfortran/LOG_005fGAMMA.html says that the input cannot be 0.

harperjf:
This version of @HarshitsKalani's log_gamma program has the same error as the original one but it can't be detected at compile time. Both gfortran and lfortran do detect it at run time. As compilers are free to choose their own format for 
 ```print *, log_gamma(0.0)``` I used an explicit format, hoping for the output ```Inf``` with an upper case ```I``` required by the F2023 standard. This is what I got. Lfortran seems to have a bug with format ```"(F0.0)"```
```
(lf) john:~/Test$ cat badlog_gamma.F90
program main
  implicit none
  real x
  x = 0.0
  print "(F0.0)", log_gamma(x)
end program main
(lf) john:~/Test$ gfortran-13 badlog_gamma.F90 && ./a.out
Inf
(lf) john:~/Test$ lfortran badlog_gamma.F90
-9223372036854775807.f
(lf) john:~/Test$

**Comment by HarshitaKalani** [comment]
Does `Infinity` work too? Or we need to print `Inf`?
```console
 % lfortran b.f90
Infinity
% gfortran b.f90 && ./a.out
Inf
```

harperjf:
F2023 standard 13.7.2.3.2 para 9 explains when to write 'Inf', 'Infinity' 
or '***' with or without '+' or '-' if using F format. ES, EN, G or * 
formats all refer back to Fw.d format with appropriate values of w.

n Sun, 15 Mar 2026, Harshita Kalani wrote:

> Date: Sun, 15 Mar 2026 08:57:53 +0000
> From: Harshita Kalani ***@***.***>
> Reply-To: lfortran/lfortran
>     ***@***.***>
> To: lfortran/lfortran ***@***.***>
> Cc: harperjf ***@***.***>, Comment ***@***.***>
[...]

PR #9037, opened 2025-12-12
Fix: Allow passing unallocated allocatable arrays to subroutines

amritamishra01:
Description:
Summary This PR fixes a bug where passing an unallocated allocatable array (specifically a derived type member) to a subroutine caused a compiler crash (Segmentation Fault) or an incorrect Runtime Error. This behavior diverged from standard Fortran (e.g., gfortran), where passing an unallocated array is valid.
The Issue Previously, when compiling code like call set_list(mylist%items), visit_ArrayItem would attempt to:
Enforce a strict "is allocated" runtime check, causing a Runtime Error.
Generate indexing/slicing code that loaded the array's data pointer. Since the array was unallocated, this pointer was NULL, leading to a Segmentation Fault during execution.
The Fix I updated src/libasr/codegen/asr_to_llvm.cpp with the following changes:
visit_ArrayItem: Added a check for ASRUtils::is_allocatable. If the array is allocatable, we now return the descriptor pointer directly (tmp = array; return;) instead of attempting to load the value or generate indexing code. This ensures the subroutine receives the correct address of the unallocated descriptor.
visit_DebugCheckArrayBounds: Relaxed the assignment check to allow unallocated targets if the variable type is allocatable.
Test Plan
Added a new integration test: tests/test_allocatable_pass.f90 (based on the MRE).
Verified the output matches gfortran:
list is UNALLOCATED
Ran ctest to ensure no regressions in existing tests.
Closes: #4547

### DIFF
--- a/tests/reference/run-matmul_unallocated_01-39026d3.json
+++ b/tests/reference/run-matmul_unallocated_01-39026d3.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "run-matmul_unallocated_01-39026d3.stderr",
-    "stderr_hash": "d3367abc6f61984f911c5138d0a113d38e39d64503264300bf151f65",
+    "stderr_hash": "e15d4ac88994cfa2076bcbf0fbe3f42b6c07725838ac1a4d8be7feac",
     "returncode": 1
 }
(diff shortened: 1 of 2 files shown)


integration_tests/test_allocatable_pass.f90
```diff
@@ -0,0 +1,26 @@
...
+    end subroutine set_list
+end module m
+
+program demo
+    use m
+    implicit none
+    type(string_list_t) :: mylist
+    ! This causes the crash because the compiler tries to read 'items' data before passing it
+    call set_list(mylist%items)
+end program demo
```
certik, line 26: If this test is valid Fortran (runs with GFortran, for example), then put it into integration_tests and register in cmake there, and make it run with LFortran as well.

Issue #7952, opened 2025-07-02
move_alloc wrong


harperjf:
This program worked correctly with gfortran, ifx and AMD flang but lfortran gave wrong output. 
```
program badalloc ! file badalloc.f90 
  implicit none ! based on @Ncarlson 12-31-2024 bug report to Intel
  type,abstract :: foo
  end type foo
  type, extends(foo) :: bar
  end type bar
  type(bar), allocatable :: x
  class(foo), allocatable :: y
  call move_alloc(x, y)
  if(allocated(y)) error stop 'Y SHOULD BE UNALLOCATED!'
  print "(A)", 'Y IS UNALLOCATED.'
end program badalloc
```
Wrong output from lfortran  0.54.0:
```
ERROR STOP Y SHOULD BE UNALLOCATED!
```
Correct output from 3 other compilers:
```
Y IS UNALLOCATED.
```

HarshilShah1804:
With the latest main, I am getting a segfault, rather than `ERROR STOP Y SHOULD BE UNALLOCATED!`

```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
LCOMPILERS_ASSERT failed: /home/harshil-shah/lfortran/src/libasr/asr_utils.h
function replace_StructType(), line number 4609 at 
derived_type_sym != nullptr
ERROR MESSAGE:
derived_type_sym cannot be nullptr
Traceback (most recent call last):
  Binary file "/home/harshil-shah/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2729, in main
    return main_app(argc, argv);
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2653, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
[...]

PR #9039, opened 2025-12-12
fix: runtime error for double allocation in Debug mode

krystophny:
Merge order: #9049 -> #9050 -> #9048 -> #9039

## Summary
- Add a runtime error for double `allocate()` of **allocatable** variables in debug/bounds-checking mode.
- Avoid false-positives for non-allocatable dummy descriptors (pointer/descriptor arrays).

## CI note (FAST)
`integration_tests/allocate_29` is a negative test: it must abort on the 2nd `allocate()`.
The `--fast` test variant can skip runtime allocation checks, so `allocate_29_FAST` can reach the should-not-reach-here prints.
Mark `allocate_29` as `NO_FAST` so the expected-failure behavior is still tested in non-FAST runs.

## Prerequisites (split out for review)
- #9049
- #9050
- #9048

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1358,8 +1358,14 @@ public:
                           ASR::is_a<ASR::Complex_t>(*curr_arg_m_a_type) ||
                           ASR::is_a<ASR::Logical_t>(*curr_arg_m_a_type)) {
                     llvm::Type* llvm_arg_type = llvm_utils->get_type_from_ttype_t_util(curr_arg.m_a, curr_arg_m_a_type, module.get());
-                    // Skip double allocation check for intent(out) arguments since Fortran
-                    // standard requires automatic deallocation at procedure entry (not yet implemented)
+                    // Skip double allocation check for INTENT(OUT) allocatables.
+                    //
+                    // Fortran requires allocatable INTENT(OUT) dummy arguments to be automatically
+                    // deallocated on procedure entry (F2018 15.5.2.13). LFortran currently inserts
+                    // this deallocation only in limited cases (see the `insert_deallocate` pass),
+                    // and it is not complete yet (#9097). Until INTENT(OUT) deallocation is fully
+                    // implemented, keeping the double-allocation runtime check enabled here would
+                    // report false positives in valid programs.
                     bool is_intent_out = false;
                     if (ASR::is_a<ASR::Var_t>(*tmp_expr)) {
                         ASR::symbol_t* sym = ASRUtils::symbol_get_past_external(
@@ -1656,8 +1662,14 @@ public:
                 if (x_arr && x_arr->getType() == nullptr) {
                     ptr_val = llvm::ConstantPointerNull::get(static_cast<llvm::PointerType*>(i8_ptr_ty));
                 }
-                // Skip double allocation check for intent(out) arguments since Fortran
-                // standard requires automatic deallocation at procedure entry (not yet implemented)
+                // Skip double allocation check for INTENT(OUT) allocatables.
+                //
+                // Fortran requires allocatable INTENT(OUT) dummy arguments to be automatically
+                // deallocated on procedure entry (F2018 15.5.2.13). LFortran currently inserts
+                // this deallocation only in limited cases (see the `insert_deallocate` pass),
+                // and it is not complete yet (#9097). Until INTENT(OUT) deallocation is fully
+                // implemented, keeping the double-allocation runtime check enabled here would
+                // report false positives in valid programs.
                 bool is_intent_out = false;
                 if (ASR::is_a<ASR::Var_t>(*tmp_expr)) {
                     ASR::symbol_t* sym = ASRUtils::symbol_get_past_external(
--- a/src/libasr/pass/insert_deallocate.cpp
+++ b/src/libasr/pass/insert_deallocate.cpp
@@ -286,7 +286,22 @@ class LoopTempVarDeallocateVisitor : public ASR::BaseWalkVisitor<LoopTempVarDeal
         }
 };
 
-// Deallocate intent(out) allocatable arguments at function entry
+// Deallocate allocatable `intent(out)` dummy arguments at function entry.
+//
+// Notes / limitations:
+//
+// - This pass only inserts an explicit `deallocate()` statement guarded by
+//   `allocated()` (and additionally by `present()` for optional dummies).
+//   Correct finalization / deep deallocation semantics are handled downstream by
+//   the runtime/codegen and are currently incomplete in some cases (#9097).
+//
+// - It only considers dummy arguments that appear as `Var` entries in
+//   `Function_t::m_args` and are `allocatable` with `intent(out)`. It does not
+//   handle pointers, components, or more complex argument expressions.
+//
+// - We intentionally skip compiler-generated intrinsic implementations
+//   (`deftype == Implementation`) to avoid changing their internal ownership
+//   conventions.
 class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDeallocateVisitor>
 {
     Allocator &al;
@@ -408,9 +423,6 @@ public:
 
 void pass_insert_deallocate(Allocator &al, ASR::TranslationUnit_t &unit,
                                 const PassOptions &/*pass_options*/) {
-    // InsertDeallocate v(al);
-    // v.visit_TranslationUnit(unit);
-
     // Deallocate intent(out) allocatable arguments at function entry
     IntentOutDeallocateVisitor iod(al);
     iod.visit_TranslationUnit(unit);


krystophny:
## Technical summary

The patch adds runtime checks in `visit_AllocateUtil` (LLVM codegen) that trigger before `allocate` statements:

**Scalars** (Integer, Real, Complex, Logical):
- Load the current pointer value from the allocatable variable
- Compare against null pointer
- If non-null → variable already allocated → emit runtime error

**Arrays** (n_dims > 0):
- Load the descriptor pointer from variable storage
- Call `arr_descr->get_is_allocated_flag()` which checks:
  1. Is descriptor pointer non-null?
  2. Is data pointer within descriptor non-null?
- If both true → array already allocated → emit runtime error

**Conditions:**
- Only when `!realloc` (skip for `reallocate`)
- Only when `compiler_options.po.bounds_checking` is enabled (default on, disable with `--no-array-bounds-checking`)

krystophny:
## CI Fix Required: Update Reference File

The CI is failing because the new allocation check code changes the LLVM IR output, causing a reference file mismatch for `tests/reference/llvm-allocate_03-495d621.stdout`.

### To fix with LLVM 11 (matching CI):

```bash
# 1. Set up micromamba environment with LLVM 11
curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba
./bin/micromamba create -p ./lf11 -c conda-forge -f ci/environment.yml llvmdev=11 -y

# 2. Build lfortran with LLVM 11
./bin/micromamba run -p ./lf11 cmake -S . -B build11 -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON
./bin/micromamba run -p ./lf11 cmake --build build11 -j

# 3. Update the reference file
./bin/micromamba run -p ./lf11 bash -c "cd tests && PATH=../build11/src/bin:\$PATH python ../run_tests.py -t allocate_03 -u"

# 4. Commit and push the updated reference
git add tests/reference/llvm-allocate_03-495d621.*
git commit -m "Update llvm-allocate_03 reference for new allocation check"
git push
```

The key is using LLVM 11 because it produces typed pointers (`i32*`) matching CI, while newer LLVM versions produce opaque pointers (`ptr`).

certik:
Nice, thanks for fixing that! Let's see if tests pass, then we'll figure out how to handle the commented out tests.

krystophny:
> Nice, thanks for fixing that! Let's see if tests pass, then we'll figure out how to handle the commented out tests.

Yeah, as usual one change uncovered two more bugs.

krystophny:
> The following tests FAILED:
> Errors while running CTest
> Output from these tests are in: /Users/runner/work/lfortran/lfortran/integration_tests/build-lfortran-llvm/Testing/Temporary/LastTest.log
> Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
> 	1163 - derived_types_82 (Subprocess aborted)

Is this a test runner issue or a real bug?

krystophny:
Cherry-picked fix for #7952 (move_alloc destination deallocation). Exact patch:

```diff
commit 9d93f375860dfcd6681eab027f836ac093f3423f
Author: Christopher Albert <albert@tugraz.at>
Date:   Sat Dec 13 01:35:00 2025 +0100

    Fix move_alloc: deallocate destination first
    
    - src/libasr/pass/intrinsic_subroutines.h: ensure move_alloc() deallocates already-allocated destination
    - integration_tests/CMakeLists.txt: re-enable intrinsics_334/intrinsics_378 LLVM labels (was #7952)

diff --git a/integration_tests/CMakeLists.txt b/integration_tests/CMakeLists.txt
index df5951798..86a2503fa 100644
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1219,8 +1219,7 @@ RUN(NAME intrinsics_330 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_331 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # cshift
 RUN(NAME intrinsics_332 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME intrinsics_333 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-# FIXME #7952: move_alloc doesn't deallocate destination first. Re-add: llvm llvm_wasm llvm_wasm_emcc
-RUN(NAME intrinsics_334 LABELS gfortran) # move_alloc
+RUN(NAME intrinsics_334 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # move_alloc
 RUN(NAME intrinsics_335 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #kind, real
 RUN(NAME intrinsics_336 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #any
 RUN(NAME intrinsics_337 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # dble, dfloat, float, shifta
@@ -1264,8 +1263,7 @@ RUN(NAME intrinsics_374 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME intrinsics_375 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # move_alloc for string
 RUN(NAME intrinsics_376 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # present
 RUN(NAME intrinsics_377 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # minval, maxval
-# FIXME #7952: move_alloc doesn't deallocate destination first. Re-add: llvm llvm_wasm llvm_wasm_emcc
-RUN(NAME intrinsics_378 LABELS gfortran) # move_alloc
+RUN(NAME intrinsics_378 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # move_alloc
 RUN(NAME intrinsics_379 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # get_command_argument
 RUN(NAME intrinsics_380 LABELS gfortran llvm) # get_environment_variable
 RUN(NAME intrinsics_381 LABELS gfortran llvm) # get_environment_variable
diff --git a/src/libasr/pass/intrinsic_subroutines.h b/src/libasr/pass/intrinsic_subroutines.h
index dec01ccaf..a1424faaf 100644
--- a/src/libasr/pass/intrinsic_subroutines.h
+++ b/src/libasr/pass/intrinsic_subroutines.h
@@ -1260,13 +1260,31 @@ namespace MoveAlloc {
             fill_func_arg_sub("to", arg_types[1], InOut);
         }
 
-        Vec<ASR::expr_t*> allocated_args; allocated_args.reserve(al, 1);
-        allocated_args.push_back(al, args[0]);
+        Vec<ASR::expr_t*> allocated_args_from; allocated_args_from.reserve(al, 1);
+        allocated_args_from.push_back(al, args[0]);
 
         // ASR node for `allocated(from)`
         ASR::expr_t* is_from_allocated = ASRUtils::EXPR(ASR::make_IntrinsicImpureFunction_t(al, loc,
                 static_cast<int64_t>(IntrinsicImpureFunctions::Allocated),
-                allocated_args.p, allocated_args.n, 0, logical, nullptr));
+                allocated_args_from.p, allocated_args_from.n, 0, logical, nullptr));
+
+        Vec<ASR::expr_t*> allocated_args_to; allocated_args_to.reserve(al, 1);
+        allocated_args_to.push_back(al, args[1]);
+
+        // ASR node for `allocated(to)`
+        ASR::expr_t* is_to_allocated = ASRUtils::EXPR(ASR::make_IntrinsicImpureFunction_t(al, loc,
+                static_cast<int64_t>(IntrinsicImpureFunctions::Allocated),
+                allocated_args_to.p, allocated_args_to.n, 0, logical, nullptr));
+
+        // If `to` is already allocated, deallocate it first.
+        // This matches standard MOVE_ALLOC semantics and avoids double-allocation at runtime.
+        Vec<ASR::expr_t*> deallocate_to_args; deallocate_to_args.reserve(al, 1);
+        deallocate_to_args.push_back(al, args[1]);
+        ASR::stmt_t* deallocate_to = ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(
+            al, loc, deallocate_to_args.p, deallocate_to_args.n));
+        std::vector<ASR::stmt_t*> deallocate_to_body;
+        deallocate_to_body.push_back(deallocate_to);
+        body.push_back(al, b.If(is_to_allocated, deallocate_to_body, {}));
 
         std::vector<ASR::stmt_t*> if_body;
 
```

krystophny:
Cherry-picked fix for #9042 (matmul / intrinsic array function reallocation). Exact patch:

```diff
commit 106d5ec644151f8cb041724dfb75eaeebf5d23d9
Author: Christopher Albert <albert@tugraz.at>
Date:   Sat Dec 13 01:37:07 2025 +0100

    Fix matmul: deallocate allocatable result before allocate
    
    - src/libasr/pass/intrinsic_array_function_registry.h: handle reallocation for allocatable result argument (fixes #9042)
    - integration_tests/CMakeLists.txt: re-enable matmul_01 LLVM labels (was #9042)

diff --git a/integration_tests/CMakeLists.txt b/integration_tests/CMakeLists.txt
index df5951798..449543944 100644
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2488,8 +2488,7 @@ RUN(NAME precision_03 LABELS gfortran llvm)
 RUN(NAME array_unbounded_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME array_unbounded_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 
-# FIXME #9042: Intrinsic array functions don't handle reallocation. Re-add: llvm llvm_wasm llvm_wasm_emcc llvm_nopragma
-RUN(NAME matmul_01 LABELS gfortran)
+RUN(NAME matmul_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvm_nopragma)
 RUN(NAME matmul_02 LABELS gfortran)
 RUN(NAME simd_01 LABELS gfortran c llvm llvm_nopragma c_nopragma)
 RUN(NAME simd_02 LABELS gfortran c llvm llvm_nopragma c_nopragma)
diff --git a/src/libasr/pass/intrinsic_array_function_registry.h b/src/libasr/pass/intrinsic_array_function_registry.h
index f7b8175be..6dc092110 100644
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -4028,6 +4028,21 @@ namespace MatMul {
             assert_msg += "`matrix_a(i, k)` and `matrix_b(k, j)`";
         }
         if (is_allocatable(result)) {
+            Vec<ASR::expr_t*> allocated_args; allocated_args.reserve(al, 1);
+            allocated_args.push_back(al, result);
+            ASR::expr_t* is_result_allocated = ASRUtils::EXPR(ASR::make_IntrinsicImpureFunction_t(al, loc,
+                static_cast<int64_t>(ASRUtils::IntrinsicImpureFunctions::Allocated),
+                allocated_args.p, allocated_args.n, 0, logical, nullptr));
+
+            Vec<ASR::expr_t*> deallocate_args; deallocate_args.reserve(al, 1);
+            deallocate_args.push_back(al, result);
+            ASR::stmt_t* explicit_deallocate = ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(
+                al, loc, deallocate_args.p, deallocate_args.n));
+
+            std::vector<ASR::stmt_t*> if_body;
+            if_body.push_back(explicit_deallocate);
+            body.push_back(al, b.If(is_result_allocated, if_body, {}));
+
             body.push_back(al, b.Allocate(result, alloc_dims));
         }
         body.push_back(al, STMT(ASR::make_Assert_t(al, loc, dim_mismatch_check,
```

certik:
I think that was a real bug.

krystophny:
Fix macOS abort in allocated_04_STD_F23.

Root cause: when the allocatable array descriptor pointer is NULL (common for allocatable members in derived types), codegen created a fresh descriptor but did not initialize it. Under --std=f23 this can take the realloc-based allocation path, so the uninitialized data pointer field was passed to _lfortran_realloc and aborted on macOS.

Fix: zero-initialize the fresh descriptor before filling the dimension descriptor; also initialize the array-string descriptor for character arrays. Commit: cb4adb12a.

krystophny:
Pushed b7b5332da on `fix-double-allocate-error`:

- Fix LLVM codegen crash for `character(len=*), allocatable :: strings(:)` dummies (hit by M_CLI2 + `--realloc-lhs-arrays`) by handling `ASR::AssumedLength` in `setup_string_length()`.
- Limit array-descriptor zero-init to the realloc-based path only (avoids `allocate_03` show-llvm/reference churn).
- Add integration reproducer `integration_tests/allocate_30.f90` (runs with `--realloc-lhs-arrays`).

Local targeted checks:
- `./run_tests.py -t ../integration_tests/allocate_03.f90 -b llvm -s`
- `lfortran --realloc-lhs-arrays integration_tests/allocate_30.f90 && ./allocate_30`}

certik:
@krystophny thanks for this. The other PRs will soon be merged, then you can rebase this / resolve conflicts and let's get it in.

certik:
All prerequisites are in, so this can now be rebased and conflicts fixed.

integration_tests/CMakeLists.txt
```diff
@@ -1742,6 +1742,7 @@ RUN(NAME allocate_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME allocate_26 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME allocate_27 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME allocate_29 FAIL LABELS gfortran llvm)
```
certik: It fails in the FAST mode ---- does the NOFAST not work?

certik: I think these bounds checks are better tested in tests/ not reference tests. Go ahead and move this tests to `tests/errors/`, and add a test like:
```
[[test]]
filename = "errors/scalar_allocation_check_01.f90"
run = true
```
Which will also add a reference test, so that we can see the actual error.

So that will be a much better solution anyway.

integration_tests/allocate_29.f90
```diff
@@ -0,0 +1,13 @@
...
+    integer, allocatable :: x
+    integer, allocatable :: arr(:)
+
+    allocate(x)
+    allocate(x)
+    print *, "Should not reach here - scalar"
+
+    allocate(arr(10))
+    allocate(arr(20))
+    print *, "Should not reach here - array"
```
certik: These re not tested, since it will fail for the first one --- you need to create two separate tests.

ci/test_third_party_codes.sh
```diff
@@ -227,12 +227,14 @@ time_section "🧪 Testing POT3D with fortran_mpi" '
 # Section 1: stdlib (Less Workarounds)
 ##########################
 time_section "🧪 Testing stdlib (Less Workarounds)" '
-  git clone https://github.com/Pranavchiku/stdlib-fortran-lang.git
+  # Temporarily use krystophny fork with double-allocation fix
+  # until https://github.com/Pranavchiku/stdlib-fortran-lang/pull/1 is merged
```
certik: This is a bug in LFortran: https://github.com/Pranavchiku/stdlib-fortran-lang/pull/1#issuecomment-3660905003

We need to deallocate all `intent(out)` arguments before entering the subroutine.

You can keep this temporarily to get the CI passing in this PR first, then we'll decide if we need to fix this in this PR, or the next PR. There might be more cases like this one, in which case we should just fix it here.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1626,6 +1645,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            ASR::down_cast<ASR::Var_t>(tmp_expr)->m_v);
+                        var_name = ASRUtils::symbol_name(sym);
+                    }
+                    llvm_utils->generate_runtime_error(is_allocated,
+                        "Runtime Error: Attempting to allocate already allocated variable '%s'\n",
+                        infile,
+                        x.base.base.loc,
+                        location_manager,
+                        LCompilers::create_global_string_ptr(context, *module, *builder, var_name));
+                }
```
certik, line 1697: I see two of these errors reporting in this PR --- is one for scalars, the other for arrays? After fixing https://github.com/lfortran/lfortran/pull/9039/changes#r2623572392, both should be then tested.

certik:
Looks good so far. Ping me once this is ready for another review.

ci/test_third_party_codes.sh
certik: In fact it might just be a bug in this PR, not in main, see https://github.com/lfortran/lfortran/issues/9097#issuecomment-3661620094.

certik:
Current failure:
```
2025-12-16T17:24:08.7391333Z At 152:4 of file build/dependencies/toml-f/src/tomlf/structure/ordered_map.f90
2025-12-16T17:24:08.7392371Z Runtime Error: Attempting to allocate already allocated variable 'list'
```


I would examine how you check for allocation. Do it like `allocate` does. Then it should work.

krystophny:
> Looks good so far. Ping me once this is ready for another review.

Unfortunately we need to fix #9097 here because we have no diagnostic otherwise, and also fpm test fails due to this now. I try to do it in single commit to review.

certik:
> Unfortunately we need to fix #9097 here because we have no diagnostic otherwise, and also fpm test fails due to this now. I try to do it in single commit to review.

Ok --- just a heads up https://github.com/lfortran/lfortran/issues/9097#issuecomment-3661620094 that I think there is nothing to fix there in main, it's not a bug in main, but rather in this PR.

krystophny:
> > Unfortunately we need to fix #9097 here because we have no diagnostic otherwise, and also fpm test fails due to this now. I try to do it in single commit to review.
> 
> Ok --- just a heads up [#9097 (comment)](https://github.com/lfortran/lfortran/issues/9097#issuecomment-3661620094) that I think there is nothing to fix there in main, it's not a bug in main, but rather in this PR.

All double checked and issue is updated. It is complex as usual. In any case all should be fine after this PR and all tests pass now.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1357,12 +1357,42 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                           ASR::is_a<ASR::Real_t>(*curr_arg_m_a_type) ||
                           ASR::is_a<ASR::Complex_t>(*curr_arg_m_a_type) ||
                           ASR::is_a<ASR::Logical_t>(*curr_arg_m_a_type)) {
+                    llvm::Type* llvm_arg_type = llvm_utils->get_type_from_ttype_t_util(curr_arg.m_a, curr_arg_m_a_type, module.get());
+                    // Skip double allocation check for intent(out) arguments since Fortran
+                    // standard requires automatic deallocation at procedure entry (not yet implemented)
```
certik: This comment says `intent(out)` deallocation is not yet implemented and it is skipping a double allocation check.

The comment https://github.com/lfortran/lfortran/issues/9097#issuecomment-3664846450 says it *is* implemened in this PR and everything works. 

Finally, the `insert_deallocate.cpp` does have code to insert this deallocation, but doesn't say if it works in all cases or not.

Can you please clarify exactly what the current status is and what needs to be done?

src/libasr/pass/array_op.cpp
```diff
@@ -108,8 +108,37 @@ class ArrayVarAddressCollector: public ASR::CallReplacerOnExpressionsVisitor<Arr
...
+            ASR::ttype_t* char1_type = ASRUtils::TYPE(ASR::make_String_t(al, loc, 1,
+                ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, loc, 1, int32_type)),
+                ASR::string_length_kindType::ExpressionLength,
+                ASR::string_physical_typeType::DescriptorString));
+            // Convert source to StringItem(source, loop_index)
+            xx.m_source = ASRUtils::EXPR(ASR::make_StringItem_t(
+                al, loc, xx.m_source, loop_index, char1_type, nullptr));
+            // Fix result type from array to element type
+            xx.m_type = ASRUtils::extract_type(xx.m_type);
+        }
```
certik: Why is this here? It seems unrelated to this PR and it is not tested?

src/libasr/pass/array_op.cpp
```diff
@@ -1047,7 +1080,16 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+            ASR::ttype_t* source_type = ASRUtils::expr_type(bit_cast->m_source);
+            ASR::ttype_t* result_element_type = ASRUtils::extract_type(bit_cast->m_type);
+            // Looping IS necessary for string→integer_array (transfer(string, int_array))
+            // because we need element-wise conversion from characters to integers
+            // For string→character_array, looping is NOT needed (byte-wise copy)
+            if (ASRUtils::is_string_only(source_type) &&
+                ASRUtils::is_integer(*result_element_type)) {
+                return true;
+            }
+            return !ASRUtils::is_array(source_type);
```
certik: It seems all changes in this file are not related to this PR?

src/libasr/pass/insert_deallocate.cpp
```diff
@@ -285,12 +286,135 @@ class LoopTempVarDeallocateVisitor : public ASR::BaseWalkVisitor<LoopTempVarDeal
         }
 };
 
+// Deallocate intent(out) allocatable arguments at function entry
```
certik: Document here the limitations of this implementation.

Given that main implements it in some cases, but not all, does this implementation implement it in even more cases?

Given that we skip double deallocation check for intent(out), could it happen that this implementation adds deallocate, but there is also another deallocate in `main`, so it's duplicated now?

certik:
Thanks for getting it passing!

I left several comments. Based on the answers we can decide what to do.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -1357,12 +1357,48 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                           ASR::is_a<ASR::Logical_t>(*curr_arg_m_a_type)) {
+                    llvm::Type* llvm_arg_type = llvm_utils->get_type_from_ttype_t_util(curr_arg.m_a, curr_arg_m_a_type, module.get());
+                    // Skip double allocation check for INTENT(OUT) allocatables.
+                    //
+                    // Fortran requires allocatable INTENT(OUT) dummy arguments to be automatically
+                    // deallocated on procedure entry (F2018 15.5.2.13). LFortran currently inserts
+                    // this deallocation only in limited cases (see the `insert_deallocate` pass),
+                    // and it is not complete yet (#9097). Until INTENT(OUT) deallocation is fully
+                    // implemented, keeping the double-allocation runtime check enabled here would
+                    // report false positives in valid programs.
```
certik, line 1368: Excellent, I think this is clear now.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: Clarified in-code. `insert_deallocate` inserts entry deallocation only in limited cases (allocatable `intent(out)` dummies in `m_args`, optional guarded by `present()`), and full deallocation/finalization coverage is still incomplete (#9097). Because of that, the `intent(out)` double-allocation runtime check remains skipped for now to avoid false positives; next step is completing `intent(out)` deallocation/finalization semantics so the check can be re-enabled.

src/libasr/pass/array_op.cpp
krystophny: Removed

src/libasr/pass/array_op.cpp
krystophny: Removed

src/libasr/pass/insert_deallocate.cpp
krystophny: Added a limitations block comment in `insert_deallocate.cpp` describing scope (`allocatable intent(out)` dummies, optional guarded by `present()`, deallocate guarded by `allocated()`, skip intrinsic implementations) and pointing to remaining gaps (#9097). The inserted `deallocate()` is guarded by `allocated()`, so even if another deallocation exists, it should not double-free; the longer-term fix is to complete `intent(out)` semantics so we can re-enable the runtime double-allocation check.

krystophny:
> I think that this is fine. It's an improvement and we'll iterate on this further in subsequent PRs.

Great! I started #9108 for now - we can enable stricter checks step by step the more of #9097 we resolve. These issues are really like a hydra, you fix one, three more pop up.

PR #9040, opened 2025-12-12
Fix debug Vec indexing + empty WRITE(*,*) handling

krystophny:
Merge order: #9048 -> #9040

## Summary
- Add a debug-only bounds assert for `Vec::operator[]`.
- Fix a concrete out-of-bounds trigger when compiling `WRITE(*,*)` with an empty item list.

## Notes
This is debug-path hardening; it should not change normal codegen behavior.


certik:
Let's wait for some exhaustive tests to pass.

Issue #9021, opened 2025-12-12
Error : unreported error (allocatable + initial value)


assem2002:
```.f90
program string_example
    implicit none
    character(:), allocatable :: greeting = "H"
    print *, greeting(1)

end program string_example
```
```.f90
program string_example
  implicit none
  integer, allocatable :: greeting = 1
end program string_example
```
refer to this comment for further explanation https://github.com/lfortran/lfortran/issues/7414#issuecomment-3644966840

xaerru:
Fixed in https://github.com/lfortran/lfortran/pull/9047.

PR #9047, opened 2025-12-13
Reject initialization of ALLOCATABLE variables

opixdown:
## **Issue Fixed**
#9021  Error : unreported error (allocatable + initial value)

## **Summary**

This PR reports a missing semantic error when an ALLOCATABLE variable is declared with an initialization expression. Such declarations are invalid in Fortran and must be rejected during semantic analysis.

## **Reference Code**
```
program string_example
    implicit none
    character(:), allocatable :: greeting = "H"
    print *, greeting(1)
end program string_example
```
```
program string_example
  implicit none
  integer, allocatable :: greeting = 1
end program string_example
```

```
  store i32 1, i32** @test_bug.x, align 4
 i32*
```

```
semantic error: An ALLOCATABLE variable cannot have an initialization expression
 --> ./t.f90:3:34
  |
3 |     character(:), allocatable :: greeting = "H"
  |                                  ^^^^^^^^^^^^^^ 

```
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran ./t1.f90
semantic error: An ALLOCATABLE variable cannot have an initialization expression
 --> ./t1.f90:3:27
  |
3 |   integer, allocatable :: greeting = 1
  |                           ^^^^^^^^^^^^ 
  
```

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -3761,14 +3761,6 @@ public:
                     }
                 }
 
-                if (is_allocatable && s.m_initializer != nullptr) {
-                    diag.add(Diagnostic(
-                        "An ALLOCATABLE variable cannot have an initialization expression",
-                        Level::Error, Stage::Semantic, {
-                            Label("", {s.loc})
-                        }));
-                    throw SemanticAbort();
-                }
                 if (is_char_type && sym_type->n_kind == ASR::string_length_kindType::DeferredLength &&  is_allocatable) {
                     bool has_fixed_shape = false;
                     for (size_t d = 0; d < s.n_dim; d++) {
@@ -3999,7 +3991,16 @@ public:
                         )));
                         throw SemanticAbort();
                     }
-
+    
+                    if (is_allocatable && s.m_initializer != nullptr) {
+                        diag.add(Diagnostic(
+                            "An 'allocatable' variable cannot have an initialization expression",
+                            Level::Error, Stage::Semantic, {
+                                Label("", {s.loc})
+                            }));
+                        throw SemanticAbort();
+                    }
+                    
                     if (s_intent == ASRUtils::intent_out && value_attr) {
                         diag.add(Diagnostic(
                             "`value` attribute conflicts with `intent(out)` attribute",
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -209,6 +209,8 @@ program continue_compilation_1
 
 
 
+
+
     ! Use the space above to insert new declarations, and remove the line, so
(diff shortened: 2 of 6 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -173,10 +173,8 @@ program continue_compilation_1
     integer :: j2, i2, k2(2), x2(2), y2(3)    
     integer::tt = b'01' * 3
     integer :: fmt_i1, fmt_i2, fmt_i3 ! for issue #8925
-
-
-
-
+    integer, allocatable :: allocate_int = 1
+    character(:), allocatable :: allocate_char = "H"
```
certik, line 177: You need to remove exactly the number of lines that you added, to keep the diff below minimal. So remove only 2, not 4.

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -147,11 +147,11 @@ semantic error: `parameter` attribute conflicts with `pointer` attribute
...
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 
-semantic error: `parameter` attribute conflicts with `allocatable` attribute
-   --> tests/errors/continue_compilation_2.f90:101:5
+semantic error: An ALLOCATABLE variable cannot have an initialization expression
+   --> tests/errors/continue_compilation_2.f90:101:40
     |
 101 |     integer, allocatable, parameter :: v=1
-    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+    |                                        ^^^
```
certik: We have to be careful here. Now we lost the test for the "`parameter` attribute conflicts with `allocatable` attribute." error.

So either we need to figure out some other test for it, or if that condition is now not testable, then maybe we can remove that error message. Or we can rearrange the code to test for the conflict first (that might be the best).

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3761,6 +3761,14 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     }
                 }
 
+                if (is_allocatable && s.m_initializer != nullptr) {
+                    diag.add(Diagnostic(
+                        "An ALLOCATABLE variable cannot have an initialization expression",
```
certik: We'll have to go over all error messages eventually and make them consistent. But I would use lowercase generally, for modern Fortran. You can use backticks to "highlight", like this:

```suggestion
                        "An `allocatable` variable cannot have an initialization expression",
```

certik:
@opixdown thanks for the PR! I left a few comments, you also need to rebase I think.

Issue #9028, opened 2025-12-12
Two allocations in a row should give runtime error in Debug mode


certik:
```fortran
program test_alloc
integer, allocatable :: x
allocate(x)
allocate(x)
print *, "Passed, but should not."
end program
```
This gives:
```console
$ gfortran a.f90 && ./a.out
At line 4 of file a.f90
Fortran runtime error: Attempting to allocate already allocated variable 'x'

Error termination. Backtrace:
#0  0x100e63cf7
#1  0x100e64b07
#2  0x100e64ebf
#3  0x10087bd43
#4  0x10087be03
$ lfortran a.f90
Passed, but should not.
```

I would say definitely in Debug mode, but possibly in Release mode too, like GFortran does it.

xaerru:
This was fixed in https://github.com/lfortran/lfortran/pull/9039.

```fortran
program test_alloc
integer, allocatable :: x
allocate(x)
allocate(x)
print *, "Passed, but should not."
end program
```
This gives:
```shell
At 4:1 of file a.f90
Runtime Error: Attempting to allocate already allocated variable 'x'
```

PR #9050, opened 2025-12-13
Fix MATMUL: handle allocatable result reallocation

krystophny:
Split out of PR #9039.

Changes:
- src/libasr/pass/intrinsic_array_function_registry.h: handle reallocation semantics for allocatable result in MATMUL.

Merge order: merge this before PR #9039.

### DIFF
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -4030,6 +4030,21 @@ namespace MatMul {
             assert_msg += "`matrix_a(i, k)` and `matrix_b(k, j)`";
         }
         if (is_allocatable(result)) {
+            Vec<ASR::expr_t*> allocated_args; allocated_args.reserve(al, 1);
+            allocated_args.push_back(al, result);
+            ASR::expr_t* is_result_allocated = ASRUtils::EXPR(ASR::make_IntrinsicImpureFunction_t(al, loc,
+                static_cast<int64_t>(ASRUtils::IntrinsicImpureFunctions::Allocated),
+                allocated_args.p, allocated_args.n, 0, logical, nullptr));
+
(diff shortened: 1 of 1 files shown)


certik:
I can't figure out a test case that fails in main and works now.

Also, I noticed in main we already automatically allocate the LHS, but it should only happen with `--realloc-lhs-arrays`: https://github.com/lfortran/lfortran/issues/9086.

certik:
@krystophny is the issue that due to https://github.com/lfortran/lfortran/issues/9028, we are unable to find a test case in main, but once we fix https://github.com/lfortran/lfortran/issues/9028, then the test case that this PR fixes is that you allocate the LHS, and then assign matmul to it with `--realloc-lhs-arrays`?

Oh -- is https://github.com/lfortran/lfortran/pull/9039 fixing exactly that problem?

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -4030,6 +4030,21 @@ namespace MatMul {
...
+                allocated_args.p, allocated_args.n, 0, logical, nullptr));
+
+            Vec<ASR::expr_t*> deallocate_args; deallocate_args.reserve(al, 1);
+            deallocate_args.push_back(al, result);
+            ASR::stmt_t* explicit_deallocate = ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(
+                al, loc, deallocate_args.p, deallocate_args.n));
+
+            std::vector<ASR::stmt_t*> if_body;
+            if_body.push_back(explicit_deallocate);
+            body.push_back(al, b.If(is_result_allocated, if_body, {}));
```
certik, line 4046: Shouldn't this only happen with `--realloc-lhs-arrays`, otherwise give a bonds check error if it is allocated to the wrong size?

certik:
Let's wait for a few exhaustive tests to pass first.

PR #9073, opened 2025-12-14
Fix assumed-size array error handling and code style

### LINKED ISSUE #9068: Add test for assumed-size dimension position error
## Follow-up from #9024

The error "Assumed-size '*' is only permitted in the last dimension" only triggers when `intent` is explicitly specified, but according to the Fortran standard it should apply to ALL dummy arguments.

### Standard Reference (Fortran 2023)

- **R825**: `assumed-size-spec is explicit-shape-spec-list, assumed-implied-spec` - the `*` is syntactically only allowed in the last position
- **C835**: "An object whose array bounds are specified by an assumed-size-spec shall be a dummy data object" - must be a dummy argument, no intent requirement

A "dummy data object" is simply a dummy argument that is a data object - intent is not required.

### Current Bug

In `ast_common_visitor.h`, the `is_argument` parameter passed to `process_dims` only considers explicit intents:

```cpp
process_dims(al, dims, s.m_dim, s.n_dim, is_compile_time, is_char_type,
    (s_intent == ASRUtils::intent_in || s_intent == ASRUtils::intent_out ||
    s_intent == ASRUtils::intent_inout), s.m_name);
```

This misses dummy arguments without explicit intent (which have `intent_unspecified`).

### Test Cases

Should error (currently only first one errors):
```fortran
subroutine foo(a)
  real, intent(in) :: a(*, 10)  ! Error triggered
end subroutine

subroutine bar(a)
  real :: a(*, 10)  ! BUG: No error, but should error
end subroutine
```

### Fix

Change the check to include `intent_unspecified` for variables in the procedure argument list:

```cpp
bool is_dummy_arg = (s_intent == ASRUtils::intent_in || 
                     s_intent == ASRUtils::intent_out ||
                     s_intent == ASRUtils::intent_inout ||
                     s_intent == ASRUtils::intent_unspecified);
process_dims(al, dims, s.m_dim, s.n_dim, is_compile_time, is_char_type,
    is_dummy_arg, s.m_name);
```

Or check if the variable name is in `current_procedure_args`.

### Deliverables

1. Fix the `is_argument` check in `process_dims` call
2. Add test cases to `tests/errors/continue_compilation_1.f90` for both with and without explicit intent

### LINKED ISSUE #9069: Code style: else if should be on same line in ast_common_visitor.h
## Follow-up from #9024

Two places in `src/lfortran/semantics/ast_common_visitor.h` have `} else` followed by `if` on the next line. These should be written as `} else if` on the same line for consistency.

### Location 1 (around line 4213):
```cpp
if (is_argument && is_dimension_star) {
    symbol_variable->m_type = type;
} else
if ( ASR::is_a<ASR::Array_t>(*symbol_variable->m_type) ) {
```

Should be:
```cpp
if (is_argument && is_dimension_star) {
    symbol_variable->m_type = type;
} else if ( ASR::is_a<ASR::Array_t>(*symbol_variable->m_type) ) {
```

### Location 2 (around line 4133):
Same pattern needs fixing.

### LINKED ISSUE #9070: Clarify comment about assumed-size dimension handling in asr_to_llvm.cpp
## Follow-up from #9024

In `src/libasr/codegen/asr_to_llvm.cpp`, there's an unclear comment about assumed-size dimensions:

```cpp
} else {
    // Assumed-size dimension - ubound is undefined, avoid crash and use lbound.
    ubs.p[i] = lbound;
}
```

The comment should be clarified:
- What kind of crash are we avoiding?
- Is using lbound just a temporary workaround, or is that the correct solution?
- If it's the correct solution, the comment shouldn't mention ubound since that was incorrect anyway.

Please update the comment to explain the actual semantics.

### LINKED ISSUE #9071: Inline function pass incorrectly skips assumed-shape arrays
## Follow-up from #9024

In `src/libasr/pass/inline_function_calls.cpp`, the current code stops inlining for both assumed-shape and assumed-size arrays:

```cpp
// Don't inline functions with assumed-size array parameters
// (PointerArray with empty dimensions)
ASR::ttype_t* var_type = ASR::down_cast<ASR::Variable_t>(sym.second)->m_type;
if( ASRUtils::is_array(var_type) &&
    ASRUtils::is_dimension_empty(var_type) ) {
    return false;
}
```

This is too broad. We should only stop inlining for **assumed-size** arrays (e.g., `A(*)`), not assumed-shape arrays (e.g., `A(:)`).

The check needs to be refined to distinguish between these two cases.

### LINKED ISSUE #9072: Assumed-size arrays should use UnboundedPointerArray physical type
## Follow-up from #9024

In `src/libasr/asr_utils.h`, the physical type for assumed-size arrays `A(*)` should be `UnboundedPointerArray`, not `PointerArray`.

### The Problem

**Before PR #9024:**
```cpp
} else if ( is_dimension_star && ASRUtils::is_only_upper_bound_empty(m_dims[n_dims-1]) ) {
    physical_type = ASR::array_physical_typeType::UnboundedPointerArray;
```

**After PR #9024:**
```cpp
} else if (is_dimension_star && is_argument) {
    physical_type = ASR::array_physical_typeType::PointerArray;  // <-- Wrong!
```

### Why This Matters

1. **`PointerArray`** is meant for arrays with **known/specified dimensions** (like `A(10,20)` passed by reference)
2. **`UnboundedPointerArray`** is meant for **assumed-size** arrays (`A(*)`) where the last dimension is truly unknown
3. The PR changed assumed-size to use `PointerArray` with **empty dimensions** (`m_start=nullptr, m_length=nullptr`)
4. This creates an inconsistent state: `PointerArray` normally expects dimensions to be specified, but we're leaving them empty

### Cascading Issues

This causes problems in other places:
- The inline pass (#9071) checks `is_dimension_empty()` which catches both assumed-shape AND assumed-size arrays
- Other code may expect `PointerArray` to always have valid dimensions

### Fix

Change back to `UnboundedPointerArray`:

```cpp
} else if (is_dimension_star && is_argument) {
    physical_type = ASR::array_physical_typeType::UnboundedPointerArray;
```

Then #9071 (inline pass) can check for `UnboundedPointerArray` physical type instead of checking for empty dimensions, which properly distinguishes assumed-size from assumed-shape.

The overall approach in #9024 (handling star dimensions, sequence association, etc.) is correct - we just picked the wrong physical type.

krystophny:
## Summary

Followup fixes from PR #9024 review:

- Fix #9069: `else if` on same line in ast_common_visitor.h
- Fix #9070: Add semantic errors for ubound on assumed-size arrays (last dimension undefined)
- Fix #9072: Use `UnboundedPointerArray` physical type for assumed-size arrays
- Fix #9071: Update inline pass to check `UnboundedPointerArray` instead of empty dimensions
- Fix #9068: Assumed-size `*` position check now works without explicit intent

## Changes

**ast_common_visitor.h:**
- Fixed code style: `} else if` on same line (2 locations)
- Added semantic error when `ubound(a, dim)` queries the last dimension of assumed-size array
- Added semantic error when `ubound(a)` is called without DIM on assumed-size array
- Fixed detection to use `UnboundedPointerArray` physical type instead of null `m_length`
- Fixed `process_dims` call to include `|| is_argument` so assumed-size star check works for dummy args without explicit intent

**asr_utils.h:**
- Changed assumed-size arrays to use `UnboundedPointerArray` physical type instead of `PointerArray`

**inline_function_calls.cpp:**
- Updated check to use `UnboundedPointerArray` physical type instead of `is_dimension_empty()`
- This correctly excludes only assumed-size arrays from inlining, not assumed-shape arrays

**asr_to_llvm.cpp:**
- Cleaned up comments (removed TODO notes, simplified)

**Tests:**
- Added 5 test cases to `continue_compilation_1.f90` covering all error scenarios

## Rationale

PR #9024 made semantics able to represent legacy array-section rewriting and LAPACK-style assumed-size/sequence-association cases, but LLVM codegen still had a few missing `UnboundedPointerArray`/assumed-size paths (assert/ICE or incorrect shape handling).

This PR finishes that work by making LLVM lowering handle:
- Descriptor array to unbounded pointer physical casts
- Array section association for `UnboundedPointerArray` via the data-only descriptor path
- Assumed-size/sequence association cases without triggering `ArrayBound` evaluation on unbounded pointers

These checks are gated on the relevant physical types and do not broaden `--legacy-array-sections` behavior beyond the cases where the ASR already uses `UnboundedPointerArray`.

## Verification

- `micromamba run -p .micromamba/envs/lf bash -lc 'unset LD_LIBRARY_PATH; cd integration_tests && ./run_tests.py -b llvm -j16'`
- `micromamba run -p .micromamba/envs/lf bash -lc 'unset LD_LIBRARY_PATH; ./run_tests.py -u'`

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3132,7 +3132,21 @@ public:
                     } else {
                         dim_start = llvm::ConstantInt::get(context, llvm::APInt(32, 1));
                     }
+                    llvm::Value* dim_size = nullptr;
+                    if (m_dims[idim].m_length) {
+                        ptr_loads = 2 - !LLVM::is_llvm_pointer(*ASRUtils::expr_type(m_dims[idim].m_length));
+                        if (is_intent_in) {
+                            this->visit_expr_wrapper(m_dims[idim].m_length, true);
+                        } else {
+                            load_array_size_deep_copy(m_dims[idim].m_length);
+                        }
+                        dim_size = tmp;
+                    } else {
+                        // Last dimension of assumed-size array has no length
+                        dim_size = llvm::ConstantInt::get(context, llvm::APInt(32, 0));
+                    }
                     llvm_diminfo.push_back(al, dim_start);
+                    llvm_diminfo.push_back(al, dim_size);
                 }
                 ptr_loads = ptr_loads_copy;
             }
@@ -3145,7 +3159,7 @@ public:
                                                     true,
                                                     false,
                                                     llvm_diminfo.p, is_polymorphic, current_select_type_block_type,
-                                                    true, false, array_name, infile);
+                                                    false, false, array_name, infile);
             } else {
                 llvm::Type* type;
                 bool is_fixed_size = (array_t->m_physical_type == ASR::array_physical_typeType::FixedSizeArray ||


tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -20,6 +20,38 @@ semantic error: Dummy argument 'c' not defined
...
+
+semantic error: Dummy argument 'a' not defined
+  --> tests/errors/continue_compilation_1.f90:92:5 - 94:18
+   |
+92 |        subroutine assumed_size_star_pos_1(a)
+   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
+...
+   |
+94 |        end subroutine
+   | ...^^^^^^^^^^^^^^^^^^
```
certik, line 37: Later: it should highlight just `a`, not the whole subroutine.

Also the argumetn `a` *is* defined, just incorrectly defined. So the error message is misleading --- I would not print it at all, since the user defined `a`, but made a mistake in the definition, which we reported already (below).

tests/errors/continue_compilation_1.f90
```diff
@@ -74,28 +74,28 @@ subroutine s(c) bind(c)
...
+    end subroutine
 
+    subroutine ubound_assumed_size_3(a)
+        real :: a(10, *)
+        print *, ubound(a, 2)
+    end subroutine
 
+    subroutine assumed_size_star_pos_1(a)
+        real, intent(in) :: a(*, 10)
+    end subroutine
```
certik, line 94: Maybe these can be merged into just one subroutine and multiple arguments, to make the tests more concise.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7590,6 +7588,20 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            if (bound == ASR::arrayboundType::UBound &&
+                ASR::is_a<ASR::Array_t>(*v_Var_type) &&
+                ASR::down_cast<ASR::Array_t>(v_Var_type)->m_physical_type ==
+                    ASR::array_physical_typeType::UnboundedPointerArray) {
+                diag.add(Diagnostic(
+                    "The DIM argument must be present when calling UBOUND "
+                    "on an assumed-size array",
+                    Level::Error, Stage::Semantic, {
+                        Label("", {x.base.base.loc})}));
+                throw SemanticAbort();
```
certik, line 7623: Can you show an example Fortran code which allows to call ubound on assumed-size array at all, even with the `dim` argument? Is the idea that the `dim` argument allows you to select a dimension that is not assumed-size? If so, then the error message is ok, and let's also add a test that passes for ubound with assumed-size.

certik:
The current failure is:
```
FAILED: SRC/CMakeFiles/lapack.dir/sgbsvx.f.o 
/home/runner/work/lfortran/lfortran/src/bin/lfortran -I/tmp/tmp.zhiDFOAEeT/assert/modern_minpack_01/modern_minpack_02/dftatom/fastGPT/stdlib/SNAP/src/lapack/build/lapack-vanilla/SRC --fixed-form-infer --implicit-interface --legacy-array-sections --separate-compilation -c SRC/CMakeFiles/lapack.dir/sgbsvx.f-pp.f -o SRC/CMakeFiles/lapack.dir/sgbsvx.f.o
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/codegen/asr_to_llvm.cpp
function visit_ArrayPhysicalCastUtil(), line number 8158 at 
false
Aborted (core dumped)
```

Let's create an MRE of this, and add to integration tests, since clearly we do not exercise our code correctly using integration tests only.

krystophny:
@certik tests pass now, follow-up issues are #9105, #9106, #9107

Issue #8182, opened 2025-07-30
Handle assignment of string from string array using transfer


jinangshah21:
MRE:
```fortran
program test_transfer
  character(kind=1), dimension(100) :: lhs
  character(len=*), parameter :: rhs = "hello world"
  integer :: length
  length = len_trim(rhs)
  lhs(1:length) = transfer(rhs, lhs)
  print *, lhs(1:length)
end program test_transfer
```

LFortran Error:
```console
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store %string_descriptor* %111, %string_descriptor* %arr_element, align 8
 %string_descriptor = type <{ i8*, i64 }>
```

assem2002:
that's due to the return of  `bitcast` after applying the passes.
```clj
 (Assignment
(ArrayItem
    (Var 2 __libasr_created__array_section_)
    [(()
    (Var 2 __libasr_index_0_)
    ())]
    (String 1 (IntegerConstant 1 (Integer 4) Decimal) ExpressionLength DescriptorString)
    ColMajor
    ()
)
(BitCast
    (Var 2 rhs)
    (ArrayItem
        (Var 2 lhs)
        [(()
        (Var 2 __libasr_index_0_1)
        ())]
        (String 1 (IntegerConstant 1 (Integer 4) Decimal) ExpressionLength DescriptorString)
        ColMajor
        ()
    )
    ()
    (Array
        (String 1 (IntegerConstant 1 (Integer 4) Decimal) ExpressionLength DescriptorString)
        [((IntegerConstant 1 (Integer 4) Decimal)
        (IntegerConstant 64 (Integer 4) Decimal))]
        PointerToDataArray
    )
    ()
)
()
.false.
)
```
`strcpy` gets invoked only on `RHS` of string type (not an array of strings).

PR #9078, opened 2025-12-14
fix: add character array assignment logic for transfer functions

### LINKED ISSUE #9076: bug: sematic error in transfer assignment from character to array due to incorrect mold calculation
MRE:
```
program test_transfer
  implicit none
  character(len=*), parameter :: rhs = "xyzw"
  character(kind=1), dimension(4) :: temp
  temp = transfer(rhs, temp)
  print *, temp
  print *, rhs
end program test_transfer
```

Main:
```
$ lfortran a.f90 
semantic error: Different shape for array assignment on dimension 1(4 and 1)
  --> a.f90:21:3
   |
21 |   temp = transfer(rhs, temp)
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

Gfortran, ifx, flang:
```
$ gfortran a.f90 & ./a.out
[1] 94821
 xyzw
 xyzw
```

YashNagda17:
Fixes #9076
Towards #8182

These changes create an initial fix for assignment of bitcast outputs to character arrays.

Changes Made:
* Adding an helper function in ASR to LLVM Conversion to assign bitcast return value to character arrays of arbitrary length and kind
* Assigning correct mold values to strings & character-array cases, in AST to ASR Conversions.


YashNagda17:
This PR solves #8182 , so that we can go ahead fixing other issues for FPM where this particular issue might be a blocker. Still, there are few cases where this particular operation might fail. Once the Approach is finalized, we can go about fixing them in subsequent PRs.

Issue #9062, opened 2025-12-13
Handle pushing back a new element in struct array member using `ArrayConstructor`


jinangshah21:
MRE:
```fortran
module m_mod
    implicit none

    type :: node
        integer :: id
    end type node

    type :: target_t
        type(node), allocatable :: dependencies(:)
    end type target_t

end module m_mod


program test_append
    use m_mod
    implicit none

    type(target_t) :: target

    allocate(target%dependencies(2))
    target%dependencies = [target%dependencies, node(1)]

end program test_append
```

LFortran Error:
```console
lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran b.f90 --realloc-lhs-arrays
At 23:27 of file b.f90
Runtime error: Array '__libasr__created__var__0__array_constructor_' index out of bounds.

Tried to access index 3 of dimension 1, but valid range is 1 to 2.
```

Brijesh-Thakkar:
@jinangshah21 Hello, Can i work on this issue
If yes, please assign it to me

certik:
This works for me:
```
$ lfortran a.f90 --realloc-lhs-arrays
$
```

certik:
This was fixed by https://github.com/lfortran/lfortran/pull/9035, test it added.

PR #9079, opened 2025-12-14
fix for incorrect runtime allocation (issue #9062)

Atharv1405:
issue #9062 

MRE
```
module m_mod
    implicit none

    type :: node
        integer :: id
    end type node

    type :: target_t
        type(node), allocatable :: dependencies(:)
    end type target_t

end module m_mod


program test_append
    use m_mod
    implicit none

    type(target_t) :: target

    allocate(target%dependencies(2))
    target%dependencies = [target%dependencies, node(1)]

end program test_append
```

**problem**
```
if (ASR::is_a<ASR::ArrayConstructor_t>(*value) &&
               !PassUtils::is_args_contains_allocatable(value)) {
            ASR::ArrayConstructor_t* arr_constructor = ASR::down_cast<ASR::ArrayConstructor_t>(value);
            value_m_dims->m_length = ASRUtils::get_ArrayConstructor_size(al, arr_constructor);
        }
```
i think `is_args_contains_allocatable` check rejects the runtime allocation as it remains true for both compile time resolution and realloc constructor so parameters never updates for realloc . check seems unnecessary for runtime codegen 
by removing that condition MRE passes 
please review and provide feedback if needed

### DIFF
--- a/src/libasr/pass/array_passed_in_function_call.cpp
+++ b/src/libasr/pass/array_passed_in_function_call.cpp
@@ -235,11 +235,11 @@ public:
         }
         // dimensions can be different for an ArrayConstructor e.g. [1, a], where `a` is an
         // ArrayConstructor like [5, 2, 1]
-        if (ASR::is_a<ASR::ArrayConstructor_t>(*value) &&
-               !PassUtils::is_args_contains_allocatable(value)) {
+        if (ASR::is_a<ASR::ArrayConstructor_t>(*value)) {
             ASR::ArrayConstructor_t* arr_constructor = ASR::down_cast<ASR::ArrayConstructor_t>(value);
             value_m_dims->m_length = ASRUtils::get_ArrayConstructor_size(al, arr_constructor);
         }
+
 
         /*
             Handle character type with assumed length OR functioncall (to avoid duplicate calls).


certik:
Add this MRE into integration tests and enable it.

For me locally your PR does not fix it:
```
$ lfortran a.f90
At 22:5 of file a.f90
Runtime Error: Array shape mismatch in assignment

Tried to match size 2 of dimension 1 of LHS with size 3 of dimension 1 of RHS.
```

In fact one must run it with `lfortran a.f90 --realloc-lhs-arrays`, then it works already in main.

certik:
@Atharv1405 I think the test already works in the latest main, but please let me know if that is not the case.

Atharv1405:
in my local system the mre fails for both `./src/bin/lfortran -test.f90` and `./src/bin/lfortran -test.f90 --realloc-lhs-arrays` but after update `./src/bin/lfortran -test.f90 --realloc-lhs-arrays` does not give error . i will add it in integrations test and check it

HarshitaKalani:
Yeah, it works with the latest main
```console
% cat a.f90
module m_mod
    implicit none

    type :: node
        integer :: id
    end type node

    type :: target_t
        type(node), allocatable :: dependencies(:)
    end type target_t

end module m_mod


program test_append
    use m_mod
    implicit none

    type(target_t) :: target

    allocate(target%dependencies(2))
    target%dependencies = [target%dependencies, node(1)]
    print *, target%dependencies
    print *, size(target%dependencies)

end program test_append
% gfortran a.f90 && ./a.out          
           0           0           1
           3
% lfortran a.f90 --realloc-lhs-arrays
0    0    1
3
```

PR #9080, opened 2025-12-14
ci: pin package versions with build hash

kmr-srbh:
- Pin exact package versions in `environment_linux_llvm.yml` with build hash and order them alphabetically.
- Clean up `Exhaustive-Checks-CI.yml` to remove redundancy with package installations based on LLVM version.


kmr-srbh:
This change only affects the Exhaustive Checks that are run on Linux.

PR #9087, opened 2025-12-15
CI: enable more fpm test packages

jinangshah21:
Towards #6840


certik:
You have to run exhaustive tests for 3rd-party tests now.

PR #9089, opened 2025-12-15
Fix MATMUL: respect --realloc-lhs-arrays flag

### LINKED ISSUE #9086: Bounds check: matmul to unallocated LHS (without `--realloc-lhs-arrays`)
This program:
```fortran
program test_matmul_allocatable
  implicit none
  integer, parameter :: n = 3, m = 4, p = 2
  real, allocatable :: A(:,:), B(:,:), C(:,:)
  integer :: i, j

  ! Allocate matrices A (n x m) and B (m x p)
  allocate(A(n, m))
  allocate(B(m, p))

  ! Initialize A and B with some values
  call random_seed()
  call random_number(A)
  call random_number(B)

  print *, "Allocation before"
  print *, allocated(A)
  print *, allocated(B)
  print *, allocated(C)

  ! Multiply: C = A * B  (result is n x p)
  ! The result of matmul is assigned to an allocatable array
  C = matmul(A, B)

  print *, "Allocation after"
  print *, allocated(A)
  print *, allocated(B)
  print *, allocated(C)

  ! Note: If C was not yet allocated, the above line would automatically
  ! allocate it with the correct shape. If it was already allocated with
  ! a different shape, it would be deallocated and reallocated.

  ! Print the result
  print *, 'Matrix C = A * B:'
  do i = 1, size(C,1)
     print '(100f10.4)', (C(i,j), j=1, size(C,2))
  end do

end program test_matmul_allocatable
```
Currently behaves like this:
```console
$ lfortran b.f90
Allocation before
T
T
F
Allocation after
T
T
T
Matrix C = A * B:
    1.2486    0.9687
    0.8542    0.5363
    0.7294    0.5320
Allocation before
T
T
F
Allocation after
T
T
T
Matrix C = A * B:
    0.8574    0.7576
    1.1769    1.1479
    0.9019    0.7143
```

Instead, it must give a bounds check error without `--realloc-lhs-arrays`, since the LHS is not allocated.

After fixing it, also add a test that it is allocated, but to the wrong size/shape, which should also give an error without `--realloc-lhs-arrays`. Finally, ensure that with `--realloc-lhs-arrays` it works in either case.

krystophny:
Fixes #9086.

Follow-up to #9050: remove unconditional allocation from `instantiate_MatMul`. The allocation/reallocation of allocatable LHS should be handled by the assignment pass based on `--realloc-lhs-arrays`, not inside the intrinsic.

Behavior now matches regular array assignment and other intrinsics.

### DIFF
--- a/tests/reference/run-matmul_unallocated_01-39026d3.stderr
+++ b/tests/reference/run-matmul_unallocated_01-39026d3.stderr
@@ -1,0 +1,2 @@
+At 10:5 of file tests/errors/matmul_unallocated_01.f90
+Runtime Error: Array 'result' is indexed but not allocated.
--- a/tests/errors/matmul_unallocated_01.f90
+++ b/tests/errors/matmul_unallocated_01.f90
@@ -1,0 +1,13 @@
+program matmul_unallocated_01
+implicit none
+real, allocatable :: A(:,:), B(:,:), C(:,:)
+
+allocate(A(3, 4), B(4, 2))
+A = 1.0
+B = 1.0
+
+! C is not allocated - should fail without --realloc-lhs-arrays
+C = matmul(A, B)
+
+print *, C(1,1)
+end program
--- a/tests/reference/run-matmul_unallocated_01-39026d3.json
+++ b/tests/reference/run-matmul_unallocated_01-39026d3.json
@@ -1,0 +1,13 @@
+{
+    "basename": "run-matmul_unallocated_01-39026d3",
+    "cmd": "lfortran --no-color {infile}",
+    "infile": "tests/errors/matmul_unallocated_01.f90",
+    "infile_hash": "69fa6b271c924d4944c45f3f5d8f6a569ff838a441c57aa9cbb649e0",
+    "outfile": null,
+    "outfile_hash": null,
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "run-matmul_unallocated_01-39026d3.stderr",
+    "stderr_hash": "a94a1b0a8f4c857b17918aaa24c9ff8f59a7ce6552ca339b4d6a64bd",
+    "returncode": 1
+}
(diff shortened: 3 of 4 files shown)


krystophny:
... still have to wait how this plays with #9039 , let's merge it after that one.

integration_tests/matmul_03.f90
```diff
@@ -0,0 +1,17 @@
+program matmul_03
+implicit none
+real, allocatable :: A(:,:), B(:,:), C(:,:)
+real :: expected
+
+allocate(A(3, 4), B(4, 2), C(3, 2))
```
certik, line 6: Let's also add a test where `C` is not allocated, and ensure it works with `--realloc-lhs-arrays`, and that it gives a runtime error without i.

krystophny:
Added `matmul_04.f90` test where C is not pre-allocated:

- With `--realloc-lhs-arrays`: works correctly, C is auto-allocated
- Without the flag: runtime error as expected:
  ```
  Runtime Error: Array 'result' is indexed but not allocated.
  ```

The integration test runs with the flag. The error case is verified manually but not added as a reference test due to test framework complications with runtime errors.

krystophny:
Update: Both tests now added as requested:

1. `integration_tests/matmul_04.f90` - C not pre-allocated, works with `--realloc-lhs-arrays` 
2. `tests/errors/matmul_unallocated_01.f90` - runtime error without the flag

Both tests pass locally.

HarshitaKalani (approved):
Looks good to me. Thanks!

tests/reference/run-matmul_unallocated_01-39026d3.stderr
```diff
@@ -0,0 +1,2 @@
+At 10:5 of file tests/errors/matmul_unallocated_01.f90
+Runtime Error: Array 'result' is indexed but not allocated.
```
certik, line 2: I had to think a while about what this error means: I think we transform matmul to loops, then we index A, and that's where this error comes from. That's fine for now.

certik (approved):
Looks great, let's see if tests pass.

Issue #9064, opened 2025-12-13
ICE: create_gep2 type mismatch for array of derived type in LLVM backend


krystophny:
### Bug
LLVM codegen aborts with a `create_gep2()` type mismatch when printing an array of a derived type.

### MRE
```fortran
module m
  implicit none
  type :: t
     real :: a
  end type t
contains
  function f() result(res)
    type(t) :: res(2,2)
    res%a = reshape([1.0,2.0,3.0,4.0],[2,2])
  end function f
end module m

program p
  use m
  implicit none
  print *, f()
end program p
```

### LFortran
Fails with:

```
LCOMPILERS_ASSERT failed: .../llvm_utils.cpp
function create_gep2(), line number 1480 at ds_pointee_type == t
Type mismatch in create_gep2: ... Target type: %t = type { float }, Pointer pointee type: [4 x %t]
```

Evidence:
- Source: `/tmp/lfortran-dev/fortplot/mre_create_gep2_array_of_struct_2025-12-13_212357.f90`
- LFortran log: `/tmp/lfortran-dev/fortplot/mre_create_gep2_array_of_struct_lfortran_2025-12-13_212357.txt`

### Control (gfortran)
GFortran compiles and prints `1 2 3 4`:
- Build log: `/tmp/lfortran-dev/fortplot/mre_create_gep2_array_of_struct_gfortran_2025-12-13_212357.txt`
- Run log: `/tmp/lfortran-dev/fortplot/mre_create_gep2_array_of_struct_gfortran_run_2025-12-13_212357.txt`

Brijesh-Thakkar:
@krystophny Hello, can i work on this issue??

krystophny:
> [@krystophny](https://github.com/krystophny) Hello, can i work on this issue??

You dont have to ask for my permission just do it, file a draft PR when ready, and remove draft status when all existing and new tests are passing. Looking forward!

Brijesh-Thakkar:
> > [@krystophny](https://github.com/krystophny) Hello, can i work on this issue??
> 
> You dont have to ask for my permission just do it, file a draft PR when ready, and remove draft status when all existing and new tests are passing. Looking forward!

Okk, will do it.

krystophny:
Fixed. The reproducer now compiles and runs correctly, printing '1 2 3 4'.

PR #9093, opened 2025-12-16
Fix: Incorrect GEP type for struct member access in array of derived types 

opixdown:
## **Issue Fixed**
#9064 ICE: create_gep2 type mismatch of derived type in LLVM backend 

## **Summary**
The issue occurred because the GEP instruction was using the wrong type it used the struct type instead of the array of structs type, leading to incorrect pointer .

## **Reference Code**
```
module m
  implicit none
  type :: t
     real :: a
  end type t
contains
  function f() result(res)
    type(t) :: res(2,2)
    res%a = reshape([1.0,2.0,3.0,4.0],[2,2])
  end function f
end module m

program p
  use m
  implicit none
  print *, f()
end program p

```

```
ICE: Type mismatch in create_gep2: ... Target type: %t = type { float }, Pointer pointee type: [4 x %t]

```

```
1.00000000e+00    2.00000000e+00    3.00000000e+00    4.00000000e+00
  
```

### DIFF
--- a/integration_tests/struct_type_07.f90
+++ b/integration_tests/struct_type_07.f90
@@ -1,4 +1,4 @@
-module m
+module struct_type_07_m
   implicit none
   type :: t
      real :: a
@@ -8,10 +8,10 @@ contains
     type(t) :: res(2,2)
     res%a = reshape([1.0,2.0,3.0,4.0],[2,2])
   end function f
-end module m
+end module struct_type_07_m
 
 program struct_type_07
-  use m
+  use struct_type_07_m
   implicit none
   type(t) :: result(2,2)
   
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3589,14 +3589,6 @@ public:
         member_idx = name2memidx[current_der_type_name][member_name];
 
         xtype = name2dertype[current_der_type_name];
-        // Check if tmp points to an array of the struct type
-        if (tmp->getType()->isPointerTy()) {
-            llvm::Type* pointee_type = llvm::cast<llvm::PointerType>(tmp->getType())->getElementType();
-            if (pointee_type->isArrayTy() && pointee_type->getArrayElementType() == xtype) {
-                // tmp points to an array of structs, use the array type for GEP
-                xtype = pointee_type;
-            }
-        }
         tmp = llvm_utils->create_gep2(xtype, tmp, member_idx);
         ASR::ttype_t* member_type = ASRUtils::type_get_past_pointer(
             ASRUtils::type_get_past_allocatable(member->m_type));
--- a/src/libasr/pass/array_struct_temporary.cpp
+++ b/src/libasr/pass/array_struct_temporary.cpp
@@ -1352,6 +1352,30 @@ class ArgSimplifier: public ASR::CallReplacerOnExpressionsVisitor<ArgSimplifier>
 
     void visit_Assignment(const ASR::Assignment_t& x) {
         ASR::Assignment_t& xx = const_cast<ASR::Assignment_t&>(x);
+        // Handle case where LHS is StructInstanceMember over an array
+        // e.g., res%a = reshape([1.0,2.0,3.0,4.0],[2,2]) where res is an array
+        if (ASR::is_a<ASR::StructInstanceMember_t>(*xx.m_target)) {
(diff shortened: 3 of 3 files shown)


opixdown:
@certik here i am using
` llvm::Type* pointee_type = llvm::cast<llvm::PointerType>(tmp->getType())->getElementType();`

but the problem is LLVM 15 enabled opaque pointers by default, and LLVM 17 removed typed pointers entirely so it fails for llvm > 15 so can you suggest any other approach?

integration_tests/struct_type_07.f90
```diff
@@ -0,0 +1,26 @@
+module m
```
certik: You have to make `m` unique such as by prepending the `struct_type_07` to it, otherwise CI fails.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3589,6 +3589,14 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
 
         xtype = name2dertype[current_der_type_name];
+        // Check if tmp points to an array of the struct type
+        if (tmp->getType()->isPointerTy()) {
+            llvm::Type* pointee_type = llvm::cast<llvm::PointerType>(tmp->getType())->getElementType();
+            if (pointee_type->isArrayTy() && pointee_type->getArrayElementType() == xtype) {
+                // tmp points to an array of structs, use the array type for GEP
+                xtype = pointee_type;
+            }
+        }
```
certik: I think this doesn't work in all LLVM versions. You have to use ASR to figure out what types we have of everything, and then you need to use the variables in the ASR->LLVM backend that map ASR type to LLVM type to determine the LLVM type.

Do not use LLVM to query types, since this is LLVM version specific and later LLVM versions do not even have that information in pointers.

assem2002:
Actually the bug is within our ASR passes. 
```
res%a = reshape([1.0,2.0,3.0,4.0],[2,2])
```
When you dump the passes, You'll find that we don't iterate res + we don't create a temp for `reshape`.
After passes we should end up with something like
```.f90
do while (__libasr_index_0_ + 1 <= 4 + 1 - 1)
        res(__libasr_index_0_)%a = reshape_temp_array(__libasr_index_0_1)
    end do
```
I think a good place to start is to go for `array_struct_temporary` pass and create a temp for reshape when the LHS is a ` structInstanceMember over an array` -- After that guarantee that `array_op` will do the loop of the array `res(__libasr_index_0_)%a`

assem2002:
current pass dump for the MRE mentioned in the issue
```f90
base) opixdown@192 lfortran % cat pass_fortran_28_insert_deallocate.f90
module m
implicit none
type :: t
    real(4) :: a
end type t

contains

subroutine f(res)
    real(4), dimension(2, 2), parameter :: __libasr_created__reshape_temp_ = [1.00000000e+00, 2.00000000e+00,&
         3.00000000e+00, 4.00000000e+00]
    integer(4) :: __libasr_index_1_
    integer(4) :: __libasr_index_1_1
    type(t), dimension(2, 2), intent(out) :: res
    ! CheckBounds LHS = res%a, RHS = __libasr_created__reshape_temp_,
    __libasr_index_1_1 = lbound(__libasr_created__reshape_temp_, 2)
    __libasr_index_1_ = int(1, kind=4) - 1
    do while (__libasr_index_1_ + 1 <= 2 + 1 - 1)
        __libasr_index_1_ = __libasr_index_1_ + 1
        __libasr_index_0_1 = lbound(__libasr_created__reshape_temp_, 1)
        __libasr_index_0_ = int(1, kind=4) - 1
        do while (__libasr_index_0_ + 1 <= 2 + 1 - 1)
            res(__libasr_index_0_, __libasr_index_1_)%a = __libasr_created__reshape_temp_(__libasr_index_0_1,&
         __libasr_index_1_1)
        end do
        __libasr_index_1_1 = __libasr_index_1_1 + 1
    end do
end subroutine f

end module m

program p
use m, only: f
use m, only: t
implicit none
type(t), dimension(2, 2) :: __libasr_created_string_format
call f(__libasr_created_string_format)
print *, __libasr_created_string_format
end program p
```

I think that is correct.

assem2002 (approved):
Looks good. Thanks!

certik (approved):
Looks good, thanks!

Issue #9095, opened 2025-12-16
Consolidate assumed-size error test subroutines


krystophny:
Consolidate separate test subroutines into one with multiple arguments for conciseness.

**Current:** `tests/errors/continue_compilation_1.f90` lines 77-98
```fortran
subroutine ubound_assumed_size_1(a)
    real :: a(*)
    print *, ubound(a, 1)
end subroutine

subroutine ubound_assumed_size_2(a)
    real :: a(*)
    print *, ubound(a)
end subroutine
! ... 3 more similar subroutines
```

**Suggested:**
```fortran
subroutine assumed_size_errors(a, b, c, d, e)
    real :: a(*), b(*), c(10, *)
    real, intent(in) :: d(*, 10)
    real :: e(*, 10)
    print *, ubound(a, 1)  ! error: last dim assumed-size
    print *, ubound(b)     ! error: DIM required
    print *, ubound(c, 2)  ! error: last dim assumed-size
    ! d, e trigger star-position errors
end subroutine
```

Reference: PR #9073 review comment

PR #9098, opened 2025-12-16
tests: consolidate assumed-size error subroutines

amritamishra01:
This PR consolidates multiple assumed-size error test subroutines in
tests/errors/continue_compilation_1.f90 into a single subroutine with
multiple arguments, as suggested in #9095.

This is a refactor for conciseness only; no behavior or expected error
output is changed.

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "afef5466980767e641ddb48778aefc3b1f4f68e6e5103c44562d8ec852e98b76",
+    "stderr_hash": "6c059bd9b4879bb20f0a8e595776d33fa3d41b67c42bf7f0d5b6f2709e8f8e5a",
     "returncode": 1
 }


certik:
I think you need to update the reference tests to show the errors.

tests/errors/continue_compilation_1.f90
```diff
@@ -74,7 +74,15 @@ subroutine s(c) bind(c)
...
-
+    subroutine assumed_size_errors(a, b, c, d, e)
+        real :: a(*), b(*), c(10, *)
+        real, intent(in) :: d(*, 10)
+        real :: e(*, 10)
+        print *, ubound(a, 1)  ! error: last dim assumed-size
+        print *, ubound(b)     ! error: DIM required
+        print *, ubound(c, 2)  ! error: last dim assumed-size
+        ! d, e trigger star-position errors
+    end subroutine assumed_size_errors
```
certik, line 85: You have to remove as many lines as you added, to keep the diff below minimal.

certik:
Also, I would wait until https://github.com/lfortran/lfortran/pull/9073 is merged, since that's the PR that actually introduces the errors I think.

certik:
Ok, the other PR is merged, you can now rebase and get it finished if you want.

amritamishra01:
hi @certik I have updated the branch with the latest changes and fixed the reference mismatches as requested. All CI checks are now passing, and this is ready for review.

certik:
Thank you!

The diff is still not minimal -- isn't there a way to adjust the empty lines, so that the other unrelated error messages do not change? Then it will be obvious that we didn't drop any error message.

certik:
Once you fix it, go ahead and click on "Ready for review".

tests/errors/continue_compilation_1.f90
```diff
@@ -74,28 +74,34 @@ subroutine s(c) bind(c)
...
+
+
+
+
+
+
+
+
+
+
```
HarshitaKalani, line 100:
```suggestion
```
If you remove these 6 additional lines, the references updated will be minimal. (This is because you've removed 18 lines but added 24.

tests/errors/continue_compilation_1.f90
```diff
@@ -455,3 +461,6 @@ program continue_compilation_1
     13 format ()
     read (5, fmt_i3)
 end program 
+
+
+
```
HarshitaKalani: This is not needed.

PR #9101, opened 2025-12-16
fix: don't create array item from array section if start and end variables are same

### LINKED ISSUE #9099: Handle Reallocation of array to an `ArraySection` with only 1 element
MRE:
```fortran
program mre
    implicit none
    integer :: i0
    integer, dimension(:), allocatable :: subpass
    integer :: temp(2)
    i0 = 1
    subpass = [2,3]
    subpass = temp(i0:i0)
    print *, size(subpass)
end program mre
```

Output: 
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran a.f90 && ./a.out
           1
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 --realloc-lhs-arrays
2
```

HarshilShah1804:
Fixes #9099 

We can have another approach of directly reallocating the array in the `array_op` pass where we handle reallocation for array sections. But, I believe this is a more accurate fix because the behavior of array sections with same `m_start` and `m_end` is different from that of array item. 

For example if we consider the MRE in the issue:
```fortran
program mre
    implicit none
    integer :: i0
    integer, dimension(:), allocatable :: subpass
    integer :: temp(2)
    i0 = 1
    subpass = [2,3]
    subpass = temp(i0:i0)
    print *, size(subpass)
end program mre
```

The expected output here is `1`. But if we change the array section with same `m_start` and `m_end`, from `temp(i0:i0)` to `temp(i0)` as we were doing earlier, the expected output should be `2` only as we are getting.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2243,6 +2243,7 @@ RUN(NAME array_section_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStac
 RUN(NAME array_section_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME array_section_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran)
 RUN(NAME array_section_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
+RUN(NAME array_section_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --realloc-lhs-arrays)
 
 RUN(NAME nested_vars_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 
--- a/integration_tests/array_section_07.f90
+++ b/integration_tests/array_section_07.f90
@@ -1,0 +1,11 @@
+program array_section_07
+    implicit none
+    integer :: i0
+    integer, dimension(:), allocatable :: subpass
+    integer :: temp(2)
+    i0 = 1
+    subpass = [2,3]
+    subpass = temp(i0:i0)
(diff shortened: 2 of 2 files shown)


certik:
It failed with:
```
[1612/2119] Building Fortran object SRC/CMakeFiles/lapack.dir/lsamen.f.o
FAILED: SRC/CMakeFiles/lapack.dir/lsamen.f.o 
/home/runner/work/lfortran/lfortran/src/bin/lfortran -I/tmp/tmp.DrsWvaC0ev/assert/modern_minpack_01/modern_minpack_02/dftatom/fastGPT/stdlib/SNAP/src/lapack/build/lapack-vanilla/SRC --fixed-form-infer --implicit-interface --legacy-array-sections --separate-compilation -c SRC/CMakeFiles/lapack.dir/lsamen.f-pp.f -o SRC/CMakeFiles/lapack.dir/lsamen.f.o
ASR verify pass error: ASR::ttype_t in ASR::FunctionType cannot be tied to a scope.
   --> SRC/CMakeFiles/lapack.dir/lsamen.f-pp.f:108:19
    |
108 |          IF( .NOT.LSAME( CA( I: I ), CB( I: I ) ) )
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


```

Please first create an MRE for this failure, add to integration tests, then let's fix it in this PR.

HarshilShah1804:
I have opened a PR for adding the MRE #9140. I am now working to fix the changes in this PR to make it compatible with the given MRE.

HarshilShah1804:
I think it was just a small issue. We were treating string section also in the same way as array section. But, I guess the intended changes were for arrays only and not string. So, we keep the string section handling as it is and just modify the array section handling.

I have tested the LAPACK MRE and the MRE in the original issue. Both work now.

certik:
Excellent. The test in #9140 still works with this PR?

PR #8894, opened 2025-11-23
Fix implied-do loop handling for deferred-length CHARACTER array constructors

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/8179

PR #9059, opened 2025-12-13
Implement position specifier for fileopen statement

opixdown:
closes #8994 
extension of #9004

PR #9004, opened 2025-12-08
Feat : Implement `position` specifier for fileopen statement

opixdown:
## **Issue Fixed**
#8994  Handle position inside FileOpen

## **Summary**

This PR adds support for the position parameter including "asis" ,"rewind" ,"append" in fortran OPEN statement, allowing users to control file positioning when opening files.

## **Problem**

LFortran did not support the position specifier in OPEN statements. 
"append" - Position at end of file
"rewind" - Position at beginning of file 
"asis" - Keep current position 

## **Reference Code**
```
program test_inquire_pos
    implicit none
    integer :: io
    integer :: length
    character(len=*), parameter :: txt = "ABC"

    open(newunit=io, file="pos_test.bin", access="stream", form="unformatted")
    write(io) txt
    close(io)
    open(file="pos_test.bin", &
      & status="old", &
      & access="stream", & 
      & position="append", &
      & newunit=io)
    inquire(unit=io, pos=length)
    close(io)
    print *, "Position =", length

end program test_inquire_pos
```

## **Behaviour before this PR**

```
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran test.f90 && ./a.out
 Position =           4
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90
warning: Argument `position` isn't supported yet
  --> test.f90
   |
10 |        open(file="pos_test.bin", &
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
14 |          & newunit=io)
   | ...^^^^^^^^^^^^^^^^^^ ignored for now


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
Position =    1
```

## **Behaviour after this PR**
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran test.f90 && ./a.out
 Position =           4
 
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran test.f90
Position =    4
```

## **Support added**

- [✓] position="append" 
- [✓]  position="rewind" 
- [✓]  position="asis" 
- [✓]  No position specified - defaults to "asis" 
- [✓]  Invalid position value - properly reports erro

PR #9102, opened 2025-12-16
Implement position specifier for fileopen statement 

opixdown:
closes #8894 
fix error of #9059 
extension of #9004

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -454,4 +454,12 @@ program continue_compilation_1
     assign 13 to fmt_i3
     13 format ()
     read (5, fmt_i3)
+
+    ! open_duplicate_position
+    open(unit=10, file='test.txt', position='rewind', position='append')
+    ! open_invalid_position_type_integer
+    q1 = 1
+    open(unit=20, file='test2.txt', position=q1)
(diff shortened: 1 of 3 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -276,6 +276,13 @@ program continue_compilation_1
...
     ! open_invalid_kwarg1
     OPEN(file="numbers", hello="world")
+    ! open_duplicate_position
+    open(unit=10, file='test.txt', position='rewind', position='append')
+    ! open_invalid_position_type_integer
+    q1 = 1
+    open(unit=20, file='test2.txt', position=q1)
+    ! open_invalid_position_type_real
+    open(unit=30, file='test3.txt', position=r1)
     !parameter_01
```
certik, line 279: You need to add these at the end, so that it keeps the diff below minimal.

opixdown: okay ondrej

certik:
I think we have to merge https://github.com/lfortran/lfortran/pull/9004 first, correct?

opixdown:
i think there are 3 pr connected to each other i will try to add all the changes into one branch and then make a PR with that

PR #9108, opened 2025-12-17
fix: finalize derived-type arrays on deallocate

### LINKED ISSUE #9097: Complete allocatable INTENT(OUT) semantics to enable strict double-allocation checks
## Summary
After #9039 (double-allocation runtime checks in debug mode), we need complete Fortran semantics for allocatable `intent(out)` dummy arguments on procedure entry: deallocation plus correct finalization/deep deallocation where required.

## Minimal Reproducer
```fortran
program test_intent_out
  implicit none

  allocate(arr(5))
  call reset_array(arr)

contains
  subroutine reset_array(x)
    integer, allocatable, intent(out) :: x(:)
    allocate(x(3))
  end subroutine
end program
```

## Follow-ups
- Deep finalization for allocatable derived-type arrays: #9109 (implemented in PR #9108)
- Deep finalization for allocatable derived-type scalars: #9110
- Optional allocatable `intent(out)` entry semantics: #9111
- Allocatable polymorphic/class `intent(out)` deallocation/finalization: #9112
- Re-enable `intent(out)` allocation checks from #9039 incrementally: #9113

krystophny:
## Summary
Fixes #9097 by ensuring derived-type arrays are finalized before explicit `deallocate`, so nested allocatable components get deallocated before the outer array storage is freed.

## What changed
- LLVM backend: add `LLVMFinalize::finalize_before_deallocate()` in `src/libasr/codegen/llvm_utils.h` and wire it from `src/libasr/codegen/asr_to_llvm.cpp` for explicit `deallocate` on derived-type arrays.
- Re-enable the runtime double-allocation check for `intent(out)` allocatable arguments now that nested deallocation is implemented.
- Update the single affected LLVM reference test `integration_tests/allocate_03.f90`.

## Evidence (local)
- Reproduced the failing reference mismatch and updated only the `allocate_03` LLVM references:
  - `/tmp/lfortran-dev/9108/test_allocate_03_2025-12-18.txt`
  - `/tmp/lfortran-dev/9108/test_allocate_03_rerun_2025-12-18.txt`
- Targeted intent(out) nested dealloc integration test:
  - `/tmp/lfortran-dev/9108/test_intent_out_nested_dealloc_2025-12-18.txt`

## CI
CI is running for the latest push.

### DIFF
--- a/tests/reference/asr-modules_29-dc71c55.json
+++ b/tests/reference/asr-modules_29-dc71c55.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-modules_29-dc71c55.stdout",
-    "stdout_hash": "652c496f5adc797204edcb432d61d8e2fd18729af483178040154d5a",
+    "stdout_hash": "d9a6ff0cf55bdf9c52dbf53afa894d4319d48bd59e738ea21b6878b3",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/asr-modules_29-dc71c55.stdout
+++ b/tests/reference/asr-modules_29-dc71c55.stdout
@@ -1436,7 +1436,43 @@
                                     (Var 6 table)
                                     (Var 6 error)]
                                     [(ImplicitDeallocate
-                                        [(Var 6 error)]
+                                        [(StructInstanceMember
+                                            (Var 6 self)
+                                            6 1_executable_config_t_dependency
+                                            (Allocatable
+                                                (Array
+                                                    (StructType
+                                                        [(Allocatable
+                                                            (StructType
+                                                                [(Allocatable
+                                                                    (String 1 () DeferredLength DescriptorString)
+                                                                )
+                                                                (Allocatable
+                                                                    (String 1 () DeferredLength DescriptorString)
+                                                                )]
+                                                                []
+                                                                .true.
+                                                                .false.
+                                                            )
+                                                        )
+                                                        (Allocatable
+                                                            (String 1 () DeferredLength DescriptorString)
+                                                        )
+                                                        (Allocatable
+                                                            (String 1 () DeferredLength DescriptorString)
+                                                        )]
+                                                        []
+                                                        .true.
+                                                        .false.
+                                                    )
+                                                    [(()
+                                                    ())]
+                                                    DescriptorArray
+                                                )
+                                            )
+                                            ()
+                                        )
+                                        (Var 6 error)]
                                     )
                                     (SubroutineCall
                                         4 new_dependencies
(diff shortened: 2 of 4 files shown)


assem2002:
Need to update reference tests.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1341,6 +1346,27 @@ if(get_struct_sym(member_variable) == struct_sym /*recursive declaration*/){cont
...
 /*>>>>>>>>>>>>>>>>>>>>> Entry <<<<<<<<<<<<<<<<<<<<<<< */
 
+        void finalize_before_deallocate(llvm::Value* const ptr, ASR::ttype_t* const t,
+                ASR::Struct_t* const struct_sym, bool in_struct){
+            ASR::ttype_t* const t_past = ASRUtils::type_get_past_allocatable_pointer(t);
+            switch(t_past->type) {
+                case ASR::Array: {
+                    ASR::Array_t* const arr_t = ASR::down_cast<ASR::Array_t>(t_past);
+                    if (arr_t->m_type->type != ASR::StructType) { return; }
+                    if (!is_finalizable_type(arr_t->m_type, struct_sym)) { return; }
```
assem2002, line 1355: **MINOR**
```suggestion
                    if (!is_finalizable_type(&arr_t->base, struct_sym)) { return; }
```

krystophny:
Re: the MINOR suggestion in `finalize_before_deallocate()`: `is_finalizable_type(&arr_t->base, ...)` is always true for DescriptorArray/PointerArray (because the array itself needs finalization/deallocation at scope exit). Here we intentionally gate on `arr_t->m_type` (the element StructType) so we only emit the per-element finalization loop when the derived type actually contains finalizable components.

certik:
Thanks for fixing this! @assem2002 and I discussed this PR, we think the approach is correct.

Later we also need to implement user-defined finalizers and call them also, at the right moment. And it should all happen in the backend, when the symbol table goes out of scope, recursively through structs and their members as needed.

krystophny:
Thanks, @assem2002 and @certik !

integration_tests/intent_out_nested_dealloc.f90
```diff
@@ -0,0 +1,57 @@
+! Test for intent(out) nested allocatable deallocation
+! Issue #9097: nested allocatables should be deallocated at procedure entry
+!
+! Verification: Run with valgrind to detect memory leak
+!   gfortran:   All heap blocks freed (correct)
+!   ifx:        All heap blocks freed (correct)
+!   nvfortran:  No leaks (correct)
+!   lfortran:   160 bytes definitely lost (BUG - nested data not freed)
```
certik: Is this after this PR or before this PR?

krystophny: > Is this after this PR or before this PR?

Before the PR.

krystophny:
Full branch moved to #9163 and just the one commit enabling checks lives here. Will be rebased on main as patches arrive.

krystophny:

PR #9114, opened 2025-12-17
CI: enable all test packages of fpm at CI

jinangshah21:
Towards #6840 

With this PR, 223/238 fpm tests are passing, with workarounds.


certik:
I rebased it, since I merged a PR that checks double allocation, so that it doesn't fail.

certik (approved):

Issue #5733, opened 2024-12-16
bug: procedure call on a derived type `StructInstanceMember` throws error


kmr-srbh:
### Minimal Reproducible Example

```fortran
module main_module

   type :: primary
      integer :: x
   contains
      procedure :: call_fun => fun
   end type

   type :: secondary
      type(primary) :: t
   end type secondary

contains

   function fun(x) result(val)
      class(primary), intent(in) :: x
      integer :: val
      val = 1
   end function fun

end module main_module


program main

   use main_module

   type(secondary) :: vals
   print *, vals%t%call_fun()

end program main
```

### Output

```console
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran test.f90 && ./a.out
           1
(base) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran test.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

%primary_polymorphic = type { i64, %primary* }
%primary = type { i32 }
%__vtab_primary = type { i64 }
%secondary = type { %primary }

@0 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@1 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define i32 @__module_main_module_fun(%primary_polymorphic* %x) {
.entry:
  %0 = alloca %__vtab_primary, align 8
  %1 = getelementptr %__vtab_primary, %__vtab_primary* %0, i32 0, i32 0
  store i64 0, i64* %1, align 4
  %val = alloca i32, align 4
  store i32 1, i32* %val, align 4
  br label %return

return:                                           ; preds = %.entry
  %2 = load i32, i32* %val, align 4
  ret i32 %2
}

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %vals = alloca %secondary, align 8
  %2 = getelementptr %secondary, %secondary* %vals, i32 0, i32 0
  %3 = getelementptr %primary, %primary* %2, i32 0, i32 0
  %4 = alloca %primary_polymorphic, align 8
  %5 = getelementptr %primary_polymorphic, %primary_polymorphic* %4, i32 0, i32 0
  store i64 0, i64* %5, align 4
  %6 = getelementptr %secondary, %secondary* %vals, i32 0, i32 0
  %7 = load %primary_polymorphic*, %primary* %6, align 8
[...]

PR #9118, opened 2025-12-18
test: add regression #5733

Nayan-Kute21:
fixes lfortran/lfortran#5733
Adds a simple test that reproduces the original problem (calling a method on a nested field) and verifies it runs correctly.


HarshilShah1804:
Thank you for the PR @Nayan-Kute21. The failure in the CI was a network issue. I have rerun the failed check.

certik:
Thanks! I think this is good, tests pass, I think we don't need to run full tests.

Issue #9100, opened 2025-12-16
throw error for mismatched dimensions while assigning using array sections


HarshilShah1804:
MRE
```fortran
program main
    integer :: a(2) = [2, 3]
    integer :: temp(2)
    integer :: i0 = 1

    a = temp(i0:i0)
    print *, size(a)
end program 
```

LFortarn
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
2
```

GFortran
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
./examples/expr2.f90:17:4:

   17 |     a = temp(i0:i0)
      |    1
Error: Different shape for array assignment at (1) on dimension 1 (2 and 1)
```

Brijesh-Thakkar:
@HarshilShah1804 I have closed that PR, actually i messed up things, really sorry for that.
I will raise a new PR shortly and will attach the PR link here and also notify you regarding the same

HarshilShah1804:
@Brijesh-Thakkar No worries. These things happen. Please go ahead and open a new PR.

opixdown:
@HarshilShah1804  i think the latest main somehow fixed the error we can add test and close it. 

```lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran ./test.f90 
At 6:5 of file ./test.f90
Runtime Error: Array shape mismatch in assignment to 'a'

Tried to match size 2 of dimension 1 of LHS with size 1 of dimension 1 of RHS.```

HarshilShah1804:
Yes, @opixdown, you are correct. But I believe a semantic error at compile time only would be better than the runtime error. But we can improve it later.

PR #9120, opened 2025-12-18
Fix lowering of single-element array slices (i:i) to ArraySection (#9100)

Brijesh-Thakkar:
Problem

Before this change, LFortran treated:

temp(i:i)

as equivalent to:

temp(i)

so assignments like:

a = temp(i:i)

(where a has size 2 and the slice has size 1) compiled without reporting
a shape mismatch error.

GFortran rejects this, but LFortran accepted it.

Fix

In create_ArrayRef():

Removed the logic that collapsed lb == ub slices into scalar accesses.

Added explicit detection of colon usage (m_start/m_end non-null).

When a colon is present, the reference is now always lowered as an
ArraySection, regardless of whether the bounds are equal.

Result

LFortran now behaves correctly:

Runtime Error: Array shape mismatch in assignment to 'a'
Tried to match size 2 of LHS with size 1 of RHS.

This matches GFortran behavior.

Tests

Added regression test:

tests/errors/array_section_shape_mismatch_01.f90

to ensure incorrect temp(i:i) assignments correctly report shape mismatch.


HarshilShah1804:
Hello @Brijesh-Thakkar Thank you for the PR. I'll request you to please keep the changes in the PR as minimal as possible. Currently, the PR is having a lot of formatting related changes. With minimal changes, it would be easy to review why the PR is failing the CI checks.

I have marked it as draft now. You can click on ready to review once you remove the redundant changes from the PR.

PR #9121, opened 2025-12-18
fix: bounds check in array_op for rank-mismatched assignments

krystophny:
Merge order: #9119 -> #9121 -> #9122 -> #9124

## Summary
Fix compile-time segfault in `array_op` when generating loops for assignments involving rank-mismatched arrays (e.g. 3D target assigned from a rank-1 temporary created from a DATA ArrayConstant).

## Diff vs #9119
- `src/libasr/pass/array_op.cpp`: bounds-check `var2indices[i].p[loop_depth]` accesses for lower-rank arrays
- `integration_tests/equivalence_12.f90`: MRE that previously triggered the segfault

## Test plan
- `integration_tests/equivalence_12.f90`

PR #9122, opened 2025-12-18
fix: COMMON block member aliases across program units

krystophny:
Merge order: #9119 -> #9121 -> #9122 -> #9124

## Summary
Handle COMMON blocks that use different local variable names for the same storage locations across program units (valid Fortran).

## Diff vs #9121
- `integration_tests/common_15.f90`: MRE for COMMON variable name aliases
- `integration_tests/CMakeLists.txt`: enable `common_15`
- `src/lfortran/semantics/*`: track COMMON member position so members can be resolved by index when local names differ
- `integration_tests/equivalence_12.f90`: removed (re-added in #9124)

## Test plan
- `integration_tests/common_15.f90`

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1442,6 +1442,7 @@ RUN(NAME select_type_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST)
 RUN(NAME select_type_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_type_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME program_02 LABELS gfortran llvm)
 RUN(NAME program_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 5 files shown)


certik:
I think it's fine, but I'll review carefully once the first two PRs are in.

certik:
One has to use a union approach. Is the implementation here effectively implementing a union approach, with the restriction that all members have the same length (possibly the same type), so they only differ by a name?

That might fix this problem (by adding complexity to the implementation), but it will not fix if the two common blocks have both different names and different size/types. Right? 

The proper fix is to use a union of the common block structs, if it is in the same translation unit.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2253,17 +2258,34 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                // Variable name differs - look up by position index
+                auto idx_it = common_variables_member_index.find(hash);
+                if (idx_it != common_variables_member_index.end()) {
+                    size_t member_idx = idx_it->second;
+                    if (member_idx < struct_type->n_members) {
+                        actual_member_name = struct_type->m_members[member_idx];
+                        struct_member_sym = struct_type->m_symtab->resolve_symbol(actual_member_name);
+                    }
+                }
+            }
```
certik, line 2276: I think you can't find it by the position index, but rather by the size of each variable, in the union fashion, right?

krystophny: Position index works here because we're restricting to same-shape declarations - position N in unit A and position N in unit B have the same type/size by assumption. For different shapes, we could use byte offsets instead.

certik: Ok. Should we merge this temporary workaround, or implement the proper solution?

certik: Also, is the above logic only happening for common block, or for any struct now?

We only want to do it for a common block, as a temporary workaround.

krystophny: Full solution now on #9199 with a „blob“ kind of storage. We can use that one and close #9122 .

certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/9199.

PR #9127, opened 2025-12-18
formatted READ: support FORMAT labels + mixed types

krystophny:
Summary
- Fix formatted `READ` when the format is given via a statement label (e.g. `10 FORMAT(...)`) and the I/O list contains mixed types.

Key changes
- `src/libasr/codegen/asr_to_llvm.cpp`: `visit_FileRead` extracts labeled FORMAT and routes formatted input through the mixed-type path.
- `src/libasr/runtime/lfortran_intrinsics.c` + `src/libasr/runtime/lfortran_intrinsics.h`: add `_lfortran_formatted_read2`.
- `integration_tests/read_11.f90`: regression test (reduced from LAPACK).

How to run
- `./run_tests.py -t read_11 -s`

PR #9124, opened 2025-12-18
test: add MRE for EQUIVALENCE 3D array DATA issue

krystophny:
Merge order: #9119 -> #9121 -> #9122 -> #9124

## Summary
Add an MRE for the remaining EQUIVALENCE + DATA 3D-array issue tracked in #9125.

## Diff vs #9122
- `integration_tests/equivalence_12.f90`: MRE (not added to CMakeLists since it currently fails)

## Test plan
- Compile/run manually while working on #9125

PR #9128, opened 2025-12-18
COMMON: fix character substring sections

krystophny:
Summary
- Fix substring sections on `COMMON`-backed character variables (e.g. `srnamt(1:3)`), avoiding ASR verification errors.

Key changes
- `src/lfortran/semantics/ast_common_visitor.h`: handle `StringSection` when the base variable comes from a `COMMON` block.
- `integration_tests/common_substring_01.f90`: regression test.

How to run
- `./run_tests.py -t common_substring_01 -s`

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2412,7 +2412,7 @@ RUN(NAME common_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME common_substring_01.f90 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME common_substring_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME statement_01 LABELS llvm gfortran llvmImplicit)
 RUN(NAME statement_02 LABELS gfortran llvmImplicit)


certik:
I think this depends on https://github.com/lfortran/lfortran/pull/9122

krystophny:
> I think this depends on #9122

Yes indeed. The merge train is #9119 - #9121 - #9122 - #9124 - #9127 - #9128 . If you are fine with this minimal fix that helps LAPACK to compile further, we can do a more general fix later.

certik (approved):

certik:
If tests pass, we can merge it.

integration_tests/common_substring_01.f90
```diff
@@ -0,0 +1,29 @@
...
+    character(len=32) :: srnamt
+    integer :: nout
+    common /srname/ srnamt
+    common /iounit/ nout
+
+    ! Substring read triggers StringSection on COMMON-backed character
+    if (srnamt(1:1) /= "X") error stop "substring mismatch"
+
+    write(nout, *) srnamt(1:6)
+end subroutine
```
certik, line 29: Hold on, this is not registered....

certik:
I tested it locally, the test works and it doesn't work in main, so it looks good.

certik:
Some other test fails, so I think this PR has to be improved.

certik:
Actually, the main fails, this PR was passing before, so it's probably ok.

certik:
I reverted the offending commit and rebased this again. It should pass now hopefully.

integration_tests/common_substring_01.f90
certik: I registered it.

integration_tests/CMakeLists.txt
```diff
@@ -2412,6 +2412,7 @@ RUN(NAME common_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME common_substring_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
certik, line 2415: Note: this should just use a numerical name, to make the tests uniform, so:

```suggestion
RUN(NAME common_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```

Issue #7304, opened 2025-05-14
Code generation fails for subroutine with implicit interface


difference-scheme:
The following valid code compiles with both ifx, gfortran, and flang-20. With lfortran the result is as follows:
$ cat test.f90
```
subroutine some_subroutine()
   real(8), allocatable :: a(:,:)      
   call other_subroutine(a)
end subroutine some_subroutine

```
$ lfortran -c --implicit-interface --generate-object-code test.f90
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

%array = type { double*, i32, %dimension_descriptor*, i1, i32 }
%dimension_descriptor = type { i32, i32, i32 }

define void @some_subroutine() {
.entry:
  %a = alloca %array*, align 8
  store %array* null, %array** %a, align 8
  %arr_desc = alloca %array, align 8
  %0 = getelementptr %array, %array* %arr_desc, i32 0, i32 2
  %1 = alloca i32, align 4
  store i32 2, i32* %1, align 4
  %2 = load i32, i32* %1, align 4
  %3 = alloca %dimension_descriptor, i32 %2, align 8
  store %dimension_descriptor* %3, %dimension_descriptor** %0, align 8
  %4 = getelementptr %array, %array* %arr_desc, i32 0, i32 4
  store i32 2, i32* %4, align 4
  %5 = getelementptr %array, %array* %arr_desc, i32 0, i32 0
  store double* null, double** %5, align 8
  store %array* %arr_desc, %array** %a, align 8
  %6 = load %array*, %array** %a, align 8
  %7 = getelementptr %array, %array* %6, i32 0, i32 0
  %8 = load double*, double** %7, align 8
  %9 = getelementptr %array, %array* %6, i32 0, i32 1
  %10 = load i32, i32* %9, align 4
  %11 = getelementptr inbounds double, double* %8, i32 %10
  call void @other_subroutine(double* %11)
  %12 = load %array*, %array** %a, align 8
  %13 = getelementptr %array, %array* %12, i32 0, i32 0
  %14 = load double*, double** %13, align 8
  %15 = ptrtoint double* %14 to i64
  %16 = icmp ne i64 %15, 0
  br i1 %16, label %then, label %else

then:                                             ; preds = %.entry
  %17 = getelementptr %array, %array* %12, i32 0, i32 0
  %18 = load double*, double** %17, align 8
  %19 = alloca i8*, align 8
  %20 = bitcast double* %18 to i8*
[...]

PR #9129, opened 2025-12-18
test: allocatable array with implicit interface

krystophny:
Summary
- Add a regression test for an LLVM type mismatch when passing an allocatable array to a procedure with an implicit interface.

Key changes
- `integration_tests/implicit_interface_20.f90`: new reduced test (LAPACK-derived; issue #7304 pattern).
- `integration_tests/CMakeLists.txt`: register the new test.

How to run
- `./run_tests.py -t implicit_interface_20 -s`


certik:
@krystophny this PR doesn't make any changes, just adding a test --- is the test somehow fixed now, and before it was failing?

certik:
The test didn't run in full tests, so I reverted it: https://github.com/lfortran/lfortran/pull/9217.

Issue #9060, opened 2025-12-13
ICE: verify fail in intrinsic_subroutine with character(len=dummy) allocatable + move_alloc (fortplot trigger)


krystophny:
### Summary
LFortran crashes while compiling code that uses an allocatable character array dummy with a length that depends on another dummy (`character(len=string_len), allocatable :: buffer(:)`) and then uses `move_alloc()`.

In fortplot this is hit while building `app/update_example_index.f90` via fpm, where the crash currently shows up as an `expr_value0()` assert.

### Minimal reproducer (standalone, no fortplot dependencies)
Save as `mre.f90`:

```fortran
program update_example_index_mre
    implicit none

    character(len=:), allocatable :: names(:)

    call list_examples(names)
    print *, size(names)

contains

    subroutine list_examples(names)
        character(len=:), allocatable, intent(out) :: names(:)
        integer, parameter :: entry_len = 256
        integer, parameter :: initial_capacity = 64
        character(len=entry_len), allocatable :: entries(:)
        integer :: capacity

        capacity = initial_capacity
        call ensure_entry_capacity(entries, entry_len, capacity)
        allocate(character(len=1) :: names(0))
    end subroutine list_examples

    subroutine ensure_entry_capacity(buffer, string_len, new_capacity)
        integer, intent(in) :: string_len
        integer, intent(in) :: new_capacity
        character(len=string_len), allocatable, intent(inout) :: buffer(:)
        character(len=string_len), allocatable :: tmp(:)

        if (new_capacity <= 0) then
            allocate(character(len=string_len) :: tmp(1))
        else
            allocate(character(len=string_len) :: tmp(new_capacity))
        end if
        tmp = 
        call move_alloc(tmp, buffer)
    end subroutine ensure_entry_capacity

end program update_example_index_mre
```

Compile:

```bash
lfortran -c mre.f90
```

### Observed output (MRE)
This MRE currently fails with a verify error and an ICE:

```
ASR verify pass error: The symbol table was not found in the scope of `symtab`.
ASR verify pass error: Var::m_v `string_len` cannot point outside of its symbol table
Internal Compiler Error: Unhandled exception
LCompilersException: Verify failed in the pass: intrinsic_subroutine
```

Log from my run:
- `/tmp/mre4_build.txt`

### Fortplot trigger (context)
- fortplot: `lazy-fortran/fortplot` commit `252c3c7`
- build: `FPM_FC=/path/to/lfortran fpm build --profile release`
[...]

PR #9130, opened 2025-12-18
fix: handle deferred-length character arrays in move_alloc

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9060


certik:
Let's run the full tests just in case.

Issue #9045, opened 2025-12-13
Bug : call to intrinsic `allocated` with wrong output


assem2002:
```.f90
module node_mod

  type :: dependency_config_t
      integer, allocatable :: arr(:)
  end type dependency_config_t
  type, extends(dependency_config_t) :: dependency_node_t
  end type dependency_node_t

end module node_mod


program demo
  use node_mod
  implicit none

  type(dependency_node_t), allocatable :: a, b
  allocate(a, b)
  b = a
  print *, allocated(b%arr)
end program demo
```
```
T
```

It should return `F`

opixdown:
on it !!

HarshilShah1804:
It only happens for scalars like integer, real etc. and not for struct type variables. For example, if we just change the type of `arr` from `integer` to `type(dependency_config_t)`, it works same as GFortran,
```fortran
module node_mod

  type :: dependency_config_t
      type(dependency_config_t), allocatable :: arr(:)
  end type dependency_config_t
  type, extends(dependency_config_t) :: dependency_node_t
  end type dependency_node_t

end module node_mod


program demo
  use node_mod
  implicit none

  type(dependency_node_t), allocatable :: a, b
  allocate(a, b)
  b = a
  print *, allocated(b%arr), allocated(a%arr)
end program demo
```

Output:
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
F    F
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
 F F
```

What I feel is that the core problem of this issue is similar to #9013, automatic reallocation of scalar variables while assignment, but not for struct type members.

HarshilShah1804:
@assem2002 The MRE now works with the latest main. I think we can close this issue by adding the test for it.

PR #9131, opened 2025-12-18
fix: allocatable member handling in derived type assignment 

opixdown:
## **Issue Fixed**
#9045 Bug : call to intrinsic allocated with wrong output

## **Summary**
The issue occurred because allocatable members (scalar strings and arrays) in derived types were being copied unconditionally without checking their allocation status. This led to runtime errors when attempting to deepcopy unallocated members

## **Reference Code**
```
module node_mod

  type :: dependency_config_t
  end type dependency_config_t
  type, extends(dependency_config_t) :: dependency_node_t
  end type dependency_node_t

end module node_mod


program demo
  use node_mod
  implicit none

  type(dependency_node_t), allocatable :: a, b
  allocate(a, b)
  b = a
  print *, allocated(b%arr)
end program demo

```

```
(lf) opixdown@192 lfortran % lfortran t.f90
T

```

```
(lf) opixdown@192 lfortran % lfortran t.f90
F
  
```


opixdown:
closing the pr as the issue was solved with the latest main

PR #9133, opened 2025-12-19
Add dimension mismatch check for array section assignments

### LINKED ISSUE #9100: throw error for mismatched dimensions while assigning using array sections

Brijesh-Thakkar:
Fixes #9100

This commit adds compile-time validation to catch dimension mismatches when assigning array sections to arrays. Previously, LFortran would accept invalid code like:

    integer :: a(2), temp(2)
    a = temp(1:1)  ! Assigning 1 element to 2-element array

This now correctly produces an error at compile-time, matching GFortran's behavior:

    semantic error: Different shape for array assignment at (1) on
    dimension 1 (2 and 1)

The fix checks dimensions before creating ArrayBroadcast nodes during semantic analysis, catching both constant and variable array indices.

Changes:
- Modified ast_body_visitor.cpp to validate array dimensions before broadcasting in assignments
- Added test case array_section_mismatch_01.f90 to verify the fix

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -560,6 +560,7 @@ RUN(NAME functions_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME functions_50 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_51 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_52 LABELS llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME functions_53 LABELS llvm llvm_wasm llvm_wasm_emcc)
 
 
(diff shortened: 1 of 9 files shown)


certik:
Please don't format the source code, it obscures the actual changes.

Issue #9134, opened 2025-12-19
Register the equivalence_12.f90 test


certik:
> You need to register this test.
> 
> It's obvious in the final diff: https://github.com/lfortran/lfortran/commit/6c975448495e03e02da9f7f4d9582921ae91a246. 

 _Originally posted by @certik in [#9121](https://github.com/lfortran/lfortran/pull/9121/changes#r2633103617)_

opixdown:
done!

certik:
@opixdown thank you!

It turns out the test doesn't actually pass. Here is the diff that was committed: https://github.com/lfortran/lfortran/commit/6c975448495e03e02da9f7f4d9582921ae91a246.

@krystophny is the commit good --- what bug does it fix?

Let's figure out how to get `equivalence_12` passing.

ANAS727189:
I have reproduced the issue locally and the test actually fails with a runtime error. 
The asr representation currently is - 
```
(lf) anas@anas-mint:/lfortran$ src/bin/lfortran --show-asr integration_tests/equivalence_12.f90
(TranslationUnit
    (SymbolTable
        1
        {
            equivalence_12:
                (Program
                    (SymbolTable
                        2
                        {
                            dt19x:
                                (Variable
                                    2
                                    dt19x
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Array
                                        (Real 8)
                                        [((IntegerConstant 1 (Integer 4) Decimal)
                                        (IntegerConstant 2 (Integer 4) Decimal))
                                        ((IntegerConstant 1 (Integer 4) Decimal)
                                        (IntegerConstant 2 (Integer 4) Decimal))
                                        ((IntegerConstant 1 (Integer 4) Decimal)
                                        (IntegerConstant 4 (Integer 4) Decimal))]
                                        FixedSizeArray
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                    .false.
                                    .false.
                                    ()
                                    .false.
                                    .false.
                                ),
                            dt19xa:
                                (Variable
                                    2
                                    dt19xa
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Pointer
                                        (Array
                                            (Real 8)
                                            [(()
                                            ())
                                            (()
                                            ())
                                            (()
                                            ())]
                                            DescriptorArray
                                        )
                                    )
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
[...]

PR #9137, opened 2025-12-19
Register test equivalence_12.f90

opixdown:
closes #9134

PR #9141, opened 2025-12-19
Register equivalence_12.f90 integration test

shlok3640:
This PR registers the equivalence_12.f90 test case as requested in #9134.


certik:
This PR doesn't register it. This was already tried in https://github.com/lfortran/lfortran/pull/9137, see that PR how to do it. But the test doesn't pass.

PR #9149, opened 2025-12-21
chore: add assert check for struct size

### LINKED ISSUE #8778: Add an assert check for struct size in openmp

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/8776#issuecomment-3483087667_

adityajha-iitj:
Closes #8778

The assertion validates at compile time that the constant stays synchronized between openmp.cpp and LLVM's struct layout, and will require force updates to both implementations together if the struct size ever changes.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1899,6 +1899,7 @@ RUN(NAME string_86 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_88 LABELS llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays) #remove gfortran label, as functionality gave incorrect results
 RUN(NAME string_89 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME string_90 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
@@ -2084,6 +2085,7 @@ RUN(NAME class_83 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --rea
 RUN(NAME class_84 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME class_85 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME class_86 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME class_procedure_args_01 LABELS gfortran llvm)
 
@@ -2345,6 +2347,7 @@ RUN(NAME procedure_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME procedure_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRAFILES procedure_22_a.f90)
 RUN(NAME procedure_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME procedure_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME procedure_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 
 RUN(NAME allocated_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/class_87.f90
+++ b/integration_tests/class_87.f90
@@ -1,0 +1,62 @@
+module class_87_mod
+    implicit none
+
+    type, abstract :: toml_value
+    end type toml_value
+
+    type :: xx
+        integer :: v = 0
+    end type xx
+
+    type, extends(toml_value) :: toml_table
+        class(xx), allocatable :: data
+    end type toml_table
+
+contains
+
+    function cast_to_table(ptr) result(table)
+        class(toml_value), intent(in), target :: ptr
+        type(toml_table), pointer :: table
+
+        nullify(table)
+        select type (ptr)
+        type is (toml_table)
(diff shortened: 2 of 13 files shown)


adityajha-iitj:
Adding the fix in some time

adityajha-iitj:
The `STRUCT_VTABLE_SIZE=8` used in openmp.cpp makes it easy to forget updating both places if the LLVM struct layout changed. I moved it to asr_utils.h, a shared header both the pass layer and backend can safely use, so there's only one source. 
Now openmp.cpp uses the shared constant, and a compile-time assertion in llvm_utils.cpp ensures the value stays correct. 
If LLVM struct size changes then both the assertion and the constant defined in the shared header file should be changed to keep it in sync with openmp.cpp.

adityajha-iitj:
Ready for Review!

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -5,6 +5,10 @@
 #include <libasr/codegen/llvm_compat.h>
 #include <llvm/Support/raw_ostream.h>
 
+// STRUCT_VTABLE_SIZE used in openmp.cpp through asr_utils.h file must match the size of LLVM_STRUCT_SIZE
+static_assert(LCompilers::STRUCT_VTABLE_SIZE == 8, 
+    "STRUCT_VTABLE_SIZE must be 8 bytes (size of pointer-to-pointer on 64-bit systems)");
+
```
certik, line 15: Why is this check here? Shouldn't this rather go into `openmp.cpp`? It's there where we require the size to be 8, since other code in that file requires it, right?

adityajha-iitj: I am very sorry, i actually saw that the llvm struct size is being handled through this file (llvm_utils.cpp) so i kept the assert statement here, but i see it would be more logically accurate to keep this statement in openmp.cpp only.
Please give me a moment i will add the changes.

certik: After your change the `STRUCT_SIZE` is not defined twice as 8. Why not just declaring it once, but it's not used in `llvm_utils`, or is it?

certik: Ah I see --- you now define the constant twice with a different name. Ok.

src/libasr/pass/openmp.cpp
```diff
@@ -3224,7 +3223,6 @@ class ParallelRegionVisitor :
 
         int64_t compute_task_data_size(const ASR::symbol_t* task_data_struct_sym) {
             int64_t total_size = 0;
-            const int STRUCT_VTABLE_SIZE = 8;
```
certik, line 3227: I was thinking of adding an assert here that it's equal to 8, since the rest of the code here depends on it.

src/libasr/codegen/llvm_utils.cpp
adityajha-iitj: If the LLVM struct size changes, the programmer updates it in `llvm_utils.cpp`, due to the assertion in this file checking against, the programmer is forced to update the constant in `asr_utils.h`. Since `openmp.cpp` uses this constant from `asr_utils.h`, the change automatically propagates, that's why i don't think we need another assertion here.

src/libasr/pass/openmp.cpp
adityajha-iitj: right now, this `STRUCT_VTABLE_SIZE` is used only at this `compute_task_data_size` function in openmp so i have directly used the variable from `asr_utils.h` to keep this in sync with changes in llvm struct size, in future if this gets used in this file elsewhere or any other file too, this can be used the same way.

Please let me know if this is fine

certik:
I need to understand where all these sizes are coming from:

* is openmp assuming length 8? Your PR makes it just use a global variable, and I currently don't know if this will work for any size, or if the size is required to be 8. So it would need to be investigated.
* How is the `STRUCT_SIZE` determined to be 8? What does it mean?

adityajha-iitj:
According to my understanding and the discussion in PR #8776 with new classes LFortran's polymorphic types (class types) have a vtable pointer as their first member. On 64-bit systems, this vtable pointer is 8 bytes.
The function `compute_task_data_size` in `openmp.cpp`needs to account for this 8-byte vtable pointer when calculating struct sizes for OpenMP task data structures.

The reason why it is exactly 8 bytes is maybe it assumes 64-bit architecture where pointers are 8 bytes.

To prevent future bugs where the LLVM vtable pointer size changes but the OpenMP code doesn't get updated:
1. In `llvm_utils.cpp` file, I added STRUCT_SIZE = 8 (a constant representing LLVM vptr type size) with an assertion that verifies that `STRUCT_VTABLE_SIZE == STRUCT_SIZE`
2. In `asr_utils.h` global constant STRUCT_VTABLE_SIZE = 8 is defined
3. In `openmp.cpp` the code now uses the global variable to keep it with sync

certik:
I am confused what STRUCT_SIZE means in llvm_utils.cpp. What does that represent exactly? How is it different to STRUCT_VTABLE_SIZE?

adityajha-iitj:
STRUCT_SIZE in llvm_utils.cpp is same what STRUCT_VTABLE_SIZE is, both of them stores this value 8 in them and they mean the same, in llvm backend i have named it STRUCT_SIZE cause i have to add the check comparing this value updated in llvm backend is equal to the global variable
I am sorry if it is too vague or making confusions to understand, if you want i will rename this to something LLVM_STRUCT_VTABLE_SIZE which makes more sense

certik:
Don't merge main, it's straining our CI and it's not necessary to test it at this point. Once we agree how to proceed, we'll finish the PR and only then rebase.

certik:
> STRUCT_SIZE in llvm_utils.cpp is same what STRUCT_VTABLE_SIZE is, both of them stores this value 8 in them and they mean the same, in llvm backend i have named it STRUCT_SIZE cause i have to add the check comparing this value updated in llvm backend is equal to the global variable
I am sorry if it is too vague or making confusions to understand, if you want i will rename this to something LLVM_STRUCT_VTABLE_SIZE which makes more sense

So if they mean the same, as you say, what is the reason to have two different global variables for that? We should have just one.

adityajha-iitj:
I had kept it twice cause i wanted to have one variable right at the source of change in the backend, so if there are changes, this one must be updated cause this is right at the source and the assertion here makes it to update the global variable in asr_utils such that other usages of this value remain in sync

certik:
> I had kept it twice cause i wanted to have one variable right at the source of change in the backend, so if there are changes, this one must be updated cause this is right at the source and the assertion here makes it to update the global variable in asr_utils such that other usages of this value remain in sync

I still don't understand how this solution actually helps. Your `STRUCT_SIZE` variable is not used anywhere, so we will not be forced to update it. Also the name of it doesn't make sense to me, the number 8 represents the size of the vptr pointer, not any struct.

Rather, we should use one variable, and we should check in the LLVM backend that it has the right size.

I just implemented it here: https://github.com/lfortran/lfortran/pull/9187.

adityajha-iitj:
No problem, now i understand what was needed, I'll make sure next time

PR #9150, opened 2025-12-21
solve issue #9147

### LINKED ISSUE #9147: Handle reallocation of string array to arraysection with single element
MRE:
```fortran
module m_dep
  type :: downloader_t
     character(len=:), allocatable :: name
    end type downloader_t
end module m_dep


program demo
  use m_dep
  type(downloader_t), allocatable :: dl(:)
    integer :: i
        i = 1
  allocate(dl(2))
    dl(1)%name = "first"
    dl(2)%name = "second"
    dl = dl(1:i)
    print *, allocated(dl(1)%name)
end program demo
```

Outputs:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran a.f90 && ./a.out
 T
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 --realloc-lhs-arrays
F
```

Atharv1405:
fixes #9147 
*MRE*
```
module m_dep
  type :: downloader_t
     character(len=:), allocatable :: name
    end type downloader_t
end module m_dep


program demo
  use m_dep
  type(downloader_t), allocatable :: dl(:)
    integer :: i
        i = 1
  allocate(dl(2))
    dl(1)%name = "first"
    dl(2)%name = "second"
    dl = dl(1:i)
    print *, allocated(dl(1)%name)
end program demo
```

issue
earlier at runtime the string was being  deallocated before being copied which causes missing data hence F was the output

*correction*
i have introduced a temporary when same array appears in lhs and rhs even if allocatable
and in array_op added logic for correct deepcopy of derived types
added a test for the same as well

*After correction*
```
(lf) /lfortran$ gfortran t1.f90 && ./t1.out
T
(lf) /lfortran$ ./src/bin/lfortran t1.f90 --realloc-lhs-arrays
T
```

If there is any discrepancy in the fix or any improvement needed please let me know

### DIFF
--- a/tests/reference/pass_array_op-modules_43-6930881.json
+++ b/tests/reference/pass_array_op-modules_43-6930881.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "pass_array_op-modules_43-6930881.stdout",
-    "stdout_hash": "5409788fab5988407b1fa2e569270aeeb029d05b53d37e2be1c97ccd",
+    "stdout_hash": "e9a18f609c0f92799e012cd627b0d06e3c26ceb712d92f69096abe2d",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/pass_array_op-modules_43-6930881.stdout
+++ b/tests/reference/pass_array_op-modules_43-6930881.stdout
@@ -98,48 +98,6 @@
                                     (SymbolTable
                                         4
                                         {
-                                            __1__libasr_index_:
-                                                (Variable
-                                                    4
-                                                    __1__libasr_index_
-                                                    []
-                                                    Local
-                                                    ()
-                                                    ()
-                                                    Default
-                                                    (Integer 4)
-                                                    ()
-                                                    Source
-                                                    Public
-                                                    Required
-                                                    .false.
-                                                    .false.
-                                                    .false.
-                                                    ()
-                                                    .false.
-                                                    .false.
-                                                ),
-                                            __1_t:
-                                                (Variable
-                                                    4
-                                                    __1_t
-                                                    []
-                                                    Local
-                                                    ()
-                                                    ()
-                                                    Default
-                                                    (Integer 4)
-                                                    ()
-                                                    Source
-                                                    Public
-                                                    Required
-                                                    .false.
-                                                    .false.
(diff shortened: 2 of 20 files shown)


jinangshah21:
Use `--dump-all-passes-fortran` to print simplified codes after each ASR Pass. Without this PR code after `array_struct_temporary` is:
```fortran
__libasr_created__array_section_pointer_ => dl(1:i)
dl = __libasr_created__array_section_pointer_
```
This PR does:
```fortran
__libasr_created__array_section_pointer_ => dl(1:i)
deallocate(__libasr_created__assignment_value_)
allocate(__libasr_created__assignment_value_(size(__libasr_created__array_section_pointer_)))
__libasr_created__assignment_value_ = __libasr_created__array_section_pointer_
dl = __libasr_created__assignment_value_
```

But here instead of this array section pointer we need to have allocatable array as temporary, so we don't use 2 temporaries as done in this PR. We need to fix this in `array_struct_temporary` pass.

Atharv1405:
ok i will work on it

Atharv1405:
@jinangshah21 i tried correcting that issue but the extra temporaries is being created by helper functions and changing them causes ICE at other places .although its structural not ideal but will it cause any error ?

integration_tests/reallocation_of_string_array.f90
```diff
@@ -0,0 +1,20 @@
+! RUN: %lfortran %s --realloc-lhs-arrays
```
certik: This test is not registered.

certik:
I think this PR is not ready to be merged, so I marked it as draft for now.

assem2002:
@Atharv1405  I think this PR could be a valid fix. Go ahead and handle the reference files -- Use llvm 11

Atharv1405:
@assem2002 thanks for assurance , i will work with reference files as suggested

HarshitaKalani:
Can you please try these commands and then push the updated references. I hope it should work then
```
git clean -dfx
./build0.sh && ./build1.sh
./run_tests -u
./integration_tests/run_tests.py
```

Atharv1405:
i think i will have to raise a new PR this one seems to got messed up with merge conflicts .sorry for the trouble

assem2002:
No problem. Go for it.

PR #9151, opened 2025-12-21
Fix cascading dummy argument error after invalid assumed-size dimension

### LINKED ISSUE #9094: Suppress spurious 'Dummy argument not defined' error after dimension errors
When a dimension error occurs (e.g., `*` not in last position), `SemanticAbort` prevents variable registration, causing spurious "Dummy argument not defined" errors.

**Current behavior:**
```fortran
subroutine foo(a)
    real :: a(*, 10)  ! error: star not in last dim
end subroutine
```
Produces TWO errors:
1. `Assumed-size '*' is only permitted in the last dimension` (correct)
2. `Dummy argument 'a' not defined` (spurious - `a` IS defined, just incorrectly)

**Location:** `src/lfortran/semantics/ast_common_visitor.h:2137`
```cpp
diag.add(diag::Diagnostic(
    "Assumed-size '*' is only permitted in the last dimension", ...));
throw SemanticAbort();  // <-- causes variable to not be registered
```

**Fix needed:** Continue processing after the error so the variable gets registered (even if invalid), preventing cascading errors. Simply removing the throw causes ICE - need to ensure dimension processing handles the error gracefully.

amritamishra01:
Fixes #9094

This PR fixes a cascading semantic error where an invalid assumed-size * in a non-last dimension triggered a SemanticAbort. This prevented the dummy argument from being registered in the symbol table, producing a secondary spurious error: "Dummy argument not defined".

Changes:

Removed the early SemanticAbort in ast_common_visitor.h.

Implemented safe error recovery by inserting a neutral explicit dimension (1:1).

This preserves the array rank and allows symbol table registration to complete, ensuring the user sees only the relevant dimension error.

A regression test has been added in tests/errors/repro_9094.f90

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2089,6 +2089,7 @@ RUN(NAME class_84 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --rea
 RUN(NAME class_85 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME class_86 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_88 LABELS llvm)
 
 RUN(NAME class_procedure_args_01 LABELS gfortran llvm)
 
--- a/integration_tests/class_88.f90
+++ b/integration_tests/class_88.f90
@@ -1,0 +1,47 @@
+module class_88_m
+  type t
+    integer :: n = 0
+   contains
+    procedure :: tbp => f
+  end type
+ contains
+  function f(this)
+    class(t), pointer, intent(in) :: this
+    integer, pointer :: f
+    f => this%n
+  end function f
+end module
+
+program test
+  use class_88_m
+  implicit none
+  type(t), target :: xt
+  class(t), pointer :: xp
+  integer, pointer :: result_ptr
+
+  ! Initialize
+  xt%n = 42
+  xp => xt
+
+  ! Test type-bound procedure call with type variable
+  result_ptr => xt%tbp()
+  print *, result_ptr, xt%tbp()
+  if (result_ptr /= 42) error stop "Failed: xt%tbp() should return 42"
+
+  ! Test type-bound procedure call with class pointer
+  result_ptr => xp%tbp()
+  print *, result_ptr, xp%tbp()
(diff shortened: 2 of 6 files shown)


amritamishra01:
Hi @certik ,
All CI checks are passing now! 
I have updated the reference files for continue_compilation_1.f90 because the fix successfully removed the spurious "Dummy argument not defined" errors from that test as well.
Ready for review.

tests/errors/repro_9094.f90.out
```diff
@@ -0,0 +1,9 @@
+[0;31;1msemantic error[0;0m[0;1m: Assumed-size '*' is only permitted in the last dimension[0;0m
+ [0;34;1m-->[0;0m tests/errors/repro_9094.f90:2:15
+  [0;34;1m|[0;0m
+[0;34;1m2 |[0;0m     real :: a(*, 10) ! error: star not in last dim
+  [0;34;1m|[0;0m               [0;31;1m^ [0;0m
+
+
+[0;1mNote[0;0m: Please report unclear, confusing or incorrect messages as bugs at
+https://github.com/lfortran/lfortran/issues.
```
certik: I think these two files should not be there?

certik:
Otherwise I think this PR is fine.

amritamishra01:
@certik Sir, thanks for pointing that out ........you’re right. The .out file was unintended. I’ve removed it and pushed the update.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2128,14 +2128,37 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                // Create the dimension object
+                ASR::dimension_t dim_dummy;
+                dim_dummy.loc = m_dim[i].loc;
+                dim_dummy.m_start = one;
+                dim_dummy.m_length = one; // start=1, length=1 implies end=1
+                
+                dims.push_back(al, dim_dummy);
+
+                // 3. Skip complex processing
+                continue;
```
HarshitaKalani: I think these comments are a bit redundant since the code is already self-explanatory. We can remove them to reduce noise and keep the code cleaner.

tests/errors/repro_9094.f90
HarshitaKalani, line 1: Can we remove this test and add it in continue_compilation_*.f90?

tests/reference/asr-continue_compilation_1-04b6d40.stderr
HarshitaKalani, line 1: Is this expected?

HarshitaKalani: I see https://github.com/lfortran/lfortran/pull/9151#issuecomment-3678729523
Should we remove those lines from continue_compilation_1.f90 then if they don't throw any error?

PR #9158, opened 2025-12-21
fix: finalize nested allocatables before deallocate

### LINKED ISSUE #9097: Complete allocatable INTENT(OUT) semantics to enable strict double-allocation checks

krystophny:
## Summary

Fixes #9097 - Nested allocatable components in derived-type arrays were not being freed before the outer array was deallocated on intent(out) re-entry.

## What changed

- Add `finalize_before_deallocate()` in `llvm_utils.h` to free nested allocatable components
- Call it from `visit_Deallocate()` in `asr_to_llvm.cpp` before freeing the main allocation
- Add minimal integration test `intent_out_array_01.f90`

## Verification

```bash
# Without fix: 192 bytes definitely lost
# With fix: All heap blocks freed
valgrind --leak-check=full ./test_intent_out
```

## Test plan

- [ ] CI passes
- [ ] Valgrind shows no leaks on intent_out_array_01

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2844,3 +2844,4 @@ RUN(NAME test_ord LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 # test for polymorphic select type
 RUN(NAME polymorphic_select_type_01 LABELS gfortran llvm)
 RUN(NAME defined_op_match_01 LABELS gfortran llvm)
+RUN(NAME intent_out_array_01 LABELS gfortran llvm)
--- a/integration_tests/intent_out_array_01.f90
+++ b/integration_tests/intent_out_array_01.f90
@@ -1,0 +1,27 @@
+! Test for intent(out) allocatable array of derived type with nested allocatables
+! Issue #9097: nested allocatables must be deallocated before outer array on re-entry
+program intent_out_array_01
+    implicit none
+    type :: node_t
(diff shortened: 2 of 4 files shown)


krystophny:
This is just the core part of PR #9108 and doesn't switch on the double allocation check yet (this led to a massive PR). The integration test passes on main but produces a memory leak which can be checked manually with valgrind. This leak is fixed here. @certik @assem2002 if you agree, we could split PR #9108 which is fixing many things into smaller PRs starting with this one here?

assem2002:
Split PRs are the best, so let's go for it.

PR #9159, opened 2025-12-22
allowed operator overloading to handle unaalocated arrays

### LINKED ISSUE #9082: Handle unallocated arrays in operator overloaded functions
MRE:
```fortran 
module m_mod
    implicit none

    interface operator(.in.)
        module procedure in_array
    end interface
contains

    logical function in_array(str, arr) result(res)
        character(len=*), intent(in) :: str
        character(len=*), intent(in) :: arr(:)
        res = .false.
    end function in_array

end module m_mod

program demo_in_operator
    use m_mod
    implicit none
    character(len=100), allocatable :: modules_provided(:)
    s = 'mod_a'
    if (s .in. modules_provided) error stop
end program demo_in_operator
```

LFortran Error:
```console
ase) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 --realloc-lhs-arrays
At 25:16 of file a.f90
Runtime error: Argument 2 of subroutine in_array_integer____1 is unallocated.
```

PrakharGoyal28:
This PR fixes #9082

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -266,9 +266,9 @@ time_section "🧪 Testing FPM" '
   git clone https://github.com/jinangshah21/fpm.git
   cd fpm
   export PATH="$(pwd)/../src/bin:$PATH"
-  git checkout lf-14
+  git checkout lf-15
   micromamba install -c conda-forge fpm
-  git checkout ed292d67826c17e91503941ddd3d2ad7fa5d2139
+  git checkout e709edb424807d8d0b565169fea494c1f8f02471
   fpm --compiler=$FC build --flag "--cpp --realloc-lhs-arrays --use-loop-variable-after-loop"
   fpm --compiler=$FC test --flag "--cpp --realloc-lhs-arrays --use-loop-variable-after-loop"
   print_success "Done with FPM"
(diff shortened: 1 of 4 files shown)


certik:

assem2002:
On it

assem2002:
Here is my intake on the issue.
We should actually keep the reported runtime error as shown in the MRE in the linked [issue](https://github.com/lfortran/lfortran/issues/9082).
From Fortran standard perspective  (Fortran standards 5.4.10) -> **"An unallocated allocatable variable shall not be referenced or defined"**

***

I think that's not a valid Fortran code, hence no bug.

_**[option 1]**_ The bug is reported as an fpm bug. We can tackle it by straightening that line in the fpm code , or turn off the checks.

gfortran & ifort raise runtime error (with `--check=all`) when an unallocated deferred-size array is passed to `dimension(*)` or `dimension(10)`, but doesn't raise an error when passed to `dimension(:)`. I think it's nothing more than a technical constraint in the compiler.

I think there's no reason to not report that error.
_**[option 2]**_ but still the door is open to not report any errors when a **descriptorarray** is passed to a **descriptorarray** (and I think there shouldn't be undefined behavior from our side if we do that -- size=0).

@certik I think option 1 is valid. What do you think?

certik:
@assem2002 excellent catch. I replied in https://github.com/lfortran/lfortran/issues/9082#issuecomment-3684642144.

PR #9160, opened 2025-12-22
fix: cast assumed rank arrays to descriptor before comparing

HarshilShah1804:

With this PR, we test assumed rank arrays for probably all kinds of operations. Now, we can find some codes which are using assumed rank arrays and test them to handle the corner cases which, if any, may have been missed now.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -761,6 +761,7 @@ RUN(NAME select_rank_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 
 RUN(NAME global_allocatable_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME global_allocatable_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/select_rank_08.f90
+++ b/integration_tests/select_rank_08.f90
@@ -1,0 +1,28 @@
+program select_rank_08
+    implicit none
+
+    integer :: a(2) = [4, 2]
+    integer :: b(2, 3) = reshape([1, 2, 3, 4, 5, 6], [2, 3])
+    
+    call check(a)
+    call check(b)
+contains
+
+    subroutine check(x)
+        integer, intent(inout) :: x(..)
+        integer, save :: a(2) = [5, 3]
+        integer, save :: b(2, 3) = reshape([6, 5, 4, 3, 2, 1], [2, 3])
+    
+        select rank(x)
+        rank (1)
+            x = a
+            print *, x
+            if (any(x /= a)) error stop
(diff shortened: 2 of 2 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -11695,6 +11695,22 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         ASR::expr_t *left = ASRUtils::EXPR(tmp);
         this->visit_expr(*x.m_right);
         ASR::expr_t *right = ASRUtils::EXPR(tmp);
+        if (ASRUtils::is_assumed_rank_array(ASRUtils::expr_type(left))) {
+            std::string array_name = ASRUtils::symbol_name(ASR::down_cast<ASR::Var_t>(left)->m_v);
+            if (assumed_rank_arrays.find(array_name) == assumed_rank_arrays.end()) {
+                diag.add(Diagnostic("Comparison operations are not allowed on assumed-rank arrays ('" + array_name + "')",
+                    Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
+                throw SemanticAbort();
```
certik, line 11703: Let's add a test for this error into continue compilation.

integration_tests/CMakeLists.txt
```diff
@@ -761,6 +761,7 @@ RUN(NAME select_rank_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
```
certik, line 764: Why is `--realloc-lhs-arrays` needed? I don't see any allocatable arrays in the code.

HarshilShah1804: We create a temporary after the `array_struct_temporary` named `_libasr_integer_compare_left` which is an allocatable array. In other cases we would have dimensions available for the allocation, so we insert an allocate statent ourselves but for assumed rank arrays we didn't had the dimensions so we don't insert the allocate statement. So, I had used the `--realloc-lhs-arrays`. 

But I think your concern is right. I'll check for the implementation on how we handle deferred shape arrays like `a(:)`. Mostly it would be involving using the runtime dimensions using the `size()` function. 

What do you suggest shall we meege this now and correct for the inserting the allocate statement using runtime size later. Or shall we fix this now only and then merge this. Please give your suggestions on how to move forward with this. I believe it is a small fix and not much difficult to handle.

src/lfortran/semantics/ast_common_visitor.h
HarshilShah1804: Oh right. I will add it in some time.

integration_tests/CMakeLists.txt
certik: We can merge this, and fix later. Go ahead and open up issues. Let's also run full tests first.

HarshilShah1804: Perfect.

HarshilShah1804: I have opened #9165 for that.

src/lfortran/semantics/ast_common_visitor.h
HarshilShah1804: Since, half of the CI checks have completed and others are completed about 50 %, is it okay if I just add this in a subsequent PR.

HarshilShah1804:
I have the fix for the `--realloc-lhs-arrays` ready on my local branch. I will push it and open the PR once this gets merged. I will also add the error test in that PR.

Issue #8870, opened 2025-11-17
Bug : Calling `allocate` with multiple levels of `Type` nesting fails


assem2002:
```.f90
program pp
    type t1
        integer, allocatable :: ll
    end type

    type t2
        type(t1),allocatable :: arr(:)
    end type t2

    type t3
        type(t2) ,allocatable :: z(:)
    end type

    call ss

    contains 
    subroutine ss
        type(t3), allocatable :: t2
        allocate(t2)
        allocate(t2%z(10))
        allocate(t2%z(1)%arr(5))
    end subroutine

end program
```

```terminal
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ ../../../LFortranUpStream/lfortran/src/bin/lfortran tt7.f90 
ASR verify pass error: ExternalSymbol::m_external '1_t2' is not in a module or struct type, owner: pp
 --> tt7.f90:7:33
  |
7 |         type(t1),allocatable :: arr(:)
  |                                 ^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues. 
```

HarshitaKalani:
This does not fail anymore. Let's add a test to close this.
```
% lfortran x.f90 
% echo $?
0
% gfortran x.f90 && ./a.out
```

PR #9161, opened 2025-12-22
updated utils to solve issue(#8870)

Atharv1405:
solves issue #8870 
*MRE*
```
program pp
    type t1
        integer, allocatable :: ll
    end type

    type t2
        type(t1),allocatable :: arr(:)
    end type t2

    type t3
        type(t2) ,allocatable :: z(:)
    end type

    call ss

    contains 
    subroutine ss
        type(t3), allocatable :: t2
        allocate(t2)
        allocate(t2%z(10))
        allocate(t2%z(1)%arr(5))
    end subroutine

end program
```

*correction*
added a guard in `import_struct_instance_member` to prevent formation of any external symbol for derived type

*after correction*
MRE didnt produced any error

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1917,6 +1917,8 @@ RUN(NAME nested_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME nested_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+
 
 RUN(NAME nested_vars1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)
 RUN(NAME nested_vars2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)


Atharv1405:
@HarshitaKalani sure , in the integration tests ?

HarshitaKalani:
Yes, add it in integration_tests and register it in CMakeLists.txt

integration_tests/nested_19.f90
```diff
@@ -0,0 +1,24 @@
+program pp
```
HarshitaKalani:
```suggestion
program nested_19
```

integration_tests/nested_19.f90
```diff
@@ -0,0 +1,24 @@
...
+
+    contains 
+    subroutine ss
+        type(t3), allocatable :: t2
+        allocate(t2)
+        allocate(t2%z(10))
+        allocate(t2%z(1)%arr(5))
+    end subroutine
+
+end program
```
HarshitaKalani:
```suggestion
end program nested_19
```

Atharv1405:
thanks for correction

integration_tests/nested_19.f90
```diff
@@ -0,0 +1,24 @@
...
+    type t3
+        type(t2) ,allocatable :: z(:)
+    end type
+
+    call ss
+
+    contains 
+    subroutine ss
+        type(t3), allocatable :: t2
+        allocate(t2)
```
HarshitaKalani, line 19: Can you also add error stop conditions to make the test more robust. For example:
```suggestion
        allocate(t2)
        if (.not. allocated(t2)) error stop 
```

integration_tests/nested_19.f90
```diff
@@ -0,0 +1,24 @@
...
+        allocate(t2%z(10))
+        allocate(t2%z(1)%arr(5))
```
HarshitaKalani, line 22: Do the same for these.

Atharv1405: i tried on these two but i am getting ICE . i think `.not.allocated()` does not work on them (probably another bug)

Atharv1405:
i think `allocated()` does not work on allocatable components like arr

Atharv1405:
i have done the rest of the changes as suggested

HarshitaKalani:
You need to register the test in integration_tests/CMakeLists.txt

Atharv1405:
@HarshitaKalani i have added test in cmake list can you please confirm if its the right way to register a test ?

integration_tests/nested_19.f90
HarshitaKalani: Can you please open an issue for it?

HarshitaKalani (approved):
Looks good to me.

certik:
Let's run full tests. Looks good.

certik (approved):
Looks good. Can you please open up an issue for this: https://github.com/lfortran/lfortran/pull/9161/changes#r2642386294 and link it from the comment?

Issue #8889, opened 2025-11-23
Code generation error from calling passed in subroutine


w6ws:
Calling a passed in subroutine two slightly different ways (first element vs array) causes a `code generation error`:
```
      SUBROUTINE callsub2 (Y, idimy, W3, xub)
      implicit none
      external xub
      integer :: idimy
      integer       W3(*),Y(IDIMY,*)
         CALL xub (Y(1,1))
         CALL xub (W3)
      RETURN
      END
```
Compiling with --implicit-interface produces:
```
$ lfortran -c --implicit-interface callsub2.f90 | more
; ModuleID = 'LFortran'
source_filename = "LFortran"

code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
%array* %w3
 i32*  call void %xub(%array* %w3)



Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
%array = type { i32*, i32, %dimension_descriptor*, i1, i32 }
%dimension_descriptor = type { i32, i32, i32 }

@0 = private unnamed_addr constant [2 x i8] c"y\00", align 1
@1 = private unnamed_addr constant [149 x i8] c"At 6:20 of file callsub2.f90\0AR
untime error: Array '%s' index out of bounds.\0A\0ATried to access index %d of d
imension %d, but valid range is %d to %d.\0A\00", align 1
@2 = private unnamed_addr constant [2 x i8] c"y\00", align 1
@3 = private unnamed_addr constant [149 x i8] c"At 6:20 of file callsub2.f90\0AR
untime error: Array '%s' index out of bounds.\0A\0ATried to access index %d of d
imension %d, but valid range is %d to %d.\0A\00", align 1

define void @callsub2(%array* %y, i32* %idimy, %array* %w3, void (i32*)* %xub) {
.entry:
  %0 = getelementptr %array, %array* %y, i32 0, i32 2
  %1 = load %dimension_descriptor*, %dimension_descriptor** %0
  %2 = getelementptr inbounds %dimension_descriptor, %dimension_descriptor* %1, 
i32 0
  %3 = getelementptr %dimension_descriptor, %dimension_descriptor* %2, i32 0, i3
2 1
  %4 = load i32, i32* %3
  %5 = getelementptr %dimension_descriptor, %dimension_descriptor* %2, i32 0, i3
2 2
  %6 = load i32, i32* %5
  %7 = sub i32 1, %4
[...]

Issue #6543, opened 2025-03-09
segv when calling a passed-in procedure


w6ws:
$ cat callsub.f90
subroutine callsub (xub, i)
  call xub (i)
end subroutine
$
$ lfortran -c --implicit-typing callsub.f90
Segmentation fault (core dumped)
$

If I add _both_ --implicit-typing and --implicit-interface it will successfully compile.  Issue is that the compiler should never segv.

$ lfortran --version
LFortran version: 0.42.0
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$

harperjf:
Giving @w6ws's subroutine callsub an actual subroutine argument makes lfortran --std=f23 compile and run this program:
```
! lfortran #6543 dummysub.f90
program callingsub
  call callsub(square,2)
contains
    
  subroutine callsub (xub, i)
    call xub (i)
  end subroutine callsub

  subroutine square(i)
    print "(I0)",i**2
  end subroutine square

end program callingsub
````
Output from gfortran, ifx, ifort or lfortran 0.47.0 --std-F23
```
4
```
Without --std=f23, lfortran gave this polite compile-time output:
```
semantic error: Dummy argument 'xub' not defined
 --> dummysub.f90:6:3 - 8:24
  |
6 |      subroutine callsub (xub, i)
  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
  |
8 |      end subroutine callsub
  | ...^^^^^^^^^^^^^^^^^^^^^^^^ 
```

PR #9162, opened 2025-12-22
fix: prevent segfault when calling non-procedure variable without --implicit-interface

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6543
closes: https://github.com/lfortran/lfortran/issues/8889

### DIFF
--- a/tests/reference/asr-implicit_call_01-e42aa0d.json
+++ b/tests/reference/asr-implicit_call_01-e42aa0d.json
@@ -1,0 +1,13 @@
+{
+    "basename": "asr-implicit_call_01-e42aa0d",
+    "cmd": "lfortran --show-asr --implicit-typing --no-color {infile} -o {outfile}",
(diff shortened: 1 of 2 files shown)


tests/errors/implicit_call_01.f90
```diff
@@ -0,0 +1,5 @@
+subroutine callsub (xub, i)
+    ! xub is a variable (implicitly typed as real),
+    ! trying to call it should fail without --implicit-interface
+    call xub (i)
+end subroutine
```
certik, line 5: Let's put this into continue compilation tests.

HarshitaKalani: I need to set `asr_implicit_typing = true` for this test, that's why I've added it separately.

certik (approved):
I see. Looks good then.

certik:
Let's make sure full tests run.

PR #9163, opened 2025-12-22
Enable double-allocation runtime check for allocatable INTENT(OUT)

krystophny:
Now that allocatable INTENT(OUT) dummies are deallocated at procedure entry on main (PR #9158 / issue #9097), re-enable the allocate-already-allocated runtime error check for INTENT(OUT) allocatables.

This is a draft PR intended to see what breaks in CI.

### DIFF
--- a/tests/reference/llvm-allocate_03-495d621.json
+++ b/tests/reference/llvm-allocate_03-495d621.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-allocate_03-495d621.stdout",
-    "stdout_hash": "a6d94939e549e30aa404fd127f300cba774e5c73f5bf2b6758352a1e",
+    "stdout_hash": "0b1a494e0d8e747a0ee55402e2168d041ca4f32a15447422c4fb0e28",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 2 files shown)


krystophny:
@certik @HarshitaKalani this passes quick checks now with latest merged fixes. Can you activate full CI run to see where thirdparty tests break? Then we can subsequently replace piece by piece here by merging small PRs and rebasing on main every time, until the full test suite passes with the stricter checks enabled, and we replaced the bulk of this large PR (it lived on #9108 before, but there we do it the opposite, we have just the single stricter check commit there and see when it passes)

HarshitaKalani:
Thanks @krystophny , the exhaustive checks will run once the conflicts are resolved. Could you please do that? Thank you!

src/libasr/compiler_tester/tester.py
```diff
@@ -555,8 +555,10 @@ def tester_main(compiler, single_test, is_lcompilers_executable_installed=False)
     global no_color
     no_color = args.no_color
 
-    # Remove all old test references while updating
-    if update_reference:
+    # While updating references, only wipe the whole reference directory if the
+    # user is updating the entire suite (no -t filters). For a targeted update
+    # (via -t), deleting everything is unexpected and breaks unrelated tests.
+    if update_reference and not specific_tests:
         log.debug("REMOVE: old test references")
```
certik, line 562: Good point

certik (approved):
Excellent, I think this is great.

PR #9164, opened 2025-12-22
fix: deallocate allocatable struct-member actuals for intent(out)

krystophny:
Fixes a missing implicit-deallocate insertion when an allocatable struct component is passed to an allocatable INTENT(OUT) dummy, which can surface as failures in third-party stdlib logger tests under stricter allocation checks.

Related: #9108

### DIFF
--- a/tests/reference/asr-modules_29-dc71c55.json
+++ b/tests/reference/asr-modules_29-dc71c55.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-modules_29-dc71c55.stdout",
-    "stdout_hash": "d9a6ff0cf55bdf9c52dbf53afa894d4319d48bd59e738ea21b6878b3",
+    "stdout_hash": "652c496f5adc797204edcb432d61d8e2fd18729af483178040154d5a",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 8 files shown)


krystophny:
The `modules_29` ASR reference changed because this PR intentionally changes ASR generation for INTENT(OUT) allocatables.

Specifically: when an allocatable derived-type component (a `StructInstanceMember`, e.g. `self%dependency`) is passed as an actual argument to an allocatable `INTENT(OUT)` dummy, we now emit an `ImplicitDeallocate` for that component before the call (same as we already do for plain allocatable variables).

In `modules_29`, the call to `new_dependencies(deps, ...)` is made with `deps` coming from `self%dependency`, so the `ImplicitDeallocate` list now includes that `StructInstanceMember` (in addition to `error`).

This is expected and is why the `asr-modules_29-...` reference output needed updating.

certik:
I think the integration tests do not run with this PR.

krystophny:
> I think the integration tests do not run with this PR.

Thanks @certik , should be fixed! I usually leave draft status until CI is passing to save you review time.

certik:
The failure is:
```
FOUND:    fpm_scratch_CC/app/main.f90
FOUND:    fpm_scratch_CC/test/check.f90
FOUND:    fpm_scratch_CC/src/fpm_scratch_CC.f90
new TEST TALLY=TTTTTTTTTTT
PASSED: all 11 tests passed 
<ERROR> Execution for object " fpm-test " returned exit code  139
STOP 139
```
Which is caused by this PR.

certik:
The CI failed on:
```
The following tests FAILED:
	1183 - derived_types_87_STD_F23 (Subprocess aborted)
```
Not sure if it is related. I restarted it.

certik:
@assem2002, @jinangshah21, do you mind reviewing it please? I think it might be ok, but I would need to spend more time on it.

jinangshah21:
I think this is fine!

PR #9166, opened 2025-12-22
test: add and register test

### LINKED ISSUE #9045: Bug : call to intrinsic `allocated` with wrong output

HarshilShah1804:
Closes #9045

### DIFF
--- a/integration_tests/allocated_05.f90
+++ b/integration_tests/allocated_05.f90
@@ -1,0 +1,21 @@
+module node_mod
+
+  type :: dependency_config_t
+      integer, allocatable :: arr(:)
+  end type dependency_config_t
+  type, extends(dependency_config_t) :: dependency_node_t
+  end type dependency_node_t
+
+end module node_mod
+
+
+program demo
+  use node_mod
+  implicit none
+
+  type(dependency_node_t), allocatable :: a, b
+  allocate(a, b)
+  b = a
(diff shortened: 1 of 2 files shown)


integration_tests/allocated_05.f90
```diff
@@ -0,0 +1,21 @@
...
+
+program demo
+  use node_mod
+  implicit none
+
+  type(dependency_node_t), allocatable :: a, b
+  allocate(a, b)
+  b = a
+  print *, allocated(b%arr)
+  if (allocated(b%arr)) error stop
```
certik, line 20: I am a little confused here --- I would think that `allocate(a)` allocates the struct, but not `arr`, but here you test `arr` is allocated, which is not what I would expect as a user --- but perhaps it is correct.

certik: Ah I see --- this tests that it is *not* allocated. So all good.

PR #9168, opened 2025-12-22
fix: string array initialization padding

### LINKED ISSUE #9167: Bad initialization
This program was in bug report #3786 when it would not compile. Now it does but the output is wrong.
```
program initialtest
  implicit none
  character:: c(6)*2 = 'a'
print *, c
end program initialtest
```
Output from lfortran 0.58.0 --std=f23 (wrong)
```
 aa    aa    aa
```
Output from gfortran, ifx or AMD flang (right):
```
 a a a a a a
```

opixdown:
## **Issue Fixed**
fixes #9167  Bad initialization
## **Summary**

Previously the scalar character array initialization generated incorrect ASR:
`["aa", "aa", "aa", "", "", ""]`
After the pr the ASR:
`["a ", "a ", "a ", "a ", "a ", "a "]`

## **Reference Code**
```
program initialtest
  implicit none
  character:: c(6)*2 = 'a'
print *, c
end program initialtest
```

```
 aa    aa    aa
```

```
a     a     a     a     a     a 
  
```

### DIFF
--- a/integration_tests/string_90.f90
+++ b/integration_tests/string_90.f90
@@ -3,10 +3,9 @@ program string_test
     character(len=2) :: c(6) = 'a'
     
     !  c(1) should be 'a ' not 'a'
-    if (len(c(1)) /= 2) then
-        print *, "Error: length mismatch"
-        stop 1
-    end if
+    if (len(c(1)) /= 2) error stop
+    if ((c(1)(1:1) /= "a")) error stop
+    if ((c(1)(2:2) /= " ")) error stop
     
     print *, c
     


integration_tests/string_90.f90
```diff
@@ -0,0 +1,13 @@
+program string_test
+    implicit none
+    character(len=2) :: c(6) = 'a'
+    
+    !  c(1) should be 'a ' not 'a'
+    if (len(c(1)) /= 2) then
+        print *, "Error: length mismatch"
+        stop 1
+    end if
```
assem2002:
```suggestion
    !  c(1) should be 'a ' not 'a'
    if (len(c(1)) /= 2) error stop
    if ((c(1)(1:1) /= "a")) error stop
    if ((c(1)(2:2) /= " ")) error stop
```
let's test something that already fails in main

opixdown:
thank you assem!! applied your suggestion..

PR #9172, opened 2025-12-23
fix: deallocate intent(out) struct allocatable components at function entry

krystophny:
## Summary

Extend `IntentOutDeallocateVisitor` to handle derived types with allocatable components.

When an `intent(out)` argument is a StructType, iterate through its members and generate deallocation code for each allocatable component.

## Scope

**Single file:** `src/libasr/pass/insert_deallocate.cpp`

Changes:
- Line 341-342: Allow StructType in addition to allocatable
- After line 349: Add conditional branch to handle struct members

## Test

`integration_tests/intent_02.f90` - tests that allocatable struct components are deallocated for intent(out) parameters

## Verification

```bash
./run_tests.py -s -t integration_tests/intent_02.f90
./run_tests.py -s -t integration_tests/intent*
```

Both pass locally.

## Relation to PR #9108

This fix is independent and targets upstream/main. After CI passes here, it will be cherry-picked onto PR #9108 to test with strict double-allocation checks enabled.

### DIFF
--- a/src/libasr/pass/insert_deallocate.cpp
+++ b/src/libasr/pass/insert_deallocate.cpp
@@ -429,7 +429,7 @@ public:
                         ASR::expr_t* member_expr = ASRUtils::EXPR(
                             ASRUtils::getStructInstanceMember_t(al, loc,
                             (ASR::asr_t*)var_expr, const_cast<ASR::symbol_t*>(arg_sym),
-                            struct_member.second, nullptr));
+                            struct_member.second, x.m_symtab));
 
                         // Create: if (allocated(arg%member)) deallocate(arg%member)
                         Vec<ASR::expr_t*> allocated_args;


certik:
Thanks for this. Let's run exhaustive tests and see.

integration_tests/CMakeLists.txt
```diff
@@ -2848,3 +2848,4 @@ RUN(NAME test_ord LABELS llvm llvm_wasm llvm_wasm_emcc) # No gfortran
 RUN(NAME polymorphic_select_type_01 LABELS gfortran llvm)
 RUN(NAME defined_op_match_01 LABELS gfortran llvm)
 RUN(NAME intent_out_array_01 LABELS gfortran llvm)
+RUN(NAME intent_out_allocatable_struct_with_components LABELS gfortran llvm)
```
HarshitaKalani, line 2851: Does it fail with `llvm_wasm llvm_wasm_emcc` flags?

HarshitaKalani (approved):
I've a minor question. Otherwise, looks good to me. Thanks!

PR #9174, opened 2025-12-23
fix: guard intent(out) dealloc for lowered optional args

krystophny:
This PR fixes a crash in optional-argument lowering when a compiler-generated optional wrapper is created and later treated like a real optional dummy.

What changes
- Makes the intent(out) allocatable deallocation guard robust for transformed optional arguments so we do not unconditionally deallocate when the original argument is absent.

Why
- Prevents runtime crashes and incorrect deallocation for code paths created by transform_optional_argument_functions.

Where
- Primarily in the optional-argument transformation and intent(out) deallocation logic, plus an integration test.

Test
- Adds integration test intentout_optional_deallocate_01 exercising the missing-optional path.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2375,6 +2375,7 @@ RUN(NAME allocate_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME allocate_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME dealloc_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME intentout_optional_deallocate_01 LABELS gfortran llvm)
 RUN(NAME dealloc_02 LABELS gfortran)
 RUN(NAME dealloc_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME dealloc_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/intentout_optional_deallocate_01.f90
+++ b/integration_tests/intentout_optional_deallocate_01.f90
@@ -1,0 +1,35 @@
+module intentout_optional_deallocate_01_mod
+    implicit none
+
+    type :: msg_t
+        character(len=:), allocatable :: msg
+    end type msg_t
+
+contains
+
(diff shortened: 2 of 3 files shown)


jinangshah21:
I think reference file is not correctly updated here. I has changes not related to this PR

krystophny:
> I think reference file is not correctly updated here. I has changes not related to this PR

Thanks for checking! Still drafting, so things still a mess.

krystophny:
@jinangshah21 actually the reference change was correct. The `transform_optional_argument_functions` pass creates new ASR function signatures but forgot to copy the m_presence field from original arguments, causing Optional arguments to be downgraded to Required. The fix adds `s_arg->m_value->m_presence = orig_arg->m_presence` to preserve the optional status. The reference files now correctly show Optional instead of Required in the transformed ASR output.

src/libasr/pass/transform_optional_argument_functions.cpp
```diff
@@ -206,9 +206,10 @@ class TransformFunctionsWithOptionalArguments: public PassUtils::PassVisitor<Tra
...
                 ASR::Variable_t* sym_ = ASR::down_cast<ASR::Variable_t>(sym);
                 if (sym_->m_presence == ASR::presenceType::Optional) {
-                    if( set_presence_to_required ) {
-                        sym_->m_presence = ASR::presenceType::Required;
-                    }
+                    // Keep m_presence as Optional to preserve semantic information
+                    // for downstream passes (e.g., insert_deallocate).
+                    // The presence bit argument handles the runtime check.
+                    (void)set_presence_to_required;  // Intentionally unused
                     return true;
```
certik, line 212: This seems like a hack. Either we don't need to do `set_presence_to_required`, in which case let's remove it completely. Or only sometimes we don't need it, so then let's rework the algorithm so that `set_presence_to_required` has the proper value.

What MRE is this trying to fix? It seems once optional arguments are transformed, we don't have any more optional arguments and in the backend we should never see it. So the above change seems wrong?

krystophny: This was a too hacky solution. PR is now completely rewritten with follow-up PRs that can be reviewed independently. All of them now have the purpose to avoid problems in `stdlib-logger` in the thirdparty test once the strict checks of #9163 are armed. I always test them locally on top of #9163 and see how far we come with the stdlib.

krystophny: Ah I see, #9163 is already merged. I check further if this is even needed.

krystophny:

PR #9178, opened 2025-12-24
Fix : Resolve var to a callback function

### LINKED ISSUE #8777: Procedure pointer assignment to dummy argument causes ‘undefined reference’ at link time
The following program does not compile/link with LFortran 0.57

```fortran
module bench_mod
   implicit none
   private
   public :: run

contains

   subroutine run(f)
      implicit none
      interface
         integer function f()
         end function f
      end interface

      procedure(f), pointer :: func_ptr
      integer :: x

      func_ptr => f          ! <-- LFortran fails to resolve this binding
      x = func_ptr()
      print *, "Result:", x
   end subroutine run

end module bench_mod


program main
   use bench_mod
   implicit none

   call run(test_func)

contains

   integer function test_func()
      test_func = 42
   end function test_func

end program main
```

assem2002:
Fixes #8777 

***
## Issue
when `Var` is pointing to a callback function (symbol), `Var` resolves to an external function symbol, hence symbol definition doesn't exist when linking; 
Instead, it should check first if that symbol is bounded to an argument (callback function argument)
## Fix
Check `llvm_symtab_fn_arg` map when resolving `Var` node.


certik (approved):
I think it's fine.

PR #9179, opened 2025-12-24
fix : mismatched argument and parameter doesn't raise error

### LINKED ISSUE #9177: Bug : mismatched argument and parameter doesn't raise error
```.f90
module bench_mod
   interface
      integer function f()
      end function f
   end interface
contains
   subroutine run(m)
      procedure(f) :: m
   end subroutine run

end module bench_mod


program main
   use bench_mod
   implicit none
   call run(1)
end program main
```
```
  %call_arg_value = alloca i32, align 4
 i32 ()*  call void @__module_bench_mod_run(i32* %call_arg_value)

```

opixdown:
## **Issue Fixed**
fixes #9177  mismatched argument and parameter doesn't raise error 

## **Summary**

After this PR, the type mismatch is caught during semantic analysis and reports a clear error message.

## **Reference Code**
```
module bench_mod
   interface
      integer function f()
      end function f
   end interface
contains
   subroutine run(m)
      procedure(f) :: m
   end subroutine run

end module bench_mod


program main
   use bench_mod
   implicit none
   call run(1)
end program main
```

```
  %call_arg_value = alloca i32, align 4
```

```
lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran ./t.f90
semantic error: Type mismatch in argument `m`: expected a procedure but got `integer`
  --> ./t.f90:17:13
   |
17 |    call run(1)
   |             ^ 
```

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -4776,19 +4776,19 @@ public:
                         ASR::expr_t* passed_arg = args[i].m_value;
                         ASR::ttype_t* passed_type = ASRUtils::expr_type(passed_arg);
                         ASR::ttype_t* param_type = v->m_type;
-                    // Check if parameter expects a procedure type 
-                    if (ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(param_type))){
-                        if (!ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(passed_type))){
-                            std::string passed_str =ASRUtils::type_to_str_fortran_expr(passed_type, nullptr);
-                            diag.add(diag::Diagnostic(
-                                "Type mismatch in argument `" + std::string(v->m_name) +
-                                "`: expected a procedure but got `" + passed_str + "`",
-                                diag::Level::Error, diag::Stage::Semantic, {
-                                    diag::Label("", {passed_arg->base.loc})
-                                }));
-                            throw SemanticAbort();
+                        // Check if parameter expects a procedure type 
+                        if (ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(param_type))){
+                            if (!ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(passed_type))){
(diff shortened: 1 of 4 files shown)


HarshilShah1804:
Thanks @opixdown for the PR. Please add the error test in `continue_compilation_*.f90`

opixdown:
> Thanks @opixdown for the PR. Please add the error test in `continue_compilation_*.f90`

 on it !

HarshitaKalani (approved):
Looks good to me. Thanks!

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4776,6 +4776,19 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                            std::string passed_str =ASRUtils::type_to_str_fortran_expr(passed_type, nullptr);
+                            diag.add(diag::Diagnostic(
+                                "Type mismatch in argument `" + std::string(v->m_name) +
+                                "`: expected a procedure but got `" + passed_str + "`",
+                                diag::Level::Error, diag::Stage::Semantic, {
+                                    diag::Label("", {passed_arg->base.loc})
+                                }));
+                            throw SemanticAbort();
+                        }
+                    }
```
certik: The indentation is incorrect.

certik (approved):

src/lfortran/semantics/ast_body_visitor.cpp
certik, line 4800: I think the above check should go into `check_equal_type`, we should not be special casing this particular case above. The `ASRUtils::check_equal_type` should return false for `integer` and `procedure`.

certik:
@opixdown if you plan to implement the proper fix, we can merge the above as a temporary workaround. Otherwise I would leave the PR open until it's implemented properly, so that we don't increase technical debt.

opixdown:
@certik thank you for the clarification. I think we can merge this as a temporary workaround and follow up with a proper fix in ASRUtils::check_equal_type in a separate PR shortly. Please let me know if that works.

certik:
That works. The PR is in (thanks!) and you can now go ahead and implement the proper fix.

PR #9180, opened 2025-12-24
Extension: Allow `pointer, intent(in)` passed objects

certik:
Implements the proposal https://github.com/j3-fortran/fortran_proposals/issues/354.

### DIFF
--- a/integration_tests/class_88.f90
+++ b/integration_tests/class_88.f90
@@ -1,4 +1,4 @@
-module m
+module class_88_m
   type t
     integer :: n = 0
    contains
@@ -10,10 +10,10 @@ module m
     integer, pointer :: f
     f => this%n
   end function f
-end module m
+end module
 
 program test
-  use m
+  use class_88_m
   implicit none
   type(t), target :: xt
   class(t), pointer :: xp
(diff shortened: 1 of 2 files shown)


certik:
@jinangshah21 do you mind reviewing it please? It fixes a couple related bugs that I think happen even without the extension.

integration_tests/class_88.f90
```diff
@@ -0,0 +1,47 @@
...
+  if (result_ptr /= 42) error stop "Failed: xt%tbp() should return 42"
+
+  ! Test type-bound procedure call with class pointer
+  result_ptr => xp%tbp()
+  print *, result_ptr, xp%tbp()
+  if (result_ptr /= 42) error stop "Failed: xp%tbp() should return 42"
+
+  ! Test assignment through pointer-returning TBP with type variable
+  xt%tbp() = 99
+  print *, xt%tbp()
```
jinangshah21, line 38: Small doubt: `xt%btp()` returns a pointer to `n` then why we are assigning it, why not associating?

certik: We are associating it on line 27, so that use case is already tested above. But you can also assign to it, which I didn't know you can do before and there were some bugs that I fixed in this PR to make this work, so we also test that.

certik:
Already twice when I pushed into this PR, it would close it. Then I reopen, but it wouldn't run exhaustive checks (despite the label), so I also had to remove and re-add it.

Issue #9042, opened 2025-12-12
Intrinsic array functions don't handle reallocation of allocatable output arguments


krystophny:
When intrinsic array functions like `matmul` are called in a loop with an allocatable output, the allocation fails because the output is already allocated from the previous iteration.

**MRE:**
```fortran
program test
    real, allocatable :: A(:,:), B(:,:), C(:,:)
    integer :: i
    allocate(A(10,10), B(10,10))
    call random_number(A)
    call random_number(B)
    do i = 1, 2
        C = matmul(A, B)  ! Second iteration fails
    end do
    print *, 'OK'
end program
```

**Error:**
```
At X:Y of file test.f90
Runtime Error: Attempting to allocate already allocated variable 'result'
```

**Root cause:**
In `instantiate_MatMul` (`intrinsic_array_function_registry.h:4030`):
```cpp
if (is_allocatable(result)) {
    body.push_back(al, b.Allocate(result, alloc_dims));
}
```

This allocates `result` (which is the passed-in `C`) without first deallocating it if already allocated.

**Suggested fix:**
Before the `Allocate`, add a conditional deallocate:
```fortran
if (allocated(result)) deallocate(result)
allocate(result, ...)
```

This affects all intrinsic array functions that allocate their output argument.

certik:
If a function returns an allocatable array, then the correct behavior should be that if the LHS is either not allocated, or allocated to the right size, it will transfer the allocation to it.

If a function returns an array that is not allocatable (like matmul), then we require the LHS to be allocated and the right size.

So the above code in this PR should only work with `--realloc-lhs-arrays`. Otherwise it should fail on the first call to matmul, as the LHS is not allocated. If you add `allocate(C(10,10))` before the loop, then it should work without `--realloc-lhs-arrays`.


All three cases already work, so to fix this issue, let's add three tests:

* Check that the above test *works* with `--realloc-lhs-arrays`
* Check that the above test *fails* without `--realloc-lhs-arrays`
* Check that the above test with `allocate(C(10,10))` *works* without `--realloc-lhs-arrays`.

certik:
Note: the issue says that it's the second iteration that fails without `--realloc-lhs-arrays`, but it's actually already the first iteration as can be checked by:
```fortran
program test
    real, allocatable :: A(:,:), B(:,:), C(:,:)
    integer :: i
    allocate(A(10,10), B(10,10))
    !allocate(C(10,10))
    call random_number(A)
    call random_number(B)
[...]

PR #9182, opened 2025-12-24
fix: handle allocatable results in matmul intrinsic array function

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/9042


HarshitaKalani:
Similar issue is there for cshift, eoshift, transpose and maybe other intrinsic array functions.
```console
% 
cat a.f90
program test
    real, allocatable :: A(:,:), B(:,:), C(:,:)
    integer :: i
    allocate(A(10,10), B(10,10))
    call random_number(A)
    call random_number(B)
    do i = 1, 2
        C = transpose(A)  ! Second iteration fails
    end do
    print *, 'OK'
end program%                                                            
(lf) harshitakalani@Harshitas-MacBook-Air:~/repos/lfortran(matmul_01) % 
lfortran a.f90
At 8:9 of file a.f90
Runtime Error: Array 'c' is not allocated.

Use '--realloc-lhs-arrays' option to reallocate LHS automatically.
(lf) harshitakalani@Harshitas-MacBook-Air:~/repos/lfortran(matmul_01) % 
cat a.f90     
program test
    real, allocatable :: A(:,:), B(:,:), C(:,:)
    integer :: i
    allocate(A(10,10), B(10,10))
    call random_number(A)
    call random_number(B)
    do i = 1, 2
        C = cshift(A, 2)  ! Second iteration fails
    end do
    print *, 'OK'
end program%                                                            
(lf) harshitakalani@Harshitas-MacBook-Air:~/repos/lfortran(matmul_01) % 
lfortran a.f90
At 8:13 of file a.f90
Runtime Error: Array 'result' is indexed but not allocated.
```

certik:
Let's think about this. If a function returns an allocatable array, then the correct behavior should be that if the LHS is either not allocated, or allocated to the right size, it will transfer the allocation to it.

If a function returns an array that is not allocatable (like matmul), then we require the LHS to be allocated and the right size.

So I think this PR implements an incorrect behavior?

certik:
Here is how to fix that issue: https://github.com/lfortran/lfortran/issues/9042#issuecomment-3689898085.

certik:
I think this PR is not the correct approach, so I am going to close it. It's effectively reverting https://github.com/lfortran/lfortran/pull/9089, but see my comment above for the correct fix.

PR #9184, opened 2025-12-24
fix: handle forward referenced procedure pointers in derived types

### LINKED ISSUE #9063: Bug: abstract interface declared after derived type not visible for procedure pointer component
### Bug
LFortran reports `Procedure type ... not declared` when a derived type has a procedure pointer component referencing an abstract interface that is declared later in the same module.

### MRE
`mre.f90`:

```fortran
module m
  implicit none

  type :: ctx_t
     procedure(f_iface), pointer, nopass :: fn => null()
  end type ctx_t

  abstract interface
     real function f_iface(x)
       real, intent(in) :: x
     end function f_iface
  end interface

end module m
```

### Expected
Compiles (gfortran accepts it).

### Actual
LFortran fails with:

```
semantic error: Procedure type f_iface not declared
```

### Evidence
- Source: `/tmp/lfortran-dev/fortplot/mre_proc_interface_order_2025-12-13_211921.f90`
- LFortran log: `/tmp/lfortran-dev/fortplot/mre_proc_interface_order_lfortran_2025-12-13_211921.txt`
- gfortran (success): `gfortran -std=f2018 -c <file>`

### Context
This blocks fortplot builds with LFortran (`--std=f23`) in `src/plotting/fortplot_streamline.f90` (type `velocity_function_context_t` vs later `abstract interface`).

Fortplot workaround tracker: lazy-fortran/fortplot#1427.

ANAS727189:
Fixes #9063 

This PR resolves Issue #9063, where forward-referenced procedure pointers within derived types caused a compiler crash.

First, in `ast_common_visitor.h`, when a procedure pointer refers to an undeclared symbol, we now generate a temporary placeholder `Function_t`. Crucially, a fresh SymbolTable is created for this placeholder to serve as its scope owner, preventing the ASR verification crash associated with incorrect parent scope assignment. 
Second, in `ast_symboltable_visitor.cpp`, the symbol re-definition logic was updated to bypass the signature mismatch error; if the existing symbol is identified as an empty placeholder (0 arguments, no return), the compiler now correctly allows the actual interface definition to overwrite it.

issue_9063.f90 - 
```bash
module m
  implicit none

  type :: ctx_t
     procedure(f_iface), pointer, nopass :: fn => null()
  end type ctx_t

  abstract interface
     real function f_iface(x)
       real, intent(in) :: x
     end function f_iface
  end interface

end module m

program test_main
    use m
    print *, "Test Passed Successfully!"
end program test_main
```
Output - 

```bash
(lf) anas@anas-mint:/lfortran$ ./inst/bin/lfortran issue_9063.f90
Test Passed Successfully!
(lf) anas@anas-mint:/mnt/MasterDrive/Fortran-dev/lfortran$
```


src/lfortran/semantics/ast_common_visitor.h
kmr-srbh: I think we need this error message for other cases when a procedure is not declared.

tests/issue_9063.f90
```diff
@@ -0,0 +1,19 @@
...
+       real, intent(in) :: x
+     end function f_iface
+  end interface
+
+end module m
+
+program test_main
+    use m
+    print *, "Test Passed Successfully!"
+end program test_main
```
certik: Add this to integration tests and register it.

Then add tests for all new error messages to tests/errors/continue_compilation_1

certik:
Also we need to add tests for cases when the variable is not actually declared --- right now it seems you just removed the error message, but we still need to give the error message (and test it in continue_compilation) if it is not declared later.

ANAS727189:
> Also we need to add tests for cases when the variable is not actually declared --- right now it seems you just removed the error message, but we still need to give the error message (and test it in continue_compilation) if it is not declared later.

@certik I have implemented the changes you requested
However, currently, I haven't implemented the check to verify that the variable is actually defined. As a result, truly undefined interfaces (like in the new `ctx_missing_t` test case) currently do not trigger an error.

Could you point me to the best place to add this verification? (should I iterate over the symbol table at the end of the module scope to check for remaining placeholders?)

certik:
> Could you point me to the best place to add this verification? (should I iterate over the symbol table at the end of the module scope to check for remaining placeholders?)

Either that, or store these placeholders in an array, then only iterate over this array at the end and check that they are all defined.

A third option is to remove every placeholder from this array when it gets defined, and at the end you just check if this array is empty, otherwise give an error message for each item in there.

tests/errors/continue_compilation_1.f90
```diff
@@ -20,6 +20,36 @@ module continue_compilation_1_mod
...
+    type :: ctx_missing_t
+        procedure(f_missing), pointer, nopass :: fn => null()
+    end type
+
+
+
+
+
+
+
```
certik, line 39: You need to remove equal amount of empty lines that you added, to keep the diff below small, see the comments in this file. Keep the above together (max 1 empty line between statements) to not "waste" lines.

certik:
The file `asr-operator_overloading_12-b71f657.stdout` was removed by a mistake and the CI then fails.

certik:
The approach is roughly correct. Go ahead and finish the PR, I think this will work.

certik:
You have to merge main and resolve conflicts.

ANAS727189:
Done @certik . I have added the check for undeclared variables and a new test case for it.

integration_tests/issue_9063.f90
```diff
@@ -0,0 +1,20 @@
+module m
```
certik: Call this `issue_9063_m`, otherwise the module name can clash in parallel compilation with other tests.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -389,6 +389,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
         in_module = true;
         visit_ModuleSubmoduleCommon<AST::Module_t, ASR::Module_t>(x);
+        ASR::symbol_t *t = ASR::down_cast<ASR::symbol_t>(tmp);
+        ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(t);
+        for (auto &[name, placeholder_sym] : pending_proc_placeholders) {
+            // Check the module's symbol table
+            ASR::symbol_t *current_sym = m->m_symtab->resolve_symbol(name);
+            if (current_sym == placeholder_sym) {
+                diag.add(diag::Diagnostic(
+                    "Procedure '" + name + "' is declared but not defined",
```
certik: What does it mean to be declared but not defined? Can you explain that to me in a simple sentence?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1769,19 +1796,24 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                                diag::Level::Error, diag::Stage::Semantic, {
+                                    diag::Label("", {tmp->loc})}));
+                            throw SemanticAbort();
+                        }
                     }
-                    // Previous declaration will be shadowed
                     parent_scope->erase_symbol(sym_name);
-                } else {
+                }
+                else {
```
certik: The previous formatting was better.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -52,6 +52,12 @@ semantic error: Dummy argument 'a' not defined
 98 |        end subroutine
    | ...^^^^^^^^^^^^^^^^^^ 
 
+semantic error: Procedure 'f_missing' is declared but not defined
+  --> tests/errors/continue_compilation_1.f90:31:9
+   |
+31 |         procedure(f_missing), pointer, nopass :: fn => null()
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Declared here
```
certik: If you have access to `f_missing`'s location, then use it in the error, to only underline that.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -131,13 +137,7 @@ semantic error: Expecting a scalar integer or parameter annotated integer variab
 163 |     logical :: mask_size(size(arr_size))
     |                          ^^^^^^^^^^^^^^ 
 
-semantic error: `protected` attribute is only allowed in specification part of a module
-   --> tests/errors/continue_compilation_1.f90:164:5
-    |
-164 |     integer, protected :: protected_attr_var
-    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
-
```
certik, line 139: It looks like we lost this error --- we need to investigate why it doesn't show anymore.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -131,13 +137,7 @@ semantic error: Expecting a scalar integer or parameter annotated integer variab
...
-semantic error: `protected` attribute is only allowed in specification part of a module
+semantic error: `parameter` attribute conflicts with `protected` attribute
```
certik: This error changed, we need to investigate, likely due to the previous error missing.

tests/errors/procedure_interface_missing.f90
```diff
@@ -0,0 +1,13 @@
...
+  type :: ctx_t
+     ! This should trigger an error because 'f_missing' is not defined
+     procedure(f_missing), pointer, nopass :: fn => null()
+  end type ctx_t
+
+end module m
+
+program test
+    use m
+end program
```
certik: I think this file is not used anywhere, so can be removed.

src/lfortran/semantics/ast_symboltable_visitor.cpp
ANAS727189: It means the code references an interface, but the actual definition of that interface was never found in the code.
Should I update it ?

ANAS727189:
Addressed all the feedback and fixed everything. I fixed the f_missing location to underline just the declaration and restored the missing error.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -464,7 +478,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 if ( !compiler_options.continue_compilation ) throw e;
             }
         }
-        in_program = false;
+        // in_program = false;
```
certik, line 481: Should this line be removed?

tests/errors/continue_compilation_1.f90
```diff
@@ -16,20 +16,20 @@ module continue_compilation_1_mod
...
-
+    type :: ctx_fail_t
+        procedure(f_fail), pointer, nopass :: fn => null()
+    end type
+    
+    abstract interface
+        subroutine f_fail(x)
+            real, intent(in) :: x
+        end subroutine
+    end interface
```
certik, line 27: What is the purpose of these lines? I don't see any error message being emitted for these. Is this a test for forward declaration? If so, that should go into integration tests (however you already test it there).

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1769,17 +1795,21 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+
+                    if (!is_placeholder) {
+                       if (!ASRUtils::types_equal(f2->m_function_signature, func->m_function_signature, 
+                                ASRUtils::get_expr_from_sym(al, f1), ASRUtils::get_expr_from_sym(al, func_sym))) {
+        
+                            diag.add(diag::Diagnostic(
+                                "Argument(s) or return type mismatch in interface and implementation",
+                                diag::Level::Error, diag::Stage::Semantic, {
+                                    diag::Label("", {tmp->loc})}));
+                            throw SemanticAbort();
```
certik, line 1810: This error message is not tested --- it wasn't tested before, but now with extra logic in there, we should ensure that it actually triggers when it should.

Also we need to ensure that it triggers when `is_placeholder` is true ---- later, right?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -534,6 +548,18 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
         add_generic_procedures();
         evaluate_postponed_calls_to_genericProcedure();
+        for (auto &[name, placeholder_sym] : pending_proc_placeholders) {
+             ASR::symbol_t *current_sym = current_scope->resolve_symbol(name);
+             if (current_sym == placeholder_sym) {
+                 diag.add(diag::Diagnostic(
+                    "Procedure '" + name + "' is declared but not defined",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("Declared here", {placeholder_sym->base.loc})
+                    }));
```
certik, line 558: This error message is not tested --- it still uses the old phrasing.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -389,6 +389,20 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        ASR::symbol_t *t = ASR::down_cast<ASR::symbol_t>(tmp);
+        ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(t);
+        for (auto &[name, placeholder_sym] : pending_proc_placeholders) {
+            // Check the module's symbol table
+            ASR::symbol_t *current_sym = m->m_symtab->resolve_symbol(name);
+            if (current_sym == placeholder_sym) {
+                diag.add(diag::Diagnostic(
+                    "Interface '" + name + "' is referenced but not defined",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("Declared here", {placeholder_sym->base.loc})
```
certik, line 401: Maybe:
```suggestion
                        diag::Label("Referenced here", {placeholder_sym->base.loc})
```
To be consistent.

ANAS727189:
Thanks for the merge! I will address all the cleanup items, fix the error labels, and add the missing test case in a follow-up PR shortly.

src/lfortran/semantics/ast_symboltable_visitor.cpp
ANAS727189: Yes, that was a leftover. I will remove the commented out line in the follow-up PR.

tests/errors/continue_compilation_1.f90
ANAS727189: You are right, these are redundant now since I added the proper positive test case in integration_tests. I will remove these lines in the follow-up PR

PR #9187, opened 2025-12-24
Check VTABLE_PTR_SIZE in the LLVM backend

### LINKED ISSUE #8778: Add an assert check for struct size in openmp

certik:
Later we can even upgrade it to use the runtime value. There is an issue that this is needed in the openmp pass that runs before (or even without) LLVM, so for now the current solution is good enough and we can further improve it later.

Fixes #8778.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -15497,7 +15497,7 @@ Result<bool> check_vtable_ptr_size(llvm::Module* module, diag::Diagnostics &diag
     uint64_t actual_ptr_size = data_layout.getPointerSize();
     if (actual_ptr_size != ASRUtils::VTABLE_PTR_SIZE) {
         std::string msg = "VTable pointer size mismatch: expected " +
-            std::to_string(ASRUtils::VTABLE_PTR_SIZE) + " bytes (64-bit), but target uses " +
+            std::to_string(ASRUtils::VTABLE_PTR_SIZE) + " bytes, but target uses " +
             std::to_string(actual_ptr_size) + " bytes";
         diagnostics.diagnostics.push_back(diag::Diagnostic(msg,
             diag::Level::Error, diag::Stage::CodeGen));


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -193,6 +193,15 @@ namespace LCompilers {
     }
     void LLVMUtils::set_module(llvm::Module* module_) {
         module = module_;
+
+        // Verify that the vtable pointer size matches the architecture assumption.
+        // The VTABLE_PTR_SIZE constant (8 bytes) assumes 64-bit pointers.
+        llvm::DataLayout data_layout(module->getDataLayout());
+        uint64_t actual_ptr_size = data_layout.getPointerSize();
+        LCOMPILERS_ASSERT_MSG(actual_ptr_size == ASRUtils::VTABLE_PTR_SIZE,
```
assem2002:
```suggestion
        llvm::DataLayout data_layout(module->getDataLayout());
        LCOMPILERS_ASSERT_MSG( data_layout.getPointerSize() == ASRUtils::VTABLE_PTR_SIZE,
```
To avoid the release mode error.

assem2002 (approved):
Looks good to me. Thanks!
I've an observation. I can remember using that constant because we always used to add a vptr with all structs;
 Didn't we change that lately to only use vptr when we only need it?, in other words, openmp structs should be C structs with no vptr.

src/libasr/codegen/llvm_utils.cpp
certik: I reworked it, now you need this variable.

certik:
> I've an observation. I can remember using that constant because we always used to add a vptr with all structs;
Didn't we change that lately to only use vptr when we only need it?, in other words, openmp structs should be C structs with no vptr.

I don't know --- if so, then the openmp code has to be updated.

src/libasr/pass/openmp.cpp
```diff
@@ -3251,7 +3250,7 @@ class ParallelRegionVisitor :
                     total_size += 8;
                 }
             }
-            return total_size + STRUCT_VTABLE_SIZE;
+            return total_size + ASRUtils::VTABLE_PTR_SIZE;
```
assem2002, line 3253:
```suggestion
            return total_size;
```
If vptr doesn't exist in openmp-special structs (C structs) -- This change should work. despite the fact, that everything works just fine with this code. I'm just sharing my concern

certik: That's easy to check!

https://github.com/lfortran/lfortran/pull/9190

certik: It worked.

certik:
The simpler approach in https://github.com/lfortran/lfortran/pull/9190 worked, so we merged it. If we every need to know the size in ASR, this PR shows how to do it.

PR #9190, opened 2025-12-24
openmp: do not add 8 bytes

certik:
The struct is now a C struct, without a vtable pointer.

Fixes https://github.com/lfortran/lfortran/issues/8778.


certik:
Locally the integration tests run for me, so we'll see at the CI.

certik:
Tests passed, so apparently it works. Thanks @assem2002 !

PR #9199, opened 2025-12-25
fix: COMMON block union-based storage association

krystophny:
## Summary

Implements true union-based storage association for COMMON blocks. Different program units can now view the same storage with different variable names, types, and layouts.

**Alternative to #9122** - uses byte-offset tracking instead of name-based lookup.

## How It Works

1. **Byte offset tracking** (`common_variables_byte_offset` map):
   - Each variable's byte offset within its COMMON block is recorded at declaration time
   - Calculated using `get_type_byte_size()` which handles all ASR types

2. **Member resolution by byte offset** (`create_StructInstanceMember`):
   - First tries direct name lookup
   - Falls back to byte-offset range matching when names differ
   - Finds the struct member whose `[offset, offset+size)` range contains the target

3. **Type aliasing via bitcast** (`asr_to_llvm.cpp`):
   - When local variable type differs from struct member type, bitcasts the pointer
   - Works on both LLVM 11 (typed pointers) and LLVM 15+ (opaque pointers)

## Files Changed

| File | Changes |
|------|---------|
| `ast_common_visitor.h` | +`common_variables_byte_offset` map, `get_type_byte_size()`, byte-offset member resolution |
| `asr_to_llvm.cpp` | Bitcast when types differ |
| `ast_body_visitor.cpp` | Pass new map |
| `ast_symboltable_visitor.cpp` | Pass new map |
| `ast_to_asr.cpp` | Initialize new map |

## Test Coverage (9 new tests)

| Test | Scenario |
|------|----------|
| `common_15` | Different variable names, same layout |
| `common_16` | Alignment edge case (int(1), real(8), int(4)) |
| `common_17` | Type aliasing (real(8) array vs int array) - **verified** |
| `common_18` | Reordered members |
| `common_19` | CHARACTER variables |
| `common_20` | Bidirectional value sharing |
| `common_21` | Mixed types with type punning - **values verified across views** |
| `common_22` | Multiple independent COMMON blocks |
| `common_23` | Blank (unnamed) COMMON block |

## Verification

All tests pass locally (LLVM 11) and in CI (LLVM 11 + 21):
```
scripts/lf.sh test -t common_15 -s  # PASS
scripts/lf.sh test -t common_16 -s  # PASS  
scripts/lf.sh test -t common_17 -s  # PASS (with verification)
# ... all common_15-23 pass
```

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -3747,6 +3747,10 @@ public:
         // the struct member type (member->m_type), we must bitcast the pointer
         // to match the local view. Example: real(8) array(4) in one unit vs
         // integer array(8) in another - both occupy 32 bytes of shared storage.
+        //
+        // On LLVM 15+ with opaque pointers, all pointers are just `ptr` so
+        // bitcast is unnecessary - the type info is used by load/store instead.
+#if LLVM_VERSION_MAJOR < 15
         ASR::ttype_t* expected_type = x.m_type;
         ASR::ttype_t* actual_member_type = member->m_type;
         llvm::Type* expected_llvm_type = llvm_utils->get_type_from_ttype_t_util(
@@ -3759,6 +3763,7 @@ public:
             // Bitcast pointer to the local program unit's view of the storage
             tmp = builder->CreateBitCast(tmp, expected_llvm_type->getPointerTo());
         }
+#endif
     }
 
     void visit_StructConstant(const ASR::StructConstant_t& x) {


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2254,17 +2315,58 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                actual_member_name = member_name_str;
+                                struct_member_sym = mem_sym;
+                                break;
+                            }
+                            current_offset += member_size;
+                        }
+                    }
+                }
+            }
+            LCOMPILERS_ASSERT(struct_member_sym != nullptr);
```
certik, line 2355: Is this block of code only involved for common blocks, or for any struct?

(We need to make sure we didn't make these loose rules for any struct, only for common blocks.)

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3740,6 +3740,30 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            const_cast<ASR::expr_t*>(reinterpret_cast<const ASR::expr_t*>(&x)),
+            expected_type, module.get());
+        llvm::Type* actual_llvm_type = llvm_utils->get_type_from_ttype_t_util(
+            const_cast<ASR::expr_t*>(reinterpret_cast<const ASR::expr_t*>(&x)),
+            actual_member_type, module.get());
+        if (expected_llvm_type != actual_llvm_type) {
+            // Bitcast pointer to the local program unit's view of the storage
+            tmp = builder->CreateBitCast(tmp, expected_llvm_type->getPointerTo());
+        }
+#endif
```
certik, line 3766: This should be hidden behind some LLVM-version-independent API. We can create a function for this in llvm utils. We can do this later, this is minor.

certik (approved):
I think this looks good, I don't see anything major.

Obviously to clean this up completely will take time, but we'll do that later. For example the `get_type_byte_size` computation in ast_common_visitor depends on the backend, and it's probably platform dependent. But we can figure out later how to do this, for example by creating some API that we can call in ASR, but it would ask the current backend to fill in the sizes, or whatever. For now it's good, it gets things going, that's what matters.

Also, we have a Union type, and I think it does exactly what you implemented here. So we can later figure out how to just create a Union, and then implement the Union using the techniques in this PR.

src/lfortran/semantics/ast_common_visitor.h
certik: I would say it's ok to merge this, to get us unblocked, and tighten it up later, if needed.

certik:
Don't push any more commits, let's see if full tests pass. If so, let's merge, to lock this win.

src/lfortran/semantics/ast_common_visitor.h
krystophny: It’s `COMMON`-only. The rewrite happens in `CommonVisitor::replace_with_common_block_variables()` which calls `create_StructInstanceMember()`, and that function returns non-null only if `common_variables_hash` contains the variable hash (if (`common_variables_hash.find(hash) != common_variables_hash.end())`). So non-`COMMON` variables are untouched (no relaxed rules applied). The name/byte-offset fallback is additionally gated by `common_variables_byte_offset`.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: Agreed.

PR #9202, opened 2025-12-25
extension mismatched argument and parameter doesn't raise error

opixdown:
extension https://github.com/lfortran/lfortran/pull/9179

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1922,7 +1922,7 @@ RUN(NAME nested_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-
+RUN(NAME nested_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME nested_vars1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)
 RUN(NAME nested_vars2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c)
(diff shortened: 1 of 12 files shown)


opixdown:
here for the code `implicit_interface_04.f90`:
```program main

integer :: b(3) = [10, 20, 30]
integer :: n = 3

call driver(implicit_interface_check, b, n)

contains

subroutine driver(fnc, arr, m)
integer, intent(in) :: m, arr(m)
print *, abs(-1.0)
call fnc(arr, m, arr(3))
end subroutine

subroutine implicit_interface_check(arr1, m, c)
integer, intent(in) :: m, arr1(m), c
if (m /= 3) error stop
if (c /= 30) error stop
if (arr1(1) /= 10) error stop
if (arr1(2) /= 20) error stop
if (arr1(3) /= 30) error stop
end subroutine

end program
```

it also catches the error of proc missmatch 
is my approach is not correct?

tests/reference/llvm-implicit_interface_04-9b6786e.stderr
```diff
@@ -0,0 +1,5 @@
+[0;31;1msemantic error[0;0m[0;1m: Type mismatch in argument `fnc`: expected `real` but got `FunctionType(integer[:], integer, integer, )`[0;0m
+ [0;34;1m-->[0;0m tests/../integration_tests/implicit_interface_04.f90:6:13
+  [0;34;1m|[0;0m
+[0;34;1m6 |[0;0m call driver(implicit_interface_check, b, n)
+  [0;34;1m|[0;0m             [0;31;1m^^^^^^^^^^^^^^^^^^^^^^^^ [0;0m
```
certik, line 5: This case now gives an error, so it needs to be investigated.

src/lfortran/semantics/ast_body_visitor.cpp
certik: I see --- I missed that we are skipping it for functions.

I don't know why, I think we should check all arguments.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4816,6 +4803,11 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                                             ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(passed_type)) ||
                                             ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(param_type));
                         // Check if types are equal
+                        if ((ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(passed_type)) !=
+                                                    ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(param_type)) &&
+                                                    !compiler_options.implicit_argument_casting)) {
+                                                    skip_check = false; // force checking to catch the mismatch
+                                                }
```
certik: Is this equivalent to just remove skipping functions above for `skip_check`? If so, I would just do that.

opixdown: Agreed. I was not sure why function types were being skipped. With the procedure type check now in check_equal_type, there is no reason to skip them.

Issue #9203, opened 2025-12-26
Wrong output with EN0.0E0 or ES0.0E0 format


harperjf:
This program is the same one as in #3934 when EN format was not yet allowed. The program now compiles and runs but gives wrong output.
```
program testen0
  write(*,"(EN0.0E0)") 3.14159
end program testen0
```
Output from Lfortran 0.58.0, wrong
```
3.141590118
```
Output from gfortran or ifx, right
```
3.E0
```
If the format is changed to ES0.0E0,  lfortran gives the same wrong output, and gfortran and ifx give the same right output.

PR #9205, opened 2025-12-26
fix: give correct output with EN and ES format

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9203


certik (approved):
I think this is fine, thanks!

Issue #9194, opened 2025-12-25
Formatted I/O: S, SP, and SS edit descriptors not recognized


w6ws:
Consider:
```
program sspss
  implicit none

  print 100, -42.43, 42.43
  print 101, -42.43, 42.43
  print 102, -42.43, 42.43

100 format (s, 2f10.3)
101 format (sp, 2f10.3)
102 format (ss, 2f10.3)

end program
```
With gfortran:
```
$ gfortran sspss.f90
$ ./a.out
   -42.430    42.430
   -42.430   +42.430
   -42.430    42.430
$
```
With lfortran:
```
$ lfortran -c --continue-compilation sspss.f90
tokenizer error: Token 's' is not recognized in `format` statement
 --> sspss.f90:8:13
  |
8 | 100 format (s, 2f10.3)
  |             ^ 

tokenizer error: Token 's' is not recognized in `format` statement
 --> sspss.f90:9:13
  |
9 | 101 format (sp, 2f10.3)
  |             ^ 

tokenizer error: Token 's' is not recognized in `format` statement
  --> sspss.f90:10:13
   |
10 | 102 format (ss, 2f10.3)
   |             ^ 

tokenizer error: Token 's' is not recognized in `format` statement
  --> sspss.f90:10:14
   |
10 | 102 format (ss, 2f10.3)
   |              ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```

w6ws:
@HarshitaKalani - looks great.  Thanks!

PR #9207, opened 2025-12-26
fix: Add support for sign control format descriptors (S, SP, SS)

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9194


certik (approved):
I think that this is good. I want to merge this after https://github.com/lfortran/lfortran/pull/9199.

PR #9208, opened 2025-12-26
fix: handle character parameter declaration using `type()` syntax

### LINKED ISSUE #7783: Character parameter declared using `type` doesn't allow to initialize with string
MRE:
```fortran
program main
    type(character(len=*)),parameter    :: all='(*(g0))'
end program main
```

LFortran Error:
```console
semantic error: Only function call assignment is allowed for now
  --> example/demo17.f90:18:1
   |
18 | type(character(len=*)),parameter    :: all='(*(g0))'
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

YashNagda17:
Fixes #7783
Changes made:
- Handling `type()` syntax for character parameters in AST-ASR conversions.
- Move the current logic to set and validate character lengths for parameter strings to a separate function to avoid code duplication.

### DIFF
--- a/integration_tests/declaration_03.f90
+++ b/integration_tests/declaration_03.f90
@@ -2,11 +2,10 @@ program declaration_03
     implicit none
     type(character(len=4)) :: str1 = 'abcd'
     type(real):: r1 = 1.53
-    ! TODO: Uncomment for parameter initializations
-    ! type(character(len=*)),parameter:: str2='(*(g0))'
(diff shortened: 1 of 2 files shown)


certik (approved):
I think this looks good, thanks for fixing it. Let's run full tests first.

certik:
Looks like the full tests will pass, so I merged it.

PR #9209, opened 2025-12-26
fix: update error label and add interface mismatch test

ANAS727189:
This PR is a follow-up to #9184 (addressing issue #9063). It addresses two main points:

- It adds a regression test for Interface Mismatch: As requested in the review, I added a test case (tests/errors/interface_mismatch.f90) to verify that the error "Argument(s) or return type mismatch in interface and implementation" triggers correctly.

Running it will give this output - 
```
(lf) anas@anas-mint:/lfortran$ ./src/bin/lfortran tests/errors/interface_mismatch.f90
semantic error: Argument(s) or return type mismatch in interface and implementation
  --> tests/errors/interface_mismatch.f90:14:5 - 17:16
   |
14 |        module function f() result(r)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
17 |        end function
   | ...^^^^^^^^^^^^^^^^ 


(lf) anas@anas-mint:/lfortran$ 

```

- It also fixes an error message label - In` ast_symboltable_visitor.cpp`, the error label for "Interface '...' is referenced but not defined" was changed from "Declared here" to "Referenced here".

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -458,5 +458,3 @@ program continue_compilation_1
     !passing non procedure to procedure parameter
     call proc_param(42)
 end program
-
-
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,7 +2,7 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
-    "infile_hash": "4a4d2c87789202cead1ee5b076f6d14eee0190ea3277035bfe1dfcc8",
+    "infile_hash": "c60f0ee0e9a05253b4e97f06999db79980e6a1b68ac4d7e69e7ad6f1",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,


tests/errors/interface_mismatch.f90
```diff
@@ -0,0 +1,23 @@
...
+    module function f() result(r)
+        real :: r
+        r = 5.5
+    end function
+
+end module
+
+program test
+    use mismatch_function
+end program
```
certik: Move this into continue_compilation_1.f90, no need to create a separate test.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -552,9 +551,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
              ASR::symbol_t *current_sym = current_scope->resolve_symbol(name);
              if (current_sym == placeholder_sym) {
                  diag.add(diag::Diagnostic(
-                    "Procedure '" + name + "' is declared but not defined",
+                    "Interface '" + name + "' is referenced but not defined",
                     diag::Level::Error, diag::Stage::Semantic, {
-                        diag::Label("Declared here", {placeholder_sym->base.loc})
+                        diag::Label("Referenced here", {placeholder_sym->base.loc})
```
certik, line 556: I think this error is not tested, correct?

ANAS727189: No, I updated it just for consistency.

certik: Let's figure out how to test it.

tests/errors/interface_mismatch.f90
ANAS727189: I have updated it.

src/lfortran/semantics/ast_symboltable_visitor.cpp
ANAS727189: ok, will add a case for it in continue_compilation_1.f90

tests/errors/continue_compilation_1.f90
```diff
@@ -457,4 +443,18 @@ program continue_compilation_1
...
+        module function f() result(i)
+            integer :: i
+        end function
+    end interface
+contains
+    module function f() result(r)
+        real :: r
+        r = 5.5
+    end function
+end module
```
certik: Just move this into the module above, we already have a module there for this case.

ANAS727189:
@certik I've implemented the changes. I had to remove the new mismatch test case for now because it was triggering a compiler crash in the symbol lookup logic rather than the intended error message. We will need to fix the C++ lookup logic to handle mismatches gracefully before I can add that test.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -1,200 +1,206 @@
 semantic error: Derived type `another_undefined_type` is not defined
-  --> tests/errors/continue_compilation_1.f90:50:9
+  --> tests/errors/continue_compilation_1.f90:49:9
    |
-50 |         type(another_undefined_type) :: s3_in_subroutine
+49 |         type(another_undefined_type) :: s3_in_subroutine
```
certik: You have to adjust the empty lines so that diffs like these are not here.

src/lfortran/semantics/ast_symboltable_visitor.cpp
certik: There should be 2 errors. Do the errors differ at all? Maybe you can adjust the error message to reflect the unique code path that led to it, then it will be easy to check that both are tested.

ANAS727189: @certik  I can tweak the messages to verify which path is hit. However, right now I can't even trigger this mismatch error because a signature mismatch causes a hard crash (segfault in down_cast) before it reaches the check. The compiler fails to look up the symbol entirely if the signatures differ.

certik: Right, that's why I want to test it. You can do this in another PR. If the error cannot be triggered, then we should remove it, replace it with an ICE error.

tests/errors/continue_compilation_1.f90
```diff
@@ -16,21 +16,21 @@ module continue_compilation_1_mod
...
-            real, intent(in) :: x
-        end subroutine
-    end interface
-
     ! Test for Missing Declaration:
     type :: ctx_missing_t
         procedure(f_missing), pointer, nopass :: fn => null()
     end type
 
+    procedure(missing_global_interface), pointer :: p => null()
```
certik, line 24: Note: I think the above two error tests actually test exactly the same error. But since one is inside the derived type, it's probably ok to test it without the derived type also.

certik:
Btw, I don't see any mismatch test in the PR, so I removed it from the commit description.

src/lfortran/semantics/ast_symboltable_visitor.cpp
ANAS727189: @certik  It’s definitely a semantic error, not an internal inconsistency, so an ICE wouldn't be right. It's only unreachable right now because the lookup segfaults first.
The issue seems to be in the C++ lookup logic.

ANAS727189: I guess this part of code in ast_body_visitor.cpp causes that crash- 
```
template <typename T>
    void visit_SubmoduleModuleCommon(const T& x) {
        SymbolTable *old_scope = current_scope;
        ASR::symbol_t *t = current_scope->get_symbol(to_lower(x.m_name));
        ASR::Module_t *v = ASR::down_cast<ASR::Module_t>(t);
        current_module_dependencies.clear(al);
        current_scope = v->m_symtab;
        current_module = v;

            if(x.m_decl[i]->type == AST::unit_decl2Type::Template){
            }
        }
```

In this if t is null then it will cause issue, we need to add a check to ensure t is not NULL before trying to cast it

ANAS727189: @certik Its working now, I checked it by running that case now the compiler is not crashing and its testing correctly also, I just had to add a nullptr check in the ast_body_visitor file, and the error now triggers correctly for the test case.
Shall I add the changes in a new PR ?

PR #9223, opened 2025-12-28
fix: transfer(character) to int8 arrays

krystophny:
This PR fixes LFortran codegen for TRANSFER from CHARACTER to INTEGER(INT8) arrays, which is used heavily by stdlib hashing and logger code.

What changes
- Ensures transfer(character, mold=int8, size) produces and propagates a correct byte-array view.
- Fixes the compile-to-object (-c) path where the temporary result is a PointerArray and was previously handled as a descriptor, triggering an LLVM type assertion.
- Prevents array_op from lowering this transfer into an element-wise loop that would assign the string data pointer into each int8 element.

Why
- Unblocks stdlib logger and hash-related code patterns that rely on TRANSFER for byte-level hashing.

Where
- asr_to_llvm.cpp (BitCast/TRANSFER handling and int8 assignment fast path)
- array_op.cpp (skip invalid element-wise lowering for this pattern)
- Adds integration test transfer_string_to_int8_01

Test
- transfer_05 (hash transfer pattern)
- transfer_string_to_int8_01

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -12798,6 +12798,16 @@ public:
                 fixed_arr_ptr = llvm_utils->CreateLoad2(target_llvm_type->getPointerTo(), target_addr);
             }
             target_data = builder->CreateBitCast(fixed_arr_ptr, i8->getPointerTo());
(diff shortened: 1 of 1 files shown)


krystophny:

PR #9226, opened 2025-12-28
fix: pass_array_by_data keep internal calls in-scope

krystophny:
Fix `pass_array_by_data` ASR verification failures in `--separate-compilation` when a parent procedure is duplicated and contains recursive internal procedure calls.

What changed:
- `pass_array_by_data`: re-resolve `Call` symbols (`m_name` / `m_original_name`) into the current scope before rewriting arguments.
- Add integration test `pass_array_by_data_13` covering a recursive internal subroutine call that previously ICEd.


krystophny:

PR #9236, opened 2025-12-28
test: add lapack_testing_02 implicit interface regression

krystophny:
## Summary
Add regression test for allocatable array passed to implicit interface.

## Test
`lapack_testing_02`: Allocatable 1D array passed to external subroutine with assumed-size dummy.

## Merge Order
Depends on #9232. Merge after #9232.


krystophny:
Superseded by #9232 and #9240 which include the same tests and code changes.

PR #9238, opened 2025-12-28
recognize `dimension(..)` as assumed rank declaration

### LINKED ISSUE #9237: Recognize `dimension(..)` variables also as assumed rank variables
Currently, we were only identifying assumed rank variables as that having variable dimension as `..` like `x(..)`. Instead if we declare them as `dimension(..) :: x`, they were not being recognized as assumed rank variables.

MRE
```fortran
program select_rank_test
    implicit none

    integer :: b(2) = [1,2]

    call check(b, 1)

contains

    subroutine check(x, expected)
        integer, dimension(..), intent(in) :: x
        integer, intent(in) :: expected

        select rank(x)
            rank(1)
                if (expected /= 1) error stop

        end select
    end subroutine check
end program select_rank_test

```

HarshilShah1804:
Fixes #9237

### DIFF
--- a/integration_tests/select_rank_01.f90
+++ b/integration_tests/select_rank_01.f90
@@ -13,7 +13,7 @@ program select_rank_test
 contains
 
     subroutine check(x, expected)
-        integer, intent(in) :: x(..)
+        integer, dimension(..), intent(in) :: x
         integer, intent(in) :: expected
 
         select rank(x)


integration_tests/select_rank_01.f90
```diff
@@ -13,7 +13,7 @@ program select_rank_test
 contains
 
     subroutine check(x, expected)
-        integer, intent(in) :: x(..)
+        integer, dimension(..), intent(in) :: x
```
certik, line 16: I think this is equivalent --- is this change done so that we test both approaches?

HarshilShah1804: Yes. As it was a small issue, so instead of adding a new test, I just changed the declaration style in this test.

PR #9240, opened 2025-12-28
codegen: fix bind(c) interface array argument ABI

krystophny:
## Summary

Fix bind(c) interface array argument ABI when calling C functions from Fortran.

When a Fortran array is passed to a bind(c) function that expects a pointer to the array data (e.g., `double*`), LFortran was previously passing a pointer to the entire descriptor structure rather than just the data pointer. This caused ABI mismatch with C libraries.

## Changes

- `libasr/codegen/llvm_array_utils.cpp`: Extract and pass only the data pointer for bind(c) array arguments

## Test Plan

- [x] `bindc6.f90` + `bindc6_sum.c` - calls a C function that sums array elements
- [x] Unblocks LAPACK `SDRVLS` test routine which passes arrays to BLAS functions

### DIFF
--- a/integration_tests/bindc6.f90
+++ b/integration_tests/bindc6.f90
@@ -4,8 +4,6 @@ program bindc6
     real(c_double) :: arr(3)
     real(c_double) :: result
 
-    arr = [1.0d0, 2.0d0, 3.0d0]
-
     interface
         function sum_arr(a, n) bind(c)
             import :: c_double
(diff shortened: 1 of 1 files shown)


integration_tests/bindc6.f90
```diff
@@ -0,0 +1,21 @@
+program bindc6
+    use iso_c_binding, only: c_double
+    implicit none
+    real(c_double) :: arr(3)
+    real(c_double) :: result
+
+    arr = [1.0d0, 2.0d0, 3.0d0]
```
certik: This must go below the interface below.

krystophny: fixed!

Issue #9193, opened 2025-12-25
Formatted I/O: BN and BZ edit descriptors not recognized and cause an ICE.


w6ws:
Consider:
```
program bnbz
  implicit none

  character(:), allocatable :: cdata
  integer :: i1, i2

  cdata = '1   1   '
  read (cdata, 100) i1, i2
  print *, i1, i2

 100 format (bn, i4, bz, i4)

end program
```
Compiling and running with gfortran gives:
```
$ gfortran bnbz.f90
$ ./a.out
           1        1000
$
```
Compiling with lfortran gives:
```
$ lfortran -c bnbz.f90
tokenizer error: Token 'b' is not recognized in `format` statement
  --> bnbz.f90:11:14
   |
11 |  100 format (bn, i4, bz, i4)
   |              ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
If I specify --continue_compilation, additional errors are reported.  Then the compiler ICEs:
```
$ lfortran -c --continue-compilation bnbz.f90 2>&1 
tokenizer error: Token 'b' is not recognized in `format` statement
  --> bnbz.f90:11:14
   |
11 |  100 format (bn, i4, bz, i4)
   |              ^ 

tokenizer error: Token 'n' is not recognized in `format` statement
  --> bnbz.f90:11:15
   |
11 |  100 format (bn, i4, bz, i4)
   |               ^ 

tokenizer error: Token 'b' is not recognized in `format` statement
  --> bnbz.f90:11:22
   |
11 |  100 format (bn, i4, bz, i4)
   |                      ^ 

tokenizer error: Token 'z' is not recognized in `format` statement
  --> bnbz.f90:11:23
   |
11 |  100 format (bn, i4, bz, i4)
   |                       ^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
[...]

PR #9244, opened 2025-12-29
feat: Add BN/BZ format descriptor support 

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9193
fixes: https://github.com/lfortran/lfortran/issues/8967


HarshitaKalani:
```
 393/2040 Test  #393: array_constructor_02 ...............................Subprocess aborted***Exception:   0.00 sec
.
.
.
Errors while running CTest
The following tests FAILED:
Output from these tests are in: /Users/runner/work/lfortran/lfortran/integration_tests/build-lfortran-llvm/Testing/Temporary/LastTest.log
	393 - array_constructor_02 (Subprocess aborted)
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5430,6 +5430,231 @@ LFORTRAN_API void _lfortran_formatted_read(
...
+        default:
+            break;
+        }
+
+        if (*iostat != 0) break;
+    }
+
+    va_end(args);
+    *chunk = (int32_t)src_pos;
+}
```
certik, line 5755: This is now duplicating all the code, right? Is there no way to share the formatted read code between files and strings?

certik:
We keep fixing these formatting descriptors and until now we had to do the fixes at one place. I think after this PR we need to fix them at two places. I think that's not optimal. Let's figure out how to share the code?

src/libasr/runtime/lfortran_intrinsics.c
HarshitaKalani: The existing helper functions (handle_read_I, handle_read_A, etc.) are tightly coupled to FILE*. It was difficult to directly use them for string buffers. I'll think of a better solution which does not introduce much duplication.

certik:
Yes, something along these lines look better.

Is it possible to submit just a PR that does the refactoring? Then it's clear that nothing got changed and what exactly the refactoring is. 

Then you submit the actual fix for BN/BZ on top.

HarshitaKalani:
Yes, I'm working on it.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11932,7 +11932,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
 
     void emit_formatted_read(const ASR::FileRead_t &x, llvm::Value *unit_val,
             llvm::Value *iostat, llvm::Value *read_size, llvm::Value *advance,
-            llvm::Value *advance_length) {
+            llvm::Value *advance_length, bool is_string = false) {
```
certik, line 11935: Let's not use keyword arguments, just set it explicitly every time the function is called.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5538,6 +5570,190 @@ LFORTRAN_API void _lfortran_formatted_read(
     }
 }
 
+// Internal string formatted read - reads from a string buffer instead of a file
+LFORTRAN_API void _lfortran_string_formatted_read(
```
certik, line 5574: This function is duplicated, right? 

So the BN/BZ handling is now implemented twice. That also means our testing methodology of using internal reads from strings are not sufficient, because now we have a separate code path for reading from a file. 

I really don't like it. 

Let's figure out how to share all this code, so that we only have to fix and test bugs once, not twice.

Baitato:
@certik @HarshitaKalani Please check out #9355. I've handled internal string read without code duplication. Note that my PR doesn't contain BN, BZ changes.

PR #9249, opened 2025-12-29
enh: misleading 'not defined' error for assumed-size

### LINKED ISSUE #9105: Misleading 'not defined' error for assumed-size array with wrong star position
When an assumed-size array has `*` in the wrong position, two errors are emitted:

```fortran
subroutine assumed_size_star_pos_1(a)
    real, intent(in) :: a(*, 10)  ! wrong: * must be last
end subroutine
```

Output:
```
semantic error: Assumed-size '*' is only permitted in the last dimension
  --> file.f90:2:25
   |
2  |     real, intent(in) :: a(*, 10)
   |                           ^

semantic error: Dummy argument 'a' not defined
  --> file.f90:1:1 - 3:14
   |
1  |    subroutine assumed_size_star_pos_1(a)
   |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

Issues:
1. The second error highlights the entire subroutine instead of just `a`
2. The message "not defined" is misleading - `a` IS defined, just incorrectly

Suggestion: suppress the second error when the first error already identifies the problem.

From PR #9073 review.

opixdown:
fixes #9105

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -2142,7 +2142,6 @@ public:
 	                "Assumed-size '*' is only permitted in the last dimension",
 	                diag::Level::Error, diag::Stage::Semantic, {
 	                    diag::Label("", {m_dim[i].loc})}));
-	            throw SemanticAbort();
 	        }
 	        ASR::dimension_t dim_dummy; dims.push_back(al, dim_dummy);
 	        ASR::dimension_t &dim  = const_cast<ASR::dimension_t&>(dims[dims.size()-1]);
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "4a7cb161af7e459768210ff6aaaa9e985f451b19fd35dedfb2f61472",
+    "stderr_hash": "836f0339a329130c5131a482b3ed031162513765a240f4a13f8c385b",
     "returncode": 1
 }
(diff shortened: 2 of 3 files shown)


HarshitaKalani (commented):
I think this will cause problem with the --continue-compilation feature.
For the following program,
```
subroutine assumed_size_star_pos_1(a)
    real, intent(in) :: a(*, 10) 
    real, intent(in) :: b(*, 10)  ! wrong: * must be last
    print *, min(1, hujkb)
end subroutine
```
the output should be 
```
 lfortran a.f90
semantic error: Assumed-size '*' is only permitted in the last dimension
 --> a.f90:2:27
  |
2 |     real, intent(in) :: a(*, 10) 
  |                           ^ 


```
whereas it is the following with this PR:
```
% lfortran a.f90
semantic error: Assumed-size '*' is only permitted in the last dimension
 --> a.f90:2:27
  |
2 |     real, intent(in) :: a(*, 10) 
  |                           ^ 

semantic error: Assumed-size '*' is only permitted in the last dimension
 --> a.f90:3:27
  |
3 |     real, intent(in) :: b(*, 10)  ! wrong: * must be last
  |                           ^ 

semantic error: Variable 'hujkb' is not declared
 --> a.f90:4:21
  |
4 |     print *, min(1, hujkb)
  |                     ^^^^^ 'hujkb' is undeclared


```

opixdown:
@HarshitaKalani  agreed, i am thinking of adding a flag when the error is detected then continue processing all 
dimensions normally so the argument gets registered, then throw SemanticAbort()  at the end of the function.
is this correct ?

HarshitaKalani:
I think this may conflict with --continue-compilation. That flag already controls whether we report multiple errors or abort just after the first one, removing SemanticAbort() here seems to cause cascading diagnostics.
Adding a local flag feels like working against the intent of continue-compilation and could make error handling harder to reason about.

PR #9255, opened 2025-12-29
runtime: accept T/F for logical input

krystophny:
## Summary

Accept single-character `T`/`F` (case-insensitive) for logical input, in addition to the existing `.true.`/`.false.` and `true`/`false` formats.

## Motivation

Fortran standard allows `T`/`F` for logical input. This is required for running LAPACK test suite which uses `T`/`F` in input files.

## Changes

- `runtime/impure/lfortran_intrinsics.c`: Extend logical input parsing to accept `T`/`F`

## Test Plan

- [x] `read_08.f90` - reads logical values using T/F format

## Merge Order

**Merge first** - #9261 and #9258 depend on this for test numbering.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -765,6 +765,7 @@ RUN(NAME select_rank_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 
 RUN(NAME global_allocatable_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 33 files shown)


HarshitaKalani:
Hey @krystophny , thanks for the PR! Can you please add a test for the same?

krystophny:
Added integration test `read_logical_01` that verifies T/F (case-insensitive) can be read as logical values. Thanks for the review!

integration_tests/CMakeLists.txt
```diff
@@ -2289,6 +2289,7 @@ RUN(NAME read_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME read_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc COPY_TO_BIN read_07_data.txt)
+RUN(NAME read_logical_01 LABELS gfortran llvm)
```
certik: Let's be consistent with the test namings
```suggestion
RUN(NAME read_08 LABELS gfortran llvm)
```

krystophny: Renamed in order together with #9258 (read_09) and #9261 (read_10). With all the PRs currently filed, LAPACK compiles now with TESTING enabled. But not all tests pass yet, so there will be a few more. The whole read/write stuff is for the I/O of LAPACK tests which is quite antique but valid.

HarshitaKalani (approved):
Looks great to me. Thanks!

PR #9262, opened 2025-12-29
fix: convert class arrays to struct type correctly in function arguments

### LINKED ISSUE #7773: Handle `class(*)` array passed to function in select type statement to function which expects concrete type

jinangshah21:
Fixes #7773


certik:
Let's make sure all the tests run.

certik:
Looks like they will run.

PR #9264, opened 2025-12-29
fix: string concat uses declared length for CHARACTER*1 dummies

### LINKED ISSUE #9263: codegen: string concat uses actual length instead of declared length for CHARACTER*1
## Summary

When a `CHARACTER*1` (or `CHARACTER`) dummy argument receives a longer string literal, string concatenation uses the actual passed length instead of the declared length, causing buffer overflows.

## MRE Branch

**Branch**: [`issue/string-concat-char1-overflow`](https://github.com/krystophny/lfortran/tree/issue/string-concat-char1-overflow)

**Test file**: `integration_tests/string_concat_char1_01.f90`

```fortran
program string_concat_char1_01
    implicit none
    call test_concat('Upper', 'Diag')
    print *, "PASS"
contains
    subroutine test_concat(a, b)
        character, intent(in) :: a, b  ! Both are CHARACTER*1
        character(2) :: result
        ! a should be 'U', b should be 'D', result should be 'UD'
        ! Bug: uses len('Upper')=5 + len('Diag')=4 = 9 chars
        result = a // b
        if (result /= 'UD') error stop "result should be 'UD'"
    end subroutine
end program
```

**Results**:
- gfortran: `len(a)=1, len(b)=1, result='UD'` → **PASS**
- lfortran: `malloc(): corrupted top size` → **CRASH**

### 1. ASR is correct

The ASR shows CHARACTER*1 with explicit constant length:
```
a: (Variable ... (String 1 (IntegerConstant 1 ...) ExpressionLength DescriptorString))
b: (Variable ... (String 1 (IntegerConstant 1 ...) ExpressionLength DescriptorString))
```

### 2. `get_string_length()` has correct logic

In `asr_to_llvm.cpp:669-672`:
```cpp
if(str->m_len && ASR::is_a<ASR::IntegerConstant_t>(*str->m_len)){
    ASR::IntegerConstant_t* len = ASR::down_cast<ASR::IntegerConstant_t>(str->m_len);
    return llvm::ConstantInt::get(context, llvm::APInt(64, len->m_n));
}
```

### 3. The bug: `_lcompilers_stringconcat` uses runtime lengths

In `intrinsic_functions.h:5187-5193`:
```cpp
fill_func_arg("s1", b.String(nullptr, ASR::AssumedLength))
fill_func_arg("s2", b.String(nullptr, ASR::AssumedLength))
// Result length computed from runtime StringLen:
b.String(b.Add(b.StringLen(args[0]), b.StringLen(args[1])), ASR::ExpressionLength)
```

The `_lcompilers_stringconcat` function uses `AssumedLength` for parameters, so it reads lengths from runtime descriptors. When called from code like `result = a // b`, the descriptors for `a` and `b` still contain their original lengths (5 and 4 from 'Upper' and 'Diag') instead of the declared CHARACTER*1 length (1).

### 4. Missing: descriptor length adjustment in call setup

When passing strings to subroutines, `convert_call_args()` doesn't adjust the string descriptor length to match the dummy argument's declared length. The original descriptor (with length 5) is passed directly, instead of creating a view with length 1.

## Related Issues

- #8620 - argument strings have issue with variable length (same underlying descriptor length mismatch)
- #6892 - Progress: Refactor string and string array (LLVM backend) - ongoing string handling refactoring
- #5801 - array of characters doesn't do actual stringcopying (pointer vs copy semantics)
- #346 - Implement Assumed Length (foundational assumed-length character support)

## Impact

This blocks running LAPACK test suite. LAPACK code pattern:
```fortran
CALL STRTRI('Upper', 'Non-unit', N, A, LDA, INFO)

SUBROUTINE STRTRI(UPLO, DIAG, ...)
  CHARACTER UPLO, DIAG  ! CHARACTER*1
  NB = ILAENV(1, 'STRTRI', UPLO // DIAG, N, -1, -1, -1)  ! Overflow here
END SUBROUTINE
```

## Proposed Fix

Two potential approaches:

1. **At call site**: When passing a string to a CHARACTER(n) dummy, create a new descriptor with `len = min(actual_len, n)` and same data pointer

2. **In `_lcompilers_stringconcat`**: Instead of using `AssumedLength`, compute result length from the ASR type's declared length when it's a constant

The first approach is more general as it would fix all operations on the dummy argument, not just concatenation.

krystophny:
## Summary
- Fix string concatenation buffer overflow when CHARACTER*1 dummies receive longer actual arguments
- Pass declared type lengths as explicit parameters to `_lcompilers_stringconcat` instead of reading from descriptor at runtime

## Details
When passing a longer string (e.g., `'Upper'`) to a `CHARACTER(1)` dummy parameter, the string descriptor retains the actual argument length (5), not the declared length (1). The previous implementation used `StringLen` to read from the descriptor at runtime, producing incorrect results and buffer overflows.

The fix modifies `instantiate_StringConcat` in `intrinsic_functions.h` to:
1. Check if the declared length (`m_len`) is known from the ASR type
2. If known, pass it as an explicit parameter to the concat function
3. Otherwise, fall back to `StringLen` for runtime determination

## Test plan
- [x] MRE test added: `integration_tests/string_concat_char1_01.f90`
- [x] Reference tests pass (string_03 reference updated for new LLVM IR)
- [x] LAPACK (Reference-LAPACK v3.12.0, BUILD_TESTING=ON) builds 100% and passes several test suites

Fixes #9263.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1909,6 +1909,7 @@ RUN(NAME string_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_88 LABELS llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays) #remove gfortran label, as functionality gave incorrect results
 RUN(NAME string_89 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME string_90 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME string_concat_char1_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # #9263 concat with CHARACTER*1 dummies
 
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -5172,54 +5172,66 @@ namespace StringConcat {
 
     
     inline ASR::expr_t* instantiate_StringConcat(Allocator &al, const Location &loc,
-        SymbolTable *scope, Vec<ASR::ttype_t*>& /*arg_types*/, ASR::ttype_t* /*return_type*/,
+        SymbolTable *scope, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t* /*return_type*/,
         Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/){
+        // Fix for #9263: Pass explicit lengths instead of using descriptor lengths
+        // When CHARACTER*1 dummy receives longer actual (e.g., 'Upper'), the descriptor
+        // still carries the actual argument length (5), not the declared length (1).
+        // StringLen reads from descriptor, giving wrong result. Pass declared lengths explicitly.
         char intrinsic_fn_name[] = "_lcompilers_stringconcat";
-        if(ASR::symbol_t* f_sym = scope->resolve_symbol(intrinsic_fn_name)){ //Avoid duplication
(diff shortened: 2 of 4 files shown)


krystophny:
Superseded by branch fix/string-concat-char1-length with correct test naming (string_91.f90)

PR #9265, opened 2025-12-29
semantics: fix string concatenation for CHARACTER*1 dummies

### LINKED ISSUE #9263: codegen: string concat uses actual length instead of declared length for CHARACTER*1

krystophny:
## Summary

Fixes #9263

Fix string concatenation buffer overflow when concatenating CHARACTER*1 (declared with single character specifier) arguments.

**Before**: `result = a // b` with CHARACTER*1 arguments would use `len(a)` which returned the full string length passed by caller (e.g., 5 for "Upper"), causing buffer overflow.

**After**: Use `type->m_len` (declared length = 1) for allocation size instead of `ASR::StringLen` which returns actual string length.

## Changes

- `libasr/pass/intrinsic_functions.h`: Use declared length from type for CHARACTER*1 dummies in StringConcat

## Test Plan

- [x] `string_91.f90` - test concatenation with CHARACTER*1 dummies
- Reference test `string_03` updated to reflect correct behavior

### DIFF
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -5171,22 +5171,42 @@ namespace StringConcat {
     }
 
     
+    // Compute string length expression without embedding nested StringConcats.
+    // For StringConcat expressions, recursively compute len(arg1) + len(arg2).
+    // For other expressions, use StringLen directly.
+    inline ASR::expr_t* get_safe_string_len(Allocator &al, const Location &loc,
+            ASR::expr_t* expr, ASR::ttype_t* expr_type, ASRBuilder& b) {
+        ASR::String_t* str_type = ASR::down_cast<ASR::String_t>(ASRUtils::extract_type(expr_type));
+        if (str_type->m_len) {
+            return b.i2i_t(str_type->m_len, ASRUtils::TYPE(ASR::make_Integer_t(al, loc, 4)));
+        }
+        // For deferred-length strings, check if expr is a StringConcat.
+        // If so, recursively compute the length to avoid embedding the
+        // StringConcat expression in a StringLen node (which causes infinite
+        // recursion when the visitor processes the call arguments).
+        if (ASR::is_a<ASR::IntrinsicElementalFunction_t>(*expr)) {
+            ASR::IntrinsicElementalFunction_t* ief = ASR::down_cast<ASR::IntrinsicElementalFunction_t>(expr);
+            if (ief->m_intrinsic_id == static_cast<int64_t>(IntrinsicElementalFunctions::StringConcat)) {
+                ASR::expr_t* len1 = get_safe_string_len(al, loc, ief->m_args[0],
+                    ASRUtils::expr_type(ief->m_args[0]), b);
+                ASR::expr_t* len2 = get_safe_string_len(al, loc, ief->m_args[1],
+                    ASRUtils::expr_type(ief->m_args[1]), b);
+                return b.Add(len1, len2);
+            }
+        }
+        return b.StringLen(expr);
+    }
+
     inline ASR::expr_t* instantiate_StringConcat(Allocator &al, const Location &loc,
         SymbolTable *scope, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t* /*return_type*/,
         Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/){
         char intrinsic_fn_name[] = "_lcompilers_stringconcat";
         declare_basic_variables(intrinsic_fn_name)
 
-        ASR::String_t* s1_type = ASR::down_cast<ASR::String_t>(ASRUtils::extract_type(arg_types[0]));
-        ASR::String_t* s2_type = ASR::down_cast<ASR::String_t>(ASRUtils::extract_type(arg_types[1]));
-
-        // Use declared type length only if it is a compile-time constant
-        ASR::expr_t* s1_len_arg = (s1_type->m_len && ASRUtils::is_value_constant(s1_type->m_len))
-            ? b.i2i_t(s1_type->m_len, int32)
-            : b.StringLen(new_args[0].m_value);
-        ASR::expr_t* s2_len_arg = (s2_type->m_len && ASRUtils::is_value_constant(s2_type->m_len))
-            ? b.i2i_t(s2_type->m_len, int32)
-            : b.StringLen(new_args[1].m_value);
+        // Compute argument lengths safely, avoiding circular references for
+        // nested StringConcat expressions.
+        ASR::expr_t* s1_len_arg = get_safe_string_len(al, loc, new_args[0].m_value, arg_types[0], b);
+        ASR::expr_t* s2_len_arg = get_safe_string_len(al, loc, new_args[1].m_value, arg_types[1], b);
 
         // Build call_args with explicit lengths
         Vec<ASR::call_arg_t> call_args;


certik:

src/libasr/pass/intrinsic_functions.h
```diff
@@ -5172,54 +5172,66 @@ namespace StringConcat {
...
-        
         declare_basic_variables(intrinsic_fn_name)
 
-        /* Args */
+        ASR::String_t* s1_type = ASR::down_cast<ASR::String_t>(ASRUtils::extract_type(arg_types[0]));
+        ASR::String_t* s2_type = ASR::down_cast<ASR::String_t>(ASRUtils::extract_type(arg_types[1]));
+
+        // Use declared type length (m_len) if known, else fall back to runtime StringLen
+        // All lengths use int32 to match StringLen return type
+        ASR::expr_t* s1_len_arg = s1_type->m_len
```
assem2002: let's Use the `len` expression only when it's a compile-time expression -- use utility `ASR::is_value_constant`.
If it's an expression that needs runtime evaluation (e.g. function call) -- it wouldn't be the behavior we want (calling the function again), though it fortran a runtime length expression should be pure -- but let's just not do that.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -5172,54 +5172,66 @@ namespace StringConcat {
...
-        /* Body */
-        /*
-            function lcompilers_stringconcat(s1, s2) result(concat_result)
-                character(*) :: s1
-                character(*) :: s2
-                character(len(s1) + len(s2)) :: concat_result
-                concat_result(1 : len(s1)) = s1
-                concat_result(len(s1) + 1 : len(concat_result)) = s2
-            end function
-        */
```
assem2002, line 5205: can we keep that?
It needs modification btw according to the new signature.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -5172,54 +5172,66 @@ namespace StringConcat {
 
     
     inline ASR::expr_t* instantiate_StringConcat(Allocator &al, const Location &loc,
-        SymbolTable *scope, Vec<ASR::ttype_t*>& /*arg_types*/, ASR::ttype_t* /*return_type*/,
+        SymbolTable *scope, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t* /*return_type*/,
         Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/){
+        // Fix for #9263: Pass explicit lengths instead of using descriptor lengths
+        // When CHARACTER*1 dummy receives longer actual (e.g., 'Upper'), the descriptor
+        // still carries the actual argument length (5), not the declared length (1).
+        // StringLen reads from descriptor, giving wrong result. Pass declared lengths explicitly.
```
assem2002: let's remove this.

src/libasr/pass/intrinsic_functions.h
krystophny: done!

src/libasr/pass/intrinsic_functions.h
krystophny: kept and modified for new 4-arg signature

src/libasr/pass/intrinsic_functions.h
krystophny: done!

certik:
Not sure what is going on with the CI failure:
```
[100%] Project compiled successfully.
++ fpm --compiler=lfortran test --flag '--cpp --realloc-lhs-arrays'
[  0%]                  testdrive.F90
[  9%]                  testdrive.F90  done.
[  9%]                    libtoml-f.a
[ 18%]                    libtoml-f.a  done.
[ 18%]                      utils.f90
[ 27%]                      utils.f90  done.
[ 27%]                      build.f90
[ 36%]                      build.f90  done.
[ 36%]                      lexer.f90
[ 45%]                      lexer.f90  done.
[ 45%]                     parser.f90
[ 54%]                     parser.f90  done.
[ 54%]                       sort.f90
[ 63%]                       sort.f90  done.
[ 63%]                      json2toml
[ 72%]                      json2toml  done.
[ 72%]                      toml2json
[ 81%]                      toml2json  done.
[ 81%]                       main.f90
Error: Process completed with exit code 143.
```
Almost looks like a segfault in LFortran itself.

certik:
Ok, the failure is real. It's either caused by this PR, or (hopefully not) by a failure in main.

krystophny:
> Ok, the failure is real. It's either caused by this PR, or (hopefully not) by a failure in main.

Yes this was some infinite recursion. Hope this is resolved now.

certik (approved):

Issue #9222, opened 2025-12-27
bug: `SemanticAbort` caught at translation unit level for errors in modules


kmr-srbh:
`SemanticAbort` thrown in this case, and some other errors inside `visit_Function()` also, are being caught by the `catch` block inside `visit_TranslationUnit()`. This happens with functions defined inside modules:

https://github.com/lfortran/lfortran/blob/233f6a48c86adb2fea9bccd32c5ffdb005d58772/src/lfortran/semantics/ast_symboltable_visitor.cpp#L71-L115
post which the assignment `tmp = tmp0` resets the translation unit to a dummy empty translation unit with no symbols. The fix is to ensure these errors are caught and processed inside `visit_Function()` only.

Turns out, this is part of a bigger issue, where errors are being caught late at the translation unit level. Here is another example:

### With `module` (segfault)
```fortran
module test_crash
    implicit integer(a-f), real(e-z)
end module
```
```console
(lf) saurabhkumar@Mac lfortran % lfortran test.f90                       
semantic error: Implicit typing is not allowed, enable it by using --implicit-typing 
 --> test.f90:2:5
  |
2 |     implicit integer(a-f), real(e-z)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
(lf) saurabhkumar@Mac lfortran % lfortran test.f90 --continue-compilation
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2805
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2732
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 1193
    t2 = std::chrono::high_resolution_clock::now();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
[...]

PR #9266, opened 2025-12-29
fix: fixed the compiler crash in SymbolTableVisitor and added a test

ANAS727189:
**Fixes:** #9222 

This PR fixes a compiler crash that occurred when semantic errors were encountered inside module, function, or subroutine blocks while using the --continue-compilation flag.

Previously, if a SemanticAbort was thrown inside these visitors, the scope management logic was interrupted, leaving current_scope pointing to an invalid or unfinished symbol table. This caused the compiler to crash or segfault when attempting to process subsequent errors.

Basically I added try-catch blocks in ast_symboltable_visitor.cpp and also added a regression test case for interface mismatches in tests/errors/continue_compilation_1.f90

Input: test_program.f90 - 
```
program test_crash
    implicit integer(a-f), real(e-z)
end program
```
Output - 
```
(lf) $ src/bin/lfortran test_program.f90 --continue-compilation
semantic error: Implicit typing is not allowed, enable it by using --implicit-typing 
 --> test_program.f90:2:5
  |
2 |      implicit integer(a-f), real(e-z)
  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Input - test_crash.f90 - 
```
module test_crash
    implicit integer(a-f), real(e-z)
end module
```
Output - 
```
(lf) $ src/bin/lfortran -c test_crash.f90 --continue-compilation
semantic error: Implicit typing is not allowed, enable it by using --implicit-typing 
 --> maintainer_crash.f90:2:5
  |
2 |      implicit integer(a-f), real(e-z)
  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Input - test_fix.f90 
```
module test_crash
    implicit none
    interface
        module function f() result(i)
            integer :: i
        end function
    end interface
contains
    module function f() result(r)
        real :: r
        r = 5.5
    end function
end module
```

Output - 
```
(lf) $ src/bin/lfortran test_fix.f90 --continue-compilation
semantic error: Argument(s) or return type mismatch in interface and implementation
  --> test_fix.f90:9:5 - 12:16
   |
 9 |         module function f() result(r)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
12 |         end function
   | ...^^^^^^^^^^^^^^^^ 

semantic error: Variable 'r' is not declared
  --> test_fix.f90:11:9
   |
11 |          r = 5.5
   |          ^ 'r' is undeclared
```

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -285,95 +285,105 @@ public:
         class_procedures.clear();
         SymbolTable *parent_scope = current_scope;
         current_scope = al.make_new<SymbolTable>(parent_scope);
-        current_module_dependencies.reserve(al, 4);
-        generic_procedures.clear();
-        ASR::asr_t *tmp0 = nullptr;
-        if( x.class_type == AST::modType::Submodule ) {
-            ASR::symbol_t* submod_parent = (ASR::symbol_t*)(ASRUtils::load_module(al, global_scope,
-                                                parent_name, x.base.base.loc, false, loaded_submodules,
-                                                compiler_options.po, true,
-                                                [&](const std::string &msg, const Location &loc) {
-                                                    diag.add(diag::Diagnostic(
-                                                        msg, diag::Level::Error, diag::Stage::Semantic, {
-                                                            diag::Label("", {loc})}));
-                                                    throw SemanticAbort();}, lm, compiler_options.separate_compilation
-                                                ));
-            ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(submod_parent);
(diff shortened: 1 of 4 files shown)


certik:
Thanks for fixing it!

I don't like all these try/except clauses everywhere. I need to think some more about this, if this is the right approach.

ANAS727189:
Thanks @certik . I agree that putting try-catch blocks isn't ideal for readability. I did it as I thought it would be better for fixing the crash.
Though a cleaner fix surely can be made using scope guards but I thought it would be slightly complex.

certik:
Let's keep this PR open. But let's figure out a solution without try/catch all over the place. If no other solution is better, we can come back to this PR and merge it.

PR #9269, opened 2025-12-29
fix: treat common block string members as allocatable in strcpy

krystophny:
## Summary

Fix "Non-allocatable string isn't allocated" error when passing string literals to `CHARACTER(*)` parameters in separate compilation with common blocks containing CHARACTER variables.

## Problem

When a subroutine has a common block with a CHARACTER variable and passes a string literal to another subroutine with a `CHARACTER(*)` parameter, the LLVM backend crashes with:

```
Non-allocatable string isn't allocated
```

This occurs because common block structs are initialized with `zeroinitializer`, which sets string descriptor pointers to NULL. The `_lfortran_strcpy` runtime function then fails when it encounters the NULL pointer for non-allocatable strings.

## Fix

In `asr_to_llvm.cpp`, treat `StructInstanceMember` targets (which includes common block variables) as allocatable when calling `lfortran_str_copy`. This allows the runtime to allocate memory for the string descriptor if the pointer is NULL.

## Test

Added `separate_compilation_28` integration test that reproduces the exact scenario:
- Main program calls subroutines
- Subroutines have common block with CHARACTER variable
- Subroutines pass string literals to a handler with `CHARACTER(*)` parameter


krystophny:
Superseded by #9270 which includes this fix plus the CommonLinkage changes and declared length storage.

PR #9274, opened 2025-12-30
perf: use LLVM intrinsics for abs/max/min instead of function calls

krystophny:
## Summary

Optimize critical numerical intrinsics by using efficient LLVM codegen instead of
generating function calls with branching logic.

### Changes

**abs()**
- Real types: use `llvm.fabs` intrinsic (single instruction)
- Integer types: use branchless `select` with `icmp`
- Complex types: unchanged (sqrt formula)

**max()/min()**
- Real types: use branchless `FCmpOGT`/`FCmpOLT` + `select`
- Integer types: use branchless `select` with `icmp`
- String types: unchanged (function with branching)

### NaN Handling

For max/min, we use `FCmpOGT`/`FCmpOLT` comparisons instead of `llvm.maxnum`/`llvm.minnum`
intrinsics. This preserves Fortran's comparison-based NaN propagation semantics:
- `max(NaN, x) = x` (NaN > x is false)
- `max(x, NaN) = NaN` (x > NaN is false, result unchanged)

The LLVM intrinsics have IEEE 754-2008 maxNum semantics which suppress NaN propagation,
causing LAPACK SQK test failures (36,885 out of 241,365 tests). With `FCmp` + `select`,
all SQK tests pass.

### Performance Impact

Profiling LAPACK tests showed:
- `_lcompilers_abs_f32`: **15%** of runtime before fix
- `_lcompilers_max0_f32`: **37.84%** of runtime before fix

After this change, these functions are inlined as branchless comparisons,
eliminating function call overhead entirely.

### Test Plan

- [x] All existing abs tests pass (`abs_01.f90`, `abs_03.f90`)
- [x] All max/min tests pass (`max_01.f90`, `max_02.f90`, `min_01.f90`, `min_02.f90`)
- [x] LAPACK SQK tests pass (241,365 tests)
- [x] Real, integer, and complex types verified

### DIFF
--- a/src/libasr/codegen/asr_to_wasm.cpp
+++ b/src/libasr/codegen/asr_to_wasm.cpp
@@ -2645,23 +2645,24 @@ class ASRToWASMVisitor : public ASR::BaseVisitor<ASRToWASMVisitor> {
                     }
                 } else if (ASRUtils::is_integer(*arg_type)) {
                     // abs(x) = x >= 0 ? x : -x (branchless with select)
-                    // Stack order for select: [if_false, if_true, condition]
+                    // Stack order for select: [val1, val2, cond]
+                    // Returns val1 if cond != 0, else val2
                     if (kind == 4) {
                         m_wa.emit_global_set(m_compiler_globals[tmp_reg_i32]);
+                        m_wa.emit_global_get(m_compiler_globals[tmp_reg_i32]);  // x (val1)
                         m_wa.emit_i32_const(0);
                         m_wa.emit_global_get(m_compiler_globals[tmp_reg_i32]);
-                        m_wa.emit_i32_sub();  // -x (if_false)
-                        m_wa.emit_global_get(m_compiler_globals[tmp_reg_i32]);  // x (if_true)
+                        m_wa.emit_i32_sub();  // -x (val2)
                         m_wa.emit_global_get(m_compiler_globals[tmp_reg_i32]);
                         m_wa.emit_i32_const(0);
                         m_wa.emit_i32_ge_s();  // x >= 0 (condition)
                         m_wa.emit_select();
                     } else if (kind == 8) {
                         m_wa.emit_global_set(m_compiler_globals[tmp_reg_i64]);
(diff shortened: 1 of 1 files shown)


certik:
@assem2002 do you mind reviewing this one please?

src/libasr/pass/intrinsic_functions.h
```diff
@@ -833,7 +868,18 @@ namespace Abs {
 
     static inline ASR::expr_t* instantiate_Abs(Allocator &al, const Location &loc,
             SymbolTable *scope, Vec<ASR::ttype_t*>& arg_types, ASR::ttype_t *return_type,
-            Vec<ASR::call_arg_t>& new_args, int64_t /*overload_id*/) {
+            Vec<ASR::call_arg_t>& new_args, int64_t overload_id) {
+        // For integer/real without FunctionCall in args: use IntrinsicElementalFunction
+        // which gets lowered to LLVM fabs/select intrinsics
+        if ((is_integer(*arg_types[0]) || is_real(*arg_types[0])) &&
+            !contains_function_call(new_args[0].m_value)) {
```
assem2002, line 875: Why can't we use the new-adopted approach with 'functionCall' args?
I think with LLVM backend should work properly, Is there a problem with other backends?

src/libasr/pass/intrinsic_functions.h
```diff
@@ -221,6 +221,41 @@ typedef void (*verify_function)(
 
 typedef ASR::expr_t* (*get_initial_value_func)(Allocator&, ASR::ttype_t*);
 
+// Check if expression contains FunctionCall - used to determine if we need
+// FunctionCall wrapper (for proper pass handling) vs direct IntrinsicElementalFunction
+static inline bool contains_function_call(ASR::expr_t* expr) {
```
assem2002, line 226: This should be a visitor -- using a switch statement usually breaks.
I think it's fine for now.

assem2002 (approved):
looks good. The changes are useful to performance for sure. Thanks!
I don't have experience with wasm backend -- if CI passes, I think it's fine then.
Please address my comments, If you think we got no issue then let's merge.

certik (approved):
I think it's good enough to merge then.

PR #9277, opened 2025-12-30
fix: Add implicit typing support for implied-do loop variables

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9008


certik:
Once quick tests run, let's also run full tests.

PR #9283, opened 2025-12-30
fix: fix assertion failure in type_to_str_fortran_expr when expr is null

### LINKED ISSUE #8953: Assertion in type_to_str_fortran_expr with StructType + cpp macros
# Bug: type_to_str_fortran_expr assertion with StructType + cpp macros

## Summary

On current `upstream/main` (LLVM 11 build in the `lf-llvm11` environment), compiling a small preprocessed module that defines a `type` via a `#define` macro and then using that type from another module causes an internal assertion failure in `type_to_str_fortran_expr`:

```
LCOMPILERS_ASSERT failed: src/libasr/asr_utils.h
function type_to_str_fortran_expr(), line number 992 at 
expr != nullptr
ERROR MESSAGE:
`expr` should be non-null for `StructType`
```

This happens both for `--show-errors --continue-compilation` and for `--lookup-name` on the main program that uses the chain of modules.

## Environment

- Branch: `upstream/main` (tested at `0.58.0-180-g2fea229b9-dirty` from `run_inside.sh`)
- Toolchain: `lf-llvm11` micromamba env as in Quick-Checks CI
- LLVM: 11.1.0
- Command path: `run_inside.sh` (CI-style build) + in-tree `lfortran` from that env.

## Reproducer

Create the following three files (e.g., in `/tmp/lfortran-pre-chain/`).

`mod_pre_chain_a.f90`:
```fortran
#define FIELDS integer :: a ; real :: b

module mod_pre_a
  implicit none
  type :: T
    FIELDS
  end type T
end module mod_pre_a
```

`mod_pre_chain_b.f90`:
```fortran
module mod_pre_b
  use mod_pre_a, only: T
  implicit none
contains
  subroutine s(x)
    type(T), intent(in) :: x
  end subroutine s
end module mod_pre_b
```

`main_pre_chain.f90`:
```fortran
program main_pre_chain
  use mod_pre_b, only: s
  implicit none
  integer :: r
  call s(r)
end program main_pre_chain
```

Then, using the LLVM 11 build from the `lf-llvm11` env (after `run_inside.sh`):

```bash
cd /tmp/lfortran-pre-chain
lfortran --cpp -c mod_pre_chain_a.f90
lfortran -c mod_pre_chain_b.f90
lfortran --show-errors --continue-compilation --no-color main_pre_chain.f90
```

Expected:
- A normal semantic error (type mismatch in the call to `s`) or a clear diagnostic, but no internal assertion.

Actual:
- LFortran aborts with:

```text
LCOMPILERS_ASSERT failed: .../src/libasr/asr_utils.h
function type_to_str_fortran_expr(), line number 992 at 
expr != nullptr
ERROR MESSAGE:
`expr` should be non-null for `StructType`
```

The same assertion is hit if you instead run:

```bash
lfortran --lookup-name --no-color --line=5 --column=10 main_pre_chain.f90
```

after compiling `mod_pre_chain_a.f90` and `mod_pre_chain_b.f90` as above.

## Notes

- The bug reproduces on a clean `upstream/main` build (no local patches), using the CI-style `lf-llvm11` environment.
- The assertion is about pretty-printing a `StructType` in `type_to_str_fortran_expr`, so this appears to be an ASR printing / diagnostics bug rather than a location-manager issue.

ANAS727189:
Fixes #8953

This addresses the segmentation fault in type_to_str_fortran_expr that occurred when dealing with a StructType alongside a nullptr expression typically triggered during type mismatch errors involving derived types.

I added a check to handle the null case safely; if expr is missing, the function now falls back to returning "type(derived)" instead of crashing. This prevents the compiler abort and ensures it prints the actual diagnostic message as intended.

The compiler no longer crashes and correctly outputs diagnostics for the reproduction case provided in the issue.
```
(lf) anas@anas-mint:/lfortran --cpp -c mod_pre_chain_a.f90
(lf) anas@anas-mint:/lfortran -c mod_pre_chain_b.f90
(lf) anas@anas-mint:/lfortran --show-errors --continue-compilation --no-color main_pre_chain.f90
{"uri":"file://main_pre_chain.f90","diagnostics":[{"source":"lpyth","range":{"start":{"line":5,"character":10},"end":{"line":5,"charac(lf) 
anas@anas-mint:/lfortran --lookup-name --no-color --line=5 --column=10 main_pre_chain.f90
```

### DIFF
--- a/tests/errors/issue_8953.f90
+++ b/tests/errors/issue_8953.f90
@@ -1,0 +1,24 @@
+#define FIELDS integer :: a ; real :: b
+
+module mod_pre_a
+  implicit none
+  type :: T
+    FIELDS
+  end type T
+end module mod_pre_a
+
+module mod_pre_b
+  use mod_pre_a, only: T
+  implicit none
+contains
+  subroutine s(x)
+    type(T), intent(in) :: x
+  end subroutine s
+end module mod_pre_b
+
+program main_pre_chain
+  use mod_pre_b, only: s
+  implicit none
+  integer :: r
+  call s(r)
+end program main_pre_chain
--- a/tests/reference/asr-issue_8953-6b39f0d.stderr
+++ b/tests/reference/asr-issue_8953-6b39f0d.stderr
@@ -1,0 +1,5 @@
+semantic error: Type mismatch in argument `x`: expected `type(derived)` but got `integer`
+  --> tests/errors/issue_8953.f90:23:10
+   |
+23 |   call s(r)
+   |          ^ 
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -4924,3 +4924,8 @@ llvm = true
 [[test]]
 filename = "../integration_tests/transfer_09.f90"
 asr = true
+
+[[test]]
+filename = "errors/issue_8953.f90"
+asr = true
+options = "--cpp"


kmr-srbh (changes requested):
I think this is not the correct change. In a well functioning code, we should not get `expr` as `nullptr`. @ANAS727189 could you please try debugging more to see why it is a `nullptr`? One possible reason can be output of `ASRUtils::get_struct_sym_from_struct_expr()` being passed to the function call. In some cases, the output of the above call is `nullptr`. We can handle it to return a non-null value instead. In the same case, this can be very well due to the derived type symbol being an `ExternalSymbol` due to the chain of modules.

PR #9284, opened 2025-12-30
perf: optimize intrinsic function instantiation

### LINKED ISSUE #9285: perf: intrinsic function instantiation creates duplicate functions
## Problem

Many intrinsic function instantiation functions in `src/libasr/pass/intrinsic_functions.h` do not check if the function already exists before creating it. This causes duplicate functions to be created with incrementing names.

### How it happens

The `declare_basic_variables` macro calls `get_unique_name()`:

```cpp
std::string fn_name = scope->get_unique_name(name, false);
```

`get_unique_name()` returns incrementing names if the original exists:
- First call: `_lcompilers_sign_i32`
- Second call: `_lcompilers_sign_i321`
- Third call: `_lcompilers_sign_i322`

Functions without existence checks create a NEW identical function every time, wasting memory and compilation time.

### Correct pattern

```cpp
// Fast path: reuse existing function
if (ASR::symbol_t *s = scope->get_symbol(func_name)) {
    ASRBuilder b(al, loc);
    return b.Call(s, new_args, ...);
}

// Slow path: create function only once
declare_basic_variables(func_name);
// ... create function body ...
```

### Affected functions

Approximately 80+ functions in `intrinsic_functions.h` are missing the existence check, including:
- Sign, SignFromValue
- Aint, Anint, Ceiling, Floor, Nint, Idnint
- Iand, Ior, Ieor, Not, And, Or, Xor
- Ibits, Ibclr, Ibset, Btest
- Dprod, Dreal, Scale, Exponent, Fraction
- Leadz, Trailz, Popcnt, Poppar, Maskl, Maskr
- Shiftl, Shiftr, Ishft, Dshiftl, Dshiftr
- StringLenTrim, StringTrim, Adjustl, Adjustr
- And many more...

### Related

- #3277 - Refactor ASRBuilder (mentions removing macros)
- #8172 - Caching/reusing ttypes (similar concept)

### Solution

Add existence checks to all affected functions following the correct pattern above.

krystophny:
## Summary

Optimize intrinsic function instantiation by checking if the function already exists BEFORE calling `declare_basic_variables`.

Fixes #9285

## Problem

The `declare_basic_variables` macro creates expensive objects on every call:
- Symbol table allocation
- ASRBuilder object
- Multiple vector allocations

Many intrinsic instantiation functions called this macro BEFORE checking if the function already existed (or never checked at all), wasting resources when the function was already available.

## Solution

Move the existence check before `declare_basic_variables`. In the fast path (function exists), create only a minimal ASRBuilder to build the call.

## Test plan

- [ ] All integration tests pass
- [ ] Build succeeds


krystophny:
Closing after review - the wrapper function generation approach should be reconsidered in favor of direct IR operations or runtime library calls (similar to GFortran's approach) rather than adding complexity with caching/deduplication logic.

PR #9286, opened 2025-12-30
fix: handle Complex literal assigned to Real/Integer PARAMETER

krystophny:
## Summary
LAPACK test files use a legacy GFortran-accepted pattern where REAL parameters are initialized with complex literal syntax:

```fortran
REAL ZERO, ONE
PARAMETER ( ZERO = ( 0.0E+0, 0.0E+0 ), ONE = ( 1.0E+0, 0.0E+0 ) )
```

GFortran accepts this as an extension, extracting the real part. LFortran was parsing the complex literal correctly but then failing in LLVM codegen with type mismatch (`complex_4` stored to `float*`).

## Fix
Handle `ComplexConstant` and `ComplexConstructor` init values when the target variable is Real or Integer by extracting the real part.

## Test Plan
- Added `complex_26.f90` integration test
- Tested with both gfortran and llvm backends


integration_tests/complex_26.f90
```diff
@@ -0,0 +1,13 @@
+! Test: Complex literal assigned to Real PARAMETER
+! LAPACK legacy pattern where REAL parameters use complex literal syntax
+program complex_26
+    implicit none
+    real :: zero, one
+    parameter (zero = (0.0E+0, 0.0E+0), one = (1.0E+0, 0.0E+0))
+
+    ! GFortran accepts this as an extension, extracting the real part
```
certik, line 8: This is not an extension. This is simply casting a complex number to a real number by a simple assignment, which is allowed.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3575,6 +3575,24 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                                 } else if (ASR::is_a<ASR::IntegerConstant_t>(*init_val)) {
                                                     ASR::IntegerConstant_t* ic = ASR::down_cast<ASR::IntegerConstant_t>(init_val);
                                                     init_val = ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, x.base.base.loc, ic->m_n, v->m_type));
+                                                } else if (ASR::is_a<ASR::ComplexConstant_t>(*init_val) ||
+                                                           ASR::is_a<ASR::ComplexConstructor_t>(*init_val)) {
+                                                    // Complex literal assigned to Real/Integer variable
+                                                    // Extract real part (GFortran extension for legacy code)
```
certik: Remove this comment, which I think is not correct:
```suggestion
```

certik:
Why not just casting this from Complex to Real (`ComplexToReal`)? Then everything should work.

certik:
However to get us unblocked, let's merge this PR as-is, if it passes tests.

Then let's iterate on this in subsequent PRs. So let's not push anymore commits here, to give the CI a chance to pass this.

certik:
There are some conflicts, let's resolve them.

krystophny:
Rebased on upstream/main. The refactoring to use Cast nodes (from PR #9292) has been incorporated into this PR. #9292 can be closed.

Issue #9259, opened 2025-12-29
Class Type assoc variable throws ASR Verify Pass Error in Select Rank


HarshilShah1804:
MRE
```fortran
module example_mod
  implicit none
contains
  subroutine check_arg(arg)
    class(*), dimension(..), intent(in) :: arg

    select rank (assoc => arg)
      rank (1)
        print *, "Received an array of rank 1"
    end select
  end subroutine check_arg
end module example_mod

program test
  use example_mod
  implicit none
  integer :: int_array1(3) = [1, 2, 3]

  call check_arg(int_array1)
end program test
```

LFortran
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90
ASR verify pass error: Variable assoc of type StructType must have a type declaration.
  --> a.f90:7:5 - 10:14
   |
 7 |        select rank (assoc => arg)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
10 |        end select
   | ...^^^^^^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

GFortran
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 && ./a.out
 Received an array of rank 1
```

PR #9296, opened 2025-12-31
fix: Set type_decl for select rank associate with class(*)

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9259

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -767,7 +767,7 @@ RUN(NAME select_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME select_rank_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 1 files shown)


assem2002:
failure similar to https://github.com/lfortran/lfortran/pull/9297 -- Check if gfortran version is the problem.

HarshitaKalani:
I don't think this is an issue with gfortran version. The error that we see at the CI is because of wrong fortran code generated with --show-fortran which is then compiled by gfortran and hence the error.
The following is same as what we see on the CI.
```console
 % lfortran a.f90 --show-fortran > p.f90
% gfortran p.f90 && 
./a.out
p.f90:7:11:

    7 |     type(~unlimited_polymorphic_type), dimension(..), intent(in) :: arg
      |           1
Error: Invalid character in name at (1)
p.f90:8:21:

    8 |     select rank (arg)
      |                     1
Error: The SELECT RANK selector at (1) must be an assumed rank variable
p.f90:9:13:

    9 |         rank(1)
      |             1
Error: Unexpected RANK statement at (1)
p.f90:12:13:

   12 |         rank(2)
      |             1
Error: Unexpected RANK statement at (1)
p.f90:15:13:

   15 |         rank default
      |             1
Error: Unexpected RANK statement at (1)
p.f90:18:7:

   18 |     end select
      |       1
Error: Expecting END SUBROUTINE statement at (1)
p.f90:6:22:

    6 | subroutine check_arg(arg)
      |                      1~~
Error: Symbol 'arg' at (1) has no IMPLICIT type
p.f90:10:1:

   10 | assoc => arg
      | 1~~~~
Error: Symbol 'assoc' at (1) has no IMPLICIT type
p.f90:24:5:

   24 | use select_rank_12_mod, only: check_arg
      |     1
Fatal Error: Cannot open module file 'select_rank_12_mod.mod' for reading at (1): No such file or directory
compilation terminated.
```

PR #9299, opened 2025-12-31
fix: mark COMMON blocks declared before processing contained procedures

krystophny:
## Summary
- Fixes a bug where COMMON block variables in contained (internal) procedures saw zeroed values instead of values set by the host program
- The issue was that `mark_common_blocks_as_declared()` was called after processing contained procedures instead of before

## Root Cause
In `visit_Program` in `ast_symboltable_visitor.cpp`, the `mark_common_blocks_as_declared()` call was placed after the loop that processes contained procedures. This caused contained procedures to create new struct members instead of mapping their COMMON block variables to the existing ones by byte offset.

## Changes
- Added `mark_common_blocks_as_declared()` call before the loop processing contained procedures
- Added integration test `common_25` that verifies COMMON block sharing between host and contained procedures

## Verification
```fortran
program common_25
    integer :: val1, val2
    common /testblock/ val1, val2
    val1 = 42
    call modify_common()
    if (val1 /= 99) error stop  ! Previously failed here with LFortran
contains
    subroutine modify_common()
        integer :: v1, v2
        common /testblock/ v1, v2
        v1 = 99  ! Should modify val1 in host
    end subroutine
end program
```

Test passes with both gfortran and llvm backends after fix.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2446,7 +2446,7 @@ RUN(NAME common_23 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_24 LABELS gfortran llvm llvm2
     EXTRAFILES common_24a.f90 common_24b.f90
     EXTRA_ARGS --implicit-interface --separate-compilation)
-RUN(NAME common_25 LABELS gfortran)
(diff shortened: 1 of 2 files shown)


certik (approved):
I think this is fine, if tests pass.

PR #9305, opened 2026-01-01
fix: honor READ END=/ERR= (EOF control flow)

krystophny:
Implements proper END=/ERR= behavior for READ by lowering it via IOSTAT checks.

Adds integration test read_12 reproducing EOF END= label handling.


krystophny:
I looked into the failing CI for this PR and couldn't find any reason why this should be affected. Maybe a glitch and one could re-run CI checks?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1181,8 +1183,90 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                ASR::expr_t* err_expr = ASRUtils::EXPR(tmp);
+                if( !ASR::is_a<ASR::IntegerConstant_t>(*err_expr) ) {
+                    diag.add(Diagnostic(
+                        "`err` must be a literal integer label",
+                        Level::Error, Stage::Semantic, {
+                            Label("",{kwarg.loc})
+                        }));
+                    throw SemanticAbort();
+                }
+                m_err_label = ASR::down_cast<ASR::IntegerConstant_t>(err_expr)->m_n;
```
certik, line 1257: Later: let's add tests for all these error messages.

PR #9310, opened 2026-01-01
fix: honor READ END=/ERR= (EOF/ERR control flow)

krystophny:
Follow-up to #9305 (reverted in #9314).

This reintroduces READ(..., END=, ERR=) control flow cleanly.

Fix:
- For list-directed READ (no explicit FMT), when IOSTAT is present (either user-provided or inserted by the semantic pass for END/ERR), codegen now calls runtime read helpers that accept an iostat pointer.
- Added runtime *_iostat read entry points that set iostat < 0 on EOF and iostat > 0 on read errors, and do not terminate the program when iostat is provided.

Tests:
- Integration regression: integration_tests/read_12.f90 (registered via integration_tests/CMakeLists.txt); locally you can run: lfortran integration_tests/read_12.f90
- Diagnostics coverage consolidated into errors/continue_compilation_1.f90; run: ./run_tests.py -t continue_compilation_1 -s

Continue-compilation:
- Avoid aborting the whole statement during END=/ERR= kwarg validation when `--continue-compilation` is enabled (emit diagnostics and keep going), preventing a compiler crash.

PR #9311, opened 2026-01-01
fix: hoist complex binop scratch allocas to entry block

krystophny:
## Summary

Moves the three temporary alloca instructions used for complex binary
operations (add, sub, mul, div) from the current builder position
to the function entry block. This prevents stack growth when complex
operations occur inside loops.

## Change

Uses the existing `CreateAlloca` overload that hoists to entry block
(without the builder parameter) instead of the one that allocates
at the current insertion point.

## Related

- See #9329 for additional stack optimizations (heap allocation for large arrays)

### DIFF
--- a/tests/reference/llvm-arrays_01_complex-c90dbdd.json
+++ b/tests/reference/llvm-arrays_01_complex-c90dbdd.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-arrays_01_complex-c90dbdd.stdout",
-    "stdout_hash": "29062000e29ed944f4c7b3cfb1a50b7a7590b03e6dbd438fd1b160c7",
+    "stdout_hash": "a5a0df21b3492ab758659a13e0feba6fdd10ef111a1153969026aac9",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-arrays_01_complex-c90dbdd.stdout
+++ b/tests/reference/llvm-arrays_01_complex-c90dbdd.stdout
@@ -105,6 +105,18 @@ source_filename = "LFortran"
 
 define i32 @main(i32 %0, i8** %1) {
 .entry:
+  %2 = alloca %complex_4, align 8
+  %3 = alloca %complex_4, align 8
+  %4 = alloca %complex_4, align 8
+  %5 = alloca %complex_4, align 8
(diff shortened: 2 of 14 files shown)


src/runtime/legacy/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set_target_properties(lfortran_runtime PROPERTIES
 add_library(lfortran_runtime_static STATIC ${SRC})
 target_include_directories(lfortran_runtime_static BEFORE PUBLIC ${libasr_SOURCE_DIR}/..)
 target_include_directories(lfortran_runtime_static BEFORE PUBLIC ${libasr_BINARY_DIR}/..)
-target_link_libraries(lfortran_runtime PRIVATE ${MATH_LIBRARIES})
+target_link_libraries(lfortran_runtime_static PRIVATE ${MATH_LIBRARIES})
```
certik: Why is this needed? I wouldn't touch it, unless needed.

certik:
I think we need to be careful here.

* Let's figure out on a simple example how to test this. The reference changes are too complex for me to see that it actually works. Let's create a simple reference test, and document how it should behave.
* What are "large fixed-size locals"? Local arrays? I thought we already allocate them on a heap --- or is that not the case? Let's separate this case first --- just allocate all local arrays on a heap by default, unless "--stack-arrays" is used.
* What other case is needed to handle --- this `--max-stack-var-size`, is it implemented correctly? I would keep it simple, just an option to control this. I would use stack only for what is truly needed, the rest should go on a heap --- eventually with a custom-optimized runtime based on arenas.

krystophny:
**Next steps per review feedback:**

1. Split this PR into atomic changes:
   - **PR A**: Alloca hoisting (move complex binop scratch allocas to function entry block)
   - **PR B**: Heap arrays by default (add `--stack-arrays` flag, allocate local arrays on heap by default)

2. Create simple reference test demonstrating behavior (not 16 reference file updates)

3. Remove `--max-stack-var-size` option per feedback - keep it simple with just `--stack-arrays`

Will address after #9310 is merged (LAPACK test enablement blocker).

krystophny:
@certik I believe the issue are large fixed size arrays that are still allocated in stack. What should we do about them?

krystophny:
Anyway I split this PR now in two and then wait for your feedback.

certik (approved):
The change is simple, I think it's fine. If you say it helps, we can keep it.

PR #9313, opened 2026-01-01
fix: make dcmplx default to complex(8)

krystophny:
This fixes `DCMPLX(x)` kind handling so it consistently produces `complex(8)` (and an `imag=0.0d0` when needed), avoiding implicit-interface signature drift when a later argument is `complex(8)` (e.g. `DCONJG(z)`).

- Fix: force `dcmplx` to default to kind=8 in semantics.
- Test: add `integration_tests/dcmplx_01.f90` (requires `--implicit-interface`).

### DIFF
--- a/integration_tests/dcmplx_01.f90
+++ b/integration_tests/dcmplx_01.f90
@@ -5,24 +5,25 @@ program dcmplx_01
     real(dp) :: x, y
     complex(dp) :: z
 
-    x = 1.5d0
-    y = 2.5d0
+    ! Use values that require double precision for accurate representation
+    ! These values would lose precision if converted to single precision
+    x = 1.123456789012345d0
+    y = 2.987654321098765d0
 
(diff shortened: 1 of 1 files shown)


integration_tests/dcmplx_01.f90
```diff
@@ -0,0 +1,17 @@
...
+    real(dp) :: cs
+    complex(dp) :: sn, r
+
+    external :: foo
+
+    x = 1.0_dp
+    z = (1.0_dp, 2.0_dp)
+
+    call foo(dcmplx(x), dconjg(z), cs, sn, r)
+    call foo(-dconjg(z), dconjg(z), cs, sn, r)
```
certik: You need to register this test. And then use some double precision number and test for 1e-15, that will ensure that it is represented as double precision. Just run it as a regular integration test, don't bother with reference tests.

PR #9316, opened 2026-01-01
fix: cast complex init constants to target kind

krystophny:
Fix LLVM verifier failures when initializing complex(kind=8) variables from complex(kind=4) literal constants (common in fixed-form LAPACK code, e.g. COMPLEX*16 PARAMETER ONE=(1.0E+0,0.0E+0)).

Adds an integration test (complex_27) reproducing the verifier error and verifies it now runs.

Tests:
- ./run_tests.py -t complex_27 -s

### DIFF
--- a/integration_tests/complex_29.f90
+++ b/integration_tests/complex_29.f90
@@ -1,5 +1,5 @@
 ! Test: complex*16 parameter with complex(4) literal - kind cast
-program complex_27
+program complex_29
     implicit none
 
     complex*16 zero, one, z


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5232,6 +5232,34 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    im = builder->CreateFPTrunc(im, llvm::Type::getFloatTy(context));
+                } else {
+                    std::string msg = "Conversion from " + std::to_string(init_kind) +
+                                      " to " + std::to_string(target_kind) +
+                                      " not implemented yet.";
+                    throw CodeGenError(msg);
+                }
+                init_value = complex_from_floats(re, im, dst_complex_type);
+            }
+        }
```
certik: Same as the other PR --- I think casting should be done in ASR explicitly, then we just implement it in a straight

certik:
Let's run full tests, to make sure we didn't break anything.

integration_tests/complex_29.f90
```diff
@@ -0,0 +1,14 @@
+! Test: complex*16 parameter with complex(4) literal - kind cast
+program complex_27
```
HarshitaKalani:
```suggestion
program complex_29
```
We can do this with subsequent PRs so as to not disturb the CI.

HarshitaKalani (approved):

PR #9317, opened 2026-01-01
fix: coerce complex parameter constants to declared kind

krystophny:
Problem: fixed-form complex*16 PARAMETER constants written as (0.0,0.0) can be lowered as complex(4) values, which then causes LLVM verifier failures when storing/comparing against complex(8) locations (store type mismatch / mixed-type fcmp). This shows up in Reference-LAPACK complex builds (e.g., ZTPQRT2).

Fix: in LLVM codegen, coerce complex values to the destination complex type when storing; and when fetching Parameter variables, coerce the constant expression to the variable declared type.

Test: add integration test ../integration_tests/complex_28.f (fixed-form) that used to trigger the verifier error.

Follow-up:
- Fix LLVM 15+ (opaque pointers) build by avoiding `getPointerElementType()` and deriving the store target type from the underlying alloca/global/GEP.


tests/reference/run-complex_28-21c1cf0.stdout
```diff
@@ -0,0 +1 @@
+PASS
```
certik: Since it runs, just make it an integration test, no need to add it as a reference.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5225,6 +5225,77 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            target_type = global_var->getValueType();
+        } else if (llvm::GEPOperator* gep_op = llvm::dyn_cast<llvm::GEPOperator>(base_ptr)) {
+            target_type = gep_op->getResultElementType();
+        }
+        if (target_type == nullptr) {
+            return value;
+        }
+        return coerce_complex_value_to_type(value, target_type);
+#endif
+    }
```
certik: I think we should have a proper cast in ASR, and in ASR->LLVM we should not be figuring out any target types --- we should know them from ASR.

PR #9329, opened 2026-01-02
fix: heap allocation for large fixed-size arrays (4KB threshold)

krystophny:
## Summary

Large fixed-size arrays (>= 4KB) are now allocated on the heap instead of the stack to prevent stack overflow. This is simpler and faster than arena allocation.

## Changes

- Arrays > `MAX_STACK_ARRAY_SIZE` (4095 bytes) use `_lfortran_malloc`
- BLOCK constructs always use heap (can be in loops)
- Respects `--stack-arrays` flag to force stack allocation
- Memory automatically freed at function/BLOCK exit

## Performance vs Arena (PR #9367)

| Benchmark | Heap | Arena | Notes |
|-----------|------|-------|-------|
| Small arrays (100 elem) | 0.058s | 0.090s | Heap uses stack |
| Below threshold (511 elem) | 0.060s | 0.080s | Heap uses stack |
| Above threshold (512 elem) | 0.074s | 0.074s | Same |
| Nested calls (3-level) | 0.141s | 0.174s | Heap 20% faster |
| BLOCK in loop | 0.079s | 0.095s | Heap 17% faster |

**Why heap is faster:**
- Modern malloc (glibc/tcmalloc) is highly optimized with thread-local caches
- Arena has per-function save/restore overhead
- Arena's `arena_reset()` iterates chunks to validate - O(n) per function exit
- Our pattern (allocate at entry, free at exit) matches malloc's optimization target

## LAPACK Test Results

| Approach | Pass Rate |
|----------|-----------|
| Heap (this PR) | 84/93 (90%) |
| Arena (PR #9367) | 73/95 (77%) |

## Technical Details

- Threshold: 4095 bytes (`MAX_STACK_ARRAY_SIZE = 4 * 1024 - 1`)
- Consistent with GFortran's `-fmax-stack-var-size` semantics
- Thread-safe: each call gets independent heap memory
- Recursion-safe: no shared global state
- BLOCK arrays freed at BLOCK exit (not function exit) to handle loops

## Verification

Integration test `arrays_101.f90` verifies:
- Small arrays (< 4KB) use stack allocation
- Large arrays (>= 4KB) use heap allocation  
- BLOCK arrays in loops don't leak memory
- Memory properly freed (LLVM IR shows `_lfortran_free` calls)

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4718,8 +4718,8 @@ public:
             builder->CreateCall(fn, {});
         }
         start_new_block(proc_return);
-        free_heap_fixed_size_arrays();
         llvm_symtab_finalizer.finalize_symtab(x.m_symtab);
+        free_heap_fixed_size_arrays();
         llvm::Value *ret_val2 = llvm::ConstantInt::get(context,
             llvm::APInt(32, 0));
         builder->CreateRet(ret_val2);
@@ -6100,8 +6100,8 @@ public:
     inline void define_function_exit(const ASR::Function_t& x) {
         if (x.m_return_var) {
             start_new_block(proc_return);
-            free_heap_fixed_size_arrays();
             llvm_symtab_finalizer.finalize_symtab(x.m_symtab);
+            free_heap_fixed_size_arrays();
             ASR::Variable_t *asr_retval = EXPR2VAR(x.m_return_var);
             uint32_t h = get_hash((ASR::asr_t*)asr_retval);
             llvm::Value *ret_val = llvm_symtab[h];
@@ -6158,8 +6158,8 @@ public:
             builder->CreateRet(ret_val2);
         } else {
             start_new_block(proc_return);
-            free_heap_fixed_size_arrays();
             llvm_symtab_finalizer.finalize_symtab(x.m_symtab);
+            free_heap_fixed_size_arrays();
             builder->CreateRetVoid();
         }
     }
--- a/tests/reference/llvm-arrays_101-8ed52ae.json
+++ b/tests/reference/llvm-arrays_101-8ed52ae.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-arrays_101-8ed52ae.stdout",
-    "stdout_hash": "f8e6556be2b0f69faf1051e3dab4b15f40e6b8d304c892b590ea891b",
+    "stdout_hash": "da9fcde6324fb6775d81fd28ac9389d1e31b411e15fdbe5d6fde3927",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-arrays_101-8ed52ae.stdout
+++ b/tests/reference/llvm-arrays_101-8ed52ae.stdout
@@ -339,10 +339,10 @@ ifcont55:                                         ; preds = %else54, %then53
   br label %return
 
 return:                                           ; preds = %ifcont55
-  call void @_lfortran_free(i8* %2)
   br label %FINALIZE_SYMTABLE_static_large_array
 
 FINALIZE_SYMTABLE_static_large_array:             ; preds = %return
+  call void @_lfortran_free(i8* %2)
   ret i32 0
 }
 


certik:
I'll review carefully after tests pass. The idea is good.

certik:
Here are some questions I would like to see answered:

* We already have `--stack-arrays`. When it is on, are all arrays still allocated on stack in this PR, or are some on the heap?
* When `--stack-arrays` is off, are all arrays always allocated on the heap?
* It seems that arrays were already allocated on the heap --- what kind of cases were not?
* What mechanism is `main` using for this, and why cannot we use it as well?

I've been trying to figure out the answers by reading the PR, but so far I don't quite understand it.

krystophny:
Thanks for reviewing! This PR concerns large local fixed size arrays that were (as far as I understood) always allocated on the stack. All others were already heap allocated unless you use —-stack-arrays on main, which would affect descriptor-based arrays but not locals that are bare arrays in memory.

> We already have --stack-arrays. When it is on, are all arrays still allocated on stack in this PR, or are some on the heap?                                       
                                                                                     
Some are on the heap: large fixed-size locals with physical_type=FixedSizeArray and size > 65536 bytes are allocated via _lfortran_malloc (and freed at exit). Others keep existing behavior.                                                   
                                                                                     
> When --stack-arrays is off, are all arrays always allocated on the heap?         
                                                                                     
No. This PR only moves the large FixedSizeArray locals to heap. Small fixed-size locals still use stack alloca.                                                   
                                                                                     
> It seems that arrays were already allocated on the heap --- what kind of cases were not?                                                                        
                                                                                     
Descriptor-backed arrays (allocatable, assumed-shape, many runtime-shaped cases) already allocate their data on heap by default when --stack-arrays is off. Fixed-
size local arrays were not; they were stack alloca.

> What mechanism is main using for this, and why cannot we use it as well?

main already has a structured exit path that runs finalize_symtab and existing tracked frees. That mechanism does not cover these new raw malloc buffers for fixed-size locals, so this PR adds explicit tracking (heap_fixed_size_arrays) frees them at the same exit points.


Final questions:

1. Should —-stack-arrays force all on stack again including locals?
2. Should we use the pattern of the „put small ones on stack“ also on descriptor based arrays?
3. Should we make the 64kB threshold configurable like in gfortran?
4. Or should we ditch this all for an arena based heap storage?

certik:
Thanks for the answers. I don't know the answers what's the best way. I think all these choices should be made configurable, the 64kB must be made configurable for sure. I think in the future once we do a fast arena-based allocation in our runtime (supported by ASR->LLVM), I think it might be as fast as stack arrays, or very close. But I don't know what users would like to change or not, so best is to not encode these choices in LFortran, but expose them, and choose some defaults. 

I think we can probably merge this.

@jinangshah21, @assem2002 can you please also review this PR?

krystophny:
> Thanks for the answers. I don't know the answers what's the best way. I think all these choices should be made configurable, the 64kB must be made configurable for sure. I think in the future once we do a fast arena-based allocation in our runtime (supported by ASR->LLVM), I think it might be as fast as stack arrays, or very close. But I don't know what users would like to change or not, so best is to not encode these choices in LFortran, but expose them, and choose some defaults.
> 
> I think we can probably merge this.
> 
> @jinangshah21, @assem2002 can you please also review this PR?

I found another issue regarding stack in LAPACK complex number part better wait a bit.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5480,19 +5484,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        type_size = data_layout.getTypeAllocSize(type);
+                        if (type_size > 65536) {
+                            use_heap_allocation = true;
+                        }
+                    }
+
+                    if (use_heap_allocation) {
+                        // Allocate on heap to avoid stack overflow (recursion-safe)
+                        llvm::Value* malloc_size = llvm::ConstantInt::get(
+                            llvm_utils->getIntType(4), llvm::APInt(32, type_size));
```
assem2002:
```suggestion
                            llvm_utils->getIntType(8), llvm::APInt(64, type_size));
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5480,19 +5484,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    // This is recursion-safe unlike static storage, as each call gets its own copy.
+                    // Memory is freed at function exit.
+                    bool use_heap_allocation = false;
+                    uint64_t type_size = 0;
+                    if (ASRUtils::is_array(v->m_type) &&
+                        ASRUtils::extract_physical_type(v->m_type) ==
+                            ASR::array_physical_typeType::FixedSizeArray) {
+                        llvm::DataLayout data_layout(module->getDataLayout());
+                        type_size = data_layout.getTypeAllocSize(type);
+                        if (type_size > 65536) {
```
assem2002: MINOR
```suggestion
                       const int STACKARRAY_MAX_SIZE = 64 * 1024;
                        if (type_size > STACKARRAY_MAX_SIZE) {
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6034,10 +6066,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         declare_local_vars(x);
     }
 
+    inline void free_heap_fixed_size_arrays() {
+        // Free all heap-allocated large fixed-size arrays
+        for (size_t i = 0; i < heap_fixed_size_arrays.n; i++) {
+            llvm_utils->lfortran_free(heap_fixed_size_arrays[i]);
+        }
+        heap_fixed_size_arrays.n = 0;
```
assem2002, line 6097: I think that's fine. We can later move that into the finalization utility class (if fixedSizeArray > 64kb --> free)

assem2002 (approved):
It looks good to me. Thanks!
please address the 32bit -> 64bit comment.
I want to also point out that the issue will persist with global variables --  those are handled by `visit_variable` function (this PR handles only function's variables).
but I think you're more concerned about the recursive calls -- a very large global array isn't a usual use case, and could be reported as a warning to the user at compile time.

jinangshah21 (approved):
I think this is fine other than above comments

krystophny:
Thank you very much for reviewing, since you anyway started I set it now to ready. I tried another arena approach in another PR #9358 after discussion with @certik , but heap seems to be quite performant, so I think we can go with this one for now. I tried to take all your comments into account, and let's now look if CI passes.

integration_tests/CMakeLists.txt
```diff
@@ -2955,3 +2955,4 @@ RUN(NAME intent_out_array_01 LABELS gfortran llvm)
 RUN(NAME intent_out_allocatable_component_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_struct_member_no_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_module_dealloc LABELS gfortran llvm)
+RUN(NAME static_large_array LABELS gfortran llvm llvmStackArray)
```
assem2002:
```suggestion
RUN(NAME arrays_101 LABELS gfortran llvm llvmStackArray)
```

krystophny: Done!

certik (approved):
I think that's fine to start, we can later iterate on it to improve it.

@assem2002 do you mind also please reviewing it?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8457,13 +8503,24 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         SymbolTable* current_scope_copy = current_scope;
         current_scope = block->m_symtab;
 
+        // BLOCK arrays always use heap allocation (can be in loops)
```
assem2002, line 8506: That means any fixedsize array will be allocated on stack when insde block -- that's not optimal but easily fixable.

assem2002:
Looks good. Let's merge

Issue #8593, opened 2025-09-18
Internal Error : default formatting error


certik:
```
program expr2
implicit none

integer, allocatable :: x

print *, x
end
```
Gives:
```
Internal Error : default formatting error
```

It should say "x" not allocated in Debug mode.

assem2002:
Gfortran prints `0` for this case as the variable isn't allocated yet, Should we follow the same case?
The current ICE is just an assertion, it could be removed (The check isn't quite necessary)

certik:
I think this must give a runtime error that `x` is not allocated. When a variable is unallocated, then it can't be used at all, except a few cases like an optional argument (I think), argument to `allocated`, `select type`, etc.

PR #9330, opened 2026-01-02
fix: throw runtime error for unallocated arguments in StringFormat

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8593

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14124,6 +14124,39 @@ public:
         }
     }
 
+    llvm::Value* expr_is_unallocated(ASR::expr_t* arg_expr) {
+        ASR::ttype_t* arg_expr_type = ASRUtils::expr_type(arg_expr);
+        LCOMPILERS_ASSERT(ASRUtils::is_allocatable(arg_expr_type))
+        int64_t ptr_loads_copy = ptr_loads;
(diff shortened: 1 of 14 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14294,34 +14327,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                    } else {
-                        cond = builder->CreateICmpEQ(
-                            builder->CreatePtrToInt(tmp,
-                                llvm::Type::getInt64Ty(context)),
-                            builder->CreatePtrToInt(llvm::ConstantPointerNull::get(arg_expr_llvm_type->getPointerTo()),
-                                llvm::Type::getInt64Ty(context)));
-                    }
-
-                    llvm_utils->generate_runtime_error(cond,
+                    llvm_utils->generate_runtime_error(expr_is_unallocated(arg_expr),
```
certik, line 14330: Question: does this happen all the time, or only if bounds_checking is on?

We need to ensure it only happens when bounds checking is on, but not in Release mode.

certik (approved):
Otherwise I think this is fine. I think it's good enough to merge. But see my comment.

src/libasr/codegen/asr_to_llvm.cpp
xaerru: Only when bounds_checking is on. The function `bounds_check_call` is called after the check for the compiler option.

PR #9341, opened 2026-01-03
fix: ILP64 type consistency in array descriptor operations

krystophny:
## Summary

When `-fdefault-integer-8` is used (ILP64 mode), INTEGER types become i64 in LLVM IR. However, array dimension descriptors are always i32. This caused ICmpInst assertion failures when comparing i32 descriptor values with i64 user expressions in array section operations.

This PR fixes the type mismatch by using `CreateSExtOrTrunc` to ensure consistent types in:
- `llvm_array_utils.cpp`: stride multiplication
- `asr_to_llvm.cpp`: array shape/size comparisons for subroutine calls

## Test Plan

- [x] New integration test `array_section_09.f90` with `-fdefault-integer-8`
- [x] All existing array tests pass
- [x] LAPACK files (slaqz3.f, slaqz4.f) compile with ILP64


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9714,6 +9740,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            left_val->getType()->isIntegerTy() && right_val->getType()->isIntegerTy()) {
+            llvm::Type* target_type = llvm_utils->getIntType(
+                ASRUtils::extract_kind_from_ttype_t(x.m_type));
+            if (left_val->getType() != target_type) {
+                left_val = builder->CreateSExtOrTrunc(left_val, target_type);
+            }
+            if (right_val->getType() != target_type) {
+                right_val = builder->CreateSExtOrTrunc(right_val, target_type);
+            }
+        }
```
certik: Both operands should be exactly the same type. If not, that's a bug in ASR --- we should not be fixing it here, but rather insert proper casts in ASR.

krystophny:
Thanks for the review. I've added type promotion in `asr_builder.h` for ASR-level comparisons, which handles the general case.

The LLVM-level changes (lines 14356, etc.) are specifically for comparing user dimension expressions with internal descriptor values. The array descriptor format uses i32 for sizes, while in ILP64 mode user expressions evaluate to i64. These comparisons happen at LLVM codegen time because one operand (`descriptor_length` from `arr_descr->get_array_size()`) is an LLVM-internal value, not an ASR expression.

Would it be acceptable to keep the LLVM truncation for this specific case? The alternative would be to change the descriptor format to match the user's integer kind, which is a larger change.

certik:
Yes, I think that this is fine --- the descriptor details are a backend thing, so it can cast things there as needed. As long as ASR itself is consistent and all types always match, which I think they do now.

certik (approved):

PR #9342, opened 2026-01-03
fix: handle integer kind mismatch in function calls with implicit argument casting

krystophny:
## Summary
- Handle integer kind mismatch when calling functions with `--implicit-argument-casting`
- Pass an integer(8) where integer(4) is expected (and vice versa) by creating a temporary with the correct kind and converting the value
- Enables LAPACK ILP64 build where external functions have explicit integer(4) parameters

## Changes
- Add `implicit_argument_casting` parameter to `make_FunctionCall_t_util` (asr_utils.h)
- Pass `compiler_options.implicit_argument_casting` at function call sites (ast_common_visitor.h)
- Add integer kind conversion in `convert_call_args` when `orig_arg` is known (asr_to_llvm.cpp)
- Add integer kind conversion in `visit_FunctionCall` using LLVM function parameter types (asr_to_llvm.cpp)
- Add integration test `implicit_argument_casting_03.f90`

## Test plan
- [x] Test passes: `lfortran --implicit-interface --implicit-argument-casting implicit_argument_casting_03.f90`
- [x] Reference tests pass
- [x] CI passes

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -147,7 +147,6 @@ public:
     std::map<std::pair<uint64_t, SymbolTable*>, llvm::Value*> llvm_symtab_deep_copy;
     std::map<uint64_t, llvm::Function*> llvm_symtab_fn;
     std::map<std::string, uint64_t> llvm_symtab_fn_names;
-    std::map<std::string, ASR::Function_t*> llvm_symtab_fn_impl; // Maps fn name to Implementation ASR
     std::map<uint64_t, llvm::Value*> llvm_symtab_fn_arg;
     std::map<uint64_t, llvm::BasicBlock*> llvm_goto_targets;
     std::unordered_map<const ASR::symbol_t*, llvm::BasicBlock*> symbol_to_returnBlock; /// Get Symbol's Return Block -- Used for Finalization. See LLVMFinalize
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13595,6 +13604,68 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            x.m_args[i].m_value, arg_type, module.get());
+                        llvm::Value* loaded = llvm_utils->CreateLoad2(arg_llvm_type, tmp);
+                        // Convert (truncate or extend) to target type
+                        llvm::Value* converted = builder->CreateSExtOrTrunc(loaded, target_llvm_type);
+                        // Store in temporary
+                        builder->CreateStore(converted, temp_var);
+                        tmp = temp_var;
+                    }
+                }
+            }
```
certik: This should not happen in the LLVM backend. Rather, we should insert explicit cast nodes in ASR itself.

krystophny:
Moved integer kind casting from LLVM backend to ASR per review feedback. The cast is now inserted in `Call_t_body` when types don't match and `implicit_argument_casting` is enabled.

certik:
Thanks, I think this is much better now.

certik (approved):
I think this is fine, thanks!

PR #9343, opened 2026-01-03
# PR: Fix EQUIVALENCE+DATA Array Shape Descriptor Construction

dhruv0985:
## Summary

This PR fixes shape mismatch runtime errors when EQUIVALENCE and DATA statements interact with multi-dimensional arrays. The root cause was that the CPtrToPointer shape array descriptor only stored the first dimension extent instead of all array rank dimensions.

## Problem Context

EQUIVALENCE creates memory aliases between variables. When combined with DATA, LFortran generates:

A C pointer to the aliased memory region

A CPtrToPointer ASR node that reshapes this memory as a Fortran descriptor array

A shape array containing the runtime extents of each dimension

The bug was in step (3): the shape array length was implicitly 1, so only the first extent was stored. This caused runtime “Array shape mismatch” errors for multi-dimensional arrays.

## Solution

The shape array now correctly allocates space for all rank dimensions by setting its length to the array rank. As a result, all dimension extents are preserved when constructing the descriptor at runtime.

ASR::expr_t* rank_size = ASRUtils::EXPR(
    ASR::make_IntegerConstant_t(al, loc, (int64_t)arr->n_dims, int_type));
dim.p[0].m_length = rank_size;

## Changes Made

Fix shape array dimension length to match array rank

Make descriptor dimensions deferred as required by ASR verification

Store the shape array as a fixed-size embedded array

Register equivalence_12.f90 in the integration test suite

## Test Results

equivalence_10.f90: PASS (shape mismatch resolved)

equivalence_11.f90: Shape fixed; fails due to unrelated abs() intrinsic issue

equivalence_12.f90: Shape fixed; fails due to unrelated abs() intrinsic issue

## Known Limitation

The abs() intrinsic currently fails during LLVM code generation when applied to elements of equivalenced descriptor arrays. This appears to be a separate backend issue and is not addressed in this PR.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,49 @@
+# Fix EQUIVALENCE+DATA Array Shape Descriptor Construction
+
+## Summary
+Fixes shape mismatch runtime errors when EQUIVALENCE and DATA statements interact with multi-dimensional arrays.
+
+Addresses part of issue #9134.
+
+## Problem
+The CPtrToPointer shape array descriptor only stored the first dimension extent (hardcoded length=1) instead of all array rank dimensions. This caused runtime errors:
+```
+Array shape mismatch in assignment to 'dt19xa':
+LHS should have size 8, but got 1 instead
+```
+
+## Solution
+1. Set shape array dimension length to array rank (not hardcoded to 1)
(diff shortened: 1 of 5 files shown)


certik:
CI fails

PR #9349, opened 2026-01-03
fix: handle CHARACTER array type mismatch in implicit interface calls

krystophny:
## Summary

When calling through an implicit interface with CHARACTER array arguments,
the interface may have a different array physical type (PointerArray) than
the actual implementation (DescriptorArray). This caused LLVM verification
errors.

**Fix approach (AST->ASR level):**

When handling a call through an implicit interface, look up the actual
implementation in parent scopes. If found, use the implementation's array
physical type for the ArrayPhysicalCast comparison, ensuring the correct
cast is inserted.

This is a minimal fix in `Call_t_body` in `asr_utils.h` that:
1. Detects when calling through an Interface (implicit interface)
2. Searches parent scopes for the actual Implementation
3. Uses the implementation's physical type for array cast decisions

## Test Plan

- [x] All reference tests pass (`scripts/lf.sh test`)
- [x] New integration test `implicit_argument_casting_04.f90` passes with LFortran
- [x] New integration test passes with gfortran (validates test correctness)
- [x] Test covers 1D CHARACTER(1), 1D CHARACTER(10), and 2D CHARACTER(5) arrays

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2582,6 +2582,10 @@ RUN(NAME implicit_argument_casting_03 LABELS gfortran llvm
     EXTRA_ARGS --implicit-interface --implicit-argument-casting
     GFORTRAN_ARGS -fallow-argument-mismatch)
 
+RUN(NAME implicit_argument_casting_04 LABELS gfortran llvm
+    EXTRA_ARGS --implicit-interface --implicit-argument-casting
+    GFORTRAN_ARGS -fallow-argument-mismatch)
+
 RUN(NAME specfun_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc
     EXTRA_ARGS --use-loop-variable-after-loop --implicit-typing)
 
--- a/integration_tests/implicit_argument_casting_04.f90
+++ b/integration_tests/implicit_argument_casting_04.f90
@@ -1,0 +1,96 @@
+! Test CHARACTER array type conversion for implicit interface
+! When caller has CHARACTER(n) array with PointerArray physical type
+! and callee has CHARACTER(*) array with DescriptorArray physical type,
+! the codegen must create proper array descriptor.
+!
+! This test covers:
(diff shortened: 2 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13702,6 +13739,127 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         ASRUtils::type_get_past_pointer(ASRUtils::expr_type(x.m_args[i].m_value))));
             }
 
+            // Handle CHARACTER array physical type mismatch for implicit interface
+            // When caller has PointerArray (string_descriptor) but callee expects
+            // DescriptorArray (array descriptor), create descriptor from string_descriptor
```
certik: I think this ASR is malformed then. I think but the callee and the caller have to have the same physical type.

I think this should be fixed in AST->ASR, not here.

krystophny:
@certik Thanks for the review. I've addressed your comment:

**Your feedback:** The fix should be in AST->ASR, not in codegen (`asr_to_llvm.cpp`).

**Updated implementation:** I moved the fix from `asr_to_llvm.cpp` (codegen) to `Call_t_body` in `asr_utils.h` (AST->ASR level).

The fix now works during ASR construction: when a call goes through an implicit interface, it looks up the actual implementation in parent scopes and uses its array physical type for the ArrayPhysicalCast decision. This ensures the correct cast is inserted before reaching codegen.

The old codegen code you commented on has been removed.

krystophny:
Closing in favor of #9384.

After analysis, these two PRs address the same issue (#9381) with different approaches:
- **#9384** (kept): Fixes the issue at the source by using DescriptorArray for CHARACTER arrays when creating implicit interfaces
- **#9349** (this PR): Attempted to fix at call-site by looking up implementation's type

#9384 is the more correct approach as noted by @certik: "the callee and the caller have to have the same physical type" - fixing at interface creation ensures this.

I've merged the comprehensive test from this PR into #9384 as `character_17.f90`.

PR #9337, opened 2026-01-02
Fix create_gep2 type mismatch for UnboundedPointerArray in FileRead

krystophny:
## Summary
- Fix ICE when compiling READ statements with implied do loops on assumed-size arrays
- For `PointerArray` and `UnboundedPointerArray`, load the pointer to get actual data pointer
- Use `create_ptr_gep2` for pointer arrays (single index) instead of `create_gep2` (two indices)

## Test
- Added `read_12.f90` integration test

Fixes #9333

PR #9350, opened 2026-01-03
fix: handle READ implied-do for assumed-size arrays

krystophny:
## Summary
- Fix READ with implied-do loop crashing for assumed-size array parameters
- Add integration test for the fixed behavior

## Details
The implied-do loop code for READ was incorrectly handling PointerArray and UnboundedPointerArray (assumed-size arrays). For these array types, `visit_expr` with `ptr_loads=0` returns the data pointer directly, not a pointer to a descriptor.

### Changes
- Handle PointerArray/UnboundedPointerArray explicitly by using `arr_ptr` directly as the data pointer
- For DescriptorArray, load the data pointer from the descriptor
- Use `create_ptr_gep2` instead of `create_gep2` for pointer arithmetic on flat data pointers

### LAPACK Pattern Fixed
Fixes LAPACK sdrgvx.f READ pattern:
```fortran
REAL STRU(*)
READ(NIN, FMT=*) (STRU(I), I=1,N)
```

## Test Plan
- [x] New integration test `read_13.f90` passes with both gfortran and llvm
- [ ] CI passes

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2326,6 +2326,7 @@ RUN(NAME read_10 LABELS gfortran llvm)
 RUN(NAME read_11 LABELS gfortran llvm)
 RUN(NAME read_12 LABELS gfortran llvm)
 RUN(NAME read_13 LABELS gfortran llvm)
+RUN(NAME read_14 LABELS gfortran llvm)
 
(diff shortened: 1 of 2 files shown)


krystophny:
Closing as duplicate - the codegen fix was already merged in PR #9337. The integration test can be added in a separate PR if needed.

Issue #9323, opened 2026-01-02
Character parameter declaration causes ICE (Wrong physicalType for global string declaration)


w6ws:
```
program char_param
  implicit none

  character*(*) hello
  parameter (hello = 'Hello world!')

  print *, hello

end program
```
```
$ lfortran -c char_param.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2808, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2674, in main_app(int, char**)
    result = compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1249, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 380, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
    compiler_options, run_fn, "", infile, lm);
[...]

PR #9351, opened 2026-01-03
fix: support assumed-length character parameters in declare_global_string

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9323


certik (approved):

certik:
If the exhaustive tests pass, then we can merge it.

PR #9353, opened 2026-01-03
fix: handle whitespace in complex number input parsing

krystophny:
## Summary

Fixes incorrect parsing of complex numbers in Fortran list-directed format when whitespace appears within parentheses. For example:
```
( 0.1000E+01, 0.2000E+01)
```

The previous implementation used `fscanf` with `%s` which stops at whitespace, causing `"("` to be read as the first token. This led to incorrect values being parsed (e.g., `(0.0, 1.0)` instead of `(1.0, 2.0)`).

## Changes

- Adds `read_complex_expr()` helper that reads the entire parenthesized expression character-by-character until the closing parenthesis
- Adds `convert_d_to_e()` helper to handle Fortran D exponent notation before parsing
- Updates `_lfortran_read_complex_float`, `_lfortran_read_complex_double`, and their array variants to use these helpers
- Adds integration test `read_14.f90` covering various complex input formats

## Test plan

- [x] Test passes with gfortran
- [x] Test passes with lfortran (llvm backend)
- [x] All integration tests pass (2082/2082)

### DIFF
--- a/integration_tests/read_22.f90
+++ b/integration_tests/read_22.f90
@@ -1,0 +1,43 @@
+! Test complex number input parsing with whitespace inside parentheses.
+! Fortran list-directed format allows: ( 0.1000E+01, 0.2000E+01)
+program read_21
+    implicit none
+    complex :: c1, c2
+    complex(8) :: c3
+    complex :: arr(2)
+
+    ! Test single-precision complex with spaces inside parentheses
+    open(10, file='read_21_data.txt', status='replace')
+    write(10, '(A)') '( 0.1000E+01, 0.2000E+01)'
+    write(10, '(A)') '(3.0, 4.0)'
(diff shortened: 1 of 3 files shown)


certik:
I think the change is fine, just needs to be rebased.

integration_tests/read_22.f90
```diff
@@ -0,0 +1,43 @@
+! Test complex number input parsing with whitespace inside parentheses.
+! Fortran list-directed format allows: ( 0.1000E+01, 0.2000E+01)
+program read_21
```
certik, line 3:
```suggestion
program read_22
```

HarshitaKalani: taken this with https://github.com/lfortran/lfortran/pull/9375

PR #9354, opened 2026-01-03
fix: copy all UNIT_FILE fields in remove_from_unit_to_file

krystophny:
## Summary
When closing a file, `remove_from_unit_to_file()` shifts remaining entries but was only copying 4 of 8 struct fields (`unit`, `filename`, `filep`, `unit_file_bin`). This left `access_id`, `read_access`, `write_access`, and `delim` uninitialized, causing pre-connected units like stdout (unit 6) to inherit wrong access flags from subsequent array entries.

## Symptom
After closing any file, writes to unit 6 fail with:
```
Runtime Error: Write operation not allowed on unit 6 (opened with action='read').
```

## Test
- `file_40.f90` - Opens/closes a file, then writes to stdout

## Verification
```bash
cd integration_tests && ./run_tests.py -b llvm -t file_40
```


krystophny:
This PR is now redundant - the fix was already merged in upstream using struct assignment (`unit_to_file[i] = unit_to_file[i + 1]`), which copies all fields.

PR #9355, opened 2026-01-03
Added internal string reads

Baitato:
Fixes part of #9193 

Credit/Related PR: #9244


certik:
@Baitato thanks for refactoring it! I like this. Let's see if tests pass.

integration_tests/CMakeLists.txt
```diff
@@ -2326,6 +2326,7 @@ RUN(NAME read_10 LABELS gfortran llvm)
 RUN(NAME read_11 LABELS gfortran llvm)
 RUN(NAME read_12 LABELS gfortran llvm)
 RUN(NAME read_13 LABELS gfortran llvm)
+RUN(NAME read_14 LABELS gfortran llvm)
```
certik, line 2333: Did you forget to commit the read_14 file?

Baitato: Yeah, my bad. Pushed it now.

certik:
The CI failed with:
```
The following tests FAILED:
	1705 - read_07 (Failed)
	1712 - read_14 (Failed)
```
So the new test doesn't work, and also old test `read_07` stopped working.

Let's first fix this bug, so that all integration tests pass. Then we'll run the exhaustive tests.

HarshitaKalani (commented):
Thanks @Baitato! The approach looks good. Let's get the CI to pass.

Baitato:
Looks like it's passing now.
Changes since last review:

1. Updated read_14 to read a string variable instead of a raw string since raw string was not allowed on gfortran. 
gfortran error message: UNIT specification must be an INTEGER expression or a CHARACTER variable
Might need to look into that separately.

2. va_list is apparently handled differently based on OS due to which read_07 and read_14 were failing. Fixed by passing pointer instead of value

How do i run exhaustive checks?

certik:
Perfect! We implemented iostat in the meantime, do you mind please resolving the conflicts? After that we can run the full tests.

Baitato:
> Perfect! We implemented iostat in the meantime, do you mind please resolving the conflicts? After that we can run the full tests.

Resolved conflicts

PR #9358, opened 2026-01-03
feat: arena allocator for fixed-size local arrays

krystophny:
## Summary

Replace stack allocation with thread-safe arena allocation for FixedSizeArray locals to prevent stack overflow in deep call stacks (e.g., LAPACK).

## Design

- Single 64MB arena per thread (thread-local storage)
- Save/restore at function entry/exit recycles memory
- Inline LLVM IR operations (no function call overhead, fully inlineable)
- Parent arrays stay valid because children allocate after saved position

## Implementation

- Add `_lfortran_arena_ptr` thread-local global in runtime
- Add `_lfortran_arena_init()` called at program start
- Generate inline arena ops in LLVM codegen:
  - save: load current pointer
  - alloc: bump pointer with 16-byte alignment
  - restore: store saved pointer
- Pre-scan functions for FixedSizeArray to decide if arena save/restore needed

## Comparison with PR #9329 (heap allocation)

| Aspect | Heap (#9329) | Arena (this PR) |
|--------|--------------|-----------------|
| Alloc cost | malloc() call | 1 load + 1 add + 1 store |
| Free cost | free() call | 1 store |
| Inlineable | No | Yes |
| Memory | Per-array malloc | Single 64MB block/thread |

## Test plan

- [x] Reference tests pass
- [x] Integration test with 12MB arrays (arrays_101.f90)
- [x] Nested function calls with passed arrays
- [x] 11-level recursion with 400KB/level
- [x] Sequential calls reusing arena space
- [ ] LAPACK tests (in progress on stacked branch)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3404,7 +3404,61 @@ LFORTRAN_API void _lfortran_free(char* ptr) {
...
+LFORTRAN_API void* _lfortran_arena_alloc(int64_t size) {
+    ensure_arena_initialized();
+
+    // Align size to 16 bytes
+    size = (size + 15) & ~15;
+
+    void* ptr = _lfortran_arena_ptr;
+    _lfortran_arena_ptr += size;
+    return ptr;
+}
```
certik, line 3452: There are several problems with this implementation:

* The unnecessary `ensure_arena_initialized()` call for every alloc (this alloc will be called many times, so must be as fast as possible)
* No check if you run out of memory
* No means to increase the memory

At the very least we need to check that we can fit the allocation into the preallocated memory, and fail hard if we can't.

krystophny:
Superseded by new implementation with full two-arena conflict avoidance pattern, comprehensive tests, and all critical issues fixed.

PR #9246, opened 2025-12-29
fix: add support for substrings inside write statements

YashNagda17:
Fixes #7787 & Fixes #8369
Add support for writing into StringSections inside write statements by marking them as assignment targets.

PR #9359, opened 2026-01-04
fix: sync scalar variables on cycle statements inside do loops

### LINKED ISSUE #9327: Handle nested function call from do while condition which uses nested vars
MRE:
```fortran
module temp

contains
  subroutine demo()
    implicit none
    integer :: a = 0
    do while(f()) 
       if (a == 0) then 
          a = 1
          print *, a
          cycle
        end if
    end do

  contains
     logical function f()
      f = (a == 0)
      print *, a
    end function f
  end subroutine demo 
end module

program nested_func_mre
  use temp
  implicit none
  call demo()
end program nested_func_mre
```

Outputs:
```console
0
1           
1
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran d.f90
1
0
0
0
0
.. (Infinite loop)
```

YashNagda17:
Fixes #9327
Addressing [Comment in #9246](https://github.com/lfortran/lfortran/pull/9246/files#r2655632705)

Changes made:
- Insert sync for scalar variables on cycle statements inside do loops.


src/runtime/lfortran_compiler.stamp
HarshitaKalani, line 1: Where is this file coming from? Do we need this? 
I also see it getting generated in my local.

YashNagda17: Ohk, let me look into it...

YashNagda17: I tried to build the current main on 2 separate machines, both having Ubuntu 24.04... and I got the same file being created there as well during builds... Not sure why this behavior is happening...

Baitato: I think this is the commit which introduced it: https://github.com/lfortran/lfortran/commit/f3b9c57e28056b7a896403319fcf01a3a654dba4

certik: Let's remove this file. We should add it to .gitignore also I think.

YashNagda17:
Closing this PR, and moving fixes to new PR

Issue #9319, opened 2026-01-01
semantic error: Named initialization not supported


w6ws:
```
program type_init
  implicit none

  type int_t
    integer :: i
  end type

  type(int_t), parameter :: T_ONE = int_t(1)

  type(int_t) :: my_int = T_ONE

  print *, my_int

end program
```
Compiling with lfortran:
```
$ lfortran -c type_init.f90
semantic error: Named initialization not supported with: t_one
  --> type_init.f90:10:3
   |
10 |   type(int_t) :: my_int = T_ONE
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
```
$ gfortran type_init.f90
$ ./a.out
           1
$
```
(@certik - This is needed for ESMF.)

opixdown:
@HarshitaKalani  we can close this isse

w6ws:
@certik - this can be closed.  Thanks!

certik:
Looks like a test was added in https://github.com/lfortran/lfortran/pull/9361

PR #9361, opened 2026-01-04
fix: add support for initializations using scalar parameters for structs inside `type()` syntax

YashNagda17:
Towards #9319 

Changes Made:
- Adding initial support for named initializations from structs having scalar parameters
- Initializations for custom types, arrays etc can be built further on this work.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4826,12 +4826,68 @@ public:
                             }
 
                         } else {
-                            diag.add(Diagnostic(
-                                "Named initialization not supported with: " + sym_name,
-                                Level::Error, Stage::Semantic, {
-                                    Label("",{x.base.base.loc})
-                                }));
-                            throw SemanticAbort();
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4826,12 +4826,68 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            if (ASR::is_a<ASR::StructConstant_t>(*param_init)) {
+                                ASR::StructConstant_t* struct_const = ASR::down_cast<ASR::StructConstant_t>(param_init);
+                                // Create StructConstructor with the constant as its value
+                                init_expr = ASRUtils::EXPR(ASR::make_StructConstructor_t(
+                                    al, x.base.base.loc, struct_const->m_dt_sym,
+                                    struct_const->m_args, struct_const->n_args,
+                                    struct_const->m_type, param_init));
+                            } else {
+                                init_expr = param_init;
+                            }
```
certik, line 4890: We should add tests for all these new error messages. Can be done later.

PR #9304, opened 2025-12-31
CI: run and require passing Reference-LAPACK xlintsts (BUILD_TESTING)

krystophny:
This runs Reference-LAPACK `v3.12.0` with `BUILD_TESTING=ON` in third-party CI and requires the `xlintsts` linear test suite to pass.

Fixes #1232.

Implementation details:
- Builds with `--use-loop-variable-after-loop` (needed by Reference-LAPACK drivers).
- CI fails if `xlintsts` exits non-zero, prints any `failed to pass the threshold` lines, or records any `error messages recorded`.

Prerequisites (merge these first):
- #9280 (self-referencing allocatable array sections)
- #9286 (complex literal assigned to real/integer PARAMETER)
- #9299 (COMMON blocks in contained procedures)
- #9302 (legacy-array-sections double-free)
- #9305 (READ `END=`/`ERR=` EOF control flow)
- #9307 (DATA broadcast compares character constants)

Related:
- #9267 is an older attempt to enable LAPACK `BUILD_TESTING`; this PR supersedes it.

PR #9280, opened 2025-12-30
fix: create temporary for self-referencing allocatable array sections

krystophny:
## Summary

When assigning an array section back to itself (e.g., `arr = arr(2:3)`) with `--realloc-lhs-arrays`, realloc may free/move the source memory before the copy completes, causing use-after-free.

The fix detects self-referencing assignments BEFORE `call_replacer()` transforms the value expression, and forces temporary creation for ALL allocatable arrays (not just struct arrays).

## Test

Added `array_section_08.f90` integration test that covers:
- Simple allocatable arrays: `arr = arr(3:5)`
- Struct arrays: `items = items(2:3)`

## Verification

```
$ valgrind ./test_array_section
PASS
ERROR SUMMARY: 0 errors from 0 contexts
```

Fixes #9275

PR #9267, opened 2025-12-29
ci: enable Reference-LAPACK BUILD_TESTING in third-party tests

krystophny:
## Summary

Enable Reference-LAPACK `BUILD_TESTING=ON` in the third-party test script.

This PR adds Section 15 to `ci/test_third_party_codes.sh` that:
1. Builds Reference-LAPACK v3.12.0 with `BUILD_TESTING=ON`
2. Compiles the `xlintsts` test executable
3. Runs initial LAPACK linear algebra tests

## Prerequisites

This PR requires the following PRs to be merged first:

- [x] #9255 - T/F logical input
- [x] #9261 - Variable-bounds implied-do in READ
- [x] #9258 - Constant-bounds implied-do in READ
- [x] #9254 - Pre-connect units 5, 6, 0
- [x] #9243 - CommonLinkage for common blocks
- [x] #9241 - PointerArray to UnboundedPointerArray cast
- [x] #9240 - bind(c) array ABI
- [x] #9233 - Flatten dimensions for rank mismatch
- [x] #9232 - Strip allocatable wrapper
- [x] #9265 - CHARACTER*1 concatenation
- [x] #9286 - Complex literal to Real PARAMETER
- [x] #9287 - 2D array implied-do in READ

## How to Enable

Once all prerequisite PRs are merged:
1. Rebase this branch on main: `git rebase upstream/main`
2. The LAPACK testing section will then work

## Test Coverage

With all fixes applied, the following LAPACK test suites pass:
- **SGE** - General matrices
- **SGB** - General band matrices
- **SGT** - General tridiagonal matrices
- **SPO** - Positive definite matrices

Total: ~77,000+ individual tests

PR #9363, opened 2026-01-04
CI: run and require passing Reference-LAPACK xlintsts (BUILD_TESTING)

### LINKED ISSUE #1232: Compile Lapack
Here is how to get started:
```console
git clone https://github.com/Reference-LAPACK/lapack
cd lapack
mkdir build
cd build
FC=gfortran cmake -DCBLAS=no -DLAPACKE=no -DBUILD_TESTING=no -DBUILD_DOUBLE=no -DBUILD_COMPLEX=no -DBUILD_COMPLEX16=no ..
make
```

krystophny:
This runs Reference-LAPACK `v3.12.0` with `BUILD_TESTING=ON` in third-party CI and requires the `xlintsts` linear test suite to pass.

Fixes #1232.

Implementation details:
- Builds with `--use-loop-variable-after-loop` (needed by Reference-LAPACK drivers).
- CI fails if `xlintsts` exits non-zero, prints any `failed to pass the threshold` lines, or records any `error messages recorded`.

Supersedes #9304 (closed) and #9267 (older attempt).

All prerequisite PRs (#9280, #9286, #9299, #9302, #9305, #9307, #9310) have been merged.

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -970,133 +970,36 @@ time_section "🧪 Testing Reference-LAPACK v3.12.0 with BUILD_TESTING" '
     # Build BLAS, LAPACK, and test executables
     cmake --build build -j8
 
-    # Helper function to run a LAPACK test and verify results
-    run_lapack_test() {
-        local test_exe=$1
-        local test_input=$2
-        local test_name=$3
-
-        print_subsection "Running LAPACK ${test_name} (${test_exe})"
-        set +e
-        timeout 120 ./bin/${test_exe} < ../TESTING/${test_input} 2>&1 | tee ${test_exe}_${test_input%.in}.out
-        local test_status=${PIPESTATUS[0]}
-        set -e
-
-        echo "${test_exe} ${test_input} exit code: ${test_status}"
-        if [ "${test_status}" -ne 0 ]; then
-            echo "ERROR: ${test_exe} ${test_input} exited with non-zero status"
-            return 1
-        fi
-
-        if grep -qE "failed to pass the threshold" ${test_exe}_${test_input%.in}.out; then
-            echo "ERROR: ${test_exe} ${test_input} threshold failures detected"
-            return 1
-        fi
-
-        local error_messages=$(awk "/error messages recorded/{sum+=\\$1} END{print sum+0}" ${test_exe}_${test_input%.in}.out)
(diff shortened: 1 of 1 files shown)


krystophny:
@certik this enables thirtparty LAPACK tests now that #9310 is merged. Can you set exhaustive ci?

certik:
@krystophny awesome! Let's do it.

certik (approved):
The change is good. Let's hope the CI passes!

We should figure out some easier way to run the tests, the shell script is quite complex, but for now it's fine.

certik:
Ok, here is the error:
```
 End of tests
 Total time used =        24.22 seconds

xlintsts stest.in exit code: 0
+++ awk '/error messages recorded/{sum+=\xlintsts} END{print sum+0}' xlintsts_stest.out
awk: cmd. line:1: /error messages recorded/{sum+=\xlintsts} END{print sum+0}
awk: cmd. line:1:                                ^ backslash not last character on line
++ local error_messages=
xlintsts stest.in error messages recorded: 
++ echo 'xlintsts stest.in error messages recorded: '
✔ Single Real Linear Equations passed
++ '[' '' -ne 0 ']'
ci/test_third_party_codes.sh: line 95: [: : integer expression expected
++ print_success 'Single Real Linear Equations passed'
++ echo -e '\033[0;32m✔ Single Real Linear Equations passed\033[0m'
++ return 0
++ (( PASSED++ ))
```
Something is wrong in the script, although line 95 seems wrong -- but it might count the lines in the `eval` separately.

krystophny:
Fixed the shell script issue. The problem was:

```bash
local error_messages=$(awk "/error messages recorded/{sum+=\\$1} END{print sum+0}" ...)
```

When inside `time_section`'s `eval "$BLOCK"`, the `\\$1` gets expanded incorrectly - first `\\` becomes `\`, then `$1` gets substituted with the first positional parameter (which was `xlintsts`), resulting in `\xlintsts` which awk rejects.

**Fix:** Simplified the LAPACK testing section to:
- Remove all helper functions with complex quoting
- Run single `xlintsts stest.in` as the smoke test  
- Use simple `grep -E "[1-9][0-9]* error messages recorded"` instead of awk

This is much more robust and avoids the eval quoting hell.

certik:
I think it works. I am going to merge it.

certik:
I am testing that the CI tests actually work in https://github.com/lfortran/lfortran/pull/9369.

Issue #1999, opened 2023-07-21
Support namelist


certik:
https://github.com/certik/fastGPT/commit/11ead54d4943be4a1e189ce97b5f31c6a182dcb5

Shaikh-Ubaid:
How do we support this? I guess we need a new data structure/node at the ASR Level.

certik:
Duplicate: https://github.com/lfortran/lfortran/issues/1851

certik:
Let's figure out how to represent namelists at the ASR level. It would be good to design them in such a way as to allow (later) to represent any user-defined custom readers/writers for other formats like JSON, TOML, NPY, HDF5, GGUF, etc.

It looks like the `read(u, input_fastGPT)` line will read a structure format from the textual input file. One way is to transform it to something like `_lfortran_namelist_read(u, "A", A, "B", B, etc.)`, and it would populate those variables if present in the input file. Similar to how we handle printing.

See also: https://github.com/j3-fortran/fortran_proposals/issues/331.

certik:
One way that might work: `read(u, input_fastGPT)` where `input_fastGPT` is a namelist that contains arrays `A`, `B` and an integer `i` gets transformed into
```c
_lfortran_namelist_context ctx;
_lfortran_namelist_read(u, &ctx);
_lfortran_namelist_read_array_f32(ctx, A);
_lfortran_namelist_read_array_f64(ctx, B);
_lfortran_namelist_read_i64(ctx, i);
```

Let's check how other compilers do this.

certik:
In the first implementation, let's do the following:

* Introduce `FileReadNamelist` node that accepts the file (handle) and list of variables to fill in (`expr *vars`).
* In the LLVM backend translate this node into `_lfortran_namelist_read`. Possibly it might translate it to a sequence of calls per https://github.com/lfortran/lfortran/issues/1999#issuecomment-2007586757.
* Later, if the namelist itself can be a global module variable, then we need to put it into the symbol table. But when the line `read(u, some_namelist)` is encountered by the frontend, it would translate it to `FileReadNamelist` with explicit vars, so it would reuse the previous two points.

ianfr:
Is this feature being actively worked on?

certik:
Not right now. If you want to send a PR, you can.

PR #9364, opened 2026-01-04
feat: introduce `Namelist` ASR node

kmr-srbh:
Towards #1999

Implement namelist support in ASR and update `FileRead`, `FileWrite` ASR nodes with `nml` field.


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3653,18 +3654,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
         current_struct_type_var_expr = nullptr;
 
-        if (x.m_vartype == nullptr &&
-                x.n_attributes == 1 &&
-                AST::is_a<AST::AttrNamelist_t>(*x.m_attributes[0])) {
-            //char *name = down_cast<AttrNamelist_t>(x.m_attributes[0])->m_name;
-            diag.add(Diagnostic(
-                "Namelists not implemented yet",
-                Level::Error, Stage::Semantic, {
-                    Label("",{x.base.base.loc})
```
assem2002, line 3663: Now ASR has proper namlist representation, that's great! Are we sure backend raises an error while processing namelists? if not, let's keep that error for now.

src/libasr/ASR.asdl
```diff
@@ -27,6 +27,7 @@ symbol
     | Block(symbol_table symtab, identifier name, stmt* body)
     | Requirement(symbol_table symtab, identifier name, identifier* args, require_instantiation* requires)
     | Template(symbol_table symtab, identifier name, identifier* args, require_instantiation* requires)
+    | Namelist(symbol_table parent_symtab, identifier group_name, symbol* var_list)
```
assem2002, line 30: Do we need `parent_symtab` with this node?

assem2002 (approved):
Looks good! Good start.
I think it's good to merge, please just address my minor comments above.

src/libasr/ASR.asdl
certik: I think every symbol has to have either a symtab or a parent symtab.

src/lfortran/semantics/ast_common_visitor.h
certik: It should give the usual error that NameList node is not handled in ASR->LLVM.

tests/namelist_01.f90
```diff
@@ -0,0 +1,12 @@
...
+
+   integer :: count = 10
+   integer :: max_iter = 100
+   real :: tolerance = 1.0e-6
+   real :: factor = 1.5
+   logical :: verbose = .false.
+   character(len=30) :: method = 'default'
+
+   namelist /config/ count, max_iter, tolerance, factor, verbose, method
+end program namelist_01
```
certik, line 12: This can go to integration tests and test it with GFortran only, to ensure the code is correct.

certik (approved):
I think that this is good to merge. We can then improve upon it with subsequent PRs.

PR #9365, opened 2026-01-04
fix: sync scalar variables on cycle statements inside do loops

### LINKED ISSUE #9327: Handle nested function call from do while condition which uses nested vars

YashNagda17:
Fixes #9327
Addressing [Comment in #9246](https://github.com/lfortran/lfortran/pull/9246/files#r2655632705)
Addressing [Comment in #9359](https://github.com/lfortran/lfortran/pull/9359#discussion_r2659678520)

Changes made:
- Insert sync for scalar variables on cycle statements inside do loops.
- update .gitignore for .stamp files generated during builds

### DIFF
--- a/tests/reference/llvm-program_03-374e848.json
+++ b/tests/reference/llvm-program_03-374e848.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-program_03-374e848.stdout",
-    "stdout_hash": "9c7d7e22a1411f2b90da8d139fc05ce3a50ff551a2e7f5ad9db989b9",
+    "stdout_hash": "8ea8d1e2dc9f3da27eeea53ed95f580dd37b5d54fdfc525f25879d4c",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-program_03-374e848.stdout
+++ b/tests/reference/llvm-program_03-374e848.stdout
@@ -55,8 +55,10 @@ loop.body:                                        ; preds = %loop.head
   br label %loop.head
 
 loop.end:                                         ; preds = %loop.head
-  %22 = load i32, i32* @__module___lcompilers_created__nested_context__closuretest____lcompilers_created__nested_context__closuretest_z, align 4
-  store i32 %22, i32* %z1, align 4
+  %22 = load i32, i32* %z1, align 4
+  store i32 %22, i32* @__module___lcompilers_created__nested_context__closuretest____lcompilers_created__nested_context__closuretest_z, align 4
+  %23 = load i32, i32* @__module___lcompilers_created__nested_context__closuretest____lcompilers_created__nested_context__closuretest_z, align 4
+  store i32 %23, i32* %z1, align 4
   call void @_lpython_free_argv()
   br label %return
 


certik:
I am confused what the "sync" does --- is it inserting some scalar/array copies before cycle? Why is it necessary?

YashNagda17:
This fix intends for cases where there's a function in the loop test, and there are common variables in main loop body & function.

In current main, after executing all of the loop body functions, we perform a  memcopy (sync) of such variables, between main loop and function for test evaluation for next loop execution. 

Statements like `cycle()` don't reach loop end, and directly jump back to test condition at the start. So here, for scalar variables, we need to perform a memcpy when cycle is executed, just as loop end. Since arrays are referenced by pointers, they work fine even without this fix.

certik:
It might be the correct fix for now. 

But I wonder if we truly need all such copies. Can you give me an example on this code:
```fortran
  subroutine demo()
    implicit none
    integer :: a = 0
    do while(f()) 
       if (a == 0) then 
          a = 1
          print *, a
          cycle
        end if
    end do

  contains
     logical function f()
      f = (a == 0)
      print *, a
    end function f
  end subroutine demo 
```

How does the transformed code look like? My understanding is that this has to do with correctly decomposing `do while(f()) ` into using a temporary and evaluating `f()` explicitly --- once before the loop and then always at the end of the loop. And has to work with `cycle` correctly.

Here is one way to do it:
```fortran
    do 
       tmp = f()
       if (.not. tmp) exit 
       if (a == 0) then 
          a = 1
          print *, a
          cycle
        end if
    end do
```
I think that should work with `cycle` correctly, no need for other temporaries and copies. No? Is it that in our current design we do something more complicated, so it requires a more complex fix?

YashNagda17:
The problem here is state of variables of main referenced inside function... For scalars, we are using temporary variables to referenced them in both spaces

Pseudocode of current state here:
```
a = 0
__lcompilers_created__nested_context__demo_a = a
while (f()) do 
    if (a==0) then
        a = 1
        ## This line generated after this fix
        __lcompilers_created__nested_context__demo_a = a
        cycle
    end if

    ## memory copies at loop end by main
    __lcompilers_created__nested_context__demo_a = a
    a = __lcompilers_created__nested_context__demo_a

```
Due to cycle, we are unable to execute assignments at function end... hence this fix...
I think these assignments are introduced [here (1464)](https://github.com/lfortran/lfortran/pull/1464)

Reference:  ASR after applying this pass, here `__lcompilers_created__nested_context__demo_a` acts as a link between main and function for variable a:
```
[(Assignment
          (Var 3 a)
          (IntegerConstant 0 (Integer 4) Decimal)
          ()
          .false.
          .false.
      )
      (Assignment
          (Var 3 __lcompilers_created__nested_context__demo_a)
          (Var 3 a)
          ()
          .false.
          .false.
      )
      (WhileLoop
          ()
          (FunctionCall
              3 f
              ()
              []
              (Logical 4)
              ()
              ()
          )
          [(If
              ()
              (IntegerCompare
                  (Var 3 a)
                  Eq
                  (IntegerConstant 0 (Integer 4) Decimal)
                  (Logical 4)
                  ()
              )
              [(Assignment
                  (Var 3 a)
                  ()
                  .false.
                  .false.
              )
              (Print
                  (StringFormat
                      ()
                      [(Var 3 a)]
                      FormatFortran
                      (Allocatable
                      )
                      ()
                  )
              )
              (Assignment
                  (Var 3 __lcompilers_created__nested_context__demo_a)
                  (Var 3 a)
                  ()
                  .false.
                  .false.
              )
              (Cycle
                  ()
              )]
              []
          )]
          []
      )
      (Assignment
          (Var 3 __lcompilers_created__nested_context__demo_a)
          (Var 3 a)
          ()
          .false.
          .false.
      )
      (Assignment
          (Var 3 a)
          (Var 3 __lcompilers_created__nested_context__demo_a)
          ()
          .false.
          .false.
      )]
```

certik:
I see, I get it now --- it's about the nested function accessing parent scope, where we indeed have to make copies --- we use a pointer usually for arrays, to avoid a copy. Yes, that is needed, but only when nested functions are used.

certik:
Last two questions:

* Do we only do this before exit / cycle, and only if nested functions are used? So that if nested functions are not used, or cycle/exit is not used, we do not insert anything, so things are fast.
* For arrays, is it using a pointer array, so the assignment is just a pointer change? Or is it actually copying the full array?

YashNagda17:
Yes, we are taking care of both the conditions here

1. We only generate this optimization when either loop body or loop test condition has function call ([Reference to set `calls_present` ](https://github.com/lfortran/lfortran/pull/9365/files#diff-5e0c30dde8ff153c28b9445c8626c39b4df60247ff191676a6fe140e3617a0d7R989-R1003) ) & [Iterating only when `calls_present == true`](https://github.com/lfortran/lfortran/pull/9365/files#diff-5e0c30dde8ff153c28b9445c8626c39b4df60247ff191676a6fe140e3617a0d7R713-R722)

2. Inside [nested_vars](https://github.com/lfortran/lfortran/pull/9365/files#diff-5e0c30dde8ff153c28b9445c8626c39b4df60247ff191676a6fe140e3617a0d7R849-R852) , we only push the variables for assignments which are not arrays... Since arrays are maintained by pointers, current state using associate statement works fine (confirmed by test_file compiles for arrays as well without any issues)

ASR for confirming point 2 for following module (no assignments here for arrays):

```
subroutine demo()
    implicit none
    integer :: arr(2)
    arr(2) = 0
    do while(f()) 
       if (arr(2) == 0) then 
          arr(2) = 1
          cycle
        end if
    end do

  contains
     logical function f()
      f = (arr(2) == 0)
      print *, arr(2)
    end function f
  end subroutine demo
```

```
(Associate
(Var 3 __lcompilers_created__nested_context__demo_arr)
(ArrayPhysicalCast
    (Var 3 arr)
    FixedSizeArray
    DescriptorArray
    (Array
        (Integer 4)
        DescriptorArray
    )
    ()
)
)
(WhileLoop
()
(FunctionCall
    3 f
    ()
    []
    (Logical 4)
    ()
    ()
)
[(If
    ()
    (IntegerCompare
        (ArrayItem
            (Var 3 arr)
            [(()
            (IntegerConstant 2 (Integer 4) Decimal)
            ())]
            (Integer 4)
            ColMajor
            ()
        )
        Eq
        (IntegerConstant 0 (Integer 4) Decimal)
        (Logical 4)
        ()
    )
    [(Assignment
        (ArrayItem
            (Var 3 arr)
            [(()
            (IntegerConstant 2 (Integer 4) Decimal)
            ())]
            (Integer 4)
            ColMajor
            ()
        )
        ()
        .false.
        .false.
    )
    (Cycle
        ()
    )]
    []
)]
```

certik:
Regarding

> We only generate this optimization when either loop body or loop test condition has function call ([Reference to set calls_present ](https://github.com/lfortran/lfortran/pull/9365/files#diff-5e0c30dde8ff153c28b9445c8626c39b4df60247ff191676a6fe140e3617a0d7R989-R1003)) & [Iterating only when calls_present == true](https://github.com/lfortran/lfortran/pull/9365/files#diff-5e0c30dde8ff153c28b9445c8626c39b4df60247ff191676a6fe140e3617a0d7R713-R722)

I am trying to figure out from this what the answer is to my question above:

> Do we only do this before exit / cycle, and only if nested functions are used?

My understanding based on what you wrote is:

* "Do we only do this before exit / cycle": no, this happens every time there is any function call
* "and only if nested functions are used?", no, this happens for all functions

If so, then this is not the right design. However, I think your analysis of the code in the PR might not be correct, the code might actually do what I was asking, but I am not 100% sure, especially after your answer. :)

I'll try to play with the actual PR locally to figure out the precise answers to my questions above.

certik:
Ok, I played with this branch locally and here is what I found:

* If I remove the function `f()` from the conditions, then the sync copies disappear. As they should.
* If I keep `f()`, but make it not access parent function variables, then the sync copies disappear, as they should.
* I also found that for arrays we just associate before the loop as a pointer. As it should.

So I think this design is good and it is behaving correctly.

PR #9366, opened 2026-01-04
Throwing semantic error for undefined err and end labels

### LINKED ISSUE #9362: Undefined END=/ERR= label should produce semantic error, not codegen crash
## Summary

When using an undefined label in READ `END=` or `ERR=` specifiers, LFortran crashes with an internal LLVM error instead of providing a clear semantic diagnostic.

## Minimal Reproducer

```fortran
program test_undefined
    integer :: x, u
    open(newunit=u, status='scratch')
    write(u, '(I4)') 42
    rewind(u)
    read(u, *, end=999) x   ! label 999 does not exist
    print *, x
end program
```

## Current Behavior (LFortran)

```
Referring to a basic block in another function!
  br label %goto_target

```

## Expected Behavior (gfortran)

```
Error: END tag label 999 at (1) not defined
```

## Suggested Fix

Add validation in semantic analysis (`src/lfortran/semantics/`) to check that labels used in `END=` and `ERR=` specifiers are defined within the current scope. This should emit a clear diagnostic like:

```
semantic error: END= label 999 is not defined
```

## Related

- PR #9310 fixed runtime handling of END=/ERR= control flow
- This issue addresses compile-time validation of undefined labels

Baitato:
Fixes #9362


tests/tests.toml
```diff
@@ -4528,6 +4528,14 @@ asr = true
 filename = "errors/iostat_constant_integer.f90"
 asr = true
 
+[[test]]
+filename = "errors/undefined_end_label.f90"
+asr = true
+
+[[test]]
+filename = "errors/undefined_err_label.f90"
+asr = true
```
certik: Let's move these into continue_compilation_1.f90, no need for separate files

certik:
I think the PR is fine otherwise, but let's fix the testing.

tests/tests.toml
Baitato: Done

certik (approved):

PR #9367, opened 2026-01-04
feat: arena allocator for large fixed-size arrays (alternative to #9329)

krystophny:
## Summary

Alternative approach to PR #9329 using arena allocation instead of heap.

**Recommendation: PR #9329 (heap) is simpler and faster. This PR is kept for reference.**

## Performance Comparison

| Benchmark | Arena (this PR) | Heap (#9329) | Difference |
|-----------|-----------------|--------------|------------|
| Small arrays | 0.090s | 0.058s | Arena 55% slower |
| Below threshold | 0.080s | 0.060s | Arena 33% slower |
| Above threshold | 0.074s | 0.074s | Same |
| Nested calls | 0.174s | 0.141s | Arena 23% slower |
| BLOCK in loop | 0.095s | 0.079s | Arena 20% slower |

**LAPACK:** Arena 73/95 (77%) vs Heap 84/93 (90%)

## Why Heap Wins

Arena allocators excel at **batch deallocation** (allocate many, free all at once). Our pattern is different:
- Allocate at function entry
- Free at function exit (per-array)

For this pattern:
1. Modern malloc has thread-local caches optimized for exactly this
2. Arena adds per-function save/restore overhead
3. Arena's `arena_reset()` validates by iterating chunks - O(n) per exit
4. Heap is simpler: just malloc/free

## Design (for reference)

- Two thread-local scratch arenas (Ryan Fleury/wasi pattern)
- Chunk-based growth starting at 4KB
- Position save/restore at function boundaries
- `MAX_STACK_ARRAY_SIZE = 4095` bytes threshold

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -305,7 +305,7 @@ public:
         if (compiler_options.stack_arrays) {
             return false;  // Arena disabled when stack_arrays is true
         }
-        constexpr uint64_t STACK_ARENA_THRESHOLD = 4096;
+        constexpr uint64_t MAX_STACK_ARRAY_SIZE = 4 * 1024 - 1;  // 4KB arrays and above use arena
         for (auto& item : x.m_symtab->get_scope()) {
             if (ASR::is_a<ASR::Variable_t>(*item.second)) {
                 ASR::Variable_t* v = ASR::down_cast<ASR::Variable_t>(item.second);
@@ -315,13 +315,13 @@ public:
                     ASRUtils::is_array(v->m_type) &&
                     ASRUtils::extract_physical_type(v->m_type) ==
                         ASR::array_physical_typeType::FixedSizeArray) {
-                    // Check if array size >= threshold
+                    // Check if array size > threshold (4KB+ uses arena)
                     ASR::dimension_t* dims = nullptr;
                     size_t n_dims = ASRUtils::extract_dimensions_from_ttype(v->m_type, dims);
                     int64_t fixed_size = ASRUtils::get_fixed_size_of_array(dims, n_dims);
                     uint64_t elem_size = estimate_element_size(v->m_type);
                     uint64_t total_size = fixed_size * elem_size;
-                    if (total_size >= STACK_ARENA_THRESHOLD) {
+                    if (total_size > MAX_STACK_ARRAY_SIZE) {
                         return true;
                     }
                 }
@@ -5714,11 +5714,10 @@ public:
                             v->base.base.loc);
                     }
                     uint64_t total_size = fixed_size * el_size;
-                    // Threshold: use stack for arrays < 4KB, arena for >= 4KB
+                    // Threshold: use stack for arrays <= 4KB-1, arena for >= 4KB
                     // Exception: BLOCKs always use arena (can be in loops)
-                    // This matches ARENA_MIN_CHUNK_SIZE from lfortran_arena.h
-                    constexpr uint64_t STACK_ARENA_THRESHOLD = 4096;
-                    if (!in_block_context && total_size < STACK_ARENA_THRESHOLD) {
+                    constexpr uint64_t MAX_STACK_ARRAY_SIZE = 4 * 1024 - 1;  // 4KB arrays and above use arena
+                    if (!in_block_context && total_size <= MAX_STACK_ARRAY_SIZE) {
                         // Small array in function - use stack allocation
 #if LLVM_VERSION_MAJOR >= 15
                         bool is_llvm_ptr = false;


krystophny:
## Arena Allocator: Current State and Future Direction

### What This PR Does (Current Scope)

Implements a two-arena scratch allocator for **local FixedSizeArray variables**:
- Two thread-local arenas alternate by call depth to prevent caller/callee conflicts
- `scratch_begin()` saves current position, `scratch_end()` restores it
- Each function's locals are allocated sequentially on its arena
- Position save/restore ensures proper cleanup on function exit

### What Works ✓

**Pass-by-reference intent(out) FixedSizeArrays** through any call topology:

```fortran
subroutine f()
    integer :: A(100)     ! arena 0, bytes 0-400
    call g(A)
end subroutine

subroutine g(B)
    integer, intent(out) :: B(100)  ! NOT allocated - just pointer to A
    integer :: C(100)     ! arena 1, bytes 0-400
    call h(B)
end subroutine

subroutine h(D)
    integer, intent(out) :: D(100)  ! pointer to A
    integer :: E(100)     ! arena 0, bytes 400-800 (AFTER A, no conflict)
end subroutine
```

Key insight: **intent(out) FixedSizeArrays are not allocated by callee** - they're pointers to caller's memory. The two-arena alternation + position save/restore correctly handles:
- Linear call chains
- Diamond patterns (multiple callees modifying same array)
- Recursive calls (each level adds to arena, unwind restores)
- Sibling calls (reuse arena space after restore)

### Potential Issues ⚠️

**1. Function Results**
```fortran
function make_array() result(R)
    integer :: R(100)  ! intent_return_var - allocated on arena?
    R = 42
end function  ! scratch_end() here - R freed before return?
```
The code includes `intent_return_var` in `scope_uses_arena()`. If R is arena-allocated, `scratch_end()` frees it before the value is used. Need to verify if LLVM sret convention handles this (caller provides storage).

**2. BLOCK Locals in Loops**
```fortran
do i = 1, 1000000
    block
        integer :: B(1000)  ! Allocated each iteration, never freed
    end block
end do
```
Current implementation only emits scratch_begin/end for Functions and Programs, not BLOCKs. This causes unbounded arena growth.

### Future Design Discussion

For a more comprehensive arena system that handles **all** array types:

1. **Arena pointer in descriptors**: intent(out) allocatable arrays could carry parent's arena reference
2. **Compile-time arena count**: Analyze call graph to determine how many arenas needed for rotation (may need >2)
3. **All arrays on arenas**: Local fixed-size, dynamic-size (`real :: A(n)`), and allocatable
4. **Deallocation semantics**: 
   - Non-allocatable → no-op
   - Allocatable → set flag in descriptor, no actual free
   - Memory reclaimed when scratch arena resets
5. **Loop corner case**: Allocate/deallocate in loop needs loop-scoped arena
6. **Lifetime analysis**: Compiler may need to determine array lifetimes within functions

### Immediate Next Steps

1. Verify function result handling (check sret codegen)
2. Decide if BLOCK-scoped arenas are needed for this PR or can be deferred

krystophny:
## Performance Benchmark Results (with 4KB Threshold)

Tested on Linux x86_64. All compilers use `-O0` for fair allocation overhead comparison.

### 4KB Threshold Implementation

- **< 4096 bytes**: Stack allocation (fast, no overhead)
- **>= 4096 bytes**: Arena allocation (prevents stack overflow)
- **BLOCK constructs**: Always arena (can be in loops)

### Full Comparison Table (seconds, lower is better)

| Benchmark | Size | lf-arena | lf-main | gfortran | nvfortran | ifx |
|-----------|------|----------|---------|----------|-----------|-----|
| Small arrays (100 elem, 1M iter) | 400 B | 0.0595 | 0.0563 | 0.0876 | 0.0788 | 0.1094 |
| Below threshold (511 elem, 100K iter) | 4088 B | 0.0625 | 0.0570 | 0.0400 | 0.0397 | 0.0363 |
| Above threshold (512 elem, 100K iter) | 4096 B | 0.0613 | 0.0638 | 0.0368 | 0.0421 | 0.0344 |
| Medium arrays (1000 elem, 100K iter) | 8 KB | 0.1231 | 0.1228 | 0.0700 | 0.0776 | 0.0665 |
| Large arrays (10000 elem, 10K iter) | 80 KB | 0.1105 | 0.1102 | 0.0786 | 0.0796 | 0.0571 |
| Nested calls (3-level, 100K iter) | mixed | 0.1410 | 0.1357 | 0.1038 | 0.0911 | 0.0917 |
| BLOCK in loop (500 elem, 100K iter) | 4 KB | 0.0811 | **CRASH** | 0.0448 | 0.0456 | 0.0315 |

### Arena vs Main Branch Analysis

| Benchmark | Arena | Main | Difference |
|-----------|-------|------|------------|
| Small (stack) | 0.0595 | 0.0563 | +5.7% |
| Below threshold (stack) | 0.0625 | 0.0570 | +9.6% |
| Above threshold (arena) | 0.0613 | 0.0638 | -3.9% (arena faster!) |
| Medium (arena) | 0.1231 | 0.1228 | +0.2% |
| Large (arena) | 0.1105 | 0.1102 | +0.3% |
| Nested (stack) | 0.1410 | 0.1357 | +3.9% |
| BLOCK (arena) | 0.0811 | CRASH | Arena prevents crash |

### Key Findings

1. **Threshold works correctly**: Verified with LLVM IR - below 4KB uses `alloca`, at/above uses `scratch_alloc`
2. **Small arrays**: ~5-10% overhead vs main (stack allocation overhead from threshold check)
3. **Large arrays**: Negligible overhead (~0.2-0.3%) - arena is actually faster at threshold boundary!
4. **BLOCK crash fixed**: Main branch segfaults, arena branch works
5. **vs traditional compilers**: ~1.5-2x slower due to LLVM optimization differences, not arena overhead

krystophny:
Closing in favor of #9329 but keeping branch for reference.

PR #9369, opened 2026-01-05
CI: test that Lapack tests fail

certik:
This PR reverts 3ed9cffbc0429b369b5745cce77914e8a6f3b53c from https://github.com/lfortran/lfortran/pull/9286, which was one of the prerequisites for https://github.com/lfortran/lfortran/pull/9363, so now the tests should fail.


certik:
@krystophny do you mind looking at the logs if the failures look good? They are of the type:
```
xlintsts exit code: 0
++ echo 'ERROR: threshold failures detected'
ERROR: threshold failures detected
++ grep 'failed to pass the threshold' xlintsts_stest.out
 SQK:  36885 out of 241365 tests failed to pass the threshold
++ exit 1
```
Which looks good to me. However it's concerning that LLVM 16, 18, 19 and 20 passes. But 17 fails, as well as 8-15. Some of the lower LLVM versions fail with LLVM verify error.

Overall I think it's fine, the CI tests work.

PR #9372, opened 2026-01-05
Fix: Preserve embedded null bytes in string formatting

### LINKED ISSUE #9371: CI: LAPACK failures
https://github.com/lfortran/lfortran/actions/runs/20702112236/job/59425986197

```
++ exit_code=0
++ set -e
++ echo 'xlintsts exit code: 0'
++ '[' 0 -ne 0 ']'
++ grep -qE 'failed to pass the threshold' xlintsts_stest.out
xlintsts exit code: 0
++ echo 'ERROR: threshold failures detected'
++ grep 'failed to pass the threshold' xlintsts_stest.out
ERROR: threshold failures detected
 SQK:  36885 out of 241365 tests failed to pass the threshold
++ exit 1
```

krystophny:
## Summary

Fix runtime string formatting to preserve embedded null bytes in CHARACTER variables.

The root cause of LAPACK SQK threshold test failures was that the runtime string formatting function `_lcompilers_string_format_fortran` used `strlen(result)` to track result length, which stops at the first null byte. This caused incorrect output when formatting CHARACTER variables containing embedded nulls.

## Changes

- Track `result_len` explicitly throughout `_lcompilers_string_format_fortran`
- Use `append_to_string_NTI` with explicit lengths instead of `append_to_string`
- Use temp buffer pattern for handler functions (handle_integer, handle_float, etc.)

## Context

Fortran CHARACTER variables that are declared but not explicitly initialized contain undefined values (often null bytes). When these variables are formatted and concatenated with other values, the null bytes must be preserved to maintain correct positioning of subsequent data.

## Verification

- All reference tests pass
- All integration tests pass (2098/2098)
- New integration test `format_45.f90` verifies null byte preservation in formatted output
- LFortran output now matches gfortran for CHARACTER variables with embedded null bytes

Fixes #9371

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -871,6 +871,7 @@ RUN(NAME format_41 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_42 LABELS gfortran llvm)
 RUN(NAME format_43 LABELS gfortran llvm)
 RUN(NAME format_44 LABELS gfortran llvm)
+RUN(NAME format_45 LABELS gfortran llvm)
 
 RUN(NAME submodule_01 LABELS gfortran)
 RUN(NAME submodule_02 LABELS gfortran fortran)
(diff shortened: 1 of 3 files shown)


krystophny:
@certik this is supposed to resolve flakey behavior on main in LAPACK tests.

certik:
The CI in this PR failed for LLVM 16, 19, 20 with:

image

All the same error:

PR #9384, opened 2026-01-05
fix: use DescriptorArray for CHARACTER arrays in implicit interfaces

### LINKED ISSUE #9381: Assumed-size CHARACTER(*) array descriptor has garbage bounds via implicit interface
When passing a fixed-size CHARACTER array to a subroutine with an assumed-size CHARACTER(*) dummy argument via implicit interface, the array descriptor bounds are not properly initialized. This causes garbage bounds and either segfaults or array bounds errors.

## Minimal Reproducing Example

```fortran
      PROGRAM TEST
      CHARACTER ADUMMA(1)
      ADUMMA(1) = ' '
      CALL CALLEE(ADUMMA)
      END

      SUBROUTINE CALLEE(EI)
      CHARACTER EI(*)
      IF (EI(1) .EQ. ' ') THEN
         PRINT *, 'PASS'
      ELSE
         PRINT *, 'FAIL'
      END IF
      END
```

## Commands

```bash
lfortran test.f --fixed-form-infer --implicit-interface --implicit-typing
```

## Expected Output

```
PASS
```

## Actual Output

With bounds checking: 
```
Array 'ei' index out of bounds. Valid range is -1981445816 to -9472572
```

Without bounds checking: Segfault

## Context

This blocks several LAPACK eigenvalue tests (nep.in, sed.in, sgd.in, dgd.in, ded.in for both single and double precision).

The pattern is common in LAPACK where a dummy CHARACTER array like `ADUMMA(1)` is passed to routines like DLATME/SLATME that declare `CHARACTER EI(*)`.

## Root Cause

When creating the array descriptor for the assumed-size CHARACTER(*) dummy, the bounds are not properly initialized from the actual argument. The garbage bounds indicate uninitialized memory.

krystophny:
## Summary

Fix CHARACTER(*) assumed-size array passing through EXTERNAL declarations by using DescriptorArray physical type for CHARACTER arrays in implicit interfaces.

## Problem

When passing CHARACTER(*) assumed-size arrays through EXTERNAL declarations, the array descriptor bounds were not properly initialized. This caused garbage bounds and either segfaults or array bounds errors.

Example pattern from LAPACK:
```fortran
SUBROUTINE SCHKHS(ADUMMA)
    CHARACTER ADUMMA(1)
    EXTERNAL SLATME
    CALL SLATME(ADUMMA)  ! Implicit interface call
END

SUBROUTINE SLATME(EI)
    CHARACTER EI(*)      ! Assumed-size CHARACTER array
    ! EI bounds were garbage here
END
```

## Root Cause

Implicit interfaces used PointerArray for all arrays. This works for numeric arrays (passed as simple pointers), but CHARACTER assumed-size arrays require DescriptorArray for proper bounds handling.

## Solution

Use DescriptorArray physical type specifically for CHARACTER arrays in implicit interfaces, while keeping PointerArray for other arrays to maintain BindC compatibility with separately-compiled code.

Key insight from review: BindC ABI must be preserved for binary compatibility with separately-compiled Fortran code (like LAPACK). Only the physical type for CHARACTER arrays changes.

## Test Plan

- [x] New integration test `character_16.f90` - MRE from issue
- [x] Non-CHARACTER arrays unchanged (PointerArray + BindC)

Fixes #9381

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -954,16 +954,20 @@ namespace LCompilers {
                     n_dims, a_kind, is_array_type, arg->m_intent,
                     module, false);
                 if (ASRUtils::get_FunctionType(x)->m_abi == ASR::abiType::BindC && is_array_type) {
-                    // For bind(c) array dummies (including implicit interfaces), the ABI is
-                    // a raw pointer. `get_arg_type_from_ttype_t()` already returns the
-                    // correct pointer type, so do not add an extra level of indirection.
-                    type = type_original;
-                    is_array_type = false;
-                    if (ASRUtils::is_character(*arg->m_type)) {
-                        // For bind(c) character arrays, use raw i8* instead of any
-                        // string/descriptor representation.
+                    // For bind(c) array dummies (including implicit interfaces), handle
+                    // based on the physical type specified in the ASR.
+                    ASR::array_physical_typeType phys_type = ASRUtils::extract_physical_type(arg->m_type);
+                    if (phys_type == ASR::array_physical_typeType::DescriptorArray) {
+                        // DescriptorArray: pass pointer to descriptor struct
+                        type = type_original->getPointerTo();
+                    } else if (ASRUtils::is_character(*arg->m_type)) {
+                        // PointerArray CHARACTER: use raw i8*
                         type = llvm::Type::getInt8Ty(context)->getPointerTo();
+                    } else {
+                        // PointerArray non-CHARACTER: use element pointer
+                        type = type_original;
                     }
+                    is_array_type = false;
                 } else if (is_array_type) {
                     type = type_original->getPointerTo();
                 } else {


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10532,7 +10542,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     current_scope, s2c(al, arg_name), variable_dependencies_vec.p,
                     variable_dependencies_vec.size(), ASRUtils::intent_unspecified,
                     nullptr, nullptr, ASR::storage_typeType::Default, var_type, ASRUtils::get_struct_sym_from_struct_expr(var_expr),
-                    ASR::abiType::BindC, ASR::Public, ASR::presenceType::Required,
+                    ASR::abiType::Source, ASR::Public, ASR::presenceType::Required,
```
certik: I think this might be problematic. The BindC approach ensures binary compatibility when you compile code separately. With Source, we have to have access to the source code, which we often do not. I think Source gives us the ability to choose any ABI we want. But we actually can't choose any ABI we want here, because we must be precisely ABI compatible with a separately-compiled file.

What is the motivation to go from BindC to Source for these implicit interfaces? Let me understand the core issue first.


For character arrays I think we need special handling, not just a C pointer. For other arrays I think we still want to keep BindC?

certik:
It fails 3rd-party tests with:
```
  %506 = getelementptr inbounds float, float* %503, i32 %505
 float**  call void @sgels(%string_descriptor* %trans, i32* %m, i32* %n, i32* %nrhs, float* %a, i32* %lda, float* %b, i32* %ldb, float* %506, i32* %call_arg_value77, i32* %info)
  %557 = getelementptr inbounds float, float* %554, i32 %556
 float**  call void @sgetsls(%string_descriptor* %trans, i32* %m, i32* %n, i32* %nrhs, float* %a, i32* %lda, float* %b, i32* %ldb, float* %557, i32* %call_arg_value103, i32* %info)
  %923 = getelementptr inbounds float, float* %920, i32 %922
 float**  call void @sgels(%string_descriptor* %trans, i32* %m, i32* %n, i32* %nrhs, float* %a, i32* %lda, float* %b, i32* %ldb, float* %923, i32* %lwork, i32* %info)
  %1661 = getelementptr inbounds float, float* %1658, i32 %1660
 float**  call void @sgetsls(%string_descriptor* %trans, i32* %m, i32* %n, i32* %nrhs, float* %a, i32* %lda, float* %b, i32* %ldb, float* %1661, i32* %lwork, i32* %info)
```

krystophny:
@certik Thanks for the review. I've updated the PR based on your feedback:

1. **Reverted ABI back to BindC** - You're right that BindC is needed for binary compatibility with separately-compiled code. The original PR incorrectly changed to Source ABI which broke LAPACK third-party tests.

2. **Only CHARACTER arrays use DescriptorArray** - For CHARACTER assumed-size arrays, we use DescriptorArray physical type for proper bounds handling. All other arrays continue using PointerArray.

3. **Removed the old LAPACK patching commits** - Those were from a different branch that has since been merged to main.

The fix is now minimal:
- Keep BindC ABI (unchanged)
- Use DescriptorArray for CHARACTER arrays only
- Keep PointerArray for non-CHARACTER arrays (unchanged)

This should fix the original issue (#9381) while maintaining compatibility with LAPACK and other separately-compiled Fortran code.

PR #9386, opened 2026-01-05
fix: stdin list-directed READ skips blank lines

krystophny:
Summary:
- Fix list-directed READ from stdin so blank records are skipped (F2018 13.10.2.1).
- Adds a regression test that pipes stdin with blank lines.

Verification:
- ./run_tests.py -t stdin_read_blank_lines_01
- ./integration_tests/run_tests.py -b llvm -j 16


ci/test_third_party_codes.sh
```diff
@@ -913,6 +913,16 @@ time_section "🧪 Testing Reference-LAPACK v3.12.1 with BUILD_TESTING" '
...
 
+    # Patch schkqp3rk.f to fix uninitialized RESULT(4) bug (upstream LAPACK bug)
+    # Bug: RESULT(4) is only set to BIGNUM on failure, never initialized to ZERO
+    # This causes flaky test failures when garbage stack value >= THRESH
+    # See: https://github.com/lfortran/lfortran/issues/9371
+    sed -i '/IF( MIN(KFACT, MINMN).GE.2 ) THEN/i\                     RESULT( 4 ) = ZERO' \
+        ../TESTING/LIN/schkqp3rk.f
+
+    # Rebuild xlintsts with the patched source
+    cmake --build . --target xlintsts -j8
```
certik: Why not patch this before the main build, so that you don't need to rebuild?

certik (approved):
I think this is good.

ci/test_third_party_codes.sh
certik: Also probably better to maintain this patch as a `lapack.patch` file and just apply it using `apply`.

certik:
See if this passes CI first, we can fix the above later.

ci/test_third_party_codes.sh
```diff
@@ -913,6 +913,16 @@ time_section "🧪 Testing Reference-LAPACK v3.12.1 with BUILD_TESTING" '
 
     cd build
 
+    # Patch schkqp3rk.f to fix uninitialized RESULT(4) bug (upstream LAPACK bug)
+    # Bug: RESULT(4) is only set to BIGNUM on failure, never initialized to ZERO
+    # This causes flaky test failures when garbage stack value >= THRESH
+    # See: https://github.com/lfortran/lfortran/issues/9371
+    sed -i '/IF( MIN(KFACT, MINMN).GE.2 ) THEN/i\                     RESULT( 4 ) = ZERO' \
```
certik: This syntax is incorrect:
```
ci/test_third_party_codes.sh: line 920: syntax error near unexpected token `('
```
Just use a regular patch file like I suggested below.

krystophny: Fixed, this is now not needed anymore since LAPACK version bump.

certik (approved):

PR #9387, opened 2026-01-06
fix: Allow blank common blocks to vary in size between program units

### LINKED ISSUE #9378: Blank COMMON size varying between program units is allowed by the Standard
Per section 8.10.2.5 of F2023, all the way back to Fortran 66 (section 7.2.1.3), blank `common` need not be the same size between program units.  Lfortran issues an error message anyway.  The correct behavior is to distinguish between named and blank `common`, and only issue the message for named `common`.
```
program common_size
  implicit none

  ! All Fortran standards allow the size of blank common to vary between
  ! program units - from F66 (7.2.1.3) to F2023 (8.10.2.5).

  real :: pi, epsilon, sq2
  common // pi, epsilon, sq2

  pi = 3.1415654
  epsilon = 2.718281828
  sq2 = 1.414213562

  call print_pi ()
  call print_eps ()
  call print_sq2 ()

end program

subroutine print_pi ()
  implicit none

  real :: pi
  common  pi

  print *, 'pi =', pi

end subroutine

subroutine print_eps ()
  implicit none

  real :: dummy, eps
  common  dummy, eps

  print *, 'epsilon =', eps

end subroutine

subroutine print_sq2 ()
  implicit none

  real :: dummy(2), sq2
  common  dummy,    sq2

  print *, 'sqrt (2.0) =', sq2

end subroutine
```
```
$ gfortran common_size.F90
$ ./a.out
 pi =   3.14156532    
 epsilon =   2.71828175    
 sqrt (2.0) =   1.41421354    
$ lfortran common_size.F90
semantic error: COMMON block storage size mismatch: this declaration has 4 bytes but previous declaration has 12 bytes
  --> common_size.F90:24:3
   |
24 |   common  pi
   |   ^^^^^^^^^^ 


$ 
```

opixdown:
fixes #9378

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2484,6 +2484,7 @@ RUN(NAME common_24 LABELS gfortran llvm llvm2
     EXTRAFILES common_24a.f90 common_24b.f90
     EXTRA_ARGS --implicit-interface --separate-compilation)
 RUN(NAME common_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME common_26 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME common_substring_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME statement_01 LABELS llvm gfortran llvmImplicit)
--- a/integration_tests/common_26.f90
+++ b/integration_tests/common_26.f90
@@ -1,0 +1,50 @@
+program common_26
+  implicit none
+
+  ! All Fortran standards allow the size of blank common to vary between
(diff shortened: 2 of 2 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -3366,9 +3366,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 			common_variables_byte_offset[hash] = byte_offset;
 			byte_offset += get_type_byte_size(var__->m_type);
 		    }
-
+            constexpr char BLANK_COMMON_BLOCK[] = "blank#block";
```
certik, line 3369: I think this is matching the same constant somewhere else in this file, so the constant should be global and reused.

Issue #8548, opened 2025-09-07
A test case for `class(*), dimension(..)`


certik:
```fortran
module example_mod
  implicit none
contains
  subroutine check_arg(arg)
    class(*), dimension(..), intent(in) :: arg

    select rank (assoc => arg)
      rank (0)
        select type (assoc)
          type is (integer)
            print *, "Received a scalar integer: ", assoc
          class default
            print *, "Received a scalar of some other type"
        end select
      rank (1)
        select type (assoc)
          type is (integer)
            print *, "Received an array of integers (rank: 1)"
            ! If needed, access as 1D: print *, assoc
          class default
            print *, "Received an array of some other type"
        end select
      rank (2)
        select type (assoc)
          type is (integer)
            print *, "Received an array of integers (rank: 2)"
            ! If needed, access as 2D: print *, assoc
          class default
            print *, "Received an array of some other type"
        end select
      rank default
        print *, "Unsupported array rank: ", rank(arg)
    end select
  end subroutine check_arg
end module example_mod

program test
  use example_mod
  implicit none
  integer :: scalar_int = 42
  integer :: int_array1(3) = [1, 2, 3]
  integer :: int_array2(2,2) = reshape([1,2,3,4], [2,2])

  call check_arg(scalar_int)    ! Outputs: Received a scalar integer: 42
  call check_arg(int_array1)    ! Outputs: Received an array of integers (rank: 1)
  call check_arg(int_array2)    ! Outputs: Received an array of integers (rank: 2)
end program test
```
This should print:
```console
Received a scalar integer:           42
Received an array of integers (rank: 1)
Received an array of integers (rank: 2)
```
As in GFortran.

difference-scheme:
Holy  [nesting](https://www.youtube.com/watch?v=8bZh5LMaSmE), Batman!

HarshilShah1804:
Currently, we are able to compile and run the test with a small workaround of removing the assoc statement `(assoc => arg)`. Below is the modified version:
```fortran
module example_mod
  implicit none
contains
  subroutine check_arg(arg)
    class(*), dimension(..), intent(in) :: arg

    select rank (arg)
      rank (1)
        select type (arg)
          type is (integer)
            print *, "Received an array of integers (rank: 1)"
            ! If needed, access as 1D: print *, arg
          class default
            print *, "Received an array of some other type"
        end select
      rank (2)
        select type (arg)
          type is (integer)
            print *, "Received an array of integers (rank: 2)"
[...]

PR #9390, opened 2026-01-06
fix: load unlimited polymorphic type pointers correctly for select type

HarshilShah1804:
Towards #8548 

Now, except for `rank(0)` block, we can compile and run the test without any modifications

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -786,6 +786,7 @@ RUN(NAME select_rank_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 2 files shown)


certik:
Excellent. Get the CI passing, then I'll do a final review.

HarshilShah1804:
For `rank(0)` on my local system, I have managed to get it compiled. But it is giving a segmentation fault. I am working on fixing that. 

Also, CI would probably pass, I have tested it thoroughly on my local system.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -16281,7 +16322,7 @@ Result<std::unique_ptr<LLVMModule>> asr_to_llvm(ASR::TranslationUnit_t &asr,
         v.module->print(os, nullptr);
         std::cout << os.str();
         msg = "asr_to_llvm: module failed verification. Error:\n" + err.str();
-        std::cout << msg << std::endl;
+        std::cout << msg << std::endl;
```
certik, line 16325: No need to add the trailing whitespace.

certik (approved):
I think it's fine. Let's finish the exhaustive tests, just to be safe.

src/libasr/codegen/asr_to_llvm.cpp
HarshilShah1804: Sure, will take care next time onwards.

Issue #9388, opened 2026-01-06
Handle multi-length string passed to function accepting less length string


jinangshah21:
MRE:
```fortran
program isdigit_mre
    implicit none
    character(len=4) :: test_char = "2021"
    print *, isdigit(test_char)
contains
    logical function isdigit(c)
        character(len=1), intent(in) :: c
        isdigit = index("12",c)>0
    end function isdigit
end program isdigit_mre
```

Outputs:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran d.f90  && ./a.out
 T
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran d.f90
F
```

PR #9391, opened 2026-01-06
fix: handle character length mismatch in function arguments

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9388


src/libasr/asr_utils.h
```diff
@@ -6623,6 +6623,38 @@ static inline void Call_t_body(Allocator& al, ASR::symbol_t* a_name,
...
+        if( func_type->m_abi != ASR::abiType::BindC &&
+            ASRUtils::is_string_only(orig_arg_type) &&
+            ASRUtils::is_string_only(arg_type) ) {
+            ASR::String_t* arg_str_type = get_string_type(arg_type);
+            ASR::String_t* orig_arg_str_type = get_string_type(orig_arg_type);
+            
+            if (arg_str_type->m_len_kind == ASR::string_length_kindType::ExpressionLength &&
+                orig_arg_str_type->m_len_kind == ASR::string_length_kindType::ExpressionLength &&
+                arg_str_type->m_len && orig_arg_str_type->m_len &&
+                ASRUtils::is_value_constant(arg_str_type->m_len) &&
```
jinangshah21, line 6636: Here `arg` may not have constant length. 
```fortran
character(len=:), allocatable :: test_char
test_char = "2021"
print *, isdigit(test_char)
```

I think we need to create `stringView` at runtime `convert_call_args`.

certik:
Looks like more work is needed here.

certik:
Conflicts have to be resolved, sorry about that.

certik:
@jinangshah21 can you please review it?

jinangshah21 (approved):
I think this is fine! 
Can you add test case where small length string is passed to function expecting more length in upcoming PR

certik:
It looks like that the CI will pass. I'll wait to get a few more exhaustive tests passing, then we can merge.

assem2002:
looks good! and it's good call to use stringsections -- it's the only valid fix in my mind now.
I had a use case in mind that could fail with this fix 
```.f90
program p
    character(10),allocatable :: str
    allocate(str)
    call ss(str) ! This PR makes it equivalent to `call ss(str(1:5))` -- gfortran forbids btw
    print *, allocated(str) ! prints FALSE
    contains 
    subroutine ss(s)
        character(5), allocatable :: s
        deallocate(s)
    end subroutine

end program 
```

We should raise an error with this use case (though gfortran allows it), because we have to forbid the user from depending on that feature. Let's push a PR with that error, I think no code in our test suite uses that case, and no code should use it from the first place (What if we reallocate the string inside the function with length 5?).

certik:
Two followups to fix:

* https://github.com/lfortran/lfortran/pull/9391#pullrequestreview-3639859501
* https://github.com/lfortran/lfortran/pull/9391#issuecomment-3725178622

Issue #8757, opened 2025-10-27
Bug : arraySlice on module's parameter-array variable


assem2002:
```.f90
module mm
  integer,parameter :: arr(3) = [1,2,3]
end module mm
program pp
  use mm
  integer,parameter :: arr2(3) = [1,2,3]
  print *, arr(1:1)  ! Error
  print *, arr2(1:1) ! Works
end program
```
snippet of the stack-trace error.
```console
  File "/home/assem/LFortranUpStream/lfortran/src/libasr/pass/array_op.cpp", line 405, in LCompilers::ArrayOpVisitor::call_replacer()
    replacer.replace_expr(*current_expr);
  File "/home/assem/LFortranUpStream/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 2386, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplaceArrayOp>::replace_expr(LCompilers::ASR::expr_t*)
    self().replace_ArraySection(down_cast<ASR::ArraySection_t>(x));
  File "/home/assem/LFortranUpStream/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 1180, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplaceArrayOp>::replace_ArraySection(LCompilers::ASR::ArraySection_t*)
    self().replace_expr(x->m_v);
  File "/home/assem/LFortranUpStream/lfortran/src/libasr/../libasr/asr_expr_base_replacer_visitor.h", line 2378, in LCompilers::ASR::BaseExprReplacer<LCompilers::ReplaceArrayOp>::replace_expr(LCompilers::ASR::expr_t*)
    self().replace_ArrayConstant(down_cast<ASR::ArrayConstant_t>(x));
AssertFailed: result_expr != nullptr

```

harperjf:
Relevant issues here may be #4726 (bug fixed now though its variable ```nkr``` uses sections of a parameter array) and #8324.

assem2002:
thanks @harperjf good insights!

PR #9395, opened 2026-01-06
fix: skip array constant replacement when result_expr is null

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8757


certik (approved):
I think that's fine.

PR #9397, opened 2026-01-06
fix: don't change allocatable to pointer arrays in nested_vars

### LINKED ISSUE #9396: Handle extension of nested subroutine allocatable array var
MRE:
```fortran
module tmp_mod
    type :: tmp_type
        integer :: a
    end type tmp_type

contains
    subroutine tmp_sub()
        type(tmp_type), allocatable :: x(:)
        allocate(x(1))
        call inner_sub()
        print *, x(2)%a
    contains
        subroutine inner_sub()
            x = [x, tmp_type(0)]
            x(2)%a = 42
        end subroutine inner_sub
    end subroutine tmp_sub
end module tmp_mod

program main
    use tmp_mod
    implicit none
    call tmp_sub()
end program main
```

LFortran Error:
```console
At 14:13 of file d.f90
Runtime Error: Array shape mismatch in assignment to '__lcompilers_created__nested_context__tmp_sub_x'

Tried to match size 1 of dimension 1 of LHS with size 2 of dimension 1 of RHS.
```

jinangshah21:
Fixes #9396

### DIFF
--- a/src/libasr/pass/nested_vars.cpp
+++ b/src/libasr/pass/nested_vars.cpp
@@ -814,16 +814,127 @@ public:
                             ASR::is_a<ASR::Function_t>(*ASRUtils::symbol_get_past_external
                             (ASR::down_cast<ASR::Variable_t>(sym_)->m_type_declaration));
                         if( ASRUtils::is_array(ASRUtils::symbol_type(sym)) || ASRUtils::is_pointer(ASRUtils::symbol_type(ext_sym)) ) {
-                            ASR::stmt_t *associate = ASRUtils::STMT(ASRUtils::make_Associate_t_util(al, t->base.loc,
-                                                        target, val));
-                            body.push_back(al, associate);
-                            // TODO : Remove the following if block (See integration test `arrays_87.f90`)
-                            if(ASRUtils::is_array(ASRUtils::symbol_type(sym)) &&
-                                is_ext_sym_allocatable_or_pointer && is_sym_allocatable_or_pointer
-                                && ASRUtils::EXPR2VAR(val)->m_storage != ASR::storage_typeType::Parameter ) {
-                                associate = ASRUtils::STMT(ASRUtils::make_Associate_t_util(al, t->base.loc,
-                                    val, target));
-                                assigns_at_end.push_back(associate);
+                            if( ASRUtils::is_allocatable(ASRUtils::symbol_type(sym)) && ASRUtils::is_allocatable(ASRUtils::symbol_type(ext_sym)) ) {
+                                // TODO: Improve this by using Associate only when no reallocation is needed
+                                // For allocatable arrays, use Assignment instead of Associate
+                                // to properly handle reallocation in nested functions
+                                // Wrap in if(allocated) check since the array may not be allocated yet
+                                ASR::stmt_t *assignment = ASRUtils::STMT(ASRUtils::make_Assignment_t_util(al, t->base.loc,
+                                                            target, val, nullptr, true, false));
+                                // First push allocate stmt for LHS
+                                ASRUtils::ASRBuilder builder(al, t->base.loc);
+                                Vec<ASR::expr_t*> to_be_deallocated;
+                                to_be_deallocated.reserve(al, 1);
+                                to_be_deallocated.push_back(al, target);
+                                body.push_back(al, ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(
+                                    al, t->base.loc, to_be_deallocated.p, to_be_deallocated.size())));
+                                ASRUtils::ExprStmtDuplicator d(al);
+                                ASR::expr_t* realloc_var = d.duplicate_expr(ASRUtils::get_expr_size_expr(val));
+                                ASR::dimension_t* m_dims;
+                                int n_dims = ASRUtils::extract_dimensions_from_ttype(ASRUtils::symbol_type(sym), m_dims);
+                                Vec<ASR::dimension_t> dims_vec;
+                                dims_vec.reserve(al, n_dims);
+                                for( int i = 0; i < n_dims; i++ ) {
+                                    ASR::dimension_t dim;
+                                    dim.loc = m_dims[i].loc;
+                                    dim.m_start = builder.i32(1);
+                                    dim.m_length = ASRUtils::EXPR(ASR::make_ArraySize_t(
+                                        al, dim.loc, realloc_var, builder.i32(i + 1), 
+                                        ASRUtils::TYPE(ASR::make_Integer_t(al, dim.loc, 4)), nullptr));
+                                    dims_vec.push_back(al, dim);
+                                }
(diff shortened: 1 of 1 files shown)


certik:
This test failed:
```
class_75_STD_F23
```
Not sure if it is caused by this PR or not.

certik:
The following failed in LLVM 20:
```
	1423 - nested_22_FAST (SEGFAULT)
```
I think it's worth investigating, since it was added in this PR.

integration_tests/CMakeLists.txt
```diff
@@ -742,7 +742,7 @@ RUN(NAME arrays_83 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_84 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME arrays_85 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_86 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME arrays_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME arrays_87 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
```
certik: That doesn't seem right. I can't see anything in `arrays_87` that requires LHS reallocation. Can you point me to the line in that file that requires it?

integration_tests/CMakeLists.txt
```diff
@@ -1410,7 +1409,7 @@ RUN(NAME modules_45 LABELS gfortran)
 RUN(NAME modules_47 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_STD_F23)
 RUN(NAME modules_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME modules_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME modules_50 LABELS gfortran llvm)
+RUN(NAME modules_50 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
```
certik: This file also doesn't seem to need it, it contains:
```
modules_used = [modules_used]
        ! modules_used = [modules_used, modules_used(1)]
```
Which I am guessing would require it if the commented line was uncommented, but as it is written, it shouldn't require it.

integration_tests/CMakeLists.txt
```diff
@@ -1964,6 +1963,7 @@ RUN(NAME nested_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_21 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME nested_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
```
certik, line 1966: This file requires it for `x = [x, tmp_var]`, so that's fine.

integration_tests/CMakeLists.txt
jinangshah21: Consider transformation after `pass_nested_vars`:
```fortran
!! temp_modules_used = global temporary 
subroutine prune_build_targets()
    type(int_t), allocatable :: modules_used(:)

    allocate(modules_used(1))

    temp_modules_used = modules_used    !!!! Here we need `--realloc-lhs-arrays`
    call collect_used_modules()
    module_used = temp_modules_used

    contains

    recursive subroutine collect_used_modules()
        temp_modules_used(1)%i = 2
        temp_modules_used = [temp_modules_used]
    end subroutine collect_used_modules
```

This is similar case with `arrays_87`

certik: So that seems like a bug in pass_nested_vars to not allocate the array properly. Also, ideally we should not copy arrays at all, but use pointers. But if we need to copy, then we need to allocate properly. We should not introduced the `--realloc-lhs-arrays` dependency in a pass from ASR that didn't need it.

src/libasr/pass/nested_vars.cpp
```diff
@@ -810,16 +814,105 @@ private :
...
-                                is_ext_sym_allocatable_or_pointer && is_sym_allocatable_or_pointer
-                                && ASRUtils::EXPR2VAR(val)->m_storage != ASR::storage_typeType::Parameter ) {
-                                associate = ASRUtils::STMT(ASRUtils::make_Associate_t_util(al, t->base.loc,
-                                    val, target));
-                                assigns_at_end.push_back(associate);
+                            if( ASRUtils::is_allocatable(ASRUtils::symbol_type(sym)) && ASRUtils::is_allocatable(ASRUtils::symbol_type(ext_sym)) ) {
+                                // TODO: Improve this by using Associate only when no reallocation is needed
+                                // For allocatable arrays, use Assignment instead of Associate
+                                // to properly handle reallocation in nested functions
+                                // Wrap in if(allocated) check since the array may not be allocated yet
```
certik, line 821: Right now this does a copy for any allocatable array, correct?

Can we only do the copy for allocatable arrays if we actually change the allocation status inside the nested function? Or is that too hard for this PR?

jinangshah21: It think this may be done if the array doesn't appear at LHS of any assignment inside nested subroutine

certik:
I will need to play with it by hand to see exactly the behavior. If this slows down any nested function that uses an array that happens to be allocatable, then I think we have a problem and will have to find a solution before merging I think, since these kinds of things will be hard to fix later.

certik:
I didn't have time yet, will investigate this hopefully later today.

integration_tests/nested_22.f90
```diff
@@ -0,0 +1,27 @@
...
+        type(tmp_type), allocatable :: x(:)
+        allocate(x(1))
+        call inner_sub()
+        print *, x(2)%a
+        if (size(x) /= 2) error stop
+        if (x(2)%a /= 42) error stop
+    contains
+        subroutine inner_sub()
+            type(tmp_type) :: tmp_var
+            x = [x, tmp_var]
```
certik, line 17: We want to make this case fast as well (no copy!):
```suggestion
            x(:) = 3
```
And I think to make it fast, maybe we can just use `move_alloc` both before and after the `inner_sub` is called. The `move_alloc` can be implemented by "stealing" the allocation and memory, no data copy.

certik:
So the answer seems to be that we currently always do a data copy of the arrays if they are allocatable in this PR. However it seems we can fix it by doing `move_alloc` instead of this copy, this PR (both before and after), and only for allocatable arrays.

certik:
This was fixed by https://github.com/lfortran/lfortran/pull/9549.

PR #9400, opened 2026-01-06
fix: preserve member subscripts in name resolution

certik:
Handle reduced member counts safely without dropping member array args in visit_NameUtil, fixing derived-type array member access while keeping single- member access safe. Regression confirmed with derived_types_43 and multi-file member access.

The MRE for this is only triggered with `lfortran file1.f90 file2.f90`, our integration tests don't have a workflow for this, so I didn't add a test yet.

### DIFF
--- a/.github/workflows/Quick-Checks-CI.yml
+++ b/.github/workflows/Quick-Checks-CI.yml
@@ -240,6 +240,7 @@ jobs:
             ./run_tests.py -m
             ./run_tests.py -b llvm
             ./run_tests.py -b llvm -f -nf16
+            ./run_tests.py -b llvm_single_invocation
 
       - name: Test GFortran, Debug Build, Fortran, OpenMP, C/C++ backend, Upload Tarball, CPP Build, WASM - Test OpenMP
         shell: bash -e -l {0}


certik:
The MRE is:
```console
$ cat file1.f90
module jtypes
   implicit none

contains
   subroutine dummy(message, line)
      character(len=*), intent(in), optional :: message
      integer, intent(in), optional :: line
      if (present(line)) error stop 1
   end subroutine
end module jtypes
```
and
```console
$ cat file2.f90
module jstringbuffer
   implicit none

   type :: StringBuffer
      integer, private :: size = 0
   contains
      procedure :: number_of_elems
   end type StringBuffer

contains

   function number_of_elems(self) result(result)
      class(StringBuffer), intent(in) :: self
      integer :: result
      result = self%size
   end function

end module jstringbuffer

program main
print *, "OK"
end program
```

certik:
Getting this error in fpm 3rd-party tests:
```
[ 59%]       fpm_compile_commands.F90  done.
<ERROR> Compilation failed for object " src_fpm_compile_commands.F90.o "
STOP 1

style suggestion: Use 'end if' instead of 'endif'
  --> ././src/fpm_compile_commands.F90:78:9
   |
78 |         endif
   |         ^^^^^ help: write this as 'end if'

semantic error: Arguments do not match for any generic procedure, load
   --> ././src/fpm_compile_commands.F90:406:17
    |
406 |                 call self%command(i)%load(elem, error)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```
We'll need to extract an MRE and fix it.

certik:
Tests were added, everything passes. We can iterate on this further, but I think it's good enough to merge.

PR #9401, opened 2026-01-06
fix: fold real/int to complex PARAMETER init

### LINKED ISSUE #9399: codegen: negative real literal to COMPLEX PARAMETER generates invalid LLVM IR
## Description

When a negative real literal is assigned to a COMPLEX PARAMETER, LFortran generates invalid LLVM IR that fails verification.

## Minimal Reproducer

```fortran
subroutine test_complex_param
    complex, parameter :: ONE = 1.0E+0
    complex, parameter :: NEG_ONE = -1.0E+0
    complex, parameter :: ZERO = 0.0E+0
    print *, ONE, NEG_ONE, ZERO
end subroutine
```

## Commands

```bash
lfortran --fixed-form-infer --implicit-interface --implicit-typing --separate-compilation test.f90
```

## Error

```
  store float -1.000000e+00, %complex_4* %neg_one, align 4
 %complex_4 = type <{ float, float }>
```

## Analysis

Looking at the generated LLVM IR:
- `ONE` correctly generates: `store %complex_4 <{ float 1.000000e+00, float 0.000000e+00 }>, %complex_4* %one`
- `ZERO` correctly generates: `store %complex_4 zeroinitializer, %complex_4* %zero`
- `NEG_ONE` incorrectly generates: `store float -1.000000e+00, %complex_4* %neg_one` (stores scalar, not complex struct)

The issue is that the UnaryOp negation path for PARAMETER initialization doesn't properly wrap the result in a complex type when the target is COMPLEX.

## Impact

This blocks building LAPACK with `BUILD_COMPLEX=ON` - fails at ~51% of the build on files like `SRC/clarft.f` which use:
```fortran
COMPLEX           ONE, NEG_ONE, ZERO
PARAMETER(ONE=1.0E+0, ZERO = 0.0E+0, NEG_ONE=-1.0E+0)
```

## Expected Behavior

Per Fortran standard, assigning a real value to a complex variable should implicitly convert to `(real_value, 0.0)`.

krystophny:
## Summary

Fix `COMPLEX, PARAMETER` initialization from constant `REAL`/`INTEGER` expressions (including negative literals like `-1.0`) by folding the conversion in semantics/ASR construction, so codegen/backends only need to lower a well-typed constant.

## Problem

Valid Fortran like:

```fortran
complex :: neg_one
parameter (neg_one = -1.0)
```

should produce `neg_one = (-1.0, 0.0)`. Previously, LFortran could end up with a `COMPLEX` parameter whose constant value was still a scalar `REAL` (or a cast node whose precomputed `m_value` was not a `ComplexConstant`), which then caused incorrect lowering.

## Fix (ASR-level, clean lowering)

In `src/lfortran/semantics/ast_common_visitor.h`, when finalizing a `PARAMETER` variable:

- If the target type is `COMPLEX` and the initializer is a **constant** `REAL`/`INTEGER` expression, wrap it using `ASRUtils::make_Cast_t_value(... RealToComplex/IntegerToComplex ...)`.
- This guarantees the cast carries a proper `ComplexConstant` value `(re, 0.0)` (including for negative literals).

Minimal sketch:

```c++
if (is_complex(v->m_type) && !is_complex(*expr_type(init_val)) && expr_value(init_val)) {
    init_val = make_Cast_t_value(... RealToComplex/IntegerToComplex ...);
}
```

## Tests

- Added `integration_tests/parameter_16.f90` and registered it.

Fixes #9399.


HarshitaKalani (approved):
lgtm. Thanks

Issue #9382, opened 2026-01-05
Stdin list-directed READ fails on blank lines (file I/O works)


krystophny:
List-directed READ from stdin (`READ(*,*)`) fails when input contains blank lines, even though the Fortran standard requires blank records to be skipped. File I/O works correctly.

## Minimal Reproducing Example

```fortran
program test_stdin
    implicit none
    integer :: m, n, i
    do i = 1, 2
        read(*,*) m
        read(*,*) n
        print *, 'Loop', i, ': m=', m, ' n=', n
    end do
end program
```

## Commands

```bash
echo -e "10\n5\n\n20\n10" | lfortran test.f90
```

Input breakdown:
```
10       <- m for loop 1
5        <- n for loop 1
         <- blank line (should be skipped)
20       <- m for loop 2
10       <- n for loop 2
```

## Expected Output

```
Loop 1 : m= 10  n= 5
Loop 2 : m= 20  n= 10
```

## Actual Output

```
Loop 1 : m= 10  n= 5
Error: Invalid input for int32_t.
```

## Note: File I/O Works Correctly

The same input read from a file (not stdin) works:

```fortran
program test_file
    implicit none
    integer :: m, n, i
    open(10, file='input.txt')
    do i = 1, 2
        read(10, *) m
        read(10, *) n
        print *, 'Loop', i, ': m=', m, ' n=', n
    end do
end program
```

With `input.txt` containing the same data (including blank line), this correctly outputs:
```
Loop 1 : m= 10  n= 5
Loop 2 : m= 20  n= 10
```

## Context

This blocks the LAPACK DMD (Dynamic Mode Decomposition) tests (sdmd.in, ddmd.in) which read test parameters from stdin with blank lines between test blocks.

## Fortran Standard Reference

F2018 13.10.2.1 states that blank records are skipped for list-directed input.

harperjf:
13.10.2.1 doesn't exist but 13.10.2 paragraph 1 says among other things "The end
of a record has the same effect as a blank character, unless it is within a character constant." That may be different from skipping a record, for example if the variable to be read is of type character.

krystophny:
> 13.10.2.1 doesn't exist but 13.10.2 paragraph 1 says among other things "The end of a record has the same effect as a blank character, unless it is within a character constant." That may be different from skipping a record, for example if the variable to be read is of type character.

[...]

PR #9404, opened 2026-01-07
Runtime: Fix stdin list-directed READ to skip blank lines (#9382)

Adityazzzzz:
fix issue #9382 

This PR fixes a runtime I/O regression where list-directed READ(*,*) from stdin would crash with "Invalid input" when encountering blank lines.
The Fix

- Replaced the line-based fgets/strtok implementation in _lfortran_read_int32 with a stream-based fscanf implementation.
- fscanf automatically consumes leading whitespace, including newlines (\n, \r), complying with standard Fortran list-directed I/O behavior.
- Added buffer overflow protection using %127s.

Impact
- Resolves the blocker for LAPACK DMD tests (sdmd.in, ddmd.in) which rely on reading parameters separated by blank lines.
- Ensures stdin behaves consistently with file I/O.

Testing
- Verified locally with the reproduction case provided in the issue.
- Confirmed that input 10\n5\n\n20\n10 is now read correctly as 10, 5, 20, 10.


Adityazzzzz:
@HarshitaKalani I have added a regression test (tests/test_issue_9382.f90) along with the corresponding .in file containing blank lines. I verified locally that the runtime handles the blank lines correctly now, producing the expected output. The CI should pass.
Please review, if any other changes required

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4521,29 +4521,29 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num, int32_t *io
 {
     if (iostat) *iostat = 0;
 
+    // --- START OF FIX FOR #9382 ---
```
certik: I think we don't need such comments --- every change in LFortran is some start of some fix. We can use git history to see how each line came to be. Otherwise I think it doesn't add much value.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4553,7 +4553,9 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num, int32_t *io
         *p = (int32_t)long_val;
         return;
     }
+    // --- END OF FIX ---
```
```suggestion
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4562,7 +4564,8 @@ LFORTRAN_API void _lfortran_read_int32(int32_t *p, int32_t unit_num, int32_t *io
         fprintf(stderr, "Internal Compiler Error: No file found with given unit number %d.\n", unit_num);
         exit(1);
     }
-
+    
+    // ... (rest of the function remains the same)
```
certik: Remove all these comments.

tests/test_issue_9382.f90
```diff
@@ -0,0 +1,9 @@
+program test_stdin_9382
+    implicit none
+    integer :: m, n, i
+    do i = 1, 2
+        read(*,*) m
+        read(*,*) n
+        print *, 'Loop', i, ': m=', m, ' n=', n
+    end do
+end program
```
certik, line 9: You need to register this file in tests/tests.toml. Then update reference tests via `./run_tests.py -u`.

tests/test_issue_9382.f90
HarshitaKalani, line 1: Can you please give a better name to the test? Something related to what it solves rather than issue number. Maybe something like read_*.f90

tests/tests.toml
```diff
@@ -4950,6 +4950,6 @@ filename = "../integration_tests/arrays_101.f90"
 llvm = true
 
 [[test]]
-filename = "stdin_read_blank_lines_01.f90"
+filename = "test_issue_9382.f90"
 run = true
-options = "< tests/stdin_read_blank_lines_01.in"
+options = "< tests/test_issue_9382.in"
```
HarshitaKalani, line 4955: Why are we removing stdin_read_blank_lines_01?

PR #9412, opened 2026-01-08
Fix broken tuple tests

swamishiju:
Some tuple tests were failing because deferred-length character elements in tuples were not properly initializing their string descriptors on the heap. 

This change ensures that for descriptor strings with deferred length in tuples:
- The descriptor is initialized to null.
- Memory is allocated on the heap for the string data as needed.

@assem2002  Could you please review if the malloc calls in `set_string_memory_on_heap` are not redundant or could lead to memory leaks?


swamishiju:
I'll push fix for LP after this is merged 👍

certik:
I resolved the conflicts and merged.

PR #9405, opened 2026-01-07
fix: recl with tests added

yuvrajsinghgmx:
Towards #9019 

@HarshitaKalani  i have added and updated the tests.

PR #9417, opened 2026-01-08
test: add and register error test for recl

yuvrajsinghgmx:
Adds error test for #9405


yuvrajsinghgmx:
Is this fine or i need to change the position of added to tests to make the diff minimal ?.

PR #9420, opened 2026-01-08
fix: array section shape mismatch diagnostic

krystophny:
## Summary
- Fix semantics: detect array section shape mismatch in assignment
- Keep `continue_compilation_1` line stability by moving the new array-section checks into the reserved blank lines near the end of the executable section

## Why
Array section assignments with mismatched shapes should be diagnosed reliably.

The `continue_compilation_1` error test is line-sensitive; keeping new checks in the reserved space avoids shifting existing diagnostics.

**Stage:** Semantics

## Changes
- [`src/lfortran/semantics/ast_to_asr.cpp#L1`](https://github.com/lfortran/lfortran/blob/4ed74717fe5690945b4879a89e10dc5761c9d412/src/lfortran/semantics/ast_to_asr.cpp#L1): semantics: array section shape mismatch assignment diagnostics
- [`tests/errors/continue_compilation_1.f90#L1`](https://github.com/lfortran/lfortran/blob/4ed74717fe5690945b4879a89e10dc5761c9d412/tests/errors/continue_compilation_1.f90#L1): move `cc_a3/cc_a4` slice checks into reserved blank lines near end

## Tests
- `asr-continue_compilation_1`

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -185,20 +185,9 @@ program continue_compilation_1
     integer :: cc_a4(2)
     integer :: cc_temp4(5)
     integer :: cc_i1 = 1
-    
-    
-    
-    
-    
-    
-    
-    
-    
-    
-    
-    
-    
-    
+
+
+
 
 
 
@@ -231,10 +220,6 @@ program continue_compilation_1
     a5 = 8
     b5 = 12_8
     c5 = 2
-    cc_a3 = cc_temp3(cc_i0:cc_i0)
-    print *, size(cc_a3)
-    cc_a4 = cc_temp4(cc_i1+1:cc_i1+1)
-    print *, size(cc_a4)
 
     !loop_test
     do i=1,3
@@ -438,14 +423,14 @@ program continue_compilation_1
     allocate(arr5, mold = arr4)
 
     print *, ["aa", "aaa"]
-
+    cc_a3 = cc_temp3(cc_i0:cc_i0)
     print *, pack(arr2, mask1)
-
+    print *, size(cc_a3)
     ! assigning to a *PROTECTED* struct instance member, not allowed
     protected_module_my_class_obj%value = 42
-
+    cc_a4 = cc_temp4(cc_i1+1:cc_i1+1)
     arr = [type(MyClass) :: v1, v2, v3]
-
+    print *, size(cc_a4)
     arr = [NonExistingType :: v1, v2, v3]
 
     !Data Statements with different number of arguments on LHS and RHS
@@ -469,8 +454,10 @@ program continue_compilation_1
     assign 13 to fmt_i3
     13 format ()
     read (5, fmt_i3)
+
     !passing non procedure to procedure parameter
     call proc_param(42)
+
     x = 9010
     read (*, end=x) x
     read (*, end=9011.0) x
@@ -481,8 +468,10 @@ program continue_compilation_1
 9014 continue
     write (*, err=9015) x
 9015 continue
+
     read(*, *, end=999) x   
     read(*, *, err=500) x
+    
     OPEN(unit=10, recl=10, recl=20)
     OPEN(unit=10, recl=10.5)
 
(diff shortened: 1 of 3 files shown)


krystophny:
@HarshitaKalani thx! (you see I do a few small fixes now, Claude was a bit too eager to interpret them as relevant for beta in #3806) with the goal of getting them out of the way so you can focus on the larger blockers.

krystophny:
Added some more generalization.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3712,11 +3712,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                         int dim_b_int {-1};
                         // 'm_length' isn't assigned for allocatable arrays
                         // let them be valid for now atleast
```
certik: These two comments can be removed now I think.

tests/reference/run-array_shape_05-429216b.stderr
```diff
@@ -1,4 +0,0 @@
-At 5:5 of file tests/errors/array_shape_03.f90
-Runtime Error: Array shape mismatch in assignment to 'a'
-
-Tried to match size 2 of dimension 1 of LHS with size 1 of dimension 1 of RHS.
```
certik, line 4: We now lost the test for the runtime check.

Let's figure out how to keep testing the runtime. One way to do it is via `temp(i:j)` and ensure the values of `i` and `j` are complicated enough that the compiler can't figure it out at compile time --- probably just a simple assignment would do. But in the worst case we can load it from a file.

We need to keep testing all the runtime checks.

krystophny: Fixed in `array_shape_05`

src/lfortran/semantics/ast_body_visitor.cpp
krystophny: Done

tests/errors/array_shape_05.f90
```diff
@@ -0,0 +1,12 @@
+program array_shape_05
+    implicit none
+    integer :: a(2)
+    integer :: temp(5)
+    integer :: i
+    integer :: j
+
+    call system_clock(count=i)
+    i = 1 + mod(i, size(temp))
```
certik: Why not just:
```suggestion
    i = 1
```
That doesn't work?

krystophny:
We keep i runtime (system_clock) so the slice length can’t be folded; setting i=1 would turn this into a compile-time shape error and wouldn’t test the runtime check.

certik:
I think
```
i = 1
j = i
a = temp(i:j)
```
would also work, since you are not doing a full program analysis, just a simple expression comparison. And if we later improve the analysis, we can make it more complicated later.

But it's fine I think.

krystophny:
> I think
> 
> ```
> i = 1
> j = i
> a = temp(i:j)
> ```
> 
> would also work, since you are not doing a full program analysis, just a simple expression comparison. And if we later improve the analysis, we can make it more complicated later.
> 
> But it's fine I think.

Since tests are anyway queued I replaced it now as you suggested.

tests/tests.toml
```diff
@@ -4696,6 +4696,14 @@ asr = true
 
 [[test]]
 filename = "errors/array_shape_03.f90"
+asr = true
+
+[[test]]
+filename = "errors/array_shape_04.f90"
+asr = true
+
```
certik: These should just go into continue_compilation. Can be done in the next PR also.

certik:
This is good. Just needs to merge main and regenerate the reference tests.

tests/errors/continue_compilation_1.f90
```diff
@@ -215,12 +212,15 @@ program continue_compilation_1
     ! Only put statements below. If you need to call a function, put it into a
     ! module above.
 
-    a = 1
-    print *, a(10)
```
certik, line 219: I think removing these is a mistake, we lost an error message below due to that.

tests/errors/continue_compilation_1.f90
```diff
@@ -221,6 +216,11 @@ program continue_compilation_1
     b5 = 12_8
     c5 = 2
 
+    cc_a3 = cc_temp3(cc_i0:cc_i0)
+    print *, size(cc_a3)
+    cc_a4 = cc_temp4(cc_i1+1:cc_i1+1)
+    print *, size(cc_a4)
```
certik: This shifted the above error, typically you just put these at the very end. But it's ok I think, let's just merge this PR finally.

krystophny: had to rebase anyway, managed to fit this still in the gaps at the end.

certik (approved):

PR #9431, opened 2026-01-09
Proper allocating for StrRepeat

swamishiju:
This commit just makes `StrRepeat` work properly for allocatable strings. This is an LPython only function which doesn't have tests for in LFortran.


certik:
This is fine --- do you think you could please add a test for this to LFortran? We already have a bunch of lpython intrinsics, so just add another one and add a test.

PR #9441, opened 2026-01-09
test: add test for issue #7780

### LINKED ISSUE #7780: Handle formatted read from a string
MRE:
```fortran
program read_nan_string
  implicit none
  character(len=3) :: nan_string
  real :: valu

  nan_string = 'NaN'
  read(nan_string, '(f3.3)') valu

  print *, 'Read value:', valu
end program read_nan_string
```

LFortran Error:
```console
  %4 = load i8*, i8** %nan_string, align 8
 i32  call void (i32, i32*, i32*, i8*, i8*, i32, ...) @_lfortran_formatted_read(i8* %4, i32* %5, i32* %6, i8* getelementptr inbounds ([4 x i8], [4 x i8]* @1, i32 0, i32 0), i8* getelementptr inbounds ([7 x i8], [7 x i8]* @2, i32 0, i32 0), i32 1, float* %valu)
```

opixdown:
fixes #7780

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2391,6 +2391,9 @@ RUN(NAME read_22 LABELS gfortran llvm)
 RUN(NAME read_23 LABELS gfortran llvm)
 RUN(NAME read_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
+
+RUN(NAME read_27 LABELS gfortran llvm)
+
 RUN(NAME write_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME write_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME write_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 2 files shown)


integration_tests/read_25.f90
```diff
@@ -0,0 +1,12 @@
+program read_nan_string
+  implicit none
+  character(len=3) :: nan_string
+  real :: valu
+
+  nan_string = 'NaN'
+  read(nan_string, '(f3.3)') valu
+
+  if (valu == valu) error stop "wrong value read"
```
certik: This should trigger a failure. I think this condition is incorrect.

certik: The tests passed, but I would expect them to fail, since the condition `valu == valu` should evaluate to true.

So something is wrong.

opixdown: You’re right  the test should only check that the READ doesn’t crash

certik: No, the test needs to test what read should read with an if statement.

certik:
Add an if test per https://github.com/lfortran/lfortran/pull/9441#discussion_r2678751205. Then it should be good. You can use some ieee intrinsic to obtain NaN, which should be the correct answer I think.

certik:
Looks good now, thanks!

PR #9444, opened 2026-01-10
parser: ignore ! directives unless --openmp

### LINKED ISSUE #6449: OpenMP directive is neither recognized nor treated as a comment
Code example:
```
module omp_module

   real (kind=8), dimension(:), allocatable :: arr
   !$omp threadprivate( arr )

end module omp_module

```
LFortran output:
```
syntax error: Token '!$omp threadprivate( arr )' (of type 'pragma') is unexpected here
 --> testlfort.f90:4:4
  |
4 |    !$omp threadprivate( arr )
  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```

krystophny:
## Summary
- Treat OpenMP sentinel directives ! ... as comments when OpenMP is not enabled (no --openmp).
- Add integration test openmp_74 covering ! threadprivate in a module with OpenMP disabled.

- Fixes #6449 

## Tests
- integration_tests/openmp_74

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -155,10 +155,13 @@ def single_test(test: Dict, verbose: bool, no_llvm: bool, skip_run_with_dbg: boo
                 verify_hash,
                 extra_args)
         else:
+            ast_cmd = "lfortran --show-ast --no-color {infile} -o {outfile}"
+            if asr_openmp:
+                ast_cmd = "lfortran --show-ast --no-color --openmp {infile} -o {outfile}"
             run_test(
                 filename,
                 "ast",
-                "lfortran --show-ast --no-color {infile} -o {outfile}",
+                ast_cmd,
                 filename,
                 update_reference,
                 verify_hash,
--- a/tests/reference/ast-pragma1-f8a3fde.json
+++ b/tests/reference/ast-pragma1-f8a3fde.json
@@ -1,12 +1,12 @@
 {
-    "basename": "ast-pragma1-6b56efa",
+    "basename": "ast-pragma1-f8a3fde",
     "cmd": "lfortran --show-ast --no-color --openmp {infile} -o {outfile}",
     "infile": "tests/pragma1.f90",
     "infile_hash": "0b9eea06186cb7eb1866b55256263d4e562835c40920f35b92ed18ee",
     "outfile": null,
     "outfile_hash": null,
-    "stdout": "ast-pragma1-6b56efa.stdout",
-    "stdout_hash": "cdb4205e19ae9c8ec854226781ea62e31657a920c19e0af61438c4ec",
+    "stdout": "ast-pragma1-f8a3fde.stdout",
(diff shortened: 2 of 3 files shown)


src/lfortran/parser/parser.cpp
```diff
@@ -237,6 +237,7 @@ Result<AST::TranslationUnit_t*> parse(Allocator &al, const std::string &s,
         diag::Diagnostics &diagnostics, const CompilerOptions &co)
 {
     Parser p(al, diagnostics, co.fixed_form, co.continue_compilation);
+    p.m_tokenizer.openmp_enabled = co.openmp;
```
certik: Pass all these options into the constructor.
```suggestion
Parser p(al, diagnostics, co.fixed_form, co.continue_compilation, co.openmp);
```

certik:
This is probably the way to fix it, otherwise we have to fiddle with the Bison grammar, which is more complex / fragile. So I think the above fix is good.

tests/reference/ast-pragma1-6b56efa.stdout
```diff
@@ -94,17 +94,16 @@
...
-        [(Pragma
-            0
-            OMPPragma
-            .false.
-            "parallel do"
-            []
-            ()
-        )
-        (DoLoop
+        [(DoLoop
```
certik: Investigate if the pragma1 test was created to test the Pragma AST node. If so, then you need to provide the `--openmp` option to this test, to keep testing that. Since it now tests that pragma just becomes a comment, which is a totally different test.

run_tests.py
```diff
@@ -155,10 +155,13 @@ def is_included(backend):
...
                 extra_args)
         else:
+            ast_cmd = "lfortran --show-ast --no-color {infile} -o {outfile}"
+            if asr_openmp:
+                ast_cmd = "lfortran --show-ast --no-color --openmp {infile} -o {outfile}"
             run_test(
                 filename,
                 "ast",
-                "lfortran --show-ast --no-color {infile} -o {outfile}",
+                ast_cmd,
```
certik, line 164: I am trying to figure out what this change is doing --- is it changing the meaning of `ast`, based on if `asr_openmp` is present? If so, that doesn't seem like a good design.

Let's create `ast_openmp` if we need that.

certik: You can do it in the next PR, so that we can get this one merged.

certik:
I think this is good enough to merge. If you could please fix https://github.com/lfortran/lfortran/pull/9444#discussion_r2679111872 in a subsequent PR, I would really appreciate it.

krystophny:
> I think this is good enough to merge. If you could please fix [#9444 (comment)](https://github.com/lfortran/lfortran/pull/9444#discussion_r2679111872) in a subsequent PR, I would really appreciate it.

Done in #9463

Issue #9088, opened 2025-12-15
Convert Descriptor derived type array to class type array conversion in subroutine args


jinangshah21:
MRE:
```fortran
program mre
    implicit none
    type :: package_t
        integer :: name
    end type package_t
    type(package_t), allocatable :: pkg(2)
    pkg(1)%name = 1
    pkg(2)%name = 2
    call display_packages(pkg)
contains 
    subroutine display_packages(packages)
        class(package_t), intent(in) :: packages(:)
        integer :: i
        do i = 1, size(packages)
            write(*,*) "Package Name: ", packages(i)%name
        end do
    end subroutine display_packages
end program mre
```

Outputs:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran a.f90 && ./a.out
 Package Name:            1
 Package Name:            2
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 
Package Name:     1
Package Name:     1438894480
```

HarshitaKalani:
The above MRE gives the following:
```console
% lfortran a.f90
At 7:5 of file a.f90
Runtime Error: Array 'pkg' is indexed but not allocated.
(lf) harshitakalani@Harshitas-MacBook-Air:~/repos/lfortran(main) % gfortran a.f90 && ./a.out
a.f90:6:37:

    6 |     type(package_t), allocatable :: pkg(2)
      |                                     1~~~~~
Error: Allocatable array 'pkg' at (1) must have a deferred shape or assumed rank
```
Here is the correct MRE:
```console
% gfortran a.f90 && ./a.out
 Package Name:            1
 Package Name:            2
% lfortran a.f90           
Package Name:     1
Package Name:     -1455326211
% cat a.f90
program mre
    implicit none
    type :: package_t
        integer :: name
    end type package_t
    type(package_t), allocatable :: pkg(:)
    allocate(pkg(2))
    pkg(1)%name = 1
    pkg(2)%name = 2
    call display_packages(pkg)
contains
    subroutine display_packages(packages)
        class(package_t), intent(in) :: packages(:)
        integer :: i
        do i = 1, size(packages)
            write(*,*) "Package Name: ", packages(i)%name
        end do
    end subroutine display_packages
end program mre                           
```

assem2002:
I'll handle it.

jinangshah21:
Here we need to handle class ArrayItems correctly, 
[...]

PR #9462, opened 2026-01-11
Fix : Allocate struct memory for each element in an array of type `class`

### LINKED ISSUE #9461: Bug : segfault on array of classes allocation
```.f90
program derived_types_92
    implicit none
    type :: package_t
        integer :: num
        character(:),allocatable :: str
    end type package_t

    class(package_t), allocatable :: instance(:)

    allocate(instance(1))
end program 
```

assem2002:
fixes #9461


certik:
I don't know if this failure is caused by this PR:
```
2026-01-11T04:49:57.9966620Z 	1552 - class_47 (Subprocess aborted)
```

certik:
I opened up an issue for the failure: https://github.com/lfortran/lfortran/issues/9478

assem2002:
@certik This PR is allocating a struct (the actual struct e.g. `{i32, f64}`) and inserts it into the class structure (`{Vtable*, struct*}`).
so yes each and every element in the array allocates a struct. I'm investigating this approach seeing what gfortran does.
This PR is a piece towards fixing this #9088

PR #9465, opened 2026-01-11
fix: handle scalar variables passing for functions inside do-loop conditions

### LINKED ISSUE #9437: Handle nested vars updated in function call from a condition of while loop
MRE:
```fortran
module temp

contains
  subroutine demo()
    implicit none
    integer :: a
    a = 0
    GET_ARGS: do while(f()) 
       if (a == 2) exit GET_ARGS
       print *, a
    end do GET_ARGS

  contains
     logical function f()
      f = .true.
      a = a + 1
    end function f
  end subroutine demo 
end module

program nested_func_mre
  use temp
  implicit none
  call demo()
end program nested_func_mre
```

Outputs:
```console
^C(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran d.f90 && ./a.out
           1
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran d.f90
0
0
0
0
0
....... Infinite loop
```

Related to #9327

YashNagda17:
Fixes #9437

Building on #8952, Modify the logic for this task to account for following cases for functions inside do-loop conditions
- Assign scalar variables from main to function temporary variables before function execution
- After the function executes, assign temporaries from function to main before starting loop execution
- At the end of the loop, again assign scalar from main to temporaries for next iteration

Similarly, for if block, perform the following:
- Assign scalar variables from main to function temporary variables
- After the function executes, assign temporaries from function to main before starting if block execution

### DIFF
--- a/integration_tests/nested_18.f90
+++ b/integration_tests/nested_18.f90
@@ -1,37 +1,71 @@
-! Checking Nested Variables Pass: Host variable from a subroutine 
-! used in nested function, Passed through Do-While loop check & If block checks
-module temp
+! Test for nested function global variables passes 
+! with function calls in conditions
+! Test for both, scalars and arrays
+! Test 1: Do-while loop with function call in condition (3 iterations)
+! Test 2: IF block with function call in condition
+
+module nested_18_temp
   contains
     subroutine demo()
       implicit none
-      integer :: a       ! Host variable
+      integer :: a
+      integer :: arr(5)
+      integer :: counter
+
+      ! Test 1: Do-While loop - runs 6 times
+      ! Check Upodates from Main to Function
+      ! And from function to Main
+      a = 0
+      arr(1:5) = 0
+      counter = 0
+      GET_ARGS: do while(increment())
+        ! Check updates from function to main
+        counter = counter + 1
+        if (counter == 1 .and. a /= 1 .and. arr(1) /= 1) error stop
+        if (counter == 3 .and. a /= 3 .and. arr(3) /= 3) error stop
+        if (counter == 5 .and. a /= 5 .and. arr(5) /= 3) error stop
+        print *, "Iteration", counter, ": a =", a, ": arr =", arr
+        
+        ! Update a inside main to check from main to function
+        a = a + 1
+        arr = arr + 1
+        counter = counter + 1
+        print *, "Iteration", counter, ": a =", a, ": arr =", arr
+        if (counter == 6) exit GET_ARGS
+      end do GET_ARGS
+      print *, "a =", a
+      print *, "arr =", arr
+      if (a /= 6) error stop
+      if (arr(1) /= 6) error stop
+
+      ! Test 2: IF block with function call in condition
       a = 10
-      do while(f())  ! While Loop access
-        a = 5
-      end do
-      if (f2()) then ! If Block executes
-        a = 10
+      if (set_value(25)) then
+        print *, "In IF body: a =", a
+        if (a /= 25) error stop
+        a = a + 5
       end if
-    contains
-      logical function f()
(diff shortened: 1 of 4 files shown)


YashNagda17:
For this MRE:
```
module temp
  contains
    subroutine demo()
      implicit none
      integer :: a, counter
      a = 0
      counter = 0
      GET_ARGS: do while(increment())
        ! Check updates from function to main
        counter = counter + 1
        a = a + 1
        if (counter == 4) exit GET_ARGS
      end do GET_ARGS
      print *, "a =", a
      
    contains
      logical function increment()
        a = a + 1
        increment = .true.
      end function increment
    end subroutine demo
end module
```

Two things can be verified:
- Assignments for Scalars are created as intended, before loop, at loop start and loop end
- There is no temporary copy for `counter`. So, we are only creating copies for variables used inside function, and not all scalars.

ASR Created for Do-Loop
```
(Assignment
        (Var 3 __lcompilers_created__nested_context__demo_a)
        (Var 3 a)
        ()
        .false.
        .false.
    )
    (WhileLoop
        GET_ARGS
        (FunctionCall
            3 increment
            ()
            []
            (Logical 4)
            ()
            ()
        )
        [(Assignment
            (Var 3 a)
            (Var 3 __lcompilers_created__nested_context__demo_a)
            ()
            .false.
            .false.
        )
        (Assignment
            (Var 3 counter)
            (IntegerBinOp
                (Var 3 counter)
                Add
                (Integer 4)
                ()
            )
            ()
            .false.
            .false.
        )
        (Assignment
            (Var 3 a)
            (IntegerBinOp
                (Var 3 a)
                Add
                (Integer 4)
                ()
            )
            ()
            .false.
            .false.
        )
        (If
            ()
            (IntegerCompare
                (Var 3 counter)
                Eq
                (IntegerConstant 4 (Integer 4) Decimal)
                (Logical 4)
                ()
            )
            [(Exit
                GET_ARGS
            )]
            []
        )
        (Assignment
            (Var 3 __lcompilers_created__nested_context__demo_a)
            (Var 3 a)
            ()
            .false.
            .false.
        )]
        []
    )
```

jinangshah21:
At which places are we doing end loop assignments (I.e. `__lcompilers_created__nested_context__demo_a =a`? (For eg. do we handle for CYCLE)..

YashNagda17:
`CYCLE` is handled by #9365 
I don't feel we need these assignments at `EXIT`, since there won't be any function calls of loop test after exit executions. 
This PR shall add these assignments at loop end... I am unable to think of any other cases, which might require this handling, for now... 

#8952 only handled reads from main to function, not other way around... So I have modified that test case here to check for both ways & avoid duplication of tests unnecessarily...

Function calls in loop body are working fine, I think if any edge cases arise, we can fix them then

integration_tests/nested_18.f90
```diff
@@ -1,37 +1,71 @@
...
       a = 10
-      do while(f())  ! While Loop access
-        a = 5
-      end do
-      if (f2()) then ! If Block executes
-        a = 10
+      if (set_value(25)) then
+        print *, "In IF body: a =", a
+        if (a /= 25) error stop
+        a = a + 5
```
certik, line 46: Should we add a new test, instead of modifying the old one?

certik:
@jinangshah21 do you think this is good to merge, or should we improve it? The logic is a bit complex, but in our current loop design it might be necessary. There might be a simpler loop design, but that would be a separate PR anyway.

jinangshah21:
I think we can merge this for now. We can do improvement in upcoming PRs

certik:
Perfect, let's do it.

PR #9467, opened 2026-01-11
fix: serialize string constant also while serializing a module

### LINKED ISSUE #9454: Handle `achar(0)` in strings created in different files than main program
MRE:
Module File:
```fortran
module mre_read_string_fmt_mod
contains
    subroutine temp()
        str = "'control char "//achar(0)//"','normal literal'"
        print *, str
    end subroutine temp
end module mre_read_string_fmt_mod
```

Program File:
```fortran
program test
    use mre_read_string_fmt_mod
    implicit none
    call temp()
end program
```

Output:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90 d.f90
'control char �
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran c.f90 d.f90 && ./a.out
 'control char ','normal literal'
```

This works if we shift module to main file. So issue may be in serialising/deserialising.

HarshilShah1804:
Fixes #9454


jinangshah21 (approved):
I think this is fine!

Issue #9328, opened 2026-01-02
Handle `trim(generic)` in select type stmt


jinangshah21:
MRE:
```fortran
program select_type_char_mre
    implicit none

    call print_any("  Hello World  ")

contains

    subroutine print_any(generic)
        class(*), intent(in) :: generic
        select type (generic)
        type is (character(len=*))
            print *, trim(generic)
        end select
    end subroutine print_any

end program select_type_char_mre
```

LFortran Error:
```console
ASR verify pass error: Unexpected args, StringTrim expects (char) as arguments
  --> d.f90:12:22
   |
12 |             print *, trim(generic)
   |                      ^^^^^^^^^^^^^ failed here
```

Adityazzzzz:
@jinangshah21 
I have been working on fixing this issue. I found that visit_SelectType was missing the logic to handle TypeStmtType correctly (it was defaulting to Class(*) instead of creating a shadow variable for Character).

I implemented the fix in ast_body_visitor.cpp to create a shadow variable inside the select type block. However, I am hitting a backend assertion in asr_to_llvm.cpp:

```
LCOMPILERS_ASSERT failed: src/libasr/codegen/asr_to_llvm.cpp
function setup_string_length(), line number 773
ERROR MESSAGE: Shouldn't define assumed length string variable (They're only arguments)
The Problem: The shadow variable we create for type is (character(len=*)) inherits the len=* property.

```
Since it is inside a block, it is treated as a Local Variable.

The LLVM backend refuses to allocate len=* variables on the stack unless they are Dummy Arguments (intent(in)).

What I tried:

Pointer Wrapper: I wrapped the shadow variable in ASR::make_Pointer_t. I expected this to work (since pointers have fixed size), but the backend still checks the underlying type's length and crashes.

Intent(In): I tried marking the shadow variable as Intent::In to mimic a dummy argument, but this causes other backend assertions because it's not actually in the function signature.

Question: How should we represent the shadow variable for type is (character(len=*)) in ASR?

Should asr_to_llvm.cpp be relaxed to allow len=* if the variable is a Pointer?

Or is there a specific way to set up the Associate node so the backend knows the storage comes from the selector?

Here is the diff of my current attempt in ast_body_visitor.cpp:
```

// Inside visit_SelectType case TypeStmtType:
// ...
selector_type = determine_type(...);

// I tried wrapping it in a pointer to avoid stack allocation of unknown length
[...]

PR #9471, opened 2026-01-11
fix: handle polymorphic variables inside intrinsic `trim()` intrinsic

YashNagda17:
Towards #9328

Changes made:
- Add support for polymorphic variables inside `trim()`
- Add guardrails for using this behavior only inside select statements for characters, throw semantic error for using directly

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -101,10 +101,10 @@ contains
         deallocate(s)
     end subroutine modify_and_deallocate
 
-
-
-
-
+    subroutine intrinsic_polymorphic(generic)
+        class(*), intent(in) :: generic
+        print *, trim(generic)
+    end subroutine intrinsic_polymorphic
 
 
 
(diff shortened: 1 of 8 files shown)


certik (approved):
I think that this is good. @HarshilShah1804 can you please also review it?

HarshilShah1804 (approved):
Yes, the fix seems good to me.

HarshilShah1804:
@YashNagda17 Please rebase the PR and resolve the conflicts in the `continue_compilation_1.f90`. Then we can merge it.

tests/errors/continue_compilation_1.f90
```diff
@@ -475,6 +475,8 @@ program continue_compilation_1
     OPEN(unit=10, recl=10, recl=20)
     OPEN(unit=10, recl=10.5)
 
+    call intrinsic_polymorphic("  Hello World  ")
```
HarshilShah1804: Just a small suggestion. You can add this statement at last so that the `stderr` would not have changes for the error tests in the following lines.

You can change this while you rebase the PR

PR #9474, opened 2026-01-11
feat: add --mangle-underscore-external

### LINKED ISSUE #9446: itpplasma/spline: OpenMP sentinels + LAPACK (dptsv) external resolution/mangling
## Summary
`itpplasma/spline` does not currently build cleanly with LFortran `main` due to a sequence of blockers:
1) `!$omp threadprivate(...)` is treated as an OpenMP pragma even when OpenMP is not enabled.
2) After the OpenMP-sentinel parsing problem is addressed, `dptsv` (LAPACK) is rejected unless implicit interfaces are enabled.
3) Even with `--implicit-interface` and `-llapack -lblas`, linking fails because LFortran calls `dptsv` (no underscore) while system LAPACK typically exports `dptsv_`.

We are tracking this as part of compiling ITP Plasma codes (`spline` and `vode` first).

## Environment
- LFortran: `ef70dd43011b30b006d758c1c114b19c47d24fb3` (`upstream/main` as of 2026-01-10)
- Reproducer: `itpplasma/spline` @ `1c40553a2cc6c930804cc8ea9cc9393106345a6e`

## Reproducer
```bash
git clone https://github.com/itpplasma/spline.git
cd spline

# 1) OpenMP sentinels should be ignored when OpenMP is off
lfortran spline.f90

# 2) After OpenMP sentinel handling is fixed/ignored, compilation proceeds to LAPACK call
#    but LFortran requires either explicit interface or --implicit-interface:
#
# lfortran spline.f90 spline_test.f90
#   semantic error: Function 'dptsv' not found (not user defined nor intrinsic)

# 3) With implicit interfaces and system LAPACK/BLAS:
lfortran spline.f90 spline_test.f90 --implicit-interface -llapack -lblas -o spline_test_lfortran
#   /usr/bin/ld: ... undefined reference to `dptsv'
```

### (1) OpenMP sentinels
```
syntax error: Token '!$omp threadprivate(j_start)' (of type 'pragma') is unexpected here
 --> spline.f90:9:3
```

### (3) Link error against system LAPACK
```
/usr/bin/ld: ... undefined reference to `dptsv'
```

## Evidence for underscore mismatch
- System LAPACK exports `dptsv_`, not `dptsv`.
- LFortran currently references `dptsv` (no underscore) from generated objects.

(Concrete nm output:)

```
# System LAPACK provides dptsv_ (trailing underscore)
0000000000272780 T dptsv_

# LFortran-generated object references
                 U dptsv
```

## Related PRs
- https://github.com/lfortran/lfortran/pull/9444 (treat OpenMP sentinels as comments unless `--openmp`)

## Related
- OpenMP sentinel directives when OpenMP is off: #6449
- LAPACK tracking: #5482
- External/implicit-interface policy for procedure calls: #1170
- External subroutine not found (similar symptom): #6623
- Underscore/mangling discussion and flags (`--mangle-underscore`, `--separate-compilation`): #9067
- External BLAS/LAPACK + implicit interface corner cases (examples): #4251, #2924

## Notes / desired behavior
- Current workaround for users: provide an explicit interface/wrapper for `dptsv` (e.g. via `iso_c_binding` and `bind(C, name="dptsv_")`) or use a LAPACK built with the expected symbol names.
- `--mangle-underscore` currently does not affect the emitted external symbol for this LAPACK call (object still references `dptsv`).
- Without `--openmp`, OpenMP sentinels like `!$omp ...` should be treated as comments.
- For LAPACK/BLAS calls, either:
  - `--mangle-underscore` should also apply to external procedure calls so `dptsv` resolves as `dptsv_`, or
  - provide a documented/robust alternative path (e.g., recommended `iso_c_binding` wrapper pattern) that works for common system BLAS/LAPACK.

### LINKED ISSUE #9473: Bug: --all-mangling does not apply to external interface functions
Bug: --all-mangling does not apply to external interface functions

## Summary
`--all-mangling` and `--mangle-underscore` flags do not affect external function calls (e.g., LAPACK/BLAS routines with implicit interfaces). This prevents linking with system libraries that use underscore-mangled symbols (the Fortran standard convention).

## Reproducer
```fortran
! test_external_mangle.f90
program test_external
    implicit none
    integer :: n, info
    real(8) :: d(3), dl(2), b(3)

    n = 3
    d = [1.0d0, 2.0d0, 3.0d0]
    dl = [0.5d0, 0.5d0]
    b = [1.0d0, 2.0d0, 3.0d0]

    ! dptsv is a LAPACK routine (external, implicit interface)
    call dptsv(n, 1, d, dl, b, n, info)

    print *, 'info =', info
end program
```

Build with mangling flags:
```bash
lfortran test_external_mangle.f90 --implicit-interface --all-mangling --mangle-underscore -llapack
```

Expected: Links successfully (calls `dptsv_` to match system LAPACK)
Actual: Link error - `undefined reference to 'dptsv'`

Verification:
```bash
# System LAPACK exports dptsv_ (with underscore)
nm -D /usr/lib/liblapack.so | grep dptsv
# Output: 0000000000272780 T dptsv_

# LFortran emits call to dptsv (without underscore)
lfortran --show-llvm test_external_mangle.f90 --implicit-interface --all-mangling --mangle-underscore | grep dptsv
# Output: declare void @dptsv(...)
```

## Root Cause
In `src/libasr/codegen/asr_to_llvm.cpp:6009-6011`, the `instantiate_function()` method bypasses mangling for external interface functions:

```cpp
if (BindC) {
    fn_name = sym_name;
} else if (Interface && !Intrinsic && !module) {
    fn_name = sym_name;  // BUG: Ignores --all-mangling and --mangle-underscore
} else {
    fn_name = mangle_prefix + sym_name;
}
```

External functions with implicit interfaces (like LAPACK `dptsv`) take the second branch and unconditionally use the raw symbol name, ignoring all mangling flags.

## Expected Behavior
When `--all-mangling` is enabled, external interface functions should be mangled according to the active mangling scheme (e.g., with `--mangle-underscore`, `dptsv` should become `dptsv_`).

## Proposed Fix
Check mangling flags before taking the Interface bypass branch. When mangling is active, allow external interfaces to proceed to the third branch for proper mangling.

## Impact
Blocks usage of system LAPACK/BLAS and other Fortran libraries that follow standard underscore mangling convention. See #9446 for real-world use case (itpplasma/spline library).

## Related
- #9446 - itpplasma/spline compilation (LAPACK mangling blocker)
- #9067 - Mangling discussion
- #5482 - LAPACK support tracking

krystophny:
## Summary
Add --mangle-underscore-external flag to append trailing underscore to external interface functions for LAPACK/BLAS compatibility.

Fixes #9473. Fixes #9446.

## Why
Fortran libraries like LAPACK/BLAS use underscore-mangled symbols. LFortran needs to generate matching names when calling through implicit interfaces.

## Changes
- [`src/libasr/codegen/asr_to_llvm.cpp#L71-L156`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/src/libasr/codegen/asr_to_llvm.cpp#L71-L156): Extract mangling to helper functions (eliminates duplication, single source of truth)
- [`src/libasr/codegen/asr_to_llvm.cpp#L6089-L6093`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/src/libasr/codegen/asr_to_llvm.cpp#L6089-L6093): Apply underscore mangling when flag set
- [`src/bin/lfortran_command_line_parser.cpp#L171`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/src/bin/lfortran_command_line_parser.cpp#L171): Add --mangle-underscore-external flag
- [`src/libasr/utils.h#L57`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/src/libasr/utils.h#L57): Add mangle_underscore_external field
- `tests/mangle_underscore_external_01.f90` and `integration_tests/external_13.f90`: Test coverage

## Tests
- [`tests/mangle_underscore_external_01.f90`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/tests/mangle_underscore_external_01.f90): Verifies LLVM output shows foo_() declaration
- [`integration_tests/external_13.f90`](https://github.com/lfortran/lfortran/blob/ad2fb50d0920bc60e95bf781128dda352f70af9c/integration_tests/external_13.f90): End-to-end test with --implicit-interface --mangle-underscore-external

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -85,22 +85,11 @@ static inline bool is_external_interface_function(ASR::FunctionType_t* ftype) {
 }
 
 /**
- * Check if a function name has a reserved LFortran/LPython/LCompilers prefix.
- * These functions should not be mangled by user-facing flags.
- */
-static inline bool has_reserved_prefix(const std::string& name) {
-    return name.rfind("_lfortran", 0) == 0 ||
-           name.rfind("_lpython", 0) == 0 ||
-           name.rfind("_lcompilers", 0) == 0;
-}
-
-/**
  * Compute the final mangled LLVM function name.
  *
  * This function encapsulates all mangling logic in one place:
  * 1. Determines base name (BindC, external interface, or prefixed)
  * 2. Adds parent function prefix if applicable
- * 3. Applies underscore mangling for external interface functions if requested
  *
  * @param sym_name Original symbol name from ASR
  * @param ftype Function type information
@@ -116,13 +105,13 @@ static std::string compute_llvm_function_name(
     const std::string& mangle_prefix,
     const ASR::Function_t* parent_function
 ) {
-    // Check function properties once
+    (void)compiler_options;
     bool is_external_interface = is_external_interface_function(ftype);
     bool is_bindc = (ftype->m_abi == ASR::abiType::BindC);
 
     std::string fn_name;
 
-    // Step 1: Determine base name based on function type
+    // Determine base name based on function type
     if (is_bindc) {
         fn_name = ftype->m_bindc_name ? ftype->m_bindc_name : sym_name;
     } else if (is_external_interface) {
@@ -131,21 +120,12 @@ static std::string compute_llvm_function_name(
         fn_name = mangle_prefix + sym_name;
     }
 
-    // Step 2: Add parent function prefix for nested functions
+    // Add parent function prefix for nested functions
     if (parent_function != nullptr &&
         !is_external_interface &&
         ftype->m_abi != ASR::abiType::Intrinsic &&
         !is_bindc) {
         fn_name = std::string(parent_function->m_name) + "." + fn_name;
-    }
-
-    // Step 3: Apply underscore mangling for external interface functions
-    if (compiler_options.po.mangle_underscore_external && is_external_interface) {
-        if (!has_reserved_prefix(fn_name) &&
-            !fn_name.empty() &&
-            fn_name.back() != '_') {
-            fn_name += "_";
-        }
     }
 
     return fn_name;
--- a/src/libasr/pass/unique_symbols.cpp
+++ b/src/libasr/pass/unique_symbols.cpp
@@ -48,14 +48,15 @@ class SymbolRenameVisitor: public ASR::BaseWalkVisitor<SymbolRenameVisitor> {
     bool bindc_mangling = false;
     bool fortran_mangling;
     bool c_mangling;
+    bool mangle_underscore_external;
     bool should_mangle = false;
     std::vector<std::string> parent_function_name;
     std::string module_name = "";
     SymbolTable* current_scope = nullptr;
 
-    SymbolRenameVisitor(bool imm, bool mm, bool gm, bool im, bool am, bool bcm, bool fm, bool cm) :
+    SymbolRenameVisitor(bool imm, bool mm, bool gm, bool im, bool am, bool bcm, bool fm, bool cm, bool mue) :
     intrinsic_module_name_mangling(imm), module_name_mangling(mm), global_symbols_mangling(gm), intrinsic_symbols_mangling(im),
-    all_symbols_mangling(am), bindc_mangling(bcm), fortran_mangling(fm), c_mangling(cm) {}
+    all_symbols_mangling(am), bindc_mangling(bcm), fortran_mangling(fm), c_mangling(cm), mangle_underscore_external(mue) {}
 
 
     const std::unordered_set<std::string> reserved_keywords_c = {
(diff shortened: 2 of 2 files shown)


certik:
The CI now fails with:
```
E   ImportError: /home/runner/micromamba/envs/scipy-dev/lib/cdflib.so: undefined symbol: _lfortran_dlog_
```
Possibly the `--all-mangling` should be used there.

certik:
It still fails, something with the mangling is now wrong:
```
    from . import _ufuncs
E   ImportError: /home/runner/micromamba/envs/scipy-dev/lib/cdflib.so: undefined symbol: _lfortran_dlog_
```

krystophny:
Should be fixed now.

krystophny:
Should be fixed now.

integration_tests/external_13.f90
HarshitaKalani, line 1: We'll need to register this test in CMakeLists.txt.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -68,6 +68,91 @@ using ASRUtils::determine_module_dependencies;
...
+    }
+
+    // Step 3: Apply underscore mangling for external interface functions
+    if (compiler_options.po.mangle_underscore_external && is_external_interface) {
+        if (!has_reserved_prefix(fn_name) &&
+            !fn_name.empty() &&
+            fn_name.back() != '_') {
+            fn_name += "_";
+        }
+    }
```
certik: This logic should not be here, but rather in `src/libasr/pass/unique_symbols.cpp`. That way we have all mangling at one place, and that way ASR itself is then explicit about this and we can check it easily. In LLVM we do not do any special replacement, we just use whatever name is in ASR.

certik:
I think that this is good, but the actual mangling logic should go into ASR, per https://github.com/lfortran/lfortran/pull/9474/changes#r2686962349.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: done

Issue #9447, opened 2026-01-10
itpplasma/vode: OpenMP sentinels + implicit-interface ICE + generic resolution


krystophny:
## Summary
`itpplasma/vode` does not currently compile/run with LFortran `main`. There are multiple sequential blockers:
1) OpenMP sentinels (`!$omp threadprivate(...)`) are treated as pragmas even when OpenMP is not enabled.
2) With OpenMP sentinels ignored, the code requires `--implicit-interface` due to legacy `EXTERNAL` usage.
3) With `--implicit-interface`, LFortran hits an ICE on `PRESENT(J_FCN)` where `J_FCN` is an optional procedure dummy.
4) After addressing the `present()` ICE, compilation reaches a generic-resolution error at the example call:
   `semantic error: No matching function found for the call to generic procedure, dvode_f90`.

We are tracking this as part of compiling ITP Plasma codes (`spline` and `vode` first).

## Environment
- LFortran: `ef70dd43011b30b006d758c1c114b19c47d24fb3` (`upstream/main` as of 2026-01-10)
- Reproducer: `itpplasma/vode` @ `5797785f0829334e758bcd5fe650b26abbe3db90`

## Reproducer
```bash
git clone https://github.com/itpplasma/vode.git
cd vode

# 1) OpenMP sentinels should be ignored when OpenMP is off
lfortran dvode_f90_m.f90

# 2) After OpenMP sentinel handling is fixed/ignored, the example build wants implicit interfaces
lfortran dvode_f90_m.f90 example1.f90
#   semantic error: function interface must be specified explicitly; you can enable implicit interfaces with `--implicit-interface`

# 3) With implicit interfaces enabled
lfortran dvode_f90_m.f90 example1.f90 --implicit-interface
#   ICE / assert (see below)
```

### (1) OpenMP sentinels
```
syntax error: Token '!$omp threadprivate(USE_JACSP, LIKE_ORIGINAL_VODE)' (of type 'pragma') is unexpected here
 --> dvode_f90_m.f90:2065:1
```

### (3) ICE with `--implicit-interface`
This appears to be triggered by `PRESENT(J_FCN)` where `J_FCN` is an optional procedure dummy:
- `dvode_f90_m.f90:2275`: `IF (PRESENT(J_FCN)) THEN`

Observed failure:
```
LCOMPILERS_ASSERT failed: .../src/libasr/asr.h
function down_cast(), line number 41 at is_a<T>(*f)
```

### (4) Generic resolution
[...]

PR #9475, opened 2026-01-11
Fix generic resolution for implicit interface procedures

krystophny:
## Summary
- Fix generic procedure resolution when passing EXTERNAL procedures to generics expecting explicit interfaces
- Treat implicit interfaces (n_arg_types=0) as compatible with any interface

Contributes to #9447 (VODE compilation)

## Why
When calling a generic with an EXTERNAL procedure argument, `types_equal()` compared the implicit interface (empty arg types) against the explicit interface (with typed args) and rejected the match. The actual interface compatibility should be checked at the call site, not during generic resolution.

**Stage:** ASR utilities (generic procedure matching)

## Changes
- [`src/libasr/asr_utils.cpp#L2527-L2544`](https://github.com/lfortran/lfortran/blob/39806e9511133c0ea8509cae0887864e86057478/src/libasr/asr_utils.cpp#L2527-L2544): Check for implicit interface in `argument_types_match()` before calling `types_equal()`

## Tests
- [`integration_tests/implicit_interface_22.f90`](https://github.com/lfortran/lfortran/blob/39806e9511133c0ea8509cae0887864e86057478/integration_tests/implicit_interface_22.f90): Generic call with EXTERNAL procedure argument

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7273,6 +7273,104 @@ public:
 
     void visit_Assignment(const ASR::Assignment_t &x) {
         if (compiler_options.emit_debug_info) debug_emit_loc(x);
+
+        // Special-case: transfer(character, int8_array, size) lowered as BitCast.
+        // When scalarized into element-wise assignments, extract the corresponding
+        // byte from the source string.
+        if (ASR::is_a<ASR::BitCast_t>(*x.m_value)) {
+            ASR::BitCast_t* bc = ASR::down_cast<ASR::BitCast_t>(x.m_value);
+            if (ASR::is_a<ASR::ArrayItem_t>(*x.m_target) &&
+                ASRUtils::is_integer(*ASRUtils::expr_type(x.m_target)) &&
+                ASR::down_cast<ASR::Integer_t>(ASRUtils::expr_type(x.m_target))->m_kind == 1 &&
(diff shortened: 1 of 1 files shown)


certik:
I think this is fine, but what is the commit 58b71d0efd11448bc00a527147dadedc479c3e4e having to do with this PR? 

Is it needed to pass tests? Otherwise it should go to another PR, together with tests.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6970,27 +6970,38 @@ void get_local_info_dwarfdump(struct Stacktrace *d) {
...
+    fseek(fp, 0, SEEK_SET);
+    if (file_size <= 0) {
+        fclose(fp);
+        free(filename);
         return;
     }
-    char *file_contents = _lpython_read(fd, size);
-    _lpython_close(fd);
+
+    char *file_contents = malloc((size_t)file_size + 1);
```
certik, line 6989: What bug are these changes fixing? I don't think these are tested by the test above.

certik:
I think this PR might be mixing 3 unrelated bug fixes. I would separate them into 3 separate PRs, and let's ensure we have tests for each, so that it doesn't get broken.

krystophny:
Split into separate PRs per review feedback:

- #9498: Generic resolution for implicit interface procedures (with test)
- #9499: Transfer string to int8 array assignment (with test)

The stacktrace C runtime change was dropped as it lacked a clear purpose and test coverage.

Closing this PR in favor of the focused splits above.

Issue #8926, opened 2025-11-28
findloc fails bounds check, and bounds check doesn't trigger for auto-realloc-lhs


certik:
The code is from https://fortran-lang.discourse.group/t/euler-conjecture-and-cdc-6600/10501/67:
```fortran
! 27^5 + 84^5 + 110^5 + 133^5 = 144^5
program euler
use, intrinsic :: iso_fortran_env, only: int64
implicit none
integer(int64) :: i,j, k, l, m,n5(145)
character(*), parameter :: cfmt = '(*(g0,1x))'
outer: do i = 1, 145
  n5(i) = i**5
  do j = 1, i; do k = 1, j; do l = 1, k
     m=findloc(n5(j)+n5(k)+n5(l)+n5(1:l),value=n5(i),dim=1)
     if ( m > 0 ) exit outer
     end do; end do; end do
end do outer
print cfmt, "i^5 = ", n5(i)
print cfmt, j, k, l, m, i
end program euler
```
This gives:
```console
$ gfortran -fcheck=all a.f90 && ./a.out
i^5 =  61917364224
133 110 84 27 144
$ lfortran a.f90
At 10:16 of file a.f90
Runtime Error: Array shape mismatch in assignment to '__libasr_created__intrinsic_array_function_FindLoc'

Tried to match size 145 of dimension 1 of LHS with size 1 of dimension 1 of RHS.

Use '--realloc-lhs-arrays' option to reallocate LHS automatically.
$ lfortran a.f90 --realloc-lhs-arrays
$ echo $?
11
```
It looks like there are several bugs here that we need to create MREs for and fix:

* the main one is likely incorrect findloc instantiation, which creates incorrect arrays that require allocatable LHS
* The segfault with `--realloc-lhs-arrays` means our bounds checking is not strong enough --- create an MRE and ensure our bounds checker catches it

HarshitaKalani:
Here is an MRE for the issue with findloc:
```console
% lfortran integration_tests/findloc_array_expr_01.f90
At 13:13 of file integration_tests/findloc_array_expr_01.f90
Runtime Error: Array shape mismatch in assignment to '__libasr_created__intrinsic_array_function_FindLoc1'

Tried to match size 10 of dimension 1 of LHS with size 5 of dimension 1 of RHS.

Use '--realloc-lhs-arrays' option to reallocate LHS automatically.
% gfortran integration_tests/findloc_array_expr_01.f90 && ./a.out
[...]

PR #9479, opened 2026-01-12
fix: remove allocas from array deallocation and allocation

xaerru:
Fix https://github.com/lfortran/lfortran/issues/8926

Reduced MRE from the issue:
```fortran
program allocate_34
    implicit none
    integer :: i
    allocate(arr(1))
    do i = 1, 1000000
        deallocate(arr)
        allocate(arr(1))
    end do
end program
```

In llvm generation of `deallocate` and `allocate` we had `alloca` statements. This caused a large amount of memory allocation on the stack. This led to a segfault in `malloc.c`.

### DIFF
--- a/integration_tests/allocate_34.f90
+++ b/integration_tests/allocate_34.f90
@@ -1,0 +1,10 @@
+program allocate_34
+    implicit none
+    integer, allocatable :: arr(:)
+    integer :: i
+    allocate(arr(1))
+    do i = 1, 1000000
+        deallocate(arr)
+        allocate(arr(1))
+    end do
+end program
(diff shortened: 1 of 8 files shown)


integration_tests/allocate_34.f90
```diff
@@ -0,0 +1,10 @@
+program allocate_34
+    implicit none
+    integer, allocatable :: arr(:)
+    integer :: i
+    allocate(arr(1))
+    do i = 1, 1000000
+        deallocate(arr)
+        allocate(arr(1))
+    end do
```
certik, line 9: @xaerru do you know if this test would crash before? It's a little bit slow to run:
```
1286/2154 Test #1272: allocate_34 ........................................   Passed    0.03 sec
```
Ideally all tests should run in 0.00s (i.e., instantly), since we have lots of tests and we need them to run quickly.

This kind of issue is better checked with reference tests, which we do in this PR also.

However, for now it's fine. We can easily reduce the iteration count later.

xaerru: Yes, it segfaults before the changes made in this PR.

I'll make another PR with a reference test instead of an integration test.

PR #9484, opened 2026-01-12
fix: handle C comments with ifdef C preprocessing directive

### LINKED ISSUE #9483: Handle C comments (`/**/`) with `ifdef` C preprocessing directive
MRE:
```fortran
program test_stringify
    implicit none
#ifdef __GFORTRAN__      /* ABC-comment */
#define X 4
#else                      /* XYZ-comment */
#define X 5
#endif
    print *, X
end program test_stringify
```

LFortran Error:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90 --cpp
C preprocessor error: #else encountered outside of #ifdef or #ifndef
 --> c.f90:6:1
  |
6 | #define X 5
  | ^ 
```

HarshilShah1804:
Fixes #9483

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2652,6 +2652,9 @@ RUN(NAME cpp_pre_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c wasm
 RUN(NAME cpp_pre_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c wasm
     EXTRA_ARGS --cpp
     GFORTRAN_ARGS -cpp)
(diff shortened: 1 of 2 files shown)


opixdown:
@HarshilShah1804 I was working on the same issue and noticed that `#ifndef` has the same problem as `#ifdef`. 
```cpp
"#" whitespace? "ifndef" whitespace @t1 name @t2 whitespace? newline {
```
i think we should fix this too in this pr

certik (approved):
Thanks!

HarshilShah1804:
The MacOS LLVM 21 runner was stuck on `Test(Linux/MacOS)` test. I have re-run the test.

certik:
If the test passes, it will auto merge, so we are good. I think we don't need to run the full tests, since this only enables more code to compile, so existing code shouldn't break.

PR #9503, opened 2026-01-13
llvm: eliminate allocas in complex helper functions

krystophny:
## Summary
- Refactor complex helper functions to avoid stack allocations
- Prevents stack growth when complex operations are called in loops

## Changes
- `convert_complex_vector_to_struct`: use extract/insert element instead of alloca+bitcast
- `complex_re`, `complex_im`: use extractvalue instead of alloca+GEP+load
- `complex_from_floats`: use insertvalue instead of alloca+GEP+store+load

## Test
- `complex_32.f90`: 500k iterations of complex operations (would overflow stack with old code)


integration_tests/complex_32.f90
```diff
@@ -0,0 +1,17 @@
...
+    use, intrinsic :: iso_fortran_env, only: dp => real64
+    implicit none
+
+    integer :: i
+    real(dp) :: sum
+    complex(dp) :: z
+
+    sum = 0.0_dp
+    z = cmplx(1.0_dp, 2.0_dp, kind=dp)
+    do i = 1, 500000
```
certik, line 11: Have to be careful here, might need to lower the number to run fast.

PR #9504, opened 2026-01-13
llvm: only hoist constant-size allocas to entry block

krystophny:
## Summary
- Modify CreateAlloca to only hoist constant-size allocas to entry block
- Variable-size allocas stay at current insertion point to avoid stack growth in loops

## Changes
- [`llvm_utils.cpp#L1589-L1608`](https://github.com/lfortran/lfortran/blob/a879dcd6abc79a67bb3792efe6b2318c751cee16/src/libasr/codegen/llvm_utils.cpp#L1589-L1608): Add check for constant size before hoisting to entry block

## Tests
- [`array_section_11.f90`](https://github.com/lfortran/lfortran/blob/a879dcd6abc79a67bb3792efe6b2318c751cee16/integration_tests/array_section_11.f90): 100k iterations of array section operations


krystophny:
Closing - unable to reproduce stack overflow issue. Recent codegen changes may have eliminated the problematic alloca patterns.

PR #9508, opened 2026-01-13
fix: implement array reading from internal files for i32, i64, f32, f64 types

### LINKED ISSUE #9442: Can't read into an array
This program has been valid ever since F95:
```
program readarray
  implicit none
  character(12):: clist = '1.0 42. -42.'
  real list(3)
  read(clist,*) list
  print "(3F6.1)",list
end program readarray
```
Lfortran output at compile time:
```
LFORTRAN ERROR: Reading into an array of float is not supported.
```
```
   1.0  42.0 -42.0
```

opixdown:
fixes #9442
additionally added support for i32,i64 types


certik:
It should work with all tests, but just in case let's run it.

assem2002:
llvm Wasm experienced some **stdio** buffer issues and raised an error, it was related to  `error stop + String`. We should open up an issue for that.

certik:
Go ahead and open an issue for this please.

PR #9509, opened 2026-01-14
Enh : Use special way of allocation for arrays of classes (single class with 1 VTable + Consecutive structs).

### LINKED ISSUE #9088: Convert Descriptor derived type array to class type array conversion in subroutine args

assem2002:
Please read the added document before review. 
It's probably not the best approach from the ASR perspective (and maybe llvm also) , but we'll figure out later how to represent those special array types properly in ASR. 

***
It tries to do what gfortran does; Instead of allocating consecutive classes and inserting VTable+struct, We just allocate 1 single class and insert 1 single VTable along with allocated-consecutive structs.

* Use single Class structure to hold single VTable, and the consecutive structs allocated. It saves memory + makes casts between CStructs and Fortran classes seamless.
* Create a class View when accessing any element of the array (insert desired struct + VTable)
* Forbid classes to appear in a fixedSizeArray type.
* Add struct api helpers to store struct into class, Create StructView, Allocate an array of classes properly. *
***
Closes #9088
relate to this comment https://github.com/lfortran/lfortran/pull/9462#pullrequestreview-3648125579


certik:
It would be good to extract an MRE for this bug at the CI.

assem2002:
Identified MRE for failing CI
```.f90
program mre
    type:: package_t
        integer :: name
    end type package_t

    class(package_t), allocatable :: pkg(:)
    
    call display_packages(pkg)
contains 
    subroutine display_packages(arg)
      class(package_t), intent(in):: arg(:)
    end subroutine display_packages
end program mre
```

jinangshah21 (approved):
Look Good to me!

certik (approved):
I think that this is fine, thanks for fixing it!

Is the MRE that you extracted (https://github.com/lfortran/lfortran/pull/9509#issuecomment-3758166285) part of the integration tests? If not, let's put it there, so that the integration tests fail first before a 3rd-party.

assem2002:
added it already, so we're good.

Issue #8307, opened 2025-08-09
Better error message of out of bounds errors


certik:
Here is the current message:
```
$ lfortran a.f90
Runtime error: Array 'arr' index out of bounds.

Tried to access index 87 of dimension 1, but valid range is 1 to 8.
```
Here is how it should look like instead (with colors):
```
$ lfortran a.f90
runtime error: Array 'arr' index out of bounds
  --> a.f90:10:13
   |
10 |     x = arr(87)
   |             ^^ note: valid range is 1:8, got 87
```
Or something like that.

For multidimensional arrays it should highlight exactly the right index. We also need to figure out a good message if the source code is not available.

For other kinds of out of bounds errors (array assignment, function call, etc.) we need to figure out similar appropriately tailored error messages.

certik:
Related: https://github.com/lfortran/lfortran/issues/9526.

PR #9514, opened 2026-01-14
Implement nicer runtime error formatting

certik:
Towards #8307.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1911,7 +1911,7 @@ public:
                         var_name = ASRUtils::symbol_name(sym);
                     }
                     llvm_utils->generate_runtime_error(is_allocated,
-                        "Attempting to allocate already allocated variable '%s'\n",
+                        "Attempting to allocate already allocated variable '%s'",
                         infile,
                         x.base.base.loc,
                         location_manager,
--- a/tests/reference/llvm-allocate_02-4f6b634.json
+++ b/tests/reference/llvm-allocate_02-4f6b634.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "llvm-allocate_02-4f6b634.stdout",
-    "stdout_hash": "2be5b33ab78323d3cb93955d1742fb367436ec3ffce2e74ca3dffe74",
+    "stdout_hash": "2f7dbc0f9edeb4b44e1b9c9c0be1bbefc98b5bcb7331d46ef77345d3",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/llvm-allocate_02-4f6b634.stdout
+++ b/tests/reference/llvm-allocate_02-4f6b634.stdout
@@ -5,9 +5,9 @@ source_filename = "LFortran"
 %dimension_descriptor = type { i32, i32, i32 }
 
 @0 = private unnamed_addr constant [4 x i8] c"arr\00", align 1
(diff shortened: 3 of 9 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2281,13 +2319,229 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, int64_t
...
+        free(filename);
+        return 0;
+    }
+    const char *runtime_prefix = "Runtime error:";
+    if (strncmp(message_start, runtime_prefix, strlen(runtime_prefix)) == 0) {
+        message_start += strlen(runtime_prefix);
+        while (*message_start == ' ') {
+            message_start++;
+        }
+    }
```
xaerru: Instead of checking here we can remove the `"Runtime Error:"` prefix from all the `llvm_utils->generate_runtime_error()` calls. We can also remove the newlines we have in the calls.

Also this code doesn't work all the time. Some references have `runtime error: Runtime Error: message`.

tests/reference/run-allocated_04-b4ea2f8.stderr
```diff
@@ -1,2 +1,5 @@
-At 6:5 of file tests/errors/allocated_04.f90
-Runtime Error: Attempting to allocate already allocated variable 'x'
+[0;31;1mruntime error[0;0m[0;1m: Runtime Error: Attempting to allocate already allocated variable 'x'[0;0m
+ [0;34;1m-->[0;0m tests/errors/allocated_04.f90:6:5
```
xaerru: Here "Runtime Error:" is not skipped.

src/libasr/runtime/lfortran_intrinsics.c
certik: Yes, exactly, I also noticed it.

certik:
Also I want to remove the colors from tests --- it's hard to read. But I want colors in the output. I think we can make it configurable via a compiler option, as well as possibly via an environment variable.

certik:
Related: https://github.com/lfortran/lfortran/issues/9526

src/libasr/runtime/lfortran_intrinsics.c
certik: This is now fixed.

tests/reference/run-allocated_04-b4ea2f8.stderr
certik: Fixed.

certik:
@xaerru I cleaned it up. It's not perfect, there is one error that doesn't show the file line because the location information cannot be extracted. We currently have to extract it from the string. In subsequent PRs, we need to refactor the error printing using the new API (#9526), so until then it's a best-effort basis.

tests/reference/run-scalar_allocation_check_01-a82b6b0.stderr
```diff
@@ -1 +1 @@
-Runtime Error: Variable 'x' is not allocated.
+runtime error: Variable 'x' is not allocated.
```
xaerru, line 1: For this error we don't set the location information.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -10947,7 +10956,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                         llvm::Type::getInt64Ty(context)));
                                 // TODO: Set location here
                                 llvm_utils->generate_runtime_error(cond,
-                                    "Runtime Error: Variable '%s' is not allocated.\n",
+                                    "Variable '%s' is not allocated.",
                                     infile,
                                     {0, 0},
                                     location_manager,
```
xaerru, line 10962: There is a TODO about setting the location here.

certik:
Thanks for the review. Yes, we need to improve that message with location information, and also to change the API to pass the location information as data, not formatted inside the string.

Issue #9409, opened 2026-01-07
Meta: LAPACK xlintstrfc stack overflow (Single Complex RFP)


krystophny:
## Summary

Meta issue: Reference-LAPACK Single Complex RFP test `xlintstrfc` can segfault with the default stack (~8MB) when compiled with LFortran.

## What Is Done (merged)

Complex-temporaries / complex-lowering stack growth fixes:
- #9311
- #9416
- #9503
- #9505

Call-argument temporary alloca proliferation reduction:
- #9531 (reuses `call_arg_value` allocas by LLVM type; reduces `ctfsm` from 77 allocas to 2)

## What Is Still Left

- [ ] **Implicit-interface dimension descriptors**
  - LFortran still materializes many small descriptors for implicit-interface calls (for `ctfsm`: 163 × 32B dimension-descriptor allocas), which keeps the stack frame much larger than GFortran.
  - Goal: avoid per-dimension descriptor allocas for implicit-interface calls (for example by passing raw pointers when safe, or representing the needed physical casts/ABI earlier).

- [ ] **End-to-end verification**
  - Re-test `xlintstrfc` with the default stack after the merged PRs above; it should pass without requiring `ulimit -s 65536`.

## Evidence (ctfsm)

- LFortran stack frame: ~22,632B before #9531, ~17,064B after #9531
- GFortran stack frame: ~104B

## Reproducer

```bash
cd /tmp && git clone --branch v3.12.1 --depth 1 https://github.com/Reference-LAPACK/lapack.git lapack-src
mkdir lapack-build && cd lapack-build
cmake ../lapack-src -G Ninja -DCMAKE_Fortran_COMPILER=lfortran -DBUILD_TESTING=ON -DBUILD_COMPLEX=ON
ninja xlintstrfc
./bin/xlintstrfc < ../lapack-src/TESTING/ctest_rfp.in
```

krystophny:
Here’s a concrete plan to finish the remaining **implicit-interface dimension-descriptor** stack bloat.

## Diagnosis (what is still happening)
The remaining stack growth is coming from **legacy sequence association** patterns like `call foo(a(i))` where `foo` has an **assumed-size** dummy (e.g. `real :: x(*)`) and is compiled via **implicit interface / separate compilation**.

[...]

PR #9531, opened 2026-01-15
Reuse allocas for call argument values by LLVM type

krystophny:
## Summary

- Reuse allocas for implicit interface call argument values by LLVM type
- Reduces excessive stack allocation in functions with many implicit interface calls

Addresses #9409 (partially - addresses call_arg_value alloca proliferation)

## Why

When calling subroutines/functions through implicit interfaces, LFortran must store expression values in memory for pass-by-reference. Previously, each expression argument created a new `call_arg_value` alloca, even when a previous call used the same type. In LAPACK's `ctfsm` function, this caused 77 allocas for a function with only a few distinct argument types.

**Stage:** Codegen

## Changes

- [`src/libasr/codegen/asr_to_llvm.cpp#L288-L312`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L288-L312): Add alloca pool infrastructure - `call_arg_alloca_pool` map and helper functions to get/reset pooled allocas
- [`src/libasr/codegen/asr_to_llvm.cpp#L4816-L4817`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L4816-L4817): Clear pool at function entry (visit_Program)
- [`src/libasr/codegen/asr_to_llvm.cpp#L6043-L6044`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L6043-L6044): Clear pool at function entry (visit_Function)
- [`src/libasr/codegen/asr_to_llvm.cpp#L13851`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L13851): Reset pool at start of convert_call_args
- [`src/libasr/codegen/asr_to_llvm.cpp#L13773`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L13773), [`#L13866`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L13866), [`#L14103`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/src/libasr/codegen/asr_to_llvm.cpp#L14103): Replace direct CreateAlloca with get_call_arg_alloca
- `tests/reference/*`: Updated LLVM IR references (fewer allocas, renumbered labels)

## Tests

- [`integration_tests/implicit_interface_24.f90`](https://github.com/lfortran/lfortran/blob/be92c629318e9fff0abe339425fa338547d064fd/integration_tests/implicit_interface_24.f90): Tests that alloca reuse works correctly with multiple calls using expression arguments

## Verification

For LAPACK's `ctfsm` function:
- Before: 77 `call_arg_value` allocas, 22,632-byte stack frame
- After: 2 `call_arg_value` allocas, 17,064-byte stack frame (97% alloca reduction, 24% stack reduction)

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -287,6 +287,7 @@ public:
     // same type across different calls.
     std::map<llvm::Type*, std::vector<llvm::AllocaInst*>> call_arg_alloca_pool;
     std::map<llvm::Type*, size_t> call_arg_alloca_idx;
+    int convert_call_args_depth = 0;
 
     // Get or create an alloca for a call argument of the given type.
     // Reuses allocas from the pool when possible.
@@ -4818,6 +4819,7 @@ public:
         loop_or_block_end_names.clear();
         call_arg_alloca_pool.clear();
         call_arg_alloca_idx.clear();
+        convert_call_args_depth = 0;
         strings_to_be_deallocated.reserve(al, 1);
         heap_fixed_size_arrays.reserve(al, 1);
         SymbolTable* current_scope_copy = current_scope;
@@ -4878,6 +4880,10 @@ public:
         llvm_goto_targets.clear();
 
         builder->SetInsertPoint(BB);
+        // Clear alloca pool after nested functions are processed, before main body
+        call_arg_alloca_pool.clear();
+        call_arg_alloca_idx.clear();
+        convert_call_args_depth = 0;
         if (compiler_options.emit_debug_info) {
             debug_current_scope = SP;
             builder->SetCurrentDebugLocation(nullptr);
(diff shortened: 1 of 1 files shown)


certik (approved):
I think that this is fine.

Generating more "tight" LLVM is always good. I thought that LLVM has an efficient mem2reg pass that translates these allocas to registers, but you are saying LLVM can't actually figure out some of this itself? I think in general it's always better to use the information we have at a higher level and generate tighter / faster code, so I think this PR is good.

krystophny:
> I think that this is fine.
> 
> Generating more "tight" LLVM is always good. I thought that LLVM has an efficient mem2reg pass that translates these allocas to registers, but you are saying LLVM can't actually figure out some of this itself? I think in general it's always better to use the information we have at a higher level and generate tighter / faster code, so I think this PR is good.

Yes, looks like this is better done earlier in the pipeline. This seems to be one of the last blockers for LAPACK complex number support, where the tests run code that still takes too much stack memory.

PR #9539, opened 2026-01-16
fix: convert DescriptorArrays in `convert_to_polymorphic_type`

### LINKED ISSUE #9515: Handle 0 sized array passes in operator overloaded function
MRE:
```fortran
module string_mod
    implicit none
    public :: string_t, operator(.in.)

    type :: string_t
    end type string_t

    interface operator(.in.)
        module procedure string_in_array
    end interface

contains

    logical function string_in_array(lhs, rhs)
        character(len=*), intent(in) :: lhs
        type(string_t), intent(in) :: rhs(:)
        integer :: i

        string_in_array = .false.
        do i = 1, size(rhs)
            ! if (allocated(rhs(i)%s) .and. allocated(lhs)) then
                if (trim(lhs) == trim(rhs(i)%s)) then
                    string_in_array = .true.
                    return
                end if
            ! end if
        end do
    end function string_in_array

end module string_mod

program mre_in_operator
    use string_mod, only: string_t, operator(.in.)
    implicit none

    type :: source_t
        type(string_t), allocatable :: modules_provided(:)
        type(string_t), allocatable :: modules_used(:)
    end type source_t

    type(source_t) :: source
    integer :: j
    allocate(source%modules_provided(0))
    allocate(source%modules_used(2))
    source%modules_used(1)%s = "stdlib"
    source%modules_used(2)%s = "iso_c_binding"
    ! call check(source)
    do j = 1, size(source%modules_used)
        if (source%modules_used(j)%s .in. source%modules_provided) then
            print *, trim(source%modules_used(j)%s), " is PROVIDED"
        else
            print *, trim(source%modules_used(j)%s), " is NOT provided"
        end if
    end do

end program mre_in_operator
```

Error:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran c.f90 && ./a.out
 s is NOT provided
 i is NOT provided
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90 
At 57:41 of file c.f90
Runtime Error: Array '__libasr_created__function_call_string_in_array@~~in' is not allocated.

Use '--realloc-lhs-arrays' option to reallocate LHS automatically.
```

This works with `--realloc-lhs-arrays`., but should work without it

jinangshah21:
Fixes #9515

### DIFF
--- a/integration_tests/class_80.f90
+++ b/integration_tests/class_80.f90
@@ -7,9 +7,9 @@ module class_80_mod
   end type
 contains 
   subroutine cast_to_keyval(ptr, val)
-   class(class_value), intent(in) :: ptr(:)
-   type(temp_type), intent(in), optional :: val(:)
-!    val(1)%value = ptr(1)%value   ! TODO 
+   class(class_value), intent(inout) :: ptr(:)
+   type(temp_type), intent(inout), optional :: val(:)
+   val(1)%value = ptr(1)%value
 end subroutine cast_to_keyval 
 end module
 
@@ -21,5 +21,5 @@ program class_80
   allocate(y(1))
   y(1)%value = "Hello World"
   call cast_to_keyval(y, x)
-!   if (x(1)%value /= "Hello World") error stop  ! TODO
+  if (x(1)%value /= "Hello World") error stop
 end program class_80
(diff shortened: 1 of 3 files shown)


integration_tests/class_80.f90
```diff
@@ -7,9 +7,9 @@ module class_80_mod
   end type
 contains 
   subroutine cast_to_keyval(ptr, val)
-   class(class_value), intent(in) :: ptr(:)
-   type(temp_type), intent(in), optional :: val(:)
-!    val(1)%value = ptr(1)%value   ! TODO 
+   class(class_value), intent(inout) :: ptr(:)
```
certik, line 10: I think `ptr` doesn't need to be `inout`, correct?

PR #9540, opened 2026-01-16
Fix stack growth in BindC complex returns

krystophny:
## Summary
- Avoid stack growth when lowering BindC complex return conversions by hoisting temporary allocas to the entry block.
- Add a regression test for BindC complex return values in a loop.

Fixes a stack overflow seen when compiling/running complex-heavy third-party code.

## Why
Some BindC complex returns are lowered as vectors (e.g. <2 x float>) and then converted to the internal complex representation using a temporary stack slot. If that alloca is emitted inside a loop, the stack grows per-iteration.

**Stage:** Codegen

## Changes
- https://github.com/krystophny/lfortran/blob/9f6270f6454b053cd1d4cad3f731e9d94f244dad/src/libasr/codegen/asr_to_llvm.cpp#L16005-L16074
- https://github.com/krystophny/lfortran/blob/9f6270f6454b053cd1d4cad3f731e9d94f244dad/integration_tests/bindc_11.f90#L1-L37

## Tests
- ./run_tests.py

### DIFF
--- a/integration_tests/bindc_11.f90
+++ b/integration_tests/bindc_11.f90
@@ -1,0 +1,41 @@
+module bindc_11_mod
+use, intrinsic :: iso_c_binding, only: c_float
+implicit none
+contains
+
+    complex(c_float) function twice(z) result(r) bind(c)
+        complex(c_float), value, intent(in) :: z
+        r = z + z
+    end function twice
+
+end module bindc_11_mod
+
+program bindc_11
+use, intrinsic :: iso_c_binding, only: c_float
+use bindc_11_mod, only: twice
+implicit none
+
+    complex(c_float) :: z, r
+    real(c_float) :: re, im
+    integer :: i
+
+    z = cmplx(5.0_c_float, 7.0_c_float, kind=c_float)
+
+    r = twice(z)
+    re = real(r, kind=c_float)
+    im = aimag(r)
+    if (abs(re - 10.0_c_float) > 1.0e-6_c_float) error stop
+    if (abs(im - 14.0_c_float) > 1.0e-6_c_float) error stop
+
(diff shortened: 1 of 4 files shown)


tests/tests.toml
```diff
@@ -3190,6 +3190,10 @@ asr = true
 filename = "../integration_tests/bindc_06.f90"
 asr = true
 
+[[test]]
+filename = "../integration_tests/bindc_11.f90"
+run = true
```
certik: Remove this reference test. Instead add `if` tests into the integration test to ensure correct results are returned.

integration_tests/bindc_11.f90
```diff
@@ -0,0 +1,37 @@
...
+    r = twice(z)
+    re = real(r, kind=c_float)
+    im = aimag(r)
+    if (abs(re - 10.0_c_float) > 1.0e-6_c_float) error stop
+    if (abs(im - 14.0_c_float) > 1.0e-6_c_float) error stop
+
+    r = cmplx(0.0_c_float, 0.0_c_float, kind=c_float)
+    do i = 1, 700000
+        r = r + twice(z)
+    end do
```
certik, line 33: Does this fail in main?

krystophny: Yes, verified. On main (commit 83d865ba8), the test crashes with a segfault (exit code 139) after running the 700k iteration loop. The stack overflow from unfixed alloca-in-loop behavior causes the crash.

I have also:
- Removed the reference test entry from tests/tests.toml
- Added `if` verification tests to ensure correct results
- Deleted the reference files

All tests pass locally.

certik (approved):
Looks great --- let's run the full tests to make sure.

Issue #9528, opened 2026-01-15
SEGV: EQUIVALENCE and LOGICAL


w6ws:
My PR #8913 described a similar problem and was closed.  It still is an issue:
```
program equiv2
  implicit none
  integer :: ia(5), ic
  logical :: la(2), lc
  equivalence (la(1),lc),  (ia(5), ic)

  ic = -42
  lc = .false.

  ia = [1, 2, 3, 4, 5]
  print *, 'ia(5) =', ia(5), ic, merge (': pass', ': fail', ia(5) == ic)

  la = .true.
  print *, 'la(1), lc =', la(1), lc, merge (': pass', ': fail', lc)

end program
```
```
$ lfortran equiv2.f90
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2808, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2732, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1192, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    result = fe.get_asr2(input, lm, diagnostics);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 276, in LCompilers::FortranEvaluator::get_asr2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
[...]

Issue #8913, opened 2025-11-28
EQUIVALENCE with logical data causes SEGV


w6ws:
```
$ cat log_equiv.f90
      subroutine log_equiv ()
      logical la(2)
      logical l1
      equivalence (la(1), l1)
      end subroutine
$ lfortran -c log_equiv.f90
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2740, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2614, in main_app(int, char**)
    return compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1137, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    result = fe.get_asr2(input, lm, diagnostics);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 276, in LCompilers::FortranEvaluator::get_asr2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::diag::Diagnostics&)
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 298, in LCompilers::FortranEvaluator::get_asr3(LCompilers::LFortran::AST::TranslationUnit_t&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&)
    compiler_options.symtab_only, compiler_options, lm);
[...]

PR #9541, opened 2026-01-16
fix: Handle null types in equivalence statements to prevent segfault

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9528
fixes: https://github.com/lfortran/lfortran/issues/8913

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -245,9 +245,7 @@ static inline const ASR::symbol_t *symbol_get_past_external(const ASR::symbol_t 
 
 static inline ASR::ttype_t *type_get_past_pointer(ASR::ttype_t *f)
 {
-    if (!f) {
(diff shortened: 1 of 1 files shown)


src/libasr/asr_utils.h
```diff
@@ -245,6 +245,9 @@ static inline const ASR::symbol_t *symbol_get_past_external(const ASR::symbol_t
 
 static inline ASR::ttype_t *type_get_past_pointer(ASR::ttype_t *f)
 {
+    if (!f) {
+        return nullptr;
+    }
```
certik: I wouldn't do that --- you can add an assert here that `f` is not null. But I wouldn't accept null, I think it's hiding bugs, isn't it? I would rather check this in the caller.

certik (approved):

PR #9545, opened 2026-01-17
Add regression test for issue 4901

### LINKED ISSUE #4901: substring trouble
This program won't compile with lfortran but 5 other compilers write ```666``` to unit ```*```:
```
program linetest
  character(80):: line
  write(line(1:3),'(I3)') 666
  write(*,'(A)') line(1:3)
end program linetest
```
Compile-time output from lfortran 0.40.0 ends with
```
  %5 = call i8* @_lfortran_str_slice(i8* %4, i32 0, i32 3, i32 1, i1 true, i1 true)
 i8**  call void (i8**, i32*, i8*, ...) @_lfortran_string_write(i8* %5, i32* %6, i8* getelementptr inbounds ([3 x i8], [3 x i8]* @3, i32 0, i32 0), i8* %7)



```

ANAS727189:
Closes #4901.

This PR adds a regression test for the issue reported in #4901 regarding substring formatting. The issue was identified as working in the current `main` branch. This test ensures that writing to a substring (e.g., `write(line(1:3),'(I3)')`) continues to compile and execute correctly in the future.

**Changes**
* Added `integration_tests/issue_4901.f90` containing the reproduction code.
* Updated `integration_tests/CMakeLists.txt` to include the new test.

### DIFF
--- a/integration_tests/issue_4901.f90
+++ b/integration_tests/issue_4901.f90
@@ -1,5 +1,11 @@
 program linetest
-  character(80):: line
-  write(line(1:3),'(I3)') 666
-  write(*,'(A)') line(1:3)
+  character(80) :: line
+  write(line(1:3), '(I3)') 666
+  
+  if (line(1:3) /= "666") then
+    write(*, *) "Test Failed: Expected '666', got '", line(1:3), "'"
+    error stop 1
+  end if
+  
+  write(*, '(A)') line(1:3)
 end program linetest


integration_tests/issue_4901.f90
```diff
@@ -0,0 +1,5 @@
+program linetest
+  character(80):: line
+  write(line(1:3),'(I3)') 666
+  write(*,'(A)') line(1:3)
```
certik: Add a test that `line` contains the correct string.

ANAS727189: I have updated the test to verify that line(1:3) contains the correct string using an if statement.

certik (approved):

PR #9549, opened 2026-01-17
fix: don't change allocatable to pointer arrays in nested_vars and use `move_allocation` for assignment

### LINKED ISSUE #9396: Handle extension of nested subroutine allocatable array var

jinangshah21:
Fixes #9396


certik:
The CI fails with:
```
	969 - modules_50_FAST (SEGFAULT)
```
See if you can reproduce it locally. Some other runs fail with:
```
At 826:13 of file ././src/fpm_targets.f90
Runtime Error: Array 'targets' is indexed but not allocated.
```

certik (approved):
I think this is fine.

PR #9559, opened 2026-01-17
Fix wrong any/count under --fast for logical masks

### LINKED ISSUE #9558: arrays_op_{19,22,23,24}_FAST fail in main on Apple M4
Must apply:
```diff
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1490,7 +1490,7 @@ RUN(NAME select_type_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME select_type_11 LABELS gfortran llvm)
+#RUN(NAME select_type_11 LABELS gfortran llvm)
 RUN(NAME select_type_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST)
 RUN(NAME select_type_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
Or use https://github.com/lfortran/lfortran/pull/9557 which fixes it.

Then:
```
$ ./run_tests.py -j16 -f
...
The following tests FAILED:
        257 - arrays_op_19_FAST (Failed)
        260 - arrays_op_22_FAST (Failed)
        261 - arrays_op_23_FAST (Failed)
        262 - arrays_op_24_FAST (Failed)
Errors while running CTest
Command failed.
```

For example:
```
$ lfortran arrays_op_19.f90
1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01
$ lfortran arrays_op_19.f90 --fast
1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01    1.00000001e-01
ERROR STOP
```
The other three fail in a similar way.

krystophny:
## Summary
- Fix wrong `any`/`count` results under `--fast` for logical masks.
- Fix invalid LLVM IR when converting/passing `class(*)` descriptor arrays in `select type`/`select rank` (LLVM 11 typed-pointers and LLVM 15+ opaque-pointers).

Fixes #9558

## Why
`--fast` produced incorrect results for some logical-mask reductions. While validating across LLVM versions, CI exposed invalid IR in polymorphic `class(*)` array conversion paths (e.g. `select_type_19`).

**Stage:** Codegen

## Changes
- [`src/libasr/codegen/asr_to_llvm.cpp#L14152`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/src/libasr/codegen/asr_to_llvm.cpp#L14152): convert `class(*)` descriptor-array actuals to typed dummy arrays at call sites.
- [`src/libasr/codegen/asr_to_llvm.cpp#L14557`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/src/libasr/codegen/asr_to_llvm.cpp#L14557): ensure descriptor-array GEP base types match the underlying array-descriptor struct type (past allocatable/pointer), so the IR stays well-formed on LLVM <15 and LLVM 15+.
- [`src/libasr/codegen/llvm_array_utils.cpp#L887`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/src/libasr/codegen/llvm_array_utils.cpp#L887): use array element *storage* type (logical arrays are i8-backed) where needed.
- [`src/libasr/codegen/llvm_utils.cpp#L1521`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/src/libasr/codegen/llvm_utils.cpp#L1521) and [`src/libasr/codegen/llvm_utils.h#L255`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/src/libasr/codegen/llvm_utils.h#L255): route GEPs through the safe helpers.
- [`integration_tests/arrays_intrin_09.f90#L1`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/arrays_intrin_09.f90#L1) and [`integration_tests/CMakeLists.txt#L838`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/CMakeLists.txt#L838): add/register a regression integration test.
- `tests/reference/*`: update LLVM references for affected logical/polymorphic array cases.

## Tests
- [`integration_tests/arrays_intrin_09.f90#L1`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/arrays_intrin_09.f90#L1)
- [`integration_tests/select_rank_13.f90#L1`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/select_rank_13.f90#L1)
- [`integration_tests/select_type_18.f90#L1`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/select_type_18.f90#L1)
- [`integration_tests/select_type_19.f90#L1`](https://github.com/lfortran/lfortran/blob/c66d590119fa3406f38919765ffde0b0ce567b72/integration_tests/select_type_19.f90#L1)
- Ran full LFortran unit + integration test suite (LLVM 11) and smoke-ran the select-type/rank cases on LLVM 21.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -6639,19 +6639,13 @@ public:
             ASR::ttype_t* fptr_type = ASRUtils::expr_type(fptr);
             llvm::Type* llvm_fptr_type = llvm_utils->get_type_from_ttype_t_util(fptr,
                 ASRUtils::get_contained_type(fptr_type), module.get());
-            llvm::Value* fptr_array = llvm_utils->CreateAlloca(*builder, llvm_fptr_type);
+            llvm_fptr = llvm_utils->CreateLoad2(llvm_fptr_type->getPointerTo(), llvm_fptr);
             builder->CreateStore(llvm::ConstantInt::get(context, llvm::APInt(32, 0)),
-                arr_descr->get_offset(llvm_fptr_type, fptr_array, false));
+                arr_descr->get_offset(llvm_fptr_type, llvm_fptr, false));
             ASR::dimension_t* fptr_dims;
             int fptr_rank = ASRUtils::extract_dimensions_from_ttype(
                                 ASRUtils::expr_type(fptr),
                                 fptr_dims);
-            llvm::Value* llvm_rank = llvm::ConstantInt::get(context, llvm::APInt(32, fptr_rank));
-            llvm::Value* dim_des = llvm_utils->CreateAlloca(*builder, arr_descr->get_dimension_descriptor_type(), llvm_rank);
-            builder->CreateStore(dim_des, arr_descr->get_pointer_to_dimension_descriptor_array(llvm_fptr_type, fptr_array, false));
-            arr_descr->set_rank(llvm_fptr_type, fptr_array, llvm_rank);
-            builder->CreateStore(fptr_array, llvm_fptr);
-            llvm_fptr = fptr_array;
             ASR::ttype_t* fptr_data_type = ASRUtils::duplicate_type_without_dims(al, ASRUtils::get_contained_type(fptr_type), fptr_type->base.loc);
             llvm::Type* llvm_fptr_data_type = nullptr;
             if (ASRUtils::is_descriptorString(fptr_data_type)) {


certik:
In `main` are logical arrays stored as `i1` bit arrays? So for example `logical :: a(8)` has 8 elements and will occupy 8 bits = 1 byte?

And somehow the LLVM optimizer has some bugs in it and doesn't handle such arrays correctly?

And the fix is to instead store logical arrays as arrays of bytes for 1 byte per element, so `logical :: a(8)` will have 8 elements and will occupy 8 bytes: each element is 1 byte?

If so, I think that looks like the right fix. Bitpacking is an optimization (and it's actually not completely obvious that bit array will always be faster in all circumstances than a byte array) and since LLVM has some issues with it, it would have to be done more carefully, so we'll do it later.

certik:
I also tried to fix this in https://github.com/lfortran/lfortran/pull/9567, but I didn't manage to get all tests to run, so I am giving up for now, but posting the PR for inspiration if it is helpful.

krystophny:
> In `main` are logical arrays stored as `i1` bit arrays? So for example `logical :: a(8)` has 8 elements and will occupy 8 bits = 1 byte?
> 
> And somehow the LLVM optimizer has some bugs in it and doesn't handle such arrays correctly?
> 
> And the fix is to instead store logical arrays as arrays of bytes for 1 byte per element, so `logical :: a(8)` will have 8 elements and will occupy 8 bytes: each element is 1 byte?
> 
> If so, I think that looks like the right fix. Bitpacking is an optimization (and it's actually not completely obvious that bit array will always be faster in all circumstances than a byte array) and since LLVM has some issues with it, it would have to be done more carefully, so we'll do it later.

As far as I understand, this is the case, so I hope that tests will pass in this pull request. I tried to split it already, but it seems that all is needed to get tests to pass together. And then we can look how to do it in ASR in a follow-up, as you say.

krystophny:
> > In `main` are logical arrays stored as `i1` bit arrays? So for example `logical :: a(8)` has 8 elements and will occupy 8 bits = 1 byte?
> > And somehow the LLVM optimizer has some bugs in it and doesn't handle such arrays correctly?
> > And the fix is to instead store logical arrays as arrays of bytes for 1 byte per element, so `logical :: a(8)` will have 8 elements and will occupy 8 bytes: each element is 1 byte?
> > If so, I think that looks like the right fix. Bitpacking is an optimization (and it's actually not completely obvious that bit array will always be faster in all circumstances than a byte array) and since LLVM has some issues with it, it would have to be done more carefully, so we'll do it later.
> 
> As far as I understand, this is the case, so I hope that tests will pass in this pull request. I tried to split it already, but it seems that all is needed to get tests to pass together. And then we can look how to do it in ASR in a follow-up, as you say.

See also #9574 as an alternative fix. Pick the one you prefer and we finish polishing then.

certik:
Can you summarize the differences between the two PRs?

I think the approach in this PR seems ok --- the changes are not too bad. I need to review carefully, hopefully later today.

@jinangshah21, @assem2002 do you think you could please also review this PR?

krystophny:
For reference: PR #9574 addresses the same bugs but takes a different approach—it fixes at the semantics level by inserting explicit `ArrayPhysicalCast` nodes in ASR (adding ~160 lines to `ast_body_visitor.cpp`), so the type narrowing is properly represented in the IR before codegen. This PR (#9559) instead patches at the codegen level using storage types in LLVM lowering. #9574 is architecturally cleaner since it encodes the semantics upstream rather than working around it downstream.

krystophny:
> For reference: PR #9574 addresses the same bugs but takes a different approach—it fixes at the semantics level by inserting explicit `ArrayPhysicalCast` nodes in ASR (adding ~160 lines to `ast_body_visitor.cpp`), so the type narrowing is properly represented in the IR before codegen. This PR (#9559) instead patches at the codegen level using storage types in LLVM lowering. #9574 is architecturally cleaner since it encodes the semantics upstream rather than working around it downstream.

But one could also argue that this PR here is better because it addresses the storage and the codegen level, and doesn't add implementation details to the ASR. So I leave the choice up to you what you prefer.

certik:
I think ASR cast is better, however the PR https://github.com/lfortran/lfortran/pull/9574 also has a bunch of changes in the backend, so I need to carefully review everything to figure it out.

assem2002:
Reviewing...

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7189,17 +7222,132 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                     ASR::ttype_t* target_type_ = ASRUtils::type_get_past_pointer(target_type);
                     switch( ASRUtils::extract_physical_type(target_type_) ) {
                         case ASR::array_physical_typeType::DescriptorArray: {
+                            // class(*) array pointer association (new classes):
+                            //   class(*), pointer :: generic(:)
+                            //   <concrete>, target :: x(:)
+                            //   generic => x
+                            //
+                            // Here, the array descriptor data field points to the polymorphic wrapper
+                            // (e.g. %"~unlimited_polymorphic_type"*) and NOT directly to the concrete data.
```
assem2002, line 7239: I think that's the old -wrong-logic, We should remove that comment,  if that's the case.

assem2002: Ah I see. I think that's fine.

krystophny:
Thank you! Let's wait if the ASR level fix works out first.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7191,17 +7231,132 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            // The wrapper then stores (vptr, i8* data). We must set both correctly.
+                            if (compiler_options.new_classes &&
+                                ASRUtils::is_unlimited_polymorphic_type(target_type_) &&
+                                !ASRUtils::is_unlimited_polymorphic_type(value_type)) {
+
+                                llvm::Type* llvm_target_type = llvm_utils->get_type_from_ttype_t_util(
+                                    x.m_target, target_type_, module.get());
+                                llvm::Value* llvm_target_ = llvm_utils->CreateAlloca(*builder, llvm_target_type);
```
assem2002, line 7247: Here we're creating a new whole array descriptor and then later storing it into target. I think array descriptor  are wrappers to underlying content (data, dimension, etc.) so we shouldn't allocate a new one to hold the pointee information.
It's no problem so far, but should be refactored later.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8117,6 +8294,14 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
         load_unlimited_polymorpic_value(x.m_value, tmp);
         value = tmp;
+        // Logical arrays use byte-backed storage (i8) in memory; convert scalar i1 values
+        // to i8 before storing into an array element.
+        if (x.m_target->type == ASR::exprType::ArrayItem &&
+            ASRUtils::is_logical(*ASRUtils::expr_type(x.m_target)) &&
+            value->getType()->isIntegerTy(1)) {
+            value = builder->CreateZExt(value, llvm::Type::getInt8Ty(context));
+        }
+
```
assem2002, line 8303: Can't we use `i8` with everything? I can see multiple due to this issue.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14211,6 +14434,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                         use_value = true;
                     }
+                    if (use_value) {
+                        // Logical arrays use i8 storage, but scalar logical arguments
+                        // are passed as i1*. If an array element is passed as an
+                        // actual argument, bitcast its pointer to the expected type.
+                        if (ASRUtils::is_logical(*arg_type_) && value->getType()->isPointerTy()) {
+                            llvm::Type* expected_ptr_type = llvm::Type::getInt1Ty(context)->getPointerTo();
+                            if (value->getType() != expected_ptr_type) {
+                                tmp = builder->CreateBitCast(value, expected_ptr_type);
```
assem2002, line 14453: I think that's safe (i8 -> i1)

assem2002 (approved):
Looks good to me.

- the fix related to i8 -> i1, could be hugely minimized by always using i8 and not using i8 with any logical typed variable.

-Class assoicate fix : As ASR still needs further work with classes/types, This fix seems reasonable.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: I started on #9596 with using i8 for logicals in general. more generally one may allow KIND parameter where 1 is 1 Byte, 4 is 4 Bytes like for integers - but not sure of anyone doesnt use default kind for logicals. Somehow llvm wants i1 = 1 BIT for logicals. For scalars this works fine, but in arrays it causes troubles apparently. The other option would be to fix everything with i1 . What do you think is the best option?

assem2002: I'm not sure if we have tests where `kind != default` in our test suite and I think we don't need to support it now, hence I think we can stick with `i8` with any logical-based type (scalar, arrays), then we can extend to support different kinds later.

assem2002: I think the fix would be in one **single spot** eliminating all of these checks to decide whether to cast or not + using `get_el_type` everywhere; If the change got larger, we can stick with the current approach.

certik:
Recommendation: once a PR passes, never rebase. Only merge into it, that way you know what change caused the CI failure. Now the CI is failing, so we don't know what caused it.

Also for review --- it's a huge PR and when you rebase, we now have to review from scratch. If you only push into it, we only have to review what is new.

certik:
Regarding the change: it's doing casts in the backend, while we should be doing all these casts in ASR itself. As a temporary workaround it's fine until we refactor it later, however it does increase the technical debt, so I am not a big fan, but I do want to fix this of course.

krystophny:
> Recommendation: once a PR passes, never rebase. Only merge into it, that way you know what change caused the CI failure. Now the CI is failing, so we don't know what caused it.
> 
> Also for review --- it's a huge PR and when you rebase, we now have to review from scratch. If you only push into it, we only have to review what is new.

Sorry about that, I saw the merge conflict with main but was only thinking about the tests but not the extra effort to review. Next time I just merge from main then.

krystophny:
> > Recommendation: once a PR passes, never rebase. Only merge into it, that way you know what change caused the CI failure. Now the CI is failing, so we don't know what caused it.
> > Also for review --- it's a huge PR and when you rebase, we now have to review from scratch. If you only push into it, we only have to review what is new.
> 
> Sorry about that, I saw the merge conflict with main but was only thinking about the tests but not the extra effort to review. Next time I just merge from main then.

PS: I could restore the state before rebase from local history. So all back and good.

certik:
No worries. Yes, just keep merging main. At the very end after everything passes and it's reviewed, we can do a clean rebase to a few nice commits before we merge.

Thanks for restoring it, much cleaner! It even shows the last passing commit.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: Fixed in 03b54c5d6: stop allocating a fresh array descriptor in CPtrToPointer (use/update the existing descriptor like upstream/main). This was also the cause of the c_ptr_07 CI abort after merging upstream/main; reproduces locally before, now passes on LLVM21 and full LLVM11 test suite.

src/libasr/codegen/asr_to_llvm.cpp
krystophny: Agree this would be cleaner. To keep this PR focused (it’s already large) I didn’t try to refactor all logical lowering to always use i8 in one spot yet. If you want it done here (instead of a follow-up), point me to the preferred “single spot” and I’ll rework it accordingly.

assem2002: I think probably this spot https://github.com/lfortran/lfortran/blob/902635a57254ecd693396cfffa989ade6e590082/src/libasr/codegen/llvm_utils.cpp#L1343-L1347

You can also check `getInt1Ty()` usage across the backend,  if it's reflecting ASR's logical type, we'll probably need to change it to `getInt8Ty()`

krystophny:
@jinangshah21 can you check once more, including the latest commit and merge commit where conflicts were resolved?

integration_tests/CMakeLists.txt
```diff
@@ -835,6 +835,7 @@ RUN(NAME arrays_intrin_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStac
 RUN(NAME arrays_intrin_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran) # minval
 RUN(NAME arrays_intrin_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran) # min
 RUN(NAME arrays_intrin_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray fortran)
+RUN(NAME arrays_intrin_09 LABELS gfortran llvm EXTRA_ARGS --fast) # any/count logical mask
```
certik, line 838: Why do you need this `--fast` argument? Is it not working without it? Each test is tested with and without `--fast` automatically.

certik:
TODO for subsequent PRs:

* https://github.com/lfortran/lfortran/pull/9559/changes#r2709433300
* https://github.com/lfortran/lfortran/pull/9559#discussion_r2713267256
* Figure out how to clean these casts up in ASR itself, to simplify the backend

PR #9562, opened 2026-01-18
fix: correct handling of C pointer array + string descriptors

### LINKED ISSUE #9552: Handle `c_f_pointer` with character array pointer passed to it with c_ptr
MRE:
```fortran

module temp_file_mod
    use iso_c_binding
    implicit none

contains

    function get_temp_filename() result(tempfile)
        integer, parameter :: MAX_FILENAME_LENGTH = 32768
        character(:), allocatable :: tempfile

        type(c_ptr) :: c_tempfile_ptr
        character(len=1), pointer :: c_tempfile(:)

        interface
            function c_tempnam(dir, pfx) bind(C, name="tempnam")
                import :: c_ptr
                type(c_ptr), value :: dir
                type(c_ptr), value :: pfx
                type(c_ptr) :: c_tempnam
            end function c_tempnam

            subroutine c_free(ptr) bind(C, name="free")
                import :: c_ptr
                type(c_ptr), value :: ptr
            end subroutine c_free
        end interface

        c_tempfile_ptr = c_tempnam(C_NULL_PTR, C_NULL_PTR)
        call c_f_pointer(c_tempfile_ptr, c_tempfile, [MAX_FILENAME_LENGTH])
        print *, "Buffer size =", size(c_tempfile)
        print *, "c_tempfile =", c_tempfile
    end function get_temp_filename

end module temp_file_mod


program test_tempnam
    use temp_file_mod
    implicit none

    character(:), allocatable :: name

    name = get_temp_filename()
end program test_tempnam
```

LFortran segfaults

jinangshah21:
Fixes #9552

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4313,14 +4313,6 @@ public:
                                         
                                         ASR::ttype_t* arg_type2 = ASRUtils::type_get_past_allocatable(
                                             ASRUtils::type_get_past_pointer(ASRUtils::expr_type(asr_eq2)));
-                                        ASR::ttype_t* pointer_type_2 = ASRUtils::TYPE(ASR::make_Pointer_t(
-                                            al, asr_eq2->base.loc, ASRUtils::type_get_past_array(arg_type2)));
-                                        ASR::asr_t* get_pointer2 = ASR::make_GetPointer_t(
-                                            al, asr_eq2->base.loc, asr_eq2, pointer_type_2, nullptr);
-                                        ASR::ttype_t *cptr = ASRUtils::TYPE(ASR::make_CPtr_t(al, asr_eq2->base.loc));
-                                        ASR::asr_t* pointer_to_cptr = ASR::make_PointerToCPtr_t(
-                                            al, asr_eq2->base.loc, ASRUtils::EXPR(get_pointer2), cptr, nullptr);
-
                                         Vec<ASR::dimension_t> dim;
                                         dim.reserve(al, 1);
                                         ASR::dimension_t dim_;


certik:
@jinangshah21 the pushed in commits make it pass for me. But I would split it into a separate PR and clean it up first.

Here is some explanation of the issue, however I am not convinced this is the best way to fix it. Either way this should be fixed in a separate PR, together with more tests for the equivalence. Your PR probably just exposed this bug and only on macOS.

# Bug Explanation and Fix

## The Bug

The original code for `equivalence (ia1, ia2(3))` where `ia1(4)` and `ia2(4)` only made `ia1` point to `ia2(3)`, which only provided access to 2 elements (`ia2(3)` and `ia2(4)`). However, `ia1` needs 4 elements, so accessing `ia1(3)` and `ia1(4)` went beyond `ia2`'s allocated memory, causing crashes or incorrect behavior.

**Memory layout problem:**
```
Original ia2:     [1][2][3][4]
                        ↑
ia1 pointed here: [?][?][?][?]  ← Only 2 elements available, but needs 4!
```

In Fortran, `equivalence (ia1, ia2(3))` with these array sizes requires **storage extension** - the total storage must be 6 elements:
```
Correct layout:  [1][2][3][4][5][6]
                  └ia2─┘ └─ia1──┘
```

## The Fix

The fix creates a **shared storage array** when storage extension is needed:

1. **Calculate total storage needed:**
   ```cpp
   total_size = max(size2, offset + size1 - 1)
              = max(4, 3 + 4 - 1) = 6
   ```

2. **Create shared storage variable:**
   - Allocate a new hidden array `__equiv_storage_2_0` of size 6
   - This provides enough space for both arrays

3. **Point both variables into shared storage:**
   - `ia2` → points to `storage(1)` with size 4
   - `ia1` → points to `storage(3)` with size 4

**Result:**
```
__equiv_storage: [1][2][3][4][5][6]
                  └ia2(1:4)┘
                        └ia1(1:4)┘
```

Now:
- `ia1(1) ≡ storage(3) ≡ ia2(3)` ✓
- `ia1(2) ≡ storage(4) ≡ ia2(4)` ✓
- `ia1(3) ≡ storage(5)` ✓ (extended storage)
- `ia1(4) ≡ storage(6)` ✓ (extended storage)

## Key Code Changes

The fix is in `src/lfortran/semantics/ast_common_visitor.h:4368-4538`:

- **Detects when extension is needed:** `if (total_size > size2)`
- **Creates shared storage:** `ASR::make_Variable_t(..., storage_array_type, ...)`
- **Sets up both pointers:** Two `CPtrToPointer` statements pointing to different offsets in the shared storage

certik:
```
wasm-ld: error: CMakeFiles/c_ptr_07.dir/c_ptr_07.f90.o: undefined symbol: tempnam
```
I would remove wasm tests for now.

I think the above thing fixes it, but we need to clean things up.

jinangshah21:
Cleaned up PR: #9585

Issue #9534, opened 2026-01-16
Failing Overwrite with `rec` in FileWrite


yuvrajsinghgmx:
**MRE**

```fortran
program overwrite_with_rec
    use iso_fortran_env, only: int32
    implicit none

    integer(int32) :: x, y ,z
				
    x = 654321
    z = 456465
    open(10, file="data.bin", access="direct", recl=4)
    write(10, rec=1) x
    write(10,rec=1) z
    close(10)

    open(10, file="data.bin", access="direct", recl=4)
    read(10, rec=1) y
    close(10)
    print *,y
end program

```

Current Response : 

`654321`

Expected Response & `gfortran` Response:

`456465`

PR #9563, opened 2026-01-18
fix: recl stored after open with direct acess

yuvrajsinghgmx:
Towards #9534


certik:
Is this fixing something that fails in main but now works (if so, please add this test), or is this a refactoring? If it is a refactoring, can you describe the motivation? So that I can review it.

yuvrajsinghgmx:
> Is this fixing something that fails in main but now works (if so, please add this test), or is this a refactoring? If it is a refactoring, can you describe the motivation? So that I can review it.



We need to store recl(record length) from open() with direct access so that it can be used with write for write()  to work properly with rec because we need record length to calculate the offset value.

I will be adding an  integration test with the implementation of  rec in write() in a subsequent PR.

yuvrajsinghgmx:
@certik I think storing recl is the only suitable option for calculating the offset for write () ?

certik:
I see. If we need it, then indeed we need to pass it to the runtime. I am running full tests.

Go ahead and try to actually implement the feature in a separate PR. That might triggers some rewrites here. It's always better to implement a feature end-to-end if you can, that way you know the design actually works.

yuvrajsinghgmx:
@certik  I have implemented recl in file_inquire() also for which i have added the  integration test  and this successfully verifies the proper implementation of recl storage during file open.

PR #9568, opened 2026-01-18
fix: skip bounds check for absent optional array arguments

### LINKED ISSUE #9561: optional argument mishandled
This program has an absent array optional argument in a subroutine that calls another subroutine where that argument is also absent.
```
program optionaltest1 ! absent optional arguments
  implicit none
  integer :: x(666)
  call outer( x(:) )
contains

  subroutine outer( x, opt )
    integer, intent(inout)        :: x(:)
    integer, intent(in), optional :: opt(:)
    call inner( x(:), opt )
  end subroutine outer

  subroutine inner( x, opt )
    integer, intent(inout)        :: x(:)
    integer, intent(in), optional :: opt(size(x))
    print *,'opt present?',present(opt)
  end subroutine inner

end program optionaltest1
```
Output from lfortran 0.59.0
```
At 10:23 of file optionaltest1.f90
Runtime error: Array shape mismatch in subroutine 'inner'

Tried to match size 1 of dimension 1 of argument number 2, but expected size is 666
```
```
 opt present? F
```

ANAS727189:
Fixes #9561 

This PR fixes the runtime "Array shape mismatch" error for absent optional array arguments passed through nested subroutine calls.

### Problem
When an optional array argument is absent and passed through nested calls, LFortran incorrectly performs bounds checking on the dummy array (size 1) against the expected size, causing a false runtime error.

### Approach
Modified bounds_check_call in `asr_to_llvm.cpp` to:

- Detect the is_X_present_ boolean parameter added by transform_optional_argument_functions pass
- Skip bounds checking when the optional argument is not present

Example :
```fortran
program optionaltest1 ! absent optional arguments
  implicit none
  integer :: x(666)
  call outer( x(:) )
contains

  subroutine outer( x, opt )
    integer, intent(inout)        :: x(:)
    integer, intent(in), optional :: opt(:)
    call inner( x(:), opt )
  end subroutine outer

  subroutine inner( x, opt )
    integer, intent(inout)        :: x(:)
    integer, intent(in), optional :: opt(size(x))
    print *,'opt present?',present(opt)
  end subroutine inner

end program optionaltest1

```

That's a great PR description! It's clear, concise, and follows good practices. Just a minor formatting suggestion - you could add a "Verification" section at the end. Here's a slightly polished version:

Before (error):
```bash
At 10:23 of file optionaltest1.f90
Runtime error: Array shape mismatch in subroutine 'inner'

Tried to match size 1 of dimension 1 of argument number 2, but expected size is 666
```

After (correct):
```bash
lfortran [ fix-issue-9561][+][△ v3.29.1][🐍 v3.12.12][🅒 lf][🧚 ]
❯ ./src/bin/lfortran ./integration_tests/optional_07.f90
opt present?    F 
```

### DIFF
--- a/tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.json
+++ b/tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.stdout",
-    "stdout_hash": "a15882aff6a5a6484c42ff7798c414f3e996c2a9cb66ab91abf2a870",
+    "stdout_hash": "716a234fdb66f5c69792e95289f641a8513eb99ec318c44be3de2d30",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.stdout
+++ b/tests/reference/pass_pass_array_by_data_transform_optional_argument_functions-modules_39-906d6cb.stdout
@@ -340,6 +340,69 @@
                                     (SymbolTable
                                         6
                                         {
+                                            __libasr_is_present_a3:
+                                                (Variable
+                                                    6
+                                                    __libasr_is_present_a3
+                                                    []
+                                                    In
+                                                    ()
+                                                    ()
+                                                    Default
+                                                    (Logical 4)
+                                                    ()
+                                                    Source
+                                                    Public
+                                                    Required
+                                                    .false.
+                                                    .false.
+                                                    .false.
+                                                    ()
+                                                    .false.
+                                                    .false.
+                                                ),
+                                            __libasr_is_present_a4:
+                                                (Variable
+                                                    6
+                                                    __libasr_is_present_a4
+                                                    []
+                                                    In
+                                                    ()
+                                                    ()
+                                                    Default
+                                                    (Logical 4)
+                                                    ()
+                                                    Source
+                                                    Public
+                                                    Required
+                                                    .false.
+                                                    .false.
+                                                    .false.
+                                                    ()
+                                                    .false.
+                                                    .false.
+                                                ),
+                                            __libasr_is_present_a5:
+                                                (Variable
+                                                    6
+                                                    __libasr_is_present_a5
+                                                    []
+                                                    In
+                                                    ()
+                                                    ()
+                                                    Default
+                                                    (Logical 4)
+                                                    ()
+                                                    Source
+                                                    Public
+                                                    Required
+                                                    .false.
+                                                    .false.
+                                                    .false.
+                                                    ()
+                                                    .false.
+                                                    .false.
+                                                ),
                                             a1:
                                                 (Variable
                                                     6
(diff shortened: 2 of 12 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14981,13 +14981,41 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            is_present_flag = tmp; 
+                            
+                            llvm::Function *fn = builder->GetInsertBlock()->getParent();
+                            llvm::BasicBlock *present_checkBB = llvm::BasicBlock::Create(context, "optional_present", fn);
+                            optional_check_mergeBB = llvm::BasicBlock::Create(context, "optional_merge");
+                            builder->CreateCondBr(is_present_flag, present_checkBB, optional_check_mergeBB);
+                            builder->SetInsertPoint(present_checkBB);
+                        }
+                    }
+                }
```
certik, line 15015: Can you explain what this code is doing exactly? It looks like it is inserting a basic block if an optional argument is among the arguments, but what is the purpose and why does it fix the bug?

ANAS727189: @certik  The code detects when an optional array argument has is_X_present_ boolean flag and creates a conditional branch that skips the bounds checking code when the optional argument is absent.
If found, it create a conditional branch:
present_checkBB: Contains the existing bounds checking code (only runs when is_X_present_ == true)
optional_check_mergeBB: It contains the merge point that skips bounds checking (when is_X_present_ == false)

Its basically this - 
```
if (is_opt_present_) {
     Its executing existing bounds checking code
} else {
    Its skipping it
}
```
Thus when the optional argument is absent (is_opt_present_ == false), we jump directly to the merge block and skip all the bounds validation code that would fail .

certik: Perfect, thanks! @xaerru can you please review this PR, to ensure it's ok with regards to bounds checking?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14981,13 +14981,41 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         bool is_method = x.m_dt && !is_nopass;
         for (size_t i = 0; i < x.n_args; i++) {
             ASR::expr_t* arg_expr = x.m_args[i].m_value;
+            if (arg_expr == nullptr) {
+                continue;
+            }
```
xaerru: Is there some case where `arg_expr` is nullptr? Ideally it should never be nullptr.

Integration tests pass if I remove this diff.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14981,13 +14981,41 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                llvm::BasicBlock *optional_check_mergeBB = nullptr;
+                if (i + 1 < x.n_args && x.m_args[i + 1].m_value != nullptr) {
+                    ASR::expr_t* next_arg = x.m_args[i + 1].m_value;
+                    if (ASR::is_a<ASR::Var_t>(*next_arg)) {
+                        ASR::Variable_t* next_var = ASRUtils::EXPR2VAR(next_arg);
+                        std::string var_name = std::string(next_var->m_name);
+                        ASR::ttype_t* var_type = ASRUtils::type_get_past_pointer(next_var->m_type);
+                        if (var_name.size() > 9 && 
+                            var_name.substr(var_name.size() - 9) == "_present_" &&
+                            ASR::is_a<ASR::Logical_t>(*var_type)) {
```
xaerru: This is correct, but we should improve it.

We can modify the name mangling in `transform_optional_argument_functions.cpp`, maybe something like `__libasr_is_present_{name}`. Then only the check for the prefix `__libasr_is_present_` and `Logical` type is sufficient.

src/libasr/codegen/asr_to_llvm.cpp
ANAS727189: I originally added it as safeguard, but you are right it shouldn't be null here, will remove it.

src/libasr/codegen/asr_to_llvm.cpp
ANAS727189: Done @xaerru  . Updated the naming convention to use `__libasr_is_present_{name}` prefix

src/libasr/codegen/asr_to_llvm.cpp
certik: You can change it to an assert.

certik:
@xaerru please go ahead and review again.

src/libasr/codegen/asr_to_llvm.cpp
xaerru: On the next line we take `ASRUtils::expr_type()` of `arg_expr`, this asserts that `arg_expr` is not nullptr. We don't need to add an extra assert.

PR #9569, opened 2026-01-19
Allow to associate to a function result

certik:


certik:
Some tests are disabled, we still need to fix all those.

PR #9574, opened 2026-01-19
Typed views for class(*) arrays + --fast logical masks

### LINKED ISSUE #9573: Clean fix proposal: typed views for class(*) arrays + robust ArrayPhysicalCast lowering
### Summary
We currently have a set of LLVM backend failures/wrong-results that show up in `--fast` mode and/or in `select type` over `class(*)` arrays.

This issue proposes an alternative, cleaner fix: make the required representation conversions explicit in ASR (via existing `ArrayPhysicalCast`/`Associate` patterns) and make the LLVM lowering of `ArrayPhysicalCast` robust to element-type changes (especially when casting a `class(*)` descriptor array view to a typed descriptor array view).

### Minimal repros
1) Wrong result under `--fast` (logical mask with array intrinsics):
```fortran
program arrays_intrin_09
    implicit none
    real :: x(10, 10)
    x = 6.0
    if (count(abs(x - 6.0) > 1e-6) /= 0) error stop
    if (any(abs(x - 6.0) > 1e-6)) error stop
    print *, "ok"
end program
```

2) `select type` + `class(*)` array pointer association (segfault on current `main` when run with LLVM backend):
```fortran
program select_type_18
    implicit none
    logical, target :: l(3)
    class(*), pointer :: generic(:)
    logical, pointer :: p(:)
    l = [.true., .false., .true.]
    generic => l
    select type (generic)
    type is (logical)
        p => generic
        if (count(p) /= 2) error stop
    class default
        error stop
    end select
    print *, "ok"
end program
```

3) `select type` + passing `class(*)` array to typed dummy (LLVM verifier error on current `main` when run with LLVM backend):
```fortran
module select_type_19_m
contains
  subroutine takes_logical_array(a)
    logical, intent(in) :: a(:)
    if (count(a) /= 2) error stop
  end subroutine
end module

program select_type_19
  use select_type_19_m, only: takes_logical_array
  implicit none
  logical, target :: l(3)
  class(*), pointer :: generic(:)
  l = [.true., .false., .true.]
  generic => l
  select type (generic)
  type is (logical)
    call takes_logical_array(generic)
  class default
    error stop
  end select
  print *, "ok"
end program
```

#### A) `select type` narrowing does not survive to codegen (no assoc name)
In `visit_SelectType` the selector’s declared symbol type is temporarily mutated during AST→ASR lowering to perform type-guard checks/validation, but it is restored immediately after the block is transformed. Since expressions refer back to the symbol table at codegen time, the narrowed type is not preserved.

When the selector has **no associate name** and is `class(*)`, the block body ends up calling a typed procedure with an actual argument that still has `class(*)` descriptor type, which later forces codegen to “guess”/patch descriptor element pointer types.

#### B) `ArrayPhysicalCast` lowering assumes source/destination element types match
`ArrayPhysicalCast` already supports `DescriptorArray -> DescriptorArray` casts, but the LLVM lowering path currently copies the data pointer using a type derived from the source expression’s element type and stores it into a destination descriptor whose data field expects a different element pointer type. This leads to LLVM verifier errors like “stored value type does not match pointer operand type” (and can also cascade into “call parameter type does not match function signature”).

#### 1) Preserve `select type` narrowing by introducing a block-local typed view symbol
For each `type is (T)` arm:
- If the user provided an associate name, continue to use it.
- If there is **no associate name**, create a **block-local symbol that shadows the selector name** and gives it the narrowed type.

Then insert an explicit `Associate` statement at the top of the block that binds this symbol to a casted view of the selector expression.

For arrays, use the already-existing ASR node:
- `ArrayPhysicalCast(selector_expr, old_phys, new_phys, type=<narrowed array/pointer type>)`

This matches existing patterns already used elsewhere (e.g. `select rank` uses an `Associate` with `ArrayPhysicalCast` to cast assumed-rank to a descriptor view).

#### 2) Make LLVM lowering of `ArrayPhysicalCast` robust for element-type changes
In the `DescriptorArray -> DescriptorArray` branch of `visit_ArrayPhysicalCastUtil`:
- Load the source data pointer using the *source* descriptor type.
- Convert/bitcast that pointer to the destination element pointer type before storing into the destination descriptor.
- Use `LLVMUtils::get_el_type()` for element types instead of re-deriving types in an inconsistent way.

With this, `class(*)` descriptor arrays can be safely “reinterpreted” into typed descriptor arrays via an explicit ASR cast node, rather than ad-hoc call-site fixes.

#### 3) `--fast` logical masks: avoid `i1` memory for logical arrays
Keep scalar logical semantics as boolean, but for **logical arrays** under `--fast`:
- Use byte-backed storage for elements (e.g. treat element storage type as `i8`) and convert on load/store.
- Make this a single, centralized policy in the LLVM backend (e.g. via `LLVMUtils::get_el_type()` guarded by the `--fast` option), instead of scattered special-cases.

This removes the fragility of bit-packed `i1` arrays under aggressive optimization.

### Benefits
- Reduces backend special-casing: ASR explicitly encodes when a typed view is required.
- Keeps fixes early and uniform: `select type` narrowing becomes structurally visible in ASR.
- Makes `ArrayPhysicalCast` a single, reusable mechanism for descriptor view conversions.
- Limits behavioral changes to `--fast` where applicable.

krystophny:
## Summary
- Preserve `select type` narrowing for `class(*)` array selectors by creating an explicit typed view in ASR.
- Fix LLVM lowering for descriptor-array view casts when the element type changes (notably `class(*)` -> intrinsic).
- Make `--fast` logical array masks use byte-backed storage to avoid wrong results with `any`/`count`.

Fixes #9573

**Stage:** Semantics / ASR + LLVM lowering

## Changes
- `src/lfortran/semantics/ast_body_visitor.cpp`: represent typed selector array views via `ArrayPhysicalCast` (with explicit bounds when required by ASR verify).
- `src/libasr/asr_utils.h`: keep necessary descriptor view casts for `class(*)` arrays; avoid ICE when comparing derived types without expressions.
- `src/libasr/codegen/asr_to_llvm.cpp`, `src/libasr/codegen/llvm_utils.cpp`, `src/libasr/codegen/llvm_array_utils.cpp`: robust lowering for descriptor->descriptor casts and `--fast` logical arrays.
- `integration_tests/select_type_18.f90`, `integration_tests/select_type_19.f90`, `integration_tests/arrays_intrin_09.f90`: regression coverage.

## Tests
- Built LFortran and ran the full unit + integration test suite.
- Verified: `select_type_18`, `select_type_19`, `arrays_intrin_09` (with `--fast`).

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -456,6 +456,11 @@ public:
         // head
         start_new_block(loophead); {
             llvm::Value* cond = condition();
+            // If condition is i8 (Logical), convert to i1 by comparing != 0
+            if (cond->getType()->isIntegerTy(8)) {
+                cond = builder->CreateICmpNE(cond,
+                    llvm::ConstantInt::get(cond->getType(), 0));
+            }
             builder->CreateCondBr(cond, loopbody, loopend);
         }
 
@@ -1977,7 +1982,9 @@ public:
     void visit_ReAlloc(const ASR::ReAlloc_t& x) {
         LCOMPILERS_ASSERT(x.n_args == 1);
         handle_allocated(x.m_args[0].m_a);
-        llvm::Value* is_allocated = tmp;
+        // Convert i8 logical to i1 for boolean operations
+        llvm::Value* is_allocated = builder->CreateICmpNE(tmp,
+            llvm::ConstantInt::get(tmp->getType(), 0));
         llvm::Value* size = llvm::ConstantInt::get(
             llvm::Type::getInt32Ty(context), llvm::APInt(32, 1));
         int64_t ptr_loads_copy = ptr_loads;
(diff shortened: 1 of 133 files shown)


certik:
It looks like there is a reproducible failure at the CI:
```
	429 - arrays_intrin_09 (Failed)
```

certik:
I would think this approach is probably better, but it's still a lot of changes that are hard to review. First let's fix the CI, and then we can maybe split this PR into smaller ones to do these individually:


*    Preserve select type narrowing for class(*) array selectors by creating an explicit typed view in ASR.
*    Fix LLVM lowering for descriptor-array view casts when the element type changes (notably class(*) -> intrinsic).
*    Make --fast logical array masks use byte-backed storage to avoid wrong results with any/count.

Then we can review it and get it in more easily I think.

krystophny:
Splitting into smaller ones right away... see #9591 and #9596

PR #9576, opened 2026-01-19
fix: propagate ABI from Implementation to implicit interface

### LINKED ISSUE #9575: Implicit interface complex function returns lose imaginary part when calling LFortran code
## Problem

Complex-returning Fortran functions lose their imaginary part when called via implicit interfaces under separate compilation.

## Root Cause

**ABI mismatch between Source and BindC calling conventions for complex return values:**

- **Source ABI**: Uses `{float, float}` struct for complex*4 returns
- **BindC ABI**: Uses platform-specific types:
  - Linux: `<2 x float>` (vector type)
  - Windows: `i64` (packed integer)
  - macOS ARM: `{float, float}` (struct)

When calling a function via an implicit interface (which defaults to BindC ABI), the caller expects the platform-specific return type. But if the callee uses Source ABI internally, it returns `{float, float}`, causing the call site to misinterpret the result.

## Solution

Make Source ABI use the same complex return representation as BindC (platform-specific types). This ensures callers and callees agree on the calling convention regardless of how the interface is specified.

## Status

- [x] Root cause identified
- [x] Fix implemented: #9578
- [ ] CI validation

## MRE

```fortran
! func.f90
complex function test_complex()
    test_complex = (21.0, -3.0)
end function

! main.f90
program test
    implicit none
    complex :: c
    complex, external :: test_complex
    c = test_complex()
    print *, c  ! Expected: (21.0, -3.0), got: (21.0, 0.0) before fix
end program
```

Compile with: `lfortran --implicit-interface --separate-compilation func.f90 main.f90 -o test`

krystophny:
## Summary
- Fix ABI mismatch for complex function returns when calling LFortran-compiled functions via implicit interfaces
- When an Implementation is found in the global scope, use its ABI instead of defaulting to BindC

Fixes #9575

## Why
Implicit interfaces default to BindC ABI for external linkage compatibility. However, when calling internal LFortran-compiled functions, this causes an ABI mismatch: BindC uses `<2 x float>` for complex4 returns on Linux, while Source ABI uses `{float, float}`. This results in the imaginary part being lost.

**Stage:** Semantics

## Changes
- [`ast_common_visitor.h#L11118-L11138`](https://github.com/lfortran/lfortran/blob/254385642802d0f4cb987b6fcfa542b4f163dced/src/lfortran/semantics/ast_common_visitor.h#L11118-L11138): Look up Implementation and extract its ABI
- [`ast_common_visitor.h#L11267`](https://github.com/lfortran/lfortran/blob/254385642802d0f4cb987b6fcfa542b4f163dced/src/lfortran/semantics/ast_common_visitor.h#L11267): Use `impl_abi` for return variable
- [`ast_common_visitor.h#L11284`](https://github.com/lfortran/lfortran/blob/254385642802d0f4cb987b6fcfa542b4f163dced/src/lfortran/semantics/ast_common_visitor.h#L11284): Use `impl_abi` for function creation
- `tests/reference/*`: Updated reference output where ABI changed from BindC to Source

## Tests
- [`implicit_interface_27.f90`](https://github.com/lfortran/lfortran/blob/254385642802d0f4cb987b6fcfa542b4f163dced/integration_tests/implicit_interface_27.f90): Tests complex function return via implicit interface calling internal function

## Limitations
This fix works when the Implementation is visible in the same compilation unit (e.g., multiple functions in the same file). It does NOT fix the LAPACK test failures because:
1. LAPACK is compiled with `--separate-compilation`
2. BLAS functions (CDOTC, etc.) are in separate files from the test drivers
3. At semantic analysis time, the Implementation is not in the global scope

Future work may require:
- A compiler flag to force Source ABI for all implicit interfaces (e.g., `--assume-source-abi-external`)
- Link-time ABI matching
- Module files that record ABI information

## Verification
- All unit tests pass
- All 2197 LLVM integration tests pass

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1574,6 +1574,7 @@ RUN(NAME implicit_interface_23 LABELS gfortran llvm EXTRA_ARGS --implicit-interf
 RUN(NAME implicit_interface_24 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_25 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_26 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
+RUN(NAME implicit_interface_27 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 
 
 RUN(NAME implicit_typing_01 LABELS gfortran llvmImplicit)
--- a/integration_tests/implicit_interface_27.f90
+++ b/integration_tests/implicit_interface_27.f90
@@ -1,0 +1,44 @@
+! Test that complex function return values work correctly with implicit interfaces
+! when calling LFortran-compiled Fortran functions.
+! This tests the ABI propagation fix: implicit interfaces should use the
+! Implementation's ABI (Source) when calling internal functions, not BindC.
+
+program test_implicit_interface_complex_return
+    implicit none
+    complex :: result, cdotc_test
+    external cdotc_test
+    complex :: x(3), y(3)
+
+    x = [(1.0, 2.0), (3.0, 4.0), (5.0, 6.0)]
+    y = [(1.0, 1.0), (1.0, 1.0), (1.0, 1.0)]
+
+    result = cdotc_test(3, x, 1, y, 1)
+
+    ! Expected: conjugate(x) dot y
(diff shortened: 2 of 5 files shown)


krystophny:
Closing in favor of a codegen fix that unifies Source and BindC complex return representations. The semantic approach here requires cross-unit visibility which isn't available with separate compilation.

Issue #9525, opened 2026-01-15
Handle multiple nested allocatable statements


YashNagda17:
Related to #9458 

MRE:
```
program mre_modules_assign
    implicit none
    type :: string_t
        character(len=:), allocatable :: s
    end type string_t

    type :: source_t
        type(string_t), allocatable :: modules_provided(:)
    end type source_t

    type :: package_t
        type(source_t), allocatable :: sources(:)
    end type package_t

    type :: model_t
        type(package_t), allocatable :: packages(:)
    end type model_t

    type(model_t) :: model
    type(string_t) :: module
    allocate(model%packages(1))
    allocate(model%packages(1)%sources(2))
    print *, 4
    allocate(model%packages(1)%sources(1)%modules_provided(2))
    print *, 5
    module = model%packages(1)%sources(1)%modules_provided(1)
    contains
    
end program mre_modules_assign
```

Main:
```
$ lfortran a.f90
4
At 27:5 of file a.f90
Runtime Error: Attempting to allocate already allocated variable ''
```

Gfortran:
```
$ gfortran a.f90 & ./a.out
[1]+  Done                    gfortran a.f90
[1] 21313
           4
           5
```

Issue #9458, opened 2026-01-10
Handle multiple nested struct arrayItem in assignment


jinangshah21:
MRE:
```fortran
program mre_modules_assign
    implicit none
    type :: string_t
        character(len=:), allocatable :: s
    end type string_t

    type :: source_t
        type(string_t), allocatable :: modules_provided(:)
    end type source_t

    type :: package_t
        type(source_t), allocatable :: sources(:)
    end type package_t

    type :: model_t
        type(package_t), allocatable :: packages(:)
    end type model_t

    type(model_t) :: model
    ! type(source_t) :: src
    type(string_t) :: module

    allocate(model%packages(1))
    call alloc_src(model%packages(1))
    ! src = model%packages(1)%sources(1)
    module = model%packages(1)%sources(1)%modules_provided(1)
contains
    subroutine alloc_src(package)
        type(package_t), intent(inout) :: package
        allocate(package%sources(2))
        allocate(package%sources(1)%modules_provided(2))
    end subroutine alloc_src
end program mre_modules_assign
```

Outputs:
```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran d.f90 && ./a.out
 Module name: module_one
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran d.f90 --realloc-lhs-arrays
Segmentation fault (core dumped)
```

Adityazzzzz:
I investigated this issue using `gdb` and `--show-llvm`.

**Diagnosis:**
The crash is a Segmentation Fault in the generated LLVM IR in the main program, immediately after returning from `alloc_src`.

**Root Cause:**
It appears to be a failure in propagating the allocation of nested components back to the caller.
1. `alloc_src` correctly performs the allocation for `package%sources(1)%modules_provided`.
2. However, the updated descriptor (pointer) for this deeply nested component is not correctly reflected in the `model` variable in `main`.
3. When `main` attempts the assignment `module = ...`, it loads a `NULL` or invalid pointer for `modules_provided`, resulting in a segfault.

This likely requires a fix in `asr_to_llvm` regarding how nested allocatable derived types are passed via `intent(inout)`.

YashNagda17:
For the statements like:
` model%packages(1)%sources(1)%modules_provided(1)`

The issue here is, for the intermediate struct access, like accessing sources(1) above,  we need to have an additional dereference. 

[...]

Issue #9186, opened 2025-12-24
Bug: allocatable component allocation is incorrect for derived-type arrays


Atharv1405:
MRE
```
program check_nested_alloc
    implicit none

    type :: t1
        integer, allocatable :: arr(:)
    end type t1

    type :: t2
        type(t1), allocatable :: arr(:)
    end type t2

    type :: t3
        type(t2), allocatable :: z(:)
    end type t3

    type(t3) :: obj

    if (.not. allocated(obj%z)) then
        print *, "z NOT allocated"
        allocate(obj%z(2))
    end if

    if (.not. allocated(obj%z(1)%arr)) then
        print *, "z(1)%arr NOT allocated"
        allocate(obj%z(1)%arr(3))
    end if

    if (.not. allocated(obj%z(1)%arr(1)%arr)) then
        print *, "z(1)%arr(1)%arr NOT allocated"
        allocate(obj%z(1)%arr(1)%arr(5))
    end if

    print *, "Final size =", size(obj%z(1)%arr(1)%arr)

end program check_nested_alloc
```

output
```
lfortran$ gfortran t1.f90 && ./a.out
 z NOT allocated
 z(1)%arr NOT allocated
 z(1)%arr(1)%arr NOT allocated
 Final size =           5
lfortran$ ./src/bin/lfortran t1.f90
z NOT allocated
z(1)%arr NOT allocated
z(1)%arr(1)%arr NOT allocated
At 30:9 of file t1.f90
Runtime Error: Attempting to allocate already allocated variable ''
```

PR #9577, opened 2026-01-19
fix: nested ArrayItem and StructInstanceMember

xaerru:
Fix https://github.com/lfortran/lfortran/issues/9186, Fix https://github.com/lfortran/lfortran/issues/9525, Fix https://github.com/lfortran/lfortran/issues/9458

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2672,6 +2672,7 @@ RUN(NAME c_ptr_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME c_ptr_07 LABELS gfortran llvm)
 
 RUN(NAME arrayitem_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
+RUN(NAME arrayitem_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME array_indices_array_item LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME array_indices_array_item_assignment LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


YashNagda17:
Ahh, I was trying on `asr_to_llvm` conversions, struggling to figure out how to distinguish intermediate accesses... After looking here on the solution, it feels very simple... I was on incorrect level all along, which made my life difficult😅

assem2002:
> Ahh, I was trying on `asr_to_llvm` conversions, struggling to figure out how to distinguish intermediate accesses... After looking here on the solution, it feels very simple... I was on incorrect level all along, which made my life difficult😅

@YashNagda17 Yeah that's sometimes the case :). The ASR is just not correct, and it survives every stage of the compiler.
Try to Check the last ASR dump (the one just before backend stage) before you dive deeper and start fixing the backend.

certik:
Let's run full tests just in case.

certik:
Yes, fixing ASR is always better, much simpler than fixing the backend.

certik:
Let's create an MRE for the failure, and add it to integration tests.

xaerru:
MRE:
```fortran
program test2
    implicit none

    type :: a1
        integer, allocatable :: x(:)
    end type

    type :: a2
        type(a1) :: y(4)
    end type

    type :: a3
        type(a2) :: z(3)
    end type

    type :: a4
        type(a3) :: w(2)
    end type

    type(a4) :: obj

    allocate(obj%w(1)%z(2)%y(3)%x(50))
    print *, size(obj%w(1)%z(2)%y(3)%x) ! malloc fails here
end program
```
```shell
$ ./src/bin/lfortran a.f90                                                                   
runtime_error: malloc failed.
```

This has something to do with the arguments passed into `visit_NameUtil()`, we are accessing array index out of bounds. I'll look into it.

assem2002:
Pushed a minor change + added a test.

certik:
The CI looks promising, but let's wait until everything passes, just in case.

PR #9578, opened 2026-01-19
Use platform-specific complex return ABI for both Source and BindC

### LINKED ISSUE #9575: Implicit interface complex function returns lose imaginary part when calling LFortran code

krystophny:
## Summary
- Unify complex return value ABI between Source and BindC to fix implicit interface calls
- All complex-returning functions now use platform-specific ABI representation

Fixes #9575

## Why
When calling Fortran functions via implicit interfaces (which default to BindC ABI), 
the caller expects platform-specific return types (`<2 x float>` on Linux) but Source 
ABI callees returned `{float, float}`, causing the imaginary part to be lost.

**Stage:** Codegen

## Changes
- [`llvm_utils.cpp#L1071-L1100`](https://github.com/lfortran/lfortran/blob/4e51c6f63a23516fd2090a0e810baa6333b7d2c3/src/libasr/codegen/llvm_utils.cpp#L1071-L1100): Remove BindC check, use platform-specific return type for all ABIs
- [`asr_to_llvm.cpp#L6364-L6365`](https://github.com/lfortran/lfortran/blob/4e51c6f63a23516fd2090a0e810baa6333b7d2c3/src/libasr/codegen/asr_to_llvm.cpp#L6364-L6365): Apply return value conversion for all ABIs
- [`asr_to_llvm.cpp#L16165-L16199`](https://github.com/lfortran/lfortran/blob/4e51c6f63a23516fd2090a0e810baa6333b7d2c3/src/libasr/codegen/asr_to_llvm.cpp#L16165-L16199): Apply call result conversion for all ABIs

## Tests
- All LFortran unit and integration tests pass
- LAPACK complex test improvement: 71 passed → 82 passed (11 remaining failures down from 22)

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1575,6 +1575,7 @@ RUN(NAME implicit_interface_23 LABELS gfortran llvm EXTRA_ARGS --implicit-interf
 RUN(NAME implicit_interface_24 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_25 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_26 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
+RUN(NAME implicit_interface_27 LABELS gfortran llvm EXTRA_ARGS --implicit-interface EXTRAFILES implicit_interface_27b.f90)
 
 
 RUN(NAME implicit_typing_01 LABELS gfortran llvmImplicit)
--- a/integration_tests/implicit_interface_27.f90
+++ b/integration_tests/implicit_interface_27.f90
@@ -1,0 +1,9 @@
+program implicit_interface_27
+implicit none
+complex :: c
+complex, external :: test_complex
+c = test_complex()
+if (abs(real(c) - 21.0) > 1e-6) error stop
+if (abs(aimag(c) - (-3.0)) > 1e-6) error stop
+print *, "ok"
+end program
--- a/integration_tests/implicit_interface_27b.f90
+++ b/integration_tests/implicit_interface_27b.f90
@@ -1,0 +1,3 @@
+complex function test_complex()
+    test_complex = (21.0, -3.0)
+end function


certik:
Can you confirm that this PR makes the MRE from https://github.com/lfortran/lfortran/issues/9575 pass, while fails on main? If so, let's add it in a subsequent PR. If not, then maybe the changes in this PR are incorrect?

The MRE apparently passes in main: https://github.com/lfortran/lfortran/issues/9575#issuecomment-3773097495.

certik:
Let's hold on the merging. Let's clarify what exactly this PR fixes with MREs, and ideally add them. I don't like merging changes that add some complexity, if they are not needed. Also the Windows changes are tested, or just best effort?

krystophny:
Found an MRE that **fails on main** - the key is compiling to separate `.o` files:

```fortran
! func.f90
complex function test_complex()
    test_complex = (21.0, -3.0)
end function

! main.f90
program test
    implicit none
    complex :: c
    complex, external :: test_complex
    c = test_complex()
    print *, c
end program
```

**Compile separately to object files:**
```bash
lfortran -c func.f90 -o func.o
lfortran --implicit-interface -c main.f90 -o main.o
lfortran func.o main.o -o test
./test
```

**Result on main:**
```
(2.10000000e+01, 0.00000000e+00)   ← imaginary part lost!
```

**Expected:**
```
(2.10000000e+01, -3.00000000e+00)
```

The LLVM IR confirms the ABI mismatch:
- `func.f90` defines: `define %complex_4 @test_complex()` (returns `{float, float}` struct)
- `main.f90` declares: `declare <2 x float> @test_complex()` (expects `<2 x float>` vector)

Note: Using `--separate-compilation` flag passes because it compiles both files together, allowing consistent ABI choices. The bug only manifests with truly separate object file compilation.

krystophny:
... the Windows changes are covered by existing CI that failed a few times until it was finally fixed. Hope that the new integration test confirms this too.

certik:
It's weird that `--separate-compilation` wouldn't catch this. It does create separate `.o`.

certik:
I can't reproduce the MRE:
```
(lf) ~/repos/lfortran(main)$ vi func.f90
(lf) ~/repos/lfortran(main)$ vi main.f90
(lf) ~/repos/lfortran(main)$ lfortran -c func.f90 -o func.o
(lf) ~/repos/lfortran(main)$ lfortran --implicit-interface -c main.f90 -o main.o

(lf) ~/repos/lfortran(main)$ lfortran func.o main.o -o test
(lf) ~/repos/lfortran(main)$ ./test
(2.10000000e+01, -3.00000000e+00)
```

However I am on Apple M4, which has different ABI than your Linux?

krystophny:
> I can't reproduce the MRE:
> 
> ```
> (lf) ~/repos/lfortran(main)$ vi func.f90
> (lf) ~/repos/lfortran(main)$ vi main.f90
> (lf) ~/repos/lfortran(main)$ lfortran -c func.f90 -o func.o
> (lf) ~/repos/lfortran(main)$ lfortran --implicit-interface -c main.f90 -o main.o
> 
> (lf) ~/repos/lfortran(main)$ lfortran func.o main.o -o test
> (lf) ~/repos/lfortran(main)$ ./test
> (2.10000000e+01, -3.00000000e+00)
> ```
> 
> However I am on Apple M4, which has different ABI than your Linux?

Yes, on Apple the ABI is {float,float} while on Linux it's <2 x float> so you will catch it on Linux and Windows.

krystophny:
> It's weird that `--separate-compilation` wouldn't catch this. It does create separate `.o`.

Ah it seems it also fails if we use `external` interfaces. The problem is that caller expects `BindC` ABI while callee provides `Source` ABI. With interface blocks it works with `--separate-compilation`, and with single compilation too.

PR #9579, opened 2026-01-19
fix: use numerically stable algorithm for complex abs()

### LINKED ISSUE #9580: Complex abs() overflows for large values
## Problem

Complex `abs()` intrinsic overflows for large values, returning infinity instead of the correct result.

## Reproducer

```fortran
program test
    complex :: z
    z = (5.07e30, 2.54e30)
    print *, abs(z)  ! Returns inf, should be ~5.67e30
end program
```

## Root Cause

The implementation uses the naive formula `sqrt(re^2 + im^2)` which overflows when `re` or `im` exceeds ~1e19 (single precision), even though the correct result is representable.

## Impact

This causes LAPACK "type 2" tests (scaled near overflow) to fail because norm computations return infinity.

## Fix

PR #9579 implements the numerically stable algorithm:
```
if |re| >= |im|: abs(z) = |re| * sqrt(1 + (im/re)^2)
else:            abs(z) = |im| * sqrt((re/im)^2 + 1)
```

krystophny:
## Summary
- Use hybrid algorithm for complex `abs()` to handle large values without overflow
- Add integration test `intrinsics_407` for complex abs overflow case

Fixes #9580

## Why
The naive formula `sqrt(re^2 + im^2)` overflows when squaring large values, even when the result is representable. This caused LAPACK "type 2" tests (scaled near overflow) to fail.

**Stage:** Semantics

## Changes
- [`src/libasr/pass/intrinsic_functions.h`](https://github.com/lfortran/lfortran/blob/0cfcf65a2/src/libasr/pass/intrinsic_functions.h#L905-L969): Use hybrid algorithm for complex abs():
  - For normal values: use naive `sqrt(re^2 + im^2)` (more accurate)
  - For large values (|re| or |im| > threshold): use scaled algorithm `|re| * sqrt(1 + (im/re)^2)` to avoid overflow
  - Threshold: `sqrt(HUGE/2)` (~1.3e19 for single, ~9.5e153 for double precision)

## Tests
- [`integration_tests/intrinsics_407.f90`](https://github.com/lfortran/lfortran/blob/0cfcf65a2/integration_tests/intrinsics_407.f90): Tests complex abs() with:
  - Normal values (3,4) -> 5
  - Large values near overflow (5.07e30, 2.54e30)
  - Negative components
  - Asymmetric values
  - Double precision large values

### DIFF
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -4637,21 +4637,17 @@ namespace Hypot {
         fill_func_arg("x", arg_types[0]);
         fill_func_arg("y", arg_types[1]);
         auto result = declare(fn_name, arg_types[0], ReturnVar);
-        // Numerically stable hypot(x,y):
-        // a = max(|x|, |y|), b = min(|x|, |y|)
-        // hypot = a * sqrt(1 + (b/a)^2)  (or 0 if a==0)
+        // Hybrid algorithm:
+        // - Normal values: pow(x**2 + y**2, 0.5) (matches existing `abs(complex)` lowering behavior)
+        // - Large values that might overflow: scaled algorithm
         ASR::ttype_t* real_type = arg_types[0];
         int kind = ASRUtils::extract_kind_from_ttype_t(real_type);
-        double huge = (kind == 4) ? std::numeric_limits<float>::max()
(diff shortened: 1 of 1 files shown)


HarshitaKalani (approved):
Looks good. Thanks @krystophny

PR #9585, opened 2026-01-19
fix: correct handling of C pointer array + string descriptors 

### LINKED ISSUE #9552: Handle `c_f_pointer` with character array pointer passed to it with c_ptr

jinangshah21:
Fixes #9552

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2646,7 +2646,7 @@ RUN(NAME c_ptr_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME c_ptr_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME c_ptr_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME c_ptr_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME c_ptr_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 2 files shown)


certik:
How does this differ to https://github.com/lfortran/lfortran/pull/9562 ? It is fixing an equivalence bug, so we should add tests and explain in the git commit logs how it works and what it is doing. It's ok to merge this to unblock us, but can you then send a PR with some ASR tests for the equivalence change introduced in this PR?

Issue #9519, opened 2026-01-14
ICE when FORMAT precedes READ statement


w6ws:
```
program format_pre
  implicit none

      integer :: itest(3)

! FORMAT statement precedes the READ
100 format (I3,I2,I2)
  read (5, 100) itest

end program
```
```
$ lfortran -c format_pre.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/wws/computer/fortran/lfortran/src/bin/lfortran", in _start
  File "/build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c", line 308, in __libc_start_main
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2808, in main
    return main_app(argc, argv);
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 2674, in main_app(int, char**)
    result = compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/wws/computer/fortran/lfortran/src/bin/lfortran.cpp", line 1249, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
  File "/home/wws/computer/fortran/lfortran/src/lfortran/fortran_evaluator.cpp", line 380, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
    compiler_options, run_fn, "", infile, lm);
[...]

PR #9586, opened 2026-01-19
ICE when FORMAT precedes READ statement (#9519)

henilp105:
Fixes: #9519

When a READ statement referenced a FORMAT label that was already processed (FORMAT before READ), the code set `a_fmt_constant` but never wrapped the values in a `StringFormat` expression for READ statements.Added the same StringFormat wrapping logic for READ statements when `a_fmt_constant` is set.

Added minor fix and integration test.


certik (approved):

certik:
Let's run all tests just in case.

PR #9591, opened 2026-01-20
semantics/codegen: fix class(*) array view casts in SELECT TYPE

krystophny:
## Summary

Fixes `SELECT TYPE` on unlimited polymorphic (`class(*)`) arrays. Before this PR, accessing elements inside `type is (T)` blocks produced garbage or crashed. Now the typed view correctly reads/writes actual data.

```fortran
class(*), pointer :: arr(:)
select type (x => arr)
type is (integer)
    x(1) = 42  ! Now works correctly - was garbage/crash before
end select
```

Related: #9573

### Background: How `class(*)` Arrays Are Stored

In Fortran, `class(*)` (unlimited polymorphic) can hold any type. LFortran represents `class(*)` arrays by storing each element as a **boxed struct**:

```
┌─────────────────────────────────────────────┐
│ class(*) array descriptor                   │
│   data_ptr ──────────────────┐              │
│   dims, offset, rank, ...    │              │
└──────────────────────────────┼──────────────┘
                               ▼
┌─────────────┬─────────────┬─────────────┐
│  Box[0]     │  Box[1]     │  Box[2]     │  ◄─ Array of boxed elements
│ ┌─────────┐ │ ┌─────────┐ │ ┌─────────┐ │
│ │type_id  │ │ │type_id  │ │ │type_id  │ │
│ │.data ───┼─┼▶│.data ───┼─┼▶│.data ───┼─┼▶ actual values
│ └─────────┘ │ └─────────┘ │ └─────────┘ │
└─────────────┴─────────────┴─────────────┘
```

Each box has a `.data` pointer to the actual value. This indirection allows storing any type.

### What Was Broken

When you write:
```fortran
select type (x => generic)
type is (logical)
    print *, x(1)  ! Should print the actual logical value
```

The old code had two problems:

1. **Semantics:** The association `x => generic` was created, but `x` still pointed to the boxed array. No cast was inserted to tell codegen "unwrap the boxes".

2. **Codegen:** Even if a cast existed, the LLVM lowering just bitcast the pointer without extracting `.data` from each box.

Result: `x(1)` returned garbage (the box struct's first field, `type_id`) instead of the actual value.

## Solution Overview

The fix ensures **explicit representation in ASR** (the intermediate representation) that a type conversion is happening, then **correct lowering in codegen** to extract the actual data.

### Part 1: Semantics - Create Explicit Typed Views

**Location:** `ast_body_visitor.cpp` (SELECT TYPE handling)

**What changed:**
- For `type is (T)` guards on arrays, always create a block-local typed view variable
- Insert an `ArrayPhysicalCast` node that represents "convert `class(*)` array → `T` array"
- The cast preserves array shape (dimensions, bounds) but changes element type

**Key additions:**
- `make_typed_selector_view_type()`: Builds the view's type (e.g., `logical(:)` from `class(*)(3)`)
- `element_types_differ()`: Detects when cast is needed (class(*) element ≠ concrete element)

**Before:** No cast in ASR - codegen had to guess
**After:** Explicit `ArrayPhysicalCast(generic, DescriptorArray→DescriptorArray, logical(:))` in ASR

### Part 2: ASR Utils - Preserve Cast Nodes

**Location:** `asr_utils.h` (`make_ArrayPhysicalCast_t_util`)

**What changed:** The utility was eliding (removing) casts when source and target had the same physical layout (`DescriptorArray → DescriptorArray`). But for `class(*) → concrete`, we **need** the cast even though both are descriptor arrays.

**Fix:** Check if we're narrowing from unlimited polymorphic to concrete. If so, keep the cast node.

**Reviewer question answered:** 
> "When does that happen? What is the API of this function?"

The `types_equal(a, b, a_expr, b_expr)` function compares types structurally. The `a_expr`/`b_expr` parameters are optional expressions that provide context for resolving symbol references (e.g., which struct declaration a `StructType_t` refers to).

- When non-null: uses the expression to look up the actual struct/union/function symbols
- When null: falls back to structural comparison of the type's fields (new code in this PR)

The null case now handles: `StructType_t` (compares `m_is_unlimited_polymorphic`, member types), `UnionType_t`, `FunctionType_t`. This is needed because `make_ArrayPhysicalCast_t_util` calls `check_equal_type` without expressions to decide whether to elide the cast.

### Part 3: Codegen - Unwrap Boxed Elements

**Location:** `asr_to_llvm.cpp`

**What changed:** When lowering `ArrayPhysicalCast` from `class(*)` to concrete:

1. Get pointer to first boxed element
2. **Extract `.data` field** from the box (the actual value pointer)
3. Bitcast to concrete element pointer type
4. Store in new descriptor

**Key additions:**
- `unwrap_polymorphic_box_data()`: Loads `.data` field from a box struct
- `get_typed_array_data_ptr()`: Orchestrates unwrap + bitcast for poly→concrete casts

Also tracks `current_selector_type_decl` so element access inside SELECT TYPE blocks knows the original class(*) type declaration.

### Part 4: Array Utils - Fix Element Access

**Location:** `llvm_array_utils.cpp` (`get_single_element`)

**What changed:** When indexing an array inside a SELECT TYPE block where the selector is `class(*)`:

1. Reload the data pointer as `class_type*` (box type)
2. Index to get `box[i]`
3. Extract `box[i].data` 
4. Bitcast to concrete element type

This ensures `x(1)` correctly returns the actual value, not the box.

## Changes Summary

| File | Lines | Purpose |
|------|-------|---------|
| `ast_body_visitor.cpp` | ~80 | Build typed view types, insert ArrayPhysicalCast |
| `asr_utils.h` | ~60 | Structural type comparison, preserve cast for poly→concrete |
| `asr_to_llvm.cpp` | ~100 | Unwrap box data in ArrayPhysicalCast lowering |
| `llvm_array_utils.cpp` | ~45 | Unwrap box data in element access |
| `select_type_20.f90` | 22 | Integration test |

## Test

[`integration_tests/select_type_20.f90`](https://github.com/lfortran/lfortran/blob/76358a9f301ca5590030eca0125983d7677962f8/integration_tests/select_type_20.f90): Tests the full cycle:
- Read element: `if (.not. x(1)) error stop`
- Write element: `x(2) = .true.`
- Pass to intrinsic: `if (count(x) /= 2) error stop`
- Verify aliasing works: writing through `x` modifies original `l`

## Splitting Consideration

The PR could theoretically be split:
- **PR A:** Semantics + ASR utils (preserve ArrayPhysicalCast nodes)
- **PR B:** Codegen + array utils (unwrap boxed elements)

However, PR A alone would produce correct ASR but codegen would crash/produce wrong results. PR B alone wouldn't help because no cast nodes would exist to trigger the unwrapping. They're tightly coupled - both parts are needed for any user-visible improvement.


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -2634,24 +2634,115 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                    al, x.base.base.loc, current_scope, x.m_assoc_name,
-                    nullptr, 0, ASR::intentType::Local, nullptr, nullptr,
-                    ASR::storage_typeType::Default, nullptr, nullptr, ASR::abiType::Source,
+
+	        auto make_typed_selector_view_type = [&](const Location& loc,
+	                ASR::ttype_t* selector_ttype, ASR::ttype_t* guard_type) -> ASR::ttype_t* {
+	            bool is_ptr = ASR::is_a<ASR::Pointer_t>(*selector_ttype);
+	            bool is_alloc = ASR::is_a<ASR::Allocatable_t>(*selector_ttype);
+	            ASR::ttype_t* base = ASRUtils::type_get_past_allocatable(
+	                ASRUtils::type_get_past_pointer(selector_ttype));
```
certik: This looks like it's incorrectly indented, isn't it?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -2634,24 +2634,115 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    ASR::StructType_t* to_st = ASR::down_cast<ASR::StructType_t>(to_el);
+                    return from_st->m_is_unlimited_polymorphic != to_st->m_is_unlimited_polymorphic;
+                }
+                default: {
+                    return false;
+                }
+            }
+        };
+
+	        for( size_t i = 0; i < x.n_body; i++ ) {
```
certik: Also incorrect indentation?

src/libasr/asr_utils.h
```diff
@@ -4106,6 +4106,29 @@ inline bool types_equal(ASR::ttype_t *a, ASR::ttype_t *b, ASR::expr_t* a_expr, A
                 return types_equal(a2->m_type, b2->m_type, a_expr, b_expr);
             }
             case (ASR::ttypeType::StructType) : {
+                // `types_equal()` is sometimes called without providing expressions. In that
+                // case, fall back to a structural comparison to avoid ICE/segfault.
+                if (a_expr == nullptr || b_expr == nullptr) {
```
certik, line 4128: When does that happen? What is the API of this function? I.e., what is the meaning of these arguments:
```
inline bool types_equal(ASR::ttype_t *a, ASR::ttype_t *b, ASR::expr_t* a_expr, ASR::expr_t* b_expr,
```
Let's document it. When do you pass `a_expr` and when do you pass null there?

src/lfortran/semantics/ast_body_visitor.cpp
krystophny: fixed

src/libasr/asr_utils.h
krystophny: Added documentation: [`asr_utils.h#L4015-L4031`](https://github.com/lfortran/lfortran/blob/68b21eab3/src/libasr/asr_utils.h#L4015-L4031)

- Non-null `a_expr`/`b_expr`: looks up actual symbols via `get_struct_sym_from_struct_expr()`
- Null: structural comparison of type fields (needed when expressions unavailable, e.g., in `make_ArrayPhysicalCast_t_util`)

src/libasr/asr_utils.h
```diff
@@ -4019,6 +4019,23 @@ inline bool dimensions_compatible(ASR::dimension_t* dims_a, size_t n_dims_a,
     return (total_a == -1) || (total_b == -1) || (total_a >= total_b);
 }
 
+// Compares two ASR types for structural equality.
+//
+// Parameters:
+//   a, b: The types to compare (required, but nullptr is allowed for both).
```
certik, line 4025: Good documentation. Why are nullptr allowed? What is the advantage? I would disallow it. This doesn't need to be done in this PR.

certik:
The PR description is very helpful. I will review this carefully.

certik:
Is https://github.com/lfortran/lfortran/pull/9673 handling a similar issue?

krystophny:
## Merge Conflict Resolution

Merged upstream/main and resolved conflicts with PR #9673.

### What was in conflict

Both this PR and #9673 addressed issues in SELECT TYPE handling:

| Location | PR #9673 (upstream) | This PR |
|----------|---------------------|---------|
| `ast_body_visitor.cpp` | Inline array type mutation: duplicates type and sets `arr_ty->m_type = selector_type` | Uses `make_typed_selector_view_type()` helper to build proper typed view |
| `asr_to_llvm.cpp` | Inline workaround with `// REMOVE` comments for unlimited poly → concrete | Uses `get_typed_array_data_ptr()` helper for clean polymorphic box unwrapping |

### Resolution

Kept this PR's approach because:
1. **Explicit ASR representation**: The view type is built in semantics, not worked around in codegen
2. **Helper functions**: `make_typed_selector_view_type()` and `get_typed_array_data_ptr()` are reusable and documented
3. **Descriptor metadata**: Uses bitcast + proper `reset_array_details` for copying dims/offset/rank

Also updated `reset_array_details` calls to match new 5-argument signature (added `source_arr_type` parameter from upstream).

### What this PR solves (in addition to resolving the conflict)

This PR provides a more comprehensive fix for `class(*)` array access in SELECT TYPE:
- Semantics creates explicit `ArrayPhysicalCast` nodes (not just type mutation)
- Codegen properly unwraps boxed polymorphic elements via `.data` field extraction
- Array element access inside type guards correctly returns actual values, not box metadata

All tests pass locally.

certik:

assem2002:
Reviewing..

src/libasr/asr_utils.h
assem2002: The Function's is expecting an expression (mostly a `Var`->`Variable` sym) to get a `Struct` symbol from, hence know information about the passed `structType` types(name of member, member functions, etc..)
 That's unfortunately due to how we're representing `structType` -- It doesn't point to the struct symbol, hence we need to pass an expression if the type is `structType` otherwise `nullptr` passed as expression is permissible with other types (it's not needed). 
It's not the best way to represent a type (depending on another symbol and not even point to it), but we can refactor that later.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7393,13 +7406,33 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                            target_desc,
+                                            llvm::ConstantPointerNull::get(target_array_desc_type->getPointerTo())),
+                                        [&]() {
+                                            llvm::Value* new_desc = llvm_utils->CreateAlloca(*builder, target_array_desc_type);
+                                            // Allocate space for dimension descriptors
+                                            llvm::Value* dim_ptr = llvm_utils->CreateAlloca(*builder, dim_desc_type, n_dims_val);
+                                            builder->CreateStore(dim_ptr, llvm_utils->create_gep2(target_array_desc_type, new_desc, 2));
+                                            // Store rank
+                                            builder->CreateStore(n_dims_val,
+                                                llvm_utils->create_gep2(target_array_desc_type, new_desc, 4));
```
assem2002, line 7442: Any ArrayDescriptor Variable should be already setup _ed_  . Fine to use, but should be refactored.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8908,34 +8984,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-            }
+            ASR::ttype_t* src_asr_type = ASRUtils::expr_type(m_arg);
+            ASR::ttype_t* dst_asr_type = m_type;
+            llvm::Type* dst_el_type = llvm_utils->get_el_type(
+                m_arg, ASRUtils::extract_type(ASRUtils::type_get_past_allocatable_pointer(dst_asr_type)), module.get());
+            llvm::Value* src_data_ptr = llvm_utils->CreateLoad2(data_type->getPointerTo(),
+                arr_descr->get_pointer_to_data(m_arg, src_asr_type, arg, module.get()));
+            llvm::Value* src_first_el_ptr = llvm_utils->create_ptr_gep2(data_type, src_data_ptr,
+                arr_descr->get_offset(arr_type, arg));
+            tmp = get_typed_array_data_ptr(src_asr_type, dst_asr_type, data_type, dst_el_type, src_first_el_ptr);
```
assem2002, line 9005: Good step towards refactoring our `class(*)` -> `T` logic.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -2880,30 +2933,64 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                                    assoc_arr->m_type, dims.p, dims.size(),
+                                    ASR::abiType::Source, false, assoc_arr->m_physical_type, true);
+                            }
+                            ASR::array_physical_typeType old_ptype =
+                                ASRUtils::extract_physical_type(ASRUtils::expr_type(m_selector));
+                            ASR::array_physical_typeType new_ptype =
+                                ASRUtils::extract_physical_type(assoc_variable->m_type);
+                            assoc_value = ASRUtils::EXPR(ASRUtils::make_ArrayPhysicalCast_t_util(
+                                al, type_stmt_type->base.base.loc, m_selector,
+                                old_ptype, new_ptype, cast_type, nullptr));
```
assem2002, line 3017: Probably we should introduce some new node to cast Arrays. The physicalCast here is playing the role of actual-type cast, We should separate those concerns.

krystophny:
Thank you! I filed a follow-up issue #9687 for your comments.

certik:
Thanks @assem2002 for the review! Let's merge it then.

PR #9594, opened 2026-01-20
ICE: when READ with implied DO (#9518)

henilp105:
fixes: #9518 

When using READ with an implied DO loop and an explicit format string like `READ(5, '(100i5)') IVI, (I1I(JVI),JVI=1,IVI)`, the compiler crashes with "visit_ImpliedDoLoop() not implemented".

Cause: In `emit_formatted_read`  function in `asr_to_llvm.cpp`, when iterating through values to read, the code called `this->visit_expr(*val_expr)` directly. However, there's no visit_ImpliedDoLoop implementation in the LLVM visitor - implied DO loops should be expanded into actual loops at codegen time.

Approach:

modified: `emit_formatted_read` in `asr_to_llvm.cpp` to:

- Detect implied DO loops - Check if any value in the READ statement is an ImpliedDoLoop
- Handle implied DO loops specially - When detected:
- Generate a loop that iterates over the DO range
- Read each array element individually using _lfortran_formatted_read with no_of_args=1
- Use ADVANCE='NO' to keep reading from the same line

After the loop, call _lfortran_empty_read to consume the rest of the line

Added fix and integration test.


certik:
Based on the PR description this implements the do loops in the backend. However, I think we expand them in an ASR pass, correct? If so, then this bug should be fixed there rather than in the backend?

henilp105:
yes correct, it should be better handled at ASR pass  `src/libasr/pass/implied_do_loops.cpp`  rather than the backend, I will go through code  (and try better approach) and open another PR.

the simplest approach that i can currently think of that will work is to modify the ASR pass to flatten the ImpliedDoLoop when the bounds are constant, and leave it for the backend when they're not. For the backend case, we need a proper runtime solution. does this look like a suitable approach?

PR #9595, opened 2026-01-20
Add support for `call abort`

Leon-Degel-Koehn:
ABORT is an extension to the fortran standard.
It is however supported by many standard fortran compilers such as gfortran (https://gcc.gnu.org/onlinedocs/gfortran/ABORT.html) and commonly used.
I have personally ran into this problem while trying to compile eCLM (https://github.com/HPSCTerrSys/eCLM) and wanted to share my changes to the compiler which removed this issue in my path towards compilation.

### DIFF
--- a/integration_tests/abort_01.f90
+++ b/integration_tests/abort_01.f90
@@ -1,0 +1,4 @@
+program abort1
+  implicit none
+  if (.false.) call abort()
+end program
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -292,6 +292,10 @@ asr = true
 julia = true
 
 [[test]]
+filename = "../integration_tests/abort_01.f90"
+llvm = true
+
+[[test]]
 filename = "errors/array_size_01.f90"
 asr = true
 
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -355,6 +355,7 @@ RUN(NAME program_cmake_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 
 
 RUN(NAME internal_subr_host_vars_01 LABELS gfortran llvm)
 
+RUN(NAME abort_01 LABELS gfortran llvm)
 RUN(NAME error_stop_01 FAIL LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm llvm2 fortran)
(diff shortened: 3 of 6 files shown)


certik:
@Leon-Degel-Koehn thanks for implementing it! Do other compilers warn about using an extension? I checked on godbolt and gfortran and flang do not seem to emit a warning, so for now the PR is fine.

We also need to figure out how to test it, in our integration tests we can test that a given test should fail, see `integration_tests/error_stop_01.f90` and its registration in integration_tests/CMakeLists.txt. Let's add such a test for `abort`.

@HarshitaKalani, do you mind reviewing the details of the new intrinsic function implementation?

keyvizsla:
It seems to me that there is a limitation with CTest that you can't check if a program aborts.
Simply adding the FAIL flag as we do for the `error stop` tests does not suffice.
From what I am able to tell from researching the topic online, the only workaround is to add a custom script that is executed instead, which runs the fortran program under test (containing call abort), and handles the logic itself.
To remain as minimally invasive to the current infrastructure as possible I have put this logic in the already existing python script.
Let me know what you think, I suppose there are different ways of going about this and I am of course open for suggestions.

certik:
In that case, just make it a passing test: it will never hit the `abort` line at runtime, so it will pass. But it will still test that it compiles, links and runs. We can also add an LLVM reference test that will show that abort() is generated and called correctly --- just add `llvm=true` test into `tests/tests.toml`.

Let's not extend the testing infrastructure with custom scripts or changes, because those changes are easy to get wrong and we need to test it at the CI that they break when they should, which adds to our plate. :)

keyvizsla:
All set now, rebased and added the LLVM reference test. Happy to adjust if needed.

tests/tests.toml
```diff
@@ -291,6 +291,10 @@ ast = true
 asr = true
 julia = true
 
+[[test]]
+filename = "abort1.f90"
```
```suggestion
filename = "../integration_tests/abort1.f90"
```

integration_tests/abort_01.f90
```diff
@@ -0,0 +1,4 @@
+program abort1
+  implicit none
+  if (.false.) call abort()
+end program
```
certik, line 4: Move this to integration_tests and register add `gfortran llvm` labels.

certik (commented):
Test it in both integration tests and reference tests, per my comments. The integration tests still check that it can line properly, i.e., that `_lcompilers_abort` is defined properly.

Otherwise it looks great now.

certik:
@Leon-Degel-Koehn thank you for your first contribution --- send us more PRs! :)

PR #9600, opened 2026-01-20
enh: implement pos for FileRead

jinangshah21:
Towards #6840


integration_tests/read_30.f90
```diff
@@ -0,0 +1,47 @@
...
+         form="unformatted", position="append", status="old", iostat=stat)
+    if (stat /= 0) error stop "open(read) failed"
+
+    inquire(unit=io_r, size=filesize)
+    if (filesize <= 1) error stop "config.toml too small"
+
+    allocate(character(len=filesize) :: buffer)
+    buffer(filesize:filesize) = "?"
+
+    read(io_r, pos=1, iostat=stat) buffer(:filesize-1)
```
certik, line 30: Should we also test some other `pos` than 1?

The file is already at `pos=1`, so this doesn't do anything. But for `pos` other than 1, it will actually have to seek.

Also we should check that `buffer` contains the correct characters as read from the file.

certik (approved):
I think it looks good. I think the testing could be improved per my comment. Can be done in a subsequent PR.

integration_tests/read_30.f90
jinangshah21: File is opened in `append` mode so originally it is at last position, that's why I checked for `pos=1`. 
Also at last I have checked content of buffer.

certik: Perfect. Good to merge then.

certik:
Resolve the conflict, and if a few tests pass, we can merge.

Issue #9587, opened 2026-01-20
CI failure: 	1017 - select_type_15 (Subprocess aborted)


certik:
https://github.com/lfortran/lfortran/actions/runs/21144207290/job/60805653084

```
	1017 - select_type_15 (Subprocess aborted)
```

certik:
Here again:
```
	1018 - select_type_15 (Subprocess aborted)
```
https://github.com/lfortran/lfortran/actions/runs/21188859512/job/60950396551?pr=9606

PR #9606, opened 2026-01-20
Implement "-t" and ninja support to run integration tests

certik:
Implemented `-t` for selecting a particular test, for example:
```
./run_tests.py -t namelist_01
```
Will only compile and run this test, which speeds up testing a lot.

Currently ninja off by default, must enable with `--ninja`. Nothing has thus changed by default.

The change in `integration_tests/modules_50.f90` was discovered by `ninja`, it would fail to build. It's an actual bug in our tests even for make (`.mod` file race condition), which this PR fixes.

### DIFF
--- a/integration_tests/run_tests.py
+++ b/integration_tests/run_tests.py
@@ -18,6 +18,8 @@ LFORTRAN_PATH = f"{BASE_DIR}/../src/bin:$PATH"
 fast_tests = "no"
 nofast_llvm16 = "no"
 separate_compilation = "no"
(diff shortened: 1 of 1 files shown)


certik:
Thanks to the flaky test https://github.com/lfortran/lfortran/issues/9587, it nicely tested that the still CI fails for failing tests. So this PR is now well tested and can be merged.

certik:
Let's run the full checks just in case to make sure the extra flags work.

PR #9639, opened 2026-01-21
Allow nested functions with namelists

### LINKED ISSUE #9636: Handle namelists from parent program in subroutine
MRE:
```fortran
program namelist_15
    implicit none

    ! Variables of different intrinsic types
    integer           :: act_i
    real              :: act_r
    character(len=10) :: act_c
    logical           :: act_l

    integer :: lun

    namelist /act_cli/ act_i
    namelist /act_cli/ act_r, act_c
    namelist /act_cli/ act_l

    act_i = 42
    act_r = 3.5
    act_c = 'hello'
    act_l = .true.
    call tmp_error()

contains 

    subroutine tmp_error()
        open(file='_test_cli', newunit=lun)
        write(lun,nml=act_cli)
        close(unit=lun)
    end subroutine
end program namelist_15
```

certik:
Fixes #9636.

  subroutine_from_function.cpp

  - What: get_resolved_symbol() now checks whether a symbol actually comes from
    a module before creating an ExternalSymbol. If there’s no module owner, it
    returns the local symbol instead of forcing an external.
  - Why: The previous logic always assumed module ownership; for local or host
    symbols that breaks name resolution (or creates redundant externals).
  - How the fix works: Grab the owner module once, short‑circuit when missing,
    and only build an external when the original symbol is truly module‑scoped.
    This keeps local symbols local and avoids accidental externals.

  nested_vars.cpp

  - What: This is the main change: nested variable capture now handles namelists
    in nested procedures, avoids creating duplicate externals, and only treats
    “nested calls” as special when they actually refer to host‑nested
    procedures.
    capture namelist vars, and it also over‑eagerly replaced variables for any
    call, plus it duplicated externals or used names that collided.
  - How the fix works (high level):
        the already‑captured external vars, and rewrites namelist references in
        nested scopes to that new symbol.
      - Reuses a single external symbol per captured variable
        (module_var_to_external) and avoids duplicating externals when a symbol
        is already in the scope chain.
      - Restricts “calls_present” and replacement behavior to actual nested
        calls (not every call), preventing unnecessary replacement of vars.

  asr_to_llvm.cpp

  - What: Two adjustments:
      1. Move _lpython_call_initial_functions to the very top of main, before
         variable declarations/initializations (restoring the original order).
      2. When lowering namelists, use the original variable name if the var is
         an external symbol (important for nested‑module externals), and resolve
         the variable symbol past externals before lookup.
  - Why: The main‑init call must run first (command line + seed initialization)
    — it was moved later by the simplification. For namelists, nested capture
    produces external symbols whose display name can differ; using the original
    name ensures runtime namelist lookup matches source names.
  - How the fix works: Emit the init call right after the entry block begins,
    then declare vars; and use a helper to get the original name from
    ExternalSymbol when building the namelist descriptor.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -871,6 +871,7 @@ RUN(NAME namelist_19 LABELS gfortran llvm)
 RUN(NAME namelist_20 LABELS gfortran llvm)
 RUN(NAME namelist_21 LABELS gfortran llvm)
 RUN(NAME namelist_22 LABELS gfortran llvm)
+RUN(NAME namelist_23 LABELS gfortran llvm)
 
 RUN(NAME nested_namelist_01 LABELS gfortran llvm)
 
--- a/integration_tests/namelist_23.f90
+++ b/integration_tests/namelist_23.f90
@@ -1,0 +1,43 @@
+program test_namelist_internal_proc
+    implicit none
+
+    integer           :: act_i
+    real              :: act_r
+    character(len=10) :: act_c
+    logical           :: act_l
+
+    integer :: lun
+
+    namelist /act_cli/ act_i
+    namelist /act_cli/ act_r, act_c
+    namelist /act_cli/ act_l
+
+    act_i = 42
+    act_r = 3.5
+    act_c = 'hello'
+    act_l = .true.
+
+    call write_namelist()
+
+    act_i = 0
+    act_r = 0.0
+    act_c = ''
+    act_l = .false.
+
+    open(unit=10, file='namelist_23.dat', status='old', form='formatted')
+    read(10, nml=act_cli)
+    close(10)
+
+    if (act_i /= 42) error stop
+    if (abs(act_r - 3.5) > 1.0e-5) error stop
+    if (act_c /= 'hello') error stop
+    if (.not. act_l) error stop
+
+contains
+
+    subroutine write_namelist()
+        open(file='namelist_23.dat', newunit=lun, status='replace', form='formatted')
+        write(lun, nml=act_cli)
+        close(unit=lun)
+    end subroutine write_namelist
+end program test_namelist_internal_proc


certik:
It should be reasonably clean.

tests/reference/llvm-allocate_02-4f6b634.stdout
```diff
@@ -31,7 +29,8 @@ define i32 @main(i32 %0, i8** %1) {
   %9 = getelementptr %array, %array* %arr_desc, i32 0, i32 0
   store i32* null, i32** %9, align 8
   store %array* %arr_desc, %array** %arr, align 8
-  %i1 = alloca i32, align 4
+  %i = alloca i32, align 4
+  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
```
certik: This function should be called first, this change seems wrong to me.

certik: I fixed it.

certik:
Let's extract the nested_vars change into a separate PR, with the LLVM changes.

certik:
We get an error at the CI:
```
# Starting library-module-use ... (1/21)
runtime error: Array 'targets' is indexed but not allocated.
   --> ././src/fpm_targets.f90:850:43
    |
850 |                 call collect_used_modules(targets(i)%ptr, modules_used)
    |                                           ^^^^^^^^^^^^^^^ 
<ERROR> Execution for object " fpm-test " returned exit code  1
STOP 1
```

We'll have to extract an MRE for it, unless it's a bug upstream.

certik:
I created a simple PR from this: https://github.com/lfortran/lfortran/pull/9666, which we can get in first.

certik:
The added test `integration_tests/namelist_23.f90` now passes in main, so it seems the changes in this PR are not needed.

certik:
The LLVM IR however is simpler, so might be worth doing anyway.

certik:
The test passes in main, so submitted separately: https://github.com/lfortran/lfortran/pull/9693.

certik:
Another good change in this PR is to remove pre-declaration that was never used, so I extracted just that change into https://github.com/lfortran/lfortran/pull/9694.

certik:
The last good change was a namelist bug fix, which I extracted into https://github.com/lfortran/lfortran/pull/9705. It required a new test that is not part of this PR 9639, but the new test indeed fails in main and passes with the new PR 9705.

The remaining change left in this PR is a rearrangement of declarations but that seems to be introducing potential bugs and doesn't seem to fix anything, so we will not use that change.

I am closing this PR, as all the good parts were extracted out.

PR #9643, opened 2026-01-21
fix: E format exponent calculation for values near powers of 10

### LINKED ISSUE #9641: E format output produces incorrect exponent for very small numbers
E format output produces incorrect exponent for very small numbers when mantissa rounds to near 1.0.

## Reproducer

```fortran
program test_e_format
    use, intrinsic :: iso_fortran_env, only: int64, real64
    implicit none
    real(real64) :: val
    integer(int64) :: bits

    val = 1.0d-200
    bits = transfer(val, bits)
    print '(A,Z16)', 'Hex: ', bits
    print '(A,E25.17)', 'E25.17: ', val
end program
```

## Expected (gfortran)

```
Hex: 16687E92154EF7AC
E25.17:   0.99999999999999998-200
```

## Actual (LFortran)

```
Hex: 16687E92154EF7AC
E25.17:  0.99999999999999998-199
```

## Analysis

The binary representation is identical (hex `16687E92154EF7AC`), but the printed exponent differs by 1:
- gfortran: `-200` (correct)
- LFortran: `-199` (incorrect)

The bug appears when:
1. The value is very small (extreme negative exponent)
2. High precision format is used (E25.17, E30.22, etc.)
3. The mantissa rounds to a value very close to 1.0 (e.g., `0.99999999999999998`)

Lower precision formats like E20.13 produce correct output because the mantissa displays as `0.1...` instead of `0.999...`.

## Additional observations

For ES format, the same exponent-off-by-one issue occurs:
- gfortran ES25.17: `9.99999999999999982-201`
- LFortran ES25.17: `9.99999999999999982-200`

The issue likely stems from exponent adjustment logic not properly accounting for mantissa rounding that pushes the displayed value toward 1.0.

krystophny:
## Summary
- Fix E format exponent calculation for values near powers of 10
- Fix E format indentation for 3-digit exponents
- Add integration test that fails on main, passes with fix

Fixes #9641

## Bug 1: Exponent off-by-one (main)
```
$ lfortran format_49.f90
FAIL: 1.0d-200 exponent should be -200, got -199
Output was: 0.99999999999999998-199
```

## Bug 2: Indentation off-by-one (main)
```
$ gfortran a.f90 && ./a.out
  0.10000000000000000+101
$ lfortran a.f90
 0.10000000000000000+101
```

## Fixed (this PR)
```
$ lfortran format_49.f90
PASS: E format exponent correctly adjusted for 1.0d-200

$ lfortran a.f90
  0.10000000000000000+101
```

**Stage:** Runtime

## Changes
- `src/libasr/runtime/lfortran_intrinsics.c`:
  - Use decimal/integer_length instead of log10 for exponent calculation
  - Fix `abs(exponent_value >= 100)` to `abs(exponent_value) >= 100`
  - Adjust FIXED_CHARS_LENGTH when 'E' is dropped for 3-digit exponents
- `integration_tests/format_49.f90`: Integration test for 1.0d-200 exponent


tests/reference/run-format8-4e5a9c0.stdout
```diff
@@ -0,0 +1,5 @@
+  0.10000000000000000E-09
+  0.10000000000000000E-99
+  0.10000000000000000E+01
+  0.10000000000000000E+11
+ 0.10000000000000000+101
```
certik: Here is the behavior I get from GFortran and LFortran main:
```console
$ gfortran a.f90 && ./a.out
  0.10000000000000000E-09
  0.10000000000000000E-99
  0.10000000000000000E+01
  0.10000000000000000E+11
  0.10000000000000000+101
$ lfortran a.f90
  0.10000000000000000E-09
  0.10000000000000000E-99
  0.10000000000000000E+01
  0.10000000000000000E+11
 0.10000000000000000+101
```
Two observations:
* The behavior in main seems identical to this PR. Am I missing any difference? If not, then this is not a good test, it doesn't actually test the change in this PR.
* The last number is incorrectly indented by one column to the left. That's a bug both in main and in this PR.

krystophny: Thanks for the hint, now we got more extreme values that trigger it.

tests/reference/run-format_32-b407b58.stdout
```diff
@@ -7,15 +7,15 @@
...
 0.30000+201
  0.30000+201
+  0.30000+201
 0.33333E-200
 *********
 .33333-200
-0.33333-200
 0.33333-200
  0.33333-200
+  0.33333-200
```
certik, line 18: We need to verify that this change is correct. Can you check gfortran?

krystophny:
@certik Verified with gfortran. The change is correct.

```
$ gfortran format_32.f90 && ./a.out
...
0.33333E-200     <- E0.5
*********        <- E9.5
.33333-200       <- E10.5
0.33333-200      <- E11.5 (no leading space)
 0.33333-200     <- E12.5 (1 leading space)
  0.33333-200    <- E13.5 (2 leading spaces)
...
```

The PR fixes the indentation to match gfortran - before, E12.5 and E13.5 were each missing one leading space.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -699,7 +707,7 @@ void handle_decimal(char* format, double val, int scale, char** result, char* c,
     } else if (is_s_format && abs(exponent_value) >= 10) {
         int abs_exp = abs(exponent_value);
         exp = (abs_exp == 0) ? 2 : (int)log10(abs_exp) + 1;
-    } else if (abs(exponent_value >= 100)) {
+    } else if (abs(exponent_value) >= 100) {
```
certik, line 710: Good catch.

integration_tests/format_49.f90
```diff
@@ -0,0 +1,36 @@
...
+    read(output(exp_start:), '(I4)') exp_val
+
+    if (exp_val /= -200) then
+        print *, "FAIL: 1.0d-200 exponent should be -200, got ", exp_val
+        print *, "Output was: ", trim(output)
+        error stop
+    end if
+
+    print *, "PASS: E format exponent correctly adjusted for 1.0d-200"
+end program
```
certik, line 36: I verified that this fails in main.

Issue #9648, opened 2026-01-22
Implement 64-bit array descriptor


krystophny:
## Summary

Implement 64-bit array descriptor using the parameterized infrastructure.

## Changes

When `-fdescriptor-index-size=64` or `native` on 64-bit:

```
Dimension Descriptor:
┌─────────────────────────────────────┐
│ [0] stride       : i64              │
│ [1] lower_bound  : i64              │
│ [2] extent       : i64              │
└─────────────────────────────────────┘
```

## Key Implementation Points

1. Pass `i64` type to `SimpleCMODescriptor` constructor
2. Update all `CreateSExtOrTrunc` calls to handle i64 properly
3. Update `get_lower_bound`, `get_stride`, `get_dimension_size` return types
4. Ensure array indexing arithmetic uses i64

## Acceptance Criteria

- [ ] 64-bit descriptor works with `-fdescriptor-index-size=64`
- [ ] ILP64 mode (`-fdefault-integer-8`) works without type mismatches
- [ ] Large array indices (>2^31) work correctly
- [ ] All existing tests pass with default (32-bit)
- [ ] New integration tests for 64-bit mode

## Test Cases

```fortran
! Test large array bounds in ILP64 mode
program test_large_bounds
    implicit none
    integer, parameter :: n = 3000000000_8  ! > 2^31
    real, allocatable :: arr(:)
    allocate(arr(n))
    arr(n) = 42.0
    if (arr(n) /= 42.0) error stop
    print *, "PASS"
end program
```

## Part of

Flexible Array Descriptors meta-issue

krystophny:
Completed in PR #9656. 64-bit array descriptors implemented via --descriptor-index-64 flag.

Issue #9646, opened 2026-01-22
Refactor: Parameterize array descriptor index type


krystophny:
## Summary

Refactor `SimpleCMODescriptor` to use a configurable index type instead of hardcoded `i32`.

## Current State

```cpp
// llvm_array_utils.cpp line 102-109
dim_des(llvm::StructType::create(
    context,
    std::vector<llvm::Type*>(
        {llvm::Type::getInt32Ty(context),   // stride
         llvm::Type::getInt32Ty(context),   // lower_bound
         llvm::Type::getInt32Ty(context)}), // extent
         "dimension_descriptor")
)
```

## Proposed Change

Add an `index_type` member that defaults to `i32` but can be configured:

```cpp
class SimpleCMODescriptor : public Descriptor {
    llvm::Type* index_type;  // i32 or i64
    // ...
};
```

## Acceptance Criteria

- [ ] `SimpleCMODescriptor` accepts index type as constructor parameter
- [ ] Default remains `i32` (no behavior change)
- [ ] All dimension descriptor field accesses use the configurable type
- [ ] All existing tests pass
- [ ] No performance regression

## Scope

Files to modify:
- `src/libasr/codegen/llvm_array_utils.h`
- `src/libasr/codegen/llvm_array_utils.cpp`

## Part of

Flexible Array Descriptors meta-issue

krystophny:
Completed in PR #9656. SimpleCMODescriptor now has configurable index_type member.

PR #9656, opened 2026-01-22
feat: add --descriptor-index-64 flag and parameterize descriptor index type

krystophny:
## Summary

Add `--descriptor-index-64` flag and parameterize array descriptor index type for ILP64 support:
- Default: i32 descriptor indices
- With `--descriptor-index-64` or `-fdefault-integer-8`: i64 indices

Refs: #9646, #9648

## Why

Enables large array support for ILP64 mode and prepares for CFI descriptor interoperability. The independent flag allows future flexibility for users who need 64-bit descriptors without full ILP64 mode.

**Stage:** Codegen

## Changes
- [`utils.h#L144`](https://github.com/lfortran/lfortran/blob/836dd2e608bd725f6cd8df4e22c4aa62d4769bdf/src/libasr/utils.h#L144): Add `descriptor_index_64` option to CompilerOptions
- [`lfortran_command_line_parser.cpp#L185`](https://github.com/lfortran/lfortran/blob/836dd2e608bd725f6cd8df4e22c4aa62d4769bdf/src/bin/lfortran_command_line_parser.cpp#L185): Add `--descriptor-index-64` flag
- [`lfortran_command_line_parser.cpp#L329-L330`](https://github.com/lfortran/lfortran/blob/836dd2e608bd725f6cd8df4e22c4aa62d4769bdf/src/bin/lfortran_command_line_parser.cpp#L329-L330): Make `-fdefault-integer-8` imply `descriptor_index_64`
- [`llvm_array_utils.h#L371-L377`](https://github.com/lfortran/lfortran/blob/836dd2e608bd725f6cd8df4e22c4aa62d4769bdf/src/libasr/codegen/llvm_array_utils.h#L371-L377): Add index_type parameter to SimpleCMODescriptor
- [`llvm_array_utils.cpp#L89-L92`](https://github.com/lfortran/lfortran/blob/836dd2e608bd725f6cd8df4e22c4aa62d4769bdf/src/libasr/codegen/llvm_array_utils.cpp#L89-L92): Select index type based on `descriptor_index_64` option
- All descriptor operations updated to use parameterized index_type

## Tests
- All existing tests pass
- Tested both 32-bit (default) and 64-bit (`--descriptor-index-64` / `-fdefault-integer-8`) modes locally


certik:
@assem2002 or @jinangshah21 could you please review this PR?

src/libasr/asr_builder.h
```diff
@@ -1088,8 +1095,8 @@ class ASRBuilder {
         return ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(al, loc, m_a.p, m_a.n));
     }
 
-    #define UBound(arr, dim) PassUtils::get_bound(arr, dim, "ubound", al)
-    #define LBound(arr, dim) PassUtils::get_bound(arr, dim, "lbound", al)
+    #define UBound(arr, dim, ...) PassUtils::get_bound(arr, dim, "ubound", al, ##__VA_ARGS__)
+    #define LBound(arr, dim, ...) PassUtils::get_bound(arr, dim, "lbound", al, ##__VA_ARGS__)
```
certik, line 1099: I think this doesn't work for `UBound(arr, dim)`, as it will get translated to `PassUtils::get_bound(arr, dim, "ubound", al,)` which I think is a syntax error.

We should not be using macros at all.

PR #9658, opened 2026-01-22
fix: don't use ftruncate for stdout and stderr

### LINKED ISSUE #9638: Handle stdout to file which is appending
MRE:
```fortran
program mre_lfortran_ftruncate_final
    use, intrinsic :: iso_fortran_env, only: stdout => output_unit
    implicit none
    
    character(len=:), allocatable :: line
    ! integer :: i, n
    line = "This is a test line to append to the file."
    
    write(stdout,'(g0)') line
end program
```

Command: `lfortran a.f90 >> output.txt`

Output.txt: 

Image

jinangshah21:
Fixes #9638

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -7030,7 +7030,11 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const 
         if(iostat != NULL) *iostat = 0;
         va_end(args);
     }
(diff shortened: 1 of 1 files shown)


jinangshah21:
I don't know how to test it, as this is tested by `lfortran a.f90 >> output.txt` and we need to check content of output.txt.

certik:
You can test it in reference tests as `run = true` which saves stdout/stderr.

certik:
Ah I see -- append. Yeah, I wouldn't worry about it then.

PR #9661, opened 2026-01-22
fix: ieee_value returning uninitialized results

### LINKED ISSUE #8714: Undefined reference to `ieee_quiet_nan` when linking separate compilation units
An MRE is here: https://github.com/JorgeG94/mre_ieee_nan_lfortran- 

Simply: 
```
git clone git@github.com:JorgeG94/mre_ieee_nan_lfortran-.git
cd mre_ieee_nan_lfortran-
FC=lfortran make single !< this will work 
FC=lfortran make mre !< this won't 

lfortran -O2 -Wall -Wextra -c mre_nan.f90
lfortran -O2 -Wall -Wextra -c main.f90
lfortran  -o mre mre_nan.o main.o
/bin/ld: main.o: in function `__module_mre_nan_get_quiet_nan':
LFortran:(.text+0x64): undefined reference to `ieee_quiet_nan'
collect2: error: ld returned 1 exit status
The command 'gcc -o mre mre_nan.o main.o  -L"/home/jorge/install/miniconda3/envs/pic/bin/../share/lfortran/lib" -Wl,-rpath,"/home/jorge/install/miniconda3/envs/pic/bin/../share/lfortran/lib" -llfortran_runtime -lm' failed.
make: *** [Makefile:29: mre] Error 10

```

opixdown:
fixes #8714


certik:
There is probably faster / better way to do it, but for now this is ok, it fixes the bug. Let's run full tests.

opixdown:
clossing this pr, working with latest main

PR #9670, opened 2026-01-23
Minimal fix for a nested namelist

certik:


certik:
PRIMA fails with:
```
✔ Success: ./build/fortran/example_bobyqa_fortran_2_exe
→ Running: ./build/fortran/example_cobyla_fortran_1_exe
ci/test_third_party_codes.sh: line 28: 158521 Segmentation fault      (core dumped) ./$1
Error: Process completed with exit code 139.
```
I skipped it for now to see if the rest passes or also fails.

certik:
Here is the reproducer for PRIMA:
```fortran
module linalg_mod
use iso_fortran_env, only : real64
implicit none
private
public :: inprod, matprod12

contains

real(real64), intent(in) :: x(:)
real(real64), intent(in) :: y(:)
real(real64) :: z
integer :: i

print *, "K1"
z = 0.0_real64
print *, size(x), size(y)
do i = 1, int(size(x), kind(i))
    z = z + x(i) * y(i)
end do
print *, "K3"
end function inprod

function matprod12(x, y) result(z)
real(real64), intent(in) :: x(:)
real(real64), intent(in) :: y(:, :)
real(real64) :: z(size(y, 2))
integer :: j

print *, "G1"
do j = 1, int(size(y, 2), kind(j))
    print *, "G2"
    z(j) = inprod(x, y(:, j))
    print *, "G3"
end do
end function matprod12

end module linalg_mod


module caller_mod
use iso_fortran_env, only : real64
implicit none
private
public :: call_calcfc

abstract interface
    subroutine objcon(x, constr)
        import real64
        real(real64), intent(in) :: x(:)
        real(real64), intent(out) :: constr(:)
    end subroutine objcon
end interface

contains

subroutine call_calcfc(calcfc, x, constr)
procedure(objcon) :: calcfc
real(real64), intent(in) :: x(:)
real(real64), intent(out) :: constr(:)

call calcfc(x, constr)
end subroutine call_calcfc

end module caller_mod


program mre_inprod_bounds
use iso_fortran_env, only : real64
use linalg_mod, only : matprod12
use caller_mod, only : call_calcfc
implicit none

real(real64) :: amat(2, 1)
real(real64) :: bvec(1)
real(real64) :: x(2)
real(real64) :: constr(1)

amat = 1.0_real64
bvec = 0.0_real64
x = [1.0_real64, 2.0_real64]

 call call_calcfc(calcfc_internal, x, constr)
 print *, constr

contains

subroutine calcfc_internal(x_internal, constr_internal)
real(real64), intent(in) :: x_internal(:)
real(real64), intent(out) :: constr_internal(:)

constr_internal = matprod12(x_internal, amat) - bvec
end subroutine calcfc_internal

end program mre_inprod_bounds
```
Which does:
```console
$ gfortran -cpp mre_cobylb_reduced_procs.f90 && ./a.out
 G1
 G2
 K1
           2           2
 K3
 G3
   3.0000000000000000
$ echo $?
0
$ lfortran --cpp mre_cobylb_reduced_procs.f90
G1
G2
K1
2    1
$ echo $?
11
```

certik:
There is probably a way to simplify this and there seem to be more corner cases to handle:

• Findings

  - Medium: get_nested_namelist_symbol only pushes variables present in
    nested_var_to_ext_var, which is populated by capture_namelist_vars after
    skipping module variables; this drops module vars from the namelist used in
    nested scopes, so READ/WRITE (nml=...) can silently lose entries (see src/
    libasr/pass/nested_vars.cpp:282 and src/libasr/pass/nested_vars.cpp:676).
  - Low: maybe_replace_namelist always calls make_external_symbol with sym_name
    even if a same-name local symbol already exists in the current scope; if
    that ever happens, add_symbol will collide (see src/libasr/pass/
    nested_vars.cpp:698).

  Questions / Assumptions

  - Are namelist groups allowed to contain module variables that should still
    participate in nested READ/WRITE? If yes, the current filtering/rewiring
    likely needs to keep them.
  - Do you want to preserve the existing duplicate-name behavior for namelist
    group symbols, or should it mirror the “unique-name” handling used for other
    externals?

  Change Summary

  - nested_vars.cpp adds namelist capture/rewire logic, new helpers for external
    symbol creation/scope checks, and refactors call detection to be nested-
    scope aware.

  Simplification Ideas (to shorten the new code)

  1. Centralize external-symbol creation:
      - Add a helper like get_or_make_external_symbol(scope, target, owner_name,
        original_name, cache*) that:
          - checks for an existing external symbol that already points at
            target,
          - generates a unique name on collision,
          - uses make_external_symbol under the hood,
          - optionally caches in module_var_to_external.
      - Then use it in ReplacerNestedVars::replace_Var (src/libasr/pass/
        nested_vars.cpp:326), ReplaceNestedVisitor::visit_SubroutineCall (src/
        libasr/pass/nested_vars.cpp:764), and both creation sites in
        AssignNestedVars::transform_stmts (src/libasr/pass/nested_vars.cpp:924
        and src/libasr/pass/nested_vars.cpp:1120).
  2. Extract the repeated “import symbol into current scope” logic:
      - The sym_ externalization block (checking is_sym_in_scope_chain and
        creating an ExternalSymbol_t) appears twice in AssignNestedVars; moving
        it into a small helper would delete ~20–25 lines in each block.
  3. Namelist variable mapping helper:
      - If you decide to keep module vars in nested namelists, a
        map_namelist_vars(nml, nested_var_to_ext_var) helper can handle “use
        external if present, else keep original” in one place and shorten
        get_nested_namelist_symbol.

-------


But I think it's good enough to merge, it fixes several bugs and doesn't break anything, so let's merge it and then iterate upon it as needed.

PR #9671, opened 2026-01-23
fix: handle length calculation using expressions for allocatable strings

### LINKED ISSUE #8620: argument strings have issue with variable length
```.f90
program string_74
  character(:), allocatable :: str

  str = ff(3)
  print *, str

  contains 
  function ff(n) result(res)
    integer :: n
    character(n), allocatable :: res
    allocate(res)
    ! res = ff2()
    print *, len(res)
    res = "hi"
  end function
end program string_74
```

```bash
(lf) assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt3.f90 && ./a.out 
           3
 hi 
(lf) assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt3.f90 
0

(lf) assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran tt3.f90 --skip-pass=subroutine_from_function
3
hi 
```

Why it fails is due to the work of `subroutine_from_function` as it changes **ReturnVar** into argument. LFortran doesn't change the runtime value representing the length for string arguments making it always having the same length of the caller.

YashNagda17:
Fixes #8620 

Changes made:
- calculate length for `ExpressionLength` strings having allocatable attributes

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2025,6 +2025,7 @@ RUN(NAME string_91 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_92 LABELS gfortran llvm)
 RUN(NAME string_93 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_94 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc  EXTRA_ARGS --realloc-lhs-arrays)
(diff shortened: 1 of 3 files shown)


certik:
The change is simple, so I merged it. @assem2002 if you find issues, we can improve it later.

assem2002:
I think we should've used a stringview. it's the best to make the runtime length reflect the ASR length. I can figure out some test cases that could still fail, but that's fine for now.

PR #9674, opened 2026-01-23
Fix : fix regeressions presented in #9673

assem2002:
Based on top of #9673
- workaround in `array_passed_in_function_call` to bypass a case.
- workaround in the backend to handle arrayphysicalCast.

All workarounds should be removed once a proper ASR representation of unlimited-polymorphic types and array casting is done.


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9030,13 +9071,33 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                llvm::Type* target_type_past_array = llvm_utils->get_type_from_ttype_t_util(m_arg,
+                    ASRUtils::type_get_past_allocatable(
+                        ASRUtils::type_get_past_pointer(ASRUtils::extract_type(m_type))), module.get());
+                data_ptr = builder->CreateBitCast(data, target_type_past_array->getPointerTo());
+            } else {
+                data_ptr = llvm_utils->create_ptr_gep2(data_type,
+                        llvm_utils->CreateLoad2(data_type->getPointerTo(), 
+                            arr_descr->get_pointer_to_data(m_arg, m_type, tmp, module.get())),
+                        arr_descr->get_offset(arr_type, tmp));
+            }
```
certik, line 9092: This if/else statement seems to be duplicated 3x, right? If so, let's refactor into a function.

PR #9678, opened 2026-01-24
Fix string arrays with namelists

certik:
Adjusted namelist descriptor handling to normalize array descriptor pointers before GEPs and to extract string-array data via the descriptor path, avoiding the mismatched pointee type assert for allocatable character arrays.


certik:
Merged as part of https://github.com/lfortran/lfortran/pull/9682.

PR #9680, opened 2026-01-24
fix: use dynamic index type for descriptor field operations (cleanup)

krystophny:
## Summary

Replace hardcoded i32 types with `arr_descr->get_index_type()` in descriptor field operations for consistency when using 64-bit descriptor indices (`--descriptor-index-64`).

Refs: #9646, #9648

## Why

This is a consistency/cleanup fix for ILP64 support. When using 64-bit descriptor indices, the descriptor fields (bounds, strides, etc.) are i64, so operations should use the matching type rather than hardcoded i32.

**Stage:** Codegen

**Note:** This is a preventive fix. Basic functionality already works due to LLVM's automatic type coercion, but using the correct types is more robust.

## Changes

In `asr_to_llvm.cpp`:
- Array section lower bound constants use dynamic index type
- Array section upper bound computation uses dynamic index type
- Array section step constants use dynamic index type
- Dimension info for PointerArray uses dynamic index type
- ComplexIm offset uses dynamic index type


krystophny:
# Automatic Review

## Status: NEEDS WORK

### Checks
- [x] PR size: 20+/10-, 1 file - small
- [ ] **Test exists: NO**
- [ ] **Test fails on main: NOT VERIFIED - no test provided**
- [x] Pipeline stage: Codegen
- [x] No global variables
- [x] Code style consistent

### Clean Code Violation: YAGNI

PR description states:
> "This is a preventive fix. Basic functionality already works due to LLVM automatic type coercion"

**YAGNI (You Are Not Gonna Need It) Violation:**
- "Preventive" fixes without demonstrated failures are explicitly rejected per review policy
- If LLVM already handles this via automatic coercion, what actually breaks without this fix?
- Cannot verify the fix is needed or working without a test

**Question:** What breaks without this change?

### Questions for Author
1. Can you provide a test case that fails without this fix when using `--descriptor-index-64`?
2. If no failure can be demonstrated, what is the concrete benefit?
3. Is this addressing a real bug or theoretical correctness?

### Architecture Assessment
The changes are consistent - replacing `APInt(32, ...)` with dynamic `APInt(idx_bits, ...)`. However, without a failing test, this is unverified code change.

### Verdict
**HAND BACK** - Cannot approve without a test demonstrating the fix is needed. Either:
1. Add a test that fails on main without this fix, OR
2. Document why no test is possible and what concrete problem this solves

Per review policy (YAGNI): "Preventive fixes without demonstrated failures = REJECT. Ask: What breaks without this change?"

krystophny:
# Automatic Review Update

## Status: TEST CASE FOUND

### Reproducible Failure on Main

The following test fails on current main with `--descriptor-index-64 --realloc-lhs-arrays`:

```fortran
program test_array_section
    implicit none
    real, allocatable :: arr(:,:)
    real, allocatable :: section(:,:)
    integer :: i, j
    
    allocate(arr(4,4))
    do j = 1, 4
        do i = 1, 4
            arr(i, j) = real(i + j * 10)
        end do
    end do
    
    section = arr(2:3, 2:3)
    
    if (abs(section(1,1) - 22.0) > 0.001) error stop 1
    if (abs(section(2,1) - 23.0) > 0.001) error stop 2
    if (abs(section(1,2) - 32.0) > 0.001) error stop 3
    if (abs(section(2,2) - 33.0) > 0.001) error stop 4
    
    print *, "PASS"
end program
```

### Error on Main

```
  %183 = mul i32 1, i64 %182
  %190 = mul i32 %183, i64 %189
```

The hardcoded `APInt(32, 1)` creates an `i32` constant that gets multiplied with an `i64` value from the descriptor, causing LLVM verification failure.

### Updated Assessment

This PR **fixes a real LLVM verification error** when using 64-bit descriptor indices with array sections. It is NOT a "preventive" fix - it addresses an actual failure that can be reproduced.

### Remaining Work

1. PR branch needs rebase on current main (build0.sh changes caused API mismatch)
2. Add the test case above as an integration test (e.g., `array_section_11.f90`)
3. Register test with labels `gfortran llvm` and `EXTRA_ARGS --descriptor-index-64 --realloc-lhs-arrays`

krystophny:
## Review Feedback Addressed

### Changes Made

1. **Merged main into branch** (merge commit `e2dd8cb46`)
   - Branch now up-to-date with upstream/main

2. **Added integration test** (`e02482816`)
   - New file: `integration_tests/array_section_11.f90`
   - Tests array section assignment with `--descriptor-index-64`
   - Registered in `CMakeLists.txt` with labels `gfortran llvm` and `EXTRA_ARGS --descriptor-index-64`

### Test Result

```
1/1 Test #1826: array_section_11_LLVM_GOC ........   Passed    0.26 sec
```

### Note on Original Test Case

The original test case from the review used `--realloc-lhs-arrays` combined with `--descriptor-index-64`. During testing, I discovered this combination triggers a separate type mismatch issue (i32 * i64 in shape comparison) that is unrelated to this PR's scope. The test was modified to pre-allocate the target array, which isolates this PR's fix for array section descriptor fields while avoiding the separate realloc issue.

The core fix for hardcoded `APInt(32, ...)` being replaced with dynamic `arr_descr->get_index_type()` is validated by this test.

Ready for re-review.

certik:
The CI fails.

krystophny:
> The CI fails.

Passing again

certik:
Let us merge the fpm PRs first, to lock in the win.

Issue #9640, opened 2026-01-21
ILP64: LLVM verification error in array section bound computation with -fdefault-integer-8


krystophny:
## Summary

When using `-fdefault-integer-8` (ILP64 mode) with `--legacy-array-sections`, LLVM verification fails due to type mismatch in array section bound computation.

## Minimal Reproducer

```fortran
C     File: ilp64_mre.f
      SUBROUTINE CALLER(WORK, LWORK, N)
      INTEGER LWORK, N
      REAL WORK(LWORK)
C     Pass array element to implicit interface - creates array section
      CALL CALLEE(WORK(N+1), LWORK-N)
      END SUBROUTINE

      SUBROUTINE CALLEE(X, M)
      INTEGER M
      REAL X(M)
      X(1) = 1.0
      END SUBROUTINE
```

```bash
lfortran --fixed-form-infer --implicit-interface --implicit-typing \
         --legacy-array-sections --separate-compilation \
         -fdefault-integer-8 -c ilp64_mre.f
```

## Error

```
code generation error: asr_to_llvm: module failed verification. Error:
Both operands to a binary operator are not of the same type!
  %23 = sub i64 %22, i32 1
```

## Root Cause

In array section bound computation for implicit interface calls, the literal `1` in subtraction is typed as `i32` instead of `i64`:

```llvm
%20 = load i64, i64* %lwork      ; load 64-bit integer
%21 = add i64 %20, 1             ; add works (i64 + i64)
%22 = add i64 %21, 1             ; add works (i64 + i64)
%23 = sub i64 %22, i32 1         ; ERROR: i64 - i32 type mismatch
```

The `add` operations correctly use `i64` for the literal, but `sub` uses `i32`.

## Impact

This prevents building LAPACK with `BUILD_INDEX64=ON` (ILP64 mode). Affected files include `sgesvj.f`, `sgejsv.f`, and others with implicit interface array section passing patterns like:

```fortran
CALL SGSVJ0(..., WORK(N+1), LWORK-N, ...)
```

## Environment

- LFortran: 0.59.0 (upstream/main at 70a1917ba)
- LLVM: 11.1.0
- Platform: Linux x86_64

krystophny:
Related to #2844 (Add option `-fdefault-integer-8`). This is a remaining bug in the ILP64 implementation.

krystophny:
[...]

Issue #2844, opened 2023-11-10
Add option `-fdefault-integer-8`


meow464:
Possibly add other similar options. Lapack uses this options in its build system.
Where should I look at to implement this option?

certik:
It would be implemented in the frontend. This would be part of:
```console
$ lfortran -h
[...]
  -f TEXT ...            All `-f*` flags (only -fPIC supported for now)
[...]
```

meow464:
I was under the impression that kind selection happens during the asr lowering?
https://github.com/lfortran/lfortran/blob/f345516553a3b1c36485de9879bfbb146cda602f/src/libasr/ASR.asdl#L386

certik:
This would happen in the frontend: AST->ASR, which would choose `integer` to become kind=8 in ASR. No change will be needed in the lowering (ASR->LLVM).

meow464:
Is there any documentation on the lfortran internals? I'm walking through the code but I can't find where the lowering from AST integers to ASR integers happen. If I print the AST and ASR for a simple example I can see they are different but I can't get much further than that.
I found ASR::make_integer_t but I think that's for lowering to IR right?

certik:
Browse the documentation here: https://docs.lfortran.org/en/, and also you can watch this video which shows how the internals work: https://www.youtube.com/watch?v=yuYsyM08bss.

krystophny:
## ILP64 Status Update (2026-01-30)

ILP64 support (`-fdefault-integer-8`) is now substantially complete. LAPACK builds and all 92 tests pass with ILP64.

### Merged PRs
Core infrastructure and fixes:
- #9656 - `--descriptor-index-64` flag and infrastructure
- #9679 - ILP64 intrinsic return types (MAXLOC/MINLOC)
- #9684 - Array index temporaries
- #9729, #9731, #9732, #9733 - realloc-lhs, stop/error, equivalence, kind handling

### In-Flight Draft PRs
Final cleanup items pending merge:
- #9769 - Complex literal kind fix (fixes #9768)
- #9771 - Logical read for all kinds (fixes #9770)
- #9773 - File unit number conversion (fixes #9772)
- #9680 - Descriptor field operations (approved)
- #9681 - Array kind mismatch validation

### Test Status
- Regular build: 92 passed, 0 failed
- ILP64 build: 92 passed, 0 failed
- All LAPACK tests pass with `-fdefault-integer-8`

PR #9681, opened 2026-01-24
fix: reject integer array kind mismatch in implicit argument casting

krystophny:
## Summary

Reject integer array kind mismatch when using `--implicit-argument-casting`.

Part of the ILP64/LAPACK support effort (#2844, #9640).

## Why

When using `-fdefault-integer-8 --implicit-argument-casting`, scalar integers with different kinds can be cast implicitly. However, integer arrays with different kinds cannot be cast element-by-element - they have different memory layouts.

Previously passing an `integer(8)` array to an `integer(4)` parameter caused an assertion failure in `Call_t_body`. Now it produces a clear semantic error.

**Stage:** Semantics

## Changes

- `ast_body_visitor.cpp`: Refine `implicit_argument_casting` check to reject arrays of integers with different kinds (scalars still allowed)
- `ast_common_visitor.h`: Simplify type formatting in error messages
- `asr_utils.h`: Add `type_to_str_fortran_symbol` for kind-aware type formatting

## Tests

- `implicit_call_02.f90`: Verifies that passing integer(8) array to integer(4) parameter produces proper error (assertion failure on main, semantic error with fix)


krystophny:
# Automatic Review

## Status: PASSED

### Checks
- [x] CI status: GREEN (21/21 jobs passing)
- [x] PR size: 111+/29-, 21 files (3 source + 1 test + 17 reference updates)
- [x] No unrelated changes - reference updates due to improved error formatting
- [x] Test exists: `tests/errors/implicit_call_02.f90`
- [x] **Test fails on main**: VERIFIED - assertion failure:
  ```
  LCOMPILERS_ASSERT failed: asr_utils.h
  function Call_t_body(), line number 7070
  ASRUtils::check_equal_type(i64[:], i32[:])
  ```
- [x] Test passes with fix: CI green - produces semantic error instead
- [x] Pipeline stage: Semantics (correct - catches before `Call_t_body`)
- [x] No global variables
- [x] Code style consistent

### Architecture Assessment
Fix adds check in `ast_body_visitor.cpp` for `implicit_argument_casting`:
- Still skips type checks for most cases (intended behavior)
- BUT rejects arrays of integers with different kinds
- Reason: `Call_t_body` cannot handle `i64[:]` to `i32[:]` casting

Semantic error is better than assertion failure.

### Reference File Changes
17 reference files updated due to `type_to_str_fortran_symbol()` improvement:
- Shows `integer` instead of `integer(4)` for default kind
- Consistent formatting change

### Issues
None.

### Verdict
LGTM - Test verified to fail on main with assertion. Fix produces proper semantic error.

tests/reference/asr-array_constructor_with_different_kind-80c5bf4.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: Element in `real(8)` array constructor is `real(4)`
+semantic error: Element in `real(8)` array constructor is `real`
```
certik: I liked the previous error better: it makes it clear that the issue is with kind. Now it says `real(8)` vs `real`, but `real(8)` is also a `real`. It could possibly say `default real`, but let's say `real(4)` as it was instead, I think that was clear.

tests/reference/asr-implicit_call_02-00d0664.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type mismatch in argument `x`: expected `integer[:]` but got `integer(8)[:]`
```
certik: We need to obviously improve the array printing for types, but the same issue as above: let's use `integer(4)` vs `integer(8)` in these cases, will make it better.

certik:
Otherwise good. Also, do we need to add an option to skip these checks, I thought some old codes intentionally break these?

tests/reference/asr-array_constructor_with_different_kind-80c5bf4.stderr
krystophny: Done. Now shows `real(8) vs real(4)` for clarity.

tests/reference/asr-implicit_call_02-00d0664.stderr
krystophny: Done. Now shows `integer(4)[:] vs integer(8)[:]`.

krystophny:
Addressed the inline review feedback - error messages now show explicit kind specifiers:
- `real(8)` vs `real(4)` instead of `real(8)` vs `real`
- `integer(4)[:]` vs `integer(8)[:]` instead of `integer[:]` vs `integer(8)[:]`

About old codes I am not sure if this is a problem. I didn't notice it anywhere up to now and would address this in a follow up PR if it pops up.

tests/reference/asr-template_lapack_01-0d30f43.stdout
```diff
@@ -3426,10 +3426,10 @@
                                                     .false.
                                                     ()
                                                 ),
-                                            real:
+                                            real(4):
```
certik: This doesn't seem right. The variable name should not have parentheses in it.

tests/reference/llvm-unlimited_polymorphic_intrinsic_type_allocate-3e26601.stdout
```diff
@@ -166,7 +166,7 @@ declare i8* @_lfortran_malloc(i64)
 ; Function Attrs: argmemonly nounwind willreturn writeonly
 declare void @llvm.memset.p0i8.i32(i8* nocapture writeonly, i8, i32, i1 immarg) #0
 
-define void @_copy_integer_8(i8* %0, i8* %1) {
+define void @"_copy_integer(8)_8"(i8* %0, i8* %1) {
```
certik: I think this is not good, let's not use `()` in variable names.

tests/reference/run_dbg-runtime_stacktrace_01-c6f5e06.stderr
```diff
@@ -1,9 +1,2 @@
...
-    print *, main()
-  File "tests/errors/runtime_stacktrace_01.f90", line 10
-    print *, g()
-  File "tests/errors/runtime_stacktrace_01.f90", line 20
-    call f()
-  File "tests/errors/runtime_stacktrace_01.f90", line 14
-    stop 1
-STOP 1
+sh: llvm-dwarfdump: command not found
+Error in creating the files used to generate the debug information. This might be caused because either `llvm-dwarfdump` or `Python` are not available. Please activate the CONDA environment and compile again.
```
certik: You might need to install the conda environment --- it will have `llvm-dwarfdump` in it.

tests/reference/asr-flush2-9767ece.stdout
```diff
@@ -40,6 +40,7 @@
                         ()
                         ()
                         ()
+                        ()
```
certik, line 43: This is concerning, as ASR.asdl was not modified, so this change shouldn't be there?

krystophny:
@certik all should be fine now. I merged main once more to be sure.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5819,17 +5819,32 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                        bool skip_check = compiler_options.implicit_argument_casting ||
-                                            ASRUtils::is_class_type(ASRUtils::type_get_past_array(passed_type)) ||
+                        // Skip type checking for polymorphic types (class), function types
+                        bool skip_check = ASRUtils::is_class_type(ASRUtils::type_get_past_array(passed_type)) ||
                                             ASRUtils::is_class_type(ASRUtils::type_get_past_array(param_type)) ||
                                             ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(passed_type)) ||
                                             ASR::is_a<ASR::FunctionType_t>(*ASRUtils::type_get_past_array(param_type));
+                        // For implicit_argument_casting, skip type checking in most cases
+                        // EXCEPT for arrays where both are integers but with different kinds -
+                        // Call_t_body cannot handle this case and would assert
```
certik, line 5829: Qeustion: is this a Fortran restriction or LFortran limitation? If this is a Fortran requirement, then don't mention `Call_t_body`. If this is LFortran limitation then this should be mentioned in the error message.

certik (approved):
I think this change is fine. But wait until fpm PRs are in.

PR #9684, opened 2026-01-25
Fix integer types for ArrayUBound in ILP64 mode

krystophny:
## Summary

Fix integer type mismatches in ILP64 mode (`-fdefault-integer-8`) for legacy array section operations.

Part of ILP64 support for LFortran/LAPACK. Tracked in issue #9640.

## Why

Legacy array sections used hardcoded `Integer(4)` constants for array bound operations, causing LLVM verification errors when `-fdefault-integer-8` is used:

```
  %23 = sub i64 %22, i32 1
```

**Stage:** Semantics

## Changes

- `src/lfortran/semantics/ast_common_visitor.h`: Pass `default_integer_kind` to `ArrayUBound` calls and replace `b.i32()` with `b.i_t()` in legacy array section code
- `src/lfortran/semantics/ast_body_visitor.cpp`: Pass `default_integer_kind` parameter to visitor classes
- `src/libasr/asr_utils.h`: Add `index_kind` parameter to `get_bound` template

## Tests

- `legacy_array_sections_15.f90`: Legacy array sections with `-fdefault-integer-8` (verified: fails on main, passes with fix)

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2820,6 +2820,9 @@ RUN(NAME legacy_array_sections_13 LABELS gfortran llvm EXTRA_ARGS --legacy-array
 RUN(NAME legacy_array_sections_14 LABELS gfortran llvm
     EXTRAFILES legacy_array_sections_14_foo.f90
     EXTRA_ARGS --legacy-array-sections --implicit-interface --separate-compilation)
+RUN(NAME legacy_array_sections_15 LABELS gfortran llvm
+    EXTRA_ARGS --legacy-array-sections -fdefault-integer-8
+    GFORTRAN_ARGS -fdefault-integer-8)
 
 RUN(NAME cmake_minimal_test_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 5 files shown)


krystophny:
# Automatic Review

## Status: PASSED

### Checks
- [x] PR size: 61+/29-, 5 files - atomic change
- [x] Test exists: `legacy_array_sections_15.f90`
- [x] **Test fails on main**: VERIFIED - LLVM type mismatch:
  ```
    %23 = sub i64 %22, i32 1
  ```
- [x] Test passes with fix: CI green
- [x] Pipeline stage: Semantics (correct - not codegen workaround)
- [x] No global variables
- [x] Code style consistent

### Architecture Assessment
Fix passes `default_integer_kind` through semantics layer:
- `replace_ArrayItem_in_SubroutineCall()` now takes `default_integer_kind`
- `get_bound()` template has `index_kind` parameter (default=4)
- Uses `b.i_t(value, int_type)` instead of hardcoded `b.i32(value)`

Correct approach: fix type at ASR construction, not codegen.

### Issues
None.

### Verdict
LGTM - Test verified to fail on main. Semantics-level fix is correct.

certik (approved):
I think that looks good.

Issue #9526, opened 2026-01-15
Figure out an API for runtime errors similar to compile time errors


certik:
The API should allow multiple messages, primary and secondary labels, etc.

PR #9686, opened 2026-01-25
feat: new runtime error API

xaerru:
Fix https://github.com/lfortran/lfortran/issues/9526

I've modified one call of `llvm_utils->generate_runtime_error()` to the new api call `llvm_utils->generate_runtime_error2()` for demonstration. In some cases the location info is wrong, that needs fixing.

### DIFF
--- a/tests/reference/run-array_bounds_check_04-2e3d8fb.stderr
+++ b/tests/reference/run-array_bounds_check_04-2e3d8fb.stderr
@@ -2,4 +2,7 @@ runtime error: Array shape mismatch in subroutine 'my'. Tried to match size 4 of
  --> tests/errors/array_bounds_check_04.f90:5:13
   |
 5 |     call my(x)
-  |             ^^ 
+  |             ^
+   |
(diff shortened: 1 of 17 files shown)


tests/reference/run-array_bounds_check_04-2e3d8fb.stderr
```diff
@@ -2,4 +2,7 @@ runtime error: Array shape mismatch in subroutine 'my'. Tried to match size 4 of
  --> tests/errors/array_bounds_check_04.f90:5:13
   |
 5 |     call my(x)
-  |             ^^ 
+  |             ^
+   |
+10 |             integer, intent(in) :: x(3, 1)
+   |                                         ~ Expected size
```
certik, line 8: As a user it's confusing what the "Expected size" means. But that can be changed later. I am now realizing that the string will be runtime, it will say something like "Expected range 1:4, got 8", so we'll have to do the formatting at runtime for labels also. But that can be improved later.

certik:
Let's run all tests.

PR #9695, opened 2026-01-26
feat: Add support for sign specifier in OPEN statement

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6349


integration_tests/file_open_06.f90
```diff
@@ -0,0 +1,10 @@
+program open_sign_01
+    implicit none
+    
+    ! Test sign='plus' - should show plus sign for positive numbers
+    open(6, sign='plus')
+    print '(F5.1)', 1.0
+    
+    ! Don't close unit 6 (stdout) as subsequent prints would have no output
+    print *, "PASS"
```
certik, line 9: Can you add some tests for this --- maybe a run test?

I think currently the fix in this PR is not tested, right?

certik:
Add a reference test for the test above --- right now the runtime I think is not tested at all, so we don't know that the changes actually work.

certik:
Once you add a test, please click on Ready for review.

PR #9697, opened 2026-01-26
fix: throw error for unequal string lengths in merge intrinsic

### LINKED ISSUE #9696: Throw error in `merge` intrinsic even when any argument is not compile time constant
MRE
```fortran
program test_inquire_fail
  implicit none
  integer :: ios
  character(len=200) :: nm

  print *, "iostat =", ios, merge("fail caught", "no fail", ios /= 0)
end program
```

GFortran
```console
(lf) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ gfortran a.f90 
a.f90:6:34:

    6 |   print *, "iostat =", ios, merge("fail caught", "no fail", ios /= 0)
      |                                  1
Error: Unequal character lengths (11/7) in MERGE intrinsic at (1)
```

HarshilShah1804:
Fixes #9696

### DIFF
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -202,7 +202,7 @@ program continue_compilation_3
     print *, "8356" .eqv. 8356.00
     print *, ['c', 'o', 'd', 'e'] .or. ['m', 'a', 's']
     print *, ["welcome", "to", "lf"] .and. "contributors"  !even size diff of array element must be caught
-
+    print *, merge("fail caught", "no fail", merge_i /= 0)
     contains 
     subroutine bpe()
         print *, size(bpe)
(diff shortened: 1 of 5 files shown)


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -749,12 +749,6 @@
             "kind_arg": True
         }
     ],
-    "Merge": [
-        {
-            "args": [("any", "any", "bool")],
-            "ret_type_arg_idx": 0
-        }
-    ],
```
HarshilShah1804, line 757: @certik Regarding removing merge from the auto-generation script, what are your thoughts? Should we modify the Python script to include length checking, or would it be better to handle this case manually?

Since merge is the only function that requires length checking, modifying the Python script might introduce unnecessary overhead compared to treating this case separately.

PR #9703, opened 2026-01-26
CI: test FPM on CI with minimal workarounds

jinangshah21:
Testing with minimal workarounds(2) needed for current main: 1 namelist test + MCLI2 character array issue


certik:
Looks like this will pass. The LLVM 12 is just a network error.

certik:
I merged it: the LLVM 12 is a network error and the quick check that is still queued and didn't run yet doesn't test third party anyway.

Issue #9692, opened 2026-01-26
Change default printing back to print as 0.14, not 1.4e-01.


certik:
```console
$ cat b.f90
print *, 0.14
end
$ gfortran b.f90 && ./a.out
  0.140000001
$ lfortran b.f90
1.40000001e-01
```

This applies to both `print *, 0.14` and `write(*,*) 0.14`. I think it's more readable and also compatible with other compilers.

PR #9714, opened 2026-01-27
fix: print list-directed real/complex output with correct format

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9692


integration_tests/print_09.f90
```diff
@@ -0,0 +1,43 @@
...
+    print *, c
+    if (abs(c - 0.0001d0) > 1.0d-10) error stop "c value incorrect"
+    
+    ! Test write(*,*)
+    print *, "Testing write(*,*) format:"
+    write(*,*) x
+    write(*,*) y
+    write(*,*) a, b
+    
+    print *, "All tests passed"
```
certik: This file seems too long. Why not write it like this:
```fortran
program print_09
print *, 0.14, 123.456, 0.001, 1.5e-5
print *, 0.14_dp, 456.789_dp, 0.0001_dp
write(*,*) 0.14, 123.456, 0.001, 1.5e-5
write(*,*) 0.14_dp, 456.789_dp, 0.0001_dp
end program

File: src/libasr/runtime/lfortran_intrinsics.c [hunk 2]
```diff
@@ -362,17 +362,20 @@ void handle_logical(char* format, bool val, char** result) {
     }
 }
 
+static void format_float_fortran(char* result, float val);
+static void format_double_fortran(char* result, double val);
+
 void handle_float(char* format, double val, int scale, char** result, bool use_sign_plus, char rounding_mode) {
     val = val * pow(10, scale); // scale the value
     if (strcmp(format,"f-64") == 0) { //use c formatting.
```
certik, line 373: Later: we should not be using string comparison, but rather enums to distinguish `f-64` and `f-32`, which is faster and less error prone.

integration_tests/print_09.f90
certik: This can be changed after this PR is merged.

certik:
Let's run full tests. If they pass, we can merge this as is.

HarshitaKalani:
>The changes in reference results are all correct?

Yes, they were wrong earlier, now they're all correct.

integration_tests/print_09.f90
HarshitaKalani: Yes, I'll do that.

src/libasr/runtime/lfortran_intrinsics.c
HarshitaKalani: Let me open an issue for this.

HarshitaKalani: Done at https://github.com/lfortran/lfortran/issues/9717

certik:
We are getting:
```
1553/1988 Test #1549: sin_02 .............................................***Failed    0.36 sec
10.500000000000000    -0.87969575997167027    -0.87969575997167004
Error: failed to run main module `/home/runner/work/lfortran/lfortran/integration_tests/test-llvm_wasm/sin_02`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
           0:  0x3d8 - <unknown>!signature_mismatch:abs
           1: 0x2a50 - <unknown>!format_double_fortran
           2: 0xe164 - <unknown>!print_into_string
           3: 0xee53 - <unknown>!default_formatting
           4: 0xf641 - <unknown>!_lcompilers_string_format_fortran
           5:  0x541 - <unknown>!main
           6:  0x99f - <unknown>!_start
           7: 0x21463 - <unknown>!_start.command_export
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: wasm trap: wasm `unreachable` instruction executed
```
I think it would be worth investigating this failure in WASM. The stacktrace shows it is running the new code from this PR and then the line `signature_mismatch:abs` seems concerning, as well as `wasm trap: wasm `unreachable` instruction executed`. Might be a bug in the C code, or possibly it's showing some runtime error?

HarshitaKalani:
With the current change, I’ve replaced the existing occurrences of abs() with conditional expressions. This seems to work for now, but another possible approach would be to add proper abs support for WASM, which I will need to explore.

PR #9718, opened 2026-01-27
Extend the namelist test to be more comprehensive

certik:
This checks that we can parse and read more than one value. No compiler changes needed, all features have already been implemented, but it was not tested.

The new test now prints:
```console
$ lfortran integration_tests/namelist_25.f90
a =    42
b =    100
c =    -7
x =    3.14000010e+00
name = "Hello"
All tests passed!
```

### DIFF
--- a/integration_tests/namelist_25.f90
+++ b/integration_tests/namelist_25.f90
@@ -1,20 +1,41 @@
 program namelist_internal_read_array
     implicit none
 
-    integer :: a
-    character(len=20) :: test
-    character(len=256) :: readme(3)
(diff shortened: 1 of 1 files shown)


certik:
@jinangshah21 I think this might be a better test, just in case. It's not testing the `readme(2) = ' ' // test` string concatenation anymore, but I don't think that was an issue in the bug.

I don't know if we need to test both a single member namelist and multiple members, or just multiple members is enough, as in this PR. I think it's probably enough, but let me know.

PR #9733, opened 2026-01-28
fix: cast KIND argument to int32 in CMPLX for ILP64 mode

krystophny:
## Summary
- Cast KIND argument to int32 in CMPLX intrinsic for ILP64 compatibility

## Why
In ILP64 mode (`-fdefault-integer-8`), `INTEGER, PARAMETER :: WP = real32` becomes `INTEGER(8), PARAMETER :: WP = 4`. When `KIND=WP` is passed to intrinsics like `CMPLX(r, z, KIND=WP)`, the argument is `i64` but the intrinsic expects `i32`, causing a type mismatch assertion.

**Stage:** Intrinsic function registry (semantic analysis)

## Changes
- [`src/libasr/pass/intrinsic_functions.h`](https://github.com/lfortran/lfortran/blob/fix/ilp64-kind-arg/src/libasr/pass/intrinsic_functions.h#L4328-L4334): Cast KIND argument to int32 if its integer kind is larger than 4

## Tests
- [`integration_tests/ilp64_kind_arg_01.f90`](https://github.com/lfortran/lfortran/blob/fix/ilp64-kind-arg/integration_tests/ilp64_kind_arg_01.f90): Tests CMPLX with KIND parameter from INTEGER PARAMETER under ILP64 mode

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3169,3 +3169,6 @@ RUN(NAME intent_out_allocatable_component_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_struct_member_no_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_module_dealloc LABELS gfortran llvm)
 RUN(NAME array_shape_func_call LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+
+# ILP64 KIND argument bug - intrinsics with KIND parameters that are INTEGER PARAMETER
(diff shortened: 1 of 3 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -3169,3 +3169,6 @@ RUN(NAME intent_out_allocatable_component_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_struct_member_no_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_module_dealloc LABELS gfortran llvm)
 RUN(NAME array_shape_func_call LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+
+# ILP64 KIND argument bug - intrinsics with KIND parameters that are INTEGER PARAMETER
```
certik, line 3173: No need for such comments I think.

Issue #2293, opened 2023-08-30
SIMD backend


certik:
```diff
diff --git a/src/libasr/ASR.asdl b/src/libasr/ASR.asdl
index 26e60e172..d6a29ecef 100644
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -420,6 +420,7 @@ array_physical_type
     = DescriptorArray
     | PointerToDataArray
     | FixedSizeArray
+    | SIMDArray
     | NumPyArray
     | ISODescriptorArray
```
We'll use:
```
from lpython import SIMD
x: Annotated[f32[64], SIMD]
```
In ASR we use `SIMDArray` physical type, and then in the LLVM backend (or ASR->ASR pass) we ensure all such arrays get vectorized, otherwise we give a compile time error message. The conditions are:

* Must be 1D array
* Array element type and all operations must directly map into hardware instructions. For example on Apple M1 CPU, f32 plus and multiply is supported (it will compile), but f16 multiply is not (compile time error on Apple M1)
* Fixed compile time size
* The size must be a multiple of the hardware vector length. If we have 512 bits (AVX-512), the sizeof(element)*size must be equal to 512*n for n=1, 2, 3, .... If n=1, then the array is directly stored in a register. For n>2, the loop is unrolled (since the size is known at compile time), ensuring we hit maximum compute throughput (hide IO and latency).

dacarnazzola:
Extensions!

certik:
I think we need both an extension and a pragma syntax. For pragma, probably this is the way to go:
```fortran
!LF$ attributes simd :: x
real(sp) :: x(8)
```
I asked at https://fortran-lang.discourse.group/t/fortran-syntax-for-pragmas/6532 for feedback.

certik:
With #2452 this now runs: `lfortran --show-c integration_tests/matmul_01.f90`.

Now we have several parallel efforts:
* [ ] Get `matmul_01` working with the C backend
* [ ] Get the `simd` feature working with the C backend
* [ ] Get `matmul_01` working with the LLVM backend
* [ ] Get the `simd` feature working with the LLVM backend

Ordered roughly by the order of priority.

rebcabin:
> Extensions!

Doesn't the question of syntax extensions depend on whether you want code to be compilable by gfortran and flang?

rebcabin:
> Extensions!

[...]

PR #9735, opened 2026-01-28
enh: Add semantic validation for SIMD arrays

### LINKED ISSUE #9734: Complete TODO's for SIMD array semantic validation and testing
In `ast_symboltable_visitor.cpp,` the function `process_simd_variables()` contains an open TODO to fully validate SIMD array 
requirements:

```
// TODO: check all the SIMD requirements here:
// * 1D array
// * the right, compile time, size, compatible type
// * Not allocatable, or pointer
```

opixdown:
fixes #9734 
added checks as proposed in https://github.com/lfortran/lfortran/issues/2293#issuecomment-3805649675

## Enhancements
Added semantic validation :
	•	Must be an array
	•	Must be 1D
	•	Must have compile time constant size 
	•	Must not be allocatable or pointer
	•	Must have real or integer element type


opixdown:
Should these tests go into continue_compilation_1?

certik:
The LLVM files are huge, for example `tests/reference/llvm-matmul_01-90b582d.stdout` has 17K lines. Let's not commit such large files.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1948,23 +1948,77 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                    !ASR::is_a<ASR::Integer_t>(*elem_type)) {
+                    diag.add(diag::Diagnostic(
+                        "SIMD arrays must have Real or Integer element type, but `" +
+                        var.first + "` has an incompatible type",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {t->base.loc})}));
+                    throw SemanticAbort();
+                }
+                // Mark as SIMD array
+                a->m_physical_type = ASR::array_physical_typeType::SIMDArray;
```
certik, line 2023: All these should be tested. Yes, continue compilation is fine.

opixdown: I tried adding a test cases in continue_compilation_1, but the expected error was not triggered. should i move these  tests to tests/errors/?

certik: If the error was not triggered, then there is a bug in the logic that we need to fix.

opixdown: yes i checked, the tests are causing ice with the --continue-compilation flag once fixed i will add the test.

tests/tests.toml
```diff
@@ -4115,6 +4119,14 @@ c = true
...
 asr = true
 
+[[test]]
+filename = "../integration_tests/simd_01.f90"
+asr = true
+
+[[test]]
+filename = "../integration_tests/simd_02.f90"
+asr = true
+
```
certik: Do we need these? If not, I would remove them for now, to save resources.

tests/reference/asr-continue_compilation_simd-4dab78a.stderr
```diff
@@ -0,0 +1,29 @@
...
+ --> tests/errors/continue_compilation_simd.f90:8:5
+  |
+8 |     real :: simd_scalar
+  |     ^^^^^^^^^^^^^^^^^^^ 
+
+semantic error: SIMD arrays must have Real or Integer element type, but `simd_char` has an incompatible type
+ --> tests/errors/continue_compilation_simd.f90:9:5
+  |
+9 |     character :: simd_char(8)
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
```
certik, line 29: This looks great. Once can further iterate on these errors later, such as only underlying the offensive variable, e.g., in `real, simd :: bad, good(8)`. It should not underline the whole thing, since it will not be clear which one is bad. But that can be done in subsequent PRs.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1948,26 +1948,86 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                            diag::Label("", {t->base.loc})}));
+                    if ( !compiler_options.continue_compilation ) throw SemanticAbort();
+                    continue;
+                }
+                // array
+                if (!ASR::is_a<ASR::Array_t>(*t)) {
                     diag.add(diag::Diagnostic(
                         "The SIMD variable `" + var.first + "` must be an array",
                         diag::Level::Error, diag::Stage::Semantic, {
                             diag::Label("", {t->base.loc})}));
```
certik, line 1974: This error is not tested yet, correct? Can be fixed later.

certik:
Let me know on https://github.com/lfortran/lfortran/pull/9735#discussion_r2751870105.

tests/tests.toml
opixdown: we have one test which was already there to verify the path of simdarray so we can remove these

tests/reference/asr-continue_compilation_simd-4dab78a.stderr
opixdown: Yes we can open a issue for that and fix it in subsequent pr

opixdown:
commited all the suggestes changes i think it is good now.

certik:
Perfect, thanks. Let's run all tests to ensure we didn't break anything.

certik:
Thank you. Go ahead and fix the remaining comments in subsequent PRs.

PR #9736, opened 2026-01-28
Fix a struct initialization issue

### LINKED ISSUE #9730: Handle array members of extended type passed to subroutine as polymorphic arg
MRE:
```fortran
program mre_namelist
    type :: base
    end type base

        character(len=20), allocatable :: names(:)
    end type derived

    class(base), allocatable :: obj

    call get_command_line_settings(obj)

    select type(obj)
        type is(derived)
            print *, size(obj%names)
            print *, obj%names
    end select

contains 

     subroutine get_command_line_settings(cmd_settings)
        class(base), allocatable, intent(out) :: cmd_settings
        character(len=20), allocatable :: global_names(:)
        allocate(character(len=20) :: global_names(2))
        global_names = ['global_name1', 'global_name2']
        allocate(derived :: cmd_settings)
        cmd_settings = derived(names = global_names)
    end subroutine get_command_line_settings

end program mre_namelist
```

Outputs:
```console
           2
 global_name1        global_name2        
-1102163292
runtime error: Argument 1 is unallocated.
  --> d.f90:16:22
   |
16 |             print *, obj%names
   |  
```

certik:
Root Cause: Three issues combined to cause the runtime error:

1. src/libasr/pass/pass_utils.h (line 743): The structure constructor pass decomposed cmd_settings = derived(names = global_names) into member assignments to a temp variable, but didn't set realloc_lhs for allocatable array members. The temp var's allocatable array member was initialized with size 1, and the assignment tried to write size-2 data into it without reallocating.
2. src/libasr/pass/array_op.cpp (line 967/1164): The insert_realloc_for_target function only checked the global realloc_lhs_arrays compiler option, ignoring the per-assignment m_realloc_lhs flag. So even when the structure constructor pass set m_realloc_lhs = true on member assignments, the ReAlloc statement wasn't being generated.
3. src/libasr/codegen/llvm_utils.cpp (line 3052): The deepcopy for allocatable character arrays in struct copies had a TODO and just did a shallow pointer copy (CreateStore). This caused dangling pointers when the source temp variable was destroyed after the function returned.

Fixes:
- pass_utils.h: Force realloc_lhs = true for assignments where the target member is allocatable
- array_op.cpp: Pass the per-assignment m_realloc_lhs flag to insert_realloc_for_target so it generates the ReAlloc statement
- llvm_utils.cpp: Implement proper deep copy for allocatable character arrays — heap-allocate a new string_descriptor, allocate and copy the character data,
 and copy dimension descriptors

Test: Added integration_tests/class_allocate_03.f90 registered with gfortran and llvm labels.

Fixes #9730.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2274,6 +2274,7 @@ RUN(NAME derived_type_member_procedure_call_01 LABELS gfortran llvm)
 
 RUN(NAME class_allocate_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME class_allocate_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME class_allocate_03 LABELS gfortran llvm)
 
 RUN(NAME kwargs_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME kwargs_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/class_allocate_03.f90
+++ b/integration_tests/class_allocate_03.f90
@@ -1,0 +1,33 @@
+program class_allocate_03
+    implicit none
+
+    type :: base
+    end type base
+
+    type, extends(base) :: derived
+        character(len=20), allocatable :: names(:)
+    end type derived
+
+    class(base), allocatable :: obj
+
+    call get_command_line_settings(obj)
+
+    select type(obj)
+        type is(derived)
+            if (size(obj%names) /= 2) error stop
+            if (obj%names(1) /= 'global_name1') error stop
+            if (obj%names(2) /= 'global_name2') error stop
+    end select
+
+contains
+
+     subroutine get_command_line_settings(cmd_settings)
+        class(base), allocatable, intent(out) :: cmd_settings
+        character(len=20), allocatable :: global_names(:)
+        allocate(character(len=20) :: global_names(2))
+        global_names = ['global_name1', 'global_name2']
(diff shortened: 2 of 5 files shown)


certik:
Note about `--realloc-lhs-arrays` and if we should allocate the allocatable array inside the struct automatically even without `--realloc-lhs-arrays` (as this PR does):

- `--realloc-lhs-arrays` controls direct array assignment: `my_alloc_array = expr`.
  Without the flag, the array must already be allocated with matching shape.
  This is the only thing the flag governs.
- Structure constructor initialization is unconditional: `derived(names = arr)`
  constructs a new object. Its allocatable components must be allocated to hold
  the provided values — this is construction semantics, not assignment to an
  existing array. The standard (F2008 4.5.2.4 / 7.2.1.5) treats this as
  initialization of a freshly-constructed object.
- Derived type assignment with allocatable components is also unconditional:
  When you do `a = b` where both are derived types with allocatable components,
  the standard (F2008 7.2.1.3) requires the allocatable components of `a` to be
  (re)allocated to match `b`. This is intrinsic to derived-type assignment
  semantics — it's not an "array assignment" in the sense the flag controls.

So the decomposition in the class_constructor pass (`temp%names = global_names`)
   is really an implementation artifact of lowering a structure constructor. The
   semantic intent is "construct an object with names initialized from
  global_names", which always allocates. Forcing `realloc_lhs = true` on those
  generated member assignments is the correct way to express that intent through
   the existing infrastructure.


There is one corner case to consider: if the LHS is not allocatable, but has allocatable members, then maybe we should not automatically reallocate them without `--realloc-lhs-arrays`; I created a dedicated issue for this https://github.com/lfortran/lfortran/issues/9738.

Issue #6346, opened 2025-02-18
USE statement with a renamed variable misused


harperjf:
Gfortran had a bug with this program in 2007 but not now. Lfortran has the bug now. The reason why it's a bug is subtle, but explained in the comments after the ```END PROGRAM``` statement. I wrote them after discussion with various gfortran developers, who initially disagreed on whether it was a bug.
```

MODULE xmod! gfortran bug report 33541 Sep 2007
  REAL(kind(1d0)) :: x = -666
END MODULE xmod

PROGRAM test2uses
  USE xmod
  USE xmod, ONLY: xrenamed => x ! kind(1d0)
  x = 666                       ! kind(1.0): see below
  PRINT *,'kind(xrenamed)==kind(x)?',kind(xrenamed)==kind(x)
  PRINT *,'That should have printed',.FALSE.
END PROGRAM test2uses

! X appears as a use-name in the only-rename for the second USE
! statement, therefore X in module XMOD is inaccessible in the main
! program under the name X, so the X that = 666 there is default real.
```
Output from gfortran 14.2.0
```
 kind(xrenamed)==kind(x)? F
 That should have printed F
```
Output from lfortran 0.45.0 --std=f23
```
 kind(xrenamed)==kind(x)?    T
 That should have printed    F

PR #9739, opened 2026-01-28
fix: make original symbol inaccessible when renamed in USE ONLY statement

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6346

### DIFF
--- a/integration_tests/modules_64.f90
+++ b/integration_tests/modules_64.f90
@@ -1,0 +1,15 @@
+MODULE xmod
+  REAL(kind(1d0)) :: x = -666
+END MODULE xmod
+
+PROGRAM modules_64
+  USE xmod                      
+  USE xmod, ONLY: xrenamed => x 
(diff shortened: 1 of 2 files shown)


integration_tests/modules_64.f90
```diff
@@ -0,0 +1,15 @@
...
+PROGRAM modules_64
+  USE xmod                      
+  USE xmod, ONLY: xrenamed => x 
+  REAL :: x                     
+  
+  x = 666                       
+  
+  if (kind(xrenamed) == kind(x)) error stop
+  if (kind(xrenamed) /= 8) error stop
+  if (kind(x) /= 4) error stop
```
certik, line 14: Also check the value of `xrenamed` and `x` here.

certik: https://github.com/lfortran/lfortran/issues/9752

PR #9743, opened 2026-01-28
Fix a bug in parsing `,` in namelists

certik:
Depends on https://github.com/lfortran/lfortran/pull/9742.


certik:
Should be rebased before merging.

PR #9746, opened 2026-01-28
Implement stringification in cpp

### LINKED ISSUE #9745: Support default stringification in cpp
```fortran
#ifdef __GFORTRAN__
# define STRINGIFY_START(X) "&
# define STRINGIFY_END(X) &X"
#else /* default stringification */
# define STRINGIFY_(X) #X
# define STRINGIFY_START(X) &
# define STRINGIFY_END(X) STRINGIFY_(X)
#endif

program test
implicit none
character (len=:), allocatable :: astring
astring = STRINGIFY_START(MYMACRO)
STRINGIFY_END(MYMACRO)
write (*, *) astring
end program test
```
This works with ifx and flang. GFortran requires the first ifdef. LFortran works with the first ifdef, but we should really support the standard stringification like ifx and flang. Currently:
```console
$ lfortran --cpp a.f90
warning: Unsupported macro
  --> a.f90:14:1
   |
14 | STRINGIFY_END(MYMACRO)
   | ^^^^^^^^^^^^^^^^^^^^^^^ Ignored

  --> a.f90:15:9
   |
15 | write (*, *) astring
   |         ^
```

certik:
Fixes #9745.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2769,6 +2769,9 @@ RUN(NAME cpp_pre_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c wasm
 RUN(NAME cpp_pre_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c wasm
     EXTRA_ARGS --cpp
     GFORTRAN_ARGS -cpp)
+RUN(NAME cpp_pre_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c wasm
+    EXTRA_ARGS --cpp
+    GFORTRAN_ARGS -cpp)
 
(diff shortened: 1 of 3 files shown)


certik:
@jinangshah21 this should allow to get rid of another workaround in fpm:
```
f042e0625 chore: use traditional stringification for LFortran
```
However it is not quite working yet for me:
```
<ERROR> *fpm*:internal error: cannot get version - Invalid character in version number
  | FPM_RELEASE_VERSION
  |-^
STOP 1
```

Fixed this extra issue in https://github.com/lfortran/lfortran/pull/9748.

And with this second PR indeed the f042e0625 can be removed and it works for me locally.

jinangshah21:
Gfortran fails on newly added test

PR #9774, opened 2026-01-29
Fix pointer association for arrays of structs

certik:
Fixes #8538.

Depends on https://github.com/lfortran/lfortran/pull/9707.

PR #9707, opened 2026-01-26
Fix a bug in associate

certik:
Test added.

Fixes #9704.

PR #9775, opened 2026-01-30
Remove last fpm workaround

certik:
Now the only commit is:

```
* 5a88c62d1  XX: unallocated struct member in string concat
```

Which is an upstream bug fix that has since been merged.

This PR contains other open PRs:

* [x] https://github.com/lfortran/lfortran/pull/9707
* [ ] https://github.com/lfortran/lfortran/pull/9774
* [x] https://github.com/lfortran/lfortran/pull/9763

After they are all merged, we just need to rebase this PR which is just the tiny fpm commit change.

Locally on Apple M4 everything builds and all tests pass:
```console
$ fpm --compiler=$FC test --flag "--cpp --realloc-lhs-arrays --use-loop-variable-after-loop"
...
FOUND:    fpm_scratch_CC/.git/refs/tags
FOUND:    fpm_scratch_CC/src/fpm_scratch_CC.f90
new TEST TALLY=TTTTTTTTTTT
PASSED: all 11 tests passed
```


certik:
It might not actually depend on https://github.com/lfortran/lfortran/pull/9774, but let's just keep it as is, I think we can merge both. If https://github.com/lfortran/lfortran/pull/9774 fails, then we'll rebase and just merge the fpm commit, which might work already with the current main.

But just in case, I also submitted just the fpm change as a separate PR: https://github.com/lfortran/lfortran/pull/9781, if it passes, we'll merge it. We just need to lock in the win as soon as we can.

PR #9780, opened 2026-01-30
fix: update ArrayConstructor to include struct var for struct type

jinangshah21:
Towards #6840

Before this issue was for `[string_t::]` we was not able to extract struct symbol from this because we store only `string_t` structType in `ArrayConstructor`. So I added symbol in it.


src/libasr/ASR.asdl
```diff
@@ -153,7 +153,7 @@ expr
     | DictLen(expr arg, ttype type, expr? value)
     | Var(symbol v)
     | FunctionParam(int param_number, ttype type, expr? value)
-    | ArrayConstructor(expr* args, ttype type, expr? value, arraystorage storage_format)
+    | ArrayConstructor(expr* args, ttype type, expr? value, arraystorage storage_format, symbol? type_declaration)
```
certik: This ties expressions to the symbol table. I think this might not be the right design.

Currently the only tie is via Var, a special expression. We do not reference `symbol` from any other node. We can store a type however, so if you need some other type to store, we can.

jinangshah21: But we need expr or symbol in `ArrayConstructor` node to get corresponding Struct Symbol

jinangshah21: Consider below test:

```fortran
module m_config
  type :: string_t
  end type string_t
contains

  subroutine new(args)
    type(string_t), intent(in) :: args(:)
  end subroutine new

  subroutine tmp()
    call new([string_t::])
  end subroutine
end module m_config


program mre_cpp_config
  use m_config
  implicit none
  call tmp()
end program mre_cpp_config
```


`ArrayConstructor` node:
```clojure
(ArrayConstructor
                                                []
                                                (Array
                                                    (StructType
                                                        [(Allocatable
                                                        )]
                                                        []
                                                        .true.
                                                        .false.
                                                    )
                                                    (IntegerConstant 0 (Integer 4) Decimal))]
                                                    FixedSizeArray
                                                )
                                                ()
                                                ColMajor
                                            )
```

Here we don't have struct symbol information (in main we use 1st expr to get struct symbo and we don't have herel), so we need expr or symbol to get struct symbol

certik: Can you use `Var` for that?

jinangshah21: Yes we can, but eventually it will point to struct symbol

certik: Use `Var` then. There is special handling for `symbol` type in serialization and stuff, so I want to isolate it to ideally just Var (I think also FunctionCall). So just use `expr`, and you can check in verify() that it is always Var.

jinangshah21: I have update PR to use expr var instead of symbol

certik: Perfect, that works.

certik:
I think the CI will pass so I merged it.

PR #9781, opened 2026-01-30
Remove last fpm workaround

certik:
Now the only commit is:

	* 5a88c62d1  XX: unallocated struct member in string concat

Which is an upstream bug fix that has since been merged.


certik:
I canceled the CI, to speedup the CI on the other PRs that I think will succeed. If they fail, we restart the CI here.

PR #9787, opened 2026-01-30
fix: check the logical condition inside while loop also

### LINKED ISSUE #9785: Handle intrinsic function call like `all`, `any` in while loop condition
MRE:
```fortran
program mre_do_while_all
    implicit none
    logical :: finished(4)
    finished = .true.

    do while (all(finished))
        finished = .false.
    end do

end program mre_do_while_all
```

LFortran goes into infinite loop here.

HarshilShah1804:
Fixes #9785

### DIFF
--- a/src/libasr/pass/array_struct_temporary.cpp
+++ b/src/libasr/pass/array_struct_temporary.cpp
@@ -2414,16 +2414,14 @@ class ReplaceExprWithTemporaryVisitor:
         
         // Create condition variable 'c' which is a scalar logical
         ASR::ttype_t* logical_type = ASRUtils::TYPE(ASR::make_Logical_t(al, x.base.base.loc, 4));
-        ASR::symbol_t* c_sym = current_scope->resolve_symbol("_while_cond");
-        if (!c_sym) {
-            c_sym = (ASR::symbol_t*)ASR::make_Variable_t(al, x.base.base.loc, 
-                    current_scope, s2c(al, current_scope->get_unique_name("_while_cond")), 
-                    nullptr, 0, ASR::intentType::Local, nullptr, nullptr, 
-                    ASR::storage_typeType::Default, logical_type, nullptr, 
-                    ASR::abiType::Source, ASR::accessType::Public, 
-                    ASR::presenceType::Required, false, false, false, nullptr, false, false);
-            current_scope->add_symbol("_while_cond", c_sym);
-        }
+        char* c_name = s2c(al, current_scope->get_unique_name("_while_cond"));
+        ASR::symbol_t* c_sym = (ASR::symbol_t*)ASR::make_Variable_t(al, x.base.base.loc, 
+                current_scope, c_name, 
+                nullptr, 0, ASR::intentType::Local, nullptr, nullptr, 
+                ASR::storage_typeType::Default, logical_type, nullptr, 
+                ASR::abiType::Source, ASR::accessType::Public, 
+                ASR::presenceType::Required, false, false, false, nullptr, false, false);
+        current_scope->add_symbol(std::string(c_name), c_sym);
         ASR::expr_t* c_var = ASRUtils::EXPR(ASR::make_Var_t(al, x.base.base.loc, ASRUtils::symbol_get_past_external(c_sym)));
 
         current_body = parent_body;


certik:
Can you post the ASR or equivalent Fortran code after the transformation? Let's review that it is correct. It is important to get this right.

integration_tests/while_07.f90
```diff
@@ -0,0 +1,10 @@
+program mre_do_while_all
+    implicit none
+    logical :: finished(4)
+    finished = .true.
+
+    do while (all(finished))
+        finished = .false.
```
certik: Let's also add some counter here so that we don't get infinite loops if we make a mistake. If the counter gets to 2, then error stop. That way it will not hang our CI by going into infinite loop.

HarshilShah1804:
This is the fortran code generated after the `arrat_struct_tempraro` pass:
```fortran
! Fortran code after applying the pass: array_struct_temporary
program mre_do_while_all
implicit none
logical(4) :: __libasr_created__intrinsic_array_function_All
logical(4) :: __libasr_created__intrinsic_array_function_All1
logical(4), dimension(4) :: finished
finished = .true.
__libasr_created__intrinsic_array_function_All = All(finished)
do while (__libasr_created__intrinsic_array_function_All)
    __libasr_created__intrinsic_array_function_All1 = All(finished)
    if (__libasr_created__intrinsic_array_function_All1 == .false.) then
        exit
    end if
    finished = .false.
end do
end program mre_do_while_all
```

integration_tests/while_07.f90
HarshilShah1804: The CI checks have almost completed successfully and aren’t hanging, so I suggest adding the test in a follow-up PR. Adding it now would require rerunning all the checks. 
What do you suggest ?

jinangshah21 (approved):

integration_tests/while_07.f90
certik: Yes, definitely in the next PR.

certik:
Hm, I think the logic is not correct.

In the original loop we have:
```
do while (f())
end do
```
After refactoring we get:
```
c = f()
do while (c)
    c = f()
    if (c == .false.) exit
    // rest of the loop
end do
```

This is not an equivalent rewrite, here are the issues:

* It evaluates `f()` twice in the first loop run, instead of once. So if `f()` has side effects or is not deterministic this is not equivalent. It's also 2x slower if `f()` is expensive.
* The condition `f()` must be evaluated AFTER the loop, not BEFORE the loop. If the loop itself changes what `f()` returns, then the above is logically wrong.

Here is one way to rewrite it correctly:
```
c = f()
do while (c)
    // rest of the loop
    c = f()
end do
```
Then I think it is equivalent.

HarshilShah1804:
Perfect. I have updated the changes as per your suggestions. This is the fortran code generated after the new changes:
```fortran
! Fortran code after applying the pass: array_struct_temporary
program while_07
implicit none
logical(4) :: __libasr_created__intrinsic_array_function_All
logical(4) :: __libasr_created__intrinsic_array_function_All1
integer(4), save :: count = 0
logical(4), dimension(4) :: finished
finished = .true.
__libasr_created__intrinsic_array_function_All = All(finished)
do while (__libasr_created__intrinsic_array_function_All)
    finished = .false.
    count = count + 1
    __libasr_created__intrinsic_array_function_All1 = All(finished)
    __libasr_created__intrinsic_array_function_All = __libasr_created__intrinsic_array_function_All1
end do
print *, count
if (count /= 1) then
    error stop
end if
end program while_07


```

integration_tests/while_07.f90
HarshilShah1804: As, we had to make changes in the logic itself, I have updated the test as well in this PR while adding the new fix.

integration_tests/while_07.f90
```diff
@@ -0,0 +1,14 @@
+program while_07
+   implicit none
+   logical :: finished(4)
+   integer :: count = 0
+   finished = .true.
+
+   do while (all(finished))
+      finished = .false.
+      count = count + 1
```
certik, line 9: Add:
```
if (count > 1) error stop
```
To prevent infinite loops.

certik:
The answer is now good. The code looks ok. It fails in third-party tests:
```
<ERROR> Compilation failed for object " app_main.f90.o "
STOP 1
[ 99%]                       main.f90  done.

code generation error: Assignment target must be a variable or subscripted expression, got: type 35

```
Go ahead and extract an MRE for this, add it to our integration tests (whatever it is) and fix it as part of this PR. Then iterate this process until everything passes.

HarshilShah1804:
Yes, probably the error was coming as we were creating assignment statements between two temporaries. Now, I declared a new variable for evaluating the while condition and use that variable. I had also tried building `fpm`. It builds successfully with these changes.

integration_tests/while_07.f90
```diff
@@ -0,0 +1,14 @@
...
+   end do
+   print *, count
+   if (count > 1) error stop
```
certik, line 12:
```suggestion
   if (count /= 1) error stop
```

integration_tests/while_07.f90
certik: This is still not fixed.

certik:
I merged it, can you please fix the test in the next PR? And post the Fortran code. Hopefully it's good. If not, we can always revert this PR.

PR #9789, opened 2026-01-30
ci: add LAPACK ILP64 testing to third party codes

krystophny:
## Summary
Run the complete LAPACK test suite in both 32-bit and 64-bit integer (ILP64) modes.

## Changes
- Refactor Section 14 to run full test suite twice (32-bit and ILP64) without code duplication
- Add missing eigenvalue condition tests: `dec`, `cec`, `zec`
- Add mixed precision linear equation tests: `dstest`, `zctest`
- Fix double real eigenvalue list (was incorrectly using `sec` instead of `dec`)

## Test Coverage
- **Linear equations**: 8 tests (single/double × real/complex × standard/RFP)
- **Mixed precision**: 2 tests (double-single real and complex)
- **Eigenvalue**: ~100 tests across all precisions (nep, sep, svd, ec, ed, gg, gd, sb, sg, bal, bak, gbal, gbak, bb, glm, gqr, gsv, csd, lse, dmd)

All tests run twice: once with 32-bit integers, once with ILP64 (`-fdefault-integer-8` + `BUILD_INDEX64=ON`).

## Verified locally
All tests pass in both modes.

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -877,9 +877,9 @@ time_section "🧪 Testing LAPACK" '
 
 
 ##########################
-# Section 14: Reference-LAPACK with BUILD_TESTING (Full Suite)
+# Section 14: Reference-LAPACK Full Test Suite (32-bit and 64-bit integers)
 ##########################
-time_section "🧪 Testing Reference-LAPACK v3.12.1 with BUILD_TESTING (Full Suite)" '
+time_section "🧪 Testing Reference-LAPACK v3.12.1 Full Test Suite" '
     export PATH="$(pwd)/../src/bin:$PATH"
     git clone --depth 1 --branch v3.12.1 https://github.com/Reference-LAPACK/lapack.git lapack-testing
     cd lapack-testing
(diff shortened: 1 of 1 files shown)


certik (approved):
So this change is impossible to review. I have no idea if it is correct. I checked the logs, it seems to be running lapack twice, once in i32, once in i64, so it probably works.

So let's merge this.

Can you then please submit a PR that reverts some of your i64 fixes and let's see if this test actually breaks? We need to ensure that if lapack fails that the CI fails and I can't guarantee that this PR doesn't break this.

When you do it, just skip integration tests (which would also break if you revert a i64 fix), but ensure third party tests run and we need to ensure they break as they should. Then we also need to do the same for i32 --- that it actually breaks as it should.

krystophny:
> So this change is impossible to review. I have no idea if it is correct. I checked the logs, it seems to be running lapack twice, once in i32, once in i64, so it probably works.
> 
> So let's merge this.
> 
> Can you then please submit a PR that reverts some of your i64 fixes and let's see if this test actually breaks? We need to ensure that if lapack fails that the CI fails and I can't guarantee that this PR doesn't break this.
> 
> When you do it, just skip integration tests (which would also break if you revert a i64 fix), but ensure third party tests run and we need to ensure they break as they should. Then we also need to do the same for i32 --- that it actually breaks as it should.

Done in #9809

krystophny:
## Verification: LAPACK ILP64 CI catches regressions

Per @certik's request, I created a test PR that reverts an ILP64 fix to verify the CI catches regressions:

**Test PR:** #9809 (intentionally failing, do not merge)

This reverts PR #9733 (cast KIND argument to int32 in CMPLX for ILP64 mode). The `Tests::Run-Exhaustive` label has been added to trigger the full LAPACK test suite.

**Expected result:** CI should fail on LAPACK ILP64 tests, proving the CI detection works.

Will update with results once CI completes.

PR #9791, opened 2026-01-31
cmake: add FILE parameter to RUN macro for test reuse

### LINKED ISSUE #9759: CI: allow to reuse the same test twice
This seems equivalent to `realloc_lhs_16.f90`, what is the purpose of this duplicate test?

Is the idea that it is tested with `--descriptor-index-64` and not `-fdefault-integer-8`?

Why are the tests not identical exactly? Let's document this in the comment that it is identical?

Also let's improve the cmake in the future, so that we do not need to duplicate files, but can use the same file for both tests.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/9729#discussion_r2742921046_

krystophny:
## Summary
Add optional FILE parameter to the RUN macro that allows specifying a different source file than the test NAME. This enables running the same test file with different configurations without duplicating source.

Fixes #9759

## Changes
- Add `FILE` to `oneValueArgs` in the RUN macro
- Use FILE if provided, otherwise fall back to NAME as the file name
- Consolidate `realloc_lhs_17` to use `realloc_lhs_16.f90` via FILE parameter
- Delete duplicate `realloc_lhs_17.f90`

## Example
```cmake
RUN(NAME realloc_lhs_16 LABELS gfortran llvm
    EXTRA_ARGS -fdefault-integer-8 --realloc-lhs-arrays
    GFORTRAN_ARGS -fdefault-integer-8)
RUN(NAME realloc_lhs_17 FILE realloc_lhs_16 LABELS llvm
    EXTRA_ARGS --descriptor-index-64 --realloc-lhs-arrays)
```

## Test plan
- [x] realloc_lhs_16 test passes
- [x] realloc_lhs_17 test passes (using realloc_lhs_16.f90 source)


certik (approved):
I think that's fine. Did you try to fail the test to ensure everything works correctly? Just do it locally by hand.

krystophny:
**Verified: FILE parameter mechanism works correctly**

I intentionally broke `realloc_lhs_16.f90` by adding an error stop:

```fortran
    print *, "INTENTIONALLY BROKEN - FILE PARAMETER TEST"
    error stop "This should fail"
```

Then ran `realloc_lhs_17` test (which uses `FILE realloc_lhs_16`):

```
$ ./run_tests.py -t realloc_lhs_17 -s

Building Fortran object CMakeFiles/realloc_lhs_17_LLVM_GOC.dir/realloc_lhs_16.f90.o
Linking Fortran executable realloc_lhs_17_LLVM_GOC
Built target realloc_lhs_17_LLVM_GOC
Test project /home/ert/code/lfortran-dev/lfortran/integration_tests/test-llvm
    Start 1349: realloc_lhs_17_LLVM_GOC
1/1 Test #1349: realloc_lhs_17_LLVM_GOC ..........***Failed    0.00 sec
INTENTIONALLY BROKEN - FILE PARAMETER TEST

The following tests FAILED:
	1349 - realloc_lhs_17_LLVM_GOC (Failed)                  llvm
```

Key observations:
1. CMake shows it's building `realloc_lhs_16.f90.o` (not `realloc_lhs_17.f90.o`) for the `realloc_lhs_17` test
2. The test correctly fails with our intentional error message
3. After restoring the file, the test passes

The FILE parameter mechanism is working as intended.

Issue #9626, opened 2026-01-21
Follow-up work after #9559


certik:
TODO for subsequent PRs:

* https://github.com/lfortran/lfortran/pull/9559/changes#r2709433300
* https://github.com/lfortran/lfortran/pull/9559#discussion_r2713267256
* Figure out how to clean these casts up in ASR itself, to simplify the backend

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/9559#issuecomment-3779255333_

PR #9798, opened 2026-01-31
fix: add ASR pass for logical array i8<->i1 casts

krystophny:
## Summary
- Add new ASR pass `logical_array_cast` that makes i8<->i1 conversions for logical arrays explicit
- Remove codegen workarounds that handled these conversions implicitly
- Scalar logicals remain i1, array elements use i8 storage
- Fix for pass-by-reference logical array elements (intent out/inout)

Refs #9626

## Why
The previous approach had scattered workarounds in LLVM codegen to convert between i1 (scalar logical) and i8 (logical array element storage). This made the code complex and error-prone. Moving the conversions to ASR level via explicit Cast nodes provides a cleaner architecture.

This addresses the core request from #9626: "Figure out how to clean these casts up in ASR itself, to simplify the backend" and comment [r2709433300](https://github.com/lfortran/lfortran/pull/9559#discussion_r2709433300): "I think the fix would be in one single spot eliminating all of these checks."

**Stage:** ASR Pass

**Note:** This is a refactoring/cleanup PR. Existing tests continue to pass with the cleaner implementation.

## Changes
- New files: `src/libasr/pass/logical_array_cast.cpp` and `.h`
- Modified `pass_manager.h`: Register and enable the pass
- Modified `CMakeLists.txt`: Add new source file
- Modified `asr_to_llvm.cpp`: Remove two workaround blocks (~15 lines)
- New test: `integration_tests/logical_array_01.f90`

## How it works
The pass transforms logical ArrayItem nodes:
1. Changes type from `Logical(k)` to `Integer(1)` (maps to i8)
2. Wraps value usages in `Cast(IntegerToLogical)` for i8->i1
3. Wraps assignment RHS in `Cast(LogicalToInteger)` for i1->i8
4. **Skips transformation for pass-by-reference** (intent out/inout) - codegen handles these transparently

Codegen then handles these uniformly - no special cases needed.

## Tests
- `logical_array_01.f90` - comprehensive test for logical array operations including intent(out/inout)
- `arrays_01_logical.f90` passes (normal and --fast)
- `where_07.f90`, `where_08.f90`, etc. pass

### Test fails before fix (intent out/inout bug)
```
$ git checkout 034ee226a  # PR commit before fix
$ scripts/lf.sh build
$ ./lfortran/build/src/bin/lfortran /tmp/test_intent_out_array.f90 -o /tmp/test && /tmp/test
ERROR STOP FAILED
Exit code: 1
```

### Test passes after fix
```
$ git checkout b84628558  # PR commit with fix
$ scripts/lf.sh build
$ ./lfortran/build/src/bin/lfortran /tmp/test_intent_out_array.f90 -o /tmp/test && /tmp/test
All tests passed!
```

Test file (`/tmp/test_intent_out_array.f90`):
```fortran
program test_intent_out_array
    implicit none
    logical, allocatable :: vals(:)
    integer :: i

    allocate(vals(10))
    do i = 1, 10
        call set_val(vals(i), mod(i, 2) == 1)
    end do

    if (.not. vals(1)) error stop "FAILED"
    print *, "All tests passed!"

contains
    subroutine set_val(val, expected)
        logical, intent(out) :: val
        logical, intent(in) :: expected
        val = expected
    end subroutine
end program
```

### DIFF
--- a/src/libasr/pass/logical_array_cast.cpp
+++ b/src/libasr/pass/logical_array_cast.cpp
@@ -32,15 +32,23 @@ public:
     Allocator& al;
     bool in_assignment_target;
     bool in_get_pointer;
+    bool in_pass_by_ref;
 
     LogicalArrayCastReplacer(Allocator& al_) : al(al_), in_assignment_target(false),
-                                               in_get_pointer(false) {
+                                               in_get_pointer(false), in_pass_by_ref(false) {
         call_replacer_on_value = false;
     }
 
     void replace_ArrayItem(ASR::ArrayItem_t* x) {
         ASR::ttype_t* elem_type = x->m_type;
         if (!ASRUtils::is_logical(*elem_type)) {
+            return;
+        }
+
+        if (in_pass_by_ref) {
+            // For pass-by-reference (intent out/inout), don't transform at all.
+            // Keep the ArrayItem as Logical type since the callee expects Logical.
+            // The codegen handles the i8 storage transparently via get_el_type().
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/libasr/pass/logical_array_cast.cpp
```diff
@@ -0,0 +1,334 @@
...
+under certain optimizations. Scalar logicals remain i1 for efficiency.
+
+This pass makes the i1 <-> i8 conversions explicit at ASR level:
+1. ArrayItem of logical type is changed to Integer(kind=1) type
+2. When used as a value, the ArrayItem is wrapped in Cast(IntegerToLogical)
+3. When storing to a logical ArrayItem, the RHS is wrapped in Cast(LogicalToInteger)
+4. For GetPointer (used in EQUIVALENCE), ArrayItem is transformed but NOT wrapped
+   in Cast, since we need the raw address, not the value
+
+This removes the need for scattered workarounds in codegen.
```
certik, line 22: So this must happen already in AST->ASR and we need to add verify() to ensure those casts are always present.

The ASR must always be correct, so when we now make the change that casts are required in ASR, they must be there from the beginning, so that at each phase the ASR is correct. For the same reason we do not insert other casts as ASR passes either.

Why does ArrayItem for a logical have to be an integer? That doesn't seem right?

krystophny: I try to find a cleaner solution. Maybe we can even keep i1 and remove the previous codegen hack if we get LLVM optimization to work cleanly with i1 arrays for logicals

krystophny:
# Automatic Analysis: i1 vs i8 for Logical Arrays

## Investigation Summary

I conducted an extensive investigation into implementing a pure i1 solution for logical arrays while avoiding LLVM O3 optimization issues with `--fast`. Here are the findings:

## Key Findings

### 1. LLVM Memory Layout
LLVM stores `[N x i1]` arrays as **byte arrays** - each element takes 1 byte, not 1 bit. Verified via assembly inspection:

```asm
; Both [100 x i1] and [100 x i8] produce identical storage:
mov  x9, #72340172838076673    ; 0x0101010101010101
stp  x9, x9, [x29, #-200]      ; Store 16 bytes at a time
```

**Conclusion**: There is **no memory savings** from using i1 over i8.

### 2. The `freeze` Instruction Approach (FAILED)

I implemented a solution using LLVM's `freeze` instruction (available since LLVM 10) to wrap all i1 loads from arrays. The implementation:

```cpp
// In get_el_type():
case ASR::ttypeType::Logical: {
    el_type = llvm::Type::getInt1Ty(context);  // Use i1
    break;
}

// In array element loads:
llvm::Value* v_i1 = CreateLoad2(llvm::Type::getInt1Ty(context), ptr);
tmp = builder->CreateFreeze(v_i1);  // Wrap with freeze
```

**Result**: Tests still **failed** with `--fast`. The freeze instruction was properly preserved through O3 optimization (verified in optimized IR), but the program still produced wrong results:

```
=== With --fast (i1 + freeze) ===
1    1    TRUE
2    1    TRUE
3    1    TRUE
1    2    FALSE   <- Should be TRUE
2    2    FALSE   <- Should be TRUE
3    2    TRUE
...
FAIL: not all true
```

### 3. Root Cause Analysis

The LLVM O3 optimizer or ARM64 backend has deep-seated issues with i1 arrays that cannot be fixed by:
- Adding `freeze` to loads
- Proper array initialization  
- Correct index calculations (verified)

The issue appears to be in how LLVM's store merging or ARM64 codegen handles i1 arrays during optimization, causing memory corruption where some elements appear as FALSE when they should be TRUE.

### 4. Why i8 Works

The i8-based implementation from PR #9559:
- Uses the **same amount of memory** as i1 (both byte-backed)
- Avoids **all LLVM optimization issues** by using well-defined i8 semantics
- Works **correctly with `--fast`** on all platforms
- Uses explicit zext/trunc at boundaries for clean i1↔i8 conversion

## Recommendation

The i8 storage approach is the **correct solution**. The preference for i1 is based on "semantic purity" rather than practical benefits:

| Aspect | i1 | i8 |
|--------|----|----|
| Memory usage | 1 byte per element | 1 byte per element |
| LLVM O3 safety | ❌ Fails with `--fast` | ✅ Works correctly |
| Semantic clarity | Looks like boolean | Explicit byte storage |

**The ASR-based approach in this PR is architecturally cleaner** than the codegen-level fix in #9559, as it encodes the type conversion upstream rather than working around it downstream. The key is that the underlying storage must remain i8 to avoid LLVM optimization bugs.

## References
- [LLVM Undefined Behavior Manual](https://llvm.org/docs/UndefinedBehavior.html)
- [Taming Undefined Behavior in LLVM (freeze instruction)](https://sf.snu.ac.kr/freeze/)
- [LLVM D29011 - freeze instruction](https://reviews.llvm.org/D29011)

krystophny:
Closing in favor of the existing codegen-only approach on main.

After thorough analysis of both this PR and #9851, the ASR-level approach (whether as a pass or in semantics) creates a viral unwrapping problem: every place that checks `is_a<ArrayItem_t>` needs `unwrap_logical_byte_cast()` first, which is fragile and affects all backends for what is purely an LLVM concern.

The 3 codegen workarounds already on main (store i1->i8, load i8->i1, function arg bitcast) are correct, localized, and work with both normal and --fast modes. The i1/i8 mismatch is an LLVM backend implementation detail that does not belong in ASR.

See #9851 (also closing) for the same reasoning.

PR #9807, opened 2026-01-31
feat: add --help=category option for grouped help output

krystophny:
## Summary
- Adds `--help=<category>` option to display options from a specific category
- Available categories: warnings, language, preprocessing, output, pass, backend, symbol, mangling, misc, lsp

Example: `lfortran --help=warnings`

Refs #7937 (partially addresses - adds grouped help output)

## Why
Users requested a way to view options by category without scrolling through the entire help output. This provides a GCC-style `--help=<topic>` interface.

## Changes
- [`src/bin/lfortran_command_line_parser.cpp`](https://github.com/lfortran/lfortran/blob/6320979a9580eabc3aedea117e5387b2e8150cdc/src/bin/lfortran_command_line_parser.cpp#L33-L96): Added `print_category_help()`, `get_category_map()`, `find_help_category_arg()`, and `handle_help_category()` functions

## Example output
```
$ lfortran --help=warnings
Warning Options:
  --no-warnings                  Turn off all warnings
  --no-style-suggestions         Turn off style suggestions
  --style-suggestions            Enable style suggestions
  --no-error-banner              Turn off error banner
  --error-format                 Control how errors are produced (human, short)
```

```
$ lfortran --help=invalid
Unknown help category: invalid
Available categories: backend, language, lsp, mangling, misc, output, pass, preprocessing, symbol, warnings
```

## Tests
- All reference tests pass
- All integration tests pass
- Manual testing of `--help=<category>` for all categories

### DIFF
--- a/src/bin/lfortran_command_line_parser.cpp
+++ b/src/bin/lfortran_command_line_parser.cpp
@@ -1,9 +1,10 @@
 #include <libasr/exception.h>
 #include <libasr/string_utils.h>
 
+#include <algorithm>
+#include <cctype>
 #include <iomanip>
 #include <iostream>
-#include <map>
 
 #include <lfortran/utils.h>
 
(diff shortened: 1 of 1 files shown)


src/bin/lfortran_command_line_parser.cpp
```diff
@@ -28,7 +30,220 @@ namespace LCompilers::CommandLineInterface {
...
+        app.add_flag("--descriptor-index-64", dummy_compiler_options.descriptor_index_64, "Use 64-bit indices in array descriptors (implied by -fdefault-integer-8)")->group(group_miscellaneous_options);
+        app.add_flag("--array-bounds-checking", dummy_compiler_options.po.bounds_checking, "Enables runtime array bounds checking")->group(group_miscellaneous_options);
+        app.add_flag("--no-array-bounds-checking", dummy_bool, "Disables runtime array bounds checking")->group(group_miscellaneous_options);
+        app.add_flag("--strict-array-bounds-checking", dummy_compiler_options.po.strict_bounds_checking, "Enables strict runtime array bounds checking: Array passed into subroutine must exactly match the expected size")->group(group_miscellaneous_options);
+
+        app.add_flag("--show-errors", dummy_opts.show_errors, "Show errors when LSP is running in the background")->group(group_lsp_options);
+        app.add_flag("--show-document-symbols", dummy_opts.show_document_symbols, "Show symbols in lfortran file")->group(group_lsp_options);
```
certik: All this code is now duplicated, right?

Let's figure out how to define a given compiler option just once.

krystophny:
Addressed the code duplication review comment.

The refactoring eliminates the duplicate `setup_options()` function that contained ~150 lines of duplicated option definitions with dummy variables. 

The key insight is that we can use the already-configured `app` instance after options are set up in `parse()` but before `app.parse()` is called. The `--help=X` check happens at this point and uses the same app that will be used for normal parsing.

**Changes:**
- Removed `setup_options()` function (which had dummy static variables)
- Removed `static void setup_options(CLI::App &app);` declaration from header
- Changed `handle_help_category()` to accept the already-configured `app` as a parameter
- Added the help category check after options are configured but before `app.parse()`

**Net result:** -162 lines of code deleted, options defined exactly once.

src/bin/lfortran_command_line_parser.cpp
krystophny: Fixed!

certik:
Looks better. It's still kind of fragile with all these strings comparisons. Let me think how to achieve this robustly.

krystophny:
Addressed: categories are now derived from the app's option groups at runtime via `get_group()`. The duplicate `category_map` is gone. Matching uses the lowercased first word with prefix support (`--help=warn`, `--help=misc`, etc.).

PR #9841, opened 2026-02-01
feat: add runtime formatting for labels

xaerru:
Address comment from https://github.com/lfortran/lfortran/pull/9686

Create a new `LLVMUtils::RuntimeLabel`. Added runtime formatting for labels.

### DIFF
--- a/src/libasr/diagnostics.h
+++ b/src/libasr/diagnostics.h
@@ -1,7 +1,6 @@
 #ifndef LFORTRAN_DIAGNOSTICS_H
 #define LFORTRAN_DIAGNOSTICS_H
 
-#include <llvm/IR/Value.h>
 #include <tuple>
 #include <libasr/location.h>
 #include <libasr/stacktrace.h>
@@ -51,23 +50,6 @@ struct Label {
 
     Label(const std::string &message, const std::vector<Location> &locations,
             bool primary=true) : primary{primary}, message{message} {
-        for (auto &loc : locations) {
-            spans.emplace_back(loc);
-        }
-    }
-};
-
-/*
- * A Label for runtime error messages
- */
-struct RuntimeLabel {
-    bool primary; // primary or secondary label
-    std::string message; // format string message
-    std::vector<diag::Span> spans; // one or more spans
-    std::vector<llvm::Value*> args; // arguments for format string
-
-    RuntimeLabel(const std::string &message, const std::vector<Location> &locations, const std::vector<llvm::Value*> &args = {},
-            bool primary=true) : primary{primary}, message{message}, args{args} {
         for (auto &loc : locations) {
             spans.emplace_back(loc);
         }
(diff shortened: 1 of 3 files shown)


src/libasr/diagnostics.h
```diff
@@ -1,6 +1,7 @@
 #ifndef LFORTRAN_DIAGNOSTICS_H
 #define LFORTRAN_DIAGNOSTICS_H
 
+#include <llvm/IR/Value.h>
```
certik: This adds LLVM dependency in our diagnostics. Let's not do that. If anything, we need to remove dependencies. LLVM should only be used in the LLVM backend, nowhere else. In the future we will hide it behind an API, so we would never use it directly.

certik: To move forward, we need to create whatever data structures are needed ourselves.

src/libasr/diagnostics.h
```diff
@@ -56,6 +57,23 @@ struct Label {
...
 };
 
+/*
+ * A Label for runtime error messages
+ */
+struct RuntimeLabel {
+    bool primary; // primary or secondary label
+    std::string message; // format string message
+    std::vector<diag::Span> spans; // one or more spans
+    std::vector<llvm::Value*> args; // arguments for format string
```
certik: This must work when LFortran is compiled without LLVM as well.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2735,13 +2735,39 @@ LFORTRAN_API void _lcompilers_runtime_error(Label *labels, uint32_t n_labels, co
...
                            label->message, use_colors);
+            free(label->message);
         }
     }
 
     fflush(stderr);
     free(error_msg);
 }
 
+LFORTRAN_API char* _lcompilers_snprintf(const char* format, ...) {
```
certik, line 2746: Put this into the header file as well.

tests/reference/run-array_bounds_check_05-d584fab.stderr
```diff
@@ -5,4 +5,4 @@ runtime error: Array shape mismatch in subroutine 'my'. Tried to match size 3 of
   |             ^
   |
 7 |     call my(x, 5)
-  |             ~ Expected size
+  |             ~ Expected range 1:5, got 3
```
certik, line 8: This error should say: "Parameter is size 5, argument is size 3". Or: "Expected size 5, passed in size 3"

certik: The `Expected range 1:5, got 3` style error is good for things like:
```
x = [1, 2, 3]
x(4) = 1
```

certik: I made this https://github.com/lfortran/lfortran/issues/9890.

Issue #9845, opened 2026-02-02
Handle nested function call inside `implied_do_loop` correctly


jinangshah21:
MRE:
```fortran
program test_hidden
    implicit none

    type :: string_t
        character(len=:), allocatable :: s
    end type string_t

    type(string_t), allocatable :: file_names(:)
    logical, allocatable :: is_source(:)

    integer :: i

    allocate(file_names(3))

    file_names(1)%s = "main.f90"
    file_names(2)%s = ".gitignore"
    file_names(3)%s = "utils.f90"

    i = 100
    is_source = [(is_hidden_file(basename(file_names(i)%s)), i = 1, size(file_names))]

    do i = 1, size(is_source)
        print *, trim(file_names(i)%s), " -> ", is_source(i)
    end do

contains
    function basename(path) result(tmp)
        character(len=*), intent(in) :: path
        character(len=:), allocatable :: tmp
        tmp = path
    end function basename

    logical function is_hidden_file(name)
        character(len=*), intent(in) :: name
        is_hidden_file = (len(name) > 0 .and. name(1:1) == ".")
    end function is_hidden_file

end program test_hidden
```

LFortran Error:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran c.f90 --realloc-lhs-arrays
runtime error: Array 'file_names' index out of bounds. Tried to access index 100 of dimension 1, but valid range is 1 to 3.
  --> c.f90:20:43
   |
20 |     is_source = [(is_hidden_file(basename(file_names(i)%s)), i = 1, size(file_names))]
```

PR #9846, opened 2026-02-02
CI: update CI with 2 workarounds on latest FPM main

jinangshah21:
Towards #6840

With this 2 MREs are left to run FPM main without any workaround: #9785 and #9845


certik:
I would be careful here: the old commit works without workarounds. Now we don't test it anymore, so we don't actually know if things work. The latest main requires workarounds, and it will take some time to get it working. Even though the above two PRs are fixed, it still doesn't work for me: https://github.com/lfortran/lfortran/issues/6840#issuecomment-3836468746.

Issue #9719, opened 2026-01-27
Direct access READ: Compiler error with logical reads


w6ws:
Follow-on to PRs #9517 and #9518:
```
program daio4
  implicit none

  logical :: l_1d(8), l_2d(2,2), l_3d(2,2,2)
  logical :: larray(8)

  integer :: i10, i, j, k

  open (newunit=i10, file='direct.dat', access='direct', recl=40, status='unknown')
  larray = .true.
  write (i10, rec=1) larray
  larray = .false.
  write (i10, rec=2) larray

! Element by element
  l_1d = .false.; l_2d = .false.; l_3d = .false.
  read (i10, rec = 1) l_1d(1), l_1d(2), l_2d(1,2), l_2d(2,2), l_3d(1,1,2),  &
   l_3d(2,1,2), l_1d(7), l_1d(8)
  print *, 'rec 1 test: ',  &
    pf (all (l_1d(1:2)) .and. all (l_2d(1:2,2)) .and. all (l_3d(1:2,1,2))  &
    .and. all (l_1d(7:8)))

! Implied loop
  l_3d = .true.
  read (i10, rec = 2) (((l_3d (J,K,I), I=1,2), K=1,2), J=1,2)
  print *, 'rec 2 test: ', pf (.not. any (l_3d))

  close (i10)

contains

  elemental function pf (l)
    character(4) :: pf
    logical, intent(in) :: l

    pf = merge ('pass', 'fail', l)

  end function

end program
```
Attempting to compile with lfortran gives the attached dump.

[dump.txt](https://github.com/user-attachments/files/24892391/dump.txt)

w6ws:
Tested and works great.  Thanks!

PR #9847, opened 2026-02-02
fix: cast in read

yuvrajsinghgmx:
This PR introduces fix in `visit_FileRead` to auto cast the `var_to_read_into` to required type.

a.f90
```fortran
program cast_test
    logical :: arr(2) 
    open(10, file='read_34.txt', access='direct', recl=4, status='replace')
    write(10, rec=1) .true.
    arr = .false.
    ! This read requires a cast for read to work correctly
    read(10, rec=1) arr(1)
    if (.not. arr(1)) error stop
    close(10, status='delete')
end program
```

Before FIX dump on `a.f90` : -

```

declare i64 @_lfortran_open(i32, i8*, i64, i8*, i64, i8*, i64, i8*, i64, i8*, i64, i32*, i8*, i64, i8*, i64, i8*, i64, i8*, i64, i8*, i64, i32*, i8*, i64)

declare void @_lfortran_seek_record(i32, i32, i32*)

declare void @_lfortran_file_write(i32, i32*, i8*, i64, ...)

declare void @_lcompilers_print_error(i8*, ...)

declare void @exit(i32)

declare void @_lfortran_read_logical(i1*, i32, i32*)

declare void @_lfortran_empty_read(i32, i32*)

declare void @_lfortran_close(i32, i8*, i64, i32*)

  %28 = getelementptr [2 x i8], [2 x i8]* %arr, i32 0, i32 0
 i1*  call void @_lfortran_read_logical(i8* %28, i32 10, i32* null)

  %28 = getelementptr [2 x i8], [2 x i8]* %arr, i32 0, i32 0
 i1*  call void @_lfortran_read_logical(i8* %28, i32 10, i32* null)



```

After this fix program in #9719 gives the following error instead of dump : 

```
/lfortran$ ./src/bin/lfortran a.f90
rec 1 test:     fail
runtime error: Runtime error: Array 'l_3d' index out of bounds.

Tried to access index 0 of dimension 1, but valid range is 1 to 2.

  --> a.f90:25:26
   |
25 |   read (i10, rec = 2) (((l_3d (J,K,I), I=1,2), K=1,2), J=1,2)
   |                          ^^^^ 

```

### DIFF
--- a/integration_tests/read_34.f90
+++ b/integration_tests/read_34.f90
@@ -1,10 +1,29 @@
 program cast_test
-    logical :: arr(2) 
-    open(10, file='read_34.txt', access='direct', recl=4, status='replace')
-    write(10, rec=1) .true.
-    arr = .false.
-    ! This read requires a cast for read to work correctly
-    read(10, rec=1) arr(1)
-    if (.not. arr(1)) error stop
+    implicit none
+    integer, parameter :: i1 = 1
+    logical(kind=i1) :: arr(2)
+    logical(kind=i1) :: scalar_val
+    
+    open(10, file='read_34.bin', access='direct', recl=2, status='replace')
+
+    ! Test 1: Scalar Read 
+    
+    write(10, rec=1) .true._i1
+    scalar_val = .false._i1
+    read(10, rec=1) scalar_val
+    if (.not. scalar_val) then
+        print *, "FAILED: Scalar Read"
+        error stop
+    end if
+
+    ! Test 2: Array Element Read 
+    write(10, rec=2) .false._i1
+    arr(1) = .true._i1
+    read(10, rec=2) arr(2)
+    if (arr(2)) then
+        print *, "FAILED: Array Element Read"
+        error stop
+    end if
+
     close(10, status='delete')
 end program


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13820,6 +13820,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                 module.get())->getPointerTo();
                             var_to_read_into = llvm_utils->CreateLoad2(t, var_to_read_into);
                         }
+                        llvm::Type* expected_arg_type = fn->getFunctionType()->getParamType(0);
+                        if (var_to_read_into->getType() != expected_arg_type) {
+                            var_to_read_into = builder->CreateBitCast(var_to_read_into, expected_arg_type);
+                        }
```
jinangshah21, line 13826: Ah, we don't need to bitcast in every case, currently we extended logical types in llvm to i8 so we need to change type of `_lfortran_logical_read` argument from i1 to i8 I guess, @certik can you please confirm?

certik: So logical scalars are i1, but logical arrays are i8.

Yes, this casting should be limited to logical only and be done right to work for both scalars and arrays. Let's improve the test to test both? And then do a proper fix.

yuvrajsinghgmx: For storing `logical_array` we can cast it to `i8` while storing in `_lfortran_read_array_logical` but what if the read var type is `i32` or `i64` for that do we need new APIs like `_lfortran_read_array_logical_i16` ,`_lfortran_read_array_logical_i32`, `_lfortran_read_array_logical_i64`. ???

Could you please explain this @certik ??

certik: Can you give a Fortran example of an input type i32 that is read into logical? Is that allowed?

yuvrajsinghgmx: > Can you give a Fortran example of an input type i32 that is read into logical? Is that allowed?

Got it so that is not technically possible.

certik: I don't know if it is not technically possible, since I don't understand exactly what you are asking. But the best way to ask a question is with a concrete Fortran or ASR code.

PR #9849, opened 2026-02-02
feat: allow character arrays as format statement in read/write statements

### LINKED ISSUE #9844: Formatted I/O: character array used as format gives unbalanced parenthesis error
```
program char_array_fmt
  implicit none

! The Fortran standard (all of them from F77 to present) allow the use of a
! character array, and treated as if the individual elements were concatenated.
! See, for example chapter 13.2.2 of F2023.
!
! This is actually the character data type version of an older F66 capability that allowed an array
! of Hollerith constants to be used as a format.  (F66 7.2.3.10)

  character(4) :: fmt(4)

  fmt = [ '("he', 'llo ', 'worl', 'd!")' ]
  write (*, fmt)

end program
```
```
$ lfortran char_array_fmt.f90
Error: Unbalanced paranthesis in format string
$
```
```
$ gfortran char_array_fmt.f90
$ ./a.out
hello world!
$
```

HarshilShah1804:
Fixes #9844


certik:
If all tests pass, it's probably fine. We can later refactor it to have a more faithful representation in ASR.

PR #9851, opened 2026-02-02
asr: add LogicalByteToLogical and LogicalToLogicalByte cast kinds

krystophny:
## Summary
- Add new cast_kind values: `LogicalByteToLogical`, `LogicalToLogicalByte`
- Handle these casts in all backends (LLVM, C++, Julia, WASM)
- **Insert `LogicalToLogicalByte` cast during semantics** when storing to logical array elements

Refs #9626

## Why
This implements explicit ASR-level representation of logical array byte storage conversion, as discussed in PR #9559 and issue #9626.

**Key architectural points from the discussion:**
1. certik: "I think ASR cast is better" - prefers ASR-level solution
2. certik: "we should be doing all these casts in ASR itself" 
3. certik (on PR #9798): "this must happen already in AST->ASR" - casts should be in semantics, not an ASR pass
4. certik (on PR #9798): "Why does ArrayItem for a logical have to be an integer? That doesn't seem right?" - ArrayItem must stay Logical type

**This PR:**
- Defines the cast kinds that represent byte↔logical conversion
- **Inserts `LogicalToLogicalByte` casts in `make_Assignment_t_util`** during AST→ASR phase
- Keeps ArrayItem type as Logical (not Integer!)
- Maintains backward compatibility with existing codegen fallback for edge cases

**Stage:** Semantics (AST→ASR)

## What these casts represent
- `LogicalByteToLogical`: Convert byte-backed (i8) logical array element to scalar i1 (for reads) - handler ready, not yet used
- `LogicalToLogicalByte`: Convert scalar i1 logical to byte (i8) for storage (for writes) - actively used

## Changes
- `ASR.asdl`: Add `LogicalByteToLogical`, `LogicalToLogicalByte` cast kinds
- `asr_to_llvm.cpp`: Handle casts + clarified comments on fallback logic
- `asr_to_c_cpp.h`, `asr_to_julia.cpp`, `asr_to_wasm.cpp`: Handle casts (no-ops, bool is byte-sized)
- `asr_utils.h`: Add helper `wrap_logical_value_for_array_store()`
- `asr_utils.cpp`: Insert `LogicalToLogicalByte` cast in `make_Assignment_t_util`
- `pass/pass_utils.cpp`: Same for the local `make_Assignment_t_util`

## Tests
- Updated `tests/reference/asr-arrays_01_logical-*` to reflect new ASR with explicit casts
- All tests pass locally

## Future work
- The `LogicalByteToLogical` cast kind and handler are ready for when logical array reads are moved from codegen to ASR level
- Current read handling via `get_el_type()` works correctly

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -2306,6 +2306,7 @@ public:
                 }
             }
             // Assume that tmp is an `ArraySection` or `ArrayItem`
+            tmp_stmt = ASRUtils::unwrap_logical_byte_cast(tmp_stmt);
             if( ASR::is_a<ASR::ArraySection_t>(*tmp_stmt) ) {
                 ASR::ArraySection_t* array_ref = ASR::down_cast<ASR::ArraySection_t>(tmp_stmt);
                 new_arg.m_a = array_ref->m_v;
@@ -4436,9 +4437,9 @@ public:
                     }
                 }
             }
-            // For assigning Array Variable    
-            else if (ASR::is_a<ASR::ArrayItem_t>(*target)) {
-                ASR::ArrayItem_t *array_item = ASR::down_cast<ASR::ArrayItem_t>(target);
+            // For assigning Array Variable
+            else if (ASR::is_a<ASR::ArrayItem_t>(*ASRUtils::unwrap_logical_byte_cast(target))) {
+                ASR::ArrayItem_t *array_item = ASR::down_cast<ASR::ArrayItem_t>(ASRUtils::unwrap_logical_byte_cast(target));
                 if (ASR::is_a<ASR::Real_t>(*array_item->m_type)){
                     current_variable_type_ = array_item->m_type;
                 }
@@ -4510,7 +4511,9 @@ public:
         }
         if (ASR::is_a<ASR::Cast_t>(*target)) {
             ASR::Cast_t* cast = ASR::down_cast<ASR::Cast_t>(target);
-            if (cast->m_kind == ASR::cast_kindType::ComplexToReal) {
+            if (cast->m_kind == ASR::cast_kindType::LogicalByteToLogical) {
+                target = cast->m_arg;
+            } else if (cast->m_kind == ASR::cast_kindType::ComplexToReal) {
                 /*
                     Case: x%re = y
                     we do: x = cmplx(y, x%im)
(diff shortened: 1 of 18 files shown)


krystophny:
## Architectural Note

This PR implements the foundation for the ASR-level approach @certik requested in #9559 and #9626.

### Key Design Decisions

1. **ArrayItem stays Logical** - Unlike PR #9798 which changed ArrayItem type to Integer (which certik objected to), this approach keeps ArrayItem as Logical type.

2. **New cast kinds, not Integer casts** - We define `LogicalByteToLogical` and `LogicalToLogicalByte` instead of using `IntegerToLogical`/`LogicalToInteger`. This explicitly represents the byte storage boundary without implying the ASR type is Integer.

3. **Foundation, not full implementation** - The casts are defined and handled, but not yet inserted during AST→ASR. The existing codegen (`get_el_type()` + direct i8↔i1 conversion) still works.

### Why not insert casts automatically?

The challenge is that `ArrayItem` is used for both:
- **Read (rvalue)**: Needs `Cast(ArrayItem, LogicalByteToLogical)` 
- **Write (lvalue)**: Target stays as ArrayItem, value needs `Cast(value, LogicalToLogicalByte)`

Automatically wrapping in `make_ArrayItem_t_util` breaks assignment targets. A proper implementation would need to:
1. Track whether ArrayItem is used as lvalue vs rvalue
2. Handle the different cases in semantics

This is more involved, so this PR just provides the foundation. The full implementation can be done incrementally.

### Alternative considered

I also tried using LLVM's `freeze` instruction with i1 storage, but it still fails under `--fast` on ARM64. The i8 storage is necessary (see my analysis in PR #9798).

krystophny:
Closing in favor of the existing codegen-only approach on main.

The LogicalByteToLogical/LogicalToLogicalByte ASR cast approach creates a viral unwrapping problem: make_ArrayItem_t_util wraps every logical array element in Cast, then make_Assignment_t_util immediately unwraps it for targets and re-wraps the value. The array_op pass needs a recursive walker to fix assignments. 15+ call sites need unwrap_logical_byte_cast(). Any future code checking is_a<ArrayItem_t> on a logical array silently breaks until someone adds the unwrap. Meanwhile C++, Julia, and WASM backends handle these casts as no-ops, proving they are not semantically meaningful.

The 3 codegen workarounds on main (store i1->i8 ZExt, load i8->i1 ICmpNE, function arg bitcast) are correct, localized to the one backend that needs them, and work with both normal and --fast. All logical array tests pass on main.

Also closing #9798 for the same reasoning.

PR #9854, opened 2026-02-03
fpm: build the latest main without workarounds

certik:
Depends on https://github.com/lfortran/lfortran/pull/9853.


certik:
If tests pass, it should be squashed merged.

PR #9861, opened 2026-02-03
fix: create pointer type for `intent(inout)` cptr arg

jinangshah21:
Towards #9858


jinangshah21:
Created #9866 as exhaustive checks were not working here.

PR #9871, opened 2026-02-04
fix: load cptr even for externalUndefined abi procedure

### LINKED ISSUE #9870: Handle character pointer array inside call with separate compilation
MRE:
`a.f90`:
```fortran
module mre_strings
    use iso_c_binding, only: c_ptr, c_char, c_f_pointer
    implicit none

    interface f_string
        module procedure f_string_cptr
    end interface f_string

contains

    function f_string_cptr(cptr) result(s)
        type(c_ptr), intent(in), value :: cptr  ! Note: value attribute
        character(len=:,kind=c_char), allocatable :: s
        character(len=1,kind=c_char), pointer :: sptr
        call c_f_pointer(cptr, sptr)
        s = sptr
    end function f_string_cptr

end module mre_strings
```

`b.f90`:
```fortran
module mre_filesystem
    use iso_c_binding, only: c_ptr
    use mre_strings, only: f_string
    implicit none

    interface
        function get_ptr() result(r) bind(c)
            import c_ptr
            type(c_ptr) :: r
        end function get_ptr
    end interface

contains

    subroutine test()
        s = f_string(get_ptr())  ! <-- Error here
        print *, s
    end subroutine test

end module mre_filesystem
```

commands: 
```console
lfortran -c a.f90
lfortran --separate-compilation -c b.f90
```

LFortran Error:
```console
  %lfortran_tmp_DAUrdjYYMPpDJ1sXQn9MWc = alloca void*, align 8
 void*  call void @__module_mre_strings_f_string_cptr(void** %lfortran_tmp_DAUrdjYYMPpDJ1sXQn9MWc, %string_descriptor* %__libasr__created__var__0_return_slot_DAUrdjYYMPpDJ1sXQn9MWc)
```

jinangshah21:
Fixes #9870

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2482,6 +2482,7 @@ RUN(NAME external_10 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME external_11 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME external_12 LABELS gfortran llvm EXTRAFILES external_12_module.f90)
 RUN(NAME external_13 LABELS gfortran llvm EXTRA_ARGS --implicit-interface --mangle-underscore-external)
+RUN(NAME external_14 LABELS gfortran llvm EXTRAFILES external_14_module.f90)
 
 
 RUN(NAME interface_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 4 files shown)


certik:
The change is fine, but the tests are not registered.

certik:
You can merge it and register in the next PR, if you are sure that the fix works. It doesn't break anything, so you can merge as is to unblock us.

Issue #4925, opened 2024-09-27
ES11.2E4 treated as ES11.2E2


harperjf:
This program is part of an effort to modify my program that finds properties of real and integer kinds so that lfortran will
accept it. It already works well with 4 other compilers and finds bugs in 2 more, including lfortran which gives a wrong answer for ```tiny``` and a seg.fault with ```huge```, partly because it treats ES11.2E4 format as if it were ES11.2E2. It looks like the opposite bug to the one reported in #4530.
```
program estest
  implicit none
  integer,parameter:: sp = kind(1e0), dp = kind(1d0),&
       sep = selected_real_kind(precision(1.0_dp)+1),&
       ep = merge(sep,dp,sep>0) ! extended precision if compiler allows
  real(sp),parameter:: xsp = 1
  real(dp),parameter:: xdp = 1 
  real(ep),parameter:: xep = 1
  write(*,"(A)")            'kind             sp         dp         ep'
  write(*,"(A,3(ES11.2E4))")'epsilon:',epsilon(xsp),epsilon(xdp),epsilon(xep)
  write(*,"(A,3(ES11.2E4))")'tiny   :',   tiny(xsp),   tiny(xdp),   tiny(xep)
  write(*,"(A,3(ES11.2E4))")'huge   :',   huge(xsp),   huge(xdp),   huge(xep)
end program estest
```
Output from lfortran 0.40.0, in which dp == ep:
```
kind             sp         dp         ep
epsilon:   1.19E-07   2.22E-16   2.22E-16
tiny   :   1.18E-38   0.00E-01   0.00E-01
Segmentation fault (core dumped)
```
Output from gfortran or g95, in which dp /= ep:
```
kind             sp         dp         ep
epsilon: 1.19E-0007 2.22E-0016 1.08E-0019
tiny   : 1.18E-0038 2.23E-0308 3.36E-4932
huge   : 3.40E+0038 1.80E+0308 1.19E+4932
```
Output from ifx or ifort, in which ep is different from gfortran's:
```
kind             sp         dp         ep
epsilon: 1.19E-0007 2.22E-0016 1.93E-0034
tiny   : 1.18E-0038 2.23E-0308 3.36E-4932
huge   : 3.40E+0038 1.80E+0308 1.19E+4932
```
Output from AMD flang, revealing bugs both like and different from lfortran's:
```
kind             sp         dp         ep
epsilon:   1.19E-07   2.22E-16   1.93E-34
[...]

PR #9872, opened 2026-02-04
fix: preserve non-constant expressions in character length declarations

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4925


certik:
Let's run full tests.

PR #9873, opened 2026-02-04
fix:  select_type

yuvrajsinghgmx:
Description of Changes :

- `ast_body_visitor`   
        - Removed forced selector type to always be a pointer
        - Assign the type to  `assoc_variable` dynamically.

- `asr_to_llvm` 
      -Enable passing of a single array item to ASSERT
      -Fix size calculation for  individual objects

- `utils`
     - Removed strict `size` check


certik:
Before you submit a PR, make sure to run integration tests locally. If they fail, fix them first, to save our CI.

Issue #4748, opened 2024-09-03
Compile neural-fortran


certik:
https://github.com/modern-fortran/neural-fortran

Issues:
* [ ] https://github.com/lfortran/lfortran/issues/4745

gxyd:
With the latest main, this repository fails with an error with MRE described in https://github.com/lfortran/lfortran/issues/4759

Pranavchiku:
```fortran
module nf_activation
    type :: softmax
    contains
      procedure :: eval_1d
    end type softmax
  
  contains
  
    pure function eval_1d(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
    end function eval_1d
  
    pure function eval_1d_prime(self, x) result(res)
      class(softmax), intent(in) :: self
      real, intent(in) :: x(:)
      real :: res(size(x))
      res = self%eval_1d(x)
    end function eval_1d_prime
  
end module nf_activation
  
program main
end program main

```

With https://github.com/lfortran/lfortran/pull/4828 it gives:

```console
(lf) pranavchiku@pranavchikus-MacBook-Air:~/repos/lfortran(gh4759) % lfortran a.f90 --show-llvm
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2624, in main
    return main_app(argc, argv);
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 2466, in main_app(int, char**)
    return emit_llvm(arg_file, lfortran_pass_manager,
  File "/Users/pranavchiku/repos/lfortran/src/bin/lfortran.cpp", line 928, in (anonymous namespace)::emit_llvm(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, LCompilers::PassManager&, LCompilers::CompilerOptions&)
    = fe.get_llvm(input, lm, pass_manager, diagnostics);
  File "/Users/pranavchiku/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 298, in LCompilers::FortranEvaluator::get_llvm(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, LCompilers::LocationManager&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&)
[...]

PR #9877, opened 2026-02-04
fix: handle class default statements with assoc variables

### LINKED ISSUE #9876: Handle assoc variable with class default in select type statements
MRE
```fortran
module m
   implicit none
   type :: t
   contains
      procedure :: init
   end type
contains
   subroutine init(self)
      class(t), intent(inout) :: self
      print *, "init"
   end subroutine
end module

program p
   use m
   class(t), allocatable :: x
   allocate(t :: x)

   select type(y => x)
    class default
      call y%init()
   end select
end program
```

LFortran
```console
semantic error: Variable 'y' is not declared
  --> /app/example.f90:21:7
   |
21 |       call y%init()
   |       ^^^^^^^^^^^^^ 'y' is undeclared


```

GFortran
```console
init
```

HarshilShah1804:
Fixes #9876
Towards #4748

### DIFF
--- a/tests/reference/asr-derived_types_19-26385d4.json
+++ b/tests/reference/asr-derived_types_19-26385d4.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-derived_types_19-26385d4.stdout",
-    "stdout_hash": "915e420d37dbf77016d6ea858d95ca292b0d34325bc019f63ac61edc",
+    "stdout_hash": "68bce5f0ae51a2b42182fc0d6afa2cd716acb34b830a0a82e62b4a7d",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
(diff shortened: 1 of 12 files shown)


certik:
Can you describe what this change does?

It seems it is wrapping the insides of select type to a block? Why is that needed? I am fine either way, just want to understand the motivation and design.

HarshilShah1804:
This is done to ensure proper scope isolation, since each select type branch has its own local scope and the associate variable behaves like a local variable within that scope. Therefore, we introduce a Block to correctly model this behavior.

To clarify, the primary change in the PR is not the addition of the block itself, but the introduction of an Associate statement specifically within the class default branch’s block which allows the use of associate variable which was previously not getting recognized.

Additionally, this approach maintains consistency with the existing implementation of other select type branches, such as `type is ()` and `class is ()`, which are already handled by creating a block for each branch.

PR #9881, opened 2026-02-05
CI: test fpm with `--separate-compilation`

certik:


certik:
It works, but fpm tests now take 23 minutes...

PR #9882, opened 2026-02-05
CI: use Release mode for exhaustive tests

certik:


jinangshah21 (approved):

certik:
It seems the FPM tests went from 23 min to 7 min. Overall things got faster too. So let's merge it.

PR #9888, opened 2026-02-05
refactor: consolidate logical array i8/i1 codegen into helpers

krystophny:
## Summary
- Extract three scattered i8/i1 conversion workarounds into two helper methods
- No behavioral change; all tests pass identically

Refs #9626

## Why
The logical array i8 storage (vs i1 scalar) boundary was handled by three inline workaround blocks scattered across `asr_to_llvm.cpp`. This consolidates them into two helpers (`logical_store_val`, `logical_load_val`) and simplifies the function argument bitcast, making the pattern explicit and easy to find.

This replaces the ASR-level approaches in #9798 and #9851, which created a viral unwrapping problem affecting all backends for what is purely an LLVM concern.

**Stage:** Codegen (LLVM only)

## Changes
- [`asr_to_llvm.cpp#L8110-L8131`](https://github.com/lfortran/lfortran/blob/3611516d5f5be0a62a4c0e0f16255fd504fe5151/src/libasr/codegen/asr_to_llvm.cpp#L8110-L8131): Add `logical_store_val` (i1->i8 for array store) and `logical_load_val` (i8->i1 for array load)
- [`asr_to_llvm.cpp#L8858`](https://github.com/lfortran/lfortran/blob/3611516d5f5be0a62a4c0e0f16255fd504fe5151/src/libasr/codegen/asr_to_llvm.cpp#L8858): Replace inline store workaround with `logical_store_val`
- [`asr_to_llvm.cpp#L9777`](https://github.com/lfortran/lfortran/blob/3611516d5f5be0a62a4c0e0f16255fd504fe5151/src/libasr/codegen/asr_to_llvm.cpp#L9777): Replace inline load workaround with `logical_load_val`
- [`asr_to_llvm.cpp#L16193`](https://github.com/lfortran/lfortran/blob/3611516d5f5be0a62a4c0e0f16255fd504fe5151/src/libasr/codegen/asr_to_llvm.cpp#L16193): Simplify function arg bitcast block

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8107,6 +8107,28 @@ public:
         this->visit_expr(*x.m_overloaded);
     }
 
+    // Logical arrays use i8 storage; scalar logicals use i1.
+    // These two helpers handle the boundary conversions in one place.
+    llvm::Value* logical_store_val(llvm::Value* v) {
+        if (v->getType()->isIntegerTy(1))
+            return builder->CreateZExt(v, llvm::Type::getInt8Ty(context));
+        return v;
+    }
+
+    llvm::Value* logical_load_val(llvm::Value* ptr, ASR::expr_t* x,
+                                  bool is_volatile = false) {
+        if (x->type == ASR::exprType::ArrayItem &&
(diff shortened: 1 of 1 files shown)


krystophny:
@certik after much back and forth, I believe this is the right solution and not a hack. What do you think?

PR #9914, opened 2026-02-06
CI: test fpm with `--fast`

certik:
Depends on #9913.


certik:
@Pranavchiku you can't do automerge, because that doesn't wait for the exhaustive tests, which are the only ones that can fail here...

So we need to wait until everything passes.

The good news is that it seems it is working at the CI for LLVM 21, so we just need to see if it works for every other version also.

PR #9897, opened 2026-02-05
fix: externalize struct type in select type class default

krystophny:
## Summary
- Fix modfile deserialization crash (assertion failure) when `class default` in `select type` references polymorphic types from external modules

Fixes #9896

## Why
When `select type` has a `class default` clause on a polymorphic component from an external module (e.g., `class(base_t), allocatable`), the association variable's `m_type_declaration` was set to the internal struct symbol directly, bypassing `ExternalSymbol`. During modfile serialization, this wrote a symtab ID from the external module. During deserialization in another compilation unit, that symtab ID was not present in the deserializer's `id_symtab_map`, causing an assertion failure.

**Stage:** Semantics

## Changes
- [`src/lfortran/semantics/ast_body_visitor.cpp#L3264-L3268`](https://github.com/lfortran/lfortran/blob/fcdc56cf1103eeea70027681fc6f9d8991dc25d2/src/lfortran/semantics/ast_body_visitor.cpp#L3264-L3268): Use `import_struct_type()` to externalize the type declaration symbol in the `ClassDefault` case, matching how `ClassStmt` and `TypeStmtName` branches already handle their symbols

## Tests
- [`integration_tests/modules_65.f90`](https://github.com/lfortran/lfortran/blob/fcdc56cf1103eeea70027681fc6f9d8991dc25d2/integration_tests/modules_65.f90): Multi-module test with `select type` + `class default` on polymorphic external type

### Test fails on main
```
$ git checkout upstream/main
$ cd /tmp/mre_9896 && rm -f *.o *.mod
$ lfortran -c mre_serialization_crash_types.f90
$ lfortran -c mre_serialization_crash_main.f90
LCOMPILERS_ASSERT failed: /home/ert/code/lfortran-dev/lfortran/src/libasr/serialization.cpp
function read_symbol(), line number 133 at
id_symtab_map.find(symtab_id) != id_symtab_map.end()
Aborted (core dumped)
```

### Test passes after fix
```
$ git checkout fix/modfile-select-type-9896
$ cd /tmp/mre_9896 && rm -f *.o *.mod
$ lfortran -c mre_serialization_crash_types.f90
$ lfortran -c mre_serialization_crash_main.f90
$ # No error, exit code 0

$ scripts/lf.sh itest -b llvm -t modules_65
1/1 Test #1040: modules_65 .......................   Passed    0.00 sec
100% tests passed, 0 tests failed out of 1

$ scripts/lf.sh itest -b gfortran -t modules_65
[...]

PR #9915, opened 2026-02-06
fix: use structural flag for unlimited polymorphic type checks

krystophny:
## Summary

- Replace ~15 scattered `"~unlimited_polymorphic_type"` string comparisons with the existing `StructType_t::m_is_unlimited_polymorphic` flag
- Fix serialization crash in `import_struct_type` where a foreign module's unlimited polymorphic symbol was returned as-is, causing a dangling symtab reference during modfile deserialization
- Fix copy-paste bug in `check_class_assignment_compatibility` that checked the target name twice instead of target vs value

Follow-up to #9897 (certik review comment about avoiding string special cases).

## Why

PR #9897 fixed a serialization crash for bounded polymorphic types (`class(base_t)`) by externalizing struct references. However, it added a string-based special case for `~unlimited_polymorphic_type` that returned the foreign symbol as-is. When a third module deserializes the second module's modfile, the foreign symbol's symtab ID is not in the local `id_symtab_map`, causing an assertion failure.

**Stage:** Semantics + Codegen (string cleanup)

## Changes

- [`src/libasr/asr_utils.h`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/src/libasr/asr_utils.h): Add `is_unlimited_polymorphic_type(Struct_t*)` overload; rewrite `symbol_t*` and `expr_t*` overloads to use structural flag; fix `import_struct_type` to create local unlimited poly type instead of returning foreign symbol; fix `is_derived_type_similar` and `check_class_assignment_compatibility`
- [`src/libasr/codegen/asr_to_llvm.cpp`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/src/libasr/codegen/asr_to_llvm.cpp): Replace 2 string comparisons with structural checks
- [`src/libasr/codegen/llvm_utils.cpp`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/src/libasr/codegen/llvm_utils.cpp): Replace 5 string comparisons with structural checks
- [`src/lfortran/semantics/ast_symboltable_visitor.cpp`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/src/lfortran/semantics/ast_symboltable_visitor.cpp): Replace 2 name-based checks with structural checks
- [`src/lfortran/semantics/ast_common_visitor.h`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/src/lfortran/semantics/ast_common_visitor.h): Simplify; one string check remains because `v` is a placeholder `ExternalSymbol_t` with null external at this creation point

## Tests

- [`integration_tests/modules_66.f90`](https://github.com/lfortran/lfortran/blob/ff97968bb51ce2dd8427120636984366fdc9d495/integration_tests/modules_66.f90): Three-module test with `class(*)` in external module, `select type` with `class default` in worker module, main program using worker -- triggers the serialization crash path

### Test fails on main
```
$ git checkout upstream/main
$ cd integration_tests && lfortran --separate-compilation modules_66_base.f90 modules_66_worker.f90 modules_66.f90 -o /tmp/m66_test
LCOMPILERS_ASSERT failed: src/libasr/serialization.cpp
function read_symbol(), line number 133 at 
id_symtab_map.find(symtab_id) != id_symtab_map.end()
Aborted (core dumped)
```

### Test passes after fix
```
$ git checkout fix/unlimited-poly-import-struct
$ cd integration_tests && lfortran --separate-compilation modules_66_base.f90 modules_66_worker.f90 modules_66.f90 -o /tmp/m66_test && /tmp/m66_test
PASSED: modules_66

$ scripts/lf.sh itest -b llvm -t modules_66

$ scripts/lf.sh test
TESTS PASSED
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1522,6 +1522,8 @@ RUN(NAME modules_63 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc
 RUN(NAME modules_64 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME modules_65 LABELS gfortran llvm
         EXTRAFILES modules_65_base.f90 modules_65_child.f90 modules_65_worker.f90)
+RUN(NAME modules_66 LABELS gfortran llvm
+        EXTRAFILES modules_66_base.f90 modules_66_worker.f90)
 
 RUN(NAME associate_06 LABELS gfortran EXTRAFILES
         associate_06_module.f90)
--- a/integration_tests/modules_66.f90
+++ b/integration_tests/modules_66.f90
@@ -1,0 +1,14 @@
+program modules_66
+    use modules_66_base, only: container_t
+    use modules_66_worker, only: extract_int
+    implicit none
+
+    type(container_t) :: c
+    integer :: r
+
+    allocate(c%item, source=42)
+    r = extract_int(c)
+    if (r /= 42) error stop
+
+    print *, "PASSED: modules_66"
+end program modules_66
--- a/integration_tests/modules_66_base.f90
+++ b/integration_tests/modules_66_base.f90
@@ -1,0 +1,8 @@
+module modules_66_base
+    implicit none
+
+    type :: container_t
+        class(*), allocatable :: item
+    end type container_t
+
+end module modules_66_base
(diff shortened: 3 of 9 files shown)


certik:
Much better! Thanks for the refactoring.

Issue #8029, opened 2025-07-11
Segfault passing a string via a class(*) pointer


harperjf:
This program from the Intel Fortran Forum 07-09-2025 ran correctly with gfortran but caused a segfault with lfortran and gave wrong run-time output with ifx.
```
module m1
  character(:), pointer, private :: string => null()
contains
  function get()
    class(*), pointer :: get
    if (.not.associated(string)) allocate(string, source='foo')
    get => string
  end function
  subroutine set(arg)
    class(*), intent(in) :: arg
    select type (arg)
    type is (character(*))
      print *, 'set received string "', arg, '"'
      if (arg /= 'foo') stop 'ERROR!'
    end select
  end subroutine
end module

use m1
select type (s => get())
type is (character(*))
  print *, 'passing string "', s, '" to set'
  call set(s)
end select
end
```
Gfortran 15.1.0 output (correct):
```
 passing string "foo" to set
 set received string "foo"
```
Lfortran 0.54.0 output:
```
Segmentation fault      (core dumped)
```
ifx 2025.2 output (wrong):
```
 passing string "foo" to set
 set received string ""
ERROR!
```

PR #9924, opened 2026-02-07
fix: add support for association of polymorphic pointer variable to character variable

### LINKED ISSUE #9923: Handle association from polymorphic class pointers to character vairables
Related to #8029
MRE:
```
program a
  implicit none
  ! Declare an unlimited polymorphic pointer
  class(*), pointer :: ptr
  character(5), target :: t
  t = 'abcde'
  ptr => t
  select type (p => ptr)
    type is (character(*))
      print *, p
    class default
      print *, "Unknown"
      error stop
  end select
end program a
```

Gfortran:
```
$ gfortran a.f90 & ./a.out
[1] 141926
 abcde
```

Main:
``` $ lfortran a.f90
File "/home/yash/Desktop/lfortran/lfortran/src/libasr/../libasr/asr_base_visitor.h", line 58, in void LCompilers::ASR::visit_stmt_t<LCompilers::ASRToLLVMVisitor>(LCompilers::ASR::stmt_t const&, LCompilers::ASRToLLVMVisitor&)
    case stmtType::Associate: { v.visit_Associate((const Associate_t &)x); return; }
AssertFailed: ASRUtils::is_string_only(target_type)
```

YashNagda17:
Fixes #9923 
Towards #8029 

Changes made:
- For strings declared inside select type block associates, change it's type from AssumedLength to DeferredLength.
- Add support for association of character variables to polymorphic class pointer variables, and vice-versa.
- Update tests for #9700  to make them more robust, by checking for variable value change using associate alias variable

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -820,6 +820,8 @@ RUN(NAME pointer_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME pointer_03 LABELS gfortran llvm)
 RUN(NAME pointer_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME pointer_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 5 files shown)


certik:
Let's see if all tests pass. Then I'll review one more time carefully.

PR #9927, opened 2026-02-08
Fix symbol collision handling in pass_array_by_data

### LINKED ISSUE #9909: Separate compilation segfaults on code with user-defined constructors
The following is a show-stopper bug for me (i.e. of high priority to compile an OO code base). Separate compilation fails on the following, valid, code that is split over two files:
`$ cat mytypemod.f90`
```
module mytypemod
   
   implicit none
   private

   type, public, abstract :: AbsType
   end type AbsType

   type, public :: MyType
      integer, allocatable :: ints(:)
   end type MyType

   interface MyType
      procedure :: constructor
   end interface MyType

contains

   function constructor(ints,obj) result(self)
      integer,        intent(in) :: ints(:)
      self%ints = ints
      self%obj  = obj
   end function constructor

end module mytypemod
```
`$ cat clientmod.f90`
```
module clientmod

   use mytypemod, only: AbsType, MyType

   implicit none
   private

   type, public :: Client
   contains
      procedure :: method
   end type Client

   interface Client
      procedure :: constructor
   end interface Client
   
contains

   function constructor(b) result(self)
      integer, intent(in) :: b(:)
      type(Client)        :: self
   end function constructor
   
   
   subroutine method(self,arr)
      class(Client),              intent(in)  :: self
      class(MyType), allocatable, intent(out) :: arr(:)

      integer,        allocatable :: ints(:)

      allocate(MyType :: arr(1))
      arr = MyType(ints,obj)      

   end subroutine method

end module clientmod
```
Trying to compile this with
`$ lfortran -c --separate-compilation  mytypemod.f90`
`$ lfortran -c --separate-compilation  clientmod.f90`
gives:
```
```

certik:
Fix pointer mismatches.

Fixes #9909.


certik:
So the commit b98c8f981afd2f416199308dbdde2973e43e16cc is one way to fix it, but the `coerce_call_args_to_function_type()` function doesn't belong into the backend, it should be in ASR with explicit casts. So let's wait with merging this PR, let's see if we can improve the fix. An improved version is in https://github.com/lfortran/lfortran/pull/9931.

certik:
Fixed in https://github.com/lfortran/lfortran/pull/9931.

PR #9931, opened 2026-02-08
Fix symbol collision handling in pass_array_by_data

certik:
This PR (alternative to https://github.com/lfortran/lfortran/pull/9927) fixes two issues. First, it fixes a symbol-collision bug in `pass_array_by_data`. After that fix, a deeper type-conversion bug becomes visible: passing `ArrayItem(class(T) array)` to a formal argument expecting `type(T)`.

In our lowering model, `class(T)` is represented as a wrapper (roughly `{ vptr, T* }`), while `type(T)` expects direct `T*` storage. So this call path requires an explicit ASR cast from class-wrapper to struct data. We did not have such a cast before.

This PR introduces `Cast(ClassToStruct)`, implements its LLVM lowering, and inserts it in the currently affected rewrite paths (`pass_array_by_data` and `array_passed_in_function_call`).

Follow-up work: make cast insertion complete for all pass-generated cases (including any frontend-emitted cases where applicable) and enforce the invariant in ASR verification.

Fixes https://github.com/lfortran/lfortran/issues/9909.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -12664,7 +12664,7 @@ public:
                     builder->CreateStore(class_value, class_tmp);
                     class_value = class_tmp;
                 }
-                #if LLVM_VERSION_MAJOR >= 17
+                #if LLVM_VERSION_MAJOR >= 15
                 // Opaque pointers collapse pointer-depth information, so
                 // type-based while loops can become non-terminating.
                 if ((ASRUtils::is_allocatable(ASRUtils::expr_type(x.m_arg)) ||
@@ -12690,7 +12690,7 @@ public:
                 llvm::Type* target_base_type = llvm_utils->get_type_from_ttype_t_util(
                     const_cast<ASR::expr_t*>(&x.base), target_type, module.get());
                 llvm::Type* target_ptr_type = target_base_type->getPointerTo();
-                #if LLVM_VERSION_MAJOR < 17
+                #if LLVM_VERSION_MAJOR < 15
                 while (tmp->getType() == target_ptr_type->getPointerTo()) {
                     tmp = llvm_utils->CreateLoad2(target_ptr_type, tmp);
                 }


certik:
@jinangshah21 can you please review this PR? How do we currently handle this class to struct cast? I know we were facing exactly this issue before --- do we currently only handle it in the LLVM backend? So this PR does the first step in refactoring it into the ASR? Or do we already do it in ASR somehow, in which case this PR should use that mechanism?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12688,7 +12632,74 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                #if LLVM_VERSION_MAJOR < 15
+                while (tmp->getType() == target_ptr_type->getPointerTo()) {
+                    tmp = llvm_utils->CreateLoad2(target_ptr_type, tmp);
+                }
+                #endif
+                if (tmp->getType() != target_ptr_type) {
+                    tmp = builder->CreateBitCast(tmp, target_ptr_type);
+                }
+                break;
+            }
```
certik, line 12702: I think this part can be improved, we need to look into `convert_class_to_type()`, `llvm_utils->getStructType()`, etc. and see if one of them cannot be used here.

jinangshah21:
Reviewing it now.

certik:
@jinangshah21 to get things moving, I am going to merge this. I think the approach is quite clean with the ASR cast. We now need to extend it to insert the cast appropriately at all places and possibly simplify the LLVM implementation of the cast.

certik:
Thanks for the review. I made a follow up issue in https://github.com/lfortran/lfortran/issues/9950 for the cast handling.

src/libasr/codegen/asr_to_llvm.cpp
certik: Fixed in https://github.com/lfortran/lfortran/pull/9951.

Issue #9916, opened 2026-02-06
Wrong output for real with '(g0)' descriptor


edvinolo:
The following program, when compiled with lfortran 0.59, prints the wrong output to the terminal
```fortran
program t
write (*,'(g0)') 1.0
end program t
```
The program outputs ```0E+1```, which is clearly incorrect.

CE reproducer: https://gcc.godbolt.org/z/f4r473d9c

certik:
Nice one!  Thanks for the report.
```console
$ lfortran b.f90
0E+1
$ gfortran b.f90 && ./a.out
1.00000000
```

PR #9934, opened 2026-02-08
fix: handle G0 format descriptor for real values

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9916


integration_tests/format_61.f90
```diff
@@ -0,0 +1,21 @@
+program format_60
```
certik, line 1:
```suggestion
program format_61
```

PR #9946, opened 2026-02-09
enh: handle nested select type blocks with selector details in map

### LINKED ISSUE #9908: Separate compilation fails for `select type` statement
Separate compilation fails on the `select type` statement of the following, valid, code:
`$ cat test.f90`
```
module testmod
   
   type, abstract :: AbsType
   contains
      procedure(add), deferred :: add
   end type AbsType

      real(8), dimension(:,:), allocatable :: arr
   contains
      procedure :: add => addition
   end type MyType

   abstract interface
      function add(a,b) result (r)
         import
         class(AbsType), intent(in)  :: a, b
         class(AbsType), allocatable :: r
      end function add
   end interface

contains

   function addition(a,b) result(r)
      class(MyType),  intent(in)  :: a
      class(AbsType), intent(in)  :: b
      class(AbsType), allocatable :: r      
      select type (b)
      type is (MyType)
         allocate(MyType :: r)
         select type (r)
         type is (MyType)
            r%arr = a%arr + b%arr
         end select
      end select
   end function addition

end module testmod
```
The following wrong error message is produced in this case:
```
code generation error: abstype doesn't have any member named arr
  --> test.f90:35:29
   |
35 |             r%arr = a%arr + b%arr
   |                             ^^^^^ 


```

jinangshah21:
Fixes #9908

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -248,40 +248,49 @@ public:
     std::map<ASR::symbol_t*, std::map<std::string, int64_t>> vtabtype2procidx;
     std::map<ASR::symbol_t*, std::map<std::string, int64_t>> struct_vtab_function_offset;
 
-    // Context for a select type block - maps selector variable name to its type info
+    // Context for a select type block - maps selector variable symbol to its type info
     struct SelectTypeContext {
         llvm::Type* block_type = nullptr;
         ASR::ttype_t* block_type_asr = nullptr;
         std::string der_type;
         ASR::symbol_t* selector_type_decl = nullptr;
     };
-    // Map from selector variable name to its select type context
-    std::map<std::string, SelectTypeContext> select_type_context_map;
+    // Map from selector variable symbol to its select type context
+    std::map<ASR::symbol_t*, SelectTypeContext> select_type_context_map;
 
     // Helper methods to access select type context for a specific variable
-    SelectTypeContext* get_select_type_context(const std::string& var_name) {
-        auto it = select_type_context_map.find(var_name);
+    SelectTypeContext* get_select_type_context(ASR::symbol_t* sym) {
+        if (sym == nullptr) return nullptr;
+        auto it = select_type_context_map.find(sym);
         return (it != select_type_context_map.end()) ? &it->second : nullptr;
     }
 
-    llvm::Type* get_select_type_block_type(const std::string& var_name) {
-        auto ctx = get_select_type_context(var_name);
+    // Helper to extract symbol from expression (returns nullptr if not Var_t)
+    ASR::symbol_t* get_symbol_from_expr(ASR::expr_t* expr) {
+        if (expr && ASR::is_a<ASR::Var_t>(*expr)) {
+            return ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::Var_t>(expr)->m_v);
+        }
+        return nullptr;
+    }
+
+    llvm::Type* get_select_type_block_type(ASR::expr_t* expr) {
+        auto ctx = get_select_type_context(get_symbol_from_expr(expr));
         return ctx ? ctx->block_type : nullptr;
     }
 
-    ASR::ttype_t* get_select_type_block_type_asr(const std::string& var_name) {
-        auto ctx = get_select_type_context(var_name);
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -15443,12 +15484,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
         if (ASRUtils::is_unlimited_polymorphic_type(expr)) {
-            if (current_select_type_block_type_asr) {
-                type = current_select_type_block_type_asr;
-            } else if (!current_select_type_block_der_type.empty()) {
+            std::string var_name = ASRUtils::EXPR2VAR(expr)->m_name;
+            ASR::ttype_t* ctx_type_asr = get_select_type_block_type_asr(var_name);
+            const std::string& ctx_der_type = get_select_type_block_der_type(var_name);
+            if (ctx_type_asr) {
+                type = ctx_type_asr;
+            } else if (!ctx_der_type.empty()) {
```
certik, line 15513: Code like this should not be in the LLVM backend. We should not be using any strings to look up things.
We should not be using strings as keys to store metadata in the LLVM backend. Because that forces you to then use strings to look it up, but strings are not unique, especially if they are derived from `ASRUtils::EXPR2VAR(expr)->m_name`.

Rather, one has to use a pointer (hash) for the given ASR node and attach metadata to it.

The LLVM backend should only use pointers (which are fast) and lookup associated metadata. We need to improve our ASR to store all the information that is needed locally in each node, or a simple pointer lookup from there (for example in the symbol table).

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -247,11 +247,44 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+    // Context for a select type block - maps selector variable name to its type info
+    struct SelectTypeContext {
+        llvm::Type* block_type = nullptr;
+        ASR::ttype_t* block_type_asr = nullptr;
+        std::string der_type;
+        ASR::symbol_t* selector_type_decl = nullptr;
+    };
+    // Map from selector variable name to its select type context
+    std::map<std::string, SelectTypeContext> select_type_context_map;
```
certik: It doesn't seem right to use strings as keys in this map. See https://github.com/lfortran/lfortran/pull/9946/changes#r2783395777.

Let's figure out a way to use pointers or hashes. See how we already do this elsewhere:
```c++
            uint32_t h = get_hash((ASR::asr_t*)asr_target);
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -17992,8 +18036,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
 
             // Convert function args
             std::vector<llvm::Value*> args;
-            if (current_select_type_block_type && ASR::is_a<ASR::Var_t>(*x.m_dt) &&
-                    ASRUtils::EXPR2VAR(x.m_dt)->m_name == current_selector_var_name) {
+            std::string _dt_var_name = ASR::is_a<ASR::Var_t>(*x.m_dt) ? ASRUtils::EXPR2VAR(x.m_dt)->m_name : "";
+            llvm::Type* ctx_block_type = get_select_type_block_type(_dt_var_name);
+            const std::string& ctx_der_type = get_select_type_block_der_type(_dt_var_name);
+            if (ctx_block_type && ASR::is_a<ASR::Var_t>(*x.m_dt)) {
```
certik: You can see this kind of code repeated throughout. Is it possible to refactor it into a function that accepts `x.m_dt` as the argument? In other words, effectively index it based on the variable itself or the symbol table entry?

certik:
So @difference-scheme said this is not a blocker, so let's fix it the right way, we don't need to rush.

jinangshah21:
Currently I have updated map to use Varibale symbol* as index.
Well proper fix is to have casts at AST --> ASR, which is next step I will be looking at

certik (approved):
I think this is good enough to merge.

certik:
Let's merge https://github.com/lfortran/lfortran/pull/9953 first, it's a higher-priority fix.

PR #9948, opened 2026-02-09
Fix: validate deferred length consistency for allocatable character dummy arguments

### LINKED ISSUE #6776: Error : Report error for not matching string array types

Jatinagarwal24:
Fixes #6776

### DIFF
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -35,8 +35,8 @@ contains
     end subroutine assumed_rank
 
     subroutine ss(x)
-        character(:), allocatable :: x
-        x(1) = "AB"
+    character(:), allocatable :: x
+    x = "AB"          ! allocates x with length 2
     end subroutine ss
 
 
--- a/tests/reference/asr-continue_compilation_3-435a232.json
+++ b/tests/reference/asr-continue_compilation_3-435a232.json
@@ -2,7 +2,7 @@
     "basename": "asr-continue_compilation_3-435a232",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_3.f90",
-    "infile_hash": "10141ffec15aee6ccd6fa16b9b3549a25ab1d0c723f667cc19d7cf70",
+    "infile_hash": "2f2ecf3f1a2aca025759e9b6b07ba8843148ba198b8b5c0beb5c4ec3",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,


tests/errors/continue_compilation_3.f90
```diff
@@ -34,10 +34,10 @@ subroutine assumed_rank(x)
         print *, (x /= [1, 2])
     end subroutine assumed_rank
 
-
-
-
-
+    subroutine ss(x)
+        character(:), allocatable :: x
+        x(1) = "AB"
```
certik: Can you write `x(1) = "AB"` ? The `x` is not an array, so this doesn't work, right?

I asked Grok, it agrees: https://grok.com/share/bGVnYWN5_891c27b5-c2ea-4dd8-a8c8-038a61d96d89 and provides examples of what you maybe intended to do here, so pick one of those.

tests/reference/asr-continue_compilation_3-435a232.stderr
```diff
@@ -439,6 +439,12 @@ semantic error: Different `character` lengths 7 and 2 in array constructor
 204 |     print *, ["welcome", "to", "lf"] .and. "contributors"  !even size diff of array element must be caught
     |                          ^^^^ 
 
+semantic error: Actual argument to allocatable or pointer dummy argument must have a deferred length type parameter if and only if the dummy has one
+   --> tests/errors/continue_compilation_3.f90:205:13
+    |
+205 |     call ss("hello")
+    |             ^^^^^^^ type mismatch at this call
```
certik: Let's improve the error message. After reading it it's not clear what I am supposed to do as a user to fix it.

Rather, let's say something like "cannot pass unallocatable string to allocatable argument" (or whatever the actual problem is) as the error description. Then the hint by the label should say "make this expression allocatable" (if this is the right fix).

certik:
After you fix it, go ahead and click on "Ready for review".

Jatinagarwal24:
Hi,@certik 
Can you pls review this.

PR #9910, opened 2026-02-06
Fix polymorphic array element addressing

yuvrajsinghgmx:
Fixes TEST 2 from #9548 

Error - Individual array element addressing was not implemented which caused test failure.

Description of Fix: 

------

` llvm_array_utils` -  Previously static size was assigned to each array element which is wrong for polymorphic arrays 

Fix - Calculates dynamic size of the elements 

-------

`llvm_utils` - Same kind of fix as in `llvm_array_utils` for correct element addressing and new runtime API for `get_class_type_size_from_vptr`

------

`intrinsics.h` - Modify to include size in `type_info`.

PR #9949, opened 2026-02-09
Fix missing quotes and apostrophes on output in FORMAT statement

### LINKED ISSUE #9902: FORMAT statement: missing quotes and apostrophes on output
```
program apos_fmt
  implicit none

  print *, '''Hello world!'''
  print *, """Hello world!"""
  write (*,100)
  write (*,101)

100 format ('''Hello world!''')
101 format ("""Hello world!""")

end program
```
```
$ lfortran apos_fmt.f90
'Hello world!'
"Hello world!"
Hello world!
Hello world!
$
``` 
```
$ gfortran apos_fmt.f90
$ ./a.out
 'Hello world!'
 "Hello world!"
'Hello world!'
"Hello world!"
$
```

yuvrajsinghgmx:
Fixes #9902 

Description for the fix : 

1) Remove spaces from raw format string 
2) In case of `' '` this `if (cformat[index+1] == '\'')` , `index += 2` skips the pair.
3) `unescape_quoted_literal` whenever it finds `' ' `  copies only one `'` to the final result and ` "` is copied as it is because it does not match this `'`.

Fix : Comment from PR #9910

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -8926,7 +8926,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al, 
             llvm_utils->i8_ptr
         };
         std::vector<llvm::Constant*> type_info_member_values;
-        type_info_member_values.reserve(2); // A type-info object has minimum 2 member.
+        type_info_member_values.reserve(2); // A type-info object has minimum 2 members.
 
         if (struct_t->m_parent) {
             create_type_info_for_struct(struct_t->m_parent, module);


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -8795,7 +8795,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
             llvm_utils->i8_ptr
         };
         std::vector<llvm::Constant*> type_info_member_values;
-        type_info_member_values.reserve(2); // A type-info object has minimum 1 member.
+        type_info_member_values.reserve(2); // A type-info object has minimum 2 member.
```
certik:
```suggestion
        type_info_member_values.reserve(2); // A type-info object has minimum 2 members.
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -8926,7 +8926,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
```
certik:
```suggestion
        type_info_member_values.reserve(2); // A type-info object has minimum 2 members.
```

Issue #6459, opened 2025-03-02
Possibly non-standard behavior on explicit re-declaration of a variable that was previously implicitly typed


difference-scheme:
Here's a truly weird one, that I found in an old random number generator. 

The Fortran standard _appears_ to allow the following, strange, explicit "re-declaration" of a variable, that was already declared with the same type by means of implicit typing:
 ```
subroutine random( )

   implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)

   save luxlev

   integer(kind=4) ::  luxlev                                        

end subroutine random

```
I don't know what the standard truly says regarding such cases, but both gfortran, ifx, and flang-20 compile the above code without any complaints. Whereas compiling this with `lfortran -c --implicit-typing random.f90` gives:
```
semantic error: Symbol is already declared in the same scope
 --> random.f90:7:24
  |
7 |    integer(kind=4) ::  luxlev                                        
  |                        ^^^^^^ redeclaration
  |
5 |    save luxlev
  |         ~~~~~~ original declaration
```
You may wish to investigate this, because I am truly at a loss here.

**Comment by difference-scheme** [comment]
The following also fails to compile with LFortran, whereas it is accepted by the other mentioned compilers:
```
subroutine random( )

   implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)

   save notyet

   logical notyet                                                    

end subroutine random

```
So, re-declaration is apparently allowed regardless of the originally declared type! 

It seems that the "philosophy" here is to allow bulk declarations through the use of implicit typing, and then to make it possible to override such declarations by explicit ones, as required ...

harperjf:
An ```implicit``` statement saying what type a variable beginning with ```n```  will be, can be superseded by a type declaration like ```logical notyet```, but the standard seems not to care whether a ```save notyet``` statement comes before or after  the declaration (provided, as here, that they are both in the same set of specification statements).

But I would combine them as ```logical, save:: notyet```. And I think ```--std=f23``` includes ```---implicit-typing``` as well as some other things allowed by the standard but not the default for lfortran.

difference-scheme:
> but the standard seems not to care whether a save notyet statement comes before or after the declaration (provided, as here, that they are both in the same set of specification statements). The full rules on this matter are in F2023, 5.1, R508 and R513.

Thanks for pointing this out. This is what caught me by surprise. 
(It also goes to show for how long I've been avoiding these legacy features).

[...]

PR #9956, opened 2026-02-09
fix: preserve init value when explicit decl overrides implicit save variable

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6459

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -6258,7 +6258,9 @@ public:
                             variable_added_to_symtab->m_type = type;
                         }
                     } else {
-                        variable_added_to_symtab->m_type = type;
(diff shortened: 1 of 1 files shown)


integration_tests/save_12.f90
```diff
@@ -0,0 +1,7 @@
+program save_12
+   implicit real(kind=8) (a-h,o-z), integer(kind=4) (i-n)
+   save notyet
+   logical :: notyet = .true.
+   if (.not. notyet) error stop
+   print *, notyet
+end program save_12
```
certik, line 7: Note: my understanding is that this is not valid Fortran, because the implicit line and `save notyet` declares `notyet` as `n`, and thus `integer(4)`. Then later you declare it as `logical`, which conflicts. So this shouldn't compile. However GFortran, Flang and ifx all accept it. So I think we should as well.

PR #9960, opened 2026-02-09
fix: throw error for wrong delim parameter

Chaitanya140904:
Throws error for wrong value in `delim` specifier for `open` statements

### DIFF
--- a/tests/reference/run-file_01-ddd5a8e.json
+++ b/tests/reference/run-file_01-ddd5a8e.json
@@ -2,12 +2,12 @@
     "basename": "run-file_01-ddd5a8e",
     "cmd": "lfortran --no-color {infile}",
     "infile": "tests/errors/file_01.f90",
-    "infile_hash": "64fc9896fe5bb684a68ac8ecbf6fee5708a1e12b66c8244c6beefdf0",
+    "infile_hash": "df2ab80816f6bc7cb189a3da2e7e98937b98b73294f0394db3a821a0",
     "outfile": null,
     "outfile_hash": null,
     "stdout": "run-file_01-ddd5a8e.stdout",
-    "stdout_hash": "6c112c7e2df9f7ebe7ff408a2199b6d0d269f76f4c09e2bde834a6b6",
+    "stdout_hash": "41203d879b3214481570be2efa98693395e4feb916c8e440111ec40a",
     "stderr": null,
     "stderr_hash": null,
-    "returncode": 1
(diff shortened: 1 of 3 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4964,7 +4840,19 @@ _lfortran_open(int32_t unit_num,
     } else if (streql(delim_c, "quote")) {
         delim_value = 2;
     } else {
-        // TODO: Throw error
+        if (iostat != NULL) {
+            *iostat = 5002;
+            if ((iomsg != NULL) && (iomsg_len > 0)) {
+                char* temp = "DELIM specifier in OPEN statement has invalid value.";
+                snprintf(iomsg, iomsg_len, "%s", temp);
+                pad_with_spaces(iomsg, strlen(iomsg), iomsg_len);
```
certik, line 4848: Let's also add a test in integration tests with iostat to ensure this part works.

certik: This is still not tested, correct? The way to test this is to check that iostat returns non-zero. Also this message that you created in `iomsg`, that goes back to Fortran? If so, test it. If not, where does it go?

certik:
Please fix https://github.com/lfortran/lfortran/pull/9960/changes#r2794135630. Then it should be good. Click on "Ready for review" when ready.

tests/reference/run-file_01-ddd5a8e.json
```diff
@@ -2,12 +2,12 @@
...
+    "infile_hash": "df2ab80816f6bc7cb189a3da2e7e98937b98b73294f0394db3a821a0",
     "outfile": null,
     "outfile_hash": null,
     "stdout": "run-file_01-ddd5a8e.stdout",
-    "stdout_hash": "6c112c7e2df9f7ebe7ff408a2199b6d0d269f76f4c09e2bde834a6b6",
+    "stdout_hash": "41203d879b3214481570be2efa98693395e4feb916c8e440111ec40a",
     "stderr": null,
     "stderr_hash": null,
-    "returncode": 1
+    "returncode": 0
```
certik, line 12: Now this error test passes. I think that is still incorrect.

You want the error test to fail. But then you want to have an integration test that gets the right message, simply check it using `if (iomsg /= "...") error stop`.

So keep the original error test. And then add a new integration test that passes.

PR #9964, opened 2026-02-10
Fix for procedure dummy arguments captured by internal functions

### LINKED ISSUE #9963: Nested function call fails to compile
From https://fortran-lang.discourse.group/t/variable-scoping-with-nvidia-fortran-compiler-on-cpu-code-nvhpcsdk/10699:
```fortran
module my_module
  implicit none

  abstract interface
    pure real function fun_interface()
    end function fun_interface
  end interface
contains
  real function my_wrapper(fun)
    implicit none
    procedure(fun_interface) :: fun

    my_wrapper = wrapper_wrapper()

    ! This always works on GCC and on NVHPC:
    ! my_wrapper = fun()
  contains
    real function wrapper_wrapper()
        wrapper_wrapper = fun()
    end function wrapper_wrapper
  end
end module my_module

program example
  use my_module
  implicit none

  ! This works on GCC and on NVHPC
  ! real, parameter :: target_value = .5

  ! This works on GCC, not on NVHPC
  real :: target_value
  target_value = 0.5

  print*, "Via wrapper (should equal .5): ", my_wrapper(my_fun)
  print*, "Direct call (should equal .5): ", my_fun()
contains
  pure real function my_fun()
    my_fun = target_value
  end function my_fun

end program example
```

This gives:
```console
$ lfortran c.f90
code generation error: Function code not generated for 'fun_interface'


$ gfortran c.f90 && ./a.out
 Via wrapper (should equal .5):   0.500000000
 Direct call (should equal .5):   0.500000000
```

certik:
Fixes #9963.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -315,71 +315,6 @@ public:
         }
     }
 
-    std::string get_nested_procedure_context_name(const ASR::Variable_t& proc_var) {
-        if (proc_var.m_parent_symtab == nullptr || proc_var.m_parent_symtab->asr_owner == nullptr) {
-            return "";
-        }
-        ASR::symbol_t* owner = ASR::down_cast<ASR::symbol_t>(proc_var.m_parent_symtab->asr_owner);
-        if (!ASR::is_a<ASR::Function_t>(*owner)) {
-            return "";
-        }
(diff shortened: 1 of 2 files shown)


certik:
I am marking this as a draft, because the right fix should go into the nested functions pass and in ASR, not special handling in the backend like this. But at least this PR shows one way to fix this.

certik:
The fix is now good. Should be squashed.

Issue #9548, opened 2026-01-17
Remove comments and fix `class_94.f90` test


jinangshah21:
There are many commented lines in `class_94.f90` which don't work with LFortran. 
Uncomment them and run with LFortran to close this issue.

PR #9967, opened 2026-02-10
fix: includes more checks in promotion of allocatable to non-allocatable (static).

yuvrajsinghgmx:
Description for the fix :

1) In case of `Multiple Allocation / Implicit or Explicit Deallocation` a variable cannot be promoted to non-allocatable so i have introduced checks for these cases. 
This fixes TEST 6 from #9548

2) TEST 5 from #9548 was already working on `main` so i have uncommented it.

### DIFF
--- a/integration_tests/class_94.f90
+++ b/integration_tests/class_94.f90
@@ -106,8 +106,8 @@ program class_94
     total = sum_class_values(base_arr)
     if (total /= 63) error stop "Sum of base_arr failed"  ! (11 + 21 + 31)
     
-    ! total = sum_class_values(ext_arr)
-    ! if (total /= 63) error stop "Sum of ext_arr failed"   ! (11 + 21 + 31)
+    total = sum_class_values(ext_arr)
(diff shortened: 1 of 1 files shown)


integration_tests/class_94.f90
```diff
@@ -118,6 +118,6 @@ program class_94
     base_arr(2)%value = 200
     total = sum_class_values(base_arr)
     print *, "Total:", total
-    ! if (total /= 300) error stop "Same type test failed"  !! TODO: fails with --fast
-    ! print *, "Test 6 passed"
+    if (total /= 300) error stop "Same type test failed"  !! Works with --fast
+    print *, "Test 6 passed"
```
HarshitaKalani, line 122: We can just remove this comment since fast is already enabled for class_94.

HarshitaKalani: It's fine, we can do this in subsequent PRs.

yuvrajsinghgmx: Sure.
Thanks for the review @HarshitaKalani

Issue #9814, opened 2026-02-01
Direct access I/O: Non-sequential WRITEs can create a bad file


w6ws:
```
program daio5
  implicit none

  integer, parameter :: unit_no = 24
  character(*), parameter :: fn = 'fort.24'
  integer :: ioerr, iorl

  integer :: i, i1
  logical :: l, pf
  character(5) :: s
  real :: r
  complex :: c
  complex, parameter :: cval = (12.34, 56.78)

  integer, parameter :: wr_order(10) =  &
    [ 5, 3, 6, 2, 8, 1, 9, 7, 10, 4 ]

! Make sure we are starting with a fresh file
  open (unit_no, file=fn, status='old', iostat=ioerr)
  if (ioerr == 0) then
    print *, 'deleting existing file: ', fn
    close (unit_no, status='delete')
  end if

!
  inquire (iolength=iorl) l, s, i, r, c
  print *, 'recl =', iorl

  open (unit_no, file=fn, access='direct', recl=iorl)

! Write some records in pseudo-random order
  do, i=1, 10
    i1 = wr_order (i)
    l = mod (i1,2) == 0
    write (unit_no, rec=i1) l, '12345', i1, i1+0.5, cval
  end do

! Read them back sequentially and check for validity
  do, i=1, 10
    l = .false.; s = 'uini'; i1 = -42; r = -42.42
    read (unit_no, rec=i) l, s, i1, r, c
    pf = (l .eqv. mod (i, 2) == 0) .and. &
         (s == '12345') .and.  &
         (i == i1) .and.  &
         (abs (r - (i + 0.5)) < 0.0001) .and.  &
         (c == cval)
    print *, 'l, s, i1, r =', l, s, i1, r, c
    print *, 'read', i, merge ('pass', 'fail', pf)
  end do

  close (unit_no)

end program
```
```
$ lfortran daio5.f90
deleting existing file:     fort.24
recl =    25
l, s, i1, r =    F    12345    1    1.50000000    (12.3400002,56.7799988)
read    1    pass
l, s, i1, r =    F        0    0.00000000    (0.00000000,0.00000000)
read    2    fail
l, s, i1, r =    F        0    0.00000000    (0.00000000,0.00000000)
read    3    fail
l, s, i1, r =    T    12345    4    4.50000000    (12.3400002,56.7799988)
read    4    pass
Error: Failed to read logical from binary file.
$ od -c fort.24
[...]

PR #9971, opened 2026-02-10
fix: direct access file I/O for unformatted read/write

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9814
fixes: https://github.com/lfortran/lfortran/issues/9719


certik (approved):
Thanks for the fix, I think that this is fine to merge.

PR #9972, opened 2026-02-10
feat: Implement c_f_procpointer in  iso_c_binding

### LINKED ISSUE #9959: bug: `c_f_procpointer` not implemented inside `iso_c_binding`

> testc.c
```c
#include <stdio.h>

void c_callback(int x) {
    printf("C function called with: %d\n", x);
}
```

> test.f90
```fortran
program test_c_f_procpointer
   use iso_c_binding, only: c_funptr, c_funloc, c_f_procpointer
   implicit none
   
   abstract interface
      subroutine sub_interface(x) bind(c)
         use iso_c_binding
         integer(c_int), value :: x
      end subroutine
   end interface
   
   procedure(sub_interface), pointer :: f_proc
   type(c_funptr) :: c_proc
   
   interface
      subroutine c_callback(x) bind(c, name='c_callback')
         use iso_c_binding
         integer(c_int), value :: x
      end subroutine
   end interface
   
   ! Get C function pointer using c_funloc
   c_proc = c_funloc(c_callback)
   
   ! Convert to Fortran procedure pointer
   call c_f_procpointer(c_proc, f_proc)
   
   ! Call C function via Fortran procedure pointer
   call f_proc(42)
   
end program test_c_f_procpointer
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gcc -c testc.c -o testc.o 
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 testc.o -o test.o && ./test.o
C function called with: 42
(lf) saurabhkumar@Mac lfortran % lfortran test.f90 testc.o -o test.o && ./test.o
semantic error: The symbol 'c_f_procpointer' not found in the module 'iso_c_binding'
 --> test.f90:2:49
  |
2 |    use iso_c_binding, only: c_funptr, c_funloc, c_f_procpointer
  |                                                 ^^^^^^^^^^^^^^^ 


```

opixdown:
fixes #9959

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -247,11 +247,53 @@ public:
     std::map<ASR::symbol_t*, int> type2vtabid;
     std::map<ASR::symbol_t*, std::map<std::string, int64_t>> vtabtype2procidx;
     std::map<ASR::symbol_t*, std::map<std::string, int64_t>> struct_vtab_function_offset;
-    llvm::Type* current_select_type_block_type;
-    ASR::ttype_t* current_select_type_block_type_asr;
-    std::string current_select_type_block_der_type;
-    std::string current_selector_var_name;
-    ASR::symbol_t* current_selector_type_decl;
+
+    // Context for a select type block - maps selector variable symbol to its type info
+    struct SelectTypeContext {
+        llvm::Type* block_type = nullptr;
+        ASR::ttype_t* block_type_asr = nullptr;
+        std::string der_type;
+        ASR::symbol_t* selector_type_decl = nullptr;
+    };
+    // Map from selector variable symbol to its select type context
+    std::map<ASR::symbol_t*, SelectTypeContext> select_type_context_map;
+
+    // Helper methods to access select type context for a specific variable
+    SelectTypeContext* get_select_type_context(ASR::symbol_t* sym) {
+        if (sym == nullptr) return nullptr;
+        auto it = select_type_context_map.find(sym);
+        return (it != select_type_context_map.end()) ? &it->second : nullptr;
+    }
+
+    // Helper to extract symbol from expression (returns nullptr if not Var_t)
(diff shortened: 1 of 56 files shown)


opixdown:
I looked at how existing iso_c_binding functions like c_f_pointer are implemented and followed the same approach. I reused the existing lowering logic, with a small backend change to correctly handle procedure pointers. Please let me know if this looks good and correct, and if the test registration and reference updates also look fine.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6903,10 +6848,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                ASR::ttype_t* fptr_asr_type = ASRUtils::expr_type(fptr);
+                ASR::ttype_t* fptr_contained = ASRUtils::type_get_past_pointer(fptr_asr_type);
+                bool is_proc_ptr = ASR::is_a<ASR::FunctionType_t>(*fptr_contained);
+                llvm::Type* llvm_fptr_elem_type =llvm_utils->get_type_from_ttype_t_util(fptr, fptr_contained, module.get());
+                if (is_proc_ptr) {
+                    // procedure pointer cast to function pointer
+                    llvm_cptr = builder->CreateBitCast(llvm_cptr,llvm_fptr_elem_type);
+                    builder->CreateStore(llvm_cptr, llvm_fptr);
+                } else {
+                    // existing behavior
```
certik: This comment doesn't make sense --- if somebody is reading the new code there is no "existing behavior".

certik:
```suggestion
```

certik: We'll apply this change right before merging. Let's run full tests now.

certik:
I think this is good. Let's see if all tests pass.

src/libasr/codegen/asr_to_llvm.cpp
opixdown: full test passed adn applied the change

Issue #9717, opened 2026-01-27
Refactor float format handling to use enums instead of string comparisons


HarshitaKalani:
ref: https://github.com/lfortran/lfortran/pull/9714#discussion_r2733019677

PR #9975, opened 2026-02-10
refactor: use enum instead of string comparison in handle_float

Pranavchiku:
fixes: #9717 

Replace string comparison with enum for float format type dispatch in handle_float() for better performance and type safety.

- Add FloatFormatType enum with F64, F32, and CUSTOM values
- Update handle_float signature to take FloatFormatType as first param
- Replace strcmp() calls with enum comparisons
- Update caller to determine format type and pass enum value

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -15,6 +15,13 @@
 #include <stddef.h>  /* ptrdiff_t */
 
 #define PI 3.14159265358979323846
+// Enum for float format types to avoid string comparison
+typedef enum {
+    FLOAT_FORMAT_F64,     // for default 64-bit float formatting
+    FLOAT_FORMAT_F32,     // for default 32-bit float formatting
+    FLOAT_FORMAT_CUSTOM   // for custom format strings like "F10.5"
+} FloatFormatType;
+
 #if defined(_WIN32)
 #  include <winsock2.h>
 #  include <io.h>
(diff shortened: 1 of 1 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2540,7 +2384,15 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(const char* format, int64_t
...
                     char* temp_buf = (char*)malloc(1); temp_buf[0] = '\0';
-                    handle_float(value, double_val, scale, &temp_buf, is_SP_specifier, rounding_mode);
+                    FloatFormatType float_fmt_type;
+                    if (strcmp(value, "f-64") == 0) {
+                        float_fmt_type = FLOAT_FORMAT_F64;
+                    } else if (strcmp(value, "f-32") == 0) {
+                        float_fmt_type = FLOAT_FORMAT_F32;
+                    } else {
+                        float_fmt_type = FLOAT_FORMAT_CUSTOM;
+                    }
```
certik, line 2394: Later on this should be replaced by passing the enum as an argument from ASR->LLVM, instead of a string.

certik (approved):
I think that looks good. Hopefully we don't need to run full tests.

src/libasr/runtime/lfortran_intrinsics.c
certik: https://github.com/lfortran/lfortran/issues/10029

PR #9976, opened 2026-02-10
Implements recursive nested function support

### LINKED ISSUE #9979: "Man or boy" recursive nested function test
From https://rosettacode.org/wiki/Man_or_boy_test#Fortran.

This currently does not work:
```fortran
module man_or_boy

implicit none

contains

  recursive integer function A(k,x1,x2,x3,x4,x5) result(res)
    integer, intent(in) :: k
    interface
      recursive integer function x1()
      end function
      recursive integer function x2()
      end function
      recursive integer function x3()
      end function
      recursive integer function x4()
      end function
      recursive integer function x5()
      end function
    end interface
    integer :: m
    if ( k <= 0 ) then
      res = x4()+x5()
    else
      m = k
      res = B()
    end if
  
  contains
  
    recursive integer function B() result(res)    
      m = m-1
      res = A(m,B,x1,x2,x3,x4)
    end function B
  
  end function A


  recursive integer function one() result(res)
    res = 1
  end function

  recursive integer function minus_one() result(res)
    res = -1
  end function

  recursive integer function zero() result(res)
    res = 0
  end function

end module man_or_boy

program test
  use man_or_boy
  write (*,*) A(10,one,minus_one,minus_one,one,zero)
end program test
```

It requires multi-level nested function recursion. A prototype implementation in #9976 makes it work:
```console
$ gfortran b.f90 && ./a.out
         -67
$ lfortran b.f90
-67
```

certik:
The code passes integration and reference tests. But needs refactoring, possibly complete redesign. But it does get the basics working, and we can use it to figure out a proper design now when we know the requirements.

The document `doc/src/passes/nested_vars_recursive_context_stack.md` explains the current design and how it works. It's using a pointer+index pairs that have to be kept in sync. The pointer is typically and argument, while the index is a global variable.

An alternative design is with a trampoline stub: we pregenerate N stubs (say N=16), which remember the index inside them, and then they copy the environment from the index and call the actual nested function. The trampolines in GCC work just like that, but generate those stubs at runtime as JIT, but that typically requires executable stack which is not always enabled for security reasons, so we don't want to use that approach. The pregenerated stubs don't require executable stacks, so it's cleaner, but since one must pregenerate them, it's limited to N recursion levels. The design in this PR is not using stubs, but the pointer+index pairs, which is another way to implement this. 

I am guessing the pregenerated stubs is the best way, as I think we can do all of this in ASR. This PR uses a hybrid, some stuff is done in the LLVM backend. 

Fixes #9979.

### DIFF
--- a/doc/src/passes/nested_vars_recursive_context_stack.md
+++ b/doc/src/passes/nested_vars_recursive_context_stack.md
@@ -65,6 +65,68 @@ closure frame stack model in ASR-level transformation:
 Backend logic remains minimal and implements only required runtime behavior for
 procedure-variable calls that consume ASR-generated context metadata.
 
+## Trampoline vs Current Model
+
+This implementation does **not** use a trampoline/stub closure ABI.
+
+- The procedure variable still stores a raw function pointer (for example,
+  `x4` stores pointer to `one`, `minus_one`, `B`, etc.).
+- The closure environment binding is stored separately in an integer context id
+  variable `<proc>__ctx` (for example, `x4__ctx`).
+- Captured values and captured procedure contexts are stored in module-level
+  stack arrays (`<captured>__stack`, `<captured>__ctx__stack`) indexed by the
+  context id.
+
+So the effective callable object is represented as:
+
+- pointer: `<proc>`
+- environment id: `<proc>__ctx`
+
+not as a synthesized trampoline function pointer.
+
+### How Indirect Calls Get the Right Environment
+
+At ASR level, the pass guarantees that procedure capture assignment updates both
+the pointer and its context id:
+
+- pointer sync: `<proc> = <source_proc>`
+- context sync: `<proc>__ctx = <source_ctx or current stack ptr>`
+
+At LLVM lowering of indirect procedure calls:
+
+1. Load `saved_stack_ptr` from `__lfortran_nested_ctx_stack_ptr`.
+2. Load `ctx_val` from `<proc>__ctx`.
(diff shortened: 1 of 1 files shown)


assem2002:
Reviewing..

assem2002:
I reviewed almost half of the changes (the general approach), Will handle the rest tomorrow ASAP.

certik:
@assem2002 thank you. We need to figure out a good way forward how to fix it without increasing the complexity too much.

assem2002:
Per my understanding, This approach is extremely linked to the idea of host/nested-function, that means any usage beyond that wouldn't work.

The problem here is trying to mimic a fraction of what functional programming is doing (function object -> `{Env + fn ptr}`). All other compilers are using trampolines (runtime created wrappers) to fallback to the ENV (Flang is warning about deprecating this btw) ; 
The approach in this PR is attempting to only fix recursive-nested-call by analyzing the ASR to know the call chain and insert some logic to set/unset the ENV; If we can't analyze the ASR to track the chain, the approach will fail.

It's a weird feature and no one would love to go into that recursive-enclosed env. headache unless he loves functional programming.
The trampoline approach is controlling the narrative, We have to mimic what trampoline is doing if we want to support that.

consider [this example](https://gist.github.com/assem2002/dce379aa49974ce19a365fe10d25e806) (A -> B -> C -> D -> A` -> B). It's showing a deficit with the current PR ( I think it's fundamental not just an unhandled case -- correct me if I'm wrong) 

***
Another intake is that I think the resulting ASR is very redundant + I couldn't find the logic responsible for actually setting the env correctly, I think it's being handled in the backend which contradicts the idea of using ASR to handle things explicitly (which this PR is trying to attempt).

PR #9977, opened 2026-02-10
fix cascading error by recovering from invalid assumed-size dimensions (#9094)

### LINKED ISSUE #9094: Suppress spurious 'Dummy argument not defined' error after dimension errors

amritamishra01:
This patch implements semantic recovery for dummy arguments with assumed-size dimensions (*) in a non-last position. Previously, an invalid dimension would trigger an immediate abort, preventing the compiler from completing the symbol table registration for that variable.

Previous Behavior
For code like integer :: a(*, 1), LFortran would:

Emit the correct diagnostic for the invalid dimension.

Immediately throw SemanticAbort.

Fail to register the dummy argument in the symbol table.

Produce a secondary, confusing error: "Dummy argument 'a' not defined".

New Behavior
Diagnostic Retention: Still emits the required error: "Assumed-size '*' is only permitted in the last dimension".

Resilient Recovery: Instead of aborting, the compiler now:

Inserts a safe "dummy" dimension (1:1).

Continues semantic analysis of the remaining code.

Allows symbol table registration to complete successfully.

Result
Elimination of Cascading Errors: The secondary "not defined" error is now suppressed because the compiler "remembers" the variable.

Improved UX: Matches GFortran-style recovery, providing a cleaner developer experience.

Regression Tested: Added tests/errors/repro_9094.f90 to ensure this behavior is preserved.

Reference Updates: Updated continue_compilation_1.f90 and other affected references to reflect the cleaner, more accurate diagnostic output.

Closes #9094


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2547,14 +2547,24 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        diag::Label("", {m_dim[i].loc})}));
+
+                ASR::ttype_t* int_type = ASRUtils::TYPE(ASR::make_Integer_t(al, m_dim[i].loc, 4));
+                ASR::expr_t* one = ASRUtils::EXPR(ASR::make_IntegerConstant_t(
+                    al, m_dim[i].loc, 1, int_type));
+
+                ASR::dimension_t dim_dummy;
+                dim_dummy.loc = m_dim[i].loc;
+                dim_dummy.m_start = one;
+                dim_dummy.m_length = one;
```
HarshitaKalani: Maybe the below makes more sense?
```
dim_dummy.m_start = one;
dim_dummy.m_length = nullptr;
```

tests/errors/repro_9094.f90
HarshitaKalani, line 1: Can we name the test better?

tests/errors/repro_9094.f90
```diff
@@ -0,0 +1,5 @@
+subroutine repro_9094(a, b)
```
HarshitaKalani: same here

tests/errors/repro_9094.f90:2:30
HarshitaKalani, line 1: can we remove this?

amritamishra01:
All CI checks are green now.
I implemented a surgical fix by validating the position of the assumed-size * before the process_dims call, which prevents cascading errors while preserving the existing ASR behavior for unaffected tests (e.g., array14.f90).
The regression test has also been renamed to assumed_size_nonlast_dim.f90 as requested.

tests/errors/assumed_size_nonlast_dim.f90
```diff
@@ -0,0 +1,5 @@
+subroutine assumed_size_nonlast_dim(a, b)
+    integer, intent(in) :: a(*, 1)
+    integer, intent(in) :: b
+    print *, a(1,1), b
+end subroutine
```
certik: This file is not tested. Let's add this into continue_compilation_1.f90?

PR #9980, opened 2026-02-10
Fix crash on deferred type-bound dispatch

### LINKED ISSUE #9978: Indexing of an object array leads to segfault upon method invocation
The following code when compiled with flang-20  prints `shape(b):  2 4`.
`$ cat arrayshape.f90 `
```
module MyTypeMod
   
   type, abstract :: AbsType
   contains
      procedure(alloc), deferred  :: alloc
   end type AbsType

   abstract interface
      subroutine alloc(self,b,n)
         import
         class(AbsType),       intent(in)  :: self
         real(8), allocatable, intent(out) :: b(:,:)
         integer,              intent(in)  :: n
      end subroutine alloc
   end interface
   
   contains
      procedure, public  :: alloc => allocation
   end type MyType
   
contains
   
   subroutine allocation(self,b,n)
      class(MyType),        intent(in)  :: self
      real(8), allocatable, intent(out) :: b(:,:)
      integer,              intent(in)  :: n
      allocate( b(2,n) )
   end subroutine allocation
   
end module MyTypeMod


program arrayshape
   
   use MyTypeMod
   
   class(AbsType), allocatable :: objarr(:)
   
   allocate(MyType :: objarr(2))
   objarr = MyType()

   call allocator(objarr(1))
   
contains
   
   subroutine allocator(obj)
      class(AbsType), intent(in)  :: obj
      real(8),        allocatable :: b(:,:)
      call obj%alloc(b,4)
      write(*,*) 'shape(b): ', shape(b)      
   end subroutine allocator

end program arrayshape

```
With LFortran the result is a segfault:
`$ lfortran arrayshape.f90`
`Segmentation fault (core dumped)
`

certik:
The `objarr(1)` was wrapped as `class(AbsType)` with the static abstract vptr, not the element’s runtime vptr.  That made `obj%alloc(...)` resolve to a null slot and crash.

To fix it, we now preserve runtime vptr when creating class views from class-array elements.

Fixes #9978.

### DIFF
--- a/src/libasr/pass/promote_allocatable_to_nonallocatable.cpp
+++ b/src/libasr/pass/promote_allocatable_to_nonallocatable.cpp
@@ -155,6 +155,7 @@ class PromoteAllocatableToNonAllocatable:
                     *ASRUtils::expr_type(alloc_arg.m_a)) &&
                     ASRUtils::is_array(ASRUtils::expr_type(alloc_arg.m_a));
                 bool is_deferred_len_character_array = false;
+                bool is_class_array = false;
                 if (is_allocatable_array) {
                     ASR::ttype_t* element_type = ASRUtils::type_get_past_array(
                         ASRUtils::type_get_past_allocatable(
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -8632,9 +8632,19 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
     
-    llvm::Value* LLVMStruct::create_class_view(ASR::Struct_t* const class_symbol, llvm::Value* const viewed_struct){
+    llvm::Value* LLVMStruct::create_class_view(ASR::Struct_t* const class_symbol,
+            llvm::Value* const viewed_struct, llvm::Value* vptr){
         llvm::Value* const allocated_class_structure = builder->CreateAlloca(llvm_utils->getClassType(class_symbol));
-        store_class_vptr(&class_symbol->base, allocated_class_structure, llvm_utils->module);
+        if (vptr) {
+            if (vptr->getType() != llvm_utils->vptr_type) {
+                vptr = builder->CreateBitCast(vptr, llvm_utils->vptr_type);
+            }
```
jinangshah21, line 8641: This check might not be needed. But it is ok for now.

certik:
I am just going to merge it, since the last test that didn't run yet already passed on the previous commit, so likely it will pass again. If not, we'll have to fix it of course.

PR #9985, opened 2026-02-11
Fix integer(8) handling in modulo

certik:


certik:
@HarshitaKalani let me know if the change looks good in the intrinsic function.

certik:
Tests passed, so I merged it. If there are fixes that we should do, we can do them in a subsequent PR.

HarshitaKalani (commented):
This is good.

PR #9986, opened 2026-02-11
LLVM: allocate class wrappers before a loop

certik:
Instead of in the loop, allocate them in the function entry point.


certik:
@jinangshah21 this moves alloca out of the loop to prevent a stack overflow, the usual issue that we fixed a few times already. Let me know if the code looks ok to you.

jinangshah21 (commented):
I think this is fine!

PR #9988, opened 2026-02-11
fix: allocate destination data in struct descriptor

certik:
When deep-copying descriptor arrays of derived types, codegen could enter the element-copy loop with a null destination data pointer. This happened in component-section paths like `v(:)%n` when the source element had allocatable components, causing `_copy_*` to write through a null pointer at runtime.

Allocate destination data when source data is allocated but destination data is null, then store the pointer in the descriptor, reset offset, and reload the data pointer before copying elements.

### DIFF
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -9188,7 +9188,7 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al, 
                 LCOMPILERS_ASSERT(false);
             }
 
-            if (llvm_utils->compiler_options.po.realloc_lhs_arrays && ASRUtils::is_allocatable(src_expr)) {
+            if (ASRUtils::is_allocatable(src_expr)) {
                 // Check if src_data is not null before realloc operations
                 llvm::Value* src_data_not_null = builder->CreateICmpNE(
                     src_data, llvm::Constant::getNullValue(src_data->getType()));
--- a/src/libasr/pass/array_struct_temporary.cpp
+++ b/src/libasr/pass/array_struct_temporary.cpp
@@ -1983,12 +1983,11 @@ class ReplaceExprWithTemporary: public ASR::BaseExprReplacer<ReplaceExprWithTemp
     ASR::ttype_t* simd_type;
     ASR::expr_t* parent_expr;
     ASR::expr_t* lhs_var;
-    bool associate_only_array_section;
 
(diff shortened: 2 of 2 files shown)


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -9213,6 +9213,31 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
+                    llvm::Value* allocated_data = builder->CreateBitCast(
+                        malloc_ptr, llvm_data_type->getPointerTo());
+                    builder->CreateStore(allocated_data,
+                        llvm_utils->arr_api->get_pointer_to_data(llvm_array_type, dest));
+                    llvm::Value* offset_val = llvm_utils->create_gep2(llvm_array_type, dest, 1);
+                    builder->CreateStore(llvm::ConstantInt::get(context,
+                            llvm::APInt(index_bit_width, 0)), offset_val);
+                }, []() {});
+                dest_data = llvm_utils->CreateLoad2(llvm_data_type->getPointerTo(),
+                    llvm_utils->arr_api->get_pointer_to_data(llvm_array_type, dest));
```
jinangshah21: This will add runtime check for all descriptor arrays, but we want to allocate descriptor only for data member which is actually scalar but struct arraysection made it an array. I think we have temporary variable for this and we need to add allocate statement in ASR.

certik: Yes, I didn't like the if statement either! I'll try to come up with a better solution. Thanks for the review.

certik: Fixed in d5e865c32bf97eb3de3d315dd5bdc8ad902cc7e3.

certik:
@jinangshah21 let me know if the fix is better now.

src/libasr/codegen/llvm_utils.cpp
jinangshah21:
```diff
- if (llvm_utils->compiler_options.po.realloc_lhs_arrays && ASRUtils::is_allocatable(src_expr)) {
+ if (ASRUtils::is_allocatable(src_expr)) {
```

jinangshah21, line 9192: Oh my bad! Above suggested change will fix this issue. Previously we were allocating member array only if realloc_lhs_arrays was on

certik:
@jinangshah21 Thanks for the fix, that indeed works. Let me know now.

PR #9992, opened 2026-02-11
fix: Enforce INTEGER type requirement for dim argument in sum intrinsic

### LINKED ISSUE #4306: Throw error when sequence of args is incorrect
Here the signature of sum intrinsic is: sum(array, dim, mask). We should throw error if the args are provided without keywords and the sequence of args is incorrect 

```fortran
program main
	print *, sum([1, 2], [.true., .false.], 1)
end program
```
```console
$ gfortran a.f90 && ./a.out
a.f90:2:22:

    2 |         print *, sum([1, 2], [.true., .false.], 1)
Error: ‘dim’ argument of ‘sum’ intrinsic at (1) must be INTEGER
$ lfortran a.f90
1
```

Jatinagarwal24:
Fixes #4306

### DIFF
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -202,7 +202,7 @@ program continue_compilation_3
     print *, "8356" .eqv. 8356.00
     print *, ['c', 'o', 'd', 'e'] .or. ['m', 'a', 's']
     print *, ["welcome", "to", "lf"] .and. "contributors"  !even size diff of array element must be caught
-    print *,  [( '*',i , i = 1, 10 )] 
+    print *, sum([1, 2], [.true., .false.], 1)
 
     contains 
     subroutine bpe()
--- a/tests/reference/asr-continue_compilation_3-435a232.json
+++ b/tests/reference/asr-continue_compilation_3-435a232.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_3-435a232",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_3.f90",
-    "infile_hash": "20d0f63666bb303112b06f8fae345185c8f86b7bb4f86a5f4a11948a",
+    "infile_hash": "6674783f3e43934be5dbadae0af7c8670a49953df6c9e877ae8cf3fc",
     "outfile": null,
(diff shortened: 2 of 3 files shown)


tests/errors/continue_compilation_3.f90
```diff
@@ -202,6 +202,7 @@ program continue_compilation_3
     print *, "8356" .eqv. 8356.00
     print *, ['c', 'o', 'd', 'e'] .or. ['m', 'a', 's']
     print *, ["welcome", "to", "lf"] .and. "contributors"  !even size diff of array element must be caught
+    print *,  [( '*',i , i = 1, 10 )]
```
HarshitaKalani: I didn't understand the point of adding this line. Could you please explain?

HarshitaKalani (commented):
Cna you please address the comment, otherwise I think this is good. Thanks!

tests/errors/continue_compilation_3.f90
Jatinagarwal24: Hi,
I was working on some other issue simultaneously due to which i by mistake added this.Sorry for that.
Thanks.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -618,23 +618,23 @@ semantic error: The ranks of the `array` and `mask` arguments of the `Sum` intri
...
     |                      ^^^^  ^^^^^ `array` has shape [2,3], but `mask` has shape [3,2]
 
-semantic error: The shapes of the `array` and `mask` arguments of the `Iparity` intrinsic must be the same
-   --> tests/errors/continue_compilation_1.f90:389:22
+semantic error: `dim` argument of `Iparity` intrinsic must be INTEGER
+   --> tests/errors/continue_compilation_1.f90:389:28
     |
 389 |     print *, iparity(arr3, mask5, 3)
-    |                      ^^^^  ^^^^^ `array` has shape [2,1,3], but `mask` has shape [2,3,1]
+    |                            ^^^^^
```
certik, line 637: I think that there are several problems with these changes:

* The old error is now not tested --- so we need to figure out another test to test the old errors
* The new error seems to point to `mask5` as a dim? Maybe it's correct, but let's check.

Jatinagarwal24:
fixed via #10681

Issue #9950, opened 2026-02-09
Use ClassToStruct cast consistently


certik:
This means: adding tests in verify that this cast is always used when required. And this in turn will force to add it in AST->ASR right away and maintain it throughout the process.

We used to do this cast implicitly in the backend, here is an example how a fix using this approach looks like: https://github.com/lfortran/lfortran/pull/9927. We then refactored this fix to use a cast in ASR in: https://github.com/lfortran/lfortran/pull/9931, which makes the backend code a lot simpler, we do not need to track such a mismatch at all.

PR #10002, opened 2026-02-11
Refactor: make ClassToStruct + ClassToClass casting at ASR level inside `select type`

jinangshah21:
Towards #9950

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4026,6 +4026,11 @@ public:
         // Look up select type context by expression
         llvm::Type* ctx_block_type = get_select_type_block_type(x.m_v);
         const std::string& ctx_der_type = get_select_type_block_der_type(x.m_v);
+        if (ASR::is_a<ASR::Cast_t>(*x.m_v) && LLVM::is_llvm_pointer(*x_m_v_type)) {
+            llvm::Type* type = llvm_utils->get_type_from_ttype_t_util(
+                x.m_v, ASRUtils::type_get_past_allocatable_pointer(x_m_v_type), module.get());
+            tmp = llvm_utils->CreateLoad2(type->getPointerTo(), tmp);
+        }
         if (ASRUtils::is_unlimited_polymorphic_type(x.m_v)) {
             if( compiler_options.new_classes && ctx_block_type) {
                 llvm::Type* x_mv_llvm_type = llvm_utils->get_type_from_ttype_t_util(
@@ -4132,7 +4137,7 @@ public:
 
         llvm::Type *xtype = name2dertype[current_der_type_name];
         if (LLVM::is_llvm_pointer(*x_m_v_type) && 
-            (ASR::is_a<ASR::StructInstanceMember_t>(*x.m_v) || ASR::is_a<ASR::Cast_t>(*x.m_v)) &&
+            ASR::is_a<ASR::StructInstanceMember_t>(*x.m_v) &&
             !ASRUtils::is_class_type(ASRUtils::extract_type(x_m_v_type))) {
             tmp = llvm_utils->CreateLoad2(xtype->getPointerTo(), tmp);
         }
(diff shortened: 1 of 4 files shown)


src/libasr/ASR.asdl
```diff
@@ -176,7 +176,7 @@ expr
     | OverloadedBoolOp(expr left, logicalbinop op, expr right, ttype type, expr? value, expr overloaded)
     | OverloadedUnaryMinus(expr arg, ttype type, expr? value, expr overloaded)
     | OverloadedStringConcat(expr left, expr right, ttype type, expr? value, expr overloaded)
-    | Cast(expr arg, cast_kind kind, ttype type, expr? value)
+    | Cast(expr arg, cast_kind kind, ttype type, expr? value, expr? dest)
```
certik, line 179: Document in doc/ what the `dest` does.

certik:
I think this is great, thanks for adding these casts.

src/libasr/ASR.asdl
jinangshah21: Done

jinangshah21:
In this PR I handled all class/struct type cast inside `select_type` statements.
NEXT: Do similar casts for `class(*)` variables, then backend will not need any context storing, all things will be inside casts.

jinangshah21:
See `select_type_*` test references changes to watch update in ASR from this PR.

certik:
I rebased it, let's run tests.

certik:
There is some linker issue but it doesn't say what exactly:
```
[ 26%] Linking Fortran executable intrinsics_27_STD_F23
[ 26%] Linking Fortran executable intrinsics_28_STD_F23
[ 26%] Linking Fortran executable intrinsics_29_STD_F23
[ 26%] Linking Fortran executable intrinsics_30_STD_F23
[ 26%] Linking Fortran executable intrinsics_31_STD_F23
[ 26%] Built target intrinsics_25_STD_F23
[ 26%] Building Fortran object CMakeFiles/intrinsics_33_STD_F23.dir/intrinsics_33.f90.o
[ 26%] Linking Fortran executable intrinsics_32_STD_F23
[ 26%] Built target intrinsics_26_STD_F23
The command 'clang -o intrinsics_27_STD_F23 CMakeFiles/intrinsics_27_STD_F23.dir/intrinsics_27.f90.o  -L"/Users/runner/micromamba/envs/lf/bin/../lib" -Wl,-rpath,"/Users/runner/micromamba/envs/lf/bin/../lib" -llfortran_runtime -lm' failed.
[ 26%] Building Fortran object CMakeFiles/intrinsics_34_STD_F23.dir/intrinsics_34.f90.o
make[2]: *** [intrinsics_27_STD_F23] Error 10
make[1]: *** [CMakeFiles/intrinsics_27_STD_F23.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

jinangshah21:
It passed on re-run, also failure don't seems to be from this PR

certik:
> It passed on re-run, also failure don't seems to be from this PR

Can you please open up an issue for this flaky test?

doc/src/asr/asr_nodes/cast_kind_nodes/cast_kind.md
```diff
@@ -659,4 +661,109 @@ ASR:
 )
 ```
 
+Example of `ClassToStruct/ClassToClass`:
+
+```fortran
+select type(var)
+    type is (base)
+        print *, var%x
```
certik, line 669:
```suggestion
        print *, var%x   ! ClassToStruct
```

doc/src/asr/asr_nodes/cast_kind_nodes/cast_kind.md
```diff
@@ -659,4 +661,109 @@ ASR:
...
+    class is (derived)
+        print *, var%y
```
certik, line 671:
```suggestion
        print *, var%y  ! ClassToClass
```

doc/src/asr/asr_nodes/cast_kind_nodes/cast_kind.md
```diff
@@ -659,4 +661,109 @@ ASR:
...
+end select
+```
```
certik, line 673: Then explain here what the difference is and when to use each. It is actually not clear to me at all. Also explain what the cast does (that in the backend we get rid of the class wrapper and extract the type directly I think).

doc/src/asr/asr_nodes/expression_nodes/Cast.md
```diff
@@ -18,6 +18,7 @@ Cast(expr arg, cast_kind kind, ttype type, expr? value)
 |`kind` | cast to kind |
```
certik, line 18:
```suggestion
|`kind` | the cast kind enum |
```

doc/src/asr/asr_nodes/expression_nodes/Cast.md
```diff
@@ -18,6 +18,7 @@ Cast(expr arg, cast_kind kind, ttype type, expr? value)
...
 |`type` | table entry type |
```
certik, line 19:
```suggestion
|`type` | type after the cast |
```
Is that correct?

doc/src/asr/asr_nodes/expression_nodes/Cast.md
```diff
@@ -18,6 +18,7 @@ Cast(expr arg, cast_kind kind, ttype type, expr? value)
...
 |`value`| expression value |
```
certik, line 20:
```suggestion
|`value`| compile time value of the result if available |
```

doc/src/asr/asr_nodes/expression_nodes/Cast.md
```diff
@@ -18,6 +18,7 @@ Cast(expr arg, cast_kind kind, ttype type, expr? value)
...
+|`dest`| use to get struct symbol of destination struct type |
```
certik, line 21: Underneath can you explain when to use `dest`, that is, when to "get struct symbol of destination struct type"? I don't understand it at all. Maybe some example would help.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -3969,6 +3969,50 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        llvm::Type::getInt64Ty(context)),
+                    builder->CreatePtrToInt(llvm::ConstantPointerNull::get(
+                        llvm::cast<llvm::PointerType>(class_instance_ptr->getType())),
+                        llvm::Type::getInt64Ty(context)));
+                llvm_utils->generate_runtime_error(cond,
+                        "Tried to access member of unallocated variable '%s'",
+                        infile,
+                        cast_expr->m_arg->base.loc,
+                        location_manager,
+                        LCompilers::create_global_string_ptr(context, *module, *builder, var_name));
```
certik, line 4012: Let's add an error test for this new check.

doc/src/asr/asr_nodes/expression_nodes/Cast.md
certik: I can see it is used in the LLVM backend to greatly simplify the cast implementation to use
```
                if (x.m_dest) {
                    struct_sym = ASRUtils::symbol_get_past_external(
                        ASRUtils::get_struct_sym_from_struct_expr(x.m_dest));
                }
```
But I still don't understand what it does and why it is needed.

src/libasr/codegen/asr_to_llvm.cpp
jinangshah21: It is already tested by `scalar_allocation_check_2.f90`, it is just I need to throw this error before unwrapping in cast if class is unallocated.

Issue #8752, opened 2025-10-24
Error compiling do loop with both construct name and numeric statement label


w6ws:
```
program do_label
  implicit none

  integer :: i

! Construct name, but no numeric statement label
do_no: do i=1,5
    print *, i
  end do do_no

! Construct name plus numeric statement label
do10: do 10 i=1,5
    print *, i
10  end do do10

end program
```
```
$ lfortran -c do_label.f90
syntax error: Token 'do10' (of type 'identifier') is unexpected here
  --> do_label.f90:14:12
   |
14 | 10  end do do10
   |            ^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```
Version of lfortran is from the trunk a few days ago:
```
$ lfortran --version
LFortran version: 0.57.0-168-g6bd06aa68
Platform: Linux
LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$
```

harperjf:
4 other compilers ran the program, printing the expected 10 lines, though g95 said correctly that the nonblock DO statement is obsolescent.

w6ws:
Note that this is also an issue with construct names and IF blocks.  Slightly expanded example:
```
program do_label
  implicit none

  integer :: i

! Construct name, but no numeric statement label
do_no: do i=1,5
    print *, i
  end do do_no

! Construct name plus numeric statement label
do10: do 10 i=1,5
    print *, i
10  end do do10

! Similar tests with IF

if_no:  &
  if (i == 6) then
    print *, 'i =', i
  end if if_no

ifcn:  &
  if (i == 6) then
    print *, 'i =', i
20  end if ifcn

end program
```
```
$ lfortran --continue-compilation do_label_8752.f90
syntax error: Token 'do10' (of type 'identifier') is unexpected here
  --> do_label_8752.f90:14:12
   |
14 | 10  end do do10
   |            ^^^^ 

syntax error: Token 'end if' is unexpected here
  --> do_label_8752.f90:26:5
   |
26 | 20  end if ifcn
   |     ^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
$
```

PR #10007, opened 2026-02-11
fix: handle construct name on labeled end do statement

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/8752

### DIFF
--- a/src/lfortran/parser/tokenizer.re
+++ b/src/lfortran/parser/tokenizer.re
@@ -162,11 +162,34 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
         enddo_insert_count--;
         if (enddo_insert_count > 0) {
             enddo_state = 1;
+            return yytokentype::TK_NEWLINE;
         } else {
-            enddo_state = 0;
-            token_loc(loc); line_num++; cur_line=cur;
-            last_token = yytokentype::TK_NEWLINE;
+            if (enddo_saved_name_start != nullptr) {
+                // Emit construct name before the final newline
+                enddo_state = 4;
+                tok = enddo_saved_name_start;
+                cur = enddo_saved_name_end;
+                enddo_saved_name_start = nullptr;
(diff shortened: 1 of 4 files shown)


src/lfortran/parser/tokenizer.re
```diff
@@ -386,13 +409,33 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
...
                 if (enddo_newline_process) {
+                    // Save optional trailing construct name for later emission
+                    unsigned char *p = cur;
+                    while (*p == ' ' || *p == '\t') p++;
+                    if ((*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z') || *p == '_') {
+                        enddo_saved_name_start = p;
+                        p++;
+                        while ((*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z') || *p == '_' || (*p >= '0' && *p <= '9')) p++;
+                        enddo_saved_name_end = p;
+                        cur = p;
```
certik, line 420: Let's not do manual handling like this, but rather use re2c mechanisms. You can extract it to a function (also it's duplicated with the change below). But rather let's figure out the tokenizer logic so that we can use re2c directly.

src/lfortran/parser/tokenizer.re
```diff
@@ -162,11 +162,34 @@ int Tokenizer::lex(Allocator &al, YYSTYPE &yylval, Location &loc, diag::Diagnost
...
+                cur = enddo_saved_name_end;
+                enddo_saved_name_start = nullptr;
+                enddo_saved_name_end = nullptr;
+                Str s;
+                s.p = (char*) tok;
+                s.n = cur - tok;
+                yylval.string = s;
+                token_loc(loc);
+                last_token = yytokentype::TK_NAME;
+                return yytokentype::TK_NAME;
```
certik, line 180: This also seems quite complex.

PR #10008, opened 2026-02-11
fix: preserve derived-type declaration on lowered call temporaries

### LINKED ISSUE #10006: Compilation segfaults when concatening a user-defined constructor with a function returning an `allocatable` object
The following valid code compiles with both flang-20, gfortran, and ifx:
`$ cat test.f90 `
```
module mytypemod

   type, abstract :: AbsType
   end type AbsType

   type :: MyType
   end type MyType

   interface MyType
      procedure :: constructor
   end interface MyType
   
contains
   
   end function constructor

end module mytypemod

module clientmod

   use mytypemod,  only: AbsType, MyType
   
contains

   function get_my_obj() result(myobj)
      class(MyType), allocatable :: myobj
      myobj = MyType(get_obj())
   end function get_my_obj
   
   function get_obj() result(obj)
   end function get_obj
   
end module clientmod

```
With LFortran, trying to compile this gives a segfault:
`$ lfortran -c --separate-compilation  test.f90 `

certik:
Fixes #10006.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -2211,7 +2211,6 @@ public:
     }
 
     void visit_Associate(const AST::Associate_t& x) {
-        tmp = nullptr;
         this->visit_expr(*(x.m_target));
         ASR::expr_t* target = ASRUtils::EXPR(tmp);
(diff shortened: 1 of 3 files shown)


certik:
Should be squash merged.

PR #10021, opened 2026-02-12
fix: use local ExternalSymbol for type_declaration on call temporaries

### LINKED ISSUE #10020: Segfault upon compilation of modules and main program combination
This issue is of higher priority then my remaining reports. It's a variation of #10006 that still fails. The following combination of three files leads to a segfault upon compilation.

`$ cat mytypemod.f90`
```
module mytypemod

   type, abstract :: AbsType
   end type AbsType

   type :: MyType
   end type MyType

   interface MyType
      procedure :: constructor
   end interface MyType
   
contains
   
   end function constructor

end module mytypemod
```
`$ cat getmyobjmod.f90 `

```
module getmyobjmod

   use mytypemod, only: AbsType, MyType
   
contains

   function get_my_obj() result(myobj)
      class(MyType), allocatable :: myobj
      myobj = MyType(get_obj())
   end function get_my_obj

   function get_obj() result(obj)
   end function get_obj

end module getmyobjmod
 ```
`$ cat main.f90 `
```
program main
   use getmyobjmod, only: get_my_obj
end program main

```
`$ lfortran -c --separate-compilation mytypemod.f90`
`$ lfortran -c --separate-compilation getmyobjmod.f90`
`$ lfortran -c --separate-compilation main.f90`
```
```

certik:
The previous commit (b4eb10eb2) used symbol_get_past_external() to set m_type_declaration on lowered call temporaries, which resolved to the actual Struct symbol in the foreign module's symbol table. When the module was serialized to a modfile, the foreign symtab ID was written, causing an assertion failure during deserialization in separate compilation (id_symtab_map lookup failed).

Use import_struct_type() instead, which finds or creates an ExternalSymbol in the current module's scope, ensuring m_type_declaration always references a symbol within the serialized module's own symbol table hierarchy.

Fixes #10020.


certik:
Unfortunately, even though this passed all the tests, it broke tests after merging.

certik:
However it does pass at the CI: https://github.com/lfortran/lfortran/actions/runs/21976983959

PR #10023, opened 2026-02-12
fix: initialize data buffer for non-allocatable character descriptor …

### LINKED ISSUE #10019: Segfault in accessing a substring
The following is a valid Fortran program.
`$ cat string.f90 `
```
program test

   character(*), parameter :: str = 'This is a string'
   integer :: sig

   print *, 'sum_string(str): ',  sum_string(str)
      
contains

   function sum_string(str) result(sig)
      character(len=*), intent(in) :: str
      integer                      :: sig

      character, dimension(len(str)) :: tmp
      integer :: i

      do i = 1, len(str)
         tmp(i) = str(i:i)
      end do

      sig = sum(ichar(tmp))

   end function sum_string

end program test

```
When compiled with flang-20 it prints the following:
`$ flang-20 string.f90 -o string`
`$ ./string`
 ```
sum_string(str):  1484
```

With LFortran the result presently is as follows:
`$ lfortran string.f90`
```
```

certik:
…arrays

When declaring a non-allocatable character array with runtime size (e.g., `character, dimension(len(str)) :: tmp`), the LLVM codegen allocated an array of string_descriptor structs but left the data pointers inside those descriptors uninitialized. Any subsequent element access (e.g., `tmp(i) = str(i:i)`) would pass a NULL pointer to `_lfortran_strcpy`, triggering:

    LFORTRAN ERROR: Runtime Error : Non-allocatable string isn't allocated.

After `fill_array_details` sets up the descriptor array, allocate a contiguous character data buffer of `total_elements * char_length` bytes and store it in `descriptors[0].data` with the proper length, matching the access pattern used by element indexing codegen.

Fixes #10019.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -5167,43 +5167,35 @@ public:
             fill_array_details(ptr_typ, ptr, llvm_data_type, m_dims, n_dims, is_data_only);
             // For non-allocatable DescriptorArray character arrays,
             // fill_array_details allocated an array of string_descriptors
-            // but their data pointers are uninitialized. Allocate a
-            // contiguous character data buffer and store it in
-            // descriptors[0].data so element accesses work correctly.
-            if (ASRUtils::is_character(*m_type) && !is_data_only) {
+            // but their data pointers are uninitialized. Set the string
+            // length on descriptors[0], then allocate a contiguous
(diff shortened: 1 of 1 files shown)


certik:
@assem2002 let me know if you think this is the right fix for this string bug.

certik:
Let's merge it to keep things moving. @assem2002 if you find any issues here, just let me know, we can fix it later.

PR #10024, opened 2026-02-12
Fix FunctionType comparison with null pointers

### LINKED ISSUE #10018: Compilation segfaults for a subroutine with a procedure pointer argument
The following valid code presently doesn't compile with LFortran, but compiles with flang-20.
`$ cat test.f90 `
```
module test

   abstract interface
      subroutine factory()
      end subroutine factory
   end interface
     
contains
   
   subroutine client(val)
      procedure(factory), pointer, intent(out) :: val
      val => null()      
   end subroutine client

end module test
```
`$ lfortran -c test.f90 `

certik:
get_function_from_expr() returns nullptr for PointerNullConstant expressions (e.g. `val => null()`). The FunctionType cases in types_equal(), types_equal_with_substitution(), and check_equal_type() dereferenced this result without a null check, causing a segfault.

Add null guards: when either function pointer is nullptr, fall back to type-only comparison, consistent with the existing expr==nullptr paths.

Fixes #10018.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -4254,21 +4254,7 @@ inline bool types_equal(ASR::ttype_t *a, ASR::ttype_t *b, ASR::expr_t* a_expr, A
                 ASR::Function_t* right_func = const_cast<ASR::Function_t*>(ASRUtils::get_function_from_expr(b_expr));
 
                 if (left_func == nullptr || right_func == nullptr) {
-                    // Fall back to type-only comparison (e.g. when expr is PointerNullConstant)
-                    if( a2->n_arg_types != b2->n_arg_types ||
-                        (a2->m_return_var_type != nullptr && b2->m_return_var_type == nullptr) ||
-                        (a2->m_return_var_type == nullptr && b2->m_return_var_type != nullptr) ) {
-                        return false;
-                    }
-                    for (size_t i = 0; i < a2->n_arg_types; i++) {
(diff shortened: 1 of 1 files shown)


src/libasr/asr_utils.h
```diff
@@ -4396,6 +4414,23 @@ inline bool types_equal_with_substitution(ASR::ttype_t *a, ASR::ttype_t *b,
                 ASR::Function_t* left_func = const_cast<ASR::Function_t*>(ASRUtils::get_function_from_expr(a_expr));
                 ASR::Function_t* right_func = const_cast<ASR::Function_t*>(ASRUtils::get_function_from_expr(b_expr));
 
+                if (left_func == nullptr || right_func == nullptr) {
+                    if( a2->n_arg_types != b2->n_arg_types ||
+                        (a2->m_return_var_type != nullptr && b2->m_return_var_type == nullptr) ||
+                        (a2->m_return_var_type == nullptr && b2->m_return_var_type != nullptr) ) {
```
jinangshah21: Here also we can use similar as below to compare types instead of writing repeated code:
```cpp
if (left_func == nullptr || right_func == nullptr) {
            return types_equal(x, y, nullptr, nullptr, true);
     }
```

certik: Nice, indeed! Fixed in 37bdc28c3b2cfb468a53d5f61e9cf295bddf5677.

jinangshah21 (approved):

PR #10025, opened 2026-02-12
tests: consolidate assumed-size error subroutines in continue_compilation_1

amritamishra01:
This PR consolidates the assumed-size error subroutines in
`tests/errors/continue_compilation_1.f90` into a single
`ubound_assumed_size` subroutine.
Changes:
- Merged `assumed_size_star_pos_1` and `assumed_size_star_pos_2`
  into `ubound_assumed_size`
- Preserved file length to avoid shifting line numbers
- Regenerated the corresponding reference files using the current upstream build

No functional behavior is changed — this is a refactor to reduce duplication
while keeping the expected diagnostics intact.

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -74,7 +74,7 @@ contains
         character(len=2, kind=c_char), intent(in) :: c
     end subroutine s
 
-    subroutine assumed_size_errors(a, b, c, d, e)
+    subroutine ubound_assumed_size(a, b, c, d, e)
         real :: a(*)
         real :: b(*)
         real :: c(10, *)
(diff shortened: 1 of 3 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -74,23 +74,23 @@ subroutine s(c) bind(c)
...
         print *, ubound(b)
         print *, ubound(c, 2)
     end subroutine
 
-    subroutine assumed_size_star_pos_1(a)
-        real, intent(in) :: a(*, 10)
-    end subroutine
 
-    subroutine assumed_size_star_pos_2(a)
-        real :: a(*, 10)
```
Pranavchiku, line 92: These are some combinations which we intentionally put to validate `--continue-compilation` is working or not.

Pranavchiku (commented):
Thank you for working on this! However, I feel this change is not required, hence marking this PR as draft.

amritamishra01:
Thank you for the feedback, @Pranavchiku . I understand that the redundancy is necessary to test the error recovery of --continue-compilation. I'm closing this PR and will look for other issues to contribute to.

PR #10022, opened 2026-02-12
fix: external symbol lookup for procedure variables

Pranavchiku:
Fixes https://github.com/lfortran/lfortran/issues/9996. Fixes #10000. Continuation of https://github.com/lfortran/lfortran/pull/9962 and #9998

PR #10032, opened 2026-02-13
fix: external symbol lookup for procedure variables

certik:
A simplified version of https://github.com/lfortran/lfortran/pull/10022.


certik:
It's weird that the tests failed here, but passed in https://github.com/lfortran/lfortran/pull/10038. I merged that one, and we'll watch the CI.

PR #10037, opened 2026-02-13
fix: Throw error when associated() is called with non-pointer argument

### LINKED ISSUE #6617: Error : `associated` should only accept pointers
MRE
```.f90
program associate____
    real(8), pointer :: p2(:, :) => null()
    integer, allocatable :: ppp(:)


    print *, associated(p2)
    print *, associated(ppp) ! <<<

end program
```

```Console
(lf) assem@assem-PC:~/lfortran_test/lfortran/integration_tests$ gfortran associate_07.f90 
associate_07.f90:13:24:

   13 |     print *, associated(ppp)
      |                        1
Error: ‘pointer’ argument of ‘associated’ intrinsic at (1) must be a POINTER
(lf) assem@assem-PC:~/lfortran_test/lfortran/integration_tests$ ../../../LFortranUpStream/lfortran/src/bin/lfortran associate_07.f90 
F
T

```

Jatinagarwal24:
fixes #6617

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -262,8 +262,8 @@ program continue_compilation_1
     integer :: bad_x
     type(ieee_class_type) :: ieee_cls
 
-
-
+    real(8), pointer :: p2(:, :) => null()
+    integer, allocatable :: ppp(:)
 
 
 
@@ -564,7 +564,8 @@ program continue_compilation_1
 
     integer, intent(out) :: out_intent
     integer, intent(in) :: in_intent
-    
+    print *, associated(p2)
+print *, associated(ppp)   
     contains
     subroutine sub(f)
         interface
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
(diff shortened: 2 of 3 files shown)


tests/reference/asr-array12-cb81afc.json
```diff
@@ -5,9 +5,9 @@
...
-    "stdout": "asr-array12-cb81afc.stdout",
-    "stdout_hash": "c41a28f623bb20eb87710bb292e72107ecee5d935c8fab002e17faab",
-    "stderr": null,
-    "stderr_hash": null,
-    "returncode": 0
+    "stdout": null,
+    "stdout_hash": null,
+    "stderr": "asr-array12-cb81afc.stderr",
+    "stderr_hash": "d4990d7f1aa0f884750937a1f9e3bb25833eb05e2fa8720c28a4e63f",
+    "returncode": 2
```
certik: I think this is problematic --- must investigate what this test was testing, since it now gives an error.

HarshilShah1804: I believe the issue is that `c_ptr` is not currently allowed in the associated statement after the changes in the PR. All the additional changes in the PR seem to be related to handling c_ptr. It looks like a small fix — if we allow c_ptr, everything should work fine.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -11200,6 +10942,13 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         std::vector<std::string> kwarg_names = {"pointer", "target"};
         handle_intrinsic_node_args(x, args, kwarg_names, 1, 2, "associated");
         ASR::expr_t *ptr_ = args[0], *tgt_ = args[1];
+        if (!ASRUtils::is_pointer(ASRUtils::expr_type(ptr_))) {
```
HarshilShah1804, line 10945: You can allow `c_ptr` here when the function is `c_associated` instead of associated. This should fix it.

Issue #10056, opened 2026-02-13
CI: Move all wasm and llvm_* integration tests to quick tests


certik:
So that quick tests are testing all integration tests with full configuration (all options).

PR #10060, opened 2026-02-14
CI: migrate integration tests for llvm* backend to quick checks

jinangshah21:
Towards #10056
With this integration tests will run only in quick checks (So LLVM 11 and 21 only, not for all llvm). Is that fine?

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -436,30 +436,6 @@ jobs:
 
             cmake --build . -j16 --target install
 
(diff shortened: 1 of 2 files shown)


jinangshah21:
It takes 5 minutes more on Ubuntu LLVM-11/21 tests and 10 minutes more on macOS LLVM-11/21 to run all integration test with all llvm* backend

certik (approved):
But the exhaustive checks are now faster --- I think they effectively become "third-party" tests. So this distinction is more logical. And we can focus on the quick tests to make them faster too --- possibly running the test suite in release mode most of the time.

PR #10064, opened 2026-02-14
fix: multi-dimensional pointer reference to 1D array

### LINKED ISSUE #10063: Segfault in referencing 2D pointer to 1D array
MRE
```fortran
program main
  implicit none

  integer :: n1, n2, i
  real, allocatable, target :: data(:)
  real, pointer :: p(:,:)

  n1 = 3
  n2 = 2

  allocate(data(n1*n2))

  do i = 1, n1*n2
     data(i) = real(i)
  end do

  p(1:n1, 1:n2) => data(1:n1*n2)

end program main
```

HarshilShah1804:
Fixes #10063 
Towards #4748


HarshilShah1804:
I don't think we need exhaustive checks on this PR as no third party code uses pointer section target association. What do you suggest @certik ?

certik:
Better safe than sorry. We have lots of PRs and any failure in main then derails everything, so let's run it.

PR #10065, opened 2026-02-14
rem: remove implicit deallocate before procedure call

jinangshah21:
This removes `ImplicitDeallocate` before a procedure call because we already deallocate `intent(out)` arguments at function entries.


jinangshah21:
I found this while working on #10002, this doesn't caused any failure previously but it removes one abundant deallocate stmt

jinangshah21:
Not at backend, in function body at starting we have deallocate node.

PR #10073, opened 2026-02-14
fix: prevent double-wrap of m_dt in SubrCall

certik:
When a FunctionCall returning a struct via a procedure pointer component (e.g., `self%get_result()`) was lowered to a SubroutineCall by the subroutine_from_function pass, make_SubroutineCall_t_util unconditionally wrapped a_dt in a StructInstanceMember node. If a_dt was already a StructInstanceMember (as it is for procedure pointer components), this created a double nesting, causing a GEP type mismatch assertion failure in LLVM codegen.

Add a guard to skip the wrapping when a_dt is already a StructInstanceMember.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -7317,7 +7317,8 @@ static inline ASR::asr_t* make_SubroutineCall_t_util(
 
     if( a_dt && ASR::is_a<ASR::Variable_t>(
         *ASRUtils::symbol_get_past_external(a_name)) &&
-        ASR::is_a<ASR::FunctionType_t>(*ASRUtils::symbol_type(a_name)) ) {
+        ASR::is_a<ASR::FunctionType_t>(*ASRUtils::symbol_type(a_name)) &&
+        !ASR::is_a<ASR::StructInstanceMember_t>(*a_dt) ) {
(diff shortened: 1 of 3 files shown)


certik:
This fails in main and now passes. Extracted from https://github.com/lfortran/lfortran/pull/10051.

src/libasr/asr_utils.h
```diff
@@ -7317,7 +7317,8 @@ static inline ASR::asr_t* make_SubroutineCall_t_util(
 
     if( a_dt && ASR::is_a<ASR::Variable_t>(
         *ASRUtils::symbol_get_past_external(a_name)) &&
-        ASR::is_a<ASR::FunctionType_t>(*ASRUtils::symbol_type(a_name)) ) {
+        ASR::is_a<ASR::FunctionType_t>(*ASRUtils::symbol_type(a_name)) &&
+        !ASR::is_a<ASR::StructInstanceMember_t>(*a_dt) ) {
```
jinangshah21, line 7321: I think there may be possibility of cases like: `call a%b%c%func()`. So maybe we nested StructInstanceMember

certik: Thanks, I made this https://github.com/lfortran/lfortran/issues/10085.

PR #10074, opened 2026-02-14
Fix associate with runtime-sized arrays

certik:
When an associate block binds a name to an implied-do array constructor (e.g., [(expr(i), i = 1, n)]), the resulting array has dimensions determined at runtime. The target variable needs to be Allocatable so the assignment can allocate memory for it.

Wrap non-lvalue array expressions with empty dimensions in Allocatable_t, and unconditionally enable realloc_lhs for the generated assignment so the target is properly allocated regardless of compiler options.


certik:
The test fails in main, now works. Extracted from https://github.com/lfortran/lfortran/pull/10051.

certik:
All tests passed, I resolved conflicts in integration tests name conflict, ran tests locally and merged to save CI.

PR #10075, opened 2026-02-14
Fix struct temp missing sub-expr visit

certik:
In traverse_call_args, the array temporary branch calls visit_call_arg() to recursively simplify sub-expressions before creating a temporary, but the struct temporary branch was missing this step. This caused an assertion failure in the verifier when a struct call argument contained nested array expressions (e.g., a user-defined operator calling pack() with a StructInstanceMember array argument like diagnoses%passed), since those inner expressions were never simplified to Var nodes.

### DIFF
--- a/integration_tests/operator_overloading_23.f90
+++ b/integration_tests/operator_overloading_23.f90
@@ -1,0 +1,68 @@
+module operator_overloading_23_mod
+  implicit none
+  type :: string_t
+    character(len=:), allocatable :: s
+  end type
+  interface operator(.cat.)
+    pure function concat(strings) result(r)
+      import string_t
+      type(string_t), intent(in) :: strings(:)
+      type(string_t) :: r
+    end function
+  end interface
+
+  type :: diag_t
+    logical :: passed = .false.
+  end type
+  interface diag_t
+    pure function make_diag(test_passed, diagnostics_string) result(d)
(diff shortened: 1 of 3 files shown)


certik:
Extracted from https://github.com/lfortran/lfortran/pull/10051. The test fails in main, and now passes.

integration_tests/operator_overloading_23.f90
```diff
@@ -0,0 +1,68 @@
...
+  logical, intent(in) :: test_passed
+  type(string_t), intent(in) :: diagnostics_string
+  type(diag_t) :: d
+  d%passed = test_passed
+end function
+
+program operator_overloading_23
+  use operator_overloading_23_mod, only: diag_t, aggregate
+  implicit none
+  type(diag_t) :: diagnoses(2), result
```
czgdp1807, line 55: A small suggestion - `type(diag_t) :: diagnoses(2)` lies on the simpler side. A corner case usually is, `type(struct_t) :: a(4)` and then the usage looks like, `a(3)%b(2)%member = .false.`. If the logic works with these corner cases, then it's assured to be correct.

certik: Thanks for the suggestion, I made this https://github.com/lfortran/lfortran/issues/10084.

PR #10077, opened 2026-02-14
doc: update cast_kind and Cast documentation

jinangshah21:
As suggested in #10002, I have updated doc, let me know if it is fine or not.

### DIFF
--- a/doc/src/asr/asr_nodes/expression_nodes/Cast.md
+++ b/doc/src/asr/asr_nodes/expression_nodes/Cast.md
@@ -7,7 +7,7 @@ Cast to different type and/or kind.
 ### Syntax
 
 ```fortran
-Cast(expr arg, cast_kind kind, ttype type, expr? value, expr? dest)
+Cast(expr arg, cast_kind kind, ttype type, expr? value, expr? dest_struct)
 ```
 
 ### Arguments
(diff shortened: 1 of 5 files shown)


doc/src/asr/asr_nodes/expression_nodes/Cast.md
```diff
@@ -7,18 +7,40 @@ Cast to different type and/or kind.
...
+
+class(base), allocatable :: var
+select type(var)
+    type is (derived)
+        print *, var%y  ! ClassToStruct
+    class is (derived)
+        print *, var%y  ! ClassToClass
+end select
+```
+Here inside select type block, `var` will get wrapped inside a Cast node. This Cast node will have `class(base)` as source type and `type(derived)` as destination type. Now in backend we need both llvm types and for struct llvm types we can't get it from just destination `ttype`, we need their symbol, so for have struct symbol of destination type, we store that symbol in `dest_struct` (i.e. in this case is `derived` symbol). Similar for `ClassToClass` casting (i.e. in above example `class(base)` to `class(derived)`)
```
certik, line 42: So `dest_struct` is either null, or a Var that points to a symbol in the symbol table for the destination type?

If so, then also add a verify check in the verify pass to enforce this.

certik (approved):
I think this looks great, much better!

PR #10078, opened 2026-02-14
fix: update procedure variable ttype in subroutine_from_function pass

### LINKED ISSUE #10069: Separate compilation segfaults on code that uses a procedure pointer
`$ cat test.f90 `
```
module absmod

   type, abstract :: AbsType
   end type AbsType

   abstract interface
      function func_intfc(c) result(res)
         import; implicit none
         class(AbsType), intent(in)  :: c
         class(*),       allocatable :: res
      end function func_intfc
   end interface

end module absmod

module mytypemod

   use absmod, only: AbsType, func_intfc

   type, public, extends(AbsType) :: MyType
   contains
      procedure :: method
   end type MyType

contains

   function method(self) result(res)
      class(MyType), intent(in)  :: self
      class(*),      allocatable :: res
      procedure(func_intfc), pointer :: func
      res = func(self)
   end function method

end module mytypemod
```

```
```

jinangshah21:
Fixes #10069


certik:
Rename procedure_27.f90 to procedure_28.f90.

PR #10079, opened 2026-02-14
fix: Throw error when integer exponentiation overflows its kind at compile time

### LINKED ISSUE #7074: Integer exponentiation overflow not detected at compile-time
```fortran
program main
    integer(8) :: mask
    mask = 2_8**63
    print *,  mask
end program
```
```console
$ gfortran b.f90 && ./a.out
b.f90:3:14:

    3 |     mask = 2_8**63
      |              1
Error: Result of exponentiation at (1) exceeds the range of INTEGER(8)
$ lfortran b.f90
-9223372036854775808
```

Jatinagarwal24:
fixes #7074

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -19,7 +19,6 @@
 #include <set>
 #include <map>
 #include <limits>
-#include <limits>
 
 using LCompilers::diag::Level;
 using LCompilers::diag::Stage;


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13339,6 +13319,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
             }
 
+            if (op == ASR::Pow) {
+                int64_t kind = 4;
+                if (ASR::is_a<ASR::Integer_t>(*dest_type)) {
+                    kind = ASR::down_cast<ASR::Integer_t>(dest_type)->m_kind;
+                }
+                return ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, left->base.loc,
+                        power_check_overflow(left_value, right_value, loc, kind), dest_type));
+            }
```
certik, line 13328: Is this replacing `2**4` with 16? If so, we can't do it like this --- rather, we need to store the constant as the `value` of the `2**4` expr.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13286,6 +13195,77 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         return false;
     }
 
+    int64_t power_check_overflow(int64_t base, int64_t exp, const Location& loc, int64_t kind) {
+        if (exp < 0) {
+            if (base == 0) {
+                diag.add(Diagnostic("Division by zero", Level::Error, Stage::Semantic, {Label("", {loc})}));
+                throw SemanticAbort();
```
certik, line 13202: Add a test for this error.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13286,6 +13195,77 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                int64_t tmp;
+                if (safe_mul(b, b, tmp)) {
+                     diag.add(Diagnostic("Integer overflow in exponentiation", Level::Error, Stage::Semantic, {Label("", {loc})}));
+                     throw SemanticAbort();
+                }
+                b = tmp;
+            }
+        }
+        return result;
+    }
```
certik, line 13267: This function has a lot of logic that is hard for me to easily verify. In any case, all code paths in this function must be tested, right now you only test one path.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13286,6 +13195,77 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                } else {
+                    if (a < max_val / b) overflow = true;
+                }
+            }
+            if (overflow) return true;
+            res = a * b;
+            return false;
+        };
+        
+        while (e > 0) {
```
certik, line 13249: What is the purpose of this loop? How does it work?

PR #10082, opened 2026-02-15
Create a temporary for select type expr

### LINKED ISSUE #10080: Separate compilation segfaults on code that assigns to a `class(*)` variable using a `select type` statement
`$ cat test.f90 `
```
module test

   type, abstract :: AbsGetter
   contains
      procedure(get), deferred :: get
   end type AbsGetter

   abstract interface
      function get(self,key) result(res)
         import
         class(AbsGetter), intent(in)  :: self
         character(*),     intent(in)  :: key
         class(*),         allocatable :: res      
      end function get
   end interface
   
   type, abstract :: AbsType
   end type AbsType
   
   type :: MyType
      class(AbsType), allocatable :: ob
   end type MyType
   
contains
   
   function create(c) result(res)
      class(AbsGetter), intent(in)  :: c
      class(*),         allocatable :: res

      select type( obj => c%get("AbsType") )
      class is (AbsType)
         res = MyType(obj)
      end select

   end function create

end module test
```

```
```

certik:
The codegen expects Var or StructInstanceMember as the selector, not a FunctionCall. In this commit we make this transformation already in AST->ASR. It has the advantage that the bug fix is short. The disadvantage is that this transformation is lowering and we should not be lowering in AST->ASR. The right place for this is either an ASR pass, or possibly the backend if it needs to be extended to handle FunctionCall. So far all my solutions in ASR or LLVM were much more complex. So I am submitting this PR to get us unblocked and we can easily replace this code in ast_body_visitor.cpp with another solution in an ASR pass or LLVM and we can actually easily judge if it is more maitainable by comparing the code removed and code added. So this PR might be a good first step.

Fixes #10080.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1609,6 +1609,7 @@ RUN(NAME select_type_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_23 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME select_type_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_type_25 LABELS gfortran llvm)
+RUN(NAME select_type_26 LABELS gfortran llvm)
 
 RUN(NAME program_02 LABELS gfortran llvm)
 RUN(NAME program_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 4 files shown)


certik:
To see what it would take to move the logic from AST->ASR to an ASR pass, see https://github.com/lfortran/lfortran/pull/10083. It works, but it adds a lot more code to maintain, so I actually think this PR #10082 is a better solution for now.

jinangshah21 (approved):
I think this is fine

PR #10087, opened 2026-02-15
perf: optimize integer power operations for exponents 0-8

### LINKED ISSUE #9989: euler0.f90 is very slow to run
From https://fortran-lang.discourse.group/t/euler-sum-of-powers-conjecture-tutorial/10678/7:

```fortran
program euler0
   ! 27^5 + 84^5 + 110^5 + 133^5 = 144^5 = 61917364224
   implicit none

   integer, parameter :: ik = selected_int_kind(18)  ! extended precision integer kind.
   integer, parameter :: imax = 50 ! 6207                 ! search for solutions i=1:imax
   integer, parameter :: findmax = 1                 ! max number of solutions.

   real(selected_real_kind(14)) :: time0, time1
   integer :: nfound
   integer(ik) :: i, j, k, l, m
   character(*), parameter :: tfmt='(/a,1x,f0.4)'
   character(*), parameter :: fmtf = '(i0,": ",3(i0,"^5 + "),i0,"^5 = ",i0,"^5 = ",i0)'

   call cpu_time( time0 )
   nfound = 0
   outer: do i = 1, imax
      do j = 1, i
         do k = 1, i
            do l = 1, i
               do m = 1, i
                  if ( j**5 + k**5 + l**5 + m**5 == i**5 ) then
                     nfound = nfound + 1
                     print fmtf, nfound, j, k, l, m, i, i**5
                     if ( nfound >= findmax ) exit outer
                  endif
               enddo
            enddo
         enddo
      enddo
   enddo outer
   call cpu_time( time1 )
   print tfmt, 'cpu time:', (time1-time0)
end program euler0
```
Notice that I lowered `imax` from 6207 to just 50. Then:
```console
$ lfortran --no-style-suggestions --fast euler0.F90

cpu time: 1.2585
$ gfortran euler0.F90 && ./a.out

cpu time: .0979
$ gfortran -march=native -O3 -ffast-math euler0.F90 && ./a.out

cpu time: .0213
```

Something is very very slow, so we need to investigate and fix it.

opixdown:
fixes #9989 

for investigation outcome see https://github.com/lfortran/lfortran/issues/9989#issuecomment-3903547698

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -11290,7 +11290,14 @@ public:
                 int64_t exponent_const =  INT64_MAX;
                 ASRUtils::extract_value(x.m_right, exponent_const);
                 // Handle simple-common exponent cases for faster computation.
-                if (exponent_const == 2) {
+                if (exponent_const == 0) {
+                    // x^0 = 1
+                    tmp = llvm::ConstantInt::get(left_val->getType(), 1);
+                } else if (exponent_const == 1) {
+                    // x^1 = x
+                    tmp = left_val;
+                } else if (exponent_const == 2) {
+                    // x^2 = x * x
                     tmp = builder->CreateMul(left_val, left_val, "simplified_pow_operation");
                 } else if (exponent_const == 3) {
                     tmp = builder->CreateMul(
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11356,7 +11307,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    llvm::Value* x2 = builder->CreateMul(left_val, left_val, "simplified_pow_operation");
+                    llvm::Value* x4 = builder->CreateMul(x2, x2, "simplified_pow_operation");
+                    tmp = builder->CreateMul(x4, left_val, "simplified_pow_operation");
+                } else if (exponent_const > 0 && exponent_const <= 10) {
+                    // For small positive exponents (6-10), use repeated multiplication
+                    // This is faster than calling the pow function
+                    tmp = left_val;
+                    for (int64_t i = 1; i < exponent_const; i++) {
+                        tmp = builder->CreateMul(tmp, left_val, "simplified_pow_operation");
+                    }
```
certik, line 11325: Why not use the same trick as you did for floating point for powers 6-8? Is it not as fast for integers?

certik:
Let's not touch this PR anymore, let's run all tests.

certik:
In a new PR, why don't you submit a benchmark, figure out where to put it. And let's benchmark different powers and see.

Also possibly we need to make this configurable, possibly only do it with `--fast`. But for now this is good, to unblock us and fix that slow benchmark.

src/libasr/codegen/asr_to_llvm.cpp
opixdown: you are right i initially thought the loop would be simpler for integers but it will make it slow for ex. x**6 is 3 muls (x²→x⁴→x⁶) but it will 5 sequential muls in the loop. we should do the same here too.

opixdown:
> In a new PR, why don't you submit a benchmark, figure out where to put it. And let's benchmark different powers and see.
> 
> Also possibly we need to make this configurable, possibly only do it with `--fast`. But for now this is good, to unblock us and fix that slow benchmark.

okayy i will see how to do benchmark then i will share it

src/libasr/codegen/asr_to_llvm.cpp
opixdown: made a pr for it i think now we can proceed for benchmarking

PR #10094, opened 2026-02-15
Fix select type modfile serialization

### LINKED ISSUE #10093: Separate compilation segfaults on code that assigns to a class(*) variable and is used by a main program
This is an extension of #10080, contained in two files, that still fails to compile.

`$ cat test.f90 `

```
module test

   type, abstract :: AbsGetter
   contains
      procedure(get), deferred :: get
   end type AbsGetter

   abstract interface
      function get(self,key) result(res)
         import
         class(AbsGetter), intent(in)  :: self
         character(*),     intent(in)  :: key
         class(*),         allocatable :: res      
      end function get
   end interface
   
   type, abstract :: AbsType
   end type AbsType

   type :: MyType
      class(AbsType), allocatable :: ob
   end type MyType

contains

   function create(c) result(res)
      class(AbsGetter), intent(in)  :: c
      class(*),         allocatable :: res

      select type( obj => c%get("AbsType") )
      class is (AbsType)
         res = MyType(obj)
      end select
      
   end function create

end module test
```

`$ cat main.f90 `

```
program main
   use test, only: create
end program main
```
`$ lfortran -c --separate-compilation main.f90 `

```
```

certik:
When creating a temp variable for a select type selector involving a FunctionCall, the type_declaration was obtained from the called function's scope rather than the current scope. This cross-scope reference caused deserialization to fail when loading the module file.

Resolve type_declaration to the current scope by looking up the symbol by name, creating it locally if needed.  Also improve the error message in read_symbol() to report a clear ICE instead of an assert on missing symtab IDs.

Fixes #10093.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3221,24 +3221,15 @@ public:
                 if (local_decl != nullptr) {
                     type_decl = local_decl;
                 } else {
-                    // Symbol not found in scope hierarchy; create it locally.
-                    // This happens for ~unlimited_polymorphic_type when the
-                    // current scope has no class(*) variable declaration yet.
-                    SymbolTable *parent_scope = current_scope;
-                    current_scope = al.make_new<SymbolTable>(parent_scope);
-                    ASR::asr_t* dtype = ASR::make_Struct_t(al, x.base.base.loc,
-                        current_scope, s2c(al, decl_name), nullptr,
-                        nullptr, 0, nullptr, 0, nullptr, 0,
-                        ASR::abiType::Source, ASR::accessType::Public,
-                        false, true, nullptr, 0, nullptr, nullptr);
-                    ASR::symbol_t* struct_symbol = ASR::down_cast<ASR::symbol_t>(dtype);
-                    ASR::ttype_t* struct_type = ASRUtils::make_StructType_t_util(
-                        al, x.base.base.loc, struct_symbol, false);
-                    ASR::Struct_t* struct_ = ASR::down_cast<ASR::Struct_t>(struct_symbol);
-                    struct_->m_struct_signature = struct_type;
-                    parent_scope->add_symbol(decl_name, struct_symbol);
-                    current_scope = parent_scope;
-                    type_decl = struct_symbol;
+                    // Symbol not found in current scope; add an ExternalSymbol
+                    // referencing the original struct from its owner scope.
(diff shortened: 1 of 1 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3211,6 +3211,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                ASR::symbol_t* local_decl = current_scope->resolve_symbol(decl_name);
+                if (local_decl != nullptr) {
+                    type_decl = local_decl;
+                } else {
+                    // Symbol not found in scope hierarchy; create it locally.
+                    // This happens for ~unlimited_polymorphic_type when the
+                    // current scope has no class(*) variable declaration yet.
+                    SymbolTable *parent_scope = current_scope;
+                    current_scope = al.make_new<SymbolTable>(parent_scope);
+                    ASR::asr_t* dtype = ASR::make_Struct_t(al, x.base.base.loc,
```
jinangshah21: Why we are creating new scope here, I think we need to check if external symbol to original struct is present in current scope or not. If not we need to add ExternalSymbol in current scope.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3211,6 +3211,36 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             ASR::ttype_t* selector_type = ASRUtils::expr_type(m_selector);
             std::string tmp_name = current_scope->get_unique_name("~select_type_selector_tmp_");
             ASR::symbol_t* type_decl = ASRUtils::get_struct_sym_from_struct_expr(m_selector);
+            // The type_declaration obtained above may reference a symbol
+            // in a sibling scope (e.g. the called function's scope). We must
+            // resolve it to a symbol accessible from current_scope so that
+            // serialization (modfile) does not produce cross-scope references.
+            if (type_decl != nullptr) {
+                std::string decl_name = ASRUtils::symbol_name(type_decl);
+                ASR::symbol_t* local_decl = current_scope->resolve_symbol(decl_name);
```
jinangshah21:
```suggestion
                ASR::symbol_t* local_decl = current_scope->get_symbol(decl_name);
```

I think this way we will now if external symbol to struct is present in current scope or not.

certik: Good idea, fixed.

src/lfortran/semantics/ast_body_visitor.cpp
certik: Good catch. I greatly simplified this part.

PR #10101, opened 2026-02-16
Fix assignment overload deserialization

certik:
When a submodule contains a function that uses the parent module's overloaded assignment(=), the SubroutineCall in m_overloaded referenced the parent module's symbol table directly. During .smod deserialization, that symbol table ID was not available, causing:

  LCompilersException: symbol
  'assign_character_to_string_t' references symbol
  table with ID 4 which has not been deserialized yet

Fix: In process_overloaded_assignment_function(), create an ExternalSymbol in the current scope when the matched subroutine is not locally available, matching the pattern already used by process_overloaded_read_write_function(). Use get_symbol() (local-only lookup) instead of resolve_symbol() (parent-traversing lookup) to ensure the ExternalSymbol is always created when needed.

### DIFF
--- a/integration_tests/intrinsics_346.f90
+++ b/integration_tests/intrinsics_346.f90
@@ -1,4 +1,4 @@
-module asciimod ! f95-compliant file asciiB.f90: [] only in character strings
+module intrinsics_346_asciimod ! f95-compliant file asciiB.f90: [] only in character strings
     implicit none
     public iascii,ascii
     interface ascii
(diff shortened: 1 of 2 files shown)


certik:
The two tests fail with duplicate module name, however it is not clear what the duplicate is:
```
FAIL: Found a duplicate module name
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 206, in <module>
    main()
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 182, in main
Name: procedure assign_character_to_string_t
Filename: submodule_19c.f90
    check_module_names()
  File "/home/runner/work/lfortran/lfortran/integration_tests/./run_tests.py", line 152, in check_module_names
    raise Exception("Duplicate module names")
Exception: Duplicate module names
```
Possibly the checking script is wrong.

certik:
This was indeed a bug in the check duplicate module script. I also rebased and resolved the conflict.

PR #10103, opened 2026-02-16
Fix ASR verify error for re-exported module symbols

certik:
When module A defines a symbol and module B re-exports it via
'use A, only: symbol', and module C uses it via 'use B, only: symbol',
the ExternalSymbol in C has m_module_name pointing to A (the origin).
However, C's dependencies only listed B (the direct import).

This caused the ASR verify pass to fail with:
      Module C dependencies must contain A because a function present
      in it is getting called in C.

Fix: when importing a re-exported ExternalSymbol (in both import_all
and import_symbols_util), add the origin module as a transitive
dependency if it differs from the directly imported module.

Depends on https://github.com/lfortran/lfortran/pull/10102.


certik:
The dependency list is a list of strings, so I think we have to do a string check?

Pranavchiku:
Yes, exactly.

PR #10104, opened 2026-02-16
perf: optimize power operations for exponents 6-8

opixdown:
addresses https://github.com/lfortran/lfortran/pull/10087#discussion_r2809348708

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -11372,13 +11372,22 @@ public:
                     llvm::Value* x2 = builder->CreateMul(left_val, left_val, "simplified_pow_operation");
                     llvm::Value* x4 = builder->CreateMul(x2, x2, "simplified_pow_operation");
                     tmp = builder->CreateMul(x4, left_val, "simplified_pow_operation");
-                } else if (exponent_const > 0 && exponent_const <= 10) {
-                    // For small positive exponents (6-10), use repeated multiplication
-                    // This is faster than calling the pow function
-                    tmp = left_val;
-                    for (int64_t i = 1; i < exponent_const; i++) {
-                        tmp = builder->CreateMul(tmp, left_val, "simplified_pow_operation");
-                    }
+                } else if (exponent_const == 6) {
+                    // x^6 = x^4 * x^2 = (x*x)*(x*x) * (x*x)
+                    llvm::Value* x2 = builder->CreateMul(left_val, left_val, "simplified_pow_operation");
+                    llvm::Value* x4 = builder->CreateMul(x2, x2, "simplified_pow_operation");
+                    tmp = builder->CreateMul(x4, x2, "simplified_pow_operation");
+                } else if (exponent_const == 7) {
+                    // x^7 = x^4 * x^2 * x
+                    llvm::Value* x2 = builder->CreateMul(left_val, left_val, "simplified_pow_operation");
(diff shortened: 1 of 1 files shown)


certik (approved):
I think that looks good.

Do you think you could please benchmark this to make sure it is actually faster? It should, but let's test it.

opixdown:
@certik  after this PR result

Device: Macbook air M2 8GB ram
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran --version
LFortran version: 0.60.0-175-gb27e97e62
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin25.2.0
```
x**8 Performance Comparison 

| Compiler              | BEFORE (with loop) | AFTER (optimized) | Improvement |
|-----------------------|-------------------|-------------------|-------------|
| LFortran --fast       | 0.016831 s        | 0.007689 s        | 2.2x faster  |
| LFortran (no opt)     | 0.011710 s        | 0.006525 s        | 1.8x faster |
| gfortran -O3          | 0.005429 s        | 0.006069 s        | ~same       |
| gfortran -O3 -ffast   | 0.005369 s        | 0.006076 s        | ~same       |



i will share whole benchmark shortly.

certik:
Let's check x^7 via:
```fortran
program benchmark_
   implicit none
   integer, parameter :: ik = selected_int_kind(18)
   integer, parameter :: imax = 50
   integer, parameter :: findmax = 1
   real(selected_real_kind(14)) :: time0, time1
   integer :: nfound
   integer(ik) :: i, j, k, l, m
   character(*), parameter :: tfmt='(a,i2,a,f0.4,a)'
   
  ! Test x**7
   call cpu_time(time0)
   nfound = 0
   outer7: do i = 1, imax
      do j = 1, i
         do k = 1, i
            do l = 1, i
               do m = 1, i
                  if (j**7 + k**7 + l**7 + m**7 == i**7) then
                     nfound = nfound + 1
                     if (nfound >= findmax) exit outer7
                  endif
               end do
            end do
         end do
      end do
   end do outer7
   call cpu_time(time1)
   print tfmt, 'x**', 7, ': ', (time1-time0), ' seconds'
end program benchmark_
```

On Apple M4 before this PR:

`lfortran`: 0.1628
`lfortran --fast`: 0.0233

After this PR:

`lfortran`: 0.1059
`lfortran --fast`: 0.0233

And gfortran:

`gfortran`: 0.1124
`gfortran -O3 -march=native -ffast-math`: 0.0000

certik:
So it looks like this doesn't affect `--fast`, presumably because LLVM itself does exactly the same optimization. What does GFortran before and after mean in your benchmark? Why would the numbers be different? Did you run it couple times and took the fastest number? One must be very careful when doing benchmarks.

opixdown:
> So it looks like this doesn't affect `--fast`, presumably because LLVM itself does exactly the same optimization. What does GFortran before and after mean in your benchmark? Why would the numbers be different? Did you run it couple times and took the fastest number? One must be very careful when doing benchmarks.

I just reran the code in the exact same sequence as before and after the PR. I agree that for strict comparison I technically shouldn’t need to rerun gfortran i will keep it in mind,I did it  because I felt it was clearer to present everything together in a single table for reference.
I didn’t pick the fastest results these are straight reruns under the same conditions.

certik:
You have to run several times and pick the fastest result. The numbers that you presented for gfortran are not reliable, as they differ too much.

PR #10106, opened 2026-02-16
Fix unary def op with array constructor

certik:
The `def_unary_operand` grammar rule did not include array constructors (`[...]`), causing a syntax error when a user-defined unary operator was applied to one, e.g. `.cat. [t(1), t(2)]`. Add the three array constructor alternatives to `def_unary_operand`.


certik:
The test doesn't compile with GFortran, need to investigate.

certik:
So the test was actually invalid Fortran and LFortran should have given an error (https://github.com/lfortran/lfortran/issues/10117). I fixed the code, now it works.

PR #10108, opened 2026-02-16
Fix associate with deferred-length string

certik:
Wrap deferred-length string expressions in Allocatable when used in non-lvalue associate constructs (e.g., `associate(x => get_str() // "bar")`). Previously, only arrays with empty dimensions were wrapped, leaving scalar deferred-length strings without the required Allocatable wrapper, which triggered an ASR verify error.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2025,7 +2025,7 @@ RUN(NAME associate_28 LABELS gfortran llvm)
 RUN(NAME associate_29 LABELS gfortran llvm)
 RUN(NAME associate_30 LABELS gfortran llvm)
 RUN(NAME associate_31 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
-RUN(NAME associate_32 LABELS gfortran llvm)
+RUN(NAME associate_32 LABELS llvm)
 
 RUN(NAME attr_dim_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME attr_dim_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)


certik:
The test fails with GFortran:
```
1551/2527 Test #1506: associate_32 .......................................Subprocess aborted***Exception:   0.10 sec
free(): double free detected in tcache 2

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x7fc558223e59 in ???
#1  0x7fc558222e75 in ???
#2  0x7fc557e4532f in ???
#3  0x7fc557e9eb2c in ???
#4  0x7fc557e4527d in ???
#5  0x7fc557e288fe in ???
#6  0x7fc557e297b5 in ???
#7  0x7fc557ea8ff4 in ???
#8  0x7fc557eab55e in ???
#9  0x7fc557eaddcd in ???
#10  0x55afb1cf3549 in associate_32
	at /home/runner/work/lfortran/lfortran/integration_tests/associate_32.f90:14
#11  0x55afb1cf3590 in main
	at /home/runner/work/lfortran/lfortran/integration_tests/associate_32.f90:13
```

Need to investigate if the test is valid Fortran or not. If it is, then we can disable gfortran test. Otherwise we need to fix it.

integration_tests/CMakeLists.txt
```diff
@@ -2025,6 +2025,7 @@ RUN(NAME associate_28 LABELS gfortran llvm)
 RUN(NAME associate_29 LABELS gfortran llvm)
 RUN(NAME associate_30 LABELS gfortran llvm)
 RUN(NAME associate_31 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME associate_32 LABELS gfortran llvm)
```
```suggestion
RUN(NAME associate_32 LABELS llvm)
```

certik:
I investigated and tested Flang, ifx, gfortran and consulted the Fortran standard. It seems this is valid Fortran, so I am only going to test it with lfortran.

HarshitaKalani (approved):

PR #10105, opened 2026-02-16
Fix false error for zero-length arrays

certik:
Skip the array element count check when the function parameter has unknown dimensions (e.g., assumed-rank). Previously, unknown dimensions defaulted to size 1, causing a spurious error when passing zero-length arrays like [real::] to assumed-rank parameters.

PR #10109, opened 2026-02-16
Simplify the logic for assumed-rank handling

certik:
Simplification of https://github.com/lfortran/lfortran/pull/10105. Instead of using `bool rhs_size_known`, we now inline the logic directly.

If merged, this should be squashed-merged.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -13910,10 +13910,18 @@ public:
         if (first_struct != nullptr && operator_sym == nullptr) {
             operator_sym = first_struct->m_symtab->resolve_symbol("~def_op~" + op);
             if (operator_sym == nullptr) {
-                diag.add(Diagnostic("`" + op
-                    + "` is not defined in the Struct: `" + first_struct->m_name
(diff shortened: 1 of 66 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -8701,7 +8700,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                 break;
                             }
                         }
-                        if( lhs_ele < rhs_ele ){
+                        if( arr_rhs->n_dims == 0 && rhs_ele < 0 && lhs_ele < rhs_ele ){
```
Pranavchiku:
```suggestion
                        if( !(arr_rhs->n_dims == 0 || rhs_ele < 0) && lhs_ele < rhs_ele ){
```

certik:
As noted in https://github.com/lfortran/lfortran/pull/10105#issuecomment-3908953986, it's not clear to me that the newer code is actually simpler to understand, but let's see if tests pass, then we can think about it.

Pranavchiku:
Its less readable, agreed. Its just simpler.

certik:
Ok, then let's not merge it.

PR #10112, opened 2026-02-16
docs: fix misleading real*8 statement in user guide (Fixes #10035)

### LINKED ISSUE #10035: Misleading item in User Guide
In ```https://docs.lfortran.org/en/usage``` the last paragraph says

"GNU extension declarations real*8 xvalue are accepted but deprecated. This is valid Fortran-77 but not Fortran-2018."

I have just looked in the Fortran 77 standard; real*8 is not there. I think the last sentence quoted above should be something like

"This was never standard-conforming Fortran."

amritamishra01:
Fixes #10035

This PR updates the user guide to clarify that `real*8` is not part of the Fortran standard and is an old IBM extension.


certik:
Thanks!

PR #10120, opened 2026-02-16
Fix class(*) allocatable assignment

### LINKED ISSUE #10119: Can't assign to an unlimited polymorphic variable
`$ cat assign.f90`

```
program main

   integer,  parameter :: i = 4
   class(*), allocatable :: obj

   obj = i
   
end program main
```
`$ lfortran assign.f90`
```
```

certik:
When assigning a non-polymorphic value (e.g. integer) to an allocatable class(*) target, the LLVM codegen did not dereference the extra pointer level from the Allocatable wrapper before calling create_gep2, causing a type mismatch assertion failure.

Add a load to strip the allocatable indirection before accessing the unlimited polymorphic struct fields.

Fixes #10119.


certik:
This bug is fixed by https://github.com/lfortran/lfortran/pull/10129 out of the box, and it doesn't work with this patch and that PR. So let's wait on this one after we merge #10129.

certik:
Fixed by https://github.com/lfortran/lfortran/pull/10130 in a cleaner way.

PR #10122, opened 2026-02-16
Fix class(*) assignment from variables

certik:
When assigning a non-parameter variable (e.g., integer, real)
to an allocatable `class(*)` target, the LLVM codegen passed the
alloca pointer directly to deepcopy instead of the loaded
value, causing a type mismatch:
  store i32* %i, i32* %32  (pointer vs value)

Add a load to obtain the scalar value from the variable
pointer before passing it to deepcopy.

Depends on #10120.


certik:
Fixed by https://github.com/lfortran/lfortran/pull/10130 in a cleaner way.

PR #10123, opened 2026-02-16
Fix real/aimag on class(*) in select type

certik:
Just the last commit (93c97c9a9fecb8235aa3ae8fe2c032fc9709dfa3). It works, but it's not the right fix.

The fundamental problem is that we need to use casts in the select type blocks and cast `class(*)` to a particular concrete type. Then the concrete type will work in all intrinsic functions or other operations out of the box. Currently we do not have explicit casts, and as such we need to fix every place that uses the `class(*)` type and handle the casting implicitly there, both in the backend and in intrinsic functions like in this PR, which is is complex and never complete. This work we would now have to extend to all intrinsics. So this is not the way. Another example where we also just fixed it only for print is here: https://github.com/lfortran/lfortran/pull/8733, as a temporary hotfix to unblock us is fine, but it's not the right fix.

The solution is to have explicit casts for the select type variable. Two cases can happen:

* Read only usage, such as in arithmetic operations, most intrinsic functions, `intent(in)` arguments, etc. There is no issue here, we implement the cast in the LLVM backend and it will work in every case.
* Write and read/write, such as `intent(inout/out)`, LHS of assignments, etc. In this case we have to handle it just like we do descriptor to pointer array casts: the LLVM backend has to create a temporary and then copy it back, in the general case. In our `class(*)` to concrete type cast, it should be possible to just unpack it and modify inplace, so no temporary copy is needed, but if there is some case when it is needed, we can still do a temporary copy: the LLVM backend should decide and do it.

----

Inside a select type block with type is (complex(4)), calling real(obj)
or aimag(obj) on a `class(*)` variable failed with ASR verify error
because the variable type is restored to class(*) after semantic
analysis, but verify_args only accepts int/real/complex.

Fix by:
- Generator: add allow_polymorphic_arg and
  encode_arg_type_as_overload_id for Real; add allow_polymorphic_arg for
  Aimag. This makes verify_args accept class(*) and encodes the original
  arg type (0=int, 1=real, 2=complex) in overload_id during create_Real.
- instantiate_Real: for `class(*)` args, emit a direct Cast
  (ComplexToReal/RealToReal/IntegerToReal) based on overload_id, since
  visit_Cast already handles extracting values from class(*) containers.
- instantiate_Aimag: for `class(*)` args, emit a direct ComplexIm
  expression.
- visit_ComplexIm: add load_unlimited_polymorpic_value and select type
  context lookup for `class(*)` args.


certik:
As an example, here is the same fix using the new design:

* https://github.com/lfortran/lfortran/pull/10126/changes/9ea79c9f573af75bcb1e8de00fe13e492b439b88

vs the old design:

* https://github.com/lfortran/lfortran/commit/93c97c9a9fecb8235aa3ae8fe2c032fc9709dfa3

PR #10126, opened 2026-02-16
Initial implementation of ClassToIntrinsic cast

certik:


certik:
Cleaned up version in https://github.com/lfortran/lfortran/pull/10129.

PR #10128, opened 2026-02-16
fix: enforce intent call-arg invariants in ASR verify

### LINKED ISSUE #8005: Note about intent(out) arguments, enforce in verify
Options for intent:

* `intent(in)` arguments can be any expression, we enforce at compile time inside the function that we never write into the parameter
* `intent(inout)` and `intent(out)` arguments must be one of:
  * a `Var` expr node, which the function modifies directly
  * `ArraySection` of a `Var` array, possibly non-contiguous; in the backend we create a temporary array, the function modifies the contiguous array and we copy it back to the `Var`
  * `Cast` or `ArrayPhysicalCast` or `StringPhysicalCast` of a `Var`. In the backend we have to ensure that if the function modifies the parameter, that it propagates correctly to the `Var`. Typically the cast reuses a pointer internally, so it works out, but it requires special care, the backend cannot (for example) make a copy in the cast operation.
* `unspecified`: if the function for a given runtime value of arguments is only reading the parameter, then it is like `intent(in)` and it can be any expression; if the function is writing to the parameter, then it is like `intent(inout)` or `intent(out)` and the same restrictions apply as above. The function can do both just reading or also writing and decide at runtime based on the value of other arguments, or even global variables or file input. So at compile time we do not know. So one way to solve this is to examine the expression passed at a call site: 
  * if the cases for `intent(inout/out)` from above are present at the call site, then we handle them the same way (including a temporary as needed), assuming that the argument could be `intent(inout)`. That way if the function happens to write into the parameter, then it will work. If it happens to not write into it, then it could be wasteful to create a temporary, but we need to allow for the possibility, because we fundamentally cannot tell at compile time, the function can decide based on a reading a file at runtime if it will write back into the argument. The user can always do `intent(in)` to avoid copying the temporary back.
  * if other expression is present, then by definition it cannot be `intent(inout)` or `intent(out)`, so the function *must* be treating it as `intent(in)` (i.e., not writing into it), so we do not need to do special handling in the backend. If the function happens to write into the parameter anyway, then the result is undefined (segfault). In principle it could be checked in Debug mode at runtime by passing a flag for the argument if it contains a general expression and thus cannot be written to, and for every write into the argument inside the function we insert a check for this flag, and give a runtime error if the flag is "ready only".

Note: I think `unspecified` is equivalent to deciding at each call site if the expression is the second case above, then it's `intent(in)`, otherwise for the first case it's `intent(inout)`. We know for each call site at compile time which of the two cases it is, but inside the function we do not know at compile time which one it is. However, we could in principle generate two versions of the function, one `intent(in)`, one `intent(inout)` and call each version appropriately for each call site. The `intent(in)` version would simply replace all argument writes with a runtime error. But it would have to recursively propagate to all functions that we call, and for each argument, so it seems like an explosion of cases, so likely not practical.

Let's enforce this in verify() and document.

krystophny:
## Summary
- Enforce ASR `verify()` call-site intent contract: `intent(out/inout)` actual arguments must be writable.
- Keep cast-wrapper support (`Cast`, `ArrayPhysicalCast`, `StringPhysicalCast`) while tightening helper scope to issue-defined cases.
- Fix CI regression in `modules_30`: avoid unsafe `ArrayItem` down-cast in `is_modifiable_actual_argument_expr()`.

Fixes #8005

## Why
Issue #8005 asks to enforce this in `verify()` and document the contract. Invalid ASR could previously pass verifier checks.

The new helper originally called `is_variable()` for `ArrayItem` expressions, which can hit an invalid down-cast for non-`Var` array bases and abort ASR verification (`modules_30` in CI). This PR now handles `ArrayItem` safely.

**Stage:** ASR Verify

## Changes
- [`src/libasr/asr_utils.h#L1519`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/libasr/asr_utils.h#L1519): `is_modifiable_actual_argument_expr()` now classifies modifiable forms directly and recurses safely on `ArrayItem` base expressions.
- [`src/libasr/asr_utils.h#L6889`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/libasr/asr_utils.h#L6889): semantic intent mismatch check uses the shared helper.
- [`src/libasr/asr_verify.cpp#L1003`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/libasr/asr_verify.cpp#L1003): verify call-arg intent invariants for subroutine/function calls.
- [`src/libasr/asr_verify.cpp#L1015`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/libasr/asr_verify.cpp#L1015): require `m_dt` for non-`nopass` method calls.
- [`src/lfortran/tests/test_ast.cpp#L108`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/lfortran/tests/test_ast.cpp#L108): base regression (`intent(out)` + non-variable).
- [`src/lfortran/tests/test_ast.cpp#L153`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/lfortran/tests/test_ast.cpp#L153): `ArrayPhysicalCast` path coverage.
- [`src/lfortran/tests/test_ast.cpp#L202`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/lfortran/tests/test_ast.cpp#L202): `StringPhysicalCast` path coverage.
- [`src/lfortran/tests/test_ast.cpp#L262`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/src/lfortran/tests/test_ast.cpp#L262): pass-method `m_dt` regression.
- [`doc/src/asr/asr.md#L81`](https://github.com/lfortran/lfortran/blob/f8df1c879833bb0de92b07653a4ef2a69a9357c7/doc/src/asr/asr.md#L81): call argument intent contract docs.

### Fails on main
```bash
$ git -C lfortran checkout -b tmp/issue8005-main-check upstream/main
$ git -C lfortran cherry-pick 7e32169ea
$ scripts/lf.sh build
$ ./lfortran/build/src/lfortran/tests/test_lfortran --test-case="ASR Verify INTENT(OUT) actual argument"
```
Observed:
```text
ERROR: CHECK( !asr_verify(*asr, true, diagnostics) ) is NOT correct!
[doctest] Status: FAILURE!
```

### Fails before CI regression fix
```bash
$ git -C lfortran checkout f00b06c74daa650d3537e706b82a4f10c50f05d6
$ scripts/lf.sh build
$ scripts/lf.sh test -t modules_30 -s
```
Observed:
```text
../integration_tests/modules_30.f90 * asr
LCOMPILERS_ASSERT failed: $DIR/src/libasr/asr.h
function down_cast(), line number 41 at
is_a<T>(*f)
```

### Passes after fix
```bash
$ git -C lfortran checkout fix/issue-8005-intent-verify
$ scripts/lf.sh build
$ scripts/lf.sh test -t modules_30 -s
```
Observed:
```text
../integration_tests/modules_30.f90 * asr ✓
TESTS PASSED
```

### Additional checks
```bash
$ scripts/lf.sh --llvm=11 pr-check
```
Observed local failure is unrelated to this change (`subroutine3.f90` LLVM reference drift: `i32*` vs `ptr`).

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3512,6 +3512,9 @@ RUN(NAME infer_walrus_shadow_01 LABELS llvm_infer NO_STD_F23)
 RUN(NAME infer_walrus_shadow_02 LABELS llvm_infer NO_STD_F23)
 RUN(NAME infer_walrus_shadow_03 LABELS llvm_infer NO_STD_F23)
 
+RUN(NAME verify_intent_external_01 LABELS gfortran llvm EXTRAFILES
+        verify_intent_external_01_mod1.f90 verify_intent_external_01_mod2.f90)
+
 # UTF-8 BOM handling
 RUN(NAME utf8_bom_01 LABELS gfortran llvm)
 
--- a/integration_tests/verify_intent_external_01.f90
+++ b/integration_tests/verify_intent_external_01.f90
@@ -1,0 +1,13 @@
+program verify_intent_external_01
+    use data_mod, only: global_count, global_values
+    use worker_mod, only: do_work
+    implicit none
+
+    call do_work()
+
+    if (global_count /= 5) error stop
+    if (abs(global_values(1) - 10.0) > 1.0e-6) error stop
+    if (abs(global_values(5) - 50.0) > 1.0e-6) error stop
+
+    print *, "ok"
+end program verify_intent_external_01
--- a/integration_tests/verify_intent_external_01_mod1.f90
+++ b/integration_tests/verify_intent_external_01_mod1.f90
@@ -1,0 +1,5 @@
+module data_mod
+    implicit none
+    integer :: global_count
+    real :: global_values(10)
+end module data_mod
--- a/integration_tests/verify_intent_external_01_mod2.f90
+++ b/integration_tests/verify_intent_external_01_mod2.f90
@@ -1,0 +1,25 @@
+module worker_mod
+    use data_mod, only: global_count, global_values
+    implicit none
+contains
+
+    subroutine set_count(n)
+        integer, intent(out) :: n
+        n = 5
+    end subroutine set_count
+
+    subroutine fill_values(arr, n)
+        integer, intent(in) :: n
+        real, intent(out) :: arr(n)
+        integer :: i
+        do i = 1, n
+            arr(i) = real(i * 10)
+        end do
+    end subroutine fill_values
+
+    subroutine do_work()
+        call set_count(global_count)
+        call fill_values(global_values, global_count)
+    end subroutine do_work
+
+end module worker_mod
--- a/src/libasr/asr_verify.cpp
+++ b/src/libasr/asr_verify.cpp
@@ -1049,7 +1049,7 @@ public:
                     continue;
                 }
 
-                if (enforce_modifiable_check &&
+                if (check_external && enforce_modifiable_check &&
                     !ASR::is_a<ASR::FunctionType_t>(*callee_param->m_type) &&
                     (callee_param->m_intent == ASR::intentType::Out ||
                      callee_param->m_intent == ASR::intentType::InOut)) {


certik:
Two tests failed:
```
lfortran  -c outer.f90
LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/../libasr/asr.h
function down_cast(), line number 40 at 
f != nullptr
make: *** [Makefile:180: outer.o] Aborted (core dumped)
```

certik:
This is good, we need to make verify more strict. Just need to investigate why tests failed, create an MRE, add it to integration tests and fix.

krystophny:
@certik I updated it. Unfortunately Claude force-pushed right away despite the policy.

certik:
Yes, this is the right fix: https://github.com/lfortran/lfortran/pull/10128/changes/54a3023bd22ab0a7b38e03e7c2b1085c2819827e

For some reason github is not showing my comment, but I think you saw it.

certik:
It will take me some time to carefully review this.

jinangshah21 (approved):
I think this is fine!

src/libasr/asr_utils.h
```diff
@@ -1566,6 +1566,17 @@ static inline bool is_modifiable_actual_argument_expr(ASR::expr_t* a_value) {
...
+static inline bool enforce_modifiable_actual_argument_check(const ASR::Function_t* f) {
+    if (f == nullptr) {
+        return false;
+    }
+    std::string fn_name = f->m_name;
+    if (fn_name.rfind("_lfortran_", 0) == 0 || fn_name.rfind("_lcompilers_", 0) == 0) {
+        return false;
+    }
+    return true;
+}
```
certik, line 1594: I don't like this. We should not be calling `_lfortran` in libasr at all, it must be frontend independent.

Also we should not be doing string comparison at all. Must encode all the info in ASR itself as enums or some other fast way.

The most concerning however: why do we need to add special casing at all for this?

certik:
I rebased it on top of the latest main, let's make sure this still passes.

certik:
@krystophny do you mind please fixing the above comment?

PR #10129, opened 2026-02-17
Implement ClassToIntrinsic casting

certik:
TODO for the future:

13. Extend `ClassToIntrinsic` to **arrays**, eliminating the fallback to
    the old mechanism (rewriting `assoc_variable->m_type`).
    **Investigation result:** This is hard because `create_ArrayRef` in
    `ast_common_visitor.h` creates bare `Var_t` nodes directly (bypassing
    `resolve_variable()` and `select_type_casts_map`), so array element
    accesses like `x(i)` wouldn't get ClassToIntrinsic wrapping. Also,
    `symbol_type(v)` is checked for array nature in `visit_FuncCallOrArray`,
    and keeping the type as class(*) would break this check. The current
    approach (rewriting `assoc_variable->m_type` for arrays) works correctly
    and the old mechanisms (`select_type_context_map`, `load_unlimited_
    polymorphic_value`) are already removed/disabled for TypeStmtType under
    `new_classes`. Deferring to a future cleanup pass.
14. Consider converting `TypeStmtName` to also use `select_type_casts_map`
    for all value extraction (it already does for `!assoc_sym`).
    **Deferred:** Would require not rewriting `assoc_variable->m_type`
    for TypeStmtName, handling ClassToStruct on LHS (store-through for
    struct fields), and modifying the Associate statement. Complex and
    low-priority since the current approach works correctly.
15. Consider removing `select_type_context_map` entirely (for `new_classes`).
    **Deferred:** Depends on step 14 — TypeStmtName and ClassStmt still
    populate and use the context map. Can only be removed when those are
    also migrated to use `select_type_casts_map` for all value extraction.


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8482,6 +8473,52 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            if (ASRUtils::is_character(*target_type)) {
+                // String assignment needs lfortran_str_copy
+                llvm_utils->deepcopy(x.m_value, rhs_value, data_ptr,
+                    target_type, target_type, module.get());
+            } else {
+                builder->CreateStore(rhs_value, data_ptr);
+            }
+            return;
+        }
+
```
assem2002, line 8521: if `ptr_load = 0` while visiting LHS, We shouldn't be in need of this piece of code, right?

jinangshah21: Actually, we can remove above special handling, as if LHS is casted to correct type, then `visit_Assignment` should handle it. For eg. 
```fortran
class(*), allocatable :: x
integer :: y
......

select type(x)
type is(integer)
     x = y      !!! (Here if we have cast on x to integer then from `visit_cast` we will get `i32*` and then it will store RHS in it)
```

certik:
Ok, let's merge as is, and then improve upon it with subsequent PRs.

src/libasr/codegen/asr_to_llvm.cpp
certik: Made this https://github.com/lfortran/lfortran/issues/10136

PR #10130, opened 2026-02-17
ClassToIntrinsic cast branch PR

certik:
Built upon https://github.com/lfortran/lfortran/pull/10129. Merge that one first, then we'll rebase / clean this one.

PR #10133, opened 2026-02-17
Implement and test most select type features

certik:
Depends on https://github.com/lfortran/lfortran/pull/10129 and https://github.com/lfortran/lfortran/pull/10130.

This uses the cast approach to get most select type features working. There is no blocker, everything works nicely. Each commit is standalone with tests, all tests pass locally, and can be submitted as a separate PR.


certik:
Tests pass here, so I'll leave this untouched for reference. I rebased this PR on top of the latest main in https://github.com/lfortran/lfortran/pull/10135.

PR #10134, opened 2026-02-17
Fix class(*) intrinsic scalar store

### LINKED ISSUE #10119: Can't assign to an unlimited polymorphic variable

### LINKED ISSUE #10121: Can't assign unlimited polymorphic expression to associate-name of a `select type` statement
This is most likely caused by the same bug as #10119. But I am reporting it here nevertheless, just in case something additional is involved.

`$ cat assign2.f90 `

```
program main

   integer,  parameter :: i = 4
   
   select type (obj => unlimited_polymorphic(i))
   type is (integer)
      print *, obj
   end select

contains

   function unlimited_polymorphic(i) result(res)
      integer,  intent(in) :: i
      class(*), allocatable :: res
      res = i
   end function unlimited_polymorphic
   
end program main
```
`$ lfortran assign2.f90`
 ```
```

certik:
Load the value from pointer only when the LLVM value is actually a pointer. With ptr_loads=0, variables produce a pointer while constants (e.g. complex literals) produce a value directly. The previous code unconditionally loaded, which caused LLVM verification failures for complex constants.

Fixes #10121. Fixes #10119.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2404,6 +2404,9 @@ RUN(NAME class_108 LABELS gfortran llvm)
 RUN(NAME class_109 LABELS gfortran llvm)
 RUN(NAME class_110 LABELS gfortran llvm EXTRAFILES class_110_module.f90)
 RUN(NAME class_111 LABELS gfortran llvm)
+RUN(NAME class_112 LABELS gfortran llvm)
+RUN(NAME class_113 LABELS gfortran llvm)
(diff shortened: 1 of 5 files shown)


certik:
Let's merge this as part of https://github.com/lfortran/lfortran/pull/10135, to keep the git history clean.

PR #10145, opened 2026-02-17
Fix bugs for the Julienne code

certik:
With this PR it gets to 93%. Each commit here is independent and should pass all tests. I recommend reviewing it commit by commit.

### DIFF
--- a/src/libasr/pass/pass_array_by_data.cpp
+++ b/src/libasr/pass/pass_array_by_data.cpp
@@ -245,6 +245,77 @@ class PassArrayByDataProcedureVisitor : public PassUtils::PassVisitor<PassArrayB
         }
 
 
+        /*
+         * Check whether `func_name` is bound to a Struct via
+         * StructMethodDeclaration in `scope` or, when `scope` belongs
+         * to a submodule, in the parent module's scope.
+         *
+         * When a type-bound procedure call goes through a
+         * StructMethodDeclaration the call site is intentionally NOT
+         * rewritten by this pass (see visit_Call).  Therefore the
+         * *definition* must also keep its original descriptor-based
+         * calling convention; otherwise the linker will see a mangled
+         * name on the definition side but an un-mangled reference on
+         * the call side, producing an "undefined reference" error.
+         */
+        bool is_struct_bound_procedure(const std::string& func_name,
+                                       SymbolTable* scope) {
+            auto has_struct_method_decl = [&](SymbolTable* s) -> bool {
+                for (auto& item : s->get_scope()) {
+                    if (ASR::is_a<ASR::Struct_t>(*item.second)) {
+                        ASR::Struct_t* struct_t =
+                            ASR::down_cast<ASR::Struct_t>(item.second);
+                        for (auto& member : struct_t->m_symtab->get_scope()) {
+                            if (ASR::is_a<ASR::StructMethodDeclaration_t>(
+                                    *member.second)) {
+                                ASR::StructMethodDeclaration_t* smd =
+                                    ASR::down_cast<
+                                        ASR::StructMethodDeclaration_t>(
+                                            member.second);
+                                ASR::symbol_t* proc =
+                                    ASRUtils::symbol_get_past_external(
+                                        smd->m_proc);
+                                if (std::string(ASRUtils::symbol_name(proc))
+                                        == func_name) {
+                                    return true;
+                                }
+                            }
+                        }
+                    }
+                }
+                return false;
+            };
+
+            // Check the current (module) scope
+            if (has_struct_method_decl(scope)) return true;
+
+            // If in a submodule, also check the parent module
+            if (scope->asr_owner
+                    && ASR::is_a<ASR::symbol_t>(*scope->asr_owner)) {
+                ASR::symbol_t* owner_sym =
+                    ASR::down_cast<ASR::symbol_t>(scope->asr_owner);
+                if (ASR::is_a<ASR::Module_t>(*owner_sym)) {
+                    ASR::Module_t* mod =
+                        ASR::down_cast<ASR::Module_t>(owner_sym);
+                    if (mod->m_parent_module && scope->parent) {
+                        ASR::symbol_t* parent_sym =
+                            scope->parent->get_symbol(
+                                mod->m_parent_module);
+                        if (parent_sym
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -19024,6 +19025,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                // the result is passed via the last output argument.
+                llvm::Type* result_type = llvm_utils->get_type_from_ttype_t_util(
+                    ASRUtils::EXPR(ASR::make_Var_t(al, x.base.base.loc,
+                        ASR::down_cast<ASR::Var_t>(s->m_args[s->n_args - 1])->m_v)),
+                    x.m_type, module.get());
+                tmp = llvm_utils->CreateAlloca(result_type, nullptr, "struct_ret_tmp");
+                builder->CreateStore(llvm::Constant::getNullValue(result_type), tmp);
+                args.push_back(tmp);
+                builder->CreateCall(fn, args);
+            } else {
```
assem2002, line 19039: I think this should be handled by the pass `subroutnie_from_function`, As I think it's doing part of the passes job in the backend.

assem2002: If that's the case, then this commit a28311baf071af49ae3bb9ff0fb079eeaeadb894 needs a rework

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -7937,6 +7937,36 @@ LFORTRAN_API void _lfortran_string_read_bool(char *str, int64_t len, char *forma
...
+            buf[i] = 'e';
+        }
+    }
+}
+
+LFORTRAN_API void _lfortran_string_read_c32(char *str, int64_t len, char *format, struct _lfortran_complex_32 *c) {
+    char *buf = (char*)malloc(len + 1);
+    if (!buf) return;
+    memcpy(buf, str, len);
+    buf[len] = '\0';
```
assem2002, line 7954: Minor
```suggestion
    char *buf = to_c_string(str, len)
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -7937,6 +7937,36 @@ LFORTRAN_API void _lfortran_string_read_bool(char *str, int64_t len, char *forma
...
+    _lfortran_replace_d_exponent(buf);
+    sscanf(buf, format, &c->re, &c->im);
+    free(buf);
+}
+
+LFORTRAN_API void _lfortran_string_read_c64(char *str, int64_t len, char *format, struct _lfortran_complex_64 *c) {
+    char *buf = (char*)malloc(len + 1);
+    if (!buf) return;
+    memcpy(buf, str, len);
+    buf[len] = '\0';
```
assem2002, line 7964: Also here
Minor 
```suggestion
    char *buf = to_c_string(str, len)
```

src/libasr/pass/pass_array_by_data.cpp
```diff
@@ -245,6 +245,77 @@ class PassArrayByDataProcedureVisitor : public PassUtils::PassVisitor<PassArrayB
...
+         * to a submodule, in the parent module's scope.
+         *
+         * When a type-bound procedure call goes through a
+         * StructMethodDeclaration the call site is intentionally NOT
+         * rewritten by this pass (see visit_Call).  Therefore the
+         * *definition* must also keep its original descriptor-based
+         * calling convention; otherwise the linker will see a mangled
+         * name on the definition side but an un-mangled reference on
+         * the call side, producing an "undefined reference" error.
+         */
```
assem2002, line 260: Bypassing the work of the pass with this case is fine.
but I think the `StructMethodDeclaration` is just pointing to the old function symbol (which should raise an error by ASR verify btw), it just needs an update.

assem2002 (approved):
Good to merge. Thanks!
Just address my comments on the `subroutine_from_function` related fix pushed at https://github.com/lfortran/lfortran/commit/a28311baf071af49ae3bb9ff0fb079eeaeadb894 .
I think it pushing some passes' work into the backend, which we always try to avoid.

Let me know, if you need help with those suggested fixes.

src/libasr/pass/pass_array_by_data.cpp
assem2002: I think it worth the job, if it's all about update, It should be couple of line of change.

src/libasr/codegen/asr_to_llvm.cpp
assem2002: I think it'd be couple of line of changes.

certik: Yes, let's rework it. We have time, this is not urgent, so let's rework it before merging. I'll try to yank this commit out of the PR and merge the rest.

src/libasr/pass/pass_array_by_data.cpp
certik: So what should be the right fix here?

certik:
I yanked the two problematic commits and fixed the runtime and rebased into https://github.com/lfortran/lfortran/pull/10149. Then we need to fix the two commits above as separate PRs.

src/libasr/pass/pass_array_by_data.cpp
assem2002: We shouldn't be bypassing function used with `StructMethodDeclaration` . I'll try to use AI to see if we can do that, otherwise it's still fine to bypass it -- Things are contained in the pass and easily fixable.

src/libasr/codegen/asr_to_llvm.cpp
certik: Refactored in https://github.com/lfortran/lfortran/pull/10181.

src/libasr/pass/pass_array_by_data.cpp
certik: Here is the original code: https://github.com/lfortran/lfortran/pull/10184 and here is the alternative fix: https://github.com/lfortran/lfortran/pull/10185, which seems more complicated, let me know.

certik:
All of this was resubmitted as separate PRs.

PR #10177, opened 2026-02-18
fix: resolve BLOCK-local USE imports in semantics

### LINKED ISSUE #9061: Bug: BLOCK with USE fails to resolve derived type from used module
### Summary
LFortran reports a semantic error for a derived type that is correctly imported via a `use` statement inside a `block` construct.

### Minimal reproducer
`mre.f90`:

```fortran
module legend_mod
    implicit none
    type :: legend_entry_t
        integer :: x
    end type legend_entry_t
end module legend_mod

program mre
    implicit none

    block
        use legend_mod, only: legend_entry_t
        type(legend_entry_t), allocatable :: new_entries(:)
        allocate(new_entries(0))
    end block

    print *, "ok"
end program mre
```

### Expected
Compiles.

### Actual
LFortran fails with:

```
semantic error: Derived type `legend_entry_t` is not defined
```

### Evidence
- LFortran log: `/tmp/lfortran-dev/fortplot/mre_block_use_legend_entry_lfortran_2025-12-13_204534.txt`
- Reproducer source: `/tmp/lfortran-dev/fortplot/mre_block_use_legend_entry_2025-12-13_204534.f90`

### Control (gfortran)
GFortran compiles and runs:
- Build log: `/tmp/lfortran-dev/fortplot/mre_block_use_legend_entry_gfortran_2025-12-13_204534.txt`
- Run log: `/tmp/lfortran-dev/fortplot/mre_block_use_legend_entry_gfortran_run_2025-12-13_204534.txt`

### Versions
- LFortran: `db51e58c799e` (upstream/main)

### Context
This blocks building fortplot, where `src/figures/management/fortplot_figure_initialization.f90` uses:

```fortran
block
    use fortplot_legend, only: legend_entry_t
    type(legend_entry_t), allocatable :: new_entries(:)
    ...
end block
```

Fortplot tracker/workaround issue: lazy-fortran/fortplot#1425 (filed separately).

krystophny:
## Summary
- Implement `USE` import handling in `BodyVisitor` so `USE` inside `BLOCK` is semantically processed before block declarations.
- Add integration regression test `block_11` that reproduces the fortplot pattern (`BLOCK` + `USE, ONLY` + derived type declaration).
- Register `block_11` in integration test CMake.

Fixes #9061

## Why
`USE` statements inside `BLOCK` were ignored in the body semantic pass, so derived types imported only within the block were unresolved.

**Stage:** Semantics

## Changes
- [`src/lfortran/semantics/ast_body_visitor.cpp#L45-L195`](https://github.com/krystophny/lfortran/blob/f3a5368bb11372163f2a126316a4c5b767b7cd66/src/lfortran/semantics/ast_body_visitor.cpp#L45-L195): add helpers for module symbol import in block scope (`USE` symbol filtering, access checks, import-all/only handling).
- [`src/lfortran/semantics/ast_body_visitor.cpp#L4002-L4060`](https://github.com/krystophny/lfortran/blob/f3a5368bb11372163f2a126316a4c5b767b7cd66/src/lfortran/semantics/ast_body_visitor.cpp#L4002-L4060): replace body-pass `visit_Use` no-op with actual import logic.
- [`integration_tests/block_11.f90#L1`](https://github.com/krystophny/lfortran/blob/f3a5368bb11372163f2a126316a4c5b767b7cd66/integration_tests/block_11.f90#L1): new regression case for block-local derived-type import via `USE`.
- [`integration_tests/CMakeLists.txt#L2018`](https://github.com/krystophny/lfortran/blob/f3a5368bb11372163f2a126316a4c5b767b7cd66/integration_tests/CMakeLists.txt#L2018): add `RUN(NAME block_11 ...)`.

## Tests
- [`integration_tests/block_11.f90`](https://github.com/krystophny/lfortran/blob/f3a5368bb11372163f2a126316a4c5b767b7cd66/integration_tests/block_11.f90)

### Test fails on main
```bash
$ git -C lfortran checkout upstream/main
$ scripts/lf.sh build
$ scripts/lf.sh run /tmp/mre_9061.f90
Compiling: /tmp/mre_9061.f90
semantic error: Derived type `legend_entry_t` is not defined
  --> /tmp/mre_9061.f90:13:9
```

### Test passes after fix
```bash
$ git -C lfortran checkout fix/fortplot-block-use-import
$ scripts/lf.sh build
$ scripts/lf.sh run /tmp/mre_9061.f90
Compiling: /tmp/mre_9061.f90
Running: /tmp/lfortran_run_6822
ok

$ scripts/lf.sh itest -b llvm -t block_11
1/1 Test #1433: block_11 ... Passed
```

### Additional regression check
```bash
$ scripts/lf.sh itest -b llvm -t block_0
100% tests passed, 0 tests failed out of 9

$ scripts/lf.sh itest -b llvm -t block_10
```


certik:
This will take time to review. It seems to add lots of duplicate code. Ideally the use handling will be the same everywhere.

krystophny:
> This will take time to review. It seems to add lots of duplicate code. Ideally the use handling will be the same everywhere.

Actually I had a version with less duplication but this moved the whole `use` processing into the common visitor to cover also BLOCK. Would this be an option?

certik:
I think so --- if the use has to be processed in the body visitor, then we don't have a choice I think.

To make it easy to review, create one commit that just moves the code. And then make modifications to it on top. That way the modifications are small and easy to review and then the big move is in one commit that is just a move and no other changes, so also easy to review.

krystophny:
> I think so --- if the use has to be processed in the body visitor, then we don't have a choice I think.
> 
> To make it easy to review, create one commit that just moves the code. And then make modifications to it on top. That way the modifications are small and easy to review and then the big move is in one commit that is just a move and no other changes, so also easy to review.

Done

PR #10180, opened 2026-02-18
codegen: CFI_cdesc_t-compatible inline array descriptor

### LINKED ISSUE #7316: Change array descriptor to be inlined
https://github.com/gcc-mirror/gcc/blob/9a2083f025fae3559474b961943484d764c2e6dd/libgfortran/ISO_Fortran_binding.h

```
typedef struct CFI_dim_t
  {
    CFI_index_t lower_bound;
    CFI_index_t extent;
    CFI_index_t sm;
  }
CFI_dim_t;

/* CFI_cdesc_t, C descriptors are cast to this structure as follows:
   CFI_CDESC_T(CFI_MAX_RANK) foo;
   CFI_cdesc_t * bar = (CFI_cdesc_t *) &foo;
 */
typedef struct CFI_cdesc_t
 {
    void *base_addr;
    size_t elem_len;
    int version;
    CFI_rank_t rank;
    CFI_attribute_t attribute;
    CFI_type_t type;
    CFI_dim_t dim[];
 }
CFI_cdesc_t;
```

I think our LLVM descriptor should just be an equivalent of `CFI_cdesc_t`, i.e., the dimension "subdescriptor" is inlined into the main descriptor, by value. So no pointer dereference involved. It means we will have different descriptors per array rank.

krystophny:
## Summary
- Change LLVM array descriptor layout to match ISO C Fortran interop `CFI_cdesc_t` structure
- Inline dimension subdescriptors at fixed `CFI_MAX_RANK=15` (no pointer indirection)
- Dimension fields always `i64` (`CFI_index_t`), removing `--descriptor-index-64` flag dependency

Fixes #7316

## Why
The current descriptor uses a pointer-to-dimension-descriptor array and flag-dependent index types (i32/i64). This is incompatible with the Fortran 2018 ISO C binding `CFI_cdesc_t` layout and adds unnecessary pointer indirection for dimension access.

**Stage:** Codegen (LLVM backend only, C backend unchanged)

## Changes
- [`llvm_array_utils.h`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/llvm_array_utils.h): New `FIELD_*` constants for CFI layout, `set_rank()` method for i8 rank
- [`llvm_array_utils.cpp`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/llvm_array_utils.cpp): CFI descriptor type, always-i64 dim fields, inline dims access
- [`asr_to_llvm.cpp`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/asr_to_llvm.cpp): Fix raw GEP accesses for new field indices, fix c_f_pointer shape loading for i64 index_type
- [`llvm_utils.cpp`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/llvm_utils.cpp): Fix offset field accesses (old index 1 -> FIELD_OFFSET)
- [`llvm_utils.h`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/llvm_utils.h): Remove heap-allocated dim descriptor free in struct finalizer
- [`c_utils.h`](https://github.com/lfortran/lfortran/blob/127303b1c/src/libasr/codegen/c_utils.h): Revert to upstream/main (C backend keeps its own layout)

## New Descriptor Layout
```
struct LFortran_cdesc_t {
    element_type* base_addr;    // 0: typed data pointer
    int64_t       elem_len;     // 1: element size in bytes
    int32_t       version;      // 2: descriptor version
    int8_t        rank;         // 3: CFI_rank_t
    int16_t       type;         // 4: CFI_type_t
    int8_t        attribute;    // 5: CFI_attribute_t
    int64_t       offset;       // 6: linear offset
    bool          is_allocated; // 7: allocation flag
    CFI_dim_t     dim[15];      // 8: {stride, lower_bound, extent} always i64
};
```

## Verification

This is a codegen refactoring, not a bug fix. No user-visible behavior changes.

### Descriptor layout: before and after
```
# Old layout (pointer-to-dim, i32 dims)
%array = type { i32*, i32, %dimension_descriptor*, i1, i32, i32, i32, i32, i32 }

# New CFI layout (inline dims, always i64)
%array = type { i32*, i64, i32, i8, i16, i8, i64, i1, [15 x %dimension_descriptor] }
%dimension_descriptor = type { i64, i64, i64 }
```

### Integration tests: zero regressions
```
$ cd integration_tests && FC=lfortran cmake ... && make -j && ctest -j --timeout 30
99% tests passed, 6 tests failed out of 2457
```
All 6 failures are pre-existing on upstream/main.
LLVM reference tests updated to reflect new descriptor layout.

## Follow-up Fix (CI regression on this PR)

After the descriptor-layout change, Ubuntu LLVM11 CI failed in `llvm_omp` with:
`CreateInBoundsGEP2: Type mismatch - pointer pointee type (i32) != type parameter (i64)`.

### Root cause
In `visit_CPtrToPointer`, lower-bound arrays were indexed using descriptor `index_type` (`i64`) instead of the actual lower-bound element type (often `i32` in OpenMP helper contexts), causing typed-pointer GEP mismatch on LLVM 11.

### Fix
- [`src/libasr/codegen/asr_to_llvm.cpp`](https://github.com/lfortran/lfortran/blob/7793e2b6e/src/libasr/codegen/asr_to_llvm.cpp): in `visit_CPtrToPointer`, use the real lower-bound element LLVM type for GEP/Load, then cast to descriptor index type.

### Regression verification
```bash
# Before fix: openmp_12 crashes with GEP type mismatch
$ git -C lfortran checkout 90168846e && scripts/lf.sh itest -b llvm_omp -t openmp_12
LCOMPILERS_ASSERT failed: CreateInBoundsGEP2: Type mismatch

# After fix: passes
$ git -C lfortran checkout 7793e2b6e && scripts/lf.sh itest -b llvm_omp -t openmp_12
```

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7007,8 +7007,12 @@ public:
                     ptr_loads = 2;
                     this->visit_expr_wrapper(x.m_lower_bounds, true);
                     ptr_loads = ptr_loads_copy;
-                    tmp = llvm_utils->create_ptr_gep2(idx_type, tmp,  i);
-                    new_lb = llvm_utils->CreateLoad2(idx_type, tmp);
+                    llvm::Type* lower_bound_type = llvm_utils->get_type_from_ttype_t_util(
+                        nullptr, ASRUtils::extract_type(ASRUtils::expr_type(x.m_lower_bounds)),
+                        module.get());
+                    tmp = llvm_utils->create_ptr_gep2(lower_bound_type, tmp, i);
+                    new_lb = builder->CreateSExtOrTrunc(
+                        llvm_utils->CreateLoad2(lower_bound_type, tmp), idx_type);
                 }
                 llvm::Value* new_size = nullptr;
                 if( ASRUtils::extract_physical_type(asr_shape_type) == ASR::array_physical_typeType::DescriptorArray ||


certik:
Awesome, thanks for fixing this. We need to carefully review.

HarshitaKalani (approved):
I think it looks great. Thanks!
In the description, we mention one test which fails with main and works with this PR. Can we add that in the integration_tests?

krystophny:
Thanks for the review! This is an internal codegen refactoring (descriptor layout), not a bug fix -- no user-visible behavior changes. The existing integration tests already validate correctness (zero regressions). A new integration test would only check LLVM IR layout details, which is fragile and already covered by the reference tests (updated `llvm-*` outputs in this PR).

certik:
Let me review this myself and play with it by hand. This is an important change, so we must get it right.

jinangshah21 (approved):
I didn't find any downside of this PR. Looks fine to me.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1013,13 +1012,7 @@ class ASRToLLVMVisitor;
             }
         }
 
-        /**
-         * @param arr llvm ptr to the array (descriptorArray, PointerArray, etc.)
-         * @param t array ASR type
-         * @param struct_sym if it's an array of struct. nullptr otherwise.
-         * @param in_struct is this array in some struct `(StructType(Array()))`.
-         */
```
certik, line 1021: Why was the documentation removed?

src/libasr/codegen/llvm_utils.h
```diff
@@ -945,15 +945,14 @@ class ASRToLLVMVisitor;
 
         }
 
-        /// Dispatches to the correct finalizer based on type.
```
certik, line 948: Keep docs?

src/libasr/codegen/llvm_utils.h
krystophny: Yes this must have been a mistake let's keep docs here and below.

certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/10676.

PR #10181, opened 2026-02-18
Fix FunctionCall with StructType

certik:


certik:
I sent a cleaner PR here: https://github.com/lfortran/lfortran/pull/10182.

Issue #4637, opened 2024-08-12
Two bugs: READ from terminal


harperjf:
This program is supposed to read an integer from the terminal and check it, twice.. If the first input is not an integer, lfortran neither gives an ```iomsg``` (bug 1) nor waits for the user to enter the next integer (bug 2). The ```i = 1,2``` in the ```do```  statement prevented an infinite loop at run time.
Gfortran does the right thing: if the first input is not an integer, the ```iomsg``` is correct, and the second input is correctly treated. The program below resembles the one in #4357 but that one did not mention ```iomsg```. See also #4438.
```
program readtesta
  implicit none ! lfortran problem with iostat in read statement
  integer  :: i,n,ios
  character:: msg*200
  character(:),allocatable:: iomsg
do i = 1,2
  print *,'Enter an integer'
  read(*,* ,iostat=ios,iomsg=msg) n
  print *,'n=',n
  iomsg = '"'//trim(msg)//'"'
  if(ios>0) print *,'Error reading integer: iostat=',ios,'iomsg=',iomsg
  if(ios<0) print *,'Insufficient input: iostat=',ios,'iomsg=',iomsg
end do
end program readtesta
```
Output from lfortran 0.38.0:
```
Enter an integer
BAD
n= 832483328
Enter an integer
n= 832483328
```
Output from gfortran:
```
 Enter an integer
BAD
 n=           0
 Error reading integer: iostat=        5010 iomsg="Bad integer for item 1 in list input"
 Enter an integer
666
 n=         666
```

harperjf:
If the reading is from a character string lfortran puts ```iostat=0``` even with invalid input (a different bug 1) and still omits the ```iomsg```. Program to reveal that:
```
program readtestb
  implicit none ! lfortran bug with iostat and iomsg in READ
  integer  :: i,n,ios
  character:: msg*200, input(2)*3 = ['BAD','666']
  character(:),allocatable:: iomsg
do i = 1,2
  read(input(i),*,iostat=ios,iomsg=msg) n
  print *,'n=',n
  iomsg = '"'//trim(msg)//'"'
  if(ios>0) print *,'Error reading integer: iostat=',ios,'iomsg=',iomsg
  if(ios<0) print *,'Insufficient input: iostat=',ios,'iomsg=',iomsg
end do
end program readtestb
```
Output from lfortran:
```
n= 32764
n= 666
```
Output from gfortran:
``` 
 n=           0
[...]

Issue #3936, opened 2024-04-25
exit or iostat=io interpreted wrongly


harperjf:
This program has a read statement with ```iostat=io``` to deal with bad data, but lfortran has a compile-time error. Or is the trouble with the ```exit``` statement? Evidence:
```
(lf2) john:~/Test/Glist$ cat testio.f90
program testio
  implicit none
  real x
  integer io
  character(3):: data = 'bad'
  do
     read(data,*,iostat=io) x
     if (io/=0) exit ! out of loop
     print *,'Loop should have exited!'
     exit
  end do
  print *,'Program ending.'
end program testio
(lf2) john:~/Test/Glist$ lfortran testio.f90 > testio.out
code generation error: asr_to_llvm: module failed verification. Error:
Call parameter type does not match function signature!
  %x = alloca float, align 4
 i32*  call void @_lfortran_string_read(i8* %4, i8* getelementptr inbounds ([3 x i8], [3 x i8]* @1, i32 0, i32 0), float* %x)
```
The ```testio.out``` file was
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

@testio.data = internal global i8* null
@0 = private unnamed_addr constant [4 x i8] c"bad\00", align 1
@1 = private unnamed_addr constant [3 x i8] c"%d\00", align 1
@2 = private unnamed_addr constant [2 x i8] c" \00", align 1
@3 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@4 = private unnamed_addr constant [25 x i8] c"Loop should have exited!\00", align 1
@5 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1
@6 = private unnamed_addr constant [2 x i8] c" \00", align 1
@7 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@8 = private unnamed_addr constant [16 x i8] c"Program ending.\00", align 1
@9 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_set_argv(i32 %0, i8** %1)
  %2 = call i8* (i32, ...) @_lfortran_malloc(i32 4)
  call void (i32, i8*, ...) @_lfortran_string_init(i32 4, i8* %2)
  store i8* %2, i8** @testio.data, align 8
[...]

PR #10183, opened 2026-02-18
Add iostat error handling to _lfortran_string_read_* runtime functions

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4637
fixes: https://github.com/lfortran/lfortran/issues/3936

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2698,7 +2698,7 @@ RUN(NAME read_35 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_36 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_37 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_38 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME read_39 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME read_39 LABELS gfortran llvm)
 RUN(NAME read_40 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME write_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)


integration_tests/read_39.f90
```diff
@@ -0,0 +1,17 @@
...
+
+  read(input(1), *, iostat=ios) n
+  if (ios <= 0) error stop "Test 1 failed: iostat should be positive for bad input"
+  read(input(2), *, iostat=ios) n
+  if (ios /= 0) error stop "Test 2 failed: iostat should be 0 for valid input"
+  if (n /= 666) error stop "Test 2 failed: n should be 666"
+  read(input(1), *, iostat=ios) r
+  if (ios <= 0) error stop "Test 3 failed: iostat should be positive for bad real input"
+  read(float_str, *, iostat=ios) r
+  if (ios /= 0) error stop "Test 4 failed: iostat should be 0 for valid real input"
```
certik: I think this doesn't test all the types that you implemented in the C runtime, correct? Let's test everything.

HarshitaKalani: I'll do that.

integration_tests/read_40.f90
```diff
@@ -0,0 +1,14 @@
...
+  implicit none
+  real :: x
+  integer :: io
+  character(3) :: data = 'bad'
+  do
+     read(data, *, iostat=io) x
+     if (io /= 0) exit
+     print *, 'Loop should have exited!'
+     exit
+  end do
```
certik, line 11: I don't like this infinite loop. What is the purpose of it? Let's not do infinite loops, so that it doesn't hang our tests if it fails. Although in this case I can see the exit right in front of `end do`, but what is the point?

HarshitaKalani: I took this test from: https://github.com/lfortran/lfortran/issues/3936
in the main, io is 0 which is incorrect. Since bad is not a valid real literal, io should be positive. Once it is positive, the infinte loop will exit. Even if it is not, it'll exit after printing 'Loop should have exited!'.
I can still remove this and add a better test if you want.

certik: Instead of using "print tests" to check logic, let's figure out how to check what is needed to be checked using regular integration tests, I don't think we need reference tests for this. Reference tests should be used to check print output itself, which we are not doing here.

PR #10184, opened 2026-02-18
Fix submodule nopass type-bound linker error

certik:
The pass_array_by_data pass was renaming submodule functions (e.g. check -> check_integer____0) while calls through StructMethodDeclaration were intentionally left untransformed. This caused an undefined reference at link time because the call site used the original name but the definition had the mangled name.

Skip the pass_array_by_data transformation for functions that are bound to a Struct via StructMethodDeclaration, either in the same module or in a parent module (for submodules). This ensures the definition keeps the same calling convention and symbol name that the untransformed call site expects.


certik:
The alternative way to fix this is in https://github.com/lfortran/lfortran/pull/10185, but it it's not clear that it is simpler.

Pranavchiku (approved):

certik:
@krystophny did you post this comment to the wrong PR?

krystophny:
> @krystophny did you post this comment to the wrong PR?

yes should be #10185

certik:
Fixed in https://github.com/lfortran/lfortran/pull/10216.

PR #10185, opened 2026-02-18
Fix submodule nopass type-bound linker error

certik:
An alternative fix for https://github.com/lfortran/lfortran/pull/10184.


certik:
Merged as part of https://github.com/lfortran/lfortran/pull/10216.

PR #10166, opened 2026-02-18
Infer mode: default realloc-lhs-arrays semantics

krystophny:
## Summary
- make infer mode enable `realloc_lhs_arrays` by default
- keep explicit `--disable-realloc-lhs-arrays` precedence over infer defaults
- add run tests for infer default-ON and explicit disable behavior

Fixes #10158
Part of #10162
Depends on #10189

## Restack Note
- Restacked on top of `#10189` (canonical infer base).
- Closed predecessors `#10163` and `#10165` (closed on 2026-02-18) are superseded by `#10189`.

## Merge Order
1. `#10189` (`feature/infer-mode`)
2. `#10190` (`feature/walrus-syntax`)
3. `#10166` (this PR)

## Changes
- `src/bin/lfortran_command_line_parser.cpp`
  - after infer-mode resolution, set `compiler_options.po.realloc_lhs_arrays = true`
  - reapply `--disable-realloc-lhs-arrays` override after infer defaulting
- `tests/infer_realloc_defaults_01.f90`
- `tests/errors/infer_realloc_disabled_01.f90`
- `tests/tests.toml`
- reference updates under `tests/reference/`

## Verification
```bash
$ scripts/lf.sh test -t infer_realloc_defaults_01 -s
infer_realloc_defaults_01.f90 * run ✓

$ scripts/lf.sh test -t infer_realloc_disabled_01 -s
errors/infer_realloc_disabled_01.f90 * run ✓
```

PR #10189, opened 2026-02-18
feat: add --infer mode with type inference and realloc-lhs defaults

krystophny:
## Summary

Adds `--infer` CLI flag that enables type inference on first assignment. When a bare variable name is assigned to without prior declaration, the compiler infers its type from the RHS expression. Also defaults `realloc-lhs-arrays` to on in infer mode (natural for exploratory use).

- New `--infer` flag in CLI, mutually exclusive with `--std`
- Interactive mode (`lfortran` with no input files) enables infer mode by default
- Parser skips `fix_program_without_program_line` wrapping in infer mode
- Supports intrinsic types: integer, real, complex, logical, character
- Realloc-lhs-arrays defaults to on (can be disabled with `--no-realloc-lhs-arrays`)

Absorbs #10166 (realloc defaults).

**Stage:** Semantics + CLI

## Changes

- [`src/bin/lfortran_command_line_parser.cpp`](https://github.com/lfortran/lfortran/blob/9c489e9f47f1351532b7740e78877d6ba58f26fe/src/bin/lfortran_command_line_parser.cpp): `--infer` flag, mutual exclusion with `--std`
- [`src/lfortran/semantics/ast_body_visitor.cpp`](https://github.com/lfortran/lfortran/blob/9c489e9f47f1351532b7740e78877d6ba58f26fe/src/lfortran/semantics/ast_body_visitor.cpp): `infer_and_declare()` + `try_infer_and_declare()` for type inference on assignment
- [`src/lfortran/parser/parser.cpp`](https://github.com/lfortran/lfortran/blob/9c489e9f47f1351532b7740e78877d6ba58f26fe/src/lfortran/parser/parser.cpp): skip program line wrapping in infer mode
- [`src/libasr/utils.h`](https://github.com/lfortran/lfortran/blob/9c489e9f47f1351532b7740e78877d6ba58f26fe/src/libasr/utils.h): `infer_mode` field in CompilerOptions

## Tests

- `tests/infer_first_assignment_01.f90`: basic scalar inference (integer, real, complex, logical, character)
- `tests/infer_mode_smoke_01.f90`: end-to-end run test
- `tests/errors/infer_first_assignment_conflict_01.f90`: error on conflicting redeclaration
- `tests/errors/infer_first_assignment_derived_01.f90`: error on unsupported derived types
- `tests/errors/infer_mode_conflict_01.f90`: `--infer` + `--std` mutual exclusion
- `tests/infer_realloc_defaults_01.f90`: realloc-lhs-arrays on by default
- `tests/errors/infer_realloc_disabled_01.f90`: explicit disable works

## Review status

certik has provided review comments on the initial version. The `c2d2ca85f` commit addresses the feedback by refactoring `infer_type_and_declare` into `infer_and_declare` (reusable core) + `try_infer_and_declare` (Assignment visitor wrapper).


src/bin/lfortran_command_line_parser.cpp
```diff
@@ -435,6 +440,13 @@ namespace LCompilers::CommandLineInterface {
             }
         }
 
+        // Interactive mode defaults to infer mode (`lfortran` with no input file),
+        // unless an explicit mode is selected.
```
certik: How do you select an explicit mode?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4472,6 +4472,70 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
     integer :: A(3, 5)
     A(i,j) = i*j
     */
```
certik: This comment applies to the function below --- you inserted try_infer_and_declare at a wrong place.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4472,6 +4472,70 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+    // Returns true if a variable was declared via type inference.
+    // Called from visit_Assignment when compiler_options.infer_mode is true
+    // and LHS is an undeclared bare name.
```
certik: This documentation is unclear.

From the code, it seems that try_infer_and_declare returns false if `--infer` is not enabled, or the variable already exists, or the LHS is not a variable. Otherwise it always returns true.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4472,6 +4472,70 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            return false;
+        }
+
+        AST::Name_t* target_name = AST::down_cast<AST::Name_t>(x.m_target);
+        std::string var_name = to_lower(target_name->m_id);
+
+        ASR::symbol_t* existing = current_scope->resolve_symbol(var_name);
+        if (existing != nullptr) {
+            return false;
+        }
```
certik: Given that this function is only called once, I would remove the `bool` return. Make the function always succeed. And yank the above logic into the caller where I think it belongs. Rename the function from `try_infer_and_declare` to `infer_type_and_declare`.

tests/reference/asr-infer_first_assignment_derived_01-940ee27.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Type inference only supports intrinsic types (integer, real, complex, logical, character)
```
certik: I would say "for now", right? We should support all types, no?

krystophny:
Addressed review feedback in commit c2d2ca85f:
- Clarified parser comment for REPL infer default behavior.
- Renamed/refactored `try_infer_and_declare` -> `infer_type_and_declare` (void), with preconditions moved to `visit_Assignment`.
- Moved the statement-function doc block back above `is_statement_function`.
- Updated diagnostic wording to "For now, ..." and refreshed reference stderr/hash.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -435,6 +440,19 @@ namespace LCompilers::CommandLineInterface {
...
 
+        // REPL invocation (`lfortran` with no input files) defaults to infer mode,
+        // unless a language mode was explicitly selected via `--std`.
+        if (!opts.arg_infer && opts.arg_standard.empty() && opts.arg_files.empty()) {
+            opts.arg_infer = true;
+        }
+        compiler_options.infer_mode = opts.arg_infer;
+        if (compiler_options.infer_mode) {
+            compiler_options.po.realloc_lhs_arrays = true;
+        }
```
certik: I don't know if this is what we want, similar how we do not want to redeclare on `:=`, we probably do not want to reallocate on `=`. Although to automatically allocate might be helpful, but not needed with `--infer` most of the time?

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -435,6 +440,19 @@ namespace LCompilers::CommandLineInterface {
             }
         }
 
+        // REPL invocation (`lfortran` with no input files) defaults to infer mode,
+        // unless a language mode was explicitly selected via `--std`.
```
certik: I would not enable `--infer` by default for now, until we get it working well. I think right now it only works for some types, correct?

certik (commented):
More review is needed, I need to play with it locally. I will get to it soon. It looks quite good overall.

tests/infer_realloc_defaults_01.f90
```diff
@@ -0,0 +1,6 @@
+program infer_realloc_defaults_01
+    integer, allocatable :: x(:)
+    x = [1, 2, 3]
+    x = [4, 5]
+    if (size(x) /= 2) error stop
+end program infer_realloc_defaults_01
```
certik: I would move these to integration test and hook it in cmake. It's a mode of LFortran, one of many.

krystophny:
Addressed in 6bb658143: removed REPL default and realloc-lhs coupling. Both are now explicit opt-in only.

tests/infer_realloc_defaults_01.f90
krystophny: Done in 5e1abfd. Moved to `integration_tests/realloc_lhs_01.f90` with `--realloc-lhs-arrays` in CMake.

tests/tests.toml
```diff
@@ -5051,11 +5051,6 @@ filename = "infer_global_scope_file_01.f90"
 ast = true
 options = "--infer"
 
-[[test]]
-filename = "infer_realloc_defaults_01.f90"
-run = true
-options = "--infer --realloc-lhs-arrays"
```
certik, line 5057: Hold on, now you don't check `--infer` in cmake.

integration_tests/CMakeLists.txt
```diff
@@ -1995,6 +1995,8 @@ RUN(NAME allocate_38 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_39 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocate_40 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
+RUN(NAME realloc_lhs_01 LABELS gfortran llvm
+    EXTRA_ARGS --realloc-lhs-arrays)
```
certik, line 1999: You have to put `--infer` here. Likely as a `llvm_infer` label.

The same for all the other tests.

PR #10190, opened 2026-02-18
feat: := walrus syntax with scope shadowing, array, and derived type inference

krystophny:
## Summary
- Add `:=` walrus assignment for explicit declaration + type inference in regular code (Go-style declare-once semantics).
- Keep `--infer` and `:=` behavior distinct: `:=` always declares, while `=` only infers in `--infer` mode.
- Final review cleanup removes obsolete `infer_first_assignment_derived_01` error-test artifacts.

## Final Review Follow-ups (2026-02-20)
- Removed obsolete error-test entry from [`tests/tests.toml#L5073`](https://github.com/lfortran/lfortran/blob/cc72d6ccad71fa1d10682f071e431f384bcad55d/tests/tests.toml#L5073).
- Removed obsolete references:
  - [`tests/reference/asr-infer_first_assignment_derived_01-940ee27.json`](https://github.com/lfortran/lfortran/blob/523109463ab5db42a2b0c297a2cdf805d20b8bd9/tests/reference/asr-infer_first_assignment_derived_01-940ee27.json)
  - [`tests/reference/asr-infer_first_assignment_derived_01-940ee27.stdout`](https://github.com/lfortran/lfortran/blob/523109463ab5db42a2b0c297a2cdf805d20b8bd9/tests/reference/asr-infer_first_assignment_derived_01-940ee27.stdout)
- Interactive mode still does **not** make infer default:
  - `arg_infer` default is false in [`src/bin/lfortran_command_line_parser.h#L32`](https://github.com/lfortran/lfortran/blob/cc72d6ccad71fa1d10682f071e431f384bcad55d/src/bin/lfortran_command_line_parser.h#L32)
  - infer mode is only enabled from `--infer` in [`src/bin/lfortran_command_line_parser.cpp#L469`](https://github.com/lfortran/lfortran/blob/cc72d6ccad71fa1d10682f071e431f384bcad55d/src/bin/lfortran_command_line_parser.cpp#L469)

## Verification
```bash
$ scripts/lf.sh check
PASSED: All checks OK.

$ scripts/lf.sh build --clean
FAILED (pre-existing branch build inconsistency in generated parser/AST artifacts, e.g. missing InferAssignment symbols in generated paths)
```


tests/reference/asr-infer_walrus_error_01-87be84d.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Variable 'x' is already declared; use `=` for assignment to existing variables
+ --> tests/errors/infer_walrus_error_01.f90:4:5
+  |
+4 |     x := 5
+  |     ^ already declared
```
certik: So I think this might be very annoying. Imagine having a jupyter notebook cell:
```
x = 5
print *, x
```
So with `--infer` I think it should declare it on first run, but reuse it on the second run.

If you change it to:
```
x := 5
print *, x
```
Then you can't rerun it.

Well, actually we run every cell in a new nested symbol table, so you probably can redefine. So probably it's fine.

Still worth figuring out the right design here.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5822,6 +5821,42 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        if (existing != nullptr) {
+            diag.add(Diagnostic(
+                "Variable '" + var_name + "' is already declared; "
+                "use `=` for assignment to existing variables",
+                Level::Error, Stage::Semantic, {
+                    Label("already declared", {target_name->base.base.loc})
+                }
+            ));
+            throw SemanticAbort();
+        }
```
certik, line 5946: I think we should give a warning that `:=` is an LFortran extension, unless `--infer` is used (or another option).

All extensions should give a warning. Also in this case say it's experimental (subject to change).

tests/reference/asr-infer_walrus_error_01-87be84d.stderr
certik: @difference-scheme, what do you think is the right design here? Should `:=` always declare and thus fail if it is already declared? Or should it declare on first use, but reuse on second use?

krystophny: Jumping in with some justification: I started this to be as close to Go behavior as possible. In notebooks one would use the actual infer mode anyway and even omit `:=`. In production code, letting `:=` do non-declaration assignment too will make it harder to reason about code. If we allow this, in order not to confuse people used to Rust, we would need to adopt rust immutable behavior. Think of `intent(out)` and 

```fortran
y := f(x)
y := g(y)
```

In Rust this is perfectly legal with `let`, as the second line would mutate the first `y` and generate an immutable new one. In Fortran the second becomes equivalent to `y = g(y)` and I think this is not well defined. Maybe there are more issues, but this is the first one I could come up with. So to sum up: If we want Go style `:=` and not Rust style `let`, the easy way is to trust the Go design to be well thought-out.

krystophny: ... also, `:=` that doesn't allow acting as `=` sends a strong signal *This variable is declared here*

difference-scheme: I agree with @krystophny that we should try to keep it as close to the Go design as possible (at least as far as the regular mode is concerned), for the reasons he mentioned (having to adopt Rust immutable behavior would make things very complicated). I'll try to think more about this over the weekend.

certik: I see --- yes, let's use the Go design and make `:=` always declare. And thus the above error is correct.

And I see --- this is to be used in production codes, not interactively.

certik: So I think this behaves like the auto keyword in C++ I think. I only use it for very complicated types. I don't like using it for simple types, as I like to see what the types are from reading the source code, for clarity and for checking (if the function changes its return type, it will fail to compile, as it should).

So do we want this in Fortran? I think @difference-scheme had a use case for it in the generic proposal. Here is what Grok says: https://grok.com/share/bGVnYWN5_5aa57159-8cd8-45f1-8044-5759aaf77faf, which I find very reasonable. Let me know if you agree with it.

I would say as long as this gives a warning every time, then I think it's no harm to prototype and experiment with it.

Important: let's write documentation what the intent is of this and how we recommend using it and when not to use it. This will be perhaps the most important part to get right. And then let's guide users with warnings an hints.

difference-scheme: I think we should have it. As you said there are use cases of it in the Traits-proposal, and I think these demonstrate that it is a feature that makes Fortran code more concise, with a "feel" much closer to what modern languages offer. 

I found myself enjoying writing Go code with that feature. The original request (of providing something like C++'s auto) actually came from @meow464, and since then we had generally positive feedback on the idea to adopt this "walrus operator" for this purpose. So I think we should at least prototype and test it.

certik: Yes, we'll definitely prototype and test it. It's compelling and the only way to know is to use it. And if we convince ourselves after using it that it is not worth it, we can always remove it.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -461,6 +466,19 @@ namespace LCompilers::CommandLineInterface {
...
+        if (!opts.arg_infer && opts.arg_standard.empty() && opts.arg_files.empty()) {
+            opts.arg_infer = true;
+        }
+        compiler_options.infer_mode = opts.arg_infer;
+        if (compiler_options.infer_mode) {
+            compiler_options.po.realloc_lhs_arrays = true;
+        }
+        if (disable_realloc_lhs) {
+            compiler_options.po.realloc_lhs_arrays = false;
+        }
```
certik: Let's not make `--infer` the default in interactive mode, because there is no way to turn it off, and we can always turn it on later without breaking things. This can be fixed in the next PR.

certik (approved):
I played with it locally. I think it works. It doesn't work in interactive mode:
```
(lf) ~/repos/lfortran/integration_tests(feature/walrus-syntax)$ lfortran
LFortran version: 0.60.0-274-g4fbfbeb70
  * Use Ctrl-D to exit
  * Use Enter to submit
    - History (Keys: Up, Down)
>>> x := 5TOKEN: 293                                                     1,7   ]
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2805
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2518
    return prompt(opts.arg_v, compiler_options);
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 432
    if (input.size() == 1 && input[0] == CTRL_KEY('d')) {
  File "/Users/ondrej/repos/lfortran/src/bin/tpl/cpp-terminal/prompt0.h", line 229
    not_complete = !iscomplete(concat(m.lines));
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__functional/function.h", line 995
    return __f_(std::forward<_ArgTypes>(__arg)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__functional/function.h", line 436
    return (*__f_)(std::forward<_ArgTypes>(__args)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__functional/function.h", line 319
    return __f_(std::forward<_ArgTypes>(__arg)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__functional/function.h", line 167
    return std::__invoke_r<_Rp>(__func_, std::forward<_ArgTypes>(__arg)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__type_traits/invoke.h", line 273
    return __invoke_void_return_wrapper<_Ret>::__call(std::forward<_Args>(__args)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__type_traits/invoke.h", line 243
    return std::__invoke(std::forward<_Args>(__args)...);
  File "/Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr/include/c++/v1/__type_traits/invoke.h", line 179
    { return static_cast<_Fp&&>(__f)(static_cast<_Args&&>(__args)...); }
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 312
    int tok_ret = emit_tokens2(command, token_strings, toks, stypes);
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 298
    tok_strings.push_back(LCompilers::LFortran::pickle(toks[i], stypes[i]));
  File "/Users/ondrej/repos/lfortran/src/lfortran/pickle.cpp", line 59
    t += token2text(token);
  File "/Users/ondrej/repos/lfortran/src/lfortran/parser/parser.cpp", line 1138
    throw LCompilersException("Token conversion not implemented yet.");
LCompilersException: Token conversion not implemented yet.
```

As I mentioned below, let's not make `--infer` the default in interactive mode yet. We'll do it later.

certik:
Let's merge https://github.com/lfortran/lfortran/pull/10189 first. Then we'll rebase this one, you can address the feedback above and let's do one final review.

krystophny:
> Let's merge #10189 first. Then we'll rebase this one, you can address the feedback above and let's do one final review.

Clear. I'll push already the rebased version with all points you mentioned fixed once #10189 is on main.

tests/reference/asr-infer_first_assignment_derived_01-940ee27.stderr
```diff
@@ -1,5 +0,0 @@
-semantic error: For now, type inference only supports intrinsic types (integer, real, complex, logical, character)
- --> tests/errors/infer_first_assignment_derived_01.f90:4:5
-  |
-4 | a = t(1)
-  |     ^^^^ non-intrinsic type
```
certik, line 5: If this is removed now, it means we support it? I think this is a mistake to remove this test reference.

tests/tests.toml
```diff
@@ -5079,3 +5079,15 @@ options = "--infer"
...
+filename = "errors/infer_walrus_error_01.f90"
+asr = true
+
+[[test]]
+filename = "errors/infer_walrus_error_02.f90"
+asr = true
+
+[[test]]
+filename = "errors/infer_walrus_redecl_01.f90"
+asr = true
```
certik: Add all these into just one file: tests/errors/continue_compilation_walrus.f90

tests/reference/asr-infer_walrus_error_02-3990bfd.stdout
```diff
@@ -0,0 +1,120 @@
...
+                        (Var 2 a)
+                        (Var 2 obj)
+                        ()
+                        .false.
+                        .false.
+                    )]
+                )
+        })
+    []
+)
```
certik: I think we don't need to check asr here. But this will be fixed by the move to continue compilation below.

certik:
The code changes are good, let's fix the tests, per above.

krystophny:
Addressed review feedback in 56bedb88d:

- **Consolidated walrus error tests** into `tests/errors/continue_compilation_walrus.f90` with `semantics_only_cc = true` (no ASR check), as requested
- **Removed** individual `infer_walrus_error_01`, `infer_walrus_error_02`, `infer_walrus_redecl_01` test files and their references
- **Kept `infer_first_assignment_derived_01` stderr removal** — verified that `a = t(1)` with derived type inference now works correctly under `--infer` (returncode 0, valid ASR output)

Still open from review (deferred to next PR per discussion):
- `--infer` not made default in interactive mode

krystophny:
Addressed feedback from #10189 review in 213ca5063:

- **Renamed** `infer_and_declare` → `infer_type_and_declare` (per [review comment](https://github.com/lfortran/lfortran/pull/10189#discussion_r2082839893))

krystophny:
Fixed in 597679fe0: added `EXTRA_ARGS --infer` to all 12 walrus integration tests in CMakeLists.txt (per #10189 [review comment](https://github.com/lfortran/lfortran/pull/10189#discussion_r2085652303)). Tests now run warning-free.

krystophny:
Fixed in 5d962430d: added `llvm_infer` backend label per #10189 [review](https://github.com/lfortran/lfortran/pull/10189#discussion_r2085652303).

- New `llvm_infer` backend in `CMakeLists.txt` and `run_tests.py` (same pattern as `llvmImplicit`, `llvm_integer_8`)
- All 12 walrus tests use `LABELS llvm_infer` — no more `EXTRA_ARGS --infer`
- Run with `./run_tests.py -b llvm_infer`

tests/tests.toml
```diff
@@ -5079,3 +5079,7 @@ options = "--infer"
 filename = "errors/infer_mode_conflict_01.f90"
 run = true
 options = "--infer --std=f23"
+
+[[test]]
+filename = "errors/continue_compilation_walrus.f90"
+semantics_only_cc = true
```
certik, line 5080: Can we run this with `--infer`, or would it require modification to Python? If so, don't do it. Otherwise do it, so that we don't have those warnings there.

tests/errors/continue_compilation_walrus.f90
```diff
@@ -0,0 +1,12 @@
+program continue_compilation_walrus
+    implicit none
+    integer :: x
+    type :: t
+        integer :: n
+    end type
+    type(t) :: obj
+    x := 5
+    a := obj
```
certik: This line didn't produce an error. So it is now supported? If so, remove it from here.

krystophny:
Fixed in 523109463:

- Added `options = "--infer"` to the `continue_compilation_walrus` test — warnings suppressed, only actual errors in stderr now
- Removed `a := obj` line — struct inference from typed variable works (not an error case)

tests/reference/asr-infer_first_assignment_derived_01-940ee27.stdout
```diff
@@ -0,0 +1,94 @@
...
+                .true.
+                .false.
+            )
+            ()
+        )
+        ()
+        .false.
+        .false.
+    )]
+)
```
certik: This file can be removed.

tests/reference/asr-infer_first_assignment_derived_01-940ee27.json
certik: remove this file, since it works now?

tests/tests.toml
certik: This can be removed, since it's not an error anymore.

krystophny:
Final follow-up pushed in cc72d6cca:

- removed obsolete infer_first_assignment_derived_01 reference files (.json/.stdout)
- removed its tests.toml error-test entry
- confirmed interactive mode still keeps infer opt-in only, not default

PR description is updated accordingly.

PR #10191, opened 2026-02-18
Fix crash on block array with deferred dim

### LINKED ISSUE #10186: Code generation fails on deferred type-bound procedure call within a `block` statement
This is a variation of #2882 that still fails to compile, apparently because the (deferred) method call is made from within a `block` statement.

`$ cat testmod.f90 `

```
module testmod

   type, abstract :: AbsType
   contains
      procedure(nelements), deferred, nopass :: nelements
   end type AbsType

   abstract interface
      pure function nelements() result(n)
         import
         integer :: n
      end function nelements
   end interface
   
   type :: MyType
   contains
      procedure :: caller
   end type MyType

contains

   subroutine caller(self)
      class(MyType), intent(in) :: self
      block
        real(8), dimension(self%obj%nelements()) :: a
      end block
   end subroutine caller

end module testmod
```
`$ lfortran -c --separate-compilation testmod.f90 `
```
code generation error: x is nullptr in load_array_size_deep_copy()


```

certik:
Resolve postponed type-bound procedure calls in dimension specifications inside Block constructs. The body visitor's visit_Block was not resolving these deferred calls, leaving m_length as nullptr and causing an assertion failure in the LLVM codegen's fill_array_details.

Fixes #10186.


krystophny:
I had a look and tested locally, and it appears fine to me.

PR #10201, opened 2026-02-18
Initial impl. of floating point exception trapping

### LINKED ISSUE #10179: Implement trapping floating point exceptions
When a NaN is created or arithmetic on it happens, stop the code. GFortran and most other compilers can do that. This greatly helps with debugging. Make it optional with a compiler option.

certik:
Add a new --fpe-trap compiler option that enables hardware floating point exception trapping at program startup, similar to GFortran's -ffpe-trap.

The option accepts a comma-separated list of exception types:
  invalid, zero, overflow, underflow, inexact, denormal

Example usage:
  lfortran --fpe-trap invalid,zero,overflow program.f90

Implementation details:
- Add fpe_traps field to CompilerOptions (utils.h)
- Register --fpe-trap CLI option with comma-separated validation
- Add _lfortran_enable_fpe_traps() runtime function that configures hardware FPE traps using a bitmask encoding
- Emit call to the runtime function in LLVM codegen (visit_Program) when --fpe-trap is specified
- Cross-platform support:
  - Linux: feenableexcept() + MXCSR for denormals
  - macOS x86_64: fenv_t manipulation + MXCSR
  - macOS ARM64: FPCR register via inline asm
  - Windows: _controlfp_s()
  - WASM: warning (not supported)

Fixes #10179.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -1,3 +1,6 @@
+#if defined(__linux__) && !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -13,6 +16,9 @@
 #include <limits.h>
 #include <stdint.h>
 #include <stddef.h>  /* ptrdiff_t */
+#if !defined(COMPILE_TO_WASM)
+#include <fenv.h>
+#endif
 
 #define PI 3.14159265358979323846
 // Enum for float format types to avoid string comparison
@@ -100,6 +106,7 @@ static int _lfortran_use_runtime_colors = 0;  // disabled by default
 #ifdef HAVE_LFORTRAN_LINK
 // For dl_iterate_phdr() functionality
 #  include <link.h>
+#ifndef _GNU_SOURCE
 struct dl_phdr_info {
     ElfW(Addr) dlpi_addr;
     const char *dlpi_name;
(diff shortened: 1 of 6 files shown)


certik:
See the example `tests/fpe_trap_test.f90` how to use this.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -100,6 +106,7 @@ static int _lfortran_use_runtime_colors = 0;  // disabled by default
 #ifdef HAVE_LFORTRAN_LINK
 // For dl_iterate_phdr() functionality
 #  include <link.h>
+#ifndef _GNU_SOURCE
```
certik, line 109: I forgot what this does and why it is needed. Either remove it or document why it is needed and what it does.

krystophny: `_GNU_SOURCE` is needed because `feenableexcept()` is a GNU extension only declared in `<fenv.h>` when that macro is set. The `#ifndef` guard around `dl_phdr_info prevents` duplicate definitions since `<link.h>` already provides them when `_GNU_SOURCE` is active. The cleanest fix is to add `target_compile_definitions(... PRIVATE _GNU_SOURCE)` in CMake for Linux, then delete both the `#define` at the top and the manual `struct/extern`.

certik:
Tests seem to pass, so let's merge this and then we'll improve upon this initial implementation.

src/libasr/runtime/lfortran_intrinsics.c
certik: I made this https://github.com/lfortran/lfortran/issues/10214.

PR #10202, opened 2026-02-18
Add check_trailing_whitespace script

certik:


certik:
Perfect, the CI works and fails as it should. Now we'll wait until most PRs are in and in the right moment push in the whitespace cleanup commit and merge it.

certik:
We have a lot of open PRs now, so we'll wait with this change until our queue is small.

check_trailing_whitespace.py
HarshitaKalani, line 1: We need to remove this file.

PR #10203, opened 2026-02-19
fix: add missing file_ends entries in kernel evaluation paths

krystophny:
## Summary

Every `lm.files.push_back()` must be paired with `lm.file_ends.push_back()` or `LocationManager::bisection()` will index out of bounds when rendering diagnostics. The Jupyter kernel had 7 sites missing the `file_ends` entry.

**Stage:** Driver (kernel frontend)

## Changes

- [`src/lfortran/fortran_kernel.cpp`](https://github.com/lfortran/lfortran/blob/d516b6283/src/lfortran/fortran_kernel.cpp): add `lm.file_ends.push_back(code0.size())` after each `lm.files.push_back(fl)` in all 7 kernel evaluation paths

## Why

`LocationManager` uses `bisection(file_ends, out_pos)` to map error positions back to source files. When `files` and `file_ends` vectors are out of sync, any diagnostic (e.g. a syntax error in a Jupyter cell) causes an out-of-bounds access. Every other call site in the codebase (`lfortran.cpp`, `lfortran_accessor.cpp`, tests, passes) pairs these correctly -- the kernel was the only place missing it.

### DIFF
--- a/src/lfortran/fortran_kernel.cpp
+++ b/src/lfortran/fortran_kernel.cpp
@@ -129,6 +129,7 @@ namespace LCompilers::LFortran {
                     std::ofstream out("input");
                     out << code0;
                     lm.files.push_back(fl);
+                    lm.file_ends.push_back(code0.size());
                 }
                 diag::Diagnostics diagnostics;
                 Result<std::string>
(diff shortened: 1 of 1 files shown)


krystophny:
Closing: file_ends is empty but bisection() returns 0 for empty vectors, so files[0] is accessed correctly. The kernel creates a fresh LocationManager with exactly one file per cell evaluation, so this is not a real bug. Hygiene fix not worth a CI run.

Issue #6097, opened 2025-01-23
interactive lfortran bug with --std=f23


harperjf:
Strange result of trying a nonexistent Fortran statement interactively:
```
lf) john:~$ lfortran-0.44.0/inst/bin/lfortran --std=f23
Interactive Fortran. Experimental prototype, not ready for end users.
LFortran version: 0.44.0
  * Use Ctrl-D to exit
  * Use Enter to submit
  * Use Alt-Enter or Ctrl-N to make a new line
    - Editing (Keys: Left, Right, Home, End, Backspace, Delete)
    - History (Keys: Up, Down)
>>> foobar                                                               1,7   ]
0
>>>                                                                      1,1   ]

Exiting.
(lf) john:~$
```
Trying the same nonexistent statement in this 2-line program:
```
 foobar
end program
```
gave this correct result:
```
(lf) john:~$ lfortran-0.44.0/inst/bin/lfortran --std=f23 badstmt.F90

syntax error: Statement or Declaration expected inside program, found Variable name
 --> badstmt.F90:1:3
  |
1 |   foobar
  |   ^^^^^^ 
```
Without the ``--std=f23``` options the interactive version gave a semantic instead of syntax error, but also correct:
```
semantic error: Variable 'foobar' is not declared
 --> input:1:1
  |
1 | foobar
  | ^^^^^^ 'foobar' is undeclared
```
and the 2-line program gave the same syntax error as before.

harperjf:
I would have expected what ```lfortran-0.44.0/inst/bin/lfortran --std=f23 badstmt.F90``` did, i.e. a syntax error starting
```
syntax error: Statement or Declaration expected inside program, found Variable name
```
because  ```foobar``` is now a valid variable name but it can't be a statement by itself.

harperjf:
As I pressed Enter after typing foobar I wondered if lfortran --std=f23 was waiting for more of the statement. So I tried
```
>>> foobar;                                                              1,8   ]
0
>>> foobar; end                                                          1,12  ]
0
>>> foobar; end program                                                  1,20  ]
0
>>>  
```
Those did not give the expected syntax error. If foobar was interpreted as a real variable its value could have been ```0.0``` or ```0.``` or ```.0``` but ```0``` is an integer value.

kmr-srbh:
Thanks for reporting this. This looks like an issue with printing. If we look at the ASR for the code in interactive mode:

```fortran
foobar
```

```console
[...]

PR #10208, opened 2026-02-19
fix: show real values with decimals in interactive REPL

krystophny:
## Summary
- print REPL `real` / `real(8)` evaluation results with `std::showpoint` so decimal output is explicit

Addresses #6097

## Why
Interactive REPL output currently prints real-valued expressions like `5.0` and implicit `f23` variables as `5` / `0`, which looks like integer output and is misleading.

**Stage:** Driver / interactive REPL frontend

## Changes
- [`src/bin/lfortran.cpp#L501-L513`](https://github.com/lfortran/lfortran/blob/38c5c490bfd2d40b30d4cc43aab4e54dfa8260a4/src/bin/lfortran.cpp#L501-L513): enable `showpoint` in real result printing branches and restore stream state afterward

## Tests
- REPL output checks in `--std=f23` mode for:
  - `5.0`
  - `foobar` (implicitly typed)
- reference test sanity check: `interactive_parse_without_program_line`

### Test fails on main
```bash
$ git -C lfortran switch --detach upstream/main
$ scripts/lf.sh build
$ printf '5.0\n' | ./lfortran/build/src/bin/lfortran --std=f23
...
5

$ printf 'foobar\n' | ./lfortran/build/src/bin/lfortran --std=f23
...
0
```

### Test passes after fix
```bash
$ git -C lfortran switch fix/repl-real-output-format
$ scripts/lf.sh build
$ printf '5.0\n' | ./lfortran/build/src/bin/lfortran --std=f23
...
5.0000000

$ printf 'foobar\n' | ./lfortran/build/src/bin/lfortran --std=f23
...
0.0000000
```

```bash
$ scripts/lf.sh test -t interactive_parse_without_program_line -s
TESTS PASSED
```

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -501,13 +501,15 @@ int prompt(bool verbose, CompilerOptions &cu)
             case (LCompilers::FortranEvaluator::EvalResult::real4) : {
                 if (verbose) std::cout << "Return type: real" << std::endl;
                 if (verbose) section("Result:");
-                std::cout << std::setprecision(8) << r.f32 << std::endl;
+                std::cout << std::showpoint << std::setprecision(8)
+                          << r.f32 << std::noshowpoint << std::endl;
                 break;
             }
             case (LCompilers::FortranEvaluator::EvalResult::real8) : {
                 if (verbose) std::cout << "Return type: real(8)" << std::endl;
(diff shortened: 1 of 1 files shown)


krystophny:
Closing: REPL output formatting preference, not a bug fix.

PR #10213, opened 2026-02-19
Fix : Apply pass `pass_array_by_data` on class methods instead of bypassing

assem2002:
Improve #10185
- Make sure to use the correct index on fun/sub calls while using a class method, as the first parameter is the `dt` while the funcall itself doesn't pass it it's implicitly passed.
- Make sure to update `StructMethodDeclaration->m_proc` properly.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -999,6 +999,7 @@ RUN(NAME submodule_20a LABELS gfortran llvm_submodule EXTRAFILES submodule_20b.f
 RUN(NAME submodule_21a LABELS gfortran llvm_submodule EXTRAFILES submodule_21b.f90 submodule_21c.f90)
 
 RUN(NAME submodule_22 LABELS gfortran llvm)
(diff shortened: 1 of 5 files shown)


krystophny:
PR checks fail because the code now rewrites type-bound method signatures in `pass_array_by_data` but not all dispatch call paths are rewritten the same way, so callers and callees disagree on argument layout (causing `operator_overloading_25/class_101` segfaults and `splpak` "Incorrect number of arguments passed" LLVM errors).

The fix is to apply one canonical rewritten signature consistently to the concrete proc, StructMethodDeclaration aliases, vtable entries, and every direct/deferred/procedure-pointer call lowering (including implicit PASS handling), or temporarily disable this transform for PASS methods until that is fully implemented.

certik:
I am trying to fix this in https://github.com/lfortran/lfortran/pull/10216.

certik:
Merged as part of https://github.com/lfortran/lfortran/pull/10216.

PR #10222, opened 2026-02-19
ci: defer checks for low-priority PRs until queue is idle

krystophny:
## Summary
- PRs tagged `Tests::Run-When-Idle` skip CI until runners are idle
- Hourly cron job retriggers one deferred PR by removing the label
- Normal PRs (no label) run CI immediately as before

## Why
Low-priority PRs compete for the same GitHub-hosted runners as urgent
work. This label-based mechanism defers their CI to off-peak windows.

## Changes
- [`Quick-Checks-CI.yml`](https://github.com/lfortran/lfortran/blob/ed54479c57bc034a937811c7316929c2108a94c4/.github/workflows/Quick-Checks-CI.yml): add `labeled`/`unlabeled` trigger types, add `if:` guard to both jobs
- [`Exhaustive-Checks-CI.yml`](https://github.com/lfortran/lfortran/blob/ed54479c57bc034a937811c7316929c2108a94c4/.github/workflows/Exhaustive-Checks-CI.yml): AND idle guard into all 8 gated jobs
- [`Retrigger-Low-Priority-CI.yml`](https://github.com/lfortran/lfortran/blob/ed54479c57bc034a937811c7316929c2108a94c4/.github/workflows/Retrigger-Low-Priority-CI.yml) (new): hourly cron that checks queue, removes label from one PR when idle

## Design: retrigger via label removal

The cron job retriggers CI by **removing** the `Tests::Run-When-Idle` label. This is the
only mechanism that produces proper green/red check results:
- `workflow_dispatch` has no PR context, so checks don't appear on the PR
- Re-requesting check suites doesn't help because the `if:` condition still sees the label
- Removing the label fires the `unlabeled` event, the `!contains` guard passes, and jobs run normally

The label is consumed when cron fires. To re-defer after new commits, re-add the label.

## Orthogonality matrix

| Labels on PR | Quick Checks | Exhaustive Checks |
|---|---|---|
| (none) | immediate | skipped |
| `Tests::Run-Exhaustive` | immediate | immediate |
| `Tests::Run-When-Idle` | deferred | skipped |
| both labels | deferred | deferred |

## Verification
Once the `Tests::Run-When-Idle` label is created on the repo and applied to this PR,
Quick Checks should be skipped. Removing the label manually will retrigger CI immediately.


certik:
Good idea. But I am worried about the complexity of this.

I think rather we might need to do a three stage CI runs: just a very simple run in Ubuntu for new PRs. Then the Quick tests with a label and then all tests with another label.

krystophny:
Simpler variant in #10227

PR #10223, opened 2026-02-19
Fix select rank(0) scalar assignment

### LINKED ISSUE #10221: Implement select rank(0)
MRE:
```fortran
program p
  implicit none

  type :: test_diagnosis_t
    integer :: i = 0
  end type

contains

  function aggregate_diagnosis(diagnoses) result(diagnosis)
    type(test_diagnosis_t), intent(in) :: diagnoses(..)
    type(test_diagnosis_t) :: diagnosis

    select rank (diagnoses)
      rank (0)
        diagnosis = diagnoses
      rank (1)
        diagnosis = diagnoses(1)
    end select
  end function

end program
```
A full test:
```fortran
program select_rank_14
    implicit none

    type :: test_diagnosis_t
        integer :: i = 0
    end type

    type(test_diagnosis_t) :: scalar_diag
    type(test_diagnosis_t) :: array_diag(2)
    type(test_diagnosis_t) :: result_diag

    scalar_diag%i = 5
    array_diag(1)%i = 11
    array_diag(2)%i = 13

    result_diag = aggregate_diagnosis(scalar_diag)
    if (result_diag%i /= 5) error stop

    result_diag = aggregate_diagnosis(array_diag)
    if (result_diag%i /= 11) error stop

contains

    function aggregate_diagnosis(diagnoses) result(diagnosis)
        type(test_diagnosis_t), intent(in) :: diagnoses(..)
        type(test_diagnosis_t) :: diagnosis

        select rank (diagnoses)
            rank (0)
                diagnosis = diagnoses
            rank (1)
                diagnosis = diagnoses(1)
            rank default
                diagnosis%i = -1
        end select
    end function aggregate_diagnosis

end program select_rank_14
```

certik:
Inside a select rank block with rank(0), an assumed-rank variable should be treated as a scalar. Previously, using such a variable on the RHS of an assignment produced a type mismatch error because its type remained as an assumed-rank array.

Fix by creating an implicit shadow variable with scalar Pointer type in the rank(0) block scope (when no explicit association name is given). This mirrors the existing m_assoc_name mechanism. Also generalize the m_assoc_name rank-0 path to work for all types (not just StructType), and relax the codegen guard in visit_Associate so the assumed-rank-to-scalar extraction works for non-class types as well.

Fixes #10221.

### DIFF
--- a/tests/reference/asr-continue_compilation_simd-4dab78a.json
+++ b/tests/reference/asr-continue_compilation_simd-4dab78a.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_simd-4dab78a.stderr",
-    "stderr_hash": "532e8662b9443ad4ca33c7e9949d4baf710a2dbf2220870530159759",
+    "stderr_hash": "7dbe5dce0f92a089dd83bcff73234bedef218e3d10328e83bf78c9cb",
     "returncode": 1
 }
--- a/tests/reference/asr-continue_compilation_simd-4dab78a.stderr
+++ b/tests/reference/asr-continue_compilation_simd-4dab78a.stderr
@@ -1,35 +1,35 @@
 semantic error: SIMD arrays must be 1 dimensional, but `simd_2d` has 2 dimensions
- --> tests/errors/continue_compilation_simd.f90:5:13
+ --> tests/errors/continue_compilation_simd.f90:5:5
   |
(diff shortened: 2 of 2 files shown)


certik:
It turned out the initial fix had several bugs, which should now be fixed, but I don't like the fix overall.

Let's figure out how to use a cast, I think that will be cleaner.

certik:
Here is a better fix: https://github.com/lfortran/lfortran/pull/10226.

PR #10225, opened 2026-02-20
Fix ICE in pass_array_by_data

certik:


certik:
I don't like this fix in the LLVM backend. Things are passing for me locally, but we need to figure out a much cleaner fix.

PR #10226, opened 2026-02-20
Fix select rank(0) type mismatch for scalars

certik:
In a `select rank(rank(0))` block, the selector variable should behave as a scalar. Previously, using it directly in an assignment (`y = x`) caused a type mismatch because the semantic layer had no handling for the assumed-rank → scalar conversion.

Changes:
- semantics: insert ArrayPhysicalCast(AssumedRankArray → DescriptorArray, scalar_type) for rank(0) assumed-rank variables used as RHS values in assignments
- asr_verify: allow ArrayPhysicalCast where m_type is scalar when m_old is AssumedRankArray (rank-0 case)
- asr_to_llvm: handle rank(0) in ArrayPhysicalCastUtil by loading the data pointer from the assumed-rank descriptor and dereferencing it
- pass_array_by_data: skip the Array_t downcast for rank-0 scalar ArrayPhysicalCast nodes

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -817,6 +817,7 @@ RUN(NAME select_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_12 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_13 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME select_rank_14 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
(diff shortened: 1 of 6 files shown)


certik:
For rank(0) we now insert this cast:
```
(ArrayPhysicalCast
                                                            (Var 4 x)
                                                            AssumedRankArray
                                                            DescriptorArray
                                                            (Real 4)
                                                            ()
                                                        )
```
For rank(1) this cast:
```
(ArrayPhysicalCast
                                                                (Var 4 x)
                                                                AssumedRankArray
                                                                DescriptorArray
                                                                (Array
                                                                    (Real 4)
                                                                    [(()
                                                                    ())]
                                                                    DescriptorArray
                                                                )
                                                                ()
                                                            )
```
So we still do AssumedRankArray -> DescriptorArray for rank(0), but the type is just `(Real 4)` instead of `(Array (Real 4) ...)`. It's a little confusing that the result is not a `DescriptorArray`, but ultimately the type is what matters and the type is correct. We could introduce a new cast AssumedRankArray -> Scalar, but we can do int later. The above seems to work. The most important is that types are correct and that the cast is inserted.

HarshitaKalani (approved):
lgtm. Thanks!

PR #10227, opened 2026-02-20
ci: gate quick checks behind Tests::Quick label

krystophny:
## Summary
- Unlabeled PRs run only Ubuntu LLVM 11 + 21 (build + unit/integration tests)
- Adding `Tests::Quick` label triggers the full quick-checks suite (all platforms, all backends)
- Push to main always runs the full suite
- Removes stale `unlabeled` trigger from Exhaustive checks

Implements the three-tier CI approach discussed in #10222.

## Why
Reduce CI load for new PRs. Most issues are caught by Ubuntu LLVM 11 + 21
alone. The full matrix (macOS, Windows, extra backends) runs on demand via label.

## Orthogonality matrix

| Labels on PR | Default (Ubuntu 11+21) | Full Quick (all platforms) | Exhaustive |
|---|---|---|---|
| (none) | immediate | skipped | skipped |
| `Tests::Quick` | skipped (covered by Full) | immediate | skipped |
| `Tests::Run-Exhaustive` | immediate | skipped | immediate |
| `Tests::Quick` + `Tests::Run-Exhaustive` | skipped | immediate | immediate |

## Changes
- `Quick-Checks-CI.yml`: Add `Build_Ubuntu` job (default tier), gate `Build` + WASM behind `Tests::Quick`
- `Exhaustive-Checks-CI.yml`: Remove `unlabeled` from PR trigger types

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -7,7 +7,7 @@ on:
     tags:
       - 'v*'
   pull_request:
-    types: [labeled, unlabeled, synchronize]
+    types: [labeled, synchronize]
     branches:
       - main
 
--- a/.github/workflows/Quick-Checks-CI.yml
+++ b/.github/workflows/Quick-Checks-CI.yml
@@ -7,6 +7,7 @@ on:
     tags:
       - 'v*'
   pull_request:
+    types: [labeled, unlabeled, synchronize, opened, reopened]
     branches:
       - main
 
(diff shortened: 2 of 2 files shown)


certik:
I added the `Tests::Run-Quick` label, it seems it is rerunning the tests. I canceled it to save the CI. Also it's only rerunning the two of them?

So I think it doesn't work. Also it is duplicating the CI logic it seems. 

Any change to a CI has to be carefully checked, which is quite time consuming.

krystophny:
> I added the `Tests::Run-Quick` label, it seems it is rerunning the tests. I canceled it to save the CI. Also it's only rerunning the two of them?
> 
> So I think it doesn't work. Also it is duplicating the CI logic it seems.
> 
> Any change to a CI has to be carefully checked, which is quite time consuming.

OK we can also defer this for later and just be more disciplined with PR amount for now. At some point we have to run all tests anyway.

Issue #4427, opened 2024-07-01
seg.fault from complex allocatable array 


harperjf:
When trying this program with lfortran
```
program cmplxinf
  implicit none
  complex,allocatable :: z(:)
  real                :: zero = 0
  allocate(z(3))
  z = cmplx(1/zero,1/zero)
   print "(A,3F9.0)", 'z%im =', z%im
end program cmplxinf
```
in my x86_64 Ubuntu 22.04 system, the result is:
```
(lf) john:~/Test$ lfortran cmplxinf.f90
Segmentation fault (core dumped)
(lf) john:~/Test$
```
The same program runs happily with gfortran, ifx and ifort, printing 
```
z%im = Infinity Infinity Infinity
```

HarshitaKalani:
There's an issue even when the array is not allocatable:
```fortran
program cmplxinf
    implicit none
    complex :: z(3)
    z = (1.0, 2.0)
    print *, z%re
end program
```
```console
$ lfortran b.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

%array.0 = type { float*, i32, %dimension_descriptor*, i1, i32 }
%dimension_descriptor = type { i32, i32, i32 }
%array = type { %complex_4*, i32, %dimension_descriptor*, i1, i32 }
%complex_4 = type <{ float, float }>

@0 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@serialization_info = private unnamed_addr constant [5 x i8] c"[R4]\00", align 1
@1 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define i32 @main(i32 %0, i8** %1) {
.entry:
  %array_descriptor = alloca %array.0, align 8
  %array_bound14 = alloca i32, align 4
  %array_bound10 = alloca i32, align 4
  %array_bound = alloca i32, align 4
  %__libasr_index_0_ = alloca i32, align 4
  %__libasr_index_0_1 = alloca i32, align 4
  %__libasr_index_0_2 = alloca i32, align 4
  %__libasr_index_0_3 = alloca i32, align 4
  %__libasr_index_0_4 = alloca i32, align 4
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %__libasr_created__complex_re_ = alloca %array, align 8
  %2 = getelementptr %array, %array* %__libasr_created__complex_re_, i32 0, i32 1
  store i32 0, i32* %2, align 4
[...]

PR #10228, opened 2026-02-20
fix: apply array indexing before ComplexRe/ComplexIm in resolve_variable2

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4427

### DIFF
--- a/integration_tests/complex_33.f90
+++ b/integration_tests/complex_33.f90
@@ -7,7 +7,6 @@ program complex_33
     integer :: i
     character(len=9) :: buf
 
-    ! --- Test from a.f90: Infinity formatting with %im on allocatable array ---
     allocate(z(3))
     z = cmplx(1.0/zero, 1.0/zero)
     do i = 1, 3
@@ -15,7 +14,6 @@ program complex_33
         if (adjustl(buf) /= "Infinity") error stop
     end do
 
-    ! --- Test from b.f90: whole-array %re on fixed-size array ---
     w = (1.0, 2.0)
     re_vals = w%re
     im_vals = w%im
@@ -24,7 +22,6 @@ program complex_33
         if (abs(im_vals(i) - 2.0) > 1e-6) error stop
     end do
 
-    ! Whole-array %re/%im on allocatable array
     z(1) = cmplx(1.0, 4.0)
     z(2) = cmplx(2.0, 5.0)
     z(3) = cmplx(3.0, 6.0)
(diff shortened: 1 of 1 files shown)


integration_tests/complex_33.f90
```diff
@@ -0,0 +1,46 @@
...
+
+    re_vals = z%re
+    im_vals = z%im
+
+    do i = 1, 3
+        if (abs(re_vals(i) - real(i)) > 1e-6) error stop
+        if (abs(im_vals(i) - real(i + 3)) > 1e-6) error stop
+    end do
+
+    ! --- Test from c.f90: scalar indexing z(i)%re and z(i)%im ---
```
certik: What are `a.f90` -- `c.f90`?

certik:
These look like 2 fixes intermixed.

integration_tests/complex_33.f90
HarshitaKalani: oops, removed them.

HarshitaKalani:
Yes, one MRE had two issues.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -9402,23 +9402,38 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    dim_vec.from_pointer_n_copy(al, m_dims, n_dims);
+                    ASR::ttype_t *real_type = ASR::down_cast<ASR::ttype_t>(
+                        ASR::make_Real_t(al, loc, kind));
+                    ASR::ttype_t* complex_arr_ret_type = ASRUtils::duplicate_type(al, real_type, &dim_vec,
+                        ASR::array_physical_typeType::DescriptorArray, true);
+                    if (var_name == "re") {
+                        return ASR::make_ComplexRe_t(al, loc, desc_arr, complex_arr_ret_type, nullptr);
+                    } else {
+                        return ASR::make_ComplexIm_t(al, loc, desc_arr, complex_arr_ret_type, nullptr);
+                    }
```
certik, line 9433: Can you explain what this does? I don't understand it. There is `val` which comes from `v_var` which is related to `v_variable`, but you already check that `v_variable` is an array. So what does `ASRUtils::is_array(ASRUtils::expr_type(val))` do and what is happening in this code?

HarshitaKalani: v_variable->m_type tells us the declared type is an array, but after applying any subscript arguments, the expression's type (expr_type(val)) may no longer be an array. The old code didn't account for indexing before %re/%im, so arr(i)%re was handled incorrectly.

Issue #10230, opened 2026-02-20
Allow to figure out the type of the variable in interactive mode


certik:
Especially with `--infer` we now need some way to print the type of the variable:
```
(lf) ~/repos/lfortran(main)$ lfortran --infer
Interactive Fortran. Experimental prototype, not ready for end users.
LFortran version: 0.60.0-276-g0319559ed
  * Use Ctrl-D to exit
  * Use Enter to submit
  * Use Alt-Enter or Ctrl-N to make a new line
    - Editing (Keys: Left, Right, Home, End, Backspace, Delete)
    - History (Keys: Up, Down)
>>> x = 4.5                                                              1,8   ]
>>> x                                                                    1,2   ]
4.5
>>>                                                                      1,1   ]
```
Let's investigate how exactly Julia does it. In Python we have "repr" and "str". For "print *, x" we would use the "str" approach which is what we already do. But for just "x" we can now use the "repr" approach and actually print the type together with the value. I think this might be what Julia does also.

krystophny:
Some ideas. Let me know what you prefer, or if we should go into completely different direction. (actually, Julia and Python don't print types automatically for simple types like int, float, etc. and Julia prints the array type usually)

Variable `it` would be printed as follows. Some looks a bit verbose. Maybe we could also just print non-default kinds as `1.0_8` or `1.0d0` and get away with minimal explicit type printing? 

### 1) Fortranic Decl-Style  

  a)
                                                                                                                               
  ```fortran                                                                                                                                                                  
  it :: real(8) = 4.5                                                                                                                                                         
  it :: integer(4) = 42                                                                                                                                                       
  it :: type(point) = point(x=1.0, y=2.0)                                                                                                                                     
  it :: integer(4), dimension(3) = [1, 3, 2] 
  ```

  b) a variant closer to declaration syntax

  ```fortran
  real(8) :: it = 4.5
  integer(4) :: it = 42
  logical(4) :: it = .true.
  character(12) :: it = "hello"
  type(point) :: it = point(x=1.0, y=2.0)
  integer(4) :: it(3) = [1, 3, 2]
   ```                           

  c) value first

  ```
  4.5 :: real(8)
  42 :: integer(4)
  .true. :: logical(4)
  "hello" :: character(12)
  point(x=1.0, y=2.0) :: type(point)
  [1, 3, 2] :: integer(4), dimension(3)
  [[1.0, 2.0], [3.0, 4.0]] :: real(8), dimension(2,2)
[...]

PR #10231, opened 2026-02-20
feat: add type_name() intrinsic for compile-time type inspection

krystophny:
## Summary
- Add `type_name(x)` intrinsic for compile-time type introspection (LFortran extension).
- Keep extension warnings in default mode and suppress them under `--infer`.
- Add `llvm_infer` backend label support and default integration-test parallelism to host CPU count.

Partially addresses #10230

## Why
`--infer` makes type declarations implicit, so users need a direct way to inspect inferred types. `type_name()` provides that introspection while we design the fuller interactive `repr` behavior discussed in #10230.

**Stage:** Semantics

## Changes
- [`src/libasr/pass/intrinsic_functions.h#L5142`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/src/libasr/pass/intrinsic_functions.h#L5142): add `TypeName` intrinsic implementation.
- [`src/libasr/asr_utils.h#L1014`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/src/libasr/asr_utils.h#L1014): extend `type_to_str_with_kind()` with missing `Logical` and `UnsignedInteger` handling.
- [`src/lfortran/semantics/ast_common_visitor.h#L11790`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/src/lfortran/semantics/ast_common_visitor.h#L11790): extension warning for `type_name` outside `--infer`; include in inquiry-function handling.
- [`src/libasr/intrinsic_func_registry_util_gen.py#L887`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/src/libasr/intrinsic_func_registry_util_gen.py#L887) and [`src/libasr/pass/intrinsic_function_registry.h#L669`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/src/libasr/pass/intrinsic_function_registry.h#L669): register intrinsic.
- [`integration_tests/type_name_01.f90`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/integration_tests/type_name_01.f90): coverage for scalar, logical, character, allocatable/fixed arrays.
- [`integration_tests/CMakeLists.txt#L3403`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/integration_tests/CMakeLists.txt#L3403) and [`integration_tests/run_tests.py#L8`](https://github.com/lfortran/lfortran/blob/8d28df21d91afc3dcae1b304d797493433859b56/integration_tests/run_tests.py#L8): add `llvm_infer` backend test path and default test thread count to `os.cpu_count()`.

## Design Notes
- `type_name()` uses `type_to_str_with_kind()` for most kinds.
- String spelling for `type_name()` is scoped locally (`"character"`) to avoid broad diagnostic wording changes in shared utilities.
- `create_TypeName` returns a plain `StringConstant` (no `TypeInquiry` wrapper) to keep string length semantics correct.

### Test fails on main
```bash
$ git -C lfortran checkout upstream/main
$ echo 'program t; implicit none; integer :: i; print *, type_name(i); end program' > /tmp/t.f90
$ ./lfortran/build/src/bin/lfortran /tmp/t.f90
semantic error: Function 'type_name' not found
```

### Test passes on this branch
```bash
$ git -C lfortran checkout feature/type-name-intrinsic
$ scripts/lf.sh build --clean
$ scripts/lf.sh test -t continue_compilation_1 -s
$ scripts/lf.sh itest -b llvm -t type_name_01
$ scripts/lf.sh itest -b llvm_infer -t type_name_01
```

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -1042,9 +1042,6 @@ static inline std::string type_to_str_with_kind(const ASR::ttype_t* t, ASR::expr
             ASR::Logical_t* logical_t = ASR::down_cast<ASR::Logical_t>(t);
             return "logical(" + std::to_string(logical_t->m_kind) + ")";
         }
-        case ASR::ttypeType::String: {
-            return "character";
-        }
         case ASR::ttypeType::Pointer: {
             return type_to_str_with_kind(ASRUtils::type_get_past_pointer(
                         const_cast<ASR::ttype_t*>(t)), expr) + " pointer";
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -5145,6 +5145,9 @@ namespace TypeName {
             ASR::ttype_t* /*t1*/, Vec<ASR::expr_t*> &args, diag::Diagnostics& /*diag*/) {
         ASR::ttype_t* arg_type = ASRUtils::extract_type(ASRUtils::expr_type(args[0]));
         std::string type_str = ASRUtils::type_to_str_with_kind(arg_type, args[0]);
+        if (ASR::is_a<ASR::String_t>(*arg_type)) {
+            type_str = "character";
+        }
         int len = type_str.size();
         return make_ConstantWithType(make_StringConstant_t, s2c(al, type_str),
             ASRUtils::TYPE(ASR::make_String_t(al, loc, 1,
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "4b02fadbdfc80d882ff1f1e73458c0f0d8250f2ed552509209400b52",
+    "stderr_hash": "4e3d8c9b0fa8d822456db4645f7f73d71e775d378c959b98a148b77a",
     "returncode": 1
 }
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -618,11 +618,11 @@ semantic error: Expected logical expression in if statement, but recieved real(4
 380 |     if (r1) r1 = 1.0
     |         ^^ real(4) expression, expected logical
 
-semantic error: Expected logical expression in if statement, but recieved character instead
+semantic error: Expected logical expression in if statement, but recieved string instead
    --> tests/errors/continue_compilation_1.f90:381:9
     |
 381 |     if (c1) c1 = 'a'
-    |         ^^ character expression, expected logical
+    |         ^^ string expression, expected logical
 
 semantic error: The ranks of the `array` and `mask` arguments of the `Sum` intrinsic must be the same
    --> tests/errors/continue_compilation_1.f90:386:18


certik:
Interesting idea. We should add a warning in default mode, and no warning with `--infer`.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -5138,6 +5139,67 @@ namespace NewLine {
...
+            case ASR::ttypeType::Logical: {
+                return "logical(" + std::to_string(ASR::down_cast<ASR::Logical_t>(t)->m_kind) + ")";
+            }
+            case ASR::ttypeType::String: {
+                return "character";
+            }
+            default:
+                return ASRUtils::type_to_str_with_kind(t, expr);
+        }
+    }
```
certik: I think we already have a function to convert an ASR type to Fortran. We should use that instead.

krystophny: Done! Also the warning should be present already.

src/libasr/asr_utils.h
```diff
@@ -1034,6 +1038,13 @@ static inline std::string type_to_str_with_kind(const ASR::ttype_t* t, ASR::expr
             ASR::Complex_t* complex_t = ASR::down_cast<ASR::Complex_t>(t);
             return "complex(" + std::to_string(complex_t->m_kind) + ")";
         }
+        case ASR::ttypeType::Logical: {
+            ASR::Logical_t* logical_t = ASR::down_cast<ASR::Logical_t>(t);
+            return "logical(" + std::to_string(logical_t->m_kind) + ")";
+        }
+        case ASR::ttypeType::String: {
+            return "character";
+        }
```
certik: This one is already handled in `type_to_str_fortran_symbol`. You can change it there, but for now let's not change it, that's a separate PR.

certik:
However I will note that this doesn't quite fix the problem in #10230. It should print the type automatically I think, not manually. But being able to get this manually is also helpful.

krystophny:
Follow-up pushed in 8d28df21d:

- scoped "character" spelling to `type_name()` only
- reverted shared `type_to_str_with_kind()` string wording change
- refreshed `continue_compilation_1` references accordingly
- kept `type_name_01` passing on `llvm` and `llvm_infer`

Also updated PR description to `Partially addresses #10230`.

certik:
Instead of `type_name`, why not `typeof(x)` which is already an existing syntax in Fortran? We just need to teach `print *, typeof(x)` to print it, so print, or StringFormat, has to accept the result of `typeof(x)`. The `typeof(x)` is not an intrinsic, only part of a declaration, but making it an intrinsic would make sense. It should not return a string, but the type itself, and right now we don't have machinery to manipulate types at runtime, so the only thing allowed with it should be to print it. It's might be tricky to actually define what the result type is. I don't know if it should be a string, we might regret that later.

krystophny:
> Instead of `type_name`, why not `typeof(x)` which is already an existing syntax in Fortran? We just need to teach `print *, typeof(x)` to print it, so print, or StringFormat, has to accept the result of `typeof(x)`. The `typeof(x)` is not an intrinsic, only part of a declaration, but making it an intrinsic would make sense. It should not return a string, but the type itself, and right now we don't have machinery to manipulate types at runtime, so the only thing allowed with it should be to print it. It's might be tricky to actually define what the result type is. I don't know if it should be a string, we might regret that later.

Hah, I did exactly the opposite to *avoid* using the intrinsic `typeof(x)`. But thinking about it, what you say would be a much more elegant solution that anyone will understand. We just have to introduce a type of a type. E.g. in Go language runtime representation of a type is `reflect.Type`. In Python the type of a type is `type` :) in Julia it is `DataType` and in C++ `std::type_info`.

certik:
Right. And I think the runtime type would be powerful to have anyway. And we already have it in some form, for example for printing user derived types. Finally, there might also be some applications of runtime types for the generics (cc @difference-scheme).

krystophny:
Now we add reflection also. So in the end our flavor of Fortran will indeed be the most powerful and elegant language ;-)

difference-scheme:
> We just have to introduce a type of a type.

This is exactly what we have to do! Yes, we need to introduce meta-types (to some extent) to take full advantage also of the generics. This is what I wanted to do next, after the Traits-proposal is being implemented. 

Before proceeding with this, please look up how this is implemented in Swift. In my opinion their design and syntax are great because it seamlessly matches with the generics (that we've already adopted from Swift)! 

It allows you to access (among other things) also the initializers of types, which will be absolutely crucial in OO programming, in order to write dependency injection containers by combining the generics with this basic reflection capability. This is something that is very needed to do proper (traits-based) OOP in large codes (where it's extremely tedious to do dependency injection manually for hundreds of classes) and which is impossible with the present language.

krystophny:
> > We just have to introduce a type of a type.
> 
> This is exactly what we have to do! Yes, we need to introduce meta-types (to some extent) to take full advantage also of the generics. This is what I wanted to do next, after the Traits-proposal is being implemented.
> 
> Before proceeding with this, please look up how this is implemented in Swift. In my opinion their design and syntax are great because it seamlessly matches with the generics (that we've already adopted from Swift)!
> 
> It allows you to access (among other things) also the initializers of types, which will be absolutely crucial in OO programming, in order to write dependency injection containers by combining the generics with this basic reflection capability. This is something that is very needed to do proper (traits-based) OOP in large codes (where its extremely tedious to do dependency injection manually for hundreds of classes) and which is impossible with the present language.

Exactly! I am tracking this now in https://github.com/krystophny/lfortran/pull/8 to avoid noise in the upstream repo. Let's discuss there further. Maybe I will also split the PR further and then when we are happy with it, submit piece by piece so @certik has a bit time to breathe.

PR #10237, opened 2026-02-21
CI: use the latest main of Julienne

certik:
This is now using the latest main, no workarounds needed. For now the main tests do not run yet, but it fully compiles and builds, which is what this PR is testing.


ci/test_third_party_codes.sh
certik:
```suggestion
  git clone https://github.com/BerkeleyLab/julienne.git
```

certik:
Already merged separately.

PR #10239, opened 2026-02-21
fix: handle allocatable & parameter arrays for string-concat

### LINKED ISSUE #7047: Allocatable character array segfault
Lfortran can't compile this program with an allocatable character array.
```
program allocationbug
  implicit none
  character(:),allocatable:: string(:)
  string = ['abc ','de  ']
  print "(A)",'"'//string(:)//'"'
end program allocationbug
```
Output from lfortran 0.51.0 with or without ```--std=f23```
```
```
Output from gfortran, ifx, or AMD flang
```
"abc "
"de  "
```

### LINKED ISSUE #9657: Concatenation error
This issue has the same program as #7047.  The lfortran version was then 0.51.0, which could not compile it. Version 0.59.0 now runs it but with wrong output.
```
program allocationbug
  implicit none
  character(:),allocatable:: string(:)
  string = ['abc ','de  ']
  print "(A)",'"'//string(:)//'"'
end program allocationbug
```
Lfortran 0.59.0 appears to ignore the second ```//``` in the print statemet:
```
"abc
"de 
```
Correct output from gfortran, ifx or AMD flang:
```
"abc "
"de  "
```
Note that the bug appears only if ```string``` is an array with at least 2 elements.

I wrote a new issue instead of a comment to #7047 because its headline does not apply to version 0.59.0 (and the program name is now misleading, but I could not change that while claiming it was the same program).

YashNagda17:
Fixes #7047 
Fixes #9657 

Changes made:
- Handle ReAlloc node for string-concat, which might require run-time calculation of length of strings, used in case of allocatable strings with DeferredLength
- Handle Compile-Time Length calculation for string-concat for Parameter string arrays.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1445,7 +1445,8 @@ RUN(NAME intrinsics_411 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # iachar 
 RUN(NAME intrinsics_412 LABELS gfortran llvm) # max_loc FAST
 RUN(NAME intrinsics_413 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # iachar(char(0))
 RUN(NAME intrinsics_414 LABELS gfortran llvm) # system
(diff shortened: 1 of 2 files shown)


certik:
Can you please resolve the conflicts?

Issue #4312, opened 2024-06-19
stacktrace_01 fails on Intel Mac


certik:
See https://github.com/lfortran/lfortran/pull/4276 for details.

Issue #307, opened 2022-08-15
Get dwarfdump based stacktrace working on linux


certik:
*Original issue*: https://gitlab.com/lfortran/lfortran/-/issues/289

After https://gitlab.com/lfortran/lfortran/-/merge_requests/714, we now use `llvm-dwarfdump` on macOS by default. Investigate if it can be used on Linux also, since it is part of LLVM which we depend on anyway, and it would eliminate the `binutils` dependency and simplify the code, as we could remove `binutils` support.

If it is possible, then we should probably move the `lines.txt` and `lines.dat` files into `share/lfortran/debug/`.

certik:
This will allow us to not depend on the BFD library which often has linking issues in practice.

certik:
Related: https://github.com/lfortran/lfortran/issues/2538.

PR #10258, opened 2026-02-21
fix: use exe path for debug map lookup to fix Ninja stacktraces

krystophny:
## Summary
Fix debug map file lookup in runtime stacktraces. Two functions changed in `lfortran_intrinsics.c`:

1. **`get_local_info_dwarfdump`**: use binary executable path instead of `source_filename`
2. **`get_local_address_mac`**: fix dangling pointer, memory leaks, and `source_filename` mismatch

## Why
The runtime used `source_filename` (the Fortran source path baked into IR) to find `_lines.dat.txt`, but the compiler generates that file from the output binary name (`-o` flag). These diverge in two cases:

1. **Ninja preprocessing** renames sources (e.g. `foo.f90` becomes `foo.f90-pp.f90`)
2. **Any `-o` name differing from the source** (e.g. `lfortran -g foo.f90 -o my_app`)

Additionally, `get_local_address_mac` had three pre-existing bugs:
- **Dangling pointer**: `realpath` wrote into a stack buffer, pointer stored in `d->binary_filename` became invalid after function return
- **Memory leaks**: two `get_base_name` calls per frame in `strcmp`, never freed
- **`source_filename` mismatch**: same exe!=source divergence as the Linux side

## Changes
- `get_local_info_dwarfdump`: resolve exe path from `d->binary_filename` / `binary_executable_path` via `realpath`, extract stem inline, append suffix
- `get_local_address_mac`: compare against dyld image 0 basename (always the main executable) instead of `source_filename`; properly free `get_base_name` results; remove dangling `realpath` into stack buffer

## Tests
- Existing unit test `runtime_stacktrace_01` (`run_with_dbg = true`) compiles with `-g --no-color`, runs, and checks stderr against reference with exact file names and line numbers - passes on both main and branch.
- Manual verification of the two bug scenarios (not covered by existing tests):

### Fails on main (Ninja-style preprocessed name)
```
$ lfortran -g trace_test.f90-pp.f90 -o trace_test && ./trace_test
trace_test.f90-pp_lines.dat.txt: No such file or directory
Error in opening the file!
```

### Fails on main (exe name != source name)
```
$ lfortran -g trace_test2.f90 -o my_app && ./my_app
trace_test2_lines.dat.txt: No such file or directory
Error in opening the file!
```

### Both pass on branch
```
$ lfortran -g trace_test.f90-pp.f90 -o trace_test && ./trace_test
  File "trace_test.f90-pp.f90", line 3
    error stop "deliberate crash"
ERROR STOP deliberate crash

$ lfortran -g trace_test2.f90 -o my_app && ./my_app
  File "trace_test2.f90", line 3
    error stop "deliberate crash"
ERROR STOP deliberate crash
```

### Integration tests
```
$ scripts/lf.sh itest -b llvm
100% tests passed, 0 tests failed out of 2489
```

## Related upstream issues
- Related to https://github.com/lfortran/lfortran/issues/307
- Related to https://github.com/lfortran/lfortran/issues/4312

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -8508,6 +8508,17 @@ int shared_lib_callback(struct dl_phdr_info *info,
 
 #ifdef HAVE_LFORTRAN_MACHO
 void get_local_address_mac(struct Stacktrace *d) {
+    // Use image 0 (always the main executable on macOS) instead of
+    // source_filename, which can diverge from the binary name.
+    char *main_base = NULL;
+    if (_dyld_image_count() > 0) {
+        main_base = get_base_name((char *)_dyld_get_image_name(0));
+    }
+    if (main_base == NULL) {
+        printf("The executable name could not be resolved for stacktrace.\n"
+            "Aborting...\n");
+        abort();
+    }
     for (uint32_t i = 0; i < _dyld_image_count(); i++) {
         const struct mach_header *header = _dyld_get_image_header(i);
         intptr_t offset = _dyld_get_image_vmaddr_slide(i);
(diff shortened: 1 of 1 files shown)


certik (approved):
I think that's fine. We can further iterate on this as needed.

krystophny:
Great! This unblocks replacement of the python scripts

PR #10261, opened 2026-02-22
fix: externalize inherited generic bindings 

certik:
Depends on https://github.com/lfortran/lfortran/pull/10260. The three commits are needed for the same bug fix.

PR #10262, opened 2026-02-22
fix: avoid submodule procedure decl segfault

certik:
Guard external symbol owner resolution in submodule procedure handling when resolving m_type_declaration, preventing a null dereference during semantic analysis.

Add integration test submodule_24 to cover module procedures with procedure(interface) dummy arguments implemented in a submodule.

Depends on https://github.com/lfortran/lfortran/pull/10261. Just the last commit is the contents of this PR.


certik:
Rebased in https://github.com/lfortran/lfortran/pull/10266.

Issue #5629, opened 2024-12-09
Code to compile: AoC-2024


certik:
How to get started: https://github.com/certik/AoC-2024/tree/lf1

This branch contains some initial workarounds, then:
```
(env) repos/AoC-2024(lf1)$ ./build.sh
+ FPM_FFLAGS=--cpp
+ fpm build --compiler=lfortran
aoc_utilities.F90                      done.
libAoC-2023.a                          done.
problem_08.f90                         failed.
[ 16%] Compiling...
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x100024433
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x100013a5f
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x10001cf63
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x10081628f
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x1024cf977
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x1000bebfb
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x1000bf207
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c115d7
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c11827
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c66497
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c665c7
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c667eb
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c6773f
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c67917
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c67ad7
  Binary file "/Users/ondrej/repos/lfortran/src/bin/lfortran", local address: 0x101c687b7
[...]

PR #10265, opened 2026-02-22
fix: handle int64 values in read iostat

### LINKED ISSUE #10264: Handle int64 values for iostat read statements
Related to #5629

MRE:
```fortran
program tmp
      use iso_fortran_env, only: int64
      implicit none

      type :: string_t
      end type

      type(string_t) :: val
      integer(int64) :: i1, istat

      val%str = "111"
      read(val%str, *, iostat=istat) i1  
      print*, i1, istat
      if (i1 /= 111) error stop
      if (istat /= 0) error stop

  end program
```

```
$ gfortran a.f90 & ./a.out
[1] 253181
                  111                    0

$ lfortran a.f90
  %istat = alloca i64, align 8
 i32*  call void @_lfortran_string_read_i64(i8* %10, i64 %12, i8* getelementptr inbounds ([4 x i8], [4 x i8]* @1, i32 0, i32 0), i64* %i1, i64* %istat)
```

YashNagda17:
Fixes #10264 
Towards #5629 

changes made:
- add llvm logic to handle int64 values inside iostat for read statements

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14244,12 +14244,14 @@ public:
                                 module.get())->getPointerTo();
                             var_to_read_into = llvm_utils->CreateLoad2(t, var_to_read_into);
                         }
-                        llvm::Type* i32ptr = llvm::Type::getInt32Ty(context)->getPointerTo();
                         llvm::Value* iostat_arg = iostat;
                         llvm::AllocaInst* tmp_iostat = nullptr;
-                        if (!llvm::isa<llvm::ConstantPointerNull>(iostat) && iostat->getType() != i32ptr) {
-                            tmp_iostat = builder->CreateAlloca(llvm::Type::getInt32Ty(context), nullptr, "tmp_iostat");
-                            iostat_arg = tmp_iostat;
+                        if (!llvm::isa<llvm::ConstantPointerNull>(iostat)) {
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14244,7 +14244,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                        builder->CreateCall(fn, { src_data, src_len, fmt, var_to_read_into, iostat });
+                        llvm::Value* iostat_arg = iostat;
+                        llvm::AllocaInst* tmp_iostat = nullptr;
+                        if (!llvm::isa<llvm::ConstantPointerNull>(iostat)) {
+                            llvm::AllocaInst* iostat_alloca = llvm::dyn_cast<llvm::AllocaInst>(iostat);
+                            if (iostat_alloca && !iostat_alloca->getAllocatedType()->isIntegerTy(32)) {
+                                tmp_iostat = builder->CreateAlloca(llvm::Type::getInt32Ty(context), nullptr, "tmp_iostat");
+                                iostat_arg = tmp_iostat;
+                            }
+                        }
```
certik, line 14255: Is there a way to rewrite this logic using ASR types, instead of querying it the LLVM IR?

YashNagda17: Ohk, looking into it...

Issue #6948, opened 2025-04-09
Intent(out) and optional forbidden in specification expressions


harperjf:
F2023 10.1.11 says in part

1 A specification expression is an expression with limitations that make it suitable for use in specifications such as length type parameters  and array bounds. A _specification-expr_ shall be a constant expression unless it is in an interface body (15.4.3.2), the specification part of a subprogram or BLOCK construct, a derived type definition, or the _declaration-type-spec_ of a FUNCTION statement (15.6.2.2).

R1029 _specification-expr_ **is** _scalar-int-expr_

C1011 (R1029) The _scalar-int-expr_ shall be a restricted expression.

2 A restricted expression is an expression in which each operation is intrinsic or defined by a specification function and each primary is

(1) a constant or subobject of a constant,
(2) an object designator with a base object that is a dummy argument that has neither the OPTIONAL nor the INTENT (OUT) attribute,

The last requirement quoted above should have prevented the program below from being compiled, but it ran with lfortran.
```
module foo ! in file blockproblem.f90
  type :: bar
    integer :: n
  end type

contains
  subroutine init(this, n)
    type(bar), intent(out) :: this
    integer, intent(in) :: n
    this%n = n
    block
      real :: array(this%n) ! DISALLOWED
      integer :: i
      array = [ (i,i = 1,this%n) ]
      print "(A,*(1X,F0.3))",'array =',array
    end block
  end subroutine

end module foo

program blockproblem
  use foo
  type(bar):: this
  call init(this,4)
  print "(A,*(1X,I0))", 'this%n =',this%n
end program blockproblem
```
Illegal output from lfortran 0.50.0 or 0.51.0
```
array = 1.000 2.000 3.000 4.000
this%n = 4
```

harperjf:
Recent discussion in Fortran Discourse casts doubt on the illegality of my program with a block construct. Replacing the block construct by an internal subroutine turns the program into
``` 
! file subproblem.f90
module foo 
  type :: bar
    integer :: n
  end type

contains
  subroutine init(this, n)
    type(bar), intent(out) :: this
    integer, intent(in) :: n
    this%n = n
    call blocksub
  contains
    subroutine blocksub ! replacing block in blockproblem.f90 
      real :: array(this%n) ! bad if intent(out), good if host-associated
      integer :: i
      array = [ (i,i = 1,this%n) ]
      print "(A,*(1X,F0.3))",'array =',array
    end subroutine blocksub
  end subroutine init

end module foo

[...]

PR #10267, opened 2026-02-22
throw error when INTENT(OUT) is used with specification expr

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6948

### DIFF
--- a/tests/errors/continue_compilation_2.f90
+++ b/tests/errors/continue_compilation_2.f90
@@ -82,44 +82,44 @@ module continue_compilation_2_mod
         integer :: x(.., 5)
     end subroutine arank
 
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
     
 end module continue_compilation_2_mod
+
+module intent_out_array_bounds_mod
+    type :: bar
+        integer :: n
+    end type
+contains
+    subroutine init_bar(this, n)
+        type(bar), intent(out) :: this
+        integer, intent(in) :: n
+        this%n = n
+        block
+            real :: array(this%n)
+            integer :: i
+            array = [ (i,i = 1,this%n) ]
+            print *, array
+        end block
+    end subroutine
+
+    subroutine optional_bound(n)
+        integer, optional, intent(in) :: n
+        real :: arr(n)
+        arr = 0.0
+        print *, arr
+    end subroutine
+end module intent_out_array_bounds_mod
+
+
+
+
+
+
+
+
+
+
+
 
 
 
--- a/tests/reference/asr-continue_compilation_2-a6145a1.stderr
+++ b/tests/reference/asr-continue_compilation_2-a6145a1.stderr
@@ -137,6 +137,12 @@ semantic error: Dummy argument 'x' not defined
 83 |        end subroutine arank
    | ...^^^^^^^^^^^^^^^^^^^^^^^^ 
 
+semantic error: OPTIONAL dummy argument 'n' cannot be used for array bounds declaration
+   --> tests/errors/continue_compilation_2.f90:107:21
+    |
+107 |         real :: arr(n)
+    |                     ^ hint: use a non-optional argument, or declare the array as `allocatable`
+
 semantic error: Implicit typing is not allowed, enable it by using --implicit-typing 
    --> tests/errors/continue_compilation_2.f90:132:5 - 134:49
     |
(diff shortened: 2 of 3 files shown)


tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -979,3 +979,9 @@ warning: Format string missing comma between descriptors (F2023 constraint C1302
     |
 474 |     print "(aai6)", a,"hi",15
     |           ^^^^^^^^ 
+
+semantic error: Dummy argument 'this' with INTENT(OUT) cannot be used in a specification expression
```
certik: Can you explain what this error message means? I don't understand it. I had to ask Grok:

Why INTENT(OUT) is forbidden
An INTENT(OUT) dummy argument is undefined on entry to the procedure (see 15.5.2.5 Argument association and 8.5.10 INTENT attribute). The whole point of INTENT(OUT) is that the procedure defines the argument; the caller does not supply a meaningful value. Reading it (even indirectly through %n) before any executable code has run would be undefined behaviour, so the standard simply prohibits it.
This is a semantic (constraint-violation) error, not a syntax error.

------

It proposed the following improvement:
```
semantic error: Dummy argument 'this' has INTENT(OUT) and cannot be referenced in a specification expression
   --> tests/errors/continue_compilation_2.f90:500:21
    |
500 | real :: array(this%n)
    |         ^^^^^^^
Note: Specification expressions (used for array bounds, character lengths, etc.) are evaluated on entry to the procedure. An INTENT(OUT) dummy argument is undefined at that point (ISO/IEC 1539-1:2023, 10.1.11 and constraint C1011).

Hint: 
  • Use INTENT(IN) or INTENT(INOUT) if you need to read 'this%n' in the declaration, or
  • Declare 'array' as ALLOCATABLE and ALLOCATE it after 'this' has been defined, or
  • Pass the size as a separate INTENT(IN) dummy argument.
```

Do we allow to have a longer Note like above?. Let's add a hint next to the underline with something like:
```
hint: use `intent(in/inout) / allocatable` for `this`, or pass the size as a separate `intent(in)` argument
```
Also instead of the first line, maybe something like:
```
semantic error: Dummy argument 'this' has `intent(out)` and cannot be used for array bounds declaration
```

I don't like the term "specification expression", because it is not clear to somebody who doesn't know the standard what this means. So it's not a good term. We should try each error message to be so good that somebody without knowing anything about the standard or Fortran could understand what the problem is.

The "specification expression" can be used at many places (I think). But rather, we know exactly what the problem is in the compiler --- so in this case it's not just some "specification expression", but it's an array bound (declaration) and nothing else. So let's be very specific and use simple direct language to explain what the problem is. And sometimes we might need to do some more investigation in the compiler to figure out exactly what specific case we have at hand.

certik:
I think this PR is simple, so let's get the error message right first.

tests/errors/continue_compilation_2.f90
```diff
@@ -485,3 +485,23 @@ program second_program_test
...
+    this%n = n
+    block
+      real :: array(this%n)
+      integer :: i
+      array = [ (i,i = 1,this%n) ]
+      print *, array
+    end block
+  end subroutine
+
+end module spec_expr_intent_out_01_mod
```
certik: I think in proper Fortran, the program goes last, and anything after it can be ignored.

Instead of adding a new module per error, is there no way to integrate this error into our existing file?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2400,7 +2400,57 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                if (var->m_presence == ASR::presenceType::Optional) {
+                    diag.add(Diagnostic(
+                        "OPTIONAL dummy argument '" +
+                        std::string(var->m_name) +
+                        "' cannot be used in a specification expression",
+                        Level::Error, Stage::Semantic, {
+                            Label("", {expr->base.loc})
+                        }));
+                    throw SemanticAbort();
+                }
```
certik, line 2441: I think this is not tested.

tests/errors/continue_compilation_2.f90
HarshitaKalani: Let me try thi.

HarshitaKalani:
I need to resolve conflics to so I'll make the changes in this PR itself.

tests/errors/continue_compilation_2.f90
```diff
@@ -83,7 +86,60 @@ subroutine arank(x)
...
+
+
+
+
+
+
+
+
+
+
```
HarshitaKalani, line 125: I've intentionally added extra lines so that next time we need to add something, there're not a lot of updates in the references.

certik:
Can you resolve the conflicts one more time please?

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -80,914 +80,263 @@ semantic error: Global name is already being used
...
-    |              ^^^^^^^ 
-
-semantic error: No custom operator exists for this expression
-   --> tests/errors/continue_compilation_2.f90:235:15
-    |
-235 |     str_t_2 = prefix // 10
-    |               ^^^^^^^^^^^^ 
-
-semantic error: Type mismatch in assignment, the types must be compatible
-   --> tests/errors/continue_compilation_2.f90:235:5
```
certik: I think this is removing a bunch of warnings and errors. 

Let's do this the slow way: submit a PR that extends with new lines.

Then rebase this PR on top.

tests/reference/asr-continue_compilation_2-a6145a1.stderr
```diff
@@ -824,12 +842,6 @@ semantic error: Type mismatch in assignment, the types must be compatible
 490 |     tm2_x = 5 + "x"
     |     ^^^^^       ^^^ type mismatch (integer and string)
 
-semantic error: Variable 'foo' is not declared
-   --> tests/errors/continue_compilation_2.f90:492:13
-    |
-492 |     print *,foo
-    |             ^^^ 'foo' is undeclared
-
```
HarshitaKalani, line 832: This is an improvement. This error was a false positive before. 
There's this line - `integer, parameter, target :: foo=4`
This declaration fails because parameter conflicts with target (that error is still reported). Previously, because the declaration threw SemanticAbort in the middle of processing, foo was never added to the symbol table at all. So when print *,foo was encountered at line 492, the compiler also reported a cascading "undeclared" error.

HarshitaKalani:
I think we can merge this now

Issue #10270, opened 2026-02-23
Run-time hang with --std=f23 but semantic error without it


harperjf:
With this program 
```
  program snooze
    implicit none
    call sleep(3)
    call execute_command_line('date')
  end program snooze
```
lfortran 0.60.0 with no options in a Linux Ubuntu system gave the correct error message
  ```
  semantic error: Function 'sleep' not found (not user defined nor intrinsic)
 --> snooze.f90:3:5
  |
8 |      call sleep(2)
  |      ^^^^^^^^^^^^^ 
```
But with the bash command ```date``` before and after lfortran --std=f23 and on stopping the program manually after nearly a minute I got
```
(lf) john:~/Gstd23/later$ date; lfortran --std=f23  snooze.f90 ; date
Mon Feb 23 11:34:18 AM NZDT 2026
^CMon Feb 23 11:35:10 AM NZDT 2026
```
Gfortran has an intrinsic subroutine ```sleep()``` as an extension to the Fortran standard and  the program's result was then, as expected,
```
(lf) john:~/Gstd23/later$ date; gfortran snooze.f90 && ./a.out ; date
Mon Feb 23 11:35:40 AM NZDT 2026
Mon Feb 23 11:35:43 AM NZDT 2026
```
Of course, if a sleep() subroutine were added to Lfortran, calling it should cause a semantic error with --std=f23 and run like the gfortran one without it.

PR #10272, opened 2026-02-23
Add sleep() as an extension intrinsic subroutine

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/10270

### DIFF
--- a/tests/reference/asr-continue_compilation_2-a6145a1.json
+++ b/tests/reference/asr-continue_compilation_2-a6145a1.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_2-a6145a1",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_2.f90",
-    "infile_hash": "c11fa263de85c97a20f26a765475828f5be96466f2c20bb3cec20889",
+    "infile_hash": "57ead231c8d005eae8bf6ea2fcebb140073d6e7121dd6b69cbd558d7",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_2-a6145a1.stderr",
-    "stderr_hash": "2fba8021c871520a6e05b054dd5c19ea268a4dd1497d60dbf180b077",
+    "stderr_hash": "01a72f674adfca6ebce4e47650806081410ad46cf4b1b52692d82490",
     "returncode": 1
 }


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5464,6 +5464,21 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        extension_subroutines.count(var_name)) {
+                    diag.add(Diagnostic(
+                        "`" + var_name + "` is a non-standard extension; not"
+                        " available with `--std=f23`",
+                        Level::Error, Stage::Semantic, {
+                            Label("remove the `--std=f23` flag to enable"
+                                  " non-standard extensions",
+                                  {x.base.base.loc})
+                        }));
+                    throw SemanticAbort();
```
certik: Currently I think we allow extensions with f23. I would make it a warning --- the error might be really annoying for codes that you want to make work with f23, but maybe use a few extensions.

We can add another option to be "strict", which changes these warnings into errors.

certik: In fact that is why our current tests fail:
```
semantic error: `abort` is a non-standard extension; not available with `--std=f23`
 --> /home/runner/work/lfortran/lfortran/integration_tests/abort_01.f90:3:16
  |
3 |   if (.false.) call abort()
  |                ^^^^^^^^^^^^ remove the `--std=f23` flag to enable non-standard extensions


```


Just make it a warning in this PR.

HarshitaKalani:
Yes, the CI also fails because of this error in abort.
```
semantic error: `abort` is a non-standard extension; not available with `--std=f23`
 --> /home/runner/work/lfortran/lfortran/integration_tests/abort_01.f90:3:16
[  0%] Linking Fortran executable internal_subr_host_vars_01_STD_F23
  |
3 |   if (.false.) call abort()
  |                ^^^^^^^^^^^^ remove the `--std=f23` flag to enable non-standard extensions


```

src/lfortran/semantics/ast_body_visitor.cpp
HarshitaKalani: Alright. I'll do that.

integration_tests/intrinsics_416.f90
```diff
@@ -37,4 +37,4 @@ program intrinsics_416
   write(str6, "(*(A))") '"'//str4(:)//str5//'"'
   if (str6 /= '"abc xy""de  xy"') error stop
 
-end program intrinsics_416
\ No newline at end of file
+end program intrinsics_416
```
certik: Don't put unrelated whitespace changes into a PR. Leave it unless you need to make more changes, then you can fix it.

integration_tests/intrinsics_417.f90
```diff
@@ -0,0 +1,8 @@
+program intrinsics_417
+    implicit none
+    integer :: t1, t2, rate
+    call system_clock(t1, rate)
+    call sleep(1)
+    call system_clock(t2)
+    if ((t2 - t1) < rate / 2) error stop
```
certik: What does this condition do? What does `rate` do?

integration_tests/intrinsics_417.f90
```diff
@@ -0,0 +1,8 @@
+program intrinsics_417
+    implicit none
+    integer :: t1, t2, rate
+    call system_clock(t1, rate)
+    call sleep(1)
```
certik: This will now be the longest running test by far. Each test should take 0.00s, so less than 10ms. Does `sleep` only accept integers? If so, then you'll have to call it with 0 here.

tests/tests.toml
```diff
@@ -4540,6 +4540,11 @@ asr = true
 filename = "warnings/doloop_01.f90"
 asr = true
 
+[[test]]
+filename = "warnings/sleep_01.f90"
+asr = true
+options = "--std=f23"
```
certik: I think this should always be a warning, no? Every extension should give a warning at all times.

HarshitaKalani: I thought this could be helpful when we introduce errors for extensions with --std=f23. But if we're aiming to introduce a `strict` flag, then I think I can remove this. I'll do that.

integration_tests/intrinsics_417.f90
HarshitaKalani: Yes, it takes integer for seconds. Let me make it 0.

integration_tests/intrinsics_417.f90
HarshitaKalani: It checks that at least ~0.5 seconds elapsed between t1 and t2 (since rate is the clock ticks per second returned by system_clock, so rate/2 ≈ half a second in ticks).
I'll use the following version now:
```
call system_clock(t1, rate)
call sleep(0)
call system_clock(t2)
if (t2 < t1) error stop
```

integration_tests/intrinsics_417.f90
```diff
@@ -0,0 +1,8 @@
+program intrinsics_417
+    implicit none
+    integer :: t1, t2, rate
+    call system_clock(t1, rate)
+    call sleep(0)
+    call system_clock(t2)
+    if (t2 < t1) error stop
```
certik, line 7: Let's merge this before merging into main:
```suggestion
    ! if (t2 < t1) error stop
```
If things run fast, then it could happen that `t1 == t2` and we'll get a flaky bug at the CI.

certik: Wait for the CI to finish, then ping me. I'll merge the above change and we'll merge into main right after that.

certik: Oh, I guess you allow both `t1 == t2` and `t1 < t2`, so the test is fine. Ignore my comment.

HarshitaKalani: Yes

HarshitaKalani:
I think we can merge this now. @certik

Issue #10275, opened 2026-02-23
Separate compilation fails on type-bound procedure that uses array slices for a method call


difference-scheme:
`$ cat mytypemod.f90 `
```
module mytypemod
   
   type :: MyType
   contains
      procedure :: method1
      procedure :: method2
   end type MyType

contains

   subroutine method1(self)
      class(MyType), intent(inout) :: self      
      real(8), dimension(140) :: a
      real(8), dimension(140,4) :: b      
      call self%method2(a(1:140),b(1:140,:))
   end subroutine method1


   subroutine method2(self,a,b)
      class(MyType), intent(in)    :: self
      real(8),       intent(in)    :: a(:)
      real(8),       intent(inout) :: b(:,:)
   end subroutine method2
   
end module mytypemod

```
 `lfortran -c --separate-compilation mytypemod.f90 `
```
; ModuleID = 'LFortran'
source_filename = "LFortran"

%dimension_descriptor = type { i32, i32, i32 }
%array = type { ptr, i32, ptr, i1, i32 }
%mytype_class = type <{ ptr, ptr }>

@_Name_mytype = private unnamed_addr constant [7 x i8] c"mytype\00", align 1
@_Type_Info_mytype = linkonce_odr unnamed_addr constant { ptr, ptr, ptr } { ptr @_Name_mytype, ptr null, ptr null }, align 8
@_VTable_mytype = linkonce_odr unnamed_addr constant { [6 x ptr] } { [6 x ptr] [ptr null, ptr @_Type_Info_mytype, ptr @_copy_mytypemod_mytype, ptr @_allocate_struct_mytypemod_mytype, ptr @__module_mytypemod_method1, ptr @__module_mytypemod_method2_real____1_real_______2] }, align 8

define void @__module_mytypemod_method1(ptr %self) {
.entry:
  %0 = alloca %dimension_descriptor, i32 2, align 8
  %array_section_descriptor2 = alloca %array, align 8
  %1 = alloca %dimension_descriptor, align 8
  %array_section_descriptor = alloca %array, align 8
  %__libasr_created__subroutine_call_1_mytype_method21 = alloca ptr, align 8
  store ptr null, ptr %__libasr_created__subroutine_call_1_mytype_method21, align 8
  %arr_desc = alloca %array, align 8
  %2 = getelementptr %array, ptr %arr_desc, i32 0, i32 2
  %3 = alloca i32, align 4
  store i32 2, ptr %3, align 4
[...]

PR #10278, opened 2026-02-23
Fix vtable call argument mismatch

certik:
When calling a type-bound procedure with array section arguments through
virtual dispatch (class(T)), the pass_array_by_data pass would update
the StructMethodDeclaration to point to the lowered function (with
unpacked array args) but skip expanding the call arguments. This
happened because can_edit_call() returned false for Pointer-typed
temporaries created by earlier passes, and the early return was taken
since m_module was false.

The fix skips the early return for StructMethodDeclaration calls,
ensuring call arguments are expanded to match the lowered function
signature stored in the vtable.

Fixes https://github.com/lfortran/lfortran/issues/10275.

Depends on #10273.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2457,6 +2457,7 @@ RUN(NAME class_117 LABELS gfortran llvm)
 RUN(NAME class_118 LABELS gfortran llvm)
 RUN(NAME class_119 LABELS gfortran llvm)
 RUN(NAME class_120 LABELS gfortran llvm)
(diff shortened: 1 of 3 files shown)


certik:
@jinangshah21 can you please review just the last commit 7f83931b798d2a9c325117b394ddfca84ee86762 that it is the correct fix, consistent with your comment in https://github.com/lfortran/lfortran/issues/10275#issuecomment-3946875660?

jinangshah21:
Yes, I think we can't transform struct-methods for now

PR #10281, opened 2026-02-24
Fix ICE on inherited generic binding

certik:
When a child type has a generic binding that resolves to a non_overridable procedure inherited from a parent type, the symbol lookup in the child's symtab returns an ExternalSymbol pointing to the parent's StructMethodDeclaration. The codegen was missing symbol_get_past_external() before downcasting, causing an assertion failure. Apply the resolution in both
visit_RuntimePolymorphicSubroutineCall and
visit_RuntimePolymorphicFunctionCall.


certik:
This is a quick fix, but a possibly better fix is in https://github.com/lfortran/lfortran/pull/10282.

PR #10282, opened 2026-02-24
Avoid using ExternalSymbol inside a module

certik:
An alternative fix for https://github.com/lfortran/lfortran/pull/10281, probably better.


certik:
@jinangshah21 what do you think of the fix in 3b802fae39a40b3d813bfdf399f4e81c8f1301e7? I think we should not be using ExternalSymbol within the same module, only when the symbol is truly external.

certik:
We should also improve our verify to enforce that ExternalSymbol always points to an "external symbol" in another module, never within the same module.

PR #10296, opened 2026-02-24
CI: test Formal

certik:


integration_tests/cpp_pre_07.f90
```diff
@@ -0,0 +1,35 @@
...
+x = 1
+#else
+x = 2
+#endif
+
+if (x /= 2) error stop
+
+#define DEFINED_MACRO
+
+#ifdef DEFINED_MACRO && OTHER
```
certik, line 18: It's the correct fix to ignore `&& OTHER`. However given how common mistake this is, we should actually give a warning that we are ignoring the extra characters.

Issue #5496, opened 2024-11-26
dwarf_convert.py cannot parse the new format


certik:
It emits empty output.

certik:
With #5497, the stacktrace almost works:
```console
$ lfortran -c tests/string1.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2686
    LCompilers::print_stack_on_segfault();
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2583
    return compile_src_to_object_file(arg_file, outfile, false,
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 1003
    bool has_error_w_cc = compiler_options.continue_compilation && diagnostics.has_error();
  File "/Users/ondrej/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 275
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics);
  File "/Users/ondrej/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 297
    compiler_options.symtab_only, compiler_options);
  File "/Users/ondrej/repos/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 72
    if (res.ok) {
  File "/Users/ondrej/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 3835
    v.visit_TranslationUnit(ast);
  File "/Users/ondrej/repos/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 132
    visit_ast(*x.m_items[i]);
  File "/Users/ondrej/repos/lfortran/src/lfortran/ast.h", line 4884
    void visit_ast(const ast_t &b) { visit_ast_t(b, self()); }
  File "/Users/ondrej/repos/lfortran/src/lfortran/ast.h", line 4840
    case astType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/ondrej/repos/lfortran/src/lfortran/ast.h", line 4887
    void visit_mod(const mod_t &b) { visit_mod_t(b, self()); }
  File "/Users/ondrej/repos/lfortran/src/lfortran/ast.h", line 4471
    case modType::Module: { v.visit_Module((const Module_t &)x); return; }
[...]

PR #10302, opened 2026-02-25
stacktrace: replace python+dwarfdump debug-map pipeline with in-process DWARF reader

### LINKED ISSUE #307: Get dwarfdump based stacktrace working on linux

### LINKED ISSUE #2538: Runtime stacktrace: do not depend on external tools
Currently it seems we require at least two external tools:
* `llvm-dwarfdump`
* `./dwarf_convert.py`
We should implement both as part of LFortran itself, so that we do not have dependencies that can fail. If not practical to implement, we need to ensure the tool is always in PATH, one way or the other. Probably as part of the conda package binary install, so that `conda install lfortran` just works for runtime stacktraces on all platforms.

krystophny:
## Summary
- Remove the Python debug-map pipeline (`dwarf_convert.py`, `dat_convert.py`) and external `llvm-dwarfdump` shelling
- Generate `*_lines.dat` directly in-process from DWARF using LLVM APIs at compile/link time
- Runtime reads packed binary `*_lines.dat` directly (no text parsing bridge)
- Fix out-of-bounds read in `bisection()` that caused garbled line numbers

Fixes #307
Fixes #2538
Related to #5496

## Why
The old flow depended on external tools and fragile text parsing:

```
compile -> llvm-dwarfdump output -> Python parse -> text -> binary -> text for runtime
```

This collapses that into one in-process DWARF read and one binary map format consumed directly by the runtime.

**Stage:** Codegen + Runtime

## Changes
- [`src/libasr/stacktrace.cpp#L675-L782`](https://github.com/krystophny/lfortran/blob/30efa23614c64b61519ff189f7d48c397b52ea64/src/libasr/stacktrace.cpp#L675-L782): new `write_runtime_debug_map()` reads DWARF line tables via LLVM APIs; LLVM 8/9 vs 10+ address-field compatibility
- [`src/bin/lfortran.cpp#L2052-L2080`](https://github.com/krystophny/lfortran/blob/30efa23614c64b61519ff189f7d48c397b52ea64/src/bin/lfortran.cpp#L2052-L2080): call runtime map builder from linker path; macOS `dsymutil` + dSYM DWARF source selection
- [`src/libasr/runtime/lfortran_intrinsics.c`](https://github.com/krystophny/lfortran/blob/30efa23614c64b61519ff189f7d48c397b52ea64/src/libasr/runtime/lfortran_intrinsics.c): consume packed binary map directly via `fread()`; fix `bisection()` off-by-one (`return size` -> `return size-1`)
- Removed: `dwarf_convert.py`, `dat_convert.py`, `get_dwarf_scripts_dir()`, `WITH_DWARFDUMP` CMake option, related install/discovery/plumbing

### Test passes on branch (LLVM 11)
```
$ scripts/lf.sh test -t runtime_stacktrace_01 -s
errors/runtime_stacktrace_01.f90 * run_dbg ✓
TESTS PASSED
```

### Test passes on branch (LLVM 21)
```
$ scripts/lf.sh --llvm21 test -t runtime_stacktrace_01 -s
errors/runtime_stacktrace_01.f90 * run_dbg ✓
TESTS PASSED
```


krystophny:
Btw liric offers also this LLVM DWARF API via a thin layer. I used the LLVM C++ API for now until we decide if or how one would replace all by a C one directly.

certik:
This is great, thanks for implementing it. This has been bothering me for a long time. Need to review by hand locally.

HarshitaKalani (approved):
I think this is great. Thanks @krystophny !

certik:
Thanks for updating it. Let me test it.

Issue #8878, opened 2025-11-19
Finalization TODO


assem2002:
After merging this #8871, We'll be able to say that we have a good finalization process that's working properly (for fortran types).
Here's a full todo to perfect it.

- [x] Add a utility to print out the **amount of allocated VS Freed heap memory** and what is the purpose of allocating that memory -- To find leaks and investigate memory usages.
- [x] **Use Functions to finalize types** -- (e.g. `finalize_alloatable_scalar`, `finalize_derivedType_tt` , `finalize_descriptorArray`). This will make us have smaller binaries + will resolves the issue of recursive derived types declaration ( reference `integration_tests/derived_types_78.f90`).
- [ ] **Handle the case of block symbol having a `return` statement**. We need to finalize any sub-constructs in a super-construct if we use `return` in them (reference `integration_tests/finalization_01.f90`).
- [x] **Remove unnecessary labels** #8844.
- [x] **Add Finalization functions in the VTable** of the types -- To handle the case of unlimited polymorphic type in Fortran `class(*)` -- remember that it could be an array.

assem2002:
I think we also need to use the same utilities of finalization with `deallocate` statement, as it for sure leaks now. 

As we'll end up having LLVM-finalization function per type, things should get easier.

PR #10304, opened 2026-02-25
Cache finalize instructions as functions

assem2002:
towards #8878
# FPM Benchmark (Is it actually 30% less compile time?):
- Without caching
```
LLVM -> BIN                                        34706.336
------------------------------------------------------------
Total time                                         46234.434
------------------------------------------------------------
```
- With Caching 
```
LLVM -> BIN                                        21773.760
------------------------------------------------------------
Total time                                         32397.344
------------------------------------------------------------
```
***
## Why this PR ?
- Finalization (in main) is an inline process -- meaning it inserts instructions for each and every variable when exiting the scope (LLVM function)
-  With big codes, We face lots of variables with exact same type. If we insert inline instructions for each variable redundantly, that means we're creating bigger and bigger binary for no reason.
## How to fix ?
- Let's cache the types we know are producing very redundant instructions like (arrays, structs, etc.).
- The cache is basically functions that you call.
## Approach
- All changes are done in `llvm_utils.h` in `LLVMFinalize` class.
- We use 2 functions `START_CACHE`, `END_CACHE`.
- The goal is to start a caching mode (basically just a function body)
- `START_CACHE` : takes a cache key (unique key per type or set of instruction) + variadic arguments representing the parameters of the function. After that, we just build the function and direct the passed `LLVM::Value` variadic argument to the function parameter. It returns `BasicBlock` to the current insert basicblock to be used later by `END_CACHE` to revert back.
- `END_CACHE` : After finishing with intended instruction that needed cache, We just revert back to the basic block returned by `START_CACHE`.

### DIFF
--- a/src/libasr/codegen/llvm_utils.h
+++ b/src/libasr/codegen/llvm_utils.h
@@ -1515,7 +1515,7 @@ if(get_struct_sym(member_variable) == struct_sym /*recursive declaration*/){cont
             for(auto &fn_arg : finalizer_fn->args()) {
                 *(args[i++]) = &fn_arg;
             }
-            LCOMPILERS_ASSERT(i == args.size())
+            LCOMPILERS_ASSERT(i == (int)args.size())
             }
 
             return saved_BB;


assem2002:
Merge squash is needed.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1368,11 +1433,111 @@ if(get_struct_sym(member_variable) == struct_sym /*recursive declaration*/){cont
...
+            return key; 
+        }
+
+        bool is_cached(const std::string& cache_key) {
+            return type_finalizer_cache_.find(cache_key) != type_finalizer_cache_.end();
+        }
+
+        /**
+         * Starts CACHE mode for the following set of instructions until `END_CACHE` is called.
+         * In CACHE mode, we create a finalizer function for the type we're finalizing,
```
certik, line 1473: Let's explain this better. It looks like this is doing the following:

* Instead of inserting LLVM instructions to finalize each type every time, we now instead call a new LLVM function and reuse this function for variables of the same type, greatly reducing the amount of LLVM instructions that we have to emit.
* By "caching" we mean that we "reuse" the same instructions the second time we need them, by putting them into a function and calling them.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1368,11 +1433,111 @@ if(get_struct_sym(member_variable) == struct_sym /*recursive declaration*/){cont
...
+                key += ASRUtils::get_type_code(t_past, false, false, false) +"__" + struct_sym->m_name;
+                if(auto module = ASRUtils::get_sym_module(&struct_sym->base)) {
+                    key += "_of_"; 
+                    key += module->m_name;
+                }
+            } else {
+                key += ASRUtils::get_type_code(t_past, false, false, false);
+            }
+            return key; 
+        }
```
certik, line 1465: There might be a more efficient way to do this than using strings. Can be fixed later.

PR #10332, opened 2026-02-27
Fix procedure pointer dummy arguments

### LINKED ISSUE #10325: Segmentation fault upon trying to assign a value to a procedure pointer
`$ cat test.f90 `
```
program test

   abstract interface
      function intfc() result(res)
         class(*), allocatable :: res
      end function intfc
   end interface

   procedure(intfc), pointer :: ptr
   
   call set(ptr)
   
contains
   
   subroutine set(ptr)
      procedure(intfc), pointer :: ptr
      ptr => null()
   end subroutine set

end program test
```
`$ lfortran test.f90 `
```
```
The above assignment actually works if in `subroutine set` the declaration of `ptr` is changed as follows:
```
procedure(intfc), pointer, intent(inout) :: ptr
```
Both versions should work, though.

certik:
Wrap procedure pointer variable types in Pointer(FunctionType) instead of bare FunctionType in the ASR, so that procedure pointer dummies are properly distinguished from non-pointer procedure dummies.

Changes:
- Semantic analysis: wrap FunctionType in Pointer() when is_pointer is set
- Codegen: handle Pointer(FunctionType) in type conversion, argument passing, fetch_var, and convert_call_args
- ASR passes: preserve Pointer wrapper when updating procedure variable types in subroutine_from_function, pass_array_by_data, transform_optional_argument_functions, and ast_body_visitor
- ASR utilities: update is_symbol_procedure_variable and get_FunctionType to look through Pointer wrapper
- ASR verifier: accept Pointer(FunctionType) for procedure variable checks

Fixes #10325.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2857,7 +2857,7 @@ RUN(NAME procedure_27 LABELS gfortran llvm)
 RUN(NAME procedure_28 LABELS gfortran llvm)
 RUN(NAME procedure_29 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
-RUN(NAME procedure_30 LABELS gfortran llvm)
+RUN(NAME procedure_30 LABELS llvm)
 RUN(NAME procedure_31 LABELS gfortran llvm)
 
 


certik:
Some third-party codes fail, we'll have to investigate.

PR #10305, opened 2026-02-25
enh: handling of nested submodules

Pranavchiku:
built on top of #10300.

PR #10335, opened 2026-02-27
fix: support module procedure implementation in nested submodules

Pranavchiku:
Part of #10305

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -871,7 +871,8 @@ public:
         switch(str_type->m_len_kind){
             case ASR::ExpressionLength:{
                 LCOMPILERS_ASSERT(len);
-                int ptr_load_cpy = ptr_loads;ptr_loads = 1;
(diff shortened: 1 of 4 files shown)


Pranavchiku:
The failure is due to failure in main.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5065,7 +5066,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        ASR::Module_t* parent_mod = ASR::down_cast<ASR::Module_t>(parent_sym);
+                        if (parent_mod->m_parent_module) {
+                            root_module = std::string(parent_mod->m_parent_module);
+                            parent_sym = tu_symtab->get_symbol(root_module);
+                        } else {
+                            break;
+                        }
+                    }
+                }
+                mangle_prefix = "__module_" + root_module + "_";
```
certik, line 5083: Logic like this can be refactored into a function `get_root_module`

PR #10350, opened 2026-02-28
fix: reject character length that depends on user-defined function in main program

### LINKED ISSUE #6344: Bad specification expression in main program
A user-defined pure function is not allowed in a specification expression in a main program (F2023 standard 10.1.11 paragraph 1) but lfortran did in this program:
```
MODULE halvestring ! Can half be in a main program's specification expression?
CONTAINS           ! gfortran bug report 32386 Jun 2007 
  PURE FUNCTION half(s1) RESULT (s2)
    CHARACTER(*),INTENT(IN):: s1
    CHARACTER(len(s1)/2) :: s2
    s2 = s1
  END FUNCTION half
END MODULE halvestring
! -----------
PROGRAM testspec
USE  halvestring, only: half
CHARACTER(*),PARAMETER :: stringa = 'abcdeedcba'
CHARACTER(len(half(stringa))) :: stringb

stringb = half(stringa)
PRINT "(A)", '"'//stringb//'"' !   5 "abcde"
END PROGRAM testspec
```
Good gfortran output:
```
gf32386.f90:13:40:

   13 | CHARACTER(len(half(stringa))) :: stringb
      |                                        1
Error: ‘stringb’ at (1) must have constant character length in this context
```
Bad lfortran 0.45.0 --std=f23 output:
```
"abcde"
```

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6344


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -2403,6 +2403,65 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            case ASR::exprType::IntegerUnaryMinus:
+                return find_user_defined_func_call_name(
+                    ASR::down_cast<ASR::IntegerUnaryMinus_t>(expr)->m_arg);
+            case ASR::exprType::RealUnaryMinus:
+                return find_user_defined_func_call_name(
+                    ASR::down_cast<ASR::RealUnaryMinus_t>(expr)->m_arg);
+            default:
+                return "";
+        }
+    }
```
certik: This will silently return `""` for nodes that are not supported --- that seems fragile. Is it like a best effort to try to find a user function and if we can't, then we'll silently skip the error message? I would rather put an ICE that we don't support some node; but even better let's implement this correctly for all nodes. And it might be that this needs to be a visitor, to keep the code small. But either way, let's implement the full solution, not a half-done solution that is not clear that it will lead to full solution when it is fully implemented for all cases.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -15821,6 +15893,19 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                diag.add(Diagnostic(
+                                    "Character length in the main program must be "
+                                    "a compile-time constant, but it depends on a "
+                                    "call to the function '" + func_name + "'",
+                                    Level::Error, Stage::Semantic, {
+                                        Label("", {len_expr->base.loc})
+                                    }));
+                                throw SemanticAbort();
+                            }
+                        }
```
certik, line 17176: There are two errors above, but only one test below, so by definition one of these two is not tested. Let's improve the test to test both and ensure they are both actually tested, by commenting each one out individually and ensuring the test fails.

certik (commented):
The following program works with this PR:
```
module halvestring_mod
contains
    pure function half(s1) result(s2)
        character(*), intent(in) :: s1
        character(len(s1)/2) :: s2
        s2 = s1
    end function half
end module halvestring_mod

program X
use halvestring_mod, only: half

contains

    subroutine f()
    character(*), parameter :: spec_str = 'abcdeedcba'
    character(len(half(spec_str))) :: spec_stringb
    print *, len(spec_stringb)
    end subroutine

end program
```
As it does with other compilers.

So `character(len(half(spec_str)))` is allowed in subroutines, just not the main or modules.

Can you please add this program into integration tests, to ensure it works?

Next, also add a test for modules. This should fail:
```fortran
module halvestring_mod
contains
    pure function half(s1) result(s2)
        character(*), intent(in) :: s1
        character(len(s1)/2) :: s2
        s2 = s1
    end function half
end module halvestring_mod

module Xm
use halvestring_mod, only: half

character(*), parameter :: spec_str = 'abcdeedcba'
character(len(half(spec_str))) :: spec_stringb

contains

end module

program X
use Xm

end program
```
Like this:
```console
e.f90:14:46:

   14 | character(len(half(spec_str))) :: spec_stringb
      |                                              1
Error: ‘spec_stringb’ at (1) must have constant character length in this context
e.f90:21:5:

   21 | use Xm
      |     1
Fatal Error: Cannot open module file ‘xm.mod’ for reading at (1): No such file or directory
compilation terminated.
```

But currently there is an ICE:
```console
$ lfortran e.f90
LCOMPILERS_ASSERT failed: /Users/ondrej/repos/lfortran/src/libasr/codegen/llvm_utils.cpp
function declare_global_string(), line number 2503 at
ASRUtils::is_value_constant(str->m_len)
ERROR MESSAGE:
Global variable should have constant-compile-time known length
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2805
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 2732
  File "/Users/ondrej/repos/lfortran/src/bin/lfortran.cpp", line 1250
  File "/Users/ondrej/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 381
    if (res.ok) {
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 19846
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 336
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 363
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 22
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 1551
    visit_symbol(*mod);
  File "/Users/ondrej/repos/lfortran/src/libasr/asr_base_visitor.h", line 365
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5077
    visit_Variable(*v);
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 4615
    ptr = llvm_utils->declare_global_string(str, initial_string_value,
  File "/Users/ondrej/repos/lfortran/src/libasr/codegen/llvm_utils.cpp", line 2504
    }
  Binary file "/usr/lib/system/libsystem_c.dylib", local address: 0x18037c9e3
  Binary file "/usr/lib/system/libsystem_pthread.dylib", local address: 0x180473847
```

This might be an unrelated error in codegen --- however it should not get to codegen in the first place... It must give the error.

So all in all, this PR is a good first step, but it's not finished:

* Doesn't catch all errors on two fronts:
  * More complicated expressions
  * Modules
* The testing only tests one of the two errors added

certik:
See also https://github.com/lfortran/lfortran/issues/6344#issuecomment-3977317737 for what we need to do to fix it.

certik:
Thanks, let me review this carefully, to ensure we get it right.

certik:
Bug: adding "needs review" triggered adding all exhaustive checks. I don't mind actually in this case, but it's still a bug. https://github.com/lfortran/lfortran/issues/10378

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -17057,6 +17100,23 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                         _processing_char_len = true;
                         this->visit_expr(*len_item->m_value);
                         ASR::expr_t* len_expr = ASRUtils::EXPR(tmp);
+                        if (!in_Subroutine && scope_forbids_automatic_data(current_scope)) {
+                            FindUserDefinedFuncCall finder;
```
assem2002, line 17104: We probably don't need a visitor to go look for a functionCall.
A simple check with `ASRUtils::is_value_constant()` should be enough.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -29,6 +29,56 @@ using LCompilers::diag::Diagnostic;
...
+            }
+            func_name = ASRUtils::symbol_name(fn_sym);
+            return;
+        }
+        ASR::BaseWalkVisitor<FindUserDefinedFuncCall>::visit_FunctionCall(x);
+    }
+};
+
+static inline bool scope_forbids_automatic_data(SymbolTable* scope) {
+    while (scope && scope->asr_owner) {
```
assem2002, line 66: Upleveling to parent scope here, doesn't seem right.
I think we can change that into `is_scope_considered_global_scope` + remove the upleveling to parent+ maybe with some proper commenting also, or drop the function and condition on `program` and `module` on the spot

PR #10357, opened 2026-02-28
fix: handle pack() with allocatable result and nested implied-do-loops

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6309

### DIFF
--- a/src/libasr/pass/intrinsic_function.cpp
+++ b/src/libasr/pass/intrinsic_function.cpp
@@ -305,66 +305,6 @@ class ReplaceFunctionCallReturningArray: public ASR::BaseExprReplacer<ReplaceFun
         new_arg.m_value = this->result_var_;
         new_args.push_back(al, new_arg);
 
-        ASRUtils::IntrinsicArrayFunctions intrinsic_id = func2intrinsicid[x_m_name];
-        if (intrinsic_id == ASRUtils::IntrinsicArrayFunctions::Pack &&
-            ASRUtils::is_allocatable(ASRUtils::expr_type(this->result_var_)) &&
-            x->n_args > 1 && x->m_args[1].m_value &&
-            ASRUtils::is_array(ASRUtils::expr_type(x->m_args[1].m_value))) {
-            const Location& loc = x->base.base.loc;
-            ASR::ttype_t* int32_type = ASRUtils::TYPE(
-                ASR::make_Integer_t(al, loc, 4));
-
-            ASR::expr_t* alloc_size = nullptr;
-            ASR::expr_t* mask_arg = x->m_args[1].m_value;
-            SymbolTable* scope = ASRUtils::symbol_parent_symtab(
-                ASRUtils::symbol_get_past_external(x->m_name));
-
-            Vec<ASR::ttype_t*> count_arg_types;
-            count_arg_types.reserve(al, 1);
-            count_arg_types.push_back(al, ASRUtils::expr_type(mask_arg));
-
-            Vec<ASR::call_arg_t> count_m_args;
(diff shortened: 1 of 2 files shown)


src/libasr/pass/implied_do_loops.cpp
```diff
@@ -157,6 +157,27 @@ class ReplaceArrayConstant: public ASR::BaseExprReplacer<ReplaceArrayConstant> {
...
+        if (ASR::is_a<ASR::IntegerUnaryMinus_t>(*expr)) {
+            return expr_references_symbol(
+                ASR::down_cast<ASR::IntegerUnaryMinus_t>(expr)->m_arg, sym);
+        }
+        if (ASR::is_a<ASR::Cast_t>(*expr)) {
+            return expr_references_symbol(
+                ASR::down_cast<ASR::Cast_t>(expr)->m_arg, sym);
+        }
+        return false;
+    }
```
certik: This function is copied twice, right?

And it is not complete, it only handles a subset of nodes, correct?

Let's figure out how to create a complete implementation, or alternatively return a hard error on all nodes that are not supported yet. Don't silently return "false" just because the node is not handled, that will cause a lot of bugs later to report, isolate and fix.

src/libasr/asr_utils.cpp
```diff
@@ -3163,6 +3163,27 @@ ASR::expr_t* get_ArrayConstant_size(Allocator& al, ASR::ArrayConstant_t* x) {
...
+    if (ASR::is_a<ASR::IntegerUnaryMinus_t>(*expr)) {
+        return expr_references_symbol(
+            ASR::down_cast<ASR::IntegerUnaryMinus_t>(expr)->m_arg, sym);
+    }
+    if (ASR::is_a<ASR::Cast_t>(*expr)) {
+        return expr_references_symbol(
+            ASR::down_cast<ASR::Cast_t>(expr)->m_arg, sym);
+    }
+    return false;
+}
```
certik: Second copy.

certik:
Requires careful review also.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -304,6 +304,68 @@ class ReplaceFunctionCallReturningArray: public ASR::BaseExprReplacer<ReplaceFun
...
+            dealloc_args.reserve(al, 1);
+            dealloc_args.push_back(al, this->result_var_);
+            pass_result.push_back(al, ASRUtils::STMT(ASR::make_ExplicitDeallocate_t(
+                al, loc, dealloc_args.p, dealloc_args.size())));
+
+            pass_result.push_back(al, ASRUtils::STMT(ASR::make_Allocate_t(
+                al, loc, alloc_args.p, alloc_args.size(), nullptr, nullptr, nullptr)));
+        }
+
         ASR::stmt_t* subrout_call = ASRUtils::STMT(ASRUtils::make_SubroutineCall_t_util(
```
jinangshah21, line 309: I think above change (special reallocation) is not needed because `--realloc-lhs-arrays` flag will take care of that.

certik:
I think this is good. @HarshitaKalani do you mind please checking https://github.com/lfortran/lfortran/pull/10357#discussion_r2869593030 ? That might further simplify it.

Otherwise just resolve conflicts and it should be ready to merge.

src/libasr/pass/intrinsic_function.cpp
HarshitaKalani: Yes, right. Updating.

certik:
I squashed it, as there was a commit that adds code, then another commit that removes this very code, so I think it's better to just have a simpler commit.

PR #10353, opened 2026-02-28
fix: resolve external procedures in statement function scope

Pranavchiku:
Fixes https://github.com/lfortran/lfortran/issues/10030. The second commit fixes main failure.

PR #10360, opened 2026-02-28
fix: ICE in visit_Associate with stale tmp

certik:
In visit_Associate, after visiting target and value expressions, tmp held a stale ASR expr node. When no branch produced an Associate statement (e.g. mismatched FunctionType in continue-compilation mode), the stale expr was passed to STMT() in transform_stmts, causing an assertion failure.

Fix by setting tmp = nullptr after extracting the expressions, so transform_stmts safely skips the unhandled case.


certik:
Should fix the failure in main hopefully.

certik:
The added tests works in main, so it's not a good test. A minimal test for the failure is:
```fortran
program main

procedure(undef), pointer :: p
p => f

contains

    integer function f()
    f = 1
    end function

end program
```
The crash only triggers when `procedure(undef)` references something that doesn't exist. The undefined interface causes a semantic error, but instead of reporting it cleanly, the subsequent `p => f` pointer assignment leaves tmp as a non-statement ASR node, hitting the assertion. The file `tests/errors/continue_compilation_1.f90` already tests exactly this problem (that's why it failed), so only the fix is needed, which was done in https://github.com/lfortran/lfortran/pull/10353.

PR #10362, opened 2026-02-28
fix : read an internal file if a substring is involved

### LINKED ISSUE #8582: Reading an internal file fails if a substring is involved
Case 1: reading from a whole character variable into a substring.
Case 2: reading from a substring into a whole character variable.

Program for Case 1:
```
  implicit none ! file substringint1.f90
  character(8):: string = 'abcd', newstring = 'ABCD'
  read(string,"(A)") newstring(5:8)
  print "(A)", newstring
end program
```
```
ABCDabcd
```
Output from lfortran 0.56.0 is long and ends with
```
ptr @string
 i32  call void (i32, ptr, ptr, ptr, i64, ptr, i32, ...) @_lfortran_formatted_read(ptr @string, ptr %2, ptr %3, ptr @0, i64 3, ptr %4, i32 2, ptr %5, i64 4)



```
Program for Case 2:
```
  implicit none ! file substringint2.f90
  character(8):: string = 'abcd', newstring = 'ABCD'
  read(string(1:4),"(A)") newstring
  print "(A)", newstring
end program
```
```
abcd
```
Output from lfortran 0.56.0 is long and ends with
```
ptr @string_const
 i32  call void (i32, ptr, ptr, ptr, i64, ptr, i32, ...) @_lfortran_formatted_read(ptr @string_const, ptr %2, ptr %3, ptr @0, i64 3, ptr %4, i32 2, ptr %5, i64 8)



```

yuvrajsinghgmx:
Closes #8582

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -813,10 +813,13 @@ RUN(NAME arrays_102 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --r
 RUN(NAME arrays_103 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME arrays_104 LABELS gfortran llvm)
 RUN(NAME arrays_105 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME arrays_106 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME arrays_107 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 # DISABLED: #8115 - ICE: get_struct_sym_from_struct_expr returns nullptr for empty struct array constructors [tp ::]
 # RUN(NAME array_init LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME assumed_rank_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME select_rank_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME select_rank_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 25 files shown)


yuvrajsinghgmx:
I have resolved merge conflicts.

HarshitaKalani (commented):
The test works for me in main. What is the code addition for? Can you add a test which fails in main and works with the changes in this PR? Mark it as ready for review once done.

yuvrajsinghgmx:
> The test works for me in main. What is the code addition for? Can you add a test which fails in main and works with the changes in this PR? Mark it as ready for review once done.

The test was passing but CI was failing due to `--fast` and the changes are made to fix the same.

certik:
The test passes both with and without `--fast` on Apple M4 on main:
```
(lf) ~/repos/lfortran(main)$ lfortran a.f90
(lf) ~/repos/lfortran(main)$ lfortran a.f90 --fast
(lf) ~/repos/lfortran(main)$
```
Can you show the failure that you are getting with `--fast`?

It seems this is not a good test for the change. 

Update: my bad, it does *not* pass with `--fast`:
```
(lf) ~/repos/lfortran(main)$ lfortran a.f90
(lf) ~/repos/lfortran(main)$ echo $?
0
(lf) ~/repos/lfortran(main)$ lfortran a.f90 --fast
(lf) ~/repos/lfortran(main)$ echo $?
10
```

src/libasr/pass/replace_with_compile_time_values.cpp
```diff
@@ -202,6 +202,47 @@ class ExprVisitor: public ASR::CallReplacerOnExpressionsVisitor<ExprVisitor> {
...
+            this->visit_expr(*xx.m_values[i]);
+            replacer.inside_prohibited_expression = inside_copy;
+        }
+        if (xx.m_fmt) {
+            current_expr = &(xx.m_fmt);
+            replacer.current_expr = current_expr;
+            ASR::CallReplacerOnExpressionsVisitor<ExprVisitor>::visit_expr(**current_expr);
+        }
+        current_expr = current_expr_copy;
+    }
```
certik, line 244: Can you explain what this part is doing? Is it replacing some arguments to FileRead with their compile-time values?

It seems that is hiding some bug, since things must work both for compile-time and runtime values.

certik: I think this was the problem:

The replace_with_compile_time_values optimization pass saw that `newstring(5:8)` had a compile-time value of `""` (from initialization) and replaced the read target with that constant. So instead of reading into the actual newstring substring, the runtime wrote into read-only constant memory, and the original newstring was never modified — causing the subsequent if check to fail.

So this might be a good fix.

PR #10366, opened 2026-02-28
Fix several stdlib bugs

certik:


certik:
Let's get https://github.com/lfortran/lfortran/pull/10367 to merge first, then we can merge this one.

PR #10367, opened 2026-02-28
Fix passing function to procedure pointer arg

### LINKED ISSUE #10364: Call of procedure pointer from a subroutine fails
`$ cat main.f90`
```
program main

   abstract interface
      subroutine intfc()
      end subroutine intfc
   end interface
   
   call output(hello)
   
contains
   
   subroutine output(ptr)
      procedure(intfc), pointer, intent(in) :: ptr
      call ptr()
   end subroutine output
   
   subroutine hello()
      print *, 'Hello world!'
   end subroutine hello

end program main
```
`$ flang-20 main.f90 -o main ; ./main`
 ```
Hello world!
```
`$ lfortran main.f90 `
```
```

### PR DESCRIPTION
**Pull request by certik**
When a bare function name is passed as an actual argument to a subroutine expecting a procedure pointer (procedure(...), pointer), the LLVM codegen was passing the function pointer directly (void()*) instead of wrapping it in an alloca to produce the expected void()** type. This caused an LLVM module verification failure.

The fix stores the function pointer into an alloca and passes the address when the target parameter has a pointer type.

Fixes #10364.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2873,7 +2873,7 @@ RUN(NAME procedure_29 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME procedure_30 LABELS llvm)
 RUN(NAME procedure_31 LABELS gfortran llvm)
-RUN(NAME procedure_32 LABELS gfortran llvm)
+RUN(NAME procedure_32 LABELS llvm)
 
 
 RUN(NAME allocated_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)

### CODE REVIEW

File: integration_tests/CMakeLists.txt [hunk 1]
```diff
@@ -2873,6 +2873,7 @@ RUN(NAME procedure_29 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME procedure_30 LABELS llvm)
 RUN(NAME procedure_31 LABELS gfortran llvm)
+RUN(NAME procedure_32 LABELS gfortran llvm)
```
```suggestion
RUN(NAME procedure_32 LABELS llvm)
```

PR #10372, opened 2026-03-01
Fix more stdlib bugs

certik:
Now we compile it to 41%.


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -6097,10 +6097,25 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+        {
+            ASR::symbol_t *sym = current_scope->resolve_symbol(sub_name);
+            if (sym) {
+                ASR::symbol_t *s = ASRUtils::symbol_get_past_external(sym);
+                if (ASR::is_a<ASR::Function_t>(*s) ||
+                    ASR::is_a<ASR::GenericProcedure_t>(*s)) {
+                    user_procedure_found = true;
+                }
+            }
+        }
```
certik, line 6112: This should be refactored into a function and also I think the same issue exists with FunctionCall, so we need to add a test and fix it there. Can be done after this PR is merged.

Issue #6109, opened 2025-01-24
transfer bug


harperjf:
Two-line test program:
```
  print "(A)",transfer(['x','y','z'],'abc')
  end program
```
Output from lfortran 0.44.0:
```
qB^
```
Output from a second try with lfortran 0.44.0:
```
���HV
```
Output from gfortran, g95, ifort, ifx or AMD flang (correct)::
```
xyz
```

gxyd:
With the latest LFortran in development, for the program:
```f90
  print "(A)",transfer(['x','y','z'],'abc')
  end program
```

I get the below error:
```console
> lfortran --version
LFortran version: 0.44.0-148-g1368799ca-dirty
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin23.6.0

> lfortran test.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

@0 = private unnamed_addr constant [2 x i8] c"x\00", align 1
@1 = private unnamed_addr constant [2 x i8] c"y\00", align 1
@2 = private unnamed_addr constant [2 x i8] c"z\00", align 1
@3 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@4 = private unnamed_addr constant [4 x i8] c"(A)\00", align 1
@5 = private unnamed_addr constant [5 x i8] c"%s%s\00", align 1

define i32 @main(i32 %0, i8** %1) {
.entry:
  %bitcast_source = alloca [3 x i8*], align 8
  %array_bound = alloca i32, align 4
  %__1_k = alloca i32, align 4
  call void @_lpython_call_initial_functions(i32 %0, i8** %1)
  %__1_k1 = alloca i32, align 4
  %__libasr__created__var__0__array_constant_ = alloca [3 x i8*], align 8
  br i1 true, label %then, label %else

then:                                             ; preds = %.entry
  store i32 1, i32* %array_bound, align 4
  br label %ifcont

else:                                             ; preds = %.entry
  br label %ifcont

ifcont:                                           ; preds = %else, %then
  %2 = load i32, i32* %array_bound, align 4
  store i32 %2, i32* %__1_k1, align 4
  %3 = load i32, i32* %__1_k1, align 4
  %4 = sub i32 %3, 1
  %5 = mul i32 1, %4
  %6 = add i32 0, %5
  %7 = getelementptr [3 x i8*], [3 x i8*]* %__libasr__created__var__0__array_constant_, i32 0, i32 %6
[...]

PR #10376, opened 2026-03-01
Fix transfer() compile-time eval for character array sources

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6109


jinangshah21 (approved):
Looks Good

PR #10380, opened 2026-03-01
fix: disable functions duplication having custom type select statements

### LINKED ISSUE #10375: Compilation segfaults on `select type` statement that accesses an array of an extended type
`$ cat test.f90`
```
module clientmod
   
   type, abstract :: AbsType
   end type AbsType
   
      real(8), allocatable :: arr(:,:)
   end type MyType

contains

   subroutine reshaper(obj,n)
      integer,        intent(in) :: n(:)      
      real(8), dimension(n(1)*n(2)) :: a
      select type (obj)
      type is (MyType)
         a = reshape(obj%arr(1:n(1),1:n(2)),[n(1)*n(2)])      
      end select
   end subroutine reshaper

end module clientmod
```

`$ lfortran -c --separate-compilation test.f90`

```
```

YashNagda17:
Fixes #10375

Changes made:
- For functions having blocks which can't be duplicated into new symbol table, disable its duplication inside `pass_array_by_data` pass.

### DIFF
--- a/src/libasr/pass/pass_array_by_data.cpp
+++ b/src/libasr/pass/pass_array_by_data.cpp
@@ -45,6 +45,34 @@ Dis-advantages:
 
 namespace LCompilers {
 
+// After duplicating a function body and its symbol table, BlockCall statements
+// still reference the original Block symbols. This visitor remaps them to the
+// corresponding duplicated Blocks in the new symbol table.
+class BlockCallRemapper : public ASR::BaseWalkVisitor<BlockCallRemapper> {
+    SymbolTable* new_symtab;
+public:
+    BlockCallRemapper(SymbolTable* new_symtab_) : new_symtab(new_symtab_) {}
+
+    void visit_BlockCall(const ASR::BlockCall_t& x) {
+        ASR::BlockCall_t& xx = const_cast<ASR::BlockCall_t&>(x);
+        if (ASR::is_a<ASR::Block_t>(*xx.m_m)) {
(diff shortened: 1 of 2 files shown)


src/libasr/pass/pass_array_by_data.cpp
```diff
@@ -112,7 +112,19 @@ class PassArrayByDataProcedureVisitor : public PassUtils::PassVisitor<PassArrayB
...
+            // into new_symtab, skip duplicating this function. Otherwise the duplicated
+            // body's BlockCall would still reference the original Block in the wrong scope.
+            for( auto& item: x->m_symtab->get_scope() ) {
+                if (ASR::is_a<ASR::Block_t>(*item.second)) { 
+                    std::string block_name = std::string(ASR::down_cast<ASR::Block_t>(item.second)->m_name);
+                    if (!new_symtab->get_symbol(block_name)) {
+                        return nullptr;
+                    }
+                }
+            }
```
certik: Can you explain exactly when a block can be duplicated and why it would be duplicated and what is "new symtab" in this case?

I don't understand the fix.

certik: My best understanding is that this skips duplicating the function if it contains a block. So it is preventing this pass to run, arbitrarily. That doesn't seem like the right fix.

Instead, either let's fix the pass, or let's get rid of it, but I don't like these hacks/workarounds just to get something working.

certik: I implemented a fix in 0f4976fa2a4d751e940d20d8731f3eba1abea485.

PR #10382, opened 2026-03-01
enh: change runtime errors from "no-realloc-lhs-arrays" to warnings

### LINKED ISSUE #10331: Proposal: change runtime errors from "no-realloc-lhs-arrays" to warnings
For:
```fortran
integer, allocatable :: x(:)
allocate(x(2))
print *, size(x)
x = [1, 2, 3]
print *, size(x)
end
```
The current behavior is:
```console
$ lfortran b.f90 
2
runtime error: Array shape mismatch in assignment to 'x'. Tried to match size 2 of dimension 1 of LHS with size 3 of dimension 1 of RHS. Use '--realloc-lhs-arrays' option to reallocate LHS automatically.
 --> b.f90:4:1
  |
4 | x = [1, 2, 3]
  | ^ LHS size is 2
  |
4 | x = [1, 2, 3]
  |     ^^^^^^^^^ RHS size is 3
2
3
```

This proposal is to make the behavior as:
```console
$ lfortran b.f90 
2
runtime warning: The array `x` is being automatically reallocated, potentially slowing performance, due to LHS being size 2, and RHS being size 3. If this is intended, allocate explicitly before the assignment using `allocate(x(2))` or use the '--realloc-lhs-arrays' option to suppress this warning.
 --> b.f90:4:1
  |
4 | x = [1, 2, 3]
  | ^ LHS size is 2
  |
4 | x = [1, 2, 3]
  |     ^^^^^^^^^ RHS size is 3
3
2
3
```
That is, in both default mode and `'--realloc-lhs-arrays` mode, the program would still run and produce the same output `2\n3` in both cases, except that in the default mode it gives a runtime warning.

The other consequence of this proposal is that `--fast` must check and perform runtime reallocation by default, otherwise it will break for the above example. So to get even faster results (equivalent to current `--fast`), one would use `--fast --no-realloc-lhs-arrays`. When using `--no-realloc-lhs-arrays` without `--fast`, it will give a runtime error instead of a warning. The new contract is: "if the code compiles and runs without errors and without warnings", you can use `--fast --no-realloc-lhs-arrays`.

We can change the names of these options and modes, but hopefully the proposed behavior is clear from this proposal. The motivation is to compile people's codes in default mode (possibly with warnings). The default warnings might be annoying to some, but less annoying than hard errors (those are really annoying). At the same time, the new behavior still preserves the ability to run fast with no checks and no reallocation at runtime: in default mode you get warnings so you can fix your code; you can also set `--no-realloc-lhs-arrays` and then you get hard errors, for example at the CI. So it's a compromise that will allow existing codes and users who like this automatic reallocation feature to be able to use it by default, while still by default guiding and allowing to easily get maximum performance with no hassle.

opixdown:
closes #10331 

## **New Behaviour**
| Mode                          | Behavior                                      |
|-------------------------------|-----------------------------------------------|
| default                       | reallocates + prints runtime warning          |
| --realloc-lhs-arrays          | reallocates silently, no warning              |
| --no-realloc-lhs-arrays       | hard runtime error, no reallocation           |
| --fast                        | reallocates silently (same as --realloc-lhs-arrays) |

## **MRE**
```
integer, allocatable :: x(:)
allocate(x(2))
print *, size(x)
x = [1, 2, 3]
print *, size(x)
end
```

## **Outputs**
```
% lfortran t.f90 
2
runtime warning: The array 'x' is being automatically reallocated, potentially slowing performance, due to LHS being size 2, and RHS being size 3. If this is intended, allocate explicitly before the assignment or use the '--realloc-lhs-arrays' option to suppress this warning.
 --> t.f90:4:1
  |
4 | x = [1, 2, 3]
  | ^ LHS size is 2
  |
4 | x = [1, 2, 3]
  |     ^^^^^^^^^ RHS size is 3
3
```
```
 % lfortran t.f90 --realloc-lhs-arrays
2
3
```
```
% lfortran t.f90 --no-realloc-lhs-arrays
2
runtime error: Array shape mismatch in assignment to 'x'. LHS size 2 dim 1, RHS size 3 dim 1.
 --> t.f90:4:1
  |
4 | x = [1, 2, 3]
  | ^ LHS size is 2
  |
4 | x = [1, 2, 3]
  |     ^^^^^^^^^ RHS size is 3
 ```
```
 % lfortran t.f90 --realloc-lhs-arrays --no-realloc-lhs-arrays
Cannot use --realloc-lhs-arrays and --no-realloc-lhs-arrays at the same time
 ```


opixdown:
we need to review the updated test references, documentation and any test cases referencing the old ` --disable-realloc-lhs-arrays` flag or expecting a hard runtime error on LHS shape mismatch may need to be updated.

PR #10383, opened 2026-03-01
fix: Resolve procedure placeholders in struct symbol

### LINKED ISSUE #10377: Compilation fails for procedure pointer association through a pointer to a derived type
`$ cat test.f90`
```
module test
   
   type :: MyType
      procedure(pintfc), pointer, nopass :: ptr => null()
   end type MyType

   abstract interface
      function pintfc() result(res)
         real(8) :: res
      end function pintfc
   end interface

contains

   subroutine pointers()
      class(MyType),     pointer :: myptr => null()
      procedure(pintfc), pointer :: ptr => null()
      ptr => myptr%ptr
   end subroutine pointers

end module test
```
```
warning: Assuming implicit save attribute for variable declaration
  --> test.f90:16:37
   |
16 |       class(MyType),     pointer :: myptr => null()
   |                                     ^^^^^^^^^^^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement

warning: Assuming implicit save attribute for variable declaration
  --> test.f90:17:37
   |
17 |       procedure(pintfc), pointer :: ptr => null()
   |                                     ^^^^^^^^^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement

semantic error: Interface mismatch in procedure pointer assignment at (1): 'unknown' is not a function
  --> test.f90:18:7
   |
18 |       ptr => myptr%ptr
   |       ^^^^^^^^^^^^^^^^ (1)
```
Both the warnings and the semantic error are bugs (pointer initialization via association to `null()` is considered good practice since Fortran 95).

jinangshah21:
Fixes #10377

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2899,7 +2899,7 @@ RUN(NAME procedure_29 LABELS gfortran llvm
 RUN(NAME procedure_30 LABELS llvm)
 RUN(NAME procedure_31 LABELS gfortran llvm)
 RUN(NAME procedure_32 LABELS llvm)
-
+RUN(NAME procedure_33 LABELS gfortran llvm)
 
 RUN(NAME allocated_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocated_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


jinangshah21:
First we visit struct symbol where procedure variable doesn't have symbol for its declaration type function, so we create placeholder there and when we get actual function interface, then we need update procedure variable's type_declaration which this PR does.

PR #10385, opened 2026-03-01
test: add test for issue #8581

### LINKED ISSUE #8581: Reading an external file into a substring fails
This program shows that lfortran can write a substring into an external file but cannot read from an external file into a substring.
```
  implicit none ! substringext.f90
  character(8):: string = 'abcd'
  integer:: unit = 42
  open(newunit=unit, file='foobar', status='new')
  write (unit, "(A)") string(1:4)
  rewind unit
  read  (unit,"(A)") string(5:8)
  close (unit,status = 'delete')
  print "(A)",string
end program
```
If the file ```foobar``` does not exist before the program runs, output from gfortran, g95, ifx or AMD flang is
```
abcdabcd
```
but output from lfortran 0.56.0 is
```
```
The file ```foobar``` contained ```abcd``` after the lfortran run, but all the other compilers had deleted it.

opixdown:
closes #8581

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2802,6 +2802,7 @@ RUN(NAME read_45 LABELS gfortran llvm)
 RUN(NAME read_46 LABELS gfortran llvm)
 RUN(NAME read_47 LABELS gfortran llvm)
 RUN(NAME read_48 LABELS gfortran llvm)
+RUN(NAME read_49 LABELS gfortran llvm)
 
 RUN(NAME write_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME write_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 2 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -2802,6 +2802,7 @@ RUN(NAME read_45 LABELS gfortran llvm)
 RUN(NAME read_46 LABELS gfortran llvm)
 RUN(NAME read_47 LABELS gfortran llvm)
 RUN(NAME read_48 LABELS gfortran llvm)
+RUN(NAME read_49 LABELS gfortran llvm NOFAST) // test fails with --fast
```
HarshitaKalani: PR fails because of the usage of // here. You can use # to comment, but it's not needed here as it's clear from the usage of NOFAST.
```suggestion
RUN(NAME read_49 LABELS gfortran llvm NOFAST)
```

HarshitaKalani: Although, it works with fast for me. 
```
 % lfortran a.f90 --fast    
abcdabcd
Test passed
```

opixdown: okayy thankyou i'll remove the no fast tag

Issue #6877, opened 2025-04-06
Array bug: elements with different type parameters


harperjf:
Lfortran 0.50.0 printed
```
Lengths: 5 6
```
with this program:
```
  implicit none
  character(:),allocatable::a,b,words(:)
  a = 'first'
  b = 'second'
  allocate(character(max(len(a),len(b))) :: words(2))
  words(1) = a
  words(2) = b
  print "(A,2I2)",'Lengths:',len(words(1)),len(words(2))
end program
```
Gfortran, ifx and AMD flang all printed
```
Lengths: 6 6
```
and they were right, because the F2023 standard 3.3 says an array is a set of scalar data, all of the same type and type parameters, and 7.2 para 2 says a type parameter is either a kind type parameter or a length type parameter.

harperjf:
A shorter program showing what may be the same bug:
```
 character(:), allocatable:: x(2)
  x = [ 'aa','bb' ]
  print "(2A)", x
end program
```
Gfortran, ifx and AMD flang won't compile it. Lfortran 0.50.0 does, and prints
```
aabb
```

assem2002:
Great MREs! 
First MRE is mainly due to how we reserve space for array of strings. It works, but doesn't meet the standards as you mentioned and fails hard for some specific cases. I like how you mention the actual Fortran standards, it sets robust rules while approaching the open issue when read for the first time.

Second MRE is just a semantic error, Feel free to open another issue for it.

czgdp1807:
The fix for this is simple, we need to allocate, all the elements of character array to the size provided in `allocate`. Can be done by generating while loop. :-).

HarshitaKalani:
This works now:
```console
% lfortran a.f90
Lengths: 6 6
]% gfortran a.f90 && ./a.out
Lengths: 6 6
```

harperjf:
Good!

On Mon, 26 Jan 2026, Harshita Kalani wrote:

> Date: Mon, 26 Jan 2026 09:02:44 +0000
> From: Harshita Kalani ***@***.***>
> Reply-To: lfortran/lfortran
>     ***@***.***>
> To: lfortran/lfortran ***@***.***>
> Cc: harperjf ***@***.***>, Author ***@***.***>
> Subject: Re: [lfortran/lfortran] Array bug: elements with different type
>     parameters (Issue #6877)
> 
> HarshitaKalani left a comment (lfortran/lfortran#6877)
>
> This works now:
> ```console
[...]

Issue #6783, opened 2025-03-30
Regression with EXTERNAL


harperjf:
This program runs with ifx, gfortran and lfortran 0.49.0 but not 0.50.0
```
! file testspecific1.f90 
module specificmod
  implicit none
  integer, parameter:: dp=kind(1d0)
contains

  subroutine root2sqdp(sqrt)
    real(dp),external::sqrt 
    print "(A,F0.16)",'check sqrt(2.0_dp)**2 = ',sqrt(2.0_dp)**2
  end subroutine root2sqdp
  
end module specificmod

program testspecific1
use specificmod
  implicit none
  intrinsic dsqrt
  call root2sqdp(dsqrt)
end program testspecific1
```
Here, dsqrt is the obsolescent but still standard double precision sqrt.
Output from  ifx, gfortran and lfortran 0.49.0
```
check sqrt(2.0_dp)**2 = 2.0000000000000004
```
Output from lfortran 0.50.0
```
emantic error: Variable 'dsqrt' is not declared
  --> testspecific1.f90:20:18
   |
20 |   call root2sqdp(dsqrt)
   |                  ^^^^^ 'dsqrt' is undeclared
```

Pranavchiku:
I'm able to reproduce it with latest main as well. Thanks for reporting!

HarshitaKalani:
Works with the latest main. Let's add a test to close this issue.
```console
 % lfortran a.f90 --implicit-interface
check sqrt(2.0_dp)**2 = 2.0000000000000004
% gfortran a.f90 && ./a.out
check sqrt(2.0_dp)**2 = 2.0000000000000004
```

harperjf:
OK with lfortran 0.59.0 - not now the latest.

On Thu, 29 Jan 2026, Harshita Kalani wrote:

> Date: Thu, 29 Jan 2026 15:30:24 +0000
> From: Harshita Kalani ***@***.***>
> Reply-To: lfortran/lfortran
>     ***@***.***>
> To: lfortran/lfortran ***@***.***>
> Cc: harperjf ***@***.***>, Author ***@***.***>
> Subject: Re: [lfortran/lfortran] Regression with EXTERNAL (Issue #6783)
> 
> HarshitaKalani left a comment (lfortran/lfortran#6783)
>
> Works with the latest main. Let's add a test to close this issue.
> ```console
> % lfortran a.f90 --implicit-interface
> check sqrt(2.0_dp)**2 = 2.0000000000000004
[...]

Issue #6702, opened 2025-03-21
External function problem


harperjf:
#6690 dealt with an external subroutine with the same name as an intrinsic. This program reveals a problem if an external function has the same name as an intrinsic function. (I have not tested against an intrinsic subroutine.)
```
! externalfun.f90
  implicit none
  character(8), external:: sin
  print "(A)", sin()
end program

character(8) function sin()
  sin = 'Peccavi!'
end function sin
```
Output from gfortran after a warning
```
Peccavi!
```
Output from lfortran 0.49.0 using --std=f23
```
Semantic error: No matching signature found for intrinsic sin
 --> externalfun.f90:4:16
  |
4 |   print "(A)", sin()
  |                ^^^^^ 
```

jinangshah21:
This is working good with current LFortran main, I think it is solved by #6697.

Pranavchiku:
```console
% cat a.f90 
! externalfun.f90
  implicit none
  character(8), external:: sin
  print "(A)", sin()
end program

character(8) function sin()
  sin = 'Peccavi!'
end function sin
% lfortran a.f90 --implicit-interface
Peccavi!
```

Yes, this works with latest main. It will be available with release `0.50.0`.

HarshitaKalani:
Let's close this by adding a test.

harperjf:
OK

On Thu, 29 Jan 2026, Harshita Kalani wrote:

> Date: Thu, 29 Jan 2026 15:31:03 +0000
> From: Harshita Kalani ***@***.***>
> Reply-To: lfortran/lfortran
>     ***@***.***>
> To: lfortran/lfortran ***@***.***>
> Cc: harperjf ***@***.***>, Author ***@***.***>
> Subject: Re: [lfortran/lfortran] External function problem (Issue #6702)
> 
> HarshitaKalani left a comment (lfortran/lfortran#6702)
>
> Let's close this by adding a test.
>
>
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/lfortran/lfortran/issues/6702#issuecomment-3818425211
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>
>


-- John Harper, School of Mathematics and Statistics
Victoria Univ. of Wellington, PO Box 600, Wellington 6140, New Zealand.
e-mail ***@***.***

PR #10386, opened 2026-03-01
test: add test for fixed issue

### LINKED ISSUE #4675: Unhelpful seg.fault with --std=f23 of program OK without it
This program was in bug report #3960. The lfortran bug has been fixed if --std=f23 is not used (hurrah!) but --std=f23 gives a seg.fault that doesn't say where the problem was in the program.
```
program fmtbuga
  character fmt*10, line*80, letters*26
  integer i
  letters = 'abcdefghijklmnopqrstuvwxyz'
  fmt = '(99(1X,A))'
  write(line,fmt) (letters(i:i), i=1,26)
  print "(A)", trim(line)
end program fmtbuga
```
Result from lfortran 0.39.0 without --std=f23
```
 a b c d e f g h i j k l m n o p q r s t u v w x y z
```
Result with --std=f23
```
```

opixdown:
added test and closes #4675  #6877 #6702  #6783


integration_tests/CMakeLists.txt
```diff
@@ -2266,6 +2267,7 @@ RUN(NAME string_100 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME string_101 LABELS gfortran llvm)
 RUN(NAME string_102 LABELS gfortran llvm)
 RUN(NAME string_103 LABELS gfortran llvm)
+RUN(NAME string_104 LABELS gfortran llvm EXTRA_ARGS --std=f23)
```
certik: I believe every test is tested with `--std=f23`. Can you verify that is the case? If so, then remove this, so that it runs with all modes.

HarshitaKalani: Yes, that's correct. CI runs all the tests with --std=f23. We need not add this explicitly here.

```suggestion
RUN(NAME string_104 LABELS gfortran llvm)
```

integration_tests/arrays_16.f90
```diff
@@ -0,0 +1,14 @@
+implicit none
```
HarshitaKalani: Can you add a program name here? And same for other tests as well.

opixdown: done

integration_tests/CMakeLists.txt
opixdown: i removed in current commit

opixdown:
> Can you please address the comments and mark it as ready for review once done? We can merge the PR then. Thanks!

changes done we can merge it

certik:
It might have already been merged in the other PR. If there is something left, please rebase it.

opixdown:
> 4675

all tests have been merged in #10431 so clossing this pr

PR #10392, opened 2026-03-02
integration_tests: use newunit for scratch I/O tests

krystophny:
## Summary
Use `newunit=` instead of fixed literal unit numbers in scratch-file integration tests to avoid unit collisions in parallel or shared-runtime execution.

## Scope
- integration_tests/read_08.f90
- integration_tests/read_10.f90
- integration_tests/read_11.f90
- integration_tests/logical_testing.f90
- integration_tests/format_33.f90
- integration_tests/file_49.f90

## Why
Fixed units (for example 10, 42) can conflict when tests are executed in parallel contexts that share runtime process state. `newunit=` allocates a safe free unit and removes that flake class.

## Verification
- gfortran integration_tests/read_08.f90 && ./a.out
- gfortran integration_tests/read_10.f90 && ./a.out
- gfortran integration_tests/read_11.f90 && ./a.out
- gfortran integration_tests/logical_testing.f90 && ./a.out
- gfortran integration_tests/format_33.f90 && ./a.out
- gfortran integration_tests/file_49.f90 && ./a.out
- Parallel smoke for read_11: 24 concurrent runs all passed

### DIFF
--- a/integration_tests/file_49.f90
+++ b/integration_tests/file_49.f90
@@ -3,12 +3,13 @@
 ! Exact MRE from issue body
 program file_49
     implicit none
+    integer :: unit_no
     character :: q*12
-    open(42, status='scratch', form='unformatted')
-    write(42) 'Hello world!'
-    rewind 42
-    read(42) q
-    close(42)
+    open(newunit=unit_no, status='scratch', form='unformatted')
+    write(unit_no) 'Hello world!'
+    rewind unit_no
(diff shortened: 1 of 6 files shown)


krystophny:
... this was an issue for [liric](https://github.com/krystophny/lfortran/pull/37), where multiple tests are run in parallel in the same process. Otherwise scratch should be fine with fixed unit numbers.

PR #10395, opened 2026-03-02
Codegen: fix type mismatches in array constants, copysign, complex returns, and class deallocation

krystophny:
## Summary

Fix several type mismatch issues in LLVM IR codegen discovered through
strict IR validation:

- **Array constants**: when `type` is already an `ArrayType`, extract
  element type before creating the constant array to prevent nested
  `[N x [M x T]]` types
- **Copysign operand casting**: check operand types before casting; use
  `FPCast` for float-to-float, `SIToFP` only for int-to-float, skip
  if already correct type
- **Complex64 returns**: add explicit `BitCast` from `{float,float}*`
  to `i64*` / `<2xfloat>*` before loading, instead of relying on
  opaque pointer semantics
- **Class deallocation**: use `getStructElementType(1)` instead of
  `getPointerElementType()` (deprecated) for class field type lookup

## Test plan

- [ ] Run `./run_tests.py -j16` (reference tests)
- [ ] Run `cd integration_tests && ./run_tests.py -b llvm -j16` (integration tests)
- [ ] Verify no regressions on existing test suite

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -4383,6 +4383,12 @@ public:
         std::vector<llvm::Constant*> arr_elements;
         size_t arr_const_size = (size_t) ASRUtils::get_fixed_size_of_array(arr_const->m_type);
         arr_elements.reserve(arr_const_size);
+        llvm::ArrayType* arr_type = nullptr;
+        llvm::Type* elem_type = type;
+        if (type->isArrayTy()) {
+            arr_type = llvm::cast<llvm::ArrayType>(type);
+            elem_type = arr_type->getElementType();
+        }
         int a_kind;
         for (size_t i = 0; i < arr_const_size; i++) {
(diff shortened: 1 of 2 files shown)


certik:
@krystophny Was the previous LLVM IR invalid? If so, it should fail for some Fortran code, right?

krystophny:
This was just because liric is more sensitive to this. I couldn't find Fortran code that fails due to these issues in the current LFortran version with LLVM. The previous IR was technically incorrect in some cases (e.g., load `<2 x float>`, `{float,float}*` without bitcast for the `complex64` argument ABI conversion), but LLVM's verifier doesn't flag it on either LLVM 11 or 21. The copysign fix is or a codegen handler (`visit_RealCopySign`) that appears to be dead code as the `sign_from_value` pass always transforms it before codegen.

certik:
Thanks!

I really want to move to a nice C-style API directly in our ASR->LLVM, I think it would simplify many things. Once you are happy with liric, let me know, we can iterate on the API design.

Issue #3935, opened 2024-04-24
entry statement OK in an external subroutine, but not in a module


harperjf:
This program with an ```entry``` statement won't compile with lfortran:
```
(lf2) john:~/Test/Glist$ cat testentry2.f90
! Is entry OK? Must be in an external or module subprogram
module entrymod
  implicit none
contains
  function firstname(n)
    integer,intent(in)::n
    character(24):: firstname, entryname
    write(firstname,"(A,I0)") 'via function name: ',n
    return
    entry entryname(n)
    write(entryname,"(A,I0)") 'via entry: ',n
  end function firstname
end module entrymod

program testentry2
  use entrymod
  implicit none
  print *, entryname(42)
  print *, firstname(666)
end program

(lf2) john:~/Test/Glist$ lfortran testentry2.f90
Segmentation fault (core dumped)
```
But this one compiles and runs correctly:
```
(lf2) john:~/Test/Glist$ cat testentry.f90
! Is entry OK? Must be in an external or module subprogram
  implicit none
  call indirect
  call direct
end program

subroutine direct
  print *,'entering at direct'
  return
  entry indirect
  print *,'entering at indirect'
end subroutine direct
(lf2) john:~/Test/Glist$ lfortran testentry.f90
entering at indirect
entering at direct
(lf2) john:~/Test/Glist$ 
```

Pranavchiku:
This is not working with latest main:

```console
% lfortran a.f90 --show-asr
ASR verify pass error: ASR verify: The symbol table was not found in the scope of `symtab`.
 --> a.f90:6:21
  |
6 |     character(24):: firstname, entryname
  |                     ^^^^^^^^^ failed here

ASR verify pass error: Var::m_v `firstname` cannot point outside of its symbol table
 --> a.f90:7:11
  |
7 |     write(firstname,"(A,I0)") 'via function name: ',n
  |           ^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #10402, opened 2026-03-02
fix: handle ENTRY statement inside module functions

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/3935


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -4293,11 +4293,24 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                    ASR::stmt_t* s = ASRUtils::STMT(x);
+                    if (return_var != nullptr && s->type == ASR::stmtType::FileWrite) {
+                        ASR::FileWrite_t* fw = ASR::down_cast<ASR::FileWrite_t>(s);
+                        if (fw->m_unit && ASR::is_a<ASR::Var_t>(*fw->m_unit)) {
+                            std::string unit_name = ASRUtils::symbol_name(
+                                ASR::down_cast<ASR::Var_t>(fw->m_unit)->m_v);
+                            if (original_function_name == unit_name ||
+                                entry_functions[original_function_name].find(unit_name) !=
+                                    entry_functions[original_function_name].end()) {
+                                fw->m_unit = return_var;
```
assem2002: I think this is deciding the return type of the ASR function; If that's the case, I think we should find another way to do that.

PR #10404, opened 2026-03-02
Fix more stdlib bugs

certik:
Gets stdlib compiling to 60%. All commits are independent and should pass all tests.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -5951,7 +5951,7 @@ public:
             bool is_proc_pointer_var = false;
             if (ASR::is_a<ASR::Variable_t>(*f2)) {
                 ASR::Variable_t* v = ASR::down_cast<ASR::Variable_t>(f2);
-                ASR::ttype_t* v_type = ASRUtils::type_get_past_pointer(v->m_type);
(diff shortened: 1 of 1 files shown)


certik:
One test was failing in GFortran, but I think the code might have been invalid, so I removed that commit. There might still be a failure in some third party tests, if so we need to investigate and fix it.

certik:
I fixed one third-party test failure, we'll see if there is another one.

src/libasr/pass/promote_allocatable_to_nonallocatable.cpp
```diff
@@ -60,7 +61,8 @@ class IsAllocatedCalled: public ASR::CallReplacerOnExpressionsVisitor<IsAllocate
 
         void visit_SubroutineCall(const ASR::SubroutineCall_t& x) {
             ASR::FunctionType_t* func_type = ASRUtils::get_FunctionType(x.m_name);
-            for( size_t i = 0; i < x.n_args; i++ ) {
+            size_t n = std::min(x.n_args, func_type->n_arg_types);
+            for( size_t i = 0; i < n; i++ ) {
```
certik, line 65: These two changes are suspect: I think this can be fixed later, but we should add strict checking to ASR verify that the type and arguments are equal and correspond to each other.

certik:
@jinangshah21 thank you! Much appreciated. Hopefully the CI will pass now.

Issue #9662, opened 2026-01-22
INQUIRE: iostat= and nextrec= not returning results


w6ws:
Follow-on to PR #9434:
```
program inquire2
  implicit none

! Test IOSTAT= and NEXTREC=

  integer :: ioerr, nxtrec, recl, unit_no

  inquire (iolength=recl) 42, 42.0, 'xyzzy'
  print *, 'iolength =', recl, ' :', merge ('pass', 'fail', recl == 13)

  ioerr = -42
  open (newunit=unit_no, file='random.data', status='unknown',  &
        access='direct', recl=recl, form='unformatted', iostat=ioerr)
  if (ioerr /= 0) error stop

  ioerr = -42
  nxtrec = -43
  inquire (unit_no, iostat=ioerr, nextrec=nxtrec)
  print *, 'iostat  =', ioerr,  ' :', merge ('pass', 'fail', ioerr == 0)
  print *, 'nextrec =', nxtrec, ' :', merge ('pass', 'fail', nxtrec == 1)

  close (unit_no, status='delete')

end program
```
```
$ lfortran inquire2.f90
iolength =    13     :    pass
iostat  =    -42     :    fail
nextrec =    -43     :    fail
$
```
```
$ gfortran inquire2.f90
$ ./a.out
 iolength =          13  :pass
 iostat  =           0  :pass
 nextrec =           1  :pass
$
```

w6ws:
While in there, here are a couple more undefined issues with `INQUIRE`:
```
program inquire4
  implicit none
  
! Test F2018 tidbits.  These were 'undefined' prior to F2018.

  integer :: irl

  irl = -99
  inquire (42, recl=irl)
  print *, 'unconnected recl=', irl, merge (': pass', ': fail', irl == -1)

  open (42, file='fort.42', access='stream')
  inquire (42, recl=irl)
  print *, 'stream recl=', irl, merge (': pass', ': fail', irl == -2)

  close (42, status='delete')

end program
```
```
$ lfortran inquire4.f90
unconnected recl=    -99    : fail
stream recl=    -99    : fail
$
```
```
$ gfortran inquire4.f90
$ ./a.out
 unconnected recl=          -1 : pass
 stream recl=          -2 : pass
$
```

certik:
The first code now prints:
```
$ lfortran a.f90
iolength =    13     :    pass
iostat  =    0     :    pass
nextrec =    1     :    pass
```

but the second code still prints:
```
[...]

PR #10410, opened 2026-03-03
fix: recl parameter in inquire for unconnected unit and stream access

HarshilShah1804:
Towards #9662


jinangshah21 (approved):
Looks Good!

PR #10425, opened 2026-03-03
Compiling neural-fortran Fix smod deserialization ice

jassem-manita:
building neural-fortran crashes at 74-75% :

```
  File "asr_deserialization_visitor.h", line 3543, in deserialize_StructInstanceMember()
    m_m = self().read_symbol();
LCompilersException: Deserialization failed: symbol '1_multihead_attention_layer_sequence_length'
references symbol table with ID 21 which has not been deserialized yet.
```
The crash happens whenever separately compiled files combine:
1. A utility module defining a struct with integer fields
2. A parent module declaring a function whose return array uses those struct fields as dimensions (`real :: output(t%rows, t%cols)`)
3. A submodule implementing that function
4. A main program that `use`s the parent module, which triggers loading the `.smod`

the root cause is that thte 5-arg `deserialize_asr()` in `serialization.cpp` never actually used its `external_symtab` parameter the name was commented out in the signature (`SymbolTable & /*external_symtab*/`) and the body just forwarded to the 4-arg version. So the deserializer always started with an empty symbol table and had no way to find symbols from previously compiled modules.

fix: 
- Added `find_symbol_in_tree()` helper that look up symbols in the external symbol table tree by name
- In `read_symbol()` before throwing on an unknown ID, try resolving the symbol by name with `find_symbol_in_tree()`
- Made the 5-arg `deserialize_asr()` pass `external_symtab` through instead of ignoring it
- Added integration tests that reproduces the crash

Neural-fortran : After the fix it jumps from 75 to 78%

### DIFF
--- a/src/libasr/serialization.cpp
+++ b/src/libasr/serialization.cpp
@@ -95,6 +95,30 @@ public:
 #endif
             DeserializationBaseVisitor(al, load_symtab_id, offset) {}
 
+    SymbolTable *external_symtab_root = nullptr;
+
+    ASR::symbol_t *find_symbol_in_tree(SymbolTable *symtab, const std::string &name) {
+        if (!symtab) return nullptr;
+        ASR::symbol_t *found = symtab->get_symbol(name);
+        if (found) return found;
+        for (auto &pair : symtab->get_scope()) {
+            ASR::symbol_t *sym = pair.second;
+            SymbolTable *child = nullptr;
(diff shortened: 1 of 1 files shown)


jassem-manita:
is there anything that i need to explain ?

jinangshah21:
Thanks, but already fixed in main, you can close this PR.

PR #10427, opened 2026-03-04
Fix several Fiats bugs

certik:


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7161,6 +7161,45 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        ASR::symbol_t* ext = ASR::down_cast<ASR::symbol_t>(
+                            ASR::make_ExternalSymbol_t(al, loc, current_scope,
+                                s2c(al, monomorphized_name), pdt_existing,
+                                ASRUtils::symbol_name(module_sym),
+                                nullptr, 0, s2c(al, monomorphized_name),
+                                ASR::accessType::Public));
+                        current_scope->add_symbol(monomorphized_name, ext);
+                        type_decl_sym = ext;
+                    }
+                }
```
jinangshah21, line 7193: I think we have utility function in ASRUtils for named something like `import_struct` which will make above changes to be single line. However this is fine, can be done in separate PR.

certik:
There is something wrong with the module/submodule changes in this PR, had to fix a lot of things due to that.

Let me try to remove those commits into a dedicated PR and fix it properly. The rest should merge.

src/lfortran/semantics/ast_common_visitor.h
certik: https://github.com/lfortran/lfortran/issues/10453

certik:
Finally everything passed. I merged it. We can fix the rest later.

Issue #4357, opened 2024-06-23
Nonzero iostat value from a read statemnt with no error, EOR or EOF.


harperjf:
This program reads an integer ```n``` but gives a wrong (nonzero) value of ```iostat``` in the read statement, and it is a different wrong value the next time the program runs with the same input. But the value of ```n``` printed by the program was correct. The program:
```
program readtest
  implicit none ! lfortran problem with iostat in read statement
  integer n,ios
  print *,'Enter an integer'
  read(*,*,iostat=ios) n
  if(ios>0) print *,'Error reading integer; iostat =',ios
  if(ios<0) print *,'Unexpected end of input; iostat =',ios
  print *,'You said',n
end program readtest
```
The results of two program runs with identical inputs in my Linux Ubuntu 22.04 system:
```
(lf) john:~/Test$ echo 666 | lfortran readtest.f90
Enter an integer
Error reading integer; iostat = 21884
You said 666
(lf) john:~/Test$ echo 666 | lfortran readtest.f90
Enter an integer
Error reading integer; iostat = 21871
You said 666
```

harperjf:
This program, which is a variant of readtest.f90, asks for ```iomsg``` to be printed when there is a nonzero value of ```iostat``` in  a ```read``` statement but lfortran produces an empty character string.
```
program readmsg
  implicit none ! try both iostat and iomsg in read statement
  integer n,ios
  character(200):: msg,fmt = "(A,I0,1X,A,'""',A,'""')"
  print *,'Enter an integer'
  read(*,*,iostat=ios,iomsg=msg) n
  if(ios>0) print fmt,'Error reading integer; ios=',ios,'msg=',trim(msg)
  if(ios<0) print fmt,'Unexpected end of input; ios=',ios,'msg=',trim(msg)
  print *,'You said',n
end program readmsg
```
Results from 3 runs of that program: the first should have been error-free, the others should have printed non-empty values of ```msg```. 
```
(lf) john:~/Test$ lf readmsg.f90
Enter an integer
666
Error reading integer; ios=1096621737 msg=""
You said 666
(lf) john:~/Test$ lf readmsg.f90
Enter an integer
2222222222222222222222222222222222222
Error reading integer; ios=469445769 msg=""
You said -1
(lf) john:~/Test$ lf readmsg.f90
Enter an integer
x
Unexpected end of input; ios=-1282212455 msg=""
You said 32766
(lf) john:~/Test$
```

PR #10433, opened 2026-03-04
Fix iostat and add iomsg support for read statements

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4357

### DIFF
--- a/integration_tests/read_50.f90
+++ b/integration_tests/read_50.f90
@@ -1,0 +1,66 @@
+program read_50
+    implicit none
+    integer :: n, ios
+    real :: r
+    character(len=20) :: str
+    character(len=100) :: msg
+
+    ! Test 1: successful read from string, iostat should be 0
+    ios = 999
+    str = "42"
+    read(str, *, iostat=ios) n
+    if (ios /= 0) error stop
+    if (n /= 42) error stop
+
+    ! Test 2: successful read of real from string
+    ios = 999
+    str = "3.14"
+    read(str, *, iostat=ios) r
+    if (ios /= 0) error stop
+    if (abs(r - 3.14) > 0.001) error stop
+
+    ! Test 3: successful read from file
+    ios = 999
+    open(unit=10, file="read_50_test.txt", status="replace", action="write")
+    write(10, '(I5)') 666
+    close(10)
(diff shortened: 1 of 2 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -7676,6 +7676,26 @@ LFORTRAN_API void _lfortran_empty_read(int32_t unit_num, int32_t* iostat) {
     }
 }
 
+LFORTRAN_API void _lfortran_set_read_iomsg(int32_t iostat, char* iomsg, int64_t iomsg_len) {
+    if (iomsg == NULL || iomsg_len <= 0 || iostat == 0) return;
```
assem2002, line 7680: I think it's good idea having a function responsible of iomsg based on `iostat`.
I think a further improvement is a `const char*[]` with index being `iostat` and value is the messge, To fixate the error message throughout the codebase (to use it whether iostat is provided or not).

integration_tests/read_50.f90
```diff
@@ -0,0 +1,66 @@
...
+    if (ios /= 0) error stop
+    if (n /= 77) error stop
+
+    ! Test 6: read error with iomsg, msg should contain error description
+    ios = 999
+    msg = ""
+    str = "xyz"
+    read(str, *, iostat=ios, iomsg=msg) n
+    if (ios <= 0) error stop
+    if (len_trim(msg) == 0) error stop
```
assem2002, line 53: Should we test `msg`?

HarshitaKalani:
Thanks @assem2002! I'll address the suggestions with a new PR.

src/libasr/runtime/lfortran_intrinsics.c
HarshitaKalani: For the lookup table idea, a plain const char*[] indexed by iostat won't work directly since the values aren't contiguous (-2, -1, 1, 5001-5006, etc.), but we could use a struct array or switch-case to map iostat codes to messages.

assem2002: I see. I think the error numbers are compiler specific, but still the end-of-file is indicated by negative values, so yeah we'll face some inconsistencies.

Issue #3900, opened 2024-04-18
ICE from -g --debug-with-line-column


harperjf:
The program below deals correctly with an allocatable kind=c_char scalar if no options are specified but lfortran complains if -g is used and gives an ICE if --debug-with-line-column is also used. The semantic error in the output seems hard to reconcile with the error-free output when no options were used. Evidence:
```
(lf2) john:~$ cat msgfromc.f90
program msgfromc
  use,intrinsic :: iso_c_binding, only: c_char
  implicit none
  character(kind=c_char,len=:), allocatable:: msg
  call hello(msg)
  print *,msg
contains
  subroutine hello(msg) bind(c)
    use, intrinsic :: iso_c_binding, only: c_char
    character(kind=c_char,len=:), allocatable, intent(out) :: msg
    msg = "Hello from Fortran"
  end subroutine hello
end program msgfromc
(lf2) john:~$ lfortran msgfromc.f90
Hello from Fortran
(lf2) john:~$ lfortran -g msgfromc.f90
semantic error: The `emit_debug_line_column` is not enabled; please use the `--debug-with-line-column` option to get the correct location information


Note: Please report unclear or confusing messages as bugs at
https://github.com/lfortran/lfortran/issues.
(lf2) john:~$ lfortran -g --debug-with-line-column msgfromc.f90
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x726858
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29e3f
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29d8f
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x600a58
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x7a0a5c
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x78ded9
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x85f731
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x1c0d4b8
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x1c06968
  Binary file "/home/john/conda_root/envs/lf2/bin/lfortran", local address: 0x1c04df5
[...]

PR #10434, opened 2026-03-04
fix: handle allocatable character(kind=c_char) in LLVM codegen

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/3900

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -2025,7 +2025,7 @@ public:
         // Avoid touching the descriptor if the array is not allocated
         {
             llvm::Type* index_type = arr_descr->get_index_type();
-            llvm::Value* size_var = llvm_utils->CreateAlloca(*builder, index_type, nullptr, "arg_array_size_tmp");
+            llvm::Value* size_var = llvm_utils->CreateAlloca(index_type, nullptr, "arg_array_size_tmp");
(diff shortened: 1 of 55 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -16885,6 +16885,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                         ASRUtils::expr_type(x.m_args[i].m_value)))) ) {
                 this->visit_expr_wrapper(x.m_args[i].m_value, true);
+            } else if (ASR::is_a<ASR::StringPhysicalCast_t>(*x.m_args[i].m_value) &&
+                       orig_arg && ASRUtils::is_allocatable(orig_arg->m_type)) {
+                ASR::StringPhysicalCast_t* cast =
+                    ASR::down_cast<ASR::StringPhysicalCast_t>(x.m_args[i].m_value);
+                if (cast->m_old == ASR::string_physical_typeType::DescriptorString &&
+                    cast->m_new == ASR::string_physical_typeType::CChar) {
+                    this->visit_expr_load_wrapper(cast->m_arg, 0);
+                    tmp = llvm_utils->get_string_data(
```
jinangshah21: I think this should be done in `visit_StringPhysicalCast` only we don't need to add special handling here

HarshitaKalani: Done.

PR #10440, opened 2026-03-05
fix: set nonzero iostat when reading non-integer string with 'I' format

### LINKED ISSUE #10438: Error not found when reading non-integer with I format
This program prints one line with gfortran, g95, AOCC flang and ifx but three with lfortran because lfortran gave ```ios``` a zero value when reading a non-integer string with ```I``` format.
```
  implicit none
  integer i,n,ios
  character(4):: input(3) = ['42  ','3.14','abcd']
  
  do i = 1,3
     read(input(i),"(I4)",iostat=ios) n
     if (ios/=0) exit
     print "(A,I0)",'You said ',n
  end do
end program
```
Wrong output from lfortran 0.60.0
```
You said 42
You said 3
You said 0
```
Correct output from all the other compilers
```
You said 42
```

opixdown:
fixes #10438


HarshitaKalani:
You'll need to resolve the conflict.

Issue #6801, opened 2025-03-31
pure keyword ignored


harperjf:
The F2023 standard C15107 says
```
A pure subprogram shall not contain a print-stmt, open-stmt, close-stmt, backspace-stmt, endfile-stmt,
rewind-stmt, flush-stmt, wait-stmt, or inquire-stmt.
```
This program disobeys the standard by printing in a pure subroutine. Gfortran and ifx detect the error, Lfortran does not. 
```
program badpure
  implicit none
  call sub(666)
contains
  pure subroutine sub(  b)
    integer,intent(in)::b
    print "(A,I0)",'b = ',b
  end subroutine sub
end program badpure
```
Gfortran 14.2.0 output, which is correct:
```
badpure.f90:7:27:

    7 |     print "(A,I0)",'b = ',b
      |                           1
Error: PRINT statement at (1) not allowed within PURE procedure
```
Lfortran 0.50.0 output (the same with or without --std=f23 option) is wrong:
```
b = 666
```

PR #10448, opened 2026-03-05
Enforce F2023 C15107: disallow I/O statements in pure procedures

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6801

### DIFF
--- a/src/lfortran/tests/test_llvm.cpp
+++ b/src/lfortran/tests/test_llvm.cpp
@@ -378,7 +378,7 @@ end function)";
     LCompilers::LocationManager lm;
     LCompilers::ASR::TranslationUnit_t* asr = TRY(LCompilers::LFortran::ast_to_asr(al, *tu,
         diagnostics, nullptr, false, compiler_options, lm));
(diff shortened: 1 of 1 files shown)


certik:
I am not sure if we want to fix it this way. Rather, let's implement https://github.com/lcompilers/lpython/issues/1266. Then for each function call and subroutine call (as well as IO) we add a function `is_deterministic` and `is_side_effect_free` and give it the function or IO node. In Fortran, pure functions are side-effect-free and simple functions are both side-effect-free and deterministic. The requirement is that every statement / function that they call must also be side-effect-free. So you simply call `is_side_effect_free` on each node that you call, and if not, give an error.

Also a frequently requested feature is to be able to turn these into just warnings, so that you *can* print from pure functions for debugging.

@HarshitaKalani do you want to give it a shot to implement a proper side-effect-free and deterministic analysis and enforce it for pure and simple functions?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -90,6 +90,33 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        Level::Error, Stage::Semantic, {
+                            Label("", {loc})
+                        }));
+                    throw SemanticAbort();
+                }
+                break;
+            }
+            scope = scope->parent;
+        }
+    }
```
certik, line 118: I don't like this approach, as I mentioned. I think we need to have a loop that checks every statement from the "caller".

certik: Maybe this is the only way to make it work, but I would like to see first the solution with user pure/impure functions and how that works. Then integrate the IO checks and other nodes into the same framework.

HarshitaKalani: Continued this at: https://github.com/lfortran/lfortran/pull/10476

HarshitaKalani:
closing this as continued with https://github.com/lfortran/lfortran/pull/10476

PR #10467, opened 2026-03-06
fix: struct member shadowing outerscope kind parameter in visit_Real

### LINKED ISSUE #10445: Compiler bug when initializing structure component with same name as kind
## MRE

```Fortran
module t
    integer, parameter :: q = selected_real_kind(10)
    type ty
        real(q) :: q = 1._q
    end type
end module
```

## lfortran

```Bash
lfortran -c tmp.f90
semantic error: Variable 'q' is not constant
 --> tmp.f90:4:24
  |
4 |         real(q) :: q = 1._q
  |                        ^^^^
```

## Other compilers

Both `gfortran -c tmf.f90` and `ifx -c -std18 tmp.f90` produce no warnings or errors.

opixdown:
fixes #10445


certik:
You need to add a test.

integration_tests/struct_type_16.f90
```diff
@@ -0,0 +1,15 @@
+module test_struct_kind_shadow
+    implicit none
+    integer, parameter :: q = selected_real_kind(10)
+    type ty
+        real(q) :: q = 1._q
```
certik, line 5: Will the same issue happen with integer, character and other types?

opixdown: Tested all types  real, complex, character, and logical all work correctly with this fix. Only integer still needs to be fixed.

certik: Do we have a test for the other ones? We need to have a test, otherwise it will get broken in the future.

opixdown: added test for all types

integration_tests/struct_type_16.f90
```diff
@@ -0,0 +1,15 @@
...
+    end type
+end module
+
+program test
+    use test_struct_kind_shadow
+    implicit none
+    type(ty) :: t
+    if (t%q /= 1.0) error stop "value incorrect"
+    print *, "PASS"
+end program test
```
certik, line 15: Do all these need to be separate tests? It adds to our test runtime, which we want to minimize.

certik:
@opixdown Can you please consolidate the tests in a subsequent PR? Ensure that the consolidated test fails before this PR was merged and works now.

I made this https://github.com/lfortran/lfortran/issues/10634.

PR #10476, opened 2026-03-07
Implement deterministic and side_effect_free analysis for functions

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6801
continued: https://github.com/lfortran/lfortran/pull/10448


HarshitaKalani:
The CI fails for the following MRE:
```fortran
program pure_side_effects_02
    implicit none
    integer, allocatable :: a(:), b(:)

    allocate(a(3))
    a = [1, 2, 3]
    call pure_move(a, b)
    if (size(b) /= 3) error stop
    if (b(1) /= 1) error stop
    if (b(2) /= 2) error stop
    if (b(3) /= 3) error stop
    if (allocated(a)) error stop

contains

    pure subroutine pure_move(src, dst)
        integer, allocatable, intent(inout) :: src(:)
        integer, allocatable, intent(out) :: dst(:)
        call move_alloc(src, dst)
    end subroutine pure_move

end program pure_side_effects_02
```
This is because move_alloc is a pure intrinsic subroutine per the Fortran standard and must be allowed inside PURE procedures.  Currently LFortran represents it as an IntrinsicImpureSubroutine ASR node, and the SideEffectFinder
 in ast_body_visitor.cpp unconditionally rejects every such node inside a PURE procedure.  The fix needs a way for the SideEffectFinder to distinguish genuinely impure intrinsics from pure ones (move_alloc, mvbits) without introducing a dependency from the semantics layer on the pass layer (where the IntrinsicImpureSubroutines enum lives).

@certik could you please have a look at the changes in this PR and let me know if this is in the right direction? I'm also trying to find a way to solve this issue in the meantime. Any inputs would be appreciated. Thanks!

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -19,6 +19,119 @@
...
+                    return;
+                }
+            }
+        }
+        BaseWalkVisitor::visit_SubroutineCall(x);
+    }
+
+    void visit_IntrinsicImpureSubroutine(const ASR::IntrinsicImpureSubroutine_t &x) {
+        if (found) return;
+        mark_found(x.base.base.loc, "Call to impure intrinsic subroutine");
```
certik: Here you can special case move_alloc and other subroutines that are actually "pure". Later we'll move them to some other ASR node, but for now just hardwiring their knowledge here should be ok.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -19,6 +19,119 @@
 
 namespace LCompilers::LFortran {
 
+class SideEffectFinder : public ASR::BaseWalkVisitor<SideEffectFinder> {
```
certik: I would move this into a separate file, as we also want to use it in ASR verify to ensure that ASR is always correct (even between passes, etc.) regarding side effects.

certik:
Yes, I think this is a good way. Let's focus on side effects only (we'll do determinism later), and get the "checker" working, by hardwiring what you need. Pass all tests.

src/libasr/asr_side_effect.h
```diff
@@ -0,0 +1,113 @@
+#ifndef LIBASR_ASR_SIDE_EFFECT_H
+#define LIBASR_ASR_SIDE_EFFECT_H
+
+#include <libasr/asr_utils.h>
+#include <libasr/pass/intrinsic_subroutines.h>
```
HarshitaKalani, line 5: This header file is quite heavy. A lighter approach can be to just use the raw integer ids in is_side_effect_free_intrinsic_impure_subroutine directly and avoid the heavy include entirely. So MoveAlloc = 11 and Mvbits = 12 (0-indexed). Do you think that would be fine?

certik: I think this can be done later, or is there an immediate downside of the current approach?

HarshitaKalani: it's fine for now

HarshitaKalani:
Yep, we can merge this. What are the next steps here?

certik:
Let's merge this after https://github.com/lfortran/lfortran/pull/10529, to lock in the win first.

certik:
Are now both labels "deterministic" and "side-effect-free" correctly determined and tested?

HarshitaKalani:
Yes. 
I've verified the updated references. The existing reference test callback_02 (registered with asr = true) covers both the true and false cases for the labels: its nested subroutine f (pure math) correctly gets .true./.true., while cb (has Print) and foo (calls impure cb) correctly stay .false./.false.. The interface declaration also correctly gets .false./.false.. The bulk of the updated reference tests are simple Implementation functions with no I/O, which correctly changed from .false./.false. to .true./.true.. The error test in continue_compilation_2 also verifies that pure procedures with side effects are rejected.

certik:
Ok, let's merge it!

PR #10483, opened 2026-03-07
Get Fiats working

certik:
All tests pass. Tests added to CI.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1441,7 +1441,7 @@ RUN(NAME intrinsics_354 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) #
 RUN(NAME intrinsics_355 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # system_clock
 RUN(NAME intrinsics_356 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME intrinsics_357 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME intrinsics_358 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # getseed
+RUN(NAME intrinsics_358 LABELS gfortran llvm) # getseed
 RUN(NAME intrinsics_359 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # pack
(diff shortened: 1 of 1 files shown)


certik:
It should pass now.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -10325,14 +10348,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        // are allocated via alloca each time the block executes; without
+        // stackrestore, an associate block inside a loop leaks stack
+        // every iteration.
+#if LLVM_VERSION_MAJOR >= 18
+        llvm::Value *saved_stack = builder->CreateStackSave();
+#else
+        llvm::Function *stacksave_fn = llvm::Intrinsic::getDeclaration(
+            module.get(), llvm::Intrinsic::stacksave);
+        llvm::Value *saved_stack = builder->CreateCall(stacksave_fn);
+#endif
```
certik, line 10362: We should rework this later on, but I think it's not a blocker to merge this PR.

Once we start working on leaks, I think we should allocate all variables at the beginning of the function.

ci/test_third_party_codes.sh
```diff
@@ -137,6 +137,19 @@ time_section "🧪 Testing splpak" '
...
 
+time_section "🧪 Testing Fiats" '
+  git clone https://github.com/BerkeleyLab/fiats
+  cd fiats
+  export PATH="$(pwd)/../src/bin:$PATH"
+  micromamba install -c conda-forge fpm
+
+  git checkout 0a2ff33cf8c06c6379d8e8883e846577a29f2f5e
+  fpm build --compiler=lfortran --flag --cpp --flag --separate-compilation --flag --realloc-lhs-arrays
+
```
jinangshah21, line 148: We are not running tests here?

jinangshah21 (approved):
Looks Good! Great achievement.

ci/test_third_party_codes.sh
certik: No, because they are flaky at the CI. They run locally, but not always and I thought I fixed it, but not quite. This way we lock in a partial win and unblock other PRs and now we need to investigate what is going on.

certik: Should be fixed in https://github.com/lfortran/lfortran/pull/10488.

Issue #8495, opened 2025-08-30
bug: LFortran incorrectly allows (real, imag) as complex literal


HarshitaKalani:
```fortran
program cmplx_02
    complex :: a = (real(1, kind=4), 0.0)
    print *, a
end program
```
```console
% lfortran c.f90
(1.00000000e+00, 0.00000000e+00)
% gfortran c.f90 && ./a.out
c.f90:2:36:

    2 |     complex :: a = (real(1, kind=4), 0.0)
      |                                    1
Error: Expected a right parenthesis in expression at (1)
```

PR #10485, opened 2026-03-07
Reject non-constant expressions in complex literal parts

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/8495

### DIFF
--- a/tests/reference/asr-continue_compilation_2-a6145a1.json
+++ b/tests/reference/asr-continue_compilation_2-a6145a1.json
@@ -8,6 +8,6 @@
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_2-a6145a1.stderr",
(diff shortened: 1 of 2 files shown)


HarshilShah1804 (approved):
Looks Good!

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -15905,7 +15905,37 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         tmp = ASR::make_RealConstant_t(al, x.base.base.loc, r, type);
     }
 
+    bool is_valid_complex_literal_part(const AST::expr_t &expr) {
+        if (AST::is_a<AST::Num_t>(expr)) return true;
+        if (AST::is_a<AST::Real_t>(expr)) return true;
+        if (AST::is_a<AST::Name_t>(expr)) return true;
```
HarshilShah1804, line 15911: Maybe we should also check here that the variable is also a compile-time constant. But we can do that in a subsequent PR.

Issue #10359, opened 2026-02-28
Segfaults when running do-loops for very large subroutines with allocatable variables & subroutine calls


YashNagda17:
Towards #5629

MRE:
```fortran
program problem_12_standalone

    implicit none
    integer :: isum, i, j
    integer,dimension(:,:),allocatable :: price2
    integer,dimension(:,:,:),allocatable :: perimeter

    integer, parameter :: NTOTAL = 200  
    allocate(price2(NTOTAL, NTOTAL))

    if (.not. allocated(perimeter)) then
        allocate(perimeter(4, NTOTAL, NTOTAL)); perimeter = 0
    end if

    do i = 1, NTOTAL
        do j = 1, NTOTAL
            call go()
            price2(i,j) = compute_n_sides()
            print *,"i", i, j
        end do
    end do
    write(*,*) '12b:', sum(price2)

    contains
        recursive subroutine go()
            perimeter = 1     
        end subroutine go

        function compute_n_sides() result(iper)
            integer :: iper
            iper = count(perimeter==1)
        end function compute_n_sides

end program problem_12_standalone
```
Changing NTOTAL to even higher values again causes this segfault, although exact loop iteration might change

Main:
```
$lfortran a.f90
[...]
i    187    171
i    187    172
Segmentation fault (core dumped)
```

YashNagda17:
Summary from running Valgrind:
```
187    180
==202914== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==202914== 
==202914== Process terminating with default action of signal 11 (SIGSEGV)
==202914==  Access not within mapped region at address 0x1FFE801FF8
==202914== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==202914==    at 0x49E4D83: memcpy (string_fortified.h:29)
==202914==    by 0x49E4D83: __printf_buffer_write (Xprintf_buffer_write.c:39)
==202914==  If you believe this happened as a result of a stack
==202914==  overflow in your program's main thread (unlikely but
==202914==  possible), you can try to increase the size of the
==202914==  main thread stack using the --main-stacksize= flag.
==202914==  The main thread stack size used in this run was 8388608.
==202914== 
==202914== HEAP SUMMARY:
==202914==     in use at exit: 801,334 bytes in 13 blocks
==202914==   total heap usage: 635,483 allocs, 635,470 frees, 197,383,471,055 bytes allocated
==202914== 
==202914== LEAK SUMMARY:
[...]

PR #10511, opened 2026-03-09
code-refactor: optimize llvm backend to reduce memory leaks inside do-loops

YashNagda17:
Towards #10359

Here, in this PR; Changes are made to llvm backend to initialize few allocations at function entry, rather than function body.

This reduces the stack allocations, particularly helpful inside do-loops... As the temporaries allocated are reused in each iteration.
In current main, these temporaries are allocated inside each loop.

Each commit represents a separate allocation optimization


YashNagda17:
After these optimizations, stack usage was reduced to around 3.1MB from 8MB (stack overflow limit) for program mentioned in #10359 (With same bounds)

certik (approved):

YashNagda17:
Some issue with Fortran_MPI, I shall debug it offline and push the correct fix

PR #10517, opened 2026-03-09
iso_c_binding

nilocj:
Additional named constants to lfortran_intrinsic_iso_c_binding.f90. Integration tests passed. Before this fix, I coud not build https://github.com/scivision/ffilesystem with lfortran.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -6,7 +6,6 @@ echo "Running SHELL"
 
 echo "CONDA_PREFIX=$CONDA_PREFIX"
 llvm-config --components
-LFORTRAN_CMAKE_GENERATOR=Ninja
 # Generate the `version` file
 bash ci/version.sh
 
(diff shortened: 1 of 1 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -2643,6 +2643,7 @@ RUN(NAME kwargs_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME test_iso_c_binding LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME test_iso_c_binding_constants LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
+RUN(FAIL NAME test_iso_c_binding_constants_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
```
certik: Instead of this, just add a test that checks the new constants that you added into iso_c_binding.

certik:
Now you need to update reference tests via `./run_tests.py -u`.

PR #10538, opened 2026-03-11
Fix class(*) bugs for petaca

certik:
Refactor `class(*), allocatable :: x(:)` to use one wrapper for the whole array, not each element individually, because all the elements must have exactly the same type.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -440,11 +440,7 @@ public:
 
         // Allocate zero-initialized contiguous data buffer
         llvm::Value* total_bytes = builder->CreateMul(num_elems_64, elem_size);
-        llvm::Value* data_buf = LLVMArrUtils::lfortran_malloc(
-            context, *module, *builder, total_bytes);
-        builder->CreateMemSet(data_buf,
-            llvm::ConstantInt::get(context, llvm::APInt(8, 0)),
-            total_bytes, llvm::MaybeAlign());
+        llvm::Value* data_buf = llvm_utils->allocate_zeroed_bytes(total_bytes);
 
         // Store data in the ACTUAL ONE-wrapper (descriptor.field0)
         llvm::Value* wrapper_ptr_ptr = arr_descr->get_pointer_to_data(
--- a/src/libasr/codegen/llvm_array_utils.cpp
+++ b/src/libasr/codegen/llvm_array_utils.cpp
@@ -984,31 +984,20 @@ namespace LCompilers {
                                 // Allocate temp wrapper on stack
                                 llvm::Value* temp_wrapper = llvm_utils->CreateAlloca(*builder, class_type);
 
+                                // Store vptr in field 0 (same for both cases)
+                                builder->CreateStore(vptr, llvm_utils->create_gep2(class_type, temp_wrapper, 0));
+
                                 // If vptr is non-NULL, compute element via byte offset
                                 // If vptr is NULL (empty wrapper for assignment target),
-                                // return wrapper with NULL data (assignment will initialize it)
-                                llvm::BasicBlock* vptr_ok = llvm::BasicBlock::Create(
-                                    context, "vptr_ok");
-                                llvm::BasicBlock* vptr_null = llvm::BasicBlock::Create(
-                                    context, "vptr_null");
-                                llvm::BasicBlock* merge = llvm::BasicBlock::Create(
-                                    context, "merge_elem");
-                                builder->CreateCondBr(vptr_is_null, vptr_null, vptr_ok);
-
-                                llvm_utils->start_new_block(vptr_ok);
-                                llvm::Value* element_ptr_i8 = llvm_utils->get_polymorphic_array_data_ptr(data_ptr, idx, vptr);
-                                builder->CreateStore(vptr, llvm_utils->create_gep2(class_type, temp_wrapper, 0));
-                                builder->CreateStore(element_ptr_i8, llvm_utils->create_gep2(class_type, temp_wrapper, 1));
-                                builder->CreateBr(merge);
-
-                                llvm_utils->start_new_block(vptr_null);
-                                builder->CreateStore(vptr, llvm_utils->create_gep2(class_type, temp_wrapper, 0));
-                                builder->CreateStore(
-                                    llvm::ConstantPointerNull::get(llvm::cast<llvm::PointerType>(data_field_type)),
-                                    llvm_utils->create_gep2(class_type, temp_wrapper, 1));
-                                builder->CreateBr(merge);
-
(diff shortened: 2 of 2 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -17955,21 +18072,103 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        vptr = struct_api->get_intrinsic_type_vptr(arg_type,
+                            ASRUtils::extract_kind_from_ttype_t(arg_type), module.get());
+                    }
+
+                    // Loop: fill each wrapper element with {vptr, &data[i]}
+                    llvm::BasicBlock *loophead = llvm::BasicBlock::Create(context, "poly_wrap.head");
+                    llvm::BasicBlock *loopbody = llvm::BasicBlock::Create(context, "poly_wrap.body");
+                    llvm::BasicBlock *loopend = llvm::BasicBlock::Create(context, "poly_wrap.end");
+
+                    llvm::Value* idx = builder->CreateAlloca(llvm::Type::getInt32Ty(context));
```
jinangshah21, line 18141: I think here logic seems incorrect as it copies each element with wrapper, while it should have only one class wrapper and store whole array in data pointer of that wrapper.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -405,6 +405,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        ptr_loads = LLVM::is_llvm_pointer(*target_arr_type);
+        if (ASRUtils::is_unlimited_polymorphic_type(target_arr_type) && ptr_loads > 0) {
+            ptr_loads--;
+        }
+        this->visit_expr_wrapper(target_ai->m_v, true);
+        ptr_loads = saved_ptr_loads;
+        llvm::Value* target_arr = tmp;
+        tmp = saved_tmp;
+
+        llvm::Type* arr_desc_type = arr_descr->get_array_type(
```
jinangshah21, line 430: This allocation during assigning `ArrayItem` of class(*) array not seems got to me becasue, consider below example:
```fortran
subroutine set_value(this, value)
        class(container), intent(out) :: this
        class(*), intent(in) :: value(:)
        this%value = value
    end subroutine
````
After all ASR passes:
```fortran
subroutine set_value(this, value)
    integer(4) :: __do_loop_end
    class(container), intent(out) :: this
    class(~unlimited_polymorphic_type), dimension(:), intent(in) :: value
    if (Allocated(this%value)) then
        deallocate(this%value)
    end if
    reallocate(this%value(size(value, 1)))   !!!! This realloction should handle allocation of lhs arrays
    ! CheckBounds LHS = this%value, RHS = value, 
    __libasr_index_0_1 = lbound(value, 1)
    __do_loop_end = ubound(this%value, 1)
    __libasr_index_0_ = lbound(this%value, 1) - 1
    do while (__libasr_index_0_ + 1 <= __do_loop_end)
        this%value(__libasr_index_0_) = value(__libasr_index_0_1)
    end do
end subroutine set_value
```

During reallocate we need to use type size from RHS `class(*)` array which is currently missing in reallocation.

 `this%value(__libasr_index_0_) = value(__libasr_index_0_1)` this line should copy data pointer and vptr (if needed) not reallocate.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8528,13 +8572,36 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                        // ONE-wrapper layout: extract data pointer from wrapper field 1.
+                                        // For unlimited polymorphic, data is already contiguous.
+                                        if (ASRUtils::is_unlimited_polymorphic_type(value_type)) {
+                                            loaded_data_ptr = llvm_utils->CreateLoad2(
+                                                llvm_utils->i8_ptr,
+                                                llvm_utils->create_gep2(value_el_type, loaded_data_ptr, 1));
+                                            // Character arrays: the ONE-wrapper data is an array of N
+                                            // individual string_descriptors (from the copy function),
+                                            // but character(*) access expects a single descriptor with
+                                            // a flat contiguous buffer. Consolidate here.
```
jinangshah21, line 8584: Here also for character arrays in `class(*)`, I think we only have to allocate only one class wrapper whose data pointer will have only one `string_descriptor` whose data pointer will have character array as we handle normal character arrays.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9327,27 +9397,96 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                }
+                llvm::Value* data_field_ptr = llvm_utils->create_gep2(target_llvm_type, target, 1);
+                llvm::Value* data_ptr = llvm_utils->CreateLoad2(llvm_utils->i8_ptr, data_field_ptr);
+                llvm::Value* is_data_null = builder->CreateICmpEQ(data_ptr,
+                    llvm::ConstantPointerNull::get(
+                        llvm::cast<llvm::PointerType>(data_ptr->getType())));
+                llvm_utils->create_if_else(is_data_null, [&]() {
+                    llvm::DataLayout data_layout(module->getDataLayout());
+                    uint64_t sz = data_layout.getTypeAllocSize(value_llvm_type);
+                    llvm::Value* alloc_size = llvm::ConstantInt::get(
```
jinangshah21, line 9482: Here also we should not allocate while array element assignment because polymorphic arrays can never be assigned using its `ArrayItem`, it is assigned as whole array, but in ASR we break it down into arrayItem with loops (maybe we should not do that at ASR and directly handle array assignments at backend for polymorphic arrays.

```fortran
module class_131_mod
    implicit none
contains
    subroutine set_value(this, value)
        class(*), allocatable, intent(out) :: this(:)
        class(*), intent(in) :: value(:)
        this(1) = value(1)
    end subroutine
end module

program test
    use class_131_mod
end program
```

```console
(lf) jinang_shah@JNSLAP:~/Desktop/lfortran$ gfortran c.f90
c.f90:7:8:

    7 |         this(1) = value(1)
      |        1
Error: Nonallocatable variable must not be polymorphic in intrinsic assignment at (1) - check that there is a matching specific subroutine for ‘=’ operator
```

src/libasr/codegen/asr_to_llvm.cpp
certik: This code is not there, see the last few commits that refactored it.

jinangshah21:
I think we can merge this as is for now and later we can address these comments for better fix.

certik:
See https://github.com/lfortran/lfortran/issues/10542 for the follow up work.

PR #10558, opened 2026-03-11
 initial implementation for memory debugger

assem2002:
We can test it by using any global variable (module variable that uses heap allocation) -- these leak for now.

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -4,7 +4,6 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include "mem_debug.h"
 #include "hashtable.h"
 #include <math.h>
 #include <complex.h>
@@ -94,6 +93,21 @@ void* dbg_malloc(void *context, int64_t size) {
     return ptr;
 
 }
+void dbg_free(void *context, void *ptr) {
+    (void)context;
+    if (!ptr) return;
+    if(!mem_dbg_hashTable.buckets){
+        fprintf(stderr, "Error: at dbg_free -- hastable not initialized\n");
+        exit(1);
+    }
+    Alloc_info *found = mem_debugger_get(&mem_dbg_hashTable, ptr);
+    if (!found) {
+        fprintf(stderr, "Error: at dbg_free -- ptr not found\n");
+        exit(1);
+    }
+    free(ptr);
+    mem_debugger_remove(&mem_dbg_hashTable, ptr);
+}
 
 void* dbg_calloc(void *context, int64_t nmemb, int64_t size) {
     (void)context;
@@ -145,21 +159,6 @@ void  *dbg_realloc(void *context, void *ptr, int64_t size){
 
 
 
-void dbg_free(void *context, void *ptr) {
-    (void)context;
-    if (!ptr) return;
-    if(!mem_dbg_hashTable.buckets){
-        fprintf(stderr, "Error: at dbg_free -- hastable not initialized\n");
-        exit(1);
-    }
-    Alloc_info *found = mem_debugger_get(&mem_dbg_hashTable, ptr);
-    if (!found) {
-        fprintf(stderr, "Error: at dbg_free -- ptr not found\n");
-        exit(1);
-    }
-    free(ptr);
-    mem_debugger_remove(&mem_dbg_hashTable, ptr);
-}
 
 // Called to report any leaks
 void dbg_report() {


src/libasr/runtime/mem_debug.h
```diff
@@ -0,0 +1,25 @@
...
+
+/*  Macro overrides — only when MEMORY_DEBUG is defined AND we are not    */
+/*  inside the debugger's own implementation file.                     */
+
+#if defined(MEMORY_DEBUG) && !defined(MEM_DEBUG_IMPL)
+#  define malloc(s)      dbg_malloc (s)
+#  define calloc(n, s)   dbg_calloc (n, s)
+#  define realloc(p, s)  dbg_realloc(p, s)
+#  define free(p)        dbg_free   (p)
+#endif
```
certik: Rename internal malloc/free to `internal_malloc`, `internal_free`, to make it clear those pointer should never be returned to the compiler.

Things that should be returned to the compiler should be reworked, so that we do the allocation inside `asr_to_llvm.cpp`. I think we need to pass an "allocator" that is implemented in asr_to_llvm.

src/libasr/runtime/mem_debug.c
```diff
@@ -0,0 +1,131 @@
...
+    while (*node) {
+        if ((*node)->ptr == ptr) {
+            Alloc *found = *node;
+            *node = found->next;
+            free(found); // Free `found` node
+            free(ptr);  // Free allocated memory
+            return;
+        }
+        node = &((*node)->next);
+    }
```
certik: I think we should be using a hashtable, or a vector, not a linked list, that's very slow.

Also I think we need to to check that the pointer `ptr` is found there at all, otherwise give an error.

assem2002:
I used the suggested[ hashtable implementation](https://github.com/certik/wasi/blob/88e936eb8c21a2e08ecb90ae7a6dfe1aa540875d/base/hashtable.h); I just added a remove functionality.


`--detect-leaks` now works for compiler's side of the memory -- it's off by default.

assem2002:
with the flag ON we get this number of leak errors `91% tests passed, 253 tests failed out of 2922`

src/libasr/runtime/mem_debug.c
```diff
@@ -0,0 +1,140 @@
...
+    if (leaks){
+        fprintf(stderr, "\n----------------------------------------------------\n");
+        fprintf(stderr, "TOTAL LEAKS FOUND : %zu, %zu bytes total\n", leaks, total_bytes);
+        fprintf(stderr, "----------------------------------------------------\n");
+        exit(1);
+    }
+    else {
+        fprintf(stdout, "NO LEAKS FOUND\n");
+    }
+}
```
certik: Given how short the file is, I would keep everything in one file `lfortran_intrinsics.c` to ease distribution / compilation.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4408,15 +4420,19 @@ LFORTRAN_API void* _lfortran_string_malloc_alloc(lfortran_allocator_t* al, int64
     if(length < 0) lfortran_error("Allocating string with length < 0");
     return ALLOCATOR_ALLOC(al, MAX(length, 1));
 }
-
+// Same as above but uses dbg_malloc
+LFORTRAN_API void* _lfortran_string_dbg_malloc(int64_t length) {
+    if(length < 0) lfortran_error("Allocating string with length < 0");
+    return ALLOCATOR_ALLOC(&compiler_mem_dbg_allocator, MAX(length, 1));
```
certik: This function should accept the allocator as an argument, not use it internally.

assem2002: I think I can even revert this change. And use the original one

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -48,6 +49,144 @@ typedef enum {
...
+    if (size == 0) {
+        dbg_free(context, ptr);
+        return NULL;
+    }
+
+    size_t alloc_size = (size_t)size;
+    void *new_ptr = realloc(ptr, alloc_size);
+    if (alloc_size && !new_ptr) {
+        fprintf(stderr, "ERROR : Unexpected error at dbg_realloc\n");
+        return NULL;
```
certik, line 135: Exit with an error

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -48,6 +49,144 @@ typedef enum {
...
+    }
+
+    if(!mem_dbg_hashTable.buckets){
+        fprintf(stderr, "ERROR : Unexpected error at dbg_realloc:hashtable not initialized\n");
+        return NULL;
```
certik, line 140: Exit with an error

PR #10574, opened 2026-03-12
Fix array section substring s1(:)(1:4)

certik:
When a character array section was combined with a substring operation like s1(:)(1:4), the substring was silently ignored and the full-length strings were passed instead.

The fix decomposes s1(:)(1:4) into an ArrayConstructor with an ImpliedDoLoop: [(s1(i)(1:4), i=lb,ub)], reusing the existing StringSection(ArrayItem(...)) code path that already handles single-element substrings correctly.


certik:
The failure is unrelated, but we might as well fix it as part of this PR.

PR #10575, opened 2026-03-12
Fix copy-in for assumed-size dummy args

certik:
The array_passed_in_function_call pass only handled DescriptorArray -> PointerArray casts, missing the DescriptorArray -> UnboundedPointerArray case used for assumed-size dummy arguments. This caused non-contiguous arrays (e.g. a(1,:)) passed through assumed-shape to assumed-size dummies to skip the contiguous temporary copy, reading wrong memory locations.


certik:
I think this change might be causing a failure in LAPACK. We'll need to investigate.

certik:
It should work now.

PR #10577, opened 2026-03-12
fix: handle real128 kind correctly in SELECT TYPE

### LINKED ISSUE #10576: Bug: real128 kind not handled correctly in SELECT TYPE
MRE 
```
program expr2
	use, intrinsic :: iso_fortran_env, only: r16 => real128
	implicit none

	allocate(value, source=1.0)

	select type(x => value)
	type is (real(r16))
		print *, x
	class default
		print *, 0
	end select
end program expr2
```
LFortran 
```
semantic error: Kind -1 is not supported for Real
 --> examples/expr2.f90:9:16
  |
9 |  type is (real(r16))
  |                ^^^ 


```
Gfortran 
```
           0
```

Jatinagarwal24:
fix #10576

### DIFF
--- a/integration_tests/select_type_41.f90
+++ b/integration_tests/select_type_41.f90
@@ -1,0 +1,15 @@
+program select_type_41
+    use, intrinsic :: iso_fortran_env, only: real128
+    implicit none
+
+    class(*), allocatable :: value
+    allocate(value, source=1.0)
+
+    select type(x => value)
+    type is (real(real128))
+        ! If real128 is unavailable (=-1), this branch should be ignored.
+        error stop
+    class default
+        if (1 /= 1) error stop
+    end select
+end program select_type_41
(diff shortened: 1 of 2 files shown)


integration_tests/select_type_41.f90
```diff
@@ -0,0 +1,15 @@
+program select_type_41
+    use, intrinsic :: iso_fortran_env, only: real128
+    implicit none
+
+    class(*), allocatable :: value
+    allocate(value, source=1.0)
+
+    select type(x => value)
+    type is (real(real128))
```
certik, line 9: Is `real(-1)` allowed? I thought that's an error. So I don't know if the above is valid Fortran.

certik:
This is not a good fix. The `real(-1)` is not valid Fortran, so the current LFortran behavior is correct.

PR #10579, opened 2026-03-12
Support allocate with mold for poly arrays

certik:
Handle `allocate(class_star_array(N), mold=class_star_var)` where the concrete type is only known at runtime.

Semantics: propagate mold type info to alloc_arg for unlimited polymorphic targets; emit String type for character type-spec allocations.

Codegen: copy vptr from mold, allocate data based on runtime element size, and initialize string descriptors when the mold's dynamic type is character.

New llvm_utils helpers: init_mold_upoly_array_data(), init_string_descriptors(), get_class_type_tag_from_vptr().


certik:
This failure was fixed in the second commit in https://github.com/lfortran/lfortran/pull/10574.

Issue #6665, opened 2025-03-18
bug: throw error when mask shape mismatches the array shape in `minloc`


HarshitaKalani:
```fortran
program main
    print*, minloc([2, 1], 1, [.true.])  
end program
```
```console
$ gfortran b.f90 && ./a.out
b.f90:2:19:

    2 |     print*, minloc([2, 1], 1, [.true.])
      |                   1
Error: Different shape for arguments 'array' and 'mask' for intrinsic minloc at (1) on dimension 1 (2 and 1)
$ lfortran b.f90
2
```
Same is the case with `maxloc`

PR #10593, opened 2026-03-13
Add shape conformance check between array and mask for minloc, maxloc, findloc

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/6665


certik:
Please rebase

Issue #4716, opened 2024-08-31
When `repeatable` in `random_init` is set to false, `random_number` should give different numbers as output


HarshitaKalani:
```fortran
program test_random_seed
  implicit none
  real x(1), y(1)
  call random_init(.false., .true.)
  call random_number(x)
  call random_init(.true., .true.)
  call random_number(y)
  ! x and y are the same sequence
  print* , x
  print* , y
end program test_random_seed

```
```console
$ gfortran b.f90 && ./a.out
  0.740848601    
  0.825262189    
$ lfortran b.f90
8.40187728e-01
8.40187728e-01
```

kargl:
It is better to check the seeds of your prng as that is what `random_init` is actually affecting.
```
program test_random_seed
   implicit none

   integer i, j
   integer, allocatable :: seeds(:)

   call random_seed(size=i)
   allocate(seeds(i))
   j = 2
   if (i < 2) j = 1

   do i = 1, 5
      call random_init(.false., .true.)
      call random_seed(get=seeds)
      write(*,'(2I12,3X)',advance='no') seeds(1:j)

      call random_init(.true., .true.)
      call random_seed(get=seeds)
      write(*,'(2I12,3X)') seeds(1:j)
   end do

end program test_random_seed
```
with gfortran one sees
 
    gfcx -o z a.f90 && ./z
    -1402382843 -1601489764      516270050  1116796470
     1972852821 -1052452947      516270050  1116796470
       446454374     913552963      516270050  1116796470
    -1723403232  1463135888      516270050  1116796470
      -857271102    148852851      516270050  1116796470

PR #10611, opened 2026-03-14
Fix random_init producing identical seeds for non-repeatable mode

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/4716


certik (approved):
I think that's fine. We can further iterate on it as needed if we discover any bugs.

PR #10613, opened 2026-03-14
fix: Reject implied-do loops in assignments without array constructor brackets

### LINKED ISSUE #4520: bug : implied do loops should raise error if not wrapped in array constructor
**MRE**

```fortran
PROGRAM example
    integer :: i
    integer ::arr(4)
    arr =  (  i , i = 1, 4 )
    print *,arr
 END PROGRAM example
```

**GFortran Error**

```shell
tt4.f90:7:17:

    7 |     arr =  (  i , i = 1, 4 )
      |                 1
Error: Expected a right parenthesis in expression at (1)
```
LFortran proceeds with the compilation with no errors as it wraps the `impliedDoLoop` node in `ArrayBroadcast`.
For now, handling the `ArrayBroadcast` of `impliedDoLoop` to fixedarray isn't done right but it could be fixed. Should we stick with GFortran error or make it work and consider it a feature in LFortran :)  ?

Notice : I think we should stick with the error, It should make handling impliedDoLoop with read,write,constructing arrays more convenient and obvious (at least to us while creating ASR).

Jatinagarwal24:
fixes #4520

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -294,7 +294,7 @@ program continue_compilation_1
     type(Derived) :: derived_var
     class(type_t) :: inst_tt
     real(8), parameter :: erfc_param = erfc(40.12_8)
-
+    integer :: arr_idl(4)
 
 
 
@@ -626,7 +626,7 @@ program continue_compilation_1
     type(container(4)) :: obj1
     type(container) :: obj2
 
-
+    arr_idl = (i, i = 1, 4)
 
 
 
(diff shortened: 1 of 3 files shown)


certik:
@HarshilShah1804 please run all tests before merging --- when adding an error message we need to check that all third-party codes work, it might easily happen that we missed some corner case and they now break.

Since it's already merged, please watch the CI in main, if it fails, then revert this PR.

HarshilShah1804:
Sure, I will take care from next time onward. But, I had checked from the fortran standard that `implied-do-loops` are not allowed as the value for assignment variables. So, I thought no third-party codes could be have it. That's why I merged the PR directly to save the CI.

PR #10615, opened 2026-03-14
fix: correct the `loc` for throwing error for `advance` specifier

### LINKED ISSUE #3295: Incorrect data member "Location" for keyword arguments
### Minimal reproducible example:

```fortran
program write
    implicit none
    write (*, "(a)", advance="hello") "Dothraki culture"
end program
```

`--show-asr` (which raises a semantic error, which is the right thing to do):
```
semantic error: ADVANCE= specifier must have value = YES or NO
 --> tests/write4.f90:1:1
  |
1 | program write
  | ^ 
```

### Issue

but the "location" ( at which it points to isn't correct). See: https://github.com/lfortran/lfortran/blob/c4cde93ab20d5a80575be80939f5fe2e0608d8dc/src/lfortran/semantics/ast_body_visitor.cpp#L554 (`kwarg.loc` has an incorrectly assigned value).

It so happens that `kwarg.loc` has both `first` and `last` always zero. Same is the case for all kwargs to all functions (not limited to `read/write`)

HarshilShah1804:
Fixes #3295

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -629,6 +629,8 @@ program continue_compilation_1
     arr_idl = (i, i = 1, 4)
     integer :: minloc_shape_mismatch = minloc([2, 1, 3], 1, [.true., .false.])
     integer :: maxloc_shape_mismatch = maxloc([2, 1, 3], 1, [.true., .false.])
+    write (*, "(a)", advance="hello") "Dothraki culture"
+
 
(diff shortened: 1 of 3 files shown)


HarshilShah1804:
I think we don't require exhaustive checks for this PR since it jsut corrected the `loc` attribute for `kwargs`.

HarshitaKalani:
Please resolve the conflicts.

PR #10618, opened 2026-03-14
Add warning for single/double precision corruption

### LINKED ISSUE #10614: Warn for single/double precision corruption
Add a warning:
```
real(dp) :: x
x = 1.3
```
Something like: "warning: this implies single precision, use kind designator to make it explicit" and underline with "hint: 1.3 is equal to 1.30000002423424234, use 1.3_dp to make it equal to 1.3000000000000000". Lookup the kind parameter, `dp` in this case. I think currently we don't store it in ASR, so we should.

himanshumahajan138:
Fixes #10614

## Summary
- Adds a semantic warning when a single precision real literal (e.g., `1.3`) is implicitly
  converted to double precision, as requested in #10614
- The warning looks up named kind parameters (e.g., `dp`) in scope and suggests them as
  the suffix in the hint message
- Adds a reference test and an integration test

## Example output

```
warning: This implies single precision; use a kind suffix to make precision explicit
--> test.f90:4:5
|
4 | x = 1.3
|     ^^^ hint: this is 1.2999999523162842 in single precision, use _dp suffix for double precision
```

## Test plan
- [x] Reference test `tests/warnings/precision_real_01.f90` verifies warning output
- [x] Integration test `integration_tests/precision_warning_01.f90` compiles and runs with gfortran and llvm
- [x] Warning is suppressed with `--no-warnings`
- [x] All existing warning and cast tests


tests/reference/asr-precision_real_01-fdb950b.stderr
```diff
@@ -0,0 +1,11 @@
+warning: This implies single precision; use a kind suffix to make precision explicit
+ --> tests/warnings/precision_real_01.f90:4:5
+  |
+4 | x = 1.3
+  |     ^^^ hint: this is 1.2999999523162842 in single precision, use _dp suffix for double precision
```
certik, line 5: We'll have to iterate on the exact message. When you say:

> this is 1.2999999523162842 in single precision

The key is that this is actually `1.2999999523162842` in *double* precision, that's why we don't want it.

In single precision it is equal to `1.2999999`, which is fine.

certik:
Update reference tests via `./run_tests.py -u`.

This fix is good at a high level, now we just need to get all the details right and more reference tests to test things that should *not* trigger and things that *should* trigger this warning. All combinations of single to double, double to single, with and without `_dp` as well as with `_4` and `_8` directly, etc.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5560,6 +5560,61 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        ASR::ttype_t *dst_real = ASRUtils::type_get_past_array(
+                            ASRUtils::type_get_past_pointer(target_type));
+                        if (ASR::is_a<ASR::Real_t>(*src_real) &&
+                                ASR::is_a<ASR::Real_t>(*dst_real)) {
+                            int src_kind = ASRUtils::extract_kind_from_ttype_t(src_real);
+                            int dst_kind = ASRUtils::extract_kind_from_ttype_t(dst_real);
+                            if (src_kind == 4 && dst_kind == 8) {
+                                ASR::RealConstant_t *rc =
+                                    ASR::down_cast<ASR::RealConstant_t>(value);
+                                double single_val = rc->m_r;
```
certik, line 5591: If it is a single precision, that I would do:

```suggestion
                                float single_val = rc->m_r;
```

I think internally we store it as double, but without losing any precision to convert to single.

tests/reference/asr-precision_real_01-fdb950b.stderr
certik: I still like what I wrote in the issue better:
```
hint: 1.3 is equal to 1.2999999523162842, use 1.3_dp to make it equal to 1.3000000000000000
```

So extract the `1.3` in the error message printing. And use the actual double precision number, which will not be exactly `1.3000`, there will be some "change" at the end. Also to print it, I would print it in scientific notation with 17 significant digits, that way ALL the digits are printed, so when the decimal is converted back to binary, it will be bit-to-bit equal.

PR #10619, opened 2026-03-14
Fix segfault in allocate for class arrays

### LINKED ISSUE #10617: Sourced allocation segfaults for nested derived types
`$ cat mymod.f90 `
```
module mymod
   
   type :: MyType
   end type MyType
   
   type :: WrapperType
   end type WrapperType
   
   type :: ClientType
      class(WrapperType), allocatable :: arr(:)
   end type ClientType
   
   interface ClientType
      procedure :: constructor
   end interface ClientType

contains
   
      class(MyType), intent(in) :: obj
      type(ClientType)          :: self
      allocate( self%arr(1) )
      allocate( self%arr(1)%obj, source = obj )
   end function constructor

end module mymod
```
`$ cat main.f90 `
```
program main

   use mymod

   type(MyType) :: obj
   class(ClientType), allocatable :: client
   
   allocate( client, source = ClientType(obj) )
      
end program main
```
`$ lfortran -c --separate-compilation mymod.f90`
`$ lfortran -c --separate-compilation main.f90`
`$ lfortran *.o -o Main; ./Main`
```
```

certik:
The deep copy of polymorphic (class) allocatable array components used the wrong memory layout model. Class arrays use a ONE-wrapper layout: a single {vptr, data_ptr} wrapper where data_ptr points to a contiguous buffer of N elements. The deep copy code incorrectly treated them as N separate wrappers, causing out-of-bounds access for arrays with 2+ elements.

Fix: add a dedicated class array deep copy path (similar to the existing unlimited polymorphic path) that correctly handles the ONE-wrapper layout with compile-time element size. Skip the N-wrapper realloc for class arrays since they only have one wrapper.

An integration test (allocate_56) is added covering allocate with source= from both function results and variables, with multi-element arrays and deep copy independence verification.

Fixes #10617.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2193,6 +2193,7 @@ RUN(NAME allocate_53 LABELS gfortran llvm)
 RUN(NAME allocate_54 LABELS gfortran llvm)
 RUN(NAME allocate_55 LABELS gfortran llvm)
 RUN(NAME allocate_56 LABELS gfortran llvm)
+RUN(NAME allocate_57 LABELS gfortran llvm)
 
 RUN(NAME realloc_lhs_01 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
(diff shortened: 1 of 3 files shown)


certik:
The CI passes, I just rebased it and fixed test filename conflict, so I merged it.

PR #10623, opened 2026-03-14
Fix list-directed internal read with multiple values

certik:
The list-directed internal `read (read(string, *) x, y, ...)` only read the first value and returned, ignoring subsequent values. This was caused by a premature 'return' in the is_string codegen path in visit_FileRead.

Additionally, each runtime `_lfortran_string_read_*` function always parsed from position 0 of the source string, so even without the early return, the second value would re-read from the start instead of continuing.

Fix:
- Add `int64_t *offset` parameter to all scalar `_lfortran_string_read_*` runtime functions to track the current read position across calls
- Change 'return' to 'continue' in the codegen loop so all values are read
- Allocate a shared offset variable in the codegen before the loop
- Skip `_lfortran_empty_read` for internal (string) reads since there is no file position to advance

Integration test read_56.f90 covers: int+char, int+int, int+real, comma-separated, three mixed values, and single-value regression.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3015,13 +3015,11 @@ RUN(NAME read_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_50 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_51 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_52 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-
-
-
 RUN(NAME read_53 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_54 LABELS gfortran llvm)
 RUN(NAME read_55 LABELS gfortran llvm)
 RUN(NAME read_56 LABELS gfortran llvm)
+RUN(NAME read_57 LABELS gfortran llvm)
 
 RUN(NAME write_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME write_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
--- a/integration_tests/read_57.f90
+++ b/integration_tests/read_57.f90
@@ -1,0 +1,47 @@
+program read_57
+  ! Test list-directed internal read with multiple values of mixed types
+  implicit none
+  character(len=20) :: line
+  integer :: x, y
+  character(1) :: c
+  real :: r
+  character(len=5) :: s
+
+  ! Integer then character
+  line = '1 A'
+  read(line, *) x, c
+  if (x /= 1) error stop
+  if (c /= 'A') error stop
+
+  ! Two integers
+  line = '10 20'
+  read(line, *) x, y
+  if (x /= 10) error stop
+  if (y /= 20) error stop
+
+  ! Integer then real
+  line = '3 2.5'
+  read(line, *) x, r
+  if (x /= 3) error stop
+  if (abs(r - 2.5) > 1.0e-6) error stop
+
+  ! Comma-separated values
+  line = '7,8'
+  read(line, *) x, y
+  if (x /= 7) error stop
+  if (y /= 8) error stop
+
+  ! Three values: integer, real, character
+  line = '42 3.14 Z'
+  read(line, *) x, r, c
+  if (x /= 42) error stop
+  if (abs(r - 3.14) > 0.01) error stop
+  if (c /= 'Z') error stop
+
+  ! Single value (regression: should still work)
+  line = '99'
+  read(line, *) x
+  if (x /= 99) error stop
+
+  print *, 'PASS'
+end program
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14844,6 +14844,18 @@ public:
             emit_formatted_read(x, unit_val, iostat, read_size, advance, advance_length, is_string);
         } else {
             llvm::Value* var_to_read_into = nullptr; // Var expression that we'll read into.
+            // For multi-value list-directed internal reads, track position
+            llvm::Value *str_offset = nullptr;
+            llvm::Value *str_src_data = nullptr, *str_src_len = nullptr;
+            if (is_string) {
+                str_offset = llvm_utils->CreateAlloca(*builder,
+                    llvm::Type::getInt64Ty(context), nullptr, "str_read_offset");
+                builder->CreateStore(
+                    llvm::ConstantInt::get(llvm::Type::getInt64Ty(context), 0),
+                    str_offset);
+                std::tie(str_src_data, str_src_len) = llvm_utils->get_string_length_data(
+                    ASRUtils::get_string_type(x.m_unit), unit_val);
+            }
             for (size_t i=0; i<x.n_values; i++) {
                 // Handle ImpliedDoLoop: read(10,*) (vals(j), j=1,n)
                 // Transform to reading n elements into vals starting at start index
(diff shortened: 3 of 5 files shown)


YashNagda17:
I'm sorry, I didn't look into this PR earlier

Issue #10635 is related, I was working on it, but this PR seems to already solve it

So, I'm dropping the fix I wrote, and will register the tests once this PR is merged

I have also worked upon the rest 2 fixes related to read, #10637 & #10638, will send a PR once this PR is merged and rebasing on top of it, to avoid potential conflicts.

certik:
@YashNagda17 perfect! The CI is almost done here.

YashNagda17:
Here's an alternate design:
Instead of using llvm variables for buffer, trying to use c++ variables to keep track of buffer in same read:
It works for reads of same variable type, but fails for reads into different types of variables (eg, string and int)

```diff

commit c1263c63d8096129c186b38405084cb27fa69d45
Author: Yash Nagda <yashnagdasub17@gmail.com>
Date:   Sun Mar 15 13:40:55 2026 +0530

    fix: add support for multiple variables inside single read() statement string.
    
    Issue: When performing an internal read from a string (e.g., read(string, *) a, b), only the first
    variable was being read correctly. The remaining variables were ignored because the backend triggered
    an early return after the first read operation.
    
    Fix:
    - Inside the visit_FileRead() node, in asr to llvm generation, add support for multiple variables
    - Here, instead of early return, adding support to let the loop run for multiple reads when we read from same string (is_string) branch.
    
    - Also, for multi variable reads from same string, adding a mechanism to store memory between 2 reads, to avoid each read always starting from beginining of the string. _str_read_offset variable inside lfortran_intrinsics is used to keep track of offset in current string.
    
    - At the start of each read() statement, we reset the offset to zero, using the _lfortran_string_read_offset_reset() api introduced.

diff --git a/src/libasr/codegen/asr_to_llvm.cpp b/src/libasr/codegen/asr_to_llvm.cpp
index 74202aff7..ddaee6c56 100644
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -14843,6 +14843,19 @@ public:
         if (x.m_fmt) {
             emit_formatted_read(x, unit_val, iostat, read_size, advance, advance_length, is_string);
         } else {
+            if (is_string) {
+                // Reset the string-read cursor so each read statement
+                // starts scanning from position 0 of the source string.
+                std::string reset_fn_name = "_lfortran_string_read_offset_reset";
+                llvm::Function *reset_fn = module->getFunction(reset_fn_name);
+                if (!reset_fn) {
+                    llvm::FunctionType *ft = llvm::FunctionType::get(
+                        llvm::Type::getVoidTy(context), {}, false);
+                    reset_fn = llvm::Function::Create(ft,
+                        llvm::Function::ExternalLinkage, reset_fn_name, module.get());
+                }
+                builder->CreateCall(reset_fn, {});
+            }
             llvm::Value* var_to_read_into = nullptr; // Var expression that we'll read into.
             for (size_t i=0; i<x.n_values; i++) {
                 // Handle ImpliedDoLoop: read(10,*) (vals(j), j=1,n)
@@ -15140,7 +15153,7 @@ public:
                         builder->CreateStore(extended, iostat_user);
                     }
                     emit_set_read_iomsg();
-                    return;
+                    continue;
                 } else {
                     fn = get_read_function(type);
                 }
@@ -15233,6 +15246,10 @@ public:
             // read, and anything after that will be skipped.
             // Here, we can use `_lfortran_empty_read` function to move to the
             // pointer to the next line.
+            // For internal string reads (is_string=true), unit_val is a
+            // string descriptor, not an integer unit number, so empty_read
+            // does not apply and must be skipped.
+            if (is_string) return;
             std::string runtime_func_name = "_lfortran_empty_read";
             llvm::Function *fn = module->getFunction(runtime_func_name);
             if (!fn) {
diff --git a/src/libasr/runtime/lfortran_intrinsics.c b/src/libasr/runtime/lfortran_intrinsics.c
index 0e9630297..4c6f05316 100644
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -8257,9 +8257,20 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
     if(iostat != NULL) *iostat = 0;
 }
 
+// Cursor for list-directed reads from a string (read(str, *) v1, v2, ...).
+// Reset by _lfortran_string_read_reset(), called once per read statement from codegen.
+static int _str_read_offset = 0;
+
+LFORTRAN_API void _lfortran_string_read_offset_reset() { _str_read_offset = 0; }
+
 LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format, int32_t *i, int32_t *iostat) {
     char *buf = to_c_string((const fchar*)str, len);
-    int rc = sscanf(buf, format, i);
+    int rc = sscanf(buf + _str_read_offset, format, i);
+    if (rc == 1) {
+        int dummy, advance = 0;
+        sscanf(buf + _str_read_offset, " %d%n", &dummy, &advance);
+        _str_read_offset += advance;
+    }
     free(buf);
     if (rc != 1) {
         if (iostat) { *iostat = 5010; return; }
@@ -8272,7 +8283,12 @@ LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format
 
 LFORTRAN_API void _lfortran_string_read_i64(char *str, int64_t len, char *format, int64_t *i, int32_t *iostat) {
     char *buf = to_c_string((const fchar*)str, len);
-    int rc = sscanf(buf, format, i);
+    int rc = sscanf(buf + _str_read_offset, format, i);
+    if (rc == 1) {
+        long long dummy; int advance = 0;
+        sscanf(buf + _str_read_offset, " %lld%n", &dummy, &advance);
+        _str_read_offset += advance;
+    }
     free(buf);
     if (rc != 1) {
         if (iostat) { *iostat = 5010; return; }
@@ -8285,7 +8301,12 @@ LFORTRAN_API void _lfortran_string_read_i64(char *str, int64_t len, char *format
 LFORTRAN_API void _lfortran_string_read_f32(char *str, int64_t len, char *format, float *f, int32_t *iostat) {
     char *buf = to_c_string((const fchar*)str, len);
     convert_fortran_d_exponent(buf);
-    int rc = sscanf(buf, format, f);
+    int rc = sscanf(buf + _str_read_offset, format, f);
+    if (rc == 1) {
+        float dummy; int advance = 0;
+        sscanf(buf + _str_read_offset, " %f%n", &dummy, &advance);
+        _str_read_offset += advance;
+    }
     free(buf);
     if (rc != 1) {
         if (iostat) { *iostat = 5010; return; }
@@ -8298,7 +8319,12 @@ LFORTRAN_API void _lfortran_string_read_f32(char *str, int64_t len, char *format
 LFORTRAN_API void _lfortran_string_read_f64(char *str, int64_t len, char *format, double *f, int32_t *iostat) {
     char *buf = to_c_string((const fchar*)str, len);
     convert_fortran_d_exponent(buf);
-    int rc = sscanf(buf, format, f);
+    int rc = sscanf(buf + _str_read_offset, format, f);
+    if (rc == 1) {
+        double dummy; int advance = 0;
+        sscanf(buf + _str_read_offset, " %lf%n", &dummy, &advance);
+        _str_read_offset += advance;
+    }
     free(buf);
     if (rc != 1) {
         if (iostat) { *iostat = 5010; return; }
diff --git a/src/libasr/runtime/lfortran_intrinsics.h b/src/libasr/runtime/lfortran_intrinsics.h
index 9ecdeb768..fb713a81c 100644
--- a/src/libasr/runtime/lfortran_intrinsics.h
+++ b/src/libasr/runtime/lfortran_intrinsics.h
@@ -294,6 +294,7 @@ LFORTRAN_API void _lfortran_string_write(char **str_holder, bool is_allocatable,
         int64_t* len, int32_t* iostat, const char* format,
         int64_t format_len, ...);
 LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const char* format_data, int64_t format_len, ...);
+LFORTRAN_API void _lfortran_string_read_offset_reset();
 LFORTRAN_API void _lfortran_string_read_i32(char *str, int64_t len, char *format, int32_t *i, int32_t *iostat);
 LFORTRAN_API void _lfortran_string_read_i32_array(char *str, int64_t len, char *format, int32_t *arr);
 LFORTRAN_API void _lfortran_string_read_i64(char *str, int64_t len, char *format, int64_t *i, int32_t *iostat);


```

PR #10631, opened 2026-03-14
Codegen: handle AssumedLength strings in associate variables

### LINKED ISSUE #10629: ICE: setup_string_length asserts on assumed-length string from modfile
## Summary

When compiling a program that uses modules with assumed-length string arguments (`character(len=*)`), lfortran crashes with an internal assertion in codegen:

```
LCOMPILERS_ASSERT failed: src/libasr/codegen/asr_to_llvm.cpp
function setup_string_length(), line number 951 at
false
ERROR MESSAGE:
Shouldn't define assumed length string variable (They're only arguments)
```

## Reproduction

Triggered when compiling fortplot (via `fpm build --compiler=lfortran`). The crash occurs on any compilation unit that instantiates modules containing type-bound procedures with `character(len=*)` arguments via modfiles.

Minimal pattern that triggers it (separate compilation required):

```fortran
! mod.f90
module mymod
  implicit none
  type :: fig_t
    integer :: w = 0
  contains
    procedure :: save => fig_save
  end type
contains
  subroutine fig_save(self, filename)
    class(fig_t), intent(inout) :: self
    character(len=*), intent(in) :: filename
    print *, filename, self%w
  end subroutine
end module
```

```fortran
! main.f90
program main
  use mymod, only: fig_t
  implicit none
  type(fig_t) :: fig
  call fig%save("output.png")
end program
```

The crash occurs when the modfile-instantiated function's assumed-length arguments go through `setup_string_length()`, which asserts that `AssumedLength` should never appear for variable declarations. For arguments, the length is set by the caller, so `setup_string_length` should simply skip them.

## Location

`src/libasr/codegen/asr_to_llvm.cpp`, `setup_string_length()`, line ~951

## Context

Fortplot compatibility blocker. See meta-issue #9895.

krystophny:
## Summary

When `associate` aliases a `character(len=*)` argument, the associate variable
is a local (`intent Local`) that inherits the `AssumedLength` kind from the
argument. The variable declaration path in codegen calls `setup_string()` which
calls `setup_string_length()`, and that function had an assert rejecting
`AssumedLength` on the assumption it could only appear on dummy arguments.
The `associate` construct is a legitimate counter-example.

Fixes #10629

## What changed

**`setup_string_length()`**: Replace the assert with a no-op `break`, matching
the existing `DeferredLength` case directly below. The associate variable's
length will be filled in later by `visit_Associate`, which copies the source
descriptor. Semantic validation of `character(len=*)` locals belongs in
semantics (which already rejects them), not in codegen.

**`setup_string()`**: Skip `set_string_memory_on_heap()` for `AssumedLength`
strings. The associate variable is an alias — its data pointer will be set by
`visit_Associate` to point at the source's data. Allocating heap memory here
would leak because the pointer gets overwritten immediately.

## Changes

- [`asr_to_llvm.cpp#L949-L951`](https://github.com/lfortran/lfortran/blob/41037a64b/src/libasr/codegen/asr_to_llvm.cpp#L949-L951): no-op break for AssumedLength in setup_string_length
- [`asr_to_llvm.cpp#L974-L975`](https://github.com/lfortran/lfortran/blob/41037a64b/src/libasr/codegen/asr_to_llvm.cpp#L974-L975): skip heap allocation for AssumedLength in setup_string
- [`separate_compilation_45a.f90`](https://github.com/lfortran/lfortran/blob/41037a64b/integration_tests/separate_compilation_45a.f90): module with `associate` on a `character(len=*)` argument
- [`separate_compilation_45.f90`](https://github.com/lfortran/lfortran/blob/41037a64b/integration_tests/separate_compilation_45.f90): program using the module via separate compilation

### Fails on main (abort in codegen assert)
```
$ git checkout upstream/main
$ lfortran -c separate_compilation_45a.f90 -o sc45a.o  # OK
$ lfortran -c separate_compilation_45.f90 -o sc45.o
LCOMPILERS_ASSERT failed: src/libasr/codegen/asr_to_llvm.cpp
function setup_string_length(), line number 951 at
false
ERROR MESSAGE:
Shouldn't define assumed length string variable (They're only arguments)
```

### Passes after fix
```
$ git checkout fix/10629-assumed-length-ice
$ lfortran -c separate_compilation_45a.f90 -o sc45a.o  # OK
$ lfortran -c separate_compilation_45.f90 -o sc45.o    # OK
$ scripts/lf.sh itest -b llvm -t separate_compilation_45
$ scripts/lf.sh itest -b gfortran -t separate_compilation_45
```

### Broader suites
```
$ scripts/lf.sh test
TESTS PASSED
$ scripts/lf.sh itest -b llvm -t separate_compilation
100% tests passed, 0 tests failed out of 35
$ scripts/lf.sh itest -b llvm -t string_
100% tests passed, 0 tests failed out of 113
```

### Fortplot (original trigger)
```
# On main - aborts at 67%:
[ 67%]  test_pdf_scale_regression.f90
LCOMPILERS_ASSERT failed: asr_to_llvm.cpp function setup_string_length()

# On fix branch - gets past this file:
[ 67%]  test_pdf_scale_regression.f90  done.
```

### DIFF
--- a/integration_tests/separate_compilation_45.f90
+++ b/integration_tests/separate_compilation_45.f90
@@ -3,6 +3,7 @@ program separate_compilation_45
     implicit none
     type(greeter_t) :: g
     g%id = 42
+    if (g%id /= 42) error stop "expected greeter id"
     call g%greet("hello")
-    print *, "ok"
+    if (g%id /= 42) error stop "unexpected greeter id after greet"
 end program separate_compilation_45
--- a/integration_tests/separate_compilation_45a.f90
+++ b/integration_tests/separate_compilation_45a.f90
@@ -14,8 +14,9 @@ contains
         character(len=*), intent(in) :: msg
         ! associate creates a local variable that inherits AssumedLength
         associate(unused => msg)
+            if (trim(unused) /= "hello") error stop "expected greeting"
+            if (self%id /= 42) error stop "expected greeter id"
         end associate
-        print *, trim(msg), self%id
     end subroutine greet
 
 end module separate_compilation_45a_mod


HarshitaKalani:
This is due to a network issue, Unrelated to this PR.
```console
Cloning into 'fpm'...
fatal: unable to access 'https://github.com/fortran-lang/fpm.git/': Failed to connect to github.com port 443 after 135590 ms: Couldn't connect to server
```

integration_tests/separate_compilation_45.f90
```diff
@@ -0,0 +1,8 @@
+program separate_compilation_45
+    use separate_compilation_45a_mod, only: greeter_t
+    implicit none
+    type(greeter_t) :: g
+    g%id = 42
+    call g%greet("hello")
+    print *, "ok"
```
HarshitaKalani: Can we have error stop conditions here to make this more robust?

integration_tests/separate_compilation_45a.f90
```diff
@@ -0,0 +1,21 @@
...
+
+contains
+
+    subroutine greet(self, msg)
+        class(greeter_t), intent(in) :: self
+        character(len=*), intent(in) :: msg
+        ! associate creates a local variable that inherits AssumedLength
+        associate(unused => msg)
+        end associate
+        print *, trim(msg), self%id
```
HarshitaKalani: here as well?

krystophny:
done

Issue #10640, opened 2026-03-15
ASR verify error: "ArrayItem must be an array, not a scalar" when assigning to derived-type component array


Jatinagarwal24:
MRE 
```
module expr2_mre
	implicit none

	type :: TokenStack
		integer :: dummy = 0
	end type TokenStack

	type, public :: EquationParser
		character(:), allocatable :: equation
		character(:), allocatable :: variableName
		character(:), allocatable :: inFixFormula
		integer :: nIndepVars
		type(IndepVar), dimension(:), allocatable :: indepVars
		type(TokenStack) :: inFix
		type(TokenStack) :: postFix
	contains
		procedure :: Touch
	end type EquationParser

	type :: IndepVar
		character(:), allocatable :: value
	end type IndepVar

	interface EquationParser
		procedure Construct_EquationParser
	end interface EquationParser

contains

	function Construct_EquationParser(indepVars) result(parser)
		type(EquationParser) :: parser
		character(*)         :: indepVars(:)
		integer :: i
		integer :: nIndepVars

		nIndepVars = size(indepVars)
		allocate(parser%indepVars(1:nIndepVars))
		parser%nIndepVars = nIndepVars

		do i = 1, nIndepVars
			parser%indepVars(i)%value = trim(indepVars(i))
		end do
	end function Construct_EquationParser

	subroutine Touch(parser)
		class(EquationParser), intent(inout) :: parser
		if (parser%nIndepVars > 0) then
			parser%nIndepVars = parser%nIndepVars
		end if
	end subroutine Touch

end module expr2_mre

program main
	use expr2_mre
	implicit none

	type(EquationParser) :: parser
	character(10) :: names(2)

	names = [character(10) :: 'x', 'y']
	parser = EquationParser(names)
	call parser%Touch()

	print *, parser%indepVars(1)%value
end program main
```
Lfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran examples/expr2.f90 
ASR verify pass error: The variable in ArrayItem must be an array, not a scalar
  --> examples/expr2.f90:41:4
   |
41 |    parser%indepVars(i)%value = trim(indepVars(i))
   |    ^^^^^^^^^^^^^^^^^^^^^^^^^ failed here


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
(lf) root@LAPTOP-8SLTQKJO:~/lfortran#
```
Gfortran 
```
[...]

PR #10641, opened 2026-03-15
fix: Resolve ASR verification error for derived type array element component access

Jatinagarwal24:
fix: Recursively handle wrapped types in deferred struct type resolution

Previously, the code only handled Allocatable_t and Pointer_t wrappers at the
top level, missing nested cases where StructType_t itself needs updating.

The fix introduces replace_deferred_struct_type() helper that recursively
handles all type wrappers:
- Direct StructType_t (returns updated StructType)
- Array_t (updates element type recursively)
- Pointer_t (updates pointee type recursively)
- Allocatable_t (updates base type recursively)

This ensures type signatures stay in sync when accessing components of
allocatable/pointer derived type arrays (e.g., parser%indepVars(i)%value).

Resolves: #10640

### DIFF
--- a/integration_tests/allocate_62.f90
+++ b/integration_tests/allocate_62.f90
@@ -1,4 +1,4 @@
-module allocate61_mod
+module allocate62_mod
     implicit none
 
     type :: token_stack
@@ -49,10 +49,10 @@ contains
         end if
     end subroutine touch
 
-end module allocate61_mod
+end module allocate62_mod
 
-program allocate_61
-    use allocate61_mod
+program allocate_62
+    use allocate62_mod
     implicit none
 
     type(equation_parser) :: parser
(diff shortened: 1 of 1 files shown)


Jatinagarwal24:
Hi,@certik pls review this.

integration_tests/allocate_58.f90
HarshitaKalani, line 1: This test works for me in main. Can you please add a test which fails in main and works with this PR?

Jatinagarwal24:
Hi,@certik pls review this and pls suggest if any changes are required.

Jatinagarwal24:
Hi,@HarshitaKalani can you review this once.

Jatinagarwal24:
> Looks good to me. Thanks @Jatinagarwal24 .

Hi,@HarshitaKalani I think this is good to merge now.

PR #10648, opened 2026-03-15
fix: handle in allocate for string variables

### LINKED ISSUE #10647: `mold` argument in allocate not supported to string variables
MRE
```fortran
program allocate_58
    character(len=5) :: a
    character(len=:), allocatable :: b

    allocate(b, mold=a)

    if (len(b) /= 5) error stop
end program
```

LFortran
```console
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/lfortran$ lfortran a.f90 
semantic error: The type of the argument is not supported yet for mold.
 --> a.f90:5:22
  |
5 |     allocate(b, mold=a)
  |                      ^ 


```

HarshilShah1804:
Fixes #10647


HarshilShah1804:
I believe we don't require exhaustive checks on this PR. Since, it adds a new feature, so it wouldn't have been used in any third party codes.

PR #10654, opened 2026-03-16
Fix : Make `array_struct_temporary` pass create a temporary for function with class return

assem2002:
fixes #10644

PR #10656, opened 2026-03-16
Fix polymorphic allocatable function in overloaded assignment

### LINKED ISSUE #10644: Compilation segfaults on type-bound assignment of a function result
This one still fails with respect to type-bound assignments.

`$ cat mymod.f90 `
```
module mymod

   type, abstract :: MyType
   contains
      procedure(op), deferred :: assign
      generic :: assignment(=) => assign
   end type MyType
   
   abstract interface
      subroutine op(self,b)
         import
         class(MyType), intent(out) :: self
         class(MyType), intent(in) :: b
      end subroutine op
   end interface
   
contains

   subroutine test()
      class(MyType), allocatable :: a
      a = client()
   end subroutine test
   
   function client() result(b)
      class(MyType), allocatable :: b
   end function client

end module mymod
```
`$ lfortran -c --separate-compilation  mymod.f90 `
```
```

certik:
The subroutine_from_function pass was emitting the m_overloaded SubroutineCall without first visiting its argument expressions. When the RHS of an assignment with user-defined assignment(=) was a FunctionCall returning a polymorphic allocatable (class) type, Phase 1 converted the function to a subroutine (void return), but the FunctionCall inside the overloaded call's arguments was never replaced, leaving an invalid FunctionCall to a void-returning function.

Fix: visit the overloaded SubroutineCall before emitting it so the expression replacer converts any nested FunctionCalls to SubroutineCall + temp variable.

Integration test added.

Fixes #10644.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3553,6 +3553,7 @@ RUN(NAME module_function_without_nopass LABELS gfortran llvm llvm_wasm llvm_wasm
 RUN(NAME polymorphic_arguments_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME polymorphic_arguments_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocatable_polymorphic_mold_01 LABELS gfortran llvm)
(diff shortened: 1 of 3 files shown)


assem2002:
Looks good to me. both this PR and https://github.com/lfortran/lfortran/pull/10654 have same effect, this is PR is minimal so it's better.

PR #10663, opened 2026-03-16
Reject legacy arrays by default

certik:


certik:
There is one issue:

**Title:** `--legacy-array-sections` generates heavier LLVM IR than direct sequence association codegen

**Description:**

When passing a scalar array element to an array dummy argument via sequence association (F2018 15.5.2.11), the codegen previously handled this efficiently by passing a direct pointer to the element. For example:

```fortran
subroutine a(w)
  double precision :: w(5)
  integer icon
  icon = 2
  call b(w(icon))
end
```

The old LLVM IR for the call was a simple GEP + pointer pass.

After adding a semantic error that requires `--legacy-array-sections` for this pattern, the semantic pass transforms `w(icon)` into an `ArraySection` (`w(icon:5)`), which generates significantly more LLVM IR: array descriptor allocation, dimension descriptor setup, bounds metadata, and `ArrayPhysicalCast`. This is functionally correct but adds unnecessary runtime overhead for a case that could be a simple pointer.

The `legacy_array_sections_helper` already has an efficient path for external symbols that emits just a `GetPointer`:

```cpp
if (callee_is_external_symbol && expected_phys_type == PointerArray && !is_character) {
    arg.m_value = GetPointer(arg_expr, cast_target_type);
    continue;
}
```

**Possible solutions:**

1. Extend the `GetPointer` path in `legacy_array_sections_helper` to also apply to internal/known calls when a simple pointer suffices.
2. Add an optimization pass that simplifies `ArraySection(a, i:ubound)` back to a pointer when the section is contiguous and only used as a call argument.

**Visible in:** `tests/reference/llvm-legacy_array_sections_01-2515c0f.stdout` (went from ~89 lines to ~723 lines of LLVM IR).

PR #10667, opened 2026-03-16
Fix incorrect dependency detection for character(a%len()) in ASR verifier

### LINKED ISSUE #10660: ASR verify error: Incorrect dependency detection for character(a%len()) with polymorphic component
MRE 
```
module expr2_string
	implicit none

	type :: string
		character(:), allocatable :: chars
	contains
		procedure :: len => string_len
	end type string

contains

	pure integer function string_len(self)
		class(string), intent(in) :: self
		string_len = len(self%chars)
	end function string_len

end module expr2_string


module expr2_mre
	use expr2_string, only: string
	implicit none

	abstract interface
		pure subroutine upper_x(str, res)
			character(*), intent(in)         :: str
			character(len(str)), intent(out) :: res
		end subroutine upper_x
	end interface

contains

	subroutine upper_caller(f, a)
		procedure(upper_x)   :: f
		class(*), intent(in) :: a

		select type (a)
		type is (string)
			block
				! MRE trigger: local char length depends on polymorphic component.
				character(a%len()) :: res
				call f(a%chars, res)
			end block
		end select
	end subroutine upper_caller

end module expr2_mre

program expr2_main
	use expr2_mre, only: upper_caller
	implicit none
end program expr2_main
```
Lfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/benchmark.f# lfortran expr2.f90
ASR verify pass error: Variable res doesn't depend on a but is found in its dependency list.
  --> expr2.f90:41:27
   |
41 |     character(a%len()) :: res
   |                           ^^^ failed here


```
Gfortran
```
(lf) root@LAPTOP-8SLTQKJO:~/benchmark.f# gfortran expr2.f90
(lf) root@LAPTOP-8SLTQKJO:~/benchmark.f# ./a.out
```

Jatinagarwal24:
Fix verifier false positive for variable dependency in select type with type-bound function call

The ASR verifier only visited FunctionCall::m_dt when it was a
StructInstanceMember, but the dependency collector in semantics
walks m_dt unconditionally. This caused a mismatch: for
character(a%len()) inside select type, 'a' was recorded as a
dependency of 'res' but the verifier never found it when
recomputing dependencies.

Fix: visit m_dt unconditionally in the verifier's visit_FunctionCall.

Fixes #10660.

### DIFF
--- a/integration_tests/allocate_59.f90
+++ b/integration_tests/allocate_59.f90
@@ -1,4 +1,4 @@
-module expr2_string
+module allocate59_string
 	implicit none
 
 	type :: string
@@ -14,11 +14,11 @@ contains
 		string_len = len(self%chars)
 	end function string_len
 
-end module expr2_string
+end module allocate59_string
 
 
-module expr2_mre
-	use expr2_string, only: string
+module allocate59_mre
+	use allocate59_string, only: string
 	implicit none
 
 	abstract interface
@@ -52,11 +52,11 @@ contains
 		end select
 	end subroutine upper_caller
 
-end module expr2_mre
+end module allocate59_mre
 
 program allocate_59
-	use expr2_mre, only: upper_caller, to_upper_identity
-	use expr2_string, only: string
+	use allocate59_mre, only: upper_caller, to_upper_identity
+	use allocate59_string, only: string
 	implicit none
 
 	type(string) :: s


integration_tests/allocate_59.f90
```diff
@@ -0,0 +1,68 @@
...
+
+	pure integer function string_len(self)
+		class(string), intent(in) :: self
+		string_len = len(self%chars)
+	end function string_len
+
+end module expr2_string
+
+
+module expr2_mre
```
jinangshah21: Please rename module name to be unique name w.r.t file name.

Jatinagarwal24: Hi,@jinangshah21 I have renamed this.Please check once and also run exhaustive tests.Thanks!

certik:
Why was the ASR check incorrect?

certik:
@Jatinagarwal24 when you submit a PR, please create a good git commit message what the bug was and why the old code was incorrect and what the fix is doing.

certik:
Also the bug and MRE has nothing to do with allocate: https://github.com/lfortran/lfortran/issues/10660#issuecomment-4076506627.

certik:
I think it's fine to merge since CI passes. I wrote a better commit message.

Jatinagarwal24:
> @Jatinagarwal24 when you submit a PR, please create a good git commit message what the bug was and why the old code was incorrect and what the fix is doing.

Ok Sir ,I will remember from next time.

PR #10670, opened 2026-03-16
Speed up third-party CI tests (~42 min → ~19 min)

certik:
Remove SciPy tests (no longer uses Fortran), remove stdlib Less Workarounds (too slow), simplify PRIMA and POT3D to compile-only, and drop --fast/--separate-compilation/--std=f23 variant builds for FPM, fortran_mpi, dftatom, fastGPT, SNAP, Numerical Methods Fortran, Fortran-Primes, and Modern Minpack. Keep all quick fpm-based codes, stdlib, and LAPACK (both sections) unchanged.


certik:
The speedup seems to be huge. For example LLVM 22 went from 39:48 to 16:20. More should be possible, but this is a great start.

Yes, I thought about the `--fast` tests, and we do test all our integration tests with `--fast`, so I think that's enough. Once we start working on performance, we can re-enable it, or just manually extract MREs. For example we can have some optional CI run that runs with `--fast`, and if it fails we extract MREs manually.

assem2002:
Yeah I had the same label idea.

PR #10671, opened 2026-03-16
Fix 2 more stdlib runtime bugs

jinangshah21:


certik:
I'll risk it without exhaustive tests, hopefully we don't need to revert it.

PR #10675, opened 2026-03-16
fix: handle multiple namelists declared on same line

### LINKED ISSUE #10674: Parser Error on declaring multiple namelists on same line
MRE
```fortran
program namelist_29
implicit none
character(len=:), allocatable :: type
character(len=100) :: buffer1, buffer2

namelist /args/ type, /long/ type
end program namelist_29
```

LFortran
```console
syntax error: Token '/' is unexpected here
 app/bookkeeper.f90:10:68
 character(len=:),allocatable :: type    ; namelist /args/ type,    /long/ type

```

HarshilShah1804:
Fixes #10674

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -992,6 +992,7 @@ RUN(NAME namelist_26 LABELS gfortran llvm)
 RUN(NAME namelist_27 LABELS gfortran llvm)
 RUN(NAME namelist_28 LABELS gfortran llvm)
 RUN(NAME namelist_29 LABELS gfortran llvm)
+RUN(NAME namelist_30 LABELS gfortran llvm)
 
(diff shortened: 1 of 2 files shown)


src/lfortran/parser/parser.yy
```diff
@@ -1711,7 +1723,7 @@ sep_one
     ;
 
 decl_statements
-    : decl_statements decl_statement { $$ = $1; LIST_ADD($$, $2); }
+    : decl_statements decl_statement { $$ = $1; LIST_ADD($$, $2); DRAIN_EXTRA($$); }
```
certik, line 1726: I don't like this approach.

Is there a way to parse it using Bison? Usually there is a way to rewrite the grammar, perhaps to simpler cases and then reconstruct the proper AST out of it. See how we parse the common block, I think it is a similar problem.

PR #10679, opened 2026-03-17
Speed up macOS CI: increase and enforce NPROC, skip redundant tests

certik:
- Increase NPROC from 2 to 3 to match macOS ARM64 runner core count
- Pass -j${NPROC} to all run_tests.py calls to prevent the default of 8 threads from overcommitting the 3-core runner
- Skip -sc (separate compilation) and -f (fast) test passes on macOS since they are already covered by Ubuntu CI
- Apply same -j3 limit and -f skip to the f23 integration test step

- Exhaustive tests: Only test llvm_wasm with LLVM 19


certik:
I think we should implement `-j` for the reference tests also. Too many bugs due to that. And to control the parallelism.

PR #10680, opened 2026-03-17
Refactor the runtime to use an allocator

certik:
Just the last 4 commits. Now the runtime uses either an allocator supplied by the caller (compiler-generated code) or internal_malloc/free. Nothing else. After this it is thus easy to implement a leak checker in both the internal runtime, or the new allocator.


certik:
I think it's fine. There is a tiny overhead of calling the runtime function to get the allocator, but even that can be improved later. Now it is explicit where runtime must allocate for the compiler and we can likely slowly even remove such allocations completely and do it in ASR.

PR #10684, opened 2026-03-17
feat: implement `atan2d` intrinsic function

HarshilShah1804:


HarshilShah1804:
We don't require the exhaustive checks on this PR since it is a new feature for lfortran and no third-party code that we compile, would have used it.

HarshitaKalani:
Let's resolve the conflicts.

PR #10689, opened 2026-03-17
Add a leak checker to internal allocs in the runtime library

certik:
And fix all leaks. Now all the internal allocations are always freed (exactly once). The internal and compiler-generated allocations are now strictly separated.

The remaining task is to now also leak check the compiler-generated allocations and fix all bugs.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8663,7 +8663,8 @@ public:
                                     int dim_desc_size = (int)data_layout.getTypeAllocSize(dim_desc_type);
                                     builder->CreateMemCpy(target_dim_ptr, llvm::MaybeAlign(8), value_dim_ptr, llvm::MaybeAlign(8), dim_desc_size*n_dims);
                                     // Copy offset
-                                    builder->CreateStore(arr_descr->get_offset(value_array_desc_type, llvm_value),
+                                    llvm::Value* src_offset = arr_descr->get_offset(value_array_desc_type, llvm_value);
+                                    builder->CreateStore(src_offset,
                                         arr_descr->get_offset(target_array_desc_type, llvm_target_, false));
                                     // Other fields of the array descriptor should be already set.
                                     return;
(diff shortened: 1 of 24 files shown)


src/libasr/utils.h
```diff
@@ -143,6 +143,7 @@ struct CompilerOptions {
     bool legacy_array_sections = false;
     bool ignore_pragma = false;
     bool stack_arrays = false;
+    bool internal_alloc_check = false;
```
assem2002, line 146: If we're using macros this won't be needed.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5429,6 +5429,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         start_new_block(proc_return);
         llvm_symtab_finalizer.finalize_symtab(x.m_symtab);
         free_heap_fixed_size_arrays();
+        {
```
assem2002, line 5432: can't this call to `_lfortran_internal_alloc_finalize` be wrapped around the macro to avoid that call in normal mode (no checks mode)?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -17468,6 +17480,28 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         print_error(context, *module, *builder, args);
 
         /* EXIT WITH CODE */
+        {
+            {
+                llvm::Function *fn_free_argv = module->getFunction("_lpython_free_argv");
```
assem2002, line 17486: I think we should use arena (the new allocator interface) and not internal allocations with `_lpython_free_argv` and `_lpython_call_initial_functions` -- so inserting this free before every exist doesn't seem necessary.

src/libasr/codegen/asr_to_llvm.cpp
certik: It used to be, but then reference results will change based on compilation option, which will make a big mess --- you generate reference tests locally, then they fail because you had the other option than the CI. So I changed this to always generate it.

src/libasr/codegen/asr_to_llvm.cpp
certik: Or we can just use a fixed array, to avoid any allocation completely.

src/libasr/utils.h
certik: Yes, I used to switch based on this, but reverted it.

src/libasr/codegen/asr_to_llvm.cpp
assem2002: I think it's fair compromise then.

certik:
Tests passed, I resolved conflicts and merged it.

src/libasr/codegen/asr_to_llvm.cpp
certik: I fixed this in https://github.com/lfortran/lfortran/pull/10692.

certik:
This enables it at the CI: https://github.com/lfortran/lfortran/pull/10693

Issue #1966, opened 2023-07-10
Implement ISO_Fortran_binding.h


certik:
The Fortran 2018 standard implements ISO_Fortran_binding.h. Now when we have physical arrays, it's the `ISODescriptorArray` of `array_physical_type`, so we just have to provide casts, and then some C header file I think.

certik:
Yes, we handle that by the new [ArrayPhysicalCast](https://github.com/lfortran/lfortran/blob/4a8f2eb04408f6623ad58d02e39b65a4c51ef1b2/src/libasr/ASR.asdl#L335) node. We already do this for pointer arrays vs descriptor arrays and now we just add support for `ISODescriptorArray`.

awvwgk:
For this it could be useful to consolidate the headers in GFortran, LLVM flang, Intel, NAG and co. and decide whether there is an ABI for the CFI descriptors LFortran should be compatible with.

ivan-pi:
Great point. Given LFortran can reuse gfortran module files, it would make sense to have compatibility here.

certik:
Is Flang binary compatible with GFortran for this ISO descriptor?

ivan-pi:
Short answer is NO, based on the layout of the descriptor struct:

The flang descriptor is:
```c
/* 18.5.3 generic data descriptor */
typedef struct CFI_cdesc_t {
  /* These three members must appear first, in exactly this order. */
  void *base_addr;
  size_t elem_len; /* element size in bytes */
  int version; /* == CFI_VERSION */
  CFI_rank_t rank; /* [0 .. CFI_MAX_RANK] */
  CFI_type_t type;
  CFI_attribute_t attribute;
  unsigned char f18Addendum;
#ifdef __cplusplus
  cfi_internal::FlexibleArray<CFI_dim_t> dim;
#else
  CFI_dim_t dim[]; /* must appear last */
#endif
} CFI_cdesc_t;
```
whereas the `gfortran` descriptor is
```c
typedef struct CFI_cdesc_t
 {
    void *base_addr;
    size_t elem_len;
    int version;
    CFI_rank_t rank;
    CFI_attribute_t attribute;
    CFI_type_t type;
    CFI_dim_t dim[];
 }
CFI_cdesc_t;
```

The NAG descriptor is:
```c
typedef struct {
  void *base_addr;
  size_t elem_len;
  int version;
  CFI_rank_t rank;
  CFI_type_t type;
  CFI_attribute_t attribute;
[...]

PR #10694, opened 2026-03-18
Extend bindc_18 CFI descriptor test to cover multiple types and ranks

certik:
Towards https://github.com/lfortran/lfortran/issues/1966.


certik:
Here is a more advanced version: https://github.com/lfortran/lfortran/pull/10695.

PR #10699, opened 2026-03-18
feat: implement `sinpi`, `cospi`, `tanpi`, `asinpi`, `acospi`, `atanpi`, `atan2pi`

HarshilShah1804:


HarshilShah1804:
These intrinsics are currently not supported by `gfortran`, `flang` or `ifort`. It is only supported by `ifx` for now.

Issue #3725, opened 2024-03-28
[stats/mean]: Fix number of arg handling for `reshape`


HarshitaKalani:
Ref: https://github.com/lfortran/lfortran/issues/3724
```fortran
program main
  integer(8) , parameter :: d1_int8(18) = [-10, 2, 3, 4, -6, 6, -7, 8, 9, 4, 1, -20, 9, 10, 14, 15, 40, 30]
  integer(8) :: d4_int8(3, 2, 3, 2) = reshape(d1_int8, [3, 2, 3, 2], [integer(8) :: 3])
end
```
```console
$ gfortran a.f90 && ./a.out
$ lfortran a.f90 
semantic error: reshape accepts only 2 arguments, got 3 arguments instead.
 --> a.f90:3:39
  |
3 |   integer(8) :: d4_int8(3, 2, 3, 2) = reshape(d1_int8, [3, 2, 3, 2], [integer(8) :: 3])
  |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear or confusing messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Pranavchiku:
Try to see if it can be worked around, else fix this.

Issue #3117, opened 2024-01-15
Support `ieee_is_nan()` function


HarshitaKalani:
```fortran
function median_1_rdp_dp(x, dim) result(res)
    integer, parameter :: dp = selected_real_kind(15)
    real(dp), intent(in) :: x(:)
    integer, intent(in) :: dim
    real(dp) :: res

    integer :: n
    real(dp), allocatable :: x_tmp(:)

    n = size(x, dim)

    allocate(x_tmp(n))
    x_tmp(:) = x(:)
    if (any(ieee_is_nan(x_tmp))) then
        print*, "okay"
        return
    end if
end function median_1_rdp_dp
program main
end
```
```console
$ lfortran a.f90
ASR verify pass error: ASR verify: Input to Any intrinsic must be of logical type, found: Allocatable[r64[:]]
  --> a.f90:14:9
   |
14 |     if (any(ieee_is_nan(x_tmp))) then
   |         ^^^^^^^^^^^^^^^^^^^^^^^ failed here


Note: Please report unclear or confusing messages as bugs at
https://github.com/lfortran/lfortran/issues.
$ gfortran a.f90 && ./a.out
a.f90:14:12:

   14 |     if (any(ieee_is_nan(x_tmp))) then
      |            1
Error: ‘mask’ argument of ‘any’ intrinsic at (1) must be a logical array
```

MRE:
```fortran
function median_1_rdp_dp(x, dim) result(res)
    real(8), intent(in) :: x(:)
    integer, intent(in) :: dim
    real(8) :: res
    integer :: n
    real(8), allocatable :: x_tmp(:)
    n = size(x, dim)
    allocate(x_tmp(n))
    x_tmp(:) = x(:)
    print*, ieee_is_nan(x_tmp)
end function median_1_rdp_dp
program main
end
```
```console
$ gfortran a.f90 && ./a.out
/usr/bin/ld: /tmp/cc5RKpq5.o: in function `median_1_rdp_dp_':
a.f90:(.text+0x360): undefined reference to `ieee_is_nan_'
collect2: error: ld returned 1 exit status
$ lfortran a.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"
.
.
code generation error: asr_to_llvm: module failed verification. Error:
Stored value type does not match pointer operand type!
  store i1 %246, double* %227, align 1
 double
```

HarshitaKalani:
```fortran
program main
    print*, ieee_is_nan(1.0)
end
```
[...]

PR #10701, opened 2026-03-18
fix bugs 

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/3152
fixes: https://github.com/lfortran/lfortran/issues/3725
fixes: https://github.com/lfortran/lfortran/issues/3117
fixes: https://github.com/lfortran/lfortran/issues/3475

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -631,7 +631,7 @@ program continue_compilation_1
     integer :: maxloc_shape_mismatch = maxloc([2, 1, 3], 1, [.true., .false.])
     write (*, "(a)", advance="hello") "Dothraki culture"
     print *, sum(arr1, dim = mask1)
-
+    print*, ieee_is_nan(1.0)
 
 
 
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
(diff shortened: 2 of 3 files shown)


certik:
Please rebase

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -20884,7 +20884,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             if (compiler_options.po.bounds_checking) {
                 for (size_t i = 0; i < x.n_args; i++) {
                     ASR::ttype_t* arg_expr_type = ASRUtils::expr_type(x.m_args[i]);
-                    if (ASRUtils::is_allocatable(arg_expr_type)) {
+                    bool is_alloc_string = ASRUtils::is_allocatable(arg_expr_type) &&
+                        ASRUtils::is_character(*ASRUtils::type_get_past_allocatable_pointer(arg_expr_type));
+                    if (ASRUtils::is_allocatable(arg_expr_type) && !is_alloc_string) {
```
certik, line 20889: This part is removing a bounds check, is this a good change?

HarshitaKalani: You're right. This change removes a bounds check which is bad.
The issue actually comes from two bugs in the original handling of allocatable CHARACTER arguments:
- Double evaluation of expressions:
expr_is_unallocated evaluates the argument expression, and the same expression is evaluated again later (e.g., for length and formatting).
This leads to function calls being executed multiple times, which is incorrect.

- Incorrect null check for strings:
The code extracts the raw i8* data pointer from the string descriptor, but compares it against a null pointer of the descriptor type, causing a type mismatch in LLVM IR.

I'll push a PR to fix this shortly.

certik: https://github.com/lfortran/lfortran/issues/10734

PR #10708, opened 2026-03-18
More F18 bind(c) tests and fixes

certik:
All tests work with LFortran, Flang and GFortran.


certik:
Ubuntu LLVM 21 bindc_23 fails with `--fast`.

PR #10716, opened 2026-03-19
Fix allocate with source calling function multiple times

### LINKED ISSUE #10698: Sourced allocation of array by array-valued function leads to superfluous function calls
`$ cat main.f90`
```
program main

   type MyType
   end type MyType
      
   type(MyType), allocatable :: arr(:)

   allocate(arr, source = allocator())

contains
   
   function allocator() result(arr)
      type(MyType), allocatable :: arr(:)      
      write(*,*) 'This should be printed once!'
      allocate(arr(9))
   end function allocator
   
end program main
```
`$ lfortran main.f90`
```
This should be printed once!
This should be printed once!
This should be printed once!
This should be printed once!
```
With flang-20 the output is:
```
 This should be printed once!
```

certik:
When allocate(arr, source=func_call()) was used, the function call expression was duplicated into ArrayBound, ArraySize, Allocate source, and Assignment nodes, causing the function to be evaluated 4 times instead of once.

The fix materializes the source expression into a temporary variable when it is a FunctionCall, so the function is called exactly once and the temporary is reused for dimension extraction, allocation, and data copying.

An integration test (allocate_60) is added to verify the fix.

Fixes #10698.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2260,6 +2260,7 @@ RUN(NAME allocate_56 LABELS gfortran llvm)
 RUN(NAME allocate_57 LABELS gfortran llvm)
 RUN(NAME allocate_58 LABELS gfortran llvm)
 RUN(NAME allocate_59 LABELS gfortran llvm)
+RUN(NAME allocate_60 LABELS gfortran llvm)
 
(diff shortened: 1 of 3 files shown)


certik:
Note: this PR introduced a bug which is fixed by https://github.com/lfortran/lfortran/pull/10726.

PR #10727, opened 2026-03-19
fix: correctly skip unallocated bounds check for allocatable strings in print

### LINKED ISSUE #10734: Add the bounds check back
> You're right. This change removes a bounds check which is bad.
> The issue actually comes from two bugs in the original handling of allocatable CHARACTER arguments:
> - Double evaluation of expressions:
> expr_is_unallocated evaluates the argument expression, and the same expression is evaluated again later (e.g., for length and formatting).
> This leads to function calls being executed multiple times, which is incorrect.
> 
> - Incorrect null check for strings:
> The code extracts the raw i8* data pointer from the string descriptor, but compares it against a null pointer of the descriptor type, causing a type mismatch in LLVM IR.
> 
> I'll push a PR to fix this shortly. 

Let's also add a test for the bounds check once you add it back. It wasn't tested, so we didn't notice it was removed.

 _Originally posted by @HarshitaKalani in [#10701](https://github.com/lfortran/lfortran/pull/10701/changes#r2960858654)_

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/pull/10701/changes#r2960445612
fixes: https://github.com/lfortran/lfortran/issues/10734


tests/errors/scalar_allocation_check_07.f90
```diff
@@ -0,0 +1,6 @@
+program scalar_allocation_check_07
+    implicit none
+    character(len=:), allocatable :: s
+
+    print *, s
```
certik, line 5: Is this the error that got skipped?

certik: I just tested it and indeed in main we don't print anything.

certik (approved):
Looks good.

tests/errors/scalar_allocation_check_07.f90
HarshitaKalani: Yes

HarshitaKalani:
Unrelated:
```
Confirm changes: [Y/n] 
Transaction starting
error    libmamba Failed to download package from https://conda.anaconda.org/conda-forge/linux-64/fpm-0.12.0-h3b12eaf_0.conda (status 403)
error    libmamba Failed to download package from https://conda.anaconda.org/conda-forge/linux-64/fpm-0.12.0-h3b12eaf_0.conda (status 403)
critical libmamba Transfer finalized, status: 403 [https://conda.anaconda.org/conda-forge/linux-64/fpm-0.12.0-h3b12eaf_0.conda] 4530 bytes
```

certik (approved):

tests/reference/run-scalar_allocation_check_05-e3b524c.stderr
```diff
@@ -1,5 +1 @@
-runtime error: Argument 1 is unallocated.
-  --> tests/errors/scalar_allocation_check_05.f90:10:14
-   |
-10 |     print *, m1%x
-   |              ^^^^ This is unallocated
+Internal Error : default formatting error
```
certik: This doesn't seem right.

certik:
Sorry, I totally missed it. This PR adds one good bounds check, but breaks another bounds check, so technically a wash, but let's not break things. Let's only fix the bounds check without breaking anything.

https://github.com/lfortran/lfortran/pull/10727#discussion_r2966319383

HarshitaKalani:
I think it's good now. Can you please review? Thanks!

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -21784,13 +21780,24 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            ASR::symbol_t* var_sym = ASR::down_cast<ASR::Var_t>(x.m_args[i])->m_v;
+                            var_sym = ASRUtils::symbol_get_past_external(var_sym);
+                            if (!ASR::is_a<ASR::Variable_t>(*var_sym)) {
+                                continue;
+                            }
+                            ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(var_sym);
+                            const std::string var_name(var->m_name);
+                            if (var_name.find("__libasr__created__var__") == 0) {
+                                continue;
+                            }
```
certik, line 21833: I don't like this string comparison. Why do we need to skip some variables and only with this name? Why not check everything?

And if we need to skip some, maybe let's encode it more explicitly in ASR as flags?

HarshitaKalani: This is to detect compiler-generated temporary variables and skip the bounds check for them. I can introduce `bool is_compiler_generated` as an explicit flag directly to Variable in ASR.asdl. Would that be better?

Issue #10728, opened 2026-03-19
Formatted WRITE: E format printing asterisks, when a value could be output


w6ws:
```
program write_fmt1
  implicit none

  real :: a = -0.0044
  character(6) :: a_string

  write (a_string, '(e6.1e1)') a
  print *, 'a_string = ', a_string, merge (': pass', ': fail', a_string == '-.4E-2')

end program
```
```
$ lfortran write_fmt1.f90
a_string = ******: fail
$
```
```
$ gfortran write_fmt1.f90
(base) wws@w6ws-4:~/computer/fortran/fcvs/fcvs21_f95/wws$ ./a.out
 a_string = -.4E-2: pass
$ flang write_fmt1.f90
$ ./a.out
 a_string = -.4E-2: pass
$ 
```

Jatinagarwal24:
Hi,@w6ws I am on it.

w6ws:
Tested and looks good.  Thanks!

PR #10731, opened 2026-03-19
fix: Fix E format descriptor width calculation to prevent unnecessary asterisk output

Jatinagarwal24:
fix: Correctly calculate E format field width to avoid unnecessary asterisk output

Previously, the code used an overly strict overflow check:
  if (strlen(formatted_value) - width > 1)
This caused valid values to be rejected with asterisks when they could fit.

The fix properly handles edge cases:
- Values that fit by dropping the leading zero (e.g., "0.4" → ".4")
- Negative values that fit by compacting notation (e.g., "-0.4" → "-.4")
- Only outputs asterisks when the value genuinely cannot fit

Example: -0.0044 with format '(e6.1e1)' now correctly outputs '-.4E-2'
instead of '******'.

Resolves: #10728


Jatinagarwal24:
Hi,@certik pls review this.

Issue #10729, opened 2026-03-19
Formatted READ: Incorrect scaling with 0P scale factor


w6ws:
```
program read_fmt14
  implicit none

  real :: a
  character(9) :: a_string

  a_string = '  987.654'
  read (a_string, '(0PF9.4)') a
  print *, 'a =', a, pf (abs (a - 987.654) < 0.001)

contains

  function pf (l)
    character(6) :: pf
    logical, intent(in) :: l
    pf = merge (': pass', ': fail', l)

  end function

end program
```
```
$ lfortran read_fmt14.f90
a =    98.7654037    : fail
$
```
```
$ gfortran read_fmt14.f90
$ ./a.out
 a =   987.653992     : pass
$
$ flang read_fmt14.f90
$ ./a.out
 a =    987.6540     : pass
$
```

Jatinagarwal24:
Hi,@w6ws I am on it.

w6ws:
Tested and working great.  Thanks!

PR #10733, opened 2026-03-19
fix: parse explicit 0P scale factor in formatted F input to preserve decimal values

Jatinagarwal24:
fix: Handle 0P scale factor in formatted read by tracking explicit repeat counts

Previously, the code unconditionally reset repeat_count from 0 to 1:
  if (repeat_count == 0) repeat_count = 1;

This broke the 0P scale factor, which uses repeat_count = 0 to indicate
"scale factor = 0", not "repeat 0 times".

The fix introduces has_repeat_count to distinguish:
- Explicit repeat counts (has_repeat_count = true)
- Scale factors with 0 (has_repeat_count = false, repeat_count = 0)

Only apply the default repeat_count = 1 when has_repeat_count is false,
preserving the 0P scale factor behavior.

Result: '(0PF9.4)' now correctly reads 987.654 without incorrect scaling.

Resolves: #10729

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1077,7 +1077,7 @@ RUN(NAME format_72 LABELS gfortran llvm)
 RUN(NAME format_73 LABELS gfortran llvm)
 RUN(NAME format_74 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_75 LABELS gfortran llvm)
-
+RUN(NAME format_76 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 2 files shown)


HarshitaKalani:
```
The following tests FAILED:
	616 - format_71 (Failed)
	2869 - formatted_read_02 (Failed)
```

Jatinagarwal24:
Hi,@certik pls review this.

certik:
Please resolve conflicts. Then let's run full tests.

Jatinagarwal24:
> Please resolve conflicts. Then let's run full tests.

Done,Pls run full tests.

Jatinagarwal24:
Hi,@certik I think this is good to merge.

PR #10736, opened 2026-03-20
Fix last bind(c) test

certik:


certik:
After this we should fix the descriptor: https://github.com/lfortran/lfortran/issues/10737

Issue #10730, opened 2026-03-19
Formatted WRITE: False 'Unbalanced paranthesis in format string' error at run-time


w6ws:
(Also note the incorrect spelling of 'parenthesis' in the error message.)
```
program write_fmt2
  implicit none

  write (*,100) 1, 2, 3, 4

100 format (27X,I4," **",1(27X,I4," ''",(I4," ((")))

end program
```
```
$ lfortran write_fmt2.f90
Error: Unbalanced paranthesis in format string
$
```
```
$ gfortran write_fmt2.f90
$ ./a.out
                              1 **                              2 ''   3 ((
                              4 ''
$
$ flang write_fmt2.f90
$ ./a.out
                              1 **                              2 ''   3 ((
                              4 ''
$
```

Jatinagarwal24:
Hi, @w6ws I am on it.

w6ws:
Tested and looks good.  Thanks!

PR #10738, opened 2026-03-20
fix:handle parentheses inside quoted literals in format strings

Jatinagarwal24:
fix: Ignore parentheses inside quoted strings when validating format balance

Previously, the code counted all '(' and ')' characters in format strings,
including those inside quoted strings like "(^)" or "***".

Example failure: format (27X,I4," ***",1(27X,I4," ''',(I4," (^)))
The ')' inside "(^)" was incorrectly counted, causing false
"Unbalanced paranthesis" errors.

The fix tracks quote state:
- Set in_quotes = true when entering ' or " delimited strings
- Set in_quotes = false when exiting quoted strings
- Only count '(' and ')' when NOT in quotes
-Also corrected the spelling of parenthesis.
Result: Format strings with quoted parentheses now validate correctly.

Resolves: #10730

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1075,7 +1075,7 @@ RUN(NAME format_71 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_72 LABELS gfortran llvm)
 
 RUN(NAME format_73 LABELS gfortran llvm)
-RUN(NAME format_74 LABELS gfortran llvm)
+RUN(NAME format_74 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME submodule_01 LABELS gfortran)
 RUN(NAME submodule_02 LABELS gfortran fortran)


integration_tests/CMakeLists.txt
```diff
@@ -1075,6 +1075,7 @@ RUN(NAME format_71 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME format_72 LABELS gfortran llvm)
 
 RUN(NAME format_73 LABELS gfortran llvm)
+RUN(NAME format_74 LABELS gfortran llvm)
```
```suggestion
RUN(NAME format_74 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
```
Lets always add wasm labels as well.

Jatinagarwal24:
Hi,@certik Pls review this.I think this is good to merge.

integration_tests/format_74.f90
```diff
@@ -0,0 +1,23 @@
...
+
+  open (newunit=u, file='format_74_data.txt', status='old', form='formatted')
+  read (u, '(A)') line1
+  read (u, '(A)') line2
+  close (u)
+
+  if (index(line1, '1 **') == 0) error stop "missing first item"
+  if (index(line1, "2 ''") == 0) error stop "missing quoted literal"
+  if (index(line1, '3 ((') == 0) error stop "missing literal parentheses"
+  if (index(line2, "4 ''") == 0) error stop "missing reverted-format item"
```
certik, line 19: I think these only add tests for what it should not be, but do not test what it should be.

Issue #9276, opened 2025-12-30
Semantic error highlights entire DO loop instead of undeclared loop variable


HarshitaKalani:
Ref: https://github.com/lfortran/lfortran/pull/9009#discussion_r2614711232
```console
% cat b.f90
  integer n(3)
  do i = 1,size(n)
     n(i) = 42
  end do
  print "(3I3)", n
end program
% lfortran b.f90
semantic error: Variable 'i' is not declared
 --> b.f90:2:3 - 4:8
  |
2 |      do i = 1,size(n)
  |      ^^^^^^^^^^^^^^^^...
...
  |
4 |      end do
  | ...^^^^^^^^ 'i' is undeclared


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

PR #10741, opened 2026-03-20
fix: point undeclared DO loop variable error to the variable, not the whole loop

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/9276

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -654,4 +654,11 @@ program continue_compilation_1
             end function
         end interface
     end subroutine
+    subroutine sub_do_undeclared()
+        implicit none
+        integer :: n(3)
+        do k = 1, 3
+            n(k) = 42
+        end do
+    end subroutine
 end program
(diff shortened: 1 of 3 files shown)


src/lfortran/parser/semantics.h
```diff
@@ -2060,11 +2060,11 @@ void add_ws_warning(const Location &loc,
 #define DO2(i, a, b, trivia, body, l) make_DoLoop_t(p.m_a, l, 0, nullptr, 0, \
         name2char(i), EXPR(a), EXPR(b), nullptr, \
         /*body*/ STMTS(body), \
-        /*n_body*/ body.size(), trivia_cast(trivia), nullptr)
+        /*n_body*/ body.size(), trivia_cast(trivia), nullptr, &((i)->loc))
```
certik, line 2063: Is this using a pointer to a location struct? Who owns the struct? Will it become dangling?

certik:
Otherwise it's good, but I am worried about the dangling pointer.

src/lfortran/parser/semantics.h
HarshitaKalani: This is safe. `i` is allocated in the parser arena (p.m_a), which lives through both parsing and semantic analysis. The pointer is only used during the AST → ASR pass, so it does not outlive the arena.

This pattern is also already used elsewhere (e.g., start_name / end_name storing &(name->loc) and used during semantic analysis), so this follows the existing design.

certik: The ASR should not depend on anything in AST's arena. Right now we might be using the same arena, but later on when we compile the whole project at once, we parse, append to ASR and likely free the AST, at least we want to have the option to free it.

If this is already used elsewhere, then it's probably fine for now. But I am not yet convinced this is the right approach. I think it's not a blocker though.

Issue #10745, opened 2026-03-20
Formatted READ: Misaligned read of complex values


w6ws:
```
program read_fmt15
  implicit none

  complex :: c1, c2, c3
  complex :: c_diff

  character(30) :: cdata

  !                 1         2
  !        1234567890123456789012345678
  cdata = '1.05.522.066.633.123455.0789'
  read (cdata, 100) c1, c2, c3
100 format  ( 2(F3.1) , 2(F4.1), 2(F7.4))

  c_diff = abs (c1 - (1.0,5.5))
  print *, 'c1 =', c1, pf (real (c_diff) < 0.0001 .and. aimag (c_diff) < 0.0001)
  c_diff = abs (c2 - (22.0,66.6))
  print *, 'c2 =', c2, pf (real (c_diff) < 0.0001 .and. aimag (c_diff) < 0.0001)
  c_diff = abs (c3 - (33.1234,55.0789))
  print *, 'c3 =', c3, pf (real (c_diff) < 0.0001 .and. aimag (c_diff) < 0.0001)


contains

  function pf (l)
    character(6) :: pf
    logical, intent(in) :: l
    pf = merge (': pass', ': fail', l)
  end function

end program
```
```
$ lfortran read_fmt15.f90
c1 =    (1.00000000,5.50000000)    : pass
c2 =    (22.0000000,6.59999990)    : fail
c3 =    (0.633000016,0.123000003)    : fail
$
```
```
$ gfortran read_fmt15.f90
$ ./a.out
 c1 =             (1.00000000,5.50000000) : pass
 c2 =             (22.0000000,66.5999985) : pass
 c3 =             (33.1234016,55.0788994) : pass
$ flang read_fmt15.f90
$ ./a.out
 c1 =  (1.000000,5.500000) : pass
 c2 =  (22.00000,66.60000) : pass
 c3 =  (33.12340,55.07890) : pass
$
```

Jatinagarwal24:
Hi,@w6ws I am on it.

w6ws:
Tested and working much better.  Thanks!

PR #10754, opened 2026-03-20
fix: Handle complex types correctly in formatted read operations

Jatinagarwal24:
Previously, formatted read operations didn't support complex types, causing
incorrect values when reading complex numbers with format descriptors.

The fix adds:
1. Complex type code generation (lines 16003-16012)
    Splits complex into real/imaginary components
    Creates separate pointers for each component
    Pushes both components to args list

2. get_formatted_read_arg_count() helper (lines 16018-16021)
   Returns 2 for complex (real + imaginary)
   Returns 1 for other types

3. Proper arg counting for complex arrays and scalars
   Uses arg_count multiplier throughout
   Ensures each complex consumes 2 format positions

Result: Format '(2(F3.1), 2(F4.1), 2(F7.4))' now correctly reads:
  c1 = (1.0, 5.5), c2 = (22.0, 66.6), c3 = (33.1234, 55.0789)

Resolves: #10745


Jatinagarwal24:
Hi @certik I think this is good to merge

PR #10761, opened 2026-03-21
Fix CFI descriptor handling: correct load type, hoist alloca, remove unnecessary memset

certik:
## Summary

Three targeted fixes to the LLVM codegen for CFI (C Fortran Interoperability) descriptors:

1. **Use descriptor type for namelist allocatable/pointer load** — When loading an allocatable/pointer array in namelist I/O, use the actual descriptor type instead of `i8*`. This ensures the loaded pointer has the correct LLVM type for subsequent operations.

2. **Hoist `internal_to_cfi` alloca to function entry block** — Use `CreateAlloca` without the builder argument so the alloca is placed in the function entry block rather than inside a loop or conditional, following LLVM best practices for stack allocations.

3. **Remove unnecessary memset from `allocate_descriptor_on_heap`** — The descriptor fields are individually initialized immediately after allocation, making the zero-fill memset redundant.

## Scope

- `src/libasr/codegen/asr_to_llvm.cpp` — Fix the load type for allocatable/pointer arrays in namelist handling.
- `src/libasr/codegen/llvm_array_utils.cpp` — Hoist alloca; remove memset.
- Updated reference test output for `allocate_03` to reflect the removed memset.

## Verification

Reference tests updated. The `allocate_03` LLVM output no longer contains the `memset` call or the associated `argmemonly` attribute declaration.

### DIFF
--- a/tests/reference/llvm-allocate_03-495d621.stdout
+++ b/tests/reference/llvm-allocate_03-495d621.stdout
@@ -1134,6 +1134,8 @@ then12:                                           ; preds = %ifcont11
   %48 = bitcast i8* %47 to %array.3*
   %49 = getelementptr %array.3, %array.3* %48, i32 0, i32 3
   store i8 3, i8* %49, align 1
+  %50 = getelementptr %array.3, %array.3* %48, i32 0, i32 0
+  store i32* null, i32** %50, align 8
   store %array.3* %48, %array.3** %c, align 8
   br label %ifcont14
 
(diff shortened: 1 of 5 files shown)


tests/reference/llvm-allocate_03-495d621.stdout
```diff
@@ -1131,249 +1131,250 @@ ifcont11:                                         ; preds = %merge_allocated8
 then12:                                           ; preds = %ifcont11
   %46 = call i8* @_lfortran_get_default_allocator()
   %47 = call i8* @_lfortran_malloc_alloc(i8* %46, i64 104)
-  call void @llvm.memset.p0i8.i32(i8* %47, i8 0, i32 104, i1 false)
   %48 = bitcast i8* %47 to %array.3*
   %49 = getelementptr %array.3, %array.3* %48, i32 0, i32 3
   store i8 3, i8* %49, align 1
+  %50 = getelementptr %array.3, %array.3* %48, i32 0, i32 0
+  store i32* null, i32** %50, align 8
```
certik, line 1138: I wonder if the nullptr store pretty much replaces the memset and either one of them is necessary.

certik:
This is good. We should refactor the alloca: https://github.com/lfortran/lfortran/issues/10789.

PR #10762, opened 2026-03-21
Switch internal array descriptor to rank-sized dim arrays and fix LLVM type issues

### LINKED ISSUE #10758: Plan: Switch internal descriptor to dim[n_dims]
## Problem

The internal array descriptor currently uses `dim[0]` (flexible array member
pattern, implemented in https://github.com/lfortran/lfortran/pull/10757), requiring a custom `create_descriptor_alloca` that manually computes
`header_size + n_dims * dim_size`. The CFI descriptor already uses
`dim[n_dims]` (fully sized type). Since the rank is always known at compile
time for both, the internal descriptor should use the same approach.

## Proposed Change

Change the internal array descriptor struct from `[0 x %dimension_descriptor]`
to `[n_dims x %dimension_descriptor]`, producing a different struct type per
rank (e.g., `array.1`, `array.2`, `array.3`).

## Benefits

- Eliminates `create_descriptor_alloca` — regular `llvm_utils->CreateAlloca(type)` works
- One consistent pattern for both internal and CFI descriptors
- LLVM can reason about struct sizes — better optimization and bounds checking
- Simpler code at every allocation site

## Scope

- `llvm_array_utils.cpp`: Change `create_array_type` to accept rank and emit `dim[n_dims]`
- `llvm_array_utils.cpp`: Remove `create_descriptor_alloca`, replace all call sites with regular alloca
- `llvm_array_utils.cpp`: Update `allocate_descriptor_on_heap` — type is now fully sized
- `asr_to_llvm.cpp`: Update all descriptor allocation sites (~20) to use the rank-specific type
- Reference tests: Update for new IR patterns

certik:
## Summary

Switch the internal LLVM array descriptor struct from using `dim[0]` (zero-length trailing array) to `dim[n_dims]` (rank-sized arrays), producing one distinct LLVM struct type per rank (e.g. `%array.1`, `%array.2`). This aligns the internal descriptor layout with the CFI descriptor and eliminates manual byte-size calculations for descriptor allocation.

The foundational change required several follow-up fixes to ensure correct LLVM types are used throughout the codegen, particularly around opaque pointer GEPs and descriptor casts.

## Changes

1. **Switch internal array descriptor from `dim[0]` to `dim[n_dims]`** — the core change. Descriptor struct types are now cached by rank, named `array.N`. Simplifies `create_descriptor_alloca` / `allocate_descriptor_on_heap` (no longer need `n_dims`). Adds bitcasts at call sites when ranks differ (e.g. assumed-rank parameters). Finalizer cache keys now include rank.

2. **Remove `getPointerElementType()` calls from `llvm_array_utils`** — this API was removed in newer LLVM versions (opaque pointers). Descriptor struct types are now passed explicitly to `get_pointer_to_data`, `fill_descriptor_for_array_section`, and `reshape`.

3. **Fix `get_pointer_to_data` to strip Allocatable/Pointer wrappers** — `get_type_from_ttype_t_util` returns a pointer type for Allocatable arrays, but `get_pointer_to_data` needs the descriptor struct type for GEP.

4. **Fix StructInstanceMember load for allocatable array sections** — load the pointer (not the full struct) when dereferencing allocatable arrays inside structs for array sections.

5. **Fix AssumedRankArray→PointerArray cast using wrong type for GEP** — use the source descriptor struct type, not the target type.

6. **Fix FixedSizeArray←DescriptorArray cast using wrong type for `get_pointer_to_data`**.

7. **Cache array descriptor types in `get_array_type_for_rank`** — avoid recreating identical struct types.

## Files Changed

- `src/libasr/codegen/asr_to_llvm.cpp` — descriptor allocation, array casts, call-site fixups
- `src/libasr/codegen/llvm_array_utils.cpp` — descriptor struct creation, GEP operations, section fills
- `src/libasr/codegen/llvm_array_utils.h` — API changes (explicit type params, caching)
- `src/libasr/codegen/llvm_utils.cpp`, `llvm_utils.h` — minor signature updates
- `tests/reference/` — updated LLVM IR reference outputs (5 tests)

Fixes #10758


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -341,6 +341,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+    // With rank-sized array descriptors, different ranks produce different
+    // LLVM struct types. Bitcast when types are structurally compatible
+    // but nominally different (e.g., %array.1* vs %array.15* for assumed-rank).
+    void fixup_call_arg_types(std::vector<llvm::Value*>& args, llvm::FunctionType* fntype) {
+        for (size_t i = 0; i < args.size() && i < fntype->getNumParams(); i++) {
+            if (args[i]->getType() != fntype->getParamType(i)) {
+                args[i] = builder->CreateBitCast(args[i], fntype->getParamType(i));
+            }
+        }
+    }
```
certik: This should not be there, we should use casts in ASR. I am working on it.

certik: Fixed in 3d345409f9.

certik:
The assumed-rank will still need to be refactored, but we can do that later: https://github.com/lfortran/lfortran/issues/10775.

certik:
Another follow-up issue: https://github.com/lfortran/lfortran/issues/10778.

Issue #10772, opened 2026-03-21
Formatted READ: Double precision value not scaled correctly


w6ws:
```
program read_fmt17
  implicit none

  character(10) :: s
  double precision :: da

  s = '12345D+01'
  read (s, '(f10.5)') da
  print *, 'da =', da, merge (': pass', ': fail', abs (da - 1.2345d0) < 0.0001d0)
  write (*, '(f6.4)') da

end program
```
```
$ lfortran read_fmt17.f90
da =    123450.00000000000    : fail
******
$
```
```
$ gfortran read_fmt17.f90
$ ./a.out
 da =   1.2344999999999999     
1.2345
$
$ flang read_fmt17.f90
$ ./a.out
 da =    1.234500000000000      : pass
1.2345
$
```

Jatinagarwal24:
Hi ,@w6ws on it

w6ws:
Tested and works fine now.  Thanks!

PR #10773, opened 2026-03-21
fix: Apply scale factor correctly to double precision values in formatted read

Jatinagarwal24:
Previously, total_scale was initialized to scale_factor and then added
decimal_places, causing incorrect scaling of double precision values.

The old logic:
  total_scale = scale_factor + (has_decimal ? decimal_places : 0)

This incorrectly treated scale_factor and decimal_places as additive.

The fix initializes total_scale = 0, then:
1. Adds decimal_places if has_decimal (for precision)
2. Adds scale_factor (for P descriptor scaling)

This separates precision (from w.d) and scaling (from nP descriptor).

Result: '(f10.5)' format now correctly reads '12345D+01' as 1.2345
instead of 123450.0.

Resolves: #10772


integration_tests/CMakeLists.txt
```diff
@@ -3118,6 +3118,7 @@ RUN(NAME read_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME read_59 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME read_60 LABELS gfortran llvm)
+RUN(NAME read_61 LABELS gfortran llvm)
```
```suggestion
RUN(NAME read_60 LABELS gfortran llvm lvm_wasm llvm_wasm_emcc)
RUN(NAME read_61 LABELS gfortran llvm lvm_wasm llvm_wasm_emcc)
```
Can you please add wasm labels as well?

HarshitaKalani (approved):
Looks good. let's address the comment in subsequent PR. Thanks!

Jatinagarwal24:
Hi ,@certik I think this is good to merge.

Jatinagarwal24:
> Looks good. let's address the comment in subsequent PR. Thanks!

Hi,@HarshitaKalani I think this is good to merge now.

PR #10798, opened 2026-03-22
feat: implement `decimal` specifier in I/O functions

HarshilShah1804:
Towards #10677


HarshilShah1804:
Almost all the reference changes are because of the introduction of `decimal_mode` specifier in `_lcompilers_string_format_fortran` function. There is also another approach to solve this:
- declare a new global variable `decimal_mode` in the `lfortran_intrinsics.c` file and populate it before calling the `_lcompilers_string_format_fortran` function. 

This would remove the reference changes for LLVM. @certik Please share your thoughts thoughts on if we shall go with the alternative approach or this one is fine

src/libasr/ASR.asdl
```diff
@@ -51,13 +51,13 @@ stmt
     | If(identifier? name, expr test, stmt* body, stmt* orelse)
     | IfArithmetic(expr test, int lt_label, int eq_label, int gt_label)
     | Print(expr text)
-    | FileOpen(int label, expr? newunit, expr? filename, expr? status, expr? form, expr? access, expr? iostat, expr? iomsg, expr? action, expr? delim, expr? recl, expr? position, expr? blank, expr? encoding, expr? sign)
+    | FileOpen(int label, expr? newunit, expr? filename, expr? status, expr? form, expr? access, expr? iostat, expr? iomsg, expr? action, expr? delim, expr? recl, expr? position, expr? blank, expr? encoding, expr? sign, expr? decimal)
```
certik, line 54: Document what this new field does in doc/src/asr/asr_nodes/statement_nodes/, you might need to add a new file since FileOpen doesn't seem documented yet.

certik:
I think this is the right approach. Go ahead and rebase it and document it, then it's good.

src/libasr/ASR.asdl
HarshilShah1804: Will do it in a subsequent PR for both `FileOpen` and `FileInquire` nodes.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -810,6 +811,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                 }
+            } else if (m_arg_str == std::string("decimal")) {
+                if (a_decimal != nullptr) {
+                    diag.add(Diagnostic(
+                        R"""(Duplicate value of `decimal` found)""",
+                        Level::Error, Stage::Semantic, {
+                            Label("",{x.base.base.loc})
+                        }));
+                    throw SemanticAbort();
+                }
```
certik, line 822: Add a test for this error.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -810,6 +811,27 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                this->visit_expr(*kwarg.m_value);
+                a_decimal = ASRUtils::EXPR(tmp);
+                ASR::ttype_t *a_decimal_type = ASRUtils::expr_type(a_decimal);
+                if (!ASRUtils::is_character(*a_decimal_type)) {
+                    diag.add(Diagnostic(
+                        "`decimal` must be of type, String or StringPointer",
+                        Level::Error, Stage::Semantic, {
+                            Label("",{x.base.base.loc})
+                        }));
+                    throw SemanticAbort();
```
certik, line 832: Add a test

Issue #10805, opened 2026-03-23
ASR verify failure with random_seed inside block using outer-scope variable for array size


HarshitaKalani:
```console
% cat a.f90                
! Test random_seed with array arg dimensioned by outer-scope variable inside a block
program test_random_seed_block
    implicit none
    integer :: sz
    call random_seed(size=sz)
    if (sz <= 0) error stop "random_seed size must be positive"
    block
        integer :: vals(sz)
        vals = 42
        call random_seed(put=vals)
        call random_seed(get=vals)
    end block
end program
% gfortran a.f90 && ./a.out
% lfortran a.f90           
ASR verify pass error: The symbol table was not found in the scope of `symtab`.

ASR verify pass error: Var::m_v `sz` cannot point outside of its symbol table
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2809
    LCompilers::print_stack_on_segfault();
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2736
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 1250
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 397
    if (res.ok) {
  File "/Users/harshitakalani/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 22101
    }
  File "/Users/harshitakalani/repos/lfortran/src/libasr/pass/pass_manager.h", line 318
    apply_passes(al, asr, _passes, pass_options, diagnostics, cummulative_time_taken_by_passes_in_microseconds);
  File "/Users/harshitakalani/repos/lfortran/src/libasr/pass/pass_manager.h", line 193
    throw LCompilersException("Verify failed in the pass: "
LCompilersException: Verify failed in the pass: intrinsic_subroutine
```

PR #10810, opened 2026-03-23
fix: use assumed-shape dims for random_seed put/get dummy arguments

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/10805


HarshilShah1804 (approved):
Looks Good.

Issue #10803, opened 2026-03-23
ICE on deferred-length allocatable character with intent(out), bind(C) during deallocation


HarshitaKalani:
```
 % cat a.f90   
! Test: allocated() and deallocate() on character(len=:),allocatable
! intent(out) bind(C) dummy argument
module allocate_63_mod
    implicit none
contains
    subroutine cleanup(errmsg) bind(C)
        character(len=:), intent(out), allocatable :: errmsg
        if (allocated(errmsg)) then
            deallocate(errmsg)
        end if
    end subroutine
end module

program test_allocate_63
    use allocate_63_mod
    implicit none
    character(len=:), allocatable :: msg

    allocate(character(5) :: msg)
    msg = "hello"
    if (.not. allocated(msg)) error stop "msg should be allocated before cleanup"
    call cleanup(msg)
    if (allocated(msg)) error stop "msg should be deallocated after cleanup"
    print *, "passed"
end program

% lfortran a.f90 
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2809
    LCompilers::print_stack_on_segfault();
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 2736
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/harshitakalani/repos/lfortran/src/bin/lfortran.cpp", line 1250
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/harshitakalani/repos/lfortran/src/lfortran/fortran_evaluator.cpp", line 397
    if (res.ok) {
  File "/Users/harshitakalani/repos/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 22121
    std::string buf;
  File "/Users/harshitakalani/repos/lfortran/src/libasr/asr_base_visitor.h", line 363
    void visit_asr(const asr_t &b) { visit_asr_t(b, self()); }
  File "/Users/harshitakalani/repos/lfortran/src/libasr/asr_base_visitor.h", line 337
    case asrType::unit: { v.visit_unit((const unit_t &)x); return; }
  File "/Users/harshitakalani/repos/lfortran/src/libasr/asr_base_visitor.h", line 364
    void visit_unit(const unit_t &b) { visit_unit_t(b, self()); }
[...]

PR #10814, opened 2026-03-23
fix: allocated/deallocate and call site for bind(C) allocatable character

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/10803

### DIFF
--- a/integration_tests/allocate_63.f90
+++ b/integration_tests/allocate_63.f90
@@ -1,0 +1,24 @@
+
+module allocate_63_mod
+    implicit none
+contains
+    subroutine cleanup(errmsg) bind(C)
+        character(len=:), intent(out), allocatable :: errmsg
+        if (allocated(errmsg)) then
+            deallocate(errmsg)
+        end if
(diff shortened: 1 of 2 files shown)


integration_tests/allocate_63.f90
```diff
@@ -0,0 +1,24 @@
...
+
+program allocate_63
+    use allocate_63_mod
+    implicit none
+    character(len=:), allocatable :: msg
+
+    allocate(character(5) :: msg)
+    msg = "hello"
+    if (.not. allocated(msg)) error stop "msg should be allocated before cleanup"
+    call cleanup(msg)
```
certik, line 21: In a subsequent PR I would also add a test when this function `cleanup` is called from C: you create the allocatable character in C, allocate there, call this function and check that it's deallocated.

certik: In fact you *have* to do this test I described, otherwise we are not sure that we have actually generated the correct C interface.

PR #10817, opened 2026-03-23
feat: implement `sign` specifier for I/O statements 

HarshilShah1804:
Towards #10677 

As suggested in #10798, I will add the documentation for the FileOpen and FileInquire ASR Nodes in a subsequent PR.

### DIFF
--- a/tests/reference/asr-rewind_inquire_flush-72b7f97.json
+++ b/tests/reference/asr-rewind_inquire_flush-72b7f97.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "asr-rewind_inquire_flush-72b7f97.stdout",
-    "stdout_hash": "0c20da1a188b0c6a07f0ad6b355bc18bc68886c5e7b52be6f508c0a1",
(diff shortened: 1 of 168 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2713,10 +2713,10 @@ void free_serialization_info(Serialization_Info* s_info){
 FILE* get_file_pointer_from_unit(int32_t unit_num, bool *unit_file_bin, int *access_id, bool *read_access, bool *write_access, int *delim, bool *blank_zero, int32_t *recl, int *sign_mode, int *decimal_mode);
 
 LFORTRAN_API char* _lcompilers_string_format_fortran(lfortran_allocator_t* al, const char* format, int64_t format_len, const char* serialization_string,
-    int64_t *result_size, int32_t array_sizes_cnt, int32_t string_lengths_cnt, int decimal_mode, ...)
+    int64_t *result_size, int32_t array_sizes_cnt, int32_t string_lengths_cnt, int decimal_mode, int sign_mode, ...)
```
certik, line 2716: Why is the `decimal_mode` variable here? It is not used, right?

certik: Ah I see, it's used deep inside the function.

PR #10820, opened 2026-03-23
Update reference results

certik:


certik:
This should unblock main.

PR #10821, opened 2026-03-23
Make self argument explicit in method call args

certik:
Prepend self as `args[0]` in `make_FunctionCall_t_util` and `make_SubroutineCall_t_util` for method calls (`dt != null`, `pass_attr == Pass`). Remove offset patterns from passes and LLVM backend that compensated for implicit self. Add verifier check that method calls have self as `args[0]`. Update Fortran printer to skip self when printing method calls.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3670,6 +3670,8 @@ RUN(NAME elemental_18 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME elemental_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME elemental_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME elemental_21 LABELS gfortran llvm)
+RUN(NAME elemental_22 LABELS gfortran llvm)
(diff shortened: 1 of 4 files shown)


tests/reference/llvm-call_subroutine_without_type_01-1c100d1.stdout
```diff
@@ -21,6 +21,10 @@ source_filename = "LFortran"
 @5 = private unnamed_addr constant [63 x i8] c"tests/../integration_tests/call_subroutine_without_type_01.f90\00", align 1
 @6 = private unnamed_addr constant [22 x i8] c"'%s' unallocated here\00", align 1
 @7 = private unnamed_addr constant [52 x i8] c"Tried to access member of unallocated variable '%s'\00", align 1
+@8 = private unnamed_addr constant [15 x i8] c"1_mytype_get_i\00", align 1
+@9 = private unnamed_addr constant [63 x i8] c"tests/../integration_tests/call_subroutine_without_type_01.f90\00", align 1
+@10 = private unnamed_addr constant [20 x i8] c"This is unallocated\00", align 1
+@11 = private unnamed_addr constant [45 x i8] c"Argument %d of subroutine %s is unallocated.\00", align 1
```
certik, line 27: The change in this file adds automatic checking for the "self" argument due to it being treated as a regular argument. Which is both good.

PR #10823, opened 2026-03-24
fix: default lf implicit argument casting, avoid crash in optional-arg call lowering

HarshitaKalani:


integration_tests/integer_kind_offset.f90
```diff
@@ -0,0 +1,20 @@
...
+end module module_integer_kind_offset
+
+program integer_kind_offset
+  use module_integer_kind_offset
+#if defined(__GFORTRAN__)
+  call sub(0_c_size_t)
+#else
+  call sub(0)
+#endif
+end program integer_kind_offset
```
certik, line 20: This code is not valid: https://grok.com/share/bGVnYWN5_09fca5c2-99fb-4010-a8a9-aca1fe2306f3. So should not work by default.

integration_tests/CMakeLists.txt
```diff
@@ -3437,6 +3437,10 @@ RUN(NAME implicit_argument_casting_03 LABELS gfortran llvm
     EXTRA_ARGS --implicit-interface --implicit-argument-casting
     GFORTRAN_ARGS -fallow-argument-mismatch)
 
+RUN(NAME integer_kind_offset LABELS gfortran llvm
+    EXTRA_ARGS --cpp
```
certik, line 3441: This code is not conforming, so you must add:
```suggestion
    EXTRA_ARGS --cpp --implicit-argument-casting
```

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -345,7 +345,10 @@ namespace LCompilers::CommandLineInterface {
         }
 
         if (opts.arg_standard == "" || opts.arg_standard == "lf") {
-            // The default LFortran behavior, do nothing
+            compiler_options.implicit_argument_casting = true;
+            if (disable_implicit_argument_casting) {
+                compiler_options.implicit_argument_casting = false;
+            }
```
certik, line 351: We do not want to make this the default.

certik:
I don't think we want to make the implicit argument casting the default. It's not standard and it's bad practice anyway, so I don't think there is any advantage.

Issue #10824, opened 2026-03-24
Semantic error: Procedure declarations without an explicit interface (procedure()) are not yet supported.


Jatinagarwal24:
MRE
```
module m
    implicit none

    type :: my_type
        procedure(), pointer, nopass :: caller => null()
        procedure(), pointer, nopass :: f => null()
    end type

contains

    subroutine dummy_caller(func, arg)
        procedure() :: func
        integer, intent(in) :: arg
        ! Simply returning here still tests the compiler's ability
        ! to parse "call this%caller(this%f, arg1)" inside invoke()
        
    end subroutine dummy_caller

    subroutine dummy_f(arg)
        integer, intent(in) :: arg
        print *, "Value is: ", arg
    end subroutine dummy_f

    subroutine invoke(this, arg1)
        class(my_type), intent(in) :: this
        integer, intent(in) :: arg1
        call this%caller(this%f, arg1)
    end subroutine invoke

end module

program expr2
    use m
    implicit none

    type(my_type) :: t

    t%caller => dummy_caller
    t%f => dummy_f

    call invoke(t, 42)
end program expr2
```
Lfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran examples/expr2.f90 --implicit-interface
semantic error: Procedure declarations without an explicit interface (procedure()) are not yet supported. Please use procedure(interface_name) or declare an interface block.
 --> examples/expr2.f90:5:9
  |
5 |         procedure(), pointer, nopass :: caller => null()
  |         ^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# 
```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran examples/expr2.f90 && ./a.out 
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# 
```

Jatinagarwal24:
I am on it.

certik:
Fixed in https://github.com/lfortran/lfortran/pull/10850.

PR #10827, opened 2026-03-24
fix: support procedure pointer declarations without explicit interface

Jatinagarwal24:
feat: Support procedure() with implicit interface in derived type components

Previously, the compiler rejected procedure pointer declarations with empty
parentheses (procedure()) in derived type components, reporting:
  "Procedure declarations without an explicit interface (procedure()) are 
   not yet supported. Please use procedure(interface_name) or declare an 
   interface block."

This is valid Fortran syntax for procedure pointers with implicit interfaces,
commonly used for callback patterns and generic programming where the specific
interface isn't known at type definition time.

The fix implements comprehensive support for procedure() across multiple phases:

## 1. AST Body Visitor (ast_body_visitor.cpp, lines 2582-2591)

Added handling for implicit interface procedure pointers in assignments:
- Line 2584: Check if target is an implicit interface (target_func_type->n_arg_types == 0)
- Lines 2585-2587: If implicit, perform additional type compatibility checks
  comparing target_type_underlying with value_type_underlying
- Lines 2590-2591: Also check with direct target_type and value_type
- This ensures type safety when assigning to procedure() pointers

## 2. AST Common Visitor (ast_common_visitor.h, lines 6844-6993)

Creates interface declarations for procedure() components:
- Lines 6844-6850: Detect procedure variables with nullptr type_declaration
- Lines 6964-6993: Handle final_sym (the callee in procedure calls)
  * Lines 6965-6968: Check if callee_var is a procedure variable with null type
  * Lines 6969-6973: Extract callee type information
  * Lines 6974-6987: Infer argument types from call sites:
    - Reserve space for inferred argument types
    - Push each argument's type into inferred_arg_types vector
  * Lines 6981-6986: Create inferred FunctionType from call context
  * Line 6987-6988: Call create_interface_for_procedure_variable

Lines 6989-6992: Fallback for non-inferred cases:
  - Create interface with callee function type directly

This inference mechanism allows procedure() pointers to have their interface
determined from usage context rather than requiring explicit declaration.

## 3. Symbol Table Visitor (ast_symboltable_visitor.cpp, lines 1313, 1785)

Track procedure declarations without explicit interfaces:
- Line 1313: Added condition to check for procedure() with empty name
  `(type->m_name != nullptr && std::string(type->m_name != nullptr ? type->m_name : "") == sym_name)`
- Line 1785: Similar check for matching procedure declarations
  `(type->m_name != nullptr && std::string(type->m_name != nullptr ? type->m_name : "") == sym_name)`

These checks allow procedure_decl_indices to track implicit interface procedures
so they can be properly resolved during semantic analysis.

## 4. ASR Utils (asr_utils.cpp, lines 720-724, 744-748, 730, 754)

Handle procedure() in function extraction and type resolution:
- Lines 720-724: Return nullptr for variables with null m_type_declaration
  (implicit interface procedure variables without concrete function attached)
- Lines 744-748: Similar handling in another code path
- Line 730: Remove exception for undeclared Var_t symbols
- Line 754: Remove exception for non-Function_t symbols

This allows get_function_from_expr to gracefully handle procedure() pointers
that don't have explicit type declarations yet.

## 5. LLVM Code Generation (asr_to_llvm.cpp, lines 9087-9098, 18196, 20499-20535, 20748-20800)

Generate proper LLVM IR for procedure() pointer calls:

Lines 9087-9098 (LLVM 15+): Handle procedure() pointer member stores
- Line 9090: Check if both source and target are function pointers
- Lines 9092-9094: Verify destination type matches function pointer type
- Line 9095: Cast concrete procedure pointer to `void ()*` before storing

Lines 20499-20535, 20748-20800: Handle procedure() pointer invocations
- Lines 20503-20516: Detect procedure pointer with pass argument
- Lines 20518-20620: Build argument type list by iterating call args
- Lines 20518-20526: Handle interface function with pass (self) argument
- Lines 20623-20634: Create new FunctionType from inferred arg types
- Line 20624: BitCast callee to new function type pointer
- Line 20625: Update fntype for CreateCall

Lines 20748-20800: Handle non-method procedure() pointer calls
- Similar logic for inferring argument types from call site
- Create new FunctionType matching actual call signature
- BitCast procedure pointer and invoke

Lines 18196: Check arg count before accessing function parameters
- `LCOMPILERS_ASSERT(arg_idx < func_subrout->n_args)`
- Prevents crashes when calling procedure() with implicit interfaces

## 6. LLVM Utils (llvm_utils.cpp, lines 1640, 1642-1643)

Handle void function type for procedure() pointers:
- Line 1640: Check if arg_expr can be extracted as function
- Lines 1642-1643: Use void function type when no explicit type available
  `llvm_type = llvm::FunctionType::get(llvm::Type::getVoidTy(context), false)->getPointerTo()`

## 7. Subroutine from Function (subroutine_from_function.cpp, lines 95-99)

Extract procedure variable information correctly:
- Line 96: Get symbol from m_type_declaration for procedure variables
- Line 97: Check if symbol is actually a FunctionType (not just Variable)
- Line 98-99: Extract and use function signature from symbol

This fixes the case where procedure() variables need their type information
retrieved from the resolved interface rather than the variable declaration.

The implementation allows valid Fortran patterns like:
```fortran
type :: my_type
    procedure(), pointer, nopass :: caller => null()
    procedure(), pointer, nopass :: f => null()
end type

t%caller => dummy_caller
t%f => dummy_f
call this%caller(this%f, arg1)  ! Infers types from call site
```

Resolves: #10824

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -9054,16 +9054,8 @@ public:
                             llvm::Type* llvm_data_type = llvm_utils->get_el_type(x.m_target, data_type, module.get());
                             fill_array_details(llvm_target_type, llvm_target_, llvm_data_type, m_dims, n_dims, false, false);
                             llvm::Value* target_data_ptr = arr_descr->get_pointer_to_data(llvm_target_type, llvm_target_);
-#if LLVM_VERSION_MAJOR < 15
-                            // Typed-pointer LLVM (<15) requires the stored pointer type to match
-                            // the destination field type exactly.
-                            if (llvm_value->getType() != target_data_ptr->getType()->getPointerElementType()) {
-                                LCOMPILERS_ASSERT(llvm_value->getType()->isPointerTy());
-                                LCOMPILERS_ASSERT(target_data_ptr->getType()->getPointerElementType()->isPointerTy());
-                                llvm_value = builder->CreateBitCast(
-                                    llvm_value, target_data_ptr->getType()->getPointerElementType());
-                            }
-#endif
+                            llvm_value = LLVM::cast_pointer_to_store_element_type_if_needed(
+                                *builder, llvm_value, target_data_ptr);
                             builder->CreateStore(llvm_value, target_data_ptr);
                             llvm_value = llvm_target_;
                             break;
@@ -9091,18 +9083,8 @@ public:
                         llvm_target_contents_type->getPointerTo(), llvm_target);
                     builder->CreateStore(llvm_value, loaded_target);
                 } else {
-#if LLVM_VERSION_MAJOR < 15
-                    // procedure() pointer members are stored as `void ()*` placeholders.
-                    // Cast concrete procedure pointers before storing into that slot.
-                    if (llvm_value->getType()->isPointerTy() && llvm_target->getType()->isPointerTy()) {
-                        llvm::Type* dst_type = llvm_target->getType()->getPointerElementType();
-                        if (dst_type->isPointerTy() && llvm_value->getType() != dst_type &&
-                                llvm_value->getType()->getPointerElementType()->isFunctionTy() &&
-                                dst_type->getPointerElementType()->isFunctionTy()) {
-                            llvm_value = builder->CreateBitCast(llvm_value, dst_type);
-                        }
-                    }
-#endif
+                    llvm_value = LLVM::cast_function_pointer_to_store_element_type_if_needed(
+                        *builder, llvm_value, llvm_target);
                     builder->CreateStore(llvm_value, llvm_target);
                 }
             }
--- a/src/libasr/codegen/llvm_utils.cpp
+++ b/src/libasr/codegen/llvm_utils.cpp
@@ -16,6 +16,40 @@ namespace LCompilers {
             return builder.CreateStore(x, y);
         }
 
+        llvm::Value* cast_pointer_to_store_element_type_if_needed(
+                llvm::IRBuilder<> &builder, llvm::Value *value, llvm::Value *target_ptr) {
+#if LLVM_VERSION_MAJOR < 15
+            if (value->getType()->isPointerTy() && target_ptr->getType()->isPointerTy()) {
+                llvm::Type* dst_type = target_ptr->getType()->getPointerElementType();
+                if (dst_type->isPointerTy() && value->getType() != dst_type) {
+                    value = builder.CreateBitCast(value, dst_type);
+                }
+            }
+#else
+            (void)builder;
+            (void)target_ptr;
+#endif
+            return value;
+        }
+
+        llvm::Value* cast_function_pointer_to_store_element_type_if_needed(
+                llvm::IRBuilder<> &builder, llvm::Value *value, llvm::Value *target_ptr) {
+#if LLVM_VERSION_MAJOR < 15
+            if (value->getType()->isPointerTy() && target_ptr->getType()->isPointerTy()) {
+                llvm::Type* dst_type = target_ptr->getType()->getPointerElementType();
+                if (dst_type->isPointerTy() && value->getType() != dst_type &&
+                        value->getType()->getPointerElementType()->isFunctionTy() &&
+                        dst_type->getPointerElementType()->isFunctionTy()) {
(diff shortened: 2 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -18254,7 +18266,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                                         std::string& orig_arg_name, ASR::intentType& arg_intent,
                                         size_t arg_idx) {
         m_h = get_hash((ASR::asr_t*)func_subrout);
-        LCOMPILERS_ASSERT(arg_idx < func_subrout->n_args);
+        if (arg_idx >= func_subrout->n_args) { x_abi = ASRUtils::get_FunctionType(func_subrout)->m_abi; return; }
```
certik, line 18251: What is this doing? Why would `arg_idx >= func_subrout->n_args` be allowed?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9091,6 +9091,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    // Cast concrete procedure pointers before storing into that slot.
+                    if (llvm_value->getType()->isPointerTy() && llvm_target->getType()->isPointerTy()) {
+                        llvm::Type* dst_type = llvm_target->getType()->getPointerElementType();
+                        if (dst_type->isPointerTy() && llvm_value->getType() != dst_type &&
+                                llvm_value->getType()->getPointerElementType()->isFunctionTy() &&
+                                dst_type->getPointerElementType()->isFunctionTy()) {
+                            llvm_value = builder->CreateBitCast(llvm_value, dst_type);
+                        }
+                    }
+#endif
```
certik: This must be refactored into llvm_utils. asr_to_llvm should not contain any ifdefs like this.

integration_tests/procedure_49.f90
```diff
@@ -0,0 +1,44 @@
+module procedure_49_mod
+    implicit none
+
+    integer :: seen = -1
+
+    type :: my_type
+        procedure(), pointer, nopass :: caller => null()
+        procedure(), pointer, nopass :: f => null()
```
certik, line 8: I didn't know that this is allowed.

integration_tests/procedure_49.f90
```diff
@@ -0,0 +1,44 @@
...
+
+    subroutine dummy_f(arg)
+        integer, intent(in) :: arg
+        if (arg /= 42) error stop
+    end subroutine dummy_f
+
+    subroutine invoke(this, arg1)
+        class(my_type), intent(in) :: this
+        integer, intent(in) :: arg1
+        call this%caller(this%f, arg1)
```
certik, line 27: How can the compiler check this?

If this cannot be checked, then I think we need to hide this usage behind `--implicit-interface`, not allow by default.

Jatinagarwal24: Done

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: Done

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: convert_call_args() iterates over actual call arguments (x.n_args) in [asr_to_llvm.cpp:18281]
For some paths (especially incomplete/implicit procedure interfaces like procedure() and certain indirect calls), the callee’s formal signature can be missing or shorter than the actual call site.
In that case, indexing [func_subrout->m_args[arg_idx]] would be out-of-bounds and crash.

certik: I think I fixed this exact issue in https://github.com/lfortran/lfortran/pull/10838. The call site and the callee should now exactly agree on the number of arguments. If not, it's a bug that we should fix.

certik:
Requires a rebase. Is this still needed?

Jatinagarwal24:
> Requires a rebase. Is this still needed?

I think this is not needed.As the issue is solved now we are getting different error for which I will open new pr.

PR #10831, opened 2026-03-24
fix(llvm): emit [0 x t] instead of <0 x t> for empty SIMD array shapes

HarshitaKalani:


integration_tests/assume_rank_shape_product_01.f90
```diff
@@ -0,0 +1,31 @@
...
+  use iso_c_binding, only: c_int, c_size_t, c_null_ptr, c_ptr
+  implicit none
+  integer :: x(2, 3)
+  x = 1
+  call sub_with_product_shape(x)
+contains
+  subroutine sub_with_product_shape(a)
+    type(*), intent(inout), target, contiguous :: a(..)
+    interface
+      subroutine caf_dummy(a, sz, team) bind(c)
```
certik, line 12: We should test this from C as well (separate test, subsequent PR).

PR #10837, opened 2026-03-24
fix: handle pass(arg) where arg is not the first parameter

certik:
When a type-bound procedure uses pass(arg) where the named argument is not the first parameter (e.g., procedure, pass(pt) :: add_named with add_named(scale, pt)), LFortran was always inserting the passed object at position 0 instead of at the correct parameter index.

Fix the argument insertion in three layers:
- asr_utils.h: add get_pass_arg_index() helper; fix make_FunctionCall_t_util and make_SubroutineCall_t_util to insert self at the correct position
- ast_common_visitor.h: fix create_StructMethodDeclaration to build intermediate arg lists with self at the correct position
- asr_to_llvm.cpp: fix convert_call_args to skip the correct arg index; fix visit_FunctionCall and visit_SubroutineCall to use the correct formal parameter for polymorphic conversion and insert the pass arg at the correct position

Test added.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -889,6 +889,28 @@ static inline bool get_class_proc_nopass_val(ASR::symbol_t* func_sym) {
     return nopass;
 }
 
+// Returns the index of the pass argument in the function's formal parameter
+// list. Returns 0 when pass_arg is unspecified (default pass = first arg).
+static inline size_t get_pass_arg_index(ASR::symbol_t* a_name) {
+    ASR::symbol_t* sym = symbol_get_past_external(a_name);
+    if (!ASR::is_a<ASR::StructMethodDeclaration_t>(*sym)) {
+        return 0;
+    }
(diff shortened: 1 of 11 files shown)


src/libasr/asr_utils.h
```diff
@@ -889,6 +889,28 @@ static inline bool get_class_proc_nopass_val(ASR::symbol_t* func_sym) {
...
+    if (clss_proc->m_self_argument == nullptr) {
+        return 0;
+    }
+    ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(clss_proc->m_proc);
+    for (size_t i = 0; i < func->n_args; i++) {
+        ASR::Variable_t* v = EXPR2VAR(func->m_args[i]);
+        if (strcmp(v->m_name, clss_proc->m_self_argument) == 0) {
+            return i;
+        }
+    }
```
certik, line 910: It would be good to speed this up by having this more explicit in ASR.

PR #10844, opened 2026-03-25
Support standalone pointer attribute declaration

certik:
Add support for standalone 'pointer :: var' attribute declarations in Fortran's implicit interface mode. This enables procedure pointer patterns like:

    pointer :: ifp
    external :: ifp, isquare
    ifp => isquare
    print *, ifp(7)

Changes:
- Handle AttrPointer in standalone attribute declarations with deferred tracking via assgnd_pointer set
- Create procedure pointer Variables (with Pointer(FunctionType) type) when both pointer and external attributes are present
- Create backing interface Function_t for procedure pointers
- Refactor needs_interface to update existing Function_t in place when m_type_declaration exists, avoiding overwriting Implementation functions
- Relax return type check in visit_Associate for implicit interface pointer assignment
- Error with clear message when pointer :: used without --implicit-interface

Integration test proc_ptr_17.f90 added.


certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/10846.

PR #10855, opened 2026-03-25
Fix reshape of derived type array in parameter declarations

certik:
LFortran rejected code like:
  type(pair), parameter :: v(*,*) = reshape([pair(1,2.), ...], [2,2])
with 'Initialization of v must reduce to a compile time constant'.

Root causes and fixes:

1. StructConstructor with all-constant args had m_value=nullptr instead of m_value=StructConstant. The else branch in create_DerivedTypeConstructor now computes all_args_const and sets value=StructConstant when all args are constant, while preserving the is_const flag for the direct StructConstant return path.

2. compare_helper threw a fatal error for types it cannot evaluate at compile time (e.g. StructType). Changed to return nullptr so that overloaded operator resolution proceeds normally.

3. Implied-shape dimension inference for TypeType parameter declarations was missing. Added parameter type adjustment for physical type and dimension inference from ArrayConstant/ ArrayConstructor values.

4. fetch_ArrayConstant_value, fetch_ArrayConstant_value_helper, and get_struct_sym_from_struct_expr lacked StructType/ArrayConstant handling, causing crashes in passes that process struct array constants. Added the missing cases.

5. pickle.cpp computed wrong element count for struct ArrayConstants (used n_data/kind instead of n_data/sizeof(expr_t*)). Fixed.

6. types_equal crashed on down_cast when get_struct_sym_from_struct_expr returned nullptr for either expression. Added null check with fallback to structural type comparison.

7. implied_do_loops pass did not pass the original result_var when creating temporary variables for array constant expansion, so struct temporaries lacked m_type_declaration. Passed result_var_copy.

Integration test added: arrays_reshape_40.f90


certik:
Looks like CI fails.

PR #10856, opened 2026-03-25
Add do_concurrent_to_openmp pass for CUDA target offloading

certik:
Add a new ASR pass that transforms DoConcurrentLoop into OMPRegion(Target > Teams > DistributeParallelDo > DoLoop) with auto-generated OMPMap(ToFrom) clauses for array variables.

This reuses the existing C backend CUDA codegen that already handles OpenMP target regions (used by openmp_70/71/72), enabling do concurrent loops to be offloaded to NVIDIA GPUs via the same --target-offload pipeline.

The pass runs before the openmp pass in the pipeline and only activates when --target-offload is set. Register do_concurrent_14 as a target_offload integration test.


certik:
The next step should be to move the OpenMP offloading from the C backend to a proper ASR pass https://github.com/lfortran/lfortran/issues/10857.

PR #10865, opened 2026-03-26
Fix several bugs for Caffeine

certik:


HarshitaKalani:
```
1284 - bindc_iso_fb_06 (Failed)
```

certik:
Turned out the failure was in gfortran's test, due to a new test that was not portable. Fixed in 1e36cea6e65cbe70afd9c27b32fedbb1fb39c969.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8442,14 +8446,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                     ASRUtils::type_get_past_pointer(value_array_type)),
                     module.get());
+#if LLVM_VERSION_MAJOR < 15
+                // On typed pointers, only GEP if value_desc points to the
+                // array type. For parameters whose m_value was visited via
+                // visit_ArrayConstant, the pointer is already at element level.
+                if (value_desc->getType()->getPointerElementType() == val_type) {
+                    value_desc = llvm_utils->create_gep2(val_type, value_desc, 0);
+                }
+#else
```
certik, line 8462: I don't like this, but probably ok for now to unblock us, but we'll have to figure out a better solution later.

PR #10869, opened 2026-03-27
fix: handle slash descriptor after repeated format groups to preserve newline in formatted output

### LINKED ISSUE #10866: Formatted WRITE: slash (newline) edit descriptor ignored after parenthesized group
```
program write_fmt3
  implicit none

  write (*,100) 1.0, 2.0, 3.0, 4.0

100 format ('values are: ', 4(f4.1)/'next line')

end program
```
```
$ lfortran write_fmt3.f90
values are:  1.0 2.0 3.0 4.0next line
$
```
```
$ gfortran write_fmt3.f90
$ ./a.out
values are:  1.0 2.0 3.0 4.0
next line
$
```

Jatinagarwal24:
fixes #10866


HarshitaKalani (approved):

certik:
Must fix third-party codes.

Jatinagarwal24:
> Must fix third-party codes.

Hi ,@certik I fix it pls check

certik:
I am worried about a7e0a98dbbf57303392510d88137c851afd90833, as we need to be calling finalizers. Is that introducing a regression?

Jatinagarwal24:
> I am worried about [a7e0a98](https://github.com/lfortran/lfortran/commit/a7e0a98dbbf57303392510d88137c851afd90833), as we need to be calling finalizers. Is that introducing a regression?

Hi @certik , This change is not intended to skip finalization semantics. It only prevents emitting invalid direct FINAL calls when the value is not a scalar struct-like object (which was producing LLVM verifier errors due to signature mismatch). Finalization still runs through the normal finalize pipeline, including element-wise finalization for arrays. So this is a type-correctness fix in dispatch, not a removal of FINAL behavior. To be safe, I can add/verify regression coverage for both scalar and array cases so we confirm FINAL is still invoked where required.

certik:
The first thing to do is extract an MRE from the third-party test, and commit it as an integration test.

Second, if you could double check that it is doing the right thing, that would be great.

Third, what caused this failure? The commit https://github.com/lfortran/lfortran/pull/10869/changes/8c909b67bca0b6f0e8b55b96a288635da1547cc8 is a tiny modification to our runtime, how did it cause such a failure in all third-party tests?

PR #10880, opened 2026-03-28
memory-leak : Handle pointer types finalization instead of bypassing them 

assem2002:
- Minor refactors in the finalization class.
- Handle pointer types finalization instead of bypassing them entirely .
- Dump the runtime branching (based on `in_struct` boolean) and did the job at compile time.


assem2002:
Yeah they're not freed automatically, but yet our compiler might create descriptors (array descriptor in this PR) that we are the responsible ones of freeing it-- in other words we free the descriptors but never free the data.

certik:
Got it, it is some kind of a heap descriptor or something.

PR #10881, opened 2026-03-28
CI : Run `--detect_leaks` at the CI

assem2002:
262 tests are disabled when `--detect-leaks` is tested by using tag `NO_DETECT_LEAK`


ci/test.sh
```diff
@@ -82,6 +82,7 @@ if [[ $WIN != "1" ]]; then
     if [[ $MACOS != "1" ]]; then
         ./run_tests.py -b llvm_submodule -sc -j${NPROC}
     fi
+    ./run_tests.py -b llvm --detect-leaks
```
certik, line 85: This will add to our tests time. Do we need to run it *in addition* to the regular test? Probably at least at the beginning we do, since it's using a different path in the compiler, so we probably have to test both.

integration_tests/run_tests.py
```diff
@@ -189,6 +193,8 @@ def get_args():
                 help="Run tests with the requested Fortran standard: ".join(SUPPORTED_STANDARDS))
     parser.add_argument("-f", "--fast", action='store_true',
                 help="Run supported tests with --fast")
+    parser.add_argument("--detect-leak", "--detect-leaks", dest="detect_leak", action='store_true',
```
certik, line 196:
```suggestion
    parser.add_argument("--detect-leaks", dest="detect_leaks", action='store_true',
```

integration_tests/run_tests.py
```diff
@@ -77,16 +78,19 @@ def run_test(backend, std, test_pattern=None):
...
         run_cmd(f"FC=lfortran cmake -DLFORTRAN_BACKEND={backend} "
-            f"-DFAST={fast_tests} -DLLVM_GOC={separate_compilation} -DNOFAST_LLVM16={nofast_llvm16} "
+            f"-DFAST={fast_tests} -DDETECT_LEAK={detect_leak_tests} "
+            f"-DLLVM_GOC={separate_compilation} -DNOFAST_LLVM16={nofast_llvm16} "
             f"-DCMAKE_Fortran_FLAGS=\"-fPIC\" {std_string}" + common,
                 cwd=cwd)
     else:
         run_cmd(f"FC=lfortran cmake -DLFORTRAN_BACKEND={backend} -DFAST={fast_tests} "
-                f"-DLLVM_GOC={separate_compilation} {std_string} -DNOFAST_LLVM16={nofast_llvm16} " + common,
+                f"-DDETECT_LEAK={detect_leak_tests} -DLLVM_GOC={separate_compilation} "
```
certik, line 92:
```suggestion
                f"-DDETECT_LEAKS={detect_leaks_tests} -DLLVM_GOC={separate_compilation} "
```

certik:
The comments can be fixed later. I would use the term "detect leaks", so I would name the options the same way.

PR #10882, opened 2026-03-28
memory-leaks : finalize globals when `detect-leaks` in ON

assem2002:


src/libasr/codegen/llvm_utils.h
```diff
@@ -976,7 +976,6 @@ class ASRToLLVMVisitor;
      * @class LLVMFinalize
      * @brief Finalize variables before exiting their scope.
      * @details 
-     * Notice Global variables aren't finalized; They live till program ends.
```
certik, line 979: This now depends on `--detect-leaks`, correct? So I would update the comment.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1730,9 +1729,8 @@ class ASRToLLVMVisitor;
                                       && ASRUtils::get_FunctionType(sym)->m_deftype == ASR::Interface;
             const bool is_external_abi = sym && ASR::is_a<ASR::Function_t>(*sym)
                                       && ASRUtils::get_FunctionType(sym)->m_abi == ASR::ExternalUndefined;
-            const bool is_module = sym && ASR::is_a<ASR::Module_t>(*sym);
             const bool is_TU = !sym && ASR::is_a<ASR::unit_t>(*s) && ASR::is_a<ASR::TranslationUnit_t>(*(ASR::unit_t*)s);
-            return is_TU || is_module || is_interface || is_external_abi ;
+            return is_TU || is_interface || is_external_abi ;
```
certik, line 1733: What is the implication of this change?

assem2002: modules used to trigger as non finalizable symbol table, Now it doesn't.
I might add some affirmiation step from the class user to make sure he's aware that he's finalizing a module symbolTable; That should prevent unintended usage, as the best way of doing finalization of any fortran code is to not free globals.

PR #10863, opened 2026-03-26
LSP: Show type kind information in hover preview

himanshumahajan138:
Fixes https://github.com/lfortran/lfortran-vscode-client/issues/49

## Summary

- Display kind information in LSP hover preview (e.g., `real(4)`, `integer(8)`, `character(len=10)`) instead of bare type names (`real`, `integer`, `string`)
- Added backward-compatible `show_kind` parameter to `type_to_str_fortran_symbol()` to avoid breaking internal call sites that use the output for function name generation
- Updated LSP hover test expectation to verify kind info is shown

## Before

Hovering over a function showed types without kind:
```fortran
function eval_1d(self, x) result(res)
    class(softmax), intent(in) :: self
    real[:], intent(in) :: x
    real[:] :: res
end function
```

## After
Hovering over a function now shows types with kind:

```fortran
function eval_1d(self, x) result(res)
    class(softmax), intent(in) :: self
    real(4)[:], intent(in) :: x
    real(4)[:] :: res
end function
```

## Test plan

- [x] LSP hover test passes: LFORTRAN_PATH=build/src/bin/lfortran pytest tests/server/tests/test_features.py::test_document_hover -v
- [x] Verified test fails before fix and passes after fix

PR #10891, opened 2026-03-28
LSP: Show dimension and allocatable/pointer attributes in hover

himanshumahajan138:
## Summary
- Extend `symbol_to_str_fortran` to emit `dimension(:)`, `allocatable`, and `pointer` attributes for variable hover previews
- Fix type extraction to use `extract_type` before checking for `StructType_t`, avoiding incorrect type dispatch on wrapper types like `Array_t`
- Update LSP hover test to match the new `dimension(:)` output format

## Context
Extends #10863 (LSP: Show type kind information in hover preview).
Related to lfortran/lfortran-lsp#49 (show declaration line when hovering over a variable).

Previously, hovering over an array variable like `real(4) :: x(:)` would display `real(4)[:]` instead of the standard Fortran `real(4), dimension(:)` syntax. Allocatable and pointer attributes were also missing from the hover output.

## Test plan
- [x] Verify `tests/server/tests/test_features.py::test_document_hover` passes with updated expected output
- [x] Confirm hover preview now shows `real(4), dimension(:), intent(in) :: x` instead of `real(4)[:], intent(in) :: x`
- [x] Run `./run_tests.py -j16 &> log` to ensure no regressions

## Screenshot of Result
Screenshot 2026-03-28 at 9 21 23 PM


himanshumahajan138:
Hi @certik , 

all test passed could u please review it once available

himanshumahajan138:
Hi @certik, any update ?

himanshumahajan138:
@certik 
mentioning as follow up

certik (approved):
That looks good, let's run full tests.

himanshumahajan138:
just synced with main and then ready to merge
Thanks...

certik (approved):
I think this is fine. Thanks!

PR #10892, opened 2026-03-28
enh: handle `order` argument in `reshape` at runtime

jinangshah21:
Needed for stdlib tests

### Algorithm Followed (Simple Flow)
For each output element index `i` from `0` to `target_size-1`:
1) Convert `i` to result subscripts `I[0..n-1]` using result shape in normal Fortran `order`.
2) Read permutation `order[0..n-1]`.
3) Build source linear index:
start `source_index = 0`, `stride = 1`

For each position `j` in permutation order:
```    
    dim = order[j] - 1
    source_index += I[dim] * stride
    stride *= shape[dim]
```
4) Get element from source array with modified offset value


certik:
You should likely split the test, as the new additions do not pass some of the auxiliary tests, not sure if it is gfortran or wasm.

certik (approved):

Issue #3697, opened 2024-03-25
Codegen error when loop iterator is `double` type


Shaikh-Ubaid:
```console
% cat examples/expr3.f90     
program main

    real :: i

    do i = 1.10, 3.5
        print *, i
    end do

end program
% gfortran examples/expr3.f90
examples/expr3.f90:5:7:

    5 |     do i = 1.10, 3.5
      |       1
Warning: Deleted feature: Loop variable at (1) must be integer
examples/expr3.f90:5:10:

    5 |     do i = 1.10, 3.5
      |          1
Warning: Deleted feature: Start expression in DO loop at (1) must be integer
examples/expr3.f90:5:16:

    5 |     do i = 1.10, 3.5
      |                1
Warning: Deleted feature: End expression in DO loop at (1) must be integer
% ./a.out 
   1.10000002    
   2.09999990    
   3.09999990    
% lfortran examples/expr3.f90
; ModuleID = 'LFortran'
source_filename = "LFortran"

@0 = private unnamed_addr constant [2 x i8] c" \00", align 1
@1 = private unnamed_addr constant [2 x i8] c"\0A\00", align 1
@2 = private unnamed_addr constant [9 x i8] c"%13.8e%s\00", align 1

define i32 @main(i32 %0, i8** %1) {
.entry:
  call void @_lpython_set_argv(i32 %0, i8** %1)
  %i = alloca float, align 4
  store float sub (float 0x3FF19999A0000000, i32 1), float* %i, align 4
  br label %loop.head

loop.head:                                        ; preds = %loop.body, %.entry
  %2 = load float, float* %i, align 4
  %3 = add float %2, i32 1
  %4 = icmp sle float %3, 3.500000e+00
  br i1 %4, label %loop.body, label %loop.end

loop.body:                                        ; preds = %loop.head
  %5 = load float, float* %i, align 4
  %6 = add float %5, i32 1
  store float %6, float* %i, align 4
  %7 = load float, float* %i, align 4
  %8 = fpext float %7 to double
  call void (i8*, ...) @_lfortran_printf(i8* getelementptr inbounds ([9 x i8], [9 x i8]* @2, i32 0, i32 0), double %8, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @1, i32 0, i32 0))
  br label %loop.head

loop.end:                                         ; preds = %loop.head
  ret i32 0
}

declare void @_lpython_set_argv(i32, i8**)

[...]

PR #10893, opened 2026-03-28
fix: Support real-typed loop variables in DoLoop to WhileLoop lowering pass

Jatinagarwal24:
Previously, the compiler crashed with an LLVM ICmpInst assertion failure when
compiling do-loops with real loop variables (e.g., `do i = 1.0, 3.5` where `i`
is `real`). The semantic analysis correctly allowed this pattern (with warnings
matching gfortran behavior via cast_as_loop_var), but the DoLoop→WhileLoop
lowering pass in pass_utils.cpp assumed integer types everywhere, causing type
mismatches during LLVM code generation when float values were passed to integer
comparison instructions.

Root cause: The replace_doloop() and insert_stmts_in_loop_body() functions
hardcoded:
- IntegerConstant_t for default increment of 1
- IntegerBinOp_t for all arithmetic (Add, Sub)  
- IntegerCompare_t for loop condition
- Integer_t for expression types
- is_integer() check that threw "Do loop increment type should be an integer"

The fix adds real type support throughout the DoLoop lowering pass:

## 1. replace_doloop() function (lines 1356-1516)

Added type checking to branch between integer and real ASR nodes:

- Check loop variable type: `ASR::is_a<ASR::Real_t>(*var_type)`

- Default increment (lines 1388-1390):
  * Integer: IntegerConstant_t(1)
  * Real: RealConstant_t(1.0) with Real_t type

- Non-constant increment saving (lines 1398-1416):
  * Extended IntegerConstant_t/IntegerUnaryMinus_t checks
  * Also handle RealConstant_t/RealUnaryMinus_t for real loops

- Increment direction check (lines 1420-1490):
  * Integer: Check IntegerConstant_t value > 0 / < 0
  * Real: Check RealConstant_t value > 0.0 / < 0.0
  * Use RealCompare_t instead of IntegerCompare_t for real types
  * Use RealBinOp_t instead of IntegerBinOp_t for real arithmetic

- Type creation (line 1494):
  * Integer loops: Integer_t type
  * Real loops: Real_t type

- Loop initialization, increment, and condition (lines 1496-1514):
  * Integer: IntegerBinOp_t for arithmetic, IntegerCompare_t for conditions
  * Real: RealBinOp_t for arithmetic, RealCompare_t for conditions

## 2. insert_stmts_in_loop_body() function (lines 1313-1354)

Fixed type handling for loop increment statements:

- Line 1324: Use loop variable's actual type (var_type) instead of always Integer_t
  
- Lines 1332-1334: Branch on loop variable type:
  * Integer: Use IntegerBinOp_t for increment arithmetic
  * Real: Use RealBinOp_t for increment arithmetic

Implementation approach:
- Introduced is_real_loop boolean flag to branch between Integer and Real ASR
  node constructors
- Keeps logic self-contained within the lowering pass
- Avoids larger architectural refactors

This ensures that real-typed do-loops like:
```fortran
real :: i
do i = 1.0, 10.0, 0.5
    print *, i
end do
```
are correctly lowered to while-loops with RealBinOp_t/RealCompare_t nodes
instead of IntegerBinOp_t/IntegerCompare_t, preventing LLVM type assertion
failures during code generation.

Resolves: #3697

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -567,6 +567,7 @@ RUN(NAME doloop_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --us
 RUN(NAME doloop_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --use-loop-variable-after-loop)
 RUN(NAME doloop_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME doloop_18 LABELS gfortran llvm)
+RUN(NAME doloop_19 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME cycle_and_exit1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME cycle_and_exit2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
--- a/integration_tests/doloop_19.f90
+++ b/integration_tests/doloop_19.f90
@@ -1,0 +1,29 @@
+! Test do loop with real loop variable and real bounds
+program doloop_19
+    implicit none
+    real :: x
+    integer :: count
+
+    ! Basic real do loop (default increment of 1.0)
+    count = 0
+    do x = 1.10, 3.5
+        count = count + 1
+    end do
+    if (count /= 3) error stop
+
+    ! Real do loop with explicit real increment
+    count = 0
+    do x = 0.0, 2.0, 0.5
+        count = count + 1
+    end do
+    if (count /= 5) error stop
+
+    ! Real do loop counting down (negative increment)
+    count = 0
+    do x = 3.0, 1.0, -1.0
+        count = count + 1
+    end do
+    if (count /= 3) error stop
+
+    print *, "ok"
+end program


certik:
Does this print a warning?

YashNagda17:
I feel this isn't allowed by default, as different systems might produce numerical inconsistencies for the same code. 

Below is an excerpt from Fortran 2018, section 11.1.7, where it only says these variable types as integer scalars

image

Jatinagarwal24:
> Does this print a warning?

examples/expr2.f90:5:7:

    5 |     do i = 1.10, 3.5
      |       1
Warning: Deleted feature: Loop variable at (1) must be integer
examples/expr2.f90:5:10:

    5 |     do i = 1.10, 3.5
      |          1
Warning: Deleted feature: Start expression in DO loop at (1) must be integer
examples/expr2.f90:5:16:

    5 |     do i = 1.10, 3.5
      |                1
Warning: Deleted feature: End expression in DO loop at (1) must be integer
   1.10000002    
   2.09999990
   3.09999990

warning: Start expression in DO loop must be integer
 --> examples/expr2.f90:5:12
  |
5 |     do i = 1.10, 3.5
  |            ^^^^ 

warning: End expression in DO loop must be integer
 --> examples/expr2.f90:5:18
  |
5 |     do i = 1.10, 3.5
  |                  ^^^ 


1.10000002
2.09999990
3.09999990
(lf) root@LAPTOP-8SLTQKJO:~/lfortran#

Jatinagarwal24:
So @certik Yes it prints warning messages in both gfortran and lfortran.

certik:
What is the intended behavior of a loop with real variables? Is it the same as casting those to integer? If so, I would insert proper casts in ASR.

Which third-party code requires this feature?

If none, then I would hide this Fortran extension behind a compiler flag, not enable by default.

Jatinagarwal24:
> What is the intended behavior of a loop with real variables? Is it the same as casting those to integer? If so, I would insert proper casts in ASR.
> 
> Which third-party code requires this feature?
> 
> If none, then I would hide this Fortran extension behind a compiler flag, not enable by default.

The intended behavior of a real-controlled DO loop is a legacy extension where loop control is evaluated in the loop variable’s type. That means start/end/step are converted to the real loop variable's type, and progression/comparison use real arithmetic.

No, it is not the same as casting to an integer. If you simply compiled the expressions as integers in the ASR, a loop like do x = 0.0, 1.0, 0.1 would effectively become do x = 0, 1, 0, which strips out the fractional steps, changes the logic, and causes a step-zero error. The loop variable must retain its floating-point type and value inside the loop.

I think no third party currently require this feature.And we can use the a flag to complile this.

PR #10896, opened 2026-03-29
semantics: remove process-global mutable caches

krystophny:
## Summary
- move `vars_with_deferred_struct_declaration` and `assumed_rank_arrays` from file-scope `static` globals to `CommonVisitor` member variables
- eliminates process-global mutable state entirely, making ownership explicit per visitor instance

## Why
This is a generic LFortran cleanup extracted from the downstream LIRIC integration work in https://github.com/krystophny/lfortran/pull/41.

## Changes
- [`src/lfortran/semantics/ast_common_visitor.h`](https://github.com/lfortran/lfortran/pull/10896/files): remove two `static` globals from namespace scope, add them as `CommonVisitor` member variables

The change is minimal: both maps keep their names and usage. They are now default-constructed as part of each visitor instance instead of living as process-global statics.

### Reference tests pass
```
$ scripts/lf.sh test
TESTS PASSED
```

```
$ scripts/lf.sh itest -b llvm
```

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1596,8 +1596,8 @@ inline void validate_format_string(const std::string& fmt_str, const Location& l
 
 
 struct CommonVisitorState {
-    std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
-    std::map<std::string, int> assumed_rank_arrays;
+    std::map<std::string, std::vector<ASR::Variable_t*>> deferred_struct_vars;
+    std::map<std::string, int> assumed_rank_array_map;
 };
 
 template <class Derived>
@@ -2080,8 +2080,8 @@ public:
         CommonVisitorState &visitor_state,
             LCompilers::LocationManager &lm
     ): diag{diagnostics},
-          vars_with_deferred_struct_declaration{visitor_state.vars_with_deferred_struct_declaration},
-          assumed_rank_arrays{visitor_state.assumed_rank_arrays},
+          vars_with_deferred_struct_declaration{visitor_state.deferred_struct_vars},
+          assumed_rank_arrays{visitor_state.assumed_rank_array_map},
           al{al}, compiler_options{compiler_options},
           current_scope{symbol_table}, implicit_mapping{implicit_mapping},
           common_variables_hash{common_variables_hash},
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -29,8 +29,18 @@ using LCompilers::diag::Diagnostic;
 
 namespace LCompilers::LFortran {
 
-static std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
-static std::map<std::string, int> assumed_rank_arrays;
+inline std::map<std::string, std::vector<ASR::Variable_t*>> &
+get_vars_with_deferred_struct_declaration() {
+    static auto *vars =
+        new std::map<std::string, std::vector<ASR::Variable_t*>>();
+    return *vars;
+}
+
+inline std::map<std::string, int> &get_assumed_rank_arrays() {
+    static auto *arrays = new std::map<std::string, int>();
+    return *arrays;
+}
+
 static int PDT_SENTINEL = 1000;
 
 template <typename T>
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -29,8 +29,18 @@ using LCompilers::diag::Diagnostic;
...
+get_vars_with_deferred_struct_declaration() {
+    static auto *vars =
+        new std::map<std::string, std::vector<ASR::Variable_t*>>();
+    return *vars;
+}
+
+inline std::map<std::string, int> &get_assumed_rank_arrays() {
+    static auto *arrays = new std::map<std::string, int>();
+    return *arrays;
+}
```
certik: These should not be global at all, but rather local variables in the visitor, and since they need to be shared between common and symbol, should be passed explicitly in constructors of these visitors.

I think this change doesn't make it thread safe, does it? So it hides this global usage, but still stays global, right?

If so, let's refactor it properly into local visitor variables passed via a constructor.

krystophny: Right, the accessor wrapper was just hiding the same global. Made them plain member variables of `CommonVisitor` instead -- default-constructed, no constructor changes needed. Both maps are only used within a single visitor instance lifetime (deferred-struct within SymbolTableVisitor, assumed-rank within BodyVisitor), so instance members are the natural home. The PR now touches only `ast_common_visitor.h`: remove the two `static` globals, add the same maps as member variables.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -1601,6 +1599,8 @@ template <class Derived>
 class CommonVisitor : public AST::BaseVisitor<Derived> {
 public:
     diag::Diagnostics &diag;
+    std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
+    std::map<std::string, int> assumed_rank_arrays;
```
certik: If data has to be shared between the visitors, then we also need to pass it via a constructor. Let's first see if tests pass.

krystophny: Tests are passing, I add fix.

krystophny:
@certik what do you say?

certik:
Why not just:
```diff
diff --git a/src/lfortran/semantics/ast_common_visitor.h b/src/lfortran/semantics/ast_common_visitor.h
index d9337bbfc6..9e9552ee29 100644
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -29,8 +29,6 @@ using LCompilers::diag::Diagnostic;

 namespace LCompilers::LFortran {

-static std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
-static std::map<std::string, int> assumed_rank_arrays;
 static int PDT_SENTINEL = 1000;

 template <typename T>
@@ -1608,6 +1606,8 @@ template <class Derived>
 class CommonVisitor : public AST::BaseVisitor<Derived> {
 public:
     diag::Diagnostics &diag;
+    std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
+    std::map<std::string, int> assumed_rank_arrays;
     std::map<AST::operatorType, std::string> binop2str = {
         {AST::operatorType::Mul, "~mul"},
         {AST::operatorType::Add, "~add"},
```
I think it's equivalent and much easier to understand?

I am testing it here: https://github.com/lfortran/lfortran/pull/11530.

krystophny:
> Why not just:
> 
> ```diff
> diff --git a/src/lfortran/semantics/ast_common_visitor.h b/src/lfortran/semantics/ast_common_visitor.h
> index d9337bbfc6..9e9552ee29 100644
> --- a/src/lfortran/semantics/ast_common_visitor.h
> +++ b/src/lfortran/semantics/ast_common_visitor.h
> @@ -29,8 +29,6 @@ using LCompilers::diag::Diagnostic;
> 
>  namespace LCompilers::LFortran {
> 
> -static std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
> -static std::map<std::string, int> assumed_rank_arrays;
>  static int PDT_SENTINEL = 1000;
> 
>  template <typename T>
> @@ -1608,6 +1606,8 @@ template <class Derived>
>  class CommonVisitor : public AST::BaseVisitor<Derived> {
>  public:
>      diag::Diagnostics &diag;
> +    std::map<std::string, std::vector<ASR::Variable_t*>> vars_with_deferred_struct_declaration;
> +    std::map<std::string, int> assumed_rank_arrays;
>      std::map<AST::operatorType, std::string> binop2str = {
>          {AST::operatorType::Mul, "~mul"},
>          {AST::operatorType::Add, "~add"},
> ```
> 
> I think it's equivalent and much easier to understand?
> 
> I am testing it here: #11530.

Agreed. I thought the current solution was needed for performance reasons to avoid teardown cost, but this didn't hold. closing in favor of #11530 .

PR #10897, opened 2026-03-29
runtime: harden build-tree header and deferred-length handling

krystophny:
## Summary
- replace filesystem probing in `get_runtime_library_c_header_dir()` with a `LIBASR_RUNTIME_DIR` CMake define

## Why
This is the generic runtime/build robustness slice extracted from the downstream LIRIC integration work in https://github.com/krystophny/lfortran/pull/41.

The downstream branch exposed these clean-build-tree issues while wiring the LIRIC parity jobs, but the fixes themselves are not LIRIC-specific.

## Changes
- [`CMakeLists.txt`](https://github.com/lfortran/lfortran/pull/10897/files): add `LIBASR_RUNTIME_DIR` set to `${CMAKE_SOURCE_DIR}/src/libasr/runtime`
- [`src/lfortran/config.h.cmakein`](https://github.com/lfortran/lfortran/pull/10897/files): add `#cmakedefine LIBASR_RUNTIME_DIR`
- [`src/lfortran/utils.cpp`](https://github.com/lfortran/lfortran/pull/10897/files): use `LIBASR_RUNTIME_DIR` for Development/Ctest modes instead of filesystem probing

### Reference tests pass
```
$ scripts/lf.sh test
TESTS PASSED
```

```
$ scripts/lf.sh itest -b llvm
```


src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
```diff
@@ -59,12 +59,12 @@ module lfortran_intrinsic_iso_fortran_env
...
 contains
 function compiler_version() result(version)
-    character(len=:), allocatable :: version
+    character(:), allocatable :: version
     version = _lfortran_compiler_version() ! note: LFortran takes this and creates an IntrinsicElementalFunction
 end function compiler_version
 
 function compiler_options() result(options)
-    character(len=:), allocatable :: options
+    character(:), allocatable :: options
```
certik: This change is equivalent, right? What is the purpose of this change?

src/lfortran/utils.cpp
```diff
@@ -97,10 +97,28 @@ std::string get_runtime_library_c_header_dir()
 
     switch (execution_mode)
     {
-        case ExecutionMode::LFortranDevelopment:
+        case ExecutionMode::LFortranDevelopment: {
+            std::string header_dir = lfortran_exec_path_dir + "/../libasr/runtime";
+            if (std::filesystem::exists(header_dir + "/ISO_Fortran_binding.h")) {
+                return header_dir;
+            }
```
certik: That means we need to touch the filesystem even multiple times, so it will slow things down.

Let's figure out a solution that doesn't need to touch a filesystem.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
krystophny: The forms are equivalent in standard Fortran, but lfortran's parser doesn't support `character(len=:)` yet (fails with "Token 'len' is unexpected here"). Since the runtime is built with the just-compiled lfortran binary via `--backend=cpp`, this change is needed for the runtime build to succeed. Kept it in this PR since it showed up during the same build-tree work.

src/lfortran/utils.cpp
krystophny: Agreed. Replaced the filesystem probing with a `LIBASR_RUNTIME_DIR` cmakedefine that bakes `${CMAKE_SOURCE_DIR}/src/libasr/runtime` at configure time. Development and Ctest modes now both return that constant directly -- no filesystem access at all. The existing `CMAKE_INSTALL_INCLUDEDIR_RELATIVE` pattern for the installed case stays as-is.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
certik: Can you create an MRE for this parsing bug? 

I don't see any failure at our CI due to this. Should our CI fail too?

krystophny: Investigated further: the parse failure was from a stale binary before running build0.sh after rebase. The current parser on main handles `character(len=:)` fine -- CI would not fail. This change is cosmetic. Happy to revert it if you prefer keeping the `len=` form.

krystophny:
@certik can this be merged?

certik:
Let me test this by hand to ensure we can use absolute paths and don't need to use relative paths.

PR #10898, opened 2026-03-29
fix : diagnose interface function/subroutine attribute conflict

### LINKED ISSUE #6281: bug: function/subroutine with same name allowed when declared in different interfaces

```fortran
module functions_33

implicit none

interface frexp
   function frexp(x,n) result(r)
      real r
      real, intent(in), value :: x
      integer, intent(out) :: n
   end function frexp
end interface frexp

interface frexp_duplicate
   subroutine frexp(x,n)
      real r
      real, intent(in), value :: x
      integer, intent(out) :: n
   end subroutine frexp
end interface

end module functions_33
```

### Output

```console
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ gfortran -c test.f90
test.f90:14:19:

   14 |    subroutine frexp(x,n)
      |                   1
Error: FUNCTION attribute conflicts with SUBROUTINE attribute in ‘frexp’ at (1)
test.f90:15:12:

   15 |       real r
      |            1
Error: Unexpected data declaration statement in INTERFACE block at (1)
test.f90:16:34:

   16 |       real, intent(in), value :: x
      |                                  1
Error: Unexpected data declaration statement in INTERFACE block at (1)
test.f90:17:31:

   17 |       integer, intent(out) :: n
Error: Unexpected data declaration statement in INTERFACE block at (1)
test.f90:18:6:

   18 |    end subroutine frexp
      |      1
Error: Expecting END INTERFACE statement at (1)
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ lfortran -c test.f90
(lf) saurabh-kumar@Awadh:~/Projects/System/lfortran$ 
```

This happens because the subroutine `frexp` is declared with the original name as there is no name conflict with the interface.

Jatinagarwal24:
fixes #6281 
Detect function/subroutine attribute conflict in interface blocks

In ast_symboltable_visitor.cpp
Adds a post-rename symbol-table check in visit_Subroutine (after ~genericprocedure key resolution).
Detects when an existing symbol is actually a function (has return variable) and the new declaration is a subroutine.
Emits a semantic diagnostic:
FUNCTION attribute of '<name>' conflicts with SUBROUTINE attribute
Respects continue-compilation behavior:
Normal mode: throws SemanticAbort.
Continue mode: records error, erases existing symbol, continues.

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.json
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_1-04b6d40",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_1.f90",
-    "infile_hash": "e8df9d7e47c6a13a18f2172bd29cc35090562793178c00db2b87e5c0",
+    "infile_hash": "3209cb7e6af4258d27131534fa53acb436c9379ff626b9b3e7e1bb1a",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
     "stderr": "asr-continue_compilation_1-04b6d40.stderr",
-    "stderr_hash": "5d326eab5609e5da55154ce4fdaa04f256c949b6bfb908d55ade86ad",
+    "stderr_hash": "7f3b1c8b78756f22831306ed9dd7a8e2b2f357359125dec14203149a",
     "returncode": 1
 }


Jatinagarwal24:
Hi @certik I think this is good to merge.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1425,11 +1425,59 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                current_procedure_args.clear();
+                current_procedure_abi_type = ASR::abiType::Source;
+                current_function_dependencies = current_function_dependencies_copy;
+                in_Subroutine = false;
+                is_template = false;
+                is_global_save_enabled = is_global_save_enabled_copy;
+                return;
+            }
+            throw SemanticAbort();
+        };
```
kmr-srbh: We can declare interfaces inside a subroutine and possibly also a template so setting the variables to `false` is not quite right. Also, we should not hardcode the ABI type. Why do we need this? I think we can abort normally like other errors?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1425,11 +1425,59 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            if (ASR::is_a<ASR::Function_t>(*f_alt)) {
+                ASR::Function_t* f_alt_fn = ASR::down_cast<ASR::Function_t>(f_alt);
+                if (ASRUtils::get_FunctionType(f_alt_fn)->m_deftype == ASR::deftypeType::Interface &&
+                    f_alt_fn->m_return_var != nullptr) {
+                    std::vector<Location> locs = {f_alt->base.loc};
+                    std::vector<diag::Label> labels;
+                    labels.push_back(diag::Label("", {x.base.base.loc}));
+                    labels.push_back(diag::Label("is already declared as a function here", locs, false));
+                    diag.add(diag::Diagnostic(
+                        "function attribute conflicts with subroutine attribute in '" + sym_name + "'",
```
kmr-srbh:
```suggestion
                        "`function` attribute conflicts with `subroutine` attribute in '" + sym_name + "'",
```

Please highlight keywords in error messages.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1425,11 +1425,59 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                        diag::Level::Error, diag::Stage::Semantic, labels));
+                    abort_interface_conflict_subroutine();
+                    return;
```
kmr-srbh: I think we don't need a `return` here since we throw a `SemanticAbort()` anyways?

kmr-srbh (changes requested):
Please add a small comment (possibly with a small example) stating the case we are throwing an error for since function/function interfaces, subroutine/subroutine interfaces can be of the same name, like:

```fortran
interface frexp
   subroutine frexp(x,n)
      real r
      real, intent(in), value :: x
      integer, intent(out) :: n
   end subroutine frexp
end interface
```

Jatinagarwal24:
Hi, @kmr-srbh I have changed the fix pls once check now.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1496,6 +1496,32 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        ASR::symbol_t* existing_sym = parent_scope->get_symbol(sym_name);
+        if (existing_sym != nullptr) {
+            ASR::symbol_t* existing_past_ext = ASRUtils::symbol_get_past_external(existing_sym);
+            if (ASR::is_a<ASR::Function_t>(*existing_past_ext)) {
+                ASR::Function_t* existing_func = ASR::down_cast<ASR::Function_t>(existing_past_ext);
+                if (existing_func->m_return_var != nullptr) {
+                    std::string original_name = to_lower(std::string(x.m_name));
+                    diag.add(diag::Diagnostic(
+                        "FUNCTION attribute of '" + original_name +
+                        "' conflicts with SUBROUTINE attribute",
```
kmr-srbh:
```suggestion
                        "`function` attribute of '" + original_name +
                        "' conflicts with `subroutine` attribute",
```

kmr-srbh (approved):
A small change, otherwise looks good.

Jatinagarwal24:
Hi , @kmr-srbh I think now this is good to merge.

kmr-srbh:
Please resolve the conflicts.

kmr-srbh:
Thanks! This looks good.

PR #10899, opened 2026-03-29
feat: add support for namelist declarations

yuvrajsinghgmx:
Adds support for `namelist` declarations

`namelist_03`

```
    Start 277: namelist_03
1/1 Test #277: namelist_03 ......................   Passed    0.00 sec

Label Time Summary:
fortran     =   0.00 sec*proc (1 test)
gfortran    =   0.00 sec*proc (1 test)
llvm        =   0.00 sec*proc (1 test)

Total Test time (real) =   0.03 sec
```

### DIFF
--- a/src/libasr/codegen/asr_to_fortran.cpp
+++ b/src/libasr/codegen/asr_to_fortran.cpp
@@ -321,38 +321,23 @@ public:
     }
 
     void append_namelist_declarations(SymbolTable *symtab, std::string &r) {
-        std::vector<std::string> namelist_order;
         for (auto &item : symtab->get_scope()) {
             if (is_a<ASR::Namelist_t>(*item.second)) {
-                namelist_order.push_back(item.first);
-            }
-        }
-        // Insertion sort ideal for smaller arrays
-        for (size_t i = 1; i < namelist_order.size(); i++) {
-            std::string key = namelist_order[i];
-            int j = i - 1;
-            while (j >= 0 && namelist_order[j] > key) {
-                namelist_order[j + 1] = namelist_order[j];
-                j--;
-            }
-            namelist_order[j + 1] = key;
-        }
-        for (auto &name : namelist_order) {
-            ASR::symbol_t *sym = symtab->get_symbol(name);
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -320,6 +321,33 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
...
     }
 
+    void append_namelist_declarations(SymbolTable *symtab, std::string &r) {
+        std::vector<std::string> namelist_order;
+        for (auto &item : symtab->get_scope()) {
+            if (is_a<ASR::Namelist_t>(*item.second)) {
+                namelist_order.push_back(item.first);
+            }
+        }
+        std::sort(namelist_order.begin(), namelist_order.end());
```
jinangshah21: We shouldn't be using `algorithm` library if possible.

src/libasr/codegen/asr_to_fortran.cpp
```diff
@@ -320,6 +320,42 @@ class ASRToFortranVisitor : public ASR::BaseVisitor<ASRToFortranVisitor>
...
+        // Insertion sort ideal for smaller arrays
+        for (size_t i = 1; i < namelist_order.size(); i++) {
+            std::string key = namelist_order[i];
+            int j = i - 1;
+            while (j >= 0 && namelist_order[j] > key) {
+                namelist_order[j + 1] = namelist_order[j];
+                j--;
+            }
+            namelist_order[j + 1] = key;
+        }
```
yuvrajsinghgmx: @jinangshah21  i have implemented simple insertion sort best for smaller arrays as in this case for the `namelist`.

jinangshah21: But why do we need namelists name in order?

yuvrajsinghgmx: Agree with you, so removed it.
Thanks for pointing it out.

PR #10900, opened 2026-03-29
perf: reduce asr_verify frequency in Debug builds

krystophny:
## Summary
- Run `asr_verify()` once after all ASR passes instead of after every single pass
- Add `WITH_LFORTRAN_VERIFY_EVERY_PASS` CMake option for developers who need per-pass verification

## Why
In Debug builds (`WITH_LFORTRAN_ASSERT`), the pass manager runs `asr_verify()` after each of 29 passes. Each verification walks the entire ASR tree. This is the dominant cost in the ASR pass phase and accounts for ~60% of total compilation time in Debug mode.

The per-pass verification was added in commit `1030499` as a debug convenience to localize which pass corrupts the ASR tree. Moving to a single final verification still catches all corruption; the tradeoff is reduced debuggability (you know a pass corrupted ASR but not which one). The new CMake option `WITH_LFORTRAN_VERIFY_EVERY_PASS=yes` restores the per-pass behavior for developers debugging a specific pass.

**Stage:** Pass Manager

## Changes
- [`CMakeLists.txt`](https://github.com/lfortran/lfortran/pull/10897/files): add `WITH_LFORTRAN_VERIFY_EVERY_PASS` option (default `no`)
- [`src/libasr/config.h.in`](https://github.com/lfortran/lfortran/pull/10897/files): add `#cmakedefine WITH_LFORTRAN_VERIFY_EVERY_PASS`
- [`src/libasr/pass/pass_manager.h`](https://github.com/lfortran/lfortran/pull/10897/files): change per-pass verify guard from `WITH_LFORTRAN_ASSERT` to `WITH_LFORTRAN_VERIFY_EVERY_PASS`; add final-verify after all passes under `WITH_LFORTRAN_ASSERT && !WITH_LFORTRAN_VERIFY_EVERY_PASS`

### Speedup evidence (Debug build, LLVM 11, Linux x86_64)

Per-file compilation (`-c`, `--time-report`, median of 3 runs):

| File | Before | After | Speedup |
|------|--------|-------|---------|
| `print_01.f90` (10 lines) | 14ms | 6.7ms | **52% faster** |
| `intrinsics_170.f90` (371 lines) | 445ms | 177ms | **60% faster** |
| `bindc_iso_fb_05.f90` (1279 lines) | 1490ms | 544ms | **63% faster** |

ASR passes phase (medium file): **149ms -> 20ms** (7.4x faster)

Integration test build (`make -j8`, 3069 tests): **3m47s -> 1m17s** (66% faster)

### Reference tests pass
```
$ scripts/lf.sh test
TESTS PASSED
```

```
$ scripts/lf.sh itest -b llvm
```

### Integration tests with --fast pass (3069/3069)
```
$ scripts/lf.sh itest -b llvm -f
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,12 @@ endif()
 set(WITH_LFORTRAN_ASSERT ${WITH_LFORTRAN_ASSERT_DEFAULT}
     CACHE BOOL "Enable LFORTRAN_ASSERT macro")
 
+
+# WITH_LFORTRAN_VERIFY_EVERY_PASS: run asr_verify after every single ASR pass
+# (slow, useful only when debugging a specific pass that corrupts the ASR tree).
+# By default, Debug builds verify once after all passes complete.
+set(WITH_LFORTRAN_VERIFY_EVERY_PASS no
+    CACHE BOOL "Run asr_verify after every ASR pass (slow, for debugging passes)")
 # LFORTRAN_STATIC_BIN
 set(LFORTRAN_STATIC_BIN no CACHE BOOL "Build LFortran as a static binary")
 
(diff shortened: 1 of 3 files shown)


certik:
> Moving to a single final verification still catches all corruption; the tradeoff is reduced debuggability (you know a pass corrupted ASR but not which one).

Actually, it can easily happen that a pass creates an incorrect ASR, but it gets corrected later, so we'll never notice. It's a good idea to run this after every pass.

---

Regarding speed, use Release mode for speed. If you want checking, you can enable asserts in Release mode. It should still be very fast.

krystophny:
> > Moving to a single final verification still catches all corruption; the tradeoff is reduced debuggability (you know a pass corrupted ASR but not which one).
> 
> Actually, it can easily happen that a pass creates an incorrect ASR, but it gets corrected later, so we'll never notice. It's a good idea to run this after every pass.
> 
> Regarding speed, use Release mode for speed. If you want checking, you can enable asserts in Release mode. It should still be very fast.

understood

PR #10901, opened 2026-03-29
Fix 2 stdlib runtime test bugs

jinangshah21:


integration_tests/legacy_array_sections_18.f90
```diff
@@ -0,0 +1,60 @@
...
+        end if
+        if (a(1, 2) /= 221.0 .or. a(2, 2) /= 222.0) then
+            error stop "legacy_array_sections_18: 3d->2d column 2 mismatch"
+        end if
+        if (a(1, 3) /= 231.0 .or. a(2, 3) /= 232.0) then
+            error stop "legacy_array_sections_18: 3d->2d column 3 mismatch"
+        end if
+    end subroutine chunk_kernel_from_3d
+
+end program legacy_array_sections_18
```
certik, line 60: Why does this test require `--legacy-array-sections`? I can't see the usage.

jinangshah21: `call chunk_kernel_4(4, data(:, 1))`

Passed arg: `data(:, 1) `
Expected arg: `a(4, n) `

So need to convert `data(:, 1)` => `data(:, 1:) `

PR #10910, opened 2026-03-30
Leak fixes

certik:


src/libasr/codegen/llvm_utils.h
```diff
@@ -1712,9 +1712,26 @@ class ASRToLLVMVisitor;
...
-                || v->m_storage == ASR::Parameter
-                || v->m_storage == ASR::Save /*Neglect - Lives till program ends*/;
+            if (v->m_intent != ASR::Local) return true;
+            if (v->m_storage == ASR::Parameter) return true;
+            if (v->m_storage == ASR::Save) {
+                // Save variables in functions persist across calls and must not
+                // be finalized at function exit.  In the main Program, however,
+                // save struct variables have string members whose data is
+                // heap-allocated during struct initialization.  These must be
+                // finalized at program exit to avoid leaking that memory.
```
certik, line 1722: This should only happen when:

* `--detect-leaks` is on
* when the user wants to free everything when using other tools, like valgrind, or other reasons; likely we want to add an option `--free-all-memory` or something like that and enable this option for `--detect-leaks`

However, at the same time I think we can do this later as well. I think this is not a blocker to just free everything now, ensure all tests pass our leak detection, and then introduce `--free-all-memory` and otherwise not free it, which is a performance optimization, but not a correctness issue.

certik: I made this https://github.com/lfortran/lfortran/issues/10916.

Issue #7195, opened 2025-05-06
bug: `ClassProcedure` cannot point to a `subroutine` defined in a module


kmr-srbh:
### Minimal Reproducible Example

```fortran
module procedure_mod
   contains
   subroutine square(a)
      integer, intent(out) :: a
      a = a * 2
   end subroutine square
end module procedure_mod

module class_mod
   use procedure_mod

   type :: my_type
   contains
      procedure, public, nopass :: square
   end type my_type
  
end module

program main
   use class_mod
   implicit none

   type(my_type) :: t
   integer :: i

   i = 2
   call t%square(i)
   print *, "Value: ", i

end program main
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
 Value:            4
(lf) saurabhkumar@Mac lfortran % lfortran test.f90           
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2685
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2616
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 1111
    lcompilers_unique_ID = compiler_options.generate_object_code ? get_unique_ID() : "";
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 276
    Result<ASR::TranslationUnit_t*> res2 = get_asr3(*ast, diagnostics, lm);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/fortran_evaluator.cpp", line 298
    compiler_options.symtab_only, compiler_options, lm);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/semantics/ast_to_asr.cpp", line 78
    if (res.ok) {
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 4167
    v.visit_TranslationUnit(ast);
  File "/Users/saurabhkumar/OpenSource/lfortran/src/lfortran/semantics/ast_symboltable_visitor.cpp", line 82
    visit_ast(*x.m_items[i]);
[...]

PR #10911, opened 2026-03-30
fix: Resolve wrapper symbols to functions in class procedure bindings to prevent ICE

Jatinagarwal24:
Bug:
A type-bound procedure binding could carry a symbol wrapper (for example from use/external association). We were directly downcasting that wrapper to Function_t, which could trigger an ICE instead of resolving the actual underlying procedure symbol.

Fix:
Call symbol_get_past_external(proc_sym) before the Function_t downcast in class procedure handling, so semantic analysis operates on the real symbol and avoids the crash.
Resolves: #7195

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3333,7 +3333,7 @@ RUN(NAME procedure_46 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME procedure_47 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME procedure_48 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME procedure_pointer_abstract_01 LABELS gfortran llvm)
-
+RUN(NAME procedure_49 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocated_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocated_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME allocated_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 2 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3868,7 +3868,35 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                        proc_sym_underlying = ASRUtils::symbol_get_past_external(method_decl->m_proc);
+                    }
+                    if (proc_sym_underlying && ASR::is_a<ASR::Function_t>(*proc_sym_underlying)) {
+                        func = ASR::down_cast<ASR::Function_t>(proc_sym_underlying);
+                        func_sym = proc_sym_underlying;
+                    } else if (proc_sym_underlying && ASR::is_a<ASR::GenericProcedure_t>(*proc_sym_underlying)) {
+                        ASR::GenericProcedure_t* gp = ASR::down_cast<ASR::GenericProcedure_t>(proc_sym_underlying);
+                        for (size_t i = 0; i < gp->n_procs; i++) {
+                            ASR::symbol_t* gp_proc = ASRUtils::symbol_get_past_external(gp->m_procs[i]);
+                            if (gp_proc && ASR::is_a<ASR::Function_t>(*gp_proc)) {
```
jinangshah21: Can you provide example for `GenericProcedure` case? and why we are considering 1st function from it?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -3893,7 +3921,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 }
                 ASR::asr_t *v = ASR::make_StructMethodDeclaration_t(al, loc,
                     clss->m_symtab, name, pass_arg_name,
-                    proc_name, proc_sym, ASR::abiType::Source,
+                    proc_name, func_sym, ASR::abiType::Source,
```
jinangshah21: above whole change doesn't seem good. ICE will just go if you do `symbol_get_past_external` on proc_sym.

Jatinagarwal24: > above whole change doesn't seem good. ICE will just go if you do `symbol_get_past_external` on proc_sym.

Thanks for your suggestion.I did this and ICE is gone.

PR #10913, opened 2026-03-30
fix: add support for logical to integer cast

### LINKED ISSUE #3436: Logical to integer casting raises semantic error
### Minimal reproducible example:

For the below fortran program:
```f90
program ImplicitLogicalToIntCastExample
    implicit none
    
    logical :: logicalValue
    integer :: integerValue
    
    ! Assign a logical value
    logicalValue = .TRUE.
    
    ! Perform an implicit logical to integer cast
    integerValue = logicalValue
    print *, integerValue

end program ImplicitLogicalToIntCastExample
```

```
>>> gfortran a.f90 -o a.out && ./a.out

   11 |     integerValue = logicalValue
      |                   1
Warning: Extension: Conversion from LOGICAL(4) to INTEGER(4) at (1)
ld: warning: ignoring duplicate libraries: '-lemutls_w', '-lgcc'

1

>>> ./src/bin/lfortran a.f90 -o a.out && ./a.out
  --> a.f90:11:5
   |
11 |     integerValue = logicalValue
   |     ^^^^^^^^^^^^   ^^^^^^^^^^^^ type mismatch (integer and logical)


```

HarshilShah1804:
Fixes #3436

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2104,6 +2104,7 @@ RUN(NAME complex_div_test LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME complex_pow_test LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME complex_unary_minus_03 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME dcmplx_01 LABELS gfortran llvm) # dcmplx, dconjg
+RUN(NAME logical_to_integer_cast LABELS gfortran llvm EXTRA_ARGS --logical-casting NO_STD_F23)
 
 RUN(NAME logical1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran)
 RUN(NAME logical2 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm fortran)
--- a/integration_tests/logical_to_integer_cast.f90
+++ b/integration_tests/logical_to_integer_cast.f90
@@ -1,0 +1,25 @@
+program logical_to_integer_cast
+    implicit none
+    
+    logical :: logicalValue
+    integer :: integerValue
+
+    logical :: logicalArray(3)
+    integer :: integerArray(3)
+    
+    ! Assign a logical value
+    logicalValue = .TRUE.
+    
+    ! Perform an implicit logical to integer cast
+    integerValue = logicalValue
+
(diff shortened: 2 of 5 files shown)


src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -561,6 +561,58 @@ class ImplicitCastRules {
...
+                            new_array[i] = data[i] ? 1 : 0;
+                        }
+                        new_data = new_array;
+                    } else if (dest_kind == 1) {
+                        int8_t *new_array = al.allocate<int8_t>(array_size);
+                        for (size_t i = 0; i < array_size; i++) {
+                            new_array[i] = data[i] ? 1 : 0;
+                        }
+                        new_data = new_array;
+                    }
```
kmr-srbh, line 607: We can make the if-else blocks only check the kind and allocate the array, then run the loop at the end.

integration_tests/integer_to_logical_cast.f90
```diff
@@ -0,0 +1,16 @@
...
+    
+    ! Assign a logical value
+    logicalValue = .TRUE.
+    
+    ! Perform an implicit logical to integer cast
+    integerValue = logicalValue
+    print *, integerValue
+
+    if (integerValue /= 1) error stop
+
```
kmr-srbh: Please add tests for arrays.

src/lfortran/semantics/asr_implicit_cast_rules.h
HarshilShah1804: I think it would complicate things up as all the arrays are of different types. Also, we handle every other casting in a similar way. So, I think it's fine.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5812,6 +5812,16 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                             array_reshape->m_type = ASRUtils::duplicate_type(al, array_reshape->m_type, &array_reshape_dims, ASR::array_physical_typeType::DescriptorArray,true);
                         }
                     }
+                    if (ASRUtils::is_logical(*value_type) && ASRUtils::is_integer(*target_type)) {
+                        if (!compiler_options.logical_casting) {
+                            diag.add(Diagnostic(
+                                "Implicit casting from LOGICAL to INTEGER is not allowed by default. Use `--logical-casting` flag to enable it.",
```
kmr-srbh, line 5837: Please add an error test for this.

PR #10917, opened 2026-03-31
Leak Fixes

Jatinagarwal24:


Jatinagarwal24:
Hi @certik pls review this.

Jatinagarwal24:
Hi @assem2002 Pls review this.

assem2002:
Thanks for the fix. I can see it fixed `test_str` + `test_rep` which are python-ported tests.
It does the job, but we're trying to avoid the approach of registering pointers and then finalize them. Now we only care about the ASR variables. It's the cleanest so far.

I see no harm from merging this, but I'd love to see backend not caring about monitoring pointers around.
@certik What do you think?

certik (approved):
I think it's ok to merge, since it fixes lots of leaks. Later we can improve it to not use pointers in the first place.

Issue #10916, opened 2026-03-31
Introduce `--free-all-memory` to free all unused memory on program exit


certik:
This should only happen when:

* `--detect-leaks` is on
* when the user wants to free everything when using other tools, like valgrind, or other reasons; likely we want to add an option `--free-all-memory` or something like that and enable this option for `--detect-leaks`

However, at the same time I think we can do this later as well. I think this is not a blocker to just free everything now, ensure all tests pass our leak detection, and then introduce `--free-all-memory` and otherwise not free it, which is a performance optimization, but not a correctness issue.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/10910#discussion_r3012881840_

PR #10931, opened 2026-04-01
Add --free-all-memory flag (implied by --detect-leaks)

devalgupta404:
issue - #10916 
- Added new CLI Flag --free-all-memory to free all unused memory on program exit
- Updates --detect-leaks to imply --free-all-memory
- It keeps the current behavior unchanged so memory is still freed in any condition


certik:
Implement the actual freeing also, to see if this approach will work.

PR #10933, opened 2026-04-01
fix: reject DO loop variable as INTENT(OUT/INOUT) subroutine argument

### LINKED ISSUE #6006: Bug: Modification of iterating variable in do_loop
MRE

```fortran
program iterator
   implicit none

   integer :: i,sum_sub=0
   do  i = 1,4
      call SUB(i)
      sum_sub = sum_sub + i
   end do

   print *,"SUB sum->", sum_sub

contains

   pure subroutine SUB(x)
      integer, intent(inout) :: x
       x = x * x
   end subroutine SUB

end program iterator
```
Gives  output on LFortran and error in GFortran As Discussed in [Here on Zulip](https://lfortran.zulipchat.com/#narrow/channel/452774-Internal/topic/Modification.20of.20iterating.20var.20in.20loops)

```clojure
 aditya-trivedi   lfortran    pure ↑10 ↓6  ~1    lfortran ./integration_tests/iterator.f90 
SUB sum->    5
ERROR STOP
 aditya-trivedi   lfortran    pure ↑10 ↓6  ~1    gfortran ./integration_tests/iterator.f90 
./integration_tests/openmp_43.f90:6:15:

    5 |    do  i = 1,4
      |              2 
    6 |       call SUB(i)
      |               1
Error: Variable ‘i’ at (1) not definable inside loop beginning at (2) as INTENT(INOUT) argument to subroutine ‘sub’
```

Documentation can be found here
https://j3-fortran.org/doc/year/18/18-007r1.pdf and https://support.nag.com/nagware/np/r72_doc/nag_f2018.html.

Jatinagarwal24:
Fixes #6006 
Fortran requires that a DO loop control variable must not be defined or become undefined inside the loop body.
LFortran already rejected direct assignments to the loop variable, but still allowed passing it to subroutines with intent(out) or intent(inout), which can redefine it.

This change adds a semantic check in subroutine-call handling:
detect actual arguments that are DO loop variables
 
map them to callee dummy arguments
if dummy intent is OUT or INOUT, emit a semantic error and abort semantic pass for that statement
Also adds/updates continue-compilation coverage with a case that passes a DO loop variable to an intent(inout) dummy argument.

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -1390,6 +1390,12 @@ semantic error: Duplicate value of `decimal` found
 636 |     open(unit=7, decimal="POINT", decimal="comma")
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 
+semantic error: Variable 'i' is not definable inside a DO loop as INTENT(INOUT) argument to subroutine 'sub_do_intent_loop'
+   --> tests/errors/continue_compilation_1.f90:644:32
+    |
+644 |        call sub_do_intent_loop(i)
+    |                                ^ 
+
 semantic error: Variable 'k' is not declared
    --> tests/errors/continue_compilation_1.f90:668:12
     |
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -639,10 +639,10 @@ program continue_compilation_1
     interface undeclared_iface
         module procedure undeclared_proc  ! {Error} Symbol 'undeclared_proc' not declared
     end interface
-
-
-
-
+    ! do_loop_intent_inout
+    do i=1,3
+       call sub_do_intent_loop(i)
+    end do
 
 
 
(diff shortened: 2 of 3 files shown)


jinangshah21:
Sorry I just realized flang allows this. so MRE you added works with flang. I also think this shouldn't give error.

Jatinagarwal24:
> Sorry I just realized flang allows this. so MRE you added works with flang. I also think this shouldn't give error.

https://github.com/lfortran/lfortran/issues/6006#issuecomment-2591071655
Here @certik said that this should report error.

Jatinagarwal24:
@jinangshah21 So what to do pls suggest.

Jatinagarwal24:
> Sorry I just realized flang allows this. so MRE you added works with flang. I also think this shouldn't give error.

Is it like that there is bug in gfortran also?

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -1390,6 +1390,12 @@ semantic error: Duplicate value of `decimal` found
 636 |     open(unit=7, decimal="POINT", decimal="comma")
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 
+semantic error: Variable 'i' is not definable inside a DO loop as INTENT(INOUT) argument to subroutine 'sub_do_intent_loop'
```
certik, line 1393: "Loop variable cannot be modified inside the loop."

You can then explain that it is being modified inside the subroutine.

certik:
Research when exactly this error message should be given --- for any function with intent out/inout arguments on the loop variable, or only when the function actually modify the variable.

Run all tests, ensure they pass.

PR #10936, opened 2026-04-01
fix Memory-Leak : Save variables leak

### LINKED ISSUE #10928: Memory-Leak : Save variables leak
```.f90
program pointer_04
    implicit none
    call ss()
    contains
    subroutine ss()
        integer, allocatable, save :: arr(:)
        allocate(arr(10))
    end subroutine 
end program pointer_04
```
```
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran leak_01.f90 --detect-leaks

   ==> LEAK: 40 bytes

TOTAL LEAKS FOUND : 1, 40 bytes total
```
The leak is for the data allocated  (10*4)

***
I'd love to flag those kind of variables as leakable variables instead of attempting to free them as they add unnecessary overhead in the finalization functions -- but still finalizing those still a valid option.

jassem-manita:
fix #10928
implemented the desired leakable approach for save allocatables

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1765,12 +1765,12 @@ public:
     }
 
     void maybe_mark_save_array_allocation_leakable(ASR::expr_t* target_expr,
-            ASR::ttype_t* array_type, llvm::Type* desc_type, llvm::Value* desc_slot,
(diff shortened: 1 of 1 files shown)


assem2002:
Can you please try to push a fix where we actually finalize the save variables? instead of bypassing them.
We can compare the 2 approaches and see how it goes.

We'd need to collect all save variables across different symboltables, then finalize them through `finalize_variable`

PR #10937, opened 2026-04-01
 fix: handle sequential character and int64 read in unformatted read

jinangshah21:
Towards `stdlib`


certik:
```
2145/3077 Test #2033: file_open_01_FAST .......................................***Exception: SegFault  0.18 sec
```

certik (approved):

PR #10938, opened 2026-04-01
Memory-Leak : Use heap allocate class wrapper instead of stack + free it at finalization

### LINKED ISSUE #10927: bug : dangling pointer as class wrapper
```.f90
program pointer_04
    implicit none

    type :: base
        integer :: v
    end type

    class(base), pointer :: y
    class(base), pointer :: mmm
    call ss(y, 1000)
    call ss(mmm, 2222)
    print *, y%v
    if(y%v /= 1000) error stop
    print *, mmm%v
    if(mmm%v /= 2222) error stop
    deallocate(y)
    deallocate(mmm)

    contains
    subroutine ss(xxx, inn)
        integer :: inn
        class(base), pointer :: xxx
        type(base), pointer :: t
        allocate(t)
        t%v = inn
        xxx => t ! Created class wrap {vtable*, struct*} -- It's being created on stack, which results in a dangling pointer.
    end subroutine 
end program pointer_04
```
```
2222
ERROR STOP
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt3.f90 && ./a.out 
        1000
        2222
```
You can refer to this code snippet [HERE](https://github.com/lfortran/lfortran/pull/10914/changes/BASE..286a04c86795e7ef60b9461b466edbac9dbce21d#r3014356211) where the dangling issue got introduced.

### LINKED ISSUE #10929: Memory-Leak : Associate class TO strings leak
```.f90
program a
  implicit none
  class(*), pointer :: ptr
  
  call ss(ptr)
  contains 
  subroutine ss(xx)
    class(*), pointer :: xx
    character(len=10), pointer :: ii
    allocate(ii)
    xx => ii
    deallocate(ii)
    
  end subroutine ss
end program a
```
```
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ lfortran leak_02.f90 --detect-leaks 

   ==> LEAK: 16 bytes

TOTAL LEAKS FOUND : 1, 16 bytes total
```
The class wrapper leaks.
refer to https://github.com/lfortran/lfortran/pull/10914/changes#r3014403535

assem2002:
fixes #10929
fixes #10927 
***
### What is in the PR
- Use heap allocated class wrapper for class types instead of mixing stack and heap choices.
- Always free that wrapper in finalizaion process.
- Register tests

### DIFF
--- a/tests/reference/pass_nested_vars-nested_call_filter_01-5a87413.json
+++ b/tests/reference/pass_nested_vars-nested_call_filter_01-5a87413.json
@@ -6,7 +6,7 @@
     "outfile": null,
     "outfile_hash": null,
     "stdout": "pass_nested_vars-nested_call_filter_01-5a87413.stdout",
-    "stdout_hash": "9183130475b060da25f4dd479d11a10d560c7a13bcf83d1f1ec7122c",
+    "stdout_hash": "dc7d5357b3eff840f72227c6cc6052971efc0aa6da8206e57334298e",
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
--- a/tests/reference/pass_nested_vars-nested_call_filter_01-5a87413.stdout
+++ b/tests/reference/pass_nested_vars-nested_call_filter_01-5a87413.stdout
@@ -12,7 +12,7 @@
                                     7
                                     a
                                     []
-                                    Unspecified
+                                    Local
                                     ()
                                     ()
                                     Default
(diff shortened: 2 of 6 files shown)


certik:
I think the change would be ok, but let's get the tests to pass. Then run all tests.

assem2002:
Most of the new-added work is properly allocating a class wrapper on heap and freeing it properly.

assem2002:
Yes you can say so. It's not complex by nature;
It's just the codebase is assuming the wrapper is not an atomic piece of the type, so we have to allocate on heap now (don't do a mix between stack and heap allocation) to just proceed with leak fixes.
It'd be great to use a stack-allocated one, it shouldn't be a hard refactor btw.

certik:
Ok. Let's use a heap one for now, and later refactor.

PR #10944, opened 2026-04-02
fix: handle assumed-size array in implicit interface function call

certik:
When an assumed-size array (e.g., real Y(*)) is passed through an implicit interface procedure argument, type duplication in ASR passes incorrectly changes UnboundedPointerArray to DescriptorArray because make_Array_t_util cannot detect assumed-size arrays from empty dimensions alone.

This causes an assertion failure (m_dim.m_start != nullptr) in the LLVM codegen when PointerToData_to_Descriptor tries to fill array details with null dimension bounds.

Fix:
- In Call_t_body, correct DescriptorArray back to UnboundedPointerArray when the actual argument is UnboundedPointerArray and the formal parameter has all-empty dimensions (indicating a corrupted type).
- In PointerToData_to_Descriptor, skip fill_array_details when dimension bounds are null as a safety net.

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -7801,6 +7801,16 @@ static inline void Call_t_body(Allocator& al, ASR::symbol_t* a_name,
                 ASRUtils::type_get_past_pointer(arg_type));
             ASR::Array_t* orig_arg_array_t = ASR::down_cast<ASR::Array_t>(
                 ASRUtils::type_get_past_pointer(orig_arg_type));
+            // When the actual argument is UnboundedPointerArray and the formal
+            // parameter was incorrectly changed to DescriptorArray (e.g. by
+            // type duplication that doesn't preserve UnboundedPointerArray),
+            // skip inserting a cast — the argument already matches the
+            // intended calling convention for assumed-size arrays.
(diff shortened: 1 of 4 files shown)


src/libasr/asr_utils.h
```diff
@@ -7631,6 +7621,16 @@ static inline void Call_t_body(Allocator& al, ASR::symbol_t* a_name,
...
+            // When the actual argument is UnboundedPointerArray and the formal
+            // parameter was incorrectly changed to DescriptorArray (e.g. by
+            // type duplication that doesn't preserve UnboundedPointerArray),
+            // skip inserting a cast — the argument already matches the
+            // intended calling convention for assumed-size arrays.
+            if (arg_array_t->m_physical_type == ASR::array_physical_typeType::UnboundedPointerArray &&
+                orig_arg_array_t->m_physical_type == ASR::array_physical_typeType::DescriptorArray &&
+                ASRUtils::is_dimension_empty(orig_arg_array_t->m_dims, orig_arg_array_t->n_dims)) {
+                orig_arg_array_t->m_physical_type = ASR::array_physical_typeType::UnboundedPointerArray;
+            }
```
certik: This seems to be treating the symptoms, not the cause.

PR #10946, opened 2026-04-02
Fix Metal VLA codegen: use device buffer workspace for variable-lengt…

certik:
…h arrays

Metal Shading Language does not support variable-length arrays (VLAs). When a do concurrent block contains a local array with a runtime-dependent dimension (e.g., real :: a(n)), the Metal codegen was emitting float a[n] which the Metal compiler rejects.

Fix: convert each block-local VLA into a per-thread slice of an additional device buffer parameter. The Metal codegen emits a device pointer offset (device float* a = __vla_a + __thread_id * n) instead of a stack VLA. The LLVM host-side codegen allocates the workspace buffer via malloc, passes it as an extra buffer arg, and frees it after kernel launch.

Also add missing _lcompilers_int_/_lcompilers_nint_ intrinsic handling in the Metal FunctionCall emitter.

### DIFF
--- a/src/libasr/utils.h
+++ b/src/libasr/utils.h
@@ -100,19 +100,6 @@ struct CompilerOptions {
     bool target_offload_enabled = false;
     std::string gpu_backend = "";
     std::string gpu_metal_source = "";
-    // Metadata for VLA workspace buffers in Metal GPU kernels.
-    // Populated by asr_to_metal, consumed by asr_to_llvm.
-    struct GpuVlaDim {
-        bool is_constant;
-        int64_t constant_value;
-        size_t call_arg_index;
-    };
-    struct GpuVlaWorkspace {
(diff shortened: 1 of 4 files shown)


src/libasr/utils.h
```diff
@@ -100,6 +100,19 @@ struct CompilerOptions {
...
+        bool is_constant;
+        int64_t constant_value;
+        size_t call_arg_index;
+    };
+    struct GpuVlaWorkspace {
+        int buffer_index;
+        int elem_size;
+        std::vector<GpuVlaDim> dims;
+    };
+    std::vector<GpuVlaWorkspace> gpu_vla_workspaces;
```
certik: I don't like this fix, we are now passing state other than in ASR, so likely this is not the right answer.

PR #10961, opened 2026-04-04
Fix Leaks 

Jatinagarwal24:


Jatinagarwal24:
Hi @certik @assem2002 Pls review this.

src/libasr/codegen/asr_to_llvm.cpp
assem2002, line 8552: If you can remove this else branch without having any failing tests, it'd be great. We shouldn't allocate any descriptors on heap.

I think it should work. The preceding `if` clause is true with all cases but a `ReturnVar` intent. In the backend we mostly don't have `ReturnVar` arrays/structs/strings (any aggregate type)

certik:
Let's merge this, since the change I think is ok. Then let's investigate @assem2002's suggestion if we can simplify things further.

PR #10966, opened 2026-04-05
Better GPU Fortran printing support

certik:


src/bin/lfortran.cpp
```diff
@@ -974,6 +975,20 @@ int dump_all_passes(const std::string &infile, CompilerOptions &compiler_options
         compiler_options.po.run_fun = "f";
         pass_manager.dump_all_passes(al, asr.result, compiler_options.po, diagnostics, lm);
         std::cerr << diagnostics.render(lm, compiler_options);
+
+        if (compiler_options.gpu_backend == "metal") {
+            LCompilers::diag::Diagnostics metal_diag;
+            LCompilers::Result<std::string> metal_res =
+                LCompilers::asr_to_metal(al, *asr.result, metal_diag, compiler_options);
```
certik, line 982: I think this is running the backend twice, we should improve it to only run once.

PR #10970, opened 2026-04-05
LSP: Fix document outline to show derived-type members

### LINKED ISSUE #39: Bug : Structs doesn't identify its members
It doesn't work for both of the following cases : 
Screenshot from 2024-12-19 16-39-58
Screenshot from 2024-12-19 16-40-50
It partly works here, but the symbol points to the `externalSymbol` found in the `program` symbol Table. 
Screenshot from 2024-12-19 16-41-09

himanshumahajan138:
Fixes [lfortran/lfortran-vscode-client#39](https://github.com/lfortran/lfortran-vscode-client/issues/39)

## Summary
Derived-type members (and Enum/Union members) were missing from the VSCode outline because `populateSymbolLists` never recursed into `Struct_t`/`Enum_t`/`Union_t`. On top of that, synthetic `ExternalSymbol` entries (e.g. `1_aa_x`) leaked into the CLI output.

Fix: recurse into Struct/Enum/Union symbol tables and skip `ExternalSymbol` entries in [src/bin/lfortran_accessor.h](src/bin/lfortran_accessor.h).

## Before 
Screenshot 2026-04-05 at 3 02 01 PM

## After 
Screenshot 2026-04-05 at 3 04 02 PM

## Verification
- New regression test [tests/document_symbols2.f90](tests/document_symbols2.f90) registered in [tests/tests.toml](tests/tests.toml)
- `export PATH="$PWD/build/src/bin:$PATH" ./run_tests.py -t document_symbols -s` passes
- Rebaselined [document_symbols1](tests/reference/asr-document_symbols1-8238851.stdout) to drop the 4 stale `ExternalSymbol` lines (client already filtered these by filename, so no user-visible change)


himanshumahajan138:
Hi @certik tests passing needs review

src/bin/lsp_cli.cpp
```diff
@@ -115,6 +115,14 @@ namespace LCompilers {
...
 
         for (auto symbol : symbol_lists) {
+            // ExternalSymbol entries are compiler artifacts (mirror of
+            // `use`-imported symbols and synthetic member-access lookups).
+            // They should not appear in the document outline (see
+            // lfortran/lfortran-vscode-client#39). They remain in the
+            // underlying symbol list so completion can still reach them.
+            if (symbol.symbol_type == ASR::symbolType::ExternalSymbol) {
+                continue;
+            }
```
certik, line 122: I think we want to include them. For example:
```
program test
use A, only: x
end
```
Should have `x` in the outline for the main program.

Because it's a symbol that is available to be used. On `use A`, it imports all the symbols into scope, so it pollutes the scope, but I think we should still print them.

How does the Python LSP handle `from A import x` and `from A import *`?

himanshumahajan138: Good point @certik ,  the `1_<parent>_<member>` mangled name is the actual artifact, while `use A, only: x` ExternalSymbols are user visible and should stay. I'll narrow the filter to that name pattern and restore the use imported entries in the `document_symbols1` baseline.

src/bin/lfortran_lsp_language_server.cpp
```diff
@@ -736,6 +736,14 @@ namespace LCompilers::LanguageServerProtocol {
...
             roots.reserve(symbols.size());
             for (const auto &symbol : symbols) {
+                // ExternalSymbol entries are compiler artifacts (mirror of
+                // `use`-imported symbols and synthetic member-access lookups).
+                // They should not appear in the document outline (see
+                // lfortran/lfortran-vscode-client#39). They remain in the
+                // underlying symbol list so completion can still reach them.
+                if (symbol.symbol_type == ASR::symbolType::ExternalSymbol) {
+                    continue;
+                }
```
certik, line 743: ExternalSymbol is what the user imported into the given scope, so I think we need to include them.

src/bin/lsp_cli.cpp
certik: We should not special case `1_`, that's not going to cover everything and we should investigate why we have these in the first place and create a proper fix.

himanshumahajan138:
@certik ready for another review

PR #10973, opened 2026-04-05
Leak Fix

Jatinagarwal24:


src/libasr/codegen/llvm_utils.h
```diff
@@ -1162,6 +1162,38 @@ class ASRToLLVMVisitor;
...
+                                    llvm_utils_->character_type,
+                                    llvm_utils_->create_gep2(llvm_utils_->string_descriptor, str_desc, 0));
+                                auto const char_not_null = builder_->CreateICmpNE(
+                                    char_ptr,
+                                    llvm::ConstantPointerNull::get(llvm_utils_->character_type));
+                                llvm_utils_->create_if_else(char_not_null,
+                                    [this, char_ptr]() { llvm_utils_->lfortran_free_nocheck(char_ptr); },
+                                    [](){});
+                            }, [](){});
+                            llvm_utils_->lfortran_free_nocheck(data_ptr);
```
assem2002: I think you can `finalize` -- Shouldn't it be doing the same work?

Jatinagarwal24: Hi @assem2002 .Yes we can do it by finalize i have implemented that now pls once again review it.
Thanks!

PR #10979, opened 2026-04-06
More GPU fixes

certik:


src/libasr/pass/gpu_offload.cpp
```diff
@@ -3738,6 +4508,154 @@ class GpuOffloadVisitor : public ASR::StatementWalkVisitor<GpuOffloadVisitor>
...
+                                        std::string content;
+                                        if (!read_file(
+                                                file.path().string(),
+                                                content) ||
+                                            content.empty())
+                                            continue;
+                                        LocationManager lm_tmp;
+                                        auto res = load_modfile(
+                                            al, content, false,
+                                            *tu.m_symtab, lm_tmp);
```
certik, line 4602: This is ugly, we should not be loading submodules from here, but at the same time there is no good option how to allow submodule functions inside do concurrent when compiled separately, since we need their source code to offload. I think for now it's fine, we can change that later easily. I created an issue https://github.com/lfortran/lfortran/issues/10983 to figure out a good solution.

PR #10982, opened 2026-04-06
fix: update module handling

yuvrajsinghgmx:


yuvrajsinghgmx:
@HarshitaKalani please review this.

src/bin/lfortran.cpp
```diff
@@ -1821,7 +1840,27 @@ int compile_to_binary_fortran(const std::string &infile,
...
+    if (ec) {
+        std::cout << "Unable to create module directory '" + gfortran_mod_dir.string()
+            + "': " + ec.message() << std::endl;
+        return 11;
+    }
+
+    std::string gfortran_mod_dir_s = gfortran_mod_dir.string();
+    std::string out_abs = std::filesystem::absolute(outfile).string();
+    std::string in_abs = std::filesystem::absolute(in_file).string();
+    std::string cmd = "cd '" + gfortran_mod_dir_s
```
HarshitaKalani: I guess this introduces a dependency on shell behavior. Would it be safer to pass -J and -I with absolute paths instead of changing directories?

yuvrajsinghgmx: Fixed, I have removed the shell dependency now.

Passing -J and -I and was not working as expected.

HarshitaKalani (approved):
I think this is good. Thanks!

src/bin/lfortran.cpp
```diff
@@ -1823,6 +1799,25 @@ int compile_to_binary_fortran(const std::string &infile,
...
+            fe.get_asr2(input, asr_lm, asr_diagnostics);
+        std::cerr << asr_diagnostics.render(asr_lm, compiler_options);
+        if (!asr.ok) {
+            LCOMPILERS_ASSERT(asr_diagnostics.has_error())
+            return 1;
+        }
+        asr_diagnostics.diagnostics.clear();
+        int err = save_mod_files(*asr.result, compiler_options, asr_lm);
+        if (err) return err;
+    }
```
certik, line 1820: Are we doing the compilation twice? Why is this needed? When exactly is this used?

src/bin/lfortran.cpp
```diff
@@ -1845,8 +1840,31 @@ int compile_to_binary_fortran(const std::string &infile,
...
+        return 11;
+    }
+    std::string out_abs = std::filesystem::absolute(outfile).string();
+    std::string in_abs = std::filesystem::absolute(in_file).string();
+    std::filesystem::path original_dir = std::filesystem::current_path();
+    std::filesystem::current_path(gfortran_mod_dir, ec);
+    if (ec) {
+        std::cout << "Unable to enter module directory: " << ec.message() << std::endl;
+        return 11;
+    }
```
certik, line 1864: This is only used when compiling with gfortran?

certik:
I left some questions, we have to be careful when modifying the main driver, so let's make sure these are the correct changes first.

PR #10984, opened 2026-04-06
LSP: Fix filter document symbols to current file only

### LINKED ISSUE #39: Bug : Structs doesn't identify its members

### LINKED ISSUE #60: The outline is incorrect for expr2.f90/a.f90
For the code from https://github.com/lfortran/lfortran-lsp/issues/55, here is the outline that I can see for expr2.f90:

image

There are several issues:
* The root node is "a", but rather the root node should be "expr2".
* There are two symbol "x". The first one points to line 2, as it should. The second one points to line 4, which is empty. It likely is pointing to line 4 in file `a.f90`, but this symbol should not be part of the outline at all.

himanshumahajan138:
## Summary

Fixes lfortran/lfortran-vscode-client#60

Depends on #10970 — should be merged after that PR lands.

When viewing the document outline for a file that `use`s a module from another file, symbols from the module file were incorrectly included in the outline. This PR filters document symbols to only include symbols belonging to the current document.

- Added filename filter in the `--show-document-symbols` CLI path (`lsp_cli.cpp`)
- Added filename filter in the flat `SymbolInformation` LSP response path (`lfortran_lsp_language_server.cpp`)
- Note: The hierarchical `DocumentSymbol` path already had this filter (commit 7e329ebc5d)

## Relationship with #10970

PR #10970 fixes lfortran/lfortran-vscode-client#39 (derived-type members missing + ExternalSymbol leak in `lfortran_accessor.h`). This PR addresses a separate but related issue: symbols from imported module *files* appearing in the current file's outline. Both PRs modify the `document_symbols1` references and add a `document_symbols2` test, so this PR will need a rebase after #10970 merges.

## Before
Screenshot 2026-04-06 at 7 17 27 PM

## After
Screenshot 2026-04-06 at 7 20 41 PM

## Test Plan

- Added `document_symbols2.f90` test (exact MRE from the issue: program that uses `a module` from another file)
- Updated `document_symbols1` reference output (was including symbols from the module file)
- All document symbols reference tests pass:


certik:
I think we need to include all symbols from the current document. Is this including another module, or what exactly is it including that this PR removes?

PR #10986, opened 2026-04-06
GPU Metal offloading fixes

certik:


src/libasr/asr_utils.h
```diff
@@ -4551,7 +4551,18 @@ inline bool types_equal(ASR::ttype_t *a, ASR::ttype_t *b, ASR::expr_t* a_expr, A
...
+                        return true;
+                    }
+                    // The GPU offload pass may duplicate Struct definitions
+                    // into kernel scope, producing distinct Struct_t pointers
+                    // for what is logically the same type.  Accept as equal
+                    // when both structs share the same name.
+                    if (std::string(x_struct->m_name) ==
+                        std::string(y_struct->m_name)) {
+                        return true;
+                    }
```
certik: I think this might be problematic, possibly a bug, since two different types could have the same name in separate modules I think.

certik: Fixed in b812ce3c7bb6570adc0ce78a26ae5b241139df5e.

PR #10987, opened 2026-04-06
Leak Fix

Jatinagarwal24:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2266,6 +2280,17 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    int64_t saved_ptr_loads = ptr_loads;
+                    ptr_loads = 1;
+                    visit_expr_load_wrapper(x.m_args[0].m_a, 1, true);
+                    llvm::Value* desc_ptr = tmp;
+                    ptr_loads = saved_ptr_loads;
+                    llvm_symtab_finalizer.finalize_before_deallocate(
+                        desc_ptr, realloc_arg_type, realloc_struct_sym, false);
+                }, [](){});
+            }
             visit_AllocateUtil(x, nullptr, true);
```
assem2002, line 2329: We shouldn't finalize before realloc, Realloc sometimes just extends current pointer which is good for speed.

I can see we must finalize due to nature of the type -- it could contain types that allocate on heap which the reallocation will leak for sure.
***
### How to proceed on
The best way is to set `realloc_struct_sym` to TRUE only with : 
An array of struct type that has : 
1- any allocatable type 
2- a character type.

I think those are the ones that really require de-allocation.

assem2002: We should be smarter on that later, We can check the returned pointer from `realloc` against the one we have.
- Equal -> don't deallocate 
- Not Equal -> deallocate (only specific set of type)

assem2002:
Please address my comment. I'll put it on draft for now.

Jatinagarwal24:
Hi @assem2002 ,Thanks for your suggestions . I have implemented them pls check once again. Thanks!

assem2002 (approved):
Looks good to me. Thanks.

PR #10989, opened 2026-04-07
Athena fixes

HarshitaKalani:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -13,9 +13,51 @@ namespace LCompilers {
...
+                                    llvm::IRBuilder<> B(CI);
+                                    llvm::Value *cast = B.CreateBitCast(arg, param_ty);
+                                    CI->setArgOperand(i, cast);
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
certik: This doesn't look right. We should not be fixing pointer mismatches in LLVM by querying LLVM. Instead, we should know in ASR that there are mismatches and generate correct LLVM code on the first try. Likely there are some missing or incorrect casts in ASR.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -13,9 +13,51 @@ namespace LCompilers {
 
         llvm::Value* CreateStore(llvm::IRBuilder<> &builder, llvm::Value *x, llvm::Value *y) {
             LCOMPILERS_ASSERT(y->getType()->isPointerTy());
+            if (x->getType()->isPointerTy() && y->getType()->getNumContainedTypes() > 0) {
+                llvm::Type* dest_pointee = y->getType()->getContainedType(0);
+                if (x->getType() != dest_pointee) {
+                    x = builder.CreateBitCast(x, dest_pointee);
+                }
+            }
```
certik: I don't think we should be doing this cast here. Instead, let's figure out how to do the proper cast in ASR itself.

src/libasr/codegen/llvm_array_utils.cpp
```diff
@@ -9,6 +9,22 @@ namespace LCompilers {
...
+            if (arr->getType() != expected) {
+                if (arr->getType()->getNumContainedTypes() > 0 &&
+                        arr->getType()->getContainedType(0) == expected) {
+                    arr = llvm_utils->CreateLoad2(expected, arr);
+                } else {
+                    arr = builder->CreateBitCast(arr, expected);
+                }
+            }
+            return arr;
+        }
```
certik: I don't like this either --- why is the casting needed? Is it missing some ASR casts?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -24260,6 +24618,7 @@ Result<std::unique_ptr<LLVMModule>> asr_to_llvm(ASR::TranslationUnit_t &asr,
         Error error;
         return error;
     }
+    LLVM::fix_pointer_type_mismatches(*v.module);
```
certik: I would remove this.

certik:
I think the casting issues above have to be fixed first. After they are fixed, I'll do another review (I haven't seen anything major, but want to review again).

certik:
This wasn't fixed yet I think: https://github.com/lfortran/lfortran/pull/10989/changes#r3045892498.

src/libasr/codegen/llvm_utils.cpp
HarshitaKalani: When we remove this fix_pointer_type_mismatches, the athena compilation fails at the LLVM verification stage with 318 Stored value type does not match pointer operand type! errors and 34 Call parameter type does not match function signature! errors.

The root cause: the pass_array_by_data ASR pass transforms function signatures from descriptor-array arguments (%array.2.2*) to flat pass-by-data arguments (float*, i32*, i32*). However, procedure pointer types stored in struct components (e.g., procedure(iface), pass(this), pointer :: op_ptr) are not updated by this pass. When the codegen stores the transformed function pointer into the untransformed slot, the LLVM types mismatch.

certik: Yes, so pass_array_by_data must be fixed, we should not be casting it later.

certik:
@HarshitaKalani I think this was not fixed: https://github.com/lfortran/lfortran/pull/10989/changes#r3045892498, we should not be doing casts like that in the LLVM backend.

src/libasr/codegen/llvm_utils.cpp
HarshitaKalani: Yes, I'm trying to do that.

HarshitaKalani: The lfortran tests still pass after removing these changes but there are ~30 remaining pre-existing LLVM type mismatches from three separate root causes unrelated to pass_array_by_data which results in athena failure:

- Class/polymorphic type confusion (e.g., %diffstruc__types.array_type* stored into %array.2.9**)
- Finalizer function call type mismatches
- FixedSizeArray-to-pointer conversion mismatches ([2 x i32]* vs i32*)

These are codegen-level bugs that cannot be addressed in pass_array_by_data.  Let me try to fix them, else I'll remove them for this PR and then will take that as a separate PR.

certik: Yes, remove complex commits from this PR, let's get the rest in. And we'll focus on the complex fixes one at a time and figure out a clean fix for each.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1752,23 +1752,19 @@ namespace LCompilers {
...
-        // For struct types, the target type must match the pointer's pointee type
-        std::string target_type_str = LLVM::get_type_as_string(t);
-        std::string pointee_type_str = LLVM::get_type_as_string(ds_pointee_type);
-        LCOMPILERS_ASSERT_MSG(ds_pointee_type == t,
-            "Type mismatch in create_gep2: GEP target type does not match pointer's pointee type. "
-            "This would cause crashes in LLVM <= 8 constant folder. "
-            "Target type: " + target_type_str +
-            ", Pointer pointee type: " + pointee_type_str);
-
-        // Verify index is within bounds for struct types
```
certik, line 1771: The above two changes do not seem right.

Why was the assert wrong that was removed?

The logic above is adding a  bitcast `ds = builder->CreateBitCast(ds, t->getPointerTo());`, but we should not be silently adding a bitcast. Instead, we need to cast explicitly in ASR.

certik:
@HarshitaKalani I recommend you split this PR into simple commits that don't have casting issues and let's merge those.

Then create an MRE for the casting issue and I can help trying to fix it. So far any of the solutions above are not the right answer as they silently insert a cast in the LLVM backend one way or the other, in a general case. Instead, we need to have explicit casts in ASR and in the backend we can insert asserts to fail if the types do not agree.

HarshitaKalani:
Closing this.

PR #11016, opened 2026-04-09
fix: show user-facing allocatable LHS name in runtime error

Jatinagarwal24:
resolves
https://github.com/lfortran/lfortran/pull/10999#issuecomment-4210776244


HarshilShah1804 (approved):
Looks good. Thanks.

HarshitaKalani (approved):

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -203,6 +203,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        if (ASR::is_a<ASR::ExternalSymbol_t>(*sym)) {
+            ASR::ExternalSymbol_t* ext = ASR::down_cast<ASR::ExternalSymbol_t>(sym);
+            if (ext->m_original_name && ext->m_original_name[0] != '\0') {
+                return std::string(ext->m_original_name);
+            }
+            sym = ext->m_external;
+        }
+
+        return std::string(ASRUtils::symbol_name(sym));
+    }
```
certik, line 220: This whole function is very fragile and bloated:

* We should not silently exit when `sym` is nullptr. The function should error out if `sym` is a nullptr. Pass it by reference, to enforce this.
* Use `ASRUtils::symbol_get_past_external`

Is this function simply equivalent to `ASRUtils::symbol_name(ASRUtils::symbol_get_past_external(sym))`?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -203,6 +203,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        if (ASR::is_a<ASR::Var_t>(*expr)) {
+            ASR::Var_t* v = ASR::down_cast<ASR::Var_t>(expr);
+            return get_symbol_display_name(v->m_v);
+        }
+
+        if (ASR::is_a<ASR::StructInstanceMember_t>(*expr)) {
+            ASR::StructInstanceMember_t* sim = ASR::down_cast<ASR::StructInstanceMember_t>(expr);
+            std::string base_name = get_expr_name_for_runtime_message(sim->m_v);
+            std::string member_name = get_symbol_display_name(sim->m_m);
+            if (base_name.empty()) {
```
certik, line 236: When would this be empty?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -203,6 +203,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                return member_name;
+            }
+            return base_name + "%" + member_name;
+        }
+
+        return "";
+    }
```
certik, line 243: Is the logic of this function to only give a name for Var or StructInstanceMember, otherwise we should not give a name?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11403,11 +11442,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                {LLVMUtils::RuntimeLabel("'%s' not allocated here", {x.m_target->base.loc}, {target_expr_name_llvm})},
+                                infile,
+                                location_manager,
+                                target_expr_name_llvm);
+                        } else {
+                            llvm_utils->generate_runtime_error(is_not_allocated,
+                                "Array is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.",
+                                {LLVMUtils::RuntimeLabel("LHS not allocated here", {x.m_target->base.loc}, {})},
+                                infile,
+                                location_manager);
```
certik, line 11459: This is not tested.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11403,11 +11442,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                            location_manager);
+                        std::string target_expr_name = get_expr_name_for_runtime_message(x.m_target);
+                        if (!target_expr_name.empty()) {
+                            llvm::Value* target_expr_name_llvm = LCompilers::create_global_string_ptr(context, *module, *builder, target_expr_name);
+                            llvm_utils->generate_runtime_error(is_not_allocated,
+                                "Array '%s' is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.",
+                                {LLVMUtils::RuntimeLabel("'%s' not allocated here", {x.m_target->base.loc}, {target_expr_name_llvm})},
+                                infile,
+                                location_manager,
+                                target_expr_name_llvm);
```
certik, line 11453: Is this tested?

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: In pr #11024
Addressed by removing the extra symbol-display helper and simplifying logic.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: It will be empty for unsupported/non-Var/non-StructInstanceMember forms (and null), which triggers generic fallback.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: yes, that is exactly how it behaves now through pr #11024

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: yes

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: for this i have opened a pr #11041

PR #11031, opened 2026-04-10
fix: avoid segfault on unallocated allocatable array assignment

### LINKED ISSUE #10812: Bounds check does not trigger in types_24.f90
```console
$ lfortran integration_tests/types_24.f90
$ echo $?
11
$ lfortran integration_tests/types_24.f90 --realloc-lhs-arrays
Input:
1.00000000    2.00000000    3.00000000    4.00000000    5.00000000    6.00000000
Gradient reshaped back to 2D:
10.0000000    20.0000000    30.0000000    40.0000000    50.0000000    60.0000000
```

In the first run we should get a runtime error that something is out of bounds or not allocated.

jassem-manita:
Fixes #10812

This fixes a segfault when assigning to an unallocated allocatable array without '--realloc-lhs-arrays'.

The compiled program previously crashed instead of raising a runtime error. This pr adds a guard in the assignment path to check whether the target is an unallocated allocatable descriptor array and emits "not allocated" error.

- '--realloc-lhs-arrays' behavior is unchanged

This also matches the expected behavior for the related examples from #11014:
- `xalloc.f90`
- `xalloc_2d.f90`
- `xalloc_func.f90`


jassem-manita:
@certik do I need to add a test ?

certik:
Yes, first create an MRE, then add a test for it. Also you added two more runtime errors, so add a test for both too.

jassem-manita:
@certik I reduced the original case to an MRE and added it as a test for the named runtime error.

While trying to test the unnamed fallback I ran into a related case with `x(1)%arr = [1, 2, 3]` that still segfaults so I left it out.

```fortran
program allocatable_component_unnamed_unallocated_01
    implicit none

    type :: t
    end type t

    type(t) :: x(1)

    x(1)%arr = [1, 2, 3]
end program allocatable_component_unnamed_unallocated_01
```
```Console
lfortran fix-10812 ? ❯ lfortran tests/errors/allocatable_component_unnamed_unallocated_01.f90; echo $?
139
```


Should I handle it in this PR or in a follow-up?

jassem-manita:
This adds coverage for the unnamed runtime error case and fixes the segfault by checking for an unallocated allocatable descriptor target earlier in `visit_DebugCheckArrayBounds()`.
The later unallocated-error blocks in that function is removed since the early check now handles that case first.

jassem-manita:
@certik what do you think ?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11033,6 +11033,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        location_manager,
+                        target_name_llvm);
+                } else {
+                    llvm_utils->generate_runtime_error(
+                        is_not_allocated,
+                        "Array is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.",
+                        {LLVMUtils::RuntimeLabel("LHS not allocated here", {x.m_target->base.loc}, {})},
+                        infile,
+                        location_manager);
+                }
```
certik, line 11057: I don't like this split for two errors, now we have to test it twice (which you did), but it still adds complexity. Why cannot we simply print the actual expression every time? Then we only need to test it once.

Since this PR is a refactor of existing code, it's fine for now. But we should fix this in a subsequent PR.

integration_tests/allocatable_component_unnamed_unallocated_01.f90
```diff
@@ -0,0 +1,11 @@
...
+    implicit none
+
+    type :: t
+        integer, allocatable :: arr(:)
+    end type t
+
+    type(t) :: x(1)
+
+    x(1)%arr = [1, 2, 3]
+end program allocatable_component_unnamed_unallocated_01
```
certik, line 15: Add both tests into integration tests and test with `--realloc-lhs-arrays` there to ensure they work. Then keep the reference tests without `--realloc-lhs-arrays` for the error message.

certik (approved):
I think this is fine. The comments should be fixed, but that can be done in a subsequent PR.

src/libasr/codegen/asr_to_llvm.cpp
jassem-manita: > I don't like this split for two errors, now we have to test it twice (which you did), but it still adds complexity.

Agree on that. i will address it in the next PR.

> Why cannot we simply print the actual expression every time? Then we only need to test it once.

I followed the existing `get_expr_name_for_runtime_message()` .

integration_tests/allocatable_component_unnamed_unallocated_01.f90
certik: And now remove them from here, reference them directly from integration tests.

tests/tests.toml
```diff
@@ -5086,6 +5086,14 @@ filename = "errors/infer_realloc_disabled_01.f90"
 run = true
 options = "--infer"
 
+[[test]]
+filename = "errors/allocatable_component_reshape_unallocated_01.f90"
```
```suggestion
filename = "../integration_tests/allocatable_component_reshape_unallocated_01.f90"
```

tests/tests.toml
```diff
@@ -5086,6 +5086,14 @@ filename = "errors/infer_realloc_disabled_01.f90"
...
+run = true
+
+[[test]]
+filename = "errors/allocatable_component_unnamed_unallocated_01.f90"
```
certik: Same here. And update reference results.

jassem-manita: Done.

certik (approved):

Issue #11023, opened 2026-04-09
Simplify struct handling in GPU offloading


certik:
<div style="font-family: "Segoe UI", -apple-system, BlinkMacSystemFont, "Noto Sans", Helvetica, Arial, sans-serif; font-size: 16px; line-height: 1.5; color: #1f2328; word-wrap: break-word;">
<p style="margin: 0 0 16px;">The unified memory on Apple Silicon <em style="font-style: italic;">does</em> make a much simpler approach possible. The reason it&#39;s not done today is one specific constraint, but it&#39;s solvable. Let me explain.</p>
<h2 style="font-size: 1.5em; font-weight: 600; margin: 24px 0 16px; line-height: 1.25; color: #1f2328; border-bottom: 1px solid #d1d9e0b3; padding-bottom: 0.3em;">Why it doesn&#39;t &quot;just work&quot; today</h2>
<p style="margin: 0 0 16px;">The Fortran array descriptor (CFI-compatible) looks like this on the CPU:</p>
<div style="background-color: #f6f8fa; border-radius: 6px; padding: 16px; margin: 0 0 16px; overflow: auto;"><pre style="margin: 0; font-family: ui-monospace, SFMono-Regular, "SF Mono", Menlo, Consolas, "Liberation Mono", monospace; font-size: 85%; line-height: 1.45; color: #1f2328;"><code>struct descriptor {
    float*   base_addr;    // pointer to heap data
    int64_t  elem_len;
    int32_t  version;
    int8_t   rank, type, attribute, extra;
    int64_t  offset;
    struct { int64_t lower, extent, stride; } dim[N];
};</code></pre></div>
[...]

PR #11032, opened 2026-04-10
Implement more GPU Metal offloading

certik:
This is about as far as we can get with the current design. It offloads Fiats in the inference mode, but the design is not good enough to do training. For that we first need to refactor along the lines of
https://github.com/lfortran/lfortran/issues/11023.


certik:
Just LLVM 15 fails in fpm third-party test compilation.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -12198,6 +12198,65 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                llvm::Type *ptr_type = alloca_inst->getAllocatedType();
+                if (!ptr_type->isPointerTy()) continue;
+#if LLVM_VERSION_MAJOR >= 15
+                llvm::Type *elem_type =
+                    llvm_utils->get_type_from_ttype_t_util(
+                        nullptr, pointee_type, module.get());
+#else
+                llvm::Type *elem_type =
+                    ptr_type->getPointerElementType();
+#endif
```
certik, line 12252: We should refactor this ifdef later.

certik:
I am going to merge this, it wraps up the Metal offloading for inference. Now we setup fiats to be tested at the CI for this. And after that we need to refactor the backend.

PR #11033, opened 2026-04-10
CI: add macOS runner with LLVM 22 for third-party tests

certik:


certik:
Now we need to fix or skip third-party codes that fail on macOS, the first failure is in fortran_mpi.

PR #11034, opened 2026-04-10
fix: initialize `access_id` for FILE object with correct value

HarshilShah1804:
Towards #10677 

Splitting the large test present in #10677 to smaller tests, each checking for a particular connection state and access method.


Jatinagarwal24 (approved):
Looks good to me. Thanks !

PR #11037, opened 2026-04-10
Leak Fix

Jatinagarwal24:


Jatinagarwal24:
Hi I am aware of regression i will fix that.

certik:
Please resolve conflicts, then let's run all tests.

certik:
Can you explain how it works? It seems for save variables you now use a global descriptor, or something like that?

Jatinagarwal24:
> Can you explain how it works? It seems for save variables you now use a global descriptor, or something like that?

Yes, exactly: for class targets that are SAVE variables, the code now uses one module-global wrapper object (internal, zero-initialized) per symbol and reuses it, instead of heap-allocating wrapper storage each time; for non-SAVE variables, behavior is unchanged and it still does malloc plus memset.

assem2002 (approved):
Looks good. Thanks

PR #11038, opened 2026-04-10
CI : added featom build in the third party tests

Jatinagarwal24:


certik:
That looks good.

certik:
It added `2026-04-11T08:09:06.6160555Z + print_subsection '⏱ Duration: 107 seconds'` 107s, so I think this is ok for now. 

Now let's get the tests running, then we'll enable them (if they don't take too much time). And then we can run it for a while at the CI. 

We can always remove it later once we have other codes that will be higher priority for the CI time.

PR #11036, opened 2026-04-10
feat: implement `position` and `action` specifier in inquire

HarshilShah1804:
Towards #10677

PR #11044, opened 2026-04-11
feat: support `delim` specifier in file I/O operations

HarshilShah1804:
After #11036 is merged, we need to rebase this and then merge.

### DIFF
--- a/integration_tests/file_62.f90
+++ b/integration_tests/file_62.f90
@@ -1,0 +1,38 @@
+program file_62
+    implicit none
+
+    integer :: unit
+    character(len=20) :: delim_val
+    character(len=*), parameter :: fname = "file_62_data.txt"
+
+    unit = 10
+
+    open(unit, file=fname, status="replace", &
+         form="formatted", delim="apostrophe")
+
+    inquire(unit=unit, delim=delim_val)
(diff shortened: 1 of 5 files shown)


integration_tests/file_62.f90
```diff
@@ -0,0 +1,90 @@
...
+  call pf (iostat == 0)
+  call pf (access == 'UNDEFINED')
+  call pf (action == 'UNDEFINED')
+  call pf (blank  == 'UNDEFINED')
+  call pf (delim  == 'UNDEFINED')
+  call pf (direct == 'UNKNOWN')
+  call pf (.not. exist)
+  call pf (form   == 'UNDEFINED')
+  call pf (formatted == 'UNKNOWN')
+  ! call pf (name   == 'UNDEFINED')
```
HarshilShah1804: Every compiler that I checked on godbolt, returned the value present in the `lfn` variable instead of "UNDEFINED". But as per, the original test in #10677, it should have been "UNDEFINED". Also, as per the documentation that I checked, I found that if a file name is provided by the `file` specifier, the `name` specifier takes the same value.

@certik Please share your thought on this.

certik (approved):

PR #11046, opened 2026-04-11
Some stdlib fixes

certik:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14533,8 +14548,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             return;
         }
         this->visit_expr_wrapper(x.m_arg, true);
-        ASR::ttype_t* curr_type = extract_ttype_t_from_expr(x.m_arg);
-        int arg_kind = ASRUtils::extract_kind_from_ttype_t(curr_type);
+        int arg_kind = ASRUtils::extract_kind_from_ttype_t(x.m_type);
+        unwrap_polymorphic_complex_arg(x.m_arg, arg_kind);
```
certik: I think that this is actually missing a cast in ASR, so this fix should not be needed.

PR #11051, opened 2026-04-11
fix: finalize and free class(*) payload memory during allocatable cleanup

Jatinagarwal24:


src/libasr/pass/insert_deallocate.cpp
```diff
@@ -291,6 +291,58 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
+
+            allocatable_vars.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(
+                al, var->base.base.loc, sym)));
+        }
+
+        if (allocatable_vars.size() > 0) {
+            ASR::stmt_t *implicit_deallocation_node = ASRUtils::STMT(
+                ASR::make_ImplicitDeallocate_t(
+                    al, xx.base.base.loc,
+                    allocatable_vars.p, allocatable_vars.size()));
```
certik: I think we don't want to insert deallocate into the main program. Rather, in LLVM when things go out of scope we want to deallocate, and in the main program only when we are detecting leaks.

CC @assem2002

assem2002: Agree. Any work that's considered as variable finalization work should be done in the backend.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2663,6 +2663,41 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                     llvm::Value *cond = arr_descr->get_is_allocated_flag(tmp, tmp_expr);
                     llvm_utils->create_if_else(cond, [=]() {
                         llvm_symtab_finalizer.finalize_before_deallocate(tmp, cur_type, struct_sym, in_struct);
+
+                        // For class(*) arrays, descriptor base_addr points to a
+                        // wrapper whose field 1 owns the actual payload buffer.
+                        // Free payload first, then free wrapper via call_lfortran_free.
```
assem2002: Shouldn't `llvm_symtab_finalizer.finalize_before_deallocate()` do the work? it should use the code you implemented in file `llvm_utils.h`

Jatinagarwal24:
Hi @assem2002 @certik pls review this i have made suggested changes.
Thanks !

src/libasr/codegen/llvm_utils.h
```diff
@@ -1639,21 +1642,44 @@ class ASRToLLVMVisitor;
...
-                // Class => {VTable* , underlying_struct*}
-                // Array of non polymorphic classes is speical handled.
-                // Array doesn't allocate consecutive classes, It allocates only one (on stack)
-                // and inserts consecutive structs (heap).
-                // `free_array_data` handles that clean up.
-                return;
-            }
-            
-            llvm_utils_->lfortran_free_nocheck(ptr);
+        void free_array_ptr_to_consecutive_data(llvm::Value* const ptr, ASR::ttype_t* const t, [[maybe_unused]] ASR::Struct_t* const struct_sym){
```
certik: What is the purpose of this `struct_sym`, it doesn't seem to be used inside this function?

src/libasr/codegen/llvm_utils.h
```diff
@@ -2142,6 +2168,40 @@ class ASRToLLVMVisitor;
...
+                                payload_ptr,
+                                llvm::ConstantPointerNull::get(llvm_utils_->i8_ptr));
+                            llvm_utils_->create_if_else(payload_not_null, [this, payload_ptr, payload_ref]() {
+                                llvm_utils_->lfortran_free_nocheck(payload_ptr);
+                                builder_->CreateStore(
+                                    llvm::ConstantPointerNull::get(llvm_utils_->i8_ptr), payload_ref);
+                            }, [](){});
+                        }, [](){});
+                        return;
+                    }
```
certik, line 2203: @assem2002 is this the right place to free the wrapper? This function does finalization, but I don't know if we should also free here as part of the process, or if we should do it elsewhere.

src/libasr/codegen/llvm_utils.h
Jatinagarwal24: Hi @certik Thanks for your suggestion  I have removed the struct_sym.

src/libasr/codegen/llvm_utils.h
```diff
@@ -1640,20 +1643,43 @@ class ASRToLLVMVisitor;
...
-                // and inserts consecutive structs (heap).
-                // `free_array_data` handles that clean up.
-                return;
-            }
-            
-            llvm_utils_->lfortran_free_nocheck(ptr);
+            llvm::Value* const ptr_not_null = builder_->CreateICmpNE(
+                builder_->CreatePtrToInt(ptr, llvm::Type::getInt64Ty(builder_->getContext())),
+                llvm::ConstantInt::get(llvm::Type::getInt64Ty(builder_->getContext()), 0));
+            llvm_utils_->create_if_else(ptr_not_null, [this, ptr, t]() {
```
assem2002, line 1649: Please move this if_else check to where it actually needs to be used.
i.e. in the check below 
```cpp
else if (ASRUtils::is_class_type(t) && ASRUtils::is_unlimited_polymorphic_type(t))
```

src/libasr/codegen/llvm_utils.h
assem2002: Yes, Your doubts are true. The PR already implements this mechanism ( as part of `finalize_array()` ).
BUT,
We still need this code in this spot, As `finalize_before_deallocate` doesn't go through `finalize_array()` it jumps to internal pieces making it not go through the already implemented mechanism .

Bottom Line : Not the right spot because `finalize_before_deallocate` is already not a good function to use. We'll iterate on that later (`visit_Deallocate()` will call the finalization process rightaway with no additional code)

assem2002 (approved):
I think it's fine. Thanks.
Please address my comment in another small PR.
We'll can iterate later to clean up things.

PR #11059, opened 2026-04-13
fix: handle access of private type bound procedures correctly

### LINKED ISSUE #11053: `private` attribute of type-bound procedure is ignored
The following code produces a compiler error message if one tries to compile it with flang-20, because it attempts to call a `private` type-bound procedure outside the scope of its hosting module:
`$ cat access.f90 `
```
module typemod

   type :: MyType
   contains
      procedure, private, nopass :: inaccessible
   end type MyType
   
contains

   subroutine inaccessible()
   end subroutine inaccessible

end module typemod

program test
   
   use typemod
   
   obj = MyType()   
   call obj%inaccessible()
   
end program test
```
LFortran erroneously compiles and runs this code without complaint.

opixdown:
fixes #11053

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -7016,6 +7016,25 @@ public:
                 break;
             }
             case (ASR::symbolType::StructMethodDeclaration) : {
+                {
+                    ASR::StructMethodDeclaration_t* smd = ASR::down_cast<ASR::StructMethodDeclaration_t>(original_sym);
+                    if (smd->m_is_private) {
+                        ASR::symbol_t* struct_sym = ASR::down_cast<ASR::symbol_t>(smd->m_parent_symtab->asr_owner);
+                        ASR::symbol_t* defining_module = struct_sym ? ASRUtils::get_asr_owner(struct_sym) : nullptr;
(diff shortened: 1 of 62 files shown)


opixdown:
I added is_private in StructMethodDeclaration, similar to is_nopass and is_deferred Is this the correct approach, or should I follow some other approach?

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7073,6 +7092,21 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        bool caller_is_same_module = (current_module != nullptr && defining_module != nullptr &&
+                            ASR::is_a<ASR::Module_t>(*defining_module) && current_module == ASR::down_cast<ASR::Module_t>(defining_module));
+                        if (!caller_is_same_module) {
+                            diag.add(Diagnostic(
+                                "Type-bound procedure '" + std::string(class_proc->m_name)
+                                    + "' is PRIVATE and inaccessible outside its defining module",
+                                Level::Error, Stage::Semantic, {
+                                    Label("", {x.base.base.loc})
+                                }));
+                            throw SemanticAbort();
```
certik, line 7107: Add tests for both error messages into continue_compilation_1.f90

Issue #10886, opened 2026-03-28
Compile M_io


HarshilShah1804:
Package Link: https://github.com/urbanjost/M_io

Current Compilation Status:
```bash
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/M_io$ fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
M_io.F90                               failed.
[  1%] Compiling...


semantic error: Invalid argument `stream` supplied
   --> ././src/M_io.F90:443:10 - 454:41
    |
443 |             inquire(unit=lun,                                                                               &
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
    |
454 |         &   iostat=ios,err=999,iomsg=message)
    | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
<ERROR> Compilation failed for object " src_M_io.F90.o "
<ERROR> stopping due to failed compilation
STOP 1

```

HarshilShah1804:
After #11683, #11726, #11752, the package compiles and all tests run successfully.

PR #10977, opened 2026-04-05
doc: add documentation for newly inmplemented specifiers for file I/O statements

HarshilShah1804:

PR #11064, opened 2026-04-13
feat: support `iomsg` specifier for `rewind`

HarshilShah1804:
Towards #10886

### DIFF
--- a/integration_tests/rewind_01.f90
+++ b/integration_tests/rewind_01.f90
@@ -1,0 +1,21 @@
+program rewind_01
+    implicit none
+    integer :: ios
+    integer, parameter :: u = 10
+    character(len=128) :: msg
+
+    open(unit=u, file="rewind_01_file.txt", status="replace", access="direct", recl=4, iostat=ios)
+    if (ios /= 0) error stop
+
+    msg = "sentinel"
+    rewind(unit=u, iostat=ios, iomsg=msg)
+
+    if (ios == 0) error stop
+#if defined(__LFORTRAN__)
+    if (len_trim(msg) == 0) error stop
+    if (trim(msg) /= "REWIND cannot be used on UNIT 10 opened with DIRECT access.") error stop
+#endif
+    close(unit=u, status="delete")
+
+
+end program rewind_01
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -57,7 +57,7 @@ stmt
     | FileClose(int label, expr? unit, expr? iostat, expr? iomsg, expr? err, expr? status)
     | FileRead(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? advance, expr? size, expr? id, expr? pos, expr* values, stmt? overloaded, bool is_formatted, symbol? nml, expr? rec, expr? pad)
     | FileBackspace(int label, expr? unit, expr? iostat, expr? err)
-    | FileRewind(int label, expr? unit, expr? iostat, expr? err)
+    | FileRewind(int label, expr? unit, expr? iostat, expr? err, expr? iomsg)
     | FileEndfile(int label, expr? unit, expr? iostat, expr? err)
     | FileInquire(int label, expr? unit, expr? file, expr? iostat, expr? err, expr? exist, expr? opened, expr? number, expr? named, expr? name, expr? access, expr? sequential, expr? direct, expr? form, expr? formatted, expr? unformatted, expr? recl, expr? nextrec, expr? blank, expr? position, expr? action, expr? read, expr? write, expr? readwrite, expr? delim, expr? pad, expr? flen, expr? blocksize, expr? convert, expr? carriagecontrol, expr? size, expr? pos, expr? iolength, expr* iolength_vars, expr? decimal, expr? sign, expr? encoding, expr? stream, expr? iomsg, expr? round, expr? pending, expr? asynchronous)
     | FileWrite(int label, expr? unit, expr? iomsg, expr? iostat, expr? id, expr* values, expr? separator, expr? end, stmt? overloaded, bool is_formatted, symbol? nml, expr? rec, expr? pos)
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5997,8 +5997,8 @@ LFORTRAN_API void _lfortran_flush(int32_t unit_num, int32_t* iostat, char* iomsg
                 }
                 return;
             } else {
-                printf("Specified UNIT %d in FLUSH is not connected.\n", unit_num);
-                exit(1);
+            printf("Specified UNIT %d in FLUSH is not connected.\n", unit_num);
+            exit(1);
             }
         }
         fflush(filep);
@@ -6730,13 +6730,44 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
     }
 }
 
-LFORTRAN_API void _lfortran_rewind(int32_t unit_num)
+LFORTRAN_API void _lfortran_rewind(int32_t unit_num, int32_t* iostat, char* iomsg, int64_t iomsg_len)
 {
+    if (iostat != NULL) {
+        *iostat = 0;
+    }
     bool unit_file_bin;
-    FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+    int access_id = -1;
+    FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin, &access_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
     if( filep == NULL ) {
         printf("Specified UNIT %d in REWIND is not created or connected.\n", unit_num);
         exit(1);
+    }
+    if (access_id == 2) {
+        if (iostat != NULL) {
+            *iostat = 5002;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char msg[100];
+                snprintf(msg, sizeof(msg), "REWIND cannot be used on UNIT %d opened with DIRECT access.", unit_num);
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND cannot be used on UNIT %d opened with DIRECT access.\n", unit_num);
+            exit(1);
+        }
+    }
+    if (fseek(filep, 0, SEEK_SET) != 0) {
+        if (iostat != NULL) {
+            *iostat = -1;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char *msg = strerror(errno);
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND failed for UNIT %d.\n", unit_num);
+            exit(1);
+        }
     }
     rewind(filep);
     seq_char_state_reset(unit_num);
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3032,6 +3032,7 @@ RUN(NAME loop_var_use_after_loop LABELS gfortran llvm llvm_wasm llvm_wasm_emcc w
 RUN(NAME sign_from_value LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 
 RUN(NAME rewind_inquire_flush LABELS gfortran)
+RUN(NAME rewind_01 LABELS gfortran llvm EXTRA_ARGS --cpp GFORTRAN_ARGS -cpp)
 RUN(NAME flush_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc COPY_TO_BIN file_01_data.txt)
 RUN(NAME flush_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc COPY_TO_BIN file_01_data.txt file_02_data.txt)
 RUN(NAME flush_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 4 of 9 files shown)


Jatinagarwal24 (commented):
Total Test time (real) =   5.57 sec

The following tests FAILED:
Errors while running CTest
	2377 - rewind_01 (Failed)

tests/reference/run-rewind_01-4e7b212.stdout
```diff
@@ -0,0 +1 @@
+Cannot REWIND a file opened for DIRECT access
```
certik: I am confused by this file --- the reference test says it passes, but it prints this? The integration test also seems to pass.

Is this a bug, or is this file a left over?

HarshilShah1804: Yes, it is passes because we have iostat specifier. So, the exit code is not 0. The reference has just been added to capture the error message that we got from the iomsg. If you feel, this is not required, I can remove it.

integration_tests/rewind_01.f90
```diff
@@ -0,0 +1,19 @@
...
+
+    open(unit=u, file="rewind_01_file.txt", status="replace", access="direct", recl=4, iostat=ios)
+    if (ios /= 0) error stop
+
+    msg = "sentinel"
+    rewind(unit=u, iostat=ios, iomsg=msg)
+
+    if (ios == 0) error stop
+    if (len_trim(msg) == 0) error stop
+    print *, trim(msg)
```
certik: I see ---- we are testing this line using reference tests, but not integration tests.

Instead, test it directly here via `if (trim(msg) /= "...") error stop`. That way it will be tested by all LLVM versions and we can remove the now redundant integration test run.

tests/reference/run-rewind_01-4e7b212.stdout
certik: I see now. In that case here is how to do this: https://github.com/lfortran/lfortran/pull/11064/changes#r3080925282.

integration_tests/rewind_01.f90
HarshilShah1804: Okay. The only reason of doing it using test run was that different compilers would give different error. So, the fortran code might fail with other compilers.

certik: Ok, so in that case use preprocessor macros to test the exact error for LFortran, but for other compilers skip the actual error message.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6730,15 +6730,53 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+    }
     bool unit_file_bin;
-    FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+    int access_id = -1;
+    FILE* filep = get_file_pointer_from_unit(unit_num, &unit_file_bin, &access_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
     if( filep == NULL ) {
-        printf("Specified UNIT %d in REWIND is not created or connected.\n", unit_num);
-        exit(1);
+        if (iostat != NULL) {
+            *iostat = -5005;
```
kmr-srbh: We need to be careful here. Apart from a few values, the Fortran standard does not provide standard status codes for errors. Runtime errors have compiler dependent exit codes. LFortran can have its own status code constants and can use it at appropriate places. I do not mean that the change is incorrect, just that it is non-standard.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6730,15 +6730,53 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+            if (iomsg != NULL && iomsg_len > 0) {
+                char *msg = "Specified UNIT is not connected.";
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("Specified UNIT %d in REWIND is not created or connected.\n", unit_num);
+            exit(1);
```
kmr-srbh: I think the error message and exit code should be same for an error regardless of the `IOSTAT` specifier.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6730,15 +6730,53 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+        if (iostat != NULL) {
+            *iostat = 5002;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char *msg = "Cannot REWIND a file opened for DIRECT access";
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND cannot be used on UNIT %d opened with DIRECT access.\n", unit_num);
+            exit(1);
```
kmr-srbh, line 6756: Same as https://github.com/lfortran/lfortran/pull/11064/changes#r3089420044.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6730,15 +6730,53 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+        if (iostat != NULL) {
+            *iostat = -1;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char *msg = strerror(errno);
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND failed for UNIT %d.\n", unit_num);
+            exit(1);
```
kmr-srbh, line 6769: Same as https://github.com/lfortran/lfortran/pull/11064/changes#r3089420044.

kmr-srbh (commented):
Otherwise looks good.

src/libasr/runtime/lfortran_intrinsics.c
HarshilShah1804: I think exit code would not be the same. In case where, iostat is present we don't exit with non-zero exit code.

src/libasr/runtime/lfortran_intrinsics.c
certik: Can you explain how this change is non-standard? It seems this is the right change that we should do?

src/libasr/runtime/lfortran_intrinsics.c
certik: I think it's fine to exit with exit code 1 always on error. But the error message I think should be the same for both cases, no?

src/libasr/ASR.asdl
```diff
@@ -57,7 +57,7 @@ stmt
     | FileClose(int label, expr? unit, expr? iostat, expr? iomsg, expr? err, expr? status)
     | FileRead(int label, expr? unit, expr? fmt, expr? iomsg, expr? iostat, expr? advance, expr? size, expr? id, expr? pos, expr* values, stmt? overloaded, bool is_formatted, symbol? nml, expr? rec, expr? pad)
     | FileBackspace(int label, expr? unit, expr? iostat, expr? err)
-    | FileRewind(int label, expr? unit, expr? iostat, expr? err)
+    | FileRewind(int label, expr? unit, expr? iostat, expr? err, expr? iomsg)
```
certik, line 60: Document in doc/

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: I see, when `IOSTAT` is not specified, the exit status is that of a regular runtime error instead of an I/O one:

```console
(base) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
clang: warning: overriding deployment version from '16.0' to '26.0' [-Woverriding-deployment-version]
At line 11 of file test.f90 (unit = 10, file = 'rewind_01_file.txt')
Fortran runtime error: Cannot REWIND a file opened for DIRECT access

Error termination. Backtrace:
#0  0x10256a103
#1  0x10256acd7
#2  0x10256b5a7
#3  0x102674a33
#4  0x10214c75f
#5  0x10214c82b
(base) saurabhkumar@Mac lfortran % echo $?
2
(base) saurabhkumar@Mac lfortran % flang test.f90 && ./a.out   

fatal Fortran runtime error(/Users/saurabhkumar/OpenSource/lfortran/test.f90:11): REWIND(UNIT=10) on non-sequential file
zsh: abort      ./a.out
(base) saurabhkumar@Mac lfortran % echo $?
134
```
So I think the exit code is correct.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: Quoting from the standard:

> 12.11.5 IOSTAT= specifier
1 Execution of an input/output statement containing the IOSTAT= specifier causes the stat-variable in the IO-
STAT= specifier to become defined with
• a zero value if neither an error condition, an end-of-file condition, nor an end-of-record condition occurs,
• the processor-dependent positive integer value of the constant IOSTAT_INQUIRE_INTERNAL_UNIT
from the intrinsic module ISO_FORTRAN_ENV (16.10.2) if a unit number in an INQUIRE statement
identifies an internal file,
• a processor-dependent positive integer value different from IOSTAT_INQUIRE_INTERNAL_UNIT if any
other error condition occurs,
• the processor-dependent negative integer value of the constant IOSTAT_END(16.10.2.16) from the intrinsic
module ISO_FORTRAN_ENV if an end-of-file condition occurs and no error condition occurs,
• the processor-dependent negative integer value of the constant IOSTAT_EOR(16.10.2.17) from the intrinsic
module ISO_FORTRAN_ENV if an end-of-record condition occurs and no error condition or end-of-file
condition occurs, or
• a processor-dependent negative integer value different from IOSTAT_EOR and IOSTAT_END, if the IO-
STAT= specifier appears in a FLUSH statement and the processor does not support the flush operation for
the specified unit.

The non-standard compiler dependent behaviour is due to the 3rd point. Different compilers have different error codes for the various I/O errors that can occur. They even provide usage guides for them, like:
* Oracle - https://docs.oracle.com/cd/E19957-01/805-4941/6j4m2soba/index.html
* IBM - https://www.ibm.com/docs/en/openxl-fortran-aix/17.1.4?topic=inputoutput-conditions-iostat-values

kmr-srbh: For the test added in the PR, here is the difference in `IOSTAT` values between `gfortran` and `flang`:

```console
(base) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
clang: warning: overriding deployment version from '16.0' to '26.0' [-Woverriding-deployment-version]
 IOSTAT:         5002
(base) saurabhkumar@Mac lfortran % flang test.f90 && ./a.out   
 IOSTAT:  1016
```

certik: Yes. But can you explain why this code in this PR is thus not correct? Each compiler chooses some codes, so we chose the code above.

certik:
@HarshilShah1804 let's wrap this up.

src/libasr/runtime/lfortran_intrinsics.c
HarshilShah1804: I have updated the error message to be same as there in case of exit with non-zero exit code.

src/libasr/ASR.asdl
HarshilShah1804: It is done already in #10977

HarshilShah1804:
@certik I have made the required change for updating the iomsg to be same as that in the non-zero exit code. Documentation was already done in #10977. I think this is good to merge.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: > Yes. But can you explain why this code in this PR is thus not correct? Each compiler chooses some codes, so we chose the code above.

No no, I did not mean to say that it is incorrect. Just that it is non-standard and the code is that of GFortran. We will need to make sure we stay consistent if we choose to follow GFortran codes.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6740,17 +6730,47 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+        if (iostat != NULL) {
+            *iostat = 5002;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char msg[100];
+                snprintf(msg, sizeof(msg), "REWIND cannot be used on UNIT %d opened with DIRECT access.", unit_num);
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND cannot be used on UNIT %d opened with DIRECT access.\n", unit_num);
```
certik, line 6755: You can assign the message to a variable and reuse at both places, right?

Also this `else` branch is not tested. Add a reference "run" test to test it that it works.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6740,17 +6730,47 @@ LFORTRAN_API void _lfortran_inquire(const fchar* f_name_data, int64_t f_name_len
...
+    if (fseek(filep, 0, SEEK_SET) != 0) {
+        if (iostat != NULL) {
+            *iostat = -1;
+            if (iomsg != NULL && iomsg_len > 0) {
+                char *msg = strerror(errno);
+                _lfortran_copy_str_and_pad(iomsg, iomsg_len, msg, strlen(msg));
+            }
+            return;
+        } else {
+            printf("REWIND failed for UNIT %d.\n", unit_num);
```
certik, line 6768: The logic to create the message should be reused for both.

The `else` branch is not tested, add a reference "run" test.

certik (approved):
I think it's good enough to merge, but more work is needed to finish it. @HarshilShah1804 do you mind please fixing the two comments in a subsequent PR? I made https://github.com/lfortran/lfortran/issues/11121 for that.

PR #11002, opened 2026-04-08
Memory-leak(Initial impl) : Create finalize fn for upoly types and register in type's VTable

assem2002:
### Goal of this PR 
**Finalize -> `class(*), allocatable :: x`.**
Start handling the case of finalizing an unlimited polymorphic type (the runtime type, not the descriptors).
**HOW :** We'll use similar approach to how we allocate and copy upoly types at runtime, by using functions.
We'll register the finalization type into each type VTable (e.g. `i32_VTable = [5 x i8*] = { .. , .. , copy , allocate , finalize_fn}`)
and call it at finalization. 
***
### What's in this PR
- Register initial finalization functions into type's VTable.
- Call that empty function at finalizatio process
***
### What's next
properly insert the finalization instructions into the finalization functions we registered. and making sure tests pass.

PR #11073, opened 2026-04-14
Leak fix

Jatinagarwal24:


src/libasr/codegen/llvm_utils.h
```diff
@@ -1154,7 +1156,32 @@ class ASRToLLVMVisitor;
                                     llvm_utils_->create_if_else(char_not_null,
                                         [this, char_ptr]() { llvm_utils_->lfortran_free_nocheck(char_ptr); },
                                         [](){});
-                                }, [](){});
+                                }, [this, vptr, data_ptr]() {
+                                    // For class(*) payloads with concrete struct dynamic type,
+                                    // run the concrete struct finalizer before freeing payload.
+                                    for (auto &vtab_entry : llvm_utils_->get_known_class_vtables()) {
```
assem2002, line 1162: It's checking against every Vtable that exists (it's done at runtime) -- I think it's still a valid approach to finalize Upoly types by checking against all VTables (if cached instruction, otherwise we'll end up with massive binary file).
But It's faster to use finalization functions that we insert into Vtable

assem2002: It's implemented in here https://github.com/lfortran/lfortran/pull/11002

Jatinagarwal24: So with pr #11002 are no leaks found for test class_allocate_02.f90
If yes then  we can merge #11002 first and then with this we can only remove the NO_DETECT_LEAK label .

assem2002: The approach here is using runtime check through almost every fortran type (almost), That's invalid approach to use.
Instead of comparing VTables at runtime, We'll have the finalization function already registered in the Vtable and just call it ( You can call it runtime polymorphic finalization).

PR #11078, opened 2026-04-14
fix: free class(*) character temporary buffers in LLVM codegen

Jatinagarwal24:


certik:
Fine with me for now.

PR #11081, opened 2026-04-14
feat: Implement writting of namelist into internal string array by creating a new LFortran Intrinsic API function

### LINKED ISSUE #10905: ICE when writing namelist to internal character array
When building the `M_intrinsics` with LFortran compilation fails with ice  The issue is triggered by a test that writes a namelist to an internal character array.

### MRE
```
module m
implicit none
  integer :: x = 123
end module m

program test
use m, y => x
implicit none
character(len=80) :: page(4)
namelist /nl/ y
write(page, nl)
end program test

```
### Current Behaviour
```
lf) opixdown@Atharvs-MacBook-Air M_intrinsics % lfortran t.f90 
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 2809
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 2736
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 1250
  File "/Users/opixdown/lfortran/src/lfortran/fortran_evaluator.cpp", line 407
    if (res.ok) {
  File "/Users/opixdown/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 22496
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 364
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 338
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 365
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 22
  File "/Users/opixdown/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 1703
    visit_symbol(*item.second);
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 367
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 30
  File "/Users/opixdown/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 5725
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 384
  File "/Users/opixdown/lfortran/src/libasr/asr_base_visitor.h", line 80
    case stmtType::FileInquire: { v.visit_FileInquire((const FileInquire_t &)x); return; }
  File "/Users/opixdown/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 17676
    unit_val = llvm_utils->convert_kind(tmp, llvm::Type::getInt32Ty(context));
  File "/Users/opixdown/lfortran/src/libasr/codegen/llvm_utils.cpp", line 3780
    LCOMPILERS_ASSERT(
AssertFailed: (val->getType()->isIntegerTy() && target_type->isIntegerTy()) || (val->getType()->isFloatingPointTy() && target_type->isFloatingPointTy())
```

AnshMehta1:
Fix #10905

### DIFF
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -11443,28 +11443,82 @@ LFORTRAN_API int _lfortran_exec_command(fchar *cmd, int64_t len) {
 // Namelist I/O Support
 // ============================================================================
 
+typedef struct {
+    FILE *fp;
+    char *data;
+    int64_t elem_len;
+    int64_t n_elems;
+    int64_t elem_idx;
+    int64_t pos;
+    bool is_file;
+} nml_writer_t;
+
+static void write_char(nml_writer_t *w, char c) {
+    if (w->is_file) {
+        fputc(c, w->fp);
+        return;
+    }
+
+    if (w->elem_idx >= w->n_elems) return;
+
+    char *dest = w->data + w->elem_idx * w->elem_len;
+
+    if (c == '\n') {
+        while (w->pos < w->elem_len) {
+            dest[w->pos++] = ' ';
+        }
+        w->elem_idx++;
+        w->pos = 0;
+        return;
+    }
+
+    if (w->pos < w->elem_len) {
+        dest[w->pos++] = c;
+    }
+}
+
+static void write_str(nml_writer_t *w, const char *s) {
+    while (*s) {
+        write_char(w, *s++);
+    }
+}
+
 // Helper function to write a single namelist item value
-static void write_nml_value(FILE *fp, const lfortran_nml_item_t *item, int64_t offset) {
+static void write_nml_value(nml_writer_t *w, const lfortran_nml_item_t *item, int64_t offset) {
+    char buf[128];
     void *ptr = (char*)item->data + offset;
+
     switch (item->type) {
         case LFORTRAN_NML_INT1:
-            fprintf(fp, "%d", *(int8_t*)ptr);
+            snprintf(buf, sizeof(buf), "%d", *(int8_t*)ptr);
+            write_str(w, buf);
             break;
+
         case LFORTRAN_NML_INT2:
-            fprintf(fp, "%d", *(int16_t*)ptr);
+            snprintf(buf, sizeof(buf), "%d", *(int16_t*)ptr);
+            write_str(w, buf);
             break;
+
         case LFORTRAN_NML_INT4:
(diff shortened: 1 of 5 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -11474,6 +11483,86 @@ static void write_nml_value(FILE *fp, const lfortran_nml_item_t *item, int64_t o
...
+                }
+            }
+
+            if (pos < max_len) out[pos++] = '\'';
+            if (pos < max_len) out[pos] = '\0';
+
+            break;
+        }
+    }
+}
```
certik, line 11567: So of all these cases only one is tested so far, correct?

certik:
Is this using the same mechanism as writing to a file, or do we now have to different writing mechanism in our runtime, one for files, the other for internal writes to strings?

AnshMehta1:
> Is this using the same mechanism as writing to a file, or do we now have to different writing mechanism in our runtime, one for files, the other for internal writes to strings?

It uses different writting mechanism for internal writes to strings. For namelist read, we followed the similar structure due to which I generalised things in a new function in https://github.com/lfortran/lfortran/pull/11081#discussion_r3088382694.

certik:
Why cannot the two use cases be unified? What is preventing it? We should not ideally maintain two separate parsers and printers which are identical except where they write to.

certik:
I don't like this, let's figure out how to reuse the same logic for writing into files as well as strings, the same with reading.

AnshMehta1:
Ok. Firstly, I will unify things in this PR. In a subsequent PR, I will modify things for read. I guess this should work.

AnshMehta1:
I unified write mechanism for file and array of strings. But still, we will need 2 different Lfortran API functions for them due to difference in arguments. Have implemented things accordingly in asr to llvm file.

certik (approved):
I think this is much better. It's ok if the highest-level API is different (if needed, or even just temporarily) as long as the actual implementation is reused for both, which I think now it is. That's important because the mechanism of both writing and reading the namelist can be complicated, so we only want to maintain one version and use and test it for both.

AnshMehta1:
> I think this is much better. It's ok if the highest-level API is different (if needed, or even just temporarily) as long as the actual implementation is reused for both, which I think now it is. That's important because the mechanism of both writing and reading the namelist can be complicated, so we only want to maintain one version and use and test it for both.

Yep, I definitely agree with this point. Will see if reading part needs unification and work on it next week once my exams are completed.

PR #11087, opened 2026-04-14
Fix nested associate variable comparison in codegen

### LINKED ISSUE #11085: Nested associate type bug
Program:
```fortran
! nested_associate.f90
associate(pi => 4.0*atan(1.0))
    associate(d => sin(pi/6))   ! sin(pi/6) = 0.5
        associate(g => d)
            print *, g, d, g == d
        end associate
    end associate
end associate
end
```
Compiler Explorer: https://godbolt.org/z/EdYbWqGzf

Error with LFortran 0.59:
```
Both operands to FCmp instruction are not of the same type!
  %9 = fcmp oeq float* %7, float %8
```

If I remove the expression `g == d`, the program works correctly.

certik:
When a variable associated with another associate variable (e.g., associate(g => d) where d is also an associate) was used in a comparison or arithmetic expression inside a print statement, the LLVM codegen produced a type mismatch. The pointer-typed associate variable was only dereferenced once (to float*) instead of twice (to float), causing 'fcmp oeq float* %x, float %y' errors.

The root cause was that several comparison and binary operation visitors (RealCompare, UnsignedIntegerCompare, ComplexCompare, LogicalCompare, LogicalBinOp, RealBinOp) used visit_expr_wrapper without explicitly setting ptr_loads, inheriting an insufficient value from parent contexts like StringFormat.

Fixed by following the existing IntegerCompare/IntegerBinOp pattern: use visit_expr_load_wrapper with appropriate ptr_loads based on whether each operand is a pointer type, and add the load_non_array_non_character_pointers safety net for struct members.

An integration test is added.

Fixes #11085.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2498,6 +2498,7 @@ RUN(NAME associate_46 LABELS gfortran llvm)
 RUN(NAME associate_47 LABELS gfortran llvm)
 RUN(NAME associate_48 LABELS gfortran llvm NO_DETECT_LEAK) # USER_LEAK
 RUN(NAME associate_49 LABELS gfortran llvm NO_DETECT_LEAK) # USER_LEAK
+RUN(NAME associate_50 LABELS gfortran llvm)
 
 RUN(NAME attr_dim_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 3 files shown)


certik:
LLVM 21 in Ubuntu gives:
```
  %44 = load ptr, ptr %g, align 8
 float  %46 = call float @llvm.fma.f32(ptr %44, float %45, float -2.500000e-01)
```
Must fix it.

PR #11091, opened 2026-04-15
Fix: Optional value dummy argument in bind(c) procedure cause ICE

### LINKED ISSUE #11080: Optional value dummy argument in bind(c) procedure
Test program (thanks to Ron Shepard):

```fortran
program foo
   call bar(42)
   call bar()
contains
   subroutine bar(a) bind(c)
      integer, value, optional :: a  ! <- illegal for a bind(c) procedure
      if ( .not. present(a) ) a = 100
      write(*,*) 'present(a)=', present(a), 'a=', a
   end subroutine bar
end program foo
```

Compiler Explorer link: https://godbolt.org/z/b9vTP4h7n

Reported error:

```txt
  store i32 100, i32 %a, align 4
```

Ideally, the compiler would print a clear message, similar to other compilers.

Source: https://fortran-lang.discourse.group/t/trick-for-optional-arguments-with-defaults/10852/18?u=ivanpribec

opixdown:
fixes #11080


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -23631,6 +23631,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 for (size_t i = 0; i < subrout_called->n_args; i++) {
                     ASR::expr_t* expected_arg = subrout_called->m_args[i];
                     ASR::expr_t* passed_arg = x.m_args[i].m_value;
+                    if (passed_arg == nullptr || expected_arg == nullptr) continue;
```
assem2002, line 23634: What is that actually tackling? Do we need this? is it correct?

assem2002 (approved):
Looks good. Thanks.
Please check me comment above.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -439,6 +439,16 @@ warning: Assuming implicit save attribute for variable declaration
...
+semantic error: Variable 'a' cannot have both the OPTIONAL and VALUE attribute because procedure 'bindc_optional_value' is BIND(C)
+   --> tests/errors/continue_compilation_1.f90:691:5 - 694:18
+    |
+691 |        subroutine bindc_optional_value(a) bind(c)
+    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
+...
+    |
+694 |        end subroutine
+    | ...^^^^^^^^^^^^^^^^^^ 
+
```
kmr-srbh, line 451: I think this should highlight the variable instead?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1391,6 +1391,17 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 }
             }
             ASR::symbol_t *var = current_scope->get_symbol(arg_s);
+            if (current_procedure_abi_type == ASR::abiType::BindC && ASR::is_a<ASR::Variable_t>(*var)) {
+                ASR::Variable_t *var_sym = ASR::down_cast<ASR::Variable_t>(var);
+                if (var_sym->m_value_attr && var_sym->m_presence == ASR::presenceType::Optional) {
+                    diag.add(diag::Diagnostic(
+                        "Variable '" + arg_s + "' cannot have both the OPTIONAL and VALUE attribute"
+                        " because procedure '" + std::string(x.m_name) + "' is BIND(C)",
```
kmr-srbh, line 1399:
```suggestion
                        "Variable `" + arg_s + "` cannot have both the OPTIONAL and VALUE attribute"
                        " because procedure `" + std::string(x.m_name) + "` is BIND(C)",
```

Please use backticks for symbol names.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1391,6 +1391,17 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                        diag::Level::Error, diag::Stage::Semantic, {diag::Label("", {x.m_args[i].loc})}));
+                    if (compiler_options.continue_compilation) {continue;
+                    } else {throw SemanticAbort();}
```
kmr-srbh, line 1402:
```suggestion
                    if (!compiler_options.continue_compilation) {
                        throw SemanticAbort();
                    }
```

kmr-srbh (commented):
Also Assem's question above, otherwise fine.

PR #11092, opened 2026-04-15
fix: leak of per-element string descriptors for character array passed to class(*)

### LINKED ISSUE #11050: memleak: character array of size N passed to unlimited polymorphic subroutine parameter leaks N string descriptors worth of memory

```fortran
program select_type_37
    implicit none
    character(5) :: a(10)
    call sub(a)
contains
    subroutine sub(x)
        class(*), intent(in) :: x(:)
        print *, "PASS"
    end subroutine
end program
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % lfortran test.f90 --detect-leaks
PASS

   ==> LEAK: 160 bytes

TOTAL LEAKS FOUND : 1, 160 bytes total
```

PS: We are not allocating N string descriptors, but N string descriptors worth of memory.

himanshumahajan138:
Fixes #11050

> Before fix: `LEAK SUMMARY: 160 bytes`
> After fix: `NO LEAKS FOUND`

`integration_tests/select_type_48.f90` passes


assem2002 (approved):
Looks good. Thanks!

assem2002:
CI needs to pass.

himanshumahajan138:
Hi @assem2002 ci paased

assem2002 (approved):
I think it's fine to merge. We should go through `expand_flat_to_char_descriptors()` -- This function doesn't fit our design choices when it comes to array of some type that's physically a descriptor (stringDescriptor, class wrapper descriptor, ..)

certik:
Thanks for the fix!

Issue #11090, opened 2026-04-15
Compiling Netlib packages


ivan-pi:
I had a go at creating a CMake driven build of many Netlib packages: https://codeberg.org/ivan-pi/netlib-fortran.git

After cloning, the project can be built with:

```
cmake -B build-lf -DCMAKE_Fortran_COMPILER="/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran" -DCMAKE_Fortran_FLAGS="--separate-compilation --linker=gcc --fixed-form-infer --infer --implicit-typing --implicit-interface --no-color" -DCMAKE_COLOR_DIAGNOSTICS=OFF
cmake --build build-lf -- -k
```
Note the `-k` to "keep going" in Make. 

I'm attaching the build log: [netlib_lf_63cd68.txt](https://github.com/user-attachments/files/26738043/netlib_lf_63cd68.txt)

It contains a few different classes of errors:

```
$ grep -n "tokenizer" netlib_lf_63cd68.txt
50:tokenizer error: Expecting terminating symbol for subroutine
63:tokenizer error: Token '4' is not recognized in `format` statement
94:tokenizer error: Token '1' is not recognized in `format` statement
119:tokenizer error: Token '1' is not recognized in `format` statement
335:tokenizer error: Token '1' is not recognized in `format` statement
348:tokenizer error: Token '4' is not recognized in `format` statement
362:tokenizer error: Token '3' is not recognized in `format` statement
376:tokenizer error: Token '4' is not recognized in `format` statement
390:tokenizer error: Token '1' is not recognized in `format` statement
404:tokenizer error: Token '2' is not recognized in `format` statement
418:tokenizer error: Token '4' is not recognized in `format` statement
432:tokenizer error: Token '4' is not recognized in `format` statement
446:tokenizer error: Token '2' is not recognized in `format` statement
460:tokenizer error: Token '4' is not recognized in `format` statement
474:tokenizer error: Token '2' is not recognized in `format` statement
488:tokenizer error: Token '3' is not recognized in `format` statement
502:tokenizer error: Token '4' is not recognized in `format` statement
516:tokenizer error: Token '1' is not recognized in `format` statement
530:tokenizer error: Token '4' is not recognized in `format` statement
544:tokenizer error: Token '2' is not recognized in `format` statement
558:tokenizer error: Token '4' is not recognized in `format` statement
572:tokenizer error: Token '4' is not recognized in `format` statement
[...]

PR #11101, opened 2026-04-15
fix (fixed-form tokenizer): handle assignments after all other tokens

### LINKED ISSUE #11096: Implicit statement newline (fixed-form)
Example taken from a the Netlib pltmg package of Randolph Bank (https://www.netlib.org/pltmg/):

```fortran
        module mthdef
cx
        integer(kind=4), parameter :: isngl=4
        integer(kind=4), parameter :: idble=8
        integer(kind=4), parameter :: rsngl=4
        integer(kind=4), parameter :: rdble=8
        integer(kind=4), parameter :: rquad=16
c
        integer(kind=4), parameter :: iknd=isngl
        integer(kind=4), parameter :: rknd=rdble
cy
        end module

        program atest
cx
            use mthdef
            implicit real(kind=rknd) (a-h,o-z)
            implicit integer(kind=iknd) (i-n)
        
        end program
```

Compiling with `lfortran --fixed-form` I get the error:

```
  --> /app/example.f90:17:47
   |
17 |             implicit real(kind=rknd) (a-h,o-z)
   |                                               ^ 
```

Compiler Explorer: https://godbolt.org/z/dTGfb458K

This is not a very common incantation of `implicit` syntax, but for refactoring it is nicer if one can start from a code that compiles.

kmr-srbh:
Fixes #11096 
Towards #11090 

* Fixed form tokenizer was incorrectly processing simple variable fields like `kind=4` as assignments, leading to incorrect token generation. For more info, please refer https://github.com/lfortran/lfortran/issues/11096#issuecomment-4254427821. We fix it by handling assignment statements before some statements in Fixed-Form Fortran.
* Correctly handle incorrect subroutine call syntax and `parameter` variable declarations
These were previously being handled through an incorrect use of the fragile assignment check inside the tokenizer. Coincidentally, the tests did not catch it because the subroutine call and variable declaration contained a "=" sign. For an example, see test `tests/fixed_form_call3.f`.
* Implicitly declared variables were unable to access variables imported from other modules for fields like `kind`. This was due to the implicit variables being processed before `use` statements. We fix this by handling `use` statements before to enable access.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2017,6 +2017,7 @@ RUN(NAME implicit_typing_05 LABELS gfortran llvm  EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_07 LABELS gfortran llvm EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_08 LABELS gfortran llvm EXTRA_ARGS --implicit-typing)
+RUN(NAME implicit_typing_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing --fixed-form GFORTRAN_ARGS -ffixed-form)
 
 
 RUN(NAME generic_name_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/src/lfortran/parser/fixedform_tokenizer.cpp
+++ b/src/lfortran/parser/fixedform_tokenizer.cpp
@@ -1090,6 +1090,13 @@ struct FixedFormRecursiveDescent {
             return true;
         }
 
+        // assignment
+        // TODO: this is fragile
+        if (is_possible_assignment(cur, cur)) {
+            tokenize_line(cur);
+            return true;
+        }
+
         if (lex_io(cur)) return true;
         if (next_is(cur, "if(")) {
             lex_cond(cur);
@@ -1266,13 +1273,6 @@ struct FixedFormRecursiveDescent {
         if (next_is(cur, "cycle")){
             push_token_advance(cur, "cycle");
             tokenize_line(cur);
-        }
-
-        // assignment
-        // TODO: this is fragile, check after all tokens
-        if (contains(cur, nline, '=')) {
-            tokenize_line(cur);
-            return true;
         }
 
         if (l != -1) {
--- a/src/lfortran/parser/fixedform_tokenizer.cpp
+++ b/src/lfortran/parser/fixedform_tokenizer.cpp
@@ -276,6 +276,7 @@ static const std::unordered_map<std::string, yytokentype> &identifier_token_map(
 const std::vector<std::string> declarators{
             "integer*",
             "integer",
+            "parameter",
 	    "real*",
             "real",
 	    "complex*",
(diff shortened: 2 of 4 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -575,17 +588,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                 }
             }
         }
```
kmr-srbh, line 590: The diff is being rendered incorrectly here. The actual change is the addition of the block below.

integration_tests/CMakeLists.txt
```diff
@@ -2013,6 +2013,7 @@ RUN(NAME implicit_typing_05 LABELS gfortran llvm  EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_07 LABELS gfortran llvm EXTRA_ARGS --implicit-typing)
 RUN(NAME implicit_typing_08 LABELS gfortran llvm EXTRA_ARGS --implicit-typing)
+RUN(NAME implicit_typing_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran EXTRA_ARGS --implicit-typing --fixed-form)
```
certik: You need to also add flags to GFortran to use fixed form.

kmr-srbh: Done in https://github.com/lfortran/lfortran/commit/ac9cd15542c702565e88496be4bade19540802d1.

kmr-srbh:
The CI failure is:

```console
FAILED: SRC/CMakeFiles/lapack_obj.dir/slarfgp.f.o 
/home/runner/work/lfortran/lfortran/src/bin/lfortran -I/tmp/tmp.q9j5Ojkcv7/assert/modern_minpack_01/modern_minpack_02/dftatom/fastGPT/stdlib/SNAP/src/lapack/build/lapack-testing/SRC --fixed-form-infer --implicit-interface --implicit-typing --legacy-array-sections --separate-compilation --use-loop-variable-after-loop -fPIC -c SRC/CMakeFiles/lapack_obj.dir/slarfgp.f-pp.f -o SRC/CMakeFiles/lapack_obj.dir/slarfgp.f.o
syntax error: Token 'alpha' (of type 'identifier') is unexpected here
   --> SRC/CMakeFiles/lapack_obj.dir/slarfgp.f-pp.f:190:14
    |
190 |          SAVEALPHA = ALPHA
    |              ^^^^^ 


```

Fixing it.

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1090,16 +1096,24 @@ struct FixedFormRecursiveDescent {
...
         }
 
+        if (is_do_loop(cur)) {
+            lex_do(cur);
+            return true;
+        }
+
+        // assignment
+        // TODO: this is fragile
+        if (is_possible_assignment(cur, cur)) {
```
certik, line 1106: I think this is not fragile anymore, the function `is_possible_assignment` is pretty solid.

certik (approved):
I think the changes are fine, because the tests pass, so I think we can merge it.

src/lfortran/parser/fixedform_tokenizer.cpp
kmr-srbh: I'll remove the comment in a different PR.

PR #11102, opened 2026-04-15
Tag container image with release version on tag pushes

### LINKED ISSUE #11100: Provide release tags on the container package
The [LFortran container package](https://github.com/lfortran/lfortran/pkgs/container/lfortran) is fabulously useful for CI and validating recent changes in LFortran.

Unfortunately it currently only has a single tag, the `latest` tag which changes with every run of the GitHub action:

Image

This is great for tracking the latest in LFortran development, but does not provide a stable basis for workflow reproducibility.

It would be really great if there was a container package tag corresponding to each LFortran release.

I think this mostly boils down to running the [build-and-push-image Job of your Exhaustive Checks Action](https://github.com/lfortran/lfortran/blob/main/.github/workflows/Exhaustive-Checks-CI.yml#L651..L682) for each release tag, and pushing it to a _unique_ package tag named for the release version. 

It looks like the Action is already set-up to run for release tags, but the [container push](https://github.com/lfortran/lfortran/blob/149977a2d74c6ba5d49ca4386287931bd63ff61c/.github/workflows/Exhaustive-Checks-CI.yml#L682) keeps overwriting the `latest` container tag at each push, so the container build corresponding to the release is currently overwritten upon the next push to the `main` branch.

CC: @certik @phhargrove

certik:
When pushing a release tag (v*), the container image is now tagged with both 'latest' and the version tag (e.g. v0.45.0). This provides stable, reproducible tags that persist after subsequent pushes to main.

Fixes #11100

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -674,9 +674,19 @@ jobs:
           username: ${{ env.USER }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Determine image tags
+        id: image-tags
+        run: |
+          TAGS="${{ env.REGISTRY }}/${{ env.USER }}/${{ env.PROJECT }}:latest"
+          if [[ "${{ github.ref }}" == refs/tags/v* ]]; then
+            VERSION="${{ github.ref_name }}"
+            TAGS="$TAGS,${{ env.REGISTRY }}/${{ env.USER }}/${{ env.PROJECT }}:${VERSION}"
+          fi
+          echo "tags=$TAGS" >> "$GITHUB_OUTPUT"
+
       - name: Build and push
         uses: docker/build-push-action@v7
         with:
           context: .
           push: ${{ github.event_name == 'push' }}
-          tags: ${{ env.REGISTRY }}/${{ env.USER }}/${{ env.PROJECT }}:latest
+          tags: ${{ steps.image-tags.outputs.tags }}


.github/workflows/Exhaustive-Checks-CI.yml
bonachea: IIUC the second part of this condition implies that any pull request labelled with `Tests::Run-Exhaustive` runs this `build-and-push-image` job, which also has a side-effect of overwriting the container tagged as `ghcr.io/lfortran/lfortran:latest`, _even though_ the commit is not actually yet in the `main` branch. 

This interpretation seems to be confirmed by the [Actions log for this PR](https://github.com/lfortran/lfortran/actions/runs/24483115335/job/71551951425?pr=11102):
```
#20 exporting to image
#20 exporting layers
#20 exporting layers 1.1s done
#20 writing image sha256:156504b5d5f6a90c9f7b85c911e1ff6f94f1718530f5343dba0b0dc5331fc4a9 done
#20 naming to ghcr.io/lfortran/lfortran:latest done
#20 DONE 1.1s
```
I think this means end users of the `:latest` container can never know whether they are getting a snapshot of `main` or a random commit in some labelled PR, which seems non-ideal.

Was that really your intent?

It probably makes more sense for PRs to either skip the container push step, or possibly write to a tag named for the PR?

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -674,9 +674,19 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Determine image tags
+        id: image-tags
+        run: |
+          TAGS="${{ env.REGISTRY }}/${{ env.USER }}/${{ env.PROJECT }}:latest"
+          if [[ "${{ github.ref }}" == refs/tags/v* ]]; then
+            VERSION="${{ github.ref_name }}"
+            TAGS="$TAGS,${{ env.REGISTRY }}/${{ env.USER }}/${{ env.PROJECT }}:${VERSION}"
+          fi
```
bonachea, line 684: Has this been tested?

IIUC you should be able to push a tag named `v0.whatever` to this commit your fork, and it should (at least try) to package a container with that tag?

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -674,9 +674,19 @@ jobs:
...
+          echo "tags=$TAGS" >> "$GITHUB_OUTPUT"
+
       - name: Build and push
         uses: docker/build-push-action@v7
         with:
           context: .
           push: ${{ github.event_name == 'push' }}
```
bonachea, line 691: IIUC this line implies that when this action runs on a pull request (due to label 'Tests::Run-Exhaustive'), the final container push step is skipped (which is the _right_ behavior when there was only one tag, ie `:latest`).

Now that you're adding more container tags, would there be any value to pushing PR containers to an appropriate tag (e.g. `:pr11102`)? This would allow PR reviewers to easily test labelled PRs using that container, without the effort to rebuild the compiler themselves. The downside is you'd probably need to periodically clean out old containers for closed PRs.

I don't see this as a requirement or even necessarily valuable, I just thought I'd point out that a few more lines of script in the new image-tags step could probably accomplish that.

bonachea (commented):
@certik Thanks for working on this!

I'm excited to see this capability deployed for future releases (and maybe even the current release, if it can be manually triggered?)

.github/workflows/Exhaustive-Checks-CI.yml
certik: No, I don't know how to test it other than running at the CI. The change seems ok, so I was going to merge it and hopefully it works. 

The alternative is to build a container at every PR etc., like you suggested below, but I think it will add another thing to worry about and that can break, so we'll have to maintain it.

I am trying to simplify things to maintain.

PR #11103, opened 2026-04-16
fix : false positive "DerivedType already defined" for member sharing type name

### LINKED ISSUE #11098: Allow a member to have the same name as the type
```fortran
program mre_member_same_name
   implicit none
   type :: foo
      integer :: foo
   end type foo
   type(foo) :: x
   x%foo = 42
   if (x%foo /= 42) error stop
   print *, "PASS"
end program mre_member_same_name
```
This gives:
```console
$ flang mre.f90 && ./a.out
 PASS
$ lfortran mre.f90
semantic error: DerivedType already defined
 --> mre.f90:3:4 - 5:16
  |
3 |       type :: foo
  |       ^^^^^^^^^^^...
...
  |
5 |       end type foo
  | ...^^^^^^^^^^^^^^^^
```

Jatinagarwal24:
Bug: When a derived type had a component with the same name as the type
itself (e.g., `type :: foo` containing `integer :: foo`), LFortran
incorrectly raised "DerivedType already defined". This is valid Fortran
and compiles successfully with gfortran.
Root cause: In visit_DerivedType, the duplicate-name check on line 2616
used `current_scope->get_symbol(sym_name)`, but at that point
`current_scope` points to the struct's child symbol table (where member
variables are stored), not the parent scope where the type symbol is
actually registered. A member variable sharing the type's name was found
in the child scope, triggering the false positive.
Fix: Change the check from `current_scope` to `parent_scope`, which is
the correct scope to verify against since the DerivedType symbol is
added to the parent scope.

fixes #11098

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -2613,7 +2613,7 @@ public:
             visit_procedure_decl(*x.m_contains[i]);
         }
         std::string sym_name = to_lower(x.m_name);
-        if (current_scope->get_symbol(sym_name) != nullptr) {
+        if (parent_scope->get_symbol(sym_name) != nullptr) {
             diag.add(diag::Diagnostic(
                 "DerivedType already defined",
                 diag::Level::Error, diag::Stage::Semantic, {
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -2613,7 +2613,7 @@ public:
             visit_procedure_decl(*x.m_contains[i]);
         }
         std::string sym_name = to_lower(x.m_name);
-        if (parent_scope->get_symbol(sym_name) != nullptr) {
+        if (!compiler_options.implicit_typing && parent_scope->get_symbol(sym_name) != nullptr) {
             diag.add(diag::Diagnostic(
                 "DerivedType already defined",
                 diag::Level::Error, diag::Stage::Semantic, {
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2613,7 +2613,7 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
             visit_procedure_decl(*x.m_contains[i]);
         }
         std::string sym_name = to_lower(x.m_name);
-        if (current_scope->get_symbol(sym_name) != nullptr) {
+        if (parent_scope->get_symbol(sym_name) != nullptr) {
```
kmr-srbh:
```suggestion
        if (!compiler_options.implicit_typing && parent_scope->get_symbol(sym_name) != nullptr) {
```

This should fix the CI failure I think. With implicit typing enabled (`--std=f23` also enables it), scope declarations like `public :: myType` can also declare implicitly typed variables. We overwrite these with type declarations that we encounter later, so we should allow this case.

kmr-srbh (approved):
I think the change is correct. However, there is a special case that we need to handle and disallow. Extended derived types contain their parent type as a member, so we still need to report an error for cases where a variable is named such.

Example:

```fortran
program derived_types_136
   implicit none
   type :: foo
      ! integer :: foo
   end type foo

   type, extends(foo) :: bar
      integer :: bar
      integer :: foo  ! parent type name
   end type bar

   type(bar) :: x
end program derived_types_136
```
```console
(base) saurabhkumar@Mac lfortran % gfortran test.f90
test.f90:9:20:

    7 |    type, extends(foo) :: bar
      |                            2
    8 |       integer :: bar
    9 |       integer :: foo
      |                    1
Error: Component 'foo' at (1) already declared at (2)
(base) saurabhkumar@Mac lfortran % flang test.f90
error: Semantic errors in test.f90
./test.f90:9:18: error: 'foo' is a parent type of this type and so cannot be a component
        integer :: foo
                   ^^^
./test.f90:7:18: Previous declaration of 'foo'
     type, extends(foo) :: bar
```

certik (approved):
I think it's fine, further refinement can be done in subsequent PRs.

PR #11104, opened 2026-04-16
fix(runtime): handle null and repeat tokens in list-directed scalar reads

### LINKED ISSUE #11094: List-directed READ: null input values not accepted
```
program read_list2
  implicit none

  character(*), parameter :: cdata = ", (2.0, 3.0),,6.0D0, 2*,"

  double precision :: d1, d2, d3
  complex :: c1, c2, c3

  open (42, file='fort.42', status='replace', form='formatted')
  write (42,'(a)') cdata
  rewind (42)
  d1 = 1.0d0
  c2 = (4.0, 5.0)
  c3 = (7.0, 8.0)
  d3 = 9.0d0
  read (42, *) d1, c1, c2, d2, c3, d3
  print *,     d1, c1, c2, d2, c3, d3

  print *, 'd1 = ', d1, merge (' : pass', ' : fail', abs (d1 - 1.0d0) < 0.0001d0)
  print *, 'c1 = ', c1, merge (' : pass', ' : fail',  &
      abs (real  (c1) - 2.0) < 0.0001 .and.  &
      abs (aimag (c1) - 3.0) < 0.0001)
  print *, 'c2 = ', c2, merge (' : pass', ' : fail',  &
      abs (real  (c2) - 4.0) < 0.0001 .and.  &
      abs (aimag (c2) - 5.0) < 0.0001)
  print *, 'd2 = ', d2, merge (' : pass', ' : fail', abs (d2 - 6.0d0) < 0.0001d0)
  print *, 'c3 = ', c3, merge (' : pass', ' : fail',  &
      abs (real  (c3) - 7.0) < 0.0001 .and.  &
      abs (aimag (c3) - 8.0) < 0.0001)
  print *, 'd3 = ', d3, merge (' : pass', ' : fail', abs (d3 - 9.0d0) < 0.0001d0)

  close (42, status='delete')

end program
```
```
$ lfortran read_list2.f90
Error: Invalid input from file.
$
```
```
$ gfortran read_list2.f90
$ ./a.out
   1.0000000000000000                  (2.00000000,3.00000000)             (4.00000000,5.00000000)   6.0000000000000000                  (7.00000000,8.00000000)   9.0000000000000000     
 d1 =    1.0000000000000000       : pass
 c1 =              (2.00000000,3.00000000)  : pass
 c2 =              (4.00000000,5.00000000)  : pass
 d2 =    6.0000000000000000       : pass
 c3 =              (7.00000000,8.00000000)  : pass
 d3 =    9.0000000000000000       : pass
$
```

Jatinagarwal24:
fix(runtime): handle null and repeat tokens in list-directed scalar reads

Correct parsing of list-directed input in lfortran_intrinsics.c where
leading separators and repeat-null tokens were previously mis-handled,
causing errors such as "Invalid input from file" for inputs with
comma/null sequences.

- add per-unit tracking for null-repeat tokens
- implement helpers to detect and consume pending null repeats
- treat leading commas as null values (leave variable unchanged)
- support n* null-repeat tokens in scalar reads
- ensure consistent handling of trailing separators
- update complex scalar parsing to return explicit null-value status
- consume trailing commas after logical reads to avoid stale separators
fixes #11094

### DIFF
--- a/integration_tests/read_75.f90
+++ b/integration_tests/read_75.f90
@@ -11,7 +11,7 @@ program read_75
   double precision :: d1, d2, d3
   complex :: c1, c2, c3
 
-  open (42, file='fort.42', status='replace', form='formatted')
+  open (42, file='read_75.dat', status='replace', form='formatted')
   write (42,'(a)') cdata
   rewind (42)
   d1 = 1.0d0
--- a/integration_tests/read_75.f90
+++ b/integration_tests/read_75.f90
@@ -1,0 +1,32 @@
+program read_75
+  implicit none
+
+  ! Test list-directed read with null values and repeat-count syntax.
+  ! Input: ", (2.0, 3.0),,6.0D0, 2*,"
+  !   d1 -> null (keep 1.0), c1 -> (2.0,3.0), c2 -> null (keep (4.0,5.0)),
+  !   d2 -> 6.0, c3 -> null via 2* (keep (7.0,8.0)), d3 -> null via 2* (keep 9.0)
+
+  character(*), parameter :: cdata = ", (2.0, 3.0),,6.0D0, 2*,"
+
+  double precision :: d1, d2, d3
+  complex :: c1, c2, c3
+
+  open (42, file='fort.42', status='replace', form='formatted')
+  write (42,'(a)') cdata
+  rewind (42)
+  d1 = 1.0d0
+  c2 = (4.0, 5.0)
+  c3 = (7.0, 8.0)
+  d3 = 9.0d0
+  read (42, *) d1, c1, c2, d2, c3, d3
+
+  if (abs(d1 - 1.0d0) > 0.0001d0) error stop "d1 failed"
+  if (abs(real(c1) - 2.0) > 0.0001 .or. abs(aimag(c1) - 3.0) > 0.0001) error stop "c1 failed"
+  if (abs(real(c2) - 4.0) > 0.0001 .or. abs(aimag(c2) - 5.0) > 0.0001) error stop "c2 failed"
+  if (abs(d2 - 6.0d0) > 0.0001d0) error stop "d2 failed"
+  if (abs(real(c3) - 7.0) > 0.0001 .or. abs(aimag(c3) - 8.0) > 0.0001) error stop "c3 failed"
+  if (abs(d3 - 9.0d0) > 0.0001d0) error stop "d3 failed"
+
+  close (42, status='delete')
+  print *, "ok"
+end program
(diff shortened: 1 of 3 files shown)


Jatinagarwal24:
Hi @certik Pls review this and pls tell if any changes is required.

integration_tests/read_75.f90
```diff
@@ -0,0 +1,32 @@
...
+  ! Input: ", (2.0, 3.0),,6.0D0, 2*,"
+  !   d1 -> null (keep 1.0), c1 -> (2.0,3.0), c2 -> null (keep (4.0,5.0)),
+  !   d2 -> 6.0, c3 -> null via 2* (keep (7.0,8.0)), d3 -> null via 2* (keep 9.0)
+
+  character(*), parameter :: cdata = ", (2.0, 3.0),,6.0D0, 2*,"
+
+  double precision :: d1, d2, d3
+  complex :: c1, c2, c3
+
+  open (42, file='fort.42', status='replace', form='formatted')
```
certik: Use a name that is unique to this test to avoid race conditions with other tests when run in parallel.

certik (approved):
I think it's fine, but there might be a bug in https://github.com/lfortran/lfortran/pull/11104/changes#r3094734358, let's figure it out first.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -6886,6 +6886,36 @@ static void skip_list_directed_comma(FILE *filep) {
     if (c != EOF) ungetc(c, filep);
 }
 
+// List-directed read: track remaining null-value repeats per unit.
+#define LF_LIST_DIR_MAX_UNITS 1024
+static int lf_list_dir_null_remaining[LF_LIST_DIR_MAX_UNITS] = {0};
```
certik: How is this initialized and then finalized when the unit is closed and reopened for a new file?

Is there a mode that this will fail, due to the global state not being synced properly?

Jatinagarwal24: Thanks @certik for pointing this out. Right now this array is only process-initialized, not unit-lifecycle initialized/finalized.
So yes, there is a potential stale-state case if a unit is closed and reused for another file without resetting this per-unit slot.

I’ll fix it by making this state per-unit lifecycle managed (reset on open/reopen/close, same pattern as existing unit state), so no global-state carryover can occur

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5201,12 +5201,20 @@ struct UNIT_FILE unit_to_file[MAXUNITS];
 static int32_t seq_char_pending[MAXUNITS];
 static int32_t seq_char_record_len[MAXUNITS];
 
+// Pending list-directed null-repeat state per unit.
+static int32_t lf_list_dir_null_remaining[MAXUNITS];
```
certik, line 5205: I think all these should be moved into the UNIT_FILE struct, then there shouldn't be a problem, right?

That can be done in the next PR (https://github.com/lfortran/lfortran/issues/11107).

certik (approved):

PR #11114, opened 2026-04-17
Fix fixed-form parsing of variables starting with IO keywords

### LINKED ISSUE #11108: Token '=' is unexpected here
This is a MWE from the comment here: https://github.com/lfortran/lfortran/issues/11090#issuecomment-4253497808

```
      SUBROUTINE SETUP()
C     .. Common Block for General Workspace Pointers ..
      INTEGER           PRTHRS, PRERST, PRWT, PRYOLD, PRSCR, PRY, PRYP,
     &                  PRSTGS, PRINTP, LNINTP
      COMMON /RKCOM3/   PRTHRS, PRERST, PRWT, PRYOLD, PRSCR, PRY, PRYP,
     &                  PRSTGS, PRINTP, LNINTP
      SAVE   /RKCOM3/
      PRINTP = 1
      RETURN
      END
```

```
 --> /app/example.f90:8:14
  |
8 |       PRINTP = 1
  |              ^ 
```

CE: https://godbolt.org/z/eYzWzz1Pn

I can actually reduce the example all the way down to, 

```
      SUBROUTINE SETUP()
      INTEGER PRINTP
      COMMON /RKCOM3/ PRINTP
      SAVE /RKCOM3/
      PRINTP = 1
      RETURN
      END
```

certik:
In fixed-form Fortran, the prescanner removes spaces and lowercases everything, so PRINTP = 1 becomes printp=1. The fixed-form tokenizer's lex_io() was matching 'print' as a prefix and incorrectly treating the line as a PRINT statement.

Add is_io_keyword_assignment() to check whether the text after an IO keyword prefix forms an assignment (identifier continuation followed by '=' at parenthesis depth 0 before any ','). If so, lex_io() returns false and lets the assignment handler process the line.

An integration test is added covering variables starting with print, read, write, and close keywords.

Fixes #11108.

### DIFF
--- a/src/lfortran/parser/fixedform_tokenizer.cpp
+++ b/src/lfortran/parser/fixedform_tokenizer.cpp
@@ -906,9 +906,33 @@ struct FixedFormRecursiveDescent {
         return false;
     }
 
+    // Check if the line starting at `cur` is an assignment to a variable
+    // whose name starts with an IO keyword (e.g., PRINTP = 1 in fixed-form
+    // becomes printp=1 after prescanning). Returns true if there is a '='
+    // (not '==') before any ',' at parenthesis depth 0, indicating an
+    // assignment rather than an IO statement.
+    bool is_io_keyword_assignment(unsigned char *p) {
+        if (*p == '=' && *(p+1) != '=') return true;
+        if (!is_char(*p) && !is_digit(*p) && *p != '_') return false;
+        int depth = 0;
+        while (*p != '\n' && *p != '\0') {
+            if (*p == '(') depth++;
+            else if (*p == ')') depth--;
+            else if (depth == 0) {
+                if (*p == '=' && *(p+1) != '=') return true;
+                if (*p == ',') return false;
+            }
+            p++;
+        }
+        return false;
+    }
+
     bool lex_io(unsigned char *&cur) {
         for(const auto &io_str: io_names) {
             if (next_is(cur, io_str)) {
+                if (is_io_keyword_assignment(cur + io_str.size())) {
+                    return false;
+                }
                 if (io_str == "format") {
                     unsigned char *format_start = cur;
                     cur += io_str.size();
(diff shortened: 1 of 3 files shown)


src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -906,9 +906,33 @@ struct FixedFormRecursiveDescent {
         return false;
     }
 
+    // Check if the line starting at `cur` is an assignment to a variable
+    // whose name starts with an IO keyword (e.g., PRINTP = 1 in fixed-form
+    // becomes printp=1 after prescanning). Returns true if there is a '='
+    // (not '==') before any ',' at parenthesis depth 0, indicating an
+    // assignment rather than an IO statement.
+    bool is_io_keyword_assignment(unsigned char *p) {
```
assem2002, line 914: won't a simple length check do the job?
If you see it's fine, let's merge then.

assem2002 (approved):
It looks good to me.

kmr-srbh:
The changes in this PR are fine. The linked issue works with the tokeniser changes in #11101 actually. The CI fails there though as it reveals a few more bugs in the tokenizer.

HarshitaKalani (approved):

src/lfortran/parser/fixedform_tokenizer.cpp
certik: Can you write what kind of simple length check you are thinking?

Like `printp` being longer than `print`? That will not work, since that's an IO statement `print p`, which is what we were parsing before. We need to distinguish `print p` from `printp =`, which is what this code is doing.

There is probably a more systematic way to do this "backtracking" from a grammar directly, and we can later redo the fixed-form parser/tokenizer to be generated from a grammar with backtracking. Until then we need to do this logic ourselves.

assem2002: Oh missed that. Removing white spaces makes thing go in-intuitively.
It looks good enough for now.

PR #11122, opened 2026-04-17
Leak Fix

Jatinagarwal24:


src/libasr/codegen/llvm_utils.h
```diff
@@ -1649,14 +1626,115 @@ class ASRToLLVMVisitor;
          * @param data_type should be the underlying ASR type of the array.
          * @param array_size is lambda object returning array's size. Lazy evaluate as only array of structs requires looping on each element (for now).
          */
+        void free_upoly_string_payload_chars(llvm::Value* payload_ptr, llvm::Value* n_elems_i64) {
```
certik, line 1631: I think you put this function between a comment and another function, so now the above comment is out of place.

certik: While at it, document what `free_upoly_string_payload_chars` does.

Jatinagarwal24:
Hi @certik I think we can close this as this fixed via pr #11128 which does the same job.

certik:
@Jatinagarwal24 thanks! Is https://github.com/lfortran/lfortran/pull/11128 fixing it the same way, or is this PR better?

PR #11124, opened 2026-04-17
enh: add support for `read` without input-item-list

### LINKED ISSUE #11097: read without input/output-item-list
This issue originates from the MGGHAT (https://netlib.org/pdes/mgghat/) code on Netlib. A second link is here: https://codeberg.org/ivan-pi/netlib-fortran/src/commit/0fc6f3488b321c0d2c511ef54571e7f3260848ee/pdes/mgghat/src/mgghat.f#L72

```
$ lfortran --fixed-form --implicit-typing --implicit-interface -c --separate-compilation pdes/mgghat/src/mgghat.f
  --> pdes/mgghat/src/mgghat.f:72:19
   |
72 |             read *
   |                   ^
```

According to the standard (12.6.1, J3/24-007), the read-stmt can take the form,

```
R1210 read-stmt is READ ( io-control-spec-list ) [ input-item-list ]
                or READ format [ , input-item-list ]
```
so the item-list is optional. In case of `*` format, essentially the process waits for the user to press enter, IIUC.

kmr-srbh:
Resolves #11097
Towards #11090

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -1915,7 +1915,8 @@ read_statement
     | KW_READ "(" write_arg_list ")" "," expr_list { $$ = READ($3, $6, @$); }
     | KW_READ "(" write_arg_list ")" { $$ = READ0($3, @$); }
     | KW_READ TK_INTEGER "," expr_list { $$ = READ2($2, $4, @$); }
-    | KW_READ "*" "," expr_list { $$ = READ3($4, @$); }
+    | KW_READ "*" "," expr_list_opt { $$ = READ3($4, @$); }
+    | KW_READ "*" { $$ = READ6(@$); }
     | KW_READ TK_INTEGER { $$ = READ4($2, @$); }
     | KW_READ TK_STRING "," expr_list { $$ = READ5($2, $4, @$); }
     ;
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -10043,7 +10043,13 @@ static void common_formatted_read(InputSource *inputSource,
 LFORTRAN_API void _lfortran_empty_read(int32_t unit_num, int32_t* iostat) {
     if (iostat) *iostat = 0;
     if (unit_num == -1) {
+        return;
+    } else if (unit_num == -2) {
         // Read from stdin
+        int inp = 0;
+        do {
+            inp = fgetc(stdin);
+        } while (inp != '\n' && inp != EOF);
         return;
     }
 
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3356,12 +3356,11 @@ RUN(NAME read_75 LABELS gfortran llvm)
 RUN(NAME read_76 LABELS gfortran llvm)
 RUN(NAME read_77 LABELS gfortran llvm)
 RUN(NAME read_78 LABELS gfortran llvm)
-
 RUN(NAME read_79 LABELS gfortran llvm)
 RUN(NAME read_80 LABELS gfortran llvm)
-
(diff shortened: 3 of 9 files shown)


src/lfortran/parser/parser.yy
```diff
@@ -1914,7 +1914,8 @@ read_statement
     | KW_READ "(" write_arg_list ")" "," expr_list { $$ = READ($3, $6, @$); }
     | KW_READ "(" write_arg_list ")" { $$ = READ0($3, @$); }
     | KW_READ TK_INTEGER "," expr_list { $$ = READ2($2, $4, @$); }
-    | KW_READ "*" "," expr_list { $$ = READ3($4, @$); }
+    | KW_READ "*" "," expr_list_opt { $$ = READ3($4, @$); }
+    | KW_READ "*" { $$ = READ6(@$); }
```
certik, line 1919: In the next PR add a test for this syntax into reference tests. Just test ASR.

certik: Actually, create a regular integration test, put `read *` into an if statement that is not executed by default (perhaps only with a command line argument), and run the tests as usual, test with gfortran, llvm, etc.

Then add llvm test into reference tests.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -10040,7 +10040,13 @@ static void common_formatted_read(InputSource *inputSource,
 LFORTRAN_API void _lfortran_empty_read(int32_t unit_num, int32_t* iostat) {
     if (iostat) *iostat = 0;
     if (unit_num == -1) {
+        return;
+    } else if (unit_num == -2) {
```
certik, line 10047: Document this please:
```
-1 ... move the file pointer ahead
-2 ... taking the input from stdin and redirecting to null
```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -10040,7 +10040,13 @@ static void common_formatted_read(InputSource *inputSource,
...
         // Read from stdin
+        int _ = 0;
```
certik: I would not use `_` as a variable name.

src/lfortran/parser/parser.yy
kmr-srbh: I found a bug when I executed my tests. Will fix it and mark this ready.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: Done in https://github.com/lfortran/lfortran/pull/11124/commits/8f69dec53e98e6a29363d4cbda6d4e802deaf834.

src/lfortran/parser/parser.yy
kmr-srbh: Done in https://github.com/lfortran/lfortran/pull/11124/commits/596ac22dab4b9d66945d52fcb9ce1dc217b9c071.

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: I noticed we probably might not be using `-1` at all, but need to verify it once. I will address this in a different PR.

integration_tests/read_80.f90
```diff
@@ -0,0 +1,24 @@
+program read_80
+   implicit none
+
+   character(len=32) :: empty_read
+   integer :: unit_num
+
+   call get_command_argument(1, empty_read)
```
certik: When there are no arguments passed, will this return an empty string?

certik (approved):
I think this looks good now.

certik:
Please resolve the conflict. Use 82.

integration_tests/read_80.f90
kmr-srbh: Yes.

PR #11131, opened 2026-04-18
fix: accept !$omp threadprivate in specification parts (safe diagnostic only)

### LINKED ISSUE #11130: OpenMP: threadprivate directive not parsed with --openmp
## Summary
With `--openmp` enabled, LFortran's parser rejects `!$omp threadprivate(...)` as an unexpected pragma token. Without `--openmp`, the sentinel is silently ignored (correct behavior since #9444/#6449), but the `--openmp`-on path does not implement the directive at all.

gfortran accepts and implements the directive under `-fopenmp`.

## Reproducer
```fortran
module mw_mod
  implicit none
  integer :: iuser_mw
  real :: ruser_mw
  !$omp threadprivate(iuser_mw, ruser_mw)
end module mw_mod

program test_tp
  use mw_mod
  implicit none
  iuser_mw = 1
  ruser_mw = 2.0
  print *, iuser_mw, ruser_mw
end program
```

## Actual behavior
```
$ lfortran --openmp mre.f90
syntax error: Token '!$omp threadprivate(iuser_mw, ruser_mw)' (of type 'pragma') is unexpected here
 --> mre.f90:5:3
  |
5 |   !$omp threadprivate(iuser_mw, ruser_mw)
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

```
$ lfortran mre.f90          # without --openmp
(compiles; sentinel ignored)

$ gfortran -fopenmp mre.f90 && ./a.out
           1   2.00000000
```

## Expected behavior
Under `--openmp`, `!$omp threadprivate(...)` should at minimum parse (and ideally be honored). The existing handling when `--openmp` is off (#9444, treat as comment) should not regress.

## Environment
- LFortran: `0.61.0-921-gba8c34df3` (upstream/main-ish)
- Platform: Linux, LLVM 11

## Related
- #9447: itpplasma/vode — mentions this failure alongside other blockers; here the concern is isolated to the `--openmp`-on parse gap.
- #6449 / PR #9444: fixed the `--openmp`-off path (treat as comment). Not the same path.

krystophny:
## Summary

The parser rejected `!\$omp threadprivate(...)` at tokenizer-level in
specification parts when `--openmp` was enabled, before the user could see
any semantic diagnostic. Extend the grammar so the sentinel is recognized in
module/submodule specification parts, then emit a **clear semantic error**
pointing at the implementation tracker instead of silently producing shared
storage.

Not implementing real thread-local storage in this PR is deliberate. A silent
no-op under `--openmp` would compile user code but serialise every thread
onto the same storage — a correctness hazard worse than a compile error.
Per-thread lowering is tracked in #11132 and will land as a stacked follow-up.

Fixes #11130. Related: #11132 (implementation tracker).

**Stage:** Parser + Semantics.

## Changes
- [`src/lfortran/parser/parser.yy`](https://github.com/lfortran/lfortran/blob/666c733f2d9296888c4c25e596a01e161cbd1652/src/lfortran/parser/parser.yy#L1126-L1132): accept `TK_OMP sep` in `decl` so OpenMP sentinels parse in module/submodule specification parts.
- [`src/lfortran/parser/semantics.h`](https://github.com/lfortran/lfortran/blob/666c733f2d9296888c4c25e596a01e161cbd1652/src/lfortran/parser/semantics.h#L264-L274): new `VAR_DECL_OMP_PRAGMA` helper building an AST `DeclarationPragma` tagged with `pragma_type=OMPPragma`.
- [`src/lfortran/semantics/ast_symboltable_visitor.cpp`](https://github.com/lfortran/lfortran/blob/666c733f2d9296888c4c25e596a01e161cbd1652/src/lfortran/semantics/ast_symboltable_visitor.cpp): under `--openmp`, reject `DeclarationPragma` of type `OMPPragma` with a diagnostic pointing at #11132. Without `--openmp`, silently accept (sentinel behaviour per #9444).
- [`src/lfortran/semantics/ast_body_visitor.cpp`](https://github.com/lfortran/lfortran/blob/666c733f2d9296888c4c25e596a01e161cbd1652/src/lfortran/semantics/ast_body_visitor.cpp): same diagnostic for `threadprivate` when it appears in executable sections (previously fell through to a generic "construct not supported yet" message).

## Tests
- [`integration_tests/openmp_74.f90`](https://github.com/lfortran/lfortran/blob/666c733f2d9296888c4c25e596a01e161cbd1652/integration_tests/openmp_74.f90): unchanged labels (`gfortran llvm`). Confirms no regression on the no-`--openmp` path.

The `--openmp` path is intentionally left without an integration test here because Option 2 rejects compilation. A positive integration test comes with the stacked TLS PR once per-thread storage actually works.

### Parse error on main (module-spec MRE with --openmp)
```
$ cat /tmp/omp_tp_min.f90
module mw_mod
  implicit none
  integer :: iuser_mw
  real :: ruser_mw
  !$omp threadprivate(iuser_mw, ruser_mw)
end module mw_mod

program test_tp
  use mw_mod
  iuser_mw = 1; ruser_mw = 2.0
  print *, iuser_mw, ruser_mw
end program

$ git checkout upstream/main
$ ./lfortran/build/src/bin/lfortran --openmp /tmp/omp_tp_min.f90
syntax error: Token '!$omp threadprivate(iuser_mw, ruser_mw)' (of type 'pragma') is unexpected here
 --> /tmp/omp_tp_min.f90:5:3
exit=1
```

### Clear semantic error on this branch
```
$ git checkout fix/openmp-threadprivate-parse
$ ./lfortran/build/src/bin/lfortran --openmp /tmp/omp_tp_min.f90
semantic error: OpenMP `!$omp threadprivate(iuser_mw, ruser_mw)` is not yet implemented;
compiling under `--openmp` would produce shared storage and silently break correctness.
See https://github.com/lfortran/lfortran/issues/11132
 --> /tmp/omp_tp_min.f90:5:3
exit=1
```

### Compilation still works without --openmp (unchanged)
```
$ ./lfortran/build/src/bin/lfortran /tmp/omp_tp_min.f90 && ./a.out
           1   2.00000000
exit=0
```

### Executable-section variant — same clean diagnostic instead of the pre-existing "construct not supported yet"
```
$ cat /tmp/omp_tp_exec.f90
program foo
  integer :: x
  x = 1
  !$omp threadprivate(x)
  print *, x
end program

$ ./lfortran/build/src/bin/lfortran --openmp /tmp/omp_tp_exec.f90
semantic error: OpenMP `threadprivate` is not yet implemented; ...
See https://github.com/lfortran/lfortran/issues/11132
exit=1
```

### Reference and integration suites
```
$ scripts/lf.sh test
TESTS PASSED

$ scripts/lf.sh itest -b llvm -t openmp
100% tests passed, 0 tests failed out of 6
```

## Stacked follow-up

Implementation of real thread-local storage (the fix that makes `--openmp`
compilation of these codes correct) is tracked in #11132 and will come as a
separate draft PR stacked on this branch.


krystophny:
sorry, let's better do #11133 right away. closing

PR #11136, opened 2026-04-18
fix: detect unexpected token after if condition in fixed-form

### LINKED ISSUE #11125: bug: fixed-form tokenizer cannot correctly detect invalid `if` statement syntax

```fortran
      subroutine idd_frm(a)
         character a*10
         if (a == sayhello())) error stop    ! notice the extra ')' at the condition end

      contains

         function sayhello() result(res)
            character res*5
            res = "hello"
         end function sayhello
      end
```

### Output

```console
(lf) saurabhkumar@Mac lfortran % gfortran test.f90 -ffixed-form -c
test.f90:3:30:

    3 |          if (a == sayhello())) error stop
      |                              1
Error: Missing '(' in statement at or before (1)
(lf) saurabhkumar@Mac lfortran % flang test.f90 -ffixed-form -c
error: Could not scan test.f90
./test.f90:3:30: error: Unmatched ')'
           if (a == sayhello())) error stop
                               ^
(lf) saurabhkumar@Mac lfortran % lfortran test.f90 --fixed-form -c
 --> test.f90:5:15 - 7:10
  |
5 |          contains
  |                  ...
...
  |
7 |             function sayhello() result(res)
  | ...^^^^^^^^^^ 


```

himanshumahajan138:
Fixes #11125.

In fixed-form, `if (a == sayhello())) error stop` used to produce a
misleading error pointing at unrelated lines. The tokenizer now reports
the stray `)` at its exact location.

Added MRE test `tests/errors/fixed_form_if_extra_paren.f`.

### Before:
Screenshot 2026-04-18 at 7 26 29 PM

### After: 
Screenshot 2026-04-18 at 7 25 31 PM


tests/errors/fixed_form_if_extra_paren.f
```diff
@@ -0,0 +1,11 @@
...
+         character a*10
+         if (a == sayhello())) error stop
+
+      contains
+
+         function sayhello() result(res)
+            character res*5
+            res = "hello"
+         end function sayhello
+      end
```
certik: Can you try to put this into continue_compilation_ff_1.f and see if we can recover from this error?

himanshumahajan138: Sure

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -790,6 +790,15 @@ struct FixedFormRecursiveDescent {
...
                 end++;
+                if (*end == ')') {
+                    Location loc;
+                    loc.first = end - string_start;
+                    loc.last = end - string_start;
+                    diag.add(diag::Diagnostic(
+                        "Unmatched `)` after the condition of the if statement",
+                        diag::Level::Error, diag::Stage::Tokenizer, {diag::Label("", {loc})}));
+                    throw parser_local::TokenizerAbort();
+                }
```
kmr-srbh: This will check for exactly the ')' character after the condition is checked so other invalid characters will still be skipped. I think the correct fix for this, is to report an error if the token we encounter before scanning the if statement's body, is not a "character". Note that digits are still allowed due to the arithmetic if statement (we already check this, so nothing more required), but we should disallow anything else, like punctuation here. This will ensure that the control goes to the `lex_body_statement()` with the first token always being a character.

kmr-srbh (changes requested):

tests/errors/fixed_form_if_extra_paren.f
himanshumahajan138: sir it worked but do i need to keep it in the file mentioned like i placed this here in the file (bottom)
`IF (i == ICHAR(x_2))) error stop`


```
c     Use the space above to insert new declarations, and remove the line, so
c     that the lines below do not shift, to keep the diff minimal.
c     Only put statements below. If you need to call a function, put it into a
c     module above.      

      CLOSE(UNIT=200)
      I == 10
      J = 20..5
      PRINT *, "Value of I is" I
      CHARACTER*10 STR = 'Hello  '
      WRITE(*,10 FORMAT(A))
      A = B +  * C
      INTEGER K L M = 5
      READ *, N 5
      REAL X Y Z 10.5
      COMPLEX C = (1.0,2.0
      I === 10
      J = .20E
      REAL*8 A = 1.23.45
      INTEGER*4 VAR/5/
      INTEGER*4 VAR/5/
      EQUIVALENCE (X Y)

      x_2 = 'u'
      i = 10
      IF (i > ICHAR(x_2)) THEN
      END IF

      size_a = SIZE(a_2, DIM=1)

      a_5 = CMPLX(1)
      IF I .EQ. 10 THEN PRINT *, "Ten"
      CALL FUNC( 5, 6,
      IF (i == ICHAR(x_2))) error stop
      END PROGRAM continue_compilation_ff
```

himanshumahajan138: keeping it in this commit...

himanshumahajan138:
CI passed i guess its ready to review now...

tests/errors/fixed_form_if_extra_paren.f
kmr-srbh, line 1: I think we do not need a separate test file for this as we are testing only a single independent statement. I think the test-case added in `tests/errors/continue_compilation_ff_1.f` is fine.

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -790,6 +790,28 @@ struct FixedFormRecursiveDescent {
...
+                        // Skip stray chars so the body below starts with a
+                        // valid token (letter, digit, or end-of-line).
+                        while (*end != '\n' && *end != '\0'
+                                && !is_char(*end) && !is_digit(*end)) {
+                            end++;
+                        }
+                    } else {
+                        throw parser_local::TokenizerAbort();
+                    }
+                }
```
kmr-srbh, line 814: This is perfect!

src/lfortran/parser/fixedform_tokenizer.cpp
kmr-srbh, line 818: We need to throw a similar error here in an `else` block if the token after `then` is not a new line or comment. Just a simple error that the token is unexpected. It is not related to this PR. You can open a different PR for this.

kmr-srbh (approved):
Just a small change, otherwise looks good. Thanks for working on this!

tests/errors/fixed_form_if_extra_paren.f
himanshumahajan138: ok so the separate test created as tests/errors/fixed_form_if_extra_paren.f is not required and so as its err and json file

src/lfortran/parser/fixedform_tokenizer.cpp
himanshumahajan138: okie will raise another pr

kmr-srbh:
Please also update the PR title.

himanshumahajan138:
DONE !!!

PR #11137, opened 2026-04-18
Fix memory leaks in set/dict/tuple/list

certik:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -410,6 +414,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+        // Zero-initialize struct allocas (e.g. list, tuple) at entry block
+        // so null-guarded free is safe on the very first use in a loop.
+        if (type->isStructTy()) {
+            llvm::BasicBlock &entry_block = builder->GetInsertBlock()->getParent()->getEntryBlock();
+            llvm::IRBuilder<> entry_builder(context);
+            entry_builder.SetInsertPoint(&entry_block, std::next(alloca->getIterator()));
+            uint64_t type_size = module->getDataLayout().getTypeAllocSize(type);
+            entry_builder.CreateMemSet(alloca, entry_builder.getInt8(0), type_size,
+                llvm::MaybeAlign(alloca->getAlign()));
+        }
```
certik, line 430: This is fine for now to fix all leaks, but we need to investigate why this is needed, as we shouldn't need to zero-initialize memory. This applies to all struct-type allocas, not just containers.

certik:
TODO for subsequent PRs (https://github.com/lfortran/lfortran/issues/11140):

* Set and tuples will leak arguments
* Investigate the above zero-initialization (https://github.com/lfortran/lfortran/issues/11139)
* Unify the call-arg cleanup
* Unify the ownership cleanup

None of this is a blocker.

src/libasr/codegen/asr_to_llvm.cpp
certik: https://github.com/lfortran/lfortran/issues/11139

PR #11138, opened 2026-04-18
Remove all NO_DETECT_LEAK by adding explicit deallocate() statements

certik:
Add explicit deallocate() calls to all 30 integration tests that had NO_DETECT_LEAK labels due to user-level pointer leaks. Each test now properly frees all allocated pointer memory before program exit.

Tests modified: arrays_10, arrays_34, associate_08, associate_16, associate_17, associate_19, associate_22, associate_48, associate_49, bindc_04, class_141, class_142, class_21, class_28, class_35, class_51, derived_types_43, derived_types_61, derived_types_100, file_14, functions_18, hashmap_derived_pointer_associated,
hashmap_nested_dealloc_derived_pointer, intrinsics_412, nullify_07, pointer_01, pointer_09, select_type_10, select_type_32, string_88


certik:
The upsides of this PR:

* No more `NO_DETECT_LEAK`, nor `# USER_LEAK` present (no special cases, all new tests must naturally pass leak detection, no tempting to append the `NO_DETECT_LEAK` label)
* All tests are now tested for leak detection, including all these pointer tests, which previously weren't: so there could have been a compiler-caused memory leak, on top of the user leak. Now we know 100% there are no leaks.

Downsides:

* Some of these pointer tests might have been testing a code path in our compiler or a test which only triggers if the pointer is allocated; we now explicitly deallocate before `nullify` or pointer reassignment, which can possibly hide a bug that previously would have been caught. But I think if we hit such a bug in the future, we can write a dedicated test, and use `#ifdef` to only call `deallocate` when for `--detect-leaks`, but not for regular run, thus still being able to test the bug.

I think the upsides outweigh the downsides, and we can modify these tests or add tests in the future as needed.

PR #11145, opened 2026-04-18
fix: print array-item lhs names in runtime errors

jassem-manita:
Follow-up to #11031: this makes the runtime error print `x(1)%arr` instead of falling back to the generic message.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -214,6 +214,37 @@ private:
             return ASRUtils::symbol_name(ASRUtils::symbol_get_past_external(v->m_v));
         }
 
+        if (ASR::is_a<ASR::IntegerConstant_t>(*expr)) {
+            ASR::IntegerConstant_t* i = ASR::down_cast<ASR::IntegerConstant_t>(expr);
+            return std::to_string(i->m_n);
+        }
+
+        if (ASR::is_a<ASR::ArrayItem_t>(*expr)) {
+            ASR::ArrayItem_t* array_item = ASR::down_cast<ASR::ArrayItem_t>(expr);
+            std::string base_name = get_expr_name_for_runtime_message(array_item->m_v);
+            if (base_name.empty()) {
+                return "";
+            }
+
+            std::string result = base_name + "(";
+            for (size_t i = 0; i < array_item->n_args; i++) {
+                ASR::array_index_t& idx = array_item->m_args[i];
+                if (idx.m_left != nullptr || idx.m_right == nullptr || idx.m_step != nullptr) {
+                    return "";
+                }
+                std::string index_name = get_expr_name_for_runtime_message(idx.m_right);
+                if (index_name.empty()) {
+                    return "";
+                }
+                if (i > 0) {
+                    result += ", ";
+                }
+                result += index_name;
+            }
+            result += ")";
+            return result;
+        }
+
         if (ASR::is_a<ASR::StructInstanceMember_t>(*expr)) {
             ASR::StructInstanceMember_t* sim = ASR::down_cast<ASR::StructInstanceMember_t>(expr);
             std::string base_name = get_expr_name_for_runtime_message(sim->m_v);
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -229,9 +229,6 @@ private:
             std::string result = base_name + "(";
             for (size_t i = 0; i < array_item->n_args; i++) {
                 ASR::expr_t* idx_expr = array_item->m_args[i].m_right;
-                if (idx_expr == nullptr) {
-                    return "";
-                }
                 std::string index_name = get_expr_name_for_runtime_message(idx_expr);
                 if (index_name.empty()) {
                     return "";
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -228,11 +228,11 @@ private:
 
             std::string result = base_name + "(";
             for (size_t i = 0; i < array_item->n_args; i++) {
-                ASR::array_index_t& idx = array_item->m_args[i];
-                if (idx.m_left != nullptr || idx.m_right == nullptr || idx.m_step != nullptr) {
+                ASR::expr_t* idx_expr = array_item->m_args[i].m_right;
+                if (idx_expr == nullptr) {
                     return "";
                 }
-                std::string index_name = get_expr_name_for_runtime_message(idx.m_right);
+                std::string index_name = get_expr_name_for_runtime_message(idx_expr);
                 if (index_name.empty()) {
                     return "";
                 }
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -261,18 +261,24 @@ private:
     void generate_unallocated_array_runtime_error(llvm::Value* is_not_allocated,
                                                   ASR::expr_t* target_expr) {
         std::string target_expr_name = get_expr_name_for_runtime_message(target_expr);
-        if (target_expr_name.empty()) {
-            target_expr_name = "LHS";
+        if (!target_expr_name.empty()) {
+            llvm::Value* target_expr_name_llvm = LCompilers::create_global_string_ptr(
+                context, *module, *builder, target_expr_name);
+            llvm_utils->generate_runtime_error(
+                is_not_allocated,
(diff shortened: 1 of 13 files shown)


jassem-manita:
@certik what do you think?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -214,6 +214,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            if (base_name.empty()) {
+                return "";
+            }
+
+            std::string result = base_name + "(";
+            for (size_t i = 0; i < array_item->n_args; i++) {
+                ASR::array_index_t& idx = array_item->m_args[i];
+                if (idx.m_left != nullptr || idx.m_right == nullptr || idx.m_step != nullptr) {
+                    return "";
+                }
```
kmr-srbh: I am not sure if we reach this `if` statement. For an `ArrayItem`, I think we always have a `m_right` value. I know you meant to write a recursive function, but I think the return would always be from the one outside the loop? Please let me know if I am missing something.

HarshilShah1804 (approved):
Looks good.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -214,6 +214,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                std::string index_name = get_expr_name_for_runtime_message(idx.m_right);
+                if (index_name.empty()) {
+                    return "";
+                }
```
kmr-srbh, line 235: Same as https://github.com/lfortran/lfortran/pull/11145/changes#r3148020321.

tests/reference/run-allocatable_component_unnamed_unallocated_01-9346375.stderr
```diff
@@ -1,5 +1,5 @@
-runtime error: Array is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.
+runtime error: Array 'x(1)%arr' is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.
   --> tests/../integration_tests/allocatable_component_unnamed_unallocated_01.f90:10:5
    |
 10 |     x(1)%arr = [1, 2, 3]
-   |     ^^^^^^^^ LHS not allocated here
+   |     ^^^^^^^^ 'x(1)%arr' not allocated here
```
kmr-srbh: I feel naming in the secondary label is probably redundant since we already have it in the primary label?

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh, line 207: We are working with the ASR here, so I think this function should go inside `asr_utils.cpp`. You can make the changes in a different PR.

kmr-srbh (commented):
Thanks for working on this Jassem! Just a few questions and suggestions, otherwise this looks good.

src/libasr/codegen/asr_to_llvm.cpp
jassem-manita: >Please let me know if I am missing something.

You're definitely right. I overcomplicated it for this PR case

jassem-manita:
@kmr-srbh what do you think

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -227,6 +258,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         return "";
     }
 
+    void generate_unallocated_array_runtime_error(llvm::Value* is_not_allocated,
+                                                  ASR::expr_t* target_expr) {
+        std::string target_expr_name = get_expr_name_for_runtime_message(target_expr);
+        if (target_expr_name.empty()) {
+            target_expr_name = "LHS";
+        }
```
kmr-srbh: I think this will cause error messages to print like `Array 'LHS' is not allocated...` when we cannot get the expression name. That is ambiguous since a user code can declare a variable of name `LHS`. For cases when we don't know the name, I think just saying `Array is not allocated...` as before is better.

jassem-manita: Done.

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: Can `index_name` be empty sometimes? If the function `get_expr_name_for_runtime_message()` encounters an expression that an `ArrayItem` index can contain, but does not implement for it, then I think yes. In that case, we might have incorrect array item strings like `arr()` or `arr(var, )`. Could you please check if that is possible or we already handle it?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -214,6 +214,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            if (base_name.empty()) {
+                return "";
+            }
+
+            std::string result = base_name + "(";
+            for (size_t i = 0; i < array_item->n_args; i++) {
+                ASR::expr_t* idx_expr = array_item->m_args[i].m_right;
+                if (idx_expr == nullptr) {
+                    return "";
+                }
```
kmr-srbh: As I earlier said, I think `idx_expr` cannot be `nullptr`.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -214,6 +214,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            ASR::IntegerConstant_t* i = ASR::down_cast<ASR::IntegerConstant_t>(expr);
+            return std::to_string(i->m_n);
+        }
+
+        if (ASR::is_a<ASR::ArrayItem_t>(*expr)) {
+            ASR::ArrayItem_t* array_item = ASR::down_cast<ASR::ArrayItem_t>(expr);
+            std::string base_name = get_expr_name_for_runtime_message(array_item->m_v);
+            if (base_name.empty()) {
+                return "";
+            }
```
kmr-srbh, line 227: Same question as https://github.com/lfortran/lfortran/pull/11145/changes#r3210127366. I think this cannot be empty.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -227,6 +258,29 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                is_not_allocated,
+                "Array '%s' is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.",
+                {LLVMUtils::RuntimeLabel("not allocated here", {target_expr->base.loc}, {})},
+                infile,
+                location_manager,
+                target_expr_name_llvm);
+        } else {
+            llvm_utils->generate_runtime_error(
+                is_not_allocated,
+                "Array is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.",
```
kmr-srbh, line 274: This is good.

src/libasr/codegen/asr_to_llvm.cpp
jassem-manita: i checked 


```console
lfortran fix-runtime-lhs-name ❯ ./src/bin/lfortran /tmp/alloc_comp_binop_idx.f90; echo $?
runtime error: Array is not allocated. Allocate it manually or use the '--realloc-lhs-arrays' option to allocate it automatically.
  --> /tmp/alloc_comp_binop_idx.f90:12:5
   |
12 |     x(i+1)%arr = [1, 2, 3]
   |     ^^^^^^^^^^ not allocated here
1
```
ill be working on x(i+1) next

src/libasr/codegen/asr_to_llvm.cpp
jassem-manita: we need empty() to safely fallback if we can’t fully format the base expression, we return "" instead of partial/invalid name.

jassem-manita:
@kmr-srbh really appreciate the quick reply

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: Thanks for the example. This shows that we are currently not handling not all expressions inside `ArrayItem`'s indices, which is fine for now. I think this is right to do now.

kmr-srbh (approved):
Thanks for sharing an example. The implementation makes sense now. I think this looks good to merge.

PR #11148, opened 2026-04-18
Add fuse_array_reduction pass to eliminate temporary arrays

certik:
Fuse scalar reductions (maxval, minval, sum) over element-wise array expressions into a single loop, eliminating heap-allocated temporaries.

Before this pass, an expression like:
    delta = maxval(abs(xnew(1:n,1:n) - x(1:n,1:n)))
would create 2 temporary arrays (~720KB each) and 3 separate loops.

After this pass, it becomes:
    acc = -huge(0d0)
    do j = 1, n
      do i = 1, n
        acc = max(acc, abs(xnew(i,j) - x(i,j)))
      end do
    end do
    delta = acc

The pass runs before array_struct_temporary and only when --fast is enabled. On the Jacobi benchmark (300x300 grid, ~57K iterations), this reduces runtime from 6.2s to 4.3s (1.44x speedup).


certik:
Right now this "pattern matches" this kind of an expression. We should rather figure out a design that does this for any array operation. So for now I'll leave this as a draft, as a proof of concept.

certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/11156.

PR #11161, opened 2026-04-19
fix: throw error when non-character arguments are passed to `Len` Intrinsic

### LINKED ISSUE #9490: Handle `len()` intrinsic inside select type statements for polymorphic arguments
Related to https://github.com/lfortran/lfortran/issues/9328, https://github.com/lfortran/lfortran/issues/3689

MRE:
```
program temp
    implicit none

    call print_any("  Hello World  ")
    contains
    subroutine print_any(generic)
        class(*), intent(in) :: generic
        select type (generic)
        type is (character(len=*))
            print *, len(generic)
        end select        
    end subroutine print_any
end program temp
```

Main:
```
$ gfortran a.f90 & ./a.out
[1] 155686
          15

$ lfortran a.f90

File "/home/yash/Desktop/lfortran/lfortran/src/libasr/../libasr/asr_base_visitor.h", line 165, in void LCompilers::ASR::visit_expr_t<LCompilers::ASRToLLVMVisitor>(LCompilers::ASR::expr_t const&, LCompilers::ASRToLLVMVisitor&)
    case exprType::StringLen: { v.visit_StringLen((const StringLen_t &)x); return; }
  File "/home/yash/Desktop/lfortran/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 9737, in LCompilers::ASRToLLVMVisitor::visit_StringLen(LCompilers::ASR::StringLen_t const&)
    tmp = get_string_length(x.m_arg); // `int64` value.
AssertFailed: ASR::is_a<ASR::String_t>(* ASRUtils::extract_type(exp_type))
```

YashNagda17:
Fixes #9490

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -12393,20 +12393,32 @@ public:
 
     ASR::asr_t* create_StringLen_from_expr(ASR::expr_t* v, ASR::ttype_t* type, const Location& loc) {
         ASR::expr_t* len_compiletime = nullptr;
-        if( !ASRUtils::is_character(*ASRUtils::expr_type(v)) ) {
-            diag.add(Diagnostic(
-                "Unexpected args, Len expects (char) as arguments",
-                Level::Error, Stage::Semantic, {Label("", {loc})}));
-            throw SemanticAbort();
+        ASR::ttype_t* v_type = ASRUtils::expr_type(v);
+        if( !ASRUtils::is_character(*v_type) ) {
+            if (ASRUtils::is_unlimited_polymorphic_type(v_type)) {
+                diag.add(Diagnostic(
+                    "Character intrinsic 'len' with class(*) argument must be inside " 
+                    "'type is (character(len=*))' guard", Level::Error, Stage::Semantic, {
+                        Label("invalid use of character intrinsic with polymorphic argument",
+                        {v->base.loc})
+                    }));
+                throw SemanticAbort();
+            } else {
+                diag.add(Diagnostic(
+                    "Argument of 'len' intrinsic must be CHARACTER, found " + 
+                    ASRUtils::type_to_str_fortran_expr(v_type, v) + " instead.", 
+                    Level::Error, Stage::Semantic, {Label("", {v->base.loc})}));
+                throw SemanticAbort();
+            }
         }
-        if( ASRUtils::is_array(ASRUtils::expr_type(v)) ) {
+        if( ASRUtils::is_array(v_type) ) {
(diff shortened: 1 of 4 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -12393,6 +12393,12 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 
     ASR::asr_t* create_StringLen_from_expr(ASR::expr_t* v, ASR::ttype_t* type, const Location& loc) {
         ASR::expr_t* len_compiletime = nullptr;
+        if( !ASRUtils::is_character(*ASRUtils::expr_type(v)) ) {
+            diag.add(Diagnostic(
+                "Unexpected args, Len expects (char) as arguments",
+                Level::Error, Stage::Semantic, {Label("", {loc})}));
```
HarshilShah1804: I think we can also add the found type of the argument in the error message.

HarshilShah1804 (commented):
I think it is fine. Just a small suggestion. Then we can merge it.

src/lfortran/semantics/ast_common_visitor.h
YashNagda17: Yes, it can be done,

Current format follows the same format as we throw for other arg type mismatches. They are mostly located in `intrinsic_function_registry_util.h`. Here's an example at line 14111 in the same file:


https://github.com/lfortran/lfortran/pull/11161/changes#diff-dbb410d8367953a79cc0b140c516bbe133e816d9f5c44360a76a030397a7fda5L14111

YashNagda17: I am honestly not sure here, I followed the same format as seen elsewhere... Please let me know if it's fine or should I change it to show input argument dtypes as well...

certik: Every time we add a new error message, let's take time to make it perfect. We have utilities that can print a Fortran type, so just print the type of `ASRUtils::expr_type(v)` as Fortran. Also you are highlighting the whole len() expression, instead only highlight the actual argument.

HarshilShah1804: Agreed.

YashNagda17: Ohk, got it
So, I shall work upon two changes here:

1. Showing the user-input dtypes in error message
2. Highlighting only the variable rather than expression

Also, I feel another refactoring work for next PRs can be applying the same logic for other error messages in same format as well, whom I referred to in the beginning. This shall make all of them more verbose & consistent

certik (approved):

PR #11163, opened 2026-04-19
Fix descriptor OOB write in sequence association

certik:
When passing a multi-dimensional assumed-size array element via sequence association to a 1D assumed-size dummy (e.g., call callee(arr(1, idx)) where callee expects arr(*)), the codegen allocated a 1D descriptor but wrote dimension descriptors for all sliced dimensions of the source section, causing an out-of-bounds stack write and a Bus error at runtime.

The fix pre-counts the number of sliced dimensions (target_rank) before allocating the intermediate section descriptor, and creates a descriptor type with enough room for all sliced dimensions. A bitcast is added for the store to handle typed-pointer LLVM builds where the section descriptor type may differ from the callee's expected descriptor type.

An integration test (array_section_24) is added to cover this case.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3208,6 +3208,7 @@ RUN(NAME array_section_20 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_section_21 LABELS gfortran llvm)
 RUN(NAME array_section_22 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_section_23 LABELS gfortran llvm)
+RUN(NAME array_section_24 LABELS gfortran llvm EXTRA_ARGS --legacy-array-sections)
 
 RUN(NAME nested_vars_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray)
 RUN(NAME nested_vars_02 LABELS gfortran llvm)
--- a/integration_tests/array_section_24.f90
+++ b/integration_tests/array_section_24.f90
@@ -6,31 +6,46 @@ program array_section_24
     integer :: i, j
 
     ! Fill with known values: arr(i,j) = i + 10*j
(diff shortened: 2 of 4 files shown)


certik:
This fix works, but it's not the right fix for two reasons:

* it leaks the legacy array section handling into general code
* it adds bitcast in LLVM which should have been handled in ASR as a cast node

More details:

Here's the full picture:

**The function is used in general**, not just `--legacy-array-sections`. Any `Associate` where the value is an `ArraySection` goes through it — that includes normal pointer association like `ptr => arr(1:n, :)`.

**How the rank mismatch arises**: In `pass_array_by_data` (line 858), when the section has more sliced dimensions than the callee's parameter rank, `set_allocation_size` returns `false` (because `allocate_dims.size() != target_n_dims`), so it falls through to create a raw `Associate` with mismatched ranks (e.g., 1D target ← 2D `ArraySection`). This only happens with `--legacy-array-sections` sequence association.

**On the rest of the changes**: For non-legacy code, the old `target_type` was derived from `duplicate_type_with_empty_dims` (preserving the target's rank). The new code builds `target_type` from `make_Array_t_util` using `target_rank` (counted from sliced dims). When ranks match (always true without `--legacy-array-sections`), these should be equivalent — but it's a different construction path, so it introduces unnecessary risk for non-legacy code.

**On the BitCast itself**: You're right that this violates the architecture — type coercion belongs in ASR, not in the LLVM backend. The real bug is that `pass_array_by_data` emits an `Associate` with mismatched ranks and leaves the backend to paper over it with a BitCast. The proper fix would be upstream in the pass — e.g., linearizing the multi-dimensional section into a 1D section before creating the `Associate`, so by the time the LLVM backend sees it, ranks already match and no BitCast is needed.

In summary: the OOB write was a real bug, but the fix is in the wrong layer.

certik:
This is a minimal fix. To create a better fix would require to add or repurpose some ASR nodes, which is more complex and we can do it later, here is the issue for it: https://github.com/lfortran/lfortran/issues/11172.

PR #11181, opened 2026-04-20
A few fixes for Caffeine

certik:


certik:
Once this is in, I have a couple more fixes on top.

bonachea:
Wow that's alot of bug fixes!
Thanks for all the great work on this!

certik:
I am not quite sure how it happened that we required all these fixes, but they all look good and going forward we'll test caffeine at the CI, so it shouldn't break anymore.

PR #11190, opened 2026-04-20
ci: add caffeine third-party test (macOS)

certik:
Build GASNet (bitbucket.org/certik/gasnet, build branch, 659f6a1c064622968667a8d123515a43a4d30b6c) and then build and run caffeine (github.com/certik/caffeine, build4 branch, 0529b3c) via 'pixi r -e lfortran test'.

Gated to macOS since both pixi.toml files only declare osx-arm64.


ci/test_third_party_codes.sh
```diff
@@ -85,6 +85,39 @@ fi
...
+  git checkout 659f6a1c064622968667a8d123515a43a4d30b6c
+  pixi install
+  pixi r build
+  pixi r install
+  cd ..
+
+  # Now build and test caffeine with LFortran
+  git clone -b build4 https://github.com/certik/caffeine.git
+  cd caffeine
+  git checkout 0529b3c
```
bonachea: Is there a reason to hard-code a version hash here?

I see this `build4` branch is already > 30 commits behind `caffeine:main` and this gap notably includes the PR that [made changes to `install.sh` to support LFortran](https://github.com/BerkeleyLab/caffeine/pull/316).

Is the intent that you'll periodically update this branch to mirror the upstream Caffeine `main`?  It might be less work to simply pull Caffeine main and run `install.sh` as documented.. That way you also don't have to mess around installing GASNet, as the install script automates that too.

bonachea (changes requested):
Thanks for looking into this! 

I strongly support an integration test to ensure things keep working as LFortran advances, although I have some concerns about keeping up with Caffeine advances.

ci/test_third_party_codes.sh
certik: Yes, I would like to simplify it, I just didn't succeed locally without my patch, but maybe at the CI, possibly on linux instead of macos, it will work.

bonachea: Note there are also some fixes in `caffeine:main` (not yet in a tagged release) to the Homebrew support for `install.sh`. However use of Homebrew is _entirely_ optional: if you don't want to use Homebrew you can instead manually install the basic package dependencies (Fortran/C compilers, pkg-config, make, git, fpm, curl). 

Either way the `install.sh` script will still download and install GASNet for you (using curl, not Homebrew).

bonachea:
`caffeine:main` is no longer passing `-g` to LFortran (as of `9e86eb7b` on Apr 13), so this is also addressed in the upstream `install.sh`

certik:
Ok, got it to compile, all tests pass, it adds 1m 39s to our our CI time, which is fine for now.

certik:
Closing in favor of https://github.com/lfortran/lfortran/pull/11195.

@bonachea now when we have it up and running, we can send PRs to simplify the build and to make it easier to update.

Issue #11143, opened 2026-04-18
Compiling CFGIO package


ivan-pi:
Package: https://github.com/pkgpl/cfgio (commit: ec2f18e)

Steps to reproduce (in Bash):

```
git clone https://github.com/pkgpl/cfgio && cd cfgio
cd src && make FC=`which lfortran`
cd ../test && make FC=`which lfortran`
/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran  -c test_write.f90 -I../include
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000be147
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000a6313
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000b11f7
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1004045c7
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x100401823
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000ca77b
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000cabdb
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1000fc41b
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x10013afc7
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1010ceb73
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x1010d0057
  Binary file "/opt/homebrew/Cellar/micromamba/2.5.0_4/bin/lfortran", local address: 0x10114c3af
LCompilersException: read_string: String is too short for deserialization.
make: *** [Makefile:15: test_write.o] Error 1
```

ivan-pi:
[...]

PR #11194, opened 2026-04-20
fix: handle complex return type functions for runtime polymorphic function calls

### LINKED ISSUE #11193: CodeGenError with complex return type function in runtime polymorphic function calls
MRE
```fortran
module functions_59_mod
implicit none

type :: t
contains
   procedure :: get_c_impl
end type

contains

function get_c_impl(this) result(v)
   class(t), intent(in) :: this
   complex :: v

   v = cmplx(1.0, 2.0)
end function get_c_impl

subroutine get_c(this, val)
   class(t), intent(in) :: this
   complex, intent(out) :: val

   val = this%get_c_impl()
end subroutine get_c

end module functions_59_mod


program functions_59
   use functions_59_mod
   implicit none

   type(t) :: x
   complex :: v

   call get_c(x, v)

   if (abs(real(v) - 1.0) > 1e-6 .or. abs(aimag(v) - 2.0) > 1e-6) error stop

end program functions_59
```

LFortran
```console
  store <2 x float> %4, %complex_4* %val, align 8
 %complex_4 = type <{ float, float }>
  store <2 x float> %4, %complex_4* %val, align 8
 %complex_4 = type <{ float, float }>


```

HarshilShah1804:
Fixes #11193 
Towards #11143

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -24580,6 +24580,21 @@ public:
             vtable_ptr, struct_api->struct_vtab_function_offset[struct_sym][proc_sym_name]));
         fn = llvm_utils->CreateLoad2(fnPtrTy, fn);
         tmp = builder->CreateCall(fnTy, fn, args);
+
+        if (func->m_return_var) {
+            ASR::ttype_t *return_var_type0 = EXPR2VAR(func->m_return_var)->m_type;
+            if (is_a<ASR::Complex_t>(*return_var_type0)) {
+                int a_kind = down_cast<ASR::Complex_t>(return_var_type0)->m_kind;
+                if (a_kind == 4) {
+                    llvm::Type* type_fx2 = FIXED_VECTOR_TYPE::get(llvm::Type::getFloatTy(context), 2);
+                    llvm::AllocaInst *p_fx2 = llvm_utils->CreateAlloca(type_fx2, nullptr, "complex_ret_tmp");
+                    builder->CreateStore(tmp, p_fx2);
+                    tmp = builder->CreateBitCast(p_fx2, complex_type_4->getPointerTo());
+                    tmp = llvm_utils->CreateLoad2(complex_type_4, tmp);
+                }
+            }
+        }
+
         return;
     }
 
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -24580,6 +24580,33 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    } else if (compiler_options.platform == Platform::macOS_ARM) {
+                        // pass - already {float, float}
+                    } else {
+                        // tmp is <2 x float>, convert to {float, float}
+                        llvm::Type* type_fx2 = FIXED_VECTOR_TYPE::get(llvm::Type::getFloatTy(context), 2);
+                        llvm::AllocaInst *p_fx2 = llvm_utils->CreateAlloca(type_fx2, nullptr, "complex_ret_tmp");
+                        builder->CreateStore(tmp, p_fx2);
+                        tmp = builder->CreateBitCast(p_fx2, complex_type_4->getPointerTo());
+                        tmp = llvm_utils->CreateLoad2(complex_type_4, tmp);
+                    }
```
certik: Why are the types different on all platforms? This is for our internal ABI (right?), so we should use the same types, to simplify things.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -24549,6 +24580,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+        if (func->m_return_var) {
+            ASR::ttype_t *return_var_type0 = EXPR2VAR(func->m_return_var)->m_type;
+            if (is_a<ASR::Complex_t>(*return_var_type0)) {
+                int a_kind = down_cast<ASR::Complex_t>(return_var_type0)->m_kind;
+                if (a_kind == 4) {
+                    llvm::Type* type_fx2 = FIXED_VECTOR_TYPE::get(llvm::Type::getFloatTy(context), 2);
+                    llvm::AllocaInst *p_fx2 = llvm_utils->CreateAlloca(type_fx2, nullptr, "complex_ret_tmp");
+                    builder->CreateStore(tmp, p_fx2);
+                    tmp = builder->CreateBitCast(p_fx2, complex_type_4->getPointerTo());
```
certik, line 24592: I fundamentally do not like this cast here, per my previous comment. Why do we need to cast --- usually it means the ASR types do not agree somehow?

PR #11197, opened 2026-04-21
Merge generic interfaces imported from multiple modules

### LINKED ISSUE #11160: Matching generic procedure not found
The code 
```
module submesh_m
  implicit none
  type smesh_t
    integer :: ndim = 0
  end type smesh_t
  interface delete
    module procedure delete_smesh
  end interface delete
contains
  subroutine delete_smesh ( smesh )
    type(smesh_t), intent(out) :: smesh
  end subroutine delete_smesh
end module submesh_m
module intmesh_m
  implicit none
  type imesh_t
    integer :: ndim = 0
  end type imesh_t
  interface delete
    module procedure delete_imesh
  end interface delete
contains
  subroutine delete_imesh ( imesh )
    type(imesh_t), intent(out) :: imesh
  end subroutine delete_imesh
end module intmesh_m
module eltree_basic_m
  use submesh_m
  use intmesh_m
  implicit none
  type eltree_t
    type(smesh_t) :: smesh
    type(imesh_t) :: imesh
  end type eltree_t
contains
  subroutine delete_eltree ( eltree )
    type(eltree_t), intent(inout) :: eltree
        call delete ( eltree%smesh )
        call delete ( eltree%imesh )
  end subroutine delete_eltree
end module eltree_basic_m
```

gives

```
semantic error: Arguments do not match for any generic procedure, delete
  --> t.f90:41:9
   |
41 |         call delete ( eltree%imesh )
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

Note, if the first two modules are combined into a single module the error disappears.

certik:
When two modules define a generic interface (or user-defined operator) with the same name and both are brought into scope via `use`, the Fortran standard requires the specific procedures to be merged into a single generic at the use-point (F2018 C1515). Previously `import_all` skipped the second import whenever a symbol with that name already existed, so only the specifics from the first module were visible; calling the generic with an argument matching a specific from the second module produced 'Arguments do not match for any generic procedure'.

This change detects the GenericProcedure/CustomOperator-with-same-name case in `import_all` and delegates merging to the existing `process_generic_proc_custom_op` helper already used by the `use, only:` path, so both `use` forms behave the same.

Adds integration_tests/generic_name_07.f90.

Fixes #11160.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2047,6 +2047,7 @@ RUN(NAME generic_name_04 LABELS gfortran llvm)
 RUN(NAME generic_name_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME generic_name_06 LABELS gfortran llvm)
 RUN(NAME generic_name_07 LABELS gfortran llvm)
+RUN(NAME generic_name_08 LABELS gfortran llvm)
 
 RUN(NAME operator_overloading_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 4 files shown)


certik:
There is a way to simplify this: https://github.com/lfortran/lfortran/pull/11198.

certik:
The simpler way doesn't work, so I'll merge this one.

PR #11200, opened 2026-04-21
athena fixes 06

HarshitaKalani:
with this, athena compiles 76%
```
[ 76%] Compiling...
code generation error: Assignment target must be a variable or subscripted expression, got: type 93

```


src/libasr/pass/array_op.cpp
```diff
@@ -273,6 +273,28 @@ class FixTypeVisitor: public ASR::CallReplacerOnExpressionsVisitor<FixTypeVisito
...
+
+    void replace_StructInstanceMember(ASR::StructInstanceMember_t* x) {
+        ASR::BaseExprReplacer<CleanupDegenerateArraySection>::replace_StructInstanceMember(x);
+        if (ASRUtils::is_array(x->m_type) &&
+            !ASRUtils::is_array(ASRUtils::expr_type(x->m_v)) &&
+            !ASRUtils::is_array(ASRUtils::symbol_type(x->m_m))) {
+            x->m_type = ASRUtils::extract_type(x->m_type);
+        }
+    }
+};
```
certik, line 296: What is this visitor doing? I can't figure it out by reading the source code. Is it replacing some types in structs?

Is the ASR invalid before this and how did it get invalid?

certik (approved):
I think it's ok to merge, but if you could please answer my comment that would be great.

src/libasr/pass/array_op.cpp
HarshitaKalani: This visitor is a cleanup step after loop generation. Earlier in the array_op pass, array expressions are rewritten into scalar accesses inside loops (e.g., arr to arr(i)). During that process, some nodes still carry old array information even though they now represent scalars.

This visitor fixes those leftovers:
- ArraySection cleanup:
If something still looks like an array slice (ArraySection), but its base is no longer an array (it’s now a scalar like arr(i)), then the slice wrapper is meaningless, so it just removes it.
- Struct member cleanup:
If a struct member is marked as an array type, but neither the base object nor the member is actually an array anymore, it removes that incorrect array type info.

The ASR is valid before the array_op pass starts. It becomes temporarily inconsistent during the transformation, because changes (like converting arrays to scalars) don’t update all type info at once.
This visitor restores consistency.

Without this cleanup, stale array types can leak into codegen and cause failures (especially with struct members).

PR #11227, opened 2026-04-24
fmt: drop redundant labeled CONTINUE terminating DO loops

### LINKED ISSUE #11223: Unnecessary `continue` line remains after `lfortran fmt`
I am using lfortran 0.63.0. Consider doconv.f, an example program I want to convert to modern Fortran:

          PROGRAM DOCONV
          INTEGER I
          DO 1 I = 1, 10
              PRINT *, I
        1 CONTINUE
          END

I run `lfortran --fixed-form fmt doconv.f` and get the following:

    program doconv
    integer :: i
    do i = 1, 10
        print *, i
        1 continue
    end do
    end program doconv

This is nearly what I want, aside from the `1 continue` line, which serves no purpose after the loop was converted to use `end do`.

certik:
When the fixed-form tokenizer rewrites the legacy
`DO <label> ... <label> CONTINUE` idiom into a structured `do ... end do`, the labeled `CONTINUE` remains as a separate statement in the AST and `lfortran fmt` prints a redundant `<label> continue` line right before `end do`.

Fix this in the AST-to-source pretty-printer only: collect labels referenced as branch targets in the current program unit (GO TO, computed GO TO, ASSIGN, arithmetic IF, FORMAT labels in PRINT/READ, and I/O err=/end=/eor=/fmt= kwargs). When visiting a DoLoop, omit a trailing labeled `continue` whose label has no such reference. Labels are unit-scoped, so the collector does not descend into nested CONTAINS subprograms; format_unit_body saves and restores the set across program-unit boundaries.

If the label is referenced (e.g. `GO TO 1` used as a cycle), the labeled `continue` is preserved so semantics are not changed.

Fixes #11223


certik:
The original approach (modify the fixed-form tokenizer to drop the `continue` token and let the already-emitted `TK_LABEL` attach to the synthesized `end_do`) failed for two reasons:

**1. The label does not actually propagate to ASR through `end_do`.**

The grammar in `parser.yy:2231` accepts `TK_LABEL KW_END_DO`:
```
enddo
    : KW_END_DO
    | TK_LABEL KW_END_DO
```
…but the action discards the label — it doesn't feed it into the `DoLoop_t` AST node's `m_label` or any end-of-loop slot. Even if it did reach the AST, `ast_body_visitor.cpp`'s `visit_DoLoop` doesn't forward `m_label` into the ASR `DoLoop`. So the token sequence `TK_LABEL(1), end_do, \n` results in the label being silently *dropped* — not moved. In my test I observed exactly that: a `go to 1` referencing the label ended up with no basic block to target, producing the LLVM verification error `Referring to a basic block in another function`.

So my "move the label to `end_do`" claim was false: it was really "throw the label away".

**2. Even if the label *did* attach to `end_do`, it would silently change pre-F90 semantics.**

`GO TO <loop-terminator-label>` in classic Fortran branches to the terminator statement *inside* the loop's extent, causing the loop to iterate (CYCLE semantics). If the label instead sits on `end_do`, the same `GO TO` exits the loop. That's a silent behavior change — exactly the kind of "hack" the request wanted to avoid. You flagged this yourself in the prompt phrasing; the rubber-duck critique confirmed it's a real concern independent of (1).

**Why the formatter-only approach is sound instead:**

- The tokenizer/parser/ASR keep producing the faithful AST (labeled `continue` inside the loop body, where branch-to-terminator semantics are preserved).
- Only the pretty-printer elides the trailing labeled `continue`, and only when a label-reference scan over the program unit proves the label is dead. If it's referenced, the `continue` stays.
- This is a pure source-to-source rewrite with a correctness condition; nothing downstream (semantics, codegen) is touched.

So the original fix traded a cosmetic improvement for silent label loss and a semantic change — not acceptable. The formatter-level fix provides the same cosmetic benefit with a proof obligation (label unreferenced) that makes it safe.

certik:
This is a better fix:https://github.com/lfortran/lfortran/pull/11231.

PR #11231, opened 2026-04-25
fix: drop redundant CONTINUE terminator on legacy DO <label>

certik:
Towards #11223.


certik:
I still don't like it, I am improving it.

certik:
I think the right answer is that the label should just be on the "DoLoop" node and any goto should just use it, and in AST->ASR we correctly transform it into GoToTarget as if it was pointing to a label. We need to update both fixed-form and free-form parser/tokenizers to do the same.

PR #11232, opened 2026-04-25
CI: Third party code updates

bonachea:
Update third-party testing in Exhaustive checks to use more recent, upstream versions of:

1. Caffeine
2. Assert
3. Julienne

In addition, the Caffeine build is updated to:

* Test on both Linux and macOS

* Use the Caffeine install script, which takes care of building GASNet

* Ensure we perform multi-image (paralllel) testing, as single-image (serial) execution trivializes much of Caffeine's functionality.

### DIFF
--- a/ci/test_third_party_codes.sh
+++ b/ci/test_third_party_codes.sh
@@ -85,38 +85,43 @@ fi
 TMP_DIR=$(mktemp -d)
 cd "$TMP_DIR"
 
-if [[ "$(uname)" == "Darwin" ]]; then
 time_section "🧪 Testing caffeine" '
-  which lfortran
-  lfortran --version
+  git clone -b main https://github.com/BerkeleyLab/caffeine.git
+  cd caffeine
 
-  # Install pixi (needed to drive gasnet and caffeine builds)
-  if ! command -v pixi >/dev/null 2>&1; then
-    curl -fsSL https://pixi.sh/install.sh | bash
-    export PATH="$HOME/.pixi/bin:$PATH"
-  fi
-  pixi --version
+  micromamba install -c conda-forge fpm=0.12.0
 
-  # Build GASNet first (caffeine depends on it)
-  git clone -b build https://github.com/certik/gasnet.git
-  cd gasnet
-  git checkout 2949970115991c3eb71ba7c16edc87807d687e97
-  pixi install
-  pixi r build
-  pixi r install
-  cd ..
+  # To debug https://github.com/lfortran/lfortran/issues/7732:
+  which fpm
+  realpath $(which fpm)
+  ls -l $(dirname $(realpath $(which fpm)))/../lib
+  ls -l $CONDA_PREFIX/lib
+  fpm --version
+
+  export CC=clang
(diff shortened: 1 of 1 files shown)


ci/test_third_party_codes.sh
```diff
@@ -239,10 +237,9 @@ time_section "🧪 Testing Formal" '
 time_section "🧪 Testing Julienne" '
   git clone https://github.com/BerkeleyLab/julienne.git
   cd julienne
-  export PATH="$(pwd)/../src/bin:$PATH"
   micromamba install -c conda-forge fpm
 
-  git checkout a75b5a831e303315304db52ec9dd70c9badc08cd
+  git checkout 3.6.2
```
certik: I think this is fine. There is a danger if you accidentally changed the tag to point to a different commit in julienne, then this could break. But assuming this will not happen, then it should not break.

I researched it sometime ago to see what the easiest way to "assert" a specific commit is, but the best approach was to just check out that specific commit. You could do some complicated if statements, but it's not worth it I think.

ci/test_third_party_codes.sh
```diff
@@ -85,62 +85,60 @@ fi
...
+
+  export CC=clang
+  export CXX=clang++
+  echo PATH="$PATH"
+
+  for tool in ${FC} ${CC} ${CXX} fpm ; do
+    if command -v $tool > /dev/null 2>&1 ; then
+       ( echo ; set -x ; w=$(which $tool) ; ls -al $w ; ls -alhL $w ; $tool --version )
+    fi
+  done
```
certik, line 110: This is just for debugging, but not needed for the CI run to work, correct?

certik (approved):
Looks great! Thanks for cleaning it up. I think it works, even on linux. I checked the duration time, it's about the same, so I think this will work.

ci/test_third_party_codes.sh
bonachea: `3.6.2` is a release tag, we will never change that.

We could use the corresponding hash instead if that makes you more comfortable, but this is more self-documenting.

ci/test_third_party_codes.sh
bonachea: Correct, this is entirely for debugging, mostly so we can see the details of the C/C++ compilers.

ci/test_third_party_codes.sh
bonachea: Resolved in b98e1c59f

ci/test_third_party_codes.sh
bonachea: Comment added in b98e1c59f

certik (approved):
Thanks for this. It seems to pass in all our tests again, with LLVM 8-22 on Ubuntu and LLVM 22 on macOS, so I think it's robust enough.

ci/test_third_party_codes.sh
certik: I might have found a solution to this: https://github.com/lfortran/lfortran/pull/11235.

PR #11239, opened 2026-04-26
CI: update fpm to 0.13.0

certik:


ci/test_third_party_codes.sh
```diff
@@ -405,9 +405,9 @@ time_section "🧪 Testing FPM" '
   git clone https://github.com/fortran-lang/fpm.git
```
bonachea: This clones the entire fpm repo, which is rather large and slow.

Since you only care about one commit in a given run, this command retrieves the same working files but results in a working directory that is about 20x smaller and quite a bit faster:

```suggestion
  git clone -b v0.13.0 --depth 1 https://github.com/fortran-lang/fpm.git
```

certik: Nice, thanks for this. Let's try it!

ci/test_third_party_codes.sh
```diff
@@ -402,12 +402,12 @@ time_section "🧪 Compiling POT3D with fortran_mpi" '
...
 ##########################
 time_section "🧪 Testing FPM" '
-  git clone https://github.com/fortran-lang/fpm.git
+  git clone -b v0.13.0 --depth 1 https://github.com/fortran-lang/fpm.git
   cd fpm
   export PATH="$(pwd)/../src/bin:$PATH"
-  git checkout main
   micromamba install -c conda-forge fpm
-  git checkout d0f89957541bdcc354da8e11422f5efcf9fedd0e
+  git checkout v0.13.0
```
certik, line 409: I guess this line is not needed anymore?

bonachea: It's redundant, but should be harmless.
Probably best to remove it for maintainability.

PR #11240, opened 2026-04-26
Initialize rank in fill_malloc_array_details

certik:
When allocating an array via the simple CMO descriptor's malloc path, the dim 0..n-1 fields and offset were set, but the rank field was never written. For module-level allocatable arrays compiled with --separate-compilation, the descriptor lives in .bss (zero-initialized) and the rank therefore remained 0. Subsequent passes that walk over the array's elements (e.g. nullifying allocatable components of every element of a derived-type array) would then see size 1 instead of the real size, leaving later elements with garbage allocatable-component pointers. This manifested, for instance, as a runtime 'already allocated' error on the second allocate(d(2)%a(1)) when format processing left non-zero bytes in the heap region returned by malloc.

Set the rank explicitly in fill_malloc_array_details, mirroring what fill_array_details and fill_dimension_descriptor already do.

Adds integration test separate_compilation_49 that triggers the scenario via a write(*,'(4x)') format followed by allocation of a module-level allocatable array of derived type with allocatable components, called from a program that does not 'use' the module.


certik:
I opened https://github.com/lfortran/lfortran/issues/11242 for the leak detection issue.

PR #11241, opened 2026-04-27
Get all NIST FORTRAN78 test suite to compile

certik:
It does not run yet, but all files now compile to a binary without failing.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -17862,8 +17862,11 @@ public:
 
             // For equivalenced variables (Pointer-typed Var), var_ptr is a
             // pointer to the pointer. Load it to get the actual data address.
+            // Skip this for descriptor arrays (e.g., array section pointers)
+            // which are handled by push_descriptor_array_args.
             if (ASR::is_a<ASR::Var_t>(*val_expr) &&
-                ASR::is_a<ASR::Pointer_t>(*ASRUtils::type_get_past_array(expr_type_full))) {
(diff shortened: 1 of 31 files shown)


HarshitaKalani (approved):
I’ve gone through the changes, they look good. However, it looks like a few tests are already working on main. I guess that's fine.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13630,7 +13641,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            llvm::BasicBlock *cont = llvm::BasicBlock::Create(context,
+                "goto_target.cont");
+            llvm_fn_insert_bb(fn, cont);
+            if (last_bb->getTerminator() == nullptr) {
+                builder->CreateBr(cont);
+            }
+            builder->SetInsertPoint(cont);
+        } else {
+            start_new_block(target);
+        }
```
certik, line 13659: This one is a workaround, I think there is a better way to fix this, but we can do it in the next PR.

certik:
Yes, the individual fixes can be improved, including the tests. But I think this gets us over the finish line and we can use this merge commit as a basis for future work to improve it. I think the fixes are fine, so we can merge it.

src/libasr/codegen/asr_to_llvm.cpp
certik: https://github.com/lfortran/lfortran/issues/11245

PR #11250, opened 2026-04-27
Get most of the NIST Fortran test suite running

certik:
Few tests still fail, but most are running correctly with this PR.


integration_tests/CMakeLists.txt
```diff
@@ -2345,7 +2354,20 @@ RUN(NAME derived_types_120 LABELS gfortran llvm)
...
+        COMMAND lfortran -g -c ${CMAKE_CURRENT_SOURCE_DIR}/derived_types_123.f90
+                -o ${CMAKE_CURRENT_BINARY_DIR}/_probe_g.o
+        RESULT_VARIABLE _g_result
+        OUTPUT_QUIET ERROR_QUIET)
+    if (_g_result EQUAL 0)
+        RUN(NAME derived_types_123 LABELS gfortran llvm EXTRA_ARGS -g GFORTRAN_ARGS -g)
+    endif()
+else()
+    RUN(NAME derived_types_123 LABELS gfortran llvm EXTRA_ARGS -g GFORTRAN_ARGS -g)
+endif()
```
certik: This should be changed, always test with `-g`, or handle it with labels. And NIST shouldn't need `-g` in the first place.

src/libasr/utils.h
```diff
@@ -44,7 +44,7 @@ struct PassOptions {
     bool dump_fortran = false; // For developer debugging
     bool pass_cumulative = false; // Apply passes cumulatively
     bool disable_main = false;
-    bool use_loop_variable_after_loop = false;
+    bool use_loop_variable_after_loop = true;
```
certik: This should be reverted.

certik:
CI passes, now need to clean the PR up.

integration_tests/CMakeLists.txt
certik: Fixed.

certik:
I carefully reviewed every change and I think it's fine now to merge.

PR #11255, opened 2026-04-28
fix: correct ArrayPhysicalCast handling in assignment target

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/11248

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -16530,6 +16530,68 @@ public:
         return fn;
     }
 
+    void emit_string_read_loop(ASR::ttype_t* elem_type, llvm::Type* llvm_elem_type,
+            llvm::Value* section_ptr, llvm::Value* size,
+            llvm::Value* str_src_data, llvm::Value* str_src_len,
+            llvm::Value* iostat, llvm::Value* str_offset,
+            ASR::expr_t* value_expr) {
+        std::string rt_name = "_lfortran_string_read_" +
+            ASRUtils::type_to_str_python_expr(elem_type, value_expr);
+        llvm::Function* str_fn = module->getFunction(rt_name);
+        if (!str_fn) {
+            llvm::FunctionType *ft = llvm::FunctionType::get(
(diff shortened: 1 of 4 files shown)


certik:
I am confused by the select rank test and how it corresponds to the fixes in this PR?

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -16963,15 +17049,21 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                                    llvm::Value* stride_val = llvm::ConstantInt::get(
-                                        llvm::Type::getInt32Ty(context), 1);
-                                    if (ASR::is_a<ASR::Logical_t>(*elem_type)) {
-                                        int a_kind = ASRUtils::extract_kind_from_ttype_t(elem_type);
-                                        llvm::Value* kind_val = llvm::ConstantInt::get(context, llvm::APInt(32, a_kind));
-                                        builder->CreateCall(fn, {section_ptr, size, kind_val, stride_val, unit_val, iostat});
+                                    if (is_string) {
+                                        emit_string_read_loop(elem_type, llvm_elem_type,
+                                            section_ptr, size, str_src_data, str_src_len,
+                                            iostat, str_offset, idl->m_values[0]);
```
certik, line 17055: In main we were handling these reads uniformly.

Now we dispatch and handle strings by inserting a loop in LLVM instead of handling it at runtime. Cannot this be handled uniformly for all types?

certik (approved):
I think this is fine to merge, we can improve upon it later.

src/libasr/codegen/asr_to_llvm.cpp
certik: I think this was the issue:

In the LLVM backend's fast-path for implied-do reads (arr(j), j=start,
end), the descriptor branch had two issues: (1) it didn't load through the
allocatable/pointer descriptor before calling get_pointer_to_data, causing a
type-mismatch GEP assertion, and (2) it used the file-unit array read
function for internal reads where the unit is a string buffer.

PR #11256, opened 2026-04-28
fix: handle select type inside select rank for class(*) assumed-rank arrays

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/11248


HarshitaKalani:
I believe this is unrelated: 
```
609 - format_14_FAST (Failed)
```

HarshilShah1804 (approved):
Looks good to me

certik (approved):
I think this is fine, thanks!

PR #11257, opened 2026-04-28
fix: fix incorrect formatted write calculations

YashNagda17:
fix for flaky `format_14` failure

Previous behavior: This solution is for cases with both TR(Row) and TL(Column) in write statements,
for example, (TR5, TL3, A).
The issue is TR moves the cursor right, but it doesn't write spaces into buffer yet.
So buffer might be empty. Previous logic used buffer contents to find the column value,
which can lead to reading incorrect/ garbage values, and wrong calculations of TL.
This made the formatting incorrect in the target variable.

So, instead of buffer, use current cursor value & use only the part
of the buffer which is already written to calculate the starting value for TL.
So, adding a guardrail check for the calculation of
starting position for TL parameter correctly.


YashNagda17:
#11250 solves this issue. So closing this PR

certik:
I think this commit https://github.com/lfortran/lfortran/commit/fe02f1125cdca9a93df3402d50dea3650e96d8ad fixed it.

PR #11281, opened 2026-04-30
runtime: use minimal widths for list-directed output

certik:
LFortran's runtime print_into_string previously padded list-directed `print *, …` / `write(*,*) …` numeric output to fixed Gw.d-style field widths (`%11d` for INTEGER(4), width 25 for REAL(8), etc.). This produced leading whitespace by default, even with `--print-leading-space` disabled.

The `--print-leading-space` flag is meant to be the only mechanism that introduces an explicit leading space character; it is already correctly gated at the semantics layer and is auto-enabled by `-std=f23` and `-std=legacy`.

Switch the runtime list-directed path to minimal widths: no field padding, no leading whitespace. Items continue to be separated by a single space (default_formatting unchanged). Formatted (Iw.d, Fw.d, Gw.d, …) output is unaffected.

Adds integration test print_15 pinning the new behavior. Updates reference outputs that encoded the old widths.


certik:
@HarshitaKalani let me know if this looks good. I think it looks better without the space and it's only used for `print *` and `write(*,*)`, the formatted output stays as before.

HarshitaKalani:
Sure, it looks great

PR #11284, opened 2026-04-30
bind(c) character in struct fixes

certik:


integration_tests/bindc_inline_char_01.f90
```diff
@@ -0,0 +1,52 @@
...
+program bindc_inline_char_01
+  use iso_c_binding, only: c_ptr, c_loc, c_size_t
+  implicit none
+
+  type :: object_t
+    sequence
+    integer :: i
+    character(len=5) :: actor
+    integer :: j
+  end type
```
bonachea, line 13: The comment doesn't match what the test is doing, object_t is not `bind(c)`

integration_tests/bindc_inline_char_02.f90
```diff
@@ -0,0 +1,52 @@
...
+! Companion to bindc_inline_char_01 (which uses SEQUENCE).
+program bindc_inline_char_02
+  use iso_c_binding, only: c_ptr, c_loc, c_size_t
+  implicit none
+
+  type, bind(c) :: object_t
+    integer :: i
+    character(len=5) :: actor
+    integer :: j
+  end type
```
bonachea, line 13: `bind(c)` prohibits character members with len > 1

> 18.3.1 Interoperability of intrinsic types
> ... If the type is character, the length type parameter is interoperable if and only if its value is one.

bonachea: Also, the kinds need to be `c_char` and `c_int` to be guaranteed interoperability and hence be valid for `bind(c)`

bonachea (changes requested):
I've just started looking at this PR, but already see some concerning changes.

I apologize if my issue sent you off in the wrong direction..

integration_tests/bindc_inline_char_02.f90
bonachea: Accepting this program as written would be an extension to the standard, so it should probably issue a portability warning because I believe other compilers will reject it.

integration_tests/bindc_inline_char_01.f90
bonachea:
```suggestion
! Verify that sequence derived types with character(len>1) members are
! laid out inline (matching the C ABI), so c_loc + memcpy works
! correctly and storage_size reports the full inline byte count.
program bindc_inline_char_01
  use iso_c_binding, only: c_ptr, c_loc, c_size_t
  implicit none

  type :: object_t
    sequence
    integer :: i
    character(len=5) :: actor
    integer :: j
  end type
```

integration_tests/bindc_inline_char_02.f90
bonachea: Confirmed, gfortran rejects the type as written:
```
$ gfortran bindc_inline_char_02.f90
bindc_inline_char_02.f90:11:29:

   11 |     character(len=5) :: actor
Error: Component 'actor' of BIND(C) type at (1) must have length one
```

flang allows it (and passes the test as written), but warns in `-pedantic` mode:
```
$  flang -pedantic bindc_inline_char_02.f90
./bindc_inline_char_02.f90:11:25: portability: A CHARACTER component of an interoperable type should have length 1 [-Wbind-c-char-length]
      character(len=5) :: actor
                          ^^^^^
```

bonachea:
@certik as discussed in our meeting today, this PR should probably be split into the fixes needed for standards conforming implementation of `sequence`, and (optionally) a second PR to add extensions for `bind(c)` derived types.

bonachea:
This PR was superseded by https://github.com/lfortran/lfortran/pull/12161 and https://github.com/lfortran/lfortran/pull/12162

PR #11285, opened 2026-04-30
fix: internal list-directed read for arrays 

### LINKED ISSUE #11261: Reading into an array of double is not supported
Get `Reading into an array of double is not supported` with Godbolt LFortran 0.59.0 for

```fortran
module xread_vec_real_section_mod
implicit none
contains

subroutine read_vec_real(text, xx, nread)
character(len=*), intent(in) :: text
real(kind=dp), intent(out), optional, allocatable :: xx(:)
integer, intent(out), optional :: nread
real(kind=dp), allocatable :: xread(:)
integer :: i, ierr, nread_, max_read_

max_read_ = 4
allocate(xread(max_read_))
nread_ = 0
do i = max_read_, 1, -1
   read(text, *, iostat=ierr) xread(:i)
   if (ierr == 0) then
      nread_ = i
      exit
   end if
end do
if (present(nread)) nread = nread_
if (present(xx)) then
   allocate(xx(nread_))
   xx = xread(:nread_)
end if
deallocate(xread)
end subroutine read_vec_real

end module xread_vec_real_section_mod

program xread_vec_real_section
use xread_vec_real_section_mod, only: read_vec_real, dp
implicit none
real(kind=dp), allocatable :: x(:)
integer :: nread

call read_vec_real("1.5 2.5 3.5", x, nread)
print *, nread
print *, x
end program xread_vec_real_section
```

Jatinagarwal24:
Bug: internal list-directed reads into numeric arrays consumed the entire input string at once, so reads past the available values did not set iostat and left extra elements (e.g., a trailing zero), unlike GFortran.
Fix: route numeric array internal reads through the element-wise string reader with proper descriptor-to-data handling so [iostat] is raised at end-of-record and the read count is correct.
fixes #11261

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -17630,7 +17630,8 @@ public:
                             }
                             llvm::Value* data_ptr = arr_data;
                             if (data_ptr->getType()->isPointerTy() &&
-                                data_ptr->getType()->getPointerElementType()->isStructTy()) {
+                                llvm::cast<llvm::PointerType>(data_ptr->getType())
+                                    ->getElementType()->isStructTy()) {
                                 llvm::Type* arr_type = llvm_utils->get_type_from_ttype_t_util(
                                     x.m_values[i],
                                     ASRUtils::type_get_past_allocatable_pointer(type),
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -17610,7 +17610,46 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                    x.m_values[i],
+                                    ASRUtils::type_get_past_allocatable_pointer(type),
+                                    module.get());
+                                data_ptr = builder->CreateGEP(arr_type, data_ptr,
+                                    { llvm::ConstantInt::get(llvm::Type::getInt32Ty(context), 0),
+                                      llvm::ConstantInt::get(llvm::Type::getInt32Ty(context), 0) });
+                            }
+                            emit_string_read_loop(elem_type, llvm_elem_type,
+                                data_ptr, array_size, str_src_data, str_src_len,
+                                iostat, str_offset, x.m_values[i]);
```
certik, line 17652: Can you explain what this code is doing? It seems complex.

certik:
Isn't this a better fix: https://github.com/lfortran/lfortran/pull/11291 ?

The test is a lot simpler at least.

Jatinagarwal24:
> Isn't this a better fix: #11291 ?
> 
> The test is a lot simpler at least.

Yes i think #11291 is a better fix.

PR #11294, opened 2026-05-01
  fix: handle scalar struct pointer target with class assumed-rank value in visit_Associate

HarshitaKalani:
towards: https://github.com/lfortran/lfortran/issues/11292


certik (approved):
Needs rebase

PR #11304, opened 2026-05-02
fix: remove files opened with `STATUS=scratch` on normal program termination

### LINKED ISSUE #11299: Scratch files not deleted during normal program termination
Per 12.5.6.19 of F2023: a file opened with `status='scratch'` is "deleted at the execution of a CLOSE statement referring to the same unit or at the normal termination of the program."  Currently the file is being deleted by an explicit CLOSE.  However if the file is not explicitly closed, the file is not being deleted during normal termination:
```
program open_scratch
  implicit none

  character(128) :: scr_fn
  logical :: exists

  integer, parameter :: lun_scr = 42

  open (lun_scr, status='scratch', form='formatted')
  inquire (lun_scr, name=scr_fn)
  print *, 'scratch filename = ', trim (scr_fn)
  write (lun_scr, *) 'hello world!'

! No CLOSE statement, so scratch file(s) should be deleted at
! normal program termination.

end program
$
```
```
$ ls *txt
ls: cannot access '*txt': No such file or directory
$ lfortran open_scratch.f90
scratch filename = _lfortran_generated_file_TJKM2K7BMGEIDU17AJ2G5CTMW.txt
$ ls -l *.txt
-rw-rw-r-- 1 wws wws 13 May  1 11:16 _lfortran_generated_file_TJKM2K7BMGEIDU17AJ2G5CTMW.txt
$
```
```
$ gfortran open_scratch.f90
$ ./a.out
 scratch filename = K8��E��E�K8��E�
$ ls -l K*
ls: cannot access 'K*': No such file or directory
$
```

kmr-srbh:
Fixes #11299

### DIFF
--- a/integration_tests/file_69.f90
+++ b/integration_tests/file_69.f90
@@ -1,0 +1,12 @@
+program file_69
+   implicit none
+
+   character(128) :: scr_fn
+   integer, parameter :: lun_scr = 42
+
+   open (lun_scr, status='scratch', form='formatted')
+   inquire (lun_scr, name=scr_fn)
+   print *, 'scratch filename = ', trim (scr_fn)
+   write (lun_scr, *) 'hello world!'
+
+end program
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5748,8 +5748,15 @@ void remove_from_unit_to_file(int32_t unit_num) {
 }
 
 static void _lfortran_close_all_units(void) {
+    const char *scratch_prefix = "_lfortran_generated_file";
+    const size_t scratch_prefix_len = 24; // strlen("_lfortran_generated_file")
     for (int i = 0; i <= last_index_used; i++) {
         if (unit_to_file[i].filename != NULL) {
+            // Delete scratch files at normal program termination
+            if (strncmp(unit_to_file[i].filename, scratch_prefix,
+                        scratch_prefix_len) == 0) {
+                remove(unit_to_file[i].filename);
+            }
             internal_free(unit_to_file[i].filename);
             unit_to_file[i].filename = NULL;
         }
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2854,6 +2854,7 @@ RUN(NAME file_65 LABELS gfortran llvm)
 RUN(NAME file_66 LABELS gfortran llvm)
 RUN(NAME file_67 LABELS gfortran llvm)
 RUN(NAME file_68 LABELS gfortran llvm)
+RUN(NAME file_69 LABELS gfortran llvm)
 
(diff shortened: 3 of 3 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5748,8 +5748,15 @@ void remove_from_unit_to_file(int32_t unit_num) {
 }
 
 static void _lfortran_close_all_units(void) {
+    const char *scratch_prefix = "_lfortran_generated_file";
```
certik, line 5751: This hardwired name doesn't seem right --- is this also used somewhere else? I see it used three times already:
```
$ git grep -n _lfortran_generated_file
src/libasr/runtime/lfortran_intrinsics.c:5925:        char *prefix = "_lfortran_generated_file", *format = "txt";
src/libasr/runtime/lfortran_intrinsics.c:12109:    const char *scratch_file = "_lfortran_generated_file";
src/libasr/runtime/lfortran_intrinsics.c:12110:    const int64_t scratch_file_len = sizeof("_lfortran_generated_file") - 1; // exclude '\0'
```
We should have this name there just once. Can be fixed later.

certik (approved):

integration_tests/file_69.f90
```diff
@@ -0,0 +1,12 @@
...
+
+   character(128) :: scr_fn
+   integer, parameter :: lun_scr = 42
+
+   open (lun_scr, status='scratch', form='formatted')
+   inquire (lun_scr, name=scr_fn)
+   print *, 'scratch filename = ', trim (scr_fn)
+   write (lun_scr, *) 'hello world!'
+
+end program
```
certik, line 12: This program passes in main:
```
$ lfortran a.f90
scratch filename = _lfortran_generated_file_JI04XMR5ZP66LL6VC2UB00HJ2.txt
$ echo $?
0
```
Is it thus testing anything? Is it needed at all? Don't we already test scratch files?

certik:
I am going to merge this, as it fixes a bug. But I am not sure about the test. We should also fix up the fix per my comment above.

integration_tests/file_69.f90
kmr-srbh: Yes the scratch files functionality already worked before in main. This test just opens a scratch file and leaves it open for the compiler to close. Other tests with `status=scratch` all manually close the file. 

I think another way to test this can be, by running another test (after all file tests) that runs the `ls` command to search for `.txt` files with the `_lfortran_generated_file_` prefix and counts the outputs? Something like:
```console
(base) saurabhkumar@Mac lfortran % ls "_lfortran_generated_file_"*.txt        
_lfortran_generated_file_abcd.txt       _lfortran_generated_file_efgh.txt
(base) saurabhkumar@Mac lfortran % ls "_lfortran_generated_file_"*.txt | wc -l
       2
```

src/libasr/runtime/lfortran_intrinsics.c
kmr-srbh: I'll fix this. 👍

PR #11306, opened 2026-05-03
fix: character array allocation length for function results

### LINKED ISSUE #11273: Semantic error: Only String type is supported as of now
Godbolt LFortran 0.59.0 says `Build failed` for code below and lfortran says

```
semantic error: Only String type is supported as of now.
  --> xdefined_operator_char_array_constructor.f90:10:1
   |
10 | character(len=len(values(1))), allocatable :: kept(:)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```
for
```
lfortran --version
LFortran version: 0.45.0
Platform: Windows
LLVM: 19.1.1
Default target: x86_64-pc-windows-msvc
```

```fortran
module xdefined_operator_char_array_constructor_mod
implicit none
interface operator (.exclude.)
   module procedure char_exclude
end interface
contains

function char_exclude(values, drop) result(kept)
character(len=*), intent(in) :: values(:), drop(:)
character(len=len(values(1))), allocatable :: kept(:)
integer :: i, n
n = 0
do i = 1, size(values)
   if (.not. any(values(i) == drop)) n = n + 1
end do
allocate(kept(n))
n = 0
do i = 1, size(values)
   if (.not. any(values(i) == drop)) then
      n = n + 1
      kept(n) = values(i)
   end if
end do
end function char_exclude

end module xdefined_operator_char_array_constructor_mod

program xdefined_operator_char_array_constructor
use xdefined_operator_char_array_constructor_mod, only: operator(.exclude.)
implicit none
write (*,"(100(a,1x))") ["dog","cat","boy"] .exclude. ["cat "]
end program xdefined_operator_char_array_constructor
```

opixdown:
fixed #11273

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3713,6 +3713,7 @@ RUN(NAME character_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS -
 RUN(NAME character_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME character_26 LABELS gfortran llvm)
 RUN(NAME character_27 LABELS gfortran llvm)
+RUN(NAME character_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME char_array_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME char_array_indexing LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME char_array_indexing_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/character_28.f90
+++ b/integration_tests/character_28.f90
@@ -1,0 +1,43 @@
+module char_28_mod
+implicit none
+interface operator (.exclude.)
+   module procedure char_exclude
+end interface
+contains
+
+function char_exclude(values, drop) result(kept)
+character(len=*), intent(in) :: values(:), drop(:)
+character(len=len(values(1))), allocatable :: kept(:)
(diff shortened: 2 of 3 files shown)


HarshitaKalani (approved):

assem2002:
Please resolve CI failure

integration_tests/CMakeLists.txt
```diff
@@ -3713,6 +3713,7 @@ RUN(NAME character_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS -
 RUN(NAME character_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME character_26 LABELS gfortran llvm)
 RUN(NAME character_27 LABELS gfortran llvm)
+RUN(NAME character_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
```
HarshitaKalani: fails with the fortran flag

integration_tests/character_28.f90
```diff
@@ -0,0 +1,43 @@
...
+
+end module char_28_mod
+
+program char_28
+use char_28_mod, only: operator(.exclude.)
+implicit none
+
+character(len=3), allocatable :: res(:)
+
+res = ["dog","cat","boy"] .exclude. ["cat "]
```
certik, line 34: This requires `--realloc-lhs-arrays`, and we need to add a test that fails, then we need to add `--realloc-lhs-arrays` into CMake for this.

certik: @opixdown can you fix this in the next PR?

opixdown: @certik i'll fix it in upcoming pr

PR #11312, opened 2026-05-03
Fix : character-array as arraysection

### LINKED ISSUE #11262: Only string(a:b) supported
Get `Only string(a:b) supported for now` with Godbolt LFortran 0.59.0 for

```fortran
module xunique_char_index_section_mod
implicit none
contains

subroutine unique_char(ivec, ipos, nuniq)
character(len=*), intent(in) :: ivec(:)
integer, intent(out) :: ipos(:)
integer, intent(out) :: nuniq
integer :: i, n

n = size(ivec)
if (n == 0) then
   nuniq = 0
   return
end if
ipos(1) = 1
nuniq = 1
do i = 2, n
   if (all(ivec(i) /= ivec(ipos(:nuniq)))) then
      nuniq = nuniq + 1
      ipos(nuniq) = i
   end if
end do
end subroutine unique_char

end module xunique_char_index_section_mod

program xunique_char_index_section
use xunique_char_index_section_mod, only: unique_char
implicit none
character(len=3) :: x(5)
integer :: ipos(5), nuniq

x = [character(len=3) :: "aa", "bb", "aa", "cc", "bb"]
call unique_char(x, ipos, nuniq)
print *, nuniq
print *, ipos(:nuniq)
end program xunique_char_index_section
```

assem2002:
fix #11262

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3319,6 +3319,7 @@ RUN(NAME array_section_27 LABELS gfortran llvm llvm_single_invocation
     EXTRAFILES array_section_27_mod.f90 array_section_27_sub.f90
     EXTRA_ARGS --separate-compilation)
 RUN(NAME array_section_28 LABELS gfortran llvm)
+RUN(NAME array_section_29 LABELS gfortran llvm)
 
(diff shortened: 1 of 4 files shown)


certik:
The test is not registered, so none of the code in this PR is tested.

PR #11328, opened 2026-05-04
fix: suppress false shadowing warning for reimported identical symbols

### LINKED ISSUE #11286: Warning about symbols that shadow the same symbol in the current scope
I am not sure this should prevoke a warning, and ifx and gfortran agree: (filed in a.f90)
```fortran
module a
use, intrinsic :: ISO_FORTRAN_ENV, only: int64
implicit none
end module a

module b
use a
use, intrinsic :: ISO_FORTRAN_ENV, only: int64
implicit none
end module b

program c
use b
implicit none
end program c
```

Output
test$ ifx a.f90
test$ gfortran a.f90
test$ lfortran a.f90
warning: Symbol 'int64' from module 'lfortran_intrinsic_iso_fortran_env' shadows 'int64' in the current scope
 --> a.f90:8:42
  |
8 | use, intrinsic :: ISO_FORTRAN_ENV, only: int64
  |                                          ^^^^^

perhaps a pedantic flag could bring this to the users attention but not be default behaviour?  I am no expert but the program is standard conforming?

Keep up the great work.

opixdown:
fixes #11286

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -19765,7 +19765,19 @@ public:
         }
         if (ASR::is_a<ASR::Function_t>(*t) &&
             ASR::down_cast<ASR::Function_t>(t)->m_return_var == nullptr) {
-            if (current_scope->get_symbol(local_sym) != nullptr) {
+            bool is_already_defined = false;
+            ASR::symbol_t *imported_sub_sym = current_scope->get_symbol(local_sym);
+            if (imported_sub_sym != nullptr) {
+                if (ASR::is_a<ASR::ExternalSymbol_t>(*imported_sub_sym)) {
+                    ASR::ExternalSymbol_t *ext_sym = ASR::down_cast<ASR::ExternalSymbol_t>(imported_sub_sym);
+                    if (ext_sym->m_external != t) {
+                        is_already_defined = true;
+                    }
+                } else {
+                    is_already_defined = true;
+                }
+            }
+            if (is_already_defined) {
                 diag.add(Diagnostic(
                     "Symbol '" + local_sym + "' from module '" + m->m_name + "' shadows '" + local_sym + "' in the current scope",
                     Level::Warning, Stage::Semantic, {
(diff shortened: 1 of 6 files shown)


HarshitaKalani (approved):

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19790,7 +19802,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 m->m_name, nullptr, 0, msub->m_name,
                 dflt_access
                 );
-            current_scope->add_symbol(local_sym, ASR::down_cast<ASR::symbol_t>(sub));
+            current_scope->add_or_overwrite_symbol(local_sym, ASR::down_cast<ASR::symbol_t>(sub));
```
certik: The overwrite approach is asking for trouble. At the very least add an assert that we are overwriting with the same symbol. But I would just skip it if the symbol is already there, because perhaps it's already referenced somewhere so it can become dangling when we overwrite?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19765,7 +19765,19 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                if (ASR::is_a<ASR::ExternalSymbol_t>(*imported_sub_sym)) {
+                    ASR::ExternalSymbol_t *ext_sym = ASR::down_cast<ASR::ExternalSymbol_t>(imported_sub_sym);
+                    if (ext_sym->m_external != t) {
+                        is_already_defined = true;
+                    }
+                } else {
+                    is_already_defined = true;
+                }
+            }
+            if (is_already_defined) {
```
certik: Let's add a reference test for "run = True" which will show empty, thus not emitting this warning anymore.

src/lfortran/semantics/ast_common_visitor.h
opixdown: Thanks for the review You are right, overwriting is risky. I will update it

PR #11330, opened 2026-05-04
fix: add ExternalSymbol fallback resolution

opixdown:
while compiling `benchmark.f`  with workaround the `Run.f90` was causing ice. to solve that I
Added a fallback that searches loaded modules for struct/enum/function owners resolves symbols when m_module_name points to a nested owner (e.g., arg_base) prevents ICE on unresolved ExternalSymbol owners

### DIFF
--- a/src/libasr/serialization.cpp
+++ b/src/libasr/serialization.cpp
@@ -382,6 +382,26 @@ public:
         }
         return result;
     }
+    ASR::symbol_t* owner_in_symtab(SymbolTable *symtab,const std::string &owner_name) {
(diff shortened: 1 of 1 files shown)


Jatinagarwal24:
Hi @opixdown Thanks for raising the pr. 
I think you miss adding the test so for that i have raised the pr #11333.

certik:
Add a test. Also a fix in serialization requires careful understanding what is going on and we have to be convinced this is the right fix.

Jatinagarwal24:
Hi @certik, @opixdown  this pr is fixed in #11333 so this can be closed now
Thanks!

PR #11349, opened 2026-05-05
Add support  of real128 

### LINKED ISSUE #11329: handle real128 kind correctly in SELECT TYPE type is(real(kind))
while exploring  `benchmark.f ` found these

### mre 
```
program repro
  use, intrinsic :: iso_fortran_env, only: real128
  implicit none
  integer, parameter :: r16 = real128
  class(*), allocatable :: x
  allocate(x, source=1.0)
  select type (x)
  type is (real(r16))
    print *, x
  class default
    print *, "test passed"
  end select
end program
```
### Output
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % lfortran a.f90 
semantic error: Kind -1 is not supported for Real
 --> a.f90:8:17
  |
8 |   type is (real(r16))
  |                 ^^^ 

```

### Gfortran
```
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran a.f90&& ./a.out
 test passed
```

opixdown:
fixes #11329

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -8849,7 +8849,7 @@ public:
             }
         }
         if (sym_type->m_type == AST::decl_typeType::TypeReal) {
-            if (!is_derived_type && a_kind != 4 && a_kind != 8) {
+            if (!is_derived_type && a_kind != 4 && a_kind != 8 && a_kind != 16) {
                 diag.add(Diagnostic(
                     "Kind " + std::to_string(a_kind) + " is not supported for Real",
                     Level::Error, Stage::Semantic, {
@@ -8917,7 +8917,7 @@ public:
                     ASRUtils::type_get_past_allocatable(type)));
             }
         } else if (sym_type->m_type == AST::decl_typeType::TypeComplex) {
-            if (!is_derived_type && a_kind != 4 && a_kind != 8) {
+            if (!is_derived_type && a_kind != 4 && a_kind != 8 && a_kind != 16) {
                 diag.add(Diagnostic(
                     "Kind " + std::to_string(a_kind) + " is not supported for Complex",
                     Level::Error, Stage::Semantic, {
@@ -18383,6 +18383,8 @@ public:
             r = ASRUtils::extract_real_4(x.m_n);
         } else if ( r_kind == 8 ) {
             r = ASRUtils::extract_real_8(x.m_n);
+        } else if ( r_kind == 16 ) {
+            r = ASRUtils::extract_real_16(x.m_n);
         } else {
             diag.add(Diagnostic("Kind not supported",
                 Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -105,6 +105,12 @@ static inline double extract_real_8(const char *s) {
     return std::strtod(r_str.c_str(), nullptr);
 }
 
+static inline double extract_real_16(const char *s) {
+    std::string r_str = ASRUtils::extract_real(s);
+    long double r = std::strtold(r_str.c_str(), nullptr);
+    return (double)r;
+}
+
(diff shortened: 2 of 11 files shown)


src/libasr/asr_utils.h
```diff
@@ -105,6 +105,12 @@ static inline double extract_real_8(const char *s) {
     return std::strtod(r_str.c_str(), nullptr);
 }
 
+static inline double extract_real_16(const char *s) {
+    std::string r_str = ASRUtils::extract_real(s);
+    long double r = std::strtold(r_str.c_str(), nullptr);
```
assem2002, line 110: Please detect the overflow. We should raise error if we have a compile-time 64bit-float overflow as C++ can't handle it.

assem2002:
Can we do simple arithmetic operations with this PR?

assem2002 (approved):

certik:
We have to be careful here, the minute we not return -1 on quadruple precision (real128) we should support it. I think the above doesn't actually support real128, nor has a path how to do that.

I would recommend to first create a prototype that actually does some arithmetic on all platforms, to ensure we have the design right and that we are confident we can extend to full support (as people report bugs).

With the above PR people will start reporting bugs, but we have no path towards fixing them.

assem2002:
If LLVM takes care of the heavy work (that's the backend).
LFortran frontend actually can't handle it with the current flow.
I think we can support it but with some restrictions on the frontend side. and maybe limitations with some operations :)

certik:
My understanding was that it only works in Ubuntu not macOS due to some LLVM limitations.

opixdown:
> Member

i tried running test but it fails ,with this pr its just like we are just accepting the syntax

opixdown:
> We have to be careful here, the minute we not return -1 on quadruple precision (real128) we should support it. I think the above doesn't actually support real128, nor has a path how to do that.
> 
> I would recommend to first create a prototype that actually does some arithmetic on all platforms, to ensure we have the design right and that we are confident we can extend to full support (as people report bugs).
> 
> With the above PR people will start reporting bugs, but we have no path towards fixing them.

you are right, I’ll focus on making working prototype first, and once that is good we can expand support further.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -8849,7 +8849,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             }
         }
         if (sym_type->m_type == AST::decl_typeType::TypeReal) {
-            if (!is_derived_type && a_kind != 4 && a_kind != 8) {
+            if (!is_derived_type && a_kind != 4 && a_kind != 8 && a_kind != 16) {
```
certik, line 8852: You can enable this 16 kind only when a compiler option is passed, otherwise do not allow it.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
```diff
@@ -11,7 +11,7 @@ module lfortran_intrinsic_iso_fortran_env
 integer, parameter :: real16 = -1     ! F2023: half-precision (not supported)
 integer, parameter :: real32 = 4
 integer, parameter :: real64 = 8
-integer, parameter :: real128 = -1
+integer, parameter :: real128 = 16
```
certik, line 14: We have to use -1 for kinds that we do not support, so for now leave it as -1 until f128 works reasonably well.

src/runtime/pure/lfortran_intrinsic_iso_fortran_env.f90
```diff
@@ -26,7 +26,7 @@ module lfortran_intrinsic_iso_fortran_env
 
 ! Kind arrays (F2008/F2018)
 integer, parameter :: integer_kinds(4) = [1, 2, 4, 8]
-integer, parameter :: real_kinds(2) = [4, 8]
+integer, parameter :: real_kinds(3) = [4, 8, 16]
```
certik, line 29: Same here.

PR #11355, opened 2026-05-05
fix: reject negative key codes in interactive prompt input filter

### LINKED ISSUE #938: Failure on non-ascii input
When I have my keyboard on a German layout, it sometimes happens that I accidentally hit an umlaut (ä, ü, ö), usually this is not a problem as I just backspace. In LFortran's interactive session however this results in a runtime error:

```
❯ lfortran --version
LFortran version: 0.14.0
Platform: Linux
Default target: x86_64-unknown-linux-gnu
❯ lfortran
  * Use Ctrl-D to exit
  * Use Enter to submit
    - History (Keys: Up, Down)
>>> runtime_error: Invalid byte in UTF8 encoded string                   1,1   ]
```

s1291:
Fixes: https://github.com/lfortran/lfortran/issues/938

The REPL was crashing when you type non-ascii characters like `ù`, `ä`. This PR fixes the crash by tightening the filter to only accept printable ascii (32..127), so unknown keys are silently ignored instead of corrupting the buffer.


assem2002 (approved):
LGTM! Thanks

src/bin/tpl/cpp-terminal/prompt0.h
```diff
@@ -124,7 +124,7 @@ std::string prompt0(const Terminal &term, const std::string &prompt_string,
         key = term.read_key();
         if (  (key >= 'a' && key <= 'z') ||
               (key >= 'A' && key <= 'Z') ||
-              (!iscntrl(key) && key < 128)  ) {
+              (key >= 32 && key < 128 && !iscntrl(key)) ) {
```
assem2002: This should be enough. Conditions above aren't needed

s1291: Thank you. I simplified the condition

certik:
@s1291 did you close this by a mistake? I can't see any changes.

assem2002:
Minor conflict. re-submitted here #11359

s1291:
> @s1291 did you close this by a mistake? I can't see any changes.

Sorry, I made a mistake with git push --force-with-lease which messed with PR. re-opened here: https://github.com/lfortran/lfortran/pull/11359

PR #11359, opened 2026-05-05
fix: ignore non-ASCII keystrokes in REPL instead of crashing

### LINKED ISSUE #938: Failure on non-ascii input

s1291:
Fixes: https://github.com/lfortran/lfortran/issues/938

The REPL was crashing when you type non-ascii characters like `ù`, `ä`. This PR fixes the crash by tightening the filter to only accept printable ascii (32..127), so unknown keys are silently ignored instead of corrupting the buffer.

### DIFF
--- a/src/bin/tpl/cpp-terminal/prompt0.h
+++ b/src/bin/tpl/cpp-terminal/prompt0.h
@@ -122,9 +122,7 @@ std::string prompt0(const Terminal &term, const std::string &prompt_string,
     bool not_complete = true;
     while (not_complete) {
         key = term.read_key();
-        if (  (key >= 'a' && key <= 'z') ||
-              (key >= 'A' && key <= 'Z') ||
-              (key >= 32 && key < 128 && !iscntrl(key)) ) {
+        if  (key >= 32 && key <= 126) {
             std::string before = m.lines[m.cursor_row-1].substr(0,
                     m.cursor_col-1);
             std::string newchar; newchar.push_back(key);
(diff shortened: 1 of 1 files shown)


s1291:
@assem2002  Sorry about the noise, I ran into an issue with my git push and had to recreate the PR. This new PR contains the same changes. Thanks again for reviewing the previous one.

assem2002:
No worries. Thanks for re-submitting

certik:
Let's figure out how to add a test for this, this will go into `src/lfortran/tests/test_llvm.cpp` as a `FortranEvaluator` test.

Actually, this is a terminal handling issue. We have tests for that in the upstream third-party project, but we kind of use our own fork now to fit our needs better. So it's probably fine for now, we just need to verify that the logic is correct.

s1291:
> Let's figure out how to add a test for this, this will go into `src/lfortran/tests/test_llvm.cpp` as a `FortranEvaluator` test.
> 
> Actually, this is a terminal handling issue. We have tests for that in the upstream third-party project, but we kind of use our own fork now to fit our needs better. So it's probably fine for now, we just need to verify that the logic is correct.

The current check simply verifies that the input key a printable ASCII character, in the range 32 to 126. I don't see an obvious way to automate this in the test setup.

certik (approved):
Let's try it. If this creates issues, we'll have to iterate on it.

PR #11365, opened 2026-05-05
fix: scope-qualify finalizer cache key for local derived types

HarshitaKalani:


src/libasr/codegen/llvm_utils.h
```diff
@@ -2024,9 +2024,9 @@ class ASRToLLVMVisitor;
                 key += std::to_string(n_dims) + "_";
                 key += get_type_key(ASRUtils::extract_type(t_past), struct_sym);
             } else if(struct_sym != nullptr) { // StructType or structType Class
-                key += ASRUtils::get_type_code(t_past, false, false, false) +"__" + struct_sym->m_name;
+                key += ASRUtils::get_type_code(t_past, false, false, false) +"__" + LCompilers::get_type_key(struct_sym);
```
assem2002: I think the best is to insert the symboltable code of struct_sym -- I can think of cases where the containing construct isn't enough to distinguish struct symbols

HarshitaKalani: Good point. Done. Thanks Assem!

assem2002 (approved):

PR #11375, opened 2026-05-06
Updated Fortran template examples syntax

odeladelplays-cmd:
I have updated the syntax in some of the template examples to match the current iteration of the proposed 2028 standard for template syntax in Fortran. The purpose of this is so you may update LFortran to support the modern syntax. I may update more of the examples in the coming days/weeks.


assem2002:
I think we have syntax error here, We might need to modify the parser.
also update reference tests using `python3 run_tests.py -u` in the root directory of the repo

odeladelplays-cmd:
The syntax errors in the changed tests are the point of the pull request. The Fortran Template syntax in LFortran right now is old and outdated, so I have updated some examples to the up to date syntax so LFortran may be improved to support Templates in their "official" capacity with the next Fortran Standard.

When running the python script mentioned, I got an error message, and 2000+ new files were generated. Is your wish for me to commit all of them for review?

odeladelplays-cmd:
I have now updated most of, if not all of the template examples I could find in integration tests to match the 2028 Standard. Hopefully this should be sufficient to help along any efforts to update the compiler to support the upcoming standard. :)

odeladelplays-cmd:
Found something I had missed in the standard and updated everything to include it. Hopefully everything should be up to the standard now.

PR #11376, opened 2026-05-06
fix: handle AssumedRankArray in descriptor dimension access

HarshitaKalani:


assem2002 (approved):
I can reproduce the error locally. 
LGTM. Thanks

Issue #11353, opened 2026-05-05
ICE: Cannot extract the physical type of 0 type


Beliavsky:
For the code at https://gist.github.com/Beliavsky/2fa41e4f3eb78327e4879f9216105d61
godbolt lfortran 0.59.0 gives the results below. It's a large code that exercises many Fortran features.

```
warning: `--generate-object-code` is deprecated and will be removed in a future release; use `--separate-compilation` instead.
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x70a2c4
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29e3f
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x29d8f
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x6886bb
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x7a285f
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x793634
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x9379f9
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x168a1d0
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x16835c4
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x1681a96
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x13a813c
  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x13c35bb
[...]

PR #11377, opened 2026-05-06
Fix : Use `BaseWalkVisitor` where needed in `function_call_in_declaration` pass

assem2002:
Fix : Don't explicitly check sub-expression looking for things, instead use a class extending `BaseWalkVisitor` and add the required visitor logic + Expose only a static function to make the class aimed for a single purpose

Introduced 2 class :
- `get_arg_indices_used`
- `expr_contains_functionCall_with_Nonscalar_return`

Gets https://github.com/lfortran/lfortran/issues/11353 to compile and run (with the code invoking error due to missing read file) -- we should check if issue is resolved after merging


certik (approved):

certik:
Looks great, I like removing total number of lines.

PR #11386, opened 2026-05-07
tests: add --realloc-lhs-arrays flag to character_28

### LINKED ISSUE #11325: This requires `--realloc-lhs-arrays`, and we need to add a test that fails, then we need to add `--realloc-lhs-arrays` into CMake for this.
This requires `--realloc-lhs-arrays`, and we need to add a test that fails, then we need to add `--realloc-lhs-arrays` into CMake for this.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/11306#discussion_r3179208789_

srinjoy933:
fixes #11325


integration_tests/CMakeLists.txt
```diff
@@ -3758,8 +3757,7 @@ RUN(NAME character_24 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS -
 RUN(NAME character_25 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME character_26 LABELS gfortran llvm)
 RUN(NAME character_27 LABELS gfortran llvm)
-RUN(NAME character_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
-RUN(NAME character_29 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+RUN(NAME character_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
```
kmr-srbh, line 3780:
```suggestion
RUN(NAME character_28 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays)
RUN(NAME character_29 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
```

kmr-srbh (approved):
The change is fine -- we are adding the required flag to the test. But I would not link the issue with the PR, since that calls for a fix to a bug where this test should have failed without this flag.

certik:
This PR doesn't fix the issue. To fix the issue, we must add two tests: one that fails without the flag and one that passes with it. This PR just adds a test to pass with it, which is not enough.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -753,12 +754,14 @@ class ReplaceFunctionCallWithSubroutineCallVisitor:
...
-                        // semantics for an allocatable LHS receiving a function result).
+                        
                         assignment->m_move_allocation = true;
-                        assignment->m_realloc_lhs = true;
+                     
+                        if (ASRUtils::is_character(*ASRUtils::expr_type(target))) {
+                            assignment->m_realloc_lhs = pass_options.realloc_lhs_arrays;
+                        } else {
+                            assignment->m_realloc_lhs = true;
+                        }
```
kmr-srbh, line 763: I think we should this for all cases, not just for character arrays.

kmr-srbh (approved):
The PR fixes the case where the `realloc_lhs` field was always set to `true` by the pass, to now correctly use the value of flag `--realloc-lhs-arrays`, which looks correct. Left just a minor comment.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -294,9 +293,9 @@ private :
 public:
 
     ReplaceFunctionCallWithSubroutineCall(
-        Allocator                                           &al_,
-        SymbolTable*                                        &current_scope,
-        Vec<ASR::stmt_t*>                                   &pass_result_,
+        Allocator                                                   &al_,
+        SymbolTable* &current_scope,
+        Vec<ASR::stmt_t*>                                           &pass_result_,
```
kmr-srbh: This looks like just code formatting. Please revert this change to minimise the PR diff.

srinjoy933:
Hi @kmr-srbh ,
I have updated the PR to enforce the --realloc-lhs-arrays flag requirement universally across all array types, exactly as you mentioned. As a  result of making it universal rather than restricting it just to character arrays, the allocate_69 and do_loop_07 tests naturally began failing because they rely on implicit LHS reallocation for their integer and real arrays. To preserve their intended testing behavior under the new compiler logic, I added the EXTRA_ARGS --realloc-lhs-arrays flag to both of their configurations in CMakeLists.txt.

kmr-srbh:
The other tests do not need the flag. I see they have the proper `allocate()` statements.

srinjoy933:
but without those flags ,the tests are failing .previously when i made that universal rather than only to character array , I got this test failure message in CI.
 Total Test time (real) = 5.36 sec
The following tests FAILED:
1874 - allocate_69 (Failed)

2789 - do_loop_07 (Failed)

Errors while running CTest

Output from these tests are in: /home/runner/work/lfortran/lfortran/integration_tests/build-lfortran-llvm/Testing/Temporary/LastTest.log

Error: Process completed with exit code 8. 

after I added the flag to those particular test it ran smoothly .One more point,  when I am applying to only the character array those tests are not failing,

kmr-srbh:
Ah, I see. You modified the pass that converts the function calls, so it might be affecting the allocation in some way. Can you check what is the output after the pass without the `--realloc-lhs-arrays` flag? Use `--dump-all-passes-fortran` flag for debugging, which will print the intermediate representation in Fortran for the various passes. Check if the `allocate` statement is there. The issue could be something also, but first we need to check this.

srinjoy933:
@kmr-srbh, since this original issue was specifically about character arrays, could we revert the logic back to only the character array case and remove the --realloc-lhs-arrays flag from the other two tests? This would allow us to resolve this particular issue now. In my previous work, I ran into situations where those exact two tests failed when applying changes to all arrays without the flag, and I had to specifically limit the logic to character arrays to get the test suite to pass. Because of that, I would prefer to create a new tracking issue and handle the broader 'all arrays' case in a separate PR.

Does this sound good to you?

srinjoy933:
to solve the issue for all type of array  I have created a separate issue https://github.com/lfortran/lfortran/issues/11609

srinjoy933:
This PR solves issue #11325.
Before this, the test (integration_tests/character_28.f90) worked on the main branch without the --realloc-lhs-arrays flag. Now, the test fails without it. However, when you add the --realloc-lhs-arrays flag, it works completely fine

certik:
If after this PR the test fails without `--realloc-lhs-arrays` as it should, then add a reference test for the runtime failure.

PR #11388, opened 2026-05-07
Prototype for real_128

opixdown:


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -2104,6 +2114,47 @@ void handle_hexadecimal(const char* format, Primitive_Types type,
...
+                char *padded = (char*)internal_malloc((width + 1) * sizeof(char));
+                memset(padded, ' ', width - _olen);
+                memcpy(padded + (width - _olen), _out, _olen + 1);
+                internal_free(_out);
+                *result = padded;
+            } else {
+                *result = _out;
+            }
+            return; /* early return: skip the generic uint64 path below */
+        }
```
certik, line 2157: Is the lfortran_float128.h/c only called from this block?

src/libasr/ASR.asdl
```diff
@@ -115,7 +115,7 @@ expr
     | UnsignedIntegerBitNot(expr arg, ttype type, expr? value)
     | UnsignedIntegerCompare(expr left, cmpop op, expr right, ttype type, expr? value)
     | UnsignedIntegerBinOp(expr left, binop op, expr right, ttype type, expr? value)
-    | RealConstant(float r, ttype type)
+    | RealConstant(float r, string? n, ttype type)
```
certik, line 118: Here is how to fix this: https://github.com/lfortran/lfortran/issues/11418, so remove the `string? n` and just use `r` as a NaN-boxed double.

certik: Even better: I just realized that we already store the kind in `type`. So for kind=8 we interpret `r` as double and use as-is. And for kind 16 we interpret it as a pointer, which will point to 128 bits in memory (allocated on an arena). I think we might not even need to box it. The boxing is effectively embedding the type into the double as well, but we already store it separate, so it's just about how to interpret the `r`.

PR #11389, opened 2026-05-07
Refactor shared DO loop terminator workaround

### LINKED ISSUE #11245: Fix the workaround done in  #11241.
This one is a workaround, I think there is a better way to fix this, but we can do it in the next PR.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/11241#discussion_r3148233566_

srinjoy933:
Fixes #11245.
 This is a follow-up to PR #11241 to properly clean up the shared DO loop terminator workaround.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -13912,6 +13912,20 @@ public:
         llvm::BasicBlock *bb = llvm::BasicBlock::Create(context, "unreachable_after_cycle");
         start_new_block(bb);
     }
+    void safe_transition_to_block(llvm::BasicBlock *target) {
+        if (target->getTerminator()) {
+            llvm::BasicBlock *last_bb = builder->GetInsertBlock();
+            llvm::Function *fn = last_bb->getParent();
+            llvm::BasicBlock *cont = llvm::BasicBlock::Create(context, "goto_target.cont");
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -13938,22 +13952,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                "goto_target.cont");
-            llvm_fn_insert_bb(fn, cont);
-            if (last_bb->getTerminator() == nullptr) {
-                builder->CreateBr(cont);
-            }
-            builder->SetInsertPoint(cont);
-        } else {
-            start_new_block(target);
-        }
+        safe_transition_to_block(target);
```
kmr-srbh, line 13955: We are only extracting this block to a function and calling it. I think we don't want to do that. Instead, we should figure out why this new block creation logic was introduced. That would cause us to either fix a different bug that we might find, or improve the design here.

I think the current change is not what we want to do.

srinjoy933: @kmr-srbh That makes complete sense. I misinterpreted the original scope of the issue and agree that we need to fix the root cause rather than just refactoring the workaround. I'll close this PR for now and open a fresh one once I have the proper architectural fix . Thanks for the guidance!

PR #11392, opened 2026-05-07
Fix warning message for sizeof()

### LINKED ISSUE #10868: Change the warning for sizeof() to recommend storage_size
Currently the warning is:
```
warning: `sizeof` is an LFortran extension
   --> integration_tests/intrinsics_450.f90:148:9
    |
148 |     if (sizeof(ac4) /= (storage_size(ac4(1)) / 8) * size(ac4)) &
    |         ^^^^^^^^^^^ use `c_sizeof` from `iso_c_binding` for portable code
```
But this is incorrect since `sizeof()` can be used on any Fortran expression, while `c_sizeof()` can only be used on C interoperable types. The `storage_size()` however can be used. The conversion formulas are:

For scalar we have `SIZEOF(X) ≡ STORAGE_SIZE(X) / 8` and for an array we have `SIZEOF(X) ≡ [STORAGE_SIZE(X(1)) / 8] × SIZE(X)`.

So I think it should just say "use the `storage_size` intrinsic function". Alternatively it can also mention `c_sizeof()`, but it only works on `bind(c)` types, so not all types.

srinjoy933:
fixes #10868


assem2002 (approved):
LGTM

PR #11394, opened 2026-05-08
fix: NEWUNIT to assign negative unit numbers and resolve runtime IO bounds

### LINKED ISSUE #11228: OPEN: newunit= returns illegal unit value
The unit returned by newunit= must be negative, and not -1.  Gfortran starts at -10, and flang at -13:
```
program open_newunit
  use iso_fortran_env
  implicit none

  integer :: io_lun

  open (newunit=io_lun, file='open_newunit.dat', status='replace', form='formatted')
  print *, 'file opened on unit', io_lun

! Test for the first few conditions specified by F2023 12.5.6.13:
  if (io_lun >= -1) error stop 'newunit >= -1'
  if (io_lun == ERROR_UNIT) error stop 'newunit == ERROR_UNIT'
  if (io_lun == INPUT_UNIT) error stop 'newunit == INPUT_UNIT'
  if (io_lun == OUTPUT_UNIT) error stop 'newunit == OUTPUT_UNIT'

  close (io_lun, status='delete')

end program
```
```
$ lfortran open_newunit.f90
file opened on unit    1
ERROR STOP newunit >= -1
$
```
```
$ gfortran open_newunit.f90
$ ./a.out
 file opened on unit         -10
$ flang open_newunit.f90
$ ./a.out
 file opened on unit          -13
$
```

Jatinagarwal24:
fixes #11228

Bug:
The `OPEN(NEWUNIT=...)` statement was incorrectly assigning positive unit
numbers (0 to 1000). The Fortran standard explicitly requires `NEWUNIT`
to assign a unique negative integer to prevent collisions with user-specified
or pre-connected standard units. Furthermore, the IO runtime inside 
`lfortran_intrinsics.c` incorrectly assumed unit numbers strictly fell
between `0` and `999`. It used the unit number directly as an array index 
for internal IO tracking (e.g., `seq_unf_pending`) and hardcoded 0-999 loops,
which caused `INQUIRE` statements to fail and crashed the compiler when
forced to handle negative units.

Fix:
- Modified `get_valid_newunit()` in `lfortran_intrinsic_custom.f90` to
  allocate negative unit identifiers starting from `-10` down to `-1000`, 
  safely avoiding common standard units (`-1`, `0`, `5`, `6`).
- Fixed `newunit_int_1` overflow bounds checking to safely validate the
  lower limit (`< -128`) instead of the upper limit, and optimized it
  to call `get_valid_newunit()` only once.
- Updated the C runtime IO arrays (`seq_unf_pending`, etc.) to double 
  their capacity (`MAXUNITS * 2`) and introduced a `STATE_INDEX` macro. 
  This seamlessly offsets negative unit numbers to safe array indices 
  without breaking existing performance or structs.
- Replaced the hardcoded `0` to `999` iteration loop inside `_lfortran_inquire` 
  with a dynamic iteration over `unit_to_file` tracking, resolving `INQUIRE(FILE=)` 
  test failures.

### DIFF
--- a/integration_tests/file_open_09.f90
+++ b/integration_tests/file_open_09.f90
@@ -1,0 +1,18 @@
+program open_newunit
+  use iso_fortran_env
+  implicit none
+
+  integer :: io_lun
+
+  open (newunit=io_lun, file='open_newunit.dat', status='replace', form='formatted')
+  print *, "PASS"
+
+! Test for the first few conditions specified by F2023 12.5.6.13:
+  if (io_lun >= -1) error stop 'newunit >= -1'
+  if (io_lun == ERROR_UNIT) error stop 'newunit == ERROR_UNIT'
+  if (io_lun == INPUT_UNIT) error stop 'newunit == INPUT_UNIT'
+  if (io_lun == OUTPUT_UNIT) error stop 'newunit == OUTPUT_UNIT'
+
+  close (io_lun, status='delete')
+
+end program
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -5520,33 +5520,34 @@ int32_t last_index_used = -1;
 struct UNIT_FILE unit_to_file[MAXUNITS];
 
 // Pending bytes for sequential unformatted records per unit (shared by all types).
-static int32_t seq_unf_pending[MAXUNITS];
-static int32_t seq_unf_record_len[MAXUNITS];
+#define STATE_INDEX(u) ((u) >= 0 ? (u) : (MAXUNITS - (u)))
+static int32_t seq_unf_pending[MAXUNITS*2];
+static int32_t seq_unf_record_len[MAXUNITS*2];
 
 // Pending list-directed null-repeat state per unit.
-static int32_t lf_list_dir_null_remaining[MAXUNITS];
+static int32_t lf_list_dir_null_remaining[MAXUNITS*2];
 
 // Child I/O mode per unit (for DTIO: suppress record terminators in child writes).
-static int32_t lf_child_io_mode[MAXUNITS];
+static int32_t lf_child_io_mode[MAXUNITS*2];
 
 static inline void seq_unf_state_reset(int32_t unit_num) {
-    if (unit_num < 0 || unit_num >= MAXUNITS) return;
-    seq_unf_pending[unit_num] = 0;
-    seq_unf_record_len[unit_num] = 0;
+    if (unit_num <= -MAXUNITS || unit_num >= MAXUNITS) return;
+    seq_unf_pending[STATE_INDEX(unit_num)] = 0;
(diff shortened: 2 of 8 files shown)


Jatinagarwal24:
Hi @certik Pls review this.
Pls tell if any changes are required.

src/runtime/custom/lfortran_intrinsic_custom.f90
```diff
@@ -10,11 +10,11 @@ module lfortran_intrinsic_custom
 
 contains
     function get_valid_newunit() result(unit)
-        integer, parameter :: LUN_MIN=0, LUN_MAX=1000
+        integer, parameter :: LUN_MIN=-1000, LUN_MAX=-10
```
certik, line 13: Is this file still being used? If so, please open up an issue to migrate the logic here into ASR itself, as we handle other intrinsics.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -5520,47 +5520,48 @@ int32_t last_index_used = -1;
 struct UNIT_FILE unit_to_file[MAXUNITS];
 
 // Pending bytes for sequential unformatted records per unit (shared by all types).
-static int32_t seq_unf_pending[MAXUNITS];
-static int32_t seq_unf_record_len[MAXUNITS];
+#define STATE_INDEX(u) ((u) >= 0 ? (u) : (MAXUNITS - (u)))
+static int32_t seq_unf_pending[MAXUNITS*2];
+static int32_t seq_unf_record_len[MAXUNITS*2];
```
certik, line 5525: Please document what is going on here, and how this should be used.

integration_tests/file_open_09.f90
```diff
@@ -0,0 +1,18 @@
...
+  use iso_fortran_env
+  implicit none
+
+  integer :: io_lun
+
+  open (newunit=io_lun, file='open_newunit.dat', status='replace', form='formatted')
+  print *, "PASS"
+
+! Test for the first few conditions specified by F2023 12.5.6.13:
+  if (io_lun >= -1) error stop 'newunit >= -1'
```
certik, line 11: You are right, the io_lun should be negative.  I didn't know that.

certik (approved):
I think this is fine. You can document the runtime in a subsequent PR.

Jatinagarwal24:
Hi @certik I think this is good to merge. I will do the documentation in the next pr.
Thanks!

src/runtime/custom/lfortran_intrinsic_custom.f90
Jatinagarwal24: Yes, this file is definitely still being used. Currently, ast_body_visitor.cpp explicitly generates a subroutine call to the newunit interface defined here to fetch the unit number. As requested, I have opened an issue to track migrating this logic natively into ASR so we can eventually remove it from the custom runtime file : 
https://github.com/lfortran/lfortran/issues/11424

PR #11405, opened 2026-05-08
fix: handle ArrayPhysicalCast correctly in realloc target resolution

HarshitaKalani:


assem2002 (approved):
I just resolved conflicts.
Looks good to me

PR #11410, opened 2026-05-09
fix: deallocate components of intent(out) array-of-struct dummies

HarshitaKalani:


src/libasr/pass/insert_deallocate.cpp
```diff
@@ -503,6 +509,121 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
+                    var_expr_full, idx_vars, al, xx.m_symtab);
+
+                Vec<ASR::stmt_t*> innermost_body;
+                innermost_body.reserve(al, 1);
+
+                SymbolTable* sym_table_of_struct = struct_type->m_symtab;
+                while (sym_table_of_struct != nullptr) {
+                    for (auto& struct_member : sym_table_of_struct->get_scope()) {
+                        if (ASR::is_a<ASR::Variable_t>(*struct_member.second) &&
+                            ASRUtils::is_allocatable(
```
assem2002: We could have a non-allocatable that needs to go through this functionality again

assem2002: like this 
```.f90
program p
   type tt2
    integer, allocatable :: xx
   end type tt2

   type tt
      type(tt2) :: internal
   end type

   type(tt) :: instance(10)

   allocate(instance(1)%internal%xx)
   call ss(instance)
   contains 
   subroutine ss(kk)
      type(tt), intent(out) :: kk(10)
      print *, allocated(kk(1)%internal%xx)
   end subroutine
   
end
```

assem2002: If test above fails with this PR, please separate the logic into 2 functions -- One handling the cleanup of a struct (I think we already have similar thing in that pass), the other handles the array of struct (it'll just loop and use the first function) and they should call each other to handle cases like [arrayOFstruct, struct has an array].

assem2002: If you need this to get unblocked on some package, Open an issue for the above and feel free to merge.

assem2002:
I even this we should introduce like a node that we call `ResetVariable` that does what the fortran standards says :
- Deallocate the variable or the internals if exist (We could use a subClass of the finalizer with virtual functions)
- Set any members to its initial state if provided

certik:
The CI fails.

certik:
@assem2002 does this fit with your insert deallocate refactor?

certik (approved):
I think it's fine. We can iterate on this later as needed.

PR #11414, opened 2026-05-09
fix : Do not free the Pointer in derived type during finalization

### LINKED ISSUE #11401: Double free on code that contains a pointer as a derived-type component
The following could be a regression, as I did not notice this behavior previously with LFortran.

`$ cat mytypemod.f90 `
```
module mytypemod

   type :: MyType
      real(8), pointer :: ptr(:)
   end type MyType

   interface MyType
      procedure :: constructor
   end interface MyType
   
contains

   function constructor() result(self)
      type(MyType) :: self   
   end function constructor

end module mytypemod
```
`$ cat wrappermod.f90 `
```
module wrappermod
   
   use mytypemod, only: MyType
   
   type :: WrapperType
      private
   end type WrapperType
   
   interface WrapperType
      procedure :: constructor
   end interface WrapperType

contains
   
      class(MyType), intent(in) :: obj
      type(WrapperType)         :: self
      self%obj = obj
   end function constructor

end module wrappermod
```
`$ cat main.f90 `
```
program main

   use wrappermod, only: WrapperType
   use mytypemod,  only: MyType
   
   call wrapping()
   
contains
   
   subroutine wrapping()
      class(WrapperType), allocatable :: wrap
      wrap = WrapperType(MyType())
   end subroutine wrapping

end program main
```
`$ lfortran mytypemod.f90 wrappermod.f90 main.f90 `
```
free(): double free detected in tcache 2
Aborted (core dumped)

```

AnshMehta1:
Fixes #11401

This PR fixes a double-free issue during finalization of pointer components in derived types. The crash occurred because pointer descriptors were shallow-copied during assignment (self%obj = obj), causing multiple objects to free the same descriptor memory during finalization. The fix prevents freeing pointer descriptor during finalization to avoid this issue.

This can lead to small descriptor-memory leaks. But other option is to create a deep copy during assignments instead of shallow copy which again leads to high memory utilization and so, I think that this PR provides a feasible approach to solve this double free issue.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2403,6 +2403,7 @@ RUN(NAME derived_types_138 LABELS gfortran llvm)
 RUN(NAME derived_types_139 LABELS gfortran llvm)
 RUN(NAME derived_types_140 LABELS gfortran llvm)
 RUN(NAME derived_types_141 LABELS gfortran llvm)
+RUN(NAME derived_types_142 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME derived_types_apply_clip_01 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
 
--- a/integration_tests/derived_types_142.f90
+++ b/integration_tests/derived_types_142.f90
@@ -1,0 +1,58 @@
+module mytypemod_derived_types_142
+
+   type :: MyType
+      real(8), pointer :: ptr(:)
+   end type MyType
+
+   interface MyType
+      procedure :: constructor
+   end interface MyType
+   
+contains
+
+   function constructor() result(self)
+      type(MyType) :: self   
+   end function constructor
+
+end module mytypemod_derived_types_142
+
+
+module wrappermod_derived_types_142
+   
+   use mytypemod_derived_types_142, only: MyType
+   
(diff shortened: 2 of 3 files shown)


certik:
Leaks memory.

certik:
> This can lead to small descriptor-memory leaks.

We can't have memory leaks.

> But other option is to create a deep copy during assignments instead of shallow copy which again leads to high memory utilization

I implemented this option in https://github.com/lfortran/lfortran/pull/11417. Can you explain how it leads to high memory utilization? 

I think each struct has to have its own descriptor, the descriptor can never be shared by two variables, as that fundamentally doesn't work.

AnshMehta1:
> > This can lead to small descriptor-memory leaks.
> 
> We can't have memory leaks.
> 
> > But other option is to create a deep copy during assignments instead of shallow copy which again leads to high memory utilization
> 
> I implemented this option in #11417. Can you explain how it leads to high memory utilization?
> 
> I think each struct has to have its own descriptor, the descriptor can never be shared by two variables, as that fundamentally doesn't work.

I was talking about high memory utilization with respect to current approach of shallow copying. I thought, for every assignment, if we create a new descriptor, then it can lead to a lot of memory usage. Other than that, I am fine with deep copying approach. I was also favouring deep copy as it would fix a lot of other fundamental issues.

PR #11421, opened 2026-05-09
Assert 1D restriction and perfect divisibility for component arrays

### LINKED ISSUE #9784: Improve associate bug fix
We should assert to avoid non integer -- `struct_size % component_size == 0`

_Originally posted by @assem2002 in https://github.com/lfortran/lfortran/pull/9707#discussion_r2744274314_
            

I don't actually know. Do you think this is correct? If not let's add an assert to handle only 1D for now

_Originally posted by @assem2002 in https://github.com/lfortran/lfortran/pull/9707#discussion_r2744283023_
            

" I think it would've been been better to use that exact same approach but with visit_StructInstanceMember itself."

srinjoy933:
fixes #9784

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8906,11 +8906,16 @@ public:
         // Get rank from base array
         int n_dims = ASRUtils::extract_n_dims_from_ttype(base_struct_array_type);
 
+        LCOMPILERS_ASSERT(n_dims == 1 && "Multi-dimensional component arrays are not yet supported");
+
         // Compute the stride multiplier: sizeof(struct) / sizeof(component)
         // This is the number of component-sized elements between consecutive struct elements
         llvm::DataLayout data_layout(module->getDataLayout());
         uint64_t struct_size = data_layout.getTypeAllocSize(struct_llvm_type);
         uint64_t component_size = data_layout.getTypeAllocSize(component_llvm_type);
+    
+        LCOMPILERS_ASSERT(component_size > 0 && struct_size % component_size == 0);
+
         uint64_t stride_multiplier = struct_size / component_size;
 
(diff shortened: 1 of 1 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8957,11 +8957,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
 
+        LCOMPILERS_ASSERT(n_dims == 1 && "Multi-dimensional component arrays are not yet supported");
+
         // Compute the stride multiplier: sizeof(struct) / sizeof(component)
         // This is the number of component-sized elements between consecutive struct elements
         llvm::DataLayout data_layout(module->getDataLayout());
         uint64_t struct_size = data_layout.getTypeAllocSize(struct_llvm_type);
         uint64_t component_size = data_layout.getTypeAllocSize(component_llvm_type);
+    
+        LCOMPILERS_ASSERT(component_size > 0 && struct_size % component_size == 0);
```
kmr-srbh:
```suggestion
        LCOMPILERS_ASSERT(component_size > 0);
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8957,11 +8957,16 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         // Get rank from base array
         int n_dims = ASRUtils::extract_n_dims_from_ttype(base_struct_array_type);
 
+        LCOMPILERS_ASSERT(n_dims == 1 && "Multi-dimensional component arrays are not yet supported");
```
kmr-srbh:
```suggestion
        LCOMPILERS_ASSERT(n_dims == 1);
```

srinjoy933:
hey @kmr-srbh , I have made the changes as you have suggested. Thank you for your suggestion

PR #11429, opened 2026-05-10
prototype adding support of real_128 using pointer approach

opixdown:


tests/reference/llvm-allocate_03-495d621.stdout
```diff
@@ -2,8 +2,8 @@
 source_filename = "LFortran"
 target datalayout = "..."
 
-%string_descriptor = type <{ i8*, i64 }>
-%array.3 = type { i32*, i64, i32, i8, i8, i8, i8, i64, [3 x %dimension_descriptor] }
+%string_descriptor = type <{ ptr, i64 }>
+%array.3 = type { ptr, i64, i32, i8, i8, i8, i8, i64, [3 x %dimension_descriptor] }
```
certik: I think you updated reference results with LLVM >= 16. You need to update them with LLVM 11.

PR #11432, opened 2026-05-10
fix: --detect-leaks misses module-level allocatables across separate compilation units

### LINKED ISSUE #11242: Fix leak detector for module variables
The test code in https://github.com/lfortran/lfortran/pull/11240 requires manual deallocation, otherwise the leak detector fails.

When leak detection is enabled, we should free all module variables.

---

### Summary
With `--separate-compilation`, allocatable variables declared at module scope leak under `--detect-leaks` if the compilation unit defining the program does not directly `use` the module. The leaks appear even though the user code is correct Fortran (module-level allocatables are not required by the standard to be finalized at program termination, so no user `deallocate` is needed).

### Reproducer
`mod.f90`:
```fortran
module m
   integer, allocatable :: d(:)
end module
subroutine sub()
   use m
   allocate(d(2))
end subroutine
```

`prog.f90`:
```fortran
program p
   interface
      subroutine sub()
      end subroutine
   end interface
   call sub()
end program
```

```
$ lfortran --separate-compilation --detect-leaks mod.f90 prog.f90 -o x && ./x
   ==> LEAK: 8 bytes
TOTAL LEAKS FOUND : 1, 8 bytes total
```

If `prog.f90` instead `use m`s the module directly, no leak is reported. If both files are concatenated into a single source, no leak is reported either.

### Root cause
`asr_to_llvm.cpp` `visit_Program`, lines ~6088-6097:

```cpp
if (compiler_options.detect_leaks) {
    SymbolTable* tu = ASRUtils::get_tu_symtab(x.m_symtab);
    for (auto &kv : tu->get_scope()) {
        if (ASR::is_a<ASR::Module_t>(*kv.second)) {
            llvm_symtab_finalizer.finalize_symtab(
                ASR::down_cast<ASR::Module_t>(kv.second)->m_symtab);
        }
    }
}
```

This walks only modules present in the **program file's** translation-unit symtab. Under `--separate-compilation`, the program file is parsed without seeing the contents of modules used only by externally-linked subroutines, so those modules are not visited and their global allocatables are never freed before `dbg_report` runs.

### Note on user-defined FINAL procedures
This is a leak-reporting issue, not a semantic bug. User-defined FINAL procedures are not invoked on module-level allocatables at program termination by lfortran (whole-program build either) — and neither does Flang. F2018 §7.5.6.3 doesn't mandate finalization of module variables at program end (they don't go out of scope), so this matches expected behavior. If a user wants their FINAL invoked, they must explicitly `deallocate` the variable.

### Suggested fix
Emit a per-module finalizer function (e.g. `_lfortran_module_finalize_<modname>`) into each module's compilation unit when `--detect-leaks` is on. Register it via `llvm.global_dtors` with a high priority so it runs before `dbg_report`, or via `atexit()` registered from a `llvm.global_ctors` ctor — choosing priorities so `dbg_report` runs last.

srinjoy933:
fixes #11242


srinjoy933:
hey @certik  @kmr-srbh @assem2002 could you please once go through this PR when you get a chance

The fix emits a per-module finalizer function _lfortran_module_finalize_<modname>() in visit_Module and registers it via llvm.global_dtors at priority 200, and moves dbg_report to global_dtor at priority 255 so it always runs after all module finalizers.

Regarding the CI failures — the 39 failing DETECT_LEAKS tests such as goto_02_DETECT_LEAKS, global_allocatable_01_DETECT_LEAKS, arrays_87_DETECT_LEAKS and others are not related to the components touched by this PR. This PR only modifies the module finalizer emission logic in visit_Module and the dbg_report registration in visit_Program inside asr_to_llvm.cpp.

 Could you please check if there is any issue with the implementation approach as the tests are failing and suggest any changes ?

srinjoy933:
https://github.com/lfortran/lfortran/actions/runs/25636854707/job/75250179720?pr=11432
Screenshot (674)
Screenshot (675)

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8962,15 +8998,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         // Get rank from base array
         int n_dims = ASRUtils::extract_n_dims_from_ttype(base_struct_array_type);
 
-        LCOMPILERS_ASSERT(n_dims == 1 );
```
assem2002, line 9247: If this isn't a blocker. Please revert it

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8962,15 +8998,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
 
         // Compute the stride multiplier: sizeof(struct) / sizeof(component)
         // This is the number of component-sized elements between consecutive struct elements
         llvm::DataLayout data_layout(module->getDataLayout());
         uint64_t struct_size = data_layout.getTypeAllocSize(struct_llvm_type);
         uint64_t component_size = data_layout.getTypeAllocSize(component_llvm_type);
     
-        LCOMPILERS_ASSERT(component_size > 0);
```
assem2002, line 9255: Here also

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6218,13 +6254,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                fn = llvm::Function::Create(function_type,
-                    llvm::Function::ExternalLinkage, "dbg_report", module.get());
+            if (!fn) {
+                llvm::FunctionType *ft = llvm::FunctionType::get(
+                    llvm::Type::getVoidTy(context),false);
+                fn = llvm::Function::Create(ft,
+                    llvm::GlobalValue::ExternalLinkage, "dbg_report", module.get());
             }
-            builder->CreateCall(fn, {});
+            llvm::appendToGlobalDtors(*module, fn, 255);
```
assem2002:
```suggestion
            llvm::appendToGlobalDtors(*module, fn, 0);
```
Apparently `low_priority = low_number` -- before that we used to call the memory dbg report before actually cleaning anything (and it would exit the whole program)

assem2002:
I tested that locally, and it fails with 25 instead of 39.
I can see some test are failing due to double finalization (as I think). Feel free to investigate those, and let me know if you need any help.
and also thanks for working on this, That's much cleaner approach to use

src/libasr/codegen/asr_to_llvm.cpp
assem2002: and maybe use some const values, instead of those random numbers.

PR #11437, opened 2026-05-10
fix: formatted internal character read and IDL loop variables

### LINKED ISSUE #11433: Formatted READ: Multi-record internal file not working
```
program do_implied3
  implicit none

! With lfortran, compile with --use-loop-variable-after-loop.

  character(60), parameter :: input_data(2) =  &
      ['246801357912345678901234      ', '.10203040506070809010E+0233.33']
  character(60) :: str_data(2)
  real :: rdata(5), r_expected(5)
  integer :: i

  str_data = input_data
  rdata = 0.0
  i = -42
  read (str_data,100) (rdata(i),i=1,size (rdata))
  print *, 'i after read  =', i, merge (': pass', ': fail', i == size (rdata) + 1)

  r_expected = [246.8, 135.79, 1.234567890E13, 10.20304, 33.33]
  print *, 'rdata values:'
  do, i=1, size (rdata)
    print *, 'rdata(', i, ') =', rdata(i),  &
        merge (': pass', ': fail', abs (rdata(i) - r_expected(i)) < 0.0001)
  end do

100   format (2F5.2, F14.0 / E25.20, F5.2)

end program
$
```
```
$ lfortran --use-loop-variable-after-loop do_implied3.f90
i after read  = -42 : fail
rdata values:
rdata( 1 ) = 0.00000000 : fail
rdata( 2 ) = 0.00000000 : fail
rdata( 3 ) = 0.00000000 : fail
rdata( 4 ) = 0.00000000 : fail
rdata( 5 ) = 0.00000000 : fail
$
```
```
$ gfortran do_implied3.f90
$ ./a.out
 i after read  =           6 : pass
 rdata values:
 rdata(           1 ) =   246.800003     : pass
 rdata(           2 ) =   135.789993     : pass
 rdata(           3 ) =   1.23456790E+13 : pass
 rdata(           4 ) =   10.2030401     : pass
 rdata(           5 ) =   33.3300018     : pass
$
```

Jatinagarwal24:
fixes #11433

This commit resolves three closely related defects that caused `READ` statements
from internal character strings with implied-do loops (IDLs) to fail.

1. Incorrect Temporary Target Variables (ast_body_visitor.cpp)
   Bug: Formatted `READ` statements were incorrectly wrapping their target 
   variables in a `StringFormat` ASR node. Since `StringFormat` is strictly for 
   `WRITE` operations, ASR passes (like array_struct_temporary) generated 
   temporary variables for the targets. The runtime then read data into these 
   temporaries instead of the actual user variables (e.g., `rdata`).
   Fix: Prevent `READ` statements from wrapping target variables in `StringFormat`. 
   Instead, correctly map `m_fmt` to the resolved format string directly in both 
   the standard and deferred format resolution paths.

2. Internal Read Position Reset (asr_to_llvm.cpp)
   Bug: For internal file reads with implied-do loops, the codegen was falling back 
   to the `emit_formatted_read_with_idl` path which generated one runtime call per 
   loop iteration. Each call created a new `InputSource`, resetting the string read 
   position to 0 on every iteration and causing it to read the first value repeatedly.
   Fix: Expand simple array implied-do loops `(array(i), i=1,N)` into direct array 
   arguments. This falls through to the non-IDL codegen path, ensuring all values 
   are populated in a single runtime call, advancing the internal string position correctly.

3. IDL Loop Variable Update (asr_to_llvm.cpp)
   Bug: The loop variable used in the `READ` IDL wasn't updated to its final state 
   (end + step) after the read completed.
   Fix: Explicitly emit the final assignment to the loop variable (`i = end + step`)
   after IDL expansion to match expected Fortran semantics.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2875,6 +2875,7 @@ RUN(NAME file_66 LABELS gfortran llvm)
 RUN(NAME file_67 LABELS gfortran llvm)
 RUN(NAME file_68 LABELS gfortran llvm)
 RUN(NAME file_69 LABELS gfortran llvm)
+RUN(NAME file_70 LABELS gfortran llvm EXTRA_ARGS --use-loop-variable-after-loop)
 
 RUN(NAME file_close_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME file_close_02 LABELS gfortran llvm)
--- a/integration_tests/file_70.f90
+++ b/integration_tests/file_70.f90
@@ -1,0 +1,23 @@
+program do_implied3
+  implicit none
+
+  character(60), parameter :: input_data(2) =  &
+      ['246801357912345678901234      ', '.10203040506070809010E+0233.33']
+  character(60) :: str_data(2)
+  real :: rdata(5), r_expected(5)
+  integer :: i
+
+  str_data = input_data
+  rdata = 0.0
+  i = -42
+  read (str_data,100) (rdata(i),i=1,size (rdata))
+  if (i /= size (rdata) + 1) error stop 1
+
+  r_expected = [246.8, 135.79, 1.234567890E13, 10.20304, 33.33]
+  do, i=1, size (rdata)
+    if (abs (rdata(i) - r_expected(i)) >= 0.0001) error stop 2
+  end do
+
+100   format (2F5.2, F14.0 / E25.20, F5.2)
+
+end program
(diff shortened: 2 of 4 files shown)


Jatinagarwal24:
Hi @certik pls review this and suggest if any changes are required or not.

HarshitaKalani (approved):
I think this is good. Thanks!

Jatinagarwal24:
hi @HarshitaKalani is this good to merge?

certik (approved):
I think it's fine.

PR #11442, opened 2026-05-10
fix:Improve an error message for unary operators

### LINKED ISSUE #9708: Improve an error message for unary operators
Shouldn't the error rather say that the `.bad.` operator is defined, but a procedure in it was not found?

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/9685#discussion_r2728963951_

srinjoy933:
fixes #9708


kmr-srbh (changes requested):
No error references changed, so it looks like this error might not be getting triggered by any of our existing error tests. So we need to add a test case inside `tests/errors/continue_compilation_1.f90` that triggers this error.

It also does not affect the test reference for the error test added in #9685, so I think this does not yet fix the linked issue.

PR #11446, opened 2026-05-11
fix: Replace Function Param in type through Function call in array struct temporary pass

### LINKED ISSUE #11413: ICE with elemental character function call using array arguments
## MRE

```fortran id="mre6419"
module bar
  implicit none
contains

elemental function trim_append(xx,yy) result(xy)
  character(len=*), intent(in) :: xx,yy
  character(len=len(xx)+len(yy)) :: xy
  xy = trim(xx) // yy
end function

function same(xx) result(yy)
  character(len=*), intent(in) :: xx(:)
  character(len=len(xx)) :: yy(size(xx))
  yy = [xx]
end function

subroutine foo(labels)
  character(len=*), intent(in) :: labels(:)
  print *, "size(labels)=", size(labels)
end subroutine

subroutine xmain()
  call foo(trim_append(["a"], same(["b"])))
end subroutine

end module bar

program main
  use bar
  call xmain()
end program
```

## Output

```text id="out6419"

Internal Compiler Error:
LCompilersException: visit_FunctionParam() not implemented
```

```text id="out6420"

 size(labels)=           1
```

AnshMehta1:
Fixes #11413


src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -167,6 +167,49 @@ inline ASR::ttype_t* normalize_string_temporary_type(
...
+void replace_function_param_in_type(Allocator &al, ASR::ttype_t *t, ASR::FunctionCall_t *f_call) {
+
+    if (!t) return;
+    ASRUtils::ReplaceFunctionParamWithArg replacer(al, f_call->m_args, f_call->n_args);
+
+    switch (t->type) {
+        case ASR::ttypeType::String: {
+            ASR::String_t *s = ASR::down_cast<ASR::String_t>(t);
+            if (s->m_len) {
+                s->m_len = replacer.replace_FunctionParam_with_arg(s->m_len);
```
assem2002: This duplicates the expression everytime even when there's no functionParam to replace.
 Can you modify this class `ASRUtils::ReplaceFunctionParamWithArg` to make it only duplicate the functionCall-return-type node only when we hit a functionParam? + I think you can then remove this switch block and call the class rightaway and maybe make the class return if the user passes a type other than (string or array) 

What do you think?

assem2002: *sorry I reviewed it yesterday without submitting, and got it drafted.

AnshMehta1: > This duplicates the expression everytime even when there's no functionParam to replace. Can you modify this class `ASRUtils::ReplaceFunctionParamWithArg` to make it only duplicate the functionCall-return-type node only when we hit a functionParam? + I think you can then remove this switch block and call the class rightaway and maybe make the class return if the user passes a type other than (string or array)
> 
> What do you think?

Yes, I will make the necessary changes. That would make the things more cleaner and simpler. Thanks for the reviews.

AnshMehta1: @assem2002, I got rid of switch and moved things directly to the class `ReplaceFunctionParamWithArg`.
I didn't resolve that expression duplication thing even when there's no functionParam. Because, it involves calling `replace_expr` two times which would cause two traversals. One to detect any FunctionParam. Once a function param is detected, then we need another traversal after duplicating our original expr for replacing FunctionParam. 

whats your say on this??

assem2002: We still got the same performance issue.
You can merge this one atleast we have the logic compacted in one class.
Please open an issue mentioning the performance issue so we keep being aware of it and can fix it later.

assem2002: Reported here https://github.com/lfortran/lfortran/issues/11494

PR #11456, opened 2026-05-11
fix: memory leak for intent(out) structs with nested allocatable members

### LINKED ISSUE #11455: Bug (Memory-Leak) : intent(out) stores null in the backend without freeing
```.f90
program p
   type tt2
    integer, allocatable :: xx
end type tt2
   type tt
      type(tt2) :: internal
   end type
   type(tt) :: instance

   allocate(instance%internal%xx)
   call ss(instance)
   contains 
   subroutine ss(kk)
      type(tt), intent(out) :: kk
      print *, allocated(kk%internal%xx)
   end subroutine
end
```
```
   ==> LEAK: 4 bytes

TOTAL LEAKS FOUND : 1, 4 bytes total

```

Jatinagarwal24:
fixes #11455

This commit resolves a 4-byte memory leak that occurred when passing a 
derived type with intent(out) that contains allocatable components nested 
within non-allocatable sub-structs.

The fix addresses two separate issues working in tandem:

1. ASR Pass (`insert_deallocate.cpp`): 
   The `IntentOutDeallocateVisitor` previously only checked for direct 
   members of an intent(out) struct. It has been updated to use a recursive 
   traversal (`collect_nested_deallocs`) to correctly find and generate 
   ExplicitDeallocate statements for allocatable components at any nesting 
   depth.

2. LLVM Codegen (`asr_to_llvm.cpp`): 
   During local variable declaration (`declare_vars`), the codegen recursively 
   initializes struct members via `allocate_array_members_of_struct`. However, 
   the recursive calls failed to propagate the `is_intent_out` flag. This 
   caused nested allocatable members of intent(out) dummy arguments to be 
   incorrectly null-initialized at function entry, zeroing out the pointer 
   before the newly inserted deallocation statements could free the memory.
   The `is_intent_out` flag is now properly passed through the recursive calls.

### DIFF
--- a/src/libasr/pass/insert_deallocate.cpp
+++ b/src/libasr/pass/insert_deallocate.cpp
@@ -453,13 +453,14 @@ public:
                     }
                 }
 
-                // Recursively collect deallocation statements for all allocatable
-                // components nested at any depth within the struct hierarchy.
-                // parent_expr is the base access expression (e.g., arg, arg%member, etc.)
-                // parent_sym is the symbol corresponding to parent_expr
-                std::function<void(ASR::Struct_t*, ASR::expr_t*, ASR::symbol_t*)>
-                    collect_nested_deallocs = [&](ASR::Struct_t* st,
-                        ASR::expr_t* parent_expr, ASR::symbol_t* parent_sym) {
+                 // collect deallocation statements for all allocatable
+                // components nested at any depth within the struct hierarchy are handled recursively.
+                // @param parent_expr is the base access expression (e.g., arg, arg%member, etc.)
+                // @param parent_sym is the symbol (structSymbol) corresponding to parent_expr
+                // @param self pass self lambda to recurise call.
+                // TODO : Clean into regular function
+                auto collect_deallocs_of_struct = [&](auto self,
+                    ASR::Struct_t* st, ASR::expr_t* parent_expr, ASR::symbol_t* parent_sym) -> void {
                     ASR::Struct_t* current_st = st;
                     SymbolTable* sym_table = current_st->m_symtab;
                     while (sym_table != nullptr) {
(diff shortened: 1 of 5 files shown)


src/libasr/pass/insert_deallocate.cpp
```diff
@@ -453,56 +453,82 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
-                                al, loc, nullptr, is_allocated, if_body.p, if_body.n, nullptr, 0));
-
-                            // Wrap in optional presence check if needed (reuse var_expr from above)
-                            ASR::stmt_t* wrapped_stmt = wrap_optional_check(loc, var_expr, arg_var->m_presence, if_stmt);
-                            dealloc_stmts.push_back(al, wrapped_stmt);
+                // Recursively collect deallocation statements for all allocatable
+                // components nested at any depth within the struct hierarchy.
+                // parent_expr is the base access expression (e.g., arg, arg%member, etc.)
+                // parent_sym is the symbol corresponding to parent_expr
+                std::function<void(ASR::Struct_t*, ASR::expr_t*, ASR::symbol_t*)>
```
assem2002:
```suggestion
                auto
```

assem2002 (approved):
Looks good to me. Thanks!
Feel free to merge after CI passes.

assem2002:
3rd party code fails. 
It'd be awesome if you can extract an MRE for it, and register it into our integration tests.

assem2002 (approved):
LGTM! Thanks.

assem2002:
I think a related PR is https://github.com/lfortran/lfortran/pull/11410, It should handle another case also :+1:

Issue #11476, opened 2026-05-12
Build OpenFAST


s1291:
[OpenFAST](https://github.com/OpenFAST/openfast) is a large code base (roughly 702k lines according to `cloc` command).

I am documenting the attemps to build OpenFAST using lfortran to identify the current limitations:

- [x] NWTC relies on some F77 behavior that requires using:
  - [x]  `--implicit-interface` flag. 
  - [x]  `--fixed-form-infer` flag. 
  - [x]  `--implicit-typing` flag. 
  - [x]  `--legacy-array-sections` flag.
- [x] Fixed: https://github.com/lfortran/lfortran/issues/11479
- [x] Fixed: https://github.com/lfortran/lfortran/issues/11478 
- [x] https://github.com/lfortran/lfortran/issues/11482
- [x] Fixed: https://github.com/lfortran/lfortran/issues/11495
- [x] https://github.com/lfortran/lfortran/issues/11498
- [ ] https://github.com/lfortran/lfortran/issues/11499
- [ ] https://github.com/lfortran/lfortran/issues/11500
- [ ] #11501 
- [ ] #11502 
- [x] #11503 
- [x] #11504

PR #11480, opened 2026-05-12
fix: compute correct byte size for ArrayConstant of c_ptr/c_funptr to…

### LINKED ISSUE #11479: crash with std::bad_alloc when using a module that broadcasts c_null_ptr into an array of c_ptr
The parent issue: https://github.com/lfortran/lfortran/issues/11476

lfortran crashes with `std::bad_alloc` when a program does use on a separately compiled module that contains an array of c_ptr or c_funptr assigned from c_null_ptr / c_null_funptr. Compiling the module alone succeeds; the crash fires on the use when the .mod file is loaded.

Example:

```fortran
! m.f90
module m
   use, intrinsic :: iso_c_binding, only: c_ptr, c_null_ptr
   implicit none
contains
   subroutine clear(a)
      type(c_ptr), intent(out) :: a(4)
      a = c_null_ptr
   end subroutine
end module
```

```fortran
! p.f90
program p
   use m, only: clear
   use, intrinsic :: iso_c_binding, only: c_ptr
   type(c_ptr) :: a(4)
   call clear(a)
end program
```

```bash
$ lfortran -c m.f90
$ lfortran p.f90 m.o -o demo
std::exception: std::bad_alloc
```

s1291:
fixes: https://github.com/lfortran/lfortran/issues/11479

compiling a module that assigns c_null_ptr to a c_ptr array then using it from a seprately compiled program, crashes the compiler with std::bad_alloc. This fix would allow us to build NWTC (part of OpenFAST, see: https://github.com/lfortran/lfortran/issues/11476)


HarshitaKalani (approved):
lgtm. Thanks!

certik (approved):
I think that looks good, thanks!

PR #11485, opened 2026-05-13
fix: allocate underlying assumed-rank dummy on realloc-LH S in select rank

HarshitaKalani:


certik (approved):
Does this correctly fail without realloc LHS?

PR #11486, opened 2026-05-13
fix: pad each element when initializing char array with shorter literal

HarshitaKalani:


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2921,6 +2921,25 @@ llvm::Value* LLVMUtils::handle_global_nonallocatable_stringArray(Allocator& al,
...
+                int64_t base = i * src_elem_len;
+                int64_t take = std::min(src_elem_len,
+                    (int64_t)sequence.size() - base);
+                if (take > 0) padded.append(sequence, base, take);
+                else take = 0;
+                padded.append(actual_len - take, ' ');
+            }
+            sequence = padded;
+        }
+    }
```
kmr-srbh: I think we already have a function in AST -> ASR that does this? The function is `adjust_character_length()`.

certik (approved):

PR #11490, opened 2026-05-13
fix: refresh dependencies after dead code removal

### LINKED ISSUE #11489: ICE with `--fast` option due to failure of dead code removal
`$ cat test.f90 `
```
module test
contains
   subroutine with_side_stepped_code()
      integer(4), parameter :: iflag = 0
      integer(4) :: j
      real(8) :: tmp
      if ( iflag.eq.1 ) then
         do j = 1, 10
            tmp = real(j) / real(10)
         end do
      else
         tmp = 0.d0
      end if
   end subroutine with_side_stepped_code
end module test
```
`$ lfortran -c --fast --std=f23 --separate-compilation test.f90 `
```
ASR verify pass error: Function with_side_stepped_code doesn't depend on _lcompilers_real_i32_22817806770928_dIVvJu8HRfdg4W15EdTQ2o but is found in its dependency list.
LCompilersException: Verify failed in the pass: dead_code_removal

```

Jatinagarwal24:
The dead_code_removal pass can remove a constant-false branch that contains
intrinsic helper calls such as real(i). After the branch is removed, the
function body no longer references the generated helper symbol, but the
function dependency list can still contain it.

This leaves stale dependencies in ASR and causes verification to fail with
messages like:

Function with_side_stepped_code doesn't depend on _lcompilers_real_i32
but is found in its dependency list.

Recompute ASR dependencies after dead_code_removal rewrites the function body.
fixes #11489

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -566,6 +566,7 @@ RUN(NAME if_03 FAIL LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c
 RUN(NAME if_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)
 RUN(NAME if_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp wasm c fortran)
 RUN(NAME if_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # select case range
+RUN(NAME if_07 LABELS gfortran llvm EXTRA_ARGS --fast NO_FAST)
 
 RUN(NAME while_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)
 RUN(NAME while_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp wasm c fortran)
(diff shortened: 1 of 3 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -566,6 +566,7 @@ RUN(NAME if_03 FAIL LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c
 RUN(NAME if_04 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp x86 wasm c fortran)
 RUN(NAME if_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp wasm c fortran)
 RUN(NAME if_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # select case range
+RUN(NAME if_07 LABELS gfortran llvm EXTRA_ARGS --fast NO_FAST)
```
kmr-srbh:
```suggestion
RUN(NAME if_07 LABELS gfortran llvm)
```

kmr-srbh (approved):
Just a minor change, otherwise looks good.

integration_tests/CMakeLists.txt
Jatinagarwal24: Done.
Thanks!

Jatinagarwal24:
Hi @certik Is there any changes required?

certik (approved):

PR #11493, opened 2026-05-13
enh: allow data statement list without comma also

### LINKED ISSUE #11109: Commas in DATA statement should be optional
This case comes from an older state estimation code once used in space mission planing: https://codeberg.org/ivan-pi/netlib-fortran/src/commit/0fc6f3488b321c0d2c511ef54571e7f3260848ee/a/esl/tmprnt.f#L144

Also discussed in: https://github.com/lfortran/lfortran/issues/11090#issuecomment-4253581260

```fortran
      PROGRAM TMPRNT
      IMPLICIT DOUBLE PRECISION (A-H,O-Z)
      CHARACTER*8 V(4), VFMT(6), V2I
      CHARACTER*9 V2A   
C   
      DATA  V/'('' ''','A16,1PX,',' ','D17.7)'/,
     1 VFMT/'6','17X,5','34X,4','51X,3','68X,2','85X,1'/
     2 V2A/'A16,1P1X,'/,V2I/'I5,1P2X,'/ 
C
      print '(4(A8,:,1X))', V
      print *, VFMT
      print *, V2A
      END   
```

LFortran says,

```
syntax error: Token 'v2a' (of type 'identifier') is unexpected here
 --> /app/example.f90:8:8
  |
8 |      2 V2A/'A16,1P1X,'/,V2I/'I5,1P2X,'/ 
  |        ^^^ 
```

The expected output is something close to:

```
 (' '    A16,1PX,        D17.7)  
 6       17X,5   34X,4   51X,3   68X,2   85X,1   
 A16,1P1X,
```

The problem appears to stem from a missing comma in the DATA statement, between VFMT and V2A. I checked the standard (Section 8.6.7, J3/24-007) and the comma is actually optional,

> R840 *data-stmt*       is `DATA` *data-stmt-set* [ [ , ] *data-stmt-set* ]
> R841 *data-stmt-set* is *data-stmt-object-list* / *data-stmt-value-list* /

### LINKED ISSUE #11110: GOTO within numbered do-loop - code generation error
Related to https://github.com/lfortran/lfortran/issues/11090#issuecomment-4253366975

The problem appears within the function:

```fortran
      subroutine balance_companion(n,a)
      implicit none
      integer n
      double precision a(n,n)
      integer b
*>>>
*     parameter(b=16)
      parameter(b=2)
*>>>
Comment b is the base of the floating point representation on the machine.
*       It is 16 for base 16 float : for example, IBM system 360/370.
*       It is 2  for base  2 float : for example, IEEE float.
*End comment;
*
      integer b2
      parameter(b2=b**2)
      integer i,j
      double precision c,f,g,r,s
      logical noconv

* If n<=1, then do nothing.
      if(n.le.1)then
        return
      endif
* iteration:
    1 continue
        noconv=.false.
        do 100 i=1,n
C Generic code.
C         c=0.d0
C         r=0.d0
C         do j=1,i-1
C           c=c+abs(a(j,i))
C           r=r+abs(a(i,j))
C         enddo
C         do j=i+1,n
C           c=c+abs(a(j,i))
C           r=r+abs(a(i,j))
C         enddo
C begin specific code. Touch only non-zero elements of companion.
          if(i.ne.n)then
            c=abs(a(i+1,i))
          else
            c=0.d0
            do j=1,n-1
              c=c+abs(a(j,n))
            enddo
          endif
          if(i.eq.1)then
            r=abs(a(1,n))
          elseif(i.ne.n)then
            r=abs(a(i,i-1))+abs(a(i,n))
          else
            r=abs(a(i,i-1))
          endif
C end specific code.
*
          if(c.eq.0.d0.or.r.eq.0.d0) goto 100
*
          g=r/b
          f=1.d0
          s=c+r
*label L3:
    3     if(c.lt.g)then
            f=f*b
            c=c*b2
            goto 3
          endif
          g=r*b
*label L4:
    4     if(c.ge.g)then
            f=f/b
            c=c/b2
            goto 4
          endif
*Comment the preceding four lines may be replaced by a machine language
*        procedure computing the exponent sig such that 
*        sqrt(r/(c*b))<=b**sig<sqrt(r*b/c). Now balance;
          if((c+r)/f .lt. 0.95*s) then
            g=1.d0/f
            noconv=.true.
C Generic code.
C           do j=1,n
C             a(i,j)=a(i,j)*g
C           enddo
C           do j=1,n
C             a(j,i)=a(j,i)*f
C           enddo
C begin specific code. Touch only non-zero elements of companion.
            if(i.eq.1)then
              a(1,n)=a(1,n)*g
            else
              a(i,i-1)=a(i,i-1)*g
              a(i,n)=a(i,n)*g
            endif
            if(i.ne.n)then
              a(i+1,i)=a(i+1,i)*f
            else
              do j=1,n
                a(j,i)=a(j,i)*f
              enddo
            endif
C end specific code.
          endif
* Comment: 
*     the j joops may be done by exponent modification
*     in machine Language;
  100   enddo
      if(noconv) goto 1
      end
```

which fails with

```
Referring to a basic block in another function!
  br label %goto_target
```

The error dissappears if I comment the `if (...) goto 100` statement. 
If I'm not mistaken the `goto 100` is supposed to act like the `cycle` statement here.

jinangshah21:
Fixes #11109 
Fixes #11110


certik (approved):
Looks good.

Issue #10884, opened 2026-03-28
Compile M_strings


HarshilShah1804:
Package Link: https://github.com/urbanjost/M_strings

Current Compilation Status:
```bash
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/M_strings$ fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
 Warning: Dependency M_io does not enforce module naming, but project does. 
 Warning: Dependency M_CLI2 does not enforce module naming, but project does. 
M_strings__chars.f90                   done.
M_strings.F90                          done.
gfortran_bug.f90                       done.
M_io.F90                               failed.
[  1%] Compiling...

semantic error: Invalid argument `stream` supplied
   --> build/dependencies/M_io/src/M_io.F90:443:10 - 454:41
    |
443 |             inquire(unit=lun,                                                                               &
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
    |
454 |         &   iostat=ios,err=999,iomsg=message)
    | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
<ERROR> Compilation failed for object " build_dependencies_M_io_src_M_io.F90.o "
<ERROR> stopping due to failed compilation
STOP 1

```

PR #11496, opened 2026-05-13
Fix : intrinsic subroutine `Mvbit` with array arguments 

assem2002:
towards #10884
Make `array_op` pass generate a loop over the `MVbits` subrotuine call to iterate over each array element.
**Basically** : Handle some intrinsicImpureSubroutine as same as we handle IntrinsicElementalFunction


src/libasr/pass/array_op.cpp
```diff
@@ -7,6 +7,7 @@
 #include <libasr/pass/pass_utils.h>
 #include <libasr/pass/intrinsic_function_registry.h>
 #include <libasr/pass/intrinsic_array_function_registry.h>
+#include <libasr/pass/intrinsic_subroutines.h>
```
assem2002, line 10: I don't like this as the header has so much code -- but I don't know how to access the `intrinsicImpureSubroutine` class inside

assem2002: I think we can move the enum class into file `intrinsic_subroutine_registery`

certik (approved):

src/libasr/pass/array_op.cpp
certik: https://github.com/lfortran/lfortran/issues/11529

PR #11511, opened 2026-05-14
fix: valid assignment from real pointer array is rejected 

### LINKED ISSUE #11503: valid assignment from real pointer array is rejected
Parent issue: #11476 

```fortran
program p
   use iso_c_binding, only: C_FLOAT
   implicit none
   real(C_FLOAT), target  :: data(4) = [1.0, 2.0, 3.0, 4.0]
   real(C_FLOAT), pointer :: src(:) => null()
   real(8)                :: dst(4)
   src => data
   dst = src 
   print *, dst
end program
```

```sh
$ gfortran -o demo p.f90

$  ./demo
   1.0000000000000000        2.0000000000000000        3.0000000000000000        4.0000000000000000

$ lfortran -o demo p.f90
 --> p.f90:8:4
  |
8 |    dst = src   ! real(4) pointer  ->  real(8) array
  |    ^^^ type mismatch (real[:] and real[:] pointer)
  |
8 |    dst = src   ! real(4) pointer  ->  real(8) array
  |    ^^^^^^^^^ type mismatch (real[:] and real[:] pointer)
```

opixdown:
fixes #11503

### DIFF
--- a/integration_tests/pointer_17.f90
+++ b/integration_tests/pointer_17.f90
@@ -1,0 +1,15 @@
+program pointer_17
+    use iso_c_binding, only: c_float
+    implicit none
+    real(c_float), target :: data(4) = [1.0, 2.0, 3.0, 4.0]
+    real(c_float), pointer :: src(:) => null()
+    real(8) :: dst(4)
+    real(8) :: expected(4)
+
+    expected = [1.0d0, 2.0d0, 3.0d0, 4.0d0]
+    src => data
+    dst = src
+
+    if (any(dst /= expected)) error stop
+    print *, dst
+end program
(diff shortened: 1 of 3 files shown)


jinangshah21 (approved):

integration_tests/pointer_17.f90
```diff
@@ -0,0 +1,15 @@
+program pointer_17
+    use iso_c_binding, only: c_float
```
s1291, line 2: Could we use `real32` and `real64` from `iso_fortran_env` here instead of `c_float`? The issue does not seem specific to `iso_c_binding`, but to assignment from a real pointer array to a real array of a different kind.

assem2002: Yeah it's quite unrelated. but there's no harm at using it here, it even stretches the test more.

PR #11512, opened 2026-05-14
fix: Convert the template name to lowercase before adding it to symbol table

### LINKED ISSUE #11454: Capital letters in a Template name will cause a segmentation fault or Nullpointer error
No matter where in the name the capital letter(s) are, it will end in a segmentation fault when compiled with a regularly downloaded lfortran compiler. Using a fork of the project and compiling using the interactive ./src/bin/lfortran method will result in a Nullptr error:
`AssertFailed: f != nullptr`

[nullptrError.txt](https://github.com/user-attachments/files/27598199/nullptrError.txt)

A template name with only lowercase will compile and print "Test drive" without issue.

```
module Uppercase_Error_m
    implicit none
    private
    public :: error_T

    template error_T(T, U)
        type, deferred :: T
        type, deferred :: U
        private
        public :: mycopy
      contains
        subroutine mycopy(lhs, rhs)
            type(T), intent(out) :: lhs
            type(T), intent(in) :: rhs
            lhs = rhs
        end subroutine
    end template

end module

program p
    print *, "Test drive"
end program
```

AnshMehta1:
Fixes #11454

### DIFF
--- a/integration_tests/template_06.f90
+++ b/integration_tests/template_06.f90
@@ -1,0 +1,23 @@
+module mod_template_06
+    implicit none
+    private
+    public :: error_T
+
+    template error_T(T, U)
+        type, deferred :: T
+        type, deferred :: U
+        private
+        public :: mycopy
+      contains
+        subroutine mycopy(lhs, rhs)
+            type(T), intent(out) :: lhs
+            type(T), intent(in) :: rhs
+            lhs = rhs
+        end subroutine
+    end template
+
+end module
+
+program template_06
+    print *, "Test drive"
+end program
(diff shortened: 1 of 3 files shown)


integration_tests/template_06.f90
```diff
@@ -0,0 +1,23 @@
...
+            type(T), intent(out) :: lhs
+            type(T), intent(in) :: rhs
+            lhs = rhs
+        end subroutine
+    end template
+
+end module
+
+program template_06
+    print *, "Test drive"
```
certik, line 22: I checked that this test fails in main, so it's a good test.

However in the subsequent PR can you also instantiate this template in the main program, to ensure everything works?

certik (approved):
Thanks for the fix!

PR #11513, opened 2026-05-14
fix: valid end label inside if block is not recognized

### LINKED ISSUE #11499: valid end label inside if block is not recognized
Parent issue: #11476 

lfortran does not recognize a valid `end=` label on a statement inside the same `if` block.

Example:

```fortran
subroutine s()
   integer :: i

   if (i > 0) then
      read(86, *, end=20) i
20    print *, 'a'
   end if
end subroutine
```

```bash
$ lfortran -c p.f90
semantic error: end= label 20 is not defined
 --> p.f90:5:23
  |
5 |       read(86, *, end=20) i
  |                       ^^

```

`gfortran -c p.f90` accepts this code.

opixdown:
fixes  #11499

### DIFF
--- a/integration_tests/read_93.f90
+++ b/integration_tests/read_93.f90
@@ -1,0 +1,13 @@
+program read_93
+   call s()
+contains
+   subroutine s()
+      integer :: i
+      open(86, file="/dev/null")
+      if (i > 0) then
+         read(86, *, end=20) i
+20       print *, 'a'
+      end if
+      close(86)
+   end subroutine
+end program
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -1336,16 +1336,97 @@ public:
             body.p, body.size(), nullptr, 0));
     }
 
-    void collect_labels() {
-        labels.clear();
-        if (starting_m_body == nullptr) return;
-        for (size_t i = 0; i < starting_n_body; ++i) {
-            int64_t label = stmt_label(starting_m_body[i]);
+    void collect_labels_from_body(AST::stmt_t** m_body, size_t n_body) {
+        for (size_t i = 0; i < n_body; ++i) {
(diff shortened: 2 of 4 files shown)


integration_tests/read_93.f90
```diff
@@ -0,0 +1,13 @@
+program read_93
+   call s()
+contains
+   subroutine s()
+      integer :: i
+      open(86, file="/dev/null")
```
s1291: `/dev/null` is posix/linux/macOS specific, it won't work on windows

s1291:
> fixes #11499

Thanks for the fix. Please let's coordinate in the future to avoid duplicating efforts 😄

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1336,17 +1336,98 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+            } else if (AST::is_a<AST::Block_t>(*m_body[i])) {
+                AST::Block_t* s = AST::down_cast<AST::Block_t>(m_body[i]);
+                collect_labels_from_body(s->m_body, s->n_body);
+            } else if (AST::is_a<AST::ChangeTeam_t>(*m_body[i])) {
+                AST::ChangeTeam_t* s = AST::down_cast<AST::ChangeTeam_t>(m_body[i]);
+                collect_labels_from_body(s->m_body, s->n_body);
+            } else if (AST::is_a<AST::Critical_t>(*m_body[i])) {
+                AST::Critical_t* s = AST::down_cast<AST::Critical_t>(m_body[i]);
+                collect_labels_from_body(s->m_body, s->n_body);
+            }
```
certik, line 1420: Can you explain how this works? This code cannot possibly be tested by the simple test in this PR.

I would say either let's add comprehensive testing to ensure this is actually correct, or figure out an algorithm that is general and short, and then we can just have the simple test above.

PR #11522, opened 2026-05-14
Refactor : Remove unnecessary class in `insert_deallocate`

assem2002:
I ran tests locally and it just passes.
If CI passes, that's great!


s1291:
I only skimmed this quickly, but did you already check whether it is compatible with `do_loop_7` in the integration tests?

assem2002:
What do you exactly mean. The test passes at the CI, so I think we're good then, right?

PR #11525, opened 2026-05-15
fix: scope array constructor implied-do variables

### LINKED ISSUE #11448: implied do variable in parameter array constructor reported as undeclared
## MRE

```fortran id="mre7521"
program test
  integer ii
  integer iii

  call two

contains

  subroutine one
    i = 99
    ii = 99
    iii = 999
  end subroutine

  subroutine two
    i = 0
    ii = 0
    iii = 0
    call one

    if (i .ne. 0) stop 1
    if (ii .ne. 99) stop 2
    if (iii .ne. 999) stop 3
  end subroutine

end program
```

## Output

```text id="out7521"

 --> t.f90:3:42
  |
```

Jatinagarwal24:
fixes #11448
Array constructor implied-do control variables were declared in the
surrounding symbol table when implicit typing was enabled. This allowed a
variable such as `i` from `(/ (i, i = 1, 10) /)` to leak into the enclosing
program scope, so contained procedures could incorrectly resolve their own
implicit `i` to that constructor loop variable and produce wrong runtime
behavior.

Use a hidden temporary symbol for array-constructor implied-do variables and
expose it only while semantically visiting the implied-do expression. Restore
the previous visible symbol afterward so the control variable remains local to
the implied-do construct.

Add an integration test covering implied-do parameter constructors together
with contained procedures that use implicit local variables.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3278,6 +3278,7 @@ RUN(NAME implied_do_loops36 LABELS gfortran llvm)
 
 RUN(NAME implied_do_loops37 LABELS gfortran llvm)
 RUN(NAME implied_do_loops38 LABELS gfortran llvm EXTRA_ARGS --use-loop-variable-after-loop)
+RUN(NAME implied_do_loops39 LABELS gfortran llvm EXTRA_ARGS --implicit-typing)
 
 RUN(NAME minpack_01 LABELS gfortran llvm_rtlib EXTRAFILES
         minpack_01_module.f90 minpack_01_func.f90)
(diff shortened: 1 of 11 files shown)


Jatinagarwal24:
Hi @certik Pls review this .

certik:
I think this fix is too complicated. Why not something like https://github.com/lfortran/lfortran/pull/11534 ?

certik:
Here is even simpler logic: https://github.com/lfortran/lfortran/pull/11535.

Jatinagarwal24:
Closing this as this pr #11535 got merged !

PR #11527, opened 2026-05-15
fix class array dealloc

### LINKED ISSUE #11526: ICE on deallocation of allocatable polymorphic array component
## MRE

```fortran id="mre6183"
program mre

  type base_t
    integer, allocatable :: buf(:)
  end type base_t

  type holder_t
    class(base_t), allocatable :: obj(:)
  end type holder_t

  type(holder_t) :: x

  allocate(x%obj(2))
  deallocate(x%obj)

end program mre
```

## Output

```text id="out6183"

Internal Compiler Error:
AssertFailed: ds->getType()->isPointerTy()
```

## gfortran Behavior

```text id="out6184"

(compiles successfully)
```

ambikeesshh:
deallocate() could hit an LLVM assertion for allocatable polymorphic arrays when the array descriptor was passed as a value during finalization.

I made the finalization path use an addressable descriptor before reading array fields, and added an integration test for the issue case.

Tested with `cmake --build build -j2` and `cd integration_tests && ./run_tests.py -b llvm -t derived_types_144`.

Fixes #11526

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2421,6 +2421,7 @@ RUN(NAME derived_types_apply_clip_01 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
 RUN(NAME derived_types_143 LABELS gfortran llvm
     EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME derived_types_144 LABELS gfortran llvm)
 
 RUN(NAME derived_type_with_default_init LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


kmr-srbh (commented):
The added test already works with `main`. Can you please verify if the error is still reproducible?

ambikeesshh:
Thanks, you're right. I checked current main with LLVM enabled, and both the original MRE from #11526 and this test pass there.

I'll look for a current reproducer; if I can't find one, I'll close this PR.

ambikeesshh:
I couldn't reproduce this anymore.

I checked current main with LLVM enabled, the original MRE from #11526, this added test, and a few nearby variants around class/class(*) allocatable array deallocation. They all pass here. I also checked the old PR base locally and the original MRE passes there too.

Closing this for now since I don't have a current failing reproducer.

PR #11534, opened 2026-05-16
Scope array-constructor implied-do variables

### LINKED ISSUE #11448: implied do variable in parameter array constructor reported as undeclared

certik:
Array-constructor implied-do control variables were declared in the surrounding symbol table when implicit typing was enabled. A variable such as 'i' from (/ (i, i = 1, 10) /) leaked into the enclosing program scope, so contained procedures could incorrectly resolve their own implicit 'i' to that constructor loop variable and produce wrong runtime behavior.

Track array-constructor nesting with a small depth counter set by visit_ArrayInitializer. In visit_ImpliedDoLoop, if the loop variable did not exist in the enclosing scope before this implied-do was visited and we are inside an array constructor, rename the implicitly created symbol to a hidden, unique name after the visit. ImpliedDoLoop value nodes refer to the symbol by pointer, so the rename keeps them valid while removing the user-facing name from the enclosing scope.

Add integration test implied_do_loops39.f90 covering parameter array constructors with implied-do variables alongside contained procedures that use implicit local variables of the same names.

Fixes #11448

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -1961,6 +1961,7 @@ public:
     bool _declaring_variable = false;
     bool _processing_common_block_object = false;
     bool is_implicit_interface = false;
+    int array_initializer_depth = 0;
     Vec<ASR::stmt_t*> *current_body = nullptr;
 
     std::map<std::string, ASR::ttype_t*> implicit_dictionary;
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10636,6 +10637,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         bool use_descriptorArray = false; // Set to true if any argument has no fixed size (array arguments).
         ASR::ttype_t* extracted_type { type ? ASRUtils::extract_type(type) : nullptr };
         size_t n_elements = 0;
+        struct ArrayInitializerDepthGuard {
+            int &depth;
+            ArrayInitializerDepthGuard(int &depth) : depth(depth) { depth++; }
+            ~ArrayInitializerDepthGuard() { depth--; }
+        } array_initializer_depth_guard(array_initializer_depth);
```
certik, line 10644: I don't like this complexity. I think here is an even simpler fix: https://github.com/lfortran/lfortran/pull/11535.

PR #11535, opened 2026-05-16
Fix implicit-typing leak of array-constructor implied-do iterator

### LINKED ISSUE #11448: implied do variable in parameter array constructor reported as undeclared

certik:
With --implicit-typing, an implied-do iterator used only inside an array constructor (e.g. `(/ (i, i = 1, 10) /)`) was implicitly declared in the enclosing program/module scope. Contained procedures then host-associated to that variable instead of getting their own implicit local declaration, producing wrong runtime behavior.

Per the Fortran standard, an array-constructor implied-do variable has scope local to the implied-do. After processing the implied-do, rename the implicitly-created symbol to a unique
`__implicit_idl_<name>` so that name lookups in contained procedures no longer resolve to it and instead trigger proper local implicit typing. Existing ASR `Var_t` references remain valid since they are direct pointers.

Adds integration test `implicit_typing_12` covering the bug.

Fixes https://github.com/lfortran/lfortran/issues/11448.


certik:
@Jatinagarwal24 thanks for the fix!

PR #11537, opened 2026-05-16
fix: support case insensitive max/min in DO CONCURRENT reduction

### LINKED ISSUE #11536: DO CONCURRENT reduce(max:...) fails with lowercase operator
### Minimal Reproducer

```fortran
program test
    integer :: i
    real :: mx
    mx = -1.0

    do concurrent (i=1:10) reduce(max:mx)
        mx = max(mx, real(i))
    end do

    print *, mx
end program
```

### Error

```text
(base) opixdown@Atharvs-MacBook-Air lfortran %  t.f90
syntax error: Unsupported operation in reduction
 --> a.f90:7:35
  |
7 |     do concurrent (i=1:10) reduce(max:mx)
  |                                   ^^^
```
### Gfortran
```
(base) opixdown@Atharvs-MacBook-Air lfortran % gfortran a.f90&& ./a.out                                                                 
   10.0000000    
```

opixdown:
fixes #11536

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -543,6 +543,7 @@ RUN(NAME slash_init_01 LABELS gfortran llvm)
 
 RUN(NAME max_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME max_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
+RUN(NAME max_03 LABELS  llvm llvm_wasm llvm_wasm_emcc) #gfortran does not support reduce()
 RUN(NAME min_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME min_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME minmax_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c fortran)
--- a/src/lfortran/parser/semantics.h
+++ b/src/lfortran/parser/semantics.h
@@ -923,18 +923,17 @@ static inline char** REDUCE_ARGS(Allocator &al, const Vec<ast_t*> args)
 static inline reduce_opType convert_id_to_reduce_type(
         const Location &loc, const ast_t *id, LCompilers::diag::Diagnostics &diagnostics)
 {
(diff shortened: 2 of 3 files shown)


src/lfortran/parser/semantics.h
```diff
@@ -924,15 +924,17 @@ static inline reduce_opType convert_id_to_reduce_type(
         const Location &loc, const ast_t *id, LCompilers::diag::Diagnostics &diagnostics)
 {
     std::string s_id = down_cast2<Name_t>(id)->m_id;
-    if (s_id == "MIN") {
+    std::string s_upper = s_id;
+    std::transform(s_upper.begin(), s_upper.end(), s_upper.begin(), ::toupper);
```
certik: Cannot you use the existing `tolower` function that we already have, and use lowercase below?

opixdown: yes we should do that only

src/lfortran/parser/semantics.h
```diff
@@ -923,16 +923,17 @@ static inline char** REDUCE_ARGS(Allocator &al, const Vec<ast_t*> args)
 static inline reduce_opType convert_id_to_reduce_type(
         const Location &loc, const ast_t *id, LCompilers::diag::Diagnostics &diagnostics)
 {
-    std::string s_id = down_cast2<Name_t>(id)->m_id;
-    if (s_id == "MIN") {
+        std::string s_id = down_cast2<Name_t>(id)->m_id;
+        std::string s_lower = LCompilers::to_lower(s_id);
+        if (s_lower == "min" ) {
```
certik, line 928: The indentation is incorrect.

integration_tests/CMakeLists.txt
```diff
@@ -543,6 +543,7 @@ RUN(NAME slash_init_01 LABELS gfortran llvm)
 
 RUN(NAME max_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME max_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
+RUN(NAME max_03 LABELS  llvm llvm_wasm llvm_wasm_emcc) #gfortran does not support reduce()
```
certik, line 546:
```suggestion
RUN(NAME max_03 LABELS llvm llvm_wasm llvm_wasm_emcc) #gfortran does not support reduce()
```

certik (approved):
You can fix it in a subsequent PR.

PR #11539, opened 2026-05-16
feat: add support for `IAND` `IEOR` `IOR`

### LINKED ISSUE #11538: DO CONCURRENT: IAND/IOR/IEOR reduction operators not supported
## Description

`IAND`, `IOR`, `IEOR`

Currently, using any of these bitwise operators in a reduction clause fails  at the parser level.

## MRE

```fortran
program test_ieor
    integer :: i
    integer :: x
    x = 0

    do concurrent (i=1:4) reduce(ieor:x)
        x = ieor(x, i)
    end do
    print *, x
end program
```

## Current behavior
```
base) opixdown@Atharvs-MacBook-Air lfortran % lfortran a.f90                                                                       
syntax error: Unsupported operation in reduction
 --> a.f90:7:34
  |
7 |     do concurrent (i=1:4) reduce(ieor:x)
  |                                  ^^^^ 

```

### Gfortran
```
(base) opixdown@Atharvs-MacBook-Air lfortran % gfortran a.f90 && ./a.out                                                            
           4
```

opixdown:
fixes #11538


src/lfortran/parser/semantics.h
```diff
@@ -924,10 +924,16 @@ static inline reduce_opType convert_id_to_reduce_type(
         const Location &loc, const ast_t *id, LCompilers::diag::Diagnostics &diagnostics)
 {
     std::string s_id = down_cast2<Name_t>(id)->m_id;
-    if (s_id == "MIN" ) {
+    if (s_id == "MIN") {
         return reduce_opType::ReduceMIN;
     } else if (s_id == "MAX") {
         return reduce_opType::ReduceMAX;
+    } else if (s_id == "IAND" || s_id == "iand") {
```
certik, line 931: This will fail for `Iand` and other cases.

Rather, we use `tolower` to normalize all such strings. So likely you have to do `s_id = tolower(...)` and then just compare lower case.

certik (approved):
I think this is good. Can you please fix the lowercase issue in the next PR?

PR #11545, opened 2026-05-16
Fix : Remove zero-initialization hack for call-args

### LINKED ISSUE #11139: Figure out how not to zero-initialize memory for call-args
This is fine for now to fix all leaks, but we need to investigate why this is needed, as we shouldn't need to zero-initialize memory. This applies to all struct-type allocas, not just containers.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/11137#discussion_r3105360007_

srinjoy933:
fixes #11139


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -465,20 +465,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         }
         // Need to create a new alloca
         llvm::AllocaInst* alloca = llvm_utils->CreateAlloca(type, nullptr, "call_arg_value");
-        // Zero-initialize struct allocas (e.g. list, tuple) at entry block
+         // Zero-initialize struct allocas (e.g. list, tuple) at entry block
```
```suggestion
        // Zero-initialize struct allocas (e.g. list, tuple) at entry block
```

certik (approved):

PR #11556, opened 2026-05-18
Introduce the Wasm kernel

anutosh491:


wasm-build1.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+# WASM build for xlfortran (JupyterLite kernel).
+# Prerequisites:
+#   1. emsdk 4.0.9 activated through xeus-lfortran-wasm-host environment (see environment-wasm-host.yml)
+#   2. xeus-lfortran-wasm-host conda env created (environment-wasm-host.yml)
+#   3. ./build0.sh already run (generated parser/ASR headers in source tree)
```
anutosh491, line 6: Hey @Jatinagarwal24 here are the steps to reproduce this 

Step 0:  Install [micromamba](https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html#automatic-install)

Step 1 : 
```
micromamba create -f environment-wasm-build.yml -y
micromamba activate xeus-lfortran-wasm-build
```

Step 2 : 
```
micromamba create -f environment-wasm-host.yml \
--platform=emscripten-wasm32 \
-c https://prefix.dev/emscripten-forge-4x \
-c https://prefix.dev/conda-forge
```

Step 3 : Run build0.sh

Step 4 : Run wasm-build1.sh (I've named this like such so that it doesn't interfere with build1.sh but this can be changed accordingly and places at a better location if we have one)

Step 5 : Check if your prefix has xlfortran.js/wasm
```
micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge
micromamba activate xeus-lite-host
jupyter lite serve --XeusAddon.prefix=$PREFIX
```

This should give you a jupyterlite instance for you to try out !

Jatinagarwal24: Hi @anutosh491 This looks amazing I will try this out.
But Can you tell me what i need to fix in this pr now.

anutosh491: Ahhh no fix as such but 

1) Try the native kernel i.e the xeus-lfortan kernel's native build and cofirm it works

2) Then try the wasm kernel through the above instructions and confirm that works

3) Then add CI for the same. Very similar to what xeus-cpp will have (and technically replicating the instructions from above)

[main.yml](https://github.com/compiler-research/xeus-cpp/blob/main/.github/workflows/main.yml#L252-L312)

& 

[deploy-github-page.yml](https://github.com/compiler-research/xeus-cpp/blob/main/.github/workflows/deploy-github-page.yml)

So that once this PR goes in we would have a link to try out the interactive repl in the browser through github pages.

Jatinagarwal24: Ok cool I will check and add to CI.
Thanks !

Jatinagarwal24: image
Hey @anutosh491! I went through all the steps and everything works perfectly. Here is the update:



 **Task 1: Native Kernel Build**
      Confirmed working. (Note: Had to update the local conda environment to `nlohmann_json>=3.12.0` for `xeus-6.0.0` compatibility, then built successfully with `-DWITH_XEUS=yes`).
**Task 2: WASM Kernel Build**
      Confirmed working. Followed the micromamba instructions, built `xlfortran.js/wasm`, and successfully served the local JupyterLite instance.
**Task 3: Add CI Workflows**
      Added `Wasm-Kernel-CI.yml` and `deploy-github-page.yml`. Both workflows handle the two-stage build by first running `build0.sh` in the standard Linux env (for re2c/bison), then switching to the WASM env for the emscripten build.

The commits have been pushed to the `wasm-kernel` branch.

anutosh491: Thanks for checking it out !

build1.sh
```diff
@@ -19,4 +19,4 @@ cmake \
     -DCMAKE_CXX_FLAGS="${CXXFLAGS} -fdiagnostics-color=always" \
     -G Ninja \
     .
-cmake --build .
+cmake --build . --target install
```
anutosh491: Ohhh does this involve changes to build1.sh too ?

I thought about providing a separate wasm-build1.sh for the same reason (to not affect any native step just yet)

What I thought is we could guide the users in the Readme to do the env setups -> then run build0.sh and then wasm-build1.sh

anutosh491:
Made a small cleanup commit too !

.github/workflows/deploy-github-page.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      # ── Step 4: Build JupyterLite site ──
+      - name: Build JupyterLite
+        shell: bash -l {0}
+        run: |
+          set -ex
+          micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge -y
+          micromamba run -n xeus-lite-host jupyter lite build \
+              --XeusAddon.prefix=${{ env.PREFIX }} \
+              --output-dir dist
```
anutosh491, line 77: Okay so basically the point of deploy-github-page.yml is to give you a link which we shall be hosting in the README for users to try out. For eg we have this for xeus-cpp

Image

So what's better is if we can add some demo notebooks here just like we do for xeus-cpp

https://github.com/compiler-research/xeus-cpp/blob/main/.github/workflows/deploy-github-page.yml#L150-L154

I see that there are already some ipynb files that LFortran provides we can probably just share them or make new ones.

.github/workflows/deploy-github-page.yml
```diff
@@ -0,0 +1,100 @@
...
+        shell: bash -l {0}
+        run: |
+          set -ex
+          ./build0.sh
+
+      # ── Step 2: Set up the Emscripten/WASM build environment ──
+      - name: Install WASM build environment
+        shell: bash -l {0}
+        run: |
+          micromamba create -f environment-wasm-build.yml -y
```
anutosh491, line 48: Just curious if this is enough to activate the env too. Don't we need `micromamba activate` too ?

anutosh491:
I think this much work here is all we might like. 

1) Here we are using quite some "#ifdef Emscripten" checks. If we want we can split up files for native evaluator & wasm evaluator in separate .h/cpp files. And in cmake only include stuff based on if we're building with emscritpen or no. So yeah we can find some refactoring work here.

2) If this is good time to release (maybe even a patch release) we can do it. That would help me host the same build on emscripten-forge and once that is done I can add the lfortran Kernel to Jupyter's official website. That would be good traction and users in the community can then see that a Fortran repl exists that runs entirely in the browser !

3) Then we might want to look into unlocking Jupyter's Rich Display utilities (display html, latex, images, audio etc)

Check cells 14 onwards here : https://compiler-research.org/xeus-cpp-wasm/lab/index.html

and also the small pt notebook for ray tracing using SDL : https://compiler-research.org/xeus-cpp-wasm/lab/index.html

We've been able to use Jupyter widgets to do offscreen canvas + mouse click events in the C++ kernel

https://github.com/user-attachments/assets/b379da6a-081e-4cc3-b768-c05a19106659

We would need this for https://dev.lfortran.org/  to display the mandelbrot plot as a result of a cell output.

environment-wasm-host.yml
```diff
@@ -0,0 +1,10 @@
+name: xeus-lfortran-wasm-host
+channels:
+  - https://prefix.dev/emscripten-forge-4x
+  - https://prefix.dev/conda-forge
+dependencies:
+  - xeus-lite
+  - xeus
+  - nlohmann_json
+  - zlib
```
anutosh491: Check if we need zlib here. I guess I just added it for safety.

I disable it in wasm-build1.sh : https://github.com/lfortran/lfortran/pull/11556/changes#diff-b7cb7c3087fed649097921b32401a851e880624e3ef45f464d33827cadbeddeeR27

Jatinagarwal24:
Hi @anutosh491 this test `[Quick checks / Build LFortran to WASM and Upload (pull_request)]` which is failing with the following error 
We have to fix this right
/home/runner/ext/emsdk/node/18.20.3_64bit/bin/node
v18.20.3
+ node src/lfortran/tests/test_lfortran.js
  throw err;
  ^

Error: Cannot find module '/home/runner/work/lfortran/lfortran/src/lfortran/tests/test_lfortran.js'
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.20.3

anutosh491:
> We have to fix this right

Yess. Not sure why it fails tho :|

To be fair I see this 

https://github.com/lfortran/lfortran/blob/413e337a4eede9bdd05fae791982217f7f719768/.github/workflows/Quick-Checks-CI.yml#L373

This emsdk is quite outdated. Emscripten has already reached version 5. Package managers like emscripten-forge & pyodide use emscripten 4 (we have around 400-500 C++, R, Rust, python pkgs built using that and is pretty trusted and stable)

emsdk 3 comes with node 18
emsdk 4 comes with node 24 (which is much better for errors like memory out of bounds etc)

So instead of downloading emsdk 3 and activating it -> we can just fetch emsdk 4 from our environment-wasm-build.yml file isn't it ? Worth a shot !

Jatinagarwal24:
> > We have to fix this right
> 
> Yess. Not sure why it fails tho :|
> 
> To be fair I see this
> 
> https://github.com/lfortran/lfortran/blob/413e337a4eede9bdd05fae791982217f7f719768/.github/workflows/Quick-Checks-CI.yml#L373
> 
> This emsdk is quite outdated. Emscripten has already reached version 5. Package managers like emscripten-forge & pyodide use emscripten 4 (we have around 400-500 C++, R, Rust, python pkgs built using that and is pretty trusted and stable)
> 
> emsdk 3 comes with node 18 emsdk 4 comes with node 24 (which is much better for errors like memory out of bounds etc)
> 
> So instead of downloading emsdk 3 and activating it -> we can just fetch emsdk 4 from our environment-wasm-build.yml file isn't it ? Worth a shot !

Yes let me try.

anutosh491:
That being said 

We already have a [recipe](https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/lfortran/build.sh) building test_lfortran.js with emsdk 4 on emscripten-forge 

Everything passes for lfortran 0.63.0 : https://github.com/emscripten-forge/recipes/actions/runs/24813667787/job/72724239019#step:7:607

So yeah moving to emscripten 4 using the emsdk from the env should be harmless !

Jatinagarwal24:
> That being said
> 
> We already have a [recipe](https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/lfortran/build.sh) building test_lfortran.js with emsdk 4 on emscripten-forge
> 
> Everything passes for lfortran 0.63.0 : https://github.com/emscripten-forge/recipes/actions/runs/24813667787/job/72724239019#step:7:607
> 
> So yeah moving to emscripten 4 using the emsdk from the env should be harmless !

Pls check my latest commit 78766d5fb2e6b8929eeb25a88f9250d89da7dc31.I think that will fix it.

Jatinagarwal24:
> Made a small cleanup commit too !

Is your this commit got removed?
If yes sorry for that.I think i forgot to pull and directly forced pushed it.

anutosh491:
> Is your this commit got removed?
If yes sorry for that.I think i forgot to pull and directly forced pushed it.

Yes I think so. 

This was the diff. Make sure you apply it
```
anutosh491@Anutoshs-MacBook-Air lfortran % git diff
diff --git a/src/libasr/codegen/evaluator.cpp b/src/libasr/codegen/evaluator.cpp
index a41287a09..ba265c9a4 100644
--- a/src/libasr/codegen/evaluator.cpp
+++ b/src/libasr/codegen/evaluator.cpp
@@ -650,16 +650,6 @@ void WasmLFortranExecutor::add_module(std::unique_ptr<LLVMModule> lm, int eval_c
     llvm::raw_string_ostream lld_errs_stream(lld_errs);
     const ::lld::DriverDef WasmDriver = {::lld::Flavor::Wasm, &::lld::wasm::link};
     ::lld::Result Result = ::lld::lldMain(LinkerArgs, llvm::outs(), lld_errs_stream, {WasmDriver});
-    // lld::wasm::link calls cl::ResetAllOptionOccurrences() which clears all
-    // LLVM command-line option state (e.g. -wasm-enable-eh). Re-parse the
-    // stored args so the next TargetMachine creation sees the correct state.
-    if (!stored_llvm_args.empty()) {
-        std::vector<const char *> arg_ptrs;
-        arg_ptrs.push_back("xlfortran (restoring LLVM options)");
-        for (const std::string &arg : stored_llvm_args)
-            arg_ptrs.push_back(arg.c_str());
-        llvm::cl::ParseCommandLineOptions(arg_ptrs.size(), arg_ptrs.data());
-    }
     if (Result.retCode)
         throw LCompilersException("WasmLFortranExecutor: wasm-ld failed for "
                                   + stem + ": " + lld_errs_stream.str());
@@ -667,17 +657,15 @@ void WasmLFortranExecutor::add_module(std::unique_ptr<LLVMModule> lm, int eval_c
     void *handle = dlopen(wasmFile.c_str(), RTLD_NOW | RTLD_GLOBAL);
     if (!handle)
         throw LCompilersException(std::string("WasmLFortranExecutor: dlopen failed: ") + dlerror());
-    loaded_modules.push_back(handle);
+    (void)handle;
 }
 
 intptr_t WasmLFortranExecutor::get_symbol_address(const std::string &name)
 {
-    for (auto it = loaded_modules.rbegin(); it != loaded_modules.rend(); ++it) {
-        void *sym = dlsym(*it, name.c_str());
-        if (sym)
-            return reinterpret_cast<intptr_t>(sym);
-    }
-    throw LCompilersException("WasmLFortranExecutor: symbol not found: " + name);
+    void *sym = dlsym(RTLD_DEFAULT, name.c_str());
+    if (!sym)
+        throw LCompilersException("WasmLFortranExecutor: symbol not found: " + name);
+    return reinterpret_cast<intptr_t>(sym);
 }
 
 #endif // __EMSCRIPTEN__
diff --git a/src/libasr/codegen/evaluator.h b/src/libasr/codegen/evaluator.h
index 5ecc9d74f..763b9eba5 100644
--- a/src/libasr/codegen/evaluator.h
+++ b/src/libasr/codegen/evaluator.h
@@ -122,11 +122,7 @@ public:
 
 private:
     std::unique_ptr<llvm::LLVMContext> context;
-    // SmallString stored as std::string because SmallString is not in scope here
     std::string TempDir;
-    std::vector<void*> loaded_modules;
-    // LLVM cl options to re-apply after lld wipes them (mirrors xeus-cpp Wasm.cpp)
-    std::vector<std::string> stored_llvm_args;
 };
 
 #endif // __EMSCRIPTEN__
```

src/libasr/codegen/evaluator.cpp
```diff
@@ -560,4 +572,102 @@ std::string LLVMEvaluator::get_default_target_triple()
...
+
+    llvm::legacy::PassManager PM;
+    if (TM->addPassesToEmitFile(PM, ObjOut, nullptr, llvm::CodeGenFileType::ObjectFile))
+        throw LCompilersException("WasmLFortranExecutor: wasm32 backend cannot emit object file");
+    PM.run(*mod);
+    ObjOut.close();
+
+    std::vector<const char *> LinkerArgs = {
+        "wasm-ld", "-shared",
+        "--import-memory", "--stack-first", "--allow-undefined",
```
anutosh491:
```suggestion
        "--import-memory", "--experimental-pic", "--stack-first", "--allow-undefined",
```

anutosh491: We need this to counter this warning 

Image

We are currently using llvm 21.1.8 which needs `--experimental-pic`  & `--shared` working together.

Check here in the 21.x release branch : https://github.com/llvm/llvm-project/blob/release/21.x/lld/wasm/Driver.cpp#L805-L809

This flag has been removed upstream and in the upcoming versions of llvm we would need need this flag.

https://github.com/llvm/llvm-project/pull/196566/changes#diff-02986149f6567f53d19aafde4514e237f72072e3eb96c068adf00e6fc7a4fbc5L806

**A TODO can be added for its removal by llvm 23**

anutosh491:
Hey @Jatinagarwal24 could you please resolve the comments that you think have been addressed ?

Jatinagarwal24:
Hi @anutosh491 CI passed finally :)
Is there any fixes/changes still required?

Jatinagarwal24:
> Hey @Jatinagarwal24 could you please resolve the comments that you think have been addressed ?
> 

Sure i will do that.

anutosh491:
> Is there any fixes/changes still required?

Yeah couple unresolved comments I opened above can be addressed

environment-wasm-host.yml
Jatinagarwal24: You're right, zlib is not needed here. wasm-build1.sh passes -DWITH_ZLIB=no so it's never used in the WASM build. Safe to remove it.
I will do that.
Thanks!

build1.sh
Jatinagarwal24: Good catch! The build1.sh change was accidental — reverted it. The WASM kernel build uses its own separate wasm-build1.sh, so build1.sh stays untouched.
Thanks !

.github/workflows/deploy-github-page.yml
Jatinagarwal24: No, micromamba activate is not needed here. We use micromamba run -n xeus-lfortran-wasm-build in all subsequent steps (see the "Build WASM kernel" step at line 76), which temporarily activates the environment for just that command. micromamba run -n <env> <cmd> is the equivalent of activate + cmd + deactivate — it's actually the recommended approach in CI because it avoids shell activation issues.

.github/workflows/deploy-github-page.yml
Jatinagarwal24: Yes, there's a Demo.ipynb in the [lfortran-binder](https://gitlab.com/lfortran/web/lfortran-binder) GitLab repo. We could copy it into a notebooks/ directory and pass --contents notebooks/Demo.ipynb in the jupyter lite build step. Though some of its features (plotting, magic commands like %%showasm) may need updating for the new WASM kernel. Should we do this as a follow-up PR or include it here?

Jatinagarwal24:
> > Is there any fixes/changes still required?
> 
> Yeah couple unresolved comments I opened above can be addressed

Hi @anutosh491 I have resolved all the above comments.Kindly go through that.
Thanks.

.github/workflows/deploy-github-page.yml
anutosh491: Hmm well we might not want %%showasm but rather %%showwasm or something here !

Let's just start with a demo notebook with everything we can achieve (a subset of the one above I guess) and that way we know what work are we left with !

Makes sense ?

Jatinagarwal24: Yes that makes sense.Can you add that demo notebook please.

.github/workflows/Quick-Checks-CI.yml
```diff
@@ -350,61 +350,74 @@ jobs:
...
         run: |
             set -ex
-            source $HOME/ext/emsdk/emsdk_env.sh # Activate Emscripten
-            which node
-            node -v
-            node src/lfortran/tests/test_lfortran.js
+            micromamba run -n xeus-lfortran-wasm-build node src/lfortran/tests/test_lfortran.js
 
       - name: Upload to wasm_builds
+        if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
```
certik: This should not be needed, it's masking a bug in the script. The `ci/upload_lfortran_wasm.sh` already has this condition. It must be able to run on a PR, to ensure that everything works, except the final upload that only happens on main/tag.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -153,40 +165,47 @@ Result<FortranEvaluator::EvalResult> FortranEvaluator::evaluate(
...
         result.c32.im = r.imag();
     } else if (return_type == "complex8") {
-        std::complex<double> r = e.execfn<std::complex<double>>(run_fn);
+        std::complex<double> r = reinterpret_cast<std::complex<double>(*)()>(sym_addr(run_fn))();
         result.type = EvalResult::complex8;
         result.c64.re = r.real();
         result.c64.im = r.imag();
     } else if (return_type == "logical") {
-        int32_t r = e.execfn<int32_t>(run_fn);
+        int32_t r = reinterpret_cast<int32_t(*)()>(sym_addr(run_fn))();
```
certik: I'll have to understand what this is doing.

anutosh491: Ahh it's trying to fetch the symbol address (emscripten's dlfcn.h header file provides good support for dlopen and dlsym which can been put to use) and invoke it !

CMakeLists.txt
```diff
@@ -30,15 +30,15 @@ if (NOT CMAKE_BUILD_TYPE)
 endif ()
 
 # xeus does use `dynamic_cast` and thus needs rtti (otherwise it segfaults)
-if (NOT WITH_XEUS)
+if (NOT WITH_XEUS AND NOT XEUS_LFORTRAN_WASM_BUILD)
```
mcbarton, line 33: Why have `XEUS_LFORTRAN_WASM_BUILD` when Emscripten provides `Emscripten` inside cmake? In xeus-cpp we got rid of a similar naming option for being pointless https://github.com/compiler-research/xeus-cpp/pull/450

CMakeLists.txt
```diff
@@ -358,6 +368,62 @@ if (WITH_XEUS)
             PATTERN "*.in" EXCLUDE)
 endif()
 
+# XEUS_LFORTRAN_WASM_BUILD — build xlfortran.js/wasm JupyterLite kernel
+set(XEUS_LFORTRAN_WASM_BUILD no CACHE BOOL "Build xlfortran JupyterLite WASM kernel")
```
mcbarton, line 369: Why have XEUS_LFORTRAN_WASM_BUILD when Emscripten provides Emscripten inside cmake? In xeus-cpp we got rid of a similar naming option for being pointless compiler-research/xeus-cpp#450

.github/workflows/deploy-github-page.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Setup ncpus
+        run: |
+          echo "ncpus=$(nproc --all)" >> $GITHUB_ENV
```
mcbarton: You define this but don't use it as far as I can see.

.github/workflows/Wasm-Kernel-CI.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      # ── Step 2: Set up the Emscripten/WASM build environment ──
+      - name: Install WASM build environment
+        shell: bash -l {0}
+        run: |
+          micromamba create -f environment-wasm-build.yml -y
+
+      - name: Setup ncpus
+        run: |
+          echo "ncpus=$(nproc --all)" >> $GITHUB_ENV
```
mcbarton: You define this but don't use it as far as I can see.

.github/workflows/Wasm-Kernel-CI.yml
```diff
@@ -0,0 +1,92 @@
...
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      # ── Step 1: Run build0.sh to generate parser/ASR headers ──
+      # build0.sh needs re2c, bison, and python with numpy, which are
+      # NOT part of the wasm-build environment. We use the standard
+      # lfortran CI environment for this step.
+      - name: Install build0 prerequisites
+        uses: mamba-org/setup-micromamba@v2
```
mcbarton, line 41: There is newer version of this action.

wasm-build1.sh
```diff
@@ -0,0 +1,34 @@
...
+    -DWITH_STACKTRACE=no \
+    -DWITH_WHEREAMI=no \
+    -DWITH_ZLIB=no \
+    -DCMAKE_INSTALL_PREFIX="$PREFIX" \
+    -DCMAKE_FIND_ROOT_PATH="$PREFIX" \
+    -DCMAKE_PREFIX_PATH="$PREFIX" \
+    -DLLVM_DIR="$PREFIX/lib/cmake/llvm" \
+    -DLLD_DIR="$PREFIX/lib/cmake/lld"
+
+emmake make -C build-wasm install -j8
```
mcbarton: Hardcoded 8 number of jobs. Will differ from machine to machine. Using `nproc` will only work on unix systems unless the person on macos has coreutils installed.

wasm-build1.sh
```diff
@@ -0,0 +1,34 @@
...
+
+mkdir -p build-wasm
+
+# Use explicit -S/-B so cmake never confuses the source-tree's in-source
+# CMakeCache.txt (from build1.sh) with the wasm binary directory.
+emcmake cmake -S . -B build-wasm \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLFORTRAN_BUILD_ALL=no \
+    -DWITH_LLVM=yes \
+    -DXEUS_LFORTRAN_WASM_BUILD=yes \
```
mcbarton, line 21: Option feels unneeded

.github/workflows/deploy-github-page.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Generate parser/ASR headers (build0.sh)
+        shell: bash -l {0}
+        run: |
+          set -ex
+          ./build0.sh
+
+      # ── Step 2: Set up the Emscripten/WASM build environment ──
+      - name: Install WASM build environment
+        shell: bash -l {0}
```
mcbarton: Might worth setting defaults for run to stop repeated use of shell https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrun

CMakeLists.txt
```diff
@@ -329,24 +329,34 @@ if (WITH_LLVM)
     endif()
 endif()
 
+# Helper: configure the Fortran kernelspec.
+# On Emscripten, processes wasm_kernel.json.in; otherwise kernel.json.in.
+# Mirrors xeus-cpp's configure_kernel() pattern.
```
mcbarton: Might be some confusing as to why another projects name gets mentioned in lfortran.

CMakeLists.txt
```diff
@@ -329,24 +329,34 @@ if (WITH_LLVM)
...
+function(configure_lfortran_kernel)
```
mcbarton, line 332: Surely there won't be configuring of kernels for other projects within the cmake? Wouldn't it make more sense to just call it `configure_kernel`

mcbarton (commented):
Another contributor to the xeus-cpp jupyter kernel here. I decided to do a quick review of this PR after I stumbled accross it. Hope you find my review comments helpful. Here are some possible improvements that could be made to the PR based on what xeus-cpp has

- [ ] Add tests for an Emscripten build like https://github.com/compiler-research/xeus-cpp/pull/277 and https://github.com/compiler-research/xeus-cpp/pull/294
- [ ] Add documentation about how to do the Emscripten build
- [ ] Add read the docs preview of jupyter lite site which is build like https://github.com/compiler-research/xeus-cpp/pull/361
- [ ] Add the ability to run the lfortran kernel in jupyter lite such as in https://github.com/compiler-research/xeus-cpp/pull/415/changes

cc @certik

anutosh491:
> Another contributor to the xeus-cpp jupyter kernel here. I decided to do a quick review of this PR after I stumbled accross it. Hope you find my review comments helpful. Here are some possible improvements that could be made to the PR based on what xeus-cpp has

Hi, 

I think this will be the final state in which the PR will be merged. As we did some back & forth on LFortran's Zulip, this is a working version of the PR and we wouldn't like to disturb it. We land something that works first and perfect it one by one.

Any refactoring will be done through atomic commits that follow. Feel free to contribute if you think it might help.

Though let me look at your reviews and comment on the ones that are legit.

CMakeLists.txt
anutosh491: Unlike xeus-cpp, the xeus-lfortan kernel is built inside the LFortran compiler itself.

So the essence of "EMSCRIPEN" extends beyond just having a wasm build for JupyterLite. 

The compiler uses it to generate wasm binaries like this

 ```
 lfortran main.f90 --target=wasm32-unknown-emscripten -o main.js

node main.js
 ```
 
 Read : https://lfortran.org/blog/2024/05/fortran-on-web-using-lfortran/
 
 So for starters it's good to use `XEUS_LFORTRAN_WASM_BUILD`

CMakeLists.txt
anutosh491: Same as : https://github.com/lfortran/lfortran/pull/11556#discussion_r3288394688

wasm-build1.sh
anutosh491: Same as : https://github.com/lfortran/lfortran/pull/11556#discussion_r3288394688

CMakeLists.txt
anutosh491: Yes makes sense. Got rid of it !

CMakeLists.txt
anutosh491: This is fine.

anutosh491:
@Jatinagarwal24 could you take care of these : https://github.com/lfortran/lfortran/pull/11556#discussion_r3287775053
https://github.com/lfortran/lfortran/pull/11556#discussion_r3287776344
https://github.com/lfortran/lfortran/pull/11556#discussion_r3287785643
and maybe this https://github.com/lfortran/lfortran/pull/11556#discussion_r3287818437 

Should be pretty small changes.

.github/workflows/Wasm-Kernel-CI.yml
anutosh491: Ahh fine for this PR.

.github/workflows/deploy-github-page.yml
Jatinagarwal24: Thanks for reporting this.Pushing a fix to remove this.

.github/workflows/Wasm-Kernel-CI.yml
Jatinagarwal24: Thanks for reporting this.Pushing a fix to remove this

wasm-build1.sh
Jatinagarwal24: Thanks for telling this.
I am updating wasm-build1.sh to dynamically detect the correct number of CPU cores based on the operating system instead of using a hardcoded 8. It now uses `nproc` on Linux and `sysctl` on macOS.

.github/workflows/deploy-github-page.yml
Jatinagarwal24: Thanks for telling this. I will implement this by Instead of repeating shell `bash -l {0}` on every single step, I will add a global defaults block at the job level for both `.github/workflows/deploy-github-page.yml `and `.github/workflows/Wasm-Kernel-CI.yml`

.github/workflows/Quick-Checks-CI.yml
Jatinagarwal24: Hi @certik I have fixed this now.
Thanks !

.github/workflows/deploy-github-page.yml
anutosh491: Done.

src/libasr/bwriter.h
```diff
@@ -40,14 +40,14 @@ std::string static inline uint64_to_string(uint64_t i) {
...
-    bytes[6] = (i >>  8) & 0xFF;
-    bytes[7] =  i        & 0xFF;
+    bytes[0] = (((uint64_t)i) >> 56) & 0xFF;
+    bytes[1] = (((uint64_t)i) >> 48) & 0xFF;
+    bytes[2] = (((uint64_t)i) >> 40) & 0xFF;
+    bytes[3] = (((uint64_t)i) >> 32) & 0xFF;
+    bytes[4] = (((uint64_t)i) >> 24) & 0xFF;
+    bytes[5] = (((uint64_t)i) >> 16) & 0xFF;
+    bytes[6] = (((uint64_t)i) >>  8) & 0xFF;
+    bytes[7] =  ((uint64_t)i)        & 0xFF;
```
certik, line 50: I see, in wasm the uintptr_t is 32 bit. I think this change should work.

src/libasr/bwriter.h
```diff
@@ -88,14 +88,15 @@ uintptr_t static inline string_to_uintptr(const char *s) {
...
-                       p[7];
+    return (uintptr_t)(
+           (((uint64_t)p[0]) << 56) |
+           (((uint64_t)p[1]) << 48) |
+           (((uint64_t)p[2]) << 40) |
+           (((uint64_t)p[3]) << 32) |
+           (((uint64_t)p[4]) << 24) |
+           (((uint64_t)p[5]) << 16) |
+           (((uint64_t)p[6]) <<  8) |
+                       p[7]);
```
certik, line 99: If the number doesn't fit into the 32-bit `uintptr_t` (in wasm), then I guess it gets truncated. I think that's fine for now.

.github/workflows/Wasm-Kernel-CI.yml
```diff
@@ -0,0 +1,87 @@
...
+          set -ex
+          ls -la ${{ env.PREFIX }}/bin/xlfortran.js
+          ls -la ${{ env.PREFIX }}/bin/xlfortran.wasm
+
+      # ── Step 5: Test JupyterLite integration ──
+      - name: JupyterLite integration
+        run: |
+          set -ex
+          micromamba create -n xeus-lite-host jupyterlite-xeus -c conda-forge -y
+          micromamba run -n xeus-lite-host jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }}
```
certik, line 87: Can this workflow be merged with our existing workflows, either the Quick Checks or the Exhaustive Checks?

CMakeLists.txt
```diff
@@ -358,6 +365,62 @@ if (WITH_XEUS)
...
+
+    configure_lfortran_kernel()
+
+    set(XJUPYTER_DATA_DIR "share/jupyter" CACHE STRING "Jupyter data directory")
+    set(XEUS_CALCSPEC_DIR ${CMAKE_CURRENT_BINARY_DIR}/share/jupyter/kernels)
+    install(DIRECTORY ${XEUS_CALCSPEC_DIR}
+            DESTINATION ${XJUPYTER_DATA_DIR}
+            PATTERN "*.in" EXCLUDE)
+
+    add_executable(xlfortran
```
certik, line 401: This creates a new binary, is this just the xeus kernel + lfortran evaluator (so no other functionality from the usual lfortran binary)?

And this binary can be build only as a `.wasm` file, or it also can be compiled natively?

Cannot this be merged with our existing Jupyter kernel, which I think is just the usual lfortran binary itself? That simplifies the distribution.

src/lfortran/fortran_evaluator.cpp
certik: I know, but that's what the `execfn` was going, and you define a new `execfn` below. Why cannot it be used? It hides some of the ugly `(*)()` business.

.github/workflows/Quick-Checks-CI.yml
```diff
@@ -350,61 +350,73 @@ jobs:
...
 
       - name: Test built lfortran.wasm
         shell: bash -l {0}
         run: |
             set -ex
-            source $HOME/ext/emsdk/emsdk_env.sh # Activate Emscripten
-            which node
-            node -v
-            node src/lfortran/tests/test_lfortran.js
+            micromamba run -n xeus-lfortran-wasm-build node src/lfortran/tests/test_lfortran.js
```
certik, line 414: We'll have to make sure this works.

.github/workflows/Quick-Checks-CI.yml
```diff
@@ -350,61 +350,73 @@ jobs:
...
+                  -DLFORTRAN_BUILD_ALL=yes \
+                  -DLFORTRAN_BUILD_TO_WASM=yes \
+                  -DWITH_STACKTRACE=no \
+                  -DWITH_LSP=no \
+                  -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH_LFORTRAN;$CONDA_PREFIX" \
+                  -DCMAKE_INSTALL_PREFIX=$(pwd)/inst \
+                  -DCMAKE_EXPORT_COMPILE_COMMANDS=yes \
+                  .
+              cmake --build . -j$(nproc)
+            '
```
certik, line 408: I would try to invoke the `./build_to_wasm.sh` script (possibly modifying it as needed), so that we can easily build it locally with exactly the same configuration.

.github/workflows/Quick-Checks-CI.yml
```diff
@@ -350,61 +350,73 @@ jobs:
...
-          mv emsdk-4.0.13 emsdk
-          export EMSDK_PATH=$HOME/ext/emsdk
-          echo $EMSDK_PATH
-          cd $EMSDK_PATH
-          ./emsdk install 3.1.35
-          ./emsdk activate 3.1.35
-          ./emsdk install node-18.20.3-64bit
-          ./emsdk activate node-18.20.3-64bit
+          micromamba create -f environment-wasm-build.yml -y
+          micromamba install -n xeus-lfortran-wasm-build re2c bison -c conda-forge -y
```
certik, line 359: We have to pin the exact versions here, otherwise it will break in the future.

.github/workflows/Quick-Checks-CI.yml
```diff
@@ -350,61 +350,73 @@ jobs:
           variant: sccache
           key: ${{ github.job }}-${{ matrix.os }}
 
-      - name : Remove existing node
-        shell: bash -e -l {0}
-        run : |
-            which node
-            node -v
-            sudo rm -rf /usr/local/bin/node /usr/local/bin/npm
```
certik, line 358: I think we should still do that, to avoid possible collisions.

.github/workflows/deploy-github-page.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
```
certik, line 100: This we'll have to audit carefully to understand what it is doing, and when to run it.

certik (commented):
Ok, I think most of these are not a blocker. Just the CI we need to figure out.

Is this now tested at the CI, or not? That is the main issue. If it is tested, then we can merge this soon, and then iterate on top.

If this is not tested, then let me at least try to build it locally first to ensure it actually works, and we'll be testing it by hand until we figure out how to test at the CI.

anutosh491:
> Ok, I think most of these are not a blocker. Just the CI we need to figure out.

> If this is not tested, then let me at least try to build it locally first to ensure it actually works, and we'll be testing it by hand until we figure out how to test at the CI.

Yess, we can possibly go ahead and land this and keep improving in subsequent commits. That being said let me reply to some of the above comments which I am sure of.

> Is this now tested at the CI, or not? That is the main issue. If it is tested, then we can merge this soon, and then iterate on top.
> If this is not tested, then let me at least try to build it locally first to ensure it actually works, and we'll be testing it by hand until we figure out how to test at the CI.

Currently there is no testing yet. But we can start with unittests that should work for both the native case and the wasm case.

See what we do for clang-repl : https://github.com/llvm/llvm-project/tree/main/clang/unittests/Interpreter

The simplest file to start with is : https://github.com/llvm/llvm-project/blob/main/clang/unittests/Interpreter/InterpreterTest.cpp

Basically clang provides `clang-repl` the tool & `libclangInterpreter.a`, the library. It exposes `Interpreter.h` and creates  the `ClangReplInterpreterTests` executable linking libclangInterpreter into it. For the emscripten case we just build [ClangReplInterpreterTests.js/wasm](https://github.com/llvm/llvm-project/blob/main/clang/unittests/Interpreter/CMakeLists.txt#L77C23-L92) and test it with node.

We can do something similar. Thinking about it we already had the repl working natively right ! So did we have anything to test it just yet ? If not this should be the way ahead !

We should do this first. This is more fundamental.

After which if we want to test out the JupyterLite notebooks itself in the CI, we need to consider one of the methods mentioned here : https://github.com/scientific-python/summit-2025-nov/issues/13 (I'd prefer the method taken for the official Jupyter website : https://github.com/jupyter/try-jupyter/pull/73)

Can guide @Jatinagarwal24 on the above testing framework discussed.

.github/workflows/deploy-github-page.yml
anutosh491: I think the deploy-github-page.yml is run everytime a commit goes into main !

So if something doesn't work with the wasm kernel but a commit fixing the bug goes in ... we should be able to see it right away.

anutosh491: We can track the fresh builds for the github deployments similar to how we see them in xeus-cpp (https://github.com/compiler-research/xeus-cpp/actions/workflows/deploy-github-page.yml)

.github/workflows/Wasm-Kernel-CI.yml
anutosh491: I guess so. Let's give it a try in the next PR !?

src/lfortran/fortran_evaluator.cpp
anutosh491: Oops, yess we won't be needing that. Didn't realize this earlier. Now it's fixed.

anutosh491:
cc @certik I addressed your doubts regarding using `execfn` and removing the ugly reinterpret_cast operations. Sorry I ended up overlooking that but its fixed now.

CMakeLists.txt
anutosh491: > This creates a new binary, is this just the xeus kernel + lfortran evaluator (so no other functionality from the usual lfortran binary)?

Yes — it's the xeus kernel + the lfortran evaluator pipeline. It links lfortran_lib which contains the full parser → semantics → LLVM codegen stack, but exposes none of the CLI functionality — only the REPL protocol. You're right that most of lfortran_lib's surface area goes unused from xlfortran's perspective.

> And this binary can be build only as a .wasm file, or it also can be compiled natively?

No, and intentionally so. The WASM entry point in fortran_kernel.cpp is [EMSCRIPTEN_BINDINGS](https://github.com/lfortran/lfortran/pull/11556/changes#diff-c660e972f3068c7f7cb8ae9045289748d1d78d1bc42a0be74f7f4e01795544ebR560) — Emscripten's JavaScript interop mechanism. It has no main(). It can't be invoked as a CLI. The native path already does the right thing: WITH_XEUS adds fortran_kernel.cpp to lfortran_lib and the lfortran binary gains a kernel subcommand. So for native, the kernel is the lfortran binary.

> Cannot this be merged with our existing Jupyter kernel, which I think is just the usual lfortran binary itself? That simplifies the distribution.


xeus-lite and xeus-zmq are fundamentally different runtime models:

- Native (WITH_XEUS): xeus-zmq, driven by ZeroMQ sockets, CLI main() dispatches to run_kernel()
- WASM (XEUS_LFORTRAN_WASM_BUILD): xeus-lite, driven by JavaScript postMessage, entry is EMSCRIPTEN_BINDINGS — no main(), no sockets, no command-line parsing 

The WASM kernel is a JS module that JupyterLite's worker thread calls via bindings. You can't graft that onto a CLI binary. Every xeus-lite kernel follows this pattern: xcpp.js, xpython.js, xr.js — all are separate binaries from their native CLI counterparts. xlfortran.js follows the same convention (I named it so just to follow how other kernels have been named)

anutosh491: So once this PR goes and we have a release ... on emscripten-forge we currently build lfortran.js/wasm (and test_lfortran.js/wasm and run the [test using node](https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/lfortran/recipe.yaml#L39))

But I guess the main usecase for hosting lfortran on emscripten-forge would be to fetch it inside a Jupyterlite env.

So I shall update the recipe to build xlfortran.js/wasm (and if we want we can also build lfortran.js/wasm if people want to use the wasm build for the compiler itself)

anutosh491:
I think this should be ready now. @Jatinagarwal24 would might like to go through some of the reviews from above and see if somethings can be addressed.

Possibly these ones 

https://github.com/lfortran/lfortran/pull/11556#discussion_r3307242622
https://github.com/lfortran/lfortran/pull/11556#discussion_r3307238383
https://github.com/lfortran/lfortran/pull/11556#discussion_r3307244549

CMakeLists.txt
certik: I see, let's keep the binary separate for now then.

Jatinagarwal24:
> I think this should be ready now. @Jatinagarwal24 would might like to go through some of the reviews from above and see if somethings can be addressed.
> 
> Possibly these ones
> 
> [#11556 (comment)](https://github.com/lfortran/lfortran/pull/11556#discussion_r3307242622) [#11556 (comment)](https://github.com/lfortran/lfortran/pull/11556#discussion_r3307238383) [#11556 (comment)](https://github.com/lfortran/lfortran/pull/11556#discussion_r3307244549)

Hi @certik @anutosh491 Should I resolve these comments in this pr or subsequent pr ?I think it will be better to wrap this PR.

anutosh491:
> Should I resolve these comments in this pr or subsequent pr ?

Hmm we are 25 commits in ... so maybe let's save refactoring work for upcoming Prs ? What say !?

My aim now is to to confirm that once this is merged, the github deployment works or no !

Jatinagarwal24:
> > Should I resolve these comments in this pr or subsequent pr ?
> 
> Hmm we are 25 commits in ... so maybe let's save refactoring work for upcoming Prs ? What say !?
> 
> My aim now is to to confirm that once this is merged, the github deployment works or no !

Yes if that dosen't then we can fix it in new pr.

certik:
@mcbarton thank you so much for the review!

certik:
@anutosh491 please don't merge PRs without a review, especially big ones like this one.

.github/workflows/Quick-Checks-CI.yml
certik: This still applies

.github/workflows/Quick-Checks-CI.yml
certik: This still applies

.github/workflows/Quick-Checks-CI.yml
certik: This applies

anutosh491:
> please don't merge PRs without a review, especially big ones like this one.

Ahh I'm sorry.

I was actually waiting to hear from everyone if there's any more reviews for the past couple days. Was tempted to check if the github deployment page works hence thought of merging.

Do you think anything more should be done through this PR ?

certik (approved):
I thing it's fine, if others can also review it. The reported bugs are relativly                                                                              minor and can be fixed by subsequent PRs.

anutosh491:
Thank you !

Shall be merging it and checking if the deployment works fine.

anutosh491:
Hey @certik @Jatinagarwal24 

The Build passes but the [Deployment fails](https://github.com/lfortran/lfortran/actions/runs/26677970778/job/78633925058) cause we don't have Github Pages Enabled

image



I don't think I have permissions to change settings of the repo. 

Could one of y'all maybe
- Open Settings → Pages
- Under Build and deployment → Source, select GitHub Actions
- Save, then rerun the workflow.

Thanks.

Jatinagarwal24:
> Hey @certik @Jatinagarwal24
> 
> The Build passes but the [Deployment fails](https://github.com/lfortran/lfortran/actions/runs/26677970778/job/78633925058) cause we don't have Github Pages Enabled
> 
> image
> I don't think I have permissions to change settings of the repo.
> 
> Could one of y'all maybe
> 
> * Open Settings → Pages
> * Under Build and deployment → Source, select GitHub Actions
> * Save, then rerun the workflow.
> 
> Thanks.

Hi @anutosh491 I think only @certik can do this.
Maybe you can ping on zulip for this.

certik:
I removed the wasm workflow, it has at least three blocker issues in it: https://github.com/lfortran/lfortran/pull/11691. 

@anutosh491 go ahead and submit a PR with a proper fix.

PR #11558, opened 2026-05-18
fix: function reference incorrectly treated as array reference

### LINKED ISSUE #11541: function reference incorrectly treated as array reference
## MRE

```fortran id="mre8421"
program counter
  implicit none
  integer get_count
  print *, get_count()
end program counter

integer function get_count()
  integer cnt
  save
  data cnt /0/
  cnt = cnt + 1
  get_count = cnt
end function get_count
```

## Output

```text id="out8421"
(base) opixdown@Atharvs-MacBook-Air lfortran % lfortran a.f90

semantic error: Array reference is not allowed on scalar variable

 --> a.f90:6:12
  |
6 |   print *, get_count()
  |            ^^^^^^^^^^^
```

## gfortran Behavior

```text id="out8422"
(base) opixdown@Atharvs-MacBook-Air lfortran % gfortran a.f90 && ./a.out
           1
```

opixdown:
fixes #11541

### DIFF
--- a/integration_tests/functions_62.f90
+++ b/integration_tests/functions_62.f90
@@ -1,0 +1,14 @@
+program functions_62
+  implicit none
+  integer get_count
+  if(get_count() /= 1) stop "get_count() should return 1 "
+  print *, "test passed"
+end program functions_62
+
+integer function get_count()
+  integer cnt
+  save
+  data cnt /0/
+  cnt = cnt + 1
+  get_count = cnt
+end function get_count
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -694,6 +694,7 @@ RUN(NAME functions_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_59 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_61 LABELS gfortran llvm fortran)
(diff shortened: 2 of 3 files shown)


integration_tests/functions_62.f90
```diff
@@ -0,0 +1,14 @@
...
+  print *, "test passed"
+end program functions_62
+
+integer function get_count()
+  integer cnt
+  save
+  data cnt /0/
+  cnt = cnt + 1
+  get_count = cnt
+end function get_count
```
certik, line 14: I think this could must give an error message.

And you must use `--implicit-interface` to enable.

opixdown: @harperjf commented on issue (https://github.com/lfortran/lfortran/issues/11541#issuecomment-4468745955)
"The program seems to be standard-conforming so I think it should compile by default."

certik: Yes, this already works if you enable standard conformance:
```
$ lfortran a.f90 --std=f23
 test passed
```
It's enough to enable implicit interfaces:
```
$ lfortran a.f90 --implicit-interface
test passed
```
But it should not work by default:
```
$ lfortran a.f90
semantic error: Array reference is not allowed on scalar variable
 --> a.f90:4:6
  |
4 |   if(get_count() /= 1) stop "get_count() should return 1 "
  |      ^^^^^^^^^^^
```

We could possibly improve the error to mention `--implicit-interface` if it can be done with little code, but that might be hard.

Implicit interfaces is a feature that we should not enable by default, as the compiler cannot check the arguments, it must heuristically guess what the arguments are from the usage and create an interface implicitly, so it can easily segfault.

opixdown: opened #11576
for improving error message

PR #11492, opened 2026-05-13
fix: read list-directed stdin values token by token

Jatinagarwal24:
fixes #11491
Update runtime stdin reads for scalar integer, logical, real, and double
values to consume only the next list-directed token instead of a whole
line. This allows reads like `read (*,*) ok, year, month, day, hour` to
consume multiple values from the same input line correctly.

Add an integration test covering mixed logical, integer, and real input
from redirected stdin.

PR #11567, opened 2026-05-19
test: Add reference tests for list-directed read EOF errors

Jatinagarwal24:
resolves all the comments of pr #11492

Added tests to verify the runtime behavior when reading from standard
input hits EOF, specifically verifying the `iostat` supplied vs 
not supplied conditions.

- Added `read_error_01.f90` to test integer read (crashes with "Error: Failed to read input.")
- Added `read_error_02.f90` to test double read (crashes with "Error: Failed to read double from stdin.")
- Added `read_iostat_01.f90` to test double read with `iostat` supplied (returns gracefully with iostat=-1)
- Created `read_error_01.in` to pipe immediate EOF into standard input using `<` in `tests.toml`.
- Updated reference outputs for all three scenarios.


Jatinagarwal24:
Hi @certik I dont think we need exhaustive tests here.

certik (approved):

PR #11569, opened 2026-05-19
 Skip runtime shape-conformance checks for assumed-shape dummy assignments

### LINKED ISSUE #11563: Athena: Incorrect shape handling for intent(out) array assignment in type-bound call
```fortran
module m_mre2f
  use, intrinsic :: iso_fortran_env, only: real32
  implicit none

  type :: array_type
    integer :: rank = 0
    integer, dimension(:), allocatable :: shape
    real(real32), dimension(:,:), allocatable :: val
    type(array_type), pointer :: grad => null()
    type(array_type), pointer :: left_operand => null()
    type(array_type), pointer :: right_operand => null()
  contains
    procedure, pass(this) :: allocate => allocate_array
    procedure, pass(this) :: get_partial_left_val
  end type array_type

contains

  subroutine allocate_array(this, array_shape)
    class(array_type), intent(inout), target :: this
    integer, dimension(:), intent(in) :: array_shape
    integer :: n
    n = size(array_shape)
    if (n == 1) then
      allocate(this%val(1, array_shape(1)))
      allocate(this%shape(0))
    else
      allocate(this%val(product(array_shape(1:n-1)), array_shape(n)))
      this%shape = array_shape(1:n-1)
    end if
  end subroutine allocate_array

  subroutine get_partial_left_val(this, upstream_grad, output)
    class(array_type), intent(in) :: this
    real(real32), dimension(:,:), intent(in) :: upstream_grad
    real(real32), dimension(:,:), intent(out) :: output
    output = upstream_grad * 2.0_real32
  end subroutine get_partial_left_val

  recursive subroutine reverse_mode(array, upstream_grad, depth)
    class(array_type), intent(inout) :: array
    real(real32), dimension(:,:), intent(in) :: upstream_grad
    integer, intent(in) :: depth
    integer :: num_samples

    if (associated(array%left_operand)) then
      num_samples = max(size(array%left_operand%val, 2), size(upstream_grad, 2))
      call accumulate_gradient_samples(array%left_operand, array, &
           upstream_grad, num_samples, .true., depth+1)
    end if
  end subroutine reverse_mode

  recursive subroutine accumulate_gradient_samples( &
       array, parent, upstream_grad, num_samples, is_left_operand, depth )
    class(array_type), intent(inout) :: array
    class(array_type), intent(inout) :: parent
    real(real32), dimension(:,:), intent(in) :: upstream_grad
    integer, intent(in) :: num_samples
    logical, intent(in) :: is_left_operand
    integer, intent(in) :: depth
    real(real32), dimension(size(array%val, 1), num_samples) :: grad
    logical :: needs_recurse

    if(depth > 5) return

    needs_recurse = associated(array%left_operand) .or. &
         associated(array%right_operand)

    if(.not. associated(array%grad))then
       allocate(array%grad)
       call array%grad%allocate(array_shape=[array%shape, num_samples])
       if(is_left_operand)then
          call parent%get_partial_left_val(upstream_grad, array%grad%val)
       end if
       if(needs_recurse) call reverse_mode(array, array%grad%val, depth+1)
       return
    end if
  end subroutine accumulate_gradient_samples

end module m_mre2f


program mre2f
  use, intrinsic :: iso_fortran_env, only: real32
  use m_mre2f
  implicit none

  type(array_type), target :: a, b
  real(real32) :: ug(8, 1)

  call a%allocate(array_shape=[8, 1])
  call b%allocate(array_shape=[1])
  a%left_operand => b

  ug = 1.0_real32
  call reverse_mode(a, ug, 0)

  if (.not. associated(b%grad)) then
     print *, "FAIL: b%grad not associated"
     error stop 1
  end if
  if (size(b%grad%val, 1) /= 1) then
     print *, "FAIL: b%grad%val dim1 = ", size(b%grad%val, 1), " expected 1"
     error stop 2
  end if

  print *, "PASS"

end program mre2f
```

```console
lfortran /tmp/mres/mre2f.f90 --realloc-lhs-arrays
runtime error: Array shape mismatch in assignment to 'output'. Tried to match size 1 of dimension 1 of LHS with size 8 of dimension 1 of RHS.
  --> /tmp/mres/mre2f.f90:37:5
   |
37 |     output = upstream_grad * 2.0_real32
   |     ^^^^^^ LHS size is 1
   |
37 |     output = upstream_grad * 2.0_real32
   |              ^^^^^^^^^^^^^ RHS size is 8
% gfortran /tmp/mres/mre2f.f90 && ./a.out          
 PASS
```

HarshitaKalani:
fixes: https://github.com/lfortran/lfortran/issues/11563

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4443,6 +4443,7 @@ RUN(NAME intent_out_struct_member_no_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_module_dealloc LABELS gfortran llvm)
 RUN(NAME intent_out_nested_allocatable_01 LABELS gfortran llvm)
 RUN(NAME intent_out_array_structs_01 LABELS gfortran llvm)
+RUN(NAME intent_out_assumed_shape_dummy_01 LABELS gfortran llvm)
 RUN(NAME array_shape_func_call LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_neg_extent_01 LABELS gfortran llvm)
 
--- a/integration_tests/intent_out_assumed_shape_dummy_01.f90
+++ b/integration_tests/intent_out_assumed_shape_dummy_01.f90
@@ -1,0 +1,51 @@
+! Issue #11563: lfortran should not raise a runtime shape-mismatch
+! error when the LHS of an assignment is a non-allocatable assumed-shape
+! intent(out) dummy argument. The shape is fixed by the caller and a
+! shape mismatch there is undefined behavior per the Fortran standard,
+! but gfortran's default (without -fcheck=bounds) silently accepts it.
+! This test only verifies that the program reaches the end without
+! aborting; it does not inspect the contents of the dummy after the
+! mismatched assignment (which would be UB).
+module m_intent_out_assumed_shape_dummy_01
+    use, intrinsic :: iso_fortran_env, only: real32
+    implicit none
+
+    type :: array_type
+        real(real32), dimension(:,:), allocatable :: val
+    contains
+        procedure, pass(this) :: write_partial
+    end type array_type
+
+contains
+
+    subroutine write_partial(this, upstream_grad, output)
+        class(array_type), intent(in) :: this
+        real(real32), dimension(:,:), intent(in) :: upstream_grad
+        real(real32), dimension(:,:), intent(out) :: output
+        output = upstream_grad * 2.0_real32
+    end subroutine write_partial
+
+end module m_intent_out_assumed_shape_dummy_01
+
+
+program intent_out_assumed_shape_dummy_01
+    use, intrinsic :: iso_fortran_env, only: real32
+    use m_intent_out_assumed_shape_dummy_01
+    implicit none
+
+    type(array_type) :: a
+    real(real32) :: ug(8, 1)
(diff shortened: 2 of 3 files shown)


certik:
I think the code is invalid Fortran and the arrays are not the same size, so it should fail when assigning. LFortran correctly fails:
```
$ lfortran a.f90
runtime error: Array shape mismatch in assignment to 'output'. Tried to match size 1 of dimension 1 of LHS with size 8 of dimension 1 of RHS.
  --> a.f90:17:9
   |
17 |         output = upstream_grad * 2.0_real32
   |         ^^^^^^ LHS size is 1
   |
17 |         output = upstream_grad * 2.0_real32
   |                  ^^^^^^^^^^^^^ RHS size is 8
```
So everything behaves correctly.

This PR removes a good check. However, the fact that tests pass even with removing this check means it is not tested in this case. So please add a test for this runtime error message. But do not remove it.

Unless I misunderstand something.

HarshitaKalani:
Yes, this is a bug in athena. GFortran without -fcheck=bounds silently performs an out-of-bounds write which makes this test pass. I'll workaround it on the Athena side. Opened a PR for the test here: https://github.com/lfortran/lfortran/pull/11784

HarshitaKalani:
Closing this: https://github.com/lfortran/lfortran/pull/11784

Issue #11494, opened 2026-05-13
Performance : `ReplaceFunctionParamWithArg` at `asr_utils.h`


assem2002:
### Problem 
`ReplaceFunctionParamWithArg` class is being used through its `replace_FunctionParam_with_arg` that attempts to duplicate expressions even if the replacement is not gonna take place.
Refer to discussion here https://github.com/lfortran/lfortran/pull/11446#discussion_r3219375612
### Solution
Find a way to only duplicate when a hit functionParam (also without doing 2 tree traversals -- one for checking if funcParam exists , and another for replacement)

PR #11571, opened 2026-05-19
perf: optimize ReplaceFunctionParamWithArg and fix upstream memory scopes

srinjoy933:
hey @assem2002 
I have thoroughly gone through the  Issue #11494  which suggested avoiding the unconditional duplication of expressions inside ReplaceFunctionParamWithArg to improve performance.

To resolve this, I initially implemented a fast read-only visitor pass to check the tree first, combined with a single-pass Copy-on-Write mechanism. The goal was to only allocate new memory if a FunctionParam was actually present in the expression tree. While this successfully reduced memory allocations, it caused the ASR verify pass to fail on tests like arrays_47.f90 with the error cannot point outside of its symbol table.

Further debugging revealed that skipping this duplication is fundamentally incompatible with LFortran's memory architecture. When the compiler creates new parent nodes—such as array struct temporaries—it must assign them their own unique, deep-copied bounds expressions. If we skip the duplication step to save performance, the new struct temporary and the original variable end up pointing to the exact same expression node in memory. This transforms the AST from a strict tree into a Directed Acyclic Graph (DAG), which violates the compiler's scope-tracking invariants and triggers the verifier failures.

Because unconditional duplication is an architectural requirement to maintain safe memory separation between scopes, the specific optimization requested in Issue #11494 cannot be safely implemented.

However, the fast-pass visitor built during this investigation is highly valuable. Rather than discarding the work, I have extracted it into a standalone, zero-allocation utility called expr_has_function_param(). This read-only visitor pass can be reused across the compiler for tasks like purity and bounds checking without incurring any memory overhead.


assem2002:
It's good utility. Thanks for that.
but we wanted to eliminate the unnecessary duplication we do. Using a checker is actually a better way compared to duplicating (visits the tree) + walking down the tree again searching for functionParam.

but we can make it better.

Here is what I have in mind :-
- `replace_FunctionParam_with_arg` is called when we want to replace a `functionParam` (if it exists) -- the function happens to duplicate right away even if there's no `functionParam`

### Fix
- Preserve expression that gets passed to `replace_FunctionParam_with_arg` ( the main original one )
- If you hit a `functionParam`, Do the replacement on the original expression + preserve the functionParam 
- Now duplicate the original expression
- revert the original expression to have back its `functionParam`
- Return to the user of the class -- the duplicate expression.
- If you'll use some class members to hold some information, maybe use `Singleton` pattern on the class to avoid wrong usage of the class by managing the class state.

### Outcome
- We don't duplicate if not functionParam exists (no unnecessary memory usage from our arena, no unnecessary visiting when duplicating)

src/libasr/asr_utils.h
```diff
@@ -3385,17 +3385,33 @@ static inline bool expr_references_symbol(ASR::expr_t* expr, ASR::symbol_t* sym)
     return visitor.found;
 }
 
+class HasFunctionParamVisitor : public ASR::BaseWalkVisitor<HasFunctionParamVisitor> {
```
certik: This class is never used, correct? So it can be removed.

srinjoy933: Hi @certik, I'm actually using this class in the latest commits I just pushed! It's now a part of the updated logic.

src/libasr/asr_utils.h
```diff
@@ -3405,22 +3420,37 @@ class ReplaceFunctionParamWithArg: public ASR::BaseExprReplacer<ReplaceFunctionP
...
+            
+            replacements.push_back({current_expr, *current_expr});
             *current_expr = m_args[n].m_value;
         }
     }
 
     ASR::expr_t* replace_FunctionParam_with_arg(ASR::expr_t* t) {
+        if (!t) return nullptr;
+
+        if (!expr_has_function_param(t)) {
```
assem2002: I think we don't need this -- let's depend on `replacements.empty()` to check if we need to return without duplication

src/libasr/asr_utils.h
```diff
@@ -3384,18 +3386,31 @@ static inline bool expr_references_symbol(ASR::expr_t* expr, ASR::symbol_t* sym)
...
+class HasFunctionParamVisitor : public ASR::BaseWalkVisitor<HasFunctionParamVisitor> {
+public:
+    bool has_param = false;
+    
+    void visit_FunctionParam(const ASR::FunctionParam_t & /*x*/) {
+        has_param = true;
+    }
+};
 
+inline bool expr_has_function_param(ASR::expr_t *t) {
```
assem2002: I think we an remove this

src/libasr/asr_utils.h
```diff
@@ -3384,18 +3386,31 @@ static inline bool expr_references_symbol(ASR::expr_t* expr, ASR::symbol_t* sym)
     visitor.visit_expr(*expr);
     return visitor.found;
 }
+class HasFunctionParamVisitor : public ASR::BaseWalkVisitor<HasFunctionParamVisitor> {
```
assem2002: We can remove this also.

src/libasr/asr_utils.h
```diff
@@ -3405,22 +3420,37 @@ class ReplaceFunctionParamWithArg: public ASR::BaseExprReplacer<ReplaceFunctionP
             if (n >= n_args) {
                 LCOMPILERS_ASSERT("FunctionParam param number not in range.");
             };
+
```
assem2002:
```suggestion
LCOMPILERS_ASSERT_MSG(n < n_args,"FunctionParam param number not in range.");

            
```

src/libasr/asr_utils.h
```diff
@@ -3405,22 +3420,37 @@ class ReplaceFunctionParamWithArg: public ASR::BaseExprReplacer<ReplaceFunctionP
...
     }
 
     ASR::expr_t* replace_FunctionParam_with_arg(ASR::expr_t* t) {
+        if (!t) return nullptr;
+
+        if (!expr_has_function_param(t)) {
+            return t;
+        }
+
+        replacements.clear();
```
assem2002:
```suggestion
                LCOMPILERS_ASSERT(replacements.empty()) 

```

assem2002: This is better. I think we shouldn't expect `replacement` to be non-empty at thing point (we're clearning at the end)

src/libasr/asr_utils.h
```diff
@@ -8013,32 +8043,45 @@ static inline void Call_t_body(Allocator& al, ASR::symbol_t* a_name,
...
+                      for (size_t i = 0; i < dimension_.size(); i++) {
+                          dimension_.p[i].loc = arg->base.loc;
+        
+                          // 2. Safely extract the replaced memory (optimized or mutated)
+                          ASR::expr_t* replaced_length = r.replace_FunctionParam_with_arg(dimension_[i].m_length);
+                          ASR::expr_t* replaced_start = r.replace_FunctionParam_with_arg(dimension_[i].m_start);
+        
+                          // 3. Caller takes responsibility for duplication to prevent the DAG crash
+                          dimension_.p[i].m_length = replaced_length ? caller_dup.duplicate_expr(replaced_length) : nullptr;
+                          dimension_.p[i].m_start = replaced_start ? caller_dup.duplicate_expr(replaced_start) : nullptr;
```
assem2002:
```suggestion
                          // 3. Caller takes responsibility for duplication to prevent the DAG crash
                          dimension_.p[i].m_length = (replaced_length == dimension_.p[i].m_length) ? caller_dup.duplicate_expr(replaced_length) : replaced_length;
                          dimension_.p[i].m_start = (replaced_start == dimension_.p[i].m_start) ? caller_dup.duplicate_expr(replaced_start) : replaced_start;
```

assem2002: don't duplicate if we already duplicated it.

src/libasr/asr_utils.cpp
```diff
@@ -2746,17 +2770,29 @@ bool use_overloaded(ASR::expr_t* left, ASR::expr_t* right,
...
+                                        // 1. Initialize the caller's duplicator OUTSIDE the loop
+                                        ASRUtils::ExprStmtDuplicator caller_dup(al);
+                                        caller_dup.allow_procedure_calls = true;
+                                        caller_dup.success = true;
+
+                                     for (size_t di = 0; di < n_ret_dims; di++) {
+                                         if (ret_dims[di].m_length) {
+                                            // 2. Extract and safely duplicate m_length
+                                            ASR::expr_t* replaced_length = r.replace_FunctionParam_with_arg(ret_dims[di].m_length);
+                                            ret_dims[di].m_length = replaced_length ? caller_dup.duplicate_expr(replaced_length) : nullptr;
```
assem2002:
```suggestion
                                            ret_dims[di].m_length = (replaced_length == ret_dims[di].m_length) ? caller_dup.duplicate_expr(replaced_length) : replaced_length;
```

src/libasr/asr_utils.cpp
```diff
@@ -2746,17 +2770,29 @@ bool use_overloaded(ASR::expr_t* left, ASR::expr_t* right,
...
+                                         }
+        
+                                         if (ret_dims[di].m_start) {
+                                            // 3. Extract and safely duplicate m_start
+                                            ASR::expr_t* replaced_start = r.replace_FunctionParam_with_arg(ret_dims[di].m_start);
+                                            ret_dims[di].m_start = replaced_start ? caller_dup.duplicate_expr(replaced_start) : nullptr;
```
assem2002: same

src/libasr/asr_utils.cpp
```diff
@@ -1861,13 +1873,25 @@ void process_overloaded_unary_minus_function(ASR::symbol_t* proc, ASR::expr_t* o
...
+                       // 1. Initialize the caller's duplicator OUTSIDE the loop
+                       ASRUtils::ExprStmtDuplicator caller_dup(al);
+                       caller_dup.allow_procedure_calls = true;
+                       caller_dup.success = true;
+
+                       for (size_t di = 0; di < n_ret_dims; di++) {
+                           if (ret_dims[di].m_length) {
+                              // 2. Extract and safely duplicate m_length
+                              ASR::expr_t* replaced_length = r.replace_FunctionParam_with_arg(ret_dims[di].m_length);
+                              ret_dims[di].m_length = replaced_length ? caller_dup.duplicate_expr(replaced_length) : nullptr;
```
assem2002: same

src/libasr/asr_utils.cpp
```diff
@@ -1760,14 +1760,26 @@ bool use_overloaded(ASR::expr_t* left, ASR::expr_t* right,
...
+                                       // 1. Initialize the caller's duplicator OUTSIDE the loop
+                                       ASRUtils::ExprStmtDuplicator caller_dup(al);
+                                       caller_dup.allow_procedure_calls = true;
+                                       caller_dup.success = true;
+
+                                       for (size_t di = 0; di < n_ret_dims; di++) {
+                                          if (ret_dims[di].m_length) {
+                                          // 2. Extract and safely duplicate m_length
+                                          ASR::expr_t* replaced_length = r.replace_FunctionParam_with_arg(ret_dims[di].m_length);
+                                          ret_dims[di].m_length = replaced_length ? caller_dup.duplicate_expr(replaced_length) : nullptr;
```
assem2002: same

certik:
> This transforms the AST from a strict tree into a Directed Acyclic Graph (DAG), which violates the compiler's scope-tracking invariants and triggers the verifier failures.

Can you explain what this sentence means? We don't use AST it this PR, but ASR. And ASR is indeed a DAG, due to the symbol table nodes. So I don't understand your sentence.

certik:
> Rather than discarding the work, I have extracted it into a standalone, zero-allocation utility called expr_has_function_param(). 

I can't find it in this PR. I am guessing it was removed? If so, go ahead and update the description.

certik:
@assem2002 can you please review once more? I don't quite understand all the details, but I think you do.

srinjoy933:
Hi @certik , I tried two or three different approaches for this issue ,after suggesting with @assem2002 .I  also implemented them  but those approaches failed the tests and had some architectural errors.so I discarded them and implemented the current approach and made the changes according to it.The comments you have mentioned are of those discarded approaches and changes .it has nothing to do with the current pr.I think those comments are making confusion so I am having only the comment about the current changes and deleting the previous ones.

assem2002 (approved):
LGTM.

PR #11577, opened 2026-05-19
Fix: ICE in template instantiation when intrinsic function is inapplicable

### LINKED ISSUE #11575: internal compiler error: unhandled exception, when instantiating min with complex, complex
I know min on complex is illegal, but I want an error message saying that instead of giving me an internal compiler error: unhandled exception.

code:

```
module Requirement_ordering_m
    implicit none

    requirement op_R ( T, V, op_func )
        type, deferred :: T
        type, deferred :: V
            type(T), intent(in) :: lhs
            type(T), intent(in) :: rhs
            type(V) :: res
        end function
    end requirement

    template op_t ( T, V, op_func )
      require :: op_R (T, V, op_func )
    contains
      pure elemental function call_op_func ( x ) result(res)
        type(V) :: res
        type(T) :: x
        res = op_func(x,x)
      end function
    end template

end module

program p
  use Requirement_ordering_m

 instantiate op_t ( complex, complex, min ), only: com_min    => call_op_func
   print *, " com_min   ((0,1))=", com_min   ((0,1))
```

Jatinagarwal24:
fixes #11575
The `create_func` for intrinsic elemental functions (e.g., `min`) can
return `nullptr` when the argument types are unsupported (e.g., complex).
Previously, the result was passed directly to `ASRUtils::EXPR()`, which
asserts non-null, causing an internal compiler error. Now we check for
`nullptr` and throw `SemanticAbort()` so the diagnostic is reported
cleanly.

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -300,7 +300,7 @@ program continue_compilation_1
     class(Derived), allocatable :: derived_cls
     integer, parameter :: z(1) = 2
     integer, parameter :: qval(2) = reshape([7, 8], -[z])
-
+    print *, min(c, c)
 
 
 
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -300,7 +300,7 @@ program continue_compilation_1
     class(Derived), allocatable :: derived_cls
     integer, parameter :: z(1) = 2
     integer, parameter :: qval(2) = reshape([7, 8], -[z])
-    print *, min(c, c)
+
 
 
 
@@ -509,7 +509,7 @@ program continue_compilation_1
     print *, present()
     print *, ieor(x)
     print *, ieor()
-
+    print *, min(c, c)
     exit
 
     ! calling function with less arguments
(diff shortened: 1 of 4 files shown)


certik (approved):
I think this is fine. This has nothing to do with templates, correct?

tests/errors/continue_compilation_1.f90
```diff
@@ -300,7 +300,7 @@ program continue_compilation_1
     class(Derived), allocatable :: derived_cls
     integer, parameter :: z(1) = 2
     integer, parameter :: qval(2) = reshape([7, 8], -[z])
-
+    print *, min(c, c)
```
certik: This must go below, not here. See the comment below.

Jatinagarwal24: Yes i will move it down.

Jatinagarwal24:
> I think this is fine. This has nothing to do with templates, correct?

Actually, it does have to do with templates.
This fix is specifically in the template instantiation code path (visit_Instantiate in ast_symboltable_visitor.cpp). The ICE occurs when an intrinsic function like min is passed as a template argument and create_func returns nullptr because the argument types are unsupported (e.g., complex for min). The nullptr was then passed directly to ASRUtils::EXPR() which asserts non-null, causing the crash.

certik (approved):

PR #11587, opened 2026-05-20
fix: bound index() outer loop by len(string) - len(substring) + 1

HarshitaKalani:


jinangshah21 (approved):

assem2002:
Please resolve conflicts

PR #11588, opened 2026-05-20
fix: skip entry FINAL for converted function result variables

HarshitaKalani:


certik:
Looks like it is failing the finalization tests.

src/libasr/asr_verify.cpp
```diff
@@ -810,15 +810,19 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
         verify_unique_dependencies(x.m_dependencies, x.n_dependencies,
                                    x.m_name, x.base.base.loc);
 
-        // Verify dependencies
+        const std::string marker_prefix = "__lcompilers_marker_";
         for( size_t i = 0; i < x.n_dependencies; i++ ) {
+            std::string dep = std::string(x.m_dependencies[i]);
+            if (dep.rfind(marker_prefix, 0) == 0) {
+                continue;
+            }
```
certik: Can you explain what this is doing? That doesn't seem right to special case some strings in ASR verify.

src/libasr/pass/intent_out_deallocate.cpp
```diff
@@ -175,6 +176,41 @@ class LHSFinalizationVisitor : public PassUtils::PassVisitor<LHSFinalizationVisi
...
+        }
+        if (st->m_parent != nullptr) {
+            st = ASR::down_cast<ASR::Struct_t>(
+                ASRUtils::symbol_get_past_external(st->m_parent));
+        } else {
+            st = nullptr;
+        }
+    }
+    return false;
+}
```
certik: @assem2002 does this fit with the plan to refactor insert deallocate?

certik:
My comment still applies: https://github.com/lfortran/lfortran/pull/11588#discussion_r3289016679.

src/libasr/pass/intent_out_deallocate.cpp
assem2002: As we discussed before `intent_out` related issues is fine to fix here.
I still don't know what the issue to be fixed here.

HarshitaKalani: The issue is that after subroutine_from_function converts a function result into a trailing intent(out) argument, intent_out_deallocate mistakenly treats it like a normal output argument and emits entry-finalization code for it. At that point the result object has not been initialized yet, so we end up finalizing an uninitialized object before the function body executes (finalization_08.f90).

The fix is to run entry-finalization before subroutine_from_function, while the function result is still recognized as m_return_var, which naturally excludes it from this logic.

src/libasr/asr_verify.cpp
HarshitaKalani: You're right. I have now fixed this by moving entry-finalization into pass_intent_out_finalize_on_entry that runs before subroutine_from_function. At that point the function result is still in m_return_var, so it is skipped and no special-casing is needed.

src/libasr/pass/intent_out_deallocate.cpp
```diff
@@ -415,10 +430,19 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
+
+    PassUtils::UpdateDependenciesVisitor u(al);
+    u.visit_TranslationUnit(unit);
+}
+
 void pass_intent_out_deallocate(Allocator &al, ASR::TranslationUnit_t &unit,
                                 const PassOptions &/*pass_options*/) {
-    // Deallocate intent(out) allocatable arguments at function entry
-    IntentOutDeallocateVisitor iod(al);
+    IntentOutDeallocateVisitor iod(al, /*emit_final_only=*/false);
```
certik: This seems really convoluted. So now we have two final visitors, one is `emit_final_only = true`, the other one is false, otherwise they are identical? And we always run them together one after the other? 

Why do we need two visitors like this, why isn't one enough?

assem2002:
currently we're setting reutrn-var that are passed as an argument (return slot) - as `INTENT(OUT)`.
I think we can just avoid setting them as `intent(out)`.
@HarshitaKalani What do you think?

HarshitaKalani:
Closing as not needed anymore.

PR #11590, opened 2026-05-20
fix: reject BOZ literals as substring indices

### LINKED ISSUE #11583: missing semantic error and bounds check for BOZ literal substring index
## MRE

```fortran id="mre48614"
program mre
  implicit none

  character(len=23) :: string

  string = ''
  print *, string(1:Z'100000003')

end program mre
```

## Output

```text id="out48614"

(compiles successfully)
```

## gfortran Behavior

```text id="out48615"

Error: Substring end index at (1) must be of type INTEGER
```

## Expected

LFortran should reject BOZ literals as substring indices.

Additionally, if such expressions are accepted, the compiler should still emit an error because the substring upper bound exceeds the character length.

Jatinagarwal24:
fixes #11583 

BOZ literals were lowered as integer constants, so LFortran accepted them
as substring indices. Fixed substring bound validation to reject BOZ
literals and report out-of-range upper bounds for fixed-length strings.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -10140,11 +10140,6 @@ public:
                     ASRUtils::type_get_past_allocatable_pointer(root_v_type));
                 int64_t str_length;
                 if (ASRUtils::extract_value(s_type->m_len, str_length)) {
-                    if (start <= 0) {
-                        diag.add(Diagnostic("Substring `start` is less than one",
-                            Level::Error, Stage::Semantic, {Label("", {loc})}));
-                        throw SemanticAbort();
-                    }
                     if (end > str_length) {
                         diag.add(Diagnostic("Substring end index exceeds the string length",
                             Level::Error, Stage::Semantic, {Label("", {loc})}));
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -9932,6 +9932,14 @@ public:
         return 1; // default
     }
 
+    bool is_boz_integer_constant(ASR::expr_t* expr) {
+        if (!expr || !ASR::is_a<ASR::IntegerConstant_t>(*expr)) {
+            return false;
+        }
+        ASR::IntegerConstant_t* int_const = ASR::down_cast<ASR::IntegerConstant_t>(expr);
+        return int_const->m_intboz_type != ASR::integerbozType::Decimal;
+    }
+
     ASR::asr_t* create_ArrayRef(const Location &loc, AST::fnarg_t* m_args,
         size_t n_args, AST::fnarg_t* m_subargs, size_t n_subargs,
         ASR::expr_t* v_expr, ASR::symbol_t *v, ASR::symbol_t *f2) {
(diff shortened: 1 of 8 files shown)


HarshitaKalani (approved):

tests/reference/asr-continue_compilation_3-435a232.stderr
```diff
@@ -247,7 +247,7 @@ semantic error: Binary numeric operators cannot be used on strings
 144 |     print *, "a" + "b"
     |              ^^^^^^^^^ help: use '//' for string concatenation
 
-semantic error: The first index in string section is less than 1
+semantic error: Substring `start` is less than one
```
Jatinagarwal24: hi @HarshitaKalani should i keep this change or restore the previous error message.
In meaning both are same.
Thanks !

tests/reference/asr-string_negative_start_index-643e9c0.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: The first index in string section is less than 1
+semantic error: Substring `start` is less than one
```
Jatinagarwal24: Similarly for this also.

tests/reference/asr-continue_compilation_3-435a232.stderr
HarshitaKalani: I think this is fine. However, does that make the code that throws `The first index in string section is less than 1` error unused? If so, let's remove that?

Jatinagarwal24: Good point. I removed the new start-index diagnostic from this patch and kept the existing one unchanged. The fix now only rejects BOZ substring bounds and adds the missing fixed-length upper-bound check.
Thanks!

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10071,6 +10079,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             Level::Error, Stage::Semantic, {Label("", {m_left_expr->base.loc})}));
                         throw SemanticAbort();
                     }
+                    if (is_boz_integer_constant(m_left_expr)) {
+                        diag.add(Diagnostic("Substring start index must be of type integer",
+                            Level::Error, Stage::Semantic, {Label("", {m_left_expr->base.loc})}));
+                        throw SemanticAbort();
+                    }
```
certik, line 10086: This error is not tested.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10106,10 +10124,29 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             Level::Error, Stage::Semantic, {Label("", {m_step_expr->base.loc})}));
                         throw SemanticAbort();
                     }
+                    if (is_boz_integer_constant(m_step_expr)) {
+                        diag.add(Diagnostic("Substring stride must be of type integer",
+                            Level::Error, Stage::Semantic, {Label("", {m_step_expr->base.loc})}));
+                        throw SemanticAbort();
+                    }
```
certik, line 10131: Not tested.

certik (approved):
Can you please add the two tests in a subsequent PR?

PR #11591, opened 2026-05-20
fix: ICE for invalid mixed-type CASE range selector 

### LINKED ISSUE #11582: ICE for invalid mixed-type CASE range selector
## MRE

```fortran id="mre7315"
program invalid_case_range
    implicit none

    integer :: value

    select case (value)
        case ('3' : 1.)
    end select
end program invalid_case_range
```

## Output

```text id="out7315"

Internal Compiler Error:
AssertFailed: ASRUtils::extract_type(l_type)->type == ASRUtils::extract_type(r_type)->type
```

## gfortran Behavior

```text id="out7316"

Error: Expression in CASE selector at (1) cannot be REAL(4)
```

opixdown:
fixes #11582

Also added blankspaces in `continue_compilation_1.f90`

### DIFF
--- a/tests/errors/continue_compilation_1.f90
+++ b/tests/errors/continue_compilation_1.f90
@@ -654,6 +654,10 @@ program continue_compilation_1
     dimension :: m(3)
     open(newunit=u, file="test.dat", status="replace", asynchronous=1)
     open(newunit=u, file="test.dat", status="replace", asynchronous="yes", asynchronous="no")
+    select case (bad_x)
+        case ('3' : 1.)
+    end select
+
     contains
     subroutine test_uminus_struct()
         use continue_compilation_1_mod, only: MyClass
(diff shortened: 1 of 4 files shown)


tests/errors/continue_compilation_1.f90
```diff
@@ -652,6 +652,21 @@ program continue_compilation_1
...
+    
+
+
+
+
+
+
+
+
+
```
certik: This changes the diff. I don't know if we need all the "contains" below. Maybe it can be moved into a module. Then we just just keep appending at the end of the file. The above solution doesn't scale, as we'll hit this issue again.

opixdown: i tried moving all the subroutines in the `contains` to module part but it is causing ice

tests/errors/continue_compilation_1.f90
certik, line 684: I think we need to refactor this file. Can you send a PR that moves these subroutines into a module at the top? Then we should be able to keep appending statements at the bottom without disturbing the line numbers in error messages.

opixdown: i tried it but it is causing ice i opened
#11597

PR #11593, opened 2026-05-20
Fix some netlib bugs

jinangshah21:
Towards #11090


integration_tests/CMakeLists.txt
```diff
@@ -2106,6 +2105,7 @@ RUN(NAME implicit_interface_35 LABELS gfortran llvm EXTRA_ARGS --implicit-typing
 RUN(NAME implicit_interface_36 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_37 LABELS gfortran llvm EXTRA_ARGS --implicit-interface --legacy-array-sections)
 RUN(NAME implicit_interface_38 LABELS gfortran llvm EXTRA_ARGS --implicit-interface EXTRAFILES implicit_interface_38b.f90)
+RUN(NAME implicit_interface_39 LABELS gfortran llvm EXTRA_ARGS --implicit-interface EXTRAFILES implicit_interface_39b.f90 implicit_interface_39c.f90)
```
certik, line 2108: I see that in main we get for this test:
```console
(lf) ~/repos/lfortran/integration_tests(issue_2)$ ./run_tests.py -t implicit_interface_39
...
[100%] Linking Fortran executable implicit_interface_39
duplicate symbol '__lcompilers_optimization_mod_i32' in:
    /Users/ondrej/repos/lfortran/integration_tests/test-llvm/CMakeFiles/implicit_interface_39.dir/implicit_interface_39c.f90.o
    /Users/ondrej/repos/lfortran/integration_tests/test-llvm/CMakeFiles/implicit_interface_39.dir/implicit_interface_39b.f90.o
ld: 1 duplicate symbols
The command 'clang -o implicit_interface_39 CMakeFiles/implicit_interface_39.dir/implicit_interface_39.f90.o CMakeFiles/implicit_interface_39.dir/implicit_interface_39b.f90.o CMakeFiles/implicit_interface_39.dir/implicit_interface_39c.f90.o  -L"/Users/ondrej/repos/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/ondrej/repos/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
...
```
So the symbol `__lcompilers_optimization_mod_i32` is duplicated. In the past we solved these issues with proper mangling, we add a unique hash to the function name, that way it is not duplicated. Why doesn't that approach work here?

The way it was solved in this PR is to use `internal` linkage, however I don't know what implications that has for all the backends. It seems the CI passes, so maybe it works.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7889,6 +7889,20 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
     void instantiate_function(const ASR::Function_t &x){
         llvm::DIScope* debug_current_scope_copy = debug_current_scope;
         uint32_t h = get_hash((ASR::asr_t*)&x);
+        auto get_function_linkage = [](const ASR::Function_t& fn) {
+            const std::string fn_name = fn.m_name;
+            ASR::FunctionType_t *ftype = ASRUtils::get_FunctionType(fn);
+            const bool is_compiler_generated_helper =
+                fn_name.rfind("_lcompilers_", 0) == 0 &&
```
certik, line 7896: This string comparison is fragile (we should not be dispatching based on string names of functions) and slow (we should not be doing string comparisons at all).

The proper solution is to encode this "compiler_generated_helper" flag in ASR itself, then dispatch based on that.

However, as mentioned in my other comment, we should not be special casing compiler generated helpers in any way. 

So let's figure out a more general solution, either a simple flag in ASR, perhaps "unique". And investigate how we already handle this exact case using mangling and why it didn't trigger in this case.

certik (commented):
You might need to separate the hollerith fix and let's get it in.

The duplicate symbol issue I think might not be the right fix.

jinangshah21:
Opened PR #11602 for hollerith format

PR #11595, opened 2026-05-21
Fix : check derivedtypes properly when matching arguments against parameters

assem2002:
use `can_pass_derviedtype_arg_to_parameter()` instead of ` is_deriatved_type_similar()` to check if I can pass derivedtype argument to a functionparameter.
It should tolerate the idea of parameter being of type `class(*)`

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -84,6 +84,7 @@ ASR::symbol_t* get_struct_sym_from_struct_expr(ASR::expr_t* expression);
 void set_struct_sym_to_struct_expr(ASR::expr_t* expression, ASR::symbol_t* struct_sym);
 
 ASR::symbol_t* get_union_sym_from_union_expr(ASR::expr_t* expression);
+static inline bool is_unlimited_polymorphic_type(ASR::Struct_t* st);
 
 static inline std::string extract_real(const char *s) {
     // TODO: this is inefficient. We should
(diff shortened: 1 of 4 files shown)


jinangshah21 (approved):

src/libasr/asr_utils.h
```diff
@@ -4409,15 +4410,20 @@ inline bool is_parent(ASR::Struct_t* a, ASR::Struct_t* b) {
...
+ */
 inline bool is_derived_type_similar(ASR::Struct_t* a, ASR::Struct_t* b) {
-    auto is_upoly = [](ASR::Struct_t* s) {
-        return s->m_struct_signature != nullptr &&
-            ASR::down_cast<ASR::StructType_t>(
-                s->m_struct_signature)->m_is_unlimited_polymorphic;
-    };
     return a == b || is_parent(a, b) || is_parent(b, a) ||
-        (is_upoly(a) && is_upoly(b));
+        (is_unlimited_polymorphic_type(a) && is_unlimited_polymorphic_type(b));
```
certik, line 4421: What is the implication of changing this rule for other usages of `is_derived_type_similar` that we have?

Is this relaxing it or making it more strict?

If it is more strict now, then since tests pass, we can merge.

If it is relaxing it, then I need to understand exactly how it is relaxing it and if we want to do it. If not, the alternative is to create a new function that you can use just for this PR.

assem2002: If you're referring to why we changed a call to `is_upoly()` --> to a call to `is_unlimited_polymorphic_type`, the answer is that we gain nothing from that change; 
`is_unlimited_polymorphic_type` == `is_upoly()` 100%. I just reused a utility that we already have -- No more than a minor maintainability for the codebase :) .

certik (approved):
Ok, it's probably ok to merge.

PR #11599, opened 2026-05-21
Fix some netlib bugs

jinangshah21:
Towards #11090

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -6948,6 +6948,27 @@ public:
                 }
                 return llvm::ConstantStruct::get(complex_type, {re, im});
             }
+            case ASR::exprType::ArrayConstant: {
+                // Infer LLVM element/array type from the ASR expression
+                llvm::Type* elem_type = nullptr;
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6948,6 +6948,27 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                llvm::Value* v = llvm_utils->declare_string_constant(sc);
+                if (!v) break;
+                if (llvm::GlobalVariable* gv = llvm::dyn_cast<llvm::GlobalVariable>(v)) {
+                    if (gv->hasInitializer()) {
+                        return gv->getInitializer();
+                    }
+                }
+                if (llvm::Constant* cc = llvm::dyn_cast<llvm::Constant>(v)) {
+                    return cc;
+                }
```
certik: I don't like these dynamic casts at runtime. Slow and fragile.

Is there a direct way to infer this from ASR?

certik (approved):
Otherwise good. If you know how to fix it, you can merge this PR and fix in a subsequent PR.

src/libasr/codegen/asr_to_llvm.cpp
jinangshah21: I have removed that.

PR #11614, opened 2026-05-23
fix: handle array constructor constant size correctly

### LINKED ISSUE #11578: Shape function invocation in reshape's shape argument rejected
The code below uses `reshape` to create an array that has the same values as `reshape`'s `source` argument, the result array has a higher rank than the `source` argument.  Four other compilers accept the code below.  I suspect the reason that `lfortran` rejects the code relates to the following description in the text in the [Fortran 2023 interpretation document](https://go.lbl.gov/fortran-2023) clause 16.9.175 line 26: 

"SHAPE   shall be a rank-one integer array. SIZE (x), where x is the actual argument corresponding to SHAPE, shall be a constant expression whose value is positive and less than 16. It shall not have an element whose value is negative."

I believe the relevant aspect of the code below is that `size([shape(rank1),1])` has the constant value 2 so the code meets the above requirement.

```
% cat add-dimension.f90 
program add_dimension
  implicit none
  integer, allocatable :: rank1(:)
  allocate(rank1(1), source=0)
  associate(rank2 => reshape(rank1, [shape(rank1),1]))
  end associate
end program

% lfortran add-dimension.f90 
semantic error: `shape` array in reshape intrinsic should be of constant size
 --> add-dimension.f90:5:37
  |
5 |   associate(rank2 => reshape(rank1, [shape(rank1),1]))
  |                                     ^^^^^^^^^^^^^^^^ not a constant size array


% lfortran --version
LFortran version: 0.60.0-421-ge2c448c79
Platform: macOS ARM
LLVM: 11.1.0
Default target: arm64-apple-darwin25.4.0
```

jinangshah21:
Fixes #11578


src/libasr/asr_utils.h
```diff
@@ -7676,9 +7676,71 @@ inline ASR::asr_t* make_ArrayConstructor_t_util(Allocator &al, const Location &a
...
+        if (ASR::is_a<ASR::ImpliedDoLoop_t>(*arg) ||
+            ASR::is_a<ASR::ArrayConstructor_t>(*arg) ||
+            ASR::is_a<ASR::ArrayConstant_t>(*arg)) {
+            all_scalars = false;
+        }
+        if (ASRUtils::is_array(ASRUtils::expr_type(arg))) {
+            all_scalars = false;
+            if (!ASRUtils::is_fixed_size_array(ASRUtils::expr_type(arg))) {
+                can_compute_size = false;
+                break;
```
assem2002, line 7702: Can we use `compute_array_constructor_element_count` at `ast_common_visitor`?
maybe port the function into `ASR_utils.h`

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -5899,6 +5901,14 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
                         if( temp_type->type == ASR::ttypeType::Array ) {
                             int64_t arr_size = ASRUtils::get_fixed_size_of_array(temp_type);
+                            int64_t start, end;
+                            if (ASR::is_a<ASR::ImpliedDoLoop_t>(*temp)) {
+                                ASR::ImpliedDoLoop_t* impl_do = ASR::down_cast<ASR::ImpliedDoLoop_t>(temp);
+                                if (ASRUtils::extract_value(ASRUtils::expr_value(impl_do->m_start), start) &&
+                                    ASRUtils::extract_value(ASRUtils::expr_value(impl_do->m_end), end)) {
+                                    arr_size *= end - start + 1;
+                                }
+                            }
```
assem2002, line 5911: I think we can call `compute_implied_do_element_count` from `ast_common_visitor` also

assem2002 (approved):
LGTM. Logic is correct.
Please address my suggestions.

certik (approved):
I think @assem2002's suggestions can be fixed in a subsequent PR.

PR #11615, opened 2026-05-23
Refactor : Change `insert_deallocate` into `intent_out_deallocate`

assem2002:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -11433,28 +11433,6 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 }
             }
             return;
-        } else if( is_target_struct && is_value_struct ) {
```
assem2002, line 11436: This is dead code (in main) -- I just removed it

certik (approved):
This looks really good, thanks for refactoring this!

src/libasr/pass/intent_out_deallocate.cpp
certik, line 27: Is this comment still accurate to what this pass does?

assem2002: Yes. The pass apparently skips some fortran functions that LFortran creates (helpers or intrinsics).

Is it right ? I don't think so, Why LFortran functions with any INTENT_OUT happen to have problems?

Is it wrong ? No, the user get what he expects. His function are well handled

PR #11616, opened 2026-05-23
Fix two netlib bugs

jinangshah21:
Towards #11090


certik:
While reviewing this, I noticed this bug: https://github.com/lfortran/lfortran/pull/11618.

PR #11619, opened 2026-05-24
fix: handle positional string constant as format specifier in READ

### LINKED ISSUE #11206: bug: `format` misinterpreted as `unit` in `FileRead` when only the `format` is provided
Found while working on #11124 

```fortran
program main
   integer :: a
   a = 5
   read ("(I10)")
end program main
```

#### LFortran ASR

See `FileRead` node's `unit` field.
<details>

```json
{
    "node": "TranslationUnit",
    "fields": {
        "symtab": {
            "node": "SymbolTable1",
            "fields": {
                "main": {
                    "node": "Program",
                    "fields": {
                        "symtab": {
                            "node": "SymbolTable2",
                            "fields": {
                                "a": {
                                    "node": "Variable",
                                    "fields": {
                                        "parent_symtab": 2,
                                        "name": "a",
                                        "dependencies": [],
                                        "intent": "Local",
                                        "symbolic_value": [],
                                        "value": [],
                                        "storage": "Default",
                                        "type": {
                                            "node": "Integer",
                                            "fields": {
                                                "kind": 4
                                            }
                                        },
                                        "type_declaration": [],
                                        "abi": "Source",
                                        "access": "Public",
                                        "presence": "Required",
                                        "value_attr": false,
                                        "target_attr": false,
                                        "contiguous_attr": false,
                                        "bindc_name": [],
                                        "is_volatile": false,
                                        "is_protected": false,
                                        "pass_attr": "NotMethod",
                                        "self_argument": [],
                                        "corank": 0
                                    }
                                }
                            }
                        },
                        "name": "main",
                        "dependencies": [],
                        "body": [
                            {
                                "node": "Assignment",
                                "fields": {
                                    "target": {
                                        "node": "Var",
                                        "fields": {
                                            "v": "a (SymbolTable2)"
                                        }
                                    },
                                    "value": {
                                        "node": "IntegerConstant",
                                        "fields": {
                                            "n": 5,
                                            "type": {
                                                "node": "Integer",
                                                "fields": {
                                                    "kind": 4
                                                }
                                            },
                                            "intboz_type": "Decimal"
                                        }
                                    },
                                    "overloaded": [],
                                    "realloc_lhs": false,
                                    "move_allocation": false
                                }
                            },
                            {
                                "node": "FileRead",
                                "fields": {
                                    "label": 0,
                                    "unit": {
                                        "node": "StringConstant",
                                        "fields": {
                                            "s": "(I10)",
                                            "type": {
                                                "node": "String",
                                                "fields": {
                                                    "kind": 1,
                                                    "len": {
                                                        "node": "IntegerConstant",
                                                        "fields": {
                                                            "n": 5,
                                                            "type": {
                                                                "node": "Integer",
                                                                "fields": {
                                                                    "kind": 4
                                                                }
                                                            },
                                                            "intboz_type": "Decimal"
                                                        }
                                                    },
                                                    "len_kind": "ExpressionLength",
                                                    "physical_type": "DescriptorString"
                                                }
                                            }
                                        }
                                    },
                                    "fmt": [],
                                    "iomsg": [],
                                    "iostat": [],
                                    "advance": [],
                                    "size": [],
                                    "id": [],
                                    "pos": [],
                                    "values": [],
                                    "overloaded": [],
                                    "is_formatted": false,
                                    "nml": [],
                                    "rec": [],
                                    "pad": []
                                }
                            }
                        ]
                    }
                }
            }
        },
        "items": []
    }
}

```
</details>

#### GFortran's `original` GIMPLE IR

```c
__attribute__((fn spec (". ")))
void MAIN__ ()
{
  integer(kind=4) a;

  a = 5;
  {
    struct __st_parameter_dt dt_parm.0;

    dt_parm.0.common.filename = &"test.f90"[1]{lb: 1 sz: 1};
    dt_parm.0.common.line = 4;
    dt_parm.0.format = &"(I10)"[1]{lb: 1 sz: 1};
    dt_parm.0.format_len = 5;
    dt_parm.0.common.flags = 4096;
    dt_parm.0.common.unit = 5;
    _gfortran_st_read (&dt_parm.0);
    _gfortran_st_read_done (&dt_parm.0);
  }
}


__attribute__((externally_visible))
integer(kind=4) main (integer(kind=4) argc, character(kind=1) * * argv)
{
  static integer(kind=4) options.1[7] = {10308, 16383, 0, 1, 1, 0, 31};

  _gfortran_set_args (argc, argv);
  _gfortran_set_options (7, &options.1[0]);
  MAIN__ ();
  return 0;
}
```

The text inside `read` is treated as `format`. `unit` defaults to `5` if nothing is provided. 5 is the value for `stdin`.

ubaidsk:
fixes https://github.com/lfortran/lfortran/issues/11206

Edit: Note that in the following description `read("(I10)"), x` is not valid fortran syntax, hence in the PR changes (including the test case), we have used `read "(I10)", x` which reproduces the issue and this PR fixes it. This PR handles both the cases `read "(I10)", x` and `read("(I10)"), x`.

As per the `main` branch lfortran, here is the issue with `read "(I10)", x`:

```shell
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % cat integration_tests/read_97.f90 
! Test READ fmt [, iolist] abbreviated form where the format specifier
! is given as a positional string constant (not an explicit fmt= keyword).
! E.g. read "(I10)", x  is equivalent to  read(*, "(I10)") x
program read_97
    implicit none
    integer :: x
    ! The string literal "(I10)" here is the format specifier, not a unit.
    ! This exercises the semantic fix that moves it to a_fmt and sets
    ! a_unit = nullptr (stdin).
    read "(I2)", x
    print *, x
end program
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % ./src/bin/lfortran integration_tests/read_97.f90 -o lfout && echo "12345" | ./lfout
12345
```

Here are the AST and the buggy ASR for observation:

<details><summary>AST and ASR</summary>

```shell
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % ./src/bin/lfortran integration_tests/read_97.f90 --show-a
st     
(TranslationUnit
    [(Program
        read_97
        ()
        []
        [(ImplicitNone
            []
            ()
        )]
        [(Declaration
            (AttrType
                TypeInteger
                []
                ()
                ()
                None
            )
            []
            [(x
            []
            []
            ()
            ()
            None
            ())]
            (TriviaNode
                []
                [(EndOfLine)
                (Comment
                    "! The string literal \"(I10)\" here is the format specifier, not a unit."
                )
                (Comment
                    "! This exercises the semantic fix that moves it to a_fmt and sets"
                )
                (Comment
                    "! a_unit = nullptr (stdin)."
                )]
            )
        )]
        [(Read
            0
            (String "(I2)" ())
            []
            []
            [x]
            ()
        )
        (Print
            0
            ()
            [x]
            ()
        )]
        []
    )]
)
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % ./src/bin/lfortran integration_tests/read_97.f90 --show-a
sr
(TranslationUnit
    (SymbolTable
        1
        {
            read_97:
                (Program
                    (SymbolTable
                        2
                        {
                            x:
                                (Variable
                                    2
                                    x
                                    []
                                    Local
                                    ()
                                    ()
                                    Default
                                    (Integer 4)
                                    ()
                                    Source
                                    Public
                                    Required
                                    .false.
                                    .false.
                                    .false.
                                    ()
                                    .false.
                                    .false.
                                    NotMethod
                                    ()
                                    []
                                )
                        })
                    read_97
                    []
                    [(FileRead
                        0
                        ()
                        ()
                        ()
                        ()
                        ()
                        ()
                        ()
                        ()
                        [(Var 2 x)]
                        ()
                        .false.
                        ()
                        ()
                        ()
                    )
                    (Print
                        (StringFormat
                            ()
                            [(Var 2 x)]
                            FormatFortran
                            (Allocatable
                            )
                            ()
                        )
                    )]
                )
        })
    []
)
```

</details>

## Summary

- `read("(I10)"), x` was miscompiled: the string `"(I10)"` was assigned to the unit field instead of the format field, causing incorrect runtime behaviour.
- Fixed `init_record_state` in the runtime to guard against non-seekable streams (e.g. stdin piped from a shell) where `record_start_pos` is `-1`.
- Added `integration_tests/read_95.f90` with a reference ASR test that confirms the fix.

## The Bug

Fortran allows two forms of a formatted READ from stdin:

```fortran
read(*, "(I10)") x   ! explicit: unit=*, fmt="(I10)"
read("(I10)"), x     ! abbreviated: string literal is the format
```

Both are standard Fortran (R910 / R912 in the spec). The second form is shorthand — when the control list contains only a single string-literal expression, that string is the format specifier and the unit defaults to stdin.

## Why the Fix Is in the Semantics Layer, Not the Parser

One might ask: why not add a grammar rule in the parser to handle `read("(I10)"), x` as a distinct case?

**The parser already handles the no-parentheses abbreviated form.** In `parser.yy`, rule `KW_READ TK_STRING "," expr_list` (the `READ5` macro) fires when the tokenizer sees a bare string literal immediately after `READ`. This is unambiguous at the grammar level.

**The parenthesised form, however, is syntactically identical to three completely different things:**

```fortran
read(unit_int),  x   ! unit is an integer variable → reads from a file
read(char_var),  x   ! unit is a character variable → internal read (from a string buffer)
read("(I10)"), x   ! unit slot holds a string literal → abbreviated fmt, stdin implied
```

All three match the same grammar production: `KW_READ "(" write_arg_list ")" "," expr_list`. The parser processes `write_arg_list` uniformly — it has no way to inspect whether the single positional argument is an integer, a character variable, or a string literal, because `write_arg_list` reduces to a list of generic AST expressions.

Disambiguating them requires type information, which is only available after semantic analysis (AST → ASR lowering). In the semantics visitor, we can inspect the lowered ASR node and check `ASR::is_a<ASR::StringConstant_t>(*a_unit)`. A parser-level fix would either require duplicating grammar productions to cover every combination (fragile and contradicts the spec's uniform treatment of the control list), or changing the AST structure — a far larger refactor.

The semantic layer is therefore the correct and minimal place for this fix, consistent with the project's architecture principle that type coercion and context-sensitive reinterpretation belong in AST→ASR.

## Runtime Fix

`init_record_state` pre-scans from `record_start_pos` to find the record length. When stdin is a pipe (non-seekable), `ftell` returns `-1`. Calling `fseek`/`fgetc` on a non-seekable stream silently consumed input that could never be recovered, corrupting subsequent reads. The fix skips the pre-scan when `record_start_pos < 0` and sets `record_length = LONG_MAX` as a sentinel.

## Verification

Reference ASR test confirms `FileRead` has unit=nil (stdin) and fmt=`StringConstant "(I10)"`:

```
(FileRead
    0
    ()                         ← unit = nil (stdin)
    (StringConstant "(I10)"  ← fmt = the string literal
    ...
    .true.                     ← formatted = true
```

```
cd tests && PATH=../src/bin:$PATH python run_tests.py -t read_95 -s
../integration_tests/read_95.f90 * asr ✓
TESTS PASSED
```


ubaidsk:
Rebased on top of the latest main and conflicts resolved.

tests/tests.toml
```diff
@@ -5227,6 +5227,10 @@ asr = true
 filename = "../integration_tests/read_83.f90"
 llvm = true
 
+[[test]]
+filename = "../integration_tests/read_97.f90"
+asr_json = true
```
certik: Why not just `asr = true` like all the other tests?

integration_tests/read_97.f90
```diff
@@ -0,0 +1,13 @@
...
+program read_97
+    implicit none
+    integer :: x
+    x = 42
+    ! The string literal "(I10)" here is the format specifier, not a unit.
+    ! This exercises the semantic fix that moves it to a_fmt and sets
+    ! a_unit = nullptr (stdin).
+    read ("(I10)"), x
+    print *, x
+end program
```
certik, line 12: According to Grok, this test is not valid Fortran: https://grok.com/share/bGVnYWN5_79f7a734-4d9c-4d27-8c46-f5e5c14a2998, and the comment is also incorrect.

certik:
I think this PR shows the danger of only testing LFortran via reference tests, and not with other compilers. According to Grok, this is not valid Fortran, so we should not support it, unless it's an extension, then we should give a warning, etc. Unless of course Grok is incorrect, in which case let's figure out what the right answer is.

integration_tests/read_97.f90
ubaidsk: It works for me locally:

```shell
lfortran % cat integration_tests/read_97.f90
! Test READ fmt [, iolist] abbreviated form where the format specifier
! is given as a positional string constant (not an explicit fmt= keyword).
! E.g. read("(I10)"), x  is equivalent to  read(*, "(I10)") x
program read_97
    implicit none
    integer :: x
    x = 42
    ! The string literal "(I10)" here is the format specifier, not a unit.
    ! This exercises the semantic fix that moves it to a_fmt and sets
    ! a_unit = nullptr (stdin).
    read ("(I10)"), x
    print *, x
end program
lfortran % gfortran integration_tests/read_97.f90 -o gfout && echo "10" | ./gfout
clang: warning: overriding deployment version from '16.0' to '26.0' [-Woverriding-deployment-version]
          10
lfortran % ./src/bin/lfortran integration_tests/read_97.f90 -o lfout && echo "10" | ./lfout
10
ubaid@Mohammeds-MacBook-Pro lfortran % 
```

ubaidsk: The clang warning is I believe some version mismatch at my end. I get that for any other fortran code compilation using gfortran.

Here are the clang and gfortran versions on my system:

```shell
lfortran % clang --version
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: arm64-apple-darwin25.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
lfortran % gfortran --version
GNU Fortran (Homebrew GCC 14.2.0_1) 14.2.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

lfortran % 
```

tests/tests.toml
ubaidsk: Yes, we can do that. I used the JSON format, as that is more explicit in terms of the different ASR node keys and the values that are being assigned to it. In this case, with the help of the reference test, we want to confirm that the format string Iis correctly being interpreted and assigned to the `fmt` field in ASR and not the `unit` field in ASR as it happens for the `main` branch as seen in the `Details` section here in the original issue description [here](https://github.com/lfortran/lfortran/issues/11206#issue-4307879802).

ubaidsk: Updated to use `asr`.

integration_tests/read_97.f90
ubaidsk: I have updated this to be valid fortran code. 

Here is the working of it:

```shell
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % cat integration_tests/read_97.f90 
! Test READ fmt [, iolist] abbreviated form where the format specifier
! is given as a positional string constant (not an explicit fmt= keyword).
! E.g. read "(I10)", x  is equivalent to  read(*, "(I10)") x
program read_97
    implicit none
    integer :: x
    ! The string literal "(I10)" here is the format specifier, not a unit.
    ! This exercises the semantic fix that moves it to a_fmt and sets
    ! a_unit = nullptr (stdin).
    read "(I2)", x
    print *, x
end program
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % gfortran integration_tests/read_97.f90 -o gfout && echo "12345" | ./gfout          
clang: warning: overriding deployment version from '16.0' to '26.0' [-Woverriding-deployment-version]
          12
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % ./src/bin/lfortran integration_tests/read_97.f90 -o lfout && echo "12345" | ./lfout
12
(lf) ubaid@Mohammeds-MacBook-Pro lfortran % 
```

We can not yet add this as an integration test because this test needs a stdin value to work and our integration tests infrastructure doesn't yet support specifying a stdin for tests.

certik: Yes, this is valid Fortran now.

certik (approved):
I think that's fine now.

PR #11622, opened 2026-05-24
fix: Update Function Call for requirement function through its ExternalSymbol by resolving it in the updated scope

### LINKED ISSUE #11574: symbol table not found in scope when instantiating with an operator
When compiling the code I get the following errors:

```

ASR verify pass error: The symbol table was not found in the scope of `symtab`.
  --> ImpureElementalInstantiationError.f90:17:5 - 21:16
   |
17 |        elemental function lt_my_type(lhs, rhs) result(res)
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
...
   |
21 |        end function
   | ...^^^^^^^^^^^^^^^^ failed here

ASR verify pass error: FunctionCall::m_name `lt_my_type` cannot point outside of its symbol table
  --> ImpureElementalInstantiationError.f90:50:25
   |
50 |             res = all ( lt(lhs,rhs) )
   |                         ^^^^^^^^^^^ failed here
```
code:
```
module lt_m
    implicit none
    private
    public :: my_type, lt_my_type

    type my_type
        real :: d
    end type

    interface operator(<)
       procedure lt_my_type
    end interface

contains

    impure function lt_my_type(lhs, rhs) result(res)
        type(my_type), intent(in) :: lhs, rhs
        logical :: res
        res = lhs%d < rhs%d
    end function

end module

module template_apply_m
    implicit none
    private
    public :: apply_t

    requirement op_r(T, U, V, op_func)
        type, deferred :: T
        type, deferred :: U
        type, deferred :: V
            type(T), intent(in) :: lhs
            type(U), intent(in) :: rhs
            type(V) :: res
        end function
    end requirement

    template apply_t(T, lt)
        require :: op_r(T, T, logical, lt)
        private
        public :: apply_lt
    contains
        pure function apply_lt(lhs, rhs) result(res)
            type(T), intent(inout) :: lhs(:)
            type(T), intent(inout) :: rhs(:)
            type(logical) :: res
            res = all ( lt(lhs,rhs) )
        end function
    end template
end module

program test
        use lt_m
        use template_apply_m

        instantiate apply_t(my_type,operator(<)), only : my_apply => apply_lt


        type(my_type) :: a(5), b(5)
        type(my_type) :: x
        type(my_type) :: y

        x = my_type(5)
        y = my_type(6)
        a = x
        b = y
        print *, "TEST"
        print *, x < y
        ! print *, a < b
        print *, "a=", a
        print *, "b=", b
        print *, my_apply(a,b)

end program
```

AnshMehta1:
Fixes #11574

### DIFF
--- a/src/libasr/pass/instantiate_template.cpp
+++ b/src/libasr/pass/instantiate_template.cpp
@@ -1567,6 +1567,12 @@ public:
         // requirement function
         if (symbol_subs.find(call_name) != symbol_subs.end()) {
             name = symbol_subs[call_name];
+
+            if (ASRUtils::symbol_parent_symtab(name)->get_counter() != new_scope->get_counter()) {
+                std::string sym_name = ASRUtils::symbol_name(name);
+                ASR::symbol_t* scoped_sym = new_scope->resolve_symbol(sym_name);
+                name = scoped_sym;
+            }
         }
 
         // function call found in body that needs to be instantiated
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3233,6 +3233,7 @@ RUN(NAME template_03 LABELS llvm llvm_wasm llvm_wasm_emcc wasm)
 # RUN(NAME template_04 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME template_05 LABELS llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME template_06 LABELS llvm llvm_wasm llvm_wasm_emcc wasm fortran)
+RUN(NAME template_07 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME template_add_01 LABELS llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME template_add_01b LABELS llvm llvm_wasm llvm_wasm_emcc)             # FIXME: add visit_OverloadedBinOp to wasm
 RUN(NAME template_add_01c LABELS llvm llvm_wasm llvm_wasm_emcc wasm)
--- a/integration_tests/template_07.f90
+++ b/integration_tests/template_07.f90
@@ -1,0 +1,78 @@
+module lt_m_template_07
+    implicit none
+    private
+    public :: my_type, lt_my_type
+
+    type my_type
+        real :: d
+    end type
+
+    interface operator(<)
+       procedure lt_my_type
+    end interface
+
+contains
+
+    pure elemental function lt_my_type(lhs, rhs) result(res)
+        type(my_type), intent(in) :: lhs, rhs
+        logical :: res
+        res = lhs%d < rhs%d
+    end function
+
+end module
+
+module template_apply_m_template_07
+    implicit none
+    private
+    public :: apply_t
+
+    requirement op_r(T, U, V, op_func)
+        type, deferred :: T
+        type, deferred :: U
+        type, deferred :: V
+        pure elemental function op_func(lhs, rhs) result(res)
+            type(T), intent(in) :: lhs
+            type(U), intent(in) :: rhs
+            type(V) :: res
+        end function
+    end requirement
+
+    template apply_t(T, lt)
+        require :: op_r(T, T, logical, lt)
+        private
+        public :: apply_lt
+    contains
+        pure function apply_lt(lhs, rhs) result(res)
+            type(T), intent(inout) :: lhs(:)
+            type(T), intent(inout) :: rhs(:)
+            logical :: res
+            res = all ( lt(lhs,rhs) )
+        end function
+    end template
+end module
+
+program template_07
+        use lt_m_template_07
+        use template_apply_m_template_07
+
+        instantiate apply_t(my_type,operator(<)), only : my_apply => apply_lt
+
+
+        type(my_type) :: a(5), b(5)
+        type(my_type) :: x
+        type(my_type) :: y
+
+        x = my_type(5)
+        y = my_type(6)
+        a = x
+        b = y
+
+        print *, x < y
+        if (.not. (x < y)) error stop 
+
+        print *, "a=", a
+        print *, "b=", b
+
+        print *, my_apply(a,b)
+        if (.not. my_apply(a, b)) error stop
+end program


src/libasr/pass/instantiate_template.cpp
```diff
@@ -1567,6 +1567,19 @@ class BodyInstantiator : public ASR::BaseExprStmtDuplicator<BodyInstantiator>
...
+                ASR::symbol_t* existing_sym = new_scope->resolve_symbol(sym_name);
+
+                if (existing_sym) {
+                    name = existing_sym;
+                } else {
+                    ASRUtils::SymbolDuplicator sd(al);
+                    sd.duplicate_symbol(name, new_scope);
+                    name = new_scope->resolve_symbol(sym_name);
+                }
+            }
```
kmr-srbh, line 1575: Could you please explain how this fixes the issue?

AnshMehta1: Yes. So, just one line above the changes, we have `name = symbol_subs[call_name];`. This basically retrieves the concrete instantiated symbol corresponding to the template/requirement function. However, the retrieved symbol may belong to a different symbol table (for example, the original module scope), while the currently instantiated function body is being generated inside new_scope. Due to this, we had ASR verify pass error of `symbol out of current SymbolTable`. Thus, this fix uses existing symbol from new scope (if present) or duplicates the symbol according to new scope.

kmr-srbh: I checked the ASR for this, I see an incorrect `FunctionCall` node:

```clojure
[(Assignment
    (Var 24 res)
    (IntrinsicArrayFunction
        All
        [(FunctionCall
            14 lt_my_type
            ()
            [((Var 24 lhs))
            ((Var 24 rhs))]
            (Array
                (Logical 4)
                (ArraySize
                    (Var 24 lhs)
                    (Integer 4)
                    ()
                ))]
                PointerArray
            )
            ()
            ()
        )]
        0
        (Logical 4)
        ()
    )
    ()
    .false.
    .false.
)]
```

Here, the function call is made to symbol `14 lt_my_type`, which is out of scope (14 is module `lt_m`). We have an external symbol `23 lt_my_type` for it, so we can use it here.

certik: We should strengthen our ASR verify to catch this.

AnshMehta1: > I checked the ASR for this, I see an incorrect `FunctionCall` node:
> 
> ```clojure
> [(Assignment
>     (Var 24 res)
>     (IntrinsicArrayFunction
>         All
>         [(FunctionCall
>             14 lt_my_type
>             ()
>             [((Var 24 lhs))
>             ((Var 24 rhs))]
>             (Array
>                 (Logical 4)
>                 [((IntegerConstant 1 (Integer 4) Decimal)
>                 (ArraySize
>                     (Var 24 lhs)
>                     (IntegerConstant 1 (Integer 4) Decimal)
>                     (Integer 4)
>                     ()
>                 ))]
>                 PointerArray
>             )
>             ()
>             ()
>         )]
>         0
>         (Logical 4)
>         ()
>     )
>     ()
>     .false.
>     .false.
> )]
> ```
> 
> Here, the function call is made to symbol `14 lt_my_type`, which is out of scope (14 is module `lt_m`). We have an external symbol `23 lt_my_type` for it, so we can use it here.

For me the ASR is correct 

```clojure
[(Assignment
    (Var 12 res)
    (IntrinsicArrayFunction
        All
        [(FunctionCall
            11 lt_my_type
            ()
            [((Var 12 lhs))
            ((Var 12 rhs))]
            (Array
                (Logical 4)
                (ArraySize
                    (Var 12 lhs)
                    (Integer 4)
                    ()
                ))]
                PointerArray
            )
            ()
            ()
        )]
        0
        (Logical 4)
        ()
    )
    ()
    .false.
    .false.
)]
```

I think the FunctionCall is correct as in ASR, I am having `11 lt_my_type` where 11 is the correct SymbolTable. Your ASR is the faulty one which I was getting prior to this fix which was correctly detected by the ASR.

AnshMehta1: This `11 lt_my_type` is 

```clojure
lt_my_type:
(ExternalSymbol
    11
    lt_my_type
    2 lt_my_type
    lt_m_template_07
    []
    lt_my_type
    Public
),
```

I think, this is correct and in accordance with your previous reveiw.

kmr-srbh: Yes, the ASR with the fix will be correct since `existing_sym` is not `nullptr`, but I think we do not need to use the Symbol Duplicator here. I think we need to fix the symbol resolution logic for the symbol which is already present in either current, or an enclosing scope.

AnshMehta1: Yep, I actually added the symbol duplication logic mainly as a precautionary fallback, but for this specific issue it does not seem necessary since the correct `ExternalSymbol` is already present in the enclosing scope.

Currently, I’ll remove the symbol duplication part and instead focus on fixing the symbol resolution logic so that it correctly resolves and reuses the existing symbol from the enclosing scope.

AnshMehta1: Updated the PR with the required changes. I think, this if fine for this particular issue. Any subsequent updations could be made in another pr as per the requirement. Let me know if it is fine for now @kmr-srbh or any other changes that could be made.

kmr-srbh: I noticed the `duplicate_FunctionCall()` method of the `SymbolInstantiator` class looks identical. I think it is a more complete version and we can write this function like that.

certik:
@kmr-srbh is this ready to be merged, or requires changes (and what exact changes)?

src/libasr/pass/instantiate_template.cpp
AnshMehta1: The Function you are talking is also not able to handle out of scope ExternalSymbol. Thus, I think the current fix is okay for now.

kmr-srbh (approved):
I think the current change is fine.

PR #11623, opened 2026-05-24
Detect FINAL procedures through parent struct hierarchy

HarshitaKalani:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -6638,7 +6654,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         }
                     }
                 }
-                if( ASR::is_a<ASR::Variable_t>(*sym) && initialize_val) {
+                if( ASR::is_a<ASR::Variable_t>(*sym) && initialize_val &&
+                    !(is_intent_out && top_struct_has_finalizer)) {
```
assem2002: What are we attempting to do here?

HarshitaKalani: Good catch, It was a workaround, but I've updated the implementation now.

intent_out_deallocate now walks the inheritance chain to detect FINAL procedures, emits the entry-FINAL call(s), and then emits the corresponding default-initialization statements in the function body, in the correct order.
allocate_array_members_of_struct still keeps the gate, but it is no longer a "skip and forget" case. It is now a deliberate hand-off: for intent(out) struct dummies with FINAL procedures, codegen avoids doing prologue default initialization because the pass has already emitted the same initialization after the FINAL call.

integration_tests/intent_out_finalize_01.f90
```diff
@@ -0,0 +1,47 @@
+module intent_out_finalize_01_mod
```
assem2002, line 1: PR is attempting to call **FINAL** of the parent chain. 
I think we need a test for that.

src/libasr/pass/intent_out_deallocate.cpp
```diff
@@ -287,63 +346,84 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
+                    while (st_for_final != nullptr) {
+                        for (size_t fi = 0;
+                                fi < st_for_final->n_member_functions;
+                                fi++) {
+                            std::string final_proc_name =
+                                st_for_final->m_member_functions[fi];
+                            ASR::symbol_t* final_sym =
+                                st_for_final->m_symtab->parent->get_symbol(
+                                    final_proc_name);
+                            if (!final_sym) continue;
```
assem2002: Assert here. The behavior isn't defined here.

assem2002 (commented):
LGTM. Please address the comments below.

assem2002:
also we would be great if you can add a test for MRE of the failing CI.

assem2002:
Please note what the PR is attempting to fix with the current changes just pushed. Is it about Final functions in parent/child hierarchy ?

integration_tests/intent_out_finalize_03.f90
```diff
@@ -0,0 +1,51 @@
...
+      integer :: cid = 11
+   contains
+      final :: child_final
+   end type child_t
+contains
+   subroutine base_final(this)
+      type(base_t), intent(inout) :: this
+      g_last_finalized_bid = this%bid
+   end subroutine base_final
+
```
assem2002, line 23: This base_final won't get invoked

src/libasr/pass/intent_out_deallocate.cpp
```diff
@@ -287,26 +348,24 @@ class IntentOutDeallocateVisitor : public ASR::BaseWalkVisitor<IntentOutDealloca
...
-                //    finalization occurs before the procedure body executes."
-                if (struct_type->n_member_functions > 0) {
+                bool has_finalizer =
+                    struct_hierarchy_has_finalizer(struct_type);
+                if (has_finalizer) {
                     ASR::expr_t* var_expr = ASRUtils::EXPR(
                         ASR::make_Var_t(al, loc, arg_sym));
-                    for (size_t fi = 0; fi < struct_type->n_member_functions; fi++) {
+                    for (size_t fi = 0;
+                            fi < struct_type->n_member_functions; fi++) {
```
assem2002, line 361: Collect finalizer function in the chain (child -> parent -> grandparent -> ..) and invoke them in that order.

integration_tests/intent_out_finalize_03.f90
assem2002: You can see the test is checking `g_last_finalized_cid`, and not `g_last_finalized_bid`

assem2002:
## IMPORTANT
If current changes unblocks you, Merge this and open an issue for finalize-function-invoking chain

HarshitaKalani:
That would be good. I've opened an issue here: https://github.com/lfortran/lfortran/issues/11783.

PR #11625, opened 2026-05-24
fix: handle implicit procedure args in type-bound generics

### LINKED ISSUE #11624: semantic error: Arguments do not match for any generic procedure
mre1.f90
```
module test_mod
    implicit none

    type :: runner
    contains
        procedure :: run_a1
        generic :: run => run_a1
    end type

contains

    subroutine run_a1(this, a1, f)
        class(*), intent(in) :: a1
        procedure() :: f
    end subroutine

end module

program main
    use test_mod
    implicit none
    type(runner) :: br

    interface
        subroutine compute_pi(n)
            class(*), intent(in) :: n
        end subroutine
    end interface

    ! This call fails to match the generic procedure interface 'run' because of the 'procedure()' argument
    call br%run(100000000000_8, compute_pi)

end program
```
Lfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran -c mre1.f90 --implicit-interface
semantic error: Arguments do not match for any generic procedure, run
  --> mre1.f90:32:5
   |
32 |     call br%run(100000000000_8, compute_pi)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 


```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran -c mre1.f90 
```

Jatinagarwal24:
fixes #11624
This PR fixes the semantic error "Arguments do not match for any generic procedure" that occurs when calling a type-bound generic procedure where the formal argument is declared as `procedure() :: f `(implicit interface).

Changes
`src/libasr/asr_utils.cpp`

Updated `argument_types_match` to recognize implicit interface procedures (`procedure() :: f `and `external :: f`) as compatible with explicit interface actual arguments during generic dispatch resolution.
For the `Variable_t `branch: if the formal argument's type declaration is an implicit interface (identified by `_iface_implicit` in the name), skip strict type matching.
For the `Function_t` branch: if the actual argument is an implicit interface procedure (FunctionType with 0 arg types and Interface deftype), skip strict type matching — actual compatibility is checked at the call site.

`src/libasr/codegen/asr_to_llvm.cpp`

Fixed instantiate_function to only skip pre-declaration of Interface functions with class(*) args for submodule cases, not for program-scoped interfaces. The existing condition !m_module is false for both submodules and programs; added a parent scope check to exclude Program_t.
integration_tests/type_bound_generic_member_access_03.f90

New integration test covering generic dispatch with procedure() dummy arguments, including multiple overloads and internal procedures passed as actual arguments.

### DIFF
--- a/integration_tests/type_bound_generic_member_access_03.f90
+++ b/integration_tests/type_bound_generic_member_access_03.f90
@@ -1,0 +1,33 @@
+module type_bound_generic_member_access_03_m
+  implicit none
+  logical :: called = .false.
+  type :: runner
+  contains
+    procedure :: run_a1
+    generic :: run => run_a1
+  end type
+contains
+  subroutine run_a1(this, a1, f)
+    class(runner), intent(inout) :: this
+    class(*), intent(in) :: a1
+    procedure() :: f
+    call f()
+  end subroutine
+end module
+
+program type_bound_generic_member_access_03
+  use type_bound_generic_member_access_03_m
+  implicit none
+  type(runner) :: br
+  interface
+    subroutine my_sub()
+    end subroutine
+  end interface
+  call br%run(1, my_sub)
+  if (.not. called) error stop 1
+end program
+
+subroutine my_sub()
+  use type_bound_generic_member_access_03_m, only: called
+  called = .true.
+end subroutine
--- a/integration_tests/type_bound_generic_member_access_03.f90
+++ b/integration_tests/type_bound_generic_member_access_03.f90
@@ -4,7 +4,8 @@ module type_bound_generic_member_access_03_m
   type :: runner
   contains
     procedure :: run_a1
-    generic :: run => run_a1
+    procedure :: run_a2
+    generic :: run => run_a1, run_a2
   end type
 contains
   subroutine run_a1(this, a1, f)
@@ -13,21 +14,35 @@ contains
     procedure() :: f
     call f()
   end subroutine
+
+  subroutine run_a2(this, a1, f2, arg)
+    class(runner), intent(inout) :: this
+    class(*), intent(in) :: a1
+    procedure() :: f2
+    integer, intent(in) :: arg
+    call f2(arg)
+  end subroutine
 end module
 
 program type_bound_generic_member_access_03
   use type_bound_generic_member_access_03_m
   implicit none
   type(runner) :: br
-  interface
-    subroutine my_sub()
-    end subroutine
-  end interface
+  logical :: called_with_arg = .false.
+
   call br%run(1, my_sub)
   if (.not. called) error stop 1
-end program
 
-subroutine my_sub()
-  use type_bound_generic_member_access_03_m, only: called
-  called = .true.
-end subroutine
+  call br%run(1, my_sub_with_arg, 42)
+  if (.not. called_with_arg) error stop 2
+
+contains
+  subroutine my_sub_with_arg(x)
+    integer, intent(in) :: x
+    if (x == 42) called_with_arg = .true.
+  end subroutine
+  subroutine my_sub()
+    use type_bound_generic_member_access_03_m, only: called
+    called = .true.
+  end subroutine
+end program
--- a/src/libasr/asr_utils.cpp
+++ b/src/libasr/asr_utils.cpp
@@ -2906,21 +2906,6 @@ bool is_op_overloaded(ASR::logicalbinopType op, std::string& intrinsic_op_name,
 template <typename T>
 bool argument_types_match(const Vec<ASR::call_arg_t>& args,
         const T &sub) {
-    auto is_implicit_procedure = [](ASR::symbol_t* sym) -> bool {
-        if (!sym) return false;
-        if (ASR::is_a<ASR::Variable_t>(*sym)) {
-            ASR::Variable_t* v = ASR::down_cast<ASR::Variable_t>(sym);
-            if (v->m_type_declaration) {
-                ASR::symbol_t* type_decl = ASRUtils::symbol_get_past_external(v->m_type_declaration);
-                if (ASR::is_a<ASR::Function_t>(*type_decl)) {
-                    std::string decl_name = ASR::down_cast<ASR::Function_t>(type_decl)->m_name;
-                    return (decl_name.find("__") == 0 && decl_name.find("_iface_implicit") != std::string::npos);
-                }
-            }
-        }
-        return false;
-    };
-
     if (args.size() <= sub.n_args) {
         size_t i;
         for (i = 0; i < args.size(); i++) {
@@ -2941,6 +2926,28 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
                 }
                 ASR::ttype_t *arg1 = ASRUtils::expr_type(args[i].m_value);
                 ASR::ttype_t *arg2 = v->m_type;
+
+                // Check if formal argument is an implicit interface procedure (e.g. procedure() :: f)
+                // If it is, it accepts any explicit interface procedure as actual argument.
+                if (v->m_type_declaration) {
+                    ASR::symbol_t* type_decl = ASRUtils::symbol_get_past_external(v->m_type_declaration);
+                    if (ASR::is_a<ASR::Function_t>(*type_decl)) {
+                        std::string decl_name = ASR::down_cast<ASR::Function_t>(type_decl)->m_name;
+                        if (decl_name.find("__") == 0 && decl_name.find("_iface_implicit") != std::string::npos) {
+                            continue;
+                        }
+                    }
+                }
+
+                // Check if actual argument is an implicit interface procedure (e.g. external :: f)
+                // If it is, it is compatible with any explicit interface formal argument.
+                if (ASR::is_a<ASR::FunctionType_t>(*arg1)) {
+                    ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
+                    if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
+                        continue;
+                    }
+                }
+
                 ASR::symbol_t* s1 = ASRUtils::symbol_get_past_external(ASRUtils::get_struct_sym_from_struct_expr(args[i].m_value));
                 ASR::symbol_t* s2 = nullptr;
                 ASR::ttype_t* arg2_ext = ASRUtils::extract_type(arg2);
@@ -2971,28 +2978,7 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
                         }
                     }
                 } else {
-                    // Implicit interface procedures (procedure() with no arg types)
-                    // are compatible with any explicit interface — skip types_equal.
-                    ASR::symbol_t* actual_sym = nullptr;
-                    if (args[i].m_value) {
-                        if (ASR::is_a<ASR::Var_t>(*args[i].m_value)) {
-                            actual_sym = ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::Var_t>(args[i].m_value)->m_v);
-                        } else if (ASR::is_a<ASR::StructInstanceMember_t>(*args[i].m_value)) {
-                            actual_sym = ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::StructInstanceMember_t>(args[i].m_value)->m_m);
-                        }
-                    }
-
-                    bool arg_is_implicit_procedure = is_implicit_procedure(sub_arg_sym) || is_implicit_procedure(actual_sym);
-                    
-                    if (!arg_is_implicit_procedure && ASR::is_a<ASR::FunctionType_t>(*arg1)) {
-                        ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
-                        if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
-                            arg_is_implicit_procedure = true;
-                        }
-                    }
-
-                    if (!arg_is_implicit_procedure &&
-                        !types_equal(arg1, arg2, args[i].m_value, sub.m_args[i], !ASRUtils::get_FunctionType(sub)->m_elemental)) {
+                    if (!types_equal(arg1, arg2, args[i].m_value, sub.m_args[i], !ASRUtils::get_FunctionType(sub)->m_elemental)) {
                         return false;
                     }
                 }
@@ -3006,33 +2992,19 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
                 ASR::ttype_t *arg1 = ASRUtils::expr_type(args[i].m_value);
                 ASR::ttype_t *arg2 = f->m_function_signature;
 
-                // Check if actual argument is an implicit interface procedure
-                // (e.g. procedure() dummy passed down).
-                // Implicit interfaces are compatible with any explicit interface
-                // - actual compatibility is checked at call site.
-                ASR::symbol_t* actual_sym = nullptr;
-                if (args[i].m_value) {
-                    if (ASR::is_a<ASR::Var_t>(*args[i].m_value)) {
-                        actual_sym = ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::Var_t>(args[i].m_value)->m_v);
-                    } else if (ASR::is_a<ASR::StructInstanceMember_t>(*args[i].m_value)) {
-                        actual_sym = ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::StructInstanceMember_t>(args[i].m_value)->m_m);
-                    }
-                }
-                bool arg1_is_implicit = is_implicit_procedure(actual_sym);
-
-                if (!arg1_is_implicit && ASR::is_a<ASR::FunctionType_t>(*arg1)) {
+                // Check if actual argument is an implicit interface procedure (e.g. external :: f)
+                // If it is, it is compatible with any explicit interface formal argument.
+                if (ASR::is_a<ASR::FunctionType_t>(*arg1)) {
                     ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
                     if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
-                        arg1_is_implicit = true;
+                        continue;
                     }
                 }
 
-                if (!arg1_is_implicit) {
-                    Allocator al(512);
-                    if (!types_equal(arg1, arg2, args[i].m_value,
-                        ASRUtils::get_expr_from_sym(al, &f->base), false)) {
-                        return false;
-                    }
+                Allocator al(512);
+                if (!types_equal(arg1, arg2, args[i].m_value,
+                    ASRUtils::get_expr_from_sym(al, &f->base), false)) {
+                    return false;
                 }
             }
         }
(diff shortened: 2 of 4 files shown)


src/libasr/asr_utils.cpp
```diff
@@ -2955,8 +2955,26 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
...
+                        ASR::is_a<ASR::FunctionType_t>(*arg2)) {
+                        ASR::FunctionType_t* arg1_func_type =
+                            ASR::down_cast<ASR::FunctionType_t>(arg1);
+                        ASR::FunctionType_t* arg2_func_type =
+                            ASR::down_cast<ASR::FunctionType_t>(arg2);
+                        arg_is_implicit_procedure =
+                            (arg1_func_type->n_arg_types == 0 &&
+                             arg1_func_type->m_deftype == ASR::deftypeType::Interface) ||
+                            (arg2_func_type->n_arg_types == 0 &&
+                             arg2_func_type->m_deftype == ASR::deftypeType::Interface);
```
kmr-srbh: I am not sure this is correct. This will match and allow `compute_pi` as argument for function interfaces declared with 0 arguments, like this:

```fortran
interface
  subroutine my_interface()
  end subroutine
end interface
```

with argument `f` declared as `procedure(my_interface)` inside `run_a1`.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -22747,7 +22747,24 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                llvm_symtab_fn_names[fn_name] = h;
+                                llvm_symtab_fn[h] = llvm::Function::Create(
+                                    function_type, llvm::Function::ExternalLinkage,
+                                    fn_name, module.get());
+                            } else {
+                                uint32_t old_h = llvm_symtab_fn_names[fn_name];
+                                llvm_symtab_fn[h] = llvm_symtab_fn[old_h];
+                            }
+                        }
                         tmp = llvm_symtab_fn[h];
```
kmr-srbh, line 22761: Won't `tmp` contain a null value here when it was not found inside `llvm_symtab_fn`?

src/libasr/asr_utils.cpp
Jatinagarwal24: Ok I will check it.
Thanks!

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: Actually,` tmp `won't be null. If `h` is not found in `llvm_symtab_fn`, it enters the `if` block where `llvm_symtab_fn[h]` is explicitly assigned a value (either via `llvm::Function::Create` or copied from `llvm_symtab_fn[old_h]`). So by the time we hit `tmp = llvm_symtab_fn[h]`;, it is guaranteed to be populated with a valid pointer.

src/libasr/asr_utils.cpp
Jatinagarwal24: Hi @kmr-srbh you are absolutely right.
Because we were checking both the formal and actual arguments with the n_arg_types == 0 && m_deftype == Interface heuristic, it incorrectly allowed explicit 0-argument formal parameters (like procedure(my_interface)) to bypass type-checking.

I've pushed a fix that resolves this. Now, for formal dummy arguments, it rigorously checks if the m_type_declaration specifically contains the _iface_implicit suffix (which is how the parser safely flags true procedure() dummies).

We still keep the 0 arguments + Interface heuristic exclusively for actual arguments, because LFortran currently doesn't easily distinguish things like external :: my_f from explicit 0-arg interfaces in the ASR, but restricting the bypass on the formal side completely protects against the compute_pi mismatch you highlighted.

Thanks for pointing this out!

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -22747,7 +22747,24 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                    llvm_symtab_fn_names.end()) {
+                                llvm_symtab_fn_names[fn_name] = h;
+                                llvm_symtab_fn[h] = llvm::Function::Create(
+                                    function_type, llvm::Function::ExternalLinkage,
+                                    fn_name, module.get());
+                            } else {
+                                uint32_t old_h = llvm_symtab_fn_names[fn_name];
+                                llvm_symtab_fn[h] = llvm_symtab_fn[old_h];
+                            }
+                        }
```
kmr-srbh: I think we would not need this since this is checking the passed argument, which is an already declared function.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -22756,6 +22773,28 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            orig_arg_intent != ASR::intentType::InOut &&
+                            orig_arg_intent != ASR::intentType::Out &&
+                            !ASRUtils::is_pointer(orig_arg->m_type)) {
+                        llvm::Type* expected_type = llvm_utils->get_type_from_ttype_t_util(
+                            ASRUtils::EXPR(ASR::make_Var_t(al, orig_arg->base.base.loc, &orig_arg->base)),
+                            orig_arg->m_type, module.get());
+                        if (tmp->getType() != expected_type) {
+                            tmp = builder->CreateBitCast(tmp, expected_type);
+                        }
+                    }
```
kmr-srbh: Why do we need this? I think we might already be handling this elsewhere since the implementation works without it.

src/libasr/asr_utils.cpp
```diff
@@ -2969,14 +3007,24 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
...
+                        actual_sym = ASRUtils::symbol_get_past_external(ASR::down_cast<ASR::StructInstanceMember_t>(args[i].m_value)->m_m);
+                    }
+                }
+                bool arg1_is_implicit = is_implicit_procedure(actual_sym);
+
+                if (!arg1_is_implicit && ASR::is_a<ASR::FunctionType_t>(*arg1)) {
+                    ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
+                    if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
+                        arg1_is_implicit = true;
+                    }
```
kmr-srbh, line 3001: I think we might not need this for this fix since `procedure() :: f` is `ASR::Variable_t`.

src/libasr/asr_utils.cpp
```diff
@@ -2955,8 +2970,31 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
...
+                        ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
+                        if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
+                            arg_is_implicit_procedure = true;
+                        }
+                    }
+
+                    if (!arg_is_implicit_procedure &&
+                        !types_equal(arg1, arg2, args[i].m_value, sub.m_args[i], !ASRUtils::get_FunctionType(sub)->m_elemental)) {
+                        return false;
+                    }
```
kmr-srbh: Here, let's just check if the logic written inside the `is_implicit_procedure` lambda holds for variable `v` (declared above in the if block), and return `true`, else continue with the general type check. We don't need the lambda, just the implementation inside it. That way, we handle this similar to how we handle `class(*)` arguments during type-checking.

integration_tests/type_bound_generic_member_access_03.f90
```diff
@@ -0,0 +1,33 @@
...
+    end subroutine
+  end interface
+  call br%run(1, my_sub)
+  if (.not. called) error stop 1
+end program
+
+subroutine my_sub()
+  use type_bound_generic_member_access_03_m, only: called
+  called = .true.
+end subroutine
```
kmr-srbh: Please declare the subroutine `my_sub` inside the program and use.

kmr-srbh (changes requested):

integration_tests/type_bound_generic_member_access_03.f90
```diff
@@ -0,0 +1,33 @@
...
+
+program type_bound_generic_member_access_03
+  use type_bound_generic_member_access_03_m
+  implicit none
+  type(runner) :: br
+  interface
+    subroutine my_sub()
+    end subroutine
+  end interface
+  call br%run(1, my_sub)
```
kmr-srbh, line 33: Please also add a test case where we pass a subroutine that takes some argument, and call it where the `procedure() :: ...` argument is declared.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: While the function is declared in the ASR, its `llvm::Function*` is sometimes NOT generated ahead of time. Inside `instantiate_function `(around line 7936), LFortran explicitly skips pre-declaring Interface functions if they contain class(*) arguments (like `compute_pi` in expr2.f90). Because it is skipped, it is missing from `llvm_symtab_fn`. When we try to pass it as an actual argument later, llvm_symtab_fn.find(h) fails, resulting in a direct `ICE (AssertFailed)`. The lazy block catches this and creates the LLVM function just-in-time.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: If we remove the bitcast block, LLVM module verification immediately fails with Call parameter type does not match function signature!. In expr2.f90, run_a1 expects procedure() :: f, which translates in LLVM to a function taking zero arguments: void ()*. We are passing compute_pi, which translates to void (%"~unlimited_polymorphic_type"*)*. Because the interface is implicit, our semantic pass allows this mismatch to bypass. However, LLVM is strictly typed and refuses to compile if you pass a mismatched function pointer. We must apply the bitcast in the codegen to cast @compute_pi to void ()* so that LLVM accepts it.

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: Here is the PR that introduced the change you mentioned - https://github.com/lfortran/lfortran/pull/8013. The change you are mentioning is related to submodules, where `ASRUtils::get_FunctionType(x)->m_module` is false. I am still not sure we need the above changes. Could you please try checking with an example using submodules? Please refer `integration_tests/separate_compilation_19a.f90` and `integration_tests/separate_compilation_19b.f90`.

integration_tests/type_bound_generic_member_access_03.f90
Jatinagarwal24: Done

integration_tests/type_bound_generic_member_access_03.f90
Jatinagarwal24: Done

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: I see, the MRE for this is an example where it fails. But there, the interface `compute_pi` is declared inside a `program`, so I think it is strange. Also, why is class type special cased here? @AnshMehta1 could you please help?

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: I checked this. Looking closely, we notice that when there is no `call f...` statement inside the subroutine declaring `procedure() :: f`, the signature of the subroutine contains `f` typed as `void ()* %f`. But with a call to f (with an arg), the subroutine signature changes to match the passed argument, like - `void (i32*)* %f`. With `class(*)` there is a different ASR verify pass error.

Please try to investigate this since it is something we have already been doing and related to your PR.

Here is the example:

```fortran
module test_mod
   implicit none

   type :: runner
   contains
      procedure :: run_a1
      generic :: run => run_a1
   end type

contains

   subroutine run_a1(this, a1, f)
      class(*), intent(in) :: a1
      procedure() :: f

      call f(2)
   end subroutine

end module

program main
   use test_mod
   implicit none
   type(runner) :: br

   ! This call fails to match the generic procedure interface 'run' because of the 'procedure()' argument
   call br%run(100000000000_8, compute_pi)

contains

   subroutine compute_pi(n)
      integer, intent(in) :: n
      print *, "Computing pi with n =", n
   end subroutine

end program
```

```console
 Computing pi with n =           2
(lf) saurabhkumar@Mac lfortran % lfortran test.f90 --implicit-interface
Computing pi with n = 2
```

Jatinagarwal24: Thanks for pointing that out . I removed that .

Jatinagarwal24:
Hi @kmr-srbh I think now all the comments have resolved and we can run exhaustive.

kmr-srbh (approved):
The changes look good, but we need to understand the LLVM function instantiation issue before merging. I am not sure we would want to do it like that.

Jatinagarwal24:
Thanks @kmr-srbh for reviewing it.
Let @certik also review it.

certik:
> The changes look good, but we need to understand the LLVM function instantiation issue before merging. I am not sure we would want to do it like that.

@Jatinagarwal24 can you please address this?

Jatinagarwal24:
> > The changes look good, but we need to understand the LLVM function instantiation issue before merging. I am not sure we would want to do it like that.
> 
> @Jatinagarwal24 can you please address this?

The lazy instantiation block and bitcast block were only needed for the `class(*) case (where compute_pi takes class(*), intent(in) :: n) `which can be seen for this mre 
```
module test_mod
    implicit none

    type :: runner
    contains
        procedure :: run_a1
        generic :: run => run_a1
    end type

contains

    subroutine run_a1(this, a1, f)
        class(*), intent(in) :: a1
        procedure() :: f
    end subroutine

end module

program main
    use test_mod
    implicit none
    type(runner) :: br

    interface
        subroutine compute_pi(n)
            class(*), intent(in) :: n
        end subroutine
    end interface

    ! This call fails to match the generic procedure interface 'run' because of the 'procedure()' argument
    call br%run(100000000000_8, compute_pi)

end program
```
 For the core fix which is currently required for benchmark.f package — internal procedures passed through procedure() arguments neither block is needed because:

No lazy block needed: Internal procedures (declared in contains) are correctly pre-declared by instantiate_function and are always present in llvm_symtab_fn. The skip logic in instantiate_function (from PR #8013, line ~7932) only triggers for Interface functions with class(*) args where m_module is false (the submodule case) — not for internal procedures.

No bitcast needed: Internal procedures have their LLVM types fully known at declaration time, so there's no type mismatch to bitcast.

 The class(*) case is a separate, deeper issue and therefore we can open a separate issue to resolve that.

src/libasr/codegen/asr_to_llvm.cpp
kmr-srbh: To unblock us on this, I would say, please make a small change to the implementation where we skip this instantiation, to work correctly for the case of submodules. It currently checks if the `m_module` field is `false`, but that field is also `false` for `program`, so I think it is incorrect.

Jatinagarwal24: Done.
Thanks !

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -7932,13 +7927,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                            debug_current_scope = debug_current_scope_copy;
+                // Skip pre-declaration only for submodule interfaces with class(*) args,
+                // not for program-scoped interfaces (where m_module is also false).
+                bool parent_is_program = false;
+                if (x.m_symtab->parent && x.m_symtab->parent->asr_owner &&
+                        x.m_symtab->parent->asr_owner->type == ASR::asrType::symbol) {
+                    parent_is_program = ASR::is_a<ASR::Program_t>(
+                        *ASR::down_cast<ASR::symbol_t>(x.m_symtab->parent->asr_owner));
+                }
+                if (!parent_is_program) {
```
kmr-srbh, line 7938: This is fine.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -22752,12 +22754,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
-                        LCOMPILERS_ASSERT(llvm_symtab_fn.find(h) != llvm_symtab_fn.end());
-                        tmp = llvm_symtab_fn[h];
-                        LCOMPILERS_ASSERT(tmp != nullptr)
+                    } else if (ASRUtils::get_FunctionType(fn)->m_deftype == ASR::deftypeType::Interface) {
+                        if (llvm_symtab_fn_arg.find(h) != llvm_symtab_fn_arg.end()) {
+                            tmp = llvm_symtab_fn_arg[h];
+                        } else {
+                            LCOMPILERS_ASSERT(llvm_symtab_fn.find(h) != llvm_symtab_fn.end());
+                            tmp = llvm_symtab_fn[h];
+                        }
```
kmr-srbh: This can now be reverted I think?

integration_tests/type_bound_generic_member_access_03.f90
```diff
@@ -0,0 +1,48 @@
...
+program type_bound_generic_member_access_03
+  use type_bound_generic_member_access_03_m
+  implicit none
+  type(runner) :: br
+  logical :: called_with_arg = .false.
+
+  call br%run(1, my_sub)
+  if (.not. called) error stop 1
+
+  call br%run(1, my_sub_with_arg, 42)
```
kmr-srbh, line 36: Now that we have fixed the function interface instantiation issue with `class(*)` inside program, please add the MRE test case here to check it.

kmr-srbh (approved):

integration_tests/type_bound_generic_member_access_03.f90
Jatinagarwal24: The instantiate_function fix correctly pre-declares compute_pi now, but adding the class(*) MRE to the integration test still hits an LLVM type mismatch:
```
void (%"~unlimited_polymorphic_type"*)* @compute_pi
 void ()*  call void @__module_type_bound_generic_member_access_03_m_run_a1(%type_bound_generic_member_access_03_m.runner.3_class* %3, %"~unlimited_polymorphic_type"* %2, void (%"~unlimited_polymorphic_type"*)* @compute_pi)

````
`run_a1` declares `procedure() :: f `which lowers to `void ()*`, but `compute_pi `takes `class(*)`,` intent(in) :: n` which lowers to `void (%"~unlimited_polymorphic_type"*)`. This mismatch needs either a bitcast at the call site or a fix in semantic analysis to propagate the actual procedure's type through the implicit interface.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: Done.
Thanks !

integration_tests/type_bound_generic_member_access_03.f90
Jatinagarwal24: Hi @kmr-srbh Maybe you can once pull my pr and see what i am saying to better understand or correct me if i am wrong.

kmr-srbh: If it is not required for the package, we could skip it, else we need to understand this https://github.com/lfortran/lfortran/pull/11625#discussion_r3299035523.

Jatinagarwal24: Yes it is not required.So we can skip and move forward.

Jatinagarwal24:
Hi @kmr-srbh We can merge this now.

kmr-srbh (approved):

kmr-srbh:
Please also update the PR description.

Issue #7629, opened 2025-06-08
feat: support polymorphic arrays


kmr-srbh:
There are a few known bugs that will have to be fixed to support this, like:

- improvement in handling of allocation in `handle_allocated()` function in `asr_to_llvm.cpp`
- assignment of `type` variable to `class`
- handle `select type(a => arr(i))` correctly

We need to study more to know what else may be required.

An example using a wrapper type

```fortran
module main_mod
   implicit none

   type :: Animal
   contains
      procedure :: speak => animal_speak
   end type Animal

   type, extends(Animal) :: Dog
   contains
      procedure :: speak => dog_speak
   end type Dog

   type, extends(Animal) :: Cat
   contains
      procedure :: speak => cat_speak
   end type Cat

contains

   subroutine animal_speak(self)
      class(Animal), intent(in) :: self
      print *, "Some animal sound."
   end subroutine

   subroutine dog_speak(self)
      class(Dog), intent(in) :: self
      print *, "Woof!"
   end subroutine

   subroutine cat_speak(self)
      class(Cat), intent(in) :: self
      print *, "Meow!"
   end subroutine

end module main_mod

program main
    use main_mod
    implicit none

    integer :: i, n

    type :: AnimalWrapper
        class(Animal), allocatable :: ref
    end type AnimalWrapper

    type(AnimalWrapper), allocatable :: arr(:)

    n = 3

    allocate(arr(n))

    allocate(Cat :: arr(1)%ref)
    allocate(Dog :: arr(2)%ref)

    do i = 1, n
        call arr(i)%ref%speak()
    end do

    do i = 1, n
        select type(a => arr(i)%ref)
         type is(Dog)
            print *, "It's a dog!"
         type is(Cat)
            print *, "It's a cat!"
        class default
            print *, "Some animal."
        end select
    end do

end program main
```
```console
(base) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
 Meow!
 Woof!
 Some animal sound.
 It's a cat!
 It's a dog!
 Some animal.
(base) saurabhkumar@Mac lfortran % lfortran test.f90
Traceback (most recent call last):
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2726
    LCompilers::print_stack_on_segfault();
  File "/Users/saurabhkumar/OpenSource/lfortran/src/bin/lfortran.cpp", line 2653
[...]

PR #11635, opened 2026-05-25
fix : polymorphic method dispatch and allocation checks on unallocated receivers

Jatinagarwal24:
Towards #7629
* The Bug:
LFortran previously generated unconditional runtime allocation errors when 
an unallocated polymorphic allocatable variable was passed as a type-bound 
receiver. Bypassing this allocation check alone was insufficient because 
unallocated polymorphic objects possess a null VTable pointer, leading 
to runtime segmentation faults during dynamic dispatch. Furthermore, 
previous attempts to bypass VTable lookups caused code-generation crashes 
or linking errors (`undefined reference`) when dealing with deferred 
methods on abstract types.

* The Solution:
1. Allocation Check Bypass: Refined the condition in `visit_SubroutineCall` 
   to correctly identify type-bound receivers (`x.m_dt == alloc_check_expr`). 
   This safely bypasses the strict non-allocatable bounds check without 
   relying on brittle, post-resolution AST type assumptions.

2. F2008 Dynamic Type Fallback: Implemented dynamic type resolution logic in 
   both `visit_RuntimePolymorphicSubroutineCall` and `visit_RuntimePolymorphicFunctionCall`. 
   Because the dynamic type of an unallocated polymorphic object defaults to 
   its declared type, the code now generates an LLVM conditional branch to 
   safely dispatch to the statically declared binding when a null receiver 
   is detected.

3. Deferred Method Safety: Added a strict guard (`!class_proc->m_is_deferred`) 
   before attempting to dispatch to a declared binding. If a receiver is 
   unallocated and the method is deferred (abstract), the compiler now 
   emits a clean runtime abort instead of generating dead calls to abstract 
   interfaces that fail during linkage.

4. LLVM SSA Form Preservation: Introduced `Alloca` memory slots in 
   polymorphic function branches to safely store and load return values 
   across the divergent `vtable_dispatch` and `declared_binding` basic 
   blocks, satisfying LLVM's Static Single Assignment constraints.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3117,6 +3117,7 @@ RUN(NAME class_146 LABELS gfortran llvm)
 RUN(NAME class_147 LABELS gfortran llvm)
 RUN(NAME class_148 LABELS gfortran llvm)
 RUN(NAME class_149 LABELS gfortran llvm)
+RUN(NAME class_150 LABELS gfortran llvm)
 
 RUN(NAME class_procedure_args_01 LABELS gfortran llvm)
 
--- a/integration_tests/class_150.f90
+++ b/integration_tests/class_150.f90
@@ -1,0 +1,66 @@
+module class_150_mod
+   implicit none
+
+   integer :: animal_calls = 0
+   integer :: dog_calls = 0
+   integer :: cat_calls = 0
+
+   type :: animal_t
+   contains
+      procedure :: speak => animal_speak
+   end type
+
+   type, extends(animal_t) :: dog_t
+   contains
+      procedure :: speak => dog_speak
+   end type
+
+   type, extends(animal_t) :: cat_t
+   contains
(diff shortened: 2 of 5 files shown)


Jatinagarwal24:
Hi @kmr-srbh pls once review this.

kmr-srbh:
I think we support polymorphic arrays already. The MRE works after adding a required allocate statement for `arr(3)%ref`:

```console
(base) saurabhkumar@Mac lfortran % gfortran test.f90 && ./a.out
 Meow!
 Woof!
 Some animal sound.
 It's a cat!
 It's a dog!
 Some animal.
(base) saurabhkumar@Mac lfortran % lfortran test.f90           
Meow!
Woof!
Some animal sound.
It's a cat!
It's a dog!
Some animal.
```
Without the allocate statement, we get a runtime error, which I think is correct. Is there a different bug that the PR helps fix?

PR #11636, opened 2026-05-25
fix: avoid struct temp for ArrayItem of non-pointer/non-allocatable struct args

HarshitaKalani:


src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -1467,7 +1467,11 @@ class ArgSimplifier: public ASR::CallReplacerOnExpressionsVisitor<ArgSimplifier>
                 x_m_args_vec.push_back(al, array_var_temporary);
             } else if( ASRUtils::is_struct(*ASRUtils::expr_type(x_m_args[i])) &&
                        !ASR::is_a<ASR::Var_t>(
-                            *ASRUtils::get_past_array_physical_cast(x_m_args[i])) ) {
+                            *ASRUtils::get_past_array_physical_cast(x_m_args[i])) &&
+                       !(ASR::is_a<ASR::ArrayItem_t>(
+                            *ASRUtils::get_past_array_physical_cast(x_m_args[i])) &&
+                         !ASRUtils::is_allocatable(ASRUtils::expr_type(x_m_args[i])) &&
+                         !ASRUtils::is_pointer(ASRUtils::expr_type(x_m_args[i]))) ) {
```
assem2002: Now that we're here;
Please Use `is_directly_addressable_array_expr` instead of this check.
It will cover more cases in the future.
also change the name of the function : 
`is_directly_addressable_array_expr` -> `is_directly_addressable_expr`

src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -1268,12 +1268,17 @@ ASR::expr_t* create_and_allocate_temporary_variable_for_struct(
...
     value = ASRUtils::get_past_array_physical_cast(value);
     switch( value->type ) {
         case ASR::exprType::Var: {
             return true;
         }
+        case ASR::exprType::ArrayItem: {
+            return !ASRUtils::is_array(ASRUtils::expr_type(value)) &&
+                   !ASRUtils::is_allocatable(ASRUtils::expr_type(value)) &&
+                   !ASRUtils::is_pointer(ASRUtils::expr_type(value));
+        }
```
assem2002, line 1281:
```suggestion
        case ASR::exprType::Var:
        case ASR::exprType::ArrayItem: {
            return true;
        }
```

assem2002: tested it locally. didn't work.

PR #11638, opened 2026-05-25
fix: handle scalar field argument in unpack intrinsic

### LINKED ISSUE #11637: ICE: `unpack` intrinsic segfaults with scalar `field` argument
## MRE

```fortran
program test_unpack_scalar
    implicit none
    integer :: a(5)
    logical :: mask(5)
    mask = [.true., .false., .true., .false., .true.]
    a = unpack([100, 200, 300], mask, 0)
    print *, a
end program
```

## Output

```text
$ lfortran a.f90
  File "src/bin/lfortran.cpp", line 2997, in main
  File "src/bin/lfortran.cpp", line 2910, in main_app
  File "src/bin/lfortran.cpp", line 1222, in compile_src_to_object_file
  File "src/lfortran/fortran_evaluator.cpp", line 302, in get_asr2
  File "src/lfortran/fortran_evaluator.cpp", line 324, in get_asr3
  File "src/lfortran/semantics/ast_to_asr.cpp", line 114, in ast_to_asr
  File "src/lfortran/semantics/ast_body_visitor.cpp", line 273
  ...
  File "src/libasr/pass/intrinsic_array_function_registry.h", line 6574, in create_Unpack
    extract_value(field_dims[0].m_length, field_dim);
```

## Expected

The `field` argument of `unpack` can be scalar per the Fortran standard. When scalar, it broadcasts to fill positions where mask is false.

Expected output: `100 0 200 0 300`

Note: `unpack` with an array-valued `field` works correctly. The crash only occurs when `field` is a scalar.

ambikeesshh:
fixes #11637

The `unpack` intrinsic segfaulted when the `field` argument was a scalar
(e.g. `unpack([100, 200, 300], mask, 0)`). The `create_Unpack` function
unconditionally dereferenced `field_dims[0]` to extract dimension info,
which is invalid for scalar arguments (rank-0).

Added a `field_is_scalar` check to skip dimension extraction and
validation when `field` is rank-0. The scalar value broadcasts to the
mask shape at runtime. Compile-time evaluation via `eval_Unpack` is
skipped for scalar field (returns nullptr), letting the runtime path
handle it.

## Test plan
- [x] New integration test `unpack_scalar_field_01.f90` covering integer, real, logical scalar field and 2D mask
- [x] Existing `intrinsics_149` (array-field unpack) still passes

### DIFF
--- a/src/libasr/pass/intrinsic_array_function_registry.h
+++ b/src/libasr/pass/intrinsic_array_function_registry.h
@@ -6411,6 +6411,11 @@ namespace Unpack {
         ASR::ttype_t *type_mask = ASRUtils::type_get_past_pointer(ASRUtils::type_get_past_allocatable(expr_type(mask)));
         ASR::ttype_t* type_a = ASRUtils::type_get_past_array(type_vector);
 
+        ASR::ttype_t *type_field = ASRUtils::type_get_past_pointer(ASRUtils::type_get_past_allocatable(expr_type(field)));
+        if (!ASRUtils::is_array(type_field)) {
+            return nullptr;
+        }
+
         int kind = ASRUtils::extract_kind_from_ttype_t(type_a);
         int dim_mask = ASRUtils::get_fixed_size_of_array(type_mask);
         int dim_vector = ASRUtils::get_fixed_size_of_array(type_vector);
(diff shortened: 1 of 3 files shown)


ambikeesshh:
@certik PTAL

ambikeesshh:
@HarshitaKalani Could you add the Tests::Run-Exhaustive label to trigger full CI? 
 Thanks!

HarshitaKalani (approved):

ambikeesshh:
@assem2002 PTAL

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -6411,6 +6411,11 @@ namespace Unpack {
         ASR::ttype_t *type_mask = ASRUtils::type_get_past_pointer(ASRUtils::type_get_past_allocatable(expr_type(mask)));
         ASR::ttype_t* type_a = ASRUtils::type_get_past_array(type_vector);
 
+        ASR::ttype_t *type_field = ASRUtils::type_get_past_pointer(ASRUtils::type_get_past_allocatable(expr_type(field)));
+        if (!ASRUtils::is_array(type_field)) {
+            return nullptr;
+        }
+
```
assem2002, line 6418: I opened an issue #11775 for this.
returning nullptr here would cause that test to fail.
Can you please take a look at that?

assem2002 (approved):
LGTM. Please address my comment

src/libasr/pass/intrinsic_array_function_registry.h
ambikeesshh: will take a look, thanks for catching

Issue #11428, opened 2026-05-10
Reproducibility problem: symbol name changes


amckinstry:
Hi

I'm having a problem getting reproducible builds with lfortran:

https://reproduce.debian.net/amd64/api/v1/builds/249037/artifacts/629879/diffoscope
Bit-reproducible builds are now required for Debian.

It comes down to the symbol names in the libraries changing, eg.
```
 +_lcompilers_len_trim_str1_281472862009456_YWIo0BwbpRGo4IKPeSljP6 in src_assert_m.F90.o
````
Where does the `281472862009456_YWIo0BwbpRGo4IKPeSljP6` come from, and can it be made constant?

Thanks
Alastair

**Comment by assem2002** [comment]
That ID is being added due to using compiler flag --> `--separate-compilation` only .
I think LFortran shouldn't use unique ID, instead, mark its own library as `weak_odr` (Use one function and avoid redefinition conflicts at link).
I think we can fix that.

**Comment by amckinstry** [comment]
Any idea of a fix or patch?
Sorry, this is a blocker for my testing on Debian.

**Comment by assem2002** [comment]
I'm putting this on my TODO this week. Let's see how we can handle this.

**Comment by assem2002** [comment]
An idea :
Since the problem is about `separate-compilation` (not compiling the whole code as a single unit), We can use the hash of the file_path string and use that as unique identifier. 
currently we're using this :
```cpp
std::string get_unique_ID() {
    static std::random_device dev;
    static std::mt19937 rng(dev());
    std::uniform_int_distribution<int> dist(0, 61);
    const std::string v =
        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
    std::string res;
    for (int i = 0; i < 22; i++) {
        res += v[dist(rng)];
    }
    return res;
}
```

assem2002:
Hi @amckinstry. I'm sorry for the delay.
 https://github.com/lfortran/lfortran/pull/11812 is attempting to depend on source_file_path to give symbols a unique ID, that should make things consistent instead of just random id.
***
I think the reproducible-build tests happen to have consistent file path names, right?
Will we need to path mapping?
*** 
Also I checked minimal example from the `assert` package by hand, there seems to be no diff with mentioned PR.

amckinstry:
Hi @assem2002 : yes reproducible-builds have consistent pathnames. 
I've tested the patch against 0.63 and it looks good. Thanks.

PR #11643, opened 2026-05-26
fix duplicate _lcompilers_* symbols in separate compilation

### LINKED ISSUE #11642: Duplicate __lcompilers_* helper symbols across separately compiled objects
Current `main` emits duplicate helper symbols into separately compiled object files. Linking those object files together then fails with duplicate `__lcompilers_*` symbols.

Version tested:

```text
Platform: macOS ARM
LLVM: 21.1.0
```

Minimal reproducer:

`a.f90`:

```fortran
subroutine a(x)
  real, intent(inout) :: x
  x = sin(x)
end subroutine
```

`b.f90`:

```fortran
subroutine b(x)
  real, intent(inout) :: x
  x = sin(x)
end subroutine
```

`main.f90`:

```fortran
program main
  real :: x
  x = 1.0
  call a(x)
  call b(x)
  print *, x
end program
```

Commands:

```sh
lfortran -c --linker=gcc -o a.o a.f90
lfortran -c --linker=gcc -o b.o b.f90
lfortran -c --implicit-interface --linker=gcc -o main.o main.f90
lfortran --linker=gcc -o main main.o a.o b.o
```

Observed:

```text
duplicate symbol '__lcompilers_sin_f32' in:
    a.o
    b.o
ld: 1 duplicate symbols
```

Expected:

Separately compiled object files should link successfully. Compiler-generated helper symbols should not collide across translation units.

gfortran accepts and links this reproducer.

ambikeesshh:
Fixes #11642

_lcompilers_* intrinsic helpers were emitted with ExternalLinkage,
causing duplicate symbol errors when linking separately compiled
objects that use the same intrinsic (e.g. sin).

Use LinkOnceODRLinkage so the linker merges identical copies across
translation units.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7949,8 +7949,11 @@ public:
             );
             if (llvm_symtab_fn_names.find(fn_name) == llvm_symtab_fn_names.end()) {
                 llvm_symtab_fn_names[fn_name] = h;
-                F = llvm::Function::Create(function_type,
-                    llvm::Function::ExternalLinkage, fn_name, module.get());
+                auto linkage = llvm::Function::ExternalLinkage;
+                if (fn_name.rfind("_lcompilers_", 0) == 0 || fn_name.rfind("__lcompilers", 0) == 0) {
+                    linkage = llvm::Function::LinkOnceODRLinkage;
+                }
+                F = llvm::Function::Create(function_type, linkage, fn_name, module.get());
             } else {
                 uint32_t old_h = llvm_symtab_fn_names[fn_name];
                 F = llvm_symtab_fn[old_h];
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -7950,9 +7950,6 @@ public:
             if (llvm_symtab_fn_names.find(fn_name) == llvm_symtab_fn_names.end()) {
                 llvm_symtab_fn_names[fn_name] = h;
                 auto linkage = llvm::Function::ExternalLinkage;
-                if (fn_name.rfind("_lcompilers_", 0) == 0 || fn_name.rfind("__lcompilers", 0) == 0) {
-                    linkage = llvm::Function::LinkOnceODRLinkage;
-                }
                 F = llvm::Function::Create(function_type, linkage, fn_name, module.get());
             } else {
                 uint32_t old_h = llvm_symtab_fn_names[fn_name];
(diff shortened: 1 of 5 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8454,8 +8454,15 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 for (size_t i=0; i<x.n_body; i++) {
                     this->visit_stmt(*x.m_body[i]);
                 }
-                
+
                 define_function_exit(x);
+
+                std::string fn_name = x.m_name;
+                if (compiler_options.separate_compilation &&
```
konradha, line 8461: I think

```
if (compiler_options.separate_compilation && x.m_name.starts_with("_lcompilers")) {
    builder->GetInsertBlock()->getParent()->setLinkage(
        llvm::Function::LinkOnceODRLinkage);
}
```

is a bit leaner

ambikeesshh: thanks..!

starts_with would miss __lcompilers prefixed functions though (double underscore). we need both same as unique_symbols.cpp:184

ambikeesshh:
hey @assem2002 can you PTAL at this PR and review it
thanks

assem2002 (commented):
Apparently the test works in main.
`--separate-compilation` handles creating object file properly.
***
the real issue addressed here is that we create redundant fortran-intrinsic functions, The correct approach for real is to use `LinkOnceODRLinkage` like what we're doing here in this PR.

The problem I have in mind is : 
- in this PR, we're setting all functions starting with `_lcompilers` as linkOnceODRLinkage`
- What if we have 2 helper functions that starting with `_lcompilers` and have the same names but different signature in 2 different source files?
- What if we happen to be creating intrinsic function like `sin_f32` and `sin_f64` but we miss the type id? at link time they link against single function while they're different.

assem2002:
I think we create different functions for the same intrinsic (e.g. `unpack`) based on the arguments + return_type.
because we use `_lcompilers` prefix extensively, I think we might need to flag any of lfortran intrinsic-library functions with unique identifier (even maybe have flag within `ASR::function` symbol) + differentiate between the same function-name-but-different-arguments-or-return by making sure we suffix the name with those arguments and return type.

***
That way, we have unique identifier for the intrinsic functions + unique-consistent name for the function name without relying on `get_unique_name` which isn't consistent.

assem2002:
related to https://github.com/lfortran/lfortran/issues/11428#issuecomment-4541284559

PR #11646, opened 2026-05-26
fix : support derived type component array sections in intrinsics and pointer association

### LINKED ISSUE #11639: MAXLOC rejects derived-type component array section with mask
Current `main` rejects a `MAXLOC` call where the array argument is a component of an array of derived-type instances and the mask has the same shape.

Version tested:

```text
Platform: macOS ARM
LLVM: 21.1.0
```

Minimal reproducer:

```fortran
module maxloc_component_mask
  implicit none

  type :: point
     real(8) :: x(2)
  end type

  type(point), allocatable :: top(:)

contains

  subroutine get_segment(r, idx)
    real(8), intent(in) :: r
    integer, intent(out) :: idx
    integer :: loc(1)

    loc = maxloc(top(:)%x(1), top(:)%x(1) < r)
    idx = loc(1)
  end subroutine

end module

program main
  use maxloc_component_mask
  implicit none
  integer :: idx

  allocate(top(3))
  top(1)%x = [0.0d0, 0.0d0]
  top(2)%x = [1.0d0, 0.0d0]
  top(3)%x = [2.0d0, 0.0d0]

  call get_segment(1.5d0, idx)
  print *, idx
end program
```

Command:

```sh
```

Observed:

```text
semantic error: `array` argument of `MaxLoc` must be an array
  --> repro.f90:17:11
   |
17 |     loc = maxloc(top(:)%x(1), top(:)%x(1) < r)
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Expected:

This should compile. `top(:)%x(1)` is an array expression over the array of derived-type instances, and the mask has the same shape.

gfortran accepts this reproducer.

Jatinagarwal24:
fixes #11639

- Semantics (ast_common_visitor.h):
  When indexing a member of a derived-type array (e.g. `top(:)%x(1)`), return an `ArraySection_t` node instead of an `ArrayItem_t` node. Ensure that the dimensions and physical type of the base struct array are preserved and duplicated into the final type. This resolves the semantic error where component array sections were not recognized as arrays (e.g., `argument of MaxLoc must be an array`).

- Codegen (asr_to_llvm.cpp):
  - In `visit_StructInstanceMember`, load intermediate pointer-to-pointer temporaries representing descriptors (like dynamic `ArraySection` nodes) before performing GEP. This fixes GEP target type mismatches under LLVM 15+ (opaque pointers).
  - In `handle_array_section_association_to_pointer`, add support for derived-type component array sections. Since LFortran uses element-based strides, scale the stride of the target descriptor by:
    (base_stride * base_element_size) / target_element_size.
    This resolves runtime out-of-bounds indexing and incorrect mask evaluations during array slice operations.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1784,6 +1784,7 @@ RUN(NAME intrinsics_463 LABELS llvm) # get_command (optional keyword arguments) 
 RUN(NAME intrinsics_464 LABELS gfortran llvm) # cshift with array shift
 RUN(NAME intrinsics_465 LABELS gfortran llvm) # index() must not read past declared length of deferred-length string
 RUN(NAME intrinsics_466 LABELS gfortran llvm) # eoshift with array-valued shift
+RUN(NAME intrinsics_467 LABELS gfortran llvm) # maxloc over derived-type array component section
 RUN(NAME intrinsics_len_trim_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME la_constants LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # LAPACK constants
--- a/integration_tests/intrinsics_467.f90
+++ b/integration_tests/intrinsics_467.f90
@@ -1,0 +1,18 @@
+program intrinsics_467
+    implicit none
+
+    type :: point
+        real(8) :: x(2)
+    end type
+
+    type(point), allocatable :: top(:)
+    integer :: loc(1)
+
+    allocate(top(3))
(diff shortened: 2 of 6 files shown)


kmr-srbh:
The added integration test fails with `--fast`.

Jatinagarwal24:
Thanks @kmr-srbh for the review.
I am aware of the fact that the test which i added fails with fast.
But i need some suggestions so as to how to fix that. I tried to fix it but was facing some issue with that.

Jatinagarwal24:
Hi @assem2002 I need your advice  like how to fix this test added so that it dosent fail with `fast` flag.

PR #11649, opened 2026-05-26
fix :ICE and runtime allocation errors for character(len=*) function returns

### LINKED ISSUE #11644: ICE on external character function with deferred length
## MRE

```fortran id="mre73184"
character(len=*) function make_token()
  make_token = "A"
end function make_token

subroutine demo(n)
  integer :: n
  character(len=n), external :: make_token

  print *, make_token()
end subroutine demo

program main
  call demo(3)
end program main
```

## Output

```text id="out73184"

```

## gfortran Behavior

```text id="out73185"
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran t.f90 && ./a.out

 A
```

Jatinagarwal24:
fixes #11644 

This commit resolves two major issues related to `character(len=*)` function 
returns and implicit interfaces: a segmentation fault during semantic analysis 
and a runtime error where strings were not allocated at the call site.

1. Fixed SIGSEGV in Semantic Analysis: 
   In `ast_symboltable_visitor.cpp`, `character(len=*)` functions evaluate to 
   `AssumedLength` strings, which lack a valid `m_value` for their length expression. 
   The compiler previously crashed trying to extract a length expression from a null 
   pointer. We added checks to correctly handle `AssumedLength` (a_len == -3) and 
   `DeferredLength` (a_len == -1) without requiring constant length evaluations.

2. Fixed "Non-allocatable string isn't allocated" Runtime Error:
   When `subroutine_from_function` converts a function call into a subroutine call, 
   the caller must allocate a temporary variable for the return slot. However, the 
   implicit interface generation in `ast_common_visitor.h` (`create_external_function`)
   was incorrectly overwriting the explicitly provided local interface (e.g. `character(len=n)`)
   with the global `AssumedLength` signature. By giving precedence to the locally
   `determined_type`, the compiler now correctly preserves `ExpressionLength`. This 
   allows `allocate_stmt_needed_for_return_slot` to recognize the string and emit 
   the required `Allocate` statement at the call site.

3. Updated ASR Transformation Logic:
   In `subroutine_from_function.cpp`, added logic to correctly map internal `AssumedLength`
   return variables to `DeferredLength` + `Allocatable` descriptors to ensure 
   compatibility with the caller's dynamically allocated return slot.


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2030,14 +2030,38 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
 
             if (return_type->m_kind != nullptr) {
                 if (return_type->n_kind == 1) {
-                    visit_expr(*return_type->m_kind->m_value);
-                    len_expr = ASRUtils::EXPR(tmp);
-                    if (return_type->m_type == AST::decl_typeType::TypeCharacter) {
-                        a_len = ASRUtils::extract_len<SemanticAbort>(len_expr, x.base.base.loc, diag);
-                        a_kind = ASRUtils::extract_kind_from_ttype_t(ASRUtils::expr_type(len_expr));
+                    if (return_type->m_kind->m_type == AST::kind_item_typeType::Star) {
```
assem2002, line 2033: This piece of the function (handling characters) needs to be refactored.
We can do that later.

assem2002 (approved):
LGTM. Thanks!

PR #11650, opened 2026-05-26
Fix ICE and unallocated errors for nested array-returning intrinsics

### LINKED ISSUE #11647: ICE when using spread inside index intrinsic call
## MRE

```fortran id="mre59211"
module spread_index
contains
  function hit(deferred_text) result(found)
    logical :: found
    character(len=1), parameter :: tokens(1) = ["A"]

    found = any(index(tokens, spread(deferred_text, 1, 1)) /= 0)
  end function hit
end module spread_index

program demo
  use spread_index
  print *, hit("A")
end program demo
```

## Output

```text id="out59211"

ASR verify pass error:
FunctionCall::m_name _lcompilers_spread must be returning a non-void value.

Internal Compiler Error:
LCompilersException: Verify failed in the pass: intrinsic_function
```

## gfortran Behavior

```text id="out59212"
(lf) opixdown@Atharvs-MacBook-Air lfortran % gfortran t.f90 -c

(compiles successfully)
```

Jatinagarwal24:
fixes #11647

This commit resolves an Internal Compiler Error (ICE) and subsequent
runtime "unallocated argument" errors that occurred when array-returning
intrinsic functions (like `spread`) were used as nested arguments inside
other expressions (e.g., `any(index(tokens, spread(...)))`).

The `intrinsic_function` pass handles array intrinsics by replacing the
`FunctionCall` with a `SubroutineCall` that takes an explicit output
variable. Previously, this pass incorrectly assumed that the tracked 
`result_var_` (the parent assignment target) could always be used as the
output buffer, and discarded the parent statement after substitution.

To fix this, the following changes were made in `intrinsic_function.cpp`:

1.  **Type Compatibility Validation**: Added strict type checking
    (`ASRUtils::types_equal`) between the target `result_var_` and the
    intrinsic's return type. If they are incompatible, the compiler now
    correctly falls back to generating an intermediate temporary variable.

2.  **Valid Assumed-Length String Conversions**: When creating the
    temporary variable for string operations, `AssumedLength` traits
    are properly converted to `DeferredLength` to satisfy ASR verifier
    constraints for local variables.

3.  **Dynamic Runtime Allocations**: Emits robust `Allocate` ASR
    structures for the temporary variables. Missing static bounds are now
    dynamically resolved at runtime using `ArraySize` on the input
    arguments, and deferred string lengths are resolved via `StringLen`.

4.  **AST Statement Preservation**: Introduced a `was_temp_created`
    tracker to `ReplaceFunctionCallReturningArray`. The statement 
    transformer (`transform_stmts`) now preserves the modified parent 
    assignment statement alongside the injected `Allocate` and 
    `SubroutineCall` statements, ensuring the AST remains intact.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2799,6 +2799,7 @@ RUN(NAME string_115 LABELS gfortran llvm)
 
 RUN(NAME substring_read LABELS gfortran llvm)
 
+RUN(NAME nested_array_intrinsics_01 LABELS gfortran llvm)
 RUN(NAME nested_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME nested_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/nested_array_intrinsics_01.f90
+++ b/integration_tests/nested_array_intrinsics_01.f90
@@ -1,0 +1,21 @@
+program test_nested_array_intrinsics
+    implicit none
+
+    character(len=1), parameter :: tokens(1) = ["A"]
+
+    ! Test spread inside index and any with character arrays
+    if (.not. hit("A")) error stop 1
+    if (hit("B")) error stop 2
+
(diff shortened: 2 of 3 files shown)


Jatinagarwal24:
Hi @certik I mistakely auto merged this pr so I reverted that back in pr #11651.
Sorry for that.

PR #11652, opened 2026-05-26
fix: ICE and unallocated errors for nested array-returning intrinsics

### LINKED ISSUE #11647: ICE when using spread inside index intrinsic call

Jatinagarwal24:
fixes #11647

Root Cause: create_Spread was propagating the source argument's AssumedLength string type directly into the IntrinsicArrayFunction's return type via duplicate_type. character(*) (AssumedLength) is only valid for dummy arguments — it is invalid ASR for a function return type. This caused downstream passes to fail when trying to create temporaries or allocate variables for the result.

Fix: In create_Spread (intrinsic_array_function_registry.h), when the source is an assumed-length string, convert the return type's element type from AssumedLength to ExpressionLength with len = StringLen(source). This produces a semantically correct return type that the existing array_struct_temporary pass can handle properly when creating temporaries for nested intrinsic calls.

In instantiate_Spread, remap the string length expression to reference the generated function's own source parameter (instead of the caller's variable) to avoid cross-scope symbol table references.

No changes to intrinsic_function.cpp are needed — the array_struct_temporary pass already unwraps nested intrinsics into temp = intrinsicCall statements with proper types.

### DIFF
--- a/src/libasr/pass/intrinsic_function.cpp
+++ b/src/libasr/pass/intrinsic_function.cpp
@@ -186,13 +186,15 @@ class ReplaceFunctionCallReturningArray: public ASR::BaseExprReplacer<ReplaceFun
     public:
     ASR::expr_t* result_var_; // Declared in array_struct_temporary
     SymbolTable* current_scope;
+    bool was_temp_created;
 
     ReplaceFunctionCallReturningArray(Allocator& al_, Vec<ASR::stmt_t*>& pass_result_,
     std::map<ASR::symbol_t*, ASRUtils::IntrinsicArrayFunctions>& func2intrinsicid_) :
     al(al_), pass_result(pass_result_), result_counter(0),
     func2intrinsicid(func2intrinsicid_),
     result_var_(nullptr),
-    current_scope(nullptr) {}
+    current_scope(nullptr),
+    was_temp_created(false) {}
 
     // Not called from anywhere but kept for future use.
     // Especially if we don't find alternatives to allocatables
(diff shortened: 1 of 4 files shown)


src/libasr/pass/intrinsic_function.cpp
```diff
@@ -293,10 +295,114 @@ class ReplaceFunctionCallReturningArray: public ASR::BaseExprReplacer<ReplaceFun
...
+            ASR::dimension_t* ret_dims = nullptr;
+            size_t ret_n_dims = ASRUtils::extract_dimensions_from_ttype(
+                ASRUtils::type_get_past_allocatable(x->m_type), ret_dims);
+            ASR::alloc_arg_t alloc_arg;
+            // For deferred-length strings, provide the string length
+            // from the source argument (first arg of spread).
+            if( ASR::is_a<ASR::String_t>(*element_type) &&
+                ASR::down_cast<ASR::String_t>(element_type)->m_len_kind ==
+                    ASR::string_length_kindType::DeferredLength &&
+                x->n_args > 0 && x->m_args[0].m_value ) {
```
assem2002: that's hard coded to handle the case of spread as I think.

assem2002 (commented):
Based on what I understand, The PR is attempting to create proper temporary to handle a functioncall (that should be converted into subroutinecall) that returns an assumed-length-string array (`character(*) :: arr(1)`).

It's attempting to create a deferred-allocatable-character-array-variable and allocate it properly, which is correct temporary, but still why does a functionCall return an assume-length? that's not a valid fortran nor ASR

***
Here is my take on this :
- `Create_spread` should create proper return type -- the return of the function shouldn't return assumed-length rather the length should be an expressionLength + the length is `stringLen(source)`
- I think the current functionality can handle this type and create/allocate proper temp variable for it. and the current changes that handles assumed-length won't be needed.
- [Maybe not now] Assert in ASR_verify that we shouldn't have any functionCall with `assumed_length` string return. because we don't know how to handle that on ASR (because it's invalid semantic)

Do you think we'll still face problem

Jatinagarwal24:
Thanks @assem2002 for the review.
This commit c47b123e1836622b1015a9e1be477f0a264081e7 resolves your comments by the following way
fix it by create_Spread returns ExpressionLength instead of AssumedLength for string source

what was the old fix doing : create_Spread was propagating AssumedLength from the source's type into
the return type of the IntrinsicArrayFunction. AssumedLength is only
valid for dummy arguments and is invalid ASR for a function return.

Fix by converting AssumedLength to ExpressionLength with
StringLen(source) as the length expression in create_Spread. Also fix
instantiate_Spread to remap the string length to reference the
function's own source parameter instead of the caller's variable.

Remove the downstream workaround in intrinsic_function.cpp that was
converting AssumedLength to DeferredLength when creating temp variables,
as it is no longer needed.

src/libasr/pass/intrinsic_function.cpp
```diff
@@ -293,10 +295,82 @@ class ReplaceFunctionCallReturningArray: public ASR::BaseExprReplacer<ReplaceFun
...
         }
 
-        if( !this->result_var_ ) {
-            // No result variable (e.g., parameter initialization);
-            // the value is compile-time constant, skip replacement.
-            return;
+        ASR::expr_t* effective_result_var = this->result_var_;
+        // Check if result_var_ is usable: it must be set AND its type must
+        // be compatible (i.e., it must be an array of the same element type).
+        // When the intrinsic is nested (e.g., index(tokens, spread(...))),
```
assem2002: I think these changes won't be needed if we correct how `spread` creates the return, what do you think?

Jatinagarwal24: I think these changes handle the nested intrinsic call case (e.g., index(tokens, spread(...))), which is independent of the return type fix. When spread is nested, result_var_ points to the outer assignment target (a logical scalar), which is incompatible with spread's string array return. The type check detects this mismatch and creates a proper temporary. Without this, the old code either crashes (LCOMPILERS_ASSERT) or passes the wrong variable as the result output. This is needed regardless of how spread creates its return type.

Jatinagarwal24: The original code before this PR was 
```
if( !this->result_var_ ) {
    // No result variable (e.g., parameter initialization);
    // the value is compile-time constant, skip replacement.
    return;    // ← just skips, leaving FunctionCall unprocessed
}
// ...
LCOMPILERS_ASSERT(this->result_var_)   // ← crashes if null
new_arg.m_value = this->result_var_;   // ← uses outer target directly
```
For `any(index(tokens, spread(deferred_text, 1, 1)) /= 0)`:

`result_var_ `points to the outer assignment target (a logical scalar)
spread returns a string array — completely incompatible
Without the type check, the outer logical target would be passed as spread's output → wrong types, crash
Even with the correct ExpressionLength return type from `create_Spread`, the nested call problem remains — you still need a temp variable because the outer target has a different type.

assem2002: I think `array_struct_temporary` unwrapps this nesting into `temp = intrinsicFunctionCall` statements.
LHS should be there with proper type. May you check?

Jatinagarwal24: Thanks @assem2002 You are correct we dont need changes in `intrinsic_function.cpp`.I have reverted that.

Jatinagarwal24:
Hi @assem2002 I think this is good to merge.

assem2002 (approved):
LGTM. 
I think a deferredLength would've been cleaner as functionReturn with funcParam is messy.

PR #11655, opened 2026-05-27
fix: character kind propagation for string declarations, literals, and slices

### LINKED ISSUE #11648: Wrong type mismatch for kind=4 character assignment
## MRE

```fortran
program p
    type :: t
        character(len=:, kind=4), pointer :: s
    end type

    type(t) :: x

    allocate(character(len=5, kind=4) :: x%s)

    x%s(4:4) = achar(int(z'0031'), kind=4)

    deallocate(x%s)
end program
```

## Output

```text

 --> t.f90:8:5
  |
8 |     x%s(4:4) = achar(int(z'0031'), kind=4)
  |     ^^^^^^^^   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch (string and string)
```

## gfortran Behavior

```text

(compiles successfully)
```

Jatinagarwal24:
fixes #11648

**Bug:**
The ASR construction phase was silently ignoring the `kind` parameter for `character` declarations, string slices (`StringSection`), and string literal constants, hardcoding them to default to `kind=1`. When assigning or comparing a `kind=4` string (e.g., `achar(..., kind=4)`) to a `kind=4` string slice `x%s(4:4)` or another `kind=4` literal like `4_"1"`, the compiler incorrectly threw a semantic error ("Type mismatch in assignment" or "Operands of comparison operator are of different types") because the generated ASR nodes were improperly marked as `kind=1`.

**Fix:**
- **Character Declarations:** Updated `determine_char_len_and_kind` in `ast_common_visitor.h` to explicitly evaluate `kind_item->m_value` and extract it into `str->m_kind` instead of leaving it at the default.
- **String Sections/Slices:** Modified string section logic (e.g., in `visit_ArrayRef` and `visit_Substring`) to dynamically extract and propagate the character kind from the root base variable (`ASRUtils::extract_kind_from_ttype_t(root_v_type)`) rather than hardcoding `1` during `ASR::make_String_t` creation.
- **String Literals:** Updated `visit_String` to parse and apply the AST's `x.m_kind` (e.g., for `4_"1"`) to the ASR `StringConstant` node, utilizing the same symbol resolution and constant evaluation logic used for integer bounds.


HarshitaKalani (approved):

Jatinagarwal24:
Hi @HarshitaKalani I think this is good to merge.
Thanks!

PR #11667, opened 2026-05-27
fix: ICE in struct implied do loop IO expansion

### LINKED ISSUE #11659: READ implied-do over derived-type array elements still hits visit_ImpliedDoLoop
Current `main` still fails on `READ` with an implied-do loop over whole derived-type array elements.

Tested on current `main` commit `1c791f9fc`.

This is different from reading into a derived-type **component** inside the implied-do loop. The reproducer below reads whole derived-type records.

Minimal reproducer:

```fortran
program main
  implicit none

  type :: rec_t
     real(8) :: theta, r, phi
  end type

  type(rec_t), allocatable :: recs(:)
  integer :: i, n

  n = 2
  allocate(recs(n))

  read(*, *) (recs(i), i = 1, n)
  print *, recs(1)%theta, recs(2)%phi
end program
```

Command:

```sh
```

Observed:

```text
LCompilersException: visit_ImpliedDoLoop() not implemented
```

Expected:

This should compile. At runtime, input like:

```text
1 2 3 4 5 6
```

should populate the two records and print `1` and `6`.

`gfortran` compiles and runs this reproducer successfully.

Jatinagarwal24:
fixes #11659

The struct member expansion for IO statements (read/write) only
triggered when a_values_vec contained exactly one value. This caused
two bugs:

1. Variable-bound implied do loops like `read(*, *) (recs(i), i=1,n)`
   produced a single ImpliedDoLoop with StructType. The struct expansion
   then wrapped StructInstanceMember around the entire ImpliedDoLoop
   (producing the invalid `StructInstanceMember(ImpliedDoLoop(...))`),
   which hit the unimplemented visit_ImpliedDoLoop in LLVM codegen.

2. Constant-bound implied do loops like `read(*, *) (recs(i), i=1,3)`
   were expanded into multiple struct-typed ArrayItems (recs(1), recs(2),
   recs(3)). Since size > 1, struct expansion was skipped entirely,
   causing an assertion failure in codegen when it tried to generate a
   read function for StructType.

Fix: generalize the struct expansion to iterate over all IO values,
expanding struct members for each one. For ImpliedDoLoop values, struct
members are expanded inside the loop values (producing correct ASR like
`ImpliedDoLoop(values=[recs(i)%x, recs(i)%y, ...])`) instead of
wrapping outside.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3323,6 +3323,7 @@ RUN(NAME implied_do_loops36 LABELS gfortran llvm)
 
 RUN(NAME implied_do_loops37 LABELS gfortran llvm)
 RUN(NAME implied_do_loops38 LABELS gfortran llvm EXTRA_ARGS --use-loop-variable-after-loop)
+RUN(NAME implied_do_loops39 LABELS gfortran llvm)
 
 RUN(NAME minpack_01 LABELS gfortran llvm_rtlib EXTRAFILES
         minpack_01_module.f90 minpack_01_func.f90)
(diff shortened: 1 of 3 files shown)


HarshitaKalani (approved):

Jatinagarwal24:
Hi @HarshitaKalani I think this is good to merge.
Thanks !

PR #11669, opened 2026-05-27
Fix segfault in implicit interface resolution for derived-type array sections

### LINKED ISSUE #11668: --implicit-interface segfault on generic procedure call with derived-type component array section
Current  main segfaults when `--implicit-interface` is enabled and a generic procedure is called with a derived-type component array section actual argument.

Tested on current `main` commit `1c791f9fc`.

Minimal reproducer:

```fortran
module monotonicmod
  implicit none
  interface monotonic
     module procedure monotonic_dble
  end interface monotonic
contains
  function monotonic_dble(x, n)
    logical :: monotonic_dble
    integer, intent(in) :: n
    real(8), dimension(n), intent(in) :: x
    monotonic_dble = .true.
    if (n == 1) return
    if (any(x(2:n) <= x(1:n-1))) monotonic_dble = .false.
  end function monotonic_dble
end module monotonicmod

module m
  use monotonicmod
  implicit none
  type :: bdrypt
     real(8) :: x(2)
  end type
  type(bdrypt), allocatable :: top(:)
contains
  subroutine test()
    if (.not. monotonic(top%x(1), size(top))) then
       print *, 'not monotonic'
    end if
  end subroutine
end module

program main
  use m
  implicit none
  allocate(top(2))
  top(1)%x = [1.0d0, 0.0d0]
  top(2)%x = [2.0d0, 0.0d0]
  call test()
end program
```

Command:

```sh
lfortran --implicit-interface --linker=gcc -c repro.f90
```

Observed:

```text
```

Stack trace ends in:

```text
.../src/lfortran/semantics/ast_common_visitor.h
ASR::symbol_t *f2 = ASRUtils::symbol_get_past_external(fn);
```

`gfortran` compiles and runs this reproducer successfully.

srinjoy933:
Fixes #11668

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2115,7 +2115,7 @@ RUN(NAME implicit_interface_35 LABELS gfortran llvm EXTRA_ARGS --implicit-typing
 RUN(NAME implicit_interface_36 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_37 LABELS gfortran llvm EXTRA_ARGS --implicit-interface --legacy-array-sections)
 RUN(NAME implicit_interface_38 LABELS gfortran llvm EXTRA_ARGS --implicit-interface EXTRAFILES implicit_interface_38b.f90)
-RUN(NAME implicit_interface_39 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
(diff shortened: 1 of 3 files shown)


konradha:
@srinjoy933 did this get fix3d in another PR?

srinjoy933:
@konradha Not actually ,but I was having some problem with test fails in this pr, so I closed this pr,and will create a new pr shortly to solve this issue .Thank you ,for reporting the issue.

PR #11677, opened 2026-05-29
Fixes for the recet f128 initial implementation

certik:


src/libasr/pass/intrinsic_functions.h
```diff
@@ -4718,13 +4725,13 @@ namespace Real {
             ASR::ttype_t* t1, Vec<ASR::expr_t*> &args, diag::Diagnostics& diag) {
         if (ASR::is_a<ASR::IntegerConstant_t>(*args[0])) {
             double i = ASR::down_cast<ASR::IntegerConstant_t>(ASRUtils::expr_value(args[0]))->m_n;
-            return make_ConstantWithType(make_RealConstant_t, i, t1, loc);
+            return ASRUtils::make_RealConstant_util(al, loc, i, t1);
         } else if (ASR::is_a<ASR::RealConstant_t>(*args[0])) {
             ASR::RealConstant_t *r = ASR::down_cast<ASR::RealConstant_t>(ASRUtils::expr_value(args[0]));
-            return make_ConstantWithType(make_RealConstant_t, r->m_r, t1, loc);
+            return ASRUtils::make_RealConstant_util(al, loc, r->m_r, t1);
```
opixdown, line 4731: From what I understand, after this PR `RealConstant_t::m_r` means two different things depending on kind:

- kind 4/8: actual numeric double value
- kind 16: pointer

it works for real 4,8 but for real 16 input ,it passes the pointer as if it were the numeric value.

I tried this small case locally:
```
program main
    real(16), parameter :: a = 1.25_16
    real(8), parameter :: b = real(a, kind=8)
    print *, b
end program
```

the output is garbage value. ASR shows `b` folded to `0.000000` instead of `1.25`.

 we should handle kind 16 here by using `ASRUtils::real_constant_get_r16(r)` instead of reading `r->m_r` directly

src/libasr/asr_utils.cpp
```diff
@@ -3141,34 +3142,50 @@ ASR::asr_t* make_Cast_t_value(Allocator &al, const Location &a_loc,
...
             int64_t int_value = ASR::down_cast<ASR::IntegerConstant_t>(ASRUtils::expr_value(a_arg))->m_n;
-            // skip constant folding for kind=16 (pointer-in-double not safe here)
             int dest_kind = ASRUtils::extract_kind_from_ttype_t(a_type);
             if (dest_kind != 16) {
                 value = ASR::down_cast<ASR::expr_t>(ASR::make_RealConstant_t(al, a_loc, (double)int_value, a_type));
+            } else {
+                // int -> real(16) is always exact for int64; build the binary128
+                // bytes directly on the arena and pack them into m_r.
+                uint8_t* bytes = (uint8_t*)al.alloc(16);
+                lf_float128 v = lf_f128_from_double((double)int_value);
```
opixdown, line 3184: this converts int64 -> double -> real(16). For large int64 values above 2^53, the double step can lose precision before we create the real(16)

opixdown:
CI is still segfaulting while compiling `real128_arith_07` on Ubuntu It  pass on macOS arm64, but still segfault like it was earlier. we need to fix it

src/libasr/pass/intrinsic_functions.h
certik: > From what I understand, after this PR RealConstant_t::m_r means two different things depending on kind:

That was implemented by your PR, correct? I am just trying to fix some leftover bugs.

PR #11679, opened 2026-05-29
feat: add legacy F77 aliases iargc and getarg

### LINKED ISSUE #10764: Support legacy intrinsics iargc and getarg
While looking at benchmarking code `FASTA` I noticed that it uses the legacy intrinsics iargc and getarg. These func have been renamed in the standard to `command_argument_count` and `get_command_argument,` but such legacy usage still exists in benchmark codes. so we should remap these internally in LFortran for better compatibility.

## **MRE**
```
program test_getarg_iargc
    implicit none
    integer :: n, i
    character(len=100) :: arg

    n = iargc()
    print *, "Number of arguments:", n
    do i = 0, n
        call getarg(i, arg)
        print *, "arg(", i, ") = '", trim(arg), "'"
    end do

end program test_getarg_iargc
```
### **Output with LFortran**
```
base) opixdown@Atharvs-MacBook-Air lfortran % lfortran t.f90 --continue-compilation
semantic error: Function 'iargc' not found (not user defined nor intrinsic)
 --> t.f90:8:9
  |
8 |     n = iargc()
  |         ^^^^^^^ 

semantic error: Function 'getarg' not found (not user defined nor intrinsic)
  --> t.f90:14:9
   |
14 |         call getarg(i, arg)
   |         ^^^^^^^^^^^^^^^^^^^ 
```
### **Output with GFortran**
```
 Number of arguments:           0
 arg(           0 ) = './a.out'
```

Zaneham:
Map iargc() to command_argument_count and getarg(pos, value) to get_command_argument by registering the legacy names in the intrinsic registries. Both follow the existing lshift -> shiftl alias pattern, sharing the modern create_* functions verbatim.

Real F77 benchmark code (FASTA and similar) uses these spellings; without the aliases the user has to rewrite source to compile under LFortran while gfortran/ifort accept them as-is.

Adds integration_tests/cmd_04.f90 that asserts the legacy and modern intrinsics agree for every command-line position from 0 to iargc(), labelled gfortran + llvm.

Fixes #10764

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3311,6 +3311,7 @@ RUN(NAME boz_02 LABELS llvm llvm_wasm llvm_wasm_emcc) #remove gfortran label, as
 RUN(NAME cmd_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME cmd_02 LABELS gfortran llvm)
 RUN(NAME cmd_03 LABELS gfortran llvm) # get_command_argument loop until empty
+RUN(NAME cmd_04 LABELS gfortran llvm) # iargc/getarg legacy F77 aliases
 
 RUN(NAME flip_sign LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME div_to_mul LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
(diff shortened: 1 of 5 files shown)


certik (approved):
Thanks for implementing it. I think it looks good, the old CI passed. I rebased it to check the CI again, if it passes, we can merge it.

PR #11680, opened 2026-05-29
fix: handle elemental intrinsics on allocatable arrays

### LINKED ISSUE #11678: ICE for nint on allocatable 2d array in array constructor
## MRE

```fortran
program ice_test
  implicit none

  real(kind=8), allocatable :: data(:,:)
  logical :: result_all

  allocate(data(2,3))

  data(1,:) = [1.0d0, -5.0d0, 3.0d+2]
  data(2,:) = [4.0d0, 2.0d0, 6.0d0]

  result_all = all([nint(data)] == [1, 4, -5, 2, 300, 6])

  print *, result_all
end program ice_test
```

## output

```text

Internal Compiler Error:
LCompilersException: Expression type, Allocatable[f64[:,:]] not yet supported
```

## gfortran behavior

```text

 T
```

Jatinagarwal24:
fixes #11678
Skip ArrayConstructor with array function calls in implied_do_loop pass

The implied_do_loop pass was incorrectly expanding ArrayConstructor
nodes containing function calls (FunctionCall, IntrinsicElementalFunction,
IntrinsicArrayFunction) returning arrays into do-loops. This generated
invalid code like `lbound(Nint(data), ...)` and caused functions to be
called multiple times instead of once.

The fix skips these cases in the implied_do_loop pass, letting the
downstream passes (array_struct_temporary + array_op) handle them
correctly. Also reverts the previous workaround in intrinsic_function.cpp
(replace_ArrayItem override) that was treating the symptom rather than
the root cause.

### DIFF
--- a/src/libasr/pass/implied_do_loops.cpp
+++ b/src/libasr/pass/implied_do_loops.cpp
@@ -577,6 +577,30 @@ class ReplaceArrayConstant: public ASR::BaseExprReplacer<ReplaceArrayConstant> {
     }
 
    void replace_ArrayConstructor(ASR::ArrayConstructor_t* x) {
+        // Skip ArrayConstructor replacement when any argument is a function
+        // call (FunctionCall or IntrinsicElementalFunction) returning an array.
+        // The implied_do_loop pass incorrectly expands these into do-loops,
+        // generating invalid code (e.g. lbound(Nint(data), ...) or calling
+        // functions multiple times). The downstream passes
+        // (array_struct_temporary + array_op) handle these cases properly.
+        for( size_t i = 0; i < x->n_args; i++ ) {
+            ASR::expr_t* arg = x->m_args[i];
+            if( arg == nullptr ) continue;
+            bool is_array_func_call = false;
+            if( ASR::is_a<ASR::FunctionCall_t>(*arg) &&
+                ASRUtils::is_array(ASRUtils::expr_type(arg)) ) {
+                is_array_func_call = true;
+            } else if( ASR::is_a<ASR::IntrinsicElementalFunction_t>(*arg) &&
+                       ASRUtils::is_array(ASRUtils::expr_type(arg)) ) {
+                is_array_func_call = true;
+            } else if( ASR::is_a<ASR::IntrinsicArrayFunction_t>(*arg) &&
+                       ASRUtils::is_array(ASRUtils::expr_type(arg)) ) {
+                is_array_func_call = true;
+            }
+            if( is_array_func_call ) {
+                return;
+            }
+        }
         const Location& loc = x->base.base.loc;
         ASR::expr_t* result_var_copy = result_var;
         bool is_result_var_fixed_size = false;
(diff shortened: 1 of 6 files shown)


assem2002:
Here is an MRE
```.f90
program ice_test
  implicit none

  real :: data(2,3)
  integer :: data2(6)
  data2 =  [nint(data)]
end program ice_test
```

look at the output of pass number 5 `implied_do_loop` :
```.f90
! Fortran code after applying the pass: implied_do_loops
program ice_test
implicit none
integer(4) :: __1__libasr_index_
integer(4) :: __1_t
integer(4) :: __2_t
integer(4), dimension(6) :: __libasr__created__var__0__array_constructor_
real(4), dimension(2, 3) :: data
integer(4), dimension(6) :: data2
__1__libasr_index_ = lbound(__libasr__created__var__0__array_constructor_, 1)
do __2_t = lbound(Nint(data), 2), ubound(Nint(data), 2)
    do __1_t = lbound(Nint(data), 1), ubound(Nint(data), 1)
        __libasr__created__var__0__array_constructor_(__1__libasr_index_) = Nint(data)(__1_t, __2_t)
        __1__libasr_index_ = __1__libasr_index_ + 1
    end do
end do
data2 = __libasr__created__var__0__array_constructor_
end program ice_test
```


*** 
Implied do loop happens to mess with `arrayConstructor` with functionCall inside. You can even use `--skip-pass=implied_do_loop` and the linked MRE would just work
***
another example 
```.f90
program ice_test
  implicit none

  integer :: data2(6)
  data2 =  [f()]
  contains 
  function f() result(r)
    integer :: r(2,3)
    print *, "heere"
  end function f
end program ice_test
```
You'll see multiple prints for `heere` -- while if you use`--skip-pass=implied_do_loop`  you'll have single print.

***
## What do to
apparently we're fixing the symptoms not the actual issue.
We need to avoid whatever impliedDoLoop is trying to do when it comes to ArrayConstructor with the argument being a functionCall -- the other passes (array_struct_temporary + array_op) handle those things properly.
It even weird that `implied_do_loop` pass is attempting to handle the arrayConstructor while it shouldn't be its job from the first place.

Jatinagarwal24:
Hi @assem Thanks for the review. I realized that was a bad fix i have reverted that and applied the fix you have suggested.
Thanks !

Jatinagarwal24:
Hi @assem2002 third party code failed 
image
So i am fixing that by adding test so that we dont have regression.
Thanks

Jatinagarwal24:
Hi @assem2002 I think this is good to merge now.

src/libasr/pass/implied_do_loops.cpp
```diff
@@ -577,6 +577,30 @@ class ReplaceArrayConstant: public ASR::BaseExprReplacer<ReplaceArrayConstant> {
...
+        // call (FunctionCall or IntrinsicElementalFunction) returning an array.
+        // The implied_do_loop pass incorrectly expands these into do-loops,
+        // generating invalid code (e.g. lbound(Nint(data), ...) or calling
+        // functions multiple times). The downstream passes
+        // (array_struct_temporary + array_op) handle these cases properly.
+        for( size_t i = 0; i < x->n_args; i++ ) {
+            ASR::expr_t* arg = x->m_args[i];
+            if( arg == nullptr ) continue;
+            bool is_array_func_call = false;
+            if( ASR::is_a<ASR::FunctionCall_t>(*arg) &&
```
assem2002, line 590: I'm not sure what `replace_ArrayConstructor` is doing in this pass so we can properly condition things, I think we might need some iterations over this issue.

but for now, that's fine and the comments are descriptive.

assem2002 (approved):
Looks good to me. Thanks!

PR #11684, opened 2026-05-29
fix: auto-detect fixed-form from .f/.F extension by default

### LINKED ISSUE #11653: ```--std=f23``` not using ```--fixed-form-infer```
Other compilers print 'OK' with this program but not lfortran. I think the
```--std=f23``` option should include ```--fixed-form-infer``` because this program is still valid f2023 (but obsolescent).
The program:
```
*     Comment line if fixed source form
      PRINT *,'OK'
      END
```
Lfortran version:
LFortran version: 0.63.0
Status: alpha (expected to fail on third-party codes)
Platform: Linux
LLVM: 19.1.1
LSP Version: 3.17.0
JSON_RPC Version: 2.0
```
Compile-time output using ```lfortran --std=f23```
```
syntax error: Token '*' is unexpected here
 --> fixedform.f:1:1
  |
1 | *     Comment line if fixed source form
  | ^ 
```

ambikeesshh:
fixes #11653

LFortran was not inferring fixed-form source from `.f`/`.F` file extensions unless `--fixed-form-infer` was explicitly passed. Every other major compiler (gfortran, ifort, nvfortran) auto-detects this by default.

This made `--std=f23` fail on valid fixed-form `.f` files, since fixed-form source is still standard-conforming (obsolescent, not deleted) in Fortran 2023.

**Changes:**
- Extension-based inference is now the default (matching gfortran behavior)
- `--free-form` flag added to explicitly force free-form parsing
- Removed unconditional `fixed_form=true` from `--std=legacy` (inference handles it correctly now)
- Conflict check for `--fixed-form` + `--free-form`

**Before:**
```
$ lfortran --std=f23 fixedform.f
syntax error: Token '*' is unexpected here
```

**After:**
```
$ lfortran --std=f23 fixedform.f
 OK
```


ambikeesshh:
Re-triggering CI after force-push.

certik:
Let's not do this right now. This can always be done later, but if we do this now, it will break people's builds to undo it later. So it's a one way street. I think the default should be to work with modern codes, and to use options for old codes, and one quite popular ideas it just use `.f` for modern codes instead of `.f90`, which is tying the Fortran 90 standard to it, which is not ideal.

ambikeesshh:
thanks for the context, that makes sense
@certik would the narrower fix of just enabling --fixed-form-infer when --std=f23 is passed be acceptable? that way the original issue is resolved without changing any default behavior

People using .f for modern code are unaffected unless they explicitly pass --std=f23

happy to go with whatever you think is best

PR #11686, opened 2026-05-30
Fixes for the recent f128 initial implementation

opixdown:
follow up on #11677


src/libasr/runtime/lfortran_float128.c
```diff
@@ -1894,10 +1898,45 @@ void pub(void) {                                             \
...
+    __asm__ volatile(                                        \
+        "stp x29, x30, [sp, #-16]!\n\t"                     \
+        "mov x29, sp\n\t"                                    \
+        "fmov x0, d0\n\t"                                    \
+        "mov  x1, v0.d[1]\n\t"                               \
+        "bl  " _LF_SYM(impl) "\n\t"                         \
+        "ldp x29, x30, [sp], #16\n\t"                       \
+        "ret\n\t"                                            \
+    );                                                       \
+}
```
certik, line 1929: This asm style was already in main, but I don't like it. Is there a way to do this in C directly?

certik (commented):
I am still reviewing it.

certik (approved):
I think it's fine, we can further improve it later as we go.

Issue #9687, opened 2026-01-26
Introduce dedicated ArrayTypeCast node separate from ArrayPhysicalCast


krystophny:
From PR #9591 review by @assem2002:

> Probably we should introduce some new node to cast Arrays. The physicalCast here is playing the role of actual-type cast, We should separate those concerns.

Currently `ArrayPhysicalCast` is used for both:
1. Physical layout changes (e.g., `PointerArray` → `DescriptorArray`)
2. Type narrowing (e.g., `class(*)` → `T`)

These are conceptually different operations:
- **Physical cast**: changes memory representation
- **Type cast**: changes element type interpretation

**Proposal:** Introduce `ArrayTypeCast` for element type changes, keeping `ArrayPhysicalCast` for layout changes only.

Related: #9591, #9573

PR #11687, opened 2026-05-30
fix: don't double-point bitcast for class(*) array LHS of concrete call

HarshitaKalani:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -23997,6 +23998,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         }
         llvm::Type* dest_llvm_type = llvm_utils->get_type_from_ttype_t_util(
                 dest_arg, dest_type, module.get());
+        if (ASRUtils::is_array(dest_type)) {
+            return builder->CreateBitCast(class_value, dest_llvm_type);
+        }
```
certik, line 24003: Can you explain how this works and why it's a good fix? Is this masking a bug in our ASR, or is this the right fix?

HarshitaKalani: This is not masking an ASR bug; it fixes a type consistency issue in convert_class_to_type.

get_type_from_ttype_t_util() returns different LLVM type levels for scalars and arrays by design: scalars return the base type, while arrays already return a pointer type. The previous unconditional ->getPointerTo() therefore worked for scalars but produced an extra pointer level for arrays (%T** instead of %T*).

In the class(\*), allocatable :: a(:,:) case, an ArrayPhysicalCast changes the physical type to PointerArray, bypassing the earlier descriptor-specific branch and reaching the final bitcast. Without this fix, the generated call uses %sample_t** where the callee expects %sample_t*, causing LLVM verifier failures. The guard ensures the correct pointer level is used.

assem2002: I think that's fine

assem2002: I do think that our ASR lack proper casting here. 
The issue here is probably addressed here https://github.com/lfortran/lfortran/issues/9687

assem2002 (approved):
LGTM.

Issue #9081, opened 2025-12-14
Handle unallocated struct instance member in string concat


jinangshah21:
MRE:
```fortran
program mre
    implicit none
    type :: string_t
        character(:), allocatable :: s
    end type string_t

    type(string_t) :: t
    character(:), allocatable :: result
    result = "Hello " // t%s
    print *, result
end program mre
```

LFortran Error:
```console
(base) jinang_shah@JNSLAP:~/Desktop/lfortran$ lfortran a.f90 --realloc-lhs-arrays
At 9:26 of file a.f90
Runtime error: Argument 2 of subroutine _lcompilers_stringconcat is unallocated.
```

GFortran passes and it is used in fpm codebase.

harperjf:
I think the MRE is not standard Fortran. The f2023 standard 9.2 paragraph 2 says "A reference is permitted only if the variable is defined." In the MRE program, ```t%s``` is a variable which was not yet defined when it was referenced as an operand of ```//```. Gfortran was therefore allowed to do anything with the program including running it and assuming that ```t%s``` was an empty string. So I think the lfortran was right to give an error message but that it would have been better as
```
LFORTRAN ERROR: Run-time Error : Copying from undefined RHS string.
```
because merely allocating ```t%s``` would not have given it a value.

harperjf:
If ```t%s``` is declared with a given length but not evaluated the MRE program becomes
```
program mre
    implicit none
    type :: string_t
        character(5) :: s
    end type string_t

    type(string_t) :: t
    character(:), allocatable :: result
    result = "Hello " // t%s
    print "(2A,I0)", result, ' has length ',len(result)
  end program mre
```
It is still not standard-conforming but lfortran, gfortran and ifx give no error messages and run it, with these run-time outputs:
gfortran 
```
Hello I_�U has length 11
```
ifx and lfortran 
```
Hello  has length 11
```

assem2002:
I think we should push a fix to fpm upstream handling this case -- This isn't valid fortran.

certik:
So the first code snipped in https://github.com/lfortran/lfortran/issues/9081#issue-3727827449 is not valid Fortran, because `t%s` is unallocated, so LFortran must give a runtime error in Debug mode.

[...]

PR #11688, opened 2026-05-30
fix: allow zero-length copy from NULL rhs in _lfortran_copy_str_and_pad

HarshitaKalani:


assem2002 (commented):
The test here is exercising an invalid fortran behavior.
We can't use unallocated values and depend on it.
Look discussion here https://claude.ai/share/16554e6e-0b91-4481-af23-b155a31a8970
***
If this is related to athena, Please try to push a fix upstream wrapping this in `allocated-if-else` clause.
***
similar issue https://github.com/lfortran/lfortran/issues/9081

PR #11693, opened 2026-05-31
Add CI for building JupyterLite deployment

anutosh491:
I've added the CI for building the JupyterLite deployment in the exhaustive checks file. Hope that's correct.

Finally as I've mentioned it multiple times, we would need the github pages to be enabled !
On a PR level the build section of the CI would run & the deploy section of the CI would be **skipped** as seen [here](https://github.com/lfortran/lfortran/actions/runs/26702652659/job/78698900247?pr=11693)
The deployment only happens works for commits that go into main.

- So the build section at a PR level would confirm that the build is correct considering the new changes on the PR
- And then once we take the PR in, the deploy section at the main level would just host the artifacts i.e xlfortran.js/wasm on github pages.

So we would be seeing this error on main
```
Fetching artifact metadata for "github-pages" in this workflow run
Found 1 artifact(s)
Creating Pages deployment with payload:
{
    "artifact_id": 7308734146,
    "pages_build_version": "380db30371c4abaf7b89631c043da12bed711578",
    "oidc_token": "***"
}
Error: Creating Pages deployment failed
Error: HttpError: Not Found
    at /home/runner/work/_actions/actions/deploy-pages/v4/node_modules/@octokit/request/dist-node/index.js:124:1
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at createPagesDeployment (/home/runner/work/_actions/actions/deploy-pages/v4/src/internal/api-client.js:125:1)
    at Deployment.create (/home/runner/work/_actions/actions/deploy-pages/v4/src/internal/deployment.js:74:1)
    at main (/home/runner/work/_actions/actions/deploy-pages/v4/src/index.js:30:1)
Error: Error: Failed to create deployment (status: 404) with build version 380db30371c4abaf7b89631c043da12bed711578. Request ID 3018:32C252:6113267:172AA3B2:6A1AF29D Ensure GitHub Pages has been enabled: https://github.com/lfortran/lfortran/settings/pages
```

Unless we have github pages enabled.

This is how you enable it 
- Open Settings → Pages https://github.com/lfortran/lfortran/settings/pages
- Under Build and deployment → Source, select GitHub Actions

I don't have rights to enable it.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -696,3 +696,73 @@ jobs:
           context: .
           push: ${{ github.event_name == 'push' }}
           tags: ${{ steps.image-tags.outputs.tags }}
+
+  build_xlfortran_jupyterlite:
+    name: Build xlfortran JupyterLite WASM kernel
+    runs-on: ubuntu-24.04
+    if: >-
+      ${{ github.event_name == 'push' ||
+          (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'Tests::Run-Exhaustive')) }}
+    defaults:
+      run:
+        shell: bash -l {0}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Install build0 prerequisites
+        uses: mamba-org/setup-micromamba@v2
+        with:
+          environment-file: ci/environment_linux.yml
+          init-shell: bash
+          environment-name: lf
+
+      - name: Generate parser/ASR headers (build0.sh)
+        run: ./build0.sh
+
+      - name: Install WASM build environment
+        run: micromamba create -f environment-wasm-build.yml -y
+
+      - name: Build WASM kernel
+        run: |
+          set -ex
+          micromamba create -f environment-wasm-host.yml -y \
+            --platform=emscripten-wasm32 \
+            -c https://prefix.dev/emscripten-forge-4x \
+            -c https://prefix.dev/conda-forge
+          export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-lfortran-wasm-host
+          echo "PREFIX=$PREFIX" >> $GITHUB_ENV
+          micromamba run -n xeus-lfortran-wasm-build ./wasm-build1.sh
+
+      - name: Build JupyterLite site
+        run: |
+          set -ex
+          micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge -y
+          micromamba run -n xeus-lite-host jupyter lite build \
+              --XeusAddon.prefix=${{ env.PREFIX }} \
+              --contents share/lfortran/nb/Demo1.ipynb \
+              --contents share/lfortran/nb/Demo2.ipynb \
+              --contents share/lfortran/nb/Variables.ipynb \
+              --output-dir dist
+
+      - name: Upload pages artifact
+        uses: actions/upload-pages-artifact@v4
+        with:
+          path: ./dist
+
+  deploy_jupyterlite:
+    name: Deploy JupyterLite to GitHub Pages
+    needs: build_xlfortran_jupyterlite
+    if: github.ref == 'refs/heads/main'
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4


certik:
It's skipping the deploy job, which is the one that failed in main. We should deploy it, to some testing environment, or testing pages.

anutosh491:
Hey @certik, 

Yes, that's exactly what I'm trying to convey here & the last paragraph of @mcbarton's comment on the other PR conveys too : https://github.com/lfortran/lfortran/pull/11694#issuecomment-4587648657

- Github pages does not work on a PR level cause they don't give us that provision. I think it's too expensive for github pages to host a deployment for every PR on a repo.  Even for xeus-cpp or any other xeus kernel the deployment is skipped on the PR level and only runs on a commit on main.

- The good part here is when targeting emscripten-wasm32 or simply building for wasm .... the wasm contract is that the executable is Build Platform agnostic. We would never really hit a "Works on my machine" case here. So there can't be a difference between me building a deployment for a commit locally vs a PR's cI building a deployment vs main building a deployment for the same commit. 

- The failure on main was not exactly a CI failure based on what I've implemented. It failed saying "Ensure GitHub Pages has been enabled" (shared the whole error in the PR description). So we just need to go ahead and enable github deployments in lfortran's settings. That's all that remains. 
Jupyterlite as provides a template repo for running wasm kernels. Please check the demo here https://github.com/jupyterlite/xeus-lite-demo#-visual-guide and it tells you how github pages being enabled helps.

- If we still think that having a deployment on every PR is essential we need to either leverage Read The Docs or some framework. The github maintaining Jupyter's official website does that and has a bot setup: check https://github.com/jupyter/try-jupyter/pull/82#issuecomment-4500462017 which builds the deployment at a PR level.

- We can choose to go the above route but even this involve ReadTheDocs working parallely with Github pages, so we need to Enable Github Pages to have a deployment anyways. This is just more CI eventually. Check : https://github.com/jupyter/try-jupyter/tree/main/.github/workflows

My understanding here is that having a deployment on top of main should be enough. The other PR I've made : https://github.com/lfortran/lfortran/pull/11694 is working towards a testing effort for anything being run through the wasm kernel. 

So if we want to test a PR's commit claiming to fix something 
i) either you can have a ReadTheDocs based preview and check it in a jupyter deployment by hand
OR
ii) You can have a TestCase through doctest doing 

- Setup FortranEvaluator
- call evaluate2( code_snippet_1)
- call evaluate2( code_snippet_2)
- .....

And if the test works we have our answer.

Though if you still think we should have a way for every PR to build it's own deployment, we can think through it. We might need to switch our approach from standard Github pages support being offer though.

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -696,3 +696,73 @@ jobs:
...
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
```
certik, line 768: What setup does this require to work?

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -696,3 +696,73 @@ jobs:
...
+              --XeusAddon.prefix=${{ env.PREFIX }} \
+              --contents share/lfortran/nb/Demo1.ipynb \
+              --contents share/lfortran/nb/Demo2.ipynb \
+              --contents share/lfortran/nb/Variables.ipynb \
+              --output-dir dist
+
+      - name: Upload pages artifact
+        uses: actions/upload-pages-artifact@v4
+        with:
+          path: ./dist
```
certik, line 752: This part looks good.

anutosh491: Thank you.

.github/workflows/Exhaustive-Checks-CI.yml
anutosh491: This requires GitHub Pages to be enabled for the repository with **Settings → Pages → Build and deployment → Source** set to **GitHub Actions**.

No separate deployment credentials are required. The workflow uploads the `github-pages` artifact using `actions/upload-pages-artifact`, and `actions/deploy-pages` deploys it using the job's `pages: write` and `id-token: write` permissions.

The `github-pages` environment will be created automatically if it does not already exist.

certik: Ok. We need to ensure that ALL things that can break are tested by a PR, so that the extra things that only happen when merged into main are minor and won't break in practice. So I think running `deploy_jupyterlite` in main only is probably fine, I think the action just copies over some files and publishes to github pages.

But the `build_xlfortran_jupyterlite` must run for both main and a PR, to ensure things build cleanly. The build can really easily break. For some reason it is not being tested at the PR:

Image

So we need to fix it.

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -696,3 +696,73 @@ jobs:
           context: .
           push: ${{ github.event_name == 'push' }}
           tags: ${{ steps.image-tags.outputs.tags }}
+
+  build_xlfortran_jupyterlite:
+    name: Build xlfortran JupyterLite WASM kernel
+    runs-on: ubuntu-24.04
+    if: >-
+      ${{ github.event_name == 'push' ||
+          (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'Tests::Run-Exhaustive')) }}
```
certik, line 705: Is this condition correct? According to it, it should run in a pull request with the `Run-Exhaustive` label, but it doesn't.

.github/workflows/Exhaustive-Checks-CI.yml
certik: Oh my bad! I think it's the `deploy_jupyterlite` that is skipped (as it should) and `build_xlfortran_jupyterlite` runs.

.github/workflows/Exhaustive-Checks-CI.yml
certik: It seems correct, and it runs, I made a mistake.

.github/workflows/Exhaustive-Checks-CI.yml
certik: Ok, I set

image

so we should be good to go.

certik: It works!

https://lfortran.github.io/lfortran/

PR #11694, opened 2026-05-31
Test WasmLFortranExecutor & FortranEvaluator being built for wasm

anutosh491:
Built on top of #11693


anutosh491:
### Working 

I realize that we are already testing `LLVMEvaluator` & `FortranEvaluator` through `test_llvm.cpp` 

- When building for Jupyterlite i.e `XEUS_LFORTRAN_WASM_BUILD` we don't really want to check other tests but only check working of `WasmLFortranExecutor` & `FortranEvaluator` (which would use `WasmLFortranExecutor` underneath when built for Jupyterlite). Basically we only want to check if the interpreter would work in the browser.

That's being done through these changes in the PR
```
# For the JupyterLite kernel build we only need to exercise the evaluator path
# (FortranEvaluator -> WasmLFortranExecutor). The other tests are already
# covered by the HAVE_BUILD_TO_WASM CI job.
if (XEUS_LFORTRAN_WASM_BUILD)
    set(SRC test_llvm.cpp)
endif()

if (XEUS_LFORTRAN_WASM_BUILD)
    # WasmLFortranExecutor uses dlopen — test binary must be a MAIN_MODULE.
    target_link_options(test_lfortran PRIVATE
        "SHELL:-fwasm-exceptions"
        "SHELL:-s MAIN_MODULE=1"
        "SHELL:-s ALLOW_MEMORY_GROWTH=1"
        "SHELL:-s STACK_SIZE=32mb"
        "SHELL:-s INITIAL_MEMORY=128mb"
    )
endif()
```

So just like while running the kernel in Jupyterlite where
- we build xlfortran.js/wasm as the Main Module  
- each cell gives us a unique Side module loaded on top of the main module xlfortran

We use a similar approach for the tests
- We build test_lfortran.js/wasm as the Main Module
- each test would instantiate a `WasmLFortranExecutor` that produces side modules loaded on top of the above main module generated.

Notes 
- The LLVMEvalutor based tests would be skipped for emscripten cause ORC JIT won't run in wasm & that's exactly why we have the WasmLFortranExecutor
- But the `FortranEvaluator` has this block 
https://github.com/lfortran/lfortran/blob/3cc3a7e212e524471fa0c4392cbea3de1fa709a7/src/lfortran/fortran_evaluator.cpp#L167-L174
and picks the evalutor we need for the specific platform.
- So we should be able to run all `FortranEvaluator` test even while building for wasm.

src/lfortran/tests/test_llvm.cpp
```diff
@@ -1401,3 +1425,19 @@ TEST_CASE("FortranEvaluator pass_array_by_data on global random_number") {
...
+    auto m = we.parse_module2(R"""(
+define i64 @__lfortran_evaluate_1()
+{
+    ret i64 42
+}
+    )""", "test.ll");
+    we.add_module(std::move(m), 1);
+    CHECK(we.execfn<int64_t>("__lfortran_evaluate_1") == 42);
+}
+#endif
```
anutosh491, line 1443: Technically all the tests that are framed for `LLVMEvaluator` can also be replicated (or basically duplicated) for `WasmLFortranExecutor` 

I have started with a simple one here.

src/lfortran/tests/test_llvm.cpp
```diff
@@ -520,20 +526,24 @@ end function
...
+integer subroutine fn_sub(i, j, r)
 integer, intent(in) :: i, j
 integer, intent(out) :: r
 r = i + j
 end subroutine
 )");
     e.evaluate2("integer :: r");
-    e.evaluate2("call fn(2, 3, r)");
+    e.evaluate2("call fn_sub(2, 3, r)");
     LCompilers::Result<FortranEvaluator::EvalResult>
```
anutosh491, line 547: You shall see that in some tests (total 3 in number) I have changed some var/function names. These are harmless changes but required.

Reason :

When we Run the kernel in Jupyterlite:
- 1 Main Module -> comes from fortran_kernel.cpp that maintains the FortranEvaluator (so only 1 `WasmLFortranExecutor`)
- Cells produce side modules loaded on top of the main module

For tests
- 1 Main Module -> coming from test_llvm.cpp 
- each test creates a `WasmLFortranExecutor` which creates a Side Module per `evaluate2` call. So there are multiple `WasmLFortranExecutor` instances being created here.
- So basically each side module (possibly created by different instances of the `WasmLFortranExecutor`) is loaded on top of the main module

https://github.com/lfortran/lfortran/blob/3cc3a7e212e524471fa0c4392cbea3de1fa709a7/src/libasr/codegen/evaluator.cpp#L657-L660

So one would expect us to call dlclose on the side modules already loaded through `WasmLFortranExecutor`'s destructor to start a fresh everytime we create a `WasmLFortranExecutor` in a test case.

But emscripten's dlclose is very poor (just cause the usecase for dlclose in wasm is very minmalisitic). Read more : https://github.com/emscripten-core/emscripten/issues/23939

So what would happen is emscripten's dynamic linker uses first-defined-wins semantics for RTLD_GLOBAL: once a symbol name is loaded into the global namespace, any later side-module defining the same name is silently ignored (or triggers a WebAssembly type-mismatch LinkError if the signatures differ). 

So when two test cases both define a Fortran function named fn (with different arities/types), the second test's copy is never loaded, causing wrong results or a hard crash. Renaming to fn_sub, fn_redecl, ad etc. gives each test case its own globally unique symbols in the WASM process.

and hence my comment here.

https://github.com/lfortran/lfortran/pull/11694/changes#diff-a38fe738abc7cbb7941365c99a4c7b94a55e0cd3ad445a3e17e411e331db460eR529-R532

anutosh491:
Here's the tests running through node https://github.com/lfortran/lfortran/actions/runs/26709159936/job/78716038789?pr=11694#step:7:1

This works exactly as expected
```
+ /home/runner/micromamba-bin/micromamba run -n xeus-lfortran-wasm-build node build-wasm/src/lfortran/tests/test_lfortran.js
5 0 16 841 39948248 39948248 0 0 39947456 3
5 5 5
[doctest] test cases:  28 |  28 passed | 0 failed | 0 skipped
[doctest] assertions: 183 | 183 passed | 0 failed |
[doctest] Status: SUCCESS!
```

src/lfortran/tests/test_llvm.cpp
```diff
@@ -1376,6 +1397,9 @@ TEST_CASE("FortranEvaluator kind parameter from global symtab") {
     CHECK(r.ok);
 }
 
+#if !defined(__EMSCRIPTEN__)
+// TODO: The 80 MB stack allocation (real(dp) :: x(10**7)) exceeds the WASM
+// shadow-stack budget and causes a memory-access-out-of-bounds at runtime.
```
anutosh491, line 1402: Only test being skipped and added as a TODO now but it works as expected in Jupyterlite

Image

Through node while being tested it cause a memory out of bounds issue which I'll try to figure out through subsequent PRs
```
(xeus-lfortran-wasm-build) anutosh491@Anutoshs-MacBook-Air lfortran % node build-wasm/src/lfortran/tests/test_lfortran.js
5 0 16 35 15 140606976 762605412 1751343469 761622121 6776419
5 5 5
wasm://wasm/0ed16f8a:1


RuntimeError: memory access out of bounds
    at wasm://wasm/0ed16f8a:wasm-function[172]:0x76e607
    at wasm://wasm/0ed16f8a:wasm-function[172]:0x76e60a
    at wasm://wasm/0ed16f8a:wasm-function[172]:0x76e60a
    at wasm://wasm/0ed16f8a:wasm-function[137]:0x75c16f
    at wasm://wasm/0ed16f8a:wasm-function[2795]:0xaae70b
    at wasm://wasm/0ed16f8a:wasm-function[165]:0x76d5a5
    at wasm://wasm/0ed16f8a:wasm-function[4463]:0xd2f31b
    at wasm://wasm/0ed16f8a:wasm-function[4482]:0xd3109d
    at callMain (/Users/anutosh491/work/lfortran/build-wasm/src/lfortran/tests/test_lfortran.js:1:540330)
    at doRun (/Users/anutosh491/work/lfortran/build-wasm/src/lfortran/tests/test_lfortran.js:1:540749)

Node.js v25.8.2
```

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -690,3 +690,79 @@ jobs:
...
+      - name: Build JupyterLite site
+        run: |
+          set -ex
+          micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge -y
+          micromamba run -n xeus-lite-host jupyter lite build \
+              --XeusAddon.prefix=${{ env.PREFIX }} \
+              --contents share/lfortran/nb/Demo1.ipynb \
+              --contents share/lfortran/nb/Demo2.ipynb \
+              --contents share/lfortran/nb/Variables.ipynb \
+              --output-dir dist
```
certik, line 753: This runs the notebooks in jupyter-lite? How does that work, is that using nodejs to run the WASM, at the CI?

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -690,3 +690,79 @@ jobs:
...
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
```
certik, line 774: You can just skip this part for now, that way we can merge the big win: that the wasm is now being tested by jupyterlite (if my understanding is correct above), which is a big win.

mcbarton:
@certik Here is some info to help you with reviewing this PR, since it appears from your comments your incorrectly thinking it tests the jupyter lite website.

This PR does not run the notebooks inside Jupyter-lite. This `jupyter lite build` command purely builds the Jupyter lite website, and outputs it to a folder dist that needs to be created so that it can be uploaded as an artifact of the ci (which in turn gets deployed if the `deploy_jupyterlite` section of this ci is included and the Github repos settings are correct). 

If you see here https://github.com/lfortran/lfortran/actions/runs/26709159936/job/78716038789?pr=11694#step:7:24 you might get a better of what this PR is trying to achieve. It is essentially building the doctest tests you have in `test_llvm.cpp` using Emscripten. This produces a javascript executable which can be run in node (see the `micromamba run -n xeus-lfortran-wasm-build node build-wasm/src/lfortran/tests/test_lfortran.js` command). If you want these doctest tests inside a browser (not in this PR) you can also add the `--emrun` flag to the link options for `test_lfortran`. You can then run the doctest tests inside a browser using the emrun python script which comes with Emscripten (see xeus-cpps readme https://github.com/compiler-research/xeus-cpp).

It is possible to test the deployment itself (e.g. run the notebooks including in the deployment), but there is no settled upon way in which to do this. You can see my suggested approach in this xeus-cpp PR https://github.com/compiler-research/xeus-cpp/pull/415 which uses `Selenium` to interact with a temporarily hosted Jupyter-Lite site, to run the notebook, download it, and then compare it to a reference notebook to check the output is as expected. The quant stack approach is to use playwright and can be seen here https://github.com/jupyter/try-jupyter/pull/73 . I am biased, but the current implementation in try-jupyter doesn't appear to actually check the notebooks come out as expected. It only checks that the notebooks can run, and that any warnings outputted do not match a very particular list of warnings (see https://github.com/jupyter/try-jupyter/blob/79317c49db01ba2bdc86e9f8c120d335f09a332d/ui-tests/utils.py#L6).

There are hacky ways around it, but deployment previews in PRs in not really a thing using Githubs infrastructure (you'll see there is a preview option in the deploy-pages action but it will be forever only available to Github admins is my understanding). I don't really know how lfortrans documentation is generated, but with read the docs you have pr previews which can be enabled, so might be worth looking into the documenation of whoever hosts lfortrans documentation to see if they provide such an option. If they do you can kind of build the deployment as part of the documentation and deploy as part of a pr preview, to allow manual testing as a stop gap until automatic ci testing can be provided.

certik:
@mcbarton Thanks a lot for the clarification. What are these lines doing: https://github.com/lfortran/lfortran/pull/11694#discussion_r3330367399:
```
          micromamba run -n xeus-lite-host jupyter lite build \
              --XeusAddon.prefix=${{ env.PREFIX }} \
              --contents share/lfortran/nb/Demo1.ipynb \
              --contents share/lfortran/nb/Demo2.ipynb \
              --contents share/lfortran/nb/Variables.ipynb \
              --output-dir dist
```
They are just copying over all the notebooks and wasm, but does _not_ actually run it, correct?

I actually like neither Selenium nor Playwright. Those would be needed to test the jupyter-lite itself, so you would use it. But for us, we trust that jupyter-lite works (if it doesn't, then we'll report a bug and it gets fixed upstream). But we do not trust that our own lfortran jupyter-lite kernel works, so that's what needs to be tested. I would personally test it with wasmtime or nodejs. I think emscripten-compiled code doesn't work in wasmtime, so we must use nodejs, which is fine. We should simply call it like jupyter-lite would and ensure we can "execute" cells with the emulated test calls. I think that is all that we need to do.

anutosh491:
> I actually like neither Selenium nor Playwright

Me too. I don't really think we need to take that route.

We don't really need to test xeus-lfortran kernel in jupyterlite but we need to test the xeus-lfortran kernel in wasm right ?

Which is what this PR is trying to achieve

- Made a note of the working [here](https://github.com/lfortran/lfortran/pull/11694#issuecomment-4586258895)
- A simple test case like this means that we are 
i) Setting up the FortranEvaluator (which would choose WasmLFortranExecutor executor underneath as we're building for wasm using emscripten)
ii) and every `evaluate2` call replicate running a cell in jupyterlite. So the below is 3 cells running.

```
TEST_CASE("FortranEvaluator 1") {
    CompilerOptions cu;
    cu.interactive = true;
    cu.po.runtime_library_dir = LCompilers::LFortran::get_runtime_library_dir();
    FortranEvaluator e(cu);
    LCompilers::Result<FortranEvaluator::EvalResult>
    r = e.evaluate2("integer :: i");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::none);
    r = e.evaluate2("i = 5");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::statement);
    r = e.evaluate2("i");
    CHECK(r.ok);
    CHECK(r.result.type == FortranEvaluator::EvalResult::integer4);
    CHECK(r.result.i32 == 5);
}
```

So as we know 1 cell executed is equivalent to 1 wasm side module. So 3 cells gives us __lfortran_evaluate_0_1.wasm, __lfortran_evaluate_0_1.wasm, __lfortran_evaluate_0_1.wasm as shown below

image

So if required (maybe in Debug mode while running tests) we should be able to dump these wasm modules generated through the tests in a unique location and compare them to what we see in Jupyterlite. This is what I was trying yesterday
```
(xeus-lfortran-wasm-build) anutosh491@Anutoshs-MacBook-Air lfortran % node build-wasm/src/lfortran/tests/test_lfortran.js
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--6079e0
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--d69c05
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--5230de
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--1a30f9
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--bbc00d
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--f495c6
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--7f314f
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--6d3d76
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--0052e6
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--e13e03
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--e4bf4e
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--d6fceb
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--e0efd9
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--0b3558
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--ae0174
5 0 16 43 0 0 41106960 1 160 0
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--cd45f5
5 5 5
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--ba91fc
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--38d7dc
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--ae671d
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--7b2fe6
[WasmLFortranExecutor] temp dir: /tmp/xlfortran-wasm-exec--525928
[doctest] test cases:  22 |  22 passed | 0 failed | 0 skipped
[doctest] assertions: 152 | 152 passed | 0 failed |
[doctest] Status: SUCCESS!
```

So every test got its unique location and then we can compare the wasm generated whenever in doubt.

But all in all, we don't really need to test stuff within Jupyterlite itself unless it's a UI test i.e the cells are working or failing. Comparing wasms should equate comparing outputs. If the wasm module is same, the output would be same.

mcbarton:
> @mcbarton Thanks a lot for the clarification. What are these lines doing: [#11694 (comment)](https://github.com/lfortran/lfortran/pull/11694#discussion_r3330367399):
> 
> ```
>           micromamba run -n xeus-lite-host jupyter lite build \
>               --XeusAddon.prefix=${{ env.PREFIX }} \
>               --contents share/lfortran/nb/Demo1.ipynb \
>               --contents share/lfortran/nb/Demo2.ipynb \
>               --contents share/lfortran/nb/Variables.ipynb \
>               --output-dir dist
> ```
> 
> They are just copying over all the notebooks and wasm, but does _not_ actually run it, correct?
> 

Yes you are right. This command purely builds the website using the jupyter lite command. 

`jupyter lite build` - tells jupyter lite to build a jupyter lite website with the options given 
`--XeusAddon.prefix` - tells jupyter lite where to look for the kernel and its dependencies
`--contents` - tells jupyter lite where the files are that you want uploaded to home directory of the jupyter lite filesystem (in this case the notebooks.  it doesn't have to be just that though. for example other projects upload a readme file)
`--output-dir` - folder name where the jupyter lite website will be outputted to

If you was trying to test the jupyter lite website locally, all you would do is run this command but change `build` to `serve` which runs a http server locally (with serve the output of the command ends with something similar to the below)

```

        Serving JupyterLite Debug Server from:
            /Users/test/xeus-cpp/build/dist
        on:
            http://127.0.0.1:8000/index.html

        *** Exit by: ***
            - Pressing Ctrl+C
            - Visiting /shutdown
```

If your not interested in having the jupyter lite site until the testing is in place, then the last section of the ci you need is `Run WASM evaluator tests`

anutosh491:
Hey @certik, 

We should get this PR in next which frames tests for the WasmLFortranExecutor & FortranEvaluator while being built for wasm. 

I've been verbose with my approach and also the comments in the PR. Let me know if something need clarification.

certik (approved):
I think that this is fine, but the symbol redifinitions should work --- I think it creates a new scope for each cell, so you are free to redefine any variable as you want.

So it's a bug and we need to fix it.

anutosh491:
Thanks for the merge. Yess we can do better here. I shall do some ideation soon on this !

PR #11698, opened 2026-05-31
fix: maxloc/minloc compile-time eval for multi-dim arrays with mask

### LINKED ISSUE #11696: comparison operator not implemented for maxloc parameter array
## MRE

```fortran
program main
  implicit none

  integer, parameter :: x(1, 1) = 1
  integer, parameter :: a(2) = maxloc(x, mask=x > 1)

  print *, a /= [0, 0]
end program main
```

## output

```text

semantic error: Comparison operator not implemented
 --> t.f90:7:12
  |
7 |   print *, a /= [0, 0]
  |            ^^^^^^^^^^^
```

## gfortran behavior

```text

 F F
```

Jatinagarwal24:
fixes #11696
The compile-time evaluation of maxloc/minloc (eval_MaxMinLoc) had two
bugs when operating on multi-dimensional parameter arrays:

1. The mask argument was not properly resolved when wrapped in an
   ArrayPhysicalCast node. Since ArrayPhysicalCast has a null m_value,
   expr_value() returned nullptr, causing the mask to be ignored and
   defaulting to all-true.

2. When no mask element was true (flag==0), the result ArrayConstant
   was constructed with only 1 zero element regardless of the original
   array's dimensionality. For a 2D array, maxloc should return a
   2-element array [0, 0], not [0].

The fix unwraps ArrayPhysicalCast before resolving the mask value, and
constructs result arrays with the correct number of elements based on
the return type size. For multi-dimensional arrays with a valid match,
the flat index is decomposed into per-dimension 1-based indices using
column-major (Fortran) ordering.


Jatinagarwal24:
Hi,I am aware of regression i will fix that.
Thanks !

src/libasr/pass/array_struct_temporary.cpp
```diff
@@ -660,36 +660,51 @@ bool set_allocation_size(
                 case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Product):
                 case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::MaxVal):
                 case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::MinVal):
-                case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Norm2): {
-                    size_t n_dims = ASRUtils::extract_n_dims_from_ttype(
-                        intrinsic_array_function->m_type);
-                    allocate_dims.reserve(al, n_dims);
-                    for( size_t i = 0; i < n_dims; i++ ) {
+                case static_cast<int64_t>(ASRUtils::IntrinsicArrayFunctions::Norm2):
```
assem2002, line 663: Just add the switch's case handling for maxloc/minloc (or maybe don't at all for now), don't use if/else just add the case and keep same exact implementation.
***
- maxloc/minloc -> with no dim parameter specified -> always returns fixedsize array of rank=1,size=rank_of_passed_array -> hence this function won't be touched. We should actually assert that all these functions (maxloc, sum, all, ..) reach here only when dim specified.

- maxloc/minloc -> with dim specified -> will touch this function and use the dim handling (reducing) properly.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1365,12 +1371,36 @@ static inline ASR::expr_t *eval_MaxMinLoc(Allocator &al, const Location &loc,
...
+                    int result_size = ASRUtils::get_fixed_size_of_array(type);
+                    std::vector<ASR::expr_t*> zeros;
+                    for (int i = 0; i < result_size; i++) {
+                        zeros.push_back(b.i32(0));
+                    }
+                    return b.ArrayConstant(zeros, extract_type(type), false);
+                }
+            }
             mask = ASR::down_cast<ASR::ArrayConstant_t>(b.ArrayConstant({b.bool_t(mask_val, logical)}, logical, false));
+        } else if (mask_arg && ASR::is_a<ASR::ArrayConstant_t>(*mask_arg)) {
```
assem2002, line 1393: this branch could've been avoided, but i think it's fine.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1379,10 +1409,11 @@ static inline ASR::expr_t *eval_MaxMinLoc(Allocator &al, const Location &loc,
             mask = ASR::down_cast<ASR::ArrayConstant_t>(b.ArrayConstant(mask_data, logical, false));
         }
         ASR::LogicalConstant_t *back = ASR::down_cast<ASR::LogicalConstant_t>(ASRUtils::expr_value(args[4]));
+        int mask_size = ASRUtils::get_fixed_size_of_array(mask->m_type);
         int index = 0;
         int flag = 0;
         for (int i = 0; i < arr_size; i++) {
-            if (((bool*)mask->m_data)[i] != 0) {
+            if (((bool*)mask->m_data)[(mask_size == 1) ? 0 : i] != 0) {
```
assem2002: is that needed?

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1365,12 +1371,36 @@ static inline ASR::expr_t *eval_MaxMinLoc(Allocator &al, const Location &loc,
...
-            if (mask_val == false) return b.i_t(0, type);
+        // Resolve the mask via expr_value() first, since args[2] may be
+        // a wrapped expression (e.g. ArrayPhysicalCast of IntegerCompare)
+        // whose value is the actual ArrayConstant.
+        ASR::expr_t* mask_arg = args[2];
+        // Unwrap ArrayPhysicalCast, since its m_value is typically null
+        // and we need the inner expression's value.
+        if (mask_arg && ASR::is_a<ASR::ArrayPhysicalCast_t>(*mask_arg)) {
+            mask_arg = ASR::down_cast<ASR::ArrayPhysicalCast_t>(mask_arg)->m_arg;
+        }
```
assem2002, line 1382: Find where we create that cast and actually make arrayphysicalCast->m_value = m_arg's value if exist.
That's the most clean approach here.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1365,12 +1371,36 @@ static inline ASR::expr_t *eval_MaxMinLoc(Allocator &al, const Location &loc,
...
+                    return b.i_t(0, type);
+                } else {
+                    int result_size = ASRUtils::get_fixed_size_of_array(type);
+                    std::vector<ASR::expr_t*> zeros;
+                    for (int i = 0; i < result_size; i++) {
+                        zeros.push_back(b.i32(0));
+                    }
+                    return b.ArrayConstant(zeros, extract_type(type), false);
+                }
+            }
```
assem2002, line 1391: remove this short circuit. same logic is done below.

src/libasr/pass/intrinsic_array_function_registry.h
assem2002: also for the same below?

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -1488,7 +1524,32 @@ static inline ASR::expr_t *eval_MaxMinLoc(Allocator &al, const Location &loc,
...
+                // indices using column-major (Fortran) ordering.
+                ASR::dimension_t* orig_dims = nullptr;
+                int n_orig_dims = ASRUtils::extract_dimensions_from_ttype(orig_array_type, orig_dims);
+                std::vector<int64_t> dim_sizes(n_orig_dims);
+                for (int d = 0; d < n_orig_dims; d++) {
+                    int64_t ds = 1;
+                    if (orig_dims[d].m_length != nullptr) {
+                        ASRUtils::extract_value(ASRUtils::expr_value(orig_dims[d].m_length), ds);
+                    }
+                    dim_sizes[d] = ds > 0 ? ds : 1;
```
assem2002: assert that (ds>0)

assem2002 (commented):
Great!, looks good.
Please address my comment

Jatinagarwal24:
Hi @assem2002 I have addressed your comments.Pls once review it 
Thanks for your suggestions.

src/libasr/asr_utils.h
```diff
@@ -7291,6 +7291,9 @@ static inline ASR::asr_t* make_ArrayPhysicalCast_t_util(Allocator &al, const Loc
         }
     }
 
+    if (!a_value && a_arg) {
+        a_value = ASRUtils::expr_value(a_arg);
+    }
```
Jatinagarwal24: I tried setting `m_value` in` make_ArrayPhysicalCast_t_util` as suggested, but it actually causes regressions in  `complex_array_member_access.f90` when compiled with `fast` flag.
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran integration_tests/complex_array_member_access.f90 --fast
LCOMPILERS_ASSERT failed: /root/lfortran/src/libasr/codegen/llvm_utils.cpp
function create_gep2(), line number 1811 at
ds_pointee_type == t
ERROR MESSAGE:
Type mismatch in create_gep2: GEP target type does not match pointer's pointee type. This would cause crashes in LLVM <= 8 constant folder. Target type: [2 x %complex_4], Pointer pointee type: %complex_4 = 

Aborted (core dumped)
```
The issue is that `expr_value()` returns an `ArrayConstant_t` (which uses a fixed-size LLVM array literal), but the physical cast is converting it to a DescriptorArray struct. If m_value is populated with the raw constant, the LLVM backend (`visit_ArrayPhysicalCast`) shortcuts to emitting the literal instead of a descriptor, which crashes downstream code expecting the descriptor fields.

Because of this type mismatch, it's safer to leave `m_value` null for descriptors and just manually unwrap the cast inside `eval_MaxMinLoc`. 
So, I've reverted the m_value assignment and kept the manual unwrap so as to fix the regression.

assem2002: actually that's great observation. the casting node is meant to be there only for the sake of casting -- We intentionally in need of casting that array, that probably mean we should opt-out the `value` member.

assem2002: #12119

assem2002 (approved):
LGTM

PR #11704, opened 2026-06-01
Fix: internal writes to character array sections

### LINKED ISSUE #11697: ICE for write to character array section with negative stride
## MRE

```fortran
program main
  implicit none

  character(len=2) :: c(5)

  c = "?"
  write(unit=c(5:1:-2), fmt="(A)") "x", "y", "z"

  if (c(1) /= "z" .or. c(2) /= "?" .or. c(3) /= "y" .or. &
      c(4) /= "?" .or. c(5) /= "x") error stop
end program main
```

## output

```text

LCOMPILERS_ASSERT failed:
Type mismatch in create_gep2: GEP target type does not match pointer's pointee type.

```

## gfortran behavior

```text

(compiles and runs successfully)
```

opixdown:
fixes #11697

### DIFF
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -666,7 +666,20 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
         if (fmt->n_args == 0) return;
         ASR::expr_t* val_arg = fmt->m_args[0];
         bool is_do_loop = ASR::is_a<ASR::ImpliedDoLoop_t>(*val_arg);
-        if (!is_do_loop && !ASRUtils::is_array(ASRUtils::expr_type(val_arg))) return;
+        bool is_array_section_unit_with_scalar_values =
+            ASR::is_a<ASR::ArraySection_t>(*x.m_unit) && fmt->n_args > 1;
+        if (is_array_section_unit_with_scalar_values) {
+            for (size_t i = 0; i < fmt->n_args; i++) {
+                if (ASR::is_a<ASR::ImpliedDoLoop_t>(*fmt->m_args[i]) ||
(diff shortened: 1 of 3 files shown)


src/libasr/pass/array_op.cpp
```diff
@@ -701,6 +714,38 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
                 ub_lhs = sec->m_args[0].m_right;
             }
         }
+
+        if (is_array_section_unit_with_scalar_values) {
```
assem2002, line 718: Good workaround for now -- but we need to find proper way of handling print/write.
The current approach doesn't handle proper record advancement 
***
Here is simple rule from Calude :
```
output list not exhausted + format exhausted  →  reversion  →  new record
```
You can see that formatting plays a vital rule here -- we handle formatting and argument work at the runtime, there's no way to invoke a **next-record** action in the ASR.

We should do that in the future.

****
An example to showcase the issue we have :
```.f90
write(c(5:1:-2), "(A, I0)") "x", 2, "z"
write(c(5:1:-2), "(A,A)") "x", "V", "z"
```

src/libasr/pass/array_op.cpp
```diff
@@ -701,6 +714,38 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+            ASR::ArraySection_t* sec = ASR::down_cast<ASR::ArraySection_t>(x.m_unit);
+            if (sec->n_args != 1) return;
```
assem2002, line 720: We should assert here with an ICE -- The return here results in undetermined behavior.

assem2002 (approved):
LGTM. Thanks!
Good for now and it resolves some edge case for now -- no problem.

PR #11706, opened 2026-06-01
fix : ICE for internal read from equivalenced character array

### LINKED ISSUE #11705: ICE for internal read from equivalenced character array
## MRE

```fortran
program p
  implicit none

  character(len=4) :: x(8) = (/ &
      "abcd", "efgh", "ijkl", "mnop", &
      "qrst", "uvwx", "yz01", "2345" /)
  character(len=4) :: y(2, 4)
  character(len=8) :: z

  equivalence (x, y)

  read(y, "(a8)") z
  print *, z
end program p
```

## output

```text

Internal Compiler Error:
AssertFailed: f != nullptr
```

## gfortran behavior

```text

 abcd
```

Jatinagarwal24:
fixes #11705

Bug 
1. The init_expr pass prepended initializer assignments (x(1)="abcd",
   x(2)="efgh", ...) before the CPtrToPointer statement that sets up
   the equivalence pointer. At that point x's descriptor had size=0,
   so bounds checks failed and the program segfaulted.

2. In visit_ArrayConstant (pass_utils.cpp), extracting dimension info
   via expr_value(m_dims[i].m_length) crashed for pointer/equivalence
   variables whose type carries empty descriptor dims [(() ())].

Fix:
- init_expr.cpp: hoist leading CPtrToPointer statements before the
  generated init assignments, so equivalence pointers are set up
  before any indexing into them.
- pass_utils.cpp: use arr_var's dimensions as primary source (needed
  for multi-dim data statements like ABC(2,3,4,5)), but fall back to
  the ArrayConstant's own type when arr_var's dims are null (pointer/
  equivalence variables).

### DIFF
--- a/src/libasr/pass/init_expr.cpp
+++ b/src/libasr/pass/init_expr.cpp
@@ -142,6 +142,18 @@ class InitExprVisitor : public ASR::CallReplacerOnExpressionsVisitor<InitExprVis
             Vec<ASR::stmt_t*> body;
             body.reserve(al, n_body);
 
+            // Collect leading CPtrToPointer statements from the existing body
+            // first (e.g. equivalence pointer setup). These must execute before
+            // any init-expr assignments that index into the pointer arrays.
+            size_t i = 0;
+            for (; i < n_body; i++) {
+                if (ASR::is_a<ASR::CPtrToPointer_t>(*m_body[i])) {
+                    body.push_back(al, m_body[i]);
+                } else {
+                    break;
+                }
+            }
+
             if( symtab2decls.find(current_scope) != symtab2decls.end() ) {
                 Vec<ASR::stmt_t*>& decls = symtab2decls[current_scope];
                 for (size_t j = 0; j < decls.size(); j++) {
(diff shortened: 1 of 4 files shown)


src/libasr/pass/init_expr.cpp
```diff
@@ -142,6 +142,18 @@ class InitExprVisitor : public ASR::CallReplacerOnExpressionsVisitor<InitExprVis
             Vec<ASR::stmt_t*> body;
             body.reserve(al, n_body);
 
+            // Collect leading CPtrToPointer statements from the existing body
+            // first (e.g. equivalence pointer setup). These must execute before
+            // any init-expr assignments that index into the pointer arrays.
+            size_t i = 0;
+            for (; i < n_body; i++) {
+                if (ASR::is_a<ASR::CPtrToPointer_t>(*m_body[i])) {
```
assem2002, line 150: It's depending on the fact that equivalence gets lowered into `CPtrToPointer_t` and inserted as the first couple of statements in the body.
I think that's fine

assem2002 (approved):
LGTM. Thanks.

PR #11709, opened 2026-06-01
Fix ASR: Create temporary for struct function returns to avoid LHS aliasing

### LINKED ISSUE #8082: Bug : `subroutine_from_function` should create its own temp variable instead of relying on exisiting LHS
```.f90
program p
    type :: t
    end type t
    
    type(t) :: varr

    varr%arr = [1,2,3]

    varr = ff(varr)

    contains 

    function ff(s) result(ret) 
       type(t),intent(in) :: s
       type(t) :: ret
       print *, s%arr
       allocate(ret%arr(3))

       ret%arr = s%arr
       print *, ret%arr
    end function 
end program
```

srinjoy933:
fixes #8082


srinjoy933:
Hey @assem2002 , please have a look on this pr , Thank You!

assem2002 (commented):
I think the issue already works on main.
Do we still need the changes introduced in this PR?

srinjoy933:
Hey @assem2002 ,as the current test works on the main ,I dont think this changes are required for this issue , so I am reverting the changes and only adding the test ,Thanks!

assem2002 (approved):
Thank you!

integration_tests/struct_type_15.f90
```diff
@@ -0,0 +1,23 @@
+program p
+    type :: t
+        integer, allocatable :: arr(:)
+    end type t
+    
+    type(t) :: varr
+    allocate(varr%arr(3))
+    
+    varr%arr = [1,2,3]
+    varr = ff(varr)
```
assem2002, line 10:
```suggestion
    varr = ff(varr)
    if (.not. all(varr%arr == [1, 2, 3])) error stop
```

PR #11715, opened 2026-06-02
fix: Reject array values in select case labels

### LINKED ISSUE #11714: code generation error for array expression in case range bound
## MRE

```fortran
program p
  implicit none

  integer :: n = 1

  print *, "select case array bound MRE"

  select case (n)
  case (:[2])
  end select
end program p
```

## output

```text

Both operands to ICmp instruction are not of the same type!
  %30 = icmp sle i32 %29, [1 x i32]* %__libasr__created__var__0__array_constant_

```

## gfortran behavior

```text

Error: Expression in CASE selector at (1) must be scalar
```

opixdown:
fixes #11714

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -3852,7 +3852,16 @@ public:
                 AST::CaseCondExpr_t *condexpr
                     = AST::down_cast<AST::CaseCondExpr_t>(Case_Stmt->m_test[i]);
                 this->visit_expr(*condexpr->m_cond);
-                expr_accum.push_back(ASRUtils::EXPR(tmp));
+                ASR::expr_t* case_expr = ASRUtils::EXPR(tmp);
+                if (ASRUtils::is_array(ASRUtils::expr_type(case_expr))) {
+                    diag.add(Diagnostic(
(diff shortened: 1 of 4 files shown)


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -3852,7 +3852,16 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                 AST::CaseCondExpr_t *condexpr
                     = AST::down_cast<AST::CaseCondExpr_t>(Case_Stmt->m_test[i]);
                 this->visit_expr(*condexpr->m_cond);
-                expr_accum.push_back(ASRUtils::EXPR(tmp));
+                ASR::expr_t* case_expr = ASRUtils::EXPR(tmp);
+                if (ASRUtils::is_array(ASRUtils::expr_type(case_expr))) {
```
assem2002, line 3856: We should only accept `integer`, `logical`, `character` and only constant values and only scalars.

It's fine for now.

assem2002 (approved):
LGTM. Thanks.

assem2002:
https://github.com/lfortran/lfortran/issues/11736#issue-4580446156

PR #11717, opened 2026-06-02
fix: ICE for structure constructor with allocatable component

### LINKED ISSUE #11716: ICE for structure constructor with allocatable component
## MRE

```fortran
program null_allocatable_component_mre
  implicit none

  type :: box
    integer, allocatable :: x
  end type

  type(box) :: y

  y = box(null())

  if (allocated(y%x)) error stop
  print *, "ok"
end program null_allocatable_component_mre
```

## output

```text

Internal Compiler Error:
AssertFailed: current_variable_type_ != nullptr
```

## gfortran behavior

```text

 ok
```

opixdown:
fixes #11716

### DIFF
--- a/integration_tests/derived_types_146.f90
+++ b/integration_tests/derived_types_146.f90
@@ -1,0 +1,16 @@
+program derived_types_146
+    implicit none
+
+    type :: box
+        integer, allocatable :: x
+    end type
+
+    type(box) :: a, b
+
+    a = box(null())
+    b = box(x=null())
+
+    if (allocated(a%x)) error stop
+    if (allocated(b%x)) error stop
+    print *, "ok"
+end program
--- a/integration_tests/derived_types_146.f90
+++ b/integration_tests/derived_types_146.f90
@@ -12,5 +12,13 @@ program derived_types_146
 
     if (allocated(a%x)) error stop
     if (allocated(b%x)) error stop
+    if (func(null()) /= 0) error stop
+    if (func(v=null()) /= 0) error stop
     print *, "ok"
+
+contains
+    integer function func(v)
+        integer, intent(in), pointer :: v
+        func = 0
+    end function
 end program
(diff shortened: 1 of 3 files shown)


kmr-srbh (changes requested):
The changes are correct, but I think we should fix this issue in general, through a fix into `visit_expr_list()`. For example, a function call accepting `null()` is a case that will fail even with the current changes.

```fortran
program null_allocatable_component_mre
   implicit none

   print *, func(null())

contains

   integer function func(v)
      integer, intent(in), pointer :: v
      func = 0
   end function
end program null_allocatable_component_mre
```

The approach of a temp val like `ASR::ttype_t* temp_current_variable_type_ = current_variable_type_;` seems correct as we use this approach inside `asr_to_llvm` also.

integration_tests/derived_types_146.f90
```diff
@@ -0,0 +1,16 @@
...
+    implicit none
+
+    type :: box
+        integer, allocatable :: x
+    end type
+
+    type(box) :: a, b
+
+    a = box(null())
+    b = box(x=null())
```
kmr-srbh, line 11: Passing `null()` to `allocatable` is probably not the best thing. Both GFortran and flang warn about this behaviour in other contexts like function calls. Let's use `pointer` instead here.

ivan-pi: It is allowed by the standard though,

> If a component of a derived type is allocatable, the corresponding constructor expression shall be a reference to the intrinsic function NULL with no arguments, an allocatable entity of the same rank, or shall evaluate to an entity of the same rank. If the expression is a reference to the intrinsic function NULL, the corresponding component of the constructor has a status of unallocated.

Page 104 of the PDF (https://j3-fortran.org/doc/year/24/24-007.pdf).

kmr-srbh: @ivan-pi Thanks for this, I was not aware. This change is fine then.

PR #11719, opened 2026-06-02
fix: drop stale assumed-rank entry when entering rank default

HarshitaKalani:


src/lfortran/semantics/ast_body_visitor.cpp
assem2002: I think you can move this up (inside the loop) and drop the changes made above.

assem2002 (approved):
LGTM. Thanks
Please address me changes and feel free to merge.

PR #11720, opened 2026-06-02
fix: set vptr when wrapping derived-type scalar for class(*) assumed-rank dummy

HarshitaKalani:


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -23296,9 +23296,28 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                     ASR::ttype_t* arg_type = ASRUtils::expr_type(x.m_args[i].m_value);
                     ASR::ttype_t* arg_type_unwrapped = ASRUtils::type_get_past_allocatable(
                         ASRUtils::type_get_past_pointer(arg_type));
+                    bool dummy_is_assumed_type = false;
```
assem2002, line 23299: `~assume_type` shouldn't have a `structType` node with member `m_is_unlimited_polymorphic` set to true` , in other words, we shouldn't care about avoiding assumed_type once we got into `is_type_upoly`

assem2002 (approved):
LGTM. Thanks

PR #11726, opened 2026-06-02
fix: handle module-level EQUIVALENCE pointer aliases in semantics and LLVM Codegen

### LINKED ISSUE #11725: Segfault when compiling module level equivalence alias with variable initialization
MRE
```fortran
module mymod
    implicit none
    integer, save, public :: a = 0
    integer, public :: b
    equivalence (a, b)
end module mymod

program main
    use mymod
    implicit none
    b = 42
    print *, "b is", b
end program main
```

LFortran
```console
```

HarshilShah1804:
Fixes #11725


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6253,62 +6284,74 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                                                ASR::ttype_t *cptr = ASRUtils::TYPE(ASR::make_CPtr_t(al, asr_eq2->base.loc));
-                                                ASR::asr_t* pointer_to_cptr = ASR::make_PointerToCPtr_t(
-                                                    al, asr_eq2->base.loc, ASRUtils::EXPR(get_pointer2), cptr, nullptr);
-                                                ASR::asr_t* c_f_pointer = ASR::make_CPtrToPointer_t(
-                                                    al, asr_eq2->base.loc, ASRUtils::EXPR(pointer_to_cptr),
-                                                    asr_eq1, ASRUtils::EXPR(array_constant), nullptr);
 
-                                                ASR::stmt_t *stmt = ASRUtils::STMT(c_f_pointer);
-                                                data_structure[current_scope->counter].push_back(stmt);
+                                                bool is_module = in_module && !in_Subroutine;
```
Jatinagarwal24: Hi @HarshilShah1804 i can see this identical check bool `is_module = in_module && !in_Subroutine` is copy pasted in 4 separate locations within the EQUIVALENCE handling. This should be computed once at the top of the equivalence processing block and reused.
This can minimise the changes.

integration_tests/equivalence_38.f90
```diff
@@ -0,0 +1,14 @@
...
+    equivalence (a, b)
+end module mymod
+
+program main
+    use mymod
+    implicit none
+    b = 42
+    if (a /= 42) error stop 
+    if (b /= 42) error stop
+end program main
```
Jatinagarwal24: This test only tests the simplest case (two integer scalars).
You can add following cases 
Array-to-array equivalence at module scope
Mixed-type equivalence `(real/integer) `at module scope
Chained equivalence `(equivalence(a,b) + equivalence(b,c))` at module scope
Equivalence inside a module subroutine (verifying the runtime path is still taken)

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5471,10 +5474,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            visit_Variable(*target_var);
+                        }
+                        if (llvm_symtab.find(target_h) != llvm_symtab.end()) {
+                            alias_target = llvm::dyn_cast<llvm::Constant>(llvm_symtab[target_h]);
+                            if (ASR::is_a<ASR::ArrayItem_t>(*get_ptr->m_arg)) {
+                                ASR::ArrayItem_t* arr_item = ASR::down_cast<ASR::ArrayItem_t>(get_ptr->m_arg);
+                                if (arr_item->n_args > 0 && arr_item->m_args[0].m_right && ASR::is_a<ASR::IntegerConstant_t>(*arr_item->m_args[0].m_right)) {
+                                    int64_t offset = ASR::down_cast<ASR::IntegerConstant_t>(arr_item->m_args[0].m_right)->m_n;
+                                    offset -= 1; // 1-based to 0-based
+                                    if (offset > 0) {
```
Jatinagarwal24: This condition `if (offset > 0)` skips the GEP instruction for index 0. If a Fortran array has a custom lower bound of 0 (e.g., `integer :: a(0:10))`, the index `0` becomes offset `-1 `and would be skipped here, which is incorrect. Consider changing this to` if (offset != 0) or if (offset >= 0)` (if you expect only 1-based arrays, `!= 0 `is safer if negative offsets occur).

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5471,10 +5474,53 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                        llvm::Constant* indices[] = {
+                                            llvm::ConstantInt::get(context, llvm::APInt(32, 0)),
+                                            llvm::ConstantInt::get(context, llvm::APInt(32, offset))
+                                        };
+                                        llvm::Type* target_ty = llvm_utils->get_type_from_ttype_t_util(x.m_symbolic_value, target_var->m_type, module.get());
+                                        alias_target = llvm::ConstantExpr::getGetElementPtr(target_ty, alias_target, indices);
```
Jatinagarwal24, line 5543: If the `dyn_cast<llvm::Constant>` on line 5499 returns nullptr (e.g., if the symbol table entry isn't a Constant), passing a null `alias_target` into `getGetElementPtr` will cause a crash. It would be safer to wrap this block in `if (alias_target && ASR::is_a<ASR::ArrayItem_t>...)`

src/lfortran/semantics/ast_common_visitor.h
Jatinagarwal24: My suggestion so as to how to reduce the code duplication even further,would be that you can create a tiny helper lambda right next to `add_variable_dependency `that handles setting up the alias vs. generating the `CPtrToPointer` statement. This would shrink the four `if (is_module)` blocks down significantly.

Jatinagarwal24 (approved):
Now looks good to me
Thanks for the fix @HarshilShah1804  .

PR #11731, opened 2026-06-03
fix :Keyword argument resolution for type-bound procedure

### LINKED ISSUE #11730: Semantic error: Keyword arguments fail for type-bound procedure pointer with pass attribute
MRE 
```
program main
    type :: test_type
        procedure(func), pointer :: compute_function
    end type
    abstract interface
        subroutine func(me, x, y)
            import :: test_type
            class(test_type) :: me
            integer :: x, y
        end subroutine
    end interface
    type(test_type) :: me
    call me%compute_function(1, y=2)
end program

```
Lfortran 
```
semantic error: Argument was not specified
  --> examples/expr2.f90:13:5
   |
13 |     call me%compute_function(1, y=2)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 1-th argument not specified for func


```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran examples/expr2.f90 -c
```

Jatinagarwal24:
fixes #11730
Keyword argument resolution was broken for type-bound procedure
pointers using the default pass attribute. The bug was in
process_call_args_and_kwargs where is_proc_pointer_var (always true
for any procedure pointer variable) was passed as the is_nopass
argument to visit_kwargs. This incorrectly treated all procedure
pointer variables as nopass, causing the implicit self argument
passed via % to be counted as an explicit argument. For example,
call me%compute_function(1, y=2) would fail with "1-th argument
not specified for func" because the compiler expected 3 explicit
arguments instead of 2.

Fixed by checking the Variable_t's actual m_pass_attr field to
determine whether the procedure pointer is nopass, instead of
assuming all procedure pointer variables are nopass.

### DIFF
--- a/integration_tests/procedure_pointer_keyword_02.f90
+++ b/integration_tests/procedure_pointer_keyword_02.f90
@@ -1,0 +1,57 @@
+module procedure_pointer_keyword_02_mod
+  implicit none
+
+  type :: obj_type
+    integer :: val = 0
+    procedure(sub_iface), pointer :: compute => null()
+  end type
+
+  abstract interface
+    subroutine sub_iface(me, x, y)
+      import :: obj_type
+      class(obj_type), intent(inout) :: me
+      integer, intent(in) :: x, y
+    end subroutine
+  end interface
+
+contains
+
+  subroutine my_add(me, x, y)
+    class(obj_type), intent(inout) :: me
+    integer, intent(in) :: x, y
+    me%val = x + y
+  end subroutine
+
+  subroutine my_mul(me, x, y)
+    class(obj_type), intent(inout) :: me
+    integer, intent(in) :: x, y
+    me%val = x * y
+  end subroutine
+
+end module
+
+program procedure_pointer_keyword_02
+  use procedure_pointer_keyword_02_mod
+  implicit none
+  type(obj_type) :: obj
+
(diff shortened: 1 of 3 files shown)


Jatinagarwal24:
Hi @assem2002 Pls review this.

integration_tests/procedure_pointer_keyword_02.f90
```diff
@@ -0,0 +1,57 @@
...
+    end subroutine
+  end interface
+
+contains
+
+  subroutine my_add(me, x, y)
+    class(obj_type), intent(inout) :: me
+    integer, intent(in) :: x, y
+    me%val = x + y
+  end subroutine
```
assem2002, line 23:
```.f90
  subroutine my_div(me, x, y)
    class(obj_type), intent(inout) :: me
    integer, intent(in) :: x, y
    me%val = x / y
  end subroutine
me%my_div(y=10, x =100)
if(val /= 10) error stop
```
This test makes sure order is correct.

assem2002 (approved):
LGTM. Thanks for the fix.

integration_tests/procedure_pointer_keyword_02.f90
assem2002: Opened an issue #11790

PR #11739, opened 2026-06-04
fix: internal write to strided character array sections

### LINKED ISSUE #11735: Bug : Formatting with mutliple records (arraysection)
```.f90
program main
  implicit none

  character(2) :: c(5)

  c = "??"
  write(c(5:1:-2), "(A, I0)") "x", 2, "z"
  print "(10A)", c

  c = "??"
  write(c(5:1:-2), "(A,A)") "x", "V", "z"
  print "(10A)", c
end program main
```

```.sh
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt5.f90 && ./a.out 
????z ??x2
????z ??xV
z ?? ??x 
z ??V ??x 
```

Jatinagarwal24:
fixes #11735

Two bugs caused incorrect output for writes like
`write(c(5:1:-2), "(A,I0)") "x", 2, "z"`:

1. The array_op pass split format args one-per-element, creating
   separate writes like `write(c(5),"(A,I0)") "x"` for each arg.
   This is wrong for multi-item formats like (A,I0) where the format
   engine must group multiple args into a single record.

   Fix: materialize a contiguous temporary array, copy-in the original
   values, perform the write to the temp (letting the format engine
   handle record grouping), then copy-out back to the strided positions.

2. The runtime `_lfortran_string_write` padded all remaining array
   records with spaces after writing, even those beyond the output
   list. Per the Fortran standard, untouched records must remain
   unchanged.

   Fix: remove the remaining-record padding loop.

### DIFF
--- a/src/libasr/pass/array_op.cpp
+++ b/src/libasr/pass/array_op.cpp
@@ -722,27 +722,93 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
             if (step == nullptr) {
                 step = make_ConstantWithKind(make_IntegerConstant_t, make_Integer_t, 1, ikind, loc);
             }
-            for (size_t i = 0; i < fmt->n_args; i++) {
-                ASR::expr_t* offset = make_ConstantWithKind(make_IntegerConstant_t, make_Integer_t, i, ikind, loc);
-                ASR::expr_t* scaled_offset = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(
-                    al, loc, offset, ASR::binopType::Mul, step, int_type, nullptr));
-                ASR::expr_t* idx = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(
-                    al, loc, lb_lhs, ASR::binopType::Add, scaled_offset, int_type, nullptr));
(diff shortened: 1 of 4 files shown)


src/libasr/pass/array_op.cpp
assem2002: We operate only when this flag is true -- I think this flag is not related anymore.
Can we remove it and only do the temporary array trick when unit is arraysection?

assem2002: If any troubles found, let's proceed with what we have now

assem2002 (approved):
Great work! 
I like how we handled it using plain array.

assem2002:
Also you can leverage ASRBuilder, it has great utilities you can use to minimize the code.

Jatinagarwal24:
Hi @assem2002 I have addressed your comments 
This commit 1f34d492ee3246a15c58ef8c5fbe1bb2285d79e6  does the following things.
- Remove `is_array_section_unit_with_scalar_values` flag; apply the
  contiguous temp array materialization whenever the write unit is
  an ArraySection, regardless of value types.
- Use ASRBuilder utilities (Variable, Add, Sub, Mul, Div, Assignment,
  DoLoop) to reduce boilerplate in the copy-in/write/copy-out logic.

If anything still i want to correct pls tell.
Thanks !

assem2002 (approved):

PR #11740, opened 2026-06-04
fix :  ICE when calling a subroutine as a function via use-association

### LINKED ISSUE #11737: Bug : using subroutine call as an expression should raise a semtanic error
```.f90
module spread_index
contains
  subroutine hit(deferred_text) 
  end subroutine hit
end module spread_index

program demo
  use spread_index
  print *, hit("A")
end program demo
```
```
  File "/home/assem/lfortran_test/lfortran/src/libasr/asr_utils.h", line 509, in LCompilers::ASRUtils::EXPR2VAR(LCompilers::ASR::expr_t const*)
    ASR::down_cast<ASR::Var_t>(f)->m_v));
AssertFailed: f != nullptr
assem@assem-PC:~/Desktop/fortran/compile_gsnap$ gfortran tt5.f90 
tt5.f90:10:15:

   10 |   print *, hit("A")
      |               1
Error: Unexpected use of subroutine name ‘hit’ at (1)

```

Jatinagarwal24:
Fixes #11737

When a subroutine imported via `use` was called in a function context
(e.g., `print *, hit("A")`), lfortran crashed with an AssertFailed
instead of producing a semantic error. The subroutine-as-function check
in visit_FuncCallOrArray only checked the direct symbol, but
use-associated symbols are wrapped in ExternalSymbol, so the check
never matched and the compiler proceeded to create_Function which
hit an assertion on the null return_var.

Fix: resolve through ExternalSymbol via symbol_get_past_external before
checking for a subroutine (null m_return_var). Also update the error
message to match gfortran's wording: "Unexpected use of subroutine
name 'X'".

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -16527,7 +16527,7 @@ public:
         if (v_past_ext && ASR::is_a<ASR::Function_t>(*v_past_ext)){
             ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(v_past_ext);
             if (func->m_return_var == nullptr){
-                diag.add(Diagnostic("Unexpected use of subroutine name '" + std::string(x.m_func) + "'",
+                diag.add(Diagnostic("Subroutine `" + var_name + "` called as a function",
                                     Level::Error, Stage::Semantic, {Label("", {x.base.base.loc})}));
                 throw SemanticAbort();
             }
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -16523,10 +16523,11 @@ public:
             }
         }
         // if v is a function which has null pointer return type, give error
-        if (ASR::is_a<ASR::Function_t>(*v)){
-            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(v);
+        ASR::symbol_t *v_past_ext = v ? ASRUtils::symbol_get_past_external(v) : nullptr;
+        if (v_past_ext && ASR::is_a<ASR::Function_t>(*v_past_ext)){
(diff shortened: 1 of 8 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -16510,10 +16510,11 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             }
         }
         // if v is a function which has null pointer return type, give error
-        if (ASR::is_a<ASR::Function_t>(*v)){
-            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(v);
+        ASR::symbol_t *v_past_ext = v ? ASRUtils::symbol_get_past_external(v) : nullptr;
+        if (v_past_ext && ASR::is_a<ASR::Function_t>(*v_past_ext)){
+            ASR::Function_t* func = ASR::down_cast<ASR::Function_t>(v_past_ext);
             if (func->m_return_var == nullptr){
-                diag.add(Diagnostic("Subroutine `" + var_name + "` called as a function",
```
HarshitaKalani, line 16529: I think this error message was better, no?

HarshitaKalani (commented):
Otherwise it is good.

src/lfortran/semantics/ast_common_visitor.h
Jatinagarwal24: I changed it to the new message specifically to match gfortran's error message

Jatinagarwal24: If you prefer previous one i will revert this.

Jatinagarwal24:
Hi @HarshitaKalani I think this is good to merge now.

HarshitaKalani (approved):

HarshitaKalani:
Please feel free to merge this after resolving the conflicts. Thanks!

PR #11759, opened 2026-06-05
Fix ICE caused by optional dummy arguments in bind(c) procedure

### LINKED ISSUE #11080: Optional value dummy argument in bind(c) procedure

Jatinagarwal24:
fixes #11080

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -1492,6 +1492,18 @@ public:
                 }
             }
             ASR::symbol_t *var = current_scope->get_symbol(arg_s);
+            if (current_procedure_abi_type == ASR::abiType::BindC && ASR::is_a<ASR::Variable_t>(*var)) {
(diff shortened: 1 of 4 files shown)


kmr-srbh (commented):
I think we already have a PR for this at #11091.

Jatinagarwal24:
Yes this completes that pr with the suggestion applied.

kmr-srbh (approved):
For now, I think we can merge this since the change is small, but please try to search for open PRs before opening new ones.

PR #11760, opened 2026-06-05
Resolve subtle memory bug with realloc and self-referencing

### LINKED ISSUE #9170: Bug : subtle memory bug (realloc + self referencing)
Using `--realloc-lhs-arrays`
```.f90
module m_dep
  type :: downloader_t
     character(len=:), allocatable :: name
     integer :: nn
    end type downloader_t
end module m_dep


program demo
  use m_dep
  type(downloader_t), allocatable :: dl(:)
  allocate(dl(2))
  allocate(character(len=1000000) :: dl(2)%name)
  dl = dl(2:2)
  print *, dl(1)%NAME
end program demo
```

Check how this line `dl = dl(2:2)` gets transformed into the following snippet (after applying all passes)
```.f90
__libasr_created__array_section_pointer_ => dl(2:2)
reallocate(dl(size(__libasr_created__array_section_pointer_, 1)))
__libasr_index_0_1 = lbound(__libasr_created__array_section_pointer_, 1)
__libasr_index_0_ = lbound(dl, 1) - 1
do while (__libasr_index_0_ + 1 <= ubound(dl, 1))
    dl(__libasr_index_0_) = __libasr_created__array_section_pointer_(__libasr_index_0_1)
end do
```

`realloc` will shrink the original memory `dl`, while it still needs to fetch memory from itself -- that memory that just got handed over to the OS. It doesn't raise errors now due to our immediate access to memory after realloc, but you can still observe the wrong memory access using `valgrind`

srinjoy933:
Fixes #9170


srinjoy933:
Hey @assem2002 , please have a look on this pr , Thank You!

srinjoy933:
Modified the visit_Assignment function to safely decouple the RHS evaluation from the LHS reallocation:
1.Split the array variable collection into two separate tracking vectors: target_vars (LHS) and value_vars (RHS).
2.Introduced a get_base_sym utility to drill down through ArrayItem_t and ArraySection_t nodes to find the underlying Var_t symbol for both sides.
3.Implemented an intersection check. If the target's base symbol matches any of the value's base symbols, the assignment is flagged as is_aliased.
4.If an assignment requires reallocation and is aliased, the compiler now generates a temporary variable (aliasing_tmp). It creates a new assignment (aliasing_tmp = dl(2:2)) and recursively passes it through visit_stmt to safely evaluate and buffer the RHS data.
5.The RHS of the original assignment is then swapped to this safe temporary variable. When the compiler finally reaches insert_realloc_for_target, it reallocates the LHS based on the temporary array, completely avoiding the use-after-free condition.

assem2002:
I can see that the original issue is now resolved.
The final pass output shows a temporary fetching the data from the arraysection and saving it.
then the `reallocate` happens on variable `dl` and then assigning into the new `dl` from the temporary that got created.

***
The issue is now resolved -- array_op pass is properly creating a temporary when it encounters an arraysection/arrayitme.
***
Do you think the changes here fix more stuff? @srinjoy933

srinjoy933:
@assem2002 ,I don't think it does more stuff other than solving this issue ,as it is working fine without my changes ,I think i should revert the changes and add the test only.

assem2002 (approved):

PR #11770, opened 2026-06-06
fix: ICE in select rank codegen for pointers and assumed rank subroutine arguments

### LINKED ISSUE #11767: ICE for select rank on assumed rank pointer dummy
## MRE

```fortran
program main
  integer, pointer :: a(:)

  allocate(a(3))
  print *, pointer_rank(a)

contains

  function pointer_rank(x) result(r)
    integer, pointer, intent(in) :: x(..)
    integer :: r

    select rank (x)
    rank (1)
      r = 1
    rank default
      r = -1
    end select
  end function pointer_rank

end program main
```

## output

```text

Assertion failed:
Ty && "Invalid GetElementPtrInst indices for type!"

```

## gfortran behavior

```text

           1
```

Jatinagarwal24:
fixes #11767
This commit resolves two interrelated LLVM codegen bugs causing Internal Compiler Errors (ICE) when dealing with pointer arrays and assumed rank parameters:

1. `visit_SelectRank` Pointer Handling:
Previously, `select rank` codegen relied strictly on `ASRUtils::is_allocatable()` to determine if a target's type needed to be loaded. When passed a `pointer` array instead, the codegen failed to load the underlying descriptor. This caused `arr_descr->get_rank()` to operate on an invalid LLVM pointer, triggering an assertion failure (`llvm::checkGEPType`). The logic was updated to use a more robust `LLVM::is_llvm_pointer()` check and `ASRUtils::type_get_past_allocatable_pointer()` for proper descriptor type extraction.

2. `convert_call_args` Assumed Rank Scalar Pointer Bug:
When passing a scalar `pointer` (e.g., `integer, pointer :: p0`) to an assumed rank subroutine argument `x(..)`, the codegen was incorrectly storing the variable's address (e.g. `i32**`) directly into the descriptor's data field (which expects `i32*`). This caused a type mismatch during LLVM IR construction. We fixed this by correctly checking if the argument is a pointer/allocatable using `LLVM::is_llvm_pointer()` and loading the value to its element type (`CreateLoad2`) before storing it in the descriptor. We also ensured the correct descriptor struct type is calculated using `ASRUtils::type_get_past_allocatable_pointer()`.


HarshitaKalani (approved):

Jatinagarwal24:
Hi @HarshitaKalani pls see if this good to merge?

PR #11771, opened 2026-06-06
fix : semantic error for standalone allocatable/pointer attributes on function return variables

### LINKED ISSUE #11765: allocatable function result attribute not supported
## MRE

```fortran
program main
  print *, value()

contains

  integer function value()
    allocatable :: value

    allocate(value)
    value = 42
  end function value

end program main
```

## output

```text

 --> t.f90:5:9
  |
5 |         allocatable :: value
  |         ^^^^^^^^^^^^^^^^^^^^
```

## gfortran behavior

```text

          42
```

Jatinagarwal24:
fixes #11765 

LFortran was throwing errors when an `allocatable` (or `pointer`) attribute was declared as a standalone attribute statement for a function's return variable (e.g., `integer function value() \n allocatable :: value`), specifically causing collisions when implicit typing was enabled (`--std=f23`).

Because the return variable doesn't fully exist in the symbol table until after `visit_Function` processes the function signature, the standalone `Attribute` node was either failing to resolve the symbol (throwing an unsupported attribute error) or prematurely creating an implicitly typed local variable that caused a "Cannot specify the return type twice" collision later in the visitor.

**Fix:**

1. **Supported Standalone Attributes:** Introduced `assgnd_allocatable` in `ast_common_visitor.h` to track symbols that received a standalone allocatable attribute prior to being fully declared. Updated the `visit_Function` logic to intercept the return variable's type construction and apply `ASR::Allocatable_t` or `ASR::Pointer_t` based on `assgnd_allocatable`/`assgnd_pointer`.
2. **Prevented Return Variable Collision:** Added `current_function_return_var_name` state tracking to explicitly prevent `declare_implicit_variable2` from prematurely generating a local implicitly typed variable for the function's return variable during attribute parsing.
3. **Fixed Implicit Scope Inheritance:** Replaced the unreliable hash-based implicit dictionary mapping with an `implicit_stack`. This cleanly saves and restores `implicit_dictionary` contexts when traversing nested procedures (e.g., internal subroutines), ensuring scoping rules strictly mirror standard Fortran behavior without leaking state.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -701,6 +701,7 @@ RUN(NAME functions_58 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_59 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_60 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME functions_61 LABELS gfortran llvm fortran)
+RUN(NAME functions_62 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME common_01 LABELS gfortran)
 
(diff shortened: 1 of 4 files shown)


Jatinagarwal24:
It will be fixed by pr #11813

PR #11776, opened 2026-06-07
fix: compile-time eval of unpack with scalar field

### LINKED ISSUE #11775: Bug : compile-value of unpack not evaluted
```.f90
program test_unpack_scalar
    implicit none
    logical,parameter :: mask(5)= [.true., .false., .true., .false., .true.]
    integer,parameter :: a(5) = unpack([100, 200, 300], mask, 0)
    if(a(1) /= 100 .or. a(2) /= 0 .or. a(3) /= 200 .or. a(4) /= 0 .or. a(5) /= 300) error stop   
end program
```

ambikeesshh:
fixes #11775

The `unpack` intrinsic failed to evaluate at compile time when `field`
was a scalar (e.g. `unpack([100, 200, 300], mask, 0)` in a `parameter`
declaration). `eval_Unpack` returned nullptr for scalar field, which
works at runtime but breaks constant folding.

The fix makes `populate_vector` and `populate_vector_complex` handle
scalar constants by broadcasting the value across the output vector,
instead of asserting `ArrayConstant_t`. The early `return nullptr` in
`eval_Unpack` is removed since scalars are now handled downstream.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1468,7 +1468,7 @@ RUN(NAME intrinsics_147 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # pack
 RUN(NAME intrinsics_148 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST_MATH) # pack
 RUN(NAME intrinsics_149 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran NO_FAST_MATH) # unpack
 RUN(NAME unpack_scalar_field_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
-RUN(NAME unpack_scalar_field_02 LABELS llvm llvm_wasm llvm_wasm_emcc fortran)
+RUN(NAME unpack_scalar_field_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME intrinsics_150 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # maskr
 RUN(NAME intrinsics_151 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # maskl
 RUN(NAME intrinsics_152 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran) # selected_real_kind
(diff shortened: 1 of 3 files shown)


src/libasr/pass/intrinsic_array_function_registry.h
assem2002:
```suggestion
 lcompilers_assert(ASRUtils::is_value_constant(vector_a))
for (int i = 0; i < dim; i++) {
ASR::expr_t* arg_a {};
if(ASR::is_a<ASR::ArrayConstant>(vector_a)){
    ASR::ArrayConstant_t *a_const = 
        ASR::down_cast<ASR::ArrayConstant_t>(vector_a);
    arg_a = ASRUtils::fetch_ArrayConstant_value(al, a_const, i);
} else {
    arg_a = vector_a;
}
```

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -6461,7 +6472,21 @@ namespace Unpack {
...
+                                static_cast<typename T::first_type>(c_a->m_im)};
+                std::fill(a.begin(), a.end(), scalar_val);
+            } else {
+                LCOMPILERS_ASSERT(false);
+            }
+            return;
+        }
         ASR::ArrayConstant_t *a_const = ASR::down_cast<ASR::ArrayConstant_t>(vector_a);
 
         for (int i = 0; i < dim; i++) {
```
assem2002: Do same as suggested above.

assem2002 (approved):
LGTM. Thanks for the fix.
Please address my changes for concise code.

src/libasr/pass/intrinsic_array_function_registry.h
ambikeesshh: fixed

src/libasr/pass/intrinsic_array_function_registry.h
ambikeesshh: done

ambikeesshh:
> LGTM. Thanks for the fix. Please address my changes for concise code.

thanks for the review!
pushed the fix

integration_tests/CMakeLists.txt
```diff
@@ -1469,6 +1468,7 @@ RUN(NAME intrinsics_147 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc) # pack
 RUN(NAME intrinsics_148 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc NO_FAST_MATH) # pack
 RUN(NAME intrinsics_149 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran NO_FAST_MATH) # unpack
 RUN(NAME unpack_scalar_field_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
+RUN(NAME unpack_scalar_field_02 LABELS llvm llvm_wasm llvm_wasm_emcc fortran)
```
HarshitaKalani: Is there any specific reason to not include the `gfortran` label?

HarshitaKalani: I see, this is because in the added test, the declarations are done after executable statements. I think it is better to add the declarations at the top and add gfortran label here.

HarshitaKalani (commented):
Can you please address the comments? Thanks!

integration_tests/CMakeLists.txt
ambikeesshh: fixed and pushed. thanks!

ambikeesshh:
> Can you please address the comments? Thanks!

done 
thanks for review!

PR #11787, opened 2026-06-08
Fix : Free heap-allocated array descriptors to prevent memory leaks

srinjoy933:
Fixes#10759
This PR fixes the memory leak  where heap-allocated array descriptors were orphaned during variable deallocation and pointer reassignment. 
To resolve this, the LLVM IR generation in asr_to_llvm.cpp was updated to explicitly free these structs at their termination points. 

visit_Deallocate: Captures the descriptor pointer for struct members, frees the underlying array data and the descriptor struct itself, then safely nullifies the pointer.
handle_pointer_section_target: Adds a null-check for non-local pointers to explicitly free any existing heap-allocated descriptor before reassigning and allocating a new one.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -2695,7 +2695,6 @@ public:
             ASR::symbol_t* curr_obj = nullptr;
             ASR::abiType abt = ASR::abiType::Source;
             
-            // --- FIX: Track address for nullification ---
             llvm::Value* address_to_nullify = nullptr;
 
             if( ASR::is_a<ASR::Var_t>(*tmp_expr) ) {
@@ -2806,6 +2805,7 @@ public:
                         tmp = llvm_utils->CreateLoad2(typ, tmp);
                     }
                     if (ASRUtils::is_class_type(ASRUtils::extract_type(cur_type))) {
+                        // If it is a class type, we need to get the pointer to the struct
                         llvm::Type* class_type = llvm_utils->get_type_from_ttype_t_util(
                             tmp_expr,
                             ASRUtils::type_get_past_pointer(ASRUtils::type_get_past_allocatable(cur_type)),
@@ -2825,6 +2825,7 @@ public:
                             llvm::ConstantPointerNull::get(llvm_data_type->getPointerTo()),
                             llvm::Type::getInt64Ty(context)) );
                     llvm_utils->create_if_else(cond, [=]() {
+                        // Call user-defined FINAL procedures (Fortran 2018 §7.5.6.3)
                         if (struct_sym != nullptr && struct_sym->n_member_functions > 0) {
                             for (size_t fi = 0; fi < struct_sym->n_member_functions; fi++) {
                                 std::string final_proc_name = struct_sym->m_member_functions[fi];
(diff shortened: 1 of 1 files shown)


assem2002:
As I understand, This should fix a leak.
If it fixes a leak, it'd be great to have a test showing that.
if it doesn't and and straightens things up (refactor), please explain further.

***

As I can remember the linked issue was opened when we had problems with leaks, now we have solid finalization-process and leak detection. Maybe the bugs were already fixed.

Check `LLVMFinalize` class at `llvm_utils.h` -- it properly cleans things, and if you want to modify things related to cleaning memory, it should be done there as I think. 

*** 
@srinjoy933 Do we have tests that show leak when we `--detect-leaks` flag?

srinjoy933:
Hey @assem2002 ,I explicitly don't have any tests that show leak when we --detect-leaks flag ,but except that, this pr also does some refactors.
Regarding the refactor: While LLVMFinalize handles scope-exit cleanup, it cannot capture mid-loop pointer reassignments that orphan descriptors before the scope ends. This PR refactors the IR generation to align descriptor lifetimes with data buffer lifecycles: I now explicitly free() descriptors during visit_Deallocate and intercept assignments in handle_pointer_section_target to free() existing descriptors before overwriting them. Please note that I lack an explicit test case for --detect-leaks, as that tool only tracks user-payload allocations and remains blind to these internal, compiler-generated descriptor structs. This refactor ensures consistent memory management at the IR level, preventing descriptor loss during AST operations.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -2869,23 +2876,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+
+                        if (in_struct && address_to_nullify && is_implicit_dealloc) {
+#if LLVM_VERSION_MAJOR < 15
+                            llvm::Value* desc_i8 = builder->CreateBitCast(tmp, llvm::Type::getInt8PtrTy(context));
+#else
+                            llvm::Value* desc_i8 = tmp; // Opaque pointers in LLVM 15+
+#endif
+                            builder->CreateCall(free_fn, {allocator, desc_i8});
+                            llvm::Type* desc_ptr_type = tmp->getType();
+                            builder->CreateStore(
```
assem2002, line 2905: I don't think this is needed. We already use `reset_is_allocated_flag` do that

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9363,6 +9384,32 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                if (is_struct_member) {
+                    llvm::Function* free_fn = llvm_utils->_Deallocate();
+                    llvm::Value* allocator = llvm_utils->get_allocator(module.get());
+                    
+#if LLVM_VERSION_MAJOR < 15
+                    llvm::Value* desc_i8 = builder->CreateBitCast(old_desc, llvm::Type::getInt8PtrTy(context));
+#else
+                    llvm::Value* desc_i8 = old_desc; // Opaque pointers in LLVM 15+
+#endif
+                    builder->CreateCall(free_fn, {allocator, desc_i8});
```
assem2002, line 9409: I think this branch is dead code (I tested it locally), this branch is when the variable is an array returnVar variable, We convert those variable into `intent(out)` argument and pass it at caller site.

if we happen to go into that branch somehow, things are gonna leak anyways because we shouldn't have gone there

PR #11797, opened 2026-06-09
feat: multiple namelist group declarations on one statement

### LINKED ISSUE #11786: parser rejects multiple namelist group declarations on one statement
## MRE

```fortran
program namelist
  implicit none

  integer :: x, y

  namelist /left/ x /right/ y

  x = 1
  y = 2

  print *, x, y
end program namelist
```

## output

```text

 --> t.f90:4:32
  |
4 |     namelist /left/ x /right/ y
  |                                ^
```

## gfortran behavior

```text

           1           2
```

opixdown:
fixes #11786


HarshitaKalani (approved):

Jatinagarwal24 (approved):
Looks good to me.
Thanks !

assem2002:
I don't have proper knowledge with bison. @certik Can you please help here ?

src/lfortran/parser/parser.yy
```diff
@@ -1716,6 +1740,7 @@ sep_one
 
 decl_statements
     : decl_statements decl_statement { $$ = $1; LIST_ADD($$, $2); }
+    | decl_statements namelist_decl { $$ = $1; LIST_CONCAT($$, $2); }
```
certik, line 1743: This doesn't seem right. The `decl_statements` is just a list of statements, but now we are hardwiring there some namelist support. We should figure out how to make it work in `decl_statement` somehow.

src/lfortran/parser/parser.yy
```diff
@@ -1351,6 +1355,7 @@ use_modifier
 // var_decl*
 var_decl_star
     : var_decl_star var_decl { $$ = $1; LIST_ADD($$, $2); }
+    | var_decl_star namelist_decl { $$ = $1; LIST_CONCAT($$, $2); }
```
certik, line 1358: This also doesn't seem right.

src/lfortran/parser/parser.yy
```diff
@@ -1115,6 +1118,7 @@ temp_decl
 
 decl_star
     : decl_star decl { $$ = $1; LIST_ADD($$, $2); }
+    | decl_star namelist_decl { $$ = $1; LIST_CONCAT($$, $2); }
```
certik, line 1121: These two also don't seem right.

certik:
See how the common block is handled, I think it has exactly the same problem of needing to parse multiple on the same line. We need to cleanly parse it as a declaration, not hack it into the "list of declarations" as a special case.

PR #9445, opened 2026-01-10
fix: avoid present() ICE for optional procedure dummies

krystophny:
## Summary
- Fix assertion failure when evaluating `present()` on optional procedure dummies
- Fix codegen crashes when passing procedure arguments through implicit interface calls (scipy quadpack pattern)
- Implicit interfaces now correctly include argument types from call site

Fixes #9418

## Why
Optional procedure dummies (declared with `external + optional`) caused an ICE in `present()` evaluation. Additionally, passing external procedure arguments through implicit interface subroutine calls (common in scipy quadpack) crashed during LLVM codegen due to null `type_declaration` and out-of-bounds array access.

**Stage:** Semantics + Codegen

## Changes
- [`src/lfortran/semantics/ast_common_visitor.h#L13164-L13225`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/src/lfortran/semantics/ast_common_visitor.h#L13164-L13225): Create interface for procedure variable passed as argument
- [`src/lfortran/semantics/ast_body_visitor.cpp#L5265-L5279`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/src/lfortran/semantics/ast_body_visitor.cpp#L5265-L5279): Populate interface args when calling procedure variables
- [`src/libasr/codegen/llvm_utils.cpp#L904-L929`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/src/libasr/codegen/llvm_utils.cpp#L904-L929): Handle null `type_declaration` for FunctionType in codegen
- [`src/libasr/codegen/asr_to_llvm.cpp`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/src/libasr/codegen/asr_to_llvm.cpp): Add null checks for `m_type_declaration` and bounds checks for `m_args` access
- `tests/reference/asr-external4-*`: Updated reference (interface now has proper args)

## Tests
- [`integration_tests/present_04.f90`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/integration_tests/present_04.f90): `present()` with optional external procedure
- [`integration_tests/present_05.f90`](https://github.com/lfortran/lfortran/blob/6aad1733713819a72830c323af0dd7e6ecabc09a/integration_tests/present_05.f90): Optional procedure with interface block
[...]

PR #11806, opened 2026-06-10
fix: LLVM verification errors for date_and_time and procedure pointers

### LINKED ISSUE #11805: LLVM codegen for date_and_time 64-bit arrays and implicit-interface procedure pointers
MRE
```
module m_compute
    implicit none
contains
    subroutine compute(n)
        class(*), intent(in) :: n
        select type(n)
        type is (integer)
            print *, 'value:', n
        end select
    end subroutine
end module

module m
    implicit none
    type :: method
        private
        integer :: nargs
        procedure(), nopass, pointer :: f => null()
    end type
    interface method
        module procedure :: method_create
    end interface
contains
    function method_create(f, a1) result(that)
        procedure() :: f
        class(*), intent(in) :: a1
        type(method) :: that
        that%f => f
        that%nargs = 1
    end function
end module

program test
    use m
    use m_compute
    implicit none
    integer, parameter :: i8 = selected_int_kind(18)
    type(method) :: mtd
    integer(i8) :: values(8)

    mtd = method(compute, 42)
    print *, values
end program


```
LFortran 
```
  store i32 %26, i64* %25, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %44, i64* %43, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %62, i64* %61, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %80, i64* %79, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %98, i64* %97, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %116, i64* %115, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %134, i64* %133, align 4
 i64Stored value type does not match pointer operand type!
  store i32 %152, i64* %151, align 4
 i64Call parameter type does not match function signature!
void (%"~unlimited_polymorphic_type"*)* @__module_m_compute_compute
 void ()*  call void @__module_m_method_create(void (%"~unlimited_polymorphic_type"*)* @__module_m_compute_compute, %"~unlimited_polymorphic_type"* %3, %m.method.6* %mtd)
````
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran examples/expr2.f90 
                 2026                    6                   10                  330                   18                   39                   23                  807
```

### PR DESCRIPTION
**Pull request by Jatinagarwal24**
fixes #11805

This commit resolves two distinct LLVM IR verification failures encountered when building `benchmark.f`:

1. Fix `date_and_time` intrinsic i32 to i64 type mismatch:
   - Bug: When the user provided an `integer(8)` (64-bit) array for the `values` argument, the generated LLVM IR attempted to directly store 32-bit integer values returned by the runtime into the 64-bit array constant, causing a `Stored value type does not match pointer operand type!` LLVM verification error.
   - Fix: Updated `instantiate_DateAndTime` in `src/libasr/pass/intrinsic_subroutines.h` to explicitly cast the runtime `i32` values to the target array's element type before populating the `ArrayConstant`.

2. Fix implicit-interface procedure pointer codegen:
   - Bug: When passing a procedure to a struct constructor containing an implicit-interface procedure pointer field, the generic resolution could select the wrong interface or generate a mismatched LLVM function type signature, resulting in a `Call parameter type does not match function signature!` error.
- fix : wip

### DIFF
--- a/src/lfortran/semantics/asr_implicit_cast_rules.h
+++ b/src/lfortran/semantics/asr_implicit_cast_rules.h
@@ -114,10 +114,23 @@ public:
                                   ASR::ttype_t *source_type,
                                   ASR::ttype_t *dest_type, diag::Diagnostics &diag) {
       if ((ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(source_type))
-           || ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(dest_type)))
-          || ((ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(source_type))
-               || ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(dest_type))))) {
-          // No casting supported currently for `StructType` and `FunctionType`
+           || ASR::is_a<ASR::StructType_t>(*ASRUtils::extract_type(dest_type)))) {
+          // No casting supported currently for `StructType`
+          return;
+      }
+
+      if (ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(source_type))
+          && ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(dest_type))) {
+          // FunctionType-to-FunctionType reconciliation is handled in
+          // Call_t_body for call arguments.  Do not insert a cast here
+          // because structural comparison of polymorphic class arguments
+          // can give false negatives (anonymous struct vs named class type),
+          // leading to spurious casts and LLVM type mismatches.
+          return;
+      }
+
+      if ((ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(source_type))
+           || ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(dest_type)))) {
           return;
       }
       if (ASRUtils::types_equal(source_type, dest_type, nullptr, nullptr, true)) {
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -4878,22 +4878,19 @@ public:
             } else if (compiler_options.implicit_typing) {
                 type = implicit_dictionary[std::string(1,sym[0])];
                 if (!type) {
-                    // There exists an `implicit none` statement, here compiler has
-                    // no information about type of symbol hence keeping it real*4.
-                    type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, 4));
+                    is_subroutine = true;
                 }
             } else {
-                // Here compiler has no information about type of symbol hence keeping it real*4.
-                type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, 4));
+                is_subroutine = true;
             }
             // add return var
             std::string return_var_name = sym + "_return_var_name";
             SetChar variable_dependencies_vec;
             variable_dependencies_vec.reserve(al, 1);
-            ASRUtils::collect_variable_dependencies(al, variable_dependencies_vec, type);
             ASR::asr_t *return_var = nullptr;
             ASR::expr_t *to_return = nullptr;
             if (!is_subroutine) {
+                ASRUtils::collect_variable_dependencies(al, variable_dependencies_vec, type);
                 return_var = ASRUtils::make_Variable_t_util(al, loc,
                     current_scope, s2c(al, return_var_name), variable_dependencies_vec.p,
                     variable_dependencies_vec.size(), ASRUtils::intent_return_var,
@@ -7527,7 +7524,27 @@ public:
                     LCOMPILERS_ASSERT( sym_ != nullptr );
                     // set function return type as `type`
                     ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(sym_);
-                    ASRUtils::EXPR2VAR(f->m_return_var)->m_type = type;
+                    if (f->m_return_var == nullptr) {
+                        std::string return_var_name = sym + "_return_var_name";
+                        SetChar variable_dependencies_vec;
+                        variable_dependencies_vec.reserve(al, 1);
+                        ASRUtils::collect_variable_dependencies(al, variable_dependencies_vec, type);
+                        ASR::asr_t *return_var = ASRUtils::make_Variable_t_util(al, x.base.base.loc,
+                            current_scope, s2c(al, return_var_name), variable_dependencies_vec.p,
+                            variable_dependencies_vec.size(), ASRUtils::intent_return_var,
+                            nullptr, nullptr, ASR::storage_typeType::Default, type, nullptr,
+                            ASR::abiType::BindC, ASR::Public, ASR::presenceType::Required,
+                            false);
+                        current_scope->add_symbol(return_var_name, ASR::down_cast<ASR::symbol_t>(return_var));
+                        f->m_return_var = ASRUtils::EXPR(ASR::make_Var_t(al, x.base.base.loc,
+                            ASR::down_cast<ASR::symbol_t>(return_var)));
+                        ASR::FunctionType_t *ft = ASR::down_cast<ASR::FunctionType_t>(f->m_function_signature);
+                        ft->m_return_var_type = type;
+                    } else {
+                        ASRUtils::EXPR2VAR(f->m_return_var)->m_type = type;
+                        ASR::FunctionType_t *ft = ASR::down_cast<ASR::FunctionType_t>(f->m_function_signature);
+                        ft->m_return_var_type = type;
+                    }
                 }
                 current_variable_type_ = type;
                 if (is_argument && current_scope->asr_owner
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: src/libasr/codegen/asr_to_llvm.cpp [hunk 1]
```diff
@@ -22916,6 +22916,19 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            orig_arg_intent != ASR::intentType::Out &&
+                            !ASRUtils::is_pointer(orig_arg->m_type)) {
+                        llvm::Type* expected_type = llvm_utils->get_type_from_ttype_t_util(
+                            ASRUtils::EXPR(ASR::make_Var_t(al, orig_arg->base.base.loc, &orig_arg->base)),
+                            orig_arg->m_type, module.get());
+                        if (tmp->getType() != expected_type) {
+                            tmp = builder->CreateBitCast(tmp, expected_type);
+                        }
+                    }
+
```
kmr-srbh: This is not the correct way to do this. The type semantics should match through the ASR. Please see https://github.com/lfortran/lfortran/pull/9445/changes#r2692856278.

Also, since we are encountering this again, it might be worth fixing this correctly.

src/libasr/pass/intrinsic_subroutines.h
kmr-srbh, line 1: @HarshitaKalani could you please review this intrinsic code changes?

kmr-srbh (dismissed):

src/libasr/codegen/asr_to_llvm.cpp
certik: I agree, this doesn't seem right. We should not be querying the LLVM for a type and then just casting it, without understanding.

We should know in ASR what is going on and just generate the right code directly.

Jatinagarwal24: Hi @kmr-srbh @certik I am able to remove this and also similar kind of this bitcasting from above

HarshitaKalani (approved):
This looks good now. Thanks!

src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -114,10 +114,23 @@ class ImplicitCastRules {
...
+          // FunctionType-to-FunctionType reconciliation is handled in
+          // Call_t_body for call arguments.  Do not insert a cast here
+          // because structural comparison of polymorphic class arguments
+          // can give false negatives (anonymous struct vs named class type),
+          // leading to spurious casts and LLVM type mismatches.
+          return;
+      }
+
+      if ((ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(source_type))
+           || ASR::is_a<ASR::FunctionType_t>(*ASRUtils::extract_type(dest_type)))) {
```
assem2002, line 133: That's fine. Comments describing the reason is good.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -4835,22 +4835,19 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             } else if (compiler_options.implicit_typing) {
                 type = implicit_dictionary[std::string(1,sym[0])];
                 if (!type) {
-                    // There exists an `implicit none` statement, here compiler has
-                    // no information about type of symbol hence keeping it real*4.
-                    type = ASRUtils::TYPE(ASR::make_Real_t(al, loc, 4));
+                    is_subroutine = true;
```
assem2002, line 4881: Why is that needed?
I checked where this function (create_external_function) is being used, and I think it's unrelated to what we're fixing.

assem2002: Okay I think it's fine. it's still unrelated but fine change

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7433,7 +7427,27 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     LCOMPILERS_ASSERT( sym_ != nullptr );
                     // set function return type as `type`
                     ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(sym_);
-                    ASRUtils::EXPR2VAR(f->m_return_var)->m_type = type;
+                    if (f->m_return_var == nullptr) {
```
assem2002, line 7527: I think that's a consequence of the change above, I think it's fine.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -22951,6 +22942,34 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                             }
                         }
                     }
+#if LLVM_VERSION_MAJOR < 15
```
assem2002, line 23244: This should be an ASR casting node (functionTypeTofunctionType).
***
It looks fine for now, but the backend will do the heavy lifting, when we have 2 different functionTypes. I can't count down the number of cases where we'd need such a cast; but I think the cast is 100% needed

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -23000,6 +23005,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    // Must happen BEFORE alloca wrapping to preserve the
+                    // correct pointer indirection level.
+                    if (orig_arg && orig_arg->m_type_declaration &&
+                            ASR::is_a<ASR::FunctionType_t>(
+                            *ASRUtils::type_get_past_pointer(orig_arg->m_type))) {
+                        llvm::Type* expected_type = llvm_utils->get_type_from_ttype_t_util(
+                            ASRUtils::EXPR(ASR::make_Var_t(al, orig_arg->base.base.loc, &orig_arg->base)),
+                            ASRUtils::type_get_past_pointer(orig_arg->m_type), module.get());
+                        if (tmp->getType() != expected_type) {
+                            tmp = builder->CreateBitCast(tmp, expected_type);
```
assem2002, line 23320: Yes here also :) -- 2 spots for same logic

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1430,7 +1430,12 @@ namespace LCompilers {
                     throw CodeGenError("Type not implemented " + ASRUtils::type_to_str_python_expr(return_var_type0, x.m_return_var));
             }
         } else {
-            return_type = llvm::Type::getVoidTy(context);
+            ASR::FunctionType_t *func_type = ASR::down_cast<ASR::FunctionType_t>(x.m_function_signature);
+            if (func_type->m_return_var_type != nullptr) {
+                return_type = get_type_from_ttype_t_util(nullptr, func_type->m_return_var_type, module);
+            } else {
+                return_type = llvm::Type::getVoidTy(context);
+            }
```
assem2002: This is probably bypassing some failing test.
ASR::function's `m_return_var` should reflect what ASR::FunctionType's `m_return_var_type` member -- In other words they either should be nullptr or !nullptr.
***
Please try to find the failing test (I assume there's a failing test), and find why there's mismatch, and try to fix that.

src/libasr/asr_utils.cpp
```diff
@@ -2978,9 +2978,12 @@ bool argument_types_match(const Vec<ASR::call_arg_t>& args,
 
                 // Check if actual argument is an implicit interface procedure (e.g. external :: f)
                 // If it is, it is compatible with any explicit interface formal argument.
-                if (ASR::is_a<ASR::FunctionType_t>(*arg1)) {
+                // Similarly, if formal argument is an implicit interface, it accepts an explicit interface actual argument.
+                if (ASR::is_a<ASR::FunctionType_t>(*arg1) && ASR::is_a<ASR::FunctionType_t>(*arg2)) {
                     ASR::FunctionType_t* arg1_func_type = ASR::down_cast<ASR::FunctionType_t>(arg1);
-                    if (arg1_func_type->n_arg_types == 0 && arg1_func_type->m_deftype == ASR::deftypeType::Interface) {
+                    ASR::FunctionType_t* arg2_func_type = ASR::down_cast<ASR::FunctionType_t>(arg2);
```
assem2002, line 3039: This shouldn't be the case with proper casting as discussed above.

src/lfortran/semantics/ast_common_visitor.h
assem2002: Maybe good consequence

assem2002 (commented):
I think it looks fine for now!
***
Please address my comments, I think there's 1 or 2 comments that require attention.
***
I think we still need refactoring on 2 things : 
1- How we create interfaces and manage the modifications whilst we build ASR.
2- How we handle `procedure()` parameter/argument and the casting.

Things are completely scattered, no separate functions, no obvious pipeline on how implicit interfaces is created/modified, no ASR verification on functions.

src/libasr/codegen/llvm_utils.cpp
assem2002: I think this is the only change needed

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -23289,6 +23304,23 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         tmp = llvm_symtab_fn_arg[h];
                         LCOMPILERS_ASSERT(tmp != nullptr)
                     }
+#if LLVM_VERSION_MAJOR < 15
```
assem2002, line 23307: just a minor note, using if/else predecessors isn't needed, if we came across similar case in the future, let's avoid using it as it attracts reader's attention for no reason.

assem2002 (approved):
Looks good. Thanks. 
We should come back to this PR later when we have proper functiontypeTOfunctiontype casting in ASR.

PR #11812, opened 2026-06-10
Improve : Make lfortran reproducible-build friendly

### LINKED ISSUE #11428: Reproducibility problem: symbol name changes

assem2002:
fixes #11428
- Change `get_unique_id` which wass used to return a unique id per lfortran invokation (random), into a function that reutrn a unique id per source file path name (consistent from build-to-build with same env)
- now the rebuilds for the same file is consistent not random.

### DIFF
--- a/src/libasr/utils2.cpp
+++ b/src/libasr/utils2.cpp
@@ -14,18 +14,13 @@
 #include <libasr/string_utils.h>
 
 namespace LCompilers {
-
-std::string get_unique_ID() {
-    static std::random_device dev;
-    static std::mt19937 rng(dev());
-    std::uniform_int_distribution<int> dist(0, 61);
-    const std::string v =
-        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
-    std::string res;
-    for (int i = 0; i < 22; i++) {
-        res += v[dist(rng)];
-    }
-    return res;
+    
+// Get a unique ID for the given file path.
+// Used for separate compilation symbol mangling.
+// Makes sure symbol mangling is stable across different runs.
+// helpful for reproducible builds
+std::string get_unique_ID(const std::string &file_path) {
+    return std::to_string(std::hash<std::string>{}(file_path));
 }
 
 bool present(Vec<char*> &v, const char* name) {
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -2090,11 +2090,10 @@ int link_executable(const std::vector<std::string> &infiles,
                 if (backend == Backend::llvm &&
                         compiler_options.po.enable_gpu_offloading &&
                         std::regex_match(s, std::regex(R"(.*\.tmp_\w+\.o)"))) {
-                    std::string file_path = std::filesystem::path(s.substr(0, s.size() - 2)).string();    // strip ".o" from end
-                    std::string mlir_tmp_o = std::filesystem::path(file_path).replace_extension(
-                        ".mlir.tmp_" + LCompilers::get_unique_ID(s) + ".o").string();
-                    compile_cmd += mlir_tmp_o + " ";
-                    mlir_temp_object_files.push_back(mlir_tmp_o);
+                        std::string obj_file_path = std::filesystem::path(s).string();// "do_concurrent_01.tmp_yDR0nDW17GGvAia86INkc2.o"
+                        obj_file_path.insert(obj_file_path.find('.'), ".mlir");
+                        compile_cmd += obj_file_path + " ";
+                        mlir_temp_object_files.push_back(obj_file_path);
                 }
             }
             if(!extra_library_flags.empty()) {
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -120,7 +120,6 @@ std::string LFORTRAN_TEMP_DIR = get_system_temp_dir();
 
 // The unique compilation ID for this invocation of the compiler.
 // Used in naming unique intermediate object files during both compilation modes.
-const std::string LCOMPILERS_UNIQUE_ID = LCompilers::get_unique_ID();
 
 void print_one_component(std::string component) {
     std::istringstream ss(component);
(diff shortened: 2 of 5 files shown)


assem2002:
maybe it's a good to have a test for that in our CI

src/libasr/utils2.cpp
```diff
@@ -14,18 +14,13 @@
...
-        res += v[dist(rng)];
-    }
-    return res;
+    
+// Get a unique ID for the given file path.
+// Used for separate compilation symbol mangling.
+// Makes sure symbol mangling is stable across different runs.
+// helpful for reproducible builds
+std::string get_unique_ID(const std::string &file_path) {
+    return std::to_string(std::hash<std::string>{}(file_path));
```
assem2002, line 23: Should we make sure we have the full_path?

kmr-srbh (commented):
I think we can implement this in a different way? The current approach of hashing the file path does not work well with parallel executions actually. That was the original reason we added unique hashes.

> test.f90

```fortran
program main
   implicit none
   print *, "Hello, World!"
end program main
```

> test_compiler.sh

```bash
seq 1 10 | parallel -j 10 "
    mkdir -p compilertest_{};
    cd compilertest_{};
    # change path below
    cp /Users/saurabhkumar/OpenSource/lfortran/test.f90 .;
    lfortran test.f90
"
```

```console
(lf) saurabhkumar@Mac lfortran % ./test_compiler.sh
Hello, World!
Hello, World!
Hello, World!
Hello, World!
Hello, World!
Hello, World!
ld: file cannot be open()ed, errno=2 (No such file or directory) path=/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o in '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
Hello, World!
Hello, World!
Hello, World!
(lf) saurabhkumar@Mac lfortran % ./test_compiler.sh
ld: file is empty in '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
Hello, World!
Hello, World!
Hello, World!
ld: file cannot be open()ed, errno=2 (No such file or directory) path=/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o in '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
Hello, World!
Hello, World!
ld: file cannot be open()ed, errno=2 (No such file or directory) path=/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o in '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
ld: file cannot be open()ed, errno=2 (No such file or directory) path=/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o in '/var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o'
The command 'clang -o test.out /var/folders/f7/1xrqkz0d2t10r5dftx8zyfzm0000gn/T/test.tmp_1808878356599694219.o  -L"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -Wl,-rpath,"/Users/saurabhkumar/OpenSource/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
Hello, World!
```


Another thing is that the `lcompilers_unique_ID_separate_compilation` variable is used inside function `get_unique_name()` for naming symbols in separate compilation, so symbol name collisions may occur in some case while linking. I did not check this though.

assem2002:
@kmr-srbh 
That's nice observation.
You pointed out to 2 separate modification concerns here.
1. object file name (when not specified) now has unique id = <hash_of_inputfile_name>, instead of unique id = random hash per compiler instance. Example -> `testfile_313123123123.o`
2. Some symbols in the symbol table gets a suffix of unique id = <hash_of_inputfile_name>, instead of unique id = random hash per compiler instance. `lcompilers_sin_i32_253455323123.o`

 ***
For issue 1 -> We can revert things and use a unique id per compiler invokation. 
For issue 2 -> I can think of very rare uses cases that could be problematic with the new hashing way. and we can fix that in the future with more robust-efficient way in the future. Do you have serious case where hashing with symbols would be problematic ?

kmr-srbh:
For case 2, I think the issue can occur when symbols of same name, in files of same name, but located inside different directories are compiled separately, with the compilation directory set to the directory holding the file.

For example when:
`./abc/def/test.f90` and `efgh/test.f90` are compiled separately with the compilation directory set to `def` and `efgh` respectively. In this case, the `get_unique_ID()` function receives path as `test.f90`, so the hashes become same, which can then affect the symbol names, later causing linker symbol collisions.

src/libasr/utils2.cpp
certik: Yes, let's use a full path.

PR #11813, opened 2026-06-10
Support standalone allocatable attribute declarations

### LINKED ISSUE #11765: allocatable function result attribute not supported

srinjoy933:
Fixes #11765 
This PR resolves semantic and implicit typing collisions (--std=f23) caused when an allocatable or pointer attribute is declared as a standalone statement for a function's return variable. Previously, because the return variable isn't fully registered in the symbol table until visit_Function processes the signature, these standalone attributes would either fail to resolve or trigger premature implicit variable creation, resulting in a "Cannot specify the return type twice" error. To fix this, assgnd_allocatable tracking was introduced in ast_common_visitor.h to properly apply Allocatable_t/Pointer_t wrappers during return type construction. Additionally, current_function_return_var_name state tracking was added to block premature implicit typing of the return variable, and the hash-based implicit dictionary was replaced with a robust implicit_stack to ensure clean, standard-compliant state restoration across nested procedures.

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -5813,16 +5813,30 @@ public:
                                         assgnd_pointer.insert(sym);
                                     }
                                 } else if (sa->m_attr == AST::simple_attributeType::AttrAllocatable) {
-                                    ASR::symbol_t* sym_ = current_scope->get_symbol(sym);
+                                    // Use resolve_symbol to find the function in parent scopes if necessary
+                                    ASR::symbol_t* sym_ = current_scope->resolve_symbol(sym);
                                     if (sym_) {
                                         ASR::symbol_t* sym_past_external =
                                             ASRUtils::symbol_get_past_external(sym_);
+                                        
                                         if (ASR::is_a<ASR::Variable_t>(*sym_past_external)) {
+                                            // Standard variable case
                                             ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(
                                                 sym_past_external);
                                             if (!ASRUtils::is_allocatable(v->m_type)) {
                                                 v->m_type = ASRUtils::TYPE(
                                                     ASR::make_Allocatable_t(al, x.base.base.loc, v->m_type));
+                                            }
+                                        } else if (ASR::is_a<ASR::Function_t>(*sym_past_external)) {
+                                            // Function result case: grab the internal return variable
+                                            ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(sym_past_external);
+                                            if (f->m_return_var) {
+                                                ASR::symbol_t* ret_sym = ASR::down_cast<ASR::Var_t>(f->m_return_var)->m_v;
+                                                ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(ret_sym);
+                                                if (!ASRUtils::is_allocatable(v->m_type)) {
+                                                    v->m_type = ASRUtils::TYPE(
+                                                        ASR::make_Allocatable_t(al, x.base.base.loc, v->m_type));
+                                                }
                                             }
                                         }
                                     } else {
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -5813,45 +5813,13 @@ public:
                                         assgnd_pointer.insert(sym);
                                     }
                                 } else if (sa->m_attr == AST::simple_attributeType::AttrAllocatable) {
-                                    ASR::symbol_t* sym_ = current_scope->resolve_symbol(sym);
+                                    ASR::symbol_t* sym_ = current_scope->get_symbol(sym);
                                     if (sym_) {
-                                        ASR::symbol_t* sym_past = ASRUtils::symbol_get_past_external(sym_);
-                                        
-                                        // Case 1: Symbol resolved to the local Variable
-                                        if (ASR::is_a<ASR::Variable_t>(*sym_past)) {
-                                            ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym_past);
+                                        ASR::symbol_t* sym_past_external = ASRUtils::symbol_get_past_external(sym_);
+                                        if (ASR::is_a<ASR::Variable_t>(*sym_past_external)) {
+                                            ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym_past_external);
                                             if (!ASRUtils::is_allocatable(v->m_type)) {
                                                 v->m_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, v->m_type));
-                                            }
-                                            
-                                            // Prevent reverting: update the function signature if this is the return variable!
-                                            if (current_scope->asr_owner && ASR::is_a<ASR::symbol_t>(*current_scope->asr_owner)) {
-                                                ASR::symbol_t* owner_sym = ASR::down_cast<ASR::symbol_t>(current_scope->asr_owner);
-                                                if (ASR::is_a<ASR::Function_t>(*owner_sym)) {
-                                                    ASR::Function_t* fn = ASR::down_cast<ASR::Function_t>(owner_sym);
-                                                    if (fn->m_return_var && ASR::down_cast<ASR::Var_t>(fn->m_return_var)->m_v == sym_past) {
-                                                        ASR::FunctionType_t *f_type = ASR::down_cast<ASR::FunctionType_t>(fn->m_function_signature);
-                                                        if (f_type->m_return_var_type && !ASRUtils::is_allocatable(f_type->m_return_var_type)) {
-                                                            f_type->m_return_var_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, f_type->m_return_var_type));
-                                                        }
-                                                    }
-                                                }
-                                            }
-                                        } 
-                                        // Case 2: Symbol resolved directly to the Function (from parent scope)
-                                        else if (ASR::is_a<ASR::Function_t>(*sym_past)) {
-                                            ASR::Function_t *f = ASR::down_cast<ASR::Function_t>(sym_past);
-                                            if (f->m_return_var) {
-                                                ASR::symbol_t* ret_sym = ASR::down_cast<ASR::Var_t>(f->m_return_var)->m_v;
-                                                ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(ret_sym);
-                                                if (!ASRUtils::is_allocatable(v->m_type)) {
-                                                    v->m_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, v->m_type));
-                                                }
-                                            }
-                                            // Prevent reverting here too
-                                            ASR::FunctionType_t *f_type = ASR::down_cast<ASR::FunctionType_t>(f->m_function_signature);
-                                            if (f_type->m_return_var_type && !ASRUtils::is_allocatable(f_type->m_return_var_type)) {
-                                                f_type->m_return_var_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, f_type->m_return_var_type));
                                             }
                                         }
                                     } else {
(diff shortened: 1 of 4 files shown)


srinjoy933:
hey, @Jatinagarwal24 can you please review this pr , when you get some time.

Jatinagarwal24:
Hi @srinjoy933 here is my suggestion to you it is a good practice to write good messages in commit also there is intially only two commits when we open a pr which is first commit of fix and second of test.
Also it is easy to track changes if we do correct commit.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5738,30 +5750,25 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             }
                         }
                     }
-		    // enable sole `dimension` attribute
+            // enable sole `dimension` attribute
```
Jatinagarwal24: correct this tab.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2175,9 +2175,9 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
                             dims.reserve(al, 0);
                             std::string sym = "";
                             ASR::ttype_t *contained_type = determine_type(x.base.base.loc, sym, 
-                                                                return_attr_type->m_attr, false, 
-                                                                false, dims, nullptr /*TODO : pass var_sym of return*/,
-                                                                type_declaration, current_procedure_abi_type);
+                                                                        return_attr_type->m_attr, false, 
+                                                                        false, dims, nullptr /*TODO : pass var_sym of return*/,
+                                                                        type_declaration, current_procedure_abi_type);
```
Jatinagarwal24: correct this tab.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2213,8 +2213,37 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+            if (is_alloc_standalone && !ASRUtils::is_allocatable(type)) {
+                type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, type));
+            }
+            if (is_ptr_standalone && !ASRUtils::is_pointer(type)) {
+                type = ASRUtils::TYPE(ASR::make_Pointer_t(al, x.base.base.loc, type));
+            }
+
             SetChar variable_dependencies_vec;
             variable_dependencies_vec.reserve(al, 1);
+
```
Jatinagarwal24: remove this blank line.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2225,25 +2254,38 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
-                    diag::Level::Error, diag::Stage::Semantic, {
-                        diag::Label("", {x.base.base.loc})}));
-                throw SemanticAbort();
+                if (!is_alloc_standalone && !is_ptr_standalone) {
+                    diag.add(diag::Diagnostic(
+                        "Cannot specify the return type twice",
+                        diag::Level::Error, diag::Stage::Semantic, {
+                            diag::Label("", {x.base.base.loc})}));
+                    throw SemanticAbort();
+                }
```
Jatinagarwal24, line 2267: see if there exist  any test for this in `continue_compilation_1.f90` otherwise add the test for this.

srinjoy933:
Thank you @Jatinagarwal24 ,for your suggestion , I will surely follow this from the next pr.

Jatinagarwal24:
Now when you will fix my comments above remember to pull my commit first otherwise if you didn't pull them you will lost them.

Jatinagarwal24:
Hi @assem2002 Pls once review this. If something else also needs to cleanup.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5796,30 +5757,25 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             }
                         }
                     }
-		    // enable sole `dimension` attribute
+           // enable sole `dimension` attribute
```
Jatinagarwal24: This is not corrected @srinjoy933 Pls fix this whitespace also

Jatinagarwal24 (changes requested):

src/lfortran/semantics/ast_common_visitor.h
Jatinagarwal24: This is still not fixed yet.

srinjoy933: Now it is fixed @Jatinagarwal24 .

Jatinagarwal24 (approved):
Now looks good to me. 
@assem2002 Pls once review this.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7004,6 +7018,9 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 if (assgnd_pointer.count(sym)) {
                     is_pointer = true;
                 }
+                if (assgnd_allocatable.count(sym)) {
```
assem2002: I don't think it's needed here.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -5811,8 +5812,20 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                    if (sym_) {
+                                        ASR::symbol_t* sym_past_external = ASRUtils::symbol_get_past_external(sym_);
+                                        if (ASR::is_a<ASR::Variable_t>(*sym_past_external)) {
+                                            ASR::Variable_t *v = ASR::down_cast<ASR::Variable_t>(sym_past_external);
+                                            if (!ASRUtils::is_allocatable(v->m_type)) {
+                                                v->m_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, v->m_type));
+                                            }
+                                        }
+                                    } else {
+                                        assgnd_allocatable.insert(sym);
```
assem2002:
```suggestion
                                        assgnd_allocatable.insert(to_lower(sym));
```

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2213,6 +2213,40 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                            type = ASRUtils::TYPE(ASR::make_Pointer_t(al, x.base.base.loc, type));
+                        }
+                    }
+                }
+            }
+
+            // Catch standalone declarations (e.g., `allocatable :: value`) parsed earlier in this pass
+            bool is_alloc_standalone = false;
+            bool is_ptr_standalone = false;
+            for (const auto& a_name : assgnd_allocatable) {
```
assem2002: Use set's `count()`, it's faster than looping.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2213,6 +2213,40 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                if (to_lower(a_name) == return_var_name) is_alloc_standalone = true;
+            }
+            for (const auto& p_name : assgnd_pointer) {
```
assem2002: same

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2225,25 +2259,51 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
         } else {
+            bool is_alloc_standalone = false;
+            bool is_ptr_standalone = false;
+            for (const auto& a_name : assgnd_allocatable) {
+                if (to_lower(a_name) == return_var_name) is_alloc_standalone = true;
+            }
+            for (const auto& p_name : assgnd_pointer) {
+                if (to_lower(p_name) == return_var_name) is_ptr_standalone = true;
+            }
+
```
assem2002: Probably this not needed here, May you check?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2225,25 +2259,51 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
             if (return_type && !(x.n_attributes == 0 && compiler_options.implicit_typing && compiler_options.implicit_interface)) {
-                diag.add(diag::Diagnostic(
-                    "Cannot specify the return type twice",
-                    diag::Level::Error, diag::Stage::Semantic, {
-                        diag::Label("", {x.base.base.loc})}));
-                throw SemanticAbort();
+                if (!is_alloc_standalone && !is_ptr_standalone) {
```
assem2002, line 2261: I'm not sure if this is needed? Can you point out to some example?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -2225,25 +2259,51 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+                ASR::symbol_t* dummy_type_decl = nullptr;
+                return_variable->m_type = determine_type(x.base.base.loc, return_var_name, (AST::decl_attribute_t*)return_type, false, false, dummy_dims, nullptr, dummy_type_decl, current_procedure_abi_type);
+            }
+            
+            // Upgrade the implicitly typed variable to Allocatable or Pointer
+            if (is_alloc_standalone && !ASRUtils::is_allocatable(return_variable->m_type)) {
+                return_variable->m_type = ASRUtils::TYPE(ASR::make_Allocatable_t(al, x.base.base.loc, return_variable->m_type));
+            }
+            if (is_ptr_standalone && !ASRUtils::is_pointer(return_variable->m_type)) {
+                return_variable->m_type = ASRUtils::TYPE(ASR::make_Pointer_t(al, x.base.base.loc, return_variable->m_type));
```
assem2002, line 2285: Is that needed? This code exists in a branch where the return variable is defined as a variable, so I think the return type would've already had the proper type+pointer OR allocatable attribute set

assem2002 (commented):
The changes are in the good direction, Thanks You! 
Please address my comments.

src/lfortran/semantics/ast_symboltable_visitor.cpp
srinjoy933: I've updated both of these to use .count(return_var_name) to optimize the set lookups.

src/lfortran/semantics/ast_symboltable_visitor.cpp
srinjoy933: Yes, this is needed for the functions_62.f90 test case (integer function value() \n allocatable :: value). Because allocatable :: value is parsed as a declaration inside the body, the compiler creates a local symbol for it. When visit_Function processes the integer header, it enters this else block (since the symbol already exists) and incorrectly assumes the user explicitly declared the return type twice, throwing the F23 duplicate-type error. This bypass prevents that false positive when the symbol was generated purely from a standalone attribute.

src/lfortran/semantics/ast_symboltable_visitor.cpp
srinjoy933: I actually tested removing the type assignment here, but it resulted in an FPToSI LLVM backend crash!
It turns out it is absolutely needed because of Fortran's implicit typing rules. When the standalone attribute allocatable :: value is parsed, it creates the local variable, but because it starts with 'v', the compiler implicitly types it as Real.
If we don't explicitly call determine_type here to overwrite it, the return variable stays an allocatable Real instead of taking the Integer type specified in the function header. So I have kept the determine_type call and type upgrades here to ensure the implicitly typed variable takes the correct signature.

srinjoy933:
Hey @assem2002 , made the changes  that you have suggested ,can you please have a look on this now. Thank You!

src/lfortran/semantics/ast_symboltable_visitor.cpp
assem2002: Declaring variable should have cleaner pipeline, We'll need to refactor later

assem2002 (approved):

PR #11824, opened 2026-06-11
Fix: BitCast ASR verification crash

HarshilShah1804:
Towards #10886

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -15749,7 +15749,45 @@ public:
                 if (ASRUtils::is_character(*element_type)) {
                     // For character arrays: num_elements = source_bytes / element_bytes
                     // where element_bytes = kind * length for character(kind, len)
-                    llvm::Value* source_length = get_string_length(x.m_source);
+                    llvm::Value* source_length = nullptr;
+                    if (ASRUtils::is_character(*ASRUtils::expr_type(x.m_source))) {
+                        source_length = get_string_length(x.m_source);
+                        if (!source_length->getType()->isIntegerTy(64)) {
+                            source_length = builder->CreateZExt(source_length, llvm::Type::getInt64Ty(context));
+                        }
+                    } else {
+                        llvm::DataLayout dl(module->getDataLayout());
+                        if (ASRUtils::is_array(ASRUtils::expr_type(x.m_source))) {
+                            ASR::ttype_t* src_arr_type = ASRUtils::type_get_past_allocatable_pointer(ASRUtils::expr_type(x.m_source));
+                            ASR::Array_t* arr_t = ASR::down_cast<ASR::Array_t>(src_arr_type);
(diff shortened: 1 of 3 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -15784,7 +15784,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            } else {
+                                uint64_t n_elems = 1;
+                                for (size_t i = 0; i < arr_t->n_dims; i++) {
+                                    ASR::dimension_t dim = arr_t->m_dims[i];
+                                    if (dim.m_length && ASR::is_a<ASR::IntegerConstant_t>(*dim.m_length)) {
+                                        n_elems *= ASR::down_cast<ASR::IntegerConstant_t>(dim.m_length)->m_n;
+                                    }
+                                }
+                                num_elems = llvm::ConstantInt::get(llvm::Type::getInt64Ty(context), n_elems);
+                            }
```
Jatinagarwal24: I think this incorrectly assume that array dimensions are always compile-time constants (`IntegerConstant_t`). For an `explicit-shape` array with a `dynamic size (like integer :: arr(n))`, the dimension length is a variable (`Var_t`), not a constant. The loop silently skips it, resulting in `n_elems = 1`.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -15784,7 +15784,45 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            llvm::Type* src_el_llvm = llvm_utils->get_type_from_ttype_t_util(x.m_source, src_el_type, module.get());
+                            uint64_t el_sz = dl.getTypeAllocSize(src_el_llvm);
+                            llvm::Value* elem_size = llvm::ConstantInt::get(num_elems->getType(), el_sz);
+                            source_length = builder->CreateMul(num_elems, elem_size);
+                        } else {
+                            llvm::Type* src_type = llvm_utils->get_type_from_ttype_t_util(x.m_source, ASRUtils::expr_type(x.m_source), module.get());
+                            uint64_t sz = dl.getTypeAllocSize(src_type);
+                            source_length = llvm::ConstantInt::get(llvm::Type::getInt64Ty(context), sz);
+                        }
+                    }
```
Jatinagarwal24, line 15775: Please see that we can avoid using this manual iteration by using existing utility for eg in `llvm_utils.cpp` we  already has `LLVMUtils::get_array_size(...)` to minimise the code.

HarshilShah1804:
Thank you for the suggestions @Jatinagarwal24. I have updated the changes. Please check.

Jatinagarwal24 (approved):
Looks good to me now.
Thanks @HarshilShah1804 for the fix.
Pls feel free to merge once the CI is green.

Issue #11597, opened 2026-05-21
ICE when subroutines are moved


opixdown:
In `continue_compilation_1.f90` when subroutines int he contains part are moved to the module part for ex. 

at the end there is contains
```
    contains
    subroutine test_uminus_struct()
        use continue_compilation_1_mod, only: MyClass
        implicit none
        type(MyClass) :: tt
        print *, -tt
    end subroutine

    subroutine sub(f)
        interface
            function f(x)
                integer :: x, f
            end function
        end interface
    end subroutine
    subroutine sub_do_undeclared()
        implicit none
```
all the subroutines are moved above so one can direct add statements or call to function at the eof , currently it is causing ice 
i.e

```
  function op_clash_f(x) result(y)
        integer, intent(in) :: x
        integer :: y
        y = x
    end function op_clash_f

    subroutine test_uminus_struct()
        use continue_compilation_1_mod, only: MyClass
        implicit none
        type(MyClass) :: tt
        print *, -tt
    end subroutine

    subroutine sub(f)
        interface
            function f(x)
                integer :: x, f
            end function
        end interface
    end subroutine
    subroutine sub_do_undeclared()
        implicit none

```

originally posted by @certik 
https://github.com/lfortran/lfortran/pull/11591#discussion_r3278736130

PR #11826, opened 2026-06-12
fix :ICE when passing assumed-size arrays to incompatible dummy arguments

### LINKED ISSUE #11800: ICE for passing assumed size array to assumed rank pointer dummy
## MRE

```fortran
program mre
  integer :: x(1)
  call f(x)
contains

  subroutine f(a)
    integer, intent(in) :: a(*)
    call g(a)
  end subroutine f
  subroutine g(a)
    integer, pointer, intent(in) :: a(..)
    print *, "ok"
  end subroutine g

end program mre
```

## output

```text

Internal Compiler Error:
```

## gfortran behavior

```text

Error: Actual argument for 'a' cannot be an assumed-size array at (1)
```

Jatinagarwal24:
fixes #11800
**The Bug:**
Previously, passing an assumed-size array (e.g., `a(*)`) as an actual argument to a procedure expecting an assumed-shape array (e.g., `a(:)`), a deferred-shape pointer array, or an assumed-rank array would cause LFortran to crash during the LLVM code generation phase. The compiler would attempt to generate an `ArrayPhysicalCast` but fail with `AssertFailed: m_dim.m_start != nullptr` because assumed-size arrays lack the required dimension bounds metadata.

**The Fix:**
Added semantic validation inside `check_simple_intent_mismatch` in `src/libasr/asr_utils.h` to intercept this invalid array passing before it reaches codegen. 

The new logic explicitly checks if the actual argument's physical type is `UnboundedPointerArray` (assumed-size). If it is, it verifies the dummy argument's physical type. If the dummy argument is a `DescriptorArray` (assumed-shape), `PointerArray`, or an `AssumedRankArray` (with pointer/allocatable attributes), the compiler now gracefully aborts with a semantic error: 
`"Actual argument for '<a>' cannot be an assumed-size array"`

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -94,13 +94,13 @@ semantic error: Procedure declarations without an explicit interface (procedure(
     |         ^^^^^^^^^^^ 
 
 semantic error: Procedure 'op_clash_f' is already defined as an interface body
-   --> tests/errors/continue_compilation_1.f90:186:5 - 190:27
+   --> tests/errors/continue_compilation_1.f90:193:5 - 197:27
     |
-186 |        function op_clash_f(x) result(y)
+193 |        function op_clash_f(x) result(y)
     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
 ...
     |
-190 |        end function op_clash_f
+197 |        end function op_clash_f
     | ...^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
     |
  35 |            function op_clash_f(x) result(y)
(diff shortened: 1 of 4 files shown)


tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -94,13 +94,13 @@ semantic error: Procedure declarations without an explicit interface (procedure(
...
-   --> tests/errors/continue_compilation_1.f90:186:5 - 190:27
+   --> tests/errors/continue_compilation_1.f90:193:5 - 197:27
     |
-186 |        function op_clash_f(x) result(y)
+193 |        function op_clash_f(x) result(y)
     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
 ...
     |
-190 |        end function op_clash_f
+197 |        end function op_clash_f
```
Jatinagarwal24, line 103: I know this is not correct but I am not able to correct this ,if and to prevent this i tried adding test below this function then we are getting error in `stderr` file.

HarshitaKalani: This is fine. However, can you please create an issue for error you see?

Jatinagarwal24: I think one issue is already opened for that pls see this 
https://github.com/lfortran/lfortran/issues/11597

HarshitaKalani:
Please resolve conflicts.

Jatinagarwal24:
> Please resolve conflicts.

done.

PR #11837, opened 2026-06-12
fix: compile-time evaluation and ICEs for array-scalar operations in implied-do loops

### LINKED ISSUE #11833: ICE for implied_do array constructor with array expression
## MRE

```fortran
program implied_do
  implicit none

  integer :: i
  integer, parameter :: dims(2) = [2, 2]
  integer, parameter :: values(2) = [(dims*i, i=1,1)]

  print *, values
end program implied_do
```

## output

```text

Internal Compiler Error:
AssertFailed: !ASR::is_a<ASR::Array_t>(*e->m_type)
```

## gfortran behavior

```text

           2           2
```

Jatinagarwal24:
fixes #11833

This commit fixes a bug where array constructors involving array-scalar operations within implied-do loops (e.g., `[(dims * i, i = 1, 1)]`) would either cause an Internal Compiler Error (ICE) or produce garbage output at runtime.

**Root Causes:**
1. The semantic analyzer failed to extract the base type during initialization, resulting in the creation of illegal nested `Array(Array(...))` ASR nodes in `visit_ArrayInitializer` and `visit_ImpliedDoLoop`.
2. The `ImpliedDoLoopValuesVisitor` lacked implementations for evaluating array constants and broadcasts. Because of this, it was incorrectly falling back to `BaseWalkVisitor` on array nodes, corrupting the visitor state and failing to evaluate compile-time array arithmetic correctly.

**Fixes:**
- **Flattened Type Assignments:** Applied `ASRUtils::extract_type()` in `visit_ArrayInitializer` and `visit_ImpliedDoLoop` to ensure that 1D Array types are correctly constructed without illegal nesting.
- **Element-wise Array Evaluation Strategy:** Rather than making massive changes to every binary operation (`visit_IntegerBinOp`, `visit_RealBinOp`, etc.) to support array arithmetic, we updated the visitor to evaluate expressions element-by-element:
  - Added an `element_idx` parameter to `ImpliedDoLoopValuesVisitor`.
  - Implemented `visit_ArrayConstant` and `visit_ArrayBroadcast` to extract and evaluate only the specific scalar element requested by `element_idx`. This cleanly leverages the existing scalar binary operation handlers.
  - Refactored `populate_compiletime_array_for_idl` to calculate the array dimensions using `ASRUtils::get_fixed_size_of_array()` and then loop over the expression using `element_idx`.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3363,6 +3363,7 @@ RUN(NAME implied_do_loops37 LABELS gfortran llvm)
 RUN(NAME implied_do_loops38 LABELS gfortran llvm EXTRA_ARGS --use-loop-variable-after-loop)
 RUN(NAME implied_do_loops39 LABELS gfortran llvm)
 RUN(NAME implied_do_loops40 LABELS gfortran llvm)
+RUN(NAME implied_do_loops41 LABELS gfortran llvm)
 
 RUN(NAME minpack_01 LABELS gfortran llvm_rtlib EXTRAFILES
         minpack_01_module.f90 minpack_01_func.f90)
--- a/integration_tests/implied_do_loops41.f90
+++ b/integration_tests/implied_do_loops41.f90
@@ -1,0 +1,22 @@
+program implied_do_loops41
+    implicit none
+
+    integer :: i
+    integer, parameter :: dims(2) = [2, 2]
+    integer, parameter :: values_mul(2) = [(dims*i, i=1,1)]
+    integer, parameter :: values_add(2) = [(dims+i, i=2,2)]
+    integer, parameter :: values_sub(2) = [(dims-i, i=1,1)]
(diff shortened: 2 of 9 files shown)


HarshitaKalani (approved):

Jatinagarwal24:
Hi @HarshitaKalani Pls see if this can be merge.

HarshitaKalani (changes requested):
I think we should avoid adding array arithmetic logic to visit_IntegerBinOp. Instead, let's use an element_idx-based approach in visit_ArrayConstant/visit_ArrayBroadcast and evaluate each array element through the existing scalar evaluation path. This will automatically provide support for all existing types and operations.

Jatinagarwal24:
Hi @HarshitaKalani I have addressed your comment.
Pls once check it .
Thanks for pointing that out !

HarshitaKalani (approved):
lgtm. Thanks!
With this approach, the program also works when the values are real. Please feel free to add a test with real values with subsequent PRs.

PR #11843, opened 2026-06-13
fix: ICE for internal write to character array section

### LINKED ISSUE #11842: ICE for internal write to character array section
## MRE

```fortran
program character_section_io
  implicit none

  integer :: value = 7
  character(12) :: buffer(4, 1)

  write(buffer(:, 1), "(i12)") value
  print *, trim(adjustl(buffer(1, 1)))
end program character_section_io
```

## output

```text

LCOMPILERS_ASSERT failed:
Type mismatch in create_gep2: GEP target type does not match pointer's pointee type.

```

## gfortran behavior

```text

 7
```

opixdown:
fixes  #11842


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -20105,11 +20106,37 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            continue;
+                        }
+                        llvm::Value* left = get_i64_value(section->m_args[i].m_left);
+                        llvm::Value* right = get_i64_value(section->m_args[i].m_right);
+                        llvm::Value* step = section->m_args[i].m_step ?
+                            get_i64_value(section->m_args[i].m_step) :
+                            llvm::ConstantInt::get(context, llvm::APInt(64, 1));
+                        llvm::Value* extent = builder->CreateAdd(
+                            builder->CreateSDiv(builder->CreateSub(right, left), step),
+                            llvm::ConstantInt::get(context, llvm::APInt(64, 1)));
```
Jatinagarwal24, line 20133: For an empty slice like buffer(3:1), the math `(1 - 3)/1 + 1` results in `-1`.

Because a negative `array_size` is passed to the runtime function `_lfortran_string_write`, the runtime falls back into its scalar `else` block (since `-1 > 0` is false). This causes the compiler to completely ignore the slice boundaries and silently overwrite the base memory of the array, while reporting a successful write.
See this mre 
```
program write_empty
    implicit none

    integer :: value = 7
    character(12) :: buffer(4, 1)
    integer :: stat

    buffer = "unset"

    ! This is a zero-sized slice. We use iostat to catch the End-Of-File error.
    write(buffer(3:1, 1), "(i12)", iostat=stat) value

    print *, "iostat: ", stat
    print *, "buffer(1, 1): ", buffer(1, 1)
end program write_empty
```

Jatinagarwal24: Please fix this and also add the test case for this case

opixdown: thanks for pointing it out :)), pushing commit

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -11910,7 +11910,12 @@ LFORTRAN_API void _lfortran_string_write(lfortran_allocator_t* al, char **str_ho
     // For character array internal files, each '\n' in formatted text denotes
     // the next record (array element).
     if (*str_holder != NULL) {
-        if (is_array_unit && array_size > 0) {
+        if (is_array_unit && array_size <= 0) {
+            if (iostat != NULL) *iostat = -1;
+            internal_free(s);
```
Jatinagarwal24, line 12036: Just one very minor catch here currently, if the user doesn't provide an iostat= argument (meaning `iostat == NULL`), the code silently returns instead of aborting. Standard Fortran requires the program to crash when an unhandled EOF occurs.

Could you add the fprintf and exit(1) for the NULL case, like this?
```
if (is_array_unit && array_size <= 0) {
    if (iostat == NULL) {
        fprintf(stderr, "Fortran runtime error: End of file\n");
        exit(1);
    } else {
        *iostat = -1;
    }
    internal_free(s);
    va_end(args);
    return;
}
```
After this change feel free to merge.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -20128,6 +20128,9 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                         llvm::Value* extent = builder->CreateAdd(
                             builder->CreateSDiv(builder->CreateSub(right, left), step),
                             llvm::ConstantInt::get(context, llvm::APInt(64, 1)));
+                        llvm::Value* zero = llvm::ConstantInt::get(context, llvm::APInt(64, 0));
+                        extent = builder->CreateSelect(
+                            builder->CreateICmpSLT(extent, zero), zero, extent);
```
Jatinagarwal24, line 20136: This looks good.
Thanks !

Jatinagarwal24 (approved):
Looks good now.
Thanks @opixdown for the fix.
Just minor change required after that feel free to merge.

src/libasr/runtime/lfortran_intrinsics.c
opixdown: done

HarshitaKalani (approved):

PR #11844, opened 2026-06-13
Fix: Support derived-type component array sections in MAXLOC

### LINKED ISSUE #11639: MAXLOC rejects derived-type component array section with mask

srinjoy933:
Fixes #11639 

This pull request resolves a bug that caused the compiler to erroneously reject valid MAXLOC operations when evaluating array sections embedded inside arrays of derived types (such as top(:)%x(1)).

 (ast_common_visitor.h): Refactored the array resolution logic. The compiler now actively detects and preserves the dimensionality of parent struct arrays, successfully outputting an ArraySection_t node rather than losing shape context and defaulting to a scalar item.

 (asr_to_llvm.cpp & llvm_utils): Fixed how the compiler handles arrays that are nested inside custom data structures. We added a new helper function (load_pointer_element) so the compiler can safely find the exact memory addresses and figure out the correct shape and size of these nested arrays without crashing.


srinjoy933:
Hey @HarshitaKalani ,can you please have a look on this pr, Thank You!

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -5209,12 +5209,38 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                            llvm::Type* ptr_type = array_type->getPointerTo();
+                            tmp = llvm_utils->CreateLoad2(ptr_type, tmp);
+                        }
+                        
+                        llvm::Value* desc_ptr = llvm_utils->get_array_descriptor_ptr(tmp, array_type, false);
+                        llvm::Value* base_gep = llvm_utils->create_gep2(array_type, desc_ptr, 0);
+                        tmp = llvm_utils->load_pointer_element(base_gep, array_type);
+                    } else {
+                        // For fixed-size arrays, no descriptor exists. We use the raw array pointer.
+#if LLVM_VERSION_MAJOR < 15
```
HarshitaKalani: Could we avoid the direct LLVM_VERSION_MAJOR check here and use a version-independent helper instead?

srinjoy933:
Hey @HarshitaKalani , I have now removed the version specifications ,please have a look at this now.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9662,6 +9682,150 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        
+                    builder->CreateStore(adjusted_stride, dst_stride_ptr);
+                }
+
+                // Finalize Target Mapping
+                builder->CreateStore(allocated_target, dest_descriptor);
+                return;
+            }
+        }
+        // --- END NEW LOGIC ---
```
```suggestion
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9662,6 +9682,150 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         ASR::ArraySection_t* array_section = ASR::down_cast<ASR::ArraySection_t>(x.m_value);
         ASR::ttype_t* value_array_type = ASRUtils::expr_type(array_section->m_v);
 
+        // --- START NEW LOGIC FOR DERIVED TYPE COMPONENT ARRAYS ---
```
```suggestion
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -1973,7 +1981,7 @@ namespace LCompilers {
 #endif
         return value;
     }
-
+
```
certik: Don't add trailing whitespace

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -9662,6 +9682,150 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    adjusted_stride = builder->CreateUDiv(adjusted_stride,
+                        llvm::ConstantInt::get(arr_descr->get_index_type(), size_target));
+                        
+                    builder->CreateStore(adjusted_stride, dst_stride_ptr);
+                }
+
+                // Finalize Target Mapping
+                builder->CreateStore(allocated_target, dest_descriptor);
+                return;
+            }
```
certik, line 9819: This is a long blob of logic that looks AI generated. It would be good to understand exactly what it does and then simplify it to make it ideally 10 lines or less.

If not possible, then we need to understand exactly what is going on, and possibly split it into functions / reuse existing functions?

srinjoy933: yes @certik ,you are correct ,actually while writing this block of code , everytime I am facing some test failures ,so I took some help of AI while writing this particular block of code .As per your suggestion I reverted the block and wrote in a much simpler way and short in size  , please have a look and suggest if it looks good to you now, Thank You!

PR #11848, opened 2026-06-13
fix: reject assumed-size array for pointer/allocatable dummy

### LINKED ISSUE #11846: ICE for passing assumed size array to assumed rank pointer dummy
## MRE

```fortran
program main
  call forward([1])

contains

  subroutine forward(items)
    integer :: items(*)

    call consume(items)
  end subroutine forward

  subroutine consume(items)
    integer, pointer :: items(..)
  end subroutine consume

end program main
```

## output

```text

Internal Compiler Error:
```

## gfortran behavior

```text

Error: Actual argument for 'items' cannot be an assumed-size array at (1)
```

ambikeesshh:
fixes #11846

## Summary
Passing an assumed-size array (`integer :: x(*)`) to a pointer or allocatable dummy caused an ICE, `AssertFailed: m_dim.m_start != nullptr`, in `PointerToData_to_Descriptor`. An assumed-size array has no known extent for its last dimension, so it cannot be built into a descriptor. gfortran rejects this with a semantic error, so this adds the same check.

## Scope
`visit_SubroutineCall` in `src/lfortran/semantics/ast_body_visitor.cpp`. In the existing dummy-argument validation loop, when the dummy has the `pointer` or `allocatable` attribute and the actual is an assumed-size array (`UnboundedPointerArray`), emit `Actual argument for '<name>' cannot be an assumed-size array` instead of ICE'ing.

The trigger is the dummy being pointer/allocatable, not assumed-rank itself. Verified against gfortran across several variants (pointer vs allocatable dummy, assumed-rank vs fixed rank, `target`/`intent` on the actual). A non-pointer assumed-rank dummy is unaffected; that combination compiles in gfortran and is a separate pre-existing codegen limitation.

## Verification
- The issue MRE now reports the same error as gfortran.
- Added an error case to `tests/errors/continue_compilation_1.f90`; reference updated.
- Full integration suite on the llvm backend: 3706/3706 pass.

## Rationale
Matches gfortran's behavior and message. Kept in semantics per the rule that codegen must not infer or fix types.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -7551,6 +7551,24 @@ public:
                     ASR::Variable_t* dummy_var = ASR::down_cast<ASR::Variable_t>(dummy_sym);
                     ASR::ttype_t* dummy_type = dummy_var->m_type;
                     ASR::ttype_t* actual_type = ASRUtils::expr_type(args[i].m_value);
+                    // An assumed-size array has no known extent for its last
+                    // dimension, so it cannot be used where the dummy requires
+                    // a descriptor (pointer or allocatable).
+                    if (ASRUtils::is_pointer(dummy_type) ||
+                        ASRUtils::is_allocatable(dummy_type)) {
+                        ASR::ttype_t* actual_array =
+                            ASRUtils::type_get_past_allocatable_pointer(actual_type);
+                        if (ASR::is_a<ASR::Array_t>(*actual_array) &&
+                            ASR::down_cast<ASR::Array_t>(actual_array)->m_physical_type ==
+                                ASR::array_physical_typeType::UnboundedPointerArray) {
+                            std::string dummy_name = dummy_var->m_name;
+                            diag.semantic_error_label(
+                                "Actual argument for '" + dummy_name +
+                                "' cannot be an assumed-size array",
+                                {args[i].m_value->base.loc}, "");
+                            throw SemanticAbort();
+                        }
+                    }
                     if (ASRUtils::is_allocatable(dummy_type)) {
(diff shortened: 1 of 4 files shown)


ambikeesshh:
@opixdown mind reviewing it when you get a chance ?

ambikeesshh:
cc : @HarshitaKalani

Jatinagarwal24 (approved):
Looks good to me.
Thanks !

ambikeesshh:
ready to merge?

opixdown (approved):

opixdown:
LGTM thanks!!

Jatinagarwal24:
Hi @ambikeesshh According to the` Fortran Standard (2018, 15.5.2.4, paragraph 14 and 16)`, an assumed-size array also cannot be passed to an assumed-shape array `(items(:))` or an assumed-rank array `(items(..))`, even if they don't have pointer/allocatable attributes.

If you pass an assumed-size array to a plain assumed-shape dummy array, LFortran will still crash with the exact same ICE (`AssertFailed: m_dim.m_start != nullptr`) because assumed-shape and assumed-rank arrays also require an array descriptor, which an assumed-size array cannot provide.

So I am opening new issue for that pls fix that also.

ambikeesshh:
> Hi @ambikeesshh According to the` Fortran Standard (2018, 15.5.2.4, paragraph 14 and 16)`, an assumed-size array also cannot be passed to an assumed-shape array `(items(:))` or an assumed-rank array `(items(..))`, even if they don't have pointer/allocatable attributes.
> 
> If you pass an assumed-size array to a plain assumed-shape dummy array, LFortran will still crash with the exact same ICE (`AssertFailed: m_dim.m_start != nullptr`) because assumed-shape and assumed-rank arrays also require an array descriptor, which an assumed-size array cannot provide.
> 
> So I am opening new issue for that pls fix that also.

sure, working on it

PR #11849, opened 2026-06-13
Fix achar(kind=4) and ichar() on UCS-4 strings in LLVM backend

### LINKED ISSUE #11845: ICE for ichar on kind 4 character slice
## MRE

```fortran
program value
  implicit none

  integer, parameter :: marker = int(Z'2603')
  character(len=1, kind=4) :: glyph

  glyph(1:1) = achar(marker, kind=4)

  if (ichar(glyph(1:1)) /= marker) error stop
  print *, "slice value:", ichar(glyph(1:1))
end program value
```

## output

```text

LCOMPILERS_ASSERT failed:
ASRUtils::check_equal_type(str, str)

```

## gfortran behavior

```text

 slice value:        9731
```

Jatinagarwal24:
fixes #11845
Key updates:
- Enforced logical length of 1 for achar() and char() intrinsics.
- Updated declare_string_constant to safely read physical UTF-8 byte lengths.
- Unified runtime API by adding char_kind to _lfortran_strcpy_alloc and 
  _lfortran_copy_str_and_pad, removing duplicated _kind variants.
- Updated all runtime I/O call sites to properly pass the character kind.


src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2252,21 +2252,24 @@ namespace LCompilers {
             const int64_t compileTime_len = llvm::dyn_cast<llvm::ConstantInt>(len)->getValue().getSExtValue();
             if(compileTime_len < 0) {throw LCompilersException("String length cannot be negative.");}
             len = llvm::ConstantInt::get(context,
```
assem2002:
```suggestion
            mem_to_allocate = llvm::ConstantInt::get(context,
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2275,7 +2278,11 @@ namespace LCompilers {
         } else {
             throw LCompilersException("Unhandled string physical type");
         }
-        llvm::Value *s_alloc = builder->CreateAlloca(llvm::Type::getInt8Ty(context), builder->CreateSExtOrTrunc(len, llvm::Type::getInt32Ty(context)));
+        llvm::Value *alloc_len = builder->CreateSExtOrTrunc(len, llvm::Type::getInt32Ty(context));
```
assem2002:
```suggestion
        llvm::Value *mem_to_allocate = builder->CreateSExtOrTrunc(len, llvm::Type::getInt32Ty(context));
```

src/libasr/codegen/llvm_utils.h
```diff
@@ -499,13 +499,13 @@ class ASRToLLVMVisitor;
              * Allocate heap memory for string.
              * Notice : It doesn't set the length.
             */
-            void set_string_memory_on_heap(ASR::string_physical_typeType str_physical_type, llvm::Value* str, llvm::Value* len);
+            void set_string_memory_on_heap(ASR::string_physical_typeType str_physical_type, llvm::Value* str, llvm::Value* len, int char_kind = 1);
```
assem2002:
```suggestion
            void set_string_memory_on_heap(ASR::string_physical_typeType str_physical_type, llvm::Value* str, llvm::Value* len, int64_t char_kind = 1);
```

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2817,6 +2847,48 @@ namespace LCompilers {
...
+        std::vector<uint32_t> unicode;
+        size_t i = 0;
+        while (i < str.length()) {
+            uint32_t codepoint = 0;
+            unsigned char c = str[i];
+            if (c <= 0x7F) {
+                codepoint = c;
+                i += 1;
+            } else if ((c & 0xE0) == 0xC0) {
+                if (i + 1 < str.length()) {
```
assem2002: remove all these nested if/else checks -- and just assert that `lcompilers(i+X < str.length())`

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2852,16 +2924,34 @@ namespace LCompilers {
...
+                    } else if (kind == 4) {
+                        std::vector<uint32_t> unicode = utf8_to_utf32_local(initial_data_padded);
+                        unicode.resize(len, 0x20); // Pad with spaces (0x20)
+                        std::vector<uint8_t> bytes(len * 4);
+                        for (int64_t idx = 0; idx < len; ++idx) {
+                            uint32_t cp = unicode[idx];
+                            bytes[idx * 4 + 0] = cp & 0xFF;
+                            bytes[idx * 4 + 1] = (cp >> 8) & 0xFF;
+                            bytes[idx * 4 + 2] = (cp >> 16) & 0xFF;
+                            bytes[idx * 4 + 3] = (cp >> 24) & 0xFF;
```
assem2002: move this logic into `utf8_to_utf32_local` and maybe change function's name from `utf8_to_utf32_local` ->`utf8_to_unicode_bytes` and please give it a description.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2852,16 +2924,34 @@ namespace LCompilers {
...
+                        }
+                        const_data_as_array = llvm::ConstantDataArray::get(context, bytes);
+                    } else {
```
assem2002, line 2955: let's handle `kind=2`. I think with `utf8_to_unicode_bytes` adding a case for  `(kind=2)` would be easy to do.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6205,8 +6205,35 @@ namespace Achar {
...
+        } else if (i <= 0xFFFF) {
+            svalue += (char)(0xE0 | (i >> 12));
+            svalue += (char)(0x80 | ((i >> 6) & 0x3F));
+            svalue += (char)(0x80 | (i & 0x3F));
+        } else if (i <= 0x10FFFF) {
+            svalue += (char)(0xF0 | (i >> 18));
+            svalue += (char)(0x80 | ((i >> 12) & 0x3F));
+            svalue += (char)(0x80 | ((i >> 6) & 0x3F));
+            svalue += (char)(0x80 | (i & 0x3F));
+        }
```
assem2002, line 6244: I think we shouldn't encode into utf8 here. we just take the integer and return a unicode based on.

+assert that reutrn length =1 always.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4926,6 +4926,74 @@ LFORTRAN_API void _lfortran_strcpy_alloc(
...
+            *lhs = (char*)ALLOCATOR_REALLOC(al, *lhs, MAX(rhs_len, 1) * char_kind);
+        }
+        *lhs_len = rhs_len;
+        memcpy(*lhs, rhs, rhs_len * char_kind);
+    } else if(is_lhs_deferred && !is_lhs_allocatable) {
+        lfortran_assert(*lhs != NULL, "Runtime Error : Non-allocatable string isn't allocated.")
+        _lfortran_copy_str_and_pad_kind(*lhs, *lhs_len, rhs, rhs_len, char_kind);
+    }
+}
+
```
assem2002: We already have those functions. Why do we duplicate with different name but same implementation here?

let's use the original ones, if it's possible

also make sure to pass the kind of lhs

assem2002: related https://github.com/lfortran/lfortran/issues/11893

src/libasr/pass/intrinsic_functions.h
Jatinagarwal24: I added the `LCOMPILERS_ASSERT(len == 1)`  as requested.

Regarding the `UTF-8` encoding: we intentionally keep the `UTF-8` encoding step here because ASR `StringConstant_t` relies on `char* m_s` (a standard C-string). If we were to store raw UCS-4 bytes directly (e.g., `0x60 0x4F 0x00 0x00` for `年`), the embedded null bytes would prematurely terminate the C-string. This would break the ASR dumper and other downstream consumers that expect `m_s` to be a valid null-terminated string.

To handle this safely, we encode to UTF-8 here in ASR, and the LLVM backend (`llvm_utils.cpp`) then decodes it back to raw bytes using the new `utf8_to_unicode_bytes` function when generating the global arrays.

Jatinagarwal24:
Hi @assem2002 I have addressed all your comments.
Thanks for reviewing it.
Pls once check it again if something still needs to be corrected pls tell me.

assem2002 (approved):
LGTM. Thanks for working on this.

PR #11853, opened 2026-06-14
fix: ICE in pointer bounds remapping with FixedSizeArray targets

### LINKED ISSUE #11847: ICE for pointer remapping to array section with explicit lower bound
## MRE

```fortran
program pointer_mre
  implicit none

  integer, target :: values(4)
  integer, pointer :: p(:)

  values = [10, 20, 30, 40]

  p(5:) => values(2:4)

  print *, size(p)
end program pointer_mre
```

## output

```text

Internal Compiler Error:
AssertFailed: m_v
```

## gfortran behavior

```text

           3
```

Jatinagarwal24:
fixes #11847
* The Bug:
When performing pointer bounds remapping on an array section (e.g.,
`p(5:) => values(2:4)`), the compiler crashed with an Internal Compiler Error: 
`AssertFailed: arr_ty != nullptr` inside `handle_pointer_section_target`. 
This occurred because the codegen evaluated the section `values(2:4)` as a 
`DescriptorArray` and incorrectly assumed the underlying array `values` was 
also a `DescriptorArray`. When it tried to dynamically read the strides and
bounds of `values` using `get_pointer_to_dimension_descriptor_array`, the cast
failed because `values` was actually a `FixedSizeArray` with no dynamic descriptors.

* The Fix:
1. Updated `handle_pointer_section_target` to check the physical type of the 
   *underlying* array (`value_desc_asr_type`) instead of the section itself.
2. Added a code path to explicitly handle the `FixedSizeArray` and `PointerArray` 
   cases during pointer section targeting.
3. For fixed-size arrays, the section offset is now manually computed using 
   the statically available dimension sizes (`m_dims`) and slice indices 
   (`vidx.m_step`), avoiding any dependency on runtime descriptors.
4. Used `ASRUtils::make_ArraySize_t_util` instead of `ASR::make_ArraySize_t`
   to safely wrap sizes.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1022,6 +1022,7 @@ RUN(NAME pointer_14 LABELS gfortran llvm)
 RUN(NAME pointer_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME pointer_16 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME pointer_17 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME pointer_remapping_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
 RUN(NAME array_section_is_non_allocatable LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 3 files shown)


HarshitaKalani (approved):

assem2002 (approved):
LGTM

PR #11859, opened 2026-06-14
fix: diagnose assumed-size actuals to assumed-shape dummies

### LINKED ISSUE #11851: ICE: Passing assumed-size array to assumed-shape or assumed-rank dummy array
MRE for Assumed-Shape Array
```
program main
  call forward([1])

contains

  subroutine forward(items)
    integer :: items(*)
    
    ! Passing assumed-size array to assumed-shape dummy
    call consume_assumed_shape(items)
  end subroutine forward

  subroutine consume_assumed_shape(items)
    integer :: items(:)
  end subroutine consume_assumed_shape

end program main
```
MRE for Assumed-Rank Array
```
program main
  call forward([1])

contains

  subroutine forward(items)
    integer :: items(*)
    
    ! Passing assumed-size array to assumed-rank dummy
    call consume_assumed_rank(items)
  end subroutine forward

  subroutine consume_assumed_rank(items)
    integer :: items(..)
  end subroutine consume_assumed_rank

end program main
```
```
Expected Behavior (gfortran output for both):

Error: Actual argument for 'items' cannot be an assumed-size array at (1)
```
```
Current LFortran Behavior:

Internal Compiler Error:

```

ambikeesshh:
## Summary
- Diagnose assumed-size actual arrays passed to assumed-shape dummy arguments before descriptor lowering.
- Keep plain assumed-rank forwarding valid, while preserving the pointer/allocatable assumed-rank rejection already on main.
- Add semantic regression coverage for subroutine and function calls.

## Verification
- `ninja -C build lfortran`
- `PATH="$PWD/build/src/bin:$PATH" NO_COLOR=1 ./run_tests.py -t errors/continue_compilation_1.f90 -s`
- Checked plain assumed-rank `items(*) -> items(..)` still passes semantic analysis.

Fixes #11851


ambikeesshh:
cc: @Jatinagarwal24

Jatinagarwal24:
> cc: @Jatinagarwal24

Pls update the test references.
Thats why tests are failing and then mark ready to review once the CI is green.

ambikeesshh:
@Jatinagarwal24

tests/errors/continue_compilation_1.f90
```diff
@@ -779,6 +779,44 @@ subroutine assumed_size_to_pointer_dummy(x)
...
+    subroutine assumed_size_to_assumed_rank_function_forward(items)
+        implicit none
+        integer :: items(*)
+        print *, consume_assumed_rank_function(items)
+    end subroutine
+    integer function consume_assumed_rank_function(items)
+        implicit none
+        integer :: items(..)
+        consume_assumed_rank_function = 0
+    end function
```
Jatinagarwal24: I think these all tests are unecessary please remove them as the error is triggered at line 785 and 813 which i can see from `continue_compilation_1..90 stderr` file.

tests/errors/continue_compilation_1.f90
```diff
@@ -779,6 +779,44 @@ subroutine assumed_size_to_pointer_dummy(x)
...
+    subroutine assumed_size_to_assumed_shape_function_forward(items)
+        implicit none
+        integer :: items(*)
+        print *, consume_assumed_shape_function(items)  ! {Error} actual argument for 'items' cannot be an assumed-size array
+    end subroutine
+    integer function consume_assumed_shape_function(items)
+        implicit none
+        integer :: items(:)
+        consume_assumed_shape_function = size(items)
+    end function
```
Jatinagarwal24: same as above pls remove this also.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -11812,6 +11818,51 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                    "' cannot be an assumed-size array",
+                    Level::Error, Stage::Semantic, {
+                        Label("", {args[actual_i].m_value->base.loc})
+                    }));
+                throw SemanticAbort();
+            }
+            actual_i++;
+        }
+    }
+
```
Jatinagarwal24: I think this is not needed because there is already a dedicated loop checking actual arguments against dummy arguments! It even explicitly checks if an assumed-size actual argument is being passed to an allocatable or pointer dummy argument. A minimal fix would have just added is_assumed_shape to that existing if condition.

Jatinagarwal24: So once you drop this function remember to remove all the changes related to where you are calling this function(`validate_assumed_size_actual_arguments`).

src/libasr/asr_utils.h
```diff
@@ -5179,6 +5179,11 @@ inline bool check_class_assignment_compatibility(ASR::expr_t* target, ASR::expr_
         sym_target = ASRUtils::symbol_get_past_external(sym_target);
         ASR::symbol_t* sym_value = ASRUtils::get_struct_sym_from_struct_expr(value);
         sym_value = ASRUtils::symbol_get_past_external(sym_value);
+        if (sym_target == nullptr || sym_value == nullptr ||
+                !ASR::is_a<ASR::Struct_t>(*sym_target) ||
+                !ASR::is_a<ASR::Struct_t>(*sym_value)) {
+            return false;
+        }
```
Jatinagarwal24: Are these changes required pls check if we can remove them.

src/libasr/asr_utils.cpp
```diff
@@ -2168,7 +2173,12 @@ bool use_overloaded_assignment(ASR::expr_t* target, ASR::expr_t* value,
             }
             expr_dt = target;
         } else if( ASR::is_a<ASR::StructType_t>(*value_type) ) {
-            ASR::Struct_t* value_struct = ASR::down_cast<ASR::Struct_t>(ASRUtils::symbol_get_past_external(ASRUtils::get_struct_sym_from_struct_expr(value)));
+            ASR::symbol_t* value_sym = ASRUtils::symbol_get_past_external(
+                ASRUtils::get_struct_sym_from_struct_expr(value));
+            if (value_sym == nullptr || !ASR::is_a<ASR::Struct_t>(*value_sym)) {
+                return false;
+            }
+            ASR::Struct_t* value_struct = ASR::down_cast<ASR::Struct_t>(value_sym);
```
Jatinagarwal24: Here also pls check if we want these changes or not.
I think they are not needed,

Jatinagarwal24:
Hi @ambikeesshh Basically fix is simple first  eliminate  your new `validate_assumed_size_actual_arguments` function entirely to avoid unnecessary loop duplication. Instead, in `ast_body_visitor.cpp`, you can simply add the `is_assumed_shape condition` to the already existing argument validation loop that currently checks for `pointer/allocatable` dummies. Similarly, in `ast_common_visitor.h`, you can follow the same logic directly inside the existing `validate_missing_required_arguments() function`, which is already called at every function instantiation point.

Thanks !

tests/errors/continue_compilation_1.f90
ambikeesshh: I kept these intentionally as positive coverage. The issue has two different cases: `items(*) -> items(:)` should be rejected, but `items(*) -> items(..)` should not be rejected because assumed-rank can accept assumed-size actuals. These tests make sure we do not accidentally reject the valid assumed-rank path while fixing the assumed-shape diagnostic

src/lfortran/semantics/ast_common_visitor.h
ambikeesshh: Since the helper is still needed for the function/generic call paths, I think the call sites should stay too. They cover the places where calls are resolved to the final callee before ASR construction.

src/libasr/asr_utils.h
ambikeesshh: I checked this locally after merging current `main`. Without this guard, `continue_compilation_1` aborts with `down_cast(nullptr)` in `check_class_assignment_compatibility()`. So this looks needed for the merged branch.

src/libasr/asr_utils.cpp
ambikeesshh: I checked this one locally too. Without this guard, `continue_compilation_1` aborts with `down_cast(nullptr)` in `use_overloaded_assignment()`. So I kept it to avoid the post-merge crash.

ambikeesshh:
hey @Jatinagarwal24 
I checked all the reviews. The current structure is intentional: the helper covers function/generic paths that the existing subroutine-call loop does not, the assumed-rank tests protect the valid `items(*) -> items(..)` case, and the `asr_utils` guards prevent real `down_cast(nullptr)` crashes after merging current `main`. 
let me know if im incorrect anywhere

thanks for the review!

tests/errors/continue_compilation_1.f90
ambikeesshh: For this test block, same reasoning as above: this is intentional positive coverage for the valid assumed-rank function-call path, so we do not accidentally reject `items(*) -> items(..)` while fixing the assumed-shape case.

ambikeesshh:
hey @HarshitaKalani can you please review it
thanks!

HarshilShah1804:
@ambikeesshh, I think @Jatinagarwal24 is correct to point out removing the correctly working cases. The test `errors/continue_compilation_1` is meant to store the invalid cases and to ensure we don't merge any change that would wrongly compile invalid cases. If you feel some valid cases are missing from the tests, you can add them in the `integration_tests/`, but please check if similar tests are already present to avoid duplicacy.

assem2002:
@ambikeesshh Do you need help with this PR? Any clarifications ?

ambikeesshh:
hey @assem2002, thanks for checking in! all good. I'm reworking it per the review feedback (folding the check into the existing validation loops and cleaning up the tests), plus resolving a fresh conflict with main. will push the update soon.

ambikeesshh:
@assem2002 @Jatinagarwal24 could you review and add the exhaustive label check?

ambikeesshh:
cc: @assem2002

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7652,6 +7652,20 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                             throw SemanticAbort();
```
assem2002: I believe we shouldn't trigger an error when the parameter is (allocatable OR poitner OR ASR::physicalArrayType::DescriptorArray), rather we should trigger the deferredsize case (in other words when all dimensions are nullptr).
let's not rely on ASR's `allocatable`, `pointer`, `physicalArrayType::DescriptorArray` as an indication that dimensions are nullptr because sometimes they don't enforce that.
***
## What to do?
 refactor the already present semantic error to be triggered when parameter (function's arg) is an array of empty dims -- You can find utility at `asr_utils.h` to do that. (notice the code you introduce here won't be needed as the refactor will cover all these)

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -12355,6 +12355,26 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 ASR::Variable_t* dummy_var = ASR::down_cast<ASR::Variable_t>(dummy_sym);
                 dummy_name = dummy_var->m_name;
                 is_required = dummy_var->m_presence != ASR::presenceType::Optional;
+                if (i < args.size() && args[i].m_value != nullptr) {
```
assem2002, line 12359: same as discussed above. Trigger error on empty dimensions.

tests/errors/continue_compilation_1.f90
assem2002: I can't see the case of `item(*)` -> `item(..).
I can see it's covering the case of both function AND subroutine, which is good.
***
Thing to notice in the future : Add those erroneous functions at the end of file, to introduce minimal change to the reference file, so we can review it easily.

integration_tests/CMakeLists.txt
```diff
@@ -970,6 +970,7 @@ RUN(NAME assumed_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME assumed_rank_12 LABELS gfortran) # move to llvm once assumed-size -> assumed-rank codegen is supported (#11851)
```
assem2002:
```suggestion
COMPILE(NAME assumed_rank_12 LABELS gfortran) # move to llvm once assumed-size -> assumed-rank codegen is supported (#11851)
```

If it works just fine, let's remove the reference test.

assem2002 (commented):
looks to be in the good direction. Thanks
Please address my comments.

src/lfortran/semantics/ast_common_visitor.h
ambikeesshh: done, same unified condition here. The separate validate_assumed_size_actual_arguments helper is gone the check now lives inline in validate_missing_required_arguments(), which is already called at every function instantiation point, so the function/generic paths are covered without a duplicate loop

integration_tests/CMakeLists.txt
ambikeesshh: Applied, with one small tweak: the COMPILE macro takes COMPILERS rather than LABELS, so it's registered as COMPILE(NAME assumed_rank_12 COMPILERS gfortran). It compiles fine with gfortran, so I also dropped the tests.toml entry and deleted the ASR reference files as suggested.

tests/errors/continue_compilation_1.f90
ambikeesshh: You're right, `items(*) -> items(..)` is valid Fortran so it doesn't belong in an errors file. I removed those cases and moved the coverage to `integration_tests/assumed_rank_12.f90` as @HarshilShah1804 suggested, to guard the valid assumed-rank path from #11851. It's registered with `COMPILE` for gfortran only since llvm codegen isn't supported yet. The error cases are now at the end of the file to keep the reference diff minimal. Noted for the future, thanks.

src/lfortran/semantics/ast_body_visitor.cpp
ambikeesshh: Done, refactored in the latest push. Both call paths now use one unified condition instead of relying on pointer/allocatable alone.

One thing I found while doing it: empty dims alone can't be the whole condition. Assumed-size dummies (x(*)) and implicit-interface synthesized dummies also carry empty dimensions, and both legally accept assumed-size actuals a pure is_dimension_empty check broke implicit_interface_20.f90 with a false positive. So the final condition is: assumed-size actual, and the dummy is pointer/allocatable OR (empty dims AND lowered to DescriptorArray). The physical type is what separates a real assumed-shape dummy from the assumed-size/implicit cases that share the empty-dims representation. Added a comment in the code explaining this

ambikeesshh:
@assem2002 pushed the rework. Both checks are now unified into one condition triggered on empty dims plus descriptor physical type (needed to avoid a false positive on implicit-interface dummies) The separate helper is removed, error cases moved to the end of continue_compilation_1.f90, assumed-rank coverage moved to integration_tests/assumed_rank_12.f90 via COMPILE with gfortran, and the reference test dropped. Also merged latest main

thanks for the review!

ambikeesshh:
ci's all green

integration_tests/CMakeLists.txt
```diff
@@ -970,6 +970,7 @@ RUN(NAME assumed_rank_08 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_10 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME assumed_rank_11 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+COMPILE(NAME assumed_rank_12 COMPILERS gfortran) # move to RUN with llvm once assumed-size -> assumed-rank codegen is supported (#11851)
```
assem2002:
```suggestion
COMPILE(NAME assumed_rank_12 COMPILERS gfortran llvm llvm_wasm) # move to RUN with llvm once assumed-size -> assumed-rank codegen is supported (#11851)
```

src/lfortran/semantics/ast_body_visitor.cpp
assem2002: Nice catch! I dropped that fact.

assem2002 (approved):
LGTM. Thanks

PR #11860, opened 2026-06-14
Fix: Skip Warning about symbols that shadow the same symbol in the current scope

### LINKED ISSUE #11286: Warning about symbols that shadow the same symbol in the current scope

AnshMehta1:
Fixes #11286

### DIFF
--- a/integration_tests/shadow_same_symbol.f90
+++ b/integration_tests/shadow_same_symbol.f90
@@ -1,0 +1,15 @@
+module a_shadow_same_symbol
+    use, intrinsic :: ISO_FORTRAN_ENV, only: int64
+    implicit none
+end module
+
+module b_shadow_same_symbol
+    use a_shadow_same_symbol
+    use, intrinsic :: ISO_FORTRAN_ENV, only: int64
+    implicit none
+end module
+
+program shadow_same_symbol
+    use b_shadow_same_symbol
+    implicit none
+end program
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -20469,7 +20469,13 @@ public:
                 );
             current_scope->add_or_overwrite_symbol(local_sym, ASR::down_cast<ASR::symbol_t>(fn));
         } else if (ASR::is_a<ASR::Variable_t>(*t)) {
-            if (current_scope->get_symbol(local_sym) != nullptr) {
+            ASR::symbol_t* existing = current_scope->get_symbol(local_sym);
+            if (existing != nullptr) {
+                existing = ASRUtils::symbol_get_past_external(existing);
+                ASR::symbol_t* incoming = ASRUtils::symbol_get_past_external(t);
+                if (existing == incoming) {
+                    return;
+                }
                 diag.add(Diagnostic(
                     "Symbol '" + local_sym + "' from module '" + m->m_name + "' shadows '" + local_sym + "' in the current scope",
                     Level::Warning, Stage::Semantic, {
--- a/tests/tests.toml
+++ b/tests/tests.toml
@@ -5221,3 +5221,7 @@ run = true
 [[test]]
 filename = "errors/read_eof_01.f90"
 run = true
+
+[[test]]
+filename = "../integration_tests/shadow_same_symbol.f90"
+run = true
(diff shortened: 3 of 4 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -20469,7 +20469,13 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            ASR::symbol_t* existing = current_scope->get_symbol(local_sym);
+            if (existing != nullptr) {
+                existing = ASRUtils::symbol_get_past_external(existing);
+                ASR::symbol_t* incoming = ASRUtils::symbol_get_past_external(t);
+                if (existing == incoming) {
+                    return;
+                }
                 diag.add(Diagnostic(
                     "Symbol '" + local_sym + "' from module '" + m->m_name + "' shadows '" + local_sym + "' in the current scope",
                     Level::Warning, Stage::Semantic, {
```
HarshitaKalani, line 20481: Do we have a test for this warning btw?

HarshitaKalani (approved):

src/lfortran/semantics/ast_common_visitor.h
AnshMehta1: I don't thinks so but am not sure about this.

integration_tests/shadow_same_symbol.f90
```diff
@@ -0,0 +1,15 @@
...
+module b_shadow_same_symbol
+    use a_shadow_same_symbol
+    use, intrinsic :: ISO_FORTRAN_ENV, only: int64
+    implicit none
+end module
+
+program shadow_same_symbol
+    use b_shadow_same_symbol
+    implicit none
+end program
```
certik, line 15: This is not registered in cmake.

tests/tests.toml
```diff
@@ -5221,3 +5221,7 @@ run = true
 [[test]]
 filename = "errors/read_eof_01.f90"
 run = true
+
+[[test]]
+filename = "../integration_tests/shadow_same_symbol.f90"
+run = true
```
certik, line 5227: What is the purpose of this reference test? Should this just be an integration test?

integration_tests/shadow_same_symbol.f90
AnshMehta1: Earlier, this code wasn't producing any errors; it was only generating unnecessary warnings. If it starts producing warnings again in the future, I don't think CMake will catch them.

tests/tests.toml
AnshMehta1: If in future, if we have this warning again, then this fails the reference test. I don't think integration test has any concerns with warnings as far as code is compiling without errors.

src/lfortran/semantics/ast_common_visitor.h
certik: We should add a test that this still warns if the symbols are not the same, but have the same name. We still want to warn. Just not if it is actually the same symbol.

certik: Follow up issue: https://github.com/lfortran/lfortran/issues/12448

tests/tests.toml
certik: This is a compile-time warning, so we should not be running the test via `run=true`, but only do ASR via `asr=true`.

Also this must be more comprehensive: https://github.com/lfortran/lfortran/issues/12448

integration_tests/shadow_same_symbol.f90
certik: Then it should still be tested as an integration test in addition to reference test.

PR #11861, opened 2026-06-14
fix: strip pointer attribute from struct arguments in shape()

HarshitaKalani:


assem2002 (commented):
Thanks for the fix.
The fix here seems fine. shape's source being a pointer or not isn't a big deal.
***
 but a proper fix would be from the backend side where we properly load a `ASR::StructInstanceMember`
***
See this similar bug ( it still fail with this PR) :
```.f90
program p
   implicit none

   type :: inner_t
      integer :: rank
   end type inner_t

   type :: outer_t
      type(inner_t), pointer :: arr(:,:)
      end type outer_t
      
   type(outer_t) :: x
   integer :: s(2)

   call shape_clone(x%arr)
   
   contains
    subroutine shape_clone(x)
    type(inner_t), pointer :: x(:,:)
  end subroutine

end program 
```
***
If things are hard to fix, feel free to merge this and open an issue with MRE above.

assem2002 (approved):
Thanks for the fix

PR #11862, opened 2026-06-14
Prevent complex pow crash

### LINKED ISSUE #3823: Bug in complex power
```fortran
program mandelbrot
implicit none
real(dp), parameter :: D = 0.035_dp
integer :: ix, iy, k
complex(dp) :: c, z
logical escaped
do iy=-15,15
    do ix=-35,35
        c = cmplx(-0.75_dp+D*ix, -1.66_dp*D*iy, dp)
        z = 0
        escaped = .false.
        do k=1,200
            z = z**2 + c
            if (z%re**2 + z%im**2 > 4) then
                escaped = .true.
                exit
            end if
        end do
        if (escaped) then
            WRITE (*,"(a)",advance="no") " "
        else
            WRITE (*,"(a)",advance="no") "*"
        end if
    end do
    WRITE (*,*)
end do
end program
```
This gives:
```console
$ lldb ./m2.out              
(lldb) target create "./m2.out"
Current executable set to '/Users/ondrej/repos/mandel2/m2.out' (arm64).
(lldb) run
Process 77243 launched: '/Users/ondrej/repos/mandel2/m2.out' (arm64)
                                                     *                 
                                                   ****                
                                                  ******               
                                                   *****               
                                              * *********              
                                         *** ****************          
                                           ******************** **     
                                        *************************      
                                       ****************************    
                                     *******************************   
                                     ******************************    
                        * *****     ********************************   
                       ***********  *Process 77243 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603ff0)
    frame #0: 0x000000019ed41d08 libsystem_m.dylib`clog + 8
libsystem_m.dylib`clog:
->  0x19ed41d08 <+8>:  stp    d11, d10, [sp, #0x20]
    0x19ed41d0c <+12>: stp    d9, d8, [sp, #0x30]
    0x19ed41d10 <+16>: stp    x29, x30, [sp, #0x40]
    0x19ed41d14 <+20>: add    x29, sp, #0x40
Target 0: (m2.out) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603ff0)
  * frame #0: 0x000000019ed41d08 libsystem_m.dylib`clog + 8
    frame #1: 0x000000019ed421a8 libsystem_m.dylib`cpow + 28
    frame #2: 0x00000001001e58d0 liblfortran_runtime.0.dylib`_lfortran_complex_pow_64(a=0x000000016f6040e0, b=0x000000016f6040d0, result=0x000000016f6040c0) at lfortran_intrinsics.c:855:29
    frame #3: 0x0000000100003e38 m2.out`main + 480
    frame #4: 0x00000001000110f4 dyld`start + 520
```

srinjoy933:
Fixes #3823

### DIFF
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,0 +1,15 @@
+program complex_pow_io
+    implicit none
+    complex :: z32
+    complex(8) :: z64
+
+    z32 = (0.0, 0.0)
+    z64 = (0.d0, 0.d0)
+    
+    print *, z32**2
+    print *, z64**2
+
+    write(*, "(a)", advance="no") "A"
+    write(*, "(a)", advance="no") " "
+    write(*, "(a)") "B"
+end program
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,7 +1,13 @@
 program complex_pow_io
     implicit none
-    complex :: base32, res32
-    complex(8) :: base64, res64
+    complex :: z32, base32, res32
+    complex(8) :: z64, base64, res64
+
+    z32 = (3.0, 4.0)
+    if (z32**2 /= (-7.0, 24.0)) error stop 1
+
+    z64 = (3.0d0, 4.0d0)
+    if (z64**2 /= (-7.0d0, 24.0d0)) error stop 2
 
     base32 = (0.0, 0.0)
     res32 = base32 ** 0.0
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,13 +1,23 @@
 program complex_pow_io
     implicit none
-    complex :: z32
-    complex(8) :: z64
+    complex :: z32, base32, res32
+    complex(8) :: z64, base64, res64
 
-    z32 = (0.0, 0.0)
-    z64 = (0.d0, 0.d0)
-    
-    print *, z32**2
-    print *, z64**2
+    z32 = (0.12345, 0.6789)
+    if (z32**2 /= z32*z32) error stop 1
+
+    z64 = (0.12345d0, 0.6789d0)
+    if (z64**2 /= z64*z64) error stop 2
+
+    base32 = (0.0, 0.0)
+    res32 = base32 ** 0.0
+    res32 = base32 ** (-2.0)
+
+    base64 = (0.0d0, 0.0d0)
+    res64 = base64 ** 0.0d0
+    res64 = base64 ** (-2.0d0)
+
+    if (real(res32) == 999.0) print *, "dummy"
 
     write(*, "(a)", advance="no") "A"
     write(*, "(a)", advance="no") " "
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,13 +1,7 @@
 program complex_pow_io
     implicit none
-    complex :: z32, base32, res32
-    complex(8) :: z64, base64, res64
-
-    z32 = (0.12345, 0.6789)
-    if (z32**2 /= z32*z32) error stop 1
-
-    z64 = (0.12345d0, 0.6789d0)
-    if (z64**2 /= z64*z64) error stop 2
+    complex :: base32, res32
+    complex(8) :: base64, res64
 
     base32 = (0.0, 0.0)
     res32 = base32 ** 0.0
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -3,11 +3,12 @@ program complex_pow_io
     complex :: z32, base32, res32
     complex(8) :: z64, base64, res64
 
+
     z32 = (3.0, 4.0)
-    if (z32**2 /= (-7.0, 24.0)) error stop 1
+    if (abs(z32**2 - (-7.0, 24.0)) > 1e-4) error stop 1
 
     z64 = (3.0d0, 4.0d0)
-    if (z64**2 /= (-7.0d0, 24.0d0)) error stop 2
+    if (abs(z64**2 - (-7.0d0, 24.0d0)) > 1e-10) error stop 2
 
     base32 = (0.0, 0.0)
     res32 = base32 ** 0.0
@@ -16,6 +17,7 @@ program complex_pow_io
     base64 = (0.0d0, 0.0d0)
     res64 = base64 ** 0.0d0
     res64 = base64 ** (-2.0d0)
+
 
     if (real(res32) == 999.0) print *, "dummy"
 
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,27 +1,33 @@
-program complex_pow_io
+program mandelbrot
     implicit none
-    complex :: z32, base32, res32
-    complex(8) :: z64, base64, res64
+    integer, parameter :: dp = kind(0.d0)
+    real(dp), parameter :: D = 0.035_dp
+    integer :: ix, iy, k
+    complex(dp) :: c, z
+    logical :: escaped
 
-
-    z32 = (3.0, 4.0)
-    if (abs(z32**2 - (-7.0, 24.0)) > 1e-4) error stop 1
-
-    z64 = (3.0d0, 4.0d0)
-    if (abs(z64**2 - (-7.0d0, 24.0d0)) > 1e-10) error stop 2
-
-    base32 = (0.0, 0.0)
-    res32 = base32 ** 0.0
-    res32 = base32 ** (-2.0)
-
-    base64 = (0.0d0, 0.0d0)
-    res64 = base64 ** 0.0d0
-    res64 = base64 ** (-2.0d0)
-
-
-    if (real(res32) == 999.0) print *, "dummy"
-
-    write(*, "(a)", advance="no") "A"
-    write(*, "(a)", advance="no") " "
-    write(*, "(a)") "B"
-end program
+    do iy = -15, 15
+        do ix = -35, 35
+            c = cmplx(-0.75_dp + D*ix, -1.66_dp * D*iy, dp)
+            z = (0.0_dp, 0.0_dp)
+            escaped = .false.
+            
+            do k = 1, 200
+                z = z**2 + c
+                
+                if (z%re**2 + z%im**2 > 4.0_dp) then
+                    escaped = .true.
+                    exit
+                end if
+            end do
+            
+            if (escaped) then
+                write(*, "(a)", advance="no") " "
+            else
+                write(*, "(a)", advance="no") "*"
+            end if
+        end do
+        
+        write(*, *) 
+    end do
+end program mandelbrot
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,33 +1,25 @@
-program mandelbrot
+program test_complex_pow_precision
     implicit none
     integer, parameter :: dp = kind(0.d0)
-    real(dp), parameter :: D = 0.035_dp
-    integer :: ix, iy, k
-    complex(dp) :: c, z
-    logical :: escaped
+    complex(dp) :: z, expected
+    real(dp) :: error_re, error_im
 
-    do iy = -15, 15
-        do ix = -35, 35
-            c = cmplx(-0.75_dp + D*ix, -1.66_dp * D*iy, dp)
-            z = (0.0_dp, 0.0_dp)
-            escaped = .false.
-            
-            do k = 1, 200
-                z = z**2 + c
-                
-                if (z%re**2 + z%im**2 > 4.0_dp) then
-                    escaped = .true.
-                    exit
-                end if
-            end do
-            
-            if (escaped) then
-                write(*, "(a)", advance="no") " "
-            else
-                write(*, "(a)", advance="no") "*"
-            end if
-        end do
-        
-        write(*, *) 
-    end do
-end program mandelbrot
+    z = cmplx(-1.5_dp, 0.0_dp, dp)
+    expected = cmplx(2.25_dp, 0.0_dp, dp)
+    
+    z = z**2
+    
+    error_re = abs(z%re - expected%re)
+    error_im = abs(z%im - expected%im)
+    
+    if (error_re > 1.0e-15_dp .or. error_im > 0.0_dp) then
+        print *, "Test Failed!"
+        print *, "Expected: ", expected
+        print *, "Got:      ", z
+        print *, "Error Re: ", error_re
+        print *, "Error Im: ", error_im
+        error stop 1
+    end if
+    
+    print *, "Test Passed!"
+end program test_complex_pow_precision
--- a/integration_tests/complex_pow_io.f90
+++ b/integration_tests/complex_pow_io.f90
@@ -1,34 +1,20 @@
 program complex_pow_io
     implicit none
     integer, parameter :: dp = kind(0.d0)
-    real(dp), parameter :: D = 0.035_dp
-    integer :: ix, iy, k
     complex(dp) :: c, z
-    logical :: escaped
 
-    do iy = -15, 15
-        do ix = -35, 35
-            c = cmplx(-0.75_dp + D*ix, -1.66_dp * D*iy, dp)
-            z = (0.0_dp, 0.0_dp)
-            escaped = .false.
-            
-            do k = 1, 200
-                
-                z = z**2 + c
-                
-                if (z%re**2 + z%im**2 > 4.0_dp) then
-                    escaped = .true.
-                    exit
-                end if
-            end do
-            
-            if (escaped) then
-                write(*, "(a)", advance="no") " "
-            else
-                write(*, "(a)", advance="no") "*"
-            end if
-        end do
-        
-        write(*, *) 
-    end do
+
+    c = cmplx(-0.75_dp, 0.1_dp, dp)
+    z = (0.0_dp, 0.0_dp)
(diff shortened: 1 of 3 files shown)


srinjoy933:
Hey @HarshitaKalani @Jatinagarwal24  ,can you please have a look on this pr, Thank You!

HarshitaKalani (approved):

integration_tests/complex_pow_io.f90
```diff
@@ -0,0 +1,15 @@
...
+    z32 = (0.0, 0.0)
+    z64 = (0.d0, 0.d0)
+    
+    print *, z32**2
+    print *, z64**2
+
+    write(*, "(a)", advance="no") "A"
+    write(*, "(a)", advance="no") " "
+    write(*, "(a)") "B"
+end program
```
Jatinagarwal24: The test which you added works perfectly in main.
There is issue on i think around 16 th line output of lfortran as i can see.
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran complex_pow_io.f90 
(0.00000000,0.00000000)
(0.0000000000000000,0.0000000000000000)
A B
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran complex_pow_io.f90 
             (0.00000000,0.00000000)
               (0.0000000000000000,0.0000000000000000)
A B

(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran mre.f90 
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
   *  *   *   * ************************************************
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************   
                        * *****     ********************************
                                     ******************************
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran mre.f90 
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
****************************************************************
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************
                        * *****     ********************************
                                     ******************************
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *


```

Jatinagarwal24: Hi @srinjoy933 i checked that your fix didnt fixed the output  produced by lfortran. Can you pls once check it.
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran mre.f90 
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **     
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************    
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
   *  *   *   * ************************************************       
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************
                        * *****     ********************************
                                     ******************************    
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran mre.f90 
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
****************************************************************
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************
                        * *****     ********************************   
                                     ******************************
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *

```

srinjoy933: hey @Jatinagarwal24 thank you for looking at this , now I have made the changes for the correct output. Can you please have a look at this .

Jatinagarwal24:
I can see this is fixed 
```
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
****************************************************************
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************   
                        * *****     ********************************
                                     ******************************
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran examples/expr2.f90 
                                                     *
                                                   ****
                                                  ******
                                                   *****
                                              * *********
                                         *** ****************
                                           ******************** **
                                        *************************
                                       ****************************
                                     *******************************
                                     ******************************
                        * *****     ********************************
                       ***********  ********************************
                      ************* *******************************
                  ** ************** ******************************
****************************************************************
                  ** ************** ******************************
                      ************* *******************************
                       ***********  ********************************
                        * *****     ********************************
                                     ******************************    
                                     *******************************
                                       ****************************
                                        *************************
                                           ******************** **
                                         *** ****************
                                              * *********
                                                   *****
                                                  ******
                                                   ****
                                                     *
```

integration_tests/complex_pow_io.f90
Jatinagarwal24: pls address this comment https://github.com/lfortran/lfortran/pull/11862/changes#r3415798929

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4174,6 +4174,24 @@ LFORTRAN_API void _lfortran_complex_div_64(struct _lfortran_complex_64* a,
...
+    }
+
+    if (b->im == 0.0f && b->re == (float)((int32_t)b->re)) {
+        int32_t n = (int32_t)b->re;
+        if (n == 2) { 
+            result->re = (a->re * a->re) - (a->im * a->im);
+            result->im = 2.0f * a->re * a->im;
+            return;
+        }
+    }
```
Jatinagarwal24: If `b->re` is a very large floating-point number (e.g., `3e10` which exceeds the maximum value of a 32-bit integer), casting it to int32_t is Undefined Behavior (UB) in C. This can cause unexpected crashes or garbage values on different architectures. The Fix: Since the block only optimizes the specific case where $n = 2$, there is no need to cast to an integer at all. You should simplify it to directly check for 2.0:
```
// 32-bit version
if (b->im == 0.0f && b->re == 2.0f) {
    result->re = (a->re * a->re) - (a->im * a->im);
    result->im = 2.0f * a->re * a->im;
    return;
}

// 64-bit version
if (b->im == 0.0 && b->re == 2.0) {
    result->re = (a->re * a->re) - (a->im * a->im);
    result->im = 2.0 * a->re * a->im;
    return;
}

```

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4200,9 +4236,9 @@ LFORTRAN_API void _lfortran_complex_pow_64(struct _lfortran_complex_64* a,
         double complex cb = CMPLX(b->re, b->im);
         double complex cr = cpow(ca, cb);
     #endif
-        result->re = creal(cr);
-        result->im = cimag(cr);
-
+    
+    result->re = creal(cr);
+    result->im = cimag(cr);
```
Jatinagarwal24: This change should not come.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -11972,6 +12008,11 @@ LFORTRAN_API void _lfortran_file_write(int32_t unit_num, int32_t* iostat, const
             }
         }
 
+        
+
```
Jatinagarwal24: remove these blank lines.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4185,12 +4203,30 @@ LFORTRAN_API void _lfortran_complex_pow_32(struct _lfortran_complex_32* a,
...
+    }
+
+    if (b->im == 0.0 && b->re == (double)((int64_t)b->re)) {
+        int64_t n = (int64_t)b->re;
+        if (n == 2) { 
+            result->re = (a->re * a->re) - (a->im * a->im);
+            result->im = 2.0 * a->re * a->im;
+            return;
+        }
+    }
```
Jatinagarwal24: pls change this also see the comments above.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4174,6 +4174,24 @@ LFORTRAN_API void _lfortran_complex_div_64(struct _lfortran_complex_64* a,
...
         struct _lfortran_complex_32* b, struct _lfortran_complex_32 *result)
 {
+    if (a->re == 0.0f && a->im == 0.0f) {
+        // 0 raised to a positive real power is 0. 
+        if (b->re > 0.0f) {
+            result->re = 0.0f;
+            result->im = 0.0f;
+            return;
+        }
+    }
```
Jatinagarwal24, line 4197: This intercepts the MSVC crash for positive powers, but it leaves negative and zero powers exposed (e.g., `(0.0, 0.0) ** (0.0, 0.0)`). If it falls through, `cpow(0, 0)` will still crash MSVC. 

Let's catch the negative and zero cases explicitly to fully protect the runtime:

```suggestion
    if (a->re == 0.0f && a->im == 0.0f) {
        if (b->re > 0.0f) {
            result->re = 0.0f;
            result->im = 0.0f;
        } else if (b->re == 0.0f && b->im == 0.0f) {
            result->re = NAN;
            result->im = NAN;
        } else {
            result->re = INFINITY;
            result->im = NAN;
        }
        return;
    }

File: src/libasr/runtime/lfortran_intrinsics.c [hunk 7]
```diff
@@ -4185,12 +4203,30 @@ LFORTRAN_API void _lfortran_complex_pow_32(struct _lfortran_complex_32* a,
...
 {
+   
+    if (a->re == 0.0 && a->im == 0.0) {
+        // 0 raised to a positive real power is 0. 
+        if (b->re > 0.0) {
+            result->re = 0.0;
+            result->im = 0.0;
+            return;
+        }
+    }
```
Jatinagarwal24: Same issue here as the 32-bit version. We should catch negative and zero powers to ensure MSVC never executes `cpow()` when the base is `(0, 0)`:

```suggestion
    if (a->re == 0.0 && a->im == 0.0) {
        if (b->re > 0.0) {
            result->re = 0.0;
            result->im = 0.0;
        } else if (b->re == 0.0 && b->im == 0.0) {
            result->re = NAN;
            result->im = NAN;
        } else {
            result->re = INFINITY;
            result->im = NAN;
        }
        return;
    }

**Review by Jatinagarwal24 — CHANGES_REQUESTED** [review]

File: integration_tests/complex_pow_io.f90 (hunk 1 above)
Comment (srinjoy933) [reply integration_tests/complex_pow_io.f90]: Hey @Jatinagarwal24 can you please clarify ,exactly what changes you suggest that you want to address through the comment.

Comment (Jatinagarwal24) [reply integration_tests/complex_pow_io.f90]: This test which you added  already works in main.Lets add a test for this issue such that it fails in main and runs with your fix.
This we do to prevent future regressions.

Comment (srinjoy933) [reply integration_tests/complex_pow_io.f90]: Hey @Jatinagarwal24 , I know this tests works in main, but I previously added this ,for this particular case ,lfortran giving wrong output, and I don't think there is any need to add a test ,for this particular part .What do  you think, if you suggest I can remove the test part and only have the changes to get the correct output.

Comment (Jatinagarwal24) [reply integration_tests/complex_pow_io.f90]: No, we dont fix any issue without having test for that.
If we do that it will have future repercussions which we don't want.
So , unfortunately you have to find a test similar to the mre with a test case also.

Comment (srinjoy933) [reply integration_tests/complex_pow_io.f90]: Ok,I understand

**Comment by srinjoy933** [comment]
Hey @Jatinagarwal24 ,I have made the changes that you have suggested and also added the test which fails in main and passes with the changes ,can you please have a look on this pr ,Thank You!

File: integration_tests/complex_pow_io.f90 (hunk 1 above)
Comment (srinjoy933) [reply integration_tests/complex_pow_io.f90]: Hey @Jatinagarwal24 ,as you have suggested I have added a new test ,which is related to the test mentioned in the original issue.

File: integration_tests/complex_pow_io.f90 [hunk 8]
```diff
@@ -0,0 +1,33 @@
...
+            if (escaped) then
+                write(*, "(a)", advance="no") " "
+            else
+                write(*, "(a)", advance="no") "*"
+            end if
+        end do
+        
+        write(*, *) 
+    end do
+end program mandelbrot
```
Jatinagarwal24:
```
program test_complex_pow_precision
    implicit none
    complex(dp) :: z, expected
    real(dp) :: error_re, error_im
    
    ! Start with a negative real value represented as complex
    ! This simulates the center row of the Mandelbrot set where the bug occurred.
    z = cmplx(-1.5_dp, 0.0_dp, dp)
    expected = cmplx(2.25_dp, 0.0_dp, dp)
    
    ! Perform the complex power which triggered cpow() previously
    z = z**2
    
    error_re = abs(z%re - expected%re)
    error_im = abs(z%im - expected%im)
    
    if (error_re > 1.0e-15_dp .or. error_im > 0.0_dp) then
        print *, "Test Failed!"
        print *, "Expected: ", expected
        print *, "Got:      ", z
        print *, "Error Re: ", error_re
        print *, "Error Im: ", error_im
        error stop 1
    end if
    
    print *, "Test Passed!"
end program test_complex_pow_precision

```
I think you can use this test this fails on main and passes with your fix pls once check it,

srinjoy933: I will @Jatinagarwal24

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4200,35 +4174,100 @@ LFORTRAN_API void _lfortran_complex_div_64(struct _lfortran_complex_64* a,
...
+        result->im = 2.0f * a->re * a->im;
+        return;
+    }
+
+    // 3. Fallback to standard library
     #ifdef _MSC_VER
-        _Fcomplex ca = _FCOMPLEX_(a->re, a->im);
-        _Fcomplex cb = _FCOMPLEX_(b->re, b->im);
-        _Fcomplex cr = cpowf(ca, cb);
+        // MSVC's _Fcomplex is buggy. We upcast to double (_Dcomplex)
```
certik: What bug is there for `_Fcomplex` ? Let's reference a github issue with a reproducer here.

src/libasr/runtime/lfortran_intrinsics.c
certik: This seems resolved?

integration_tests/complex_pow_io.f90
```diff
@@ -0,0 +1,34 @@
+program complex_pow_io
+    implicit none
+    integer, parameter :: dp = kind(0.d0)
+    real(dp), parameter :: D = 0.035_dp
+    integer :: ix, iy, k
+    complex(dp) :: c, z
+    logical :: escaped
+
+    do iy = -15, 15
+        do ix = -35, 35
```
certik: Do we need to run this test for all these iterations?

I would only commit a minimal test that actually checks the numbers where it used to fail, not a full mandelbrot calculation whose result is not even checked. That is wasteful and doesn't improve our test coverage.

integration_tests/complex_pow_io.f90
certik: Does this new test fail in main, but now passes?

certik:
I am marking it as draft. @srinjoy933 when you fix or address the comments, please mark it ready for review and we'll review again.

src/libasr/runtime/lfortran_intrinsics.c
srinjoy933: yes ,it is resolved.

src/libasr/runtime/lfortran_intrinsics.c
srinjoy933: I don't find any previous issue but working on this issue I found some problem where MSVC's single-precision cpowf function (using _Fcomplex) is buggy because it violently crashes the Windows compiler on edge cases like zero-base exponentiation by throwing fatal floating-point exceptions. The reliable fix is to bypass cpowf entirely by upcasting to the more stable 64-bit _Dcomplex, performing the math with cpow, and safely casting the result back to a 32-bit float.

integration_tests/complex_pow_io.f90
srinjoy933: I have changed this with a minimal test

PR #11871, opened 2026-06-15
fix: reject real and complex types in select case expression

### LINKED ISSUE #11736: Case and case stmt should only accept limited set of types
```.f90
program p
  implicit none

  complex :: nn = 1


  select case (nn)
  case (nn)
  end select

end program p
```

Refer to this https://claude.ai/share/f47f702a-a33b-4898-a0b1-54efe6b209f2

Madhav1729:
## Summary

`SELECT CASE` expression is now restricted to integer, character, and logical types per Fortran 2018 constraint C1145. Real and complex selectors previously compiled silently without error.

## Changes

- `src/lfortran/semantics/ast_body_visitor.cpp` — added type check in `visit_Select()` rejecting any selector that is not integer, character, or logical
- `tests/errors/continue_compilation_1.f90` — two new subroutines verifying complex and real selectors are rejected

## Verification

Before fix — both compile silently (exit 0, no error):

```fortran
complex :: nn = (1.0, 0.0)
select case (nn)  ! should be an error
```

After fix:

```
semantic error: select case expression must be of type integer, character, or logical
 --> test.f90:4:16
4 |   select case (nn)
  |                ^^
```

Valid types (integer, character, logical) continue to work correctly.

## Test Results

- Reference tests: all pass
- Integration tests (llvm): 3959/3959 passed

fixes #11736


HarshitaKalani (approved):
lgtm. Thanks!

tests/errors/continue_compilation_1.f90
```diff
@@ -752,7 +752,7 @@ subroutine bindc_optional_value(a) bind(c)
     subroutine sub_alternate_return_intrinsic()
         call cpu_time(*1)
 1       continue
-    end subroutine 
+    end subroutine
```
Jatinagarwal24: Hi @Madhav1729 Pls fix this whitespace i think this should not come.

Jatinagarwal24 (changes requested):

tests/errors/continue_compilation_1.f90
Madhav1729: fixed it, thank you!

integration_tests/select_case_04.f90
```diff
@@ -0,0 +1,30 @@
...
+  b = .true.
+  select case (b)
+  case (.true.)
+    print *, "default not taken"
+  case default
+    error stop "expected true branch, not default"
+  end select
+
+  print *, "ok"
+end program
```
Jatinagarwal24: The test which you added is i think not required for this pr as we are solving issue of 
`error not reported` and also this test is working on main
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran select_case_04.f90 
true branch
false branch
default not taken
ok
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran select_case_04.f90 
 true branch
 false branch
 default not taken
 ok
```

Jatinagarwal24 (changes requested):

integration_tests/select_case_04.f90
Madhav1729: I added this as a positive test for valid logical SELECT CASE behavior to complete the test cases for SELECT CASE. You’re right that it doesn’t directly test the error-reporting issue and not relevant to this PR. I will remove it

Madhav1729: Done

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -488,6 +488,18 @@ semantic error: Variable `a` cannot have both the OPTIONAL and VALUE attribute b
...
+    |
+784 |         complex :: nn = (1.0, 0.0)
+    |                    ^^^^^^^^^^^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement
+
+warning: Assuming implicit save attribute for variable declaration
+   --> tests/errors/continue_compilation_1.f90:791:17
+    |
+791 |         real :: x = 1.0
+    |                 ^^^^^^^ help: add explicit save attribute or parameter attribute or initialize in a separate statement
+
```
Jatinagarwal24: Actually these warnings are not needed maybe they are correct but they distract from the actual purpose of this pr  (which is to verify the `select case `semantic error).
You can  remove them by just simplfying the above the test added in `continue_compilation_1.f90`
```
    subroutine select_case_complex()
        implicit none
        complex :: nn
        select case (nn)
        case default
        end select
    end subroutine

    subroutine select_case_real()
        implicit none
        real :: x
        select case (x)
        case default
        end select
    end subroutine
```
If you we will add this we will not get these warnings.

Jatinagarwal24:
@Madhav1729 This pr misses  the type validation for the `case` values themselves. According to the `Fortran Standard (F2018 11.1.9.1), "For a given case-construct, each case-value shall be of the same type as case-expr`".
This can be handled in subsequent pr.I will open a new issue for that.

Jatinagarwal24 (approved):
Looks good to me now.
Thanks @Madhav1729 for the fix.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
Madhav1729: Got it, changing it now.

PR #11872, opened 2026-06-15
fix: reject length specifier (*n) on non-character types

### LINKED ISSUE #3855: invalid length specifier for Integer declaration
For the below Fortran code:
```fortran
program main
    INTEGER, parameter :: i*2 = 1 !> '2' here doesn't mean anything for integers (though valid for 'character string')
    print *, i
end program
```

### What happens currently?

LFortran ignores the length '2' and prints the output '1' without any error/warning.

### What we should do?
We should raise a semantic error, which says that length specifier isn't valid. The same error also for when "parameter" is absent as well.

This is a low priority issue and easy to fix.

Madhav1729:
## Summary

The `*n` length specifier is only valid for character type per the Fortran standard. LFortran was silently accepting it on integer, real, logical, and other types. It now emits a semantic error.

## Changes

- `src/lfortran/semantics/ast_common_visitor.h` — added `else if` branch after the existing `is_char_type && s.m_length` check to reject `*n` on non-character types
- `tests/errors/continue_compilation_1.f90` — new subroutine verifying integer with `*n` is rejected

## Verification

Before fix — silently compiles:

```fortran
integer :: i*2    ! no error
real :: x*4       ! no error
logical :: flag*1 ! no error
```

After fix:

```
semantic error: length specifier (*n) is only valid for character type
```

Valid character usage still works:

```fortran
character :: s*10   ! accepted correctly
```

## Test Results

- Reference tests: all pass
- Integration tests: 3712/3713 passed (`file_open_10` is a pre-existing platform-specific failure unrelated to this fix)

fixes #3855

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -7381,7 +7381,15 @@ public:
                     } else {
                         lhs_type->m_len = char_length;
                     }
-                }                
+                } 
+                else if (!is_char_type && s.m_length) {
+                    diag.add(Diagnostic(
+                        "length specifier (*n) is only valid for character type",
+                        Level::Error, Stage::Semantic, {
+                            Label("", {s.m_length->base.loc})
+                        }));
+                    throw SemanticAbort();
+                }               
                 ASR::Variable_t* variable_added_to_symtab = nullptr;
                 if( std::find(excluded_from_symtab.begin(), excluded_from_symtab.end(), sym) == excluded_from_symtab.end() ) {
                     if ( !is_implicitly_declared && !is_external) {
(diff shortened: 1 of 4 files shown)


HarshitaKalani (approved):
lgtm. Thank you!

opixdown:
@Madhav1729  please resolve the conflicts

HarshitaKalani:
I think you'll need to resolve the conflicts again. Sorry about that. Please ping me once done. I'll merge this.

Madhav1729:
@HarshitaKalani can you merge #11884 first, so that I will fix the conflicts here once done?

Madhav1729:
Closing this, it was already merged by another PR

PR #11876, opened 2026-06-15
fix : added runtime bounds check for pointer bounds/rank remapping

### LINKED ISSUE #11875: Support runtime error for arraysections with incorrect bounds
```.f90
program main
  implicit none

  integer, pointer :: t(:)
  integer, pointer :: u(:)

  allocate(t(5))
  t = 1
  u(0:5) => t !! Should raise runtime error 
  u => t(0:) !! Should raise runtime error 
  print *, u
end program main
```

Jatinagarwal24:
fixes #11875 

* `src/libasr/codegen/asr_to_llvm.cpp`: Modified `handle_pointer_section_target` to dynamically evaluate the target's capacity (`value_size`) and compare it against the requested size of the new pointer section (`current_stride`). If the target is too small, a runtime error (matching `gfortran`'s standard error format) is raised, guarded by the `bounds_checking` option.
* `tests/errors/array_bounds_check_16.f90`: Added a regression test validating that an undersized target raises the appropriate error when pointer remapping is attempted.
* `tests/tests.toml`: Registered the new test to run with the test suite.

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -9535,6 +9535,88 @@ public:
             current_stride = builder->CreateMul(current_stride, size);
         }
         
+        if (compiler_options.po.bounds_checking) {
+            llvm::Type* value_desc_type = nullptr;
+            if (value_physical_type == ASR::array_physical_typeType::DescriptorArray) {
+                ASR::ttype_t* value_desc_asr_type = value_type;
+                if (ASR::is_a<ASR::ArraySection_t>(*x.m_value)) {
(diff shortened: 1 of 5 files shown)


src/libasr/sttHRHbG
HarshitaKalani, line 1: This appears to be an accidentally committed temporary file. Is this required by the implementation, or should it be removed from the PR?

Jatinagarwal24: Yes it is needed.I have removed it.
Thanks for pointing out.

Jatinagarwal24:
hi @HarshitaKalani pls see if this can be merge now or not.

PR #11878, opened 2026-06-16
Fix Trim/Len_Trim function duplication and Shape pointer mismatch

HarshitaKalani:


HarshitaKalani:
1287 - intrinsics_470_DETECT_LEAKS (Failed)

PR #11889, opened 2026-06-16
fix: call inherited finalizers up the type hierarchy

### LINKED ISSUE #11783: Complete finalize-function invocation chain implementation
Address: https://github.com/lfortran/lfortran/pull/11623#issuecomment-4643197900

Madhav1729:
## Summary

Fixes a bug where only the derived type's FINAL procedure was called for intent(out) arguments, skipping parent finalizers. Now all available finalizers are called from child to parent, matching gfortran.

Closes #11783

## Verification

Added intent_out_finalize_05.f90 to verify finalizers run in the correct order across a three-level hierarchy, including a type without a finalizer. The test fails on main, passes with this fix, and matches gfortran. Existing tests still pass.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3174,6 +3174,7 @@ RUN(NAME intent_out_finalize_01 LABELS gfortran llvm)
 RUN(NAME intent_out_finalize_02 LABELS gfortran llvm)
 RUN(NAME intent_out_finalize_03 LABELS llvm) # does not work with GFortran version at CI
 RUN(NAME intent_out_finalize_04 LABELS gfortran llvm EXTRA_ARGS --realloc-lhs-arrays)
+RUN(NAME intent_out_finalize_05 LABELS llvm) # does not work with GFortran version at CI
 RUN(NAME empty_array_01 LABELS gfortran llvm)
 RUN(NAME empty_array_02 LABELS gfortran llvm)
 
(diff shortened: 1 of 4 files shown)


Madhav1729:
There is change in indentation since new loop is added outside the for loop, only minor changes were done.

Madhav1729:
The CI has failed, as far as I can see from the log, I guess it is due to the fact that when we pass var_expr to the finalizer, the type of var_expr is always child_type, when we pass the same var_expr to the parent finalizer, the ASR is generated faitfully but the LLVM IR is not, it is strict about types. So, I have two alternatives, 
1. when we are about to call a finalizer in the LLVM backend (asr_to_llvm.cpp), check if the object type is a child of finalizer's expected parameter type. if so, insert a bitcast before the call. 
2. instead of bitcast, just extract the parent part of the child before calling but I am not sure of how to do this. 

@HarshitaKalani could you please let me know about this?

Madhav1729:
So the CI uses llvm 11 ig and that is why the issue came. I was compiling with llvm19 and then it worked mainly due to non-typed pointers.

Madhav1729:
So here we have a cast in the code generation phase, Although I know it is not recommended, I have just done it as a workaround to test the code. ASR is expected to be fully type correct but in order to do that, whoever creates a` base_final(this)`, should attach a cast so the argument's ASR type is already `base_t`. 

The correct fix would be: wrap the argument in a cast node so the ASR node is type correct. the problem is ASR has no cast kind for a non polymorphic derived type -> parent type. so we need a bigger change. 

I am kindly awaiting for any comments upon this on how to take this further.

Issue #11777, opened 2026-06-07
ICE for passing target array to assumed rank pointer dummy


opixdown:
## MRE

```fortran
program main
  implicit none
  integer, target :: a(2)

  call take_pointer_rank(a)
  print *, "ok"
contains
  subroutine take_pointer_rank(x)
    integer, pointer, intent(in) :: x(..)

    if (any(lbound(x) /= [1])) error stop
  end subroutine take_pointer_rank

end program main
```

## output

```text
(base) opixdown@Atharvs-MacBook-Air lfortran % lfortran t.f90

Internal Compiler Error:
AssertFailed: x->getType()->isPointerTy()
```

## gfortran behavior

```text
(base) opixdown@Atharvs-MacBook-Air lfortran % gfortran t.f90 && ./a.out

 ok
```

PR #11890, opened 2026-06-16
Fix ASR descriptor generation and LLVM lowering for target arrays to assumed-rank pointers

srinjoy933:
The Issue:
When passing a static target array to an assumed-rank pointer dummy argument (e.g., x(..)), LFortran crashes with an Internal Compiler Error in the LLVM backend: AssertFailed: x->getType()->isPointerTy().

This occurs because the ASR optimization passes mistakenly skip generating a temporary array descriptor for pointer arguments, assuming they are already correctly formatted. As a result, the LLVM backend is handed a raw array/by-value struct instead of the expected memory pointer to a descriptor, triggering the assertion failure.

About The PR:
This PR addresses both the root cause in the ASR phase and adds a safeguard in the LLVM backend:

src/libasr/pass/array_struct_temporary.cpp: Updated traverse_call_args to explicitly check for the edge case where the dummy argument is an assumed-rank pointer and the actual argument has the target attribute. This ensures the compiler does not skip the is_temporary_needed block and successfully wraps the target array in an ArrayPhysicalCast descriptor.

src/libasr/codegen/asr_to_llvm.cpp: Added a safeguard to the assumed-rank bitcast block inside convert_call_args. If the backend receives a value that is not an LLVM pointer, it now safely wraps it using CreateAlloca before casting, preventing the isPointerTy() assertion from failing.

Testing
Added a new integration test: integration_tests/assumed_rank_10.f90.

The test specifically verifies that a target array can be successfully lowered and passed to an integer, pointer :: x(..) dummy argument without crashing the compiler.

Partially related to #11777


srinjoy933:
Hey @HarshitaKalani ,can you please have a look on this pr, Thank You!

integration_tests/assumed_rank_15.f90
HarshitaKalani, line 1: This test already works with main. Can you please add a test which works with the changes in this PR and fails in main?

srinjoy933: @HarshitaKalani as you have suggested ,added a test.

PR #11894, opened 2026-06-16
fix: semantic error with string assignment of differing kinds

### LINKED ISSUE #11893: Bug : character different kind assignment not supported
```.f90
program p
    character(len=1,kind=1) :: x
    character(len=1,kind=4) :: y
    y = x
end program 
```

Jatinagarwal24:
fixes #11893
This commit resolves a semantic abort caused by a type mismatch when assigning
character strings with different kinds (e.g., `character(kind=1)` to `character(kind=4)`).

Changes made:
- Added `StringToString` implicit cast kind to `ASR.asdl`.
- Updated `rule_map` and added `StringToString` cast handling in `asr_implicit_cast_rules.h` to implicitly allow conversions between strings of different kinds.
- Handled `StringToString` node lowering in `visit_Cast` inside `asr_to_llvm.cpp`. Since strings in LFortran natively use UTF-8 representation, no underlying conversion is needed for LLVM type lowering during the assignment.


HarshitaKalani (approved):

assem2002:
I think a an ASR reference test here would've been great. If you have time, please add a test

Jatinagarwal24:
> I think a an ASR reference test here would've been great. If you have time, please add a test 

Ok i will do that.

PR #11905, opened 2026-06-17
fix: detect dummy procedure argument interface mismatch (#11778)

### LINKED ISSUE #11778: missing semantic error for dummy procedure argument type mismatch
## MRE

```fortran
program main
  type :: box
  end type

  call takes_proc(actual)

contains

  real function actual(arg)
    type(box) :: arg
    actual = 0.0
  end function actual

  subroutine takes_proc(proc)
    interface
      real function proc(arg)
        import :: box
        class(box) :: arg
      end function proc
    end interface
  end subroutine takes_proc

end program main
```

## output

```text

```

## gfortran behavior

```text

Error: Interface mismatch in dummy procedure 'proc' at (1): Type mismatch in argument 'arg' (CLASS(box)/TYPE(box))
```

Madhav1729:
Fixes #11778 

LFortran did not verify a procedure passed as an actual argument against its dummy procedure's interface. The mismatch in the issue (passing a `type(box)` procedure where `class(box)` is expected) was silently accepted and crashed during LLVM IR verification.

This adds a signature check on procedure arguments, reporting a semantic error for mismatches in argument type/kind, polymorphism (class vs type), distinct derived types, argument count, return type, and function vs subroutine.


Madhav1729:
There are more changes that have to be done to support further mismatches like rank etc, I will create a seperate issue for that

PR #11948, opened 2026-06-22
fix: ICE for mod by zero in array initialization

### LINKED ISSUE #11947: ICE for mod by zero in array initialization
## MRE

```fortran
program p
  logical :: flag(1) = mod([2], 0) == 0
end program p
```

## output

```text

Internal Compiler Error:
AssertFailed: x.m_symbolic_value == nullptr
```

## gfortran behavior

```text

Error: Argument 'P' of MOD at (1) shall not be zero
```

opixdown:
fixes #11947

### DIFF
--- a/src/libasr/intrinsic_func_registry_util_gen.py
+++ b/src/libasr/intrinsic_func_registry_util_gen.py
@@ -1190,6 +1190,18 @@ def add_create_func_return_src(func_name):
     src += indent * 2 + f"Vec<ASR::expr_t*> m_args; m_args.reserve(al, {no_of_args});\n"
     for _i in range(no_of_args):
         src += indent * 2 + f"m_args.push_back(al, args[{_i}]);\n"
+    if func_name in ("Mod", "Modulo"):
+        src += indent * 2 + "ASR::expr_t* divisor_value = ASRUtils::expr_value(m_args[1]);\n"
+        src += indent * 2 + "if (divisor_value && ASR::is_a<ASR::IntegerConstant_t>(*divisor_value) &&\n"
+        src += indent * 4 + "ASR::down_cast<ASR::IntegerConstant_t>(divisor_value)->m_n == 0) {\n"
+        src += indent * 3 + f'append_error(diag, "Second argument of {func_name.lower()} cannot be 0", loc);\n'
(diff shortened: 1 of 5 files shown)


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -1190,6 +1190,18 @@ def add_create_func_return_src(func_name):
...
+        src += indent * 2 + "if (divisor_value && ASR::is_a<ASR::IntegerConstant_t>(*divisor_value) &&\n"
+        src += indent * 4 + "ASR::down_cast<ASR::IntegerConstant_t>(divisor_value)->m_n == 0) {\n"
+        src += indent * 3 + f'append_error(diag, "Second argument of {func_name.lower()} cannot be 0", loc);\n'
+        src += indent * 3 + "return nullptr;\n"
+        src += indent * 2 + "}\n"
+        src += indent * 2 + "if (divisor_value && ASR::is_a<ASR::RealConstant_t>(*divisor_value) &&\n"
+        src += indent * 4 + "ASR::down_cast<ASR::RealConstant_t>(divisor_value)->m_r == 0) {\n"
+        src += indent * 3 + f'append_error(diag, "Second argument of {func_name.lower()} cannot be 0", loc);\n'
+        src += indent * 3 + "return nullptr;\n"
+        src += indent * 2 + "}\n"
```
HarshitaKalani: I don't like doing this. This should be a generic code generation file. Let's not make it specific to some intrinsics.
Can we find a better alternative instead? Let me know if anything else does not work. I'll take a look in that case.

opixdown: okayy harshita i will look into this

ambikeesshh:
hey @opixdown 

what if we add a generic `nonzero_arg` directive instead, mirroring `same_kind_arg`? then set `"nonzero_arg": 1` on Mod and Modulo.
that way the generator stays free of intrinsic-specific code and any future intrinsic can reuse it

HarshitaKalani:
I believe the CI failure is unrelated to this PR : https://github.com/lfortran/lfortran/pull/11975#issuecomment-4791440825

PR #11954, opened 2026-06-22
Add co_max, co_min coarray collectives and refactor prif calls

### LINKED ISSUE #11920: co_sum accepts invalid argument types without a semantic error
`co_sum` currently does not perform semantic type checking on its first argument. Calls with `logical` or `character` arguments compile successfully without any diagnostic, even though `co_sum` should only accept `integer`, `real`, or `complex` arguments according to the PRIF spec.

This happens both in normal compilation and with `--coarray=true`.

## MRE

```fortran
program cosum_bad_type
    logical :: b
    b = .true.
    call co_sum(b)
end program
```

```console
$ lfortran cosum_bad_type.f90"
```

```console
$ lfortran --coarray=true -c cosum_bad_type.f90"
```

A `character` argument is also accepted silently.

This is because the `verify_args` is dead. It is not even called by verify_asr even then it is not a user facing error. the correct fix would be to have type checking in the create function itself. 

Towards #11869

Madhav1729:
Adds the CO_MAX and CO_MIN collectives subroutine for integer, real and character arguments

- Added the frontend ASR Representation
- Did the semantic checks and it fails for types other than what mentioned above
- Added PRIF lowering support for integer, real and character types.
- Added integration test
- Verified execution end to end using Caffeine
- towards #11869 
- Also fixed #11920


tests/coarrays_11.f90
```diff
@@ -0,0 +1,10 @@
+program coarrays_11
+    character(3) :: s
+
+    if (this_image() == 1) s = "abc"
+    if (this_image() == 2) s = "xyz"
```
bonachea: As written, this test only has the well-defined/expected behavior for exactly 2 images.

There's nothing inherently wrong with that assumption, but it only takes a tiny bit more effort to ensure this test runs correctly with 2 **or more** images, which might be a valuable long-term feature for the test suite.




```suggestion
    character(3) :: s = "xyz"

    if (this_image() == 2) s = "abc"
```

tests/coarrays_13.f90
```diff
@@ -0,0 +1,10 @@
+program coarrays_13
+    character(3) :: s
+
+    if (this_image() == 1) s = "xyz"
+    if (this_image() == 2) s = "abc"
```
bonachea: Similarly here


```suggestion
    character(3) :: s = "abc"

    if (this_image() == 2) s = "xyz"
```

src/libasr/pass/coarray.cpp
```diff
@@ -844,6 +844,376 @@ class PRIFInterface {
...
+                fn_symtab, loc, "errmsg", str_type, ASR::intentType::InOut, nullptr,
+                ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Optional, false);
+            ASR::expr_t *errmsg = ASRUtils::EXPR(ASR::make_Var_t(al, loc, errmsg_sym));
+
+            ASR::symbol_t *errmsg_alloc_sym = declare_variable(
+                fn_symtab, loc, "errmsg_alloc", alloc_str_type, ASR::intentType::InOut, nullptr,
+                ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Optional, false);
+            ASR::expr_t *errmsg_alloc = ASRUtils::EXPR(ASR::make_Var_t(al, loc, errmsg_alloc_sym));
```
bonachea: If we're seeking opportunities for factorization, these three arguments (`stat, errmsg, errmsg_alloc`) are common to dozens of PRIF entry points with exactly the same declarations, so this block of code creating those three dummy arguments could probably be factored.

bonachea (commented):
Thanks for working on this @Madhav1729 !

I agree with the comments from @Jainam-not-a-robot, and I've made a few additional minor observations.

src/libasr/pass/intrinsic_subroutines.h
```diff
@@ -1796,6 +1798,155 @@ namespace CoSum {
...
+        ASRUtils::require_impl(
+            ASRUtils::is_integer(*ASRUtils::expr_type(x.m_args[0])) ||
+            ASRUtils::is_real(*ASRUtils::expr_type(x.m_args[0])) ||
+            ASRUtils::is_character(*ASRUtils::expr_type(x.m_args[0])) ,
+            "First argument must be of integer, real or character type",
+            x.base.base.loc, diagnostics);
+    }
+
+    static inline ASR::asr_t* create_CoMax(Allocator& al, const Location& loc,
+            Vec<ASR::expr_t*>& args, diag::Diagnostics& diag) {
```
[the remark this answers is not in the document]
Madhav1729: yes since it should give compilation error if we try to use co_* with any other types than specified

Madhav1729:
@Jainam-not-a-robot, @bonachea I have implemented all of your suggestions, Please look into it. Thank you.

src/libasr/pass/coarray.cpp
```diff
@@ -928,20 +928,40 @@ class PRIFInterface {
...
+            args.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, loc, errmsg_sym)));
+
+            ASR::symbol_t *errmsg_alloc_sym = declare_variable(
+                fn_symtab, loc, "errmsg_alloc", allocatable_deferred_string(), ASR::intentType::InOut, nullptr,
+                ASR::abiType::Source, ASR::accessType::Public,
+                ASR::presenceType::Optional, false);
+            args.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, loc, errmsg_alloc_sym)));
+        }
+
+        ASR::symbol_t* get_or_create_prif_co_reduction_sub(const Location &loc, const std::string &prif_name) {
```
bonachea: Minor: The "co_reduction" part of this name is misleading, because it doesn't handle `co_reduce`. `prif_co_reduce` has a completely different signature and will probably need its own dedicated declaration function once we get there.

Maybe `get_or_create_prif_co_minmaxsum_sub`?

src/libasr/pass/coarray.cpp
```diff
@@ -974,30 +994,83 @@ class PRIFInterface {
...
+            call_args.push_back(al, arg2);
+            call_args.push_back(al, arg3);
+            call_args.push_back(al, arg4);
+            call_args.push_back(al, arg5);
+
+            return ASRUtils::STMT(ASR::make_SubroutineCall_t(
+                al, loc, sub, nullptr, call_args.p, call_args.n, nullptr, false));
+        }
+
+        ASR::symbol_t* get_or_create_prif_co_reduction_character_sub(const Location &loc, const std::string &prif_name) {
```
bonachea: Similarly here, perhaps `get_or_create_prif_co_minmax_character_sub`?

src/libasr/pass/coarray.cpp
```diff
@@ -1011,14 +1084,16 @@ class PRIFInterface {
             return ASR::down_cast<ASR::symbol_t>(fn);
         }
 
-        ASR::stmt_t* make_prif_co_sum_call(const Location &loc,
+        ASR::stmt_t* make_prif_co_reduction_call(const Location &loc, const std::string &prif_name,
```
bonachea: And here maybe `make_prif_co_minmax_call`, assuming you only intend to handle those two.

However I suspect you could also delete `make_prif_co_sum_call` and have `co_sum` call this function as well (in which case maybe name it `make_prif_co_minmaxsum_call`?)

bonachea (commented):
Thanks for the improvements @Madhav1729 !

I've added a few more minor (mostly cosmetic) suggestions. Feel free to ignore if you disagree.

Madhav1729:
> The changes look better now, just the function name seems confusing as @bonachea already said. The CI failure seems independent of this PR, so that's not an issue. Just make the minor changes and then we can run exhaustive tests and merge it

Done.

bonachea (approved):
Thanks for all the improvements!

LGTM based on source inspection.

Madhav1729:
Can we run exhaustive tests now @Jainam-not-a-robot ?

Madhav1729:
@Jainam-not-a-robot @bonachea Ready to be merged

Madhav1729:
@Jainam-not-a-robot Thank you!

PR #11965, opened 2026-06-23
fix: ASR constant evaluation for parameter arrays of derived types

### LINKED ISSUE #11951: component array of derived type parameter not folded
## MRE

```fortran
program component_parameter_array
  implicit none

  type :: item
    character(1) :: tag
  end type

  type(item), parameter :: items(2) = [item("a"), item("b")]
  character(1), parameter :: tags(*) = items%tag

  print *, tags

  if (tags(1) /= "a" .or. tags(2) /= "b") error stop
end program component_parameter_array
```

## output

```text

ASR verify pass error: Initialisation of tags must reduce to a compile time constant.
 --> t.f90:9:30
  |
9 |   character(1), parameter :: tags(*) = items%tag
  |                              ^^^^^^^^^^^^^^^^^^^ failed here
```

## gfortran behavior

```text

 ab
```

Jatinagarwal24:
fixes #11951 

This commit resolves an issue where parameter arrays initialized from derived-type
component references (e.g., `items%tag` where `items` is an array of structs) 
failed to evaluate at compile-time, triggering an ASR verification error.

Changes:
- Updated `getStructInstanceMember_t` in `asr_utils.cpp` to fall back to 
  `m_symbolic_value` when retrieving constant array parameters during semantic analysis.
- Implemented iteration logic to correctly extract structural members from nested 
  `ArrayConstant_t` and `ArrayConstructor_t` AST nodes.
- Introduced an isolated `value_type` for generating constant arrays so that we 
  generate validly-typed `ArrayConstant_t` elements without globally polluting 
  `member_type`, preserving downstream semantic visitor checks.

### DIFF
--- a/integration_tests/struct_parameter_array_01.f90
+++ b/integration_tests/struct_parameter_array_01.f90
@@ -1,0 +1,14 @@
+program struct_parameter_array_01
+  implicit none
+
+  type :: item
+    character(1) :: tag
+  end type
+
+  type(item), parameter :: items(2) = [item("a"), item("b")]
+  character(1), parameter :: tags(*) = items%tag
+
+  print *, tags
+
+  if (tags(1) /= "a" .or. tags(2) /= "b") error stop
+end program struct_parameter_array_01
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1842,6 +1842,7 @@ RUN(NAME parameter_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME parameter_16 LABELS gfortran llvm)
 RUN(NAME parameter_17 LABELS gfortran llvm)
 RUN(NAME parameter_18 LABELS gfortran llvm)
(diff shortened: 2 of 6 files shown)


integration_tests/struct_parameter_array_01.f90
czgdp1807, line 1: I would register this test with `asr=true` in [`tests.toml`](https://github.com/lfortran/lfortran/blob/main/tests/tests.toml) so that we make sure that the ASR of `items%tag` holds the constant value.

This is because the test would anyways work without your diff end-to-end, just without adding the constant value of `items%tag` and figuring out at runtime. So to make sure that the purpose of the diff is served, I would recommend to do what I suggested above.

czgdp1807 (changes requested):
Left a comment for improving testing of your change.

Thank you.

integration_tests/struct_parameter_array_01.f90
Jatinagarwal24: Hi @czgdp1807 I have registered this in `tests.toml` also.
Thanks

Jatinagarwal24:
Hi @czgdp1807 Pls see if this can be merge now .
Thanks !

PR #11975, opened 2026-06-24
fix: ASR verify error for StructType in implicit interface dummy args

### LINKED ISSUE #11974: ASR verify pass error: Variable __caller_iface_implicit_arg_1 of type StructType must have a type declaration.
mre.f90
```
module m1
    implicit none
    type :: runner
        procedure(), nopass, pointer :: caller => null()
    contains
        procedure, pass(this) :: set_caller
    end type
contains
    subroutine set_caller(this, caller)
        procedure() :: caller
        this%caller => caller
    end subroutine
end module

program p
    use m1
    implicit none
    type(runner) :: br
    call br%set_caller(upper_caller)
contains
    subroutine upper_caller(a)
        class(*), intent(in) :: a
    end subroutine
end program
```
Lfortran
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran mre.f90 --implicit-interface
ASR verify pass error: Variable __caller_iface_implicit_arg_1 of type StructType must have a type declaration.
  --> mre.f90:20:24
   |
20 |     call br%set_caller(upper_caller)
   |                        ^^^^^^^^^^^^ failed here


```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran mre.f90 
```

Jatinagarwal24:
fixes #11974

When `create_or_update_implicit_interface` generates dummy arguments
for implicit procedure interfaces, Variables with StructType (e.g.
class(*)) were created with nullptr for m_type_declaration, causing
ASR verification failures.

The fix resolves/creates the ~unlimited_polymorphic_type struct symbol
in three code paths:
- create_or_update_implicit_interface (ast_common_visitor.h)
- reverse propagation arg creation (ast_body_visitor.cpp)
- sync lambda type update (ast_body_visitor.cpp)

For class(*), if the symbol isn't found via resolve_symbol (which
can't cross sibling scopes like module vs program), a new Struct is
created in the target scope mirroring determine_type's logic.

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -8111,9 +8111,20 @@ public:
                             if (existing_ft->n_arg_types == 0 && var_has_no_args) {
                                 Vec<ASR::ttype_t*> arg_types;
                                 arg_types.reserve(al, args.size());
+                                Vec<ASR::symbol_t*> arg_type_decls;
+                                arg_type_decls.reserve(al, args.size());
                                 for (size_t i = 0; i < args.size(); i++) {
                                     if (args[i].m_value == nullptr) continue;
                                     arg_types.push_back(al, ASRUtils::expr_type(args[i].m_value));
+                                    ASR::symbol_t* td = nullptr;
+                                    if (ASR::is_a<ASR::Var_t>(*args[i].m_value)) {
+                                        ASR::symbol_t* sym = ASRUtils::symbol_get_past_external(
+                                            ASR::down_cast<ASR::Var_t>(args[i].m_value)->m_v);
+                                        if (ASR::is_a<ASR::Variable_t>(*sym)) {
+                                            td = ASR::down_cast<ASR::Variable_t>(sym)->m_type_declaration;
+                                        }
+                                    }
+                                    arg_type_decls.push_back(al, td);
                                 }
                                 SymbolTable* parent_scope = current_scope->parent
                                     ? current_scope->parent : current_scope;
@@ -8121,7 +8132,7 @@ public:
                                     proc_var, x.base.base.loc,
                                     arg_types.p, arg_types.size(),
                                     nullptr, parent_scope, sub_name,
-                                    current_scope);
+                                    current_scope, arg_type_decls.p);
                             }
                         }
                     }
@@ -8228,8 +8239,29 @@ public:
                                         }
                                     }
                                     if (!has_type_decl || (!has_arg_info && expected_ft->n_arg_types > 0)) {
+                                        // Extract type_declarations from the parameter function's args
+                                        ASR::symbol_t** param_arg_type_decls = nullptr;
+                                        if (v->m_type_declaration) {
+                                            ASR::symbol_t* param_decl = ASRUtils::symbol_get_past_external(v->m_type_declaration);
+                                            if (ASR::is_a<ASR::Function_t>(*param_decl)) {
+                                                ASR::Function_t* param_func = ASR::down_cast<ASR::Function_t>(param_decl);
+                                                if (param_func->n_args == expected_ft->n_arg_types) {
+                                                    param_arg_type_decls = al.allocate<ASR::symbol_t*>(expected_ft->n_arg_types);
+                                                    for (size_t pi = 0; pi < expected_ft->n_arg_types; pi++) {
+                                                        param_arg_type_decls[pi] = nullptr;
+                                                        if (ASR::is_a<ASR::Var_t>(*param_func->m_args[pi])) {
+                                                            ASR::symbol_t* ps = ASR::down_cast<ASR::Var_t>(param_func->m_args[pi])->m_v;
+                                                            if (ASR::is_a<ASR::Variable_t>(*ps)) {
+                                                                param_arg_type_decls[pi] = ASR::down_cast<ASR::Variable_t>(ps)->m_type_declaration;
+                                                            }
+                                                        }
+                                                    }
+                                                }
+                                            }
+                                        }
                                         create_interface_for_procedure_variable(
-                                            proc_var, passed_arg->base.loc, expected_ft);
+                                            proc_var, passed_arg->base.loc, expected_ft,
+                                            param_arg_type_decls);
                                     } else if (has_arg_info && expected_ft->n_arg_types == 0) {
                                         // Reverse: passed has arg_types but param doesn't.
                                         // Update the parameter's type to match what we're passing.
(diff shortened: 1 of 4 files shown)


Jatinagarwal24:
Hi @assem2002  I am not able to update this test reference i think due to macos thing and i am on linux or if there is issue is with fix then i will resolve that pls once check it.
```
Reference JSON: tests/reference/run-stop_03-897cacd.json
Output JSON:    tests/output/run-stop_03-897cacd.json
Omitting 10 identical items
ACTUAL != REFERENCE
Differing items:
{'stderr_hash': '57bdb7be40ead0f89a7e235da8f6a8befdef69d6dda10848b239d9e5'} != {'stderr_hash': '3e23cf1cb50e4b97d71757a6cc871271723a4247df92d3c07b24a0b0'}
=== STDERR DIFF ===
Reference: tests/reference/run-stop_03-897cacd.stderr
Output:    tests/output/run-stop_03-897cacd.stderr
0a1
> ld: warning: object file (/private/var/folders/8j/sfr9qqcj73j4p6nhwcfpr0th0000gn/T/stop_03.tmp_v7TNztVZhXAtjeCV9ywTis.o) was built for newer 'macOS' version (16.0) than being linked (15.0)
```

HarshitaKalani:
I'm on mac and tested this locally. I think the failure is unrelated to this PR.

HarshitaKalani:
Can you please resolve the conflicts?

Jatinagarwal24:
> Can you please resolve the conflicts?

done.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -9999,11 +9999,42 @@ Result<ASR::TranslationUnit_t*> body_visitor(Allocator &al,
...
+                                                for (auto& sp : search_scope->get_scope()) {
+                                                    ASR::symbol_t* past_ext = ASRUtils::symbol_get_past_external(sp.second);
+                                                    if (ASR::is_a<ASR::Struct_t>(*past_ext)) {
+                                                        arg_type_decl = sp.second;
+                                                        break;
+                                                    }
+                                                }
+                                            }
+                                        }
                                         ASR::symbol_t* arg_sym = ASR::down_cast<ASR::symbol_t>(
```
assem2002: I believe this whole code in this spot is doing some hard-wiring for some specific test case, I believe `implicit_interface_26.f90`.
 ***
I think it's not needed for our case here, let's not use. We'll probably need to refactor that anyways

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -9999,11 +9999,42 @@ Result<ASR::TranslationUnit_t*> body_visitor(Allocator &al,
...
+                                            } else if (search_scope) {
+                                                for (auto& sp : search_scope->get_scope()) {
+                                                    ASR::symbol_t* past_ext = ASRUtils::symbol_get_past_external(sp.second);
+                                                    if (ASR::is_a<ASR::Struct_t>(*past_ext)) {
+                                                        arg_type_decl = sp.second;
+                                                        break;
+                                                    }
+                                                }
+                                            }
+                                        }
```
assem2002: This fetches the very first struct it finds in the containing scope, This is basically incorrect.

assem2002 (changes requested):
I see 2 major changes : 
- Syncing variable's functionType `procedure() :: f_` with the binded function's FuntionType.
- Creating parameters in the interface that gets created for the sake of `procedure() :: f` variable and making sure the `m_type_declaration` member is properly set to the correct `Struct` symbol.
***
We can't follow the current approach as both changes above try to deduce the struct symbol from `StructType` which has no information about the Struct symbol, hence the changes awkwardly fetches a random `Struct` symbol.
***
### What to do ?
- We must pass the `type_declaration` symbols to `create_or_update_implicit_interface`, and from there we'll create and update the interface properly. pass nullptr when there's not and pass the symbol where there's a symbol, pass those symbols for every spot that uses `create_or_update_implicit_interface`.

- For the syncing issue, We'll just assert if we see that we're syncing a type of `StructType` -- I think we don't have a test that will fail, if we see that we'll fail a test, we should find out where the outsync happen and do it there and not wait for the sync. IF THINGS GOT MESSY, LET'S NOT SYNC IF TYPE IS `STRUCTTYPE` AND JUST PROCEED.

Jatinagarwal24:
Hi @assem2002 I have applied the suggested changes.
Pls once review it.
Thanks for the review .

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -8133,8 +8144,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                                         }
                                     }
                                     if (!has_type_decl || (!has_arg_info && expected_ft->n_arg_types > 0)) {
+                                        // Extract type_declarations from the parameter function's args
+                                        ASR::symbol_t** param_arg_type_decls = nullptr;
+                                        if (v->m_type_declaration) {
+                                            ASR::symbol_t* param_decl = ASRUtils::symbol_get_past_external(v->m_type_declaration);
+                                            if (ASR::is_a<ASR::Function_t>(*param_decl)) {
+                                                ASR::Function_t* param_func = ASR::down_cast<ASR::Function_t>(param_decl);
+                                                if (param_func->n_args == expected_ft->n_arg_types) {
```
assem2002:
```suggestion
                                                                                                LCOMPILERS_ASSERT(param_func->n_args == expected_ft->n_arg_types)
```
and let's remove the `if` statement

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -8168,17 +8200,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
-                                            // This handles both "interface exists" and "no interface" cases.
-                                            // For existing interfaces, it shares the source arg_types array
-                                            // directly for cross-scope type propagation (see #11924).
+                                            // Extract type_declarations from the passed function's args.
                                             SymbolTable* callee_scope = f->m_symtab;
                                             SymbolTable* iface_parent = callee_scope->parent ? callee_scope->parent : callee_scope;
                                             std::string var_name = v->m_name;
                                             ASR::ttype_t* return_type = passed_ft->m_return_var_type;
+                                            ASR::symbol_t** fn_arg_type_decls = nullptr;
+                                            if (passed_func->n_args == passed_ft->n_arg_types) {
```
assem2002: same here.
let's  remove the `if` statement and assert instead.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19423,7 +19460,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
             ASR::ttype_t** arg_type_arr, size_t n_arg_types,
             ASR::ttype_t* return_type, SymbolTable* parent_scope,
             const std::string& var_name,
-            SymbolTable* owner_scope = nullptr) {
+            SymbolTable* owner_scope = nullptr,
+            ASR::symbol_t** arg_type_decls = nullptr) {
```
assem2002: I don't want to be picky but.
- I think we should make `arg_type_decls` as a required parameter (we'd need to move it before parameter `owner_scope`.
- Use ` Vec<ASR::symbol_t*>` and assert that its size is == to `n_arg_types`

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19446,10 +19484,12 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
         arg_types_vec.reserve(al, n_arg_types);
         for (size_t i = 0; i < n_arg_types; i++) {
             std::string arg_name = iface_name + "_arg_" + std::to_string(i);
+            // Use the type_declaration passed by the caller (nullptr when not available)
+            ASR::symbol_t* arg_type_decl = (arg_type_decls ? arg_type_decls[i] : nullptr);
```
assem2002: if we'll make `arg_type_decls` as a required parameter, we won't need this.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -19542,7 +19582,8 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
 
     // Convenience wrapper: creates interface from an existing FunctionType.
     void create_interface_for_procedure_variable(ASR::Variable_t* proc_var,
-            const Location& loc, ASR::FunctionType_t* expected_type = nullptr) {
+            const Location& loc, ASR::FunctionType_t* expected_type = nullptr,
```
assem2002: same here.
- required parameter 
- `ASR::Vec` containter and assert

assem2002 (approved):
Very nice! Thanks for working on this.
I mentioned minor comments, please give them a look.

Jatinagarwal24:
Hi @assem2002 I have addressed all the comments.
Thanks for the review !

assem2002 (approved):
LGTM. Thanks.

PR #12005, opened 2026-06-26
fix: lower array assignment of derived types through user-defined assignment(=)

HarshitaKalani:


src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,144 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+        ASR::stmt_t* loop = call_stmt;
+        for (size_t k = 0; k < rank; k++) {
+            size_t dim = rank - 1 - k;
+            ASR::do_loop_head_t head;
+            head.loc = loc;
+            head.m_v = idx_vars[dim];
+            head.m_start = PassUtils::get_bound(target, dim + 1, "lbound",
+                al, 4);
+            head.m_end = PassUtils::get_bound(target, dim + 1, "ubound",
+                al, 4);
```
czgdp1807: Are we still using these? `ASRUtils` also has `get_bound` defined. 

**`PassUtils::get_bound`**

https://github.com/lfortran/lfortran/blob/4b5191b9712d0cac5a854ce2c81fc02ff3ce9f00/src/libasr/pass/pass_utils.cpp#L973-L1015

**`ASRUtils::get_bound`**

https://github.com/lfortran/lfortran/blob/4b5191b9712d0cac5a854ce2c81fc02ff3ce9f00/src/libasr/asr_utils.h#L6762-L6847

The ASRUtils looks more exhaustive and better in its implementation. Are you using `PassUtils::get_bound` because of the `template <typename SemanticAbort>` in `ASRUtils::get_bound`?

src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,144 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+            body.reserve(al, 1);
+            body.push_back(al, loop);
+            loop = ASRUtils::STMT(ASR::make_DoLoop_t(al, loc, nullptr, head,
+                body.p, body.size(), nullptr, 0));
+        }
+        pass_result.push_back(al, loop);
+        return true;
+    }
+
+    ASR::expr_t* make_indexed_expr(ASR::expr_t* expr,
```
czgdp1807: May be you can consider using `PassUtils::create_array_ref`,

https://github.com/lfortran/lfortran/blob/4b5191b9712d0cac5a854ce2c81fc02ff3ce9f00/src/libasr/pass/pass_utils.cpp#L108-L291

czgdp1807 (commented):
Left some suggestions. TY.

src/libasr/asr_utils.cpp
```diff
@@ -2178,6 +2178,41 @@ bool use_overloaded_assignment(ASR::expr_t* target, ASR::expr_t* value,
             expr_dt = value;
         }
     }
+    if(!sym) {
```
certik: **[suggestion]** The new array-of-struct `~assign` lookup is essentially a copy of the existing scalar `StructType` block (resolve `~assign`, walk parents, set `expr_dt`). Because `types_equal` already peels arrays when matching formals, the only real gap was that the first branch required `is_a<StructType_t>(*target_type)` on a non-array type. Duplicating ~30 lines makes future fixes easy to miss on one path only (the new path null-checks `get_struct_sym_from_struct_expr`; the old path does not).

**Suggestion:** Collapse both paths by looking up on `extract_type(target_type)` / `extract_type(value_type)` once, with a shared parent-walk helper, e.g. resolve `~assign` from `get_struct_sym_from_struct_expr` when the extracted type is a struct. That should remove the entire second `if (!sym)` block while covering scalar and array cases.

src/libasr/asr_utils.h
```diff
@@ -6759,9 +6759,8 @@ static inline bool is_pass_array_by_data_possible(ASR::Function_t* x, std::vecto
     return v.size() > 0;
 }
 
-template <typename SemanticAbort>
 static inline ASR::expr_t* get_bound(ASR::expr_t* arr_expr, int dim,
```
certik, line 6807: **[suggestion]** The non-throwing `ASRUtils::get_bound` overload (and the subsequent template wrapper that forwards to it) exists only so `array_op` can call `ASRUtils::get_bound(...)` without diagnostics. The rest of `array_op.cpp` already uses `PassUtils::get_bound`, which is the pass-layer API for bounds. This expands the PR surface in a core header without clear reuse benefit.

**Suggestion:** Drop the `asr_utils.h` `get_bound` split entirely and call `PassUtils::get_bound(target, dim, "lbound"|"ubound", al, get_index_kind())` from the new lowering. That both simplifies the diff and aligns with existing pass conventions.

src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,83 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+        ASR::expr_t* value = x.m_value;
+        ASR::expr_t* value_unbroadcast = ASRUtils::get_past_array_broadcast(value);
+        ASR::ttype_t* target_type = ASRUtils::expr_type(target);
+        size_t rank = ASRUtils::extract_n_dims_from_ttype(target_type);
+        if (rank == 0) return false;
+        bool value_is_scalar = !ASRUtils::is_array(
+            ASRUtils::expr_type(value_unbroadcast));
+
+        Vec<ASR::expr_t*> idx_vars;
+        PassUtils::create_idx_vars(idx_vars, rank, loc, al, current_scope,
```
certik: **[bug]** Index variables are created with the default `index_kind=4`, and loop bounds from `ASRUtils::get_bound` also default to kind 4. Elsewhere in this file (`generate_loop`, file-write lowering, etc.) bounds and index vars use `get_index_kind()` so they respect `descriptor_index_64`. On 64-bit descriptor builds this can produce kind-mismatched loop heads / indices relative to the rest of array lowering.

**Suggestion:** Pass `get_index_kind()` into `PassUtils::create_idx_vars` and into `PassUtils::get_bound` (same pattern as `generate_loop` around the `get_index_kind()` uses).

src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,83 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+            "__libasr_ov_idx_");
+
+        ASR::expr_t* indexed_target = PassUtils::create_array_ref(target,
+            idx_vars, al, current_scope);
+        ASR::expr_t* indexed_value = value_is_scalar
```
certik: **[bug]** When the RHS is a scalar broadcast (`ArrayBroadcast`), `value_is_scalar` is true via `value_unbroadcast`, but `indexed_value` is set to `value` (still the broadcast/array-typed expr) rather than `value_unbroadcast`. The matched assignment procedure has scalar formals, so the call should receive the scalar element, not an `ArrayBroadcast` node. Array-to-array (the test) is fine; scalar-to-array defined assignment is not.

**Suggestion:** Use `value_unbroadcast` when `value_is_scalar` is true: `indexed_value = value_is_scalar ? value_unbroadcast : PassUtils::create_array_ref(...)`.

src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,83 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
         }
     }
 
+    bool try_lower_overloaded_scalar_array_assignment(
```
certik: **[suggestion]** `try_lower_overloaded_scalar_array_assignment` reimplements nested `DoLoop` construction, index vars, and per-element refs, while `generate_loop` / the elemental overloaded path already exist for the same idea. The early placement is justified (array-section assignments otherwise take `generate_loop_for_array_indexed_with_array_indices` and never reach the elemental `m_overloaded` arg collection), but the body still duplicates loop infrastructure.

**Suggestion:** Prefer a thinner special case: (1) detect non-elemental scalar-formal overloaded array assignment early; (2) rewrite/prepare `m_overloaded` args (and `m_dt` if array) and either call a small shared “wrap stmt in rank loops” helper or route into `generate_loop` with the overloaded `SubroutineCall` as the body. At minimum, reuse `PassUtils::get_bound` + `get_index_kind()` and avoid a one-off loop nest. Also consider renaming: the helper handles array-to-array as well as scalar-to-array, so “scalar_array” is misleading.

src/libasr/pass/array_op.cpp
```diff
@@ -1830,6 +1830,83 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
...
+            ? value
+            : PassUtils::create_array_ref(value_unbroadcast, idx_vars, al,
+                current_scope);
+
+        Vec<ASR::call_arg_t> new_args;
+        new_args.reserve(al, sc->n_args);
+        for (size_t i = 0; i < sc->n_args; i++) {
+            ASR::call_arg_t a;
+            a.loc = sc->m_args[i].loc;
+            if (i == 0) {
```
certik: **[suggestion]** Args are rewritten by hard-coded positions (`i == 0` → target, `i == 1` → value) and `sc->m_dt` is left unchanged. That matches the common 2-arg / `m_dt == nullptr` form produced for `type(t)` targets, but is fragile for type-bound / class forms where `insert_self_arg` can place the passed object at `pass_idx` and set `m_dt` to the whole array. Elemental `visit_SubroutineCall` already indexes array `m_dt`.

**Suggestion:** For each `sc->m_args[i].m_value` that is array-typed (or equal to target/value), create an element ref; if `sc->m_dt` is array-typed, index it the same way (or set it to the indexed pass object). Avoid assuming fixed (0,1) layout.

integration_tests/derived_types_149.f90
```diff
@@ -0,0 +1,61 @@
...
+
+   allocate(arr(3))
+   do i = 1, 3
+      allocate(arr(i)%leaf)
+      arr(i)%leaf = i * 100
+      arr(i)%is_temporary = .false.
+   end do
+
+   allocate(copy, source = get_data(arr))
+
```
certik: **[suggestion]** Coverage is a single 1-D section-to-section path inside a function, then `allocate(copy, source=...)`. That is a good Athena-style MRE for pointer association, but does not exercise whole-array assignment (`a = b`), multi-rank, scalar-to-array defined assignment, non-type-bound `interface assignment(=)`, or parent-type `~assign` inheritance (the C++ walks parents but no test does).

**Suggestion:** Add at least one whole-array and one scalar-to-array case (can stay in this file or a sibling). That would also lock in the `value_unbroadcast` fix from Issue 4.

src/libasr/pass/array_op.cpp
```diff
@@ -1863,6 +1940,10 @@ class ArrayOpVisitor: public ASR::CallReplacerOnExpressionsVisitor<ArrayOpVisito
             (ASRUtils::is_simd_array(xx.m_target) && ASRUtils::is_simd_array(xx.m_value)) ) {
             return ;
         }
+        if (try_lower_overloaded_scalar_array_assignment(x)) {
+            remove_original_stmt = true;
```
certik: **[nit]** `remove_original_stmt = true` is redundant here: `transform_stmts` already drops the original statement whenever `pass_result` is non-empty. Other paths in this file set the flag when replacing without always pushing, but after `pass_result.push_back` it has no effect.

**Suggestion:** Either omit the flag (rely on non-empty `pass_result`) or keep it only for consistency with nearby early-exit styles—low priority.

certik (commented):
Here is some review from Grok. If it is not useful, you can ignore it.

certik:
I fixed the comments and resolved conflicts.

certik:
@HarshitaKalani this is now starting to look pretty good, the fix is minimal, and it seems to do the right thing.

Your original fix was lots of code added and it seems it is not needed.

certik:
There was another issue that LFortran behaved differently from Flang and the standard behavior seems to be what Flang did. 

Now the tests should be solid, and the implementation as well, I tried to keep it as simple as possible.

PR #12007, opened 2026-06-26
feat: add support of namelist arrays of derived types

### LINKED ISSUE #12006: namelist arrays of derived types not supported
## MRE

```fortran
program namelist_derived_array_mre
  implicit none

  type :: pair
    integer :: value
  end type pair

  type(pair) :: items(2)

  namelist /settings/ items

  items = [pair(4), pair(9)]

  open(10, status="scratch")
  write(10, nml=settings)
  rewind(10)

  items = [pair(0), pair(0)]

  read(10, nml=settings)

  print *, "items(1)%value = ", items(1)%value
  print *, "items(2)%value = ", items(2)%value
end program namelist_derived_array_mre
```

## output

```text

code generation error: Namelist arrays of derived types are not supported yet
```

## gfortran behavior

```text

 items(1)%value =            4
 items(2)%value =            9
```

opixdown:
fixes  #12006


czgdp1807 (approved):
Looks good.

certik (approved):

certik (approved):
All new errors should have a test.

We should also test more than one member of the struct, and also this case that AI found:

"This represents each component as a separate rank-N item, which changes the value order when the input names the whole derived-type array. For `type(pair) :: items(2)` where `pair` has components `a,b`, the standard input `items=1,2,3,4` assigns `(a,b)` as `(1,2),(3,4)` (Flang does this), but `collect_derived_items()` consumes each descriptor completely before moving to the next, so these descriptors assign `a=[1,2]`, then `b=[3,4]`, producing `(1,3),(2,4)`. Please preserve element-major component order for whole-object assignments and cover a type with at least two components." 

^^^ I think this is a bug.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -16651,8 +16594,22 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                     llvm::Value* member_ptr = llvm_utils->create_gep2(llvm_struct_type, current_ptr, member_idx);
-                    add_namelist_item(prefix + "%" + member_name, member_var->m_type, member_ptr,
+                    ASR::ttype_t *member_type = member_var->m_type;
+                    if (rank > 0) {
+                        member_type = ASRUtils::duplicate_type_with_empty_dims(al, member_type);
+                    }
+                    add_namelist_item(prefix + "%" + member_name, member_type, member_ptr,
                                       member_var->m_type_declaration);
+                    if (rank > 0) {
+                        llvm::Value *item = nml_items.back();
```
certik, line 16604: AI found:

"`add_namelist_item()` can append more than one item here when this member is itself a derived type, but only `nml_items.back()` receives the outer array's rank, stride, and shape. For an array of an outer type containing an inner type with two scalar components, the first inner component remains rank 0 while only the last is treated as an array, so reads/writes silently process the wrong number of elements. Record the item range added by the recursive call and propagate the outer metadata to every leaf descriptor (with a nested-derived-array test)."

certik:
@opixdown  I am going to merge it, to move forward, but please fix the above issues in subsequent PRs. For each issue find an MRE, then fix it.

opixdown:
@certik  i have opened two pr #12195 #12198 for the issues, can you please take a look at them

PR #12013, opened 2026-06-26
fix : ASR verify failure for array intrinsic parameters

### LINKED ISSUE #12008: comparison operator not implemented
## MRE

```fortran
program minloc_mre
  implicit none

  integer, parameter :: grid(2, 2) = reshape([1, 2, 2, 1], [2, 2])
  integer, parameter :: where_is_two(2) = minloc(grid, dim=1, mask=grid > 1, back=.true.)

  print *, "minloc:", where_is_two

  if (any(where_is_two /= [2, 1])) error stop
end program minloc_mre
```

## output

```text

semantic error: Comparison operator not implemented
 --> t.f90:7:11
  |
7 |   if (any(where_is_two /= [2, 1])) error stop
  |           ^^^^^^^^^^^^^^^^^^^^^^
```

## gfortran behavior

```text

 minloc:           2           1
```

Jatinagarwal24:
fixes #12008

This patch resolves an Internal Compiler Error (ICE) triggered by "Non-variable expression in variable definition context" during the ASR verify pass. 

Previously, when parameter initializations involving array intrinsics (e.g., `minloc`) were pulled out into assignments, the `ReplaceFunctionCallReturningArrayVisitor` pass would directly pass the parameter as an `INTENT(OUT)` argument into the generated `SubroutineCall`. This violates the strict verification rule that parameters are non-modifiable and cannot be bound to `INTENT(OUT)` dummies.

The fix introduces a check within `replace_FunctionCall`. If the target variable has `Parameter` storage, the pass now dynamically allocates a temporary intermediate variable, binds that temporary to the subroutine's `INTENT(OUT)` argument, and properly appends an assignment from the temporary back to the parameter.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1839,6 +1839,7 @@ RUN(NAME parameter_14 LABELS gfortran llvm EXTRA_ARGS -fdefault-integer-8 GFORTR
 RUN(NAME parameter_15 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME parameter_16 LABELS gfortran llvm)
 RUN(NAME parameter_17 LABELS gfortran llvm)
+RUN(NAME parameter_18 LABELS gfortran llvm)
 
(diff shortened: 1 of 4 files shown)


czgdp1807 (approved):
Looks good.

PR #12018, opened 2026-06-27
fix: Internal read into assumed shape array fails with LLVM type mismatch

### LINKED ISSUE #12017: Internal read into assumed shape array fails with LLVM type mismatch
## MRE

```fortran
module m
contains
    subroutine read_assumed_shape(values)
        real :: values(:)
        read(*, *) values
        print *, values
    end subroutine
end module

program main
    use m
    real :: x(3)
    call read_assumed_shape(x)
end program
```

## Output

```text

  %35 = getelementptr %array.1, %array.1* %array_descriptor_local, i32 0, i32 0
 float*  call void @_lfortran_read_array_float(float** %35, i32 %51, i32 %34, i32 -1, i32* null)

```

## gfortran behavior

```text
1 2 3
   1.00000000       2.00000000       3.00000000
```

(Input provided: `1 2 3`)

opixdown:
fixes #12017

### DIFF
--- a/integration_tests/read_97.f90
+++ b/integration_tests/read_97.f90
@@ -1,0 +1,29 @@
+module read_97_mod
+contains
+    subroutine read_assumed_shape(unit, values)
+        integer, intent(in) :: unit
+        real, intent(out) :: values(:)
+
+        read(unit, *) values
+    end subroutine
+end module
+
+program read_97
+    use read_97_mod, only: read_assumed_shape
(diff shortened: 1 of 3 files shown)


integration_tests/read_97.f90
HarshitaKalani, line 1: This test already works with main branch. Can you please add a test which fails in main and works with this PR?

PR #12020, opened 2026-06-27
fix: inquire with integer(8) unit causes llvm type mismatch

### LINKED ISSUE #12019: inquire with integer(8) unit causes llvm type mismatch
## MRE

```fortran
program stream_position_inquire
  implicit none

  integer(8) :: unit_number
  integer(8) :: position

  unit_number = 21_8

  open(unit=unit_number, file="stream_position.tmp", access="stream", status="replace")
  write(unit_number) "abcd"
  inquire(unit=unit_number, pos=position)
  close(unit_number, status="delete")

  if (position /= 5_8) error stop 1
  print *, "stream position ok"
end program stream_position_inquire
```

## output

```text

  %22 = load i64, i64* %unit_number, align 8
 i32  call void @_lfortran_inquire(..., i64 %22, ...)

```

## gfortran behavior

```text

stream position ok
```

opixdown:
fixes #12019


HarshilShah1804 (approved):
Looks good. This might be the case for other parameters too, but you can take that up later.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -19201,6 +19162,7 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         if (x.m_unit) {
             this->visit_expr_wrapper(x.m_unit, true);
             unit = tmp;
+            unit = llvm_utils->convert_kind(unit, llvm::Type::getInt32Ty(context));
```
HarshilShah1804, line 19165: I think we should perform this operation only when the the kind is not 4. It is not required in other cases.

HarshitaKalani:
Hi @opixdown , would you like to complete it?

PR #12025, opened 2026-06-28
fix: select type segfault on non-polymorphic selector

### LINKED ISSUE #12023: Segfault for select type on a non-polymorphic variable
## MRE

```fortran
program p
  integer :: a
  select type (a)
  type is (integer)
    print *, "int"
  end select
end program
```

## output

```text
$ lfortran t.f90

```

## gfortran behavior

```text
$ gfortran t.f90
Error: Selector shall be polymorphic in SELECT TYPE statement at (1)
```

This segfaults for any non-polymorphic selector (integer, real, logical, complex); the selector must be polymorphic (`class(*)` or `class(derived)`).

ambikeesshh:
fixes #12023


czgdp1807 (approved):
Looks good. TY.

HarshitaKalani (approved):

PR #12028, opened 2026-06-28
fix: cycle segfault outside a loop

### LINKED ISSUE #12027: Segfault for cycle statement outside a loop
## MRE

```fortran
program p
  cycle
end program
```

## output

```text
$ lfortran t.f90

```

## gfortran behavior

```text
$ gfortran t.f90
Error: CYCLE statement at (1) is not within a construct
```

This segfaults for any `cycle` outside a loop construct (bare `cycle`, or `cycle` inside an `if`/`block`). The sibling `exit` statement is already validated with a clean error; `cycle` is missing the equivalent check.

ambikeesshh:
fixes #12027


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -8725,6 +8725,7 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
 
     void visit_WhileLoop(const AST::WhileLoop_t &x) {
         all_loops_blocks_nesting += 1;
+        loop_nesting += 1;
```
czgdp1807: Is it necessary to maintain counter? I saw below that you are just testing for `loop_nesting == 0`. May be a bool type for `loop_nesting` can be a better alternative to achieve the same?

czgdp1807 (commented):
Left a minor comment.

ambikeesshh:
thanks for the review! replied to the inline comment

src/lfortran/semantics/ast_body_visitor.cpp
ambikeesshh: a bool would break nested loops. when an inner loop exits it'd reset to false even though the outer loop is still active. i checked with `do i / do j / cycle / end do / cycle` and the counter correctly stays at 1 (not 0) after the inner loop ends. a save/restore bool would work but ends up the same number of lines as the counter, so i kept it. happy to change it if you prefer

czgdp1807: > a bool would break nested loops. when an inner loop exits it'd reset to false even though the outer loop is still active.

I remember dealing with this kind of stuff. I used to do the following for setting/unsetting,

```cpp
// setting
bool loop_nesting_copy = loop_nesting;
loop_nesting = true;
.
.
.
// do your stuff
.
.
.
// unsetting or better to say resetting
loop_nesting = loop_nesting_copy;
```

So if `loop_nesting` was originally `true` then `loop_nesting_copy` will hold `true` and give it back to `loop_nesting` at the time of unsetting/resetting.

czgdp1807: Also note that the `loop_nesting` should be initialised as `false` before starting to traverse the tree.

czgdp1807: In addition, to the is just a nice to have but not necessary to apply. The problem with counters is that they can lead to `< 0` stuff (by mistake like double decrements). I leave the decision of applying my suggestion upto you.

czgdp1807 (approved):
https://github.com/lfortran/lfortran/pull/12028#discussion_r3488145283

src/lfortran/semantics/ast_body_visitor.cpp
ambikeesshh: good point on the `< 0` risk, done. switched to a bool with the save/restore pattern you suggested

ambikeesshh:
@czgdp1807 
applied the bool save/restore as suggested, pushed
thanks for the review!

czgdp1807 (approved):
Looks good. TY.

HarshitaKalani (approved):
When you add something in continue_compilation_*.f90 files, let's remove the same number of lines so that the updated reference diff is small. For now, I've checked the diff and it looks good. Thanks!

opixdown:
@ambikeesshh please resolve the conflicts

ambikeesshh:
@opixdown

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -371,139 +371,139 @@ semantic error: LEN parameters in parameterized derived types are not supported
     |         ^^^^^^^^^^^^^^^^^ 
 
 semantic error: Intent attribute can only be applied to procedure arguments
-   --> tests/errors/continue_compilation_1.f90:614:14
+   --> tests/errors/continue_compilation_1.f90:615:14
     |
-614 |     integer, intent(out) :: out_intent
+615 |     integer, intent(out) :: out_intent
```
certik: The diff here is too large. This should be done by first making space, then use it.

ambikeesshh:
moved the cycle test to eof to keep the diff minimal

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -1,3 +1,9 @@
+syntax error: Token 'cycle' is unexpected here
+   --> tests/errors/continue_compilation_1.f90:896:5
+    |
+896 |     cycle
+    |     ^^^^^
```
opixdown: it is a syntax error not the error you added, it is changing the whole stderr file

tests/errors/continue_compilation_1.f90
```diff
@@ -511,7 +511,6 @@ program continue_compilation_1
     print *, ieor()
     print *, min(c, c)
     exit
-    cycle
```
opixdown: i think this was the correct place, when you add one line just remove one line to keep the diff minimum,

ambikeesshh: done exactly that

opixdown: thnk you LGTM!!

certik (approved):

PR #12042, opened 2026-06-29
fix: len_trim and repeat ICE with character(kind=4)

### LINKED ISSUE #12041: ICE for len_trim and repeat with character(kind=4)
## MRE

```fortran
program p
  character(kind=4, len=5) :: s = 4_"ab"
  print *, len_trim(s)
  print "(A)", trim(s)
  print "(A)", repeat(4_"x", 3)
end program
```

## output

```text
$ lfortran t.f90

LCOMPILERS_ASSERT_MSG(check_equal_type(arg_type, orig_arg_type, arg_expr, orig_arg_expr), ...)
```

## gfortran behavior

```text
$ gfortran t.f90 && ./a.out
           2
 ab
 xxx
```

This ICEs for `len_trim`, `trim`, and `repeat` when the argument is `character(kind=4)`. The instantiated function hardcodes `kind=1` for the string parameter, so passing a `kind=4` argument triggers a type-mismatch assertion. gfortran handles all three correctly.

ambikeesshh:
fixes #12041

### DIFF
--- a/integration_tests/char4_string_intrinsics_01.f90
+++ b/integration_tests/char4_string_intrinsics_01.f90
@@ -1,0 +1,12 @@
+program char4_string_intrinsics_01
+  implicit none
+  character(kind=4, len=5) :: s = 4_"ab"
+
+  ! len_trim with character(kind=4) previously ICE'd
+  if (len_trim(s) /= 2) error stop 1
+
+  ! repeat with character(kind=4) previously ICE'd
+  if (repeat(4_"x", 3) /= 4_"xxx") error stop 2
+
+  print *, "ok"
+end program
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4636,6 +4636,7 @@ RUN(NAME pure_side_effects_03 LABELS gfortran llvm)
 RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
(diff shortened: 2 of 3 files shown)


opixdown:
LGMT, thnak you @ambikeesshh

ambikeesshh:
@assem2002 PTAL

integration_tests/char4_string_intrinsics_01.f90
```diff
@@ -0,0 +1,12 @@
+program char4_string_intrinsics_01
+  implicit none
+  character(kind=4, len=5) :: s = 4_"ab"
```
assem2002, line 3: character(kind=4, len=1) :: s2

```suggestion
  character(kind=4, len=5) :: s = 4_"ab"
  character(kind=4, len=1) :: s2
```

integration_tests/char4_string_intrinsics_01.f90
```diff
@@ -0,0 +1,12 @@
...
+
+  ! len_trim with character(kind=4) previously ICE'd
+  if (len_trim(s) /= 2) error stop 1
+
+  ! repeat with character(kind=4) previously ICE'd
+  if (repeat(4_"x", 3) /= 4_"xxx") error stop 2
```
assem2002, line 9:
```suggestion
  s2 = "x"
  if (repeat(s2, 3) /= 4_"xxx") error stop 2
```

assem2002 (approved):
LGTM! Thanks
***
I can see that `repeat` still fails when we use runtime variables; The current test pass for `repeat` as we rely on compiler's compile time evaluation, while using runtime variable actually calls a function to repeat the string, and there we happen to have a bug. I opened issue #12139

PR #12052, opened 2026-06-30
Support for co_broadcast

### LINKED ISSUE #11579: co_broadcast not implemented
This program
```
PROGRAM P
IMPLICIT NONE

INTEGER :: I[*], STAT
CALL S(I, STAT)

CONTAINS

SUBROUTINE S(I_ARG, STAT_ARG)
IMPLICIT NONE
INTEGER, INTENT(IN) :: I_ARG[*], STAT_ARG
CALL CO_BROADCAST (I_ARG, 1, STAT_ARG)    
END SUBROUTINE S

END PROGRAM P
```
found a bug in ifx, and lfortran 0.63.0 would not compile it.
```
(lf) john:~/Gstd23/later$ lfortran cobroadcast.f90
semantic error: Coarrays are not supported yet
  --> cobroadcast.f90:12:1
   |
12 | CALL CO_BROADCAST (I_ARG, 1, STAT_ARG)    
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
```

Madhav1729:
Support for co_broadcast collective support for intrinsic types

 - Rejects types other than intrinsic types
 - Pending support for derived types (and with allocatables) 
 - Added tests

Towards #11869 

Fixes #11579


Madhav1729:
@bonachea I'd like to discuss `co_broadcast` in this minimal PR, which adds support for `co_broadcast` on intrinsic types.

First, I refactored `make_prif_co_minmaxsum_call` into `make_prif_co_broadcast_call`. With a few additional arguments, the same helper can now be reused for both the collective reduction intrinsics and `co_broadcast`, reducing code duplication.

I'm happy to further modify the refactoring based on your feedback if you think there's a better approach.

tests/coarrays_17.f90
bonachea, line 1: These two examples are currently virtually identical, and leave most important cases untested. 

I'd like to see at least:
* test coverage for arrays of intrinsics (assuming that's accepted)
* test coverage for `character` types
* test coverage for `real` and/or `complex`

src/libasr/pass/coarray.cpp
```diff
@@ -953,7 +953,9 @@ class PRIFInterface {
             args.push_back(al, ASRUtils::EXPR(ASR::make_Var_t(al, loc, errmsg_alloc_sym)));
         }
 
-        ASR::symbol_t* get_or_create_prif_co_minmaxsum_sub(const Location &loc, const std::string &prif_name) {
+        ASR::symbol_t* get_or_create_prif_collective_sub(const Location &loc, const std::string &prif_name,
+                const std::string &image_arg_name = "result_image",
+                ASR::presenceType image_presence = ASR::presenceType::Optional) {
```
bonachea, line 996: I like this factorization!

However IMHO when it eventually comes time to do `prif_co_reduce` (the remaining F23 collective) in a future PR, we should _not_ try to squeeze `prif_co_reduce` into this same function, because the interface is very different.

bonachea (changes requested):
Thanks for working on this!

I'd like to see slightly broader test coverage of the permitted types, especially given co_broadcast is frequently called with array arguments.

Otherwise LGTM!

src/libasr/pass/coarray.cpp
Madhav1729: Yes, I was aware of that while changing it, we cannot use the same for `co_reduce`. I believe it is better to make it usable for co_broadcast since we already have one for max,min,sum :)

So, I will keep it intact for now

tests/coarrays_17.f90
Madhav1729: oh mb, I missed the actual test file while pushing. I can do one thing - I will keep the test 16 as minimal baseline and make 17 a substantial one covering all others

bonachea (approved):
LGTM!

bonachea:
@assem2002 this PR needs the Tests::Run-Exhaustive label to exercise the new changes

Madhav1729:
Yes, it is valid but the current plan is to implement it just for intrinsic
types as discussed with Dan. I will do a separate PR for that this week
after discussing the implementation plan.

Madhav1729:
Okay

src/libasr/pass/intrinsic_subroutines.h
```diff
@@ -1956,6 +1957,84 @@ namespace CoMin {
...
+
+    static inline ASR::asr_t* create_CoBroadcast(Allocator& al, const Location& loc,
+            Vec<ASR::expr_t*>& args, diag::Diagnostics& diag) {
+        ASR::ttype_t* arg_type = ASRUtils::expr_type(args[0]);
+        if (!ASRUtils::is_integer(*arg_type) && !ASRUtils::is_real(*arg_type)
+                && !ASRUtils::is_complex(*arg_type) && !ASRUtils::is_character(*arg_type)
+                && !ASRUtils::is_logical(*arg_type)) {
+            diag.add(diag::Diagnostic(
+                "`a` argument of `co_broadcast` must be of integer, real, complex, character or logical type, but got " +
+                    ASRUtils::type_to_str_fortran_expr(arg_type, args[0]),
```
bonachea: Towards @Jainam-not-a-robot 's point, I think the near-term issue here is that derived type arguments to `co_broadcast` are permitted by Fortran, but are not yet implemented in LFortran.

I suggest we resolve this by altering this error message to clarify the status, something like:

```suggestion
                "`a` argument of `co_broadcast` must currently be of integer, real, complex, character or logical type, but got " +
                    ASRUtils::type_to_str_fortran_expr(arg_type, args[0]) +
                    " which is not yet supported",
```

src/libasr/pass/intrinsic_subroutines.h
```diff
@@ -1956,6 +1957,84 @@ namespace CoMin {
...
+        m_args.push_back(al, args[0]);
+        for (size_t i = 1; i < args.size(); i++) {
+            if (args[i]) {
+                m_args.push_back(al, args[i]);
+            }
+        }
+        return ASR::make_IntrinsicImpureSubroutine_t(al, loc,
+            static_cast<int64_t>(IntrinsicImpureSubroutines::CoBroadcast),
+            m_args.p, m_args.n, 0);
+    }
```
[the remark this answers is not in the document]
bonachea: @Madhav1729 please do open a new dedicated issue for `co_broadcast` on derived types, to track status and design discussions after this PR is merged.

bonachea (changes requested):
I'd definitely like to see us merge support for `co_broadcast` on intrinsic types first; this is relatively straightforward and very very impactful to end users.

A future PR can tackle `co_broadcast` on derived types, which introduces many subtle challenges and design questions. See [this issue comment](https://github.com/lfortran/lfortran/issues/11869#issuecomment-4811641451) for more details.

Madhav1729:
Done, I have removed the MRE from continue_compilation for now, as asked.

src/libasr/pass/intrinsic_subroutines.h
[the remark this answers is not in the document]
bonachea: #12107

bonachea (approved):
LGTM!

assem2002:
I'll give it a quick review, sorry for the delay

tests/tests.toml
```diff
@@ -2922,6 +2922,18 @@ fortran = true
 pass = "coarray"
 options = "--coarray=true"
 
+[[test]]
+filename = "coarrays_17.f90"
+fortran = true
+pass = "coarray"
+options = "--coarray=true"
```
assem2002, line 2929: I think that doesn't need the fortran dump

assem2002 (commented):
LGTM. Thanks for PR.
***
I put it on auto merge, please resolve conflicts.

tests/tests.toml
bonachea: I very much appreciate the fortran dumps and would strongly advocate that we continue using them liberally. 

This is the primary mechanism I've been using to identify correctness problems in these PRIF PRs by direct review, rather than having to go build the compiler from a PR branch and then try to figure out why tests fail at runtime.

assem2002: Yeah. I strongly agree that fortran dump is great visual correctness check.
I think it's okay to keep the dump for now at this development stage.

when we see that we have numerous coarray tests, we'll cut the number of dump tests as it will be costly at the CI for no reason. maybe let's add like a comment or something to differentiate between ones that actually aim for a reference check and ones that are aiming for the runtime check, so we can know them later easily.

Madhav1729:
@bonachea @Jainam-not-a-robot @assem2002 I have resolved the conflicts, ready to be merged!

PR #12056, opened 2026-06-30
fix : ICE for elemental int on zero-sized parameter array  

### LINKED ISSUE #12051: ICE for elemental int on zero-sized parameter array
## MRE

```fortran
program test
  integer, parameter :: xs(0) = 1

  print *, int(xs)
end program test
```

## output

```text

Internal Compiler Error:
AssertFailed: f != nullptr
```

## gfortran behavior

```text

(compiles and runs successfully with no printed values)
```

Jatinagarwal24:
fixes #12051
Elemental intrinsics (e.g. `int()`, `abs()`) applied to zero-sized
parameter arrays caused an AssertFailed (f != nullptr) because the
broadcast loop never executed, producing an empty ArrayConstructor
whose expr_value() was nullptr.

Add an early return when max_array_size == 0 to keep the original
(empty) result_array unchanged.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3200,6 +3200,7 @@ RUN(NAME empty_array_01 LABELS gfortran llvm)
 RUN(NAME empty_array_02 LABELS gfortran llvm)
 
 RUN(NAME empty_array_03 LABELS gfortran llvm)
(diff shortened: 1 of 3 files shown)


opixdown:
LGTM thank you @Jatinagarwal24

PR #12057, opened 2026-07-01
Fix SHAPE semantic checking for scalar and assumed-rank arguments

### LINKED ISSUE #11674: ICE for dot_product with shape of scalar argument
## MRE

```fortran
program dot_product
  implicit none

  integer, parameter :: folded = dot_product(shape(42), shape(42))

  if (folded /= 0) error stop
end program dot_product
```

## output

```text

Internal Compiler Error:
AssertFailed: ASR::is_a<ASR::ArrayConstant_t>(*vector_a) && ASR::is_a<ASR::ArrayConstant_t>(*vector_b)
```

mohitdhingra22:
FIXES #11674 

````markdown
# Fix semantic checking of SHAPE intrinsic for scalar arguments

## Description

This PR fixes incorrect semantic checking of the `SHAPE` intrinsic.

Previously, `SHAPE` validation was using the number of dimensions (`n_dims`) to determine whether the argument was an array.

The existing logic treated:

```cpp
n_dims == 0
````

as a scalar argument.

While this works for normal scalar values, it is incorrect for assumed-rank arrays because assumed-rank arrays do not have compile-time rank information available in ASR.

As a result, valid assumed-rank array usage of `SHAPE` was incorrectly rejected.

---

## Problem

For example, this should be valid:

```fortran
subroutine test(input)

    real :: input(..)

    print *, shape(input)

end subroutine
```

`input(..)` is an assumed-rank array.

However, internally it also has:

```
n_dims == 0
```

because the rank is determined at runtime.

The previous implementation could not distinguish between:

### A scalar:

```fortran
shape(42)
```

where:

```
n_dims = 0
```

and

### An assumed-rank array:

```fortran
shape(input)
```

where:

```
n_dims = 0
```

but the value is still an array.

This caused valid assumed-rank tests such as:

* `integration_tests/assumed_rank_07.f90`
* `integration_tests/assumed_rank_09.f90`

to fail.

---

## Solution

Instead of using:

```cpp
n_dims == 0
```

as a scalar check, this PR uses the existing ASR utility:

```cpp
ASRUtils::is_assumed_rank_array()
```

to correctly identify assumed-rank arrays.

The new logic checks:

```cpp
if (n_dims == 0 &&
    !ASRUtils::is_assumed_rank_array(expr_type(args[0]))) {
    
    // reject scalar argument
}
```

Now the compiler distinguishes between:

### Scalar input

Example:

```fortran
shape(42)
```

Result:

```
n_dims == 0
assumed_rank == false
```

The argument is rejected correctly.

---

### Assumed-rank array input

Example:

```fortran
subroutine foo(a)

real :: a(..)

print *, shape(a)

end
```

Result:

```
n_dims == 0
assumed_rank == true
```

The argument is accepted correctly.

---

---

## Regression Test

Added a semantic error regression test:

```
tests/errors/shape_01.f90
```

with:

```fortran
program shape_error
    implicit none

    integer, parameter :: folded = dot_product(shape(42), shape(42))

    if (folded /= 0) error stop

end program shape_error
```

This verifies that invalid scalar usage:

```fortran
shape(42)
```

continues to produce the expected semantic error:

```
semantic error: The argument of SHAPE must be an array
```

Registered the test in:

```
tests/tests.toml
```

using:

```toml
[[test]]
filename = "errors/shape_01.f90"
asr = true
```

---

## Testing Done

Ran the new regression test:

```bash
./run_tests.py -t shape_01 -u
```

and then:

```bash
./run_tests.py -t shape_01
```

Result:

```
TESTS PASSED
```

Also verified integration tests related to SHAPE and assumed-rank behavior.

---


opixdown:
the ci fails, please fix it

mohitdhingra22:
> the ci fails, please fix it

Can u help with it

PR #12058, opened 2026-07-01
fix: reject length specifier on non-character type

### LINKED ISSUE #3855: invalid length specifier for Integer declaration

VossProject:
Fixes #3855

## Summary

The `*2` in `integer :: i*2` is a character-length specifier, which is only valid on `CHARACTER`. On any other type, LFortran silently drops it and compiles the line as `integer :: i`. It now raises a semantic error instead, whether or not `parameter` is present.

## Scope

- `src/lfortran/semantics/ast_common_visitor.h`: when a declaration puts a length specifier on a non-character type, raise a semantic error and stop, instead of dropping it. The `CHARACTER` length handling is unchanged.
- `tests/errors/continue_compilation_1.f90`: adds the two cases above (`integer :: i*2` and `integer, parameter :: j*2 = 1`) with their regenerated reference output.

## Verification

- On `main`, both cases compile and print `1` (exit 0).
- On this branch, both raise `semantic error: length specifier is not allowed for type 'integer'`.
- `./run_tests.py -t continue_compilation_1` passes, and no other reference output changed.

## Rationale

- A length specifier only means something on `CHARACTER`. Silently ignoring one on another type hides a mistake the programmer would want caught.
- The error points at the entity, so `integer :: a, i*2, b` underlines just `i*2` rather than the whole line.
- The message leaves out the kind on purpose: the problem is the length syntax, not whether the type is `integer(4)` or `integer(8)`.


tests/errors/continue_compilation_1.f90
```diff
@@ -836,4 +836,10 @@ subroutine select_type_nonpolymorphic()
             print *, a
         end select
     end subroutine
+
+    subroutine length_specifier_non_character()
+        implicit none
+        integer :: i*2  ! {Error} length specifier is not allowed for type 'integer'
```
opixdown: you can remove the `! {Error} length specifier is not allowed for type 'integer'` comment as one can see the error in stderr file

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -494,6 +494,18 @@ semantic error: Function `c_loc` is not permitted in an initialization expressio
...
+semantic error: length specifier is not allowed for type 'integer'
+   --> tests/errors/continue_compilation_1.f90:842:20
+    |
+842 |         integer :: i*2  ! {Error} length specifier is not allowed for type 'integer'
+    |                    ^^^ 
+
+semantic error: length specifier is not allowed for type 'integer'
+   --> tests/errors/continue_compilation_1.f90:843:31
+    |
+843 |         integer, parameter :: j*2 = 1  ! {Error} length specifier is not allowed for type 'integer'
```
opixdown: we are checking same thing 2 times

opixdown:
@VossProject  thankyou for the pr, just address the comment and ping me when done

VossProject:
> @VossProject thankyou for the pr, just address the comment and ping me when done

@opixdown 
Thanks! I'm not seeing the specific comment on my end, could you point me to it or re-submit the review?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7554,7 +7554,15 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                     } else {
                         lhs_type->m_len = char_length;
                     }
-                }                
+                } else if (s.m_length) {
+                    diag.add(Diagnostic(
+                        "length specifier is not allowed for type '"
```
opixdown: i think better error message could be something like
`length specifier is only valid for character type`

VossProject:
Thanks for the feedback @opixdown, addressed all three in my latest commit.

opixdown:
@VossProject  thank you LGTM

VossProject:
@opixdown thanks again for the review.

Tests came back good and the PR is free of conflict now.

Is there anything else you need from me, or is it just waiting on a maintainer to merge at this point?

Issue #10939, opened 2026-04-02
Plan: GPU offloading for `do concurrent` via a new GPU ASR dialect


certik:
## Summary
 
Add a small set of GPU primitives to ASR and use them to lower `do concurrent` directly to Metal/CUDA/HIP device code + LLVM host code, without routing through OpenMP. This gives us a clean, minimal GPU offloading path that we fully control, and that can deliver working `do concurrent` GPU execution quickly.
 
## Motivation
 
We currently have one partial path for GPU offloading:
 
- **do concurrent → OpenMP transformation** (unfinished PR #10856) and **OpenMP via C backend** (hardwired in the C backend (#10857), not composable and using C for both host and device)
 
The fundamental issue is that OpenMP is a directive language that prescribes *how* to parallelize, while `do concurrent` asserts *that* iterations are independent and lets the compiler choose. Routing through OpenMP loses information (the independence guarantee, multi-dimensional index structure, purity invariants) and adds complexity. The idea is that in the do concurrent -> OpenMP pass we would choose how to parallelize it, but we are already making some choices early (before a transformation to a kernel).
 
A direct path — `do concurrent` → GPU ASR primitives → device code + host LLVM IR — is simpler, gives us full control, and maps naturally to the `do concurrent` semantics. The GPU surface area needed for `do concurrent` is tiny: about 6 new ASR node types. The idea is that we transform to a kernel first, without losing semantics.

Ultimately both approaches should work and we can keep developing both, to see which one ends up better. But the direct path (this issue) should allow us to deliver faster, since we control the whole stack and do concurrent is simpler than OpenMP. The advantage of going via OpenMP is that if done well, it should allow to reuse the work for both CPU and GPU parallelism (currently however the CPU OpenMP transformation doesn't seem to be directly reusable, so would need refactoring anyway).
 
### Architecture
 
```
do concurrent
    │
    ├── CPU path: do concurrent → OpenMP parallel do → LLVM libopenmp (existing)
    │
    └── GPU path: do concurrent → GPU ASR dialect → device source (Metal/CUDA/HIP)
                                                   + host LLVM IR
                                                   + platform-independent runtime
```
 
Selected by compiler flag: `--gpu=metal`, `--gpu=cuda`, `--gpu=hip`. The CPU OpenMP path remains a separate effort for multicore CPU parallelism.
 
### New ASR nodes (the GPU dialect)
 
Six new node types, nothing more:
 
| ASR Node | Purpose |
|----------|---------|
[...]

PR #12073, opened 2026-07-02
test: cover gpu_offload lowering for simple do concurrent

jassem-manita:
Related to #10939
This PR adds an ASR-level regression test for the existing `gpu_offload` pass using a simple SAXPY-style `do concurrent` loop.


jassem-manita:
@certik

PR #12078, opened 2026-07-02
fix : return type for count intrinsic with kind argument

### LINKED ISSUE #12077: comparison of count result is not treated as logical mask for any
## MRE

```fortran
program test
  integer, parameter :: k = kind(0)
  logical :: a(4, 5)

  a = .true.

  print *, count(a, 2, kind=k)

  if (any(count(a, 2, kind=k) /= 5)) stop 1
end program test
```

## output

```text

semantic error: `mask` argument of `Any` intrinsic must be a logical array
 --> t.f90:7:11
  |
7 |   if (any(count(a, 2, kind=k) /= 5)) stop 1
  |           ^^^^^^^^^^^^^^^^^^^^^^^^
```

## gfortran behavior

```text

           5           5           5           5
```

Jatinagarwal24:
fixes #12077 

This commit fixes a bug where `count` invocations returning an array were incorrectly typed as a scalar if the `kind` argument was provided. 

Bug:
In `Count::create_Count`, when the `kind` argument was supplied, the `return_type` was unconditionally overwritten to a scalar `Integer_t` of the given kind. This ignored the `dim` argument's effect on the return shape. Consequently, expressions like `count(a, dim, kind=k)` returned a scalar integer instead of an integer array, which in turn caused subsequent operations (like `count(...) /= 5`) to evaluate to a scalar logical instead of a logical array, triggering a semantic error when passed to `any()`.

Fix:
Modified the logic in `Count::create_Count` to first determine the `base_type` from the `kind` argument (defaulting to `int32`), and then correctly construct the `return_type` either as a scalar of `base_type` or an array of `base_type` depending on the presence of the `dim` argument and the rank of the mask.


HarshitaKalani (approved):
lgtm. Thanks!

Issue #11870, opened 2026-06-15
Compile json-fortran


opixdown:
### Package link
https://github.com/jacobwilliams/json-fortran
### Here are the steps for building with lfortran:
```
git clone https://github.com/jacobwilliams/json-fortran
cd json-fortran
fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
```
### Current Status
```
(base) opixdown@Atharvs-MacBook-Air json-fortran % fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
 + mkdir -p build/dependencies
json_kinds.F90                         done.
test_iso_10646_support.f90             done.
json_parameters.F90                    done.
json_string_utilities.F90              done.
json_value_module.F90                  done.
json_file_module.F90                   done.
json_module.F90                        done.
libjson-fortran.a                      done.
jf_test_50.F90                         failed.
[  7%] Compiling...
ASR verify pass error: Allocating a variable that's a string of deferred length requires providing a length to allocate with
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 2994
    LCompilers::print_stack_on_segfault();
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 2849
    result = compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/Users/opixdown/lfortran/src/bin/lfortran.cpp", line 1294
    std::cerr << diagnostics.render(lm, compiler_options);
  File "/Users/opixdown/lfortran/src/lfortran/fortran_evaluator.cpp", line 430
    if (res.ok) {
  File "/Users/opixdown/lfortran/src/libasr/codegen/asr_to_llvm.cpp", line 26734
    pass_manager.apply_passes(al, &asr, co.po, diagnostics);
  File "/Users/opixdown/lfortran/src/libasr/pass/pass_manager.h", line 324
    apply_passes(al, asr, _passes, pass_options, diagnostics, cummulative_time_taken_by_passes_in_microseconds);
  File "/Users/opixdown/lfortran/src/libasr/pass/pass_manager.h", line 197
    throw LCompilersException("Verify failed in the pass: "
LCompilersException: Verify failed in the pass: nested_vars
<ERROR> Compilation failed for object " test_jf_test_50.F90.o "
[...]

PR #12083, opened 2026-07-02
fix: append trailing newline before preprocessing source

HarshitaKalani:
https://github.com/lfortran/lfortran/issues/11870

### DIFF
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -245,7 +245,14 @@ Result<LFortran::AST::TranslationUnit_t*> FortranEvaluator::get_ast2(
     if (compiler_options.c_preprocessor) {
         // Preprocessor
         LFortran::CPreprocessor cpp(compiler_options);
-        Result<std::string> res = cpp.run(code_orig, lm, cpp.macro_definitions, diagnostics);
+        const std::string *cpp_input = &code_orig;
+        std::string cpp_input_with_newline;
+        if (!code_orig.empty() && code_orig.back() != '\n') {
+            cpp_input_with_newline = code_orig;
+            cpp_input_with_newline.push_back('\n');
+            cpp_input = &cpp_input_with_newline;
+        }
+        Result<std::string> res = cpp.run(*cpp_input, lm, cpp.macro_definitions, diagnostics);
         if (res.ok) {
             tmp = res.result;
         } else {
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4029,6 +4029,10 @@ RUN(NAME cpp_pre_16 LABELS gfortran llvm
     EXTRA_ARGS --cpp
     GFORTRAN_ARGS -cpp)
 
+RUN(NAME cpp_pre_17 LABELS gfortran llvm
+    EXTRA_ARGS --cpp
+    GFORTRAN_ARGS -cpp)
+
 RUN(NAME dabs_01 LABELS gfortran llvmImplicit)
 
 RUN(NAME minpack_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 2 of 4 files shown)


src/lfortran/fortran_evaluator.cpp
```diff
@@ -245,7 +245,14 @@ Result<LFortran::AST::TranslationUnit_t*> FortranEvaluator::get_ast2(
     if (compiler_options.c_preprocessor) {
         // Preprocessor
         LFortran::CPreprocessor cpp(compiler_options);
-        Result<std::string> res = cpp.run(code_orig, lm, cpp.macro_definitions, diagnostics);
+        const std::string *cpp_input = &code_orig;
+        std::string cpp_input_with_newline;
+        if (!code_orig.empty() && code_orig.back() != '\n') {
```
certik, line 250: **[suggestion]** Design note before “move into `run()`”: centralizing an unconditional trailing-newline append at the entry of `CPreprocessor::run()` looks tempting, but it is **incorrect as a blanket change**. `run()` is also invoked recursively for macro argument expansion and recursive macro expansion (`preprocessor.re` ~731 and ~773) on short fragments such as `1` or `a+b`. Appending `\n` there would change expansion text and inject newlines mid-line.

The include path already normalizes **before** calling `run()` (`preprocessor.re` ~670–672) because that path is file-like; macro paths intentionally do not. This PR’s call-site layer (file-level inputs only) is therefore the right simple layer.

**Suggestion:** Keep the fix at file-level entry points (as here), or factor the include check + the two new call sites into one tiny helper used only for file-like buffers. Do **not** unconditionally append at the start of `run()`. Optionally note the include precedent in the PR description so reviewers see this rediscovers an existing pattern.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -245,7 +245,14 @@ Result<LFortran::AST::TranslationUnit_t*> FortranEvaluator::get_ast2(
...
+            cpp_input_with_newline = code_orig;
+            cpp_input_with_newline.push_back('\n');
```
certik, line 252: **[nit]** In `get_ast2`, the synthetic `\n` is appended to a temporary copy passed to `cpp.run`, while callers have already set `lm.file_ends` from the original `code_orig.size()`. Inside `run()`, `get_newlines` / final `in_start0` see size N+1 against `file_ends` of N. By contrast, `emit_c_preprocessor` appends **before** `lm.file_ends.push_back(input.size())`, so the two paths are slightly inconsistent.

Unlikely to break normal diagnostics (bisection clamps at `file_ends.back()`, and the synthetic newline sits on a non-re-emitted directive). Same pattern already exists for includes.

**Suggestion:** Either bump `lm.file_ends.back()` when appending if non-empty, or leave a one-line comment that the extra `\n` is preprocessor-only. Not a merge blocker.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -245,7 +245,14 @@ Result<LFortran::AST::TranslationUnit_t*> FortranEvaluator::get_ast2(
     if (compiler_options.c_preprocessor) {
         // Preprocessor
         LFortran::CPreprocessor cpp(compiler_options);
-        Result<std::string> res = cpp.run(code_orig, lm, cpp.macro_definitions, diagnostics);
+        const std::string *cpp_input = &code_orig;
```
certik, line 248: **[nit]** The `const std::string *` + optional `cpp_input_with_newline` dance avoids a copy when the source already ends in `\n`. Fine, but denser than needed for a rare path; when the C preprocessor is enabled, always copying is dominated by preprocessing itself.

**Suggestion:** Optional simplification:
```cpp
std::string cpp_input = code_orig;
if (!cpp_input.empty() && cpp_input.back() != '\n') {
    cpp_input.push_back('\n');
}
Result<std::string> res = cpp.run(cpp_input, lm, cpp.macro_definitions, diagnostics);
```
Pure style; not required for correctness.

certik (approved):
I think this is fine.

PR #12086, opened 2026-07-02
fix: propagate string length for captured assumed-length allocatable char arrays

HarshitaKalani:
https://github.com/lfortran/lfortran/issues/11870


certik:
I fixed up some issues (see the commit message) and merged main.

PR #12093, opened 2026-07-03
handle file open failures when iostat is provided

HarshitaKalani:

PR #12094, opened 2026-07-03
fix: hoist per-iteration allocas to entry block in LLVM codegen

HarshitaKalani:


integration_tests/CMakeLists.txt
```diff
@@ -3003,6 +3003,7 @@ RUN(NAME file_open_07 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME file_open_08 LABELS gfortran llvm)
 RUN(NAME file_open_09 LABELS gfortran llvm)
 RUN(NAME file_open_10 LABELS gfortran llvm)
+RUN(NAME file_open_11 LABELS gfortran llvm)
```
assem2002, line 3006: I think this test and the related changes to it are all addressed in this PR #12093 (registering this test, the test file, all the changes in file `lfortran_intrinsics.c`).
***
Let's revert them here and that PR will handle it.

integration_tests/write_internal_loop_01.f90
```diff
@@ -0,0 +1,17 @@
...
+    implicit none
+    integer :: i, ndigits, count
+    character(len=10) :: digits
+    character(len=32) :: buf
+
+    count = 0
+    do i = 1, 300000
+        ndigits = i
+        write(digits, '(I0)') ndigits
+        write(buf,    '(I0,":",I0)') i, ndigits
```
assem2002, line 11:
```suggestion
```
dump this write statement, it'll reduce the test time to almost the half.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -26759,8 +26756,18 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+#else
+            llvm::Function *string_format_stacksave_fn =
+                llvm::Intrinsic::getDeclaration(module.get(), llvm::Intrinsic::stacksave);
+            llvm::Value *string_format_saved_stack =
+                builder->CreateCall(string_format_stacksave_fn);
+#endif
+
+            // Create and Push a pointer to int64 to store the result size in.
+            llvm::Value *result_size_ptr = llvm_utils->CreateAlloca(
+                *builder, llvm::Type::getInt64Ty(context));
```
assem2002, line 26770: use `CreateAlloca`  (the one the `builder` NOT being passed, it hoists the instruction up)
```suggestion

            // Create and Push a pointer to int64 to store the result size in.
            llvm::Value *result_size_ptr = llvm_utils->CreateAlloca( llvm::Type::getInt64Ty(context));
```
There is no dynamic size here

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -26878,7 +26885,13 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
                     LCOMPILERS_ASSERT(llvm_utils->get_StringType(expr_type(x.m_args[i])) == llvm::Type::getInt8Ty(context))
                     LCOMPILERS_ASSERT(tmp->getType()->isPointerTy()); // atleast not `char`
-                    llvm::Value* tmp_ptr = builder->CreateAlloca(llvm::Type::getInt8Ty(context)->getPointerTo()->getPointerTo());
+                    // Dynamic alloca; reclaimed by llvm.stackrestore after
+                    // the variadic call (see stacksave above).  Hoisting
+                    // this into the entry block breaks the variadic call
+                    // ABI in the same way as result_size_ptr.
+                    llvm::Value* tmp_ptr = llvm_utils->CreateAlloca(
+                        *builder,
+                        llvm::Type::getInt8Ty(context)->getPointerTo()->getPointerTo());
```
assem2002, line 26894: same.
```suggestion
                    llvm::Value* tmp_ptr = llvm_utils->CreateAlloca(
                        llvm::Type::getInt8Ty(context)->getPointerTo()->getPointerTo());
```

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -26919,7 +26932,10 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
                 if(!tmp->getType()->isPointerTy() ||
                     ASR::is_a<ASR::PointerToCPtr_t>(*x.m_args[i])){
                     tmp = logical_store_val(tmp, x.m_args[i]);
-                    llvm::Value* tmp_ptr = builder->CreateAlloca(tmp->getType());
+                    // Dynamic alloca; reclaimed by llvm.stackrestore after
+                    // the variadic call (see stacksave above).
+                    llvm::Value* tmp_ptr = llvm_utils->CreateAlloca(
```
assem2002, line 26937: same

assem2002 (approved):
LGTM.
Please address my comments.

***
**If you're willing to do the following in here as it's quite relative, it'd be great refactor (otherwise we can open an issue for it).**

We have the following utilities :=
```cpp
            llvm::AllocaInst* CreateAlloca(llvm::Type* type,
                llvm::Value* size=nullptr, std::string Name="",
                bool is_llvm_ptr=false);
            llvm::AllocaInst* CreateAlloca(llvm::IRBuilder<> &builder,
                llvm::Type* type, llvm::Value* size=nullptr, std::string Name="",
                bool is_llvm_ptr=false);
```
- First one is accepting size (dynamic allocations) + it hoists the `Alloca` instruction.
- Second one is accepting size (dynamic allocations)  + it DOESN'T hoist the `Alloca` instr.

My suggestions
1- first function needs to be named into something showing that the function is hoisting, and maybe name the second function to show that it DOESN'T hoist :)
2- first function shouldn't be accepting size (dynamic size), because it hoists the allocation which may lead to llvm error `instruction not dominating` or a wrong logic.
3 - Second function doesn't need the `builder` parameter, the `LLVMUtils` class already has it.

PR #11966, opened 2026-06-23
fix: raise semantic error for pure function dummy args without INTENT(IN) or VALUE (C1592)

usefahmed07:
## Summary
Pure functions require all dummy data objects to have INTENT(IN) or VALUE attribute per Fortran standard C1592, but LFortran was silently ignoring this requirement.

## Fix
Added a check in `check_pure_function` in `ast_body_visitor.cpp` that raises a semantic error when a dummy argument of a pure function has `Unspecified` intent.

## Verification
- LFortran now raises: `semantic error: Dummy argument 'x' of pure function must have INTENT(IN) or VALUE attribute`
- Valid pure functions with INTENT(IN) still compile correctly
- Added integration test: `pure_intent_01`

fixes #11449

PR #12126, opened 2026-07-07
fix: raise semantic error for pure function dummy args without INTENT…

usefahmed07:
Re-implements PR #11966 (reverted in #12122) with a fix for the regression bonachea reported: C1592 only applies to *nonpointer* dummy data objects, so pointer dummy arguments (including procedure pointers such as 'procedure(...), pointer :: x') must be exempt.

The previous check only excluded arguments whose type was exactly FunctionType_t, which missed procedure *pointer* arguments (whose type is Pointer_t wrapping FunctionType_t). This unwraps pointer types before the FunctionType_t check, and also exempts any pointer dummy argument outright, matching the standard's wording.

Adds a regression integration test (pure_intent_02) that exercises a pure function with a procedure pointer dummy argument.


HarshitaKalani:
The CI fails:
```
coarrays_17.f90 * fortran
Reference JSON: tests/reference/fortran-coarrays_17-9932346.json
Output JSON:    tests/output/fortran-coarrays_17-9932346.json
Omitting 10 identical items
ACTUAL != REFERENCE
Differing items:
{'stdout_hash': 'ac4a70fa743d5cf9312ecb2029176e1f8862055cacd07e2d6035371f'} != {'stdout_hash': '84beafdf7622b1ad4dd981980726d55f9368095c062833b64f0697d6'}
=== DIFF ===
Reference: tests/reference/fortran-coarrays_17-9932346.stdout
Output:    tests/output/fortran-coarrays_17-9932346.stdout
10c10
< sync memory
---
> call __module_prif_prif_sync_memory()
34a35,42
> end interface
> 
> interface
>     subroutine __module_prif_prif_sync_memory(stat, errmsg, errmsg_alloc)
>         character(len=*, kind=1), intent(inout), optional :: errmsg
>         character(len=:, kind=1), allocatable, intent(inout), optional :: errmsg_alloc
>         integer(4), intent(out), optional :: stat
>     end subroutine __module_prif_prif_sync_memory

```

usefahmed07:
ok

integration_tests/pure_intent_02.f90
```diff
@@ -0,0 +1,26 @@
+! Regression test: C1592 only applies to *nonpointer* dummy data objects.
```
assem2002, line 1: Awesome you brought the test into integration test suite.
let's keep it that way, whenever we encounter a test in third party codes that's not covered in integration test suite, we bring it in

usefahmed07: Thanks! That makes sense. I'll follow this approach in future PRs and make sure to bring such third-party test cases into the integration test suite whenever applicable.

PR #12131, opened 2026-07-07
Fix  ICE for transfer with character array mold from assumed-length dummy

### LINKED ISSUE #12112: ICE for transfer with character array mold from assumed-length dummy
## MRE

```fortran
program main
  call f("a")

contains

  subroutine f(mold)
    character(*), intent(in) :: mold

    print *, size(transfer(0.0, [mold]))
  end subroutine f

end program main
```

## output

```text

Internal Compiler Error:
AssertFailed: llvm_utils->is_proper_array_of_strings_llvm_var(expr_type(expr), str)
```

## gfortran behavior

```text

           4
```

srinjoy933:
Fixes #12112


srinjoy933:
hey @opixdown please have a look at this pr ,Thank you!

integration_tests/array_130.f90
```diff
@@ -0,0 +1,10 @@
+program main
```
opixdown: change this to array_130

integration_tests/array_130.f90
```diff
@@ -0,0 +1,10 @@
...
+  call f("a")
+contains
+
+  subroutine f(mold)
+    character(*), intent(in) :: mold
+
+    print *, size(transfer(0.0, [mold]))
```
opixdown: also add check like if(/) error stop

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -15989,7 +15989,8 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
         bool is_fixed_size_array_result = ASRUtils::is_array(x.m_type) &&
             ASRUtils::extract_physical_type(x.m_type) == ASR::array_physical_typeType::FixedSizeArray;
         if ( !ASRUtils::types_equal(ASRUtils::extract_type(ASRUtils::expr_type(x.m_source)), ASRUtils::extract_type(x.m_type),
-             x.m_source, const_cast<ASR::expr_t*>(&x.base), false) && !ASRUtils::is_string_only(expr_type(x.m_mold)) &&
+             x.m_source, const_cast<ASR::expr_t*>(&x.base), false) && 
+             !ASRUtils::is_character(*ASRUtils::extract_type(expr_type(x.m_mold))) &&
```
opixdown, line 16111: can you please explain me the issue.

also with this pr the output is wrong
```
(lf) opixdown@Unknown_c6:68:ac:e5:68:0c lfortran % lfortran t.f90          
64
(lf) opixdown@Unknown_c6:68:ac:e5:68:0c lfortran % gfortran t.f90&& ./a.out 
           4
```

opixdown:
@srinjoy933 leave some comments, thank you

src/libasr/codegen/asr_to_llvm.cpp
srinjoy933: The ICE was happening because in visit_BitCast, the compiler was using !ASRUtils::is_string_only(expr_type(x.m_mold)) to protect the properly constructed str_view from being overwritten. However, because [mold] is an array of strings, is_string_only returned false. This caused the compiler to fall into the else block, overwrite tmp with an invalid pointer load (CreateLoad2), and pass a raw struct value to getelementptr, which crashed the LLVM backend. Changing the check to !ASRUtils::is_character(*ASRUtils::extract_type(...)) correctly identifies the string array's base type and safely preserves the str_view.


While this PR successfully fixes the ICE during LLVM lowering, the logical output of 64 points to a separate bug in how the frontend ASR calculates the dimension for transfer. Because mold is an assumed-length dummy argument, its element size is not known until runtime. Instead of generating a dynamically sized runtime array descriptor, the ASR seems to be assigning a hardcoded/fallback FixedSizeArray dimension of 64 at compile-time.

srinjoy933:
@opixdown  please have a look at this pr, Thanks!

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -14867,51 +14867,93 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                            al, x.base.base.loc, mold, local_int_type, nullptr));
+                        
+                        if (mold_str_type->m_kind > 1) {
+                            ASR::expr_t* kind_expr = ASRUtils::EXPR(ASR::make_IntegerConstant_t(al, x.base.base.loc, mold_str_type->m_kind, local_int_type));
+                            mold_bytes_expr = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(al, x.base.base.loc, mold_bytes_expr, ASR::binopType::Mul, kind_expr, local_int_type, nullptr));
+                        }
+                    } else if (mold_str_type->m_len && ASRUtils::expr_value(mold_str_type->m_len)) {
+                        mold_bytes = mold_str_type->m_kind * ASR::down_cast<ASR::IntegerConstant_t>(ASRUtils::expr_value(mold_str_type->m_len))->m_n;
+                    } else {
+                        mold_bytes = mold_str_type->m_kind * 1;
```
opixdown, line 14893: what about character(n) where n is a runtime expression? The current fallback appears to assume a length of 1, which would produce incorrect results for cases like:

```
character(n) :: mold
size(transfer(0.0, [mold]))
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -14867,51 +14867,93 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                        ASR::expr_t* mold_bytes_minus_one = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(
+                            al, x.base.base.loc, mold_bytes_expr, ASR::binopType::Sub, one_expr, local_int_type, nullptr));
                         ASR::expr_t* numerator = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(
-                            al, x.base.base.loc, src_len_expr, ASR::binopType::Add,
-                            mold_bytes_minus_one, int_type, nullptr));
+                            al, x.base.base.loc, src_len_expr, ASR::binopType::Add, mold_bytes_minus_one, local_int_type, nullptr));
                         result_size_expr = ASRUtils::EXPR(ASR::make_IntegerBinOp_t(
-                            al, x.base.base.loc, numerator, ASR::binopType::Div,
-                            mold_bytes_expr, int_type, nullptr));
+                            al, x.base.base.loc, numerator, ASR::binopType::Div, mold_bytes_expr, local_int_type, nullptr));
```
opixdown, line 14914: Should we guard against mold_bytes_expr == 0 before generating this division? A call like call f("") would result in len(mold) == 0, so this becomes a runtime divide by zero.

integration_tests/array_130.f90
```diff
@@ -0,0 +1,10 @@
+program array_130
+  call f("a")
+contains
+
```
opixdown, line 4: It would be good to add more test coverage for different runtime string lengths and character(n) cases.

opixdown:
left some comments otherwise it looks good to me, thank you

HarshitaKalani:
Please mark this PR as ready for review again once the comments are addressed. Thank you!

integration_tests/array_130.f90
srinjoy933: Added new test cases to integration_tests/array_130.f90 that explicitly test both the character(n) runtime length and the empty string scenario.

src/lfortran/semantics/ast_common_visitor.h
srinjoy933: I wrapped the final division node in an ASR::IfExp_t conditional to safely return 0 when mold_bytes_expr == 0 at runtime, preventing the crash for cases like call f("").Updated the fallback logic to properly route the mold_str_type->m_len runtime expression to mold_bytes_expr instead of assuming a length of 1.

srinjoy933:
> left some comments otherwise it looks good to me, thank you
Hi @opixdown , made some changes ,according to what you have suggested ,please have a look at this now ,Thank You!

PR #12134, opened 2026-07-08
fix: ICE for transfer with merge inside character function

### LINKED ISSUE #12069: ICE for transfer with merge inside character function
## MRE

```fortran
program transfer_mre
  character(len=20) :: out

  out = make_string("abc")
  print *, out

contains

  character(len=20) function make_string(text)
    character(len=*) :: text

    make_string = transfer(merge(transfer(text, "x", len(text)), &
        ["a", "b", "c"], .true.), repeat("x", len(text)))
  end function make_string

end program transfer_mre
```

## output

```text

Internal Compiler Error:
AssertFailed: is_proper_string_llvm_variable(str_type, str)
```

## gfortran behavior

```text

 abc
```

opixdown:
fixes https://github.com/lfortran/lfortran/issues/12069

### DIFF
--- a/src/libasr/intrinsic_func_registry_util_gen.py
+++ b/src/libasr/intrinsic_func_registry_util_gen.py
@@ -811,7 +811,8 @@ intrinsic_funcs_args = {
         {
             "args": [("any", "any", "bool")],
             "ret_type_arg_idx": 0,
-            "same_type_arg": 2
+            "same_type_arg": 2,
+            "return_array_arg_from_scalar_mask": True
         }
     ],
     "Mergebits": [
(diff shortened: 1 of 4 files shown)


src/libasr/intrinsic_func_registry_util_gen.py
```diff
@@ -811,7 +811,8 @@
         {
             "args": [("any", "any", "bool")],
             "ret_type_arg_idx": 0,
-            "same_type_arg": 2
+            "same_type_arg": 2,
+            "return_array_arg_from_scalar_mask": True
```
HarshitaKalani, line 815: I was thinking if this flag would ever be used for any other intrinsic. I think this behavior is only needed for MERGE, so should we keep it intrinsic-specific rather than forcing a generic flag. 
We removed "merge" specific code but this flag also does not seem to be generic. It is very merge-specific.

PR #12141, opened 2026-07-08
fix: preserve character kind in repeat

### LINKED ISSUE #12139: Bug : `repeat` with kind=4
```.f90
program char4_string_intrinsics_01
  implicit none
  character(kind=4, len=1) :: s
  s = "x"
  if (repeat(s, 3) /= 4_"xxx") error stop 2
end program
```
the check fails

jassem-manita:
fix #12139
Fixed repeat() for character(kind=4) by preserving the input string kind, and extended the existing kind-4 string intrinsics integration test.

### DIFF
--- a/integration_tests/char4_string_intrinsics_01.f90
+++ b/integration_tests/char4_string_intrinsics_01.f90
@@ -1,12 +1,15 @@
 program char4_string_intrinsics_01
   implicit none
   character(kind=4, len=5) :: s = 4_"ab"
+  character(kind=4, len=1) :: c
 
   ! len_trim with character(kind=4) previously ICE'd
   if (len_trim(s) /= 2) error stop 1
 
   ! repeat with character(kind=4) previously ICE'd
   if (repeat(4_"x", 3) /= 4_"xxx") error stop 2
+  c = "x"
+  if (repeat(c, 3) /= 4_"xxx") error stop 3
 
   print *, "ok"
 end program
--- a/integration_tests/char4_string_intrinsics_01.f90
+++ b/integration_tests/char4_string_intrinsics_01.f90
@@ -1,7 +1,8 @@
 program char4_string_intrinsics_01
   implicit none
   character(kind=4, len=5) :: s = 4_"ab"
-  character(kind=4, len=1) :: c
+  character(kind=4, len=1) :: c, glyph
+  character(kind=4, len=2) :: glyph_pair
 
   ! len_trim with character(kind=4) previously ICE'd
   if (len_trim(s) /= 2) error stop 1
@@ -10,6 +11,10 @@ program char4_string_intrinsics_01
   if (repeat(4_"x", 3) /= 4_"xxx") error stop 2
   c = "x"
   if (repeat(c, 3) /= 4_"xxx") error stop 3
+  glyph = achar(9731, kind=4)
+  glyph_pair(1:1) = achar(9731, kind=4)
+  glyph_pair(2:2) = achar(9731, kind=4)
+  if (repeat(glyph, 2) /= glyph_pair) error stop 4
 
   print *, "ok"
 end program
--- a/src/libasr/pass/intrinsic_functions.h
+++ b/src/libasr/pass/intrinsic_functions.h
@@ -6467,6 +6467,20 @@ namespace Rrspacing {
 
 namespace Repeat {
 
+    static inline ASR::ttype_t* character_type_with_kind(Allocator &al,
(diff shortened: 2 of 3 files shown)


jassem-manita:
@assem2002

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6467,6 +6467,20 @@ namespace Rrspacing {
...
+            len_kind, ASR::string_physical_typeType::DescriptorString));
+    }
+
+    static inline ASR::ttype_t* allocatable_deferred_string_with_kind(
+            Allocator &al, const Location &loc, int kind) {
+        return ASRUtils::TYPE(ASR::make_Allocatable_t(al, loc,
+            character_type_with_kind(al, loc, kind, nullptr,
+                ASR::string_length_kindType::DeferredLength)));
+    }
+
```
assem2002: Good helpers, but we have very similar ones at `asr_builder.h`.
You can find function`String()` there to use -- it will require a modification to be able to pass character_kind argument (let's set the parameter to default 1)
***
i.e.
```cpp
  ASR::ttype_t* String(...,...,..,character_kind = 1){
}
```

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6496,7 +6515,8 @@ namespace Repeat {
             ASR::string_length_kindType::AssumedLength,
             ASR::string_physical_typeType::DescriptorString)));
         fill_func_arg("y", arg_types[1]);
-        auto result = declare(fn_name, b.allocatable(b.String(nullptr, ASR::DeferredLength)), ReturnVar);
+        auto result = declare(fn_name,
```
assem2002, line 6505:
```suggestion
        auto result = declare(fn_name, b.allocatable(b.String(nullptr, char_kind ASR::DeferredLength)), ReturnVar)
```

src/libasr/pass/intrinsic_functions.h
```diff
@@ -6573,7 +6593,8 @@ namespace Repeat {
             if (diag.has_error()) return nullptr;
             return_type = expr_type(m_value);
         } else {
-            return_type = allocatable_deferred_string();
+            return_type = allocatable_deferred_string_with_kind(al, loc,
```
assem2002: Try to leverage the already existing utility in `asr_builder.h`. I think the worst case is that you'll create a wrapper for one of these utility/macro.
AI can help with that greatly.

integration_tests/char4_string_intrinsics_01.f90
```diff
@@ -1,12 +1,15 @@
...
   character(kind=4, len=5) :: s = 4_"ab"
+  character(kind=4, len=1) :: c
 
   ! len_trim with character(kind=4) previously ICE'd
   if (len_trim(s) /= 2) error stop 1
 
   ! repeat with character(kind=4) previously ICE'd
   if (repeat(4_"x", 3) /= 4_"xxx") error stop 2
+  c = "x"
+  if (repeat(c, 3) /= 4_"xxx") error stop 3
```
assem2002, line 13: Good test, but I can see when we actually use a kind=4 character, repeat still fail.
```.f90
program char4_string_intrinsics_01
  implicit none
  character(kind=4, len=1) :: s
  character(kind=4, len=2) :: m
  s = achar(9731, kind=4)
  m(1:1) = achar(9731, kind=4) 
  m(2:2) = achar(9731, kind=4) 
  if (repeat(s, 2) /= m) error stop 2
end program
```
***
## How to fix?
- implementation of `repeat` creates `ASR::stringSection` with wrong kind in return type member -- please correct that.
- Also check the `intrinsic_function` pass dump by eyes to see if we missed any spot of  `repeat` implementation where don't properly set character's kind. 
***
If you wish try to verify in `asr_verify.cpp` that stringSection should have it's argument kind matching the return kind. Note other tests start to fail due to that, just skip the idea and open an issue

assem2002 (commented):
It it's in the good direction. Thanks
Please address my comments. Set as "ready to review" when ready

jassem-manita:
> It it's in the good direction. Thanks Please address my comments. Set as "ready to review" when ready

done. i made `ASRBuilder::String()` accept an optional `character_kind` parameter with default `1`, so `repeat` can reuse the existing builder path directly through `b.String(...)` / `b.allocatable(...)`.

Since the existing utility now supports the needed kind no didn’t need to add a wrapper

assem2002 (approved):
Looks good. THANKS

PR #12144, opened 2026-07-08
fix: use immediate owner (not nearest Module) for ExternalSymbol::m_m…

### LINKED ISSUE #12138: ASR verify failure for nested character function with length depending on parameter array
## MRE

```fortran
module m
contains

  character(4) function g(k)
    integer :: k

    g = f(k)

  contains

    function f(n)
      character(3), parameter :: names(1) = ["abc"]
      integer :: n
      character(len_trim(names(n))) :: f

      f = "abc"
    end function f

  end function g

end module m

program p
  use m
  print *, g(1)
end program p
```

## output

```text

ASR verify pass error: ExternalSymbol::m_module_name `m` must match external's module name `f`

Internal Compiler Error:
```

## gfortran behavior

```text

 abc
```

usefahmed07:
Fix #12138
## Problem

Nested functions with a dynamic-length `character` in their signature (e.g. `character(len=len_trim(x))`) could crash the compiler at LLVM codegen with: Instruction does not dominate all uses!

## Root Cause

`ArrayConstantVisitor` (in `implied_do_loops.cpp`) lowers array constants into a temp variable + `Allocate` statement. It used the default traversal, which also visited a function's `m_function_signature` — type metadata, not executable code. When an array-constant-like expression appeared there, the visitor created a temp variable and tried to allocate it, but the signature has no function body to place the `Allocate` in. This left the variable used in two places with no statement dominating both, which LLVM rejects.

## Fix

Added a `visit_Function` override that only visits the function's scope, args, body, and return variable — skipping `m_function_signature` entirely, since it's handled elsewhere.

## Before : 

```fortran
subroutine collect_used_modules(target)
    type(build_target), intent(inout) :: target
    allocate(character(len=len_trim(target%name)) :: target%names(1))
end subroutine
```
$ lfortran nested_23.f90
Instruction does not dominate all uses!
LLVM ERROR: Broken function found, compilation aborted!

## After : 
$ lfortran nested_23.f90
MRE completed successfully

## Testing

- Covered by the existing `nested_23` integration test (llvm backend).
- Full `run_tests.py -b llvm` suite passes, no new regressions.

### DIFF
--- a/integration_tests/nested_28.f90
+++ b/integration_tests/nested_28.f90
@@ -1,0 +1,26 @@
+module nested_28_mod
+contains
+
+  character(4) function g(k)
+    integer :: k
+
+    g = f(k)
+
+  contains
+
+    function f(n)
+      character(3), parameter :: names(1) = ["abc"]
+      integer :: n
+      character(len_trim(names(n))) :: f
+
+      f = "abc"
+    end function f
+
+  end function g
+
+end module nested_28_mod
+
+program nested_28
+  use nested_28_mod
+  print *, g(1)
+end program nested_28
--- a/integration_tests/nested_28.f90
+++ b/integration_tests/nested_28.f90
@@ -23,4 +23,5 @@ end module nested_28_mod
 program nested_28
   use nested_28_mod
   print *, g(1)
+  if (len_trim(g(1)) /= 3) error stop
 end program nested_28
--- a/src/libasr/pass/implied_do_loops.cpp
+++ b/src/libasr/pass/implied_do_loops.cpp
@@ -937,43 +937,10 @@ class ArrayConstantVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayC
             // Do nothing, already handled in init_expr pass
         }
 
-        void visit_Function(const ASR::Function_t& x) {
-            // Skip visiting m_function_signature here. Array constants that
-            // appear inside a function's signature (e.g. inside a
-            // dynamic-length String's m_len expression, such as
-            // `character(len_trim(param_array(n)))`) are type-level
-            // metadata, not executable statements. Applying the
-            // ArrayConstant -> temp-variable transform to them creates a
-            // variable in whichever current_scope happens to be active at
-            // visit time, with no corresponding Allocate statement placed
-            // in the right function body, which later causes an LLVM
-            // "instruction does not dominate all uses" error when the
-            // signature is inspected both from the function itself and
-            // from a caller's FunctionCall type.
-            ASR::Function_t& xx = const_cast<ASR::Function_t&>(x);
-            SymbolTable* current_scope_copy = current_scope;
-            current_scope = x.m_symtab;
-            for (auto &a : x.m_symtab->get_scope()) {
-                this->visit_symbol(*a.second);
-            }
-            for (size_t i = 0; i < x.n_args; i++) {
-                ASR::expr_t** current_expr_copy_0 = current_expr;
-                current_expr = const_cast<ASR::expr_t**>(&(x.m_args[i]));
-                this->call_replacer();
-                current_expr = current_expr_copy_0;
-                if (x.m_args[i] && visit_expr_after_replacement)
-                    this->visit_expr(*x.m_args[i]);
-            }
-            this->transform_stmts(xx.m_body, xx.n_body);
-            if (x.m_return_var) {
-                ASR::expr_t** current_expr_copy_1 = current_expr;
-                current_expr = const_cast<ASR::expr_t**>(&(x.m_return_var));
-                this->call_replacer();
-                current_expr = current_expr_copy_1;
-                if (x.m_return_var && visit_expr_after_replacement)
-                    this->visit_expr(*x.m_return_var);
-            }
-            current_scope = current_scope_copy;
+        void visit_ttype(const ASR::ttype_t& /*x*/) {
+            // Do nothing. Array constants that appear inside type metadata
+            // (e.g. dynamic-length strings) are not executable statements.
+            // Replacing them here creates dangling local variables.
         }
 
         void call_replacer() {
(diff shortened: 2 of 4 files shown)


assem2002 (changes requested):
I don't think this is the root cause.
***
## Why?
Here `ExternalSymbol` is setting `m_module_name` to function's name, which is incorrect, you shouldn't be accessing a local variable from outside of a function.

`ASR::ExternalSymbols` doesn't try to bypass functions and module to get some symbol, it's mainly used to point to symbols that's not in the current ASR scope, but it also implies that current scope can access that variable anytime at runtime.
To be more accurate, it was made as a way for ASR to be able to point to symbols that's not reachable in the current ASR scope to be able to load/unload modules and/or functions easily (serialization/deserialization).
***
## What is the root cause?
Look at this MRE 
```.f90
module m
  contains 
  function f(n) result (r)
    character(3),parameter :: s(1) = ["aa"]
    integer :: n  
    character (len_trim(s(n))) :: r
  end function f
end module m  

program p
  use m
  character(4) :: s
  s = f(1)
if(len(s) /= 2) error stop
end program p
```
some how pass `implied_do_loop` misses with the `FunctionType` node of function`f` 
from this := 
```clj
(FunctionType
                                        [(Integer 4)]
                                        (String 1 (IntrinsicElementalFunction
                                            StringLenTrim
                                            [(ArrayItem
                                                (ArrayConstant
                                                    3
                                                    ["aa "]
                                                ......
``` 
To this :=
```clj
(FunctionType
                                        [(Integer 4)]
                                        (String 1 (IntrinsicElementalFunction
                                            StringLenTrim
                                            [(ArrayItem
                                                (Var 3 __libasr__created__var__0__array_constant_)
                                                [(()
............
```

You can even use flag `--skip-pass=implied_do_loops` and the MRE will work.
***
## What to do?
Find out why that variable was created, Is it there for good reason? why it's usually made and used for? How to avoid or leverage it with this MRE

usefahmed07:
okey thanks for the heads-up

Jatinagarwal24 (commented):
Hi @usefahmed07 I can see you haven't added the integration test for this pr.Please add and register that to prevent future regressions.

usefahmed07:
ok i will add the integration test

Jatinagarwal24 (commented):
Thanks @assem2002 for correctly pointing out the fix.
@usefahmed07 I think the correct fix is simple .Your cuurent attempt in this PR to skip `m_function_signature` inside `visit_Function` was on the right track, but incomplete (because `m_type` still gets visited in other nodes like `FunctionCall` or Variable).

So instead of hacking `subroutine_from_function.cpp`, we should revert those changes completely. To fix the root cause, we just need to instruct the `ArrayConstantVisitor` in `implied_do_loops.cpp` to never modify type metadata by overriding `visit_ttype`:
```
        void visit_ttype(const ASR::ttype_t& /*x*/) {
            // Do nothing. Array constants that appear inside type metadata
            // (e.g. dynamic-length strings) are not executable statements. 
            // Replacing them here creates dangling local variables.
        }
```

usefahmed07:
@Jatinagarwal24  Implemented the `visit_ttype` override as
suggested — reverted the `m_function_signature`-specific skip in
`visit_Function` entirely, replaced with a no-op `visit_ttype`
override in `ArrayConstantVisitor`.

Verified:
- Original issue #12138 MRE: compiles and runs, prints `abc ` ✅
- @assem2002's MRE: matches gfortran's own output exactly (`[  ]`,
  `len(s) = 4`) ✅
- `nested_28` (new regression test) passes ✅
- Full `nested_*` and `template_*` test suites pass with no
  regressions ✅

Thanks both for pushing on the root cause - this is a much cleaner
fix.

src/libasr/pass/implied_do_loops.cpp
```diff
@@ -937,6 +937,12 @@ class ArrayConstantVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayC
             // Do nothing, already handled in init_expr pass
         }
 
+        void visit_ttype(const ASR::ttype_t& /*x*/) {
```
assem2002, line 940: I don't know if we should use be more specific and use `visit_FunctionType` here.
but in general, Creating temporaries for arrayConstants isn't this pass job from the first place so I think the fix is good.
I created #12198 to mention that we need a refactor for that pass.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -175,26 +175,41 @@ private :
      *  to get it due to name mangling -- So just create a one.
      */
     ASR::symbol_t* get_resolved_symbol(ASR::symbol_t* sym){
-        ASR::Module_t *sym_mod = ASRUtils::get_sym_module(sym);
+        // Unwrap first -- ExternalSymbol::m_external must never itself be
```
assem2002: I don't think changes here are needed anymore.
@usefahmed07  Do you think we should keep it?

integration_tests/nested_28.f90
```diff
@@ -0,0 +1,26 @@
...
+      f = "abc"
+    end function f
+
+  end function g
+
+end module nested_28_mod
+
+program nested_28
+  use nested_28_mod
+  print *, g(1)
```
assem2002, line 25:
```suggestion
  print *, g(1)
  if(len(g(1)) /= 3) error stop
```

assem2002 (approved):
Thanks for finding the right approach.
Please address the comments below.

src/libasr/pass/subroutine_from_function.cpp
Jatinagarwal24: Yes i checked locally also no changes needed here in subroutines_.cpp file.
Thanks.

assem2002: Great! let's revert those changes.

Jatinagarwal24 (approved):
I think now looks good to me.
Thanks @usefahmed07 !

integration_tests/nested_28.f90
usefahmed07: Applied it with len_trim(g(1)) instead of len(g(1)), since g has a fixed declared length of 4, so len(g(1)) would always return 4 and never actually verify the trimmed content length of 3.

assem2002 (approved):

PR #12150, opened 2026-07-09
Fix select type with PDT kind parameters

### LINKED ISSUE #8825: Cannot compile a select type program with parametrized derived types with kind parameters
The following program:
```
module test
  implicit none

  type, abstract :: parent
  end type parent

  ! Concrete real type with default kind
  type, extends(parent) :: child_real(prec)
    integer, kind :: prec = kind(1.0d0)
    real(prec) :: a
  end type child_real

  ! Concrete int type with default kind
  type, extends(parent) :: child_int(prec)
    integer, kind :: prec = kind(1)
    integer(prec) :: a
  end type child_int

contains

  subroutine try_select(par)
    class(parent), intent(in) :: par

    ! Associate name 'p' lets us access fields of the specific type
    select type (p => par)
    type is (child_real(8))
      write(*,*) "Do something with the real (kind=8):", p%a
    type is (child_int(4))
      write(*,*) "Do something with the int (kind=4):", p%a
    class default
      write(*,*) "Unknown or other precision"
    end select

  end subroutine try_select

end module test



program main
  use test
  implicit none

  type(child_real(8)) :: r
  type(child_int(4))  :: i

  r%a = 3.141592653589793_8
  i%a = 42

  call try_select(r)
  call try_select(i)
end program main

```
cannot be compiled by lfortran 0.57 with the error:
```
syntax error: Token '(' is unexpected here
  --> a.f90:26:24
   |
26 |     type is (child_real(8))
   |                        ^


``` 

Compiling with `lfortran --std=f23 a.f90` 

This code compiles with flang, ifx, and ifort but fails to compile with gfortran and nvfortran.

certik:
LFortran rejected `type is (child(4))` as a syntax error. Accept parametrized derived types with kind arguments in type-is guards by extending the parser, and route those guards through the existing TypeStmtName/ClassToStruct path so associate names can access members.

Adds integration_tests/select_type_53.f90 (llvm only; gfortran lacks support for this construct). Regenerates ci/parser.yy.patch.

Fixes #8825.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2065,6 +2065,8 @@ RUN(NAME select_type_49 LABELS gfortran llvm)
 RUN(NAME select_type_50 LABELS gfortran llvm)
 RUN(NAME select_type_51 LABELS gfortran llvm)
 RUN(NAME select_type_52 LABELS gfortran llvm)
+# type is (pdt(kind)) — not supported by gfortran
+RUN(NAME select_type_53 LABELS llvm)
 
 RUN(NAME where_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
 RUN(NAME where_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc fortran)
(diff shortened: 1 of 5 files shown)


integration_tests/CMakeLists.txt
```diff
@@ -2065,6 +2065,8 @@ RUN(NAME select_type_49 LABELS gfortran llvm)
 RUN(NAME select_type_50 LABELS gfortran llvm)
 RUN(NAME select_type_51 LABELS gfortran llvm)
 RUN(NAME select_type_52 LABELS gfortran llvm)
+# type is (pdt(kind)) — not supported by gfortran
```
```suggestion
```

certik:
Tests passed, I just removed a redundant comment and force merged and canceled the CI.

PR #12152, opened 2026-07-09
fix: Illegal inquire statement allowed  

### LINKED ISSUE #12147: Illegal inquire statement allowed
Lfortran 0.63.0 ran this program but should not have compiled it because the INQUIRE statement must evaluate the ```delim=``` variable, not treat it as if it were INTENT(IN), and 'WHAT' is not one of its possible values.
```
program badinquire
  implicit none
  character(20):: del = 'WHAT'
  inquire(6,delim=trim(del))
  print *,'DELIM=',trim(del)
end program badinquire
```
Lfortran output:
```
 DELIM=WHAT
``` 
Ifx gives a clear error message:
```
badinquire.f90(4): error #7939: A scalar default-character variable is required in this context.   [TRIM]
  inquire(6,delim=trim(del))
```

opixdown:
fixes #12147

### DIFF
--- a/src/lfortran/semantics/ast_body_visitor.cpp
+++ b/src/lfortran/semantics/ast_body_visitor.cpp
@@ -2796,6 +2796,18 @@ public:
         ASR::expr_t *iolength = args[32], *sign=args[33], *encoding=args[34];
         ASR::expr_t *stream = args[35], *iomsg = args[36], *round = args[37];
         ASR::expr_t *pending = args[38], *asynchronous = args[39];
+        if (delim) {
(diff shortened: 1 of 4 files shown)


HarshitaKalani:
You'll need to resolve the conflicts

harperjf:
I should have said in #12147 that the only specifiers that INQUIRE may treat as if they were INTENT(IN) are ```unit=``` and ```file=```. All the others must be treated like INTENT(OUT) but lfortran treated at least one like INTENT(INOUT). Of course only a subprogram's argument may have an INTENT, and INQUIRE is a statement.

Issue #11191, opened 2026-04-20
Extract MRE for a branch of caffeine


certik:
Extract and fix MRE for this bug: https://github.com/bonachea/caffeine/pull/2#issuecomment-4282972334

bonachea:
You can ask Julienne to run only the failing test with a command like the following (assuming a build using `install.sh`):

```
$ env GASNET_BACKTRACE=1 CAF_IMAGES=2 ./run-fpm.sh test -- --contains broadcast

[100%] Project compiled successfully.

Append '-- --help' or '-- -h' to your `fpm test` command to display usage information.

Running only tests with subjects or descriptions containing 'broadcast'.

prif_init
 0 of 0 tests passed. 0 tests were skipped.

prif_num_images
 0 of 0 tests passed. 0 tests were skipped.

prif_this_image_no_coarray
 0 of 0 tests passed. 0 tests were skipped.

PRIF Image Queries
 0 of 0 tests passed. 0 tests were skipped.

PRIF Types
 0 of 0 tests passed. 0 tests were skipped.

prif_co_broadcast
double free or corruption (out)
*** Caught a fatal signal (proc 0): SIGABRT(6)
[0] Invoking GDB for backtrace...
[...]
[0] #8  <signal handler called>
[0] #9  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
[0] #10 __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
[0] #11 __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
[0] #12 0x00007bd81cc4527e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
[0] #13 0x00007bd81cc288ff in __GI_abort () at ./stdlib/abort.c:79
[0] #14 0x00007bd81cc297b6 in __libc_message_impl (fmt=fmt@entry=0x7bd81cdce8d7 "%s\n") at ../sysdeps/posix/libc_fatal.c:134
[0] #15 0x00007bd81cca8ff5 in malloc_printerr (str=str@entry=0x7bd81cdd1ac0 "double free or corruption (out)") at ./malloc/malloc.c:5775
[...]

PR #12161, opened 2026-07-09
fix: inline character layout for SEQUENCE types

certik:
LFortran stored fixed-length character components of derived types as string descriptors ({i8*, i64}). For SEQUENCE types this violated the requirement that components form a contiguous storage sequence, so storage_size under-reported the size and c_loc + memcpy corrupted memory (double free). That broke Caffeine's prif_co_broadcast test (issue #11191).

- Add is_sequence to ASR Struct and set it from the SEQUENCE attribute
- Lay out non-pointer, non-allocatable character members of SEQUENCE (and bind(C)) types as inline [len x i8]
- Fix assignment (memcpy + blank pad), read, print, and finalization for that layout
- Fix storage_size / compute_type_size_align to include character length
- Warn on c_loc of non-SEQUENCE types with descriptor-backed characters
- Add integration_tests/derived_types_150.f90 covering SEQUENCE + storage_size + memcpy round-trip

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2504,6 +2504,7 @@ RUN(NAME derived_types_146 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME derived_types_147 LABELS gfortran llvm)
 RUN(NAME derived_types_148 LABELS gfortran llvm)
 RUN(NAME derived_types_149 LABELS gfortran llvm)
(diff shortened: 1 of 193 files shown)


certik:
Once this is merged, the follow-up PR is https://github.com/lfortran/lfortran/pull/12162.

bonachea:
@certik Thanks for working on this!

I lack the expertise to review most of the changes in this PR, but the test program looks right.

PR #12164, opened 2026-07-10
fix: splice "\<newline>" in the C preprocessor

HarshitaKalani:


src/lfortran/parser/preprocessor.re
```diff
@@ -335,7 +335,23 @@ int parse_bexpr(unsigned char *string_start, unsigned char *&cur, const cpp_symt
 Result<std::string> CPreprocessor::run(const std::string &input, LocationManager &lm,
         cpp_symtab &macro_definitions, diag::Diagnostics &diagnostics) const {
     LCOMPILERS_ASSERT(input[input.size()] == '\0');
-    unsigned char *string_start=(unsigned char*)(&input[0]);
+    std::string spliced_input = input;
+    for (size_t i = 0; i + 1 < spliced_input.size(); ) {
+        if (spliced_input[i] == '\\' && spliced_input[i+1] == '\n') {
+            spliced_input[i] = ' ';
+            spliced_input[i+1] = ' ';
```
certik: I think this implementation might have a bug, from AI:

"A backslash-newline is removed during C translation phase 2, not replaced with whitespace. Replacing both characters with spaces changes tokenization: for example, `#define JOIN joined\\\n_name 1` must define `joined_name`, while this produces two tokens and leaves `_name 1` in the replacement list; likewise, a split directive such as `#def\\\nine X 1` is no longer recognized. Please splice by deletion (while preserving source-location mapping separately if needed) and add a test where the continuation occurs inside a preprocessing token."

I think the AI might be right. Can you add some tests for `#define JOIN joined\\\n_name 1` and ensure they behave correctly, that the above code didn't break it?

HarshitaKalani:
This is ready for a review.

PR #12174, opened 2026-07-10
fix : multiple definition linker errors for compiler-generated helpers

### LINKED ISSUE #12173: Multiple definition linker errors for `_lcompiler_*` / `_lcompilers_*` symbols during separate compilation
MRE

```
program separate_compilation_50
    use separate_compilation_50a_mod
    use separate_compilation_50b_mod
    implicit none
    real(8) :: x
    integer :: r1, r2
    x = 3.14d0
    call get_exp_a(x, r1)
    call get_exp_b(x, r2)
    print *, r1, r2
    if (r1 /= r2) error stop
end program separate_compilation_50
```
```
module separate_compilation_50a_mod
    implicit none
contains
    subroutine get_exp_a(x, r)
        real(8), intent(in) :: x
        integer, intent(out) :: r
        r = exponent(x)
    end subroutine get_exp_a
end module separate_compilation_50a_mod

```

```
module separate_compilation_50b_mod
    implicit none
contains
    subroutine get_exp_b(x, r)
        real(8), intent(in) :: x
        integer, intent(out) :: r
        r = exponent(x)
    end subroutine get_exp_b
end module separate_compilation_50b_mod

```
LFortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran --separate-compilation integration_tests/separate_compilation_50a.f90 integration_tests/separate_compilation_50b.f90 integration_tests/separate_compilation_50.f90 -o /tmp/sc50_test
/usr/bin/ld: /tmp/separate_compilation_50b.tmp_1hGgjE8xXzhiq1nAbh7k5d.o: in function `_lcompiler_optimization_exponent_f64':
LFortran:(.text+0x0): multiple definition of `_lcompiler_optimization_exponent_f64'; /tmp/separate_compilation_50a.tmp_1hGgjE8xXzhiq1nAbh7k5d.o:LFortran:(.text+0x0): first defined here
/usr/bin/ld: /tmp/separate_compilation_50.tmp_1hGgjE8xXzhiq1nAbh7k5d.o: in function `_lcompiler_optimization_exponent_f64':
LFortran:(.text+0x0): multiple definition of `_lcompiler_optimization_exponent_f64'; /tmp/separate_compilation_50a.tmp_1hGgjE8xXzhiq1nAbh7k5d.o:LFortran:(.text+0x0): first defined here
The command 'clang -o /tmp/sc50_test /tmp/separate_compilation_50a.tmp_1hGgjE8xXzhiq1nAbh7k5d.o /tmp/separate_compilation_50b.tmp_1hGgjE8xXzhiq1nAbh7k5d.o /tmp/separate_compilation_50.tmp_1hGgjE8xXzhiq1nAbh7k5d.o  -L"/root/lfortran/src/bin/../runtime" -Wl,-rpath,"/root/lfortran/src/bin/../runtime" -llfortran_runtime -lm' failed.
Also, if required use --linker-path=<PATH>, where PATH has location to look for the linker executable
```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran#  gfortran integration_tests/separate_compilation_50a.f90 integration_tests/separate_compilation_50b.f90 integration_tests/separate_compilation_50.f90 -o /tmp/sc50_gfortran && /tmp/sc50_gfortran
           2           2
```

Jatinagarwal24:
fixes #12173

Compiler-generated intrinsic helper functions (e.g.
_lcompiler_optimization_exponent_f64, _lcompilers_acos_f32, etc.)
were emitted with external linkage in every compilation unit that
used them. When linking multiple object files, this caused
"multiple definition" linker errors.

Fix by setting LinkOnceODRLinkage on functions whose names start
with "_lcompiler_". This tells the linker that all definitions are
identical and it may keep only one copy.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4509,6 +4509,7 @@ RUN(NAME separate_compilation_46 LABELS gfortran llvm EXTRAFILES separate_compil
 RUN(NAME separate_compilation_47 LABELS gfortran llvm EXTRAFILES separate_compilation_47a.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_48 LABELS gfortran llvm EXTRAFILES separate_compilation_48a.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_49 LABELS gfortran llvm EXTRAFILES separate_compilation_49a.f90 EXTRA_ARGS --separate-compilation)
+RUN(NAME separate_compilation_50 LABELS gfortran llvm EXTRAFILES separate_compilation_50a.f90 separate_compilation_50b.f90 EXTRA_ARGS --separate-compilation)
 
 RUN(NAME no_explicit_return_type_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc
     EXTRA_ARGS --implicit-typing)
--- a/integration_tests/separate_compilation_50.f90
+++ b/integration_tests/separate_compilation_50.f90
@@ -1,0 +1,12 @@
+program separate_compilation_50
+    use separate_compilation_50a_mod
+    use separate_compilation_50b_mod
+    implicit none
+    real(8) :: x
+    integer :: r1, r2
+    x = 3.14d0
+    call get_exp_a(x, r1)
+    call get_exp_b(x, r2)
+    print *, r1, r2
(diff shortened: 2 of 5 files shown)


kmr-srbh (approved):
I think this is fine.

PR #12176, opened 2026-07-10
fix: handle ArrayConstant in LogicalToLogical cast

### LINKED ISSUE #12175: Code Generation Error in Logical Array Casting
MRE
```fortran
program logical_array_cast_01
    implicit none
    call activate()
    print *, "Ok"
contains
    subroutine activate()
        logical(1), dimension(2) :: iremove = [.false., .true.]
        if (iremove(1) .or. .not. iremove(2)) error stop
    end subroutine activate
end program logical_array_cast_01
```

LFortran
```console
[2 x i8]* @activate.iremove



```

HarshilShah1804:
Fixes #12175

### DIFF
--- a/src/lfortran/semantics/asr_implicit_cast_rules.h
+++ b/src/lfortran/semantics/asr_implicit_cast_rules.h
@@ -564,6 +564,26 @@ public:
                     ASR::LogicalConstant_t *l = ASR::down_cast<ASR::LogicalConstant_t>(value);
                     value = (ASR::expr_t *)ASR::make_LogicalConstant_t(al, a_loc,
                         l->m_value, dest_type2);
+                } else if (ASR::is_a<ASR::ArrayConstant_t>(*value)) {
+                    ASR::ArrayConstant_t* array = ASR::down_cast<ASR::ArrayConstant_t>(value);
+                    ASR::Array_t* array_type = ASR::down_cast<ASR::Array_t>(array->m_type);
+                    bool *data = (bool*) array->m_data;
+                    size_t array_size = ASRUtils::get_fixed_size_of_array(array->m_type);
+                    int dest_kind = ASRUtils::extract_kind_from_ttype_t(dest_type2);
+                    bool *new_array = al.allocate<bool>(array_size);
+                    for (size_t i = 0; i < array_size; i++) {
+                        new_array[i] = data[i];
+                    }
+                    ASR::ttype_t* new_array_type = ASRUtils::TYPE(ASR::make_Array_t(al, dest_type2->base.loc, dest_type2,
+                                                      array_type->m_dims, array_type->n_dims, ASR::array_physical_typeType::FixedSizeArray));
+                    // NOTE: Logical ArrayConstant_t data is stored packed one byte per element (`sizeof(bool) == 1`)
+                    // regardless of Fortran kind. However, asr_verify (asr_verify.cpp:1579) computes `fixed_size * kind`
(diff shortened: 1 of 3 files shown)


src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -564,6 +564,23 @@ class ImplicitCastRules {
                     ASR::LogicalConstant_t *l = ASR::down_cast<ASR::LogicalConstant_t>(value);
                     value = (ASR::expr_t *)ASR::make_LogicalConstant_t(al, a_loc,
                         l->m_value, dest_type2);
+                } else {
```
Jatinagarwal24: See if we change this else block to this 
```
} else if (ASR::is_a<ASR::ArrayConstant_t>(*value)) {
```

src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -564,6 +564,23 @@ class ImplicitCastRules {
...
+                    LCOMPILERS_ASSERT(ASR::is_a<ASR::ArrayConstant_t>(*value));
```
Jatinagarwal24: Redundant once the `else if` above does the type check — the branch can only be entered when this holds.

src/lfortran/semantics/asr_implicit_cast_rules.h
Jatinagarwal24: `LCOMPILERS_ASSERT` compiles out in release builds, so with a bare `else` the `down_cast` on the next line is unchecked. If `value` is ever some other constant expression here (e.g. an `ArrayConstructor_t` that didn't get folded), release builds would `down_cast` it anyway and then read `(bool*)array->m_data` off unrelated memory. Guarding with `else if (ASR::is_a<ASR::ArrayConstant_t>(*value))` makes release degrade to "no constant folding" instead of UB, and costs nothing in debug.

src/lfortran/semantics/asr_implicit_cast_rules.h
```diff
@@ -564,6 +564,23 @@ class ImplicitCastRules {
...
+                    ASR::ArrayConstant_t* array = ASR::down_cast<ASR::ArrayConstant_t>(value);
+                    ASR::Array_t* array_type = ASR::down_cast<ASR::Array_t>(array->m_type);
+                    bool *data = (bool*) array->m_data;
+                    size_t array_size = ASRUtils::get_fixed_size_of_array(array->m_type);
+                    int dest_kind = ASRUtils::extract_kind_from_ttype_t(dest_type2);
+                    bool *new_array = al.allocate<bool>(array_size);
+                    for (size_t i = 0; i < array_size; i++) {
+                        new_array[i] = data[i];
+                    }
+                    if (new_array) {
```
Jatinagarwal24: Always true: `al.allocate<bool>()` never returns null (the Allocator aborts on OOM rather than returning). This looks copied from the `ComplexToComplex` branch, where the analogous `if (new_data)` is meaningful because `new_data` stays `nullptr` when `dest_kind` is neither 4 nor 8. There's no such fallthrough here, so the check does nothing. Removing it lets lines 579-582 de-indent one level.

Jatinagarwal24 (commented):
Fix looks in good direction.
Just address the comments.
Thanks

Jatinagarwal24 (approved):
Looks good to me.
Thanks.

PR #12178, opened 2026-07-11
Fix/requirement duplicate parameter 11347

### LINKED ISSUE #11347: The compiler allows multiple instances of the same parameter when declaring a requirement
When declaring a parameter, multiple instances of the same should result in a warning.

code:

```
module Requirement_ordering_m
    implicit none

    requirement op_R ( T, T, V, V, op_func )
        type, deferred :: T
        type, deferred :: V
            type(T), intent(in) :: lhs
            type(T), intent(in) :: rhs
            type(V) :: res
        end function
    end requirement

    template op_t ( T, V, op_func )
      require :: op_R ( T, V, op_func )
    contains
      pure elemental function call_op_func ( x ) result(res)
        type(T) :: x
        type(V) :: res
        res = op_func(x,x)
      end function
    end template

end module

```

A warning appears if I would pass more or less than 2 parameters in the require statement of the template declaration.
Like this:

```
template op_t ( T, V, op_func )
      require :: op_R ( T, T, V, V, op_func )
```

The warning is that the amount of parameters is incorrect. However the warning should be that I have multiple instances of the same parameter when declaring the requirement.

usefahmed07:
Fixes #11347

## Problem
LLVM codegen crashed with `"instruction does not dominate all uses"`
for nested functions. The root cause was that `ArrayConstantVisitor`
was traversing `m_function_signature` inside nested functions,
incorrectly visiting external symbols and their owning modules —
using the *nearest* module instead of the symbol's *immediate* owner,
and failing to unwrap `ExternalSymbol` before treating it as the
external target.

## Fix
- Use the immediate owner (not the nearest module) when resolving
  `ExternalSymbol::m_module`.
- Unwrap `ExternalSymbol` before using it as `ExternalSymbol::m_external`.
- Skip `m_function_signature` in `ArrayConstantVisitor::visit_Function`
  so nested function signatures are no longer incorrectly traversed.

## Testing
- Verified the crash no longer reproduces on the original failing case.
- Ran the relevant test suite to confirm no regressions were introduced.


src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -4472,8 +4406,18 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
             std::string arg = to_lower(x.m_namelist[i]);
+            if (seen_args.find(arg) != seen_args.end()) {
+                diag.add(diag::Diagnostic(
+                    "Parameter '" + arg + "' is declared more than once in requirement '"
+                    + to_lower(x.m_name) + "'",
+                    diag::Level::Error, diag::Stage::Semantic, {
+                        diag::Label("", {x.base.base.loc})}));
+                throw SemanticAbort();
+            }
+            seen_args.insert(arg);
```
certik: Isn't this all that is needed? What is all the code below doing?

usefahmed07: @certik Yup, that small block is all that's needed for this specific
fix — it's a hard error the moment a duplicate name is found in the
parameter list, before any other processing happens.

The code below it (the "Parameter X is unused" warning loop and the
"Symbol not declared in parameters" error loop) is pre-existing logic
that was already there before this PR — it's unrelated to duplicates.
It checks the opposite kind of mismatch: that every declared parameter
is actually used in the requirement body, and that every symbol
defined in the body is listed as a parameter. I didn't touch that
logic, just added the duplicate-name check earlier in the same loop.

src/libasr/pass/implied_do_loops.cpp
```diff
@@ -937,6 +937,45 @@ class ArrayConstantVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayC
...
+            if (x.m_return_var) {
+                ASR::expr_t** current_expr_copy_1 = current_expr;
+                current_expr = const_cast<ASR::expr_t**>(&(x.m_return_var));
+                this->call_replacer();
+                current_expr = current_expr_copy_1;
+                if (x.m_return_var && visit_expr_after_replacement)
+                    this->visit_expr(*x.m_return_var);
+            }
+            current_scope = current_scope_copy;
+        }
```
certik, line 944: I don't understand what this is for.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -175,26 +175,41 @@ private :
...
+        // enclosing Module -- a symbol declared inside a nested function
+        // (e.g. a PARAMETER local to a CONTAINS'd function) must produce
+        // an ExternalSymbol whose m_module_name is that function's name,
+        // not the name of some outer module several scopes up.
+        ASR::symbol_t *sym_owner = ASRUtils::get_asr_owner(sym_target);
+        if (sym_owner != nullptr &&
+            !ASR::is_a<ASR::Module_t>(*sym_owner) &&
+            !ASR::is_a<ASR::Function_t>(*sym_owner)) {
+            sym_owner = nullptr;
+        }
```
certik: And this.

src/libasr/pass/subroutine_from_function.cpp
```diff
@@ -175,26 +175,41 @@ private :
...
+        char* const unique_name = s2c(al_, current_scope_->get_unique_name(ASRUtils::symbol_name(sym_target)));
         auto ext_sym = ASR::down_cast<ASR::symbol_t>(
                 ASR::make_ExternalSymbol_t(al_, sym->base.loc, 
-                                    current_scope_, unique_name, sym,
-                                    sym_mod->m_name, nullptr, 0,
-                                    ASRUtils::symbol_name(sym), ASR::Private));
+                                    current_scope_, unique_name, sym_target,
+                                    s2c(al_, ASRUtils::symbol_name(sym_owner)), nullptr, 0,
+                                    ASRUtils::symbol_name(sym_target), ASR::Private));
         current_scope_->add_symbol(unique_name, ext_sym);
```
certik, line 198: And this change

certik (commented):
It seems this PR is mixing several unrelated fixes. Keep only the simple fix for templates, and separate the do loop fix into a dedicated PR with a proper test there.

src/lfortran/semantics/ast_symboltable_visitor.cpp
certik: No, it's not pre-existing logic, you added it in this PR, see https://github.com/lfortran/lfortran/pull/12178#pullrequestreview-4688900679.

usefahmed07: Ok no problem

usefahmed07:
@certik You're right, apologies - those changes were leftovers from another branch (PR #12144) that got mixed in here by mistake.

I've cleaned up the branch and force-pushed. It now contains only the single commit for the duplicate parameter check (#11347) — nothing else. The unrelated do-loop / subroutine_from_function changes are no longer part of this PR.

tests/reference/asr-template_error_10-7c4bf68.stderr
```diff
@@ -0,0 +1,9 @@
+semantic error: Parameter 't' is declared more than once in requirement 'op_r'
+  --> tests/errors/template_error_10.f90:4:5 - 13:1
+   |
+ 4 |        requirement op_R ( T, T, V, V, op_func )
```
certik: Can you highlight the parameter `T` directly? Not the whole requirement. Change the location information to be more precise.

usefahmed07: Done

src/lfortran/parser/semantics.h
```diff
@@ -2766,10 +2787,11 @@ ast_t* TYPEPARAMETER0(Allocator &al,
         REDUCE_ARGS(p.m_a, namelist), namelist.size(), \
         /*unit_decl2_t** a_decl*/ DECLS(decl), /*size_t n_decl*/ decl.size(), \
         /*contains*/ CONTAINS(contains), /*n_contains*/ contains.size())
-#define REQUIREMENT(name, namelist, decl, funcs, l) \
+#define REQUIREMENT(name, namelist, decl, funcs, l) ( \
+        check_duplicate_namelist_params(p, namelist, "requirement", LCompilers::to_lower(name2char(name))), \
```
certik: This should not happen here --- this is for parsing, we should always parse to AST. Then raise semantic errors in the semantic phase.

tests/errors/template_error_10.f90
```diff
@@ -0,0 +1,24 @@
...
+      require :: op_R ( T, T, V, V, op_func )
+    contains
+      pure elemental function call_op_func ( x ) result(res)
+        type(T) :: x
+        type(V) :: res
+        res = op_func(x,x)
+      end function
+    end template
+
+end module
```
certik: We should create continue_compilation_1.f90 style error tests for templates and then ensure we can recover from template errors.

usefahmed07: OK i will make it

src/lfortran/parser/semantics.h
```diff
@@ -966,6 +967,7 @@ static inline char** REDUCE_ARGS(Allocator &al, const Vec<ast_t*> args)
     return a;
 }
 
+
```
certik: You can now remove these additions

tests/tests.toml
```diff
@@ -5227,6 +5231,9 @@ run = true
 filename = "../integration_tests/shadow_same_symbol.f90"
 run = true
 
+[[test]]
+filename = "errors/continue_compilation_templates_01.f90"
+semantics_only_cc = true
```
certik, line 5246:
```suggestion
[[test]]
filename = "errors/continue_compilation_templates_01.f90"
semantics_only_cc = true

```

tests/errors/template_error_10.f90
certik: Remove this file

tests/tests.toml
```diff
@@ -498,6 +498,10 @@ asr = true
 filename = "errors/template_error_09.f90"
 asr = true
 
+[[test]]
+filename = "errors/template_error_10.f90"
+asr = true
+
```
```suggestion
```

tests/errors/continue_compilation_templates_01.f90
```diff
@@ -0,0 +1,30 @@
...
+        interface
+            function comp_func(x, y) result(z)
+                type(V), intent(in) :: x
+                type(V), intent(in) :: y
+                logical :: z
+            end function
+        end interface
+    end requirement
+
+end module continue_compilation_templates_01_mod
```
certik, line 30: Add a main program that can build in principle, so that it's a full program. You can leave the template in the module, that's fine.

Over time we'll add all the template errors here.

tests/reference/asr-template_error_10-7c4bf68.stderr
certik: It's not done. It needs to highlight just the T parameter, ideally both T's.

certik (commented):
Thanks for the improvements. I left more comments.

grammar/AST.asdl
```diff
@@ -70,7 +70,7 @@ unit_decl2
         program_unit* contains)
     | Enum(decl_attribute* attr, trivia? trivia, unit_decl2* items)
     | Instantiate(identifier name, decl_attribute* args, use_symbol* symbols)
-    | Requirement(identifier name, identifier* namelist, unit_decl2* decl,
+    | Requirement(identifier name, arg* namelist, unit_decl2* decl,
```
certik: What is the purpose of this change? Was the previous code incorrect?

src/lfortran/parser/semantics.h
```diff
@@ -2768,7 +2784,7 @@ ast_t* TYPEPARAMETER0(Allocator &al,
         /*contains*/ CONTAINS(contains), /*n_contains*/ contains.size())
 #define REQUIREMENT(name, namelist, decl, funcs, l) \
         make_Requirement_t(p.m_a, l, name2char(name), \
-        REDUCE_ARGS(p.m_a, namelist), namelist.size(), \
+        REQ_ARGS(p.m_a, namelist), namelist.size(), \
```
certik: Was the previous code incorrect?

tests/tests.toml
```diff
@@ -498,6 +498,7 @@ asr = true
 filename = "errors/template_error_09.f90"
 asr = true
 
+
```
certik, line 505: Don't add empty lines

usefahmed07: Done

src/lfortran/parser/semantics.h
usefahmed07: No the previous code wasnotincorrect  it worked fine when namelist was identifier*. Since we changed Requirement's AST namelist to arg_t* (to carry per-item Location for precise duplicate-parameter error highlighting, per your earlier feedback), REDUCE_ARGS (which only produces char**) can no longer populate it. REQ_ARGS was added to build arg_t* with both the name and its location

grammar/AST.asdl
usefahmed07: The previous code was fine identifier* just didn't carry a per-item Location. Since we now need to point the duplicate-parameter error at the exact token (not the whole requirement), I switched namelist to arg_t*, same as regular function args, so each entry keeps its own location. REQ_ARGS builds that instead of REDUCE_ARGS, which only gave plain names

assem2002: https://github.com/lfortran/lfortran/blob/1516cebacf9109e95225f9e1149cffd894c4f685/grammar/AST.asdl#L394-L397

apparently we needed to register the location for each argument, to show a good warning message

src/lfortran/parser/semantics.h
```diff
@@ -956,6 +957,20 @@ static inline arg_t* ARGS(Allocator &al, Location &l,
...
+        if (args.p[i]) {
+            a[i].loc = args.p[i]->loc;
+            a[i].m_arg = name2char(args.p[i]);
+        } else {
+            a[i].loc = Location();
+            a[i].m_arg = nullptr;
+        }
+    }
+    return a;
+}
```
assem2002: I can see that you have a better a function than the one we have `ARGS()`.
Improve `ARGS()` to match this implementation, and use that instead. 
I hope it doesn't break anything ( i don't think it will, it may even introduce more proper error messages)

src/lfortran/parser/semantics.h
```diff
@@ -11,6 +11,7 @@
 */
 
 #include <cstring>
+#include <set>
```
assem2002: not needed.

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -4472,8 +4473,27 @@ class SymbolTableVisitor : public CommonVisitor<SymbolTableVisitor> {
...
+        {
+            std::map<std::string, Location> seen_args;
+            for (size_t i=0; i<x.n_namelist; i++) {
+                std::string arg = to_lower(x.m_namelist[i].m_arg);
+                auto it = seen_args.find(arg);
+                if (it != seen_args.end()) {
+                    diag.add(Diagnostic(
+                        "Parameter '" + arg + "' is declared more than once in "
+                        "requirement '" + to_lower(x.m_name) + "'",
+                        Level::Error, Stage::Semantic, {
```
assem2002, line 4524: shouldn't we raise a warning? I think duplicates are permissible, right?

src/lfortran/semantics/ast_symboltable_visitor.cpp
```diff
@@ -1,3 +1,4 @@
+#include <set>
```
assem2002: not needed.

assem2002 (approved):
LGTM.
Pleas address my comments

src/lfortran/parser/semantics.h
usefahmed07: Removed, thanks.

src/lfortran/semantics/ast_symboltable_visitor.cpp
usefahmed07: > shouldn't we raise a warning? I think duplicates are permissible, right?

I don't think a duplicate name here should just warn it makes the symbol
table build itself ambiguous, and if we only warn, the user still hits the
same misleading downstream error that https://github.com/lfortran/lfortran/issues/11347 was originally about. Keeping
it a hard error catches it right at the source instead

assem2002 (approved):
looks good, 
I just ported your `REQ_ARGS` to the already existing `ARGS()`

PR #12179, opened 2026-07-12
fix: emit real object files for submodule library units

HarshitaKalani:


src/bin/lfortran.cpp
```diff
@@ -1233,7 +1232,19 @@ int compile_src_to_object_file(const std::string &infile,
...
     }
 
-    // Save .mod files
+    bool submodule_lib_unit = arg_c
+        && LCompilers::ASRUtils::submodule_present(*asr)
+        && !LCompilers::ASRUtils::main_program_present(*asr)
+        && !LCompilers::ASRUtils::global_function_present(*asr);
+    bool user_separate_compilation = compiler_options.separate_compilation;
+    if (submodule_lib_unit) {
+        compiler_options.separate_compilation = true;
```
certik, line 1241: From AI:

"`compiler_options` is passed by reference and the same instance is reused for every source in the multi-file compilation loop, but this forced setting is never restored. After compiling a submodule library unit, every later input is therefore parsed with separate-compilation semantics. For example, in `lfortran parent.f90 child.f90 main.f90`, the main program no longer loads the submodules/definitions from the parent module as normal, while the earlier parent object may be empty, producing unresolved symbols at link time. Please scope these forced options to this compilation or restore all mutated fields before returning."

I think this seems like a bug?

src/bin/lfortran.cpp
```diff
@@ -1242,12 +1253,22 @@ int compile_src_to_object_file(const std::string &infile,
         if (err) return err;
     }
 
+    if (submodule_lib_unit) {
+        if (!user_separate_compilation) {
+            LCompilers::ASRUtils::mark_local_regular_modules_as_external_when_submodule_present(*asr);
```
certik, line 1258: AI said:

"This externalizes every locally defined regular module whenever the source also contains a submodule. Such a module is not loaded from a `.mod` file and this object is its only compilation output, so its procedures/variables are omitted entirely. A valid file containing a helper module followed by a submodule that uses it will consequently fail to link. Only loaded modules should be externalized; locally defined regular modules still need to be emitted into this object."

It would be good to understand if this bug is present. If so, to fix it.

certik:
Can you explain what this PR is doing in terms of changing how we compile submodules?

We have to be super careful how this is done. Can you explain the failure, and then how it is fixed? It seems the current fix introduces more bugs and also possibly changes the current design when we emit code.

PR #12193, opened 2026-07-13
README: add JupyterLite demo link

anutosh491:
It will be useful for users/contributors to try lfortran directly in their browser through Jupyterlite so adding the link in the README. Also it's the latest build as the deployment updates with every commit going into main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ Website: https://lfortran.org/
 
 Try online: https://dev.lfortran.org/
 
+Try LFortran in a JupyterLite notebook:
+[![JupyterLite](https://jupyterlite.rtfd.io/en/latest/_static/badge.svg)](https://lfortran.github.io/lfortran/)
+
 # Documentation
 
 All documentation, installation instructions, motivation, design, ... is


README.md
```diff
@@ -12,6 +12,9 @@ Website: https://lfortran.org/
 
 Try online: https://dev.lfortran.org/
 
+Try LFortran in a JupyterLite notebook:
+[![JupyterLite](https://jupyterlite.rtfd.io/en/latest/_static/badge.svg)](https://lfortran.github.io/lfortran/)
+
```
Jatinagarwal24, line 17: Hi @anutosh491 for the consistency of style.
The existing "Try online" line (13 th line)  uses a bare URL format (`Try online: https://dev.lfortran.org/`). The new entry uses a badge image instead.
Consider whether a consistent style would be preferable, e.g.:
```
Try in JupyterLite notebook: https://lfortran.github.io/lfortran/
```
Or, alternatively, keep the badge but also include the raw URL for accessibility (screen readers may not read alt text from badges well) .

Jatinagarwal24: Tell me what do you think.

anutosh491: Ahh I'm fine with anything. Just thought of using the jupyterlite cause it conveys that everything is wasm powered !

anutosh491: Do you think we can just keep the badge somehow ?

Jatinagarwal24: > Ahh I'm fine with anything. Just thought of using the jupyterlite cause it conveys that everything is wasm powered !

Ah got it .The badge does make it visually clear that it's JupyterLite powered. Let's keep it as is.

Jatinagarwal24 (approved):
Looks good to me.
Thanks!

Jatinagarwal24:
@anutosh491 We don't need exhaustive tests for this pr right ?

anutosh491:
Nope nope

README.md
certik: I think the badge is fine for now. We can always adjust / remove it later if we don't like it.

PR #12194, opened 2026-07-13
fix: set target triple on LLVM module during IR generation

anutosh491:
While trying the `%%showllvm` in the xeus kernel I realized the target triple is never set. I particularly needed it to show the `wasm32-unknown-emscripten` triple for the Jupyterlite notebooks.

image

So previously `--show-llvm` and `%%showllvm` (Jupyter kernel) were missing the 'target triple' line because `setTargetTriple()` was never called in `visit_TranslationUnit()`. The data layout was already being set
using the default target triple, so this just adds the matching `setTargetTriple()` call in the same block.

### DIFF
--- a/build1.sh
+++ b/build1.sh
@@ -3,7 +3,11 @@
 set -e
 set -x
 
+BUILD_DIR="build-lfortran"
+
 cmake \
+    -S . \
+    -B "$BUILD_DIR" \
     -DCMAKE_BUILD_TYPE=Debug \
     -DWITH_LLVM=yes \
     -DLFORTRAN_BUILD_ALL=yes \
@@ -17,6 +21,6 @@ cmake \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
     -DCMAKE_C_FLAGS="${CFLAGS} -fdiagnostics-color=always" \
     -DCMAKE_CXX_FLAGS="${CXXFLAGS} -fdiagnostics-color=always" \
-    -G Ninja \
-    .
-cmake --build .
+    -G Ninja
+
+cmake --build "$BUILD_DIR"
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -1841,6 +1841,11 @@ public:
             } else {
                 module->setDataLayout("");
             }
+#if LLVM_VERSION_MAJOR >= 21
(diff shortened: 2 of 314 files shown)


anutosh491:
This is now fixed 

on the native kernel 
image

Also for a simple hello world file 
```
(lf) anutosh491@Anutoshs-MacBook-Air bin % ./lfortran --show-llvm hello_subroutine.f90                               
; ModuleID = 'LFortran'
source_filename = "LFortran"
target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-n32:64-S128-Fn32"
target triple = "arm64-apple-darwin25.5.0"

%string_descriptor = type <{ ptr, i64 }>
.....
.....
```

anutosh491:
Ahh I think this might need some update to the reference IRs in the tests. Not sure how that can be done !

Is there a command I need to execute ?

Jatinagarwal24:
> Ahh I think this might need some update to the reference IRs in the tests. Not sure how that can be done !
> 
> Is there a command I need to execute ?

Just use `./run_tests.py -u`

anutosh491:
Ahhh the tests might complain :\
```
subroutine3.f90 * llvm
Reference JSON: tests/reference/llvm-subroutine3-fb7e0b8.json
Output JSON:    tests/output/llvm-subroutine3-fb7e0b8.json
Omitting 10 identical items
ACTUAL != REFERENCE
Differing items:
{'stdout_hash': '88d4820b20ceb37e9bfd1dbabf64df2ce95a34f30e6e6dfd1acaf760'} != {'stdout_hash': 'd3a392f16ebb0a59254befd2e01e2cc97bfbd150bde27fcc942e948a'}
=== DIFF ===
Reference: tests/reference/llvm-subroutine3-fb7e0b8.stdout
Output:    tests/output/llvm-subroutine3-fb7e0b8.stdout
4c4
< target triple = "arm64-apple-darwin25.5.0"
---
> target triple = "x86_64-unknown-linux-gnu"
9c9
```

certik:
Couple things:

* You need to use LLVM 11 when updating tests, so that they don't change
* We need to sanitize the triple out in printing the tests, since we use different platforms to generate the tests (macos/linux, etc.)

Question: this triple was never set even when generating a binary? Isn't it set automatically? If it is not set, and we enable optimizations, does it mean the binary will not be as optimized? What exactly happens when you set this triple --- what advantages do you get?

PR #12196, opened 2026-07-13
fix: namelist derived array read ordering

opixdown:


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -14264,33 +14264,24 @@ static void nml_value_iter_init(nml_value_iter_t *it, lfortran_nml_item_t **item
...
-}
-
+// Namelist whole-derived-object assignments are consumed element-major:
+// for items=1,2,3,4 with components a,b, assign items(1)%a, items(1)%b,
+// then items(2)%a, items(2)%b. With n_items == 1 this is the old array walk.
 static bool nml_value_iter_advance(nml_value_iter_t *it) {
-    it->value_idx++;
-    if (it->value_idx < it->total_size) {
+    it->item_idx++;
+    if (it->item_idx < it->n_items) {
```
certik, line 14272: AI said:

"This now advances across flattened components before advancing each component's value index, while `total_size` remains the size of only the first component. That is valid for an array of derived objects, where each flattened item represents the same outer extent, but not for a scalar derived object with array components. For example, `type t; integer :: a(2); integer :: b; end type` and values `x=1,2,3` should fill `a` before `b`; this traversal instead visits `a(1), b, a(2)` and can index a scalar component beyond its extent. Please retain component-major traversal for scalar derived objects (and add coverage for mixed component extents), using element-major traversal only for derived-type arrays."

This is likely correct. Create an MRE for this failure mode first, to ensure it is a real bug. Then let's fix it.

PR #12201, opened 2026-07-13
fix: preload runtime .mod files for iso_c_binding in JupyterLite kernel

### LINKED ISSUE #11712: iso_c_binding intrinsic module not found in JupyterLite
This works fine in the native kernel

Image

But fails in the wasm kernel

Image

anutosh491:
Fixes #11712

The wasm xlfortran kernel was missing the runtime .mod files (e.g. lfortran_intrinsic_iso_c_binding.mod), causing 'modfile was not found' errors when using `use iso_c_binding` in JupyterLite cells.

image

Three changes fix this:
- utils.cpp: under `__EMSCRIPTEN__`, `get_runtime_library_dir()` now returns "/lib", matching where empack mounts $PREFIX/lib/ in the VFS.
- CMakeLists.txt: when `XEUS_LFORTRAN_WASM_BUILD=yes`, preload each .mod from $PREFIX/lib/ into the WASM VFS at /lib/ via --preload-file, and install the generated `xlfortran.data` alongside `xlfortran.js/.wasm`.
- wasm-build0.sh (new): compiles the runtime .f90 sources with the native lfortran and installs the resulting .mod files to $PREFIX/lib/ so that wasm-build1.sh finds them during the emcmake configure step.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -741,8 +741,9 @@ jobs:
       - name: Run WASM evaluator tests
         run: |
           set -ex
+          cd build-wasm/src/lfortran/tests
           micromamba run -n xeus-lfortran-wasm-build \
-            node build-wasm/src/lfortran/tests/test_lfortran.js
+            node test_lfortran.js
 
       - name: Build JupyterLite site
         run: |
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -723,16 +723,20 @@ jobs:
       - name: Install WASM build environment
         run: micromamba create -f environment-wasm-build.yml -y
 
-      - name: Build WASM kernel
+      - name: Create WASM host environment
         run: |
           set -ex
           micromamba create -f environment-wasm-host.yml -y \
             --platform=emscripten-wasm32 \
             -c https://prefix.dev/emscripten-forge-4x \
             -c https://prefix.dev/conda-forge
-          export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-lfortran-wasm-host
-          echo "PREFIX=$PREFIX" >> $GITHUB_ENV
-          micromamba run -n xeus-lfortran-wasm-build ./wasm-build1.sh
+          echo "PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-lfortran-wasm-host" >> $GITHUB_ENV
+
+      - name: Build native runtime .mod files
+        run: ./wasm-build0.sh
+
+      - name: Build WASM kernel
+        run: micromamba run -n xeus-lfortran-wasm-build ./wasm-build1.sh
 
       - name: Run WASM evaluator tests
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
     xeus_wasm_compile_options(xlfortran)
     xeus_wasm_link_options(xlfortran "web,worker")
 
+    # Preload runtime .mod files into the WASM VFS at /lib/
+    file(GLOB LFORTRAN_MOD_FILES "${CMAKE_INSTALL_PREFIX}/lib/*.mod")
+    foreach(MOD_FILE ${LFORTRAN_MOD_FILES})
+        get_filename_component(MOD_NAME ${MOD_FILE} NAME)
+        target_link_options(xlfortran PRIVATE
+            "SHELL:--preload-file ${MOD_FILE}@/lib/${MOD_NAME}"
+        )
+    endforeach()
+
     install(FILES
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.js"
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.wasm"
+        "$<TARGET_FILE_DIR:xlfortran>/xlfortran.data"
         DESTINATION ${CMAKE_INSTALL_BINDIR}
     )
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -419,16 +419,25 @@ if (XEUS_LFORTRAN_WASM_BUILD)
     foreach(MOD_FILE ${LFORTRAN_MOD_FILES})
         get_filename_component(MOD_NAME ${MOD_FILE} NAME)
         target_link_options(xlfortran PRIVATE
-            "SHELL:--preload-file ${MOD_FILE}@/lib/${MOD_NAME}"
+            "SHELL:--preload-file \"${MOD_FILE}@/lib/${MOD_NAME}\""
         )
     endforeach()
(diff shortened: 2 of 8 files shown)


anutosh491:
Now fixed 

image

certik:
Thanks for this. Need to review carefully.

wasm-build0.sh
```diff
@@ -0,0 +1,52 @@
...
+# Compile runtime modules using the freshly built lfortran.
+LFORTRAN="$(pwd)/asset_dir/src/bin/lfortran"
+MODDIR="$(pwd)/asset_dir/mods"
+RTDIR="$(pwd)/src/runtime"
+mkdir -p "$MODDIR"
+
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_iso_fortran_env.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/custom/lfortran_intrinsic_custom.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" -I "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_ieee_arithmetic.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_iso_c_binding.f90"
```
certik, line 48: Looks like there is some bug there, but probably this can be fixed later.

--------

This packages an `iso_c_binding.mod` whose source hardcodes `c_long`, `c_size_t`, `c_intptr_t`, and `c_ptrdiff_t` to kind 8. In the actual JupyterLite kernel I observed `8 8 8 8`, while Emscripten wasm32 defines both `__SIZEOF_LONG__` and `__SIZEOF_POINTER__` as 4. Consequently, `integer(c_long)`/`integer(c_size_t)` are lowered as i64 while the corresponding C ABI types are i32, breaking `bind(C)` argument and result marshalling. Recompiling the same source for WASM would not fix hardcoded values; these intrinsic kinds (and the hardcoded kind-8 `c_sizeof` result) need to become target-aware before exposing this module on wasm32.

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
     xeus_wasm_compile_options(xlfortran)
     xeus_wasm_link_options(xlfortran "web,worker")
 
+    # Preload runtime .mod files into the WASM VFS at /lib/
+    file(GLOB LFORTRAN_MOD_FILES "${CMAKE_INSTALL_PREFIX}/lib/*.mod")
```
certik, line 418: The matched `.mod` files are only embedded as link-option strings, so CMake does not track their contents as link dependencies. Already reported in the other PR.

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
...
+    foreach(MOD_FILE ${LFORTRAN_MOD_FILES})
+        get_filename_component(MOD_NAME ${MOD_FILE} NAME)
+        target_link_options(xlfortran PRIVATE
+            "SHELL:--preload-file ${MOD_FILE}@/lib/${MOD_NAME}"
```
certik: Add quotes, already reported in the other PR.

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
...
+        )
+    endforeach()
+
     install(FILES
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.js"
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.wasm"
+        "$<TARGET_FILE_DIR:xlfortran>/xlfortran.data"
```
certik: `xlfortran.data` exists only when at least one `--preload-file` option is present. Already reported in the other PR.

.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -724,16 +724,20 @@ jobs:
...
             -c https://prefix.dev/emscripten-forge-4x \
             -c https://prefix.dev/conda-forge
-          export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-lfortran-wasm-host
-          echo "PREFIX=$PREFIX" >> $GITHUB_ENV
-          micromamba run -n xeus-lfortran-wasm-build ./wasm-build1.sh
+          echo "PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-lfortran-wasm-host" >> $GITHUB_ENV
+
+      - name: Build native runtime .mod files
+        run: ./wasm-build0.sh
+
```
certik, line 737: The later Node test runs the standalone `test_lfortran` target, not `xlfortran`; it has no preload options and never executes `use iso_c_binding` from `/lib`. Thus the behavior fixed by this PR is still untested in CI. 

Is this tested in the other PR?

certik (commented):
I reported most of these in the other PR. But perhaps they have to be fixed here first.

CMakeLists.txt
anutosh491: Done.

CMakeLists.txt
anutosh491: Done.

CMakeLists.txt
anutosh491: Done.

wasm-build0.sh
anutosh491: Woah okay !!

i didn't realize this at all 

Makes sense to open an issue for it and address it on a different branch (https://github.com/lfortran/lfortran/issues/12236)

anutosh491: Won't this also play a role in the lfortran.js build 

https://github.com/lfortran/lfortran/blob/487c1335fcf09d5a9d7fe1f14818db2aad756dca/build_to_wasm.sh#L17

.github/workflows/Exhaustive-Checks-CI.yml
anutosh491: Yes absolutely a test was pending for this which I've added and it passes as required 
```
[doctest] test cases:  29 |  29 passed | 0 failed | 0 skipped
[doctest] assertions: 190 | 190 passed | 0 failed |
[doctest] Status: SUCCESS!
```

anutosh491:
Guessing this should be in a position to move ahead with now. I'll look at the reviews on Rich Display PR once this goes in (would need rebasing on this)

src/lfortran/utils.cpp
```diff
@@ -65,6 +65,11 @@ std::string get_runtime_library_dir()
 {
 #ifdef HAVE_BUILD_TO_WASM
     return "asset_dir";
+#endif
+#ifdef __EMSCRIPTEN__
+    // JupyterLite (xlfortran): empack mounts $PREFIX at / in the VFS,
+    // so runtime .mod files installed to $PREFIX/lib end up at /lib.
+    return "/lib";
```
certik, line 72: This might be problematic --- we also build with emscripten for https://dev.lfortran.org/, and this might break it?

certik (approved):
I think it's ok to merge, we'll have to iterate on it.

PR #12202, opened 2026-07-13
fix: preserve high bits in ishftc when SIZE is given

### LINKED ISSUE #12200: ishftc returns wrong value when the optional SIZE argument is given
## MRE

```fortran
program p
  integer :: i
  i = ishftc(120, 2, 4)
  print *, i
  if (i /= 114) error stop
end program
```

## LFortran behavior

```text
$ lfortran t.f90
2
ERROR STOP
```

LFortran prints `2` and then error-stops on the check.

## gfortran behavior

```text
$ gfortran t.f90 && ./a.out
         114
```

## notes

`ishftc(I, SHIFT, SIZE)` must rotate only the rightmost `SIZE` bits of `I` and leave the higher bits unchanged. For `I = 120 = 0b01111000`, `SHIFT = 2`, `SIZE = 4`: the low 4 bits `1000` rotate left by 2 to `0010`, and the high bits `01110000` (= 112) must be preserved, giving `112 | 2 = 114`. LFortran returns just `2`, having discarded the high bits.

The 2-argument form is fine:

```fortran
print *, ishftc(120, 2)           ! 480, correct
print *, ishftc(120, 2, 4)        ! 2,   wrong (expected 114)
print *, ishftc(120_1, 2_1, 4_1)  ! 2,   wrong for kinds 1/2/4/8 (expected 114)
print *, ishftc([120,7,8],[2,2,1],[4,3,4])  ! [2,7,1], wrong (expected [114,7,1])
```

The high bits are masked off with `cutoff_extra_bits(...)` and never restored, in three places:

- `src/libasr/pass/intrinsic_functions.h`, `eval_Ishftc` (compile-time fold)
- `src/libasr/runtime/lfortran_intrinsics.c`, `_lfortran_sishftc` (runtime, int32)
- `src/libasr/runtime/lfortran_intrinsics.c`, `_lfortran_dishftc` (runtime, int64)

ambikeesshh:
fixes #12200

`ishftc(i, shift, size)` was throwing away the bits of `i` above position `size` instead of leaving them alone, so e.g. `ishftc(120, 2, 4)` returned `2` rather than `114`.

the high bits are now kept off to the side and OR'd back in after the rotation. fixed in all three spots (the compile-time fold and the two runtime functions). adds `intrinsics_475` covering it.

```
$ lfortran t.f90   # print *, ishftc(120, 2, 4)
114
```

### DIFF
--- a/integration_tests/intrinsics_475.f90
+++ b/integration_tests/intrinsics_475.f90
@@ -60,6 +60,22 @@ program intrinsics_475
     res_8 = ishftc(120_8, -2_8, 4_8)
     if (res_8 /= 114_8) error stop
 
+    ! ----- SHIFT == 0: value is unchanged (early-return path) -----
+
+    if (ishftc(120_4, 0_4, 4_4)   /= 120)           error stop
+    if (ishftc(120_4, 0_4, 32_4)  /= 120)           error stop
+    if (ishftc(120_8, 0_8, 64_8)  /= 120_8)         error stop
+    if (ishftc(huge(0_8), 0_8, 64_8) /= huge(0_8))  error stop
+
+    ! ----- SIZE == BIT_SIZE(I): full-width rotation (mask == ~0ULL branch) -----
+
+    ! kind=4, SIZE=32
+    if (ishftc(120_4, 2_4, 32_4)  /= 480)           error stop
+    ! kind=8, SIZE=64 -> exercises the bits_size == 64 mask branch
+    if (ishftc(120_8, 2_8, 64_8)  /= 480_8)         error stop
+    if (ishftc(120_8, -2_8, 64_8) /= 30_8)          error stop
+    if (ishftc(huge(0_8), 2_8, 64_8) /= -3_8)       error stop
+
     print *, "pass"
 
 contains
--- a/integration_tests/intrinsics_475.f90
+++ b/integration_tests/intrinsics_475.f90
@@ -1,0 +1,79 @@
+program intrinsics_475
+    ! Verify ishftc with the optional SIZE argument preserves the bits of I
+    ! that lie above position SIZE. Only the rightmost SIZE bits are rotated;
+    ! the higher bits must be left unchanged. (See F2018 16.9.130.)
+
+    implicit none
+
+    integer(kind=1) :: res_1
+    integer(kind=2) :: res_2
+    integer(kind=4) :: res_4
+    integer(kind=8) :: res_8
+
+    integer(kind=4), parameter :: p1 = ishftc(120_4, 2_4, 4_4)
+    integer(kind=8), parameter :: p2 = ishftc(120_8, -2_8, 4_8)
+
+    ! ----- compile-time fold (parameter) path -----
+
+    ! 120 = 0b01111000: low 4 bits 1000 rotate left 2 -> 0010, high 01110000 kept
+    if (p1 /= 114) error stop
+    ! same value rotated right by 2: low 4 bits 1000 rotate right 2 -> 0010
+    if (p2 /= 114_8) error stop
+
+    if (ishftc(7_4, 2_4, 3_4)   /= 7)   error stop   ! 111 rot L2 -> 111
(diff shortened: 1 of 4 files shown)


integration_tests/intrinsics_475.f90
Jatinagarwal24, line 1: The test is good but doesn't cover:

`SIZE == BIT_SIZE(I)`   full-width circular shift (the `mask == ~0` branch). E.g. `ishftc(120_4, 2_4, 32_4)`.
`SHIFT == 0` — the code has a dedicated `shift == 0` early-return path but it's never exercised in tests.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -4448,24 +4448,26 @@ LFORTRAN_API float _lfortran_sbesselyn( int n, float x ) {
...
 }
 
-uint64_t cutoff_extra_bits(uint64_t num, uint32_t bits_size, uint32_t max_bits_size) {
-    if (bits_size == max_bits_size) {
-        return num;
-    }
-    return (num & ((1lu << bits_size) - 1lu));
-}
-
 LFORTRAN_API int _lfortran_sishftc(int val, int shift_signed, int bits_size) {
```
Jatinagarwal24, line 4469: Actually I can see `_lfortran_sishftc` and `_lfortran_dishftc` contain identical rotation logic. A shared static helper like:
```
static uint64_t ishftc_rotate(uint64_t val, uint32_t shift, bool neg, uint32_t bits_size, uint32_t max_bits) { ... }
```
would reduce duplication.

Jatinagarwal24: Pls see if we can do this .
It would be great.

src/libasr/pass/intrinsic_functions.h
```diff
@@ -5111,12 +5104,22 @@ namespace Ishftc {
             return nullptr;
         }
 
-        val = cutoff_extra_bits(val, bits_size, max_bits_size);
+        // Rotate only the low `bits_size` bits of val; leave the higher bits
+        // unchanged. The high bits are captured before the rotation and OR'd
+        // back into the result, so they survive even when bits_size < bit_width.
+        uint64_t mask = (bits_size == max_bits_size)
+            ? ~0lu
+            : ((1lu << bits_size) - 1lu);
```
Jatinagarwal24: Just check if `~0lu` and `1lu` are 64-bit safe on MSVC or not.
If not then we can change it to 
` ~0lu → ~(uint64_t)0 ` and `1lu → (uint64_t)1` in all three locations.

Jatinagarwal24 (commented):
Looks good.
Just address the comments.

src/libasr/pass/intrinsic_functions.h
ambikeesshh: switched to 0ULL/1ULL in all three spots (incl. the new helper). On LLP64 windows unsigned long is 32-bit, so ~0lu would have been a wrong mask for SIZE==64 and (1lu << bits_size) would be UB past 31.

src/libasr/runtime/lfortran_intrinsics.c
ambikeesshh: done. extracted a static ishftc_rotate() helper; both _lfortran_sishftc and _lfortran_dishftc now just compute the abs shift and call it.

integration_tests/intrinsics_475.f90
ambikeesshh: done. added SHIFT==0 cases and SIZE==BIT_SIZE(I) cases (incl. ishftc(x_8, s_8, 64_8) which hits the bits_size==64 / mask==~0ULL branch).

ambikeesshh:
@Jatinagarwal24, thanks for the review
pushed a follow-up commit addressing all three comments

Jatinagarwal24 (approved):
Thanks @ambikeesshh for the fix.
Looks good to me.

HarshilShah1804 (approved):
Looks good.

ambikeesshh:
tysm!

PR #12205, opened 2026-07-13
feat: add rich display support to xeus-lfortran kernel

anutosh491:
Had shared this recording on Zulip sharing here again !

https://github.com/user-attachments/assets/7215968b-d066-4f2e-9821-866149b4cccc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -415,29 +415,29 @@ if (XEUS_LFORTRAN_WASM_BUILD)
     xeus_wasm_link_options(xlfortran "web,worker")
 
     # Preload runtime .mod files into the WASM VFS at /lib/
-    file(GLOB LFORTRAN_MOD_FILES "${CMAKE_INSTALL_PREFIX}/lib/*.mod")
+    file(GLOB LFORTRAN_MOD_FILES CONFIGURE_DEPENDS "${CMAKE_INSTALL_PREFIX}/lib/*.mod")
+    if(NOT LFORTRAN_MOD_FILES)
+        message(FATAL_ERROR
+            "XEUS_LFORTRAN_WASM_BUILD: no runtime .mod files found in "
+            "${CMAKE_INSTALL_PREFIX}/lib. Run wasm-build0.sh first."
+        )
+    endif()
     foreach(MOD_FILE ${LFORTRAN_MOD_FILES})
         get_filename_component(MOD_NAME ${MOD_FILE} NAME)
         target_link_options(xlfortran PRIVATE
             "SHELL:--preload-file \"${MOD_FILE}@/lib/${MOD_NAME}\""
         )
     endforeach()
-    if(LFORTRAN_MOD_FILES)
-        # Track .mod file contents as link dependencies so relinking is
-        # triggered when any preloaded module changes.
-        set_property(TARGET xlfortran APPEND PROPERTY LINK_DEPENDS ${LFORTRAN_MOD_FILES})
-    endif()
+    # Track .mod file contents as link dependencies so relinking is
+    # triggered when any preloaded module changes.
+    set_property(TARGET xlfortran APPEND PROPERTY LINK_DEPENDS ${LFORTRAN_MOD_FILES})
 
     install(FILES
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.js"
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.wasm"
+        "$<TARGET_FILE_DIR:xlfortran>/xlfortran.data"
         DESTINATION ${CMAKE_INSTALL_BINDIR}
     )
-    # xlfortran.data is only generated when at least one --preload-file is used.
-    if(LFORTRAN_MOD_FILES)
-        install(FILES "$<TARGET_FILE_DIR:xlfortran>/xlfortran.data"
-                DESTINATION ${CMAKE_INSTALL_BINDIR})
-    endif()
 endif()
 
 # JSON OR LSP (`conda install rapidjson`)
--- a/share/lfortran/nb/Rich_Display.ipynb
+++ b/share/lfortran/nb/Rich_Display.ipynb
@@ -1,0 +1,270 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Rich Display in xeus-lfortran\n",
+    "\n",
+    "xeus-lfortran supports Jupyter's rich MIME display system via the `lfortran_display` module,\n",
+    "injected automatically at kernel startup.\n",
+    "\n",
+    "The single primitive is `display_data(mime_type, data)`.\n",
+    "\n",
+    "| Call | MIME type | Use for |\n",
+    "|---|---|---|\n",
+    "| `display_data('text/html', ...)` | `text/html` | Styled text, tables, widgets |\n",
+    "| `display_data('image/svg+xml', ...)` | `image/svg+xml` | Vector graphics, charts |\n",
+    "| `display_data('text/latex', ...)` | `text/latex` | Math (rendered by MathJax) |\n",
+    "| `display_data('image/bmp', b64)` | `image/bmp` | Pixel images (base64-encoded) |"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1 — HTML"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "use lfortran_display\n",
+    "call display_data('text/html', &\n",
+    "    '<div style=\"font-family:sans-serif;padding:16px;border-left:4px solid #0066cc\">' // &\n",
+    "    '<h2 style=\"color:#0066cc\">Hello from xeus-lfortran!</h2>' // &\n",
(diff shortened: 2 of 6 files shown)


anutosh491:
The way to do this is 

- Add generic `lfortran_display_data` C bridge in `fortran_kernel.cpp`
- Inject `lfortran_display` module at kernel startup (`display_data` subroutine). We need to implement `configure_request_impl `for this which is the first thing that is executed as soon as the kernel is started (before the user can execute anycell)

Added Mandelbrot.ipynb & Rich_Display.ipynb to demo this.

anutosh491:
This is built on top of #12201 and would need a rebase on that PR goes in.

share/lfortran/nb/Rich_Display.ipynb
```diff
@@ -0,0 +1,165 @@
...
...///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////AP/yAf/wAf/uAv/tA//rBP/pBf/oBv/mB//kB//jCP/hCf/fCv/eCv/cC//aDP/ZDP/XDf/VDv/UDv/SD//QEP/PEP/NEf/LEv/KEv/IE//GE//EFP/DFP/BFf+/Ff++Fv+8Fv+6Fv+4F/+3F/+1GP+zGP+xGP+wGf+uGf+sGf+qGf+pGv+nGv+lGv+jGv+hG/+gG/+eG/+cG/+aG/+YG/+WG/+UG/+TG/+RG/+PG/+NG/+LG/+JG/+HG/+FG/+DG/+BG/9/G/99G/97G/95Gv93Gv91Gv9zGv9xGf9vGf9tGf9rGf9pGP9nGP9lGP9jF/9hF/9fFv9cFv9aFv9YFf9WFf9UFP9SFP9PE/9NE/9LEv9JEv9GEf9EEP9CEP9AD/89Dv87Dv85Df82DP80DP8yC/8vCv8tCv8qCf8oCP8mB/8jB/8hBv8eBf8cBP8ZA/8XAv8UAf8SAf8PAP8N/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
certik: We probably should not commit this binary blob into git.

certik:
This is good. Will need to carefully review all the integration parts, which is most of this PR.

share/lfortran/nb/Rich_Display.ipynb
anutosh491: Oops didn't notice this. Yess we should probably not have this.

src/lfortran/fortran_kernel.cpp
```diff
@@ -33,11 +33,67 @@
...
+// WASM: defined in the MAIN_MODULE; side modules import via --allow-undefined
+// and RTLD_DEFAULT resolves them at dlopen time.
+
+extern "C" {
+
+// Generic display bridge: publish arbitrary MIME type + data to Jupyter.
+// Users can send HTML, SVG, LaTeX, Markdown, base64-encoded images, etc.
+// This is the extensibility point - all format-specific encoding (BMP, PNG, etc.)
+// should be done in Fortran user code, not hardcoded here.
+void lfortran_display_data(const char* mime_type, const char* data) {
```
certik: I think this will not work on Windows. But we can fix that later:

----------

This bridge is not exported from the native Windows executable. ORC's `GetForCurrentProcess` ultimately uses `GetProcAddress` on Windows, but this function has neither `LFORTRAN_API`/`__declspec(dllexport)` nor explicit JIT symbol registration; `src/bin/CMakeLists.txt` only enables `--export-dynamic` on ELF platforms. The kernel therefore builds successfully on Windows, but the first `display_data` call fails symbol resolution. Please export or explicitly register this symbol and add a Windows runtime test.

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
     xeus_wasm_compile_options(xlfortran)
     xeus_wasm_link_options(xlfortran "web,worker")
 
+    # Preload runtime .mod files into the WASM VFS at /lib/
+    file(GLOB LFORTRAN_MOD_FILES "${CMAKE_INSTALL_PREFIX}/lib/*.mod")
```
certik: I think this is too fragile and will break in various ways. More details:



----------

The matched `.mod` files are only embedded as link-option strings, so CMake does not track their contents as link dependencies. I modified a preloaded module and reran `wasm-build1.sh`; `xlfortran` did not relink and `xlfortran.data` retained the same SHA-256. This can silently ship stale or format-incompatible modules after an incremental rebuild. Please add the files to `LINK_DEPENDS` (and use `CONFIGURE_DEPENDS` for changes to the matched set).

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
...
+    foreach(MOD_FILE ${LFORTRAN_MOD_FILES})
+        get_filename_component(MOD_NAME ${MOD_FILE} NAME)
+        target_link_options(xlfortran PRIVATE
+            "SHELL:--preload-file ${MOD_FILE}@/lib/${MOD_NAME}"
```
certik: Use quotes.

----------

`SHELL:` reparses this fragment on whitespace, but the preload path is unquoted. With a valid `PREFIX` path containing a space, I reproduced an xlfortran link failure: `em++: error: host: No such file or directory`. Please quote the complete `${MOD_FILE}@/lib/${MOD_NAME}` argument inside the `SHELL:` fragment.

CMakeLists.txt
```diff
@@ -414,9 +414,19 @@ if (XEUS_LFORTRAN_WASM_BUILD)
...
+        )
+    endforeach()
+
     install(FILES
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.js"
         "$<TARGET_FILE_DIR:xlfortran>/xlfortran.wasm"
+        "$<TARGET_FILE_DIR:xlfortran>/xlfortran.data"
```
certik, line 438: Another fragility. 

--------------

`xlfortran.data` is generated only when at least one `--preload-file` option is present. If the configure-time glob is empty, the target builds without a `.data` file and this unconditional install fails with `file INSTALL cannot find .../xlfortran.data`. I reproduced this with a clean host prefix. Since the kernel needs these modules, please fail during configuration with a clear prerequisite error when none are found.

wasm-build0.sh
```diff
@@ -0,0 +1,52 @@
...
+LFORTRAN="$(pwd)/asset_dir/src/bin/lfortran"
+MODDIR="$(pwd)/asset_dir/mods"
+RTDIR="$(pwd)/src/runtime"
+mkdir -p "$MODDIR"
+
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_iso_fortran_env.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/custom/lfortran_intrinsic_custom.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" -I "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_ieee_arithmetic.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" "$RTDIR/pure/lfortran_intrinsic_iso_c_binding.f90"
+$LFORTRAN --backend=cpp -c -J "$MODDIR" -I "$MODDIR" "$RTDIR/openmp/omp_lib.f90"
```
certik, line 51: This will leave temporary files in the top level directory. Probably fine for now, but probably we should build this from a build directory.

share/lfortran/nb/Rich_Display.ipynb
certik: These need to be removed.

certik:
Overall it looks good and it works locally. But there are some integration issues that should be resolved.

share/lfortran/nb/Rich_Display.ipynb
anutosh491: Done.

src/lfortran/fortran_kernel.cpp
anutosh491: I think what I've pushed should work. Don't have a windows device to try it out though !

wasm-build0.sh
anutosh491: Done.

anutosh491:
Address most comments. Let me know if you think anything's pending.

src/lfortran/fortran_kernel.cpp
```diff
@@ -39,6 +39,67 @@
...
+  !   call display_data("text/html", "<h1>Hello</h1>")
+  !   call display_data("image/svg+xml", svg_string)
+  !   call display_data("image/bmp", base64_bmp_string)
+  subroutine display_data(mime_type, data)
+    character(len=*), intent(in) :: mime_type, data
+    call lf_display_data(trim(mime_type)//c_null_char, trim(data)//c_null_char)
+  end subroutine
+
+end module lfortran_display
+)fortran";
```
certik, line 101: For now this is fine. Should we later maybe ship this with all the other modules that we ship with LFortran in our runtime library?

certik (approved):
I think this is fine to merge.

The display feature is currently untested --- for example if it doesn't produce an image, we will not know. We are currently only testing that the notebooks execute, which is good. But not enough.

PR #12206, opened 2026-07-13
fix: correct ES edit descriptor handling in handle_decimal

HarshitaKalani:


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -1418,50 +1426,80 @@ void handle_decimal(char* format, double val, int scale, char** result, char* c,
...
-            sprintf(exponent, "%+02d", exponent_value);
-        } else {
-            int exp_width = exp + 1;
-            if (exp_width > 10) exp_width = 10;
-            if (exp_width < 1) exp_width = 1;
-            int len = snprintf(exponent, sizeof(exponent), "%+0*d", exp_width, exponent_value);
-            if (len < 0 || len >= (int)sizeof(exponent)) {
-                goto overflow;
-            }
+        int abs_exp2 = (exponent_value < 0 ? -exponent_value : exponent_value);
```
certik, line 1464: I think there might be a bug here:

After this carry, the exponent constraints and layout derived above must be reapplied. For example, `(ES9.2E1)` with `9.999e9` changes the exponent from 9 to 10; the initial `E1` fit check passes, and this block then emits `+10` even though an explicit one-digit exponent must overflow the field. Likewise, `(ES0.2E0)` at the same boundary keeps the pre-carry auto width and ends up returning seven asterisks instead of growing to `1.00E+10`. A 99-to-100 carry without an explicit `Ee` also leaves `drop_e` false. Please recompute the explicit-width overflow check, auto width, and `drop_e` after incrementing `exponent_value`.

HarshitaKalani:
This is ready for review.

PR #12207, opened 2026-07-13
test: register integration test for issue #12203

### LINKED ISSUE #12203: Test : register test
```.f90

module m_151
  implicit none
  contains
  
  function method_create(f) result(that)
    procedure() :: f
    integer :: that
  end function

  subroutine compute(n)
    integer :: n
  end subroutine
end module

program derived_types_151
    use m_151
    implicit none
    integer :: ll
    ll = method_create(compute)
end program
```
Register test and verify it works properly, it's derived from test found at #11806, but simpler.

usefahmed07:
Closes #12203

Registers `derived_types_152.f90` as an integration test (derived from #11806, simplified per the issue). Verifies that procedure pointer arguments work correctly with `--implicit-interface` mode.

Verified locally with `python run_tests.py -t derived_types_152` — passes.


HarshilShah1804 (approved):
Looks good. Thanks @usefahmed07

certik (approved):

PR #12215, opened 2026-07-14
fix: ICE for list directed internal read from allocatable character array  

### LINKED ISSUE #12214: ICE for list-directed internal read from allocatable character array
## MRE

```fortran
program character_array_read_mre
  implicit none

  integer :: value
  character(len=4), allocatable :: lines(:)

  allocate(lines(1))
  lines = [" 42 "]

  read(lines, *) value

  if (value /= 42) error stop
  print *, "ok"
end program character_array_read_mre
```

## output

```text

Internal Compiler Error:
AssertFailed: is_proper_string_llvm_variable(str_type, str)
```

## gfortran behavior

```text

 ok
```

opixdown:
fixes #12214

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -17922,8 +17922,28 @@ public:
                 builder->CreateStore(
                     llvm::ConstantInt::get(llvm::Type::getInt64Ty(context), 0),
                     str_offset);
-                std::tie(str_src_data, str_src_len) = llvm_utils->get_string_length_data(
-                    ASRUtils::get_string_type(x.m_unit), unit_val);
+                ASR::ttype_t* unit_type = ASRUtils::expr_type(x.m_unit);
+                if (ASRUtils::is_array(unit_type)) {
+                    ASR::ttype_t* array_type = ASRUtils::type_get_past_allocatable_pointer(unit_type);
+                    if (ASRUtils::is_allocatable_or_pointer(unit_type)) {
+                        llvm::Type* llvm_array_type = llvm_utils->get_type_from_ttype_t_util(
+                            x.m_unit, array_type, module.get());
+                        unit_val = llvm_utils->CreateLoad2(llvm_array_type->getPointerTo(), unit_val);
+                    }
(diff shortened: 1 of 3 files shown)


HarshilShah1804 (approved):
Looks good.

HarshitaKalani (approved):

HarshitaKalani:
Please merge this once the CI passes.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -17922,8 +17922,28 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    }
+                    str_src_data = llvm_utils->get_stringArray_data(array_type, unit_val);
+                    llvm::Value* elem_len = llvm_utils->get_stringArray_length(array_type, unit_val);
+                    ASR::ttype_t* type32 = ASRUtils::TYPE(ASR::make_Integer_t(al, x.base.base.loc, 4));
+                    ASR::ArraySize_t* array_size = ASR::down_cast2<ASR::ArraySize_t>(
+                        ASR::make_ArraySize_t(al, x.base.base.loc, x.m_unit,
+                            nullptr, type32, nullptr));
+                    visit_ArraySize(*array_size);
+                    llvm::Value* n_elems = builder->CreateIntCast(tmp, llvm::Type::getInt64Ty(context), true);
+                    tmp = nullptr;
```
certik, line 17941: There might be a bug here:

Multiplying the element length by the element count exposes the array as one contiguous character stream, but each character-array element is a separate internal-file record. For example, `character(2) :: lines(2) = ["42", "43"]`; `read(lines, *) a, b` should produce `42` and `43`, while this passes `"4243"` to the runtime and lets the first conversion consume `4243`. Please preserve record boundaries (for example, by making the runtime aware of element length/count) and add a multi-record test without padding between records.

certik: This can be fixed in a subsequent PR, but please don't forget.

opixdown: i will fix these in subsequent PR

opixdown:
opened #12224 for the issue

PR #12217, opened 2026-07-14
[ci] Update Exhaustive-Checks-CI.yml to use latest versions of each Github actions (pins to hash of latest version)

mcbarton:
This PR updates the versions of every Github action in Exhaustive-Checks-CI.yml to the latest version. I used the full Git hash as i've been told its best to pin actions to a particular immutable version. I've ended each line with a hash followed by the version it is pinned to, to make it can at a quick glance be seen which version is being used.

I'll follow up this PR with PRs to update the other workflow files.

### DIFF
--- a/.github/workflows/Exhaustive-Checks-CI.yml
+++ b/.github/workflows/Exhaustive-Checks-CI.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: mamba-org/setup-micromamba@d7c9bd84e824b79d2af72a2d4196c7f4300d3476 # v3.0.0
         with:
           micromamba-version: '1.5.10-0'
-          environment-file: src/ci/environment_linux.yml
+          environment-file: ci/environment_linux.yml
           create-args: >-
             python=3.10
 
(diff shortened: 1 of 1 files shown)


.github/workflows/Exhaustive-Checks-CI.yml
```diff
@@ -36,19 +36,19 @@ jobs:
...
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
           path: src
 
-      - uses: mamba-org/setup-micromamba@v1.8.0
+      - uses: mamba-org/setup-micromamba@d7c9bd84e824b79d2af72a2d4196c7f4300d3476 # v3.0.0
         with:
           micromamba-version: '1.5.10-0'
-          environment-file: src/ci/environment_linux.yml
```
mcbarton, line 47: `src/ci/environment_linux.yml` doesn't exist on main, so unsure why this doesn't fail for others. Fixed as part of this PR.

PR #12222, opened 2026-07-14
fix: prevent ICE when non-variable expr passed to intent(inout/out) a…

### LINKED ISSUE #12204: Bug : `intent(inout)` introduces ICE while we should raise semantic error
```.f90
module m1
    implicit none
    type :: runner
        procedure(), nopass, pointer :: caller => null()
    contains
    end type
contains
    subroutine set_caller(this, caller)
        procedure() :: caller
        this%caller => caller
    end subroutine
end module

program p
    use m1
    implicit none
    type(runner) :: br
    call set_caller(upper_caller)
contains
    subroutine upper_caller(a)
        class(*), intent(in) :: a
    end subroutine
end program
```
Remove `intent(inout)` and we'll raise a proper semantic error. Keep it and we have ICE while we should keep raising the same error.

usefahmed07:
fix #12204
## Problem

Calling a procedure with a missing argument, where that argument would have matched an `intent(inout)`/`intent(out)` dummy, crashes the compiler with an internal compiler error (ICE / assertion failure) instead of reporting a normal semantic error.

Example:

```fortran
module caller_mod
    implicit none
    type :: caller_t
        integer :: id
    end type caller_t
contains
    subroutine set_caller(this, caller)
        type(caller_t), intent(inout) :: this
        type(caller_t), intent(in) :: caller
        this%id = caller%id
    end subroutine set_caller
end module caller_mod

program test
    use caller_mod
    type(caller_t) :: upper_caller
    call set_caller(upper_caller)   ! missing `caller` argument
end program test
```

Here `upper_caller` gets matched positionally to `this` (the `intent(inout)` dummy). Before checking whether the actual argument is even a valid variable, the compiler tries to validate that it's "modifiable" for the `intent(inout)` dummy, which crashes.

## Root cause

`is_modifiable_actual_argument_expr` in `asr_utils.h` assumed that any `ASR::Var_t` expression must resolve to a `Variable_t` symbol, and called `EXPR2VAR` (which does an unchecked `down_cast`) without first verifying that assumption:

```cpp
case ASR::exprType::Var: {
    ASR::Variable_t* variable_t = ASRUtils::EXPR2VAR(a_value);
    return variable_t->m_storage != ASR::storage_typeType::Parameter;
}
```

That's not always true - the symbol behind a `Var_t` can also be a `Function_t`/procedure symbol (as in the example above), and `down_cast` asserts when the actual type doesn't match, aborting the whole compiler.

## Fix

**Before:**
```cpp
case ASR::exprType::Var: {
    ASR::Variable_t* variable_t = ASRUtils::EXPR2VAR(a_value);
    return variable_t->m_storage != ASR::storage_typeType::Parameter;
}
```

**After:**
```cpp
case ASR::exprType::Var: {
    ASR::symbol_t* v_sym = ASRUtils::symbol_get_past_external(
        ASR::down_cast<ASR::Var_t>(a_value)->m_v);
    if (!ASR::is_a<ASR::Variable_t>(*v_sym)) {
        // Not a plain variable (e.g. a procedure/function symbol
        // passed positionally where a variable was expected) --
        // cannot be treated as a modifiable variable.
        return false;
    }
    ASR::Variable_t* variable_t = ASR::down_cast<ASR::Variable_t>(v_sym);
    return variable_t->m_storage != ASR::storage_typeType::Parameter;
}
```

Now the function safely returns `false` for non-variable symbols instead of crashing. This lets execution reach the existing "required argument is missing" validation later in `visit_SubroutineCall`, so the compiler reports the actual


tests/tests.toml
```diff
@@ -3956,6 +3956,12 @@ asr = true
 [[test]]
 filename = "errors/intent1.f90"
 asr = true
+[[test]]
+filename = "errors/intent_inout_nonvar_arg.f90"
+asr = true
+[[test]]
+filename = "errors/intent_inout_procedure_arg.f90"
+asr = true
```
HarshilShah1804: Instead of creating new test files, add the incorrect fortran code in any of the `tests/errors/continue_compilation*.f90` files and update references

HarshilShah1804 (dismissed):

tests/tests.toml
usefahmed07: Thanks @HarshilShah1804 to this point u are rigth

usefahmed07: Hey @HarshilShah1804, tried this adding just the type + subroutine to `continue_compilation_1.f90`'s module section triggers an unrelated SIGSEGV during symbol-table construction (in `visit_Subroutine`, `GenericProcedure` check), even before adding any call site. Confirmed a fresh checkout of the file passes clean, so it's something about my addition interacting with this specific file (which is already pretty dense with generic interfaces).

Given the risk of destabilizing a shared file I can't fully explain yet, I'd rather keep my two separate test files for now. Happy to dig into the root cause if you think it's worth it let me know.

HarshilShah1804: @usefahmed07 You can do something like this:
```diff
index 5f71d849b..5e3c227d9 100644
--- a/tests/errors/continue_compilation_3.f90
+++ b/tests/errors/continue_compilation_3.f90
@@ -75,10 +75,10 @@ program continue_compilation_3
 
     print *, iachar(s1)
     
-
-
-
-
+    type :: runner
+        procedure(), nopass, pointer :: caller => null()
+    contains
+    end type
```

Becuase the main error-triggering part is the procedure declaration. This change correctly captures the error.

And the second error test that you added, is already handled correctly without the changes. So, I don't think we require that test as a part of this PR. 

So, you can just add the code for the procedure pointer in `continue_compilation_3.f90` and update the test references.

usefahmed07: ok i will try it

HarshilShah1804: See https://github.com/lfortran/lfortran/issues/12204#issuecomment-4990913288. The MRE present in the issue is not throwing ICE now. So, this is probably fixed in some other PR. Let's wait for Assem's reply. Then we will see if we require the changes in the PR.

tests/reference/asr-continue_compilation_3-435a232.stderr
```diff
@@ -28,6 +28,12 @@ semantic error: Explicit shaped array with nonconstant bounds
 69 |     integer :: arr_nc(minval(abs(a)))
    |                       ^^^^^^^^^^^^^^ 
 
+semantic error: Procedure declarations without an explicit interface (procedure()) are not yet supported. Please use procedure(interface_name) or declare an interface block.
+  --> tests/errors/continue_compilation_3.f90:79:9
+   |
+79 |         procedure(), nopass, pointer :: caller => null()
+   |         ^^^^^^^^^^^ 
+
```
assem2002, line 36: that's not the targeted error.
I think you need to add a new test file and register it at `test.toml`, and make sure `implicit_interface` will be used.

We should see an error indicating that the number of arguments isn't matching (or any similar error message)

usefahmed07: ok i will add new test file

assem2002: let's use this 
```.f90
module m1
    implicit none
    type :: runner
        procedure(), nopass, pointer :: caller => null()
    contains
    end type
contains
    subroutine set_caller(this, caller)
        procedure() :: caller
        this%caller => caller
    end subroutine
end module

program p
    use m1
    implicit none
    type(runner) :: br
    call set_caller(upper_caller)
contains
    subroutine upper_caller(a)
        class(*), intent(in) :: a
    end subroutine
end program
```
that's the one with the issue, the current registered test actually works perfectly in main

src/libasr/asr_utils.h
```diff
@@ -1740,7 +1740,15 @@ static inline bool is_modifiable_actual_argument_expr(ASR::expr_t* a_value) {
...
         case ASR::exprType::Var: {
-            ASR::Variable_t* variable_t = ASRUtils::EXPR2VAR(a_value);
+            ASR::symbol_t* v_sym = ASRUtils::symbol_get_past_external(
+                ASR::down_cast<ASR::Var_t>(a_value)->m_v);
+            if (!ASR::is_a<ASR::Variable_t>(*v_sym)) {
+                // Not a plain variable (e.g. a procedure/function symbol
+                // passed positionally where a variable was expected) --
+                // cannot be treated as a modifiable variable.
+                return false;
+            }
```
assem2002, line 1747:
```suggestion
            if (!ASR::is_a<ASR::Variable_t>(*v_sym)) {
                return false;
            }
```

usefahmed07: Done

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
```diff
@@ -0,0 +1,5 @@
+semantic error: Non-variable expression in variable definition context (actual argument to INTENT = OUT/INOUT)
```
assem2002, line 1: We should see a type mismatch error.
Raising this error is still kinda valid, but it's quite misleading for the end user.
***
## What we want?
We want to prioritize the "type mismatch error" (i.e. we want to see a type mismatch error here, rather than a wrong-intent error)
 ***
## What to do?
- I opened an issue here #12312, Please try to fix it as you'll probably be aware of how to fix it.

assem2002 (approved):
LGTM.
Please try to fix the issue I mentioned.

usefahmed07:
> LGTM. Please try to fix the issue I mentioned.

ok i will work on this , thanks for marged!

tests/tests.toml
```diff
@@ -382,6 +382,10 @@ filename = "errors/implicit_call_02.f90"
 asr_implicit_interface = true
 asr_implicit_argument_casting = true
 
+[[test]]
+filename = "errors/intent_inout_missing_arg.f90"
+asr_implicit_interface = true
```
certik, line 387: This should not be a new file, use continue compilation.

@assem2002 or @usefahmed07 can you please fix it in the next PR?

usefahmed07: Yup no problem

usefahmed07:
Thanks @certik, tried moving it into continue_compilation_3.f90, but the `runner` type there is declared mid-body (not in the normal declaration section), and adding a contained subroutine that uses it surfaces unrelated cascading errors ("Derived type 'runner' not declared", duplicate procedure-interface errors) due to how that file's error-recovery/symbol-table ordering works same fragility me hit before with continue_compilation_1.f90. Given the risk of destabilizing a shared file, I'd like to keep this as a separate test file for now. Happy to revisit if there's a safer way to fold it in

tests/tests.toml
certik: https://github.com/lfortran/lfortran/issues/12339

certik:
I don't quite understand the problem, but if there are bugs in continue compilation then they must be fixed, those are real bugs. Here is the issue to fix: https://github.com/lfortran/lfortran/issues/12339.

assem2002:
The test requires implicit-interface flag, I think `continue_compilation_01` file doesn't get compiled with that flag ON.

PR #12232, opened 2026-07-15
Fix variable name collision in coarray pass (#12225)

### LINKED ISSUE #12225: Coarray lowering generated internal variables conflict with user defined variables
When compiling with `--coarray=true`, coarray lowering introduces an internal variable named `stat` which conflicts with user code defined variable with same name.

## MRE

```fortran
program main
    implicit none

    integer :: stat = 0

    print *, stat
end program
```

## Output
```
AssertFailed: scope.find(name) == scope.end()
```

Note: stat is just 1 of such variable names, there are other possible variable names too which need the same fix

usefahmed07:
Fixes #12225

The coarray lowering pass (`src/libasr/pass/coarray.cpp`) declares internal
temp variables into the same scope as user code — for example, a temp for
`prif_init`'s exit code. When a user variable happened to share the name
the pass wanted to use (e.g. `stat`, via `sync all(stat=stat)`), the pass
would try to insert a symbol with a name already present in the scope,
triggering: 

## Fix

Uniquify the internal temp's name via `get_unique_name` before declaring
it, so it no longer collides with an existing symbol in the same scope.
The user's `stat` keeps its name; the pass-generated temp becomes `stat1`
(or the next available uniquified name).

This fix is at the shared `declare_variable`-style call site used across
the pass, so it covers all coarray keyword arguments that go through this
path (`stat`, `errmsg`, `team`, `image_num`, etc.), not just `stat`.

## Before / After

**Before:** compiling the MRE below crashed with `AssertFailed:
scope.find(name) == scope.end()` during the coarray pass.

**After:** the pass now declares two distinct symbols in the same scope
— `stat` (the user's) and `stat1` (the pass-generated temp) — and each is
used for its correct purpose:

```fortran
program test_stat
    implicit none
    integer :: stat
    stat = 0
    sync all(stat=stat)
    print *, stat
end program test_stat
```

Confirmed via `--show-asr --pass=coarray` that:
- `stat1` is used for `prif_init`'s exit code.
- `stat` is correctly threaded through to `prif_sync_all`, matching the
  user's original `sync all(stat=stat)` call.

## Testing

- Added `tests/coarray_var_name_collision_01.f90`, a minimal reproduction
  of the collision.
- Ran the full coarray integration test suite (`coarray*.f90`,
  `coarrays_*.f90`) compile-only — no regressions.
- Ran the full test suite (`./run_tests.py`) — all tests pass.
- Regenerated reference files for `coarray_saved_01` and
  `coarray_initialization_01` - no diff, confirming this fix doesn't
  alter their expected output.

## Notes for reviewers

The original crash reproduces at the ASR/compile stage. Linking a full
coarray program additionally requires the Caffeine/PRIF runtime (built
separately in CI, not part of the standard build), so this was validated
with `-c` (compile-only) and `--show-asr`, which is sufficient to exercise
the fixed code path.

### DIFF
--- a/src/libasr/pass/coarray.cpp
+++ b/src/libasr/pass/coarray.cpp
@@ -51,8 +51,13 @@ class PRIFInterface {
                                         ASR::abiType abi, ASR::accessType access,
                                         ASR::presenceType presence, bool value_attr) {
             ASRUtils::ASRBuilder b(al, loc);
-            b.VariableDeclaration(symtab, name, type, intent, type_decl, abi, value_attr);
-            ASR::symbol_t *sym = symtab->get_symbol(name);
+            // Use a unique name to avoid clashing with user-defined variables
+            // that may already exist in `symtab` (e.g. a user variable
+            // literally named "stat"); internal compiler-generated names
+            // must never collide with source-level identifiers.
+            std::string unique_name = symtab->get_unique_name(name, false);
+            b.VariableDeclaration(symtab, unique_name, type, intent, type_decl, abi, value_attr);
+            ASR::symbol_t *sym = symtab->get_symbol(unique_name);
             LCOMPILERS_ASSERT(sym);
             ASR::Variable_t *var = ASR::down_cast<ASR::Variable_t>(sym);
             var->m_access = access;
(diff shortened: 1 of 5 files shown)


bonachea (approved):
LGTM!

Thanks for the detailed description of the problem and the fix!

PR #12233, opened 2026-07-15
Add regression test: derived type used before declaration (#7417)

### LINKED ISSUE #7417: Ambigous error message
```f90
program test
    type(t_pair) :: x

    type :: t_pair
      integer :: i
      real :: x
    end type
end program test
```

gives error

```
$ src/bin/lfortran test.f90
ASR verify pass error: The symbol table was found and the symbol in it shares the name, but is not equal to `sym`
 --> test.f90:3:5
  |
3 |     type(t_pair) :: x
  |     ^^^^^^^^^^^^^^^^^ failed here

ASR verify pass error: StructType::m_derived_type 't_pair' cannot point outside of its symbol table, owner: test
 --> test.f90:3:5
  |
3 |     type(t_pair) :: x
  |     ^^^^^^^^^^^^^^^^^ failed here


```

its fixed by moving type definition above variable definition
This should be caught earlier imo or the message needs to be different

usefahmed07:
Closes #7417

## Summary
Verified that the following reproduction from #7417 no longer crashes:
```fortran
program test
    type(t_pair) :: x
    type :: t_pair
      integer :: i
      real :: x
    end type
end program test
```
On current `main`, this now correctly produces a clear semantic error:

```
semantic error: Derived type `t_pair` is not defined
 --> test.f90:2:5
  |
2 |     type(t_pair) :: x
  |     ^^^^^^^^^^^^ Type used here is not defined in any scope
```

instead of the original `AssertFailed: scope.find(name) == scope.end()` crash from the ASR verify pass reported in the issue.

## Investigation
I traced this through the derived-type resolution logic in `determine_type` (`ast_common_visitor.h`). There's already a clear error path there for an unresolved derived type name when the declared variable isn't a pointer/allocatable:
```cpp
diag.add(Diagnostic(
    "Derived type `" + derived_type_name + "` is not defined",
    ...
));
throw SemanticAbort();
```
I couldn't find anything specifically fixing this case in recent history, so my best guess is this got resolved as a side effect of other semantic analysis / symbol resolution work done since the issue was opened. No source changes were needed for this PR — just adding a regression test so this stays fixed and doesn't silently regress.

## Testing
- Added `tests/errors/type_used_before_declared_01.f90`, reproducing the exact example from the issue.
- Ran the full test suite (`./run_tests.py`) - all tests pass.


usefahmed07:
@bonachea

HarshilShah1804:
@usefahmed07 The branch has a conflict in `tests/tests.toml` file. Please resolve the conflict. Then we can review the PR. Also, instead of creating a new tests, add the error-throwing code in any of the `tests/errors/continue_compilation_*.f90` files.

tests/tests.toml
```diff
@@ -5258,7 +5258,6 @@ fortran = true
 pass = "coarray"
 options = "--coarray=true"
 
-
```
HarshilShah1804, line 5261: Please remove this line deletion change.

HarshilShah1804 (approved):
Looks good. Just remove that line deletion change and the unused files `asr-type_used_before_declaraed_01...`. Then we can merge this.

tests/tests.toml
usefahmed07: ok

Jatinagarwal24 (approved):

PR #12234, opened 2026-07-16
Add scope in do concurrent using block

HiteshGhanchi:
This change adds an owned symbol table to the ASR DoConcurrentLoop node.

A DO CONCURRENT construct can introduce variables whose scope is limited to the construct itself. The most direct example is a loop index declared in the concurrent header:

``` fortran
do concurrent (integer :: i = 1:3)
    values(i) = i * i
end do
```
The i declared in this header is not the same variable as an i declared in an enclosing procedure or program. It is a construct-local variable whose lifetime and visibility are limited to the DO CONCURRENT construct.

Previously, DoConcurrentLoop was represented only as a statement containing loop heads, locality specifications, reductions, and a body. It did not own a symbol table. As a result, construct-local declarations could not be represented with the correct lexical scope. They could incorrectly resolve to an enclosing variable and could overwrite or otherwise interfere with that variable during lowering or code generation.

The ASR representation now gives each DoConcurrentLoop its own symbol table. This brings the ASR representation in line with the Fortran scoping rules and allows the compiler to distinguish construct-local entities from variables declared in surrounding scopes.

Minimum Reproducible Error:

``` fortran
program concurrent_scope_test
    implicit none

    integer :: i
    integer :: values(3)

    i = 10

    do concurrent (integer :: i = 1:3)
        values(i) = i * i
    end do

    if (any(values /= [1, 4, 9])) error stop 1
    print *, i
end program concurrent_scope_test
```

Here, the current version shows the output as `3`, but the output should be `10`.
The PR makes sure that it happens by introducing symbol table to `do_concurrent` even though it is a `statement` and not a `symbol`.

TO DO:
- [ ] Add new tests for all the changes.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -718,7 +718,7 @@ RUN(NAME types_06 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc cpp wasm fortran
 
 # GFortran + LFortran C++
 RUN(NAME doconcurrentloop_01 LABELS gfortran cpp)
-RUN(NAME doconcurrentloop_03 LABELS llvm llvm_wasm llvm_wasm_emcc fortran) # type-spec in concurrent-header is not yet implemented by gfortran.
+RUN(NAME doconcurrentloop_03 LABELS llvm llvm_wasm llvm_wasm_emcc) # # GFortran does not yet accept a type declaration in a DO CONCURRENT control header, including through the Fortran backend.
 
 RUN(NAME arrays_01 LABELS gfortran cpp llvm llvmStackArray wasm)
 RUN(NAME arrays_01_size LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
@@ -4224,6 +4224,7 @@ RUN(NAME do_concurrent_12 LABELS llvm_omp llvm NO_FAST_MATH) # to not include is
 RUN(NAME do_concurrent_13 LABELS llvm_omp llvm) # to not include is that we do a `omp_set_num_threads(xx)` call
 RUN(NAME do_concurrent_14 LABELS llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME do_concurrent_15 LABELS llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME do_concurrent_16 LABELS llvm llvm_wasm llvm_wasm_emcc)
 
 
 RUN(NAME transfer_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/do_concurrent_16.f90
+++ b/integration_tests/do_concurrent_16.f90
@@ -1,0 +1,16 @@
+program do_concurrent_16
+    implicit none
+
+    integer :: i
+    integer :: values(3)
+
+    i = 10
+    values = 0
+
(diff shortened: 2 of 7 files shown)


HiteshGhanchi:
@certik @HarshitaKalani  
The LFortran CI (OS=ubuntu-latest, LLVM=11) failure occurs in the Fortran-backend integration tests. LFortran generates the following standard-conforming Fortran 2018 code:
``` fortran
do concurrent (integer(4) :: i = 1:verbose_levels)
```
The optional integer-type-spec :: in a DO CONCURRENT header is permitted by Fortran 2018 rule R1125. However, gfortran rejects this syntax. I reproduced the failure with both GFortran 13.3 in CI and GFortran 15.2 locally. LLVM Flang accepts the same generated source and the executable runs successfully.
Therefore, this failure is caused by missing support in the downstream GFortran compiler, rather than invalid Fortran generated by LFortran or a failure in the new DoConcurrentLoop symbol-table implementation. This test should not currently be validated through the GFortran-backed Fortran integration configuration. It could instead be skipped for that configuration or tested using Flang.

MRE:
``` fortran
program doconcurrentloop_03
implicit none
integer(4) :: verbose_levels
integer(4) :: verbose_sum
verbose_levels = 5
verbose_sum = 0
do concurrent ( integer(4) :: i = 1: verbose_levels )
    verbose_sum = verbose_sum + i
end do
print *, verbose_sum
if (verbose_sum /= 15) then
    error stop
end if
end program doconcurrentloop_03
```
Output is `15` in flang and in my lfortran build after my changes, but gfortran doesn't allow `integer` in `do concurrent`

HiteshGhanchi:
I also plan to add the same scoping support for `do_implied` loops. Before doing that, I would like to know whether you are onboard with allowing `statement nodes` to own `symbol tables`. I could not find any documentation or design guidance indicating that symbol tables should be restricted to symbol nodes.

src/libasr/ASR.asdl
```diff
@@ -41,7 +41,7 @@ stmt
     | Cycle(identifier? stmt_name)
     | ExplicitDeallocate(expr* vars)
     | ImplicitDeallocate(expr* vars)
-    | DoConcurrentLoop(do_loop_head* head, expr* shared, expr* local, reduction_expr* reduction, stmt* body)
+    | DoConcurrentLoop(symbol_table symtab, do_loop_head* head, expr* shared, expr* local, reduction_expr* reduction, stmt* body)
```
certik: We only allow to have a symbol table for Symbols. One workaround is to use Block inside `body`. The other option is to create a dedicated node that will be referenced from here.

certik:
Good catch and thanks for the fix. Probably the best way for now to unblock us is to generate a Block inside the `body`, which can have a local declaration. We can only do the block if we have to. Another option is to always do the block and enforce it in verify() that block is always present inside do concurrent.

certik:
For the CI failure, you'll have to disable the gfortran test for this case somehow.

src/libasr/ASR.asdl
HiteshGhanchi: Ok, I removed the symbol table and added the Block inside the do_concurrent instead.
Changed the fortran codegen accordingly too.

HiteshGhanchi:
@certik I think now it can be safely merged.
If there is anything more which needs to be added, do inform me.
Thanks

HiteshGhanchi:
@certik any more changed you need ?

PR #12235, opened 2026-07-16
Fix multi-image constant values in `iso_fortran_env`

bonachea:
There were several problems with the coarray-relevant constant values chosen in `iso_fortran_env` that would cause integration problems with PRIF/Caffeine in the near future: 

* Caffeine requires `atomic_{int,logical}_kind` to be 64-bit
* `STAT_*` values may not be zero
* `STAT_FAILED_IMAGES` should have a negative value while the compiler does not support failed image detection

Resolve those problems, and change the Caffeine build in test_caffeine.sh to import the repaired constant values.

Towards #12199


certik (approved):
Looks good to me, thanks!

PR #12248, opened 2026-07-16
Fix ICE when include file is not found (#12156)

### LINKED ISSUE #12156: Missing `include` file causes an Internal Compiler Error (unhandled LCompilersException) instead of a normal error diagnostic
## Description

When a Fortran source `include`s a file that cannot be located, LFortran aborts with an **Internal Compiler Error / Unhandled exception** and prints a `Traceback`, rather than emitting a normal user-facing error diagnostic (as e.g. gfortran does). A missing include is ordinary user error, not a compiler bug, so it should be reported gracefully.

This surfaces very easily on real codebases that rely on `include` files: in a 152-file third-party corpus, 17 files failed to parse — **all 17 for this single reason** (a `.fi` include not on the search path).

Same class as #8434 ("should give an error message, not an ICE").

## Minimal reproducer

`missing_include.f90`:

```fortran
program p
  implicit none
  include 'does_not_exist.fi'
  print *, "hello"
end program p
```

```console
$ lfortran missing_include.f90
  Binary file ".../bin/lfortran", local address: 0x...
  ... (5 frames) ...
LCompilersException: Include file 'does_not_exist.fi' not found. If an include path is available, please use the `-I` option to specify it.
```

Also reproduces with `--show-ast` / `--show-ast --json` (the path I hit it on).

## Expected behavior

A clean error diagnostic pointing at the `include` line, e.g.:

```
error: Include file 'does_not_exist.fi' not found (use -I to add a search path)
 --> missing_include.f90:3:3
```

and a **non-zero exit code**.

## Actual behavior

- "Internal Compiler Error: Unhandled exception" + traceback.
- The underlying `LCompilersException` message is actually informative — it just isn't caught and rendered as a normal diagnostic.
- **Exit code is `0`** despite the failure (observed on macOS), so callers cannot detect the error programmatically. (This exit-code point may be a separate bug, but noting it here.)

## Environment

```
LFortran version: 0.63.0   (latest conda-forge release)
Status: alpha
Platform: macOS ARM (arm64-apple-darwin)
LLVM: 19.1.1
```

Installed via `conda create -n lfortran -c conda-forge lfortran`.

usefahmed07:
Closes #12156

## Summary
Fixes an Internal Compiler Error that occurs when a Fortran `include` statement references a file that can't be found. Previously this crashed with a full stack trace instead of producing a normal diagnostic.

## Before
```fortran
program p
  implicit none
  include 'does_not_exist.fi'
  print *, "hello"
end program p
```
```
$ lfortran missing_include.f90
  Binary file ".../bin/lfortran", ...
  ... (several frames) ...
LCompilersException: Include file 'does_not_exist.fi' not found. If an include path is available, please use the `-I` option to specify it.
```

## After
```
$ lfortran missing_include.f90
syntax error: Include file 'does_not_exist.fi' not found. If an include path is available, please use the `-I` option to specify it.
 --> missing_include.f90:3:30
  |
3 |   include 'does_not_exist.fi'
  |                              ^

EXIT CODE: 1
```

## Root cause
`process_include()` in `parser.cpp` threw a bare `LCompilersException` when a file wasn't found. `prescan()` (which calls it) had no `diagnostics` parameter and returned a plain `std::string`, so there was no way to report the error as a normal diagnostic — it could only escape as a raw exception, which the top-level handler in `lfortran.cpp` always treats as a genuine internal compiler bug.

## Fix
- Changed `prescan()`'s signature to accept `diag::Diagnostics &diagnostics` and return `Result<std::string>`, matching the existing pattern used by `parse()`.
- Updated `process_include()` to populate `diagnostics` with a proper `diag::Diagnostic` (`Stage::Parser`) and throw the existing lightweight `parser_local::ParserAbort` marker exception instead of `LCompilersException`, mirroring the pattern already used elsewhere in `parser.cpp` (e.g. `fix_program_without_program_line`).
- Wrapped `prescan()`'s body in a `try`/`catch` for `ParserAbort`, converting it into a clean `Error()` result.
- Updated all call sites (`fortran_evaluator.cpp`, `lfortran.cpp`'s `emit_prescan`/`emit_tokens`, and the recursive `prescan()` call inside `process_include()` itself for nested includes) to pass `diagnostics` through and handle the new `Result<std::string>` return type.

## Testing
- Added `tests/errors/missing_include_01.f90` reproducing the original MRE from the issue.
- Ran the full test suite (`./run_tests.py`) — all tests pass, no regressions from the `prescan()` signature change.

### DIFF
--- a/src/lfortran/parser/parser.cpp
+++ b/src/lfortran/parser/parser.cpp
@@ -640,7 +640,7 @@ bool check_newlines(const std::string &s, const std::vector<uint32_t> &newlines)
     return true;
 }
 
-void process_include(std::string& out, const std::string& s,
+bool process_include(std::string& out, const std::string& s,
                      LocationManager& lm, size_t& pos, bool fixed_form,
                      std::vector<std::filesystem::path> &include_dirs,
                      int &col, diag::Diagnostics &diagnostics)
@@ -673,7 +673,7 @@ void process_include(std::string& out, const std::string& s,
             + "' not found. If an include path "
             "is available, please use the `-I` option to specify it.",
             diag::Level::Error, diag::Stage::Parser, {diag::Label("", {loc})}));
-        throw parser_local::ParserAbort();
+        return false;
     }
 
     LocationManager lm_tmp;
@@ -687,7 +687,7 @@ void process_include(std::string& out, const std::string& s,
     if (include_res.ok) {
         include = include_res.result;
     } else {
-        throw parser_local::ParserAbort();
+        return false;
     }
 
     // Possible it goes here
@@ -696,6 +696,7 @@ void process_include(std::string& out, const std::string& s,
     while (pos < s.size() && s[pos] != '\n') pos++;
     lm.files.back().out_start.push_back(out.size());
     lm.files.back().in_start.push_back(pos);
+    return true;
 }
 
 bool is_include(const std::string &s, uint32_t pos) {
@@ -727,7 +728,6 @@ Result<std::string> prescan(const std::string &s, LocationManager &lm,
         bool fixed_form, std::vector<std::filesystem::path> &include_dirs,
         diag::Diagnostics &diagnostics)
 {
-  try {
     if (fixed_form) {
         // `pos` is the position in the original code `s`
         // `out` is the final code (outcome)
@@ -817,8 +817,11 @@ Result<std::string> prescan(const std::string &s, LocationManager &lm,
                     pos += 7;
                     while (pos < s.size() && s[pos] == ' ') pos++;
                     if ((s[pos] == '"') || (s[pos] == '\'')) {
-                        process_include(out, s, lm, pos, fixed_form,
-                            include_dirs, col, diagnostics);
+                        if (!process_include(out, s, lm, pos, fixed_form,
+                                include_dirs, col, diagnostics)) {
+                            Error error;
+                            return error;
+                        }
                     }
                     break;
                 }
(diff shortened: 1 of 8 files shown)


usefahmed07:
@bonachea Can You Review it

bonachea:
> @bonachea Can You Review it

It looks right to me, but this is outside my area of expertise so someone else should take a look.

usefahmed07:
> > @bonachea Can You Review it
> 
> It looks right to me, but this is outside my area of expertise so someone else should take a look.

Ok no problem

src/lfortran/parser/parser.cpp
```diff
@@ -665,9 +665,15 @@ void process_include(std::string& out, const std::string& s,
...
+        Location loc;
+        loc.first = pos;
+        loc.last = pos;
+        diagnostics.add(diag::Diagnostic(
+            "Include file '" + include_filename
             + "' not found. If an include path "
-            "is available, please use the `-I` option to specify it.");
+            "is available, please use the `-I` option to specify it.",
+            diag::Level::Error, diag::Stage::Parser, {diag::Label("", {loc})}));
+        throw parser_local::ParserAbort();
```
certik: I think we should not be using exceptions, just propagate the error state via Result in the call chain.

src/lfortran/parser/parser.cpp
```diff
@@ -886,6 +900,10 @@ std::string prescan(const std::string &s, LocationManager &lm,
         lm.files.back().out_start.push_back(out.size());
         return out;
     }
+  } catch (const parser_local::ParserAbort &) {
+    Error error;
+    return error;
+  }
```
certik: Instead of try/catch, propagate the error state via Result

certik (commented):
I think this is in the right direction, thanks for fixing it.

Let's not introduce any more exceptions, but rather use the Result error handling. We should eventually get rid of all exceptions, but for now at least do not introduce new ones.

src/lfortran/parser/parser.cpp
usefahmed07: ok i will make it

certik (approved):
I think this looks good now.

usefahmed07:
> I think this looks good now.

Thanks

PR #12249, opened 2026-07-17
Tests: Rename Fortran output file for `--show-fortran` reference tests to use .f90 suffix

bonachea:
The `test/reference/*.stdout` files resulting from `--show-fortran` compilation contain only Fortran code. Rename these to `*.f90` so we automatically activate syntax highlighting in the GitHub UI and many editors, improving human readability.


HarshitaKalani (approved):
Looks good to me. Thanks!

src/libasr/compiler_tester/tester.py
```diff
@@ -199,7 +199,11 @@ def run(basename: str, cmd: Union[pathlib.Path, str],
     if not os.path.exists(outfile):
         outfile = None
     if len(r.stdout):
-        stdout_file = os.path.join(out_dir, basename + "." + "stdout")
+        if "--show-fortran" in cmd:
+            stdout_ext = ".f90"
+        else:
+            stdout_ext = ".stdout"
+        stdout_file = os.path.join(out_dir, basename + stdout_ext)
```
certik, line 206: Why not accepting the stdout_ext as an argument to this function? In this function we do not care about what we are testing, and also the check `"--show-fortran" in cmd` seems fragile. I would propagate this option all the way up.

certik (approved):
I think this is fine, but the implementation can be improved. Can be done in a subsequent PR.

PR #12265, opened 2026-07-18
fix : false impure procedure error when a PURE procedure uses an imported array bound

### LINKED ISSUE #12264: Failure to compile a program that uses an imported integer parameter as an array size in a pure procedure
This program does not compile with lfortran 0.64, I *think* it did with 0.63 but I cannot say for sure. Godbolt only has up to 0.59 and it builds there! I think this is legal. 

```fortran 
module const_mod
   implicit none
   private
   public :: NZ
   integer, parameter :: NZ = 128        ! plain compile-time named constant
end module const_mod

module kern_mod
   use const_mod, only: NZ               ! <-- imported (this is what breaks it)
   implicit none
   private
   public :: caller
contains
   pure subroutine callee(h, out)
      real, intent(in)  :: h(NZ)         ! explicit-shape dummy sized by imported param
      real, intent(out) :: out
      out = h(1)
   end subroutine callee

   pure subroutine caller(res)
      real, intent(out) :: res
      real :: buf(NZ)
      buf = 1.0
      call callee(buf, res)              ! <-- LFortran emits __lcompilers_get_nz() here
   end subroutine caller
end module kern_mod
```

Jatinagarwal24:
fixes #12264


certik:
Can you explain how the fix works --- what was wrong before?

Is the issue somehow that we made a procedure impure if it imported a bound that was a constant?

And is it still pure, if the bound can change? I think it is. But it's not deterministic then. But you set it to deterministic also in this PR?

Jatinagarwal24:
> Can you explain how the fix works --- what was wrong before?
> 
> Is the issue somehow that we made a procedure impure if it imported a bound that was a constant?
> 
> And is it still pure, if the bound can change? I think it is. But it's not deterministic then. But you set it to deterministic also in this PR?

So the issue was that  the `getter (__lcompilers_get_nz) ` was created with `m_pure=true` but `m_side_effect_free=false,` so the PURE check (which keys off `m_side_effect_free`) rejected the call. Reading a module variable is fine inside a PURE procedure, so I `set m_side_effect_free=true`.

Thanks for pointing out that it's not deterministic  , if the bound is a mutable module variable it can change between calls , so I've left `m_deterministic=false` and only flipped the side-effect-free flag

PR #12266, opened 2026-07-18
fix : ICE and segfault on class arrays passed to class array dummies

### LINKED ISSUE #12253: ICE for reshape of allocatable polymorphic array passed to class dummy
## MRE

```fortran
program main
  implicit none

  type :: t
    integer :: i
  end type t

  class(t), allocatable :: a(:)

  allocate(a(1), source=t(1))

  call check(reshape(a, [1]))

contains

  subroutine check(arg)
    class(t) :: arg(:)

    print *, arg%i
  end subroutine check

end program main
```

## output

```text

LCOMPILERS_ASSERT failed:
CreateGEP2: Type mismatch

Pointer pointee type:
%array.1 = type { %t.3_class*, ... }

Type parameter:
%t.3_class = type <{ i32 (...)**, %t.3* }>

```

## gfortran behavior

```text

           1
```

Jatinagarwal24:
fixes #12253
Passing a polymorphic array to a `class(t) :: arg(:)` dummy hit an
assert in CreateGEP2, and `reshape()` of such an array segfaulted at
runtime. Two independent bugs:

1. declare_vars() guarded the scalar-class path with
   `is_class_type(extract_type(v->m_type))`. extract_type() strips Array
   along with Allocatable/Pointer, so a class *array* passed the guard
   and indexed field 1 of what is actually an array descriptor
   (%array.1), not a class wrapper (%t_class) -- tripping the pointee
   type assert. Fixed by also requiring !is_array(). The same faulty
   predicate in allocate_array_members_of_struct() is fixed alongside it.

2. reshape()'s element-wise deep-copy path was written for plain derived
   types: it memsets the destination buffer, which is correct for
   allocatable components but leaves a class wrapper's inner struct
   pointer null, and deepcopy() then copies through that null pointer.
   Class arrays also use a one-wrapper layout -- a single {vptr,
   data_ptr} addressing a contiguous element buffer -- not one wrapper
   per element, so the element-wise loop had the wrong memory model
   entirely. Class arrays now get their own copy: propagate the vptr,
   heap-allocate the destination element buffer (the finalizer frees it,
   so stack memory would double-free), and copy elements through the
   vtable copy function. Dimension descriptors are left untouched so
   reshape's own shape setup still applies, which keeps rank-changing
   reshape working. Unlimited-polymorphic arrays have a different layout
   again and are left to the existing paths.

gfortran segfaults on the reshape case and flang-new 19 does not, so
arrays_reshape_45 is labelled llvm-only.

Tests: integration_tests/class_150.f90 (class array dummy, incl. an
extended dynamic type) and integration_tests/arrays_reshape_45.f90
(reshape of a polymorphic array, incl. rank-changing reshape to [2,2]).


YashNagda17:
I have just pushed one simplification reusing an existing helper. Please review this part once. Rest looks fine 

CC: @Jatinagarwal24

Jatinagarwal24:
> I have just pushed one simplification reusing an existing helper. Please review this part once. Rest looks fine
> 
> CC: @Jatinagarwal24

Thanks for the review and the simplification . Looks good to me.

YashNagda17 (approved):

PR #12268, opened 2026-07-18
fix: support VOLATILE attribute statement in semantic analyzer

### LINKED ISSUE #12267: Volatile is not implemented
MRE:

```fortran 
program volatile_stmt
   real :: x
   volatile :: x     ! VOLATILE attribute statement (F2008 R560 / F2018 C868)
   x = 1.0
   print *, x
end program volatile_stmt
```

Output:

```
   --> mre.f90:3:4
    |
  3 |    volatile :: x     ! VOLATILE attribute statement
    |    ^^^^^^^^^^^^^
```

LFortran 0.64 installed via conda, macos arm

Jatinagarwal24:
fixes #12267

AttrVolatile was produced by the parser but never consumed in
ast_common_visitor.h, so `volatile :: x` fell through to a generic
"Attribute declaration not supported" error. Handled as a no-op,
mirroring AttrAsynchronous. The inline `type, volatile ::` form
already worked and was already covered by volatile_01..03; the
statement form had no coverage, hence the gap. Adds volatile_04.


JorgeG94:
amazing, thanks so much for this. Very excited

HarshitaKalani (approved):

PR #12269, opened 2026-07-18
Fix parser error for coindexed derived-type component reference (#12262)

### LINKED ISSUE #12262: Parser error for coindexed derived type component reference
## MRE
```
program test
    implicit none

    type :: t
        integer :: val
    end type

    type(t) :: x[*]
    x%val = this_image()

    if (this_image() == 1) then
        print *, x[1]%val
    end if
end program
```

## LFortran's Output
```
(lf) jainam@jainam-ThinkBook-15-G5-ABP:~/Documents/gsoc/lfortran$ ./src/bin/lfortran tests/1a.f90
syntax error: Token '%' is unexpected here
  --> tests/1a.f90:12:22
   |
12 |         print *, x[1]%val
   |                      ^ 


```

usefahmed07:
Fixes #12262

`x[1]%val` (a coindexed reference followed by derived-type member access) previously failed to parse with "syntax error: Token '%' is unexpected here", since struct_member had no way to carry coarray indices.

## Before
    type(t) :: x[*]
    print *, x[1]%val

    syntax error: Token '%' is unexpected here
      --> tests/1a.f90:12:22
       |
    12 |         print *, x[1]%val
       |                      ^

## After
    type(t) :: x[*]
    print *, x[1]%val

    1

(parses and runs correctly - in single-image mode the coindex is validated then resolved to the base variable, same convention as plain `x[1]`)

## Changes
- grammar/AST.asdl: add coarrayarg*/keyword* fields to struct_member
- parser.yy: new production `id "[" coarray_arg_list "]" "%"`
- semantics.h: STRUCT_MEMBER3 macro for the new production
- ast_common_visitor.h: visit_NameUtil validates the coindex (is-coarray check, corank match) then resolves the base variable, consistent with visit_CoarrayRef's existing single-image behavior of ignoring cosubscripts after validation.

## Scope
Covers a single coindexed member followed by a trailing member/id (the reported case). A coindex in the middle of a longer chain (e.g. a%x[1]%y%z) is explicitly rejected with a clear "not yet supported" diagnostic rather than being silently mishandled.

## Testing
Added regression test: tests/coarray_member_access_01.f90


usefahmed07:
@bonachea Can u review it

tests/reference/fortran-coarray_member_access_01-3e8c480.stdout
bonachea, line 1: This file is extraneous and should be deleted.

tests/reference/fortran-coarray_member_access_01-3e8c480.f90
```diff
@@ -0,0 +1,88 @@
...
+type(c_ptr) :: x__coarray_data
+type(prif_coarray_handle) :: x__coarray_handle
+call __module_prif_prif_init(stat)
+call __module_prif_prif_allocate_coarray([1_8], [integer(8) :: ], 4_8, null(), x__coarray_handle, x__coarray_data)
+call c_f_pointer(x__coarray_data, x)
+call __module_prif_prif_sync_all()
+x%val = lcompilers_prif_this_image()
+if (lcompilers_prif_this_image() == 1) then
+    print *, x%val
+end if
```
bonachea: This is a low-fidelity translation of the original code:
```
    if (this_image() == 1) then
        print *, x[1]%val
    end if
```
It might be reasonable for single-image mode, but this was the output for multi-image compilation (`--coarray=true`).

In multi-image mode this transformation only happens to work in this specific case because of the conditional.

In general (with multiple images), the expression `x[1]%val` requires emitting a `prif_get*` call.  

The transformed expression`x%val` is only a valid replacement in multi-image mode when the compiler analysis can prove that line will only be executed by image 1 of the current team (which is difficult and unlikely in non-trivial cases).

bonachea (changes requested):
I generally support work to generalize LFortran's coarray support.

Unfortunately I have serious concerns about the current approach in this PR, which IIUC is to unconditionally hard-code assumptions that only hold in single-image mode (erasing of coindexing) that would produce silently incorrect behavior in multi-image mode.

This approach could _maybe_ proceed if changes are made to ensure it's only ever applied in single-image mode. However IMHO it would be better to pursue a real multi-image solution for component access to derived type coarrays.

CC: @Jainam-not-a-robot

usefahmed07:
> I generally support work to generalize LFortran's coarray support.
> 
> Unfortunately I have serious concerns about the current approach in this PR, which IIUC is to unconditionally hard-code assumptions that only hold in single-image mode (erasing of coindexing) that would produce silently incorrect behavior in multi-image mode.
> 
> This approach could _maybe_ proceed if changes are made to ensure it's only ever applied in single-image mode. However IMHO it would be better to pursue a real multi-image solution for component access to derived type coarrays.

No problem, I know you don't want to take the risk Thank you for your honesty.

tests/reference/fortran-coarray_member_access_01-3e8c480.stdout
usefahmed07: > This file is extraneous and should be deleted.
ok

PR #12270, opened 2026-07-18
feat : add create-mre, fix-mre and repro-issue agent skills

Jatinagarwal24:
Add three agent skills, adapted from certik/ai_skills, that automate the reproduce -> reduce -> fix loop for LFortran bugs:

  repro-issue  turn a GitHub issue into a faithful Reproducible Example
  create-mre   reduce an MRE or third-party failure to a minimal MRE
  fix-mre      fix the compiler bug and add an integration test

The skills chain into a pipeline that can be run in a loop to bring a third-party Fortran package up on LFortran: build, take the first failure, reduce it, fix it, rebuild, repeat.

Skills live in .agents/skills/, the vendor-neutral location read by Codex, Copilot and Grok. Claude Code only discovers .claude/skills/, so .claude/skills is a symlink to ../.agents/skills, mirroring the existing CLAUDE.md -> AGENTS.md symlink. The skills are stored once and every agent picks up edits.

Adapted from the originals for this repository:

- Replaced hardcoded personal conda/pixi paths with the build and test commands AGENTS.md documents.
- Made the reference compiler configurable and defaulted it to gfortran, matching the gfortran integration-test label. Fixed an undefined $GFORTRAN reference in the create-mre run.sh template.
- fix-mre now reports instead of committing, and verifies the new test fails on main and passes on the branch, per the AGENTS.md requirement.
- Pointed diagnosis at the AGENTS.md architecture rules, notably that a type workaround needed in codegen means the bug is upstream in semantics.

Reproducers are written to the repo root by convention and are gitignored; the committed deliverable is always the integration test.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,50 @@ detailed reference how to contribute to the project.
 - `tests/`, `integration_tests/`: unit/E2E suites
 - `doc/`: docs & manpages (site generated from here)
 - `examples/`, `grammar/`, `cmake/`, `ci/`, `share/`: supporting assets
+- `.agents/skills/`: agent skills (see "Agent Skills" below)
+
+## Agent Skills
+
+Reusable, agent-invocable procedures live in `.agents/skills/<name>/SKILL.md`,
+following the [Agent Skills](https://agentskills.io/) open standard. Agents load
+a skill on demand when the task matches its `description`.
+
+`.agents/skills/` is the vendor-neutral location read by Codex, Copilot, and
+Grok. Claude Code only discovers `.claude/skills/`, so `.claude/skills` is a
+symlink to `../.agents/skills` — the same approach as the `CLAUDE.md` →
+`AGENTS.md` symlink. **Skills are stored once, in `.agents/skills/`;** edit them
+there and every agent picks up the change.
+
+Available skills:
+
+| Skill | Purpose |
+| --- | --- |
+| `repro-issue` | Turn a GitHub issue into a faithful Reproducible Example (RE) |
+| `create-mre` | Reduce an RE or third-party failure to a Minimal Reproducible Example (MRE) |
+| `fix-mre` | Fix the compiler bug behind an MRE and add an integration test |
+
+### The reproduce → reduce → fix loop
+
+The three skills chain into a pipeline, each consuming the previous one's output:
+
+```
+repro-issue  ──►  create-mre  ──►  fix-mre
+ (RE: faithful)    (MRE: minimal)   (fix + integration test)
+```
+
+This is designed to be run **in a loop** to bring a third-party Fortran package
+up on LFortran: build the package, take the first failure, reduce it, fix it,
+then rebuild and repeat until the package compiles and its tests pass. Each
+iteration should produce one focused PR — `AGENTS.md`'s "one bug = one MRE =
+one PR" rule applies to every pass through the loop.
+
+Reproducers are written to the repository root by convention (`run.sh`,
+`mre_*.f90`, `re_*.f90`) and are gitignored — they are scratch inputs to
+`fix-mre`. The committed deliverable is always the integration test.
+
+Skills assume `build/src/bin` is first on `PATH` (so `lfortran` is the in-tree
+build) and that a reference compiler — `gfortran`, matching the `gfortran`
+integration-test label — is available for differential testing.
 
 ## Prerequisites
 - Tools: CMake (>=3.10), Ninja, Git, Python (>=3.8), GCC/Clang/MSVC.
--- a/check_binary_file_in_git_history.py
+++ b/check_binary_file_in_git_history.py
@@ -5,10 +5,12 @@ import subprocess
 # If the file is binary it'd conflict with
 # text mode 'r', leading to raised error `UnicodeDecodeError`.
 def is_file_binary(file_path: str) -> bool:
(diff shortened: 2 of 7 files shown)


Jatinagarwal24:
Hi @certik pls review this.

certik:
Also the binary check should be improved:
```
  File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 74, in <module>
number of commits from current PR branch against origin/main is : 2
    run_test()
  File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 53, in run_test
    if (is_file_binary(file_path)):
        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 8, in is_file_binary
    with open(file_path, 'r') as fp:
         ^^^^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: '.claude/skills'
a3f294e1f2c57073d57ca2cca2b4d15340a689e1
0661c275141db12422c22a3cd20a327751aadf2f

checked out to commit -->  a3f294e1f2c57073d57ca2cca2b4d15340a689e1

```

certik:
Adding a check in https://github.com/lfortran/lfortran/pull/12271, so in the future this will fail.

AGENTS.md
```diff
@@ -12,6 +12,50 @@ detailed reference how to contribute to the project.
...
+iteration should produce one focused PR — `AGENTS.md`'s "one bug = one MRE =
+one PR" rule applies to every pass through the loop.
+
+Reproducers are written to the repository root by convention (`run.sh`,
+`mre_*.f90`, `re_*.f90`) and are gitignored — they are scratch inputs to
+`fix-mre`. The committed deliverable is always the integration test.
+
+Skills assume `build/src/bin` is first on `PATH` (so `lfortran` is the in-tree
+build) and that a reference compiler — `gfortran`, matching the `gfortran`
+integration-test label — is available for differential testing.
```
certik, line 58: The skills also assume flang, and also they are usable in a standalone setting, not just in a loop. We can iteratively improve the documentation later.

certik (approved):
I think this is good as a start. Get the CI passing, then let's merge and we can iterate on it with subsequent PRs.

Jatinagarwal24:
> Also the binary check should be improved:
> 
> ```
> Traceback (most recent call last):
>   File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 74, in <module>
> number of commits from current PR branch against origin/main is : 2
>     run_test()
>   File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 53, in run_test
>     if (is_file_binary(file_path)):
>         ^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/runner/work/lfortran/lfortran/check_binary_file_in_git_history.py", line 8, in is_file_binary
>     with open(file_path, 'r') as fp:
>          ^^^^^^^^^^^^^^^^^^^^
> IsADirectoryError: [Errno 21] Is a directory: '.claude/skills'
> a3f294e1f2c57073d57ca2cca2b4d15340a689e1
> 0661c275141db12422c22a3cd20a327751aadf2f
> 
> checked out to commit -->  a3f294e1f2c57073d57ca2cca2b4d15340a689e1
> ```

Through this commit  0499394b1f6500adcea69ee34bcb9ded215c23fe it fixes the crash in check_binary_file_in_git_history.py. Its `is_file_binary()` only catches `UnicodeDecodeError`, so a symbolic link
pointing to a directory raises an uncaught `IsADirectoryError` and aborts
the check. Any PR adding such a symbolic link would hit this.

check_binary_file_in_git_history.py
```diff
@@ -1,9 +1,15 @@
...
 
 # Open file_path, Read it with mode 'r' which expects a utf-8 encode.
 # If the file is binary it'd conflict with
 # text mode 'r', leading to raised error `UnicodeDecodeError`.
 def is_file_binary(file_path: str) -> bool:
+    # Only regular files can be binary. Skip symbolic links: opening a
+    # symlink that points to a directory raises `IsADirectoryError`, and
+    # the link itself stores only a path, never binary content.
+    if os.path.islink(file_path) or not os.path.isfile(file_path):
+        return False
```
certik, line 14: Will this still trigger on binary files? Can some files have `.os.path.isfile() == False`? Is the `os.path.isfile` needed here?

We have to be super careful when changing these checks.

Jatinagarwal24: Oh yes `os.path.isfile` is not needed here I have removed that.

Jatinagarwal24: Will this still trigger on binary files?  Yes  I ran the original and patched `is_file_binary()` side by side over the same inputs: (presenting in the table format to better understand)
<html>
<body>

<table>

file | ORIGINAL | PATCHED | identical?
-- | -- | -- | --
elf.bin | False | False | yes
nul.dat | False | False | yes
png.png | True | True | yes
utf16.txt | True | True | yes
a.txt | False | False | yes


</table>
</body>
</html>
So for any regular file the guard is a no-op , os.path.isfile() is True, so control falls straight through to the original open()/UnicodeDecodeError logic, untouched.

certik (approved):

PR #12273, opened 2026-07-18
Fix self-named and link-name generics with mixed-case names

certik:
## Summary
Fixes generic procedure handling for mixed-case self-named generics and for specific procedures that share a generic name (correct LLVM link name).

## Scope
- Semantics: mixed-case self-named generic (`generic_name_09`)
- Codegen: link name for generic-named specific (`generic_name_10`)

## Verification
- Integration tests `generic_name_09`, `generic_name_10` (gfortran, llvm)

## Rationale

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2235,6 +2235,7 @@ RUN(NAME generic_name_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME generic_name_06 LABELS gfortran llvm)
 RUN(NAME generic_name_07 LABELS gfortran llvm)
 RUN(NAME generic_name_08 LABELS gfortran llvm)
+RUN(NAME generic_name_09 LABELS gfortran llvm)
 
 RUN(NAME operator_overloading_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2236,6 +2236,7 @@ RUN(NAME generic_name_06 LABELS gfortran llvm)
 RUN(NAME generic_name_07 LABELS gfortran llvm)
 RUN(NAME generic_name_08 LABELS gfortran llvm)
 RUN(NAME generic_name_09 LABELS gfortran llvm)
+RUN(NAME generic_name_10 LABELS gfortran llvm)
 
 RUN(NAME operator_overloading_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2207,6 +2207,7 @@ RUN(NAME generic_name_06 LABELS gfortran llvm)
 RUN(NAME generic_name_07 LABELS gfortran llvm)
 RUN(NAME generic_name_08 LABELS gfortran llvm)
 RUN(NAME generic_name_09 LABELS gfortran llvm)
+RUN(NAME generic_name_10 LABELS gfortran llvm)
 
 RUN(NAME operator_overloading_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2206,6 +2206,7 @@ RUN(NAME generic_name_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME generic_name_06 LABELS gfortran llvm)
 RUN(NAME generic_name_07 LABELS gfortran llvm)
 RUN(NAME generic_name_08 LABELS gfortran llvm)
+RUN(NAME generic_name_09 LABELS gfortran llvm)
 
 RUN(NAME operator_overloading_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME operator_overloading_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/generic_name_09.f90
+++ b/integration_tests/generic_name_09.f90
@@ -1,0 +1,23 @@
+module generic_name_09_mod
+ implicit none
+ interface addCNullChar
+  module procedure addCNullChar
+ end interface
+contains
+ function addCNullChar(x) result(y)
+  integer, intent(in) :: x
+  integer :: y
+  y = x + 1
+ end function addCNullChar
+end module generic_name_09_mod
+
+program generic_name_09
+ use generic_name_09_mod
+ implicit none
+ integer :: r
+ r = addCNullChar(3)
+ print *, r
+ if (r /= 4) error stop
+ if (addCNullChar(10) /= 11) error stop
+ print *, "PASS"
+end program generic_name_09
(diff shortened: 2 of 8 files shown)


Jatinagarwal24 (commented):
Fix looks good 
but i want to point out one thing 
 we're now using `m_bindc_name` to carry an external link name for a Source-ABI function, which isn't what that field means. Right now nothing breaks, because every place that reads m_bindc_name checks `abi == BindC ` first (`asr_to_c_cpp.h:1119/3052, asr_to_fortran.cpp:670, asr_utils.h:6749, asr_to_llvm.cpp:8214/24200`). But that's incidental , none of those guards exist for this reason, so if someone later reads the field without checking the ABI they'd get the wrong name and it would surface as a link error that's hard to trace back here.

certik:
Right. So what should be a better way to fix this?

The first iteration did it in the backend completely, but it feels hackish. This seems better, but still not quite right.

Jatinagarwal24:
@certik 
the previous fix which you did recorded the external link name in `FunctionType.bindc_name` during semantics. That worked, but it overloaded `bindc_name`  for a Source-ABI function it now meant `"link name," not the bind(c, name=)` label  and it was only safe because every current reader of the field happens to guard on `abi == BindC`. Nothing enforces that, so a future reader without the ABI check would silently pick up the wrong name.

I tried to drop  the `bindc_name` write entirely and instead strips the `~genericprocedure` suffix in the backend to recover the link name. `bindc_name` goes back to a single meaning, no reader depends on an incidental ABI guard, and the suffix convention lives in one documented helper instead of being smuggled through an unrelated field.

Pls tell if any further changes are required .

certik:
Awesome, thanks for this! I think it looks better.

Any idea why we need to first append the suffix only to later remove it? Wouldn't it be even better not to append it in the first place?

Jatinagarwal24:
> Awesome, thanks for this! I think it looks better.
> 
> Any idea why we need to first append the suffix only to later remove it? Wouldn't it be even better not to append it in the first place?

Thanks for the review ,actually the suffix isn't for the link name, it's to avoid a `symbol-table` clash the `GenericProcedure` and the specific `Function` would both be `get_text`, so the specific one is stored under `get_text~genericprocedure` to keep the key unique. Codegen just strips it back to get the real name.So we can't skip the append it is needed for symbol resolution.

certik:
Here is an AI review. It called out the same issue, the string stripping doesn't seem like the right solution. I am investigating how to best handle this.

--------

Review: PR #12273 (https://github.com/lfortran/lfortran/pull/12273) — Fix self-named and link-name generics with mixed-case names

Author: certik (+ follow-up from Jatinagarwal24)  
Scope: Two focused fixes + integration tests generic_name_09 / generic_name_10  
CI: Green (including exhaustive checks)

───

Findings

[rework] Backend recovers link name by stripping a frontend naming convention in libasr

src/libasr/asr_utils.h (new helpers)  
src/libasr/codegen/asr_to_llvm.cpp (compute_llvm_function_name)

The final approach puts ~genericprocedure and strip_genericprocedure_suffix() in ASRUtils, and the LLVM backend peels that suffix for external interfaces:

} else if (is_external_interface) {
    fn_name = ASRUtils::strip_genericprocedure_suffix(sym_name);
}

That reintroduces the problem your own intermediate commit called out: libasr must not encode frontend symbol-table naming. The earlier bindc_name design was right about where the decision belongs (semantics → explicit ASR state → mechanical backend), but wrong about which field (m_bindc_name is BindC-only; Jatin’s comment is correct).

Dropping the bindc_name write was the right correction. Replacing it with string surgery in libasr is not the right replacement.

Recommended shape:

1. Keep the symbol-table key as "<name>~genericprocedure" (the clash with GenericProcedure is real).
2. At the point where semantics invents that key for an external interface body, store the real external/link name in a dedicated, ABI-neutral place (new optional field on Function / FunctionType, e.g. link/export name — not bindc_name).
3. Backend: if that field is set, use it; otherwise use m_name. No suffix knowledge in codegen.

If an ASR schema change is too heavy for this PR, a minimal acceptable interim is still better than suffix-stripping in libasr: keep the link name only in the frontend’s construction path and pass it through an existing structured channel that already means “external name” without overloading BindC — or accept that this PR is semantics-only for the case bug and park the link-name design until the field exists. The current hybrid is the worst of both: it documents the convention in libasr and teaches every future reader that mangled names are a valid codegen API.

───

[follow-up] genericprocedure_suffix is only used in one of several append sites

src/lfortran/semantics/ast_symboltable_visitor.cpp

On the PR branch, only the visit_Function site uses ASRUtils::genericprocedure_suffix. Other sites still hardcode "~genericprocedure":

• ~1587 (GenericProcedure collision path)
• ~1622 (interface_name == sym_name / generic map)
• ~3790 (add_generic_procedures)

If the constant stays, every append/strip should go through it. Better: remove the libasr constant entirely once link names are explicit (finding above).

───

[follow-up] Mixed-case coverage for the external-interface / link-name path

integration_tests/generic_name_09.f90 exercises mixed-case module procedure self-naming (semantics lookup).  
integration_tests/generic_name_10.f90 exercises same-name external interface + link, but all-lowercase get_text.

A small mixed-case variant of generic_name_10 (e.g. GetText / getText) would pin both halves of the PR together. Not a merge blocker given Fortran names are lowercased before storage, but it would match the title (“mixed-case names”).

───

What looks good

Case fix in add_generic_procedures is correct and minimal:

if (to_lower(pname.first) == proc.first) {
    correct_pname = pname.first + "~genericprocedure";
}
// ... later lowercased before resolve_symbol

Generic map keys are already lowercased (generic_name = to_lower(...)), while module procedure spellings keep source case. The old equality failed for mixed case and produced “Symbol … not declared”. generic_name_09 is a tight MRE-style integration test with gfortran + llvm.

Tests placement: both new tests are registered in integration_tests/CMakeLists.txt with the right labels. Scope is appropriately split from the larger netcdf2 work.

Abandoning bindc_name for Source-ABI functions was the right response to review: nothing enforces “read only if abi == BindC”, so stuffing a link name there is a latent trap.

───

Architecture assessment

┌─────────────┬─────────────────────────────────────────────┐
│ Question    │ Assessment                                  │
├─────────────┼─────────────────────────────────────────────┤
│ Earliest    │ AST→ASR / symbol table for both bugs: (1)   │
│ wrong phase │ case-sensitive self-name match when         │
│             │ building the generic’s procedure list; (2)  │
│             │ internal disambiguated Function.m_name used │
│             │ as external link name.                      │
├─────────────┼─────────────────────────────────────────────┤
│ Chosen fix  │ (1) Semantics — justified. (2) LLVM name    │
│ layer       │ mangling via string strip in libasr — not   │
│             │ justified as the long-term design; earlier  │
│             │ phase should record the external name       │
│             │ explicitly.                                 │
├─────────────┼─────────────────────────────────────────────┤
│ ASR         │ Still missing: “external interface Function │
│ invariant   │ ’s link symbol is explicit and independent  │
│             │ of symbol-table disambiguation.” Today that │
│             │ fact is encoded only as “m_name may end in  │
│             │ ~genericprocedure.”                         │
├─────────────┼─────────────────────────────────────────────┤
│ Backend     │ Small mechanical change, but it adds a      │
│ complexity  │ frontend naming contract into ASRUtils      │
│             │ instead of removing backend inference.      │
└─────────────┴─────────────────────────────────────────────┘

───

Verdict

Not quite ready to treat as architecturally settled, even though behavior looks fixed and CI is green.

• Ship-blocking for a strict architecture bar: the libasr suffix strip (rework finding).  
• Pragmatic “safe enough + iterate”: only if maintainers explicitly accept the temporary string convention with a tracked follow-up for a real link-name field — the current code does not document that debt beyond comments.

I would not approve without either:

1. restoring an explicit link name in ASR (not bindc_name), with the backend only reading that field; or  
2. a clear maintainer acknowledgment that the strip helper is interim debt, plus an issue for the proper field.

The mixed-case semantics fix alone is solid and could land on its own if you want to split further.

certik:
I tried to fix all this in https://github.com/lfortran/lfortran/pull/12384. Let me know what you think.

certik:
This is not the ideal fix, but it is the simplest fix. So let's merge it to unblock us, and we'll figure out a more principled fix later.

PR #12274, opened 2026-07-18
Fix BIND(C) VALUE argument passed an array element

certik:
## Summary
Pass BIND(C) VALUE arguments that are array elements correctly (load the element value instead of passing a pointer/descriptor incorrectly).

## Scope
- LLVM codegen for BIND(C) VALUE + array element actual
- Integration test `bindc_54`

## Verification
- `bindc_54` (gfortran, llvm)

## Rationale


certik (commented):
This PR is good to go I think.

PR #12275, opened 2026-07-18
Fix EXTERNAL procedures: multi-name, arg checks, private modules, leaks

certik:
## Summary
Several EXTERNAL / implicit-interface fixes used by codes such as netcdf-fortran:
multi-name EXTERNAL statements, argument type/count checks with CONTAINS and
use-association, private module access, and leaks of external symbols from
function scopes.

## Scope
- Semantics: EXTERNAL list handling, private module external access, function-scope cleanup
- Arg count/type for use-associated and CONTAINS cases
- Integration tests `external_17`–`external_19`, `implicit_interface_44`–`47`

## Verification
- Listed integration tests with `--implicit-interface` where required

## Rationale

### DIFF
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8176,38 +8176,6 @@ public:
             } else {
                 uint32_t old_h = llvm_symtab_fn_names[fn_name];
                 F = llvm_symtab_fn[old_h];
-                // Two implicit-interface external procedures can share the
-                // same link name yet be seen with different argument lists,
-                // e.g. a module declares `external :: nf_create` (no
-                // interface, zero args) while the caller passes actual
-                // arguments (nf_create(a, b, c)). The earlier, less complete
-                // declaration would otherwise be reused for the call site and
-                // the LLVM verifier would reject the call ("Incorrect number
-                // of arguments passed to called function!"). When the current
-                // interface has more arguments, replace the earlier
-                // declaration with one matching the current signature so all
-                // call sites use the correct type; existing references are
-                // redirected through a bitcast (they refer to the same
-                // link-time symbol).
-                if (F->getFunctionType() != function_type &&
-                        function_type->getNumParams() >
-                            F->getFunctionType()->getNumParams()) {
-                    llvm::Function* old_F = F;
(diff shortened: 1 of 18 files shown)


Jatinagarwal24 (approved):
Fix looks good to me .Just pushed a minor cleanup.
@certik Still if something you feel needs to be changed tell me I will do that.
Thanks

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8176,6 +8176,38 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                        llvm::ConstantExpr::getBitCast(new_F, old_F->getType()));
+                    old_F->eraseFromParent();
+                    new_F->setName(kept_name);
+                    for (auto& kv : llvm_symtab_fn) {
+                        if (kv.second == old_F) {
+                            kv.second = new_F;
+                        }
+                    }
+                    F = new_F;
+                }
```
certik: I don't like this part --- feels hackish and fragile and a fix too late. We need to figure out how to fix this in ASR itself, so that we don't need to do such hacks in ASR->LLVM.

I don't know what the right solution is, but I feel the above is not it.

certik:
@Jatinagarwal24 awesome, thanks for this!

We need to figure out a solution to https://github.com/lfortran/lfortran/pull/12275#discussion_r3667480733. I think there is a way.

src/libasr/codegen/asr_to_llvm.cpp
Jatinagarwal24: Thanks for the review I have tried to fix it at asr level pls once check it .

certik:
I think this looks much better.

Here is an AI review: https://gist.github.com/certik/df9c868b4e8d24a3cee451187f5ab257.

I am still figuring this out -- how to handle implicit functions as module variables without hacks.

certik:
I tried to fix it here: https://github.com/lfortran/lfortran/pull/12380, but I am still not happy.

certik:
Keep this PR as is, it might be the "least bad" way to do this for now.

certik:
The follow-up is in https://github.com/lfortran/lfortran/pull/12380.

certik:
Merged with other PRs.

PR #12276, opened 2026-07-18
Accept -fPIE and macOS -install_name linker flags

certik:
## Summary
Accept the `-fPIE` driver flag and the macOS `-install_name` linker flag so
third-party build systems can pass them through to LFortran.

## Scope
- Command-line parser / driver
- Integration tests `fpie_flag_01`, `install_name_01`

## Verification
- CLI integration tests

## Rationale


Jatinagarwal24:
Hi @certik I have pushed some minor changes. Pls see if they are correct or not .ALso tell if anything is else required.
Otherwise I think it looks good.

integration_tests/install_name_01.sh
```diff
@@ -0,0 +1,26 @@
...
+"$LFORTRAN" --shared -o "$LIB" -install_name "$INAME" install_name_01.o
+
+got=$(otool -D "$LIB" | tail -1 | tr -d '[:space:]')
+echo "requested install name: $INAME"
+echo "actual install name:    $got"
+if [ "$got" != "$INAME" ]; then
+    echo "ERROR: install name was not set to the requested value"
+    exit 1
+fi
+echo "install_name_01: OK"
```
certik: This seems fragile. Maybe it might be better not to test it, not sure yet.

certik:
I need to understand what exactly `-install_name` is doing and if we need to support it or not.

Jatinagarwal24:
> I need to understand what exactly `-install_name` is doing and if we need to support it or not.

So basically  `-install_name <path>` is a `macOS` only linker flag.When you link a shared library `(.dylib)`, it records `<path>` as the dylib's install name , the path dependent binaries embed and use to find it at load time (`LC_ID_DYLIB` in the Mach-O). Basically It's the `macOS` equivalent of GNU ld's `-soname`.

CMake passes it automatically whenever it links a shared lib on `macOS`, so any third-party project that builds a `.dylib` through LFortran hits it. If we don't accept it, the link just fails on an unknown flag.

So actually we need to handle it, but minimally  we don't implement anything ourselves. We just accept the flag, forward it to the linker as `-Wl,-install_name,<name>` only on macOS when linking -shared, and drop it elsewhere (as GNU ld wouldn't understand it). and that's actually enough to unblock third-party builds without adding real complexity.

integration_tests/install_name_01.sh
Jatinagarwal24: I have fixed this . Pls once check 
Thanks !

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -390,8 +412,21 @@ namespace LCompilers::CommandLineInterface {
...
+                parse_args.push_back(argv[i]);
+            }
+            opts.install_name = extract_install_name(parse_args);
+            std::vector<const char *> parse_argv;
+            parse_argv.reserve(parse_args.size() + 1);
+            parse_argv.push_back(argv[0]);
+            for (const std::string &arg : parse_args) {
+                parse_argv.push_back(arg.c_str());
+            }
+            app.parse(static_cast<int>(parse_argv.size()), parse_argv.data());
```
certik: I don't like how we now need to hack our argument parsing just to support this.

src/bin/lfortran_command_line_parser.cpp
```diff
@@ -205,6 +205,28 @@ namespace LCompilers::CommandLineInterface {
...
+            if (args[i] == "-install_name" && i + 1 < args.size()) {
+                install_name = args[i + 1];
+                i++;
+                continue;
+            }
+            filtered.push_back(args[i]);
+        }
+        args = std::move(filtered);
+        return install_name;
+    }
```
certik: Not ideal.

integration_tests/install_name_01.sh
certik: I would remove this file.

integration_tests/CMakeLists.txt
```diff
@@ -4976,3 +4977,29 @@ RUN(NAME get_environment_variable_01 LABELS gfortran llvm)
...
+    add_custom_target(install_name_01 ALL
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/install_name_01.o)
+    add_test(NAME install_name_01
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/install_name_01.sh
+                ${CMAKE_CURRENT_SOURCE_DIR}/install_name_01.f90
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    # Exercise the freshly built lfortran (the Fortran compiler for this
+    # project), not whatever `lfortran` happens to be on PATH.
+    set_tests_properties(install_name_01 PROPERTIES
+        ENVIRONMENT "LFORTRAN=${CMAKE_Fortran_COMPILER}")
```
certik: Is there a way to test this without the bash shell hack?

certik:
It probably would be good to support `-install_name, -dynamiclib, -framework`. However, for example we don't support `-shared`, but `--shared` and cmake already knows this about LFortran. We already support `-Wl,-install_name`. So I think we should use that approach here as well, so that we don't need to be hacking our command-line parser.

CMake already does this for NAG, so let's extend it for LFortran also. Flang also doesn't seem to support `-install_name`.

Alternatively, is there a way to support this in a clean way in our command-line parser? I am worried the current solution will break in corner cases, such as `-DX="-install_name"`. It would be good to not having to hack our parser.

Jatinagarwal24:
HI @certik I have resolved all your comments .
Pls once review the pr .
Thanks

certik:
CI has to be fixed.

integration_tests/CMakeLists.txt
```diff
@@ -4994,3 +4995,24 @@ RUN(NAME get_environment_variable_01 LABELS gfortran llvm)
...
+        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/libinstall_name_01.dylib
+        COMMAND ${CMAKE_Fortran_COMPILER} --shared
+                -Wl,-install_name,@rpath/libinstall_name_01.dylib
+                -o ${CMAKE_CURRENT_BINARY_DIR}/libinstall_name_01.dylib
+                ${CMAKE_CURRENT_SOURCE_DIR}/install_name_01.f90
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/install_name_01.f90
+    )
+    add_custom_target(install_name_01 ALL
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/libinstall_name_01.dylib)
+endif()
```
certik: I don't like this code here, it should be at the top of the file, but also this seems to add a special test. I don't think it's worth it.

Remember: each cmake test path must be independently tested and maintained. So if we create one path for just one test, now we need to maintain this cmake code and ensure it actually fails when it should (to catch bugs). That's more maintenance for us.

I would not test it, it doesn't seem worth it, or I would figure out a label that we can append and test this with more tests.

certik:
For `-fPIE` this is now fine.

PR #12277, opened 2026-07-18
Support non-standard BYTE type declaration

certik:
## Summary
Add the non-standard `byte` type (equivalent to `integer(1)` / `integer*1`) in
free-form and fixed-form, and update `ci/parser.yy.patch` for `KW_BYTE`.

## Scope
- Tokenizers + grammar + semantics lowering to integer(1)
- Integration test `byte_type_01`
- CI parser patch

## Verification
- `byte_type_01` (gfortran, llvm)

## Rationale


src/lfortran/parser/semantics.h
```diff
@@ -299,6 +299,12 @@ static inline ast_t* VAR_DECL_PRAGMA2(Allocator &al, Location &loc,
             a2kind_list(p.m_a, l, INTEGER(n, l)).p, 1, \
             nullptr, nullptr, None)
 
+#define ATTR_TYPE_BYTE(l) make_AttrType_t( \
+            p.m_a, l, \
+            decl_typeType::TypeInteger, \
+            a2kind_list(p.m_a, l, make_Num_t(p.m_a, l, 1, nullptr)).p, 1, \
+            nullptr, nullptr, None)
```
certik: This should give a warning that `byte` is non-standard, to use `integer(1)` (or equivalent) instead.

src/lfortran/parser/semantics.h
```diff
@@ -299,6 +299,12 @@ static inline ast_t* VAR_DECL_PRAGMA2(Allocator &al, Location &loc,
...
+
```
Jatinagarwal24: This macro duplicates `ATTR_TYPE_INT` with the kind fixed at 1, and `make_Num_t(p.m_a, l, 1, nullptr)` is already defined as` INT1(l) ` at `semantics.h:1143`. Could we drop the macro entirely and use the existing `ATTR_TYPE_EXPR` at the grammar rule instead?
```
| KW_BYTE { $$ = ATTR_TYPE_EXPR(Integer, INT1(@$), @$); }
```

src/lfortran/parser/parser.yy
```diff
@@ -1586,6 +1587,7 @@ integer_type_spec
 
 intrinsic_type_spec
     : integer_type_spec { $$ = $1; }
+    | KW_BYTE { $$ = ATTR_TYPE_BYTE(@$); }
```
Jatinagarwal24: Shouldn't be it be grouped with `integer_type_spec ` right above where other integer forms are also grouped.I am saying this because `byte` is also just `integer(1)`.

integration_tests/byte_type_01.f90
Jatinagarwal24, line 1: pls see if this test case can be added or not ?
```
integer :: byte
byte = 42
if (byte /= 42) error stop
```
`byte ` as a variable name .
I think this has not been tested.

Jatinagarwal24 (commented):
And also I think this got missed 
`parser.cpp:991` has the token→string table used for syntax errors and `--show-tokens`, and it goes straight from `KW_BLOCK` to `KW_CALL`. Every other keyword site in this PR has a byte entry, so I think `T(KW_BYTE, "byte")` belongs there too.

src/lfortran/parser/semantics.h
certik: Although we want to give a warning, so probably need a dedicated macro.

integration_tests/byte_type_01.f90
certik: It's been added.

PR #12278, opened 2026-07-18
Fix PARAMETER statement: real value on integer and negative literals

certik:
## Summary
Handle PARAMETER with an integer variable initialized from a real value, and
negative literals in PARAMETER statements.

## Scope
- Semantics in the common visitor
- Integration tests `parameter_19`, `parameter_20`

## Verification
- Listed integration tests

## Rationale


certik (commented):
I think this is fine.

PR #12280, opened 2026-07-18
Optionally short-circuit .and./.or. in the LLVM backend

certik:
## Summary
Lower logical `.and.` / `.or.` with short-circuit evaluation in the LLVM backend
so the right operand is not evaluated when the left already determines the
result (common guard idiom; needed for netcdf-fortran with bounds checks).

## Scope
- `visit_LogicalBinOp` short-circuit codegen
- Integration test `logical_short_circuit_01`
- Updated LLVM reference outputs for affected tests

## Verification
- `logical_short_circuit_01` plus unit/reference LLVM tests

## Rationale

### DIFF
--- a/integration_tests/logical_short_circuit_01.f90
+++ b/integration_tests/logical_short_circuit_01.f90
@@ -1,0 +1,57 @@
+program logical_short_circuit_01
+    ! Fortran permits short-circuit evaluation of .and. and .or.. This test
+    ! uses the common idiom where the first operand guards evaluation of the
+    ! second so that an out-of-bounds array element (a(0)) is not accessed.
+    ! With array bounds checking enabled this must not abort at run time.
+    implicit none
+    integer :: a(3)
+    integer :: i
+    logical :: r
+
+    a = [10, 20, 30]
+
+    ! .and.: when the guard (i >= 1) is false, the second operand a(i)
+    ! must not be evaluated (it would index the out-of-bounds a(0)).
+    i = 0
+    if (i >= 1 .and. a(i) == 10) then
+        error stop "and: guarded operand must not be evaluated"
+    end if
+
+    ! .and. with the guard true: the second operand is evaluated and used.
+    i = 2
+    r = (i >= 1 .and. a(i) == 20)
+    if (.not. r) error stop "and: expected true"
+
+    i = 1
+    r = (i >= 1 .and. a(i) == 20)
+    if (r) error stop "and: expected false"
+
+    ! .or.: when the guard (i < 1) is true, the second operand must not
+    ! be evaluated (it would index the out-of-bounds a(0)).
+    i = 0
+    if (.not. (i < 1 .or. a(i) == 10)) then
+        error stop "or: guarded operand must not be evaluated"
+    end if
+
+    ! .or. with the guard false: the second operand is evaluated and used.
+    i = 3
+    r = (i < 1 .or. a(i) == 30)
+    if (.not. r) error stop "or: expected true"
+
+    i = 3
+    r = (i < 1 .or. a(i) == 99)
+    if (r) error stop "or: expected false"
+
+    ! chained .and.: left-to-right association, so the leftmost false guard
+    ! short-circuits the rest and no out-of-bounds access happens.
+    i = 0
+    if (i >= 1 .and. a(i) == 10 .and. a(i + 1) == 20) then
+        error stop "chained and: guarded operand must not be evaluated"
+    end if
+
+    ! non-short-circuit operators still evaluate both operands correctly.
+    if ((.true. .neqv. .false.) .neqv. .true.) error stop "neqv"
+    if ((.true. .eqv. .true.) .neqv. .true.) error stop "eqv"
+
+    print *, "All tests passed."
+end program logical_short_circuit_01
--- a/integration_tests/logical_short_circuit_01.f90
+++ b/integration_tests/logical_short_circuit_01.f90
@@ -1,12 +1,16 @@
 program logical_short_circuit_01
-    ! Fortran permits short-circuit evaluation of .and. and .or.. This test
-    ! uses the common idiom where the first operand guards evaluation of the
-    ! second so that an out-of-bounds array element (a(0)) is not accessed.
-    ! With array bounds checking enabled this must not abort at run time.
+    ! Fortran permits, but does not require, short-circuit evaluation of
+    ! .and. and .or., so this test is only meaningful when compiled with
+    ! --logical-short-circuit (it is registered with that flag and only for
+    ! the llvm label). It uses the common idiom where the first operand
+    ! guards evaluation of the second so that an out-of-bounds array
+    ! element (a(0)) is not accessed. With array bounds checking enabled
+    ! this must not abort at run time.
     implicit none
     integer :: a(3)
     integer :: i
     logical :: r
+    character(:), allocatable :: s
 
     a = [10, 20, 30]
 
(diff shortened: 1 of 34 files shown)


HarshitaKalani (approved):

integration_tests/logical_short_circuit_01.f90
```diff
@@ -0,0 +1,57 @@
...
+    i = 0
+    if (i >= 1 .and. a(i) == 10 .and. a(i + 1) == 20) then
+        error stop "chained and: guarded operand must not be evaluated"
+    end if
+
+    ! non-short-circuit operators still evaluate both operands correctly.
+    if ((.true. .neqv. .false.) .neqv. .true.) error stop "neqv"
+    if ((.true. .eqv. .true.) .neqv. .true.) error stop "eqv"
+
+    print *, "All tests passed."
```
certik, line 72: This whole test is problematic, as in GFortran it is not guaranteed that it will short circuit, so this code is invalid Fortran.

certik:
This must be guarded with a compiler flag, and for now probably not enabled by default.

I think it's not currently required and not proposed: https://grok.com/share/bGVnYWN5_e7ce1067-b51f-4f4b-a373-fddf49bcd70e, although Grok can be wrong of course.

Jatinagarwal24:
Hi @certik I have addressed your comment.
Pls review this pr and pls tell if something else needs to be changed or not .
Thanks .

Jatinagarwal24:
This change  is not due to the commit i added. I am getting this on current main also .
```diff
diff --git a/tests/reference/fortran-coarray_collectives_01-4fd95fc.json b/tests/reference/fortran-coarray_collectives_01-4fd95fc.json
index 2c330c1393..218b479570 100644
--- a/tests/reference/fortran-coarray_collectives_01-4fd95fc.json
+++ b/tests/reference/fortran-coarray_collectives_01-4fd95fc.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_initialization_01-23c9c70.json b/tests/reference/fortran-coarray_initialization_01-23c9c70.json
index e643af4ebd..e144fd1dda 100644
--- a/tests/reference/fortran-coarray_initialization_01-23c9c70.json
+++ b/tests/reference/fortran-coarray_initialization_01-23c9c70.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_operations_01-fb45389.json b/tests/reference/fortran-c
oarray_operations_01-fb45389.json
index 34834efb7b..344cdf8f36 100644
--- a/tests/reference/fortran-coarray_operations_01-fb45389.json
+++ b/tests/reference/fortran-coarray_operations_01-fb45389.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_saved_01-1f96b4c.json b/tests/reference/fortran-coarra
y_saved_01-1f96b4c.json
index 560d8ae695..5d4a7fc1e9 100644
--- a/tests/reference/fortran-coarray_saved_01-1f96b4c.json
+++ b/tests/reference/fortran-coarray_saved_01-1f96b4c.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_sync_01-4cab83b.json b/tests/reference/fortran-coarray
_sync_01-4cab83b.json
index 79d70fff77..ef00ceaa52 100644
--- a/tests/reference/fortran-coarray_sync_01-4cab83b.json
+++ b/tests/reference/fortran-coarray_sync_01-4cab83b.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_teams_01-442657a.json b/tests/reference/fortran-coarra
y_teams_01-442657a.json
index baa2ef40de..29c511f09d 100644
--- a/tests/reference/fortran-coarray_teams_01-442657a.json
+++ b/tests/reference/fortran-coarray_teams_01-442657a.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_teams_02-f6f38f1.json b/tests/reference/fortran-coarra
y_teams_02-f6f38f1.json
index c9e9069351..33db62632d 100644
--- a/tests/reference/fortran-coarray_teams_02-f6f38f1.json
+++ b/tests/reference/fortran-coarray_teams_02-f6f38f1.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
diff --git a/tests/reference/fortran-coarray_var_name_collision_01-94763bf.json b/tests/reference/f
ortran-coarray_var_name_collision_01-94763bf.json
index 816b793ddb..41cccf36c3 100644
--- a/tests/reference/fortran-coarray_var_name_collision_01-94763bf.json
+++ b/tests/reference/fortran-coarray_var_name_collision_01-94763bf.json
@@ -10,4 +10,4 @@
     "stderr": null,
     "stderr_hash": null,
     "returncode": 0
-}
+}
\ No newline at end of file
(END)
```

certik:
See https://github.com/lfortran/lfortran/issues/12382 for the EOF issue.

integration_tests/logical_short_circuit_01.f90
```diff
@@ -0,0 +1,73 @@
...
+    integer :: i
+    logical :: r
+    character(:), allocatable :: s
+
+    a = [10, 20, 30]
+
+    ! .and.: when the guard (i >= 1) is false, the second operand a(i)
+    ! must not be evaluated (it would index the out-of-bounds a(0)).
+    i = 0
+    if (i >= 1 .and. a(i) == 10) then
```
certik, line 20: I can confirm this fails in main:
```
$ lfortran a.f90
runtime error: Runtime error: Array 'a' index out of bounds. Tried to access index 0 of dimension 1, but valid range is 1 to 3.
  --> a.f90:20:22
   |
20 |     if (i >= 1 .and. a(i) == 10) then
   |                      ^^^^
```
as it should.

certik: Let's add this test into our reference tests, to ensure it fails without `--logical-short-circuit`.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -14518,6 +14518,83 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
             this->visit_expr_wrapper(x.m_value, true);
             return;
         }
+        // Fortran permits (but does not require) short-circuit evaluation of
+        // the .and. and .or. logical operators. Real-world code commonly
```
```suggestion
        // the .and. and .or. logical operators. Real-world code sometimes incorrectly
```

integration_tests/logical_short_circuit_01.f90
certik: Done.

src/libasr/codegen/asr_to_llvm.cpp
certik: Applied.

certik:
@Jatinagarwal24 this should be in a good shape now.

Jatinagarwal24:
> @Jatinagarwal24 this should be in a good shape now.

Yes it is looking good now I  was thinking if we should also improve our error message and tell the user to use flag `--logical-short-circuit`. Otherwise the user will not know which flag to use it while compiling . Pls tell me what do you think about this?

certik:
Which error message, this one: "runtime error: Runtime error: Array 'a' index out of bounds. Tried to access index 0 of dimension 1, but valid range is 1 to 3." ? It's not easy to tell, especially at runtime, that using this option would help.

Jatinagarwal24:
> Which error message, this one: "runtime error: Runtime error: Array 'a' index out of bounds. Tried to access index 0 of dimension 1, but valid range is 1 to 3." ? It's not easy to tell, especially at runtime, that using this option would help.

yes i was referring to this .
Oh ok we can drop this idea.
Then we are good to merge this pr.

PR #12281, opened 2026-07-19
Fix stack buffer overflow in get_local_info_dwarfdump

### LINKED ISSUE #10603: Bug in runtime stacktrace with `-g`
The segfault is in LFortran's runtime library.
  When compiled with -g, LFortran's stop 1 calls print_stacktrace_addresses()
  → get_local_info_dwarfdump() to print a stack trace, and that function
  crashes when the binary links against a library (like libpetaca.a).

  Backtrace:

   get_local_info_dwarfdump + 788   (liblfortran_runtime)
   print_stacktrace_addresses + 96  (liblfortran_runtime)
   main + 300

  Minimal reproducer:

   program mre
     use timer_tree_type  ! any module from libpetaca.a
     implicit none
     stop 1               ! crashes 100% with -g
   end program

   - -g flag → crash. No -g → no crash.

usefahmed07:
## Summary

Fix a stack buffer overflow in `get_local_info_dwarfdump` where the
`addresses`/`line_numbers` arrays were written without bounds checking
against `LCOMPILERS_MAX_STACKTRACE_LENGTH` (200). Programs whose debug
line table has more than 200 entries corrupt the stack, leading to a
segfault (or corrupted stacktrace output) when a runtime `stop`
triggers a stacktrace print with `-g`.

## Scope

- `get_local_info_dwarfdump` bounds check on `d->stack_size` before
  writing to `d->addresses` / `d->line_numbers`
- Regression test `errors/runtime_stacktrace_02.f90` (`run_with_dbg`)
- Updated reference stderr/json for the new test

## Verification

- `errors/runtime_stacktrace_02.f90` — a program with 259 subroutine
  calls, producing a debug line table with 781 entries (well past the
  200-entry limit), reliably segfaults before the fix and prints a
  correct stacktrace after it
- Full `run_tests.py` suite passes

## Rationale

Fixes #10603. The original report links the crash to linking against
an external static library (`libpetaca.a`) built with a different
compiler, which inflates the executable's debug line table. Root cause
turned out to be independent of any particular library or compiler:
any binary whose debug line table exceeds 200 entries triggers the
same out-of-bounds write, which we reproduce here with a
self-contained Fortran program (no external dependencies needed).


usefahmed07:
Note on diff size: most of the +1327 lines come from the regression
test file (`errors/runtime_stacktrace_02.f90`), which needs 259
subroutine calls to push the debug line table past the
`LCOMPILERS_MAX_STACKTRACE_LENGTH` (200) threshold and reliably
trigger the overflow. Happy to shrink the test if a smaller repro is
preferred - let me know.

Edit: now was edited

tests/errors/runtime_stacktrace_02.f90
```diff
@@ -0,0 +1,1300 @@
+program runtime_stacktrace_02
```
certik: I don't like having this large file, only to test an out of bounds error. I would not test this.

certik:
There were several bugs, I just fixed them. I removed the tests which I think is too long.

usefahmed07:
> There were several bugs, I just fixed them. I removed the tests which I think is too long.

You are right , but I disagree with this. It's a very large file. Thank you.

PR #12290, opened 2026-07-19
fix : allocate(source=) with character pointer components

### LINKED ISSUE #12285: LFortran 0.64 segfault at runtime on allocate with characters
LFortran 0.64 segfaults at runtime on allocate(dst, source=src) when the source type has a character pointer component (e.g. character(len=:), pointer :: p). The crash happens even when the pointer is null/unassociated, for both deferred- and fixed-length character pointers, and for both polymorphic (class(...), allocatable) and plain (type(...), allocatable) dst; swapping the component to a character(len=:), allocatable (non-pointer), or using a non-character pointer, works fine. gfortran, ifx, and nvfortran all run the reproducer correctly.

Tested on MacOS ARM. 

```fortran 
module m
   implicit none
   type :: t
      character(len=:), pointer :: p => null()   ! character pointer component
   end type
contains
   subroutine clone(dst, src)
      type(t), allocatable, intent(out) :: dst
      type(t), intent(in) :: src
      allocate(dst, source=src)                   ! <-- segfaults under LFortran
   end subroutine
end module
program charptr_source
   use m
   implicit none
   type(t) :: a           ! a%p is null (unassociated)
   type(t), allocatable :: b
   call clone(b, a)
   print *, "OK"
end program
````

Jatinagarwal24:
fixes #12285 

A derived type with a `character(len=:), pointer` component failed to
compile under `allocate(dst, source=src)` -- first an assertion failure in
codegen, then a runtime error once that was cleared.

Two defects:

1. Struct member init treated `=> null()` as a string value to copy.
   `visit_PointerNullConstant` emits a bare `i8*` null for string types,
   but `lfortran_str_copy` expects a `%string_descriptor*`, tripping
   `is_proper_string_llvm_variable`. The guards checked only that the
   initializer's *type* was a string -- true for `null()` -- never that it
   was a null pointer.

2. `LLVMUtils::deepcopy` value-copied character pointer components. `source=`
   propagates pointer components by association, not by value. The copy also
   passed `is_dest_allocatable=false` (a pointer isn't allocatable) with
   deferred length, hitting the `_lfortran_strcpy_alloc` branch that asserts
   a non-NULL destination -- but `dst%p` is NULL right after allocation.
   Surfaced as "Non-allocatable string isn't allocated."

Fixes:
- asr_to_llvm.cpp: skip the string copy for `PointerNullConstant_t`
  initializers on both the `initialize_val` and `!initialize_val` paths.
  `setup_string` already leaves the descriptor in the null state `null()`
  denotes.
- llvm_utils.cpp: copy the `{i8*, i64}` descriptor for non-allocatable
  character pointers so dest aliases src's target, matching the
  function-pointer, descriptor-array, and default branches of the same
  `case`. The string sub-case was the lone outlier doing a value copy. Null
  propagates naturally as a null descriptor. Allocatable character
  components keep value-copy semantics.


JorgeG94:
Thanks again, very much appreciated

HarshitaKalani (approved):
lgtm. Thanks!

PR #12293, opened 2026-07-19
fix: default .f files to fixed-form parsing (fixes #12283)

### LINKED ISSUE #12283: syntax error: Token '**' is unexpected here
While I'm attempting to compile Xfoil (https://web.mit.edu/drela/Public/web/xfoil/), I've encountered the following parser error:

```
lfortran -c -O  -fdefault-real-8  ../src/xfoil.f
syntax error: Token '**' is unexpected here
 --> ../src/xfoil.f:1:4
  |
1 | C***********************************************************************
  |    ^^
```

Here's the snippet of `xfoil.f`

```fortran
C***********************************************************************
C    Module:  xfoil.f
C
C    Copyright (C) 2000 Mark Drela
C
C    This program is free software; you can redistribute it and/or modify
C    it under the terms of the GNU General Public License as published by
C    the Free Software Foundation; either version 2 of the License, or
C    (at your option) any later version.
C
C    This program is distributed in the hope that it will be useful,
C    but WITHOUT ANY WARRANTY; without even the implied warranty of
C    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
C    GNU General Public License for more details.
C
C    You should have received a copy of the GNU General Public License
C    along with this program; if not, write to the Free Software
C    Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
C***********************************************************************
C
      PROGRAM XFOIL
```

It turns out, `lfortran` struggles to parse `C*`

```fortran
C****
      PROGRAM FOO

      END !FOO
```

```
$ lfortran -c -O -fdefault-real-8 foo.f
syntax error: Token '**' is unexpected here
 --> foo.f:1:4
  |
1 | C****
  |    ^^
```

I'm not well versed in Fortran standard, but `gfortran` does not have this limitation and it's desirable `lfortran` can be a drop-in replacement of `gfortran`.

usefahmed07:
## Summary

`lfortran` requires an explicit `--fixed-form` or `--fixed-form-infer` flag to parse `.f` files as fixed-form, unlike `gfortran` which infers this automatically from the file extension. Because of this mismatch, fixed-form comment lines like `C***...` (no space after the comment marker) get misparsed as free-form code:

    $ lfortran --show-ast foo.f
    syntax error: Token '**' is unexpected here
     --> foo.f:1:4
      |
    1 | C***********************************************************************
      |    ^^

After this fix, the same command produces a valid AST with no errors.

The comment-line detection itself (`determine_line_type` in `src/lfortran/parser/parser.cpp`) was already correct — the real bug is that `fixed_form_infer` defaults to `false`. This PR defaults it to `true` so `.f`/`.F` files are parsed as fixed-form by default, matching `gfortran`.

Fixes #12283

**Tests:** `integration_tests/fixed_form_infer_default_01.f` is a new `add_test` that runs `lfortran --show-ast` directly on a `.f` file with no flags (fails on `main`, passes on this branch). `integration_tests/fixed_form_comment_02.f90` guards fixed-form comment lines with no space after the marker (`C***`, `c---`, `*...`).


tests/parser/fixed_form_infer_default_01.f
```diff
@@ -0,0 +1,10 @@
+C***********************************************************************
+C     Regression test for https://github.com/lfortran/lfortran/issues/12283
+C     A `.f` file must be parsed as fixed-form by default, with no
+C     --fixed-form / --fixed-form-infer flag needed, matching gfortran.
+C***********************************************************************
+      PROGRAM FIXED_FORM_INFER_DEFAULT_01
+      INTEGER A
+      A = 1
+      IF (A .NE. 1) ERROR STOP
+      END
```
certik, line 10: This is not used?

integration_tests/fixed_form_comment_02.f90
```diff
@@ -0,0 +1,14 @@
+C***********************************************************************
+C     Comment lines that begin with a comment indicator (C, c, or *)
+C     immediately followed by more of the same or other characters
+C     (no space required) must still be recognized as full-line
+C     comments in fixed form, and a `.f` file must be treated as
+C     fixed-form by default, without requiring --fixed-form.
```
certik:
```suggestion
C     comments in fixed form.
```

src/bin/lfortran_command_line_parser.h
```diff
@@ -60,7 +60,7 @@ namespace LCompilers::CommandLineInterface {
         std::string linker_path{""};
         bool print_targets = false;
         bool print_c_include_dir = false;
-        bool fixed_form_infer = false;
+        bool fixed_form_infer = true;
```
certik: This need to stay false.

tests/parser/fixed_form_infer_default_01.f
usefahmed07: u are  right it wasn't wired up to anything moved it to tests/parser/ and registered it in tests/tests.toml with ast = true so it actually runs now

src/bin/lfortran_command_line_parser.h
usefahmed07: Done

assem2002 (commented):
it looks good to me, no issues.
but I can see that `fixed_form_infer` is a flag that we introduced to make the compiler switch to the `fixed_form` mode if it detects `.f` or `.F`.
With this PR, the flag has no use.
***
@certik 
- Do you think we should keep the flag functioning(neglect changes),and raise a suggestion to user when file ends with `.f`?
- If we'll proceed with this PR, we'll have a flag that has no use in the codebase, if we dump it, some users may complain.
- We may proceed with the PR, and keep the flag, but raise a warning that the flag will get removed in future versions.

certik:
No, we should treat `.f` as fixed form always, at this time. We can do that later, but right now we want to keep the door open to use `.f` for modern Fortran.

PR #12300, opened 2026-07-21
fix: support %re/%im on allocatable complex arrays in LLVM backend

### LINKED ISSUE #12299: ICE from complex allocated array
This program when reported in #4427 gave  a segfault. But #4427 has been removed from the list of issues, and now the program gives an ICE.
```
program cmplxinf
  implicit none
  complex,allocatable :: z(:)
  real                :: zero = 0
  print *,'Run time'
  allocate(z(3))
  z = cmplx(1/zero,1/zero)
  print "(A,3F9.0)", 'z%im =', z%im
end program cmplxinf
```
Lfortran 0.64.0 output:
```
code generation error: Internal error: x is nullptr
```
Gfortran or ifx output:
```
 Run time
z%im = Infinity Infinity Infinity
```

Jatinagarwal24:
fixes #12299

Compiling `z%im` where `z` is an allocatable (deferred-shape) complex
array crashed the LLVM backend with an assertion in create_gep2, and
`sum(z%re)` crashed the array_passed_in_function_call pass.

Three fixes:

1. asr_to_llvm.cpp (handle_arr_for_complex_im_re): compute descriptor
   LLVM types past the Allocatable/Pointer wrapper, so create_gep2
   receives the descriptor struct type instead of a pointer to it.

2. asr_to_llvm.cpp (handle_arr_for_complex_im_re): when the shape is
   deferred, the ASR dims have no compile-time start/length, so
   PointerToData_to_Descriptor cannot be used. Instead, build the
   real-part view's descriptor by copying dims from the complex
   array's descriptor at runtime (reset_array_details), set the CFI
   fields, and scale the source offset by 2 (complex elements ->
   real elements). The existing shared code then adds 1 to the
   offset for %im and doubles the strides, as before.

3. array_passed_in_function_call.cpp (set_allocation_size): handle
   ComplexRe/ComplexIm expressions; the temporary has the same shape
   as the complex argument, mirroring the Cast/LogicalNot cases.


HarshitaKalani (approved):
Looks good to me. Thanks!

PR #12303, opened 2026-07-21
fix: dereference scalar pointer/allocatable items in namelist I/O

### LINKED ISSUE #12302: Namelist I/O of scalar pointer/allocatable variables reads the pointer instead of its target (garbage values, segfault with derived types)
MRE
```
program mre
implicit none
integer, target :: t = 10
integer, pointer :: x => null()
namelist / config / x
x => t
print *, 'plain print of x:', x
write(*, nml=config)
end program
```
LFortrran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran mre.f90 
plain print of x: 10
 &config
  x=4210792
 /
```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran mre.f90 && ./a.out 
 plain print of x:          10
&CONFIG
 X=10         ,
 /
```

Jatinagarwal24:
fixes #12302 

In build_namelist_descriptor, the symtab value for scalar pointer and
allocatable namelist items was used directly as the data pointer, so
the runtime read the pointer variable's own storage instead of its
target. This produced garbage values for intrinsic types and a
segfault for pointers to derived types with character components.

Load through the symtab value for non-character scalar
pointer/allocatable items, matching the existing handling of
pointer/allocatable arrays.

Adds integration test namelist_36 (round-trips a scalar integer
pointer, a scalar real allocatable, and a pointer to a derived type
with a character component).

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1139,6 +1139,7 @@ RUN(NAME namelist_34 LABELS gfortran llvm)
 
 
 RUN(NAME namelist_35 LABELS gfortran llvm)
+RUN(NAME namelist_36 LABELS gfortran llvm)
 
 RUN(NAME nested_namelist_01 LABELS gfortran llvm)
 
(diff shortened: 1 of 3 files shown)


HarshilShah1804 (approved):
Looks good.

PR #12304, opened 2026-07-21
Fix ICE for array-valued coindexed reference

### LINKED ISSUE #12261: ICE for array-valued coindexed reference
## MRE
```fortran
program test
    implicit none

    integer :: a(5)[*] = 0
    integer :: i

    do i = 1, 5
        a(i) = this_image() * 10 + i
    end do

    if (this_image() == 1) then
        print *, a(:)[2]
    end if
end program
```

## LFortran's Output
```
(lf) jainam@jainam-ThinkBook-15-G5-ABP:~/Documents/gsoc/lfortran$ ./src/bin/lfortran tests/1a.f90 
....
LCompilersException: get_struct_sym_from_struct_expr() not implemented for 72
```

srinjoy933:
Fixes #12261

### DIFF
--- a/integration_tests/coarrays_28.f90
+++ b/integration_tests/coarrays_28.f90
@@ -1,0 +1,14 @@
+﻿program test
+    implicit none
+
+    integer :: a(5)[*] = 0
+    integer :: i
+
+    do i = 1, 5
+        a(i) = this_image() * 10 + i
+    end do
+
+    if (this_image() == 1) then
+        print *, a(:)[2]
+    end if
+end program
--- a/integration_tests/coarrays_28.f90
+++ b/integration_tests/coarrays_28.f90
@@ -1,4 +1,4 @@
-﻿program test
+﻿program coarrays_28
     implicit none
 
     integer :: a(5)[*] = 0
@@ -11,4 +11,4 @@
     if (this_image() == 1) then
         print *, a(:)[2]
     end if
-end program
+end program coarrays_28
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4740,6 +4740,7 @@ RUN(NAME coarrays_24 EXTRA_ARGS --coarray=true)
 
 RUN(NAME coarrays_26 EXTRA_ARGS --coarray=true)
 RUN(NAME coarrays_27 EXTRA_ARGS --coarray=true NUM_IMAGES=4)
+RUN(NAME coarrays_28 EXTRA_ARGS --coarray=true)
 
 RUN(NAME gpu_metal_01 LABELS gfortran llvm metal)
 RUN(NAME gpu_metal_02 LABELS gfortran llvm metal)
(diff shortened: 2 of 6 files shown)


srinjoy933:
hey @Jainam-not-a-robot ,please have a look at this ,need some suggestion from you  regarding the test failures.

integration_tests/coarrays_28.f90
bonachea, line 1: Please add a new reference test entry in `tests/tests.toml` to cover the new feature:
```
[[test]]
filename = "../integration_tests/coarrays_28.f90"
fortran = true
pass = "coarray"
options = "--coarray=true"
```
see the entry for `coarray_operations_01.f90` for an example.

You'll also need to use `run_tests.py` to generate the reference .json and .f90 output

integration_tests/coarrays_28.f90
```diff
@@ -0,0 +1,14 @@
+﻿program coarrays_28
+    implicit none
+
+    integer :: a(5)[*] = 0
+    integer :: i
+
+    do i = 1, 5
+        a(i) = this_image() * 10 + i
+    end do
```
bonachea, line 9: You are missing synchronization between the coarray writes and subsequent read

```suggestion
    end do
    
    sync all
```

integration_tests/coarrays_28.f90
```diff
@@ -0,0 +1,14 @@
...
+
+    if (this_image() == 1) then
+        print *, a(:)[2]
+    end if
```
bonachea, line 13: Integration tests should be self-validating and `error stop` if the test fails.

```suggestion
    if (this_image() == 1) then
        print *, a(:)[2]
    end if
    if (any(a(:)[2] /= [(20 + i, i=1,5)])) error stop
```

bonachea (changes requested):
Thanks for working on this @srinjoy933 !

I'd very much like to see LFortran add coindexed access for arrays, although I'm concerned this PR might not yet handle all the subtle details correctly.

The new test has a number of defects that must be addressed, which most importantly includes adding reference output so we can examine the  `prif_get` calls resulting from the coarray transformation pass. It's likely these calls are currently incorrect, which would explain the runtime crashes in CI.

bonachea:
@srinjoy933 is there progress on this?

I can help you debug any runtime crashes if you post the generated PRIF calls from the ASR-to-Fortran output, as suggested in my review.

Regarding the CI failures, it looks like the following assertion failure in the compiler on the new test:
```
  Running coarray test: integration_tests/coarrays_28.f90 (images: 2)
  =========================================
  + base=coarrays_28
  + lfortran integration_tests/coarrays_28.f90 --coarray=true -o coarrays_28_lf.out -L/home/runner/work/lfortran/lfortran/caffeine/inst/lib -lcaffeine -lgasnet-smp-seq
  LCOMPILERS_ASSERT failed: /home/runner/work/lfortran/lfortran/src/libasr/pass/coarray.cpp
  function get_coarray_companions(), line number 665 at 
  hsym && dsym
  ERROR MESSAGE:
  Coarray companion variables not found
```
so I would guess you need to adjust the creation of coarray companion variables in `src/libasr/pass/coarray.cpp` to handle array coarrays.

PR #12305, opened 2026-07-21
Fix support for assumed-rank character arrays and inquiry intrinsics in select rank

### LINKED ISSUE #12167: ICE for select rank on assumed-rank character array
## MRE

```fortran
program p
  character :: c(1)

  call s(c)

contains

  subroutine s(a)
    character, intent(in) :: a(..)

    select rank(a)
    rank(1)
      print *, a(1)
    end select
  end subroutine s

end program p
```

## output

```text

LCOMPILERS_ASSERT failed:
Index out of bounds in create_gep2: index 3 is out of range for struct with 2 elements

```

## gfortran behavior

```text

 
```

srinjoy933:
Fixes #12167

### DIFF
--- a/integration_tests/select_rank_39.f90
+++ b/integration_tests/select_rank_39.f90
@@ -1,4 +1,4 @@
-program p
+program select_rank_39
   implicit none
   character :: c(1)
   c(1) = 'A'
@@ -12,8 +12,8 @@ contains
 
     select rank(a)
     rank(1)
-      print *, a(1)
+      if (a(1) /= 'A') error stop
     end select
   end subroutine s
   
-end program p
+end program select_rank_39
--- a/integration_tests/select_rank_39.f90
+++ b/integration_tests/select_rank_39.f90
@@ -1,0 +1,19 @@
+program p
+  implicit none
+  character :: c(1)
+  c(1) = 'A'
+  
+  call s(c)
+  
+contains
+
+  subroutine s(a)
+    character, intent(in) :: a(..)
+
+    select rank(a)
+    rank(1)
+      print *, a(1)
+    end select
+  end subroutine s
+  
+end program p
--- a/integration_tests/select_rank_39.f90
+++ b/integration_tests/select_rank_39.f90
@@ -10,6 +10,8 @@ contains
   subroutine s(a)
     character, intent(in) :: a(..)
 
+    if (len(a) /= 1) error stop
+
     select rank(a)
     rank(1)
       if (a(1) /= 'A') error stop
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -13215,7 +13215,7 @@ public:
                             ASR::array_physical_typeType::DescriptorArray, desc_type, nullptr
                         );
                         temp = ASRUtils::EXPR(array_cast);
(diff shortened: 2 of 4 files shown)


integration_tests/select_rank_39.f90
```diff
@@ -0,0 +1,19 @@
...
+  call s(c)
+  
+contains
+
+  subroutine s(a)
+    character, intent(in) :: a(..)
+
+    select rank(a)
+    rank(1)
+      print *, a(1)
```
HarshitaKalani: Can we have an error stop condition here to ensure the output is always correct?

integration_tests/select_rank_39.f90
```diff
@@ -0,0 +1,19 @@
+program p
```
HarshitaKalani: NIT
```suggestion
program select_rank_39
```

integration_tests/select_rank_39.f90
```diff
@@ -0,0 +1,19 @@
...
+  subroutine s(a)
+    character, intent(in) :: a(..)
+
+    select rank(a)
+    rank(1)
+      print *, a(1)
+    end select
+  end subroutine s
+  
+end program p
```
HarshitaKalani:
```suggestion
end program select_rank_39
```

HarshitaKalani (approved):
I think the changes are good. Can you please address some comments I made? Thanks!

integration_tests/select_rank_39.f90
srinjoy933: Hey @HarshitaKalani , made the changes that you have suggested ,please have a look at this now. Thanks!

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13208,7 +13215,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             ASR::array_physical_typeType::DescriptorArray, desc_type, nullptr
                         );
                         temp = ASRUtils::EXPR(array_cast);
-                    } else if (!(intrinsic_name == "size" || intrinsic_name == "lbound" || intrinsic_name == "ubound" || 
+                    } else if (!(intrinsic_name == "len" ||intrinsic_name == "size" || intrinsic_name == "lbound" || intrinsic_name == "ubound" ||
```
certik:
```suggestion
                    } else if (!(intrinsic_name == "len" || intrinsic_name == "size" || intrinsic_name == "lbound" || intrinsic_name == "ubound" || 
```

certik: So this adds `len`. Is this being tested? I don't see `len` in the above test?

integration_tests/select_rank_39.f90
```diff
@@ -0,0 +1,19 @@
...
+  subroutine s(a)
+    character, intent(in) :: a(..)
+
+    select rank(a)
+    rank(1)
+      if (a(1) /= 'A') error stop
+    end select
+  end subroutine s
+  
+end program select_rank_39
```
certik, line 21: Does this fail in main and now passes?

srinjoy933: yes @certik .

src/lfortran/semantics/ast_common_visitor.h
srinjoy933: I actually missed that part , so now I added len in the above test.

HarshitaKalani (approved):
lgtm. Thanks!

PR #12306, opened 2026-07-21
Build xfoil

certik:
With LFortran:
```console
==> Running test case: NACA 0012 at alpha = 2 deg (inviscid) ...
    Cp distribution written OK (     241 lines). First rows:
      #      x          Cp
           1.00000    0.42134
           0.99632    0.28738
           0.99156    0.24271
```
With GFortran:
```console
==> Running test case: NACA 0012 at alpha = 2 deg (inviscid) ...
    Cp distribution written OK (     241 lines). First rows:
      #      x          Cp
           1.00000    0.42121
           0.99632    0.28703
           0.99156    0.24324
```
There are some slight numerical differences, larger than numerical error, so probably we should track those down as well, but at least it compiles and runs pretty much correctly.


certik:
Split into:

* [x] #12308 
* [ ] #12309 
* [ ] #12310

PR #12307, opened 2026-07-22
Fix segfault: real(16) constant in procedure imported from module

### LINKED ISSUE #12291: Segfault when using real(16) in a procedure imported from a module
## MRE

`mre_real128_mod.f90`

```fortran
module m
  implicit none

contains

  subroutine s()
    real(16) :: xl

    xl = 1.0_16
  end subroutine s

end module m
```

`mre_real128_main.f90`

```fortran
program p
  use m, only: s
  implicit none

  call s()
end program p
```

## LFortran behavior

```text

Internal Compiler Error:

ASRUtils::real_constant_get_r16_bytes(&x)
```

## gfortran behavior

```text
gfortran: OK
```

usefahmed07:
## Summary
Fixes a segfault that occurs when a real(16) variable is assigned a literal constant inside a subroutine defined in a module, when that module is imported via `use` from another file.

## Root Cause
A RealConstant node with kind=16 doesn't store its value directly in m_r (a double). Instead, since a double can't hold a 128-bit float, LFortran allocates 16 bytes on the heap for the actual binary128 payload and packs the pointer to those bytes into m_r (see real_constant_get_r16_bytes / real_constant_pack_r16 in asr_utils.h).

The generated (de)serialization code (from asdl_cpp.py) treated m_r as a plain double in all cases, so it wrote/read the raw pointer bit-pattern when saving/loading .mod files. When a program did use module_name, the module's ASR was deserialized from disk, and the pointer restored from the .mod file no longer pointed to valid memory (the original allocator that owned those 16 bytes was gone). Dereferencing it in ASRToLLVMVisitor::visit_RealConstant during codegen caused a segfault.

### Before

module m
  implicit none
contains
  subroutine s()
    real(16) :: xl
    xl = 1.0_16
  end subroutine s
end module m

program p
  use m, only: s
  implicit none
  call s()
end program p

Internal Compiler Error:
ASRUtils::real_constant_get_r16_bytes(&x)

### After
Same code compiles and runs successfully:

$ lfortran mod.f90 main.f90 -o test_run
$ ./test_run
(exits cleanly, no crash)

## Fix
Modified src/libasr/asdl_cpp.py (the generator for asr_serialization_visitor.h / asr_deserialization_visitor.h) to special-case the RealConstant.r field:
- Serialization: if kind == 16, write the actual 16 payload bytes (via real_constant_get_r16_bytes) instead of the raw pointer bit-pattern; otherwise write the plain double as before.
- Deserialization: read the 16 raw bytes first (since m_type is only known after the next field is deserialized), then once m_type is available, either pack them into a freshly-allocated heap buffer in the current allocator (real_constant_pack_r16) for kind == 16, or interpret them as a plain double otherwise.

This keeps kind=4/kind=8 constants completely unaffected and only changes behavior for kind=16.

## Testing
Added integration_tests/module_real128_01.f90, reproducing the exact pattern from the issue (module subroutine assigning a real(16) literal, called from a program via use), registered in integration_tests/CMakeLists.txt.

Verified no regressions by running all existing real128_* integration tests (real128_arith_01-10, real128_literal_01, real128_int_to_real_01, real128_intrinsic_real_01, real128_intrinsic_inquiry_01) - all pass.

Fixes #12291


usefahmed07:
@assem2002 Can you review it

src/libasr/asdl_cpp.py
assem2002, line 1: `RealConstant(float r, ttype type)`

**Serialized like this** :=
- if kind == 16 -> `r` is a pointer -> De-reference  it -> and write 16 bytes.
- Otherwise `r` is plain double type -> write also 16 bytes. (8 BYTES WASTE)

**Deserialized like this :=**
- Go and read 16 bytes (POTENTIALLY 8 UNNECESSARY READS)
- after de-serializing `type`
- Make a decision (based on kind) whether we'll only consume 8 bytes or 16 bytes.

certik:
The principled way to fix this is to always first store the type, then the payload, where the size of the payload depends on the type.

One option is to flip the order in ASR itself:
```diff
--- a/src/libasr/ASR.asdl
+++ b/src/libasr/ASR.asdl
@@ -117,7 +117,7 @@ expr
     | UnsignedIntegerBitNot(expr arg, ttype type, expr? value)
     | UnsignedIntegerCompare(expr left, cmpop op, expr right, ttype type, expr? value)
     | UnsignedIntegerBinOp(expr left, binop op, expr right, ttype type, expr? value)
-    | RealConstant(float r, ttype type)
+    | RealConstant(ttype type, float r)
     | RealUnaryMinus(expr arg, ttype type, expr? value)
     | RealCompare(expr left, cmpop op, expr right, ttype type, expr? value)
     | RealBinOp(expr left, binop op, expr right, ttype type, expr? value)
```
The other option is to just flip this order in serialization/deserialization, and keep the ASR original. Either way should work.

We need special handling in serialization anyway (the 128 bits need to be allocated separately and use a pointer), so might as well just flip the order there to avoid a large change by changing the ASR. We can change the ASR later if we want to.

integration_tests/module_real128_01.f90
```diff
@@ -0,0 +1,17 @@
+module module_real128_01_mod
+    implicit none
+contains
+    subroutine set_value(x)
+        real(16) :: x
+        x = 1.0_16
+    end subroutine set_value
+end module module_real128_01_mod
```
certik: This test doesn't exercises the changes below. You must put the module in a separate file to trigger serialization.

You can easily discover this problem by testing this test without your patch, which you should always do.

usefahmed07:
Thanks For Review @certik 
Flipped serialization to write/read type before r (fixes both the ordering and byte-waste points), and split the test into separate module/program files via EXTRAFILES so it actually exercises serialization - all 14 real128_* tests pass.

integration_tests/module_real128_01.f90
usefahmed07: ok u are right

certik:
Several blockers:

* The test doesn't fail in main
* Wasteful 8 bytes in serialization for the type
* The Python generator changes are too complex and adhoc. Must simplify.

usefahmed07:
> Several blockers:
> 
>     * The test doesn't fail in main
> 
>     * Wasteful 8 bytes in serialization for the type
> 
>     * The Python generator changes are too complex and adhoc. Must simplify.

ok thanks

certik:
I fixed everything in https://github.com/lfortran/lfortran/pull/12331.

PR #12309, opened 2026-07-22
fix: COMMON block fixes (fixed-form tokenizer, char substring, separate-compilation init)

certik:
## Summary
Three independent fixes for COMMON block handling, grouped together as one thematic change (extracted from the `xfoil` branch).

## Scope
- Treat `/` inside parentheses as division when tokenizing fixed-form `COMMON` statements, so an expression like `COMMON /C/ A(20/4)` is no longer mis-tokenized (`fixed_form_common_01`).
- Route a single-character substring (`StringItem`) of a COMMON block `CHARACTER` variable at the COMMON struct member so it resolves to the shared storage (`common_substring_02`).
- Initialize COMMON block `CHARACTER` variables that are declared across separately compiled translation units (`separate_compilation_51`).

## Verification
Built locally with `-DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON`. The three integration tests compile and run successfully with `lfortran`:
- `fixed_form_common_01.f90` (`--fixed-form --implicit-typing`)
- `common_substring_02.f90`
- `separate_compilation_51.f90` (+ `_51a`, `_51b`, with `--implicit-interface --separate-compilation`)

All are registered with at least `LABELS gfortran llvm` and also pass with gfortran.

## Rationale
All three commits address distinct COMMON-block bugs and share the `COMMON` theme, so they are grouped into a single focused PR.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4553,6 +4553,7 @@ RUN(NAME separate_compilation_48 LABELS gfortran llvm EXTRAFILES separate_compil
 RUN(NAME separate_compilation_49 LABELS gfortran llvm EXTRAFILES separate_compilation_49a.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_50 LABELS gfortran llvm EXTRAFILES separate_compilation_50a.f90 separate_compilation_50b.f90 EXTRA_ARGS --separate-compilation)
 RUN(NAME separate_compilation_51 LABELS gfortran llvm EXTRAFILES separate_compilation_51a.f90 separate_compilation_51b.f90 EXTRA_ARGS --implicit-interface --separate-compilation)
+RUN(NAME separate_compilation_52 LABELS gfortran llvm EXTRAFILES separate_compilation_52a.f90 separate_compilation_52b.f90 EXTRA_ARGS --implicit-interface --separate-compilation)
 
 RUN(NAME no_explicit_return_type_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc
     EXTRA_ARGS --implicit-typing)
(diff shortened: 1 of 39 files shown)


certik:
Ok, so I think this is done incorrectly. The common block must inline everything in it, not use descriptors and we should not be fixing descriptors in the backend like this. So we need to rather rework the common block to truly inline things in it. Then it will work.

certik:
Split this PR into five smaller, self-contained PRs, each rebased onto current `main` (this branch had gone stale — `main` has since added its own `separate_compilation_51`, so the tests here are renumbered to 52/53):

**Independent, can be reviewed and merged in any order:**

| PR | Change | Size |
| --- | --- | --- |
| #12550 | `fix: treat / inside parentheses in fixed-form COMMON as division` | 3 files, +20/-2 |
| #12551 | `fix: point COMMON block char substring StringItem at struct member` | 3 files, +24/-1 |

**Stacked, in order (each based on the previous):**

| PR | Change | Size (net) |
| --- | --- | --- |
| #12552 | `refactor: centralize inline character struct member detection` — no behavior change | 3 files, +68/-95 |
| #12553 | `fix: lay out and index inline character array struct members correctly` — bind(C)/SEQUENCE, no COMMON involved | 15 files, +335/-231 |
| #12554 | `fix: store COMMON block characters inline instead of via descriptors` | 21 files, +121/-22 |

Two things changed relative to the branch here:

- **The `953c10d` runtime workaround is dropped entirely.** It added
  `setup_common_block_string_storage`, `emit_common_block_global_initializers`,
  `struct_has_character_members` and the `__lfortran_init_common_blocks`
  constructor, and `03508b8` deleted all of it again. Only the inline layout is
  kept, so no PR in the split carries added-then-removed code. Its test
  (`separate_compilation_51`, now `separate_compilation_52`) survives, under the
  inline implementation.
- **The `is_common_block_global` extraction is dropped.** It was a leftover from
  the workaround — nothing but `needs_common_linkage_for_global` called it.

The pre-existing bind(C)/SEQUENCE bug that #12553 fixes is now demonstrated on
its own, with no COMMON in the picture (`derived_types_154`): a `SEQUENCE` type
with a `character(len=4) :: names(3)` member segfaults on `main` because
`getStructType` sized the member as `[len x i8]` and ignored the element count.

Each branch was built and tested separately. Every PR's `-b llvm` and
`-b gfortran` integration passes are 100% green, `./run_tests.py --no-llvm` is
green, and #12552's reference outputs are byte-identical to `main`'s, confirming
it is a true no-op. The ASR reference updates in #12554 were regenerated locally
and match this branch's exactly.

---

certik:
It's just a test now.

Issue #11924, opened 2026-06-19
Idea-refactor : implicit-interface with implicit-procedure variables


assem2002:
General Idea : interface in `procedure() :: foo_var` is kinda of smartly deducing the required interface (to make the backend lowering easier and not erroneous), while it shouldn't be that strict -- See example below.

***
  
I can see that for variable like `procedure( ) , intent(in) :: x` -- We create an ASR::Function node that's an INTERFACE and we set `variable->m_declaration_sym = interface_func_SYMBOL` and  `variable->m_type = type_of_interface_func`.
Those are 2 different things. 
While visiting body we start mutating that on the go. and sometimes they get mismatched.
***
I think Variables whose type is implicit procedure `procedure( )` should act a bit different.

- That variable should be identified as implicit-unknown type `void*` -- to be set in `variable->m_type` .
- On call spots use `ASR::Cast` to cast that variable into the correct functionType (proper argument/returntype specified)
- At compileTime keep an interface that shapes the assumed compile-time-function-signature, so we can raise compile-time errors -- to be set in `variable->m_declaration_sym`

That way we freely invoke calls to the variable (even when interface isn't deduced) + we don't link the 2 variable members, they point to 2 unrelated things and the mutation of the interface (while we visit the body) doesn't hurt anyone, because noone will depend on it.

assem2002:
example see this
```.f90
program p
    call upper_caller(foo, 1)
    call upper_caller(boo, 2)
contains

    subroutine upper_caller(f, b)
        procedure() :: f
        integer :: b
        real :: a

        if(a == 1) then
          a = f()
        else
          a = f(2.0)
        end if
        print *, a
    end subroutine

    function foo() result(r)
      real :: r
      r  = 1010
      print *, "hello"
    end function

    function boo(m) result(r)
      real :: r
      real :: m
      r  = m
      print *, "hello"
    end function
end program
```
it runs with gfortran -- I'm not sure that's a correct fortran code or not, but still it runs properly and no interface conflicts (it feels like calls to the variable functions are not bound to an interface).

Here we should either raise a compile-time error OR make it run, unfortunately we don't do both now

assem2002:
related #11874

srinjoy933:
Thanks @assem2002 for this idea, working on it.

Jatinagarwal24:
> Thanks [@assem2002](https://github.com/assem2002) for this idea, working on it.

[...]

Issue #5273, opened 2024-11-06
Parser and AST: unify all declarations and statements


certik:
Right now in the parser we distinguish declarations before "implicit none" and declarations after. In AST we currently store "use" statements separately from the rest of the declarations. Also we separate declarations and body.

At least in the parser itself it seems we should parse all declaration and executable statements together and allow any order. Then
* either when we construct AST we separate them (and report errors if invalid order)
* or possibly we allow any order even in AST, and do the "order" check in the semantic phase in AST->ASR.

This change would allow to report good error messages if the order is incorrect, and it would also simplify the logic both in the parser itself, as well as in constructing the AST. We currently go over all statements already to separate them: https://github.com/lfortran/lfortran/blob/fb5bfc447bf71a2515a252c5709feabdcd813c74/src/lfortran/parser/semantics.h#L1512, so the new design should not be slower than what we have now.

In the AST itself we can consider merging `unit_decl1/2` into just `unit_decl`. We can keep use statements, declarations and executable statements separate in AST, but they would all be of type `unit_decl`.

pjh40:
It may be worth defining _declaration-construct_ and/or _specification-construct_ rules in bison, given how often they can occur, and it would make it easier to add the 2023 _enumeration-type-def_

certik:
Here is a follow-up cleanup: https://github.com/lfortran/lfortran/issues/12447

PR #12310, opened 2026-07-22
fix: implicit typing / implicit interface handling

certik:
## Summary
Several fixes for implicit typing / implicit interface handling, grouped together as one thematic change (extracted from the `xfoil` branch).

## Scope
- Reconcile an implicit external interface (from `EXTERNAL` under `--implicit-interface --implicit-typing`) with the actual `SUBROUTINE` definition in the same file, dropping the guessed return value so codegen emits valid IR (`implicit_interface_39`).
- Pass LFortran's string descriptor (not the raw bind(C) data pointer) for implicit-interface character-array dummy arguments (`implicit_interface_char_array_01`).
- Isolate implicitly typed locals from external procedures by tracking entry-master wrappers, so an implicit local is not mistaken for an external (`implicit_typing_13`).
- Mangle implicit external procedure names distinctly by introducing a new `deftypeType::ImplicitInterface` and using it for implicit-interface functions; updates many ASR/LLVM reference outputs (`external_17`).
- Allow `PARAMETER` statements before `IMPLICIT` statements in the parser (`implicit_typing_14`).

## Verification
Built locally with `-DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON` (including regenerating ASR/parser sources via `build0.sh`). The five integration tests compile and run successfully with `lfortran`:
- `implicit_interface_39.f90`, `implicit_interface_char_array_01.f90` (+ mod), `implicit_typing_13.f90`, `implicit_typing_14.f90`, `external_17.f90` (+ `.c`)

All are registered with at least `LABELS gfortran llvm` and also pass with gfortran. Updated `tests/reference/` outputs are included for the ASR/LLVM tests affected by the new `ImplicitInterface` deftype.

## Rationale
All five commits concern implicit typing / implicit interface support and share the new `ImplicitInterface` machinery and related semantics, so they are grouped into a single coherent PR.


certik:
Review findings:

**[blocker] Reconciliation skips program and module scopes**  
`src/lfortran/semantics/ast_body_visitor.cpp:10261`

Both walkers recurse only into `Function_t`. An implicit interface declared directly in a program or module is therefore never reconciled. Traverse every symbol-owning scope and add program- and module-scope regressions.

**[blocker] Definitions are matched by unqualified name**  
`src/lfortran/semantics/ast_body_visitor.cpp:10267`

`definitions[name]` conflates unrelated procedures in different scopes. A same-named internal subroutine can cause an external function interface elsewhere to lose its return value, violating the `FunctionCall` ASR invariant. Match canonical symbols and defining scopes, and validate complete signatures.

**[blocker] Invalid `IMPLICIT` ordering is silently normalized**  
`src/lfortran/parser/parser.yy:1233`  
`src/lfortran/parser/semantics.h:1811`

`program_unit_statements` permits executable statements before `IMPLICIT`; `split_implicit_statements()` then erases the original order. For example:

```fortran
program p
  integer :: x
  x = 1
  implicit real(a-z)
end
```

This is normalized as though the `IMPLICIT` statement preceded the execution part. Preserve ordering and diagnose invalid placement in semantics while supporting the intended `PARAMETER`-before-`IMPLICIT` case.

**[rework] ENTRY provenance depends on generated names**  
`src/lfortran/semantics/ast_common_visitor.h:2351`

Master/wrapper relationships are reconstructed from `_main__lcompilers`, `entry__lcompilers`, locations, and dependency strings. Represent this relationship explicitly in ASR and verify it.

**[rework] `ImplicitInterface` conflicts with the ASR design**

All interfaces should be explicit in ASR. In this PR, `ImplicitInterface` does not describe an incomplete ASR signature; it records that the frontend synthesized the interface from a source-level implicit external procedure. It is then used mainly for mangling, linkage, and same-name definition lookup.

That distinction should not be another `deftype`. ASR should contain a fully explicit `FunctionType` with `deftype=Interface`. If unresolved Fortran linkage or source provenance is genuinely required downstream, represent that property independently, preferably through explicit linkage/ABI state such as `ExternalUndefined` or a narrowly named provenance field. Once matched to a definition, canonicalize the symbol and discard the provenance. Backends should not need to know whether the source originally omitted an explicit interface.

**Architecture assessment:** `ImplicitInterface` is currently mixing interface completeness with source provenance/linkage. The unresolved correctness problems are earlier: parser/AST ordering and AST-to-ASR symbol identity. The LLVM/backend changes should consume canonical explicit interfaces mechanically.

certik:
The parser bug is an old issue: https://github.com/lfortran/lfortran/issues/925

I think a better way to implement it rather than hacking the parser like we did in this PR is to do https://github.com/lfortran/lfortran/issues/5273.

integration_tests/implicit_interface_39.f90
Jatinagarwal24, line 1: This is not failing in current main.

integration_tests/implicit_typing_14.f90
Jatinagarwal24, line 1: THis is also not failing.

certik:
The parsing issue is already fixed.

src/libasr/ASR.asdl
```diff
@@ -239,7 +239,7 @@ cast_kind = RealToInteger | IntegerToReal | LogicalToReal | RealToReal | Integer
 storage_type = Default | Save | Parameter
 access = Public | Private
 intent = Local | In | Out | InOut | ReturnVar | Unspecified
-deftype = Implementation | Interface
+deftype = Implementation | Interface | ImplicitInterface
```
certik, line 242: I don't like this. I think this is related to https://github.com/lfortran/lfortran/issues/11924.

certik:
I would extract the MREs for the implicit issue, and figure out a good solution in https://github.com/lfortran/lfortran/issues/11924. The parser bug is fixed. If there is any other bug here, we can send a separate PR.

We should also consolidate the MREs for implicit typing.

PR #12314, opened 2026-07-22
fix: allow no whitespace after parenthesis in function like `#define`

### LINKED ISSUE #12313: Preprocessor fails when a #define macro body starts immediately after the closing parenthesis
MRE
```
#define FOO(x)\
print *, x; \
print *, x;
program main
implicit none
FOO('hello')
end program
```
Lfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# lfortran examples/expr2.f90 --cpp
warning: #define ignored
 --> examples/expr2.f90:1:1
  |
1 | #define FOO(x)\
  | ^^^^^^^^^^^^^^^^ help: use the '--cpp' command line option to preprocess it

tokenizer error: Token '\' is not recognized
 --> examples/expr2.f90:2:13
  |
2 | print *, x; \
  |             ^ token not recognized


```
Gfortran 
```
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# gfortran examples/expr2.f90 -cpp && ./a.out
 hello
 hello
(lf) root@LAPTOP-8SLTQKJO:~/lfortran# 
```

Jatinagarwal24:
fixes #12313

The function-like macro rules required whitespace between the closing
parenthesis and the macro body, so a definition like

    #define FOO(x)\
    print *, x

matched no rule, fell through to the "#define ignored" warning, and the
stray `\` then caused a tokenizer error. Make the whitespace optional in
both the zero-arg and named-arg rules, always capture the body (fixing
`token(t3, t4)` on unset captures when the old optional group didn't
match), and call handle_continuation_lines in the zero-arg rule so
multi-line bodies are stitched together there as well.

Add integration test cpp_pre_18 covering continuation lines and bodies
starting immediately after `)`, verified against gfortran -cpp.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4091,6 +4091,10 @@ RUN(NAME cpp_pre_17 LABELS gfortran llvm
     EXTRA_ARGS --cpp
     GFORTRAN_ARGS -cpp)
 
+RUN(NAME cpp_pre_18 LABELS gfortran llvm
+    EXTRA_ARGS --cpp
+    GFORTRAN_ARGS -cpp)
+
 RUN(NAME dabs_01 LABELS gfortran llvmImplicit)
 
 RUN(NAME minpack_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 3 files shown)


HarshilShah1804 (approved):
Looks good.

PR #12315, opened 2026-07-22
fix: prioritize type-mismatch error over wrong-intent error (#12312)

### LINKED ISSUE #12312: Error: Error message should priortize type-mismatch error
# PLEASE LOOK AT LINKED COMMENT BELOW.

> We should see a type mismatch error.
> Raising this error is still kinda valid, but it's quite misleading for the end user.
> ***
> ## What we want?
> We want to prioritize the "type mismatch error" (i.e. we want to see a type mismatch error here, rather than a wrong-intent error)
>  ***
> ## What to do? 

 _Originally posted by @assem2002 in [#12222](https://github.com/lfortran/lfortran/pull/12222/changes#r3632444500)_

usefahmed07:
Fix #12312

## Problem

When a procedure is passed positionally to an intent(out)/intent(inout) dummy argument, the compiler raised a misleading error:

call set_caller(upper_caller)

Before:
semantic error: Non-variable expression in variable definition context (actual argument to INTENT = OUT/INOUT)

This is technically true but confusing — the real issue is that upper_caller is a procedure, not a value of the expected type. The user needs to know the type is wrong, not that it's "non-variable".

## Fix

In check_simple_intent_mismatch, before checking whether the argument is modifiable, we now check whether it even resolves to a variable at all. If it resolves to something else (like a procedure symbol), we report a type mismatch instead.

After:
semantic error: Type mismatch: expected `runner` but a procedure was passed

## Testing

Updated tests/errors/intent_inout_missing_arg.f90 reference output. Ran the full test suite locally, all passing.


usefahmed07:
@assem2002 Done

opixdown (approved):

opixdown:
LGTM, thanks

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
```diff
@@ -1,4 +1,4 @@
-semantic error: Non-variable expression in variable definition context (actual argument to INTENT = OUT/INOUT)
+semantic error: Type mismatch: expected `runner` but a procedure was passed
```
certik, line 1: Please move this file to continue compilation.

src/libasr/asr_utils.h
```diff
@@ -7964,6 +7964,26 @@ inline void check_simple_intent_mismatch(diag::Diagnostics &diag, ASR::Function_
...
+                                    std::string param_type_str = ASRUtils::type_to_str_with_kind(
+                                        callee_param->m_type, f->m_args[i]);
+                                    diag.add(diag::Diagnostic(
+                                        "Type mismatch: expected `" + param_type_str +
+                                        "` but a procedure was passed",
+                                        diag::Level::Error, diag::Stage::Semantic, {
+                                            diag::Label("", {passed_arg_expr->base.loc})
+                                        }));
+                                    throw SemanticAbort();
+                                }
```
certik: This function is checking intent mismatch. Not a type mismatch. I think this error should be caught somewhere else.

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
usefahmed07: ok no problem

usefahmed07:
Done @certik

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
HarshitaKalani: I don't see this file has still been moved to continue_compilation_*.f90 files. Can you please do that and mark the PR as review after that? Thanks!

HarshitaKalani (commented):
Can you please move the error test to continue_compilation_*.f90 file? Thanks!

usefahmed07:
@HarshitaKalani I cleaned up the branch history (rebased on top of the latest upstream/main and dropped the extra merge commits), so the PR is now down to 3 commits:

- fix: prioritize type-mismatch error over wrong-intent error (#12312)
- Move type-mismatch check out of check_simple_intent_mismatch
- Move intent_inout_missing_arg test into continue_compilation_2

The error test has also been moved into `tests/errors/continue_compilation_2.f90` as requested. Let me know if there's anything else you'd like changed

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7800,6 +7799,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                            ASR::down_cast<ASR::Var_t>(args[i].m_value)->m_v);
+                        if (!ASR::is_a<ASR::Variable_t>(*passed_sym_check)) {
+                            std::string param_type_str = ASRUtils::type_to_str_with_kind(
+                                dummy_type, f->m_args[i]);
+                            diag.add(diag::Diagnostic(
+                                "Type mismatch: expected `" + param_type_str +
+                                "` but a procedure was passed",
+                                diag::Level::Error, diag::Stage::Semantic, {
+                                    diag::Label("", {args[i].m_value->base.loc})
+                                }));
```
HarshitaKalani, line 7854: I don't see this error in the updated references. Can you please check this?

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
```diff
@@ -1,5 +0,0 @@
-semantic error: Non-variable expression in variable definition context (actual argument to INTENT = OUT/INOUT)
```
HarshitaKalani, line 1: I don't see the same error in the continue_compilation_* references

tests/errors/continue_compilation_2.f90
```diff
@@ -571,6 +572,10 @@ logical function f(x)
         f = PRESENT(x)
     end function
 
+    subroutine upper_caller(a)
+        class(*), intent(in) :: a
+    end subroutine
```
HarshitaKalani: Can you please add this in a way that the references diff is minimal? Currently I see the updated references diff is very large and it contains unwanted changes as well.

HarshitaKalani (dismissed):
Can you please address the comments and mark it as ready for review once done? Thanks!

tests/reference/asr-intent_inout_missing_arg-15600a9.stderr
usefahmed07: u're right moving it there caused a different error path to trigger instead ("More actual than formal arguments"). I've reverted that move and kept intent_inout_missing_arg as its own standalone test file to avoid the interaction

src/lfortran/semantics/ast_body_visitor.cpp
usefahmed07: This message does appear correctly for the standalone test — I've reverted moving intent_inout_missing_arg into continue_compilation_2 since that move was causing a different, unintended error to surface there. Kept it as its own test file instead

tests/errors/continue_compilation_2.f90
usefahmed07: Agreed  I ended up reverting the move of intent_inout_missing_arg into continue_compilation_2 entirely, since it was both enlarging the diff and changing the actual error reported. It's back to being a standalone test now

usefahmed07:
@HarshilShah1804 pls review it

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7837,6 +7836,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
...
+                        args[i].m_value && ASR::is_a<ASR::Var_t>(*args[i].m_value)) {
+                        ASR::symbol_t* passed_sym_check = ASRUtils::symbol_get_past_external(
+                            ASR::down_cast<ASR::Var_t>(args[i].m_value)->m_v);
+                        if (!ASR::is_a<ASR::Variable_t>(*passed_sym_check)) {
+                            std::string param_type_str = ASRUtils::type_to_str_with_kind(
+                                dummy_type, f->m_args[i]);
+                            diag.add(diag::Diagnostic(
+                                "Type mismatch: expected `" + param_type_str +
+                                "` but a procedure was passed",
+                                diag::Level::Error, diag::Stage::Semantic, {
```
assem2002: I'm not a fan of this. Here things feel a bit hard coded.
I created this PR #12412 trying to make the compiler just use our mechanism of detecting arg/param mismatch.
Changes here seem minimal, compared to #12412, but I tried to just fallback to the normal mismatch mechanism without adding any additional error checks

usefahmed07: Thanks for pointing this out, and for opening #12412! I checked it out I get the concern about hardcoding a separate check instead of letting the existing arg/param mismatch mechanism handle it.
Since you mentioned only one of these should be merged, how do you want to proceed? Should I close this PR (#12315) and let #12412 take over, or would you rather I rework this one to follow the same approach (moving the check into the general mismatch loop instead of hardcoding it here)?

Happy to go either way, just want to make sure we don't duplicate effort.

assem2002: both are kinda hacking their way in. 
let's Proceed with this one.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7837,6 +7836,29 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     ASR::Variable_t* dummy_var = ASR::down_cast<ASR::Variable_t>(dummy_sym);
                     ASR::ttype_t* dummy_type = dummy_var->m_type;
                     ASR::ttype_t* actual_type = ASRUtils::expr_type(args[i].m_value);
+                    // Check whether the passed argument is even a compatible
+                    // kind of thing (e.g. a procedure symbol passed where a
+                    // variable was expected). That's a type mismatch, and
+                    // should be reported as such -- not as a misleading
+                    // "wrong intent" error.
```
assem2002:
```suggestion
                    // TODO: Dump this check and depend on our param/arg mismatch mechanism below
```

usefahmed07: Very Good Now iam Waiting CI

Issue #12319, opened 2026-07-23
LCompilersException: Stop Code should be of type [String, Integer].


YuanchengJiang:
The following code:

```fortran
subroutine exit_pointer(status)
  call exit(status)
end subroutine
```

Resulted in this output:
```
warning: Routine `exit` is a non-standard function
 --> /tmp/test.f90:2:3
  |
2 |   call exit(status)
  |   ^^^^^^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/opt/lfortran/bin/lfortran", in _start
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in __libc_start_main
  Binary file "/lib/x86_64-linux-gnu/libc.so.6", in <invalid>
  File "/home/fuzz/WorkSpace/fusion-fuzz/projects/lfortran/lfortran-src/src/bin/lfortran.cpp", line 3000, in main
    return main_app(argc, argv);
  File "/home/fuzz/WorkSpace/fusion-fuzz/projects/lfortran/lfortran-src/src/bin/lfortran.cpp", line 2913, in main_app(int, char**)
    err = compile_src_to_object_file(arg_file, tmp_o, compiler_options.time_report, false,
  File "/home/fuzz/WorkSpace/fusion-fuzz/projects/lfortran/lfortran-src/src/bin/lfortran.cpp", line 1293, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool, bool*)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
  File "/home/fuzz/WorkSpace/fusion-fuzz/projects/lfortran/lfortran-src/src/lfortran/fortran_evaluator.cpp", line 436, in LCompilers::FortranEvaluator::get_llvm3(LCompilers::ASR::TranslationUnit_t&, LCompilers::PassManager&, LCompilers::diag::Diagnostics&, LCompilers::LocationManager&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int*)
[...]

PR #12320, opened 2026-07-23
Fix: report semantic error for STOP/exit code of wrong type (#12319)

usefahmed07:
**Fixes** #12319


tests/tests.toml
```diff
@@ -505,6 +505,9 @@ asr = true
 [[test]]
 filename = "errors/real_unsupported_kind_01.f90"
 asr = true
+[[test]]
+filename = "errors/stop_code_wrong_type_01.f90"
+asr = true
```
certik: Use continue compilation.

usefahmed07: ok thanks

opixdown (approved):

opixdown:
@usefahmed07 LGTM! just move the test to continue compilation. We generally add semantic error tests under continue compilation while keeping the diff minimum. Also, if you are using VS Code, you can use the source control tab to review exactly what youare committing before pushing. Thankyou

opixdown:
remove blank line for every line you add thats the way to keep the minimal diff

tests/errors/continue_compilation_1.f90
```diff
@@ -657,6 +657,8 @@ program continue_compilation_1
     integer :: eoshift_bad_shift(2, 2)
     eoshift_bad_shift = 1
     b1 = eoshift(b1, eoshift_bad_shift)
+    stop_code_wrong_type_01_x = 5.0
```
opixdown: for this remove two lines if no space is there try to do it in subroutine and append it in eof

HarshitaKalani: Can you please address this @usefahmed07 ? This is to make the reference diff small so that it's easy to review and we make sure there're no unwanted changes.

HarshitaKalani (commented):
Can you please address the comment and then mark it ready for review?

tests/errors/continue_compilation_1.f90
```diff
@@ -657,6 +657,8 @@ program continue_compilation_1
     integer :: eoshift_bad_shift(2, 2)
     eoshift_bad_shift = 1
     b1 = eoshift(b1, eoshift_bad_shift)
+
+
```
opixdown: remove this two line

usefahmed07: Done

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -7697,6 +7706,21 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
             }
         }
         if (!original_sym || (original_sym && is_external)) {
+            if (to_lower(sub_name) == "exit") {
+                diag.semantic_warning_label(
+                    "Routine `" + sub_name + "` is a non-standard function",
+                    {x.base.base.loc},
+                    ""
+                );
```
HarshitaKalani, line 7736: This error is not tested. Can you please add a test for this too?

HarshitaKalani (commented):
There are two errors added in this PR but only one is tested. Can you please add a test for the other error as well? Thanks!

src/lfortran/semantics/ast_body_visitor.cpp
usefahmed07: ok no problem i am on it

usefahmed07:
Hi @HarshitaKalani can you review it thanks!

usefahmed07:
@assem2002

PR #12323, opened 2026-07-23
fix: treat parameter-array element as constant array bound

### LINKED ISSUE #12322: parameter-array element rejected as array bound ("nonconstant bounds")
## MRE

```fortran
program p
  integer, parameter :: dims(2) = [3, 4]
  real :: a(dims(1), dims(2))
  a = 0.0
  print *, size(a)
end program
```

## LFortran behavior

```text
$ lfortran t.f90
semantic error: Explicit shaped array with nonconstant bounds
 --> t.f90:3:13
  |
3 |   real :: a(dims(1), dims(2))
  |             ^^^^^^^
```

## gfortran behavior

```text
$ gfortran t.f90 && ./a.out
          12
```

## notes

An element of a PARAMETER (named-constant) array is a constant expression (F2018 10.1.12) and may be used as an array bound. LFortran accepts a scalar parameter (`integer, parameter :: n = 4; real :: a(n)`) but rejects `dims(1)` where `dims` is a parameter array.

The same rejection hits every form of parameter-array-element bound:

```fortran
integer, parameter :: dims(2) = [3, 4]
real :: a(dims(1))              ! rejected
real :: b(dims(2))              ! rejected
type :: t; real :: x(dims(1)); end type t   ! rejected (component bound)
```

Curiously, a parameter-array element used as a character length works:

```fortran
integer, parameter :: n(1) = [5]
character(len=n(1)) :: s        ! accepted, len(s) == 5
```

so the constant value is already being folded for `dims(1)` (its ASR `ArrayItem` carries `m_value`), but `is_value_constant` in `src/libasr/asr_utils.h` has no case for `ArrayItem`, so it returns false and the bound is treated as nonconstant.

ambikeesshh:
fixes #12322

`real :: a(dims(1))` was rejected as "nonconstant bounds" when `dims` is a parameter array, even though `dims(1)` is a constant expression and its value was already being folded.

the gap was just recognition: `is_value_constant` and `extract_value` in `asr_utils.h` had no case for `ArrayItem`, so they fell through to false. added `ArrayItem` to the two switch groups that already delegate to `expr_value` (same pattern as `IntegerBinOp` / `StringLen`). no new folding logic.

```
$ lfortran t.f90   # integer, parameter :: dims(2) = [3,4]; real :: a(dims(1),dims(2)); print *, size(a)
          12
```

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -774,6 +774,7 @@ RUN(NAME array_bound_4 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackAr
 RUN(NAME array_bound_5 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME array_bound_6 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc EXTRA_ARGS --realloc-lhs-arrays fortran)
 RUN(NAME array_bound_7 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME array_bound_8 LABELS gfortran llvm) # parameter-array element as array bound
 RUN(NAME arrays_op_1 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc llvmStackArray wasm fortran)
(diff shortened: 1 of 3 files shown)


certik (approved):
I think that's fine.

opixdown (approved):

PR #12329, opened 2026-07-24
fix: make `call exit` silent and evaluate stop codes only once

### LINKED ISSUE #12324: Type-bound procedure is erroneously executed twice from within an `exit` subroutine call
`$ cat test.f90`
```
module testmod

   type :: tester
   contains
      procedure, nopass :: test
   end type tester

contains
   
   function test() result(retcode)
      integer :: retcode
      write(*,*) 'Setting return code!'
      retcode = 0
   end function test
   
end module testmod

program check
   
   use testmod
   
   type(tester) :: obj
   
   call exit(obj%test())
   
end program check

```
`$ flang-20 test.f90 -o test; ./test`
 ```
Setting return code!
```
`$ lfortran --no-warnings test.f90`
```
Setting return code!
STOP 0
Setting return code!
```

Jatinagarwal24:
fixes #12324
Previously `call exit(n)` was lowered to an ASR `Stop` node, which
printed "STOP <n>" to stderr, unlike gfortran's silent exit. In
addition, the LLVM backend evaluated the stop code expression twice
(once for the message, once for the exit code), so a function call
with side effects like `call exit(obj%test())` ran twice.

- Implement `exit` as an IntrinsicImpureSubroutine (same pattern as
  `abort`/`sleep`) backed by a new `_lfortran_exit(status)` runtime
  function that finalizes allocations and exits silently with the
  given status; `call exit` with no argument exits with status 0.
- Remove the old `exit` -> `Stop` special case in the body visitor.
- In the LLVM backend, evaluate the code of `STOP`/`ERROR STOP`
  exactly once and reuse the value for both the printed message and
  the process exit code (including the --debug-info path).
- Add integration test stop_04 that fails if the exit code
  expression is evaluated more than once.

Fixes the case:

    call exit(obj%test())   ! test() ran twice and printed "STOP 0"


HarshitaKalani (approved):
lgtm. Thanks!

PR #12332, opened 2026-07-25
parser: unify declarations and statements into one list

### LINKED ISSUE #5271: Out-of-order USE statement misdiagnosed
```f90
program a
  integer b
  use, intrinsic :: iso_c_binding
end program
```
results in the error:
```sh
 --> 5250.f90:3:6
  |
3 |   use, intrinsic :: iso_c_binding
  |      ^ 
```
rather than reporting that `USE` statements must appear first in a _specification-part_.  Ordering the two statements correctly allows compilation.  This is on `main`, version `0.41.0-224-g11df6a825`.

### LINKED ISSUE #5272: FORMAT statement in implicit-part causes error
```f90
program a
100 format ("I0")
  implicit none
  integer b
end program
```
causes the error:
```sh
syntax error: Token 'none' is unexpected here
 --> 5250.f90:3:12
  |
3 |   implicit none
  |            ^^^^ 
```
on `main` branch, version `0.41.0-224-g11df6a825`.  F2018 standard R505 and R506 allow _implicit-stmt_ to be mixed with _parameter-stmt_ and _format-stmt_ at will.

certik:
## Summary

The Bison grammar used to encode the ordering of a specification part directly
in the rule structure of every program unit, via the `use_statement_star`,
`import_statement_star`, `implicit_statement_star` and `decl_star` prefixes,
repeated across 16 productions. Any out-of-order statement therefore produced an
unhelpful `syntax error: Token 'x' is unexpected here` (#5271), and statements
that F2018 R506 allows in the *implicit-part* (`format`, `entry`,
`parameter`) could not precede `implicit` (#5272).

All five kinds of statements are now collected into the single `decl_statements`
list. `split_decl_stmts()` in `semantics.h` splits that list back into the
`use`, `import`, `implicit`, `decl` and `body` lists of the AST and reports any
ordering or context violation as a parser error. `grammar/AST.asdl` is
unchanged; merging `unit_decl1`/`unit_decl2` into a single `unit_decl` is left
for a follow-up, as suggested in #5273.

Before:

```
 --> a.f90:3:6
  |
3 |   use, intrinsic :: iso_c_binding
  |      ^
```

After:

```
syntax error: use statement must appear before all other declarations and statements
 --> a.f90:3:3
  |
3 |   use, intrinsic :: iso_c_binding
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Scope

The set of accepted programs is unchanged, except that `format`, `entry` and
`parameter` statements may now precede `implicit` (F2018 R505, R506).
Every ordering or context violation that used to be a syntax error is now a
clear parser diagnostic:

* `use statement must appear before all other declarations and statements`
* `import statement must appear before implicit and declaration statements`
* `implicit statement must appear before declaration and executable statements`
* `import statement is not allowed in a main program`
* `implicit statement is not allowed in a block construct`
* `executable statement is not allowed in a module`

Diagnostics are non-fatal, so `--continue-compilation` keeps working and all
violations are reported in one run.

Deliberately out of scope, to avoid rejecting anything that is accepted today:

* a declaration after an executable statement is still silently reordered
* `block data` still accepts executable statements

The `.yy` actions keep their previous one-line shape; all of the splitting and
checking lives in `semantics.h`. The unified grammar also drops 29 shift/reduce
conflicts (241 -> 212) and remains free of essential conflicts, i.e.
`ci/grammar_conflicts.sh` still shows the patched grammar is LALR(1).
`ci/parser.yy.patch` was regenerated with `ci/generate_lalr1_patch.py`.

## Verification

* `./run_tests.py` — `TESTS PASSED`. Every existing reference output is
  byte-identical, which is the main regression signal since the AST shape is
  preserved.
* `cd integration_tests && ./run_tests.py -j16` — `100% tests passed, 0 tests
  failed out of 3821`.
* `cd integration_tests && ./run_tests.py -b gfortran` — the new
  `decl_order_01` passes; the only failures are pre-existing and unrelated to
  this change.
* LALR(1) check with the regenerated `ci/parser.yy.patch`: no conflicts.

New tests:

* `integration_tests/decl_order_01.f90` (labels `gfortran llvm`) — `format`
  before `implicit` in a module procedure and in a main program. Fails to
  compile on `main`, compiles and runs on this branch, and is accepted by
  GFortran with `-std=f2018`.
* `tests/errors/decl_order_01.f90` — covers all six new diagnostics. On `main`
  this file produces a single unhelpful `Token 'x' is unexpected here`.

## Rationale

`decl_statements` already mixed declarations and executable statements and was
split afterwards in `semantics.h`, and `fix_program_without_program_line()` in
`parser.cpp` already performs exactly this flat-list split for source without a
`program` line. Extending that mechanism to `use`/`import`/`implicit` removes
four near-duplicate list nonterminals, makes the grammar simpler and smaller,
and is what makes good diagnostics possible at all: the parser can only explain
that a statement is out of order if it is able to parse it in the first place.

Fixes #5271
Fixes #5272


tests/tests.toml
```diff
@@ -4963,6 +4963,14 @@ asr = true
 filename = "errors/data_implied_do3.f90"
 asr = true
 
+[[test]]
+filename = "errors/decl_order_01.f90"
+asr = true
+
+[[test]]
+filename = "errors/module_error_recovery_01.f90"
+semantics_only_cc = true
```
certik, line 4972: We'll move these into continue compilation next after this is merged.

certik: https://github.com/lfortran/lfortran/pull/12341

PR #12334, opened 2026-07-25
fix: respect delim= option for character namelist output

### LINKED ISSUE #12145: Namelist problem with delim= option
[nml_quotes_bug.txt](https://github.com/user-attachments/files/29821164/nml_quotes_bug.txt)

srinjoy933:
Fixes #12145

### DIFF
--- a/integration_tests/namelist_37.f90
+++ b/integration_tests/namelist_37.f90
@@ -5,23 +5,37 @@ program namelist_37
     character(len=256) :: text
     namelist /nums/ msg
 
-    open(unit=10, file="namelist_34_quote.dat", status="replace", &
+    ! --- QUOTE TEST ---
+    open(unit=10, file="namelist_37_quote.dat", status="replace", &
         form="formatted", delim="quote")
     write(10, nml=nums)
     close(10)
 
-    text = compact(read_file("namelist_34_quote.dat"))
+    text = compact(read_file("namelist_37_quote.dat"))
     if (index(text, 'msg="12345678901234567890"') == 0 .and. &
             index(text, 'MSG="12345678901234567890"') == 0) then
         error stop "quote-delimited namelist output mismatch"
     end if
 
-    open(unit=10, file="namelist_34_none.dat", status="replace", &
+    ! --- APOSTROPHE TEST ---
+    open(unit=10, file="namelist_37_apostrophe.dat", status="replace", &
+        form="formatted", delim="apostrophe")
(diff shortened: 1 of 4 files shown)


srinjoy933:
@assem2002 ,please have a look at this pr. Thanks!

integration_tests/namelist_37.f90
```diff
@@ -0,0 +1,64 @@
...
+    close(10)
+
+    text = compact(read_file("namelist_34_quote.dat"))
+    if (index(text, 'msg="12345678901234567890"') == 0 .and. &
+            index(text, 'MSG="12345678901234567890"') == 0) then
+        error stop "quote-delimited namelist output mismatch"
+    end if
+
+    open(unit=10, file="namelist_34_none.dat", status="replace", &
+        form="formatted", delim="none")
```
opixdown, line 34: we should also add test for `delim='apostrophe'`

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -13852,6 +13865,7 @@ LFORTRAN_API void _lfortran_namelist_write_str_array(
     w.pos = 0;
 
     w.is_file = false;
+    w.delim = 0;
```
opixdown, line 13868: Why is w.delim hardcoded to 0 here?

integration_tests/namelist_37.f90
```diff
@@ -0,0 +1,64 @@
+program namelist_37
+    implicit none
+
+    character(len=20) :: msg = "12345678901234567890"
+    character(len=256) :: text
+    namelist /nums/ msg
+
+    open(unit=10, file="namelist_34_quote.dat", status="replace", &
```
opixdown: the file should be `namelist_37....`

opixdown (approved):

opixdown:
LGTM! just address the comments thank you

src/libasr/runtime/lfortran_intrinsics.c
srinjoy933: It is hardcoded to 0  here because _lfortran_namelist_write_str_array handles internal character array writes. Currently, the compiler frontend doesn't pass a delim argument to this specific runtime function signature. Without explicitly initializing it, w.delim would contain uninitialized memory, potentially causing random quote characters to be injected into internal string writes

HarshitaKalani (approved):

PR #12348, opened 2026-07-26
feat: add --show-gpu-kernel-source to emit registrar-free GPU kernels

Zaneham:
Hi, 

I maintain [Booth](https://github.com/Zaneham/Booth), an open source GPU
compiler, and I've been spending my time lately on getting Fortran to run on
GPUs. I've also done a wee bit in the legacy fortran space within LFortran too.

I ran into this while wiring LFortran up to Booth. The plan was to let LFortran
handle the Fortran and hand the generated kernels over for a non-NVIDIA target,
so I went looking at what `--gpu=cuda` emits. The kernels are good. What gets in
the way is the registration block on the end of every generated `.cu`. It
declares and calls `lfortran_gpu_register_kernel`, so the emitted source is not
self-contained. Compile it on its own and you have an undefined symbol unless
you also link LFortran's CUDA runtime. That block is host side glue for the
`cudaLaunchKernel` path, which finds kernels by function pointer, and anything
resolving by name has to discard it.

The seam is already mostly there, mind. `lfortran_gpu_runtime.h` is eight
functions with no CUDA types in sight, which I assume is the Metal backend
having forced the issue. This is really just finishing that thought.

So this adds `--show-gpu-kernel-source`, which prints the kernel source for
whichever backend `--gpu` selected and then exits, following the same shape as
`--show-c` and `--show-julia`. The registrar is left out, since anything
reaching for this flag is not going through the CUDA runtime.

I tried making the `.cuda.cu` sidecar registrar-free first. That breaks separate
compilation, because the link step reads those sidecars back in and the runtime
resolves kernels through exactly that registry. So this is opt-in and the
compile and link paths are untouched. `asr_to_cuda` takes an `emit_registrar`
parameter defaulting to true, so the existing call site behaves as before.

Testing is a reference test on the CUDA output plus an integration test running
under gfortran and llvm. I also checked the metal path emits, and that passing
the flag without `--gpu` errors sensibly. No CUDA toolkit on this machine, so I
haven't run the output through nvcc and haven't labelled the new test `cuda` for
that reason.

Cheers!
Zane

### DIFF
--- a/doc/man/lfortran.md
+++ b/doc/man/lfortran.md
@@ -52,6 +52,7 @@ LFortran is a modern interactive Fortran compiler based on LLVM.
 - `--show-asm`: Show assembly for the given file and exit
 - `--show-wat`: Show WAT (WebAssembly Text Format) and exit
 - `--show-julia`: Show Julia translation source for the given file and exit
+- `--show-gpu-kernel-source`: Show the GPU kernel source for the backend selected by `--gpu` and exit
 - `--show-fortran`: Show Fortran translation source for the given file and exit
 - `--show-stacktrace`: Show internal stacktrace on compiler errors
 - `--symtab-only`: Only create symbol tables in ASR (skip executable stmt)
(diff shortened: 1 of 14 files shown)


certik:
Thanks! I'll review it soon.

We have to rework how we generate GPU code, we started as a subset of Fortran but realized we really have to treat GPU as a full target kind of like a CPU, and be able to lower anything (except possibly some IO stuff), so we need to figure out how to restructure our CPU lowering so that it can be shared with GPUs.

certik (approved):
All tests pass, the PR is a little bit older, but hopefully nothing broke since then, I am going to merge it.

PR #12349, opened 2026-07-26
make the GPU device compiler configurable via --device-compiler

Zaneham:
Follow-up to my other GPU PR, same background. I maintain
[Booth](https://github.com/Zaneham/Booth) and I've been getting Fortran running on GPUs.

The CUDA link path shells out to `nvcc` three times through `system()`, all hardcoded. Once for the generated kernel, once for `lfortran_gpu_cuda.cu`, and once more as the link driver. That's fine when `nvcc` is on PATH, less fine otherwise. If you have CUDA somewhere unusual, or several versions installed, or you want to point at a wrapper, there's currently no way to say so.

So this adds `--device-compiler`, defaulting to `nvcc`, and uses it for all three. Same idea as the existing `--linker` option.

However, there's no automated test here, because the path needs both nvcc and a GPU to reach, and there isn't any existing coverage of it either. I verified it by hand with a stub compiler on PATH, checking the default is still `nvcc` and that the override actually gets executed rather than just accepted. Happy to add something if you can suggest a way to test it that fits the suite.

Cheers!
Zane

### DIFF
--- a/doc/man/lfortran.md
+++ b/doc/man/lfortran.md
@@ -63,6 +63,7 @@ LFortran is a modern interactive Fortran compiler based on LLVM.
 - `--continue-compilation`: Collect error messages and continue compilation after encountering semantic errors
 - `--error-format TEXT=human`: Control how errors are produced (human, short)
 - `--backend TEXT=llvm`: Select a backend (llvm, cpp, x86, wasm, fortran)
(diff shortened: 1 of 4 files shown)


certik (approved):
This is good. Yes, it's hard to test this at the CI, we currently test CUDA code by emulating the cuda headers. We could test nvcc by having a fake `nvcc` "binary" that would use our CUDA API emulator to just test that things compile+link+run on a CPU.

PR #12351, opened 2026-07-26
semantics: check that a type-bound procedure override has a compatibl…

### LINKED ISSUE #12311: LFortran doesn't check interfaces
The following is just one of the examples that I've encountered:

`$ cat nocheck.f90 `
```
module abstypemod

   type, abstract :: AbsType
   contains
      procedure(method), deferred :: method
   end type AbsType

   abstract interface
      subroutine method(self,a)
         import; implicit none
         class(AbsType),       intent(inout) :: self
         real(8), allocatable, intent(inout) :: a(:,:)
      end subroutine method
   end interface
   
end module abstypemod

module mytypemod
   
   use abstypemod, only: AbsType

   contains
      procedure :: method
   end type MyType

contains
   
   subroutine method(self,i1,a)
      class(MyType),  intent(inout) :: self
      integer, allocatable, intent(inout) :: i1(:)
      real(8), allocatable, intent(inout) :: a(:,:)
   end subroutine method

end module mytypemod

```
Here's what one gets when trying to compile this with flang:

`$ flang-20 -c nocheck.f90 `
```
error: Semantic errors in nocheck.f90
./nocheck.f90:24:20: error: A type-bound procedure and its override must have compatible interfaces
        procedure :: method
                     ^^^^^^
./nocheck.f90:9:18: Declaration of 'method'
        subroutine method(self,a)
                   ^^^^^^
```
LFortran accepts this invalid code. It doesn't print an error.

usefahmed07:
Fixes #12311

LFortran used to silently accept a type-bound procedure override even if its arguments didn't match the parent/deferred one. Now it correctly reports an error: "A type-bound procedure and its override must have compatible interfaces."

Test: errors/interface_override_incompatible_01.f90

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4782,6 +4782,8 @@ RUN(NAME coarrays_32 EXTRA_ARGS --coarray=true)
 RUN(NAME coarrays_33 EXTRA_ARGS --coarray=true)
 RUN(NAME coarrays_34 EXTRA_ARGS --coarray=true NUM_IMAGES=4)
 
+RUN(NAME coarrays_37 EXTRA_ARGS --coarray=true)
+
 RUN(NAME gpu_metal_01 LABELS gfortran llvm metal)
 RUN(NAME gpu_metal_02 LABELS gfortran llvm metal)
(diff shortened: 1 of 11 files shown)


Jatinagarwal24:
Pls put the test in `continue_compilation_1` to throw error message.

PR #12352, opened 2026-07-26
Fix the calling convention of the binary128 runtime helpers

certik:
## Summary

Relational operators on `real(16)` return wrong answers at run time. On macOS ARM64 `2.0_16 > 1.0_16` is `.false.`; on Linux x86-64 the answer is whatever the argument registers happened to hold. On x86-64 the arithmetic and conversion helpers are wrong as well.

LLVM lowers an fp128 operation to a call to a compiler-rt helper and passes the operands with the target's native fp128 calling convention: a single SSE register on x86-64, a SIMD register on ARM64. The helpers in `lfortran_float128.c` are written against `lf_float128`, a 16-byte struct, which x86-64 System V classifies INTEGER + INTEGER and passes in a pair of general purpose registers, and which ARM64 passes in `x0`/`x1` rather than `q0`. Every helper therefore read its operands from the wrong registers and returned its result in the wrong one.

ARM64 already bridged the two conventions with naked-function shims, but only for arithmetic and conversions — the comparison helpers had none. On x86-64 nothing was bridged: a comment claimed the two conventions coincide, which is not true, and the weak-symbol fallback meant to let libgcc take over on ELF sits in a branch guarded by `LFORTRAN_HAVE_REAL128`, a macro that is never defined.

## Scope

- ARM64: a `_LF_SHIM2_TO_INT` shim for each of `__eqtf2`, `__netf2`, `__lttf2`, `__letf2`, `__gttf2`, `__getf2`, `__unordtf2`.
- x86-64: the public entry points now take a 16-byte vector of doubles, which System V classifies exactly like `__float128`, and convert to the struct the implementations expect. This covers the comparisons, `__addtf3` and friends, the conversions, and `lf_sqrtq` / `lf_powq`. The conversions are plain `memcpy`, so no fp128 arithmetic is generated in the wrappers and they cannot recurse into themselves.
- Compilers without vector extensions keep the previous struct signatures.
- The public entry points are no longer declared in `lfortran_float128_llvm.h`: their signature is per platform now, and nothing outside `lfortran_float128.c` calls them from C.

## Verification

New integration test `real128_compare_01` covers `<`, `<=`, `>`, `>=`, `==`, `/=` on runtime `real(16)` values, including equal operands, negative operands, mixed signs and signed zero. It fails on `main` with `ERROR STOP 1` and passes with this change; `gfortran` accepts and passes it too.

The ABI itself was checked with an LLVM IR harness exercising 18 comparison, arithmetic and conversion properties through `fcmp` / `fadd` / `fsub` / `fmul` / `fptrunc` on `fp128`, linked against the runtime and run natively on ARM64 and under Rosetta for x86-64:

| | ARM64 | x86-64 |
| --- | --- | --- |
| before | `0x03ff0` | `0x03faf` |
| after | `0x3ffff` | `0x3ffff` |

(mask of passing checks, `0x3ffff` = all 18). The x86-64 column shows that the comparisons *and* `__addtf3` / `__subtf3` / `__multf3` / `__trunctfdf2` were all affected there.

Suites, macOS ARM64:

- `integration_tests/run_tests.py`: `100% tests passed, 0 tests failed out of 3828`
- `./run_tests.py` (reference): `TESTS PASSED`
- `integration_tests/run_tests.py -b gfortran`: `real128_compare_01` passes; the 9 remaining failures are identical with and without this change (pre-existing on this gfortran).

## Rationale

The existing `real128_arith_*` tests only `print` their results and integration tests are judged by exit status alone, so a wrong answer from these helpers was invisible; the `real128_intrinsic_*` tests compare named constants, which are folded at compile time. `real128_compare_01` is the first test that computes a `real(16)` value at run time and checks it.


src/libasr/runtime/lfortran_float128.c
```diff
@@ -457,14 +457,17 @@ static int f128_cmp_internal(lf_float128 a, lf_float128 b) {
...
-int __lttf2   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
-int __letf2   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
-int __gttf2   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
-int __getf2   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
-int __unordtf2(lf_float128 a, lf_float128 b) { return (f128_is_nan(a)||f128_is_nan(b))?1:0; }
+int __eqtf2_lf_impl   (lf_float128 a, lf_float128 b) { return f128_cmp_internal(a,b) == 0 ? 0 : 1; }
+int __netf2_lf_impl   (lf_float128 a, lf_float128 b) { return f128_cmp_internal(a,b) != 0 ? 1 : 0; }
+int __lttf2_lf_impl   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
+int __letf2_lf_impl   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
+int __gttf2_lf_impl   (lf_float128 a, lf_float128 b) { int c=f128_cmp_internal(a,b); return c>=2?1:c; }
```
opixdown: "`__gttf2`/`__getf2` currently `return c >= 2 ? 1 : c,` which maps unordered to a positive value. LLVM lowers as `__gttf2(a, b) > 0` /` __getf2(a, b) >= 0`, so x > `NaN` (and >=) would evaluate to .true."

integration_tests/real128_compare_01.f90
```diff
@@ -0,0 +1,60 @@
...
+    if (.not. (b < a)) error stop 22
+    if (a == b) error stop 23
+
+    ! zero compares equal regardless of sign
+    a = 0.0_real128 * real(n + 1, real128)
+    b = -a
+    if (.not. (a == b)) error stop 24
+    if (a < b) error stop 25
+    if (a > b) error stop 26
+
```
opixdown, line 58: PR also changes the unordered comparison path. we should verify all comparison operators with runtime generated NaN values (using ieee_value) to ensure they behave correctly.

src/libasr/runtime/lfortran_float128.c
```diff
@@ -1651,27 +1654,73 @@ lf_float128 lf_float128_from_str(const char *s) {
...
+    return r;
+}
+
+static lf_float128 lf_f128_from_abi(lf_f128_abi v) {
+    lf_float128 r;
+    memcpy(r.bytes, &v, 16);
+    return r;
+}
+#else
+/* Compilers without vector extensions (MSVC): keep the historical
```
opixdown: The `#else` branch is not just "MSVC" it covers all non-x86-64, non-ARM64, non-WASM targets.we should update the comment to describe that more accurately

src/libasr/runtime/lfortran_float128.c
```diff
@@ -1651,27 +1654,73 @@ lf_float128 lf_float128_from_str(const char *s) {
...
+ * struct-based signatures. */
```
opixdown: Should we produce a clear error message or ice for unsupported targets here instead of silently using the struct based fallback?

opixdown:
LGTM. Also it would be good to add more float128 tests or even we can compile and run more real(16) code to see the correctness of our implementation. But it will required more work (e.g. compiling package), Thank you

src/libasr/runtime/lfortran_float128.c
certik: I think so.

certik:
@opixdown thanks for fixing it, I think this is good to merge.

Issue #10885, opened 2026-03-28
Compile M_Framework


HarshilShah1804:
Package Link: https://github.com/urbanjost/M_framework

Current Compilation Status:
```bash
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/M_framework$ fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
 + mkdir -p build/dependencies
M_framework__timing.f90                done.
M_framework__utility.f90               done.
M_framework__msg.F90                   failed.
[  3%] Compiling...
warning: Unsupported macro
  --> ././src/M_framework__msg.F90:25:1
   |
25 | #   warning  NOTE: UNKNOWN COMPILER
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Ignored

semantic error: Kind -1 is not supported for Real
   --> ././src/M_framework__msg.F90:234:20
    |
234 |       type is (real(kind=real128));     write(line(istart:),'(1pg0)') generic
    |                    ^^^^^^^^^^^^ 


Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
<ERROR> Compilation failed for object " src_M_framework__msg.F90.o "
<ERROR> stopping due to failed compilation
STOP 1
```

opixdown:
with #11755  it compiles to 31%

```
(base) opixdown@Unknown_2e:0f:6a:73:84:7a M_framework % fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
 + mkdir -p build/dependencies
M_framework__utility.f90               done.
M_framework__msg.F90                   done.
M_framework__timing.f90                done.
unit_test.f90                          done.
test_suite.f90                         done.
fid.f90                                done.
M_framework__verify.F90                done.
M_framework__journal.f90               done.
M_framework__attr.f90                  done.
demo_stderr.f90                        done.
demo_fmt.f90                           done.
demo_fstop.f90                         done.
demo_assert.f90                        done.
demo_set.f90                           done.
demo_wrt.f90                           done.
demo_str.f90                           done.
demo_pdec.f90                          done.
test_suite_M_framework__timing.f90     done.
M_framework__approx.F90                done.
demo_attr.f90                          done.
demo_attr_mode.f90                     done.
demo_M_framework__journal.f90          done.
[...]

PR #12367, opened 2026-07-27
fix: real128 runtime arithmetic, log10 routing, and string read

opixdown:
## Summary

Stacked on #12352 — this branch contains its commits and must be merged after it. Only the top three commits are new:

1. **fix: real128 soft-float arithmetic bugs** — four correctness bugs in `lfortran_float128.c`:
   - `__divtf3_lf_impl` lost fractional quotient bits whenever the dividend mantissa was >= the divisor mantissa (`3.0_16/2.0_16` returned `1.0`), which also broke `sqrt`, `log`, `exp` (all built on division).
   - The `f128_ln2` and `f128_pi`/`f128_pi_2`/`f128_pi_4` constants held misencoded bit patterns (ln2 decoded to 0.5866).
   - `lf_f128_log` reduced the exponent with conventions off by 112, so its atanh series diverged.
   - `__trunctfdf2_lf_impl` mishandled the rounding carry out of the 52-bit fraction, so values one ulp below a power of two converted to double as e.g. `1.5` instead of `1.0`.
2. **fix: route real128 log10 to an fp128 runtime entry point** — kind-16 `log10` previously called `_lfortran_dlog10` (double ABI) with an fp128 argument and returned garbage; adds `lf_log10q` with the same per-platform ABI bridging as `lf_sqrtq`.
3. **fix: implement `_lfortran_string_read_f128`** — list-directed internal reads into `real(16)` failed to link (undefined symbol); found while building M_framework.

## Scope

`src/libasr/runtime/lfortran_float128.c`, `src/libasr/runtime/lfortran_intrinsics.c`, `src/libasr/pass/intrinsic_functions.h`, plus three integration tests.

## Verification

- New integration tests `real128_arith_11`, `real128_intrinsic_math_01`, `read_99` (labels `gfortran llvm`): each fails on the base branch (the read test with a link error) and passes with its commit.
- `./run_tests.py --no-llvm`: TESTS PASSED (llvm-dump references differ locally only by opaque-pointer syntax from the local LLVM version).
- real128 integration tests: 15/15 pass, including `real128_compare_01` from #12352.
- End-to-end: with this branch, `fpm test` for M_framework (`--compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"`) compiles 100% and all test tallies pass (previously: link failure, then wrong `out_of_range`-style accuracy results from broken real128 log10).

## Rationale

The portable binary128 soft-float library is exercised on platforms without native `__float128` (notably macOS ARM64). These bugs made basic real128 arithmetic and math intrinsics silently wrong at runtime; the fixes follow the existing `lf_sqrtq`/`_lf_impl` bridging patterns introduced in #12352.


opixdown:
with this pr  `M_framework` #10885 compiles to 100%, we can split the commits in separate pr so we can review it easily

certik (approved):
I think the changes are fine.

certik:
Needs to be rebased.

PR #12380, opened 2026-07-29
Fix EXTERNAL procedures via ImplicitInterface deftype and FunctionPointerCast

certik:
## Summary

Follow-up to the EXTERNAL / implicit-interface work (supersedes the approach in #12275 for the design issues found in review).

An interface-less `external` declaration no longer pretends to be a zero-argument `Interface` function. ASR `deftype` gains `ImplicitInterface` so "arguments unknown" is stated explicitly. A module's bare declaration is not code-generated (so it cannot poison a later call site's link name). Divergent later references keep the first-inferred signature as canonical and call through a new `FunctionPointerCast` so the `SubroutineCall` always agrees with its callee.

Also fixes two frontend bookkeeping leaks: BLOCK-scoped `external` no longer leaks across program units, and external accessibility is not written back into the uncleared `assgnd_access` map.

## Scope

- ASR: `deftype += ImplicitInterface`; new expr `FunctionPointerCast(arg, to, type)`
- Semantics: emit `ImplicitInterface` for interface-less EXTERNAL; cast path for later mismatched references under `--implicit-interface`
- Backends: skip codegen of module bare declarations; LLVM lowers `FunctionPointerCast` as a bitcast
- Drop shape-guess heuristics (unused_functions pruning of placeholders, libasr Call_t_body BindC skip, use-associated zero-arg BindC guess)
- Tests: external_17-21, implicit_interface_44-47

## Verification

- `cd integration_tests && ./run_tests.py -j12` - 3844 passed
- `./run_tests.py` - TESTS PASSED
- implicit_interface_46 compiles with `--skip-pass unused_functions`
- Explicit zero-arg bind(c) interface still rejects extra arguments under `--implicit-interface`

## Rationale

Review of #12275: encoding "unknown signature" as "zero arguments" forced four competing shape heuristics and could not tell a module external apart from a genuine bind(c) zero-arg interface. FunctionPointerCast makes the separate-compilation / non-conforming second-signature case explicit without orphaning the first call's symbol pointer.

Note: implicit_interface_44 is not standard Fortran (kind disagreement on dummies). It is kept as the extension gfortran accepts with `-fallow-argument-mismatch`, implemented via cast rather than a lying single interface.


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8185,6 +8207,26 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    if (!F->use_empty()) {
+                        llvm::Value* cast = llvm::ConstantExpr::getBitCast(
+                            new_F, F->getType());
+                        F->replaceAllUsesWith(cast);
+                    }
+                    F->eraseFromParent();
+                    F = new_F;
+                    llvm_symtab_fn[old_h] = F;
+                    llvm_symtab_fn_names[fn_name] = h;
+                }
```
certik, line 8229: Don't like this.

PR #12384, opened 2026-07-29
Add Function.link_name for self-named external generics

certik:
## Summary

Adds an optional `Function.link_name` ASR field for the object-file / external linkage name when it differs from `Function.name` (symbol-table key).

This fixes self-named generic handling without backend string surgery on `~genericprocedure`:

1. **Mixed-case self-named module procedures** (`generic_name_09`): case-insensitive match when wiring generic procedure lists.
2. **External interface bodies that share the generic name** (`generic_name_10`): semantics stores the specific as `"<name>~genericprocedure"` and sets `link_name` to `"<name>"`; LLVM uses `link_name` for the external symbol.

Also adds ASR verification: interface + non-module + `~genericprocedure` suffix requires `link_name` equal to the name without the suffix.

## Scope

- ASR: `Function.link_name` (not on `FunctionType`; independent of `bindc_name`)
- Semantics + verification + LLVM naming
- Integration tests `generic_name_09`, `generic_name_10`
- Dedicated commit updating ASR reference pickles

## Design notes

See `doc/src/asr/function_link_name.md`.

| Field | Meaning |
| --- | --- |
| `Function.name` | ST / ASR identity (may be `foo~genericprocedure`) |
| `FunctionType.bindc_name` | BindC export only |
| `Function.link_name` | External link name when it differs from `name` (null = use `name`) |

## Verification

- `lfortran integration_tests/generic_name_09.f90` → PASS
- `lfortran integration_tests/generic_name_10.f90` → PASS
- Sample ASR refs: `interface_generic_procedure_same_name`, `functions_04`, `string1`

## Rationale

Follow-up to discussion on #12273: the backend must not invent or strip frontend naming conventions; the external name belongs in ASR. Placed on `Function` (not `FunctionType`) because it is not part of the procedure type.


src/libasr/ASR.asdl
```diff
@@ -11,7 +11,12 @@ unit
 symbol
     = Program(symbol_table symtab, identifier name, identifier* dependencies, stmt* body, location start_name, location end_name)
     | Module(symbol_table symtab, identifier name, identifier? parent_module, identifier* dependencies, bool loaded_from_mod, bool intrinsic, bool has_submodules, location start_name, location end_name)
-    | Function(symbol_table symtab, identifier name, ttype function_signature, identifier* dependencies, expr* args, stmt* body, expr? return_var, access access, bool deterministic, bool side_effect_free, string? module_file, location start_name, location end_name)
+    -- link_name: optional object-file / external linkage name when it differs
+    -- from Function.name (e.g. self-named generic specifics stored as
+    -- "<name>~genericprocedure" in the symbol table but link as "<name>").
+    -- Independent of bind(C): FunctionType.bindc_name remains the BindC export
+    -- name only. Null means backends should use Function.name (default).
```
certik: This must be documented in doc/, not here.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -118,11 +118,13 @@ static std::string compute_llvm_function_name(
...
     const CompilerOptions& compiler_options,
     const std::string& mangle_prefix,
-    const ASR::Function_t* parent_function
+    const ASR::Function_t* parent_function,
+    const char* link_name = nullptr
 ) {
     (void)compiler_options;
     bool is_external_interface = is_external_interface_function(ftype);
     bool is_bindc = (ftype->m_abi == ASR::abiType::BindC);
+    bool has_link_name = (link_name && link_name[0] != '\0');
```
certik: Let's enforce the `link_name[0] != '\0'` in verify(), so here we don't need to check. You can put an assert here, but do not check.

doc/src/asr/function_link_name.md
```diff
@@ -0,0 +1,51 @@
+# Function.link_name
+
+## Problem
+
+A specific procedure that shares its generic interface name cannot occupy the
```
certik: Move all this into the documentation for the Function node, which should be in doc/src/asr/asr_nodes/symbol_nodes/. It's currently not there, so add a new Function.md in there and document it, including this new `link_name` member.

certik (commented):
## Review

Built the branch in-source (Debug, `WITH_LFORTRAN_ASSERT=yes`, so the new `require`s were live) and exercised it. Findings ordered by severity.

Note for anyone building this locally: `src/libasr/asr.h` needs regeneration for the ASDL change (`python src/libasr/asdl_cpp.py src/libasr/ASR.asdl src/libasr/asr.h`); `cmake --build` alone does not do it.

---

### [rework] `link_name` is set on module-procedure interface bodies, where it is *not* the linkage name

`src/lfortran/semantics/ast_symboltable_visitor.cpp:1646` and `:2349`

Both set sites gate only on `deftype == Interface`, never on `is_module`. A `module function` / `module subroutine` inside a self-named interface therefore gets a `link_name` equal to the bare source name, while its real external symbol is the mangled module-procedure name:

```fortran
module mi
  implicit none
  interface bar
    module function bar(x) result(y)
      integer, intent(in) :: x
      integer :: y
    end function
  end interface
end module
```

produces, on this branch:

```
bar~genericprocedure: (Function ...
    (FunctionType [(Integer 4)] (Integer 4) Source Interface () .false. .false. .true. .false. .false. [] .false.)
    ... Public .false. .false. () "bar")
```

`m_module` is `.true.` and `link_name` is `"bar"` — wrong. The new verifier rule explicitly excludes `!ftype->m_module`, so it does not catch this.

LLVM escapes the bug only because `is_external_interface_function()` tests `!m_module` *before* consulting `link_name`. That means the backend is still inferring the answer rather than trusting the ASR field — which is precisely what this PR sets out to eliminate. Any consumer that follows the rule documented in the new `Function.md` ("else if link_name set → use link_name") emits the wrong symbol.

Recommended shape: add `&& !is_module` at both set sites, and extend `Function` verification to require `link_name == nullptr` when `m_module` is true. That makes the field mean what the doc says, and lets the backend drop the `is_external_interface_function` pre-check.

### [rework] The `~genericprocedure` convention moves *into* libasr

`src/libasr/asr_verify.cpp:550-577`

Before this PR, `~genericprocedure` appeared only under `src/lfortran/semantics/`. The verifier now string-matches that suffix and hard-fails on it. From `AGENTS.md`:

> libasr is frontend-independent. Never reference `_lfortran` or any frontend-specific names in libasr code. Use enums or structured types, not string comparisons.

The PR's own rationale — the backend must not invent or strip frontend naming conventions — applies here too: the convention was removed from `asr_to_llvm.cpp` and re-added to shared infrastructure, now in a form that aborts compilation.

Recommended shape: keep only the frontend-neutral invariant in `asr_verify.cpp` (`link_name` is null or non-empty), and enforce the suffix ↔ `link_name` correspondence in the Fortran frontend where the convention is defined — e.g. an assertion next to the two set sites in `ast_symboltable_visitor.cpp`.

### [follow-up] A third suffix site never sets `link_name`

`src/lfortran/semantics/ast_symboltable_visitor.cpp:1611`

The GenericProcedure-shadowing branch in `visit_Subroutine` appends `~genericprocedure` without setting `link_name`. If it ever produced an Interface non-module Function, the new `require` would abort with an internal ASR error instead of a diagnostic.

It is unreachable today only because line 1632 (`parent_scope->erase_symbol(sym_name)`) asserts first — a pre-existing bug, reproducible with a use-associated generic re-declared as an interface body:

```
AssertFailed: scope.find(name) != scope.end()   (asr_scopes.h:79, from ast_symboltable_visitor.cpp:1632)
```

Worth a comment at 1611 so the verify rule is not a landmine when that crash gets fixed.

### [follow-up] Nothing tests the new ASR state or the new verify rules

Roughly 300 reference files churn by one `()`, but **no reference anywhere contains a non-null `link_name`** — the new field is never pinned by a reference test. `generic_name_09` / `generic_name_10` are integration-only and not registered in `tests.toml`. The PR description cites `interface_generic_procedure_same_name`, `functions_04`, and `string1` as sample refs, but those only gained a null `()`.

Suggest adding `asr = true` for `generic_name_10` so `link_name = "get_text"` is pinned, plus a negative case for each `require` branch.

### [follow-up] C backend still emits the raw disambiguated key

```
$ lfortran --show-c integration_tests/generic_name_10.f90
int32_t get_text~genericprocedure();
...
r = get_text~genericprocedure();
```

Pre-existing, but this PR adds exactly the state that fixes it and wires only LLVM. Since verification now guarantees `link_name` is present for this case, `asr_to_c_cpp.h` is a one-line change — and doing it would let `generic_name_10` carry a `c` label.

### [follow-up] Two unrelated fixes in one PR

`generic_name_09` is fixed by `to_lower(pname.first) == proc.first` at line 3836 — a plain case-sensitivity bug in `add_generic_procedures` affecting a self-named **module procedure**. It involves no `link_name` and needs none. `AGENTS.md`: one bug = one MRE = one PR.

### [follow-up] `SymbolDuplicator` change is wider than `link_name`

`src/libasr/asr_utils.h:6369-6372` now also copies `m_module_file`, `m_start_name`, and `m_end_name`, which `main` silently dropped. Likely a genuine improvement, but it is an unrelated behavior change riding along untested.

Minor doc point: `Function.md`'s naming rule omits the `bind(c, name='')` branch, where `link_name` is ignored in favor of the mangled name.

---

## Architecture assessment

- **Earliest wrong phase:** AST→ASR. The symbol-table key was disambiguated without recording the linkage name, forcing later phases to guess.
- **Chosen layer:** ASR field + semantics + verification. Correct — this is state, not a transformation, so a field beats a pass, and `Function` beats `FunctionType` since linkage is not part of the procedure type.
- **Invariant added:** `link_name` is null or non-empty; suffixed non-module interface bodies must carry it. **Still missing:** the converse — module procedures must *not* carry it. That gap is what leaves the LLVM backend inferring from `m_module` instead of reading ASR.
- **Backend complexity:** net reduced. `compute_llvm_function_name` gains one parameter and loses no logic. But the `is_external_interface_function` pre-check still encodes the decision the field was meant to carry.

An alternative worth considering: rename the **GenericProcedure** rather than the specific. The generic is a compiler-level construct with no linkage; the specific has one. `add_generic_procedures` already does exactly this when the generic name collides with a `Struct` or `Function` (`sym_name_str = "~" + proc.first`). Extending that precedent would keep `Function.name` equal to the true external name and remove the need for `link_name` entirely.

---

## Verification performed

- `./run_tests.py` — **TESTS PASSED** (excluding `runtime_stacktrace_01`, which fails in my checkout because it is configured without `WITH_RUNTIME_STACKTRACE`).
- Integration suite via CMake + CTest — `generic_name_09` and `generic_name_10` both pass. Four failures, all environmental and unrelated: `derived_types_123` (needs `-g` / runtime stacktrace), `intrinsics_367/380/381` (need `LFORTRAN_TEST_ENV_VAR`, which `run_tests.py` sets and bare `ctest` does not).
- Additional cases checked, all passing: mixed-case self-named external interface; self-named external interface **subroutine**; `bind(c, name=...)` and bare `bind(c)` on a self-named generic (`declare i32 @abs2(i32)` — correct); separate-compilation `.mod` round-trip of `link_name`; submodule self-named generic.

PR #12386, opened 2026-07-29
test: use free form for implicit interface regression

### LINKED ISSUE #12385: Use free form in a test
This has nothing to do with fixed form, just use free form.

_Originally posted by @certik in https://github.com/lfortran/lfortran/pull/12359#discussion_r3677461186_

krystophny:
## Summary

- convert `implicit_interface_44.f90` from fixed-form layout to free-form source
- remove the LFortran and GFortran fixed-form runner flags while retaining `--implicit-interface`

## Why

The regression exercises user-function shadowing of an intrinsic through an implicit interface. Fixed-form parsing is unrelated to that behavior and made the test cover an extra concern unnecessarily.

## Impact

The test continues to verify that the two-argument user function `diff` is selected over the one-argument intrinsic and returns `2.0d0`, now through the normal free-form path. Compiler behavior outside the test suite is unchanged.

## Root cause

The test source was written in fixed-form column layout and its runner registration explicitly forced fixed form even though the regression did not depend on source form.

## Checks

- `PATH=/home/ert/code/lfortran-12385/build/src/bin:$PATH ./run_tests.py -b llvm -t '^implicit_interface_44$' --ninja` — 1/1 passed
- `gfortran integration_tests/implicit_interface_44.f90 -o /tmp/lfortran-12385-gfortran-test && /tmp/lfortran-12385-gfortran-test` — passed
- `git diff --check` — passed

Fixes #12385.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2190,7 +2190,7 @@ RUN(NAME implicit_interface_40 LABELS gfortran llvm EXTRA_ARGS --implicit-typing
 RUN(NAME implicit_interface_41 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_42 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME implicit_interface_43 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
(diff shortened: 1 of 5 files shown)


certik (approved):
This is good. Can you please also add a reference tests for this exact test, without `--implicit-interface`? It should still show an error.

krystophny:
Added the requested no-flag reference test, then stabilized its diagnostic across LLVM 11/21 by declaring `diff` external. Targeted tests and the full local pipeline pass.

PR #12308, opened 2026-07-22
fix: EQUIVALENCE for arrays, multidim offsets, backing storage and DATA reshape

certik:
## Summary
Several fixes to make Fortran `EQUIVALENCE` work correctly, grouped together as one thematic change (extracted from the `xfoil` branch).

## Scope
- Handle equivalence between a COMMON block scalar and a single-element array member (`equivalence_40`).
- Span all members of a COMMON-backed array in array equivalence, not just unit size (`equivalence_40`).
- Preserve multidimensional offsets when computing equivalence positions, replacing the first-index-only logic with proper linear-offset math (`equivalence_41`).
- Use the larger of the two equivalenced arrays as the backing storage so the alias cannot overrun it (`equivalence_42`).
- Reshape flat `DATA` values to the equivalenced target's rank before assigning them (`equivalence_43`).

## Verification
Built locally with `-DCMAKE_BUILD_TYPE=Debug -DWITH_LLVM=ON`. The four integration tests compile and run successfully with `lfortran`:
- `equivalence_40.f90`, `equivalence_41.f90`, `equivalence_42.f90`, `equivalence_43.f90`

All are registered with `LABELS gfortran llvm` and also pass with gfortran.

## Rationale
These commits all touch the equivalence machinery in `ast_common_visitor.h` (the multidim-offset, backing-array and DATA-reshape fixes build directly on the common/array-equivalence code introduced here), so they are kept together as one coherent, reviewable change.

PR #12387, opened 2026-07-30
fix: preserve integer kind and reject COMMON overrun in EQUIVALENCE

### LINKED ISSUE #12383: Followup for #12308
The first two issues:

* `[blocker] clone_type drops non-default integer kinds`
* `[rework] COMMON array span can overrun storage`

I would fix, but it can be done in a subsequent PR.

_Originally posted by @certik in https://github.com/lfortran/lfortran/issues/12308#issuecomment-5121663677_

certik:
Follow-ups to #12308 (fixes #12383):

- clone_type hardcoded default_integer_kind for every Integer type, so integer(8) array-array EQUIVALENCE aliases were lowered as Pointer(Array(Integer 4)), truncating 64-bit values. Preserve the source kind via down_cast<Integer_t>->m_kind, matching the Real/Complex/Logical branches.

- The common-scalar to array path aliased a local array onto a COMMON member with the array's full shape without checking that it fits in the remaining COMMON storage, silently overrunning the block. Compute the member's byte offset and the block's total size and reject with the existing "not implemented" diagnostic when the alias does not fit (extending the COMMON block is left for a future change).

Tests:
- integration_tests/equivalence_44.f90: integer(8) array-array EQUIVALENCE, covering the larger-backing-swap and equal-size cases.
- tests/errors/continue_compilation_1.f90: new overrun rejection case.


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -6834,51 +6878,6 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
-                                    if (type != nullptr) {
-                                        if (!(in_module && !in_Subroutine)) {
-                                            ASR::ttype_t* ptr = ASRUtils::TYPE(ASR::make_Pointer_t(al, asr_eq1->base.loc, type));
-                                            var__->m_type = ptr;
-                                        }
-
-                                        bool is_module = in_module && !in_Subroutine;
-                                        if (is_module) {
-                                            var__->m_symbolic_value = ASRUtils::EXPR(get_pointer);
-                                        } else {
```
Jatinagarwal24, line 6881: This was deadcode.So if removed this.

certik:
I think this looks good, thanks for fixing it! I merged it.

PR #12398, opened 2026-07-30
Report semantic error for type mismatch on class(T) dummy argument

### LINKED ISSUE #12394: Bug : passing to class parameter should report error robuslty
```.f90
module m1
    implicit none
    type :: runner
    contains
    end type

contains

    subroutine set_caller(this)
        class(runner) :: this
    end subroutine
end module

program p
    use m1
    call set_caller(1)
end program
```
This fails at codegen.
Why ? because we should've raised a semantic error.
Notice : I think passing to class parameters has some flexibility, I don't know how exactly, but I think (not sure) you can pass arguments to class types and some implicit conversion happens. bottom line, the error could be more flexible not just plain argument/parameter type mismatch error

usefahmed07:
Fixes #12394.

The SubroutineCall argument-checking loop in `ast_body_visitor.cpp` had
no general type-compatibility check for `class(T)` dummy arguments, so
passing a non-derived-type actual (e.g. an integer) reached codegen
instead of being caught at the semantic stage.

Added a check using `ASRUtils::is_class_type` and
`ASRUtils::check_class_assignment_compatibility` that reports a clear
"Type mismatch" semantic error instead.

Added a regression test case to `tests/errors/continue_compilation_1.f90`
(confirmed it fails on main without the patch, with a codegen crash
instead of a semantic error).


assem2002 (commented):
The change seems okay, but I had something else in mind.
Look 
https://github.com/lfortran/lfortran/blob/765faa6ba00f60994aa757ea9c4a6817742837f1/src/lfortran/semantics/ast_body_visitor.cpp#L8489-L8492

This spot is where we start checking arg/param types. What I had in mind is to handle the `skip_check` variable. 
***
### what to do?
- Introduce function named `can_pass_class_argument()` that checks whether argument is passable to parameter. I think a call to `check_class_assignment_compatibility` might do it.
- Pass return of that function to `skip_check`
***
also reference tests need a look

usefahmed07:
Screenshot From 2026-08-02 21-26-07
@assem2002 Done! Added `can_pass_class_argument()` in `asr_utils.h` like you suggested.

It checks compatibility between the dummy and the passed arg using `check_class_assignment_compatibility()`. Only exception is `class(*)` — kept that accepting anything since a couple existing tests rely on it.

`skip_check` now comes from that function's result for class(T) args, and if it's not compatible we throw a proper "Type mismatch" error right there instead of letting it slide to codegen.

Tested it Works Now.

usefahmed07:
@assem2002 Quick update  testing this hit a false positive on `derived_types_04.f90` (`rhs%date%to_string(lhs)`, uses `pass(rhs)`)

Turns out there's a pre-existing bug: when `pass(name)` isn't the first dummy arg, the arg-matching offset assumes it is, so args get mismatched. It was silently hidden before since any class-type dummy always skipped the check
Instead of fixing that (bigger, separate issue), I just scoped `can_pass_class_argument()` to skip self-passing bound-procedure calls and only validate direct calls like #12394's case. CI is green now on `166d4e2a0`.

Want me to open a separate issue for the `pass(name)` bug?

assem2002 (approved):
I think that's the required change.Thanks!
 I added a commit to handle the test and remove large comment sections.

PR #12403, opened 2026-07-31
fix: iany/iall intrinsics to correctly handle mask without explicit dim

### LINKED ISSUE #12357: intrinsic IANY appears to ignore MASK parameter or fails if present
IANY either segfaults or produces the same answer as if no mask were
present using version 0.64.0.

It works as expected with
  - flang version 23.0.0 
  - Intel(R) 64, Version 2025.3.2 Build 20260112
  - GCC version 16.1.0
 
And fails to compile with nvfortran and requires a special switch (not tried) for AOCC flang.

In addition, if the indicated WRITE statement is uncommented the Lfortran compile segfaults.

```fortran
     1	program demo_iany
     2	use, intrinsic :: iso_fortran_env, only : compiler_version
     3	   use, intrinsic :: iso_fortran_env, only : integer_kinds, &
     4	   & int8, int16, int32, int64
     5	   implicit none
     6	   logical,parameter :: T=.true., F=.false.
     7	   integer(kind=int8) :: a(3)
     8	   integer(kind=int8) :: answer
     9	
    10	   print '(2a)', 'This file was compiled by ', compiler_version()
    11	
    12	   ! set some values to exercise with
    13	   a(1) = int(b'00100100',int8)
    14	   a(2) = int(b'01101010',int8)
    15	   a(3) = int(b'10101010',int8)
    16	   answer=int(b'11101110',int8)
    17	   ! if any bit on in any element of A it should be on in answer
    18	
    19	   ! basic call
    20	   print '("A=")'
    21	   print '("  ",1x,b8.8)', a
    22	   print '("IANY(A)=",1x,b8.8,/)', iany(a)
    23	   print '("is it the expected value? ",1x,l1,/)', iany(a)==answer
    24	
    25	   ! select values with a mask
    26	   write(*,*)'IANY(A) with a mask'
    27	   write(*,*)'these values should be equivalent'
    28	
    29	   ! segfaults using lfortran 0.64.0
    30	   !print '(1x,b8.8)', iany(a,mask=[T,F,T])  
    31	   ! value is incorrect, not changed
    32	   print '(1x,b8.8)', iany(a,[T,F,T])  
    33	   ! value is incorrect, not changed
    34	   print '(1x,b8.8)', iany(a,dim=1,mask=[T,F,T])
    35	
    36	   print *
    37	   print '("the answer should match",1x,b8.8)', iany([a(1),a(3)])
    38	   write(*,*)'does it?'
    39	   write(*,*)iany(a,[T,F,T]) == iany([a(1),a(3)])
    40	   write(*,*)iany(a,[T,F,T]) == int(b'10101110',int8)
    41	
    42	end program demo_iany
```
```text
aoccflang520         :     1	STDERR:
aoccflang520         :     2	F90-S-0155-iany iall and iany is supported only in f2008, use -std=f2008 to enable
aoccflang520         :     3	 (/app/example.f90: 22)
aoccflang520         :    18	  0 inform,   0 warnings,   8 severes, 0 fatal for demo_iany

nvfortran_x86_26_3   :     2	NVFORTRAN-S-0079-Keyword form of argument illegal in this context for dim (/app/example.f90: 34)
nvfortran_x86_26_3   :     3	NVFORTRAN-S-0079-Keyword form of argument illegal in this context for mask (/app/example.f90: 34)
nvfortran_x86_26_3   :     4	NVFORTRAN-S-0038-Symbol, iany, has not been explicitly declared (/app/example.f90: 1)
nvfortran_x86_26_3   :     5	  0 inform,   0 warnings,   3 severes, 0 fatal for demo_iany

flangtrunk           :     2	This file was compiled by flang version 23.0.0 (https://github.com/llvm/llvm-project.git 8d61da2569a975346150371cb0d317fb9339be0c)
flangtrunk           :     3	A=
flangtrunk           :     4	   00100100
flangtrunk           :     5	   01101010
flangtrunk           :     6	   10101010
flangtrunk           :     7	IANY(A)= 11101110
flangtrunk           :     8	
flangtrunk           :     9	is it the expected value?  T
flangtrunk           :    10	
flangtrunk           :    11	 IANY(A) with a mask
flangtrunk           :    12	 these values should be equivalent
flangtrunk           :    13	 10101110
flangtrunk           :    14	 10101110
flangtrunk           :    15	
flangtrunk           :    16	the answer should match 10101110
flangtrunk           :    17	 does it?
flangtrunk           :    18	 T
flangtrunk           :    19	 T

ifxlatest            :     2	This file was compiled by Intel(R) Fortran Compiler for applications running on Intel(R) 64, Version 2025.3.2 Build 20260112
ifxlatest            :     3	A=
ifxlatest            :     4	   00100100
ifxlatest            :     5	   01101010
ifxlatest            :     6	   10101010
ifxlatest            :     7	IANY(A)= 11101110
ifxlatest            :     8	
ifxlatest            :     9	is it the expected value?  T
ifxlatest            :    10	
ifxlatest            :    11	 IANY(A) with a mask
ifxlatest            :    12	 these values should be equivalent
ifxlatest            :    13	 10101110
ifxlatest            :    14	 10101110
ifxlatest            :    15	 
ifxlatest            :    16	the answer should match 10101110
ifxlatest            :    17	 does it?
ifxlatest            :    18	 T
ifxlatest            :    19	 T

lfortranlatest       :     2	This file was compiled by LFortran version 0.64.0
lfortranlatest       :     3	A=
lfortranlatest       :     4	   00100100
lfortranlatest       :     5	   01101010
lfortranlatest       :     6	   10101010
lfortranlatest       :     7	IANY(A)= 11101110
lfortranlatest       :     8	
lfortranlatest       :     9	is it the expected value?  T
lfortranlatest       :    10	
lfortranlatest       :    11	IANY(A) with a mask
lfortranlatest       :    12	these values should be equivalent
lfortranlatest       :    13	 11101110
lfortranlatest       :    14	 11101110
lfortranlatest       :    15	
lfortranlatest       :    16	the answer should match 10101110
lfortranlatest       :    17	does it?
lfortranlatest       :    18	F
lfortranlatest       :    19	F

gfortran161          :     2	This file was compiled by GCC version 16.1.0
gfortran161          :     3	A=
gfortran161          :     4	   00100100
gfortran161          :     5	   01101010
gfortran161          :     6	   10101010
gfortran161          :     7	IANY(A)= 11101110
gfortran161          :     8	
gfortran161          :     9	is it the expected value?  T
gfortran161          :    10	
gfortran161          :    11	 IANY(A) with a mask
gfortran161          :    12	 these values should be equivalent
gfortran161          :    13	 10101110
gfortran161          :    14	 10101110
gfortran161          :    15	
gfortran161          :    16	the answer should match 10101110
gfortran161          :    17	 does it?
gfortran161          :    18	 T
gfortran161          :    19	 T
```

HarshilShah1804:
Fixes #12357


src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3592,11 +3688,19 @@ namespace IanyIall {
...
+            if (overload_id == 0) {
+                generate_body_for_scalar_output(al, loc, args[0], return_var, fn_symtab, body,
+                initial_value, elemental_operation);
+            } else {
+                generate_body_for_scalar_output_with_mask(al, loc, args[0], args[1], return_var, fn_symtab, body,
+                initial_value, elemental_operation);
+            }
         } else if( overload_id == 1 ) {
-            generate_body_for_array_output(al, loc, args[0], args[1], args[2], fn_symtab, body,
+            generate_body_for_array_output(al, loc, args[0], args[1], nullptr, fn_symtab, body,
```
Jatinagarwal24: This replaces the result out-parameter (`args[2])` with nullptr; `generate_body_for_array_output` immediately does builder.Assignment(`result`, `init_int_val`) on it, so every `iany(a, dim) `/ `iall(a, dim) `call now crashes the compiler , for rank-1 arrays return_var must be used (previously routed through the scalar generator), for rank ≥ 2 it must be `args[2]`. No test in the repo calls `iany/iall` with dim, which is why CI is green ,please also add one."

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3592,11 +3688,19 @@ namespace IanyIall {
...
+            initial_value, elemental_operation);
+        } else if( overload_id == 3 ) {
+            generate_body_for_array_output_with_mask(al, loc, args[0], args[1], args[2], return_var, fn_symtab, body,
```
Jatinagarwal24: `return_var` is only non-null when the result is scalar (rank-1 input). For `iany(a2d, dim=1, mask=m)` the result parameter is `args[3]` (filled at line 3679) and `return_var` is `nullptr` , same crash as above. Your new test only covers `1-D ` ,please add a `2-D dim+mask` case.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3469,6 +3512,55 @@ namespace IanyIall {
...
+            [=, &al, &fn_body, &builder] () {
+                ASR::stmt_t* return_var_init = builder.Assignment(result, init_int_val);
+                fn_body.push_back(al, return_var_init);
+            },
+            [=, &al, &idx_vars, &target_idx_vars, &doloop_body, &builder] () {
+                ASR::expr_t* array_ref = PassUtils::create_array_ref(array, idx_vars, al);
+                ASR::expr_t* target_var = PassUtils::create_array_ref(result, target_idx_vars, al);
+                ASR::expr_t* logical_op = (builder.*elemental_operation)(target_var, array_ref);
+                ASR::stmt_t* loop_invariant = builder.Assignment(target_var, logical_op);
+                ASR::expr_t* mask_ref = ASRUtils::is_value_constant(mask) ? mask : PassUtils::create_array_ref(mask, idx_vars, al);
```
Jatinagarwal24, line 3588: mask here is the generated function's formal parameter (a Var), so is_value_constant is always false , the first branch is dead code. The real distinction needed is scalar vs. array: verify_array_mask (line 3321) deliberately allows scalar masks (iany(a, mask=.true.), standard-conforming), and those currently fall into create_array_ref on a scalar.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3433,8 +3477,7 @@ namespace IanyIall {
                 return_type = ASRUtils::make_Array_t_util(al, loc,
                     return_type, dims.p, dims.size());
             }
-        }
-        if( mask ) {
+        } else if( mask ) {
             overload_id = 2;
         }
         value = eval_IanyIall(al, loc, return_type, arg_values, init_int_val, logical_operation);
```
Jatinagarwal24: `eval_IanyIall ` reduces over all elements of `args[0]` and never looks at the mask, so an all-constant call like `iany([36,106,170]` , `[.true.,.false.,.true.])` folds to 238 instead of 174 ,  the same wrong-silent-result class as #12357. Either skip folding when mask is present, or implement a masked fold. Please add a parameter-initializer test for this.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -764,11 +764,16 @@ static inline ASR::asr_t* create_ArrIntrinsic(
...
+            diag.add(diag::Diagnostic("'mask' argument of '" + intrinsic_func_name + "' intrinsic must be logical",
                 diag::Level::Error, diag::Stage::Semantic, {diag::Label("", {loc})}));
             return nullptr;
         }
         mask = args[2];
+        if (!ASRUtils::is_value_constant(mask)) {
+            if (!is_same_shape(array, mask, intrinsic_func_name, diag, {args[0]->base.loc, args[2]->base.loc})) {
+                return nullptr;
+            }
+        }
```
Jatinagarwal24: A scalar mask is conformable with any array per F2018 §3.36, so `sum(a, mask=some_logical_scalar_var)` is legal but now errors with `'ranks must be the same' `( `is_same_shape `, line 680, rejects rank 0 ≠ rank n). The sibling branches at lines 738 and 755 have the same pre-existing problem  , ideally fix the rank-0 case in all three.

src/libasr/pass/intrinsic_array_function_registry.h
```diff
@@ -3560,6 +3652,10 @@ namespace IanyIall {
                     fill_func_arg("result", return_type);
                 }
             } else if ( overload_id == 2 ) {
+                ASR::ttype_t* mask_type = ASRUtils::expr_type(new_args[1].m_value);
+                fill_func_arg("mask", mask_type);
```
Jatinagarwal24, line 3699: The array parameter uses `duplicate_type_with_empty_dims (line 3631)` but the mask keeps the actual's fixed shape, and the function-cache lookup (lines 3614–3616) only compares the array argument's type ,  a second call with a differently-shaped mask reuses a function declaring the first call's mask shape.

integration_tests/intrinsics_477.f90
Jatinagarwal24, line 1: Pls add the test for the following cases.
(a) `iall` with mask (title says `iany/iall `, but iall is never called , its mask path has zero coverage), (b)` iany(a, dim=1)` with no mask , (c) a 2-D array with dim+mask checked against known values , (d) a scalar mask `iany(a, mask=.true.)` (e) an all-constant `parameter :: p = iany([...], [...]`).

src/libasr/pass/intrinsic_array_function_registry.h
Jatinagarwal24, line 3444: No conformance `check: iany(a, [T,F])` with `size(a)==3 ` compiles and reads the mask out of bounds at runtime. Reuse` is_same_shape (line 680) `here, as this PR already does for the sum family.

Jatinagarwal24 (changes requested):
Pls address the comments.
Thanks.

certik:
A lot of changes that need to be carefully reviewed. Also this needs to be rebased.

PR #12404, opened 2026-07-31
fix: tokenization of fixed-form loops containing semicolons

### LINKED ISSUE #12393: Fixed form: `END DO` after a `;` on the same line is rejected
We've come across a few parsing errors in the effort of using `lfortran` on a large CFD code written in fixed form f90. First error is summarised below with a series of unittests with variations over the same issue.

## Summary

In **fixed-form** source, an `END DO` / `ENDDO` that follows a `;` on the same
line fails to parse. `;` is a statement separator (F2018 6.3.2.3), so these are
ordinary multi-statement lines. gfortran and ifort accept all of them.

Free form is unaffected, and `END IF` after a `;` is fine — the failure is
specific to the loop terminator in fixed form.

## Version

```
LFortran 0.64.0-142-g185bb63a1  (macOS ARM, LLVM 22.1.8)
Reference: GNU Fortran (Homebrew GCC 15.2.0_1) 15.2.0
```

## Unit test

```bash
./run.sh        # table
./run.sh -v     # with full diagnostics
```

Minimal case (`A1_enddo_semicolon_enddo.f`):

```fortran
      subroutine a1(n, x)
      integer n, i, j
      real x(n,n)
      do i = 1, n
      do j = 1, n
      x(i,j) = 0.0
      end do; end do
      end
```

```
$ lfortran -c --fixed-form --implicit-typing A1_enddo_semicolon_enddo.f
tokenizer error: Expected an executable statement inside a do loop
  --> A1_enddo_semicolon_enddo.f:12:7
   |
12 |       end
   |       ^

$ gfortran -fsyntax-only -ffixed-form A1_enddo_semicolon_enddo.f
```

## Expected vs actual

| # | construct | gfortran | lfortran |
|---|---|---|---|
| A1 | `end do; end do` | OK | **tokenizer error** |
| A2 | `enddo; enddo` | OK | **tokenizer error** |
| A3 | `end do; x(1) = 1.0` | OK | **syntax error** |
| A4 | `end do;`  *(stray trailing separator)* | OK | **syntax error** |
| A5 | `;end do`  *(stray leading separator)* | OK | **tokenizer error** |
| A6 | `do i = 1, n; x(i) = 0.0; end do` | OK | **tokenizer error** |
| A7 | `do l = f, 250; if (t(l+1:l+1) .eq. ' ') exit; end do` | OK | **tokenizer error** |
| A8 | `end if; x(1) = 2.0` and `x(1) = 3.0; x(1) = 4.0` | OK | OK *(control)* |
| A9 | A1 written in **free form** | OK | OK *(control)* |

A8 and A9 are controls that pass today. They localise the defect: it is not
`;` handling in general, and not `END IF`, but specifically `END DO`/`ENDDO`
in fixed form.

[A-fixed-form-enddo-after-semicolon.zip](https://github.com/user-attachments/files/30547148/A-fixed-form-enddo-after-semicolon.zip)

HarshilShah1804:
Fixes #12393

### DIFF
--- a/src/lfortran/parser/fixedform_tokenizer.cpp
+++ b/src/lfortran/parser/fixedform_tokenizer.cpp
@@ -455,6 +455,42 @@ struct FixedFormRecursiveDescent {
         if (*cur == '\n') cur++;
     }
 
+    void next_statement(unsigned char *&cur) {
+        bool in_string = false;
+        char quote = 0;
+        bool in_comment = false;
+        while (*cur != '\n' && *cur != '\0') {
+            if (*cur == '\r') {
+                cur++;
+                continue;
+            }
+            if (in_comment) {
+                cur++;
+                continue;
+            }
+            if (in_string) {
+                if (*cur == quote) {
+                    if (*(cur+1) == quote) {
+                        cur++; // skip escaped quote
+                    } else {
+                        in_string = false;
+                    }
+                }
+            } else {
+                if (*cur == '\'' || *cur == '"') {
+                    in_string = true;
+                    quote = *cur;
+                } else if (*cur == '!') {
+                    in_comment = true;
+                } else if (*cur == ';') {
+                    break;
+                }
+            }
+            cur++;
+        }
+        if (*cur == '\n' || (!in_string && !in_comment && *cur == ';')) cur++;
+    }
+
     // Push the token_type, YYSTYPE and Location of the token_str at `cur`.
     // (Does not modify `cur`.)
     void push_token_no_advance_token(unsigned char *cur, const std::string &token_str,
@@ -728,9 +764,9 @@ struct FixedFormRecursiveDescent {
      */
     void tokenize_line(unsigned char *&cur) {
         t.cur = cur;
-        next_line(cur);
+        next_statement(cur);
         tokenize_until(cur);
-        LCOMPILERS_ASSERT(*(t.cur-1) == '\n')
+        LCOMPILERS_ASSERT(*(t.cur-1) == '\n' || *(t.cur-1) == ';')
     }
 
     /*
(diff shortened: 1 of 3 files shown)


src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -728,9 +764,9 @@ struct FixedFormRecursiveDescent {
      */
     void tokenize_line(unsigned char *&cur) {
```
certik, line 765: This should not be called `tokenize_statement`

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -455,6 +455,42 @@ struct FixedFormRecursiveDescent {
...
+                } else if (*cur == '!') {
+                    in_comment = true;
+                } else if (*cur == ';') {
+                    break;
+                }
+            }
+            cur++;
+        }
+        if (*cur == '\n' || (!in_string && !in_comment && *cur == ';')) cur++;
+    }
```
certik, line 492: This is really hard to read to ensure it is correct. I would structure it into multiple functions like `parse_string`, and `parse_comment`, etc. That way we can ensure it is correct.

Also, I think we already have functions for many of these.

src/lfortran/parser/fixedform_tokenizer.cpp
certik: Consequently, we need to check every usage of `tokenize_line` that it will work when it is now `tokenize_statement`.

certik:
@HarshilShah1804 thanks for the fix. I think it's the right fix, but it needs to be cleaned up.

src/lfortran/parser/fixedform_tokenizer.cpp
certik: I don't like adding additional ad-hoc parsers like this. I want us to reuse as much as we can, and move this fixed-form tokenizer towards a single recursive descent parser, with no ad-hoc logic like this.

HarshilShah1804:
Thanks for the review @certik. Will address your comments and update the PR in some time.

PR #12409, opened 2026-08-01
fix: transfer full kind-width bytes for A editing of integers

### LINKED ISSUE #12400: "A" edit descriptor should do a binary transfer of an argument to output files in version 0.64.0
Having trouble finding the statement I want in the standard, but partly
for historical reasons before there even was a CHARACTER type the "A"
edit descriptor is rather special in that it can be used to print the
bytes of essentially any type.

So you can think of "A" as standing for "ANY" instead of "ASCII". It
allows for writing binary data to a formatted file. Although today mixing
binary and formatted data as is often done with high-level metadata
files is best done with a stream file it is still used this way. Although
it can now be done with CHAR(N) it has been a long-standing method for
writing ANSI terminal escape sequences, for example.

The case that five other compilers print this programs' output identically
hopefully bolsters this description and explains the length of the MRE
when just one executable line would suffice to show an issue exists. The
different lines are meant to bolster the idea and show some real-world
uses as well.

When used with internal WRITE statement you can perhaps think of it as
an early version of the TRANSFER() function; or as a TRANSFER() function
to writes to output instead of storing into a variable?

The lines beginning with a dash (-) are producing incorrect output.
```fortran
program mre
use, intrinsic :: iso_fortran_env, only : int8, int16, int32, int64
implicit none
integer                       :: abcd_int           
character(len=*),parameter    :: fmt=               '(g0,t30,a,t40,b32.32)'
integer,parameter             :: byte=kind(0_int8)  
integer(kind=byte)            :: iletter(26)        
integer(kind=byte),parameter  :: esc=27             
integer(kind=int64)           :: iword              
integer                       :: i                  

   ! make native integer value with bit patterns
   ! that happen to be the same as the beginning of the alphabet

   write(*,'(*(a))')65,66,67,68,'ABCD'
   write(*,'(*(a))')65_byte,66_byte,67_byte,68_byte
   write(*,'(*(a1))')65,66,67,68,'A','B','C','D'

   abcd_int=transfer('abcd',0)
   ! show the value as integer, as string, and as byte pattern
   write(*,fmt)abcd_int,abcd_int,abcd_int

   write(*,'(a,1x,a4)')abcd_int,abcd_int

   ! home cursor and clear screen on vt102-compatible (ANSI) terminals
   write(*,'(*(a))',advance='no')esc,'[H',esc,'[2J',esc,'[3J'

   iword=transfer('ABCDEFGH',0_int64)
   write(*,'(a)')iword

   do i=1,26
      iletter(i)=64+i
   enddo
   write(*,'(*(a))')iletter

end program mre
```
```diff
+aoccflang520         :     1	STDOUT:
+aoccflang520         :     2	A^@^@^@B^@^@^@C^@^@^@D^@^@^@ABCD
+aoccflang520         :     3	ABCD
+aoccflang520         :     4	ABCDABCD
+aoccflang520         :     5	1684234849                   abcd      01100100011000110110001001100001
+aoccflang520         :     6	abcd abcd
+aoccflang520         :     7	ABCDEFGH
+aoccflang520         :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ
+flangtrunk           :     2	A^@^@^@B^@^@^@C^@^@^@D^@^@^@ABCD
+flangtrunk           :     3	ABCD
+flangtrunk           :     4	ABCDABCD
+flangtrunk           :     5	1684234849                   abcd      01100100011000110110001001100001
+flangtrunk           :     6	abcd abcd
+flangtrunk           :     7	ABCDEFGH
+flangtrunk           :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ

+ifxlatest            :     2	A^@^@^@B^@^@^@C^@^@^@D^@^@^@ABCD
+ifxlatest            :     3	ABCD
+ifxlatest            :     4	ABCDABCD
+ifxlatest            :     5	1684234849                   abcd      01100100011000110110001001100001
+ifxlatest            :     6	abcd abcd
+ifxlatest            :     7	ABCDEFGH
+ifxlatest            :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ
-lfortranlatest       :     2	ABCDABCD
+lfortranlatest       :     3	ABCD
+lfortranlatest       :     4	ABCDABCD
-lfortranlatest       :     5	1684234849                   a         01100100011000110110001001100001
-lfortranlatest       :     6	a    a
-lfortranlatest       :     7	A
+lfortranlatest       :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ
+nvfortran_x86_26_3   :     2	A^@^@^@B^@^@^@C^@^@^@D^@^@^@ABCD
+nvfortran_x86_26_3   :     3	ABCD
+nvfortran_x86_26_3   :     4	ABCDABCD
+nvfortran_x86_26_3   :     5	1684234849                   abcd      01100100011000110110001001100001
+nvfortran_x86_26_3   :     6	abcd abcd
+nvfortran_x86_26_3   :     7	ABCDEFGH
+nvfortran_x86_26_3   :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ
+gfortran161          :     2	A^@^@^@B^@^@^@C^@^@^@D^@^@^@ABCD
+gfortran161          :     3	ABCD
+gfortran161          :     4	ABCDABCD
+gfortran161          :     5	1684234849                   abcd      01100100011000110110001001100001
+gfortran161          :     6	abcd abcd
+gfortran161          :     7	ABCDEFGH
+gfortran161          :     8	ABCDEFGHIJKLMNOPQRSTUVWXYZ
```

opixdown:
## Summary

The `A` edit descriptor with an integer I/O list item must do a binary transfer of the value's internal representation, as a character string whose length equals the kind width (4 bytes for `integer(4)`, 8 for `integer(8)`, etc.). The runtime formatter emitted only a single byte, so:

```fortran
integer :: word
character(4) :: out
word = transfer("abcd", 0)
write(out, "(a)") word   ! produced "a   ", expected "abcd"
```

GFortran, Flang, ifx, nvfortran and AOCC Flang all agree on the expected behavior.

Closes #12400

## Scope

Single narrow change in `_lcompilers_string_format_fortran` (`src/libasr/runtime/lfortran_intrinsics.c`): the integer-with-`A` branch now copies the argument's bytes with length equal to the kind width — leftmost bytes when the field width is narrower (`a2` → `ab`), right-justified with blanks when wider (`a6` → `  abcd`).

A separate pre-existing bug (NUL bytes truncated when writing to stdout, which also affects plain `character` items containing `char(0)`) is intentionally not addressed here.

## Verification

- New integration test `integration_tests/format_108.f90` (labels `gfortran`, `llvm`) covers default integer, `integer(int64)`, field narrower/wider than the kind width, and the zero bytes of the representation.
- Verified the test fails on `main` (`ERROR STOP 1`) and passes on this branch; it also compiles and runs clean with GFortran 16.1.
- Compared the full reproducer from #12400 against GFortran: all lines involving internal writes and non-NUL output now match byte-for-byte.


Jatinagarwal24 (approved):
Fix Looks good 
Just can you add the test cases for these 
` integer(int8), integer(int16), and integer arrays with A editing `

HarshitaKalani (approved):

PR #12410, opened 2026-08-01
fix: honor SP sign mode in G editing

### LINKED ISSUE #12408: SP edit descriptor not producing plus sign prefix on positive numbers printed with G0 format in version 0.64.0
The SP edit descriptor is being ignored by the G0 format.  All numbers
should be prefixed with a plus sign character but the ones printed with
G0 lack the sign character.

The 5.0 value should be preceded by "+" for the first line of numbers, and all the numbers on the second line are missing the plus sign.
```fortran
program mre
use, intrinsic :: iso_fortran_env, only : compiler_version
implicit none
   print '(SP,g0,1x,i0,1x,f8.2)', 5.0, 10, 23.45
   print '(SP,*(g0,1x))', 5.0, 10, 23.45
end program mre
```
## platforms
```diff
+ This file was compiled by flang AOCC AOCC_5.2.0-Build#2035 2026_04_10
+ +5.000000 +10   +23.45
+ +5.000000 +10 +23.45000
+ This file was compiled by flang version 23.0.0 (https://github.com/llvm/llvm-project.git 8d61da2569a975346150371cb0d317fb9339be0c)
+ +5. +10   +23.45
+ +5. +10 +23.45
+ This file was compiled by Intel(R) Fortran Compiler for applications running on Intel(R) 64, Version 2025.3.2 Build 20260112
+ +5.000000 +10   +23.45
+ +5.000000 +10 +23.45000
+ This file was compiled by LFortran version 0.64.0
- 5.00000000 +10   +23.45
- 5.00000000 10 23.4500008
+ This file was compiled by nvfortran 26.3-0
+ +5.000000 +10   +23.45
+ +5.000000 +10 +23.45000
+ This file was compiled by GCC version 16.1.0
+ +5.00000000 +10   +23.45
+ +5.00000000 +10 +23.4500008
```

opixdown:
Closes #12408


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3695,6 +3695,12 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(lfortran_allocator_t* al, c
                                 }
                             }
                         }
+                        // SP editing: force a plus sign on non-negative values.
+                        if (is_SP_specifier && !isnan(double_val) && double_val >= 0.0 &&
+                                strlen(formatted) + 1 < sizeof(formatted)) {
+                            memmove(formatted + 1, formatted, strlen(formatted) + 1);
+                            formatted[0] = '+';
+                        }
```
Jatinagarwal24, line 3703: Change the condition from testing the value to testing the already-formatted string, since `-0.0 >= 0.0` is true in C but `%f prints -0.00`:
```
// current (line 3699–3700)
if (is_SP_specifier && !isnan(double_val) && double_val >= 0.0 &&
        strlen(formatted) + 1 < sizeof(formatted)) {

// suggested
if (is_SP_specifier && !isnan(double_val) && formatted[0] != '-' &&
        strlen(formatted) + 1 < sizeof(formatted)) {

```

integration_tests/format_109.f90
Jatinagarwal24, line 1: Pls add an E-mode case `(write(out,"(sp,g0)") 1.0e10 `→ `"+1.00000000E+10")`, a negative-zero case (`sp,g12.4` with `-0.0` this is the regression test for item 1), and a plain s restore alongside at line 20.

src/libasr/runtime/lfortran_intrinsics.c
Jatinagarwal24, line 1: Pls once check this 
At line 3629: `if (width > 0 && (int)strlen(inf_long) > width)`. With SP, +Infinity is 9 chars, so sp,g8 overflows instead of falling back to +Inf. The check would need + (is_SP_specifier && double_val > 0 ? 1 : 0) on the length.

src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -3695,6 +3695,12 @@ LFORTRAN_API char* _lcompilers_string_format_fortran(lfortran_allocator_t* al, c
                                 }
                             }
                         }
+                        // SP editing: force a plus sign on non-negative values.
+                        if (is_SP_specifier && !isnan(double_val) && double_val >= 0.0 &&
+                                strlen(formatted) + 1 < sizeof(formatted)) {
+                            memmove(formatted + 1, formatted, strlen(formatted) + 1);
+                            formatted[0] = '+';
```
Jatinagarwal24, line 3702: The E-mode branch strips a leading 0. when the field exceeds width, but the + is added afterwards. So `sp,g11.5 `of 0.85734e8 becomes` "+0.85734E+08"` (12 chars in an 11-wide field), whereas the standard-conforming output is `"+.85734E+08"`
So  after the formatted[0] = '+'; at line 3702, re-apply the strip for the +0.
```
if (width > 0 && (int)strlen(formatted) > width &&
        formatted[1] == '0' && formatted[2] == '.') {
    memmove(formatted + 1, formatted + 2, strlen(formatted) - 1);
}
```

Jatinagarwal24 (commented):
Pls address the comments .
Thanks.

PR #12419, opened 2026-08-02
CI: test_caffeine.sh: Cherry-pick recent fixes to assertion reporting

bonachea:
Caffeine 0.8.0's assertion failure path for LFortran was previously hitting a segmentation fault (with no output). Consequently Caffeine's many assertion checks were not being correctly reported when triggered under LFortran CI.

Cherry-pick the relevant fixes from https://github.com/BerkeleyLab/caffeine/pull/353 to ensure high-quality assertion failure output with LFortran.

### DIFF
--- a/ci/test_caffeine.sh
+++ b/ci/test_caffeine.sh
@@ -74,6 +74,14 @@ git config user.email "nobody@nowhere.com"
 git config user.name  "Nobody"
 git cherry-pick 736130c4af77b4ab33e4341e6dcd32ab4c8b7f4a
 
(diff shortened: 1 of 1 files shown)


certik (approved):
This is fine. But why cannot we simply upgrade the Caffeine version we use, so that we check out a particular commit and don't need to patch things?

bonachea:
>  why cannot we simply upgrade the Caffeine version we use, so that we check out a particular commit and don't need to patch things?

That would have also worked. The cherry-pick approach keeps the code closer to the official release, but in this case the difference is not that significant in any case.

Next advance will probably use the next Caffeine patch release, or I'll switch to bare commit checkout.

Issue #12418, opened 2026-08-02
Feat: A better way to process version string instead of ci/version.sh


Syize:
## Description of this issue

At present, building and uploading LFortran requires `version` file generated by `ci/verison.sh`. But:

### `ci/version.sh` will fail if the repo contains no tag

If developer forks LFortran, then develop LFortran based on his/her own fork, the forked repository doesn't have any tag, and `ci/version.sh` will fail to generate `version`.

### `ci/version.sh` isn't necessary for CI

We can write version string to [GITHUB_ENV](https://docs.github.com/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable).

### `ci/version.sh` hardcodes `version` file path, which makes changing cmake build dir harder

`ci/version.sh` generates `version` at the root directory of LFortran project, so cmake must build LFortran at root directory since it depends on `version`. And build files from cmake will be generated to LFortran src dir directly.

## A better way to process version string

I think we can process and get version string in a more flexible way. 

- In CI, we get version string, and write it to [GITHUB_ENV](https://docs.github.com/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable).
- In cmake, try to get version string directly from git, then try `version` file. If both fail, use a default dev string like "0.1.0-dev".
- When creating tarball, generating `version` so users can build LFortran from source tarball.

Thus we can minimize the dependence on `version`, and change build path of cmake conveninetly.

## Plan to do

Here is the plan to achieve this goal:

1. Modify `CMakeLists.txt`, add if block and `execute_process` to try to get version from git first, then fallback to `version` file at project root dir.
2. Modify `.github/workflow` files, add command to get version from git and write it to `GITHUB_ENV`.
3. Update related CI scripts to read version from environmental variable.
4. Update build scripts to change cmake build path to an isolated dir `build`.

PR #12420, opened 2026-08-02
Feat: Improve version string processes of build workflow.

Syize:
This commit implements changes introduced by issue #12418:

- CMakeLists.txt is updated to try to run git command first to get version, then try to read ${PROJECT}/version, then fallback to default dev version "0.1.0-dev". If environment variable "LFORTRAN_VERSION" is defined, use it directly.
- .github/workflows are updated to store version string to GITHUB_ENV in each job, and then pass it to cmake or CI scripts. When creating tarball, $dest/version will be created.
- Related CI scripts are updated to accept new argument, or read LFORTRAN_VERSION environment variable.
- Build scripts are updated, calls to ci/version.sh are removed, and cmake is configured to build in isolated build directory.
- ci/version.sh is deleted.


assem2002:
Thanks for the PR.

Based on the description here, We 2 main points :=
* 1 ) ci/version.sh will fail if the repo contains no tag
    * It's reasonable concern. Having no tag shouldn't block us from building the project locally.
    * The fix here is trying to look for the version (either it's a env variable, or the file), and if not falls back to some default. 
    * We do have a fallback; based on Claude, we have a fallback at `src/libasr/CMakeLists.txt`. apparently, it's misplaced.
* 2) ci/version.sh hardcodes version file path, which makes changing cmake build dir harder
    * I don't exactly get the issue for this. Can you elaborate?
 
I think the first problem is the main one, Can you describe what exactly you think is a problem and the outcome you want after fixing it?

Syize:
@assem2002 Sorry for the late reply. The first problem is the main one indeed. When building locally, the local repo may don't have tags in some cases (as I have described in issue #12418, if developers fork repo on github, then clone their own forks). When I build locally, the build script is failed because it can't get any tag, and I have to set it manually in `ci/version.sh`. If we have a fallback version string, no additional change is needed.

> We do have a fallback; based on Claude, we have a fallback at `src/libasr/CMakeLists.txt`. apparently, it's misplaced.

I didn't notice this fallback. Maybe we can use this rather than add duplicate functionality.

As for the second point, it is a small problem. Currently we use project directory as cmake's build directory, and `version` file is generated to project directory. It's Okey, but cache files (`*.cmake` file, `CMakeFiles` dir, etc) will be generated to the source dir.  I think it would be better if we use isolated build directory : )

PR #12436, opened 2026-08-04
Ship the Fortran display module with the Xeus runtime

anutosh491:


certik:
@anutosh491 let's rebase this and then run full tests on this one. I also want to play with it locally, since it changes cmake.

certik:
I like what this does, I think this is in the right direction. I just need to test it manually locally, to ensure it works well.

anutosh491:
> I just need to test it manually locally, to ensure it works well.

Thank you. Let me know if you find any issues !

certik (approved):
I didn't have time to check locally, but the changes in cmake look good. So let's merge it.

PR #12439, opened 2026-08-04
Fix: RANDOM_SEED() with no args must reseed and refresh state (#12428)

### LINKED ISSUE #12428: intrinsic RANDOM_SEED() always returns an array of zeros when the seed is queried in version 0.64.0
#6250  appears to imply the intrinsic RANDOM_SEED() is fully implemented. 
If it is supposed to be ready (and comply with the f2023 standard):

There seem to be several issues with the random number procedures
conforming to the f2023 standard, but one definite issue is that
a query of the seed is always returning zeros. And if you set the
key to zeros you get a difference sequence than initially, 
proving the key is not actually set to zero. 

### Fortran source
[details="click to see more..."]
```fortran
program demo_random_seed
   implicit none
   integer, allocatable :: seed(:),initial_seed(:)
   integer :: i,n
   real :: x(3)
   !call random_seed() ! set random seed if f2023

   call random_seed(size = n)
   allocate(seed(n))
   call random_seed(get=seed)
   initial_seed=seed

   write (*, *) 'queried initial seed=',seed
   write (*,*) 'get three sets of random numbers'
   do i=1,3
      call random_number(x)
      write(*,*)x
   enddo

   ! now randomize the seed several times, query
   ! and print it, and then generate an array of PRN
   do i=1,3
      call random_seed() ! randomize seed if f2023
      call random_seed(get=seed)
      write (*, *) 'new seed=',seed
      call random_number(x)
      write(*,*)'set with new seed=',x
   enddo

   ! now go back to initial seed and should reproduce
   ! initial set
   write(*,*)'back to initial'
   call random_seed(put=initial_seed)

   ! repeat first display 
   call random_seed(get=seed)
   write (*, *) 'queried current seed=',seed
   write (*,*) 'get three sets of random numbers, should be duplicates of first set'
   do i=1,3
      call random_number(x)
      write(*,*)x
   enddo
end program demo_random_seed
```
```diff
-lfortranlatest       :     2	 queried initial seed= 0 0 0 0 0 0 0 0
+lfortranlatest       :     3	 get three sets of random numbers
+lfortranlatest       :     4	 0.352760732 0.789589584 0.905825853
+lfortranlatest       :     5	 4.38874364E-02 0.954264939 0.248271078
+lfortranlatest       :     6	 0.867982030 0.838708162 0.922090709
-lfortranlatest       :     7	 new seed= 0 0 0 0 0 0 0 0
+lfortranlatest       :     8	 set with new seed= 0.653399765 0.550067544 0.269132197
-lfortranlatest       :     9	 new seed= 0 0 0 0 0 0 0 0
+lfortranlatest       :    10	 set with new seed= 0.682013214 0.383551717 0.901754320
-lfortranlatest       :    11	 new seed= 0 0 0 0 0 0 0 0
+lfortranlatest       :    12	 set with new seed= 0.643302083 0.916349769 0.577012718
+lfortranlatest       :    13	 back to initial
-lfortranlatest       :    14	 queried current seed= 0 0 0 0 0 0 0 0
+lfortranlatest       :    15	 get three sets of random numbers, should be duplicates of first set
-lfortranlatest       :    16	 0.840187728 0.394382924 0.783099234
-lfortranlatest       :    17	 0.798440039 0.911647379 0.197551370
-lfortranlatest       :    18	 0.335222751 0.768229604 0.277774721
+gfortran161          :     2	 queried initial seed=    -2447268 -1354129540   374710663 -1240608696 -1827405339  -257388164  2056470833   269047911
+gfortran161          :     3	 get three sets of random numbers
+gfortran161          :     4	  0.728326082      0.733394623      0.807955265    
+gfortran161          :     5	  0.827496469      0.709796131      0.855553031    
+gfortran161          :     6	  0.850020826       2.29641199E-02  0.848301649    
+gfortran161          :     7	 new seed=   532671634  1589724431  -702344385  -267089641 -2127795903  1724481233 -1649777043  -673546294
+gfortran161          :     8	 set with new seed=  0.182621956      0.814420581      0.161144853    
+gfortran161          :     9	 new seed=  -588602105 -1330109958 -1909200428  2013740993   926921249   132638128 -1513550047 -1366162835
+gfortran161          :    10	 set with new seed=  0.922367275      0.226936579      0.626253545    
+gfortran161          :    11	 new seed=   701608307 -1105671482   804077484    93609417 -2030753861    94338487 -1850184744   417402487
+gfortran161          :    12	 set with new seed=  0.119894266      0.515091896      0.171295166    
+gfortran161          :    13	 back to initial
+gfortran161          :    14	 queried current seed=    -2447268 -1354129540   374710663 -1240608696 -1827405339  -257388164  2056470833   269047911
+gfortran161          :    15	 get three sets of random numbers, should be duplicates of first set
+gfortran161          :    16	  0.728326082      0.733394623      0.807955265    
+gfortran161          :    17	  0.827496469      0.709796131      0.855553031    
+gfortran161          :    18	  0.850020826       2.29641199E-02  0.848301649    
```

usefahmed07:
closes #12428

`call random_seed()` with no arguments had no effect — the internal
seed buffer was never updated, so a following `get=` always returned
zeros even though `random_number()` kept advancing.

Fixed by making the no-args case call the same runtime routine
`random_init` uses to reseed and refresh the buffer. Also removed an
unintended `srand(seed)` call that fired just from querying `size=`.

Extended `integration_tests/random_seed_01.f90` to check that `get=`
matches a prior `put=`, and that a bare `random_seed()` call actually
changes the seed.

### DIFF
--- a/integration_tests/random_seed_01.f90
+++ b/integration_tests/random_seed_01.f90
@@ -4,6 +4,22 @@ program random_seed_01
     call random_seed(size=sz)
     if (sz <= 0) error stop "random_seed size must be positive"
     block
+        integer, allocatable :: got0(:)
+        allocate(got0(sz))
+        ! random_seed(get=...) called before any prior put/bare random_seed()
+        ! must not observe the seed buffer's initial all-zero state
+        ! (regression test for #12428)
+        call random_seed(get=got0)
+        block
+            logical :: all_zero
+            all_zero = .true.
+            do i = 1, sz
+                if (got0(i) /= 0) all_zero = .false.
+            end do
+            if (all_zero) error stop "random_seed(get=...) returned all zeros before any prior seed"
+        end block
+    end block
+    block
         integer, allocatable :: vals(:), got(:)
         allocate(vals(sz), got(sz))
         vals = 42
@@ -18,12 +34,12 @@ program random_seed_01
         call random_seed()
         call random_seed(get=got)
         block
-            logical :: all_zero
-            all_zero = .true.
+            logical :: unchanged
+            unchanged = .true.
             do i = 1, sz
-                if (got(i) /= 0) all_zero = .false.
+                if (got(i) /= vals(i)) unchanged = .false.
             end do
-            if (all_zero) error stop "random_seed() did not reseed (got all zeros)"
+            if (unchanged) error stop "random_seed() did not reseed"
         end block
     end block
 end program random_seed_01
--- a/integration_tests/random_seed_01.f90
+++ b/integration_tests/random_seed_01.f90
@@ -1,12 +1,29 @@
 program random_seed_01
     implicit none
-    integer :: sz
+    integer :: sz, i
     call random_seed(size=sz)
     if (sz <= 0) error stop "random_seed size must be positive"
     block
-        integer :: vals(sz)
+        integer, allocatable :: vals(:), got(:)
+        allocate(vals(sz), got(sz))
         vals = 42
         call random_seed(put=vals)
-        call random_seed(get=vals)
+        call random_seed(get=got)
+        do i = 1, sz
+            if (got(i) /= vals(i)) error stop "random_seed get did not match put"
+        end do
+
+        ! bare `random_seed()` must reseed and refresh the internal state,
+        ! not leave it unchanged (regression test for #12428)
+        call random_seed()
+        call random_seed(get=got)
+        block
+            logical :: all_zero
+            all_zero = .true.
+            do i = 1, sz
+                if (got(i) /= 0) all_zero = .false.
+            end do
+            if (all_zero) error stop "random_seed() did not reseed (got all zeros)"
+        end block
     end block
 end program random_seed_01
--- a/src/libasr/pass/intrinsic_subroutines.h
+++ b/src/libasr/pass/intrinsic_subroutines.h
@@ -229,6 +229,34 @@ namespace RandomSeed {
             fill_func_arg_sub("get", real32, In);
         }
 
+        if (is_real(*arg_types[0]) && is_real(*arg_types[1]) && is_real(*arg_types[2])) {
+            // `call random_seed()` with no arguments must reseed the RNG and
+            // refresh the internal seed buffer, matching
+            // `random_init(repeatable=.false., image_distinct=.false.)`.
+            std::string c_func_name_2 = "_lfortran_random_init";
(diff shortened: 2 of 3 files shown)


integration_tests/random_seed_01.f90
```diff
@@ -1,12 +1,29 @@
...
+        ! not leave it unchanged (regression test for #12428)
+        call random_seed()
+        call random_seed(get=got)
+        block
+            logical :: all_zero
+            all_zero = .true.
+            do i = 1, sz
+                if (got(i) /= 0) all_zero = .false.
+            end do
+            if (all_zero) error stop "random_seed() did not reseed (got all zeros)"
```
opixdown, line 26: can we change it to something like `if (all(got == vals)) error stop "random_seed() did not reseed"` as it would directly test that the bare call actually changed the seed.

src/libasr/pass/intrinsic_subroutines.h
```diff
@@ -229,6 +229,34 @@ namespace RandomSeed {
             fill_func_arg_sub("get", real32, In);
         }
 
+        if (is_real(*arg_types[0]) && is_real(*arg_types[1]) && is_real(*arg_types[2])) {
```
opixdown, line 232: the initial `random_seed(get=seed)` issue reported in #12428 still seems to remain. The seed buffer is initially zero, so a `get` before any `put` or bare `random_seed()` will still return zeros

integration_tests/random_seed_01.f90
usefahmed07: Good suggested i am on it

opixdown:
@usefahmed07 can you address the comments, thank you for the pr.

certik:
We will need to very carefully check that the test is good and that the new behavior is correct, as the implementation changed a lot. I can't verify it just by looking at the code, so needs to be done by hand.

usefahmed07:
> We will need to very carefully check that the test is good and that the new behavior is correct, as the implementation changed a lot. I can't verify it just by looking at the code, so needs to be done by hand.

Manually verified beyond the code read: ran the built binary multiple times lazy-init GET differs each run (real time-based seed, not a stuck constant), a bare random_seed() reliably changes state, put→get round-trips stably, and put with the same seed reproduces the same random_number() output (repeatability preserved). random_seed_01 also passes. You can verify it yourself any time.

usefahmed07:
@Jainam-not-a-robot

usefahmed07:
@assem2002

PR #12442, opened 2026-08-05
Fix: allow BOZ literal arguments in MERGE_BITS() (#12416)

### LINKED ISSUE #12416: Intrinsic MERGE_BITS() will not allow BOZ literal arguments in 0.64.0
Per the Fortran 2003 standard, BOZ literals may be used for the MASK
and one of either I or J.

I think #1502 (references similar issues that no longer appear to be
a problem) could be closed as well, as so far I have found no issues
with other intrinsics that take BOZ arguments; just MERGE_BITS().

From the standard:

      16.9.140 MERGE_BITS (I, J, MASK)

      Description. Merge of bits under mask.

      Class. Elemental function.

      Arguments.

      I      shall be of type integer or a boz-literal-constant.

      J      shall be of type integer or a boz-literal-constant. If both
             I and J are of type integer they shall have
             the same kind type parameter. I and J shall not both be
             boz-literal-constants.

     MASK    shall be of type integer or a boz-literal-constant. If MASK
             is of type integer, it shall have the same
             kind type parameter as each other argument of type integer.

     Result Characteristics.

     Same as I if I is of type integer; otherwise, same as J.

     Result Value. If any argument is a boz-literal-constant, it is
     first converted as if by the intrinsic function

     INT to the type and kind type parameter of the result. The result
     has the value of

         IOR (IAND (I, MASK), IAND (J, NOT (MASK))).

```fortran
program demo_merge_bits
use,intrinsic :: iso_fortran_env,  only : int8, int16, int32, int64
implicit none
character(len=*),parameter :: fmt='(*(g0, 1X))'

  ! ! using BOZ values
   print fmt, &
   & merge_bits(32767_int16, o'12345',    32767_int16),         &
   & merge_bits(o'12345',    32767_int16, b'0000000000010101'), &
   & merge_bits(32767_int16, o'12345',    z'1234')

  ! additional intrinsics that accept BOZ
  !print *, dble(z"ff")
  !print *, real(z"ff")
  !print *, cmplx(z"ff")
  !print *, int(z"ff")
  !print *, bge(z"ff",o"777")
  !print *, ble(z"ff",o"777")
  !print *, blt(z"ff",o"777")
  !print *, iand(int(b"11111000000000000000000000000000"), b"00000111111111111111111111111111")
  !print *, ieor(int(b"11111000000000000000000000000000"), b"00000111111111111111111111111111")
  !print *, ior(int(b"11111000000000000000000000000000"), b"00000111111111111111111111111111")
  !print *, dshiftl(int(b"11111000000000000000000000000000"), b"00000111111111111111111111111111",8)
  !print *, dshiftr(int(b"11111000000000000000000000000000"), b"00000111111111111111111111111111",7)

end program demo_merge_bits
```
   + works with flang, gfortran, ifx
   + lfortran fails to compile, not allowing BOZ constants
   + AOCC flang and nvfortran fail to compile
     supporting MERGE_BITS() yet.
```diff
+flangtrunk           :    2    32767 32751 5877
+ifxlatest            :    2    32767 32751 5877
+gfortran161          :    2    32767 32751 5877
-aoccflang520         :    2    F90-S-0074-Illegal number or type of arguments to merge_bits - keyword argument i (/app/example.f90: 7)
-aoccflang520         :    3    F90-S-0074-Illegal number or type of arguments to merge_bits - keyword argument i (/app/example.f90: 7)
-aoccflang520         :    4    F90-S-0074-Illegal number or type of arguments to merge_bits - keyword argument i (/app/example.f90: 7)
-aoccflang520         :    5      0 inform,   0 warnings,   3 severes, 0 fatal for demo_merge_bits
-lfortranlatest       :     3   semantic error: Type and kind of the relevant arguments of Mergebits must be the same
-lfortranlatest       :     4    --> /app/example.f90:8:5
-lfortranlatest       :     5     |
-lfortranlatest       :     6   8 |    & merge_bits(32767_int16, o'12345',    32767_int16),         &
-lfortranlatest       :     7     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-lfortranlatest       :     8
-lfortranlatest       :     9
-lfortranlatest       :    10   Note: Please report unclear, confusing or incorrect messages as bugs at
-lfortranlatest       :    11   https://github.com/lfortran/lfortran/issues.
-nvfortran_x86_26_3   :     1   STDERR:
-nvfortran_x86_26_3   :     2   NVFORTRAN-S-0038-Symbol, merge_bits, has not been explicitly declared (/app/example.f90: 1)
-nvfortran_x86_26_3   :     3     0 inform,   0 warnings,   1 severes, 0 fatal for demo_merge_bits
```

usefahmed07:
Fixes #12416

MERGE_BITS(I, J, MASK) rejected BOZ literal arguments, even though the standard allows a BOZ literal for MASK and one of I/J.

Root cause: BOZ literals were always typed with the default integer kind, with no promotion to match a sibling integer argument's kind in the same call — so a call like merge_bits(32767_int16, o'12345', 32767_int16) failed with a kind mismatch error.

Fix: added a helper that promotes a BOZ argument's kind to match the other integer argument, applied to iand, ior, ieor, merge_bits, dshiftl, dshiftr. Also added a check rejecting merge_bits calls where I and J are both BOZ (not allowed by the standard).

Added a runtime test for the fix and a regression test for the both-BOZ rejection.


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -10553,6 +10553,55 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
     }
 
+    // Per Fortran standard, some elemental intrinsics accept a BOZ literal
+    // for one or more of their integer arguments. visit_BOZ() always types
+    // BOZ constants using compiler_options.po.default_integer_kind, so if
+    // another (non-BOZ) integer argument in the same call has a different
+    // kind, the intrinsic's equal-type/kind check fails spuriously. This
+    // promotes any BOZ argument's kind to match the first non-BOZ integer
+    // argument found in the call.
+    void fix_boz_kind_for_intrinsic(const std::string& intrinsic_name, Vec<ASR::expr_t*>& args) {
```
opixdown, line 10563: it also handles `dshiftl/dshiftr`, but the "both I and J are BOZ" restriction is only enforced for merge_bits. we should add the same diagnostic for `dshiftl/dshiftr` as required by the standard

integration_tests/merge_bits_boz.f90
```diff
@@ -0,0 +1,15 @@
+program merge_bits_boz
+use, intrinsic :: iso_fortran_env, only: int16
+implicit none
+integer(int16) :: r1, r2, r3
+
```
opixdown, line 6: Could you also add regression tests for dshiftl, dshiftr, iand, ior, and ieor?

tests/errors/continue_compilation_1.f90
```diff
@@ -951,6 +951,7 @@ end subroutine equivalence_common_array_overrun
     subroutine set_caller(this)
         class(MyClass) :: this
     end subroutine
+    subroutine merge_bits_boz_both_args(); print *, merge_bits(o'12345', z'1234', 32767); end subroutine
```
opixdown, line 954: fix the indentation, so the diff will be minimum

tests/tests.toml
```diff
@@ -5240,6 +5240,9 @@ run = true
 [[test]]
 filename = "../integration_tests/shadow_same_symbol.f90"
 run = true
+[[test]]
+filename = "../integration_tests/merge_bits_boz.f90"
```
opixdown, line 5244: we should add it in integration test

tests/errors/continue_compilation_1.f90
usefahmed07: very good suggested thanks

integration_tests/merge_bits_boz.f90
usefahmed07: on it now

tests/tests.toml
```diff
@@ -5240,6 +5240,9 @@ run = true
...
+run = true
```
opixdown, line 5245: this is a postive test means it is giving correct output so we should add it in cmake, and in test file verify it with `if(/)`

tests/errors/continue_compilation_1.f90
```diff
@@ -951,6 +951,9 @@ end subroutine equivalence_common_array_overrun
...
+    subroutine dshiftl_boz_both_args(); print *, dshiftl(o'12345', z'1234', 5); end subroutine
```
opixdown, line 955: the indentation is not correct see above(`set_caller`) how it is written

tests/tests.toml
opixdown: remove it from the toml and add it to cmake

opixdown: just do this and resolve conflict otherwise its good now, thankyou

usefahmed07:
Hi @opixdown Can you review it Thanks!

PR #12450, opened 2026-08-07
fix: expand implied-do loop bounds with non-literal size expressions 

### LINKED ISSUE #10561: (*(1X,A)) format with implied-do output bug
Test program different from the one in #10560 only in the format
```
  implicit none
  integer i
  character(8):: strings(2) = ['Hello   ','World   ']
  print "(*(1X,A))",(trim(strings(i)),i=1,size(strings)),'(Implied-do I/O)'
end program
```
Output from lfortran 0.61.0 used 2 lines and omitted the last item
```
 Hello
 World
```
Output from gfortran
```
 Hello World (Implied-do I/O)
```

HarshilShah1804:
Fixes #10561

### DIFF
--- a/src/libasr/pass/implied_do_loops.cpp
+++ b/src/libasr/pass/implied_do_loops.cpp
@@ -1118,21 +1118,68 @@ class ArrayConstantVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayC
             }
         };
         
+        // Attempt to evaluate an expression to a compile-time integer constant.
+        // Returns true and sets `val` on success.
+        bool try_eval_int_const(ASR::expr_t* expr, int64_t& val) {
+            if (!expr) return false;
+            // Direct integer constant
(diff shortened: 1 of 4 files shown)


src/libasr/pass/implied_do_loops.cpp
```diff
@@ -1118,21 +1118,68 @@ class ArrayConstantVisitor : public ASR::CallReplacerOnExpressionsVisitor<ArrayC
...
+                }
+            }
+            // Fallback: use constant-folded value stored by the semantic pass
+            ASR::expr_t* v = ASRUtils::expr_value(expr);
+            if (v && ASR::is_a<ASR::IntegerConstant_t>(*v)) {
+                val = ASR::down_cast<ASR::IntegerConstant_t>(v)->m_n;
+                return true;
+            }
+            return false;
+        }
```
certik, line 1165: What does it mean when this fails, below?

I see that expand_implied_do_loop_flat will return false? What happens next?

PR #12456, opened 2026-08-08
fix: initialize elem_len for non-allocatable ExpressionLength string dummy args

### LINKED ISSUE #12153: Bug in optional string type arguments
MRE
```fortran
program optional_string_arg_01
    implicit none
    call check(1)
contains
    subroutine check(n, s)
        integer, intent(in) :: n
        character(len=n), optional, intent(in) :: s
        if (len(s) /= 1) error stop
    end subroutine
end program optional_string_arg_01

```

LFortran
```console
ERROR STOP
```

HarshilShah1804:
Fixes #12153

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4145,6 +4145,7 @@ RUN(NAME optional_10 LABELS gfortran llvm)
 RUN(NAME optional_11 LABELS gfortran llvm)
 RUN(NAME optional_12 LABELS gfortran llvm)
 RUN(NAME optional_13 LABELS gfortran llvm)
+RUN(NAME optional_string_arg_01 LABELS gfortran llvm)
 
 RUN(NAME end_name_match LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 
--- a/integration_tests/optional_string_arg_01.f90
+++ b/integration_tests/optional_string_arg_01.f90
@@ -1,0 +1,32 @@
+program optional_string_arg_01
+   implicit none
+   character(10) :: a
+
+   ! Test absent case
+   call check(1)
+   
+   ! Test present case
+   call check(1, 'a')
+
+   ! Test caller-unchanged check
+   a = "1234567890"
+   call sub(3, a)
+   if (len(a) /= 10) error stop
+   if (a /= "1234567890") error stop
+
+contains
+   subroutine check(n, s)
+      integer, intent(in) :: n
+      character(len=n), optional, intent(in) :: s
+      if (len(s) /= 1) error stop
+      if (present(s)) then
+         if (s /= 'a') error stop
+      end if
+   end subroutine
+
+   subroutine sub(n, s)
+      integer, intent(in) :: n
+      character(len=n), optional, intent(in) :: s
+      if (len(s) /= 3) error stop
+   end subroutine
+end program optional_string_arg_01
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8516,17 +8516,22 @@ public:
             }
             // For non-allocatable ExpressionLength string dummy arguments (including optional
             // ones like `character(len=n), optional`), when the argument is absent, the call
-            // site may pass a zeroed dummy descriptor (elem_len = 0) (e.g. from 
+            // site may pass a zeroed dummy descriptor (elem_len = 0) (e.g. from
             // transform_optional_argument_functions) or a NULL pointer (e.g. on bind(C) paths).
             // We create a local copy of the descriptor to safely set the declared length
             // without mutating the caller's descriptor, and guard the copy with a null-check.
-            if (ASRUtils::get_FunctionType(x)->m_abi == ASR::abiType::BindC &&
+            // Bind(C) functions are excluded: their string dummies are not passed as
+            // by-reference string descriptors (e.g. `character, value` is a by-value
+            // argument), so there is no descriptor to copy or fix up.
+            if (ASRUtils::get_FunctionType(x)->m_abi != ASR::abiType::BindC &&
                 ASRUtils::is_string_only(symbol_type) &&
                 !ASRUtils::is_allocatable(symbol_type) &&
                 !ASRUtils::is_pointer(symbol_type)) {
                 ASR::String_t* str_t = ASRUtils::get_string_type(symbol_type);
-                if (str_t->m_len_kind == ASR::ExpressionLength && str_t->m_len) {
-                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(sym.second);
+                ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(sym.second);
+                if (str_t->m_len_kind == ASR::ExpressionLength && str_t->m_len &&
+                    str_t->m_physical_type == ASR::DescriptorString &&
+                    !var->m_value_attr) {
                     uint32_t h = get_hash((ASR::asr_t*)sym.second);
                     LCOMPILERS_ASSERT(llvm_symtab.find(h) != llvm_symtab.end());
                     llvm::Value* arg_str_desc = llvm_symtab[h];
--- a/src/libasr/codegen/asr_to_llvm.cpp
+++ b/src/libasr/codegen/asr_to_llvm.cpp
@@ -8505,15 +8505,73 @@ public:
                     allocate_array_members_of_struct(struct_sym, st_desc, ASR::down_cast<ASR::Variable_t>(sym.second)->m_type, true);
                 }
             }
-            if (ASRUtils::is_string_only(symbol_type) && ASRUtils::is_allocatable(symbol_type)) {                                                                  
-                ASR::String_t* str_t = ASRUtils::get_string_type(symbol_type);                                             
-                if (str_t->m_len_kind == ASR::ExpressionLength && str_t->m_len) {                                      
-                    uint32_t h = get_hash((ASR::asr_t*)sym.second);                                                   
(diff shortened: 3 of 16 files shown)


src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8410,15 +8410,54 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                llvm::BasicBlock::Create(context, "opt_str_set_len", fn);
+                            llvm::BasicBlock* skip_bb =
+                                llvm::BasicBlock::Create(context, "opt_str_skip_len", fn);
+                            builder->CreateCondBr(is_null, skip_bb, set_len_bb);
+                            builder->SetInsertPoint(set_len_bb);
+                            setup_string_length(str_desc, str_t, str_t->m_len);
+                            builder->CreateBr(skip_bb);
+                            builder->SetInsertPoint(skip_bb);
+                        } else {
+                            setup_string_length(str_desc, str_t, str_t->m_len);
```
Jatinagarwal24: This writes the declared length through the `caller's` descriptor. For a non-optional (or present) dummy, `str_desc` is a pointer to the caller's own variable descriptor  , the updated `llvm-string_54` reference output shows main passing %`string_descriptor* %char (its character(10) :: char)` directly, and `foo_sub` now stores `elem_len = 20` into it. After the call returns, the caller's variable claims length 20 with only 10 bytes allocated, so any later `len(char)`, assignment, or print in the caller is corrupted.

I think  instead of mutating the incoming descriptor, alloca a local descriptor copy (copy data pointer, store the declared length into the copy) and register it in llvm_symtab , the same pattern the non-allocatable DescriptorArray dummy handling uses just below. Alternatively, `lower len()` of a non-allocatable ExpressionLength dummy from the declared expression instead of reading `elem_len`, then no descriptor write is needed at all.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8410,15 +8410,54 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                                llvm::ConstantPointerNull::get(
+                                    llvm::cast<llvm::PointerType>(str_desc->getType())));
+                            llvm::Function* fn = builder->GetInsertBlock()->getParent();
+                            llvm::BasicBlock* set_len_bb =
+                                llvm::BasicBlock::Create(context, "opt_str_set_len", fn);
+                            llvm::BasicBlock* skip_bb =
+                                llvm::BasicBlock::Create(context, "opt_str_skip_len", fn);
+                            builder->CreateCondBr(is_null, skip_bb, set_len_bb);
+                            builder->SetInsertPoint(set_len_bb);
+                            setup_string_length(str_desc, str_t, str_t->m_len);
```
Jatinagarwal24: Same concern as below: the null guard only protects the absent case. When the optional argument is present, this still writes into the caller's descriptor.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8410,15 +8410,54 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+            if (ASRUtils::is_string_only(symbol_type) &&
+                !ASRUtils::is_allocatable(symbol_type) &&
+                !ASRUtils::is_pointer(symbol_type)) {
+                ASR::String_t* str_t = ASRUtils::get_string_type(symbol_type);
+                if (str_t->m_len_kind == ASR::ExpressionLength && str_t->m_len) {
+                    ASR::Variable_t* var = ASR::down_cast<ASR::Variable_t>(sym.second);
+                    uint32_t h = get_hash((ASR::asr_t*)sym.second);
+                    LCOMPILERS_ASSERT(llvm_symtab.find(h) != llvm_symtab.end());
+                    llvm::Value* str_desc = llvm_symtab[h];
+                    if (str_desc->getType()->isPointerTy()) {
```
Jatinagarwal24: On LLVM 15+ (opaque pointers) this is always true, and if it were ever false we'd silently skip the fix. Prefer LCOMPILERS_ASSERT(str_desc->getType()->isPointerTy()) or drop the check.

src/libasr/codegen/asr_to_llvm.cpp
```diff
@@ -8410,15 +8410,54 @@ class ASRToLLVMVisitor : public ASR::BaseVisitor<ASRToLLVMVisitor>
...
+                    LCOMPILERS_ASSERT(llvm_symtab.find(h) != llvm_symtab.end());
+                    llvm::Value* str_desc = llvm_symtab[h];
+                    setup_string_length(str_desc, str_t, str_t->m_len);
+                }
+            }
+            // For non-allocatable ExpressionLength string dummy arguments (including optional
+            // ones like `character(len=n), optional`), the call site may pass in a zeroed
+            // dummy descriptor (elem_len = 0) when the argument is absent. Set the length
+            // field from the declared length expression now, guarding with a null-check for
+            // optional arguments so we don't write through a null pointer.
```
Jatinagarwal24: The comment says the call site passes `"a zeroed dummy descriptor (elem_len = 0)"`  that's a non-null pointer (from `transform_optional_argument_functions`), so the null check below doesn't cover that case; NULL is only passed on other paths (e.g. bind(C)). Could you update the comment to describe both mechanisms so the guard isn't misleading?

Also note this evaluates the length expression at every function entry even when len() is never queried  the reference diffs show `_lcompilers_len_trim_str` now being called twice in `_lcompilers_trim_str`. Fine semantically (spec exprs are pure), but worth being aware of.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2374,7 +2374,17 @@ namespace LCompilers {
         switch (str_type->m_physical_type)
         {
             case ASR::DescriptorString:{
-                ptr_to_data = create_gep2(string_descriptor, str, 0);
+                if (!str->getType()->isPointerTy()) {
+                    if (get_pointer_to_data) {
+                        llvm::Value* str_ptr = CreateAlloca(str->getType(), nullptr, "str_val");
+                        builder->CreateStore(str, str_ptr);
+                        ptr_to_data = create_gep2(string_descriptor, str_ptr, 0);
```
Jatinagarwal24: When `get_pointer_to_data` is true this returns a pointer into a freshly alloca'd copy, so any store the caller performs through it is silently discarded. Since no caller can meaningfully write to a by-value descriptor, this case should throw `LCompilersException (or assert)` rather than return a dead write target.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2405,6 +2415,15 @@ namespace LCompilers {
             switch (str_type->m_physical_type)
             {
                 case ASR::DescriptorString:{
+                    if (!str->getType()->isPointerTy()) {
+                        if (get_pointer_to_len) {
+                            llvm::Value* str_ptr = CreateAlloca(str->getType(), nullptr, "str_val");
+                            builder->CreateStore(str, str_ptr);
+                            return create_gep2(string_descriptor, str_ptr, 1);
```
Jatinagarwal24: Same as in get_string_data: returning a pointer into a temp copy silently discards any store made through it (e.g. from setup_string_length / set_string_length). Better to throw/assert here.

src/libasr/codegen/llvm_utils.cpp
```diff
@@ -2374,7 +2374,17 @@ namespace LCompilers {
         switch (str_type->m_physical_type)
         {
             case ASR::DescriptorString:{
-                ptr_to_data = create_gep2(string_descriptor, str, 0);
+                if (!str->getType()->isPointerTy()) {
```
Jatinagarwal24: This branch contradicts `LCOMPILERS_ASSERT(is_proper_string_llvm_variable(str_type, str))` at the top of the function  on` LLVM < 15 ` that assert requires str to be a `string_descriptor*`, so it fires before this branch is reachable in debug builds. Where does the by-value descriptor come from? It may indicate an unintended load upstream that should be fixed at the source instead. (Same applies to the branch at line 2418 in `get_string_length`.)

integration_tests/optional_string_arg_01.f90
Jatinagarwal24, line 1: This only covers the absent path. Could you also add:

a present-argument case: call `check(1, 'a')` verifying `len(s)` and content, and
a caller-unchanged check, e.g.
```
character(10) :: a
call sub(3, a)
if (len(a) /= 10) error stop
```

integration_tests/CMakeLists.txt
Jatinagarwal24, line 1: `integration_tests/string_54.f90` has a commented-out check `if(len(char) /= 20) error stop` with a TODO saying to enable it once strings use DescriptorString , this PR makes that check pass, so it could be uncommented here as an extra regression test.

Jatinagarwal24 (changes requested):

Jatinagarwal24:
CI is failing.
To fix it i think 
Revert the `get_string_length()` hunk back to the original `constant-fast-path + descriptor-read`. It's no longer needed: your entry-block local-descriptor copy already captures the declared length at function entry, so `len()` reading `elem_len` from the descriptor returns the right value for the MRE (the absent optional arrives as a non-null stack dummy, takes the copy path, and gets `elem_len = n`). Reverting also avoids re-evaluating the spec expression at every `len()` use, which would be wrong if a variable in it is redefined after entry, and removes the double _lcompilers_len_trim_str call visible in the reference diffs. (If you want to keep it, you'd need the extra CreateLoad2 fix-up that setup_string_length has for allocatable/pointer StructInstanceMember/ArrayItem/ArraySection length exprs, best factored into a shared helper.)

Also update the test references.

Jatinagarwal24 (approved):
Now it looks good to me.
Thanks .

Issue #12453, opened 2026-08-07
ASR fuzzer: enforce verifier-or-executable contract


certik:
## Goal

Build an ASR fuzzer that generates millions of bounded random ASR graphs and compiles each one through the default ASR pass pipeline and default LLVM backend.

For every generated input, exactly two outcomes are valid:

1. Initial ASR verification rejects the input with an `ASRVerify` diagnostic.
2. Initial ASR verification succeeds and the compiler emits an executable binary.

Anything else is a compiler bug: a verifier crash, assertion failure, exception, unimplemented path, timeout, post-pass verifier failure, LLVM failure, object-emission failure, or linker failure.

This implements the randomized ASR-hardening direction tracked in #1303.

## Contract

The authoritative pipeline is:

```text
construct or decode schema-valid ASR
    |
    +-- run initial asr_verify
    |       +-- failure -> accepted verifier rejection
    |       +-- success -> continue
    |
    +-- run the default ASR passes
    |       +-- verify after every pass
    |       +-- any verifier failure -> bug
    |
    +-- LLVM lowering
    +-- LLVM module verification
    +-- object emission
    +-- production linker
            +-- executable emitted -> accepted
            +-- anything else -> bug
```

Only the initial verifier rejection is accepted. Once the initial ASR passes verification, every subsequent phase must succeed and produce a binary.

ASR verification should optionally enforce a standalone/executable profile: exactly one main program and no unresolved or separately compiled dependencies. The ordinary verifier must remain usable for modules and separately compiled ASR, while the fuzzer enables the stricter complete-program checks.

Post-pass verification must be enabled by the fuzzer independently of `WITH_LFORTRAN_ASSERT`.

## Shared post-ASR compilation API

Extract one library-level post-ASR entry point, for example `compile_asr_to_executable()`, and use it from:

- the normal `lfortran` CLI;
- the direct-ASR fixture runner; and
- the fuzzer.

It must own default backend/pass configuration, LLVM lowering, object emission, runtime/linker configuration, and executable linking. This avoids copying logic currently private to the anonymous namespace in `src/bin/lfortran.cpp` and prevents the fuzzer from drifting from normal compiler behavior.

Failures should be structured by phase. Do not catch a generic `LCompilersException` and reinterpret it as verifier rejection.

## Lossless ASR text format

Add a human-readable, round-trippable ASR text input format for minimized regression fixtures. The current positional Clojure/pickle output is write-only and lossy, so it should not be parsed as-is.

[...]

Issue #1303, opened 2023-02-14
ASR refactoring


certik:
There are several long standing issues with ASR that need refactoring. We will track of all of them here:

* [ ] https://github.com/lfortran/lfortran/issues/416
* [ ] FunctionType hardening (implemented in https://github.com/lfortran/lfortran/pull/1189)
* [ ] function argument index
* [ ] runtime library: right now we use BindC and implement all interfacing with the system using it; we need to rather introduce specific ASR nodes for interfacing the system (such as clock, file IO, Intrinsic math functions, etc.), the runtime library will eventually be just pure Fortran (not calls into C), possibly using some "LFortran builtin" functions to access specific functionality, and we can emulate them for other compilers so that we can compile the runtime library with other compilers as well
* [ ] Closures (parent scope variables)
* [ ] Improve printing of ASR to be exactly equal to Clojure

PR #12459, opened 2026-08-08
feat: add a lossless ASR text format and an ASR fuzzing harness

certik:
## Summary

Adds a lossless, round-trippable ASR text format and builds an ASR fuzzing
harness on top of it, implementing the direction tracked in #12453 (and the
randomized ASR hardening in #1303).

The format is a named-field Clojure/EDN dialect with encoder and decoder
generated from `ASR.asdl`. Unlike the previous positional Clojure output, it
is parseable, so a minimized ASR graph can be checked in as a regression
fixture and compiled directly:

    lfortran mre.asr -o mre        # .asr inputs are detected by extension
    lfortran --verify-asr mre.asr  # parse and verify, no passes

On top of that, the fuzzer generates bounded ASR graphs -- both by mutating
integration-test ASR and by constructing graphs from the schema -- and runs
each through the full pipeline. Exactly two outcomes are accepted: the initial
verifier rejects the input, or the compiler emits a linked executable.
Anything else is recorded as a bug with a replayable artifact.

## Scope

- Lossless ASR text format: generic S-expression parser, generated encoder and
  decoder, `(ASRText :version 1 ...)` envelope and `(SymbolRef 1 "x")` symbol
  references, strict errors for unknown or missing fields.
- Direct `.asr` compilation, `--from-asr`, `--verify-asr`, `--no-member-names`,
  `--verify-all-passes`.
- A shared post-ASR pipeline (`load_input_asr`, `compile_asr_to_object`) used
  by the CLI, the fixture runner and the fuzzer, so the fuzzer cannot drift
  from normal compiler behavior. Failures are reported by phase rather than by
  reinterpreting a generic exception.
- An optional standalone/executable verifier profile (`require_main_program`);
  the ordinary verifier stays usable for modules and separate compilation.
- Post-pass verification that the fuzzer can enable independently of
  `WITH_LFORTRAN_ASSERT`.
- Fuzzing harness: deterministic mutation and schema generators, isolated
  workers, timeouts, persisted artifacts, replay, structural reduction and
  coverage metrics.
- A compile/verify regression corpus with stable diagnostic codes and spans,
  an LLVM constructor coverage check, and a scheduled ASan/UBSan campaign.
- Six verifier gaps found by the harness and closed: call argument types and
  kinds, assignment value types, function signature types, and unsupported
  intrinsic kinds.

56 files changed, ~5.7k insertions.

## Verification

Built Debug with LLVM, assertions and runtime stacktraces.

| Suite | Result |
| --- | --- |
| CTest (unit, ASR corpus, fuzzer smoke, schema generator, reducer, LLVM constructor coverage) | 8/8 pass |
| Reference suite `./run_tests.py` | pass |
| Integration suite, LLVM backend | 3851/3851 pass |

Campaigns:

| Campaign | Cases | Compiled | Verifier rejections | Bugs |
| --- | --- | --- | --- | --- |
| 6 seeds x 500, Debug + assertions | 3000 | 1822 | 1178 | 0 |
| ASan + UBSan, seed 11 | 300 | 167 | 133 | 0 |

Each seed covered 36 ASR constructors, 24 enums, all 29 default passes and both
outcome phases.

The new verifier checks are covered by tests that fail before the change:
the `ASR Verify` unit test, the `derived_types_06` and
`errors/intent_inout_missing_arg` reference tests, and the
`procedure_pointer_12`, `procedure_pointer_18`, `separate_compilation_31` and
`real128_intrinsic_inquiry_01` integration tests.

## Rationale

The existing Clojure output is write-only, so minimized ASR bugs could not be
checked in as fixtures -- every ASR-level regression had to be re-derived from
Fortran source. A parseable format makes an ASR graph a first-class test input,
which in turn makes randomized ASR generation useful: the fuzzer can hand the
compiler a graph the frontend would never produce and hold it to a contract the
verifier is supposed to enforce.

Sharing one post-ASR entry point is what keeps that honest. If the fuzzer built
its own pipeline it would drift from the compiler, and every finding would need
to be re-confirmed by hand.

## User-visible changes

- `--clojure` now emits the named-field format. Three reference outputs change
  accordingly. Anything parsing the old positional output will break.
- In builds with assertions, a post-pass verifier failure now exits 3 with an
  `ASR_FUZZ_FAILURE phase=pass pass=<name>` line instead of a generic ICE.


certik:
Everything from here was implemented and merged in separate smaller PRs.

PR #12460, opened 2026-08-08
Reject invalid assumed codimensions in coarray declaration

### LINKED ISSUE #12457: Invalid assumed codimensions are being compiled in coarray declaration
## MRE 1
```fortran
program test_corank
    implicit none
    integer :: b[*, *]

end program test_corank
```

LFortran compiles this code and gives runtime segmentation fault, whereas this program should be rejected with semantic error.

## MRE 2
```fortran
program test_corank
    implicit none
    integer :: b[*, 2]

end program test_corank
```

## LFortran's Output:
```
AssertFailed: var->m_codims[ci].m_end == nullptr
```

This program should also be rejected at semantic stage itself before the coarray pass executes

usefahmed07:
Fixes #12457

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -7642,7 +7642,28 @@ public:
                                     AST::codimension_typeType::CodimensionStar
                                     ? ASR::codimension_typeType::CodimensionStar
                                     : ASR::codimension_typeType::CodimensionExpr;
+                                if (asr_codim.m_end_star ==
+                                        ASR::codimension_typeType::CodimensionStar &&
+                                        c + 1 != ac->n_codim) {
+                                    diag.add(Diagnostic(
+                                        "The upper cobound `*` must appear "
+                                        "only in the last codimension",
+                                        Level::Error, Stage::Semantic, {
+                                            Label("", {ast_codim.loc})
+                                        }));
+                                    throw SemanticAbort();
+                                }
                                 codims.push_back(al, asr_codim);
+                            }
+                            if (ac->n_codim == 0 ||
+                                    codims[codims.size() - 1].m_end_star !=
+                                        ASR::codimension_typeType::CodimensionStar) {
+                                diag.add(Diagnostic(
+                                    "The last upper cobound of a coarray must be `*`",
+                                    Level::Error, Stage::Semantic, {
+                                        Label("", {x.base.base.loc})
+                                    }));
+                                throw SemanticAbort();
                             }
                         } else {
                             diag.add(Diagnostic(
@@ -7768,6 +7789,17 @@ public:
                             AST::codimension_typeType::CodimensionStar
                             ? ASR::codimension_typeType::CodimensionStar
                             : ASR::codimension_typeType::CodimensionExpr;
+                        if (asr_codim.m_end_star ==
+                                ASR::codimension_typeType::CodimensionStar &&
+                                c + 1 != s.n_codim) {
+                            diag.add(Diagnostic(
+                                "The upper cobound `*` must appear "
+                                "only in the last codimension",
+                                Level::Error, Stage::Semantic, {
+                                    Label("", {ast_codim.loc})
+                                }));
+                            throw SemanticAbort();
+                        }
                         codims.push_back(al, asr_codim);
                     }
                 }
--- a/tests/errors/continue_compilation_coarrays.f90
+++ b/tests/errors/continue_compilation_coarrays.f90
@@ -25,4 +25,13 @@ program continue_compilation_coarrays
     real :: cod[4]
     print *,cod[5]
 
+    ! Assumed codimension `*` must be the last codimension → ERROR
+    integer :: w1[*, *]
+
+    ! Assumed codimension `*` not in the last position → ERROR
+    integer :: w2[*, 2]
+
+    ! CODIMENSION attribute: last cobound must be `*` → ERROR
+    integer, codimension[3, 2] :: w3
+
 end program continue_compilation_coarrays
--- a/tests/reference/asr-continue_compilation_coarrays-85ccad8.json
+++ b/tests/reference/asr-continue_compilation_coarrays-85ccad8.json
@@ -2,12 +2,12 @@
     "basename": "asr-continue_compilation_coarrays-85ccad8",
     "cmd": "lfortran --semantics-only --continue-compilation --no-color {infile}",
     "infile": "tests/errors/continue_compilation_coarrays.f90",
-    "infile_hash": "519fc54a139fe129a0865f6d7c1566618583a68add260fe02c6d3456",
+    "infile_hash": "5b7a5590e476849244928cf6528a575da3717461dd33b5029b8e2915",
     "outfile": null,
     "outfile_hash": null,
     "stdout": null,
     "stdout_hash": null,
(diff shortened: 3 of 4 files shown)


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7642,8 +7642,30 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                                     AST::codimension_typeType::CodimensionStar
                                     ? ASR::codimension_typeType::CodimensionStar
                                     : ASR::codimension_typeType::CodimensionExpr;
+                                if (asr_codim.m_end_star ==
+                                        ASR::codimension_typeType::CodimensionStar &&
+                                        c + 1 != ac->n_codim) {
+                                    diag.add(Diagnostic(
+                                        "The assumed codimension `*` must appear "
+                                        "only in the last codimension",
```
bonachea: Minor wording quibble: "assumed codimension" is not a term in the Fortran standard, and we should not invent a term in an error message.

AFAIK there is no standard term to specifically denote the star, the star character is simply a required part of the syntax in  *explicit-coshape-spec* and *allocate-coarray-spec*. Also it's not a codimension, it's a placeholder for only an upper cobound (one of the two bounds defining a codimension).

I suggest we instead say something like
```suggestion
                                        "The upper cobound `*` must appear "
                                        "only in the last codimension",
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7642,8 +7642,30 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                    throw SemanticAbort();
+                                }
                                 codims.push_back(al, asr_codim);
                             }
+                            if (ac->n_codim == 0 ||
+                                    codims[codims.size() - 1].m_end_star !=
+                                        ASR::codimension_typeType::CodimensionStar) {
+                                diag.add(Diagnostic(
+                                    "The last codimension of a coarray must be "
+                                    "an assumed codimension `*`",
```
bonachea: rephrasing
```suggestion
                                    "The last upper cobound of a coarray must be `*`",
```

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7768,6 +7790,17 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                             AST::codimension_typeType::CodimensionStar
                             ? ASR::codimension_typeType::CodimensionStar
                             : ASR::codimension_typeType::CodimensionExpr;
+                        if (asr_codim.m_end_star ==
+                                ASR::codimension_typeType::CodimensionStar &&
+                                c + 1 != s.n_codim) {
+                            diag.add(Diagnostic(
+                                "The assumed codimension `*` must appear "
+                                "only in the last codimension",
```
bonachea:
```suggestion
                                "The upper cobound `*` must appear "
                                "only in the last codimension",
```

tests/errors/continue_compilation_coarrays.f90
bonachea, line 1: Are the analogous errors already covered for _allocate-coarray-spec_ in ALLOCATE? Eg:
```fortran
integer :: A[:,:]
ALLOCATE(A[*,*])
ALLOCATE(A[*,2])
```

src/lfortran/semantics/ast_common_visitor.h
bonachea: Please expand the test to generate this new error message

bonachea (changes requested):
Thanks for working on this!

I have a few minor requests..

usefahmed07:
@bonachea Thanks For Review it i will address all Comments

tests/errors/continue_compilation_coarrays.f90
usefahmed07: > Are the analogous errors already covered for _allocate-coarray-spec_ in ALLOCATE? Eg:
> 
> ```fortran
> integer :: A[:,:]
> ALLOCATE(A[*,*])
> ALLOCATE(A[*,2])
> ```

Good catch that's not currently covered. `A[*,*]` and `A[*,2]` in an ALLOCATE coarray-spec go through a different code path (`visit_CoarrayRef`) that doesn't have the same validation. Can you open a follow-up issue to track this separately, to keep this PR focused on the declaration case.

or Solve it in this but i suggested open issue to keep PR Clean

src/lfortran/semantics/ast_common_visitor.h
usefahmed07: You're right, thanks for the clarification  "assumed codimension" isn't a standard term. Updated to "upper cobound" as suggested.

usefahmed07:
Hi @bonachea I've addressed all the review comments (wording changes + expanded test for the CODIMENSION attribute case). Could you take another look?

tests/errors/continue_compilation_coarrays.f90
bonachea: New issue : #12463

bonachea (commented):
> Hi @bonachea I've addressed all the review comments (wording changes + expanded test for the CODIMENSION attribute case). Could you take another look?

Thanks for the fixes.

In the future, when updating a PR please push new commits that address reviewer comments. Rewriting the existing commit makes it harder for reviewers to audit what changed, especially if you also rebase it onto the target branch during review.

usefahmed07:
> > Hi @bonachea I've addressed all the review comments (wording changes + expanded test for the CODIMENSION attribute case). Could you take another look?
> 
> Thanks for the fixes.
> 
> In the future, when updating a PR please push new commits that address reviewer comments. Rewriting the existing commit makes it harder for reviewers to audit what changed, especially if you also rebase it onto the target branch during review.

Okay, I'm noting these observations, I'll improve the quality of PRs in the future. Thanks for the heads-up.

bonachea (approved):
Changes LGTM!

tests/errors/continue_compilation_coarrays.f90
```diff
@@ -25,4 +25,13 @@ program continue_compilation_coarrays
     real :: cod[4]
     print *,cod[5]
```
bonachea, line 26: In looking this over again I just noticed something weird here. 

This explicit coshape coarray declaration for `cod` is also invalid, and _should_ be triggering the same error added in this PR (because the final ucobound is not `*`).  However we appear to accept this declaration and only error on the subsequent line in the coindexed access.

@usefahmed07 will you please investigate why that's not happening, and fix the error check to also catch the case of a single codimension where the ucobound is not `*`? We should ensure this test also covers that case (i.e. please don't delete the erroneous declaration, I suggest we expand it as shown below and adjust the expected error output). corank == 1 is the overwhelmingly common case in practice, so it's even more important to get that case right.

Perhaps @Jainam-not-a-robot can help explain why this isn't working as expected.


```suggestion
    ! last ucobound must be '*' -> ERROR
    real :: cod[4]

    ! last ucobound must be '*' -> ERROR
    real :: cod2[2:4]
    
    ! last ucobound must be '*' -> ERROR
    real, codimension[5] :: cod3
    
    ! last ucobound must be '*' -> ERROR
    real, codimension[2:5] :: cod4
```

bonachea (changes requested):
Sorry, I just noticed a puzzling problem here..

bonachea:
Superseded by #12466

PR #12462, opened 2026-08-09
Implement new ASR reader and writer

certik:
## Summary

Adds a lossless, round-trippable text format for ASR, so an ASR graph can be
written out, read back, and compiled. This is the foundation for checking in
minimized ASR regression fixtures and for the ASR fuzzing work tracked in
#12453.

The existing `--clojure` output is positional and lossy, so it can be read by a
human but not by the compiler. Every ASR-level bug therefore had to be
re-derived from Fortran source, and there was no way to hand the compiler a
graph the frontend would never produce. This PR replaces that output with a
named-field Clojure/EDN dialect that parses back into identical ASR:

```clojure
(ASRText
  :version 1
  :value (TranslationUnit
    :symtab (SymbolTable
      :id 0
      :symbols {
        "asr_text_compile_01" (Program
          :symtab (SymbolTable :id 1 :symbols {})
          :name "asr_text_compile_01"
          :dependencies []
          :body []
        )
      }
    )
    :items []
  )
)
```

An `.asr` file can then be compiled directly:

```console
lfortran mre.asr -o mre
lfortran --show-asr --clojure prog.f90 > prog.asr   # round-trips
```

## Scope

- Generic S-expression parser (`src/libasr/asr_text_parser.{h,cpp}`) with
  precise diagnostics and a bounded nesting depth.
- Printer (`src/libasr/asr_text.{h,cpp}`) plus encoder and decoder generated
  from `ASR.asdl` by `asdl_cpp.py`, so the format tracks the grammar and cannot
  silently drift from it.
- Named fields with strict errors for unknown or missing required fields, an
  explicit `(ASRText :version 1 ...)` envelope, `(SymbolRef <symtab-id> "name")`
  symbol references, and lossless constants and intrinsic modules.
- CLI: `--from-asr` (also inferred from an `.asr` extension), and
  `--no-member-names` to get the previous positional form. `--clojure` now
  emits the lossless named format.
- `ASRVerifyOptions`, so a caller can ask for the standalone/executable profile
  instead of the default one. The ordinary verifier stays usable for modules
  and separately compiled ASR.

## Verification

- New unit tests: `test_asr_text_parser.cpp` covers the parser (literals,
  escapes, UTF-8, embedded NUL, maps, comments, malformed input, nesting
  limits); `test_asr_text.cpp` covers printing, round-tripping, schema errors
  and location overrides.
- New CTest fixtures compile `tests/asr/asr_text_compile_01.asr` to an object
  and to an executable, so the format is exercised end to end.
- Eight reference outputs are updated for the new `--clojure` form, and the
  fixture is registered in `tests/tests.toml`.
- Full local run on macOS: unit tests, reference suite, and the LLVM
  integration suite (3851 tests) all pass. CTest was also run from a source
  tarball build, matching what CI does.

## Notes for review

- `--clojure` output changes shape. Anything parsing the old positional form
  will need updating; `--no-member-names` restores it.
- The last two commits are CI fixes rather than new behaviour: GCC diagnoses
  two values that Clang does not, the wasm test binary needs a larger stack
  than the 64kb Emscripten default for the parser's maximum nesting depth, and
  the source tarball has to ship `tests/asr` because CI runs CTest from it. The
  Windows job has no working end-to-end link, so only object emission is
  registered there.

This is the second PR in a stack; #12461 is merged. The next PRs add the
regression corpus, the shared post-ASR pipeline, the verifier type checks, and
the fuzzing harness.


doc/src/asr/asr_text.md
```diff
@@ -0,0 +1,142 @@
...
+It prints every non-location field using the exact field names in
+`src/libasr/ASR.asdl`:
+
+```clojure
+(ASRText
+  :version 1
+  :value (Var
+    :v (SymbolRef 1 "x")
+  )
+)
```
certik, line 32: I don't like this `ASRText` wrapper, I don't know if this is needed at all.

We also need to check if the output actually loads into clojure, or if that compatibility will not work anyway. If it doesn't work, it would be good to "define" our LISP/clojure-like output somehow, so that we know what the "correct" syntax is.

certik: This is now fixed.

src/libasr/asdl_cpp.py
```diff
@@ -2306,6 +2306,590 @@ def visitField(self, field, cons):
...
+            helper = "decode_symbol_table_ref" \
+                if field.name == "parent_symtab" else "decode_symbol_table"
+            self.emit("if (!self().%s(%s, %s)) return false;"
+                % (helper, value, target), indent)
+        elif field.type == "symbol":
+            self.emit("if (!self().decode_symbol_ref(%s, %s)) return false;"
+                % (value, target), indent)
+        elif field.type == "void":
+            self.emit("if (!self().decode_bytes(%s, %s, "
+                "static_cast<size_t>(m_n_data))) return false;"
```
certik: **[bug]** `ArrayConstant` decodes `n_data` as an unconstrained `int64` and immediately does `al.alloc(static_cast<size_t>(m_n_data))` in `decode_bytes`. A hostile `.asr` can use a huge or negative `n_data` (the latter wraps to a near-`SIZE_MAX` allocation) with a short `#asr/bytes` string; the EDN parser’s node/depth caps do not apply here. The format is explicitly an untrusted-input path (`lfortran file.asr`).

**Suggestion:** Reject `n_data < 0` and cap the payload (e.g. against remaining input size or a documented maximum) before `alloc`, and emit an `ASRParser` diagnostic. Apply the same check to any other integer-derived allocation.

src/libasr/asr_text_parser.cpp
```diff
@@ -0,0 +1,622 @@
...
+                            span_loc(esc_start, esc_start), "escape starts here");
+                }
+                char e = src[pos];
+                switch (e) {
+                    case '\\': buffer += '\\'; pos++; break;
+                    case '"':  buffer += '"';  pos++; break;
+                    case 'n':  buffer += '\n'; pos++; break;
+                    case 'r':  buffer += '\r'; pos++; break;
+                    case 't':  buffer += '\t'; pos++; break;
+                    case 'b':  buffer += '\b'; pos++; break;
```
certik: **[suggestion]** The published grammar and `asr_text.md` accept only `\t`, `\r`, `\n`, `\"`, `\\`, and `\uNNNN`, and they reject `\b`/`\f` as non-EDN. The reader still accepts `\b` and `\f`, and `\u` of surrogate code points is encoded as ill-formed UTF-8 even though the format is defined as UTF-8 only. Printed output stays conservative, but hand-written documents that the docs say must be rejected will parse.

**Suggestion:** Reject `\b`/`\f` and surrogate/`\u` values outside `[0, 0xD7FF] ∪ [0xE000, 0xFFFF]`, and reject raw string bytes that are not well-formed UTF-8.

src/lfortran/tests/CMakeLists.txt
```diff
@@ -89,7 +93,43 @@ if (XEUS_LFORTRAN_WASM_BUILD)
...
+        NAME test_asr_text_object
+        COMMAND $<TARGET_FILE:lfortran>
+            ${CMAKE_SOURCE_DIR}/tests/asr/asr_text_compile_01.asr
+            --no-color
+            -c
+            -o ${PROJECT_BINARY_DIR}/asr_text_compile_01.o
+    )
+    # The format claims to be an EDN data subset; enforce that with a reader
+    # that knows nothing about LFortran.
+    find_program(LFORTRAN_TEST_PYTHON NAMES python python3 REQUIRED)
```
certik: **[suggestion]** `find_program(LFORTRAN_TEST_PYTHON NAMES python python3 REQUIRED)` makes configuring `src/lfortran/tests` fail when Python is absent, even though `check_edn.py` already skips without `edn_format` outside CI. Python is common for this repo, but this is a new hard configure-time requirement for a subset of optional tests.

**Suggestion:** Drop `REQUIRED` and only `add_test(test_asr_text_edn_conformance ...)` when the interpreter is found.

src/libasr/asr_text_parser.h
```diff
@@ -0,0 +1,127 @@
...
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <libasr/diagnostics.h>
+#include <libasr/exception.h>
+#include <libasr/location.h>
+
+// ASRText is a generic, non-evaluating reader for the EDN subset used by the
```
certik: **[suggestion]** The file- and type-level comments narrate architecture and ownership policy at length (what the layer is not, why maps are not hash maps, preferred overloads). That belongs in `doc/src/asr/asr_text.md`; headers should keep a short contract.

**Suggestion:** Collapse to a few sentences on ownership, location spans, and the two `parse` overloads.

PR #12466, opened 2026-08-09
Enforce C827/C828: deferred vs explicit coarray coshape

### LINKED ISSUE #12457: Invalid assumed codimensions are being compiled in coarray declaration

### LINKED ISSUE #12465: Missing checks for deferred coshape vs allocatable coarrays (C827, C828)
# MRE

```fortran
! non-allocatable coarray may not have deferred coshape
integer :: A[:]
integer :: B[:,*]
integer :: C[2,:,*]

! allocatable coarray may not have explicit shape
integer, allocatable :: W[*]
integer, allocatable :: X[:,*]
integer, allocatable :: Y[:,4,:]
integer, codimension[2,10,*], allocatable :: Z

end
```

# Problem

Every coarray declaration above is invalid. They violate F2023 constraints C827 and C828:

Image

These are currently all compiled without error, but each should generate an error.

CC: @usefahmed07 , @Jainam-not-a-robot

usefahmed07:
Fixes #12465

### DIFF
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -7771,6 +7771,70 @@ public:
                         codims.push_back(al, asr_codim);
                     }
                 }
+                if (corank > 0) {
+                    // C827: A coarray with the ALLOCATABLE attribute shall have
+                    // a coarray-spec that is a deferred-coshape-spec-list (i.e.
+                    // every codimension is a bare ':', no explicit bounds or '*').
+                    // C828: A nonallocatable coarray shall have a coarray-spec
+                    // that is an explicit-coshape-spec (every codimension but the
+                    // last has an explicit upper cobound, and the last is '*').
+                    for (size_t c = 0; c < codims.size(); c++) {
+                        const ASR::codimension_t &codim = codims[c];
+                        bool is_deferred = (codim.m_end == nullptr &&
+                            codim.m_start == nullptr &&
+                            codim.m_end_star != ASR::codimension_typeType::CodimensionStar);
+                        if (is_allocatable) {
+                            if (!is_deferred) {
+                                diag.add(Diagnostic(
+                                    "A coarray with the `allocatable` attribute "
+                                    "must have a deferred coshape (every "
+                                    "codimension written as `:`)",
+                                    Level::Error, Stage::Semantic, {
+                                        Label("", {codim.loc})
+                                    }));
+                                throw SemanticAbort();
+                            }
+                        } else {
+                            if (is_deferred) {
+                                diag.add(Diagnostic(
+                                    "A nonallocatable coarray must have an "
+                                    "explicit coshape; a deferred codimension "
+                                    "(`:`) is only permitted for allocatable "
+                                    "coarrays",
+                                    Level::Error, Stage::Semantic, {
+                                        Label("", {codim.loc})
+                                    }));
+                                throw SemanticAbort();
+                            } else if (c == codims.size() - 1) {
+                                if (codim.m_end_star != ASR::codimension_typeType::CodimensionStar) {
+                                    diag.add(Diagnostic(
+                                        "The last upper cobound of a coarray "
+                                        "must be `*`",
+                                        Level::Error, Stage::Semantic, {
+                                            Label("", {codim.loc})
+                                        }));
+                                    throw SemanticAbort();
+                                }
+                            } else if (codim.m_end_star == ASR::codimension_typeType::CodimensionStar) {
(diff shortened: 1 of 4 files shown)


usefahmed07:
@bonachea Okay, I've solved the problem I was looking for. The issue was that the compiler wasn't checking if the coarray type matched its syntax meaning you could write an allocatable and use explicit numbers like [*], or write a regular coarray and use a colon (which should only be used for allocatables). Now I've added a check that verifies this after it finishes reading all the attributes, so it works correctly regardless of their order and covers both syntaxes (attribute and bracket syntax).

I tried all the examples I included in the issue, and they're all now giving me a clear error instead of passing normally.

Regarding the size of the diff, it's slightly larger than usual because it contains three things: the check itself, the new tests (7 cases), and I also found an existing test (code[4]) that was itself incorrect according to the same rule and was only working by chance because of this bagged file so I fixed it too.

There's no regression; I ran all the tests in this file, and they all passed.

Do you have any other suggestions?

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -7771,6 +7771,70 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+                                throw SemanticAbort();
+                            } else if (codim.m_end == nullptr) {
+                                diag.add(Diagnostic(
+                                    "A nonallocatable coarray's codimension "
+                                    "must have an explicit upper cobound",
+                                    Level::Error, Stage::Semantic, {
+                                        Label("", {codim.loc})
+                                    }));
+                                throw SemanticAbort();
+                            }
```
bonachea, line 7834: These checks for `CodimensionStar` seem to duplicate the checks for the same errors proposed in #12460 
We probably shouldn't check the same error in two places, as one of them would presumably become unreachable.

Are you intending this PR to supersede #12460 which we'd then close without merging?
If so, then please close that PR and copy over the new test cases to this PR so we preserve coverage of the `CodimensionStar` error messages.

bonachea (changes requested):
This PR appears to partially duplicate #12460 and I'd like clarity on how that will be resolved.

usefahmed07:
@bonachea You're right, thanks for catching that. Since this PR's check already covers the same errors, I copied over the w1/w2/w3 test cases from #12460 so we don't lose that coverage. I think it's safe to close #12460 without merging in favor of this one let me know if that works for you.

tests/errors/continue_compilation_coarrays.f90
```diff
@@ -33,4 +33,24 @@ program continue_compilation_coarrays
...
 
+    ! Assumed codimension `*` must be the last codimension → ERROR
+    integer :: w1[*, *]
+
+    ! Assumed codimension `*` not in the last position → ERROR
+    integer :: w2[*, 2]
+
+    ! CODIMENSION attribute: last cobound must be `*` → ERROR
+    integer, codimension[3, 2] :: w3
+
```
bonachea, line 44: Let's add some more checks to capture the unresolved thread from #12460 for corank==1 that was not being correctly diagnosed

```suggestion

    ! last ucobound must be '*' -> ERROR
    real :: w4[4]
    
    ! last ucobound must be '*' -> ERROR
    real :: w5[2:4]
    
    ! last ucobound must be '*' -> ERROR
    real, codimension[5] :: w6
    
    ! last ucobound must be '*' -> ERROR
    real, codimension[2:5] :: w7

```

bonachea (changes requested):
Changes LGTM, but I'd like to expand the test cases to directly cover the common case of corank==1 explicit-shape, which was not being correctly diagnosed in #12460 for unexplained reasons.

tests/errors/continue_compilation_coarrays.f90
usefahmed07: on it

bonachea (approved):
LGTM - thanks for the improvements!

usefahmed07:
@bonachea Done; added w4-w7 covering the corank==1 explicit-shape case. All four now give a clear "last upper cobound must be `*`" error across both syntaxes (bracket and codimension attribute), with and without an explicit lower bound. Verified locally, no regressions.

PR #12467, opened 2026-08-10
feat: support character results in the Fortran evaluator

anutosh491:
This PR lays the foundation for the "Pull" mechanism we were discussing on Zulip.

Pull-style rich display depends on the kernel being able to receive a value from the final expression in a cell. Character expressions were not previously supported by the evaluator, so even a simple final value such as 'hello' could not be returned.

This is a small prerequisite for implementing rich MIME results later.

Major changes being 
- Support character expressions in the global-statements wrapper.
- Return character data from the LLVM evaluator using LFortran's string descriptor.

### DIFF
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -1,3 +1,5 @@
+#include <array>
+#include <cstring>
 #include <fstream>
 
 #include <lfortran/fortran_evaluator.h>
@@ -37,6 +39,27 @@ namespace LCompilers {
 #endif
 
 namespace LCompilers {
+
+class StringDescriptor {
+    std::array<unsigned char, sizeof(char *) + sizeof(int64_t)> storage{};
+
+public:
+    void *pointer() {
+        return storage.data();
+    }
+
+    char *data() const {
+        char *data;
+        std::memcpy(&data, storage.data(), sizeof(data));
+        return data;
+    }
+
+    int64_t length() const {
+        int64_t length;
+        std::memcpy(&length, storage.data() + sizeof(char *), sizeof(length));
+        return length;
+    }
+};
 
 
 /* ------------------------------------------------------------------------- */
@@ -131,6 +154,11 @@ Result<FortranEvaluator::EvalResult> FortranEvaluator::evaluate(
         result.asr = pickle(*asr, true, false, false, false);
     }
 
+    bool character_result = asr->n_items > 0
+        && asr->m_items[asr->n_items - 1]->type == ASR::asrType::expr
+        && ASRUtils::is_character(*ASRUtils::expr_type(
+            ASRUtils::EXPR(asr->m_items[asr->n_items - 1])));
+
     // ASR -> LLVM
     Result<std::unique_ptr<LLVMModule>> res3 = get_llvm3(*asr,
         pass_manager, diagnostics, lm, lm.files.back().in_filename,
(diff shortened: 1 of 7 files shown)


anutosh491:
This should be fixed by the time the PR is merged
image

certik (approved):
I think that this is fine.

certik:
Let's run all tests just to be sure.

certik:
@anutosh491 the tests failed, so they'll have to get fixed first.

certik:
The coarray failure might be unrelated (https://github.com/lfortran/lfortran/issues/12473), but the evaluator failure for strings is likely related.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -38,6 +40,27 @@ namespace LCompilers {
...
+        std::memcpy(&data, storage.data(), sizeof(data));
+        return data;
+    }
+
+    int64_t length() const {
+        int64_t length;
+        std::memcpy(&length, storage.data() + sizeof(char *), sizeof(length));
+        return length;
+    }
+};
```
anutosh491, line 62: I was short on time to debug the failure so I asked ChatGPT about it and it educated me that the fix lied in the StringDescriptor. It says this 

### WASM descriptor layout

LFortran represents deferred-length character values using a packed descriptor containing a pointer followed by a 64-bit length.

The original evaluator used an equivalent-looking C++ struct:

```cpp
struct StringDescriptor {
    char *data;
    int64_t length;
};
```
This happened to work on native 64-bit platforms because the pointer is 8 bytes, so the length naturally starts at byte offset 8.
On wasm32, however, a pointer is only 4 bytes. LFortran's packed LLVM descriptor places the length immediately after it, at byte offset 4. The C++ ABI may add padding and place the 64-bit length at byte offset 8 instead.
As a result, the WASM evaluator read the length from the wrong memory location. It could retrieve the data pointer, but interpreted the string length incorrectly, causing the returned character result to be empty.
The evaluator now treats the descriptor as its packed ABI representation—sizeof(char*) + sizeof(int64_t) bytes—and extracts each field with memcpy. This avoids relying on target-specific C++ structure padding and works for both native 64-bit and wasm32 builds.

In very simple terms: the two sides agreed about what information was being passed, but disagreed about where the string length lived in memory.

```text
Native 64-bit:

pointer: bytes 0–7
length:  bytes 8–15

WASM packed:

pointer: bytes 0–3
length:  bytes 4–11

WASM C++ struct with padding:

pointer: bytes 0–3
padding: bytes 4–7
length:  bytes 8–15
```
Native worked accidentally because its natural C++ layout matched the packed LLVM layout. WASM exposed the portability bug.

anutosh491: I think this should do the trick and hopefully all tests pass now.

src/libasr/pass/global_stmts.cpp
```diff
@@ -107,6 +107,25 @@ void pass_wrap_global_stmts(Allocator &al,
                 fn_scope->add_symbol(std::string(var_name), down_cast<ASR::symbol_t>(return_var));
                 target = return_var_ref;
                 idx++;
+            } else if (ASRUtils::is_character(*ASRUtils::expr_type(value))) {
```
certik, line 110: Is there a reason not to follow the previous style:

```suggestion
            } else if (ASRUtils::expr_type(value)->type == ASR::ttypeType::String) {
```

It seems it should work? We could probably turn this if chain into a switch statement later.

certik (approved):
I think that this is fine to merge, we can iterate on this later.

src/lfortran/fortran_evaluator.cpp
certik: Fine for now.

PR #12472, opened 2026-08-10
Fix: reject IAND/IOR/IEOR calls with both-BOZ arguments (#12471)

### LINKED ISSUE #12471: IAND/IOR/IEOR accept two BOZ literal arguments without a semantic error
Per the Fortran standard and as documented by gfortran, IAND, IOR, and IEOR
each require that I and J not both be BOZ literal constants — if only one
argument is BOZ, it is converted to the kind of the other; if both are BOZ,
there is no "other" kind to convert to, so the standard forbids it
(see e.g. https://gcc.gnu.org/onlinedocs/gfortran/IOR.html: "I and J shall
not both be boz-literal-constants").

LFortran already rejects this correctly for MERGE_BITS, DSHIFTL, and DSHIFTR
(#12416 / #12442), but IAND, IOR, and IEOR are missing the same check.

## Reproducer

    program bug
        print *, iand(o'12345', z'1234')
        print *, ior(o'12345', z'1234')
        print *, ieor(o'12345', z'1234')
    end program

## Expected

A semantic error for each call, e.g.:

    semantic error: 'I' and 'J' arguments of 'iand' cannot both be BOZ literal constants

## Actual

LFortran compiles and runs it without any diagnostic, silently producing a
result (4132, 5877, 1745 on my build) with no defined meaning per the
standard.

## Environment

Built from source at HEAD (branch based on lfortran/lfortran main).

usefahmed07:
Fixes #12471

IAND, IOR, and IEOR each require per the Fortran standard that I and
J not both be BOZ literal constants (see gfortran's IOR docs: "I and
J shall not both be boz-literal-constants") — if only one argument
is BOZ, it's converted to the kind of the other; if both are BOZ
there's no other kind to convert to.

LFortran previously accepted iand(o'12345', z'1234') and similar
silently, with no defined result per the standard.

Added a check rejecting this, applied at both call sites that build
args for intrinsic elemental functions. Added a runtime test for the
valid one-BOZ case and a regression test (in continue_compilation_1.f90,
per repo convention) for the both-BOZ rejection.


integration_tests/iand_ior_ieor_boz.f90
```diff
@@ -0,0 +1,10 @@
+program iand_ior_ieor_boz
+implicit none
+integer :: k = 42
+if (iand(k, z'3456') /= 2) error stop
+if (iand(z'1234', k) /= 32) error stop
+if (ior(k, z'3456') /= 13438) error stop
+if (ior(z'1234', k) /= 4670) error stop
+if (ieor(k, z'3456') /= 13436) error stop
+if (ieor(z'1234', k) /= 4638) error stop
+end program
```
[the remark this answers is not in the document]
usefahmed07: > This program already works on main. Any reason to register it as an integration test in this PR?

It's a regression test for the fix makes sure rejecting both-BOZ-argument calls doesn't accidentally also reject the valid single-BOZ-argument case. Let me know if you'd rather I drop it.

src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -12907,6 +12929,7 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
                 for (size_t i = 0; i < args.size(); i++) {
                     expr_args.push_back(al, args[i].m_value);
                 }
+                reject_both_boz_args(intrinsic_name, expr_args);
```
certik: This it the wrong place to put this error check. This goes to the intrinsic function file, there is a verify phase which reports such errors. Please put it there.

usefahmed07:
> @usefahmed07 [#12472 (comment)](https://github.com/lfortran/lfortran/pull/12472#discussion_r3860509232) check this comment from @certik, after that we can merge this

I see it, btw. I was working on this too, so don’t worry — it’s done now.

tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -1903,19 +1903,19 @@ semantic error: The implied do loop variable 'r_idx' must be a scalar integer, n
...
-985 |         if (wide == narrow) print *, "eq"  ! {Error} operands of comparison operator '==' must be character with the same kind, found character(4) and character(1)
+991 |         if (wide == narrow) print *, "eq"  ! {Error} operands of comparison operator '==' must be character with the same kind, found character(4) and character(1)
     |             ^^^^^^^^^^^^^^ 
 
 semantic error: kind 3 is not supported for character, only 1 and 4 are
-   --> tests/errors/continue_compilation_1.f90:991:13
+   --> tests/errors/continue_compilation_1.f90:997:13
     |
-991 |         s = 3_"abc"  ! {Error} kind 3 is not supported for character, only 1 and 4 are
+997 |         s = 3_"abc"  ! {Error} kind 3 is not supported for character, only 1 and 4 are
```
certik, line 1920: I think the errors don't trigger now --- I don't see the errors here.

usefahmed07: > I think the errors don't trigger now --- I don't see the errors here.

Found it: moving the check into `verify_args` made it fire as an **ASR verify pass error**, not a `semantic error`:

```
ASR verify pass error: 'I' and 'J' arguments of 'Iand' cannot both be BOZ literal constants
```

`--continue-compilation` only collects `semantic error`s, so this can never show up in `continue_compilation_1.f90` no matter where I place the test.

Should I move it back to a semantic-stage check, or keep it in verify_args and test it separately instead? 🤔

PR #12478, opened 2026-08-12
Fix: skip unnecessary temp for ArrayConstant assigned directly to a target

### LINKED ISSUE #12198: Refactor : `impliedDoLoops` is creating temporaries for Arrayconstant while not needed by the pass
related to https://github.com/lfortran/lfortran/pull/12144
***
I think this pass is replacing ArrayConstants with temporaries. I think this isn't the job of the pass.
### Downsides
- 2 passes doing same responsibility (array_op pass, implied_do_loops pass)
- Bugs in `implied_do_loops` due to being not specialized in array operation defusing, unlike array_op pass.
- `ArrayConstant` replacer class in `implied_do_loops` pass could be costing us unnecessary tree traversals.
****
## What do to?
- Let's Find out why the class `ArrayConstantVisitor` and `ReplaceArrayConstant` were introduced.
- After doing the required job by these classes, Isolate their job by refactoring the visitors (maybe introduce a factory pattern, some static function trick, or any pattern that forces us to use the class properly)

usefahmed07:
Fix #12198


usefahmed07:
@certik Can you review it

certik:
Thanks. Let's run all tests first.

usefahmed07:
@certik The 2 failures were WASM-only (procedure_35/procedure_37), caused by this PR now assigning ArrayConstant directly to array vars the WASM backend's visit_Assignment wasn't handling that, it was writing the constant's buffer address into the target instead of the actual values. Added a case in asr_to_wasm.cpp to store the elements directly instead. Both tests and the full llvm_wasm suite pass locally now.

usefahmed07:
Hi @certik , WASM CI (Test LLVM 15/19) is failing on procedure_35/procedure_37 after this PR looks like asr_to_wasm.cpp's visit_Assignment doesn't handle a direct array = [array_constant] assignment (it always went through a temp before). I tried a fix but it didn't work, so I reverted it. Could you take a look or point me in the right direction?

certik:
I am not sure, we need to debug it why it fails.

PR #12480, opened 2026-08-12
Rework interactive feature

certik:

PR #12483, opened 2026-08-12
fix: run program units entered in interactive mode

certik:
## Summary

A cell containing a program unit compiled without error and then produced no
output at all. The program was translated and never called, so this looked to
the user like the compiler had silently ignored the cell.

In interactive mode a program is now compiled into an argument-less void
function named after the cell's run function, which the evaluator calls before
evaluating the cell's own statements. The name carries the cell number, so a
later cell may contain a program of its own.

The WASM executor renames it alongside the run function, so this works in the
Jupyter kernel as well as natively.

## Scope

`fortran_evaluator`, `visit_Program` in the LLVM backend, and the WASM
executor's symbol renaming.

## Verification

New test: a cell containing a program that assigns to a variable declared in an
earlier cell, checking the assignment actually happened. Fails before this
change, since nothing runs. Unit suite: 103/103.

## Note

Third of a five-PR stack replacing #12480, on top of #12482. Only the last
commit belongs to this PR. Independent of the cell-scoping work above it.


certik:
This looks good, it gets the main program working interactively, and the changes seem ok. We can further iterate / improve it later.

Issue #12479, opened 2026-08-12
Interactive: interface block does not link to a procedure from an earlier cell


certik:
## Summary

In interactive mode (REPL / Jupyter / JupyterLite lab), an `interface` block that
declares a procedure defined in an *earlier cell* does not link: the call is
emitted against the procedure's plain name, while the definition was emitted
under a cell-qualified one.

This is a regression from the in-progress work that gives each interactive cell
its own `TranslationUnit` and qualifies cell-level symbols by their cell
(`__cell<N>_<name>`), so that a redeclared name can shadow an earlier one while
both stay live. It does not affect ahead-of-time compilation, and it does not
affect ordinary cross-cell calls -- a procedure defined in an earlier cell can be
called directly without an interface block, and that works.

## Reproducer

Each block is one cell:

```fortran
integer :: z0
```

```fortran
integer function fn()
fn = 1
end function
```

```fortran
integer function call_fn()
    interface
        integer function fn()
        end function
    end interface
    call_fn = fn()
end function
```

```fortran
call_fn()
```

The last cell fails with an unresolved symbol:

```
JIT session error: Symbols not found: [ _fn ]
lookup() failed to find the symbol '__lfortran_evaluate_4'
```

Note the first cell: the problem needs the callee to be declared somewhere other
than the first cell. The first cell is deliberately left unqualified so that it
emits the same symbols as ahead-of-time compilation, so a callee defined there
links fine and the bug does not show.

## Cause

The IR emitted for the third cell shows it directly:

```llvm
define i32 @__cell3_call_fn() {
  ...
  %0 = call i32 @fn()
}
declare i32 @fn()          ; never defined -- cell 2 emitted @__cell2_fn
```

An interface block declares an *external* procedure, which binds by link name.
Cell qualification renames cell-level procedures, so the name the interface
block binds to no longer exists.

## Expected

`call_fn()` returns 1. In a notebook the natural reading of the interface block
is that it refers to the `fn` currently in scope, i.e. the one the most recent
cell declared.

## Suggested fix

Resolve an interface-block-declared external against the cell chain during
semantic analysis: if a procedure of that name is visible in an enclosing cell's
scope, bind to it, so codegen emits the qualified call. Falling back to the
plain external name when nothing in the chain matches keeps genuinely external
(e.g. C-interoperable) procedures working.

## Test

A regression test belongs in `src/lfortran/tests/test_llvm.cpp`, next to the
[...]

PR #12484, opened 2026-08-12
interactive: one TranslationUnit per cell

certik:
## Summary

Every cell was evaluated into a single, shared symbol table, so a name could
only ever have one meaning in a session: re-declaring it either clashed or
overwrote what earlier cells had been compiled against.

Each cell now gets its own `TranslationUnit`, and their symbol tables are
chained through `SymbolTable::parent`, oldest first:

```
cell 1   TranslationUnit   symtab(parent = nullptr)
cell 2   TranslationUnit   symtab(parent = cell 1's symtab)
cell 3   TranslationUnit   symtab(parent = cell 2's symtab)
```

Two properties follow from that shape. Cells are additive: a cell sees
everything declared before it, and nothing can reach forward into a later cell.
And re-declaring a name shadows it rather than replacing it, the way a nested
scope does, so code compiled earlier keeps using what it resolved to then,
while later cells resolve to the new declaration. That is what a Python
notebook does when a cell is re-run.

Two live declarations of one name need two symbols in the generated code, so
symbols declared at cell scope are qualified by their cell (`__cell<N>_`), the
way symbols in a module are qualified by the module. The first cell is left
unqualified, so it and ordinary compilation emit identical symbols.

## For reviewers writing new scope walks

Walks that looked for "the" translation unit by climbing to the root of a scope
chain now stop at the nearest one, which is the cell being compiled
(`ASRUtils::is_tu_scope`). With a single `TranslationUnit` the nearest one is
also the root, so ordinary compilation is unaffected. New code that climbs a
scope chain looking for the translation unit should stop there rather than at
`parent == nullptr`.

Ordinary compilation gets none of the chaining or copying, so it is not slowed
down by any of this.

## Scope

`asr_utils.h`, `asr_scopes`, `asr_verify`, the LLVM backend's symbol naming,
`fortran_evaluator`, one open-coded scope walk in `intrinsic_subroutine`, the
design note in `doc/src/interactive_cells.md`, tests, and reference updates.

## Verification

Tests cover shadowing a variable, a function, a subroutine and a module. Each
checks that the new binding is visible and that code compiled against the old
one still uses the old one; for a variable, function and module, an accessor
for each is also called side by side from a single cell, so both symbols are
demonstrably live at once. Unit suite: 110/110.

The reference tests that run with `--interactive-parse` shift their symbol
table counters by one, since each cell now has a scope of its own. Only the
numbering changes; the updated files contain no other difference.

## Note

Fourth of a five-PR stack replacing #12480, on top of #12483. Only the last
commit belongs to this PR.

Known limitation, filed as #12479: an `interface` block naming a procedure from
an earlier cell binds by link name, which cell qualification changes, so it
does not resolve. Ordinary cross-cell calls are unaffected -- a procedure from
an earlier cell can be called directly, without an interface block.

### DIFF
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -1,6 +1,7 @@
 #include <array>
 #include <cstring>
 #include <fstream>
+#include <set>
 
 #include <lfortran/fortran_evaluator.h>
 #include <libasr/codegen/asr_to_cpp.h>
@@ -25,6 +26,8 @@
 
 #ifdef HAVE_LFORTRAN_LLVM
 #include <libasr/codegen/evaluator.h>
+#include <llvm/IR/Function.h>
+#include <llvm/IR/Module.h>
 #include <libasr/codegen/asr_to_llvm.h>
 #ifdef HAVE_LFORTRAN_MLIR
 #include <libasr/codegen/asr_to_mlir.h>
@@ -159,7 +162,9 @@ Result<FortranEvaluator::EvalResult> FortranEvaluator::evaluate(
         && ASRUtils::is_character(*ASRUtils::expr_type(
             ASRUtils::EXPR(asr->m_items[asr->n_items - 1])));
 
-    // ASR -> LLVM
+    // ASR -> LLVM. The passes rewrite this tree; that is fine, it is this
+    // cell's own and is discarded afterwards. What later cells resolve
+    // against is the snapshot taken in get_asr3().
     Result<std::unique_ptr<LLVMModule>> res3 = get_llvm3(*asr,
         pass_manager, diagnostics, lm, lm.files.back().in_filename,
         nullptr);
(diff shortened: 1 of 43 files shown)


src/libasr/asr_scopes.h
```diff
@@ -50,13 +50,11 @@ struct SymbolTable {
...
-        while (global_scope->parent) {
-            global_scope = global_scope->parent;
-        }
-        return global_scope;
-    }
+    // The symbol table of the TranslationUnit this scope belongs to. In
+    // interactive mode there is one TranslationUnit per cell, chained through
+    // `parent`, so this stops at the innermost one: new symbols belong to the
+    // cell being compiled, not to the first cell of the session.
+    SymbolTable* get_global_scope();
```
certik: Then this function should be called `get_tu_scope()`, not `get_global_scope()`.

certik: Renamed to `get_tu_scope()`, along with its ten call sites. You are right that the name no longer described it: with one TranslationUnit per cell it returns the nearest translation unit, not the root of the chain.

`get_global_scope()` in `pass/coarray.cpp` is a different class's method and is left alone.

src/lfortran/fortran_evaluator.cpp
```diff
@@ -21,10 +22,13 @@
 #include <libasr/pickle.h>
 #include <libasr/utils.h>
 #include <libasr/asr_lookup_name.h>
+#include <libasr/serialization.h>
```
certik: Shouldn't be needed anymore.

certik:
I reviewed everything, it should be in good shape. If all tests pass, we can merge it.

src/lfortran/fortran_evaluator.cpp
certik: Removed. It was there for the earlier approach that serialized the ASR between cells, which this PR replaced with the in-memory snapshot, so nothing in the file uses it now.

PR #12485, opened 2026-08-12
interactive: use modules and procedures from earlier cells

certik:
## Summary

With one `TranslationUnit` per cell, an earlier cell's symbols live in a parent
scope of the cell being compiled. Five things did not account for that, and
together they made a notebook that declares its modules in one cell and uses
them in the next fail a different way each time one was fixed.

**`load_module` looked only in the cell being compiled.** A module declared
earlier was not found, so a modfile was searched for on disk, of which a
notebook has none. It reported `modfile was not found` naming the *dependency*
of the module being used, which was doubly confusing. It now resolves through
the chain.

**The copy taken for the next cell kept stale `ExternalSymbol` targets.** They
still pointed into the modules they were copied from, so a symbol re-exported
by another module looked absent. They are now pointed at the copies.

**Two passes only visited the cell being compiled.** `pass_array_by_data` and
`transform_optional_argument_functions` left a call to an earlier cell's
procedure in a form its already-compiled definition does not have: an
assumed-shape array argument reached a specialisation that was never emitted,
and an absent optional argument reached code generation as a null argument.
Both now visit the earlier cells, so a call is lowered the way the definition
it links to was.

**Those passes rewrite in place, so the chain needed protecting.** A cell now
hands the next one a copy of the whole chain rather than the scopes its own
passes have seen. Without this, re-running a cell that omits an optional
argument failed with the argument reported as required, because the procedure
the pass had rewritten was the one the next cell resolved against.

**Dead code elimination dropped procedures a later cell may call.** Any
procedure a cell declares may be called by a cell that does not exist yet, so
`unused_functions` is skipped in interactive mode.

## Scope

`load_module`, the evaluator's cell snapshot, `pass_array_by_data`,
`transform_optional_argument_functions`, one dedup in the LLVM backend, and a
`skip_pass` hook on the pass manager.

## Verification

Each fix has a test derived from a minimal reproducer: a module using another
module across cells, a re-exported module symbol, an optional argument,
re-running a cell that omits one, and calling a procedure no earlier cell
called. Unit suite: 116/116.

End to end, the `Plot.ipynb` demo now runs in the JupyterLite lab in Firefox:
its modules are declared in one cell and the program re-run from later cells,
repeatedly, rendering its plot each time.

## Note

Last of a five-PR stack replacing #12480, on top of #12484. Only the last
commit belongs to this PR.

The two chain-walking passes are handled individually. If a third pass needs
the same treatment, that is the point to replace this with something general
rather than extend the pattern again.

### DIFF
--- a/src/lfortran/fortran_evaluator.cpp
+++ b/src/lfortran/fortran_evaluator.cpp
@@ -428,6 +428,32 @@ void FortranEvaluator::drop_redefinitions(LLVMModule &m)
 }
 #endif
 
+namespace {
+
+// The duplicated symbols carry ExternalSymbols whose m_external still points
+// into the modules they were copied from. A later cell resolves a module by
+// name and finds the copy, so an ExternalSymbol pointing at the original does
+// not match it and the symbol looks absent. Point them at the copies.
+static void relink_external_symbols(SymbolTable *scope, SymbolTable *tu) {
+    for (auto &item : scope->get_scope()) {
+        ASR::symbol_t *s = item.second;
+        if (ASR::is_a<ASR::ExternalSymbol_t>(*s)) {
+            ASR::ExternalSymbol_t *es = ASR::down_cast<ASR::ExternalSymbol_t>(s);
+            ASR::symbol_t *mod = tu->resolve_symbol(es->m_module_name);
+            if (mod == nullptr || !ASR::is_a<ASR::Module_t>(*mod)) continue;
+            ASR::Module_t *m = ASR::down_cast<ASR::Module_t>(mod);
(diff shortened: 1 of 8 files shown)


certik:
I reviewed everything, this looks good, it copies all cells, as it should, I noticed that in the previous PR that this is likely a bug, so it was fixed here.

I am sure there will be more bugs, but I think this is a good start, and as we discover more issues, we'll add more tests.

Once all tests pass, we can merge it.

certik:
Looks good. Ready to merge if tests pass.

certik:
The CI was passing before, so I am going to merge it.

Issue #10888, opened 2026-03-28
Compile M_intrinsics package


HarshilShah1804:
Link to the package: https://github.com/urbanjost/M_intrinsics

**Dependencies**
- [x] M_CLI2: https://github.com/lfortran/lfortran/issues/7497
- [ ] M_match: https://github.com/lfortran/lfortran/issues/10883
- [ ] M_strings: https://github.com/lfortran/lfortran/issues/10884
- [ ] M_framework: https://github.com/lfortran/lfortran/issues/10885
- [ ] M_io: https://github.com/lfortran/lfortran/issues/10886
- [ ] M_attr: https://github.com/lfortran/lfortran/issues/10887

Current compilation status:
```bash
(base) harshil-shah@harshil-shah-HP-Pavilion-Laptop-15-eg2xxx:~/M_intrinsics$ fpm test --compiler=lfortran --flag "--cpp --realloc-lhs-arrays --no-style-suggestions"
M_vendor.F90                           done.
M_intrinsics.f90                       done.
fid.f90                                done.
nl-rename.f90                          failed.
[ 16%] Compiling...
Internal Compiler Error: Unhandled exception
Traceback (most recent call last):
  Binary file "/home/harshil-shah/lfortran/src/bin/lfortran", in _start
  File "./csu/../csu/libc-start.c", line 360, in __libc_start_main
  File "./csu/../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2812, in main
    return main_app(argc, argv);
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 2678, in main_app(int, char**)
    result = compile_src_to_object_file(opts.arg_file, outfile, compiler_options.time_report, false,
  File "/home/harshil-shah/lfortran/src/bin/lfortran.cpp", line 1249, in (anonymous namespace)::compile_src_to_object_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, LCompilers::CompilerOptions&, LCompilers::PassManager&, bool)
    res = fe.get_llvm3(*asr, lpm, diagnostics, lm, infile, &time_opt);
[...]

PR #12488, opened 2026-08-12
Fix Namelist formatting and USE renaming to comply with Fortran 2018 standards

HarshilShah1804:
Towards #10888

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1152,7 +1152,7 @@ RUN(NAME namelist_35 LABELS gfortran llvm)
 RUN(NAME namelist_36 LABELS gfortran llvm)
 RUN(NAME namelist_37 LABELS gfortran llvm)
 RUN(NAME namelist_38 LABELS gfortran llvm)
-
+RUN(NAME namelist_39 LABELS llvm) # Test in not compatible with gfortran 13.0 used at CI
 RUN(NAME nested_namelist_01 LABELS gfortran llvm)
 RUN(NAME nested_namelist_02 LABELS gfortran llvm)
 
--- a/src/libasr/runtime/lfortran_intrinsics.c
+++ b/src/libasr/runtime/lfortran_intrinsics.c
@@ -13779,6 +13779,14 @@ static void write_str(nml_writer_t *w, const char *s) {
     }
 }
 
+// Group and object names in namelist output shall be in upper case
+// (F2018 13.11.4.1); the descriptor keeps them lowercase for matching.
+static void write_str_upper(nml_writer_t *w, const char *s) {
+    while (*s) {
+        write_char(w, (char)toupper((unsigned char)*s++));
+    }
+}
+
 // Helper function to write a single namelist item value
 static int64_t get_element_size(const lfortran_nml_item_t *item);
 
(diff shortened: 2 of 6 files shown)


src/libasr/runtime/lfortran_intrinsics.c
```diff
@@ -14215,11 +14215,11 @@ static void parse_nml_value(const char *value_str, lfortran_nml_item_t *item, in
...
-            char lower[32];
-            strncpy(lower, value_str, 31);
-            lower[31] = '\0';
-            to_lowercase(lower);
-            bool val = (strstr(lower, "t") != NULL || strstr(lower, "1") != NULL);
+            char upper[256];
+            strncpy(upper, value_str, sizeof(upper) - 1);
+            upper[sizeof(upper) - 1] = '\0';
+            to_uppercase(upper);
+            bool val = (strstr(upper, "T") != NULL || strstr(upper, "1") != NULL);
```
certik: Why are you now converting everything to uppercase? Isn't it equivalent to the previous code that used lowercase?

HarshilShah1804: NAMELIST output uses uppercase for the group and group-object names. This is consistent with the Fortran standard's NAMELIST output requirements and with established Fortran implementations such as gfortran, Intel Fortran, and IBM XL Fortran. So, that's why I converted those to upper case.

HarshilShah1804: @certik Please share your thoughts on this.

Jatinagarwal24 (approved):
It now looks good to me.
Thanks.
@certik @HarshilShah1804 Pls once review it once.

certik (approved):
I think this looks good now.

src/libasr/runtime/lfortran_intrinsics.c
certik: @Jatinagarwal24 already fixed it. We should not be converting to uppercase at all like this.

We can convert to uppercase when printing, as we do now, that's fine.

integration_tests/CMakeLists.txt
```diff
@@ -1147,7 +1147,7 @@ RUN(NAME namelist_34 LABELS gfortran llvm)
 RUN(NAME namelist_35 LABELS gfortran llvm)
 RUN(NAME namelist_36 LABELS gfortran llvm)
 RUN(NAME namelist_37 LABELS gfortran llvm)
-
+RUN(NAME namelist_38 LABELS llvm) # Test in not compatible with gfortran 13.0 used at CI
```
certik: Why is it not compatible?

src/libasr/runtime/lfortran_intrinsics.c
HarshilShah1804: Yes, this is a cleaner approach. Thanks @Jatinagarwal24 for the fix.

integration_tests/CMakeLists.txt
certik: It works with Flang locally for me. 

However, LFortran also passes this test in main, so this is a bad test.

HarshilShah1804: It's a newer feature introduced in gfortran 14 (from what I saw at godbolt) and the test fails with gfortran 13 which is being used at our CI.

certik:
@HarshilShah1804 your original test had two bugs that made it pass in main. I fixed both, see the commit message.

HarshilShah1804:
Thanks @certik. I was also working on the test fix. But I think this is good now.

certik:
I rebased it one more time, should hopefully work now.

PR #12492, opened 2026-08-13
Implement more WASM interactive fixes

certik:


certik:
I just squashed the last few commits, no changes. All tests pass.

PR #12493, opened 2026-08-13
fix: lower character functions across interactive cells

anutosh491:
## Summary

`subroutine_from_function` transformed calls in the current cell but did not transform character-returning procedures reached through `ExternalSymbol`s or earlier cell scopes. This left the lowered call and procedure signature with different argument counts.

So we visit imported procedures and ancestor translation units so both sides use the same hidden-result ABI. 

I added a regression test covering repeated calls across cells, with and without another `use` statement.

### DIFF
--- a/src/lfortran/tests/test_llvm.cpp
+++ b/src/lfortran/tests/test_llvm.cpp
@@ -492,6 +492,43 @@ TEST_CASE("FortranEvaluator character result") {
     CHECK(r.result.str == "hello");
 }
 
+TEST_CASE("FortranEvaluator character function across cells") {
+    CompilerOptions cu;
+    cu.interactive = true;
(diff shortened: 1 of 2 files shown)


anutosh491:
On main 

image

on branch 

image

anutosh491:
I needed this for implementing the "pull" mechanism

certik:
I rebased on top of my fixes and let's run all tests.

certik:
I might have already fixed this bug in https://github.com/lfortran/lfortran/pull/12492. Let's check the new test in this PR --- if it passes in main, then it's fixed.

certik (approved):
I think that this is good, it transforms the functions everywhere, and we do a copy in the cells, so this is for the ASR that is a throw-away ASR. So I think this is the right fix.

PR #12495, opened 2026-08-13
Support rich display values as cell results (Pull mechanism)

anutosh491:
image


anutosh491:
Just like xeus-cpp/xeus-cling added a `mime_bundle()` for creating reusable rich values that Jupyter renders when evaluated as a cell result.

if i have understood it correctly. This is the pull mechanism we are looking for where in we can have an image as a value and then last value print it just like a python shell.

certik:
I think so, thanks for implementing it. I'll review it more carefully soon. I rebased it and let's run all tests.

certik:
I think the `summary` and `image` variables should be some custom type, possibly a user type that implements a method "`__show__`" or something like that. I think currently they are just strings, and strings you also want to be able to print as strings (and this should be the default I think).

anutosh491:
I think what you say above makes sense.

The current implementation overloads ordinary character values using a private prefix, which means strings can unexpectedly become rich output. I’ll change this so rich values use a derived type with a display binding, while ordinary character results always remain `text/plain`. The evaluator will recognize the display binding on the final expression and return an explicit rich result to the kernel.

anutosh491:
I’ve updated this to use a derived type with a type-bound `show()` function, so normal strings continue to be displayed as strings by default.

`mime_bundle` now provides this behaviour out of the box, but users can also define their own types with show(). Leaving an object of such a type as the last value in a cell renders its rich representation. I’ve added tests and updated the Rich Display notebook with reusable HTML, SVG and image values.

We can now do the Marie Curie example that xeus-cpp does : https://compiler-research.org/xeus-cpp/lab/index.html

image

certik:
This trips our binary detector:
```
checked out to commit -->  efa9faa3a2429cd549b972da617f77fda061487b
FAIL --- binary file detected
Binary File Name : (share/lfortran/nb/marie.png) In Commit : (efa9faa3a2429cd549b972da617f77fda061487b)
```

Instead of a pre-made png image, can we generate something on the fly? I don't like adding binary files to the repo, unless needed.

anutosh491:
> I don't like adding binary files to the repo, unless needed.

Ahh we just have the marie curie image in most of our kernels. But yeah we don't necessarily need it. Let me get rid of that example !

PR #12496, opened 2026-08-13
Fix the broken star history chart in the README

OctoBored:
The star history chart in the README is broken due to GitHub stargazer API restrictions. This change updates the chart link to a working endpoint that uses an alternative data source, so the chart renders correctly again.


certik (approved):
Thanks!

PR #12500, opened 2026-08-14
Add an ASR regression corpus and close four verifier gaps

certik:
## Summary

Makes a minimized ASR graph usable as a checked-in regression fixture, and
closes four verifier gaps found by the fuzzing work in #12453.

#12462 landed the lossless ASR text format, so an ASR graph can be printed and
read back. This PR adds what makes that useful for testing: stable diagnostic
codes, an initial-only verification mode, and fixtures registered as ordinary
reference tests. An ASR-level bug no longer has to be re-derived from Fortran
source to be regression tested.

This is the third PR in the stack, on top of #12499 (merged). The next PR adds
the fuzzing harness.

## Scope

Seven commits, ordered so that nothing changing existing behavior comes before
the infrastructure.

Infrastructure (moves no existing test):

1. `feat: add stable ASR verifier diagnostic codes and --verify-asr` — gives
   verifier diagnostics optional stable `asr.verify.*` codes rendered into the
   message, so a fixture pins the rule rather than the wording, and adds
   `--verify-asr`, which parses a standalone ASR file and runs only its initial
   verifier.
2. `refactor: share the post-ASR object pipeline` — moves the ASR-to-object path
   into `src/lfortran/pipeline.{h,cpp}` as `load_input_asr` and
   `compile_asr_to_object`. No behavior change.
3. `feat: add --verify-all-passes` — enables post-pass verification
   independently of `WITH_LFORTRAN_ASSERT`, reporting failures through a
   dedicated exception carrying the pass name.

Verifier gaps, one per commit, each with its own `.asr` fixture:

4. call argument types and kinds
5. assignment value types
6. function signature types
7. unsupported scalar type kinds

Each rejects ASR that previously verified successfully and failed later as a
pass assertion, an unknown exception or a miscompile.

## Fixtures

Fixtures live under `tests/asr/` and are registered in `tests/tests.toml` like
any other reference test — there is no separate test harness:

- `tests/asr/compile/` is registered with `asr_clojure` and `llvm`;
- `tests/asr/verify/` is registered with `asr = true`, so the stored reference
  captures the exact diagnostic, its stable code, the span and the return code.

They are stored in the printer's indented positional form, so a diagnostic
points at one short line instead of at a single very long one:

```
ASR verify pass error [asr.verify.call.actual_type_matches_formal]: Actual argument type i16 does not match formal argument type i32
   --> tests/asr/verify/call_actual_type_mismatch.asr:107:18
    |
107 |                 (Var
    |                  ^^^ failed here
```

A fixture the verifier accepts can be regenerated with
`--show-asr --clojure --no-member-names`.

## Verification

Debug build with LLVM, runtime stacktraces and assertions, on macOS.

| Suite | Result |
| --- | --- |
| CTest | 6/6 pass |
| Reference suite `./run_tests.py` | pass |
| Integration suite, LLVM backend | 3854/3854 pass |

The four new checks are covered by the fixtures added alongside them.

## Notes for review

- The only CTest changes are path updates to the three tests that already
  existed on `main` (`test_asr_text_compile`, `test_asr_text_object`, and the
  EDN conformance check), following the fixture into `tests/asr/compile/`.
- `--verify-all-passes` changes what an assertion build prints when a pass
  corrupts ASR: an exception naming the pass, rather than a generic ICE.
- `--verify-asr` currently requires the ASR to contain a main program, which
  suits the executable fixtures here; module-only ASR would need that relaxed.


certik:
Reviews:

* GLM-5.3: https://gist.github.com/certik/1881a89305deefd070461d528f251b72
* Qwen3.8-27B: https://gist.github.com/certik/7d051ff50cffcfd4053b8bdfb97bcca3

PR #12503, opened 2026-08-14
fix: rename test files so they don't fail  over parallel ctest

### LINKED ISSUE #12498: Integration tests race over a shared test.txt and fail intermittently under parallel ctest
## Summary

Three integration tests all write a file called `test.txt` in the same
directory, and `ctest` runs them in parallel, so they race: one deletes or
truncates the file while another has it open. The suite then fails on a test
that is fine on its own, and passes when the job is re-run.

## The tests

| test | what it does with `test.txt` | labels |
| --- | --- | --- |
| `integration_tests/endfile_iomsg_01.f90` | `open(..., status="replace")`, `write`, `endfile`, `close(status="delete")` | gfortran llvm llvm_wasm llvm_wasm_emcc |
| `integration_tests/backspace_iomsg_01.f90` | `open(..., status="replace")`, `write`, `backspace`, `close(status="delete")` | gfortran llvm llvm_wasm llvm_wasm_emcc |
| `integration_tests/format_78.f90` | `open(unit=10, ..., status="replace")`, writes, `close` | gfortran llvm |

They all also use unit 10, though that alone is harmless -- each test is its own
process.

## Seen in CI

`Test LLVM 19 (ubuntu-latest)`, run 31672720689 (PR #12492):

```
1697/2447 Test #1699: endfile_iomsg_01 ...................................***Failed    0.49 sec
Error in deleting file!

	1699 - endfile_iomsg_01 (Failed)
Errors while running CTest
```

`backspace_iomsg_01` was running at the same time and passed. Re-running the
job passed with no changes, which is what a race looks like.

## Fix

Give each test a filename of its own -- `endfile_iomsg_01.txt`,
`backspace_iomsg_01.txt`, `format_78.txt` -- so nothing is shared. A test that
writes a file should name it after itself; `test.txt` is a name a fourth test
will pick again.

Worth a grep for other shared fixed filenames in `integration_tests/` while
fixing this one.

yuvrajsinghgmx:
Closes #12498


opixdown (approved):

opixdown:
LGTM, thank you for the pr

opixdown:
@yuvrajsinghgmx can you please resolve the issue

yuvrajsinghgmx:
> @yuvrajsinghgmx can you please resolve the issue

Done

PR #12506, opened 2026-08-15
Fix several bugs found by the ASR fuzzer

certik:
## Summary

The ASR fuzzer (#12453) started producing graphs the verifier accepted and the compiler then could not lower: mixed-kind binops and comparisons, `ArrayItem` types that disagreed with the array element, variable initializers of the wrong kind or family, `StructConstructor` arguments that did not match their members, and a literal `dim` outside `1..rank` that crashed the array-bound pass.

This PR puts those invariants in `asr_verify`, broadens the schema generator and mutation seed corpus so the campaign actually reaches those nodes, and links every `tests/asr/compile/` fixture so a graph the verifier accepts must also produce an executable.

## Scope

- Verifier: operand kinds on numeric/logical ops and comparisons; `ArrayItem` vs element type; variable initializer vs declaration (kind and family); `StructConstructor` argument count and types, including an acyclic parent walk; literal `dim` in `1..rank` behind `check_standalone_rules` (standalone ASR only).
- Tests: negative fixtures under `tests/asr/verify/`; CTest now globs and links every `tests/asr/compile/*.asr` (`CONFIGURE_DEPENDS`). The programs are not run. Initializer/ArrayItem/constructor checks compare kinds inside one type family only: equivalence can attach another object's initializer across families (integer storage viewed as real), which is a bit reinterpretation rather than a Cast.
- Fuzzer: wider schema generator (reals, logicals, loops, calls, arrays, functions, complex) and a larger mutation seed corpus.
- Also: unique scratch files for three I/O integration tests that raced on `test.txt` under `ctest -j`; ignore `_lfortran_generated_file*`.

Follow-ups, not in this PR: constant-folded `dim` (not just a bare integer literal), and applying `check_standalone_rules` to post-`ast_to_asr` programs.

## Verification

- New `tests/asr/verify/` fixtures fail without the corresponding check and pass with it, including `struct_constructor_cyclic_parent`.
- `python3 ./run_tests.py -t asr/verify/` — all passed.
- `equivalence_39` (`real` equivalenced with an initialized `integer`) compiles again after restoring the family skip.
- I/O race: `backspace_iomsg_01`, `endfile_iomsg_01`, and `format_78` failed 2/10 concurrently before the rename and 0/20 after.

## Rationale

These are graph well-formedness rules the frontend already establishes with explicit Casts. The verifier is the right layer: LLVM, `class_constructor`, and `update_array_dim_intrinsic_calls` should only lower what ASR already made explicit. A kind or family disagreement is invalid ASR, not a backend conversion problem.

`check_standalone_rules` stays off after passes, because the array passes legitimately build intermediate `ArraySize`/`ArrayBound` nodes whose rank does not yet agree.

Refs #12453


src/lfortran/tests/CMakeLists.txt
```diff
@@ -97,27 +97,40 @@ add_test(test_lfortran ${PROJECT_BINARY_DIR}/test_lfortran)
...
-    if (NOT WIN32)
+    # Every fixture under tests/asr/compile/ must link an executable. That is
+    # the second half of the contract the fuzzer enforces, and it is what
+    # proves the ASR was well formed enough to reach a binary. The programs
+    # are deliberately not run: a generated program may fault at runtime for
+    # reasons no ASR verifier could predict, so only linking is a property of
+    # the ASR itself.
+    #
+    # Adding a fixture needs a re-configure for the glob to pick it up.
+    file(GLOB LFORTRAN_ASR_COMPILE_FIXTURES
```
certik: **[suggestion]** The new compile-fixture contract is a configure-time `file(GLOB ...)` with an explicit comment that adding a fixture needs a reconfigure. That is the same miss this commit set out to fix (a new `tests/asr/compile/*.asr` is silent until someone remembers an extra step). The top-level `CMakeLists.txt` already uses `CONFIGURE_DEPENDS` for the analogous mod-file glob.

**Suggestion:** `file(GLOB LFORTRAN_ASR_COMPILE_FIXTURES CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/tests/asr/compile/*.asr)` so Ninja/Make re-glob on build. Keep the Windows link skip as-is.

src/libasr/asr_verify.cpp
```diff
@@ -1880,6 +1943,230 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
...
+    // positionally, so a count mismatch is a memory error waiting to happen.
+    void visit_StructConstructor(const StructConstructor_t &x) {
+        ASR::symbol_t *struct_sym = x.m_dt_sym == nullptr
+            ? nullptr : ASRUtils::symbol_get_past_external(x.m_dt_sym);
+        if (!diagnostics.has_error() && struct_sym != nullptr
+                && ASR::is_a<ASR::Struct_t>(*struct_sym)) {
+            std::vector<ASR::Struct_t*> chain;
+            ASR::Struct_t *struct_type =
+                ASR::down_cast<ASR::Struct_t>(struct_sym);
+            while (struct_type != nullptr) {
```
certik: **[suggestion]** `visit_StructConstructor` walks `m_parent` with no visited set. A mutated graph whose parent chain cycles (A→B→A) never reaches `require_id` and hangs the verifier. `class_constructor` has the same walk, but that pass is not the fuzzer's first line of defense; this visitor now is. Existing `struct_has_member` has the same shape, so this is a new copy of a known unbounded walk on fuzzer-controlled pointers.

**Suggestion:** Track seen `Struct_t*` (or a small hop limit) and `require` that the parent chain is acyclic, then fail the constructor check. Do not rely on the later pass assertion.

src/libasr/asr_verify.cpp
```diff
@@ -859,6 +862,42 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
...
+            ASR::ttype_t *actual = ASRUtils::type_get_past_array(
+                ASRUtils::type_get_past_allocatable_pointer(initial_type));
+            // A character initializer is padded or truncated to the
+            // declared length, so the two legitimately differ. A kind at or
+            // above the parameterized derived type sentinel is a type
+            // parameter rather than a storage size, and a parameterized type
+            // carries it on the declaration or on the initializer depending
+            // on where it has been substituted, so it is not comparable.
+            if (is_struct_like_type(declared) || is_procedure_type(declared)
+                    || declared->type != actual->type
```
certik: **[nit]** Variable initializers, ArrayItem types, and StructConstructor arguments all skip the check when `declared->type != actual->type`. That family skip is required for binops (`real ** integer` is a frontend `RealBinOp` whose exponent is intentionally left integer). It is not required here: the frontend casts initializers, ArrayItem element types, and constructor args, so an `integer(4)` parameter whose `m_value` is a `RealConstant` is invalid ASR and still produces the LLVM store-type failure this PR is closing. The fuzzer will keep reporting those as compile bugs.

**Suggestion:** Drop the family-type early-return in the Variable / ArrayItem / StructConstructor checks (keep the binop one, and keep the String / kind≥1000 skips on initializers). Add one verify fixture that is an integer variable initialized with a real constant.

PR #12512, opened 2026-08-15
fix: verify call argument rank, allocatable, and pointer

certik:
## Summary

The ASR fuzzer accepted a scalar actual for an array dummy; LLVM then rejected the module (`Call parameter type does not match function signature`). Kind, family, omitted required arguments, extra arguments, and `intent(in)` passed to `intent(out)` were already rejected.

This PR checks rank, allocatable, and pointer on calls, and extends the schema generator so the campaign actually produces those combinations.

## Scope

- Verifier: an array dummy requires a matching rank (except assumed-rank and elemental); an allocatable or pointer dummy requires an actual with the same wrapper. The other direction is valid Fortran (an allocatable actual may be passed to a nonallocatable dummy).
- Generator: optional, allocatable, pointer, `intent(out)`, and required-plus-optional calls, plus a catalog of mismatches.
- Mutations: `:presence` and dummy `:intent` (`In`/`Out`/`InOut` only).
- Seeds: `present_01`, `elemental_21`, `functions_05`.
- Tests: `call_rank_mismatch`, `call_allocatable_mismatch`, `call_pointer_mismatch` under `tests/asr/verify/`.

## Verification

- `python3 ./run_tests.py -t asr/verify/` — all passed.
- Schema-invalid, 80 cases: 67 verify, 13 compile (valid-direction cases), 0 failures.
- Schema-valid, 40 cases: 40 compile, 0 failures.

## Rationale

`check_equal_type` strips `Allocatable` and `Pointer`, so those wrappers and rank have to be compared in `verify_args`. A rank mismatch is invalid ASR, not something LLVM should discover. An allocatable dummy has a different ABI from a nonallocatable one, so the dummy-requires-wrapper rule belongs in the verifier. Passing an allocatable actual to a nonallocatable dummy is standard Fortran and must stay accepted (`allocate_01`).

Refs #12453


src/libasr/asr_verify.cpp
```diff
@@ -1528,6 +1533,62 @@ class VerifyVisitor : public BaseWalkVisitor<VerifyVisitor>
                         !ASRUtils::is_intrinsic_symbol(x.m_name) &&
                         !is_method && !struct_argument &&
                         !procedure_argument) {
+                    // These wrapper and rank rules hold for a complete
+                    // standalone graph. After a pass the dummy may have been
+                    // rewritten (openmp turns allocatable into pointer;
+                    // pass_array_by_data changes ranks), so they are not
+                    // applied to intermediate ASR.
```
certik, line 1540: This seems like a bug to me, but it's not a blocker for this PR. I think we should fix it, so that the ASR verify is identical, no matter if it is an initial ASR, or after a pass.

certik: #12513

PR #12520, opened 2026-08-17
fix: check type-bound procedure override interfaces

### LINKED ISSUE #12311: LFortran doesn't check interfaces

certik:
An overriding type-bound procedure must present the same interface as the binding it overrides, apart from the passed-object dummy argument. AST->ASR never checked this, so a derived type could override a deferred binding with a procedure taking different arguments; a dispatch through the parent type then calls it with arguments it was not declared to take.

The ASR verifier already caught this, but it runs only in a build configured with WITH_LFORTRAN_ASSERT, so a release compiler accepted the code silently and a debug one reported an internal verifier message instead of a semantic error. The check now runs in AST->ASR, where the decision belongs.

The comparison the verifier performed moves to ASRUtils::interface_mismatch and ASRUtils::binding_override_mismatch so both callers share one implementation; the verifier keeps its existing messages and check identifiers, and its standalone ASR tests pass unchanged.

Under --continue-compilation the rejected binding is pointed back at the one it failed to override, which is what the type would have had without the override. Abandoning the binding instead would leave a deferred one nothing supplies, and keeping it would leave an override no interface describes; both make the translation unit fail verification the next time a USE of any module in it revalidates the tree.

The case added to tests/errors/continue_compilation_1.f90 is written on two lines, one for the type and one for the procedure, so that it fits the blank lines already there and no other error keeps a different line number.

Fixes #12311.

### DIFF
--- a/src/lfortran/semantics/ast_symboltable_visitor.cpp
+++ b/src/lfortran/semantics/ast_symboltable_visitor.cpp
@@ -4392,6 +4392,83 @@ public:
                 sync_pdt_specialization_symbols(clss, proc_scope);
             }
         }
+        check_class_procedure_overrides();
+    }
+
+    // The name of the derived type that declares the binding `x` overrides.
+    std::string overridden_binding_owner(
+            const ASR::StructMethodDeclaration_t &x) {
+        ASR::StructMethodDeclaration_t *base = ASRUtils::overridden_binding(x);
+        if (base == nullptr || base->m_parent_symtab == nullptr) return "";
+        ASR::symbol_t *owner = ASR::down_cast<ASR::symbol_t>(
+            base->m_parent_symtab->asr_owner);
+        return std::string(ASRUtils::symbol_name(owner));
+    }
+
+    // Points a rejected override back at the binding it failed to override,
+    // which is what the type would have had if the override were absent.
+    void adopt_overridden_binding(ASR::StructMethodDeclaration_t *x) {
+        ASR::StructMethodDeclaration_t *base = ASRUtils::overridden_binding(*x);
+        if (base == nullptr) return;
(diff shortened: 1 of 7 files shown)


certik:
Let's wait with this one, I am doing an ASR verify refactoring fix, which might conflict, and we should do the refactoring first.

certik:
Here is the refactoring PR that I would like to go in first: https://github.com/lfortran/lfortran/pull/12523.

PR #12522, opened 2026-08-17
fix: component array lowering, character ABI, and I/O

konradha:
Fixing connected failures in semantic analysis, ASR array representation, LLVM descriptor lowering, external procedure ABI, persistent storage, and IO. Fourteen new integration tests isolate the individual compiler failures, and the existing external-symbol regression now covers separate compilation. Each test runs with both GFortran and LFortran's LLVM backend.

Expressions such as `values(:,1)%component(1)` get their rank from the array of structures, even though the final component subscript selects one scalar from each structure. Semantic analysis now preserves the base dimensions and physical array type, represents the result as an `ArraySection`, and rejects references where multiple component parts retain nonzero rank.
A read-only explicit-shape dummy can receive a packed contiguous temporary. A writable dummy needs a live strided view so assignments reach the original components. The temporary pass now gets result dimensions from the structure array and uses dummy intent to select materialization or association. Pointer assignment constructs a descriptor with the correct lower bounds, extents, offset, and structure-adjusted strides. Type extension also receives explicit aggregate traversal because an inherited member index applies inside the parent object stored in field zero of the child LLVM structure.

`READ` implied-do expansion previously replaced loop variables only when the complete value was an `ArrayItem`. It could handle `a(i)` but could not reach `i` inside `records(i)%field`, casts, or nested expressions.

External character procedures now use one ABI decision across LLVM function types, call arguments, and function entry. Compiler-generated implicit interfaces keep the established `BindC` ASR representation. LLVM identifies these interfaces from their absent source-name locations and applies the GNU Fortran character ABI. Explicit C bindings, intrinsics, abstract interfaces, and module procedures retain their established conventions.

The IO changes treat descriptors and lexical rules as authoritative. Direct unformatted output checks descriptor strides before passing bytes to the runtime. Type-specific code only converts and stores a valid token.


konradha:
Rethinking a few things here as the ABI changes were too aggressive I realized.

PR #12526, opened 2026-08-18
fix: drop two ASR verify rules that do not hold

certik:
Part 2 of 7 splitting #12523, which stays open as the combined branch.

Stack: #12525 → **this** → #12531

Branches are stacked, so this targets `main` and its diff includes #12525.
**Review the last commit only.**

## Summary

Both rules were added behind `check_standalone_rules`, so neither has ever run
in the compiler. Turning them on shows that neither states an invariant.

`associate.value_is_target` reads `ASR::Associate` as a pointer assignment, but
the node also serves the ASSOCIATE construct, whose selector need be neither a
pointer nor a target. It fires 87 times on the integration suite, 22 of those
at initial verification.

`module.dependency_is_present` requires every name in a module's dependency
list to be a module in the same translation unit. That list names the modules
still to load, so a name with nothing behind it is its normal state rather than
a defect. Filtering the list in `UpdateDependenciesVisitor` to make the rule
hold broke module loading outright (`the module 'iso_fortran_env' was not
found`).

Their fixtures and fuzzer generator families go with them.

## Rationale

If `Module::dependencies` is meant to be an invariant, the rule should come
back with a definition of what belongs in it — but as the field is used today,
this rule is not that definition.

## Verification

`ctest` 13/13, `./run_tests.py` passes.


tests/asr/verify/pointer_to_nontarget.asr
```diff
@@ -1,79 +0,0 @@
...
-            (Var
-              (SymbolRef 1 "plain")
-            )
-          )
-        ]
-      )
-    }
-  )
-  []
-)
```
certik, line 79: Find for now, but after everything is merged, these two ASR files should either compiler, or give an error, so we need to put them back. They discovered a real issue.

certik: #12534

PR #12527, opened 2026-08-18
fix: keep a variable's type declaration inside its own scope

certik:
Part 3 of 7 splitting #12523, which stays open as the combined branch.

Stack: #12525 → #12526 → **this** → #12531

Branches are stacked, so this targets `main` and its diff includes the parts
below it. **Review the last commit only.**

## Summary

`type_declaration` names a symbol, so it has to be reachable from the
variable's scope the same way any other name is. `variable.type_declaration_in_scope`
and `struct.parent_in_scope` ran only for ASR read from a file; enforcing them
everywhere, 122 of the integration tests fail.

Import centrally, in `make_Variable_t_util`, rather than at each producer.
`ASRUtils::import_type_declaration` reuses a name already standing for the
type, falls back to an ExternalSymbol through the owning module or derived
type, and leaves the declaration alone when the symbol is local to a program or
a procedure and so cannot be named at all.

Producers that set `m_type_declaration` directly go through the same helper:

- select-type blocks, which used `import_struct_type` — a silent no-op for a
  Block scope;
- PDT specialisation;
- implicit-interface synthesis;
- `pass_array_by_data`, which reaches its rewritten procedure through an
  ExternalSymbol, so the specialisation is imported the same way the original
  was;
- a procedure with ENTRY points becomes several sibling functions holding
  copies of the same dummy, so an interface synthesised into one of them
  cannot be named from the others; each sibling now gets its own copy.

A generated helper has to live where it can name what it uses. One that hoists
an expression naming a derived type or procedure contained in a program cannot
go in the global scope, and no ExternalSymbol reaches it there.
`PassUtils::instantiation_scope` decides this once and the four instantiation
sites share it. `reduce` stamped the caller's own procedure onto its dummy's
type declaration; the dummy describes the interface it accepts, so it now gets
an interface built from the FunctionType — which is also what lets the helper
stay shared between call sites.

`pass_utils::create_var` no longer carries a type declaration over for a
non-derived-type temporary, where it named a symbol unrelated to the
temporary's type.

`relink_copied_symbols` pointed a copied ExternalSymbol at another
ExternalSymbol when the module's own entry for the name was itself an import.
ASR verify already forbade that chain; nothing walked it until the
type-declaration import did, at which point the Jupyter kernel died on the
second cell that used such a module. New `FortranEvaluator` test covers it.

## Verification

`ctest` 13/13, `./run_tests.py` passes, `integration_tests -j16` passes.
The new evaluator test fails on `main` and passes here.


certik:
This is a good one, need to review it more carefully, but it strengthens the ASR verify and then fixes the whole codebase to conform, previously we were creating invalid ASR and didn't know it.

certik:
I reviewed it, I think this is good to merge.

PR #12528, opened 2026-08-18
fix: check an intrinsic's dim argument at every point in the pipeline

certik:
Part 4 of 7 splitting #12523, which stays open as the combined branch.

Stack: #12525 → #12526 → #12527 → **this** → #12531

Branches are stacked, so this targets `main` and its diff includes the parts
below it. **Review the last commit only.**

## Summary

`dim` selects one of an array's dimensions, so a constant outside `1..rank` is
invalid: the pass that folds these intrinsics indexes `dims[dim - 1]` directly
and reads past the dimension array rather than diagnosing it. The rule that
says so ran only for ASR read from a file.

Enforcing it everywhere shows `spread` building an `ArraySize` over the
result's rank, which is one more than the array it spreads. That branch is only
selected when `i < dim`, so it never names the appended dimension, but the
index still has to name a dimension the array actually has for the expression
to be well formed.

## Verification

`ctest` 13/13, `./run_tests.py` passes, `integration_tests -j16` passes.


certik:
This is good to merge.

PR #12530, opened 2026-08-18
fix: check a call's arguments against the procedure at every point

certik:
Part 6 of 7 splitting #12523, which stays open as the combined branch.

Stack: #12528 → #12529 → **this** → #12531

Rebased onto `main` now that #12525, #12526 and #12527 are merged, so the
diff is down to three commits. #12528 and #12529 are still below it;
**review the last commit only.**

## Summary

The rules that compare an actual argument with the dummy it is passed to ran
only for ASR read from a file. Enforcing them everywhere shows both rules that
were stated too strongly and passes that leave a call disagreeing with the
procedure it calls.

### Rules, corrected against the standard

- an interface declaring no arguments constrains nothing — that is the shape
  `EXTERNAL f` and `procedure(), pointer` produce, and ASR cannot yet tell it
  from a genuinely argumentless interface;
- a `POINTER, INTENT(IN)` dummy also accepts a non-pointer actual that is a
  valid target for it (F2008 12.5.2.7). gfortran and flang both accept it and
  Caffeine relies on it;
- sequence association lets an explicit-shape or assumed-size dummy take an
  array element, which is a scalar (F2008 12.5.2.11); splpak and LAPACK both
  rely on it. An assumed-shape dummy takes its extents from the actual and an
  allocatable or pointer dummy carries the actual's own storage, so neither
  takes an element — new fixture `call_scalar_to_assumed_shape.asr` pins that
  boundary, and `call_rank_mismatch.asr` now takes the direction that stays
  invalid;
- the allocatable, pointer and rank rules describe how an implementation
  receives its arguments, so they apply where the procedure itself is in hand.
  Under `--implicit-interface` the frontend infers an interface from one call
  site rather than reading a declared one, and ASR carries nothing that
  separates the two, so another call site was measured against a guess (LAPACK).
  Separate compilation is unaffected: a module procedure reaches its caller
  with its body.

### Passes

- `transform_optional_argument_functions` replaces an optional dummy with a
  presence flag, rewriting the procedure and the interface it was declared
  from. A `procedure(iface)` entity holds a copy of that interface's
  FunctionType, so the copy was left declaring an argument list the procedure
  no longer has — 4 arguments against 6 in the Caffeine build.
- `openmp` rewrites a dummy into a pointer so the parallel region can reach it
  through the thread data, while callers still pass what the procedure used to
  declare. Point a pointer at the actual and pass that; both are descriptors,
  so this only states in ASR what the call already does.
- `array_op` deliberately leaves a GPU kernel's array operations alone, because
  the Metal and CUDA backends lower them directly. An elemental intrinsic there
  therefore still has its array arguments, and `intrinsic_function` instantiated
  a helper from the element types and handed it those arrays. Leave such an
  intrinsic alone in a kernel, the same way `array_op` does — but only when its
  arguments are still arrays, since the Metal tests rely on scalar-argument
  instantiation.

## Verification

`ctest` 13/13; `./run_tests.py`, `integration_tests -j16`,
`integration_tests -b llvm_omp` (81/81) and `integration_tests -b metal`
(195/195) all pass.

New integration tests `pointer_intent_in_01`, `sequence_assoc_01` and
`separate_compilation_51`; each fails on `main` and passes here.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4321,6 +4321,8 @@ RUN(NAME arrayprint_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME print_arr_05 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME print_arr_09 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME pointer_to_cptr LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
+RUN(NAME pointer_intent_in_01 LABELS gfortran llvm)
+RUN(NAME sequence_assoc_01 LABELS gfortran llvm)
 
 RUN(NAME binop_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME binop_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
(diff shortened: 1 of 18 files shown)


certik:
I think that this is fine. It fixes the ASR after we made the verify more rigorous.

PR #12531, opened 2026-08-18
refactor: remove the ASR verify option that split the rules in two

### LINKED ISSUE #12513: Tighten the ASR verify in passes
> This seems like a bug to me, but it's not a blocker for this PR. I think we should fix it, so that the ASR verify is identical, no matter if it is an initial ASR, or after a pass. 

 _Originally posted by @certik in [#12512](https://github.com/lfortran/lfortran/pull/12512/changes#r3790183000)_

certik:
Last of the 7 parts splitting #12523, and now the only one left: everything
below it is merged, so this is down to a single commit.

## Summary

`ASRVerifyOptions::check_standalone_rules` gated rules so they ran only for ASR
read from a file — never for ASR the frontend built or a pass rewrote. Every
rule added since the flag appeared was therefore dead in the compiler itself.

With the rules corrected and their producers fixed by #12526, #12527, #12528,
#12529 and #12530, nothing reads the option any more, so it goes.
`check_external` stays: a modfile being deserialized genuinely has no
ExternalSymbol targets to follow yet.

ASR verify is now identical wherever it runs.

Fixes #12513.

## Verification

Rebased onto `main` after #12530 merged. `ctest` 13/13, `./run_tests.py`
passes, `integration_tests -j16` passes.

#12523 stays open as the combined branch and can be closed once this lands.

### DIFF
--- a/src/bin/lfortran.cpp
+++ b/src/bin/lfortran.cpp
@@ -813,7 +813,6 @@ int verify_asr_input(const std::string &infile,
     LCompilers::ASRVerifyOptions verify_options;
     verify_options.check_external = true;
     verify_options.require_main_program = true;
-    verify_options.check_standalone_rules = true;
     bool verified = LCompilers::asr_verify(
         *result.result, verify_options, diagnostics);
     std::cerr << diagnostics.render(lm, compiler_options);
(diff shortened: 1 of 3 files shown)


certik:
This is good to merge if tests pass.

PR #12535, opened 2026-08-18
fix: support labeled END statement targets

### LINKED ISSUE #12516: END statements fail as branch targets in 0.64.0
END statements may be branch target statements. Per the Fortran
standard:

## 5.3.3   The END statement

```text
Each program unit, module subprogram, and internal subprogram shall have
exactly one END statement.

The end-program-stmt, end-function-stmt, end-subroutine-stmt, and
end-mp-subprogram-stmt statements are executable, and may be branch
target statements (11.2).

Executing an end-program-stmt initiates normal termination.

Executing an end-function-stmt, end-subroutine-stmt, or
end-mp-subprogram-stmt is equivalent to executing a return-stmt with
no scalar-int-expr.

The end-module-stmt, end-submodule-stmt, and end-block-data-stmt
statements are nonexecutable.
```
## using END statements as branch targets
```fortran
program mre
   print *,"Why shouldn't you trust atoms?"
   call sub()
   goto 99999
contains
   subroutine sub()
      write(*,*)'Because they make up everything.'
      goto 11111
11111 end subroutine sub
99999 end program mre
```
```diff
+aoccflang520         :     2	 Why shouldn't you trust atoms?
+aoccflang520         :     3	 Because they make up everything.
+flangtrunk           :     2	 Why shouldn't you trust atoms?
+flangtrunk           :     3	 Because they make up everything.
+ifxlatest            :     2	 Why shouldn't you trust atoms?
+ifxlatest            :     3	 Because they make up everything.
+ifxlatest_standard   :     2	 Why shouldn't you trust atoms?
+ifxlatest_standard   :     3	 Because they make up everything.
+nvfortran_x86_26_3   :     2	 Why shouldn't you trust atoms?
+nvfortran_x86_26_3   :     3	 Because they make up everything.
-lfortranlatest       :     3	syntax error: Token 'end subroutine' is unexpected here
-lfortranlatest       :     4	 --> /app/example.f90:9:7
-lfortranlatest       :     5	  |
-lfortranlatest       :     6	9 | 11111 end subroutine sub
-lfortranlatest       :     7	  |       ^^^^^^^^^^^^^^ 
-lfortranlatest       :     8	
-lfortranlatest       :     9	
-lfortranlatest       :    10	Note: Please report unclear, confusing or incorrect messages as bugs at
-lfortranlatest       :    11	https://github.com/lfortran/lfortran/issues.
-gfortran161          :     2	/app/example.f90:4:13:
-gfortran161          :     3	
-gfortran161          :     4	    4 |    goto 99999
-gfortran161          :     5	      |             1
-gfortran161          :     6	Error: Label 99999 referenced at (1) is never defined
```

certik:
Fortran allows executable END PROGRAM, END SUBROUTINE, END FUNCTION, and END PROCEDURE statements to carry labels and serve as branch targets. LFortran rejected labels before these terminators, causing valid GOTO statements to fail during parsing.

Parse and preserve END labels, then lower them to terminal branch targets immediately before implicit program-unit termination. Handle both explicit program units and implicit main programs without introducing grammar conflicts, and add integration coverage for each executable END form.

Fixes #12516.

### DIFF
--- a/src/lfortran/parser/parser.yy
+++ b/src/lfortran/parser/parser.yy
@@ -14,7 +14,7 @@ see the documentation in that script for details and motivation.
 %param {LCompilers::LFortran::Parser &p}
 %locations
 %glr-parser
-%expect    213 // shift/reduce conflicts
+%expect    193 // shift/reduce conflicts
 %expect-rr 180 // reduce/reduce conflicts
 
 // Uncomment this to get verbose error messages
@@ -414,17 +414,22 @@ void yyerror(YYLTYPE *yyloc, LCompilers::LFortran::Parser &p,
 %type <ast> instantiate_symbol
 %type <ast> enum_decl
 %type <ast> program
-%type <ast> end_program
+%type <end_stmt> end_program
 %type <ast> id_opt
 %type <ast> subroutine
-%type <ast> end_subroutine
+%type <end_stmt> end_subroutine
 %type <ast> procedure
+%type <end_stmt> end_procedure
 %type <ast> sub_or_func
 %type <vec_ast> sub_args
 %type <vec_ast> id_or_star_list
 %type <ast> id_or_star
 %type <ast> function
-%type <ast> end_function
+%type <end_stmt> end_function
+%type <contains_end> program_contains_end
+%type <contains_end> subroutine_contains_end
+%type <contains_end> procedure_contains_end
+%type <contains_end> function_contains_end
 %type <ast> use_statement
 %type <ast> use_statement1
 %type <ast> use_symbol
@@ -537,6 +542,7 @@ void yyerror(YYLTYPE *yyloc, LCompilers::LFortran::Parser &p,
 %type <ast> decl_statement
 %type <vec_ast> statements
 %type <vec_ast> decl_statements
+%type <vec_ast> contains_block
 %type <vec_ast> contains_block_opt
 %type <vec_ast> sub_or_func_plus
 %type <vec_ast> sub_or_func_star
@@ -642,6 +648,9 @@ script_unit
     | statement          %dprec 7
     | expr sep           %dprec 8
     | KW_END_PROGRAM sep { $$ = SYMBOL($1, @$); }
+    | TK_LABEL KW_END sep { $$ = LABELED_END($1, $2, @$); }
+    | TK_LABEL KW_END_PROGRAM sep { $$ = LABELED_END($1, $2, @$); }
+    | TK_LABEL KW_ENDPROGRAM sep { $$ = LABELED_END($1, $2, @$); }
     ;
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


src/lfortran/parser/parser.yy
```diff
@@ -980,92 +1012,105 @@ end_team
...
 
 procedure
     : fn_mod_plus KW_PROCEDURE id sub_args sep decl_statements
-        contains_block_opt
-        end_procedure sep {
-            LLOC(@$, @9); $$ = PROCEDURE($1, $3, $4, TRIVIA($5, $9, @$),
-                $6, $7, @$); }
+        procedure_contains_end sep {
+            LLOC(@$, @8); $$ = PROCEDURE($1, $3, $4, TRIVIA($5, $8, @$),
+                $6, $7->contains, $7->end.label, $7->end.loc, @$); }
```
certik: Not happy with these `->contains` and `->end.label` all over this patch. It leaks the details of implementation.

src/lfortran/parser/parser.yy
```diff
@@ -906,15 +921,23 @@ proc_modifier
...
-    : KW_PROGRAM id sep decl_statements
-        contains_block_opt end_program sep {
-      LLOC(@$, @6); $$ = PROGRAM($2, TRIVIA($3, $7, @$), $4, $5, $6, @$); }
+    : KW_PROGRAM id sep decl_statements program_contains_end sep {
+      LLOC(@$, @5); $$ = PROGRAM($2, TRIVIA($3, $6, @$), $4,
+          $5->contains, $5->end.name, $5->end.label, $5->end.loc, @$); }
+    ;
+
+program_contains_end
+    : end_program { $$ = make_contains_end(p.m_a, $1); }
```
certik: `make_contains_end` should be uppercase, to be consistent with the rest of the file (whether or not it is implemented as a macro).

src/lfortran/parser/parser.yy
```diff
@@ -642,6 +648,15 @@ script_unit
     | statement          %dprec 7
     | expr sep           %dprec 8
     | KW_END_PROGRAM sep { $$ = SYMBOL($1, @$); }
+    | TK_LABEL KW_END sep {
+            RESULT(make_Continue_t(p.m_a, @$, $1, nullptr));
```
certik: Should be a macro, simplify the action.

certik:
Let's see if tests pass. Some more small refactoring and cleaning up is needed, but it looks good, it seems like the right fix.

certik:
Rebased onto `main` and addressed the three review comments.

**"Should be a macro, simplify the action"** (`script_unit`) — the three
labeled-END alternatives now read:

```
| TK_LABEL KW_END sep         { $$ = LABELED_END($1, $2, @$); }
| TK_LABEL KW_END_PROGRAM sep { $$ = LABELED_END($1, $2, @$); }
| TK_LABEL KW_ENDPROGRAM sep  { $$ = LABELED_END($1, $2, @$); }
```

`LABELED_END` is an expression rather than a statement, because bison only
substitutes `$$` inside an action in `parser.yy`, not inside a macro body in
`semantics.h`.

**"`make_contains_end` should be uppercase"** — renamed to `CONTAINS_END`, and
`make_end_stmt` to `END_STMT` for the same reason. Both stay inline functions;
only the naming changes, matching `EXPR`, `CONTAINS`, `TRIVIA` and the rest of
the file.

**"Not happy with these `->contains` and `->end.label` all over this patch"** —
the grammar actions now hand over the whole value and let the macro unpack it:

```
$$ = PROGRAM($2, TRIVIA($3, $6, @$), $4, $5, @$);
```

was

```
$$ = PROGRAM($2, TRIVIA($3, $6, @$), $4,
    $5->contains, $5->end.name, $5->end.label, $5->end.loc, @$);
```

Fourteen actions changed this way; `grep -E '\->contains|\->end\.'` over
`parser.yy` is now empty. The layout is read only through `CE_CONTAINS`,
`CE_END`, `END_NAME`, `END_LABEL` and `END_LOC` in `semantics.h`.

Those five accessors are not just tidiness: the field names have to live
somewhere no macro parameter can capture them. Writing `CE_END(ce).name`
directly inside `PROGRAM(name, ...)` silently expands `.name` into the
`name` argument, which is exactly the kind of breakage this indirection
prevents.

## Verification

`bison -Wall` regenerates with the declared conflict counts unchanged, and
`ci/grammar_conflicts.sh` still reports the patched grammar as LALR(1) with no
essential conflicts (checked locally in a clean worktree; `ci/parser.yy.patch`
still applies).

`ctest` 13/13, `./run_tests.py` passes, `integration_tests -j16` passes,
`goto_07` and `goto_08` build and run.

src/lfortran/parser/semantics.h
```diff
@@ -1702,12 +1704,56 @@ Vec<ast_t*> empty_sync(Allocator &al) {
...
+// are parsed as one symbol so that a label on the END does not have to be
+// distinguished from a label starting a statement in the contains block.
+ContainsEnd* CONTAINS_END(Allocator &al, EndStmt end) {
+    ContainsEnd *result = al.make_new<ContainsEnd>();
+    result->contains.reserve(al, 0);
+    result->end = end;
+    return result;
+}
+
+ContainsEnd* CONTAINS_END(Allocator &al, Vec<ast_t*> contains, EndStmt end) {
```
certik: Let's name these two functions differently like `CONTAINS_END1` and `CONTAINS_END2` to make the dispatch explicit, like we do for other macros.

certik: Done — split into `CONTAINS_END1` (no contains block) and `CONTAINS_END2` (with one), and the eight call sites in `parser.yy` now name which they mean instead of relying on overload resolution.

certik:
It looks good now.

PR #12537, opened 2026-08-18
fix: reject pointer assignment to non-target variables

### LINKED ISSUE #12534: Put back two ASR test files
> Find for now, but after everything is merged, these two ASR files should either compiler, or give an error, so we need to put them back. They discovered a real issue. 

 _Originally posted by @certik in [#12526](https://github.com/lfortran/lfortran/pull/12526/changes/cb15e86d78601f035301e1439598a58771be902d#r3804958484)_

ambikeesshh:
fixes #12534

p1 => plain, where plain is neither a pointer nor a target, is invalid Fortran but lfortran accepted it silently. the only check was the associate.value_is_target verify rule dropped in #12526, which never ran in the compiler.

the check now lives in visit_Associate in ast_body_visitor.cpp. associate constructs are unaffected since the syntax is distinct there. procedure pointers and null() are exempt.

covered by pointer_18.f90 (valid forms) and continue_compilation_1.f90 (rejections).

$ lfortran t.f90 with: integer, pointer :: p1; integer :: plain; p1 => plain
semantic error: Pointer assignment target 'plain' is neither a pointer nor a target


ambikeesshh:
@certik PTAL

ambikeesshh:
@opixdown PTAL

PR #12539, opened 2026-08-19
Accept a statement label on end-if (R1138)

DisciplinedSoftware:
## Summary

- `end if`/`endif` had no labeled grammar alternative, so a statement label on the closing line of an if-construct (e.g. `86 endif`) was rejected with `syntax error: Newline is unexpected here`, even though `end-if-stmt` (R1138) is listed as a branch target statement in J3/18-007r1 (F2018) 11.2.1, and both gfortran and ifx accept it.
- Mirrors the existing `enddo : TK_LABEL KW_END_DO` pattern: `endif`'s grammar rule gains `TK_LABEL` alternatives, and the captured label is attached onto the already-built `If_t` AST node.
- To keep the grammar free of essential conflicts (`ci/grammar_conflicts.sh`), `endif` is consumed inside the `if_block`/`elseif_block` productions rather than once in `if_statement`: with a trailing `endif` after `if_block`, the parser would have to reduce `if_block` before shifting the label's `TK_LABEL`, which is an essential shift/reduce conflict against a labeled statement continuing the body. Inlining it — exactly the `do_statement`/`enddo` shape — lets the parser shift `TK_LABEL` and decide one token later. In an else/else-if chain the shared `endif` is consumed by the innermost production and its label hoisted onto the outermost `If_t`. This also removes the four pre-existing `endif` shift/reduce conflicts from the GLR grammar (`%expect` 197 → 189), and `If_t`'s location now spans through `end if`, as `DoLoop_t`'s already spans through `end do`.
- `If` gains an `end_label` field (analogous to `DoLoop`'s `do_label`), and AST→ASR appends a `GoToTarget` for a nonzero `end_label` so `GO TO 86` into the label resolves correctly — same mechanism already used for `do_label`.
- `ast_to_src.cpp` re-emits the label when unparsing, mirroring how `do_label` round-trips.

Found while converting MODFLOW-2005 to free form (`lmt8.f:306`, `86 ENDIF`, branched to via `GO TO 86`).

## Test plan

- [x] Added `tests/if3.f90` (labeled end-if reached via `GO TO`, plus an else-if chain closed by a labeled end-if to cover the hoist path), with `ast`, `ast_f90`, and `asr` checks
- [x] Added `tests/errors/duplicate_end_if_label1.f90` (duplicate label involving an end-if label is diagnosed)
- [x] Updated the existing `if1`, `if2`, and `fixed_form/end_label` references, whose `If` node picked up the new (always-zero-unless-labeled) field
- [x] `ci/grammar_conflicts.sh`'s check passes locally: the patched (keyword-free, LALR) grammar builds with zero conflicts
- [x] Full `run_tests.py --no-llvm` passes with a frontend-only build; the only additional reference change is the `duplicate_end_if_label1` diagnostic span, which now extends through the labeled `end if` line


src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1496,6 +1496,9 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     AST::If_t* s = AST::down_cast<AST::If_t>(stmt);
                     collect_labels_in_stmts(s->m_body, s->n_body, collect_labels_in_stmt_ref);
                     collect_labels_in_stmts(s->m_orelse, s->n_orelse, collect_labels_in_stmt_ref);
+                    if (s->m_end_label != 0) {
+                        labels.insert(std::to_string(s->m_end_label));
+                    }
```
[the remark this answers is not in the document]
DisciplinedSoftware: Fixed in 8c31263 — the `end_label` insert now emits the same "duplicate statement label" diagnostic as ordinary labels, with a regression test (`tests/errors/duplicate_end_if_label1.f90`). The analogous `do_label` insert is intentionally left unchecked: legacy nested DO loops may legitimately share one terminal statement label, so a duplicate there is not an error.

tests/errors/duplicate_end_if_label1.f90
```diff
@@ -0,0 +1,11 @@
...
+implicit none
+integer :: i
+i = 0
+86 continue
+i = i + 1
+if (i < 3) then
+   go to 86
+86 end if
+print *, i
+end program
```
certik: Please move this into continue compilation

tests/if3.f90
```diff
@@ -0,0 +1,17 @@
...
+86  end if
+    if (i == 0) then
+        i = 10
+    else if (i == 2) then
+        go to 92
+        i = 20
+    else
+        i = 30
+92  end if
+end program
```
certik, line 18: Make this a regular integration test, to ensure it works. Then reference it from here.

src/lfortran/parser/parser.yy
```diff
@@ -1986,28 +1987,38 @@ if_statement_single
             $$ = IFARITHMETIC($3, INTEGER3($5), INTEGER3($7), INTEGER3($9), @$); }
     ;
 
+// `endif` is consumed inside if_block/elseif_block (rather than once in
+// if_statement) so that its optional statement label (`86 end if`) is
+// shifted in the same production as `statements`: the parser can then
+// postpone the statement-vs-endif decision past TK_LABEL, exactly like
+// `enddo` in do_statement. A trailing `endif` after `if_block` would need
+// the block reduced before the label is seen, an essential s/r conflict.
```
certik: Remove the comments. You can put it into docs/, but not here.

src/lfortran/semantics/ast_body_visitor.cpp
```diff
@@ -1504,6 +1504,19 @@ class BodyVisitor : public CommonVisitor<BodyVisitor> {
                     AST::If_t* s = AST::down_cast<AST::If_t>(stmt);
                     collect_labels_in_stmts(s->m_body, s->n_body, collect_labels_in_stmt_ref);
                     collect_labels_in_stmts(s->m_orelse, s->n_orelse, collect_labels_in_stmt_ref);
+                    if (s->m_end_label != 0) {
+                        std::string end_label_name = std::to_string(s->m_end_label);
+                        if (!labels.insert(end_label_name).second) {
```
certik, line 1513: Why are labels being indexed as strings? I thought they can only be numbers.

certik:
I left some comments. I canceled the CI, as I need it for something else right now.

tests/errors/duplicate_end_if_label1.f90
DisciplinedSoftware: Done in 4a9cc2279: the standalone `errors/duplicate_end_if_label1` reference test is gone, and the case (`86 continue` followed by `86 end if`) is a subroutine appended to `errors/continue_compilation_2.f90`. Its reference output gains exactly that one "duplicate statement label 86" diagnostic; nothing earlier in the file moves.

tests/if3.f90
DisciplinedSoftware: Done in 4d6caf433: `integration_tests/if_08.f90` (`LABELS gfortran llvm`) runs the three shapes — a plain if block, an if/else with the branch in the else part, and an if/else-if/else chain where the label is hoisted from the inner block — and asserts the values after each `go to` to a labeled `end if`. On main the file fails to parse at the first labeled `end if`; here it passes under both backends locally. `tests/if3.f90` (the AST reference test for the new `end_label` field) now carries a comment pointing at it; its references were regenerated for that comment only.

src/lfortran/parser/parser.yy
DisciplinedSoftware: Removed in 32c2f8fa4. I did not add anything under docs/; the rationale now lives only in that commit's message.

src/lfortran/semantics/ast_body_visitor.cpp
DisciplinedSoftware: They follow the existing container: `labels` is the pre-existing `std::set<std::string>` member of the body visitor, into which the ordinary statement labels are inserted as `std::to_string(label)` just above this hunk, and into which `visit_Assign` also inserts variable names (`ASSIGN 10 TO ivar`) — that mix is why it is a set of strings. The `GO TO` / `err=` / `end=` checks all look up with `labels.find(std::to_string(...))`. The new `end_label` insert mirrors the ordinary-label one; changing the container to integers would be a separate change that also touches the ASSIGN path.

PR #12540, opened 2026-08-19
fixed-form: recognize % component access in assignment statements

DisciplinedSoftware:
## Summary

- `is_possible_assignment()` (the fixed-form statement classifier that decides "is this line an assignment") only understood array-index parentheses after the leading name, e.g. `X(1)`. It had no notion of derived-type component access (`%`), so any assignment whose target contains a `%` — `D%X = 5`, `D%X => I`, `D(1)%X(2)%Y = ...` — derailed the classifier: it returned false, the caller's dispatch loop stopped consuming statements one line early, and the enclosing construct's own `END`/`END SUBROUTINE`/etc. detection then ran against the *next* unrecognized line, producing a misleading `Expecting terminating symbol for subroutine` pointing at the wrong place entirely.
- This affects plain assignment and pointer assignment identically — the bug is `%`, not `=>`; `is_possible_assignment` only needs to decide "assignment-shaped or not", the `=` vs `=>` distinction is made later by the shared token-level lexer.
- Fix: extend the existing index-parsing loop to also consume `%name` segments, interleaved with `(...)` groups in any order/depth (so `D(1)%X(2)%Y` works too), before checking for the trailing `=`. Only changes behavior for lines with a top-level `%` in the assignment target; every other line takes the exact same path as before.

Found in MODFLOW-2005 `de47.f:986` (`DE4DAT(IGRID)%MXUP=>MXUP`, the grid-pointer save/point pattern used across the codebase — this blocks most of its 51 fixed-form files). The identical code already parses fine as free form. gfortran accepts the file with `-ffixed-form`.

## Test plan

- [x] Added `tests/fixed_form/ptr_assign1.f` (`D%X=>I` through a `USE`d module), `ast` check
- [x] Verified regression-free against `fixed_form/assign_to1-3`, `fixed_form/do1`
- [x] Verified with a frontend-only build (no LLVM) in this environment; a full `run_tests.py -u` against the complete CI matrix should be run before merge


certik:
Must be squashed merged, as I applied a change from copilot.

DisciplinedSoftware:
Superseded by and merged as #12547.

Issue #12497, opened 2026-08-13
ICE (tokenizer error) in --fixed-form: Type...end Type block followed by contains


dan-distributive:
## Summary

In `--fixed-form` mode, a `module` containing a `Type ... end Type` block
followed by `contains` (with at least one procedure) fails to tokenize, with
an internal compiler error. The identical source, parsed as free-form
(dropping `--fixed-form`), compiles successfully. Also reproduces via
`lfortran fmt --fixed-form` (same underlying tokenizer path).

This blocks essentially any real-world fixed-form (`.f`) codebase that mixes
modern derived types with module procedures -- a very common pattern in
legacy scientific Fortran being incrementally modernized in place (my case:
a ~64-file astronomy pipeline where the large majority of files hit this).

## Environment

- LFortran version: 0.64.0 (alpha)
- Installed via: conda-forge (micromamba)
- LLVM: 19.1.1
- Platform: macOS ARM (arm64-apple-darwin), Darwin 25.2.0

## Minimal Reproducible Example

`mre_fixed_form_type_contains.f`:

```fortran
      module MinTypeMod
      implicit none

      Type Beam2D
        real BeamMajorAxis
      end Type

      contains

      subroutine dummy()
      end subroutine

      end module
```

## Command and actual output

```
$ lfortran --implicit-interface --fixed-form -c mre_fixed_form_type_contains.f -o /tmp/out.o
```

```
tokenizer error: ICE: Cannot recognize global scope entity
 --> mre_fixed_form_type_contains.f:8:7
  |
8 |       contains
  |       ^

Note: Please report unclear, confusing or incorrect messages as bugs at
https://github.com/lfortran/lfortran/issues.
```

Same ICE reproduces via `lfortran --fixed-form fmt mre_fixed_form_type_contains.f`.

## Expected behavior

Successful compilation (or at worst a clear semantic/syntax diagnostic, not
an internal "ICE: Cannot recognize global scope entity").

## Confirming this is fixed-form-specific

The byte-identical structure, adapted to free-form syntax, compiles cleanly:

```fortran
module MinTypeMod
  implicit none

  Type Beam2D
    real :: BeamMajorAxis
  end Type

contains

  subroutine dummy()
  end subroutine

end module
```

```
$ lfortran --implicit-interface -c mre_free_form_type_contains.f90 -o /tmp/out.o
# exits 0, no errors
```

## Additional notes

- Removing every field from the `Type` block down to a single scalar `real`
[...]

PR #12541, opened 2026-08-19
fixed-form: recognize the bare TYPE name derived-type-def opener

DisciplinedSoftware:
## Summary

- `lex_body_statement()`'s derived-type-block detector only matched `TYPE ::` (the attribute-list form); the equally valid bare form `TYPE name` (no `::`, no attributes) fell through to `lex_declaration` unrecognized. The type-name line, its component line(s), and the `END TYPE` line then got scanned as three unrelated statements instead of one coherent block. Fixed-form's prefix-based `next_is()` then let `END TYPE name` (stripped to e.g. `endtypegt`) spuriously match the `end`-prefixed fallback meant for the enclosing construct's own terminator, so the block closed one statement early — `END TYPE` was consumed as if it were `END MODULE`/`END SUBROUTINE`, leaving whatever followed orphaned at global scope, hence `ICE: Cannot recognize global scope entity` pointing at that next line rather than at `END TYPE` itself.
- Fix: also route to `lex_derived_type()` when `TYPE` is immediately followed by a bare name and then end-of-line — i.e. exactly a derived-type-def statement with nothing else on the line. This is carefully disambiguated from `TYPE(...)` (a declaration-type-spec using a derived type, e.g. `TYPE(GT), SAVE :: DAT(10)`, which must NOT be routed here) and from a plain identifier that merely starts with "type" (e.g. `TYPEX = 5`, where a `=` follows the name instead of end-of-line). `TYPE, EXTENDS(parent) :: name` (attribute-list form) is out of scope for this fix, same as it already was for the pre-existing `type::` check.
- Same underlying gap as the currently-open #12497 ("ICE (tokenizer error) in --fixed-form: Type...end Type block followed by contains") — this reproduces the ICE without `contains`, from an ordinary declaration right after `END TYPE`.

Found in MODFLOW-2005 `gwf2gag7.f:9` (`TYPE(GWFGAGTYPE), SAVE:: GWFGAGDAT(10)`). The identical code parses fine as free form. gfortran accepts the file with `-ffixed-form`.

## Test plan

- [x] Added `integration_tests/fixed_form_type_save_01.f90` (labels `gfortran llvm`, `--fixed-form`): bare `TYPE GT` opener with a trailing comment, `TYPE(GT), SAVE ::` after `END TYPE`, and a `TYPEX` identifier that merely starts with "type", checked at runtime against gfortran
- [x] Verified `fixed_form_interface.f` (uses `TYPE :: element`, the pre-existing `::` form) still passes unchanged
- [x] CI green across the full matrix


src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1129,6 +1129,25 @@ struct FixedFormRecursiveDescent {
...
+        // A bare derived-type-def statement is exactly `TYPE type-name`
+        // with nothing else on the line, so require a NAME immediately
+        // after `type` followed immediately by end-of-line.
+        if (next_is(cur, "type") && !next_is(cur, "type(")) {
+            unsigned char *name_end = cur + 4;
+            if (try_name(name_end) && next_is_eol(name_end)) {
+                lex_derived_type(cur);
+                return true;
+            }
+        }
```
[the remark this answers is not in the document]
DisciplinedSoftware: This scenario cannot occur: trailing `!` comments never reach the fixed-form recursive-descent tokenizer. The fixed-form prescanner strips them during `copy_rest_of_line()` in `parser.cpp` (the `s[pos] == '!'` branch skips the rest of the line and emits `'\n'`), so by the time `lex_body_statement()` runs, `TYPE GT ! comment` is already `typegt\n` and `next_is_eol(name_end)` holds.

Verified empirically on this branch: a module with `TYPE GT ! trailing comment` on the opener line parses to the correct `DerivedType` AST. Locked in by cce0e3a, which adds a trailing comment to the opener line of `tests/fixed_form/type_save1.f` — the reference AST output is byte-identical.

DisciplinedSoftware:
Added a runtime integration test in 80dbc510: `integration_tests/fixed_form_type_save_01.f90` (labels gfortran + llvm, `--fixed-form`). It exercises the bare `TYPE GT` opener with a trailing comment, the `TYPE(GT), SAVE ::` declaration after `END TYPE`, and a `TYPEX` identifier that merely starts with "type", then checks the values at runtime. Verified locally with gfortran `-ffixed-form` and through lfortran AST/ASR (frontend-only build here, so the LLVM run is on CI).

tests/fixed_form/type_save1.f
```diff
@@ -0,0 +1,6 @@
+      MODULE G
+        TYPE GT ! trailing comment must not defeat opener detection
+          INTEGER N
+        END TYPE GT
+        TYPE(GT), SAVE:: DAT(10)
+      END MODULE G
```
certik, line 6: What is the point of this test --- aren't we already testing this in the integration test in this PR?

DisciplinedSoftware: The AST test (ast = true, no backend requirement) runs in the CI job that builds without LLVM, where the integration test does not; it is registered with LABELS gfortran llvm in integration_tests/CMakeLists.txt, so it is skipped there. Without the AST test, that job would have no coverage of this tokenizer fix at all. They also check different things: the AST test pins the parser output shape directly, while the integration test checks runtime behavior via gfortran comparison. Happy to drop it if you would rather simplify, but I think it is earning its keep here.

DisciplinedSoftware:
@certik Friendly follow-up: the requested gfortran-checked integration test is included as `integration_tests/fixed_form_type_save_01.f90`, and CI is green. I also replied to the remaining thread explaining why the small AST test provides coverage in the no-LLVM CI job where the integration test is skipped. Could you take another look and merge if that coverage split is acceptable?

tests/fixed_form/type_save1.f
certik: I don't understand. This job right here in integration test is exercising the same tokenizer:
```
RUN(NAME fixed_form_type_save_01 LABELS gfortran llvm EXTRA_ARGS --fixed-form GFORTRAN_ARGS -ffixed-form)
```
This is testing the AST.

certik: So remove this file.

tests/tests.toml
```diff
@@ -3651,6 +3651,10 @@ filename = "fixed_form/ptr_assign1.f"
 ast = true
 options = "--implicit-typing --implicit-interface --implicit-argument-casting"
 
+[[test]]
+filename = "fixed_form/type_save1.f"
+ast = true
```
certik: Remove.

tests/fixed_form/type_save1.f
DisciplinedSoftware: Removed in 1ba7268ae: `tests/fixed_form/type_save1.f`, its two reference files and the `tests.toml` stanza. Only the integration test remains.

certik:
Thanks!

PR #12542, opened 2026-08-19
fixed-form: error instead of hanging on an unrecognized DO WHILE body statement

DisciplinedSoftware:
## Summary

- `lex_dowhile()` ignored `lex_body_statement()`'s return value, so an unrecognized statement inside a fixed-form `DO WHILE` body (named or not) silently spun forever instead of erroring — unlike the plain-DO and labeled-DO terminator loops, which already check it. It also never passed `continue_compilation` down, so the same input hung under `--continue-compilation` as well, where every other statement in the program body recovers by tokenizing the line and letting the parser report it.
- Both are fixed: the body loop checks the return value and raises `Expected an executable statement inside a do while loop` (what an end of file or a stray `end`/`contains`/`subroutine`/`function` inside the loop now hits instead of hanging), and `continue_compilation` is threaded through so that under `--continue-compilation` an unrecognized body statement takes the regular line-level recovery and the rest of the translation unit is still compiled.

Originally found in MODFLOW-2005 `gwf2huf7.f:567` (`UPLAY: DO WHILE ( KK.LE.NLAY )`). Named `DO WHILE` recognition itself was fixed independently in #12572; this is the remaining piece, rebased on top of it.

## Test plan

- [x] Regression case appended to `tests/errors/continue_compilation_ff_1.f` (a `DO WHILE` whose body holds a malformed declaration), placed after the last existing statement so no earlier diagnostic moves. On `main` this file now hangs; with this change its reference gains exactly one syntax error (line 122) and the ASR (`.stdout`) is byte-identical.
- [x] Verified the bare reproducer: non-cc → tokenizer error, exit; cc → syntax error, ASR of the rest of the program produced; unterminated loop (`END PROGRAM` inside the body) → tokenizer error, no hang. All three hang on `main`.
- [x] All `tests/errors/` + `tests/fixed_form/` reference tests and every `.f` test in `tests.toml` pass locally (frontend build; LLVM-gated ones via CI).
- [x] `fixed_form/while_02_fixed_form.f` (existing unnamed `DO WHILE`) unchanged.


src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1652,11 +1671,21 @@ struct FixedFormRecursiveDescent {
...
+                    diag::Level::Error, diag::Stage::Tokenizer, {diag::Label("", {loc})}));
+                throw parser_local::TokenizerAbort();
+            };
         }
-        push_token_advance(cur, "enddo");
-        tokenize_line(cur);
+        // Consumes "enddo", an optional trailing construct name (e.g. the
+        // `UPLAY` in `END DO UPLAY`), and the rest of the line -- same
+        // terminator handling `try_enddo_regular` uses for a plain DO.
+        lex_enddo_line(cur);
```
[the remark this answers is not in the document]
DisciplinedSoftware: This matches the existing plain-DO terminator behavior on purpose: `try_enddo_regular` uses the same `next_is(cur, "enddo")` prefix check and carries a pre-existing `// TODO: parse things correctly to distinguish enddo = 5;` for exactly this ambiguity, and `lex_do_concurrent`/`lex_do_terminal` do the same. A token-boundary check isn't implementable in this stream: the fixed-form prescanner strips whitespace, so `END DO UPLAY` arrives as `enddouplay` — `enddo` followed by an identifier character is precisely the named-terminator case this PR fixes, indistinguishable from `enddoxyz = 5` without real parsing. The old `enddo\n` check traded this false positive for never matching a named terminator at all. Resolving `enddo=5` for all DO forms is left to the existing TODO.

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1652,11 +1671,21 @@ struct FixedFormRecursiveDescent {
...
+        while (!next_is(cur, "enddo")) {
+            if (!lex_body_statement(cur)) {
+                Location loc;
+                loc.first = cur-string_start;
+                loc.last = cur-string_start;
+                diag.add(diag::Diagnostic(
+                    "Expected an executable statement inside a do while loop",
+                    diag::Level::Error, diag::Stage::Tokenizer, {diag::Label("", {loc})}));
+                throw parser_local::TokenizerAbort();
+            };
```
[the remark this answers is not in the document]
DisciplinedSoftware: Removed in 43707b6f. (For the record, the `};` was copied verbatim from `lex_do_regular`'s identical error block, which still has one — left untouched here to keep the diff minimal.)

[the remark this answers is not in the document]
DisciplinedSoftware: Added in 43707b6f: `tests/errors/fixed_form_7.f` puts an unrecognized statement inside a fixed-form `DO WHILE` body and asserts the new "Expected an executable statement inside a do while loop" tokenizer error (the pre-fix code looped forever on this input).

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1525,6 +1532,18 @@ struct FixedFormRecursiveDescent {
...
+    // it would misparse `while` as the do-variable name, since "while" is a
+    // syntactically valid identifier.
+    bool is_named_do_while_loop(unsigned char *cur, unsigned char *&do_pos) {
+        unsigned char *tmp = cur;
+        if (!try_name(tmp)) return false;
+        if (!try_next(tmp, ":")) return false;
+        if (!next_is(tmp, "dowhile(")) return false;
+        do_pos = tmp;
+        return true;
+    }
```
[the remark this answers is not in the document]
DisciplinedSoftware: Agreed in principle, but the overlap is two lines (`try_name` + `try_next(tmp, ":")`), and `is_named_do_loop` needs the post-colon pointer anyway, so a helper would save little today. Happy to factor out a `try_construct_name_prefix()` when a third named-construct recognizer appears; keeping this fix minimal for now.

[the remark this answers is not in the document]
DisciplinedSoftware: Done after all in 83f90544: `try_construct_name_prefix()` now owns the `name:` parsing, following the same advance-on-success/leave-untouched-on-failure convention as the other `try_*()` helpers, and both `is_named_do_loop()` and `is_named_do_while_loop()` are built on it. (Considered instead merging the two predicates into one with a construct-kind out-param, or teaching `is_do_loop()` about `DO WHILE` — but `is_do_loop()` is also called bare for the unnamed case where `lex_do` can't handle a WHILE, and the merged predicate would break the file's one-`is_*`-per-construct pattern, so the small shared helper is the cleanest of the three.)

tests/tests.toml
```diff
@@ -3651,6 +3655,11 @@ filename = "fixed_form/ptr_assign1.f"
 ast = true
 options = "--implicit-typing --implicit-interface --implicit-argument-casting"
 
+[[test]]
+filename = "fixed_form/named_do_while1.f"
+ast = true
+options = "--implicit-typing --implicit-interface --implicit-argument-casting"
```
certik: Why do we need these options? We should not be structuring the test to require those.

tests/fixed_form/named_do_while1.f
```diff
@@ -0,0 +1,6 @@
+      SUBROUTINE H(N)
+      K=1
+      UPLAY: DO WHILE ( K.LE.N )
+        K=K+1
+      END DO UPLAY
+      END
```
certik: This should be a regular integration test, and test it with gfortran, to ensure the syntax is correct.

certik:
I feel this PR is very complex. I am testing a simpler solution in https://github.com/lfortran/lfortran/pull/12572.

tests/tests.toml
DisciplinedSoftware: Those options were only needed for `named_do_while1.f`'s implicit-typing repro (`K` had no declaration). Since #12572 already added a proper integration test for the positive case (`fixed_form_do_while_name_01.f90`, checked against gfortran), I've dropped `named_do_while1.f` and this tests.toml entry entirely rather than fix the options — no need for a duplicate AST-only test.

tests/fixed_form/named_do_while1.f
DisciplinedSoftware: Agreed, and #12572's `fixed_form_do_while_name_01.f90` already is that gfortran-checked integration test for the positive case. I've dropped `named_do_while1.f` from this PR rather than convert it — the only test left here is `tests/errors/fixed_form_7.f`, a negative/error test asserting a clean tokenizer error (not a hang) on an unrecognized statement inside a DO WHILE body, following the existing `tests/errors` convention rather than the gfortran-checked `integration_tests` one.

DisciplinedSoftware:
Agreed — #12572 covers the actual bug I hit in MODFLOW-2005 (`gwf2huf7.f:567`) and is simpler, so I'm glad it's in. The one thing #12572 doesn't cover is that `lex_dowhile()` still ignores `lex_body_statement()`'s return value on `main`: an unrecognized statement inside a DO WHILE body (named or not) still hangs the tokenizer instead of erroring, unlike the plain-DO and labeled-DO paths, which already check it.

I've trimmed this PR down to just that guard + its regression test (`tests/errors/fixed_form_7.f`), rebased on top of #12572, and dropped the now-redundant named-DO-WHILE recognition code and the `named_do_while1.f` test (replied on those threads too). Should be a much smaller, single-purpose diff now.

tests/errors/fixed_form_7.f
```diff
@@ -0,0 +1,6 @@
+      SUBROUTINE H(N)
+      K=1
+      DO WHILE ( K.LE.N )
+        XINTEGER :: Y
+      END DO
+      END
```
certik: Can this be moved into `tests/errors/continue_compilation_ff_1.f` ?

certik (commented):
I think this is good, thanks for the error check. After moving it, we can merge it I think.

tests/errors/fixed_form_7.f
DisciplinedSoftware: Moved, but into a new `tests/errors/continue_compilation_ff_3.f` rather than folding it into `continue_compilation_ff_1.f`.

Reason: this error isn't recoverable under `--continue-compilation` the way the errors already in `ff_1.f` are. I tried splicing it in, and the whole file's ASR output collapsed to an empty translation unit. The tokenizer error aborts the entire parse instead of being skipped, unlike, e.g., the broken nested `SUBROUTINE` in `ff_2.f`, which the recovery machinery drops cleanly while keeping the rest. So putting it in `ff_1.f` would have silently deleted that file's coverage of ~20 other error cases.

`continue_compilation_ff_3.f` follows the `ff_2.f` pattern (own small file, `fixed_form_cc_asr = true`) and keeps the test meaningful. Happy to instead have `lex_dowhile()`'s error path participate in continue-compilation recovery, so it truly could live in `ff_1.f` — but that adds more surface area to an already-approved, CI-green PR, so I'd rather do it as a follow-up PR if you want it. Let me know which you'd prefer.

DisciplinedSoftware:
@certik Friendly follow-up: the requested error test was moved in 83be606f to tests/errors/continue_compilation_ff_3.f. I kept it separate from continue_compilation_ff_1.f because this tokenizer error aborts parsing, which would otherwise erase the existing recovery coverage in that file. CI is green and the PR is mergeable. Could you confirm whether the separate ff_3 test is acceptable and merge if so?

tests/reference/asr-continue_compilation_ff_3-6bc4cf3.stderr
```diff
@@ -0,0 +1,5 @@
+tokenizer error: Expected an executable statement inside a do while loop
+ --> tests/errors/continue_compilation_ff_3.f:4:9
+  |
+4 |         XINTEGER :: Y
+  |         ^
```
certik: Add this to the already existing continue compilation for fixed form.

tests/reference/asr-continue_compilation_ff_3-6bc4cf3.stdout
```diff
@@ -0,0 +1,8 @@
+(TranslationUnit
+    (SymbolTable
+        1
+        {
+            
+        })
+    []
+)
```
certik: I think this is not needed.

tests/reference/asr-continue_compilation_ff_3-6bc4cf3.stderr
DisciplinedSoftware: Done in ef0d8058e: the case now lives in `tests/errors/continue_compilation_ff_1.f` (three lines appended after the last existing statement, so no earlier diagnostic moves) and `continue_compilation_ff_3.f` with its references is gone.

To make that possible, `continue_compilation` is now threaded through `lex_dowhile()` into `lex_body_statement()`, so under `--continue-compilation` an unrecognized DO WHILE body statement takes the same line-level recovery every other program-body statement already has (tokenize the line, let the parser report it) instead of aborting the whole unit. `ff_1`'s reference gains exactly one syntax error at line 122 and its ASR is unchanged. The tokenizer guard stays for the non-cc path and for end-of-file / a stray `END` inside the loop (which used to hang too).

tests/reference/asr-continue_compilation_ff_3-6bc4cf3.stdout
DisciplinedSoftware: Agreed, gone: with the recovery in place there is no empty-translation-unit output any more, and `ff_3` (file, references, `tests.toml` entry) is removed entirely.

tests/errors/fixed_form_7.f
DisciplinedSoftware: Superseded: the error path now participates in continue-compilation recovery (see the thread on the `.stderr` reference), so the case is in `ff_1.f` after all and `ff_3.f` is removed.

src/lfortran/parser/fixedform_tokenizer.cpp
```diff
@@ -1691,14 +1691,26 @@ struct FixedFormRecursiveDescent {
...
+            // error below is then only reached at end of file or at a
+            // stray `end`/`contains`/`subroutine`/`function`.
+            if (!lex_body_statement(cur, continue_compilation)) {
+                Location loc;
+                loc.first = cur-string_start;
+                loc.last = cur-string_start;
+                diag.add(diag::Diagnostic(
+                    "Expected an executable statement inside a do while loop",
+                    diag::Level::Error, diag::Stage::Tokenizer, {diag::Label("", {loc})}));
+                throw parser_local::TokenizerAbort();
```
certik, line 1712: Please add a test for this new error message.

certik:
Thanks!

src/lfortran/parser/fixedform_tokenizer.cpp
certik: https://github.com/lfortran/lfortran/issues/12985

PR #12544, opened 2026-08-19
fix: ICE on EQUIVALENCE between two COMMON block variables

### LINKED ISSUE #12524: ICE on EQUIVALENCE between two COMMON block variables
Hello,

This aborts on an assertion instead of giving a diagnostic. gfortran accepts it.

    subroutine s
    real :: a, b
    common /c1/ a, b
    equivalence (a, b)
    end subroutine

    LCOMPILERS_ASSERT failed: src/libasr/asr.h
    function down_cast(), line number 41 at is_a<T>(*f)

`visit_DeclarationUtil` branches on `eq1_is_common` / `eq2_is_common` and then
assumes the other operand is a plain `ASR::Var_t`. That holds when one side is a
local, not when both are in the block. Two unguarded casts, `ast_common_visitor.h:6821`
for scalars and `:6282` for array elements. The second is what JPL's DIVA hits.

Guarding both with `is_a` and falling through to the existing "not implemented yet"
warning nearby would at least make it a diagnostic.

0.64.0-74-g02efa6ac5. Related to #692 and #661.

opixdown:
Fixes #12524

### DIFF
--- a/tests/reference/asr-continue_compilation_1-04b6d40.stderr
+++ b/tests/reference/asr-continue_compilation_1-04b6d40.stderr
@@ -538,11 +538,15 @@ semantic error: equivalence between a common block variable and this array eleme
 949 |         equivalence (first, alias(1))  ! {Error} equivalence between a common block variable and this array element is not implemented
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported equivalence
 
-semantic error: equivalence for 'b' conflicts with the ordering of COMMON block 'c1'
+semantic error: cannot make two COMMON variables share the same storage
    --> tests/errors/continue_compilation_1.f90:973:22
     |
 973 |         equivalence (a, b)
     |                      ^  ^ 
+
+'a' and 'b' are both declared in COMMON block 'c1'. The COMMON statement already places them at fixed positions in memory. EQUIVALENCE is not allowed to override that layout.
+
+Hint: you may EQUIVALENCE a non-COMMON variable to a COMMON variable (this extends the common block), but never two variables that are already in a COMMON block.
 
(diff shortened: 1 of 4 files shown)


tests/reference/asr-continue_compilation_1-04b6d40.stderr
```diff
@@ -538,16 +538,22 @@ semantic error: equivalence between a common block variable and this array eleme
 949 |         equivalence (first, alias(1))  ! {Error} equivalence between a common block variable and this array element is not implemented
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported equivalence
 
+semantic error: equivalence for 'b' conflicts with the ordering of COMMON block 'c1'
+   --> tests/errors/continue_compilation_1.f90:973:22
+    |
+973 |         equivalence (a, b)
+    |                      ^  ^
```
certik, line 545: I don't understand this error message as a user.

Let's use the following message, which I think is clear:
```
semantic error: cannot make two COMMON variables share the same storage
   --> tests/errors/continue_compilation_1.f90:973:22
    |
973 |         equivalence (a, b)
    |                      ^  ^
    |
    | 'a' and 'b' are both declared in COMMON block 'c1'.
    | The COMMON statement already places them at fixed positions
    | in memory. EQUIVALENCE is not allowed to override that layout.
    |
    | Hint: you may EQUIVALENCE a non-COMMON variable to a COMMON
    | variable (this extends the common block), but never two
    | variables that are already in a COMMON block.
```

certik (approved):
I think it looks nice now!

Can you please rebase and resolve the conflicts, and then enable exhaustive tests?

PR #12552, opened 2026-08-20
refactor: centralize inline character struct member detection

certik:
## Summary

bind(C) and SEQUENCE derived types lay their non-pointer, non-allocatable
character members out inline in the struct (a flat byte blob at a fixed offset)
instead of behind a string descriptor, so that the struct matches the C/Fortran
storage layout. That condition was hand-rolled in six places across
`asr_to_llvm.cpp` and `llvm_utils.cpp`, each one respelling the same walk from a
`StructInstanceMember` to its owning `Struct`.

This PR is pure refactoring: no behavior change, no new tests.

## Scope

- `src/libasr/asr_utils.h`: add `ASRUtils::is_inline_character_struct_member`
  with two overloads — one taking a `Struct_t` plus a member type, one taking a
  member-access expression.
- `src/libasr/codegen/asr_to_llvm.cpp`, `src/libasr/codegen/llvm_utils.cpp`:
  call it from all six sites (`getStructType`,
  `handle_global_nonallocatable_stringArray`, `get_string_data_and_length`,
  `allocate_array_members_of_struct`, `visit_Assignment`, and the string-format
  argument path).

Two dead/duplicated bits fall out while collapsing the copies:

- `get_string_data_and_length` assigned `phys_type = CChar` for the inline case
  and then returned before the switch that reads it — dropped.
- the string-format path built a string descriptor by hand; it now reuses the
  existing `LLVMUtils::create_string_descriptor`.

## Verification

Local runs on this branch:

- `./run_tests.py --no-llvm` — TESTS PASSED
- `integration_tests/run_tests.py -b llvm` — 100% passed, 0 failed out of 3860
- `integration_tests/run_tests.py -b gfortran` — 100% passed, 0 failed out of 3922
- reference outputs for the tests that touch this code (`common2`, `common_03`
  … `common_13`, `derived_types_32`, `read_83`, `string_03`, `string_11`) are
  byte-identical to the ones produced by `main` with the same compiler — as
  expected for a no-op refactor, and no reference file needed updating.

## Rationale

Extracted from #12309, which added a seventh copy of this predicate. Pulling the
dedup out first keeps the behavior changes that follow readable; per `AGENTS.md`,
refactoring is not mixed with bug fixes.

This is the first of three stacked PRs extracted from the inline-COMMON part of
#12309:

1. **this PR** — the shared helper (no behavior change)
2. #12553 — fix inline character *array* members (bind(C)/SEQUENCE)
3. #12554 — store COMMON block characters inline

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -4302,6 +4302,46 @@ static inline ASR::String_t* get_string_type(ASR::expr_t* s){
 }
 
 /*
+    A character member of a bind(C) or SEQUENCE struct is laid out inline in
+    the struct (a flat byte blob at a fixed offset) instead of behind a string
+    descriptor, so that the struct matches the C/Fortran storage layout.
+    Pointer and allocatable members keep their descriptor.
+*/
+static inline bool is_inline_character_struct_member(
+        ASR::Struct_t* struct_type, ASR::ttype_t* member_type) {
+    return (struct_type->m_abi == ASR::abiType::BindC
+            || struct_type->m_is_sequence)
+        && !ASR::is_a<ASR::Pointer_t>(*member_type)
+        && !ASR::is_a<ASR::Allocatable_t>(*member_type)
+        && ASR::is_a<ASR::String_t>(*type_get_past_array(member_type));
+}
+
+// Same as above, for an expression that may be a struct member access.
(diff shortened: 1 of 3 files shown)


certik:
This is a good change just on the account of `+68 -95`.

It refactors some common logic, so I think this is good to merge if CI passes.

PR #12553, opened 2026-08-20
fix: lay out and index inline character array struct members correctly

certik:
> Rebased onto `main` (71f604b2a) now that #12552 has merged. This PR is a single
> commit and its diff no longer carries #12552's refactor. Merge before #12554.

## Summary

A non-pointer, non-allocatable character member of a bind(C)/SEQUENCE struct is
stored inline in the struct rather than behind a string descriptor, but the
layout only ever reserved one element's worth of bytes: `getStructType` sized the
member as `[len x i8]` and ignored the element count, so an array member
overlapped the members that follow it. The accessors matched that assumption —
indexing, substring and formatted read all went through the descriptor-only
paths — so a program that used such a member crashed:

```fortran
type :: t
    sequence
    character(len=4) :: names(3)
    integer :: n
end type
type(t) :: x
x%names(1) = "abcd"     ! Segmentation fault
```

## Scope

- `src/libasr/asr_utils.h`: add `ASRUtils::inline_character_storage_size`
  (`count * len * kind`).
- `src/libasr/codegen/llvm_utils.cpp`: size the inline member as
  `[count*len*kind x i8]`; add `LLVMUtils::get_inline_string_element`, which
  offsets into the blob by `idx*len*kind` and wraps the element in a string view
  descriptor.
- `src/libasr/codegen/llvm_array_utils.cpp`, `src/libasr/codegen/asr_to_llvm.cpp`:
  use it from `get_single_element` and from the formatted-read element path;
  route `visit_StringItem` and the string-section helper through the inline-aware
  `get_string_data_and_length`; skip descriptor setup for inline members in
  `allocate_array_members_of_struct` (arrays as well as scalars); and stop
  treating an inline member as an allocatable assignment destination — there is
  no data pointer for `_lfortran_strcpy` to allocate.
- Reference outputs updated for the four tests whose LLVM lowering changes
  (`derived_types_32`, `read_83`, `string_03`, `string_11`).

## Verification

New integration test `derived_types_154` (labels `gfortran llvm`). It segfaults
at runtime before this change and passes after; gfortran agrees.

The four `llvm-*` reference outputs were regenerated locally against
**llvmdev 11.1.0**, matching the LLVM version CI's ubuntu reference job uses.
That toolchain was validated first by running the full reference suite on
unmodified `main`, which passed — so it reproduces CI's reference outputs
byte-for-byte. `derived_types_32` and `read_83` needed regenerating after the
rebase because main's 10db928bf touched the same string lowering; `string_03`
and `string_11` were already correct.

Local runs on this branch:

- `./run_tests.py` under LLVM 11 — TESTS PASSED
- integration suites were run on the #12554 tip, which contains this commit:
  `-b llvm` 3874 passed / 0 failed, `-b gfortran` 3936 passed / 0 failed

## Rationale

Extracted from #12309, where this was a prerequisite for storing COMMON block
characters inline. It is a bug in its own right for bind(C)/SEQUENCE types, with
no COMMON involved, so it is reviewable and testable on its own.

Second of the three PRs split out of #12309: #12552 (merged) → **this PR** →
#12554.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2560,6 +2560,7 @@ RUN(NAME derived_types_150 LABELS gfortran llvm) # SEQUENCE char(len>1) inline +
 RUN(NAME derived_types_151 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME derived_types_152 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
 RUN(NAME derived_types_153 LABELS gfortran llvm EXTRA_ARGS --implicit-interface)
+RUN(NAME derived_types_154 LABELS gfortran llvm)
 RUN(NAME defined_assignment_01 LABELS llvm flang)
 RUN(NAME defined_assignment_free_interface_01 LABELS gfortran llvm flang)
 
--- a/integration_tests/derived_types_154.f90
+++ b/integration_tests/derived_types_154.f90
@@ -1,0 +1,29 @@
+! A character array member of a SEQUENCE (or bind(C)) derived type is stored
+! inline in the struct, as a flat blob of count*len bytes, rather than behind
+! a string descriptor. The blob has to be sized by the element count, and
+! indexing it has to step by the element length.
+program derived_types_154
+    implicit none
+    type :: t
+        sequence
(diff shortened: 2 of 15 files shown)


certik:
I think that this is good, it fixes a string descriptor.

PR #12555, opened 2026-08-20
simplify: drop the team_type import only if unused

certik:
FORM TEAM already converts the used variable. An unused team_type entity still names the import, so drop it only when nothing else does, instead of rewriting every team_type variable in the translation unit.

Follow-up on #12549.

### DIFF
--- a/integration_tests/coarrays_40.f90
+++ b/integration_tests/coarrays_40.f90
@@ -2,12 +2,7 @@ program coarrays_40
   ! Unused team_type variables must not break compilation (issue #12538)
   use iso_fortran_env, only: team_type
   implicit none
-
   type(team_type) :: subteam
   type(team_type) :: default_team
-  integer :: n
-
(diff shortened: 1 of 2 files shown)


bonachea (commented):
Thanks for the fixes on this @certik !

The approach sounds right to me, although I won't have the opportunity to test it until later today.

However I'd like to note the original problem is not specific to `FORM TEAM`: that just happens to be the statement I chose in the MRE. The coarray pass does a similar rewriting for `SYNC TEAM` and `CHANGE TEAM`, but they all use `convert_team_type()` so presumably this "just works" for those as well?

CC: @Jainam-not-a-robot

certik:
@Jainam-not-a-robot why don't you fix it for all the other issues as well? Add tests.

certik:
Here is the follow-up issue: https://github.com/lfortran/lfortran/issues/12559

PR #12562, opened 2026-08-21
fix: correct name mangling for submodule functions in self-named generics

### LINKED ISSUE #12395: Self-named generic with a `module function` implemented in a submodule fails to link
When a generic interface has a specific procedure of the same name (a "self-named" generic), LFortran stores that specific under the disambiguated symbol table key `<name>~genericprocedure`. If the specific is a **separate module function implemented in a submodule**, the call site and the implementation disagree on the mangled name and the program fails to link.

## MRE

```fortran
module m
 implicit none
 interface f
  module function f() result(y)
   integer :: y
  end function f
 end interface
end module m

submodule (m) sm
contains
 module function f() result(y)
  integer :: y
  y = 4
 end function f
end submodule sm

program p
 use m
 implicit none
 if (f() /= 4) error stop
 print *, "PASS"
end program p
```

## Expected

Prints `PASS`. `flang` compiles and runs it:

```
$ flang mre.f90 -o a.out && ./a.out
 PASS
```

## Actual

```
$ lfortran mre.f90 -o a.out
Undefined symbols for architecture arm64:
  "___module_m_f~genericprocedure", referenced from:
      _main in mre.tmp_fFYCyZLeLIVCVfFd9ATkEt.o
ld: symbol(s) not found for architecture arm64
```

## Cause

The call site mangles the disambiguated symbol table key, but the submodule implementation is emitted under the plain name:

```
$ lfortran --show-llvm mre.f90 | grep -E '^(define|declare).*_f'
declare i32 @"__module_m_f~genericprocedure"()      <-- referenced by the caller
define  i32 @__module_m_f()                          <-- defined by the submodule
```

The two never meet. The submodule's separate module procedure is not being registered under the `~genericprocedure` key that the interface body occupies.

The equivalent subroutine case gets this right — there the implementation is emitted as `__module_m_s~genericprocedure`, matching the call — which suggests the fix belongs in the separate-module-procedure path of `visit_Function` in `src/lfortran/semantics/ast_symboltable_visitor.cpp`, mirroring what `visit_Subroutine` already does.

## Conditions

All three are required; dropping any one makes it link and run correctly:

| Variant | Result |
| --- | --- |
| self-named generic + `module function` + submodule | **fails to link** |
| non-self-named generic (`interface g` / `module function f`) + submodule | passes |
| self-named generic + `module subroutine` + submodule | passes |
| self-named generic + `module function`, implemented in the same module's `contains` | passes |

## Version

```
LFortran version: 0.64.0-141-gf14d9f5fa7
Platform: macOS ARM
LLVM: 11.1.0
```

Reference compiler: `flang version 22.0.0git`.

## Notes

Found while reviewing #12384 (`Function.link_name`); it is independent of that PR and reproduces on `main`. It is also why the regression test added there uses `module subroutine` rather than `module function`.

srinjoy933:
Fixes #12395

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -1986,7 +1986,7 @@ RUN(NAME modules_70 LABELS gfortran llvm)
 RUN(NAME modules_71 LABELS gfortran llvm)
 RUN(NAME modules_72 LABELS gfortran llvm
         EXTRAFILES modules_72_stypemod.f90 modules_72_otypemod.f90)
-
+RUN(NAME module_func_generic_submodule LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
 RUN(NAME associate_06 LABELS gfortran EXTRAFILES
         associate_06_module.f90)
 
--- a/integration_tests/module_func_generic_submodule.f90
+++ b/integration_tests/module_func_generic_submodule.f90
@@ -1,0 +1,23 @@
+module m
+ implicit none
+ interface f
+  module function f() result(y)
+   integer :: y
+  end function f
+ end interface
+end module m
+
+submodule (m) sm
+contains
+ module function f() result(y)
+  integer :: y
+  y = 4
+ end function f
+end submodule sm
+
+program p
+ use m
+ implicit none
+ if (f() /= 4) error stop
+ print *, "PASS"
+end program p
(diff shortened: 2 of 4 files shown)


srinjoy933:
hey @certik ,please have a look at this one, Thanks!

certik (approved):

PR #12566, opened 2026-08-22
fix: dropped loops, untyped intrinsics and ** in CUDA do concurrent

### LINKED ISSUE #12369: Nested do loops get dropped inside do concurrent on the GPU path
Was putting some SLATEC kernels through the CUDA offload and noticed that a
`do concurrent` with a nested `do` in it comes out wrong. No error, it just
compiles and hands you zeros.

```fortran
subroutine k1(x, y, n)
  integer, intent(in) :: n
  real, intent(in) :: x(n)
  real, intent(out) :: y(n)
  real :: s
  integer :: i, j
  do concurrent (i = 1:n)
    s = 0.0
    do j = 1, 4
      s = s + x(i)
    end do
    y(i) = s
  end do
end subroutine
```

`lfortran --gpu=cuda -c k1.f90 -o k1.o`, and the `.cuda.cu` sidecar has:

```c
extern "C" __global__ void __lfortran_gpu_kernel_0(float *x, float *y, int n, int __loop_end_0) {
    ...
    i = (__flat_idx + 1);
    s = 0;
    j = (1 - 1);
    y[(i - 1)] = s;
}
```

The inner loop's init sticks around as `j = (1 - 1)` but the body is gone, so
`s` never accumulates. I linked the kernel and ran it with x = 3.0 and got 0
where 12 was due.

Same story ascending or descending, and with constant or runtime bounds. A
runtime bound leaves a `__do_loop_end = m` behind and drops the body anyway.
`if`/`else` inside a `do concurrent` is fine, so it looks specific to nested
loops.

Where I ran into it was SLATEC's special functions. They're nearly all built on
CSEVL, a Chebyshev recurrence, which is this exact shape, so it lowers to
`y = 0.5*(0 - 0)`.

No pressure on any of this, and it may well be covered by the lowering rework
you mentioned on #12348 anyway. Just seemed worth writing down somewhere since
it's quiet about it.

Version 0.64.0-74-g02efa6ac5.

Cheers!
Zane

Zaneham:
Six fixes to the CUDA `do concurrent` path, found pushing Fortran kernels through my compiler. Four were silent wrong answers rather than errors: a loop nested inside the `do concurrent` was dropped entirely, real literals lost precision and their decimal point so `1.0/3.0` became integer `0`, `real(8)` loops were never offloaded because of a restriction that only applies to Metal, and `abs`/`exp`/`sqrt` came out untyped. The other two were hard errors, `**` reaching the backend as a literal `**` and `a(i,j)` as `a[]`. The unsupported-statement and unsupported-intrinsic paths used to fall through quietly, which is how the dropped loop went unnoticed, so both raise a `CodeGenError` now.

Fixes #12369. The reproducer there gives 12 for x=3.0 now, where it gave 0.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -5011,6 +5011,11 @@ RUN(NAME gpu_metal_195 LABELS gfortran llvm metal)
 
 RUN(NAME gpu_kernel_source_01 LABELS gfortran llvm)
 RUN(NAME gpu_kernel_source_02 LABELS gfortran llvm)
+RUN(NAME gpu_kernel_source_03 LABELS gfortran llvm)
+RUN(NAME gpu_kernel_source_04 LABELS gfortran llvm)
+RUN(NAME gpu_kernel_source_05 LABELS gfortran llvm)
+RUN(NAME gpu_kernel_source_06 LABELS gfortran llvm)
+RUN(NAME gpu_kernel_source_07 LABELS gfortran llvm)
 
 RUN(NAME real128_arith_01 LABELS gfortran llvm)
 RUN(NAME real128_arith_02 LABELS gfortran llvm)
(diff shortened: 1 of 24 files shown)


certik (approved):
I think the changes look good, let's see if all tests pass.

I think going forward, we have CUDA API emulation, and we should use it to test all CUDA offloading. We just have to set it up (I think it's setup already in some form).

PR #12575, opened 2026-08-25
fix: classify BIND(C) character kind in semantics instead of patching codegen

### LINKED ISSUE #12571: code generation error in version 0.64.0
Code generation error.

This is reduced down from a larger working module
at https://github.com/urbanjost/M_regex that works
with five other compilers.

Almost any further reduction causes it to compile, 
including removing the unused variable char_temp.
```fortran
module M_regex
   use ISO_C_Binding, only: C_char
   implicit none
contains
   subroutine regcomp(this,pattern)
      character(len=*),intent(out)             :: this
      character(len=*), intent(in)             :: pattern
      character(kind=c_char,len=1),allocatable :: char_temp1(:)
      interface
         subroutine C_regcomp(preg,pattern) bind(C,name="C_regcomp")
            import
            character(kind=C_char)  :: pattern(*)
         end subroutine C_regcomp
      end interface
   end subroutine regcomp
end module M_regex
```
```diff
-lfortranlatest       :     2	; ModuleID = 'LFortran'
-lfortranlatest       :     3	source_filename = "LFortran"
-lfortranlatest       :     4	target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
-lfortranlatest       :     5	
-lfortranlatest       :     6	%lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz.lfortran_intrinsic_iso_c_binding_c_ptr_f7KyKyH8uCM4IPYRt49cRz.6 = type { i32 }
-lfortranlatest       :     7	%array.1 = type { i32*, i64, i32, i8, i8, i8, i8, i64, [1 x %dimension_descriptor] }
-lfortranlatest       :     8	%dimension_descriptor = type { i64, i64, i64 }
-lfortranlatest       :     9	%lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz.lfortran_intrinsic_iso_c_binding_c_funptr_f7KyKyH8uCM4IPYRt49cRz.5 = type { i32 }
-lfortranlatest       :    10	%string_descriptor = type <{ i8*, i64 }>
-lfortranlatest       :    11	%array.1.0 = type { i8**, i64, i32, i8, i8, i8, i8, i64, [1 x %dimension_descriptor] }
-lfortranlatest       :    12	
-lfortranlatest       :    13	declare i32 @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_associated_f7KyKyH8uCM4IPYRt49cRz(%lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz.lfortran_intrinsic_iso_c_binding_c_ptr_f7KyKyH8uCM4IPYRt49cRz.6*)
-lfortranlatest       :    14	
-lfortranlatest       :    15	declare void @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_f_pointer_f7KyKyH8uCM4IPYRt49cRz(%lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz.lfortran_intrinsic_iso_c_binding_c_ptr_f7KyKyH8uCM4IPYRt49cRz.6*, i32**, %array.1*, i32*)
-lfortranlatest       :    16	
-lfortranlatest       :    17	declare void @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_f_procpointer_f7KyKyH8uCM4IPYRt49cRz(%lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz.lfortran_intrinsic_iso_c_binding_c_funptr_f7KyKyH8uCM4IPYRt49cRz.5*, i32**)
-lfortranlatest       :    18	
-lfortranlatest       :    19	declare i32 @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_funloc_f7KyKyH8uCM4IPYRt49cRz(i32*)
-lfortranlatest       :    20	
-lfortranlatest       :    21	declare i32 @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_loc_f7KyKyH8uCM4IPYRt49cRz(i32*)
-lfortranlatest       :    22	
-lfortranlatest       :    23	declare i64 @__module_lfortran_intrinsic_iso_c_binding_lfortran_intrinsic_iso_c_binding_f7KyKyH8uCM4IPYRt49cRz_lfortran_intrinsic_iso_c_binding_c_sizeof_f7KyKyH8uCM4IPYRt49cRz(i32*)
-lfortranlatest       :    24	
-lfortranlatest       :    25	define void @__module_m_regex_regcomp(%string_descriptor* %this, %string_descriptor* %pattern) {
-lfortranlatest       :    26	.entry:
-lfortranlatest       :    27	  %arr_desc_str_desc = alloca %string_descriptor, align 8
-lfortranlatest       :    28	  %arr_desc = alloca %array.1.0, align 8
-lfortranlatest       :    29	  %char_temp1 = alloca %array.1.0*, align 8
-lfortranlatest       :    30	  store %array.1.0* null, %array.1.0** %char_temp1, align 8
-lfortranlatest       :    31	  %0 = getelementptr %string_descriptor, %string_descriptor* %arr_desc_str_desc, i32 0, i32 0
-lfortranlatest       :    32	  store i8* null, i8** %0, align 8
-lfortranlatest       :    33	  %1 = getelementptr %string_descriptor, %string_descriptor* %arr_desc_str_desc, i32 0, i32 1
-lfortranlatest       :    34	  store i64 0, i64* %1, align 8
-lfortranlatest       :    35	  %2 = getelementptr %string_descriptor, %string_descriptor* %arr_desc_str_desc, i32 0, i32 1
-lfortranlatest       :    36	  store i64 1, i64* %2, align 8
-lfortranlatest       :    37	  %3 = getelementptr %array.1.0, %array.1.0* %arr_desc, i32 0, i32 0
-lfortranlatest       :    38	  store %string_descriptor* %arr_desc_str_desc, i8*** %3, align 8
-lfortranlatest       :    39	  %4 = getelementptr %array.1.0, %array.1.0* %arr_desc, i32 0, i32 8
-lfortranlatest       :    40	  %5 = getelementptr [1 x %dimension_descriptor], [1 x %dimension_descriptor]* %4, i32 0, i32 0
-lfortranlatest       :    41	  %6 = getelementptr inbounds %dimension_descriptor, %dimension_descriptor* %5, i32 0
-lfortranlatest       :    42	  %7 = getelementptr %dimension_descriptor, %dimension_descriptor* %6, i32 0, i32 0
-lfortranlatest       :    43	  store i64 1, i64* %7, align 8
-lfortranlatest       :    44	  %8 = getelementptr %dimension_descriptor, %dimension_descriptor* %6, i32 0, i32 1
-lfortranlatest       :    45	  store i64 1, i64* %8, align 8
-lfortranlatest       :    46	  %9 = getelementptr %dimension_descriptor, %dimension_descriptor* %6, i32 0, i32 2
-lfortranlatest       :    47	  store i64 0, i64* %9, align 8
-lfortranlatest       :    48	  %10 = getelementptr %array.1.0, %array.1.0* %arr_desc, i32 0, i32 3
-lfortranlatest       :    49	  store i8 1, i8* %10, align 1
-lfortranlatest       :    50	  store %array.1.0* %arr_desc, %array.1.0** %char_temp1, align 8
-lfortranlatest       :    51	  br label %return
-lfortranlatest       :    52	
-lfortranlatest       :    53	return:                                           ; preds = %.entry
-lfortranlatest       :    54	  br label %FINALIZE_SYMTABLE_regcomp
-lfortranlatest       :    55	
-lfortranlatest       :    56	FINALIZE_SYMTABLE_regcomp:                        ; preds = %return
-lfortranlatest       :    57	  br label %Finalize_Variable_char_temp1
-lfortranlatest       :    58	
-lfortranlatest       :    59	Finalize_Variable_char_temp1:                     ; preds = %FINALIZE_SYMTABLE_regcomp
-lfortranlatest       :    60	  %11 = load %array.1.0*, %array.1.0** %char_temp1, align 8
-lfortranlatest       :    61	  call void @finalize_allocatable__Array_1_str(%array.1.0* %11)
-lfortranlatest       :    62	  ret void
-lfortranlatest       :    63	}
-lfortranlatest       :    64	
-lfortranlatest       :    65	declare void @C_regcomp(float*, i8*)
-lfortranlatest       :    66	
-lfortranlatest       :    67	define internal void @finalize_allocatable__Array_1_str(%array.1.0* %0) {
-lfortranlatest       :    68	entry:
-lfortranlatest       :    69	  %1 = icmp ne %array.1.0* %0, null
-lfortranlatest       :    70	  br i1 %1, label %is_allocated.then, label %is_allocated.else
-lfortranlatest       :    71	
-lfortranlatest       :    72	is_allocated.then:                                ; preds = %entry
-lfortranlatest       :    73	  call void @finalize_descriptorArray_Array_1_str(%array.1.0* %0)
-lfortranlatest       :    74	  br label %is_allocated.ifcont
-lfortranlatest       :    75	
-lfortranlatest       :    76	is_allocated.else:                                ; preds = %entry
-lfortranlatest       :    77	  br label %is_allocated.ifcont
-lfortranlatest       :    78	
-lfortranlatest       :    79	is_allocated.ifcont:                              ; preds = %is_allocated.else, %is_allocated.then
-lfortranlatest       :    80	  ret void
-lfortranlatest       :    81	}
-lfortranlatest       :    82	
-lfortranlatest       :    83	define internal void @finalize_descriptorArray_Array_1_str(%array.1.0* %0) {
-lfortranlatest       :    84	entry:
-lfortranlatest       :    85	  %1 = call i8* @_lfortran_get_default_allocator()
-lfortranlatest       :    86	  %2 = getelementptr %array.1.0, %array.1.0* %0, i32 0, i32 0
-lfortranlatest       :    87	  %3 = load %string_descriptor*, i8*** %2, align 8
-lfortranlatest       :    88	  %4 = getelementptr %string_descriptor, %string_descriptor* %3, i32 0, i32 0
-lfortranlatest       :    89	  %5 = load i8*, i8** %4, align 8
-lfortranlatest       :    90	  call void @_lfortran_free_alloc(i8* %1, i8* %5)
-lfortranlatest       :    91	  %6 = ptrtoint %string_descriptor* %3 to i64
-lfortranlatest       :    92	  %7 = icmp ne i64 %6, 0
-lfortranlatest       :    93	  br i1 %7, label %then, label %else
-lfortranlatest       :    94	
-lfortranlatest       :    95	then:                                             ; preds = %entry
-lfortranlatest       :    96	  br label %ifcont
-lfortranlatest       :    97	
-lfortranlatest       :    98	else:                                             ; preds = %entry
-lfortranlatest       :    99	  br label %ifcont
-lfortranlatest       :   100	
-lfortranlatest       :   101	ifcont:                                           ; preds = %else, %then
-lfortranlatest       :   102	  ret void
-lfortranlatest       :   103	}
-lfortranlatest       :   104	
-lfortranlatest       :   105	declare void @_lfortran_free_alloc(i8*, i8*)
-lfortranlatest       :   106	
-lfortranlatest       :   107	declare i8* @_lfortran_get_default_allocator()
-lfortranlatest       :   108	asr_to_llvm: module failed verification. Error:
-lfortranlatest       :   109	Stored value type does not match pointer operand type!
-lfortranlatest       :   110	  store %string_descriptor* %arr_desc_str_desc, i8*** %3, align 8
-lfortranlatest       :   111	 i8**
-lfortranlatest       :   112	STDERR:
-lfortranlatest       :   114	code generation error: asr_to_llvm: module failed verification. Error:
-lfortranlatest       :   115	Stored value type does not match pointer operand type!
-lfortranlatest       :   116	  store %string_descriptor* %arr_desc_str_desc, i8*** %3, align 8
-lfortranlatest       :   117	 i8**
-lfortranlatest       :   118	
-lfortranlatest       :   119	
-lfortranlatest       :   120	Note: Please report unclear, confusing or incorrect messages as bugs at
-lfortranlatest       :   121	https://github.com/lfortran/lfortran/issues.
```

certik:
## Summary

Fixes #12571 — **alternative to #12573**, fixing the same bug in semantics instead of in codegen. See the comparison below; only one of the two should be merged.

The reported crash:

```
  store %string_descriptor* %arr_desc_str_desc, i8*** %3
 i8**
```

The BIND(C) dummy ended up with an ASR type that contradicts itself — `Array(String(…, DescriptorString), StringArraySinglePointer)`. `DescriptorString` says the elements are `%string_descriptor`; `StringArraySinglePointer` says the whole array is one flat `i8*`. The backend resolved the contradiction on its own at `llvm_utils.cpp:900` ("`// Treat it as a DescriptorArray`", element type `i8*`), while the descriptor cache key still said `strdesc` — so the dummy and a genuine allocatable string array shared one LLVM descriptor struct and the wrong pointer type got stored.

Two independent semantic bugs produced that contradictory type.

### 1. The BIND(C) character kind was matched by spelling

`ast_common_visitor.h` string-compared the source identifier against `"c_char"`. Fortran is case insensitive and the kind can be imported under another name, so these three semantically identical declarations produced two different ASR types:

| Declaration | Before | After |
| --- | --- | --- |
| `character(kind=c_char)` | `CChar` | `CChar` |
| `character(kind=C_char)` | `DescriptorString` | `CChar` |
| `use iso_c_binding, only: my_char => c_char` … `character(kind=my_char)` | `DescriptorString` | `CChar` |

**The issue's own code uses capital `C_char`** — that is exactly why it hit this. Now the name is resolved through the symbol table and checked to be `c_char` from the intrinsic `iso_c_binding` module, which also satisfies the "use enums or structured types, not string comparisons" rule in `AGENTS.md`.

### 2. The element `String` was not kept consistent with the array

A `StringArraySinglePointer` array is one flat character buffer, so its elements are plain C characters. The element `String` kept whatever physical type it already had. It is now set where the array physical type is decided, so the pair cannot disagree.

This part also covers a plain `character :: pattern(*)` dummy, which has no `c_char` spelling to resolve but is interoperable all the same (the default character kind *is* `c_char`).

## Scope

Semantics only. No backend change — with the ASR consistent, the existing descriptor cache key is already unambiguous.

An earlier attempt classified by ABI + kind + length inside `determine_char_len_and_kind` instead. That regressed `string_29`, which has an assumed-shape `character(len=1), intent(in) :: s(:)` dummy in a BIND(C) function: assumed-shape arrays are not interoperable and legitimately keep descriptors, so an ABI-based rule marked them wrong. Deriving the string physical type from the array physical type — at the one site where the latter is computed — avoids re-deriving that condition and gets the case right by construction.

## Verification

Two integration tests, both `LABELS gfortran llvm`, both confirmed to **fail on main** with the codegen error and pass here, with output matching GFortran and Flang:

- `bindc_55` — plain `character` assumed-size dummy alongside an allocatable character array.
- `bindc_56` — the same shape via mixed-case `C_char` and via a renamed import, covering both spellings the old check missed.

```
$ cd integration_tests && ./run_tests.py -j16 &> log
100% tests passed, 0 tests failed out of 3891

$ ./run_tests.py &> log
TESTS PASSED
```

No reference outputs needed updating.

## Comparison with #12573

| | #12573 (codegen) | this PR (semantics) |
| --- | --- | --- |
| Layer | `llvm_array_utils.cpp` descriptor cache | AST→ASR |
| Fixes the issue's code | yes | yes |
| Fixes plain `character` in BIND(C) | yes | yes |
| Fixes the `C_char` / renamed-import misclassification | no — ASR stays wrong | yes |
| Contradictory ASR still constructible | yes | no, for this path |
| Tests | `bindc_55` | `bindc_55`, `bindc_56` |

`AGENTS.md` says the backend "must never infer or fix types — it should only lower what ASR gives it. If codegen needs a type workaround, the bug is upstream." That points at this PR. #12573 remains useful as a backstop, but if it lands it would be better reduced to an assertion on the cache invariant rather than a silent repair.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -2036,6 +2036,8 @@ RUN(NAME bindc_52 LABELS gfortran llvm)
 # Extension: BIND(C) character(len>1); GFortran rejects, so llvm only
 RUN(NAME bindc_53 LABELS llvm)
 RUN(NAME bindc_54 LABELS gfortran llvm)
+RUN(NAME bindc_55 LABELS gfortran llvm)
+RUN(NAME bindc_56 LABELS gfortran llvm)
 RUN(NAME bindc_iso_fb_01 LABELS gfortran llvm flang EXTRAFILES bindc_iso_fb_01c.c)
 RUN(NAME bindc_iso_fb_02 LABELS gfortran llvm flang EXTRAFILES bindc_iso_fb_02c.c)
 RUN(NAME bindc_iso_fb_03 LABELS gfortran llvm flang EXTRAFILES bindc_iso_fb_03c.c)
--- a/integration_tests/bindc_55.f90
+++ b/integration_tests/bindc_55.f90
@@ -1,0 +1,49 @@
+! An allocatable character array declared alongside an interface body for a
+! BIND(C) procedure that takes an assumed-size character dummy argument.
+! Regression test: the dummy is a StringArraySinglePointer array, i.e. one flat
+! character buffer, but its element String was left as DescriptorString. The two
+! physical types disagreed, so the backend had to pick a representation on its
+! own and ended up sharing one LLVM array descriptor between this dummy and the
+! allocatable array, whose elements really are string descriptors.
+module bindc_55_mod
+  implicit none
+contains
+  subroutine fill(n, total)
+    integer, intent(in) :: n
+    integer, intent(out) :: total
+    character, allocatable :: buf(:)
+    integer :: k
+    interface
+      subroutine c_take(pattern) bind(c, name="c_take")
+        character :: pattern(*)
+      end subroutine c_take
+    end interface
+    allocate(buf(n))
+    do k = 1, n
+      buf(k) = achar(iachar('a') + k - 1)
+    end do
+    if (buf(1) /= 'a') error stop
+    if (buf(n) /= achar(iachar('a') + n - 1)) error stop
+    total = size(buf)
+    do k = 1, n
+      total = total + iachar(buf(k))
+    end do
+    deallocate(buf)
+  end subroutine fill
+end module bindc_55_mod
+
+program bindc_55
+  use bindc_55_mod
+  implicit none
+  integer :: total
+
+  call fill(3, total)
+  print *, total
+  ! size + iachar('a') + iachar('b') + iachar('c') = 3 + 97 + 98 + 99
+  if (total /= 297) error stop
+
+  call fill(1, total)
+  print *, total
+  ! size + iachar('a') = 1 + 97
+  if (total /= 98) error stop
+end program bindc_55
--- a/integration_tests/bindc_56.f90
+++ b/integration_tests/bindc_56.f90
@@ -1,0 +1,71 @@
+! The BIND(C) character kind must be recognised from the resolved symbol, not
+! from how it is spelled: `C_char` (mixed case) and a renamed import both name
+! the same `c_char` from iso_c_binding, so all three spellings must produce the
+! same interoperable character type. Regression test: only the exact lowercase
+! spelling `c_char` was recognised, so the other two produced a different
+! physical type and the module failed LLVM verification.
+module bindc_56_mod
+  use ISO_C_Binding, only: C_char
+  use iso_c_binding, only: renamed_char => c_char
+  implicit none
+contains
+  subroutine upper_case(n, total)
+    integer, intent(in) :: n
+    integer, intent(out) :: total
+    character(kind=C_char, len=1), allocatable :: buf(:)
+    integer :: k
+    interface
+      subroutine c_upper(pattern) bind(C, name="c_upper")
+        import
+        character(kind=C_char) :: pattern(*)
+      end subroutine c_upper
+    end interface
+    allocate(buf(n))
+    do k = 1, n
+      buf(k) = achar(iachar('A') + k - 1)
+    end do
+    total = 0
+    do k = 1, n
+      total = total + iachar(buf(k))
+    end do
+    deallocate(buf)
+  end subroutine upper_case
+
+  subroutine renamed_kind(n, total)
+    integer, intent(in) :: n
+    integer, intent(out) :: total
+    character(kind=renamed_char, len=1), allocatable :: buf(:)
+    integer :: k
+    interface
+      subroutine c_lower(pattern) bind(C, name="c_lower")
+        import
+        character(kind=renamed_char) :: pattern(*)
+      end subroutine c_lower
+    end interface
+    allocate(buf(n))
+    do k = 1, n
+      buf(k) = achar(iachar('a') + k - 1)
+    end do
+    total = 0
+    do k = 1, n
+      total = total + iachar(buf(k))
+    end do
+    deallocate(buf)
+  end subroutine renamed_kind
+end module bindc_56_mod
+
+program bindc_56
+  use bindc_56_mod
+  implicit none
+  integer :: total
+
+  call upper_case(3, total)
+  print *, total
+  ! iachar('A') + iachar('B') + iachar('C') = 65 + 66 + 67
+  if (total /= 198) error stop
+
+  call renamed_kind(3, total)
+  print *, total
+  ! iachar('a') + iachar('b') + iachar('c') = 97 + 98 + 99
+  if (total /= 294) error stop
+end program bindc_56
--- a/src/lfortran/semantics/ast_common_visitor.h
+++ b/src/lfortran/semantics/ast_common_visitor.h
@@ -9986,11 +9986,22 @@ public:
                     nullptr, 0,
                     ASR::array_physical_typeType::AssumedRankArray));
             } else {
-                type = ASRUtils::make_Array_t_util(
-                    al, loc, type, dims.p, dims.size(), abi, is_argument,
+                ASR::array_physical_typeType array_physical_type =
                     dims.size() > 0 && abi == ASR::abiType::BindC && (is_dimension_star || ASRUtils::is_fixed_size_array(dims.p, dims.n)) ? ASR::array_physical_typeType::StringArraySinglePointer :
                                     ASRUtils::is_fixed_size_array(dims.p, dims.n) ? ASR::array_physical_typeType::PointerArray :
-                                    ASR::array_physical_typeType::DescriptorArray,
+                                    ASR::array_physical_typeType::DescriptorArray;
+                // A StringArraySinglePointer array is one flat character
+                // buffer, so its elements are plain C characters rather than
+                // string descriptors. Keep the two physical types consistent
+                // here, where the array one is decided: a mismatched pair
+                // leaves the representation of the array ambiguous, and any
+                // consumer would have to pick a winner on its own.
+                if (array_physical_type == ASR::array_physical_typeType::StringArraySinglePointer) {
+                    str->m_physical_type = ASR::CChar;
+                }
+                type = ASRUtils::make_Array_t_util(
+                    al, loc, type, dims.p, dims.size(), abi, is_argument,
+                    array_physical_type,
                     dims.size() > 0 ? true : false);
             }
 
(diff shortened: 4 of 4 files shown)


certik:
Follow-up in #12576 adds an ASR verify rule that makes the contradictory `Array(StringArraySinglePointer)` + `String(DescriptorString)` pair unconstructible, with a hand-written ASR test for it. It is stacked on this branch, and it subsumes the site-specific fix here by normalizing centrally in `make_Array_t_util` — which also caught a second, pre-existing producer of the same pair in `ArrayPhysicalCast`.

PR #12576, opened 2026-08-25
verify: reject inconsistent string array physical types in ASR

certik:
## Summary

Makes the ASR that caused #12571 **unconstructible**, so the next occurrence fails immediately in ASR verify instead of surfacing as an LLVM verification error much later.

**Stacked on #12575** — it contains that PR's commit. Review or merge that one first; this PR's own commit is `verify: reject inconsistent string array physical types`.

An `Array` of physical type `StringArraySinglePointer` is one flat character buffer, so its elements are plain C characters. Nothing stopped ASR from pairing it with `DescriptorString` elements, which says the elements are string descriptors instead. The pair describes two layouts at once, and every consumer had to pick one on its own — the LLVM backend picked `i8*` at `asr_to_llvm.cpp:900` while the descriptor cache key still said `strdesc`, which is exactly the wrong store behind #12571.

## Scope

**1. The verify rule** — `asr_verify.cpp`, in `visit_Array`, next to the existing "array of strings must not be `FixedSizeArray`" rule:

```
ASR verify pass error: Array of strings with physical type "StringArraySinglePointer" must have string physical type "CChar", not "DescriptorString"
  --> tests/asr/verify/bindc_string_array_physical_type.asr:21:22
   |
21 |                     (Array
   |                      ^^^^^ failed here
```

**2. The test that constructs the invalid ASR** — `tests/asr/verify/bindc_string_array_physical_type.asr`, hand-written ASR carrying the contradictory pair, registered in `tests.toml` with the verifier error captured as the reference (`returncode` 2), following the other 50 `asr/verify` tests.

**3. `make_Array_t_util` normalizes the pair.** Turning the rule on immediately caught a second producer: `ArrayPhysicalCast` from `DescriptorArray` to `StringArraySinglePointer` copied the element type across unchanged, so its result type carried `DescriptorString` elements. That predates all of this work — it is visible in the stored `array12` reference on main. Rather than patch that one site, normalize in `make_Array_t_util`, the single place `Array` types are built, so the invariant holds for every producer present and future. This also subsumes the site-specific fix from #12575, which is reverted here.

**4. One backend adjustment.** The `DescriptorArray` → `StringArraySinglePointer` cast read its operand using the cast's *result* string type, which is now `CChar` and describes the buffer being produced rather than what is being read from. It now uses the argument's string type, matching how `data_type`, `arr_type`, and the neighbouring `PointerArray` branch already derive everything from the operand.

## Verification

```
$ cd integration_tests && ./run_tests.py -j16 &> log
100% tests passed, 0 tests failed out of 3891

$ ./run_tests.py &> log
TESTS PASSED
```

One reference updated: `array12`, a single line, `DescriptorString` → `CChar` in the cast's result type — the fix being recorded.

## Why this is worth having

#12573 and #12575 each stop *one* route to the bad type. This stops the type from existing. The verify rule found a producer I had not suspected within minutes of being switched on, which is the argument for it in miniature.


certik:
@Jainam-not-a-robot I like this PR a lot --- it now catches this error in ASR verify and this caught another bug that we didn't know about before, and is now fixed.

PR #12581, opened 2026-08-25
Support ISO_10646 (UCS-4) character values

### LINKED ISSUE #12574: Unicode ISO-10646 support not working in version 0.64.0
Is Unicode support via "ISO_10646" supported yet?

Using this small code produces incorrect output:
```fortran
program testit
use iso_fortran_env
implicit none

character(len=*),parameter :: ident = "@(#) print date and time in japanese"
integer, parameter :: ascii = selected_char_kind ("ascii")
integer,parameter :: ucs4 = selected_char_kind("ISO_10646")

character(kind=ascii, len=26) :: alphabet = ascii_"abcdefghijklmnopqrstuvwxyz"
character(kind=ucs4,  len=30) :: hello_world
character(len=100,kind=ucs4) :: str
   hello_world = ucs4_'Hello World and Ni Hao -- ' &
      // char (int (z'4F60'), ucs4)                &
      // char (int (z'597D'), ucs4)
   write (*,*) alphabet
   open (output_unit, encoding='UTF-8')
   write (*,*) trim (hello_world)
   call create_date_string(str)
   write(*,*)str
contains
subroutine create_date_string(str)
   intrinsic date_and_time,selected_char_kind
   integer,parameter :: ucs4 = selected_char_kind("ISO_10646")
   character(kind=ucs4),parameter :: nen = char(int(z'5e74'),ucs4)
   character(kind=ucs4),parameter :: gatsu = char(int(z'6708'),ucs4)
   character(kind=ucs4),parameter :: nichi = char(int(z'65e5'),ucs4)
   character(len= *, kind= ucs4) str
   integer values(8)
       write(str,10) values(1),nen,values(2),gatsu,values(3),nichi
   10  format(i0,a,i0,a,i0,a)
end subroutine
end program testit
```
AOCC flang/nvfortran and ifx does not support character kind "ISO_10646",
but this program works as intended for LLVM flang and gfortran but LFortran
produces incorrect output. The output was run through "cat -vet" but is
correct for LLVM flang and gfortran and actually looks like
```text
 abcdefghijklmnopqrstuvwxyz
 Hello World and Ni Hao -- 你好
 2026年8月24日
```
```diff
+flangtrunk           :     2    abcdefghijklmnopqrstuvwxyz
+flangtrunk           :     3    Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
+flangtrunk           :     4    2026M-eM-9M-48M-fM-^\M-^H24M-fM-^WM-%

+lfortranlatest       :     2    abcdefghijklmnopqrstuvwxyz
-lfortranlatest       :     3    H
-lfortranlatest       :     4    2026t8^H24M-oM-?M-=

+gfortran161          :     2    abcdefghijklmnopqrstuvwxyz
+gfortran161          :     3    Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
+gfortran161          :     4    2026M-eM-9M-48M-fM-^\M-^H24M-fM-^WM-%

-aoccflang520         :     2   F90-S-0081-Illegal selector - KIND value must be non-negative  (/app/example.f90: 10)
-aoccflang520         :    11   F90/x86-64 Linux AOCCAOCC_5.2.0-Build#2035 2026_04_10: compilation completed with severe errors

-ifxlatest            :     2   /app/example.f90(10): error #6684: This is an incorrect value for a kind type parameter in this context.   [UCS4]
-ifxlatest            :     3   character(kind=ucs4,  len=30) :: hello_world
-ifxlatest            :     4   ---------------^
-ifxlatest            :     5   /tmp/ifx15229366272mRaFE/ifxu2kkWL.i90(39): catastrophic error: Too many errors, exiting
-ifxlatest            :     6   compilation aborted for /app/example.f90 (code 1)

-nvfortran_x86_26_3   :    10   NVFORTRAN-S-0081-Illegal selector - KIND value must be non-negative  (/app/example.f90: 30)
-nvfortran_x86_26_3   :    11   NVFORTRAN/x86-64 Linux 26.3-0: compilation completed with severe errors

### PR DESCRIPTION
**Pull request by certik**
Fixes [#12574](https://github.com/lfortran/lfortran/issues/12574).

First of three. Two commits, which have to land together — see **Why one PR** below. [#12582](https://github.com/lfortran/lfortran/pull/12582) (kind validation) and [#12584](https://github.com/lfortran/lfortran/pull/12584) (backend diagnostics) stack on this one.

## The bug

```fortran
integer,parameter :: ucs4 = selected_char_kind("ISO_10646")
character(kind=ucs4, len=30) :: hello_world
hello_world = ucs4_'Hello World and Ni Hao -- ' &
   // char (int (z'4F60'), ucs4) // char (int (z'597D'), ucs4)
open (output_unit, encoding='UTF-8')
write (*,*) trim (hello_world)
```

```
main:      H
branch:    Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
gfortran:  Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
flang:     Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
```

Byte-identical to both reference compilers, on the issue's program and on the new tests.

## Commit 1 — represent kind 4 values correctly in ASR

A kind 4 value was represented inconsistently: its ASR length was a count of UTF-8 *bytes*, while the rest of the pipeline treats a string length as a count of *characters*. The character intrinsics also dropped the kind from their results.

| | `main` | GFortran 13 / Flang 18 |
| --- | --- | --- |
| `kind(trim(padded))` | 1 | 4 |
| `kind(new_line(padded))` | 1 | 4 |
| `len_trim(padded)` | 6 | 2 |
| `ichar(ni, kind=4)` | -28 | 20320 |
| `index(padded, hao)` | 4 | 2 |
| `ichar(char(code, ucs4), kind=4)` (non-constant `code`) | 96 | 20320 |

- **Length is a character count everywhere.** A kind > 1 literal takes the number of code points its UTF-8 encodes; constant folding of `trim`, `len_trim`, `adjustl`, `adjustr`, `repeat`, `index`, `scan`, `verify`, `ichar`, `iachar` and substrings walks characters, not bytes. Substring folding now blank-pads an initializer shorter than the declared length instead of reading past the end of the value — an out-of-bounds read on `main`, masked because the folded length came out right by accident.
- **Result kinds follow the argument.** `create_StringTrim` and `create_NewLine` returned kind 1; `Char`, `Achar`, `Iachar`, `Adjustl`, `Adjustr`, `StringLenTrim`, `StringFindSet`, `StringContainsSet` and `StringConcat` built kind 1 dummies, temporaries and blanks, so a kind 4 operand was compared against a one-byte blank. `ASRBuilder::StringItem` now takes the element kind from its operand, and a new `ASRBuilder::StringBlank(kind)` replaces the open-coded kind 1 `" "` literals.
- **One UTF-8 codec instead of three.** `asr_text.cpp`, `asr_text_parser.cpp` and `llvm_utils.cpp` each carried a near-identical decoder; they now share helpers in `string_utils`, which also gain the missing UCS-4 → UTF-8 direction.

The in-memory representation is unchanged: a kind 4 value is still `len * 4` bytes of little-endian UCS-4, and `StringConstant::m_s` still holds UTF-8 (a `char*` cannot carry the embedded nulls raw UCS-4 would produce). Only the *length* changes meaning for kind > 1, from bytes to characters.

## Commit 2 — write and read kind 4 as UTF-8

Every output path treated the descriptor's character count as a byte count, so a write emitted the first byte of the record and stopped at the first embedded null. Internal `WRITE` into a kind 4 variable stored kind 1 bytes; a formatted `READ` copied raw bytes into code units.

- **The kind reaches the runtime.** `SerializeType` emits it as `S-DESC-K<kind>`, alongside the kind every other type already carried (`I4`, `R8`, `L32`) — strings were the only type whose kind was dropped. The format engine transcodes a kind 4 argument when it consumes it, counting `Aw` widths in characters.
- **The two fast paths that bypass the format engine** for a lone character value carry the kind too: `_lfortran_printf` takes it as a parameter, and `_lfortran_file_write` learns `%S`, which reads a kind after the length.
- **Internal writes.** `_lfortran_string_write` takes the internal file's kind and decodes the record back into code units, blank-padding in characters. It also drops a `sprintf("%.*s")` staging step that truncated at the first null.
- **Reads.** A kind > 1 destination uses read type code 8 and carries its kind; the field is decoded from UTF-8 when the connection has `ENCODING='UTF-8'`, and otherwise one byte becomes one character. `InputSource` gains the connection's `encoding_mode` for this. Both reference compilers agree exactly here, with and without the specifier.

`Aw` widths follow the standard rather than GFortran: GFortran's `Aw` for kind 4 is demonstrably wrong (`a10` on a six-character value emits ten blanks; `a3` and `a2` both emit all six). The tests only assert behaviour all three compilers agree on.

**`ENCODING=` on output follows Flang:** kind 4 output is transcoded regardless of the unit's encoding. GFortran instead substitutes `?` on a non-UTF-8 unit. Following Flang keeps the write path free of any per-unit encoding lookup and still gives the issue's program the output it expects. `ENCODING=` was already parsed and stored per unit; before this its only consumer was `INQUIRE`, and it now also drives decoding on input.

## Why one PR

These were opened as two ([#12583](https://github.com/lfortran/lfortran/pull/12583) held the second commit) and are now merged, because the first is a **regression on its own**.

`integration_tests/print_12.f90` is a `run =` reference test, so its stdout is compared byte for byte. On `main` its concatenation folded to kind 1, so the value's UTF-8 bytes printed correctly *by accident*, and the stored reference is correct UTF-8. Typing the result as kind 4 — which commit 1 does, correctly — makes it print raw UCS-4 until the output path understands the typing, which commit 2 adds:

```
main:       Hello = Ni Hao = M-dM-=M- M-eM-%M-=
commit 1:   H^@^@^@e^@^@^@l^@^@^@l^@^@^@o^@^@
commit 1+2: Hello = Ni Hao = M-dM-=M- M-eM-%M-=     (identical to main's reference)
```

Merging commit 1 alone would regress that output on `main`. The two are one change.

## Tests

All self-checking with `error stop`, labelled `gfortran llvm`, passing on GFortran 13, LLVM Flang 18 and this branch. Integration tests can't compare stdout — the runner only checks the exit status — so byte-exactness is verified by writing to a scratch file with `encoding='UTF-8'`, reopening it as default `character`, and checking the bytes with `ichar`.

- `char4_string_intrinsics_02.f90` — result kinds, and `trim`/`len_trim`/`adjustl`/`adjustr`/`repeat`/`index`/`scan`/`verify`/`//`/`char`/`achar`/`ichar` over CJK code points, checked at run time and as folded constants. Fails at `error stop 11` on `main`.
- `char4_io_01.f90` — `'(a)'`, `'("<",a,">")'` and list-directed output, each checked against the expected UTF-8 bytes.
- `char4_io_02.f90` — the issue's internal-write case (`write(line,'(i0,a,i0,a,i0,a)') 2026, nen, …`) checked character by character, blank padding of a short record, and a write/read round trip through a UTF-8 file.
- `char4_string_literal_01.f90` (label `llvm` only) — a kind 4 literal in UTF-8 source has a length in characters. `llvm` only because the reference compilers disagree: Flang decodes the source into code points (`len(ucs4_"你好") == 2`), GFortran treats each source byte as a character (`6`). LFortran's backend already decoded such literals, so only the length was out of step; this follows Flang.
- `print_12.f90` already exercised this path but asserted nothing beyond the reference stdout. It now checks the length and the bytes that reach the stream; its stdout reference is unchanged, only the source hash.

## Verification

```
./run_tests.py --skip-run-with-dbg -t <each of the 117 `run =` tests>   # all pass
./run_tests.py --no-llvm --skip-run-with-dbg                           # TESTS PASSED
cd integration_tests && ./run_tests.py -b llvm -j4                     # pass
cd integration_tests && ./run_tests.py -b gfortran -j4                 # pass
```

The `run =` reference tests are enumerated explicitly because `--no-llvm` skips them, which is how the `print_12` regression got past the first round of local checks. `--no-llvm`/`--skip-run-with-dbg` are needed for the rest only because this sandbox has LLVM 18 (the `llvm-*.stdout` references are LLVM 11, typed vs opaque pointers) and no runtime-stacktrace build; both are covered by CI.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,10 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
+# GFortran treats each source byte of a kind 4 literal as one character,
+# LFortran (like LLVM Flang) decodes the UTF-8 source into code points.
+RUN(NAME char4_string_literal_01 LABELS llvm)
 
 #test for coarray
 RUN(NAME coarray_01 LABELS gfortran GFORTRAN_ARGS -fcoarray=single)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,8 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_io_01 LABELS gfortran llvm)
+RUN(NAME char4_io_02 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
 # GFortran treats each source byte of a kind 4 literal as one character,
 # LFortran (like LLVM Flang) decodes the UTF-8 source into code points.
--- a/integration_tests/char4_io_01.f90
+++ b/integration_tests/char4_io_01.f90
@@ -1,0 +1,73 @@
+program char4_io_01
+    implicit none
+    integer, parameter :: ucs4 = selected_char_kind("ISO_10646")
+    ! U+4F60 and U+597D: three UTF-8 bytes each, so a byte oriented write
+    ! would truncate or mangle them.
+    character(kind=ucs4, len=1), parameter :: ni = char(int(z'4F60'), ucs4)
+    character(kind=ucs4, len=1), parameter :: hao = char(int(z'597D'), ucs4)
+    character(kind=ucs4, len=6) :: greeting
+    character(len=80) :: record
+    integer :: u
+
+    greeting = ucs4_"ab" // ni // hao
+
+    ! Write the value out with ENCODING='UTF-8', then read the file back as
+    ! default characters so the exact bytes can be checked.
+    open(newunit=u, file="char4_io_01_data.txt", status="replace", &
+         encoding="UTF-8", action="write")
+    write(u, '(a)') greeting
+    write(u, '("<",a,">")') trim(greeting)
+    write(u, *) greeting
+    close(u)
+
+    open(newunit=u, file="char4_io_01_data.txt", status="old", action="read")
+
+    ! 'a' with no width writes all six characters: a, b, then the two
+    ! ideographs as three bytes each, then two trailing blanks.
+    read(u, '(a)') record
+    call check_utf8(record, 1, "plain a")
+
+    ! trim() drops the two blanks, so only four characters are written.
+    read(u, '(a)') record
+    if (record(1:1) /= "<") error stop 10
+    call check_utf8(record, 2, "trimmed a")
+    if (record(10:10) /= ">") error stop 11
+
+    ! List directed output writes the same bytes.
+    read(u, '(a)') record
+    call check_utf8(record, 1 + count_leading_blanks(record), "list directed")
+
+    close(u, status="delete")
+
+    print *, "ok"
+
+contains
+
+    integer function count_leading_blanks(s) result(n)
+        character(len=*), intent(in) :: s
+        n = 0
+        do while (n < len(s))
+            if (s(n + 1 : n + 1) /= " ") exit
+            n = n + 1
+        end do
+    end function
+
+    ! Checks that `s`, starting at `start`, holds the UTF-8 bytes of
+    ! "ab" // U+4F60 // U+597D.
+    subroutine check_utf8(s, start, what)
+        character(len=*), intent(in) :: s
+        integer, intent(in) :: start
+        character(len=*), intent(in) :: what
+        integer :: expected(8), i, got
+        ! 'a', 'b', E4 BD A0, E5 A5 BD
+        expected = [97, 98, 228, 189, 160, 229, 165, 189]
+        do i = 1, 8
+            got = ichar(s(start + i - 1 : start + i - 1))
+            if (got /= expected(i)) then
+                print *, "byte", i, "of", what, "is", got, "expected", expected(i)
+                error stop 1
+            end if
+        end do
+    end subroutine
+
+end program
--- a/integration_tests/char4_io_02.f90
+++ b/integration_tests/char4_io_02.f90
@@ -1,0 +1,51 @@
+program char4_io_02
+    implicit none
+    integer, parameter :: ucs4 = selected_char_kind("ISO_10646")
+    ! U+5E74, U+6708 and U+65E5: the Japanese year, month and day characters.
+    character(kind=ucs4, len=1), parameter :: nen = char(int(z'5e74'), ucs4)
+    character(kind=ucs4, len=1), parameter :: gatsu = char(int(z'6708'), ucs4)
+    character(kind=ucs4, len=1), parameter :: nichi = char(int(z'65e5'), ucs4)
+    character(kind=ucs4, len=40) :: line
+    character(kind=ucs4, len=:), allocatable :: grown
+    integer :: u
+
+    ! An internal WRITE into a kind 4 variable must store code units, so the
+    ! result can be measured and indexed as characters afterwards.
+    write(line, '(i0,a,i0,a,i0,a)') 2026, nen, 8, gatsu, 24, nichi
+    if (len_trim(line) /= 10) error stop 1
+    if (line(1:4) /= ucs4_"2026") error stop 2
+    if (ichar(line(5:5), kind=4) /= int(z'5e74')) error stop 3
+    if (line(6:6) /= ucs4_"8") error stop 4
+    if (ichar(line(7:7), kind=4) /= int(z'6708')) error stop 5
+    if (line(8:9) /= ucs4_"24") error stop 6
+    if (ichar(line(10:10), kind=4) /= int(z'65e5')) error stop 7
+
+    ! The same for a shorter record: the tail must be blank padded in
+    ! characters, not in bytes.
+    write(line, '(a,a,a)') nen, gatsu, nichi
+    if (len_trim(line) /= 3) error stop 8
+    if (ichar(line(2:2), kind=4) /= int(z'6708')) error stop 9
+
+    grown = repeat(nichi, 3)
+    if (len(grown) /= 3) error stop 10
+
+    ! Round trip through a UTF-8 file: what is written as code units comes
+    ! back as code units.
+    open(newunit=u, file="char4_io_02_data.txt", status="replace", &
+         encoding="UTF-8", action="write")
+    write(u, '(a)') nen // gatsu // nichi
+    close(u)
+
+    line = ucs4_" "
+    open(newunit=u, file="char4_io_02_data.txt", status="old", &
+         encoding="UTF-8", action="read")
+    read(u, '(a)') line
+    close(u, status="delete")
+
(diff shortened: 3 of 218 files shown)


certik:
Pushed a correction. Two classes of failure that my earlier local runs could not see, both now fixed and verified against a build that matches CI exactly (Debug + LLVM 11 from conda-forge, the configuration `./run_tests.py` actually runs under — the reference suite only runs on `ubuntu` + `llvm-version: 11`).

**1. LLVM IR references (98 files).** The serialization string and `_lfortran_printf` signature change, so every IR reference that prints anything moved:

```
- @serialization_info = ... c"S-DESC-31,R4\00"
+ @serialization_info = ... c"S-DESC-K1-31,R4\00"
- declare void @_lfortran_printf(i8*, i8*, i32, i8*, i32)
+ declare void @_lfortran_printf(i8*, i8*, i32, i32, i8*, i32)
- @0 = ... c"_lcompilers_stringconcat\00"
+ @0 = ... c"_lcompilers_stringconcat_1\00"
```

I could not regenerate these with my LLVM 18 sandbox build (opaque `ptr` vs typed `i8*`), so I built against LLVM 11, confirmed that build reproduces **the entire reference suite green on `main`**, and only then regenerated. Every changed line in the 98 files falls into the three patterns above.

**2. Three ASR passes silently dropped the character kind**, caught by the `StringConcat` operand-kind requirement from [#12582](https://github.com/lfortran/lfortran/pull/12582) once a Debug build enabled ASR verification:

- `nested_vars.cpp:691,714` — a host-associated character variable was rewritten to a deferred-length string with `kind` hardcoded to 1, so `character(kind=4)` accessed from a contained procedure became kind 1. Minimal case: `u_"  " // ni` inside a contained subroutine where `ni` is a host kind-4 parameter.
- `select_case.cpp:156,253` — the temporary holding a `SELECT CASE` selector.
- `implied_do_loops.cpp:1183,1204` — the accumulator for implied-do concatenation.

Also two sites I had missed in the intrinsics: `instantiate_Adjustl`/`instantiate_Adjustr` recompute the *call-site* return type at the end of the function and still hardcoded kind 1, and `eval_NewLine` ignored the passed type and built a kind-1 constant.

All are real kind-losing rewrites that would have miscompiled kind-4 values; none are reachable with kind 1, which is why they were invisible before.

**Verified at every head of the stack** (Debug, LLVM 11): full reference suite including all `run =` and `llvm` IR tests, `-b llvm`, and `-b gfortran`.

---

certik:
CI's remaining red was `char4_string_intrinsics_02` failing at run time under `-b llvm` on Linux and macOS — but passing on Windows, and passing here in Debug+LLVM 11, Release+LLVM 18, with `WITH_INTERNAL_ALLOC_CHECK=yes`, and from a clean build tree.

The bug is in my test, not the compiler. It used four variables before defining them:

```fortran
character(kind=ucs4, len=6) :: padded
character(kind=ucs4, len=3) :: a, b
...
if (kind(trim(padded)) /= 4) error stop 1
if (kind(repeat(a, 2)) /= 4) error stop 4
if (kind(a // b) /= 4) error stop 5
```

`KIND` is an inquiry function, so an undefined argument is standard-conforming, but LFortran still evaluates the argument expression — so `trim`, `repeat` and `//` scan uninitialized kind-4 storage. That is environment-dependent by construction, which matches a failure that reproduces on CI's Linux and macOS runners and nowhere I can reach.

Fixed by giving each variable a value before the inquiries (the later redundant assignments are removed, so the diff is a move rather than an addition). Passes on GFortran 13, LLVM Flang 18 and this branch.

I want to be straight about the confidence here: I could not reproduce the failure locally, including under `MALLOC_PERTURB_`, so this is a well-motivated fix for a real defect rather than a confirmed root cause. If CI stays red I will keep digging. Re-verified at the top of the stack, clean build trees: full reference suite (all `run =` and IR tests), `-b llvm`, `-b gfortran`.

---

certik:
@Jainam-not-a-robot looks like the CI will pass here. I think this PR's implementation is the way to go, it's what most other languages do, as well as Flang (but not GFortran).

certik:
I reviewed everything, I think this is good. If CI passes, we can merge it.

certik:
Merged.

PR #12582, opened 2026-08-25
Reject unsupported character kinds and mixed-kind operands

certik:
## Summary

Second of three towards [#12574](https://github.com/lfortran/lfortran/issues/12574). **Depends on [#12581](https://github.com/lfortran/lfortran/pull/12581)** — its two commits are the parents here, so only `24d8368d` belongs to this PR. [#12584](https://github.com/lfortran/lfortran/pull/12584) stacks on this one.

The dependency is real, not just ordering: the mixed-kind checks below fire on valid code until the character intrinsics stop dropping the kind from their results. On `main`, `trim(word)` returns kind 1 for a kind 4 `word`, so `integration_tests/string_120.f90` (`trim(word) /= 4_"x"`) would be rejected by this PR's comparison check.

## Scope

**Unsupported kinds.** `character(kind=3)` was accepted silently and lowered as if the kind were meaningful. Kinds are now restricted to 1 and 4, on a declaration and on a kind-prefixed literal alike, and `asr_verify` requires the same of every `String` node — a check the numeric types have had at `asr_verify.cpp:2607` onwards but `visit_String` did not.

**Mixed-kind operands.** F2023 requires both operands of `//` and of a character comparison to have the same kind. Both were accepted silently. `libasr` also requires equal operand kinds for `StringConcat` in `check_args`, so the invariant does not rest on the frontend alone.

Measured on the reference compilers, which agree on all of this:

| | `main` | GFortran 13 | Flang 18 |
| --- | --- | --- | --- |
| `character(kind=2)` | accepted | error | accepted (UCS-2) |
| `character(kind=3)` / `(kind=8)` | accepted | error | error |
| `ucs4 // default` | accepted | error | error |
| `ucs4 == default` | accepted | error | error |
| `ucs4 = default` (assignment) | accepted | accepted | accepted |
| kind 4 internal file | accepted | accepted | accepted |

Assignment across kinds stays legal, and kind 4 internal files stay legal.

Kind 2 follows GFortran here rather than Flang: LFortran's `selected_char_kind` can only ever return 1, 4 or −1, so accepting 2 would be a kind no program can name portably. The backend paths are written in terms of the kind rather than special-cased, so allowing it later is one entry in `ASRUtils::is_supported_character_kind`.

Error wording follows the project style (lowercase, explicit kinds, no internal node names):

```
semantic error: kind 3 is not supported for character, only 1 and 4 are
semantic error: operands of // must be character with the same kind, found character(4) and character(1)
semantic error: operands of comparison operator '==' must be character with the same kind, found character(4) and character(1)
```

## Tests

Five cases appended to `tests/errors/continue_compilation_1.f90` with the reference updated: kinds 2, 3 and 8 on a declaration, kind 3 on a literal prefix, and mixed-kind `//` and `==`.

The rejected declarations are deliberately placed **last** in that file. A rejected declaration makes the symbol table visitor skip the program units after it, so putting them earlier swallows the errors the other cases are there to check. There is a comment in the test file saying so.

## Verification

```
./run_tests.py --skip-run-with-dbg -t <each of the 117 `run =` tests>   # all pass
./run_tests.py --no-llvm --skip-run-with-dbg                           # TESTS PASSED
cd integration_tests && ./run_tests.py -b llvm -j4                     # pass
cd integration_tests && ./run_tests.py -b gfortran -j4                 # pass
```

The `run =` reference tests are enumerated explicitly because `--no-llvm` skips them. `--no-llvm`/`--skip-run-with-dbg` are needed for the rest only because this sandbox has LLVM 18 (the `llvm-*.stdout` references are LLVM 11) and no runtime-stacktrace build; both are covered by CI.

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,10 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
+# GFortran treats each source byte of a kind 4 literal as one character,
+# LFortran (like LLVM Flang) decodes the UTF-8 source into code points.
+RUN(NAME char4_string_literal_01 LABELS llvm)
 
 #test for coarray
 RUN(NAME coarray_01 LABELS gfortran GFORTRAN_ARGS -fcoarray=single)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,8 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_io_01 LABELS gfortran llvm)
+RUN(NAME char4_io_02 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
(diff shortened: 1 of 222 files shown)


certik:
Merged.

PR #12583, opened 2026-08-25
Write and read kind 4 character values as UTF-8

### LINKED ISSUE #12574: Unicode ISO-10646 support not working in version 0.64.0

certik:
Fixes [#12574](https://github.com/lfortran/lfortran/issues/12574).

Third of a short series. **Depends on [#12581](https://github.com/lfortran/lfortran/pull/12581) and [#12582](https://github.com/lfortran/lfortran/pull/12582)** — they are the parent commits, so only the third commit (`404e9345`) belongs to this PR.

## Summary

A kind 4 `CHARACTER` value is held in memory as four-byte code units, but every output path treated the descriptor's *character* count as a *byte* count. So `write (*,*) trim(hello_world)` emitted the first byte of the record and stopped at the first embedded null — the `H` in the issue report. Internal `WRITE` into a kind 4 variable stored kind 1 bytes, and a formatted `READ` copied raw bytes into code units.

The issue's program, before and after, alongside the reference compilers:

```
main:      H
           2026t8^H24M-oM-?M-=

branch:    Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
           2026M-eM-9M-48M-fM-^\M-^H25M-fM-^WM-%

gfortran:  Hello World and Ni Hao -- M-dM-=M- M-eM-%M-=
           2026M-eM-9M-48M-fM-^\M-^H25M-fM-^WM-%
```

Byte-identical to GFortran 13 (and to LLVM Flang 18).

## Scope

**The character kind reaches the runtime.** `SerializeType` emits it as `S-DESC-K<kind>`, alongside the kind every other type already carried (`I4`, `R8`, `L32`) — strings were the only type whose kind was dropped. The format engine transcodes a kind 4 argument to UTF-8 when it consumes it, and counts `Aw` field widths in characters, as the standard requires.

Widths are counted in characters rather than copied from GFortran here: GFortran's `Aw` for kind 4 is demonstrably wrong (`a10` on a six-character value emits ten blanks, `a3` and `a2` both emit all six characters), so it is not a usable oracle for that one case. The new tests only assert behaviour all three compilers agree on.

**The two fast paths that bypass the format engine** for a lone character value carry the kind too: `_lfortran_printf` takes it as a parameter, and `_lfortran_file_write` learns `%S`, which reads a kind after the length. Without this, `print *, s` and `write(*,*) s` still wrote raw code units.

**Internal writes.** `_lfortran_string_write` takes the internal file's character kind and decodes the record back into code units for a kind 4 target, blank-padding in characters. It also drops a `sprintf("%.*s")` staging step that truncated the record at the first null.

**Reads.** A destination of kind > 1 uses read type code 8 and carries its kind. The field is decoded from UTF-8 when the connection has `ENCODING='UTF-8'`, and otherwise one byte becomes one character — `InputSource` gains the connection's `encoding_mode` for this. Both reference compilers agree exactly here, with and without the specifier.

**`ENCODING=` on output follows Flang**, per the design decision in the plan: kind 4 output is transcoded to UTF-8 regardless of the unit's encoding. GFortran instead substitutes `?` on a non-UTF-8 unit. Following Flang keeps the write path free of any per-unit encoding lookup and still gives the issue's program the output it expects. `ENCODING=` was already parsed, validated and stored per unit (`unit_to_file[].encoding`); before this PR its only consumer was `INQUIRE`, and it now also drives decoding on input.

## Tests

All self-checking with `error stop`, labelled `gfortran llvm`, and passing on GFortran 13, LLVM Flang 18 and this branch.

Integration tests can't compare stdout — the runner only checks the exit status — so byte-exactness is verified by writing to a scratch file with `encoding='UTF-8'`, reopening it as default `character`, and checking the bytes with `ichar`.

- `char4_io_01.f90` — `'(a)'`, `'("<",a,">")'` and list-directed output of a kind 4 value, each checked against the expected UTF-8 bytes.
- `char4_io_02.f90` — the issue's internal-write case (`write(line,'(i0,a,i0,a,i0,a)') 2026, nen, …`) checked character by character, blank padding of a short record, and a write/read round trip through a UTF-8 file.
- `print_12.f90` already exercised this path but asserted nothing, so it passed on `main` while printing garbage. It now checks the length and the bytes that reach the stream.

## Verification

```
./run_tests.py --no-llvm --skip-run-with-dbg    # TESTS PASSED
cd integration_tests && ./run_tests.py -b llvm -j4       # pass
cd integration_tests && ./run_tests.py -b gfortran -j4   # pass
```

`--no-llvm --skip-run-with-dbg` only because this sandbox has LLVM 18 (the `llvm-*.stdout` references are LLVM 11) and no runtime-stacktrace build; both are covered by CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sa9WpeKxiSd3kV1VKZXCUd)_

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,10 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
+# GFortran treats each source byte of a kind 4 literal as one character,
+# LFortran (like LLVM Flang) decodes the UTF-8 source into code points.
+RUN(NAME char4_string_literal_01 LABELS llvm)
 
 #test for coarray
 RUN(NAME coarray_01 LABELS gfortran GFORTRAN_ARGS -fcoarray=single)
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -4777,6 +4777,8 @@ RUN(NAME merge_str_01 LABELS gfortran llvm)
 RUN(NAME io_direct_slash LABELS gfortran llvm)
 RUN(NAME achar_kind_01 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_01 LABELS gfortran llvm)
+RUN(NAME char4_io_01 LABELS gfortran llvm)
+RUN(NAME char4_io_02 LABELS gfortran llvm)
 RUN(NAME char4_string_intrinsics_02 LABELS gfortran llvm)
 # GFortran treats each source byte of a kind 4 literal as one character,
 # LFortran (like LLVM Flang) decodes the UTF-8 source into code points.
--- a/integration_tests/char4_io_01.f90
+++ b/integration_tests/char4_io_01.f90
@@ -1,0 +1,73 @@
+program char4_io_01
+    implicit none
+    integer, parameter :: ucs4 = selected_char_kind("ISO_10646")
+    ! U+4F60 and U+597D: three UTF-8 bytes each, so a byte oriented write
+    ! would truncate or mangle them.
+    character(kind=ucs4, len=1), parameter :: ni = char(int(z'4F60'), ucs4)
+    character(kind=ucs4, len=1), parameter :: hao = char(int(z'597D'), ucs4)
+    character(kind=ucs4, len=6) :: greeting
+    character(len=80) :: record
+    integer :: u
+
+    greeting = ucs4_"ab" // ni // hao
+
+    ! Write the value out with ENCODING='UTF-8', then read the file back as
+    ! default characters so the exact bytes can be checked.
+    open(newunit=u, file="char4_io_01_data.txt", status="replace", &
+         encoding="UTF-8", action="write")
+    write(u, '(a)') greeting
+    write(u, '("<",a,">")') trim(greeting)
+    write(u, *) greeting
+    close(u)
+
+    open(newunit=u, file="char4_io_01_data.txt", status="old", action="read")
+
+    ! 'a' with no width writes all six characters: a, b, then the two
+    ! ideographs as three bytes each, then two trailing blanks.
+    read(u, '(a)') record
+    call check_utf8(record, 1, "plain a")
+
+    ! trim() drops the two blanks, so only four characters are written.
+    read(u, '(a)') record
+    if (record(1:1) /= "<") error stop 10
+    call check_utf8(record, 2, "trimmed a")
+    if (record(10:10) /= ">") error stop 11
+
+    ! List directed output writes the same bytes.
+    read(u, '(a)') record
+    call check_utf8(record, 1 + count_leading_blanks(record), "list directed")
+
+    close(u, status="delete")
+
+    print *, "ok"
+
+contains
+
+    integer function count_leading_blanks(s) result(n)
+        character(len=*), intent(in) :: s
+        n = 0
+        do while (n < len(s))
+            if (s(n + 1 : n + 1) /= " ") exit
+            n = n + 1
+        end do
+    end function
+
+    ! Checks that `s`, starting at `start`, holds the UTF-8 bytes of
+    ! "ab" // U+4F60 // U+597D.
+    subroutine check_utf8(s, start, what)
+        character(len=*), intent(in) :: s
+        integer, intent(in) :: start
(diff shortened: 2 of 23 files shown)


certik:
Closing: folded into [#12581](https://github.com/lfortran/lfortran/pull/12581), which now carries both commits.

CI caught why they cannot be separate. `integration_tests/print_12.f90` is also a `run =` reference test, so its stdout is compared byte for byte. On `main` its concatenation folds to kind 1, so the value's UTF-8 bytes print correctly *by accident*, and the stored reference is correct UTF-8. #12581's first commit types that result as kind 4 — correctly — which makes it print raw UCS-4 until the output path in this PR understands the typing:

```
main:       Hello = Ni Hao = M-dM-=M- M-eM-%M-=
#12581:     H^@^@^@e^@^@^@l^@^@^@l^@^@^@o^@^@
both:       Hello = Ni Hao = M-dM-=M- M-eM-%M-=     (identical to main's reference)
```

Merging the ASR commit alone would regress that output on `main`, so the two are one change. Nothing is lost — the commit is unchanged in #12581, and [#12582](https://github.com/lfortran/lfortran/pull/12582) and [#12584](https://github.com/lfortran/lfortran/pull/12584) have been restacked on top of it.

---

PR #12586, opened 2026-08-26
fix: resolve complex part designator on a derived type component

### LINKED ISSUE #1001: complex designators
LFortran produces errors at compile time for both tests. Test I is a simple test, Test II fails with several compilers but I believe is a correct test per the f2018 standard.
```fortran
program main
implicit none
complex :: x 
real,parameter:: y=sin(3.0)

   ! TEST I 
   ! x%re|mi can appear on the left-hand side of an assignment 
   x%re=100.0 
   x%im=sin(3.0) 
   write(*,*)'LHS:',x,x.eq.cmplx(100.0,y)

   ! TEST II 
   ! it can be passed as a subroutine argument and be changed 
   call trivial(x%re) 
   write(*,*)'RETURNED:',x,x.eq.(200.0,y)

contains 

subroutine trivial(r) 
real,intent(inout) :: r 
   r=r*2
end subroutine trivial

end program main
```

certik:
## Summary

`d%c%re` — a complex part designator applied to a derived type component — crashed the compiler with a segmentation fault during AST→ASR.

`visit_NameUtil` resolves a multi-part designator left to right, then resolves the trailing component and blindly down-casts the result to `StructInstanceMember`. That holds only when the last component really is a member of a derived type. When the base is an intrinsic type, `resolve_variable2` returns a `ComplexRe`/`ComplexIm`/`Kind`/`StringLen` node instead, and the down-cast read past the node — `import_struct_instance_member` then dereferenced garbage as a symbol name.

The fix resolves the trailing component against the type of the already-built base expression: when that type is not a derived or union type, the inquiry node is built directly from the base expression. This keeps the decision in semantics (AST→ASR) and leaves the backends untouched.

## Scope

- `src/lfortran/semantics/ast_common_visitor.h`: new `resolve_intrinsic_type_member` helper, called from `visit_NameUtil` before the struct-member path.

The same code path served `%kind` and `%len` on a component, so those are fixed by the same change:

| designator | before | after |
| --- | --- | --- |
| `d%c%re`, `d%c%im` | SIGSEGV | works |
| `d%s%len` | SIGSEGV | works |
| `d%i%kind`, `d%c%kind` | `std::exception: basic_string: construction from null is not valid` | works |
| `d%i%bogus` | crash | `semantic error: Variable 'i' doesn't have any member named, 'bogus'.` |

Covered: read and assign on both parts, `INTENT(INOUT)` actual arguments, `complex(8)`, complex array components, array-of-derived-type elements, allocatable components, nested derived types, and module-defined types.

## Verification

Two integration tests, both registered in `integration_tests/CMakeLists.txt` with `gfortran llvm`:

- `integration_tests/complex_38.f90` — complex part designators on components
- `integration_tests/derived_types_155.f90` — `%kind` / `%len` on components

Fails on `main`, passes on this branch (verified by rebuilding with the source change stashed):

```
$ lfortran integration_tests/complex_38.f90        # on main
Segmentation fault                        (exit 139)
$ lfortran integration_tests/derived_types_155.f90 # on main
std::exception: basic_string: construction from null is not valid   (exit 1)

$ lfortran integration_tests/complex_38.f90 -o a.out && ./a.out        # this branch
PASS
$ lfortran integration_tests/derived_types_155.f90 -o a.out && ./a.out # this branch
PASS
```

Both programs compile and run correctly under `gfortran` as well.

Suites:

```
$ cd integration_tests && ./run_tests.py -j4
100% tests passed, 0 tests failed out of 3893

$ ./run_tests.py --no-llvm
TESTS PASSED
```

The `llvm` reference outputs could not be checked in this environment: it has LLVM 18, which prints opaque pointers (`ptr`), while the project pins `llvmdev=11`, whose reference outputs use typed pointers (`i8*`, `float*`). Every difference seen was that spelling change alone, in tests unrelated to this fix; CI runs on the pinned LLVM and covers them.

## Rationale

Reported in #1001. The program in that issue compiles and runs correctly on `main` today, so the fix here targets the part of "complex designators" that is still broken: the same designators applied to a derived type component. Placing the fix in AST→ASR keeps the type decision in semantics, per the architecture rules in `AGENTS.md`.

Fixes #1001

---
_Generated by [Claude Code](https://claude.ai/code/session_01N4ridD8pPKn6nkZV4L6Nb8)_


src/lfortran/semantics/ast_common_visitor.h
```diff
@@ -13523,6 +13491,69 @@ class CommonVisitor : public AST::BaseVisitor<Derived> {
...
+            if (member_name != "len") {
+                diag.add(Diagnostic("Character variable '" + base_name + "' only has %len and %kind members, not '" + member_name + "'",
+                    Level::Error, Stage::Semantic, {Label("", {loc})}));
+                throw SemanticAbort();
+            }
+            return create_StringLen_from_expr(base, int32, loc);
+        }
+        diag.add(Diagnostic("Variable '" + base_name + "' doesn't have any member named, '" + member_name + "'.",
+            Level::Error, Stage::Semantic, {Label("", {loc})}));
+        throw SemanticAbort();
```
certik, line 13554: We should add tests for all these new error messages here.

PR #12587, opened 2026-08-26
fix: give character MAXVAL/MINVAL a reduction identity

### LINKED ISSUE #12344: MAXVAL fails when argument is a zero-length string
This is a dusty corner, but  what MAXVAL returns for
zero-sized arrays and character strings of zero length is defined in the standard but this MRE gets an ICE. The original demo of MAXVAL is also attached as it makes additional tests after the failure point that will hopefully pass once the initial ICE is resolved.

```fortran
     1	program mre
     2	implicit none
     3	character(len=:),allocatable :: strs(:)
     4	   if(allocated(strs))deallocate(strs)
     5	   allocate(character(len=0) :: strs(5))
     6	   print '(*(g0,1x))', 'is maxval of null length strings a null character? ',ichar(maxval(strs))==0
     7	end program mre
```
```text
+ cexpl -c lfortran0590 --skip-asm '--cflags=--no-style-suggestions --realloc-lhs-arrays --cpp -DCOMPILER_=lfortran0590' xx.f90
lfortran0590         :     1	STDERR:
lfortran0590         :     2	warning: `--generate-object-code` is deprecated and will be removed in a future release; use `--separate-compilation` instead.
lfortran0590         :     3	Internal Compiler Error: Unhandled exception
lfortran0590         :     4	Traceback (most recent call last):
lfortran0590         :     5	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x70a2c4
lfortran0590         :     6	  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x2a28a
lfortran0590         :     7	  Binary file "/lib/x86_64-linux-gnu/libc.so.6", local address: 0x2a1c9
lfortran0590         :     8	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x6886bb
lfortran0590         :     9	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x7a1b0b
lfortran0590         :    10	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x793634
lfortran0590         :    11	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x9379f9
lfortran0590         :    12	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x168a1d0
lfortran0590         :    13	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x16835c4
lfortran0590         :    14	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x1681a96
lfortran0590         :    15	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x1247bab
lfortran0590         :    16	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x12474b3
lfortran0590         :    17	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x1246b73
lfortran0590         :    18	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x128348b
lfortran0590         :    19	  Binary file "/cefs/75/756226c57544077aa239d172_consolidated/compilers_fortran_lfortran_0.59.0/bin/lfortran", local address: 0x1264f58
lfortran0590         :    20	LCompilersException: get_minimum_value_with_given_type: Not implemented 4
```
## demo_maxval.f90
```fortran
      program demo_maxval
      implicit none
      integer,save                 :: ints(3,5)= reshape([&
         1,  2,  3, -4,  5, &
        10, 20,-30, 40, 50, &
        11,-22, 33, 44, 55  &
      ],shape(ints),order=[2,1])
      character(len=:),allocatable :: strs(:)
      character(len=:),allocatable :: answer
      integer                      :: i
      character(len=*),parameter   :: gen='(*(g0,1x))'
      character(len=*),parameter   :: ind='(3x,*(g0,1x))'

         print gen,'Given the array'
         write(*,'(1x,*(g4.4,1x))') &
         & (ints(i,:),new_line('a'),i=1,size(ints,dim=1))
         print gen,'Basics:'
         print ind, 'biggest value in array'
         print ind, maxval(ints)
         print ind, 'biggest value in each column'
         print ind, maxval(ints,dim=1)
         print ind, 'biggest value in each row'
         print ind,  maxval(ints,dim=2)
         print ind
         print ind, 'find biggest number less than 30 with mask'
         print ind
         print ind, 'find biggest negative value'
         print ind, '(closest to zero, not biggest magnitude)'
         print ind, maxval(ints,mask=ints.lt.0)
         print ind
         print ind, 'DEALING WITH ZERO-LENGTH STRINGS AND ZERO-SIZE ARRAYS'
         print ind
         print ind, 'if zero size numeric array:'
         print ind, maxval([integer :: ]),'and -huge(0) is',-huge(0),&
         & '(often not the same!)'
         print ind
         print ind, maxval([real :: ]),'and -huge(0.0) is',-huge(0.0)
         print ind
         print ind, 'if zero-size character array all nulls'
         if(allocated(strs))deallocate(strs)
         allocate(character(len=0) :: strs(5))
         print ind, 'STRS() has a length of:', len(strs), &
          & 'a SHAPE of:',shape(strs), &
          & ':a SIZE of:',size(strs)
         print ind, 'is maxval of null length strings a null character? ',ichar(maxval(strs))==0
         print ind
         if(allocated(strs))deallocate(strs)
         allocate(character(len=5) :: strs(0))
         print ind, 'STRS() has a length of:', len(strs), &
          & 'a SHAPE of:',shape(strs), &
          & ':a SIZE of:',size(strs)
          answer=maxval(strs)
         print ind, 'is maxval of strings all null characters? ', &
          & [(answer(i:i),i=1,len(answer))].eq.char(0)
         print ind
         print ind, 'if everything in mask is false,'
         print ind, 'same as zero-size array for each subarray'
         print ind, maxval(ints,mask=.false.)
         print ind, maxval(ints,mask=.false.,dim=1)
      end program demo_maxval
```

certik:
## Summary

Fixes #12344.

`MAXVAL`/`MINVAL` of a character array had no identity value, so the generated reduction seeded its result from element 1 of the array. That reads out of bounds for a zero sized array, and for a deferred length array it declares a `DeferredLength` string variable that is neither allocatable nor a pointer, which ASR verification rejects. Before that seed was introduced the same gap surfaced as the `get_minimum_value_with_given_type: Not implemented 4` ICE reported in the issue — it is still reachable on `main` through `maxval(character_array, dim=1)`.

The reduction is now seeded with the identity the standard defines: a string as long as the elements of the array with every character equal to `char(0)` for `MAXVAL` and `char(255)` for `MINVAL` (Fortran 2018, 16.9.126 and 16.9.135). Every element compares greater than the first and smaller than the second, so a non empty array reduces exactly as before, while an empty one (zero sized, or fully masked out) yields the identity itself. When the element length is only known at runtime the result is allocated to `len(array)` and filled in a loop.

The MRE from the issue now agrees with GFortran:

```fortran
program mre
implicit none
character(len=:), allocatable :: strs(:)
allocate(character(len=0) :: strs(5))
if (len(maxval(strs)) /= 0) error stop
end program mre
```

The `demo_maxval.f90` attached to the issue also compiles and runs, and its `is maxval of strings all null characters?  T T T T T` line matches GFortran.

## Scope

The fix is in AST→ASR/ASR (the intrinsic array function registry and the ASR utilities); the backends only lower what they are given. Two supporting fixes it needs:

- `str_compare` in the runtime, and the single character comparison in the LLVM backend, ordered characters as **signed** values, so everything from `char(128)` up compared below `char(0)`. Characters are ordered by their position in the collating sequence, so they are now compared as unsigned. Without this the `MINVAL` identity of `char(255)` would never be beaten by a real element. Compile time folding already compared unsigned, so this also removes an inconsistency between constant folding and runtime.
- `types_equal` compares only the *kind* of two string types, so a reduction generated for one element length was silently reused for another (`maxval` on a `character(len=3)` array and then on a `character(len=:), allocatable` one returned the first helper). The element length is now part of the generated function's name.

Not addressed here, and unchanged by this PR: `maxval`/`minval` of a character array with `dim=` or `mask=` still fail to compile — the `dim`/`mask` body generators have no string handling at all, and `verify_args` rejects character input for these intrinsics. That is a separate gap; no working program regresses.

## Verification

New integration test `integration_tests/empty_array_05.f90` (labels `gfortran llvm`), the character counterpart to the existing `empty_array_03.f90`. It covers a non empty character array, a zero sized one, a zero sized array constructor, a deferred length array allocated with size zero, deferred length elements of zero length, and a deferred length array with non zero length elements.

Test fails on `main` and passes on this branch:

```
$ git stash && cmake --build build -j && ./build/src/bin/lfortran integration_tests/empty_array_05.f90
LCompilersException: get_minimum_value_with_given_type: Not implemented 4

$ git stash pop && cmake --build build -j && ./build/src/bin/lfortran integration_tests/empty_array_05.f90 -o t && ./t
 Pass
```

Suites run locally (Ubuntu, LLVM 18):

```
./run_tests.py --no-llvm                            TESTS PASSED
cd integration_tests && ./run_tests.py -b llvm      100% tests passed, 0 failed out of 3892
cd integration_tests && ./run_tests.py -b llvm -f -nf16   100% tests passed, 0 failed out of 3892
cd integration_tests && ./run_tests.py -b gfortran  100% tests passed, 0 failed out of 3951
```

The `llvm` reference outputs could not be checked locally — they are pinned to CI's LLVM 11 and LLVM 18 emits opaque pointers (`ptr` vs `i8**`) throughout. The `c`/`cpp` integration backends need `lfortran_intrinsics.h` on the include path and Kokkos, which this environment lacks. Both are covered by the exhaustive CI run requested on this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JSLVVoJokrwXCo5XWGUWrR)_

### DIFF
--- a/src/libasr/asr_utils.h
+++ b/src/libasr/asr_utils.h
@@ -2830,6 +2830,20 @@ static inline ASR::expr_t* get_constant_one_with_given_type(Allocator& al, ASR::
 
 void mark_modules_as_external(const LCompilers::ASR::TranslationUnit_t &u);
 
+// Returns a string as long as `string_type` with every character equal to
+// `fill`, or nullptr when that length is not known at compile time.
+static inline ASR::expr_t* get_string_filled_with_char(Allocator& al,
+        ASR::ttype_t* string_type, unsigned char fill) {
+    ASR::String_t* str_type = ASR::down_cast<ASR::String_t>(string_type);
+    int64_t len = -1;
+    if (!str_type->m_len || !ASRUtils::extract_value(str_type->m_len, len) || len < 0) {
+        return nullptr;
+    }
+    std::string value((size_t) len, (char) fill);
+    return ASRUtils::EXPR(ASR::make_StringConstant_t(al, string_type->base.loc,
+        s2c(al, value), string_type));
+}
+
 static inline ASR::expr_t* get_minimum_value_with_given_type(Allocator& al, ASR::ttype_t* asr_type) {
     asr_type = ASRUtils::type_get_past_array(asr_type);
     int kind = ASRUtils::extract_kind_from_ttype_t(asr_type);
@@ -2855,6 +2869,17 @@ static inline ASR::expr_t* get_minimum_value_with_given_type(Allocator& al, ASR:
                     throw LCompilersException("get_minimum_value_with_given_type: Unsupported real kind " + std::to_string(kind));
             }
             return ASRUtils::EXPR(ASR::make_RealConstant_t(al, asr_type->base.loc, val, asr_type));
+        }
+        case ASR::ttypeType::String: {
+            // The smallest character value is char(0), the first character of
+            // the collating sequence (Fortran 2018, 16.9.126 requires MAXVAL of
+            // a zero sized character array to be a string of char(0))
+            ASR::expr_t* min_value = get_string_filled_with_char(al, asr_type, 0);
+            if (!min_value) {
+                throw LCompilersException("get_minimum_value_with_given_type: "
+                    "string length is not known at compile time");
+            }
+            return min_value;
         }
         default: {
             throw LCompilersException("get_minimum_value_with_given_type: Not implemented " + std::to_string(asr_type->type));
@@ -2888,6 +2913,23 @@ static inline ASR::expr_t* get_maximum_value_with_given_type(Allocator& al, ASR:
                     throw LCompilersException("get_maximum_value_with_given_type: Unsupported real kind " + std::to_string(kind));
             }
             return ASRUtils::EXPR(ASR::make_RealConstant_t(al, asr_type->base.loc, val, asr_type));
+        }
+        case ASR::ttypeType::String: {
+            // The largest character value is char(n - 1), the last character of
+            // the collating sequence, where n is the number of characters
+            // representable with the given kind (Fortran 2018, 16.9.135
+            // requires MINVAL of a zero sized character array to be a string
+            // of that character)
+            if (kind != 1) {
+                throw LCompilersException("get_maximum_value_with_given_type: "
+                    "Unsupported character kind " + std::to_string(kind));
+            }
+            ASR::expr_t* max_value = get_string_filled_with_char(al, asr_type, 0xff);
+            if (!max_value) {
+                throw LCompilersException("get_maximum_value_with_given_type: "
+                    "string length is not known at compile time");
+            }
+            return max_value;
         }
         default: {
             throw LCompilersException("get_maximum_value_with_given_type: Not implemented " + std::to_string(asr_type->type));
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -3289,6 +3289,7 @@ RUN(NAME empty_array_02 LABELS gfortran llvm)
 
 RUN(NAME empty_array_03 LABELS gfortran llvm)
 RUN(NAME empty_array_04 LABELS gfortran llvm)
+RUN(NAME empty_array_05 LABELS gfortran llvm)
 
 RUN(NAME kwargs_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm)
 RUN(NAME kwargs_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc)
--- a/integration_tests/empty_array_05.f90
+++ b/integration_tests/empty_array_05.f90
@@ -1,0 +1,65 @@
+program empty_array_05
+    implicit none
+
+    character(len=3) :: c(2), c0(0), r3
+    character(len=:), allocatable :: d(:), r
+    integer :: i
+
+    c(1) = 'abc'
+    c(2) = 'xyz'
+
+    ! a non empty character array reduces to one of its elements
+    if (maxval(c) /= 'xyz') error stop
+    if (minval(c) /= 'abc') error stop
+
+    ! MAXVAL of a zero sized character array is a string of char(0) and
+    ! MINVAL of one is a string of char(255)
+    r3 = maxval(c0)
+    do i = 1, 3
+        if (ichar(r3(i:i)) /= 0) error stop
+    end do
+    r3 = minval(c0)
+    do i = 1, 3
+        if (ichar(r3(i:i)) /= 255) error stop
+    end do
+
+    ! the same holds for a zero sized array constructor
+    r3 = maxval([character(len=3) :: ])
+    do i = 1, 3
+        if (ichar(r3(i:i)) /= 0) error stop
+    end do
+    r3 = minval([character(len=3) :: ])
+    do i = 1, 3
+        if (ichar(r3(i:i)) /= 255) error stop
+    end do
+
+    ! and for a deferred length array allocated with size zero
+    allocate(character(len=4) :: d(0))
+    r = maxval(d)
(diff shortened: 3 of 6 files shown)


src/libasr/asr_utils.h
```diff
@@ -2889,6 +2914,23 @@ static inline ASR::expr_t* get_maximum_value_with_given_type(Allocator& al, ASR:
...
+            // requires MINVAL of a zero sized character array to be a string
+            // of that character)
+            if (kind != 1) {
+                throw LCompilersException("get_maximum_value_with_given_type: "
+                    "Unsupported character kind " + std::to_string(kind));
+            }
+            ASR::expr_t* max_value = get_string_filled_with_char(al, asr_type, 0xff);
+            if (!max_value) {
+                throw LCompilersException("get_maximum_value_with_given_type: "
+                    "string length is not known at compile time");
```
certik, line 2930: We should not be raising LCompilersException, as that is an ICE.

Can these be triggered by constructing invalid ASR? If so, we must strengthen our ASR verify and add tests for this to not allow it.

And if it can't be triggered, then all these should be converted to just asserts.

certik: Addressed in lfortran/lfortran#12594. I checked each of these and they turned out to be three different things.

The two I added here were reachable from valid Fortran, not just from invalid ASR — `maxval(d, dim=1)` on a `character(len=:), allocatable :: d(:,:)` hits "string length is not known at compile time". They are gone: the constant identity is only needed for constant folding, so it moved into `eval_ArrIntrinsic`, where returning `nullptr` already means "evaluate at runtime". Neither helper needs a `String` case now.

The `dim` and `mask` overloads have no string handling in their body generators, which is what reached the internal error in the first place. They are now rejected in semantics with a diagnostic, so a character `dim`/`mask` reduction says it is not implemented yet instead of reporting an ICE, an LLVM verifier failure, or an ASR verify error depending on the argument. Error test added.

With those handled, the integer kind and the type switch really are invariants (semantics restricts integer kinds to 1, 2, 4 and 8, and MaxVal/MinVal only accept integer, real and character arrays), so they became asserts.

One I deliberately left as an exception: the unsupported **real** kind is not an invariant. `maxval` of a `real(16)` array reaches it from valid Fortran and raises `Unsupported real kind 16` on `main` today, so an assert would hide a genuine bug in release builds. It is unrelated to the character work, so I left it out of that PR — happy to fix it separately.

---

PR #12589, opened 2026-08-26
fix: use the longest argument's length for character max/min

### LINKED ISSUE #12354: MAX intrinsic returns incorrect results for character array arguments
Although results between platforms vary the example program contains the very statement
used as an example in the standard. The result for the character arrays should be
```text
["BB" "Z "]
```
Lfortran produces truncated strings and an incorrect value
```text
["B" "B"]
```
From the standard
```diff
-     9  16.9.135 MAX (A1, A2 [, A3, ...])
-    10  Description. Maximum value.
-    11  Class. Elemental function.
-    12  Arguments. The arguments shall all have the same type which shall be integer, real, or character and they shall
-    13  all have the same kind type parameter.
-
-    14  Result Characteristics. The type and kind type parameter of the result are the same as those of the arguments.
-
-    15  For arguments of character type, the length of the result is the length of the longest argument.
-
-    16  Result Value. The value of the result is that of the largest argument. For arguments of character type, the
-    17  result is the value that would be selected by application of intrinsic relational operators; that is, the collating
-    18  sequence for characters with the kind type parameter of the arguments is applied. If the selected argument is
-    19  shorter than the longest argument, the result is extended with blanks on the right to the length of the longest
-    20  argument.
-
-    21  Examples. MAX (−9.0, 7.0, 2.0) has the value 7.0, MAX (’Z’, ’BB’) has the value ’Z ’, and MAX ([’A’, ’Z’],
-    22  [’BB’, ’Y ’]) has the value [’BB’, ’Z ’].
```
```fortran
     1	program demo_max
     2	implicit none
     3	
     4	   write(*,"(*(g0:,1x))")max([-2,2],[20,10])
     5	   write(*,"(*('""',g0,'""':,1x))")max(['A','Z'],['BB','Y '])
     6	
     7	end program demo_max
```
Only flang and ifx currently produce the expected results. 
```text
aoccflang520         :     2	20 10
aoccflang520         :     3	"B^@" "B^@"

flangtrunk           :     2	20 10
flangtrunk           :     3	"BB" "Z "

ifxlatest            :     2	20 10
ifxlatest            :     3	"BB" "Z "

lfortran0590         :     2	20 10
lfortran0590         :     3	"B" "B"

nvfortran_x86_26_3   :     1	STDERR:
nvfortran_x86_26_3   :     2	Lowering Error: unknown target type for ast conversion [ast=103,asttype=10,datatype=0]
nvfortran_x86_26_3   :     3	Lowering Error: target type was 22
nvfortran_x86_26_3   :     4	Lowering Error: unexpected result type for intrinsic function [ast=126,asttype=14,datatype=0]
nvfortran_x86_26_3   :     5	NVFORTRAN-F-0000-Internal compiler error. Errors in Lowering       3  (/app/example.f90: 7)
nvfortran_x86_26_3   :     6	NVFORTRAN/x86-64 Linux 26.3-0: compilation aborted

gfortran161          :     2	20 10
gfortran161          :     3	"B" "Z"
gfortran161          :     4	STDERR:
gfortran161          :     5	/app/example.f90:5:35:
gfortran161          :     6	
gfortran161          :     7	    5 |    write(*,"(*('""',g0,'""':,1x))")max(['A','Z'],['BB','Y '])
gfortran161          :     8	      |                                   1
gfortran161          :     9	Warning: CHARACTER expression at (1) is being truncated (2/1) [-Wcharacter-truncation]
gfortran161          :    10	/app/example.f90:5:35:
gfortran161          :    11	
gfortran161          :    12	    5 |    write(*,"(*('""',g0,'""':,1x))")max(['A','Z'],['BB','Y '])
gfortran161          :    13	      |                                   1
gfortran161          :    14	Warning: Legacy Extension: The CHARACTER elements of the array constructor at (1) must have the same length (2/1)
gfortran161          :    15	/app/example.f90:5:35:
gfortran161          :    16	
gfortran161          :    17	    5 |    write(*,"(*('""',g0,'""':,1x))")max(['A','Z'],['BB','Y '])
gfortran161          :    18	      |                                   1
gfortran161          :    19	Warning: CHARACTER expression at (1) is being truncated (2/1) [-Wcharacter-truncation]
gfortran161          :    20	/app/example.f90:5:35:
gfortran161          :    21	
gfortran161          :    22	    5 |    write(*,"(*('""',g0,'""':,1x))")max(['A','Z'],['BB','Y '])
gfortran161          :    23	      |                                   1
gfortran161          :    24	Warning: Legacy Extension: The CHARACTER elements of the array constructor at (1) must have the same length (2/1)
```

certik:
## Summary

Fixes #12354.

The standard (16.9.135 for `MAX`, 16.9.141 for `MIN`) requires the result of `MAX`/`MIN` with character arguments to have the length of the **longest** argument, and a shorter selected argument to be blank-padded on the right up to that length:

> For arguments of character type, the length of the result is the length of the longest argument. […] If the selected argument is shorter than the longest argument, the result is extended with blanks on the right to the length of the longest argument.
>
> `MAX(['A', 'Z'], ['BB', 'Y '])` has the value `['BB', 'Z ']`.

LFortran used the length of the **first** argument instead, so a longer selected argument was truncated.

MRE:

```fortran
program mre_max_char
implicit none
character(len=1) :: a(2)
character(len=2) :: b(2), r(2)
a(1) = 'A'
a(2) = 'Z'
b(1) = 'BB'
b(2) = 'Y '
r = max(a, b)
print *, "["//r(1)//"]["//r(2)//"]"
end program mre_max_char
```

| | output |
| --- | --- |
| gfortran | `[BB][Z ]` |
| lfortran (before) | `[B ][Z ]` |
| lfortran (after) | `[BB][Z ]` |

## Scope

`src/libasr/pass/intrinsic_functions.h`, `MAX` and `MIN`:

- **Semantics (`create_Max` / `create_Min`)** — the result type now carries the length of the longest argument instead of `expr_type(args[0])`. Constant lengths fold to the larger constant; otherwise an `IfExp` over the two length expressions is built. The array shape of the first argument is preserved.
- **`instantiate_Max` / `instantiate_Min`** — the helper function's result variable is declared with the longest dummy's length, so the existing `result = x(i)` assignment blank-pads a shorter selected argument instead of truncating a longer one. `instantiate_Min`'s call-site return type is computed the same way.
- **Compile-time folding (`eval_Max` / `eval_Min`)** — arguments are blank-padded to the longest length before being compared and returned, which is both what the collating sequence requires and what makes the folded constant carry the right length. This replaces the previous `strcmp` on unpadded values.

The fix is entirely in AST→ASR semantics and the instantiated ASR helper; no backend was touched.

## Verification

New integration test `integration_tests/minmax_02.f90` (labels `gfortran llvm`), covering arrays, scalars with the shorter and the longer argument first, more than two arguments, compile-time folding, and a runtime-length (`character(len=m)`) argument.

- Test fails on `main` (`ERROR STOP`, exit 1) and passes on this branch — verified by stashing the source change and rebuilding.
- `gfortran` compiles and runs the new test successfully, so it does not rely on LFortran-specific behavior.
- Integration suite: `cd integration_tests && ./run_tests.py -j4` → **100% tests passed, 0 tests failed out of 3892**.
- Reference suite: `./run_tests.py -u` produces **no** AST/ASR reference changes, and no reference diff touches the `max0`/`min0` helpers. (The local box has LLVM 18 while CI pins LLVM 11, so `llvm-*` references differ purely by opaque-pointer/register-numbering IR text; those were reverted and not committed.)

## Rationale

Per `AGENTS.md`, type and length decisions belong in AST→ASR rather than in codegen, so the result length is computed where the intrinsic node is built and simply lowered from there. `MIN` shares the same code path and the same defect as `MAX` — the standard's wording for both is identical — so both are fixed together rather than leaving `MIN` knowingly broken.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FGhZgYpF9RSBgd8UsXhK4i)_

### DIFF
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -561,6 +561,7 @@ RUN(NAME max_03 LABELS  llvm llvm_wasm llvm_wasm_emcc) #gfortran does not suppor
 RUN(NAME min_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME min_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME minmax_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc wasm c fortran)
+RUN(NAME minmax_02 LABELS gfortran llvm)
 RUN(NAME arithmetic_if_01 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran) # arithmetic tests use goto
 RUN(NAME arithmetic_if_02 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
 RUN(NAME arithmetic_if_03 LABELS gfortran llvm llvm_wasm llvm_wasm_emcc c fortran)
--- a/integration_tests/minmax_02.f90
+++ b/integration_tests/minmax_02.f90
@@ -1,0 +1,70 @@
+program minmax_02
+! The result of MAX/MIN with character arguments has the length of the longest
+! argument, and a shorter selected argument is blank-padded on the right.
+implicit none
+character(len=1) :: a(2)
+character(len=2) :: b(2), r(2)
+character(len=1) :: s1
+character(len=2) :: s2
+character(len=6) :: s3
+integer :: n
+
+! Arrays: max(['A','Z'], ['BB','Y ']) == ['BB', 'Z ']
+a(1) = 'A'
+a(2) = 'Z'
+b(1) = 'BB'
+b(2) = 'Y '
+r = max(a, b)
+if (r(1) /= 'BB') error stop
+if (r(2) /= 'Z ') error stop
+if ('[' // r(1) // ']' /= '[BB]') error stop
+if ('[' // r(2) // ']' /= '[Z ]') error stop
+r = min(a, b)
+if ('[' // r(1) // ']' /= '[A ]') error stop
+if ('[' // r(2) // ']' /= '[Y ]') error stop
+
+! Scalars, shorter argument first
+s1 = 'Z'
+s2 = 'BB'
+if (len(max(s1, s2)) /= 2) error stop
+if ('[' // max(s1, s2) // ']' /= '[Z ]') error stop
+if (len(min(s1, s2)) /= 2) error stop
+if ('[' // min(s1, s2) // ']' /= '[BB]') error stop
+
+! Scalars, longer argument first
+if (len(max(s2, s1)) /= 2) error stop
+if ('[' // max(s2, s1) // ']' /= '[Z ]') error stop
+if ('[' // min(s2, s1) // ']' /= '[BB]') error stop
+
+! More than two arguments
+s3 = 'cc'
+if (len(max(s1, s2, s3)) /= 6) error stop
+if ('[' // max(s1, s2, s3) // ']' /= '[cc    ]') error stop
+if ('[' // min(s1, s2, s3) // ']' /= '[BB    ]') error stop
+
+! Compile time
+if (len(max('Z', 'BB')) /= 2) error stop
+if ('[' // max('Z', 'BB') // ']' /= '[Z ]') error stop
+if (len(min('Z', 'BB')) /= 2) error stop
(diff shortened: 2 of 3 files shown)


YashNagda17 (approved):
LGTM

